PR #244094, opened 2026-01-15
[IMP] mail: show explicit error of fetch message in message list

alexkuhn:
Before this commit, when message list failed to load, it just displays a "Ann error occurred" generic message with a retry button.

This assumes that error happens rarely and when so this is temporarily. However some errors are persistent and it's frustrating to have no clue on why there's error or what may have caused it.

This commit shows the `Error.toString()` from fetch message RPC failure on UI, so that there's a clue on the reason the fetch of messages failed.

Before / After
Screenshot 2026-01-15 at 18 18 31 Screenshot 2026-01-15 at 18 17 44

### DIFF
--- a/addons/mail/static/src/core/common/thread_model.js
+++ b/addons/mail/static/src/core/common/thread_model.js
@@ -197,6 +197,8 @@ export class Thread extends Record {
     /** @type {String|undefined} */
     primary_email_field;
     hasLoadingFailed = false;
+    /** @type {Error} */
+    hasLoadingFailedError;
     canPostOnReadonly;
     /** @type {Boolean} */
     is_editable;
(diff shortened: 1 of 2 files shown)


tsm-odoo (approved):
@robodoo r+

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -380,9 +382,11 @@ export class Thread extends Record {
         let res;
         try {
             res = await this.fetchMessagesData({ after, around, before });
+            this.hasLoadingFailedCause = undefined;
```
alexkuhn, line 385: Will fix typo in forward-port of https://github.com/odoo/odoo/pull/244754

PR #244155, opened 2026-01-15
[FIX] account: apply tax when tax fiscal position is set to all

mma1377:
### Issue:
No tax will be applied if in taxes, fiscal position is set to all.

#### Steps to reproduce:
1- Create a tax, and in the tax form, leave `Fiscal Position` field blank, which in this case `all` will be shown in placeholder.
2- Set Domestic FP to be applied automatically, and set the country to `US`.
3- Create a Partner with `US` country_id.
4- Create a product, and apply the created tax to sale taxes.
5- Create a SO with created partner and the created product.
6- As you see, the tax is not applied to the line, while if you check SO's fiscal position, it is set to Domestic.

Expected: As tax's fp is set to all, we expect this tax being applied with Domestic fp.

### Cause:
In this line, if no `tax_ids` is set, it means fp has not tax_ids:
https://github.com/odoo/odoo/blob/0c3ae7f78d313885984c99a4e57485d9660dd974/addons/account/models/partner.py#L154-L158

However, this might also mean the tax has no fp because `fp.tax_ids` is a Many2Many relation.

In the forms, `tax.fiscal_position_ids` being empty is shown as `all` in the placeholder, which means when no tax applied to fp, we expect all taxes to be mapped.

### Fix:
This can be fixed by making sure the fp.tax_ids is not empty because there is no `tax.fiscal_position_ids` set.

### Remark:
In this fix we rename `TestInvoiceTaxes._create_invoice` to `_create_invoice_taxes_per_line` to avoid override of `AccountTestInvoicingCommon._create_invoice`. This is already done on saas-19.1+.

opw-5463245

### DIFF
--- a/addons/account/tests/test_invoice_taxes.py
+++ b/addons/account/tests/test_invoice_taxes.py
@@ -93,7 +93,7 @@ class TestInvoiceTaxes(AccountTestInvoicingCommon):
         base_tags = base_report_line.expression_ids._get_matching_tags()
         cls.base_tag_pos, cls.base_tag_neg = base_tags.sorted('tax_negate')
 
-    def _create_invoice(self, taxes_per_line, inv_type='out_invoice', currency_id=False, invoice_payment_term_id=False):
+    def _create_invoice_taxes_per_line(self, taxes_per_line, inv_type='out_invoice', currency_id=False, invoice_payment_term_id=False):
         ''' Create an invoice on the fly.
 
         :param taxes_per_line: A list of tuple (price_unit, account.tax recordset)
@@ -124,7 +124,7 @@ class TestInvoiceTaxes(AccountTestInvoicingCommon):
         --------------------------------------------
         21%         | /         | 100       | 21
         '''
-        invoice = self._create_invoice([(100, self.env['account.tax'])])
+        invoice = self._create_invoice_taxes_per_line([(100, self.env['account.tax'])])
         invoice.invoice_line_ids[0].tax_ids = self.percent_tax_1
         self.assertRecordValues(invoice.line_ids.filtered('tax_line_id'), [
             {'name': self.percent_tax_1.name, 'tax_base_amount': 100, 'balance': -21, 'tax_ids': []},
@@ -145,7 +145,7 @@ class TestInvoiceTaxes(AccountTestInvoicingCommon):
         21% incl    | /         | 100       | 21
         12%         | /         | 100       | 12
         '''
-        invoice = self._create_invoice([
+        invoice = self._create_invoice_taxes_per_line([
             (100, self.percent_tax_1),
             (121, self.percent_tax_1_incl),
             (100, self.percent_tax_2),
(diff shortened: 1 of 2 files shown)


kitan191 (approved):

addons/account/tests/test_invoice_taxes.py
```diff
@@ -929,3 +929,26 @@ def test_multiple_onchange_product_and_price(self):
...
+        self.env['account.fiscal.position'].create({
+            'name': 'FP',
+            'auto_apply': True,
+            'country_id': self.env.company.country_id.id,
+            'tax_ids': default_tax.ids,
+        })
+        self.assertEqual(self.product.taxes_id, default_tax)
+        default_tax.write({'fiscal_position_ids': self.env['account.fiscal.position']})
+        self.partner_a.write({'country_id': self.env.company.country_id.id})
+        move = self.init_invoice('out_invoice', self.partner_a, post=True, products=[self.product])
```
yosuanicolaus: `init_invoice` is deprecated. Use `_create_invoice` / `_create_invoice_one_line` instead.
```suggestion
        move = self._create_invoice_one_line(product_id=self.product,  partner_id=self.partner_a, post=True)
```

addons/account/tests/test_invoice_taxes.py
```diff
@@ -929,3 +929,26 @@ def test_multiple_onchange_product_and_price(self):
...
+        default_tax = self.company_data['default_tax_sale']
+        self.env['account.fiscal.position'].create({
+            'name': 'FP',
+            'auto_apply': True,
+            'country_id': self.env.company.country_id.id,
+            'tax_ids': default_tax.ids,
+        })
+        self.assertEqual(self.product.taxes_id, default_tax)
+        default_tax.write({'fiscal_position_ids': self.env['account.fiscal.position']})
+        self.partner_a.write({'country_id': self.env.company.country_id.id})
```
yosuanicolaus: how about?
```suggestion
        self.partner_a.country_id = self.env.company.country_id
```
(same for `default_tax` 's single write)

yosuanicolaus (changes requested):

addons/account/tests/test_invoice_taxes.py
mma1377: @yosuanicolaus 
`_create_invoice` does not cover my test as that explicitly set `tax_id` per line.
`_create_invoice_one_line` would have covered the test but `_create_invoice` override in here doesn't allow it to work. (Unless I add a default arg in `_create_invoice`, to allow calling parent implementation of `_create_invoice` instead)

yosuanicolaus: Ah yes, I mean the `_create_invoice` from `AccountTestInvoicingCommon`, not the `TestInvoiceTaxes`.

Can you rename the one in `TestInvoiceTaxes` to something else? I think it is renamed in `19.1`+ but the backport to 18.0 last time only brought back the methods to prevent unnecessary conflicts

yosuanicolaus (approved):
LGTM, sending to final review 👌

aboo-odoo (commented):
Hey 👋 

I have a small remark before move forward with the r+: could you mention in your commit message that you are also renaming the `_create_invoice` so that it does not override the `AccountTestInvoicingCommon:_create_invoice` method (I assume that's the reason you did it) ? Thank you

aboo-odoo (approved):
Thank you!

@robodoo r+

PR #244156, opened 2026-01-16
[IMP] website_event: Split registration new controller logic for extensibility

CarmenMiranda:
Since this controller returns raw markup, it is impossible to inherit.

By splitting the controller `registration_new`, allows to manage custom developments with the inheritance of the prepare method instead.

---


reth-odoo:
Note I cannot merge and tde is off this week so feel free to ping again later :)

PR #244158, opened 2026-01-16
[IMP] account,point_of_sale: Modified payment widget and exchange diff account

leml-odoo:
This PR targets two things:
- Improve the visualization of the exchanges on the payment widget
- Match the exchange account when an exchange entry is created for both caba and payment
reconciliations with an invoice

The invoice payment widget was modified to now show the total amount of exchange instead
of showing each exchange move separately and if it is gain or loss. It is possible to click on button to go and see 
the list of exchange. Also, now the exchange moves coming from CABA entries are also considered

Also, when a reconciliation with an invoice with caba taxes, has exchange difference. The exchange moves
created will share the same account

target: master
task-4814793

### DIFF
--- a/addons/account/models/account_move_line.py
+++ b/addons/account/models/account_move_line.py
@@ -2485,6 +2485,8 @@ class AccountMoveLine(models.Model):
             'debit_move_id': debit_aml.id,
             'credit_move_id': credit_aml.id,
         }
+        res['debit_aml'] = debit_aml
+        res['credit_aml'] = credit_aml
 
         debit_values['amount_residual'] = remaining_debit_amount
         debit_values['amount_residual_currency'] = remaining_debit_amount_curr
@@ -2810,6 +2812,7 @@ class AccountMoveLine(models.Model):
         partials_values_list = []
         exchange_diff_values_list = []
         all_plan_results = []
+        exchange_account_per_move = {}
         for plan in plan_list:
             plan_results = self\
                 .with_context(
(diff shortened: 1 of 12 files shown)


addons/account/models/account_move_line.py
```diff
@@ -2454,12 +2454,30 @@ def get_amount_range_after_rate(currency_from, currency_to, amount, rate):
...
+                ).account_id[:1]
+
                 res['exchange_values'] = exchange_lines_to_fix._prepare_exchange_difference_move_vals(
                     amounts_list,
                     exchange_date=max(
                         debit_aml._get_reconciliation_aml_field_value('date', shadowed_aml_values),
                         credit_aml._get_reconciliation_aml_field_value('date', shadowed_aml_values),
                     ),
+                    exchange_account=exchange_account,
+                    ref=','.join((debit_aml | credit_aml).move_id.filtered(lambda m: m.is_invoice() and m.name).mapped('name')),
```
smetl: I think you should do:
1) Add an optional `options` parameter to `_reconcile_plan` and add this parameter to every single method involved by the reconciliation. That way, we will have a nice way to propagate data and we won't need to add weird context key to do shitty fixes (`no_exchange_difference`, `no_exchange_difference_no_recursive`).
2) During the CABA, create new options indicating you are in the context of CABA: `{**options, 'cash_basis': True}`
3) All exchange differences are created in batch, see `exchange_diff_values_list`. Based on those dictionaries, populate `options` with the exchange account used per move.

What do you think?

smetl (commented):
shit I forgot to publish my review ~~

smetl (approved):
@robodoo r+

PR #224917, opened 2025-09-01
[ADD] l10n_lk

erle-odoo:
Implement a basic accounting package for Sri Lanka including:

- COA
- Tax groups
- Taxes
- Fiscal Positions
- Tax reports (VAT001, WHT001)

task-4352802

PR #244164, opened 2026-01-16
[IMP] l10n_th: add fiscal positions and update tax groups, and tax templates

avar-odoo:
Enhanced Thai localization by adding fiscal positions for various entity types, updating tax groups and rates, and improving tax configuration for compliance and usability.

Task-4689740

---


addons/l10n_th/data/template/account.tax-th.csv
vin-odoo, line 1: It's best to keep the xml id same if you can.
tax_wht_income_2 can stay  tax_wht_income_2, even if you rename it/...
As long as its the same tax, rate and tags.

We also always write the csv to be:
tax line + invoice base repartition line at once
invoice tax repartition line
refund base repartition line
refund tax repartition line

You should follow the same standard here. Less diff is always better if it can be avoided

addons/l10n_th/data/template/account.tax.group-th.csv
vin-odoo, line 1: Same idea here. You keep the ids, but you reorder the file.
It would be better to keep the existing order and just add any additional groups at the bottom.

vin-odoo (changes requested):

addons/l10n_th/data/template/account.tax.group-th.csv
kidoodoo: For this one, the order of the tax group matters on the order it is displayed on the invoice.
It is not a must, but in case of both being displayed, was hoping to follow VAT -> WHT order.

Is this something to be avoided?

vin-odoo: Wouldn't it be safe and more reliable to set a sequence?
It order this way because it automatically order by ID, but I wouldn't rely on that to ensure the order.

addons/l10n_th/data/template/account.tax-th.csv
```diff
@@ -1,73 +1,187 @@
...
+"","","","","","","","","100","tax","refund","l10n_th_account_213302","PND53","","","","","","","","","","","","",""
+"tax_wht_co_3","3% WH C SERV","WHT 3% (P.N.D. 53) Services / Hire of Work","","percent","-3","purchase","tax_group_3","100","base","invoice","","Income PND53","","l10n_th_fiscal_position_template_1","tax_wht_pers_3","True","service","True","False","True","on_invoice","","หัก ณ ที่จ่าย 3% (ภ.ง.ด. 53) ค่าบริการ/จ้างทำของ","","11"
+"","","","","","","","","100","base","refund","","Income PND53","","","","","","","","","","","","",""
+"","","","","","","","","100","tax","invoice","l10n_th_account_213302","PND53","","","","","","","","","","","","",""
+"","","","","","","","","100","tax","refund","l10n_th_account_213302","PND53","","","","","","","","","","","","",""
+"tax_wht_co_5","5% WH C RENT","WHT 5% (P.N.D. 53) Rentals","tax_group","percent","-5","purchase","tax_group_5","100","base","invoice","","Income PND53","","l10n_th_fiscal_position_template_1","tax_wht_pers_5","True","service","True","False","True","on_invoice","","หัก ณ ที่จ่าย 5% (ภ.ง.ด. 53) ค่าเช่า","","12"
```
kidoodoo: Found this label `tax_group`. This should be empty.

addons/l10n_th/data/template/account.fiscal.position-th.csv
```diff
@@ -0,0 +1,6 @@
+"id","name","auto_apply","country_id","name@th_TH",sequence
+"l10n_th_fiscal_position_template_1","Domestic Company","1","base.th","นิติบุคคลในประเทศ","1"
+"l10n_th_fiscal_position_template_2","Domestic Individual","1","base.th","บุคคลธรรมดาในประเทศ","2"
+"l10n_th_fiscal_position_template_3","Foreign Entity","1","base.th","นิติบุคคลต่างประเทศ","3"
+"l10n_th_fiscal_position_template_4","Foreign Individual","1","base.th","บุคคลธรรมดาต่างประเทศ","4"
```
kidoodoo: Foreign Entity and Foreign Individual should not have country set

addons/l10n_th/data/template/account.fiscal.position-th.csv
vin-odoo: You should only set the country on the domestic fiscal positions, otherwise it won't work.

addons/l10n_th/data/template/account.tax-th.csv
vin-odoo, line 1: You still don't respect the line order :)

It should follow the same structure as everywhere else in Odoo.

Image

vin-odoo (changes requested):

addons/l10n_th/data/template/account.tax-th.csv
erle-odoo: image

Oops, accidentally let my [l10n_lk](https://github.com/odoo/odoo/pull/224917/files#diff-209f54ca42b70d21cdf02b03d8c0ff980e7dd7c980fa54eedc9773263de56f53) slip through :sweat_smile:

vin-odoo: Indeed, oops. Not ideal, but at least that one was a new module, not a change in the existing file ^^

addons/l10n_th/data/template/account.fiscal.position-th.csv
vin-odoo, line 1: We should use the vat_required field.
Set it to true to only apply Domestic Company if the partner has a VAT set, otherwise default to Domestic Individual.

I think this would help make the automatic assignement a little better?

cc @kidoodoo

vin-odoo (changes requested):
Nearly LGTM. One small comment left

vin-odoo (approved):

vin-odoo:
robodoo r+

PR #244167, opened 2026-01-16
[PERF] pos_self_order: fix N+1 pricelist queries

juwu-odoo:
Currently when a pricelist is set on a POS session, N+1 queries are generated when loading the data by calling `_get_product_price()` in a loop.
This commit avoids the extra queries by performing the computation on the recordset using `_get_products_price()` and looping through the result instead.

This commit also generally cleans up the function by removing unnecessary intermediate variables, and removing the redundant product_obj check.

Benchmark opening /pos-self/data
| product.product count | Before | After | Queries Before | Queries After | 
| --------------------- | ------ | ----- | -------------- | ------------- |
| 1,500                 | 2.16s  | 0.72s | 1,711          | 185           |
| 15,000                | 24.91s | 6.52s | 17,175         | 397           |

opw-5477715

### DIFF
--- a/addons/pos_self_order/models/product_product.py
+++ b/addons/pos_self_order/models/product_product.py
@@ -94,23 +94,21 @@ class ProductProduct(models.Model):
 
     def _compute_product_price_with_pricelist(self, products, config_id):
         config = self.env['pos.config'].browse(config_id)
-        pricelist = config.pricelist_id
 
-        product_ids = [product['id'] for product in products]
-        product_objs = self.env['product.product'].browse(product_ids)
-
-        product_map = {product.id: product for product in product_objs}
-        loaded_product_tmpl_ids = list({p['product_tmpl_id'] for p in products})
-        archived_combinations = self._get_archived_combinations_per_product_tmpl_id(loaded_product_tmpl_ids)
+        archived_combinations = self._get_archived_combinations_per_product_tmpl_id(
+            list({p['product_tmpl_id'] for p in products})
+        )
+        lst_prices = config.pricelist_id._get_products_price(
+            self.env['product.product'].browse([product['id'] for product in products]),
+            quantity=1.0,
+            currency=config.currency_id,
+        )
 
         for product in products:
-            product_obj = product_map.get(product['id'])
-            if product_obj:
-                product['lst_price'] = pricelist._get_product_price(
-                    product_obj, 1.0, currency=config.currency_id
-                )
-            if archived_combinations.get(product['product_tmpl_id']):
-                product['_archived_combinations'] = archived_combinations[product['product_tmpl_id']]
+            if product['id'] in lst_prices:
+                product['lst_price'] = lst_prices[product['id']]
+            if archived_combination := archived_combinations.get(product['product_tmpl_id']):
+                product['_archived_combinations'] = archived_combination
 
     def _filter_applicable_attributes(self, attributes_by_ptal_id: Dict) -> List[Dict]:
         """


addons/pos_self_order/models/product_product.py
```diff
@@ -102,13 +102,12 @@ def _compute_product_price_with_pricelist(self, products, config_id):
...
 
         for product in products:
             product_obj = product_map.get(product['id'])
             if product_obj:
-                product['lst_price'] = pricelist._get_product_price(
-                    product_obj, 1.0, currency=config.currency_id
-                )
+                product['lst_price'] = lst_prices[product['id']][0]
             if archived_combinations.get(product['product_tmpl_id']):
                 product['_archived_combinations'] = archived_combinations[product['product_tmpl_id']]
```
pivi-odoo:
```suggestion
    def _compute_product_price_with_pricelist(self, products, config_id):
        config = self.env['pos.config'].browse(config_id)

        archived_combinations = self._get_archived_combinations_per_product_tmpl_id(
            list({product['product_tmpl_id'] for product in products})
        )

        product_ids_set = {product['id'] for product in products}
        product_objs = self.env['product.product'].browse(product_ids_set)
        lst_prices = config.pricelist_id._compute_price_rule(product_objs, 1.0, currency=config.currency_id)

        for product in products:
            product_id = product['id']
            if product_id in product_ids_set:
                product['lst_price'] = lst_prices[product_id][0]
            if archived_combination := archived_combinations.get(product['product_tmpl_id']):
                product['_archived_combinations'] = archived_combination
```

TBF: one can go further and and rework a bit the function. There are a lot of intermediate variables that are not necessary, and the `product_map` thing can be completely dropped. 
`_get_archived_combinations_per_product_tmpl_id` could take a `set` of ids directly, as it's just used in a `browse` which takes any iterable, so we could've skipped the `list()`, but in stable we can't change the signature (it's a type change of parameter, we don't know if there is custom code that expect this param to be a list only, and not any iterable). And in master this method doesn't exist anymore anyway...

pivi-odoo: Github's suggestion tool shat on itself... the proposed diff is the whole function def, if it's not really clear.

pivi-odoo (commented):
Good catch! @juwu-odoo 

The idea of the patch LGTM. I'm proposing a bit of cleanup while at it (this method is a mess), but that's up to you if you want or not. Let me know what you think.

addons/pos_self_order/models/product_product.py
juwu-odoo: I think the cleanup makes sense, I went ahead and removed the intermediate variables. The first `if` check seems to be completely redundant actually (the corresponding `project_obj` will always exist since we `browse()` every id), so I removed that as well. Also I found a cleaner function `_get_products_price()` instead of `_compute_price_rule()`

addons/pos_self_order/models/product_product.py
```diff
@@ -94,23 +94,20 @@ def _load_pos_self_data(self, data):
...
-
-        product_map = {product.id: product for product in product_objs}
-        loaded_product_tmpl_ids = list({p['product_tmpl_id'] for p in products})
-        archived_combinations = self._get_archived_combinations_per_product_tmpl_id(loaded_product_tmpl_ids)
+        archived_combinations = self._get_archived_combinations_per_product_tmpl_id(
+            list({p['product_tmpl_id'] for p in products})
+        )
+        lst_prices = config.pricelist_id._get_products_price(
+            self.env['product.product'].browse([product['id'] for product in products]),
+            1.0,
```
pivi-odoo:
```suggestion
            quantity=1.0,
```
Just for clarity, to avoid having a magic `1.0` that isn't clear what it's referring to.

pivi-odoo (approved):
LGTM now, thank you for your work @juwu-odoo.

Just left a last little suggestion, mostly for readability, but can be left as is if you want. :)

The forward port will probably be rough, I didn't manage to quickly see in which version this was removed/moved/refactored, so you should probably check around, to know up to which version this should be fw-ported.

@manv-afk This perf fix has already been pre-reviewed internally.
When you have the time to review this. Thank you in advance.

addons/pos_self_order/models/product_product.py
```diff
@@ -94,23 +94,20 @@ def _load_pos_self_data(self, data):
...
 
         for product in products:
-            product_obj = product_map.get(product['id'])
-            if product_obj:
-                product['lst_price'] = pricelist._get_product_price(
-                    product_obj, 1.0, currency=config.currency_id
-                )
-            if archived_combinations.get(product['product_tmpl_id']):
-                product['_archived_combinations'] = archived_combinations[product['product_tmpl_id']]
+            product['lst_price'] = lst_prices[product['id']]
```
manv-afk: don't you think we should make this check to avoid keyError?
```suggestion
            if product['id'] in lst_prices:
                product['lst_price'] = lst_prices[product['id']]
```

manv-afk (changes requested):

addons/pos_self_order/models/product_product.py
juwu-odoo: `_get_products_price()` *should* always return an entry for each id, but I guess it never hurts to be sure :)

manv-afk (approved):

davidmonnom:
@robodoo r+

PR #244177, opened 2026-01-16
[FIX] pos_razorpay: prevent sudo attribute error during Razorpay request

paan-odoo:
Steps:

- Configure a Razorpay payment method in POS with Indian localization.
- Open a POS session and process an order.
- Start the Razorpay transaction.

Issue:

- An error pop-up is shown: 'RazorpayPosRequest' object has no attribute 'sudo'.

Cause:

- The code attempts to call sudo() on RazorpayPosRequest, which is not a model and does not support sudo.

Fix:

- Remove the unnecessary sudo() access from RazorpayPosRequest.

Task-5501634


paso-odoo (approved):

stva-odoo:
robodoo r+

PR #244180, opened 2026-01-16
[FIX] {purchase,sale_purchase}_stock: avoid overcounting of monthly demand

pkgu-odoo:
**Steps to reproduce:**
* Install *purchase_stock* and *sale_management* module.
* Go to *Inventory > Configuration > Settings*. Enable *Multi-Step Routes*.
* Go to *Inventory > Configuration > Warehouses*.
* Set the warehouse delivery flow to *Pick + Pack + Ship (3 steps)*.
* Create a new product. Under the *Purchase* tab, add a vendor.
* Create a sales order for this product with some quantity.
* Confirm the sales order. Validate all three generated transfers (*Pick*, *Pack*, *Ship*).
* Create a purchase order for the same vendor.
* In the purchase order line, click *Catalog* and search for the product.

**Observed behavior:**
* In the catalog view, the *Monthly Demand* is shown as *3x* 
the original sales order quantity instead of the actual demand.

**Cause:**
* *Monthly Demand* is a computed field using `_compute_monthly_demand`, 
which relies on `_get_monthly_demand_moves_location_domain()`.
* In a 3-step delivery flow, all related moves have a final location with usage set to *customer*.
* The domain condition: `('location_final_id.usage', 'in', ['customer', 'production'])`
counts all intermediate pickings.
* Additionally, the fallback condition: `[('location_final_id.warehouse_id', '!=', warehouse_id)]`
is always true because *customer* locations are not linked to a warehouse.
* As a result, all three pickings are counted, inflating the demand. See: https://github.com/odoo/odoo/blob/6bbaea728dbcba49776e813c70dff649d041bdc9/addons/purchase_stock/models/product.py#L144-L157

**Fix:**
* Prevent counting intermediate pickings in 3-step delivery by 
restricting the domain to moves with `move_dest_ids = False`,
ensuring only the final move is considered for monthly demand computation.

---

opw-5453991

---

### DIFF
--- a/addons/purchase_stock/models/product.py
+++ b/addons/purchase_stock/models/product.py
@@ -141,10 +141,14 @@ class ProductProduct(models.Model):
             (We don't include returns in demand estimation - they come back on hand)
         """
         warehouse_id = self.env.context.get('warehouse_id')
+        non_return_moves_domain = ['!', ('move_dest_ids.origin_returned_move_id', '=', False)]
         if not warehouse_id:
-            return Domain.OR([
-                [('location_dest_usage', 'in', ['customer', 'production'])],
-                [('location_final_id.usage', 'in', ['customer', 'production'])],
+            return Domain.AND([
+                Domain.OR([
+                    [('location_dest_usage', 'in', ['customer', 'production'])],
+                    [('location_final_id.usage', 'in', ['customer', 'production'])],
+                ]),
+                non_return_moves_domain,
             ])
         else:
             return Domain.AND([
@@ -153,7 +157,8 @@ class ProductProduct(models.Model):
                     [('location_dest_id.warehouse_id', '!=', warehouse_id)],
                     [('location_final_id.warehouse_id', '!=', warehouse_id)]
                 ]),  # includes moves going to customer or production
-                [('location_dest_id.usage', '!=', 'inventory')]  # exclude scrap
+                [('location_dest_id.usage', '!=', 'inventory')],  # exclude scrap
+                non_return_moves_domain,
             ])
 
     def _get_quantity_in_progress(self, location_ids=False, warehouse_ids=False):
--- a/addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
+++ b/addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
@@ -2,9 +2,10 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from datetime import timedelta
+from dateutil.relativedelta import relativedelta
 
 from odoo import Command, fields
-from odoo.tests import Form, TransactionCase
+from odoo.tests import Form, TransactionCase, freeze_time
 
 
 class TestSalePurchaseStockFlow(TransactionCase):
@@ -558,3 +559,26 @@ class TestSalePurchaseStockFlow(TransactionCase):
         sale_order.action_confirm()
         purchase_order = sale_order._get_purchase_orders()
         self.assertEqual(purchase_order.order_line.analytic_distribution, {str(analytic_account.id): 100})
+
+    def test_product_monthly_demand(self):
+        """
+        Test monthly demand is counted once in a multi-step delivery flow.
+        """
+        self.warehouse.delivery_steps = 'pick_pack_ship'
+        with freeze_time(fields.Datetime.now() - relativedelta(days=1)):
+            so = self.env['sale.order'].create({
+                'partner_id': self.customer.id,
+                'warehouse_id': self.warehouse.id,
+                'order_line': [Command.create({
+                    'product_id': self.mto_product.id,
+                    'product_uom_qty': 10,
+                })],
+            })
+            so.action_confirm()
+            so.picking_ids[0].move_ids.quantity = 10
+            so.picking_ids[0].button_validate()
+            so.picking_ids[1].move_ids.quantity = 10
+            so.picking_ids[1].button_validate()
+            so.picking_ids[2].move_ids.quantity = 10
+            so.picking_ids[2].button_validate()
+        self.assertEqual(self.mto_product.monthly_demand, 10.0)


aksi-odoo (approved):

addons/purchase_stock/models/product.py
```diff
@@ -142,9 +142,12 @@ def _get_monthly_demand_moves_location_domain(self):
...
         if not warehouse_id:
-            return Domain.OR([
-                [('location_dest_usage', 'in', ['customer', 'production'])],
-                [('location_final_id.usage', 'in', ['customer', 'production'])],
+            return Domain.AND([
+                Domain.OR([
+                    [('location_dest_usage', 'in', ['customer', 'production'])],
+                    [('location_final_id.usage', 'in', ['customer', 'production'])],
+                ]),
+                [('move_dest_ids', '=', False)]
```
Sehnde:
```suggestion
                [('move_dest_ids', '=', False)],
```
you can already put a comma at the end. That way if another argument needs to be added,  the diff will only be on 1 line and not 2.

addons/purchase_stock/models/product.py
```diff
@@ -153,7 +156,8 @@ def _get_monthly_demand_moves_location_domain(self):
                     [('location_dest_id.warehouse_id', '!=', warehouse_id)],
                     [('location_final_id.warehouse_id', '!=', warehouse_id)]
                 ]),  # includes moves going to customer or production
-                [('location_dest_id.usage', '!=', 'inventory')]  # exclude scrap
+                [('location_dest_id.usage', '!=', 'inventory')],  # exclude scrap
+                [('move_dest_ids', '=', False)]
```
Sehnde:
```suggestion
                [('move_dest_ids', '=', False)],
```
same here

addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
```diff
@@ -535,3 +536,27 @@ def test_mto_cancel_multi_steps_confirmed_purchase(self):
         self.assertFalse(sale_orders.picking_ids.move_ids.move_orig_ids)
         sale_orders.picking_ids.action_assign()
         self.assertListEqual(sale_orders.picking_ids.move_ids.mapped('quantity'), [1.0, 1.0])
+
+    def test_product_monthly_demand(self):
+        """
+        Test monthly demand is counted once in a 3-step delivery flow.
+        """
+        self.env['res.config.settings'].write({'group_stock_adv_location': True})
```
Sehnde:
```suggestion
```
you don't need to activate that group if it's just to be able to set 3-steps delivery in a test.

Sehnde (commented):
only nitpicks, otherwise LGTM

Sehnde (approved):

addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
```diff
@@ -535,3 +536,26 @@ def test_mto_cancel_multi_steps_confirmed_purchase(self):
         self.assertFalse(sale_orders.picking_ids.move_ids.move_orig_ids)
         sale_orders.picking_ids.action_assign()
         self.assertListEqual(sale_orders.picking_ids.move_ids.mapped('quantity'), [1.0, 1.0])
+
+    def test_product_monthly_demand(self):
+        """
+        Test monthly demand is counted once in a 3-step delivery flow.
+        """
+        self.warehouse.delivery_steps = 'pick_pack_ship'
+        with freeze_time(fields.Datetime.now() - relativedelta(days=1)):
```
clesgow, line 568: Since `_get_monthly_demand_range` will be on 30 days by default, why do you need this `freeze_time`? 🤔

addons/purchase_stock/models/product.py
```diff
@@ -142,9 +142,12 @@ def _get_monthly_demand_moves_location_domain(self):
...
         if not warehouse_id:
-            return Domain.OR([
-                [('location_dest_usage', 'in', ['customer', 'production'])],
-                [('location_final_id.usage', 'in', ['customer', 'production'])],
+            return Domain.AND([
+                Domain.OR([
+                    [('location_dest_usage', 'in', ['customer', 'production'])],
+                    [('location_final_id.usage', 'in', ['customer', 'production'])],
+                ]),
+                [('move_dest_ids', '=', False)],
```
clesgow: If your delivery to the customer has a return, then it will have a `move_dest_ids`, and be excluded from the domain and not count towards the `monthly_demand`.

dhha-odoo: Hello @clesgow 

Just to inform you, there is already an ongoing [task](https://www.odoo.com/odoo/project/966/tasks/5490137) in our project that handles this issue. If I understood your scenario correctly, it is working fine there. Here is a [video](https://www.awesomescreenshot.com/video/50137691?key=e52ee3862815e17d68f754f297cd08ec) demonstrating it. However, in this PR, if we follow the same steps, the monthly_demand is not currently calculated.

Thanks.

addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
pkgu-odoo: In this function `_compute_monthly_demand`
https://github.com/odoo/odoo/blob/72be98d705e225f663b65e289e11d0b8642ec6f8/addons/purchase_stock/models/product.py#L115-L116
"Give all moves where date >= start_date(30 days ago) AND date < limit_date(datetime.now())"
https://github.com/odoo/odoo/blob/72be98d705e225f663b65e289e11d0b8642ec6f8/addons/purchase_stock/models/product.py#L217

Example - 
10:35:00.100  →  picking.button_validate()  →  move.date = "10:35:00.100"
10:35:00.101  →  picking.button_validate()  →  move.date = "10:35:00.101"
10:35:00.102  →  picking.button_validate()  →  move.date = "10:35:00.102"

10:35:00.102  →  _compute_monthly_demand runs
                  limit_date = datetime.now() = "10:35:00.102"

Domain: date(10:35:00.102) <  limit_date (10:35:00.102)
→ move with date "10:35:00.102" is EXCLUDED! ❌

The last move's  `date`
 can equal `limit_date = datetime.now()`. Since the filter is strictly less than, it gets excluded. The test fails (a race condition based on timing).

addons/purchase_stock/models/product.py
pkgu-odoo: Hello @clesgow,
Thank you for your review. 
I have added a check to exclude all return moves.

addons/sale_purchase_stock/tests/test_sale_purchase_stock_flow.py
clesgow: Indeed, noticed the domain too late. I guess it's fine as is then. 👍

PR #244183, opened 2026-01-16
[FIX] l10n_nl: correct NL Domestic fiscal position definition

hash-odoo:
Installing the Netherlands localisation creates two Domestic fiscal positions, both incorrectly configured. This commit removes the empty duplicate fiscal position and properly defines the NL Domestic fiscal position by setting the Country Group, leaving Country empty, and disabling VAT requirement.

task-5489829


epictete (approved):
@robodoo r+

PR #244185, opened 2026-01-16
[IMP] product, *: simplify product variant management

khaj-odoo:
* = [stock, website_sale, hr_expense, mrp, mrp_account, point_of_sale, purchase,
     repair, sale, sale_gelato, stock_account]

Product variant handling is streamlined across the backend and the eCommerce
to reduce duplication and improve overall usability. 
- Remove the variant easy-edit form and all its dependent view inheritances 
  across modules   
- Remove standalone `Product Variants` menus from Sales, Purchase, Stock,
  POS, Repair, and Website, centralizing management on product templates
- Move product_tag_ids to the General Information tab and hide it on the variant
  form
- Add additional_product_tag_ids to the variant form
- Switch product_template_variant_value_ids to many2many_tags_color_dot widget
  across all variant list and form views
- Simplify variant list views with clearer columns, consistent ordering,
  and improved attribute readability
- Remove website_ribbon_id from product.product, ribbon now comes from
  the template only
- Normalize attribute naming by removing redundant prefixes from
   display_name and user-facing descriptions
- Always show weight/volume fields on the template form (remove conditional
  invisibility based on variant count)
- Improve stock visibility by introducing a quantity-at-date widget on variant
  lists in website_sale_stock
- Cleaned up obsolete views, menus, and overrides made redundant by the
  simplified workflow
 
This aligns product variant management with a template-first approach,
reduces cross-module complexity, and provides a cleaner, more consistent
user experience across applications.

See Also:
Enterprise PR:https://github.com/odoo/enterprise/pull/104548
Upgrade PR:https://github.com/odoo/upgrade/pull/9276

---


addons/product/views/product_views.xml
```diff
@@ -7,7 +7,7 @@
         <field name="arch" type="xml">
             <list string="Update Quantities" editable="bottom" create="false">
                 <field name="display_name" readonly="1"/>
-                <field name="product_template_variant_value_ids" widget="many2many_tags" groups="product.group_product_variant" readonly="1"/>
+                <field name="product_template_variant_value_ids" widget="many2many_tags" groups="product.group_product_variant" string="Attributes" readonly="1"/>
```
kcv-odoo: I would directly update in field definition in python

addons/product/views/product_views.xml
```diff
@@ -55,11 +55,10 @@
                         class="oe_avatar"
                         options="{'convert_to_webp': True, 'preview_image': 'image_128'}"
                     />
-                    <div class="oe_title">
+                    <div class="oe_title" style="margin-bottom: 20px;">
                         <label for="name" string="Product"/>
                         <h1>
                             <div class="d-flex">
-                                <field name="is_favorite" widget="boolean_favorite" class="me-2" nolabel="1"/>
```
kcv-odoo: Where this field will be used now?

addons/product/views/product_views.xml
```diff
@@ -55,11 +55,10 @@
                         class="oe_avatar"
                         options="{'convert_to_webp': True, 'preview_image': 'image_128'}"
                     />
-                    <div class="oe_title">
+                    <div class="oe_title" style="margin-bottom: 20px;">
```
kcv-odoo: It would make more sense to add this in product_normal_form_view instead here as it is causing issue there also it would be better to use bootstrap mb-2 it fine!

addons/product/views/product_views.xml
```diff
@@ -494,29 +410,72 @@
...
-                    <field name="additional_product_tag_ids" widget="many2many_tags" optional="hide"/>
-                    <field name="type" optional="hide" readonly="1"/>
-                    <field name="qty_available" invisible="not is_storable" string="On Hand" readonly="1" optional="show"
-                           decoration-danger="virtual_available &lt; 0" decoration-warning="virtual_available == 0" decoration-bf="1"
-                           sum="Total On Hand"
+                    <field
+                        name="product_template_variant_value_ids"
+                        widget="many2many_tags"
+                        groups="product.group_product_variant"
+                        string="Attributes"
```
kcv-odoo: Not need if we do it directly on field definition, same for other place where we could update field string directly in python 🙈

addons/product/views/product_views.xml
```diff
@@ -494,29 +410,72 @@
             <field name="model">product.product</field>
             <field eval="7" name="priority"/>
             <field name="arch" type="xml">
-                <list string="Product Variants" multi_edit="1" duplicate="false" sample="1" default_order="is_favorite desc, default_code, name, id">
-                    <field name="is_favorite" widget="boolean_favorite" nolabel="1" readonly="1"/>
```
kcv-odoo, line 504: is_favorite is removed from list and form view but still reference in kanban view, shouldn't we remove it from there also? are we removing whole feature of is_favorite field as we are removing from views?

addons/product/views/product_views.xml
```diff
@@ -571,13 +530,10 @@
                     <attribute name="string">Product Variant</attribute>
                     <attribute name="duplicate">false</attribute>
                 </form>
-                <xpath expr="//div[@name='standard_price_uom']" position="after">
+                <xpath expr="//field[@name='categ_id']" position="after">
```
kcv-odoo:
```suggestion
                <field name="categ_id" position="after">
```

addons/product/views/product_attribute_value_views.xml
```diff
@@ -32,7 +32,7 @@
                                 widget="color"
                                 invisible="display_type != 'color' or image"
                             />
-                            <field name="default_extra_price"/>
+                            <field name="default_extra_price" invisible="1"/>
```
kcv-odoo: Why make it invisible and not remove it?

addons/product/models/product_template_attribute_value.py
```diff
@@ -164,7 +164,7 @@ def _compute_display_name(self):
         Eg. on exclusion rules form
         """
         for value in self:
-            value.display_name = f"{value.attribute_id.name}: {value.name}"
+            value.display_name = value.name
```
kcv-odoo: I don't think we need this override if we remove this 👀

addons/product/models/product_attribute_custom_value.py
```diff
@@ -20,6 +20,7 @@ class ProductAttributeCustomValue(models.Model):
     def _compute_name(self):
         for record in self:
             name = (record.custom_value or '').strip()
-            if record.custom_product_template_attribute_value_id.display_name:
-                name = "%s: %s" % (record.custom_product_template_attribute_value_id.display_name, name)
+            ptav = record.custom_product_template_attribute_value_id
+            if ptav:
```
kcv-odoo:
```suggestion
            if ptav := record.custom_product_template_attribute_value_id:
```
Nit

addons/product/models/product_attribute_custom_value.py
```diff
@@ -20,6 +20,7 @@ class ProductAttributeCustomValue(models.Model):
...
+                name = f"{ptav.attribute_id.name}: {ptav.name}: {name}"
```
kcv-odoo:
```suggestion
                name = f"{ptav.attribute_id.display_name}: {ptav.name}: {name}"
```

addons/product/models/product_template.py
```diff
@@ -394,17 +394,21 @@ def _search_virtual_available(self, operator, value):
 
     @api.depends('product_variant_ids.volume')
     def _compute_volume(self):
-        self._compute_template_field_from_variant_field('volume')
+        for t in self:
+            t.volume = t.product_variant_ids.volume if len(t.product_variant_ids) == 1 else t.volume
```
kcv-odoo: why we need to do this changes? 🤔

addons/stock/views/product_views.xml
```diff
@@ -209,6 +200,7 @@
                 <xpath expr="//group[@name='group_lots_and_weight']" position="before">
                     <field name="has_available_route_ids" invisible="1"/>
                     <group string="Operations" name="operations" invisible="not (has_available_route_ids or route_from_categ_ids)">
+                        <field name="responsible_id" domain="[('share', '=', False)]" widget="many2one_avatar_user" groups="stock.group_stock_user"/>
```
kcv-odoo: Don't we need to remove this field from template only form field?

addons/account/views/product_view.xml
```diff
@@ -98,7 +98,7 @@
                         <field name="tax_string" class="oe_inline"/>
                     </div>
                 </div>
-                <div name="standard_price_uom" position="after">
+                <xpath expr="//div[@name='standard_price_uom']" position="after">
```
kcv-odoo:
```suggestion
                <div name="standard_price_uom" position="after">
```

addons/website_sale_stock/static/src/widgets/product_qty_at_date_widget.js
```diff
@@ -0,0 +1,94 @@
...
+
+export const productQtyAtDateWidget = {
+    component: ProductQtyAtDateWidget,
+    fieldDependencies: [
+        { name: "qty_available", type: "float" },
+        { name: "virtual_available", type: "float" },
+    ],
+};
+
+registry.category("view_widgets").add("product_qty_at_date_widget", productQtyAtDateWidget);
```
kcv-odoo: Why we need this widget? can't we use or adept existing one?

kcv-odoo (changes requested):
Thank you for your work! Added some comments

addons/product/models/product_template.py
khaj-odoo: Needed because we removed the invisible condition on weight/volume fields (they're now always visible on template form). The inline logic keeps the current template value when multiple variants exist, instead of resetting to 0 via the helper method. This avoids data loss when editing template weight/volume with multiple variants.

For example.
1.Set 10 kg 100m3 on template
2.Modified 1 variant with 50m3
3. Template resets to  0.00 m3, it should leave 100

addons/product/views/product_views.xml
khaj-odoo: I kept it in the kanban, it's needed for the "Add to Favorites" feature in the product catalog (sales order line). It has been removed from the variant list, variant form, and regular variant kanban views as given in task. The template views remain unchanged.

addons/product/models/product_template.py
kcv-odoo: use product_variant_count

addons/product/models/product_template.py
```diff
@@ -394,17 +394,21 @@ def _search_virtual_available(self, operator, value):
...
 
     def _set_volume(self):
-        self._set_product_variant_field('volume')
+        for t in self:
+            t.product_variant_ids.volume = t.volume
```
kcv-odoo: What if here product_variant_ids has more then on value?

addons/product/views/product_views.xml
```diff
@@ -385,98 +384,15 @@
...
-                                <field
-                                    name="uom_ids"
-                                    widget="many2many_uom_tags"
-                                    options="{'no_quick_create': True, 'on_tag_click': 'open_form'}"
-                                    groups="uom.group_uom"
-                                    force_save="1"
-                                    readonly="1"
-                                    context="{'product_id': id}"/>
-                                <field name="product_tag_ids" widget="many2many_tags" readonly="1"/>
-                                <field name="additional_product_tag_ids" widget="many2many_tags" context="{'product_variant_id': id}"/>
```
kcv-odoo: I guess we also need to remove this fields and related logic from Product variant `product.product`. Wdyt?

kcv-odoo (changes requested):

addons/product/models/product_template.py
khaj-odoo: This is intentional as per the task spec, setting volume/weight on the template should update all variants, not just single-variant products.

addons/product/views/product_views.xml
```diff
@@ -169,6 +175,7 @@
                                         </span>
                                     </div>
                                     <field name="categ_id" string="Category"/>
+                                    <field name="product_tag_ids" widget="many2many_tags" context="{'product_template_id': id}"/>
```
kcv-odoo: /> 100

addons/product/views/product_views.xml
```diff
@@ -385,98 +388,15 @@
...
             <field name="res_model">product.product</field>
-            <field name="context">{'search_default_product_tmpl_id': [active_id], 'default_product_tmpl_id': active_id, 'create': False}</field>
+            <field name="domain">[('product_tmpl_id', '=', active_id)]</field>
+            <field name="context">{'default_product_tmpl_id': active_id, 'create': False}</field>
             <field name="search_view_id" ref="product_search_form_view"/>
             <field name="view_ids"
                    eval="[(5, 0, 0),
                           (0, 0, {'view_mode': 'list'}),
-                          (0, 0, {'view_mode': 'form', 'view_id': ref('product_variant_easy_edit_view')}),
                           (0, 0, {'view_mode': 'kanban'})]"/>
```
kcv-odoo: We could remove this and just add view_type in this action as we are not opening any special form from this action now, wdyt?

addons/product/views/product_views.xml
```diff
@@ -495,29 +415,71 @@
...
+                    />
+                    <field
+                        name="virtual_available"
+                        invisible="not is_storable"
+                        string="Forecasted"
+                        readonly="1"
+                        optional="hide"
+                        decoration-danger="virtual_available &lt; 0"
+                        decoration-warning="virtual_available == 0"
+                        sum="Total Forecasted"
```
kcv-odoo, line 474: Re-order the attributes, and other places as well where you updated the code

addons/product/views/product_views.xml
```diff
@@ -495,29 +415,71 @@
...
                     />
                     <field name="uom_id" groups="uom.group_uom" optional="show" readonly="1"/>
                     <field name="product_tmpl_id" readonly="1" column_invisible="True"/>
                     <field name="active" column_invisible="True"/>
```
kcv-odoo: Can you check do we need this invisible fields? same for other above `currency_id` and others

addons/product/views/product_views.xml
kcv-odoo: Not sure this view is used anywhere 🤔 can you check?

addons/product/views/product_views.xml
```diff
@@ -592,18 +553,28 @@
                 <field name="name" position="after">
                     <field name="product_tmpl_id" class="oe_inline" readonly="1" invisible="1" required="id"/>
                 </field>
+                <xpath expr="//div[hasclass('oe_title')]" position="attributes">
+                    <attribute name="style">margin-bottom: 20px;</attribute>
+                </xpath>
```
kcv-odoo: Use bootstrap and add this on below field so it would only apply when below field is visible and not add so much space when it's not there

addons/product/views/product_views.xml
```diff
@@ -592,18 +553,28 @@
...
                 <xpath expr="//div[hasclass('oe_title')]" position="inside">
-                    <field name="product_template_variant_value_ids" widget="many2many_tags" readonly="1" invisible="not product_template_variant_value_ids" groups="product.group_product_variant"/>
+                    <field
+                        name="product_template_variant_value_ids"
+                        widget="many2many_tags_color_dot"
+                        options="{'color_field': 'html_color'}"
+                        style="font-size: 1rem;"
```
kcv-odoo: use bootstrap instead fs-6 would do the job 🤔

addons/product/views/product_views.xml
```diff
@@ -592,18 +553,28 @@
...
+                        groups="product.group_product_variant"
+                        invisible="not product_template_variant_value_ids"
+                        readonly="1"
+                    />
                 </xpath>
                 <field name="product_tag_ids" position="attributes">
```
kcv-odoo, line 564: we could just remove it from this view, by possition replace

addons/product/models/product_template.py
```diff
@@ -394,17 +394,21 @@ def _search_virtual_available(self, operator, value):
 
     @api.depends('product_variant_ids.volume')
     def _compute_volume(self):
-        self._compute_template_field_from_variant_field('volume')
+        for t in self:
+            t.volume = t.product_variant_ids.volume if t.product_variant_count == 1 else t.volume
```
kcv-odoo: - I think it's better to use _compute_template_field_from_variant_field in this for loop and pass t.volume as default so it would cover all the cases
- Also using template as variable would be better instead just t
- same for weight


Update, we could just `_compute_template_field_from_variant_field` for non multi variant products, wdyt

something like
``` 
self.filtered(
    lambda t: t.product_variant_count < 2
)._compute_template_field_from_variant_field('volume')
```

addons/product/models/product_template.py
kcv-odoo: We probably have to handle this case

```
elif count == 0:
      archived_variants = self.with_context(active_test=False).product_variant_ids
      if len(archived_variants) == 1:
          archived_variants[fname] = template[fname]
```

If product have no active variant, may try to use `_set_product_variant_field` somehow or update that method to take this case in account 🤔

addons/website_sale/models/product_template.py
kcv-odoo, line 1402: Need to adept this function as well as there is not ribbon on variant now 🤔 and other related code

addons/website_sale/views/product_views.xml
```diff
@@ -67,12 +41,8 @@
             <list position="attributes">
                 <attribute name="default_order">website_sequence</attribute>
             </list>
-            <field name="is_favorite" position="before">
-                <field name="website_sequence" widget="handle"/>
-            </field>
             <field name="website_id" position="after">
                 <field name="public_categ_ids" widget="many2many_tags" string="Categories" optional="show"/>
-                <field name="is_published" string="Is Published" optional="hide"/>
```
kcv-odoo, line 75: This would be useful is user want to make multiple products published/unpublished from multi edit. Do we really want to remove this? @Jean-boje

addons/website_sale/views/product_views.xml
```diff
@@ -67,12 +41,8 @@
             <list position="attributes">
                 <attribute name="default_order">website_sequence</attribute>
             </list>
-            <field name="is_favorite" position="before">
-                <field name="website_sequence" widget="handle"/>
```
kcv-odoo, line 71: Then how user will be able to re-sequence products for their website? @Jean-boje

addons/website_sale/views/product_views.xml
```diff
@@ -240,29 +206,36 @@
         </field>
     </record>
 
-    <record id="product_product_view_form_easy_inherit_website_sale" model="ir.ui.view">
-        <field name="name">product.product.view.form.easy.inherit.website_sale</field>
+    <record id="product_normal_form_view_inherit_website_sale" model="ir.ui.view">
```
kcv-odoo:
```suggestion
    <record id="product_normal_form_view model="ir.ui.view">
```
make name simpler while we are at it, since it xml_id actually point to `website_sale`

addons/website_sale/views/product_views.xml
```diff
@@ -240,29 +206,36 @@
...
+                    add-label="Add Media"
+                    nolabel="1"
+                    colspan="2"
+                    widget="x2_many_media_viewer"
+                    options="{'convert_to_webp': True}"
+                />
+            </field>
+            <group name="product_template_images" position="attributes">
+                <attribute name="string">Extra Variant Media</attribute>
             </group>
```
kcv-odoo: Feels not right
- Making field readonly just in Ecom
- changing field product_template_image_ids to product_variant_image_ids in ecom

may be issue is duplication in `Extra info` and `Ecommerce` page, I guess it would be better to just move wanted fields and replace unwanted page 🤔

addons/mrp/views/mrp_bom_views.xml
```diff
@@ -24,7 +24,12 @@
                         </div>
                         <field name="operation_id" groups="mrp.group_mrp_routings" options="{'no_quick_create':True,'no_create_edit':True}"/>
                         <field name="possible_bom_product_template_attribute_value_ids" invisible="1"/>
-                        <field name="bom_product_template_attribute_value_ids" widget="many2many_tags" options="{'no_create': True}" groups="product.group_product_variant"/>
+                        <field
+                            name="bom_product_template_attribute_value_ids"
+                            widget="many2many_tags_color_dot"
+                            options="{'no_create': True, 'color_field': 'html_color'}"
+                            groups="product.group_product_variant"
+                        />
```
kcv-odoo: Why are we change it here? is it to be consistent? Not sure it will add much value for BOM 🤔

kcv-odoo (changes requested):
Almost there... 
Thank you for the changes

shrm-odoo (commented):
Hello @khaj-odoo 
I see there is a conflict. Could you please rebase your branch and resolve it?

addons/product/views/product_attribute_value_views.xml
```diff
@@ -32,7 +32,6 @@
                                 widget="color"
                                 invisible="display_type != 'color' or image"
                             />
-                            <field name="default_extra_price"/>
```
shrm-odoo, line 35: Why was this removed? I couldn’t find any mention of it in the specifications.

addons/product/views/product_attribute_views.xml
```diff
@@ -68,8 +68,8 @@
                                        invisible="image"/>
                                 <field name="image"
                                        widget="image"
-                                       options="{'size': [70, 70]}"
-                                       class="oe_avatar text-start float-none"
```
shrm-odoo, line 72: Why was this class removed? Wouldn’t simply adjusting the image size through the options be enough to reduce its size?

addons/product/views/product_template_views.xml
```diff
@@ -110,10 +110,10 @@
                             <field
                                     name="value_ids"
                                     placeholder="List of possible values (e.g. Blue, Green, White, ... )"
-                                    widget="many2many_tags"
+                                    widget="many2many_tags_color_dot"
                                     options="{
-                                        'on_tag_click': 'open_form',
-                                        'color_field': 'color',
+                                        'can_open': True,
```
shrm-odoo, line 116: Why? I couldn’t find any mention of it in the specifications 🤔

addons/product/views/product_template_views.xml
```diff
@@ -126,11 +126,6 @@
                                 groups="product.group_product_variant"/>
                         </list>
                     </field>
-                        <p class="opacity-50 oe_edit_only">
-                        <strong>Warning</strong>: adding or deleting attributes
-                        will delete and recreate existing variants and lead
-                        to the loss of their possible customizations.
-                    </p>
```
shrm-odoo, line 134: Same question as above.

addons/product/views/product_views.xml
```diff
@@ -556,13 +491,23 @@
                     <attribute name="string">Product Variant</attribute>
                     <attribute name="duplicate">false</attribute>
                 </form>
-                <xpath expr="//div[@name='standard_price_uom']" position="after">
+                <field name="is_favorite" position="replace"/>
```
shrm-odoo, line 524: Same question as above

addons/product/views/product_views.xml
```diff
@@ -583,23 +528,19 @@
...
                         name="additional_product_tag_ids"
-                        widget="many2many_tags"
                         invisible="product_variant_count == 1"
+                        widget="many2many_tags"
                     />
                 </field>
-                <field name="uom_ids" position="attributes">
-                    <attribute name="context">{'product_id': id}</attribute>
-                </field>
+                <field name="uom_ids" position="replace"/>
```
shrm-odoo: Same question 😅

addons/product/views/product_views.xml
```diff
@@ -733,6 +622,43 @@ action = {
         </field>
     </record>
 
+    <record id="product_kanban_view" model="ir.ui.view">
```
shrm-odoo: Why did you reintroduce the kanban view here if it was removed from above?

addons/website_sale/templates/snippets/product_snippet_template_data.xml
```diff
@@ -53,7 +53,6 @@
                         t-set="ribbon"
                         t-value="product_template._get_ribbon(
                             price_vals=data,
-                            variant=product_variant,
                         )"
```
shrm-odoo, line 57: Can be written in one line

shrm-odoo (changes requested):
Hello @khaj-odoo 
I have some questions and suggestions.

shrm-odoo (commented):
There is a typo in the commit message: in the first point, there should be a space between “and” and “all”.

addons/product/views/product_views.xml
```diff
@@ -651,7 +576,6 @@
             </field>
         </record>
 
-
         <record id="product_kanban_view" model="ir.ui.view">
```
shrm-odoo, line 618: Don’t you need to completely remove the kanban view of product.product?

addons/product/views/product_attribute_views.xml
khaj-odoo: Rounded avatar style doesn't fit, text-start float-none is kept to avoid centering.

addons/product/static/src/product_catalog/kanban_controller.js
```diff
@@ -30,6 +30,10 @@ export class ProductCatalogKanbanController extends KanbanController {
     }
 
     async setOrderStateInfo() {
+        if (!this.orderResModel) {
+            this.orderStateInfo = {};
+            return;
+        }
```
shrm-odoo: In which cases would we not get `this.orderResModel`? 🤔

khaj-odoo: When product_view_kanban_catalog is opened via a many2one field on mobile SelectCreateDialog is used. It always picks type = "kanban" when env.isSmall is true, and it only passes the many2one field's context  no order context is included.

After removing product_kanban_view from product.product, this catalog kanban becomes the only available kanban view for product.product, so it gets picked automatically for any many2one product field popup on mobile. In that case, product_catalog_order_model is not in context, so this.orderResModel is undefined.

shrm-odoo: This doesn’t seem like a clean solution, as it feels more like a defensive workaround.
Looping in @Feyensv and @kcv-odoo for the best way to handle this.

shrm-odoo (approved):
The changes in the kanban controller, record, and model seem more like a workaround than a proper solution. The next reviewer may have better insights here. Otherwise, LGTM.

addons/product/static/src/product_catalog/kanban_controller.js
kcv-odoo: I would not do this, either re-add a simple view or change priority of `product_view_kanban_catalog` (if it works), as we don't want to open catalog kanban view from m2o fields in mobile or any other place by default

addons/product/models/product_template.py
```diff
@@ -336,21 +336,22 @@ def _compute_template_field_from_variant_field(self, fname, default=False):
             else:
                 template[fname] = default
 
-    def _set_product_variant_field(self, fname):
+    def _set_product_variant_field(self, fname, update_all_variants=False):
         """Propagate the value of the given field from the templates to their unique variant.
 
         Only if it's a single variant product.
```
kcv-odoo: Need to change this also

addons/product/views/product_views.xml
```diff
@@ -342,133 +350,30 @@
...
             <field name="name">Product Variants</field>
             <field name="res_model">product.product</field>
-            <field name="context">{'search_default_product_tmpl_id': [active_id], 'default_product_tmpl_id': active_id, 'create': False}</field>
+            <field name="context">
+                {
+                    'default_product_tmpl_id': active_id,
+                    'create': False,
+                    'search_default_product_tmpl_id': active_id
+                }
+            </field>
```
kcv-odoo:
```suggestion
            <field name="context">{
                'create': False,
                'default_product_tmpl_id': active_id,    
                'search_default_product_tmpl_id': active_id,
            }</field>
```
Nit, context swap, linting comma

addons/product/views/product_views.xml
```diff
@@ -485,29 +390,70 @@
...
+                    <field
+                        string="Sales Price"
+                        name="lst_price"
+                        optional="show"
+                        options="{'currency_field': 'currency_id'}"
+                        widget="monetary"
+                    />
+                    <!-- Currency, needed for monetary widgets to display the currency symbol -->
+                    <field name="currency_id" column_invisible="True"/>
+                    <field name="cost_currency_id" column_invisible="True"/>
```
kcv-odoo: Move this above lst_price as it is used in lst_price options

addons/product/views/product_views.xml
```diff
@@ -592,33 +531,19 @@
...
+                        widget="many2many_tags"
                     />
                 </field>
-                <field name="uom_ids" position="attributes">
-                    <attribute name="readonly">product_variant_count != 1</attribute>
-                    <attribute name="context">{'product_id': id}</attribute>
-                </field>
-                <field name="uom_ids" position="after">
-                    <field
-                        name="extra_uom_ids"
```
kcv-odoo, line 631: If we are removing `extra_uom_ids` from here I don't see they will be useful, Do we really want to remove this? as I see they are recently added in task https://www.odoo.com/odoo/project/4106/tasks/5152612

As it seems useful, What do you think?

addons/product/models/product_template_attribute_value.py
```diff
@@ -157,15 +157,6 @@ def unlink(self):
...
         return True
 
-    @api.depends('attribute_id')
-    def _compute_display_name(self):
-        """Override because in general the name of the value is confusing if it
-        is displayed without the name of the corresponding attribute.
-        Eg. on exclusion rules form
-        """
-        for value in self:
-            value.display_name = f"{value.attribute_id.name}: {value.name}"
```
kcv-odoo, line 167: I am not sure removing this from all the places would be good idea as mention in docstring only attribute value does not give proper idea in lots of case like Size X, S, XL does not make sense alone

May be it would be better make this condition to decide whether to display attribute name or not like `display_default_code` but not have clear idea where we should display attribute or not

Can you check? may be discuss with Jean

kcv-odoo (changes requested):
Thank you for your work and changes!

Added some small comments and question

addons/product/models/product_template_attribute_value.py
Jean-boje: Hello @kcv-odoo we can keep something in the code so each scope can still easily decide to display the attributes names in a specific view, but I think we can still remove it from everywhere. People will let us know if this is blocking for them at review.
cc @khaj-odoo

addons/product/views/product_views.xml
Jean-boje: We can re-introduce it, with an optional hidden field in the list view @kcv-odoo @khaj-odoo

addons/product/static/src/product_catalog/kanban_controller.js
Feyensv: Same opinion, let's keep providing a default kanban view, even if we do not use it (much) in standard anymore.
The catalog kanban shouldn't appear in non catalog flows.

kcv-odoo (approved):
Thank you for the changes! LGTM

addons/product/models/product_template.py
```diff
@@ -337,21 +337,19 @@ def _compute_template_field_from_variant_field(self, fname, default=False):
             else:
                 template[fname] = default
 
-    def _set_product_variant_field(self, fname):
-        """Propagate the value of the given field from the templates to their unique variant.
-
-        Only if it's a single variant product.
-        It's used to set fields like barcode, weight, volume..
+    def _set_product_variant_field(self, fname, update_all_variants=False):
+        """Propagate the value of the given field from the templates to their variants.
```
svs-odoo, line 341: I find it's confusing to say this methods will propagate the field to the template's variant**s** while it will do it only if there is a single variant by default, and in majority of cases, we only do it if there is a single variant (weight and volume are the only exceptions.)

addons/product/models/product_template.py
```diff
@@ -337,21 +337,19 @@ def _compute_template_field_from_variant_field(self, fname, default=False):
...
         """
         for template in self:
             count = len(template.product_variant_ids)
-            if count == 1:
+            if count == 1 or (update_all_variants and count > 1):
                 template.product_variant_ids[fname] = template[fname]
             elif count == 0:
-                archived_variants = self.with_context(active_test=False).product_variant_ids
+                archived_variants = template.with_context(active_test=False).product_variant_ids
                 if len(archived_variants) == 1:
```
svs-odoo, line 353: Shouldn't we update all the archived variants if `update_all_variants` is true ?
```suggestion
                archived_count = len(archived_variants)
                if archived_count == 1 or (update_all_variants and archived_count > 1):
```

addons/product/views/product_views.xml
```diff
@@ -260,13 +267,13 @@
                         >
                             <group name="inventory">
                                 <group name="group_lots_and_weight" string="Logistics" invisible="type != 'consu'">
-                                    <label for="weight" invisible="product_variant_count &gt; 1 and not is_product_variant"/>
-                                    <div class="o_row" name="weight" invisible="product_variant_count &gt; 1 and not is_product_variant">
+                                    <label for="weight"/>
```
svs-odoo, line 270: Some remarks about this spec:
> **Weight/Volume always visible** on product template even with variants

  I find that it's error prone because the user can set the weight or the volume on each variants, than if by mistake they set it on the template, all already set values on the variants are lost and replaced.

kcv-odoo: @Jean-boje :up:

Jean-boje: Hello @svs-odoo thanks for checking :) This is the intended behavior. We assume that in the vast majority of cases, variants will share the same weight and volume, so we wanted to prevent you from having to add it to each variant manually. cc @kcv-odoo

addons/website_sale/models/product_product.py
```diff
@@ -10,7 +10,6 @@ class ProductProduct(models.Model):
     _inherit = "product.product"
     _mail_post_access = "read"
 
-    variant_ribbon_id = fields.Many2one(string="Variant Ribbon", comodel_name="product.ribbon")
```
Feyensv, line 13: Was kinda surprised to see this in the task since we have other task(s) to show variants on the shop page.
After discussion with Jean, he changed his mind and believes we should keep the ability to set a ribbon on a variant.

Could you keep the field (and remove the related changes) please ? 
thanks :pray:

addons/product/models/product_template.py
```diff
@@ -422,17 +421,21 @@ def _search_virtual_available(self, operator, value):
...
     def _set_volume(self):
-        self._set_product_variant_field('volume')
+        self._set_product_variant_field('volume', update_all_variants=True)
 
     @api.depends('product_variant_ids.weight')
     def _compute_weight(self):
-        self._compute_template_field_from_variant_field('weight')
+        self.filtered(
+            lambda t: t.product_variant_count < 2
+        )._compute_template_field_from_variant_field('weight')
```
Feyensv: Why do you add this filtered check?
As far as I know, `_compute_template_field_from_variant_field` already handles it 🤔

addons/product/views/product_views.xml
```diff
@@ -355,135 +362,27 @@
...
-                                <field name="product_tag_ids" widget="many2many_tags" readonly="1"/>
-                                <field name="additional_product_tag_ids" widget="many2many_tags" context="{'product_variant_id': id}"/>
-                            </group>
-                        </group>
-                    </sheet>
-                </form>
-            </field>
-        </record>
-
         <record id="product_variant_action" model="ir.actions.act_window">
```
Feyensv, line 376: After thinking more about the current behavior with @Jean-boje, we believe it'd be better for the list view to edit the record instead of opening its form view (when coming from this action).

It's a bit confusing to land on a view nearly identical to the template one when you come from the template.  Since the list view is meant to display fields that are variant-specific, users should be encouraged to stay on it and edit the variants from there. For those who want to access the variant form view, we should add the 'View' button (there is a framework feature for it since it can be added by studio).

Image

addons/website_sale/models/product_template.py
Feyensv, line 1: Now that the ribbon logic was removed, I believe the diff here is unnecessary.

addons/sale_stock/tests/test_sale_stock.py
```diff
@@ -2194,10 +2194,10 @@ def test_move_description(self):
                 'product_no_variant_attribute_value_ids': [Command.set(product_with_description.attribute_line_ids[1].product_template_value_ids[0].ids)],
             }
         )
-        self.assertEqual(so.order_line.name, 'Product with description (red)\nSale description\nNo variant: extra')
```
Feyensv, line 2197: This behavior shouldn't change, we should keep the no_variant information on lines description, both the attribute name and the attribute value.
Otherwise there is no way to differentiate which attribute had this value.
Jean only meant to change the display on the backend views, not remove information from order lines.

addons/website_sale/views/product_views.xml
```diff
@@ -51,9 +43,6 @@
             <field name="default_code" position="after">
                 <field name="website_id" groups="website.group_multi_website" optional="hide"/>
             </field>
-            <field name="product_tag_ids" position="after">
-                <field name="website_ribbon_id" optional="hide"/>
-            </field>
```
Feyensv, line 56: I don't see any expected changes to the product template tree view in the spec, this was maybe left over from the ribbon changes ?

addons/website_sale/tests/test_product_ribbon.py
Feyensv, line 1: The changes here are probably unnecessary now that `_get_ribbon` changes have been dropped.

addons/product/models/product_template.py
khaj-odoo: Without the filtered, `_compute_template_field_from_variant_field `hits its else branch `(template[fname] = default)` for multi-variant products, and since `default=False`, that resets the template weight/volume to 0, wrong since each variant now holds its own value. Passing a custom default wouldn't help either, it still iterates over all templates in the loop, whereas filtered skips them entirely before the call.

Feyensv: Well, considering it's a stored computed field, I've always been told it's a bad practice to not set a value to some records in the compute :thinking:

Could you at least add a comment to explain this in the code (and ideally add a test covering this custom behavior ?)

addons/sale_stock/tests/test_sale_stock.py
Feyensv, line 1: Same comment than for sale order lines description, pickings descriptions shouldn't change (and should keep the attribute name to know which attribute this value refers to)

addons/sale_purchase/tests/test_sale_purchase.py
Feyensv, line 1: same comment but for PO lines

addons/website_sale/tests/test_product_configurator.py
Feyensv, line 1: Since the description of sale order lines was supposed to be kept, why is the diff in this file still needed ?

addons/product/models/product_template_attribute_value.py
Feyensv: Considering all the other comments, it could make sense to keep this method (and default behavior) and only use the value `name` when a specific context is provided 🤔 (which could be set on the `product.product` field itself)

addons/product/models/product_template.py
Feyensv: Thought about it more, and why don't we provide some key for the `_compute_template_field_from_variant_field` method too to say that the value should be computed from all the variants ?

That way, the compute would work as expected and display a value only if all the variants have the same one.

addons/product/models/product_template_attribute_value.py
```diff
@@ -157,14 +157,11 @@ def unlink(self):
         ptav_to_archive.write({'ptav_active': False})
         return True
 
-    @api.depends('attribute_id')
```
Feyensv, line 160: the dependency on the attribute_id should be kept (handled in my fixup commit already)

addons/product/tests/test_variants.py
```diff
@@ -1624,6 +1624,35 @@ def write(self, vals):
...
+            {'name': 'Red', 'attribute_id': color_attr.id},
+            {'name': 'Blue', 'attribute_id': color_attr.id},
+        ])
+        template = self.env['product.template'].create({
+            'name': 'Multi Variant Product',
+            'attribute_line_ids': [Command.create({
+                'attribute_id': color_attr.id,
+                'value_ids': [Command.set(color_attr.value_ids.ids)],
+            })],
+        })
```
Feyensv:
```suggestion
        template = self.env['product.template'].create({
            'name': 'Multi Variant Product',
            'attribute_line_ids': [Command.create({
                'attribute_id': self.color_attribute.id,
                'value_ids': [Command.link(self.color_attribute_red.id), Command.link(self.color_attribute_blue.id))],
            })],
        })
```

you can reuse the values from the `ProductVariantsCommon` setup ;) (handled in fixup commit)

addons/product/models/product_template.py
```diff
@@ -337,26 +338,29 @@ def _compute_template_field_from_variant_field(self, fname, default=False):
             elif variant_count == 0 and self.env.context.get("active_test", True):
                 # If the product has no active variants, retry without the active_test
                 template_ctx = template.with_context(active_test=False)
-                template_ctx._compute_template_field_from_variant_field(fname, default=default)
+                template_ctx._compute_template_field_from_variant_field(fname, default=default, multi_variant=multi_variant)
+            elif multi_variant:
+                values = template.product_variant_ids.mapped(fname)
+                if len(set(values)) == 1:
+                    template[fname] = values[0]
```
Feyensv, line 348: Why can't we set the default when all the variants do not share the same value ?
If for some reason some variants do not have the template value, the template value should be reset IMHO.

Feyensv: PO requested this behavior:
If a non-default value is set, it means that either all variants have the same value, or that it was the last value set on the template, regardless of variant-specific changes done afterwards.

Feyensv:
robodoo r+

PR #244188, opened 2026-01-16
[IMP] hr: allow HR user to archive unused departure reason

xavieralt:
An HR user has currently no way to prevent use of specific /deprecated departure reason other than specifying it in the name.

This commit an active field on `hr.departure.reason` so that an HR user can archive departure reasons that should not be used (as he can't delete them if they are referenced at least once).

---


Bertrand2 (approved):
LGTM! Thanks

Bertrand2:
robodoo r+

PR #244193, opened 2026-01-16
[FIX] hr: fix JS error by adding explicit views to action

yhu-odoo:
The frontend action service requires the 'views' field to be present in the action dictionary.

Without it, `_preprocessAction` fails when trying to access `action.views`[1], causing a client-side crash. This commit ensures `views` is defined alongside `view_mode`.

[1]: https://github.com/odoo/odoo/blob/3c5e2be7eb3b22b1d5eb28c1eae37fe61dabcd46/addons/web/static/src/webclient/actions/action_service.js#L442

Task-[5368166](https://www.odoo.com/odoo/5778/tasks/5368166)
Ent PR odoo/enterprise#102547

---

### DIFF
--- a/addons/hr/models/res_partner.py
+++ b/addons/hr/models/res_partner.py
@@ -26,6 +26,7 @@ class ResPartner(models.Model):
                 'type': 'ir.actions.act_window',
                 'res_model': 'hr.employee',
                 'view_mode': 'kanban',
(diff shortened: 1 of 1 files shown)


mepe-odoo (approved):
lgtm

oomsveta (approved):
robodoo review+ delegate+

PR #244195, opened 2026-01-16
[FIX] pos_restaurant: handle split orders correctly in preparation display

jipr-odoo:
Steps to Reproduce
==================
- Create an order with Product A ×2 and Product B ×2
- Send the order to the preparation display
- Split the order and pay for one unit of each product
- On the remaining order, add another product
- Send the order to the preparation display
- The preparation display incorrectly cancels previously sent products

Fix
===
- Prevent cancellation of existing preparation order lines
- Avoid duplication of preparation lines when split orders are sent to the kitchen

Covered Scenario
================
1. Send an order with Coca-Cola ×2, Minute Maid ×2, Water ×1 to the kitchen.
2. Split 1 qty of each product into a new order, increase split quantities, and send it.
3. Increase quantities on the original order and send it again.

Expected Behavior
=================
- Initial preparation order remains unchanged (no line cancellations)
- Split order creates a new preparation card with only unsend split quantities
- Original order creates a new preparation card with only newly added quantities
- No preparation line incorrectly aggregates quantities across orders

Task-5355899

Related Ent. PR: https://github.com/odoo/enterprise/pull/104554

### DIFF
--- a/addons/pos_restaurant/static/src/app/screens/split_bill_screen/split_bill_screen.js
+++ b/addons/pos_restaurant/static/src/app/screens/split_bill_screen/split_bill_screen.js
@@ -128,6 +128,7 @@ export class SplitBillScreen extends Component {
         newOrder.floating_order_name = newOrderName;
         newOrder.uiState.splittedOrderUuid = curOrderUuid;
         originalOrder.uiState.splittedOrderUuid = newOrder.uuid;
+        const isOrdered = Object.keys(originalOrder.last_order_preparation_change.lines).length > 0;
 
         // Create lines for the new order
         const comboMap = new Map();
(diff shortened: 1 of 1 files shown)


addons/pos_restaurant/static/src/app/screens/split_bill_screen/split_bill_screen.js
```diff
@@ -185,13 +186,39 @@ export class SplitBillScreen extends Component {
...
+                // because `handlePreparationHistory` deletes prep key after split, when a line is split and sent again to the preparation display.
+                if (isOrdered) {
+                    const origSplit = (originalOrder.last_order_preparation_change.split_lines ??=
+                        {});
+
+                    if (originalOrder.last_order_preparation_change.lines[line.uuid]) {
+                        origSplit[line.uuid] = {
+                            uuid: line.uuid,
+                            splitted_qty: this.qtyTracker[line.uuid],
+                        };
```
loukwets, line 199: It means that you replace the previous splitted quantity in case you split two times the original order, in this case only the splitted_qty of the last split would be stored. It also implies the fact that if you split an order that is the result of another split you loose the original_order and in the backend it goes to the wrong computed, but also the splitted_qty would need to be computed based on the previous splitted quantity.

loukwets:
@jipr-odoo Ludo made a complete solution [here](https://github.com/odoo/odoo/pull/266636) so I don't think we still need this, thank you for your work !

jipr-odoo:
> @jipr-odoo Ludo made a complete solution [here](https://github.com/odoo/odoo/pull/266636) so I don't think we still need this, thank you for your work !

Thanks @loukwets for The review and apology for late reply 
If this issue is going to be fixed by PR #266636, then this PR is no longer needed, and I'll close it.

PR #244202, opened 2026-01-16
[FIX] hr_holidays: fix forecasted balance for hourly time off types

SurryaT10:
Steps to Reproduce:
Create an accrual plan that accrues 2 days monthly, credited at the start of the month. Create a Time Off Type with the Unit of Measure set to Hours. Create an allocation request using the above Time Off Type and Accrual Plan. Open the Time Off dashboard:
a. The current available balance is displayed in hours and is correctly computed based on the employee’s working schedule. b. However, when checking the balance for a future date, the system incorrectly adds 2 hours instead of 16 hours (i.e., 2 days converted to hours based on the working schedule).

Bug cause:
The system incorrectly identified the unit of measure because it was still referencing the request_unit field, which defaults to 'day'. Following a recent refactor where request_unit was replaced by unit_of_measure to define Time Off Type units, this specific logic flow was left unupdated, causing it to be calculated in days instead of hours.

Solution:
Updated the future leaves calculation logic to use unit_of_measure instead of request_unit. This ensures that when a Time Off Type is configured in hours, accruals defined in days are correctly converted based on the employee's working schedule.

Task: 5498292

---

### DIFF
--- a/addons/hr_holidays/models/hr_leave_allocation.py
+++ b/addons/hr_holidays/models/hr_leave_allocation.py
@@ -644,7 +644,7 @@ class HrLeaveAllocation(models.Model):
 
         fake_allocation = self.env['hr.leave.allocation'].with_context(default_date_from=accrual_date).new(origin=self)
         fake_allocation.sudo().with_context(default_date_from=accrual_date)._process_accrual_plans(accrual_date, log=False)
-        if self.holiday_status_id.request_unit in ['hour']:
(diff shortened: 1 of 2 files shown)


amry-odoo (approved):
Thanks for your work! :rocket: :raccoon:

amah-odoo (approved):
Thanks for your work 🚀 

@robodoo r+

PR #244203, opened 2026-01-16
[FIX] project_timesheet_holidays: Exclude archived employees from time-off 

VincentJanssen-Code:
[FIX] project_timesheet_holidays: Exclude archived employees from time-off 

# Description of the issue/feature this PR addresses:
## Steps to Reproduce:
1. Create a time off for Employee A (it should affect the timesheets).
2. Create a new public holiday (global time off) that overlaps with Employee A’s time off.
3. Archive Employee A.
4. Delete the public holiday created in step 2.
5. An error related to timesheet generation appears.

## Expected Behavior:
- The public holiday / global time off should be deleted without any error.

# Desired behavior after PR is merged:
## Fix (Implemented):
When regenerating timesheets due to changes in holidays or time off, leaves related to archived employees should not be taken into account. A check was added inside the `_generate_timesheets` function in `project_timesheet_holidays/models/hr_holidays.py` to exclude leaves belonging to archived employees.

## Alternative Fix (Not Implemented):
Instead of filtering out leaves linked to archived employees, we could delete those leaves when an employee is archived. However, this approach is not ideal, as archived employees may be reactivated later and would still need their previously requested time off to be preserved.

 ## Version:
 This bug appears in both version 17.0 and 19.0. I assumed that it also appears in 18.0  but didn't directly test

## Task:

[5474038](https://www.odoo.com/odoo/project/4105/tasks/5474038)

---

### DIFF
--- a/addons/project_timesheet_holidays/models/hr_holidays.py
+++ b/addons/project_timesheet_holidays/models/hr_holidays.py
@@ -68,9 +68,10 @@ class Holidays(models.Model):
         """
         vals_list = []
         leave_ids = []
-        calendar_leaves_data = self.env['resource.calendar.leaves']._read_group([('holiday_id', 'in', self.ids)], ['holiday_id'], ['id:array_agg'])
+        leaves_with_active_employee = self.filtered(lambda l: l.employee_id.active)
+        calendar_leaves_data = self.env['resource.calendar.leaves']._read_group([('holiday_id', 'in', leaves_with_active_employee.ids)], ['holiday_id'], ['id:array_agg'])
         mapped_calendar_leaves = {leave: calendar_leave_ids[0] for leave, calendar_leave_ids in calendar_leaves_data}
-        for leave in self:
+        for leave in leaves_with_active_employee:
             if leave.holiday_type != 'employee' or not leave.holiday_status_id.timesheet_generate:
                 continue
 
--- a/addons/project_timesheet_holidays/tests/test_employee.py
+++ b/addons/project_timesheet_holidays/tests/test_employee.py
@@ -120,3 +120,44 @@ class TestEmployee(TransactionCase):
         self.assertEqual(len(timesheet), 1, 'A timesheet should have been created for the global leave of the employee')
         self.assertEqual(str(timesheet.date), '2020-01-01', 'The timesheet should be created for the correct date')
         self.assertEqual(timesheet.unit_amount, 8, 'The timesheet should be created for the correct duration')
+
+    @freeze_time('2020-01-01')
+    def test_timesheet_inactive_employee(self):
+        """ Test if the timesheets representing the time off of this employee,
+            are correctly generated once the employee is set to inactive
+        """
+        employee = self.env['hr.employee'].create({
+            'name': 'Test Employee',
+            'resource_calendar_id': self.company.resource_calendar_id.id,
+            'company_id': self.company.id,
+        })
+        old_timesheet_count = len(self.env['account.analytic.line'].search([
+            ('employee_id', '=', employee.id)]))
+        generic_time_off_type = self.env['hr.leave.type'].create({
+            'name': 'Generic Time Off',
+            'requires_allocation': 'no',
+            'leave_validation_type': 'both',
+            'company_id': self.company.id,
+        })
+        leave = self.env['hr.leave'].create({
+            'employee_id': employee.id,
+            'state': 'confirm',
+            'holiday_type': 'employee',
+            'holiday_status_id': generic_time_off_type.id,
+            'request_date_from': '2020-01-02 08:00:00',
+            'request_date_to': '2020-01-07 17:00:00',
+        })
+        leave.action_validate()
+        timesheets = self.env['account.analytic.line'].search([
+            ('employee_id', '=', employee.id)])
+        self.assertEqual(len(timesheets), old_timesheet_count + 4, "4 new Timesheets should be generated for timeoff that doesn't fall on a week-end")
+
+        employee.write({'active': False})  # Archiveing an employee will delete the timesheet related to its future holidays, this will change the number of timesheets
+        old_timesheet_count = len(self.env['account.analytic.line'].search([
+            ('employee_id', '=', employee.id)]))
+
+        leave._generate_timesheets()
+        self.assertTrue(timesheets, "Timesheet should not have been regenerated and therefore old timesheet shouldn't be deleted")
+        timesheets = self.env['account.analytic.line'].search([
+            ('employee_id', '=', employee.id)])
+        self.assertEqual(len(timesheets), old_timesheet_count, "New timesheets should not have been generated due to employee being inactive/archived")


addons/project_timesheet_holidays/models/hr_holidays.py
```diff
@@ -71,6 +71,8 @@ def _generate_timesheets(self):
         calendar_leaves_data = self.env['resource.calendar.leaves']._read_group([('holiday_id', 'in', self.ids)], ['holiday_id'], ['id:array_agg'])
```
xavierbol: We could maybe create a variable to only have the leaves with active employees. By doing that, we will not fetch the resource.calendar.leaves linked to a leave with an archived employee. 🙂 

```suggestion
        leaves_with_active_employee = self.filtered(lambda l: l.employee_id.active)
        calendar_leaves_data = self.env['resource.calendar.leaves']._read_group([('holiday_id', 'in', leaves_with_active_employee.ids)], ['holiday_id'], ['id:array_agg'])
```

addons/project_timesheet_holidays/models/hr_holidays.py
```diff
@@ -71,6 +71,8 @@ def _generate_timesheets(self):
...
         mapped_calendar_leaves = {leave: calendar_leave_ids[0] for leave, calendar_leave_ids in calendar_leaves_data}
         for leave in self:
```
xavierbol:
```suggestion
        for leave in leaves_with_active_employee:
```

addons/project_timesheet_holidays/models/hr_holidays.py
```diff
@@ -71,6 +71,8 @@ def _generate_timesheets(self):
...
+            if not leave.employee_id.active:
+                continue
```
xavierbol: no longer needed if we use leaves_with_active_employee. 🙂

xavierbol (commented):

deneuvillem (commented):
Thanks for your fix! 🎖️ 
Could you add a small test to ensure the issue was correctly fixed? Also, I think you can squash your commits into a single one :) Thanks in advance!

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
         self.assertEqual(len(timesheet), 1, 'A timesheet should have been created for the global leave of the employee')
         self.assertEqual(str(timesheet.date), '2020-01-01', 'The timesheet should be created for the correct date')
         self.assertEqual(timesheet.unit_amount, 8, 'The timesheet should be created for the correct duration')
+
+    @freeze_time('2020-01-01')
+    def test_timesheet_inactif_employee(self):
```
deneuvillem:
```suggestion
    def test_timesheet_inactive_employee(self):
```

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+        """ Test the timesheets representing the time off of this employee
+            is correctly generated once the employee is set to inactif
```
deneuvillem:
```suggestion
            are correctly generated once the employee is set to inactive
```

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+        """
+        self.employee = self.env['hr.employee'].create({
```
deneuvillem:
```suggestion
        employee = self.env['hr.employee'].create({
```

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+        })
+        leave = self.env['hr.leave'].create({
+            'employee_id': self.employee.id,
+            'state': 'draft',
+            'holiday_type': 'employee',
+            'holiday_status_id': self.generic_time_off_type.id,
+            'request_date_from': '2020-01-02 08:00:00',
+            'request_date_to': '2020-01-07 17:00:00',
+        })
+        leave.action_confirm()
```
deneuvillem:
```suggestion
```
as the state will be 'confirm', no need to do this

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+            ('employee_id', '=', self.employee.id)]))
+        self.generic_time_off_type = self.env['hr.leave.type'].create({
+            'name': 'Generic Time Off',
+            'requires_allocation': 'no',
+            'leave_validation_type': 'both',
+            'company_id': self.company.id,
+        })
+        leave = self.env['hr.leave'].create({
+            'employee_id': self.employee.id,
+            'state': 'draft',
```
deneuvillem:
```suggestion
            'state': 'confirm',
```

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+            'holiday_type': 'employee',
+            'holiday_status_id': self.generic_time_off_type.id,
+            'request_date_from': '2020-01-02 08:00:00',
+            'request_date_to': '2020-01-07 17:00:00',
+        })
+        leave.action_confirm()
+        leave.action_validate()
+        timesheets = self.env['account.analytic.line'].search([
+            ('employee_id', '=', self.employee.id)])
+        self.assertGreater(len(timesheets), old_timesheet_count, "Timesheet should be generated for timeoff of active employees")
```
deneuvillem: Can't we be more accurate and check that the timesheet count has increased by 1?

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+        self.employee.write({'active': False})
+        leave._generate_timesheets()
+        self.assertTrue(timesheets, "Timesheet should not have been regenerated and therefore old timesheet shouldn't be deleted")
```
deneuvillem, line 160: Can we also check that the count of timesheets has not increased?

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+        self.assertNotEqual(len(timesheets), 0, "Timesheet should not have been regenerated and therefore old timesheet shouldn't be deleted")
```
deneuvillem:
```suggestion
```
The previous assertion already asserts that the recordset is not empty

addons/project_timesheet_holidays/tests/test_employee.py
```diff
@@ -120,3 +120,39 @@ def test_write_employee(self):
...
+            is correctly generated once the employee is set to inactif
+        """
+        self.employee = self.env['hr.employee'].create({
+            'name': 'Test Employee',
+            'resource_calendar_id': self.company.resource_calendar_id.id,
+            'company_id': self.company.id,
+        })
+        old_timesheet_count = len(self.env['account.analytic.line'].search([
+            ('employee_id', '=', self.employee.id)]))
+        self.generic_time_off_type = self.env['hr.leave.type'].create({
```
deneuvillem:
```suggestion
        generic_time_off_type = self.env['hr.leave.type'].create({
```

deneuvillem (changes requested):
Thanks for your fix, I left some comments for you :)

addons/project_timesheet_holidays/tests/test_employee.py
VincentJanssen-Code: I believe it is more reliable to verify that the count of timesheet records returned by
`self.env['account.analytic.line'].search(...)` remains unchanged after calling
`leave._generate_timesheets()`. This ensures that `leave._generate_timesheets()` has not accidentally created or removed _any_ timesheet entries.

addons/project_timesheet_holidays/tests/test_employee.py
VincentJanssen-Code: Yes we can but it won't increase by 1 but 4, as a new timesheet is created for each days off that isn't a weekend. So from 02/01/2026 to 07/01/2026 that make four business days.

deneuvillem (approved):
LGTM, thanks for your fix!

robodoo r+

PR #244204, opened 2026-01-16
[FIX] mail: allow starring persistent messages without thread

nees-odoo:
Starring a message should depend on whether the message is persistent and 
whether the current user is allowed to star it, not on whether the message is 
attached to a thread.

Some persistent messages (such as user notifications) can be displayed in Inbox 
without an associated thread. In such cases, the star action was unavailable 
even though the message itself was valid and persistent.

This change removes the dependency on thread presence when determining 
whether a message can be starred. As a result, users can now star persistent messages 
without a thread, including Inbox notifications such as sign requests.

task-[5473438](https://www.odoo.com/odoo/project/1519/tasks/5473438)

---

### DIFF
--- a/addons/mail/static/tests/message/message.test.js
+++ b/addons/mail/static/tests/message/message.test.js
@@ -1002,6 +1002,26 @@ test("toggle_star message", async () => {
     await contains(".o-mail-Message [title='Mark as Todo']" + " i.fa-star-o");
 });
 
+test("can star a persistent message without thread", async () => {
+    const pyEnv = await startServer();
+    const messageId = pyEnv["mail.message"].create({
+        body: "Test",
+        message_type: "user_notification",
+        needaction: true,
+    });
+    pyEnv["mail.notification"].create({
+        mail_message_id: messageId,
+        notification_status: "sent",
+        notification_type: "inbox",
+        res_partner_id: serverState.partnerId,
+    });
+    await start();
+    await openDiscuss();
+    await contains(".o-mail-Message:not([data-starred]):contains('Test')");
+    await click(".o-mail-Message:contains('Test') [title='Mark as Todo']");
+    await contains(".o-mail-Message[data-starred]:contains('Test')");
+});
+
 test("Name of message author is only displayed in chat window for partners others than the current user", async () => {
     const pyEnv = await startServer();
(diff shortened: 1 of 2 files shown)


addons/mail/static/tests/message/message.test.js
```diff
@@ -1002,6 +1002,26 @@ test("toggle_star message", async () => {
...
+    });
+    pyEnv["mail.notification"].create({
+        mail_message_id: messageId,
+        notification_status: "sent",
+        notification_type: "inbox",
+        res_partner_id: serverState.partnerId,
+    });
+    await start();
+    await openDiscuss();
+    await click(".o-mail-Message [title='Mark as Todo'] i.fa-star-o");
```
maad-odoo: why do we need to have two (`title` and `i.fa-star-o`) different selectors to assert one element ?
Isn't `title` or `i.fa-star-o` isn't enough ?

addons/mail/static/tests/message/message.test.js
```diff
@@ -1002,6 +1002,26 @@ test("toggle_star message", async () => {
...
+    await contains("button", { text: "Starred", contains: [".badge", { text: "1" }] });
+    await contains(".o-mail-Message [title='Mark as Todo'] i.fa-star");
```
maad-odoo: same as above. why need to selector for one element ?

addons/mail/static/tests/message/message.test.js
```diff
@@ -1002,6 +1002,26 @@ test("toggle_star message", async () => {
...
+});
```
maad-odoo, line 1023: This is a positive assertion without a corresponding negative assertion. While it works for this case, I think we can improve it by also validating the message content.

The reason is that if there are `x` messages, there will `x` similar elements in the DOM. Selecting elements in the same way as we do now does not guarantee that we are asserting against the exact message we intend, unless we explicitly target a specific child within the parent (using pseudo-selectors).

To ensure the uniqueness and reliability of the assertion step for a particular message ,in the particular test case ,I believe we should assert on the message content itself along with some negative assertions before positive assertion.

Let me know what do you think about this one :)

maad-odoo (commented):
Few comments 🙂

maad-odoo (approved):

alexkuhn (approved):
@robodoo delegate+ r+

PR #244205, opened 2026-01-16
[FIX] mail: fix traceback when opening chatter in documents

pyduf:
How to reproduce:
- Install documents
- Create a folder at the root (company) not shared to anyone (including internal user)
- Open that folder and ensure the chatter is closed
- Upload a document in it and share it with Marc Demo with view access
- Connect with Marc Demo, click on that shared document and open the chatter You get an error because the client try to update the thumbnail of the attachment for the chatter but the user has only view access to it.

The user doesn't have write access to the attachment because it is linked to a document with only read access. To solve the problem, we modify the check that trigger the thumbnail update to also check that the user has access to the related record.

Task-5360962

### DIFF
--- a/addons/mail/static/src/core/common/attachment_model.js
+++ b/addons/mail/static/src/core/common/attachment_model.js
@@ -37,7 +37,10 @@ export class Attachment extends FileModelMixin(Record) {
             if (
                 this.isPdf &&
                 !this.has_thumbnail &&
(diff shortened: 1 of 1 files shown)


alexkuhn (approved):
Would be nice a test coverage
@robodoo delegate+

awa-odoo:
@robodoo r+

PR #244217, opened 2026-01-16
[IMP] spreadsheet_dashboard_pos_restaurant: added pos.config filter

shag-odoo:
Following this commit:
====
- pos.config filter is been added for POS Restaurant menu
- Resequence the order of filters.
- Date filter label is been updated to Period.

task-5490607


davidmonnom (approved):
@robodoo r+

PR #244229, opened 2026-01-16
[FIX] html_builder: make option containers reactive

Syozik:
After the [refactoring of html_builder], containers weren't updated properly all the time. They were updated only when the options length or ids respectively have been changed, which is not the desired behavior. For example, if `isCloneDisabledReason` changes, we won't be able to see the changes right away. That's why we make containers reactive here, so the components are updated automatically whenever any of the properties change.

Steps to see the issue:
- Open website and start editing
- Drop a form
- Add a field and click on it
- Change its type to one of the existing fields, e.g. 'Alias Domain'

=> The duplicate button of the builder is not deactivated while it should be. The issue exists vice versa too (when we change the type from an existing field to a custom one).

[refactoring of html_builder]: https://github.com/odoo/odoo/commit/9fe45e2

task-5391316

### DIFF
--- a/addons/html_builder/static/src/core/builder_options_plugin.js
+++ b/addons/html_builder/static/src/core/builder_options_plugin.js
@@ -5,6 +5,7 @@ import { isClonable } from "./clone_plugin";
 import { getElementsWithOption, isElementInViewport } from "@html_builder/utils/utils";
 import { shouldEditableMediaBeEditable } from "@html_builder/utils/utils_css";
 import { OptionsContainer } from "@html_builder/sidebar/option_container";
+import { reactive } from "@odoo/owl";
 
 /** @typedef {import("@html_builder/core/utils").BaseOptionComponent} BaseOptionComponent */
 /** @typedef {import("@odoo/owl").Component} Component */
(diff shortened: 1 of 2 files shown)


addons/html_builder/static/src/core/builder_options_plugin.js
```diff
@@ -322,22 +328,24 @@ export class BuilderOptionsPlugin extends Plugin {
...
-                hasOverlayOptions: this.hasOverlayOptions(element),
-                isRemovable: isRemovable(element),
-                removeDisabledReason: this.getRemoveDisabledReason(element),
-                isClonable: isClonable(element),
-                cloneDisabledReason: this.getCloneDisabledReason(element),
-                optionsContainerTopButtons: this.getOptionsContainerTopButtons(element),
-            }));
+            .map(([element, options]) =>
+                reactive({
+                    id: previousElementToIdMap.get(element) || uniqueId(),
```
FrancoisGe: Why not make :
```js
let containers = reactive(...)
```

addons/html_builder/static/src/core/builder_options_plugin.js
```diff
@@ -247,25 +248,30 @@ export class BuilderOptionsPlugin extends Plugin {
...
+                    changedProperties.push(changes);
+                }
+                // Mutate existing reactive containers with only the changed properties
+                for (let i = 0; i < this.lastContainers.length; i++) {
+                    for (const [key, value] of Object.entries(changedProperties[i])) {
+                        this.lastContainers[i][key] = value;
+                    }
                 }
+                // Skip full dispatch since reactivity handles the updates
+                return;
```
FrancoisGe: It s a question, it looks maybe a little bit overkill to check if the value change before. If we use the reactivity, the difference will be check by the reactivity (If something change, it s will be rerender automaticly) Maybe check if I don t miss something

PR #244230, opened 2026-01-16
[IMP] mail: poll votes panel

tsm-odoo:
Like for reactions, its interresting to be able to see
who votes for which option during a poll. This PR introduces
a panel in order to show it.

task-5207317
image
image


addons/mail/static/src/core/common/poll_votes_panel.js
```diff
@@ -0,0 +1,54 @@
...
+    get contentClass() {
+        const attClass = {
+            "h-50 d-flex": true,
+            "position-absolute top-100 start-0": this.store.useMobileView,
+        };
+        return Object.entries(attClass)
+            .filter(([_, value]) => value)
+            .map(([classNames]) => classNames)
+            .join(" ");
+    }
```
alexkuhn, line 50: There's `attClassObjectToString` that does the conversation from attClass object to string.

alexkuhn (approved):
@robodoo delegate+

PR #244233, opened 2026-01-16
[IMP] web: enhance security by protecting critical databases from deletion with the new conf option protected_db

nedhirDalleji:
[IMP] web: enhance security by protecting critical databases from deletion with the new conf option protected_db

This security enhancement is made for the master version:
- Odoo 19.2a1

Before this commit:     
A person that has the admin_passwd can with or without intention delete critical databases such as the production database.

After this commit:
A person that has access to odoo.conf can disable the ability to delete critical databases (separated by commas) by declaring them in the new conf option protected_db.
- For client side, in the html served through the endpoint '/web/database/manager' the delete button will be disabled for the protected databases.
- For server side, if any attempt is made to bypass client side restriction (or via a direct HTTP request ) to delete a protected database, a second server validation through
'/web/database/drop' is made and this attempt will be still denied.

---


odoo/addons/base/tests/config/16.0.conf
```diff
@@ -38,6 +38,7 @@ max_cron_threads = 2
 osv_memory_count_limit = 0
 pg_path = 
 pidfile = 
+protected_db =
```
Julien00859: Don't change this file. It is a config was generated using --save in Odoo 16.0 and is used to make sure that an older config can load fine with a newer version.

odoo/tools/config.py
```diff
@@ -405,6 +405,8 @@ def _build_cli(self):
                             help="Disable the ability to obtain or view the list of databases. "
                                  "Also disable access to the database manager and selector, "
                                  "so be sure to set a proper --database parameter first")
+        security.add_option('--protected-db', dest='protected_db', type='comma', metavar="DATABASE,...", my_default=[],
+                            help="Disable the ability to delete protected databases (separated by commas).")
```
Julien00859:
```suggestion
                            help="A list of databases that cannot be removed by the database manager (separated by commas).")
```

addons/web/controllers/database.py
```diff
@@ -203,6 +204,10 @@ def duplicate(self, master_pwd, name, new_name, neutralize_database=False):
 
     @route('/web/database/drop', type='http', auth="none", methods=['POST'], csrf=False)
     def drop(self, master_pwd, name):
+        if name in odoo.tools.config["protected_db"]:
+            e = _("Houston, we've got you covered! even a misclick can't lead to delete a protected database.")
```
Julien00859:
```suggestion
            e = _("Houston, we've got you covered! even a miss-click can't delete a protected database.")
```

Julien00859 (changes requested):
With great powers come great responsibilities. Are you doing this PR because you, one of your admins, or a customer with admin privileges did delete a production database by accident?

Like mart-e I consider that everything related to the database-manager should not be accessible on a production database. Still, I understand that "practicality beats purity" (in PEP20 terms) and that it is an handy tool.

We don't like to had new CLI options, as there are so many of those already. In this case I see no other way (an ICP wouldn't work). An alternative, have the :wastebasket: buttons in the web interface prompts the user to type again the name of the database to delete. Or have a prompt with "You are about to delete <dbname>. Proceed?" but the "Ok" button is only accessible after like 5 seconds. Makes for more changes, but no CLI option. I can offer to create a task in our own pipeline and assign someone from Odoo to do it.

---

Below is my technical review, if we were to merge this new option which is unlikely.

There must also be a commit in the documentation repo, to adapt `content/developer/reference/cli.rst` and document this new option.

PR #244234, opened 2026-01-16
[FIX] website_sale_comparison: hide single custom value attribute

mma1377:
### Issue:
Due to this bug, an attribute with a single custom value will be shown in specifications and single value attributes.

#### Steps to reproduce:
1- Create a product. Add a `Free text` attribute.
2- Navigate to product page on the website.
3- From website editor, style tab, switch `specification` style.
4- With specification style set to `None`, you can see this attribute
   in single value attributes list.
5- With other 2 styles you can see this attributes in specifications.

Expected: In both steps 4 and 5, this attributes shouldn't be displayed there because it is already displayed in main attributes where you can set the value.

Initially fix was to filter `ProductTemplateAttributeLine` in `_prepare_single_value_for_display`, however it would cause
empty specification sections.

To avoid that we also need to filter out single custom value attributes in `_prepare_categories_for_display`.

opw-5484721

### DIFF
--- a/addons/website_sale/models/product_template_attribute_line.py
+++ b/addons/website_sale/models/product_template_attribute_line.py
@@ -18,8 +18,25 @@ class ProductTemplateAttributeLine(models.Model):
         The returned attributes are ordered as they appear in `self`, so based
         on the order of the attribute lines.
         """
+        single_value_lines = self.filtered(lambda ptal: (
+            len(ptal.value_ids) == 1
+            and ptal.attribute_id.display_type != 'multi'
+            and not ptal.value_ids.is_custom
+        ))
+        single_value_attributes = OrderedDict([(pa, self.env['product.template.attribute.line']) for pa in single_value_lines.attribute_id])
+        for ptal in single_value_lines:
+            single_value_attributes[ptal.attribute_id] |= ptal
+        return single_value_attributes
+
+    def _prepare_single_value_including_multi_type_for_display(self):
+        """On the product page group together the attribute lines that concern
+        the same attribute and that have only one value each.
+
+        Unlike `_prepare_single_value_for_display`, this method also includes
+        the attribute lines with a display type 'multi'
+        """
         single_value_lines = self.filtered(
-            lambda ptal: len(ptal.value_ids) == 1 and ptal.attribute_id.display_type != 'multi'
+            lambda ptal: len(ptal.value_ids) == 1 and not ptal.value_ids.is_custom
         )
         single_value_attributes = OrderedDict([(pa, self.env['product.template.attribute.line']) for pa in single_value_lines.attribute_id])
         for ptal in single_value_lines:
--- a/addons/website_sale_comparison/models/product_template_attribute_line.py
+++ b/addons/website_sale_comparison/models/product_template_attribute_line.py
@@ -26,3 +26,20 @@ class ProductTemplateAttributeLine(models.Model):
         for ptal in self:
             categories[ptal.attribute_id.category_id] |= ptal
         return categories
+
+    def _prepare_categories_for_display_in_specs_table(self):
+        """
+         Prepare attribute categories for display in a specs table.
+
+        Filters out attribute lines that have a single value and whose value is
+        marked as custom, then call _prepare_categories_for_display to group
+        the remaining attribute lines by category.
+
+        :return: OrderedDict [{
+        product.attribute.category: [product.template.attribute.line]
+        }]
+        """
+        filtered_self = self - self.filtered(
+            lambda ptal: len(ptal.value_ids) == 1 and ptal.value_ids.is_custom
+        )
+        return filtered_self._prepare_categories_for_display()
--- a/addons/website_sale_comparison/tests/test_website_sale_comparison.py
+++ b/addons/website_sale_comparison/tests/test_website_sale_comparison.py
@@ -254,3 +254,50 @@ class TestWebsiteSaleComparisonUi(HttpCase):
                 ]))
             ])),
         ]))
+
+    def test_single_value_attribute_specifications(self):
+        """Test that attribute with single custom value shouldn't be displayed in specifications
+        or single value attributes.
+        The attribute with 'multi type single value' attribute should be displayed in specs table
+        but not in single value attributes."""
+        custom_attribute = self.env['product.attribute'].create([{
+            'name': 'Write here',
+            'value_ids': [
+                Command.create({
+                    'name': 'Custom',
+                    'is_custom': True,
+                }),
+            ],
+        }])
+        multi_attribute = self.env['product.attribute'].create([{
+            'name': 'multi',
+            'display_type': 'multi',
+            'create_variant': 'no_variant',
+            'value_ids': [
+                Command.create({
+                    'name': 'multi type single value',
+                }),
+            ],
+        }])
+        product = self.env['product.template'].create([{
+            'name': 'T-Shirt',
+            'is_published': True,
+            'attribute_line_ids': [
+                Command.create({
+                    'attribute_id': custom_attribute.id,
+                    'value_ids': custom_attribute.value_ids,
+                }),
+                Command.create({
+                    'attribute_id': multi_attribute.id,
+                    'value_ids': multi_attribute.value_ids,
+                }),
+            ],
+        }])
+
+        self.assertFalse(product.attribute_line_ids._prepare_single_value_for_display())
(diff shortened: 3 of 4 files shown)


addons/website_sale_comparison/models/product_template_attribute_line.py
```diff
@@ -18,11 +18,14 @@ def _prepare_categories_for_display(self):
             product.attribute.category: [product.template.attribute.line]
         }]
         """
-        attributes = self.attribute_id
+        filtered_self = self - self.filtered(
+            lambda ptal: len(ptal.value_ids) == 1 and ptal.value_ids.is_custom
+        )
```
barracudapps: Not sure about it readability but could be like this:
```suggestion
        filtered_self = self.filtered(lambda ptal: not (len(ptal.value_ids) == 1 and ptal.value_ids.is_custom))
```

barracudapps (approved):
LGTM 👍 
Can you please just review your first step? Instead of "single custom value", I would write "as `Free text`" to ease the issue reproduction.
TBC during final review if this is indeed an unwanted behavior but it seems to be unexpected to me.

addons/website_sale_comparison/models/product_template_attribute_line.py
mma1377: @barracudapps That would make it more than 100 chars :D

ltinel: `_prepare_categories_for_display` isn't only used in the specs table (e.g. it's also used to conditionally show the "Add to comparison" button), so I don't think we can do this here (unless I misunderstood something?).

So can we use another condition to show/hide the specs table (and table headers)? Possibly via a new method that simply calls `_prepare_categories_for_display` but filters out single-custom-value ptals.

ltinel (changes requested):
Just one comment :) Thanks for your work!

addons/website_sale_comparison/models/product_template_attribute_line.py
ltinel: We should also call this method when displaying the specs at the bottom of the page
```suggestion
    def _prepare_categories_for_display_in_specs_table(self):
```

ltinel:
```suggestion
        Prepare attribute categories for display in a specs table.
```

addons/website_sale_comparison/views/website_sale_comparison_template.xml
ltinel: The other call in this file should also be updated.

addons/website_sale_comparison/models/product_template_attribute_line.py
ltinel, line 43: This condition is different from the one in `_prepare_single_value_for_display`, so it's still possible to have an empty specs table for PTALs with a single value and a "multi" display type (when selecting "in accordion" or "bottom of page" specs).

To be a bit more precise, on the product page, we show:
- PTALs with more than one value https://github.com/odoo/odoo/blob/618fde96d0a3526194b12cf734c999e4080e67f4/addons/website_sale_comparison/views/website_sale_comparison_template.xml#L137
- PTALs with a single non-custom value that doesn't have a "multi" display type https://github.com/odoo/odoo/blob/618fde96d0a3526194b12cf734c999e4080e67f4/addons/website_sale_comparison/views/website_sale_comparison_template.xml#L154-L157

But the method you added only filters out PTALs with a single custom value, but not PTALs with a single value and a "multi" display type, which could give an empty accordion (for "in accordion" specs) or a "Specifications" header with nothing below it (for "bottom of page" specs).

ltinel: After giving this a bit more thought, I'm not so sure we should remove PTALs with a single custom value, and PTALs with a single value and a "multi" display type from the specs table when selecting "in accordion" or "bottom of page" specs. At least not without discussing this with @Jean-boje first.

 I do agree, however, that we should remove both of them from when selecting "none" specs.

@Jean-boje @mma1377 WDYT? 🙂

addons/website_sale_comparison/tests/test_website_sale_comparison.py
ltinel: I'm not a big fan of this weird numbering in test names. Could you instead make the method name self-explanatory?

ltinel (changes requested):

addons/website_sale_comparison/models/product_template_attribute_line.py
mma1377: @ltinel  I agree that we need to filter out multi in `_prepare_categories_for_display_in_specs_table` as well.
Since in `_prepare_single_value_for_display` we already have:
https://github.com/odoo/odoo/blob/efe06fb71ca9d9376310ef9fd3f86b6a4e4b6698/addons/website_sale/models/product_template_attribute_line.py#L21-L23
IMO that attribute of single value of multi type needs to be filtered out in `_prepare_categories_for_display_in_specs_table` regardless of issue this PR is fixing. In a standard runbot, we are already having the issue that you pointed out that an empty `Specifications` is displayed at the bottom of the page when there is a single value attribute which is multi type. 😄 
I'm waiting for @Jean-boje 's response before make the modifications on this PR.

ltinel: We discussed this with @Jean-boje at the office :slightly_smiling_face: 
- In stable (18.2):
    - "none" specs should exclude PTALs with a single custom value, and PTALs with a single "multi" value (which is already the case in your PR).
    - "in accordion" and "bottom of page" specs should exclude PTALs with a single custom value, but include PTALs with a single "multi" value (which isn't the case yet in your PR).
- In master:
    - "none" specs shouldn't display anything at all.
    - "in accordion" and "bottom of page" specs should do the same as in stable.

@Jean-boje let me know if I missed something.
@mma1377 we can discuss this on Thursday if it's not clear.

addons/website_sale/models/product_template_attribute_line.py
```diff
@@ -8,7 +8,7 @@
 class ProductTemplateAttributeLine(models.Model):
     _inherit = 'product.template.attribute.line'
 
-    def _prepare_single_value_for_display(self):
+    def _prepare_single_value_for_display(self, include_multi_display_type=False):
```
ltinel: I'm afraid we can't do this in stable 🙁 You'll probably have to duplicate the method (same as you did with `_prepare_categories_for_display`)

ltinel (changes requested):

addons/website_sale_comparison/tests/test_website_sale_comparison.py
```diff
@@ -254,3 +254,28 @@ def test_03_category_order(self):
...
+            'is_published': True,
+            'attribute_line_ids': [
+                Command.create({
+                    'attribute_id': attribute.id,
+                    'value_ids': attribute.value_ids,
+                }),
+            ],
+        }])
+        self.assertFalse(product.attribute_line_ids._prepare_single_value_for_display())
+        self.assertFalse(product.attribute_line_ids._prepare_categories_for_display_in_specs_table())
```
ltinel: Optional: could you also assert that `_prepare_single_value_including_multi_type_for_display` includes multi types, and that `_prepare_single_value_for_display` excludes them?

ltinel (approved):
Nice, LGTM 😄 
@robodoo delegate+
Could you request a review from me for the FW port in master, since it will be significantly different?

addons/website_sale_comparison/tests/test_website_sale_comparison.py
mma1377: @ltinel 
Thanks for the review! Now I added the multi as well. Could you please review the test.

addons/website_sale_comparison/tests/test_website_sale_comparison.py
```diff
@@ -259,3 +254,48 @@ def test_03_category_order(self):
...
+                }),
+                Command.create({
+                    'attribute_id': multi_attribute.id,
+                    'value_ids': multi_attribute.value_ids,
+                }),
+            ],
+        }])
+
+        self.assertFalse(product.attribute_line_ids._prepare_single_value_for_display())
+
```
ltinel, line 300:
```suggestion
        self.assertIn(multi_attribute.attribute_line_ids, product.attribute_line_ids._prepare_single_value_including_multi_type_for_display().values())
```
(Id didn't test this so I'm not entirely sure I got it right, but you got the gist 😅 )

ltinel (approved):
LGTM, maybe just one more check in the test and that's all :) Feel free to r+

PR #244245, opened 2026-01-16
[FIX] project: set correct action target of project window

rapa-odoo:
Steps to reproduce:
1. Create a db with having 'sale' & 'project' installed in version 16.
2. Create a sale order having linkage to more than single project.
3. Migrate the db to version 19.
4. When clicking on the project stat button the breadcrumb traceability will not be there.

Issue:
-> In v16.4 the target defined for the action `project.open_view_project_all` is removed from [here](odoo/odoo@a92d686)

When migrating a database from v16 to v19 and opening projects from a sale order linked to multiple projects, the stat button triggers `action_view_project_ids`, which in turn calls
`project.open_view_project_all` for records having len('projects_ids') > 1 from [here]
(https://github.com/odoo/odoo/blame/19.0/addons/sale_project/models/sale_order.py#L220) Because the persisted target is `main`, breadcrumb traceability will be lost. The issue will arise in the DBs coming from version 16 or lesser. Therefore, it would be necessary to address this immediately and set correct target for window_action for databases >= v17

This commit explicitly sets the action target to `current` to restore proper breadcrumb behavior and align it with standard odoo record.

OPW-5448916

Current behavior before PR:

---

### DIFF
--- a/addons/project/views/project_project_views.xml
+++ b/addons/project/views/project_project_views.xml
@@ -600,6 +600,7 @@
             <field name="view_mode">kanban,tree,form</field>
             <field name="view_id" ref="view_project_kanban"/>
             <field name="search_view_id" ref="view_project_project_filter"/>
+            <field name="target">current</field>
             <field name="help" type="html">
                 <p class="o_view_nocontent_smiling_face">
                     No projects found. Let's create one!


rapa-odoo:
Hello @odoo/rd-services , could you please review this?

rapa-odoo:
> Hello @rapa-odoo
> 
> Thanks for your work!
> 
> However, I don't understand why we need to do that to me the default value is current and not main. So, I am a bit lost, TBH. 🤔

Actually, this PR is meant to address a behavior that exists in previous versions (< v17).

In older versions, the action `open_view_project_all` had its target explicitly set to main via [XML](https://github.com/odoo/odoo/blob/16.0/addons/project/views/project_views.xml#L987). Because this value is stored in the database, it gets carried forward during migrations, which leads to the breadcrumb issue in upgraded databases.

```sql
test_16=# select imd.name AS name,ia.name AS action_name, ia.target as tgt from ir_model_data imd join ir_act_window ia on imd.res_id = ia.id where imd.name='open_view_project_all';
         name          |      action_name      | tgt  
-----------------------+-----------------------+------
 open_view_project_all | {"en_US": "Projects"} | main
(1 row)
```

For fresh v17 databases, this problem does not occur. Since there is no legacy XML value to migrate, the target is assigned using the default value of the [field](https://github.com/odoo/odoo/blob/17.0/odoo/addons/base/models/ir_actions.py#L279) (i.e. current), and breadcrumbs work as expected.

```sql
Fresh database v17.
test_17_project=# select imd.name AS name,ia.name AS action_name, ia.target as tgt from ir_model_data imd join ir_act_window ia on imd.res_id = ia.id where imd.name='open_view_project_all';
         name          |      action_name      |   tgt   
-----------------------+-----------------------+---------
 open_view_project_all | {"en_US": "Projects"} | current
(1 row)
```

With this commit, we explicitly redefine and override the target for databases migrated from versions below v17.

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #243032, opened 2026-01-09
[FIX] base_import_module: mark if module was imported with demo

Xavier-Do:
Currently even if an imported module was imported with demo data, the 'demo' field is not set to True.

Setting this field properly will allow to have a better handling of demo data in industry tests.

PR #244246, opened 2026-01-16
[FIX] base_import_module: mark if module was imported with demo

trcazier:
Currently even if a imported module was imported with demo data, the 'demo' field is not set to True.

Setting this field proprely will allow to have a better handling of demo data in industry tests.

Backport of https://github.com/odoo/odoo/pull/243032


vava-odoo (approved):
@robodoo r+

PR #244247, opened 2026-01-16
[FIX] mail: avoid access error when archiving channel as non admin

jayp-odoo:
**Purpose of this PR:-**
 
Allow users to archive channels using `sudo` so the action is not blocked by missing discuss role access  error.

task-5478832

---


alexkuhn (approved):
@robodoo delegate+ r+

PR #244249, opened 2026-01-16
[FIX] crm: prevent error when marking lead as won without a valid stage

adip-odoo:
An error occurs when marking a lead as Won while it does not belong to any valid CRM stage.

**Steps to Reproduce:**
1. Install **CRM** module.
2. Create a new user.
3. Delete all the default crm stages.
4. Create a lead **older than 31 days** with no stage assigned (change system date if required).
5. Assign the new user as **Salesperson** and set **Expected Revenue = 0**.
6. Mark the lead as **Won**.

**Error:**
`ValueError - invalid literal for int() with base 10: 'false'`

**Cause:**
When a lead does not belong to a valid stage, the `stage_id` become False. When it tries to cast `stage_id` to an integer, an error will be raised.

**Fix:**
This commit ensures only valid numerical stage ids are processed.

sentry-7193424787

### DIFF
--- a/addons/crm/models/crm_lead.py
+++ b/addons/crm/models/crm_lead.py
@@ -1248,7 +1248,7 @@ class CrmLead(models.Model):
         # use duration tracking field to determine if the task jumped from first to last stage
         # only takes into accounts stages on which the lead has spent at least a minute,
         # to only account for valid stage movements
-        elif len(stage_ids := [int(stage_id) for stage_id, duration in self.duration_tracking.items() if duration >= 60]) == 1:
+        elif len(stage_ids := [int(stage_id) for stage_id, duration in self.duration_tracking.items() if str(stage_id).isdigit() and duration >= 60]) == 1:
             first_stage = self.env['crm.stage'].search([
                 '|', ('team_ids', 'in', False), ('team_ids', 'in', self.team_id.id),
             ], order='sequence ASC', limit=1)


addons/crm/models/crm_lead.py
```diff
@@ -1248,7 +1248,7 @@ def _get_rainbowman_message(self):
         # use duration tracking field to determine if the task jumped from first to last stage
         # only takes into accounts stages on which the lead has spent at least a minute,
         # to only account for valid stage movements
-        elif len(stage_ids := [int(stage_id) for stage_id, duration in self.duration_tracking.items() if duration >= 60]) == 1:
+        elif len(stage_ids := [int(stage_id) for stage_id, duration in self.duration_tracking.items() if str(stage_id).isdigit() and duration >= 60]) == 1:
```
maan-odoo, line 1251: We are getting `stage_id` as false because the lead does not have a stage assigned. In the `crm.lead` model, `_track_duration_field` is set to `stage_id`. The code below attempts to read the value of `stage_id`, but since it is empty, it evaluates to false.

https://github.com/odoo/odoo/blob/fe2615edb136c6b28c83a7ca2c29a810f6aff643/addons/mail/models/mail_tracking_duration_mixin.py#L116

adip-odoo: Indeed! :raised_hands: 
However, when a new stage is created at a later point, and an old lead (which previously had no stage) is moved to this newly created stage and then tries to be marked as Won, the `stage_id` remains None. This happens even though `self.duration_tracking` contains valid values.

IMO, instead of handling this at the Won action level, the fix should be applied where `self.duration_tracking` is being computed.

I’ve pushed a proposed fix to the test branch, and the runbot seems happy with it.
https://runbot.odoo.com/runbot/bundle/190-test-crm-none-stage-error-adip-431846

Please let me know your thoughts on this.
Thank you :)

PR #244253, opened 2026-01-16
[IMP] website: improve full-size popup with margin and background

assk-odoo:
The full-size option for popups could be misleading when using
transparent backgrounds, as the dialog content could visually merge
with the page and become hard to read.

This change keeps the same behavior as other popup sizes while adding a
10px margin on both sides of the snippet. It also applies a default white
background in full-size mode, ensuring better visual separation and improved readability.

| Before  | After |
| ------------- | ------------- |
| image | image |
| image | image |

task-5435802

### DIFF
--- a/addons/website/static/src/snippets/s_popup/001.scss
+++ b/addons/website/static/src/snippets/s_popup/001.scss
@@ -24,10 +24,9 @@
     .s_popup_size_full {
         padding: 0 !important;
         max-width: 100%;
+        margin: 0 10px !important;
 
         > .modal-content {
-            // Use the backdrop color as background-color
-            background-color: transparent;
             box-shadow: none;
             border-radius: 0;
         }


addons/website/static/src/snippets/s_popup/001.scss
```diff
@@ -24,10 +24,9 @@
     .s_popup_size_full {
         padding: 0 !important;
         max-width: 100%;
+        margin: 0 10px !important;
```
divy-odoo, line 27: Considering this changes, don't you think line number 37 should be moved inside `&:not(.s_popup_size_full) {` block?

```
    // Position option
    .modal-dialog {
        min-height: 100%;
        &:not(.s_popup_size_full) {
            margin: 0 0 0 auto;
            padding: $spacer !important;
        }
    }
```

divy-odoo (commented):
Hello @assk-odoo, Fix seems fine to me. But it should be target from 17.0. can you please check & re-target if needed?

Thanks : )

PR #244254, opened 2026-01-16
[FIX] pos_loyalty: prevent creating empty cards when not relevant.

sbel-odoo:
Currently, when you have multiple loyalty program but only 1 applies points on the current order, the other will have a loyalty card created with 0 points.

steps to reproduce:
-------------------
* Create a loyalty program giving you 1 point per $ on product 1
* Create a loyalty program giving you 1 point per § on product 2
* Make a pos order with just product 1, add a customer
* Check the programs in the backend
> Observation, a loyalty card with 0 points was create for the second
loyalty program

Why the fix:
------------
This creates a lot of unecessary records. We now only create cards where there are points or when there 0 points because a reward has been applied, thus keeping history.

opw-5405109


davidmonnom (approved):
@robodoo delegate+

PR #244261, opened 2026-01-16
[IMP] html_editor: Blocking browser plugin translation on editable content.

OmarKhalil2001:
Browser translation plugins were altering editable content by replacing the original content with translated versions, which caused issues when paired with auto-save. To prevent this behavior, the attribute `translate="no"` has been added to editable fields.

task-5485078

### DIFF
--- a/addons/html_editor/static/src/editor.js
+++ b/addons/html_editor/static/src/editor.js
@@ -139,6 +139,7 @@ export class Editor {
             }
         }
         editable.setAttribute("contenteditable", true);
+        editable.setAttribute("translate", "no");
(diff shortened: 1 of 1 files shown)


abd-msyukyu-odoo:
@OmarKhalil2001 Hello, sorry for the 3rd PR, but I asked Luc and we want to introduce this change as a fix starting from 17.0, let me know when it's done so I can send it :)

I don't remember but it's possible that version used the old `web_editor` module that was replaced by `html_editor` let me know if you have issues finding the correct place for the attribute and I'll have a look.

Cheers!

PR #244262, opened 2026-01-16
[FIX] sale_project: prevent stat button layout break on studio toggle

sami-odoo:
**Before this commit:**
When the user toggled studio, the `action_view_project_ids` stat button layout would break.

**After this commit:**
The stat button layout no longer breaks.

task-5480309

---


omra-odoo (approved):

kebeclibre:
@sami-odoo 
If there is an issue in Studio but not in the regular backend, I thing the fix should be in studio, and it makes sense that a piece of style is missing there....

PR #244268, opened 2026-01-16
[IMP] point_of_sale: optimize getPrice pricelist computation

rhe-odoo:
The getPrice function in product_template_accounting.js was a
performance bottleneck when using large pricelists. This commit
introduces the following optimizations:
1. Pre-index pricelist rules by product_id and product_tmpl_id at
   load time in ProductPricelist.
2. Replace the sort() operation with a single-pass scan to find
   the best matching rule (highest min_quantity that satisfies the
   current quantity).
3. Apply sequential rule precedence (Variant → Template → General),
   stopping as soon as a valid rule is found.
4. Move rule mappings from uiState to direct properties in
   ProductPricelist to avoid useless reactivity.
5. Centralize rule selection logic in ProductPricelist for better
   responsibility separation.

task-id: 5965826

---

### DIFF
--- a/addons/point_of_sale/static/src/app/models/accounting/product_template_accounting.js
+++ b/addons/point_of_sale/static/src/app/models/accounting/product_template_accounting.js
@@ -89,24 +89,43 @@ export class ProductTemplateAccounting extends Base {
             quantity = related_lines.reduce((sum, line) => sum + line.getQuantity(), 0);
         }
 
-        const tmplRules = (productTmpl.backLink("<-product.pricelist.item.product_tmpl_id") || [])
-            .filter((rule) => rule.pricelist_id.id === pricelist.id && !rule.product_id)
-            .sort((a, b) => b.min_quantity - a.min_quantity);
-        const productRules = (product?.backLink?.("<-product.pricelist.item.product_id") || [])
-            .filter((rule) => rule.pricelist_id.id === pricelist.id)
-            .sort((a, b) => b.min_quantity - a.min_quantity);
+        let rule = null;
 
-        const tmplRulesSet = new Set(tmplRules.map((rule) => rule.id));
-        const productRulesSet = new Set(productRules.map((rule) => rule.id));
-        const generalRulesIds = pricelist.getGeneralRulesIdsByCategories(this.parentCategories);
-        const rules = this.models["product.pricelist.item"]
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/static/src/app/models/accounting/product_template_accounting.js
```diff
@@ -89,24 +89,46 @@ export class ProductTemplateAccounting extends Base {
...
+            let bestRule = null;
+            for (const rule of rules) {
+                if (!rule.min_quantity || rule.min_quantity <= quantity) {
+                    if (!bestRule || rule.min_quantity > bestRule.min_quantity) {
+                        bestRule = rule;
+                    }
+                }
+            }
+            return bestRule;
+        };
```
stva-odoo: This method could be extracted to avoid recreating it on each call.

```
function findBestRule(rules, quantity){
 let bestRule = null;
    for (const rule of rules) {
        const minQty = rule.min_quantity ?? 0;

          if (minQty <= quantity && 
              (!bestRule || minQty > (bestRule.min_quantity ?? 0))) {
              bestRule = rule;
          }
      }
    return bestRule;
}
```

addons/point_of_sale/static/src/app/models/product_pricelist.js
```diff
@@ -10,12 +10,14 @@ export class ProductPricelist extends Base {
         this.uiState = {
             generalRulesIdsByCateg: {},
             generalRulesIds: {},
+            rulesByProductId: {},
+            rulesByTmplId: {},
```
stva-odoo: I’m not sure we should store this in uiState.
Could you use this.rulesByProductId = {} or this.rulesByTmplId = {} instead?
And in master, move the other fields out of uiState as well?

stva-odoo (commented):
Great improvement thanks 👍

addons/point_of_sale/static/src/app/models/product_pricelist.js
rhe-odoo: i finally move them all because it was a small change for all of them and centralized only in this specific file.

addons/point_of_sale/static/src/app/models/accounting/product_template_accounting.js
rhe-odoo: Good catch.
Instead of moving it globally in the file, i decided to move it directly to where it should belong, the pricelist itself.

PR #244270, opened 2026-01-16
[IMP] website: show image alt SEO status on translated pages

chdh-odoo:
Allow SEO optimization popup to detect and display image alt attributes on translated pages where oe branding metadata is
missing by default.

Add a fallback to derive the record from the root ir.ui.view when the page is translatable but not editable, ensuring users can see existing alt values and missing warnings across all languages.

task-5480952


chdh-odoo:
closing this due to conflict with base :cry:

PR #216737, opened 2025-06-30
[FIX] mail: use alias local parts when detecting new partners

ryce-odoo:
Sending an email that was auto-forwarded would lead to the recipient line having two different emails which would cause any emails that were not set up as aliases to be added as followers even if local part detection was enabled in the alias settings. This was due to the fact that partner detection was changed in 18.2 and now looked for exact email to alias matches. Ban emails were passed to the _find_or_create_from_emails and then directly matched to the emails in the list of recipient emails. When a match was not found for emails that did not have an exact matching alias_full_name, we would then look for or create a partner for that email. This caused erroneous followers to be added.

Changing the functionality to also look for matching local parts in order to skip partner finding and creation reverts this functionality to how it previously worked where extra recipients with matching local parts when local part detection was enabled would not add those partners as followers.

opw-4896074

PR #244272, opened 2026-01-16
[FIX] mail: ensure _find_aliases respects allowed domains for localpart aliases

jorv-odoo:
The fix introduced in https://github.com/odoo/odoo/pull/216737 can lead to "over-eager" filtering when an external email address matches a localpart (left part) alias in a input email list contains internal emails (aliases to filter) AND external email addresses (should not be filtered).

The `_find_aliases` method is used to identify internal system emails (aliases, bounces, catchalls) to prevent mail loops and ensure correct recipient filtering during notification grouping.

Before this fix, when the `mail.catchall.domain.allowed` system parameter was set, the logic for local-part aliases (where `alias_incoming_local` is True) failed to correctly associate the local part with the allowed domains. This resulted in external email addressed being returned by the system, potentially leading to incorrect notification routing.

We now use a more robust approach:
- Pre-filter local parts based on the allowed domains to reduce DB load.
- Utilize Python Sets for O(1) lookups of static and local aliases
- Explicitly validate the (local_part, domain) combo during the final filtering.

Example Scenario:
- Config: mail.catchall.domain.allowed = "test1.com,test2.com"
- Alias: "info" (alias_incoming_local=True)
- Input: ["info@test1.com", "info@test3.com"]

### Output Before Fix:
["info@test1.com", "info@test3.com"]
(The function failed to recognize info@test3.com as an external alias to be ignored based on the `mail.catchall.domain.allowed` config)

### Output After Fix:
["info@test1.com"]
(Correctly identifies the internal alias tob filtered while ignoring the external one)

OPW-5469264
OPW-5504201

---

### DIFF
--- a/addons/mail/models/mail_alias_domain.py
+++ b/addons/mail/models/mail_alias_domain.py
@@ -189,12 +189,15 @@ class MailAliasDomain(models.Model):
         if not filtered_emails:
             return filtered_emails
         all_domains = self.search([])
-        aliases = all_domains.mapped('bounce_email') + all_domains.mapped('catchall_email') + all_domains.mapped('default_from_email')
+        aliases = set(all_domains.mapped('bounce_email') +
+                    all_domains.mapped('catchall_email') +
+                    all_domains.mapped('default_from_email'))
 
-        catchall_domains_allowed = list(filter(None, (self.env["ir.config_parameter"].sudo().get_param(
-            "mail.catchall.domain.allowed") or '').split(',')))
+        # Get allowed domains and convert to a set for O(1) lookup
+        catchall_params = self.env["ir.config_parameter"].sudo().get_param("mail.catchall.domain.allowed") or ''
+        catchall_domains_allowed = set(filter(None, catchall_params.split(',')))
         if catchall_domains_allowed:
-            catchall_domains_allowed += all_domains.mapped('name')
+            catchall_domains_allowed.update(all_domains.mapped('name'))
             email_localparts_tocheck = [
                 email.partition('@')[0] for email in filtered_emails if (
                     email and email.partition('@')[2] in catchall_domains_allowed
@@ -209,15 +212,23 @@ class MailAliasDomain(models.Model):
             ('alias_full_name', 'in', filtered_emails),
             '&', ('alias_name', 'in', email_localparts_tocheck), ('alias_incoming_local', '=', True),
         ])
-        # global alias: email match
-        aliases += potential_aliases.filtered(lambda x: not x.alias_incoming_local).mapped('alias_full_name')
-        # compat-mode alias: left-part only (filter on allowed domains already done)
-        local_alias_names = potential_aliases.filtered(lambda x: x.alias_incoming_local).mapped('alias_name')
-        return [
-            email for email in filtered_emails if (
-                email in aliases or
-                email.partition('@')[0] in local_alias_names
-            )]
+        # Global aliases match by full name
+        aliases.update(potential_aliases.filtered(lambda x: not x.alias_incoming_local).mapped('alias_full_name'))
+
+        # Local aliases for validated local part + domain checking
+        local_alias_names = set(potential_aliases.filtered(lambda x: x.alias_incoming_local).mapped('alias_name'))
+
+        res = []
+        for email in filtered_emails:
+            if email in aliases:
+                res.append(email)
+                continue
+
+            local_part, _, domain = email.partition('@')
+            if local_part in local_alias_names and (not catchall_domains_allowed or domain in catchall_domains_allowed):
+                res.append(email)
+
+        return res
 
     @api.model
     def _migrate_icp_to_domain(self):
--- a/addons/mail/tests/test_mail_tools.py
+++ b/addons/mail/tests/test_mail_tools.py
@@ -101,6 +101,26 @@ class TestMailTools(MailCommon):
         self.assertEqual(found.email_normalized, 'test_no_localpart@gmail.com')
         self.assertEqual(found.name, 'Customer')
 
+        # test if ICP still works when passing a list of emails
+        test_list = [
+            '"Customer" <test_localpart@gmail.com>',
+            '"Customer" <test_localpart@tartopoils.com>',
+            '"Customer" <test_localpart@brutijus.com>',
+            '"Customer" <test_localpart@brutijus.fr.com>',
+        ]
+
+        found = self.env['mail.thread']._partner_find_from_emails_single(test_list, no_create=False)
+        self.assertEqual(
+            len(found),
+            2,
+            "Should have found 2 partners, as tartopoils.com and brutijus.com are limiting local part alias recognition, limiting alias conflict check to those domains",
+        )
+        self.assertEqual(
+            found.mapped("email_normalized"),
+            ["test_localpart@gmail.com", "test_localpart@brutijus.fr.com"],
+            "Found Partners have wrong normalized email addresses",
+        )
+
     @users('employee')
     def test_mail_find_partner_from_emails_followers(self):
         """ Test '_mail_find_partner_from_emails' when dealing with records on


addons/test_mail/tests/test_mail_thread_internals.py
```diff
@@ -457,6 +457,39 @@ def test_message_get_default_recipients(self):
                 'Mail: prioritize email should not return partner if email is found'
             )
 
+    @users('employee')
+    def test_find_localpart_aliases_vs_allowed_catchall_domain(self):
```
reth-odoo: Not a bad idea to have a specific test but could we also update the test added by the other PR?
(if it's short to do)

reth-odoo (commented):
Indeed that seems to make sense, I suppose this is an issue when conversing with other companies that might use generic names like "info" and "contact" on their side as well?

addons/test_mail/tests/test_mail_thread_internals.py
jorv-odoo: Indeed that is the main issue: any outside email recipient using "generic" local-parts (contact, help, info etc.) that are also susceptible to be used by the Odoo setup will be blacklisted (accidentally).

Back from vacay, so give me a few days to warm up and I will see if I can adapt the original test too ;)

Thx for the feedback @reth-odoo !

jorv-odoo: Hoi @reth-odoo ,

did a quick analysis of the tests changed in the original PR that introduced the `_find_aliases` changes:
* `test_mail_find_partner_from_emails_alias_localpart` -> it explicitly calls `_partner_find_from_emails_single`, i.e. will only use one email address, thus the bug in question will never occur (as it only happens when passing a n>1 list of emails)
* `test_message_get_default_recipients_banned` -> I guess I can add an explicit
```
        self.env['ir.config_parameter'].sudo().set_param(
            'mail.catchall.domain.allowed',
            f'{self.test_aliases[2].alias_domain},other.domain'
        )
  ```
  And make sure that the test cases are still correctly banned. 
  But given the design of the test (negative/empty expected outputs), I *could* add a positive outcome (default recipient is returning something when using a domain not in `mail.catchall.domain.allowed`), but that would change the overall pattern of the test. 
  
 So let me know if you want me to add that second addition, or if it's even worth the hassle ;)

Comment (reth-odoo) [reply addons/test_mail/tests/test_mail_thread_internals.py]: Doesn't sound short to do so fine like this
Thanks for looking into it :)

Comment (reth-odoo) [inline addons/test_mail/tests/test_mail_thread_internals.py]: 🔤 order 🧑‍⚖️

**Review by reth-odoo — APPROVED** [review]
@tde-banana-odoo small imp to avoid a good chunk of issues with `alias_incoming_local`

(though arguably the real fix is to not use it if you have this issue)

File: addons/test_mail/tests/test_mail_thread_internals.py (hunk 1 above)
Comment (jorv-odoo) [reply addons/test_mail/tests/test_mail_thread_internals.py]: Fixed (but now it's farther away form it's brethren tests ^)

Comment (reth-odoo) [reply addons/test_mail/tests/test_mail_thread_internals.py]: Renaming it to be closer is also an option :disguised_face: (possibly, don't remember the name of the other ones ^^')

File: addons/test_mail/tests/test_mail_thread_internals.py [hunk 2]
```diff
@@ -194,6 +194,39 @@ def test_body_escape(self):
         ticket_record._message_update_content(message, "Hello <R&D/>")
         self.assertEqual(message.body, Markup('<p>Hello &lt;R&amp;D/&gt;<span class="o-mail-Message-edited"></span></p>'))
 
+    @users('employee')
+    def test_find_localpart_aliases_vs_allowed_catchall_domain(self):
```
tde-banana-odoo: Isn't it possible to update 'test_mail_find_partner_from_emails_alias_localpart' test with a minimal changes so that we limit a bit test addition ?

addons/mail/models/mail_alias_domain.py
```diff
@@ -189,12 +189,15 @@ def _find_aliases(self, email_list):
...
         all_domains = self.search([])
-        aliases = all_domains.mapped('bounce_email') + all_domains.mapped('catchall_email') + all_domains.mapped('default_from_email')
+        aliases = set(all_domains.mapped('bounce_email') +
+                    all_domains.mapped('catchall_email') +
+                    all_domains.mapped('default_from_email'))
 
-        catchall_domains_allowed = list(filter(None, (self.env["ir.config_parameter"].sudo().get_param(
-            "mail.catchall.domain.allowed") or '').split(',')))
+        # Get allowed domains and convert to a set for O(1) lookup
+        catchall_params = self.env["ir.config_parameter"].sudo().get_param("mail.catchall.domain.allowed") or ''
```
tde-banana-odoo, line 197: Why this change (aka: useless variable ?)

tde-banana-odoo (commented):
Zboing !

addons/mail/models/mail_alias_domain.py
jorv-odoo: Mostly asthetics/comprehension.

Instead of running a big one-liner, we explicitly show that we are grabbing the chatchall ICP and then splitting it by "," when defining the set for `catchall_domains_allowed` (the next line).

I find it cleaner, but I will not die on a hill for this, if you want me to make it a one-liner again, I can :)

addons/test_mail/tests/test_mail_thread_internals.py
jorv-odoo: Indeed It seems possible.

I pushed a new commit where I added an extra section to `test_mail_find_partner_from_emails_alias_localpart`. Let me know if it looks good, then I will delete `test_find_localpart_aliases_vs_allowed_catchall_domain` and repush a clean commit ;)

addons/mail/models/mail_alias_domain.py
tde-banana-odoo: No it's ok for me, not a big deal :)

addons/mail/models/mail_alias_domain.py
```diff
@@ -209,15 +212,24 @@ def _find_aliases(self, email_list):
...
+
+        res = []
+        for email in filtered_emails:
+            if email in aliases:
+                res.append(email)
+                continue
+
+            local_part, _, domain = email.partition('@')
+            if local_part in local_alias_names:
+                if not catchall_domains_allowed or domain in catchall_domains_allowed:
```
tde-banana-odoo: Could concatenate (gain one indent level, and we understand it is a single global condition)
```suggestion
            if local_part in local_alias_names and (
                not catchall_domains_allowed or domain in catchall_domains_allowed
            ):
```

addons/mail/tests/test_mail_tools.py
```diff
@@ -101,6 +101,23 @@ def test_mail_find_partner_from_emails_alias_localpart(self):
...
+        # test if ICP still works when passing a list of emails
+        test_list = [
+            '"Customer" <test_localpart@gmail.com>',
+            '"Customer" <test_localpart@tartopoils.com>',
+            '"Customer" <test_localpart@brutijus.com>',
+            '"Customer" <test_localpart@brutijus.fr.com>',
+        ]
+
+        found = self.env['mail.thread']._partner_find_from_emails_single(test_list, no_create=False)
+        self.assertTrue(found, 'Should have created/found a partners')
```
tde-banana-odoo: Useless assert (implied by the one below)

addons/mail/tests/test_mail_tools.py
```diff
@@ -101,6 +101,23 @@ def test_mail_find_partner_from_emails_alias_localpart(self):
...
+        self.assertEqual(len(found), 2, 'Should have found 2 partners')
```
tde-banana-odoo: Message should contain a bit more explanation when possible (I know most of tests lack explanations but hey, you asked for a review :D )
```suggestion
        self.assertEqual(len(found), 2, 'Should have found 2 partners, as tartopoils.com and brutijus.com are limiting local part alias recognition, limiting alias conflict check to those domains')
```

addons/test_mail/tests/test_mail_thread_internals.py
tde-banana-odoo: Other test seems ok to me (especially if it is red without the fix and green with, obviously :D )

addons/mail/tests/test_mail_tools.py
jorv-odoo: Added suggestion!

addons/mail/models/mail_alias_domain.py
jorv-odoo: Added suggestion!

PR #76734, opened 2021-09-17
[MERGE] mail, various: support multi domains aliases

tde-banana-odoo:
RATIONALE

Currently only one alias domain is possible when using Odoo. Even if several
outgoing and/or incoming email servers can be used, all "reply-to" email
addresses belong to the same global email domain. Moreover incoming emails
cannot be cleanly limited or checked against a company as we accept all
incoming emails sent to aliases to ease notably email forwarding.

PURPOSE

Allow alias domains to be multiple, notably to be used in a multi company
environment where each company has its own alias domain.

SPECIFICATIONS

Allow companies sharing a database to handle their mailing flow independently
e.g if a travel agency group uses Odoo to handle its business. Each agency is
a company with their mail domains and servers like 'info@travel.namur.example.com'
'info@travel.liege.example.com'.

ALIAS DOMAINS

Add a new 'mail.alias.domain' model storing information previously stored
as a unique value in 'mail.catchall.domain' configuration parameter. Domains
are not completely linked to companies, allowing to have a mono-domain MC
setup, or mono-company multi-domain setup.

A main email domain is present on companies for all default domain computation
when being in that company while keeping flexibility of having other domains
defined.

Now that a new 'mail.alias.domain' model exists we move bounce and catchall
aliases definition directly on this model. Update related computation on
company model. Default_from (filter) are also moved on this model, allowing
an higher level default from computation for mail servers when having knowledge
of alias domain environment.

Usage of 'mail.catchall.domain' will soon be completely removed to be replaced
by proposed company-based email alias domain usage.

Constraints are added so that each bounce and catchall defined on domains
do not clash with existing aliases. Sanitize of bounce and catchall is also
performed to ensure they make valid emails. This matches previous behavior
of ICP parameters. Domain name is also sanitized like alias names.

ALIASES

Add 'mail.alias.domain' information on alias model. Aliases do not use global
configuration parameters anymore. Instead they are linked to an alias domain
e.g. 'sales' linked to 'mycompany.com' alias domain: 'sales@mycompany.com'.

It is now considered as a different alias compared to 'sales@mycompany.in'
which has the same alias_name but a different alias domain.

Constraints and checks are added, as

  * uniqueness of aliases is now checked inside a given domain;
  * an alias name must not clash with its domain bounce or catchall;
  * a combination of (alias_name, alias_domain_id) must be unique;

Now that alias domains are multiple and can be linked to companies it is
easy to end up with configuration where an alias is using a domain of
[...]

PR #244279, opened 2026-01-16
[FIX] mail: run sanitize for mail.catchall.domain.allowed

jorv-odoo:
Since its introduction (https://github.com/odoo/odoo/pull/76734), the "mail.catchall.domain.allowed" system parameter is normally sanitized by `_sanitize_allowed_domains` when using `set_param`.

But the method is not run when the system parameter is created manually, for example through the UI, because the current conditional logic is not triggered for write and create call.

This can be an issue if for example the use submits this value: "domain1.com, domain2.com,domain3.com"

The comma seperated list will only be split by ',', meaning that the second element is " domain2.com" (space character in front). As such, that domain will not be used as expected, since most flows using that data will end up comparing "domain2.com" vs " domain2.com", and not considering it a match.

Before this fix:
Create new record in System parameters with
key = "mail.catchall.domain.allowed"
value = "domain1.com, domain2.com,domain3.com"
-> after save, value == "domain1.com, domain2.com,domain3.com"

After this fix:
-> after save, value == "domain1.com,domain2.com,domain3.com"

OPW-5505414

---

### DIFF
--- a/addons/mail/models/ir_config_parameter.py
+++ b/addons/mail/models/ir_config_parameter.py
@@ -90,3 +90,24 @@ class IrConfigParameter(models.Model):
             value = self.env['mail.alias']._sanitize_allowed_domains(value)
 
         return super().set_param(key, value)
+
+    def _sanitize_param_value(self, key, value):
+        """ Dispatcher for sanitization logic """
+        if key == 'mail.catchall.domain.allowed' and value:
+            return self.env['mail.alias']._sanitize_allowed_domains(value)
+        return value
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get('key') and 'value' in vals:
+                vals['value'] = self._sanitize_param_value(vals['key'], vals['value'])
+        return super().create(vals_list)
+
+    def write(self, vals):
+        if 'value' in vals:
+            for record in self:
+                # Determine the key: from vals if changing, otherwise from the record
+                key = vals.get('key', record.key)
+                vals['value'] = self._sanitize_param_value(key, vals['value'])
+        return super().write(vals)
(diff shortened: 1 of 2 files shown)


addons/mail/models/ir_config_parameter.py
```diff
@@ -90,3 +90,24 @@ def set_param(self, key, value):
...
         return super().set_param(key, value)
+
+    def _sanitize_param_value(self, key, value):
+        """ Dispatcher for sanitization logic """
+        if key == 'mail.catchall.domain.allowed' and value:
+            return self.env['mail.alias']._sanitize_allowed_domains(value)
+        return value
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
tde-banana-odoo, line 101: :ghost: Unit tests !

jorv-odoo: Added the unit test, let me know if it's good enough 🦘

addons/test_mail/tests/test_mail_alias.py
```diff
@@ -47,15 +47,53 @@ def test_alias_domain_allowed_validation(self):
...
+                    "value": value,
+                })
+                self.assertEqual(
+                    self.env["ir.config_parameter"].get_param(
+                        "mail.catchall.domain.allowed"
+                    ),
+                    expected,
+                )
+
+        # test write sanitization
```
tde-banana-odoo: Could write "caca" on newly-created parameter in the "create" loop, then write old value again (that way we test create and write in a single loop on test cases)

jorv-odoo: Done (minus the explicit use of "caca")

tde-banana-odoo: But caca is great.

jorv-odoo: Yeah, but then people might think that BVE wrote the fix! Can't have that!

tde-banana-odoo: Seems legit ^^ .

PR #244280, opened 2026-01-16
[FIX] point_of_sale: rounding of tax base amounts in reports before sum

pedrambiria:
Before this commit, the total base amounts of taxes might be different from summing the base amounts of each line, due to rounding issues.

opw-5408201

---


stva-odoo (approved):
robodoo r+

PR #244281, opened 2026-01-16
[IMP] account: generalize introduce new test helper to pay invoices with bank statement lines

oco-odoo:


hupo-odoo (approved):
@robodoo r+

PR #244286, opened 2026-01-16
[FIX] hr_work_entry_holidays: Error when writing contract with multiple leaves

kareemoo:
Issue: when you change in a contract for an employee with multiple leaves and one of them is overlapping multiple contracts, you get a validation error which shouldn't be the case

Steps to reproduce:
- create an employee with a contract ending on a day x
- create 2 leaves for the employee one from x-5 to x+5 and the second from x+10 to x+15 and approve them
- create another contract with a different calendar for the employee starting on a day x+1
- when you move the second contract to the running state, you get a validation error but you shouldn't and the operation should be done normally

Fix:
- ensured that the write of vals in contract is only done after refusing leaves that span multiple contracts
- changed the employee in `test_leave_change_working_schedule` and ensured that all timezones are in Europe/Brussels (jules_emp's tz was changing between UTC and Europe/Brussels depending on demo data which caused a runbot error)
- created another leave in `test_leave_multi_contracts_split` to ensure that the split works correctly with more than one leave

task-id: 5499527

---

### DIFF
--- a/addons/hr_work_entry_holidays/models/hr_contract.py
+++ b/addons/hr_work_entry_holidays/models/hr_contract.py
@@ -92,6 +92,7 @@ class HrContract(models.Model):
         all_new_leave_origin = []
         all_new_leave_vals = []
         leaves_state = {}
+        vals_calendar = self.env['resource.calendar'].browse(vals.get('resource_calendar_id', False))
         # In case a validation error is thrown due to holiday creation with the new resource calendar (which can
         # increase their duration), we catch this error to display a more meaningful error message.
         try:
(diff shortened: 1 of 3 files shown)


addons/hr_work_entry_holidays/models/hr_contract.py
```diff
@@ -110,14 +110,17 @@ def write(self, vals):
...
                     ]).sorted(key=lambda c: {'open': 1, 'close': 2, 'draft': 3, 'cancel': 4}[c.state])
-                    if len(overlapping_contracts.resource_calendar_id) <= 1:
-                        super(HrContract, contract).write(vals)
-                        if overlapping_contracts and leave.resource_calendar_id != overlapping_contracts[0].resource_calendar_id:
-                            leave.resource_calendar_id = overlapping_contracts[0].resource_calendar_id
-                            if not leave.request_unit_hours:
-                                leave.with_context(leave_skip_date_check=True, leave_skip_state_check=True)._compute_date_from_to()
-                                if leave.state == 'validate':
-                                    leave._validate_leave_request()
+                    vals_calendar = self.env['resource.calendar'].browse(vals.get('resource_calendar_id', False))
```
Einar-eian: Can't this be moved out of the loop as it has the same value on every iteration?

addons/hr_work_entry_holidays/models/hr_contract.py
```diff
@@ -110,14 +110,17 @@ def write(self, vals):
...
+                    all_calendars = overlapping_contracts.resource_calendar_id if not vals_calendar else (overlapping_contracts - self).resource_calendar_id | vals_calendar
+                    if len(all_calendars) <= 1:
```
Einar-eian, line 115: I'm wondering here, if every leave has `all_calendars` value as 1 or less (so the execution enters this branch), the super().write() will never be called. Is that scenario possible or am I thinking of something impossible?

Einar-eian (commented):
Thank you for your work💪 😄 I left two comments/questions but otherwise looking good!

addons/hr_work_entry_holidays/models/hr_contract.py
kareemoo: similar to the other pr, the return statement at the end of the function handles this scenario

addons/hr_work_entry_holidays/tests/test_multi_contract.py
```diff
@@ -146,6 +146,11 @@ def test_leave_multi_contracts_split(self):
         self.assertEqual(leave.state, 'validate')
 
         self.contract_cdi.date_end = date(2022, 6, 15)
+        # create another leave to ensure that we only write the state
+        # to the contract after ensuring that overlapped leaves are
+        # splitted correctly and not with other leaves while looping
```
Einar-eian: Small non-important detail 👀 
```suggestion
        # split correctly and not with other leaves while looping
```

tivisse:
@robodoo help

tivisse:
@robodoo up to saas-18.3

tivisse:
@robodoo fw=skipmerge

PR #244287, opened 2026-01-16
[FIX] point_of_sale: create order for cash move with correct values

pedrambiria:
Before this commit, the order for cash moves was created without setting the session_id, company_id, and user_id fields.

opw-5500966

---


adgu-odoo (commented):
robodoo delegate +

pedrambiria:
robodoo r+

PR #244289, opened 2026-01-16
[FIX] mail: enable other views when opening records

flch-odoo:
Records of some models may not want to be shown in their form
view by default. See related ENT PR for documents.

We re-export to avoid patching order issues. In particular,
this makes sure that if we are in the webclient, the chat
window is opened before executing the "real" open.

Task-5386466

### DIFF
--- a/addons/mail/static/src/core/web/thread_model_patch.js
+++ b/addons/mail/static/src/core/web/thread_model_patch.js
@@ -75,15 +75,18 @@ patch(Thread.prototype, {
             return;
         }
         if (this.model !== "discuss.channel") {
-            this.store.env.services.action.doAction({
-                type: "ir.actions.act_window",
-                res_id: this.id,
-                res_model: this.model,
-                views: [[false, "form"]],
-            });
+            this.store.env.services.action.doAction(this.openRecordActionRequest);
             return;
         }
         super.open();
+    },
+    get openRecordActionRequest() {
+        return {
+            type: "ir.actions.act_window",
+            res_id: this.id,
+            res_model: this.model,
+            views: [[false, "form"]],
+        };
     },
     async unpin() {
         const chatWindow = this.store.ChatWindow.get({ thread: this });


addons/mail/static/src/core/web/thread_actions.js
```diff
@@ -48,14 +48,13 @@ threadActionsRegistry
             component.actionService = useService("action");
         },
         icon: "fa fa-fw fa-expand",
-        name: _t("Open Form View"),
+        name: _t("Open"),
```
alexkuhn: Changing name in stable is not a good idea, people that do not care for document scope would be affected by this new renaming that is confusing.

Couldn't you keep `Open Form View` for all but the affected flow? I don't mind this change in master at all (IMO I would name it "Open Record" for more clarity), but a change in 18.0 this late, this feels too much for such a FIX.

flch-odoo: Thanks for the comment!

> Couldn't you keep Open Form View for all but the affected flow?

No, I would need to patch multiple places instead of just one to detect it there. I can keep the current name and introduce the oddity that it opens a different view for the affected flows... I didn't imagine that some users would be puzzled by "Open" vs "Open Form View", but I stand corrected and will make the change.

(I would also personally prefer "Open Record", but I'm not sure it's that good for a common user)

alexkuhn: There's ambiguity that the conversation is already open... in a chat window. We have "Open in Discuss" and "Open Form View" to clarify this will open the conversation in another context. "Open Record" is the generic name attributed to business objects that are shown using Odoo view.

aab-odoo (changes requested):
Hello @flch-odoo 

Side not first: this doesn't look like a bug fix to me, but rather an improvement to do in master. I don't know why/how/when it has been decided to change that in stable, as it seems that originally, it was targeting master.

About the diff itself, 18 or master, I don't want to see that change done in action service, and I don't think it is necessary to do it either. Here's an alternative solution for documents that doesn't require any change in the framework:
```js
import { WebClient } from "@web/webclient/webclient";
import { router } from "@web/core/browser/router";

patch(WebClient.prototype, {
    /**
     * @override to not open documents in form view.
     */
    loadRouterState() {
        const { resId, model } = router.current.actionStack?.at(-1) || {};
        if (resId && model === "documents.document") {
            // stable-proof solution while we don't yet have document_action_preference or better
            return this.actionService.doAction({
                type: "ir.actions.act_window",
                res_model: model,
                res_id: resId,
                views: [[false, "kanban"]],
                context: {
                    documents_init_document_id: resId,
                },
            });
        }
        return super.loadRouterState(...arguments);
    },
});
```

I'm not a big fan of patching the WebClient obviously, but that's a bit better than patching the action service + the quite ugly refactoring that comes with it. And it better targets the usecase as it patches `loadRouterState`, which is exactly what you want, instead of an internal function of the action service which is also called in other situations.

We explored an alternative solution with @jpp-odoo which consists in patching `urlToState` (as you already patch `stateToUrl`), but that would require a small (but breaking) change in router.js which we will definitely not do in stable. It would look to something like this
```diff
diff --git a/documents/static/src/core/router.js b/documents/static/src/core/router.js
index d533e4aef27..1725a7c5ff8 100644
--- a/documents/static/src/core/router.js
+++ b/documents/static/src/core/router.js
@@ -47,4 +47,17 @@ patch(router, {
         }
         return url;
     },
+
+    urlToState() {
+        const state = super.urlToState(...arguments);
+        const { resId, model } = state.actionStack?.at(-1) || {};
+        if (resId && model === "documents.document") {
+            return {
+                actionStack: [{ action: "documents", documents_init_document_id: resId }],
+                action: "documents",
+                documents_init_document_id: resId,
+            };
+        }
+        return state;
+    },
 });
```

addons/mail/static/src/core/web/thread_model_patch.js
```diff
@@ -75,19 +75,24 @@ patch(Thread.prototype, {
...
+        };
+    },
     async unpin() {
         const chatWindow = this.store.ChatWindow.get({ thread: this });
         await chatWindow?.close();
         super.unpin(...arguments);
     },
 });
+
+export { Thread };
```
seb-odoo: This should not be re-exported here. Import from the original file instead, and import this file without return value to ensure patching order.

seb-odoo (commented):
Split getter to override ok.

seb-odoo (approved):
@robodoo delegate+ r+

PR #244290, opened 2026-01-16
[FIX] sale_loyalty: keep sol.price_unit after reward re-claimed

mma1377:
### Issue:
Due to this issue, by clicking twice on `Reward` button on SO, reward line unit price will be reset to zero.

#### To reproduce:
1- Create a `Buy X Get Y` program:
   rule: minimum quantity: 3, product: `Large Desk Wood`
   reward: 1 `Large Desk Wood` for free
2- Create a SO, and add a line with 3 `Large Desk Wood`
3- Click on `Reward` button. A new line should be automatically
   created. The unit price should be the product sale price and
   line `Amount` should be 0.
4- Re-click on reward. The reward line price unit is set to zero.

Expected: The reward line price unit should remain as product sale price with a discount of 100 and line.amount of 0.

### Cause and Fix:
In `_reset_loyalty`, price_unit is set to zero.
The method `compute_amount` depends on `price_unit`, which means setting `price_unit` will make amount to be recomputed: https://github.com/odoo/odoo/blob/4fa9f9b849016f312efcb73f9a76b223e429aec0/addons/sale/models/sale_order_line.py#L843-L844 https://github.com/odoo/odoo/blob/4fa9f9b849016f312efcb73f9a76b223e429aec0/addons/sale/models/sale_order_line.py#L848 https://github.com/odoo/odoo/blob/4fa9f9b849016f312efcb73f9a76b223e429aec0/addons/account/models/account_tax.py#L1613

Which leads to `_compute_price_unit`. In this compute, the unit price will not be recomputed if `technical_price_unit` and `price_unit` differ.
https://github.com/odoo/odoo/blob/4fa9f9b849016f312efcb73f9a76b223e429aec0/addons/sale/models/sale_order_line.py#L606-L611 https://github.com/odoo/odoo/blob/4fa9f9b849016f312efcb73f9a76b223e429aec0/addons/sale/models/sale_order_line.py#L588-L595

So IMO if we want to recompute `price_unit`, we need to also reset `technical_price_unit` in `_reset_loyalty`.

opw-5467623

### DIFF
--- a/addons/sale_loyalty/tests/test_loyalty.py
+++ b/addons/sale_loyalty/tests/test_loyalty.py
@@ -1278,3 +1278,25 @@ class TestLoyalty(TestSaleCouponCommon):
         reward_line = sale_order.order_line.filtered('reward_id')
         self.assertEqual(reward_line.discount, 100)
         self.assertEqual(reward_line.price_total, 0)
+
+    def test_reapplying_reward_keeps_reward_price_unit(self):
+        """
+        Ensure that re-applying a reward doesn't reset the existing reward line unit price to zero
+        """
+        self.immediate_promotion_program.active = True
+        sale_order = self.empty_order
+        sale_order.write({
+            'order_line': [
+                Command.create({
+                    'product_id': self.product_A.id,
+                    'product_uom_qty': 1,
+                }),
+            ],
+        })
+        sale_order._update_programs_and_rewards()
+        self._claim_reward(sale_order, self.immediate_promotion_program)
+        reward_line = sale_order.order_line.filtered('reward_id')
+        reward_line_price_unit = reward_line.price_unit
+        sale_order._update_programs_and_rewards()
+        self._claim_reward(sale_order, self.immediate_promotion_program)
+        self.assertEqual(reward_line.price_unit, reward_line_price_unit)
--- a/addons/sale_loyalty/models/sale_order_line.py
+++ b/addons/sale_loyalty/models/sale_order_line.py
@@ -69,6 +69,7 @@ class SaleOrderLine(models.Model):
         vals = {
             'points_cost': 0,
             'price_unit': 0,
+            'technical_price_unit': 0,
         }
         if complete:
             vals.update({


addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,45 @@ def test_free_product_sol_is_zero_price(self):
...
+            'reward_ids': [
+                Command.create({
+                    'reward_type': 'product',
+                    'reward_product_id': self.product_B.id,
+                    'reward_product_qty': 1,
+                    'required_points': 1,
+                }),
+            ],
+        })
+        self.product_B.write({'lst_price': 5.0})
```
barracudapps: Seems useless as already defined here:
https://github.com/odoo-dev/odoo/blob/6f4043a9adf16ac356dcd37f0105c73e21a7e60c/addons/sale_loyalty/tests/common.py#L83-L88

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,45 @@ def test_free_product_sol_is_zero_price(self):
...
+                Command.create({
+                    'product_id': self.product_a.id,
+                    'product_uom_qty': 1,
+                }),
+            ],
+        })
+        sale_order._update_programs_and_rewards()
+        self._claim_reward(sale_order, loyalty_program)
+        reward_line = sale_order.order_line.filtered('reward_id')
+        self.assertEqual(reward_line.price_total, 0)
```
barracudapps: This seems useless IMHO as the total price is already checked in many different ways and your flow wants to ensure the unit prie is not equal to 0.

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,45 @@ def test_free_product_sol_is_zero_price(self):
...
+        self.assertEqual(reward_line.price_unit, 5.0)
+        sale_order._update_programs_and_rewards()
+        self._claim_reward(sale_order, loyalty_program)
+        self.assertEqual(reward_line.price_total, 0)
```
barracudapps: Same here

barracudapps (changes requested):
Thank you for fixing but here are some questions about the test.

barracudapps (approved):
Thank you for this !
Indeed, locking `product_B`'s price ensures the test's stability

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,31 @@ def test_free_product_sol_is_zero_price(self):
         reward_line = sale_order.order_line.filtered('reward_id')
         self.assertEqual(reward_line.discount, 100)
         self.assertEqual(reward_line.price_total, 0)
+
+    def test_sol_free_product_price_unit(self):
```
anko-odoo:
```suggestion
    def test_reapplying_reward_keeps_reward_price_unit(self):
```

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,31 @@ def test_free_product_sol_is_zero_price(self):
...
+        self.immediate_promotion_program.active = True
+        self.product_B.write({'lst_price': 5})
```
anko-odoo: I am not sure if it's needed. What could be done instead, for this test specifically is to extract price_unit after claiming the reward the first time, and then after claiming it second time, compare it to the price_unit of the second call.

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,31 @@ def test_free_product_sol_is_zero_price(self):
...
+            ],
+        })
+        sale_order._update_programs_and_rewards()
+        self._claim_reward(sale_order, self.immediate_promotion_program)
+        reward_line = sale_order.order_line.filtered('reward_id')
+        self.assertRecordValues(reward_line, [{
+            'price_unit': 5,
+            'price_total': 0,
+            'discount': 100,
+        }])
```
anko-odoo: I think checking if price_unit and price_total and discount are correct after first claim should be done in separate test. In here, we should only extract price_unit after first call.

anko-odoo (changes requested):
Hello, good job on the task, only a few nits and one improvement.

addons/sale_loyalty/tests/test_loyalty.py
```diff
@@ -1278,3 +1278,22 @@ def test_free_product_sol_is_zero_price(self):
         reward_line = sale_order.order_line.filtered('reward_id')
         self.assertEqual(reward_line.discount, 100)
         self.assertEqual(reward_line.price_total, 0)
+
+    def test_reapplying_reward_keeps_reward_price_unit(self):
+        self.immediate_promotion_program.active = True
```
anko-odoo, line 1286: Add a short doc string.

anko-odoo (changes requested):
One little comment and we are good to go. Good job on the task.

anko-odoo (approved):
LGTM @robodoo r+

PR #244296, opened 2026-01-16
[FIX] mail: ensure company specific bounce alias on catchall

delcourtfl:
**Steps to reproduce:**
- Create two companies
- Create aliases for each company
- Receive a bounced email on the alias of the second company
- Recipient of the bounce email will be from the first company

**Issue:**
`self.env.company` is used in `message_route` for catchall mails without checking if it corresponds to the received domain, making it defaults to the 'main' company instead.

**Fix:**
Try to find the company of the given mail address using the `'mail.alias.domain'` and apply it on the body rendering and `_routing_create_bounce_email` function.

similar fix in `account` module: https://github.com/odoo/odoo/commit/b7e0d8914d35af12a96593e484889e48c0613078

opw-5180433

### DIFF
--- a/addons/mail/models/mail_thread.py
+++ b/addons/mail/models/mail_thread.py
@@ -1213,8 +1213,8 @@ class MailThread(models.AbstractModel):
 
         # 2. Handle new incoming email by checking aliases and applying their settings
         # prefetch catchall aliases as they are used several times
-        catchall_aliases = self.env['mail.alias.domain'].search([]).mapped('catchall_email')
-        self = self.with_context(mail_catchall_aliases=catchall_aliases)
+        all_aliases = self.env['mail.alias.domain'].search([])
+        self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
@@ -1223,14 +1223,23 @@ class MailThread(models.AbstractModel):
             if self._detect_write_to_catchall(message_dict):
                 _logger.info('Routing mail from %s to %s with Message-Id %s: direct write to catchall, bounce',
                              email_from, message_dict['to'], message_id)
+
+                # TODO master: merge the logic with _detect_write_to_catchall.
+                recipient_company = self.env.company
+                email_to_list = [email_normalize(e) or e for e in email_split(message_dict['to'])]
+                for alias in all_aliases:
+                    if alias.catchall_email in email_to_list and alias.company_ids and recipient_company not in alias.company_ids:
+                        recipient_company = alias.company_ids[:1]
+
                 body = self.env['ir.qweb']._render('mail.mail_bounce_catchall', {
                     'message': message,
+                    'res_company': recipient_company,
                 })
-                self._routing_create_bounce_email(
+                self.with_company(recipient_company)._routing_create_bounce_email(
                     email_from, body, message,
                     # add a reference with a tag, to be able to ignore response to this email
                     references=f'{message_id} {generate_tracking_message_id("loop-detection-bounce-email")}',
-                    reply_to=self.env.company.email)
+                    reply_to=recipient_company.email)
                 return []
 
             dest_aliases = self.env['mail.alias'].search([
--- a/addons/test_mail/tests/test_mail_gateway.py
+++ b/addons/test_mail/tests/test_mail_gateway.py
@@ -1049,6 +1049,22 @@ class TestMailgateway(MailGatewayCommon):
         self.assertNotSentEmail()
 
     @mute_logger('odoo.addons.mail.models.mail_thread', 'odoo.models')
+    def test_message_route_bounce_multi_company_alias_domain(self):
+        """Incoming emails: company-specific bounce alias from catchall"""
+        with self.mock_mail_gateway():
+            record = self.format_and_process(
+                MAIL_TEMPLATE, self.partner_1.email_formatted,
+                f'{self.alias_catchall_c2}@{self.alias_domain_c2_name}',
+                subject='Test multi-company routing alias',
+            )
+        self.assertFalse(record)
+        self.assertSentEmail(
+            f'"MAILER-DAEMON" <{self.alias_bounce_c2}@{self.alias_domain_c2_name}>',
+            ['whatever-2a840@postmaster.twitter.com'],
+            subject='Re: Test multi-company routing alias'
+        )
+
+    @mute_logger('odoo.addons.mail.models.mail_thread', 'odoo.models')
     def test_message_route_bounce_other_recipients(self):
         """Incoming email: bounce processing: bounce should be computed even if not first recipient """
         with self.mock_mail_gateway():


addons/mail/models/mail_thread.py
```diff
@@ -1212,24 +1212,30 @@ def _filter_excluded_local_part(email):
...
+        all_aliases = self.env['mail.alias.domain'].search([])
+        self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
 
+            current_company = self.env.company
+            for alias in all_aliases:
+                if alias.catchall_email in message['To'] and alias.company_ids[0]:
+                    current_company = alias.company_ids[0]
```
Ephoniz: 👋🏽 

A small note here, If we have multiple companies with the same alias, we wouldn't be using then the same company here all the time?

I mean, if the current company is using the same alias as an earlier created company, we would lose the context fo the current company right? 🤔

delcourtfl: Yes, but right now the company context might not even be from one of the company using the alias.
And I am not sure on the expected behavior if we cover this case, but it might be better to send the mail to each active company which uses it.

Ephoniz: I'm not sure about this, I'll let the team reviewer decide :) I will still fix the styling tho, even though it was not following it before, I think is good to keep it as it should after the fix?

addons/mail/models/mail_thread.py
```diff
@@ -1212,24 +1212,30 @@ def _filter_excluded_local_part(email):
...
-        self = self.with_context(mail_catchall_aliases=catchall_aliases)
+        all_aliases = self.env['mail.alias.domain'].search([])
+        catchall_self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
 
+            current_company = self.env.company
+            for alias in all_aliases:
+                if alias.catchall_email in message['To'] and alias.company_ids[0]:
```
reth-odoo: 1)
alias.company_ids[:1] (alias does not necessarily have a company)

2)
See how _detect_write_to_catchall does it. Notably "to" emails need to be normalized.

addons/mail/models/mail_thread.py
reth-odoo: Indeed I think it's preferable to keep the original company, provided it's part of the possible companies.

This avoids change in behavior for most cases. Even if it probably would not matter much.

addons/mail/models/mail_thread.py
```diff
@@ -1212,24 +1212,30 @@ def _filter_excluded_local_part(email):
...
         # prefetch catchall aliases as they are used several times
-        catchall_aliases = self.env['mail.alias.domain'].search([]).mapped('catchall_email')
-        self = self.with_context(mail_catchall_aliases=catchall_aliases)
+        all_aliases = self.env['mail.alias.domain'].search([])
+        catchall_self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
 
+            current_company = self.env.company
```
reth-odoo:
```suggestion
            recipient_company = self.env.company
```

addons/mail/models/mail_thread.py
```diff
@@ -1212,24 +1212,30 @@ def _filter_excluded_local_part(email):
 
         # 2. Handle new incoming email by checking aliases and applying their settings
         # prefetch catchall aliases as they are used several times
-        catchall_aliases = self.env['mail.alias.domain'].search([]).mapped('catchall_email')
-        self = self.with_context(mail_catchall_aliases=catchall_aliases)
+        all_aliases = self.env['mail.alias.domain'].search([])
+        catchall_self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
```
reth-odoo: Did you have to rename it because of CI?

delcourtfl: Yes, it was not happy with the `self` assignment (not sure if it's better as it could break customizations ?)

```
[PLW0642](https://docs.astral.sh/ruff/rules/self-or-cls-assignment) coll 9: Reassigned `self` variable in instance method
```

reth-odoo: yes I'd keep as is here and override ci/style instead, just to be cautious

addons/test_mail/tests/test_mail_alias.py
```diff
@@ -576,6 +577,34 @@ def test_res_company_creation_alias_domain(self):
         })
         self.assertEqual(company.alias_domain_id, self.mail_alias_domain_c2)
 
+    def test_message_route_bounce_multi_company_alias_domain(self):
```
reth-odoo: 🔤

addons/mail/models/mail_thread.py
```diff
@@ -1212,24 +1212,32 @@ def _filter_excluded_local_part(email):
...
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
 
+            recipient_company = self.env.company
+            email_to_list = [email_normalize(e) or e for e in email_split(message_dict['to'])]
+            for alias in all_aliases:
+                if alias.catchall_email in email_to_list and alias.company_ids:
+                    if recipient_company not in alias.company_ids:
```
reth-odoo: if it fits
```suggestion
                if alias.catchall_email in email_to_list and alias.company_ids and recipient_company not in alias.company_ids:
```

And I guess you can ping tde once changes are done, not much more to say :)

addons/mail/models/mail_thread.py
```diff
@@ -1213,24 +1213,31 @@ def _filter_excluded_local_part(email):
...
         # prefetch catchall aliases as they are used several times
-        catchall_aliases = self.env['mail.alias.domain'].search([]).mapped('catchall_email')
-        self = self.with_context(mail_catchall_aliases=catchall_aliases)
+        all_aliases = self.env['mail.alias.domain'].search([])
+        self = self.with_context(mail_catchall_aliases=all_aliases.mapped('catchall_email'))
         if rcpt_tos_list:
             # no route found for a matching reference (or reply), so parent is invalid
             message_dict.pop('parent_id', None)
 
+            recipient_company = self.env.company
```
tde-banana-odoo: Can be put into the if below

tde-banana-odoo: Would put a comment like "in master, to merge with _detect_write_to_catchall" (that way we don't do the same stuff twice)

addons/test_mail/tests/test_mail_alias.py
```diff
@@ -595,6 +596,34 @@ def test_assert_initial_values(self):
         )
         self.assertEqual(self.company_3.default_from_email, f'{self.alias_default_from_c3}@{self.alias_domain_c3_name}')
 
+    def test_message_route_bounce_multi_company_alias_domain(self):
```
tde-banana-odoo: This should be put in test_mail_gateway file holding mail gateway tests, there are already tests for bounce emails, probably some specific use case to insert in a test / add an new test

delcourtfl:
@tde-banana-odoo small reminder for this one

tde-banana-odoo:
@robodoo override=ci/style

Reason: better rewrite 'self' than changing all self in this method, especially fro a bugfix

tde-banana-odoo:
@robodoo r+

PR #244303, opened 2026-01-16
[IMP] survey: improve ui

pyduf:
As the restrict_user_ids field label is "Shared with", we add a placeholder "Everyone" to make it clear that if we don't select a user, it will be shared with everyone.

In the survey statistic page, we change the look of the print button and place it in a div meant to be populated by other addons.

Task-5424745

### DIFF
--- a/addons/survey/views/survey_templates_statistics.xml
+++ b/addons/survey/views/survey_templates_statistics.xml
@@ -17,14 +17,16 @@
     <template id="survey_page_statistics_header" name="Survey: result statistics header">
         <div class="o_survey_statistics_header mt32">
             <h2 t-field="survey.title"/>
-            <div class="d-flex align-items-start">
+            <div class="d-flex align-items-center">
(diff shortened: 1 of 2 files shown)


addons/survey/views/survey_templates_statistics.xml
```diff
@@ -17,14 +17,16 @@
...
+            <div class="d-flex align-items-center">
                 <div t-if="question_and_page_data" class="o_survey_results_topbar d-print-none"><t t-call="survey.survey_results_filters"/></div>
                 <h3 t-else="">
                     Sorry, no one answered this survey yet.
                 </h3>
-                <button class="btn btn-primary d-print-none o_survey_results_print mt-1 ms-auto" aria-label="Print" title="Print">
-                    <i class="fa fa-print"/>
-                </button>
+                <div class="ms-auto d-flex gap-1 align-items-center">
+                    <button class="btn btn-light d-print-none o_survey_results_print" aria-label="Print" title="Print">
```
julien-banken: We can maybe add `type="submit"` on the button

julien-banken (approved):

PR #244306, opened 2026-01-16
[FIX] im_livechat: smaller size of looking for help timers

alexkuhn:
Timers on "looking for help" live chat conversations on discuss sidebar were slightly too big. This commit reduces the size to make the overal text content of a discuss item more balanced.

Before / After
Screenshot 2026-01-16 at 16 58 50 Screenshot 2026-01-16 at 16 58 57


phenix-factory (approved):
@robodoo r+

PR #222169, opened 2025-08-07
[REF] stock_account: New inventory valuation

amoyaux:
Functional:
- Automated handling of periodic valuation
- Closing report with managment of good delivered not invoiced and bill to receive
- Realtime valuation only at invoice/bill (no more at delivery/receipt)
- Removal of input/output accounts
- Automatic periodic/realtime closing
- Error management at closing (back dating, rounding issue,...)
- Back dating
- Inventory value is justify from last documents rather than the full history (fifo)

Technical:
- Removal of `stock.valuation.layer`
- value is stored on the `stock.move`
- value is compute base on accounting documents
- value priority order is: manual val > bill/inv > po/so > std price
- value could be use with a context at date to get the value at date
- `product.value` to log manual update from the user without real justification
- It doesn't try to replay and correct the past. Instead it focus on the current situation and match it

The new valuation aims to reduce the complexity of realtime accounting and improve the simple periodic valuation.

TODO:
- Correct all the tests skipped
- Ensure l10n works correctly
- Automatic valution when in/out from valued location (in realtime only)

PR #244307, opened 2026-01-16
[REF] *: remove `uom.uom.rounding` field

mohammadhossam:
This commit removes the obsolete field `rounding` on `uom.uom` model.
This field has been obsolete since 18.1 (look at 947d18a6). This caused
some confusion for developers, which led to setting the field value in
some occurrences, while it has no real effect. This commit replaces all
the usages of the field with its real value, which is the 'Product Unit'
decimal accuracy record.

Task-5506647


addons/sale_stock/models/sale_order.py
```diff
@@ -175,11 +175,12 @@ def write(self, vals):
         if values.get('order_line') and self.state == 'sale':
             for order in self:
                 to_log = {}
+                digits = self.env['decimal.precision'].precision_get('Product Unit')
                 order.order_line.fetch(['product_uom_id', 'product_uom_qty', 'display_type', 'is_downpayment'])
                 for order_line in order.order_line:
                     if order_line.display_type or order_line.is_downpayment:
                         continue
-                    if float_compare(order_line.product_uom_qty, pre_order_line_qty.get(order_line, 0.0), precision_rounding=order_line.product_uom_id.rounding) < 0:
+                    if float_compare(order_line.product_uom_qty, pre_order_line_qty.get(order_line, 0.0), precision_digits=digits) < 0:
```
Feyensv:
```suggestion
                    if order_line.product_uom_id.compare(order_line.product_uom_qty, pre_order_line_qty.get(order_line, 0.0)) < 0:
```

there are float utils equivalent on both `res.currency` and `uom.uom` models ;)

mohammadhossam: @Feyensv Thanks for you suggestion. I have updated the commit to use the float utils where possible.

addons/mrp/report/mrp_report_mo_overview.py
```diff
@@ -264,8 +263,9 @@ def _format_state(self, record, components=False):
             return _("%(producible_qty)s Ready", producible_qty=producible_qty)
         return _("Ready")
 
-    def _get_uom_precision(self, uom_rounding):
-        return max(0, int(-(log10(uom_rounding))))
+    def _get_uom_precision(self):
+        digits = self.env['decimal.precision'].precision_get('Product Unit')
+        return digits
```
clesgow:
```suggestion
        return self.env['decimal.precision'].precision_get('Product Unit')
```
Nitpick

addons/mrp/report/mrp_report_mo_overview.py
```diff
@@ -957,9 +957,9 @@ def _add_extra_in_forecast(self, forecast_lines, extras, product_rounding):
                 new_extra_line['cost'] = extra['cost'] * ratio
                 lines_with_extras.append(new_extra_line)
                 extra['quantity'] -= forecast_line['uom_id']._compute_quantity(taken_from_extra, extra['uom'])
-                if float_compare(extra['quantity'], 0, precision_rounding=product_rounding) <= 0:
+                if extra['uom'].compare(extra['quantity'], 0) <= 0:
                     index_to_remove.append(index)
-                if float_is_zero(line_qty, precision_rounding=product_rounding):
+                if extra['uom'].is_zero(line_qty):
```
clesgow:
```suggestion
                if forecast_line['uom_id'].is_zero(line_qty):
```
Doesn't change anything rounding-wise (as it's the same precision anyway), but for semantics, `line_qty` is in the `forecast_line['uom_id']`.

addons/mrp/tests/test_bom.py
```diff
@@ -2919,7 +2918,6 @@ def test_bom_kit_rounding(self):
         """
         integer_unit = self.env['uom.uom'].create({
             'name': 'unit_int',
-            'rounding': 1.0,
```
clesgow, line 2919: If we want to keep the test relevant, then we need to add this instead:
```py
self.env['decimal.precision'].search([('name', '=', 'Product Unit')]).digits = 0
```
Otherwise it doesn't check anything with 2 digits

addons/purchase_mrp/models/stock_move.py
```diff
@@ -35,6 +35,8 @@ def _get_valuation_price_and_qty(self, related_aml, to_curr):
             }
             valuation_total_qty = self._compute_kit_quantities(related_aml.product_id, order_qty, kit_bom, filters)
             valuation_total_qty = kit_bom.uom_id._compute_quantity(valuation_total_qty, related_aml.product_id.uom_id)
-            if related_aml.product_uom_id.rounding or related_aml.product_id.uom_id.is_zero(valuation_total_qty):
+            digits = self.env['decimal.precision'].precision_get('Product Unit')
+            rounding = 10 ** -digits
+            if rounding or related_aml.product_id.uom_id.is_zero(valuation_total_qty):
```
clesgow: In which case would you have a non-truthy `rounding`? 😅 
Note the previous code was wrong too.

In any case, this is dead code since the new valuation in `19.0` [here](https://github.com/odoo/odoo/pull/222169/changes#diff-d601e0945d548cd4d323375bcb53addfbb594bd198424a9b6aac0651064ecaccL294-L314), so we can even remove this method.

addons/sale_mrp/models/sale_order_line.py
```diff
@@ -57,8 +58,8 @@ def _prepare_qty_delivered(self):
                                or (m.location_dest_id.usage != 'customer'
                                and m.state == 'done'
                                and float_compare(m.quantity,
-                                                 sum(sub_m.uom_id._compute_quantity(sub_m.quantity, m.uom_id) for sub_m in m.returned_move_ids if sub_m.state == 'done'),
-                                                 precision_rounding=m.uom_id.rounding) > 0)
+                                                 sum(sub_m.uom_id._compute_quantity(sub_m.quantity, m.product_uom) for sub_m in m.returned_move_ids if sub_m.state == 'done'),
```
clesgow: Why do we use another uom here? 🤔 

Also, can't you do a `m.uom_id.float_compare(...)` rather than having to fetch the digits?

addons/stock/models/stock_move.py
```diff
@@ -1881,8 +1881,7 @@ def _get_available_move_lines(self, assigned_moves_ids, partially_available_move
         grouped_move_lines_out = self._get_available_move_lines_out(assigned_moves_ids, partially_available_moves_ids)
         available_move_lines = {key: grouped_move_lines_in[key] - grouped_move_lines_out.get(key, 0) for key in grouped_move_lines_in}
         # pop key if the quantity available amount to 0
-        rounding = self.product_id.uom_id.rounding
-        return dict((k, v) for k, v in available_move_lines.items() if float_compare(v, 0, precision_rounding=rounding) > 0)
+        return {k: v for k, v in available_move_lines.items() if self.uom_id.compare(v, 0) > 0}
```
clesgow:
```suggestion
        return {k: v for k, v in available_move_lines.items() if self.product_id.uom_id.compare(v, 0) > 0}
```
Doesn't change the rounding, but makes more sense for semantics, as the quantities are in the uom of the product, not the uom of the move line.

addons/stock/models/stock_move.py
```diff
@@ -1967,15 +1963,15 @@ def _action_assign(self, force_qty=False):
                         continue
                     # If we don't need any quantity, consider the move assigned.
                     need = missing_reserved_quantity
-                    if float_is_zero(need, precision_rounding=rounding):
+                    if move.uom_id.is_zero(need):
```
clesgow:
```suggestion
                    if move.product_id.uom_id.is_zero(need):
```
Since we're using quantity in the product's uom, this makes it clearer. Same thing for the four others below.

addons/stock/models/stock_move.py
```diff
@@ -2654,7 +2649,7 @@ def check_available_qty(self):
         else:
             available_qty = sum(product.with_context(lot_id=lot.id).qty_available for lot in self.lot_ids)
         scrap_qty = self.uom_id._compute_quantity(self.quantity, self.product_id.uom_id)
-        return float_compare(available_qty, scrap_qty, precision_rounding=self.product_id.uom_id.rounding) >= 0
+        return self.uom_id.compare(available_qty, scrap_qty) >= 0
```
clesgow:
```suggestion
        return self.product_id.uom_id.compare(available_qty, scrap_qty) >= 0
```

addons/stock/models/stock_package.py
```diff
@@ -494,7 +494,7 @@ def _apply_dest_to_package(self, processed_package_ids=None):
             if len(new_location) > 1:
                 raise UserError(self.env._("Packages %(duplicate_names)s are moved to different locations while being in the same container %(container_name)s.",
                                             duplicate_names=packages.mapped('name'), container_name=container_package.name))
-            contained_quants = container_package.contained_quant_ids.filtered(lambda q: not float_is_zero(q.quantity, precision_rounding=q.uom_id.rounding))
+            contained_quants = container_package.contained_quant_ids.filtered(lambda q: not q.uom_id.compare(q.quantity, 0))
```
clesgow:
```suggestion
            contained_quants = container_package.contained_quant_ids.filtered(lambda q: not q.uom_id.is_zero(q.quantity))
```

addons/stock/report/report_package_barcode.xml
```diff
@@ -3,9 +3,10 @@
 <data>
 <template id="report_package_barcode">
     <t t-set="uom_unit_id" t-value="env.ref('uom.product_uom_unit').id"/>
+    <t t-set="rounding_digits" t-value="env['decimal.precision'].precision_get('Product Unit')"/>
     <t t-call="web.basic_layout"
         uom_unit_id="uom_unit_id"
-        gs1_uom_patterns="{rule.associated_uom_id.id: rule.pattern[1:4] + str(len(str('{:.10f}'.format(rule.associated_uom_id.rounding).rstrip('0')).split('.')[1])) for rule in env['barcode.rule'].search([('associated_uom_id', '!=', False), ('associated_uom_id.id', '!=', uom_unit_id), ('is_gs1_nomenclature', '=', True)])}">
+        gs1_uom_patterns="{rule.associated_uom_id.id: rule.pattern[1:4] + rounding_digits for rule in env['barcode.rule'].search([('associated_uom_id', '!=', False), ('associated_uom_id.id', '!=', uom_unit_id), ('is_gs1_nomenclature', '=', True)])}">
```
clesgow:
```suggestion
        gs1_uom_patterns="{rule.associated_uom_id.id: rule.pattern[1:4] + str(rounding_digits) for rule in env['barcode.rule'].search([('associated_uom_id', '!=', False), ('associated_uom_id.id', '!=', uom_unit_id), ('is_gs1_nomenclature', '=', True)])}">
```

addons/stock/report/stock_forecasted.py
```diff
@@ -303,10 +302,10 @@ def _get_out_move_taken_from_stock_data(out, currents, reserved_data, wh_stock_l
                 'taken_from_stock': taken_from_stock_out,
             }
 
-        def _reconcile_out_with_ins(lines, out, ins, demand, product_rounding, only_matching_move_dest=True, read=True):
+        def _reconcile_out_with_ins(lines, out, ins, demand, only_matching_move_dest=True, read=True):
```
clesgow: I'd use `product_uom` here and use that one rather than rely on the move's uom.

clesgow (commented):
A lot of nitpicks, but most of them stems from the same thing: Keep using the right `uom.uom` when using the float utils, even though it doesn't change anything for the `rounding`.

*If* things change in the future, we'll be glad to not have to hunt for the right uom again.

addons/repair/models/repair.py
```diff
@@ -317,7 +317,7 @@ def _compute_availability_boolean(self):
             elif repair.parts_availability_state == 'late':
                 repair.is_parts_late = True
 
-    @api.depends('move_ids.quantity', 'move_ids.product_uom_qty', 'move_ids.uom_id.rounding')
```
clesgow, line 348: Given the definition of the compute, I wonder if we should rather keep `move_ids.uom_id` as a `depends`. 🤔

clesgow (commented):
Just found a little something that I missed last time, but aside from that LGTM 👍

addons/repair/models/repair.py
mohammadhossam: Yes, indeed. Got it back :+1:

PR #244309, opened 2026-01-16
[FIX] account_no_followup: Set no_followup on general entries

seifgneedy:
General(MISC) entries should be no_followup by default.

task-5489772

---


hupo-odoo (approved):
@robodoo r+

PR #244314, opened 2026-01-16
[FIX] account_edi_ubl_cii: traceback at xml generation

guva-odoo:
Steps:

- Belgian localisation
- Activate peppol
- Have two fixed sales taxes (T1 3.5 and T2 4.5)
- Have 4 product:
	- P1: Any sale price, taxes 21% and T1
	- P2: Any sale price, taxes 21% and T2
	- P3: sale price 0, taxes 0% and T1
- Create an invoice, with following invoice lines:
	- P1, quantity 2
	- P2, quantity 2
	- P3, quantity -4
- Confirm and send it to peppol
-> Traceback (ZeroDivisionError)

The reason is that we try to extract emptying taxes like "Vidanges" and
aggregate them into new base lines, but we treat all these taxes as they
are the same but they are not always the same. Therefore we aggregate
both price unit and quantity and we try to divide the aggregated price
by the aggregated quantity. In our case we end up with a price unit of 2
(9 + 7 - 14) and a quantity of 0 (2 + 2 + -4) which leads to a zero
division error.

The fix adds a grouping function in order to group the extra lines by
taxes before aggregating them.

opw-5384928

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
@@ -1175,44 +1175,9 @@ class AccountEdiXmlUBL20(models.AbstractModel):
                     new_taxes_data.append(tax_data)
 
     def _turn_emptying_taxes_as_new_base_lines(self, base_lines, company, vals):
-        """ Extract emptying taxes such as "Vidanges" on bottles from the current base lines and turn them into
-        additional base lines.
-
-        :param base_lines:  The original 'base_lines' of the document.
-        :param company:     The company owning the 'base_lines'.
-        :param vals:        Some custom data.
-        """
-        AccountTax = self.env['account.tax']
         if not vals['fixed_taxes_as_allowance_charges']:
             return base_lines
-
-        def exclude_function(base_line, tax_data):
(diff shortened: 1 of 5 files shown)


addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -1172,44 +1172,7 @@ def is_recycling_contribution(tax_data):
...
-        """ Extract emptying taxes such as "Vidanges" on bottles from the current base lines and turn them into
-        additional base lines.
-
-        :param base_lines:  The original 'base_lines' of the document.
-        :param company:     The company owning the 'base_lines'.
-        :param vals:        Some custom data.
-        """
-        AccountTax = self.env['account.tax']
-        if not vals['fixed_taxes_as_allowance_charges']:
-            return base_lines
```
guva-odoo, line 1187: Shouldn't we keep that part? @smetl

smetl: @guva-odoo Indeed^^

smetl (approved):
@robodoo r+

PR #244318, opened 2026-01-16
[IMP] website_sale_mrp: Subtract qty from kits

axelmanzanilla:
When a product with a kit was added to the website cart, the quantities of the
kit components were not deducted from stock calculations. As a result, customers
could add the same product again without the kit, even though the available stock
should have been reduced.

Current behavior before PR:
Adding a product with a kit to the cart does not subtract the quantities of the
kit components. This allows customers to add the same product without the kit
regardless of actual stock availability.

When a product with a kit is added to the cart, the quantities of the kit
components are correctly deducted from stock calculations, preventing customers
from adding the same product without the kit when stock limits are reached.

---

### DIFF
--- a/addons/website_sale_mrp/models/sale_order.py
+++ b/addons/website_sale_mrp/models/sale_order.py
@@ -58,3 +58,7 @@ class SaleOrder(models.Model):
                 max_free_kit_qty = min(max_free_kit_qty, (component.free_qty - unavailable_component_qty) // qty_per_kit[component])
             unavailable_qty += free_qty - max_free_kit_qty
         return unavailable_qty
+
+    def _get_cart_and_free_qty(self, product, line=None):
+        cart_qty, free_qty = super()._get_cart_and_free_qty(product, line)
+        return cart_qty, free_qty - self._get_unavailable_quantity_from_kits(product)


addons/website_sale_mrp/models/sale_order.py
```diff
@@ -58,3 +58,7 @@ def _get_unavailable_quantity_from_kits(self, product):
                 max_free_kit_qty = min(max_free_kit_qty, (component.free_qty - unavailable_component_qty) // qty_per_kit[component])
             unavailable_qty += free_qty - max_free_kit_qty
         return unavailable_qty
+
+    def _get_cart_and_free_qty(self, product, line=None):
+        cart_qty, free_qty = super()._get_cart_and_free_qty(product, line)
+        return cart_qty, free_qty - self._get_unavailable_quantity_from_kits(product)
```
Feyensv, line 64:
```suggestion
    def _get_cart_and_free_qty(self, product, line=None):
        cart_qty, free_qty = super()._get_cart_and_free_qty(product, line)
        return cart_qty + self._get_unavailable_quantity_from_kits(product), free_qty 
```
Let's follow the same approach as the way done through the `variant` controller (the kit quantity from the cart is added to the cart quantity, not removed from the free quantity).
It's better from a functional POV, as the customer will know that some are in their cart, instead of being told that those in the cart are not available at all (and maybe see more quantities on another browser/computer/...).

Feyensv (commented):
Could you add a test to cover this fix please ? :pray:
It'll help ensuring that the issue doesn't re-appear in the future.

PR #244322, opened 2026-01-16
[FIX] website_sale: fixing test

sawer-odoo:
version: 17.0+e

Why ?
---------------

test_payment_provider_visibility_with_portal didn't pass on runbot because the sale order used doesn't reqire payment.

The fix
---------------

Force the sale order to require a payment.

runbot-237786

---


Feyensv (approved):
robodoo delegate+ up to saas-18.2

Thanks for the fix :pray:

PR #244324, opened 2026-01-16
[FIX] html_builder: Prevent dropping image in invalid locations

sben-odoo:
Steps to reproduce the issue:
=================================

1. Go to Website > Products and edit a product.
2. In the product description, add a header using the editor.
3. You can now drop an image in that place, so drop it.
4. Click on the image and try to add an anchor for it.

-> Traceback

Cause:
======

The reason an image can be dropped in that location is because the editor
added an adjacent element, such as a `<p>` or `<h1>`, which is considered
a valid drop zone by the drop zone plugin. However, when trying to
create a link, the parent element checks if it is valid to have an
anchor, but then when adding an anchor the element text is undefined,
causing a traceback.

Why can an image be dropped in that location after adding the editor?
The bug was due to passing `excludeNearParent` as a property name instead
of `excludeParent` to the `getSelectorSiblings` function, which always
evaluated to false. So it allowed dropping images in invalid locations.

Solution:
=========
Pass the correct property name.

Impact:
=======
This fix enforces the intended dropzone behavior where content snippets
(like images, buttons with `.o_snippet_drop_in_only`) cannot create
dropzones next to elements that are direct children of `.oe_structure`
containers.

Test Changes Required:
=======================
The button tests were failing after the fix because they were testing
buttons as direct children of
`<div id="wrap" class="oe_structure oe_empty">`. This was only working
before due to the bug - `excludeNearParent` was being ignored.

The fix wraps the test buttons in a `<div>` container,
making them children of a child of `.oe_structure` rather than
direct children.

opw-5489444

---

### DIFF
--- a/addons/html_builder/static/src/core/disable_snippets_plugin.js
+++ b/addons/html_builder/static/src/core/disable_snippets_plugin.js
@@ -143,7 +143,7 @@ export class DisableSnippetsPlugin extends Plugin {
                 dropAreaEls.push(
                     ...this.dependencies.dropzone.getSelectorSiblings(editableAreaEls, rootEl, {
                         selector: dropNear,
-                        excludeNearParent,
+                        excludeParent: excludeNearParent,
                     })
                 );
             }
--- a/addons/html_builder/static/tests/drop_zone.test.js
+++ b/addons/html_builder/static/tests/drop_zone.test.js
@@ -37,3 +37,29 @@ test("drop beside dropzone inserts the snippet", async () => {
     </div>
 </section>`);
 });
+
+test("snippets cannot be dropped next to elements inside excluded parent", async () => {
+    const snippetContent = [
+        `<div name="Image" data-oe-thumbnail="image.svg" data-snippet="s_image">
+            <img src="/web/image/test.png" data-snippet="s_image" alt="Test Image"/>
+        </div>`,
+    ];
+    const dropzoneSelectors = [
+        {
+            selector: "img",
+            dropNear: "p, h1",
+            excludeNearParent: ".second-div",
+        },
+    ];
+    await setupHTMLBuilder(
+        `<div class="first-div"><h1>Title</h1><p>Paragraph</p></div>
+        <div class="second-div"><h1>Title</h1><p>Paragraph</p></div>`,
+        { snippetContent, dropzoneSelectors }
+    );
+
+    await contains(".o-snippets-menu .o_snippet_thumbnail[data-snippet='s_image']").drag();
+    // Should have 3 dropzones in first-div (not excluded)
+    expect(":iframe .first-div .oe_drop_zone").toHaveCount(3);
+    // Should have no dropzones in second-div (excluded by excludeNearParent)
+    expect(":iframe .second-div .oe_drop_zone").toHaveCount(0);
+});
--- a/addons/website/static/tests/builder/website_builder/button_option.test.js
+++ b/addons/website/static/tests/builder/website_builder/button_option.test.js
@@ -39,11 +39,11 @@ test("Drag & drop a 'Button' snippet in a <div> should put it inside a <p>", asy
 
 test("Drag & drop a 'Button' snippet should align the button style with the button before it", async () => {
     const { getEditableContent } = await setupWebsiteBuilder(
-        `<a href="http://test.com" class="btn btn-fill-secondary" style="line-height: 50px;">ButtonStyled</a>`
+        `<div><a href="http://test.com" class="btn btn-fill-secondary" style="line-height: 50px;">ButtonStyled</a></div>`
     );
     const contentEl = getEditableContent();
     expect(contentEl).toHaveInnerHTML(
-        `<a href="http://test.com" class="btn btn-fill-secondary" style="line-height: 50px;">ButtonStyled</a>`
+        `<div class="o-paragraph"><a href="http://test.com" class="btn btn-fill-secondary" style="line-height: 50px;">ButtonStyled</a></div>`
     );
     expect(".o-website-builder_sidebar .fa-undo").not.toBeEnabled();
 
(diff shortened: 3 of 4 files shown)


nle-odoo (approved):
fix makes sense to me

there should probably an hoot test if this was not tested before. I'm not sure if there is already a test that test this, in worst case we could call directly `getDropAreas` with the DOM that reproduce this error, and check that if we have the correct drop zones (or just drop the widget on the area and check that it is not dropped after the change, and that before the change there is an traceback)

addons/html_builder/static/src/core/disable_snippets_plugin.js
```diff
@@ -143,7 +143,7 @@ export class DisableSnippetsPlugin extends Plugin {
                 dropAreaEls.push(
```
sobo-odoo, line 143: The issue happens in 18.4 too, so retarget your PR :slightly_smiling_face: 
Try to always find the oldest version when doing a fix :wink:

sobo-odoo, line 143: Commit message:
- Title
  - transform the "p" of "Prevent" to lowercase
  - you also modified the website module, so add it in the title too :slightly_smiling_face: 
  - rename  it to something like "prevent dropping inner content in invalid locations" (as it does not concern only images) :slightly_smiling_face: 

- Some line exceeds 72 chars
- `add a header using the editor.` => I think you meant "heading", I would also add that we have to do it with the "/" command :slightly_smiling_face:

addons/html_builder/static/tests/drop_zone.test.js
```diff
@@ -35,3 +35,46 @@ test("drop beside dropzone inserts the snippet", async () => {
     <div class="test_a"></div>
     </section>`);
 });
+
+test("content snippets cannot be dropped next to elements inside oe_structure", async () => {
```
sobo-odoo: Since we are talking about contents, move the test in `snippet_content.test.js` file :slightly_smiling_face: (but see last comment first)

sobo-odoo: Given that you don't use the real selectors for that, rename the test as `Content snippets cannot be dropped next to elements inside excluded parent`, to be more generic :+1: 

Note that you could have used any selector for your test if you want, it is not necessary to use `oe_structure` specifically, but it's fine like this :slightly_smiling_face:

addons/html_builder/static/tests/drop_zone.test.js
```diff
@@ -35,3 +35,46 @@ test("drop beside dropzone inserts the snippet", async () => {
...
+    const editableAreaEls = setupEditorPlugin.getEditableAreas();
+    const rootEl = dropzonePlugin.getDropRootElement();
+    const dropAreasBySelector = disableSnippetsPlugin.getDropAreas(editableAreaEls, rootEl);
+
+    // Find the drop areas for our image selector
+    const imageDropAreas = dropAreasBySelector.find((item) => item.selector === "img");
+
+    expect(imageDropAreas?.dropAreaEls || []).toHaveLength(0, {
+        message: "Content snippets should not have drop zones next to h1/p inside .oe_structure",
+    });
```
sobo-odoo: I don't like the fact that we are calling the plugins and such, you can simplify the test by just testing if there are dropzones in the DOM :slightly_smiling_face: 

Like drag the image snippet, then check that there is no `:iframe .oe_drop_zone` elements
Then to go further and make the test more relevant, wrap the content in a div, and then drag again => there should be dropzones because they are not direct child of `oe_structure` anymore :slightly_smiling_face: 

Aah I see that the snippet is disabled (because it works), so I would suggest to make the test another way:
- Define a builder with two divs with a `<p>` and `<h1>` in them and with classes:
  ```html
  <div class="first-div"><h1>Title</h1><p>Paragraph</p></div>
  <div class="second-div"><h1>Title</h1><p>Paragraph</p></div>
  ```
- The dropzoneSelector will therefore have `excludeNearParent: ".second-div"` (and only keep `p` and `h1`, the rest is useless for the test)
- Then you drag the image and check that there are 3 dropzones in `first-div`, and 0 in `second-div` :slightly_smiling_face: 

And then in that case, you can keep the test in `drop_zone.test.js` as it is generic enough :slightly_smiling_face:

sobo-odoo (changes requested):
The fix is okay with me :slightly_smiling_face: I have comments on the test and the commit message :+1:

addons/html_builder/static/tests/drop_zone.test.js
Syozik: Hey @sben-odoo apparently we worked on the same bug in parallel 😅 I will close my PR https://github.com/odoo/odoo/pull/248453, but you can copy the test from there, if you wish.

addons/website/static/tests/builder/website_builder/button_option.test.js
```diff
@@ -98,7 +98,7 @@ test("Drag & drop a 'Button' snippet over a dropzone should preview it correctly
     await drop(getDragHelper());
     await waitForEndOfOperation();
     expect(contentEl).toHaveInnerHTML(
-        `<a href="http://test.com" class="btn btn-fill-secondary"> ButtonStyled </a>
+        `<div><a href="http://test.com" class="btn btn-fill-secondary"> ButtonStyled </a>
```
sobo-odoo, line 101: The closing `</div>` is missing :slightly_smiling_face: (That's why the build failed)

addons/html_builder/static/tests/drop_zone.test.js
```diff
@@ -37,3 +37,38 @@ test("drop beside dropzone inserts the snippet", async () => {
...
+    ).drag();
+    await moveTo(":iframe .first-div");
+
+    expect(":iframe .first-div .oe_drop_zone").toHaveCount(3, {
+        message: "Should have 3 dropzones in first-div (not excluded)",
+    });
+
+    expect(":iframe .second-div .oe_drop_zone").toHaveCount(0, {
+        message: "Should have no dropzones in second-div (excluded by excludeNearParent)",
+    });
```
sobo-odoo: I can nitpick since I have other comments :stuck_out_tongue: : remove the empty lines, it takes space for nothing :wink: Maybe you can keep one between the setup and the drag :+1:

addons/website/static/tests/builder/website_builder/button_option.test.js
```diff
@@ -61,20 +61,20 @@ test("Drag & drop a 'Button' snippet should align the button style with the butt
     await drop(getDragHelper());
     await waitForEndOfOperation();
     expect(contentEl).toHaveInnerHTML(
-        `<a href="http://test.com" class="btn btn-fill-secondary mb-2" style="line-height: 50px;"> ButtonStyled </a> <a class="btn mb-2 btn-fill-secondary" href="#"> Button </a>`
+        `<div><a href="http://test.com" class="btn btn-fill-secondary mb-2" style="line-height: 50px;"> ButtonStyled </a> <a class="btn mb-2 btn-fill-secondary" href="/contactus"> Button </a></div>`
```
sobo-odoo: Not that it really matters but why did you modify the link ? (/contactus)

addons/html_builder/static/tests/drop_zone.test.js
```diff
@@ -37,3 +37,38 @@ test("drop beside dropzone inserts the snippet", async () => {
     </div>
 </section>`);
 });
+
+test("content snippets cannot be dropped next to elements inside excluded parent", async () => {
```
sobo-odoo: I know it was my suggestion, but in fact remove `content` and simply say `snippets cannot be...` because it concerns all snippets, not only inner contents :+1:

addons/html_builder/static/tests/drop_zone.test.js
sobo-odoo: You can just drag and check the dropzones, no need to move :slightly_smiling_face: It's either that, or do the whole flow by moving it in second div and dropping it, and check that it is in first div like in SERU's PR; but we don't really need to go that far :slightly_smiling_face: 
Also nitpick: you can just put comments before the check, no need for to use `message` :+1: (it is shorter)
```js
await contains(".o-snippets-menu .o_snippet_thumbnail[data-snippet='s_image']").drag();
// Should have 3 dropzones in first-div (not excluded)
expect(":iframe .first-div .oe_drop_zone").toHaveCount(3);
// Should have no dropzones in second-div (excluded by excludeNearParent)
expect(":iframe .second-div .oe_drop_zone").toHaveCount(0);
```

sobo-odoo (changes requested):
Some details to change :slightly_smiling_face:

addons/website/static/tests/builder/website_builder/button_option.test.js
sben-odoo: It was a merge mistake :(

addons/website/static/tests/builder/website_builder/button_option.test.js
```diff
@@ -39,11 +39,11 @@ test("Drag & drop a 'Button' snippet in a <div> should put it inside a <p>", asy
```
sobo-odoo, line 39: The build is still red, apparently a `o-paragraph` class is added on the div sometimes, can you recheck which tests need it and add the class if needed :slightly_smiling_face:

sobo-odoo (changes requested):

sobo-odoo (approved):
Thanks for the fix ! :slightly_smiling_face: 

@robodoo r+

PR #244325, opened 2026-01-16
[FIX] mail,website: offset chat hub with website editor panel

alexkuhn:
Before this commit, chat windows were overlapping website editor panel. This happens because chat windows had a hard-code starting position of 15px starting from the right of the screen, which is applied in all contexts.

This commit turns `BUBBLE_START` into computed field, so that this is patched when in website editor to offset it next to website editor panel, as to not overlap it.

Before / After
Screenshot 2026-01-16 at 18 11 50
Screenshot 2026-01-16 at 18 21 53

### DIFF
--- a/addons/mail/static/src/core/common/chat_hub_model.js
+++ b/addons/mail/static/src/core/common/chat_hub_model.js
@@ -3,6 +3,7 @@ import { fields, Record } from "@mail/model/export";
 
 import { Deferred, Mutex } from "@web/core/utils/concurrency";
 
+export const CHAT_HUB_DEFAULT_BUBBLE_START = 15; // same value as $o-mail-ChatHub-bubblesStart
 export const CHAT_HUB_KEY = "mail.ChatHub";
 const CHAT_HUB_COMPACT_LS = "mail.user_setting.chathub_compact";
 
@@ -15,7 +16,17 @@ const CHAT_HUB_COMPACT_LS = "mail.user_setting.chathub_compact";
 
 export class ChatHub extends Record {
     BUBBLE = 56; // same value as $o-mail-ChatHub-bubblesWidth
-    BUBBLE_START = 15; // same value as $o-mail-ChatHub-bubblesStart
+    recomputeBubbleStart = 0;
+    BUBBLE_START = fields.Attr(CHAT_HUB_DEFAULT_BUBBLE_START, {
+        /** @this {import("models").Chathub} */
+        compute() {
+            void this.recomputeBubbleStart;
+            return this.computeBubbleStart();
+        },
+    });
+    computeBubbleStart() {
+        return CHAT_HUB_DEFAULT_BUBBLE_START;
+    }
     BUBBLE_LIMIT = 7;
     BUBBLE_OUTER = 10; // same value as $o-mail-ChatHub-bubblesMargin
     WINDOW_GAP = 10; // for a single end, multiply by 2 for left and right together.
--- a/addons/mail/static/src/core/common/primary_variables.scss
+++ b/addons/mail/static/src/core/common/primary_variables.scss
@@ -3,7 +3,7 @@ $o-mail-Avatar-sizeSmall: 24px !default;
 $o-mail-ChatWindow-width: 380px !default; // same value as WINDOW
 $o-mail-ChatHub-bubblesWidth: 56px !default; // same value as BUBBLE
 $o-mail-ChatHub-bubblesMargin: 10px !default; // same value as BUBBLE_OUTER
-$o-mail-ChatHub-bubblesStart: 15px !default; // same value as BUBBLE_START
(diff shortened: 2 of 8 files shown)


addons/mail/static/src/core/common/chat_hub_model.js
```diff
@@ -15,7 +15,17 @@ const CHAT_HUB_COMPACT_LS = "mail.user_setting.chathub_compact";
 
 export class ChatHub extends Record {
     BUBBLE = 56; // same value as $o-mail-ChatHub-bubblesWidth
-    BUBBLE_START = 15; // same value as $o-mail-ChatHub-bubblesStart
+    recomputeBubbleStart = 0;
+    BUBBLE_START = fields.Attr(15, {
+        /** @this {import("models").Chathub} */
+        compute() {
+            void this.recomputeBubbleStart;
```
phenix-factory, line 23: What is the purpose of `void`? I get it's used to make it a dependency, but why use `void`?

addons/mail/static/src/core/common/primary_variables.scss
```diff
@@ -3,7 +3,7 @@ $o-mail-Avatar-sizeSmall: 24px !default;
 $o-mail-ChatWindow-width: 380px !default; // same value as WINDOW
 $o-mail-ChatHub-bubblesWidth: 56px !default; // same value as BUBBLE
 $o-mail-ChatHub-bubblesMargin: 10px !default; // same value as BUBBLE_OUTER
-$o-mail-ChatHub-bubblesStart: 15px !default; // same value as BUBBLE_START
+$o-mail-ChatHub-bubblesStart: 15px !default; // same value as BUBBLE_START @deprecated
```
phenix-factory, line 6: Would put the `@[...]` at the top to ensure visibility

addons/website/static/src/common/chat_hub_patch.js
```diff
@@ -0,0 +1,22 @@
...
+
+patch(ChatHub.prototype, {
+    setup() {
+        super.setup(...arguments);
+        this.wasWebsiteEdition = false;
+        useEffect(
+            (isWebsiteEdition) => {
+                if (isWebsiteEdition !== this.wasWebsiteEdition) {
+                    this.chatHub.recomputeBubbleOffset++;
+                }
```
phenix-factory: Is `this.env.services.website?.context.edition` a boolean? If so, why do we need to check the previous value? Don't we always want to recompute if the value changes?
The compute rechecks the value of `this.env.services.website?.context.edition` anyway, storing and checking is/was looks useless to me.

phenix-factory (changes requested):
Could we have a test that ensures the chathub position when the editor is opened and closed? Looks like a valuable test to me.

addons/mail/static/src/core/common/chat_hub_model.js
alexkuhn: `void` does nothing and just makes linter not panic thinking this LOC is useless.
This is a read on field to have it recompute, which is useful for the `useEffect` in diff.

Without this, when website editor is enabled initially, the chat hub can be mis-positioned until next rendering. This recompute is to observe the change and properly re-compute the position in UI.

addons/mail/static/src/core/common/primary_variables.scss
alexkuhn: It's removed in master anyway. I grep `@deprecated` in master and remove the affected logic, so no need to worry about format.
`@deprecate` above makes sense for function js docs as the autocomplete shows the `@deprecate`, but here this doesn't apply.

addons/website/static/src/common/chat_hub_patch.js
alexkuhn: We don't want infinite rendering => `recompute++` => rendering => `recompute++`, etc. And it's fine to explicitly force a recompute when needed, especially because this is shenanigans from lack of generalized reactivity that needs compensation with this code.

The value can be boolean or undefined, but the compute observes the value as long as the `services.website.context` has value, which isn't the case most likely at page load. That's why we need to help the compute function to recompute itself.

OWL 3 with signals should solve this problem once and for all, and we won't need to help reactive code with recompute like that.

addons/website/static/src/common/chat_hub_patch.js
```diff
@@ -0,0 +1,33 @@
...
+            const rootEl = this.root.el;
+            if (!rootEl) {
+                return;
+            }
+            if (this.isWebsiteEdition) {
+                rootEl.dataset.isWebsiteEdition = this.isWebsiteEdition;
+            } else {
+                delete rootEl.dataset.isWebsiteEdition;
+            }
+        });
```
tsm-odoo: Should be moved in patch during test.

tsm-odoo (approved):
@robodoo delegate+

CSS is not enough as the chat hub relies relies on `bubble_start` to compute how many chat windows fit in the view. TranslateX will make this computation incorrect, leading to overflows of chat windows.

PR #244371, opened 2026-01-16
[FIX] microsoft_calendar: support single-tenant apps

LouisGobert:
Single-tenant Azure applications could synchronize calendar with Outlook, but refresh token renewal fail.

Odoo was always using the default Microsoft token endpoint instead of the tenant-specific endpoint required for single-tenant apps.

Steps to reproduce:
- Create a single-tenant app in the Azure portal
- Configure Odoo Microsoft Calendar with this app
- Set `microsoft_account.auth_endpoint` and `microsoft_account.token_endpoint` system parameters with the specific endpoints using the tenant ID
- Open the Calendar app and sync with Outlook
- Wait for access token expiration
- Refresh token request fails

This commit fixes the issue by using the token endpoint stored in the microsoft_account.token_endpoint system parameter when requesting a refresh token.

### DIFF
--- a/addons/microsoft_calendar/models/res_users.py
+++ b/addons/microsoft_calendar/models/res_users.py
@@ -9,7 +9,6 @@ from datetime import timedelta
 from odoo import api, fields, models, _, Command
 from odoo.exceptions import UserError
 from odoo.loglevels import exception_to_unicode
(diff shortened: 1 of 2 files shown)


LouisGobert:
@arj-odoo Hello, test added :)

PR #244378, opened 2026-01-16
[FIX] website_event: Fix Value Conversion in Attendee Registration

CarmenMiranda:
Before the addition of identification questions like "name", "email", and "phone" in the commit [1] as event questions instead of having them static, we could add custom data, such as fields for the address, with static inputs in the form. After that addition, it's no longer possible because the registration gives us the following error when trying to convert data that isn't a M2o ID or an Integer value:
  
    invalid literal for int() with base 10

By adding the check for the field's type, we can still add custom fields with static fields in the template, as an alternative, given that there's no question type for other fields.

[1]: https://github.com/odoo/odoo/commit/6b8daa880c

---

### DIFF
--- a/addons/website_event/controllers/main.py
+++ b/addons/website_event/controllers/main.py
@@ -299,7 +299,10 @@ class WebsiteEventController(http.Controller):
                 registration_index, field_name = key_values
                 if field_name not in registration_fields:
                     continue
-                registrations.setdefault(registration_index, dict())[field_name] = int(value) or False
+                # Only cast when needed, as it might crash here for custom inputs in overrides
+                if isinstance(registration_fields[field_name], (fields.Many2one, fields.Integer)):
+                    value = int(value) or False
+                registrations.setdefault(registration_index, dict())[field_name] = value
                 continue
 
             if len(key_values) != 3:


reth-odoo (approved):
Indeed as technically `_get_website_registration_allowed_fields` still allows string fields it would make sense to handle this case for now. I suppose for most fields `convert_to_column` handles strings anyways so I see the broader usefulness.

@Odoonan 👋

addons/website_event/controllers/main.py
```diff
@@ -294,7 +294,9 @@ def _process_attendees_form(self, event, form_details):
                 registration_index, field_name = key_values
                 if field_name not in registration_fields:
                     continue
-                registrations.setdefault(registration_index, dict())[field_name] = int(value) or False
+                if isinstance(registration_fields[field_name], (fields.Many2one, fields.Integer)):
+                    value = int(value) or False
+                registrations.setdefault(registration_index, dict())[field_name] = value
```
Odoonan, line 305: Hi @CarmenMiranda ,

Thanks for this PR. I am just not sure of the use case, could you be a bit more specific? The list of allowed fields is protected by `_get_website_registration_allowed_fields`, do you have a concrete example where this fails and that is not supported? (what is the format of the input name e.g.) All string fields seem to have existing question types that could handle them?

Thanks!

CarmenMiranda: Hi @Odoonan 
We have a case where we need to include custom data for the registration, which, by inheriting the method `_get_website_registration_allowed_fields`, allows, such as fields for the address (like `street`, `zip_code`) of each of the customers/tickets, to be included in the `event.registration` (also, as inheritance of that model), with static inputs in the form

And now, having the conversion without checking what type of field it is, those values that are strings fail (in the `int(value)`)

Those cases can't be covered by the question types that exist right now, because those are just saved as records in `event.question.answer`, and we don't have a place to put the field that corresponds to

Cc. @luisg123v

Odoonan: Hi, thanks for the answer! What is the situation allowing such an update but not of this part of the code itself? custom modules? :) 🔍

CarmenMiranda: Hi, yes, we have a custom module that, before the mentioned update [1], worked fine

What allows us to include those fields is the inheritance of the method `_get_website_registration_allowed_fields`, the module `event.registration`, and the template

The idea of this change is to include cases that can be used elsewhere, and to avoid overwriting the whole method (because it is huge and not inheritable-friendly) or using patches.

[1]: https://github.com/odoo/odoo/commit/6b8daa880c

Cc. @luisg123v @fernandahf

CarmenMiranda: Hi @Odoonan @reth-odoo, Friendly reminder for this review 🤓 

Don't forget us 🙈 

Cc. @luisg123v @fernandahf

reth-odoo: We were a bit in a rush for the minor version freeze sorry ^^

Still seems ok to me, feel free to ping nan later if you don't get any news

Odoonan: Hi @CarmenMiranda ,

Indeed, sorry for the delay. I agree with the change. Small things left:

- Can we add a comment just above your change to remind us that this also used for extensibility?
`# Only cast when needed, as it might crash here for custom inputs in overrides`
- Also, commit title suggestion:
`[FIX] website_event: ease attendee form inputs extensibility `
- And in the commit message, feel free to explicitly mention that the fix is needed to allow _custom modules overrides_ to keep extensibility in mind when reading the commit message in the future :)
- Feel free to rebase as well, as the branch is a few weeks old :) 

Thanks! and sorry for the delay, it was indeed a bit of a rush.

CarmenMiranda: Hi @Odoonan, Changes are done, and it's also rebased 🤓

PR #244398, opened 2026-01-17
[IMP] account: remove constraint on journal accounts reconciliation

abmn-odoo:
Previously, a constraint prevented accounts from being non-reconcilable if they were used as default debit/credit accounts involved in journals. This behavior is too restrictive.

This commit removes the constraint.

task-5254202

---

### DIFF
--- a/addons/account/models/account_payment_method.py
+++ b/addons/account/models/account_payment_method.py
@@ -164,24 +164,11 @@ class AccountPaymentMethodLine(models.Model):
 
     @api.model
     def _auto_toggle_account_to_reconcile(self, account_id):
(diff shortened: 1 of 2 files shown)


addons/account/models/account_payment_method.py
william-andre, line 166: Can you add a comment here to state that it is deprecated?

abmn-odoo: Added :+1: 

Related PR: https://github.com/odoo/odoo/pull/245289

william-andre (approved):
@robodoo r+

PR #244406, opened 2026-01-19
[FIX] resource: attendance intervals batch

vin-odoo:
Fixes an issue with the attendance interval batch calculation; which would apply the
latest work entry type in the attendances to each and every work entry being generated.

If a work entry type was set on a weekend, that type would be set on every day of the month.

---


tivisse:
@robodoo r+

PR #244409, opened 2026-01-19
[IMP] website, html_builder: image alt SEO checks on translated pages

chdh-odoo:
#### [IMP] website: add branding in preview mode inside translated pages

For a logged in editor outside of edit mode, branding depended on the
language: in the default one, views and fields both carried it; in any
other, only fields did. `editable` is now dropped only when the page is
really rendered with `edit_translations`, so views are branded in every
language, which is how the dialog resolves each image to its record.

#### [IMP] website: edit image alts on translated pages

Alt attributes used to be read from branding metadata that translated
pages do not carry. They are now read with `edit_translations`, where
each alt is a marker span holding the sha of its base language term.
Saving groups the updated alts per record and sends them to
`/website/field/translation/update`, storing them as translations
instead of rewriting the view arch.

#### [IMP] website: flush delayed translations on SEO save

Saving the dialog only updated the records it edited itself; delayed
translations elsewhere stayed delayed until a later save from translation
mode. The page is now prefetched with `edit_translations=1` on opening,
un-awaited, so its `.o_delay_translation` markers can be read at save
time. Saving then follows the same three steps as translation mode, whose
last step both flows now share through `save_utils.js` and
`delayed_translation.js`.

### DIFF
--- a/addons/website/static/src/builder/plugins/translation/save_translation_plugin.js
+++ b/addons/website/static/src/builder/plugins/translation/save_translation_plugin.js
@@ -2,45 +2,23 @@ import { Plugin } from "@html_editor/plugin";
 import { withSequence } from "@html_editor/utils/resource";
 import { rpc } from "@web/core/network/rpc";
 import { registry } from "@web/core/registry";
+import { saveDelayTranslations } from "@website/js/delayed_translation";
 
 export class SaveTranslationPlugin extends Plugin {
     static id = "saveTranslation";
-    static dependencies = ["savePlugin", "websiteSavePlugin"];
+    static dependencies = ["websiteSavePlugin"];
 
     /** @type {import("plugins").WebsiteResources} */
     resources = {
-        on_will_save_handlers: this.saveDelayTranslations.bind(this),
+        on_will_save_handlers: (rootEl, groupedDirtyElements) =>
+            saveDelayTranslations(
+                rootEl,
+                this.services.website.currentWebsite.metadata.lang,
+                groupedDirtyElements
+            ),
         save_elements_overrides: withSequence(20, this.saveTranslationElements.bind(this)),
     };
 
-    async saveDelayTranslations(root = this.editable, groupedDirtyElements) {
-        // Don't take dirty elements as they will be saved
-        const cleanDelayTranslationEls = [
-            ...root.querySelectorAll(".o_delay_translation:not(.o_dirty)"),
-        ];
-        const groupedDelayTranslationElements =
-            this.dependencies.savePlugin.groupElements(cleanDelayTranslationEls);
-        const updateTranslationProms = [];
-        const currentWebsiteLang = this.services.website.currentWebsite.metadata.lang;
-        const translations = {};
-        translations[currentWebsiteLang] = {};
-        for (const [key, els] of Object.entries(groupedDelayTranslationElements)) {
-            // Keep only delay translation related to particular field that will
-            // not be updated by a modified (dirty) element
-            if (groupedDirtyElements[key]) {
-                continue;
-            }
-            updateTranslationProms.push(
-                rpc("/website/field/translation/update", {
-                    model: els[0].dataset["oeModel"],
-                    record_id: [Number(els[0].dataset["oeId"])],
-                    field_name: els[0].dataset["oeField"],
-                    translations,
-                })
-            );
-        }
-        return Promise.all(updateTranslationProms);
-    }
     /**
      * If the elements hold a translation, saves it. Otherwise, fallback to the
      * standard saving with the lang kept.
--- a/addons/website/static/src/components/dialog/seo.js
+++ b/addons/website/static/src/components/dialog/seo.js
@@ -22,6 +22,8 @@ import {
     useEffect,
 } from "@odoo/owl";
 import wUtils from "@website/js/utils";
+import { saveDelayTranslations } from "@website/js/delayed_translation";
+import { getRecordKey } from "@html_builder/core/save_utils";
 
 // This replaces \b, because accents(e.g. à, é) are not seen as word boundaries.
 // Javascript \b is not unicode aware, and words beginning or ending by accents won't match \b
@@ -1003,6 +1005,13 @@ export class OptimizeSEODialog extends Component {
                 ? this.data.website_meta_description
                 : "";
             this.isDefaultLang = this.data.lang?.code === this.data.default_lang_code;
+            // On translated pages, prefetch the page rendered with translation
+            // branding (un-awaited) so that, at save time, we can resolve every
+            // delayed translation on the page without blocking the dialog
+            // opening. Consumed only in save().
+            this.delayedTranslationsProm = this.isDefaultLang
+                ? null
+                : this.fetchDelayedTranslations();
             seoContext.description = storedDescription || this.getMeta({ name: "description" });
             this.previewDescription = _t(
                 "Your page description should be between 50 and 160 characters long."
@@ -1047,6 +1056,31 @@ export class OptimizeSEODialog extends Component {
 
     get pageDocumentElement() {
         return this.website.pageDocument.documentElement;
+    }
+
+    /**
+     * Fetch the current page rendered with `edit_translations` so its delayed
+     * translation markers (`.o_delay_translation`) can be read client-side.
+     *
+     * @returns {Promise<HTMLElement|null>}
+     */
+    async fetchDelayedTranslations() {
+        const url = new URL(this.website.pageDocument.location.href);
+        url.searchParams.set("edit_translations", "1");
+        try {
+            const response = await fetch(url, { credentials: "same-origin" });
+            if (!response.ok) {
+                throw new Error(`${response.status} ${response.statusText}`);
+            }
+            const html = await response.text();
+            // Restrict to the same root as the editable in translation mode, so
+            // that saving here confirms the delayed translations of exactly the
+            // records a save from that mode would confirm.
+            return new DOMParser().parseFromString(html, "text/html").querySelector("#wrapwrap");
+        } catch (error) {
+            console.warn(`Could not prefetch delayed translations for ${url}`, error);
+            return null;
+        }
     }
 
     async getImages() {
@@ -1142,7 +1176,7 @@ export class OptimizeSEODialog extends Component {
         if (!this.isDefaultLang) {
             const translationsByRecord = {};
             for (const img of seoContext.updatedAlts || []) {
-                const key = `${img.res_model}::${img.res_id}::${img.field}`;
+                const key = getRecordKey(img.res_model, img.res_id, img.field);
                 translationsByRecord[key] ??= {
                     model: img.res_model,
                     id: img.res_id,
@@ -1161,6 +1195,14 @@ export class OptimizeSEODialog extends Component {
                             [currentLang]: record.translations,
                         },
                     })
+                );
+            }
+            // The delayed translations are collected on the page prefetched
+            // with `edit_translations`.
+            const translationsRootEl = await this.delayedTranslationsProm;
+            if (translationsRootEl) {
+                rpcCalls.push(
+                    saveDelayTranslations(translationsRootEl, currentLang, translationsByRecord)
                 );
             }
         } else if (seoContext.updatedAlts?.length) {
--- a/addons/website/static/src/js/delayed_translation.js
+++ b/addons/website/static/src/js/delayed_translation.js
@@ -1,0 +1,38 @@
+import { rpc } from "@web/core/network/rpc";
+import { groupElements } from "@html_builder/core/save_utils";
+
+/**
+ * Save the delayed translations of the records found under `rootEl`, except the
+ * ones saved with a new value, which flushes their delayed terms anyway.
+ *
+ * @param {ParentNode} rootEl - element to search.
+ * @param {string} lang - current website language code.
+ * @param {Object<string, unknown>} [groupedDirtyElements] - the records saved
+ *        with a new value, keyed as in `groupElements`.
+ * @returns {Promise}
+ */
+export function saveDelayTranslations(rootEl, lang, groupedDirtyElements = {}) {
+    // Don't take dirty elements as they will be saved
+    const cleanDelayTranslationEls = [
+        ...rootEl.querySelectorAll(".o_delay_translation:not(.o_dirty)"),
+    ];
+    const groupedDelayTranslationElements = groupElements(cleanDelayTranslationEls);
+    const updateTranslationProms = [];
+    const translations = { [lang]: {} };
+    for (const [key, els] of Object.entries(groupedDelayTranslationElements)) {
+        // Keep only delay translation related to particular field that will
+        // not be updated by a modified (dirty) element
+        if (groupedDirtyElements[key]) {
+            continue;
+        }
+        updateTranslationProms.push(
+            rpc("/website/field/translation/update", {
+                model: els[0].dataset["oeModel"],
+                record_id: [Number(els[0].dataset["oeId"])],
+                field_name: els[0].dataset["oeField"],
+                translations,
+            })
+        );
+    }
+    return Promise.all(updateTranslationProms);
+}
--- a/addons/website/static/tests/tours/website_optimize_seo_menu.js
+++ b/addons/website/static/tests/tours/website_optimize_seo_menu.js
@@ -97,6 +97,10 @@ registerWebsitePreviewTour(
             trigger: ":iframe .s_text_image img[alt='alt text in English']",
         },
         ...switchToLang("fr"),
+        {
+            content: "The footer still shows its confirmed translation",
+            trigger: ":iframe #footer:contains('Liens utiles')",
+        },
         ...openSeoModal(),
         {
             content: "Alt text is prefilled from default language",
@@ -116,6 +120,10 @@ registerWebsitePreviewTour(
         {
             content: "Image alt attribute is updated in French",
             trigger: ":iframe .s_text_image img[alt='alt text in French']",
+        },
+        {
+            content: "The delayed footer translation has been confirmed",
+            trigger: ":iframe #footer:contains('Handy Links')",
         },
         ...switchToLang("en"),
         {
--- a/addons/html_builder/__manifest__.py
+++ b/addons/html_builder/__manifest__.py
@@ -18,6 +18,7 @@
         ],
         'web.assets_backend': [
             'html_builder/static/src/utils/keyboard_navigation.js',
+            'html_builder/static/src/core/save_utils.js',
         ],
         # this bundle is lazy loaded when the editor is ready
         'html_builder.assets': [
--- a/addons/html_builder/static/src/core/save_plugin.js
+++ b/addons/html_builder/static/src/core/save_plugin.js
@@ -1,7 +1,7 @@
 import { Plugin } from "@html_editor/plugin";
 import { withSequence } from "@html_editor/utils/resource";
-import { uniqueId } from "@web/core/utils/functions";
 import { isZWS } from "@html_editor/utils/dom_info";
+import { groupElements } from "@html_builder/core/save_utils";
 import { _t } from "@web/core/l10n/translation";
 import { EDITOR_MUTATION_TYPES } from "@html_editor/core/dom_observer_plugin";
 
@@ -10,7 +10,6 @@ import { EDITOR_MUTATION_TYPES } from "@html_editor/core/dom_observer_plugin";
  * @typedef { Object } SaveShared
  * @property { SavePlugin['save'] } save
  * @property { SavePlugin['ignoreDirty'] } ignoreDirty
- * @property { SavePlugin['groupElements'] } groupElements
  */
 
 /**
@@ -31,7 +30,7 @@ import { EDITOR_MUTATION_TYPES } from "@html_editor/core/dom_observer_plugin";
 
 export class SavePlugin extends Plugin {
     static id = "savePlugin";
-    static shared = ["save", "ignoreDirty", "groupElements"];
+    static shared = ["save", "ignoreDirty"];
     static dependencies = ["history", "domReferenceMap"];
 
     /** @type {import("plugins").BuilderResources} */
@@ -67,30 +66,12 @@ export class SavePlugin extends Plugin {
         this.canObserve = false;
     }
 
-    groupElements(toGroupEls) {
-        return Object.groupBy(toGroupEls, (toGroupEl) => {
-            const model = toGroupEl.dataset.oeModel;
-            const recordId = toGroupEl.dataset.oeId;
-            const field = toGroupEl.dataset.oeField;
-
-            // There are elements which have no linked model as something
-            // special is to be done "to save them". In that case, do not group
-            // those elements.
-            if (!model) {
-                return uniqueId("special-element-to-save-");
-            }
-
-            // Group elements which are from the same field of the same record.
-            return `${model}::${recordId}::${field}`;
-        });
-    }
-
     async save({ shouldSkipAfterSaveHandlers = async () => true } = {}) {
         let skipAfterSaveHandlers;
         try {
             // Get elements to save, then group them if possible.
             const dirtyEls = this.getResource("dirty_els_providers").flatMap((p) => [...p()]);
-            const groupedElements = this.groupElements(dirtyEls);
+            const groupedElements = groupElements(dirtyEls);
             await Promise.all(
                 this.trigger("on_will_save_handlers", this.editable, groupedElements)
             );
--- a/addons/html_builder/static/src/core/save_utils.js
+++ b/addons/html_builder/static/src/core/save_utils.js
@@ -1,0 +1,37 @@
+import { uniqueId } from "@web/core/utils/functions";
+
+/**
+ * Build the key under which the elements of a given record's field are grouped
+ * to be saved together.
+ *
+ * @param {string} model
+ * @param {string|number} recordId
+ * @param {string} field
+ * @returns {string}
+ */
+export function getRecordKey(model, recordId, field) {
+    return `${model}::${recordId}::${field}`;
+}
+
+/**
+ * Group the elements which are from the same field of the same record.
+ *
+ * @param {Iterable<HTMLElement>} toGroupEls
+ * @returns {Object.<string, HTMLElement[]>}
+ */
+export function groupElements(toGroupEls) {
+    return Object.groupBy(toGroupEls, (toGroupEl) => {
+        const model = toGroupEl.dataset.oeModel;
+        const recordId = toGroupEl.dataset.oeId;
+        const field = toGroupEl.dataset.oeField;
+
+        // There are elements which have no linked model as something
+        // special is to be done "to save them". In that case, do not group
+        // those elements.
+        if (!model) {
+            return uniqueId("special-element-to-save-");
+        }
+
+        return getRecordKey(model, recordId, field);
+    });
+}
--- a/addons/website/__manifest__.py
+++ b/addons/website/__manifest__.py
@@ -480,6 +480,7 @@
             'website/static/src/components/edit_head_body_dialog/edit_head_body_dialog.scss',
             'website/static/src/components/edit_head_body_dialog/edit_head_body_dialog.xml',
             'website/static/src/utils/**/*',
+            'website/static/src/js/delayed_translation.js',
             'website/static/src/components/dialog/*.js',
             'website/static/src/components/dialog/*.scss',
             'website/static/src/components/dialog/*.xml',
--- a/addons/website/tests/test_ui.py
+++ b/addons/website/tests/test_ui.py
@@ -914,4 +914,22 @@ class TestUi(HttpCaseWithWebsiteUser):
 
     def test_seo_multilang_alt_check(self):
         self.add_fr_language_to_website()
+
+        # Leave a delayed translation on the footer
+        footer = self.env.ref('website.footer_custom')
+        footer.update_field_translations('arch_db', {'fr_FR': {'Useful Links': 'Liens utiles'}})
+        footer.with_context(delay_translations=True).arch = footer.arch.replace(
+            'Useful Links', 'Handy Links'
+        )
+        self.assertIn(
+            'o_delay_translation',
+            footer.with_context(lang='fr_FR', edit_translations=True).arch,
+            "the footer should hold a delayed translation before the tour",
+        )
         self.start_tour(self.env['website'].get_client_action_url("/", True), "seo_multilang_alt_check", login="admin")
+        self.assertNotIn(
+            'o_delay_translation',
+            footer.with_context(lang='fr_FR', edit_translations=True).arch,
+            "saving the SEO dialog should confirm the delayed translations of the page",
+        )
+        self.assertIn('Handy Links', footer.with_context(lang='fr_FR').arch)


addons/website/static/src/components/dialog/seo.js
```diff
@@ -813,6 +814,14 @@ export class SeoChecks extends Component {
             uniqueRecords.add(`${model}||${id}||${field}||${type}`);
         });
 
+        // Fallback for translated pages
+        if (uniqueRecords.size === 0 && !htmlEl.dataset.editable && htmlEl.dataset.translatable) {
+            const viewId = parseInt(htmlEl.dataset.viewid, 10);
+            if (viewId) {
```
shsa-odoo:
```suggestion
        // Fallback for translated pages
        const {editable, translatable, viewid} = htmlEl.dataset;
         if (!(uniqueRecords.size || editable) && translatable) {
             const viewId = parseInt(viewid, 10);
```

addons/website/controllers/main.py
```diff
@@ -967,12 +968,23 @@ def get_alt_images(self, models):
...
         for img in imgs:
             record = request.env[img['res_model']].browse(img['res_id'])
+            base_lang = record._get_base_lang()
             if not record.has_access('write'):
                 continue
-            img['field'] = 'arch_db' if img['field'] == 'arch' else img['field']
-            tree = html.fromstring(str(record[img['field']]))
+            field_name = 'arch_db' if img['field'] == 'arch' else img['field']
+            if current_lang != base_lang:
+                translations = {current_lang: {img['source_sha']: img['alt']}}
```
shsa-odoo: "There's a security inconsistency: img['source_sha'] is hashed for identification, but img['alt'] is passed raw without escaping. If the goal is to securely transmit data, shouldn't img['alt'] be escaped ?"

addons/website/static/src/components/dialog/seo.js
```diff
@@ -790,7 +790,8 @@ export class SeoChecks extends Component {
         const uniqueRecords = new Set();
 
         // Select all relevant <img> elements in the editable page.
-        const imgEls = this.website.pageDocument.documentElement.querySelectorAll("#wrapwrap img");
+        const htmlEl = this.website.pageDocument.documentElement;
```
shsa-odoo: documentEl :'), htmlEl looks too vague.

shsa-odoo (commented):
@chdh-odoo Good work, Here is my initial review, please look into it.

addons/website/controllers/main.py
chdh-odoo: hmm, agree thanks for pointing it out :)

addons/website/controllers/main.py
```diff
@@ -1027,12 +1037,27 @@ def get_alt_images(self, models):
     def update_alt_images(self, imgs):
         if not request.env.user.has_group('website.group_website_restricted_editor'):
             raise werkzeug.exceptions.Forbidden()
+        current_lang = request.env.context.get('lang')
```
shsa-odoo: We dont need to request.env, could be handled by self.env.

image

check other places as well.

addons/website/static/src/components/dialog/seo.js
```diff
@@ -779,11 +779,20 @@ export class SeoChecks extends Component {
         this.seoContext.updatedAlts = this.state.altAttributes.filter((img) => img.updated);
     }
 
+    get showDecorativeOption() {
+        const { langName, defaultLangName } = this.website.currentWebsite.metadata;
+        if (langName !== undefined) {
```
shsa-odoo: In which case the langName will be undefined ?

addons/website/controllers/main.py
```diff
@@ -1000,22 +1000,32 @@ def seo_suggest(self, keywords=None, lang=None):
...
+                    if index < len(source_imgs)
+                    else alt
+                )
                 if not decorative or alt is None:
                     result.append({
                         "src": el.get("src"),
                         "alt": alt or "",
                         "decorative": False,
                         "updated": False,
+                        "source_sha": sha256((source_alt or "").encode()).hexdigest(),
```
shsa-odoo: When multiple images have the same alt text, they generate the same SHA256 hash. On translated pages, the _update_field_translations() call uses this hash to match terms, causing all images with identical alt text to be updated simultaneously when only one should be changed.
Example:

Image 1: alt="cat" → sha256("cat") = abc123
Image 2: alt="cat" → sha256("cat") = abc123 ← Same hash!
When updating Image 1's translation to "chat", both images get replaced.
```suggestion
                        "source_sha": sha256(f"{source_alt or ''}:{index}".encode()).hexdigest(),
```

chdh-odoo: We will have separate task for this since this is generic issue and we need to fix it in side bar as well

addons/website/static/src/components/dialog/seo.js
chdh-odoo: In case of default lang it will be `undefined`

gppa-odoo:
Hello @dhruv2x 
While testing i found this weird bug, Can you please check?
https://www.loom.com/share/ee9e8a81c62e46b4b81de9f212f5bfe5

addons/website/controllers/main.py
```diff
@@ -1030,26 +1030,44 @@ def seo_suggest(self, keywords=None, lang=None):
...
-                if not decorative or alt is None:
+                raw_alt = el.get('alt')
+                alt = (raw_alt or '').strip()
+                source_alt = (
+                   (source_imgs[index].get('alt') or '').strip()
+                    if index < len(source_imgs)
+                    else alt
+                )
+                # Alt tag must first be defined in the base language
+                if current_lang != base_lang and not source_alt:
```
gppa-odoo: `current_lang != base_lang` -> `is_translated`
Since you already made that variable above.

dhruv2x:
hello @loco-odoo 

I found alternative way to fetch delayed translations in the Optimize SEO modal. I just wanted your thoughts on second POC [commit](https://github.com/odoo/odoo/pull/244409/changes/a0c04a835fd7918d2eac6565c6ca93f7398739bc), since the delayed translation flow is a bit complicated! Instead of computing `source_sha` through source image index mapping, it reads the translated arch with `edit_translations=True` and extracts the hash directly from the translation markers.

loco-odoo:
Hello @dhruv2x :wave: 
> I found alternative way to fetch delayed translations in the Optimize SEO modal. I just wanted your thoughts on second POC commit,..., Instead of computing source_sha through source image index mapping, it reads the translated arch with edit_translations=True and extracts the hash directly from the translation markers

At first look, it feels that it is indeed a good idea to read the view that would have been produced in translate mode :+1:
I actually wanted to discuss the SEO dialog with the PO's at the office today. Indeed, while reviewing your PR, I landed on [this commit](https://github.com/odoo/odoo/commit/006b943a5dcc9bf988756bac333d2857b07e3be2) that explains that images of dynamic snippets should not be part of the SEO dialog (as they are added by an interaction). IMO, the concern goes beyond dynamic snippets; when we add an `alt` on an image through the SEO dialog, we actually edit the view. Because of that, we should take the concept of editable area into account. As the editable areas are build in edit mode, I think we should consider moving the SEO dialog in edit mode.
Your PR also goes in that direction; modifying the SEO dialog of a page in another language needs the view that would have been created in translate mode :+1: 

I will have a discussion with the PO's today and let you know what is the output of the discussion but maybe that can be done in another PR

dhruv2x:
sure, I've just squashed commits :)

addons/website/static/src/components/dialog/seo.js
```diff
@@ -815,6 +824,12 @@ export class SeoChecks extends Component {
             uniqueRecords.add(`${model}||${id}||${field}||${type}`);
         });
 
+        // Fallback for translated pages
+        const viewId = parseInt(documentEl.dataset.viewid, 10);
```
loco-odoo: Why adding a radix (10)?

dhruv2x:
hello @loco-odoo, may I know if we have some conclusion from the discussion you had with PO?

loco-odoo:
> hello @loco-odoo, may I know if we have some conclusion from the discussion you had with PO?

Sure, I was on an urgent task that we wanted to have for the freeze hence the delay. Concerning the discussion we had, let me do a recap; In an ideal world, the SEO edition should deal with the content page in edit mode. Indeed, we should only deal with information that are in editable areas (this is only known in edit mode as the `contenteditable` attribute is added by plugins). However, from a functional point of view, we still want the SEO optimizer to be available outside of the edit mode. The conclusion is that for the moment, we will continue to do the way is it done now (not dealing with the content of the page in edit mode). In a future task, we will work on a way to instantiate a light `Editor` outside of the edit mode; this will be usefull for the SEO optimizer but also for the edition of website related html fields in the backend etc... But again, this will be for a future task :+1: 

Concerning your task, I do not forget you and I'll review it ASAP :+1:

addons/website/static/src/components/dialog/seo.js
```diff
@@ -815,6 +824,12 @@ export class SeoChecks extends Component {
...
+        if (!this.props.isDefaultLang && !uniqueRecords.size && viewId) {
+            uniqueRecords.add(`ir.ui.view||${viewId}||arch||html`);
+        }
```
loco-odoo: This feels weird. If you do it like this you'll have:
- Go on a product page
- Add an image at the top of the page so that it is available for all the product  (when you begin to drag a dropzone appears with the "Drags block here to add them on all the products" message)
- Add another one in the `description_ecommerce` field.
- Add an `alt` on both images
- Go in another language
- Open the SEO dialog
-> There is only one image available

Actually, how will you do for the image at the top of the page as it seems that there are no information about views when it is rendered in another language than the website default one?

addons/website/static/src/components/dialog/seo.js
```diff
@@ -1131,7 +1149,45 @@ export class OptimizeSEODialog extends Component {
...
+                    id,
+                    fieldName,
+                    translations: {},
+                });
+                return translationsByRecord.get(recordKey);
+            };
+
+            for (const record of seoContext.translationRecords) {
+                addTranslationRecord(record.model, record.id, record.field);
+            }
```
loco-odoo: Why? By doing so, don't you send useless info to the `/website/field/translation/update` route?

addons/website/static/src/components/dialog/seo.js
```diff
@@ -1131,7 +1149,45 @@ export class OptimizeSEODialog extends Component {
...
+                    rpc("/website/field/translation/update", {
+                        model: record.model,
+                        record_id: [record.id],
+                        field_name: record.fieldName,
+                        translations: {
+                            [currentLang]: record.translations,
+                        },
+                    })
+                );
+            }
```
loco-odoo, line 1199: Can't we do that in only one `for` loop?

addons/website/controllers/main.py
```diff
@@ -1374,27 +1382,46 @@ def seo_suggest(self, keywords=None, lang=None):
...
     def get_alt_images(self, models):
         result = []
+        current_lang = self.env.context.get('lang')
         for model in models:
             record = request.env[model['model']].browse(model['id'])
-            model['field'] = 'arch_db' if model['field'] == 'arch' else model['field']
-            tree = html.fromstring(str(record[model['field']]))
+            field_name = 'arch_db' if model['field'] == 'arch' else model['field']
+            base_lang = record._get_base_lang()
+            is_translated = current_lang != base_lang
```
loco-odoo: Nitpick: you can check this outside of the `for` loop: you can take the current website and take the `default_lang_id` of this website 👍

addons/website/static/src/components/dialog/seo.js
```diff
@@ -775,15 +778,20 @@ export class SeoChecks extends Component {
...
         onMounted(() => {
-            this.getBrokenLinks();
+            if (this.props.isDefaultLang) {
+                this.getBrokenLinks();
+            }
         });
     }
 
     imgUpdated(img) {
+        img.alt = (img.alt || "").trim();
```
loco-odoo: Why? Should this change be in this commit? Should it be backported to stable versions?

addons/website/static/tests/tours/seo_multilang_alt_check.js
```diff
@@ -0,0 +1,104 @@
...
+    {
+        content: `Switch to ${lang}`,
+        trigger: `:iframe .js_change_lang[data-url_code='${lang}']`,
+        run: "click",
+    },
+    {
+        content: `Wait until ${lang} is applied`,
+        trigger: `:iframe html[lang*="${lang}"]`,
+    },
+    stepUtils.waitIframeIsReady(),
```
loco-odoo: Not needed?

addons/website/static/tests/tours/seo_multilang_alt_check.js
```diff
@@ -0,0 +1,104 @@
+import {
+    registerWebsitePreviewTour,
+    insertSnippet,
+    clickOnSave,
+} from "@website/js/tours/tour_utils";
+import { stepUtils } from "@web_tour/tour_utils";
+
+const switchTo = (lang) => [
```
loco-odoo: Maybe it is time to extract this in a util and use it everywhere on the tests (in another commit)

addons/website/controllers/main.py
```diff
@@ -1415,7 +1442,7 @@ def update_alt_images(self, imgs):
                         element.set('alt', '')
                         element.set('role', 'presentation')
                     else:
-                        element.set('alt', markup_escape(img['alt']))
+                        element.set('alt', markup_escape((img.get('alt') or '').strip()))
```
loco-odoo: Why?

addons/website/static/src/components/dialog/seo.js
```diff
@@ -1131,7 +1149,45 @@ export class OptimizeSEODialog extends Component {
                 })
             );
         }
-        if (seoContext.updatedAlts?.length) {
+        if (!this.isDefaultLang) {
+            const translationsByRecord = new Map();
+            const getFieldName = (field) => (field === "arch" ? "arch_db" : field);
```
loco-odoo: Needed in translate mode?

addons/website/controllers/main.py
```diff
@@ -1374,27 +1382,46 @@ def seo_suggest(self, keywords=None, lang=None):
...
+                alt = (raw_alt or '').strip()
+                values = {
+                    "src": el.get("src"),
+                    "alt": alt,
+                    "decorative": False,
+                    "updated": False,
+                    "res_model": model['model'],
+                    "res_id": model['id'],
+                    "id": f"{model['model']}-{model['id']}-{index}",
+                    "field": field_name,
```
loco-odoo, line 1477: Is there a reason you changed `model.get('field')` to `field_name` ? (Those values can be different especially in the website default language)

addons/website/static/src/components/dialog/seo.js
dhruv2x: We need two separate loops because multiple images on the same page can belong to the same translation record. For example, if two alt texts are updated on the same page, they are grouped into a single bucket:

for example, if two alt texts are updated on the same page, they are grouped into a single bucket:

```
{
  "model": "ir.ui.view",
  "id": 1247,
  "fieldName": "arch_db",
  "translations": {
    "sha1": "a",
    "sha2": "b"
  }
}
```
The first loop collects all alt translations for the same record into this bucket. The second loop sends a single RPC containing all translations :)

If we sent the RPC while building the bucket, we'd end up making two concurrent calls for the same ir.ui.view record: one containing {sha1: "a"} and another containing {sha2: "b"}. Since the translation endpoint replaces the field value rather than merging it, one call could overwrite the other and cause a translation to be lost^^

addons/website/static/src/components/dialog/seo.js
dhruv2x: This is mainly because of delayed translations.

For example, if a user adds a new image in the default language, then switches to a translated language and opens the Optimize SEO modal, we'll still show that image in translated mode through the delayed translation mechanism.(since we do show that in translation mode)

Now, if the user doesn't change anything and just clicks Save, we still need to send it through the translation flow so that the newly added image gets saved for that language as well. Otherwise, the image was visible in the modal but wouldn't get persisted in the translation.

So the extra record isn't really about a user change. It's there to keep the delayed translation behavior consistent and make sure newly added images are saved even when the user just opens the modal and hits Save :)

addons/website/static/src/components/dialog/seo.js
dhruv2x: I added this to make intent explicit, altho `documentEl.dataset.viewid` will always be integer string as far as I know, do you think I should remove radix?

addons/website/controllers/main.py
dhruv2x: The value is effectively the same, `field_name` contains the same normalized value (arch → arch_db), so the behavior is unchanged. I just reused the local variable instead of updating model['field'] and reading it back later.

addons/website/static/src/components/dialog/seo.js
dhruv2x: `getFieldName` turns "arch" into "arch_db" so both sources match.

It makes sure we send "arch_db" to the server, which is the real stored column. "arch" is computed and would be the wrong field. Just like translation mode where we update `arch_db` too.

addons/website/controllers/main.py
dhruv2x: ahh! thanks for noticing

addons/website/controllers/main.py
dhruv2x: I just added defensive dict access with trim leading/trailing whitespaces from alt text.

We already auto-trim alt text in edit mode, so I wanted to keep the same behavior here as well. except transaltion mode options which is wrong and it SHOULD trim. leading/trailing whitespaces doesn't buy much

addons/website/static/src/components/dialog/seo.js
dhruv2x: Without trimming, values like "    " get saved as actual alt text. 

> Should this change be in this commit?

hmm,  I should have a separate commit for this
 
> Should it be backported to stable versions?

I think we should, leading/trailing whitespaces don't provide any meaningful information for screen readers nor for SEO 

EDIT: for master https://github.com/odoo/odoo/pull/244409/changes/1ced09b07a96df2e618982ffa5d70353b4bc11bf and for stable I'll create backport PR after this one

addons/website/static/tests/tours/seo_multilang_alt_check.js
dhruv2x: I think we should move this to the end of `saveSeoModal()`. After saving the SEO modal, it takes a bit of time to return to the homepage, which makes this step flaky and causes occasional failures.

addons/website/static/tests/tours/seo_multilang_alt_check.js
dhruv2x: yes I just saw someone actually did yesterday 

https://github.com/odoo/odoo/pull/271303/changes/bafae1a551adc1e8b96245f5efde31bb38a0c6bc

so I can just re-use this util :)

addons/website/static/src/components/dialog/seo.js
loco-odoo: Concerning that :point_up: , I asked to chm and it seems that there is no particular valid reason for which branding of views is not displayed in languages that are different than the default one. I made a [testing PR](https://github.com/odoo/odoo/pull/271849) to check that no tests are failing when the branding is displayed. Feel free to reuse it in a commit of this PR, it should solve this:
> Actually, how will you do for the image at the top of the page as it seems that there are no information about views when it is rendered in another language than the website default one?

dhruv2x: ohh cool, but if we ended up not keeping branding on translated pages, I created [POC commit](https://github.com/odoo/odoo/pull/244409/changes/4520b43c8ee4ec08f8d4ba0e71ed76c4d7c97b0c) , worth a look :)

(Note that it's still need some cleanup)

loco-odoo: In this case, in order to avoid unnecessary rpc I would go with the fix I explained but indeed, otherwise, if it would have been a problem to show branding of view in languages different than the default one, your way of doing would have been a good solution :+1:

dhruv2x:
hello @loco-odoo 

I've added 2 more commits. Can you have a look? :)

dhruv2x:
@loco-odoo JFI: resolved merge conflicts and removed last commit as the fix is already merged by https://github.com/odoo/odoo/commit/6981dffcfbaf6978852b71f0efd9bc7b5e16fe92 commit recently

addons/website/controllers/main.py
```diff
@@ -105,6 +105,14 @@
     ]
 }
 
+# Matches the translation marker inside attribute values
+# (e.g. alt="<span data-oe-translation-source-sha="...">term</span>")
+# when a record is read with edit_translations=True.
+_TRANSLATION_MARKER_MATCHER = re.compile(
+    r'<span[^>]*data-oe-translation-source-sha="([^"]+)"[^>]*>(.*?)</span>',
```
loco-odoo: Nitpick: 
```diff
- <span[^>]
+ <span [^>]
```

addons/website/controllers/main.py
```diff
@@ -1417,7 +1443,7 @@ def update_alt_images(self, imgs):
             tree = html.fromstring(str(record[img['field']]))
             modified = False
             for index, element in enumerate(tree.xpath('//img')):
-                imgId = f"{img['res_model']}-{img['res_id']}-{index!s}"
+                imgId = f"{img['res_model']}-{img['res_id']}-{img['field']}-{index!s}"
```
loco-odoo: This will be solved by https://github.com/odoo/odoo/pull/277588 👍

addons/website/static/src/components/dialog/seo.js
loco-odoo: With the new implementation, is this part still needed? If we take the images that have been modified (`seoContext.updatedAlts`) and the `.o_delay_translation` elements, isn't it enough?

addons/website/static/src/components/dialog/seo.js
loco-odoo: Then, it is actually better to create another PR where all the forward ports will be merged rather than one that we will have to merge partly; it will be easier for future comprehension

addons/website/models/website.py
```diff
@@ -1541,7 +1541,7 @@ def _render_template(self, template, values=None):
             except AccessError:
                 pass
         translatable = has_group_restricted_editor and self.env.context.get('lang') != IrHttp._get_default_lang().code
-        editable = editable and not translatable
+        editable = editable and not (translatable and self.env.context.get('edit_translations'))
```
loco-odoo, line 1559: Maybe you can do that in a separate commit, it deserves a good commit message with an explanation like the one we had in the discussion on discord 👍

addons/website/static/src/utils/delayed_translation.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Collect the records targeted by the delayed-translation elements under
+ * `root`, deduplicated by record. Elements flagged `.o_dirty` are skipped.
+ *
+ * @param {ParentNode} root - element/document to search.
+ * @returns {Map<string, {key: string, model: string, id: number, field: string}>}
+ *
+ */
+export function collectDelayedTranslationRecords(root) {
```
loco-odoo: Instead of creating 2 new methods (`collectDelayedTranslationRecords` and `flushDelayedTranslations`) and having to be sync with `SavePlugin.groupElements`, why not extracting the entire `saveDelayTranslations` (and `groupElements`) in a util and reusing it directly for the SEO part?
Indeed, in translation, what we do is that we:
- Extract the dirty elements.
- Save the delayed translations that are not dirty.
- Save the dirty elements.

In the SEO case, it is the same logic: the dirty elements (`seoContext.updatedAlts`) need to be saved and the delayed translation elements that are not part of the dirty elements also need to be saved. As it is the same logic, let's use the same method 👍

loco-odoo:
It would be great if you can also add a test for the delay translation case (delay translation that is saved when saving the SEO dialog) :pray:

addons/website/static/src/components/dialog/seo.js
dhruv2x: hmm not needed.

addons/website/controllers/main.py
dhruv2x: yeah im just waiting for forward ports to be merged since without this the changes made in modal will not be saved

addons/website/static/src/utils/delayed_translation.js
dhruv2x: hmm makes sense that way we can avoid keeping both implementations in sync :+1:

dhruv2x:
resolved merge conflicts :crying_cat_face:

dhruv2x:
hello @loco-odoo ping :)

addons/website/static/tests/tours/website_optimize_seo_menu.js
```diff
@@ -1,6 +1,12 @@
...
+import {
+    registerWebsitePreviewTour,
+    insertSnippet,
+    clickOnSave,
+    switchToLang,
+} from "@website/js/tours/tour_utils";
+import { stepUtils } from "@web_tour/tour_utils";
 
-const openSeoModalSteps = [
+const openSeoModal = () => [
```
loco-odoo, line 9: Why this change?

addons/website/static/tests/tours/website_optimize_seo_menu.js
```diff
@@ -32,29 +51,80 @@ registerWebsitePreviewTour(
...
+            id: "s_text_image",
+            name: "Text - Image",
+            groupName: "Content",
+        }),
+        ...clickOnSave(),
+        ...openSeoModal(),
+        {
+            content: "Image is missing an alt description",
+            trigger: ".o_seo_images_check input.is-invalid",
+        },
```
loco-odoo: Not needed? (same trigger than the next instruction)

addons/website/static/src/components/dialog/seo.js
```diff
@@ -1163,6 +1194,18 @@ export class OptimizeSEODialog extends Component {
                     })
                 );
             }
+            // The delayed translations are collected on the document prefetched
+            // with `edit_translations`.
+            const translationsDoc = await this.delayedTranslationsProm;
+            if (translationsDoc) {
+                rpcCalls.push(
+                    saveDelayTranslations(
+                        translationsDoc.querySelector("#wrapwrap") || translationsDoc,
```
loco-odoo: Why `translationsDoc.querySelector("#wrapwrap")` ? Can we remove it?

addons/website/static/src/builder/plugins/translation/save_translation_plugin.js
```diff
@@ -14,32 +15,8 @@ export class SaveTranslationPlugin extends Plugin {
...
-                rpc("/website/field/translation/update", {
-                    model: els[0].dataset["oeModel"],
-                    record_id: [Number(els[0].dataset["oeId"])],
-                    field_name: els[0].dataset["oeField"],
-                    translations,
-                })
-            );
-        }
-        return Promise.all(updateTranslationProms);
+        return saveDelayTranslations(root, currentWebsiteLang, groupedDirtyElements);
```
loco-odoo: Maybe another function name to not have two `saveDelayTranslations` methods?

addons/website/static/src/js/editor/delayed_translation.js
```diff
@@ -0,0 +1,38 @@
...
+ * Save the delayed translations of the records found under `root`, except the
+ * ones saved with a new value, which flushes their delayed terms anyway.
+ *
+ * @param {ParentNode} root - element/document to search.
+ * @param {string} lang - current website language code.
+ * @param {Object.<string, any>} [groupedDirtyElements] - records saved with a
+ *        new value, keyed as in `groupElements`.
+ * @returns {Promise}
+ */
+export function saveDelayTranslations(root, lang, groupedDirtyElements = {}) {
```
loco-odoo: Maybe you can take the chance to rename `root` -> `rootEl` 👍

addons/website/static/src/js/editor/delayed_translation.js
```diff
@@ -0,0 +1,38 @@
+import { rpc } from "@web/core/network/rpc";
+import { groupElements } from "@html_builder/core/save_utils";
+
+/**
+ * Save the delayed translations of the records found under `root`, except the
+ * ones saved with a new value, which flushes their delayed terms anyway.
+ *
+ * @param {ParentNode} root - element/document to search.
+ * @param {string} lang - current website language code.
+ * @param {Object.<string, any>} [groupedDirtyElements] - records saved with a
```
loco-odoo: `{Object.<string, any>}` ?  Maybe you can check all your added docstring to be sure your `param` are valid elements?

addons/website/static/src/js/delayed_translation.js
```diff
@@ -0,0 +1,38 @@
+import { rpc } from "@web/core/network/rpc";
+import { groupElements } from "@html_builder/core/save_utils";
```
loco-odoo, line 2: Also, maybe `editor` is not the best folder to add this file 👍

addons/website/static/tests/tours/website_optimize_seo_menu.js
dhruv2x: I felt the 'Steps' suffix a bit annoying since it's already pretty clear that this is a utility that returns the steps for the tour! more of we also added util for save modal `const saveSeoModal` so just wanted to make both util consistent
(let me know if you want me to revert tho but then I have to change save util name to `saveSeoModalSteps` :cry: )

addons/website/static/src/components/dialog/seo.js
dhruv2x: This is needed, `saveDelayTranslations` scans its root for `.o_delay_translation`. In translation mode that root is the editable, which is `#wrapwrap`. here the root comes from DOMParser, so it's the whole document, a bigger scope than the flow it's meant to mirror.

That matters because the markers aren't limited to the editable. They're added to every translatable term of a view when it's read with `edit_translations`

I moved the scoping into `fetchDelayedTranslations`, which now returns the `#wrapwrap` element (or null) instead of the document, with a comment explaining why

addons/website/static/src/builder/plugins/translation/save_translation_plugin.js
dhruv2x: hmm I think inlining `on_will_save_handlers` makes more sense since we're anyway calling util method

addons/website/static/src/js/editor/delayed_translation.js
dhruv2x: For `groupedDirtyElements` I meant "an object whose keys are strings and whose values can be anything." 
since here we only care about keys and not values, maybe we can use `unknown` instead of `any`

addons/website/static/src/js/delayed_translation.js
dhruv2x: i'm bit skeptical for this one if you can help me out here ^^

My intution was that the file has to stay out of `web.assets_frontend`, it imports `@html_builder/core/save_utils`, which only exists in `web.assets_backend`. The tidier-looking homes `(static/src/utils/, static/src/core/)` are both globbed into the frontend bundle, so they'd ship it to every public visitor with a broken import. `js/editor/` isn't globbed anywhere, so listing it explicitly in `website.assets_editor` puts it in exactly one place reachable by both the SEO dialog and the translation plugin.

dhruv2x:
@loco-odoo 

accidentally rebased ( :disappointed: ) , so can't share the exact diff. but here's the change log since the last review (only last commit had changes)

1. Minor docstring updates for a few functions.
2. Renamed `root` to `rootEl` in `saveDelayTranslations`.
3. Moved `querySelector("#wrapwrap")` into `fetchDelayedTranslations` instead of passing it when calling `saveDelayTranslations` in seo.js.
4. inline `on_will_save_handlers` in the save translation plugin.

addons/website/static/src/js/delayed_translation.js
loco-odoo: Indeed, it seems there is no perfect place for it... Let's add it in the parent folder (`website/static/src/js`). It is not perfect but we will have to rework the content of the `website` folders at a moment anyway

dhruv2x: thanks for suggestion :)

pushed, here's [diff](https://github.com/odoo/odoo/compare/4e1a6001f293092078ad236abee2628e80c67d6c..f14d18a51db4b0c2362e40728baf2e33be9c6405)

loco-odoo:
Thnaks for the PR @dhruv2x !

@robodoo r+

loco-odoo:
@robodoo rebase-ff

PR #244410, opened 2026-01-19
[IMP] l10n_cn: change fapiao field size to 20

dijo-odoo:
with this commit:-
 - We're changing the char size to 20 from 8.
 - We remove the constraint that prevents us from saving the record if the size is not 8.

task-5501485


dijo-odoo:
As the task is cancelled, we're closing this PR. 
cc @hamo-odoo

PR #244412, opened 2026-01-19
[FIX] pos_loyalty: ensure loyalty is loaded in trusted pos

aksi-odoo:
Step to reproduce:
- Create a loyalty program for a specific product as for 1 $ spent gets 1 point
- Have a trusted pos for our main shop (both pos should have each other as trusted)
- Create an order from the main pos , select a partner (notice we  get LPs)
- save the order
- Open the order in the trusted pos
- we do not see the loyalty points

Cause:
- currently, when loading a order in trusted pos, there is no trigger to update rewards for selected order

Fix:
- the ticket screen in loyalty has been patched to update rewards after loading a order

opw-5412744

---

### DIFF
--- a/addons/pos_loyalty/static/tests/tours/pos_loyalty_tour.js
+++ b/addons/pos_loyalty/static/tests/tours/pos_loyalty_tour.js
@@ -691,3 +691,28 @@ registry.category("web_tour.tours").add("test_race_conditions_update_program", {
             },
         ].flat(),
 });
+
+registry.category("web_tour.tours").add("test_loyalty_in_trusted_pos_make_order", {
+    steps: () =>
+        [
+            Chrome.startPoS(),
+            Dialog.confirm("Open Register"),
+            ProductScreen.clickPartnerButton(),
+            ProductScreen.clickCustomer("AAAA"),
+            ProductScreen.addOrderline("Whiteboard Pen", "1", "100"),
+            PosLoyalty.pointsAwardedAre("100"),
+            ProductScreen.saveOrder(),
+        ].flat(),
+});
+
+registry.category("web_tour.tours").add("test_loyalty_in_trusted_pos", {
+    steps: () =>
+        [
+            Chrome.startPoS(),
+            Dialog.confirm("Open Register"),
+            Chrome.clickMenuOption("Orders"),
+            TicketScreen.selectOrder("-0001"),
+            TicketScreen.loadSelectedOrder(),
+            PosLoyalty.pointsAwardedAre("100"),
+        ].flat(),
+});
--- a/addons/pos_loyalty/static/src/overrides/components/ticket_screen/ticket_screen.js
+++ b/addons/pos_loyalty/static/src/overrides/components/ticket_screen/ticket_screen.js
@@ -51,4 +51,10 @@ patch(TicketScreen.prototype, {
         }
         return false;
     },
+    async _setOrder() {
+        await super._setOrder(...arguments);
+        if (this.pos.isOpenOrderShareable()) {
+            this.pos.updateRewards();
+        }
+    },
 });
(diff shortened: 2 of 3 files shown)


alsh-odoo (changes requested):
Thanks for the fix!

I noticed that when opening the order from the trusted device, the loyalty points are getting doubled. Could you please check this issue?

Here's the reference video (after applying your fix): https://drive.google.com/file/d/1a78-k-O_6t49RwkH1LGhQ2a5U9G8KkbV/view?usp=drive_link

alsh-odoo (approved):

addons/pos_loyalty/static/tests/tours/pos_loyalty_tour.js
```diff
@@ -674,3 +674,32 @@ registry.category("web_tour.tours").add("test_race_conditions_update_program", {
...
+            PosLoyalty.pointsAwardedAre("100"),
+            {
+                content: `save order`,
+                trigger: `.pads .fa-upload`,
+                run: "click",
+            },
+        ].flat(),
+});
+
+registry.category("web_tour.tours").add("test_loyalty_in_trusted_pos", {
```
sbel-odoo, line 708:
```suggestion
registry.category("web_tour.tours").add("test_loyalty_in_trusted_pos_load_order", {
```

addons/pos_loyalty/static/tests/tours/pos_loyalty_tour.js
```diff
@@ -674,3 +674,32 @@ registry.category("web_tour.tours").add("test_race_conditions_update_program", {
             },
         ].flat(),
 });
+
+registry.category("web_tour.tours").add("test_loyalty_in_main_pos", {
```
sbel-odoo:
```suggestion
registry.category("web_tour.tours").add("test_loyalty_in_trusted_pos_make_order", {
```

sbel-odoo (changes requested):
Hello [@aksi-odoo](https://github.com/aksi-odoo) thanks four your investigation,

Have you asked the R&D if this is something they want to fix? I have in mind that loyalty point features are not supported across different shops (/devices also). Make sure they want you to fix this.

If they want to fix this issue then I'd like you to check the issue mentioned by [@alsh-odoo](https://github.com/alsh-odoo) because I consider this part of the flow you are trying to fix. They were not shown or computed prior, now they are computed but not shown correctly, the fix is "partial" to me.

In the mean time, I've left very small nitpicking for search purposes.

sbel-odoo:
> As i said, the issue was already there, just not shown, this is already reported in opw-[5609964](https://www.odoo.com/web#id=5609964&view_type=form&model=project.task) while the fix is still in progress, #245400 i have mentioned the description and cause. **TL;DR** when ever `sync_from_ui` is called, we are calculating and reassigning the LPs, again ( IIRC, but final assignment is correct but on screeen it shows wrong value). Just trigger "sync_from_ui" to reproduce the issue, one way is to save a order.

Next time don't hesitate to share the other ticket/PR if you have one it would have made a difference :slightly_smiling_face:  Indeed I see that the problem was there without your fix. All I did was on a runbot without your fix, load the order on the second pos and trigger again the customer to notice it as well. 

I see the fix is not complicated on the other PR so if we fix this I would prefer to ensure the other fix is merged before this one if possible because it will highlight even more the other issue and I'd like to avoid having tickets created for something already ongoing.

addons/point_of_sale/static/tests/tours/utils/product_screen_util.js
```diff
@@ -729,3 +729,14 @@ export function addDiscount(discount) {
             .flatMap((key) => Numpad.click(key)),
     ].flat();
 }
+
+export function saveOrder() {
```
alsh-odoo, line 733: Looks like you forgotten to update the PR title 🙂

aksi-odoo: this function can be removed too if [this](https://github.com/odoo/odoo/pull/245400) is merged first , so not updating the title for now.

sbel-odoo (approved):
I'm ok with this PR. I'd prefer the other one to be merge before this one 👍

manv-afk (approved):

PR #244413, opened 2026-01-19
[FIX] website: remove header padding when visually separated from page

romo-odoo:
Before this PR, the left/right padding of the header would be removed when the header had no background color or when its background color matches the pages color. In these cases, the header boundaries were not visible and so we allowed the header to take more space.

However, the previous condition would not check if the page had a background image, or if the header had a shadow defined. The header would have visible boundaries and the element on the extremities would be stuck to the boundaries.

This commit updates the condition to include any possible option that would make the header boundaries visible.

task-5367502

### DIFF
--- a/addons/website/static/src/scss/website.scss
+++ b/addons/website/static/src/scss/website.scss
@@ -251,13 +251,28 @@ body {
                 max-width: 100% !important;
             }
 
-            // Vertical alignment when top-menu has visually "no background"
+            // Remove the left/right padding of the header navbar when the
+            // header has no visible separation from the body:
+            // - The header has no background color (transparent)
+            // - The header background color matches the body
+            // - The header has no shadow defined
             $-menu-color: o-color('menu-custom') or o-color('menu');
-            @if (not $-menu-color or $-menu-color == o-color('body')) {
-                > header {
-                    .navbar, .container {
-                        padding-left: 0;
-                        padding-right: 0;
+            $-header-template: o-website-value('header-template');
+            @if (
+                (not $-menu-color or ($-menu-color == o-color('body') and o-website-value('body-image') == null))
+                and (o-website-value('menu-box-shadow') == none)
+                and (index(("default", "hamburger", "stretch", "vertical", "search"), $-header-template) != null)
+            ) {
+                > header #o_main_nav {
+                    padding-left: 0;
+                    padding-right: 0;
+                    
+                    @if ($-header-template == "search") {
+                        & > div, 
+                        & > .o_header_hide_on_scroll > .container {
+                            padding-left: 0;
+                            padding-right: 0;
+                        }
                     }
                 }
             }


addons/website/static/src/scss/website.scss
bvr-odoo, line 1: - Commit message => "_Before this PR, the left/right padding of the header would be removed when the header had no background color or when its background color matches the pages color._" Also explain that is when **the header is not full** + add the steps to reproduce

addons/website/static/src/scss/website.scss
```diff
@@ -236,9 +236,18 @@ body {
                 max-width: 100% !important;
             }
 
-            // Vertical alignment when top-menu has visually "no background"
+            // Remove the left/right padding when the header has no visible
```
bvr-odoo:
```suggestion
            // Remove the left/right padding of the navbar header
```

bvr-odoo (changes requested):
- Since this is not a critical bug and it could introduce unwanted changes for some users, Lionel and I agreed to target version 19.0 for this fix instead.

- Another point: the padding removal is not only applied to the header but also to the mega menus. It probably made sense when this code was added, but it does not anymore. The default mega menu has a border, so without padding the text touches the border. In addition, with some mega menus (see the image below for example) it also looks quite bad.

You can therefore remove `.container` from the CSS selector.

Image

This `.container` class also causes an issue with some headers that use two colors. Example below.

Image


- Please check whether this rule makes sense for each header. Some headers already have their own padding defined, which overrides the rule, so we can probably exclude them since there is no point in applying a rule that has no effect.

addons/website/static/src/scss/website.scss
bvr-odoo, line 1: Adapt the PR description the same way you did for the commit message.

addons/website/static/src/scss/website.scss
```diff
@@ -251,15 +251,39 @@ body {
...
                     }
+
+                    // Header with 2 or more parts
+                    @if ((o-website-value('header-template') == 'search')
+                        or (o-website-value('header-template') == 'sales_one')
+                        or (o-website-value('header-template') == 'sales_two')
+                        or (o-website-value('header-template') == 'sales_three')
+                        or (o-website-value('header-template') == 'sales_four')
+                    ) {
+                        #o_main_nav > div, #o_main_nav > div > .container {
```
bvr-odoo:
```suggestion
                        #o_main_nav > div {
                            &, & > .container {
```

bvr-odoo: Can't we rely on the `o_header_hide_on_scroll` class (or other classes) here somewhere?
That would be more robust than targeting a `div`.

addons/website/static/src/scss/website.scss
```diff
@@ -251,15 +251,39 @@ body {
...
+            $-menu-box-shadow: o-website-value('menu-box-shadow');
+            @if (
+                (not $-menu-color or ($-menu-color == o-color('body') and o-website-value('body-image') == null))
+                and ($-menu-box-shadow == none)
+                and (o-website-value('header-template') != 'boxed')
+                and (o-website-value('header-template') != 'sidebar')
+            ) {
                 > header {
-                    .navbar, .container {
+                    .navbar, #o_main_nav {
```
bvr-odoo: I did not look into it in detail to decide whether the rule should be on `navbar`, `o_main_nav`, or both.
I tested it a bit and did not see any issue 👍🏼

addons/website/static/src/scss/website.scss
bvr-odoo: There is still something wrong with the "2 parts" headers look:

Image

With this kind of header using two colors, you cannot rely on only one of them to decide whether it needs padding or not. But if we also had to check the color of the other part of the header, it would become complicated, so I would leave it as is, unless you find a simple way to fix it 😉

bvr-odoo (changes requested):
🐈 Some remarks
Don't forget to store `(o-website-value('header-template')`

addons/website/static/src/scss/website.scss
romo-odoo: `#o_main_nav` seems to be enough when I tested it.
I let you double check (it would simplify the way the code is written)

bvr-odoo (approved):
ᯓ➤ LGTM

PR #244415, opened 2026-01-19
[FIX] website: restrict header border width to single value

romo-odoo:
Previously, the border width input for headers allowed multiple values. This was required for specific header templates (e.g. rounded box) that use a full border.

However, most headers only apply a border on the bottom. When the input had multiple values, the scss would break, resulting in full border. This change ensures that, for headers without the .o_full_border class, only the first value of the saved border width is used.

As a result, the input behaves like a single-value field (similar to font size inputs) for standard headers, while still supporting multiple values for templates that require a full border.

Steps to reproduce the issue:
- Go to Edit mode
- Click on the Header
- In the Border option, enter "1 2" and leave the input to validate
=> The input display "3" and the header has a full border.

task-5500516

### DIFF
--- a/addons/website/static/src/scss/website.scss
+++ b/addons/website/static/src/scss/website.scss
@@ -1142,15 +1142,20 @@ header {
     }
 
     nav.navbar {
-        @if o-website-value('menu-border-width') {
+        $-menu-border-width: o-website-value('menu-border-width');
+        @if $-menu-border-width {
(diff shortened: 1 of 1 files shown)


bvr-odoo (approved):
@romo-odoo LGTM, could you rebase on 17.0 + mark the PR as ready and then ping me
Thanks 👍🏼

PR #244416, opened 2026-01-19
[FIX] base: prevent cron run without valid access

GaetanVandenBergh:
Any user can be used to run a server action, even if archived or without corresponding access right.

Steps:
- Create a scheduled action running with an administrator (code: empty)
- Archive the used administrator
- Do the same steps with random user

Actual result:
- Action continue to run with an archived user
- Action continue to run with a user without corresponding access right

Expected result:
- Action should not run on a archived user (except System)
- Action can run if user has corresponding access right

opw-5475805

### DIFF
--- a/odoo/addons/base/models/ir_cron.py
+++ b/odoo/addons/base/models/ir_cron.py
@@ -445,6 +445,11 @@ class IrCron(models.Model):
             loop_count = 0
             _logger.info('Job %r (%s) starting', job['cron_name'], job['id'])
 
+            if not env.user.active and env.user != env.ref('base.user_root'):
+                _logger.warning("Forbidden server action %r executed while the user %s is archived.", job['cron_name'], env.user.login)
+                done, remaining = 0, 0
+                status = CompletionStatus.FAILED
+
             # stop after MIN_RUNS_PER_JOB runs and MIN_TIME_PER_JOB seconds, or
             # upon full completion or failure
             while status is None and (
--- a/odoo/addons/base/tests/test_ir_cron.py
+++ b/odoo/addons/base/tests/test_ir_cron.py
@@ -13,7 +13,8 @@ from unittest.mock import patch
 from freezegun import freeze_time
 
 from odoo import fields
-from odoo.tests.common import RecordCapturer, TransactionCase
+from odoo.addons.base.tests.common import TransactionCaseWithUserDemo
+from odoo.tests.common import Like, RecordCapturer, TransactionCase
 from odoo.tools import mute_logger
 
 from odoo.addons.base.models.ir_cron import (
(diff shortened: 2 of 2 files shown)


odoo/addons/base/models/ir_cron.py
```diff
@@ -445,6 +445,19 @@ def _run_job(cls, job) -> CompletionStatus:
             loop_count = 0
             _logger.info('Job %r (%s) starting', job['cron_name'], job['id'])
 
+            if not env.user.active and env.user != env.ref('base.user_root'):
+                _logger.warning("Forbidden server action %r executed while the user %s is archived.", cron.name, env.user.login)
+                status = CompletionStatus.FAILED
+            else:
+                model_name = cron.model_id.model
+                try:
+                    env[model_name].check_access_rights("write")
```
kmagusiak: Why is this needed?
What errors do you get?

odoo/addons/base/models/ir_cron.py
```diff
@@ -445,6 +445,19 @@ def _run_job(cls, job) -> CompletionStatus:
             loop_count = 0
             _logger.info('Job %r (%s) starting', job['cron_name'], job['id'])
 
+            if not env.user.active and env.user != env.ref('base.user_root'):
+                _logger.warning("Forbidden server action %r executed while the user %s is archived.", cron.name, env.user.login)
+                status = CompletionStatus.FAILED
```
kmagusiak, line 451: You have updated the status here, but the server action is still called.

kmagusiak (changes requested):

odoo/addons/base/models/ir_cron.py
GaetanVandenBergh: Action not called if the status is used in the loop condition, right?
```
while status is None and (
...
```

odoo/addons/base/models/ir_cron.py
GaetanVandenBergh: No error, the action is run without any error even if the user doesn't have access to the correspondind model. If it's expected let me know
Move to draft for now, checking to add test

odoo/addons/base/models/ir_cron.py
```diff
@@ -445,6 +445,20 @@ def _run_job(cls, job) -> CompletionStatus:
             loop_count = 0
             _logger.info('Job %r (%s) starting', job['cron_name'], job['id'])
 
+            if not env.user.active and env.user != env.ref('base.user_root'):
+                _logger.warning("Forbidden server action %r executed while the user %s is archived.", job['cron_name'], env.user.login)
+                done, remaining = 0, 0
+                status = CompletionStatus.FAILED
+            elif model_name := job.get('model_name'):
+                try:
+                    env[model_name].check_access("write")
```
kmagusiak: Why do we need this? What issue does it solve? Do you have an example of a stack-trace maybe?
The job will just crash if the user does not have the right permissions, won't it?

GaetanVandenBergh: If the code to execute pass, the job will succeed, without checking the model field. 
So maybe too much check for this PR, but notice that the test `test_cron_no_access_user` doesn't pass without that, so that means there is not access check for the "running" model?

I will remove this `elif` and the corresponding test from this PR, maybe something to check more in depth on dev side

kmagusiak (approved):
@Julien00859 Can you double-check?

Julien00859 (approved):
I feared it added a SQL query, but it does not!
So it is (a bit) safer and doesn't cost us anything! :D

@robodoo r+

PR #244417, opened 2026-01-19
[FIX] website: preserve urls when removing or reordering gallery images

malh-odoo:
Before this change, using the website editor with image galleries, removing or reordering an image caused all image links to be lost. This happened because the gallery was rebuilt using image nodes only, dropping anchor wrappers during the process.

### How to reproduce:
* Open the website editor.
* Add an Image Gallery block.
* Add links to some of the images.
* Remove or reorder an image.
* All image links are lost.

### Solution:
Pass image holders instead of raw images to setImages and handle the different gallery modes.
opw-5422081

### DIFF
--- a/addons/website/static/src/builder/plugins/options/gallery_element_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/gallery_element_option_plugin.js
@@ -43,6 +43,7 @@ export class SetGalleryElementPositionAction extends BuilderAction {
             : "GalleryImageList";
 
         // Get the items to reorder.
+        activeItemEl = activeItemEl.closest("a") || activeItemEl;
         const itemEls = [];
         for (const getGalleryItems of this.getResource("get_gallery_items_handlers")) {
             itemEls.push(...getGalleryItems(activeItemEl, optionName));
--- a/addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
@@ -116,7 +116,7 @@ class ImageGalleryOption extends Plugin {
         if (optionName === "GalleryImageList") {
             const galleryEl = activeItemEl.closest(".s_image_gallery");
             const containerEl = this.getContainer(galleryEl);
-            itemEls = this.getImages(containerEl);
+            itemEls = this.getImageHolder(containerEl);
         }
         return itemEls;
     }
@@ -133,7 +133,10 @@ class ImageGalleryOption extends Plugin {
             const galleryEl = activeItemEl.closest(".s_image_gallery");
 
             // Update the content with the new order.
-            itemEls.forEach((img, i) => (img.dataset.index = i));
+            itemEls.forEach((itemEl, i) => {
+                const imgEl = this.getImageElement(itemEl);
+                imgEl.dataset.index = i;
+            });
             const mode = this.getMode(galleryEl);
             this.setImages(galleryEl, mode, itemEls);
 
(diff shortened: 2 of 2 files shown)


addons/website/static/src/builder/plugins/options/gallery_element_option_plugin.js
```diff
@@ -43,6 +43,7 @@ export class SetGalleryElementPositionAction extends BuilderAction {
             : "GalleryImageList";
 
         // Get the items to reorder.
+        activeItemEl = activeItemEl.closest("a") || activeItemEl
```
robinlej: Can you activate your linter? (Under 19, just to see what is flagged, not to auto-apply the recommandations. Starting in 19, to enforce it in website and html_builder.)

Here, it's missing a `;`

addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
```diff
@@ -127,7 +127,10 @@ class ImageGalleryOption extends Plugin {
             const galleryEl = activeItemEl.closest(".s_image_gallery");
 
             // Update the content with the new order.
-            itemEls.forEach((img, i) => (img.dataset.index = i));
+            itemEls.forEach( (imgElement, i) => {
```
robinlej: No space between the 2 `((`

addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
```diff
@@ -127,7 +127,10 @@ class ImageGalleryOption extends Plugin {
...
+                const img = imgElement.tagName === 'IMG' ? imgElement : imgElement.querySelector('img');
```
robinlej: - You do that 3 times, it might be worth having a dedicated method
- Use double quotes instead of single quotes
- It's not obvious in this file (😬), but the convention for new variables in website-related modules is to name variables with elements `...El` (or `...Els` for an array-like) (I won't repeat it, but same comment for the rest)

addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
```diff
@@ -244,30 +247,39 @@ class ImageGalleryOption extends Plugin {
...
 
-    slideshow(imageGalleryElement, images) {
+    slideshow(imageGalleryElement, imageElements) {
         const container = this.getContainer(imageGalleryElement);
         const currentInterval = imageGalleryElement.querySelector(".carousel")?.dataset.bsInterval;
         const carouselEl = imageGalleryElement.querySelector(".carousel");
         const colorContrast =
             carouselEl && carouselEl.classList.contains("carousel-dark") ? "carousel-dark" : " ";
+
+        const imageData = imageElements.map(element => {
```
robinlej: - I'd call it `imagesData` (I know  "data" is already a plural, but I wouldn't bet it's obvious for everybody)
- Linting: `((el) => {`
Same elsewhere

addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
```diff
@@ -283,6 +295,13 @@ class ImageGalleryOption extends Plugin {
         container.replaceChildren(slideshowEl);
         slideshowEl.querySelectorAll("img").forEach((img, index) => {
             img.setAttribute("data-index", index);
+            const linkUrl = imageData[index]?.link;
+            if (linkUrl) {
```
robinlej: just do that in the `if` directly

addons/website/static/src/builder/plugins/options/image_gallery_option_plugin.js
```diff
@@ -283,6 +295,13 @@ class ImageGalleryOption extends Plugin {
...
+                const linkEl = document.createElement("a");
```
robinlej: Instead of recreating the link, I think you should store the old link and reuse it.
If you check the code before/after in the devtools, you'll see there is a class `o_translate_inline` on the link before, but not after this.

robinlej (changes requested):
Thanks for the PR!
I have one real issue, the rest are nitpick comments (but on half the lines, and it's to make forward-ports and future dev easier)

PR #244419, opened 2026-01-19
Allow image generation using AI from media dialog

abdelrahmankhaledg:
[IMP] *: allow AI image generation on media dialog

*: html_editor, html_builder, mail, product, web_splash, website,
website_sale

- Extract the MediaDialog save logic into media_dialog_utils so that it can be reused from 'ai_chat' channels to save generated images.

- Extract the X2ManyImageField and X2ManyMediaViewer save logic into image_library to be reused from 'ai_chat' channels to save generated images.

- Create a new ImageFieldWithMediaDialog to be used with Many2One fields and use it with product.template model. When editing the image of a product, the media manager will appear instead of the file browser. This will give users the ability to select from images already stored in the media manager, upload images, use URLs to retrieve images or generate images using AI.

- Some plugin actions open the media dialog in the load function and sets the save function of the media dialog to a dummy save function that only assigns the selected images from the media dialog to a variable and the actual save logic is done in the apply method (which takes the assigned variable of the load function as an input). Given that the save logic is passed from the media dialog to the 'ai_chat' channel (which is a dummy function in this case), it is not possible to save AI generated images. So, these actions are refactored such that the load method is removed and the media dialog is opened in the apply method with the full save logic.

SEE https://github.com/odoo/enterprise/pull/104712
task-5153846

### DIFF
--- a/addons/html_editor/models/ir_attachment.py
+++ b/addons/html_editor/models/ir_attachment.py
@@ -82,7 +82,15 @@ class IrAttachment(models.Model):
     def _get_media_info(self):
         """Return a dict with the values that we need on the media dialog."""
         self.ensure_one()
-        return self._read_format(['id', 'name', 'description', 'mimetype', 'checksum', 'url', 'type', 'res_id', 'res_model', 'public', 'access_token', 'image_src', 'image_width', 'image_height', 'original_id'])[0]
+        return self._read_format(self._editor_media_fields())[0]
+
+    @api.model
+    def _editor_media_fields(self):
+        return [
+            'original_id', 'id', 'name', 'description', 'mimetype',
+            'checksum', 'url', 'type', 'res_id', 'res_model',
+            'public', 'access_token', 'image_src', 'image_width', 'image_height',
+        ]
 
     def _can_bypass_rights_on_media_dialog(self, **attachment_data):
         """ This method is meant to be overridden, for instance to allow to
--- a/addons/html_editor/static/src/core/dialog_plugin.js
+++ b/addons/html_editor/static/src/core/dialog_plugin.js
@@ -19,14 +19,14 @@ export class DialogPlugin extends Plugin {
      * @param {Component} DialogClass
      * @param {Object} props
      * @param {DialogServiceInterfaceAddOptions} options
-     * @returns {Promise<void>}
+     * @returns {Promise<any>}
      */
     addDialog(DialogClass, props, options = {}) {
         return new Promise((resolve) => {
             this.services.dialog.add(DialogClass, props, {
-                onClose: () => {
+                onClose: (closeParams) => {
                     this.dependencies.selection.focusEditable();
-                    resolve();
+                    resolve(closeParams);
                 },
                 ...options,
             });
--- a/addons/html_editor/static/src/fields/image_field_with_media_dialog/image_field_with_media_dialog.js
+++ b/addons/html_editor/static/src/fields/image_field_with_media_dialog/image_field_with_media_dialog.js
@@ -1,0 +1,50 @@
+import { registry } from "@web/core/registry";
+import { useService } from "@web/core/utils/hooks";
+import { ImageField, imageField } from "@web/views/fields/image/image_field";
+import { saveSingleAttachment } from "@web/core/utils/image_library";
+import { CustomMediaDialog } from "@html_editor/fields/x2many_field/custom_media_dialog";
+
+export class ImageFieldWithMediaDialog extends ImageField {
+    static template = "html_editor.ImageFieldWithMediaDialog";
+
+    setup() {
+        super.setup();
+        this.orm = useService("orm");
+        this.dialog = useService("dialog");
+    }
+
+    onFileEdit(ev) {
+        this.dialog.add(CustomMediaDialog, this.mediaDialogProps);
+    }
+
+    get mediaDialogProps() {
+        return {
+            visibleTabs: ["IMAGES"],
+            activeTab: "IMAGES",
+            save: (el) => {}, // Simple rebound to fake its execution
+            imageSave: this.onImageSave.bind(this),
+            imageUrl: this.imageUrl,
+        };
+    }
+
+    get imageUrl() {
+        const fieldName = this.fieldType === "many2one" ? this.props.previewImage : this.props.name;
+        return this.getUrl(fieldName);
+    }
+
+    async onImageSave(attachments) {
+        await saveSingleAttachment(this.env, {
+            attachment: attachments[0],
+            targetRecord: this.props.record,
+            targetFieldName: this.props.name,
+            changeRecordName: false,
+        });
+    }
+}
+
+export const imageFieldWithMediaDialog = {
+    ...imageField,
+    component: ImageFieldWithMediaDialog,
+};
+
+registry.category("fields").add("image_with_media_dialog", imageFieldWithMediaDialog);
--- a/addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
+++ b/addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
@@ -15,22 +15,26 @@ export class CustomMediaDialog extends MediaDialog {
             return;
         }
         if (this.state.activeTab == "IMAGES") {
-            const attachments = this.selectedMedia[this.state.activeTab];
-            const preloadedAttachments = attachments.filter((attachment) => attachment.res_model);
-            this.selectedMedia[this.state.activeTab] = attachments.filter(
-                (attachment) => !preloadedAttachments.includes(attachment)
-            );
-            if (this.selectedMedia[this.state.activeTab].length > 0) {
-                await super.save();
-                const newAttachments = this.selectedMedia[this.state.activeTab];
-                this.props.imageSave(newAttachments);
-            }
-            if (preloadedAttachments.length) {
-                this.props.imageSave(preloadedAttachments);
-            }
+            await this.imageSave({
+                attachments: this.selectedMedia[this.state.activeTab],
+                superSaveFunction: () => super.save(),
+                propsSaveFunction: (attachments) => this.props.imageSave(attachments),
+            });
         } else {
             this.props.videoSave(this.selectedMedia[this.state.activeTab]);
         }
         this.props.close();
+    }
+
+    async imageSave({ attachments, superSaveFunction, propsSaveFunction }) {
+        const preloadedAttachments = attachments.filter((attachment) => attachment.res_model);
+        const nonPreloadedAttachments = attachments.filter((attachment) => !attachment.res_model);
+        if (nonPreloadedAttachments.length > 0) {
+            await superSaveFunction();
+            await propsSaveFunction(nonPreloadedAttachments);
+        }
+        if (preloadedAttachments.length) {
+            await propsSaveFunction(preloadedAttachments);
+        }
     }
 }
--- a/addons/html_editor/static/src/fields/x2many_field/x2many_image_field.js
+++ b/addons/html_editor/static/src/fields/x2many_field/x2many_image_field.js
@@ -3,6 +3,7 @@ import { useService } from "@web/core/utils/hooks";
 import { ImageField, imageField } from "@web/views/fields/image/image_field";
 import { CustomMediaDialog } from "./custom_media_dialog";
 import { getVideoUrl } from "@html_editor/utils/url";
+import { saveSingleAttachment } from "@web/core/utils/image_library";
 
 export class X2ManyImageField extends ImageField {
     static template = "html_editor.ImageField";
@@ -17,42 +18,31 @@ export class X2ManyImageField extends ImageField {
      * standard behavior of opening file input box in order to update a record.
      */
     onFileEdit(ev) {
+        this.dialog.add(CustomMediaDialog, this.mediaDialogProps);
+    }
+
+    get mediaDialogProps() {
         const isVideo = this.props.record.data.video_url;
         let mediaEl;
         if (isVideo) {
             mediaEl = document.createElement("img");
             mediaEl.dataset.src = this.props.record.data.video_url;
         }
-        this.dialog.add(CustomMediaDialog, {
+        return {
             visibleTabs: ["IMAGES", "VIDEOS"],
             media: mediaEl,
             activeTab: isVideo ? "VIDEOS" : "IMAGES",
             save: (el) => {}, // Simple rebound to fake its execution
             imageSave: this.onImageSave.bind(this),
             videoSave: this.onVideoSave.bind(this),
-        });
+        };
     }
-
-    async onImageSave(attachment) {
-        const attachmentRecord = await this.orm.searchRead(
-            "ir.attachment",
-            [["id", "=", attachment[0].id]],
-            ["id", "raw", "name"],
-            {}
-        );
-        if (!attachmentRecord[0].raw) {
-            // URL type attachments are mostly demo records which don't have any ir.attachment raw
-            // TODO: make it work with URL type attachments
-            return this.notification.add(
-                `Cannot add URL type attachment "${attachmentRecord[0].name}". Please try to reupload this image.`,
-                {
-                    type: "warning",
-                }
-            );
-        }
-        await this.props.record.update({
-            [this.props.name]: attachmentRecord[0].raw,
-            name: attachmentRecord[0].name,
+    async onImageSave(attachments) {
+        await saveSingleAttachment(this.env, {
+            attachment: attachments[0],
+            targetRecord: this.props.record,
+            targetFieldName: this.props.name,
+            changeRecordName: true,
         });
     }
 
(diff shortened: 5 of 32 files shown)


addons/product/views/product_views.xml
```diff
@@ -31,7 +31,7 @@
         <field name="model">product.template</field>
         <field name="mode">primary</field>
         <field name="arch" type="xml">
-            <form string="Product">
+            <form string="Producton">
```
Feyensv: 🤔

addons/html_editor/models/ir_attachment.py
```diff
@@ -77,7 +77,34 @@ def _compute_image_size(self):
     def _get_media_info(self):
         """Return a dict with the values that we need on the media dialog."""
         self.ensure_one()
-        return self._read_format(['id', 'name', 'description', 'mimetype', 'checksum', 'url', 'type', 'res_id', 'res_model', 'public', 'access_token', 'image_src', 'image_width', 'image_height', 'original_id'])[0]
+        return self._read_format(self._editor_media_fields())[0]
+
+    @api.model
+    def _editor_media_fields(self, split_relational=False):
```
dmo-odoo: This metho is lacking a verb. Shouldn't it be something like `_get_editor_media_fields`?

FrancoisGe: Who is using `split_relational=True` ?

addons/html_editor/static/src/fields/image_field_with_media_dialog/image_field_with_media_dialog.js
```diff
@@ -0,0 +1,50 @@
...
+        })
+    }
+}
+
+export const imageFieldWithMediaDialog = {
+    ...imageField,
+    component: ImageFieldWithMediaDialog,
+};
+
+registry.category("fields").add("image_with_media_dialog", imageFieldWithMediaDialog);
```
FrancoisGe, line 50: In what situation would you want an image field without the editing feature with AI ?

addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
```diff
@@ -15,19 +16,11 @@ export class CustomMediaDialog extends MediaDialog {
...
-                this.props.imageSave(newAttachments);
-            }
-            if (preloadedAttachments.length) {
-                this.props.imageSave(preloadedAttachments);
-            }
+            await customMediaDialogImageSave({
+                attachments: this.selectedMedia[this.state.activeTab],
+                superSaveFunction: () => super.save(),
+                propsSaveFunction: (attachments) => this.props.imageSave(attachments),
+            });
```
FrancoisGe: Why do you need to extract this function like an utils ? (You use it only one time)

addons/html_editor/static/src/fields/x2many_field/x2many_media_viewer.js
```diff
@@ -42,115 +47,12 @@ export class X2ManyMediaViewer extends X2ManyField {
...
-                }
-                record.update(updateData);
-            });
-        }
+        await saveMultipleAttachments(this.env, {
+            attachments,
+            targetRecord: this.props.record,
+            targetFieldName: this.props.name,
+            convertToWebp: this.props.convertToWebp,
+        });
```
FrancoisGe, line 55: Same question, why do you need to move it in an utils ?

addons/html_editor/static/src/main/media/media_dialog/media_dialog.js
```diff
@@ -318,12 +201,17 @@ export class MediaDialog extends Component {
...
+                orm: this.orm,
+                activeTab: this.state.activeTab,
+                availableTabs: this.tabs,
+                oldMediaNode: this.props.media,
+                selectedMedia: selectedMedia,
+                extraClassesToAdd: this.extraClassesToAdd(),
+                extraClassesToRemove: this.initialIconClasses,
+                multiImages: this.props.multiImages,
+                saveFunction: this.props.save,
+            });
```
FrancoisGe, line 217: same question

addons/html_editor/models/ir_attachment.py
abdelrahmankhaledg: It is used in `_store_attachment_fields` in `ir.attachment` in the `AI` module. It returns the fields that should be synced with the frontend. `split_relational` is needed because relational fields are wrapped inside`Store.FieldList.one` whereas non relational ones use `Store.FieldList.attr`.

addons/html_editor/static/src/fields/image_field_with_media_dialog/image_field_with_media_dialog.js
abdelrahmankhaledg: I can discuss this with the PO. For now, our main use case is Products. Maybe you don't want to use AI with employee's pictures for example?

addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
abdelrahmankhaledg: It is used again in `custom_media_dialog_patch.js` in the `AI` module to give the `ai_chat` channel the ability to execute the same save logic of the CustomMediaDialog when clicking on the `Use This` button which appears under AI generated images inside the chat channel.

addons/html_editor/static/src/fields/x2many_field/x2many_media_viewer.js
abdelrahmankhaledg: It is used a second time in `product_ai_action.js` in `ai_website_sale` module to allow the `Extra Photo` action when opening an `ai_chat` channel to change the image of a product from the `Sales` app.

addons/html_editor/static/src/main/media/media_dialog/media_dialog.js
abdelrahmankhaledg: It is similar to the case of `CustomMediaDialog`. It is used in the `media_dialog_patch.js` in the `AI` module to allow the `ai_chat` channel to execute the save logic of the MediaDialog when clicking on the `Use This` button.

addons/html_editor/models/ir_attachment.py
seb-odoo: It might be easier to simply make 2 separate lists.

If these are used for different purposes, there's probably not much point to have a common list anyway.

seb-odoo: Also note: relational fields for which you don't need any field (`[]` as 2nd param) can simply be returned as `attr` too, so you might not even need to split them at all.

addons/mail/tests/discuss/test_message_controller.py
```diff
@@ -163,44 +147,44 @@ def test_channel_message_attachments(self):
...
 
+    def _get_attachment_store_data(self, attachment):
+        store_data = {
+            "checksum": False,
+            "create_date": fields.Datetime.to_string(attachment.create_date),
+            "file_size": 0,
+            "has_thumbnail": False,
+            "id": attachment.id,
+            "mimetype": "application/octet-stream",
+            "name": attachment.name,
```
seb-odoo: Not hard-coding the values reduce the coverage of the test in case the name becomes a dummy value, so we might not want to do that change.

I'd prefer if you followed the same pattern as we do in the other tests, which is to list all the values initially and filter those that won't be applicable. See `_filter_threads_fields` and other similar methods around that one.

addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
FrancoisGe: Why do you need to extract it in an utils ?
It s not easier to have the function in `CustomMediaDialog` ? 
The actual code looks strange for me ^^

abdelrahmankhaledg: Yes you are correct but I have one concern. Currently when the AI chat is opened, the media dialog is closed. So, any usage of `this` (which refers to the instance of the media dialog that is closed) inside the function may / may not be suitable for the AI. Extracting the function as a util and passing everything as a parameter will force developers to do the same when introducing new logic. What do you think?

addons/website_sale/static/src/website_builder/product_page_option_plugin.js
```diff
@@ -317,15 +317,19 @@ export class ProductReplaceMainImageAction extends BaseProductPageAction {
         this.reload = false;
         this.canTimeout = false;
     }
-    apply({ editingElement: productDetailMainEl }) {
+    apply({ editingElement }) {
+        this.dependencies.media.openMediaDialog(this.getMediaDialogProps({ editingElement }));
```
caburj:
```suggestion
        return this.dependencies.media.openMediaDialog(this.getMediaDialogProps({ editingElement }));
```

or maybe declare "apply" as async and await the openMediaDialog call.

addons/mail/tests/discuss/test_message_controller.py
```diff
@@ -97,26 +97,32 @@ def test_channel_message_attachments(self):
...
+                'url': False,
+                "access_token": False,
+                "description": False,
+                "image_src": False,
+                "image_height": 0,
+                "image_width": 0,
+                "original_id": False,
+                "public": False,
+                "res_id": self.attachments[0].res_id,
+            }),
```
seb-odoo: Please no extra diff, keep the syntax as it is, just wrap the `_filter_attachments_fields` around it.

addons/mail/tests/discuss/test_message_controller.py
```diff
@@ -163,44 +169,58 @@ def test_channel_message_attachments(self):
...
+                "file_size": 0,
+                "has_thumbnail": False,
+                "id": self.attachments[0].id,
+                "mimetype": "application/octet-stream",
+                "name": "File 1",
+                "ownership_token": self.attachments[0]._get_ownership_token(),
+                "raw_access_token": self.attachments[0]._get_raw_access_token(),
+                "res_name": "Test channel",
+                "res_model": self.attachments[0].res_model,
+                "thread": {"id": self.channel.id, "model": "discuss.channel"},
```
seb-odoo: same remark everywhere

odoo/addons/base/tests/common.py
```diff
@@ -204,6 +204,22 @@ def setUpClass(cls):
                 'group_ids': [Command.set([cls.env.ref('base.group_user').id, cls.env.ref('base.group_partner_manager').id])],
             })
 
+    def _filter_attachments_fields(self, /, *attachments_data):
```
seb-odoo: This should normally be in `MailCommon` with other similar functions.

We're not making use of this code in `base`, and picking `HttpCaseWithUserDemo` of all the possible test classes is oddly specific.

addons/html_editor/models/ir_attachment.py
```diff
@@ -82,7 +82,15 @@ def _compute_image_size(self):
     def _get_media_info(self):
         """Return a dict with the values that we need on the media dialog."""
         self.ensure_one()
-        return self._read_format(['id', 'name', 'description', 'mimetype', 'checksum', 'url', 'type', 'res_id', 'res_model', 'public', 'access_token', 'image_src', 'image_width', 'image_height', 'original_id'])[0]
+        return self._read_format(self._editor_media_fields())[0]
+
+    @api.model
+    def _editor_media_fields(self):
```
dmo-odoo, line 88: Weird name. No verb (get?) and prefixed by `editor` for some reason. It used to be in `_get_media_info` so why not `_get_media_fields`?

addons/html_editor/static/src/core/dialog_plugin.js
```diff
@@ -19,14 +19,14 @@ export class DialogPlugin extends Plugin {
...
+     * @returns {Promise<any>}
      */
     addDialog(DialogClass, props, options = {}) {
         return new Promise((resolve) => {
             this.services.dialog.add(DialogClass, props, {
-                onClose: () => {
+                onClose: (closeParams) => {
                     this.dependencies.selection.focusEditable();
-                    resolve();
+                    resolve(closeParams);
```
dmo-odoo, line 29: I don't like this change because it feels like a hack. Prior to this PR, the only possible cases for closing where either saving or discarding. Therefore, it could have been `onSave` and `onDiscard` instead if it was needed to differentiate between both. But you introduced a _new_ one, which complicates things, and I'm not sure why.

EDIT after our discussion: I don't think the media dialog needs to know it was closed by AI etc. The AI chat can insert the generated picture as base64 with the `o_b64_image_to_save` class, just like the AI chat can already insert text content and that's it.

addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
```diff
@@ -15,22 +15,26 @@ export class CustomMediaDialog extends MediaDialog {
...
         } else {
             this.props.videoSave(this.selectedMedia[this.state.activeTab]);
         }
         this.props.close();
     }
+
+    async imageSave({ attachments, superSaveFunction, propsSaveFunction }) {
```
dmo-odoo, line 29: Weird function signature. Weird name, should be `saveImage`, and weird parameters.

addons/html_editor/static/src/fields/x2many_field/x2many_image_field.js
```diff
@@ -17,42 +18,31 @@ export class X2ManyImageField extends ImageField {
...
-                {
-                    type: "warning",
-                }
-            );
-        }
-        await this.props.record.update({
-            [this.props.name]: attachmentRecord[0].raw,
-            name: attachmentRecord[0].name,
+    async onImageSave(attachments) {
+        await saveSingleAttachment(this.env, {
```
dmo-odoo, line 41: I don't think it's useful to have both `saveSingleAttachment` and `saveMultipleAttachments`. Surely one can put a single element inside an array and the other can be a loop over one element, right?

addons/html_editor/static/src/main/media/media_dialog/media_dialog_utils.js
```diff
@@ -0,0 +1,167 @@
...
+        extraClassesToAdd,
+        extraClassesToRemove,
+        aiChannelId,
+    });
+    if (multiImages) {
+        await saveFunction(elements, selectedMedia, activeTab, oldMediaNode);
+    } else {
+        await saveFunction(elements[0], selectedMedia, activeTab, oldMediaNode);
+    }
+}
```
dmo-odoo, line 48: Weird function, feels ad-hoc. I think what should have been factorized is the save function to be able to be called with multiple images or not and then this function would have only been two function calls, which would then have been obvious that it is not very useful, so every time it's use (two times total), you could just write:
```js
const elements = await renderMedia(...);
const saveMedia(elements, ...);
```

addons/html_editor/static/src/main/media/media_dialog/media_dialog_utils.js
```diff
@@ -0,0 +1,167 @@
...
+                }
+                if (oldMediaNode.dataset.hoverEffectIntensity) {
+                    element.dataset.hoverEffectIntensity =
+                        oldMediaNode.dataset.hoverEffectIntensity;
+                }
+            }
+        }
+        if (aiChannelId) {
+            element.dataset.aiChannelId = aiChannelId;
+        }
```
dmo-odoo, line 112: Wait what? This is the html_editor module, not the ai module. I caught this line by chance. The function was modified and moved at the same time so it's very hard to see what has been changed.

addons/html_editor/static/src/main/media/media_dialog/media_dialog_utils.js
```diff
@@ -0,0 +1,167 @@
...
+    return elements;
+}
+
+export function convertAttachmentRecordToObject(attachment_record) {
+    const imageAttachmentObject = {};
+    for (const attachment_field of ATTACHMENT_FIELDS) {
+        imageAttachmentObject[attachment_field] = attachment_record[attachment_field];
+    }
+    return imageAttachmentObject;
+}
```
dmo-odoo, line 167: This function is never used in community and I don't think it should be here. It looks more like a local util than something important for html_editor to export.

addons/html_editor/static/src/main/media/media_plugin.js
```diff
@@ -254,6 +259,14 @@ export class MediaPlugin extends Plugin {
         return mediaDialogClosedPromise;
     }
 
+    getSaveFunction(params) {
+        return (...args) => {
+            const element = args[0];
+            const node = args[3] || params.node;
+            this.onSaveMediaDialog(element, { node });
+        };
+    }
```
dmo-odoo, line 268: This is a weird function to have. Why is it needed? It's not overriden anywhere, even in enterprise.

addons/html_editor/static/src/fields/x2many_field/custom_media_dialog.js
abdelrahmankhaledg: I've removed the params superSaveFunction and propsSaveFunction.
I will keep the name as `imageSave` to avoid unnecessary diff because the same name is used inside `x2many_image_field`, `x2many_media_viewer` and `image_field_with_media_dialog` which are the only use cases of the custom media dialog.

addons/html_editor/static/src/fields/x2many_field/x2many_image_field.js
abdelrahmankhaledg: Saving multiple attachments is used with x2many fields which makes the save logic a bit different.

PR #244420, opened 2026-01-19
[IMP] orm: enforce company checks on unarchive

UemuS:
Summary:

  - Ensure _check_company runs on unarchive and related inverse checks are executed consistently.
  - Cache inverse check_company fields in the registry (field_company_inverses) and use that map in _check_company_on_unarchive_inverses to avoid scanning every model per call.
  - Add test_orm models and access entries needed to exercise inverse check_company links.
  - Add tests covering unarchive and write+unarchive scenarios.
  - Reduce log noise by logging the company check warning once.

Why:

  - Users routinely correct archived records in multiple steps, some of which are transiently inconsistent by design (change company_id, then reassign parent_id, then fix inverse refs, etc.). Blocking those writes would break the
  workflow, so the check needs to fire at the point where the inconsistency actually causes harm: unarchive.
  - Cross-company inconsistencies can also hide on the inverse side (archived records pointing at ours in mid-fix state), so the sweep uses active_test=False to catch them too.
  - The inverse field map is computed once per registry and reused, keeping the check focused on relevant models/fields.
  - Added tests guard against regressions for unarchive flows.

### DIFF
--- a/odoo/addons/test_orm/tests/test_company_checks.py
+++ b/odoo/addons/test_orm/tests/test_company_checks.py
@@ -163,6 +163,23 @@ class TestCompanyCheck(common.TransactionCase):
         with self.assertRaises(UserError):
             child.action_unarchive()
 
+    def test_company_check_on_write_unarchive_inverses(self):
+        child = self.env['test_orm.model_child'].create({
+            'name': 'M1',
+            'company_id': self.company_a.id,
+        })
+        self.env['test_orm.model_child_ref'].create({
+            'name': 'M2',
+            'company_id': self.company_a.id,
+            'child_id': child.id,
+        })
+
+        child.action_archive()
+        child.company_id = self.company_b.id
+
+        with self.assertRaises(UserError):
+            child.write({'active': True})
+
     def test_check_company_write_performance(self):
         child = self.env['test_orm.model_child'].create({
             'name': 'M1',


odoo/orm/models.py
```diff
@@ -3675,6 +3701,11 @@ def write(self, vals: ValuesType) -> typing.Literal[True]:
         if not self:
             return True
 
+        active_name = self._active_name
+        inactive_before = None
+        if active_name and active_name in vals and vals[active_name]:
```
kmagusiak: Nitpick
```suggestion
        if active_name and vals.get(active_name):
```

odoo/orm/models.py
```diff
@@ -3822,6 +3853,15 @@ def write(self, vals: ValuesType) -> typing.Literal[True]:
 
         if self._check_company_auto:
             self._check_company(list(vals))
+        if inactive_before:
+            to_unarchive = inactive_before.filtered(lambda record: record[active_name])
+            if 'company_id' in self:
+                to_unarchive._validate_fields(['company_id'])
```
kmagusiak: You could validate the field even before the existing `_check_company_auto`.

odoo/orm/registry.py
```diff
@@ -521,6 +521,32 @@ def field_inverses(self) -> Collector[Field, Field]:
...
         return result
 
+    @functools.cached_property
+    def check_company_inverses(self) -> dict[str, dict[str, tuple[str, ...]]]:
+        """Map comodel names to models and their check_company field names."""
+        result: defaultdict[str, dict[str, tuple[str, ...]]] = defaultdict(dict)
+        for model_name, Model in self.models.items():
+            if Model._abstract:
+                continue
+            if not (
```
kmagusiak:
```suggestion
            if Model._abstract or not (
```

odoo/orm/registry.py
```diff
@@ -521,6 +521,32 @@ def field_inverses(self) -> Collector[Field, Field]:
                     field.setup_inverses(self, result)
         return result
 
+    @functools.cached_property
+    def check_company_inverses(self) -> dict[str, dict[str, tuple[str, ...]]]:
```
kmagusiak: This does not check anything. Can you rename it "field_company_inverses"?

odoo/orm/models.py
```diff
@@ -3344,6 +3352,24 @@ def _check_company(self, fnames=None):
...
+        ids = self.ids
+        env = self.env
+        inverse_map = env.registry.check_company_inverses.get(self._name)
+        if not inverse_map:
+            return
+        for model_name, field_names in inverse_map.items():
+            if len(field_names) == 1:
+                domain = Domain(field_names[0], 'in', ids)
+            else:
+                domain = Domain.OR(Domain(name, 'in', ids) for name in field_names)
```
kmagusiak: OR works fine if you have 1 (or even 0) field_name.
```suggestion
            domain = Domain.OR(Domain(name, 'in', ids) for name in field_names)
```

odoo/orm/models.py
```diff
@@ -3344,6 +3352,24 @@ def _check_company(self, fnames=None):
...
+            records = env[model_name].sudo().with_context(active_test=False).search(domain)
```
kmagusiak: Why do you need active test here?
Maybe we should check the consistency for inactive records as well? For example

```py
child.action_archive()
child.company_id = company2.id  # this should already raise
# otherwise
parent.with_context(active_test=False).children_ids.company_id  # will have 2 companies
```

odoo/orm/models.py
```diff
@@ -3284,6 +3284,19 @@ def _check_company(self, fnames=None):
         if not (regular_fields or property_fields):
             return
 
+        has_company_field = (
+            self._name == 'res.company'
+            or 'company_id' in self
+            or 'company_ids' in self
+        )
+        if regular_fields and not has_company_field:
```
aj-fuentes: We could prepare the way we get companies, instead of checking all this for every `record`.
```suggestion
        get_companies = None
        if self._name == 'res.company':
            get_companies = lambda rec: rec
        elif 'company_id' in self:
            get_companies = lambda rec: rec.company_id
        elif 'company_ids' in self:
            get_companies = self.company_ids
        else:
            ... the warning
            regular_fields = []
```

odoo/orm/models.py
aj-fuentes: Instead of checking `regular_fields` we could make the code here more homogeneous with what is done for `property_fields`.
```py
                for name in regular_fields:
                    corecords = record.sudo()[name]
                    companies = get_companies(record)
                    if corecords:
                        domain = corecords._check_company_domain(companies)
                        if domain and corecords != corecords.with_context(active_test=False).filtered_domain(domain):
                            inconsistencies.append((record, name, corecords))
```
It will work nicely if `regular_fields` is an empty list.

kmagusiak:
Could we have this done before 20.0?

odoo/orm/models.py
UemuS: In my opinion that would break an acceptable workflow. Users sometimes correct records in multiple steps while they're archived:  For example change company_id, then reassign parent_id, then fix a couple of inverse refs, etc.
If we blocked at write-time, they couldn't make the first fix without the others already being in place.

That workflow is also why the search uses active_test=False: archived inverse records can carry the same in-progress inconsistency, and we need to catch them too at the unarchive gate.

odoo/orm/models.py
```diff
@@ -3361,30 +3361,37 @@ def _check_company(self, fnames=None):
...
-                    ))
-                    continue
-                for name in regular_fields:
-                    corecords = record.sudo()[name]
-                    if corecords:
-                        domain = corecords._check_company_domain(companies)
-                        if domain and corecords != corecords.with_context(active_test=False).filtered_domain(domain):
-                            inconsistencies.append((record, name, corecords))
+            for name in regular_fields:
+                corecords = record.sudo()[name]
```
kmagusiak, line 3390: Shouldn't you ignore non-active records and corecords?

odoo/orm/models.py
```diff
@@ -3925,6 +3952,15 @@ def write(self, vals: ValuesType) -> typing.Literal[True]:
             # validate inversed fields
             real_recs._validate_fields(inverse_fields)
 
+        if inactive_before:
+            to_unarchive = inactive_before.filtered(lambda record: record[active_name])
+            if 'company_id' in self:
+                to_unarchive._validate_fields(['company_id'])
+            if self._check_company_auto and (
```
kmagusiak: What if `self._check_company_auto == False`, but it is True on the comodel?

odoo/orm/models.py
```diff
@@ -3925,6 +3952,15 @@ def write(self, vals: ValuesType) -> typing.Literal[True]:
             # validate inversed fields
             real_recs._validate_fields(inverse_fields)
 
+        if inactive_before:
+            to_unarchive = inactive_before.filtered(lambda record: record[active_name])
```
kmagusiak:
```suggestion
            unarchived = inactive_before.filtered(lambda record: record[active_name])
```

odoo/orm/models.py
kmagusiak: Just to clarify: company checks should ignore archived records or not?

If ignored, then do `active_test=True` since you don't want to check archived records and update `_check_company` to filter out inactive records.
If not ignored, then I don't understand the use case.

> That workflow is also why the search uses active_test=False: archived inverse records can carry the same in-progress inconsistency, and we need to catch them too at the unarchive gate.

This function is called after the write, so I don't understand this. After each unarchive, you will perform company checks, won't you?

odoo/orm/models.py
UemuS: Thanks for the catch! I kept the _check_company_auto guard to match the general check pattern, but unarchive is a different case :/
This check should also run no matter what. Fixed.

UemuS:
Hello @kmagusiak 
Right, each unarchive fires its own check, so preemptively sweeping archived neighbors doesn't add coverage. Dropping active_test=False in the inverse search.

On updating _check_company to filter out inactive records, we weren't on the same page, so I'd like your opinion:

skip the check only when both the record and its corecord are archived:

  - Archived record pointing at an active corecord: still checked, since an active record's assumptions shouldn't be silently broken.
  - Archived record pointing at an archived corecord: skipped, which unlocks the "fix both in multiple steps" workflow that master doesn't currently allow. 
 
Some background on why I want this last part: the PR idea originally came because a user archived a record to correct data, but then hit issues on his DB because no checks ran. This extra bit would also let them archive the linked records to fix inconsistencies together. And run all checks at unarchive.

Would that work, or do we want to keep the PR narrowed to checks on unarchive?

UemuS:
Need your take on a behaviour change:

The unarchive check runs _check_company() even when the model has _check_company_auto=False.

Example: hr.department has check_company=True on parent_id but _check_company_auto=False. Take two unrelated companies A and B (no parent/child or branch link). Today you can create a dept in A with a parent in B (create doesn't enforce), and unarchiving it works. On this branch, unarchiving now raises.

PR #244421, opened 2026-01-19
[IMP] core, *: fields.Binary return BinaryValue

kmagusiak:
A new class `BinaryValue` represents lazily loaded binary objects. It is
a subclass of `collections.abc.Buffer`. We use it as the cache and
record format of Binary fields. It allows to handle them while delaying
loading of the contents in the memory. At the same time, the loaded
contents are in the raw format which is 33% smaller than base64-encoded
binary values we were storing in cache and that encoding does not have
to be done anymore on the ORM side. We encode in base64 only for RPC
communications.

... see commit message for details, here are some examples:

```py
bin = fields.Binary()

# reading
assert isinstance(record.bin, BinaryValue)  # always
if record.bin:  # empty data is falsy
  record.bin.contents  # bytes
  record.bin.decode()  # like bytes.decode()
  record.bin.to_base64()  # base64 str
  with record.bin.open() as f:
    r.read()  # just like any io
  record.bin.mimetype  # helper to get the mimetype
  record.bin.size  # not len, because we don't want to load contents in memory for bool testing

# buffer interface, the following work
bytes(record.bin)
memoryview(record.bin)
io.BytesIO(record.bin)
hashlib.sha256(record.bin)

# writing
record.bin = "AAAA"  # base64 string from RPC
# wrapped explicitely because we don't know if a bytes argument was base64 encoded or not
record.bin = BinaryBytes(b'hello world')
# this still works: lot of usages and we know it is raw
attachment.raw = b'hello'

# bin_size only if you call read() - just like today
record.with_context(bin_size=True).read(['bin'])  # human readable size
record.with_context(bin_size=False).read(['bin'])  # same as record.bin.to_base64()
```

A lot of encoding/decoding of base64 can be removed saving data copy and memory. As a next step, we will remove *datas* field.

odoo/enterprise#104714
odoo/documentation#16093
task-4251301

Next PR to remove `datas`: https://github.com/odoo/odoo/pull/245916

---


odoo/orm/fields_binary.py
```diff
@@ -293,6 +234,8 @@ class Image(Binary):
     def setup(self, model):
         super().setup(model)
         if not model._abstract and not model._log_access:
+            # XXX whyyyyy? 25c6c15a06e859370d86c561068e7246effe0e6e
```
kmagusiak: Weird stuff: why we must have log access for images?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -30,6 +30,7 @@
 
 # This SANDBOX_AUTH is only used for testing purposes, and is shared to all users of the sandbox environment
 SANDBOX_AUTH = {
+    # the security token has a double base64 encoding
```
kmagusiak, line 34: l10n: seems strange, but possibly correct

odoo/orm/fields_binary.py
```diff
@@ -56,141 +52,81 @@ def _description_sortable(self, env):
...
+        if not value:
+            return None
+        if isinstance(value, BinaryValue):
+            return value
+        if isinstance(value, str):
+            # a string may come from RPC, it is base64 encoded
+            decoded_value = base64.b64decode(value, validate=validate)
+            # heuristic to detect if it is probably a base64 and not just text
+            if not all(len(s) > 70 or i == 0 for i, s in enumerate(reversed(value.split()))):
+                _logger.warning("%s: str passed to binary does not look like base64", self)
```
HydrionBurst: maybe also log the `record` to help fix the database when necessary

odoo/tools/binary.py
```diff
@@ -0,0 +1,169 @@
...
+
+class BinaryFile(BinaryValue):
+    """Lazy loaded file."""
+    __slots__ = ('__content', '__file', '__mimetype')
+
+    def __init__(self, file: io.FileIO):
+        # __content is None <=> the file is open
+        self.__content: bytes | None = None
+        self.__file = file
+        assert file.mode == 'rb' and not file.closed and file.seekable(), "file needs to be opened in 'rb' mode"
```
HydrionBurst: Since we lazily close the file, is it possible to reach the maximum file descriptor limit for a process?

What is the best practice for it?

kmagusiak: Instead of `mapped('raw')` on thousands of files iterate and either invalidate or read them.
Although, I have not really found such use cases.

odoo/orm/fields_binary.py
kmagusiak: The record is not necessarily the one on which we assign (for example in create). The record is here just to be able to get the environment. We already log the model.

odoo/tools/binary.py
HydrionBurst: Could you try this?

```py
for attachment in attachments:
    attachment.raw.content
```
when we have the cache miss for the first attachment, I think the ORM will compute the `raw` field for all `attachments` and there might be many not closed `BinaryFile.__file` in the cache

kmagusiak: I have `ulimit` set to unlimited.
Here I open them all and `ls /proc/265138/fd -l | grep filestore | wc -l` gives 198 open files.
`atts = env['ir.attachment'].search([('id', '>', 0)]).mapped('raw')`

The question is whether we want starting to load such files in the memory if we do nothing with them just to close file descriptors. If this is really an issue, we can have a global semaphore for number of lazy files open: if we cannot get the semaphore, we read the file into memory and close the descriptor.

HydrionBurst:
```py
for attachment in attachments:
    content = attachment.raw.content
    ... # any business logic to content
```

It is an example, it is not `do nothing`, we could do anything. The problem is after CacheMiss of the `attachments[0].raw` the batched `IrAttachment._compute_raw` "prefetches" files and fills not closed `BinaryFile` in the cache for all e.g 1000 `attachments`
To avoid that I think we may need to override `_to_prefetch` for binary attachment fields

odoo/addons/base/models/ir_attachment.py
```diff
@@ -60,6 +63,31 @@ def condition_values(model, field_name, domain):
     return None
 
 
+class BinaryValueAttachment(BinaryValue):
+    __slots__ = ('__attachment',)
+
+    def __init__(self, attachment: IrAttachment):
+        assert attachment.env.su and attachment._name == 'ir.attachment' and len(attachment) == 1
+        self.__attachment = attachment
```
HydrionBurst: not against the current code
TBD: we may or may not want
`self.__attachment = attachment.browse(attachment.id)`
to drop `_prefetch_ids`

odoo/orm/fields_binary.py
```diff
@@ -56,141 +52,81 @@ def _description_sortable(self, env):
...
+        if cache_value is not None:
+            assert isinstance(cache_value, BinaryValue), f"{self}: unexpected type {type(cache_value)}"
+            cache_value.size  # check if exists and raise if we have issues
+        return super()._update_cache(records, cache_value, dirty)
 
     def convert_to_record(self, value, record):
-        if isinstance(value, _BINARY):
-            return bytes(value)
-        return False if value is None else value
+        return value or BinaryValue.from_bytes(b'')
```
HydrionBurst: ~~I am thinking if we should `value.content` (and change them to cached_property) to freeze the binary value for `BinaryValueAttachment` when `convert_to_record`~~
Maybe no since we don't want to read the binary in many cases.
But the below code seems wired to me 
```py
data = record.binary_attachment_field
record.binary_attachment_field = b'new_value'
print(data.content) # 'new_value' ? what if the old value is b'' ?

data = record.binary_not_attachment_field
record.binary_not_attachment_field = b'new_value'
print(data.content) # `old_value` ?
```

kmagusiak: Indeed. The ugly case is if you had a binary attachment field that is set, and you update the value, you get the new value. This is to avoid getting the raw value as much as possible.

odoo/addons/base/models/ir_attachment.py
kmagusiak: I'd rather not because if you accessed attachments in batch, you will probably read them in batch as well.
PS: to drop prefetch `attachment.with_prefetch()`.

odoo/tools/binary.py
kmagusiak: Prefetch works for records, not really for fields, here the function is only when reading through the `__get__` method. A semaphore is nicer because it is a global limit on the process and the system is on the process.

odoo/tools/binary.py
```diff
@@ -0,0 +1,169 @@
...
+        :param path: path to the file
+        :param filter_ext: passed to `file_open` function
+        :param reader: if set, use it instead of opening; path must match the name
+        :raise OSError: if the file cannot be opened
+        """
+        if reader is None:
+            reader = file_open(path, 'rb', filter_ext=filter_ext)
+        else:
+            assert reader.name == path, "reader.name must match the path"
+        return BinaryFile(reader)
```
HydrionBurst: I don't check all use cases but maybe I am wrong.

Why `BinaryFile` cannot directly store the file name?
So that we lazily get the file descriptor and close the file after use.
If there is no security issue, can we pass the open function (default=open_file) as the parameter to gain flexibility?

Or for `io.FileIO`, we could leave the `close` to developer.
I think in python using `with open` may seems more inconvenient but it is very pythonic and explicitly prevents the leak.

odoo/tools/binary.py
```diff
@@ -0,0 +1,169 @@
...
+        return os.fstat(self.__file.fileno()).st_size
+
+    @property
+    def local_path(self):
+        return self.__file.name
+
+    def close(self):
+        self.__file.close()
+
+    def __del__(self):
```
HydrionBurst: Logicially correct.
I am just curious if it is necessary 🤔 

I think `FileIO` is only referenced by one `BinaryFile`
then if the `BinaryFile` can be garbage collected, the `FileIO` can also be garbage collected, and the python should close the file automatically.
if the same `FileIO` is also used somewhere else, the unpredictable `closed` makes it hard to use

odoo/orm/fields_binary.py
```diff
@@ -56,141 +52,81 @@ def _description_sortable(self, env):
...
-        except binascii.Error:
-            decoded_value = value
         # Detect if the binary content is an SVG for restricting its upload
-        # only to system users.
-        if validate and decoded_value[:1] == b'<':  # XML tag opening
+        # only to system users. Check plaintext XML tag opening.
+        if validate and value[:1] == b'<':
             # Full mimetype detection
-            if (guess_mimetype(decoded_value).startswith('image/svg') and
+            if ('image/svg' in data.mimetype and
```
Julien00859: why `in` and not `startswith` like before?

odoo/orm/fields_binary.py
Julien00859: Isn't `b64decode(validate=True)` doing the job already?

Julien00859: I had a look at cpython's code, `validate=True` is implemented in C, IMO we should always validate.
```c
        // this code is always executed
        unsigned char v = table_a2b_base64[this_ch];
        if (v >= 64) {
            // strict_mode is validate
            if (strict_mode && !ignorechar(this_ch, ignorechars)) {
                state = get_binascii_state(module);
                if (state) {
                    PyErr_SetString(state->Error, "Only base64 data is allowed");
                }
                goto error_end;
            }
            continue;
        }
```

odoo/orm/fields_binary.py
```diff
@@ -56,141 +52,81 @@ def _description_sortable(self, env):
...
+            # heuristic to detect if it is probably a base64 and not just text
+            if not all(len(s) > 70 or i == 0 for i, s in enumerate(reversed(value.split()))):
+                _logger.warning("%s: str passed to binary does not look like base64", self)
+            return BinaryValue.from_bytes(decoded_value)
+        # Error needed because we used to write base64 encoded data and we
+        # cannot distinguish whether bytes are encoded or not in base64.
+        if isinstance(value, bytes) and (self.related_field or self).name == 'raw':
+            # Exception for the raw field, we know bytes are raw.
+            return BinaryValue.from_bytes(value)
+        raise TypeError(f'{self}: use BinaryValue instead of {value.__class__.__name__}')
```
Julien00859, line 84: nit
```suggestion
        raise TypeError(f'{self}: use BinaryValue instead of {type(value).__name__}')
```

odoo/tools/binary.py
```diff
@@ -0,0 +1,169 @@
...
+import io
+import os
+from collections.abc import Buffer
+
+from .misc import file_open
+
+__all__ = ['BinaryValue']
+
+
+class BinaryValue(Buffer):
```
Julien00859, line 8: add a comment to make it explicit that this class is abstract

odoo/tools/binary.py
Julien00859: I wouldn't do anything just yet regarding the file descriptors

odoo/addons/base/models/avatar_mixin.py
```diff
@@ -61,8 +59,8 @@ def _compute_avatar_128(self):
 
     def _avatar_generate_svg(self):
         initial = html_escape(self[self._avatar_name_field][0].upper())
-        bgcolor = get_random_ui_color_from_seed(self[self._avatar_name_field] + str(self.create_date.timestamp() if self.create_date else ""))
-        return b64encode((
+        bgcolor = get_random_ui_color_from_seed(f"{self[self._avatar_name_field]}{self.create_date.timestamp() if self.create_date else ""}".encode())
```
Julien00859: Don't mix single quote and double quote in f-strings, this is only supported by a quite recent python
```suggestion
        bgcolor = get_random_ui_color_from_seed(f'{self[self._avatar_name_field]}{self.create_date.timestamp() if self.create_date else ""}'.encode())
```

odoo/addons/base/models/ir_actions.py
```diff
@@ -132,6 +132,7 @@ def _get_eval_context(self, action=None):
             'float_compare': float_compare,
             'b64encode': base64.b64encode,
             'b64decode': base64.b64decode,
+            'BinaryValue': BinaryValue,
```
Julien00859: I'm pretty confident we don't want to give access to `BinaryValue.from_file` inside safe-eval.
cc @odoo/rd-security

odoo/addons/base/models/ir_attachment.py
```diff
@@ -253,16 +281,13 @@ def _gc_file_store_unsafe(self):
...
-            for attach in self:
-                attach.datas = human_size(attach.file_size)
-            return
-
         for attach in self:
-            attach.datas = base64.b64encode(attach.raw or b'')
+            value = attach.raw
+            if value:
+                value = BinaryValue.from_bytes(base64.b64encode(value))
+            attach.datas = value
```
Julien00859, line 267: I am uneasy `datas` is an object with a `to_base64()` method, when it actually is base64 already. But this might be gone in the next commit...

odoo/addons/base/models/ir_model.py
```diff
@@ -40,6 +40,7 @@
 
 # base environment for doing a safe_eval
 SAFE_EVAL_BASE = {
+    'BinaryValue': BinaryValue,
```
Julien00859: my above comment about `BinaryValue` and safe-eval. cc @odoo/rd-security

odoo/tools/binary.py
kmagusiak: When we return a BinaryValue, we don't expect devs to close them (these are just values from their point of view).

We could store the file name and open as necessary in most cases. Except ir.attachment has `_file_read` which needs to open there the file. It wouldn't guarantee that the file is not deleted in the meantime, but that might be fine.

Actually, I find it a neat idea to try. Tradeoffs :)

odoo/tools/binary.py
kmagusiak: Not strictly necessary, I just wanted it to be explicit.

odoo/addons/base/models/ir_attachment.py
```diff
@@ -60,6 +63,31 @@ def condition_values(model, field_name, domain):
     return None
 
 
+class BinaryValueAttachment(BinaryValue):
+    __slots__ = ('__attachment',)
+
+    def __init__(self, attachment: IrAttachment):
+        assert attachment.env.su and attachment._name == 'ir.attachment' and len(attachment) == 1
```
Julien00859: I don't like `AssertionError: False`.
```suggestion
        assert attachment.env.su and attachment._name == 'ir.attachment' and len(attachment) == 1, \
            (attachment.env.su, attachment._name, len(attachment))
```

odoo/addons/base/models/ir_attachment.py
```diff
@@ -60,6 +63,31 @@ def condition_values(model, field_name, domain):
     return None
 
 
+class BinaryValueAttachment(BinaryValue):
```
Julien00859: why isn't `open` implemented? At the moment it inherits from `BinaryValue`.

odoo/addons/base/models/ir_attachment.py
```diff
@@ -253,38 +281,45 @@ def _gc_file_store_unsafe(self):
...
             if attach.store_fname:
-                attach.raw = attach._file_read(attach.store_fname)
+                file = attach._file_read(attach.store_fname)
+                if isinstance(file, BinaryFile):
+                    files.append(file)
+                    if len(files) == 10:
+                        for file in files:
+                            file.content
+                    elif len(files) > 10:
+                        file.content
```
Julien00859: This deserve a comment. So starting at 10 files, you eagerly load them in memory. Why?

Julien00859: I don't like magic values btw

odoo/addons/base/tests/test_avatar_mixin.py
```diff
@@ -43,27 +40,27 @@ def test_partner_has_avatar_even_if_it_has_no_image(self):
...
 
     def test_external_partner_has_default_placeholder_image_as_avatar(self):
-        expectedAvatar = (
+        expected_avatar = (
             "<?xml version='1.0' encoding='UTF-8' ?>"
             "<svg height='180' width='180' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'>"
-            "<rect fill='#A76DBC' height='180' width='180'/>"
-            "<text fill='#ffffff' font-size='96' text-anchor='middle' x='90' y='125' font-family='sans-serif'>J</text>"
+            "<rect fill='#4EA7F2' height='180' width='180'/>"
+            "<text fill='#ffffff' font-size='96' text-anchor='middle' x='90' y='125' font-family='sans-serif'>M</text>"
```
Julien00859, line 60: why it changed?

odoo/addons/base/tests/test_binary.py
```diff
@@ -0,0 +1,85 @@
...
+        val = BinaryValue.from_file(path)
+        self.assertIsInstance(val, BinaryFile)
+        self.assertIsInstance(val.content, bytes)
+        self.assertEqual(val.decode(), val.content.decode())
+        self.assertIsInstance(val.to_base64(), str)
+
+        val2 = BinaryValue.from_file(val.local_path)
+        self.assertEqual(val.local_path, val2.local_path)
+
+    def test_binary_open(self):
```
Julien00859: use subtests

odoo/addons/base/tests/test_binary.py
```diff
@@ -0,0 +1,85 @@
...
+        self.assertIs(val.content, data)
+
+        self.assertEqual(val.size, len(data))
+        self.assertTrue(val.mimetype, "determine a mimetype")
+        self.assertEqual(val.decode(), 'test')
+        self.assertEqual(val.to_base64(), b64encode(data).decode())
+
+        self.assertFalse(BinaryValue.from_bytes(b''))
+
+    def test_binary_from_file(self):
```
Julien00859: missing assertion on mimetype

odoo/addons/base/tests/test_binary.py
```diff
@@ -0,0 +1,85 @@
...
+                self.assertEqual(f.read(), data, f"Read {i}")
+
+        val = BinaryValue.from_file('base/tests/files/file.csv')
+        for i in range(1, 3):  # can open multiple times
+            with val.open() as f:
+                self.assertEqual(len(f.read()), 18, f"Read {i}")
+
+        data = val.content
+        with val.open() as f:
+            self.assertIsInstance(f, BytesIO)
```
Julien00859: why not FileIO?

odoo/addons/base/tests/test_binary.py
```diff
@@ -0,0 +1,85 @@
...
+            with val.open() as f:
+                self.assertEqual(f.read(), data, f"Read {i}")
+
+        val = BinaryValue.from_file('base/tests/files/file.csv')
+        for i in range(1, 3):  # can open multiple times
+            with val.open() as f:
+                self.assertEqual(len(f.read()), 18, f"Read {i}")
+
+        data = val.content
+        with val.open() as f:
```
Julien00859: I don't like mutable variables in tests. `val` is a `BinaryFile` at this point, please make it explicit.

odoo/addons/base/tests/test_ir_attachment.py
```diff
@@ -12,39 +12,36 @@
...
     def assertApproximately(self, value, expectedSize, delta=1):
-        # we don't used bin_size in context, because on write, the cached value is the data and not
-        # the size, so we need on each write to invalidate cache if we really want to get the size.
-        try:
-            value = base64.b64decode(value.decode())
-        except UnicodeDecodeError:
-            pass
-        size = len(value) / 1024 # kb
+        value = value.content
+        size = len(value) / 1024  # kb
```
Julien00859: 🤓 
```suggestion
        size = len(value) / 1024  # kiB
```

odoo/addons/test_web/tests/test_web_save.py
```diff
@@ -38,39 +38,39 @@ def test_web_save_write(self):
 
     def test_web_save_computed_stored_binary(self):
         [result] = self.env['test_orm.binary_svg'].web_save(
-            {'name': 'test', 'image_wo_attachment': SVG_B64},
+            {'name': 'test', 'image_wo_attachment': SVG_B64.decode()},
             {'image_wo_attachment': {}, 'image_wo_attachment_related': {}},
         )
-        self.assertEqual(result['image_wo_attachment'], '322 bytes')  # From PostgreSQL
-        self.assertEqual(result['image_wo_attachment_related'], b'322.00 bytes')  # From human_size
+        self.assertEqual(result['image_wo_attachment'], '322.00 bytes')
```
Julien00859, line 44: What would be `322.5 bytes`? The previous value (integer, not float) was better

Julien00859: `human_size` is a shitty function btw, `{:.2f}` sucks, `{:.3g}` would be better

odoo/addons/test_web/tests/test_web_save.py
```diff
@@ -38,39 +38,39 @@ def test_web_save_write(self):
...
         # check web_save() on existing record
         self.env.invalidate_all()
         [result] = record.web_save(
-            {'image_wo_attachment': JPG_B64},
+            {'image_wo_attachment': JPG_B64.decode()},
             {'image_wo_attachment': {}, 'image_wo_attachment_related': {}},
         )
-        self.assertEqual(result['image_wo_attachment'], '29 kB')  # From PostgreSQL
-        self.assertEqual(result['image_wo_attachment_related'], b'29.47 Kb')  # From human_size
+        self.assertEqual(result['image_wo_attachment'], '29.47 Kb')
```
Julien00859, line 63: `kB` was the correct unit, "kilo Byte", base 10.
`K` doesn't exist as magnitude, `b` stands for `bit`, not `byte`.

Julien00859 (commented):
Review for core and base

odoo/tools/binary.py
```diff
@@ -0,0 +1,169 @@
...
+    @staticmethod
+    def from_bytes(data: Buffer):
+        """Create a BinaryValue from bytes."""
+        data = bytes(data)
+        if not data:
+            return _EMPTY_BINARY_DATA
+        return BinaryBytes(data)
+
+    @staticmethod
+    def from_file(path: str, *, filter_ext=(), reader: io.FileIO | None = None):
```
kmagusiak: @thle-odoo Should we refactor this? Unless it is fine to use it in safe_eval. If not, the alternative is to make this static method into a function in this module.

odoo/orm/fields_binary.py
kmagusiak: Because I want to avoid security warnings.

odoo/addons/base/models/avatar_mixin.py
kmagusiak: We have minimum python 3.12 where this is supported. I agree that it was painful in previous versions.

odoo/addons/base/models/ir_attachment.py
kmagusiak: Solved here... https://github.com/odoo/odoo/pull/245916

odoo/addons/base/models/ir_attachment.py
kmagusiak: Just a POC here in a separate commit. I am not sure if we want this anyways.

odoo/addons/base/tests/test_avatar_mixin.py
kmagusiak: Because I changed the source value for the hash which was mixing bytes, str and casting all to str.

odoo/tools/binary.py
thle-odoo: This function cannot be exposed in `safe_eval` (arbitrary path and no filter on the extension).

The alternative of making it a module-level function is a good one (or using the protection of `safe_eval` with a double underscore `__`).

odoo/addons/base/tests/test_binary.py
kmagusiak: The file was already read into memory.

odoo/addons/base/tests/test_binary.py
kmagusiak: Updating the code.

odoo/addons/test_web/tests/test_web_save.py
kmagusiak: We should fix `human_size` in a separate PR. Just using it here as is.

odoo/tools/binary.py
Julien00859: I would be +1 to have `def binary_from_file(path, ...)` at the top-level, instead of a `@staticmethod`.

Julien00859: Set a goddamn profile picture!
image

kmagusiak: Done. Just using the constructor.

odoo/tools/binary.py
```diff
@@ -0,0 +1,79 @@
+import base64
+import io
+from collections.abc import Buffer
+
+__all__ = ['BinaryBytes', 'BinaryValue']
+
+
+class BinaryValue(Buffer):
+    """Lazy representation of bytes. Abstract class."""
```
rco-odoo: It's not necessarily lazy.
```suggestion
    """ Value of a binary field in cache and on records. It is a proxy object to
    the value's content as bytes, enabling getting the content in a lazy way
    (for performance). Abstract class.
    """
```

odoo/tools/binary.py
```diff
@@ -0,0 +1,79 @@
...
+
+    __slots__ = ()
+
+    def open(self) -> io.IOBase:
+        """Open the file for reading."""
```
rco-odoo:
```suggestion
        """Open the value's content as a file."""
```

odoo/tools/binary.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __copy__(self):
+        # BinaryValue is immutable from the point of view of the API.
+        return self
+
+    def __deepcopy__(self, memo):
+        return self.__copy__()
+
+    def to_base64(self) -> str:
+        """Encode to a base64 string."""
```
rco-odoo:
```suggestion
        """Return the content encoded as a base64 string."""
```

odoo/orm/fields_binary.py
```diff
@@ -390,3 +322,45 @@ def _process_related(self, value, env):
             # Avoid the following `write` to fail if the related image was saved
             # invalid, which can happen for pre-existing databases.
             return False
+
+
+class BinaryValueAttachment(BinaryValue):
+    """Lazy BinaryValue that uses an attachment's ``raw`` field as contents.
+
+    A Binary field that stores the data in attachment's will use this class in
+    it's cache. Once we request the data, the `raw` field will be computed and
```
rco-odoo:
```suggestion
    its cache. Once we request the content, the `raw` field will be computed and
```

odoo/addons/test_orm/tests/test_html_converter.py
```diff
@@ -224,26 +220,22 @@ class TestBinaryExport(TestBasicExport):
     def test_image(self):
         converter = self.env['ir.qweb.field.image']
 
-        with file_open(os.path.join(directory, 'test_vectors', 'image'), 'rb') as f:
-            content = f.read()
-
-        encoded_content = base64.b64encode(content)
-        value = converter.value_to_html(encoded_content, {})
+        file = self.read_file_contents('test_orm/tests/test_vectors/image')
```
rco-odoo: The name `file` is confusing.  It's a `BinaryValue`, not a file.  Simply call it `binary`:
```suggestion
        binary = self.read_file_contents('test_orm/tests/test_vectors/image')
```

odoo/addons/base/models/ir_attachment.py
```diff
@@ -146,15 +151,14 @@ def _get_path(self, bin_data, sha):
         return fname, full_path
 
     @api.model
-    def _file_read(self, fname, size=None):
+    def _file_read(self, fname):
```
rco-odoo: Given that there's no docstring, a type signature may provide useful documentation.

```suggestion
    def _file_read(self, fname: str) -> BinaryValue:
```

odoo/addons/base/models/ir_attachment.py
```diff
@@ -971,3 +970,56 @@ def _migrate_remote_to_local(self):
             return
         if self.type == 'url':
             raise ValidationError(_("URL attachment (%s) shouldn't be migrated to local.", self.id))
+
+
+class LocalBinaryFile(BinaryValue):
```
rco-odoo, line 976: Why "Local" ?  Why not simply `BinaryFile` ?

odoo/addons/base/models/ir_attachment.py
```diff
@@ -971,3 +970,56 @@ def _migrate_remote_to_local(self):
...
+    __slots__ = ('__content', '__mimetype', '__path', '__stat')
+
+    def __init__(self, path: str, model: IrAttachment):
+        """ Open a file as a binary value.
+
+        :param path: absolute path to the file
+        :param model: model to check the path
+        :raise OSError: if the file cannot be opened
+        """
+        path = model._full_path(path)
```
rco-odoo, line 987: This line should not be part of the BinaryFile abstraction, and actually contradicts the docstring above.  It's an implementation detail of attachments, and should remain part of _file_read() in my opinion.  Also, removing it also removes the need for parameter `model`.

odoo/addons/base/models/ir_attachment.py
```diff
@@ -971,3 +970,56 @@ def _migrate_remote_to_local(self):
...
+        if not stat.S_ISREG(self.__stat.st_mode):
+            raise FileNotFoundError(f"Path is not a regular file: {path}")
+        self.__content: bytes | None = None
+        self.__mimetype: str | None = None
+
+    def open(self):
+        if self.__content is not None:
+            return super().open()
+        # open the file
+        return open(self.__path, 'rb')
```
rco-odoo, line 1000: To check: is a file object considered safe from the point of view of `safe_eval` ?

odoo/addons/base/models/ir_attachment.py
```diff
@@ -971,3 +970,56 @@ def _migrate_remote_to_local(self):
...
+            else:
+                self.__mimetype = guess_file_mimetype(self.local_path)
+        return self.__mimetype
+
+    @property
+    def size(self):
+        return self.__stat.st_size
+
+    @property
+    def local_path(self):
```
rco-odoo: Why making this part of the public API of those objects?  It's unnecessary, and not even useful IMHO.

odoo/addons/base/models/ir_attachment.py
Julien00859: The public attributes and methods of the returned `_io.BufferedReader` are the following:
```
['close', 'closed', 'detach', 'fileno', 'flush', 'isatty', 'mode', 'name', 'peek', 'raw', 'read', 'read1', 'readable', 'readinto', 'readinto1', 'readline', 'readlines', 'seek', 'seekable', 'tell', 'truncate', 'writable', 'write', 'writelines']
```

We open the file with `"rb"` which disables half of them.

* flush, truncate, writable, write, writelines are disabled.
* close, closed, fileno, isatty, mode, read, read1, readable, readinto, readinto1, readline, readlines, seek, seekable are tell have no side-effect beyond the file and only return `bool | bytes | str | int | None` which are fine.
* **name** is a bit annoying as it leaks the path to the file on the filesystem, don't know if it's ok.
* **detach** and **raw** return a `_io.FileIO` object that has about the sames attributes and methods.

The public attributes and methods of `_io.FileIO` are the following:
```
['close', 'closed', 'closefd', 'fileno', 'flush', 'isatty', 'mode', 'name', 'read', 'readable', 'readall', 'readinto', 'readline', 'readlines', 'seek', 'seekable', 'tell', 'truncate', 'writable', 'write', 'writelines']
```
Only `closefd`, `readall` are new, and IMO pose no security problem.

addons/account/demo/account_demo.py
```diff
@@ -1,17 +1,24 @@
...
+from odoo.tools.misc import formatLang
 from odoo.exceptions import UserError, ValidationError
 from odoo.addons.account.models.chart_template import template
+from odoo.tools import BinaryBytes, file_open
+
 
 _logger = logging.getLogger(__name__)
 
 
+def read_file_contents(path: str) -> BinaryBytes:
```
Julien00859: Would make for a fine standard function in our tools, as `file_read(path)`

addons/account_peppol/models/account_move_send.py
```diff
@@ -188,15 +187,15 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
                     )
                     continue
 
-                if len(xml_file) > 64000000:
+                if xml_file.size > 64000000:
```
Julien00859, line 188: genuine question, couldn't we have `__len__()`?

addons/account/demo/account_demo.py
Julien00859: When we run Odoo with the warnings for the "unclosed file descriptor", we get a fair share of those. I never took the time to search and fix them all. A `file_read` function would make it less likely that users do `file_open(...).read()`.

addons/attachment_indexation/models/ir_attachment.py
```diff
@@ -214,6 +214,7 @@ def extract_text(content):
 
     def _index_pdf(self, bin_data):
         '''Index PDF documents'''
+        bin_data = bin_data.content
```
Julien00859: The only two places where `bin_data.content` are used are:
```
if bin_data.startswith(b"%PDF-")
```
and
```
f = io.BytesIO(bin_data)
```

We might do this instead:
```diff
diff --git a/addons/attachment_indexation/models/ir_attachment.py b/addons/attachment_indexation/models/ir_attachment.py
index 8ff1ed0c0509..8cc1c3a948f0 100644
--- a/addons/attachment_indexation/models/ir_attachment.py
+++ b/addons/attachment_indexation/models/ir_attachment.py
@@ -214,7 +214,7 @@ class IrAttachment(models.Model):
-    def _index_pdf(self, bin_data):
+    def _index_pdf(self, bin_data: BinaryValue):
         '''Index PDF documents'''
-        if not bin_data.startswith(b'%PDF-'):
+        if bin_data.mimetype != 'application/pdf':
             return ""
         try:
             if not importlib.util.find_spec('pdfminer.high_level'):
@@ -227,7 +227,6 @@ class IrAttachment(models.Model):
         except ImportError:
             # warned already during init of module
             return ""
-        f = io.BytesIO(bin_data)
         try:
             resource_manager = PDFResourceManager()
             laparams = LAParams(detect_vertical=True)
@@ -238,8 +237,9 @@ class IrAttachment(models.Model):
                 laparams=laparams
             ) as device:
                 interpreter = PDFPageInterpreter(resource_manager, device)
-                for page in PDFPage.get_pages(f):
-                    interpreter.process_page(page)
+                with bin_data.open('rb') as f:
+                    for page in PDFPage.get_pages(f):
+                        interpreter.process_page(page)
 
                 buf = content.getvalue()
             return _clean_text_content(buf)
```

odoo/addons/base/models/ir_attachment.py
kmagusiak: @thle-odoo Can you confirm?

odoo/addons/base/models/ir_attachment.py
kmagusiak: It is the implementation for the local storage. (see next comment)

odoo/addons/base/models/ir_attachment.py
kmagusiak: Actually, I put it there so that if you have the class, you cannot instantiate it with an arbitrary path and read whatever you want. Here you need to pass a path and you need to know the checksum to read anything.

odoo/addons/base/models/ir_attachment.py
thle-odoo: As discussed privately, at first glance, we don't know how to use this object for unintended purpose (the previous comment clearly shows this).
We would prefer to avoid it ending up in the evaluation context of `safe_eval` **if possible** (otherwise, it seems harmless to me).

CC: @mwath

addons/base_automation/models/base_automation.py
```diff
@@ -707,6 +707,7 @@ def _get_eval_context(self, payload=None):
         self.ensure_one()
         model = self.env[self.model_name]
         eval_context = {
+            'BinaryBytes': BinaryBytes,
```
Julien00859, line 711: Hi @odoo/rd-security, I'm listing you the lines you care, so you don't need to read all files

addons/base_import/models/base_import.py
```diff
@@ -480,7 +480,7 @@ def _read_xlsx(self, options):
         import openpyxl  # noqa: PLC0415
         import openpyxl.cell.cell as types  # noqa: PLC0415
         import openpyxl.styles.numbers as styles  # noqa: PLC0415
-        book = openpyxl.load_workbook(io.BytesIO(base64.b64decode(self.file or b'')), data_only=True)
+        book = openpyxl.load_workbook(io.BytesIO(self.file.content), data_only=True)
```
Julien00859:
```suggestion
        with self.file.open() as file:
            book = openpyxl.load_workbook(file, data_only=True)
```

addons/base_import/models/base_import.py
```diff
@@ -519,7 +519,7 @@ def _read_xlsx(self, options):
 
     def _read_ods(self, options):
         from . import odf_ods_reader  # noqa: PLC0415
-        doc = odf_ods_reader.ODSReader(file=io.BytesIO(base64.b64decode(self.file or b'')))
+        doc = odf_ods_reader.ODSReader(file=io.BytesIO(self.file.content))
```
Julien00859:
```suggestion
        with self.file.open() as file:
            doc = odf_ods_reader.ODSReader(file=file)
```

addons/base_import/models/base_import.py
```diff
@@ -1112,9 +1112,9 @@ def parse_preview(self, options, count=10):
             # preview to a list in the return.
             _logger.debug("Error during parsing preview", exc_info=True)
             preview = None
-            if self.file_type == 'text/csv':
-                preview_bytes = base64.b64decode(self.file or b'')
-                preview = preview_bytes[:ERROR_PREVIEW_BYTES].decode('iso-8859-1')
+            if self.file_type == 'text/csv' and self.file:
+                with self.file.open() as f:
```
Julien00859, line 1118: I'm reading those opens and more than once my first thought was "oh there's a missing `"rb"` here", only to realize that BinaryValue.open takes no argument, and always use `"rb"`. So here is a silly idea, feel free to reject it as total non-sense.

```py
def open(self, mode):
    if mode != "rb":
        raise ValueError('mode must be "rb"')
    ...
```

addons/l10n_sa_edi/models/account_journal.py
kmagusiak: @odoo/rd-accounting FYI

odoo/orm/fields_binary.py
kmagusiak: Added a checksum to raise if we have a concurrent modification.

addons/base_import_module/models/base_import_module.py
```diff
@@ -19,18 +17,18 @@ class BaseImportModule(models.TransientModel):
     def import_module(self):
         self.ensure_one()
         IrModule = self.env['ir.module.module']
-        zip_data = base64.decodebytes(self.module_file)
+        zip_data = self.module_file
         fp = BytesIO()
         fp.write(zip_data)
-        res = IrModule._import_zipfile(fp, force=self.force, with_demo=self.with_demo)
+        IrModule._import_zipfile(fp, force=self.force, with_demo=self.with_demo)
```
Julien00859:
```suggestion
        with self.module_file.open() as fp:
            IrModule._import_zipfile(fp, force=self.force, with_demo=self.with_demo)
```

addons/base_import_module/tests/test_import_module.py
```diff
@@ -499,7 +498,7 @@ def test_import_module_field_file(self):
         ]
         self.import_zipfile(files)
         logo_path, logo_data = files[2]
-        self.assertEqual(base64.b64decode(self.env.ref('foo.logo').datas), logo_data)
+        self.assertEqual(self.env.ref('foo.logo').raw.content, logo_data)
```
Julien00859, line 501: Another silly general question, now that I'm yet-again reading the PR in full. (the question is general, not specific to this line)

Why is it a `content` property, and not a `read()` function? Most of the time, the thing is going to be an attachment. `raw.content` makes it look like it is cheap, `raw.read()` would coin that there's an open+read under the hood. Sure the thing is put in cache, so it's only the first time that `read()` is gonna *actually* be slow and read the file from the filesystem. Do we actually want to cache the content ourselves? Isn't the kernel already keeping most read (past-tense) file in RAM, even after they are closed?

I'm very sorry to be asking the question now, and not weeks ago. Came to mind only now :(

addons/bus/websocket.py
```diff
@@ -560,7 +561,9 @@ def _send_frame(self, frame):
             )
         if isinstance(frame.payload, str):
             frame.payload = frame.payload.encode('utf-8')
-        elif not isinstance(frame.payload, (bytes, bytearray)):
+        elif isinstance(frame.payload, Buffer):
+            frame.payload = bytes(frame.payload)
```
Julien00859, line 565: @tsm-odoo in the previous code there was no copy, now in case it is a bytearray/memoryview there's gonna be a copy, can we afford that O(n) copy here?

addons/certificate/models/certificate.py
```diff
@@ -99,9 +98,9 @@ def _compute_private_key(self):
         attachments = self.env['ir.attachment'].search([
             ('res_model', '=', 'certificate.key'),
             ('res_field', '=', 'content'),
-            ('res_id', 'in', self.ids)
+            ('res_id', 'in', self.private_key_id.ids),
```
Julien00859, line 101: Sorry... This must go to a dedicated `[FIX]` commit :(

addons/certificate/models/certificate.py
```diff
@@ -113,29 +112,29 @@ def _compute_private_key(self):
 
             # Create the private key in case of PKCS12 File and no private key is set
             if certificate.content_format == 'pkcs12':
-                content = certificate.with_context(bin_size=False).content
+                content = certificate.content.content
```
Julien00859, line 115: Nit, my point above about a `read()` method instead of a `content` property x)

Julien00859 (commented):
Image

addons/bus/websocket.py
kmagusiak: Note: if payload is already bytes, there is no copy.

addons/account/demo/account_demo.py
kmagusiak: This is basically what I had with `file_open` and we removed it because not enough usages. I added that helper here, because it is not widely used. I also added a similar version on `tests.BaseCase`.

addons/account_peppol/models/account_move_send.py
kmagusiak: Because len implies bool by default in python. Doing `if something.binary_field` would read the size, but now it's just true except for the empty binary value.

addons/base_import/models/base_import.py
kmagusiak: No need. The only open we want is binary read-only (and seekable).

addons/base_import_module/tests/test_import_module.py
kmagusiak: It could be a function. I found it easier to think of it as a property. They are generally faster, but don't have to be (just look at some computed fields in the ORM).
Note that you can also do `bytes(binary)`, but often you want to write in a suffix form `binary.read()` or `binary.content`. I also found it confusing to have "read" as a name because of `binary.open().read()`.

We want to cache it ourselves to avoid a system call everytime and the transfer of data. Since we have read the entire file already in memory, we probably want to keep it until the end of the tranaction.

addons/account_edi/tests/common.py
```diff
@@ -19,6 +16,7 @@ def _mocked_get_move_applicability(edi_format, move):
         'cancel': edi_format._cancel_invoice_edi,
     }
 
+
```
jbw-odoo:
```suggestion
```

addons/l10n_sa_edi/models/account_journal.py
jbw-odoo: > @odoo/rd-accounting FYI

Seems like that's what the api expects

jbw-odoo (changes requested):

addons/l10n_dk/tests/test_nemhandel_messages.py
```diff
@@ -106,8 +104,8 @@ def _get_mock_data(cls, error=False, nr_invoices=1):
             FAKE_UUID[1]: {
                 'accounting_supplier_party': '0184:16356706',
                 'filename': 'test_incoming',
-                'enc_key': file_open(f'{FILE_PATH}/enc_key', mode='rb').read(),
-                'document': b64encode(file_open(f'{FILE_PATH}/document', mode='rb').read()),
+                'enc_key': cls.read_file_contents(f'{FILE_PATH}/enc_key').decode(),  # base64 data
```
Julien00859: explicit is better than implicit, I would prefer:
```suggestion
                with file_open(f'{FILE_PATH}/enc_key', 'rb) as file:
                    'enc_key': b64decode(file.read())
```

addons/l10n_ro_edi/tests/test_xml_ubl_ro.py
```diff
@@ -19,7 +19,7 @@ def _patch_request_ciusro_download_answer(company, key_download, session):
     answer_data = {
         '3029027561': {
             'signature': {
-                'attachment_raw': b'Y291Y291',
+                'attachment_raw': 'Y291Y291',
```
Julien00859: b"coucou"

addons/loyalty/models/product_template.py
```diff
@@ -15,9 +13,7 @@ def create(self, vals_list):
         """ Override of `product` to set a default image for gift cards. """
         templates = super().create(vals_list)
         if templates and self.env.context.get('loyalty_is_gift_card_product'):
-            with file_open('loyalty/static/img/gift_card.png', 'rb') as f:
-                gift_card_placeholder = base64.b64encode(f.read())
-            templates.image_1920 = gift_card_placeholder
+            templates.image_1920 = BinaryBytes('loyalty/static/img/gift_card.png')
```
Julien00859: bug here
```suggestion
            with file_open('loyalty/static/img/gift_card.png', 'rb') as file:
                templates.image_1920 = BinaryBytes(file.read())
```

addons/account_edi/tests/common.py
william-andre: Putting a "blocking" request changes for following PEP8 is wild :laughing: 
https://peps.python.org/pep-0008/#blank-lines

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -234,20 +234,16 @@ def _link_invoice_documents(self, invoices_data):
...
             if 'tr_nilvera' in invoice_data['extra_edis']:
-                if attachment_values := invoice_data.get('ubl_cii_xml_attachment_values'):
-                    xml_file = BytesIO(attachment_values.get('raw'))
-                    xml_file.name = attachment_values['name']
-                else:
-                    xml_file = BytesIO(invoice.ubl_cii_xml_id.raw or b'')
-                    xml_file.name = invoice.ubl_cii_xml_id.name or ''
+                attachment_values = invoice_data.get('ubl_cii_xml_attachment_values', {})
+                xml_file = BytesIO(attachment_values.get('raw'))
+                xml_file.name = attachment_values.get('name')
```
Julien00859: the `else` part with `invoice.ubl_cii_xml_id` is gone, mistake?

addons/l10n_dk/tests/test_nemhandel_messages.py
kmagusiak: We are inside a dict here.

addons/website/controllers/form.py
```diff
@@ -189,7 +189,7 @@ def extract_data(self, model_sudo, values):
                 # If it's an actual binary field, convert the input file
                 # If it's not, we'll use attachments instead
                 if field_name in authorized_fields and authorized_fields[field_name]['type'] == 'binary':
-                    data['record'][field_name] = base64.b64encode(field_value.read())
+                    data['record'][field_name] = field_value.read()
```
Julien00859: ?

odoo/addons/base/models/ir_attachment.py
```diff
@@ -812,29 +819,21 @@ def _get_raw_access_token(self):
...
-        for values in values_list:
-            # Create only if record does not already exist for checksum and size.
+    def create_unique(self, vals_list):
+        result = self.browse()
+        for vals in vals_list:
             try:
-                bin_data = base64.b64decode(values.get('datas', '')) or False
-            except binascii.Error:
+                vals = self._check_contents(vals, accept_datas=True)
+            except self.Error:
```
Julien00859: self.Error?

odoo/tests/common.py
```diff
@@ -825,6 +830,12 @@ def profile(self, description='', **kwargs):
             profile_session=self.profile_session,
             **kwargs)
 
+    @classmethod
+    def read_file_contents(cls, path: str) -> BinaryBytes:
```
Julien00859:
```suggestion
    def file_read(cls, path: str) -> BinaryBytes:
```

addons/account/demo/account_demo.py
lordkrandel: First we make the function, then we advertise it, then people adopt it, it won't go the opposite way. 🤔  To me, it does belong to Odoo's framework, after 21 years.
I mean, I just grepped `file_open` in 18.0 and:

image

Then if it's not used, you either force replace automatically, advertise to developers, make draft replace PRs for others to complete, send a mail to security reviewers who get pinged for every filesystem access, live with the double standard for a while...

addons/l10n_it_edi/models/account_move.py
```diff
@@ -2013,7 +2013,7 @@ def _l10n_it_edi_send(self, attachments_vals):
             move.l10n_it_edi_header = False
             attachment = attachments_vals[move]
             filename = attachment['name']
-            content = b64encode(attachment['raw']).decode()
+            content = attachment['raw']
```
lordkrandel: First test on the Italian EDI.
As this content is sent to the EDI, it tracebacks.

```suggestion
            content = b64encode(attachment['raw']).decode()
```

A base64 encoded string is required, not bytes, see line `2088` down here

To reproduce: ask me ...?
- install `l10n_it_edi`,
- setup the test IAP parameter
  `self.env['ir.config_parameter'].sudo().set_str('l10n_it_edi.proxy_user_edi_mode', 'test')`
- Flag the IT EDI registration in the accounting settings 
  (which creates the proxy user on the test IAP)
- create an invoice (from the Italian company)
- send the invoice

Don't get me wrong, I think this PR is awesome, even if risky... we just need to have time to do a bunch of testing.

```
[...]
[...]
  File "/home/odoo/work/odoo/addons/l10n_it_edi/models/account_move.py", line 2088, in _l10n_it_edi_upload_single
    return proxy_user._make_request(f'{server_url}/api/l10n_it_edi/2/out/SdiRiceviFile', params={'file': file})
[...]
  File "/usr/lib/python3.13/json/encoder.py", line 182, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
                    f'is not JSON serializable')
TypeError: Object of type bytes is not JSON serializable
```

odoo/orm/fields_binary.py
kmagusiak: @rco-odoo I removed that comment, but do you have an idea why we need log access for images?

addons/base_import_module/tests/test_import_module.py
kmagusiak: Discussed IRL

addons/l10n_it_edi/models/account_move.py
```diff
@@ -886,7 +886,7 @@ def grouping_function_pension_funds(base_line, tax_data):
             'balance_multiplicator': -1 if self.is_inbound() else 1,
             'abs': abs,
             'pdf_name': pdf_values['name'] if pdf_values else False,
-            'pdf': b64encode(pdf_values['raw']).decode() if pdf_values else False,
+            'pdf': pdf_values['raw'] if pdf_values else False,
```
lordkrandel:
```suggestion
            'pdf': base64.b64encode(pdf_values['raw']).decode() if pdf_values else False,
```

Blocking error here:
This goes straight to the XML and to the EDI.
image

addons/account_edi_proxy_client/models/account_edi_proxy_user.py
```diff
@@ -218,12 +219,16 @@ def _renew_token(self):
...
-        :param symmetric_key:   The symmetric_key encrypted with self.private_key_id.public_key()
+        :param symmetric_key:   The symmetric_key (base64-encoded) encrypted with self.private_key_id.public_key()
+        :return:                Decrypted raw bytes
         '''
         decrypted_key = self.sudo().private_key_id._decrypt(base64.b64decode(symmetric_key))
-        return self.env['certificate.key']._account_edi_fernet_decrypt(decrypted_key, base64.b64decode(data))
+        key = Fernet(decrypted_key)
+        if not isinstance(data, str):
+            data = bytes(data)
+        return key.decrypt(data)
```
lordkrandel:
```suggestion
        if not isinstance(data, str | bytes):
            data = bytes(data)
        return key.decrypt(base64.base64decode(data))
```

I didn't do this part of the IAP communication encryption, but:

```
  [...]
  File "/home/odoo/work/odoo/addons/l10n_it_edi/models/account_move.py", line 392, in action_check_l10n_it_edi
    self._l10n_it_edi_update_send_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/odoo/work/odoo/addons/l10n_it_edi/models/account_move.py", line 2136, in _l10n_it_edi_update_send_state
    notification['xml_content'] = proxy_user._decrypt_data(encrypted_update_content, encryption_key)
                                  ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/work/odoo/addons/account_edi_proxy_client/models/account_edi_proxy_user.py", line 234, in _decrypt_data
    return key.decrypt(data)
           ~~~~~~~~~~~^^^^^^
  File "/home/odoo/work/.venv313/lib/python3.13/site-packages/cryptography/fernet.py", line 84, in decrypt
    timestamp, data = Fernet._get_unverified_token_data(token)
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/odoo/work/.venv313/lib/python3.13/site-packages/cryptography/fernet.py", line 118, in _get_unverified_token_data
    raise InvalidToken
cryptography.fernet.InvalidToken
```

lordkrandel: Happens while pressing this button: 
image

lordkrandel: Hmmm... looks like the key comes from the IAP, so maybe we can do the b64decode here:

image

addons/l10n_it_edi/models/account_move.py
kmagusiak: Given multiple errors here, it seems like some unit tests are missing. I am looking in related placed to what you sent and I am sure that `action_l10n_it_edi_send` will also fail. I will do the fixes early next week at latest.

lordkrandel: You're definitely correct, Italian EDI has a history of bug-driven development, with changes rigorously tested in production. I (and we) have been bug-fixing it for 4 years, as more EDI/Italian domains knowledge comes out.

We sure need some IAP communication unit testing, and some attachments testing at least.
I can make a [task](https://www.odoo.com/odoo/my-tasks/5956607) and start working on it this Monday... ? FYI @epictete

addons/account_edi_proxy_client/models/account_edi_proxy_user.py
kmagusiak: I think this is due to tests mocking requests with byte contents: addons/l10n_it_edi/tests/test_edi_import.py `'file': self.fake_test_content` is bytes and probably should be a string here.

@odoo/rd-accounting can you review usages of `_decrypt_data` function?

lordkrandel: What I meant is that the values of `'file'` and `'key'` come from the IAP server

There is an encryption level in the communication between Odoo client and the IAP, so that only the correct Odoo instance can decrypt their invoices.

Data and key come in as base64 encoded bytes, inside a JSON response from the IAP controllers.

This happens in `l10n_it_edi_proxy`, `peppol_proxy` and I guess `l10n_dk`.

I'd suggest leaving this `_decrypt_data` as it was, for the moment.
The IAP controllers and their API are stable for all Odoo versions from `17.0` to `master`, so it has to simultaneously change in all versions, or we have to add another controller method in master keep the 2 API method versions working in parallel.

image
image

I don't get what you mean, an error in runtime should be due to a mock in a test?

lordkrandel: See odoo/odoo#250439

Added some tests for the flow, reverted this point back, as `b64decode` is legit, as not coming from a file, but from an IAP controller

(which in turn reads a file, yes, but we cannot change that yet)

rco-odoo:
@robodoo override=ci/security

* instanciation of XML parser: the line was modified, but not that part;
* invocation of `open()` on `BinaryValue`;
* usage of x509: same as before, we only discarded base64 encoding/decoding;
* related binary fields (implicitly sudoed): same as before;
* `os.stat` used in trusted code (to get the size of a file in the filestore);
* `open(self.__path, 'rb')`: the file path cannot be forged;
* methods `__repr__` don't leak evaluable data.

odoo/addons/base/models/ir_attachment.py
```diff
@@ -971,3 +970,49 @@ def _migrate_remote_to_local(self):
...
+        if not stat.S_ISREG(self.__stat.st_mode):
+            raise FileNotFoundError(f"Path is not a regular file: {path}")
+        self.__content: bytes | None = None
+        self.__mimetype: str | None = None
+
+    def open(self):
+        if self.__content is not None:
+            return super().open()
+        # open the file
+        return open(self.__path, 'rb')
```
antonrom1, line 1000: I would say this needs an `assert isinstance(self, LocalBinaryFile)` to make sure self is not a dotdict

PR #210094, opened 2025-05-15
[IMP] mail:  show current datetime in avatar card and DM header

mear-odoo:
**Specifications:**
• Display the current time and date in the avatar card using the timezone of the
  user only if the card user's timezone differs from the current user's.
• Show the local time below the user name in the header of direct messages .

**Purpose:**
•  Enhance cross-timezone collaboration by presenting the local time of other
  users during interactions

task-4613148

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #244422, opened 2026-01-19
[FIX] im_livechat:  correct test_channel_get_livechat_visitor_info tz assertion

mear-odoo:
This PR fixes a failing assertion in `test_channel_get_livechat_visitor_info`.
The test fails since [*] because the demo data sets the OdooBot time zone to
'Europe/Brussels', while the test was asserting it to be False.
This PR explicitly sets the time zone of the OdooBot partner record and updates 
assertions accordingly.

[*] https://github.com/odoo/odoo/pull/210094

runbot-237778

---

### DIFF
--- a/addons/im_livechat/tests/test_get_discuss_channel.py
+++ b/addons/im_livechat/tests/test_get_discuss_channel.py
@@ -25,6 +25,7 @@ class TestGetDiscussChannel(TestImLivechatCommon, MailCommon):
 
     def test_channel_get_livechat_visitor_info(self):
         self.maxDiff = None
+        self.partner_root.tz = "Europe/Brussels"
         belgium = self.env.ref('base.be')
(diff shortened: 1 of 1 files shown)


addons/im_livechat/tests/test_get_discuss_channel.py
```diff
@@ -25,6 +25,7 @@ def test_get_discuss_channel(self):
 
     def test_channel_get_livechat_visitor_info(self):
         self.maxDiff = None
+        self.env.ref("base.partner_root").tz = "Europe/Brussels"
```
tsm-odoo: :pick: 

```suggestion
        self.partner_root.tz = "Europe/Brussels"
```

tsm-odoo (approved):
@robodoo delegate+

PR #244423, opened 2026-01-19
[FIX] l10n_it_edi: prevent crash on zero-amount foreign currency invoices

khsr-odoo:
**Steps to reproduce:**
1. Install the `l10n_it_edi` and switch to IT company
2. Go to Accounting → Invoice.
3. Create an invoice add partner and change the currency to USD
4. Add a product and apply tax and discount of 100% and confirm
3. Click on Send

**Issue:**
A ZeroDivisionError is raised during the Italian EDI XML generation. 
`ZeroDivisionError: float division by zero`

**Cause:**
The conversion rate is computed by dividing `amount_total` by `amount_total_signed` without handling the case where `amount_total_signed` is `0.0`.

**Solution:**
Skip the conversion rate computation when the signed total amount is zero, preventing the division by zero

**opw-5481821**

### DIFF
--- a/addons/l10n_it_edi/models/account_move.py
+++ b/addons/l10n_it_edi/models/account_move.py
@@ -587,7 +587,7 @@ class AccountMove(models.Model):
         # Reference line for finding the conversion rate used in the document
         conversion_rate = float_repr(
             abs(self.amount_total / self.amount_total_signed), precision_digits=5,
-        ) if convert_to_euros and self.invoice_line_ids else None
+        ) if convert_to_euros and self.invoice_line_ids and not self.currency_id.is_zero(self.amount_total_signed) else None
 
         # Reduce downpayment views to a single recordset
         downpayment_moves = self.invoice_line_ids._get_downpayment_lines().move_id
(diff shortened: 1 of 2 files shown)


bhra-odoo (approved):
LGTM.

addons/l10n_it_edi/models/account_move.py
```diff
@@ -587,7 +587,10 @@ def _l10n_it_edi_get_values(self, pdf_values=None):
         # Reference line for finding the conversion rate used in the document
         conversion_rate = float_repr(
             abs(self.amount_total / self.amount_total_signed), precision_digits=5,
-        ) if convert_to_euros and self.invoice_line_ids else None
+        ) if convert_to_euros and self.invoice_line_ids and not float_is_zero(
+            self.amount_total_signed,
+            precision_rounding=self.currency_id.rounding
```
kitan191: Thanks for the fix. :+1: 

We just want to prevent the division by zero.
So you can probably just use `and self.amount_total_signed`.
Otherwise, you can use `self.currency_id.is_zero(self.amount_total_signed)` instead of `float_is_zero(...)`.

Megaaaaaa (approved):
Hello, thank you for the fix! LGTM :+1:

robodoo r+

PR #240808, opened 2025-12-19
[FIX] l10n_es_edi_facturae: incorrect tax when negative amounts

santiagopim:
Issue:

When using negative amounts, for example to explicitly show a discount, the tax calculation is incorrect due to the application of the `abs` function. Furthermore, the way to find out if a tax is of the withholding type is based on the sign of the value, which can lead to error in these cases.

Cause:

A previous change (#237235) added the `abs` function so the 'TotalTaxesWithheld' would be always with positive value. But this also affects the calculation of taxes 'TotalTaxOutputs' in some cases, such as if the invoice line has negative values.

Steps to reproduce:

- Install `l10n_es_edi_facturae`
- With the ES company, create an invoice with some standard lines and one line with negative amounts, as an explicit discount
- Confirm the invoice and send (facturae)
- Open the XML attached in the chatter
- Observe that the taxes amounts (VAT and WITHHOLDING) are erroneous

A correct invoice should be for example:

```
Product        Price       Taxes       Amount
---------------------------------------------
PRODUCT-A       1000    21%VAT 15%WHI    1000
Discount        -100    21%VAT 15%WHI    -100
---------------------------------------------
                       Untaxed amount     900
                      Withholding 15%    -135
                              VAT 21%     189
                      -----------------------
                                 TOTAL    954
```

---
I confirm I have signed the [CLA](https://github.com/odoo/odoo/pull/157955) and read the PR guidelines at www.odoo.com/submit-pr

PR #244428, opened 2026-01-19
[FIX] l10n_es_edi_facturae: incorrect tax when negative amounts

santiagopim:
Issue:

When using negative amounts, for example to explicitly show a discount, the tax calculation is incorrect due to the application of the `abs` function. Furthermore, the way to find out if a tax is of the withholding type is based on the sign of the value, which can lead to error in these cases.

Cause:

A previous change (#237235) added the `abs` function so the `TotalTaxesWithheld` would be always with positive value. But this also affects the calculation of taxes `TotalTaxOutputs` in some cases, such as if the invoice line has negative values.

Steps to reproduce:

- Install `l10n_es_edi_facturae`
- With the ES company, create an invoice with some standard lines and one line with negative amounts, as an explicit discount
- Confirm the invoice and send (facturae)
- Open the XML attached in the chatter
- Observe that the taxes amounts (VAT and WITHHOLDING) are erroneous

A correct invoice should be for example:

```
Product        Price       Taxes       Amount
---------------------------------------------
PRODUCT-A       1000    21%VAT 15%WHI    1000
Discount        -100    21%VAT 15%WHI    -100
---------------------------------------------
                       Untaxed amount     900
                      Withholding 15%    -135
                              VAT 21%     189
                      -----------------------
                                 TOTAL    954
```

This PR replaces #240808 

---
I confirm I have signed the [CLA](https://github.com/odoo/odoo/pull/157955) and read the PR guidelines at www.odoo.com/submit-pr


addons/l10n_es_edi_facturae/models/account_move.py
```diff
@@ -469,10 +469,12 @@ def grouping_function_per_base_line_tax(base_line, tax_data):
             if not tax:
                 continue
 
-            is_withholding = tax.amount < 0.0
+            is_withholding = tax.l10n_es_edi_facturae_tax_type == '04'  # IRPF
```
aboo-odoo: Withholding taxes are taxes with negative amounts. What is the reason behind this change ? Wouldn't it remove other withholding from the computation ?

aboo-odoo (commented):
Thanks for your change 😄 I have one small comment:

addons/l10n_es_edi_facturae/models/account_move.py
santiagopim: The sign of the amount does not defines the tax type. In fact in my example with a 1000 base, there is a -150 withholding, and with a -100 base because the explicit discount, there is a +15 withholding.

addons/l10n_es_edi_facturae/models/account_move.py
```diff
@@ -469,10 +469,12 @@ def grouping_function_per_base_line_tax(base_line, tax_data):
...
             tax_data = self._l10n_es_edi_facturae_get_tax_node_from_tax_data({**values, 'grouping_key': tax})
             invoice_values['TaxesWithheld' if is_withholding else 'TaxOutputs'].append(tax_data)
-            invoice_values['TotalTaxesWithheld' if is_withholding else 'TotalTaxOutputs'] += abs(values['tax_amount_currency'])
+            invoice_values['TotalTaxesWithheld' if is_withholding else 'TotalTaxOutputs'] += values['tax_amount_currency']
```
aboo-odoo: I think the correct fix for this would be something in the lines of
```suggestion
        if is_withholding:
            invoice_values['TotalTaxesWithheld'] -= values['tax_amount_currency']
        else:
            invoice_values['TotalTaxOutputs'] += values['tax_amount_currency']
```

Absolute values are error-prone in accounting. This suggestion would allow to maintain the correct inverse sign for withholding tax.

What do you think ?

addons/l10n_es_edi_facturae/models/account_move.py
aboo-odoo: > The sign of the amount does not defines the tax type.

You are right, however withholding taxes are defined as taxes having a negative amount. Hence, the previous version is more generic and allows for less code update in the future in case new withholding taxes are added.

aboo-odoo (commented):
Also, could you retarget 18.0 ? It is the version in which the issue arose 😄

addons/l10n_es_edi_facturae/models/account_move.py
santiagopim: Ok, tests passed successfully in both solutions. In this case I wrapped all related commands:
``` python
            if is_withholding:
                invoice_values['TaxesWithheld'].append(tax_data)
                invoice_values['TotalTaxesWithheld'] -= values['tax_amount_currency']
            else:
                invoice_values['TaxOutputs'].append(tax_data)
                invoice_values['TotalTaxOutputs'] += values['tax_amount_currency']
```

aboo-odoo (approved):
Thank you!

@robodoo r+

PR #244432, opened 2026-01-19
[FIX] point_of_sale: filtering single attribute value products

hadielyakhni:
Steps to reproduce
------------------
1. Create a product "P", with an attribute e.g. "Size", having only one signe possible attribute value, e.g. "Small".
2. Add it to PoS.
3. In PoS, click the "i" button to show this product's info popup, and click on the "Small" button.

We are supposed to filter out the variants having "Small" attribute value, so we expect to see one product, however, we see NO PRODUCTS.

Reason
------
When clicking the attribute value button, "Small" in this case, we filter for proucts belonging to the product template "Product", which also have "Small" in their names. This assumes that a new product have been created for that "Small" attribute value, however that's not the case, since for single-attribute-value attributes, we don't create a separate product variant. I.e. since we can only have one option for "Size" which is "Small", we don't create a separate prodduct for it.

Fix
----
In the case of single-attribute-value attributes, don't add the attribute name to the search word since no products will be found. Just list all the products from that prodcuct template, they will naturally all match that single attribute value.

opw-5497593

### DIFF
--- a/addons/point_of_sale/static/src/app/screens/product_screen/product_info_popup/product_info_popup.js
+++ b/addons/point_of_sale/static/src/app/screens/product_screen/product_info_popup/product_info_popup.js
@@ -18,6 +18,12 @@ export class ProductInfoPopup extends AbstractAwaitablePopup {
         this.pos = usePos();
         Object.assign(this, this.props.info);
     }
+    searchAttributeValue(variant, attributeValue) {
+        const attributeValueSearch = variant.values.length > 1 ? attributeValue.search : "";
+        this.searchProduct(
+            `${attributeValueSearch};product_tmpl_id:${this.props.product.product_tmpl_id}`
+        );
+    }
     searchProduct(productName) {
         this.pos.setSelectedCategoryId(0);
         this.pos.searchProductWord = productName;
--- a/addons/point_of_sale/static/tests/tours/ProductScreen.tour.js
+++ b/addons/point_of_sale/static/tests/tours/ProductScreen.tour.js
@@ -273,3 +273,20 @@ registry.category("web_tour.tours").add("test_table_stand_number_exported", {
             ReceiptScreen.isShown(),
         ].flat(),
 });
+
+registry.category("web_tour.tours").add("test_products_variants_attribute_value_filtering", {
+    test: true,
+    steps: () =>
+        [
+            ProductScreen.confirmOpeningPopup(),
+            ProductScreen.clickProductInfo("Small Shelf"),
+            ProductScreen.clickProductInfoAttributeValue("Color", "White"),
+            ProductScreen.checkProductsNumber(2),
+            ProductScreen.productIsDisplayed("Small Shelf (Small)"),
+            ProductScreen.productIsDisplayed("Small Shelf (Medium)"),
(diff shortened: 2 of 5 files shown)


hadielyakhni:
@robodoo fw=no
Filtering has changed in 18.0

addons/point_of_sale/static/tests/tours/ProductScreen.tour.js
```diff
@@ -273,3 +273,20 @@ registry.category("web_tour.tours").add("test_table_stand_number_exported", {
...
+        [
+            ProductScreen.confirmOpeningPopup(),
+            ProductScreen.clickProductInfo("Small Shelf"),
+            ProductScreen.clickProductInfoAttributeValue("Color", "White"),
+            ProductScreen.checkProductsNumber(2),
+            ProductScreen.productIsDisplayed("Small Shelf (Small)"),
+            ProductScreen.productIsDisplayed("Small Shelf (Medium)"),
+            ProductScreen.clickProductInfo("Small Shelf"),
+            ProductScreen.clickProductInfoAttributeValue("Size", "Medium"),
+            ProductScreen.checkProductsNumber(1),
```
loukwets, line 289: Could you add a check of productIsDisplayed("Small Shelf (Medium)") here just to be 100% sure ?

loukwets (commented):
Nice fix !

addons/point_of_sale/static/tests/tours/ProductScreen.tour.js
hadielyakhni: Hello @loukwets, it's already added, did I miss your point? 😄

loukwets: I did sorry my bad !

addons/point_of_sale/static/src/app/screens/product_screen/product_info_popup/product_info_popup.js
```diff
@@ -18,9 +18,10 @@ export class ProductInfoPopup extends AbstractAwaitablePopup {
         this.pos = usePos();
         Object.assign(this, this.props.info);
     }
-    searchProduct(productName) {
+    searchProduct(variant, attributeValue) {
```
adgu-odoo: I don't really like the idea of changing the signature of a method in stable. You could add a new method `searchAttributeValue(variant, attribute_value)` that would do
```js
const attributeValueSearch = variant.values.length > 1 ? attributeValue.search : "";
this.searchProduct(`${attributeValueSearch};product_tmpl_id:${this.props.product.product_tmpl_id}`);
```
Such that you do not change the signature of the existing method and overrides made on the method will not fail in stable

adgu-odoo (changes requested):

addons/point_of_sale/static/src/app/screens/product_screen/product_info_popup/product_info_popup.js
hadielyakhni: Thanks 👌🏼

adgu-odoo (approved):
robodoo r+

PR #244433, opened 2026-01-19
[FIX] mail: fix multiple click on "done" button

phenix-factory:
Before this commit, the activity done button could be clicked multiple times and generate multiple "done" notifications.
This commit disables the done button once pressed to avoid multiple clicks

task-5504259

### DIFF
--- a/addons/mail/models/mail_activity.py
+++ b/addons/mail/models/mail_activity.py
@@ -544,6 +544,7 @@ class MailActivity(models.Model):
         # marking as 'done'
         messages = self.env['mail.message']
         next_activities_values = []
+        ongoing_activities = self.filtered(lambda a: not a.date_done)
 
         # Search for all attachments linked to the activities we are about to archive. This way, we
         # can link them to the message posted and prevent their disparition. The move is done in
@@ -553,7 +554,7 @@ class MailActivity(models.Model):
             ('res_id', 'in', self.ids),
         ], ['res_id']).grouped('res_id')
 
-        for model, activity_data in self.filtered('res_model')._classify_by_model().items():
+        for model, activity_data in ongoing_activities.filtered('res_model')._classify_by_model().items():
             # Allow user without access to the record to "mark as done" activities assigned to them. At the end of the
             # method, the activity is archived which ensure the user has enough right on the activities.
             records_sudo = self.env[model].sudo().browse(activity_data['record_ids'])
--- a/addons/mail/static/src/core/web/activity_markasdone_popover.js
+++ b/addons/mail/static/src/core/web/activity_markasdone_popover.js
@@ -1,4 +1,4 @@
-import { Component, onMounted, useExternalListener, useRef } from "@odoo/owl";
+import { Component, onMounted, useExternalListener, useRef, useState } from "@odoo/owl";
 
 export class ActivityMarkAsDone extends Component {
     static template = "mail.ActivityMarkAsDone";
@@ -20,6 +20,7 @@ export class ActivityMarkAsDone extends Component {
     setup() {
         super.setup();
         this.textArea = useRef("textarea");
+        this.state = useState({ disableDoneButton: false });
         onMounted(() => {
             this.textArea.el.focus();
         });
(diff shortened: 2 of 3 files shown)


addons/mail/static/src/core/web/activity_markasdone_popover.js
```diff
@@ -33,6 +36,7 @@ export class ActivityMarkAsDone extends Component {
     }
 
     async onClickDone() {
+        this.state.disableDoneButton = true;
```
tsm-odoo, line 45: Component could be recycled, in which case the state will be kept as `true`. We should probably try/finally and reset the state.

tsm-odoo: Also, should be guarded on the `onClickDone` function as well since there could be 2 clicks before the rendering occurs.

phenix-factory: This is already `onClickDone`, I don't get it

tsm-odoo: You only ensure the button is disabled based on a state variable. However, there still is a gap between the moment you disable it on the JS side and the moment the change will be effective in the DOM. That's why the guard in the template (disabled) is not enough. And that's why we should also check that the button is enabled in the JS callback.

addons/mail/models/mail_activity.py
```diff
@@ -564,31 +564,34 @@ def _action_done(self, feedback=False, attachment_ids=None):
...
-                # TODO: Fix void res_id on attachment when you create an activity with an image
-                # directly, see route /web_editor/attachment/add
-                if message_attachments := activity_attachments.get(activity.id):
-                    message_attachments.write({
-                        'res_id': activity_message.id,
-                        'res_model': activity_message._name,
-                    })
-                    activity_message.attachment_ids = message_attachments
-                messages += activity_message
+                if not activity.date_done:
```
tsm-odoo: early continue would be less diff+less indent.

addons/mail/static/src/core/web/activity_markasdone_popover.js
```diff
@@ -41,6 +48,7 @@ export class ActivityMarkAsDone extends Component {
         await this.props.activity.markAsDone();
         this.props.onActivityChanged(thread);
         await thread.fetchNewMessages();
+        this.state.disableDoneButton = false;
```
tsm-odoo: Should probably be wrapped into try/finally to catch errors :)

addons/mail/models/mail_activity.py
```diff
@@ -558,6 +558,8 @@ def _action_done(self, feedback=False, attachment_ids=None):
             # method, the activity is archived which ensure the user has enough right on the activities.
             records_sudo = self.env[model].sudo().browse(activity_data['record_ids'])
             for record_sudo, activity in zip(records_sudo, activity_data['activities']):
+                if activity.date_done:
+                    continue
```
tsm-odoo: It would be better to filter out done activities at the beginning.

tsm-odoo (approved):

addons/mail/static/src/core/web/activity_markasdone_popover.js
```diff
@@ -33,14 +36,22 @@ export class ActivityMarkAsDone extends Component {
...
-        this.props.onActivityChanged(thread);
-        await thread.fetchNewMessages();
+        this.state.disableDoneButton = true;
+        try {
+            await this.props.activity.markAsDone();
+            this.props.onActivityChanged(thread);
+            await thread.fetchNewMessages();
+        } finally {
+            this.state.disableDoneButton = false;
+        }
```
alexkuhn, line 52: You should keep `throw e` otherwise all errors like programming errors are silently dismissed, which is not ok.

addons/mail/static/src/core/web/activity_markasdone_popover.js
```diff
@@ -20,6 +20,9 @@ export class ActivityMarkAsDone extends Component {
     setup() {
         super.setup();
         this.textArea = useRef("textarea");
+        this.state = useState({
+            disableDoneButton: false,
+        });
```
alexkuhn: Nitpick: could be in a single line.
More effective LOCs to see in editor without having to scroll, the better!

alexkuhn (commented):
LGTM otherwise.

addons/mail/static/src/core/web/activity_markasdone_popover.js
phenix-factory: The linter disagrees: 
image
Since it's not caught, I think it's still going to throw.

tsm-odoo: Yeah, finally doesn't catch the error. It's just a way to always reset the state for consistency.

alexkuhn: Ah oops I misread `finally` from `catch`. No issue here indeed.

alexkuhn (approved):
LGTM
Task says problem spotted in 19.0. Considering the diff, I'd say it's worth backport 19.0 at least.
@robodoo delegate+

PR #201360, opened 2025-03-12
[IMP] payment: copy payment providers for new companies

nipl-odoo:
**Before this commit**:

- When creating a new company, payment providers were not copied from the main company.
- The process of adding payment providers was unintuitive; users had to manually duplicate providers and change the company field.

**After this commit**:

- Payment providers from the main company are automatically copied when a new company is created, reducing manual effort.
- A clear instructional message is added to the blank page for existing companies, guiding users on how to configure payment providers.

Affected Version - master
Task - 4523394

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #244441, opened 2026-01-19
[FIX] website_sale: ensure provider and website company consistency

sawer-odoo:
What is fixed
-------------

The test_payment_provider_visibility_with_portal didn't pass due to a multi company issue. In demo data, "Demo" payment provider exists for different companies which led the search method to return every one of them. Trying to assign the shop website id to each of them then raises an error as their company_id are inconsistent.

The fix
-------------

We now ensure only one provider is selected and its company is the same as the website's company.

Runbot error : 237822

---

### DIFF
--- a/addons/website_sale/tests/test_main_controller.py
+++ b/addons/website_sale/tests/test_main_controller.py
@@ -24,7 +24,7 @@ class TestPaymentProviderVisibility(PaymentHttpCommon, SaleCommon):
         website_portal.write({'domain': base_url})
 
         self.provider.write({'website_id': website_portal.id})
-        restricted_provider = self.env['payment.provider'].sudo().search([('name', '=', 'Demo')])
+        restricted_provider = self.env['payment.provider'].sudo().search([('name', '=', 'Demo'), ('company_id', '=', website_shop.company_id.id)])
         restricted_provider.write({'state': 'test', 'website_id': website_shop.id})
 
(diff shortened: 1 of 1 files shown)


addons/website_sale/tests/test_main_controller.py
```diff
@@ -24,7 +24,8 @@ def test_payment_provider_visibility_with_portal(self):
         website_portal.write({'domain': base_url})
 
         self.provider.write({'website_id': website_portal.id})
-        restricted_provider = self.env['payment.provider'].sudo().search([('name', '=', 'Demo')])
+        restricted_provider = self.env['payment.provider'].sudo().search([('name', '=', 'Demo')], limit=1)
+        restricted_provider.company_id = website_shop.company_id
```
Feyensv: Why don't you directly filter on the company in the `search` instead of assigning it here ? 🤔

sawer-odoo: Hello ! Indeed, I was doing it first but I didn't know if the provider with the right company id was guarantee to exist.

Feyensv: Yes, and that's why the test started to fail in 18.4+ (#201360), I think `stock` adds another company in demo data leading to the other provider(s).

Feyensv (approved):
robodoo delegate+

PR #244447, opened 2026-01-19
Introduce store locator snippet

AlessandroLupo:
This PR introduces a new Store Locator snippet that displays a searchable map with multiple locations based on `res.partner`. It supports multiple markers, search, and optional opening hours (via `website_sale_collect`).

1. The `location_selector` component is moved from `website_sale_stock` to `website` for better reuse.
2. The `BuilderList` component gains two new props: `alternativeChoiceItemTemplate` and `fullWidthDropdown` for better customization.

task-4361433


addons/website/static/src/builder/plugins/store_locator_option.js
```diff
@@ -0,0 +1,78 @@
...
+        { value: "6", label: "100 km" },
+        { value: "5", label: "200 km" },
+        { value: "4", label: "400 km" },
+        { value: "3", label: "1000 km" },
+        { value: "2", label: "2000 km" },
+    ];
+
+    setup() {
+        super.setup();
+        this.orm = useService("orm");
```
robinlej: try using `usePlugin(ORM)`

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,110 @@
...
+        </BuilderRow>
+        <BuilderRow label.translate="Placeholder" t-if="this.isActiveItem('searchbar_opt')" level="2">
+            <BuilderTextInput dataAttributeAction="'mapSearchbarPlaceholder'" preview="false" default.translate="Your postal code"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Sidebar Location" t-if="this.isActiveItem('sidebar_opt')">
+            <BuilderButtonGroup dataAttributeAction="'mapSidebarLocation'" preview="false">
+                <BuilderButton title.translate="Left" dataAttributeActionValue="'left'"
+                    iconImg="'/website/static/src/img/snippets_options/pos_left.svg'" />
+                <BuilderButton title.translate="Right" dataAttributeActionValue="'right'"
+                    iconImg="'/website/static/src/img/snippets_options/pos_right.svg'" />
```
robinlej: You'll have to use a `ltrRtlMapping` key for these

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,212 @@
...
+        editingElement.dataset.locationsList = JSON.stringify(loadResult);
+    }
+
+    getValue({ editingElement }) {
+        return editingElement.dataset.locationsList;
+    }
+}
+
+export class RefreshStoreLocatorAction extends BuilderAction {
+    static id = "refreshStoreLocator";
```
robinlej, line 154: This action fails

```
UncaughtPromiseError > TypeError
Uncaught Promise > can't access property "contact_address_inline", locationsUpToDateMap.get(...) is undefined

TypeError: can't access property "contact_address_inline", locationsUpToDateMap.get(...) is undefined
    apply/<@https://122838246-master-design-theme.runbot293.odoo.com/web/assets/190/dbc8690/website.website_builder_assets.min.js:4562:313
    apply@https://122838246-master-design-theme.runbot293.odoo.com/web/assets/190/dbc8690/website.website_builder_assets.min.js:4562:235
```

addons/website/views/snippets/snippets.xml
```diff
@@ -585,6 +585,10 @@
                     t-image-preview="/website/static/src/img/snippets_previews/s_whatsapp.webp">
                     <keywords>cta, whatsapp, social, e-commerce, selling-points, prompt, forms, chatbots, call-to-action, assistants, assistance, customers, 24, 7, support, talk</keywords>
             </t>
+            <t t-snippet="website.s_store_locator" string="Store Locator" group="social"
```
robinlej: Add `label="Map"`

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,110 @@
...
+        </BuilderRow>
+        <BuilderRow label.translate="Display Searchbar" t-if="this.isActiveItem('sidebar_opt')" level="1"
+            preview="false">
+            <BuilderCheckbox dataAttributeAction="'mapSearchbar'" dataAttributeActionValue="'true'" id="'searchbar_opt'"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Placeholder" t-if="this.isActiveItem('searchbar_opt')" level="2">
+            <BuilderTextInput dataAttributeAction="'mapSearchbarPlaceholder'" preview="false" default.translate="Your postal code"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Sidebar Location" t-if="this.isActiveItem('sidebar_opt')">
+            <BuilderButtonGroup dataAttributeAction="'mapSidebarLocation'" preview="false">
```
robinlej, line 32: Not sure why `preview="false"`
I think most of your options could have a preview (and would benefit from it), I don't understand why it's prevented

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,110 @@
...
+                    iconImg="'/website/static/src/img/snippets_options/pos_left.svg'" />
+                <BuilderButton title.translate="Right" dataAttributeActionValue="'right'"
+                    iconImg="'/website/static/src/img/snippets_options/pos_right.svg'" />
+            </BuilderButtonGroup>
+        </BuilderRow>
+        <BuilderRow label.translate="Details">
+            <BuilderSelect dataAttributeAction="'mapDetails'" preview="false">
+                <BuilderSelectItem dataAttributeActionValue="'area'" title.translate="Area"
+                    id="'details_area'">Area</BuilderSelectItem>
+                <BuilderSelectItem dataAttributeActionValue="'tooltip'" title.translate="Tooltip">
```
robinlej, line 43: Issue of the OWL component: when you use `tooltip` and `showWebsite`, you can't click on the link

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,110 @@
...
+            t-if="this.isActiveItem('details_area')">
+            <BuilderCheckbox dataAttributeAction="'mapShowImage'" dataAttributeActionValue="'true'" />
+        </BuilderRow>
+        <BuilderRow>
+            <div class="d-flex flex-column w-100">
+                <p class="hb-row-label mb-0 flex-grow-0 flex-shrink-0 flex-basis-auto">
+                    <t t-out="'Store Locations'" />
+                </p>
+                <t t-if="this.state.hasAvailableRecords">
+                    <BuilderList
```
robinlej, line 67: When moving the BuilderList options, there's a traceback
Image

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,110 @@
...
+        <BuilderRow label.translate="Email" preview="false" level="1">
+            <BuilderCheckbox dataAttributeAction="'mapShowEmail'" dataAttributeActionValue="'true'" />
+        </BuilderRow>
+        <BuilderRow label.translate="Website" preview="false" level="1">
+            <BuilderCheckbox dataAttributeAction="'mapShowWebsite'"
+                dataAttributeActionValue="'true'" />
+        </BuilderRow>
+        <BuilderRow label.translate="Image" preview="false" level="1"
+            t-if="this.isActiveItem('details_area')">
+            <BuilderCheckbox dataAttributeAction="'mapShowImage'" dataAttributeActionValue="'true'" />
```
robinlej, line 59: I don't know where the issue lies, but images don't work. The src appears as `src="data:image/png;base64,[object Object]"`

addons/website/static/src/builder/plugins/store_locator_option.xml
robinlej, line 1: The default "Height" option doesn't make much sense in my view:
- If you check all details options (phone, email, website, image), it's already > 100% by default, so checking 75% height ""breaks"" the snippet (it adds another vertical scrollbar on the snippet)
- It's buggy when selecting 100%, the bottom of the map doesn't load properly:

Image

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,142 @@
...
+        this.debouncedSearchButton = useDebounced(() => {
+            this.state.locations = [];
+            this._updateLocations(this.state.zipCode);
+        }, 300);
+
+        onMounted(() => {
+            window.addEventListener("resize", this.debouncedOnResize);
+            this.updateSize();
+        });
+        onWillUnmount(() => window.removeEventListener("resize", this.debouncedOnResize));
```
robinlej: You can probably use `useListener(window, "resize", this.debouncedOnResize.bind(this))` instead

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,142 @@
...
+    setup() {
+        this.state = proxy({
+            locations: [],
+            viewMode: "list",
+            zipCode: this.props.zipCode,
+            selectedLocationId: "",
+            isSmall: this.env.isSmall,
+        });
+
+        this.debouncedOnResize = useDebounced(() => this.updateSize(), 300);
```
robinlej: `useDebounced(this.updateSize.bind(this), 300)`

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,142 @@
...
+            this.updateSize();
+        });
+        onWillUnmount(() => window.removeEventListener("resize", this.debouncedOnResize));
+
+        // Fetch new locations when the zip code is updated.
+        useEffect(
+            (zipCode) => {
+                this._updateLocations(zipCode);
+            },
+            () => [this.state.zipCode]
```
robinlej: useEffect doesn't work like that anymore
see https://odoo.github.io/owl/documentation/v3/owl/migration_owl2_to_owl3.html#3-useeffect-semantic-change

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,142 @@
...
+    //--------------------------------------------------------------------------
+    // Handlers
+    //--------------------------------------------------------------------------
+
+    /**
+     * Update displayed locations based on the zip in the searchbar. Then, if
+     * the old selected location is not anymore displayed, select the first
+     * location in the list.
+     *
+     * @private
```
robinlej: To my knowledge, we don't use `@private` anymore in JS
Likewise, methods are also not prefixed with `_` anymore

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,142 @@
...
+     * @param {String} searchedZip - The zip code used to look for close locations.
+     * @return {void}
```
robinlej: not necessary (we don't mention it anywhere else, if there's no return value, we know it returns nothing)

addons/website/static/src/components/location_selector/map_container/map_container.js
```diff
@@ -47,11 +38,15 @@ export class MapContainer extends Component {
...
              * CDN’s doesn't host the library anymore).
+             * loadCSS's targetDoc option is needed when the component is mounted inside an iframe,
+             * for example in website snippet selector (when browsing custom snippets).
              */
             try {
                 await Promise.all([
                     loadJS("https://unpkg.com/leaflet@1.9.4/dist/leaflet.js"),
-                    loadCSS("https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"),
+                    loadCSS("https://unpkg.com/leaflet@1.9.4/dist/leaflet.css", {
+                        targetDoc: this.props.containerEl?.ownerDocument,
```
robinlej: Why do you add it only for the CSS?

addons/website/static/src/components/location_selector/map/map.js
```diff
@@ -43,15 +56,41 @@ export class Map extends Component {
...
-                maxZoom: 19,
-                attribution: "&copy; <a href='http://www.openstreetmap.org/copyright'>OpenStreetMap</a>"
+
+            L.tileLayer("https://tile.openstreetmap.org/{z}/{x}/{y}.png", {
+                maxZoom: OSM_MAX_ZOOM,
+                attribution:
+                    "&copy; <a href='http://www.openstreetmap.org/copyright'>OpenStreetMap</a>",
             }).addTo(this.leafletMap);
+
+            this.leafletMap.addEventListener("moveend", () => {
```
robinlej, line 77: Don't forget to remove the listeners on unmount (or use `useListener` when possible, but I guess it's not possible in this case)

addons/website/static/src/components/location_selector/map/map.js
```diff
@@ -43,15 +56,41 @@ export class Map extends Component {
...
+                this.props.setVisibleLocations?.(this.getVisibleMarks(this.leafletMap));
+            });
+
+            if (this.props.pressControlToZoom) {
+                this.mapRef().dataset.zoomDisabledText = _t("Hold Ctrl and scroll to zoom");
+                this.leafletMap.scrollWheelZoom.disable();
+                this.leafletMap.getContainer().addEventListener("wheel", (e) => {
```
robinlej: You could add `{ passive: true }` on the wheel event

addons/website/views/snippets/s_store_locator.xml
robinlej, line 1: It doesn't work on mobile

Image

addons/website/static/src/components/location_selector/map_container/map_container.js
AlessandroLupo: If Leaflet needs to run inside a custom snippet preview, the CSS has to target the iframe, while the JS could still work from the parent document. That said, I don’t see a good reason to treat them differently, so I’ll set `targetDoc` for both the CSS and JS. :+1:

robinlej: Oh, so it's only for the custom snippets preview? Because indeed the JS is shared between the main window and the preview iframe. I don't think it would be a good idea to duplicate it in the iframe.

The issue is that it makes the fix not generic: it wouldn't work with any iframe, because not all iframes are handled that way, it's very specific to our snippet previews.
So maybe the logic should be more constrained and the comment should be clearer about the fact that it's especially made for custom snippet previews (for now you just give it as an example, while it is in fact the full scope of it).

addons/website/static/src/components/location_selector/map/map.js
```diff
@@ -53,12 +53,16 @@ export class Map extends Component {
         this.leafletMap = null;
         this.markers = [];
 
+        const boundWheelHandler = this.wheelEventHandler.bind(this);
```
robinlej: Like most scrolls/wheels etc., maybe this should be throttled (`useThrottleForAnimation`)

addons/html_builder/static/src/core/building_blocks/builder_list.js
```diff
@@ -212,17 +213,30 @@ export class BuilderList extends Component {
     }
 
     updateRecords() {
```
robinlej, line 214: If I click on the "update records" option on a fresh snippet, the records are removed from the list.
If i add a new record, then click to refresh, it only keeps that record. (same if I save in between)

robinlej (changes requested):
There's a new conflict on the dialog :/

addons/html_builder/static/src/core/building_blocks/builder_list.js
AlessandroLupo: This is intended. The two locations listed by default are demo data and they don't match any of the existing partners, even if the names are the same. The update action is supposed to drop entries that don't exist (anymore) in the database. So when you click "update" the demo data is dropped.
The two demo entries are hardcoded in the snippet  so that they show even if the user does not have demo data. The drawback is that they are actually not connected to real records. We could discuss better way to implement this of course.

addons/html_builder/static/src/core/building_blocks/builder_select.scss
```diff
@@ -131,6 +131,11 @@
         background-color: $o-we-bg-darker;
         color: var(--o-hb-select-color);
     }
+
+    &.o-hb-full-width-dropdown{
+        // Size of the website builder sidebar - 8 pixels
+        min-width: 280px!important;
```
robinlej:
```suggestion
    &.o-hb-full-width-dropdown {
        min-width: ($o-we-sidebar-width - 8px) !important;
```

robinlej: That said, given how it's displayed (i.e. not within the sidebar), I'm not sure why that min-width limit?

Image

addons/website/static/src/components/location_selector/location/location.js
robinlej, line 1: The 3rd commit title lacks a * after `website`: `[MOV] website, *: ...`

addons/website/static/src/components/location_selector/location_schedule/location_schedule.xml
```diff
@@ -1,16 +1,16 @@
...
-                <t t-call="website_sale_stock.locationSelector.dailySchedule"
+                <t t-call="website.locationSelector.dailySchedule"
                     weekday="openingHour_index"
                     openingPeriods="openingHour_value"/>
             </t>
         </span>
     </t>
 
-    <t t-name="website_sale_stock.locationSelector.dailySchedule">
+    <t t-name="website.locationSelector.dailySchedule">
```
robinlej, line 13: I'm not sure it makes sense to move that. If I'm not wrong, `opening_hours` come from the `StockWarehouse` model.
It doesn't cause any error, but it means you're referencing something from another module :/

addons/website_sale_collect/__manifest__.py
```diff
@@ -25,6 +25,7 @@
         "web.assets_frontend": [
             "website_sale_collect/static/src/**/*",
             ("remove", "website_sale_collect/static/src/js/pickup_location_many2one/**/*"),
+            'website/static/src/components/location_selector/**/*',
```
robinlej: Isn't it already in website's `web.assets_frontend`?

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,109 @@
...
+            t-if="this.isActiveItem('sidebar_opt')" level="1" preview="false">
+            <BuilderCheckbox dataAttributeAction="'mapHideOffscreenLocations'"
+                dataAttributeActionValue="'true'" action="'hideLocationsOffscreen'" />
+        </BuilderRow>
+        <BuilderRow label.translate="Display Searchbar" t-if="this.isActiveItem('sidebar_opt')" level="1"
+            preview="false">
+            <BuilderCheckbox dataAttributeAction="'mapSearchbar'" dataAttributeActionValue="'true'" id="'searchbar_opt'"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Placeholder" t-if="this.isActiveItem('searchbar_opt')" level="2">
+            <BuilderTextInput dataAttributeAction="'mapSearchbarPlaceholder'" preview="false" default.translate="Your postal code"/>
```
robinlej: The  default is not applied. The snippet shows `Zip or City` until you actually edit the placeholder.

Image

Honestly the easiest solution would be to use `default.translate="Zip or City"`, even if that doesn't solve the underlying issue 🤷

addons/website/models/website.py
```diff
@@ -263,6 +263,7 @@ def _get_blocked_third_party_domains_list(self):
     def _get_blocked_iframe_containers_classes(self):
         return {
             's_map',
+            's_store_locator',
```
robinlej: - There's no iframe in the snippet / component (leaflet creates divs and imgs), so there's no reason to add it here
- OSM's URLs are not flagged as a 3rd-party server that should be blocked by the cookies bar, so even if it was an iframe it wouldn't do anything

The choice of flagging it or not is debatable. Should we block it? Some people, mainly in Germany, think so because OSM receives at least the IP on an American server (.org) and we can't know how they handle it, but it's not clear.

Anyway, regardless of whether we should hide it or not, the class shouldn't be in this method.

addons/website/static/src/builder/plugins/store_locator_option.js
```diff
@@ -0,0 +1,78 @@
...
+    ];
+
+    setup() {
+        super.setup();
+        this.orm = usePlugin(ORM);
+        this.state = useDomState((editingElement) => {
+            const locations = JSON.parse(editingElement.dataset.locationsList || "[]");
+            return {
+                availableRecords: this.state.availableRecords ?? "[]",
+                hasAvailableRecords: this.state.hasAvailableRecords,
```
robinlej: That seems circular (this.state = this.state), how does that work?

I would have expected

```js
availableRecords: JSON.stringify(this.searchResult ?? []),
hasAvailableRecords: this.searchResult.length > 0,
```

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,109 @@
...
+                dataAttributeActionValue="'true'" />
+        </BuilderRow>
+        <BuilderRow label.translate="Image" preview="false" level="1"
+            t-if="this.isActiveItem('details_area')">
+            <BuilderCheckbox dataAttributeAction="'mapShowImage'" dataAttributeActionValue="'true'" />
+        </BuilderRow>
+        <BuilderRow>
+            <div class="d-flex flex-column w-100">
+                <p class="hb-row-label mb-0 flex-grow-0 flex-shrink-0 flex-basis-auto">
+                    <t t-out="'Store Locations'" />
```
robinlej: `t-out` is not translated
Not sure why you need a t-out anyway 🤔

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,109 @@
...
+            </div>
+        </BuilderRow>
+        <BuilderRow label.translate="Refresh Locations" preview="false"
+            t-if="this.state.hasLocations">
+            <BuilderButton
+                action="'refreshStoreLocator'"
+                title.translate="Update the selected locations with the latest data">
+                Refresh Now
+            </BuilderButton>
+        </BuilderRow>
```
robinlej: I don't see that button (and given the title attribute ("the selected locations"), I wonder how it's supposed to work)

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,109 @@
...
+    </t>
+
+    <t t-name="website.StoreLocatorChoiceItemTemplate">
+        <b t-out="choice.value.name" />
```
robinlej: Don't use `<b>`, use `<strong>` instead (or `<span>` with some class)

addons/website/static/src/builder/plugins/store_locator_option.xml
```diff
@@ -0,0 +1,109 @@
...
+        <br />
+        <div t-out="choice.value.contact_address_inline" />
```
robinlej: It produces invalid HTML (div within a span)

Image

With `<br>, you can just use a plain `span` with the same effect though.

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+        builder_actions: {
+            AddLocationToStoreLocatorAction,
+            HideLocationsOffscreenAction,
+            RefreshStoreLocatorAction,
+        },
+        on_will_save_handlers: this.onWillSave.bind(this),
+    };
+
+    onWillSave(rootEl = this.editable) {
+        // Remove store_locator snippets without entries
```
robinlej: Should it be removed? We have other snippets that can be empty, we generally don't remove them, we just hide them. (e.g. an empty dynamic snippet)

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+            this.dependencies.builderActions.getAction("scrollButtonSectionHeightClass").apply({
+                editingElement,
+                params: {
+                    mainParam: "o_half_screen_height",
+                },
+            });
+        }
+    }
+
+    clean() {}
```
robinlej: This doesn't seem necessary?

robinlej (changes requested):
I'm in the middle of the 4th commit, but I'll already publish these comments

addons/website/static/src/components/location_selector/location_schedule/location_schedule.xml
AlessandroLupo: I see your point, but do you think it is worth to keep only `LocationSchedule` in `website_sale_stock`? Another way of seeing it is that `LocationSchedule` is a "sub-component" of the `LocationSelector`, and when it is rendered on the website, "openingHours" is sourced directly from the HTML.

addons/website/static/src/builder/plugins/store_locator_option.xml
AlessandroLupo: This button is removed by the next commit "clarify BuilderList refresh behavior". The word "selected" is referred to the fact that refresh will affect only the locations already listed in the snippet (hence "selected" in the `BuilderList`). And actually the confusion about it is exactly the reason of why I changed the refresh behavior of BuilderList :smile:

robinlej: If it's removed in the next commit, either both commits can be squashed, or it shouldn't be added in this commit

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+}
+
+async function batchGeolocalize(orm, locationsMap, IDsToUpdate) {
```
robinlej: `idsToUpdate`, with lower case

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+    // locations could be added per time by the BuilderList component.
+    async load({ editingElement, value }) {
+        const newList = JSON.parse(value);
+        const oldList = JSON.parse(editingElement.dataset.locationsList || "[]");
+
+        // Early return if there are no new locations to process
+        if (oldList.length >= newList.length) {
+            return newList;
+        }
+
```
robinlej, line 118: Shouldn't you also delete the demo values once a new value is added?

I'm not entirely sure, because the way these 2 example values work feels weird anyway. Maybe it's better not to give the impression that adding a value will remove existing values. But it also seems wrong to have both example data and actual records displayed at the same time.

Maybe we shouldn't have the example values at all, drop the snippet with the "No store location yet" message by default and let the user discover the options.
That's how we do it with the embedded code for instance 🤔

addons/website/static/src/builder/plugins/store_locator_option.xml
robinlej, line 1: To make the BuilderList option more "in your face" when there's no record selected, could you hide all the map-related options in that case?

Image

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+                body: _t(
+                    "One or more selected locations could not be geolocalized and have not been added. Please make sure that the addresses are complete and correct."
+                ),
+            });
+        }
+        return newLocationsList;
+    }
+
+    apply({ editingElement, loadResult }) {
+        editingElement.dataset.locationsList = JSON.stringify(loadResult);
```
robinlej, line 145: Using the builder list's result in the data attribute leads to a problem: the BuilderList renders its items through a `t-foreach`, iterating over `item._id`. But here, two items can end up with the same `_id` value (which is not overridden anywhere).

I haven't analyzed everything deeply, but it looks like the `_id` is stored in the dataset, which then creates issues when you add items later on. The way to reproduce is not very clear to me, but I did it several times and it crashes. (The common ground from my attempts was that I dropped the snippet, changed the records, then saved, then added one or two new stores, then added them in the snippet.)

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+        if (oldList.length >= newList.length) {
+            return newList;
+        }
+
+        const oldLocationsIDs = new Set(oldList.map((l) => l.id));
+        const locationsMap = new Map(newList.map((l) => [l.id, l]));
+        const newLocationsId = Array.from(
+            locationsMap.keys().filter((id) => !oldLocationsIDs.has(id))
+        );
+        await batchGeolocalize(this.services.orm, locationsMap, newLocationsId);
```
robinlej, line 124: Let me know if I miss something, but we could avoid the ORM call when the long/lat are already set for the location.
The only case where you'd want to update them when they're already set, is if you change the address afterwards. 1) that's unlikely / rare, 2) it doesn't work anyway because the rest of the info is not updated (city, street), so you need to click on the refresh button anyway (which is fine to me for such a rare case).

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+    });
+}
+
+async function batchUpdateCalendar(orm, locationsMap, IDsToUpdate) {
+    const updateCalendarsProms = [];
+    // Tries to retrieve the opening hours for IDsToUpdate
+    let calendars;
+    try {
+        calendars = await orm.searchRead(
+            "stock.warehouse",
```
robinlej: This method should definitely not be in website. It should be a patch in stock warehouse at best.
(Or, other option, the opening_hours should be set on the partner model from another module.)

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+        const oldLocationsIDs = new Set(oldList.map((l) => l.id));
+        const locationsMap = new Map(newList.map((l) => [l.id, l]));
+        const newLocationsId = Array.from(
+            locationsMap.keys().filter((id) => !oldLocationsIDs.has(id))
+        );
+        await batchGeolocalize(this.services.orm, locationsMap, newLocationsId);
+        await batchUpdateCalendar(this.services.orm, locationsMap, newLocationsId);
+
+        let newLocationsList = Array.from(locationsMap.values());
+        const lenghtBeforeFiltering = newLocationsList.length;
```
robinlej: typo: leng_**th**_BeforeFiltering

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
```diff
@@ -0,0 +1,215 @@
...
+
+        // Update geolocations only if address has changed.
+        const locationsToGeolocalize = [];
+        locationsList.forEach((location) => {
+            const locationUpToDate = locationsUpToDateMap.get(location.id);
+            if (
+                locationUpToDate &&
+                location.contact_address_inline !=
+                locationUpToDate.contact_address_inline
+            ) {
```
robinlej, line 182: FYI, my linter is complaining:

```diff
if (
                locationUpToDate &&
-                location.contact_address_inline !=
-                locationUpToDate.contact_address_inline
+                location.contact_address_inline != locationUpToDate.contact_address_inline
)
```

addons/html_builder/static/src/core/building_blocks/builder_select.scss
AlessandroLupo: LEBL confirmed that this is a bug (affecting also `master`), so I would keep the code with your suggested change

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
AlessandroLupo: After a chat with LEBL we agreed on keeping the snippet also when empty

addons/website/static/src/components/location_selector/location_list/location_list.js
```diff
@@ -44,7 +27,7 @@ export class LocationList extends Component {
             (locations, selectedLocationId) => {
                 const selectedLocation = locations.find((l) => String(l.id) === selectedLocationId);
                 if (selectedLocation) {
-                    document.getElementById(`location-${selectedLocation.id}`).focus();
+                    document.getElementById(`location-${selectedLocation.id}`)?.focus();
```
robinlej, line 30: Why is this optional chaining necessary?

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,125 @@
...
+            locations: [],
+            viewMode: "list",
+            zipCode: this.props.zipCode,
+            selectedLocationId: "",
+            isSmall: this.uiService.isSmall,
+        });
+
+        this.debouncedOnResize = useDebounced(this.updateSize.bind(this), 300);
+        this.debouncedSearchButton = useDebounced(() => {
+            this.state.locations = [];
```
robinlej, line 41: Is this line needed? It's overridden anyway in `updateLocations`, right?

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Update displayed locations based on the zip in the searchbar. Then, if
+     * the old selected location is not anymore displayed, select the first
+     * location in the list.
+     *
+     * @param {String} searchedZip - The zip code used to look for close locations.
+     */
+    async updateLocations(searchedZip) {
```
robinlej, line 70: Not async

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,125 @@
...
+            this.state.locations = [];
+            this.updateLocations(this.state.zipCode);
+        }, 300);
+
+        useListener(window, "resize", this.debouncedOnResize.bind(this));
+
+        onMounted(() => {
+            this.updateSize();
+        });
+
```
robinlej, line 50:
```suggestion
onMounted(this.updateSize);
```

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,125 @@
...
+import { useDebounced } from "@web/core/utils/timing";
+
+export class LocationSelectorComponent extends Component {
+    static components = { LocationList, MapContainer };
+    static template = "website.locationSelector.component";
+    props = useProps({
+        mapZoom: t.string(),
+        showSidebar: t.boolean().optional(false),
+        showSearchbar: t.boolean().optional(false),
+        mapSearchbarPlaceholder: t.string().optional(_t("Zip or City")),
```
robinlej, line 15: Is it really `Zip or City`? It seems to only take the zip code into account?

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.xml
```diff
@@ -110,16 +105,27 @@
                 id="o_location_selector_search"
                 class="search-query form-control oe_search_box border-0 text-bg-light"
```
robinlej, line 106: Validating the input with enter doesn't update anything. It should have the same behavior as clicking/pressing enter on the submit button.

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
robinlej: It also does a precise search. Shouldn't it check for a "closest match" when there is no result? (Can be done in a follow-up PR.)

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
```diff
@@ -0,0 +1,125 @@
...
+
+        useListener(window, "resize", this.debouncedOnResize.bind(this));
+
+        onMounted(() => {
+            this.updateSize();
+        });
+
+        // Fetch new locations when the zip code is updated.
+        useEffect(() => {
+            this.updateLocations(this.state.zipCode);
```
robinlej, line 53: Does it have to use `useEffect`? Unless I'm mistaken, the state is never updated, the zip code always comes from the props. It looks to me it could be `onMounted` instead.

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.xml
```diff
@@ -32,75 +34,68 @@
...
-                    </div>
-                </t>
             </div>
         </t>
-
         <!-- Desktop view -->
-        <div t-else="" class="d-flex h-100 mx-0 mx-md-n3 overflow-hidden">
+        <t t-else="">
+            <div class="d-flex h-100 mx-0 mx-md-n3 overflow-hidden"
+            t-att-class="{'flex-row-reverse': this.props.sidebarLocation == 'right'}">
```
robinlej, line 51: `===`

addons/website/static/src/components/location_selector/map_container/map_container.js
```diff
@@ -47,11 +38,22 @@ export class MapContainer extends Component {
...
-             */
+             * When the component is mounted inside the custom snippet preview iframe, loadCSS should
+             * target the iframe instead of the main document, because otherwise the CSS would not be
+             * available.
+             **/
             try {
+                const isCustomSnippetPreview =
+                    this.props.containerEl?.ownerDocument.documentElement.classList.contains(
+                        "o_add_snippets_preview"
+                    );
```
robinlej, line 49: I don't like that we pass around the `containerEl` in the props, especially as it's only used to check if it's a custom snippet preview iframe. It looks very random and hackish.

To know if we're in the custom iframe, I think this could be a legit use of the env. So I suggest doing something like:

- in the interaction
```js
        this.env.windowContext = {};
        if (this.el.ownerDocument.documentElement.classList.contains("o_add_snippets_preview")) {
            this.env.windowContext = { iframePreviewDocument: this.el.ownerDocument };
        }
```

- here:
```js
                const snippetPreviewDoc = this.env.windowContext.iframePreviewDocument;
                await Promise.all([
                    loadJS("https://unpkg.com/leaflet@1.9.4/dist/leaflet.js"),
                    loadCSS("https://unpkg.com/leaflet@1.9.4/dist/leaflet.css", {
                        targetDoc: snippetPreviewDoc,
                    }),
                ]);
```

and remove the `props.containerEl`

(Btw if you want to check if it's a *custom* snippet preview, you could check `this.el.ownerDocument.defaultView.frameElement.id === "tabpanel_custom"`. Because here it just checks that it's a preview, not necessarily custom.)

addons/website/static/src/builder/plugins/store_locator_option.js
AlessandroLupo: It was indeed an overcomplicated way to provide a default value before `willStart` performed the ORM search. I simplified it, and also dropped `useDomState`

addons/website/models/website.py
AlessandroLupo: I see your point, removed the class

addons/website/static/src/components/location_selector/map/map.js
```diff
@@ -84,6 +125,24 @@ export class Map extends Component {
         );
     }
 
+    wheelEventHandler(e) {
+        if (e.metaKey || e.ctrlKey) {
```
robinlej, line 129: No need to check for both if you use `getActiveHotkey`, in that case you can check that `hotkey === "control"`.

addons/website/static/src/components/location_selector/map/map.js
```diff
@@ -40,20 +54,47 @@ export class Map extends Component {
...
+                attribution:
+                    "&copy; <a href='http://www.openstreetmap.org/copyright'>OpenStreetMap</a>",
             }).addTo(this.leafletMap);
+
+            this.leafletMap.addEventListener("moveend", () => {
+                this.props.setVisibleLocations?.(this.getVisibleMarks(this.leafletMap));
+            });
+
+            if (this.props.pressControlToZoom) {
+                this.mapRef().dataset.zoomDisabledText = _t("Hold Ctrl and scroll to zoom");
```
robinlej, line 82: Nit: should we have an alternative text for mac? (`"Hold ⌘ and scroll..."`)

addons/website/static/src/components/location_selector/map/map.scss
```diff
@@ -17,5 +17,35 @@
...
+        color: #FFFFFF;
+        font-size: 2rem;
+        justify-content: center;
+        align-items: center;
+        position: absolute;
+        display: flex;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
```
robinlej, line 37:
```suggestion
@include o-position-absolute(0, 0, 0, 0);
display: flex;
```
(place the mixin at the very start of the CSS rules)

addons/website/static/src/components/location_selector/map/map.scss
```diff
@@ -17,5 +17,35 @@
...
+        background: rgba(0, 0, 0, 0.5);
+        pointer-events:none;
+
+        // Leaflet elements could take up to z-index 700,
+        // see leafletjs.com/reference.html#map-pane
+        z-index: 800;
```
robinlej, line 43: Nit: reorganize the rules "from the “outside” in" (cf. [CSS coding guidelines](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html#properties-order) )

addons/website/static/src/components/location_selector/map/map.xml
```diff
@@ -1,10 +1,26 @@
...
-        <div class="o_location_selector_map w-100 flex-grow-1" t-ref="map"/>
+        <div class="o_location_selector_map w-100 flex-grow-1" t-ref="this.mapRef"/>
     </t>
 
     <t t-name="website.locationSelector.map.marker">
         <t t-call="website.locationSelector.map.pinIndicator" number="this.number"/>
     </t>
+
+    <t t-name="website.locationSelector.map.tooltip">
+        <b t-out="location.name"/><br/>
```
robinlej: `<strong>`

addons/website/static/src/components/location_selector/map/map.xml
```diff
@@ -1,10 +1,26 @@
...
+        <t t-out="location.street"/><br/>
+        <t t-out="location.zip_code"/> <t t-out="location.city"/>
+        <t t-if="showPhone and location.phone">
+            <br/><i class="oi oi-fw me-1" data-icon="phone"/> <t t-out="location.phone"/>
```
robinlej: `<i ... role="img" aria-hidden="true"/>`

addons/website/static/src/components/location_selector/map/map.xml
```diff
@@ -1,10 +1,26 @@
...
+        </t>
+        <t t-if="showEmail and location.email">
+            <br/><i class="oi oi-fw me-1" data-icon="mail"/> <t t-out="location.email"/>
```
robinlej: `<i ... role="img" aria-hidden="true"/>`

addons/website/static/src/components/location_selector/map/map.xml
```diff
@@ -1,10 +1,26 @@
...
+        </t>
+        <t t-if="showWebsite and location.website">
+            <br/><i class="oi oi-fw me-1" data-icon="link"/> <a t-attf-href="{{location.website}}" target="_blank"><t t-out="location.website"/></a>
```
robinlej: same

robinlej (changes requested):

addons/website/static/src/components/location_selector/map_container/map_container.xml
```diff
@@ -16,17 +26,47 @@
...
+            <div t-if="this.selectedLocation and this.props.showDetailsTextArea" class="o_location_selector_textarea d-none d-md-flex overflow-auto justify-content-between flex-column flex-lg-row gap-3 gap-lg-0 p-4">
                 <div class="col-lg-5 d-flex flex-column justify-content-between">
                     <div class="d-flex gap-2">
                         <div class="d-flex flex-column flex-grow-1">
                             <strong class="h5 fw-bold" t-out="this.selectedLocation.name"/>
                             <small t-out="this.selectedLocation.street"/>
                             <small t-out="this.getCityAndZipCode(this.selectedLocation)"/>
+                            <br/>
+                            <t t-if="(this.props.showImage and this.selectedLocation.image_256) or (this.props.showPhone and this.selectedLocation.phone) or (this.props.showEmail and this.selectedLocation.email) or (this.props.showWebsite and this.selectedLocation.website)">
+                                <b>Additional details:</b>
```
robinlej: `<strong>`

addons/website/static/src/components/location_selector/map_container/map_container.xml
```diff
@@ -16,17 +26,47 @@
...
+                            </t>
+                            <t t-if="this.props.showPhone and this.selectedLocation.phone">
+                                <small>
+                                    <i class="oi oi-fw me-1" data-icon="phone"/>
```
robinlej: Same for all the icons used like that: `role="img" aria-hidden="true"`. Or alternatively in this case: `role="img" aria-label="Phone"` (phone / email / website...), because as it's a list of additional info, having a leading label could be useful.

addons/website/static/src/components/location_selector/map_container/map_container.xml
```diff
@@ -53,7 +94,7 @@
 
             <!-- Mobile infos -->
             <button
-                t-if="this.selectedLocation"
+                t-if="this.selectedLocation and this.props.showDetailsTextArea"
                 class="btn collapsed d-flex d-md-none gap-2 gap-lg-3 w-100 border-0 p-3 bg-transparent text-start"
```
robinlej, line 98: Oh my.
That's mainly something brought from the existing template, so let's maybe not take the time now to improve it. But that collapse button is wrong on many levels. :(

- Why not use `details`/`summary` which exist for this exactly?
- There's no clear indication (as an aria-label or aria-description) of what the button does
- The `<button>` wraps the accordion response (it shouldn't)
- Which means if you add an email or a website, the `<button>` wraps one or two links => that's invalid HTML and completely inaccessible

addons/website/static/src/components/location_selector/map_container/map_container.xml
```diff
@@ -53,7 +94,7 @@
...
                 type="button"
                 data-bs-target="#map_collapseHours"
```
robinlej, line 100: Weird to add all the rest of the details but keep using IDs / controls that refer exclusively to `collapseHours`

addons/website/static/src/components/location_selector/utils.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+    const formattedOpeningHours = {
+        0: [],
+        1: [],
+        2: [],
+        3: [],
+        4: [],
+        5: [],
+        6: [],
+    };
```
robinlej, line 48:
```suggestion
const formattedOpeningHours = Array(7).fill([]);
```

An array or an object with numeric keys is (nearly) the same in JS.

addons/website/static/src/components/location_selector/utils.js
```diff
@@ -0,0 +1,62 @@
...
+    openingHours.forEach((period) => {
+        const { h: hour_from, m: minute_from } = toTimeString(period.hour_from);
+        const { h: hour_to, m: minute_to } = toTimeString(period.hour_to);
+        formattedOpeningHours[period.dayofweek].push(
+            _t("%(hour_from)s:%(minute_from)s - %(hour_to)s:%(minute_to)s", {
```
robinlej, line 53: 1. Does it make sense to translate that line?
2. Does it work as expected in RTL?

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
+.s_store_locator {
+    position: relative;
+
+    &.o_full_screen_height{
```
robinlej, line 4: Nit: keep a space between the selector and the `{` (same in all the CSS files)

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
...
+        height: 1vh !important;
```
robinlej, line 5: uh? Is that because you anticipate the loading time of the component?

- It should have a comment at least, because "full screen height -> 1vh" is a suspicious choice :D
- If it's indeed to avoid taking the full height when the component isn't loaded, I'm not even sure we should do that. At least planning the full screen height size avoids layout shifts.

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    .o_location_selector_map {
+        min-height: 40vh;
+    }
+
+    .o_location_selector_view {
+        // Force tab-pane to be displayed on desktop.
+        @include media-breakpoint-up(md) {
+            opacity: 1;
```
robinlej, line 21: I don't get it:

- Changing the opacity doesn't change the space it takes.
- I don't see anywhere `opacity: 0` that would explain a need for an override?
- Disabling the rule on my browser doesn't seem to change a thing?

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
...
+        grid-template-columns: auto 1fr;
+    }
+
+    .o_location_selector_schedule_hours {
+        gap: 0 map-get($spacers, 3);
+    }
+
+    .o_store_locator_component{
+        padding-left: 1;
+        padding-right: 1;
```
robinlej, line 55: That's invalid CSS. Did you mean to use the equivalent to Bootstrap classes `ps-1 pe-1`? If so, that's `0.25rem`

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
...
+        display: flex;
+        flex-direction: column;
+        // Required to have the map below the "scroll down" button (when present)
+        z-index: 0;
```
robinlej, line 59: Better than a hack: `isolation: isolate;`
That's exactly what it's made for

addons/website/static/src/snippets/s_store_locator/000.scss
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+}
+
+.o_add_snippets_preview .o_location_selector_map {
+    height: 80vh !important;
+}
```
robinlej, line 66: Make a `.preview.scss` file instead with

```js
.o_snippet_preview_wrap {
    [data-snippet="s_store_locator"] {
        .o_location_selector_map {
            height: 80vh !important;
        }
    }
}
```

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
AlessandroLupo: I agree that deleting the demo locations when new ones are added may give the wrong impression to the user. I lean towards emoving the demo locations: since we dropped the limitation of "is_company", the user will likely have entries ready to be inserted in the snippet. On the other hand, to be fair, we have other snippets with "demo" content (FloatingBlocks, image carousels).

addons/website/static/src/snippets/s_store_locator/store_locator.js
```diff
@@ -0,0 +1,38 @@
...
+            showEmail: dataset.mapShowEmail === "true" ? true : false,
+            showImage: dataset.mapShowImage === "true" ? true : false,
+            showPhone: dataset.mapShowPhone === "true" ? true : false,
+            showWebsite: dataset.mapShowWebsite === "true" ? true : false,
+            showSearchbar: dataset.mapSearchbar === "true" ? true : false,
+            showSidebar: dataset.mapSidebar === "true" ? true : false,
+            mapSearchbarPlaceholder: dataset.mapSearchbarPlaceholder,
+            sidebarLocation: dataset.mapSidebarLocation,
+            containerEl: this.el,
+        };
```
robinlej, line 25: move this in the setup, and store the props in `this.props`

addons/website/static/src/snippets/s_store_locator/store_locator.js
```diff
@@ -0,0 +1,38 @@
...
+        this.mountComponent(
+            this.el.querySelector(".o_store_locator_component"),
+            LocationSelectorComponent,
+            props
+        );
+    }
```
robinlej: Not absolutely necessary, but possibly move this in the dynamicContent with `t-component`

addons/website/static/src/snippets/s_store_locator/store_locator.js
```diff
@@ -0,0 +1,38 @@
...
+    static selector = ".s_store_locator";
+
+    start() {
+        const dataset = this.el.dataset;
+        const props = {
+            mapZoom: dataset.mapZoom,
+            hideOffscreenLocations: dataset.mapHideOffscreenLocations === "true" ? true : false,
+            locationsList: dataset.locationsList,
+            showDetailsTextArea: dataset.mapDetails == "area",
+            showDetailsTooltip: dataset.mapDetails == "tooltip",
```
robinlej, line 15: `===`

addons/website/static/src/snippets/s_store_locator/store_locator.js
```diff
@@ -0,0 +1,38 @@
...
+            showEmail: dataset.mapShowEmail === "true" ? true : false,
+            showImage: dataset.mapShowImage === "true" ? true : false,
+            showPhone: dataset.mapShowPhone === "true" ? true : false,
+            showWebsite: dataset.mapShowWebsite === "true" ? true : false,
+            showSearchbar: dataset.mapSearchbar === "true" ? true : false,
+            showSidebar: dataset.mapSidebar === "true" ? true : false,
```
robinlej, line 21: No need for the ternary operation. Just `showSidebar: dataset.mapSidebar === "true",`
(same for all the others)

addons/website/static/tests/tours/snippet_store_locator.js
```diff
@@ -0,0 +1,106 @@
...
+const assertLocationCount = (count) => [
+    {
+        content: "Check the number of locations listed in the option",
+        trigger: "div[data-container-title='Store Locator'] .o_we_table_wrapper table",
+        async run() {
+            await delay(100);
+            const els = document.querySelectorAll(
+                "div[data-container-title='Store Locator'] .o_we_table_wrapper table tr"
+            );
+            if (els.length != count) {
```
robinlej, line 20: `!==` (double check all your equalities please 🙏 )

addons/website/static/tests/tours/snippet_store_locator.js
```diff
@@ -0,0 +1,106 @@
...
+        {
+            content: "Check that the map is now rendered",
+            trigger: ":iframe section.o_location_selector_view",
+        },
+        {
+            content: "Check that the details textarea is displayed",
+            trigger: ":iframe div.o_location_selector_textarea",
+        },
+        {
+            content: "Check that no toltip is displayed",
```
robinlej: typo: tooltip

addons/website/__manifest__.py
```diff
@@ -19,6 +19,7 @@
         'google_recaptcha',
         'utm',
         'html_builder',
+        'base_geolocalize',
```
robinlej, line 22: We should ask FGE's opinion about adding this dependency

addons/website_sale_collect/static/src/js/location_selector/location/location.js
```diff
@@ -0,0 +1,10 @@
+import { patch } from '@web/core/utils/patch';
+
+import { Location } from '@website/components/location_selector/location/location';
+
+patch(Location, {
+    props: {
+        ...Location.props,
+        additionalData: { type: Object, optional: true },
```
robinlej, line 8: Shouldn't the props be in owl 3 syntax in the patch too?

robinlej (changes requested):

addons/website/static/src/builder/plugins/store_locator_option_plugin.js
AlessandroLupo: Geolocation is not done automatically when an address changes.
So the following can happen:
1. An address has been geolocated in the past (by the store locator or by other apps)
2. After some time, the address is changed, but no geolocation is performed
3. The user decides to add the location to the snippet
4. The option correctly fetches the new address, but no geolocation is triggered
5. When the user selects the location, this action geolocalize it

Maybe you are right that this case is rare enough, we can consider dropping it

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
AlessandroLupo: `state.zipCode` is actually updated by the searchbar input, that's why we need `useEffect`

addons/website/static/src/components/location_selector/map/map.js
AlessandroLupo: `getActiveHotkey` does not work with a `WheelEvent` because it misses the `key` property. The function only works with `keyboard` events

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.js
AlessandroLupo: Implemented a fuzzy search on zip code **and** city :+1:

addons/website/static/src/components/location_selector/location_selector_component/location_selector_component.xml
AlessandroLupo: Can't reproduce this problem, maybe it was a bug introduced by one of the fixup commits and already solved

addons/website/static/src/snippets/s_store_locator/store_locator.js
AlessandroLupo: The component should be mounted only when a condition is true (`locationList.length > 0`). This felt impossible to realize within `dynamicContent`, unless I am missing something.
I thought about tweaking Colibri to make `t-component` ignore `false`/`undefined` arguments, but in the end I decided to leave it in `start()`

addons/website/static/src/components/location_selector/location_list/location_list.js
AlessandroLupo: In some situations `selectedLocation` will default to `id = 0`. If "hide offscreen locations" is active, the element `#location-0` may not be found

PR #244449, opened 2026-01-19
[FIX] test_website_modules: shop performance tests

shrm-odoo:
Error:
AssertionError: Expected 63 queries but 64 where ran: Diff of select queries: {'product_ribbon': 1}

Cause:
Commit fc8d7bdc365a0d202b978d9a42d1471be6175725 removed the expected query count of `product.ribbon` model for
demo data installation.

Fix:
Increase the expected query count by 1 for the `product.ribbon` model when demo data is installed.

runbot-{237964, 237966}


kmagusiak (approved):
@robodoo r+
Probably lost between rebases or I did not install all modules when testing with demo data.

PR #244454, opened 2026-01-19
[FIX] html_editor: keep editor focus when clicking more options

dhba-odoo:
**Issue 1:**

Steps to Reproduce

- Click on the More options button in the power buttons.
- The powerbox opens, but the editor loses focus and the button receives focus.

Description of the issue:

- After clicking the power button, the button becomes focused and the editor loses focus.

Solution

- When clicking the power button, after the command is executed in the click event, explicitly restore focus to the editable area so the editor remains focused.

**Issue 2:**

Steps to reproduce

- Click More options in the Power Buttons to open the powerbox.
- Start typing `heading`.

Description of the issue:

- The powerbox does not filter commands and continues to show all commands.

Cause:
- In `search_powerbox_plugin`, commands are filtered only when `shouldUpdate` is true.
- `shouldUpdate` is set only when the powerbox is opened through `search_powerbox_plugin`.
- Power Buttons open the powerbox via `powerbox_plugin`, so `shouldUpdate` remains false and filtering is not triggered.

Solution:
- Introduced `openSearchPowerbox` in `searchPowerboxPlugin`.
- Updated the implementation to use this method instead of `openPowerbox` in `search_powerbox_plugin`.
- Instead of opening the powerbox via `powerbox_plugin`, it is now opened via `search_powerbox_plugin`, ensuring `shouldUpdate` is set correctly and commands are filtered on keypress.

task-5485088

### DIFF
--- a/addons/html_editor/static/src/main/power_buttons_plugin.js
+++ b/addons/html_editor/static/src/main/power_buttons_plugin.js
@@ -77,6 +77,7 @@ export class PowerButtonsPlugin extends Plugin {
             const btn = this.document.createElement("button");
             btn.className = `power_button btn px-2 py-1 cursor-pointer fa ${icon}`;
             btn.title = title;
+            this.addDomListener(btn, "pointerdown", (ev) => ev.preventDefault());
             this.addDomListener(btn, "click", () => this.applyCommand(run));
             return btn;
         };
--- a/addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
+++ b/addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
@@ -1,5 +1,8 @@
 import { fuzzyLookup } from "@web/core/utils/search";
 import { Plugin } from "../../plugin";
+import { _t } from "@web/core/l10n/translation";
+import { withSequence } from "@html_editor/utils/resource";
+import { closestElement } from "@html_editor/utils/dom_traversal";
 
 /**
  * @typedef {import("./powerbox_plugin").PowerboxCategory} CommandGroup
@@ -15,6 +18,21 @@ export class SearchPowerboxPlugin extends Plugin {
         delete_handlers: this.update.bind(this),
         post_undo_handlers: this.update.bind(this),
         post_redo_handlers: this.update.bind(this),
+        user_commands: {
+            id: "openSearchPowerbox",
+            run: () => {
+                const selection = this.dependencies.selection.getEditableSelection();
+                this.historySavePointRestore = this.dependencies.history.makeSavePoint();
+                // Anchor element for powerbox opened via power buttons.
+                this.powerButtonAnchorEl = closestElement(selection.anchorNode);
+                this.openSearchPowerbox();
+            },
+        },
+        power_buttons: withSequence(100, {
+            commandId: "openSearchPowerbox",
+            title: _t("More options"),
+            icon: "fa-ellipsis-v",
+        }),
     };
     setup() {
         const categoryIds = new Set();
@@ -26,6 +44,11 @@ export class SearchPowerboxPlugin extends Plugin {
         }
         this.categories = this.getResource("powerbox_categories");
         this.shouldUpdate = false;
+        this.addDomListener(this.editable, "pointerdown", () => {
+            if (this.powerButtonAnchorEl) {
+                this.powerButtonAnchorEl = false;
+            }
+        });
     }
     onBeforeInput(ev) {
         if (ev.data === "/") {
@@ -34,7 +57,7 @@ export class SearchPowerboxPlugin extends Plugin {
     }
     onInput(ev) {
         if (ev.data === "/") {
-            this.openPowerbox();
+            this.openSearchPowerbox();
         } else {
             this.update();
         }
@@ -49,7 +72,7 @@ export class SearchPowerboxPlugin extends Plugin {
             this.dependencies.powerbox.closePowerbox();
             return;
         }
-        const searchTerm = this.searchNode.nodeValue.slice(this.offset + 1, selection.endOffset);
+        const searchTerm = this.searchNode.nodeValue?.slice(this.offset + 1, selection.endOffset);
         if (!searchTerm) {
             this.dependencies.powerbox.updatePowerbox(this.enabledCommands, this.categories);
             return;
@@ -82,12 +105,31 @@ export class SearchPowerboxPlugin extends Plugin {
      */
     isSearching(selection) {
         return (
-            selection.endContainer === this.searchNode &&
-            this.searchNode.nodeValue &&
-            this.searchNode.nodeValue[this.offset] === "/" &&
-            selection.endOffset >= this.offset
+            this.powerButtonAnchorEl === closestElement(this.searchNode) ||
+            (selection.endContainer === this.searchNode &&
+                this.searchNode.nodeValue &&
+                this.searchNode.nodeValue[this.offset] === "/" &&
+                selection.endOffset >= this.offset)
         );
     }
+    openSearchPowerbox() {
+        const selection = this.dependencies.selection.getEditableSelection();
+        this.offset = selection.startOffset - 1;
+        this.enabledCommands = this.dependencies.powerbox.getAvailablePowerboxCommands();
+        this.dependencies.powerbox.openPowerbox({
+            commands: this.enabledCommands,
+            categories: this.categories,
+            onApplyCommand: this.historySavePointRestore,
+            onClose: () => {
+                this.shouldUpdate = false;
+                this.powerButtonAnchorEl = false;
+            },
+        });
+        this.shouldUpdate = true;
+    }
+    /**
+     * @deprecated
+     */
     openPowerbox() {
         const selection = this.dependencies.selection.getEditableSelection();
         this.offset = selection.startOffset - 1;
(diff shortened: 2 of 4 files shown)


addons/html_editor/static/src/main/power_buttons_plugin.js
```diff
@@ -173,5 +173,6 @@ export class PowerButtonsPlugin extends Plugin {
         btns.forEach((btn) => btn.classList.add("disabled"));
         await commandFn();
         btns.forEach((btn) => btn.classList.remove("disabled"));
+        this.dependencies.selection.focusEditable();
```
deso-odoo: So currently the focus goes to button and then you bring it back to editable, how about you never let button have the focus by preventing default the pointerdown.

addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
```diff
@@ -15,6 +18,21 @@ export class SearchPowerboxPlugin extends Plugin {
         delete_handlers: this.update.bind(this),
         post_undo_handlers: this.update.bind(this),
         post_redo_handlers: this.update.bind(this),
+        user_commands: {
```
deso-odoo, line 21: On principle I don't even like to have a separate `SearchPowerboxPlugin` to handle searching in powerbox. That being said I'd like to hear from final reviewer.

addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
```diff
@@ -15,6 +18,21 @@ export class SearchPowerboxPlugin extends Plugin {
...
+            id: "openPowerbox",
+            run: () => {
+                const selection = this.dependencies.selection.getEditableSelection();
+                this.historySavePointRestore = this.dependencies.history.makeSavePoint();
+                // Anchor element for powerbox opened via power buttons.
+                this.powerButtonAnchorEl = closestElement(selection.anchorNode);
```
deso-odoo, line 27: This should not work if you go on implementing what I suggested in commit 1.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        await expectElementCount(".o-we-powerbox", 1);
+        expect(editor.document.activeElement).toBe(el);
+    });
+
+    test("should filter the powerbox contents based on the search term", async () => {
+        const { el, editor } = await setupEditor("<p>[]<br></p>");
+        // Open powerbox via the More options button
+        click(".o_we_power_buttons .power_button.fa-ellipsis-v");
+        await animationFrame();
+        await expectElementCount(".o-we-powerbox", 1);
```
deso-odoo: `animationFrame` should not be need alongside `expectElementCount`.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        expect(editor.document.activeElement).toBe(el);
```
deso-odoo: You already tested this in test case above this, its okay to not test it again.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        // All commands should be available
+        await animationFrame();
```
deso-odoo: This should also not be needed.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        // All commands should be available again
+        expect(queryAllTexts(".o-we-command-name").length).toBe(28);
+    });
+
+    test("should close the powerbox on pointerdown outside and not reopen it on subsequent keydown", async () => {
+        const { editor } = await setupEditor("<p>[]<br></p>");
+        // Open powerbox via the More options button
+        click(".o_we_power_buttons .power_button.fa-ellipsis-v");
+        await animationFrame();
+        await expectElementCount(".o-we-powerbox", 1);
```
deso-odoo: same as above.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        // Click outside the powerbox
+        await pointerDown("p");
+        await animationFrame();
+        await expectElementCount(".o-we-powerbox", 0);
```
deso-odoo: same as above.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        // Typing should not reopen the powerbox
+        await insertText(editor, "a");
+        await animationFrame();
+        await expectElementCount(".o-we-powerbox", 0);
```
deso-odoo: `expectElementCount` with 0 should not be used in such case because we don't have powerbox opened, so waitForNone within `expectElementCount` would not actually wait, here it works because of animationFrame above.

addons/html_editor/static/tests/power_buttons.test.js
```diff
@@ -172,11 +172,137 @@ describe("buttons", () => {
...
+        expect(queryAllTexts(".o-we-command-name")).toEqual([
+            "Heading 1",
+            "Heading 2",
+            "Heading 3",
+        ]);
+        undo(editor);
+        await animationFrame();
+        await expectElementCount(".o-we-powerbox", 1);
     });
 });
```
deso-odoo, line 306: All this tests but you never test whether command is applied or not!

addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
dhba-odoo: No, it will not be affected by the changes in commit 1, as it is simply the anchor element from which the powerbox is opened.

deso-odoo: By `powerButtonAnchorEl` I thought of it as powerButton element. ^^'

deso-odoo: Seeing this again I don't think you even need this, you already have `this.offset` can't we use that.
change `this.offset = selection.startOffset - 1` to `this.offset = selection.startOffset` and remove the check of `this.searchNode.nodeValue[this.offset] === "/"` in `isSearching`. Am I missing anything?

dhba-odoo: There is an edge case when opening the powerbox via the power buttons. Typing **heading** correctly filters the commands, but removing the entire word (when the last “h” is deleted) closes the powerbox instead of showing all commands, unlike the **/heading** flow.

I am not fully sure which behavior is functionally better. In the **/heading** case, this behavior makes sense because removing `/` closes the powerbox. However, in the case of power buttons, there is no `/` to indicate that removing it should close the powerbox.

So to keep both flows consistent, I have implemented in a way so that removing the search term shows all powerbox commands instead of closing the powerbox.

deso-odoo: > There is an edge case when opening the powerbox via the power buttons. Typing heading correctly filters the commands, but removing the entire word (when the last “h” is deleted) closes the powerbox instead of showing all commands, unlike the /heading flow.

That is because in `isSearching` we have a check for `this.searchNode.nodeValue` and that shall be undefined in your case. Shouldn't removing that check and updating `searchTerm = this.searchNode.nodeValue?.slice(this.offset, selection.endOffset);` in `update` work?

dhba-odoo: No, this will not work for undo and redo. For example, if I open the powerbox using the power button and then perform an undo, `this.searchNode` is updated before `isSearching` is called. As a result, inside `isSearching`, the endContainer matches the `this.searchNode`, and since the offset will be 0 or greater, the condition passes, therefore, the powerbox does not get closed.

addons/html_editor/static/src/main/powerbox/search_powerbox_plugin.js
dmo-odoo: If I'm not mistaken, the Powerbox is the dropdown itself, while the SearchPowerbox filters the dropdown. From the top of my head, I know only one usage of the Powerbox that does not filter, it's when you paste an url that points to an image. The editor opens a powerbox with two options: to embed the image in the document or to paste as url. So that's the reason. I guess technically it could be a simple dropdown, but it would be reusing the style of the powerbox anyway.

With that being said, it's a bit weird to have the "openPowerbox" command in `SearchPowerboxPlugin`. If it's not used anywhere else (I think the paste image url example does not call the command), then I would suggest renaming it to "openSearchPowerbox", but we're in stable so we'll need to duplicate it and have both "openPowerbox" and "openSearchPowerbox" until master. In fact it doesn't make sense to have a user command to open the powerbox without being able to search in it. It's only useful from the system perspective (the image url example), but from a user command perspective, you always want to search in the powerbox. So in master we would remove "openPowerbox" and only keep "openSearchPowerbox". What do you think?

dhba-odoo: @dmo-odoo I have created a new method, `openSearchPowerbox`, in `searchPowerboxPlugin` and replaced the previous usage of `openPowerbox` as you have mentioned.

PR #244455, opened 2026-01-19
[IMP] point_of_sale: toggle preset control when only two presets are configured

jipr-odoo:
In this commit:
===============
When only two presets are configured on a POS config, skip the preset selection popup and make the preset button work as a toggle between the two presets.

Task-5491093


davidmonnom (approved):
@robodoo r+

PR #244460, opened 2026-01-19
[FIX] website_slide: update quiz answer tags without page reload

jeromebelpaire:
After adding new answers or editing existing ones in a quiz question
and clicking 'Save & Close', the question overview displays incorrect
data. New answers appear as empty tags, and modified answers continue
to show their old values. The correct data only appears after a full
page reload.

This regression was introduced in commit 8ceea093, where invisible
fields were removed during code cleanup. The `display_name` field
is required in the `answer_ids` list view for the web client to
correctly update its local cache. Without this field, the client
cannot refresh the display names of the tags immediately after
modification.

This commit restores the `display_name` field as `column_invisible`
in the `slide.question` form view.

Task-5449335

### DIFF
--- a/addons/website_slides/views/slide_question_views.xml
+++ b/addons/website_slides/views/slide_question_views.xml
@@ -18,6 +18,7 @@
                     </div>
                     <field name="answer_ids">
                         <list editable="bottom" create="true" delete="true">
+                            <field name="display_name" column_invisible="True"/> <!-- 'display_name' is necessary for the many2many_tags to work on slide_slide_views -->
                             <field name="text_value"/>
                             <field name="is_correct"/>
                             <field name="comment"/>


addons/website_slides/views/slide_question_views.xml
```diff
@@ -18,6 +18,7 @@
                     </div>
                     <field name="answer_ids">
                         <list editable="bottom" create="true" delete="true">
+                            <field name="display_name" column_invisible="True"/>
```
julien-banken: To pass all the tests, I think you have to add a small comment on top of that line to explain why you need to load the display name. 😃

Example in `addons/event/views/event_question_views.xml`:
```xml
<list editable="bottom">
    <!-- 'display_name' is necessary for the many2many_tags to work on the event view -->
    <field name="display_name" column_invisible="True" />
```

jeromebelpaire: Aha, I was waiting on that kind of information, thanks! 😃

jeromebelpaire: I added a comment and the runbot still raises an error, do you know what else I need to do? 🤔

julien-banken: Reading `test_views.py`, I think you actually have to add the comment right after the field on the same line.

```suggestion
                            <field name="display_name" column_invisible="True"/> <!-- 'display_name' is necessary for the many2many_tags to work on slide_slide_views -->
```

jeromebelpaire: That worked, thanks!

addons/website_slides/tests/test_slide_views.py
```diff
@@ -0,0 +1,12 @@
+from lxml import etree
+from odoo.tests import tagged, TransactionCase
+
+
+@tagged("at_install", "-post_install")
```
tde-banana-odoo: Why specifically at_install ?

addons/website_slides/tests/test_slide_views.py
```diff
@@ -0,0 +1,12 @@
...
+class TestSlideViews(TransactionCase):
+    def test_slide_question_view_form_has_display_name(self):
```
tde-banana-odoo: That being said, weird test ... would say it is ok not to have a test for this one. We can also merge the fix and hope people won't break invisible fields again ^^.

addons/website_slides/tests/test_slide_views.py
jeromebelpaire: No good reason actually, I inspired myself from other tests of survey which had the tags, but it is indeed not really necessary.

addons/website_slides/tests/test_slide_views.py
jeromebelpaire: Ok, I'll remove the test, I agree it might be a bit too much/niche, but flch has drilled me to write a test for every fix 😄

tde-banana-odoo: True jedi power is to know when not to write a test !

tde-banana-odoo:
@robodoo r+

PR #244463, opened 2026-01-19
[FIX] im_livechat: do not fetch expertise if not needed

tsm-odoo:
Before this commit, the user's live chat expertise were always returned as part of `_init_store_data`. However, this led to a warning in portal tours because the expertises are inserted into the store while the live chat models are not loaded (they are not needed).

We shouldn't return useless data. This commit fixes this issue by fetching the data when they will be needed (i.e. when the discuss app is opened with a live chat user).

runbot-234137

Current behavior before PR:

---

### DIFF
--- a/addons/im_livechat/controllers/webclient.py
+++ b/addons/im_livechat/controllers/webclient.py
@@ -37,6 +37,8 @@ class WebClient(WebclientController):
                 return
             fields_to_store = channel._get_livechat_session_fields_to_store()
             store.add(channel, fields=fields_to_store)
+        if name == "/im_livechat/fetch_self_expertise":
+            store.add(request.env.user, Store.Many("livechat_expertise_ids", ["name"]))
 
     @classmethod
     def _process_request_for_all(self, store: Store, name, params):
--- a/addons/im_livechat/static/src/core/web/store_service_patch.js
+++ b/addons/im_livechat/static/src/core/web/store_service_patch.js
@@ -9,6 +9,7 @@ const storePatch = {
     setup() {
         super.setup(...arguments);
         this.livechatChannels = this.makeCachedFetchData("im_livechat.channel");
+        this.livechatSelfExpertises = this.makeCachedFetchData("/im_livechat/fetch_self_expertise");
         this.has_access_livechat = false;
     },
     /**
@@ -18,6 +19,7 @@ const storePatch = {
         super.onStarted(...arguments);
         if (this.discuss.isActive && this.has_access_livechat) {
             this.livechatChannels.fetch();
+            this.livechatSelfExpertises.fetch();
         }
     },
     /** @returns {boolean} Whether the livechat thread changed. */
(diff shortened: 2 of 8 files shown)


addons/im_livechat/controllers/webclient.py
```diff
@@ -76,3 +76,5 @@ def _process_request_for_all(self, store: Store, name, params):
                     request.env.ref("im_livechat.action_report_livechat_conversation", False),
                 ),
             )
+        if name == "/im_livechat/fetch_self_expertises":
```
alexkuhn: Why not in `_process_request_for_internal_user` instead of `_process_request_for_all`?

addons/im_livechat/static/src/core/web/store_service_patch.js
```diff
@@ -9,6 +9,9 @@ const storePatch = {
     setup() {
         super.setup(...arguments);
         this.livechatChannels = this.makeCachedFetchData("im_livechat.channel");
+        this.livechatSelfExpertises = this.makeCachedFetchData(
+            "/im_livechat/fetch_self_expertises"
+        );
```
alexkuhn: Do we really need another cached fetch data? This looks like the coverage of this data to fetch is the same as `livechatChannels`, so we could just condition data fetching with `im_livechat.channel`/`livechatChannels` too.

tsm-odoo: They are batched together anyway, it's better to have a clear scope IMO. And I don't like this shortcut. For example, when opening the command palette, we fetch the live chat channels. It doens't mean we have to fetch the expertises as well. 

In the future, one could change the condition to fetch the channels, thus preventing the fetch of expertises as well without noticing.

addons/im_livechat/controllers/webclient.py
tsm-odoo: Updated, it should indeed be in the internal user func.

addons/im_livechat/static/src/core/web/store_service_patch.js
alexkuhn: I'd prefer to combine data when they have 100% shared flow and split only when differences are made obvious in code. But I get that `"livechatChannels"` is named in a way that can be hard to grasp expertises are also fetched there, and since this is diff in stable we don't have the luxury to rename.

It's a bit sad to have to define a named cached fetch data + add LOC of fetch + intercept named fetched data in python, etc. when all of this could have been a single LOC of `"im_livechat.channel"` cached data to enrich with expertise ids.

Anyway, not a big deal, it's alright.

addons/im_livechat/controllers/webclient.py
```diff
@@ -37,6 +37,8 @@ def _process_request_for_internal_user(self, store: Store, name, params):
                 return
             fields_to_store = channel._get_livechat_session_fields_to_store()
             store.add(channel, fields=fields_to_store)
+        if name == "/im_livechat/fetch_self_expertises":
```
alexkuhn: "expertises" is grammatically incorrect: plural of expertise is expertise since this is uncountable.
I would not bother with attempt to plural here.

alexkuhn (approved):
LGTM.
I saw runbot red but I wonder if this is fully solved by internal_users branching from recent change or not.

@robodoo delegate+

PR #244464, opened 2026-01-19
[IMP] setup: auto setup Windows IoT Box firewall rules

loouis-t:
We now automatically add the firewall exception for inbound TCP connections on ports 80,443 and 8069, to avoid manual configuration.

Task: 3748708

### DIFF
--- a/setup/package.py
+++ b/setup/package.py
@@ -428,7 +428,7 @@ class DockerIot(DockerWine):
         self.nt_service_name = "odoo-iot"
 
     def build_image(self):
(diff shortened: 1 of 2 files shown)


Ysoroko (changes requested):
Shouldn't we also allow
1) 9000 for older Ingenico integrations (if we assume 19.2 virtual iot will be compatible with older versions)
2) 9050 for Worldline
3) 7784/33334 for Six

Ysoroko (approved):
tested the installer, an outbound rule for the specified ports is created in Windows firewall settings as expected

PR #244470, opened 2026-01-19
[IMP] website_forum: remove search_count from forum pager

JKE-be:
In production, the search count take more than 100ms while it is only necessary to compute the number of page...

To improve performance, the search count is removed and the pager now displays a scope of 5 pages around the current one.

The existing `scope` parameter is reused to handle this behavior. In stable, a negative value is used to explicitly indicate that the parameter should be reused.

### DIFF
--- a/addons/portal/controllers/portal.py
+++ b/addons/portal/controllers/portal.py
@@ -29,12 +29,16 @@ def pager(url, total, page=1, step=30, scope=5, url_args=None):
     - Shows current page with -1 and +1 neighbors
     - Adds ellipses when necessary
 
+    In version 19 (current stable), you need to use a negative value for the scope parameter if you
+    want to have a fixed number of pages displayed without the first / last page and the ellipses.
+    We use the absolute value of parameter scope.
+
     :param str url : base url of the page link
     :param int total : number total of item to be splitted into pages
     :param int page : current page
     :param int step : item per page
     :param int scope : number of page to display on pager
-    :param dict url_args : additionnal parameters to add as query params to page url
+    :param dict url_args : additional parameters to add as query params to page url
     :returns dict
     """
     # Compute Pager
@@ -52,14 +56,34 @@ def pager(url, total, page=1, step=30, scope=5, url_args=None):
         return _url
 
     # Build page list based on conditions
-    if page_count <= 5:
-        page_list = list(range(1, page_count + 1))
-    elif page <= 3:
-        page_list = [1, 2, 3, 4, "…", page_count]
-    elif page >= page_count - 2:
-        page_list = [1, "…"] + list(range(page_count - 3, page_count + 1))
+    if scope < 0:
+        # To distinguish from the previous behavior, negative scope means fixed scope while positive scope
+        # means old leftover scope. From this way we don't change behavior of current running databases.
+        scope = abs(scope)
+
+        start = page - (scope // 2)
+        end = start + scope - 1
+
+        if start < 1:
+            start = 1
+            end = min(scope, page_count)
+
+        if end > page_count:
+            end = page_count
+            start = max(1, page_count - scope + 1)
+        start = int(start)
+        end = int(end) + 1
+
+        page_list = list(range(start, end))
     else:
-        page_list = [1, "…", page - 1, page, page + 1, "…", page_count]
+        if page_count <= 5:
+            page_list = list(range(1, page_count + 1))
+        elif page <= 3:
+            page_list = [1, 2, 3, 4, "…", page_count]
+        elif page >= page_count - 2:
+            page_list = [1, "…"] + list(range(page_count - 3, page_count + 1))
+        else:
+            page_list = [1, "…", page - 1, page, page + 1, "…", page_count]
 
     pages = [
         {"num": p, "url": get_url(p) if p != "…" else None, "is_current": p == page}
--- a/addons/website_forum/controllers/website_forum.py
+++ b/addons/website_forum/controllers/website_forum.py
@@ -90,7 +90,7 @@ class WebsiteForum(WebsiteProfile):
 
     def _get_forum_post_search_options(self, forum=None, tag=None, filters=None, my=None, create_uid=False, include_answers=False, **post):
         return {
-            'allowFuzzy': not post.get('noFuzzy'),
+            'allowFuzzy': False,
             'create_uid': create_uid,
             'displayDescription': False,
             'displayDetail': False,
@@ -141,11 +141,18 @@ class WebsiteForum(WebsiteProfile):
             **post
         )
 
+        pager_scope = 10
+
+        options['per_page'] = self._post_per_page
+        options['scope'] = pager_scope
+        options['current_page'] = page
+
         slug = request.env['ir.http']._slug
         question_count, details, fuzzy_search_term = request.website._search_with_fuzzy(
             "forum_posts_only", search, limit=page * self._post_per_page, order=sorting, options=options)
+
         question_ids = details[0].get('results', Post)
-        question_ids = question_ids[(page - 1) * self._post_per_page:page * self._post_per_page]
+        question_ids = question_ids[:self._post_per_page]
 
         if not forum:
             url = '/forum/all'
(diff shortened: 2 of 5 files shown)


JKE-be:
With this commit
```
[19/Jan/2026 11:28:53] "GET /forum/help-1 HTTP/1.0" 200 - 56 0.067 0.165
[19/Jan/2026 11:28:53] "GET /forum/help-1 HTTP/1.0" 200 - 56 0.067 0.163
[19/Jan/2026 11:28:54] "GET /forum/help-1 HTTP/1.0" 200 - 56 0.066 0.163
```

Before
```
[19/Jan/2026 11:31:15] "GET /forum/help-1 HTTP/1.0" 200 - 57 0.757 0.173
[19/Jan/2026 11:31:16] "GET /forum/help-1 HTTP/1.0" 200 - 57 0.768 0.173
[19/Jan/2026 11:31:17] "GET /forum/help-1 HTTP/1.0" 200 - 57 0.758 0.172
```

addons/portal/controllers/portal.py
```diff
@@ -52,14 +52,37 @@ def get_url(page):
...
-    elif page <= 3:
-        page_list = [1, 2, 3, 4, "…", page_count]
-    elif page >= page_count - 2:
-        page_list = [1, "…"] + list(range(page_count - 3, page_count + 1))
+    if scope < 0:
+        # In v19, scope has been unused and ignored until now.
+        # We reuse this parameter to allow to supply a fixed scope (number of pages to display around current page)
+        # To distinguish from the previous behavior, negative scope means fixed scope while positive scope
+        # means old leftover scope. From this way we don't change behavior of current running databases.
+        PAGER_SIZE = (-scope * 2) + 1
```
pyduf: Should we not change the docstring (:param int scope : number of page to display on pager) ? As actually we display “scope” page before the current page and “scope” page after, no ?

addons/website_forum/controllers/website_forum.py
```diff
@@ -140,13 +140,22 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
...
         )
+        options['allowFuzzy'] = False
+        fuzzy_search_term = False
+        pager_scope = 5
+
+        search_details = request.website._search_get_details("forum_posts_only", sorting, options)
+        question_count, details = request.website._search_exact(
+            search_details,
+            search,
+            (page + (2 * pager_scope)) * self._post_per_page,
```
pyduf: Maybe this formula is a little better and maybe easier to understand:
```python
# Getting enough record +1 for displaying page_scope before the page and after the page (if possible)
max(page + pager_scope, 2 * pager_scope) * self._post_per_page + 1
```
What do you think ?

addons/website/models/website.py
```diff
@@ -1512,7 +1512,7 @@ def get_template(self, template):
         return self.env['ir.ui.view']._get_template_view(template).sudo()
 
     @api.model
-    def pager(self, url, total, page=1, step=30, scope=5, url_args=None):
+    def pager(self, url, total, page=1, step=30, scope=1, url_args=None):
```
pyduf: Why modifying scope to 1 (as it is not used for positive value) ?

addons/website_forum/models/forum_post.py
```diff
@@ -923,6 +923,18 @@ def _search_render_results(self, fetch_fields, mapping, icon, limit):
...
+    def _search_fetch(self, search_detail, search, limit, order):
+        fields = search_detail['search_fields']
+        base_domain = search_detail['base_domain']
+        domain = self._search_build_domain(base_domain, search, fields, search_detail.get('search_extra'))
+        results = self.search(
+            domain,
+            limit=limit,
+            order=search_detail.get('order', order)
+        )
+        return results, len(results)
```
pyduf: Actually it override the method of the mixin (website.searchable.mixin) and do the same except the search_count. Should we not add a docstring to explain that ?

addons/website_forum/controllers/website_forum.py
```diff
@@ -140,13 +140,22 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
             my_profile=request.env.user == author,
             **post
         )
+        options['allowFuzzy'] = False
+        fuzzy_search_term = False
+        pager_scope = 5
+
+        search_details = request.website._search_get_details("forum_posts_only", sorting, options)
+        question_count, details = request.website._search_exact(
```
pyduf: We lose some feature by switching from _search_with_fuzzy to _search_exact. For example, if you search on master for “ecomerce” in the “help” forum you find a post while you don’t in the new version. Is that a lot slower with _search_with_fuzzy (which seems to perform a search_exact under the hood) ?

addons/website_forum/controllers/website_forum.py
```diff
@@ -162,7 +171,7 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
 
         pager = tools.lazy(lambda: request.website.pager(
             url=url, total=question_count, page=page, step=self._post_per_page,
-            scope=5, url_args=url_args))
+            scope=-pager_scope, url_args=url_args))  # scope broken, never used
```
pyduf: Maybe we could change the comment to something like: “Using a negative scope to trigger the optimized search for the pager in stable”. And add a comment in the pager docstring explaining that the scope is not used beside triggering the optimisation.

pyduf (commented):
Hello @JKE-be , 

Thanks for your PR. I have left you some comments. 

My main concern is the switch between _search_with_fuzzy to _search_exact as it makes the search feature less effective (no result for "ecomerce" for example while on master it returns post with "eCommerce"). 
Is switching back to fuzzy without the search_count is an option ?

addons/website/models/website.py
JKE-be: it was preview for master, but I reverted, to respect the current paging system, you will have scope of 1 + border...
But yes, to decide in master, as event the value of scope is not sure (around, or total of page)

addons/portal/controllers/portal.py
JKE-be: adapted the code to respect the same meaning.

scope of 10 = 10 pages in pager (similar to google) 5 left / 4 right

Optimized the extra part to fill in the 10 pages if needed

addons/website_forum/controllers/website_forum.py
JKE-be: Yes, that is exactly the current purpose of this PR: to remove the fuzzy logic, not only the search_count.

FP and AL want to keep it only in ecommerce / product for now.

It is a performance killer, especially with this number of posts.

addons/website_forum/controllers/website_forum.py
JKE-be: already refactored to keep the scope = total page

`extra_fill_scope = max((pager_scope / 2), pager_scope - page)`

JKE-be (commented):
thanks for review, I updated/commented

addons/website_forum/controllers/website_forum.py
```diff
@@ -140,13 +140,23 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
             my_profile=request.env.user == author,
             **post
         )
+        options['allowFuzzy'] = False
+        fuzzy_search_term = False
```
pyduf: This seems to be specific to the _search_with_fuzzy function that we drop so maybe we can get rid of them ?

addons/website_forum/controllers/website_forum.py
```diff
@@ -162,7 +172,7 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
 
         pager = tools.lazy(lambda: request.website.pager(
             url=url, total=question_count, page=page, step=self._post_per_page,
-            scope=5, url_args=url_args))
+            scope=-pager_scope, url_args=url_args))  # scope negative, to force to use it ()in stable)
```
pyduf: Typo

pyduf (commented):
Hello @JKE-be , 

Thanks for the update, it seems to work well. 
I wonder if there are other cases where the forum is searched. For example, you can search it through the global search of the website. But although it displays the older pager, it seems to work well as it is limited to 50 items per page and 500 (see my note below).

Maybe you can also update the commit message which says “now displays
a scope of 5 pages around the current one” to the new version. Also I wonder, is there a task reference ?

Note: 
–	I thought that overriding  _search_fetch in forum_post with an implementation that doesn’t follow the spec of the mixin could lead to some bugs in the full search of the website but it doesn’t seems the case as that search is limited to 500 results.
–	I haven’t test the speedup as locally even with 2000 records, it’s the same. Do you know how much record is needed to feel the difference ?

PR #192953, opened 2025-01-08
[IMP] mail: canned response with :: instead of :

alexkuhn:
Before this commit, delimiter to use canned response was ":". This puts some conflicts with emoji substitution, which has been minimized a bit by not putting the focus on the suggestion list of emojis.

However this is still bothersome, so it's worth changing it slightly. This PR minimises the impact of change by replacing it to "::", which is relatively close to the former ":".

Since the discovery of the feature was only through typing ":", we added a new item in the "+" button of composer to insert a new canned response, which simply adds the "::" in the composer to help select the canned response. This composer action only shows the shortcut to trigger canned response: it let the user learns of this feature, and in case the delimiter changes like with this PR, the user can relatively easily re-learn the new trigger thanks to this composer action.

The "Insert a Canned response" composer action is visible when the user can at least find a canned response. Also the canned response suggestions are focused by default, now that they do not conflict with emoji substitution.

Task-4110505

https://github.com/odoo/enterprise/pull/77184

Jan-14-2025 16-15-44

PR #170900, opened 2024-06-26
[IMP] mail: avoid emoji conflicts with canned responses

tsm-odoo:
Canned responses are triggered by the ":" shortcut, which can be cumbersome as it conflicts with emojis. This also leads to mistakes since the first canned response in the list is automatically focused, causing the insertion of the canned response when pressing enter to send a message.

This PR addresses this usability issue by removing the auto-focus behavior for canned response suggestions. Users can still easily access canned responses using the same shortcut, but the risk of inserting a canned response by mistake is significantly reduced.

task-356389

PR #244472, opened 2026-01-19
[FIX] im_livechat: non deterministic canned response test

tsm-odoo:
In [1], auto focus was removed for canned responses in the navigable list. The intent was to avoid conflicts with emojis.

In [2], the shortcut for canned response was changed to `::` and the auto focus behavior was restored.

However, tests added in [1] were not removed. Since they were fragile (negative assertion), they are passing most of the time.

This PR removes the tests that are now outdated.

runbot-233358

[1]: https://github.com/odoo/odoo/pull/170900
[2]: https://github.com/odoo/odoo/pull/192953

Current behavior before PR:

---

### DIFF
--- a/addons/mail/static/tests/composer/composer.test.js
+++ b/addons/mail/static/tests/composer/composer.test.js
@@ -1014,32 +1014,6 @@ test("Canned response last used changes on posting", async () => {
     expect(cannedResponse.last_used).not.toBeEmpty();
 });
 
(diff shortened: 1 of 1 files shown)


alexkuhn (approved):
@robodoo delegate+
Should re-target branch to 18.2.

alexkuhn:
@robodoo retry

PR #244477, opened 2026-01-19
[FIX] hr_attendance: prevent multiple check-ins/outs and geolocation hang

vega-odoo:
when using signInOut with geolocation, slow or temporarily unavailable network connections could cause getCurrentPosition to hang indefinitely (default timeout is infinite). This led to:

- Frontend not updating, allowing multiple clicks and creating duplicate attendance entries
- Incorrect check-in/check-out data

__Steps to reproduce:__
1. check in while online and server reachable
2. disconnect network or make server unreachable
3. check out

Currently, getCurrentPosition would hang indefinitely. till the network is restored. then it will trigger the rpc call much later than the action time. in the meantime, the user could click multiple times, creating multiple attendance records.

With this fix, getCurrentPosition will timeout after 10 seconds, then it will proceed without position. and if the server is unreachable, it will show an error notification without allowing multiple clicks.

__FIX__
- Adds a timeout to getCurrentPosition
- Uses a `_attendanceInProgress` flag to prevent multiple clicks
- Ensures only the first callback (success or error) triggers the RPC

opw-5414044

### DIFF
--- a/addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
+++ b/addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
@@ -8,6 +8,8 @@ import { deserializeDateTime } from "@web/core/l10n/dates";
 import { registry } from "@web/core/registry";
 import { useService } from "@web/core/utils/hooks";
 import { isIosApp } from "@web/core/browser/feature_detection";
+import { ConnectionLostError } from "@web/core/network/rpc_service";
+import { _t } from "@web/core/l10n/translation";
 const { DateTime } = luxon;
 
 export class ActivityMenu extends Component {
@@ -19,6 +21,7 @@ export class ActivityMenu extends Component {
         this.rpc = useService("rpc");
         this.ui = useState(useService("ui"));
         this.userService = useService("user");
+        this.notification = useService("notification");
         this.employee = false;
         this.state = useState({
             checkedIn: false,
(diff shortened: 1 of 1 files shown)


addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
```diff
@@ -55,27 +58,57 @@ export class ActivityMenu extends Component {
     async signInOut() {
         // to close the dropdown
         document.body.click()
-        // iOS app lacks permissions to call `getCurrentPosition`
-        if (!isIosApp()) {
+
+        if(this._attendanceInProgress){
```
mepe-odoo, line 87: I don't understand why you need this variable :thinking:

mepe-odoo (changes requested):

addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
vega-odoo: Hello Hello,

`_attendanceInProgress` prevents multiple executions of `signInOut()` while a previous request is still in progress, avoiding duplicate RPC calls or geolocation requests if the user clicks rapidly. Without it, a user could click 5 times in a row, each time calling `navigator.geolocation.getCurrentPosition`, even though called protects the internal callback.

But  `called` variable is optional and only adds extra safety to ensure `finalize()` is not triggered more than once by geolocation if it misbehaves. it could be removed tho

addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
```diff
@@ -55,27 +58,57 @@ export class ActivityMenu extends Component {
...
+            return;
+        }
+        this._attendanceInProgress = true;
+
+        const finalize = async (latitude = false, longitude = false) => {
```
mepe-odoo: this method name is a bit strange :/ maybe you can rename it into checking and define it like a regular method for more clarity :D

addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.js
mepe-odoo: I think we can remove called it's a bit difficult to understand what happens with many try catch and finally :)

mepe-odoo (changes requested):
thanks <3

vega-odoo:
Hello @mepe-odoo 

I've applied the changes, Could u have a look?
Thanks!

mepe-odoo:
@robodoo r+

PR #244478, opened 2026-01-19
[PERF] html_editor: reduce layout thrashing

deso-odoo:

I. The power buttons positioning logic was interleaving DOM writes and layout reads during selectionchange, causing repeated style/layout recalculations. This PR reorders the logic so geometry is read first and DOM mutations are
applied afterwards, reducing the number of forced reflows and significantly improving performance.

II. Debounce `updateHints` and `updatePowerButtons` to avoid excessive UI updates on frequent selection changes. Introduce `debounceHints` and `debouncePowerButtons` editor config options so debouncing can be disabled in tests for deterministic behavior.

III. Introduce READ helper for withSequence to explicitly order resource handlers so DOM reads run before DOM mutations.

Before:
image

After:
image

task-5499625

---

### DIFF
--- a/addons/html_editor/static/src/main/hint_plugin.js
+++ b/addons/html_editor/static/src/main/hint_plugin.js
@@ -3,13 +3,14 @@ import { isEditorTab, isEmptyBlock, isProtected } from "@html_editor/utils/dom_i
 import { removeClass } from "@html_editor/utils/dom";
 import { descendants, selectElements } from "@html_editor/utils/dom_traversal";
 import { closestBlock } from "../utils/blocks";
+import { debounce } from "@web/core/utils/timing";
 
 export class HintPlugin extends Plugin {
     static id = "hint";
     static dependencies = ["history", "selection"];
     resources = {
         /** Handlers */
-        selectionchange_handlers: this.updateHints.bind(this),
+        selectionchange_handlers: this.triggerDebouncedUpdateHints.bind(this),
         external_history_step_handlers: () => {
             this.clearHints();
             this.updateHints();
@@ -31,6 +32,12 @@ export class HintPlugin extends Plugin {
 
     setup() {
         this.updateHints(this.editable);
+        const shouldDebounce = this.config.debounceHints !== false;
+        if (shouldDebounce) {
+            this.debouncedUpdateHints = debounce(this.updateHints.bind(this), 30);
+        } else {
+            this.debouncedUpdateHints = this.updateHints.bind(this);
+        }
     }
 
     destroy() {
@@ -41,6 +48,11 @@ export class HintPlugin extends Plugin {
     normalize() {
         this.clearHints();
         this.updateHints();
+    }
+
+    triggerDebouncedUpdateHints() {
+        this.clearHints();
+        this.debouncedUpdateHints();
     }
 
     /**
--- a/addons/html_editor/static/src/main/power_buttons_plugin.js
+++ b/addons/html_editor/static/src/main/power_buttons_plugin.js
@@ -4,6 +4,7 @@ import { closestBlock } from "@html_editor/utils/blocks";
 import { isEditorTab, isEmptyBlock } from "@html_editor/utils/dom_info";
 import { closestElement, descendants } from "@html_editor/utils/dom_traversal";
 import { omit, pick } from "@web/core/utils/objects";
+import { debounce } from "@web/core/utils/timing";
 
 /** @typedef {import("./powerbox/powerbox_plugin").PowerboxCommand} PowerboxCommand */
 
@@ -54,7 +55,7 @@ export class PowerButtonsPlugin extends Plugin {
     ];
     resources = {
         layout_geometry_change_handlers: this.updatePowerButtons.bind(this),
-        selectionchange_handlers: this.updatePowerButtons.bind(this),
+        selectionchange_handlers: this.triggerDebouncedUpdatePowerButtons.bind(this),
         post_mount_component_handlers: this.updatePowerButtons.bind(this),
     };
 
@@ -63,6 +64,12 @@ export class PowerButtonsPlugin extends Plugin {
             "oe-power-buttons-overlay"
         );
         this.createPowerButtons();
+        const shouldDebounce = this.config.debouncePowerbuttons !== false;
+        if (shouldDebounce) {
+            this.debouncedUpdatePowerButtons = debounce(this.updatePowerButtons.bind(this), 30);
+        } else {
+            this.debouncedUpdatePowerButtons = this.updatePowerButtons.bind(this);
+        }
     }
 
     createPowerButtons() {
@@ -104,6 +111,11 @@ export class PowerButtonsPlugin extends Plugin {
         this.powerButtonsContainer.className = `o_we_power_buttons d-flex justify-content-center invisible position-absolute`;
         this.powerButtonsContainer.append(...this.descriptionToElementMap.values());
         this.powerButtonsOverlay.append(this.powerButtonsContainer);
+    }
+
+    triggerDebouncedUpdatePowerButtons() {
+        this.powerButtonsContainer.classList.add("invisible");
+        this.debouncedUpdatePowerButtons();
     }
 
     updatePowerButtons() {
--- a/addons/html_editor/static/src/main/power_buttons_plugin.js
+++ b/addons/html_editor/static/src/main/power_buttons_plugin.js
@@ -78,7 +78,7 @@ export class PowerButtonsPlugin extends Plugin {
             const btn = this.document.createElement("button");
             let className = "power_button btn px-2 py-1 cursor-pointer";
             if (icon) {
-                let iconLibrary = icon.includes("fa-") ? "fa" : "oi";
+                const iconLibrary = icon.includes("fa-") ? "fa" : "oi";
                 className += ` ${iconLibrary} ${icon}`;
             } else {
                 const span = this.document.createElement("span");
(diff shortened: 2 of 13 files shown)


addons/html_editor/static/src/main/hint_plugin.js
```diff
@@ -63,10 +63,15 @@ export class HintPlugin extends Plugin {
...
                         this.makeHint(target, nodeHint);
+                        hintsToApply.push({ target, nodeHint });
                     }
                 }
             }
         }
+        this.clearHints();
+        for (const { target, nodeHint } of hintsToApply) {
+            this.makeHint(target, nodeHint);
+        }
```
deso-odoo: Before:
image
After:
image

addons/html_editor/static/src/main/power_buttons_plugin.js
deso-odoo, line 1: For first commit along this, I could also debounce updatePowerButtons. But I'm not sure it will be liked from UX perspective. Because on first selectionchange the powerbuttons will be gone and on the last call it will directly be at the position it should be! Same for commit 2 for hints as well.

It's choosing between:
https://drive.google.com/file/d/1eYzb54pnH_zM6g3kGhaeM-zCXVt4-8bh/view
and
https://drive.google.com/file/d/1p4vw16-b2a6deQ8WRpfrUTN0XoNQLQJi/view

addons/html_editor/static/src/main/power_buttons_plugin.js
```diff
@@ -130,28 +129,27 @@ export class PowerButtonsPlugin extends Plugin {
...
-        this.editable.appendChild(clone);
-        const { width } = clone.getBoundingClientRect();
-        this.editable.removeChild(clone);
-        this.dependencies.history.enableObserver();
+        const hintText = block.getAttribute("o-we-hint-text");
+        const cs = getComputedStyle(block);
+        const canvas = this.canvas || (this.canvas = document.createElement("canvas"));
+        const ctx = canvas.getContext("2d");
+        ctx.font = cs.font;
+        const width = ctx.measureText(hintText).width;
```
dmo-odoo, line 163: Are you sure this is more performant? Recreating a new canvas all the time?

addons/html_editor/static/src/main/power_buttons_plugin.js
dmo-odoo: In the first one it looks like it is lagging behind the cursor, especially at the end. But maybe it's a video quirk? If it's not then I have a personnal preference for the one where it disappears and reappears at the end because I think it looks better than having it always lagging behind. But of course it's, once again, a functional decision... ^^'

addons/html_editor/static/src/core/selection_plugin.js
```diff
@@ -303,6 +303,7 @@ export class SelectionPlugin extends Plugin {
                 return;
             }
         }
+        this.dispatchTo("selectionchange_read_handlers", selectionData);
```
dmo-odoo: Good idea! Not sure about the name, though. I thought about `selectionchange_readonly_handlers` but then it kind of feels like you are not allowed to change _anything_, whereas it's the _DOM_ that should be readonly not your plugin or anything.

I also thought of something like this, which I don't know if it's a good idea or not:
```
/// Plugin 1:
resources: {
    selection_change_handlers: withSequence(READ, () => { /* reads stuff */ }),
}
/// Plugin 2:
resources: {
    selection_change_handlers: withSequence(WRITE, () => { /* reads stuff */ }),
}
```
Not sure I like it honestly, but it... exists.

The thing is, `selection_change_handlers` is probably not the only case where we might need to order function calls based on reads first, writes last, hence why I'm looking for a way to express this without duplicating the resources names for each case.

@FrancoisGe Any good idea since you've looked at similar performance issues before?

addons/html_editor/static/src/main/power_buttons_plugin.js
deso-odoo: Yes it's much more performant, also we're not creating a new canvas all the time we're reusing it.
Before:
image
After:
image

addons/html_editor/static/src/main/power_buttons_plugin.js
deso-odoo: > But maybe it's a video quirk?

No it's not a video quirk.

> I think it looks better than having it always lagging behind.

Agreed lets go with second one :+1:. I'll add debouncing.

addons/html_editor/static/src/core/selection_plugin.js
deso-odoo: I thought of `withSequence`, but usually I hate usage of `withSequence` for such cases with numbers. But I never thought of using something like READ/WRITE.

> Not sure I like it honestly, but it... exists.

I do like it!

But only issue I see with something like `withSequence(READ/WRITE)` is correctness relies on authors to classifying their handlers correctly, which I fear will be hard to maintain overtime.
eg:
```js
selectionchange_handlers: withSequence(READ, () => {
    abc(node);
});
```
and after few months someone adds `node.classList.add("foo")` in abc function. Nothing will break but it will no longer be read only.

you might say same could be true for `selectionchange_read_handlers`, but my only concern is I want some `selectionchange_handlers` to run before other so `read_handlers` would not be clear but maybe something like `pre_handlers`, A better name that suggest that this handlers should be run before handlers that updates dom.

> The thing is, selection_change_handlers is probably not the only case where we might need to order function calls based on reads first, writes last, hence why I'm looking for a way to express this without duplicating the resources names for each case.

Totally agreed, for me in a _perfect world_ all handlers wouldn’t directly mutate the DOM at all during dispatch. Instead, each handler would run in a read phase and queue its DOM mutations as callbacks, which would then be executed in a single write phase after all handlers have run.
eg:
something like
```js
dispatchTo(resourceId, ...args) {
    const handlers = this.getResource(resourceId);
    const writes = [];
    const context = {
        pushWrite(fn) {
            writes.push(fn);
        },
    };
    // READ PHASE
    for (const handler of handlers) {
        handler(...args, writes);
    }
    // WRITE PHASE
    for (const write of writes) {
        write();
    }
}

//and for handlers
resources = {
    selectionchange_handlers: (selectionData, { pushWrite }) => {
        // Code that reads the DOM.
        const block = closestBlock(selectionData.editableSelection.anchorNode);
        const rect = block.getBoundingClientRect();

        pushWrite(() => {
            // Code that mutates the DOM.
        });
    },
};
```
But that will be too much of abstraction, and I'm not even sure you'll like it. :sweat_smile:

FrancoisGe: > Good idea! Not sure about the name, though. I thought about `selectionchange_readonly_handlers` but then it kind of feels like you are not allowed to change _anything_, whereas it's the _DOM_ that should be readonly not your plugin or anything.
> 
> I also thought of something like this, which I don't know if it's a good idea or not:
> 
> ```
> /// Plugin 1:
> resources: {
>     selection_change_handlers: withSequence(READ, () => { /* reads stuff */ }),
> }
> /// Plugin 2:
> resources: {
>     selection_change_handlers: withSequence(WRITE, () => { /* reads stuff */ }),
> }
> ```
> 
> Not sure I like it honestly, but it... exists.
> 
> The thing is, `selection_change_handlers` is probably not the only case where we might need to order function calls based on reads first, writes last, hence why I'm looking for a way to express this without duplicating the resources names for each case.
> 
> @FrancoisGe Any good idea since you've looked at similar performance issues before?

To be honest, I haven't thought about it since our last discussion :/. I know that normalizers have a similar issue. There's probably a whole bunch of handlers that have this kind of problem. It would be great to find a pattern that would optimize these cases. I agree that it'll be hard to make them use it, but it still gives us a way to fix performance issues when we see them.

deso-odoo: See [this](https://www.odoo.com/mail/message/981604523) as well.

dmo-odoo: I suggest we do this `withSequence` with `READ`/`WRITE` as a first approach. It has the advantage of being explicit, reusable, and uses existing mechanisms, which means it's even valid for stable (as opposed to renaming resources for example).

You can do it for this case here first, and then once we agree on the end result then you can tackle the other task as well in the same fashion.

If we ever change our minds and want to go a different route later on, the time spent here will not be lost because it would have been spent to label the problematic handlers (the WRITE ones), which is something that would need to be done no matter what solution is chosen anyway.

deso-odoo: > and then once we agree on the end result then you can tackle the other task as well in the same fashion.

This approach is good for this task but can't work for [other task](https://www.odoo.com/odoo/project/1695/tasks/5245157) because there the WRITE handler itself is checking for `isBlock` as well as updating DOM. Handlers still remains costly for such cases:
```js
somehandler(node) {
    if (isBlock(node)) {
        node.classlist.add('my-class');
    }
}
```

dmo-odoo: Not saying we should do it, but in theory your example could easily be changed to:
```js
class MyPlugin {
    resources = {
        "normalize_handlers": [
            withSequence(READ, this.checkNode.bind(this)),
            withSequence(WRITE, this.updateNode.bind(this)),
        ]
    }
    checkNode(node) {
        if (isBlock(node)) { 
            this.nodesToNormalize.push(node)};
        }
    }
    updateNode(node) {
        if (this.nodesToNormalize.has(node)) {
            node.classlist.add('my-class');
            this.nodesToNormalize.delete(node);
        }
    }
}
```
I guess in practice the handler is more complex?

addons/html_editor/static/src/utils/resource.js
```diff
@@ -1,4 +1,6 @@
 export const resourceSequenceSymbol = Symbol("resourceSequence");
+export const READ = 0;
+export const WRITE = 100;
```
deso-odoo: @dmo-odoo PR ready for review. Though I have not adapted all selectionchange_handlers to use WRITE currently because, I'd prefer just to have READ and no WRITE, because what we actually need here is to ensure that READ handlers run before WRITE, and if READ runs first other handlers naturally runs after so we don't need WRITE. What do you think?

dmo-odoo: Ok but then people have no chance to understand why we need this READ/WRITE at all if we don't use `WRITE`, so you'll need to write a comment on top of the `const` definitions.

deso-odoo: Added the comment!

PR #244480, opened 2026-01-19
[IMP] web: worksheetpocalypse

damrOdoo:
This commit purpose is to get rid of the dependence of the worksheet template to the studio module. The goal is to let free user using the field service application with the worksheet feature without forcing them to pay extra for the studio module. 

From a technical point of view, the logic of the previous worksheet is completely replaced with the properties field. In order to be able to edit the properties definition of a worksheet template without being force to have a child model, a new widget 'properties_definition' which extends the 'properties_field' widget was added to the base web module.

task - 5167053

### DIFF
--- a/addons/web/static/src/views/fields/properties/properties_definition_field.js
+++ b/addons/web/static/src/views/fields/properties/properties_definition_field.js
@@ -1,0 +1,122 @@
+import { registry } from "@web/core/registry";
+import { exprToBoolean } from "@web/core/utils/strings";
+import { _t } from "@web/core/l10n/translation";
+import { PropertiesField } from "./properties_field";
+
+export class PropertiesDefinitionField extends PropertiesField {
+    static template = "web.PropertiesDefinitionField";
+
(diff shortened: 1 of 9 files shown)


addons/web/static/src/views/fields/properties/properties_definition_field.js
```diff
@@ -0,0 +1,114 @@
+import { registry } from "@web/core/registry";
+import { exprToBoolean } from "@web/core/utils/strings";
+import { _t } from "@web/core/l10n/translation";
+import { PropertiesField } from "./properties_field";
+
+export class PropertiesDefinitionField extends PropertiesField {
+
+    static template = "web.PropertiesDefinitionField";
```
xavierbol:
```suggestion
    static template = "web.PropertiesDefinitionField";
```
nitpicking

xavierbol (commented):
Thanks for your work, I left one small comment 🙂

PR #99849, opened 2022-09-08
[FIX] bus: fix interface error on server stop

tsm-odoo:
When stopping the server, all cursors are closed. The issue is that
the ImDispatch thread is kept alive until the main thread exits.

This can lead to errors during server stop: the ImDispatch thread
could try to poll an already closed connection thus raising a
`psycopg2.InterfaceError` exception.

This commit solves the issue by hiding the interface error when
it occurs during server stop.

Traceback :

```bash
Traceback (most recent call last):
  File "/data/build/odoo/addons/bus/models/bus.py", line 155, in loop
    conn.poll()
psycopg2.InterfaceError: connection already closed

During handling of the above exception, another exception occurred:

2022-09-08 14:51:00,385 271171 ERROR ? odoo.addons.bus.models.bus: Bus.loop error, sleep and retry 
Traceback (most recent call last):
  File "/home/odoo/Desktop/odoo/odoo/addons/bus/models/bus.py", line 198, in run
    self.loop()
  File "/home/odoo/Desktop/odoo/odoo/addons/bus/models/bus.py", line 184, in loop
    event.set()
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 162, in __exit__
    self.close()
  File "<decorator-gen-4>", line 2, in close
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 90, in check
    return f(self, *args, **kwargs)
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 379, in close
    return self._close(False)
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 415, in _close
    self.__pool.give_back(self._cnx, keep_in_pool=keep_in_pool)
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 592, in _locked
    return fun(self, *args, **kwargs)
  File "/home/odoo/Desktop/odoo/odoo/odoo/sql_db.py", line 683, in give_back
    raise PoolError('This connection does not belong to the pool')
psycopg2.pool.PoolError: This connection does not belong to the pool

```

PR #244481, opened 2026-01-19
[FIX] bus: fix pool error on server stop

tsm-odoo:
When the server stops, pool errors can occur. It's due to the server stop function calling `sql_db.py@close_all()`. If this occurs before the dispatcher loop wakes up, the `__exit__` function of the cursor context manager will raise an error, because the connection has already been popped from the pool.

As for interface errors in [1], pool errors should be ignored during server shutodwn.

runbot-229800

[1]: https://github.com/odoo/odoo/pull/99849

Current behavior before PR:

---

### DIFF
--- a/addons/bus/models/bus.py
+++ b/addons/bus/models/bus.py
@@ -10,6 +10,7 @@ import selectors
 import threading
 import time
 from psycopg2 import InterfaceError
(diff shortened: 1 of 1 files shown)


Julien00859 (approved):
@robodoo r+

PR #244482, opened 2026-01-19
[FIX] base: prevent error when unarchiving company

taqu-odoo:
**Steps to reproduce**
- Have at least two active companies
- Archive one company
- Unarchive the company
- Select the just unarchived company in the company selector
- `Access Error: Access to unauthorized or invalid companies.`

**Cause**
There was a timing issue between the cache invalidation and the call to super. After the cache invalidation, if `_get_company_ids` of `res.users` was called before the call to `super` in the `write`, it would not return the just unarchived company.

opw-5449925


kmagusiak (commented):
It looks like it is fixed in 19.0, can you confirm and update "up to" for mergebot?

kmagusiak (approved):

PR #244491, opened 2026-01-19
[FIX] web: Make Popover animation timing consistent in tests

BastienFafchamps:
Because the popover had his animation enabled in tests, it could in very rare occasion end too fast and call it's finished callback, triggering extra repositionning (and thus extra expect.steps).

---


addons/web/static/tests/core/popover/popover.test.js
```diff
@@ -375,3 +382,28 @@ test("popover fixed position", async () => {
 
     expect.verifySteps([]);
 });
+
+test("popover can animate", async () => {
+    patchWithCleanup(window.Element.prototype, {
+        animate() {
+            expect(this).toHaveClass("o_popover");
```
kebeclibre, line 389: I would avoid to do it that way, at least not by overriding every element's function.

I see that the animation transitions from opacity 0 to 1.
Maybe a waitUntil with the right condition would do the trick ?

PR #244493, opened 2026-01-19
[FIX] web: restrict group quick create to default groupby backport

Alpha9584:
Backport of commit https://github.com/odoo/odoo/commit/95f09fda03b7d2c74ea8cb33d2431d3a4cef95bf.

This commit updates the conditions for enabling group quick create in list and kanban views. In addition to existing checks, the view must now be grouped using its default groupby (if defined) for quick create to be available.

This ensures that group quick create is only shown in contexts where it is properly supported and avoids undesired behavior when grouping by arbitrary fields.

Fixes the RPC_ERROR when trying to add a salesperson while grouped by salesperson in CRM, which incorrectly passed default_type='opportunity' to res.users.name_create in the case of our issue.

Steps to reproduce:
1- Install the CRM module.
2- Go to CRM (kanban view) and groupby salesperson.
3- Add new salesperson from the view.

Ticket [link](https://www.odoo.com/odoo/project.task/5499818)
opw-5499818

---

### DIFF
--- a/addons/project/static/tests/legacy/views/project_task_kanban/project_task_kanban_view_tests.js
+++ b/addons/project/static/tests/legacy/views/project_task_kanban/project_task_kanban_view_tests.js
@@ -27,7 +27,7 @@ QUnit.module('Project', {
         ]);
         this.views = {
             "project.task,false,kanban":
-                `<kanban js_class="project_task_kanban">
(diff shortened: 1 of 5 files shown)


abz-odoo:
This is a big behavior change for 18.0, it will not be accepted.

PR #244494, opened 2026-01-19
[FIX] point_of_sale: fix customer display url

aksi-odoo:
Step to reproduce:
- start pos
- from top-right menu, click on display icon
- a dialog will appear, click on "this device" button

Observation:
- current session will be closed and open in new tab

Cause:
- Incorrect url formed for redirection which, as a fallback loads currrent pos
- base url is `undefined`, as we try to get it from `pos.session`, which now is attribute of `pos.config`.
https://github.com/odoo/odoo/blob/06ddce00115c906a4d8396387dd3332482145d7f/addons/point_of_sale/models/pos_config.py#L288

opw-5502812

---


alsh-odoo (approved):

loukwets (approved):
Nice, thanks for the test !

stva-odoo (approved):
robodoo r+

PR #244496, opened 2026-01-19
[IMP] web, survey: update certificate design

Brieuc-brd:
This commit redesigns the survey certificate templates.
Each new template has two versions: one colored and one black and white.
The colored variant uses the company’s primary color, which can be edited in the document layout settings.
It also introduces a new google font, Playball, available in common assets.

task-3775369

Co-authored-by: Dufays Pierre-Yves (PYDU) <pydu@odoo.com>

Requires:
- https://github.com/odoo/upgrade/pull/9359

---

| | Modern | Minimal | Classic 1 | Classic 2 |
|--------|--------|--------|--------|--------|
Colored | image |  image | image | image |
| Black | image | image | image | image |

---


addons/survey/models/survey_survey.py
```diff
@@ -130,13 +130,15 @@ def default_get(self, fields):
...
-        ('classic_gold', 'Classic Gold')],
-        string='Certification template', default='modern_purple')
+        ('modern_company', 'Modern'),
+        ('modern_black', 'Modern Black'),
+        ('minimal_company', 'Minimal'),
+        ('minimal_black', 'Minimal Black'),
+        ('classic-1_company', 'Classic #1'),
+        ('classic-1_black', 'Classic #1 Black'),
+        ('classic-2_company', 'Classic #2'),
+        ('classic-2_black', 'Classic #2 Black')],
```
thjo-odoo, line 140: Have you tried an upgrade from before your changes to after your changes ? 
I feel that changing values on a selection field might have undesired effects => Quid of already existing DBs?
If it only was the label part I think it wouldn't be an issue but here you changed both the key and the label.

thjo-odoo (changes requested):

addons/survey/models/survey_survey.py
Brieuc-brd: Indeed, the new designs don’t work in this case 🤔.

Since the current templates are becoming outdated, the goal of this PR is to refresh these designs and add the ability to use the company color, which wasn’t possible before.

What would you recommend to ease the replacement of the current templates without breaking anything?
Is there a way to keep the old designs for the existing DB while disabling them in the selection field?

cc . @Francois-frgi

thjo-odoo: I see 2 solutions for this one: either you reuse the old keys but change the label => in Databases there wouldn't be any need to upgrade the field. Or you keep it like you did there and you would need to write an upgrade script that would set the value of all the old records on a default value. Which I would let you decide in terms of design what fits best.

Brieuc-brd: Thanks for your time @thjo-odoo 🙂.
I don’t think the first solution is feasible, as the old key doesn’t correspond to the new design.
For example, using "classic_purple" for the "classic-1_black" template doesn’t make sense 😅 .

I think your second solution is the right approach. As a designer, I’ve never created an upgrade script before, so I'll need your or a dev support to handle that part 🙏 .

What do you think @Francois-frgi ?

thjo-odoo: @Brieuc-brd 

Hello, 
After discussing live with FRGI we decided to go for the upgrade path.
Here's the commit that I just did for that: https://github.com/odoo/upgrade/commit/e11b69910c081562013147678bb2cfcd333fc112

All you have to do is create your upgrade branch in odoo/upgrade with this commit and then create the PR.
You will just need to add the link to this PR and the task number. Then when creating the PR you will have a checklist to go through, quite easy as most of it is already done.
If you have any question or if you'd prefer I create the PR let me know.

Brieuc-brd: Thanks a lot @thjo-odoo ! 🙏  I've created the [Upgrade PR](https://github.com/odoo/upgrade/pull/9359) 👍 
I've also rebased this PR. Do I need to do anything else before you can proceed with the technical review ?

Thanks again !

thjo-odoo: To me it was my unique point of concern, for the design itself I trust as it's not my domain.
Maybe someone from @odoo/rd-sm-learn could have a look to be absolutely sure.

addons/survey/report/survey_templates.xml
```diff
@@ -2,69 +2,135 @@
...
-                             </p>
-                            <div class="oe_structure"/>
-                        </div>
-                        <div t-else="" class="certification-failed">
-                            <p>Certification Failed</p>
-                            <div class="oe_structure"/>
+             <div
+                id="o_survey_certification"
+                t-att-data-oe-model="user_input._name" t-att-data-oe-id="user_input.id"
+                t-attf-class="#{'article o_survey_certification position-relative o_layout_' + layout_template + ' o_color_' + layout_color + '_' + company_id } #{layout_template != 'minimal' and 'o_bg_color'}">
```
pyduf: Maybe you can simplify like this: 
```xml
t-attf-class="article o_survey_certification position-relative o_layout_{{layout_template}} o_color_{{layout_color}}_{{company_id}} #{layout_template != 'minimal' and 'o_bg_color'}"
```

pyduf (commented):
Hello @Brieuc-brd , 

Thanks for the explaination about wkhtmltopdf (it's hard to get a compatible version actually). The PR looks good to me. I have only added one tiny comment.

Also I wonder if we cannot extend an existing test to check that all the version contains essential information. The idea would be to add at the end of this test: def test_flow_certification(self):

```python
        # Check the rendering of all layout in html
        for layout_value, _ in self.env['survey.survey']._fields['certification_report_layout'].selection:
            with self.subTest(layout_value=layout_value):
                user_inputs.survey_id.certification_report_layout = layout_value
                layout, color = layout_value.split('_')
                html = str(self.env['ir.actions.report'].with_user(user_new_company).with_company(
                    new_company)._render_qweb_html('survey.certification_report_view', user_inputs.ids)[0])
                self.assertIn(f'o_layout_{layout}', html)
                self.assertIn(f'o_color_{color}_{self.survey_user.company_id.id}', html)
                self.assertIn('Eglantine Employee', html)
                self.assertIn(self.survey_user.company_id.name, html)
                self.assertIn(str(user_inputs.id), html)
                self.assertIn('02/15/2020', html)
                self.assertNotIn('Certification failed', html)
```

And to get the correct date, decorate the test by:
```python
# ...
@freeze_time("2020-02-15 18:00")
def test_flow_certification(self):
```
The idea is just to render in html all the different template and checking there are essential no missing part (name of the user, ...). What do you think ?

Also, can you add the upgrade PR on the task ?

Finally is there a reason to add the font in web rather than survey (like before) ? Maybe you can add a comment in the commit for that ?

pyduf (commented):
Hello @Brieuc-brd , 

Thanks for the followup. It looks good to me and I think you should take all the credits as you have done all the work!

The upgrade works well but I wonder if it is intended that all the values for the certification_report_layout are mapped to modern_company. 
Should we not map old classic template to the new classic one ? (maybe it is a detail). Sorry, didn’t see it before.

Also, a small details about the import that should be ordered alphabetically:
```python
from unittest.mock import patch
```
->
```python
from unittest.mock import patch
```

pyduf (approved):
Hello @Brieuc-brd,

Thanks for the followup, it looks good to me.

@awa-odoo, could you either r+ or delegate it to me when you have time ? Thanks
Note that the runbot is red because of the added font with a size greater than 100K: web/static/fonts/google/Playball/Playball-Regular.tt 186K

addons/web/static/fonts/google/Playball/Playball-Regular.ttf
d-fence, line 1: Please add the [license file](https://github.com/googlefonts/play-ball/blob/master/OFL.txt) alongside with the font.

d-fence (changes requested):

addons/web/static/fonts/google/Playball/Playball-Regular.ttf
Brieuc-brd: @d-fence 👋 
If I'm not mistaken, it's already there : https://github.com/odoo-dev/odoo/tree/57fcad86a0a788ae2a3ff2f05d044b4b332d615a/addons/web/static/fonts/google/Playball
or did I missed something ? 🤔

Brieuc-brd: @d-fence 👋  small reminder :)

d-fence (approved):
Sorry, I filtered out the txt files and din't see the license.
So it's all good from my point of view.

pyduf:
@robodoo rebase-merge

PR #244500, opened 2026-01-19
[FIX] website_slides: prevent merging contacts enrolled in common courses

mobar-odoo:
Expected Behaviour:
Contacts enrolled in common courses should not be merged and the merge
should fail.

Steps to reproduce:
1- Go to one of the courses
2- Add two attendees to the course
3- Go to Contacts App
4- Select the two attendees you added to the course
5- Try merging the two contacts

Actual Behaviour before the Fix:
Contacts enrolled in common courses are getting merged and the common
courses are kept in the destination contact.

Behaviour with the Fix:
Contacts enrolled in common courses are blocked from being merged and an
error message is shown to the user saying that the reason the merge is
blocked is a duplicate course.

opw-5417223

### DIFF
--- a/addons/website_slides/models/base_partner_merge.py
+++ b/addons/website_slides/models/base_partner_merge.py
@@ -1,0 +1,25 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = 'base.partner.merge.automatic.wizard'
+
+    @api.model
+    def _update_foreign_keys(self, src_partners, dst_partner):
+        # prevent merging partners enrolled in common courses
+        courses_set = set(dst_partner.sudo().slide_channel_ids.ids)
+        duplicate_courses_names_set = set()
+        for partner in src_partners:
+            for slide_channel in partner.sudo().slide_channel_ids:
+                if slide_channel.id in courses_set:
+                    duplicate_courses_names_set.add(slide_channel.name)
+                else:
+                    courses_set.add(slide_channel.id)
+        if duplicate_courses_names_set:
+            raise UserError(_(
+                "You cannot merge these contacts because multiple contacts are enrolled in the same courses: %s",
+                ', '.join(duplicate_courses_names_set)))
+        super()._update_foreign_keys(src_partners, dst_partner)
--- a/addons/website_slides/tests/test_slide_channel.py
+++ b/addons/website_slides/tests/test_slide_channel.py
@@ -156,6 +156,33 @@ class TestSlidesManagement(slides_common.SlidesCase):
                 for mail in created_mails)
         )
 
+    def test_merging_partners_with_course_memberships(self):
+        """ Test merging partners with course memberships """
+        course_1, course_2 = self.env['slide.channel'].create([{'name': 'Course 1'}, {'name': 'Course 2'}])
+        partner_1, partner_2, partner_3 = partners = self.env['res.partner'].create([
+            {'name': 'Partner 1', 'email': 'partner1@example.com'},
+            {'name': 'Partner 2', 'email': 'partner2@example.com'},
+            {'name': 'Partner 3', 'email': 'partner3@example.com'}])
+        partners.invalidate_recordset(fnames=['slide_channel_ids'])
+        course_1.sudo()._action_add_members(partner_1 | partner_2)
+        course_2.sudo()._action_add_members(partner_1 | partner_3)
+        wizard = self.env['base.partner.merge.automatic.wizard'].create({})
+
+        with self.assertRaises(UserError) as user_error:
+            wizard._merge([partner_1.id, partner_2.id], partner_1)
+
+        self.assertEqual(
+            user_error.exception.args[0],
+            "You cannot merge these contacts because multiple contacts are enrolled in the same courses: Course 1",
+            "Error message should appear that mentions the common courses that prevent the merge"
+        )
+
+        wizard._merge([partner_2.id, partner_3.id], partner_2)
+        self.assertFalse(partner_3.exists(), "Source partner should be deleted after merge")
+        self.assertTrue(partner_2.exists(), "Destination partner should exist after merge")
+        self.assertIn(course_1, partner_2.slide_channel_ids, "Course 1 should belong to destination partner")
+        self.assertIn(course_2, partner_2.slide_channel_ids, "Course 2 should belong to destination partner")
+
     def test_mail_completed_with_different_templates(self):
         """ When the completion email is generated, it must take into account different templates. """
 
--- a/addons/website_slides/models/__init__.py
+++ b/addons/website_slides/models/__init__.py
@@ -1,5 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+from . import base_partner_merge
 from . import gamification_challenge
 from . import gamification_karma_tracking
 from . import ir_binary


addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,19 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+from odoo import api, models
+from odoo.exceptions import UserError
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = 'base.partner.merge.automatic.wizard'
+
+    @api.model
+    def _update_foreign_keys(self, src_partners, dst_partner):
```
pyduf, line 11: I was wondering why not overriding _merge instead but I think this is because basic checks are already done in that method and we want to do this check afterwards. So I think you did the right choice (as there are unfortunately no method to do addition check of the parameters called). What do you think ?

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,19 @@
...
+        # check if the list of partners to merge have duplicate course
```
pyduf: Maybe " # prevent merging partners enrolled in common courses" ?

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,19 @@
...
+        courses_set = set(dst_partner.sudo().slide_channel_ids.ids)
+        for partner in src_partners:
+            for slide_channel in partner.sudo().slide_channel_ids:
+                if slide_channel.id in courses_set:
+                    raise UserError(self.env._("You cannot merge contacts that are enrolled in the same course."))
```
pyduf: Maybe we could list the common course to help solving the issue ?

addons/website_slides/tests/test_slide_channel.py
```diff
@@ -162,6 +162,38 @@ def test_mail_completed(self):
                 for mail in created_mails)
         )
 
+    def test_merge_partners_with_unique_courses(self):
+        """ Test merging partners with unique courses """
+        course1 = self.env['slide.channel'].create({'name': 'Course 1'})
+        course2 = self.env['slide.channel'].create({'name': 'Course 2'})
+        partner1 = self.env['res.partner'].create({'name': 'Partner 1', 'email': 'partner1@example.com'})
+        partner2 = self.env['res.partner'].create({'name': 'Partner 2', 'email': 'partner2@example.com'})
```
pyduf: (details) You can create them in batches: 
```python
course1, course2 = self.env['slide.channel'].create([{'name': 'Course 1'}, {'name': 'Course 2'}])
partner1, partner2 = self.env['res.partner'].create([{'name': 'Partner 1', 'email': 'partner1@example.com'},
                                                     {'name': 'Partner 2', 'email': 'partner2@example.com'}])
```

addons/website_slides/tests/test_slide_channel.py
```diff
@@ -162,6 +162,38 @@ def test_mail_completed(self):
...
+        self.assertFalse(partner2.exists(), "Source partner should be deleted after merge")
+        self.assertTrue(partner1.exists(), "Destination partner should exist after merge")
+        self.assertIn(course1, partner1.slide_channel_ids, "Course 1 should belong to destination partner")
+        self.assertIn(course2, partner1.slide_channel_ids, "Course 2 should belong to destination partner")
+
+    def test_merge_partners_with_duplicate_course(self):
+        """ Test merging partners with duplicate courses. Merging should fail in that case. """
+        course1 = self.env['slide.channel'].create({'name': 'Course 1'})
+        partner1 = self.env['res.partner'].create({'name': 'Partner 1', 'email': 'partner1@example.com'})
+        partner2 = self.env['res.partner'].create({'name': 'Partner 2', 'email': 'partner2@example.com'})
```
pyduf: Idem here (you can create them in batch)

addons/website_slides/tests/test_slide_channel.py
```diff
@@ -162,6 +162,38 @@ def test_mail_completed(self):
...
+        wizard = self.env['base.partner.merge.automatic.wizard'].create({})
+        wizard._merge([partner1.id, partner2.id], partner1)
+
+        self.assertFalse(partner2.exists(), "Source partner should be deleted after merge")
+        self.assertTrue(partner1.exists(), "Destination partner should exist after merge")
+        self.assertIn(course1, partner1.slide_channel_ids, "Course 1 should belong to destination partner")
+        self.assertIn(course2, partner1.slide_channel_ids, "Course 2 should belong to destination partner")
+
+    def test_merge_partners_with_duplicate_course(self):
+        """ Test merging partners with duplicate courses. Merging should fail in that case. """
```
pyduf: Maybe "Test that merging partners with common courses fails. " ?

pyduf (commented):
Hello @mobar-odoo , 

Thanks for the fix. It works fine. I have added some minor comments. 

I wonder mainly if we should not list the common courses in the error to help to solve the issue. What do you think ?

(detail) Also, usually we put the task reference at the end of the commit not at the start.

addons/website_slides/models/base_partner_merge.py
mobar-odoo: Yes, I thought of overriding this method because it is the method that updates the foreign keys. So, for me, it was the right method since we're checking the foreign keys and will update them if the check passes.

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = 'base.partner.merge.automatic.wizard'
+
+    @api.model
+    def _update_foreign_keys(self, src_partners, dst_partner):
+        # prevent merging partners enrolled in common courses
+        courses_set = set(dst_partner.sudo().slide_channel_ids.ids)
+        for partner in src_partners:
+            for slide_channel in partner.sudo().slide_channel_ids:
+                if slide_channel.id in courses_set:
+                    raise UserError(self.env._("You cannot merge contacts that are enrolled in the same course (%(course_name)s).")) % {'course_name': slide_channel.name}
```
pyduf: There is an error on the runbot about this line (unsupported operand type(s) for %: 'UserError' and 'dict'). Also would it be easy possible to accumulate the common course and list all the common course in the error ?

pyduf (commented):
Hello @mobar-odoo, 

Thanks for the followup. Actually there an error on the runbot (easy to fix). I also wonder if it would be easy to display all the common courses in case of error (not only one) ?

addons/website_slides/models/base_partner_merge.py
mobar-odoo: Hello @pyduf ,

I fixed the error and changed it to display all the common courses.

pyduf (commented):
Hello @mobar-odoo ,

Actually, the user error message should be translated using the underscore method and as we are in stable we need also to update the pot file with that added translation (You can generate the new translation file in odoo "Settings / Translations / Export", then you should add only the new translation using a comparing tool). Sorry not seeing it earlier.

pyduf (approved):
Hello @mobar-odoo , 

Thanks, it seems you're right about not having to update the pot file and it works well so it looks good to me.

Hello @flch-odoo , when you have sometime and if it looks also good to you, could you r+ it ? Thanks.

pyduf (commented):
Hello @mobar-odoo ,

Sorry, I've forgot to ask (thanks Florian for the advice) why we're doing this fix in 19 and not in an earlier version.
Is the bug only present in 19.0+ versions ? If so maybe you could add a small comment in the commit message (what introduce the problem).
If not, should we not fix it in earlier supported version as well ?

Thanks.

pyduf:
Hello @mobar-odoo , 

Thanks for having backported to version 17 and sorry for the delay. It works well.

The only question I have is why we need to invalidate cache in _update_foreign_keys. Actually when trying it on the interface it's not necessary because adding the attendees and merging the contact will always happens in 2 differents call. So I wonder if we could not move the invalidation of the cache in the test (as setting the course attendee is kind of a setup of the test). What do you think ?

mobar-odoo:
> Hello @mobar-odoo ,
> 
> Thanks for having backported to version 17 and sorry for the delay. It works well.
> 
> The only question I have is why we need to invalidate cache in _update_foreign_keys. Actually when trying it on the interface it's not necessary because adding the attendees and merging the contact will always happens in 2 differents call. So I wonder if we could not move the invalidation of the cache in the test (as setting the course attendee is kind of a setup of the test). What do you think ?

Hello @pyduf ,

You're correct. I thought cache needed to be invalidated in the method not just in the test. I've updated it accordingly.

addons/website_slides/tests/test_slide_channel.py
```diff
@@ -156,6 +156,42 @@ def test_mail_completed(self):
...
+
+        self.assertFalse(partner2.exists(), "Source partner should be deleted after merge")
+        self.assertTrue(partner1.exists(), "Destination partner should exist after merge")
+        self.assertIn(course1, partner1.slide_channel_ids, "Course 1 should belong to destination partner")
+        self.assertIn(course2, partner1.slide_channel_ids, "Course 2 should belong to destination partner")
+
+    def test_merge_partners_with_duplicate_courses(self):
+        """ Test that merging partners with common courses fails """
+        course1, course2 = self.env['slide.channel'].create([{'name': 'Course 1'}, {'name': 'Course 2'}])
+        partner1, partner2 = self.env['res.partner'].create([
```
flch-odoo: this way you can use `partners`, and `partners.ids` below without re-creating a recordset each time.
Same for test above.
For consistency, please also note the python "way" of naming variables (with `_` as separator), as in the rest of the file too.

```suggestion
        partner_1, _partner_2 = partners = self.env['res.partner'].create([
```

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = 'base.partner.merge.automatic.wizard'
+
+    @api.model
+    def _update_foreign_keys(self, src_partners, dst_partner):
+        # prevent merging partners enrolled in common courses
+        courses_set = set(dst_partner.sudo().slide_channel_ids.ids)
+        duplicate_courses = self.env['slide.channel']
```
flch-odoo: Perhaps a `duplicate_courses_names_set = set()` can store the `name`s directly?

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,24 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
flch-odoo, line 1: We usually leave a blank line after the license info.

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo import api, models, _
```
flch-odoo, line 3: There is also a couple things we can do on the commit message to make it clearer imho:
* switch to lowercase for `prevent` on the title, aligning with guidelines
* Isn't it more specifically an issue of "multiple memberships" than "duplicate courses"?
* No need for a hyperlink to the task on the commit message, the ticket id is already present.
* If you start with the steps to reproduce, you'll have your expected/actual/after fix paragraphs and won't need to repeat yourself with the extra 
> Before the fix, the merging will happen.
After the fix, the merging will stop and an error message will appear.

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo.exceptions import UserError
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = 'base.partner.merge.automatic.wizard'
+
+    @api.model
+    def _update_foreign_keys(self, src_partners, dst_partner):
+        # prevent merging partners enrolled in common courses
```
flch-odoo, line 12: Isn't this docstring material?

addons/website_slides/models/base_partner_merge.py
```diff
@@ -0,0 +1,24 @@
...
+        duplicate_courses = self.env['slide.channel']
+        for partner in src_partners:
+            for slide_channel in partner.sudo().slide_channel_ids:
+                if slide_channel.id in courses_set:
+                    duplicate_courses |= slide_channel
+                else:
+                    courses_set.add(slide_channel.id)
+        if duplicate_courses:
+            raise UserError(_(
+                "You cannot merge these contacts because they are enrolled in the following common courses: %s",
```
flch-odoo:
```suggestion
                "You cannot merge these contacts because multiple contacts are enrolled in the same courses: %s",

**Comment by flch-odoo** [comment]
Hello @mobar-odoo , @pyduf . Thanks for the fix and validation respectively. Here are some polish-addressing comments before merging "for ever" :)

**Comment by mobar-odoo** [comment]
> Hello @mobar-odoo , @pyduf . Thanks for the fix and validation respectively. Here are some polish-addressing comments before merging "for ever" :)

Hello @flch-odoo , you mentioned some comments but I don't see any :) Perhaps you forgot to post them

File: addons/website_slides/models/base_partner_merge.py (hunk 12 above)
Comment (mobar-odoo) [reply addons/website_slides/models/base_partner_merge.py:12]: No, that's just a small comment to explain what this method overload is doing

File: addons/website_slides/tests/test_slide_channel.py [hunk 14]
```diff
@@ -156,6 +156,46 @@ def test_mail_completed(self):
...
+        wizard = self.env['base.partner.merge.automatic.wizard'].create({})
+        wizard._merge(partners.ids, partner_1)
+
+        self.assertFalse(partner_2.exists(), "Source partner should be deleted after merge")
+        self.assertTrue(partner_1.exists(), "Destination partner should exist after merge")
+        self.assertIn(course1, partner_1.slide_channel_ids, "Course 1 should belong to destination partner")
+        self.assertIn(course2, partner_1.slide_channel_ids, "Course 2 should belong to destination partner")
+
+    def test_merge_partners_with_duplicate_courses(self):
+        """ Test that merging partners with common courses fails """
```
flch-odoo: We don't need two courses here, just one is enough and makes the test shorter (faster).

That said, it would be even faster if we had three partners in the test above, 
* `course_1`, enrolled `partner_1` and `partner_2`
* `course_2`: enrolled `partner_1` and `partner_3`,
and we'd test that we can't merge partners 1 and 2, but we can 2 and 3.

addons/website_slides/tests/test_slide_channel.py
```diff
@@ -156,6 +156,46 @@ def test_mail_completed(self):
                 for mail in created_mails)
         )
 
+    def test_merge_partners_with_unique_courses(self):
+        """ Test merging partners with unique courses """
```
flch-odoo:
```suggestion
        """ Test merging partners with memberships in distinct courses """
```

or, if we go with the other suggestion:

```suggestion
        """ Test merging partners with course memberships """
```

flch-odoo:
@robodoo r+

mobar-odoo:
@flch-odoo thanks for the review :) I see you already r+ -ed

PR #244507, opened 2026-01-19
[IMP] l10n_sa_edi: zatca onboarding bugprovement

khom-odoo:
Previously, it was assumed that if no 'error' or 'errors' key was present that means we've received a valid response for obtaining CCSID or PCSID.

But sometimes the error is not sent with those keys, and the binarySecurityToken is missing, therefore a traceback is shown to the user because the invalid requests passes through the validation unnoticed. This kind of response is the result of a new change introduced by zatca requiring csr fields to be at most 64 characters long.

This commit improves the error handling mechanism of CSID responses, to show the user an informative message, and handles the length check for csr fields on the client side before sending to zatca.

task-5347269
---

### DIFF
--- a/addons/l10n_sa_edi/models/account_journal.py
+++ b/addons/l10n_sa_edi/models/account_journal.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import requests
 from markupsafe import Markup
 from lxml import etree
@@ -41,6 +42,9 @@ SANDBOX_AUTH = {
     'binarySecurityToken': "TUlJRDFEQ0NBM21nQXdJQkFnSVRid0FBZTNVQVlWVTM0SS8rNVFBQkFBQjdkVEFLQmdncWhrak9QUVFEQWpCak1SVXdFd1lLQ1pJbWlaUHlMR1FCR1JZRmJHOWpZV3d4RXpBUkJnb0praWFKay9Jc1pBRVpGZ05uYjNZeEZ6QVZCZ29Ka2lhSmsvSXNaQUVaRmdkbGVIUm5ZWHAwTVJ3d0dnWURWUVFERXhOVVUxcEZTVTVXVDBsRFJTMVRkV0pEUVMweE1CNFhEVEl5TURZeE1qRTNOREExTWxvWERUSTBNRFl4TVRFM05EQTFNbG93U1RFTE1Ba0dBMVVFQmhNQ1UwRXhEakFNQmdOVkJBb1RCV0ZuYVd4bE1SWXdGQVlEVlFRTEV3MW9ZWGxoSUhsaFoyaHRiM1Z5TVJJd0VBWURWUVFERXdreE1qY3VNQzR3TGpFd1ZqQVFCZ2NxaGtqT1BRSUJCZ1VyZ1FRQUNnTkNBQVRUQUs5bHJUVmtvOXJrcTZaWWNjOUhEUlpQNGI5UzR6QTRLbTdZWEorc25UVmhMa3pVMEhzbVNYOVVuOGpEaFJUT0hES2FmdDhDL3V1VVk5MzR2dU1ObzRJQ0p6Q0NBaU13Z1lnR0ExVWRFUVNCZ0RCK3BId3dlakViTUJrR0ExVUVCQXdTTVMxb1lYbGhmREl0TWpNMGZETXRNVEV5TVI4d0hRWUtDWkltaVpQeUxHUUJBUXdQTXpBd01EYzFOVGc0TnpBd01EQXpNUTB3Q3dZRFZRUU1EQVF4TVRBd01SRXdEd1lEVlFRYURBaGFZWFJqWVNBeE1qRVlNQllHQTFVRUR3d1BSbTl2WkNCQ2RYTnphVzVsYzNNek1CMEdBMVVkRGdRV0JCU2dtSVdENmJQZmJiS2ttVHdPSlJYdkliSDlIakFmQmdOVkhTTUVHREFXZ0JSMllJejdCcUNzWjFjMW5jK2FyS2NybVRXMUx6Qk9CZ05WSFI4RVJ6QkZNRU9nUWFBL2hqMW9kSFJ3T2k4dmRITjBZM0pzTG5waGRHTmhMbWR2ZGk1ellTOURaWEowUlc1eWIyeHNMMVJUV2tWSlRsWlBTVU5GTFZOMVlrTkJMVEV1WTNKc01JR3RCZ2dyQmdFRkJRY0JBUVNCb0RDQm5UQnVCZ2dyQmdFRkJRY3dBWVppYUhSMGNEb3ZMM1J6ZEdOeWJDNTZZWFJqWVM1bmIzWXVjMkV2UTJWeWRFVnVjbTlzYkM5VVUxcEZhVzUyYjJsalpWTkRRVEV1WlhoMFoyRjZkQzVuYjNZdWJHOWpZV3hmVkZOYVJVbE9WazlKUTBVdFUzVmlRMEV0TVNneEtTNWpjblF3S3dZSUt3WUJCUVVITUFHR0gyaDBkSEE2THk5MGMzUmpjbXd1ZW1GMFkyRXVaMjkyTG5OaEwyOWpjM0F3RGdZRFZSMFBBUUgvQkFRREFnZUFNQjBHQTFVZEpRUVdNQlFHQ0NzR0FRVUZCd01DQmdnckJnRUZCUWNEQXpBbkJna3JCZ0VFQVlJM0ZRb0VHakFZTUFvR0NDc0dBUVVGQndNQ01Bb0dDQ3NHQVFVRkJ3TURNQW9HQ0NxR1NNNDlCQU1DQTBrQU1FWUNJUUNWd0RNY3E2UE8rTWNtc0JYVXovdjFHZGhHcDdycVNhMkF4VEtTdjgzOElBSWhBT0JOREJ0OSszRFNsaWpvVmZ4enJkRGg1MjhXQzM3c21FZG9HV1ZyU3BHMQ==",
     'secret': "Xlj15LyMCgSC66ObnEO/qVPfhSbs3kDTjWnGheYhfSs="
 }
+MAX_ALLOWED_CSR_VALUE_LENGTH = 64
+
+_logger = logging.getLogger(__name__)
 
 
 class AccountJournal(models.Model):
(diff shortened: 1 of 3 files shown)


addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -216,6 +249,27 @@ def _l10n_sa_generate_csr(self):
 
     # ====== Certificate Methods =======
 
+    def _l10n_sa_get_csid_error(self, CSID):
+        """
+            Return a formatted error string if the CSID response has an 'error' or 'errors'
+            key or doesn't have a 'binarySecurityToken'
+        """
+        errorMsg = ""
+        unknownErrorMsg = "Unknown response returned from ZATCA. Please check the logs."
```
FlorianGilbert: We are using snake_case and not camelCase according to the [guidelines](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html#symbols-and-conventions:~:text=use%20underscore%20lowercase%20notation%20for%20common%20variable.).

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -216,6 +249,27 @@ def _l10n_sa_generate_csr(self):
...
+            errorMsg = error
+        elif errors := CSID.get('errors'):
+            errorMsg = " <br/>" + " <br/>- ".join([err['message'] if isinstance(err, dict) else err for err in errors])
+        elif 'error' in [CSID.get('type', "").lower(), CSID.get('status', "").lower()]:
+            errorMsg = CSID.get('message', unknownErrorMsg)
+        elif not CSID.get('binarySecurityToken'):
+            errorMsg = unknownErrorMsg
+
+        if errorMsg:
+            _logger.warning(_("Failed to obtain CCSID: %s", CSID))
```
FlorianGilbert: No need to translate technical log.

addons/l10n_sa_edi/models/account_journal.py
FlorianGilbert: This message should be translated.

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -216,6 +249,27 @@ def _l10n_sa_generate_csr(self):
...
+            key or doesn't have a 'binarySecurityToken'
+        """
+        errorMsg = ""
+        unknownErrorMsg = "Unknown response returned from ZATCA. Please check the logs."
+        if error := CSID.get('error'):
+            errorMsg = error
+        elif errors := CSID.get('errors'):
+            errorMsg = " <br/>" + " <br/>- ".join([err['message'] if isinstance(err, dict) else err for err in errors])
+        elif 'error' in [CSID.get('type', "").lower(), CSID.get('status', "").lower()]:
+            errorMsg = CSID.get('message', unknownErrorMsg)
```
FlorianGilbert:
```suggestion
            errorMsg = CSID.get('message') or unknownErrorMsg
```
Better in case we got a key with a null value.

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -216,6 +249,27 @@ def _l10n_sa_generate_csr(self):
 
     # ====== Certificate Methods =======
 
+    def _l10n_sa_get_csid_error(self, CSID):
```
FlorianGilbert: is it necessary to have a full caps parameter name?

Without context CSID doesn't mean nothing 😅

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +138,63 @@ def _l10n_sa_csr_required_fields(self):
         """ Return the list of fields required to generate a valid CSR as per ZATCA requirements """
         return ['l10n_sa_private_key', 'vat', 'name', 'city', 'country_id', 'state_id']
 
+    def _l10n_sa_get_csr_raw_vals(self):
```
FlorianGilbert: Having this method and the other one (`_l10n_sa_get_csr_vals`) is very confusing tbh 🤯

FlorianGilbert: `self.ensure_one() `?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +138,63 @@ def _l10n_sa_csr_required_fields(self):
...
+            ("org_id", parent_company_id.vat if parent_company_id else company_id.vat, _("%sCompany VAT", "Parent " if parent_company_id else "")),
+            ("state_name", company_id.state_id.name, _("State/Province Name")),
+            ("locality_name", company_id.city, _("Locality Name")),
+            ("egs_serial", '1-Odoo|2-%s|3-%s' % (service.common.exp_version()['server_version_info'][0], self.l10n_sa_serial_number), _("Journal Serial Number")),
+            ("org_uid", company_id.vat, _("Company VAT")),
+            ("invoice_type", company_id._l10n_sa_get_csr_invoice_type(), _("Invoice Type")),
+            ("location", company_id.street, _("Street")),
+            ("industry", company_id.partner_id.industry_id.name or 'Other', _("Partner Industry Name")),
+            ("cert_tmp", CERT_TEMPLATE_NAME[company_id.l10n_sa_api_mode], _("Certificate Template Name")),
+        ]
```
FlorianGilbert: To be frank, it is a bit messy. Don't we have another way of doing this?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +138,63 @@ def _l10n_sa_csr_required_fields(self):
...
+
+    def _l10n_sa_validate_csr_vals(self):
+        error_fields = set()
+        for key, val, fname in self._l10n_sa_get_csr_raw_vals():
+            if len(val) > 64:
```
FlorianGilbert: Look like a magic number, what about using a constant or something meaningful?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +138,63 @@ def _l10n_sa_csr_required_fields(self):
...
+            ("org_uid", company_id.vat, _("Company VAT")),
+            ("invoice_type", company_id._l10n_sa_get_csr_invoice_type(), _("Invoice Type")),
+            ("location", company_id.street, _("Street")),
+            ("industry", company_id.partner_id.industry_id.name or 'Other', _("Partner Industry Name")),
+            ("cert_tmp", CERT_TEMPLATE_NAME[company_id.l10n_sa_api_mode], _("Certificate Template Name")),
+        ]
+
+    def _l10n_sa_validate_csr_vals(self):
+        error_fields = set()
+        for key, val, fname in self._l10n_sa_get_csr_raw_vals():
```
FlorianGilbert: `fname`? 🤔 
field name?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +138,63 @@ def _l10n_sa_csr_required_fields(self):
...
+            if len(val) > 64:
+                error_fields.add(fname)
+        if error_fields:
+            company_msg = _("<br/><br/>Once the journal is onboarded, please update the company name to match the one listed on the VAT Registration Certificate.") if _("Company Name") in error_fields else ""
+            raise UserError(_("Please make sure the following fields are shorter than 64 characters: %s", " <br/>- " + " <br/>- ".join(error_fields) + company_msg))
```
FlorianGilbert: Why do we have twice `<br/>- `?

FlorianGilbert (changes requested):
Hm, no test, no merge? 😄

addons/l10n_sa_edi/models/account_journal.py
khom-odoo: The first one to break the list of fields to a new line. The second one is used as a separator between the fields.

FlorianGilbert: Ok but we have twice the dash. Ok for the second one, but I don't get the first one, isn't it an error?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+                "value": company_id.country_id.code,
+                "name": _("Country Name")
+            },
+            "org_unit_name": {
+                "value": company_id.name if parent_company_id else company_id.vat[:10],
+                "name": _("Company Name")
+            },
+            "org_name": {
+                "value": parent_company_id.name if parent_company_id else company_id.name,
+                "name": _("%sCompany Name", "Parent " if parent_company_id else "")
```
FlorianGilbert: Parent is not translatable. 

What about this:

```suggestion
                "name": _("Parent Company Name") if parent_company_id else _("Company Name")
```

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+            },
+            "common_name": {
+                "value": "%s-%s-%s" % (self.code, self.name, company_id.name),
```
FlorianGilbert: personal opinion but f-string is better imo

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
         return ['l10n_sa_private_key', 'vat', 'name', 'city', 'country_id', 'state_id']
 
+    def _l10n_sa_get_csr_vals(self):
+        self.ensure_one()
+        company_id = self.company_id
+        parent_company_id = self.company_id.parent_id
+        return {
+            "country_name": {
+                "value": company_id.country_id.code,
+                "name": _("Country Name")
```
FlorianGilbert: Working for all elements in the dict but don't forget the traling comma to avoid useless futur diff 😉

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+                "value": parent_company_id.name if parent_company_id else company_id.name,
+                "name": _("%sCompany Name", "Parent " if parent_company_id else "")
+            },
+            "common_name": {
+                "value": "%s-%s-%s" % (self.code, self.name, company_id.name),
+                "name": _("Common Name")
+            },
+            "org_id": {
+                "value": parent_company_id.vat if parent_company_id else company_id.vat,
+                "name": _("%sCompany VAT", "Parent " if parent_company_id else "")
```
FlorianGilbert: Same remark concerning the Parent not translatable

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+            "state_name": {
+                "value": company_id.state_id.name,
+                "name": _("State/Province Name")
+            },
+            "locality_name": {
+                "value": company_id.city,
+                "name": _("Locality Name")
+            },
+            "egs_serial": {
+                "value": '1-Odoo|2-%s|3-%s' % (service.common.exp_version()['server_version_info'][0], self.l10n_sa_serial_number),
```
FlorianGilbert: same remark about the f-string but feel free to ignore.

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+            "invoice_type": {
+                "value": company_id._l10n_sa_get_csr_invoice_type(),
+                "name": _("Invoice Type")
+            },
+            "location": {
+                "value": company_id.street,
+                "name": _("Street")
+            },
+            "industry": {
+                "value": company_id.partner_id.industry_id.name or 'Other',
```
FlorianGilbert: `Other` shouldn't be translated?

addons/l10n_sa_edi/models/account_journal.py
```diff
@@ -135,32 +139,98 @@ def _l10n_sa_csr_required_fields(self):
...
+        }
+
+    def _l10n_sa_validate_csr_vals(self):
+        error_fields = set()
+        for data in self._l10n_sa_get_csr_vals().values():
+            if len(data['value']) > MAX_ALLOWED_CSR_VALUE_LENGTH:
+                error_fields.add(data['name'])
+        if error_fields:
+            company_msg = _("<br/><br/>Once the journal is onboarded, please update the company name to match the one listed on the VAT Registration Certificate.") if _("Company Name") in error_fields else ""
+            raise UserError(_("Please make sure the following fields are shorter than %d characters: %s", MAX_ALLOWED_CSR_VALUE_LENGTH, " <br/>- " + " <br/>- ".join(error_fields) + company_msg))
```
FlorianGilbert: In translated sentence, it's the usage to name parameter when you have more than one to help translator for rtl language.

addons/l10n_sa_edi/tests/test_edi_zatca.py
```diff
@@ -492,3 +492,90 @@ def test_company_missing_country_on_standard_invoice(self):
...
+            'type': 'sale',
+            'company_id': compliant_company.id,
+        })
+
+        compliant_journal._l10n_sa_edi_set_csr_fields()
+        try:
+            csr_string = compliant_journal._l10n_sa_get_csr_str()
+            self.assertTrue(csr_string, "a Valid CSR should not be empty")
+        except UserError as e:
+            self.fail(f"Compliant company should not raise error: {e}")
```
FlorianGilbert, line 525: Not sure this is useful but why not 😄

addons/l10n_sa_edi/tests/test_edi_zatca.py
```diff
@@ -492,3 +492,90 @@ def test_company_missing_country_on_standard_invoice(self):
...
+            "Company Name",
+            "Common Name",
+            "Street",
+            "Locality Name",
+            "State/Province Name",
+            "Partner Industry Name",
+        ]
+
+        for field_name in expected_error_fields:
+            self.assertIn(field_name, error_message,
```
FlorianGilbert: Could be onelined

FlorianGilbert (changes requested):
Even if the original ticket is in the task, may I ask you to add it in the commit message as well? Just after the task id?
thank you.

addons/l10n_sa_edi/models/account_journal.py
khom-odoo: br_demo
because `"<sep>".join()` will use `sep` in between the elements, so the first one doesn't get any

addons/l10n_sa_edi/tests/test_edi_zatca.py
```diff
@@ -492,3 +492,89 @@ def test_company_missing_country_on_standard_invoice(self):
...
+            'street': 'Short Street Name',
+            'city': 'Riyadh',
+            'zip': '12345',
+            'country_id': self.saudi_arabia.id,
+            'state_id': self.riyadh.id,
+            'l10n_sa_api_mode': 'sandbox',
+            'currency_id': self.env.ref('base.SAR').id,
+        })
+        compliant_company.partner_id.industry_id = self.env['res.partner.industry'].create({
+            'name': 'Technology'
```
FlorianGilbert:
```suggestion
            'name': 'Technology',
```
To prevent future useless diff

FlorianGilbert (approved):
LGTM. As said before, it's fixing a traceback so ok to merge it in 17.0.

Hello @odoo/rd-accounting-security or @odoo/rd-security 
May I ask you to double check this PR? We didn't change anything related to security but by modifying something on the line, we need your approval. We probably need a security check on X509 lines to be sure that everything is fine.

Thank you guys!

xmo-odoo (commented):
@robodoo override=ci/security

FlorianGilbert (approved):
@robodoo r+

PR #208005, opened 2025-04-29
[IMP] project,sale_project,sale_timesheet,web: clean settings

deneuvillem:
This PR aims to remove some global settings to only keep them in the project form view:

1. More specifically, it removes the global settings "Recurring Tasks", "Task Dependencies" and "Milestones" from the global settings.
And moves "Recurring Tasks" and "Task Dependencies" in the project form view directly. It keeps the "Milestones" settings in the project form view as it was.

- The way groups `project.group_project_milestone`, `project.group_project_recurring_tasks` and `project.group_project_recurring_tasks` are given to the user is now different, as the global settings are removed.
It now works like this:
=> When the user enables one of the settings we mentioned (e.g. "Milestones") in any project and the related group (e.g. `project.group_project_milestone`) is not already added to the base user group, it is added.
=> If the user disables the setting from a project/remove a project which had the setting enabled, if after that disabling there is no project (even archived ones) having the setting enabled, then the related group is removed from the base user group.

- The default value given to the setting of a new project is always False.

- We remove the behaviour introduced by this commit: https://github.com/odoo/odoo/commit/50ffc339991b136b48525b2d60aa34a3039bb8da

2. The "Customer Ratings" setting was removed from the global settings and the project form view. It is moved to the task stage form view.

- If the rating status of a stage is set to "stage": when a task reaches that stage, an email will be automatically sent to get the customer rating.
  If the rating status of a stage is set to "periodic": an email will be sent periodically at the given interval as long as the task remains in the stage.
- If sale_project is installed, we only show the setting if at least one of the project of the stage is billable
- The "Rating Email Template" is now required if the setting is enabled.
- Some renaming was made in the setting description to make it clearer.
- The group `project.group_project_rating` is removed so we always make visible the customer rating in all views and in the reporting.

3. When the user enables the top bar a project which only has one visible action (e.g. "Tasks"), it now directly opens the dropdown so that the user will gain time to select the actions to show.

task-4700865

PR #244508, opened 2026-01-19
[FIX] web: fix embedded action test

deneuvillem:
In this commit, we fix an embedded action test that failed because when clicking on the sliders to show the top bar, the dropdown was directly opened as there is only one visible action
(since this commit: https://github.com/odoo/odoo/pull/208005/commits/a5709e870e033b83d50a2f581c81cceb2a98d91b).

We then adapt the test to wait for the dropdown to open automatically, and then when it's the case, create the new custom embedded action.

runbot error~237752

---


aab-odoo (approved):
robodoo r+

PR #244512, opened 2026-01-19
[FIX] im_livechat: fix non-deterministic session tour

tsm-odoo:
The `test_form_view_embed_thread` ensures that the live chat list view redirects to the discuss app and that the bus subscription for the selected channel is properly made. This ensure that a live chat agent looking at a channel will receive the latest updates for the conversation.

To ensure the bus subscription is made, the bus service is patched in order to detect calls to add or remove methods. However, we only wait for `whenReady` which is not enough to guarantee the bus is already exposed via `odoo.__WOWL_DEBUG__`.

This PR ensure we wait for the list view to be displayed, which ensures the webclient is mounted and the debug info is already set.

runbot-234928

Current behavior before PR:

---


alexkuhn (approved):
@robodoo delegate+ r+

PR #244521, opened 2026-01-19
[18.0][FIX] rating messaging menu

ap-wtioit:
The rating module does not correctly extend the mail modules onClick for NotificationItem

Current behavior before PR:
* mail NotificationItem onClick calls `this.onClickThread(isMarkAsRead, thread, message)`
* rating NotificationItem onClick calls `this.onClickThread(isMarkAsRead, thread)`
  * causing the [if in onClickThread](https://github.com/odoo/odoo/blob/18.0/addons/mail/static/src/core/public_web/messaging_menu.js#L39C17-L39C24) to be missed and falling backup to opening the thread
     ```javascript
            if (message?.needaction && message.message_type === "user_notification") {
                this.store.inbox.highlightMessage = message;
                this.openDiscussion(this.store.inbox);
                return;
            }
            this.openDiscussion(thread);
            return;
     ```

* rating should not change the fallback behaviour for messages without a rating_id and still allow opening the user_notification in inbox instead of switching to the thread

---

Info @wt-io-it

### DIFF
--- a/addons/rating/static/src/core/web/messaging_menu_patch.xml
+++ b/addons/rating/static/src/core/web/messaging_menu_patch.xml
@@ -3,7 +3,7 @@
     <t t-inherit="mail.MessagingMenu.content" t-inherit-mode="extension">
         <xpath expr="//t[@name='threads']/NotificationItem" position="attributes">
             <attribute name="rating">message?.rating_id</attribute>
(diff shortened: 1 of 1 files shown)


william-andre:
@robodoo r+

PR #244527, opened 2026-01-19
[FIX] mail: fix invite by email tour

tsm-odoo:
The `test_01_invite_by_email_flow` test ensures we can invite partners to a channel using their email.

The tour opens the invitation panel, then sends an email. However, it conflicts with the auto-open of the member list. Sometimes, the member list opens after clicking on the invite panel button.

This only occurs because the tour runs very fast. It's very unlikely that this occurs to a real user. This commit fixes the tour in order to wait for the member list to open initially.

runbot-237994

Current behavior before PR:

---


alexkuhn (approved):
@robodoo delegate+ r+

PR #244529, opened 2026-01-19
[FIX] mail: tracking UID context dependent

LauraCForgeFlow:
Fields marked with `depends_context('uid')`, such as `company_ids` in the `account.account` model in this [code](https://github.com/odoo/odoo/blob/5607dbdef771527665cf5fe3cb1a15af977c9143/addons/account/models/account_account.py#L107) fragment, can produce false chatter logs when tracking is enabled. 

This happens because the initial values of tracked fields are captured in `precommit.data` under the user environment, but later, when `_track_finalize` calls `_message_track`, the records are read under `sudo()`, see [code](https://github.com/odoo/odoo/blob/79d509a0ab43039cd989b0a70a09231a4e92ba84/addons/mail/models/mail_thread.py#L573). For context-dependent fields, this creates two cached values, one under the user and one under `sudo()`, and `_mail_track` incorrectly detects a difference, generating a tracking log even though the database value has not actually changed.

This fix modifies `_mail_track` so that the current field value is fetched under the same user environment (`sudo()` or not, if possible and if we don't get an access error when trying to do so) that was used to capture the initial value. This ensures that only real changes are tracked in the chatter, while still preventing access errors and preserving multi-company behavior.

Tests have also been added to reproduce this scenario. Without the change in the `_mail_track` method, the new test fails, but it passes successfully when we add the fix.

---

### DIFF
--- a/addons/mail/models/mail_thread.py
+++ b/addons/mail/models/mail_thread.py
@@ -539,12 +539,14 @@ class MailThread(models.AbstractModel):
         prepared with ``_tracking_prepare``.
         """
         initial_values = self.env.cr.precommit.data.pop(f'mail.tracking.{self._name}', {})
+        mail_track_with_sudo = self.env.su
         ids = [id_ for id_, vals in initial_values.items() if vals]
         if not ids:
             return
         records = self.browse(ids).sudo()
         fnames = self._track_get_fields()
         context = clean_context(self._context)
+        context.setdefault('mail_track_with_sudo', mail_track_with_sudo)
         tracking = records.with_context(context)._message_track(fnames, initial_values)
         for record in records:
             changes, _tracking_value_ids = tracking.get(record.id, (None, None))
--- a/addons/test_mail/models/test_mail_corner_case_models.py
+++ b/addons/test_mail/models/test_mail_corner_case_models.py
@@ -198,6 +198,16 @@ class MailTestMultiCompany(models.Model):
     company_id = fields.Many2one('res.company')
 
 
+class MailTestMultiCompanyUIDDependent(models.Model):
+    _name = 'mail.test.multi.company.uid.dependent'
+    _description = "Test Multi Company Mail UID Dependent"
+    _inherit = 'mail.thread.main.attachment'
+
+    name = fields.Char(tracking=True)
+    company_id = fields.Many2one('res.company', tracking=True)
+    company_ids = fields.Many2many('res.company', depends_context=('uid',), tracking=True, string="Companies")
+
+
 class MailTestMultiCompanyRead(models.Model):
     """ Just mail.test.simple, but multi company and supporting posting
     even if the user has no write access. """
--- a/addons/mail/models/models.py
+++ b/addons/mail/models/models.py
@@ -8,6 +8,7 @@ from markupsafe import Markup
 
 from odoo import api, exceptions, models, tools, _
 from odoo.addons.mail.tools.alias_error import AliasError
+from odoo.exceptions import AccessError
 
 import logging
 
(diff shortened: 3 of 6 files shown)


reth-odoo (commented):
Seems a bit unfortunate that flushing does not sync or wipe caches of envs on the same transaction. Could be an oversight, though I suppose it's not generally an issue. And maybe we do want to keep caches untouched.

I'm not sure I'm getting the fix however. Is sudo actually relevant? Or is it that because sudo creates a new env, it has to fetch the flushed values again which will match those of the "correct" cache?

"try without sudo first" seems too specific of a change for this issue, imo.

LauraCForgeFlow:
Hello @reth-odoo, thanks for taking a look! You’re right that the root issue is not sudo itself, but the fact that sudo creates a new env, with its own cache, within the same transaction.

The problem arises because `_tracking_prepare` captures initial values under the user env, while `_track_finalize` later reads the record under a different env (created by sudo). For context-dependent fields (`depends_context('uid')`), this results in two different cached values for the same field, which `_mail_track` then incorrectly interprets as a change. 

Basically, the initial value of the field (user env) is not the same as the current value of the field (sudo env, and therefore another cached value), so it tracks this as a change. But of course, this is incorrect. And it only happens for these kinds of fields that are context or env dependent, so the cached value can vary depending on the used env.

The fix is therefore about ensuring that the “current” value is evaluated under the same env semantics as the initial value, whenever possible. This change restores that symmetry and avoids the false positive. The fallback to sudo is only there to guarantee flush safety in cases where the user no longer has read access at finalize time, it's not part of the primary mechanism.

I went with this approach because the issue I encountered was specifically caused by sudo / non-sudo env discrepancies for uid context dependent fields in the mail tracking workflow that I explained. If you have a broader or cleaner way to address env cache divergence in this context, I’m open to exploring that as well.

addons/test_mail/models/test_mail_corner_case_models.py
```diff
@@ -198,6 +198,16 @@ class MailTestMultiCompany(models.Model):
...
 
 
+class MailTestMultiCompanyUIDDependent(models.Model):
+    _name = 'mail.test.multi.company.uid.dependent'
+    _description = "Test Multi Company Mail UID Dependent"
+    _inherit = 'mail.thread.main.attachment'
+
+    name = fields.Char(tracking=True)
+    company_id = fields.Many2one('res.company', tracking=True)
+    company_ids = fields.Many2many('res.company', depends_context=('uid',), tracking=True, string="Companies")
```
LauraCForgeFlow, line 208: Right, this ensures that the company_ids field gets 1 cache value per env (in this case related to uid), so fetching the value as a regular user can give us different results than doing it as the sudo user. It is also briefly explained in the actual test function with comments.

reth-odoo: Forgot to remove this comment before posting, indeed this is explained in the PR message ^^

-> But actually it doesn't seem necessary 🤔

addons/mail/models/mail_thread.py
```diff
@@ -539,12 +539,14 @@ def _track_finalize(self):
...
         """
         initial_values = self.env.cr.precommit.data.pop(f'mail.tracking.{self._name}', {})
+        mail_track_with_sudo = self.env.su
         ids = [id_ for id_, vals in initial_values.items() if vals]
         if not ids:
             return
         records = self.browse(ids).sudo()
         fnames = self._track_get_fields()
         context = clean_context(self._context)
+        context.setdefault('mail_track_with_sudo', mail_track_with_sudo)
```
reth-odoo, line 549: I think it's a bit confusing as it doesn't really set the su state and won't necessarily actually use the value (just attempt to)
```suggestion
        context.setdefault('mail_track_try_unpriviledged', mail_track_with_sudo)
```

addons/mail/models/mail_thread.py
```diff
@@ -539,12 +539,14 @@ def _track_finalize(self):
         prepared with ``_tracking_prepare``.
         """
         initial_values = self.env.cr.precommit.data.pop(f'mail.tracking.{self._name}', {})
+        mail_track_with_sudo = self.env.su
```
reth-odoo, line 542: One issue is that this isn't necessarily always an indication of what the state of the recordset was originally. As precommit data can be added to the queue, but will always be processed with the env of the first recordset of this model to have been processed by `_track_prepare`.

Ideally, if we wanted to do this, we should probably include some env state (context + su + uid) for each record in "initial values" for how they were fetched. And then always try to fetch with that state first.

reth-odoo (commented):
I see. So in the end we use sudo() in _track_finalize but in practice we often just reuse the values from cache which did not bypass access checks so it happens to work out most of the time.

It seems like in theory you would have the same issue if the field you're tracking is somehow evicted from cache before it's flushed too? Which I guess is pretty rare.

-----------------------------------------------------

Honestly I would say that the "initial value" should be fetched with sudo to make sure we never apply access rules for initial but not final values. And we should clear the cache in tracking.

As a general rule I would say tracking should pretty much represent the data exactly as it is in DB. We don't care about context or access rules.

To see with @tde-banana-odoo when he gets back.
Feel free to ping again next week.

-----------------------

Thanks for the PR, test and explanation :)

PR #244530, opened 2026-01-19
[IMP] iot_drivers: allow testing every driver

loouis-t:
We create a standard status method on drivers to allow getting status (clicking "Test" in the database).
Every driver can now override this method to test the device its own way.

see odoo/enterprise#104794
Task: 5491079

### DIFF
--- a/addons/iot_drivers/iot_handlers/drivers/ctypes_terminal_driver.py
+++ b/addons/iot_drivers/iot_handlers/drivers/ctypes_terminal_driver.py
@@ -1,7 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 import ctypes
-import datetime
 import logging
 from abc import abstractmethod, ABC
 from queue import Queue
@@ -159,7 +158,6 @@ class CtypesTerminalDriver(Driver, ABC):
         self.terminal_busy = False
 
         self._actions[''] = self._action_default
-        self.next_transaction_min_dt = datetime.datetime.min
 
         self.device_manufacturer = manufacturer
         self.device_name = f"{self.device_manufacturer} terminal {self.device_identifier}"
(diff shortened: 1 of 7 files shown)


addons/iot_drivers/iot_handlers/drivers/ctypes_terminal_driver.py
```diff
@@ -149,3 +149,18 @@ def six_terminal_balance(self, transaction):
         Method implementing the terminal balance request (only for Worldline "Six")
         Not an abstract method as it remains undefined for Worldline
         """
+
+    def status(self, _):
+        self._action_default({
+            "messageType": "Transaction",
+            "transactionType": "Payment",
```
Ysoroko, line 248: `TransactionID` and `actionIdentifier` keys/values are missing for Worldline.
```suggestion
            "transactionType": "Payment",
            "TransactionID": 1,
            "actionIdentifier": 1,
```

Ysoroko (changes requested):
With the 1st commit only you could crash Odoo on the iot box.
For Worldline when "test" button was spammed (or even regularly pressed)
1) A transaction is put in a queue
2) A cancellation is executed directly, not necessarily waiting for the transaction to start
3) If cancellation starts before transaction --> crash

A 2nd commit was added which
a) Ignores the cancellation requests if there are no ongoing transactions
b) Adds a sleep between transaction and cancellation to accound for the queue waiting time
c) Removed unnecessary `_check_transaction_delay` which is already handled by `terminal_busy` variable and was causing cancellations before transaction for the test button

addons/iot_drivers/iot_handlers/drivers/ctypes_terminal_driver.py
Ysoroko: to remove

addons/iot_drivers/iot_handlers/drivers/ctypes_terminal_driver.py
```diff
@@ -256,3 +241,21 @@ def six_terminal_balance(self, transaction):
...
+            "transactionID": 1,
```
Ysoroko:
```suggestion
            "TransactionID": 1,
```

loouis-t:
robodoo help

PR #244531, opened 2026-01-19
[IMP] web: offline searchbar

aab-odoo:
This PR introduces an offline-specific version of the search
bar, in kanban and list views. The offline search bar allows to
select previous searches, for which we know that we have data in
cache. Those searches are sorted by count (# of times they have
been done) and by last visited date. It only displays searches for
which there's a result (groups or records), i.e. no searches that
would lead to an empty screen with the no content helper.

We rely on the offline service (and its `visited-ui-items` idb
table) to store all searches that are available offline. For each
search, we store the description of its facets, to be able to
render them both in the search menu dropdown and in the search bar
when that search is selected, and we store the domain and groupby
of the search, to apply in the search model when that search is
selected.

Note: storing the domain and not a description of the selected
filters allows to ensure that we can restore a search even if the
search view (arch) or favorite items changed meanwhile.

Task~5493339

### DIFF
--- a/addons/web/static/src/core/offline/offline_service.js
+++ b/addons/web/static/src/core/offline/offline_service.js
@@ -84,14 +84,15 @@ class OfflineManager extends Reactive {
             this._offlineUI();
             // Create an observer instance linked to the callback function to keep disabling
             // elements that would appear in the DOM while being offline.
-            this._observer = new MutationObserver((mutationList) => {
-                if (this._offline && mutationList.find((m) => m.addedNodes.length > 0)) {
+            this._observer = new MutationObserver(() => {
+                if (this._offline) {
                     this._offlineUI();
                 }
             });
             this._observer.observe(document.body, {
                 childList: true, // listen for direct children being added/removed
                 subtree: true, // also observe descendants (not just direct children)
+                attributeFilter: ["data-available-offline"], // listen for specific attribute change
             });
 
             // Repeatedly check if connection is back.
@@ -187,6 +188,11 @@ class OfflineManager extends Reactive {
      * @private
      */
     _offlineUI() {
+        // Re-enable elements that have been marked as available offline
+        document.querySelectorAll(".o_disabled_offline[data-available-offline]").forEach((el) => {
+            el.removeAttribute("disabled");
+            el.classList.remove("o_disabled_offline");
+        });
         document.querySelectorAll(OfflineManager.SELECTORS_TO_DISABLE.join(", ")).forEach((el) => {
             el.setAttribute("disabled", "");
             el.classList.add("o_disabled_offline");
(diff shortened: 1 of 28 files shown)


addons/web/static/src/core/offline/offline_service.js
```diff
@@ -107,19 +111,18 @@ class OfflineManager extends Reactive {
...
+                    const { action, viewType, resId } = JSON.parse(key);
+                    this._visited[action] = this._visited[action] || { views: {} };
+                    if (viewType === "form") {
+                        this._visited[action].views.form = this._visited[action].views.form || [];
+                        this._visited[action].views.form.push(resId);
                     } else {
-                        this._visited[value.action].views[value.viewType] = true;
+                        this._visited[action].views[viewType] = true;
                     }
                 }
```
jpp-odoo: refactor in a private function ?

addons/web/static/src/core/offline/offline_service.js
```diff
@@ -169,24 +199,53 @@ class OfflineManager extends Reactive {
...
      */
-    async setAvailableOffline(actionId, viewType, { resId }) {
+    async setAvailableOffline(actionId, viewType, { resId, search }) {
         if (!this.offline) {
-            const key = JSON.stringify({ action: actionId, viewType, resId });
-            this._idb.write(this._idbTable, key, true);
+            const key = this._generateKey(actionId, viewType, resId);
+            let value;
+            if (viewType === "form") {
+                value = true; // TODO: store resIds as value?
```
jpp-odoo: Not sure it's a good idea, If we store `resIds` as values the `isAvailableOffline` function for the `resIds` will become async.

jpp-odoo (approved):

addons/web/static/src/core/offline/offline_service.js
aab-odoo: Yeah I wrote this when I still used idb.getEntries, so everything was loaded in ram. It's probably no longer relevant, I'll remove the comment

aab-odoo:
robodoo rebase-ff

addons/web/static/src/search/search_model.js
```diff
@@ -501,6 +519,24 @@ export class SearchModel extends EventBus {
...
     }
 
+    /**
+     * Apply the given search. Typically called offline, with a Search that has
+     * been previously obtained through @getCurrentSearch.
+     *
+     * @param {Search} search
+     */
+    applySearch(search) {
+        this.query = []; // remove everything
```
Polymorphe57, line 529: use blockNotification to avoid intermediate notify

addons/web/static/src/search/search_model.js
```diff
@@ -2480,6 +2529,24 @@ export class SearchModel extends EventBus {
         );
     }
 
+    _toggleGroupBys(groupBys) {
+        for (const raw of groupBys) {
+            const [field, interval] = raw.split(":"); // eg. grouping on "dateorder:quarter"
```
Polymorphe57, line 2534: does it work with properties?

aab-odoo: astonishingly, it does

addons/web/static/src/search/search_model.js
aab-odoo: I don't think it's necessary

PR #244532, opened 2026-01-19
[FIX] portal: fix address form with hidden relational input values

delcourtfl:
**Steps to reproduce:**
- Create a DB with l10n_ar and l10n_ar ecommerce
- Go to Settings > Invoicing > Fiscal Localization
- Select the 'Argentina - Argentine Generic Chart of Accounts for Registered Accountants' package
- Create a website, and a portal user
- Add a main and a secondary address to the user using the website form in 'My Account'
- Go to the secondary address and change any field
- Click on Save Address
- Multiple errors will appear on the form (reproducible with other similar config)
(and in logs: `UserWarning: unsupported operand type(s) for "==": 'l10n_latam.identification.type()' == '1'`)

**Issue:**
In `address_form_fields` some hidden input field are used to add specific non-editable values to the forms.

This breaks the address validation of `CustomerPortal` in `def _validate_address_values` due to the following comparison: `partner_sudo[commercial_field_name] != address_values[commercial_field_name]`
which try to compare recordsets with the given ids.

**Fix:**
Cast relational field to their id values to ensure they can be properly compared to the website form values.

related fix which introduces the input issue: https://github.com/odoo/odoo/commit/0ee91631214c34b650342a0210ee8db27764f252

opw-5247171

### DIFF
--- a/addons/portal/controllers/portal.py
+++ b/addons/portal/controllers/portal.py
@@ -680,16 +680,22 @@ class CustomerPortal(Controller):
             # commercial partner values, and would be reset if modified on the commercial partner.
             if not (is_commercial_address := partner_sudo == partner_sudo.commercial_partner_id):
                 for commercial_field_name in partner_sudo._commercial_fields():
+                    if commercial_field_name not in address_values:
+                        continue
+                    partner_sudo_field = partner_sudo._fields[commercial_field_name]
+                    partner_sudo_value = partner_sudo_field.convert_to_cache(
+                        partner_sudo[commercial_field_name],
+                        partner_sudo,
+                    )
                     if (
-                        commercial_field_name in address_values
-                        and partner_sudo[commercial_field_name] != address_values[commercial_field_name]
+                        partner_sudo_value != address_values[commercial_field_name]
                         and (
-                            bool(partner_sudo[commercial_field_name])
+                            bool(partner_sudo_value)
                             or bool(address_values[commercial_field_name])
                         )
                     ):
                         invalid_fields.add(commercial_field_name)
-                        field_description = partner_sudo._fields[commercial_field_name]._description_string(request.env)
+                        field_description = partner_sudo_field._description_string(request.env)
                         if partner_sudo.commercial_partner_id.is_company:
                             error_messages.append(_(
                                 "The %(field_name)s is managed on your company account.",


addons/portal/controllers/portal.py
```diff
@@ -680,16 +680,22 @@ def _validate_address_values(
             # commercial partner values, and would be reset if modified on the commercial partner.
             if not (is_commercial_address := partner_sudo == partner_sudo.commercial_partner_id):
                 for commercial_field_name in partner_sudo._commercial_fields():
+                    partner_sudo_field = partner_sudo._fields[commercial_field_name]
+                    partner_sudo_value = partner_sudo_field.convert_to_read(
```
Ephoniz: How's the performance impact of executing this for every field? Maybe we can move it to an if and only execute it for the needed fields? 🤔

Ephoniz (approved):
Makes sense 👍🏽

addons/portal/controllers/portal.py
delcourtfl: Not sure of the performance impact for every type of field but we could filter with something like:
```py
if partner_sudo_field.relational:
```
(But I am not sure it's really better)

Ephoniz: Wouldn't it work with something like  `if commercial_field_name in address_values`

addons/portal/controllers/portal.py
```diff
@@ -680,16 +680,24 @@ def _validate_address_values(
             # commercial partner values, and would be reset if modified on the commercial partner.
             if not (is_commercial_address := partner_sudo == partner_sudo.commercial_partner_id):
                 for commercial_field_name in partner_sudo._commercial_fields():
+                    if not commercial_field_name in address_values:
+                        address_values.pop(commercial_field_name, None)
+                        continue
+                    partner_sudo_field = partner_sudo._fields[commercial_field_name]
+                    partner_sudo_value = partner_sudo_field.convert_to_read(
```
reth-odoo: Seems better to convert_to_cache the incoming value instead, or really just browse if the issue is just with relational fields. As then it will not depend on ordering. And just globally convert_to_cache will give you the best possible equality method for the field.

delcourtfl: I think you meant the current value as the incoming one already uses `convert_to_cache` in `_parse_form_data` to get the id.

addons/portal/controllers/portal.py
```diff
@@ -680,16 +680,23 @@ def _validate_address_values(
             # commercial partner values, and would be reset if modified on the commercial partner.
             if not (is_commercial_address := partner_sudo == partner_sudo.commercial_partner_id):
                 for commercial_field_name in partner_sudo._commercial_fields():
+                    if not commercial_field_name in address_values:
+                        address_values.pop(commercial_field_name, None)
```
tde-banana-odoo: Why pop if commercial_field_name is not in address_values ?

addons/portal/controllers/portal.py
```diff
@@ -680,16 +680,23 @@ def _validate_address_values(
             # commercial partner values, and would be reset if modified on the commercial partner.
             if not (is_commercial_address := partner_sudo == partner_sudo.commercial_partner_id):
                 for commercial_field_name in partner_sudo._commercial_fields():
+                    if not commercial_field_name in address_values:
```
tde-banana-odoo:
```suggestion
                    if commercial_field_name not in address_values:
```

tde-banana-odoo (commented):
Zboing !

addons/portal/controllers/portal.py
delcourtfl: Oups, removed it (probably some leftovers from another version of the fix)

PR #244533, opened 2026-01-19
[IMP] base: add access tokens API

thle-odoo:
Currently, when a user does not have sufficient access rights to one or more records, we grant elevated privileges to access the record(s) if a verified access token is presented.

However, each module that requires the use of an access token implements the logic for these tokens independently. The result is that there are many differences in how access tokens are managed.

This commit introduces an API that maintains a standard for access token management, thereby avoiding disparities and bad practices. From a design perspective, it is much simpler to have all access tokens centralised.

From a technical point of view, all access tokens will be managed in the `ir.access.token` model.

In order to provide a simple API for developers, the best solution is to expose the token management methods directly on the `BaseModel`.

A token is linked to a scope. This allows for greater granularity in granting privileges following token verification. An expiry date can be provided to force a token to expire. Furthermore, a token can be specific to a particular user.

Tokens generated with this API are not stored in the database. However, in order to avoid invalidating existing tokens and to maintain compatibility, a `manual_token` column is provided for this purpose.

Task-4518427

### DIFF
--- a/odoo/addons/base/models/ir_access_token.py
+++ b/odoo/addons/base/models/ir_access_token.py
@@ -1,0 +1,265 @@
+import hashlib
+import json
+import logging
+from datetime import timedelta
+from typing import Any
+
+from odoo import api, fields, models
+from odoo.exceptions import AccessError, UserError, ValidationError
+from odoo.fields import Domain
+from odoo.models import BaseModel
+from odoo.tools.constants import GC_UNLINK_LIMIT
+from odoo.tools.misc import consteq, hash_sign, verify_hash_signed
+from odoo.tools.sql import SQL, column_exists, create_column
+
+
+_logger = logging.getLogger(__name__)
+
+
+class IrAccessToken(models.Model):
+    _name = 'ir.access.token'
+    _description = 'Access Token'
+
+    res_model = fields.Char('Model', required=True, readonly=True)
+    res_id = fields.Many2oneReference('Record', model_field='res_model', required=True, readonly=True)
+    scope = fields.Char('Scope', required=True, readonly=True)
+    expiration = fields.Datetime('Expiration', readonly=True, index='btree')
+    owner_id = fields.Many2one('res.users', readonly=True, ondelete='cascade')
+    json_extra = fields.Char('JSON Extra', readonly=True)
+
+    _res_idx = models.Index('(res_model, res_id)')
+
+    def _auto_init(self):
+        super()._auto_init()
+        if not column_exists(self.env.cr, 'ir_access_token', 'manual_token'):
+            create_column(self.env.cr, 'ir_access_token', 'manual_token', 'VARCHAR')
+
+    @api.autovacuum
+    def _gc_access_token(self):
+        self.search(
+            [('expiration', '<', self.env.cr.now())],
+            limit=GC_UNLINK_LIMIT,
+        ).unlink()
+
+    def write(self, vals):
+        raise UserError(self.env._("Cannot update access tokens"))
+
+    def action_open_record(self):
+        self.ensure_one()
+        return {
+            'type': 'ir.actions.act_window',
+            'res_model': self.res_model,
+            'res_id': self.res_id,
+            'view_mode': 'form',
+            'target': 'current',
+        }
+
+    @api.model
+    def __generate(
+        self,
+        res_model: str,
+        res_id: int,
+        scope: str,
+        duration: timedelta | None = None,
+        owner_id: int | None = None,
+        extra: Any = None,
+        manual_token: str | None = None,
+    ) -> str:
+        """ Generate an access token.
+
+        :param res_model: target model name.
+        :param res_id: record id.
+        :param scope: access scope associated with the token.
+        :param duration: optional expiration timedelta for the token,
+                         if `None`, the token does not expire.
+        :param owner_id: optional id identifying the token owner.
+        :param extra: optional extra information stored in the token (must be serializable).
+        :param manual_token: optional token value which must be forced.
+        :return: signed access token string.
+        """
+        assert not (extra and manual_token), "A manual token cannot contain additional information."
+        assert self.env.su, "Sudo environment is required to generate access token."
+
+        expiration = None
+        if duration is not None:
+            if duration.total_seconds() <= 0:
+                raise ValidationError(self.env._("Duration must be positive."))
+            expiration = self.env.cr.now() + duration
+
+        json_extra = json.dumps(extra) if extra is not None else None
+
+        access_token = self.create({
+            'res_model': res_model,
+            'res_id': res_id,
+            'scope': scope,
+            'expiration': expiration,
+            'owner_id': owner_id,
+            'json_extra': json_extra,
+        })
+
+        if manual_token:
+            self.env.cr.execute(SQL("""
+                UPDATE %(table)s SET manual_token = %(manual_token)s WHERE id = %(id)s
+            """,
+                table=SQL.identifier(self._table),
+                manual_token=manual_token,
+                id=access_token.id,
+            ))
+
+        _logger.info(
+            "User %d generates access tokens (%s): %s, %s, %s, %s, %s (%s)",
+            self.env.uid, access_token.id, res_model, res_id, scope, expiration, owner_id,
+            'manual' if manual_token else 'auto',
+        )
+
+        # Include the access token ID in the signed payload to ensure we have a
+        # different token if we need to regenerate one for the same record.
+        return manual_token or hash_sign(
+            self.env, scope, (access_token.id, res_model, res_id, owner_id, json_extra and hashlib.sha256(json_extra.encode()).hexdigest()), expiration
+        )
+
+    @api.model
+    def __retrieve_sudo_record(
+        self,
+        model: str, scope: str, token: str, *,
+        res_model: str | None = None,
+        res_id: int | None = None,
+    ) -> BaseModel:
+        """ Get the referenced sudoed record by the token.
+
+        :param model: target model name.
+        :param scope: access scope associated with the token.
+        :param token: access token identifying the record.
+        :param res_model: model name (required for manual token or retrieve from the payload)
+        :param res_id: record id (required for manual token or retrieve from the payload)
+        :return: sudoed record.
+        :raises AccessError:
+            - if the token is invalid, expired, or does not match the expected scope or payload,
+            - if the model record doesn't match the target model,
+            - if the owner of the token doesn't match the user environment,
+            - if the token has been revoked.
+        """
+        assert self.env.su, "Sudo environment is required to retrieve a record from access token."
+
+        try:
+            payload = verify_hash_signed(self.env, scope, token)
+        except ValueError:
+            assert res_model and res_id, "Manual tokens require record identification."
+
+            query = self._search([
+                ('res_model', '=', res_model),
+                ('res_id', '=', res_id),
+                ('scope', '=', scope),
+                '|', ('owner_id', '=', False), ('owner_id', '=', self.env.uid),
+                '|', ('expiration', '=', False), ('expiration', '>', self.env.cr.now()),
+            ], order='owner_id NULLS FIRST, expiration DESC NULLS FIRST, id')
+
+        else:  # Correct token format
+            if payload is None:  # Token is invalid
+                raise AccessError(self.env._("Invalid token: incorrect scope or expired."))
+
+            access_token_id, res_model, res_id, owner_id, _ = payload
+            if owner_id and owner_id != self.env.uid:
+                raise AccessError(self.env._("Invalid token: incorrect owner."))
+
+            query = self._search([('id', '=', access_token_id)])
+
+        if res_model != model:
+            raise AccessError(self.env._("Invalid token: incorrect model."))
+
+        query = query.select(
+            query.table.id,
+            query.table.res_model,
+            query.table.res_id,
+            query.table.scope,
+            query.table.expiration,
+            query.table.owner_id,
+            query.table.json_extra,
+            SQL("%s", SQL.identifier(query.table._alias, 'manual_token')),
+        )
+
+        for access_token_id, res_model, res_id, scope, expiration, owner_id, json_extra, manual_token in self.env.execute_query(query):
+            token_from_db = manual_token or hash_sign(
+                self.env, scope, (access_token_id, res_model, res_id, owner_id, json_extra and hashlib.sha256(json_extra.encode()).hexdigest()), expiration
+            )
+            if consteq(token, token_from_db):
+                return self.env[model].browse(res_id).with_context(access_token_extra=json_extra and json.loads(json_extra))
+
+        raise AccessError(self.env._("Invalid token: expired (revoked) or invalid."))
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    def _grant_access_token(
+        self,
+        scope: str,
+        *,
+        duration: timedelta | None = None,
+        owner: BaseModel | None = None,
+        extra: Any = None,
+        _manual_token: str | None = None,
+    ) -> str:
+        """ Generate and get access token for the record. """
+        assert owner is None or owner._name == 'res.users'
+
+        self.ensure_one()
+        return self.env['ir.access.token']._IrAccessToken__generate(
+            self._name, self.id, scope,
+            duration, owner.id if owner else None,
+            extra,
+            _manual_token,
+        )
+
+    @api.model
+    def _get_sudo_record_from_access_token(
+        self,
+        scope: str,
+        token: str,
+        *,
+        record: BaseModel | None = None
+    ) -> BaseModel:
+        """ Retrieve the sudoed record referenced by the token. """
+        return self.env['ir.access.token']._IrAccessToken__retrieve_sudo_record(
+            self._name, scope, token,
+            res_model=record._name if record else None,
+            res_id=record.id if record else None,
+        )
+
+    @api.model
+    def _revoke_access_tokens(
+        self,
+        scope: str,
+        *,
+        owners: BaseModel | None = None,
+        revoke_shared: bool = False,
+    ) -> int:
+        """ Ensures that tokens for the records and scope are invalidated.
+
+        :param scope: access scope associated with the tokens.
+        :param owners: if `res.users` records, delete access tokens for these users.
+        :param revoke_shared: if `True`, delete shared access tokens (without user).
+        :return: number of deleted access tokens.
+        """
+        assert owners is None or owners._name == 'res.users'
+        assert self.env.su, "Sudo environment is required to revoke access tokens."
+
+        domain = Domain([
+            ('res_model', '=', self._name),
+            ('res_id', 'in', self._ids),
+            ('scope', '=', scope),
+        ])
+
+        owner_domain = Domain.FALSE
+        if owners is not None:
+            owner_domain |= Domain('owner_id', 'in', owners.ids)
+        if revoke_shared:
+            owner_domain |= Domain('owner_id', '=', False)
+        if not owner_domain.is_false():
+            domain &= owner_domain
+
+        access_tokens = self.env['ir.access.token'].search(domain)
+        if access_tokens:
+            access_tokens.unlink()
+            _logger.info("User %d revokes access tokens (%s)", self.env.uid, ', '.join(map(str, access_tokens._ids)))
+        return len(access_tokens)
--- a/odoo/addons/base/tests/test_ir_access_token.py
+++ b/odoo/addons/base/tests/test_ir_access_token.py
@@ -1,0 +1,264 @@
+from datetime import datetime, timedelta
+
+from freezegun import freeze_time
+from functools import partial
+
+from odoo.exceptions import AccessError, ValidationError
+from odoo.sql_db import BaseCursor
+from odoo.tests.common import new_test_user, TransactionCase
+
+
+class TestIrAccessToken(TransactionCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.enterClassContext(cls.registry_test_mode())
+
+        cls.IrAccessToken = cls.env['ir.access.token']
+
+        cls.admin = new_test_user(cls.env, 'Test Admin', groups='base.group_erp_manager')
+        cls.user_1 = new_test_user(cls.env, 'Test User 1')
+        cls.user_2 = new_test_user(cls.env, 'Test User 2')
+
+        cls.Partner = cls.env['res.partner'].with_user(cls.admin)
+        cls.partner = cls.Partner.create({'name': 'Test partner_sudo'})
+
+        cls.PartnerSudo = cls.Partner.sudo()
+        cls.partner_sudo = cls.partner.sudo()
+
+        cls.classPatch(BaseCursor, 'now', lambda cr: datetime.now())
+
+        cls.IrAccessToken.search([]).unlink()
+
+    def test_grant_access_to_record(self):
+        with self.assertRaises(AssertionError):  # Sudo environment required to grant access tokens
+            self.partner._grant_access_token('test.partner_phone_number_read')
+
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        self.assertTrue(token, 'An access token must be generated and returned')
+
+        with self.assertRaises(AssertionError):  # Sudo environment required to retrieve partner_sudos
+            self.Partner._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'The partner_sudo must be found')
+
+    def test_grant_access_to_record_with_manual_token(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', _manual_token='123456789')
+        self.assertEqual(token, '123456789', 'The raw access token must be returned')
+
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', '123456789', record=self.partner_sudo)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'The partner_sudo must be found')
+
+    @freeze_time('2026-01-01')
+    def test_grant_access_to_record_with_duration(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', duration=timedelta(days=10))
+
+        with freeze_time('2026-01-09'):
+            partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+            self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        with freeze_time('2026-01-11'), self.assertRaises(AccessError):  # The token is invalid
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+    @freeze_time('2026-01-01')
+    def test_grant_access_to_record_with_expiration_in_past(self):
+        with self.assertRaises(ValidationError):
+            self.partner_sudo._grant_access_token('test.partner_phone_number_read', duration=timedelta(days=-1))
+
+    def test_grant_access_to_record_with_owner(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', owner=self.user_1)
+
+        with self.assertRaises(AccessError):  # With user 2, the token is invalid
+            self.PartnerSudo.with_user(self.user_2).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+        partner_sudo = self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token is valid for the owner user_1')
+
+    def test_grant_access_to_record_with_scope(self):
+        token_read = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        token_write = self.partner_sudo._grant_access_token('test.partner_phone_number_write')
+
+        with self.assertRaises(AccessError):  # Token incorrect for the scope
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_write', token_read)
+        with self.assertRaises(AccessError):  # Token incorrect for the scope
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_write)
+
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_write', token_write)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_read)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+    @freeze_time('2026-01-01')
+    def test_ensure_database_integrity_expiration(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', duration=timedelta(days=10))
+
+        with freeze_time('2026-01-09'):
+            partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+            self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+            # Alter DB - Attempting to extend a token
+            access_token = self.IrAccessToken.search([], limit=1)
+            self.env.registry['base'].write(access_token, {'expiration': datetime(2026, 1, 20)})  # Bypass write override
+            with self.assertRaises(AccessError):  # We detect a modification in the expiration
+                self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+    def test_ensure_database_integrity_owner(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', owner=self.user_1)
+        partner_sudo = self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+        # Alter DB - Attempting to change user
+        access_token = self.IrAccessToken.search([], limit=1)
+        self.env.registry['base'].write(access_token, {'owner_id': self.user_2})  # Bypass write override
+        with self.assertRaises(AccessError):  # We detect a modification in the user
+            self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        with self.assertRaises(AccessError):  # Token completely invalid (even if we use the updated user)
+            self.PartnerSudo.with_user(self.user_2).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+    def test_ensure_database_integrity_scope(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+        # Alter DB - Attempting to change scope
+        access_token = self.IrAccessToken.search([], limit=1)
+        self.env.registry['base'].write(access_token, {'scope': 'test.partner_phone_number_write'})  # Bypass write override
+        with self.assertRaises(AccessError):  # We detect a modification in the scope
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        with self.assertRaises(AccessError):  # Token completely invalid (even if we use the updated scope)
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_write', token)
+
+    def test_revoke_access_to_record(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+        with self.assertRaises(AssertionError):  # Sudo environment required to revoke access tokens
+            self.partner._revoke_access_tokens('test.partner_phone_number_read')
+        self.partner_sudo._revoke_access_tokens('test.partner_phone_number_read')
+
+        with self.assertRaises(AccessError):  # The token is invalid
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+
+    def test_revoke_access_to_record_scope(self):
+        token_read = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        token_write = self.partner_sudo._grant_access_token('test.partner_phone_number_write')
+
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_read)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_write', token_write)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+        self.partner_sudo._revoke_access_tokens('test.partner_phone_number_write')
+
+        with self.assertRaises(AccessError):  # The token is invalid
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_write', token_write)
+
+        self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_read)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+    def test_revoke_access_to_record_scope_concurrent(self):
+        token_1 = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        token_2 = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        self.partner_sudo._revoke_access_tokens('test.partner_phone_number_read')
+        with self.assertRaises(AccessError):
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+        with self.assertRaises(AccessError):
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+
+    def test_revoke_access_to_record_according_user(self):
+        token_1 = self.partner_sudo._grant_access_token('test.partner_phone_number_read', owner=self.user_1)
+        token_2 = self.partner_sudo._grant_access_token('test.partner_phone_number_read', owner=self.user_2)
+        partner_sudo = self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+        partner_sudo = self.PartnerSudo.with_user(self.user_2).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        self.partner_sudo._revoke_access_tokens('test.partner_phone_number_read', owners=self.user_1)
+        with self.assertRaises(AccessError):
+            self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+        partner_sudo = self.PartnerSudo.with_user(self.user_2).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must always be valid')
+
+    def test_revoke_access_to_record_according_share(self):
+        token_shared = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        token_user = self.partner_sudo._grant_access_token('test.partner_phone_number_read', owner=self.user_1)
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_shared)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+        partner_sudo = self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_user)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        self.partner_sudo._revoke_access_tokens('test.partner_phone_number_read', revoke_shared=True)
+        with self.assertRaises(AccessError):
+            self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_shared)
+        partner_sudo = self.PartnerSudo.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('test.partner_phone_number_read', token_user)
+        self.assertEqual(partner_sudo, self.partner_sudo, 'Token must always be valid')
+
+    @freeze_time('2026-01-01')
+    def test_invalidate_concurrent_tokens_according_expiration(self):
+        token_1 = self.partner_sudo._grant_access_token('test.partner_phone_number_read', duration=timedelta(days=10))
+        token_2 = self.partner_sudo._grant_access_token('test.partner_phone_number_read', duration=timedelta(days=20))
+
+        with freeze_time('2026-01-05'):
+            partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+            self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+            partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+            self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        with freeze_time('2026-01-15'):
+            with self.assertRaises(AccessError):
+                self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+            partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+            self.assertEqual(partner_sudo, self.partner_sudo, 'Token must be valid')
+
+        with freeze_time('2026-01-25'):
+            with self.assertRaises(AccessError):
+                self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_1)
+            with self.assertRaises(AccessError):
+                self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token_2)
+
+    def test_retrieve_record_from_concurrent_token(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        self.partner_sudo._grant_access_token('test.partner_phone_number_read')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+        # Using manual token
+        # search domain does not use the access token ID, so filtering must be performed
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', _manual_token='123456789')
+        self.partner_sudo._grant_access_token('test.partner_phone_number_read', _manual_token='abcdefghi')
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token, record=self.partner_sudo)
+        self.assertEqual(partner_sudo, self.partner_sudo)
+
+    def test_place_extra_information_in_token(self):
+        token = self.partner_sudo._grant_access_token('test.partner_phone_number_read', extra=self.partner_sudo.id)
+        partner_sudo = self.PartnerSudo._get_sudo_record_from_access_token('test.partner_phone_number_read', token)
+        self.assertEqual(partner_sudo.env.context['access_token_extra'], self.partner_sudo.id)
+
+        # Try to use a non-serializable object
+        with self.assertRaises(TypeError):
+            self.partner_sudo._grant_access_token('test.partner_phone_number_read', extra={self.partner_sudo.id})
+
+    def test_falsy_extra_information_in_token(self):
+        grant_token = partial(self.partner_sudo._grant_access_token, 'test.partner_phone_number_read')
+
+        def get_context_extra(token):
+            return self.PartnerSudo._get_sudo_record_from_access_token(
+                'test.partner_phone_number_read', token
+            ).env.context['access_token_extra']
+
+        token_empty = grant_token()
+        token_none = grant_token(extra=None)
+        token_false = grant_token(extra=False)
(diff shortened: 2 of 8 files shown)


addons/survey/models/survey_survey.py
```diff
@@ -207,6 +198,15 @@ def default_get(self, fields):
         'A positive default time limit is required when the session rewards quick answers.',
     )
 
+    @api.depends_context('uid')
+    def _compute_access_token(self):
```
kmagusiak: You could define `_compute_access_token(self, scope, field_name='access_token')` in the mixin and just call it in `access_token = field.Char(compute=lambda s: s._compute_access_token('access-survey'))`.

kmagusiak (commented):
I am mitigated between this and the other PR. This seems to have a slightly clearer API, but there will be the question of in which module the `_inherit` should be located (we could alternatively add in on the base model directly).

addons/survey/models/survey_survey.py
thle-odoo: I moved the API to `BaseModel` (along with the compute + renaming).

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class IrAccessToken(models.Model):
+    _name = 'ir.access.token'
+    _description = 'Access Token'
+
+    res_model = fields.Char('Resource Model', required=True)
+    res_id = fields.Integer('Resource ID', required=True)
+    scope = fields.Char('Scope', required=True)
+    expiration = fields.Datetime('Expiration')
```
mwath: Add an index here (needed for the GC).

odoo/addons/base/views/ir_access_token_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                    <field name="owner_id"/>
+                    <field name="expiration"/>
+                </list>
+            </field>
+        </record>
+
+        <record id="action_access_token" model="ir.actions.act_window">
+            <field name="name">Access Tokens</field>
+            <field name="res_model">ir.access.token</field>
+            <field name="path">access-token</field>
```
mwath: nitpicking, but I think it's better for the list view
```suggestion
            <field name="path">access-tokens</field>
```

odoo/orm/models.py
```diff
@@ -6320,6 +6327,142 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+            'res_id': self.id,
+            'scope': scope,
+            'expiration': expiration.replace(microsecond=0) if expiration else False,
+            'owner_id': owner.id if owner else False,
+            'manual_token': value,
+        })
+        return access_token.token
+
+    @api.private
+    def revoke_access_token(self, scope: str, *,
```
mwath:
```suggestion
    def revoke_access_tokens(self, scope: str, *,
```

odoo/orm/models.py
```diff
@@ -6320,6 +6327,142 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+
+        # Recompute token and assert with the token used ensures the integrity
+        # of values in the database. This ensures that if the data in the
+        # database is altered, it does not change the intended behaviour of the
+        # token.
+        if not consteq(access_token.token or '', token):
+            raise AccessError(_('Invalid token'))
+
+        # Return the referenced record in sudo
+        return self.browse(access_token.res_id).sudo()
```
mwath: What happens when the record is unlinked? Maybe we could revoke access tokens during unlinks, but this will slow down the unlinks. 😕

odoo/orm/models.py
```diff
@@ -6320,6 +6327,142 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+
+    @api.depends_context('uid')  # Invalidate cache if user changes
```
mwath: Does it work correctly even inside a lambda? 🤔 
I think the lambda should be decorated, not this method.
```python
access_token = field.Char(
    compute=lambda record: record._compute_access_token('<scope>')
)
```

mwath (changes requested):
Nice PR, I prefer this API over the one on `res.users`.
However, you do not handle tokens for removed records. 😕 

Otherwise, the review is mostly nitpicking

odoo/orm/models.py
```diff
@@ -6320,6 +6327,148 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+        except ValueError:
+            # The token is not in the correct format (maybe manually entered token)
+            access_token = self.env['ir.access.token'].sudo().search_fetch([
+                ('manual_token', '=', token),
+                ('scope', '=', scope),
+                ('owner_id', 'in', (False, self.env.uid)),
+                '|',
+                    ('expiration', '>=', datetime.now()),
+                    ('expiration', '=', False),
+            ], limit=1)
```
antonrom1: This is vulnerable to timing attacks, as we discussed, to support old tokens in unstructured formats, you have 2 options:
- Either you do like stripe (public_id + secret), where you have a btree index on public_id and you do a constant time comparison on secret
- Or you can index hmac(manual_token, db_secret), so that an attacker won't be able to get any sensible information from the timing information. So on every request, you would need to first compute the hmac(manual_token), and only then do the lookup

odoo/orm/models.py
```diff
@@ -6320,6 +6327,148 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+            if payload is None:  # Expired token
+                raise AccessError(_('Invalid token'))
+            access_token_id, res_model, *_rest, owner_id = payload
+            # Ensure correct model and owner
+            if (owner_id and owner_id != self.env.uid) or (res_model != self._name):
+                raise AccessError(_('Invalid token'))
+            access_token = self.env['ir.access.token'].sudo().browse(access_token_id)
+            # Ensure the access_token record exists (+ fetch)
+            try:
+                access_token.res_model
```
antonrom1: I think it might be a good idea to put the metadata into the token and sign it, but then res_model, (and other metadata) can become a computed field (probably stored computed field for legacy tokens that don't have metadata inside the token)

odoo/orm/models.py
```diff
@@ -6320,6 +6327,148 @@ def _get_placeholder_filename(self, field: str) -> bool:
         """
         return False
 
+    # ================
```
kmagusiak: @rco-odoo  I am wondering if we should  keep such extensions here or in a separate file in base with _name = 'base'.

odoo/orm/models.py
```diff
@@ -6320,6 +6327,148 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+            'scope': scope,
+            'expiration': expiration.replace(microsecond=0) if expiration else False,
+            'owner_id': owner.id if owner else False,
+            'manual_token': value,
+        })
+
+        # /!\ Invalidate the cache for the current record,
+        # it is necessary if access token fields exist.
+        # self.invalidate_recordset()
+        # TODO: find another solution because this is too costly
```
kmagusiak: Should we remove access token fields and use this API directly?

odoo/orm/models.py
kmagusiak: All metadata is already in the token and checked a few lines below when recomputing the token.

odoo/orm/models.py
kmagusiak: Same issue as for all many2one_reference. We have multiple options: delete records on unlink, check if they exist, do nothing (and let it crash later with a MissingException). Since there is no integrity guarantee, all of these are valid depending on the use case. We might implement an ondelete=cascade for these fields in the (near) future.

odoo/orm/models.py
```diff
@@ -6320,6 +6327,148 @@ def _get_placeholder_filename(self, field: str) -> bool:
...
+        # token.
+        if not consteq(access_token.token or '', token):
+            raise AccessError(_('Invalid token'))
+
+        # Return the referenced record in sudo
+        return self.browse(access_token.res_id).sudo()
+
+    @api.depends_context('uid')  # Invalidate cache if user changes
+    def _compute_field_access_token(self, scope: str, *,
+        required: bool = False, field_name: str = 'access_token') -> None:
```
kmagusiak: At the end, with `required` it seems bugprone because you need to precompute it :/

addons/survey/models/survey_survey.py
```diff
@@ -470,6 +461,9 @@ def _check_survey_responsible_access(self):
     @api.model_create_multi
     def create(self, vals_list):
         surveys = super().create(vals_list)
+        for survey, vals in zip(surveys, vals_list):
+            if 'access_token' in vals:
```
kmagusiak: I don't think we should accept values from the create function.

addons/survey/controllers/survey_session_manage.py
```diff
@@ -18,7 +19,11 @@ class UserInputSession(http.Controller):
     def _fetch_from_token(self, survey_token):
         """ Check that given survey_token matches a survey 'access_token'.
         Unlike the regular survey controller, user trying to access the survey must have full access rights! """
-        return request.env['survey.survey'].search([('access_token', '=', survey_token)])
+        Survey = request.env['survey.survey']
+        try:
+            return Survey.get_record_from_access_token('access', survey_token).sudo(False)
+        except AccessError:
+            return Survey.search([('access_token', '=', survey_token)])
```
kmagusiak: Why? If we did not find using the token, we should just raise.

odoo/orm/models.py
thle-odoo: My opinion is that the module that wants to create a compute field rather than using the API directly must take care to implement the compute.

odoo/orm/models.py
kmagusiak: Index using hash instead of a btree should work as well

addons/survey/models/survey_survey.py
```diff
@@ -96,7 +91,7 @@ def default_get(self, fields):
         ('public', 'Anyone with the link'),
         ('token', 'Invited people only')], string='Access Mode',
         default='public', required=True)
-    access_token = fields.Char('Access Token', default=lambda self: self._get_default_access_token(), copy=False)
+    access_token = fields.Char('Access Token', compute='_compute_access_token')
```
kmagusiak: As discussed, we should remove that field and use the functions in its place.
We should not be able to set values. Manual values are reserved for migrations (and maybe operations in sudo as a last-case resort).

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+    _res_record_idx = models.Index('(res_model, res_id)')
+
+    @api.autovacuum
+    def _gc_access_token(self):
+        self.search([('expiration', '<', self.env.cr.now())]).unlink()
+
+    @api.constrains('expiration')
+    def _constraint_expiration(self):
+        for access_token in self:
+            if access_token.expiration and access_token.expiration < fields.Datetime.now():
```
kmagusiak:
```suggestion
            if access_token.expiration and access_token.expiration < self.env.cr.now():
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+import logging
+import typing
+from datetime import datetime
+
+from odoo import _, api, fields, models
```
kmagusiak: Use `self.env._`.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+    @api.model
+    @api.private
+    def retrieve_from_token(self, scope: str, token: str) -> BaseModel:
+        """ Retrieve access token record from a token and scope. """
+        try:
+            payload = verify_hash_signed(self.env, scope, token)
+
+        except ValueError:
+            domain = [
+                ('scope', '=', scope),
```
kmagusiak: Missing condition on `manual_token`.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+                self.env, access_token.scope, payload,
+                expiration=access_token.expiration,
+            )
+
+    def write(self, vals):
+        raise UserError(_('Cannot update access tokens'))
+
+    @api.model
+    @api.private
+    def retrieve_from_token(self, scope: str, token: str) -> BaseModel:
```
kmagusiak:
```suggestion
    def retrieve_from_token(self, scope: str, token: str) -> Self:
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return access_token
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    @api.private
+    def grant_access_token(self, scope: str, *,
+        expiration: datetime | None = None, owner: BaseModel | None = None,
```
kmagusiak:
```suggestion
        expiration: datetime | timedelta | None = None, owner: BaseModel | None = None,
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+        value: str | None = None) -> str:
```
kmagusiak:
```suggestion
        _manual_token: str | None = None) -> str:
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+        owners: BaseModel | None = None) -> int:
+        """ Ensures that tokens for the records and scope are invalidated. """
+        if not self.env.is_admin():
+            raise AccessError(_("Only administrators can revoke access tokens."))
+
+        access_tokens = self.env['ir.access.token'].sudo().search([
+            ('res_model', '=', self._name),
+            ('res_id', 'in', self._ids),
+            ('scope', '=', scope),
+            ('owner_id', 'in', owners._ids if owners else [False]),
```
kmagusiak:
```suggestion
            ('owner_id', 'in', owners.ids if owners else [False]),
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+    def retrieve_from_token(self, scope: str, token: str) -> BaseModel:
+        """ Retrieve access token record from a token and scope. """
+        try:
+            payload = verify_hash_signed(self.env, scope, token)
+
+        except ValueError:
+            domain = [
+                ('scope', '=', scope),
+                ('owner_id', 'in', (False, self.env.uid)),
+                '|', ('expiration', '>=', datetime.now()), ('expiration', '=', False),
```
kmagusiak:
```suggestion
                '|', ('expiration', '>=', self.env.cr.now()), ('expiration', '=', False),
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,175 @@
...
+        """ Use the token and get the referenced record (sudoed).
+
+        :raises AccessError: If the token is invalid, expired, or does not match
+                             the expected scope or payload.
+                             If the owner of the token is not correct.
+        """
+        access_token = self.env['ir.access.token'].sudo().retrieve_from_token(scope, token)
+
+        if (access_token.res_model != self._name) \
+            or (access_token.owner_id and access_token.owner_id.id != self.env.uid):
```
kmagusiak: Owner should be checked inside `retrieve_from_token`.

odoo/addons/base/models/ir_access_token.py
thle-odoo: This is intentional because, in all cases, once the records have been retrieved with the domain, we filter the access tokens retrieved with `consteq`.

This allows tokens to be compared using `consteq` (which is good practice).


Of course, the scope must be precise enough for manual tokens (this will be the case since these tokens are normally only used for backward compatibility reasons).

kmagusiak: Here your domain is large enough to retrieve all tokens for a given scope (could be 10% of the whole table). You must add manual_token with an index to avoid crashing (out of memory). If you use the hash index on postgresql side, it is constant in time.

odoo/orm/models.py
antonrom1: btree on sha256(token) is likely fine in practice because the indexed value is random-looking, so timing can't reveal anything useful unless the attacker already has a small candidate set (maybe on the legacy low-entropy tokens?). In that case, if timing were to leak the prefix of the digest of the token, they could hash their low-entropy candidate set and filter those matching the prefix offline. btree on `HMAC(token, k)` removes even that: without k they can't test candidates against whatever leaked, so the side-channel becomes completely useless even for low entropy tokens

odoo/orm/models.py
thle-odoo: As discussed, it is the token that will not be stored and the information maintained (this allows queries to be performed on `ir.access.token`, for vacuuming or other purposes).

As mentioned, the integrity of this data is verified when the token is recalculated.

addons/survey/models/survey_survey.py
thle-odoo: The adaptation of the `survey` module is for informational purposes only. It will take a longer time to adapt it completely as many modules depend on it (adapt the logic/demo data of the various modules, etc.). Leaving the compute field "for testing" (the commit will not be merged) allows to focus on the proposed API.

I agree that in future, the `access_token` field should be removed.

addons/survey/models/survey_survey.py
thle-odoo: This is to satisfy unit tests and should not exist in the complete adaptation of the `survey` module (and its dependencies).

This is related to the comment: https://github.com/odoo/odoo/pull/244533#discussion_r2726493975

odoo/orm/models.py
thle-odoo: The two solutions:
- `plain token + hash index`
- `hmac(token, secret) + btree index`

In terms of timing attacks, even if the latter is theoretical, I think it needs to be taken into account. Measurements show that the variance in our HTTP flow is already too great to extract information from it. Both methods meet this requirement.

However, the main argument that would make me use `hmac(token, secret) + btree index` rather than `plain token + hash index` is the fact that the (legacy) token is not stored in plain text in the database.

Even though the use of `sha256` allows this, as mentioned, a legacy token with low entropy may be exposed (if we take timing attacks into account).

Unfortunately, storing the legacy token in encrypted form would block the use of the `get_access_token` method (because `_compute_token` would not be able to find the original token).

These legacy tokens are already stored in the database. Moving them does not increase the risk that already exists.

Currently, access tokens (legacy and no stored) must be retrievable __in order to meet business needs__.

thle-odoo: We will use an optional parameter `record_id` that we will use in the domain. This eliminates problems (and complexity) related to timing attacks.

This is what the tokens currently present should already be doing (so adjustments will need to be made if this is not the case).

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,202 @@
...
+    # However, it is necessary to maintain existing tokens as valid
+    # and retain the ability to force the value of a token.
+    manual_token = fields.Char('Manual Token', groups=fields.NO_ACCESS)
+    token = fields.Char('Token', compute='_compute_token', groups=fields.NO_ACCESS)
+
+    _res_record_idx = models.Index('(res_model, res_id)')
+
+    @api.autovacuum
+    def _gc_access_token(self):
+        self.search([('expiration', '<', self.env.cr.now())]).unlink()
```
kmagusiak: Add: `from odoo.tools.constants import GC_UNLINK_LIMIT`

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,202 @@
...
+
+        :param scope: Access scope associated with the token.
+        :param expiration: Optional expiration datetime for the token.
+                           Optional duration timedelta for the token.
+                           If False, the token does not expire.
+        :param owner: Optional single record identifying the token owner.
+        :param _manual_token: Optional token value which must be forced.
+        :return: Signed access token string.
+        """
+        if not self.env.is_admin():
```
kmagusiak: Shouldn't this be `is_system`? or even better, just test `self.env.su` because I don't remember a use case for users calling this directly. (unless you have a use case)

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,202 @@
...
+                return self.browse()
+
+            domain = [('id', '=', access_token_id)]
+
+        access_tokens = self.search_fetch(domain)
+
+        return next((
+            access_token for access_token in access_tokens
+            if consteq(access_token.token, token)
+        ), self.browse())
```
kmagusiak: Nitpick
```suggestion
        for access_token in self.search_fetch(domain):
          if consteq(access_token.token, token):
            return access_token
        return self.browse()
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,202 @@
...
+            'res_model': self.res_model,
+            'res_id': self.res_id,
+            'view_mode': 'form',
+            'target': 'current',
+        }
+
+    @api.model
+    @api.private
+    def retrieve_from_token(self, scope: str, token: str, *,
+        record_id: int | None = None) -> Self:
```
kmagusiak: Pass a real record instead for legacy tokens to use `record._name` and hit the index. Otherwise, it would be a full table scan.
```suggestion
        record: BaseModel | None = None) -> Self:
```

-----

The other possibility is to searching using the token directly and having a hash index in the database for it - which is much simpler.

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,245 @@
...
+        cls.user_2 = new_test_user(cls.env, 'Test User 2')
+
+        cls.model = cls.env['res.partner'].sudo(False)
+        cls.record = cls.model.create({'name': 'Test Record'}).sudo(False)
+
+        cls.classPatch(BaseCursor, 'now', lambda cr: datetime.now())
+
+    def setUp(self):
+        super().setUp()
+        self.IrAccessToken.search([]).unlink()
```
kmagusiak: You can move this inside setup class because each test is ran in a savepoint and rolled back.

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,245 @@
...
+
+        with freeze_time('2026-01-11'), self.assertRaises(AccessError):  # The token is invalid
+            self.model.with_user(self.user_1).get_record_from_access_token('access-record', token)
+
+    @freeze_time('2026-01-01')
+    def test_grant_access_to_record_with_duration(self):
+        duration = timedelta(days=10)
+        token = self.record.with_user(self.admin).grant_access_token('access-record', expiration=duration)
+
+        with freeze_time('2026-01-9'):
```
kmagusiak: (same in other places)
```suggestion
        with freeze_time('2026-01-09'):
```

kmagusiak (commented):
Nice work.

kmagusiak (approved):

mwath (approved):

rco-odoo (approved):

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+class IrAccessToken(models.Model):
+    _name = 'ir.access.token'
+    _description = 'Access Token'
+    _auto = False
+
+    res_model = fields.Char('Resource Model', required=True, readonly=True)
+    res_id = fields.Many2oneReference('Resource ID', model_field='res_model', required=True, readonly=True)
+    scope = fields.Char('Scope', required=True, readonly=True)
+    expiration = fields.Datetime('Expiration', readonly=True, index='btree')
+    owner_id = fields.Many2one('res.users', readonly=True, ondelete='cascade')
```
antonrom1, line 27: Wouldn't it make sense that the owner_id would be a `res.partner` for some business flows?

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+            ))
+
+        query = SQL("""
+            SELECT id, res_model, res_id, scope, expiration, owner_id, manual_token
+            FROM %(table)s
+            WHERE
+                res_model = %(res_model)s
+                AND res_id = %(res_id)s
+                AND scope = %(scope)s
+                AND (owner_id IS NULL OR owner_id = %(uid)s)
```
antonrom1:
```suggestion
                AND (owner_id IS NULL OR owner_id = %(owner)s)
```

edit: (oops, not the right %(uid), i meant __get_token)

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+                AND (expiration IS NULL OR expiration >= %(now)s)
+            ORDER BY expiration DESC NULLS FIRST
```
antonrom1: I think `owner_id` should also deterministic and NULLS FIRST. Also add id to make it deterministic no?.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+            res_model=record._name if record else None,
+            res_id=record.id if record else None,
+        )
+
+    def _retrieve_access_token_from_record(self, scope: str, *, owner: BaseModel | None = None) -> str | None:
+        """ Retrieve the token value of the record. """
+        self.ensure_one()
+        return self.env['ir.access.token']._IrAccessToken__retrieve_token(
+            self._name, self.id, scope, owner.id if owner else None,
+        )
```
antonrom1: I am not a fan of this because:
- it incentivizes to not create new tokens whenever you need
- It doesn't allow controlling the expiration date (you will probably get a token that has no expiration date as we order by NULLS FIRST)
- Disabling a token now has more side-effects than the expected flow of having one business flow per token
- Its not logged

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def _revoke_access_tokens(self, scope: str, *, owners: BaseModel | None = None) -> int:
```
antonrom1:
```suggestion
    @api.model
    def _revoke_access_tokens(self, scope: str, *, owners: BaseModel | None = None) -> int:
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+        """ Ensures that tokens for the records and scope are invalidated. """
+        if not self.env.su:
+            raise AccessError(self.env._("Sudo environment is required to revoke access tokens."))
+
+        access_tokens = self.env['ir.access.token'].search([
+            ('res_model', '=', self._name),
+            ('res_id', 'in', self._ids),
+            ('scope', '=', scope),
+            ('owner_id', 'in', owners.ids if owners else [False]),
```
antonrom1: It's not intuitive that calling `_revoke_access_tokens('scope')` doesn't delete owned tokens

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    def _grant_access_token(
+        self,
+        scope: str,
+        *,
+        expiration: datetime | timedelta | None = None,
```
antonrom1: I think it's better to choose either datetime or timedelta, but not both

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+    _description = 'Access Token'
+    _auto = False
+
+    res_model = fields.Char('Resource Model', required=True, readonly=True)
+    res_id = fields.Many2oneReference('Resource ID', model_field='res_model', required=True, readonly=True)
+    scope = fields.Char('Scope', required=True, readonly=True)
+    expiration = fields.Datetime('Expiration', readonly=True, index='btree')
+    owner_id = fields.Many2one('res.users', readonly=True, ondelete='cascade')
+
+    _res_record_idx = models.Index('(res_model, res_id)')
```
antonrom1: Do it in in `init`, your model is auto=False

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+        """ Generate and get access token for the record. """
+        self.ensure_one()
+        return self.env['ir.access.token']._IrAccessToken__generate(
+            self._name, self.id, scope,
+            expiration, owner.id if owner else None,
+            _manual_token,
+        )
+
+    @api.model
+    def _retrieve_record_from_access_token(
```
antonrom1: Isn't there a risk that it's not clear enough this returns sudoed recordset and someone will do this? (maybe it's ok because you need to be sudo anyway to call it)

```py
record = self.sudo()._retrieve_record_from_access_token(scope, token)
record.write(vals)
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,249 @@
...
+
+        with self.assertRaises(AccessError):  # With user 2, the token is invalid
+            self.model.with_user(self.user_2).sudo()._retrieve_record_from_access_token('access-record', token)
+
+        record = self.model.with_user(self.user_1).sudo()._retrieve_record_from_access_token('access-record', token)
+        self.assertEqual(record, self.record, 'Token is valid for the owner user_1')
+
+    def test_grant_access_to_record_with_scope(self):
+        token_read = self.record._grant_access_token('access-read')
+        token_write = self.record._grant_access_token('access-write')
```
antonrom1: These tests are implicitly recommending using `'access-read', 'access-write'` scopes. This is a bad idea, and I think you should be more opinionated in namespacing these to avoid collisions in business logic. 
`'l10n_de.allow_kyc_company_write'` for example

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,294 @@
...
+
+        if res_model != model:
+            raise AccessError(self.env._("Invalid token: incorrect model."))
+
+        self.env.cr.execute(query)
+
+        for access_token_id, res_model, res_id, scope, expiration, owner_id, manual_token in self.env.cr.fetchall():
+            token_from_db = manual_token or hash_sign(self.env, scope, (access_token_id, res_model, res_id, owner_id), expiration)
+            if consteq(token, token_from_db):
+                return self.env[model].browse(res_id)
```
antonrom1: Shouldn't you do an existance check before returning at least?

odoo/addons/base/models/ir_access_token.py
thle-odoo: I think the best approach is to work exclusively with the `timedelta` type.
This has the advantage of avoiding the complexity associated with timezones.
Furthermore, a token’s validity is generally limited to a specific duration.

odoo/addons/base/models/ir_access_token.py
thle-odoo: In practice, we use an access token precisely because we don't have access to the record and expect to receive it sudoed.
However, I agree that it would be good to specify this explicitly (as with `__retrieve_sudo_record`).
But isn't the name `_retrieve_sudo_record_from_access_token` a bit too long ?

odoo/addons/base/models/ir_access_token.py
thle-odoo: The purpose of using an access token is to retrieve a record. I don't think it is the "access token's job" to check whether the record exists (and adding an additional query).
See the comment: https://github.com/odoo/odoo/pull/244533#discussion_r2715898233

odoo/addons/base/models/ir_access_token.py
thle-odoo: I'm going to update the `owners` parameter to achieve the following behavior:
- if `res.users` records, delete access token only for these users,
- if `False`, delete shared access token,
- if `None`, delete all access token associated to the scope.
What do you think ?

odoo/addons/base/models/ir_access_token.py
thle-odoo: I understand the idea of using `res.partner` instead of `res.users`.
However, in practice, this won't work.

When a public user uses an access token, it will be done via a route that doesn't require authentication. Consequently, no verification against a user is possible. Having a `partner_id` in the token doesn't provide any benefit other than traceability.

The purpose of including the `owner_id` (`user_id`) in the token is for a business flox that requires authentication (for example, to elevate privileges on a record). This ensures that other users cannot use the token.

odoo/addons/base/models/ir_access_token.py
thle-odoo: I have decided to remove this method entirely. We can always add it in the future if we need it
I agree that, in general, if we need a token, we should generate one.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+                FROM %(table)s
+                WHERE
+                    res_model = %(res_model)s
+                    AND res_id = %(res_id)s
+                    AND scope = %(scope)s
+                    AND (owner_id IS NULL OR owner_id = %(uid)s)
+                    AND (expiration IS NULL OR expiration >= %(now)s)
+                ORDER BY
+                    owner_id NULLS FIRST,
+                    expiration DESC NULLS FIRST
```
antonrom1: add id at the end so that it's deterministic ordering

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+        )
+
+    @api.model
+    def _revoke_access_tokens(self, scope: str, *, owners: BaseModel | Literal[False] | None = None) -> int:
+        """ Ensures that tokens for the records and scope are invalidated.
+
+        :param scope: access scope associated with the token.
+        :param owners: - if `res.users` records, delete access token only for these users,
+                       - if `False`, delete shared access token,
+                       - if `None`, delete all access token associated to the scope.
```
antonrom1: this is confusing imo.

These do totally different things?
_revoke_access_tokens(..., owners=False)
_revoke_access_tokens(..., owners=None)

Also I think this should be banned
_revoke_access_tokens(..., owners=env['res.partner'].browse(1))

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+        token_read = self.record._grant_access_token('<module>.<action_A>')
+        token_write = self.record._grant_access_token('<module>.<action_B>')
+
+        with self.assertRaises(AccessError):  # Token incorrect for the scope
+            self.model._get_sudo_record_from_access_token('<module>.<action_B>', token_read)
+        with self.assertRaises(AccessError):  # Token incorrect for the scope
+            self.model._get_sudo_record_from_access_token('<module>.<action_A>', token_write)
+
+        record = self.model._get_sudo_record_from_access_token('<module>.<action_B>', token_write)
+        self.assertTrue(record, self.record)
```
antonrom1:
```suggestion
        self.assertEqual(record, self.record)
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+from datetime import datetime, timedelta
+from freezegun import freeze_time
```
antonrom1:
```suggestion
from datetime import datetime, timedelta

from freezegun import freeze_time
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+        """,
+            table=SQL.identifier(self._table),
+            create_uid=self.env.uid,
+            res_model=res_model, res_id=res_id, scope=scope,
+            expiration=expiration, owner_id=owner_id, manual_token=manual_token,
+        ))
+        access_token_id = self.env.cr.fetchone()[0]
+
+        _logger.info(
+            'User %d generate access tokens (%s): %s, %s, %s, %s, %s',
```
antonrom1:
```suggestion
            "User %d generate access tokens (%s): %s, %s, %s, %s, %s",
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+            'res_model': self.res_model,
+            'res_id': self.res_id,
+            'view_mode': 'form',
+            'target': 'current',
+        }
+
+    @api.model
+    def __generate(
+        self,
+        res_model: str, res_id: int, scope: str,
```
antonrom1:
```suggestion
        res_model: str, 
        res_id: int,
        scope: str,
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+        return manual_token or hash_sign(self.env, scope, (access_token_id, res_model, res_id, owner_id), expiration)
+
+    @api.model
+    def __retrieve_sudo_record(
+        self,
+        model: str, scope: str, token: str, *,
+        res_model: str | None = None,
+        res_id: int | None = None,
+    ) -> BaseModel:
+        """ Get the referenced record (sudo-ed) referenced by the token.
```
antonrom1:
```suggestion
        """ Get the referenced sudoed record by the token.
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+            raise AccessError(self.env._(
+                "Sudo environment is required to retrieve a record from access token"
+            ))
+
+        try:
+            payload = verify_hash_signed(self.env, scope, token)
+        except ValueError:
+            # Manual token
+            if not (res_model and res_id):
+                raise AccessError(self.env._("Manual token require record identification."))
```
antonrom1:
```suggestion
                raise AccessError(self.env._("Manual tokens require record identification."))
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+    def test_ensure_database_integrity_owner(self):
+        token = self.record._grant_access_token('<module>.<action>', owner=self.user_1)
+        record = self.model.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('<module>.<action>', token)
+        self.assertEqual(record, self.record, 'Token must be valid')
+        # Alter DB - Attempting to change user
+        access_token = self.IrAccessToken.search([], limit=1)
+        self.env.registry['base'].write(access_token, {'owner_id': self.user_2})  # Bypass write override
+        with self.assertRaises(AccessError):  # We detect a modification in the user
+            self.model.with_user(self.user_1).sudo()._get_sudo_record_from_access_token('<module>.<action>', token)
+        with self.assertRaises(AccessError):  # Token completly invalid (even if we use the updated user)
```
antonrom1:
```suggestion
        with self.assertRaises(AccessError):  # Token completely invalid (even if we use the updated user)
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+    def test_ensure_database_integrity_scope(self):
+        token = self.record._grant_access_token('<module>.<action_A>')
+        record = self.model._get_sudo_record_from_access_token('<module>.<action_A>', token)
+        self.assertEqual(record, self.record, 'Token must be valid')
+        # Alter DB - Attempting to change scope
+        access_token = self.IrAccessToken.search([], limit=1)
+        self.env.registry['base'].write(access_token, {'scope': '<module>.<action_B>'})  # Bypass write override
+        with self.assertRaises(AccessError):  # We detect a modification in the scope
+            self.model._get_sudo_record_from_access_token('<module>.<action_A>', token)
+        with self.assertRaises(AccessError):  # Token completly invalid (even if we use the updated scope)
```
antonrom1:
```suggestion
        with self.assertRaises(AccessError):  # Token completely invalid (even if we use the updated scope)
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+
+        with freeze_time('2026-01-09'):
+            record = self.model._get_sudo_record_from_access_token('<module>.<action>', token)
+            self.assertEqual(record, self.record, 'Token must be valid')
+
+        with freeze_time('2026-01-11'), self.assertRaises(AccessError):  # The token is invalid
+            self.model._get_sudo_record_from_access_token('<module>.<action>', token)
+
+    @freeze_time('2026-01-01')
+    def test_grant_access_to_record_with_duration(self):
```
antonrom1, line 55: This test is duplicate of above

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+
+        cls.admin = new_test_user(cls.env, 'Test Admin', groups='base.group_erp_manager')
+        cls.user_1 = new_test_user(cls.env, 'Test User 1')
+        cls.user_2 = new_test_user(cls.env, 'Test User 2')
+
+        cls.model_no_su = cls.env['res.partner'].with_user(cls.admin)
+        cls.record_no_su = cls.model_no_su.create({'name': 'Test Record'})
+
+        cls.model = cls.model_no_su.sudo()
+        cls.record = cls.record_no_su.sudo()
```
antonrom1:
```suggestion
        cls.Partner = cls.env['res.partner'].with_user(cls.admin)
        cls.partner = cls.model_no_su.create({'name': 'Test Record'})

        cls.PartnerSudo = cls.model_no_su.sudo()
        cls.partner_sudo = cls.record_no_su.sudo()
```

odoo/addons/base/tests/test_ir_access_token.py
```diff
@@ -0,0 +1,212 @@
...
+
+        cls.classPatch(BaseCursor, 'now', lambda cr: datetime.now())
+
+        cls.IrAccessToken.search([]).unlink()
+
+    def test_grant_access_to_record(self):
+        with self.assertRaises(AccessError):  # Sudo environment required to grant access tokens
+            self.record_no_su._grant_access_token('<module>.<action>')
```
antonrom1: I prefer these not to be placeholders. Keep the placeholders in the docstrings i would say. For example

```suggestion
            self.record_no_su._grant_access_token('test.partner_phone_number_read')
```

odoo/addons/base/views/ir_access_token_views.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+
+        <record model="ir.ui.view" id="ir_access_token_view_tree">
+            <field name="name">ir.access.token.list</field>
+            <field name="model">ir.access.token</field>
+            <field name="arch" type="xml">
+                <list>
+                    <field name="res_model" string="Model"/>
```
antonrom1: Doesn't it work without overriding the string?

odoo/addons/base/views/ir_access_token_views.xml
```diff
@@ -0,0 +1,34 @@
...
+                    <field name="res_id" string="Name"/>
```
antonrom1: Why `Name`?

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        if not self.env.su:
+            raise AccessError(self.env._(
+                "Sudo environment is required to generate access token."
+            ))
+
+        expiration = None
+        if duration:
+            if duration.total_seconds() < 0:
+                raise ValidationError(self.env._('Duration cannot be negative.'))
```
antonrom1:
```suggestion
        if duration is not None:
            if duration.total_seconds() <= 0:
                raise ValidationError(self.env._("Duration must be positive"))
```

odoo/addons/base/views/ir_access_token_views.xml
thle-odoo: I rename it directly on the field

odoo/addons/base/views/ir_access_token_views.xml
thle-odoo: I rename it directly on the field

odoo/addons/base/models/ir_access_token.py
thle-odoo: As discussed, we use the boolean `revoke_shared` to delete shared access tokens (rather than using `False`).

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,271 @@
...
+        :param scope: access scope associated with the token.
+        :param duration: optional expiration timedelta for the token,
+                         if `None`, the token does not expire.
+        :param owner_id: optional id identifying the token owner.
+        :param manual_token: optional token value which must be forced.
+        :return: signed access token string.
+        """
+        if not self.env.su:
+            raise AccessError(self.env._(
+                "Sudo environment is required to generate access token."
```
odony: 🧦 normally we don't translate messages aimed at devs, it pollutes translations and our translators don't know what to do with those. Here for example users (and translators) wouldn't know what a "Sudo environment" is 😉

I guess the same is true for all the AccessError/ValueError/ValidationError messages here. I don't think users are expected to mananage access tokens manually, so only devs would ever see this in case of bugs, no? 🤔

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,271 @@
...
+
+        expiration = None
+        if duration is not None:
+            if duration.total_seconds() <= 0:
+                raise ValidationError(self.env._("Duration must be positive."))
+            expiration = self.env.cr.now() + duration
+
+        self.env.cr.execute(SQL("""
+            INSERT INTO %(table)s (create_uid, res_model, res_id, scope, expiration, owner_id, manual_token)
+            VALUES (%(create_uid)s, %(res_model)s, %(res_id)s, %(scope)s, %(expiration)s, %(owner_id)s, %(manual_token)s)
```
odony: note to self: `create_date` is handled in the SQL column's DEFAULT

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,271 @@
...
+            table=SQL.identifier(self._table),
+            create_uid=self.env.uid,
+            res_model=res_model, res_id=res_id, scope=scope,
+            expiration=expiration, owner_id=owner_id, manual_token=manual_token,
+        ))
+        access_token_id = self.env.cr.fetchone()[0]
+
+        _logger.info(
+            "User %d generate access tokens (%s): %s, %s, %s, %s, %s",
+            self.env.uid, access_token_id, res_model, res_id, scope, expiration, owner_id,
```
odony: differentiate manual tokens for greppability?

```suggestion
            "User %d generated access token (#%s): %s, %s, %s, %s, %s %s",
            self.env.uid, access_token_id, res_model, res_id, scope, expiration, owner_id,
            ' (manual)' if manual_token else ''
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,284 @@
...
+                table=SQL.identifier(self._table),
+                create_uid=self.env.uid,
+                res_model=res_model, res_id=res_id, scope=scope,
+                expiration=expiration, owner_id=owner_id, manual_token=manual_token,
+            ))
+            access_token_id = cr.fetchone()[0]
+            _logger.info(
+                "User %d generates access tokens (%s): %s, %s, %s, %s, %s",
+                self.env.uid, access_token_id, res_model, res_id, scope, expiration, owner_id,
+            )
```
odony: might be nice to also indicate the use of a manual token in the log message?

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,284 @@
...
+
+    create_uid = fields.Many2one('res.users', string='Created by', readonly=True)
+    create_date = fields.Datetime('Created on', required=True, readonly=True)
+    res_model = fields.Char('Model', required=True, readonly=True)
+    res_id = fields.Many2oneReference('Record', model_field='res_model', required=True, readonly=True)
+    scope = fields.Char('Scope', required=True, readonly=True)
+    expiration = fields.Datetime('Expiration', readonly=True, index='btree')
+    owner_id = fields.Many2one('res.users', readonly=True, ondelete='cascade')
+
+    def init(self):
```
kmagusiak: _auto_init override: create_column if it does not exist

kmagusiak:
@robodoo r+

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,270 @@
...
+                res_model=res_model, res_id=res_id, scope=scope,
+                expiration=expiration, owner_id=owner_id, manual_token=manual_token,
+            ))
+            access_token_id = cr.fetchone()[0]
+            _logger.info(
+                "User %d generates access tokens (%s): %s, %s, %s, %s, %s",
+                self.env.uid, access_token_id, res_model, res_id, scope, expiration, owner_id,
+            )
+
+        return manual_token or hash_sign(self.env, scope, (access_token_id, res_model, res_id, owner_id, extra), expiration)
```
odony: could we explicitly have a comment that we want `access_token_id` to be part of the payload that is signed, because that is the only way we can make sure that the token will be different if it needs to be regenerated. It's a key part.

odony: I'm not sure I'm comfortable with the idea of returning the payload in clear. I can see cases where that `extra` value or the `owner_id` might hold sensitive information, and devs may not realize that they're going to be leaked in plaintext inside the token. The token will look "opaque" to the unaware.

It also means that this information will *remain available forever* in the URL, even if the token gets revoked. So the leak if effectively permanent.

An alternative is to store the `extra`, and load the token data before it can be validated. Would that model work for the use cases foreseen?

odony:
@robodoo r-

odoo/addons/base/models/ir_access_token.py
kmagusiak: `AccessError` is a `UserError`, that's why a translation is required.
We can just replace it with a `RuntimeError` for devs. Alternatively, we could provide a better user message, but I think this can stay technical.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,270 @@
...
+                "A manual token cannot contain additional information."
+            ))
+
+        expiration = None
+        if duration is not None:
+            if duration.total_seconds() <= 0:
+                raise ValidationError(self.env._("Duration must be positive."))
+            expiration = self.env.cr.now() + duration
+
+        with self.env.registry.cursor() as cr:
```
kmagusiak: Why do you start a new transaction?
Also, use the ORM and just update `manual_token` in SQL.

odoo/addons/base/models/ir_access_token.py
kmagusiak: Indeed, we could store `extra` as a column (json) and sign the hash of that column.

odoo/addons/base/models/ir_access_token.py
thle-odoo: I received feedback describing a practice use case where tokens are generated to grant access to an external API. This API will immediately send a GET request to retrieve the resource. The API will send this request before the transaction that creates the token is complete, so without a commit/flush, the token will not be in the database (and the API will not have access to the resource).

I plan to continue creating the token using a query, since the `manual_token` and `extra` columns will not be accessible via the ORM. Therefore, creating the token with a query will save one query (the update).

kmagusiak: Don't break transactionality by default.
If a developer need to have their token available before the end of the transaction, they can create a new cursor that creates that token themselves. When you write "without commit/flush", the flush part does not make anything visible (it might just make a broken test work).

Use the ORM to create tokens. Don't try to save 1 query here, and most of the time manual token and extra will be empty anyways so you don't have to update anything.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,262 @@
...
+            payload = verify_hash_signed(self.env, scope, token)
+        except ValueError:
+            assert res_model and res_id, "Manual tokens require record identification."
+
+            query = self._search([
+                ('res_model', '=', res_model),
+                ('res_id', '=', res_id),
+                ('scope', '=', scope),
+                '|', ('owner_id', '=', False), ('owner_id', '=', self.env.uid),
+                '|', ('expiration', '=', False), ('expiration', '>=', self.env.cr.now()),
```
kmagusiak: Nitpick:
```suggestion
                '|', ('expiration', '=', False), ('expiration', '>', self.env.cr.now()),
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,262 @@
...
+
+        query = query.select(
+            query.table.id,
+            query.table.res_model,
+            query.table.res_id,
+            query.table.scope,
+            query.table.expiration,
+            query.table.owner_id,
+            query.table.json_extra,
+            SQL("%s", SQL.identifier(self._table, 'manual_token')),
```
kmagusiak:
```suggestion
            SQL("%s", SQL.identifier(query.table, 'manual_token')),
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,262 @@
...
+        )
+
+        for access_token_id, res_model, res_id, scope, expiration, owner_id, json_extra, manual_token in self.env.execute_query(query):
+            token_from_db = manual_token or hash_sign(
+                self.env, scope, (access_token_id, res_model, res_id, owner_id, hash(json_extra)), expiration
```
kmagusiak: You should use a hash function that is not implementation dependent.
If the python implementation even changes the `hash` function, this should not change.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+            if consteq(token, token_from_db):
+                return self.env[model].browse(res_id).with_context(access_token_extra=json.loads(json_extra))
```
kmagusiak: `json.loads` can fail il json_extra is empty.
```suggestion
                return self.env[model].browse(res_id).with_context(access_token_extra=json_extra and json.loads(json_extra))
```

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,262 @@
...
+        assert not (extra and manual_token), "A manual token cannot contain additional information."
+        assert self.env.su, "Sudo environment is required to generate access token."
+
+        expiration = None
+        if duration is not None:
+            if duration.total_seconds() <= 0:
+                raise ValidationError(self.env._("Duration must be positive."))
+            expiration = self.env.cr.now() + duration
+
+        json_extra = json.dumps(extra)
```
kmagusiak: Same remarks to what is in the token retrieval, below code for extra and then there is the hash function too,.
```suggestion
        json_extra = json.dumps(extra) if extra else False
```

odoo/addons/base/models/ir_access_token.py
thle-odoo: The current design ensures that `extra` is never empty but is instead `None`, and therefore `json_extra` is `'null'` (this is a choice). This allows to place a falsy value in the extras.

kmagusiak: But you need to store 'null' instead of just NULL (free) which is the most common case.
You also have no required constraint on the json_extra field, so I am expecting the code to work in False/None as value.

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,265 @@
...
+        assert not (extra and manual_token), "A manual token cannot contain additional information."
+        assert self.env.su, "Sudo environment is required to generate access token."
+
+        expiration = None
+        if duration is not None:
+            if duration.total_seconds() <= 0:
+                raise ValidationError(self.env._("Duration must be positive."))
+            expiration = self.env.cr.now() + duration
+
+        json_extra = extra and json.dumps(extra)
```
kmagusiak: Nitpick: so if extra == 0, then we write the int instead of None?
(the field is a char, so we should write a string or False/None only)
```suggestion
        json_extra = json.dumps(extra) if extra else False
```

thle-odoo: Only an extra with a value of `None` will have no value in the database.
To avoid any ambiguity, I added a unit test.

rco-odoo:
@robodoo override=ci/security

Legitimate use of `sha256` for hashing the extra data in tokens.

rco-odoo:
@robodoo r+

odoo/addons/base/models/ir_access_token.py
```diff
@@ -0,0 +1,265 @@
...
+            token_from_db = manual_token or hash_sign(
+                self.env, scope, (access_token_id, res_model, res_id, owner_id, json_extra and hashlib.sha256(json_extra.encode()).hexdigest()), expiration
+            )
+            if consteq(token, token_from_db):
+                return self.env[model].browse(res_id).with_context(access_token_extra=json_extra and json.loads(json_extra))
+
+        raise AccessError(self.env._("Invalid token: expired (revoked) or invalid."))
+
+
+class Base(models.AbstractModel):
```
seb-odoo, line 191: Could we move this to a dedicated file?

It's hard to see there is actually a change to all models and to find back these methods when the change is hidden inside the ir access token model file itself.

thle-odoo: I think the location of the `Base` mixin override is well chosen.
The entire access token mechanism (along with its API) is located in the `ir_access_token.py` file.
We generally don't create a `base.py` file or anything like that (to my knowledge).

If this is a problem, I will take this feedback into account and move these two classes into separate files.

seb-odoo: There are many overrides of base model in various modules and they are usually all in models.py

PR #244540, opened 2026-01-19
[FIX] hr: employee version timeline auto save

abdullahsherif:
As runbot faketimes test fail because the usage of dynamic dates

now the test uses freeze_time

Task#5717031

---


amah-odoo (approved):
Thanks for your work 💯 

@robodoo  r+

PR #244542, opened 2026-01-19
[FIX] point_of_sale: raise error when can not make order paid

pedrambiria:
After commit https://github.com/odoo/odoo/commit/dde26e5372b8328bbc9e050f0fdedadd1de649fd, when trying to validate a POS order that was not fully paid, the order state would remain in 'draft' state but the picking and other related operations would be done. This could lead to double entries and inconsistencies.

opw-5447749

---


stva-odoo (approved):
robodoo r+

PR #244543, opened 2026-01-19
[FIX] account: a non-expected error should not break the import.

gawa-odoo:
When you import a batch of invoice and one of them gets an unexpected Exception, 
the others are created but we stop the method. 
It's a problem with crons that don't expect to be interrupted in the middle. 
It creates duplicates as we fail on the same invoice each time.

Of course, we should avoid all Exceptions when we can, but we should not loop on the same error.

opw-5503069
part of task-5499871

---

### DIFF
--- a/addons/account/models/account_document_import_mixin.py
+++ b/addons/account/models/account_document_import_mixin.py
@@ -6,11 +6,10 @@ import itertools
 import logging
 from lxml import etree
 from markupsafe import Markup
-import psycopg2.errors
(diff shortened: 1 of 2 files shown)


clbr-odoo (approved):
LGTM, just curious about the version, I imagine this is broken since longer than that no ? 🤔 
@robodoo r+ delegate+

clbr-odoo:
Perfect, thanks for the double-check.

PR #244544, opened 2026-01-19
[FIX] website_sale: remove variant image tours

Oussema39:
With this commit, we fix tours:
- website_sale.remove_main_product_image_with_variant
- website_sale.add_and_remove_main_product_image_no_variant where we want to remove the product image.
This fix add a step to ensure the image is in DOM before clicking on the remove button.

error-runbot-id~237766


ltinel (approved):
LGTM, thanks :)
@robodoo delegate+

PR #244546, opened 2026-01-19
[FIX] portal: prevent mutation during testing

lipiraux:
Before this commit, the values passed in to the `res.partner.write()` method were mutated during the call by an override of `l10n_cl`. This in turn caused tests of the
`portal.tests.test_addresses.TestPortalAddresses` test class to fail as the `self.default_address_values` dictionary which was used to populate fields on partners was mutated during tests and ended in an unexpected state.

After this commit, a copy of `self.default_address_values` is passed instead of the original dictionary.

runbot-164163

### DIFF
--- a/addons/portal/tests/test_addresses.py
+++ b/addons/portal/tests/test_addresses.py
@@ -126,7 +126,7 @@ class TestPortalAddresses(BaseCommon, HttpCase):
 
         internal_partner = self.internal_user.partner_id
         # Fill the incomplete address
-        internal_partner.write(self.default_address_values)
+        internal_partner.write(self.default_address_values.copy())
 
         # Try to update the account name
         res = self._submit_address_values({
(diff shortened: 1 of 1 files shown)


aboo-odoo (commented):
Hello! I'm not sure your fix is working :| It still triggers the error when calling with the following test tag `:TestPortalAddresses`. Does it work for you ?

Image

aboo-odoo (commented):
You're right, I donno what I was doing wrong. And you are also right about the mutable dict. However, I don't think the issue comes from CL but rather from the test file itself. It appears that `test_internal_user_cannot_update_email` and `test_internal_user_cannot_update_name` pass the `default_address_values` dict as parameter to the write which modifies it. Since `default_address_values` is a class dict, it does not get recomputed and the noise remains in all subsequent test. It gets added `partner_latitude`, `partner_longitude` and `vat` with the value `False` which then gets transformed into a string `'False'` by the controller which in turn fucks up the subsequent conditional statements. Doing this fixes the issue: ```internal_partner.write(self.default_address_values.copy())```

aboo-odoo (approved):
Thank you! :D 

@robodoo r+

PR #244547, opened 2026-01-19
[FIX] mail: fix non-deterministic copy message test

tsm-odoo:
Before this commit, the "Copy Message Link" test was sometimes failing. It happens for two reasons:
- It presses "ctrl+v" with the intent of copying the message link in the composer but doesn't check if the composer is focused (it should not be as the test clicks away to copy the link).
- It then presses "enter" with the intent of posting the message, without checking if the composer is filled or even ready to send.

This test replaces the "press" steps:
- "ctrl+v" is replaced by `insertText`: we already check that the copy works and "press" is less reliable.
- We click on the send button when enabled.

runbot-237538

Current behavior before PR:

---

### DIFF
--- a/addons/mail/static/tests/message/message.test.js
+++ b/addons/mail/static/tests/message/message.test.js
@@ -2166,16 +2166,16 @@ test("Copy Message Link", async () => {
     ]);
     await start();
     await openDiscuss(channelId);
+    await contains(".o-mail-Message", { count: 2 });
     await click(".o-mail-Message:eq(0) [title='Expand']");
     await contains(".o-dropdown-item:contains('Copy Link'_", { count: 0 });
     await click(".o-mail-Message:eq(1) [title='Expand']");
     await click(".o-dropdown-item:contains('Copy Link')");
-    await expect.waitForSteps([url(`/mail/message/${messageId_2}`)]);
-    await press(["ctrl", "v"]);
-    await press("Enter");
-    await contains(
-        `.o-mail-Message a[href='${url(`/mail/message/${messageId_2}`)}']:text('channel1')`
-    );
+    const link = url(`/mail/message/${messageId_2}`);
+    await expect.waitForSteps([link]);
+    await insertText(".o-mail-Composer-input", link);
+    await click("button[title='Send']:enabled");
+    await contains(`.o-mail-Message a[href='${link}']:text('channel1')`);
 });
 
 test("deleted message should not have translate feature", async () => {


addons/mail/static/tests/message/message.test.js
```diff
@@ -2170,12 +2170,11 @@ test("Copy Message Link", async () => {
...
-    await expect.waitForSteps([url(`/mail/message/${messageId_2}`)]);
-    await press(["ctrl", "v"]);
-    await press("Enter");
-    await contains(
-        `.o-mail-Message a[href='${url(`/mail/message/${messageId_2}`)}']:text('channel1')`
-    );
+    const link = url(`/mail/message/${messageId_2}`);
+    await expect.waitForSteps([link]);
+    await insertText(".o-mail-Composer-input", link);
+    await click(".o-sendMessageActive:enabled");
```
alexkuhn: Nitpick: `.o-sendMessageActive` is a modifier class, this doesn't bear the semantics this is attached to the send button. The modifier is more than just `o-active` because this classname is attached to mail `Action` button and this conflicts with the concept of an "active" `Action` button.

I would put `button[title='Send']` so we know for sure this is the button "send" we refer to here...
I see the same mistake in `search_messages_panel.test.js`.

alexkuhn (approved):
LGTM.

A bit sad to see all the replace of `press(Enter)` by click on button... I get the proper awaiting is shown in button but in practice 99% of interactions in non-mobile is through the `press(Enter)` interaction and it will be barely (or none at all) tested, which feels weird for such a core and heavily used user interaction in Discuss.

Also the showing of button is a soft spec, in the sense that it's there but we aren't sure to keep it. Some other messaging apps don't have it in desktop, and chat window narrow width could incite us to remove the button in desktop to gain more real-estate for message composition. This becomes more costly to remove this button because the tests are made to absolutely need this button :(

@robodoo delegate+

addons/mail/static/tests/message/message.test.js
```diff
@@ -2170,12 +2170,11 @@ test("Copy Message Link", async () => {
     await contains(".o-dropdown-item:contains('Copy Link'_", { count: 0 });
     await click(".o-mail-Message:eq(1) [title='Expand']");
```
alexkuhn, line 2172: Oh yeah, should `.Message { count: 2 }` after `await openDiscuss()` for awaiting of messages fetched.
Otherwise `Message:eq(0)` could actually be the last message!

tsm-odoo:
@robodoo r+

PR #244548, opened 2026-01-19
[FIX] web: fix aggregation crash on float monetary fields

juliusc2066:
This commit resolves a traceback during list view aggregation when the view is grouped.

The crash occurred specifically when aggregating a `float` field that uses the `monetary` widget. Unlike true `monetary` fields, the server does not automatically send a `currency_id` aggregate for `float` fields. The aggregation logic previously failed when this data was missing; it now handles the case by ignoring it since it's not supported.

opw-5725692

### DIFF
--- a/addons/web/static/src/views/list/list_renderer.js
+++ b/addons/web/static/src/views/list/list_renderer.js
@@ -736,7 +736,7 @@ export class ListRenderer extends Component {
                     } else {
                         currencyId = values[0][currencyField] && values[0][currencyField].id;
                     }
-                    if (func) {
+                    if (func && type === "monetary") {
                         const currencies = this.getFieldCurrencies(fieldName);
                         // in case of multiple currencies, convert values into default currency using conversion rates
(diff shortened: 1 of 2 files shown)


addons/web/static/src/views/list/list_renderer.js
```diff
@@ -795,7 +795,7 @@ export class ListRenderer extends Component {
         }
         if (this.props.list.isGrouped && !this.props.list.selection.length) {
             return values.reduce((set, value) => {
-                value[currencyField].forEach((c) => {
+                value[currencyField]?.forEach((c) => {
```
aab-odoo: Monetary aggregate on float field is indeed not supported: https://github.com/odoo/odoo/blob/19.0/odoo/orm/models.py#L1974

However, instead of using `?`, I would restrict the calls to this function to monetary field, 'cause `?` makes it look like we don't know what we're doing.

aab-odoo (approved):
robodoo r+

PR #244550, opened 2026-01-19
[PERF] account_edi_ubl_cii: Batch queries for Peppol cron

auma-odoo:
### Description:

Retrieving Peppol documents with high line counts could cause the background cron to timeout and eventually be disabled after multiple failures. This bottleneck occurred because the system processed each invoice line through individual calls.

To resolve this, the code has been refactored to batch the creates and updates. This change significantly reduces I/O overhead and ensures that large invoices no longer block the cron from creating new documents.

### Benchmark:

| N° of lines | Before  | After  |
|-------------|---------|--------|
|        4633 | Timeout |  3 min |

### Reference:
opw-5462267

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_common.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_common.py
@@ -6,6 +6,7 @@ from odoo.tools.float_utils import float_round
 from odoo.tools.misc import clean_context, formatLang
 from odoo.tools.zeep import Client
 
+from collections import defaultdict
 from markupsafe import Markup
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -959,6 +963,78 @@ def _import_fill_invoice_form(self, invoice, tree, qty_factor):
...
+        for tree, invoice_line in zip(trees, invoice_lines):
+
+            default_code = self._find_value('./cac:Item/cac:SellersItemIdentification/cbc:ID', tree)
+            name = self._find_value('./cac:Item/cbc:Name', tree)
+            barcode = self._find_value("./cac:Item/cac:StandardItemIdentification/cbc:ID[@schemeID='0160']", tree)
+            company = invoice_line.company_id
+
+            # Product.
+            product_id = previously_retrieved_product.get((default_code, name, barcode, company))
+            if not (default_code, name, barcode, company) in previously_retrieved_product:
```
gawa-odoo:
```suggestion
            if not product_id:
```

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -721,6 +722,85 @@ def _import_retrieve_fixed_tax(self, invoice_line, fixed_tax_vals):
                     return tax
         return self.env['account.tax']
 
+    def _import_fill_invoice_line_taxes_batched(self, all_tax_nodes, invoice_lines, all_inv_line_vals, logs):
```
gawa-odoo, line 767: I would think you would do the same as for the product with the remembering to avoid doing a search for each tax node?

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -721,6 +722,85 @@ def _import_retrieve_fixed_tax(self, invoice_line, fixed_tax_vals):
...
+            self.env['account.move.line'].browse(ids).product_uom_id = product_uom_id or False
+
+        for price_unit, ids in update_price_unit.items():
+            self.env['account.move.line'].browse(ids).price_unit = price_unit
+
+        for discount, ids in update_discount.items():
+            self.env['account.move.line'].browse(ids).discount = discount
+
+        for tax_ids, ids in update_tax_ids.items():
+            self.env['account.move.line'].browse(ids).tax_ids = list(tax_ids) or False
```
gawa-odoo, line 847: Can't you do a 
```
for invoice_line in lines:
     line_update_values.append([Command.update(line.id, {specific_values_for_that_line})])
invoice.line_ids = line_update_values
```
Won't the orm batch everything the right way anyway?
(I'm not sure for that part at all. I know it batches the write in a smart way for the SQL, I don't know if it works for the update)

auma-odoo: Seems to be the case in 18.0 and above but not before: [PR](https://github.com/odoo/odoo/pull/156238)

addons/account_edi_ubl_cii/models/account_edi_common.py
auma-odoo: True, I don't know why I forgot about it. Did the changes 😁

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -763,6 +764,92 @@ def _import_retrieve_fixed_tax(self, invoice_line, fixed_tax_vals):
...
+                ]
+
+                tax = False
+                if hasattr(invoice_line, '_predict_specific_tax'):
+                    # company check is already done in the prediction query
+                    predicted_tax_id = invoice_line\
+                        ._predict_specific_tax('percent', amount, tax_type)
+                    tax = self.env['account.tax'].browse(predicted_tax_id)
+                if not tax:
+                    tax = taxes_map.get((tax_type, amount))
```
gawa-odoo: Why not do that before the predict?
I suppose the only thing that could happen is that the most likely one for this amount and tax_type will be even more likely, no?

auma-odoo: Because the `_predict_specific_tax` seems to behave differently than a simple search. So I am not 100% sure if the result would have been the same in case we want to predict the tax vs search by type and amount. But if you think it should bring the same result, I can also add the predicted tax to the map of taxes.

gawa-odoo: I think it should bring the same result.
We take the last 100 invoices that respect the criteria and take the most common one that match.
I don't see anything possible other than your previous result being even more "common"

gawa-odoo (commented):
@robodoo r+

PR #244553, opened 2026-01-19
[FIX] point_of_sale: ensure product is loaded before refunding orderline

pedrambiria:
Before this commit, it was possible the cached toRefundLines included orderlines whose products were not loaded in the POS. This would cause an error when refunding another order for the same partner.

This commit ensures that the product is loaded before refunding an orderline by using a new method in the POS store that loads the product if it is not already present in the database.

opw-5384010

---


stva-odoo (approved):
robodoo r+

PR #244554, opened 2026-01-19
[FIX] account_add_gln: Fix label of the GLN in contact view

clbr-odoo:
The field is placed weirdly in the view when opening the contact view of the delivery address.
There is no label, making it almost impossible to discover, and it is in the middle of everything.
Let's put it down before the notes field.

Before:
image

After:
image

task-none


smetl (approved):
@robodoo r+

PR #244565, opened 2026-01-19
[FIX] website: Align menu items based on chosen alignment menu

sben-odoo:
Steps to reproduce:
===================
1- Enter the website editor
2- Enable mobile view
3- Edit the alignment of the mobile menu to be center or right aligned

The group labels (e.g. "Shop", "Forum") stay left-aligned regardless
of the chosen alignment.

This can also be seen on desktop by switching to the sidebar header
template.

Cause:
======
The class .accordion-button uses `display:flex` and `text-align:left`
and that class is used for the menu groups labels.
this prevents the alignment from working.

Solution:
=========
When right-aligned (`text-end`), reverse the flex direction so the arrow
moves to the left and the text stays on the right.

When centered (`text-center`), let the text span fill the remaining
space and center its content via `text-align: center`,
keeping the arrow on its position.

The default left-aligned case is unchanged.

opw-5494765
---

### DIFF
--- a/addons/website/static/src/scss/website.scss
+++ b/addons/website/static/src/scss/website.scss
@@ -441,6 +441,22 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
             }
         }
     }
+
+    .top_menu {
+        &.text-end > .accordion .accordion-button {
+            flex-direction: row-reverse;
+
+            &::after {
+                margin-left: 0;
+                margin-right: auto;
+            }
+        }
+
+        &.text-center > .accordion .accordion-button > span {
+            text-align: center;
+            flex-grow: 1;
+        }
+    }
 }
 
 // Apply background color to the offcanvas menu


nle-odoo (approved):
the change make sense to me

but it would be better IMO to just handle our case, so for example have it in the:

```css
.o_navbar_mobile {
    /* ... */
}
```

part and target this selector inside .o_navbar_mobile:

```css
.top_menu > .accordion .accordion-button {
```

We could also have the CSS directly on the "@mixin accordion-style-vertical-menu {" but it's also used for the desktop "sidebar" menu. The option "Mobile Alignment" currently only work for mobile menu accordion + desktop menu accordion (the second menu in menu template selector in 18.2) who both have ..o_navbar_mobile.

In 19.0 with the "Desktop Menu / Mobile Menu" options there is alignment option for desktop / mobile so in the forward-port we will probably need to fix both.

addons/website/static/src/scss/website.scss
```diff
@@ -453,6 +453,15 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
         z-index: $o-mega-menu-btn-close-zindex;
     }
 
+    // Fix menu alignment: Bootstrap's .accordion-button uses display:flex and text-align:left
+    // which prevents alignment utilities from working. Override when alignment class is present.
```
bvr-odoo: comments is too long (max. 80 caract)

addons/website/static/src/scss/website.scss
```diff
@@ -453,6 +453,15 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
...
+    .top_menu.text-center, .top_menu.text-end {
+        > .accordion .accordion-button {
+            display: block;
+            text-align: inherit;
+        }
+    }
```
bvr-odoo: From a design perspective, it’s preferable to position the accordion arrows differently than how you did it.

When the text is right-aligned, put the arrow on the left so all the items stay neatly aligned on a single line on the right. See below how it should look.
Image

When it’s centered, don’t center the arrow as well; keep it on the right.
Image

bvr-odoo (changes requested):
Hi @sben-odoo 👋🏼  here is my review

Regarding the commit message, it isn’t clear because your steps don’t explain that submenus are required.

addons/website/static/src/scss/website.scss
bvr-odoo: One more thing: the animations were lost, and the arrow no longer rotates when you open an accordion when it’s aligned differently than left-aligned 😬

sben-odoo: Fixed

addons/website/static/src/scss/website.scss
bvr-odoo, line 1: commit message:

- Steps to reproduce are still not clear [see my previous review](https://github.com/odoo/odoo/pull/244565/#pullrequestreview-3814183798)

- > This can also be seen on desktop by switching to the sidebar header
template.

    It's not the "sidebar template". You probably meant "hamburger template".

addons/website/static/src/scss/website.scss
```diff
@@ -441,6 +441,26 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
...
+        text-align: end;
+
+        &::after {
+            margin-left: 0;
+            margin-right: auto;
+        }
+    }
+
+    .top_menu.text-center > .accordion .accordion-button {
+        text-align: center;
```
bvr-odoo: Apply `text-align: center;` only on the span, not on the button

addons/website/static/src/scss/website.scss
```diff
@@ -441,6 +441,26 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
...
+
+        > span {
+            flex: 1;
```
bvr-odoo: `flex-grow: 1;` is clearer

addons/website/static/src/scss/website.scss
```diff
@@ -441,6 +441,26 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
             }
         }
     }
+
+    // Fix menu alignment: Bootstrap's .accordion-button uses display:flex
+    // and text-align:left which prevents alignment utilities from working.
+    .top_menu.text-end > .accordion .accordion-button {
+        flex-direction: row-reverse;
+        text-align: end;
```
bvr-odoo: what is the purpose of this `text-align: end;` ?

addons/website/static/src/scss/website.scss
```diff
@@ -441,6 +441,26 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
             }
         }
     }
+
+    // Fix menu alignment: Bootstrap's .accordion-button uses display:flex
+    // and text-align:left which prevents alignment utilities from working.
```
bvr-odoo: This comment says that the following code was added to fix something. In general, we do not write this kind of comment. It is useful in the commit message, but not here. In the code, we usually explain what the code does instead.

addons/website/static/src/scss/website.scss
```diff
@@ -441,6 +441,26 @@ $-navbar-mobile-bg-color: o-color('menu-custom') or o-color('menu');
...
+    .top_menu.text-end > .accordion .accordion-button {
```
bvr-odoo: You can use a single `.top_menu` selector that covers both cases, `.text-end` and `.text-center`, instead of duplicating it.

bvr-odoo (changes requested):

addons/website/static/src/scss/website.scss
sben-odoo: removed

addons/website/static/src/scss/website.scss
sben-odoo: You're right. I added that comment because I noticed that most of the file already contains comments, so I thought it would be consistent to include one here as well. (I removed it )

PR #244571, opened 2026-01-19
[FIX] orm: method 'is_transient' is not public method.

florentx:
It was documented within Doc API, but it cannot be called through the Json2 API.

This change removes method from API documentation:  `/doc-bearer/<model>.json` and `/doc/<model>.json`, etc...

---


florentx:
@Julien00859 
Since it relates to api_doc, can you have a look?

PR #244577, opened 2026-01-19
[FIX] cli: restore obfuscate command

MerlinGuillaume:
It is not possible to use the obfuscate command

Steps to reproduce:
1. Initialize a database `test`
2. In a terminal, try to obfuscate the database with the command `python odoo/odoo-bin obfuscate --pwd=1234 -d test`
3. An error occurs

Issue:
`config['db_name']` returns a list

Solution:
Make sure we use obfuscate with a single database and get the first (and only) database in the list

opw-5480280

### DIFF
--- a/odoo/cli/obfuscate.py
+++ b/odoo/cli/obfuscate.py
@@ -158,7 +158,13 @@ class Obfuscate(Command):
             if opt.allfields and not opt.unobfuscate:
                 _logger.error("--allfields can only be used in unobfuscate mode")
                 sys.exit("ERROR: --allfields can only be used in unobfuscate mode")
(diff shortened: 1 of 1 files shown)


hubvd (approved):

MerlinGuillaume:
We also check that there is a database mentioned in the command like in neutralize
https://github.com/odoo/odoo/blob/0acb0927ca02e8d65b945b3fcafd103109c8c548/odoo/cli/neutralize.py#L26-L32

kmagusiak (approved):
@robodoo r+

xmo-odoo:
@robodoo override=ci/security

PR #244647, opened 2026-01-20
[FIX] l10n_vn_edi_viettel: taxAmount decimal

jand-odoo:
When sending an invoice to SInvoice (Viettel), the API raises an `INVALID_DECIMAL_POINT_TAX_MONEY` error if the tax amount contains excessive decimal places (e.g., 100.02999999999997).

This occurs because the tax amount is currently calculated using simple subtraction (`total - subtotal`). Due to standard floating-point precision issues, this can result in unrounded values that the API rejects.

The tax amount calculation is now updated to explicitly use the currency's rounding method, ensuring the value is accepted by SInvoice.

task-[5484845](https://www.odoo.com/odoo/project/967/tasks/5484845)

### DIFF
--- a/addons/l10n_vn_edi_viettel/models/account_move.py
+++ b/addons/l10n_vn_edi_viettel/models/account_move.py
@@ -706,7 +706,7 @@ class AccountMove(models.Model):
                 # Values are either: -2 (no tax), -1 (not declaring/paying taxes), 0,5,8,10 (the tax %)
                 # Most use cases will be -2 or a tax percentage, so we limit the support to these.
                 'taxPercentage': line.tax_ids and line.tax_ids[0].amount or -2,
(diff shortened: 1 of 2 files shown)


vin-odoo (commented):
Make sure to always follow the [commit message structure](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-structure)

More specifically, don't forget add the task number in the commit message.

Other than that, it looks good to me :)

jand-odoo:
Hi @vin-odoo , thank you for the feedback! I have changed the message and force pushed again. Let me know if you need anything else :)

vin-odoo (approved):

vin-odoo (commented):
robodoo r+

Thanks!

PR #244650, opened 2026-01-20
[FIX] point_of_sale: correct discount amount on session report

paso-odoo:
Steps:
----
- Create a fiscal position with 2 different taxes
- Add a line in POS
- Apply fiscal position and add line discount
- Finish the order cycle
- Download the session report

Issue:
----
- The discount amount was calculated incorrectly in the session report

Cause:
----
- The discount amount calculation used taxes before applying the fiscal position

Fix:
----
- Used `tax_ids_after_fiscal_position` for tax calculation while computing the discount amount

task-5421215

---


davidmonnom (approved):
Careful when forward porting, `tax_ids_after_fiscal_position` was removed.

PR #244652, opened 2026-01-20
[FIX] l10n_gcc_invoicing: reposition GCC format option in settings

hash-odoo:
This fix moves the GCC Format checkbox under the 'Customer Invoices' block so it remains accessible in branches where the Fiscal Localization section is hidden.

task-5723793


hamo-odoo (approved):
LGTM

jbw-odoo (approved):
Thanks !
robodoo r+

PR #244653, opened 2026-01-20
[FIX] mail: display Invitation to follow notification correctly in inbox

akth-odoo:
Steps to reproduce:

- Configure user A to receive inbox notifications.
- As user B, invite user A to follow a record with Notify recipients enabled.
- Open the inbox of user A.

The Invitation to follow notification is not displayed in the inbox when no additional comment is provided. This happens because the notification body is empty unless extra comments are added.

This commit fixes the issue by displaying only the subject when the body is empty.

Task-[5485727](https://www.odoo.com/odoo/project/1519/tasks/5485727)

---

### DIFF
--- a/addons/mail/static/src/core/common/message_model.js
+++ b/addons/mail/static/src/core/common/message_model.js
@@ -352,7 +352,7 @@ export class Message extends Record {
     }
 
     get hasTextContent() {
-        return !this.isBodyEmpty;
+        return !this.isBodyEmpty || this.subject;
     }
 
     isEmpty = Record.attr(false, {
@@ -387,7 +387,8 @@ export class Message extends Record {
             this.isBodyEmpty &&
             this.attachment_ids.length === 0 &&
             this.trackingValues.length === 0 &&
-            !this.subtype_description
+            !this.subtype_description &&
+            !this.subject
         );
     }
 
@@ -543,6 +544,7 @@ export class Message extends Record {
             attachment_ids: [],
             attachment_tokens: [],
             body: "",
+            subject: "",
             message_id: this.id,
             ...this.thread.rpcParams,
         };
--- a/addons/mail/static/src/core/common/message_model.js
+++ b/addons/mail/static/src/core/common/message_model.js
@@ -352,7 +352,7 @@ export class Message extends Record {
     }
 
     get hasTextContent() {
-        return !this.isBodyEmpty;
+        return !this.isBodyEmpty || this.subject;
     }
 
     isEmpty = Record.attr(false, {
@@ -387,7 +387,8 @@ export class Message extends Record {
             this.isBodyEmpty &&
             this.attachment_ids.length === 0 &&
             this.trackingValues.length === 0 &&
-            !this.subtype_description
+            !this.subtype_description &&
+            !this.subject
         );
     }
 
--- a/addons/mail/static/tests/discuss_app/inbox.test.js
+++ b/addons/mail/static/tests/discuss_app/inbox.test.js
@@ -159,21 +159,39 @@ test('"reply to" composer should send message if message replied to is not a not
 
 test("show subject of message in Inbox", async () => {
     const pyEnv = await startServer();
-    const messageId = pyEnv["mail.message"].create({
-        body: "not empty",
-        model: "discuss.channel",
-        needaction: true,
-        subject: "Salutations, voyageur",
-    });
-    pyEnv["mail.notification"].create({
-        mail_message_id: messageId,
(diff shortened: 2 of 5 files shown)


addons/mail/wizard/mail_wizard_invite.py
```diff
@@ -33,8 +33,15 @@ def add_followers(self):
         return {'type': 'ir.actions.act_window_close'}
 
     def _prepare_message_values(self, document, model_name):
+        body = self.env._("%(name)s (%(email)s) added you as a follower of this %(model)s.") % {
+            "name": self.env.user.name,
+            "email": self.env.user.email,
+            "model": model_name,
+        }
+        if self.message:
+            body += Markup('<div style="color:grey;">%s</div>') % self.message
```
ramh-odoo: Embedding CSS directly in the message body isn’t a good practice. If we decide to update or redesign the UI in the future, it will be difficult to modify or maintain styles that are hard-coded within the message content.

ramh-odoo (changes requested):

addons/mail/data/mail_templates_invite.xml
```diff
@@ -8,14 +8,17 @@
...
-                    <div style="color:grey;">
-                        <t t-out="message.body"/>
+                        <t t-out="body[0]"/>
                     </div>
+                    <t t-if="len(body) > 1">
+                        <br/>
+                        <div style="color:grey;">
+                            <t t-out="body[1]"/>
+                        </div>
+                    </t>
```
ramh-odoo: I think we can’t make these changes in the stable version. Maybe we could instead consider messages with a subject but no body as non-empty and display them.

addons/mail/static/tests/discuss_app/inbox.test.js
```diff
@@ -159,21 +159,39 @@ test('"reply to" composer should send message if message replied to is not a not
...
+            notification_status: "sent",
+            notification_type: "inbox",
+            res_partner_id: serverState.partnerId,
+        },
+    ]);
     await start();
     await openDiscuss();
     await contains(".o-mail-Message", { text: "Subject: Salutations, voyageurnot empty" });
+    // Empty body: display subject only
+    await contains(".o-mail-Message", { text: "Subject: Hello, wanderer" });
```
ramh-odoo: ⚒️ Could use `:text()`

ramh-odoo (approved):
For master we could move the default message into the body ( `<t t-out='message.author_id.name'/> (<t t-out='message.author_id.email'/>) added you as a follower of this <t t-out="model_description"/>.`) instead of keeping it in the template, so we won’t face this issue again and it would stay consistent with emails as well.

addons/mail/static/src/core/common/message_model.js
```diff
@@ -352,7 +352,7 @@ export class Message extends Record {
     }
 
     get hasTextContent() {
-        return !this.isBodyEmpty;
+        return !this.isBodyEmpty || this.subject;
```
seb-odoo, line 355: Isn't check on `hasTextContent` in code simply wrong instead?

In most places, and in particular where we have the bug, we should use `isEmpty` instead I think.

akth-odoo: Yes, we should use `isEmpty` instead of `hasTextContent` here. I’ve updated the code accordingly.

akth-odoo: Tour of `website_slides` `course_review_modification` is breaking in this case, as `computeIsEmpty()` considers rating-only messages as non-empty due to `rating_value`, which causes the message body to appear even when it shouldn’t. So, I reimplemented the logic by adding a check on `hasTextContent`.

ramh-odoo: Shouldn't the `this.subject` check be handled directly within the message template itself. [here](https://github.com/odoo/odoo/blob/ccbe9107b2e5d0193a482ada16873ce923338e4e/addons/mail/static/src/core/common/message.xml#L71)?

akth-odoo: Since subject is also part of the text content, I handled it within `hasTextContent`. However, it could also be checked directly in the template.

addons/mail/static/tests/message/message.test.js
alexkuhn: Would rename test to `"Can delete a message"`, so that this is very easy to find this test and it shows this is a main test for a core feature.

alexkuhn:
@robodoo r+

PR #244658, opened 2026-01-20
[IMP] mail: mark all related notifications as read when unfollowing from inbox

nees-odoo:
Previously, unfollowing a record from the inbox only marked the notification 
on which the action was performed as read.
As a result, other unread notifications related to the same record remained 
in the inbox, even though unfollowing is a global action, indicating that the 
user is no longer interested in the entire conversation.

- Marks all notifications related to the same record as read when unfollowing 
from the inbox.
- Ensures the inbox state accurately reflects the user’s intent to stop 
following the conversation.

task-[5119331](https://www.odoo.com/odoo/project/1519/tasks/5119331)

---

### DIFF
--- a/addons/mail/static/src/core/common/message_model.js
+++ b/addons/mail/static/src/core/common/message_model.js
@@ -879,10 +879,10 @@ export class Message extends Record {
     }
 
     async unfollow() {
-        if (this.needaction) {
-            await this.setDone();
-        }
         const thread = this.thread;
+        if (this.needaction) {
+            await thread.markAllMessagesAsRead();
+        }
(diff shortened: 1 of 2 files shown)


maad-odoo (approved):

maad-odoo (commented):
**Nitpick:** _Not sure but now that `Unfollow` action is doing "more" than just unfollowing the thread, I think we could improve the Notification message as well._

Image

alexkuhn (approved):
LGTM
I would retarget 19.2, to have it sooner. 19.2 is not being actively used until next week ;)

PR #244660, opened 2026-01-20
[IMP] tests: improve assertQueries()

rco-odoo:
* Show a diff-like error message to make it more usable.
* Make wildcard string `"..."` match anything.

### DIFF
--- a/odoo/addons/base/tests/test_expression.py
+++ b/odoo/addons/base/tests/test_expression.py
@@ -1683,48 +1683,49 @@ class TestQueries(TransactionCase):
                     )
                 )
             )
-            ORDER BY "res_partner"."complete_name" ASC, "res_partner"."id" DESC
+            ORDER BY ...
         ''']):
             Model.search(domain)
 
     def test_order(self):
         Model = self.env['res.partner']
-        Model.search([('name', 'like', 'foo')])
+        domain = [('name', 'like', 'foo')]
+        Model.search(domain)
 
         with self.assertQueries(['''
             SELECT "res_partner"."id"
             FROM "res_partner"
-            WHERE ("res_partner"."active" IS TRUE AND "res_partner"."name" LIKE %s)
-            ORDER BY "res_partner"."complete_name" ASC,"res_partner"."id" DESC
+            WHERE ...
+            ORDER BY "res_partner"."complete_name" ASC, "res_partner"."id" DESC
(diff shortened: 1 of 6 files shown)


odoo/tests/common.py
```diff
@@ -613,44 +615,8 @@ def assertQueries(self, expected, flush=True):
...
-                "\n".join(actual_queries), "\n".join(expected),
-            )
-        )
-        for actual_query, expect_query in zip(actual_queries, expected):
-            self.assertIn(
-                "".join(expect_query.lower().split()),
-                "".join(actual_query.lower().split()),
-                "\n---- actual query:\n%s\n---- doesn't contain:\n%s" % (actual_query, expect_query),
-            )
+        if msg := _query_diff(expected, actual_queries):
```
kmagusiak: Nitpick: usually the order is "actual, expected".

odoo/addons/base/tests/test_expression.py
```diff
@@ -1932,7 +1932,6 @@ def test_rec_names_search(self):
 
 
 @tagged('res_partner')
-@tagged('at_install', '-post_install')  # LEGACY at_install
```
kmagusiak, line 1935: Is it for testing or do we want to remove this and make the test post-install?
Just a general question because most of ORM tests should be at-install, but in this case we can assert that other modules don't change the base behaviour of the ORM which is great. :)

kmagusiak (approved):
Look superb!

It will allow to migrate `assertQueryCount(x)` where is x low so that the tests are explicit in what is going on. Hopefully, changes will be more visible; this could be used in addons/test_website_modules/tests/test_performance.py.

odoo/addons/base/tests/test_expression.py
rco-odoo: That's indeed only for testing/demonstration purposes.  That part of the diff should go away.

odoo/tests/common.py
rco-odoo: That's the order for a test comparison method.  But the logic of the diff goes the other way around: `_query_diff(left, right)` shows with `>>>` (or `+++`) the queries that are in `right` but not in `left`.

rugo-odoo (approved):

kmagusiak (approved):
@robodoo r+

PR #244664, opened 2026-01-20
[FIX] account,account_edi_ubl_cii: Fix partner and invoice line tax recognition during import.

jugurtha-gaci:
Context :
When importing an XML invoice or vendor bill, if an invoice line is predicted to have multiple taxes, the import fails, as we assume that an invoice line should have one and only one tax.

This assumption is not always correct, especially in contexts such as international commerce where Peppol is not used.

Moreover, partner recognition was not correct or rigorous, as it was based only on the VAT number.

After this fix :
- We correctly handle the case of having multiples taxes in an invoice line.
- We correctly identify the partner based on both VAT number and name.

task-5499871

---

### DIFF
--- a/addons/account/models/partner.py
+++ b/addons/account/models/partner.py
@@ -1011,6 +1011,11 @@ class ResPartner(models.Model):
         :returns:       A partner or an empty recordset if not found.
         '''
 
+        def search_with_name_and_vat(extra_domain):
+            if not name or not vat:
+                return None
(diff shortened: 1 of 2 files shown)


jugurtha-gaci:
Replaced by : 
- https://github.com/odoo/enterprise/pull/105804
- https://github.com/odoo/odoo/pull/246159

PR #244665, opened 2026-01-20
[IMP] account_edi_ubl_cii: rearrange test files subfolder

yosuanicolaus:
This commit rearranges the UBL test files' subfolder to be one level higher, from `export/bis3/be` to `export/bis3/invoice/be`, in preparation for supporting different move formats and for better ignore schema support.

task-4891206


smetl (approved):
@robodoo r+

PR #244667, opened 2026-01-20
[FIX] pos_self_order: show missing details widget when required

lcantraine:
The "Missing Required Details" widget appeared when there was required attributes to choose in combo item or on the product page AND when the height of the screen was smaller than the content size. So sometimes, it was showing when it was not required and when we clicking on it, nothing was done because we saw everything on the screen. So, what I did, is basically change the condition for displaying the widget. Now it will appear only when a title of a required attributes is hidden by the header and this required attribute is not selected.

part ot task: 5493798

---

### DIFF
--- a/addons/pos_self_order/static/src/app/components/attribute_selection/attribute_selection.xml
+++ b/addons/pos_self_order/static/src/app/components/attribute_selection/attribute_selection.xml
@@ -4,9 +4,10 @@
         <t t-foreach="props.productTemplate.attribute_line_ids" t-as="attribute" t-key="attribute.id">
             <t t-set="availableValues" t-value="availableAttributeValue(attribute)"/>
             <t t-if="availableValues.length">
-                <h2 class="text-start py-3 m-0 d-flex align-items-center" t-att-id="attribute.attribute_id.id">
+                <t t-set="attributeRequired" t-value="attribute.attribute_id.display_type !== 'multi'"/>
+                <h2 class="text-start py-3 m-0 d-flex align-items-center" t-att-id="attribute.attribute_id.id" t-att-attr-required="attributeRequired? 'required' : undefined">
                     <t t-out="attribute.attribute_id.name"/>
-                    <span t-if="attribute.attribute_id.display_type !== 'multi'" class="badge smaller rounded-pill bg-danger text-white ms-2">Required</span>
+                    <span t-if="attributeRequired" class="badge smaller rounded-pill bg-danger text-white ms-2">Required</span>
                 </h2>
                 <div class="self_order_attribute_selection row pb-3">
                     <t t-foreach="availableAttributeValue(attribute)" t-as="value" t-key="value.id">
--- a/addons/pos_self_order/static/src/app/components/missing_required_details/missing_required_details.xml
+++ b/addons/pos_self_order/static/src/app/components/missing_required_details/missing_required_details.xml
@@ -5,8 +5,8 @@
             <button class="btn btn-primary mb-2 rounded-circle d-flex align-items-center justify-content-center" t-on-click="scrollUpToRequired" style="width:60px; height: 60px;">
                 <i class="fa fa-fw fa-arrow-up" aria-hidden="true"/>
             </button>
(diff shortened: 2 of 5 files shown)


stva-odoo (commented):
The method "shouldShowMissingDetails" of the combo_page is almost a copy-paste of the one used in the other screen.
We should extract the shared logic and move it to a common place to avoid duplication.

Instead of relying on the header element, I think we can use
`this.scrollContainerRef?.el.getBoundingClientRect().`

Also, some attributes may not be displayed (see attribute_selection.js, availableAttributeValue.availableAttributeValue). 

A solution might be to adapt the Attribute Selection component to add an HTML attribute such as `t-att-required` for the required attributes. 

Then, inside the loop, we can determine required attributes directly from the DOM:

```
const requiredAttributes = product.attribute_line_ids.filter((attr) => {
    const element = document.getElementById(attr.attribute_id.id);
    return element && element.hasAttribute("required");
});
```

addons/pos_self_order/static/src/app/utils.js
```diff
@@ -16,3 +16,39 @@ export const formatProductName = (product) => {
...
+            const attributeEl = document.getElementById(attribute.attribute_id.id);
+            if (attributeEl) {
+                if (attributeEl.getBoundingClientRect().top < containerRect.top) {
+                    return true;
+                }
+            }
+        }
+    }
+
+    return false;
```
stva-odoo:
```suggestion
    const selection = selectedValues[product.id];

   return product.attribute_line_ids.some((attr) => {
        const attributeEl = document.getElementById(attr.attribute_id.id);

        return (
            attributeEl?.hasAttribute("attr-required") &&
            !selection?.hasValueSelected(attr) &&
            attributeEl.getBoundingClientRect().top < containerRect.top
        );
    });

```

addons/pos_self_order/static/src/app/components/attribute_selection/attribute_selection.xml
```diff
@@ -4,7 +4,7 @@
         <t t-foreach="props.productTemplate.attribute_line_ids" t-as="attribute" t-key="attribute.id">
             <t t-set="availableValues" t-value="availableAttributeValue(attribute)"/>
             <t t-if="availableValues.length">
-                <h2 class="text-start py-3 m-0 d-flex align-items-center" t-att-id="attribute.attribute_id.id">
+                <h2 class="text-start py-3 m-0 d-flex align-items-center" t-att-id="attribute.attribute_id.id" t-att-required="attribute.attribute_id.display_type !== 'multi' ? 'required' : undefined">
                     <t t-out="attribute.attribute_id.name"/>
                     <span t-if="attribute.attribute_id.display_type !== 'multi'" class="badge smaller rounded-pill bg-danger text-white ms-2">Required</span>
```
stva-odoo:
```suggestion
                <t t-set="attribute-required" t-value="attribute.attribute_id.display_type !== 'multi'"/>
                <h2 class="text-start py-3 m-0 d-flex align-items-center" t-att-id="attribute.attribute_id.id" t-att-attr-required="attribute-required? 'required' : undefined">
<t t-out="attribute.attribute_id.name"/>
<span t-if="attribute-required" class="badge smaller rounded-pill bg-danger text-white ms-2">Required</span>
```

I suggest to use `attr-required` instead of required to avoid potential conflict with the html required attribute

addons/pos_self_order/static/src/app/components/missing_required_details/missing_required_details.xml
```diff
@@ -6,7 +6,7 @@
                 <i class="fa fa-fw fa-arrow-up" aria-hidden="true"/>
             </button>
             <h2 class="mb-3">See what you missed</h2>
```
stva-odoo:
```suggestion
            <h2 class="mb-3" t-on-click="scrollUpToRequired">See what you missed</h2>
```

stva-odoo (commented):
Could you also modify your commit message: [FIX] pos_self_order: show missing details widget when required

stva-odoo:
robodoo r+

PR #244673, opened 2026-01-20
[IMP] mail: display priority in message inbox

phenix-factory:
task-5385387

PR enterprise: https://github.com/odoo/enterprise/pull/105614

### DIFF
--- a/addons/mail/models/mail_message.py
+++ b/addons/mail/models/mail_message.py
@@ -1215,6 +1215,17 @@ class MailMessage(models.Model):
                     predicate=lambda t: target_user and add_followers and non_channel_records,
                     value=lambda t: follower_by_record_and_partner[t, target_user.partner_id],
                 ),
+                res.attr(
+                    "priority",
+                    value=lambda t: t[t._priority_field],
+                    predicate=lambda t: hasattr(t, "_priority_field"),
+                    sudo=True,
+                ),
+                res.attr(
+                    "priority_definition",
+                    predicate=lambda t: hasattr(t, "_priority_field"),
+                    value=lambda t: t.fields_get([t._priority_field], ["selection"])[t._priority_field]["selection"],
+                ),
             ),
             as_thread=True,
             value=record_by_message.get,
--- a/addons/mail/static/src/core/common/message.xml
+++ b/addons/mail/static/src/core/common/message.xml
@@ -52,7 +52,9 @@
                                     (from <a t-att-href="message.resUrl"><t t-esc="message.thread.prefix"/><t t-esc="message.channel_id.displayName or message.default_subject"/></a>)
                                 </t>
                                 <t t-else="">
-                                    on <a t-if="message.thread.displayName" t-att-href="message.resUrl" t-esc="message.thread.displayName"/><em class="pe-1 text-decoration-line-through" t-else="">Deleted document</em>
+                                    on
+                                    <Priority t-if="message.thread.priority" thread="message.thread"/>
+                                    <t t-out="nbsp"/><a t-if="message.thread.displayName" t-att-href="message.resUrl" t-esc="message.thread.displayName"/><em class="pe-1 text-decoration-line-through" t-else="">Deleted document</em>
                                 </t>
                             </small>
                             <div t-if="props.message.scheduledDatetime" t-att-class="{ 'ms-2': (env.inChatWindow and isAlignedRight) or (isPersistentMessageFromAnotherThread) }" t-att-title="props.message.scheduledDateSimple">
--- a/addons/mail/static/src/core/common/thread_model.js
+++ b/addons/mail/static/src/core/common/thread_model.js
@@ -170,6 +170,10 @@ export class Thread extends Record {
      * when fetching newer messages.
      */
     pendingNewMessages = fields.Many("mail.message");
+    /** @type {'0'|'1'|'2'|'3'} */
+    priority;
+    /** @type {Array<[string,string]>} */
+    priority_definition;
     needactionMessages = fields.Many("mail.message", {
         inverse: "threadAsNeedaction",
         sort: (message1, message2) => message1.id - message2.id,
--- a/addons/crm/models/crm_lead.py
+++ b/addons/crm/models/crm_lead.py
@@ -96,6 +96,7 @@ class CrmLead(models.Model):
     _primary_email = 'email_from'
     _check_company_auto = True
     _track_duration_field = 'stage_id'
+    _priority_field = 'priority'
 
     # Description
     name = fields.Char(
--- a/addons/mail/static/src/core/common/message.js
+++ b/addons/mail/static/src/core/common/message.js
@@ -12,6 +12,7 @@ import { RelativeTime } from "@mail/core/common/relative_time";
 import { htmlToTextContentInline } from "@mail/utils/common/format";
 import { isEventHandled, markEventHandled } from "@web/core/utils/misc";
 import { renderToElement } from "@web/core/utils/render";
+import { nbsp } from "@web/core/utils/strings";
 
 import {
     Component,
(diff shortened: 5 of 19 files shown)


addons/mail/static/src/core/common/message.xml
```diff
@@ -52,7 +52,11 @@
                                     (from <a t-att-href="message.resUrl"><t t-esc="message.thread.prefix"/><t t-esc="message.channel_id.displayName or message.default_subject"/></a>)
                                 </t>
                                 <t t-else="">
-                                    on <a t-if="message.thread.displayName" t-att-href="message.resUrl" t-esc="message.thread.displayName"/><em class="pe-1 text-decoration-line-through" t-else="">Deleted document</em>
+                                    on
+                                    <t t-if="message.thread.priority">
+                                        <span t-foreach="Array(window.parseInt(this.message.thread.priority))" t-as="star" t-key="star_index" class="o-mail-priority-star fa fa-star"/> 
+                                    </t> 
+                                    <t t-out="nbsp"/><a t-if="message.thread.displayName" t-att-href="message.resUrl" t-esc="message.thread.displayName"/><em class="pe-1 text-decoration-line-through" t-else="">Deleted document</em>
```
alexkuhn, line 57: Instead of importing `nbsp` and using it here, couldn't you simply play with spacing through margin or gap?

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -170,6 +170,7 @@ export class Thread extends Record {
      * when fetching newer messages.
      */
     pendingNewMessages = fields.Many("mail.message");
+    priority;
```
alexkuhn, line 174: Would be nice to type the attribute, so we know whether we should expect string or value (and also the semantics behind `1` and `3`, is greater more a priority or the inverse? And what's the scale, etc.).

addons/mail/models/mail_message.py
```diff
@@ -1213,6 +1213,11 @@ def _store_message_fields(
                     predicate=lambda t: target_user and add_followers and non_channel_records,
                     value=lambda t: follower_by_record_and_partner[t, target_user.partner_id],
                 ),
+                res.attr(
+                    "priority",
+                    sudo=True,
+                    predicate=lambda t: "priority" in t._fields,
+                ),
```
alexkuhn, line 1223: I would return the max scale of star, so we show the fill and unfilled messages.
This would make the visual consistent with priority widget that shows the max amount of star.

It would be also nice to return the label part of the priority field, so that the label can have "Medium" or "Urgent" for clarity. As the amount, max scale, and label all contribute to giving the right interpretation of stars, I feel this is important here.

Also we have lots of start icons in discuss app, and we need proper label to get they are different features:
- Starred messages / mailbox
- "Add to Favorites" category
- Member owners / admins
- Looking for help conversations with self expertise

And here a 5th case with chatter record priority. That's why label + max scale with unfilled star outline is important.

alexkuhn (commented):
Would also show stars in messaging menu item when relevant. Quick example:
Image

(and unfilled star for max scale + label as with other comments)

LGTM otherwise!

addons/mail/static/src/core/common/message.xml
alexkuhn: Ok for `nsbp` here.

addons/mail/models/mail_message.py
phenix-factory: > I would return the max scale of star, so we show the fill and unfilled messages.
> This would make the visual consistent with priority widget that shows the max amount of star.

I thought of that and ruled it out since it would take space inside an already small area with a lot of info. 
The max is always 3, and as far as I know, it's consistent across Odoo.

> It would be also nice to return the label part of the priority field, so that the label can have "Medium" or "Urgent" for clarity. As the amount, max scale, and label all contribute to giving the right interpretation of stars, I feel this is important here.

I agree, but this information isn't available, only the number is stored inside the database.

alexkuhn: > I thought of that and ruled it out since it would take space inside an already small area with a lot of info. The max is always 3, and as far as I know, it's consistent across Odoo.

- `mrp_eco`, `stock_move`, `purchase_order`, `repair_order` use `0` and `1`. The other models use `0`-`3`.

Since inbox can have messages from any model, seeing the max amount of star makes sense.
And if I'm not mistaking, can be edited with web_studio or other means.
Besides for consistency with priority widget, this helps visually that this looks the same in all contexts.

I understand the argument with space constraints, but I feel that seeing the unfilled star is more helpful. When there are 2 stars with 1 unfilled, it's good to see it's not the max priority among all records of this model. And also in this case, many people spot the unfilled star instead of counting the amount of filled star in their head, this is faster to scan ;)

> I agree, but this information isn't available, only the number is stored inside the database.

Then how UI of priority widget shows it? For consistency we should copy them.

addons/mail/static/src/core/public_web/priority.xml
```diff
@@ -0,0 +1,19 @@
...
+                t-key="star_index"
+            >
+                <span
+                    class="fa"
+                    t-att-class="{
+                        'o-mail-priority-star fa-star': star_index+1 lte priority,
+                        'fa-star-o': star_index+1 gt priority,
+                    }"
+                />
+            </t>
```
seb-odoo: overkill new lines

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -170,6 +170,10 @@ export class Thread extends Record {
      * when fetching newer messages.
      */
     pendingNewMessages = fields.Many("mail.message");
+    /** @type {'0'|'1'|'2'|'3'} */
+    priority;
+    /**  @type {Array} */
+    priority_definition;
```
seb-odoo: It would be nice to clarify the shape, it's still unclear what it contains, and it doesn't really help the editor this way.

addons/mail/static/src/core/public_web/priority.xml
```diff
@@ -0,0 +1,19 @@
...
+        <div class="d-inline" t-att-aria-label="label" t-att-title="label">
+            <t t-foreach="Array(maxStar)"
+                t-as="star"
+                t-key="star_index"
+            >
+                <span
+                    class="fa"
+                    t-att-class="{
+                        'o-mail-priority-star fa-star': star_index+1 lte priority,
+                        'fa-star-o': star_index+1 gt priority,
```
seb-odoo: spaces around (+) operator

addons/mail/models/mail_message.py
```diff
@@ -1213,6 +1213,16 @@ def _store_message_fields(
...
                 ),
+                res.attr(
+                    "priority",
+                    predicate=lambda t: "priority" in t._fields,
+                    sudo=True,
+                ),
+                res.attr(
+                    "priority_definition",
+                    predicate=lambda t: "priority" in t._fields,
+                    value=lambda t: t.fields_get(["priority"], ["selection"]).get("priority").get("selection")
```
seb-odoo: `get` can be `[]` if we know the key is always present

addons/mail/static/src/core/common/message.xml
```diff
@@ -52,7 +52,11 @@
                                     (from <a t-att-href="message.resUrl"><t t-esc="message.thread.prefix"/><t t-esc="message.channel_id.displayName or message.default_subject"/></a>)
                                 </t>
                                 <t t-else="">
-                                    on <a t-if="message.thread.displayName" t-att-href="message.resUrl" t-esc="message.thread.displayName"/><em class="pe-1 text-decoration-line-through" t-else="">Deleted document</em>
+                                    on
+                                    <t t-if="message.thread.priority">
+                                        <Priority thread="message.thread"/>
+                                    </t>
```
seb-odoo: t-if can be on the component itself

addons/mail/static/src/core/common/message.xml
seb-odoo: I don't get why we need nbsp specifically?

phenix-factory: If, for whatever reason, the title would be on the next line, the stars are going to follow and always be before the title because nbsp is forcing this behavior.

phenix-factory:
> But we don't have threads in Odoo, just unread messages.

You might want to elaborate, because this doesn't make much sense to me.

seb-odoo (approved):
@robodoo r+

PR #244677, opened 2026-01-20
[FIX] im_livechat: non deterministic end session test

tsm-odoo:
This commit fixes the `visitor leaving ends the livechat conversation` test that was sometimes failing.

There is a race between the simulation of the guest leaving the chat that uses `withGuest` which is known to be problematic and the `action_unfollow` method, called when the chat window is closed.

The unfollow method deletes the agent member, but since the guest cookie is still present, the `channels_as_member` rpc retrieves the guest member.

The rest of the code doesn't expect this mismatch and a crash occurs.

This commit ensures the `channels_as_member` rpc is properly done before closing the chat window.

runbot-237947

Current behavior before PR:

---

### DIFF
--- a/addons/im_livechat/static/tests/channel_join_leave.test.js
+++ b/addons/im_livechat/static/tests/channel_join_leave.test.js
@@ -1,4 +1,4 @@
-import { Command, serverState } from "@web/../tests/web_test_helpers";
+import { Command, onRpc, serverState } from "@web/../tests/web_test_helpers";
 import { defineLivechatModels } from "@im_livechat/../tests/livechat_test_helpers";
 import {
     click,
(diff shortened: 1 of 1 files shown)


phenix-factory (approved):
@robodoo r+ delegate+

addons/im_livechat/static/tests/channel_join_leave.test.js
```diff
@@ -141,8 +141,29 @@ test("visitor leaving ends the livechat conversation", async () => {
...
+            post_data: {
+                body: "I need help!",
+                message_type: "comment",
+                subtype_xmlid: "mail.mt_comment",
+            },
+            thread_id: channel_id,
+            thread_model: "discuss.channel",
+        })
+    );
+    await expect.waitForSteps(["fetch channels_as_member"], { timeout: 3000 });
```
alexkuhn, line 156: Why not setup with at least 2 messages in conversations and assert `.Message { count: 2 }` like in many other discuss tests?

This is simplest and functionally relevant, whereas `onRpc` + making a message post that can happen before message fetch still is very prone to many other kinds of race conditions.

tsm-odoo (commented):
@robodoo r+

PR #244678, opened 2026-01-20
[FIX] base_import_module: hide "All" option from Apps search panel

chga-odoo:
Hide the "All" option in the Apps search panel to avoid mixing
Official Apps and Industries. 

Task ID: 5222706

### DIFF
--- a/addons/base_import_module/static/tests/apps_searchpanel.test.js
+++ b/addons/base_import_module/static/tests/apps_searchpanel.test.js
@@ -1,0 +1,112 @@
+import { describe, expect, test } from "@odoo/hoot";
+import { queryAllTexts } from "@odoo/hoot-dom";
+import {
+    contains,
+    defineActions,
+    defineModels,
+    defineWebModels,
+    fields,
+    getService,
+    models,
+    mountWithCleanup,
+} from "@web/../tests/web_test_helpers";
+import { WebClient } from "@web/webclient/webclient";
+
+class Module extends models.Model {
+    name = fields.Char();
+    category_id = fields.Many2one({ string: "category", relation: "category" });
+    module_type = fields.Selection({
+        selection: [
+            ["official", "Official"],
+            ["industries", "Industries"],
+        ],
+    });
+
+    _records = [
+        {
+            id: 1,
+            name: "CRM",
+            module_type: "official",
+        },
+        {
+            id: 2,
+            name: "Rental",
+            module_type: "official",
+        },
+        {
+            id: 3,
+            name: "Hotel",
+            module_type: "industries",
+        },
+    ];
+    _views = {
+        list: `<list><field name="name"/></list>`,
+        search: `
+            <search>
+                <searchpanel view_types="list" class="o_apps_searchpanel">
+                    <field name="module_type" string="Apps" expand="1"/>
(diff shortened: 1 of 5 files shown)


vava-odoo:
@robodoo squash

vava-odoo:
@kmagusiak @rco-odoo To me, this approach is fine but I'd like a second thought. Thanks in advance

kmagusiak:
@vava-odoo This is a question for @odoo/rd-framework-js 
"A quick and rather clean solution would be to hide the "All" selection value for now."

vava-odoo:
Thanks @jeanschoenlaub for the review
> LGTM code wise, but maybe add a test ? (especialy cause the selector is a bit fickle)

@chga-odoo could you add one?

> Note for AAB: Maybe we could enable the option to hide "All" from the framework side, but I asked AI to grep the codebase for other people hiding "All" and it did not find any.

I also thought about that option, but not sure it is necessary. There is already another option with the multi select which removes the "All" option.

The problem here is that none of the options work since we cannot (and don't want to) mix both module_type.

(but if you ever add that option, feel free to activate it here and remove this custom code)

addons/base_import_module/static/tests/apps_searchpanel.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+        expect(queryAllTexts`.o_search_panel_category_value`).toEqual([
+            "All",
+            "Official",
+            "Industries",
+        ]);
+
+        expect(
+            `.o_search_panel_category_value:nth-child(2) header.active`
+        ).toHaveCount(1);
```
vava-odoo: maybe check that only 2 records are displayed (hotel is not)

vava-odoo: maybe what is more important is that active one is "Official", no?

```suggestion
            `.o_search_panel_category_value header.active`
        ).toHaveText("Official");
```

addons/base_import_module/static/tests/apps_searchpanel.test.js
```diff
@@ -0,0 +1,96 @@
...
+    test("Official is selected by default", async () => {
+        await mountWithCleanup(WebClient);
+
+        await getService("action").doAction(1);
+
+        expect(".o_search_panel").toHaveCount(1);
+        expect(".o_search_panel_section").toHaveCount(1);
+
+        expect(queryAllTexts`.o_search_panel_category_value`).toEqual([
+            "All",
```
vava-odoo: make sure this is hidden (display none)

addons/base_import_module/static/tests/apps_searchpanel.test.js
```diff
@@ -0,0 +1,96 @@
...
+            "Industries",
+        ]);
+
+        expect(
+            `.o_search_panel_category_value:nth-child(2) header.active`
+        ).toHaveCount(1);
+
+        expect(
+            `.o_search_panel_category_value:nth-child(2) header`
+        ).toHaveText("Official");
```
vava-odoo: you could also
- add a click on Industries and check 1 (hotel) is displayed
- check that "All" still is visible on Categories below

PR #244683, opened 2026-01-20
[IMP] pos_sale: ignore fully delivered SO lines when settle

meji-odoo:
Before this commit:
---
- All sale order lines were considered when selling a sale order in POS, including lines that were already fully delivered.

After this commit:
---
- Exclude SO lines where the delivered quantity equals the ordered quantity.
- Prevent processing zero remaining quantity lines when settle a SO in POS.

task-5494693

---


addons/pos_sale/static/src/app/services/pos_store.js
```diff
@@ -167,6 +167,10 @@ patch(PosStore.prototype, {
                 }
             }
             newLine.setQuantityFromSOL(converted_line);
+            if (newLine.qty === 0) {
+                newLine.delete();
+                continue;
+            }
```
davidmonnom: We should be able to avoid creating the line to delete it after

davidmonnom (changes requested):

addons/pos_sale/static/src/app/services/pos_store.js
meji-odoo: @davidmonnom 

When a POS order line is initially created from a sale order line, the POS line quantity is set to `product_uom_qty`.
After that, `setQuantityFromSOL` is called. In the normal flow, the quantity is recalculated using `saleOrderLine.qty_to_invoice` and `saleOrderLine.qty_delivered`.

However, in other modules like **pos_sale_loyalty** and **pos_repair**, the logic fully depends on `product_uom_qty`.

So if we want to avoid creating a line instead of deleting it, we would need to update all three modules.
I’m not sure if this is what you were referring to, but I think we should keep the changes minimal.

What do you think?

Thanks for the review, and have a great day ahead!

davidmonnom: You can easily update `setQuantityFromSOL` to get the right quantity before creating the line :)

meji-odoo: We first create a line, and after that we call `setQuantityFromSOL` :face_with_diagonal_mouth:

davidmonnom (approved):
@robodoo r+

PR #244686, opened 2026-01-20
[IMP] hr: restrict HR Responsible to internal users of current company

vikri-odoo:
Updated the "HR Responsible" field under the Approvers group in the Employee settings to ensure better data integrity.

Changes:
Added a domain filter to the hr_responsible_id field.
Restricted selection to Internal Users only (share=False). This prevents external or portal users from being incorrectly assigned as HR responsibles

External users can be assigned to employee as HR responsible

Current behavior before PR:
External users can be assigned to employee as HR responsible

Only Internal users can be assigned as HR responsible

task-5602911

---

### DIFF
--- a/addons/hr/models/hr_employee.py
+++ b/addons/hr/models/hr_employee.py
@@ -84,6 +84,11 @@ class HrEmployee(models.Model):
         groups="hr.group_hr_user,hr_payroll.group_hr_payroll_user",
     )
 
+    def _get_hr_responsible_domain(self):
+        return "[('share', '=', False), ('company_ids', 'in', company_id), ('all_group_ids', 'in', %s)]" % self.env.ref('hr.group_hr_user').id
+
+    hr_responsible_id = fields.Many2one(related='version_id.hr_responsible_id', readonly=False, inherited=True, domain=_get_hr_responsible_domain, groups="hr.group_hr_user")
+
     @api.model
     def _lang_get(self):
         return self.env['res.lang'].get_installed()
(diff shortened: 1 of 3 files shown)


addons/hr/views/hr_employee_views.xml
```diff
@@ -496,7 +496,7 @@
                                         </div>
                                     </group>
                                     <group name="managers" string="Approvers">
-                                        <field name="hr_responsible_id" widget="many2one_avatar_user"/>
+                                        <field name="hr_responsible_id" widget="many2one_avatar_user" domain="[('share', '=', False), ('company_ids', 'in', company_id)]"/>
```
Einar-eian: This works, but I wonder if it would make more sense to define this domain directly in the field declaration (in python). I see it is already done on the version but I think the magic-inherit-related-generator that adds versions fields to employee doesn't include the domain. So you will probably need to declare this field on the employee and set the domain there.

Einar-eian (changes requested):
Thank you for your work 💪 I have one question/theory for a slightly better approach but otherwise everything looks good!

addons/hr/views/hr_employee_views.xml
vikri-odoo: thanks for the suggestion updated the changes accordingly.

addons/hr/models/hr_employee.py
```diff
@@ -84,6 +84,11 @@ class HrEmployee(models.Model):
         groups="hr.group_hr_user,hr_payroll.group_hr_payroll_user",
     )
 
+    def _get_hr_responsible_domain(self):
+        return "[('share', '=', False), ('company_ids', 'in', company_id), ('all_group_ids', 'in', %s)]" % self.env.ref('hr.group_hr_user').id
+
+    hr_responsible_id = fields.Many2one('res.users', domain=_get_hr_responsible_domain)
```
Einar-eian: If I remember correctly, you need to also explicitly define that this field is related to the version field (so they don't go out of sync).

```suggestion
    hr_responsible_id = fields.Many2one(
        related="version_id.hr_responsible_id",
        inherited=True,
        domain=_get_hr_responsible_domain,
        groups="hr.group_hr_user",
    )

```

Einar-eian (commented):
Thank you for the update 😄 I left a small comment

Einar-eian (approved):
Thank you for your work! LGTM 🚀 
@robodoo r+

PR #244687, opened 2026-01-20
[IMP] mail: check author name when searching messages

nmak-odoo:

Previously, message search only matched message content and subject, making it difficult to find messages from a specific author.

This commit expands the search scope to include the author's name, making it easier to find messages from a specific sender in a thread, with matching author names highlighted in the search results.

task-5473532

---

### DIFF
--- a/addons/mail/models/mail_message.py
+++ b/addons/mail/models/mail_message.py
@@ -967,6 +967,10 @@ class MailMessage(models.Model):
             message_domain = Domain.OR([
                 # sudo: access to attachment is allowed if you have access to the parent model
                 [("attachment_ids", "in", self.env["ir.attachment"].sudo()._search([("name", "ilike", search_term)]))],
+                # sudo: res.partner - allow searching by author name
+                [("author_id", "in", self.env["res.partner"].sudo()._search([("name", "ilike", search_term)]))],
+                # sudo: mail.guest - allow searching by guest name
+                [("author_guest_id", "in", self.env["mail.guest"].sudo()._search([("name", "ilike", search_term)]))],
                 [("body", "ilike", search_term)],
                 [("subject", "ilike", search_term)],
                 [("subtype_id.description", "ilike", search_term)],
--- a/addons/mail/static/tests/core/search_highlight.test.js
+++ b/addons/mail/static/tests/core/search_highlight.test.js
@@ -118,18 +118,24 @@ test("Search highlight", async () => {
 test("Display highlighted search in chatter", async () => {
     patchUiSize({ size: SIZES.XXL });
     const pyEnv = await startServer();
-    const partnerId = pyEnv["res.partner"].create({ name: "John Doe" });
+    const partnerId = pyEnv["res.partner"].create({ name: "Groot" });
     pyEnv["mail.message"].create({
-        body: "not empty",
+        author_id: partnerId,
+        body: "I am Groot",
         model: "res.partner",
         res_id: partnerId,
     });
     await start();
     await openFormView("res.partner", partnerId);
     await click("[title='Search Messages']");
-    await insertText(".o_searchview_input", "empty");
+    await insertText(".o_searchview_input", "Groot");
     triggerHotkey("Enter");
-    await contains(`.o-mail-SearchMessageResult .o-mail-Message span.${HIGHLIGHT_CLASS}`);
+    await contains(
+        `.o-mail-SearchMessageResult .o-mail-Message-author .${HIGHLIGHT_CLASS}:text('Groot')`
+    );
+    await contains(
+        `.o-mail-SearchMessageResult .o-mail-Message-body .${HIGHLIGHT_CLASS}:text('Groot')`
+    );
 });
 
 test("Display multiple highlighted search in chatter", async () => {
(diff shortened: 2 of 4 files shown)


maad-odoo (approved):

addons/mail/models/mail_message.py
```diff
@@ -963,6 +963,7 @@ def _message_fetch(self, domain, *, thread=None, search_term=None, is_notificati
             message_domain = Domain.OR([
                 # sudo: access to attachment is allowed if you have access to the parent model
                 [("attachment_ids", "in", self.env["ir.attachment"].sudo()._search([("name", "ilike", search_term)]))],
+                [("author_id.display_name", "ilike", search_term)],
```
phenix-factory: This field isn't stored inside the database: some ORM magic is involved.
I'm not certain it's wise to include it as a search element, as it might cause performance issues.

seb-odoo: Should we just search `name` then? As we usually display `name` anyway.

Also shouldn't it also check for `author_guest_id` ?

phenix-factory: It's probably better to use name, it's going to need to check `res_partner` table, but at least it's SQL doing the job.

> Also shouldn't it also check for author_guest_id ?

If we want to search for guest name, then yes. But aren't most of them just named "guest"?

addons/mail/models/mail_message.py
```diff
@@ -967,6 +967,10 @@ def _message_fetch(self, domain, *, thread=None, search_term=None, is_notificati
             message_domain = Domain.OR([
                 # sudo: access to attachment is allowed if you have access to the parent model
                 [("attachment_ids", "in", self.env["ir.attachment"].sudo()._search([("name", "ilike", search_term)]))],
+                # sudo: res.partner - allow searching by author name
+                [("author_id", "in", self.env["res.partner"].sudo()._search([("name", "ilike", search_term)]))],
+                # sudo: mail.guest - allow searching by guest name
+                [("author_guest_id", "in", self.env["mail.guest"].sudo()._search([("name", "ilike", search_term)]))],
```
seb-odoo, line 973: Isn't it the definition of the `any` operator rather than `in` + `_search`?

addons/mail/static/tests/core/search_highlight.test.js
```diff
@@ -191,13 +200,17 @@ test("Display multiple highlighted search in Discuss", async () => {
     await contains(`.o-mail-SearchMessagesPanel .o-mail-Message span.${HIGHLIGHT_CLASS}`, {
         count: 2,
     });
+    await insertText(".o_searchview_input", "mitchell", { replace: true });
+    triggerHotkey("Enter");
+    await contains(`.o-mail-SearchMessagesPanel .o-mail-Message-author span.${HIGHLIGHT_CLASS}`);
```
seb-odoo: I don't think we should have that assert on each test.
Either pick only one test, or change test setup so the author name is also present in body so we can assert both at once without changing search term, or rather make a new test specific for searching on name instead.

addons/mail/static/tests/core/search_highlight.test.js
```diff
@@ -211,4 +224,27 @@ test("Display highlighted with escaped character must ignore them", async () =>
...
+        model: "discuss.channel",
+        res_id: channelId,
+    });
+    await start();
+    await openDiscuss(channelId);
+    await contains(".o-mail-Message");
+    await click("button[title='Search Messages']");
+    await insertText(".o_searchview_input", "saul");
+    triggerHotkey("Enter");
+    await contains(`.o-mail-SearchMessagesPanel .o-mail-Message-author span.${HIGHLIGHT_CLASS}`);
```
seb-odoo: (new) discuss/channel related tests should be inside a `discuss` subfolder

addons/mail/models/mail_message.py
nmak-odoo: From guest contexts, the search on name requires an explicit `sudo()._search()` to bypass ACLs on `mail.guest` and `res.partner` records.

Since the `any` operator runs with the current user’s access rights, it would raise an `AccessError` for guest users.

seb-odoo: Thank you for the clarification!

addons/mail/static/tests/discuss/search_messages_panel.test.js
```diff
@@ -350,3 +350,35 @@ test("Search a message containing single quotes", async () => {
...
+        model: "discuss.channel",
+        res_id: channelId,
+    });
+    await start();
+    await openDiscuss(channelId);
+    await contains(".o-mail-Message", { count: 2 });
+    await click("button[title='Search Messages']");
+    await insertText(".o_searchview_input", "Mitchell");
+    triggerHotkey("Enter");
+    await contains(`.o-mail-SearchMessagesPanel .o-mail-Message .${HIGHLIGHT_CLASS}`);
```
tsm-odoo: This test doesn't really ensure the feature works properly. The JS test should check that the author name is properly highlighted. currently, it only shows that something in the message is. We don't check which message is either.

addons/mail/static/tests/discuss/search_messages_panel.test.js
```diff
@@ -364,3 +350,40 @@ test("Search a message containing single quotes", async () => {
     triggerHotkey("Enter");
     await contains(".o-mail-SearchMessagesPanel .o-mail-Message");
 });
+
+test.tags("desktop");
+test("Search a message by author name", async () => {
```
tsm-odoo: As I said previously this test is not useful. It only tests that the mock server is properly implemented. We already have `Display highlighted search in chatter` that ensures the author name is properly highlighted.  If we really want to, we can add an assertion on the existing `Search should be hightlighted`.

tsm-odoo (approved):
@robodoo r+ delegate+

PR #244693, opened 2026-01-20
[FIX] mail: fix inbox mark as read test

tsm-odoo:
The `Opening thread with needaction messages should mark all messages of thread as read` ensures that opening a channel marks the related inbox messages as read.

However, this test can fail in a non-deterministic fashion. There are two flows that can mark inbox messages as read: fetching the messages or a manual request after opening the channel.

The test expects the rpc to be issued, but if the messages are fetched before, it won't happen thus making the test fail.

This commit awaits the initial fetch to prevent this issue. As a bonus, the second flow is also tested.

runbot-237970

Current behavior before PR:

---


phenix-factory (approved):
@robodoo r+ delegate+

PR #244694, opened 2026-01-20
[FIX] website_profile: use _can_edit_country in the profile save route

Megaaaaaa:
In 'save_edited_profile()', we check that the user can edit his country_id if he tries. As from saas-18.4, a new helper '_can_edit_country()' is available. This commit uses that helper to perform the check.

Follow-up of #239789 

Task-5331916


Megaaaaaa:
robodoo up to saas-19.1

PR #244702, opened 2026-01-20
[FIX] point_of_sale: Refund combo product with multiple qtys

macocianradu:
When refunding a combo item, the 'To Refund' text would always show the same quantity for all the orerlines as for the combo. But combos could have orderlines with a higher quantity than the combo itself (i.e. 3 menus with 2 burgers each - 6 burgers in total. Now the POS would only let us refund up to the limit qty of the combo, so 3 instead of all 6 burgers)

After the fix, we check the quantity of each line in the combo and we refund the full quantity (i.e. if you have a 3 menus with 2 burgers each - 6 burgers in total. The burgers will be divided per combo, so each menu refund will automatically refund 2 burgers.)

Task-[5503962](https://www.odoo.com/odoo/project/1737/tasks/5503962)

---

### DIFF
--- a/addons/point_of_sale/static/src/app/screens/ticket_screen/ticket_screen.js
+++ b/addons/point_of_sale/static/src/app/screens/ticket_screen/ticket_screen.js
@@ -231,6 +231,38 @@ export class TicketScreen extends Component {
             this.setOrder(refundOrder);
         }
     }
+    _setToRefundDetail(toRefundDetail, buffer) {
+        // When already linked to an order, do not modify the to refund quantity.
+        if (toRefundDetail.destionation_order_id) {
+            return this.numberBuffer.reset();
+        }
+
+        toRefundDetail.refundableQty = toRefundDetail.line.qty - toRefundDetail.line.refundedQty;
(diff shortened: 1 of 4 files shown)


loukwets (approved):

addons/point_of_sale/static/src/app/screens/ticket_screen/ticket_screen.js
```diff
@@ -231,52 +231,66 @@ export class TicketScreen extends Component {
...
-                        });
-                    }
-                } else {
-                    toRefundDetail.qty = quantity;
-                }
-            }
+        const parentToRefundDetail = this.getToRefundDetail(orderline);
+        this._setToRefundDetail(parentToRefundDetail, buffer);
+
+        for (const comboLine of orderline.combo_line_ids) {
```
stva-odoo, line 291: I think we need to stop the process if the refunded quantity for the parent exceeds the maximum value.
 
I entered 2 as the refund quantity the "Maximum Exceeded" is displayed, but the combo line shows “To refund: 2” and I am able to create a refund.

image

stva-odoo:
robodoo retry

PR #244704, opened 2026-01-20
[FIX] web: file_upload_service handles errors from different API

kebeclibre:
Odoo controllers often don't use coherent API they respond with either json, jsonrpc or html. On the client side, we have no reliable way to know this, and sometimes it is not possible to make a guess (rpc.js makes the guess that controllers are of json type -- actually meaning jsonrpc)

The file upload service suffered from this and did not handle those error cases well.

After this commit, the file upload service handles much of the errors.

Current behavior before PR:

---

### DIFF
--- a/addons/web/static/src/core/file_upload/file_upload_service.js
+++ b/addons/web/static/src/core/file_upload/file_upload_service.js
@@ -15,7 +15,7 @@ export const fileUploadService = {
         return new window.XMLHttpRequest();
     },
 
-    start(env, { notificationService }) {
+    start(env, { notification: notificationService }) {
         const uploads = reactive({});
         let nextId = 1;
         const bus = new EventBus();
@@ -67,26 +67,76 @@ export const fileUploadService = {
             });
             // Load listener
             xhr.addEventListener("load", () => {
+                try {
+                    handleResponse();
+                } catch (e) {
+                    onError(e);
+                    return;
+                }
                 delete uploads[upload.id];
                 upload.state = "loaded";
                 bus.trigger("FILE_UPLOAD_LOADED", { upload });
             });
-            // Error listener
-            xhr.addEventListener("error", async () => {
+
+            function handleResponse() {
+                const resp = xhr.responseText ?? xhr.response;
+                let error;
+                let errorMessage = "";
+                if (!(xhr.status >= 200 && xhr.status < 300)) {
+                    error = true;
+                }
+                if (resp) {
+                    let content = resp;
+                    if (typeof content === "string") {
+                        try {
+                            content = JSON.parse(content);
+                        } catch {
+                            try {
+                                content = new DOMParser().parseFromString(content, "text/html");
+                            } catch {
+                                /** pass */
+                            }
+                        }
+                    }
+                    // Not sure what to do if the content is neither JSON nor HTML
+                    // Let's the call be successful then....
+                    if (error && content instanceof Document) {
+                        errorMessage = content.body.textContent;
+                    } else if (content instanceof Object) {
+                        if (content.error) {
+                            // https://www.jsonrpc.org/specification#error_object
+                            error = true;
+                            if (content.error.data) {
+                                // JsonRPCDispatcher.handle_error and http.serialize_exception
+                                errorMessage = `${content.error.data.name}: ${content.error.data.message}`;
+                            } else {
+                                errorMessage = content.error.message || errorMessage;
+                            }
+                        }
+                    }
+                }
+                if (error) {
+                    throw new Error(errorMessage);
+                }
+                return true;
+            }
+
+            function onError(error) {
+                const defaultErrorMessage = _t("An error occured while uploading.");
                 delete uploads[upload.id];
                 upload.state = "error";
+                const displayError = params.displayErrorNotification ?? true;
                 // Disable this option if you need more explicit error handling.
-                if (
-                    params.displayErrorNotification !== undefined &&
-                    params.displayErrorNotification
-                ) {
-                    notificationService.add(_t("An error occured while uploading."), {
+                if (displayError) {
+                    notificationService.add(error?.message || defaultErrorMessage, {
                         title: _t("Error"),
                         sticky: true,
                     });
                 }
                 bus.trigger("FILE_UPLOAD_ERROR", { upload });
-            });
+            }
+            // Error listener
+            xhr.addEventListener("error", (ev) => onError(ev.error));
             // Abort listener, considered as error
             xhr.addEventListener("abort", async () => {
                 delete uploads[upload.id];
--- a/addons/web/static/tests/core/file_upload.test.js
+++ b/addons/web/static/tests/core/file_upload.test.js
@@ -1,4 +1,4 @@
-import { expect, test } from "@odoo/hoot";
+import { expect, test, waitFor } from "@odoo/hoot";
 import {
     contains,
     getService,
@@ -61,13 +61,12 @@ test("upload renders new component(s)", async () => {
 
 test("upload end removes component", async () => {
     await mountWithCleanup(Parent);
-
+    onRpc("/test/", () => true);
     const fileUploadService = await getService("file_upload");
     fileUploadService.upload("/test/", []);
     await animationFrame();
-    fileUploadService.uploads[1].xhr.dispatchEvent(new Event("load"));
-    await animationFrame();
     expect(".file_upload").toHaveCount(0);
+    expect(".o_notification").toHaveCount(0);
 });
 
 test("upload error removes component", async () => {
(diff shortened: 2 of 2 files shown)


addons/web/static/src/core/file_upload/file_upload_service.js
```diff
@@ -15,7 +15,7 @@ export const fileUploadService = {
         return new window.XMLHttpRequest();
     },
 
-    start(env, { notificationService }) {
+    start(env, { notification: notificationService }) {
```
aab-odoo, line 18: lol

aab-odoo (commented):
I rely on @Julien00859 to validate the different faulty scenarios of controllers, but besides that, I'm ok with what you did.

addons/web/static/src/core/file_upload/file_upload_service.js
```diff
@@ -67,26 +67,74 @@ export const fileUploadService = {
             });
             // Load listener
             xhr.addEventListener("load", () => {
+                if (xhr.readyState !== 4) {
+                    return;
+                }
```
Julien00859: The `load` event is only ever fired when `readState` becomes `4` (`DONE`). The code is redundant with `onload` and can me removed.

```suggestion
```

addons/web/static/src/core/file_upload/file_upload_service.js
```diff
@@ -67,26 +67,74 @@ export const fileUploadService = {
...
-            // Error listener
-            xhr.addEventListener("error", async () => {
+
+            function handleResponse() {
+                const resp = xhr.responseText ?? xhr.response;
+                let error;
+                let errorMessage = "";
+                if (!(xhr.status >= 200 && xhr.status < 300)) {
+                    error = true;
+                }
```
Julien00859, line 87: 3xx should not be handled like 4xx and 5xx. 3xx are redirections, typically if you upload an attachment but that your session has since expired, you'll be redirected on `/web/login`. Shouldn't this code handles this `/web/login` redirect automatically somehow?

addons/web/static/src/core/file_upload/file_upload_service.js
```diff
@@ -67,26 +67,74 @@ export const fileUploadService = {
...
+                    if (typeof content === "string") {
+                        try {
+                            content = JSON.parse(content);
+                        } catch {
+                            try {
+                                content = new DOMParser().parseFromString(content, "text/html");
+                            } catch {
+                                /** pass */
+                            }
+                        }
```
Julien00859, line 99: Instead of this (kinda ugly) try/catch, cannot we access the `Content-Type` header and decide from that to use DOMParser or JSON.parse?

Julien00859:
```js
const contentType = response.getResponseHeader("Content-Type").split(";")[0].trim();
// Content-Type: application/json; charset=utf-8 -> "application/json"

switch (contentType) {
    case "application/json":  // the normal content-type
    case "text/json":  // another unofficial content-type that some devs (wrongly) use, let's be nice and support those
    case "application/json-rpc":  // another content-type we support when receiving jsonrpc, unlikely to appear in response but we never know
        content = JSON.parse(response.responseText);
        // content can be anything but will likely be an object with:
        // content["error"]["data"]["name"] and content["error"]["data"]["message"]  // jsonrpc
        // content["name"] and content["message"]  // json2
        // content["error"]  // only the error message, custom (bad) controllers
        break;
    case "text/html":
        content = new DOMParser().parseFromString(response.responseText, "text/html");
        // likely an error message produced by werkzeug, which is self-contained/self-explanatory
        break;
    default:
        content = response.responseText;
        // very bad controllers
}

File: addons/web/static/src/core/file_upload/file_upload_service.js [hunk 5]
```diff
@@ -67,26 +67,74 @@ export const fileUploadService = {
...
+                    }
+                    // Not sure what to do if the content is neither JSON nor HTML
+                    // Let's the call be successful then....
+                    if (error && content instanceof Document) {
+                        errorMessage = content.body.textContent;
+                    } else if (content instanceof Object) {
+                        if (content.error) {
+                            // https://www.jsonrpc.org/specification#error_object
+                            error = true;
+                            errorMessage = content.error.message ?? errorMessage;
```
Julien00859: Don't use `content.error.message`, it contains garbage:
```py
    def handle_error(self, exc):
        error = {
            'code': 0,  # we don't care of this code
            'message': "Odoo Server Error",  # <-------- here @LPE
            'data': serialize_exception(exc),
        }
        if isinstance(exc, NotFound):
            error['code'] = 404
            error['message'] = "404: Not Found"  # <-------- here @LPE
        elif isinstance(exc, SessionExpiredException):
            error['code'] = 100
            error['message'] = "Odoo Session Expired"  # <-------- here @LPE

        return self._response(error=error)

    def _response(self, result=None, error=None):
        response = {'jsonrpc': '2.0', 'id': self.request_id}
        if error is not None:
            response['error'] = error
        if result is not None:
            response['result'] = result

        return self.request.make_json_response(response)
```
The actually useful information are found inside `content.error.data` (which is the result of `serialiaze_exceptions`. Use `content.error.data.name` and `content.error.data.message`.
```py
def serialize_exception(exception, *, message=None, arguments=None):
    name = type(exception).__name__
    module = type(exception).__module__

    return {
        'name': f'{module}.{name}' if module else name,
        'message': exception_to_unicode(exception) if message is None else message,
        'arguments': exception.args if arguments is None else arguments,
        'context': getattr(exception, 'context', {}),
        'debug': ''.join(traceback.format_exception(exception)),
    }
```
"But the jsonrpc specs states that..." our jsonrpc implementation is not compliant with the spec. "But then we should change that" it was implemented like this long before you and me got hired in this company, and now it is too late to fix it.

Julien00859 (changes requested):
It kinda is futile to try and automatically handle those errors. There are several upload controllers in the backend and none of them care to serialize the errors like the other do.

There even are controllers that are not consistent with themselves!

```py
@http.route("/mail/attachment/upload", methods=["POST"], type="http", auth="public")
@add_guest_to_context
def mail_attachment_upload(self, ufile, thread_id, thread_model, is_pending=False, **kwargs):
    thread = self._get_thread_with_access_for_post(thread_model, thread_id, **kwargs)
    if not thread:
        raise NotFound()  # JUC, 404 response with content-type html
    try:
        ...
        res = {"data": {"store_data": store.get_result(), "attachment_id": attachment.id}}
        # JUC, 200 response with content-type json
    except AccessError:
        res = {"error": _("You are not allowed to upload an attachment here.")}
        # JUC, 200 response with content-type json
    return request.make_json_response(res)
```

If we want to improve the situation, we should do a spec on "how to do an upload controller, with proper error handling" and adapt all backend upload controllers to use that spec. This way the code in the frontend would become much saner.

addons/web/static/src/core/file_upload/file_upload_service.js
kebeclibre: indeed

addons/web/static/src/core/file_upload/file_upload_service.js
kebeclibre: I don't think so. It is not the job of that method to route the call (and the issue you are bringing forth is actually common to every single call that the client makes)

The rationale here is to say that anything different from success is failure.
In that stack, there is no ignoring, it is either success: upload did what we expected , or not.
So when when the request finishes:
- code 1XX : informational message: not sure what to do, seems odd to have those at the request's end
- code 2XX: successes
- 3XX redirects: not the job of the upload to route to login (not an upload route...) maybe to other routes though, but seems out of scope here because again, the JS doesn't do this in any part of the code

addons/web/static/src/core/file_upload/file_upload_service.js
kebeclibre: good idea but doesn't work.
We generally cannot trust the return of the controller.
In my example on         '/sale_pdf_quote_builder/quotation_document/upload',
If I do return what the code of the controllers expects (a json.dumped object)

I end up with
```js
xhr.getResponseHeader("Content-Type")
// 'text/html; charset=utf-8'
xhr.response
// `{"error": "It seems that we're not able to process this pdf inside a quotation. It is either encrypted, or encoded in a format we do not support."}`
```

addons/web/static/src/core/file_upload/file_upload_service.js
kebeclibre: héhéhéhé, right, forgot that. En route to write a test :)

addons/web/static/src/core/file_upload/file_upload_service.js
Julien00859: Indeed, because the controller returns a string, `return json.dumps(...)`, which is considered as a rendered html template. It should had `return request.make_json_response(...)`. Definitely another bug in the backend, and indeed we cannot trust the content-type so try/catch is the only way.

addons/web/static/src/core/file_upload/file_upload_service.js
Julien00859: 1xx are unlikely (will happen if you try to upload on `/websocket` :joy:) and could be considered as errors. 3xx should be passed as-is to the user who needs to handle them. Is it what the code does?

kebeclibre: With Snippet:
Screenshot from 2026-01-20 15-30-45

Redirects seem to work as expected.
Screenshot from 2026-01-20 15-30-22

But fun fact, redirecting a POST request doesn't seem to be possible
Screenshot from 2026-01-20 15-30-29

Julien00859 (approved):
Thanks for the test! It made the review much easier.

It assumes all json with a `"error"` key in the top-level object a errors, even if they are stored inside a `200 OK` http response. This is in-line with our very bad jsonrpc implementation and some other very bad controllers.

addons/web/static/src/core/file_upload/file_upload_service.js
Julien00859: > But fun fact, redirecting a POST request doesn't seem to be possible

That's because it uses status code 303 "See other" by default which is specified to use GET for the redirection. If you wanna keep the HTTP verb in between redirections you gotta use 307/308.

The use case for 303 is: you create (post) a new resource `POST /resources/new`, and you wanna redirect the user on the newly created resource `GET /resources/5`. The redirection is expected to occurs *after* the post. Here it happens *before*, you should had used 307/308 which are verb-preserving and mean "the URL you're trying to access actually is found on this other URL".

Julien00859: But I'm not interested in 307/308, they're gonna work out of the box

What I would like to see is what happen upon a redirection *triggered due to an expired session*.

1. Login
2. Remove your session-id cookie
3. Upload

kebeclibre: noiceeeee

then yes, it works as expected
Screenshot from 2026-01-20 15-44-32

kebeclibre: > But I'm not interested in 307/308, they're gonna work out of the box
> 
> What I would like to see is what happen upon a redirection _triggered due to an expired session_.
> 
>     1. Login
> 
>     2. Remove your session-id cookie
> 
>     3. Upload

Screenshot from 2026-01-20 16-07-07


seems to work fine ; doesn'it ?

kebeclibre: then clicking close actually redirects to /login

Julien00859: Great!

PR #244708, opened 2026-01-20
Master workorder kanban whe

Whenrow:

Current behavior before PR:

---

### DIFF
--- a/addons/hr/static/src/views/fields/many2many_avatar_employee_field/many2many_avatar_employee_field.js
+++ b/addons/hr/static/src/views/fields/many2many_avatar_employee_field/many2many_avatar_employee_field.js
@@ -1,3 +1,4 @@
+import { _t } from "@web/core/l10n/translation";
 import { registry } from "@web/core/registry";
 import {
     Many2ManyTagsAvatarUserField,
@@ -44,6 +45,10 @@ export class KanbanMany2ManyTagsAvatarEmployeeField extends EmployeeFieldRelatio
             (!this.env.isSmall && ["hr.employee", "hr.employee.public"].includes(this.relation)) ||
             super.displayAvatarCard(record)
         );
+    }
+
+    get placeholder() {
+        return _t("Search employee...");
     }
 }
 
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -479,11 +479,11 @@ class MrpProduction(models.Model):
         for production in self:
             production.duration = sum(production.workorder_ids.mapped('duration'))
 
-    @api.depends("workorder_ids.date_start", "workorder_ids.date_finished", "date_start")
+    @api.depends("workorder_ids.is_planned")
     def _compute_is_planned(self):
         for production in self:
             if production.workorder_ids:
-                production.is_planned = all(wo.date_start and wo.date_finished for wo in production.workorder_ids if wo.state != 'cancel')
+                production.is_planned = all(wo.is_planned for wo in production.workorder_ids if wo.state != 'cancel')
             else:
                 production.is_planned = False
 
@@ -1684,12 +1684,9 @@ class MrpProduction(models.Model):
             order.message_post(body=self.env._("The manufacturing order has been planned."), subtype_id=self.env.ref('mrp.mrp_mo_planned').id)
         return True
 
-    def _plan_workorders(self, replan=False):
+    def _plan_workorders(self):
         """ Plan all the production's workorders depending on the workcenters
         work schedule.
-
-        :param replan: If it is a replan, only ready and blocked workorder will be taken into account
-        :type replan: bool.
         """
         self.ensure_one()
 
@@ -1700,18 +1697,10 @@ class MrpProduction(models.Model):
         self._link_workorders_and_moves()
 
         # Plan workorders starting from final ones (those with no dependent workorders)
-        final_workorders = self.workorder_ids.filtered(lambda wo: not wo.needed_by_workorder_ids)
-        for workorder in final_workorders:
-            workorder._plan_workorder(replan)
-
-        workorders = self.workorder_ids.filtered(lambda w: w.state not in ['cancel'])
-        if not workorders:
-            return
-
-        self.with_context(force_date=True).write({
-            'date_start': min((workorder.leave_id.date_from for workorder in workorders if workorder.leave_id), default=None),
-            'date_finished': max((workorder.leave_id.date_to for workorder in workorders if workorder.leave_id), default=None),
-        })
+        self.workorder_ids.filtered(
+            lambda wo: not wo.is_planned
+            and not wo.needed_by_workorder_ids
+        )._plan_workorders(from_date=self.date_start)
 
     def button_unplan(self):
         self._unplan_workorders()
--- a/addons/mrp/models/mrp_workorder.py
+++ b/addons/mrp/models/mrp_workorder.py
@@ -15,7 +15,7 @@ from odoo.tools.intervals import Intervals
 class MrpWorkorder(models.Model):
     _name = 'mrp.workorder'
     _description = 'Work Order'
-    _order = 'sequence, leave_id, date_start, id'
+    _order = 'date_start, sequence, id'
 
     def _default_sequence(self):
         return self.operation_id.sequence or 100
@@ -51,6 +51,7 @@ class MrpWorkorder(models.Model):
     production_bom_id = fields.Many2one('mrp.bom', related='production_id.bom_id')
     qty_production = fields.Float('Original Production Quantity', readonly=True, related='production_id.product_qty')
     company_id = fields.Many2one(related='production_id.company_id')
+    priority = fields.Selection(related='production_id.priority')
     qty_producing = fields.Float(
         compute='_compute_qty_producing', inverse='_set_qty_producing',
         string='Currently Produced Quantity', digits='Product Unit')
@@ -130,7 +131,7 @@ class MrpWorkorder(models.Model):
     show_json_popover = fields.Boolean('Show Popover?', compute='_compute_json_popover')
     qty_reported_from_previous_wo = fields.Float('Carried Quantity', digits='Product Unit', copy=False,
         help="The quantity already produced awaiting allocation in the backorders chain.")
-    is_planned = fields.Boolean(related='production_id.is_planned')
+    is_planned = fields.Boolean(compute='_compute_is_planned')
     allow_workorder_dependencies = fields.Boolean(related='production_id.allow_workorder_dependencies')
     blocked_by_workorder_ids = fields.Many2many('mrp.workorder', relation="mrp_workorder_dependencies_rel",
                                      column1="workorder_id", column2="blocked_by_id", string="Blocked By",
@@ -140,6 +141,9 @@ class MrpWorkorder(models.Model):
                                      column1="blocked_by_id", column2="workorder_id", string="Blocks",
                                      domain="[('allow_workorder_dependencies', '=', True), ('id', '!=', id), ('production_id', '=', production_id)]",
                                      copy=False)
+    remaining_time = fields.Float('Remaining Working Time', compute='_compute_remaining_time',
+                                  help="The remaining time to finish this work order in hours.")
+    color = fields.Integer('Color', related="production_id.id")
 
     @api.depends('qty_ready')
     def _compute_state(self):
@@ -312,6 +316,14 @@ class MrpWorkorder(models.Model):
             if self.env.context.get('prefix_product'):
                 wo.display_name = f"{wo.product_id.name} - {wo.production_id.name} - {wo.name}"
 
+    @api.depends('duration_expected', 'duration', 'state')
+    def _compute_remaining_time(self):
+        for workorder in self:
+            if workorder.state in ('done', 'cancel'):
+                workorder.remaining_time = 0.0
+            else:
+                workorder.remaining_time = workorder.duration_expected - workorder.duration
+
     def unlink(self):
         # Removes references to workorder to avoid Validation Error
         (self.mapped('move_raw_ids') | self.mapped('move_finished_ids')).write({'workorder_id': False})
@@ -427,6 +439,18 @@ class MrpWorkorder(models.Model):
             else:
                 order.is_user_working = False
 
+    @api.depends('date_start', 'date_finished')
+    def _compute_is_planned(self):
+        for wo in self:
+            wo.is_planned = bool(wo.date_start and wo.date_finished)
+
+    def _search_is_planned(self, operator, value):
+        if (operator == '=' and value) or (operator == '!=' and not value):
+            return ['&', ('date_start', '!=', False), ('date_finished', '!=', False)]
+        if (operator == '=' and not value) or (operator == '!=' and value):
+            return ['|', ('date_start', '=', False), ('date_finished', '=', False)]
+        return NotImplemented
+
     @api.onchange('operation_id')
     def _onchange_operation_id(self):
         if self.operation_id:
@@ -438,14 +462,14 @@ class MrpWorkorder(models.Model):
         if self.date_start and self.workcenter_id:
             self.date_finished = self._calculate_date_finished()
 
-    def _calculate_date_finished(self, date_start=False, new_workcenter=False):
+    def _calculate_date_finished(self, date_start=False, new_workcenter=False, compute_leaves=False):
         workcenter = new_workcenter or self.workcenter_id
         if not workcenter.resource_calendar_id:
             duration_in_seconds = self.duration_expected * 60
             return (date_start or self.date_start) + timedelta(seconds=duration_in_seconds)
         return workcenter.resource_calendar_id.plan_hours(
             self.duration_expected / 60.0, date_start or self.date_start,
-            compute_leaves=True, domain=[('count_as', 'in', ['absence', 'working_time'])]
+            compute_leaves=compute_leaves, domain=[('count_as', 'in', ['absence', 'working_time'])]
         )
 
     @api.onchange('date_finished')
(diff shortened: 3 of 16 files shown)


addons/hr/static/src/views/fields/many2many_avatar_employee_field/many2many_avatar_employee_field.js
```diff
@@ -56,7 +61,6 @@ export const kanbanMany2ManyTagsAvatarEmployeeField = {
     ],
     extractProps: (fieldInfo, dynamicInfo) => ({
         ...kanbanMany2ManyTagsAvatarUserField.extractProps(fieldInfo, dynamicInfo),
-        relation: fieldInfo.options?.relation,
```
aab-odoo, line 59: are you sure about this? The EmployeeFieldRelationMixin takes a relation props.

aab-odoo: Besides that, ok for the placeholder ref

Whenrow: @aab-odoo Oupsy. I did this during the dev as the widgets failed and forgot about it. I just made a new commit to handle this correctly. This props is given to the avatar popover class which does not accept it

addons/web/static/src/views/fields/many2many_tags_avatar/many2many_tags_avatar_field.js
```diff
@@ -125,6 +125,7 @@ export class KanbanMany2ManyTagsAvatarField extends Many2ManyTagsAvatarField {
     get popoverProps() {
         const props = { ...this.props };
         delete props.isEditable;
+        delete props.relation;
```
clesgow, line 128: Not 100% sure this should be deleted here, as that props is only added through the mixin in `Many2ManyTagsAvatarEmployeeField`, but I guess we don't risk too much by removing a prop that may not exist at this point.

addons/mrp/models/mrp_production.py
```diff
@@ -479,11 +479,11 @@ def _compute_duration(self):
         for production in self:
             production.duration = sum(production.workorder_ids.mapped('duration'))
 
-    @api.depends("workorder_ids.date_start", "workorder_ids.date_finished", "date_start")
+    @api.depends("workorder_ids.is_planned")
     def _compute_is_planned(self):
         for production in self:
             if production.workorder_ids:
-                production.is_planned = all(wo.date_start and wo.date_finished for wo in production.workorder_ids if wo.state != 'cancel')
+                production.is_planned = all(wo.is_planned for wo in production.workorder_ids if wo.state != 'cancel')
```
clesgow, line 486: If all your `workorder_ids` are cancelled, then `all([]) == True`. Which seems contrary to the `else` condition.

addons/mrp/models/mrp_workorder.py
```diff
@@ -438,14 +462,14 @@ def _onchange_date_start(self):
...
-    def _calculate_date_finished(self, date_start=False, new_workcenter=False):
+    def _calculate_date_finished(self, date_start=False, new_workcenter=False, compute_leaves=False):
         workcenter = new_workcenter or self.workcenter_id
         if not workcenter.resource_calendar_id:
             duration_in_seconds = self.duration_expected * 60
             return (date_start or self.date_start) + timedelta(seconds=duration_in_seconds)
         return workcenter.resource_calendar_id.plan_hours(
             self.duration_expected / 60.0, date_start or self.date_start,
-            compute_leaves=True, domain=[('time_type', 'in', ['leave', 'other'])]
+            compute_leaves=compute_leaves, domain=[('time_type', 'in', ['leave', 'other'])],
```
clesgow: Seems like `compute_leaves` is always `False` now? Though I guess it's nice to have some extra granularity 🤔

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
-        leave = self.env['resource.calendar.leaves'].create({
-            'name': self.display_name,
-            'calendar_id': best_workcenter.resource_calendar_id.id,
-            'date_from': best_date_start,
-            'date_to': best_date_finished,
-            'resource_id': best_workcenter.resource_id.id,
-            'time_type': 'other'
+        :param from_date: An optional `datetime` object. If provided, The planning will start from
+            this date.
+        :type date: datetime.datetime or None
```
clesgow: Nitpick, but if `datetime or None`, then shouldn't be the default value be `None` rather than `False` ? 😄

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
+        :param alternative: An optional `boolean` parameter allowing the planning on alternative
+        workcenters than the one set on the workorder.
+        :type Boolean:
+        :raises UserError: If there are no available slot to plan the workorders
+        """
+        if not self:
+            return
+        # we need to keep the order of the workorder before removing the start date
+        workorders_to_plan = self.filtered(lambda wo: wo.state in ['ready', 'blocked'])
```
clesgow:
```suggestion
        # we need to keep the order of the workorder before removing the start date
        workorders_to_plan = self.filtered(lambda wo: wo.state in ['ready', 'blocked'])
        if not workorders_to_plan:
            return
```

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
+        wo_list = list(workorders_to_plan)
+        done_wo = set()
+        workorders_to_plan.leave_id.unlink()
+        workorders_to_plan.write({
+            'date_start': False,
+            'date_finished': False,
         })
-        vals['leave_id'] = leave.id
-        self.write(vals)
+        for wo in wo_list:
```
clesgow, line 606: Is there a point to using a `list` over the recordset itself? 🤔

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
+                date_start = wo.blocked_by_workorder_ids[-1].date_finished
+            # Consider workcenter and alternatives
+            workcenters = wo.workcenter_id | wo.workcenter_id.alternative_workcenter_ids
+            best_date_finished = datetime.max
+            vals = {}
+            if self.env.context.get('workcenter_to_plan_on'):
+                alternative = False
+                wo.workcenter_id = self.env['mrp.workcenter'].browse(self.env.context.get('workcenter_to_plan_on'))
+            for workcenter in workcenters:
+                if not alternative and workcenter != wo.workcenter_id:
```
clesgow, line 624: Nitpick, but I guess that if `alternative` is `False`, then it's not even useful to add the alternative workcenters in `workcenter` in the first place?

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
+                    continue
+                if not workcenter.resource_calendar_id:
+                    raise UserError(_('There is no defined calendar on workcenter %s.', workcenter.name))
```
clesgow:
```suggestion
                    raise UserError(self.env._('There is no defined calendar on workcenter %s.', workcenter.name))
```
I *think* we should use `self.env._()` over `_()` now?

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,84 @@ def _action_confirm(self):
...
+                from_date, to_date = workcenter._get_first_available_slot(date_start, duration_expected)
+                # If the workcenter is unavailable, try planning on the next one
+                if not from_date:
+                    continue
+                # Check if this workcenter is better than the previous ones
+                if to_date and to_date < best_date_finished:
+                    best_date_start = from_date
+                    best_date_finished = to_date
+                    best_workcenter = workcenter
+                    vals = {
```
clesgow, line 642: I guess we could create the right vals here and overwrite them if we find a better workcenter, rather than writing them partially here and again at the end, but it's up to you.

addons/mrp/models/mrp_workorder.py
```diff
@@ -748,20 +789,55 @@ def action_cancel(self):
...
+                continue
+            if not wo.date_start or not dec_workorders[idx - 1].date_start:
+                continue
+            # The workorder is planned later than the next one. To correct this, we set the same
+            # date_start than the next one and set a sequence number lower to stay in a sooner
+            # position
+            if wo.date_start >= dec_workorders[idx - 1].date_start:
+                vals = {
+                    'date_start': dec_workorders[idx - 1].date_start,
+                    'sequence': max_seq - idx,
```
clesgow, line 836: Shouldn't you take the `offset` into account here?

addons/mrp/models/mrp_workorder.py
```diff
@@ -748,20 +789,55 @@ def action_cancel(self):
...
+                }
+                if wo.workcenter_id:
+                    vals['date_finished'] = datetime.max
```
clesgow, line 839: Err... Is this really correct? 🤔

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
```diff
@@ -66,11 +69,18 @@ export class MOListViewDropdown extends Component {
...
 };
 
-registry.category("view_widgets").add("mo_view_list_dropdown", moListViewDropdown);
+registry.category("fields").add("mo_view_list_dropdown", {
+    ...badgeField,
+    supportedOptions: [
+        {
+            name: "display",
+            type: "boolean"
+        }
```
clesgow:
```suggestion
            type: "string",
        },
```

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
```diff
@@ -44,13 +48,12 @@ export class MOListViewDropdown extends Component {
         if (!selectedWorkorders || selectedWorkorders.length == 0) {
             selectedWorkorders = [this.props.record];
         }
-        let ids = selectedWorkorders.filter((wo) => !([state, 'done'].includes(wo.data.state) || wo.data.production_state == 'done')).map((wo) => wo.resId)  
+        let ids = selectedWorkorders.filter((wo) => !([state, 'done'].includes(wo.data.state) || wo.data.production_state == 'done')).map((wo) => wo.resId)
```
clesgow:
```suggestion
        const ids = selectedWorkorders
            .filter(
                (wo) =>
                    !([state, "done"].includes(wo.data.state) || wo.data.production_state == "done")
            )
            .map((wo) => wo.resId);
```
*cough* [linter](https://github.com/odoo/odoo/wiki/Javascript-coding-guidelines#tooling) 😄

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
```diff
@@ -35,7 +39,7 @@ export class MOListViewDropdown extends Component {
     }
 
     get statusColor() {
-        const state = this.workorderState.state;
+        const state = this.props.record.data.state;
```
clesgow, line 42: Do we still need `this.workorderState` then?

addons/mrp/views/mrp_workorder_views.xml
```diff
@@ -87,20 +87,49 @@
                   invisible="production_state in ('draft', 'done', 'cancel') or working_state == 'blocked' or not is_user_working"/>
                 <button name="button_finish" type="object" title="Done" class="btn-success px-2" icon="fa-check"
                   invisible="production_state in ('draft', 'done', 'cancel') or working_state == 'blocked' or not is_user_working"/>
-                <widget name="mo_view_list_dropdown"/>
-                <field name="state" widget="badge" decoration-info="state == 'progress'" decoration-success="state == 'done'" decoration-danger="state == 'cancel'" decoration-warning="state == 'blocked'" decoration-muted="state == 'ready'"
+                <field name="state" widget="mo_view_list_dropdown"
```
clesgow, line 90: Note that now it doesn't display anything if the workorder is done, which is not a big deal but weird.

addons/mrp/views/mrp_workorder_views.xml
```diff
@@ -87,20 +87,49 @@
...
+                <attribute name="column_invisible">1</attribute>
+            </button>
+            <button name="button_pending" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </button>
+            <button name="button_finish" position="attributes">
+                <attribute name="column_invisible">1</attribute>
+            </button>
+            <field name="state" position="attributes">
+                <attribute name="widget">badge</attribute>
```
clesgow: If we set the column as invisible, do we even care about the widget? 😄

addons/mrp/views/mrp_workorder_views.xml
```diff
@@ -328,44 +357,58 @@
...
+                        <div class="o_kanban_workorder_production_name">
+                            <field name="priority" widget="priority" class="pe-2"/>
+                            <field name="production_id"/>
+                            <field name="name" class="ps-4"/>
+                        </div>
+                        <t t-set="start_datetime" t-value="luxon.DateTime.fromISO(record.date_start.raw_value)"/>
+                        <div class="o_kanban_workorder_date"
+                            t-att-class="{'o_wo_orange': start_datetime.hasSame(luxon.DateTime.now(), 'day'),
+                                 'o_wo_red': start_datetime.toISODate() &lt; luxon.DateTime.now().toISODate(),
+                                 '': start_datetime.toISODate() &lt; luxon.DateTime.now().toISODate(),
```
clesgow, line 398:
```suggestion
                                 'o_wo_red': start_datetime &lt; luxon.DateTime.now(),
                                 '': start_datetime &lt; luxon.DateTime.now(),
```
Not sure you need those `toISODate()` here, as `now()` and `fromISO` should both create a date in the same locale.

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
```diff
@@ -1,18 +1,22 @@
...
     static template = "mrp.MOViewListDropdown";
     static components = {
         Dropdown,
         DropdownItem,
     };
-    static props = { ...standardWidgetProps };
+
+    static props = {
+        ...standardFieldProps,
+        display: { type: String },
```
clesgow:
```suggestion
        display: { type: String, validate: (val) => ["bubble", "badge"].includes(val) },
```
If you want to be extra-cautious

addons/mrp/views/mrp_workorder_views.xml
```diff
@@ -328,44 +357,58 @@
...
-                                <field t-if="!record.date_start.raw_value" name="production_date"/>
-                            </div>
-                            <div class="h2 ms-2">
-                                <span t-attf-class="badge rounded-pill #{['progress'].indexOf(record.state.raw_value) > -1 ? 'text-bg-warning' : ['ready', 'blocked'].indexOf(record.state.raw_value) > -1 ? 'text-bg-primary' : ['done'].indexOf(record.state.raw_value) > -1 ? 'text-bg-success' : 'text-bg-danger'}">
-                                    <field name="state"/>
-                                </span>
+                                <field name="product_id"/>
+                                <div class="ms-auto text-nowrap">
+                                    <field name="qty_production" class="p-2"/>
+                                    <field name="uom_id"/>
```
clesgow:
```suggestion
                                    <field name="uom_id" groups="uom.group_uom"/>
```

addons/mrp/models/mrp_workorder.py
Whenrow: the workorders are ordered by date_start. And I erase them before the loop to free the planning. I put the wo in a list to keep the panning orders

addons/mrp/models/mrp_workorder.py
Whenrow: I don't think so as I change the complete order only on the selected records with the dates. I don't care about the sequence of workorders not selected by the resequence.

addons/mrp/models/mrp_workorder.py
Whenrow: the workorder `write()` needs a value for `date_finished` to recompute the duration expected

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
Whenrow: haha *very* secured

addons/mrp/static/src/components/wo_list_view_dropdown/wo_list_view_dropdown.js
Whenrow: more LoC ➡ more bugs :nerd_face:

addons/mrp/models/mrp_production.py
amoyaux: It seems okay to tell that a MO is planned if he doesn't have any workorders or if they are all cancel. I'm not sure, in which use case we need it without MO

addons/mrp/models/mrp_workorder.py
```diff
@@ -312,6 +316,14 @@ def _compute_display_name(self):
             if self.env.context.get('prefix_product'):
                 wo.display_name = f"{wo.product_id.name} - {wo.production_id.name} - {wo.name}"
 
+    @api.depends('duration_expected', 'duration', 'state')
+    def _compute_remaining_time(self):
+        for workorder in self:
+            if workorder.state in ('done', 'cancel'):
+                workorder.remaining_time = 0.0
+            else:
+                workorder.remaining_time = workorder.duration_expected - workorder.duration
```
amoyaux, line 325: max 0 to avoid ngative?

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,86 @@ def _action_confirm(self):
...
-        date_start = max(self.production_id.date_start, datetime.now())
-        for workorder in self.blocked_by_workorder_ids:
-            workorder._plan_workorder(replan_predecessors)
-            if workorder.date_finished and workorder.date_finished > date_start:
-                date_start = workorder.date_finished
-        # Plan only suitable workorders
-        if not replan:
-            return
-        self._internal_plan_workorder(date_start, self.workcenter_id | self.workcenter_id.alternative_workcenter_ids)
+    def _plan_workorders(self, from_date=False, alternative=True):
```
amoyaux, line 584: rename in _action_plan

addons/mrp/models/mrp_workorder.py
```diff
@@ -557,67 +581,86 @@ def _action_confirm(self):
...
+        workorders_to_plan = self.filtered(lambda wo: wo.state in ['ready', 'blocked'])
+        if not workorders_to_plan:
+            return
+        # we need to keep the order of the workorder before removing the start date
+        wo_list = list(workorders_to_plan)
+        done_wo = set()
+        workorders_to_plan.leave_id.unlink()
+        workorders_to_plan.write({
+            'date_start': False,
+            'date_finished': False,
```
amoyaux, line 604: action_unplan

amoyaux:
robodoo r+ rebase-ff We merge it in order to make it available for customer in 19.2. We will apply all the comments in another PR. Feel free to add more comments if needed

PR #231948, opened 2025-10-16
[IMP] base: Currency Rates Fetching

abmn-odoo:
Currency rates are valid for the next day and should not change during the day.

Use only rates strictly earlier than the requested date.

task-5173684

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #218612, opened 2025-07-11
[ADD] purchase_alternative*: separate alternative from purchase_requisition

dhha-odoo:
*: purchase_requisition

-----------------------------------------------------------------------------------------

We have separated the `Purchase Alternative` functionality from
the `purchase_requisition` module and created a new independent module
called `purchase_alternative`. A new Boolean field, Purchase Alternative,
has been added in Purchase settings to manage this feature.

Previously, the `Purchase Alternative` option was hard to find; 
It only appeared after enabling Purchase Agreements, saving, 
and refreshing the Purchase settings view. This was very confusing 
and difficult to find for the user

Now, users can directly enable the `Purchase Alternative` feature 
from the Purchase settings without relying on the `purchase_requisition` module.
This makes the feature easier to access and more modular.

taskId :- 4737059

PR #244711, opened 2026-01-20
[FIX] purchase_alternative: fix currency rate in alternative PO test

dhha-odoo:
Issue before this commit:
=========================
The test_12_alternative_po_line_different_currency test was moved to the
purchase_alternative module in this [PR](https://github.com/odoo/odoo/pull/218612).

At the same time, the currency rate logic was updated in another [PR](https://github.com/odoo/odoo/pull/231948). 
During conflict resolution, the test was mistakenly reverted to the old behaviour, causing the test to break.

As a result, the alternative PO price comparison became incorrect.

Steps to Reproduce:
=========================
- Install the purchase_alternative module.
- Run the test: test_12_alternative_po_line_different_currency test case.
- The test fails at: self.assertEqual(best_price_ids[0], po_alt.order_line.id)

Cause of the issue:
=========================
Recent currency rate changes introduced in [PR](https://github.com/odoo/odoo/pull/231948),
That PR ensures that Currency rates are valid only starting the next day.

During conflict resolution, these changes were removed from the test, 
causing invalid currency rates, leading to incorrect price comparisons.

With This Commit:
=========================
Ensure the test runs correctly by aligning it with the updated currency rate logic.

runbot-build-error: [237989](https://runbot.odoo.com/odoo/runbot.build.error/237989)

### DIFF
--- a/addons/purchase_alternative/tests/test_purchase_alternative.py
+++ b/addons/purchase_alternative/tests/test_purchase_alternative.py
@@ -234,11 +234,11 @@ class TestPurchaseAlternative(TestPurchaseAlternativeCommon):
 
         # 1 USD = 0.5 EUR
         self.env['res.currency.rate'].create([{
-            'name': fields.Datetime.today(),
+            'name': fields.Date.today() - timedelta(days=1),
             'currency_id': self.env.ref('base.USD').id,
             'rate': 1,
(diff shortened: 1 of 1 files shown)


ticodoo:
@robodoo r+

PR #234132, opened 2025-11-03
[IMP] tests, web: start js test per module

Xavier-Do:
The web module tests became a bottleneck regarding testing and splitting. 
Moreover the js tests are run only in web, meaning that to run all tests of a module, the correct way would be to make
/module,/web.tests_unit_descktop[web],/web.tests_unit_mobile[web], ...

This pr add a new way to declare a test so that it will run in all modules, but only for the part of the module that are being tested. So /module will run all test of the module, including js tests.  

This pr combines 
#233788
#232264
#232268

Note that the current configuration format is a prototype and subject to breaking changes in a near future.

PR #244715, opened 2026-01-20
[FIX] web: increase test_unit_desktop timeout

aab-odoo:
This commit increases the timeout of the main (desktop) hoot test suite.

Since [1], the suite is split by sub-builds, such that each sub-build only runs the tests defined in the addons that are tested by that sub-build. So the suite never timeouts anymore on regular builds.

However, in nightly, we still run the whole suite "old school", i.e. all tests in the same run. In that case, the suite sometimes (in 19.1) or often (in master) exceeds the previous timeout (50m). With 1 hour, we hope that it will be enough.

Note that the number of tests keeps increasing, hence the need to increase the timeout.

[1] https://github.com/odoo/odoo/pull/234132

Current behavior before PR:

---

### DIFF
--- a/addons/web/tests/test_js.py
+++ b/addons/web/tests/test_js.py
@@ -196,7 +196,7 @@ class WebSuite(HootCommon, odoo.tests.CrossModule):
         filters = self._get_hoot_filters(addons_from_asset_bundle, modules)
         if not filters:
             return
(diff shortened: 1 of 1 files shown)


Xavier-Do:
robodoo r+

PR #244718, opened 2026-01-20
[FIX] hr_recruitment: synchronize employees count in job's smart button

ksar-odoo:
### Steps to reproduce:
- Create a job for multiple companies.
- Go to its configuration and click the Employees smart button.
- Click the smart button and compare the employee count.

### Before/Root cause:
- Currently selected companies were not considered. Instead, all companies were taken into account during the computation.

### After:
- When you click on employee smart button only currently selected companies will be considered.

### Fix:
- Added a domain filter to consider only the currently selected companies.

task-5785245

---


romaincarlier4 (approved):
LGTM

PR #244719, opened 2026-01-20
[FIX] point_of_sale: use config base_url for customer display URL

paan-odoo:
Steps:

- Install point_of_sale.
- Open a POS session on a mobile device.
- Click on the customer display menu.

Issue:

- The QR code for the customer display contains an invalid URL: undefined/pos_customer_display/${config_id}.

Issue:

- The base URL was being accessed from the POS session, where it is not available.

Fix:

- Retrieve the base URL from the POS configuration instead of the session.

task-5792308


davidmonnom (commented):
Can you write tests ?

PR #244729, opened 2026-01-20
[FIX] account: hide Auto-Complete field when Vendor is not selected

maan-odoo:
Currently, an exception is generated when the user tries to select
`Auto-Complete` without selecting `Vendor`.

Steps to produce an error
- Go to Invoicing > Vendors > Bills > Click New
- Try selecting `Auto-Complete` >>> Error occurs

Error `AssertionError: Invalid falsy real id`

This error occurs because, after the recent refactoring in commit [1], falsy
ids are no longer allowed in domains. In the `Account Entry` view
`[partner_id]` (see [2]) is passed as part of the domain. When the user has
not selected a vendor (partner_id) it evaluates to False, which results in
[False] being passed and triggers the error introduced by commit [1].

This commit will fix the above issue by hiding the `Auto-Complete`
field when `Vendor` is not selected.

[2]: https://github.com/odoo/odoo/blob/7b288aa436f669d8c8e2ddac5277bb98321a390e/addons/account/views/account_move_views.xml#L1067

Sentry-7201563878

### DIFF
--- a/addons/account/views/account_move_views.xml
+++ b/addons/account/views/account_move_views.xml
@@ -1061,9 +1061,9 @@
                                 <field name="ref" default_focus="1" invisible="move_type != 'entry'"/>
                                 <field name="tax_cash_basis_origin_move_id" invisible="not tax_cash_basis_origin_move_id"/>
                                 <label name="invoice_vendor_bill_id_label" for="invoice_vendor_bill_id" string="Auto-Complete" class="oe_edit_only"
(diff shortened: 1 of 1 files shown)


maan-odoo:
Hello @odoo/sentry_reviewers 

Could I get a review here?
Thanks

Megaaaaaa (approved):
LGTM, thank you for the fix :+1:

robodoo r+

PR #240289, opened 2025-12-16
[FIX] pos_loyalty: fix discount value specific

robinengels:
When using a promotion program applied only on specific products with
tax included prices, the discount value was incorrect.

Steps to reproduce:
-------------------
* Create a product with a price of 100$ and tax of 10% included
* Create a promotion program with:
  - Apply on: Specific Products (the product created above)
  - Discount Type: $ Fixed Discount
  - Discount Value: 10$
* Open PoS and add the product to the order
* Apply the promotion code
> Observation: The value of the discount is 9.09$ instead of 10$

The same steps also applies to the discount reward applied on the whole
order.

Why the fix:
------------
When computing the discountable amount for the promotion it would take
the amount excluding tax instead of the base price. The base price
should be the unit price of the product multiplied by the quantity. As
it was before here :
https://github.com/odoo/odoo/blob/8568baa35c9caa4c8cd3b1f5d737e1ecbec76434/addons/point_of_sale/static/src/app/models/pos_order_line.js#L470-L474
https://github.com/odoo/odoo/blob/8568baa35c9caa4c8cd3b1f5d737e1ecbec76434/addons/pos_loyalty/static/src/app/models/pos_order.js#L1112
To do so we modified the accounting helpers to provide this value
instead of computing it manually.

opw-[5260067](https://www.odoo.com/web#id=5260067&view_type=form&model=project.task)
opw-[5242194](https://www.odoo.com/web#id=5242194&view_type=form&model=project.task)

PR #244734, opened 2026-01-20
[FIX] pos_loyalty: fix cheapest tax incl discount

robinengels:
When applyin a discount by percentage on the cheapest product if the product was using a tax included price, the discount was wrongly calculated on the tax excluded price.

Steps to reproduce:
-------------------
* Create a tax of 15% included in price
* Create a product with a price of 10€ and assign the tax created before
* Create a loyalty program with a reward of 100% discount on the cheapest product
* In POS, add the product to the order
> Observation: The discount applied is of 8.7€ instead of 10€

Why the fix:
------------
We make a similar fix to this one : https://github.com/odoo/odoo/pull/240289

opw-5260067


addons/point_of_sale/static/src/app/models/accounting/pos_order_line_accounting.js
```diff
@@ -150,6 +155,10 @@ export class PosOrderlineAccounting extends Base {
         return this.qty * this.price_unit * (1 - this.getDiscount() / 100);
     }
 
+    get basePriceUnit() {
+        return this.price_unit * (1 - this.getDiscount() / 100);
+    }
```
stva-odoo, line 160:
```suggestion
   get basePrice() {
        return this.qty * this.basePriceUnit;
    }
    
    get basePriceUnit() {
        return this.price_unit * (1 - this.getDiscount() / 100);
    }
```

addons/point_of_sale/static/src/app/models/accounting/pos_order_line_accounting.js
```diff
@@ -147,7 +152,11 @@ export class PosOrderlineAccounting extends Base {
     }
 
     get basePrice() {
-        return this.qty * this.price_unit * (1 - this.getDiscount() / 100);
+        return this.qty * this.basePriceUnit;
+    }
+
+    get basePriceUnit() {
+        return this.price_unit * (1 - this.getDiscount() / 100);
```
davidmonnom, line 159: I think you've got the names mixed up.

basePrice shouldn't remain 

```js
  get basePrice() {
        return this.qty * this.price_unit * (1 - this.getDiscount() / 100);
    }
```

?

robinengels: No I think it's correct, I jut extracted basePriceUnit so that it can be used when computing the combo price. @davidmonnom

davidmonnom (approved):
@robodoo r+

PR #244735, opened 2026-01-20
[FIX] pos_loyalty: correct gift card tax handling

Nokoz:
We had a bug when redeeming a gift card when the default taxe of the company was tax_included.

Steps to reproduce:
-------------------
* Set default 15% tax to be Tax Included
* In PoS, sell a gift card
* Use that gift card in a new order

> Observation:
Deducted amount is 43.48, that's 50 without taxes

Why the fix:
------------
We now compute the gift card reward line from a tax-aware amount and choose the unit price based on whether the discount product’s tax is price-included, ensuring we deduct the full intended value while displaying the correct tax. This prevents under-deduction (untaxed base only) and avoids re-adding tax on top, keeping totals and tax lines consistent.

opw-5441106

### DIFF
--- a/addons/pos_loyalty/static/src/app/models/pos_order.js
+++ b/addons/pos_loyalty/static/src/app/models/pos_order.js
@@ -1186,11 +1186,16 @@ patch(PosOrder.prototype, {
                     special_mode: "total_included",
                 },
             });
+            const priceUnit =
+                price.total_excluded +
+                price.taxes_data
+                    .filter((taxData) => taxData.tax.price_include)
+                    .reduce((sum, taxData) => sum + taxData.tax_amount, 0);
 
             return [
                 {
                     product_id: discountProduct,
-                    price_unit: price.total_excluded,
+                    price_unit: priceUnit,
                     qty: 1,
                     reward_id: reward,
                     is_reward_line: true,
--- a/addons/pos_loyalty/static/tests/unit/models/pos_order.test.js
+++ b/addons/pos_loyalty/static/tests/unit/models/pos_order.test.js
@@ -415,4 +415,25 @@ describe("pos.order - loyalty", () => {
         const rewardLine = order._get_reward_lines()[0];
         expect(rewardLine.prices.total_included).toBe(-10);
     });
+
+    test("_getRewardLineValuesDiscount - gift card payment uses full amount tax-free", async () => {
+        const store = await setupPosEnv();
+        const models = store.models;
+        const order = store.addNewOrder();
+
+        const product = models["product.product"].get(1);
+        await addProductLineToOrder(store, order, { productId: product.id, price_unit: 50 });
+
+        const reward = models["loyalty.reward"].get(4);
+        const discountProduct = models["product.product"].get(200);
+        discountProduct.taxes_id = [];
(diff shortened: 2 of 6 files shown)


addons/pos_loyalty/static/src/app/models/pos_order.js
```diff
@@ -1184,11 +1184,12 @@ patch(PosOrder.prototype, {
                     special_mode: "total_included",
                 },
             });
-
+            const hasPriceIncludedTax = discountProduct.taxes_id[0]?.price_include;
```
robinengels: This is working only if all taxes on the product are included or excluded. If you make a mix of tax_include and tax_exclude it will not work.
I think when you compute the price you should take the base price + the tax amount of the tax included in price only. This way when the product is added the excluded tax will be added and the total amount would be correct.
Another way could be to force all the taxes to be included in the price but I don't think this is something that makes a lot of sense

addons/pos_loyalty/static/tests/unit/models/pos_order.test.js
```diff
@@ -338,4 +338,46 @@ describe("pos.order - loyalty", () => {
...
+        const models = store.models;
+        const order = store.addNewOrder();
+
+        const product = models["product.product"].get(1);
+        product.taxes_id = [];
+        product.taxes_id_after_fiscal_position = [];
+        await addProductLineToOrder(store, order, { productId: product.id, price_unit: 50 });
+
+        const reward = models["loyalty.reward"].get(4);
+        const discountTemplate = models["product.template"].create({
```
robinengels: You can probably reuse the giftcard program defined her ehttps://github.com/odoo-dev/odoo/blob/d2c5072e949e4dded5ca809508ec4de3c99966bf/addons/pos_loyalty/static/tests/unit/data/loyalty_program.data.js#L68

And change the discount product on the program

addons/pos_loyalty/static/src/app/models/pos_order.js
Nokoz: True.. We can keep it in mind for the records or later imp.
I think this is a bit overkill right now.

stva-odoo: IMO the `price_unit` should depend on the config. 
Is something like this could work ? 

```
const config = this.models["pos.config"].getFirst();
const priceUnit = config.iface_tax_included === "total" ? price.total_included : price.total_excluded;
```

Nokoz: This solution breaks some tests and from what I could see `iface_tax_included` is always "total" no matter how I change the config. With taxes included, excluded and mixed between pos default and gift card reward product

addons/pos_loyalty/static/src/app/models/pos_order.js
davidmonnom: I know there is a problem with gift cards, but this fix is not good.

Gift cards should be sold tax-free and then used to purchase other products with tax. It's a kind of eWallet.

To do it right, the gift card line should not even have tax.

Nokoz: @davidmonnom Does this mean no fix?
Or what should be looked at to make it good?

PR #244739, opened 2026-01-20
[FIX] pos_sale: create picking for imported sale orders

pedrambiria:
Before this commit, when validating a POS order linked to a sale order, if the Inventory Management was configured to create pickings at the end of the session, no picking was created for the order, and the sale order stock was not updated. This commit fixes this issue by ensuring that the picking is created for the imported sale orders in real time.

opw-5423113

---


adgu-odoo (commented):
robodoo delegate+

PR #244744, opened 2026-01-20
[IMP] hr_holiday: use Luxon method

phenix-factory:
This commit uses the Luxon method `hasSame` to compare `DateTime` years.


tsm-odoo (approved):
@robodoo r+

PR #244753, opened 2026-01-20
[IMP] hr_holidays: show future out of office

phenix-factory:
This improves the display of upcoming out of office periods and correctly
handles afternoon absences.

Task-5119038
Task-5129796

PR enterprise: https://github.com/odoo/enterprise/pull/125668


addons/hr_holidays/models/hr_employee_public.py
```diff
@@ -72,3 +72,4 @@ def action_open_time_off_calendar(self):
     def _store_avatar_card_fields(self, res: Store.FieldList):
         super()._store_avatar_card_fields(res)
         res.attr("leave_date_to")
+        res.attr("leave_date_form")
```
Atovange: typo :)

addons/hr_holidays/models/hr_employee.py
```diff
@@ -146,21 +152,23 @@ def _compute_leave_status(self):
         # Used SUPERUSER_ID to forcefully get status of other user's leave, to bypass record rule
         holidays = self.env['hr.leave'].sudo().search([
             ('employee_id', 'in', self.ids),
-            ('date_from', '<=', fields.Datetime.now()),
+            ('date_from', '<=', fields.Datetime.now() + timedelta(days=1.0)),
```
alexkuhn: The notification is the business day before, so not necessarily the day before: it could be the Friday before a week-end and / or before an holiday.

addons/hr_holidays/static/tests/out_of_office_text.test.js
```diff
@@ -26,7 +26,9 @@ test("Show 'back on' in avatar card", async () => {
     const employee = pyEnv["hr.employee"].create({
         user_id: serverState.userId,
         work_contact_id: serverState.partnerId,
-        leave_date_to: DateTime.now().plus({ days: 3 }).toISODate(),
+        leave_date_from: DateTime.now().plus({ days: 2 }).toSQL(),
+        leave_date_to: DateTime.now().plus({ days: 3 }).toSQLDate(),
```
alexkuhn: Prefer official formatters `serializeDateTime()`

alexkuhn (commented):
Haven't test the feature, but diff is quite straight-forward.

addons/hr_holidays/models/hr_employee.py
phenix-factory: Id I'm not mistaken, it's not important since we only handle "tomorrow", if the time off is planned more than a day later, it shows the exact date. This is just to grab a wider range of leave to handle "future" leaves.

addons/hr_holidays/static/tests/out_of_office_text.test.js
phenix-factory: Indeed, thats better

addons/hr_holidays/models/hr_employee.py
alexkuhn: While I get that with specs only concerned with "Tomorrow" this works, but I'm pretty sure the benefit of this task is too low if we don't take business days into account.

Taking a day or half day off here and there, this is rarely planned. And also this is only "half day" off.

Usually moments that matter to be warned of a future leave is when the leave is long, usually a week or more. The majority of 1 week+ leaves start on Monday, therefore the case of week-end is quite frequent. Its omission would reduce the benefit of this feature too much.

phenix-factory: I absolutely don't get what you are trying to explain. If you take Friday and it's Thursday, it should say "off tomorrow". What's the fuss around "business" day?

alexkuhn: If I take time off on Monday and we are on Friday, other people are expected to see that this is the last day I'm available and it should warn them of incoming leave on Friday.

Friday is 3 days before Monday but in terms of business days this is "yesterday". As I said, this is important because this scenario is extremely frequent and not handling it would mean that showing of future leaves won't work most of time which is not ok.

phenix-factory: I see, but I don't see a way to do that, or even to account for all the edge cases that will appear. For example, every country has a different "business day" calendar.
And some people work on weekends…

It's also not part of the spec, if it were, I would have avoided this nightmare.

alexkuhn: That's the crucial part of this module. The whole complexity around `hr_holidays` is dealing with business days. We have to work around it, that's how business works.

All info should be available. The work hours are modeled (see `_get_duration_based_work_hours_on_date` and related methods), the leaves of the person are encoded, so it's all about mixing all the available ingredients to make the meal.

phenix-factory: I don't see how this method helps. It computes some sort of "duration," most probably in hours, based on attendance, which, according to the naming, is probably when someone is working.
That is nice, but doesn't help much.

In fact, I don't even see how you would be able to compute something like that. Each employee might have a different business calendar, but having the number of hours they work does help much.

Maybe if you can have a personalized calendar and find a way to tell Luxon about it… Somehow?

alexkuhn: I refer to this method not to say this method needs to be used, but more to show there's logic around "work hours" that the computed fields in `hr_holidays` use, and this is an important consideration here.

My understanding of `hr_holidays` modeling in the past, and I suppose it didn't change much, is that each employee has a job and the job maps to work hours in a calendar. That means for a given user -> employee, we can retrieve the encoded leaves and the job work hours and deduce the time range of leave.

So we can deduce last active business day of a user before a leave.

You point out a good remark about other people that have different business hours and the last active business day of the person might take place during a "leave" of someone. While this case is interesting and could be nice to resolve, I think it's ok to ignore for now.

(I don't think this is that much more complex, but that's an additional layer of complexity: we need either personalized store data or we need to deduce it client-side / JS, by knowing the "yesterday" business day of the person and client-side code has the "next active business day" of the current user, and from both data we can conditionally show on UI for the current day.)

phenix-factory: Calculating the “next working day” differs for each employee and requires multiple database calls (see _get_first_work_interval). This results in many extra queries that do not scale well.

My current solution is to store this information in the database to mitigate the issue. Since leaves are not encoded frequently, this should be an acceptable solution.

phenix-factory: My bad, this is not storable since it uses "now".

addons/hr_holidays/models/hr_employee.py
```diff
@@ -45,10 +51,30 @@ class HrEmployee(models.Model):
...
+    next_working_day_on_leave = fields.Date(compute="_compute_next_working_day_on_leave")
     departure_do_cancel_time_off_requests = fields.Boolean(related='version_id.departure_do_cancel_time_off_requests',
         inherited=True, readonly=False, groups="hr.group_hr_user")
 
+    def _compute_next_working_day_on_leave(self):
+        first_leave_by_employee = dict(self.env['hr.leave'].sudo()._read_group(
+            [
+                ('employee_id', 'in', self.ids),
+                ('date_from', '>=', fields.Datetime.now()),
+                ('date_to', '<=', fields.Datetime.now() + timedelta(days=7)),
```
ThanhDodeurOdoo: shouldn't it also be `date_from`?

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
```diff
@@ -1,10 +1,52 @@
...
+import { _t } from "@web/core/l10n/translation";
+
+const { DateTime } = luxon;
 
 patch(HrEmployee.prototype, {
     setup() {
         super.setup();
         this.leave_date_to = fields.Date();
+        this.leave_date_from = fields.Datetime();
+        /** @type {'am'|pm} */
```
ThanhDodeurOdoo:
```suggestion
        /** @type {'am'|'pm'} */
```

addons/hr_holidays/static/src/discuss/core/public_web/discuss_app/sidebar/channel_patch.xml
```diff
@@ -2,7 +2,7 @@
 <templates xml:space="preserve">
     <t t-inherit="mail.DiscussSidebarChannel.main" t-inherit-mode="extension">
         <xpath expr="//*[hasclass('o-mail-DiscussSidebarChannel-itemName')]" position="inside">
-            <div t-if="channel.correspondent?.partner_id?.outOfOfficeDateEndText" class="text-warning smaller" t-out="channel.correspondent.persona.outOfOfficeDateEndText"/>
+            <div t-if="channel.correspondent?.partner_id?.employee_id?.outOfOfficeDateEndText" class="text-warning smaller" t-out="channel.correspondent.persona.employee_id.outOfOfficeDateEndText"/>
```
ThanhDodeurOdoo: could be more consistent if it was partner_id in both parts

addons/hr_holidays/models/hr_employee.py
```diff
@@ -45,10 +51,30 @@ class HrEmployee(models.Model):
...
+            ],
+            ("employee_id",),
+            ["date_from:min"],
+        ))
+        for employee in self:
+            next_working_day_on_leave = None
+            if next_leave_day := first_leave_by_employee.get(employee):
+                # get the first day that the employee should be working
+                next_working_day = employee._get_first_working_interval(fields.Datetime.now() + timedelta(days=1), compute_leaves=False)
+                if next_leave_day.date() == next_working_day.date():
```
ThanhDodeurOdoo: `next_working_day` can technically be None if we don't find a working day within the 720 days limit (not sure it can happen in practice, maybe for employees that are no longer employed, I have some of them in my odoo DM list)

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
```diff
@@ -1,10 +1,52 @@
...
+            return _t("Out of office starting at %(time)s", { time });
+        }
+        if (
+            this.leave_date_from &&
+            DateTime.now().plus({ day: 1 }).hasSame(this.leave_date_from, "day")
+        ) {
+            // A negative diff means the leave period is ongoing.
+            return _t("Out of office tomorrow");
+        }
+        const date = this.leave_date_to.toLocaleString(dateFormat(this.leave_date_to));
```
ThanhDodeurOdoo: in hr employee, `leave_data[holiday.employee_id.id]['leave_date_to'] = back_on.date() if back_on else None` seems to indicate that this can be `None`, could crash if it happens

addons/hr_holidays/models/hr_employee.py
phenix-factory: That is better I guess

addons/hr_holidays/models/hr_employee.py
phenix-factory: It's unlikely that they have encoded leave. But you are right, it's technically possible.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -206,17 +236,18 @@ def _compute_leave_status(self):
             employee.id: holiday[0].date_to
             for employee, holiday in employee_holidays.items()
         })
-
         for employee, holidays in employee_holidays.items():
             latest_holiday = holidays[0]
             employee.leave_date_from = min(holidays.mapped('date_from')).date()
```
ThanhDodeurOdoo: `leave_date_from` is now a Datetime, but datetime.date() is a day, so I think it will count as midnight

addons/hr_holidays/models/hr_employee.py
```diff
@@ -196,7 +226,7 @@ def _compute_leave_status(self):
         # Used SUPERUSER_ID to forcefully get status of other user's leave, to bypass record rule
         holidays = self.env['hr.leave'].sudo().search([
             ('employee_id', 'in', self.ids),
-            ('date_from', '<=', fields.Datetime.now()),
+            ('date_from', '<=', fields.Datetime.now() + timedelta(days=1.0)),
```
ThanhDodeurOdoo: https://github.com/odoo/odoo/blob/62f3cf65baabfdabc309a909095098e0bde925dd/addons/hr_holidays/static/src/core/common/im_status_patch.js#L1-L17

This file sets the im status to `fa-plane` only if the `leave_dat_to` is truthy, I think it should be updated

addons/hr_holidays/models/hr_employee.py
```diff
@@ -206,17 +236,18 @@ def _compute_leave_status(self):
...
-
         for employee, holidays in employee_holidays.items():
             latest_holiday = holidays[0]
             employee.leave_date_from = min(holidays.mapped('date_from')).date()
+            employee.request_date_from_period = latest_holiday.request_date_from_period
             employee.leave_date_to = employee_back_on.get(employee.id, latest_holiday.date_to).date()
             employee.current_leave_state = latest_holiday.state
             employee.is_absent = any(h.work_entry_type_id.count_as == 'absence' for h in holidays)
 
         no_data = self - holidays.employee_id
```
ThanhDodeurOdoo: not directly related to the PR but isn't it an issue that `holidays` is shadowed in the loop above? `holidays.employee_id` then just only means the last one in the loop

ThanhDodeurOdoo: reported at https://github.com/odoo/odoo/pull/254040#discussion_r3021672174

addons/hr_holidays/models/hr_employee.py
phenix-factory: date_from is also a datetime, I guess we don't need `.date()` anymore then

addons/hr_holidays/models/hr_employee.py
phenix-factory: How do you want to update it?
I'm not sure we want to display the plane if the person isn't actually absent?

addons/hr_holidays/models/hr_employee.py
```diff
@@ -210,13 +241,15 @@ def _compute_leave_status(self):
         for employee, emp_holidays in employee_holidays.items():
             latest_emp_holiday = emp_holidays[0]
             employee.leave_date_from = min(emp_holidays.mapped('date_from')).date()
```
zel-odoo: `leave_date_from` is now `Datetime`, but `.date()` truncates it to midnight — so the JS `outOfOfficeDateEndText` renders the start time as "12:00 AM" and the "starting at …" / afternoon case never works in production (the hoot tests pass only because they set the mock datetime directly). Drop `.date()`:
```suggestion
            employee.leave_date_from = min(emp_holidays.mapped('date_from'))
```

addons/hr_holidays/models/hr_employee.py
```diff
@@ -196,7 +227,7 @@ def _compute_leave_status(self):
         # Used SUPERUSER_ID to forcefully get status of other user's leave, to bypass record rule
         holidays = self.env['hr.leave'].sudo().search([
             ('employee_id', 'in', self.ids),
-            ('date_from', '<=', fields.Datetime.now()),
+            ('date_from', '<=', fields.Datetime.now() + timedelta(days=1.0)),
```
zel-odoo: Widening to `now + 1 day` also flips `is_absent` / `current_leave_state` for leaves that only start tomorrow — `hr_presence` then shows those employees as `absent` today (`hr_presence/models/hr_employee.py` L157), and `is_absent` searches/reports pick them up early. Only `leave_date_from` needs the upcoming leave; consider keeping this search on `<= now` and fetching the future `date_from` separately.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -61,6 +68,30 @@ def _compute_current_work_entry_type_id(self):
             employee = self.filtered(lambda e: e.id == holiday.employee_id.id)
             employee.current_work_entry_type_id = holiday.work_entry_type_id.id
 
+    def _compute_next_working_day_on_leave(self):
+        first_leave_by_employee = dict(self.env['hr.leave'].sudo()._read_group(
```
zel-odoo, line 91: `sudo()` here and need a comment

addons/hr_holidays/static/tests/mock_server/mock_models/res_partner.js
```diff
@@ -1,6 +1,40 @@
...
+
+    _get_store_im_status_fields() {
+        return [
+            ...super._get_store_im_status_fields(),
+            mailDataHelpers.Store.one("main_user_id", [
+                "partner_id",
+                mailDataHelpers.Store.many("all_employee_ids", "leave_date_to"),
+            ]),
+        ];
+
```
zel-odoo: no new line here

addons/hr_holidays/static/tests/mock_server/mock_models/res_users.js
```diff
@@ -3,9 +3,17 @@ import { fields } from "@web/../tests/web_test_helpers";
...
+            ["am", "Morning"],
+            ["pm", "Afternoon"],
+        ],
+    });
 
     _store_main_user_fields(res) {
         super._store_main_user_fields(res);
         res.many("all_employee_ids", ["leave_date_to"], { internal: true });
     }
+
```
zel-odoo: no extra line

addons/hr_holidays/models/hr_employee.py
phenix-factory: That's new ! Good thing

zel-odoo:
runbot fails

phenix-factory:
@zel-odoo yes, something is wrong with the store I think, I'm working on it

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
             employee.is_absent = any(e_h.work_entry_type_id.count_as == 'absence' for e_h in emp_holidays)
 
-        no_data = self - holidays.employee_id
+        for employee, emp_holidays in employee_future_holidays.items():
+            latest_emp_holiday = emp_holidays[0]
+            employee.request_date_from_period = latest_emp_holiday.request_date_from_period
+            employee.leave_date_from = min(emp_holidays.mapped('date_from'))
+            employee.leave_date_to = employee_back_on.get(employee.id, latest_emp_holiday.date_to).date()
+            employee.current_leave_state = latest_emp_holiday.state
+            employee.is_absent = False
```
ThanhDodeurOdoo: I think that this loop can override is_absent on employees that are currently on leave (like if you're on leave and have another leave planned)

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
```diff
@@ -6,19 +6,40 @@ import { _t } from "@web/core/l10n/translation";
...
+            return _t("Out of office starting on %(date)s", { date });
+        }
+        if (this.leave_date_from) {
+            if (
+                DateTime.now().hasSame(this.leave_date_from, "day") &&
+                (this.request_date_from_period === "pm" || this.leave_date_from.hour > 12)
+            ) {
+                const time = this.leave_date_from.toLocaleString(DateTime.TIME_SIMPLE);
+                return _t("Out of office starting at %(time)s", { time });
+            }
```
alexkuhn: I wonder if we should also display the time for "pm" when this is `next_working_day_on_leave`. Just stating the date can be misleading in thinking the person is only available for 1 extra day when this is 1 and a half day. The last morning might be a good slot to plan a last minute stuff with the person before leave.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -47,6 +53,7 @@ class HrEmployee(models.Model):
         ('presence_holiday_present', 'Present but on leave')])
     member_of_department = fields.Boolean('Member of Department', compute='_compute_member_of_department', search='_search_part_of_department')
     hr_responsible_id = fields.Many2one(domain=lambda self: self.env['hr.version']._get_hr_responsible_domain())
+    next_working_day_on_leave = fields.Date(compute="_compute_next_working_day_on_leave")
```
alexkuhn:
```suggestion
    next_working_day_before_leave = fields.Date(compute="_compute_next_working_day_before_leave")
```

The current naming is strange: "working day on leave" doesn't make sense as someone cannot be both on a working day and on leave.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -61,6 +68,32 @@ def _compute_current_work_entry_type_id(self):
...
+            employee.id: fields.Datetime.now() + timedelta(days=1)
+            for employee in first_leave_by_employee
+        }, compute_leaves=False)
+        for employee in self:
+            next_working_day = next_working_day_by_employee.get(employee.id)
+            next_leave_day = first_leave_by_employee.get(employee)
+            if next_working_day and next_leave_day and next_working_day.date() == next_leave_day.date():
+                employee.next_working_day_on_leave = next_leave_day
+            else:
+                employee.next_working_day_on_leave = None
```
alexkuhn, line 113: Ok to use an interval of 1 week to find the next working day before next leave, realistically no one should skip a whole week while working... but there might be exceptions that are not worth making code more complex for them?

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
```diff
@@ -6,19 +6,40 @@ import { _t } from "@web/core/l10n/translation";
...
+        this.next_working_day_on_leave = fields.Date();
     },
-    /** @returns {string} */
     get outOfOfficeDateEndText() {
-        if (!this.leave_date_to) {
-            return "";
+        if (this.next_working_day_on_leave) {
+            const date = toLocaleDateString(this.next_working_day_on_leave);
+            return _t("Out of office starting on %(date)s", { date });
+        }
```
alexkuhn: The back on should have priority over the `next_working_day_before_leave`.
This is strange to see someone in a leave but the system notifies that someone will be on leave a few days later:

Image

For this case, I'm on Wednesday and Marc Demo has approved leaves on Tuesday/Wednesday and another one on Friday. So basically Marc Demo is no leave and is available on Thursday, then again on leave on Friday. I would except to see "Back on Thursday" as text (replace Thursday by the date).

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
 
     def _compute_leave_status(self):
         # Used SUPERUSER_ID to forcefully get status of other user's leave, to bypass record rule
-        holidays = self.env['hr.leave'].sudo().search([
+        futures_holidays = self.env['hr.leave'].sudo().search([
             ('employee_id', 'in', self.ids),
-            ('date_from', '<=', fields.Datetime.now()),
+            ('date_from', '<=', fields.Datetime.now() + timedelta(days=1.0)),
```
alexkuhn: Future holiday could be in a few hours (we are in morning, next leave is on the afternoon), so I think this should be a date_from now and later to find the future holiday.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
             ('date_to', '>=', fields.Datetime.now()),
             ('state', '=', 'validate'),
         ], order='date_to desc')
+        holidays = futures_holidays.filtered(lambda h: h.date_from <= fields.Datetime.now())
```
alexkuhn: Currently on holiday should not rely on future holidays, as the employee can be already on holiday right now from a `date_from` of many weeks ago.

----

For a concrete example with Demo data:

I'm on Wednesday in the afternoon (16:00), with Marc Demo on leave on Tuesday, Wednesday, and on Friday.

I would expect Marc Demo to be on leave, showing "Back to (Thursday_datetime)". Currently this suggests the next leave (Previous comment of mine), but this also removes the fa-plane icon on Marc Demo when this should be on leave right now:

Image

addons/hr_holidays/models/hr_employee.py
phenix-factory: Well, not exactly.
This contains a `date` if the next working day is also encoded as leave. So basically expressed: 

> The next working day of the employee's schedule is also the starting day of a leave period.

Or 

> Is the next normal work day part of the next leave period

The `before` in this case is just wrong: it's not before.
I was considering `is_next_working_day_on_leave`, but it's also wrong because it's not a Boolean.

addons/hr_holidays/models/hr_employee.py
phenix-factory: Ha yes! I remember the struggle with this.

We are, in fact, using an interval of 7 days, but not for what you think: it's to cover every day of the week.
Remember that we want to show if the employee's next "normal" workday has a leave encoded. The only assumption here is that every employee works at least one day per week. 
So, if you are only working on Mondays and you encode a leave for the next one, we need to start displaying the leave alert by Tuesday, since your "next working day" has a leave encoded for it. This is actually important because it creates a 2 weeks delai before you are working again.

Now, if some people have work schedules that are one day every two weeks, it's not going to work, but I bet that is pretty rare and doesn't matter for this feature.

addons/hr_holidays/models/hr_employee.py
alexkuhn: What about `working_day_before_next_leave`?
My understanding of this field is to determine whether we should show on UI that the person is working right now but is about to be on leave soon.

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
```diff
@@ -6,19 +6,43 @@ import { _t } from "@web/core/l10n/translation";
...
+                const time = this.leave_date_from.toLocaleString(DateTime.TIME_SIMPLE);
+                return _t("Out of office starting at %(time)s", { time });
+            }
+            if (DateTime.now().plus({ day: 1 }).hasSame(this.leave_date_from, "day")) {
+                return _t("Out of office tomorrow");
+            }
+        }
+        if (this.next_working_day_on_leave) {
+            const date = toLocaleDateString(this.next_working_day_on_leave);
+            return _t("Out of office starting on %(date)s", { date });
```
alexkuhn: In discuss app as a banner above the message list that's ok, but the other UI it's quite narrow.
So I'd suggest to word it "Out tomorrow" / "Out from [date]" / "Out after [time]".

This is also easier to read when this is this short, and matches the "Back on X" we have now.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -501,7 +501,7 @@ export class DiscussChannel extends Record {
         const showTyping = !ignoreTyping && this.channel.hasOtherMembersTyping;
         return (
             (this.channel.channel_type === "chat" &&
-                (this.channel.correspondent.imStatusUI || showTyping)) ||
+                (this.channel.correspondent?.imStatusUI || showTyping)) ||
```
alexkuhn: Looks like a problem that is fixed by https://github.com/odoo/odoo/pull/277199
It's about to reach master, should remove this optional chaining because that's not the right fix.

addons/hr_holidays/models/hr_employee_public.py
```diff
@@ -19,6 +19,9 @@ class HrEmployeePublic(models.Model):
     is_absent = fields.Boolean('Absent Today', compute='_compute_leave_status', search='_search_absent_employee')
     allocation_display = fields.Char(compute='_compute_allocation_display')
     allocation_remaining_display = fields.Char(related='employee_id.allocation_remaining_display')
+    leave_date_from = fields.Datetime(compute='_compute_leave_status')
+    request_date_from_period = fields.Selection(related='employee_id.request_date_from_period')
+    next_working_day_on_leave = fields.Date(related='employee_id.next_working_day_on_leave', help="The next working day that is also encoded as leave")
```
alexkuhn, line 24: Shouldn't this be a `related` on `employee_id` too?

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
+        futures_holidays -= holidays
```
alexkuhn: I would name the variables `ongoing_leaves` and `upcoming_leaves`:
- "ongoing" / "upcoming" gives very clear understanding of the different types of leaves and what they represent temporarily
- "leaves" is more appropriate than "holidays", holiday is a type of leave but leave is the general-purpose business object to represent modeling of someone being on leave for some reasons, including holidays.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
+            employee.request_date_from_period = latest_emp_holiday.request_date_from_period
+            employee.leave_date_from = min(emp_holidays.mapped('date_from'))
+            employee.leave_date_to = employee_back_on.get(employee.id, latest_emp_holiday.date_to).date()
+            employee.current_leave_state = latest_emp_holiday.state
+            employee.is_absent = False
+
         for employee, emp_holidays in employee_holidays.items():
             latest_emp_holiday = emp_holidays[0]
-            employee.leave_date_from = min(emp_holidays.mapped('date_from')).date()
+            employee.request_date_from_period = latest_emp_holiday.request_date_from_period
```
alexkuhn: `request_date_from_period` is only for upcoming leaves, so strange to see it for ongoing leaves.
That would be nice to keep for upcoming leaves, as right now this hints the employee will be on leave tomorrow but doesn't tell more precisely this is on the afternoon. When it just says tomorrow this is misleading because it gives the impression this is on leave from the morning.

I'm ok with "Out from tomorrow" when this is implicitly in the morning, and then it's "Out from tomorrow [time]" specifically when this is on the afternoon. To promote on the UI.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
+            employee.leave_date_from = min(emp_holidays.mapped('date_from'))
```
alexkuhn: I'm not super fond of field being reused for both upcoming and ingoing, and we have to deduce from the datetime based on now to tell apart...

...But at the same time these fields are deduced from a group of `hr.leaves` so fields on `hr.employee` feels necessary and that's not great to duplicate the fields for the ongoing and upcoming versions...

So I'm ok but I would clarify in `help` of the field definition that this contains ongoing and upcoming. No `help` at all was fine when this only meant ongoing but that's not natural that this is used by both now.

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
alexkuhn: Also I feel that this is useful to know for how long the person would be on leave. If that's only half a day people can just wait for the person to come back. However if this is on leave for 3 weeks then you have another stance and quite a few people may rush to do something before this person is on leave.

Knowing when the person is back when the person on leave is too late!

Considering we don't have much space on some UI, like ctrl-k or in sidebars (messaging menu / member list), there we can omit the "Until [time]", but in Discuss app's header and on avatar card I think we can put the "Until [time]" and that's very handy to see this complete information over there.

addons/hr_holidays/models/hr_employee.py
alexkuhn: Actually the name of field is good ;)

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
phenix-factory: That is good to know

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
phenix-factory: I agree for the rename, why not. But let's expand the scope of this and add every useful thing we can think of around the leave status. We show that an employee is on leave and will be on leave soon. It's already better than before, we can improve afterwards.

addons/hr_holidays/models/hr_employee_public.py
phenix-factory: But it is? Is this comment on the right line?

addons/hr_holidays/static/src/core/common/hr_employee_model_patch.js
alexkuhn: Right now we just say when the person comes back so this is worse in the sense it doesn't warn that the person will be on leave soon.

With this PR, this is a good addition, but we should be cautious to not make it feel half-backed. Also we don't have that much opportunities to work on out-of-office, so that's why some polishing now is not too bad especially when we have good proposal for the issues we see right now ;)

addons/hr_holidays/models/hr_employee.py
phenix-factory: > request_date_from_period is only for upcoming leaves, so strange to see it for ongoing leaves.

The compute method needs to set up all the fields, otherwise, it crashes. We could set it to false, but I don't see the point.

> That would be nice to keep for upcoming leaves, as right now this hints the employee will be on leave tomorrow but doesn't tell more precisely this is on the afternoon. When it just says tomorrow this is misleading because it gives the impression this is on leave from the morning.

Did you test that? 
Because I think if we are "today" and the employee is on leave "tomorrow afternoon", it doesn't display anything about that (yet, it's an edge case to me right now, to improve later)

Bertrand2:
Hello :wave: Putting my HR cents here!
I see while testing that Marc is on leave (plane icon) while he isn't (he is off only tomorrow), I'm guessing it isn't intended? On the runbot I set him a leave starting only tomorrow.
image

Also on a side note, I wonder if instead of having multiple fields to compute when the current leave is starting/finishing, it wouldn't make sense to simply compute the message content directly?
(I haven't reviewed the PR thoroughly yet)

addons/hr_holidays/models/hr_employee.py
phenix-factory: We could do this: 

```py
 employee_leaves = {**employee_upcoming_leaves, **employee_ongoing_leaves}
  for employee, emp_holidays in employee_leaves.items():
      latest_emp_holiday = emp_holidays[0]
      employee.request_date_from_period = latest_emp_holiday.request_date_from_period
      employee.leave_date_from = min(emp_holidays.mapped('date_from'))
      employee.leave_date_to = employee_back_on.get(employee.id, latest_emp_holiday.date_to).date()
      employee.current_leave_state = latest_emp_holiday.state
      employee.is_absent = employee in employee_ongoing_leaves and any(
          e_h.work_entry_type_id.count_as == 'absence' for e_h in emp_holidays
      )
```

But it makes the code more obscure and doesn't show that the order matters. Note that the previous code doesn't do a good job either. Cf: [Than comment's](https://github.com/odoo/odoo/pull/244753#discussion_r3534216419)
Maybe just a comment inside the code?

phenix-factory:
@Bertrand2 Indeed, if it's a side effect of this PR, it's not intended. The plane should be for people who are on leave, not for upcoming leave.

phenix-factory:
@Bertrand2 
> Also on a side note, I wonder if instead of having multiple fields to compute when the current leave is starting/finishing, it wouldn't make sense to simply compute the message content directly?

If you do that, you lose the flexibility to have the exact date in the JS code, for example, they can be formatted using the browser localization (`toLocalDateString`).

addons/hr_holidays/models/hr_employee_public.py
alexkuhn: Let me check again... I wonder if I missed the comment to another line or I was blind

alexkuhn: Looks good, maybe it was missing `related` at some point or I was blind.

addons/hr_holidays/models/hr_employee.py
alexkuhn: It displays that the person will be on leave tomorrow, which is true but is somewhat misleading because usually when this says "on leave tomorrow", we assume this is inclusive to the whole day of tomorrow.

It would have been nicer to more specifically tell on the afternoon, as the availability in the morning is actually quite important to postpone last-minute stuffs to that morning rather than today.

addons/hr_holidays/models/hr_employee.py
alexkuhn: It's not really the amount of LOCs that feels a bit wrong, it's more that we have different semantics of these fields on whether this is a later or ongoing/early date, which can make code harder to read as we need to guess how to read these fields.

But duplicate columns is not necessarily better, there are already quite a lot of fields on employee. So I think that's decent as long as we clarify how to read these fields in `help` and/or `string` for now.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
+        for employee, emp_holidays in employee_upcoming_leaves.items():
+            latest_emp_holiday = emp_holidays[0]
+            employee.request_date_from_period = latest_emp_holiday.request_date_from_period
+            employee.leave_date_from = min(emp_holidays.mapped('date_from'))
+            employee.leave_date_to = employee_back_on.get(employee.id, latest_emp_holiday.date_to).date()
+            employee.current_leave_state = latest_emp_holiday.state
+            employee.is_absent = False
+
+        for employee, emp_holidays in employee_ongoing_leaves.items():
             latest_emp_holiday = emp_holidays[0]
```
alexkuhn: Should also rename variable from `holiday` to `leave` for more appropriate and consistent wording

addons/hr_holidays/models/hr_employee.py
```diff
@@ -194,29 +227,42 @@ def collect_employees(work_intervals):
...
         # Used SUPERUSER_ID to forcefully get status of other user's leave, to bypass record rule
-        holidays = self.env['hr.leave'].sudo().search([
+        upcoming_leaves = self.env['hr.leave'].sudo().search([
             ('employee_id', 'in', self.ids),
-            ('date_from', '<=', fields.Datetime.now()),
+            ('date_from', '<=', fields.Datetime.now() + timedelta(days=1.0)),
             ('date_to', '>=', fields.Datetime.now()),
             ('state', '=', 'validate'),
         ], order='date_to desc')
+        ongoing_leaves = upcoming_leaves.filtered(lambda h: h.date_from <= fields.Datetime.now())
```
alexkuhn:
```suggestion
        ongoing_leaves = upcoming_leaves.filtered(lambda l: l.date_from <= fields.Datetime.now())
```

addons/hr_holidays/static/tests/thread_icon_patch.test.js
```diff
@@ -65,3 +80,27 @@ test("thread icon of a chat when correspondent is on leave & offline", async ()
...
+        ],
+        channel_type: "chat",
+    });
+    await start();
+    await openDiscuss();
+    await contains(".o-mail-NotificationItem", {
+        contains: [".o-mail-ThreadIcon:not([data-icon='travel'])"],
+        text: "Demo",
+    });
+});
```
alexkuhn: Instead of negative assertion, you should assert the explicit icon, whether the person is online / away / busy / offline.

Again, negative assertions are poor in terms of code maintenance. In particular, if your test was written before icons syntax were changed, this would pass due to outdated selector. Avoiding negative assertions and using positive assertions is best to keep test failing when it needs to.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -32,8 +32,14 @@ class HrEmployee(models.Model):
             ('validate', 'Approved'),
             ('cancel', 'Cancelled'),
         ], groups="hr.group_hr_user")
-    leave_date_from = fields.Date('From Date', compute='_compute_leave_status', groups="hr.group_hr_user")
+    leave_date_from = fields.Datetime('From Date', compute='_compute_leave_status', groups="hr.group_hr_user")
     leave_date_to = fields.Date('To Date', compute='_compute_leave_status')
```
alexkuhn: I think it would be nice a help to tell that this is meant for ongoing and upcoming leave, depending on the datetime value.

Before this PR / task this was easier to understand:
- unset = no leave
- set = ongoing leave

Now when set this can mean either an ongoing leave or upcoming leave. I think a help would... help... to tell that there's also the upcoming aspect in the semantics of these fields.

phenix-factory:
@alexkuhn I'm working on an other way to do this that might be better/simpler

addons/hr_holidays/models/hr_employee.py
phenix-factory: There is `request_date_from_period` who handle when a multi-day leave starts (am/pm)
And there is `request_duration` that handles single-day leave.
I updated the code to handle all of those cases.

addons/hr_holidays/static/tests/thread_icon_patch.test.js
phenix-factory: I can just asset the online icon is used I guess

phenix-factory:
@alexkuhn I have updated the code. Unfortunately, it's not simpler than before. But now the morning/afternoon cases should be correctly handled.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -32,8 +32,26 @@ class HrEmployee(models.Model):
             ('validate', 'Approved'),
             ('cancel', 'Cancelled'),
         ], groups="hr.group_hr_user")
-    leave_date_from = fields.Date('From Date', compute='_compute_leave_status', groups="hr.group_hr_user")
+    leave_date_from = fields.Datetime('From Date', compute='_compute_leave_status', groups="hr.group_hr_user",
+        help="Start date of the current validated leave, or of tomorrow's if none is ongoing.")
     leave_date_to = fields.Date('To Date', compute='_compute_leave_status')
+    request_date_from_period = fields.Selection(
```
alexkuhn:
```suggestion
    leave_request_date_from_period = fields.Selection(
```

This is on `hr_employee` model, so should make it clear this is for the sake of `leave`.

addons/hr_holidays/models/hr_employee.py
```diff
@@ -32,8 +32,26 @@ class HrEmployee(models.Model):
...
+        [('am', 'Morning'), ('pm', 'Afternoon')],
+        string="Date Period Start",
+        default='am',
+        compute="_compute_leave_status"
+    )
+    request_duration = fields.Selection(
```
alexkuhn:
```suggestion
    leave_request_duration = fields.Selection(
```

Same idea.

addons/hr_holidays/models/hr_employee_public.py
```diff
@@ -19,6 +19,10 @@ class HrEmployeePublic(models.Model):
     is_absent = fields.Boolean('Absent Today', compute='_compute_leave_status', search='_search_absent_employee')
     allocation_display = fields.Char(compute='_compute_allocation_display')
     allocation_remaining_display = fields.Char(related='employee_id.allocation_remaining_display')
+    leave_date_from = fields.Datetime(compute='_compute_leave_status')
+    request_date_from_period = fields.Selection(related='employee_id.request_date_from_period')
```
alexkuhn:
```suggestion
    leave_request_date_from_period = fields.Selection(related='employee_id.leave_request_date_from_period')
```

PR #244754, opened 2026-01-20
[FIX] mail: show explicit error of fetch message in message list

alexkuhn:
Backport of https://github.com/odoo/odoo/pull/244094

Before this commit, when message list failed to load, it just displays a "Ann error occurred" generic message with a retry button.

This assumes that error happens rarely and when so this is temporarily. However some errors are persistent and it's frustrating to have no clue on why there's error or what may have caused it.

This commit shows the `Error.toString()` from fetch message RPC failure on UI, so that there's a clue on the reason the fetch of messages failed.

Before / After
Screenshot 2026-01-20 at 15 10 38 Screenshot 2026-01-20 at 15 09 07

### DIFF
--- a/addons/mail/static/src/core/common/core.scss
+++ b/addons/mail/static/src/core/common/core.scss
@@ -30,6 +30,10 @@
     font-size: smaller;
 }
 
+.o-xsmaller {
+    font-size: 0.65rem;
+}
+
 .o-text-white {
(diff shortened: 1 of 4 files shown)


phenix-factory (approved):
master should have a test for this.
Otherwise, LGTM 

@robodoo delegate+

PR #244774, opened 2026-01-20
[REV] delivery: revert the commit 6901863

DjamelTouati:
Bug introdcued in: https://github.com/odoo/odoo/commit/6901863d30f524c160971c9f5c97846ffd1fa5f5

The fix was misleading because forcing the context in the unit test caused incorrect assumptions. Passing `default_carrier_id` in the context prevents the computed method from being called, breaking expected behavior.

This revert restores the previous behavior as the fix is incorrect. The issue will be resolved in another PR.


vchu-odoo:
Thank you!

@robodoo r+ squash

PR #224553, opened 2025-08-27
[IMP] improve `mail` visual consistency 

stefanorigano:
task-5042678

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #244806, opened 2026-01-20
[FIX] mail: channel member actions are hard to find

alexkuhn:
Before this commit, the channel member actions were hard to find:
- Click on Members panel.
- Click on Member to open popover card.
- Click on "..." in top-right corner of card.

This is hard because only this avatar card from this menu has the "..." button, and the actions are hidden in this menu. This is easy to miss since avatar cards in message list or other places don't have this "...", and since the actions are hidden there many people could easily miss these actions.

This commit improves the visibility of action by their showing as a "..." on the member item on hover in the "Members" panel. This works like the "..." button in the discuss left sidebar, where the button is shown on mouse-hover. Mobile view (small or mobile device) shows the button all the time next to member, making the discoverability of the action very clear.

Task-5871730

Before / After
Screenshot 2026-01-23 at 14 45 17
Screenshot 2026-01-26 at 11 56 09

----

This PR also makes channel member actions available in guest items, so we can now "Remove Member" on guests too.

Screenshot 2026-01-27 at 18 25 54

### DIFF
--- a/addons/mail/static/src/discuss/core/common/channel_member.js
+++ b/addons/mail/static/src/discuss/core/common/channel_member.js
@@ -1,12 +1,16 @@
 import { ImStatus } from "@mail/core/common/im_status";
 import { ActionPanel } from "@mail/discuss/core/common/action_panel";
+import { useChannelMemberActions } from "@mail/discuss/core/common/channel_member_actions";
 
 import { Component, useState } from "@odoo/owl";
+import { Dropdown } from "@web/core/dropdown/dropdown";
+import { useDropdownState } from "@web/core/dropdown/dropdown_hooks";
 
 import { useService } from "@web/core/utils/hooks";
+import { ActionList } from "@mail/core/common/action_list";
 
 export class ChannelMember extends Component {
-    static components = { ImStatus, ActionPanel };
+    static components = { ActionList, ActionPanel, Dropdown, ImStatus };
     static props = ["member"];
     static template = "discuss.ChannelMember";
 
@@ -14,6 +18,8 @@ export class ChannelMember extends Component {
         super.setup();
         this.state = useState({});
         this.store = useService("mail.store");
+        this.actions = useChannelMemberActions({ member: () => this.props.member });
+        this.showingActions = useDropdownState();
     }
 
     /** @return {import("models").ChannelMember} */
--- a/addons/mail/static/src/discuss/core/common/channel_member_actions.js
+++ b/addons/mail/static/src/discuss/core/common/channel_member_actions.js
@@ -1,0 +1,101 @@
+import { registry } from "@web/core/registry";
+import { Action, ACTION_TAGS, useAction, UseActions } from "@mail/core/common/action";
+import { _t } from "@web/core/l10n/translation";
+import { ConfirmationDialog } from "@web/core/confirmation_dialog/confirmation_dialog";
+import { rpc } from "@web/core/network/rpc";
+
+export const channelMemberActionsRegistry = registry.category("discuss.channel.member/actions");
+
+/** @typedef {import("@odoo/owl").Component} Component */
+/** @typedef {import("@mail/core/common/action").ActionDefinition} ActionDefinition */
+/** @typedef {import("models").ChannelMember} ChannelMember */
+
+/**
+ * @typedef {ActionDefinition} ChannelActionDefinition
+ */
+
+/**
+ * @param {string} id
+ * @param {ChannelActionDefinition} definition
+ */
+export function registerChannelMemberAction(id, definition) {
+    channelMemberActionsRegistry.add(id, definition);
+}
+
+registerChannelMemberAction("set-admin", {
+    condition: ({ member }) => member.canSetAdmin,
+    icon: "fa fa-star text-primary",
+    name: _t("Set Admin"),
+    onSelected: ({ member }) => member.setChannelRole("admin"),
+    sequence: 10,
+});
+
+registerChannelMemberAction("remove-admin", {
+    condition: ({ member }) => member.canRemoveAdmin || member.canRemoveOwner,
+    icon: ({ member }) =>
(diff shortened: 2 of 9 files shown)


addons/mail/static/src/discuss/core/common/channel_member.js
```diff
@@ -2,18 +2,24 @@ import { ImStatus } from "@mail/core/common/im_status";
 import { ActionPanel } from "@mail/discuss/core/common/action_panel";
 
 import { Component, useState } from "@odoo/owl";
+import { Dropdown } from "@web/core/dropdown/dropdown";
+import { useDropdownState } from "@web/core/dropdown/dropdown_hooks";
 
 import { useService } from "@web/core/utils/hooks";
+import { useChannelMemberActions } from "./channel_member_actions";
```
seb-odoo: absolute import

addons/mail/static/src/discuss/core/common/channel_member_actions.js
```diff
@@ -0,0 +1,101 @@
...
+    condition: ({ member }) => member.canSetOwner,
+    icon: "fa fa-star text-warning",
+    name: _t("Set Owner"),
+    onSelected: ({ member }) => member.setChannelRole("owner"),
+    sequence: 30,
+});
+
+registerChannelMemberAction("remove-member", {
+    condition: ({ member }) => member.canRemoveMember,
+    icon: "fa fa-sign-out fa-rotate-180",
```
seb-odoo: Why rotate?

addons/mail/static/src/discuss/web/avatar_card/avatar_card_popover.js
```diff
@@ -167,6 +151,7 @@ export class AvatarCardPopover extends Component {
         this.props.close();
     }
 
+    /** @deprecated */
     onClickRemove() {
         this.dialog.add(ConfirmationDialog, {
```
seb-odoo, line 156: Could this somehow use the action method to avoid duplicating it?

seb-odoo (approved):
@robodoo delegate+

addons/mail/static/src/discuss/core/common/channel_member_actions.js
alexkuhn: This is to match with the sign-out of live chat channel.

seb-odoo: I'd rather we match the leave channel icon, as it does the same kind of action.

Live chat does a different thing.

addons/mail/static/src/discuss/web/avatar_card/avatar_card_popover.js
alexkuhn: This is harder than it looks:
- to make actions available here, it must be passed from `ChannelMember` component, through `env` or explicit `props`
- dialog doesn't propagate the `env`... So that would mean another prop just for that
- This uses generic component `ConfirmDialog`. Another prop means necessarily a custom dialog.

So I felt this was simpler to flag `@deprecated` here.

addons/mail/static/src/discuss/core/common/channel_member_actions.js
alexkuhn: All these actions should have be the same.
I'd argue the "Leave Channel" should rotate too.

The rotate was explicitly asked for live chat channel, which makes sense as this is a core action of livechat. The one from "Leave channel", very few people see and use it, and when that's the case there's the label so people don't think to much about the icon. But I'm pretty sure the intended icon of `sign-out` is the rotate version.

The rotate version makes it so that the position of arrow relative to the rectangle is the same. This is visually more pleasing:
Screenshot 2026-02-02 at 14 12 40

alexkuhn: And an arrow going away from label gives better vibes of a "leave" action.

seb-odoo: I much prefer the non-rotated version. We read from left to right. We are inside (on the left) and the arrow should point outside towards the right to leave. As the icon is intended to be.

alexkuhn: @stefanorigano The rotated `fa-sign-out` for live chat was introduced by https://github.com/odoo/odoo/pull/224553/changes#diff-d49b4c2d8dbdaeaf71c9e2d68543497aff874cd504138e63620e1eb7160a2fc1L28
Could you clarify the intention behind it? And also whether what is your opinion about the orientation of this icon in other contexts like "Leave Channel" or "Remove Member" inside dropdown? Thanks!

stefanorigano: Hello 👋

The goal is to provide a more immediate visual feedback while **toggling** 

https://github.com/user-attachments/assets/9412788c-dde2-4523-83f1-ab81f76d3005

---

Without rotation it would be slightly less clear.  Since the icon is quite small the only thing you clearly see is the arrow, without rotation it moves just slightly. 

https://github.com/user-attachments/assets/8507f1c8-8316-49fc-8fe8-642aa28a4314

The color helps, but still, we know that we cannot rely on that cause of filters (see eye-confort mode), and 'cause  of unpaired users. 

---

> orientation of this icon in other contexts like "Leave Channel" or "Remove Member" inside dropdown

I wouldn’t worry too much about consistency in this case. It’s always going to be a bit tricky for people to figure out what the symbol means, especially since the box is so thin at this font size. It’s not super reliable, and people will probably always go with the label instead.
As long as the label is clear (like “Remove Member”), I think we’re good 👍


Luckily we'll move to google symbols soon, those icons are way more cleaner.
image

alexkuhn: Thanks! I've pushed version without `fa-rotate-180` then :)

PR #244818, opened 2026-01-20
[IMP] l10n_tr_nilvera_einvoice: migrate last fetched dates to fields

alialfie:
To store the last fetched date in stable to allow for synchronization, config parameters were used. In master, these are now replaced with dedicated fields on the company model for better performance and easier maintenance.

This commit adds dedicated fields on res.company for einvoice and earchive dates.

task-5865683

Upgrade PR: https://github.com/odoo/upgrade/pull/9315


jbw-odoo (approved):
Looks good, thanks
robodoo r+

PR #244837, opened 2026-01-20
[FIX] base: fix `click_all` menus demo test access rights

Dj0ulo:
__Steps to reproduce:__

1. Create a new db with `contacts` and `account_followup` installed
   without demo data.
2. Run `TestMenusDemo` using `--test-tags click_all:TestMenusDemo`.
   => It fails when testing the "Overdue Customer Invoices" filter.

__Reason:__

`TestMenusDemo` inherits from `HttpCaseWithUserDemo`, which creates a
fake user demo if the database does not already contain one (meaning
that it has no demo data). However, when doing so, it does not include
all the access groups needed by the different modules.

Since this filter requires access to `account.move.line`, an
`AccessError` is raised and the test fails.

__Fix:__

- Inside `TransactionCaseWithUserDemo` setup, add all the access groups
  in which Marc Demo normally is included to better represent the real
  flow.
- Clean up the duplicate code.

runbot-231045

### DIFF
--- a/odoo/addons/base/tests/common.py
+++ b/odoo/addons/base/tests/common.py
@@ -161,29 +161,6 @@ class TransactionCaseWithUserDemo(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
 
-        cls.env.ref('base.partner_admin').write({'name': 'Mitchell Admin'})
-        cls.user_demo = cls.env['res.users'].search([('login', '=', 'demo')])
-        cls.partner_demo = cls.user_demo.partner_id
-
-        if not cls.user_demo:
-            cls.env['ir.config_parameter'].sudo().set_param('auth_password_policy.minlength', 4)
-            cls.partner_demo = cls.env['res.partner'].create({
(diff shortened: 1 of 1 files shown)


odoo/addons/base/tests/common.py
```diff
@@ -201,31 +178,43 @@ def setUpClass(cls):
...
-            })
-            cls.user_demo = cls.env['res.users'].create({
-                'login': 'demo',
-                'password': 'demo',
-                'partner_id': cls.partner_demo.id,
-                'group_ids': [Command.set([cls.env.ref('base.group_user').id, cls.env.ref('base.group_partner_manager').id])],
-            })
+    def _get_user_demo_access_groups(cls):
+        group_xml_ids = [
+            'account.group_account_user',
```
kmagusiak, line 187: This is not in base. You should not add all groups, you will just mask errors.

kmagusiak (changes requested):
Fix the issue instead of adding all permission in base.

PR #244912, opened 2026-01-21
[FIX] pos_loyalty: load product without gift card

aksi-odoo:
Step to reproduce
- install `pos_loyalty` and have two companies
- for this record: `loyalty.gift_card_product_50` set company_id = company1
- switch to company b and start pos

Observation:
- no products are loaded

Cause:
- `browse()` was used to load records `gift_card_product_50` and `ewallet_product_50`
- when any of them is restricted on one company it fails to load whole model

Fix:
- Instead of directly loading them, we use `search()` to avoid `AccessError` so that rest of the products can be loaded.
- Gift card is still not shown as expected

opw-5449466

---

### DIFF
--- a/addons/pos_loyalty/models/product_template.py
+++ b/addons/pos_loyalty/models/product_template.py
@@ -29,7 +29,7 @@ class ProductTemplate(models.Model):
             'loyalty.gift_card_product_50',
             'loyalty.ewallet_product_50',
         ])
-        special_display_products = self.env['product.product'].browse(loyality_products)
+        special_display_products = self.env['product.product'].search([('id', 'in', loyality_products)])
         # Include trigger products from loyalty programs of type 'gift_card' or 'ewallet'
         special_display_products += self.env['loyalty.program'].search([
             ('program_type', 'in', ['gift_card', 'ewallet']),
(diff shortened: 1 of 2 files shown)


alsh-odoo (approved):

robinengels (approved):
LGTM

addons/pos_loyalty/models/product_template.py
```diff
@@ -29,7 +29,8 @@ def _load_pos_data(self, data):
             'loyalty.gift_card_product_50',
             'loyalty.ewallet_product_50',
         ])
-        special_display_products = self.env['product.product'].browse(loyality_products)
+        # sudo(): to avoid scenario when above products are limited to a company
+        special_display_products = self.env['product.product'].sudo().browse(loyality_products)
```
adgu-odoo: If they are limited to a company, wouldn't it be better if we just do

```suggestion
        special_display_products = self.env['product.product']
        for product in loyality_products:
            try:
                special_display_products |= self.env['product.product'].sudo().browse(product)
            except AccessError:
                continue
```

aksi-odoo: Hey @adgu-odoo  thanks for review , i have update the pr.

PR #244915, opened 2026-01-21
[FIX] l10n_vn: add missing VAT 8% tag to report formula

matd-odoo:
`VAT_SALES` report line aggregates total untaxed amount from its children lines.

Previously, the formula for this line was missing `VAT_SALES_8.amount_untaxed`.
As a result, the base amount for 8% VAT transactions was excluded from the total
sales base calculation.

This commit adds the missing tag to the `VAT_SALES` formula to ensure the total
taxable base is calculated correctly.

task-5836154


khva-odoo (approved):
LGTM

vin-odoo (approved):
robodoo r+

Thank you!

PR #204805, opened 2025-04-04
[IMP] sale,*: light improvements

khaj-odoo:
Description of the issue/feature this PR addresses:
Improvements to Sales module and portal view.

Current behavior before PR:
- Delivery date always shown in PDF  
- Old demo reward for Simple Pen  
- Sales warning field had a label  
- Static quote builder video timestamp  
- No Shipping Status in SO portal

Desired behavior after PR is merged: 
- Show delivery date in PDF only if set  
- Update Simple Pen reward to 1 per $50  
- Remove label from sales warning field  
- Use dynamic timestamp for quote video  
- Show Shipping Status in SO portal view

See also:

- Enterprise PR: https://github.com/odoo/enterprise/pull/82886

Task-4402920
---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #244917, opened 2026-01-21
[FIX] {l10n_din5008_,}sale: missing Delivery Date on DIN5008 report and preview

khsr-odoo:
**Steps to reproduce:**
1. Install modules `sale_management` and `l10n_din5008_sale`
2. Go to Settings, Configure Document Layout and set layout to DIN 5008
3. Create a new Sale Order
4. Set a customer, add a product, and fill in the Delivery Date (Other Info)
5. Click on Print and Preview

**Issue:**
The Delivery Date (commitment_date) is not displayed on:
- The DIN 5008 sale order report
- The sale order preview (portal view)

Functional experts confirmed that the Delivery Date must be visible when using the DIN 5008 layout.

**Cause:**
The `commitment_date` field was not included in the DIN 5008 sale order report template nor in the preview view.

**Solution:**
This commit adds the Delivery Date information to:
- The DIN 5008 sale order report template
- The sale order portal/preview view

**opw-5490651**

**Before:**
image
image

**After:**
image
image

### DIFF
--- a/addons/sale/views/sale_portal_templates.xml
+++ b/addons/sale/views/sale_portal_templates.xml
@@ -414,6 +414,10 @@
                                 <th t-else="" class="ps-0 pb-0">Date:</th>
                                 <td class="w-100 pb-0 text-wrap"><span t-field="sale_order.date_order" t-options='{"widget": "date"}'/></td>
                             </tr>
+                            <tr t-if="sale_order.commitment_date and sale_order.state in ['sent', 'sale']">
+                                <th class="ps-0 pb-0">Delivery Date:</th>
+                                <td class="w-100 pb-0 text-wrap"><span t-field="sale_order.commitment_date" t-options='{"widget": "date"}'/></td>
+                            </tr>
                             <tr t-if="sale_order.validity_date and sale_order.state in ['draft', 'sent']">
                                 <th class="ps-0 pb-0">Expiration Date:</th>
                                 <td class="w-100 pb-0 text-wrap"><span t-field="sale_order.validity_date" t-options='{"widget": "date"}'/></td>
--- a/addons/l10n_din5008_sale/report/din5008_sale_templates.xml
+++ b/addons/l10n_din5008_sale/report/din5008_sale_templates.xml
@@ -28,6 +28,10 @@
                                 <td>Order Date:</td>
                                 <td><div t-field="doc.date_order" t-options="{'widget': 'date'}"/></td>
                             </tr>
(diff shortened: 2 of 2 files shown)


alsh-odoo (approved):

addons/sale/views/sale_portal_templates.xml
```diff
@@ -414,6 +414,10 @@
                                 <th t-else="" class="ps-0 pb-0">Date:</th>
                                 <td class="w-100 pb-0 text-wrap"><span t-field="sale_order.date_order" t-options='{"widget": "date"}'/></td>
                             </tr>
+                            <tr t-if="sale_order.commitment_date">
```
agr-odoo: The condition on sale order state needs to be added here, otherwise the line could be broken in the order cases

agr-odoo (changes requested):

agr-odoo (approved):
Looks good, it may mess up with third party customizations, but it can't be helped

addons/sale/views/sale_portal_templates.xml
```diff
@@ -414,6 +414,10 @@
                                 <th t-else="" class="ps-0 pb-0">Date:</th>
                                 <td class="w-100 pb-0 text-wrap"><span t-field="sale_order.date_order" t-options='{"widget": "date"}'/></td>
                             </tr>
+                            <tr t-if="sale_order.commitment_date and sale_order.state in ['sale', 'cancel']">
```
Feyensv: Why should we show the delivery date for cancelled orders? 🤔

Feyensv (commented):
Doesn't really seem to be a fix to me, more an improvement 🤔 

@Jean-boje do we want to add the Expected delivery date in the portal ?

addons/sale/views/sale_portal_templates.xml
khsr-odoo: Hello,
After confirming with the PO, I have updated the state in the portal.
Additionally, in 19.0 the delivery date is already present in the portal view

https://github.com/odoo/odoo/blob/53e09796a842aa020df4411d07ab4a285bd8f484/addons/sale/views/sale_portal_templates.xml#L613-L619

Implemented with: https://github.com/odoo/odoo/pull/204805/files#diff-d81ab625daf015b596e3b4c58624890e73068902abec55ce62add069c3c3a2e7R610

Feyensv (approved):
Ok for `sale` part then 👌

lordkrandel (approved):

PR #244924, opened 2026-01-21
[IMP] website: allow no default value for selection field

divy-odoo:
Previously, selection fields in website forms always included an empty option with a no label that could not be customized.

This change introduces options to control the presence of the empty option and customize its label.

Features added:
- Toggle to enable or disable the placeholder option
- Input field to define a dynamic placeholder text

task-3861126

### DIFF
--- a/addons/website/static/src/builder/plugins/form/form_option.xml
+++ b/addons/website/static/src/builder/plugins/form/form_option.xml
@@ -195,6 +195,14 @@
             action="'toggleRequired'" actionParam="'s_website_form_required'"
         />
     </BuilderRow>
+    <t t-if="domStateCurrentFieldInput.nodeName === 'SELECT'">
+        <BuilderRow label.translate="Allow Empty">
+            <BuilderCheckbox id="'allow_empty_opt'" action="'toggleAllowEmpty'"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Placeholder" level="1">
+            <BuilderTextInput t-if="isActiveItem('allow_empty_opt')" action="'setEmptyPlaceholder'" placeholder.translate="Please choose any one option"/>
+        </BuilderRow>
+    </t>
     <BuilderRow label.translate="Max # of Files">
         <BuilderNumberInput id="'max_files_number_opt'" t-if="isMaxFilesVisible"
             title.translate="The maximum number of files that can be uploaded."
--- a/addons/website/static/src/builder/plugins/form/utils.js
+++ b/addons/website/static/src/builder/plugins/form/utils.js
@@ -95,18 +95,6 @@ export function renderField(field, resetId = false) {
     if (!field.id) {
         field.id = generateHTMLId();
     }
-    if (field.records && (field.type === "many2one" || field.type === "selection")) {
-        const hasDefault =
-            field.records[0]?.["display_name"] === "" ||
-            field.records.some((value) => value.selected);
-        if (!hasDefault) {
-            field.records.unshift({
-                id: "",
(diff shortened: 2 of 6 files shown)


addons/website/static/src/builder/plugins/form/form_option.xml
```diff
@@ -195,6 +195,14 @@
...
         />
     </BuilderRow>
+    <BuilderRow label.translate="Allow Empty">
+        <BuilderCheckbox t-if="domStateCurrentFieldInput.nodeName === 'SELECT'"
+            id="'allow_empty_opt'" preview="false"
+            action="'toggleAllowEmpty'"/>
+    </BuilderRow>
+    <BuilderRow label.translate="Placeholder" level="1">
+        <BuilderTextInput t-if="isActiveItem('allow_empty_opt')" action="'setEmptyPlaceholder'" placeholder.translate="Please choose any one option"/>
+    </BuilderRow>
```
FrancoisGe:
```suggestion
<t  t-if="domStateCurrentFieldInput.nodeName === 'SELECT'">
    <BuilderRow label.translate="Allow Empty">
        <BuilderCheckbox
            id="'allow_empty_opt'" preview="false"
            action="'toggleAllowEmpty'"/>
    </BuilderRow>
    <BuilderRow label.translate="Placeholder" level="1">
        <BuilderTextInput t-if="isActiveItem('allow_empty_opt')" action="'setEmptyPlaceholder'" placeholder.translate="Please choose any one option"/>
    </BuilderRow>
    </t>
```

FrancoisGe: It's easier to read and faster with owl.

addons/website/static/src/builder/plugins/form/form_option.xml
```diff
@@ -195,6 +195,14 @@
             action="'toggleRequired'" actionParam="'s_website_form_required'"
         />
     </BuilderRow>
+    <BuilderRow label.translate="Allow Empty">
+        <BuilderCheckbox t-if="domStateCurrentFieldInput.nodeName === 'SELECT'"
+            id="'allow_empty_opt'" preview="false"
```
FrancoisGe: move `preview = false` in the BuilderAction `toggleAllowEmpty`

FrancoisGe: When your action should never preview, it's best to define it in the action. It's quite rare to define it in the XML.

addons/website/static/src/builder/plugins/form/utils.js
```diff
@@ -514,7 +507,7 @@ export function getListItems(fieldEl) {
     const multipleInputsEl = getMultipleInputs(fieldEl);
     let options = [];
     if (selectEl) {
-        options = [...selectEl.querySelectorAll("option")];
```
FrancoisGe: Did you check "setDependencyValueList" ? It has the same selector, we maybe need to fix it ?

FrancoisGe (commented):
The default value can be selected in the “Visibility Rule” option. If you use “Contains,” it works, but if you use “is equal,” it doesn't work.
I probably won't block the task for that. But do you have any ideas ?

addons/website/static/src/builder/plugins/form/utils.js
divy-odoo: That selector is used to prepare candidates for conditional visibility options. Whatever options the select menu has will be available for the conditional visibility option. IMO, we should consider the empty option for conditional visibility, so I don't think we need to modify that selector.

PR #244926, opened 2026-01-21
[IMP] account,l10n*: add `is_refund` method

hamo-odoo:
Before this commit-
We had methods on `account.move` i.e.
`is_sale_document`, `is_invoice` but
at many instance in the code base,
we did do `self.move_type in ('out_refund', 'in_refund')`

After this commit-
We add new methods on `account.move` to
get and check if the move is refund

task-5850565

---

### DIFF
--- a/addons/account_debit_note/wizard/account_debit_note.py
+++ b/addons/account_debit_note/wizard/account_debit_note.py
@@ -52,7 +52,7 @@ class AccountDebitNote(models.TransientModel):
             record.journal_type = record.move_type in ['in_refund', 'in_invoice'] and 'purchase' or 'sale'
 
     def _prepare_default_values(self, move):
-        if move.move_type in ('in_refund', 'out_refund'):
+        if move.is_refund():
             type = 'in_invoice' if move.move_type == 'in_refund' else 'out_invoice'
         else:
             type = move.move_type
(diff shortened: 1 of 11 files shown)


addons/account_debit_note/wizard/account_debit_note.py
```diff
@@ -65,7 +65,7 @@ def _prepare_default_values(self, move):
                 'debit_origin_id': move.id,
                 'move_type': type,
             }
-        if not self.copy_lines or move.move_type in [('in_refund', 'out_refund')]:
```
hamo-odoo, line 68: @jco-odoo is this something intentional, or some dev error?

addons/l10n_es_edi_facturae/models/account_move.py
```diff
@@ -479,7 +479,7 @@ def grouping_function_per_base_line_tax(base_line, tax_data):
             - invoice_values['TotalGeneralDiscounts']
             + invoice_values['TotalGeneralSurcharges']
         )
-        refund_multiplier = -1 if self.move_type in ('out_refund', 'in_refund') else 1
+        refund_multiplier = -1 if self.is_refund() else 1
```
hamo-odoo, line 512: We can do something for the sign as well, many l10ns do it for the edi purposes :D

hamo-odoo:
> refactoring = master
 
This is a technical improvement, not a refactor. There is nothing i.e. we can't add new methods in stable?

> not sure why we have is_refund public, probably just because you did similar as the other methods.

Yes, to maintain the consistency

jco-odoo:
@hamo-odoo Whatever -> master

hamo-odoo:
@jco-odoo Updated

PR #244928, opened 2026-01-21
[IMP] account_edi_ubl_cii: introduce support for allowance/charge mechanism

hatr-odoo:
This **PR** introduces support for the Allowance/Charge mechanism in UBL and CII EDI formats. Previously, allowance and charge handling was static and implicit: fixed taxes were always exported as charges with code AEO, while line discounts were exported as allowances with code 95, which did not allow correct modeling of many real-world fiscal scenarios.

A new configuration is added on the tax form to explicitly define whether a tax is treated as a regular tax or as an allowance/charge. When marked as an allowance/charge, the corresponding code and reason can be defined. The sign of the amount determines the semantic: negative values are treated as allowances, while positive values are treated as charges.

When applied on invoices, such taxes are exported as AllowanceCharge elements in UBL and CII for EDI purposes. Support is also added for importing invoices containing such allowance/charge entries.

Additionally, test cases are added to validate correct behavior across export and import flows, including scenarios combining standard taxes with allowance/charge entries.

**task**-5424216


addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_account_invoice_form_inherit" model="ir.ui.view">
+        <field name="name">account.tax.form.inherit</field>
+        <field name="model">account.tax</field>
+        <field name="inherit_id" ref="account.view_tax_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='electronic_invoicing']" position="replace">
```
jva-odoo: What if user install this module without updating `account_edi_ubl_cii_tax_extension`
this will work?

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
```diff
@@ -0,0 +1,60 @@
...
+                <page string="Electronic Invoicing" name="electronic_invoicing">
+                    <group>
+                        <field name="ubl_cii_type"
+                            widget="radio"
+                            string="Type"
+                            options="{'horizontal': True}"/>
```
jva-odoo: we can split line if we can't read in on screen.
normally it's 120 characters
```suggestion
                        <field name="ubl_cii_type" widget="radio" string="Type" options="{'horizontal': True}"/>
```

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_account_invoice_form_inherit" model="ir.ui.view">
+        <field name="name">account.tax.form.inherit</field>
+        <field name="model">account.tax</field>
+        <field name="inherit_id" ref="account.view_tax_form"/>
```
jva-odoo:
```suggestion
        <field name="inherit_id" ref="account_edi_ubl_cii_tax_extension.view_account_invoice_form_inherit"/>
```
if you change to specific view then you must inherit that

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
```diff
@@ -0,0 +1,60 @@
...
+                            string="Charge Reason"
+                            invisible="
+                                ubl_cii_type == 'tax'
+                                or not amount >= 0
+                            "/>
+
+                        <field name="ubl_cii_allowance_charge_reason"
+                            nolabel="1"
+                            invisible="ubl_cii_type == 'tax'"/>
+                    </group>
```
jva-odoo, line 32: same for this all

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
hatr-odoo: No, it won't. Thanks for pointing this out. I think then we need to change the versions of both `account_edi_ubl_cii_tax_extension` and `account_edi_ubl_cii`.

jva-odoo: That is also not effective for on premise server.

You can use `convert.convert_file` to reload `account_edi_ubl_cii_tax_extension` view file in pre init.
or
don't modify view in `account_edi_ubl_cii_tax_extension` and add page here
and hind old fields by xpath

jva-odoo (changes requested):

addons/account_edi_ubl_cii_allowance_charge_extension/models/account_edi_common.py
```diff
@@ -0,0 +1,147 @@
...
+        2. not price_include matching the reason_code and amount
+        3. not price_include matching the amount
+        4. price_include matching the reason, reason_code and amount
+        5. price_include matching the reason_code and amount
+        6. price_include matching the amount
+        """
+        amount = allowance_charge_tax_vals['tax_percent'] if allowance_charge_tax_vals['tax_percent'] else allowance_charge_tax_vals['tax_amount']
+        base_domain = [
+            ('company_id', '=', invoice_line_form.company_id.id),
+            ('amount', '=', amount),
```
jva-odoo, line 102: instead of `=` you can use < and > with +-0.01
so if you have amount like 2.0001
it's find 2% tax
that is very high chance because you divide by qty 
```py
>>> 0.1 + 0.2
0.30000000000000004
>>> 4.8 / 1.6
2.9999999999999996
>>> 
```

hatr-odoo: Using < and > with a tolerance of ±0.01 should not be required in this case.

For percentage taxes, we obtain a specific value from `allowance_charge_node` in `tax_percent`, so there is no ambiguity.

For fixed taxes, the value is already rounded when the domain is converted to SQL in `__leaf_to_sql`, through `Float.convert_to_column`, according to the precision defined on the field (`account.tax.amount` in this case). This rounding uses the HALF-UP method. An error should only occur if floating-point noise affects the digit at the rounding boundary (i.e. 4th/5th digit in our case), which should be rare.

`__leaf_to_sql`
https://github.com/odoo/odoo/blob/0593895266ce4c95a1c262b14ab4a9aca7471e3b/odoo/osv/expression.py#L1585

`Float.convert_to_column`
https://github.com/odoo/odoo/blob/0593895266ce4c95a1c262b14ab4a9aca7471e3b/odoo/fields.py#L1545-L1551

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
hatr-odoo: I've confirmed with NASTI.
We will not modify the view in `tax_extension`. Instead, we will handle the entire case within `account_edi_ubl_cii_allowance_charge_extension` itself.

addons/account_edi_ubl_cii_allowance_charge_extension/models/account_edi_common.py
jva-odoo: Indeed, we can ignore for this rare case

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
```diff
@@ -0,0 +1,37 @@
...
+        <field name="arch" type="xml">
+            <!-- Hide tax category fields in existing view. They will be relocated to Electronic Invoicing. -->
+            <xpath expr="//field[@name='ubl_cii_tax_category_code']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='ubl_cii_tax_exemption_reason_code']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//page[@name='advanced_options']" position="after">
+                <page string="Electronic Invoicing" name="electronic_invoicing">
```
jva-odoo, line 16: There is no conditions to hide in other country?

addons/account_edi_ubl_cii_allowance_charge_extension/models/account_edi_common.py
```diff
@@ -0,0 +1,147 @@
...
+                        if reason_code:
+                            domain.append(('ubl_cii_allowance_reason_code', '=', reason_code))
+                    if reason:
+                        domain.append(('ubl_cii_allowance_charge_reason', '=', reason))
+                    tax = self.env['account.tax'].search(domain, limit=1)
+                    if tax:
+                        return tax
+        return self.env['account.tax']
+
+    def _get_tax_category_list(self, invoice, taxes):
```
jva-odoo: https://github.com/odoo/odoo/blob/1c7ae9fb30441197732543ec07e0cc3d1599fe0c/addons/l10n_jo_edi/models/account_edi_xml_ubl_21_jo.py#L339-L354

https://github.com/odoo/odoo/blob/1c7ae9fb30441197732543ec07e0cc3d1599fe0c/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py#L190-L199

this two EDI create it's own vals and here you removed taxes to pass in supper base on conditions.
so must check this for this country and discuses with POs that what if user set this in this specific countries.

or you can apply this to selected countries or EDI.

jva-odoo (changes requested):
Round 2

addons/account_edi_ubl_cii_allowance_charge_extension/models/account_edi_common.py
jva-odoo: I see that you filtered this type of tax from `_get_invoice_line_item_taxes` in [account_edi_xml_ubl_20.py](https://github.com/odoo/odoo/pull/244928/changes#diff-c8ba4aee3311ef9d1f6ecfcfa4e13cb6213fb0866767afc75b72c025020fb9e2)

and this both are inherited from `ubl_21` and `ubl_21` inherit from `ubl_20` so it's not effect because it's filtered before come to this method.

addons/account_edi_ubl_cii_allowance_charge_extension/tests/common.py
```diff
@@ -0,0 +1,183 @@
...
+            'model_id': cls.env['ir.model']._get_id('account.move'),
+            'name': "Invoice: Test Sending",
+            'partner_to': "{{ object.partner_id.id }}",
+            'subject': "{{ object.company_id.name }} Invoice (Ref {{ object.name or 'n/a' }})",
+            'report_template_ids': [(4, cls.env.ref('account.account_invoices').id)],
+            'lang': "{{ object.partner_id.lang }}",
+        })
+
+    @classmethod
+    def setup_company_data(cls, company_name, chart_template=None, **kwargs):
```
jva-odoo: why not you use `TestUBLCommon` and defined same methods?

jva-odoo (changes requested):

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
hatr-odoo: Thanks, that’s a very nice catch.
However, I’m not sure about the specific list of countries that require this `Electronic Invoicing` page.

@nasti-odoo, could you please confirm the list of countries that need this requirement? That way, we can avoid showing irrelevant information to localisations such as India.

addons/account_edi_ubl_cii_allowance_charge_extension/tests/common.py
hatr-odoo: `TestUBLCommon` is defined in the separate test module `l10n_account_edi_ubl_cii_tests`, which is not part of my module’s dependency chain. Therefore, I cannot use it in my module.

So for now, I have defined the required tests within this module itself. Once the extension module is merged into the main module in master, these tests will be transferred to the `l10n_account_edi_ubl_cii_tests` module.

addons/account_edi_ubl_cii_allowance_charge_extension/models/account_edi_common.py
hatr-odoo: Thanks for pointing this out. Indeed, in this case, `_get_tax_category_list` would not have any adverse effect.

However, this method is also used in `_export_invoice_vals`. In that flow, the new `allowance_charge` mechanism would be bypassed. So to preserve the code's integrity, I have introduced an intermediate function, `_get_tax_category_vals`,  that will ensure we get `allowance_charge_vals` correctly.

addons/account_edi_ubl_cii_allowance_charge_extension/views/account_tax_views.xml
jva-odoo: I find PEPPOL_LIST here
https://github.com/odoo/odoo/blob/406b7fb890ed99acfc1159b9174e8f2160c29a18/addons/account/models/company.py#L44
but I check other place it's not hide anything so let's don't improve in 17.0
we can do it in separate task in master if it's not already hide

nasti-odoo: Hello @hatr-odoo, 

As stated in the original Excalidraw, it must only be shown if UBL is installed : 
image

hatr-odoo: @nasti-odoo that's already working. However, we were discussing how to handle it's visiblity based on countries, so that if some localisations don't require UBL, this tab can be hidden for them.

But anyways, as mentioned by JVA sir, let's do it comprehensively in the master.

addons/account_edi_ubl_cii_allowance_charge_extension/tests/common.py
hatr-odoo: It is possible to import TestUBLCommon in this module, even if it's not present in the dependency chain, because it is a valid path already present in the addons namespace. Hence, I've adapted the code accordingly.

For reference,
https://github.com/odoo/odoo/blob/540aa87b8810666037debab0e7146d4063048803/addons/l10n_dk_oioubl/tests/test_xml_oioubl_dk.py#L4
https://github.com/odoo/odoo/blob/540aa87b8810666037debab0e7146d4063048803/addons/l10n_ro_edi/tests/test_xml_ubl_ro.py#L2
https://github.com/odoo/odoo/blob/540aa87b8810666037debab0e7146d4063048803/addons/l10n_rs_edi/tests/test_xml_ubl_rs.py#L2

hatr-odoo: Helper methods have been backported to 17.0 in account_edi_ubl_cii’s common.py. Therefore, there is no longer any need to depend on l10n_account_edi_ubl_cii_tests.

PR #229706, opened 2025-10-02
[IMP] resource: refactor working schedule

romaincarlier4:
First part of the work:
- Display required full time + work time rate in debug models
- Always define the amount of hours per day
- Remove the 2 week calendar feature
- Remove the name and sequence

task-5124688

Description of the issue/feature this PR addresses:

Current behavior before PR:

Desired behavior after PR is merged:

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #244930, opened 2026-01-21
[FIX] hr_holidays: fix has mandatory days compute

romaincarlier4:
The bug occured following this refactoring PR: https://github.com/odoo/odoo/pull/229706 Since the timezone taken is no longer on the calendar but on the employee, the computation of the `has_mandatory_days` field was wrong. Indeed, the field `date_from` and `date_to` on the leave takes into account the timezone of the employee, which might be on a different calendar day in UTC. To fix this, we can use the fields `request_date_from` and `request_date_to`, which are a simple date and not datetime.

Runbot build error: https://runbot.odoo.com/odoo/runbot.build.error/238015 task-5788459

Current behavior before PR:

---

### DIFF
--- a/addons/hr_holidays/models/hr_leave.py
+++ b/addons/hr_holidays/models/hr_leave.py
@@ -493,20 +493,21 @@ class HrLeave(models.Model):
         for holiday in self:
             holiday.department_id = holiday.employee_id.department_id
 
-    @api.depends('date_from', 'date_to', 'holiday_status_id')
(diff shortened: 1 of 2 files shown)


mepe-odoo (changes requested):
Can you add a test with a strange timezone please; to ensure that your changes is correct ? :D

mepe-odoo (approved):
lgtm

PR #244933, opened 2026-01-21
[FIX] account_peppol: Fix embed attachments through peppol

roto-odoo:
[FIX] account_peppol: Fix embed attachments through peppol

Embedded attachments sent through peppol are not decoded causing a XSD
validation error

Steps to reproduce the bug:
- Create an invoice
- In the send & print wizard, select "by peppol" and attach a supported
file
- Send
- Run the scheduled action to update document status
- Go in the invoice form and see the error message in the chatter

no-task

### DIFF
--- a/addons/account_peppol/wizard/account_invoice_send.py
+++ b/addons/account_peppol/wizard/account_invoice_send.py
@@ -244,7 +244,7 @@ class AccountInvoiceSend(models.TransientModel):
             xml_attachment,
             [{
                 'name': attachment.name,
-                'raw_b64': attachment.datas,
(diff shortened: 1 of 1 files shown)


clbr-odoo (changes requested):
small typo in the module name on the commit message :)

clbr-odoo (approved):
@robodoo r+

odoostan:
Thank you @roto-odoo @clbr-odoo ! Do you know if an emergency Odoo.sh revision is in the works for this ? I guess it's affecting all users in 16.0.20260119 using UBL.

PR #244938, opened 2026-01-21
[FIX] crm: prevent error on moving won lead to another won stage in diff year

path-odoo:
Currently, a traceback occurs when a lead won in a previous year is moved to another won stage in the current year.

### **Steps to Reproduce:**
1) Install CRM without demo data.
2) From the `CRM>Configuration>Stages` make `new` stage as **'won'** stage.
 3) Create a lead with dated in the past (e.g., 30-12-2025 by changing system date)
   and with some expected_revenue.
4) Change the system date to today and move the lead to the Won stage.

Ref Video: https://drive.google.com/file/d/18OCQ4Tl6Co_oh28XNag3qjxMkXJNSMOh/view?usp=sharing

### **Error:**
`TypeError: '<' not supported between instances of 'NoneType' and 'float'`

### **Root Cause:**
When a lead is moved to a won stage, `_get_rainbowman_message` is called and computes the values for `max_{team,user}_{31,7}`. However, when the lead spans different years, the condition of SQL query at [1] fails(because 2025 != 2026) due to which SQL query return null from the MAX() Function. As a result, subsequent comparisons at [2] fail, raising an Error.

### **FIX:**
Introduce small helper method(`_is_lower_than_expected_revenue`) to ensure comparisons
across different years only happen when we have meaningful numeric values.

[1]- https://github.com/odoo/odoo/blob/fb4e08fe46c0e1865e30b0ce1eb5f4436438ea03/addons/crm/models/crm_lead.py#L1218

[2]- https://github.com/odoo/odoo/blob/fb4e08fe46c0e1865e30b0ce1eb5f4436438ea03/addons/crm/models/crm_lead.py#L1232

**opw-5484887**
**sentry-7026119584**

### DIFF
--- a/addons/crm/models/crm_lead.py
+++ b/addons/crm/models/crm_lead.py
@@ -1227,15 +1227,18 @@ class CrmLead(models.Model):
         })
         query_result = self.env.cr.dictfetchone()
 
+        def _is_lower_than_expected_revenue(value):
+            return self.expected_revenue and value is not None and value < self.expected_revenue
+
         if query_result['count_user_closed_year'] == 1:
             return _('Go, go, go! Congrats for your first deal.')
-        elif self.expected_revenue and query_result['max_team_31'] < self.expected_revenue:
+        elif _is_lower_than_expected_revenue(query_result['max_team_31']):
             return _('Boom! Team record for the past 30 days.')
-        elif self.expected_revenue and query_result['max_team_7'] < self.expected_revenue:
+        elif _is_lower_than_expected_revenue(query_result['max_team_7']):
             return _('Yeah! Best deal out of the last 7 days for the team.')
-        elif self.expected_revenue and query_result['max_user_31'] < self.expected_revenue:
+        elif _is_lower_than_expected_revenue(query_result['max_user_31']):
             return _('You just beat your personal record for the past 30 days.')
-        elif self.expected_revenue and query_result['max_user_7'] < self.expected_revenue:
+        elif _is_lower_than_expected_revenue(query_result['max_user_7']):
             return _('You just beat your personal record for the past 7 days.')
         elif query_result['count_user_closed_today'] == 5:
             return _('You\'re on fire! Fifth deal won today 🔥')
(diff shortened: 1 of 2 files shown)


alsh-odoo:
Hello @path-odoo :wave: 
Thanks for the fix!!
Since the issue is already present in sentry, I will let @maan-odoo to review this.

addons/crm/models/crm_lead.py
```diff
@@ -1229,13 +1229,13 @@ def _get_rainbowman_message(self):
 
         if query_result['count_user_closed_year'] == 1:
             return _('Go, go, go! Congrats for your first deal.')
-        elif self.expected_revenue and query_result['max_team_31'] < self.expected_revenue:
+        elif self.expected_revenue and (query_result['max_team_31'] or 0.0) < self.expected_revenue:
```
maan-odoo: Rather than changing the condition, it’s better to handle this directly in SQL.
Apply for all below

path-odoo: Hello :wave:, I've applied the suggested changes.
Could you please have a look?
Thanks :smile:!

maan-odoo (approved):
LGTM; Need to check if the behavior is expected or not with the CRM module team.

ushyme (approved):
LGTM, thanks for your work!
Minor cosmetic note: The PR description refers to `_is_lower_than_expected_revenue` (with underscore), but the implementation uses `is_lower_than_expected_revenue` (without underscore).

addons/crm/tests/test_crm_rainbowman.py
```diff
@@ -451,3 +451,24 @@ def test_leads_rainbowman_no_team(self):
                 'Yay, your first win from Xitter Post!',
                 'Benoit can still receive the message as neither he nor his team have a recorded win for this source',
             )
+
+    @users('sales_manager_casey')
+    def test_rainbowman_message_with_different_year(self):
```
tde-banana-odoo: Can't it be included in existing test, like just an extra entry or a test data change ? We already have way too much test lines for such a small features ^^

path-odoo: I've included this case in the existing test case.
Could you please have a look?

PR #244939, opened 2026-01-21
[FIX] website: improve carousel accessibility

robinlej:
[FIX] website: improve carousel accessibility
    
    To improve keyboard accessibility on carousels, the indicators container
    is not focusable anymore, and the indicators themselves now only have
    one focusable button at a time.
    The tab order is thus: previous button > active indicator > next button.
    (The previous and next button may both appear before.)
    You can still navigate among indicators with the left and right arrows,
    which also moves the focus to the newly targetted indicator.
    
    Note that other accessibility improvements remain to be done on the
    carousels (add a pause/play button on auto-sliding carousels, place the
    buttons before the carousel slide in the tab order, add some aria
    attributes (roledescription, live), adaptative labels...).

[FIX] website: make focus visible on carousel arrows
    
    When focusing manually (with tab / shift+tab) the previous/next arrows
    in carousels, it is hard to follow where the focus is, because there is
    no outline and the contrast is too small.
    
    This commit adds a specific outline if the button is `:focus-visible`
    (with both black and white to work on any background).

[IMP] website: pause carousel on focus within
    
    Carousels are paused on hover (or touchstart), but if the focus is
    inside one, it won't pause. That can make it hard to navigate within
    interactive carousels during the sliding interval (e.g. the dynamic
    products, which by default has 4 different products with 3 different
    focusable links/buttons).
    This commit applies the same behavior on focusin as Bootstrap's default
    on mouseover, and on focusout as on mouseout.

[IMP] website: support home/end keys on carousels

    When the focus is in a carousel, pressing the "home" key displays the
    first slide and pressing the "end" key displays the last slide.

task-5470023

### DIFF
--- a/addons/website/static/src/interactions/carousel/carousel_slider.js
+++ b/addons/website/static/src/interactions/carousel/carousel_slider.js
@@ -10,6 +10,17 @@ export class CarouselSlider extends Interaction {
             "t-on-slid.bs.carousel": this.onSlidCarousel,
             "t-on-focusin": () => window.Carousel.getInstance(this.el)?.pause(),
             "t-on-focusout": this.resumeCarouselCycling,
+            "t-on-keydown": (ev) => {
+                const hotkey = getActiveHotkey(ev);
+                if (/input|textarea/i.test(ev.target.tagName)) {
+                    return;
+                }
+                if (["home", "end"].includes(hotkey)) {
+                    const childToFocus = hotkey === "home" ? ":first-child" : ":last-child";
+                    ev.preventDefault();
+                    this.el.querySelector(`.carousel-indicators > ${childToFocus}`).click();
+                }
+            },
         },
         "img": {
             "t-on-load": this.computeMaxHeight,


addons/website/static/src/interactions/carousel/carousel_slider.js
```diff
@@ -86,6 +112,15 @@ export class CarouselSlider extends Interaction {
         this.loadItemsToAppear(); // Preload future items after a slide
     }
 
+    /**
+     * If the carousel should auto-slide and it has been paused, resume it.
+     */
+    resumeCarouselCycling() {
+        const bsCarousel = window.Carousel.getInstance(this.el);
```
duau-odoo: For consistency, shouldn't the variable be called `carouselBs` like in the `setup()` ?

addons/website/static/src/interactions/carousel/carousel_slider.js
```diff
@@ -76,6 +83,25 @@ export class CarouselSlider extends Interaction {
...
+        const nextActiveIndicatorEl = this.carouselIndicatorEls.item(ev.to);
+        // If before the slide, the focused element was another indicator, move
+        // the focus to the newly active indicator.
+        if (
+            document.activeElement.matches(
+                `.carousel-indicators > :is(button, li):not([data-bs-slide-to="${ev.to}"])`
+            )
+        ) {
+            nextActiveIndicatorEl.focus();
+        }
```
duau-odoo, line 116: This call grabs the focus from another carousel, shouldn't we check if the activeElement is in `this.el` ?

Steps to reproduce:
- Drop two carousel on the page
- Put the speed on the second carousel to 1sec so it's easier to test
- Focus on the indicators of the first carousel
- When the second carousel slides automatically, it will grab the focus

duau-odoo (changes requested):
Thanks for the PR, not much to say apart from the small bug, it seems a great improvement for accessibility.

duau-odoo (approved):
LGTM!

addons/website/static/src/interactions/carousel/carousel_slider.js
```diff
@@ -113,6 +115,15 @@ export class CarouselSlider extends Interaction {
         this.loadItemsToAppear(); // Preload future items after a slide
     }
 
+    /**
+     * If the carousel should auto-slide and it has been paused, resume it.
+     */
+    resumeCarouselCycling() {
+        const carouselBS = window.Carousel.getInstance(this.el);
+        if ([true, "carousel"].includes(carouselBS?._config.ride)) {
```
FrancoisGe: Could you check if `this.el.dataset.bsRide` === `carouselBS?._config.ride`. ?
If you don t want to check in the inter of bootstrap

robinlej: Done 👍

FrancoisGe:
@robodoo r+ rebase-ff

PR #244941, opened 2026-01-21
[FIX] website_sale: persist multiple attribute filters during pagination

abz-odoo:
Current behavior:
When a user selects multiple filters (attributes) that result in multiple pages of products, navigating to the second page causes some filters to be lost. Specifically, only the last selected attribute value is kept in the URL of the pager.

This happens because the `/shop` controller processes query parameters using a standard Python dictionary (**post). Since a dictionary cannot hold duplicate keys, an URL like `?attrib=1&attrib=2` is reduced to `{'attrib': '2'}`, losing all previous values.

Steps to reproduce:
1. Install `website_sale`.
2. Reduce "Products per Page" (e.g., to 4) to easily trigger pagination.
3. Go to the /shop page.
4. Select a first attribute (e.g., Color: White).
5. Select a second attribute (e.g., Size: M).
6. Ensure the result spans at least two pages.
7. Click on page "2".
8. Observation: The second attribute filter is lost, and the product list changes incorrectly.

Fix:
Ensure that `attribute_values` are stored as a list within the `url_args` passed to the pager. Since Odoo's `website.pager` uses `url_encode` internally, passing a list of values for a single key correctly generates repeated parameters in the resulting URL (e.g., `attrib=1&attrib=2`).

opw-4152637

### DIFF
--- a/addons/website_sale/controllers/main.py
+++ b/addons/website_sale/controllers/main.py
@@ -291,6 +291,7 @@ class WebsiteSale(payment_portal.PaymentPortal):
         attribute_value_ids = set(itertools.chain.from_iterable(attribute_value_dict.values()))
         if attribute_values:
             request.session['attribute_values'] = attribute_values
+            post['attribute_values'] = attribute_values
         else:
             request.session.pop('attribute_values', None)
 
--- a/addons/website_sale/static/tests/tours/website_sale_shop_filters.js
+++ b/addons/website_sale/static/tests/tours/website_sale_shop_filters.js
@@ -1,0 +1,33 @@
+import { registry } from "@web/core/registry";
+
+registry.category("web_tour.tours").add("shop_attribute_filters_remain_when_changing_page", {
+    url: "/shop",
+    steps: () => [
+        {
+            content: "Select first attribute value",
+            trigger: ".products_attributes_filters form.js_attributes input:eq(0)",
+            run: "click",
+            expectUnloadPage: true,
+        },
+        {
+            content: "Select last attribute value",
+            trigger: ".products_attributes_filters form.js_attributes input:eq(3)",
+            run: "click",
+            expectUnloadPage: true,
+        },
+        {
+            content: "Select page 2",
+            trigger: 'li a.page-link:contains("2")',
+            run: "click",
+            expectUnloadPage: true,
+        },
+        {
+            content: "First attribute value should be checked",
+            trigger: ".products_attributes_filters form.js_attributes input:eq(0):checked",
+        },
+        {
+            content: "Last attribute value should be checked",
+            trigger: ".products_attributes_filters form.js_attributes input:eq(3):checked",
+        },
+    ],
+});
--- a/addons/website_sale/tests/test_website_sale_product_filters.py
+++ b/addons/website_sale/tests/test_website_sale_product_filters.py
@@ -1,7 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo import Command
-from odoo.tests import tagged
+from odoo.tests import tagged, HttpCase
 
 from odoo.addons.product.tests.test_product_attribute_value_config import (
     TestProductAttributeValueCommon,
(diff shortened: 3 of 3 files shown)


abz-odoo:
wip: test incoming

addons/website_sale/controllers/main.py
```diff
@@ -303,6 +303,12 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
                 post['tags'] = None
                 tags = {}
 
+        # We need to transform attribute_value in list to prevent keeping only the first attribute
+        # ?attribute_values=1-2,3&attribute_value=4-5,6 will works with this
+        if attribute_values:
+            if len(attribute_values) > 1:
+                post['attribute_values'] = attribute_values
```
barracudapps: You can move the `post` up (under line 293 `request.session['attribute_values'] = attribute_values`) and revert its removal since commit 59fd420aa252b0416d3677c7bb6832bfecbfc9a7
No conditions required as already checked

barracudapps (changes requested):
Seen with R&D and PO, you can drop my fixup.
We won't try to remember the page from which the user came before clicking on a product. Indeed, as all filters are not saved (eg. `search`) the user will most likely not come back in the same behavior, even though the page is the same.

Yet, I think you can reuse the test from the fixup commit as somehow less complex.

Thanks for taking care of the Sales' scope 👍

addons/website_sale/static/tests/tours/website_sale_shop_filters.js
```diff
@@ -0,0 +1,37 @@
...
+            run: () => {
+                const attributes = Array.from(
+                    document.querySelectorAll(
+                        '.products_attributes_filters input[name="attribute_value"]'
+                    )
+                );
+                attributes.at(0).click();
+                attributes.at(-1).click();
+            },
+        },
```
ltinel, line 17: I find it cleaner to select the attribute values via triggers instead. Moreover, selecting an attribute value will reload the page, so it's possible that selecting the 2 attribute values without waiting in between doesn't work (which might explain why runbot is red :sweat_smile: )

```suggestion
        {
            content: "Select first attribute value",
            trigger: "form.js_attributes input:eq(0)",
            run: "click",
            expectUnloadPage: true,
        },
        {
            content: "Select last attribute value",
            trigger: "form.js_attributes input:eq(-1)",
            run: "click",
            expectUnloadPage: true,
        },
```

Optional: personally, I'd find the test even cleaner if we select the attribute values by name instead of by index.

addons/website_sale/static/tests/tours/website_sale_shop_filters.js
```diff
@@ -0,0 +1,37 @@
...
+        {
+            content: "Select page 2",
+            trigger: 'li a.page-link:contains("2")',
+            run: "click",
```
ltinel, line 21: Nit (I believe it's a best practice):
```suggestion
            run: "click",
            expectUnloadPage: true,
```

addons/website_sale/static/tests/tours/website_sale_shop_filters.js
```diff
@@ -0,0 +1,37 @@
...
+                const checkedAttributes = Array.from(
+                    document.querySelectorAll(
+                        '.products_attributes_filters input[name="attribute_value"]'
+                    )
+                ).filter((el) => el.checked);
+                if (checkedAttributes.length !== 2) {
+                    throw new Error("Expect 2 attributes checked, got " + checkedAttributes.length);
+                }
+            },
+        },
```
ltinel, line 31: I think it's better to use triggers here as well :)
```suggestion
        {
            content: "First attribute value should be checked",
            trigger: "form.js_attributes input:eq(0):checked",
        },
        {
            content: "Last attribute value should be checked",
            trigger: "form.js_attributes input:eq(-1):checked",
        },
```

addons/website_sale/tests/test_website_sale_product_filters.py
```diff
@@ -322,3 +322,11 @@ def test_newest_products_filter(self):
                 16,
                 "When displaying newest product templates, 16 unique templates should be shown",
             )
+
+    def test_shop_attribute_filters_remain_when_changing_page(self):
+        self.env['product.attribute'].search([]).write({'visibility': 'hidden'})
+        self.color_attribute.visibility = self.size_attribute.visibility = 'visible'
```
ltinel: Optional nit: I find it a bit more readable to split this assignment on 2 lines.

ltinel (changes requested):
Thanks for the fix :slightly_smiling_face: The fix in itself LGTM, but the test doesn't seem to pass (see comments)

ltinel (approved):
LGTM, thanks :slightly_smiling_face: 
@robodoo r+

PR #244942, opened 2026-01-21
[FIX] test_l10n_be_hr_payroll_account: fix the test of 13th month full leave

Faruk372742:
[FIX] test_l10n_be_hr_payroll_account: fix the test of 13th month full leave

Bug production: It is a runbot error occurred in test_13th_month_full_leave (version: master);
    1 - Install and update l10n_be_hr_payroll,l10n_be_hr_payroll_account,test_l10n_be_hr_payroll_account,hr_holidays,resource modules in localhost
    2 - Execute the test test_13th_month_full_leave and it will fail, also in the same class (Test13thMonth) many other tests were failing.

Bug cause:
    1 - In test test_13th_month_full_leave, for all year the sick day timeoff is allocated. (year: 2025), generate work entries for 2025 and expecting payslip amount must be 0 but it is not.
    2 - When I looked to _get_paid_amount -> _get_paid_amount_13th_month -> _compute_presence_prorated_fixed_wage: Some days in the year was not counted as sick day and since few days are not counted, the payslip amount is > 0.
    3 - By diving deeper: in function _get_calendar_days: date_to = min(date_max, leave.date_to.date()), from this line, I observed that leave.date_to is calculated 1 days before sometimes.
    4 - In hr.leave model -> _compute_date_from_to -> hour_from, hour_to = holiday._get_hour_from_to(holiday.request_date_from, holiday.request_date_to) -> In here hour_from and hour_to is (0,0) when the calendar (id = 1) standard 40h/week is used.
    5 - Since it returns (0,0) and converted to UTC in the next line, the date_to becomes 1 day before the expected date_to.
    6 - After finding that, the standard 40h/week calendar returns (0,0) for work start hour and work end hour, I looked to 'resource.calendar' model.
    7 - In function _get_default_attendance_ids, hour_from and hour_to returns 0,0 for typical 40 hours/week calendar which is not true.
    8 - This standard calendar with id = 1 is used in other 4 tests in class Test13thMonth and they were failing as well, now all of them are fixed.

Bug solution:
    1 - Fixing the hour_from and hour_to of standard 40h/week calendar from (0,0) -> (8, 16).

task - 5853076
Runbot Error

Current behavior before PR:

---


romaincarlier4 (commented):
Thanks for your work !
The investigation is quite accurate and correct, but the fix must be somewhere else.
The test is supposed to work with such a calendar :)
I think the fix must be in `_get_calendar_days`, where you should convert the date from/to of the leave to the timezone of the employee before taking the date

PR #244949, opened 2026-01-21
[IMP] mail: mark message as edited when removing attachments

Atovange:
With this commit, removing an attachment from a message will mark it as edited.

task-5357702


addons/account/tests/test_portal_attachment.py
alexkuhn, line 183: Could assert message body has `.o-mail-Message-edited`, for explicit test coverage.

alexkuhn (approved):
LGTM
I wonder if this shouldn't even been considered as a bug fix.
Having this change of behaviour in master is good at least.
@robodoo delegate+

PR #244951, opened 2026-01-21
[IMP] hr: allow sending bulk emails to employees from the Actions menu

nikj-odoo:
Before this commit, there was no easy way to send emails to multiple employees at once.

This change adds an option to send bulk emails directly from employee views making communication easier.

Task-5847145


yava-odoo (approved):
LGTM 🚀

PR #224976, opened 2025-09-01
[FIX] mail: several small discuss style fixes

alexkuhn:
Several small fixes to match with recent change of style of discuss app.

PR #235427, opened 2025-11-13
[FIX] mail: show warning when no mic permission in systray call menu

akus-odoo:
Call menu now shows a warning badge when mic permission have not been granted like in call action list. The warning is also shown in the call action dropdown list, inciting click to open the permission dialog.

This commit also makes other minor improvements to systray call menu:
- tracked action use active color of action. For example, muted mic uses red colored icon.
- Success actions now has correct style in dropdown mode (was missing, only danger had correct style)
- flash effect of the volume-up icon has been reduced.
- improve readability of the warning badge in all call actions in dark theme.
- Fix a race-condition bug where initial discuss call could show incorrect active tracked call action.
- Systray call menu has been moved to left-most systray item.

Task-5382871

PR #244954, opened 2026-01-21
[FIX] mail: danger/success effect for actions

maki-odoo:
PR #224976 fixes the demonstration of an action with a `DANGER` tag in the
dropdown menu in the action list. Such an action may not be located in the
dropdown, but rather in the quick action menu. This change ensures that the
action buttons with `DANGER` or `SUCCESS` tags are demonstrated properly in
either case.

Steps to reproduce:
- Open a document in the portal as a portal user.
- Send a message and hover over the message.
- The `delete` button is not visible in the action list.

### DIFF
--- a/addons/mail/static/src/core/common/action_list.scss
+++ b/addons/mail/static/src/core/common/action_list.scss
@@ -80,13 +80,22 @@
                 transform: translateY(.5px);
             }
         }
-        &.o-dropdown-item.btn-danger:not(.o-tag-JOIN_LEAVE_CALL) {
-            &.focus {
-                background-color: rgba($danger, 0.75);
-                color: white;
+
+        @each $status, $color in (danger: $danger, success: $success) {
+            &.o-dropdown-item.btn-#{$status}:not(.o-tag-JOIN_LEAVE_CALL) {
+                &.focus {
+                    background-color: rgba($color, 0.75);
+                    color: white;
+                }
+                &:not(.focus) {
+                    color: var(--o-mail-ActionList-#{$status}NoBgColor, #{$color});
+                }
             }
-            &:not(.focus) {
-                color: var(--o-mail-ActionList-dropdownDangerUnfocusedColor, #{$danger});
+            &.o-inline.btn-#{$status}:not(.o-tag-JOIN_LEAVE_CALL):not(.o-hasBtnBg) {
+                color: var(--o-mail-ActionList-#{$status}NoBgColor, #{$color});
+                &:active {
+                    background-color: inherit;
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


addons/mail/static/src/core/common/action_list.scss
alexkuhn: This style should not affect `.o-hasBtnBg` buttons:

Before this change (correct):
Image

After this change (incorrect):
Image

And also I would just keep
```css
color: var(--o-mail-ActionList-dropdownDangerUnfocusedColor, #{$danger});
```

The rest doesn't make the button look good on click:
Image

Just the change of opacity like the other buttons is more than enough.

addons/mail/static/src/core/common/action_list.scss
```diff
@@ -89,6 +89,12 @@
                 color: var(--o-mail-ActionList-dropdownDangerUnfocusedColor, #{$danger});
             }
         }
+        &.o-inline.btn-danger:not(.o-tag-JOIN_LEAVE_CALL):not(.o-hasBtnBg) {
+            color: var(--o-mail-ActionList-dropdownDangerUnfocusedColor, #{$danger});
+            &:active {
+                background-color: inherit;
+            }
+        }
```
alexkuhn: Same bug can be fixed for the `success` variant.

Also the message actions with danger / success variants have lower luminance on hover. I would bump them to .5:
Image

If we keep .35:
Image

With .5 on the danger variant:
Image

alexkuhn (approved):
LGTM otherwise.

@robodoo delegate+

addons/mail/static/src/core/common/action_list.scss
```diff
@@ -80,13 +80,22 @@
...
+                    color: white;
+                }
+                &:not(.focus) {
+                    color: var(--o-mail-ActionList-#{$status}NoBgColor, #{$color});
+                }
             }
-            &:not(.focus) {
-                color: var(--o-mail-ActionList-dropdownDangerUnfocusedColor, #{$danger});
+            &.o-inline.btn-#{$status}:not(.o-tag-JOIN_LEAVE_CALL):not(.o-hasBtnBg) {
+                color: var(--o-mail-ActionList-#{$status}NoBgColor, #{$color});
```
alexkuhn, line 95: And should change the ` --o-mail-ActionList-dropdown(Success/Danger)UnfocusedColor` in the `.dark.scss` bundle.

alexkuhn (commented):
Also https://github.com/odoo/odoo/pull/235427 in 19.1 introduced the Success variant of dropdown unfocused. I think it's worth backporting this color, as I suspect the readability is also poor in dark theme without it.

alexkuhn (commented):
@robodoo delegate+ r+

PR #244966, opened 2026-01-21
[IMP] base: ensure res.device.log during exception

thle-odoo:
When a device using the session is detected and a log needs to be inserted, we insert directly into the `res_device_log` table. Before this commit, if an exception occurred, the insertion was rolled back and we lost the log.
This commit ensures that the log is committed.
In addition, during retries, we authenticate the user each time. In order to avoid detecting a new device each time (and therefore inserting a log each time), it is necessary to save the session on disk.

### DIFF
--- a/odoo/addons/base/models/res_device.py
+++ b/odoo/addons/base/models/res_device.py
@@ -1,7 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 import logging
-from contextlib import nullcontext
 from datetime import datetime, timedelta
 
 from odoo import api, fields, models, tools
@@ -62,10 +61,8 @@ class ResDeviceLog(models.Model):
 
         if self.env.cr.readonly:
             self.env.cr.rollback()
-            cursor = self.env.registry.cursor(readonly=False)
-        else:
-            cursor = nullcontext(self.env.cr)
-        with cursor as cr:
+
+        with self.env.registry.cursor() as cr:
             cr.execute(SQL("""
                 INSERT INTO res_device_log (session_identifier, user_id, ip_address, user_agent, country, city, first_activity, last_activity, revoked)
                 VALUES (%(session_identifier)s, %(user_id)s, %(ip_address)s, %(user_agent)s, %(country)s, %(city)s, %(first_activity)s, %(last_activity)s, %(revoked)s)
@@ -80,7 +77,8 @@ class ResDeviceLog(models.Model):
                 last_activity=datetime.fromtimestamp(device['last_activity']),
                 revoked=False,
             ))
-        _logger.info('User %d inserts device log (%s)', user_id, session_identifier)
+            _logger.info('User %d inserts device log (%s)', user_id, session_identifier)
+            root.session_store.save(request.session)
 
     @api.autovacuum
     def _gc_device_log(self):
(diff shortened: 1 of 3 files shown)


mwath (commented):
Do you know what is the impact of this PR on PostgreSQL locks, especially under heavy load? I'm curious about how this will impact performance/concurrency.

An alternative approach could be to introduce a pre- and/or post-rollback hook(s). This might help mitigate any locking issues while maintaining the intended functionality. What do you think?

pivi-odoo (commented):
Is it intended for `cr` to be inconsistent if the `self.env.cr` is `readonly` or not?
When it's `readonly`, we get a *new* cursor, but if it was RW, it's the same cursor of the transaction.
So when we `commit`, what gets commited is different, in the case of a RW, it's also all the content before the call to `_update_device`.

Not sure if it's possible (and acceptable design wise), but I guess for this kind of logging, it would make more sense to do it in a postcommit and postrollback? This way you won't have to bifurcate when you have a RO or RW cursor, since in those hooks, you use your own cursor that can always be RW.
The INSERT shouldn't conflict with anything at first glance, it's just some data, and `user_id` isn't an FKey (it's an Integer), so no way it requires locking of any sort. (also in a post-..., everything is committed, so potential locks from the TX are released)

Ofc this proposition makes sense only if it's ok for `_update_device` to not do the logging *immediately*.

odoo/addons/base/models/res_device.py
```diff
@@ -80,7 +77,8 @@ def _update_device(self, request):
                 last_activity=datetime.fromtimestamp(device['last_activity']),
                 revoked=False,
             ))
-        _logger.info('User %d inserts device log (%s)', user_id, session_identifier)
+            _logger.info('User %d inserts device log (%s)', user_id, session_identifier)
+            root.session_store.save(request.session)
```
mwath, line 81: imo, saving the session after the commit would be preferable
```suggestion
        root.session_store.save(request.session)
```

mwath: In fact, it's a bit more complicated: even if it should never happen, either saving the session or the commit *could* fail.
So, what do we want in the worst case?
1. the device log is in the session on disk, but not in database
2. the device log is in the database, but not on the session

thle-odoo: I prefer:
> the device log is in the session on disk, but not in database

Because the feature is based on data on the filesystem.

mwath: Discussed together, it's better to save in session, and no big deal if we don't save it in db

mwath (approved):
LGTM, thank you, Thomas! 👍 
(and thanks PIVI for the review :yum:)

beledouxdenis:
robodoo r+

beledouxdenis:
robodoo retry

odoo/addons/base/models/res_device.py
```diff
@@ -62,10 +61,8 @@ def _update_device(self, request):
 
         if self.env.cr.readonly:
             self.env.cr.rollback()
-            cursor = self.env.registry.cursor(readonly=False)
-        else:
-            cursor = nullcontext(self.env.cr)
-        with cursor as cr:
+
+        with self.env.registry.cursor() as cr:
```
Julien00859, line 65: the previous code with the rollback was crazy

PR #244973, opened 2026-01-21
[FIX] purchase: allow cross-company payment matching

gteboul:
On bills, the `purchase_vendor_bill_id` field already allows matching with cross-company purchase orders. This commit extends this behavior to payment matching.

Steps to reproduce:
- Create a child company from a parent company.
- Create a purchase order in the child company.
- Create a bill in the parent company.
- In the bill’s payment matching, the child company’s purchase order should be available.

opw-5416947

### DIFF
--- a/addons/purchase/models/account_invoice.py
+++ b/addons/purchase/models/account_invoice.py
@@ -143,7 +143,8 @@ class AccountMove(models.Model):
             'res_model': 'purchase.bill.line.match',
             'domain': [
                 ('partner_id', 'in', (self.partner_id | self.partner_id.commercial_partner_id).ids),
-                ('company_id', 'in', self.env.company.ids),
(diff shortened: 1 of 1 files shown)


addons/purchase/models/account_invoice.py
```diff
@@ -143,7 +143,7 @@ def action_purchase_matching(self):
             'res_model': 'purchase.bill.line.match',
             'domain': [
                 ('partner_id', 'in', (self.partner_id | self.partner_id.commercial_partner_id).ids),
-                ('company_id', 'in', self.env.company.ids),
+                ('company_id', 'in', (self.env.company | self.env.company.child_ids).ids),
```
adwid: This will work with direct children only
If we go for this change, I guess we will expect all children levels

PR #244974, opened 2026-01-21
[IMP] calendar: time detection in calendar event title

frbin-odoo:
We to create a meeting by typing the time into the meeting title: e.g. '9am meeting'

This should:
-> uncheck all day
-> autofill start time
-> autofill end time if provided, otherwise set it to one hour after start

This behavior should only apply in the all day quick create form, and should cover a wide range of possible time formats, e.g.: 9h, 9H, 9:00, 09:00, 9 am, 9am, 21h, 21:30, 9-10, 9–11h

A regex was used to match times, time ranges, and time format suffixes. See the regex comments and tests for covered cases.

task-5380671

### DIFF
--- a/addons/calendar/models/calendar_event.py
+++ b/addons/calendar/models/calendar_event.py
@@ -3,9 +3,10 @@
 import itertools
 import logging
 import math
+import re
 import uuid
+from babel.dates import parse_time
 from datetime import datetime, timedelta, UTC
-from itertools import repeat
 from zoneinfo import ZoneInfo
 
 from markupsafe import Markup
@@ -27,7 +28,7 @@ from odoo.addons.calendar.models.calendar_recurrence import (
 from odoo.addons.calendar.models.utils import interval_from_events
 from odoo.tools.intervals import intervals_overlap
 from odoo.tools.translate import _
-from odoo.tools.misc import get_lang
+from odoo.tools.misc import get_lang, babel_locale_parse
 from odoo.tools import html2plaintext, html_sanitize, is_html_empty, single_email_re
 from odoo.exceptions import UserError, ValidationError
 
@@ -51,6 +52,29 @@ RRULE_TYPE_SELECTION_UI = [
     ('yearly', 'Yearly'),
     ('custom', 'Custom')
 ]
+
+# regex to match common ways to represent time
+# (9h, 9H, 9:00, 09:00, 9 am, 9am, 9a.m. 21h, 21:30, 9-10, 9-11h...)
+HOUR = r'(?:[01]?\d|2[0-3])'  # 0-23
+MINUTES = r'(?:[0-5]\d)'  # :00-:59
+SUFFIX = r'(?:h|hr|am|pm|a\.m\.?|p\.m\.?)'
+RANGE_DELIM = r'(?:-|–|to)'
+HOUR_WITH_MINUTES = rf'{HOUR}:{MINUTES}\s*{SUFFIX}?'
+HOUR_WITH_SUFFIX = rf'{HOUR}\s*{SUFFIX}'
+LOOSE_TIME = rf'(?:{HOUR_WITH_MINUTES}|{HOUR_WITH_SUFFIX}|{HOUR})'
+# Hour range uses lookahead, to make sure that a delimiter and a time follow.
+# Needed to disallow single hours 'meeting 5', but match ranges like '5-7pm'
+HOUR_RANGE = rf'{HOUR}(?=\s*{RANGE_DELIM}\s*{LOOSE_TIME})'
+TIME_REGEX = re.compile(
+    rf'''
+    (?:^|\s) # start of string or space
+    ({HOUR_WITH_MINUTES}|{HOUR_WITH_SUFFIX}|{HOUR_RANGE})  # Valid start times, with an optional range lookahead
+    (?:\s*{RANGE_DELIM}\s*({LOOSE_TIME}))? # Optional end time
+    (?:$|\s) # end of string or space
+    ''',
+    re.IGNORECASE | re.VERBOSE,
+)
+
 
 def get_weekday_occurence(date):
     """
@@ -590,6 +614,58 @@ class CalendarEvent(models.Model):
     def get_discuss_videocall_location(self):
         access_token = uuid.uuid4().hex
         return f"{self.get_base_url()}/{self.DISCUSS_ROUTE}/{access_token}"
+
+    @api.onchange('name')
+    def _onchange_name_extract_time(self):
+        for event in self:
+            if not event.env.context.get("is_quick_create_form") or not event.name or not event.allday:
+                continue
+
+            parsed_time = event._parse_time_from_title()
+            if not parsed_time:
+                continue
+
+            event.update({
+                'allday': False,
+                'start': parsed_time[0],
+                'stop': parsed_time[1],
+            })
+
+    def _parse_time_from_title(self):
+        """Extract datetime information from an event title.
+        Returns a tuple of (start_datetime_utc, end_datetime_utc) or None if no time is found."""
+        match = TIME_REGEX.search(self.name)
+        if not match:
+            return None
+
+        start_raw, end_raw = match.groups()
+
+        locale = babel_locale_parse(get_lang(self.env).code)
+        start_time = parse_time(start_raw, locale=locale)
+        end_time = parse_time(end_raw, locale=locale) if end_raw else None
+
+        # The user enters the meeting time in their own timezone, but we store it as UTC
+        tz_name = self.env.context.get('tz') or self.env.user.tz or 'UTC'
+        tz = ZoneInfo(tz_name)
+
+        # Combine event date and parsed time, using the user's timezone
+        start_local_dt = datetime.combine(self.start.date(), start_time, tzinfo=tz)
+        if end_time:
+            end_local_dt = datetime.combine(self.start.date(), end_time, tzinfo=tz)
+        else:
+            end_local_dt = start_local_dt + timedelta(hours=1)
+
+        # 9 to 5 should assume 9am to 5pm
+        if start_local_dt > end_local_dt and start_local_dt.hour <= 12:
+            end_local_dt += timedelta(hours=12)
+
+        # Handle overnight events edge case (e.g. 23:00 - 01:00)
+        if end_local_dt <= start_local_dt:
+            end_local_dt += timedelta(days=1)
+
+        start_utc = start_local_dt.astimezone(UTC).replace(tzinfo=None)
+        end_utc = end_local_dt.astimezone(UTC).replace(tzinfo=None)
+        return start_utc, end_utc
 
     # ------------------------------------------------------------
     # CRUD
--- a/addons/calendar/tests/test_calendar_event_time_extraction.py
+++ b/addons/calendar/tests/test_calendar_event_time_extraction.py
@@ -1,0 +1,171 @@
+from datetime import datetime
+from odoo.tests import TransactionCase, tagged, new_test_user, users
+
+
+@tagged('post_install', '-at_install')
+class TestCalendarEventTimeExtraction(TransactionCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user_utc_tz = new_test_user(
+            cls.env,
+            login='user_utc_tz',
+            name='Humphrey Appleby',
+            groups='base.group_user',
+            tz='UTC',
+        )
+        cls.user_brussels_tz = new_test_user(
+            cls.env,
+            login='user_brussels_tz',
+            name='Hercule Poirot',
+            groups='base.group_user',
+            tz='Europe/Brussels',
+        )
+        cls.user_japan_tz = new_test_user(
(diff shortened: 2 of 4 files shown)


addons/calendar/models/calendar_event.py
```diff
@@ -52,6 +53,29 @@
...
+RANGE_DELIM = r'(?:-|–|to)'
+HOUR_WITH_MINUTES = rf'{HOUR}:{MINUTES}\s*{SUFFIX}?'
+HOUR_WITH_SUFFIX = rf'{HOUR}\s*{SUFFIX}'
+LOOSE_TIME = rf'(?:{HOUR_WITH_MINUTES}|{HOUR_WITH_SUFFIX}|{HOUR})'
+# Hour range uses lookahead, to make sure that a delimiter and a time follow.
+# Needed to disallow single hours 'meeting 5', but match ranges like '5-7pm'
+HOUR_RANGE = rf'{HOUR}(?=\s*{RANGE_DELIM}\s*{LOOSE_TIME})'
+TIME_REGEX = re.compile(
+    rf'''
+    (?:^|\s) # start of string or space
```
tde-banana-odoo, line 70: Can't we use \b ?

addons/calendar/models/calendar_event.py
```diff
@@ -52,6 +53,29 @@
...
+HOUR = r'(?:[01]?\d|2[0-3])'  # 0-23
+MINUTES = r'(?:[0-5]\d)'  # :00-:59
+SUFFIX = r'(?:h|hr|am|pm|a\.m\.?|p\.m\.?)'
+RANGE_DELIM = r'(?:-|–|to)'
+HOUR_WITH_MINUTES = rf'{HOUR}:{MINUTES}\s*{SUFFIX}?'
+HOUR_WITH_SUFFIX = rf'{HOUR}\s*{SUFFIX}'
+LOOSE_TIME = rf'(?:{HOUR_WITH_MINUTES}|{HOUR_WITH_SUFFIX}|{HOUR})'
+# Hour range uses lookahead, to make sure that a delimiter and a time follow.
+# Needed to disallow single hours 'meeting 5', but match ranges like '5-7pm'
+HOUR_RANGE = rf'{HOUR}(?=\s*{RANGE_DELIM}\s*{LOOSE_TIME})'
```
tde-banana-odoo, line 67: lookahead are really bad from performance POV ...  Can't you have something like ``TIME_PART = rf'(?:{HOUR}:{MINUTES}|{HOUR}\s*{SUFFIX}|{HOUR})'`` and then ``(?:RANGE_DELIM TIME)?``, to somehow remove the lookahead ? I hate regexes, so maybe not, but feels like it can be simplified.

Something like
```
    \b
    ({TIME_PART})                 # Start Time
    (?:                           # Optional range group
        {RANGE_DELIM}
        ({TIME_PART})             # Group 2: End Time
    )?
    \b
```

addons/calendar/models/calendar_event.py
```diff
@@ -583,6 +607,84 @@ def get_discuss_videocall_location(self):
         access_token = uuid.uuid4().hex
         return f"{self.get_base_url()}/{self.DISCUSS_ROUTE}/{access_token}"
 
+    @api.onchange('name')
+    def _onchange_name_extract_time(self):
+        for event in self:
+            if not event.env.context.get("is_quick_create_form") or not event.name or not event.allday:
+                return
```
tde-banana-odoo: If you loop, then don't break the loop ^^
```suggestion
                continue
```

addons/calendar/models/calendar_event.py
```diff
@@ -583,6 +607,84 @@ def get_discuss_videocall_location(self):
...
+
+            parsed_time = event._parse_time_from_title(event.name)
+            if not parsed_time:
+                return
```
tde-banana-odoo: Same here
```suggestion
                continue
```

addons/calendar/models/calendar_event.py
```diff
@@ -583,6 +607,84 @@ def get_discuss_videocall_location(self):
...
+            event.allday = False
+
+            # The user enters the meeting time in their own timezone, but we store it as UTC
+            tz_name = self.env.context.get('tz') or self.env.user.partner_id.tz or 'UTC'
+            self_tz = event.with_context(tz=tz_name)
+
+            # Get the event date in the user's timezone
+            start_local = end_local = fields.Datetime.context_timestamp(self_tz, fields.Datetime.from_string(event.start))
+
+            # Handle day overflow like 11pm - 1am
```
tde-banana-odoo: Would do it (if necessary, a bit unclear to me) in _parse_time_from_title

addons/calendar/models/calendar_event.py
```diff
@@ -583,6 +607,84 @@ def get_discuss_videocall_location(self):
...
+            if start_time.hour > end_time.hour or \
+                (
+                    start_time.hour == end_time.hour and
+                    start_time.minute > end_time.minute
+                ):
+                end_local += timedelta(days=1)
+
+            new_start_utc = self._apply_time_in_tz(start_local, start_time)
```
tde-banana-odoo: Have to say this seems a bit complicated ... somehow in my head, you would have: parsing title information, giving you start_dt and end_dt in user's TZ using localize, then move them using astimezone. Not sure why we compute datetimes, then replace hour/minute/second, then do astimezone, managing +1 day manually ^^ code should be simpler in my opinion.

addons/calendar/tests/test_calendar_event_time_extraction.py
```diff
@@ -0,0 +1,183 @@
...
+            result = event._parse_time_from_title(title)
+
+            self.assertIsNotNone(result, f"Failed to parse time from: {title}")
+            self.assertEqual(result[0].hour, start_h, f"Wrong start hour for: {title}")
+            self.assertEqual(result[0].minute, start_m, f"Wrong start minute for: {title}")
+            self.assertEqual(result[1].hour, end_h, f"Wrong end hour for: {title}")
+            self.assertEqual(result[1].minute, end_m, f"Wrong end minute for: {title}")
+
+    def test_parse_time_from_title_ranges(self):
+        test_cases = [
```
tde-banana-odoo: Don't forget to check with starting words / end boundaries, like "Meeting 9-10!"

addons/calendar/tests/test_calendar_event_time_extraction.py
```diff
@@ -0,0 +1,183 @@
...
+            '-11 training',
+            '9--11 workshop',
+        ]
+
+        for title in test_cases:
+            event = self.CalendarEvent.new({'name': title})
+            result = event._parse_time_from_title(title)
+            self.assertIsNone(result, f"Should not find time in: {title}")
+
+    def get_event_times_in_user_tz(self, event):
```
tde-banana-odoo: Don't understand this, seems this is adding layers on test that should be as low level as possible. Instead create / use a test user with a timezone set on it and check results while giving an expected result explicitly. Don't have "variable" tests, as in the end we don't know what you are really testing :) .

addons/calendar/tests/test_calendar_event_time_extraction.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self.assertFalse(event.allday)
+        self.assertEqual(start_tz.hour, 14, "Start hour should be 14")
+        self.assertEqual(start_tz.minute, 30, "Start minute should be 30")
+        self.assertEqual(stop_tz.hour, 16, "Stop hour should be 16")
+        self.assertEqual(stop_tz.minute, 0, "Stop minute should be 0")
+        self.assertEqual(start_tz.date(), datetime(2024, 1, 15).date(), "Start date should remain unchanged")
+        self.assertEqual(stop_tz.date(), datetime(2024, 1, 15).date(), "Stop date should remain unchanged")
+
+    def test_onchange_name_without_quick_create_context(self):
```
tde-banana-odoo: Probably to merge with next test in a "no_parsing" test (to limit a bit code duplication)

addons/calendar/tests/test_calendar_event_time_extraction.py
```diff
@@ -0,0 +1,183 @@
...
+    def get_event_times_in_user_tz(self, event):
+        timezone = self.env.context.get('tz') or self.env.user.partner_id.tz or 'UTC'
+        self_tz = event.with_context(tz=timezone)
+
+        start_tz = fields.Datetime.context_timestamp(self_tz, fields.Datetime.from_string(event.start))
+        stop_tz = fields.Datetime.context_timestamp(self_tz, fields.Datetime.from_string(event.stop))
+
+        return start_tz, stop_tz
+
+    def test_onchange_name_with_quick_create_context(self):
```
tde-banana-odoo: Feels like this test and the next one could be merged (simplify test suite if possible). You may do a multi-create if you want to test several things (and then provide a list of expected results and use `subTest`).

addons/calendar/tests/test_calendar_event_time_extraction.py
```diff
@@ -0,0 +1,183 @@
...
+        event = self.CalendarEvent.with_context(is_quick_create_form=True).new({
```
tde-banana-odoo: As suggested above, use a real test user with a timezone set on it, and check timezone diff is effectively applied.

tde-banana-odoo (commented):
Zboing, thanks for your work ! Added some comments on regex (beware I hate those and might be wrong) / onchange, then some remarks to try to simplify / make tests a bit more explicit :) 

Cheers !

addons/calendar/models/calendar_event.py
frbin-odoo: There is a slight difference, so it depends on how restrictive we want the regex to be.

The current version explicitly checks for start/end or whitespace -> the more restrictive option
`\b` checks for a word boundary, meaning anything that does not match `\w` ([a-zA-Z0-9_])

The main difference is special characters. Using `\b` allows extra matches like:
- 'Meeting 9-10!'
- 'Meeting (9am)'

But also leads to potentially strange matches like:
- 'Test 19-3-2026"

In general, I preferred more restrictive rules where possible - more false negatives at the cost of less false positives seem like a better option to me in this case.

Let me know which you prefer :)

addons/calendar/models/calendar_event.py
frbin-odoo: The issue with a simplified solution is that it can match a single number which could be an hour:
`meeting 5` -> shouldn't match as a time
The lookahead ensures, that if no suffix or minutes are present, we have a full time range.

It is of course possible to do this without the lookahead, but doing so would likely require more capturing groups making the parsing later more complicated.
For example, we could check for a full time range separately from `{HOUR_WITH_MINUTES}|{HOUR_WITH_SUFFIX}`, but we'd end up with different capturing groups for the two cases, meaning we'd have to differentiate between the cases when parsing.
With the lookahead, we always have one group for start_time, and optionally one for end_time

Does the lookahead really affect performance that much?
- I wouldn't expect the meeting titles to be very long strings -> should be negligible
- Even for longer strings, we are only looking for the characters immediately after the first time part - we don't scan the entire string

addons/calendar/models/calendar_event.py
frbin-odoo: I refactored the flow. Hopefully it's a bit easier to read now.
I moved all the logic dealing with time/localization to the parse method, so that the onchange only gets the resulting date and updates the record.

In the _parse method:
- parse the matched groups with babel
- add a date from the event to the time
- handle specific edge cases
- convert back to UTC

addons/calendar/tests/test_calendar_event_time_extraction.py
frbin-odoo: `Meeting 9-10!` is not supported with the current word boundaries, but would work with '\b'.
See other comments for potential tradeoffs

frbin-odoo (commented):
Hello @tde-banana-odoo !
I refactored the tests as suggested, merging some of them together. 
I also added some users with different time-zones. Most of the tests now work with the 'utc' user, to test the basic behavior, and one of the tests checks the behavior of the other timezones (including Japan's UTC+9 which can easily cross midnight for morning meetings)

addons/calendar/models/calendar_event.py
tde-banana-odoo: Both are ok to me to be honest :) if it is a choice to be more restrictive it is ok for me, wanted to note it.  I suppose people won't type themselves "meeting 9-10!" (more something you could extract from an email for example). As it is limited to onchange in the name, I guess it is ok. Up to you :) .

addons/calendar/models/calendar_event.py
tde-banana-odoo: I am clearly not an expert of regexes, so difficult to say directly ^^ I already had the case where lookaheads exploded but that was in more complex strings (small html blobs). Here indeed titles should be short. So would say if you prefer the simple lookahead that's ok for me, at least it was discussed :) .

tde-banana-odoo (approved):
Answered ! For parsing, it is indeed way cleaner that way :) so LGTM, letting you the final decision on implementation. Once ok ping me we can r+ :) .

tde-banana-odoo:
@robodoo r+

Zboing !

PR #244989, opened 2026-01-21
[IMP] web_tour: allow macro delay only for wanted listed tours

pipu-odoo:
We noticed that allowing a delay between each step of the tours
(in the macro engine) leads to indeterministic behavior within the
tours, such as:
- letting animationframes to be calculated
- waiting for the RPC response, which sometimes arrives, sometimes not
within 50ms.

Most of the time, a tour is indeterministic because it
takes a wrong way by pointing to overly general triggers.
Removing this (arbitrary) delay is the final step to ensure the macro
engine always works the same way.

Find trigger => Do Action => Find trigger => Do Action => ...
No more delay between each step of the process.

With this commit, by default, the delay is removed in all tours by
default but it is necessary to tag the tours which are still
indeterministic.

### DIFF
--- a/addons/web/static/src/core/macro.js
+++ b/addons/web/static/src/core/macro.js
@@ -5,6 +5,7 @@ import { validate } from "@odoo/owl";
 const macroSchema = {
     name: { type: String, optional: true },
     timeout: { type: Number, optional: true },
+    allowDelayToRemove: { type: Boolean, optional: true },
     steps: {
         type: Array,
         element: {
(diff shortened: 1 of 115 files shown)


addons/web/static/src/core/macro.js
```diff
@@ -119,6 +119,10 @@ export class Macro {
             const step = this.steps[this.currentIndex];
             const timeoutDelay = step.timeout || this.timeout || 10000;
             const executeStep = async () => {
+                // To be remove ASAP because it allows non deterministic behaviors.
+                if (this.allowDelay) {
+                    await delay(50);
```
aab-odoo: why moved here?

addons/web_tour/static/src/js/tour_automatic/tour_automatic.js
```diff
@@ -119,6 +119,7 @@ export class TourAutomatic {
         this.macro = new Macro({
             name: this.name,
             steps: macroSteps,
+            allowDelay: this.config.allowDelay,
```
aab-odoo: How will this be used?

addons/web/static/src/core/macro.js
pipu-odoo: to use this.

PR #244991, opened 2026-01-21
[FIX] pos_online_payment: automatically close payment QR on order completion

meji-odoo:
Steps:
---
- Configure online payment on the POS configuration.
- Open a POS session and the customer display.
- Add a product and an online payment line.
- Validate the order and complete the payment via QR code.

Issue:
---
- The order is finalized, but the payment QR code remains visible on the customer display.
- The QR popup must be closed manually every time.

Cause:
---
- The customer display popup lifecycle depends on `onlinePaymentData`.
- This data was not cleared when the order was finalized.

Fix:
---
- Clear `onlinePaymentData` when the order is completed.

task-5502344

---


davidmonnom (approved):
@robodoo r+

PR #244992, opened 2026-01-21
[IMP] sale: add % delivered quantity

lkerroum:
In many cases, where you manually set delivered quantities on your SO, it would be easier to do it with a % of completion instead. In this commit, we add an optional field to set it. 

task-5387190

### DIFF
--- a/addons/sale/models/sale_order_line.py
+++ b/addons/sale/models/sale_order_line.py
@@ -288,6 +288,9 @@ class SaleOrderLine(models.Model):
         readonly=False,
         copy=False,
     )
+    qty_delivered_percent = fields.Float(
+        string="% Delivered", compute="_compute_qty_delivered_percent", readonly=False
+    )
 
     # Analytic & Invoicing fields
     qty_invoiced = fields.Float(
@@ -1015,6 +1018,19 @@ class SaleOrderLine(models.Model):
         for so_line in self:
             if not so_line.qty_delivered or so_line in delivered_qties:
                 so_line.qty_delivered = delivered_qties[so_line]
+
+    @api.depends("qty_delivered", "product_uom_qty")
+    def _compute_qty_delivered_percent(self):
+        for line in self:
+            if line.product_uom_qty:
+                line.qty_delivered_percent = 100 * (line.qty_delivered / line.product_uom_qty)
+            else:
+                line.qty_delivered_percent = 100
+
+    @api.onchange("qty_delivered_percent")
+    def _onchange_qty_delivered_percent(self):
+        for line in self:
+            line.qty_delivered = (line.qty_delivered_percent * line.product_uom_qty) / 100
 
     @api.depends("qty_delivered")
     @api.depends_context("accrual_entry_date")
@@ -1819,7 +1835,7 @@ class SaleOrderLine(models.Model):
         self.ensure_one()
 
         billable_lines = self.order_id.order_line.filtered(
-            lambda line: line.product_type != 'combo' and self._is_line_in_section(line)
+            lambda line: line.product_type != "combo" and self._is_line_in_section(line)
         )
 
         if display_taxes:
--- a/addons/sale/views/sale_order_views.xml
+++ b/addons/sale/views/sale_order_views.xml
@@ -780,6 +780,14 @@
                                     readonly="is_downpayment"/>
                                 <widget name="simple_qty_at_date_widget" width="20px"/>
                                 <field
+                                    name="qty_delivered_percent"
+                                    decoration-info="(not display_type and invoice_status == 'to invoice')"
+                                    decoration-bf="(not display_type and invoice_status == 'to invoice')"
+                                    readonly="qty_delivered_method != 'manual' or is_downpayment or product_uom_qty == 0"
+                                    column_invisible="parent.state != 'sale'"
+                                    optional="hide"
+                                />
+                                <field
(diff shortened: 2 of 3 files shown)


addons/sale/models/sale_order_line.py
```diff
@@ -238,6 +238,12 @@ class SaleOrderLine(models.Model):
         default=0.0,
         digits='Product Unit',
         store=True, readonly=False, copy=False)
+    qty_delivered_percent = fields.Float(
+        string="% Delivered",
+        compute='_compute_qty_delivered_percent',
+        inverse='_inverse_qty_delivered_percent',
+        readonly=False, copy=False,
```
livaios: readonly not needed, if you define an inverse function it is by default editable
```suggestion
        copy=False,
```

addons/sale/models/sale_order_line.py
```diff
@@ -954,6 +960,22 @@ def _compute_qty_delivered(self):
...
+        'product_uom_qty',
+        )
+    def _compute_qty_delivered_percent(self):
+        for line in self:
+            if line.product_uom_qty:
+                line.qty_delivered_percent = 100 * (line.qty_delivered / line.product_uom_qty)
+            else:
+                line.qty_delivered_percent = 100
+
+    @api.onchange('qty_delivered_percent')
```
livaios: Not needed for inverse function
```suggestion
```

livaios (changes requested):

addons/sale/models/sale_order_line.py
livaios: Hello @lkerroum 
It seems like the desired behavior was for it indeed to be an onchange function so in that case you need to remove the inverse function from the field, rename the function to `_onchange_qty_delivered_percent` and add the readonly back to the field

addons/sale/models/sale_order_line.py
```diff
@@ -238,6 +238,12 @@ class SaleOrderLine(models.Model):
         default=0.0,
         digits='Product Unit',
         store=True, readonly=False, copy=False)
+    qty_delivered_percent = fields.Float(
+        string="% Delivered",
+        compute='_compute_qty_delivered_percent',
+        readonly=False,
```
livaios, line 288: since we removed the inverse we now need to add the `store=True` again 🙈

addons/sale/models/sale_order_line.py
```diff
@@ -238,6 +238,10 @@ class SaleOrderLine(models.Model):
         default=0.0,
         digits='Product Unit',
         store=True, readonly=False, copy=False)
+    qty_delivered_percent = fields.Float(
+        string="% Delivered",
+        compute='_compute_qty_delivered_percent',
+        store=True, readonly=False, copy=False)
```
livaios: nitpicking but formatting 😊 
```suggestion
        store=True,
        readonly=False,
        copy=False,
    )
```

livaios (approved):
LGTM 👍 

Good work on this one!

addons/sale_project/models/product_template.py
```diff
@@ -12,7 +12,7 @@ def _selection_service_policy(self):
         service_policies = [
             # (service_policy, string)
             ('ordered_prepaid', _('Prepaid/Fixed Price')),
-            ('delivered_manual', _('Based on Delivered Quantity (Manual)')),
+            ('delivered_manual', _('Based on Delivered (Manual)')),
```
xavierbol: Maybe `Based on Delivery (Manual)` will be clearer. 🤔 🙂 

```suggestion
            ('delivered_manual', _('Based on Delivery (Manual)')),
```

xavierbol (commented):

Apart of that, LGTM. 🙂

xavierbol (approved):
LGTM, thanks for your work! 🙂

addons/sale/views/sale_order_views.xml
```diff
@@ -779,6 +779,14 @@
                                     decoration-bf="(not display_type and invoice_status == 'to invoice')"
                                     readonly="is_downpayment"/>
                                 <widget name="simple_qty_at_date_widget" width="20px"/>
+                                <field 
+                                    name="qty_delivered_percent"
+                                    decoration-info="(not display_type and invoice_status == 'to invoice')"
+                                    decoration-bf="(not display_type and invoice_status == 'to invoice')"
+                                    readonly="qty_delivered_method != 'manual' or is_downpayment"
+                                    column_invisible="parent.state != 'sale'"
+                                    optional="hide"
```
Feyensv: I'd also put the field readonly (or even invisible) when the `product_uom_qty` is 0, as modifying the percent on those lines won't have any impact (nor has much meaning).

addons/sale/views/sale_order_views.xml
```diff
@@ -779,6 +779,14 @@
                                     decoration-bf="(not display_type and invoice_status == 'to invoice')"
                                     readonly="is_downpayment"/>
                                 <widget name="simple_qty_at_date_widget" width="20px"/>
+                                <field
```
Feyensv: Noticed while testing it locally that you have an empty space after `<field`, I'd suggest configuring your editor to display (or even automatically removing) those.

Feyensv (commented):
2 little comments, thanks for the work here 👍

Feyensv (approved):
robodoo r+

PR #244993, opened 2026-01-21
[IMP] website_forum: remove search_count from pager

JKE-be:
In production, the search count take more than 100ms while it is only
necessary to compute the number of page...

To improve performance, the search count has been removed, and the pager now
displays a fixed range of 10 pages without showing the boundaries.

The existing `scope` parameter is reused to handle this behavior. In stable,
a negative value is used to explicitly indicate that the parameter should be
reused.

### DIFF
--- a/addons/website_forum/controllers/website_forum.py
+++ b/addons/website_forum/controllers/website_forum.py
@@ -141,11 +141,18 @@ class WebsiteForum(WebsiteProfile):
             **post
         )
 
+        pager_scope = 10
+
+        options['per_page'] = self._post_per_page
+        options['scope'] = pager_scope
+        options['current_page'] = page
+
         slug = request.env['ir.http']._slug
         question_count, details, fuzzy_search_term = request.website._search_with_fuzzy(
             "forum_posts_only", search, limit=page * self._post_per_page, order=sorting, options=options)
+
         question_ids = details[0].get('results', Post)
-        question_ids = question_ids[(page - 1) * self._post_per_page:page * self._post_per_page]
+        question_ids = question_ids[:self._post_per_page]
 
         if not forum:
             url = '/forum/all'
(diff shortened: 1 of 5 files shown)


addons/website_forum/controllers/website_forum.py
```diff
@@ -90,7 +90,7 @@ def sitemap_forum(env, rule, qs):
 
     def _get_forum_post_search_options(self, forum=None, tag=None, filters=None, my=None, create_uid=False, include_answers=False, **post):
         return {
-            'allowFuzzy': not post.get('noFuzzy'),
+            'allowFuzzy': False,
```
h4818: I'm not sure this is going to solve it (entirely).
The problem search query that you shared with me
`WHERE unaccent('overiw') <% unaccent("forum_post"."content") OR unaccent('overiw') <% unaccent("forum_post"."name")`

only occurs with the javascript autocomplete search. It doesn't happen when hitting enter due to https://github.com/odoo/odoo/blob/8e9c2788225aa1e4b483fc082c7bd5ebd7de4b1b/addons/website_forum/controllers/website_forum.py#L95

addons/website_forum/controllers/website_forum.py
```diff
@@ -162,22 +169,23 @@ def questions(self, forum=None, tag=None, page=1, filters='all', my=None, sortin
 
         pager = tools.lazy(lambda: request.website.pager(
             url=url, total=question_count, page=page, step=self._post_per_page,
-            scope=5, url_args=url_args))
+            scope=-pager_scope, url_args=url_args))  # scope negative, to force to use it (in stable)
 
         values = self._prepare_user_values(forum=forum, searches=post)
+
```
h4818: unnecessary

PR #244994, opened 2026-01-21
[FIX] account_edi_ubl_cii,account_peppol: better filter embed attachment

roto-odoo:
[FIX] account_edi_ubl_cii,account_peppol: better filter embed attachment

When you send an invoice by peppol with extra attachments, we want to make sure to only embed the extra attachments (not the xml nor the original PDF). To do this, a filter is applied on the attachments, this commit improves the filter to handle translations

no-task


smetl (approved):
@robodoo fw=no
@robodoo r+

PR #245007, opened 2026-01-21
[IMP] html_editor, *: substitute emojis on the fly

zel-odoo:
1. emojis should be substituted on the fly as the user types.
2. if the user presses the backspace key or undo next to an emoji, it should revert back to text

task-5478835

---

### DIFF
--- a/addons/html_editor/static/src/main/emoji_plugin.js
+++ b/addons/html_editor/static/src/main/emoji_plugin.js
@@ -1,6 +1,7 @@
 import { Plugin } from "@html_editor/plugin";
-import { EmojiPicker } from "@web/core/emoji_picker/emoji_picker";
+import { EmojiPicker, loadEmoji, loader } from "@web/core/emoji_picker/emoji_picker";
 import { _t } from "@web/core/l10n/translation";
+import { isContentEditable, isTextNode } from "@html_editor/utils/dom_info";
 
 /**
  * @typedef { Object } EmojiShared
@@ -13,6 +14,8 @@ export class EmojiPlugin extends Plugin {
     static shared = ["showEmojiPicker"];
     /** @type {import("plugins").EditorResources} */
     resources = {
+        delete_backward_overrides: this.handleDeleteBackward.bind(this),
+        input_handlers: this.detect.bind(this),
         user_commands: [
             {
                 id: "addEmoji",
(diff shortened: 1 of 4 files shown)


addons/html_editor/static/src/main/emoji_plugin.js
```diff
@@ -30,11 +33,77 @@ export class EmojiPlugin extends Plugin {
...
+
+    async prepareEmojiData() {
+        const { emojis } = await loadEmoji();
+        const emojiDict = {};
+        for (const emoji of emojis) {
+            for (const source of [...emoji.shortcodes, ...emoji.emoticons]) {
+                emojiDict[source] = emoji;
+            }
+        }
+        return emojiDict;
```
alexkuhn: That's the same `emojiDict` to prepare all the time, on all 1500 emojis.
Could be performed in `loader.loaded` as side-effect of loading emoji, same as `emojiValueToShortcodes`. This is already looping at least once on all emojis, so adding the `[source] = emoji` map is cheap.

Nitpick: I would advice using a `Map()` for reduced memory pressure.

alexkuhn (approved):
LGTM otherwise
@robodoo delegate+

addons/html_editor/static/src/main/emoji_plugin.js
```diff
@@ -30,11 +33,67 @@ export class EmojiPlugin extends Plugin {
...
-    setup() {
+    async setup() {
+        await super.setup();
         this.overlay = this.dependencies.overlay.createOverlay(EmojiPicker, {
             hasAutofocus: true,
             className: "popover",
         });
+        this.match = undefined;
+        await loadEmoji();
+        this.emojiDict = loader.loaded?.emojiSourceToEmoji ?? new Map();
```
alexkuhn: I don't think it's worth to have an attribute here.
This is more prone to not being properly synced to `loaded.loaded`. Maybe a getter if we feel we want a more readable accessor.

Note that `loaded` is filled even when the loader fails: this is added in a `finally`. So I don't think you need to `loaded?.emojiSourceToEmoji ?? new Map()`.

addons/html_editor/static/src/main/emoji_plugin.js
```diff
@@ -30,11 +33,70 @@ export class EmojiPlugin extends Plugin {
...
             hasAutofocus: true,
             className: "popover",
         });
+        this.match = undefined;
+        await loadEmoji();
+    }
+
+    get emojiDict() {
+        return loader.loaded.emojiSourceToEmoji;
+    }
```
alexkuhn, line 48: I don't know whether `resources` can be invoked when `setup()` hasn't be fully resolved, but if that's the case then this could explain why runbot fails and we need to guard `loaded?.` in the end.

PR #245011, opened 2026-01-21
[FIX] html_editor: fix blockquote behavior

sbbh-odoo:

- Pressing Backspace inside a blockquote that has visible content but no text content removes the blockquote instead of the content. The content (image or table) gets moved outside of the blockquote.
- Trailing BR was kept after tables because tables are marked as unsplittable blocks. This left unnecessary BR nodes after tables in blockquote.
- Lists could not be created inside a `blockquote`.

- Backspace removes the inner content first when blockquote contains nodes.
- Trailing BR is removed when placed table inside blockquote. Cursor can still be placed at the edge of the table without
  requiring a BR anchor.
- Lists can be created directly inside a blockquote.

task-5864080

---

### DIFF
--- a/addons/html_editor/static/src/core/dom_plugin.js
+++ b/addons/html_editor/static/src/core/dom_plugin.js
@@ -406,15 +406,19 @@ export class DomPlugin extends Plugin {
                     (node) => {
                         // Don't remove the last BR in cases where the
                         // previous sibling is an unsplittable block
-                        // (i.e. a table, a non-editable div, ...)
-                        // to allow placing the cursor after that unsplittable
-                        // element. This can be removed when the cursor
-                        // is properly handled around these elements.
+                        // (i.e. a non-editable div, ...) to allow placing the
+                        // cursor after that unsplittable element.
+                        // Tables are exception because the cursor can be
+                        // places directly at the edge of the table, so the
+                        // trailing BR is not needed.
+                        // This can be removed when the cursor is properly
+                        // handled around these elements.
                         const previousSibling = node.previousSibling;
                         return (
                             previousSibling &&
                             isBlock(previousSibling) &&
-                            this.dependencies.split.isUnsplittable(previousSibling)
+                            this.dependencies.split.isUnsplittable(previousSibling) &&
+                            previousSibling.nodeName !== "TABLE"
                         );
                     },
                 ]);
--- a/addons/html_editor/static/src/main/font/font_plugin.js
+++ b/addons/html_editor/static/src/main/font/font_plugin.js
@@ -2,7 +2,7 @@ import { Plugin } from "@html_editor/plugin";
 import { isBlock, closestBlock } from "@html_editor/utils/blocks";
 import { fillEmpty, unwrapContents } from "@html_editor/utils/dom";
 import { leftLeafOnlyNotBlockPath } from "@html_editor/utils/dom_state";
-import { isRedundantElement, isVisibleTextNode } from "@html_editor/utils/dom_info";
+import { isEmptyBlock, isRedundantElement, isVisibleTextNode } from "@html_editor/utils/dom_info";
 import {
     closestElement,
     createDOMPathGenerator,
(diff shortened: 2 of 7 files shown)


addons/html_editor/static/src/main/font/font_plugin.js
```diff
@@ -473,7 +473,7 @@ export class FontPlugin extends Plugin {
             return;
         }
         // Check if cursor is inside an empty heading, blockquote or pre.
-        if (!closestHandledElement || closestHandledElement.textContent.length) {
+        if (!closestHandledElement || !isEmptyBlock(closestHandledElement)) {
```
deso-odoo, line 476: Why not `isEmpty`? :thinking:

sbbh-odoo: Because `isEmpty` only checks if `innerHTML` is "" or `<br>`,
For example, `<blockquote><strong>[]</strong></blockquote>` is treated as non-empty, even though it is visually empty, causing Backspace to remove the blockquote instead of converting it to the base container.

addons/html_editor/static/src/core/dom_plugin.js
deso-odoo, line 422:
```suggestion
// Don't remove the last BR in cases where the
// previous sibling is an unsplittable block
// (i.e. a non-editable div, ...) to allow placing the
// cursor after that unsplittable element.
// Tables are exception because the cursor can
// be places directly at the edge of the table, so
// the trailing BR is not needed.
// This can be removed when the cursor
// is properly handled around these elements.
const previousSibling = node.previousSibling;
return (
previousSibling &&
isBlock(previousSibling) &&
(this.dependencies.split.isUnsplittable(previousSibling) && previousSibling.nodeName !== "TABLE")
);
```

addons/html_editor/static/src/core/dom_plugin.js
```diff
@@ -406,15 +406,19 @@ export class DomPlugin extends Plugin {
...
+                        // trailing BR is not needed.
+                        // This can be removed when the cursor is properly
+                        // handled around these elements.
                         const previousSibling = node.previousSibling;
                         return (
                             previousSibling &&
                             isBlock(previousSibling) &&
-                            this.dependencies.split.isUnsplittable(previousSibling)
+                            this.dependencies.split.isUnsplittable(previousSibling) &&
+                            previousSibling.nodeName !== "TABLE"
```
dmo-odoo, line 421: Why is the problem specific to tables? How comes we don't have the problem with other things?

sbbh-odoo: Because for unsplittable blocks we keep the trailing `<br>` to allow placing the cursor after them.

Tables are an exception: the cursor can be placed directly at the table edge, so the trailing `<br>` is unnecessary. That’s why the behavior differs only for tables.

EDIT: I think you’re asking why deleteBackward doesn’t cause issues for other unsplittable elements but does for tables.

deleteBackward should move the cursor into the last td’s paragraph. However, when the `<br>` is a direct sibling of the table, the editor tries to merge the `<br>` with that paragraph. Since unsplittable elements are also unmergeable, the merge fails. As a result, nothing effectively changes in the DOM and only a history step is created due to the internal `<br>` remove/add cycle.

dmo-odoo: Thanks for the details. Does the problem still exist in more recent versions that have the selection placeholders mechanism?

deso-odoo: No the problem does not exist in the recent version with selection placeholders mechanism following this commit: https://github.com/odoo/odoo/commit/0778d56957b1e

We will drop commit 2 from saas-19.1

dmo-odoo: Alright thanks, let's go then! 👍 
Edit: or not, that's a PR that I pre-reviewed, so it hasn't been greened by Lionel yet.

PR #245012, opened 2026-01-21
[FIX] iot_drivers: slow printing while waiting for cups_lock

MaxTheWhale:
Before this commit, the `printer_interface_L` and `printer_driver_L` shared a single `cups.Connection` instance guarded with a `Lock`. This meant that while the interface for checking for new printers (which can take 10-15 seconds), all printers were being blocked from printing until it was finished.

After this commit, each driver creates its own `cups.Connection` and `Lock`. This means they should never block each other, and prevents long pauses when trying to print.

---

### DIFF
--- a/addons/iot_drivers/iot_handlers/drivers/printer_driver_L.py
+++ b/addons/iot_drivers/iot_handlers/drivers/printer_driver_L.py
@@ -1,7 +1,8 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from base64 import b64decode
-from cups import IPPError, IPP_JOB_COMPLETED, IPP_JOB_PROCESSING, IPP_JOB_PENDING, CUPS_FORMAT_AUTO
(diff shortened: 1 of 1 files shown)


addons/iot_drivers/iot_handlers/drivers/printer_driver_L.py
loouis-t, line 17: I think you can remove `conn` here

MaxTheWhale: I was planning to do this in the 19.0 FW, the reason it is still here is because of the `supported` method (which is called without the class being instantiated so we can't use our own connection yet)

loouis-t: ah yes sure, makes sense

loouis-t (approved):
tested and still works as expected.

robodoo r+

PR #240945, opened 2025-12-22
[FIX] pos_sale: fix qty_delivered with refund

robinengels:
The qty_delivered on sale.order.line was not correctly computed when the original order was refunded with a ship later.

Steps to reproduce:
-------------------
* Create a sale order for 5 quantities of any product
* Confirm the sale order
* Settle the order in the PoS
* At this point the qty_delivered on the sale order line is 5
* Now go back to the PoS and refund partially the order for 3 quantities and use the "Ship Later" option
> Observation: The qty_delivered is 0 instead of 2

Why the fix:
------------
We group the pos.order.line by procurement group and then check if all pickings related to these lines are done before adding the qty to the qty_delivered. We also make sure to include the refund lines in the computation

opw-5059560

PR #245013, opened 2026-01-21
[FIX] pos_sale: fix qty_delivered with refund

robinengels:
This commit fixes an issue introduced here : https://github.com/odoo/odoo/pull/240945 If the multi steps delivery was enabled, the flow would give a traceback

The qty_delivered on sale.order.line was not correctly computed when the original order was refunded with a ship later.

Steps to reproduce:
-------------------
* Create a sale order for 5 quantities of any product
* Confirm the sale order
* Settle the order in the PoS
* At this point the qty_delivered on the sale order line is 5
* Now go back to the PoS and refund partially the order for 3 quantities and use the "Ship Later" option
> Observation: The qty_delivered is 0 instead of 2

Why the fix:
------------
This line would crash because picking would contain more than one record
https://github.com/odoo/odoo/blob/2881f049b892b3380bcab62ad5b2b0b71a9ef261/addons/pos_sale/models/sale_order.py#L95

Instead of doing a groupby we just apply the logic separately for the
normal lines and the refund lines.

opw-5059560


adgu-odoo (commented):
robodoo delegate+

PR #245016, opened 2026-01-21
[FIX] pos_online_payment: ensure a customer on order for certain providers

sbel-odoo:
Currently some providers require a customer to be registered on the order because the email address needs to be sent with the request. We the order does not have a customer the payment cannot be made using that provider.

Step to reproduce:
------------------
- Set up Amazon payment services on a online pos payment method
- Set this method as the online payment method for a self and also set is as payment method of the pos
- Place an order in the self or without a customer on the normal pos
- Try to pay it
> Observation: When the page is redirected to provider's checkout page,
an error occurs: Signature mismatch

Why the fix:
------------
On a normal pos shop we are simply ensuring that a customer with an address mail is registered on the order if the provider of an online payment method requires customer identification.
- When selecting the payment method on the payment screen if the payment method is online, requires customer identification and there's no customer on the order, the validate order button will be unavailable and the customer button is highlighted.
- When there is a customer the validate button is highlighted
- When validating the order if the customer does not have an email it will not go through and warn the cashier that the customer needs an email. This behavior is similar to the present "Delivery" which, if the cashier disregarded all popups, will not have the validate button available until a cashier is registered. Also if the selected customer doesn't have an address it will show a popup upon validation.

We are not doing anything regarding the self order as using presets like "Delivery" is compatible since the information required filled by the customer creates a partner in the db and sets it on the order. This preset ensures we always have a partner.

Other preset don't but we don't want to block flows that are currently working.

The list of providers requiring customer identification can be extended.

opw-5406501

### DIFF
--- a/addons/pos_online_payment/static/src/app/utils/order_payment_validation.js
+++ b/addons/pos_online_payment/static/src/app/utils/order_payment_validation.js
@@ -69,6 +69,18 @@ patch(OrderPaymentValidation.prototype, {
                 });
                 return false;
             }
+            if (
+                this.order.paymentsRequireCustomer(onlinePaymentLines) &&
+                !this.order.partner_id?.email
+            ) {
+                this.pos.dialog.add(AlertDialog, {
+                    title: _t("Payment provider requirement"),
+                    body: _t(
+                        "Please ensure you have a customer with an email address on this order."
+                    ),
+                });
+                return false;
+            }
             let prevOnlinePaymentLine = null;
             let lastOrderServerOPData = null;
             for (const onlinePaymentLine of onlinePaymentLines) {
--- a/addons/pos_online_payment/static/src/overrides/pos_overrides/models/pos_order.js
+++ b/addons/pos_online_payment/static/src/overrides/pos_overrides/models/pos_order.js
@@ -11,4 +11,15 @@ patch(PosOrder.prototype, {
         }
         return super.canBeValidated();
     },
+    get isCustomerRequired() {
+        const online_payments_customer_required = this.paymentsRequireCustomer(this.payment_ids);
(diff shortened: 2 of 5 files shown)


addons/pos_online_payment/static/src/overrides/pos_overrides/models/pos_order.js
```diff
@@ -11,4 +11,20 @@ patch(PosOrder.prototype, {
...
+        const res = super.isCustomerRequired;
+        return res || online_payments_customer_required;
+    },
+    payment_requires_customer(payments) {
+        return payments?.some(
+            (payment) =>
+                payment.payment_method_id.is_online_payment &&
+                payment.payment_method_id._customer_required
+        );
+    },
```
manv-afk, line 24:
```suggestion
    get isCustomerRequired() {
        const online_payments_customer_required = this.paymentsRequireCustomer(this.payment_ids);
        return super.isCustomerRequired || (!this.partner_id && online_payments_customer_required);
    },
    paymentsRequireCustomer(payments) {
        return payments?.some(
            (payment) =>
                payment.payment_method_id.is_online_payment &&
                payment.payment_method_id._customer_required
        );
    },
```

addons/pos_online_payment/static/src/app/utils/order_payment_validation.js
```diff
@@ -69,6 +69,18 @@ patch(OrderPaymentValidation.prototype, {
                 });
                 return false;
             }
+            if (
+                this.order.payment_requires_customer(onlinePaymentLines) &&
```
manv-afk:
```suggestion
                this.order.paymentsRequireCustomer(onlinePaymentLines) &&
```

manv-afk (changes requested):

manv-afk (approved):

PR #245021, opened 2026-01-21
[FIX] website: allow background color to be edited within s_tabs

romo-odoo:
Before this commit, it was not possible to edit the background color of a snippet dropped within the s_tabs snippet. The user could set the background image / video, but not the color. The user could work around the limitation by setting the background color of a snippet, saving it, then dropping it within the s_tabs snippet.

Therefore, it made no sense to disable the background color option.

Moreover, adding the background color option fixes the reverse issue: if the user dropped a snippet with a background color set by default, they cannot remove it since the colorpicker was not available.


FrancoisGe:
@robodoo r+

PR #245024, opened 2026-01-21
[FIX] account: different batches should not be grouped

gawa-odoo:
- Create a misc with 2 receivable lines with different partners
- Pay it
- You only get a single payment, for the first partner only

=> We grouped the lines per move, but we should only do that inside a same batch

Also fixes some caba tests that were incomplete

---

### DIFF
--- a/addons/account/tests/test_account_payment_register.py
+++ b/addons/account/tests/test_account_payment_register.py
@@ -2057,3 +2057,46 @@ class TestAccountPaymentRegister(AccountTestInvoicingCommon, PaymentCommon):
             {'amount_currency': 300, 'date_maturity': fields.Date.from_string('2017-01-06')},
             {'amount_currency': 600, 'date_maturity': fields.Date.from_string('2017-01-11')},
         ])
+
+    def test_payment_register_misc_different_batches(self):
+        """ Tests that payments that should be in different batches stay in different ones """
+        move1 = self.env['account.move'].create({
+            'move_type': 'entry',
+            'journal_id': self.company_data['default_journal_misc'].id,
+            'date': '2025-01-01',
+            'line_ids': [
+                Command.create({
+                    'account_id': self.company_data['default_account_receivable'].id,
+                    'partner_id': self.partner_a.id,
+                    'balance': 100.0,
+                }),
+                Command.create({
+                    'account_id': self.company_data['default_account_receivable'].id,
+                    'partner_id': self.partner_b.id,
+                    'balance': 100.0,
+                }),
+                Command.create({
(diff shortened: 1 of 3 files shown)


addons/account/tests/test_account_payment_register.py
```diff
@@ -2057,3 +2057,55 @@ def test_payment_register_wizard_without_receivable_line_due_date(self):
...
+                Command.create({
+                    'account_id': self.company_data['default_account_receivable'].id,
+                    'partner_id': self.partner_b.id,
+                    'debit': 100.0,
+                    'credit': 0.0,
+                }),
+                Command.create({
+                    'account_id': self.company_data['default_account_revenue'].id,
+                    'debit': 0.0,
+                    'credit': 200.0,
```
smetl:
```suggestion
                    'balance': 100.0,
                }),
                Command.create({
                    'account_id': self.company_data['default_account_receivable'].id,
                    'partner_id': self.partner_b.id,
                    'balance': 100.0,
                }),
                Command.create({
                    'account_id': self.company_data['default_account_revenue'].id,
                    'balance': -200.0,
```

addons/account/tests/test_account_payment_register.py
```diff
@@ -2057,3 +2057,55 @@ def test_payment_register_wizard_without_receivable_line_due_date(self):
...
+                Command.create({
+                    'account_id': self.company_data['default_account_receivable'].id,
+                    'partner_id': self.partner_a.id,
+                    'debit': 100.0,
+                    'credit': 0.0,
+                }),
+                Command.create({
+                    'account_id': self.company_data['default_account_revenue'].id,
+                    'debit': 0.0,
+                    'credit': 100.0,
```
smetl:
```suggestion
                    'balance': 100.0,
                }),
                Command.create({
                    'account_id': self.company_data['default_account_revenue'].id,
                    'balance': -100.0,
```

addons/account/tests/test_account_payment_register.py
```diff
@@ -2057,3 +2057,55 @@ def test_payment_register_wizard_without_receivable_line_due_date(self):
...
+                }),
+            ],
+        })
+
+        wizard = self.env['account.payment.register'].with_context(
+            active_model='account.move',
+            active_ids=(move1 + move2).ids
+        ).create({})
```
smetl:
```suggestion
        wizard = self._register_payment(move1 + move2)
```

gawa-odoo: put the group_payment to False but yeah, didn't know this one existed

smetl (approved):
@robodoo r+

PR #245027, opened 2026-01-21
[IMP] web, *: replace legacy grab cursor mixin with standard CSS

juliusc2066:
*: account, html_builder, html_editor, point_of_sale

The `o-grab-cursor` mixin was originally introduced to handle cross-browser compatibility for the `grab` cursor style.

Since `cursor: grab` is now widely supported by all modern browsers, the mixin is no longer necessary. This commit removes the mixin definition and replaces all its usages with the standard CSS property.


adr-odoo (approved):
@robodoo r+

PR #245028, opened 2026-01-21
[FIX] l10n_hr_edi: Error handling and UBL tax extension compatibility

nape-odoo:
- Improving error handling for various requests
- Adjusting XML generation to not conflict with `account_edi_ubl_cii_tax_extension` if it is installed
- Adding a separate test for HR:E category taxes
- Replacing skipping import of not successfully fiscalized document with warnings displayed on the moves after import
task-none

---

### DIFF
--- a/addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
+++ b/addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
@@ -34,6 +34,7 @@ class AccountEdiXmlUBLHR(models.AbstractModel):
                                     'cbc:ID': {},
                                     'cbc:Name': {},
                                     'cbc:Percent': {},
+                                    'cbc:TaxExemptionReasonCode': {},
                                     'cbc:TaxExemptionReason': {},
                                     'hrextac:HRTaxScheme': {
                                         'cbc:ID': {},
@@ -42,7 +43,7 @@ class AccountEdiXmlUBLHR(models.AbstractModel):
                             }
                         },
                         'hrextac:HRLegalMonetaryTotal': {
-                            'hrextac:HRTaxExclusiveAmount': {},
+                            'cbc:TaxExclusiveAmount': {},
                             'hrextac:OutOfScopeOfVATAmount': {},
                         }
                     }
@@ -140,14 +141,17 @@ class AccountEdiXmlUBLHR(models.AbstractModel):
                         '_text': invoice.invoice_date_due
                     }
                 })
-        # HR-BR-6: Each previous invoice reference (BG-3) must have the date of issue of the previous invoice (BT-26).
         # HR-BT-3: Note on previous invoice
-        if invoice.reversed_entry_id:
-            document_node['cac:BillingReference'].update({
+        # HR-BR-6: Each previous invoice reference (BG-3) must have the date of issue of the previous invoice (BT-26).
+        if 'refund' in invoice.move_type and invoice.reversed_entry_id:
+            document_node['cac:BillingReference'] = {
+                'cac:InvoiceDocumentReference': {
+                    'cbc:ID': {'_text': invoice.ref},
+                },
                 'cbc:IssueDate': {
                     '_text': invoice.reversed_entry_id.invoice_date
-                    }
-                })
+                }
+            }
         # Document Type Codes and Process Type Logic
         if invoice.l10n_hr_process_type in ('P4', 'P6'):
             if invoice.move_type == 'out_invoice':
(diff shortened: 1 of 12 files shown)


addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -195,8 +202,8 @@ def _add_hr_extension_node(self, document_node):
                                 'hrextac:HRTaxSubtotal': hr_tax_subtotals,
                             },
                             'hrextac:HRLegalMonetaryTotal': {
-                                'hrextac:HRTaxExclusiveAmount': document_node['cac:LegalMonetaryTotal']['cbc:TaxExclusiveAmount'],
-                                'hrextac:OutOfScopeOfVATAmount': False,     # Currently unsupported, a HR-specific workaround can potentially be made
+                                'cbc:TaxExclusiveAmount': document_node['cac:LegalMonetaryTotal']['cbc:TaxExclusiveAmount'],
+                                'hrextac:OutOfScopeOfVATAmount': out_of_scope_node,
```
antoine162, line 208:
```suggestion
                                'hrextac:OutOfScopeOfVATAmount': {'_text': '0.00', 'currencyID': vals['currency_name']},
```
Wouldn't that do the same thing?

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -179,11 +183,14 @@ def _add_hr_extension_node(self, document_node):
...
                     'cbc:Name': hr_tax_name,
                     'cbc:Percent': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:Percent'],
+                    'cbc:TaxExemptionReasonCode': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReasonCode'],
                     'cbc:TaxExemptionReason': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReason'],
                     'hrextac:HRTaxScheme': tax_subtotals[i]['cac:TaxCategory'][0]['cac:TaxScheme'] if hr_tax_name['_text'] != "HR:POVNAK" else {'_text': "OTH"},
                 }
             }
             hr_tax_subtotals.append(new_item)
+        out_of_scope_node = document_node['cac:LegalMonetaryTotal']['cbc:TaxExclusiveAmount'].copy()
+        out_of_scope_node.update({'_text': '0.00'})     # Currently unsupported, a HR-specific workaround can potentially be made
```
antoine162: What are we copying from the `cbc:TaxExclusiveAmount` node that we're not overwriting in the next line? Better be explicit IMO and just add those keys directly in the `hrextac:OutOfScopeOfVATAmount` dict.

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -273,8 +280,10 @@ def _get_tax_exemption_reason(self, customer, supplier, tax):
                 'name': hr_category.name,
             })
             if tax.amount == 0:
+                tax_extension = 'ubl_cii_tax_exemption_reason_code' in tax._fields and tax.ubl_cii_tax_exemption_reason_code
+                # If account_edi_ubl_cii_tax_extension is installed, use data from it, if not, override with HR data
                 res.update({
-                    'tax_exemption_reason': hr_category.description,
+                    'tax_exemption_reason': res['tax_exemption_reason'] if tax_extension else hr_category.description,
```
antoine162: nitpick, but personally I think it's neater to do
```py
if not tax_extension:
    res['tax_exemption_reason'] = hr_category.description
```

antoine162 (commented):
Hiiii thanks so much, just a few very small comments below 😄 
Besides that there seems to be a merge conflict in `tools.py`. 😄

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
antoine162: Oh okay, the `vals` isn't a param of the `_add_hr_extension_node` function. In that case you can indeed still get the currency name from the `cbc:TaxExclusiveAmount` node.

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
nape-odoo: The currency, while accounting for a case where a currency isn't specified. But it can be taken from the other node with a `get()` with the same effect.

antoine162: Okay, makes sense! I'm pretty sure in this case the UBL standard requires the currency to be there.

addons/l10n_hr_edi/tools.py
```diff
@@ -80,7 +80,7 @@ def _make_request(company, endpoint_type, params=False):
     # Structure-specific error handling
     if response.status_code != 200:
         try:
-            error_message = response.json().get('errors')
+            error_message = response.json().get('errors', response.json().get('message'))
```
antoine162:
```suggestion
            response_json = response.json()
            error_message = response_json.get('errors', response_json.get('message'))
```

antoine162 (approved):
@robodoo r+

PR #245034, opened 2026-01-21
[IMP] mail: unify mailboxes and add new ones

Atovange:
This commit changes the way mailboxes are shown in the Discuss' sidebar and adds new mailboxes.
This effectively changes the way the inbox is viewed, becoming the one single place for all notifications, both read and unread, with the option to filter them.
This improves the navigation and gives the user the possibility to easily check the messages in priority (mentioning them), follow conversations (notes/discussions) or follow the state of their records (tracked changes).

task-5470655


Atovange:
Task canceled

PR #245039, opened 2026-01-21
[18.0][FIX] sale: Avoid error when internal user (no sale permissions) see Orders at /my

victoralmau:
Avoid error when internal user (no sale permissions) see Orders at /my

Similar to https://github.com/odoo/odoo/commit/5ebab949a06ec338cc28e912317a74bbfb3fe6ac

@Tecnativa TT60025

---


Feyensv (approved):
LGTM, thanks for the fix 👌 

@robodoo r+

PR #245048, opened 2026-01-21
[FIX] iot_drivers: propagate session_id in scale and generic drivers

orma-odoo:
[FIX] iot_drivers: propagate session_id in scale and generic drivers

Before this commit, `session_id` was lost during scale operations because `ScaleDriver` overwrote `self.data` instead of updating it, erasing the context. Additionally, the generic Driver class stored the ID as "owner", causing a TypeError on the server which expects "session_id".

After this commit, `ScaleDriver` uses `.update()` to preserve the session context, and `Driver.action` correctly uses the key `session_id` in the response payload. This ensures requests are properly acknowledged by the Odoo server.

opw-5485278

---

### DIFF
--- a/addons/iot_drivers/driver.py
+++ b/addons/iot_drivers/driver.py
@@ -60,6 +60,7 @@ class Driver(Thread):
         session_id = data.get('session_id')
         if session_id:
             self.data["owner"] = session_id
+            self.data["session_id"] = session_id
 
         try:
             response = {'status': 'success', 'result': self._actions[action](data), 'session_id': session_id}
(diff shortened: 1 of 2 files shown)


addons/iot_drivers/driver.py
```diff
@@ -57,7 +57,7 @@ def action(self, data):
 
         session_id = data.get('session_id')
         if session_id:
-            self.data["owner"] = session_id
+            self.data["session_id"] = session_id
```
loouis-t, line 63:
```suggestion
            self.data["owner"] = session_id
            self.data["session_id"] = session_id
```
we need to keep it as longpolling still uses `owner` as key

loouis-t (approved):
tested simulating a scale, both longpolling and websocket can read the response.

robodoo r+

PR #245058, opened 2026-01-21
[FIX] l10n_it_edi: on import fill payment_reference field correctly

blackne0n:
The payment_reference field in invoices was being filled with a wrong field from the imported XML, progressivoinvio is the progressive number of invoices sent by the partner's system, not a partner's requested payment reference.

Current behavior before PR:
On import, payment_reference was being filled with ProgressivoInvio, making automated payments out to partners harder.

payment_reference is only being filled if partner specifies a payment reference in the EDI, avoiding confusion.

---

### DIFF
--- a/addons/l10n_it_edi/models/account_move.py
+++ b/addons/l10n_it_edi/models/account_move.py
@@ -1101,9 +1101,9 @@ class AccountMove(models.Model):
                 ))
                 message_to_log.append(message)
 
-            # Numbering attributed by the transmitter
-            if progressive_id := get_text(tree, '//ProgressivoInvio'):
-                self.payment_reference = progressive_id
+            # Payment code
+            if payment_code := get_text(tree, './/DettaglioPagamento[1]/CodicePagamento'):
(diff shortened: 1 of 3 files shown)


Megaaaaaa (changes requested):
Hello, :wave:

Thank you for your contribution!

First small thing, the title of your CLA signature commit is wrong and therefore none of our tests run since the minimal check doesn't pass. Can I ask you to use the expected format for that title commit which would be:
`[CLA] Signature for blackne0n`

Apart from that, the diff looks good to me and makes sense. But could I ask you to add a link to the official documentation you used to justify the change in the appropriate commit ? So i guess that would be [this](https://www.fatturapa.gov.it/export/documenti/fatturapa/v1.4/Specifiche_tecniche_del_formato_FatturaPA_V1.4.pdf) except if you have a more recent one ?

Thank you in advance!

Megaaaaaa (approved):
Looks good now, thank you!

robodoo r+

PR #245059, opened 2026-01-21
[FIX] payment_stripe: fix stripe payment failing due to currency decimal

mma1377:
### Issue:
The stripe payments are failing to due setting currency decimals.

#### Steps to reproduce:
1- In USD currency form in developer mode, set the rounding factor
   to `0.001000`, so we have 3 decimal places.
2- From `Decimal Accuracy` set Price Unit to 4 digits.
3- Refresh the page.
4- Create a SO and add a line. Set the price to `200.7647`.
5- Refresh the page again and generate a payment link.
6- Pay through stripe. As you see payments fail.

### Cause:
This is due to rounding the amount before sending to stripe:
https://github.com/odoo/odoo/blob/824e0c42c62de7f570f106860e7dcbf1f7344c14/addons/payment_stripe/models/payment_transaction.py#L160-L165

We are rounding the amount to the minor currency unit.

However, we make a compare between `tx_amount` and the amount paid through stripe:
https://github.com/odoo/odoo/blob/824e0c42c62de7f570f106860e7dcbf1f7344c14/addons/payment/models/payment_transaction.py#L726-L732

Which are not equal. We should use minor current unit as `precision_digits` to fix this issue.

opw-5496825

### DIFF
--- a/addons/payment_stripe/models/payment_transaction.py
+++ b/addons/payment_stripe/models/payment_transaction.py
@@ -9,6 +9,7 @@ from odoo import _, fields, models
 from odoo.exceptions import UserError, ValidationError
 
 from odoo.addons.payment import utils as payment_utils
+from odoo.addons.payment.const import CURRENCY_MINOR_UNITS
 from odoo.addons.payment_stripe import const
 from odoo.addons.payment_stripe import utils as stripe_utils
 from odoo.addons.payment_stripe.controllers.main import StripeController
@@ -393,7 +394,10 @@ class PaymentTransaction(models.Model):
             payment_data.get('amount', 0), self.currency_id
         )
         currency_code = payment_data.get('currency', '').upper()
-        self._validate_amount_and_currency(amount, currency_code)
+        precision_digits = CURRENCY_MINOR_UNITS.get(
+            self.currency_id.name, self.currency_id.decimal_places
+        )
+        self._validate_amount_and_currency(amount, currency_code, precision_digits=precision_digits)
 
     def _process_notification_data(self, notification_data):
         """ Override of `payment` to process the transaction based on Stripe data.
--- a/addons/payment_stripe/tests/test_stripe.py
+++ b/addons/payment_stripe/tests/test_stripe.py
@@ -5,6 +5,7 @@ from unittest.mock import patch
 
 from werkzeug.urls import url_encode, url_join
 
+from odoo.exceptions import ValidationError
 from odoo.tests import tagged
 from odoo.tools import mute_logger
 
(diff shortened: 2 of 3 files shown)


vchu-odoo (commented):
Thank you for the fix!
Could you please add a test?

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -200,3 +200,16 @@ def test_create_account_link_pass_required_parameters(self):
...
+        Test that compare_notification_data accepts amount rounded down to the currency minor unit.
+        """
+        self.amount = 20076
+        self.currency.rounding = 0.001
+        tx = self._create_transaction(
+            'dummy', operation='online_direct', tokenize=True, amount=200.769
+        )
+        data = self.notification_data['data']
+        data['payment_intent'] = self._mock_payment_intent_request()
+        tx._compare_notification_data(data)
```
vchu-odoo, line 217: self._assert_does_not_raise(ValidationError, tx._compare_notification_data, data)

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -200,3 +200,16 @@ def test_create_account_link_pass_required_parameters(self):
             call_args = mock.call_args.kwargs['payload'].keys()
             for payload_param in ('account', 'return_url', 'refresh_url', 'type'):
                 self.assertIn(payload_param, call_args)
+
+    def test_notification_data_matches_rounded_transaction_amount(self):
```
vchu-odoo:
```suggestion
    def test_compare_notification_data_uses_payment_minor_unit(self):
```

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -200,3 +200,16 @@ def test_create_account_link_pass_required_parameters(self):
...
+        """
+        Test that compare_notification_data accepts amount rounded down to the currency minor unit.
```
vchu-odoo:
```suggestion
        Test that the payment's minor unit precision is used to validate the amount, not the custom
        currency rounding that customers may have configured.
```

addons/payment_stripe/models/payment_transaction.py
```diff
@@ -393,7 +394,10 @@ def _compare_notification_data(self, notification_data):
             payment_data.get('amount', 0), self.currency_id
         )
         currency_code = payment_data.get('currency', '').upper()
-        self._validate_amount_and_currency(amount, currency_code)
+        precision_digits = CURRENCY_MINOR_UNITS.get(
+            self.currency_id.name, self.currency_id.decimal_places
+        )
```
vchu-odoo, line 399: i checked for other payment providers and there is the same issue, I think we need to do it in `_validate_amount_and_currency` itself.

vchu-odoo (commented):
Hello,
thank you for the changes, just suggested how to test not raising and noticed that for other providers, the same issue can happen.

addons/payment_stripe/models/payment_transaction.py
mma1377: @vchu-odoo 
In that case, there might be another issue about rounding. Are we sure all payment providers round down like Stripe and not round to nearest?

vchu-odoo: Normally, it should, as we send amounts rounded to vs to_minor_currency_units.
But indeed, we need to check the others to be sure. that adyen also has the same issue but not xendit.
Let us indeed fix it for each provider where there is an issue.
Thank you!

vchu-odoo (approved):
Thank you for the fix!

@robodoo r+

PR #245065, opened 2026-01-21
[FIX] mrp_subcontracting{,_purchase}: fix inconsistency between stock move and move line quantities

MaximeNoirhomme229:
**Issue**
In subcontracting, if a BOM is modified after the creation of a PO, then modifying the move quantity of the associated receipt can lead to inconsistency between move and move line quantities.

**Steps to reproduce**
- Create a subcontracting BOM of a final product using 1 component product
- Create a PO of the final product for a quantity of 10 and confirm it
- Modify the BOM to use 2 component products instead
- Go to the receipt of the PO and modify the quantity to 2 and validate it
- Click on the move line of the receipt -> The displayed quantity is 10 instead of 2

**Cause**
Setting the quantity triggers this line:
https://github.com/odoo/odoo/blob/c496235b9520b2a33040174974de7d37e74b0580/addons/mrp_subcontracting/models/stock_move.py#L78-L78 
which calls:
https://github.com/odoo/odoo/blob/c496235b9520b2a33040174974de7d37e74b0580/addons/mrp_subcontracting/models/stock_move.py#L107

Since the BOM has been modified, a `consumption_issues` is detected and `_update_finished_move()` won't be called:
https://github.com/odoo/odoo/blob/c496235b9520b2a33040174974de7d37e74b0580/addons/mrp_subcontracting/models/mrp_production.py#L86-L89

And since the returned action is not used when calling `subcontracting_record_component`, `_update_finished_move()` won't be called later neither, which is the method responsible for updating the move line quantities:
https://github.com/odoo/odoo/blob/c496235b9520b2a33040174974de7d37e74b0580/addons/mrp_subcontracting/models/mrp_production.py#L142-L146

**Solution**
Since the consumption issue actions are ignored in this case, just skipped it and avoid inconsistencies.

opw-5493577

### DIFF
--- a/addons/mrp_subcontracting/models/stock_move.py
+++ b/addons/mrp_subcontracting/models/stock_move.py
@@ -104,7 +104,7 @@ class StockMove(models.Model):
                 production.qty_producing = 1
                 if not production.lot_producing_id:
                     production.action_generate_serial()
-                production.with_context(cancel_backorder=False).subcontracting_record_component()
+                production.with_context(cancel_backorder=False, skip_consumption=True).subcontracting_record_component()
         else:
             production.qty_producing = qty
             if float_compare(production.qty_producing, production.product_qty, precision_rounding=production.product_uom_id.rounding) > 0:
@@ -115,7 +115,7 @@ class StockMove(models.Model):
             if production.product_tracking == 'lot' and not production.lot_producing_id:
                 production.action_generate_serial()
             production._set_qty_producing()
-            production.with_context(cancel_backorder=False).subcontracting_record_component()
+            production.with_context(cancel_backorder=False, skip_consumption=True).subcontracting_record_component()
 
     def copy(self, default=None):
         self.ensure_one()
--- a/addons/mrp_subcontracting_purchase/tests/test_mrp_subcontracting_purchase.py
+++ b/addons/mrp_subcontracting_purchase/tests/test_mrp_subcontracting_purchase.py
@@ -485,6 +485,58 @@ class MrpSubcontractingPurchaseTest(TestMrpSubcontractingCommon):
         self.assertEqual(account_move_credit_line.account_id.id, stock_in_acc_id)
         self.assertEqual(account_move_credit_line.credit, 5)
 
+    def test_receipt_consumption_issues_due_to_subcontract_bom_modifications(self):
+        """
+        Buy 10 subcontracted serial and non-serial products. Modify the BOM quantities.
+        Then set move quantities to 2 for both product type and validate the receipt.
+        """
+        self.finished3 = self.env['product.product'].create({
+            'name': 'SuperProduct',
+            'type': 'product',
+            'tracking': 'serial',
+        })
+        self.bom_finished3 = self.env['mrp.bom'].create({
+            'product_tmpl_id': self.finished3.product_tmpl_id.id,
+            'type': 'subcontract',
+            'subcontractor_ids': [Command.set(self.subcontractor_partner1.ids)],
+            'bom_line_ids': [Command.create({
+                'product_id': self.comp3.id,
+                'product_qty': 1,
+            })],
+        })
+        po = self.env['purchase.order'].create({
+            'partner_id': self.subcontractor_partner1.id,
+            'order_line': [
+                Command.create({
+                    'name': self.finished2.name,
+                    'product_id': self.finished2.id,
+                    'product_qty': 10,
+                    'product_uom': self.finished2.uom_id.id,
+                    'price_unit': 1,
+                }),
+                Command.create({
+                    'name': self.finished3.name,
+                    'product_id': self.finished3.id,
+                    'product_qty': 10,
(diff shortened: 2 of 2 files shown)


addons/mrp_subcontracting/models/stock_move.py
```diff
@@ -115,7 +115,7 @@ def _auto_record_components(self, qty):
             if production.product_tracking == 'lot' and not production.lot_producing_id:
                 production.action_generate_serial()
             production._set_qty_producing()
-            production.with_context(cancel_backorder=False).subcontracting_record_component()
+            production.with_context(cancel_backorder=False, skip_consumption=True).subcontracting_record_component()
```
lase-odoo, line 118: The bug is even worse when the final product is tracked by `serial`. Currently the bug can still be reproduced in that use case because an other `subcontracting_record_component` call is performed in that case:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp_subcontracting/models/stock_move.py#L103-L118
I double checked and we should also add the `skip_consumption` to that case.

I would adapt the test to add a second pol for a subcontracted and tracked product to test that change aswell.

addons/mrp_subcontracting_purchase/tests/test_mrp_subcontracting_purchase.py
```diff
@@ -485,6 +485,32 @@ def test_subcontract_product_price_change(self):
...
+
+        line = self.bom_finished2.bom_line_ids[0]
+        line.write({'product_qty': line.product_qty + 1})
+
+        receipt = po.picking_ids
+        receipt.move_ids.quantity = 2
+        receipt.button_validate()
+
+        self.assertEqual(receipt.move_line_ids.quantity, 2)
+        self.assertEqual(receipt.move_ids.quantity, 2)
```
lase-odoo: Regarding the flow, I think that the assert should also be done prior to the `button_validate`.

Taking my other comment into account you will end-up with multiple moves and move lines but I would just make `self.assertRecordValues` in that case.

addons/mrp_subcontracting_purchase/tests/test_mrp_subcontracting_purchase.py
```diff
@@ -485,6 +485,32 @@ def test_subcontract_product_price_change(self):
...
+                'product_id': self.finished2.id,
+                'product_qty': 10,
+                'product_uom': self.finished2.uom_id.id,
+                'price_unit': 1,
+            })],
+        })
+        po.button_confirm()
+
+        line = self.bom_finished2.bom_line_ids[0]
+        line.write({'product_qty': line.product_qty + 1})
```
lase-odoo: As a general rule of thumb it is best to not call write directly if you do not have a reason to (like writing on multiple fields simultaneously).
```suggestion
        line.product_qty = line.product_qty + 1
```

addons/mrp_subcontracting_purchase/tests/test_mrp_subcontracting_purchase.py
```diff
@@ -485,6 +485,32 @@ def test_subcontract_product_price_change(self):
         self.assertEqual(account_move_credit_line.account_id.id, stock_in_acc_id)
         self.assertEqual(account_move_credit_line.credit, 5)
 
+    def test_modify_subcontract_bom_after_validate_po_modify_receipt(self):
```
lase-odoo: The title suggest that you modify the bom after you validate the po. What about:
```suggestion
    def test_receipt_consumption_issues_due_to_subcontract_bom_modifications(self):
```

lase-odoo (commented):

I like the fix axis since the `skip_consumption` guarranties that we will avoid checking for `consumption_issues`:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp/models/mrp_production.py#L1553-L1564
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp_subcontracting/models/mrp_production.py#L76-L88
Which indeed does not make sense in the present case since or call is an "automatic" action that will not return its value and hence will not allow to open the consumption wizard. Note that this was already the fix axis used by the `action_confirm` of the consumption warning:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp_subcontracting/wizard/mrp_consumption_warning.py#L10-L13
Note also that, it does not invalidate the purpose of these lines: 
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp_subcontracting/models/mrp_production.py#L86-L88
in the `subcontracting_record_component` since this action can still be called from the view and in that case allows one to return the consumption issue wizard:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/mrp_subcontracting/views/mrp_production_views.xml#L49-L55

I still left a review on the fix because it should be applied twice (currently the tracked related flows is still bugged).

PR #245068, opened 2026-01-21
[IMP] crm_iap_mine: remove redundant fields in mail template

meba-odoo:
- Remove newly added fields that can be replaced by existing keys present in `iap_mail.enrich_company` template to ensure we get more data in chatter without upgrade and avoid duplicate entries

IAP PR: https://github.com/odoo/iap-apps/pull/1390


RicGR98 (approved):
@robodoo up to 19.0

https://github.com/odoo/odoo/pull/244437

@robodoo r+

PR #245074, opened 2026-01-21
[FIX] *: Fix json decode error imports

Julien00859:
Installing `requests==2.25.1` and using the following line of code:

    from requests.exceptions import JSONDecodeError

It raises the following error:

    ImportError: cannot import name 'JSONDecodeError' from 'requests.exceptions' (python3.10/site-packages/requests/exceptions.py)

It was removed from the following commit in the `requests` package:

    https://github.com/psf/requests/commit/db575eeedcfdb03bf31285afd3033e301df8b685

This change fixes this error importing the original exception from `json` package


Julien00859:
@robodoo help

aj-fuentes (approved):
@robodoo r+

PR #245141, opened 2026-01-21
[FIX] stock_account: prevent inflated and negative values in the avco justification report

rosm-odoo:
Problem:
In the `stock.avco.report`, there is a chance for floating point inaccuracies to cause avco unit cost to inflate. This occurs specifically when the remaining quantity is calculated to be a positive number extremely close to 0. There are also issues with negative avco value that are tangential to this error.

Solution:
We will mirror the iterative logic from `_run_avco`, being sure to account for whether we have moved from negative quantity to positive or vice versa. Also, before calculating avco value as `total_value / total_quantity`, we will ensure `total_quantity` is not a floating point number very close to 0 to avoid unit cost inflation.

Steps to replicate (Runbot 19):
Have storage locations enabled to make this easier to test
1. Product with tracked inventory, avco perpetual valuation, non-zero cost
2. Set the on hand quantity to -0.3 Units
3. Set the on hand quantity to -0.2 Units
4. Set the on hand quantity to 0 Units (A well-known python quirk, that .1 + .2 = 0.30000000000000004)

Go to the stock report and inspect the unit cost history for the product, note
1. The nonsense Unit Cost on the top row
2. The negative Unit Cost on the 2nd row

Note to reviewer:
I discussed this issue briefly with dafr, who pointed out that there are further adjustments to come with the `_run_avco` method, and several other issues with this report that are not addressed by this PR. However, the problems with this report are affecting a customer so we'd like to push these fixes forward for now.

opw-5430300

### DIFF
--- a/addons/stock_account/report/stock_avco_audit_report.py
+++ b/addons/stock_account/report/stock_avco_audit_report.py
@@ -1,4 +1,5 @@
 from odoo import fields, models, tools
+from odoo.tools import float_is_zero
 
 
 class StockAverageCostReport(models.AbstractModel):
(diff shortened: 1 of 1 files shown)


ticodoo:
@rosm-odoo  LGTM, can you rebase and fix the conflict?

And also remove your reviewer's note from your commit message (it can stay in the PR description if you want). Some other tips for better commit messages (doesn't have to be done for this PR):
- commit titles are typically limited to around 74 chars for readability and the same applies to the body's line wrap limit.
- keep your wording simple and concise.
  - Instead of `[FIX] stock_account: prevent inflated and negative values in the avco justification report` => `[FIX] stock_account: avoid rounding errors in avco unit cost history ` is easier/faster to read
  - I'd also drop the use of the words "tangential", "mirror", "iterative" all uses of "inflation/inflate", and "quirk" either completely when they're not needed or use a simpler word. Remember non-native English speakers read these commits and it requires extra thinking to know what you mean when you use these words (even for native speakers).
- you should write the expected vs the actual results of the calculation. "nonsense Unit Cost" doesn't mean much to someone who doesn't want to reproduce the steps. In this case, with a Cost of $10, the Unit Cost ends up being more than $100,000,000,000 when following your steps instead of the correct $10.

rosm-odoo:
Hi @ticodoo 

I just rebased and updated the commit message to remove the note to reviewer.

I will definitely implement these other notes in my future contributions, thank you for the detailed feedback.

PR #238115, opened 2025-12-01
[FIX] account: discard data from non dependent modules

william-andre:
In order to avoid declaring an exponential number of bridge modules, we can define values for fields declared in modules that are not in the dependencies of the l10n.
In order to support that, `_pre_load_data` needs to be called before `_load_data` as it is the method discarding those fields.

Reproduce:
* Create a database with only `account` and the demo data
* Install a l10n with stock account defined, for instance `l10n_mx`

PR #178113, opened 2024-08-27
[FIX] account: When changing CoA, reset company.anglo_saxon_accounting

antoine162:
To reproduce:
- Install `account` in a fresh DB
- Change the CoA to one which doesn't use anglo-saxon accounting
- Notice how anglo-saxon accounting is still activated on the company

Analysis:
- Since `anglo_saxon_accounting` has a default value of False, we don't bother to specify its value in the templates which don't use it.
- So, when switching CoA from one which uses anglo-saxon accounting to one which doesn't, False doesn't get written.

Solution:
- We explicitly reset it when changing CoA.
- We could have put a default value of False in the template values, but that doesn't combine nicely with the existing code, so this solution is preferred.

This has been causing issues for new SaaS databases, since the `saas_worker` first generates a database template for the Accounting app, and only afterwards generates database templates for each localization by changing the CoA.

As far as I can see, no other fields need the same treatment.

taskid: none

PR #245152, opened 2026-01-22
[FIX] account: set anglo saxon accounting again

jorenvo:
In 19.1 a fix was developed to avoid bridge modules for template data [1].

A couple of `ChartTemplate._pre_load_data` calls without template data were added to discard non-existent fields in the database.

It inadvertently triggers an older fix [2] that was meant to unset `anglo_saxon_accounting` for existing companies if not explicitly set in the template, which meant that in practice `anglo_saxon_accounting` wasn't being set anymore.

[1] https://github.com/odoo/odoo/pull/238115
[2] https://github.com/odoo/odoo/pull/178113

task-5785189

### DIFF
--- a/addons/account/models/chart_template.py
+++ b/addons/account/models/chart_template.py
@@ -528,8 +528,9 @@ class AccountChartTemplate(models.AbstractModel):
         if not company.country_id:
             vals['country_id'] = fiscal_country.id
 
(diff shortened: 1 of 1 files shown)


william-andre (approved):
@robodoo r+

PR #245159, opened 2026-01-22
[FIX] point_of_sale: prevent multiple prints of general note message

jipr-odoo:
Before this commit:
===================
The general note message was printed multiple times for a single order, resulting in duplicate prints on the printer.

After this commit:
====================
Ensure the general note message is printed only once, preventing duplicate prints.

Task-5502902


davidmonnom (commented):
I don't want to change this method in 18 since we don't have enough test about it.

To me is ok if generalNote is on all receipts.

PR #245162, opened 2026-01-22
[FIX] pos_restaurant: tb when merging a table with an empty order

parp-odoo:
Steps to reproduce:
- On an empty table, change the guest count
- Create an order and send it to the kitchen
- Open another table without an order
- Merge the first table with the second one

Issue:
- A server error occurs during table merge

Fix:
- Wait for the merge order to sync before returning the result

Task-5502511

Related: https://github.com/odoo/enterprise/pull/104577

### DIFF
--- a/addons/pos_restaurant/static/src/app/services/pos_store.js
+++ b/addons/pos_restaurant/static/src/app/services/pos_store.js
@@ -214,8 +214,10 @@ patch(PosStore.prototype, {
             }
         }
 
-        await this.deleteOrders([sourceOrder], [], true);
-        this.syncAllOrders({ orders: [destOrder] });
+        if (typeof destOrder.id === "number") {
+            await this.syncAllOrders({ orders: [destOrder] });
+        }
+        await this.deleteOrders([sourceOrder], [], typeof sourceOrder.id === "number");
         return destOrder;
     },
     mergeCourses(sourceOrder, destOrder) {
--- a/addons/pos_restaurant/static/tests/unit/services/pos_service.test.js
+++ b/addons/pos_restaurant/static/tests/unit/services/pos_service.test.js
@@ -351,18 +351,29 @@ describe("restaurant pos_store.js", () => {
 
     test("transferOrder", async () => {
         const store = await setupPosEnv();
+        const date = DateTime.now();
         const tableSrc = store.models["restaurant.table"].get(1);
         const tableDst = store.models["restaurant.table"].get(2);
-        const sourceOrder = store.addNewOrder({ table_id: tableSrc });
+        const sourceOrder = store.addNewOrder({
+            table_id: tableSrc,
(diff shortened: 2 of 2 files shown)


addons/pos_restaurant/static/src/app/services/pos_store.js
```diff
@@ -214,8 +214,8 @@ patch(PosStore.prototype, {
             }
         }
 
+        await this.syncAllOrders({ orders: [destOrder] });
```
davidmonnom: `await this.deleteOrders([sourceOrder], [], <check if sourceOrder is synced>);`

davidmonnom (changes requested):

addons/pos_restaurant/static/src/app/services/pos_store.js
parp-odoo: Updated! :slightly_smiling_face:

addons/pos_restaurant/static/src/app/services/pos_store.js
```diff
@@ -214,8 +214,8 @@ patch(PosStore.prototype, {
             }
         }
 
-        await this.deleteOrders([sourceOrder], [], true);
-        this.syncAllOrders({ orders: [destOrder] });
+        await this.syncAllOrders({ orders: [destOrder] });
```
davidmonnom: So you don't need to sync anymore with the check :)

parp-odoo: Then need to add same condition (order is synced or not) before calling merge_orders here:
https://github.com/odoo-dev/enterprise/blob/saas-18.3/pos_restaurant_preparation_display/static/src/override/point_of_sale/services/pos_store.js#L27

else it leads to a singleton error.

davidmonnom: The call here is not necessary if there is only one of them synced ?

parp-odoo: Yes, I'll update it

parp-odoo: > So you don't need to sync anymore with the check :)

This results in some failing cases.

For example, if we merge two already-synced orders and then perform operations that do not trigger a re-sync, the data of one order gets lost.

Reason: final order state is not updated on the server.

https://github.com/user-attachments/assets/1ef9c0ce-ef03-47ed-b5b4-758f574ab573

parp-odoo: Updated the code to sync only if the source order is already synced:
```js
if (typeof destOrder.id === "number") {
    await this.syncAllOrders({ orders: [destOrder] });
}
```

davidmonnom (approved):
@robodoo r+

PR #245164, opened 2026-01-22
[FIX] hr: avoid traceback when searching employee in appraisal

tshe-odoo:
Steps to reproduce:
1. Log in with demo user
2. Create a new appraisal
3. Search for an employee

Issue:
A traceback occurs when searching for an employee.

Cause:
The `registration_number` field is not available on the `hr.employee.public` model for public users.

Fix:
Remove `registration_number` from the employee search domain and restrict the search to fields available on the public model.

task-5864546


amah-odoo (approved):
LGTM 💯 

@robodoo r+

PR #238635, opened 2025-12-04
[REF] point_of_sale: move scales logic to enterprise

loouis-t:
To maintain consistency, we moved IoT connected scales logic from `point_of_sale` to `pos_iot`.

see odoo/enterprise#101307

PR #245165, opened 2026-01-22
[FIX] point_of_sale: product with unique serial number leads traceback

djip-odoo:
Steps to reproduce:
=
- Configure a product with `Tracking`=`By Unique Serial Number`
- Open POS and click on the configured product.
- The “Lot/Serial number(s) required” dialog opens.
- Click `Discard`.

Issue:
=
- The following error occurs: `TypeError: Cannot read properties of undefined (reading 'modifiedPackLotLines')`

Fix:
=
- Stop further steps to add the product when the lot/serial number dialog is discarded.

Reference:
=
- https://github.com/odoo/odoo/pull/238635/files#diff-9e8905c1e88dc96f9145acda2c6a165fb14a58e0d2cfb5b3b6049411918ba27dL913-L914

task-5505855

related pr: https://github.com/odoo/enterprise/pull/108263

### DIFF
--- a/addons/point_of_sale/static/src/app/services/pos_store.js
+++ b/addons/point_of_sale/static/src/app/services/pos_store.js
@@ -929,16 +929,20 @@ export class PosStore extends WithLazyGetterTrap {
         if (keepGoing === false) {
             return;
         }
-
-        const pack_lot_ids = await this.configureNewOrderLine(
-            productTemplate,
-            vals,
-            values,
-            order,
-            options,
-            configure
-        );
-
+        let pack_lot_ids = {};
+        if (this.requiresOrderLineConfiguration(configure, opts.code, values.product_tmpl_id)) {
+            pack_lot_ids = await this.configureNewOrderLine(
+                productTemplate,
+                vals,
+                values,
+                order,
+                options,
+                configure
+            );
+            if (!pack_lot_ids) {
+                return;
+            }
+        }
         // Handle price unit
         this.handlePriceUnit(values, order, vals.price_unit);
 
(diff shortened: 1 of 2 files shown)


addons/point_of_sale/tests/test_frontend.py
```diff
@@ -2029,6 +2029,12 @@ def test_only_existing_lots(self):
             'tracking': 'lot',
             'available_in_pos': True,
         })
+        self.env['product.product'].create({
+            'name': 'Product with Unique Serial Number',
+            'is_storable': True,
+            'tracking': 'serial',
+            'available_in_pos': True,
+        })
```
paso-odoo: We have existing products with tracking as a lot or serial, so you can use them!

addons/point_of_sale/tests/test_frontend.py
```diff
@@ -2042,7 +2048,7 @@ def test_only_existing_lots(self):
         }]).sudo().action_apply_inventory()
 
         self.main_pos_config.picking_type_id.write({
-            "use_create_lots": False,
+            "use_create_lots": True,
```
paso-odoo: Is there any reason for set as True. 
Tour name does not match with the boolean you have set as True.

paso-odoo (changes requested):

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -920,6 +920,9 @@ export class PosStore extends WithLazyGetterTrap {
             options,
             configure
         );
+        if (!pack_lot_ids) {
+            return;
+        }
```
davidmonnom: I understand why this works but isn't really clear, when orderline doesn't need any lot it return an empty object. So it should be cool to use

`if (values.product_tmpl_id.to_weight && this.config.iot_scale_id && configure) {`
and
`if (values.product_tmpl_id.isTracked() && (configure || code)) {`

To be sure to not return wrongly in the future.

davidmonnom (changes requested):

davidmonnom (approved):
@robodoo r+

PR #245168, opened 2026-01-22
[FIX] mass_mailing: fix non-deterministic errors

abd-msyukyu-odoo:
After the previous [fix], other non-deterministic errors were revealed. It seems
`200` ms is sometimes too little for operations involving multiple async calls
like `save` on the mass_mailing form view, and display the style option in the
`html_builder` when clicking on a block.

The `body_html` field is displayed in readonly as we never need to use the
editor for this field. This prevents selection/event conflicts between 2
editors at the same time which is something that we don't want to test in
`mass_mailing`.

[fix]: https://github.com/odoo/odoo/commit/a180d09aaf0661483670a79deedc3223a0b6e621

runbot-error-238002
runbot-error-238045

task-5866753

### DIFF
--- a/addons/mass_mailing/static/tests/mass_mailing_html_field.test.js
+++ b/addons/mass_mailing/static/tests/mass_mailing_html_field.test.js
@@ -191,7 +191,7 @@ const mailViewArch = `
     <field name="mailing_model_id" invisible="1"/>
     <field name="mailing_model_real" invisible="1"/>
     <field name="state" invisible="1"/>
-    <field name="body_html" class="o_mail_body_inline"/>
+    <field name="body_html" class="o_mail_body_inline" readonly="true"/>
     <field name="body_arch" class="o_mail_body_mailing" widget="mass_mailing_html"
         options="{
             'inline_field': 'body_html',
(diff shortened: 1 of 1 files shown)


addons/mass_mailing/static/tests/mass_mailing_html_field.test.js
thjo-odoo, line 1: Commit message:
>`200` ms is sometimes too little for operations involving multiple async calls
like `save` on the mass_mailing form view, and display the style option in the
`html_builder` when clicking on a block.

`200` ms is sometimes too **short** for operations involving multiple async calls.
E.g. saving the mass_mailing form view or displaying the style option in the
`html_builder` when a block is selected.

> This prevent selection/event conflicts between 2 editors
at the same time which is something that we don't want to test in
`mass_mailing`.

This **prevents** selection/event conflicts between 2 editors
at the same time. Which is already tested elsewhere.

thjo-odoo (approved):

addons/mass_mailing/static/tests/mass_mailing_html_field.test.js
abd-msyukyu-odoo: That's some fine nitpicking if I ever saw some! Thanks, updated.

PR #245173, opened 2026-01-22
[ADD] l10n_tr_nilvera_einvoice: adding return and withholding return …

khfa-odoo:
…invoice scenarios

This commit introduces two new invoicing scenarios to our Nilvera implementaiton. The first scenario is the "Return", which is a credit note for a basic sales invoice. The second scenario is the "Withholding Return", which is a credit note for withholding invoices.

task-id: 4774826

Current behavior before PR:

---

### DIFF
--- a/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
@@ -55,6 +55,10 @@ class AccountEdiXmlUblTr(models.AbstractModel):
 
     def _add_invoice_header_nodes(self, document_node, vals):
         super()._add_invoice_header_nodes(document_node, vals)
+        # The Nilvera Extended flow is only for out_invoice and out_refund type
+        if vals['document_type'] not in {'invoice', 'credit_note'}:
+            return
+
         invoice = vals['invoice']
 
         # Check the customer status if it hasn't been done before as it's needed for profile_id
@@ -78,7 +82,6 @@ class AccountEdiXmlUblTr(models.AbstractModel):
             'cbc:UUID': {'_text': invoice.l10n_tr_nilvera_uuid},
             'cbc:DueDate': None,
             'cbc:InvoiceTypeCode': {'_text': 'ISTISNA' if invoice.l10n_tr_is_export_invoice else invoice.l10n_tr_gib_invoice_type},
-            'cbc:CreditNoteTypeCode': {'_text': 'IADE'} if vals['document_type'] == 'credit_note' else None,
             'cbc:PricingCurrencyCode': {'_text': invoice.currency_id.name.upper()}
                 if vals['currency_id'] != vals['company_currency_id'] else None,
             'cbc:LineCountNumeric': {'_text': len(invoice.line_ids)},
@@ -103,6 +106,31 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         if vals['invoice'].currency_id.name != 'TRY':
             document_node['cbc:Note'].append({'_text': self._l10n_tr_get_amount_integer_partn_text_note(invoice.amount_residual, vals['invoice'].currency_id), 'note_attrs': {}})
             document_node['cbc:Note'].append({'_text': self._l10n_tr_get_invoice_currency_exchange_rate(invoice)})
+
+        if invoice.move_type == "out_refund":
+            # For credit notes, we need to add cac:BillingReference (i.e. reference to the original invoice)
+            self._l10n_tr_add_billing_reference_node(document_node, vals)
+
+    @api.model
+    def _l10n_tr_add_billing_reference_node(self, document_node, vals):
+        invoice = vals['invoice']
+        if not invoice.reversed_entry_id and not invoice.ref:
+            raise UserError(
+                self.env._(
+                    "The credit note must be linked to the original invoice, or include the invoice reference in the reference field."
+                )
+            )
+        _, parts = invoice._get_sequence_format_param(invoice.reversed_entry_id.name or invoice.ref)
+        prefix, year, number = parts['prefix1'][:3], parts['year'], str(parts['seq']).zfill(9)
+        reverese_entry_id = f"{prefix.upper()}{year}{number}"
+        document_node['cac:BillingReference'] = {
+            "cac:InvoiceDocumentReference": {
+                'cbc:ID': {'_text': reverese_entry_id},
+                'cbc:IssueDate': {'_text': invoice.reversed_entry_id.invoice_date or invoice.l10n_tr_original_invoice_date},
+                'cbc:DocumentTypeCode': {'_text': "İADE"},
+                'cbc:DocumentDescription': {'_text': "İade Edilen Fatura"},
+            },
+        }
 
     @api.model
     def _l10n_tr_get_amount_integer_partn_text_note(self, amount, currency):
@@ -167,6 +195,8 @@ class AccountEdiXmlUblTr(models.AbstractModel):
 
     def _add_invoice_payment_means_nodes(self, document_node, vals):
         # EXTENDS account.edi.xml.ubl_21
+        if vals["document_type"] == "credit_note":
+            return
         super()._add_invoice_payment_means_nodes(document_node, vals)
         payment_means_node = document_node['cac:PaymentMeans']
         payment_means_node['cbc:InstructionID'] = None
@@ -219,6 +249,8 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         """
         if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFAT':
             return self._add_withholding_document_line_tax_total_nodes(line_node, vals)
+        if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFATIADE':
+            return self._add_withholding_return_document_line_tax_total_nodes(line_node, vals)
         return super()._add_document_line_tax_total_nodes(line_node, vals)
 
     def _l10n_tr_get_issuer_party_line_codes(self, move_line):
@@ -298,6 +330,8 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         """
         if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFAT':
             return self._add_withholding_document_tax_total_nodes(document_node, vals)
+        if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFATIADE':
+            return self._add_withholding_return_document_tax_total_nodes(document_node, vals)
         return super()._add_document_tax_total_nodes(document_node, vals)
 
     def _add_withholding_document_tax_total_nodes(self, line_node, vals):
@@ -361,6 +395,73 @@ class AccountEdiXmlUblTr(models.AbstractModel):
                 for tax_details in aggregated_tax_details_by_l10n_tr_tax_withholding_code_id['withholding_tax'].values()
             ]
 
+    def _add_withholding_return_document_tax_total_nodes(self, document_node, vals):
+        tax_subtotals_per_currency = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: 0)))
+        for base_line in vals["base_lines"]:
+            line = base_line["record"]
+            taxes_data = base_line.get("tax_details", {}).get("taxes_data", [])
+            total_taxed_amount = sum(tax_line["tax_amount"] for tax_line in taxes_data)
+
+            percent_return = (
+                line.l10n_tr_original_quantity
+                and line.quantity * 100 / line.l10n_tr_original_quantity
+            ) or 0
+            tax_subtotals_per_currency[line.currency_id][percent_return][
+                "taxable_amount"
+            ] += line.l10n_tr_original_tax_without_withholding
+            tax_subtotals_per_currency[line.currency_id][percent_return][
+                "tax_amount"
+            ] += total_taxed_amount
+
+        tax_totals = []
+        for currency_id, subtotal_aggregates in tax_subtotals_per_currency.items():
+            currency_name = currency_id.name
+            precision = currency_id.decimal_places
+            tax_amount = 0
+            tax_subtotals = []
+            for percent_return, subtotal in subtotal_aggregates.items():
+                tax_amount += subtotal["tax_amount"]
+                tax_subtotals.append(
+                    {
+                        "cbc:TaxableAmount": {
+                            "_text": self.format_float(
+                                subtotal["taxable_amount"],
+                                precision,
+                            ),
+                            "currencyID": currency_name,
+                        },
+                        "cbc:TaxAmount": {
+                            "_text": self.format_float(
+                                subtotal["tax_amount"],
+                                precision,
+                            ),
+                            "currencyID": currency_name,
+                        },
+                        "cbc:Percent": {
+                            "_text": percent_return,
+                        },
+                        "cac:TaxCategory": {
+                            "cac:TaxScheme": {
+                                "cbc:Name": {
+                                    "_text": "Gerçek Usulde KDV",
+                                },
+                                "cbc:TaxTypeCode": {
+                                    "_text": "0015",
+                                },
+                            },
+                        },
+                    },
+                )
+            tax_totals.append({
+                "cbc:TaxAmount": {
+                    '_text': self.format_float(tax_amount, precision),
+                    'currencyID': currency_name,
+                },
+                "cac:TaxSubtotal": tax_subtotals,
+            })
+
+        document_node["cac:TaxTotal"] = tax_totals
+
     def _add_withholding_document_line_tax_total_nodes(self, line_node, vals):
         """Extend aggregation of line tax details to include Turkish (TR) withholding-specific amounts.
 
@@ -415,6 +516,56 @@ class AccountEdiXmlUblTr(models.AbstractModel):
                 for tax_details in aggregated_tax_details_by_l10n_tr_tax_withholding_code_id['withholding_tax'].values()
             ]
 
+    def _add_withholding_return_document_line_tax_total_nodes(self, line_node, vals):
+        base_line = vals["base_line"]
+        line = base_line["record"]
+        currency_name = line.currency_id.name
+        precision = line.currency_id.decimal_places
+        percent_return = (
+            line.l10n_tr_original_quantity
+            and line.quantity * 100 / line.l10n_tr_original_quantity
+        ) or 0
+
+        taxes_data = base_line.get("tax_details", {}).get("taxes_data", [])
+        total_taxed_amount = sum(tax_line["tax_amount"] for tax_line in taxes_data)
+        line_node["cac:TaxTotal"] = [
+            {
+                "cbc:TaxAmount": {
+                    "_text": self.format_float(total_taxed_amount, precision),
+                    "currencyID": currency_name,
+                },
+                "cac:TaxSubtotal": {
+                    "cbc:TaxableAmount": {
+                        "_text": self.format_float(
+                            line.l10n_tr_original_tax_without_withholding,
+                            precision,
+                        ),
+                        "currencyID": currency_name,
+                    },
+                    "cbc:TaxAmount": {
+                        "_text": self.format_float(
+                            total_taxed_amount,
+                            precision,
+                        ),
+                        "currencyID": currency_name,
+                    },
+                    "cbc:Percent": {
+                        "_text": percent_return,
+                    },
+                    "cac:TaxCategory": {
+                        "cac:TaxScheme": {
+                            "cbc:Name": {
+                                "_text": "Gerçek Usulde KDV",
+                            },
+                            "cbc:TaxTypeCode": {
+                                "_text": "0015",
+                            },
+                        },
+                    },
+                },
+            },
+        ]
+
     def _get_address_node(self, vals):
         partner = vals['partner']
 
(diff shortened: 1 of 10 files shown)


addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -219,6 +249,8 @@ def _add_document_line_tax_total_nodes(self, line_node, vals):
         """
         if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFAT':
             return self._add_withholding_document_line_tax_total_nodes(line_node, vals)
+        if vals["invoice"].l10n_tr_gib_invoice_type == 'TEVKIFATIADE':
```
malb-odoo:
```suggestion
        if vals['invoice'].l10n_tr_gib_invoice_type == 'TEVKIFATIADE':
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -328,6 +362,75 @@ def _add_withholding_document_tax_total_nodes(self, line_node, vals):
                 for tax_details in aggregated_tax_details_by_l10n_tr_tax_withholding_code_id['withholding_tax'].values()
             ]
 
+    def _add_withholding_return_document_tax_total_nodes(self, document_node, vals):
+        tax_subtotals_per_currency = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: 0)))
```
malb-odoo, line 399: wtff ahah, smell like a bad design ahah

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -328,6 +362,75 @@ def _add_withholding_document_tax_total_nodes(self, line_node, vals):
...
+        for base_line in vals["base_lines"]:
+            line = base_line["record"]
+            taxes_data = base_line.get("tax_details", {}).get("taxes_data", [])
+            total_taxed_amount = sum(tax_line["tax_amount"] for tax_line in taxes_data)
+
+            percent_return = (
+                base_line["record"].quantity
```
malb-odoo: either use the line variable or remove it ?

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -104,6 +107,31 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+    def _l10n_tr_add_billing_reference_node(self, document_node, vals):
+        invoice = vals['invoice']
+        if not invoice.reversed_entry_id and not invoice.ref:
+            raise UserError(
+                self.env._(
+                    "The credit note must be linked to the original invoice, or include the invoice reference in the reference field."
+                )
+            )
+        _, parts = invoice._get_sequence_format_param(invoice.reversed_entry_id.name or invoice.ref)
+        prefix, year, number = parts['prefix1'][:3], parts['year'], str(parts['seq']).zfill(9)
```
malb-odoo, line 124: legal requirement ?

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -167,6 +195,8 @@ def _l10n_tr_get_invoice_currency_exchange_rate(self, invoice):
 
     def _add_invoice_payment_means_nodes(self, document_node, vals):
         # EXTENDS account.edi.xml.ubl_21
+        if vals["document_type"] == "credit_note":
+            return
```
malb-odoo, line 199: sure we want to return completely and not using the super ?

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -328,6 +362,75 @@ def _add_withholding_document_tax_total_nodes(self, line_node, vals):
...
+                        },
+                        "cbc:TaxAmount": {
+                            "_text": self.format_float(
+                                subtotal["tax_amount"],
+                                precision,
+                            ),
+                            "currencyID": currency_name,
+                        },
+                        "cbc:Percent": {
+                            # The percentatge can't be a float as it is expected to return as an integer
```
malb-odoo:
```suggestion
                            # The percentage can't be a float as it is expected to return as an integer
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -104,6 +106,7 @@ class AccountMove(models.Model):
         string="Partner Nilvera Status",
         related='partner_id.l10n_tr_nilvera_customer_status',
     )
+    l10n_tr_original_invoice_date = fields.Date(string="Original Invoice Date")
```
malb-odoo, line 118: Just to be sure we don't need default value ?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -549,3 +571,23 @@ def _cron_nilvera_get_sale_pdf(self, batch_size=100):
...
+
+        When creating a credit note, an R is added by standard, so
+        we remvoe the first letter of the journal prefix to make sure it
+        remains 3 characters (e.g., RINV → RNV).
+        """
+        starting_sequence = super()._get_starting_sequence()
+        if (
+            self.company_id.country_id.code == "TR"
+            and self.journal_id.refund_sequence
+            and self.move_type in ("out_refund", "in_refund")
```
```suggestion
            and self.move_type in {"out_refund", "in_refund"}
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -431,10 +586,26 @@ def _get_document_template(self, vals):
...
         """
-        if vals['document_node']['cbc:CustomizationID']['_text'] == 'TR1.2' and vals['document_type'] == 'invoice':
+        if vals['document_node']['cbc:CustomizationID']['_text'] == 'TR1.2' and vals['document_type'] in {'invoice', 'credit_note'}:
             return TrInvoice
         return super()._get_document_template(vals)
 
+    def _get_document_nsmap(self, vals):
+        res = super()._get_document_nsmap(vals)
+        if vals["document_type"] == "credit_note":
+            res[None] = "urn:oasis:names:specification:ubl:schema:xsd:Invoice-2"
```
malb-odoo, line 625: if i'm not mistaken res[None] is a dict, so don't we want to change the value of a key ?

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -12,6 +12,29 @@ class AccountMoveLine(models.Model):
...
+    l10n_tr_original_quantity = fields.Float(
+        string='Original Quantity',
+        digits='Product Unit',
+        help="The quantity originally sold for this product.",
+    )
+
+    l10n_tr_original_tax_without_withholding = fields.Monetary(
+        string="Original Tax After Withholding",
+        help="Taxes collected after deducting withholding taxes for this product on the original invoice.",
+    )
```
malb-odoo, line 40: Rule of thumb of quotes here, either you try to stay consistent or you can follow this:
Double quotes for thing the user can see, single otherwise

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -12,6 +12,29 @@ class AccountMoveLine(models.Model):
...
+
+    @api.depends("l10n_tr_original_line_id")
+    def _compute_l10n_tr_original_quantity(self):
+        for line in self.filtered(lambda r: r.l10n_tr_original_line_id and r.move_id.state == "draft" and not r.l10n_tr_original_quantity):
```
malb-odoo: Since we could have a lot of lines in self. I think it would be better to use an if instead of a filtered otherwise you will traverse self to filter element and then again the list with filtered element

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -12,6 +12,29 @@ class AccountMoveLine(models.Model):
...
+            line.l10n_tr_original_quantity = line.l10n_tr_original_line_id.quantity
+
+    @api.depends("l10n_tr_original_line_id")
+    def _compute_l10n_tr_original_tax_without_withholding(self):
+        for line in self.filtered(lambda r: r.l10n_tr_original_line_id and r.move_id.state == "draft" and not r.l10n_tr_original_tax_without_withholding):
```
malb-odoo: same

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -152,8 +154,93 @@ def _is_valid_nilvera_name(move):
...
+        ):
+            alerts["tr_withholding_credite_note_tax_mismatch"] = (
+                tr_withholding_credit_note_tax_mismatch
+            )
+
+        if tr_credit_note_invoice_not_sent := tr_nilvera_moves.filtered(
+            lambda r: r.move_type == "out_refund"
+            and r.reversed_entry_id
+            and r.reversed_entry_id.l10n_tr_nilvera_send_status
+            not in ["sent", "waiting", "succeed"]
```
malb-odoo:
```suggestion
            not in {"sent", "waiting", "succeed"}
```

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -152,8 +154,93 @@ def _is_valid_nilvera_name(move):
...
+                "action_text": _("View Invoice(s)"),
+                "action": tr_credit_note_invoice_not_sent.reversed_entry_id._get_records_action(
+                    name=_("View Invoice(s)")
+                ),
+            }
         return alerts
 
+    def _l10n_tr_withholding_credit_note_tax_mismatch(self, moves):
+        errors = defaultdict(defaultdict)
+        for invoice in moves.filtered(
```
malb-odoo, line 235: same for the filtered here

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -152,8 +154,93 @@ def _is_valid_nilvera_name(move):
...
+                    error_message += "- [%s] - %s: %s\n" % (invoice, line, error)
+
+            error_message += _("Make sure the price and taxes match the original invoice.")
+            return {
+                'level': 'danger',
+                'message': error_message,
+                'action_text': _("Check Tax Amounts on Credit Note(s)"),
+                'action': moves._get_records_action(name=_("Check Tax Amounts on Credit Note(s)")),
+            }
+        return ""
```
malb-odoo: what about doing the opposite "if not errors: return "" so that's it's a bit more readable

addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
```diff
@@ -63,12 +63,24 @@
             </xpath>
             <xpath expr="//field[@name='invoice_line_ids']//list//field[@name='tax_ids']" position="after">
                 <field name="l10n_tr_ctsp_number" optional="hide"/>
+                <field name="l10n_tr_original_quantity"
+                    required="parent.l10n_tr_gib_invoice_type == 'TEVKIFATIADE'"
+                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'" />
```
malb-odoo:
```suggestion
                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'"/>
```

addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
```diff
@@ -63,12 +63,24 @@
...
+                <field name="l10n_tr_original_tax_without_withholding"
+                    required="parent.l10n_tr_gib_invoice_type == 'TEVKIFATIADE'"
+                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'" />
```
malb-odoo:
```suggestion
                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'"/>
```

addons/l10n_tr_nilvera_einvoice/wizard/account_move_reversal.py
```diff
@@ -0,0 +1,28 @@
+from odoo import Command, models
+
+
+class AccountMoveReversal(models.TransientModel):
+    _inherit = 'account.move.reversal'
+
+    def _prepare_default_reversal(self, move):
+        res = super()._prepare_default_reversal(move)
+        if self.country_code == 'TR':
```
malb-odoo: we could invert to gain readability also

addons/l10n_tr_nilvera_einvoice/wizard/account_move_reversal_views.xml
```diff
@@ -0,0 +1,38 @@
...
+            <field name="context">{'dialog_size': 'medium'}</field>
+        </record>
+
+        <record id="helpdesk_ticket_action_refund_form" model="ir.actions.act_window.view">
+            <field name="sequence" eval="1"/>
+            <field name="view_mode">form</field>
+            <field name="view_id" ref="view_account_move_reversal_inherit_helpdesk_account"/>
+            <field name="act_window_id" ref="helpdesk_ticket_action_refund"/>
+        </record>
+
```
malb-odoo:
```suggestion
```

malb-odoo (commented):
First quick path on this :smile: Thanks for the work !

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
khfa-odoo: Yes, the credit note must be linked to an original invoice.

In Nilver,a this format is "ABC2009123456789", but we expect the user to enter Odoo's format "ABC/2009/123456789".  If there is neither an original invoice nor a reference to it, then we raise the error

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
khfa-odoo: According to Nilvera's credit note xml samples, there are never payment means nodes. When I was testing initially, having payment means raised errors from Nilvera. I will verify again with the PO

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
khfa-odoo: https://github.com/odoo/odoo/blob/44e0c0175aec7175aa6874dcc8afee600b1ab143/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py#L73-L83

It is indexed by vals["document_type"] before returning, so in credit_note, it would be

```py
{
 None : "urn:oasis:names:specification:ubl:schema:xsd:CreditNote-2"
}
```
this is wrong because Nilvera's credit note specification is the same as the invoice's

addons/l10n_tr_nilvera_einvoice/models/account_move.py
khfa-odoo: We have a validation error to ensure this field is set in the case the credit note was created manually. If we use a default value, then it will always pass, and the user won't notice

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
khfa-odoo: I think in terms of number of operations, filtering operation is equivalent to its counterpart if condition. However, filtered does some more stuff than just a simple if, so I would agree that an if is a bit faster. The choice for the filtered is for code simplicity and using ORM functions.

I could still change it if you insist.

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
khfa-odoo: will do, but wouldn't the indexing time of a set be as expensive as searching in the array (assuming we only search one)

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
khfa-odoo: To be honest, I kinda agree, but it is a highly customized flow, and is only used for Nilvera's Withholding Returns  specifically, so it is not expected to be overridden.

The reason it had to be done this way instead of using `_aggregate_base_line_tax_details` with a custom grouping function is that the aggregates themselves are also custom and specific to this flow.

In Nilvera's Withholding Return, 

Taxable Amount : is the original's invoice tax amount after deducting withholding amounts
Tax Percentage: the percentage of the original's invoice quantity being returned

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
malb-odoo: Yeah but that means that if we have multiple companies that extend that, we could fucked up their ubl, so i suppose it would be better to call the super before

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
malb-odoo: it's a minor nitpick but since we are touching move lines it could be better perf wise, now i will not block for that

malb-odoo (commented):
Could we add some tests there to bullet proof the changes :smile:

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
khfa-odoo: but this is only for the `account.edi.xml.ubl.tr` model, so it shouldn't affect other localizations

https://github.com/odoo/odoo/blob/f6e3f07e4f5e4231f6ff9c7a4f16d9b32553c1c4/addons/l10n_tr_nilvera/models/res_partner.py#L148-L152

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -3,7 +3,7 @@
 from markupsafe import Markup
 from urllib.parse import quote, urlencode, urlparse
 
-from odoo import _, api, fields, models
+from odoo import _, api, Command, fields, models
```
malb-odoo: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports
```suggestion
from odoo import Command, _, api, fields, models
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -71,6 +71,8 @@ class AccountMove(models.Model):
             ('TEVKIFAT', "Withholding"),
             ('IHRACKAYITLI', "Registered for Export"),
             ('ISTISNA', "Tax Exempt"),
+            ("IADE", "Return"),
+            ("TEVKIFATIADE", "Withholding Return"),
```
malb-odoo: To be consistent
```suggestion
            ('IADE', "Return"),
            ('TEVKIFATIADE', "Withholding Return"),
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -533,3 +555,53 @@ def _cron_nilvera_get_sale_pdf(self, batch_size=100):
...
+
+    def _get_starting_sequence(self):
+        """
+        Generate a valid name for credit notes.
+
+        Nilvera requires invoice names in the format:
+        <3 alphanumeric characters>/<year>/<sequence number>.
+
+        When creating a credit note, an R is added by standard, so
+        we remvoe the first letter of the journal prefix to make sure it
```
```suggestion
        we remove the first letter of the journal prefix to make sure it
```

addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
```diff
@@ -65,12 +65,24 @@
                 <field name="l10n_tr_ctsp_number" optional="hide"/>
                 <field name="l10n_tr_seller_line_code" optional="hide"/>
                 <field name="l10n_tr_customer_line_code" optional="hide"/>
+                <field name="l10n_tr_original_quantity"
+                    required="parent.l10n_tr_gib_invoice_type == 'TEVKIFATIADE'"
+                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'"/>
+                <field name="l10n_tr_original_tax_without_withholding"
+                    required="parent.l10n_tr_gib_invoice_type == 'TEVKIFATIADE'"
+                    column_invisible="parent.l10n_tr_gib_invoice_type != 'TEVKIFATIADE'"/>
```
malb-odoo: small nitpick but could you align the parameter ahah

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -26,6 +26,29 @@ class AccountMoveLine(models.Model):
...
+        help="The quantity originally sold for this product.",
+    )
+
+    l10n_tr_original_tax_without_withholding = fields.Monetary(
+        string="Original Tax After Withholding",
+        help="Taxes collected after deducting withholding taxes for this product on the original invoice.",
+    )
+
+    @api.depends("l10n_tr_original_line_id")
+    def _compute_l10n_tr_original_quantity(self):
```
malb-odoo: not declared in the field ? does that work ahah

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -26,6 +26,29 @@ class AccountMoveLine(models.Model):
...
+        for line in self.filtered(lambda r: r.l10n_tr_original_line_id and r.move_id.state == "draft" and not r.l10n_tr_original_quantity):
+            line.l10n_tr_original_quantity = line.l10n_tr_original_line_id.quantity
+
+    @api.depends("l10n_tr_original_line_id")
+    def _compute_l10n_tr_original_tax_without_withholding(self):
```
malb-odoo: same here

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
khfa-odoo: oh this is embarrassing, the field isn't computed anymore, but forgot to delete the function :laughing:

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -71,6 +71,8 @@ class AccountMove(models.Model):
             ('TEVKIFAT', "Withholding"),
             ('IHRACKAYITLI', "Registered for Export"),
             ('ISTISNA', "Tax Exempt"),
+            ('IADE', 'Return'),
+            ('TEVKIFATIADE', 'Withholding Return'),
```
malb-odoo: Should be like this 😂 
```suggestion
            ('IADE', "Return"),
            ('TEVKIFATIADE', "Withholding Return"),
```

malb-odoo (approved):
Some tests are failing too 😬, once change i will r+ it

addons/l10n_tr_nilvera_einvoice/models/account_move.py
khfa-odoo: oh, sorry for that :sob:

yace-odoo:
Hi @malb-odoo, can you take a look at this? would appreciate this to make it to 19.2

malb-odoo (approved):
Let's goooo theeen
@robodoo r+

PR #245175, opened 2026-01-22
[FIX] website_sale: hide pills variant when single value

xlu-odoo:
Before this commit, when a product attribute with display_type 'pills' has only one non-custom value, the parent li.variant_attribute is correctly hidden. However, the CSS :has() selector still matched the inner li.o_variant_pills, causing the attributes section to display with empty borders.

This commit refines the selector to only match pills outside of variant_attribute elements (UOM pills).

task-5852515

| Current (19.0) | After |
|--------|--------|
| image | image |

---

### DIFF
--- a/addons/website_sale/static/src/scss/website_sale.scss
+++ b/addons/website_sale/static/src/scss/website_sale.scss
@@ -1294,7 +1294,7 @@ $-product-radius-map: (
     .o_wsale_product_details_content_section_attributes {
         display: none;
 
-        &:where(:has(li.variant_attribute:not(.d-none))), &:where(:has(li.o_variant_pills:not(.d-none))) {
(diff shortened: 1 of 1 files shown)


xlu-odoo:
Hello @chevalierv 👋 
Small one here ready for tech review, could you please check when you have some time? 
Thanks!

ltinel (changes requested):
Hello, thanks for your work :slightly_smiling_face: However, I'm not a big fan of this fix. It's a bit odd to nest `d-none` classes with the same condition. Would it be possible to update the failing selector instead?

ltinel (approved):
Thanks for the fix, LGTM :slightly_smiling_face: 
@robodoo r+

PR #245177, opened 2026-01-22
[FIX] pos_self_order: fix ticket font size for preparation printers in kiosk

lcantraine:
Before this commit, preparation tickets printed from kiosk had a different structure from those printed from the POS.

Two differences were spotted:
1. Font size was smaller than ticket printed from the pos
2. The name of the product displayed was the full product name instead of the basic one with attributes under the product name.

So, to fix that I basically used the same template structure from the point_of_sale module and adapted data to fit with.

task : 5506329

---

### DIFF
--- a/addons/pos_self_order/static/src/app/store/order_change_receipt_template.xml
+++ b/addons/pos_self_order/static/src/app/store/order_change_receipt_template.xml
@@ -25,11 +25,18 @@
             <div class="pos-receipt-title">
                 NEW
             </div>
-            <br />
             <t t-foreach="changes.new" t-as="change" t-key="change_index">
-                <div class="product-details d-flex" t-att-class="{'ms-5': change.combo_parent_id}">
-                    <span class="product-quantity me-3 mb-1" t-esc="change.qty"/>
-                    <span class="product-name" t-esc="change.full_product_name"/>
+                <div class="product-details d-flex mx-1" t-att-class="{'ms-5': change.combo_parent_id}" style="font-size: 150%;">
+                    <span class="product-quantity me-3" t-esc="change.qty"/>
+                    <span class="product-name" t-esc="change.product_id.name"/>
+                </div>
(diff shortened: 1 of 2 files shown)


addons/pos_self_order/static/src/app/store/order_change_receipt_template.xml
```diff
@@ -26,23 +26,27 @@
...
-                <t t-if="change.customer_note">
-                    <div>
-                        NOTE
-                        <span class="pos-receipt-right-align">...</span>
-                    </div>
-                    <div><span class="pos-receipt-left-padding">--- <t t-esc="change.customer_note" /></span></div>
-                    <br/>
-                </t>
+            <t t-foreach="changes.new" t-as="line" t-key="line_index">
+                <t t-call="pos_self_order.OrderChangeReceiptLine"/>
```
stva-odoo: In stable, we should keep this in the same template, since a customer module may override it.
And we should keep the note, we never know ;-) 
```suggestion
            <div t-attf-class="product-details orderline #{line.combo_parent_uuid ? 'ms-5 px-2' : 'mx-1'}" style="font-size: 150%;">
            <div class="d-flex medium">
                <span class="product-quantity me-3" t-esc="line.qty"/> <span class="product-name" t-esc="line.basic_name"/>
            </div>
            <div t-if="line.attribute_value_names?.length" class="ms-5" style="font-size: 91%;">
                <t t-foreach="line.attribute_value_names" t-as="name" t-key="name_index">
                    <p class="p-0 m-0">
                        - <t t-esc="name" />
                    </p>
                </t>
            </div>
        </div>
        </t>
         <t t-if="change.customer_note">
                    <div>
                        NOTE
                        <span class="pos-receipt-right-align">...</span>
                    </div>
                    <div><span class="pos-receipt-left-padding">--- <t t-esc="change.customer_note" /></span></div>
                    <br/>
          
```

stva-odoo (commented):
robodoo r+

stva-odoo:
robodoo r+

PR #245179, opened 2026-01-22
[FIX] l10n_tw_edi_ecpay_website_sale: add post-init hook to manifest

avar-odoo:
The extra invoicing info step for EDI was unpublished and hidden on ecommerce. This commit fixes that.

Task-5493138

---


vin-odoo (approved):
robodoo r+

Thank you!

PR #245180, opened 2026-01-22
[IMP] account: sanitize account number search, with/without spaces

pssh-odoo:
Before this commit:
- Searching on the `account_number` field failed when the search query contained
  spaces or separators ( `'-', ' ', '.'`).

After this commit:
- Added a stored computed field `sanitized_account_number`, containing the 
  sanitized value of `account_number`.
- The search on `account_number` sanitizes the search value and performs the
  lookup based on the `sanitized_account_number` field.

~~Related: https://github.com/odoo/enterprise/pull/105136~~
Related: https://github.com/odoo/upgrade/pull/9360
task-4872851

### DIFF
--- a/addons/account/models/account_bank_statement_line.py
+++ b/addons/account/models/account_bank_statement_line.py
@@ -5,6 +5,7 @@ from odoo.fields import Command, Domain
 from xmlrpc.client import MAXINT
 
 from odoo.tools import SQL
+from odoo.addons.base.models.res_partner_bank import sanitize_account_number
 
 
 class AccountBankStatementLine(models.Model):
@@ -63,8 +64,13 @@ class AccountBankStatementLine(models.Model):
         check_company=True)
 
     # Technical field used to store the bank account number before its creation, upon the line's processing
-    account_number = fields.Char(string='Bank Account Number')
+    account_number = fields.Char(string='Bank Account Number', search='_search_account_number')
 
+    sanitized_account_number = fields.Char(
+        string='Sanitized Bank Account Number',
+        compute='_compute_sanitized_account_number',
+        readonly=True, store=True,
+    )
     # This field is used to record the third party name when importing bank statement in electronic format,
     # when the partner doesn't exist yet in the database (or cannot be found).
     partner_name = fields.Char(index='btree_not_null')
@@ -308,6 +314,11 @@ class AccountBankStatementLine(models.Model):
                 # The journal entry seems reconciled.
                 st_line.is_reconciled = True
 
+    @api.depends('account_number')
+    def _compute_sanitized_account_number(self):
+        for account in self:
+            account.sanitized_account_number = sanitize_account_number(account.account_number)
+
     # -------------------------------------------------------------------------
     # CONSTRAINT METHODS
     # -------------------------------------------------------------------------
@@ -447,6 +458,13 @@ class AccountBankStatementLine(models.Model):
             for group_line in result:
                 group_line['running_balance'] = self.search(group_line['__extra_domain'] + domain, limit=1).running_balance or 0.0
         return result
+
+    def _search_account_number(self, operator, value):
+        if operator in ('in', 'not in'):
+            value = [sanitize_account_number(i) for i in value]
+        else:
+            value = sanitize_account_number(value)
+        return [('sanitized_account_number', operator, value)]
 
     # -------------------------------------------------------------------------
     # ACTION METHODS


khva-odoo (commented):
Hello :wave: 
I think may be beneficial to introduce an additional field for this purpose. As we are already sanitizing the stored account number, keeping the sanitized value in a separate field would be a more appropriate approach. Since anyhow we need the sanitised  account number to compare it with the value

addons/account/models/account_bank_statement_line.py
```diff
@@ -63,8 +64,13 @@ class AccountBankStatementLine(models.Model):
...
 
     # Technical field used to store the bank account number before its creation, upon the line's processing
-    account_number = fields.Char(string='Bank Account Number')
+    account_number = fields.Char(string='Bank Account Number', search='_search_account_number')
 
+    sanitized_account_number = fields.Char(
+        string='Sanitized Bank Account Number',
+        compute='_compute_sanitized_account_number',
+        readonly=True, store=True,
+    )
```
SaddemAmine: Wondering if it makes more sense to just use res.partner.bank
It already has both of these fields

addons/account/models/account_bank_statement_line.py
```diff
@@ -63,8 +64,13 @@ class AccountBankStatementLine(models.Model):
         check_company=True)
 
     # Technical field used to store the bank account number before its creation, upon the line's processing
-    account_number = fields.Char(string='Bank Account Number')
+    account_number = fields.Char(string='Bank Account Number', search='_search_account_number')
```
FlorianGilbert, line 67: Just asking, why don't we sanitize the value at the create/write in this field and use a search method like you did to sanitize the value we got in search?
By doing that, we avoid a new stored field. What do you think about that?

jco-odoo: It is like we had sanitized number before, it was removed and we have the sanitize vals in the write/create already.  However if you want to search, you want to search filthy and sanitized at the same time probably.

jco-odoo: Do not we rather inherit _search instead of defining a search method if it is about a stored field?  (I have not seen many there)

pssh-odoo: >What do you think about that?

I had previously gone with that approach, but the main drawback I encountered was that we wouldn’t store what the user actually typed. For example, during imports we would only store the sanitized values.
Additionally, if a user enters the account number with spaces or dashes, those would be removed during sanitization, which would affect the readability of the field.

>Do not we rather inherit _search instead of defining a search method if it is about a stored field?

Yes, however we have done something similar in `res.partner.bank` for the account number field, which is where I took the reference from. [See](https://github.com/odoo/odoo/blob/551ef8ffe5076d79fa23fc6a59722798974b020a/odoo/addons/base/models/res_partner_bank.py#L27)

pssh-odoo: @FlorianGilbert , @jco-odoo Just a reminder in case my previous reply was missed.

jco-odoo: @pssh-odoo I see your point.  You are actually doing the same as what is done on res.partner.bank.  In that regard, it is legit.  (but we might need an upgrade script to precompute the values?)  --> upgrade script is there

jco-odoo: However @pssh-odoo, I asked around and have not found any good reason to store the filthy number (like in the account_number field).  Could you create a branch where we keep only the sanitized number for both res.partner.bank and this?  So, we could have just one field account_number which already has the sanitized number and we remove the sanitized_account_number.  Instead of overriding create/write as @FlorianGilbert suggests, you could use an inverse as in base_vat: vat = fields.Char(inverse="_inverse_vat", store=True) where we kind of have a similar mechanism.   @khva-odoo

khva-odoo: I think @hsal-odoo could help us here as he was the one who introduced the field `sanitized_account_number` in `res.partner.bank`.

jco-odoo: @khva-odoo No, he just renamed.  These are old things that can have a fresh take, so please go rip it apart.

pssh-odoo: @jco-odoo Won’t the readability of the `account_number` be impacted if we always show the sanitized version to the user?

jco-odoo: If that is the case, we could just have a non-stored field with the formatted value.  At least it would be formatted consistently.

pssh-odoo:
Closing this PR as this will be handled over [here](https://github.com/odoo/odoo/pull/258860).

PR #245183, opened 2026-01-22
[FIX] l10n_dk: prevent error when unlinking bank and cash accounts

aamo-odoo:
This error occurs when attempting to delete the `Bank and Cash` account type record.

Steps to reproduce:
- Install `l10n_dk` and `Accounting` module(with demo data) > Switch to `DK Company`
- Go to `Chart of Accounts` and groupby by `Account Types` 
- Try to delete `Bank and Cash` record (if warning occurs for audit trail then disable `audit trail` in settings)

Traceback:
`AttributeError: 'account.account' object has no attribute 'read_group'`

This error occurs because `read_group` is used in the code, but it was deprecated in [commit] and replaced by `_read_group`.

[commit]: https://github.com/odoo/odoo/pull/163300/commits/191ac027dd1e5133b0bcc72ab44358431c508a1c

sentry-7207582551

### DIFF
--- a/addons/l10n_dk/models/account_account.py
+++ b/addons/l10n_dk/models/account_account.py
@@ -7,26 +7,27 @@ from odoo.exceptions import UserError
 class AccountAccount(models.Model):
     _name = 'account.account'
     _inherit = 'account.account'
-    
+
     @api.ondelete(at_uninstall=False)
     def _unlink_bank_cash_accounts(self):
         if self.env.context.get('force_delete'):
             return
         nb_account_to_delete_per_company = defaultdict(self.env['account.account'].browse)
         for account in self:
-            for company in account.company_ids:
-                if company.country_code == 'DK':
-                    nb_account_to_delete_per_company[company] |= account
+            for company in account.company_ids.filtered(lambda c: c.country_code == 'DK'):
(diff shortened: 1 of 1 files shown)


addons/l10n_dk/models/account_account.py
```diff
@@ -7,24 +7,26 @@
...
+            for company in account.company_ids.filtered(lambda c: c.country_code == 'DK'):
+                nb_account_to_delete_per_company[company] |= account
 
         if not nb_account_to_delete_per_company:
             return
 
-        grouped_counts = self.read_group(
-            domain=[('company_ids.account_fiscal_country_id.code', '=', 'DK'), ('account_type', '=', 'asset_cash')],
-            fields=['company_ids', 'id:count'],
+        grouped_counts = self._read_group(
```
maan-odoo: I think there is no need the `grouped_counts` variable since we are just converting this variable data to dict. It's good to combine both lines and use one variable.

```
dict(self._read_group(
))
```

aamo-odoo: I did this for better readability and understanding. 🙂

maan-odoo (approved):

oco-odoo (approved):
@robodoo r+

PR #157362, opened 2024-03-12
[IMP] l10n_nl: Added tags needed for PL and BS tags reports

hsal-odoo:
In the Netherlands localization, the Balance Sheet and Profit&Loss reports are only currently built on account codes. We want to add another version of the accounting reports that uses account tags.

Enterprise PR: odoo/enterprise#58487
Task link: https://www.odoo.com/web#model=project.task&id=3794536
task-3794536

PR #245185, opened 2026-01-22
[FIX] l10n_nl: correct labels for COGS and SALE

khsr-odoo:
**Steps to reproduce:**
1. Install l10n_nl_report.
2. Go to Accounting → Configuration  →  Accounting Reports.
3. Open Profit and loss report (tags).
4. Click on Cost of Goods Sold or Cost of Sales and check the related codes.

**Issue:**
In the Dutch Profit and Loss report, the "Cost of Goods Sold" and "Cost of Sales" names are swapped.
According to Dutch accounting standards:
- Cost of goods sold should represent direct costs ([mapping to NL_COGS / 7xxx](https://github.com/odoo/odoo/blob/master/addons/l10n_nl/data/template/account.account-nl.csv#L288-L317)).
- Cost of sales should represent indirect selling costs ([mapping to NL_SALE/45xx](https://github.com/odoo/odoo/blob/master/addons/l10n_nl/data/template/account.account-nl.csv#L249-L260))

**Cause:**
In [PR : #157362 ](https://github.com/odoo/odoo/pull/157362/files#diff-1ba769f7e47f7a941e45b6c662ec2a0558272e9fe8bdbfed368f261c8e43aa5dR55), The names were swapped. While the technical codes and expressions correctly targeted the intended account ranges

**Solution:**
Updated account_tag_7 and account_tag_10 to match corrected naming convention.

Related Enterprise PR : https://github.com/odoo/enterprise/pull/103150

**opw-5407885**

### DIFF
--- a/addons/l10n_nl/__manifest__.py
+++ b/addons/l10n_nl/__manifest__.py
@@ -3,7 +3,7 @@
     'name': 'Netherlands - Accounting',
     'icon': '/account/static/description/l10n.png',
     'countries': ['nl'],
-    'version': '3.4',
+    'version': '3.5',
     'category': 'Accounting/Localizations/Account Charts',
     'author': 'Onestein (http://www.onestein.eu)',
     'website': 'https://www.odoo.com/documentation/17.0/applications/finance/fiscal_localizations/netherlands.html',
--- a/addons/l10n_nl/data/account_account_tag.xml
+++ b/addons/l10n_nl/data/account_account_tag.xml
@@ -36,7 +36,7 @@
             <field name="country_id" ref="base.nl"/>
         </record>
         <record id="account_tag_7" model="account.account.tag">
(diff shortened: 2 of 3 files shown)


Megaaaaaa (changes requested):
Hello :wave:

Please put the PR ready for review so that the upgrade team gets pinged for review.

Talking about upgrade, your script doesn't run currently which is the reason for your ci/upgrade_enterprise to be red. You created an upgrade script for the version 3.5 of the module but the manifest still says 3.4. You need to up that version so the script runs.

Apart from that, I'm good with the change :+1:

Megaaaaaa (commented):
Hello :wave:

Is this still required ? It's not been moving for a while now and I suggest we close the PR if there's no need for a change. The issue gets fixed in 19.0 anyway where the translation in dutch is correct and the issue remains a translation one.

PR #245189, opened 2026-01-22
[FIX] mrp: make byproduct cost share constraint variant dependant

lase-odoo:
### Steps to reproduce:

- In the settings enable By-Products
- Create a product with an color attribute and 2 values: white, black
- Create a bom for that products and add 2 by product lines:
    - 1 x comp1 with a cost_share of 50% specific to the white att-value
    - 1 x comp2 with a cost_share of 70% specific to the Black att-value
> Try to save and you will raise a UserError: The total cost share for a BoM's by-products cannot exceed 100.

### Expected behavior:

The error should not be raised as the total cost_share is 50% for the white variant and 70% for the black one but none of them exceeds the 100% cost share.

### Cause of the Issue:

Currently the constraint does not take attribute values into accounts and simply sums the value of the cost share of all by-product lines: https://github.com/odoo/odoo/blob/bcc1397c7d694dbe61ecbd44d0320b9518df84cb/addons/mrp/models/mrp_bom.py#L201-L202

### Fix:

Just as for the total cost_share on kit products, we rely on the exclusion util and check for each existing product variant if the cost share set up is valid:
https://github.com/odoo/odoo/blob/7e81c528ae350aab4432207f5655dcfadf6ec627/addons/purchase_mrp/models/mrp_bom.py#L20-L23 see 3832793e3ce61aff0c7cf4673de84645a3469b3a

opw-5499773

---

### DIFF
--- a/addons/mrp/models/mrp_bom.py
+++ b/addons/mrp/models/mrp_bom.py
@@ -4,7 +4,7 @@
 from odoo import api, fields, models, _, Command
 from odoo.exceptions import UserError, ValidationError
 from odoo.osv.expression import AND, OR
-from odoo.tools import float_round
+from odoo.tools import float_compare, float_is_zero, float_round
 from odoo.tools.misc import clean_context
 
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


adwid:
Hey @lase-odoo 

Considering our [documentation](https://www.odoo.com/odoo/knowledge/10288) (section "Define the start version") I think you did a small mistake with the targeted version
But I'm sure it was not on purpose

Can you rectify ? 

Have a good day :)

PR #245193, opened 2026-01-22
[FIX] pos_self_order: fix empty order lines check before printing changes

lse-odoo:
Before this commit:
To reproduce (version 17 and >):
 1. Install Restaurant
 2. In PoS restaurant config allow "Self ordering" in Kiosk mode and enable "Preparation printers"
 3. Create a preparation printer with NO categories and set it as a restaurant preparation printer
 4. Open the Kiosk
 5. Make an order with any product and checkout

-> Printer will print a ticket with no product
image

After this commit:
No kitchen printer is printed (expected as no category set on the kitchen printer)


loukwets (approved):

adgu-odoo (commented):
robodoo r+

PR #245194, opened 2026-01-22
[IMP] *: cleanup redundant [role=button]  on button elements

pparidans:
The `<button>`  HTML element has by default the "role", well, "button". No need to specify it again.

Reference:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role

Current behavior before PR:

---


adr-odoo (approved):
@robodoo r+

PR #245197, opened 2026-01-22
[IMP] hr_holidays: made end date of allocation editable

djsh-odoo:
## Purpose
Initially we were unable to set the end date of an allocation once approved for 'accrual type' allocation. We need to refuse it and then we can edit it.

## Issue
- The end date for 'Regular allocation' type is editable but we can even edit it below the date of already taken leave.
- The message for refuse allocation when a leave is already taken was not appropriate.

## Steps To Reproduce
1. Make a new allocation with any type (accrual/ regular) and approve it.
2. Create timeoff for same allocation and approve it.
3. Now try to edit the date_to field.

- If the End Date is set to a date sooner(before) to any leave request approved then it will raise a warning showing approved timeoff requests.
- Also if you try to 'refuse' the allocation it will raise warning showing approved timeoff requests.

task-5472998
---

### DIFF
--- a/addons/hr_holidays/models/hr_leave_allocation.py
+++ b/addons/hr_holidays/models/hr_leave_allocation.py
@@ -829,11 +829,18 @@ class HrLeaveAllocation(models.Model):
 
         self.add_follower(employee_id)
 
-        if 'number_of_days_display' not in values and 'number_of_hours_display' not in values and 'state' not in values:
+        tracked_fields = {'number_of_days_display', 'number_of_hours_display', 'state', 'date_to'}
+        tracked_updates = tracked_fields.intersection(values)
+        if not tracked_updates:
             res = super().write(values)
             if 'accrual_plan_id' in values:
                 self._add_lastcalls()
             return res
+
+        def total_excess(extra_data):
+            excess_days = extra_data.get('excess_days', {})
+            total = sum(leave_date['amount'] for leave_date in excess_days.values())
+            return total - min(extra_data.get('exceeding_duration', 0), 0)
 
         previous_consumed_leaves = self.employee_id._get_consumed_leaves(work_entry_types=self.work_entry_type_id)
         result = super().write(values)
@@ -842,12 +849,12 @@ class HrLeaveAllocation(models.Model):
         if 'accrual_plan_id' in values:
             self._add_lastcalls()
         for allocation in self:
-            current_excess = dict(consumed_leaves[1]).get(allocation.employee_id, {}) \
-                .get(allocation.work_entry_type_id, {}).get('excess_days', {})
-            previous_excess = dict(previous_consumed_leaves[1]).get(allocation.employee_id, {}) \
-                .get(allocation.work_entry_type_id, {}).get('excess_days', {})
-            total_current_excess = sum(leave_date['amount'] for leave_date in current_excess.values() if not leave_date['is_virtual'])
-            total_previous_excess = sum(leave_date['amount'] for leave_date in previous_excess.values() if not leave_date['is_virtual'])
+            current_extra_data = dict(consumed_leaves[1]).get(allocation.employee_id, {}) \
+                .get(allocation.work_entry_type_id, {})
+            previous_extra_data = dict(previous_consumed_leaves[1]).get(allocation.employee_id, {}) \
+                .get(allocation.work_entry_type_id, {})
+            total_current_excess = total_excess(current_extra_data)
+            total_previous_excess = total_excess(previous_extra_data)
 
             if total_current_excess <= total_previous_excess:
                 continue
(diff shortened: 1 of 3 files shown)


addons/hr_holidays/views/hr_leave_allocation_views.xml
```diff
@@ -143,7 +143,7 @@
                                     invisible="not accrual_plan_id"/>
                                 <field name="date_to" nolabel="1"
                                     placeholder="No Limit"
-                                    readonly="state == 'refuse' or (accrual_plan_id and state != 'confirm')"
+                                    readonly="state == 'refuse'"
```
Bertrand2: if it's already almost always editable, wouldn't it make sense to keep it editable even in the refuse state ?

addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -829,12 +866,30 @@ def write(self, vals):
...
             return res
 
+        allocations_to_check = self.env['hr.leave.allocation']
+        if values.get('date_to'):
+            new_date_to = fields.Date.to_date(values['date_to'])
+            allocations_to_check = self.filtered(
+                lambda allocation: not allocation.date_to or new_date_to < allocation.date_to
+            )
+        if tracked_updates == {'date_to'}:
+            result = super().write(values)
```
Bertrand2: You'll be writing twice

addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -829,12 +866,30 @@ def write(self, vals):
...
+            for allocation in allocations_to_check:
+                if allocation._has_conflicting_leaves(values['date_to']):
+                    raise ValidationError(self.env._(
+                        "You cannot set this end date because it would make approved "
+                        "time off requests exceed the available allocation."
+                    ))
+            return result
```
Bertrand2: Not sure all of this is necessary 🤔 The comparison between `previous_consumed_leaves` and `consumed_leaves` should do the trick, no?
At least it should raise an error if you're bringing issues

addons/hr_holidays/tests/test_allocations.py
```diff
@@ -643,3 +643,303 @@ def test_action_generate_group_allocations(self):
...
+            pending_leave = self.env['hr.leave'].with_user(self.user_employee).create({
+                'name': 'Future Pending Leave Request',
+                'work_entry_type_id': work_entry_type.id,
+                'request_date_from': date(2024, 1, 10),
+                'request_date_to': date(2024, 1, 10),
+                'employee_id': self.employee_emp.id,
+            })
+
+            self.assertEqual(pending_leave.state, 'confirm')
+            allocation.write({'date_to': date(2024, 1, 8)})
```
Bertrand2: this shouldn't be allowed.

addons/hr_holidays/tests/test_allocations.py
Bertrand2, line 1: Nitpick, but you could compact a bit more the tests together, it's a bit too much, for example:

- create an validated regular allocation
- check date_to is editable
- add a leave
- check changing date_to after the leave is valid
- check changing date_to before the leave is invalid
- add a second covering regular allocation
- check changing the initial allocation before the leave is valid

With accruals:

- create an validated accrual allocation
- check date_to is editable
- add a leave
- check changing date_to after the leave is valid
- check changing date_to before the leave is invalid

This should be sufficient IMO :)

addons/hr_holidays/models/hr_leave_allocation.py
kais-odoo: Indeed, I simplified this. There is no date_to-specific write branch anymore, so we now keep a single super().write() call.

addons/hr_holidays/models/hr_leave_allocation.py
kais-odoo: Yes, I changed it in that direction. date_to is now included in the existing consumed leaves comparison instead of using a separate conflicting leaves helper.

I only adapted the comparison to include virtual excess when date_to changes, so pending/future requests are also blocked.

addons/hr_holidays/tests/test_allocations.py
kais-odoo: Agreed, I changed the behavior. Reducing the allocation end date before a pending request now raises a ValidationError.

addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -829,25 +829,40 @@ def write(self, vals):
...
-        if 'number_of_days_display' not in values and 'number_of_hours_display' not in values and 'state' not in values:
+        tracked_fields = {'number_of_days_display', 'number_of_hours_display', 'state', 'date_to'}
+        tracked_updates = tracked_fields.intersection(values)
+        if not tracked_updates:
             res = super().write(values)
             if 'accrual_plan_id' in values:
                 self._add_lastcalls()
             return res
 
+        check_virtual_excess = 'date_to' in values
```
Bertrand2: always check virtual excess

Bertrand2:
robodoo r+

PR #245198, opened 2026-01-22
[IMP] sale: SO signature integrity

lkerroum:
A signed order can be modified afterward while retaining the signature on the newly generated PDF. After this commit, we will only print the signature on the document generated by the signature itself and not on any generated PDF afterwards. 

We also take this opportunity to:
- Eliminate the "PDF Quote" option, leaving "Quotation/Order" as the option to print the raw quote, and "Print"(button) to print the quote with the additional hearder/footer.
- Stop storing the signature related fields and remove the Customer Signature tab

Enterprise PR: https://github.com/odoo/enterprise/pull/107529
 
task-5421716

### DIFF
--- a/addons/sale/report/ir_actions_report_templates.xml
+++ b/addons/sale/report/ir_actions_report_templates.xml
@@ -336,7 +336,10 @@
             </div>
             <div class="oe_structure"></div>
 
-            <div t-if="not doc.signature" class="oe_structure"></div>
+            <div
+                t-if="not doc.signature or not doc.env.context.get('sale_include_signature')"
+                class="oe_structure"
+            />
             <div t-else="" class="mt-4 ml64 mr4" name="signature">
                 <div class="offset-8">
                     <strong>Signature</strong>
--- a/addons/sale_pdf_quote_builder/report/ir_actions_report.xml
+++ b/addons/sale_pdf_quote_builder/report/ir_actions_report.xml
@@ -13,7 +13,7 @@
     </record>
 
     <record id="sale.action_report_saleorder" model="ir.actions.report">
-        <field name="name">PDF Quote</field>
+        <field name="binding_model_id" eval="False"/>
     </record>
 
 </odoo>
--- a/addons/sale/controllers/portal.py
+++ b/addons/sale/controllers/portal.py
@@ -376,6 +376,7 @@ class CustomerPortal(payment_portal.PaymentPortal):
             request
             .env["ir.actions.report"]
             .sudo()
+            .with_context(sale_include_signature=True)
             ._render_qweb_pdf("sale.action_report_saleorder", [order_sudo.id])[0]
         )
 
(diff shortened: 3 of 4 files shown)


addons/sale/report/ir_actions_report_templates.xml
```diff
@@ -329,7 +329,9 @@
             </div>
             <div class="oe_structure"></div>
 
-            <div t-if="not doc.signature" class="oe_structure"></div>
+            <div t-if="not doc.signature or not doc.env.context.get('sale_include_signature')"
+                class="oe_structure"
+            />
```
ltinel: Nit:
```suggestion
            <div
                t-if="not doc.signature or not doc.env.context.get('sale_include_signature')"
                class="oe_structure"
            />
```

addons/sale_pdf_quote_builder/report/ir_actions_report.xml
```diff
@@ -13,7 +13,7 @@
     </record>
 
     <record id="sale.action_report_saleorder" model="ir.actions.report">
-        <field name="name">PDF Quote</field>
+        <field name="binding_model_id" eval="False"/>
```
ltinel, line 16: I guess this has the effect of removing the "PDF Quote" option from the "Print" menu? :)

addons/sale/models/sale_order.py
```diff
@@ -158,10 +158,15 @@ def _rec_names_search(self):
     )
 
     signature = fields.Image(
-        string="Signature", copy=False, attachment=True, max_width=1024, max_height=1024
+        string="Signature",
+        copy=False,
+        attachment=True,
+        max_width=1024,
+        max_height=1024,
+        store=False
```
ltinel: I don't think you can simply mark these fields as non-stored. By doing this, you'll lose the info whenever you reload the page (or any other action that reloads the SO). And as a result, you'll never be able to pay for the order in the portal, as you'll always be asked to sign first (since signing refreshes the page) 😅

Could you tell me what you were trying to achieve with this? 🙂 There might be alternatives. If the purpose is to make the signature "single-use", then maybe the field should be removed from the model (which implies some changes in the portal and elsewhere as well).

ltinel (changes requested):
Thanks for your work :)

addons/sale_pdf_quote_builder/report/ir_actions_report.xml
lkerroum: Yes, else it's automatically created from the action defined in `sale`. This was the intended change (see Dev notes with the updated spec decisions after discussion)

addons/sale/models/sale_order.py
lkerroum: Thank you for taking to time to review my work 😃 My goal with this change was to fulfill this request https://www.odoo.com/mail/message/999233044 . Maybe it would make sense to only delete it from the view?

ltinel: Yes, for now, I'd only delete it from the view :) I had a quick chat with Jean and it's fine with him

ltinel (approved):
LGTM, thanks :)

Feyensv (approved):

Feyensv:
robodoo r+

PR #245202, opened 2026-01-22
[IMP] web: replace a[href="#"]  by actual button element

pparidans:
This commit introduces the `.btn-link-inline` class to reproduce the inline styling of a `<a href="...">` tag as a variant of Bootstrap's `.btn.btn-link`.

This allows to properly uses `<button>` element for interactive DOM elements instead of relying on a "hacky" and non-semantic `<a href="#">` (tl;dr: `<a>` HTML tag are for navigation, `<button>` are for interactivity).

task-4380519


res-odoo (approved):
@robodoo r+

PR #245205, opened 2026-01-22
[FIX] *: replace "Deco Addict" with "Acme Corporation"

reth-odoo:
A company that happens to be named "Deco Addict" has complained some of our users thought they had business with them due to test and demo data containing that name.

It will now be named Acme Corporation.

task-5865502


addons/account/views/bill_preview_template.xml
```diff
@@ -8,7 +8,7 @@
...
...F09/fb8bi5o8JCdCXhla11p9xXXdqHMff7+jocFP/vFE3Jx3H2SNJkpMAXFOr1QzXizDdrcUShoaGZDab9YQQNrXK46lTpwpm9gBgw4YNtEUIGMaY3Lbu+tVaG6UUu66bSS+KHP2dBMCWSqXGbSENFdK8PfzlAj9BEPCiRYv2BnBKkiRnWmvPsdaeLYRYSERnEtHZRPT/oiiqADjb9303CIIhIvqTEOLoIAhsW1sb+76vzjrrLDcIApvNZvf1PG8fZh5OpchmZn5USnlkEAR2YGDAphzrpBb5WzKZzAwiqm3FQGUAyGQyT1lrH2bmfBAEdsOGDTKdQzXcwC0BpJRKfN9Xb3zjG9VYjQW8rhIgjdBZpdR5AP60ZMmSb2wDLH8F8JDWejaA2wH4UsofF4vFNYVC4QcNRl20aNHeQohvSClHDLh0IzuFEHd0dHR8rqur68rG788///w9jDHf8zwvbHgNoxm/MUcQBNVisfhlIvp2sVi8p7u7e9WoOdpyudwXmbk9CIL1AGKu04ygXsE8Zi++Uq8n9xcKBbto0aJ9iegMa+1nmZnOOuss94knntD5fH5k7B133CGCIHi0o6PjZiHEV6+66qoDzjjjjBuLxeL5RPStYrH4KSJ6hIhahRDvMcZcX6vVcsw8fVSM4a729vbPCyG+2tHRcQozPwwgp5SaqbX+abVazQJ44+g1uq6bjaLIbeQXuru7v9Pe3v5WKeW1xWLxXiL6K4CpSqn3GWPuj+O4ll52pTs6Oq5RSi3t6OjIMrMD4OHu7m4/zXI27w1sMKYQYiqAr+62224/ICK+4oor4oZx1fhMnz7dAGBjTBczr1y/fv0UAOju7r5ECHEQgPsBZJl5ozHmU11dXV8E8F/W2rsa6tj3fdHT0/NNrfUBzPwLABkAQ8x8Znd39+eZ+TvW2ttHc38URf8hpXwIAAYGBiwA6unp6SSiw5j592m3kCFr7VlLliw5ftmyZQOVSsUCQFdX139orf8fEblENMzMd6JJ/3wpso3A0v95/KuZAy++8GpcJI9oLGzihg0b5NVXX528wpcvGmHd0b9rWPbTp083aVBJAXhRgcVLjS+Xy7JSqYxO+mDBggXOVVddpWkrxSOj5xgYGLBbu920XC7LO+64Q4z+nibbNqlJTWpSk5rUpLFA/x/uiqsFaH5j9gAAAABJRU5ErkJggg=="/>
```
reth-odoo, line 11: Logo was embedded in base64 here, so replaced with the new one

------------------------------------

Looked for other copies that may be hidden in base64 using these:

`grep -rhoP 'image/png;base64,\K[^\'"]+(?=[\'"])' . > res.txt`

```
i=1

while IFS= read -r line; do
  echo "$line" | base64 --decode > "${i}.png"
  i=$((i + 1))
done < res.txt
```

-----------------------------

Couldn't find any, hopefully that's all of them.

addons/account/wizard/account_tour_upload_bill.py
```diff
@@ -50,7 +50,9 @@ def _selection_values(self):
         journal_alias = self.env['account.journal'] \
             .search([('type', '=', 'purchase'), ('company_id', '=', self.env.company.id)], limit=1)
 
-        values = [('sample', _('Try a sample vendor bill')), ('upload', _('Upload your own bill'))]
+        # We don't want people putting demo data in their database if they didn't launch it in demo mode
+        values = [('sample', _('Try a sample vendor bill'))] if self.env.ref('base.res_partner_2', raise_if_not_found=False) else []
```
mart-e, line 54: I think the proper way to check for demo is `bool(self.env['ir.module.module'].search_count([('demo', '=', True)]))`

reth-odoo: :o ok I didn't know there was a demo module

Probably not worth changing at this point but noted

PR #245208, opened 2026-01-22
[FIX] point_of_sale: delivery address in POS inhibits real time COGS

martiyen:
When a delivery address is selected as the customer in the POS UI and an invoice is generated from the POS order, it results in two different `partner_id` being set on the `pos.order` and the `account.move`. It is expected, as a delivery address cannot logically be used as an invoicing address.

Anglo-saxon accounting registers the COGS with additional account.move.line on the invoice and, when using real time valuation, the balance of these lines is calculated from values found on the corresponding `stock.move`.

Currently, the `stock.move` cannot be found if the `partner_id` on the `pos.order` and the `account.move` are different. The system then falls back on the `standard_price` of the `product_id`, which is incorrect. In version 19+, the `product_id` used to calculate the unit price is fetched from the `stock.move` but, because it's an empty recordset, the balance of the COGS is 0 and the lines are not created.

An easy fix could be to remove this check of equivalence between the two `partner_id`. The invoice is created directly from the POS order anyway. I am not sure why it is necessary. Am I missing something?

Steps to reproduce:
1. Create a product with fifo and automated valuation. Make sure that the product type is storable.
2. Validate one purchase and its delivery for the product (qty 5, unit price 5).
3. Validate another purchase and its delivery for the product (qty 5, unit price 10).
4. Configure the point of sale to allow invoicing.
5. Open a pos session then sell 1 item of the product. Select a delivery adress as the customer and generate an invoice.
6. Since the product is fifo, this means that the cost of goods sold is 1 * 5.
7. The correct expense account line should have balance of 5.0.
8. The correct output account line should have balance of -5.0.
9. The COGS lines instead have a balance of 8.13 and -8.13 which corresponds to the standard price of the product at the time of the invoice.

Related ticket: opw-5467877

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -287,8 +287,7 @@ class PosOrder(models.Model):
         return invoice_lines
 
     def _get_pos_anglo_saxon_price_unit(self, product, partner_id, quantity):
-        moves = self.filtered(lambda o: o.partner_id.id == partner_id)\
-            .mapped('picking_ids.move_ids')\
+        moves = self.mapped('picking_ids.move_ids')\
             .filtered(lambda m: m.is_valued and m.product_id.valuation == 'real_time' and m.product_id.id == product.id)\
             .sorted(lambda x: x.date)
         return moves._get_price_unit()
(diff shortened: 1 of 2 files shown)


martiyen:
Hello @loukwets 

Did you have a chance to check on the PR? Thanks

loukwets (approved):

martiyen:
Hello @stva-odoo 

Did you have a chance to check on this PR? Have a nice day!

addons/point_of_sale/models/pos_order.py
```diff
@@ -279,9 +279,8 @@ def _prepare_invoice_lines(self):
 
         return invoice_lines
 
-    def _get_pos_anglo_saxon_price_unit(self, product, partner_id, quantity):
-        moves = self.filtered(lambda o: o.partner_id.id == partner_id)\
-            .mapped('picking_ids.move_ids')\
+    def _get_pos_anglo_saxon_price_unit(self, product, quantity):
```
stva-odoo: I suggest avoiding changing the method signature in stable.
Keep the partner_id even if it is no longer required in the implementation
And you could modify the method signature in master.

PR #245209, opened 2026-01-22
[FIX] account: Fixed large spacing in invoicing dashboard

hsal-odoo:
As a part of task-5258726 the icons of no-content help in bills list view were enlarged.
The icons used there were also used in the module dashboard, so enlarging them messed the dashboard styling.

This commit fixes this issue by using a prop `largeIcons` on the `BillGuide` componenet to make it with large icons when needed only (in the bills list view no content help).

task-5801970

---


hupo-odoo (approved):
@robodoo r+
Thanks for the fix!

PR #117806, opened 2023-04-05
[IMP] core: recurse in SQL

HydrionBurst:
[IMP] core: refactor dependency
refactor upstream_dependency and downstream_dependency since they do a recursion
across python and SQL.
These methods are not performance bottlenecks, but it would be simpler to only
use RECURSIVE CTE to do everything in SQL when possible.

Behavior changes
1 for downstream_dependencies and upstream_dependencies
Before this commit:

when call `dd = module_a.downstream_dependencies(known_deps=module_a)`
the module_a will be included in the returned dd
when call `dd = module_a.downstream_dependencies()`
the module_a won't be included in the returned dd
It is too tricky and useless

After this commit:
know_deps is removed and when call `dd = module_a.downstream_dependencies()`
the module_a won't be included in the returned dd

2 for upstream_dependencies
Before this commit:
The result of upstream_dependencies won't include the last layer of depends
it should be treated as a bug since it is against its definition
but it is a safe bug, since the last layer of depends is always 'base' which
should always be installed

After this commit:
the module 'base' is actually always included in the result of
upstream_dependencies

[IMP] core: check recursion by sql
use RECURSIVE CTE to check recursion
avoid potential dead loop for check_recursion in test_105_res_partner_recursion

PR #245210, opened 2026-01-22
[IMP] base: rewrite IrModule.upstream_dependencies to use an RCTE

xmo-odoo:
Looping through every module of community + enterprise (1367 modules) and getting their entire set of dependencies, the new version runs in about a third the time on my machine (1.5s to 4.2s), yielding a result that's almost identical: the new version includes `base`, the old one did not (not sure why, or how).

Now supports calling with `exclude_states=()`, which did not work previously and led to wonky workarounds e.g.

    self.upstream_dependencies(exclude_states=('',))

or

    module.upstream_dependencies(exclude_states=('uninstallable'))

(which the keen-eyed will notice does not pass a tuple so excludes the individual non-existent states `u`, `n`, `i`, `s`, `t`, `a`, `l`, `b`, and `e`).

Also drops support for `known_deps`: it's not used anywhere and doesn't really seem useful.

### DIFF
--- a/addons/website/controllers/backend.py
+++ b/addons/website/controllers/backend.py
@@ -43,7 +43,7 @@ class WebsiteBackend(http.Controller):
         show the progress between installed and yet to install features.
         """
         features_not_installed = request.env['website.configurator.feature']\
-            .browse(selected_features).module_id.upstream_dependencies(exclude_states=('',))\
+            .browse(selected_features).module_id.upstream_dependencies(exclude_states=())\
             .filtered(lambda feature: feature.state != 'installed')
 
         # On the 1st run, the total tallies the targeted, not yet installed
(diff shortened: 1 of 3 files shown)


HydrionBurst (approved):
LGTM

I had a canceled PR did the same thing
https://github.com/odoo/odoo/pull/117806/changes/2e64c2b8ec2fa92374cc8e74294594ff7c7368bf
This one seems simpler and better.

As I mentioned in its commit message

> The result of upstream_dependencies won't include the last layer of depends
it should be treated as a bug since it is against its definition
but it is a safe bug, since the last layer of depends is always 'base' which
should always be installed


`base` was not returned because previously we 
```sql
SELECT DISTINCT m.id
FROM ir_module_module_dependency d
JOIN ir_module_module m ON (d.module_id=m.id)
...
```
Since `base` module doesn't depends on any module,
there is no ir_module_module_dependency row for module_id = [module_id_for_base]
As a result, after `JOIN` the returned `m.id` will never be the id of base

PR #245214, opened 2026-01-22
[FIX] point_of_sale: close customer display popup

macocianradu:
When opening the customer display in POS, the popup would open a new window with the display and show a notification that the window is opened successfully, but the popup would remain active.

This PR is to make the popup close when the customer display is opened.

Task-[5867558](https://www.odoo.com/odoo/project/1737/tasks/5867558)

---


davidmonnom (approved):
@robodoo r+

PR #245216, opened 2026-01-22
[IMP] sale, _*: add Services and Materials upselling view

nipl-odoo:
_*= sale_project, sale_timesheet

Added a new *Services & Materials* view to allow users to create additional delivered
flat fees service or consumable product entries directly in a sale order.

This feature allows:
* Creating analytic lines from a sale order for extra services and materials.
* Automatically creating or updating related sale order lines from those
  analytic lines.
* Handling SO line creation based on product expense policy:
   * `cost` → create a new SO line.
   * `sales_price` → update an existing line if available.
* Updating delivered quantities when analytic lines are changed.
* Reverting effect/deleting SO line when analytic lines are deleted.

All analytic lines created from this flow are assigned to the analytic plan
`sale.analytic_plan_sale_orders`, which ensures an analytic account is created
and set on the sale order.

Products with reinvoicing policy `no` will only be eligible for manual delivery.

If a product has reinvoicing policy anything other than `no` it will be eligible
for upselling from our new view, and delivered quantity will be aggregated
from AALs.

If a project is selected on the sale order, the analytic account is deduced from
that project instead.

Added a stat button on invoices to easily view analytic lines linked to that
invoice.

Technical changes:

* Field `timesheet_invoice_id` is renamed to `reinvoice_move_id` as it's not
  timesheet specific anymore.
* Made `product_uom_id` field on AAL compute to make sure we get expected
  `uom_id` on AAL.
* Changed field `order_id` from `related` to `compute` since we need it to work
  as normal m2o in sale.
* Centralized analytic line reinvoicing logic in the sale module.
* Improved `_get_analytic_lines_to_link` to support timesheet-based
  subscriptions with date filtering.
* `qty_delivered` will now be sum of upsell AALs and stock moves(for
   consu products) and recorded timesheet hours(for timesheets).

task-5490517

See Also:
- https://github.com/odoo/enterprise/pull/105158
- https://github.com/odoo/upgrade/pull/9366

---

### DIFF
--- a/addons/analytic/models/analytic_line.py
+++ b/addons/analytic/models/analytic_line.py
@@ -160,7 +160,6 @@ class AccountAnalyticLine(models.Model):
 
     name = fields.Char(
         'Description',
-        required=True,
     )
     date = fields.Date(
         'Date',
@@ -180,6 +179,9 @@ class AccountAnalyticLine(models.Model):
     product_uom_id = fields.Many2one(
         'uom.uom',
         string='Unit',
+        compute='_compute_product_uom_id',
+        store=True,
+        readonly=False,
     )
     partner_id = fields.Many2one(
         'res.partner',
@@ -225,6 +227,10 @@ class AccountAnalyticLine(models.Model):
         default=lambda self: self.env['decimal.precision'].precision_get("Percentage Analytic"),
     )
 
+    def write(self, vals):
+        self._check_can_write(vals)
+        return super().write(vals)
+
     def _compute_analytic_distribution(self):
         for line in self:
             line.analytic_distribution = {line._get_distribution_key(): 100}
@@ -257,9 +263,16 @@ class AccountAnalyticLine(models.Model):
                 'message': self.env._("%s analytic lines created", len(to_create_vals)),
             })
 
+    def _compute_product_uom_id(self):
+        return
+
     def _split_amount_fname(self):
         return 'amount'
 
     def _search_fiscal_date(self, operator, value):
         fiscalyear_date_range = self.env.company.compute_fiscalyear_dates(fields.Date.today())
         return [('date', '>=', fiscalyear_date_range['date_from'] - relativedelta(years=1))]
+
+    # Hook to be shared between non related modules
+    def _check_can_write(self, vals):
+        return True
--- a/addons/sale/data/analytic_plan_data.xml
+++ b/addons/sale/data/analytic_plan_data.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record id="analytic_plan_sale_orders" model="account.analytic.plan">
+        <field name="name">Sales Orders</field>
+        <field name="default_applicability">unavailable</field>
+    </record>
+</odoo>
--- a/addons/sale/models/account_analytic_line.py
+++ b/addons/sale/models/account_analytic_line.py
@@ -1,0 +1,231 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.fields import Domain
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    reinvoice_move_id = fields.Many2one(
+        string="Invoice",
+        comodel_name='account.move',
+        readonly=True,
+        copy=False,
+        help="Invoice created from related SO line",
+        index='btree_not_null',
+    )
+    so_line = fields.Many2one(
+        string='Sales Order Item',
+        comodel_name='sale.order.line',
+        compute='_compute_so_line',
+        store=True,
+        readonly=False,
+        index='btree_not_null',
+        domain=lambda self: self._domain_so_line(),
+    )
+    order_id = fields.Many2one(
+        string="Customer Order",
+        comodel_name='sale.order',
+        compute='_compute_order_id',
+        store=True,
+        readonly=False,
+        index=True,
+    )
+
+    def _compute_so_line(self):
+        return
+
+    def _domain_so_line(self):
+        return Domain('qty_delivered_method', '=', 'analytic')
+
+    def _compute_order_id(self):
+        return
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        lines = super().create(vals_list)
+        if self.env.context.get('from_services_and_material'):
+            lines._sync_so_accounts_and_partners()
+            lines._sync_so_lines()
+        return lines
+
+    def write(self, vals):
+        if self and self.env.context.get('from_services_and_material'):
+            order_changed_aals = self.env['account.analytic.line']
+            product_changed_aals = self.env['account.analytic.line']
+
+            if vals.get('order_id'):
+                order_changed_aals = self.filtered(
+                    lambda aal: aal.order_id.id != vals['order_id']
+                )
+
+            if vals.get('product_id'):
+                product_changed_aals = self.filtered(
+                    lambda aal: aal.product_id.id != vals['product_id']
+                )
+
+            product_or_order_changed_aals = order_changed_aals | product_changed_aals
+
+            res = super().write(vals)
+
+            # if order changed then we need to reassign accounts and partners
+            order_changed_aals._sync_so_accounts_and_partners()
+
+            product_or_order_changed_aals._unsync_so_lines()
+            product_or_order_changed_aals._sync_so_lines()
+        else:
+            res = super().write(vals)
+        return res
+
+    def _check_can_write(self, vals):
+        if self.filtered(
+            lambda aal: aal.reinvoice_move_id and aal.reinvoice_move_id.state != 'cancel'
+        ):
+            if any(field_name in vals for field_name in self._restricted_fields_when_invoiced()):
+                raise UserError(self._get_invoiced_line_write_error())
+
+        if 'unit_amount' in vals and vals['unit_amount'] < 0 and self.so_line:
+            raise UserError(self.env._("You cannot set a negative quantity on services."))
+
+        super()._check_can_write(vals)
+
+    def _restricted_fields_when_invoiced(self):
+        return ['unit_amount', 'order_id', 'product_id', 'so_line', 'date', 'partner_id']
+
+    def _get_invoiced_line_write_error(self):
+        return self.env._("You cannot modify already invoiced services.")
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_so_lines_except_invoiced(self):
+        """Cleanup related sale order lines when analytic lines are deleted.
+
+        Analytic lines linked to a posted invoice cannot be removed.
+
+        For other lines, the related sale order line is deleted only if its
+        delivery method is manual. Lines originating from timesheets or
+        expenses are left unchanged.
+        """
+        if any(
+            line.reinvoice_move_id and line.reinvoice_move_id.state == 'posted' for line in self
+        ):
+            raise UserError(self._get_invoiced_line_delete_error())
+        self._unsync_so_lines()
+
+    def _get_invoiced_line_delete_error(self):
+        return self.env._("You cannot remove already invoiced services.")
+
+    def _sync_so_lines(self):
+        """Ensure that a corresponding sale order line exists and is synchronized
+        with the current analytic line.
+
+        Depending on the product's expense policy:
+
+        - For 'cost' policy:
+          A new sale order line with product's cost is always created with delivered quantity
+          equal to the analytic line's unit amount.
+
+        - For 'sales_price' policy:
+          The method first attempts to find an existing sale order line
+          matching the product. If found, its delivered quantity is recomputed.
+          Otherwise, a new sale order line is created with product's sales price.
+
+        The analytic line is then linked to the resulting sale order line.
+        """
+        for line in self:
+            if not line.order_id or not line.product_id:
+                continue
+
+            so_line = self.env['sale.order.line']
+            if line.product_id.expense_policy == 'sales_price':
+                so_line = line._get_existing_so_line()
+            line.so_line = so_line or line._create_so_line()
+
+    def _get_existing_so_line(self):
+        """Retrieve the first matching sale order line from the related order, if any."""
+        self.ensure_one()
+        return self.order_id.order_line.filtered(
+            lambda line: line.product_id == self.product_id
+            and line.product_uom_id == self.product_uom_id
+            and line._is_reinvoicing_line()
+        )[:1]
+
+    def _create_so_line(self):
+        """Create a new sale order line corresponding to this analytic line.
+
+        The created line is initialized with delivered quantity based on the
+        analytic line amount, unit price derived from the product's expense
+        policy, and an optional custom description.
+
+        :rtype: sale.order.line
+        :return: The newly created sale order line record.
+        """
+        self.ensure_one()
+        values = {
+            'order_id': self.order_id.id,
+            'product_id': self.product_id.id,
+            'product_uom_id': self.product_uom_id.id,
+            'product_uom_qty': 0,
+        }
+
+        if self.product_id.expense_policy == 'cost':
+            product = self.product_id.with_company(self.order_id.company_id)
+            values['price_unit'] = product.currency_id._convert(
+                product.standard_price, self.order_id.currency_id, round=False
+            )
+
+        return self.env['sale.order.line'].create(values)
+
+    def _unsync_so_lines(self):
+        """Revert synchronization of delivered quantities on related sale order lines.
+
+        If the linked sale order line has an expense policy of `cost` and no
+        ordered quantity, it indicates that the line was created solely for
+        reinvoicing purposes and can be safely removed.
+        If not, the analytic line update will automatically trigger the recomputation
+        of the `qty_delivered` field on the linked so lines .
+        """
+        for line in self.filtered(lambda line: line.so_line):
+            if not line.so_line._is_reinvoicing_line():
+                continue
+            if (
+                line.product_id.expense_policy == 'cost'
+                and not line.so_line.product_uom_qty
+                and line.unit_amount == line.so_line.qty_delivered
+            ):
+                line.so_line.unlink()
+
+    def _sync_so_accounts_and_partners(self):
+        """Synchronize analytic account fields and partner with the related sale order.
+
+        This method ensures that each analytic line is linked to the correct
+        analytic account derived from its associated sale order.
+
+        The analytic account is retrieved (or created if necessary) using
+        the analytic plan dedicated to sale orders. If that plan no longer
+        exists, the project plan is used as a fallback.
+
+        Since analytic plans are stored as dynamic columns on the analytic
+        line model, the method determines the effective column name based
+        on the plan actually used and assigns the corresponding analytic
+        account to that field.
+        """
+        if not self:
+            return
+
+        plan_id = self.env.ref('sale.analytic_plan_sale_orders', raise_if_not_found=False)
+        if not plan_id:
+            plan_id, _other_plans = self.env['account.analytic.plan']._get_all_plans()
+        column_name = plan_id._column_name()
+
+        for line in self:
+            line.partner_id = line.order_id.partner_id
+            analytic_account = line.order_id._get_or_create_analytic_account(plan_id)
+            used_plan_id = analytic_account.plan_id
+
+            effective_column_name = column_name
+            if used_plan_id != plan_id:
+                effective_column_name = used_plan_id._column_name()
+
+            line[effective_column_name] = analytic_account.id
--- a/addons/sale/models/account_move.py
+++ b/addons/sale/models/account_move.py
@@ -1,5 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 from odoo import _, api, fields, models
+from odoo.fields import Domain
 from odoo.tools import groupby, OrderedSet
 
 
@@ -22,6 +23,7 @@ class AccountMove(models.Model):
         "Sale Warning",
         help="Internal warning for the partner or the products as set by the user.",
         compute="_compute_sale_warning_text")
+    service_line_count = fields.Integer(compute="_compute_service_line_count")
 
     def unlink(self):
         downpayment_lines = self.mapped('line_ids.sale_line_ids').filtered(lambda line: line.is_downpayment and line.invoice_lines <= self.mapped('line_ids'))
@@ -47,6 +49,16 @@ class AccountMove(models.Model):
     def _compute_origin_so_count(self):
         for move in self:
             move.sale_order_count = len(move.line_ids.sale_line_ids.order_id)
+
+    def _domain_services_analytic_line(self):
+        return Domain('reinvoice_move_id', 'in', self.ids)
+
+    @api.depends('line_ids.sale_line_ids')
+    def _compute_service_line_count(self):
+        for move in self:
+            move.service_line_count = self.env['account.analytic.line'].search_count(
+                move._domain_services_analytic_line(),
+            )
 
     @api.depends('partner_id.name', 'partner_id.sale_warn_msg', 'invoice_line_ids.product_id.sale_line_warn_msg', 'invoice_line_ids.product_id.display_name')
     def _compute_sale_warning_text(self):
@@ -161,6 +173,12 @@ class AccountMove(models.Model):
             result = {'type': 'ir.actions.act_window_close'}
         return result
 
+    def action_view_services_analytic_lines(self):
+        self.ensure_one()
+        result = self.env['ir.actions.act_window']._for_xml_id('sale.action_service_material')
+        result['domain'] = self._domain_services_analytic_line()
+        return result
+
     def _is_downpayment(self):
         # OVERRIDE
         self.ensure_one()
@@ -204,3 +222,39 @@ class AccountMove(models.Model):
             )
             exclude_amount += order_amount_company
         return exclude_amount
+
+    def _link_analytic_lines_to_invoice(self):
+        """Link analytic lines originating from a sale order to the corresponding invoice.
+
+        When an invoice is generated from a sale order, this method associates all
+        related analytic lines with the invoice by setting their `reinvoice_move_id` field.
+
+        This allows tracking which analytic entries have already been invoiced.
+        """
+        invoices = self.filtered(lambda i: i.move_type == 'out_invoice' and i.state == 'draft')
+        for invoice in invoices:
+            invoice._get_reinvoiced_analytic_lines_to_link().reinvoice_move_id = invoice
+
+    def _get_reinvoiced_analytic_lines_to_link(self):
+        # This method is kept separate because some modules require more granular control
+        # over how analytic lines are retrieved. Instead of aggregating all sale order lines
+        # from invoice lines at once, certain implementations need to search analytic lines
+        # per invoice line individually. See the override in `sale_subscription_timesheet`.
+        so_lines = self.invoice_line_ids.sale_line_ids.filtered(lambda line: line._is_line_reinvoicable())
+
+        return self.env['account.analytic.line'].sudo().search(
+            self._analytic_line_domain_get_invoiced_lines(so_lines)
+        )
+
+    @api.model
+    def _analytic_line_domain_get_invoiced_lines(self, so_lines):
+        return (
+            Domain('so_line', 'in', so_lines.ids)
+            & (
+                Domain('reinvoice_move_id', '=', False)
+                | Domain('reinvoice_move_id', 'any', (
+                    Domain('payment_state', '=', 'reversed')
+                    | (Domain('state', '=', 'cancel') & Domain('payment_state', '!=', 'invoicing_legacy')
+                )))
+            )
+        )
--- a/addons/sale/models/sale_order.py
+++ b/addons/sale/models/sale_order.py
@@ -361,6 +361,8 @@ class SaleOrder(models.Model):
     show_update_pricelist = fields.Boolean(
         string="Has Pricelist Changed", store=False)  # True if the pricelist was changed
 
+    analytic_account_id = fields.Many2one(string="Analytic Account", comodel_name='account.analytic.account')
+
     _date_order_id_idx = models.Index("(date_order desc, id desc)")
 
     #=== COMPUTE METHODS ===#
@@ -1804,6 +1806,8 @@ class SaleOrder(models.Model):
                 moves_to_switch.action_switch_move_type()
                 self.invoice_ids._set_reversed_entry(moves_to_switch)
 
+        moves._link_analytic_lines_to_invoice()
+
         for move in moves:
             move.message_post_with_source(
                 'mail.message_origin_link',
@@ -2098,7 +2102,7 @@ class SaleOrder(models.Model):
                 user_id=order.user_id.id or order.partner_id.user_id.id,
                 note=_("Upsell %(order)s for customer %(customer)s", order=order_ref, customer=customer_ref))
 
-    def _prepare_analytic_account_data(self, prefix=None):
+    def _prepare_analytic_account_data(self, prefix=None, plan_id=None):
         """ Prepare SO analytic account creation values.
 
         :return: `account.analytic.account` creation values
@@ -2108,12 +2112,13 @@ class SaleOrder(models.Model):
         name = self.name
         if prefix:
             name = prefix + ": " + self.name
-        project_plan, _other_plans = self.env['account.analytic.plan']._get_all_plans()
+        if not plan_id:
+            plan_id, _other_plans = self.env['account.analytic.plan']._get_all_plans()
         return {
             'name': name,
             'code': self.client_order_ref,
             'company_id': self.company_id.id,
-            'plan_id': project_plan.id,
+            'plan_id': plan_id.id,
             'partner_id': self.partner_id.id,
         }
 
@@ -2268,6 +2273,15 @@ class SaleOrder(models.Model):
         for order in self:
             for line in order.order_line:
                 line.qty_delivered = line.product_uom_qty
+
+    def _get_or_create_analytic_account(self, plan_id):
+        self.ensure_one()
+
+        if not self.analytic_account_id:
+            self.analytic_account_id = self.env['account.analytic.account'].create(
+                self._prepare_analytic_account_data(plan_id=plan_id)
+            )
+        return self.analytic_account_id
 
     # === CATALOG === #
 
--- a/addons/sale/models/sale_order_line.py
+++ b/addons/sale/models/sale_order_line.py
@@ -904,7 +904,7 @@ class SaleOrderLine(models.Model):
     def _compute_customer_lead(self):
         self.customer_lead = 0.0
 
-    @api.depends('is_expense')
+    @api.depends('is_expense', 'product_id')
     def _compute_qty_delivered_method(self):
         """ Sale module compute delivered qty for product [('type', 'in', ['consu']), ('service_type', '=', 'manual')]
                 - consu + expense_policy : analytic (sum of analytic unit_amount)
@@ -914,17 +914,34 @@ class SaleOrderLine(models.Model):
             This is true when only sale is installed: sale_stock redifine the behavior for 'consu' type,
             and sale_timesheet implements the behavior of 'service' + service_type=timesheet.
         """
+        self.qty_delivered_method = 'manual'
         for line in self:
             if line.is_expense:
                 line.qty_delivered_method = 'analytic'
-            else:  # service and consu
-                line.qty_delivered_method = 'manual'
+            elif line.product_id.type == 'service':
+                line.qty_delivered_method = line.product_id.service_type
+            elif line.product_id.type == 'consu':
+                line.qty_delivered_method = line._get_consu_qty_delivered_method()
+
+            # If product has a reinvoice policy and no specific delivered method, its delivered
+            # quantities should still be computed through analytic lines.
+            # For other delivery methods, they are expected to add their own quantities to the
+            # quantities already provided by the `_prepare_qty_delivered` method, including
+            # analytic lines quantities for reinvoiceable products.
+            if line.qty_delivered_method == 'manual' and line.product_id.expense_policy != 'no':
+                line.qty_delivered_method = 'analytic'
+
+    def _get_consu_qty_delivered_method(self):
+        return 'manual'
 
     @api.depends(
         'qty_delivered_method',
+        'analytic_line_ids',
         'analytic_line_ids.so_line',
         'analytic_line_ids.unit_amount',
-        'analytic_line_ids.product_uom_id')
+        'analytic_line_ids.product_uom_id',
+        'analytic_line_ids.product_id',
+    )
     def _compute_qty_delivered(self):
         """ This method compute the delivered quantity of the SO lines: it covers the case provide by sale module, aka
             expense/vendor bills (sum of unit_amount of AAL), and manual case.
@@ -952,11 +969,18 @@ class SaleOrderLine(models.Model):
     def _prepare_qty_delivered(self):
         # compute for analytic lines
         delivered_qties = defaultdict(float)
-        lines_by_analytic = self.filtered(lambda sol: sol.qty_delivered_method == 'analytic')
-        mapping = lines_by_analytic._get_delivered_quantity_by_analytic([('amount', '<=', 0.0)])
+        lines_by_analytic = self.filtered(
+            lambda sol: sol.qty_delivered_method == 'analytic' or sol._is_reinvoicing_line()
+        )
+        domain = lines_by_analytic._get_delivered_quantity_by_analytic_domain()
+        mapping = lines_by_analytic._get_delivered_quantity_by_analytic(domain)
         for so_line in lines_by_analytic:
             delivered_qties[so_line] = mapping.get(so_line.id or so_line._origin.id, 0.0)
         return delivered_qties
+
+    @api.model
+    def _get_delivered_quantity_by_analytic_domain(self):
+        return Domain('amount', '<=', 0.0)
 
     def _get_downpayment_state(self):
         self.ensure_one()
@@ -1464,6 +1488,8 @@ class SaleOrderLine(models.Model):
         * Lines cannot be deleted if the order is confirmed.
         * Down payment lines who have not yet been invoiced bypass that exception.
         * Sections and Notes can always be deleted.
+        * Lines with expense policy `cost` and an ordered quantity 0 bypass this restriction
+          when deleted from the services and material view.
 
         :returns: Sales Order Lines that cannot be deleted
         :rtype: `sale.order.line` recordset
@@ -1473,6 +1499,12 @@ class SaleOrderLine(models.Model):
                 line.state == 'sale'
                 and (line.invoice_lines or not line.is_downpayment)
                 and not line.display_type
+                and not (
+                    self.env.context.get('from_services_and_material')
+                    and not line.product_uom_qty
+                    and len(line.analytic_line_ids) == 1
+                    and line.analytic_line_ids.unit_amount == line.qty_delivered
+                )
         )
 
     @api.ondelete(at_uninstall=False)
@@ -1838,3 +1870,25 @@ class SaleOrderLine(models.Model):
     def _get_rounding(self):
         self.ensure_one()
         return self.product_uom_id.rounding
+
+    def _is_reinvoicing_line(self):
+        self.ensure_one()
+        return (
+            not self.is_expense
+            and self.product_id.expense_policy != 'no'
+            and self.qty_delivered_method != 'manual'
+        )
+
+    def _is_line_reinvoicable(self):
+        """Determine whether this sale order line should be used to retrieve analytic lines
+        to be linked to the generated invoice.
+
+        Only non-expense lines invoiced based on delivered quantities are considered
+        valid sources for analytic line reinvoicing.
+        """
+        self.ensure_one()
+        return (
+            not self.is_expense
+            and self.product_id.invoice_policy == 'delivery'
+            and self.product_id.expense_policy != 'no'
+        )
--- a/addons/sale/tests/test_analytic_to_sale_to_invoice.py
+++ b/addons/sale/tests/test_analytic_to_sale_to_invoice.py
@@ -1,0 +1,411 @@
+from odoo.exceptions import UserError
+from odoo.fields import Command
+
+from odoo.addons.sale.tests.common import SaleCommon
+
+
+class TestAnalyticToSaleToInvoice(SaleCommon):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.reinvoice_at_cost_product = cls.env['product.product'].create({
+            'name': 'Reinvoice at Cost product',
+            'type': 'service',
+            'standard_price': 100,
+            'list_price': 110,
+            'expense_policy': 'cost',
+            'invoice_policy': 'delivery',
+        })
+        cls.reinvoice_at_sales_price_product = cls.env['product.product'].create({
+            'name': 'Reinvoice at Sales Price product',
+            'type': 'service',
+            'standard_price': 100,
+            'list_price': 110,
+            'expense_policy': 'sales_price',
+            'invoice_policy': 'delivery',
+        })
+        cls.services_sale_order = cls.env['sale.order'].create({
+            'partner_id': cls.partner.id,
+            'order_line': [
+                Command.create({
+                    'name': 'Reinvoice at Cost Line',
+                    'product_id': cls.reinvoice_at_cost_product.id,
+                }),
+                Command.create({
+                    'name': 'Reinvoice at Sales Price Line',
+                    'product_id': cls.reinvoice_at_sales_price_product.id,
+                }),
+            ],
+        })
+
+        # confirm the sale order
+        cls.services_sale_order.action_confirm()
+
+        cls.at_cost_aal = cls.env['account.analytic.line'].with_context(from_services_and_material=True).create({
+            'name': 'At cost Upsale line',
+            'product_id': cls.reinvoice_at_cost_product.id,
+            'unit_amount': 1,
+            'order_id': cls.services_sale_order.id,
+        })
+
+        cls.at_sale_price_aal = cls.env['account.analytic.line'].with_context(from_services_and_material=True).create({
+            'name': 'At sale price Upsale line',
+            'product_id': cls.reinvoice_at_sales_price_product.id,
+            'unit_amount': 1,
+            'order_id': cls.services_sale_order.id,
+        })
+
+    def test_upsale_lines_created_based_on_reinvoice_policy(self):
+        """Ensure that services upsold from analytic lines create proper sale order lines
+        depending on the product reinvoice policy:
+
+        - For products reinvoiced at cost: a new SO line is always created.
+        - For products reinvoiced at sales price: an existing SO line is reused if available.
+
+        Also verify that:
+        - Price unit is correctly set based on the reinvoice policy.
+        - Analytic account is automatically created on the sale order with the correct plan.
+        """
+        self.assertEqual(len(self.services_sale_order.order_line), 3)
+
+        reinvoice_at_cost_product_lines = self.services_sale_order.order_line.filtered(
+            lambda aal: aal.product_id == self.reinvoice_at_cost_product,
+        )
+        reinvoice_at_sales_price_product_lines = self.services_sale_order.order_line.filtered(
+            lambda aal: aal.product_id == self.reinvoice_at_sales_price_product,
+        )
+
+        self.assertEqual(
+            len(reinvoice_at_cost_product_lines),
+            2,
+            "When we upsale a service with reinvoice at cost, we always create a new line",
+        )
+
+        at_cost_upsale_order_line = reinvoice_at_cost_product_lines.filtered(
+            lambda aal: aal.product_uom_qty == 0,
+        )
+
+        self.assertEqual(
+            len(at_cost_upsale_order_line),
+            1,
+            "One line should be created with 0 ordered quantity for at cost product.",
+        )
+
+        self.assertEqual(
+            len(reinvoice_at_sales_price_product_lines),
+            1,
+            "When we upsale a service with reinvoice at sales price, we link an existing line if any exists",
+        )
+
+        at_sale_price_upsale_order_line = reinvoice_at_sales_price_product_lines
+
+        self.assertEqual(
+            at_sale_price_upsale_order_line,
+            self.at_sale_price_aal.so_line,
+            "Existing line should be linked with the analytic line",
+        )
+        self.assertEqual(
+            at_cost_upsale_order_line,
+            self.at_cost_aal.so_line,
+            "Newly created line should be linked with the analytic line",
+        )
+
+        self.assertEqual(
+            at_cost_upsale_order_line.price_unit,
+            self.reinvoice_at_cost_product.standard_price,
+            "Reinvoice at cost line should have the product cost as price unit",
+        )
+
+        self.assertEqual(
+            at_sale_price_upsale_order_line.price_unit,
+            self.reinvoice_at_sales_price_product.list_price,
+            "Reinvoice at sales price line should have the product list price as price unit",
+        )
+
+        self.assertIsNotNone(
+            self.services_sale_order.analytic_account_id,
+            "The sale order should have an analytic account created",
+        )
+
+        self.assertEqual(
+            self.services_sale_order.analytic_account_id.plan_id,
+            self.env.ref('sale.analytic_plan_sale_orders'),
+            "The sale order analytic account should have the sale orders analytic plan",
+        )
+
+    def test_quantity_update_on_analytic_line_updates_upsale_lines(self):
+        """Verify that changing the quantity (unit_amount) on analytic lines correctly
+        updates the delivered quantities on the linked sale order lines:
+
+        - Increasing quantity on an at-cost analytic line recomputes delivered qty.
+        - Setting quantity to zero on a sales-price analytic line should be recomputed.
+        """
+        self.at_cost_aal.unit_amount = 3
+        self.at_sale_price_aal.unit_amount = 0
+
+        at_cost_upsale_order_line = self.at_cost_aal.so_line
+        at_sale_price_upsale_order_line = self.at_sale_price_aal.so_line
+
+        self.assertEqual(
+            at_cost_upsale_order_line.qty_delivered,
+            3,
+            "The delivered quantity should be updated on upsale order line",
+        )
+        self.assertEqual(
+            at_sale_price_upsale_order_line.qty_delivered,
+            0,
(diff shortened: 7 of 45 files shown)


addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _get_or_create_account_from_so(self, order_id):
+        sale_order = self.env['sale.order'].browse(order_id)
+
+        if sale_order.analytic_account_id:
+            return sale_order.analytic_account_id
+
+        sale_order.analytic_account_id = self.env['account.analytic.account'].create(
+            sale_order._prepare_analytic_account_data(
+                plan_id=self.env.ref('sale.analytic_plan_sale_orders')
```
Feyensv: what if that plan was deleted?
Should/can we create the analytic account without a plan or should we raise ?

addons/sale/data/analytic_plan_data.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
```
Feyensv:
```suggestion
<odoo noupdate="1">
```

addons/sale_timesheet/models/account_move.py
```diff
@@ -9,7 +9,7 @@
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    timesheet_ids = fields.One2many('account.analytic.line', 'timesheet_invoice_id', string='Timesheets', readonly=True, copy=False, export_string_translation=False)
+    timesheet_ids = fields.One2many('account.analytic.line', 'reinvoice_id', string='Timesheets', readonly=True, copy=False, export_string_translation=False)
```
Feyensv: this field probably requires a domain now (unless it's not used anywhere)

addons/sale/models/account_analytic_line.py
Feyensv, line 1: Docstrings (and comments) would be welcome on all those new methods :eyes:

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,20 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+            link the analytic lines in this sale order to the invoice.
+            Then, we can know which analytic lines are invoiced in the sale order.
+        """
+        for line in self.filtered(lambda i: i.move_type == 'out_invoice' and i.state == 'draft').invoice_line_ids:
+            sale_line = line.sale_line_ids.filtered(self._filter_sale_lines_to_reinvoice)
+            if sale_line:
+                domain = line._analytic_line_domain_get_invoiced_lines(sale_line)
+
+                analytic_lines = self.env['account.analytic.line'].sudo().search(domain)
+                analytic_lines.write({'reinvoice_id': line.move_id.id})
```
Feyensv: Seems to be highly unefficient to do one search for each AML (also, `sale_line_ids` is multi-relational field, so might be safer to rename `sale_line` into `so_lines`)
Since what matters (and is set on the AAL) is the AM, shouldn't we iterate over the AM and process all the `AM.AML.SOL` at once ?

addons/sale_timesheet/models/hr_timesheet.py
```diff
@@ -35,12 +35,15 @@ def _domain_so_line(self):
     timesheet_invoice_type = fields.Selection(TIMESHEET_INVOICE_TYPES, string="Billable Type",
             compute='_compute_timesheet_invoice_type', compute_sudo=True, store=True, readonly=True)
     commercial_partner_id = fields.Many2one('res.partner', compute="_compute_commercial_partner")
-    timesheet_invoice_id = fields.Many2one('account.move', string="Invoice", readonly=True, copy=False, help="Invoice created from the timesheet", index='btree_not_null')
```
Feyensv, line 38: Are we sure we want to modify timesheets logic (do we need to consider this case in ours, or are they still separate except for the shared field) ?
Might be safer (and easier to differentiate both flows) to have separate fields no? 🤔

Feyensv (commented):
Didn't go too deep into the code or flows for now as it might change a lot according to our new concept (of new and separate `qty_delivered_method` values).

addons/sale_timesheet/models/hr_timesheet.py
nipl-odoo: To me I expect all analytic lines that generates an invoice via Sale order, to be linked to that invoice. Honestly, I think it's good that this logic is centralized in Sale. Not sure it's worth to keep logic separate when we will be doing exactly same thing at both the places 🤔 
> Are we sure we want to modify timesheets logic (do we need to consider this case in ours, or are they still separate except for the shared field) ?

I didn't actually modified the timesheet logic, I've just added mine in top of that, whole start_date end_date thing is still there

addons/sale/models/account_analytic_line.py
nipl-odoo: If some user deletes the plan then the view parsing can fail 🤔 Since I've passed the plan value in view's context for `auto_account_id` which automatically computes the related account from plan.

Maybe we can check if our new plan exists or not, if not then we create it under `analytic.project_plan` which is must for  analytic setup see `__get_all_plans()`. If we raise for that it could be headache for user to create a new plan by his own and setting it in technical context.

addons/sale_timesheet/models/hr_timesheet.py
Feyensv: Felt like it might have been safer to differentiate both flows (and with way less diff), but I agree it's cleaner technically :ok_hand: 
Will see what other teams say when pinged.

addons/sale/models/account_move.py
nipl-odoo: I just moved this method to `sale` it was already like that in `sale_timesheet`, I will try to optimize it 👍🏻

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,42 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+                ('reinvoice_id', '=', False),
+                '&',
+                    ('reinvoice_id.state', '=', 'cancel'),
+                    ('reinvoice_id.payment_state', '!=', 'invoicing_legacy'),
+                ('reinvoice_id.payment_state', '=', 'reversed')
+        ])
+
+    def _filter_sale_lines_to_reinvoice(self, so_line):
+        """Filter hook to get sale lines that should be re-invoiced."""
+        return not so_line.is_expense and so_line.product_id.invoice_policy == 'delivery'
```
xavierbol: why not defining that method in `sale.order.line` model to be able to use `self`? 🤷🏼

addons/sale/models/sale_order.py
```diff
@@ -2258,6 +2263,17 @@ def deliver_sold_quantity(self):
...
 
+    def _get_or_create_analytic_account(self, plan_id):
+        self.ensure_one()
+
+        if self.analytic_account_id:
+            return self.analytic_account_id
+
+        self.analytic_account_id = self.env['account.analytic.account'].create(
+            self._prepare_analytic_account_data(plan_id=plan_id)
+        )
```
xavierbol: nitpicking

```suggestion
        if not self.analytic_account_id:
            self.analytic_account_id = self.env['account.analytic.account'].create(
                self._prepare_analytic_account_data(plan_id=plan_id)
            )
```

addons/sale_timesheet/models/account_move.py
```diff
@@ -9,7 +9,15 @@
...
 
-    timesheet_ids = fields.One2many('account.analytic.line', 'timesheet_invoice_id', string='Timesheets', readonly=True, copy=False, export_string_translation=False)
+    timesheet_ids = fields.One2many(
+        'account.analytic.line',
+        'reinvoice_id',
+        string='Timesheets',
+        domain=[('project_id', '!=', False)],
+        readonly=True,
+        copy=False,
+        export_string_translation=False
```
xavierbol:
```suggestion
        export_string_translation=False,
```
😇

xavierbol (commented):

addons/sale/models/account_analytic_line.py
```diff
@@ -45,39 +46,66 @@ def create(self, vals_list):
     def write(self, vals):
         if self.env.context.get('from_services_and_material'):
             if 'order_id' in vals or 'product_id' in vals:
-                # Reset delivered quantity on old sale order line if order or product is changed
-                self._sync_so_quantity(0, self.so_line, True)
+                # Delete old sale order line if order or product is changed
+                self.so_line.unlink()
```
xavierbol: That's strange to completely remove the SOL but I don't know the flow you try to improve, so I will let @Feyensv decide. :)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+    _inherit = 'account.analytic.line'
+
+    reinvoice_id = fields.Many2one(
+        'account.move',
+        string="Invoice",
+        readonly=True,
+        copy=False,
+        help="Invoice created from related SO line",
+        index='btree_not_null',
+    )
```
JulienAlardot: `reinvoice_id` is a poor choice of name, it isn't called "a reinvoice", `reinvoicing_move_id` makes way more sense "reinvoice_id" doesn't carry any meaning of source or destination.
I actually believed at first that it was the move that would be reinvoiced and not the move that is the reinvoicing of another one

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get('from_services_and_material'):
+            for vals in vals_list:
+                if vals.get('order_id', False):
+                    plan_id = self.env.ref('sale.analytic_plan_sale_orders', raise_if_not_found=False)
+
+                    # If user deleted plan then fallback on project plan
+                    if not plan_id:
+                        plan_id, _ = self.env['account.analytic.plan']._get_all_plans()
```
JulienAlardot:
```suggestion
                        plan_id, _ trash = self.env['account.analytic.plan']._get_all_plans()
```
In odoo `_` as a dummy variable is still considered a bad practice (even if we're moving away from it)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+
+    reinvoice_id = fields.Many2one(
+        'account.move',
+        string="Invoice",
+        readonly=True,
+        copy=False,
+        help="Invoice created from related SO line",
+        index='btree_not_null',
+    )
+    so_line = fields.Many2one(
```
JulienAlardot, line 19: This pains me as a naming convention breach but as you're only moving it I guess we'll keep it that way :c

addons/sale/models/account_analytic_line.py
Feyensv: Indeed, we probably have to consider cases where the SO line cannot be deleted, and maybe set the quantity to 0 instead? :thinking:

nipl-odoo: When it's line which wasn't **created** by analytic line, maybe we can set quantity 0 on those line, but it will still magically let user to delete the line 🤷🏻‍♂️

addons/sale_timesheet/report/timesheets_analysis_report.py
```diff
@@ -12,7 +12,7 @@ class TimesheetsAnalysisReport(models.Model):
     order_id = fields.Many2one("sale.order", string="Sales Order", readonly=True)
     so_line = fields.Many2one("sale.order.line", string="Sales Order Item", readonly=True)
     timesheet_invoice_type = fields.Selection(TIMESHEET_INVOICE_TYPES, string="Billable Type", readonly=True)
-    timesheet_invoice_id = fields.Many2one("account.move", string="Invoice", readonly=True, help="Invoice created from the timesheet")
+    reinvoice_id = fields.Many2one("account.move", string="Invoice", readonly=True, help="Invoice created from the timesheet")
```
JulienAlardot: Same comment

JulienAlardot (changes requested):
Doesn't break anything in hr_expense, sale_expense, or any of those bridges so I'm happy
cc @epictete you may want to check this as it's analytic related

addons/sale/models/account_analytic_line.py
nipl-odoo: Me too, I will rename it in another PR in master maybe (There is one similar nasty field on product too ^^)

addons/analytic/models/analytic_line.py
```diff
@@ -225,6 +224,10 @@ class AccountAnalyticLine(models.Model):
         default=lambda self: self.env['decimal.precision'].precision_get("Percentage Analytic"),
     )
 
+    def write(self, vals):
+        self._check_can_write(vals)
+        return super().write(vals)
```
Feyensv, line 232: I don't really understand the need of this hook 🤔
Overriding `_check_can_write` doesn't seem better/easier than direclty overriding `write`.

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
+from odoo import api, fields, models
```
Feyensv, line 3: Missing license :eyes:

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+        string="Invoice",
+        readonly=True,
+        copy=False,
+        help="Invoice created from related SO line",
+        index='btree_not_null',
+    )
+    so_line = fields.Many2one(
+        'sale.order.line',
+        string='Sales Order Item',
+        index='btree_not_null',
```
Feyensv, line 25:
```suggestion
        index='btree_not_null',
        domain=lambda self: self._domain_so_line(),
```

```python
    def _domain_so_line(self):
        if self.env.context.get('from_services_and_material'):
            return []
        return [('qty_delivered_method', '=', 'analytic')]
```

(not fully sure contextual domains work, but it might be a safer way to keep the previous behavior in other places)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+            lines = super().create(vals_list)
+
+        return lines
+
+    def write(self, vals):
+        if self.env.context.get('from_services_and_material'):
+            if 'unit_amount' in vals:
+                # sync delivered quantity if quantity is changed
+                self._sync_so_quantity(vals['unit_amount'] - self.unit_amount, self.so_line)
```
Feyensv: Either iterate over `self`, or make it clear we do not support multi-edit in this context, because this will crash if self holds multiple records.


```suggestion
            self.ensure_one()
            if 'unit_amount' in vals:
                # sync delivered quantity if quantity is changed
                self._sync_so_quantity(vals['unit_amount'] - self.unit_amount, self.so_line)
```

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+
+            lines = super().create(vals_list)
+            lines._sync_so_lines()
+        else:
+            lines = super().create(vals_list)
+
+        return lines
+
+    def write(self, vals):
+        if self.env.context.get('from_services_and_material'):
```
Feyensv:
```suggestion
        if self and self.env.context.get('from_services_and_material'):
```

nitpicking and good practice (even more if we add the `ensure_one` check)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+                    so_line = self._create_so_line(
+                        qty_delivered=line.unit_amount,
+                        description=line.name,
+                        policy=policy,
+                    )
+
+                line.so_line = so_line
+
+    def _unsync_so_lines(self):
+        """ Revert synchronization of delivered quantities on related sale order lines.
```
Feyensv:
```suggestion
        """Revert synchronization of delivered quantities on related sale order lines.
```

https://docs.astral.sh/ruff/rules/surrounding-whitespace/

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+        if self.env.context.get('from_services_and_material'):
+            if 'unit_amount' in vals:
+                # sync delivered quantity if quantity is changed
+                self._sync_so_quantity(vals['unit_amount'] - self.unit_amount, self.so_line)
+
+            res = super().write(vals)
+
+            if 'order_id' in vals or 'product_id' in vals:
+                # remove/update old sale order line
+                self._unsync_so_lines()
```
Feyensv: Unless I'm wrong, you try to delete the linked soline regardless of whether there might have been multiple AAL targeting that same SOL (which can happen for lines based on sales price, where you try to find an existing line to update) 🤔 

Let's say I have a SO with product A (expense_policy = 'sales_price') with existing quantities.
Then I register an AAL for it, with the same product A, it'll increase the existing line delivered quantity for that line.
If I do delete/modify the linked AAL afterwards, it should (maybe?) remove that quantity from the matching SOL, but certainly not delete it 💥 

Edit: you have a test that covers this situation, but since the `product_uom_qty` of the SOL is 1, it's not deleted.

addons/sale/views/sale_menus.xml
```diff
@@ -40,13 +40,19 @@
             groups="sales_team.group_sale_salesman"
             sequence="20">
 
+            <menuitem
+                id="menu_service_material"
+                action="action_service_material"
+                sequence="10"
+            />
```
Feyensv, line 48: Does it make sense to show this menu if `analytic.group_analytic_accounting` feature group is not enabled ?
You don't see the `expense_policy` field on products unless the feature is enabled, so I'd say the menu should be hidden too.

Also, strangely, when I only install `sale_management`, I don't even see the `Analytics` setting block where one is supposed to enable the feature 🤔

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+        """ Adjust the delivered quantity on a linked sale order line based on
+        changes to the analytic line.
+
+        :param float quantity: New quantity value to synchronize with the sale order line.
+        :param sale.order.line so_line: Sale order line whose delivered quantity must be updated.
+
+        :rtype: None
+        """
+        if not so_line:
+            return
```
Feyensv: Since you already checked whether there was a line in 2 of the 3 call sites, I would stop handling it here, and check it before calling the method in the last call site too.

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+    def _get_existing_so_line(self):
+        """ Retrieve an existing sale order line from the related order that
+        matches the product of the current analytic line and has manual quantity delivered method.
+        """
+        return self.order_id.order_line.filtered(
+            lambda line: line.product_id == self.product_id
+            and line.qty_delivered_method == 'manual',
+        )[:1]
+
+    def _create_so_line(self, qty_delivered, description, policy):
```
Feyensv: Since the values are all coming from an AAL, why don't you call it directly on the AAL the SOL should be created from ?

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+                    line._sync_so_quantity(line.unit_amount, so_line)
+                    so_line.name = line.name
+                else:
+                    so_line = self._create_so_line(
+                        qty_delivered=line.unit_amount,
+                        description=line.name,
+                        policy=policy,
+                    )
+
+                line.so_line = so_line
```
Feyensv:
```suggestion
            so_line = self.env['sale.order.line']
            if (
                line.product_id.expense_policy == 'sales_price'
                and (so_line := line._get_existing_so_line())
            ):
                line._sync_so_quantity(line.unit_amount, so_line)

            line.so_line = so_line or line._create_so_line()
```

I feel like things could be simplified a bit here :eyes:

addons/sale/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,70 @@
...
+                    domain="[('expense_policy', '!=', 'no')]"
+                    required="True"
+                />
+                <field name="name" string="Description"/>
+                <field name="unit_amount" string="Quantity" optional="show"/>
+                <field name="product_uom_id" widget="many2one_uom"/>
+                <field name="so_line" string="Sales Order Line" optional="hide"/>
+                <field name="ref" optional="hide"/>
+                <field name="partner_id" optional="hide"/>
+                <field name="amount" optional="hide"/>
```
Feyensv, line 67: Missing `company_id` field in multi-company (can stay optional or not, but must be available).
In multi-company environment, you shouldn't be able to select SO's from companies different than the AAL company (is handled by the `check_company` on the field suggested in another comment), but you should also be able to select another of the enabled companies on the AAL (and unlock the SOs of the other companies)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+        copy=False,
+        help="Invoice created from related SO line",
+        index='btree_not_null',
+    )
+    so_line = fields.Many2one(
+        'sale.order.line',
+        string='Sales Order Item',
+        index='btree_not_null',
+    )
+    order_id = fields.Many2one('sale.order', string="Customer Order", index=True)
```
Feyensv:
```suggestion
    order_id = fields.Many2one('sale.order', string="Customer Order", index=True, check_company=True)
```

addons/sale/models/sale_order.py
```diff
@@ -357,6 +357,8 @@ def _rec_names_search(self):
     show_update_pricelist = fields.Boolean(
         string="Has Pricelist Changed", store=False)  # True if the pricelist was changed
 
+    analytic_account_id = fields.Many2one('account.analytic.account', string="Analytic Account")
```
Feyensv:
```suggestion
    analytic_account_id = fields.Many2one(string="Analytic Account", comodel_name='account.analytic.account')
```
`string` and `help` first (according to our team guidelines :eyes:)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,230 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get('from_services_and_material'):
+            for vals in vals_list:
+                if vals.get('order_id', False):
+                    plan_id = self.env.ref('sale.analytic_plan_sale_orders', raise_if_not_found=False)
+
+                    # If user deleted plan then fallback on project plan
+                    if not plan_id:
+                        plan_id, _ = self.env['account.analytic.plan']._get_all_plans()
```
Feyensv: Can it happen to have no `order_id` at this point ? 🤔 
It'll have raised earlier because of the required in the view.
Also, the `plan_id` lookup should be extracted outside the loop as it's not expected to change during the iterations.

addons/sale/models/account_move.py
```diff
@@ -48,6 +50,16 @@ def _compute_origin_so_count(self):
...
             move.sale_order_count = len(move.line_ids.sale_line_ids.order_id)
 
+    def _domain_services_analytic_line(self):
+        return Domain('reinvoice_move_id', 'in', self.ids)
+
+    @api.depends('line_ids.sale_line_ids')
+    def _compute_service_line_count(self):
+        for move in self:
+            move.service_line_count = self.env['account.analytic.line'].search_count(
+                self._domain_services_analytic_line(),
```
Feyensv:
```suggestion
                move._domain_services_analytic_line(),
```
:eyes:

addons/sale/models/sale_order_line.py
```diff
@@ -1477,7 +1477,7 @@ def _check_line_unlink(self):
 
     @api.ondelete(at_uninstall=False)
     def _unlink_except_confirmed(self):
-        if self._check_line_unlink():
+        if self._check_line_unlink() and not self.env.context.get('from_services_and_material'):
```
Feyensv: The context key shouldn't be sufficient to allow any deletion (ideally, you should duplicate here the additional checks done in `_unsync_so_lines`)

addons/sale/models/sale_order_line.py
```diff
@@ -1838,3 +1838,13 @@ def _can_be_edited_on_portal(self):
...
+
+    def _is_line_reinvoicable(self):
+        """ Determine whether this sale order line should be used to retrieve analytic lines
+        to be linked to the generated invoice.
+
+        Only non-expense lines invoiced based on delivered quantities are considered
+        valid sources for analytic line reinvoicing.
+        """
+        self.ensure_one()
+        return not self.is_expense and self.product_id.invoice_policy == 'delivery'
```
Feyensv: Shouldn't we also check that the product `expense_policy != 'no'` too here ?

addons/sale/tests/test_analytic_to_sale_to_invoice.py
Feyensv, line 1: `test_analytic_to_sale_to_invoice.py` ?
(not really inspired, but I don't like a functional menu name to decide the test name)

addons/sale/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,70 @@
...
+        <field name="view_mode">list</field>
+        <field name="view_id" ref="view_analytic_lines_service_list"/>
+        <field name="search_view_id" ref="view_analytic_lines_service_search"/>
+        <field name="context">{
+            'search_default_my_orders': 1,
+            'search_default_today': 1,
+            'from_services_and_material': 1,
+            'default_unit_amount': 1,
+        }</field>
+        <field name="domain"> [('product_id.expense_policy', '!=', 'no')]</field>
```
Feyensv: I fear this would show AAL not related to 'Services and materials' at all, shouldn't you also check whether the `order_id` field is set ? 🤔

addons/sale_timesheet/models/account_move.py
```diff
@@ -66,40 +81,33 @@ def action_view_timesheet(self):
...
         ])
-        timesheets_sudo.write({'timesheet_invoice_id': False})
+        timesheets_sudo.write({'reinvoice_move_id': False})
         return result
+
+    def _domain_services_analytic_line(self):
+        """To make sure we see only services analytic line that are not timesheets"""
+        domain = super()._domain_services_analytic_line()
+
+        return domain & Domain('project_id', '=', False)
```
Feyensv:
```suggestion
    def _domain_services_analytic_line(self):
        """Make sure we see only services analytic line that are not timesheets."""
        return super()._domain_services_analytic_line() & Domain('project_id', '=', False)
```

could be on one-line :eyes:

Feyensv (commented):
Deeper review :duck:

Thanks for the tests and other improvements ✨

addons/sale_timesheet/report/timesheets_analysis_report.py
```diff
@@ -12,7 +12,7 @@ class TimesheetsAnalysisReport(models.Model):
     order_id = fields.Many2one("sale.order", string="Sales Order", readonly=True)
     so_line = fields.Many2one("sale.order.line", string="Sales Order Item", readonly=True)
     timesheet_invoice_type = fields.Selection(TIMESHEET_INVOICE_TYPES, string="Billable Type", readonly=True)
-    timesheet_invoice_id = fields.Many2one("account.move", string="Invoice", readonly=True, help="Invoice created from the timesheet")
```
Feyensv, line 15: Since it's a SQL view related to timesheet, is it really necessary to rename that one? 🤔

addons/analytic/models/analytic_line.py
nipl-odoo: The hook was already there in `hr_timesheet` and was being shared by `project_timesheet_holidays`, `sale_timesheet` and `timesheet_grid`, I moved it to `analytic` so it can be used in unrelated module like `sale`, and it's not worth changing these all modules just for overriding `write` IMO, and personally I think it's nice that we keep business logic and custom raised errors separate, What do you think?

addons/sale/models/account_analytic_line.py
nipl-odoo: I guess this won't be issue after the change we discussed yesterday regarding always decrease quantity on `at sale price` line 😄

addons/sale/models/account_analytic_line.py
nipl-odoo: It's weird to have a single line method 😅 But okay

addons/sale/views/sale_menus.xml
nipl-odoo: > Does it make sense to show this menu if analytic.group_analytic_accounting feature group is not enabled ?

Nope, Adding group there ;)
> Also, strangely, when I only install sale_management, I don't even see the Analytics setting block where one is supposed to enable the feature 🤔

If you install `accountant` the enterprise one then only you get to see full accounting features and eventually analytic stuff

addons/sale/models/sale_order_line.py
nipl-odoo: I need to duplicate only the check when deleting line from `_unsync_so_lines` yes 👀

addons/sale/models/sale_order_line.py
nipl-odoo: It'll refrain timesheets from getting linked 🤔

addons/sale/views/account_analytic_line_views.xml
nipl-odoo: Indeed we can also check for `project_id == False `in sale_timesheet I guess 🤔 

Edit: Not needed `project_id` check since product is never set on timesheets 👍🏻

addons/sale_timesheet/report/timesheets_analysis_report.py
nipl-odoo: Did it purely for consistency, and I don’t have a strong reason to rename it, so I’m okay leaving it as is ;)

addons/analytic/models/analytic_line.py
Feyensv: I still think it's not useful (unless overrides impact each other, which doesn't seem to be the case here).
Anyway, can stay like this for now, can remove it someday if needed :D

addons/sale/views/sale_menus.xml
Feyensv: Yeah, but still was annoyed to have to enable manually the feature group on the current user to make it work without the accountant module :D

addons/sale/models/sale_order_line.py
Feyensv: Might still be safer to override in `sale_timesheet` to add their case than to be potentially too generic here :thinking:

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,213 @@
...
+        ordered quantity, the delivered quantity on the related sale order line
+        is decreased by the amount contributed by the analytic line being deleted.
+
+        Lines originating from timesheets or expenses (i.e., not using the manual
+        delivery method) are ignored.
+        """
+        for line in self.filtered(lambda line: line.so_line):
+            if line.so_line.qty_delivered_method != 'manual':
+                continue
+            if line.product_id.expense_policy == 'cost' and not line.so_line.product_uom_qty:
```
Feyensv:
```suggestion
            if (
                line.product_id.expense_policy == 'cost'
                and not line.so_line.product_uom_qty
                and line.unit_amount == line.so_line.qty_delivered
            ):
```
?

addons/sale/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,72 @@
...
+        <field name="view_mode">list</field>
+        <field name="view_id" ref="view_analytic_lines_service_list"/>
+        <field name="search_view_id" ref="view_analytic_lines_service_search"/>
+        <field name="context">{
+            'search_default_my_orders': 1,
+            'search_default_today': 1,
+            'from_services_and_material': 1,
+            'default_unit_amount': 1,
+        }</field>
+        <field name="domain"> [('product_id.expense_policy', '!=', 'no'), ('order_id', '!=', False)]</field>
```
Feyensv:
```suggestion
        <field name="domain">[('product_id.expense_policy', '!=', 'no'), ('order_id', '!=', False)]</field>
```
nit

addons/sale_timesheet/models/hr_timesheet.py
```diff
@@ -81,11 +85,16 @@ def _compute_so_line(self):
         for timesheet in self.filtered(lambda t: not t.is_so_line_edited and t._is_not_billed()):  # Get only the timesheets are not yet invoiced
             timesheet.so_line = timesheet.project_id.allow_billable and timesheet._timesheet_determine_sale_line()
 
-    @api.depends('timesheet_invoice_id.state')
+    @api.depends('so_line')
+    def _compute_order_id(self):
+        for timesheet in self:
+            timesheet.order_id = timesheet.so_line.order_id if timesheet.so_line else False
```
Feyensv: It's kinda a bad practice to make a field computed in a sub-module, as multiple modules doing so will conflict.
It's better to make it compute in the base module, even if the compute method is simple (or even does nothing but assign a default value).

Same comment for the `so_line` field too :eyes:

addons/sale/models/account_analytic_line.py
Feyensv, line 1: `product_uom_id` and `so_line` will have to be handled in the upgrade script (to avoid triggering the compute on thousands of records :eyes:)

Whenrow (approved):

addons/sale_stock/models/sale_order_line.py
```diff
@@ -176,26 +181,28 @@ def _compute_is_mto(self):
...
-        super(SaleOrderLine, self)._compute_qty_delivered_method()
+        super()._compute_qty_delivered_method()
 
         for line in self:
-            if not line.is_expense and line.product_id.type == 'consu':
+            if line.qty_delivered_method == 'manual' and line.product_id.type == 'consu':
                 line.qty_delivered_method = 'stock_move'
 
+    def _get_qty_delivered_method_reinvoicing(self):
+        return 'analytic_and_moves'
```
Feyensv: If the product is a service, it should stay `analytic` here (to avoid fetching meaningless stock information)

addons/sale_timesheet/models/sale_order_line.py
```diff
@@ -205,3 +197,14 @@ def _get_action_per_item(self):
     @api.model
     def _get_product_service_policy(self):
         return super()._get_product_service_policy() + ['delivered_timesheet']
+
+    def _is_line_reinvoicable(self):
+        """If line is supposed to be delivered based on a timesheet then we skip the expesne policy
```
Feyensv:
```suggestion
        """If line is supposed to be delivered based on a timesheet then we skip the expense policy
```

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+    order_id = fields.Many2one(
+        string="Customer Order",
+        comodel_name='sale.order',
+        compute='_compute_order_id',
+        store=True,
+        readonly=False,
+        index=True,
+    )
+
+    def _compute_so_line(self):
```
william-andre, line 37: If there is no dependency, this doesn't do anything since the value is already `False`

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+            plan_id = self.env.ref('sale.analytic_plan_sale_orders', raise_if_not_found=False)
+            # If user deleted plan then fallback on project plan
+            if not plan_id:
+                plan_id, _other_plans = self.env['account.analytic.plan']._get_all_plans()
+
+            column_name = plan_id._column_name()
+
+            for vals in vals_list:
+                order_id = self.env['sale.order'].browse(vals['order_id'])
+                vals[column_name] = order_id._get_or_create_analytic_account(plan_id).id
```
william-andre: Wouldn't it make sense to do all of this in an inverse function for `so_line`?

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+                line.so_line = False
+
+    def _compute_order_id(self):
+        for line in self:
+            if not line.order_id:
+                line.order_id = False
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get('from_services_and_material'):
```
william-andre, line 49: Why do we need a context, if we provide `order_id`? (or `so_line`)

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+                # create/update new sale order line for new order or product
+                self._sync_so_lines()
+        return res
+
+    def _check_can_write(self, vals):
+        if self.filtered(
+            lambda aal: aal.reinvoice_move_id and aal.reinvoice_move_id.state != 'cancel'
+        ):
+            if any(field_name in vals for field_name in self._restricted_fields_when_invoiced()):
+                raise UserError(self._get_invoiced_line_write_error())
```
william-andre, line 87: Why do we need this restriction?

For instance on an invoice, we ca still change the analytic values even if the entries are locked and hashed (highest protection possible). Both from the invoice and from the analytic items.

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+
+        if (
+            'unit_amount' in vals
+            and vals['unit_amount'] < 0
+            and self.env.context.get('from_services_and_material')
```
william-andre: A constraint using the context 🤪

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+        ):
+            raise UserError(self.env._("You cannot set a negative quantity on services."))
+
+        super()._check_can_write(vals)
+
+    def _restricted_fields_when_invoiced(self):
+        return ['unit_amount', 'order_id', 'product_id', 'so_line', 'date']
+
+    def _get_invoiced_line_write_error(self):
+        return self.env._("You cannot modify already invoiced services.")
```
william-andre, line 98: Why a separate function?

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+        expenses are left unchanged.
+        """
+        if any(
+            line.reinvoice_move_id and line.reinvoice_move_id.state == 'posted' for line in self
+        ):
+            raise UserError(self._get_invoiced_line_delete_error())
+        self._unsync_so_lines()
+
+    def _get_invoiced_line_delete_error(self):
+        return self.env._("You cannot remove already invoiced services.")
```
william-andre, line 117: Same

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+        compute='_compute_order_id',
+        store=True,
+        readonly=False,
+        index=True,
+    )
+
+    def _compute_so_line(self):
+        for line in self:
+            if not line.so_line:
+                line.so_line = False
```
william-andre: What does this do?

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+    so_line = fields.Many2one(
+        string='Sales Order Item',
+        comodel_name='sale.order.line',
+        compute='_compute_so_line',
+        store=True,
+        readonly=False,
+        index='btree_not_null',
+        domain=lambda self: self._domain_so_line(),
+    )
+    order_id = fields.Many2one(
```
william-andre, line 28:
```suggestion
    sale_order_id = fields.Many2one(
```

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+        string="Customer Order",
+        comodel_name='sale.order',
+        compute='_compute_order_id',
+        store=True,
```
william-andre, line 32: Why do we need to store both the SO and the line?

addons/sale/tests/test_analytic_to_sale_to_invoice.py
```diff
@@ -0,0 +1,415 @@
...
+        )
+        reinvoice_at_sales_price_product_lines = self.services_sale_order.order_line.filtered(
+            lambda aal: aal.product_id == self.reinvoice_at_sales_price_product,
+        )
+
+        self.assertEqual(
+            len(
+                self.services_sale_order.order_line.filtered(
+                    lambda aal: aal.product_id == self.reinvoice_at_cost_product,
+                ),
```
william-andre: Isn't this `reinvoice_at_cost_product_lines`?

addons/sale/tests/test_analytic_to_sale_to_invoice.py
```diff
@@ -0,0 +1,415 @@
...
+            "One line should be created with 0 ordered quantity for at cost product.",
+        )
+
+        self.assertEqual(
+            len(reinvoice_at_sales_price_product_lines),
+            1,
+            "When we upsale a service with reinvoice at sales price, we link an existing line if any exists",
+        )
+
+        at_sale_price_upsale_order_line = reinvoice_at_sales_price_product_lines
```
william-andre, line 101: Why? What does this do?

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,38 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+        """Link analytic lines originating from a sale order to the corresponding invoice.
+
+        When an invoice is generated from a sale order, this method associates all
+        related analytic lines with the invoice by setting their `reinvoice_move_id` field.
+
+        This allows tracking which analytic entries have already been invoiced.
+        """
+        invoices = self.filtered(lambda i: i.move_type == 'out_invoice' and i.state == 'draft')
+        for invoice in invoices:
+            invoice._get_analytic_lines_to_link().write({'reinvoice_move_id': invoice.id})
```
william-andre: Let the ORM decide when to write

```suggestion
            invoice._get_analytic_lines_to_link().reinvoice_move_id = invoice
```

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,38 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+
+    def _get_analytic_lines_to_link(self):
```
william-andre: Name too generic

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,38 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+    def _analytic_line_domain_get_invoiced_lines(self, so_lines):
+        return Domain([
+            ('so_line', 'in', so_lines.ids),
+            '|', '|',
+                ('reinvoice_move_id', '=', False),
+                '&',
+                    ('reinvoice_move_id.state', '=', 'cancel'),
+                    ('reinvoice_move_id.payment_state', '!=', 'invoicing_legacy'),
+                ('reinvoice_move_id.payment_state', '=', 'reversed')
+        ])
```
william-andre: If you use the `Domain` class, use the `Domain` class

```suggestion
        return Domain('so_line', 'in', so_lines.ids) & (
            Domain('reinvoice_move_id', '=', False)
            | Domain('reinvoice_move_id.payment_state', '=', 'reversed')
            | (
                Domain('reinvoice_move_id.state', '=', 'cancel')
                & Domain('reinvoice_move_id.payment_state', '!=', 'invoicing_legacy')
            )
        )
```

But also, use the `any` operator to avoid repeating, and make sure that we don't mix records from different moves
```suggestion
        return Domain('so_line', 'in', so_lines.ids) & (
            Domain('reinvoice_move_id', '=', False)
            | Domain('reinvoice_move_id', 'any', (
                Domain('payment_state', '=', 'reversed')
                | (Domain('state', '=', 'cancel') & Domain('payment_state', '!=', 'invoicing_legacy')
            ))
        )
```

addons/sale/models/account_analytic_line.py
william-andre: If the SO is linked to a project, don't we also want to set `account_id`?

addons/sale_timesheet/models/account_move.py
```diff
@@ -36,7 +44,14 @@ def _compute_timesheet_total_duration(self):
...
+            [
+                ('reinvoice_move_id', 'in', self.ids),
+                ('project_id', '!=', False),
+            ],
+            ['reinvoice_move_id'],
+            ['__count'],
+        )
         mapped_data = {timesheet_invoice.id: count for timesheet_invoice, count in timesheet_data}
         for invoice in self:
             invoice.timesheet_count = mapped_data.get(invoice.id, 0)
```
william-andre: Isn't it possible to get here in `onchange`?
If so, it requires a `_origin`

(and the dict comprehension above could be a simple call to `dict` instead, and we don't need to use the `.id`)

addons/sale/models/account_move.py
```diff
@@ -204,3 +222,38 @@ def _get_partner_credit_warning_exclude_amount(self):
...
+    def _get_analytic_lines_to_link(self):
+        # This method is kept separate because some modules require more granular control
+        # over how analytic lines are retrieved. Instead of aggregating all sale order lines
+        # from invoice lines at once, certain implementations need to search analytic lines
+        # per invoice line individually. See the override in `sale_subscription_timesheet`.
+        so_lines = self.invoice_line_ids.sale_line_ids.filtered(lambda line: line._is_line_reinvoicable())
+
+        return self.env['account.analytic.line'].sudo().search(
+            self._analytic_line_domain_get_invoiced_lines(so_lines)
+        )
```
william-andre, line 247: search in a loop 😬

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+            return [
+                ('qty_delivered_method', 'in', ('analytic', 'timesheet', 'stock_move')),
+                ('product_id.expense_policy', '!=', 'no'),
+            ]
+        return [('qty_delivered_method', '=', 'analytic')]
+
+    reinvoice_move_id = fields.Many2one(
+        string="Invoice",
+        comodel_name='account.move',
+        readonly=True,
```
william-andre, line 14: If it's readonly, why do we even need to store it?
Seems to be just a search with a domain?

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,203 @@
...
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    def _domain_so_line(self):
+        if self.env.context.get('from_services_and_material'):
+            return [
+                ('qty_delivered_method', 'in', ('analytic', 'timesheet', 'stock_move')),
+                ('product_id.expense_policy', '!=', 'no'),
+            ]
+        return [('qty_delivered_method', '=', 'analytic')]
```
william-andre: Why is this not in the view?

In any case if we keep it, please put this after the field declaration.

addons/sale/models/account_analytic_line.py
nipl-odoo: this compute is overridden in `sale_timesheet` and I was told it's not good practice to make fields compute in child modules so I made it compute in parent module, but totally agree that `if` condition is wrong
ref https://github.com/odoo/odoo/pull/245216#discussion_r2781752359

addons/sale/models/account_analytic_line.py
nipl-odoo: Context is not needed in create, but it was in write because of one edge case so added it just as a guard

addons/sale/models/account_analytic_line.py
nipl-odoo: Because in this view changing `product_id`, `order_id` or `product_uom_qty` will directly affect so lines and once that line is invoiced we shouldn't be changing that line at all

addons/sale/models/account_analytic_line.py
william-andre: Just do this then
```python
def _compute_so_line(self):
    return  # or continue, whatever
```

addons/sale/models/account_analytic_line.py
nipl-odoo: I was not sure adding this constrain globally in all AALs 😄, does it make sense to restrict negative quantities on all AALs?

addons/sale/models/account_analytic_line.py
william-andre: Which edge case?

addons/sale/models/account_analytic_line.py
nipl-odoo: Used in `sale_timesheet` if any AAL with `project_id` (timesheet) is being removed/changed we must show `You cannot modify already invoiced timesheets.` 
same for the comment below

addons/sale/models/account_analytic_line.py
nipl-odoo: Because in our new view we let user select order_id and based on that we determine so_line in create/write, but in timesheets it's kinda opposite

addons/sale/models/account_analytic_line.py
william-andre: Wouldn't there be a constraint in the sales order itself that would prevent that from happening?

addons/sale/models/account_analytic_line.py
william-andre: certainly not.
But what about checking if it is linked to a sale order or a reinvoice document?

addons/sale/models/account_analytic_line.py
william-andre: But it shouldn't be called in the same use cases :thinking: 
The message needs to stick with the condition that failed. One is for timesheets, the other for services.

addons/sale/models/account_analytic_line.py
nipl-odoo: If `so_line` is changed/reassigned in timesheet because of any dependencies, `order_id` will also be in `vals` of write, that was happening in one of the test case I don't recall which one is it 🤔

addons/sale/models/account_analytic_line.py
william-andre: So knowing the line is enough in both cases :thinking: 
In timesheet: they set the so line, from there computing the SO is trivial
In sale: you set the order, and you are able to get the right line immediately. So just keep the right line, and you can compute the SO like above.

addons/sale/models/account_analytic_line.py
william-andre: probably a sign that storing the order and the line is an issue :wink:

addons/sale/models/account_analytic_line.py
nipl-odoo: The field was stored in `sale_timesheet` anyways, so didn't try to change that 🤔

addons/sale/tests/test_analytic_to_sale_to_invoice.py
nipl-odoo: Only for naming conistancy TBH, but I can remove it if you think it's problematic, I just thought it's easier to read test with this variable 😄

addons/sale/models/account_analytic_line.py
nipl-odoo: It was stored in `sale_timesheet`, I just moved it here, but I would indeed like to try it in another PR 😄

addons/sale/models/account_analytic_line.py
nipl-odoo: I feel it could be confusing for users to see *“You can't modify invoiced SO line”* even though what they are actually editing is an AAL 🤔

Also, we should block the action before the write is executed. Some side effects happen after the AAL write call, and those can trigger errors on the SO lines afterward. So if we don’t restrict it earlier, the field would be written on AAL anyways, No?

addons/sale/models/account_analytic_line.py
nipl-odoo: We are checking exactly same thing at both the places 🤔 it's just when `sale_timesheet` is installed we restrict some extra fields. It'll duplicate some code if I keep these conditions separate, also I would need to define a hook to differentiate between service line and timesheet line otherwise it'll always show `You cannot edit invoiced services`, so it isn't beautiful either, no?

addons/sale_timesheet/models/account_move.py
nipl-odoo: Nope I guess we won't be here in onchange, since this field is for computing number of timesheets which were invoiced in current invoice, for new/extra timesheets a new/extra invoice is created so it's fine I guess 🤔

addons/sale/models/account_move.py
nipl-odoo: The loop is intentional here because each analytic line must be linked to its exact originating invoice. We can’t batch everything in a single search since `reinvoice_move_id` needs to be set per invoice.

I can indeed batch the AAL search, but the tricky part is mapping the results back, without searching per invoice, I don’t have a reliable way to know which lines belong to which invoice. 🤔

(It is even worse in `sale_subscription_timesheet` where I am doing search on each AML of each AM but I don't see any better way for that)

addons/sale/models/account_analytic_line.py
nipl-odoo: agreed, but we would still need to store it in `sale_timesheet` we want it for doing groupby in portal, so storing it in `sale_timesheet` but not in `sale` is a sign to store it in `sale`, No? 
https://github.com/odoo/odoo/pull/245216/changes#diff-4f5b5c045b974e5e61eafae8e89f43850468b71173a0dba5d952b844a19c61cdL42

addons/sale/models/account_analytic_line.py
nipl-odoo: Moving this in inverse means whenever so_line changes this will be called, but we just need to handle AAs when AAL is created otherwise it doesn't really matters what happens to SOL no?


Though I need to handle case when order_id is changed I must update related column name with new SO's AA.
qq: Is it technically okay if we set `x_plan_*_id` or `account_id` **after** the AAL is created? 🤔

addons/sale/models/account_analytic_line.py
william-andre: It’s confusing for the user to have to create an analytic line in order to modify a sales order in the first place...

william-andre: > So if we don’t restrict it earlier, the field would be written on AAL anyways, No?

That’s not how transactions work. Everything or nothing gets committed

william-andre: > So if we don’t restrict it earlier, the field would be written on AAL anyways, No?

That’s not how transactions work. Everything or nothing gets committed

Feyensv: > It’s confusing for the user to have to create an analytic line in order to modify a sales order in the first place...

I agree I don't get why this new flow requires AAL in the first place (and a task was cancelled that kinda did the same with SOlines directly), but that's how the spec was designed, with FP behind it (for the P project))

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,201 @@
...
+            'order_id': self.order_id.id,
+            'product_id': self.product_id.id,
+            'product_uom_id': self.product_uom_id.id,
+            'product_uom_qty': 0,
+        }
+
+        if self.product_id.expense_policy == 'cost':
+            values['price_unit'] = self.product_id.standard_price
+        elif self.product_id.expense_policy == 'sales_price':
+            values['price_unit'] = self.product_id.list_price
```
Feyensv: Missing a currency conversion (if SO is in different currency than company one) for cost lines
And for sales prices lines, I believe the `price_unit` shouldn't be set, to let the SOL `_compute_price_unit` consider the pricelist no?

Feyensv (approved):

addons/sale/models/account_analytic_line.py
```diff
@@ -0,0 +1,220 @@
...
+            'product_id': self.product_id.id,
+            'product_uom_id': self.product_uom_id.id,
+            'product_uom_qty': 0,
+        }
+
+        if self.product_id.expense_policy == 'cost':
+            values['price_unit'] = self.product_id.currency_id.with_company(
+                self.order_id.company_id
+            )._convert(
+                self.product_id.standard_price, self.order_id.currency_id, round=False
```
Feyensv:
```suggestion
                self.product_id.with_company(self.order_id.company_id)standard_price, self.order_id.currency_id, round=False
```
forgot that with_company, it matters since the cost is company dependent.


```python
            product = self.product_id.with_company(self.order_id.company_id)
values['price_unit'] =product.currency_id._convert(
                product.standard_price, self.order_id.currency_id, round=False
```
might be simpler then

addons/sale/models/sale_order.py
```diff
@@ -2269,6 +2274,15 @@ def deliver_sold_quantity(self):
...
                 line.qty_delivered = line.product_uom_qty
 
+    def _get_or_create_analytic_data(self, plan_id):
+        self.ensure_one()
+
+        if not self.analytic_account_id:
+            self.analytic_account_id = self.env['account.analytic.account'].create(
+                self._prepare_analytic_account_data(plan_id=plan_id)
+            )
+        return self.analytic_account_id, plan_id
```
Feyensv: Instead of returning a tuple, can't you rely on the `plan_id` field of the returned account instead ? 🤔 
(I find `_get_or_create_analytic_data` kinda ambiguous)

addons/sale/models/account_analytic_line.py
william-andre: > qq: Is it technically okay if we set `x_plan_*_id` or `account_id` **after** the AAL is created? 🤔

Yes.
An analytic line is supposed to just be a record holding raw information on which we do reporting and **nothing else**. It's not supposed to trigger weird behavior in other modules...

addons/sale/models/account_analytic_line.py
william-andre: Displaying a message related to services when the analytic line is related to a timesheet (or vice versa) doesn't make sense

Feyensv (approved):
@robodoo r+

addons/sale/models/account_analytic_line.py
xavierbol: @nipl-odoo no, we set the invoice created in the SO to know which timesheets are invoiced to make them readonly in timesheets app. TBH, I don't really understand why we name the field `reinvoice` since the label says `Invoice`. 😕

Anyways, that's not a big deal. 🙂

addons/sale/models/account_move.py
```diff
@@ -48,6 +50,16 @@ def _compute_origin_so_count(self):
...
 
+    def _domain_services_analytic_line(self):
+        return Domain('reinvoice_move_id', 'in', self.ids)
+
+    @api.depends('line_ids.sale_line_ids')
+    def _compute_service_line_count(self):
+        for move in self:
+            move.service_line_count = self.env['account.analytic.line'].search_count(
+                move._domain_services_analytic_line(),
+            )
```
xavierbol, line 61: why not doing a read_group instead of a search in the loop? 🤔

addons/sale/models/account_analytic_line.py
nipl-odoo: @xavierbol Honestly, renaming it to reinvoice_move_id wasn’t my call, it was mentioned in the spec by FP, and since it looked like a straightforward rename, I didn’t push back on it 😕

That said, I’m open to better naming if you have something clearer in mind. I’ll be opening a PR to rename a few fields after the freeze anyway, so we can adjust it there if needed. 😄

addons/sale/models/account_move.py
nipl-odoo: ~Will be~ fixed in https://github.com/odoo/odoo/pull/248840

addons/sale/models/account_analytic_line.py
nipl-odoo: Here it isssss 
- https://github.com/odoo/odoo/pull/252144

PR #245217, opened 2026-01-22
[FIX] html_builder, website, website_sale: avoid timeout dialog actions

blse-odoo:
Commit 6df83abb35c95ab42e55d9a08cf6c411efa64b3e added a timeout on operations, as an heuristic to detect when an operation is stuck. This timeout may be triggered when the action opens a dialog and wait for user choice.

This commit sets `canTimeout = false` on actions that open a dialog and wait for user choice in the `apply` method.

Steps to reproduce:
- Open website builder
- Click on "Theme" tab
- Click "Add a language"
- Wait a bit
- Bug: It show the error message "A technical issue occurred..."

task-5867364

### DIFF
--- a/addons/html_builder/static/src/core/operation_plugin.js
+++ b/addons/html_builder/static/src/core/operation_plugin.js
@@ -56,7 +56,7 @@ export class OperationPlugin extends Plugin {
 
         this.services.notification.add(
             _t(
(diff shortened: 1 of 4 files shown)


FrancoisGe:
@robodoo r+

PR #212173, opened 2025-05-28
[IMP] mail,*: add information in avatar card in multi-company

KevGerard:
*: hr,hr_holidays

This commit aims to add employee public information in the avatar card
popover even when the employee record is available in another company.

This commit should follow the current ACLs and only show information
that is allowed to the current user.

The result of this commit is that user have a better understanding of
the user they are interacting with even though the company might not be
activated.

Before this commit:
When following those steps:
- create user A with employee A in company A
- create user B with employee B in company B
- log into discuss with with user A
- start a conversation with user B
- click on the avatar of user B

It was not possible to see the out of office nor the employee
information (phone, email, department, building, office,...) if the
company was not activated.

task-4648745
task-4784252 (partial)

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245224, opened 2026-01-22
[FIX] hr_holidays, *: no access error on fetching portal partner data

Atovange:
* = test_discuss_full

Before this commit, mentioning a portal user from another company would result in an access error.

Steps to reproduce:
1. Install `hr_holidays` module. 2 Have a portal user in company A.
3. Switch the active company to company B.
4. In any chatter, try to mention said portal user.

This happens because portal user read access depends on the current active company (see `res_users_rule`).
The access error happens since [1], which added user information to the partner's default Store fields.

[1] https://github.com/odoo/odoo/pull/212173

task-5499827

### DIFF
--- a/addons/hr_holidays/models/res_partner.py
+++ b/addons/hr_holidays/models/res_partner.py
@@ -37,7 +37,8 @@ class ResPartner(models.Model):
     def _to_store_defaults(self, target):
         defaults = super()._to_store_defaults(target)
         if target.is_internal(self.env):
+            # sudo: res.users - to access other company's portal user leave date
             defaults.append(
-                Store.One("main_user_id", Store.Many("employee_ids", "leave_date_to"))
+                Store.One("main_user_id", Store.Many("employee_ids", "leave_date_to", sudo=True)),
             )
         return defaults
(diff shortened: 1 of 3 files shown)


addons/hr_holidays/models/res_partner.py
```diff
@@ -37,7 +37,8 @@ def _get_on_leave_ids(self):
     def _to_store_defaults(self, target):
         defaults = super()._to_store_defaults(target)
         if target.is_internal(self.env):
+            # sudo: res.users - to access other company's portal user leave date
             defaults.append(
-                Store.One("main_user_id", Store.Many("employee_ids", "leave_date_to"))
+                Store.One("main_user_id", Store.Many("employee_ids", "leave_date_to", sudo=True)),
```
alexkuhn, line 42: Basically copy of original code of `_to_store_defaults` but here this is extended with `hr_holidays` stuffs for internal users:
https://github.com/odoo/odoo/blob/19.0/addons/mail/models/res_partner.py#L288

alexkuhn (approved):
@robodoo delegate+ r+

PR #245226, opened 2026-01-22
[REF] iot_drivers: simplify serial drivers

loouis-t:
We simplify serial drivers by:
- making toledo scale the base scale driver (adam can still inherit),
- removing community specific code to maintain consistency (IoT is now enterprise only),
- simplifying scale serial port read,

see odoo/enterprise#105166

### DIFF
--- a/addons/iot_drivers/iot_handlers/drivers/serial_scale_driver.py
+++ b/addons/iot_drivers/iot_handlers/drivers/serial_scale_driver.py
@@ -37,101 +37,42 @@ Toledo8217Protocol = SerialProtocol(
 
 
 class ScaleDriver(SerialDriver):
-    """Abstract base class for scale drivers."""
-    last_sent_value = None
+    """Driver for the Toldedo 8217 serial scale."""
+    _protocol = Toledo8217Protocol
+    last_sent_value: float = None
 
-    def __init__(self, identifier, device):
+    def __init__(self, identifier: str, device: dict):
         super().__init__(identifier, device)
-        self.device_type = 'scale'
-        self._set_actions()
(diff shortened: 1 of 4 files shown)


addons/iot_drivers/iot_handlers/drivers/serial_scale_driver.py
```diff
@@ -175,9 +131,10 @@ def _read_status(self, answer):
...
-                    self.data.update({
-                        'result': 0,
-                        'status': self._status,
-                    })
-                    break
+                    _logger.debug(
+                        "Scale error: %s. Status string: %s. Scale answer: %s.",
+                        status_char_error_bits[index], binary_status_char, answer
+                    )
+                    self.data.udpate({"status": "error", "result": 0})
```
MaxTheWhale:
```suggestion
                    self.data.update({"status": "error", "result": 0})
```

MaxTheWhale (approved):
Tested and scale is still working correctly with 19.0 DB ✅ 

Just need to fix typo and rebase

addons/iot_drivers/iot_handlers/drivers/serial_scale_driver.py
```diff
@@ -155,13 +109,15 @@ def supported(cls, device):
     def _get_raw_response(connection):
```
MaxTheWhale, line 96: This method is now defined twice, I think we can just remove the looping version? (unless Adam scale is depending on it)

MaxTheWhale:
@robodoo r+

MaxTheWhale:
@robodoo r+
checksum is now fixed

PR #245237, opened 2026-01-22
[FIX] point_of_sale: fix report price for refunds

sbel-odoo:
When doing a refund the price is positive in the reporting.

Steps to reproduce:
-------------------
* Open pos session
* Make an order
* Refund the order
* Go backend and look at the order reporting
> Observation: if the product sold was 100$, the report shows
200$ in positive

Why the fix:
------------
orderSign exist only in the js side resulting in values to be positive. We know multiply the prices in the report by 1 or -1 if the order is a refund or not. This fix works because starting from 19.0 pos refund orders cannot have anything else than refund product.

opw-5483471

### DIFF
--- a/addons/point_of_sale/report/pos_order_report.py
+++ b/addons/point_of_sale/report/pos_order_report.py
@@ -66,10 +66,10 @@ class ReportPosOrder(models.Model):
                 l.id AS id,
                 1 AS nbr_lines, -- number of lines in order line is always 1
                 s.date_order AS date,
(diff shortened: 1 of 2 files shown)


sbel-odoo:
Note: it would also be possible to use the sign of the quantity. I just decided to follow the same form as here: https://github.com/odoo/odoo/commit/31c4a26b04f271c33ce2babb74afaebf574afaea

PR #216370, opened 2025-06-27
[FIX] html_editor, project: prevent error when saving after applying history

walidsahli:
Problem:
When applying a specific version from history in `task.description`, then updating and saving it, an error appears stating the content was saved from a different history model.

Cause:
When inserting versioned content, the required
`data-last-history-steps` attribute is not included.

Solution:
Ensure the latest `data-last-history-steps` is added when restoring a version.

Steps to reproduce:
- Open Project > any task
- Change the description
- Save
- Open version history and apply any version
- Update the description
- Save
- Error appears

opw-4829553

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245239, opened 2026-01-22
[FIX] web_editor: prevent error when saving after applying history

maedekey:
backport of https://github.com/odoo/odoo/pull/216370

__
opw-5405048

### DIFF
--- a/addons/web_editor/static/src/js/common/utils.js
+++ b/addons/web_editor/static/src/js/common/utils.js
@@ -582,3 +582,25 @@ export default {
     isImageCorsProtected: _isImageCorsProtected,
     isSrcCorsProtected: _isSrcCorsProtected,
 };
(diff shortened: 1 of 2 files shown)


addons/web_editor/static/src/js/common/utils.js
```diff
@@ -582,3 +582,25 @@ export default {
     isImageCorsProtected: _isImageCorsProtected,
     isSrcCorsProtected: _isSrcCorsProtected,
 };
+
+const HTML_FIELD_METADATA_ATTRIBUTES = ["data-last-history-steps"];
+export function getHtmlFieldMetadata(content) { //todo: mettre dans un fichier utils
```
walidsahli: can you remove this comment as you already did it.

PR #245240, opened 2026-01-22
[FIX] core: fix KeyError when upgrade with manual one2many fields

HydrionBurst:
When ``setup`` a manual one2many field, if its ``inverse_name`` field hasn't been ``setup`` and is also a manual field which might be ``pop`` when ``setup``, the one2many field can be ``setup`` successfully. But when computing ``setup_inverses`` when ``init_models``, the ``inverse_name`` will cause a ``KeyError``.
``invf = registry[self.comodel_name]._fields[self.inverse_name]``

Reproduce:
see https://github.com/odoo/odoo/pull/240085

This commit simply checks the ``setup`` for the inverse field of the one2many field.

Current behavior before PR:

---

### DIFF
--- a/odoo/orm/fields_relational.py
+++ b/odoo/orm/fields_relational.py
@@ -876,7 +876,8 @@ class One2many(_RelationalMulti):
             # link self to its inverse field and vice-versa
             comodel = model.env[self.comodel_name]
             try:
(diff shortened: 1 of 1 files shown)


rco-odoo (approved):
@robodoo r+

PR #245242, opened 2026-01-22
[IMP] Add optional fields to peppol invoices

yagho-odoo:
Currently, several specific UBL fields are lacking from our Peppol integration, resulting in users not being able to send compliant invoices Using studio, they can now add the optional fields that are allowed by us and their optional attributes
TODO: Add Test case
task-4963157


addons/account_edi_ubl_cii/tools/ubl_21_invoice.py
```diff
@@ -44,6 +44,8 @@
     'cbc:PricingCurrencyCode': {},
     'cbc:LineCountNumeric': {},
     'cbc:BuyerReference': {},
+    'cbc:TaxPointDate': {},
```
gawa-odoo: This one atleast is put at the wrong place for Bis3.
It fails the xsd validation, as it expects it above DocumentCurrencyCode.
You will need to check the others

That's the goal of this declaration, to define the order of the nodes

addons/account_edi_ubl_cii/tools/ubl_20_optional_fields.py
```diff
@@ -0,0 +1,71 @@
...
+        "path": ["cac:InvoicePeriod", "cbc:EndDate"],
+        "attrs": lambda invoice: {
+            '_text': invoice.x_studio_peppol_invoice_period_end_date,
+        }
+    },
+}
+
+PEPPOL_OPTIONAL_LINE_FIELDS = {
+    "x_studio_peppol_order_line_reference_id": {
+        "path": ["cac:OrderLineReference", "cbc:LineID"],
```
gawa-odoo, line 62: Don't you need to add these to the definition of `OrderLineReference`? (To define the order as well?)

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -1333,6 +1336,19 @@ def _add_invoice_monetary_total_nodes(self, document_node, vals):
             },
         })
 
+    def _add_invoice_optional_nodes(self, document_node, vals):
+        invoice = vals['invoice']
+        invoice_optional_fields = {key: invoice[key] for key in invoice._fields if key.startswith("x_studio_peppol") and invoice[key]}
+        for field in invoice_optional_fields:
+            path = PEPPOL_OPTIONAL_FIELDS[field]["path"]
```
gawa-odoo: You also need to check if the field is present in PEPPOL_OPTIONAL_FIELDS here, else you will launch a traceback if he did a typo

addons/account_edi_ubl_cii/tools/ubl_20_optional_fields.py
```diff
@@ -0,0 +1,71 @@
+PEPPOL_OPTIONAL_FIELDS = {
+    "x_studio_peppol_delivery_party_name": {
+        "path": ["cac:Delivery", "cac:DeliveryParty", "cac:PartyName", "cbc:Name"],
```
gawa-odoo: This one may be unneeded now, as https://github.com/odoo/odoo/commit/c6151558f228b3a1a9264024af3ef552f4a06a31 seem to have added it and got merged in the meantime

gawa-odoo (changes requested):
And I would appreciate you creating at least a test for it, to ensure the feature still works later. (Ideally, we would test them all, even in a single test).
You can directly create a field by hand without studio (and naming it `x_studio_peppol_contract_document_reference_id` for example)

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -1334,6 +1338,27 @@ def _add_invoice_monetary_total_nodes(self, document_node, vals):
...
+            self.add_invoice_optional_nodes(document_node, vals, PEPPOL_INVOICE_OPTIONAL_FIELDS)
+        elif (vals['document_type'] == 'credit_note'):
+            self.add_invoice_optional_nodes(document_node, vals, PEPPOL_CREDIT_NOTE_OPTIONAL_FIELDS)
+
+    def add_invoice_optional_nodes(self, document_node, vals, optional_fields):
+        move = vals['invoice']
+        invoice_optional_fields = {key: move[key] for key in move._fields if key.startswith("x_studio_peppol") and move[key]}
+        for field in invoice_optional_fields:
+            if field not in optional_fields:
+                raise ValidationError(f"Unexpected value: '{field}'. Please make sure you entered a valid field name.")
```
gawa-odoo: Should we raise here? 
I could imagine people having created fileds named "Peppol something" on their invoices for their own uses (their own customizations or the like) and I don't think we should block their sending for it.
I would just ignore that field and so, `continue` (or just replace the if - else into a `if field in optional_fields:`)

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -1334,6 +1337,26 @@ def _add_invoice_monetary_total_nodes(self, document_node, vals):
...
+        if (vals['document_type'] == 'invoice'):
+            self.add_invoice_optional_nodes(document_node, vals, PEPPOL_INVOICE_OPTIONAL_FIELDS)
+        elif (vals['document_type'] == 'credit_note'):
+            self.add_invoice_optional_nodes(document_node, vals, PEPPOL_CREDIT_NOTE_OPTIONAL_FIELDS)
+
+    def add_invoice_optional_nodes(self, document_node, vals, optional_fields):
+        move = vals['invoice']
+        invoice_optional_fields = {key: move[key] for key in move._fields if key.startswith("x_studio_peppol") and move[key]}
+        for field in invoice_optional_fields:
+            if field in optional_fields:
```
gawa-odoo: I think the check of optional_fields could be done in the dict comprehension directly, sorry.
Could be a tiny bit faster

Except that, LGTM

gawa-odoo (approved):

gawa-odoo:
@robodoo r+

PR #245243, opened 2026-01-22
[FIX] account: block deletion of used tax groups

smjo-odoo:
Before this PR:
- Group of taxes can be deleted by a user, even if they are used.

After this PR:
- If a user tries to delete a group of taxes in use, a validation
  error is raised.

- Fixed a test case in POS and deactivated the tax instead of
  deleting the tax.

Related PR: https://github.com/odoo/enterprise/pull/105174

task-5472834

### DIFF
--- a/addons/account/models/account_tax.py
+++ b/addons/account/models/account_tax.py
@@ -340,10 +340,9 @@ class AccountTax(models.Model):
                     WHERE EXISTS(
                         SELECT 1
                         FROM account_move_line_account_tax_rel AS line
-                        WHERE account_tax_id IN %s
-                        AND account_tax.id = line.account_tax_id
-                    ) """,
-                tuple(self.ids),
+                        WHERE account_tax.id = line.account_tax_id
+                    ) AND id IN %s
+                    """, tuple(self.ids),
             )))
             taxes_to_compute = set(self.ids) - used_taxes
 
@@ -356,10 +355,9 @@ class AccountTax(models.Model):
                         WHERE EXISTS(
                             SELECT 1
                             FROM account_reconcile_model_line_account_tax_rel AS reco
-                            WHERE account_tax_id IN %s
-                            AND account_tax.id = reco.account_tax_id
-                        ) """,
-                    tuple(taxes_to_compute)
+                            WHERE account_tax.id = reco.account_tax_id
+                        ) AND id IN %s
+                        """, tuple(taxes_to_compute)
                 )))
                 taxes_to_compute -= used_taxes
 
--- a/addons/account/models/account_tax.py
+++ b/addons/account/models/account_tax.py
@@ -5125,6 +5125,11 @@ class AccountTax(models.Model):
             return ''
         return html2plaintext(self.description)
 
+    @api.ondelete(at_uninstall=False)
+    def unlink_except_tax_used(self):
+        if any(self.mapped('is_used')):
+            raise ValidationError(self.env._("You cannot delete taxes that are currently in use. Consider archiving them instead."))
+
 
 class AccountTaxRepartitionLine(models.Model):
     _name = 'account.tax.repartition.line'
--- a/addons/account/tests/test_account_tax.py
+++ b/addons/account/tests/test_account_tax.py
@@ -309,6 +309,45 @@ class TestAccountTax(AccountTestInvoicingCommon):
             {'display_type': 'payment_term',    'tax_ids': [],          'balance': 100.0,   'account_id': self.company_data['default_account_receivable'].id},
         ])
 
+    def test_cannot_delete_group_tax_in_use(self):
+        """ Test that a group of taxes (parent tax) cannot be deleted when it's used. """
+
+        sales_10_perc = self.env['account.tax'].create({
+            'name': '10% Sales tax',
+            'amount': 10.0,
+            'amount_type': 'percent',
+            'type_tax_use': 'sale',
+        })
+        sales_5_perc = self.env['account.tax'].create({
+            'name': '5% Sales tax',
+            'amount': 5.0,
+            'amount_type': 'percent',
+            'type_tax_use': 'sale',
+        })
+        # Group of taxes
+        sales_15_perc = self.env['account.tax'].create({
+            'name': '15% Sales tax',
+            'amount': 15.0,
+            'amount_type': 'group',
+            'type_tax_use': 'sale',
+            'children_tax_ids': [Command.set([sales_10_perc.id, sales_5_perc.id])],
+        })
+        self.env['account.move'].create({
+            'move_type': 'out_invoice',
+            'date': '2025-01-01',
+            'invoice_line_ids': [
+                Command.create({
+                    'name': 'invoice_line',
+                    'quantity': 1.0,
+                    'price_unit': 100.0,
+                    'tax_ids': [Command.set(sales_15_perc.ids)],
+                }),
+            ],
+        })
+        self.assertTrue(sales_15_perc.is_used)
+        with self.assertRaisesRegex(ValidationError, "delete taxes that are currently in use"):
+            sales_15_perc.unlink()
+
     def test_negative_factor_percent(self):
         account_1 = self.company_data['default_account_tax_sale'].copy()
         with self.assertRaisesRegex(ValidationError, r"Invoice and credit note distribution should have a total factor \(\+\) equals to 100\."):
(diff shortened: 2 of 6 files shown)


addons/account/models/account_tax.py
```diff
@@ -5044,6 +5044,13 @@ def _get_description_plaintext(self):
             return ''
         return html2plaintext(self.description)
 
+    @api.ondelete(at_uninstall=False)
+    def unlink_except_tax_used(self):
+        used_taxes = self.filtered('is_used')
+        if used_taxes:
+            tax_names = ", ".join(used_taxes.mapped("display_name"))
+            raise UserError(self.env._("Cannot delete tax(es) '%s' because they are used in one or more journal entries.", tax_names))
```
zepa-odoo:
```suggestion
        if any(self.mapped('is_used')):
            raise ValidationError(self.env._("You cannot delete taxes that are currently in use."))
```
We don’t need to include the list of tax names here, as it can become very long. A generic error message should be sufficient.

zepa-odoo (changes requested):
Hello @smjo-odoo,
Thanks for the PR! 
I’ve made a few changes,could you please take a look?
Also Add some test-cases for it.

zepa-odoo (approved):

yosuanicolaus (approved):

addons/account/models/account_tax.py
```diff
@@ -5044,6 +5044,11 @@ def _get_description_plaintext(self):
             return ''
         return html2plaintext(self.description)
 
+    @api.ondelete(at_uninstall=False)
```
jco-odoo, line 5126: We have an ondelete="restrict" on tax_line_id.  Would not an ondelete="restrict" on tax_ids (on account_move_line) do the same trick?  Why this choice to pass through is_used?

jco-odoo: Would not this avoid the enterprise PR as well?

smjo-odoo: > Why this choice to pass through is_used?

@jco-odoo, the reason for using `is_used` is that the tax group might be used in different modules, like sale, PoS, purchase, etc.  We have the `_hook_compute_is_used` method that computes the IDs of taxes used in other modules and returns a set of IDs of taxes from that input set that are used in transactions. 

If we use `ondelete='restrict'` on tax_ids in `account.move.line`, It will only prevent the deletion in invoicing module and we would still be able to delete tax groups used in other modules.

If the `is_used` field can be used for preventing the modification of taxes. Why should we not use the same field to prevent the deletion of used tax groups ?

smjo-odoo: > Would not this avoid the enterprise PR as well?

The enterprise PR would still be needed because in that particular test case, the goal is to create a bill with "no tax", and not to delete/unlink the tax. This is the reason why we used `Command.clear()` instead of `unlink()` on the tax_ids.

smjo-odoo: Hello @jco-odoo  Do we have any update on this PR ?

jco-odoo: @smjo-odoo I had not checked that far and that is_used seems to be in handy for it.  I see that the ondelete='restrict' is also not on some other models, where we could have simply put it. (hr.expense,  purchase.order.line, sale.order.line)  But e.g. the method is not inherited for sale.order.line...  Also, such changes are huge in 18.3 for still solving a corner case (although it is not 18 stable), but it can be a good idea to enforce it this way. 

I see @aboo-odoo worked a lot on the is_used, so maybe he can already tell why it is not on sale order line or if it can be a good approach for it.

aboo-odoo: The initial idea behind `is_used` was to check if a tax was used to restrict/track modification of used taxes (restriction has been removed as per FP request). I don't recall, nor see any reason anywhere on why the `is_used` is not applied to SOs and SOLs - maybe this was said during a discussion with TSB but not written anywhere.

However, it seems that here you are making it so that any taxes (not just tax groups) cannot be deleted after it as been used. If this is the case, I'm pretty sure this should be further discussed and validated.

edit: `is_used` is quite heavy to compute but I think it's okay to use it to check whether or not a tax should or not be deleted (again - if it has been functionally agreed upon)

smjo-odoo: @jco-odoo  Should we continue with the is_used approach ?

smjo-odoo: > However, it seems that here you are making it so that any taxes (not just tax groups) cannot be deleted after it as been used. If this is the case, I'm pretty sure this should be further discussed and validated.

@aboo-odoo  Would it be fine to use context(force_delete=True) to allow a forced deletion?
This way, we could bypass the blocking error when intentionally force-deleting a group of taxes, while still keeping the restriction in normal cases.

addons/account/tests/test_account_tax.py
```diff
@@ -308,3 +308,42 @@ def test_tax_no_duplicate_in_repartition_line(self):
...
+                Command.create({
+                    'name': 'invoice_line',
+                    'quantity': 1.0,
+                    'price_unit': 100.0,
+                    'tax_ids': [Command.set(sales_15_perc.ids)],
+                }),
+            ],
+        })
+        self.assertTrue(sales_15_perc.is_used)
+        with self.assertRaises(ValidationError):
```
william-andre:
```suggestion
        with self.assertRaisesRegex(ValidationError, "delete taxes that are currently in use"):
```

addons/account/models/account_tax.py
william-andre: > edit: `is_used` is quite heavy to compute but I think it's okay to use it to check whether or not a tax should or not be deleted (again - if it has been functionally agreed upon)

Especially since the SQL logic is flawed.
```sql
SELECT id
FROM account_tax
WHERE EXISTS(
    SELECT 1
    FROM account_move_line_account_tax_rel AS line
    WHERE account_tax_id IN %s
    AND account_tax.id = line.account_tax_id
)
```

Should be 
```sql
SELECT id
FROM account_tax
WHERE EXISTS(
    SELECT 1
    FROM account_move_line_account_tax_rel AS line
    WHERE account_tax.id = line.account_tax_id
) AND id IN %s
```

Same for all the similar queries of course.

william-andre (commented):
Using `is_used` is fine for me; it makes sense to protect the taxes against edition and deletion the same way.

addons/account/models/account_tax.py
```diff
@@ -340,10 +340,9 @@ def _compute_is_used(self):
                     WHERE EXISTS(
                         SELECT 1
                         FROM account_move_line_account_tax_rel AS line
-                        WHERE account_tax_id IN %s
-                        AND account_tax.id = line.account_tax_id
-                    ) """,
-                tuple(self.ids),
+                        WHERE account_tax.id = line.account_tax_id
+                    ) AND id IN %s
+                    """, tuple(self.ids),
```
jco-odoo, line 345: The simplest would be to not use _is_used but only put this one in a separate method and call it. In principle, the tax is mainly important for accounting (in principle you would still have the tax tags but it is more difficult to track where they come from then) and if it gets deleted on expense, ... it is less of a problem.  (discussed with PO)  If we would keep the _is_used, we would need a very precise error message to tell where the user would need to go to delete the tax.  Doing this is also better in stable that the ondelete restrict as we would need an update of the module for that. (so only the tax_ids on account.move.line but use this query instead of the ondelete restrict)

addons/account/models/account_tax.py
```diff
@@ -5125,6 +5123,11 @@ def _get_description_plaintext(self):
             return ''
         return html2plaintext(self.description)
 
+    @api.ondelete(at_uninstall=False)
+    def unlink_except_tax_used(self):
+        if any(self.mapped('is_used')):
+            raise ValidationError(self.env._("You cannot delete taxes that are currently in use."))
```
william-andre: Suggest archiving instead

addons/account/models/account_tax.py
william-andre: You can simply add smart buttons on the tax to jump to the different related models.

In the meantime, this fix is alright; even if it could be improved.

jco-odoo: That would be a lot of buttons just to tell where it is used...

william-andre: You're right, but I don't think that it matters that much...

addons/account/models/account_tax.py
smjo-odoo: yes @william-andre , I have updated the validation error message. Thanks

PR #245244, opened 2026-01-22
[FIX] l10n_lk: removing . from tax id

legi-odoo:
**STEP TO REPRODUCE**
1. create a company, and set country to Sri Lanka.
2. create a 2nd company, and do the same.
3. There is a company inconsistencies error.

**CAUSE**
There is 2 taxes defined in `account.tax-lk.csv` with `.` in their id. This messes up with the function `company_xmlid()`: we end up loading the `account.tax` record of the 1st company when saving the 2nd company.

opw-5473952


vin-odoo (approved):
robodoo r+

Thank you for the fix :)
I wasn't aware this could cause issue; tho not surprised as well 🙈

PR #245245, opened 2026-01-22
[FIX] stock: prevent error when setting inventory quantity to 0

reka-odoo:
When a user sets the inventory quantity to 0 on a quant while
the product’s Inventory Location is unset, a traceback is raised.

Steps to reproduce the error:
- Install ``stock`` module with demo data
- Open ``Cabinet with Doors`` product
- In Inventory tab, unset Inventory Location > Open forecast report > click the On Hand quantity
- Select the quant > Actions > Set to 0

Traceback:
```py
ValueError: NotNullViolation('null value in column "location_dest_id"
of relation "stock_move" violates not-null constraint
```

https://github.com/odoo/odoo/blob/bc790e13ddf3ceacead40cc6ff8d27f1a5f5364d/addons/stock/models/stock_quant.py#L1005-L1016
When property_stock_inventory is unset,
the ``_get_inventory_move_values`` method assigns a NULL value to ``location_dest_id`` in ``move_vals``.
As a result, creating the stock move with a NULL ``location_dest_id`` leads to the above traceback.

sentry-7117991902

---

### DIFF
--- a/addons/stock/models/stock_quant.py
+++ b/addons/stock/models/stock_quant.py
@@ -997,22 +997,31 @@ class StockQuant(models.Model):
         # Consider the inventory_quantity as set => recompute the inventory_diff_quantity if needed
         self.inventory_quantity_set = True
         move_vals = []
+        default_loss_locations = {}
+        quants_with_missing_loss_locations = self.filtered(lambda quant: not quant.product_id.with_company(quant.company_id).property_stock_inventory)
+        if quants_with_missing_loss_locations:
+            for company in quants_with_missing_loss_locations.mapped('company_id'):
+                loss_location_id = self.env['ir.default'].with_company(company)._get_model_defaults(
+                    'product.template').get('property_stock_inventory')
+                default_loss_locations[company.id] = self.env['stock.location'].browse(loss_location_id)
         for quant in self:
             # if inventory applied from product's inverse_qty and the inventory_diff_quantity is 0,
             # we skip creating a move with 0 quantity.
             if quant.env.context.get('from_inverse_qty') and quant.product_uom_id.compare(quant.inventory_diff_quantity, 0) == 0:
                 continue
+            inventory_location = quant.product_id.with_company(quant.company_id).property_stock_inventory or\
+                default_loss_locations.get(quant.company_id.id)
             # Create and validate a move so that the quant matches its `inventory_quantity`.
             if quant.product_uom_id.compare(quant.inventory_diff_quantity, 0) > 0:
                 move_vals.append(
                     quant._get_inventory_move_values(quant.inventory_diff_quantity,
-                                                     quant.product_id.with_company(quant.company_id).property_stock_inventory,
+                                                     inventory_location,
                                                      quant.location_id, package_dest_id=quant.package_id))
             else:
                 move_vals.append(
                     quant._get_inventory_move_values(-quant.inventory_diff_quantity,
                                                      quant.location_id,
-                                                     quant.product_id.with_company(quant.company_id).property_stock_inventory,
+                                                     inventory_location,
                                                      package_id=quant.package_id))
         moves = self.env['stock.move'].with_context(inventory_mode=False).create(move_vals)
         moves.with_context(ignore_dest_packages=True)._action_done()
--- a/addons/stock/tests/test_quant.py
+++ b/addons/stock/tests/test_quant.py
@@ -1401,6 +1401,28 @@ class TestStockQuant(TestStockCommon):
             'lot_id': False,
         }])
 
+    def test_set_inventory_no_property_stock_inventory(self):
+        """
+        Test that quant inventory can be applied on products with no property_stock_inventory set.
+        """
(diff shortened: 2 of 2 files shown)


maan-odoo (approved):

reka-odoo:
Hello @ticodoo :wave: 
Can you please review it? :slightly_smiling_face: 
Thanks!

addons/stock/models/stock_quant.py
```diff
@@ -996,22 +996,28 @@ def _apply_inventory(self, date=None):
         # Consider the inventory_quantity as set => recompute the inventory_diff_quantity if needed
         self.inventory_quantity_set = True
         move_vals = []
+        loss_locations = dict(self.env['stock.location']._read_group(
+            domain=[('usage', '=', 'inventory')],
+            groupby=['company_id', 'id'],
+        ))
```
ticodoo: This is incorrect, you should be using the `ir.default` value for the `property_stock_inventory` for `product.template`. Also your solution will throw an error if there are more than one `inventory` stock locations set in the company

ticodoo (changes requested):

addons/stock/models/stock_quant.py
reka-odoo: Hello @ticodoo :wave: 
I have made the changes.
Can you have a look?
Thanks!

addons/stock/models/stock_quant.py
```diff
@@ -1001,17 +1001,20 @@ def _apply_inventory(self, date=None):
             # we skip creating a move with 0 quantity.
             if quant.env.context.get('from_inverse_qty') and quant.product_uom_id.compare(quant.inventory_diff_quantity, 0) == 0:
                 continue
+            loss_location_id = self.env['ir.default'].with_company(quant.company_id)._get_model_defaults('product.template').get('property_stock_inventory')
```
ticodoo: Every `_get_model_defaults` call does a query, you still need to minimize the performance issue this will cause (i.e. don't call it from within a `for` loop)

Since this is just a sentry error anyways though, maybe you can do a check if any products missing a `property_stock_inventory` and then call the `_get_model_defaults` and create a lookuip table to use within the for loop

reka-odoo: Hello :wave: 
I have made the requested changes.
Can you have a look?
Thanks!

addons/stock/models/stock_quant.py
```diff
@@ -996,22 +996,31 @@ def _apply_inventory(self, date=None):
         # Consider the inventory_quantity as set => recompute the inventory_diff_quantity if needed
         self.inventory_quantity_set = True
         move_vals = []
+        default_loss_locations = {}
+        quants_with_missing_loss_locations = self.filtered(lambda quant: not quant.product_id.property_stock_inventory)
```
ticodoo:
```suggestion
        quants_with_missing_loss_locations = self.filtered(lambda quant: not quant.product_id.with_company(quant.company_id).property_stock_inventory)
```

addons/stock/tests/test_quant.py
```diff
@@ -1401,6 +1401,21 @@ def test_set_on_hand_quantity_tracked_product(self):
...
+        quant = self.env['stock.quant'].create([{
+            'location_id': self.stock_location.id,
+            'product_id': self.productA.id,
+            'inventory_quantity': 10,
+        }])
+
+        self.productA.property_stock_inventory = False
+
+        quant.with_context(inventory_report_mode=True).action_set_inventory_quantity_zero()
+        self.assertEqual(quant.inventory_quantity, 0)
```
ticodoo, line 1424: you should maybe also check that the expected default location is used

addons/stock/tests/test_quant.py
```diff
@@ -1401,6 +1401,29 @@ def test_set_on_hand_quantity_tracked_product(self):
             'lot_id': False,
         }])
 
+    def test_set_inventory_quantity_zero(self):
+        """
+        test that the quant is set to 0 from forecasted report.
```
ticodoo: your docstring and test title should match the situation you're testing. E.g.: `test_set_inventory_no_property_stock_inventory`

""" Test that quant inventory can be applied on products with no `property_stock_inventory` set """

addons/stock/tests/test_quant.py
```diff
@@ -1401,6 +1401,29 @@ def test_set_on_hand_quantity_tracked_product(self):
...
+            'product_id': self.productA.id,
+            'inventory_quantity': 10,
+        }])
+
+        self.productA.property_stock_inventory = False
+
+        quant.with_context(inventory_report_mode=True).action_set_inventory_quantity_zero()
+        move_line = self.env['stock.move.line'].search([('product_id', '=', self.productA.id)])
+        loss_location_id = self.env['ir.default']._get_model_defaults('product.template').get('property_stock_inventory')
+        default_loss_location = self.env['stock.location'].browse(loss_location_id)
```
ticodoo: no need to `browse` it, just compare the id values

PR #245246, opened 2026-01-22
[FIX] sale: amount_undiscounted

odoo-pda:

Current behavior before PR:

---

### DIFF
--- a/addons/sale/models/sale_order.py
+++ b/addons/sale/models/sale_order.py
@@ -676,11 +676,16 @@ class SaleOrder(models.Model):
             )
 
     def _compute_amount_undiscounted(self):
+        AccountTax = self.env['account.tax']
         for order in self:
-            total = 0.0
+            subtotal = 0
             for line in order.order_line:
-                total += (line.price_subtotal * 100)/(100-line.discount) if line.discount != 100 else (line.price_unit * line.product_uom_qty)
-            order.amount_undiscounted = total
+                base_line = line._prepare_base_line_for_taxes_computation(discount=0)
+                # Exclude global discounts and down payments
+                if not base_line.get('special_type'):
+                    AccountTax._add_tax_details_in_base_line(base_line, order.company_id)
+                    subtotal += base_line['tax_details']['raw_total_excluded_currency']
+            order.amount_undiscounted = subtotal
 
     @api.depends('client_order_ref', 'origin', 'partner_id')
     def _compute_duplicated_order_ids(self):
--- a/addons/sale/tests/test_sale_prices.py
+++ b/addons/sale/tests/test_sale_prices.py
@@ -1050,9 +1050,9 @@ class TestSalePrices(SaleCommon):
         self.assertEqual(line.price_subtotal, 17527.41)
         self.assertEqual(line.untaxed_amount_to_invoice, line.price_subtotal)
 
-    def test_discount_and_amount_undiscounted(self):
-        """When adding a discount on a SO line, this test ensures that amount undiscounted is
-        consistent with the used tax"""
+    def test_amount_undiscounted_with_incl_excl_taxes(self):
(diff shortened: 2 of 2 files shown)


addons/sale/models/sale_order.py
```diff
@@ -675,11 +676,25 @@ def _compute_amount_paid(self):
...
+        AccountTax = self.env['account.tax']
         for order in self:
-            total = 0.0
+            subtotal = 0
             for line in order.order_line:
-                total += (line.price_subtotal * 100)/(100-line.discount) if line.discount != 100 else (line.price_unit * line.product_uom_qty)
-            order.amount_undiscounted = total
+                # Exclude discount lines
+                if line.price_subtotal <= 0:
+                    continue
```
Feyensv: You cannot reliably consider any negative line as a discount line.

addons/sale/models/sale_order.py
```diff
@@ -675,11 +676,25 @@ def _compute_amount_paid(self):
...
+                company = line.company_id or self.env.company
+                base_line = line._prepare_base_line_for_taxes_computation()
+                AccountTax._add_tax_details_in_base_line(base_line, company)
+                # Calculate the price_subtotal before discount
+                discount_factor = (
+                    (1 - base_line['discount'] / 100) if base_line['discount'] != 100 else 1
+                )
+                excl_tax_discounted_price = base_line['tax_details']['raw_total_excluded_currency']
+                excl_tax_original_price = excl_tax_discounted_price / discount_factor
+                subtotal += excl_tax_original_price
```
Feyensv: Way too detailed stuff from accounting being used here.
As I see it, it should look like `_compute_amounts` but simply with discount values removed from the base lines computation no? 🤔

addons/sale/models/sale_order.py
```diff
@@ -293,7 +293,8 @@ def _rec_names_search(self):
     # Remaining non stored computed fields (hide/make fields readonly, ...)
     amount_undiscounted = fields.Float(
         string="Amount Before Discount",
-        compute='_compute_amount_undiscounted', digits=0)
+        compute='_compute_amount_undiscounted',
+        digits="Product Price")
```
Feyensv: It's not a product price, it's a total amount of an order.
Should stay like it was (or become a monetary, but I wouldn't guarantee it's safe to do in stable)

Feyensv (commented):
I believe it's not safe and/or could be simpler 🤔

addons/sale/models/sale_order.py
odoo-pda: 👍  Switching for special_type in base_line.

addons/sale/models/sale_order.py
odoo-pda: `line._prepare_base_line_for_taxes_computation(discount=0)` is working 🤩

Feyensv: :partying_face:

addons/sale/models/sale_order.py
odoo-pda: Removed. Test adapted instead as we're in stable and the field will be removed in task-4075189.

addons/sale/tests/test_sale_prices.py
```diff
@@ -1096,11 +1096,44 @@ def test_discount_and_amount_undiscounted(self):
         line.discount = 50.0
         order.action_confirm()
 
-        # 300 with 10% incl tax -> 272.72 total tax excluded without discount
+        # 300 with 10% incl tax -> 272.727272... total tax excluded without discount
         # 136.36 price tax excluded with discount applied
-        self.assertEqual(order.amount_undiscounted, 272.72)
+        self.assertEqual(float_round(order.amount_undiscounted, 2), 272.73)
```
Feyensv: assertAlmostEqual

Feyensv (approved):
robodoo r+

PR #245247, opened 2026-01-22
[FIX] hr_attendance: Fix overtime calculation for shifts crossing UTC midnight

SurryaT10:
**Steps to reproduce:**
1. Configure Working Hours: Set up a night-shift schedule that splits at midnight (Local Time):
- Thursday: 20:00 to 24:00
- Friday: 00:00 to 04:00
2. Assign the above calendar to an employee.
3. Set the Employee’s Timezone to Asia/Kolkata (UTC+5:30).
4. Assign an active Overtime Ruleset to the employee.
5. When the attendance calendar is in Europe/Brussels TZ 
- Check-in: Jan 15, 15:30 CET
- Check-out: Jan 15, 23:30 CET

Expected Behavior: Worked Hours = 8.0, Overtime (Extra Hours) = 0.0 
Actual Behavior (Bug): Worked Hours = 8.0, Overtime = 4.0

**Bug Cause:**
1. The _update_overtime function normalized the Ruleset version periods using time.min for both the start and end of the day.
This forced the validity period of the rules to end exactly at 00:00:00 UTC on the final day.
2. The overtime recalculation logic failed to delete existing overtime records because the search domain was incorrectly computed.
Specifically, using relativedelta(SU) and relativedelta(MO(-1)) without the weekday= keyword argument did not shift the dates to the week boundaries.
This resulted in an empty or incorrect deletion range, leading to duplicated overtime hours as new records were layered on top of un-removed old ones.

**Solution:**
1. Modified the version_periods_by_employee mapping to use time.max (23:59:59) for the end of the version period.
This ensures that the ruleset remains active through the entire final calendar day in UTC, allowing shifts that cross the midnight boundary to be fully captured.
2. Corrected the date range logic by explicitly passing the weekday argument to relativedelta.
This ensures the domain correctly targets the full week window - from the preceding Monday to the following Sunday, ensuring all relevant stale overtime lines are purged before recalculation.
3. Updated Manual Edit Handling: Refined the logic to detect days with manual overrides or "To Approve" statuses before unlinking. If an attendance change triggers a recalculation on such a day, the system now replaces the manual entry with the mathematically correct value but flags the new record with a to_approve status for manager review.
4. Adjusted the expected overtime in test_weekly_overtime to 18.0 to correctly reflect the cumulative calculation of daily overtime (2h/day) plus the weekly overtime threshold reached on Friday.

Task: 5710273

---

### DIFF
--- a/addons/hr_attendance/models/hr_attendance.py
+++ b/addons/hr_attendance/models/hr_attendance.py
@@ -281,8 +281,8 @@ class HrAttendance(models.Model):
             return Domain.FALSE
         domain_list = [Domain.AND([
             Domain('employee_id', '=', employee.id),
-            Domain('date', '<=', max(attendances.mapped('check_out')).date() + relativedelta(SU)),
-            Domain('date', '>=', min(attendances.mapped('check_in')).date() + relativedelta(MO(-1))),
+            Domain('date', '<=', max(attendances.mapped('check_out')).date() + relativedelta(weekday=SU(1))),
+            Domain('date', '>=', min(attendances.mapped('check_in')).date() + relativedelta(weekday=MO(-1))),
         ]) for employee, attendances in self.filtered(lambda att: att.check_out).grouped('employee_id').items()]
         if not domain_list:
             return Domain.FALSE
@@ -291,7 +291,11 @@ class HrAttendance(models.Model):
     def _update_overtime(self, attendance_domain=None):
         if not attendance_domain:
             attendance_domain = self._get_overtimes_to_update_domain()
-        self.env['hr.attendance.overtime.line'].search(attendance_domain).unlink()
+        all_overtime_lines = self.env['hr.attendance.overtime.line'].search(attendance_domain)
+        manual_overtimes = set(all_overtime_lines.filtered(
+            lambda l: l.manual_duration != l.duration or l.status == 'to_approve'
+        ).mapped(lambda l: (l.employee_id.id, l.date)))
+        all_overtime_lines.unlink()
         all_attendances = (self | self.env['hr.attendance'].search(attendance_domain)).filtered_domain([('check_out', '!=', False)])
         if not all_attendances:
             return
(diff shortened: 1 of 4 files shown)


addons/hr_attendance/models/hr_attendance.py
```diff
@@ -276,7 +276,7 @@ def _get_overtimes_to_update_domain(self):
     def _update_overtime(self, attendance_domain=None):
         if not attendance_domain:
             attendance_domain = self._get_overtimes_to_update_domain()
-        self.env['hr.attendance.overtime.line'].search(attendance_domain).unlink()
+        self.env['hr.attendance.overtime.line'].search(attendance_domain).filtered(lambda l: l.duration == l.manual_duration).unlink()
```
mepe-odoo: I'm not sure about this; because you will add more extra hours if you don't delete them :/

addons/hr_attendance/tests/test_hr_attendance_rulesets.py
```diff
@@ -113,15 +113,16 @@ def test_no_overtime(self):
...
+                {
+                    'employee_id': self.employee.id,
                     'check_in': datetime(2021, 1, day, 8, 0),
                     'check_out': datetime(2021, 1, day, 18, 0)
-                }) for day in range(4, 9)  # Monday to Friday
+                } for day in range(4, 9)
             ]
-            self.assertAlmostEqual(self.employee.total_overtime, 10, 2, msg="He should work from 8-16h so each day he did 2 hours of overtime")
+            self.env['hr.attendance'].create(attendance_vals)
+            self.assertAlmostEqual(self.employee.total_overtime, 18, 2, msg="He should work from 8-16h so each day he did 2 hours of overtime")
```
mepe-odoo, line 125:
```suggestion
            self.assertAlmostEqual(self.employee.total_overtime, 20, 2, msg="He should work from 8-16h so each day he did 2 hours of overtime for daily rule and 10h for weekly rule")
```
For me you should have 20 hours and not 18 because you work 50h instead of 40h

mepe-odoo (changes requested):
good work :partying_face:

addons/hr_attendance/tests/test_hr_attendance_rulesets.py
SurryaT10: > For me you should have 20 hours and not 18 because you work 50h instead of 40h

It should be 18h right? because:
Mon to Thu - 2 + 2 + 2 + 2 = 8hr of daily overtime
Fri - 10hr of weekly overtime

So total_overtime = 18hr

mepe-odoo: but this employee will also work 10h on friday so he does 2h of daily overtime on friday too so it should be 20h of overtime

SurryaT10: @mepe-odoo but the 2hr on friday is already included as part of weekly rule right? arent we paying extra 2hr for the employee if we apply both rules in this case?

SurryaT10: There is a function _record_overlap_intervals in odoo/addons/hr_attendance/models/hr_attendance_overtime_rule.py which splits intervals based on rules. This prevents the extra 2hr double counting for friday.

mepe-odoo: ah yes you're right I forgot :D so you have 2hours where the two rules are applied ? :)

mepe-odoo (approved):
LGTM

addons/hr_attendance/tests/test_hr_attendance_rulesets.py
SurryaT10: @mepe-odoo Yes, exactly!

PR #245251, opened 2026-01-22
[FIX] point_of_sale: prevent rounding issues in cash difference removal

pedrambiria:
Before this commit, when calculating the cash difference in the report, the code did not account for currency rounding. This could lead to situations where a very small cash difference, due to rounding errors, was not recognized as zero, resulting in the unintended removal of cash moves.

opw-5489958

---


stva-odoo (commented):
robodoo r+

PR #238108, opened 2025-12-01
[FIX] website_sale: Maintain correct product image aspect ratio in grid

sben-odoo:
Steps to reproduce:
===================
1- Add a product with a very large image width & publish product.
2. Go to the Shop page & type product name. -> The product image is blurred.

Cause:
======
The product images have `h-100 w-100` classes which force them to fill the container dimensions exactly, ignoring their intrinsic aspect ratio.

Solution:
=========
Add the `object-fit-contain`  class to the image. This ensures the image scales to fit within the container while preserving its aspect ratio.

Side note: `object-fit-contain` class will be added only in version 17.0 In the next versions the class already exists.

opw-5258658

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245258, opened 2026-01-22
[FIX] website_sale: Auto crop of product images in product tiles

sben-odoo:
Steps to reproduce:
===================
1- Go to website > shop & open edit mode
2- Select any product and click on "paint-brush" icon 3- Click on "Image ratio" option and enable "Auto crop"

-> nothing happens to the images.

Cause:
======
After this commit [1] `object-fit-contain` was used even for auto crop which will override the `o_wsale_products_opt_thumb_cover` class object-fit value.

Solution:
=========
Apply basic fit contain only when autocrop is disabled.

[1]: https://github.com/odoo/odoo/pull/238108/commits/002a8a1ff65fd48fd69e41c9a881fddceee34bf5

opw-5868057

---

### DIFF
--- a/addons/website_sale/static/src/scss/product_tile.scss
+++ b/addons/website_sale/static/src/scss/product_tile.scss
@@ -113,7 +113,7 @@
         }
         .oe_product_image_img, .oe_product_image_img_secondary {
             position: absolute;
-            object-fit: var(--o-wsale-card-thumb-fill-mode, contain);
+            object-fit: var(--o-wsale-card-thumb-fill-mode, contain) !important;
             object-position: var(--o-wsale-card-thumb-position, center);
             background-size: var(--o-wsale-card-thumb-fill-mode, contain);
             background-position: var(--o-wsale-card-thumb-position, center);
--- a/addons/website_sale/views/product_tile_templates.xml
+++ b/addons/website_sale/views/product_tile_templates.xml
@@ -90,7 +90,7 @@
                 <!-- Primary image -->
                 <span
                     t-field="primary_image_holder.image_1920"
-                    t-options="{'widget': 'image', 'preview_image': 'image_1024', 'class': 'oe_product_image_img h-100 w-100 object-fit-contain'}"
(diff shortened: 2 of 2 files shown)


addons/website_sale/static/src/scss/product_tile.scss
```diff
@@ -119,6 +119,9 @@
             background-position: var(--o-wsale-card-thumb-position, center);
             background-repeat: var(--o-wsale-card-thumb-repeat, no-repeat);
         }
+        .auto_crop_images {
+            object-fit: var(--o-wsale-card-thumb-fill-mode, contain);
+        }
```
barracudapps:
```suggestion

        .auto_crop_images {
            object-fit: var(--o-wsale-card-thumb-fill-mode, contain);
        }

```

Maybe some spaces for readability ?

barracudapps (approved):
LGTM 👍 
Thanks for taking care of eCommerce's layout!

addons/website_sale/static/src/scss/product_tile.scss
livaios: This wont work on stable, users would need to upgrade the module and they wont necessarily know about the bug.
I propose a fix like this and in master we can implement your change.
```suggestion
        .oe_product_image_img.object-fit-contain {
            object-fit: var(--o-wsale-card-thumb-fill-mode, contain) !important;
        }
```

sben-odoo: Done

livaios (changes requested):
I appreciate your promptness in applying the needed changes.
I looked it over again and I think we can modify it directly here, by adding a `!important`
https://github.com/odoo/odoo/blob/707bde764139aedb29a77e896e59bbd9461ae988/addons/website_sale/static/src/scss/product_tile.scss#L116

We can remove `object-fit-contain` from the class in this PR, in case users install 19.0 they should have the correct layout.
And finally we want to remove `!important` from master since it won't be needed anymore.
What do you think?

livaios (approved):
robodoo r+

PR #245259, opened 2026-01-22
[FIX] stock: add filter warehouse to mobile stock report

josv-odoo:
On the mobile version the standard filter warehouses is not present.

Steps to reproduce:
-------------------
* Open odoo on a phone
* Inventory>Reporting>Stock
-> The filter Warehouses is not present

Observation:
------------
The basic warehouses filter is loaded in the stock report view :
https://github.com/odoo/odoo/blob/e99e07f2f22b0987468c45d3d7da287cdf703588/addons/stock/views/product_views.xml#L540
and defined in this file :
https://github.com/odoo/odoo/blob/bc46c9783ab00222b86cd951ff7f732b0212bdee/addons/stock/static/src/views/search/stock_report_search_panel.xml#L27-L31
Since the searchPanel is defined for .Regular views and we are in mobile, the searchPanel will not be mounted.

opw-5480119

### DIFF
--- a/addons/stock/static/src/views/search/stock_report_search_panel.js
+++ b/addons/stock/static/src/views/search/stock_report_search_panel.js
@@ -1,10 +1,11 @@
 import { SearchPanel } from "@web/search/search_panel/search_panel";
+import { useState } from "@odoo/owl";
 
 export class StockReportSearchPanel extends SearchPanel {
     static template = "stock.StockReportSearchPanel";
     setup() {
         super.setup(...arguments);
-        this.selectedWarehouse = false;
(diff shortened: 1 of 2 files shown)


lase-odoo (commented):
To be honest this looks more like a feature implementation rather than a fix. I do not think it is reasonable to implement this in stable. Also this is the only occurrence of an inheritance of the `web.SearchPanel.Small` view I can find in 18.0 which looks dubious to me particularly considering that the "category" filters is still part of the `StockReportSearchPanel` even in the mobile view (so that it is apparently added differently to the view).

I'll leave the review to the stock team If they really want to implement the feature but I personally suggest to re-target for stability reasons in any case.

PR #245261, opened 2026-01-22
[FIX] pos_hr: fix removing all employees from basic or minimal list

hadielyakhni:
Steps to reproduce
------------------
1. Enable "Log in with Employees"
2. Add employee "A" to the "Basic rights" group.
3. Add employee "B" to the "Minimal rights" group.
4. Make sure to remove all other employees from these 2 groups!!
5. Now save and go back to the main settings page, remove employee "A" from "Basic rights" group, such that this group is now empty.
6. Save the changes.

Observation -> Employee "A" is still in the group "Basic rights"!!

Why the issue
-------------
When deleting records from an `x2many` field in pos, from the main settings page, we do not pass the associated `unlink` commands in the arguments of `web_save` when clicking "Save", we just pass an empty commands list `[]`, because the backend, specifically the method `_preprocess_x2many_vals_from_settings_view` will take care of unlinking all the records before applying the new commands. So passing an empty commands list when updating an `x2many` field will be equivalent to saying "Unlink all records for that field, then apply no commands", i.e. just unlink all records for that field.

However, after https://github.com/odoo/odoo/commit/1f45b035354bb9040be3aba3a79392e603e870b4, we now filter out updates having an empty commands list, and thus now trying to unlink all records of a `x2many` does not work, since an empty command list will be dropped from the updates, before reaching `_preprocess_x2many_vals_from_settings_view`.

The fix
-------
We add stricter filtering logic, to keep valid command lists, even when
they are empty.

opw-5725879

### DIFF
--- a/addons/pos_hr/models/pos_config.py
+++ b/addons/pos_hr/models/pos_config.py
@@ -25,10 +25,11 @@ class PosConfig(models.Model):
 
         # write employees in sudo, because we have no access to these corecords
         sudo_vals = {
-            field_name: value
+            field_name: vals.pop(field_name)
             for field_name in ('minimal_employee_ids', 'basic_employee_ids', 'advanced_employee_ids')
             if not self.env.su
-            if (value := vals.pop(field_name, ()))
+            if isinstance(vals.get(field_name), list)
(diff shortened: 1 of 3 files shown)


stva-odoo (commented):
robodoo r+

PR #245272, opened 2026-01-22
[FIX] hr_work_entry_holidays: Error when writing contract with multiple leaves

kareemoo:
Issue: when you change in a contract for an employee with multiple leaves and one of them is overlapping multiple contracts, you get a validation error which shouldn't be the case

Steps to reproduce:
Case 1:
- create an employee with a contract ending on a day x
- create a leave for the employee from x-5 to x+5 and approve it
- create another contract with a different calendar for the employee starting on a day x+1 (change both the contract start date and the calendar and save once)
- when you save, you get a validation error but you shouldn't and the operation should be done normally

Case 2:
- create an employee with a contract ending on a day x
- create a leave for the employee from y-5 to y+5 inside the contract and approve it
- create another version with a different calendar for the employee starting on a day y (change both the contract start date and the calendar and save once)
- when you save, you get a validation error but you shouldn't and the operation should be done normally

Fix:
- ensured that the write of vals in contract is only done after refusing leaves that span multiple contracts
- changed the employee in `test_leave_change_working_schedule` and ensured that all timezones are in Europe/Brussels (jules_emp's tz was changing between UTC and Europe/Brussels depending on demo data which caused a runbot error)
- created another leave in `test_leave_multi_contracts_split` to ensure that the split works correctly with more than one leave

task-id: 5499527

---

### DIFF
--- a/addons/hr_holidays/models/hr_leave.py
+++ b/addons/hr_holidays/models/hr_leave.py
@@ -339,6 +339,10 @@ class HrLeave(models.Model):
 
     @api.depends('employee_id', 'request_date_from', 'request_date_to')
     def _compute_resource_calendar_id(self):
+        # Avoid recomputing the calendar in `_check_contracts`
+        # before the correct calendar is written to the version
+        if self.env.context.get('leave_skip_calendar_recompute', False):
+            return
         leaves_without_emp_or_date = self.filtered(
             lambda leave: not (leave.employee_id and leave.request_date_from and leave.request_date_to)
         )
--- a/addons/hr_holidays/models/hr_leave.py
+++ b/addons/hr_holidays/models/hr_leave.py
@@ -339,6 +339,11 @@ class HrLeave(models.Model):
 
     @api.depends('employee_id', 'request_date_from', 'request_date_to')
     def _compute_resource_calendar_id(self):
+        # skip the calendar compute if there is a new calendar being written to the leave
+        # but not written yet to the version as it might raise errors if there are overlapping
+        # leaves between this version and another one
+        if self.env.context.get('leave_skip_calendar_recompute', False):
+            return
         leaves_without_emp_or_date = self.filtered(
             lambda leave: not (leave.employee_id and leave.request_date_from and leave.request_date_to)
         )
@@ -378,14 +383,14 @@ class HrLeave(models.Model):
         self.ensure_one()
         domain = Domain.AND([
             Domain('employee_id', '=', self.employee_id.id),
-            Domain('contract_date_start', '<=', self.date_to),
+            Domain('date_start', '<=', self.date_to),
             Domain.OR([
-                Domain('contract_date_end', '>=', self.date_from),
-                Domain('contract_date_end', '=', False),
+                Domain('date_end', '>=', self.date_from),
+                Domain('date_end', '=', False),
             ])
         ])
         versions = self.env['hr.version'].sudo().search(domain)
-        return versions.filtered(lambda v: v._is_overlapping_period(self.date_from.date(), self.date_to.date()))
+        return versions
 
     @api.constrains('date_from', 'date_to')
     def _check_contracts(self):
--- a/addons/hr_holidays/models/hr_version.py
+++ b/addons/hr_holidays/models/hr_version.py
@@ -37,21 +37,15 @@ class HrVersion(models.Model):
                     created_versions |= super().create(vals)
                     continue
                 leaves = self._get_leaves_from_vals(vals)
-                is_created = False
                 for leave in leaves:
                     leaves_state = self._refuse_leave(leave, leaves_state)
-                    if not is_created:
-                        created_versions |= super().create([vals])
-                        is_created = True
+                created_versions |= super().create([vals])
+                for leave in leaves:
                     overlapping_contracts = self._check_overlapping_contract(leave)
                     if not overlapping_contracts:
                         continue
                     all_new_leave_origin, all_new_leave_vals = self._populate_all_new_leave_vals_from_split_leave(
                         all_new_leave_origin, all_new_leave_vals, overlapping_contracts, leave, leaves_state)
-                # TODO FIXME
-                # to keep creation order, not ideal but ok for now.
-                if not is_created:
-                    created_versions |= super().create([vals])
             if all_new_leave_vals:
                 self._create_all_new_leave(all_new_leave_origin, all_new_leave_vals)
         except ValidationError:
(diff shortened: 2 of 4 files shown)


addons/hr_holidays/models/hr_version.py
```diff
@@ -37,21 +37,15 @@ def create(self, vals_list):
...
                     continue
                 leaves = self._get_leaves_from_vals(vals)
-                is_created = False
                 for leave in leaves:
                     leaves_state = self._refuse_leave(leave, leaves_state)
-                    if not is_created:
-                        created_versions |= super().create([vals])
-                        is_created = True
+                created_versions |= super().create([vals])
+                for leave in leaves:
```
Einar-eian, line 43: Is there a reason why we need to split this `for leave in leaves...` loop into two loops? If I understand correctly it is to solve your main bug but is there a reason why we can't just create the versions in one batch after looping though the vals_list?

addons/hr_holidays/models/hr_version.py
```diff
@@ -78,11 +72,11 @@ def write(self, vals):
                         extra_domain=extra_domain
                     )
                     for leave in leaves:
-                        super(HrVersion, contract).write(vals)
-                        overlapping_contracts = self._check_overlapping_contract(leave)
+                        overlapping_contracts = contract.with_context(vals=vals)._check_overlapping_contract(leave)
                         if not overlapping_contracts:
                             continue
                         leaves_state = self._refuse_leave(leave, leaves_state)
+                        super(HrVersion, contract).write(vals)
```
Einar-eian, line 79: ~~Also same thing here as I mentioned in the other review. If overlapping_contracts is somehow always False, won't the super().write() call be ignored/never executed?~~
Never mind, I noticed the return value of the method should take care of it.

Einar-eian (commented):
Thank you for your work! 😄 I left one comment but looks good otherwise

addons/hr_holidays/models/hr_version.py
kareemoo: We need to refuse leaves first to avoid getting any error when creating the versions and then, in the 2nd loop, `_check_overlapping_contract` is used which expects versions to be already created. So, we can actually create the versions at the end and change the implementation of `_check_overlapping_contract` but it seems like an overkill to me

addons/hr_holidays/models/hr_leave.py
```diff
@@ -378,14 +383,14 @@ def _get_overlapping_contracts(self):
...
+            Domain('date_start', '<=', self.date_to),
             Domain.OR([
-                Domain('contract_date_end', '>=', self.date_from),
-                Domain('contract_date_end', '=', False),
+                Domain('date_end', '>=', self.date_from),
+                Domain('date_end', '=', False),
             ])
         ])
         versions = self.env['hr.version'].sudo().search(domain)
-        return versions.filtered(lambda v: v._is_overlapping_period(self.date_from.date(), self.date_to.date()))
```
Einar-eian, line 388: I'm not sure I understand the intent of this diff. Aren't you essentially checking now for overlapping versions instead of overlapping contracts? I can see an argument for it making sense, but it does sound a bit confusing when everything is named `*_contracts` but it's actually all about the versions and not the contracts fields on the version. I guess this diff can now also include versions with no contract which may or may not be desirable?

addons/hr_holidays/models/hr_version.py
```diff
@@ -74,15 +68,15 @@ def write(self, vals):
...
                     )
                     for leave in leaves:
-                        super(HrVersion, contract).write(vals)
-                        overlapping_contracts = self._check_overlapping_contract(leave)
-                        if not overlapping_contracts:
+                        overlapping_contracts = contract.with_context(vals=vals)._check_overlapping_contract(leave)
+                        if not overlapping_contracts or len(overlapping_contracts) <= 1:
                             continue
                         leaves_state = self._refuse_leave(leave, leaves_state)
+                        super(HrVersion, contract).write(vals)
```
Einar-eian, line 79: I do wonder if this write can somehow be hoisted out of this leave loop? but then again maybe the chances of multiple leaves in this loop triggering this write is pretty low?

addons/hr_holidays/models/hr_version.py
```diff
@@ -144,8 +142,8 @@ def _refuse_leave(self, leave, leaves_state):
 
     def _populate_all_new_leave_vals_from_split_leave(self, all_new_leave_origin, all_new_leave_vals, overlapping_contracts, leave, leaves_state):
         for overlapping_contract in overlapping_contracts:
-            new_request_date_from = max(leave.request_date_from, overlapping_contract.contract_date_start)
-            new_request_date_to = min(leave.request_date_to, overlapping_contract.contract_date_end or date.max)
+            new_request_date_from = max(leave.request_date_from, overlapping_contract.date_start)
+            new_request_date_to = min(leave.request_date_to, overlapping_contract.date_end or date.max)
```
Einar-eian, line 146: I guess same comment applies here, seems like we are now more focused on overlapping versions rather than contracts.

kareemoo: We need to do this if we are splitting a leave between two versions in the same contract with different calendars.

addons/hr_holidays/models/hr_version.py
kareemoo: There can be only 1 leave between 2 versions as leaves don't overlap. The only case that I think it might write twice if we are creating a version between 2 existing versions with leaves overlapping which shouldn't happen

addons/hr_holidays/models/hr_leave.py
kareemoo: I think we can try to keep it with contracts for now

addons/hr_holidays/models/hr_version.py
```diff
@@ -74,18 +68,18 @@ def write(self, vals):
...
-                        overlapping_contracts = self._check_overlapping_contract(leave)
+                        overlapping_contracts = contract.with_context(vals=vals)._check_overlapping_contract(leave)
                         if not overlapping_contracts:
                             continue
                         leaves_state = self._refuse_leave(leave, leaves_state)
+                        super(HrVersion, contract).write(vals)
                         specific_contracts += contract
                         all_new_leave_origin, all_new_leave_vals = self._populate_all_new_leave_vals_from_split_leave(
-                            all_new_leave_origin, all_new_leave_vals, overlapping_contracts, leave, leaves_state)
+                            all_new_leave_origin, all_new_leave_vals, overlapping_contracts + contract, leave, leaves_state)
```
Einar-eian: Small detail maybe:
It might possibly be smarter to use union / `|` here over `+` in case `contract` is also in `overlapping_contracts` which I think could theoretically happen.
```suggestion
                            all_new_leave_origin, all_new_leave_vals, overlapping_contracts | contract, leave, leaves_state)
```

addons/hr_holidays/models/hr_version.py
Einar-eian: Indeed, makes sense!

addons/hr_holidays/models/hr_version.py
Einar-eian: ah yes, makes sense

addons/hr_work_entry_holidays/tests/test_performance.py
Einar-eian, line 1: I think the diff in this file is not needed anymore, now that you are not checking on `date_start` in `_get_overlapping_contracts` (at least from what I see in my local testing but runbot might behave differently)

addons/hr_holidays/models/hr_version.py
```diff
@@ -74,18 +68,18 @@ def write(self, vals):
...
                     resource_calendar_id = vals.get('resource_calendar_id', contract.resource_calendar_id.id)
                     extra_domain = [('resource_calendar_id', '!=', resource_calendar_id)] if resource_calendar_id else None
-                    leaves = contract._get_leaves(
+                    leaves = contract.with_context(contract_date_start=vals.get('contract_date_start', False))._get_leaves(
                         extra_domain=extra_domain
                     )
                     for leave in leaves:
-                        super(HrVersion, contract).write(vals)
-                        overlapping_contracts = self._check_overlapping_contract(leave)
+                        overlapping_contracts = contract.with_context(vals=vals)._check_overlapping_contract(leave)
```
tivisse: Please be more specific, debugging context is already a nightmare :)

```suggestion
                        overlapping_contracts = contract.with_context(version_write_vals=vals)._check_overlapping_contract(leave)
```

tivisse:
@robodoo fw=skipmerge

PR #245278, opened 2026-01-22
[IMP] mail: mention online members with "here"

Atovange:
This commit introduces a new mention suggetion "(at)here", which notifies all non-offline members of the conversation.

task-5868541

### DIFF
--- a/addons/mail/models/discuss/discuss_channel.py
+++ b/addons/mail/models/discuss/discuss_channel.py
@@ -1067,8 +1067,16 @@ class DiscussChannel(models.Model):
         if message_type not in ["notification", "user_notification"]:
             # sudo: discuss.channel - write to discuss.channel is not accessible for most users
             self.sudo().last_interest_dt = fields.Datetime.now()
-        if "everyone" in kwargs.pop("special_mentions", []):
-            partner_ids = list(OrderedSet((partner_ids or []) + self.channel_member_ids.partner_id.ids))
+        if special_mentions := kwargs.pop("special_mentions", []):
+            partners = self.env['res.partner'].browse(partner_ids or [])
+            partner_members = self.channel_member_ids.partner_id
+            if "everyone" in special_mentions:
+                partners |= partner_members
+            if "here" in special_mentions:
+                partners |= partner_members.filtered(
+                    lambda p: any(u.im_status != "offline" for u in p.user_ids),
+                )
+            partner_ids = partners.ids
         if partner_ids:
             kwargs["partner_ids"] = self._get_allowed_message_partner_ids(partner_ids)
         # mail_post_autofollow=False is necessary to prevent adding followers
--- a/addons/mail/static/src/core/common/store_service.js
+++ b/addons/mail/static/src/core/common/store_service.js
@@ -97,7 +97,14 @@ export class Store extends BaseStore {
             label: "everyone",
             channel_types: ["channel", "group"],
             displayName: "Everyone",
-            description: _t("Notify everyone"),
+            description: _t("Notify all members of this conversation"),
+        },
+        {
+            isSpecial: true,
+            label: "here",
+            channel_types: ["channel", "group"],
+            displayName: "Here",
+            description: _t("Notify all members of this conversation who are online"),
         },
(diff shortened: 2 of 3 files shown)


addons/mail/static/src/core/common/store_service.js
```diff
@@ -83,7 +83,14 @@ export class Store extends BaseStore {
...
             displayName: "Everyone",
-            description: _t("Notify everyone"),
+            description: _t("Notify all members of this conversation"),
+        },
+        {
+            isSpecial: true,
+            label: "here",
+            channel_types: ["channel", "group"],
+            displayName: "Here",
+            description: _t("Notify all members of this conversation who are online"),
```
alexkuhn, line 107: Already problem with "Everyone" but the displayName should be `_t()`.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -1004,8 +1004,14 @@ def message_post(self, *, message_type="notification", partner_ids=None, **kwarg
...
-        if "everyone" in kwargs.pop("special_mentions", []):
-            partner_ids = list(OrderedSet((partner_ids or []) + self.channel_member_ids.partner_id.ids))
+        if special_mentions := kwargs.pop("special_mentions", []):
+            partners = self.env['res.partner'].browse(partner_ids or [])
+            partner_members = self.channel_member_ids.partner_id
+            if "everyone" in special_mentions:
+                partners |= partner_members
+            if "here" in special_mentions:
+                partners |= partner_members.filtered(lambda p: "offline" not in p.im_status)
+            partner_ids = partners.ids
```
alexkuhn, line 1079: I'm not entirely sure about people that have manual "offline" but are actually online.
In a sense I think that they should receive the notification, but at the same time other people should not see these manual offline people in the recipient list.

addons/mail/static/src/core/common/store_service.js
Atovange: IIRC we decided explicitly not to translate it, as it would be weird to type "ever" and have the translation of "Everyone" show up.

addons/mail/models/discuss/discuss_channel.py
Atovange: Indeed leaking the "fake" offline members would not be great.
I think we could avoid sending the `partner_ids` of a message, but I am not sure.

addons/mail/static/src/core/common/store_service.js
alexkuhn: Oh ok that's the display name, for a moment I thought we were showing "label" on UI.
Then ok at it is, though it looks weird to capitalize a command now I think about it... Let's ignore for now 😅

addons/mail/models/discuss/discuss_channel.py
seb-odoo: We need `partner_ids` to know who should be highlighted as mentioned.

We can't send it individually to each person as this would not scale.

It's probably easier to exclude fake offline from `@here`.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -1067,8 +1067,14 @@ def message_post(self, *, message_type="notification", partner_ids=None, **kwarg
...
             self.sudo().last_interest_dt = fields.Datetime.now()
-        if "everyone" in kwargs.pop("special_mentions", []):
-            partner_ids = list(OrderedSet((partner_ids or []) + self.channel_member_ids.partner_id.ids))
+        if special_mentions := kwargs.pop("special_mentions", []):
+            partners = self.env['res.partner'].browse(partner_ids or [])
+            partner_members = self.channel_member_ids.partner_id
+            if "everyone" in special_mentions:
+                partners |= partner_members
+            if "here" in special_mentions:
+                partners |= partner_members.filtered(lambda p: p.im_status != "offline")
```
alexkuhn: This crashes with `AttributeError: 'res.partner' object has no attribute 'im_status'`.
Should rely on `res.users` instead of `res.partner`.

Would be nice a small test coverage for this feature.

PR #245279, opened 2026-01-22
[IMP] product_combo : change free into include

mobt-odoo:

combo choice products are not free products. They are included in the combo product price. Therefore it creates misunderstanding for our users

Current behavior before PR:

Change the "free" string into "item" in the backend
and into "included" in pos frontend to align with the display in self

---


adgu-odoo (commented):
robodoo r+

PR #245280, opened 2026-01-22
[PERF] stock, mrp: speed up producing SN tracked products

Aurelienvd:
When a product is tracked by unique serial number, is set as MTO/Manufacture and has a bom producing 1 quantity with multiple operations, it becomes
increasingly slow to produce it as the quantity to produce and the number of operations grow. This is linked to multiple places in the codebase that are mostly design to work on singletons.

As the product is tracked by serial_number, there will be 1 mrp.production for each SN. As such, batching calls in the codebase leads to an overall speed improvement.

The first optimization is in `_compute_move_finished_ids`. Batching the updates on stock.moves reduces the number of calls to `__setitem__` and makes the calls to `modified` works with bigger recordsets.

Another optimization is to add to possibility to set `qty_producing` on multiple mrp.productions at once. Also, a new method called `_set_quantities_done` is added on stock.move. It sets the quantity on multiple moves at once. The point of this method is that the previous `_set_quantity_done` method might lead to the creation/deletion of existing move_lines. As it was only working for a single quantity, those calls to create/unlink were not properly batched. `_set_quantities_done` changes that by calling create/delete only once for all the moves and all the quantities in parameters.

The last optimization is in `_split_productions`. Because the product is tracked by serial number, the total number of production in self + the number of backorders = the quantity of product to produce. As the number of operations on the BOM increases so does the total number of workorder_ids. This can lead to a huge number of workorder_ids and writing on `duration_expected` becomes increasingly slow. One of the reason for that is because writing on this field will trigger a call to `modified` and a call to `_compute_duration_expected` among others.

However, since the BOM is the same for all the productions, it's safe to assume that a lot of workorders will have the same `duration_expected`. We are therefore batching the updates by the value of `duration_expected`. This leads to a further speedup.

#### speedup

A Test database with a product tracked by SN, MTO/Manufacture, a BOM with qty 1.0, 3 components and 15 operations and the Warehouse manufacturing in 2 steps

| Producing Qty | Before PR  | After PR  |
|:-------------:|:----------:|:---------:|
|       1       |    100ms   |    100ms  |
|       10      |    1.7s    |    1.1s   |
|       150     |    35s     |    17s    |
|       1000    |    7min35s |    2min25s|

Test v18 database for 150 quantities, 5 operations vs 15 operations
|   Operations | Before PR | After PR |
|:------------:|:---------:|:--------:|
|        5     |    12s    |    7.4s  |
|       15     |    35s    |    17s   |

We can see that the speedup after batching is 2 on average. Both when increasing the qty to produce and the number of operations, the version After PR scales better.

opw-5184826

---

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -453,8 +453,16 @@ class MrpProduction(models.Model):
 
     @api.depends('workorder_ids.duration_expected')
     def _compute_duration_expected(self):
-        for production in self:
-            production.duration_expected = sum(production.workorder_ids.mapped('duration_expected'))
+        res = self.env['mrp.workorder']._read_group(
+            [('production_id', 'in', self.ids)],
+            ['production_id'],
+            ['duration_expected:sum'],
+        )
+        productions_with_wo = self.env['mrp.production']
+        for production, total_duration_expected in res:
+            production.duration_expected = total_duration_expected
+            productions_with_wo |= production
+        (self - productions_with_wo).duration_expected = 0.0
 
     @api.depends('workorder_ids.duration')
     def _compute_duration(self):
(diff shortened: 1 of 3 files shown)


addons/mrp/models/mrp_production.py
```diff
@@ -2817,29 +2842,28 @@ def _prepare_merge_orig_links(self):
         return origs
 
     def _set_quantities(self):
-        self.ensure_one()
```
ticodoo, line 2825: I'm a little dubious about this change in stable since customizations might extend this method and add logic that's designed for it to always be `ensure_one` 🤔 (i.e. it might break a lot of stuff? 🤷 )

Aurelienvd: @ticodoo Sorry I completely forgot about this PR :face_with_peeking_eye: 
The issue is that if I keep the `ensure_one` then I cannot call this method in batch in `pre_button_mark_done` so `self._set_qty_producing` will be done one production at a time and the move_lines will be created/updated for every single production. I fear that it may greatly impact the speed of this patch :/

I can do a benchmark for this but I think we should get similar result as the "Before PR" numbers.

For the customization, I don't see a lot of cases where this would break stuff since there's no return value :thinking: The only thing I can think of is if an override explicitly catch the singleton `ValueError`.

Maybe a possible solution would be to add a batch version of this method, like `_set_quantities_for_productions` ? So we can keep this one with `ensure_one` ? I know Arnold is not super fan of this but I don't really see a lot of ways to go around that. Or we can also close this PR if you think this is too risky :slightly_smiling_face:

PR #245281, opened 2026-01-22
[FIX] sale_stock,purchase_stock: ministock pickings

Sehnde:
Make sure that the quantity received is in the unit of measure of the purchase order line.
Also, when installing stock, create pickings for partial and empty sale/purchase orders.
Finally, since creating we're creating more pickings, we need to raise the demo stock
of product_product_12 to not be in negative stock for other modules demo data.

task 5431550

---

### DIFF
--- a/addons/purchase_stock/__init__.py
+++ b/addons/purchase_stock/__init__.py
@@ -5,6 +5,34 @@ from . import models
 from . import report
 from . import wizard
 
+from odoo.fields import Domain
+
+
+def _post_init_hook(env):
+    _create_buy_rules(env)
+    _create_pickings_for_open_purchase_orders(env)
+
+
+def _split_partial_purchase_order_lines(env):
+    """ Pre-init hook: before installing stock, split partial order lines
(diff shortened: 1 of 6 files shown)


addons/purchase_stock/__init__.py
```diff
@@ -13,3 +20,37 @@ def _create_buy_rules(env):
...
+    for po in open_purchase_orders:
+        if po.receipt_status == 'partial':
+            partial_orders |= po
+        for line in po.order_line:
+            if 0 < line.qty_received < line.product_qty:
+                new_line = line.copy_data({
+                    'product_qty': line.product_qty - line.qty_received,
+                })
+                lines_to_create.extend(new_line)
+                line.with_context(skip_picking=True).product_qty = line.qty_received
```
Whenrow, line 47: This could be done in a `_pre_init_hook` to avoid dealing with `stock` overrides. This method could then just be creating stock moves for lines having their `qty_received == 0`

Whenrow:
robodoo r+
thank you

PR #245282, opened 2026-01-22
[FIX] web,*: fix favorite icon size in form titles

xlu-odoo:
*: product, project

Before this commit, product and project each had a duplicated CSS rule to make the favorite icon inherit the title font-size. This caused inconsistent behavior: modules like hr_recruitment that use the favorite widget in titles would show a smaller icon unless product or project were installed.

This commit moves the rule to the form_controller.scss in web, ensuring  consistent behavior across all modules regardless of which ones are installed.

task-5379834

---

### DIFF
--- a/addons/project/static/src/scss/project_form.scss
+++ b/addons/project/static/src/scss/project_form.scss
@@ -36,12 +36,8 @@
     }
 }
 
-.oe_title .o_favorite i.fa {
-    font-size: inherit;
-}
-
 .o_data_cell .o_favorite i.fa {
-    font-size: 1.35em;
+    --Favorite-font-size: 1.35em;
 }
 
 .o_project_task_kanban_view .o_kanban_quick_create .o_field_widget .o_input  {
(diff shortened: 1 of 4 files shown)


addons/project/static/src/scss/project_form.scss
```diff
@@ -36,10 +36,6 @@
     }
 }
 
-.oe_title .o_favorite i.fa {
-    font-size: inherit;
-}
-
 .o_data_cell .o_favorite i.fa {
     font-size: 1.35em;
```
chgo-odoo: Since we're removing these custom rules, I'd suggest to assign `--Favorite-font-size` here instead of defining the `font-size` property 🤓

addons/web/static/src/views/fields/boolean_favorite/boolean_favorite_field.scss
```diff
@@ -11,3 +11,7 @@
         }
     }
 }
+
+.oe_title .o_favorite i.fa {
+    font-size: inherit;
+}
```
chgo-odoo: This should be placed within `form_controller.scss` instead to keep this SCSS file clean (don't forget to update commit and PR description as you reference the file in it ⚠️) 

I'd suggest to put it here as we already have the same rule for the priority widget : https://github.com/odoo/odoo/blob/master/addons/web/static/src/views/form/form_controller.scss#L349

chgo-odoo (approved):
Hello @xlu-odoo 👋 

LGTM beside these two tiny comments

PR #243028, opened 2026-01-09
[FIX] *: propagate stock move partner from procurement

lase-odoo:
*mrp{,_subcontracting_dropshipping}, point_of_sale, sale_purchase, {purchase_,sale_,}stock

### Issue: 

The partner_id (contact field) of stock move is no longer propagated in pull chain since the procurement/reference refactoring of 19.0 see 2713876dbc70d3984e584a9037a2206dcda4e84a

#### Steps to reproduce:
- In the settings enable Multi-Steps routes
- Inventory > Configuration > Warehouse Management > Warehouses
- Set your warehouse in two step deliveries
- Modify the routes to have two pull rules: Stock -> Out -> Customer
- Create and confirm an SO, 1 unit of a storable product for a customer
### Cause of the issue:

Due to the configuration, the ship move is created by the `action_launch_stock_rule` and linked to the customer via the procurement created from the sale order line:
https://github.com/odoo/odoo/blob/4466a2684fd36ef3c9e1e7f7437ae48ff51f1941/addons/sale_stock/models/sale_order_line.py#L295 Then, the confirmation of the ship move generates a second procurement for the pick move creation. Prior to 19.0, the partner_id of this second procurement was set to the customer via the procurement group: https://github.com/odoo/odoo/blob/712cd806c38e193acdb966e4d3b339ab780f158e/addons/stock/models/stock_rule.py#L333 However, since procurement groups have been removed by the reference refactor 2713876dbc70d3984e584a9037a2206dcda4e84a, the partner id is now expected to be set from the procurement values:
https://github.com/odoo/odoo/blob/4466a2684fd36ef3c9e1e7f7437ae48ff51f1941/addons/stock/models/stock_rule.py#L332 As it is not set by any mean in this procurement values it will therefore not be propagated.

### Additional changes:

1. The `partner_id` of the procurement values needs to be harmonized are some methods set a recordset and others a record id:
https://github.com/odoo/odoo/blob/4466a2684fd36ef3c9e1e7f7437ae48ff51f1941/addons/point_of_sale/models/pos_order.py#L1707
https://github.com/odoo/odoo/blob/4466a2684fd36ef3c9e1e7f7437ae48ff51f1941/addons/mrp_subcontracting_dropshipping/models/stock_orderpoint.py#L10-L13
https://github.com/odoo/odoo/blob/4466a2684fd36ef3c9e1e7f7437ae48ff51f1941/addons/sale_stock/models/sale_order_line.py#L295
[...]

PR #245284, opened 2026-01-22
[FIX] sale_purchase,stock_dropshipping: set dest_address_id when changing picking_type

qucol-odoo:
## Issue
When setting up a product with both the MTO and the *Dropship* routes, the *Purchase Order* genereated when confirming a *Sales Order* does not contain a *Dropship Address* (`purchase.order.dest_address_id`). It is problematic because that field is both readonly and required to confirm the order.

## Steps to reproduce
1. Install *Stock* (`stock`), *Purchase* (`purchase`) and *Sales* (`sale_management`)
2. In Settings, enable *Dropshipping* and *Replenish on Order (MTO)*
3. Create a Product P
    - Set a vendor in the Purchase tab
    - Enable the *Buy*, *Dropship* and *Replenish on Order (MTO)* routes
4. Create a Sales Order
    - Any Customer
    - Product P
    - Confirm the Sales Order
5. Click on the *Purchase* smart button
6. Set the *Delivery To* (`purchase.order.picking_type_id`) field to *"Dropship"*
7. **The _Dropship Address_ (`purchase.order.dest_address_id`) field appears, but it's empty and readonly. The purchase order cannot be confirmed, as the field is required and cannot be updated.**

## Cause
When confirming a Sales Order, the created Purchase Order has a `dest_addres_id` set by `StockRule._prepare_purchase_order`:

https://github.com/odoo/odoo/blob/19.0/addons/purchase_stock/models/stock_rule.py#L350

At that point, the `picking_type_id` of the PO is set to `"Receipts"`, which `default_location_dest_id` is the user's Stock, and the `usage` of that location is set to `"internal"`. When `_compute_dest_address_id` is triggered, it starts by calling the method in `sale_purchase`:

https://github.com/odoo/odoo/blob/9d96a8a4ae23bd331296ee0fd628c2be3de4bfe3/addons/sale_purchase/models/purchase_order.py#L25-L30

Which calls the one in `purchase_stock`:

https://github.com/odoo/odoo/blob/9d96a8a4ae23bd331296ee0fd628c2be3de4bfe3/addons/purchase_stock/models/purchase_order.py#L80-L82

Which sets the `dest_address_id` to `False`. This impacts the rest of first `_compute_dest_address_id`, as the PO does not have a `dest_address_id` anymore, its value will never be updated by the `_compute_dest_address_id` methods.

## Fix
The `dest_address_id` should only be set when dropshipping. The easiest way to do so is to override the `_compute_dest_address_id` in the `stock_dropshipping` module by following a similar logic as in `sale_purchase`:

https://github.com/odoo/odoo/blob/7a39185f83d0daca207c8007512f4700537c7e88/addons/sale_purchase/models/purchase_order.py#L25-L30

opw-5426322

### DIFF
--- a/addons/stock_dropshipping/models/purchase.py
+++ b/addons/stock_dropshipping/models/purchase.py
@@ -35,6 +35,9 @@ class PurchaseOrder(models.Model):
         self.ensure_one()
         return self.picking_type_id and self.picking_type_id.code == 'dropship'
 
+    def _should_set_dest_address(self):
+        return super()._should_set_dest_address() or self._is_dropshipped()
+
 
 class PurchaseOrderLine(models.Model):
     _inherit = 'purchase.order.line'
--- a/addons/stock_dropshipping/tests/test_dropship.py
+++ b/addons/stock_dropshipping/tests/test_dropship.py
@@ -575,3 +575,28 @@ class TestDropshipPostInstall(common.TransactionCase):
             default_product_tmpl_id=self.dropship_product.product_tmpl_id.id
         ))
         self.assertNotEqual(replenish_wizard.route_id, self.env.ref('stock_dropshipping.route_drop_shipping'))
+
+    def test_dest_address_when_changing_po_to_dropship(self):
+        """
+        Check that the destination address is set on the purchase order when
+        its picking type is manually changed to the dropshipping picking type.
+        """
+        mto_route = self.env.ref('stock.route_warehouse0_mto')
+        mto_route.action_unarchive()
+        buy_route = self.env.ref('purchase_stock.route_warehouse0_buy')
+        self.dropship_product.route_ids += buy_route
+
+        so = self.env['sale.order'].create({
+            'partner_id': self.customer.id,
+            'order_line': [Command.create({
+                'product_id': self.dropship_product.id,
+                'product_uom_qty': 1.00,
+                'price_unit': 1,
+            })],
+        })
+        so.action_confirm()
+        po = so.order_line.purchase_line_ids.order_id
+        po.picking_type_id = buy_route.rule_ids.picking_type_id[-1]
+        self.assertFalse(po.dest_address_id)
+        po.picking_type_id = self.env['stock.picking.type'].search([('name', '=', 'Dropship'), ('company_id', '=', self.env.company.id)], limit=1)
+        self.assertEqual(po.dest_address_id, self.customer)
--- a/addons/sale_purchase/models/purchase_order.py
+++ b/addons/sale_purchase/models/purchase_order.py
@@ -25,9 +25,12 @@ class PurchaseOrder(models.Model):
     @api.depends('order_line.sale_order_id.partner_shipping_id')
     def _compute_dest_address_id(self):
         super()._compute_dest_address_id()
-        po_with_address = self.filtered(lambda po: po.dest_address_id and len(po._get_sale_orders().partner_shipping_id) == 1)
-        for order in po_with_address:
-            order.dest_address_id = order._get_sale_orders().partner_shipping_id
+        for po in self:
+            if not po._should_set_dest_address():
+                continue
+            sale_orders = po._get_sale_orders()
+            if len(sale_orders.partner_shipping_id) == 1:
+                po.dest_address_id = sale_orders.partner_shipping_id
 
     def action_view_sale_orders(self):
         self.ensure_one()
@@ -77,6 +80,10 @@ class PurchaseOrder(models.Model):
                     'purchase_orders': purchase_order_lines.mapped('order_id'),
                     'purchase_order_lines': purchase_order_lines,
             })
+
+    def _should_set_dest_address(self):
+        self.ensure_one()
+        return bool(self.dest_address_id)
 
 
 class PurchaseOrderLine(models.Model):


addons/stock_dropshipping/models/purchase.py
```diff
@@ -18,6 +18,10 @@ def _compute_incoming_picking_count(self):
             order.incoming_picking_count -= dropship_count
             order.dropship_picking_count = dropship_count
 
+    @api.depends('order_line.sale_order_id.partner_shipping_id', 'picking_type_id')
+    def _compute_dest_address_id(self):
+        self.dest_address_id = self._get_sale_orders().partner_shipping_id
```
lase-odoo: This does not seem right. To begin with, you do not call super. So that none of the other override for modules installed prior to this one will ever be called. Furthermore, it is likely that you do not want to trigger this call on non-dropship/sale_pruchase flows. Other than that, do not forget that the compute can happen on record set and that the assignment needs to be done in a loop. Finally the `dest_address_id` is a many to one but the `_get_sale_orders` can return multiple orders with different `partner_shipping_id`. I think that all these use cases will trigger a traceback. Imo the compute should look something alike what we can find in `sale_purchase`:
https://github.com/odoo/odoo/blob/7abd7ba2f38fdb1953c39fd3693f012c2ad1b497/addons/sale_purchase/models/purchase_order.py#L25-L30
But then I wonder what the purpose of override is since the `sale_purchase_stock` is a dependency of `stock_dropshipping` did you just want to add the `picking_type_id` dependency ?
Are you sure that the issue is still reproducible in 19.0 since this change: https://github.com/odoo/odoo/pull/243028 ?

lase-odoo (changes requested):
Thanks for the proposition !

I did not review the test yet as there are some issues with the current fix. Also I think you should double check if the issue is still reproducible in 19.0. It looks like there is already an override of the `_compute_dest_address_id` doing the exact same thing as you want to achieve.

addons/stock_dropshipping/models/purchase.py
qucol-odoo: The issue is (unfortunately) not solved by https://github.com/odoo/odoo/pull/243028. When changing the `picking_type_id` to Dropship, the `PurchaseOrder._compute_dest_address_id` is triggered:

https://github.com/odoo/odoo/blob/7abd7ba2f38fdb1953c39fd3693f012c2ad1b497/addons/sale_purchase/models/purchase_order.py#L25-L30

This first line of that method calls the `PurchaseOrder` compute method:

https://github.com/odoo/odoo/blob/7abd7ba2f38fdb1953c39fd3693f012c2ad1b497/addons/purchase_stock/models/purchase_order.py#L80-L82

This method does not have any impact, as `self.picking_type_id.default_location_dest_id.usage != 'customer'`. Then `po_with_address` is empty because `self.dest_address_id` is empty, making the `_compute_dest_address_id` completely impactless. All of that leads to the `dest_address_id` being empty.

The objective of the overwrite here is to consistently add a `dest_address_id` for dropshipped sales order, because it is not done by the `_compute` method in `PurchaseOrder`. In that model, the `dest_address_id` is only updated if it is already present, but if the `dest_address_id` is missing (which is the case here), it is not added/updated.

addons/stock_dropshipping/models/purchase.py
lase-odoo: These dependencies are already added by the `_compute_dest_address_id` defined in the `purchase_stock` and the `sale_purchase` modules and since the `stock_dropshipping` module depends on the `sale_purchase_stock` one which depends on both of the above, adding them here isn't necessary.

lase-odoo: BOOOM 💣 

Carefull with self and loops.
```suggestion
           order.dest_address_id = order._get_sale_orders().partner_shipping_id
```

addons/stock_dropshipping/tests/test_dropship.py
lase-odoo: Use `Command.create`

lase-odoo: The `dropship_route` is already set on the product and both the buy and dropship rute are already active.
When it comes to archive records they should be un-archived using the `action_unarchive`.
```suggestion
        mto_route = self.env.ref('stock.route_warehouse0_mto')
        mto_route.action_unarchive()
        buy_route = self.env.ref('purchase_stock.route_warehouse0_buy')
        dropship_route = self.env.ref('stock_dropshipping.route_drop_shipping')
        self.dropship_product.route_ids += buy_route
```

lase-odoo: The `partner_shipping_id` is already pre-computed from the so partner is this really necessary ?
https://github.com/odoo/odoo/blob/c08b6bf4ba74a90ce55eb79b4ff55a00378dbc3b/addons/sale/models/sale_order.py#L405-L408

lase-odoo: It is already pre computed by from the product name is this really necessary ?

lase-odoo: If this line is really necessary it means that the dependencies of the compute are not corect.

lase-odoo: Why not use `dropship_route.rule_ids.picking_type_id` ?

lase-odoo: Why not use `so.picking_ids.move_ids.created_purchase_line_ids.order_id` ?

lase-odoo: If this line is really necessary it means that the dependencies of the compute are not correct. Since you change a dependency of the compute the field is marked as dirty and the orm already knows that he should recompute the value the next time it needs to access it.

lase-odoo (commented):
Thanks for the proposition ! 🪼

While I agree with what you want to achieve with the fix, since your method is doing the exact same as what is done in:
https://github.com/odoo/odoo/blob/c08b6bf4ba74a90ce55eb79b4ff55a00378dbc3b/addons/sale_purchase/models/purchase_order.py#L25-L30
But instead of doing it for the po for which `po.dest_address_id` is set you want to do it for po such that `po._is_dropshipped()`, I would rather define a method used in the filter in `sale_purchase` used in:
https://github.com/odoo/odoo/blob/c08b6bf4ba74a90ce55eb79b4ff55a00378dbc3b/addons/sale_purchase/models/purchase_order.py#L28
and then you override it in `stock_dropshipping`.
The method could be something ike:
```py
    def _should_set_dest_address(self):
        self.ensure_one()
        return bool(self.dest_address_id)
```
and in `stock_dropshipping`:
```py
    def _should_set_dest_address(self):
        return super()._should_set_dest_address() or self._is_dropshipped()
```

addons/stock_dropshipping/tests/test_dropship.py
```diff
@@ -558,3 +558,29 @@ def test_so_cancel_creates_one_activity_on_po(self):
         purchase_order.button_confirm()
         sale_order._action_cancel()
         self.assertEqual(len(purchase_order.activity_ids), 1)
+
+    def test_dest_address_for_dropship_and_mto(self):
+        """
+        Test that the destination address is correctly set on the purchase order
+        when using both the MTO and the Dropship routes
+        """
```
lase-odoo: Imo the `mto` is not relevant in this description. The use case is just that you change a regular purchase order to a dropship picking type.
```suggestion
    def test_dest_address_when_changing_po_to_dropship(self):
        """
        Check that the destination address is set on the purchase order when
        its picking type is manually changed to the dropshipping picking type.
        """
```

addons/stock_dropshipping/models/purchase.py
```diff
@@ -35,6 +35,10 @@ def _is_dropshipped(self):
         self.ensure_one()
         return self.picking_type_id and self.picking_type_id.code == 'dropship'
 
+    def _should_set_dest_address(self):
+        self.ensure_one()
```
lase-odoo: It is not necessary to repeat that `self.ensure_one()` as is it already handled in the super call (and even by the `_is_dorpshipped`).

addons/stock_dropshipping/tests/test_dropship.py
```diff
@@ -558,3 +558,29 @@ def test_so_cancel_creates_one_activity_on_po(self):
...
+            'partner_id': self.customer.id,
+            'order_line': [Command.create({
+                'product_id': self.dropship_product.id,
+                'product_uom_qty': 1.00,
+                'price_unit': 1,
+            })],
+        })
+        so.action_confirm()
+        po = so.order_line.purchase_line_ids.order_id
+        po.picking_type_id = buy_route.rule_ids.picking_type_id
```
lase-odoo: This will raise a traceback if you have 2 warehouses since the buy route has a different rule for each warehouse. In particular your test fails if you run it with demo data.
```suggestion
        buy_route.rule_ids.picking_type_id[-1]
```

addons/stock_dropshipping/tests/test_dropship.py
```diff
@@ -558,3 +558,29 @@ def test_so_cancel_creates_one_activity_on_po(self):
...
+        self.assertFalse(po.dest_address_id)
+        po.picking_type_id = dropship_route.rule_ids.picking_type_id
```
lase-odoo: Even if the has a single we could also for safety add a `[-1]` for safety.

lase-odoo (commented):
Almost there !

addons/stock_dropshipping/tests/test_dropship.py
qucol-odoo: The dropship route seems ot have multiple rules (thus multiple `picking_type_id`) when Subcontracting is enable.

PR245284

I propose to use the following instead:

```py
po.picking_type_id = self.env['stock.picking.type'].search([('name', '=', 'Dropship')])
```

(and to remove the unused `dropship_route` variable)

addons/stock_dropshipping/tests/test_dropship.py
```diff
@@ -558,3 +558,28 @@ def test_so_cancel_creates_one_activity_on_po(self):
...
+                'product_id': self.dropship_product.id,
+                'product_uom_qty': 1.00,
+                'price_unit': 1,
+            })],
+        })
+        so.action_confirm()
+        po = so.order_line.purchase_line_ids.order_id
+        po.picking_type_id = buy_route.rule_ids.picking_type_id[-1]
+        self.assertFalse(po.dest_address_id)
+        po.picking_type_id = self.env['stock.picking.type'].search([('name', '=', 'Dropship')])
```
lase-odoo: The dropship picking type is company dependent: this search could end up assigning the dropship type of an other company:
https://github.com/odoo/odoo/blob/e43f6040a6aa9e8f8c10d720a760f9c92606eab5/addons/stock_dropshipping/models/res_company.py#L40-L59
```suggestion
        po.picking_type_id = self.env['stock.picking.type'].search([('name', '=', 'Dropship'), ('company_id', '=', self.env.company.id)], limit=1)
```

PR #245285, opened 2026-01-22
[IMP] mail: make current user last in mention suggestions

Atovange:
With this commit the current user will be the last suggested partner in the suggestion list, as it's very unlikely to mention yourself so the space is better used to display other more relevant suggestions.

part of task-5867464

### DIFF
--- a/addons/mail/static/tests/suggestion/suggestion.test.js
+++ b/addons/mail/static/tests/suggestion/suggestion.test.js
@@ -870,7 +870,7 @@ test("Internal user should be displayed first", async () => {
     await contains(".o-mail-Composer-suggestion:eq(3) strong:text('Person A')");
 });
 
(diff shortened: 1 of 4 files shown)


addons/mail/static/tests/suggestion/suggestion.test.js
```diff
@@ -870,7 +870,7 @@ test("Internal user should be displayed first", async () => {
     await contains(".o-mail-Composer-suggestion:eq(3) strong:text('Person A')");
 });
 
-test("[text composer] Current user that is a follower should be considered as such", async () => {
+test("[text composer] Current user is last suggested partner", async () => {
```
Atovange, line 873: Replaced the test as the new rule takes priority over being a follower.

alexkuhn (approved):
@robodoo delegate+ r+

PR #245288, opened 2026-01-22
[PERF] pos_self_order: fix N+1 pricelist queries

juwu-odoo:
Currently when a pricelist is set on a POS session, N+1 queries are generated when loading the data. These come from the search in _get_applicable_rules() in _compute_price_rule() which is often run on one record at a time by its wrappers. This commit avoids the extra queries by performing the filtering through Python, which allows the ORM to effectively cache previously fetched pricelist rules.

Benchmark opening /pos-self/data
| product.product count | Before | After | Queries Before | Queries After |
| --------------------- | ------ | ----- | ------------- | ------------- |
| 1,500                 | 2.16s  | 0.70s | 1,711          | 184           |
| 15,000                | 24.91s | 6.45s | 17,175         | 396           |

opw-5477715

### DIFF
--- a/addons/product/models/product_pricelist.py
+++ b/addons/product/models/product_pricelist.py
@@ -234,18 +234,38 @@ class Pricelist(models.Model):
 
         return results
 
-    # Split methods to ease (community) overrides
     def _get_applicable_rules(self, products, date, **kwargs):
         self and self.ensure_one()  # self is at most one record
         if not self:
             return self.env['product.pricelist.item']
 
+        def is_applicable(rule):
+            if rule.categ_id and not any(categ.parent_path.startswith(rule.categ_id.parent_path) for categ in products.categ_id):
+                return False
+
+            if products._name == 'product.template':
+                if rule.product_tmpl_id and rule.product_tmpl_id not in products:
+                    return False
+                if rule.product_id and rule.product_id.product_tmpl_id not in products:
+                    return False
+            else:
+                if rule.product_tmpl_id and rule.product_tmpl_id not in products.product_tmpl_id:
+                    return False
+                if rule.product_id and rule.product_id not in products:
+                    return False
+
+            date_val = fields.Datetime.to_datetime(date)
+            if rule.date_start and rule.date_start > date_val:
+                return False
+            if rule.date_end and rule.date_end < date_val:
+                return False
+
+            return True
+
         # Do not filter out archived pricelist items, since it means current pricelist is also archived
         # We do not want the computation of prices for archived pricelist to always fallback on the Sales price
         # because no rule was found (thanks to the automatic orm filtering on active field)
-        return self.env['product.pricelist.item'].with_context(active_test=False).search(
-            self._get_applicable_rules_domain(products=products, date=date, **kwargs)
-        ).with_context(self.env.context)
+        return self.with_context(active_test=False).item_ids.filtered(is_applicable).with_context(self.env.context)
 
     def _get_applicable_rules_domain(self, products, date, **kwargs):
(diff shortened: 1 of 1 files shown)


addons/product/models/product_pricelist.py
```diff
@@ -234,18 +234,38 @@ def _compute_price_rule(
 
         return results
 
-    # Split methods to ease (community) overrides
     def _get_applicable_rules(self, products, date, **kwargs):
         self and self.ensure_one()  # self is at most one record
         if not self:
             return self.env['product.pricelist.item']
 
+        def is_applicable(rule):
```
pivi-odoo, line 242: Even if you are not changing an external signature, this probably can't go in stable, as it's going to break 3rd party code.
Customers that have overriden `_get_applicable_rules_domain`, will see their behaviour when calling `_get_applicable_rules` reverted to default, because the inner function `is_applicable` implements the standard behaviour, and isn't overridable.
So this can't be merged in stable.

addons/product/models/product_pricelist.py
```diff
@@ -234,18 +234,38 @@ def _compute_price_rule(
...
+
+            return True
+
         # Do not filter out archived pricelist items, since it means current pricelist is also archived
         # We do not want the computation of prices for archived pricelist to always fallback on the Sales price
         # because no rule was found (thanks to the automatic orm filtering on active field)
-        return self.env['product.pricelist.item'].with_context(active_test=False).search(
-            self._get_applicable_rules_domain(products=products, date=date, **kwargs)
-        ).with_context(self.env.context)
+        return self.with_context(active_test=False).item_ids.filtered(is_applicable).with_context(self.env.context)
```
pivi-odoo, line 268:
```suggestion
        return self.with_context(active_test=False).item_ids.filtered_domain(
            self._get_applicable_rules_domain(products=products, date=date, **kwargs)
        ).with_context(self.env.context)
```
If we were to do some python filtering, we should re-use the domain (so we get 3rd party customization).

Would be interesting to see what is the Query count with this approach.

pivi-odoo (changes requested):
Hello @juwu-odoo,

I've left a few comments, would be interesting to see the results, but regardless, I don't think the patch is mergable in stable.
I prefer the proposition you made in https://github.com/odoo/odoo/pull/244167
It's less invasive.

addons/product/models/product_pricelist.py
juwu-odoo: `filtered_domain` is a good idea yeah! And it would be more compatible with stable since we keep the logic from `_get_applicable_rules_domain`

The only issue is that `filtered_domain` still generates a query due to the `('categ_id', 'parent_of', products.categ_id.ids)` leaf. So the N+1 issue would still be there.
https://github.com/odoo/odoo/blob/76e5b85bd6d0e300079de1f6610f2bcb12a64803/odoo/models.py#L6593-L6601

Maybe we could add an optimization to `filtered_domain` itself...

pivi-odoo: Yeah there is only an optimization for company related fields.
`filtered_domain` has been significantly reworked in 18.4, and I would expect it to be faster (basically it was migrated to the new Domain object, applying domain optimizations and all its related shenanigans). But I have no idea of the performance implication of it.

I think trying to fix the issue here at this level is not possible, so I suggest we just proceed with the other PR. 

Any form of viable fix here would involve you parsing the domain returned from _get_applicable_rules_domain and replace the categ leaf with ids you've explicitely searched yourself, if it even works, it would be quite the ugly "hack".

juwu-odoo: Makes sense to me, thanks :)

PR #245290, opened 2026-01-22
[IMP] hr_attendance: revamp attendance systray

RaoufGhrissi:
This PR revamps the attendance systray to provide a detailed overview of the user's attendances, rather than just aggregated statistics.

Task-5180326

### DIFF
--- a/addons/hr_attendance/__manifest__.py
+++ b/addons/hr_attendance/__manifest__.py
@@ -39,7 +39,12 @@ actions(Check in/Check out) performed by them.
         'web.assets_backend': [
             'hr_attendance/static/src/**/*.js',
             'hr_attendance/static/src/**/*.xml',
-            'hr_attendance/static/src/scss/views/*.scss'
+            'hr_attendance/static/src/scss/views/*.scss',
+            # Don't include dark mode files in light mode
+            ('remove', 'hr_attendance/static/src/**/*.dark.scss'),
+        ],
+        'web.assets_web_dark': [
+            'hr_attendance/static/src/**/*.dark.scss',
         ],
         'web.assets_unit_tests': [
             'hr_attendance/static/tests/*.test.js',
@@ -69,7 +74,6 @@ actions(Check in/Check out) performed by them.
             # Public Kiosk app and its components
             "hr_attendance/static/src/public_kiosk/**/*",
             'hr_attendance/static/src/components/**/*',
-            ('remove', 'hr_attendance/static/src/components/attendance_menu/**/*'),
 
             'hr_attendance/static/src/scss/kiosk/hr_attendance.scss',
             "web/static/src/views/fields/formatters.js",
@@ -85,7 +89,7 @@ actions(Check in/Check out) performed by them.
             "barcodes/static/src/components/barcode_scanner.scss",
             "barcodes/static/src/barcode_service.js",
 
-        ]
+        ],
     },
     'author': 'Odoo S.A.',
     'license': 'LGPL-3',
--- a/addons/hr_attendance/controllers/main.py
+++ b/addons/hr_attendance/controllers/main.py
@@ -26,8 +26,10 @@ class HrAttendance(http.Controller):
         if employee:
             response = {
                 'id': employee.id,
+                'name': employee.name,
                 'hours_today': float_round(employee.hours_today, precision_digits=2),
                 'hours_previously_today': float_round(employee.hours_previously_today, precision_digits=2),
+                'today_attendance_ids': employee.today_attendance_ids.read(['check_in', 'check_out', 'worked_hours']),
                 'last_attendance_worked_hours': float_round(employee.last_attendance_worked_hours, precision_digits=2),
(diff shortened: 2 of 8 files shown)


addons/hr_attendance/controllers/main.py
```diff
@@ -26,8 +26,10 @@ def _get_user_attendance_data(employee):
         if employee:
             response = {
                 'id': employee.id,
+                'name': employee.name,
                 'hours_today': float_round(employee.hours_today, precision_digits=2),
                 'hours_previously_today': float_round(employee.hours_previously_today, precision_digits=2),
+                'today_attendance_ids': employee.today_attendance_ids.read(['check_in', 'check_out', 'worked_hours']),
```
xavierbol, line 32: IMO, the new field is only useful here so why not just doing a search_read here instead of adding a computed field. 🤷🏼‍♂️

addons/hr_attendance/__manifest__.py
```diff
@@ -69,7 +74,7 @@
             # Public Kiosk app and its components
             "hr_attendance/static/src/public_kiosk/**/*",
             'hr_attendance/static/src/components/**/*',
-            ('remove', 'hr_attendance/static/src/components/attendance_menu/**/*'),
+            # ('remove', 'hr_attendance/static/src/components/attendance_menu/**/*'),
```
xavierbol: ? 🤔

addons/hr_timesheet/security/hr_timesheet_security.xml
```diff
@@ -52,11 +52,12 @@
...
                 ('user_id', '=', user.id),
-                ('project_id', '!=', False),
                 '|', '|',
-                    ('project_id.privacy_visibility', 'in', ['employees', 'portal']),
                     ('partner_id', '=', user.partner_id.id),
-                    ('message_partner_ids', 'in', [user.partner_id.id])
+                    ('message_partner_ids', 'in', [user.partner_id.id]),
+                    '&amp;',
+                        ('project_id', '!=', False),
+                        ('project_id.privacy_visibility', 'in', ['employees', 'portal'])
```
xavierbol:
```xml
('user_id', '=', 'user.id'),
'|',
    ('employee_id.user_id', '=', user.id),
    '&amp;',
        ('project_id', '!=', False),
        '|', '|',
            ('project_id.privacy_visibility', 'in', ['employees', 'portal']),
            ('partner_id', '=', user.partner_id.id),
            ('message_partner_ids', 'in', [user.partner_id.id])
```
I think it would be better to check if the employee_id field is the user to make sure the user cannot alter analytic lines but just timesheets when project is not set.

addons/hr_timesheet/security/hr_timesheet_security.xml
```diff
@@ -65,19 +66,20 @@
             <field name="name">account.analytic.line.timesheet.approver</field>
             <field name="model_id" ref="analytic.model_account_analytic_line" />
             <field name="domain_force">[
-                ('project_id', '!=', False),
                 '|', '|',
-                    ('project_id.privacy_visibility', 'in', ['employees', 'portal']),
-                    ('message_partner_ids', 'in', [user.partner_id.id]),
                     ('partner_id', '=', user.partner_id.id),
+                    ('message_partner_ids', 'in', [user.partner_id.id]),
```
xavierbol, line 71: We should either be sure the timesheet is own or project_id is set, otherwise the timesheet approver could also alter the basic AAL. 🙂

addons/hr_timesheet/security/hr_timesheet_security.xml
```diff
@@ -65,19 +66,20 @@
...
+                        ('project_id.privacy_visibility', 'in', ['employees', 'portal'])
             ]</field>
             <field name="groups" eval="[(4, ref('hr_timesheet.group_hr_timesheet_approver'))]" />
         </record>
 
         <record id="timesheet_line_rule_manager" model="ir.rule">
             <field name="name">account.analytic.line.timesheet.manager</field>
             <field name="model_id" ref="analytic.model_account_analytic_line"/>
-            <field name="domain_force">[('project_id', '!=', False)]</field>
+            <field name="domain_force">['|', ('project_id', '!=', False), ('user_id', '=', user.id)]</field>
```
xavierbol: it would be better to use `employee_id` instead of `user_id`, I think

xavierbol (commented):

addons/hr_timesheet/security/hr_timesheet_security.xml
MissingNoShiny: `employee_id` is not set when there is no project iirc

xavierbol: ahhh that's annoying

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #245297, opened 2026-01-22
[FIX] portal_address_extended: quick fixes

Feyensv:
* Do not consider current company country, only customer country
* Do not enforce cities if no cities are defined for the country
* Do not force city choice for all countries, feature will be improved in the coming months but it's too restrictive to impose on all the countries set to `enforce_cities` in the data (e.g. us payroll enables it for US, leading to issues when doing a demo with mitchell admin on runbot).

Current enforcement makes it impossible to buy on the ecommerce if data is incomplete (e.g. if you select Micronesia in the US states, there is no available city, so there is no way to save your address before this commit).

---

### DIFF
--- a/addons/l10n_br/controllers/portal.py
+++ b/addons/l10n_br/controllers/portal.py
@@ -2,24 +2,13 @@
 
 from odoo.http import request
 
-from odoo.addons.l10n_latam_base.controllers.portal import L10nLatamBasePortalAccount
+from odoo.addons.portal_address_extended.controllers.portal import CustomerPortalExtended
 
 
-class L10nBRPortalAccount(L10nLatamBasePortalAccount):
+class L10nBRPortalAccount(CustomerPortalExtended):
 
     def _is_brazilean_fiscal_country(self):
         return request.env.company.account_fiscal_country_id.code == 'BR'
-
-    def _prepare_address_form_values(self, partner_sudo, *args, **kwargs):
-        rendering_values = super()._prepare_address_form_values(partner_sudo, *args, **kwargs)
-        if self._is_brazilean_fiscal_country():
-            rendering_values.update({
-                'city': partner_sudo.city_id,
-                'state_cities': request.env['res.city'].sudo().search([
-                    ('country_id.code', '=', 'BR'),
-                ]),
-            })
-        return rendering_values
 
     def _get_mandatory_address_fields(self, country_sudo):
         mandatory_fields = super()._get_mandatory_address_fields(country_sudo)
--- a/addons/l10n_br/static/src/interactions/address.js
+++ b/addons/l10n_br/static/src/interactions/address.js
@@ -11,12 +11,12 @@ patch(CustomerAddress.prototype, {
 
     async onChangeZip() {
         await super.onChangeZip();
-        if (this.countryCode !== 'BR' || this._getSelectedCountryCode() !== 'BR') return;
+        if (this._getSelectedCountryCode() !== 'BR') return;
 
         const newZip = this.addressForm.zip.value.padEnd(5, '0');
 
-        for (const option of this.addressForm.querySelectorAll('select[name="city_id"] option')) {
-            const ranges = option.getAttribute('zip-ranges');
+        for (const option of this.elementCities.options) {
+            const ranges = option.dataset.l10n_br_zip_ranges;
             if (ranges) {
                 // Parse the l10n_br_zip_ranges field (e.g. "[01000-001 05999-999] [08000-000 08499-999]").
                 // Loop over each range that is enclosed in [] (e.g. "[01000-001 05999-999]" followed by "[08000-000 08499-999]").
@@ -32,7 +32,7 @@ patch(CustomerAddress.prototype, {
                             new CustomEvent('select', { detail: { value: option.value } })
                         );
                         option.selected = 'selected';
-                        this._selectState(option.getAttribute('state-id'));
+                        this._selectState(option.dataset.state_id);
                         return;
                     }
                 }
@@ -42,13 +42,11 @@ patch(CustomerAddress.prototype, {
 
     async onChangeCity() {
         await super.onChangeCity();
-        if (this.countryCode !== 'BR' || this._getSelectedCountryCode() !== 'BR') return;
+        if (this._getSelectedCountryCode() !== 'BR') return;
 
         if (this.elementCities.value) {
             this._selectState(
-                this.elementCities
-                    .querySelector(`option[value='${this.elementCities.value}']`)
-                    .getAttribute('state-id')
+                this.elementCities.selectedOptions[0].dataset.state_id
             );
         }
     },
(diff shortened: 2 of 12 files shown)


addons/portal_address_extended/models/res_country.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = 'res.country'
+
+    def _enforce_city_choice(self):
+        if not self:
+            return False
+
+        # Only enabled on frontend for those countries for now
+        # Feature has to be adapted to be more generic and less blocking
+        # before being enabled for other countries
+        if self.code not in ['BR', 'CL', 'PE', 'CO']:
```
jco-odoo: Need to see for TW with @vin-odoo

addons/portal_address_extended/controllers/portal.py
```diff
@@ -1,27 +1,31 @@
...
-            ])
+        country = rendering_values['country']
+        state_id = rendering_values['state_id']
+        if country._enforce_city_choice():
+            state_cities = (
+                request.env['res.city']
+                .sudo()
+                .search([
+                    ('country_id', '=', country.id),
+                    ('state_id', 'in', [state_id, False]),
```
jco-odoo: Is the t-att-state-id="c.state_id.id" then still important?  (on the city_id in the view)

Feyensv: Mhm, there seems to be some kind of logic to select the State according to the chosen city in the `l10n_br` loca (`onChangeCity` override).

Do they not know their state or tend to select the city first ? :thinking:

jco-odoo: @Feyensv I guess zip -> city -> state

jco-odoo: But it is overridden anyways

Feyensv: Well, will have to fix the override, as it seems that loaded cities will always be BR ones if company country is, even if the partner is in another country :thinking:

jco-odoo: @Feyensv Somehow it gets messed up with PE and BR l10n in the same db and BR handles the ZIP ranges of the cities through js. If you select PE first and then BR, the cities do not seem to come back.

jco-odoo: @Feyensv Talking to @vbe-odoo , it would not be a problem if the cities are only there for the state for BR as well, but then we need to do a call in onchangezip to change city/state (when country BR).

addons/portal_address_extended/models/res_country.py
vin-odoo: @Feyensv For Taiwan, cities are usually required in the addresses to be accurate and complete.
As we support TW e-invoicing,  it would be beneficial to enforce it there to reduce the risks of mistakes.

Thank you for the ping @jco-odoo

addons/portal_address_extended/controllers/portal.py
```diff
@@ -44,5 +40,29 @@ def _parse_form_data(self, form_data):
...
-        cities = request.env['res.city'].sudo().search([('state_id', '=', state.id)])
-        return {'cities': [(city.id, city.name) for city in cities]}
+        country = state.country_id
+
+        if country._enforce_city_choice():
+            return {
+                'cities': request.env['res.city'].sudo().search_read(
+                    [('state_id', '=', state.id)],
+                    country._get_cities_fields_to_fetch(),
+                    load="whatever",  # we only want the ids of relational fields
```
ltinel:
```suggestion
                    load='',  # we only want the ids of relational fields
```
Takes less space and seems a bit more standard 😅 Same below

addons/portal_address_extended/views/portal_address_templates.xml
```diff
@@ -2,27 +2,25 @@
...
                 <label class="col-form-label label-optional" for="city_id">City</label>
                 <select id="o_city_id" name="city_id" class="form-select">
                     <option value="">City...</option>
                     <option t-foreach="state_cities"
                             t-as="c"
                             t-att-value="c.id"
-                            t-att-selected="c.id == (city.id if city else partner_sudo.city_id.id)"
-                            t-att-state-id="c.state_id.id"
+                            t-att-selected="c.id == city.id"
+                            t-att-data-state_id="c.state_id.id"
```
ltinel, line 17: It's recommended to uses dashes over underscores (that way, dataset keys are automatically converted to camelCase in JS, e.g. `data-state-id` in XML becomes `dataset.stateId` in JS).

addons/l10n_br/views/portal_address_templates.xml
```diff
@@ -3,7 +3,7 @@
 
     <template id="address_form_fields" inherit_id="portal.address_form_fields">
         <xpath expr="//select[@name='city_id']/option[not(@value='')]" position="attributes">
-            <attribute name="t-att-zip-ranges" add="c.l10n_br_zip_ranges"/>
+            <attribute name="t-att-data-l10n_br_zip_ranges" add="c.l10n_br_zip_ranges"/>
```
ltinel, line 6: Dashes instead of underscores :)

addons/l10n_br/controllers/portal.py
```diff
@@ -2,25 +2,14 @@
 
 from odoo.http import request
 
-from odoo.addons.l10n_latam_base.controllers.portal import L10nLatamBasePortalAccount
+from odoo.addons.portal_address_extended.controllers.portal import CustomerPortalExtended
 
 
-class L10nBRPortalAccount(L10nLatamBasePortalAccount):
```
ltinel, line 8: I'm not familiar with this code, but don't we lose some functionality by not inheriting `L10nLatamBasePortalAccount` anymore?

addons/l10n_br/static/src/interactions/address.js
```diff
@@ -11,12 +11,12 @@ patch(CustomerAddress.prototype, {
 
     async onChangeZip() {
         await super.onChangeZip();
-        if (this.countryCode !== 'BR' || this._getSelectedCountryCode() !== 'BR') return;
+        if (this._getSelectedCountryCode() !== 'BR') return;
```
ltinel, line 14: Do we need to make similar changes for PE and CO?

addons/portal_address_extended/views/portal_address_templates.xml
Feyensv: Field names are kept as is to ease the implementation of `_changeOption`

addons/l10n_br/static/src/interactions/address.js
Feyensv: CO logic is purely accounting so no need if company fiscal country is not CO.
PE would require a template change as well (and districts are not the target of the recent task anyway).

addons/l10n_br/controllers/portal.py
Feyensv: Not that I see, that class is for accounting logic, whereas this one is about the address

addons/portal_address_extended/controllers/portal.py
Feyensv: Indeed :eyes:

ltinel:
@robodoo delegate+

ltinel (approved):

Feyensv:
robodoo r+

Feyensv:
robodoo r+

PR #245304, opened 2026-01-22
[FIX] mass_mailing: don't enforce deep font-family

abd-msyukyu-odoo:
There is an issue where some editor descendants have a rule overriding their
`font-family` which has more precedence than the inherited `font-family` from
the Design Tab.

How to reproduce:
- create a new mass_mailing
- add a text snippet
- select all text and change e.g. the `font-size`
- open the Design Tab and change the `font-family` of paragraphs

Issue:
- the text `font-family` does not change

Resolution:
Fix a css rule applying on every element that is not a `fa` element so that it
only applies on the `o_layout` element to let the Design Tab override the
`font-family` for specific elements (e.g. `<p>`, `<a>`, ...).

task-5869568

### DIFF
--- a/addons/mass_mailing/static/src/scss/mass_mailing_mail.scss
+++ b/addons/mass_mailing/static/src/scss/mass_mailing_mail.scss
@@ -2,7 +2,7 @@
 .o_layout * {
     box-sizing: border-box !important;
 }
-.o_layout :not(.fa) {
+.o_layout {
     font-family: Arial,Helvetica Neue,Helvetica,sans-serif;
 }
 /* Remove space around the email design. */


addons/mass_mailing/static/src/scss/mass_mailing_mail.scss
```diff
@@ -2,7 +2,7 @@
 .o_layout * {
     box-sizing: border-box !important;
 }
-.o_layout :not(.fa) {
+.o_layout {
     font-family: Arial,Helvetica Neue,Helvetica,sans-serif;
```
julien-banken, line 6: I assume we still need to use these fonts as fallback options, as they are more widely available across most operating systems and web clients compared to the Odoo fonts (-> `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Ubuntu, "Noto Sans", Arial, "Odoo Unicode Support Noto", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"`).

julien-banken (approved):

addons/mass_mailing/static/src/scss/mass_mailing_mail.scss
abd-msyukyu-odoo: From my understanding, `Arial,Helvetica Neue,Helvetica,sans-serif` is indeed the fallback font chosen historically. I think there is a more broad issue with the file `mass_mailing_mail.scss`, I'm not sure it should even be there during `convert_inline` (but it was there before the big refactoring so I let it stay since I could not measure the impact of removing it at the time).

I'm re-evaluating how to include the rules from this file in my refactoring, in the meantime this simplistic fix should do the trick

julien-banken:
@robodoo r+

PR #245306, opened 2026-01-22
[FIX] *: replace "Deco Addict" with "Acme Corporation"

reth-odoo:
A company that happens to be named "Deco Addict" has complained some of
our users thought they had business with them due to test and demo data
containing that name.

It will now be named Acme Corporation.

task-5865502


addons/account/static/src/components/bill_guide/bill_guide.js
```diff
@@ -16,10 +16,12 @@ export class BillGuide extends Component {
         this.action = useService("action");
         this.context = null;
         this.alias = null;
+        this.showSampleAction = false;
         onWillStart(this.onWillStart);
     }
 
     async onWillStart() {
+        const showSamplePromise = this.orm.call("account.journal", "should_show_sample_action");
```
awa-odoo: That is indeed annoying because it triggers every single time you load the dashboard :(

But it looks like it's only until you create your first bill and then it disappears, it's for onboarding purposes, so it should have a limited impact hopefully.
We could create some kind of cache in the localStorage but that seems overkill :thinking:

addons/account/models/account_journal_dashboard.py
```diff
@@ -898,6 +898,11 @@ def _build_no_journal_error_msg(self, company_name, journal_types):
                 journal_types=', '.join(journal_types),
             )
 
+    @api.model
+    def should_show_sample_action(self):
```
awa-odoo: Is it not possible to query (search_count) the ir.model.data on the "name" property to try to see if we have a "base.res_partner_2" in there from the JS side?

Would avoid a new model method for our niche case.

addons/account/models/account_journal_dashboard.py
```diff
@@ -908,13 +913,10 @@ def action_create_vendor_bill(self):
...
         invoice_date = fields.Date.today() - timedelta(days=12)
-        partner = self.env['res.partner'].search([('name', '=', 'Acme Corporation')], limit=1)
+        partner = self.env.ref('base.res_partner_2', raise_if_not_found=False)
         company = purchase_journal.company_id
         if not partner:
-            partner = self.env['res.partner'].create({
-                'name': 'Acme Corporation',
-                'is_company': True,
-            })
+            raise UserError(_('You may only use samples in demo mode, try uploading one of your invoices instead.'))
```
awa-odoo: Would move right after the "if not purchase_journal:" so both pre-conditions are checked first.

addons/account/static/src/components/bill_guide/bill_guide.js
```diff
@@ -36,6 +38,7 @@ export class BillGuide extends Component {
                 default_journal_id: ctx.active_id,
             }
         }
+        this.showSampleAction = await showSamplePromise;
```
awa-odoo: Can probably be a one-liner instead of splitting to a separate variable?

awa-odoo (commented):
Hello @reth-odoo 

Had a closer look and left some comments / suggestions.
Something to occupy yourself while we wait for FW :smile: 

Thanks & cheers!

addons/account/models/account_journal_dashboard.py
reth-odoo: IrModelData is restricted to admins sadly, like IrModel. This is needed as a "sudo" getter.
Unless we can pass it some other way to the front-end.

addons/account/static/src/components/bill_guide/bill_guide.js
reth-odoo: I can store the result on the class directly so it will only do it when you reload the page.

And indeed unless the "no data" helper is displayed this component won't start (or exist) so it should be quite rare on real databases. Had not thought about that actually.

Maybe it's not even worth caching on the class then...

addons/account/static/src/components/bill_guide/bill_guide.js
reth-odoo: Idea was to avoid blocking until we actually need the result, though I guess here it's the only async call.

addons/account/models/account_journal_dashboard.py
awa-odoo: That's sad.

Tried looking for a different trick, there is /web/image that takes a xmlid so we could technically query a random field on our XML ID and then see if it returns an empty result or not but that seems too hackish ^^

Let's keep it then, maybe rename to something like "is_sample_action_available".

reth-odoo: We could also assume the demo data hasn't been changed and search for the contact with the right name like they did prior but that's a lot less reliable

PR #245368, opened 2026-01-23
[FIX] l10n_au: correct BAS formulas and WET sequence and add support for line 7A

avar-odoo:
Prior to this commit, the Deferred GST (DGST) tax was configured with a positive rate. When applied to a Vendor Bill, this generated a Debit entry, incorrectly treating the tax as a Receivable rather than a Liability to the ATO.

This commit corrects the accounting and reporting logic by:

- Inverting the tax rate: The tax now generates a Credit entry on Vendor Bills, correctly recording the Deferred GST as a liability in the General Ledger.
- Adjusting BAS formulas: Updated the formulas for labels 7 and 7A to display a positive liability amount on the statement following the change in 1.
- Adding Label 7A: Added support for the Deferred GST (7A) line to BAS variations where it was missing.
- Reordering WET: Updated the Wine Equalisation Tax (WET) sequence so it no longer appears at the top of the list, as it is not applicable to all businesses.

Task-5870092

---


addons/l10n_au/data/template/account.tax-au.csv
kidoodoo:
```suggestion
"au_tax_sale_wet","WET","38","Wine Equalisation Tax","Wine Equalisation Tax Amount","sale","","percent","","29","","","","base","invoice","","","True","","l10n_au_domestic_fiscal_position",""
```

kidoodoo (commented):
No further change, let's remove enterprise branch + PR.

Good to go on my end. Once you push with below changes, will ping Nicolas once tests are all green. Thank you!

vin-odoo (changes requested):
The fixes themselves are ok, but the commit message isn't good enough.
Your sentence talks about 'these accounts', but no accounts are in the PR at all.
You also do not explain **why** you need to change the sequence, etc.

It's also a bit misleading, as you do double sign inversion here. We invert the sign on the tax repartition lines; and again in the reports to correct the entries while keeping the report signs correct.

We can move forward with this fix once the commit message is clearer 👍
Thank you for the work!

vin-odoo (approved):
robodoo r+

Thank you!

PR #245390, opened 2026-01-23
[IMP] l10n_tr_nilvera_einvoice: Add Buyer and Seller Line code for Export invoices

sham-odoo:
This commit introduces two new fields on product and invoice lines: `Customer Line Code` and `Seller Line Code`.

It introduces an exemption reason for Registered for Export type invoices with exemption code `702`.
It makes the Customer Line code on each invoice line mandatory for invoices having exemption code `702` while sending e-invoice to Nilvera.

Adds test case for the above scenario.

task-5010752

---

### DIFF
--- a/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
@@ -28,7 +28,7 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         # We'll replace the empty value by a dummy one so that the node doesn't get cleaned up and remove its content after the file generation.
         xml, errors = super()._export_invoice(invoice)
         xml_root = etree.fromstring(xml)
-        shipment_id_elements = xml_root.findall('.//cac:Shipment/cbc:ID', namespaces=xml_root.nsmap)
+        shipment_id_elements = xml_root.findall('.//cac:Shipment//cbc:ID', namespaces=xml_root.nsmap)
         for element in shipment_id_elements:
             if element.text == 'NO_ID':
                 element.text = ''
@@ -221,6 +221,21 @@ class AccountEdiXmlUblTr(models.AbstractModel):
             return self._add_withholding_document_line_tax_total_nodes(line_node, vals)
         return super()._add_document_line_tax_total_nodes(line_node, vals)
 
+    def _l10n_tr_get_issuer_party_line_codes(self, move_line):
+        return {
+            'cac:PartyIdentification': [
+                {'cbc:ID': {'_text': move_line.l10n_tr_customer_line_code, 'schemeID': 'ALICIDIBSATIRKOD'}},
+                {'cbc:ID': {'_text': move_line.l10n_tr_seller_line_code, 'schemeID': 'SATICIDIBSATIRKOD'}},
+            ],
+            'cac:PostalAddress': {
+                'cbc:CitySubdivisionName': {'_text': move_line.move_id.partner_id.city},
+                'cbc:CityName': {'_text': move_line.move_id.partner_id.state_id.with_context(lang='tr_TR').name},
+                'cac:Country': {
+                    'cbc:Name': {'_text': move_line.move_id.partner_id.country_id.with_context(lang='tr_TR').name},
+                },
+            },
+        }
+
     def _add_invoice_line_delivery_nodes(self, line_node, vals):
         """Add delivery information to an invoice line node for export invoices.
 
@@ -248,6 +263,24 @@ class AccountEdiXmlUblTr(models.AbstractModel):
                     'cbc:ID': {'_text': 'NO_ID'},
                     'cac:GoodsItem': {'cbc:RequiredCustomsID': {'_text': move_line.l10n_tr_ctsp_number or move_line.product_id.l10n_tr_ctsp_number}},
                     'cac:ShipmentStage': {'cbc:TransportModeCode': {'_text': move_line.move_id.l10n_tr_shipping_type}},
+                },
+            }
+
+        if move_line.move_id.l10n_tr_exemption_code_id.code == 702:
+            line_node['cac:Delivery'] = {
+                'cac:Shipment': {
+                    'cbc:ID': {'_text': 'NO_ID'},
+                    'cac:GoodsItem': {
+                        'cbc:RequiredCustomsID': {
+                            '_text': move_line.l10n_tr_ctsp_number or move_line.product_id.l10n_tr_ctsp_number,
+                        },
+                    },
+                    'cac:TransportHandlingUnit': {
+                        'cac:CustomsDeclaration': {
+                            'cbc:ID': {'_text': 'NO_ID'},
+                            'cac:IssuerParty': self._l10n_tr_get_issuer_party_line_codes(move_line),
+                        },
+                    },
                 },
             }
 
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
@@ -11,6 +11,20 @@ class AccountMoveLine(models.Model):
         store=True,
         readonly=False,
     )
+    l10n_tr_seller_line_code = fields.Char(
+        string="Seller Line Code",
+        compute='_compute_l10n_tr_seller_line_code',
+        store=True,
+        readonly=False,
+    )
+    l10n_tr_customer_line_code = fields.Char(
+        string="Customer Line Code",
+        help="Customer Line Code (11 or fewer characters)",
+        size=11,
+        compute='_compute_l10n_tr_customer_line_code',
+        store=True,
+        readonly=False,
+    )
 
     @api.depends("product_id.l10n_tr_ctsp_number")
     def _compute_l10n_tr_ctsp_number(self):
@@ -22,3 +36,13 @@ class AccountMoveLine(models.Model):
         for record in self:
             if record.l10n_tr_ctsp_number and len(record.l10n_tr_ctsp_number) > 12:
                 raise ValidationError(_("CTSP Number must be 12 digits or fewer."))
+
+    @api.depends("product_id.l10n_tr_seller_line_code")
+    def _compute_l10n_tr_seller_line_code(self):
+        for record in self:
+            record.l10n_tr_seller_line_code = record.product_id.l10n_tr_seller_line_code
+
+    @api.depends("product_id.l10n_tr_customer_line_code")
+    def _compute_l10n_tr_customer_line_code(self):
+        for record in self:
+            record.l10n_tr_customer_line_code = record.product_id.l10n_tr_customer_line_code
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
@@ -52,6 +52,8 @@ class AccountMoveSend(models.AbstractModel):
 
         # Filter for moves that have 'tr_nilvera' in their EDI data
         tr_nilvera_moves = moves.filtered(lambda m: 'tr_nilvera' in moves_data[m]['extra_edis'])
+        if not tr_nilvera_moves:
+            return alerts
 
         if tr_companies_missing_required_codes := tr_nilvera_moves.company_id.filtered(lambda c: c.country_code == 'TR' and not (c.partner_id.category_id.parent_id and self.env["res.partner.category"]._get_l10n_tr_official_mandatory_categories())):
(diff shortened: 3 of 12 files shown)


sham-odoo:
hello @odoo/rd-accounting-flg 
can you please review this PR?

thanks!

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -221,6 +221,26 @@ def _add_document_line_tax_total_nodes(self, line_node, vals):
...
         return super()._add_document_line_tax_total_nodes(line_node, vals)
 
+    def _l10n_tr_get_issuer_party_line_codes(self, move_line):
+        move_line_codes = {
+            'ALICIDIBSATIRKOD': move_line.l10n_tr_customer_line_code,
+            'SATICIDIBSATIRKOD': move_line.l10n_tr_seller_line_code,
+        }
+        return {
+            'cac:PartyIdentification': [
+                {'cbc:ID': {'_text': code, 'schemeID': schemeID}}
```
FlorianGilbert: It's always 2 items right?
Look a bit overkill to do that and too much lines for nothing. 

```suggestion
                {'cbc:ID': {'_text': move_line.l10n_tr_customer_line_code, 'schemeID': 'ALICIDIBSATIRKOD'}},
                {'cbc:ID': {'_text': move_line.l10n_tr_seller_line_code, 'schemeID': 'SATICIDIBSATIRKOD'}},
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -221,6 +221,26 @@ def _add_document_line_tax_total_nodes(self, line_node, vals):
             return self._add_withholding_document_line_tax_total_nodes(line_node, vals)
         return super()._add_document_line_tax_total_nodes(line_node, vals)
 
+    def _l10n_tr_get_issuer_party_line_codes(self, move_line):
+        move_line_codes = {
+            'ALICIDIBSATIRKOD': move_line.l10n_tr_customer_line_code,
+            'SATICIDIBSATIRKOD': move_line.l10n_tr_seller_line_code,
+        }
```
FlorianGilbert: Then we can remove this

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -28,7 +28,7 @@ def _export_invoice(self, invoice):
         # We'll replace the empty value by a dummy one so that the node doesn't get cleaned up and remove its content after the file generation.
         xml, errors = super()._export_invoice(invoice)
         xml_root = etree.fromstring(xml)
-        shipment_id_elements = xml_root.findall('.//cac:Shipment/cbc:ID', namespaces=xml_root.nsmap)
+        shipment_id_elements = xml_root.findall('.//cac:Shipment//cbc:ID', namespaces=xml_root.nsmap)
```
FlorianGilbert, line 31: Why?

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -11,6 +11,20 @@ class AccountMoveLine(models.Model):
...
+        compute='_compute_l10n_tr_seller_line_code',
+        store=True,
+        readonly=False,
+    )
+    l10n_tr_customer_line_code = fields.Char(
+        string="Customer Line Code",
+        help="Customer Line Code (11 or fewer characters)",
+        size=11,
+        compute='_compute_l10n_tr_customer_line_code',
+        store=True,
```
FlorianGilbert, line 25: Do we really need to store it in account.move.line and make this table even heavier? 😄

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
```diff
@@ -11,6 +11,20 @@ class AccountMoveLine(models.Model):
...
+        readonly=False,
+    )
```
FlorianGilbert, line 27: And why it's not related field?

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -136,7 +136,9 @@ def _is_valid_nilvera_name(move):
             }
 
         # Warning alert if product is missing CTSP Number
-        tr_export_moves = tr_nilvera_moves.filtered(lambda m: m.l10n_tr_is_export_invoice)
+        tr_export_moves = tr_nilvera_moves.filtered(
+            lambda m: m.l10n_tr_is_export_invoice or m.l10n_tr_exemption_code_id.code == 702,
```
FlorianGilbert: Avoid magic number, let's use a constant or something to know what does 702 mean?

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -152,6 +154,23 @@ def _is_valid_nilvera_name(move):
                 ),
             }
 
+        if moves_with_missing_line_codes := tr_nilvera_moves.filtered(
+            lambda move: move.l10n_tr_exemption_code_id.code == 702 and move.invoice_line_ids.filtered(
```
FlorianGilbert: Could also be use there

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -152,6 +154,23 @@ def _is_valid_nilvera_name(move):
...
+                lambda ml: not ml.l10n_tr_customer_line_code
+            ),
```
FlorianGilbert: filtered in filtered leads to several loops. 
What about this:

```suggestion
        if moves_with_missing_line_codes := tr_nilvera_moves.invoice_line_ids.filtered(
            lambda move_line: move_line.move_id.l10n_tr_exemption_code_id.code == 702 and not move_line.l10n_tr_customer_line_code
        ),
```
I don't think that we can even inverse the condition as we are iterating on move line and then we can return early faster. Depending of what it's the most probable case.

addons/l10n_tr_nilvera_einvoice/tests/test_xml_ubl_tr_common.py
```diff
@@ -87,6 +87,7 @@ def _generate_invoice_xml(self, partner_id, tax=None, **kwargs):
                     'quantity': 3,
                     'discount': 12,
                     'tax_ids': [Command.set(invoice_tax)],
+                    **kwargs.pop('invoice_line_data', {}),
```
FlorianGilbert, line 90: Why? Don't get this change

addons/l10n_tr_nilvera_einvoice/views/product_views.xml
```diff
@@ -13,6 +13,16 @@
...
+                        <field name="l10n_tr_customer_line_code"
+                                 invisible="product_variant_count &gt; 1
+                                        or (product_variant_count == 0
+                                        and valid_product_template_attribute_line_ids)
+                                        or 'TR' not in fiscal_country_codes"/>
+                        <field name="l10n_tr_seller_line_code"
+                                 invisible="product_variant_count &gt; 1
+                                        or (product_variant_count == 0
+                                        and valid_product_template_attribute_line_ids)
+                                        or 'TR' not in fiscal_country_codes"/>
```
FlorianGilbert: Nitpick but I think the condition on country code could be moved at the beginning of the condition to have something cleaner

FlorianGilbert (changes requested):
Thank you! 
What do you think about improving a bit the commit message? Comparing to the spec, it's not very clear to understand all the changes. Thank you!

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
yace-odoo: @FlorianGilbert These values change per invoice, any other way to achieve this? @sham-odoo

FlorianGilbert: > @FlorianGilbert These values change per invoice, any other way to achieve this? @sham-odoo

@yace-odoo Per invoice but then it's invoice line so is it necessary?

yace-odoo: Sorry, meant to say, per product and per invoice.

So each invoice line can have a different Customer/Supplier code per invoice

FlorianGilbert: Makes more sense but it couldn't differ from the product code, right?

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
sham-odoo: hello @FlorianGilbert 
Let's say this is the xml snippet:
```xml
...
<cac:Shipment>
    <cbc:ID></cbc:ID>
    <cac:TransportHandlingUnit>
        <cac:CustomsDeclaration>
            <cbc:ID></cbc:ID>
            <cac:IssuerParty>
                <cac:PartyIdentification>
                    <cbc:ID schemeID="SATICIDIBSATIRKOD">55321260090</cbc:ID>
                </cac:PartyIdentification>
                <cac:PostalAddress>
                    <cbc:CitySubdivisionName>Beyoğlu</cbc:CitySubdivisionName>
                    <cbc:CityName>İstanbul</cbc:CityName>
                    <cac:Country>
                        <cbc:Name>Türkiye</cbc:Name>
                    </cac:Country>
                </cac:PostalAddress>
            </cac:IssuerParty>
        </cac:CustomsDeclaration>
    </cac:TransportHandlingUnit>
</cac:Shipment>
...
```

with current implementation, when we search for `cac:Shipment/cbc:ID`, it will only search for the child `cbc:ID` nodes of `cac:Shipment` and not the nested nodes inside its children. Nilvera needs `cbc:ID` node for every nodes inside `cac:Shipment` (if its defined in its schema). so to get all the nodes, I replaced `/` with `//` instead.

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
yace-odoo: Codes are dependent on several variables, 

- Customer or Sellers DiiB (IPAC) certificate
- Nature of the product
- Nature of the tax
- Destination country
- Incoterms
etc. :sweat_smile:

sham-odoo: hello @FlorianGilbert 

> So each invoice line can have a different Customer/Supplier code per invoice

I also had some similar confusion regarding this before, then had a discussion with Yasin, and got to know the following:
So basically, each invoice line needs a customer line and a seller line code that can change for each line in invoice.
the reason for storing it in product as well as in invoice is that, there is a format for this codes in which a few digits remain same while just some digits changes depending on the customer.
so if we make it related and readonly=false on the invoice line, updating this will also update the code on product which we don't want.

cc: @yace-odoo

addons/l10n_tr_nilvera_einvoice/models/account_move_line.py
sham-odoo: hello.
disscussed [here](https://github.com/odoo/odoo/pull/245390#discussion_r2744756087)

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
sham-odoo: okay :+1: 
will do a env.ref() for better clarity.

addons/l10n_tr_nilvera_einvoice/tests/test_xml_ubl_tr_common.py
sham-odoo: hello @FlorianGilbert 

this function doesn't accept any args for, invoice_line_ids.
if I gave invoice_line_ids in function params to `_generate_invoice_xml`, it would override the `invoice_line_ids` already given in the function [here](https://github.com/odoo/odoo/blob/90aa32ce4494626c627c0f2315a9627459082e4e/addons/l10n_tr_nilvera_einvoice/tests/test_xml_ubl_tr_common.py#L74)
I wanted to test by giving the customer and seller line code to the invoice lines, so this change seems a flexible approach to me. it can also be used in future to add some more data for the invoice line.

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
FlorianGilbert: ok good

FlorianGilbert (approved):
Good job

@robodoo r+

PR #245392, opened 2026-01-23
[FIX] sale_project: allow users to open milestones

njor-odoo:
Steps to reproduce:
- Install the sale_project module
- Create a sale order based on milestones
- Create the project from the order
- Open the project, click the three dots, and open a milestone

Issue:
Users are unable to open milestones and get an access error.

Cause:
Users in `sales_team.group_sale_salesman`  lack read access to the related
`sale.order.line`, causing an AccessError when `sale_line_id` is accessed
during the computation of `product_uom_qty`.

Fix:
Compute `product_uom_qty` using `sudo()` to bypass record rule restrictions.

task-5477304

### DIFF
--- a/addons/sale_project/models/project_milestone.py
+++ b/addons/sale_project/models/project_milestone.py
@@ -26,7 +26,7 @@ class ProjectMilestone(models.Model):
 
     sale_line_display_name = fields.Char("Sale Line Display Name", related='sale_line_id.display_name')
     product_uom = fields.Many2one(related="sale_line_id.product_uom")
-    product_uom_qty = fields.Float("Quantity", compute="_compute_product_uom_qty", readonly=False)
+    product_uom_qty = fields.Float("Quantity", compute="_compute_product_uom_qty", compute_sudo=True, readonly=False)
 
     @api.depends('sale_line_id.product_uom_qty', 'product_uom_qty')
     def _compute_quantity_percentage(self):


addons/sale_project/models/project_milestone.py
```diff
@@ -37,9 +37,9 @@ def _compute_quantity_percentage(self):
     def _compute_product_uom_qty(self):
         for milestone in self:
             if milestone.quantity_percentage:
-                milestone.product_uom_qty = milestone.quantity_percentage * milestone.sale_line_id.product_uom_qty
+                milestone.product_uom_qty = milestone.quantity_percentage * milestone.sale_line_id.sudo().product_uom_qty
             else:
-                milestone.product_uom_qty = milestone.sale_line_id.product_uom_qty
+                milestone.product_uom_qty = milestone.sale_line_id.sudo().product_uom_qty
```
deneuvillem: We could use `compute_sudo=True`  in the field definition instead of using `sudo()` on the sale lines

addons/sale_project/models/project_milestone.py
deneuvillem, line 31: Not related to your work, but could we try to remove `product_uom_qty` from the dependencies of `_compute_quantity_percentage`? I suspect that there is a performance issue that is occurring because of the following:
When `_compute_product_uom_qty` is triggered, for instance, at the opening of a form/tree view, the value of `product_uom_qty` is computed, which then will trigger the `_compute_quantity_percentage`, and overwrites its stored value -> This could cause many useless writes

In short, it's not ideal that a stored computed field depends on a non-stored computed field.

addons/sale_project/models/project_milestone.py
```diff
@@ -22,24 +22,21 @@ def _default_sale_line_id(self):
 
     sale_line_id = fields.Many2one('sale.order.line', 'Sales Order Item', default=_default_sale_line_id, help='Sales Order Item that will be updated once the milestone is reached.',
         domain="[('order_partner_id', '=?', project_partner_id), ('qty_delivered_method', '=', 'milestones')]")
-    quantity_percentage = fields.Float('Quantity (%)', compute="_compute_quantity_percentage", store=True, help='Percentage of the ordered quantity that will automatically be delivered once the milestone is reached.')
+    quantity_percentage = fields.Float('Quantity (%)', compute="_compute_quantity_percentage", inverse='_inverse_quantity_percentage', store=True, help='Percentage of the ordered quantity that will automatically be delivered once the milestone is reached.')
 
     sale_line_display_name = fields.Char("Sale Line Display Name", related='sale_line_id.display_name')
     product_uom = fields.Many2one(related="sale_line_id.product_uom")
-    product_uom_qty = fields.Float("Quantity", compute="_compute_product_uom_qty", readonly=False)
+    product_uom_qty = fields.Float("Quantity")
```
Xavier-Do: Looks like the field was computed and is now stored you cannot do that in stable ;)

addons/sale_project/models/project_milestone.py
njor-odoo: @deneuvillem, sorry for the late reply, I was on leave.

 I have updated the code to remove product_uom_qty from the dependencies of _compute_quantity_percentage to avoid the unnecessary recomputations.

deneuvillem: Thanks for making the changes @njor-odoo but do we really need to add the `_onchange_product_uom_qty()`? 
I was thinking that just removing the dependency would do the trick (to be tested). What do you think? :)

njor-odoo: Hello @deneuvillem, i think `_onchange_product_uom_qty` is still needed because when `product_uom_qty changes`, we need to recompute `quantity_percentage`.

If we just remove the dependency without adding the onchange, `quantity_percentage` won’t be updated when `product_uom_qty` changes, which can lead to inconsistent values.

deneuvillem: Okay, I see, we will already merge your fix, and we'll think about optimising this part if we have some reported issues I guess

deneuvillem:
Thanks for your fix! :rocket: 

robodoo r+

PR #152737, opened 2024-02-05
[FIX] calendar: delete activity with event causes error

maki-odoo:
Before this PR:
Canceling an activity linked to an event produces error because of non provided 
thread.

Steps to reproduce:
- Create an activity with event (call or meeting) in chatter
- Try to cancel this activity
- There is an error about undefined thread and activity remains on the chatter

After this PR:
- Thread is provided in `onUpdate` method to use in `load()` in chatter
- `activityService` deletes the activity with event in `unlink` patch

PR #245398, opened 2026-01-23
[IMP] mail: make activity delete action explicit and safer

yagp-odoo:
Activities had a 'Cancel' action that permanently delete the records, leading to unintended data loss.

This change:
- Removes quick delete options from chatter, list, kanban, and activity views
- Adds an explicit Delete button in the activity popup form
- Removes custom added 'Cancel' button from list view header, as users
can still perform mass deletion via list view actions.

Task-5468806

### DIFF
--- a/addons/calendar/static/src/core/web/activity_patch.js
+++ b/addons/calendar/static/src/core/web/activity_patch.js
@@ -10,17 +10,4 @@ patch(Activity.prototype, {
     async onClickReschedule() {
         await this.props.activity.rescheduleMeeting();
     },
-    /**
-     * @override
-     */
-    async unlink() {
-        if (this.props.activity.calendar_event_id) {
-            const thread = this.thread;
-            this.props.activity.remove();
-            await this.orm.call("mail.activity", "unlink_w_meeting", [[this.props.activity.id]]);
-            this.props.onActivityChanged(thread);
-        } else {
-            super.unlink();
-        }
-    },
 });
--- a/addons/mail/models/mail_activity.py
+++ b/addons/mail/models/mail_activity.py
@@ -709,11 +709,6 @@ class MailActivity(models.Model):
     def action_reschedule_nextweek(self):
         self.filtered('active').date_deadline = date.today() + relativedelta(weeks=1, weekday=MO(-1))
 
-    def action_cancel(self):
-        for activity in self:
-            if activity.active:
-                activity.unlink()
-
     @api.readonly
     def activity_format(self):
         return Store().add(self, "_store_activity_fields").get_result()
--- a/addons/mail/static/src/views/web/list/archive_disabled_list_controller.js
+++ b/addons/mail/static/src/views/web/list/archive_disabled_list_controller.js
@@ -1,3 +1,4 @@
+import { _t } from "@web/core/l10n/translation";
 import { useService } from "@web/core/utils/hooks";
 import { ListController } from "@web/views/list/list_controller";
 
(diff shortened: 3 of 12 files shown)


addons/calendar/static/src/core/web/activity_patch.js
```diff
@@ -14,13 +14,8 @@ patch(Activity.prototype, {
      * @override
      */
     async unlink() {
-        if (this.props.activity.calendar_event_id) {
-            const thread = this.thread;
```
mrsr-odoo, line 18: I think it was stored locally on purpose
see - https://github.com/odoo/odoo/pull/152737#discussion_r1479758595

addons/mail/models/mail_activity.py
```diff
@@ -618,7 +618,7 @@ def action_reschedule_tomorrow(self):
     def action_reschedule_nextweek(self):
         self.filtered('active').date_deadline = date.today() + relativedelta(weeks=1, weekday=MO(-1))
 
-    def action_cancel(self):
+    def action_delete(self):
         for activity in self:
             if activity.active:
```
mrsr-odoo: shouldn't the delete feature work with done activities?
it feels bit odd when we have confirmation box yet they are not deleted
TBD with PO
maybe we can directly use `unlink`

mrsr-odoo (commented):
Hello @yagp-odoo ,
Overall LGTM 
small comment about done activities
Thanks

addons/mail/models/mail_activity.py
yagp-odoo: Checked with PO, we can indeed delete the done activities. Also, in the Actions menu, we still have the  default delete option, so we don't need that extra button as well
image

addons/mail/static/src/views/web/list/archive_disabled_list_controller.js
```diff
@@ -8,6 +9,17 @@ export class ArchiveDisabledListController extends ListController {
         this.store = useService("mail.store");
     }
 
+    get deleteConfirmationDialogProps() {
+        return {
+            title: _t("Delete Activities"),
+            body: _t(
+                "Are you sure you want to delete these activities? It will be gone forever!\nThink twice before you click that 'Delete' button!"
```
awa-odoo: They will*

addons/calendar/static/src/core/web/activity_patch.js
```diff
@@ -14,13 +14,9 @@ patch(Activity.prototype, {
...
-            this.props.activity.remove();
-            await this.orm.call("mail.activity", "unlink_w_meeting", [[this.props.activity.id]]);
-            this.props.onActivityChanged(thread);
-        } else {
-            super.unlink();
-        }
+        const thread = this.thread;
```
awa-odoo: Was the previous code an issue with the new implementation we want?

addons/calendar/static/src/core/web/activity_patch.xml
```diff
@@ -9,6 +9,9 @@
             <button t-if="props.activity.calendar_event_id" class="btn btn-link p-0 me-3" t-on-click="onClickReschedule">
                 <i class="fa fa-calendar"/> Reschedule
             </button>
+            <button t-if="props.activity.calendar_event_id" class="btn btn-link btn-danger p-0" t-on-click="unlink">
```
awa-odoo: Not sure why this is in calendar and not in the base template? Also not sure why it's related to it having a "calendar_event_id"?

awa-odoo (commented):
Hello @yagp-odoo 

Had a quick look at this one and left some comments.

Thanks & cheers!

addons/calendar/static/src/core/web/activity_patch.js
yagp-odoo: ~~There is no `super.unlink()` since we removed the delete button for all the activities except calendar meeting, so I adjusted the method accordingly.~~

addons/calendar/static/src/core/web/activity_patch.xml
yagp-odoo: ~~We don't have a standard edit form when the activity is a 'Meeting' (calendar event), so in order to remove it without switching to the calendar app, the delete button is added here~~

After further discussion, it's decided we'll remove this as well.

PR #245400, opened 2026-01-23
[FIX] pos_loyalty: add test for processing loyalty draft/cancel order

aksi-odoo:
Step to reproduce:
- have a trusted pos and a loyalty program which gives points per $ spent
- start pos and select order and a partner (he should already have some LPs)
- notice the loyalty points assgined
- save the order, you are redirected to new order
- switch back to original order

Observation:
- Notice, the loyalty points are reassigned for example :
- if initially partner's LP = 50,  product added is 100$ , LP = 50+100 = 150
- After saving, LP becomes 150 + 100 = 250

Cause:
- LP's are processed after every `sync_from_ui` call from `_postProcessLoyalty` which updates the customer's lp, even before the order is fullfilled or when order is still in `draft` state

Fix:
- Issue is fixed in https://github.com/odoo/odoo/commit/a4b37ec474656c7af23d0134251d589e9a6a61ca
- This commit adds related test for the fix
    

opw-5609964

---

### DIFF
--- a/addons/point_of_sale/static/tests/tours/utils/product_screen_util.js
+++ b/addons/point_of_sale/static/tests/tours/utils/product_screen_util.js
@@ -729,3 +729,14 @@ export function addDiscount(discount) {
             .flatMap((key) => Numpad.click(key)),
     ].flat();
 }
+
+export function saveOrder() {
+    return [
+        clickReview(),
+        {
+            content: "save order",
+            trigger: ".pads .fa-upload",
+            run: "click",
+        },
+    ];
+}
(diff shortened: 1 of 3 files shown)


alsh-odoo (approved):
Thanks for the PR! 

 I was wondering if it might make sense to merge both PRs into a single one (possibly as two separate commits or may be  bug 1 & bug 2 in a single commit), since the issues are indirectly related.

Reasoning

The issue only occurs when we have a trusted device, because the “Save order” option is visible only when the order is shareable. Given this dependency, handling both fixes together could improve clarity and traceability.

Additional benefit

By merging them:
- We can avoid writing a separate tour just for this case
- Instead, we can extend the existing tour with a few additional steps to cover this scenario

I’ll leave it to @sbel-odoo to decide which approach she prefers. 🙂

loukwets (approved):
Nice fix !

stva-odoo:
robodoo r+

PR #245403, opened 2026-01-23
[IMP] account: update due date color and adjust tests

jasa-odoo:
With this commit, the invoice due date is displayed in red in the form view
once it has passed.

Tests have been updated accordingly. When the `account_followup` module is
installed and a user performs Send & Print twice on an overdue invoice,
the second email is treated as a reminder. This causes the test to fail due
to the additional attachment of the open items report.

To avoid this, a context is passed while creating the wizard the second time
to skip the invoice reminder.

task-5241209

ent: https://github.com/odoo/enterprise/pull/105264
upg: https://github.com/odoo/upgrade/pull/9587

### DIFF
--- a/addons/account/tests/test_account_move_send.py
+++ b/addons/account/tests/test_account_move_send.py
@@ -512,7 +512,7 @@ class TestAccountMoveSendCommon(AccountTestInvoicingCommon):
                 {k: v for k, v in expected_values.items() if not check_id_needed and k != 'id'},
             )
 
-    def create_send_and_print(self, invoices, default=False, *, as_user=None, **kwargs):
+    def create_send_and_print(self, invoices, default=False, *, as_user=None, no_invoice_reminder=False, **kwargs):
         invoices.with_user(as_user).action_send_and_print()
         if len(invoices) == 1:
             if not default and not kwargs.get('sending_methods'):
@@ -520,7 +520,7 @@ class TestAccountMoveSendCommon(AccountTestInvoicingCommon):
                 # Therefore by default we deactivate sending methods, unless default parameter is set to True,
                 # or they are explicitly given.
                 kwargs['sending_methods'] = []
-            return self._create_account_move_send_wizard_single(invoices, as_user=as_user, **kwargs)
+            return self._create_account_move_send_wizard_single(invoices, as_user=as_user, no_invoice_reminder=no_invoice_reminder, **kwargs)
         else:
             return self._create_account_move_send_wizard_multi(invoices, as_user=as_user, **kwargs)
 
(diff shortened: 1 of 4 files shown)


addons/account/tests/test_account_move_send.py
```diff
@@ -572,6 +572,9 @@ def test_invoice_single(self):
         self.assertEqual(len(invoice_attachments), 1)
         self.assertTrue(self._get_mail_message(invoice))
 
+        if self.env['ir.module.module']._get('account_followup').state == 'installed':
+            self.skipTest("Following test will fail if 'account_followup' is installed")
+
```
svfu-odoo: It is not the most important; but I think we now only run enterprise on the runbot.
Maybe we could just add a context key to check in `account_followup` to not do a reminder?
Or move the `is_reminder_mail` to community so that it can be patched?

addons/account/views/account_move_views.xml
```diff
@@ -1125,7 +1125,8 @@
                                 <div class="d-flex" invisible="move_type not in ('out_invoice', 'out_refund', 'in_invoice', 'in_refund', 'out_receipt', 'in_receipt')" name="due_date">
                                     <field name="invoice_date_due" force_save="1"
                                            placeholder="Date"
-                                           invisible="invoice_payment_term_id"/>
+                                           invisible="invoice_payment_term_id"
```
svfu-odoo, line 1142: It is more of a functional question but maybe we should do the same in case the Payment Terms are shown instead?

addons/account/tests/test_account_move_send.py
```diff
@@ -842,23 +855,24 @@ def test_invoice_mail_attachments_widget(self):
...
         # Resend.
         wizard = self.create_send_and_print(invoice, sending_methods=['email'])
-        pdf_report_values['id'] = invoice.invoice_pdf_report_id.id
-        self._assert_mail_attachments_widget(wizard, [
-            pdf_report_values,
-            extra_attachment_values,
-        ])
+        # If the follow-up module is installed, a second Send & Print is handled as a reminder:
+        # it adds the open items report as an attachment and switches the email template.
+        if followup_installed:
```
svfu-odoo: It is not the most important. (Can be done at the end)
But I think we now only run enterprise on the runbot.
I think it would be better to make the tests here test the case that `account_followup` is not installed.
And create a proper test in `account_followup` for the reminder part.

Maybe we could just add a context key to check in account_followup to not do a reminder?
Or move the `is_reminder_mail` to community so that it can be patched? (I guess that is the "problem")

svfu-odoo (commented):
It seems like in the spec / the main part is in enterprise.

addons/account/views/account_move_views.xml
svfu-odoo: I checked with GUHEC and he said we do not need it, sorry

addons/account/tests/test_account_move_send.py
jasa-odoo: i think context is better option

addons/account/views/account_move_views.xml
```diff
@@ -1125,7 +1125,8 @@
...
+                                           decoration-danger="invoice_date_due and invoice_date_due &lt; context_today().strftime('%Y-%m-%d') and payment_state not in ('paid', 'in_payment', 'reversed')"/>
```
svfu-odoo, line 1143: Can we also have it for the Journal Items like QDP suggested in the chatter on the task?

svfu-odoo (commented):
I think the commit / PR message is out of date.
Otherwise LGTM

svfu-odoo (approved):

svfu-odoo:
@robodoo r+

svfu-odoo:
@robodoo r+

PR #245405, opened 2026-01-23
[FIX] web: prevent list header border disappear in Firefox

sauc-odoo:
In Firefox, the separator below the list header could disappear 
at certain zoom levels when using sticky headers.

After this commit, the list header border will not disappear.

Task-5249221

---


omra-odoo (approved):

aab-odoo (approved):
robodoo r+

PR #235832, opened 2025-11-14
[IMP] orm: fields.Binary stores bytes

kmagusiak:
Description of the issue/feature this PR addresses:
The field is used to sometimes some mutable data because the `convert_to_cache` does not force a particular type. That hack is used in multiple places where instead json files (or another form of cache) could be used.

Current behavior before PR:
```py
res.binary_field = val = {}
res.binary_field["ok"] = 4
assert "ok" in val
```

Desired behavior after PR is merged:
```py
res.binary_field = {}  # fail because not bytes (or like bytes)
```

https://github.com/odoo/enterprise/pull/99567
task-4251301

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245409, opened 2026-01-23
[FIX] l10n_tr_nilvera_einvoice: Change field type from Binary to Json

sham-odoo:
After a recent PR, #235832 the `fields.Binary` can not store data other than bytes, due to this a traceback occurs while trying to change the Invoice Type, since a compute method is triggered that computes domain for a field that is of type Binary.

Error: `TypeError: 'str' object cannot be interpreted as an integer`

This commit fixes that issue by changing field type from Binary to Json.

---


kmagusiak (approved):
@robodoo r+

Seems it got merged very recently and I did not see it.

PR #245411, opened 2026-01-23
[FIX] sale_edi_ubl, purchase_edi_ubl_bis3: Fix order import with line-level charges

wopy-odoo:
When importing orders with line-level charges, the import would fail because the `quantity` key was not being renamed to correct quantity keys (`'product_uom_qty'` for sale order and `'product_qty'` for purchase order) for allowance charge lines.

The issue occurred because the key renaming was only applied to the main lines_vals list after calling `_import_lines()`, but not to the `allowance_charges_line_vals` that were created separately.

Also the key for taxes returned in `_get_line_vals_list` method in `purchase.order` model is `'taxes_id'` which is not aligned with the `purchase.order.line` model.

This fix properly handles the quantity key renaming for both regular lines and allowance charge lines and updates `'taxes_id'` key value to `'tax_ids'`, preventing the AttributeError when accessing the quantity key during order line creation.

### DIFF
--- a/addons/test_sale_purchase_edi_ubl/tests/test_order_ubl_bis3.py
+++ b/addons/test_sale_purchase_edi_ubl/tests/test_order_ubl_bis3.py
@@ -307,3 +307,36 @@ class TestOrderEdiUbl(TestAccountEdiUblCii, SaleCommon):
         po = self.env['purchase.order'].with_context(default_partner_id=self.env.user.partner_id.id)._create_records_from_attachments(xml_attachment)
         # Should have same payment term as PO
         self.assertEqual(po.payment_term_id, payment_term)
+
+    def test_so_import_line_allowance_charges(self):
+        po_line_vals = [{
+            'product_id': self.place_prdct.id,
+            'price_unit': 30.0,
+            'product_uom_id': self.uom_units.id,
+            'product_qty': 10.0,
+            'tax_ids': self.purchase_tax.ids,
+            'discount': -10.0,
+        }]
+        xml_attachment = self.get_purchase_xml(po_line_vals)
+
+        so = self.env['sale.order'].with_context(default_partner_id=self.env.user.partner_id.id)._create_records_from_attachments(xml_attachment)
+        self.assertEqual(len(so.order_line), 2)
+        allowance_lines = so.order_line.filtered(lambda l: not l.product_id)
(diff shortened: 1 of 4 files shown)


addons/test_sale_purchase_edi_ubl/tests/test_order_ubl_bis3.py
```diff
@@ -307,3 +309,70 @@ def test_po_import_payment_terms(self):
...
+                    <Amount currencyID="EUR">20</Amount>
+                </AllowanceCharge>
+                <AllowanceCharge>
+                    <ChargeIndicator>true</ChargeIndicator>
+                    <AllowanceChargeReason>FUEL SURCHARGE</AllowanceChargeReason>
+                    <Amount currencyID="EUR">10</Amount>
+                </AllowanceCharge>
+            </xpath>"""
+        xml_with_allowance = self.with_applied_xpath(xml_root, modifying_xpath)
+        xml_attachment.raw = etree.tostring(xml_with_allowance)
```
Sehnde:
```suggestion
        po_line_vals = [{
            'product_id': self.place_prdct.id,
            'price_unit': 30.0,
            'product_uom_id': self.uom_units.id,
            'product_qty': 10.0,
            'tax_ids': self.purchase_tax.ids,
            'discount': -10.0,
        }]
        xml_attachment = self.get_purchase_xml(po_line_vals)
```
You can trigger the same error by using a negative discount. This is closer to a user flow than just adding to the raw xml. Even better would be to use the same flow as the ticket if there's one. Same for the other test.

Sehnde (changes requested):
Is there a ticket for this issue ? If yes, please indicate the ticket number in your commit and PR opening message.

addons/test_sale_purchase_edi_ubl/tests/test_order_ubl_bis3.py
wopy-odoo: There is no ticket related to this issue. I came across the issue in a different task when importing an external order document (not generated by Odoo, but still a compatible `'urn:fdc:peppol.eu:poacc:trns:order:3'` document). I have updated the code to generate the XML based on negative discount as you advised. Thank you!

Sehnde (approved):
LGTM, thanks

PR #245414, opened 2026-01-23
[FIX] web: ensure kanban <main> attributes are preserved during compilation

omra-odoo:
Before this commit:
since commit https://github.com/odoo/odoo/commit/3ad0b49b646090e51c4d36bc25480000b3bfd5ad kanban compiler is not preserving its runtime owl directives defined on the `<main>` element, such as t-att-class,t-if,etc.

After this commit:
The compiler now preserves all attributes from the original `<main>` node before converting to the 'o_record_main', restoring correct runtime behavior while keeping the expected kanban structure.

task-5870330

---

### DIFF
--- a/addons/web/static/src/views/kanban/kanban_compiler.js
+++ b/addons/web/static/src/views/kanban/kanban_compiler.js
@@ -101,7 +101,10 @@ export class KanbanCompiler extends ViewCompiler {
     compileMain(el, params) {
         const compiled = createElement("div");
 
-        compiled.setAttribute("class", `o_record_main`);
+        for (const { name, value } of el.attributes) {
+            compiled.setAttribute(name, value);
+        }
+        combineAttributes(compiled, "class", ["o_record_main"]);
         for (const child of el.childNodes) {
             append(compiled, this.compileNode(child, params));
         }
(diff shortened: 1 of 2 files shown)


addons/web/static/src/views/kanban/kanban_compiler.js
aab-odoo, line 111:
```suggestion
        return assignOwlDirectives(compiled, el);
```

aab-odoo: What about this? I didn't test it, but it looks like it does exactly what you want, with proper handling of `t-attf-` attributes

aab-odoo (commented):
Nice catch @omra-odoo 👍 

Could you write a test case as well? Thanks

addons/web/static/src/views/kanban/kanban_compiler.js
omra-odoo: it's not working it just copy pasting values of t-att-class into class without evaluting expression. like
`<div class="['1_done', '1_canceled'].includes(record.state.raw_value) ? 'opacity-50' : ''">`

aab-odoo: Ha indeed, I read to fast. It's to copy attributes on Component nodes (which then become props).

aab-odoo (approved):
robodoo r+

Thanks @omra-odoo

PR #245421, opened 2026-01-23
[FIX] im_livechat: quick action non-deterministic test

tsm-odoo:
The `Only two quick actions are shown` live chat test awaits the first posted message because to avoid conflict with the reaction being added afterwards.

To do so, the test uses the `waitForSteps` helper. However, during high load, 200ms might be too short to receive the notification, leading to the test failing.

This commit increases the timeout.

runbot-237587

Current behavior before PR:

---


alexkuhn (approved):
@robodoo delegate+ r+

PR #245423, opened 2026-01-23
[IMP] hr_attendance: improve overtime ruleset duplication

rmun-odoo:
Add the “(copy)” suffix when duplicating overtime rulesets and cloned all associated rules to copied ruleset.

task-5705117

---

### DIFF
--- a/addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
+++ b/addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
@@ -9,7 +9,7 @@ class HrAttendanceOvertimeRuleset(models.Model):
 
     name = fields.Char(required=True)
     description = fields.Html()
-    rule_ids = fields.One2many('hr.attendance.overtime.rule', 'ruleset_id')
+    rule_ids = fields.One2many('hr.attendance.overtime.rule', 'ruleset_id', copy=True)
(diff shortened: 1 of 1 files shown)


addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
```diff
@@ -78,3 +78,8 @@ def action_show_versions(self):
         action['domain'] = self._get_current_versions_domain()
         action['context'] = {'default_ruleset_id': self.id}
         return action
+
+    def copy_data(self, default=None):
+        default = dict(default or {})
```
mepe-odoo: I think this line can be removed :)

addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
```diff
@@ -78,3 +78,8 @@ def action_show_versions(self):
...
+        vals_list = super().copy_data(default=default)
+        return [dict(vals, name=self.env._("%s (copy)", line.name)) for line, vals in zip(self, vals_list)]
```
mepe-odoo:
```suggestion
        return [dict(vals, name=self.env._("%s (copy)", ruleset.name)) for ruleset, vals in zip(self, vals_list)]
```
I think line is not really a good naming as it's really general ruleset seems more appropriate :)

mepe-odoo (changes requested):

PR #245427, opened 2026-01-23
[IMP] point_of_sale: setting to install pos dpopay module

vpan-odoo:
In this commit:
===============
- We added one field in the `res_config_setting` to install the `pos_dpopay` module.

Task: 5502977

---


davidmonnom (changes requested):
You must add the same message as when activating pos_restaurant.
`Save this page and come back here to set up the feature.`

Grep in the code to see examples.

vpan-odoo:
> You must add the same message as when activating pos_restaurant. `Save this page and come back here to set up the feature.`
> 
> Grep in the code to see examples.

Hello @davidmonnom,

For the setup of payment terminals, we use an internal link that redirects to the payment method form view, where the configuration is done. This keeps the same behavior for all payment terminal fields (see image 1).

The message `Save this page and come back here to set up the feature.` is more suitable for settings where the configuration steps are done on the same page (see image 2 and image 3).

In this case, we should follow the same behaviour for all the payment terminals wdyt?

| Image 1 | Image 2 | Image 3 |
|--------|--------|--------|
| image | image | image |

davidmonnom (approved):
@robodoo r+

PR #245431, opened 2026-01-23
[FIX] hr_recruitment: prevent duplicate 'Applicant created' log

ramezlahzy:
Steps to reproduce:
1. Create a new applicant in recruitment.
2. Open the applicant’s chatter.
3. See multiple “Applicant created” messages for the same creation.

Bug cause:
The applicant creation flow posts the `mt_applicant_new` subtype more than once (create + extra write/track), and the frontend renders the subtype description, so each duplicate post shows “Applicant created” again.

Solution:
- Post the `mt_applicant_new` subtype only once during applicant creation.
- Avoid re-posting it in subsequent writes/tracking so chatter shows a single creation log.

Task Id: 5454691

### DIFF
--- a/addons/hr_recruitment/models/hr_applicant.py
+++ b/addons/hr_recruitment/models/hr_applicant.py
@@ -1115,3 +1115,9 @@ class HrApplicant(models.Model):
             if applicant.refuse_reason_id and applicant.refuse_date:
                 json[applicant.stage_id.id] -= (now - applicant.refuse_date).total_seconds()
         return json
+
+    def _creation_message(self):
+        self.ensure_one()
+        if self.is_pool_applicant:
+            return self.env._("Talent created")
+        return super()._creation_message()
--- a/addons/hr_recruitment/data/mail_message_subtype_data.xml
+++ b/addons/hr_recruitment/data/mail_message_subtype_data.xml
@@ -7,7 +7,6 @@
         <field name="res_model">hr.applicant</field>
         <field name="default" eval="False"/>
         <field name="hidden" eval="True"/>
-        <field name="description">Applicant created</field>
     </record>
     <record id="mt_applicant_stage_changed" model="mail.message.subtype">
         <field name="name">Stage Changed</field>
@@ -26,7 +25,6 @@
         <field name="name">New Talent</field>
         <field name="res_model">hr.applicant</field>
         <field name="default" eval="False"/>
-        <field name="description">Talent created</field>
     </record>
 
     <!-- Job-related subtypes for messaging / Chatter -->


Einar-eian (changes requested):
Thank you for your work! 😄  So this only partially fixes the issue, as you will still get a double message when you create a talent (go to a talent pool and click new and create one there), except it says `Talent created Applicant created` (see attached picture). 
This double message seems to come from the the method `_creation_message` from `mail.thread`, which always adds a `${model name} created` message when a record is created.  Looking through the codebase I see that there are already a few overrides of this method for some models so I would recommend to either override it to return nothing or remove `description` from `mt_applicant_new` and `mt_talent_new` and have `_creation_message` return the appropriate message depending on if an applicant is an applicant or talent (see `_creation_subtype` in `hr_applicant.py` for inspiration).
There might also be other ways to achieve your goal so if you find a better way feel free to do that 😄 


Image

addons/hr_recruitment/models/hr_applicant.py
```diff
@@ -1115,3 +1115,9 @@ def _get_duration_from_tracking(self, trackings):
             if applicant.refuse_reason_id and applicant.refuse_date:
                 json[applicant.stage_id.id] -= (now - applicant.refuse_date).total_seconds()
         return json
+
+    def _creation_message(self):
+        self.ensure_one()
+        if self.is_pool_applicant:
+            return _("Talent created")
```
Einar-eian: Small detail, we prefer `self.env._()` over `_()`
```suggestion
            return self.env._("Talent created")
```

addons/hr_recruitment/models/hr_applicant.py
```diff
@@ -1115,3 +1115,9 @@ def _get_duration_from_tracking(self, trackings):
...
+        return _("Applicant created")
```
Einar-eian: Same here, but I'm also thinking here you could also just return the super value? Then this method has at least some connection to base method and it will be more obvious if the base method gets modified and not this one, what do you think?

So basically just return this, as the value will be the same (for now)
```suggestion
        return super()._creation_message()
```

Einar-eian (changes requested):
Thank you for your work again! 😄 I noticed some small improvements you can make but otherwise looking good!

addons/hr_recruitment/models/hr_applicant.py
ramezlahzy: done 😊

Einar-eian (approved):
Thank you for your work, LGTM 🚀 
@robodoo r+

PR #245433, opened 2026-01-23
[FIX] html_editor: focus editable after deleting column

adsa-odoo:

When a user deletes a row or column from table menu, the editor loses focus. As a result, actions like Undo do not behave as expected and require multiple attempts to restore the original table state. This breaks the editing flow, causes confusion when performing table-related actions.

**Desired behavior after PR:**

This PR ensures that editable is focused after deleting row or column from table menu. This commit also makes sure that selection is set properly and hint is visible on empty cell after deleting the column.

task-5725593

---

### DIFF
--- a/addons/html_editor/static/src/main/table/table_plugin.js
+++ b/addons/html_editor/static/src/main/table/table_plugin.js
@@ -275,9 +275,8 @@ export class TablePlugin extends Plugin {
         const index = cells.findIndex((td) => td === cell);
         const siblingCell = cells[index - 1] || cells[index + 1];
         table.querySelectorAll(`tr td:nth-of-type(${index + 1})`).forEach((td) => td.remove());
-        // not sure we should move the cursor?
         siblingCell
-            ? this.dependencies.selection.setCursorStart(siblingCell)
+            ? this.dependencies.selection.setCursorEnd(lastLeaf(siblingCell))
             : this.deleteTable(table);
     }
     /**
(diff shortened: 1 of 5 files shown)


addons/html_editor/static/src/main/table/table_plugin.js
```diff
@@ -275,9 +275,8 @@ export class TablePlugin extends Plugin {
         const index = cells.findIndex((td) => td === cell);
         const siblingCell = cells[index - 1] || cells[index + 1];
         table.querySelectorAll(`tr td:nth-of-type(${index + 1})`).forEach((td) => td.remove());
-        // not sure we should move the cursor?
         siblingCell
-            ? this.dependencies.selection.setCursorStart(siblingCell)
+            ? this.dependencies.selection.setCursorEnd(lastLeaf(siblingCell))
```
deso-odoo, line 279: `setCursorEnd` uses `setSelection` and by default `normalize` is true for setSelection, so I expected the use of `lastLeaf/firstLeaf` would not be needed.

adsa-odoo: It is expected that `setSelection` doesn't normalize such cases. See this [test](https://github.com/odoo/odoo/blob/18.0/addons/html_editor/static/tests/delete/backward.test.js#L2248-L2261). Here if `<div>[<h1>abc</h1><p>def</p>]</div>` is normalized to `<div><h1>[abc</h1><p>def]</p></div>` by `setSelection` then `<h1>` will not be deleted after `deleteBackward`.

deso-odoo: For final reviewer: the above msg is more of an answer to: https://discord.com/channels/1024949713759375421/1174681835876388874/1466748068706652375

PR #245443, opened 2026-01-23
[FIX] hr_holidays: fix the scrollbar

djal-odoo:
Before:
the scrollbar of the kanban view in timeoff was not showing coz of which users
were not able to see their timeoffs easily

After:
Fixed the scrollbar of the kanban view so that the user can be able to see their
timeoffs which they were not able to do that easily

Fix:
Added the `overflow-x` as auto so that the scrollbar is visible.

Task:5502863

---


Mahmoudk3m (approved):
LGTM 🚀

amah-odoo (approved):
@robodoo r+

PR #245444, opened 2026-01-23
[18.0][FIX] product: Pricelist rule is not applied on variant

ypapouin:
- A pricelist rule never apply to a product variant, always its template when edited in "expanded" mode
- This is because the `applied_on` field is missing on the `product.pricelist.item` form view.

Current behavior before PR:
- Install `sale_management`
- Enable "Pricelist" setting
- Go to Sales - Products - Pricelists and create a new "Sample" Pricelist.
- Create 3 rules for the same product (template), one with price at 6666, a second one at 6667 and the last one at 3333 image
- Next step, edit 2 first rules for price 6666 and 6667 to set a variant **BUT before starting editing, open the form in expanded mode using two-arrows button** image
- Set the variant to the first price rule (as you can see, the rule name is immediately updated with "Variant: [REF] Product name" image
- Save the price rule (_note that the display name incorrectly resets to default_) image
- Same thing for second rule  image image
- Create a new sale quotation with pricelist set to our "Sample"
- Add product [6666] image
- Add product [6667] image
- Incorrect prices in sale order image

- Each variant must have its own price.
- The added test simply edit the pricelist item using its own form.
- The `applied_on` field is added to the pricelist item form to ensure that its value is saved.
- If we edit the rule without expanded mode, it works fine only because the `applied_on` field exists in the list view as an invisible column.

_Note 1_: `applied_on` is updated by `_onchange_rule_content` using `update` instead of `write` (for caching ?).
But if the `applied_on` field is not on the form, then its value is never sent to low level `_write_multi`.

_Note 2_: to clear the field value with the Form test component, we must set `=self.env["product.product"]` instead of `=False` ....

---

### DIFF
--- a/addons/product/tests/test_pricelist.py
+++ b/addons/product/tests/test_pricelist.py
@@ -2,13 +2,13 @@
 
 from odoo.exceptions import UserError
 from odoo.fields import Command
-from odoo.tests import tagged, Form
+from odoo.tests import Form, tagged
 
-from odoo.addons.product.tests.common import ProductCommon
+from odoo.addons.product.tests.common import ProductVariantsCommon
 
 
 @tagged('post_install', '-at_install')
-class TestPricelist(ProductCommon):
+class TestPricelist(ProductVariantsCommon):
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


Feyensv (approved):
Little push to:
1) avoid using demo data in tests (acoustic bloc screens) and use the right test common instead
2) move the invisible field with the other ones defined in the beginning of the view.

Thanks for the fix :heart:

@robodoo r+

PR #245446, opened 2026-01-23
[FIX] pos_self_order: compute price on backend

macocianradu:
The price computation is done on the backend only for combo products. This means that someone can order stuff on the self order, change the price of the products through JS on the client, and then when they go to pay the backend would generate a payment for the modified price and register everything as normal without any indication to the owner that the price was modified.

This change will force the backend to recompute the price before creating the payment based on the standard prices defined on the backend. This only applies on self orders, so cashiers should still be able to modify prices on the frontend of the normal POS

Task-[5864068](https://www.odoo.com/odoo/project/1737/tasks/5864068)

---

### DIFF
--- a/addons/pos_self_order/controllers/orders.py
+++ b/addons/pos_self_order/controllers/orders.py
@@ -114,6 +114,17 @@ class PosSelfOrderController(http.Controller):
                         'tax_ids': child_product.taxes_id,
                     })
                 lst_price = 0
+            elif not line.combo_parent_id:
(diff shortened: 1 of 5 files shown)


davidmonnom (commented):
We should add tests to compare prices in frontend and backend. To avoid to have price change when syncing.

Check here https://github.com/odoo/odoo/pull/245934

davidmonnom (approved):
@robodoo up to saas-18.4

PR #245451, opened 2026-01-23
[IMP] hr_*: adjust approver update on manager change

HashemKhaled:
*: attendance,expense,holidays

This commit adjusts the update of an employee's {attendance/expense/leave} approvers
upon changing their manager.

The update is performed only if these conditions are met:
- The new manager has a linked user.
- The current attendance approver is the previous manager.

The prevents setting the approver to the new manager when the
current approver is unset.

Task: 4804439


amry-odoo (approved):
LGTM 🚀 🦝

PR #245456, opened 2026-01-23
[IMP] tools: remove vendored Werkzeug urls

mwath:
Since Werkzeug 3.0.0, most utilities in `werkzeug.urls` have been removed in favor of the standard library `urllib.parse`.

This PR removes our vendored version of the module, and replaces all usages of `werkzeug.urls.*` with its equivalent
function from `urllib.parse`.

Here is a breakdown of the migration:
 - `werkzeug.urls.url_encode` -> `urllib.parse.urlencode`
   - Params coming from `urllib.parse.parse_qs` must specify `doseq=True`
     to work.
   - `sort=True` does not exist and must be replaced with `sorted`.

 - `werkzeug.urls.url_decode` -> `urllib.parse.parse_qs`
   - Returns `dict[str, list[str]]` instead of `MultiDict`.
   - Alternatively, use `urllib.parse.parse_qsl`

 - `werkzeug.urls.url_join` -> `odoo.tools.urls.urljoin`
   - If you know what you are doing: `urllib.parse.urljoin`

 - `werkzeug.urls.url_parse` -> `urllib.parse.urlparse`
   - Returns `ParseResult` instead of `URL`.
   - `URL.host` -> `ParseResult.hostname`
   - `URL.replace(...)` -> `ParseResult._replace(...)`
   - `URL.to_url()` -> `ParseResult.geturl()` or `urllib.parse.urlunparse(ParseResult)`
   - `URL.decode_query()` -> `urllib.parse.parse_qs(ParseResult.query)`

 - `werkzeug.urls.url_quote` -> `urllib.parse.quote`
   - Does not allow giving `unsafe='...'`, but it is usually not needed

 - `werkzeug.urls.url_unquote` -> `urllib.parse.unquote`
 - `werkzeug.urls.url_quote_plus` -> `urllib.parse.quote_plus`
 - `werkzeug.urls.url_unquote_plus` -> `urllib.parse.unquote_plus`
 - `werkzeug.urls.url_unparse` -> `urllib.parse.urlunsplit`
   - Alternatively, use `urllib.parse.urlunparse`

Enterprise: odoo/enterprise#105297

opw-5418150

---


addons/payment_buckaroo/models/payment_provider.py
```diff
@@ -1,9 +1,8 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+import urllib.parse
```
livaios:
```suggestion
from urllib.parse import unquote_plus
```

addons/payment_buckaroo/models/payment_provider.py
```diff
@@ -76,7 +75,7 @@ def _buckaroo_generate_digital_sign(self, values, incoming=True):
             # Incoming communication values must be URL-decoded before checking the signature. The
             # key 'brq_signature' must be ignored.
             items = [
-                (k, urls.url_unquote_plus(v)) for k, v in values.items()
+                (k, urllib.parse.unquote_plus(v)) for k, v in values.items()
```
livaios:
```suggestion
                (k, unquote_plus(v)) for k, v in values.items()
```

addons/portal/controllers/portal.py
```diff
@@ -3,8 +3,9 @@
 import json
 import math
 import re
+import urllib.parse
```
livaios:
```suggestion
```

addons/portal/controllers/portal.py
```diff
@@ -3,8 +3,9 @@
...
+from urllib.parse import urlencode, urlparse, parse_qsl
```
livaios:
```suggestion
from urllib.parse import urlencode, urlparse, parse_qsl, quote
```

addons/portal/controllers/portal.py
```diff
@@ -934,7 +935,7 @@ def deactivate_account(self, validation, password, **post):
                 request.env['res.users']._check_credentials(credential, {'interactive': True})
                 request.env.user.sudo()._deactivate_portal_user(**post)
                 request.session.logout()
-                return request.redirect('/web/login?message=%s' % urls.url_quote(_('Account deleted!')))
+                return request.redirect('/web/login?message=%s' % urllib.parse.quote(_('Account deleted!')))
```
livaios:
```suggestion
                return request.redirect('/web/login?message=%s' % quote(_('Account deleted!')))
```

addons/payment_iyzico/models/payment_transaction.py
```diff
@@ -44,8 +43,8 @@ def _get_specific_rendering_values(self, *args):
 
         # Extract the payment link URL and params and embed them in the redirect form.
         api_url = payment_link_data['paymentPageUrl']
-        parsed_url = urls.url_parse(api_url)
-        url_params = urls.url_decode(parsed_url.query)
+        parsed_url = urlparse(api_url)
+        url_params = parse_qs(parsed_url.query)
```
livaios: `parse_qs` doesn't have the same output type as `url_decode`
iyzico payment fails
Here is the different output types between the 2.
Screenshot 2026-01-23 at 13 58 30
expected url:
```
https://sandbox-cpp.iyzipay.com/?token=8d95d5e6-75f5-4d27-8293-5052a70b7a63&lang=en
```

current url:
```
https://sandbox-cpp.iyzipay.com/?token=%5B%278d95d5e6-75f5-4d27-8293-5052a70b7a63%27%5D&lang=%5B%27en%27%5D
```

livaios (changes requested):
Hello @mwath,

Regarding payment modules, you replace `url_decode` and `url_parse(example_url).decode_query()` with `parse_qs`.
`parse_qs` doesn't return the same result type as the 2 previous methods, causing certain functionalities to fail. (See Iyzico screenshot)

I didn't test all scenarios but I know payment redirection for `payment_iyzico` redirect to a 404 page now, and I imagine for the rest of the use cases it will be faulty as well.

addons/payment_iyzico/models/payment_transaction.py
mwath: Good catch, thank you!
The absence of `MultiDict` in urllib is the most annoying change, as it fails silently. I'll fix this quickly :+1:

addons/payment_mercado_pago/models/payment_transaction.py
```diff
@@ -45,8 +44,8 @@ def _get_specific_rendering_values(self, processing_values):
         ]
 
         # Extract the payment link URL and params and embed them in the redirect form.
-        parsed_url = url_parse(api_url)
-        url_params = url_decode(parsed_url.query)
+        parsed_url = urlparse(api_url)
+        url_params = {k: v[0] for k, v in parse_qs(parsed_url.query).items()}
```
livaios, line 49: I don't have all the usecases for the providers on top of mind but taking the first value might be risky as its not always guaranteed that its only one value.
I recommend maintaining the same structure as before we can do:
```suggestion
        url_params = MultiDict(parse_qs(parsed_url.query))
```

Another option would be to make the view in `addons/payment_iyzico/views/payment_iyzico_templates.xml` expect a tuple instead of a dict 
```xml
<input type="hidden" t-att-name="param[0]" t-att-value="param[1]"/>
```
and use `parse_qsl` instead
```suggestion
        url_params = parse_qsl(parsed_url.query)
```
but I prefer the first approach since it doesn't change the structure of the xml based on providers using multidicts or not and we do have a general structure that's shared between all providers so best to keep it unified as much as possible

mwath: While reintroducing `MultiDict` might be the simplest solution, I believe it’s a step in the wrong direction for the long term. We are currently aiming to reduce our dependency on Werkzeug, and relying on external data structures increases security risks (c.f. our security training :smile:).

Moreover, `MultiDict` has, IMHO, a flawed implicit design. It acts as a flat `dict` by default, returning only the first element for a key, even if multiple values exists (c.f. [`werkzeug.datastructures.MultiDict`](https://werkzeug.palletsprojects.com/en/stable/datastructures/#werkzeug.datastructures.MultiDict)).
This *hides* potential bugs (or even security issues) from both the developer and the reviewer. Using `parse_qs` or `parse_qsl` is a safer pattern because it forces the developer to acknowledge that a key might contain multiple values.
And the specialized methods are not used in the payments addons (`\b(getlist|setlist|setlistdefault|listvalues|to_dict|poplist|popitemlist)\b`), suggesting that a flat dictionary is sufficient.

In this case, maybe the safer option would be to assert that there is a single value? E.g. if the payment provider returns with two tokens where one is expected, it should be treated as a potential error or a malformed request, not ignored.

livaios: I went ahead and checked the different use cases for mercado_pago, iyzico and mollie (because sometimes we get the url params from their apis), and it looks like the url params shouldn't contain lists and if the existing behavior of `MultDict` was returning the first value anyways then your current approach should be good to go.

Thanks for clarifying 👍

addons/portal/controllers/mail.py
```diff
@@ -48,10 +48,10 @@ def _redirect_to_record(cls, model, res_id, access_token=None, **kwargs):
                         hash = kwargs.get('hash')
                         url = record_action['url']
                         if pid and hash:
-                            url = urls.url_parse(url)
-                            url_params = url.decode_query()
+                            url = urlparse(url)
+                            url_params = parse_qs(url.query)
```
livaios: Shouldn't we use `parse_qsl` here since we're adding tuples in the following line?

mwath: [`dict.update`](https://docs.python.org/3/library/stdtypes.html#dict.update) can take an iterable with key-value pairs, and `urlencode(..., doseq=True)` supports both `str` and `list[str]`.
```python
In [1]: urlencode({"a": "12", "b": ["34"]}, doseq=True)
Out[1]: 'a=12&b=34'
```
 
... buuuut, it doesn't give the result I expected 😅

```python
In [1]: from werkzeug.datastructures import MultiDict
In [2]: d = MultiDict()
In [3]: d.setlist("a", ["a", "b", "c"])
In [4]: d.update([("a", "x"), ("b", "x")])
In [5]: d
Out[5]: MultiDict([('a', 'a'), ('a', 'b'), ('a', 'c'), ('a', 'x'), ('b', 'x')])

In [6]: d["a"] = "x"
In [7]: d
Out[7]: MultiDict([('a', 'x'), ('b', 'x')])
```
Using `parse_qsl` is indeed much easier, thanks ^^

addons/survey/models/survey_survey.py
```diff
@@ -1077,7 +1076,7 @@ def action_send_survey(self):
     def action_start_survey(self, answer=None):
         """ Open the website page with the survey form """
         self.ensure_one()
-        url = '%s?%s' % (self.get_start_url(), werkzeug.urls.url_encode({'answer_token': answer and answer.access_token or None}))
+        url = '%s?%s' % (self.get_start_url(), urlencode({'answer_token': answer and answer.access_token or None}))
```
antonrom1: Afaik one of the behaviour changes between werkzeug and urllib is that werkzeug drops values when value_str is None, but urllib doesn't. This might be a big difference in some cases like here

```
url_encode({'a': None}) == ''
urlencode({'a': None}) == 'a=None'
```

odoo/sql_db.py
```diff
@@ -779,7 +779,7 @@ def connection_info_for(db_or_uri: str, readonly=False) -> tuple[str, dict]:
     app_name = app_name.replace('{pid}', str(os.getpid()))[:63]
     if db_or_uri.startswith(('postgresql://', 'postgres://')):
         # extract db from uri
-        us = urls.url_parse(db_or_uri)  # type: ignore
+        us = parse_url(db_or_uri)  # type: ignore
         if len(us.path) > 1:
```
antonrom1: `parse_url(...).path` might return None here

addons/test_mail_full/tests/test_portal.py
```diff
@@ -353,16 +355,16 @@ def test_redirect_to_records_norecord(self):
...
                 path, '/web/login',
                 'Failed with %s - %s' % (model, res_id)
             )
 
     def assert_URL(self, url, expected_path, expected_fragment_params=None, expected_query=None):
         """Asserts that the URL has the expected path and if set, the expected fragment parameters and query."""
-        parsed_url = url_parse(url)
-        fragment_params = url_decode(parsed_url.fragment)
+        parsed_url = parse_url(url)
+        fragment_params = parse_qs(parsed_url.fragment)
```
antonrom1: Same here, urllib3 differentiates None and empty string, and fragment can be None. Here it will break if it happens

addons/account_edi_proxy_client/models/account_edi_proxy_auth.py
```diff
@@ -36,7 +37,7 @@ def __get_payload(self, request, msg_timestamp):
             msg_timestamp,  # timestamp
             parsed_url.path,  # url path
             self.id_client,
-            json.dumps(werkzeug.urls.url_decode(parsed_url.query), sort_keys=True),  # url query params sorted by key
+            json.dumps(urllib.parse.parse_qs(parsed_url.query), sort_keys=True),  # url query params sorted by key
```
antonrom1: You will break peppol 😁 Screenshot 2026-02-02 at 16 17 23

parse_qs returns a dict of lists, so c14n will change. And this thing builds the "digest" for a couple of EDIs, so signature must be same client & server side

addons/survey/models/survey_survey.py
mwath: Indeed, that's not nice. I was considering monkey patching `urlencode` to prevent the common mistake of forgetting `doseq=True`. With the `None` behaviour, I think it's best if developers are warned about a (maybe) incorrect usage of the lib. And if you really want this behaviour, simply `str(your_not_nice_value)` :)

addons/account_edi_proxy_client/models/account_edi_proxy_auth.py
mwath: Oops
image

addons/test_mail_full/tests/test_portal.py
mwath: Actually, I don't think this function is still used :thinking:

addons/payment/wizards/payment_link_wizard.py
```diff
@@ -1,6 +1,5 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
-
```
AntoineVDV, line 2: Newline removed by mistake.

addons/payment_iyzico/models/payment_transaction.py
```diff
@@ -1,6 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
-
```
AntoineVDV, line 2: Newline removed by mistake.

addons/payment_iyzico/tests/test_payment_transaction.py
```diff
@@ -2,7 +2,7 @@
 
 from unittest.mock import patch
```
AntoineVDV, line 4: Group built-in imports.
```suggestion
```

addons/payment_iyzico/tests/test_processing_flows.py
```diff
@@ -2,7 +2,7 @@
 
 from unittest.mock import patch
```
AntoineVDV, line 4: Group built-in imports.
```suggestion
```

addons/payment_razorpay/models/payment_transaction.py
```diff
@@ -5,7 +5,7 @@
 from datetime import datetime
 
 from dateutil.relativedelta import relativedelta
-from werkzeug.urls import url_encode
+from urllib.parse import urlencode
```
AntoineVDV, line 8: Group built-in imports.

```python
from datetime import datetime
from urllib.parse import urlencode

from dateutil.relativedelta import relativedelta
```

addons/payment_stripe/models/payment_provider.py
```diff
@@ -2,7 +2,8 @@
 
 import json
```
AntoineVDV, line 4: Group built-in imports.
```suggestion
```

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -2,7 +2,7 @@
 
 from unittest.mock import patch
```
AntoineVDV, line 4: Group built-in imports.
```suggestion
```

addons/payment_xendit/models/payment_transaction.py
```diff
@@ -1,6 +1,5 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
-
```
AntoineVDV, line 2: Newline removed by mistake.

addons/payment_xendit/tests/test_payment_transaction.py
```diff
@@ -2,7 +2,7 @@
 
 from unittest.mock import patch
```
AntoineVDV, line 4: Group built-in imports.
```suggestion
```

addons/payment_xendit/tests/test_processing_flows.py
```diff
@@ -3,7 +3,7 @@
 from unittest.mock import patch
 
 from werkzeug.exceptions import Forbidden
-from werkzeug.urls import url_encode
+from urllib.parse import urlencode
```
AntoineVDV, line 6: Group built-in imports.
```python
from urllib.parse import urlencode

from werkzeug.exceptions import Forbidden
```

AntoineVDV (approved):
LGTM for `payment*` modules after fixing the import guidelines 🙂 

Edit: Some of my suggestions might not account for the fact that `urllib3` is actually not a built-in module.

PR #245463, opened 2026-01-23
[FIX] html_editor: prevent removal of unremovable non-editable element children

dhba-odoo:
**Steps to reproduce:**
- Drop a dynamic snippet, for example Recent Blogs → “Latest Stories”.
- Click on the section title.
- Press Ctrl + A.
- Press Delete.

**Description of the issue:**
- The content part is removed even though it is inside a contenteditable false container. doing discard after it crash the page.

**Cause:**
- In `removeNode`, there is no check to prevent removal of children of a contenteditable false element.
- As a result, if a contenteditable false element itself is unremovable, the parent node is preserved, but its children are still removed.

**Solution:**
- In `removeNode`, add a check to return false if node is not `isNodeEditable`.

task-5858259

### DIFF
--- a/addons/html_editor/static/src/core/delete_plugin.js
+++ b/addons/html_editor/static/src/core/delete_plugin.js
@@ -659,10 +659,13 @@ export class DeletePlugin extends Plugin {
             for (const child of [...node.childNodes]) {
                 remove(child);
             }
-            if (this.isUnremovable(node, root)) {
+            if (
+                this.isUnremovable(node, root) ||
+                !this.dependencies.selection.isNodeEditable(node)
+            ) {
                 return false;
             }
-            if (node.hasChildNodes()) {
+            if (node.hasChildNodes() && node.isContentEditable) {
                 node.before(...node.childNodes);
                 node.remove();
                 return false;
--- a/addons/html_editor/static/tests/delete/delete_range.test.js
+++ b/addons/html_editor/static/tests/delete/delete_range.test.js
@@ -156,6 +156,13 @@ describe("deleteRange method", () => {
                 contentAfter: "<div><p>abc[]def</p></div>",
             });
         });
+        test("should remove contenteditable=false element if fully selected", async () => {
+            await testEditor({
+                contentBefore: `<p>a[bc</p><div contenteditable="false">def</div><p>gh]i</p>`,
+                stepFunction: deleteSelection,
+                contentAfter: `<p>a[]i</p>`,
+            });
+        });
     });
     describe("Block + inline", () => {
         test("should merge paragraph with inline content after it", async () => {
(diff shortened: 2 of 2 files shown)


addons/html_editor/static/src/core/delete_plugin.js
```diff
@@ -659,10 +659,13 @@ export class DeletePlugin extends Plugin {
...
             }
-            if (this.isUnremovable(node, root)) {
+            if (
+                this.isUnremovable(node, root) ||
+                !this.dependencies.selection.isNodeEditable(node)
+            ) {
                 return false;
             }
-            if (node.hasChildNodes()) {
+            if (node.hasChildNodes() && node.isContentEditable) {
```
deso-odoo, line 668: I don't get this why at one point you chose to use `isNodeEditable` and at other directly `isContentEditable`?

addons/html_editor/static/tests/delete/delete_range.test.js
```diff
@@ -334,6 +334,13 @@ describe("deleteSelection", () => {
                 contentAfter: `<p>a[]</p><div class="oe_unremovable"><br></div><p>i</p>`,
             });
         });
+        test("should not remove nor clear content of unremovable contenteditable=false node", async () => {
+            await testEditor({
+                contentBefore: `<p>a[bc</p><div class="oe_unremovable" contenteditable="false">def</div><p>gh]i</p>`,
+                stepFunction: deleteSelection,
+                contentAfter: `<p>a[]</p><div class="oe_unremovable" contenteditable="false">def</div><p>i</p>`,
+            });
+        });
```
deso-odoo, line 350: Along side can you please also check if we have a test case for this
```js
await testEditor({
    contentBefore: `<p>a[bc</p><div contenteditable="false">def</div><p>gh]i</p>`,
    stepFunction: deleteSelection,
    contentAfter: `<p>a[]i</p>`,
});
```
and if we don't we should probably add one.

addons/html_editor/static/src/core/delete_plugin.js
dhba-odoo: I need both `isNodeEditable` and `isContentEditable` for the following case:
`<p>a[bc</p><div contenteditable="false">def</div><p>gh]i</p>.`
In this case [here](https://github.com/odoo/odoo/pull/245463/files#diff-aa96600c8e21897ff3ebfe1240c36a1bda0a47d1c234014c51b5007c6c22daabR664) `isNodeEditable` is required. Using `isContentEditable` would cause an early return and prevent a fully selected contenteditable="false" element from being removed.

As for why `isContentEditable` is needed [here](https://github.com/odoo/odoo/pull/245463/files#diff-aa96600c8e21897ff3ebfe1240c36a1bda0a47d1c234014c51b5007c6c22daabR668) because that if block removes only the element, not its children. When the element’s content is not editable, this would result in the element being removed while its children remain

addons/html_editor/static/tests/delete/delete_range.test.js
dhba-odoo: There is a similar test for the banner already available.

dmo-odoo: We just spent some time on this commit trying to figure out why we didn't want to remove contenteditable=false elements and in fact we realized that the fix does not do what we thought it did.

In the commit title there is no mention of "unremovable" which makes it confusing. When you take the title of the commit (and thus the PR) at face value, it seems like you always prevent the removal of contenteditable false children, which would be wrong. Could you update the commit and PR title please?

Also, I agree with @deso-odoo about the missing test. Even though this particular behavior might already be tested somewhere else due to a specific feature having the particular layout to trigger the case, I think that we should have a proper test for it in `delete_range_test`.

dhba-odoo: @dmo-odoo I have updated the commit and PR title, and added the test.

dmo-odoo:
@dhba-odoo Thank you!
@robodoo r+

PR #233472, opened 2025-10-28
[IMP] mrp: split MO while updating qty

abyo-odoo:
This commit adds the possibility to split a MO
from the `Change Quantity to Produce` wizard
while updating the quantity to produce.

This offers more flexibility to the users as they
can split MO with more precise quantity (e.g.: to split an "in progress"
MO to mark a part as done and leaving the remaining quantity
to a second MO.). A MO can be split using any quantity in any state as
long as it's not marked as done and the split quantity is less than
the currenty quantity to produce and greater than 0.

The manufacturing order will not be set as done and remains open
after splitting.

Task: 5153835

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245468, opened 2026-01-23
[IMP] mrp: split MO v2

abyo-odoo:
The new split MO was introduced in https://github.com/odoo/odoo/pull/233472
and works to allow users to split an ongoing Mo.
In this commit we add a couple of points,
to furthermore give users more flexibility
when splitting an MO while updating quantity:

- When MO is `to_close` and quantity is bigger than the
new one, quantity will be adjusted to match the demand.
- Adapt error message for standard split method in order
to clarify how to split an MO while in progress.

Furthermore, it was decided to merge it with the
standard split action to avoid having 2 actions for
splitting.

So whenever we try to split an MO in progress, it will
go through the new split logic for in progress MOs and
split based on the `qty_producing`.

Users won't be blocked when splitting an MO with done
workorders, if the workorder's done quantity is more than
the new split quantity, the workorder will just be "cancelled"
as the work is already done to keep it flexible for the user.

---

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -1958,7 +1958,7 @@ class MrpProduction(models.Model):
             'orderpoint_id': self.orderpoint_id.id,
         }
 
-    def _split_productions(self, amounts=False, cancel_remaining_qty=False, set_consumed_qty=False, skip_workorder_quantity=False):
+    def _split_productions(self, amounts=False, cancel_remaining_qty=False, set_consumed_qty=False):
         """ Splits productions into productions smaller quantities to produce, i.e. creates
         its backorders.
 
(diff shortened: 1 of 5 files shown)


addons/mrp/models/mrp_production.py
```diff
@@ -2520,6 +2519,15 @@ def action_split(self):
             action['res_id'] = wizard.id
             return action
         else:
+            if self.state not in ('draft', 'confirmed'):
+                if self.qty_producing <= 0:
+                    raise UserError(_("Please specify a quantity bigger than 0."))
```
ticodoo:
```suggestion
                    raise UserError(_("Please specify a quantity larger than 0."))
```

ticodoo (commented):
Seems good, just a small English fix

PR #245491, opened 2026-01-23
[FIX] website: fix Full-Screen label on snippet previews

bvr-odoo:
Steps to reproduce the issue:

Issue 1
- Enter the "Website" edit mode.
- Drag and drop a "Cover" snippet onto the page.
- Enable the "full-height" option for this snippet.
- Save the snippet as a custom snippet.
- Click on the "Custom" category.
- Bug: the "Full-Screen" label is not displayed on the snippet preview.

Commit [1] added a label on snippet previews to indicate if they are Carousel, Popup, Gallery, Tab, or Parallax. This label is linked directly to the snippet's original template. However, this does not work well for "Full-Screen" because it's an option that can be enabled or disabled. So it doesn't make sense to keep the label when the option is not active.

task-5156137

### DIFF
--- a/addons/website/static/src/builder/snippet_model.js
+++ b/addons/website/static/src/builder/snippet_model.js
@@ -19,19 +19,19 @@ export class WebsiteSnippetModel extends SnippetModel {
      * @override
      */
     getSnippetLabel(snippetEl) {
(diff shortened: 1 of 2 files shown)


FrancoisGe:
@robodoo r+

PR #245495, opened 2026-01-23
[IMP] portal, *: merge portal chatter controllers

maki-odoo:
*: portal_rating, project

Since #214792, the `PortalChatter` has been changed so that it inherits `ThreadController` instead of `http.Controller`. The original change was made to
 the stable, so the controllers were kept intact.

Both `PortalChatter` and `PortalThreadController` inherit from `ThreadController`.
 With this change, the two were merged as `PortalThreadController`.
`portal_message_update_is_internal` is dead code so it's been removed rather than moved.


seb-odoo (approved):
@robodoo r+ delegate+

PR #245506, opened 2026-01-23
[I18N] *: remove old Albanian translations

ticodoo:
Both the po files and the translations were old and out of date. Delete them now so we can start from scratch when the language is added into Weblate.

---


oomsveta (approved):
robodoo override=ci/style review+ delegate+

PR #245507, opened 2026-01-23
[IMP] fleet: improve vehicle and model form layouts

tshe-odoo:
This commit improves the usability and consistency of Fleet forms.

Vehicle form:
- Use license_plate as the main header (H1)
- make model_id field from header to a regular field
- Update license_plate placeholder to "Unregistered"

Model form:
- Display brand_id as a regular field
- Add a new "Other" option to vehicle_type selection field

task-5013634

### DIFF
--- a/addons/fleet/models/fleet_vehicle.py
+++ b/addons/fleet/models/fleet_vehicle.py
@@ -240,7 +240,7 @@ class FleetVehicle(models.Model):
     @api.depends('model_id.brand_id.name', 'model_id.name', 'license_plate')
     def _compute_vehicle_name(self):
         for record in self:
-            record.name = (record.model_id.brand_id.name or '') + '/' + (record.model_id.name or '') + '/' + (record.license_plate or _('No Plate'))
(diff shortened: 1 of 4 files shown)


addons/fleet/models/fleet_vehicle.py
```diff
@@ -240,7 +240,7 @@ def _compute_color(self):
     @api.depends('model_id.brand_id.name', 'model_id.name', 'license_plate')
     def _compute_vehicle_name(self):
         for record in self:
-            record.name = (record.model_id.brand_id.name or '') + '/' + (record.model_id.name or '') + '/' + (record.license_plate or _('No Plate'))
+            record.name = (record.license_plate or _('No Plate')) + ': ' + (record.model_id.brand_id.name or '') + '/' + (record.model_id.name or '')
```
amah-odoo: we are now moving to use `self.env._`
```suggestion
            record.name = (record.license_plate or self.env._('No Plate')) + ': ' + (record.model_id.brand_id.name or '') + '/' + (record.model_id.name or '')
```

amah-odoo (commented):
I think we should also update the `fleet_vehicle_view_kanban` to have the name instead of having the current logic to build the name there

PR #245518, opened 2026-01-23
[FIX] purchase: suggest the right account for downpayments

viso-odoo:
Steps to reproduce:
1/ install purchase and accountant
2/ create and setup an expense account dedicated to your downpayments (typically code 60-, account type "expense"), ACC1.
3/ setup a service type product named "downpayment"
4/ set the default expense account on that product to be ACC1.
5/ create a PO for any product other than the downpayment (PO1). take note of the partner.
6/ create a bill for the same partner as the one set on PO1. Call it BILL1.
7/ On BILL1, add one invoice line with the "downpayment" product. Set a unit price.
8/ Confirm BILL1 and match it with PO1 via the "bill matching" smart button. Add it as a downpayment.
9/ Back on PO1, receive the products. Create a bill (BILL2).

=> The account suggested for the downpayment line in BILL2 will use the default expense account instead of ACC1.

After this commit, the account suggested will be the one used in BILL1 for the downpayment line.

opw-5253877

### DIFF
--- a/addons/purchase/models/purchase_order_line.py
+++ b/addons/purchase/models/purchase_order_line.py
@@ -581,6 +581,8 @@ class PurchaseOrderLine(models.Model):
             'purchase_line_id': self.id,
             'is_downpayment': self.is_downpayment,
         }
+        if self.is_downpayment and self.invoice_lines:
+            res['account_id'] = self.invoice_lines.account_id[:1].id
         return res
 
     @api.model
--- a/addons/purchase/tests/test_purchase_downpayment.py
+++ b/addons/purchase/tests/test_purchase_downpayment.py
@@ -8,11 +8,15 @@ from odoo.tests import tagged
 class TestPurchaseDownpayment(TestPurchaseToInvoiceCommon):
 
     def test_downpayment_basic(self):
+        custom_expense_account = self.company_data['default_account_expense'].copy()
+        self.product_order.property_account_expense_id = custom_expense_account
+
         po = self.init_purchase(confirm=False, products=[self.product_order])
         po.order_line.product_qty = 10.0
         po.button_confirm()
 
         dp_bill = self.init_invoice('in_invoice', amounts=[69.00], post=True)
+        dp_bill.invoice_line_ids.account_id = custom_expense_account
 
         match_lines = self.env['purchase.bill.line.match'].search([('partner_id', '=', self.partner_a.id)])
         action = match_lines.action_add_to_po()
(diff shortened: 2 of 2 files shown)


addons/purchase/models/purchase_order_line.py
```diff
@@ -581,6 +581,8 @@ def _prepare_account_move_line(self, move=False):
             'purchase_line_id': self.id,
             'is_downpayment': self.is_downpayment,
         }
+        if self.is_downpayment and self.invoice_lines:
+            res['account_id'] = self.invoice_lines.account_id[:1].id
```
Whenrow, line 585: Could you call `get_product_accounts` to extract the desired account ?

viso-odoo: Hey thanks for the review :)
So the pre-fix situation is actually that we won't specify an `account_id` to allow the system to fall back to `get_product_accounts`.
Since those are set at the "global" config level, they won't account for local "per invoice" changes.
This leads to many human corrections required at the accounting level, which is error-prone overall.
In this situation we would want the account set at the invoice line level, not at the product level.
Also worth noting that `property_account_downpayment_categ_id` is not defined for the module purchase, so we would need to add that in order to implement an override of `get_product_accounts`. Such a change, given the low-level severity of the issue seems too big for a stable version (also i don't know if adding fields is acceptable in stable, actually 😅 )

Whenrow: I mean checking the account on another line is not robust enough. If you want to set specifically the expense account of the product, use `get_product_accounts()['expense']`

viso-odoo: Hey again
I tested using `get_product_accounts()`, but with the flow that adds a downpayment section and line to the PO, the downpayment PO lines don't have a product. This ends up breaking the invoice creation because we pass a falsy value for `account_id` on the related account move line.

> checking the account on another line

Here we're not taking any random line; we are checking this PO line's matched account move line, which is is assumed to be indication of a previous vendor bill matching.

> If you want to set specifically the expense account of the product

I realize that my description and test are a bit misleading on that front; in the test case, we set a default account for the "downpayment" product so that it becomes the account used in accounting. However when we do the bill matching and add the related PO lines to the purchase order, we do not set a dedicated product on the PO lines. So currently, the expense account for the downpayment product is irrelevant for that part.

Also in my previous comment I mentioned that accounting would fall back to `get_product_accounts()` which is incorrect. It falls back to `_get_most_frequent_account_for_partner()`, which means the error is not always reproducible in runbot :sweat_smile:

PR #245520, opened 2026-01-23
[IMP] Project: Schedule recurrent tasks only on workable days

VincentJanssen-Code:
## Previous behavior:
Recurrent tasks could still be scheduled on weekends, vacation days, or other non-working days.

## New expected behavior:
Recurrent tasks are always scheduled based on the workable days and working hours of the users assigned to them. Workable days and hours are determined using the assigned users’ shared calendars. If the assigned users do not share the same calendar, the company calendar is used to resolve scheduling conflicts. If a user is on leave, the recurrent tasks they are assigned to are still scheduled, but the user is temporarily removed from the list of assigned users for the duration of their leave.This allows other individuals to take over the task when the user is on leave. Users are also removed from recurrent tasks when their contract start or end dates fall outside the task schedule since users shouldn't work with outdated contracts.

## Exception:
This behavior can be overridden if the original recurrent task was initially scheduled on a weekend day (Saturday or Sunday). In this case, recurrent tasks behave exactly as before and may be scheduled on non-working days. This was made to ensure the end-users could have the final say on this behavior in the case it is unwanted.

## Reference:

[task-4796700](https://www.odoo.com/odoo/project/4105/tasks/4796700)

---

### DIFF
--- a/addons/project/models/project_task_recurrence.py
+++ b/addons/project/models/project_task_recurrence.py
@@ -108,6 +108,12 @@ class ProjectTaskRecurrence(models.Model):
                 'stage_id': task.sudo().project_id.type_ids[0].id if task.sudo().project_id.type_ids else task.stage_id.id,
                 'child_ids': [Command.create(vals) for vals in self._create_next_occurrences_values({child: recurrence for child in task.child_ids})]
             }
+
+            if fields_to_postpone.get('date_deadline'):
+                start_dt, end_dt, planned_data, fields_to_postpone = self._plan_task(fields_to_postpone, self._get_original_task(recurrence), recurrence)
+                copy_data = self._filter_non_working_employees(copy_data, self._get_original_task(recurrence), task, start_dt, end_dt)
+                create_values.update(planned_data)
+
             create_values.update({
                 field: value[0] if isinstance(value, tuple) else value
                 for field, value in fields_to_copy.items()
@@ -120,3 +126,51 @@ class ProjectTaskRecurrence(models.Model):
             list_create_values.append(copy_data)
 
         return list_create_values
+
+    def _filter_non_working_employees(self, data, original_task, last_task, next_task_date_begin, next_task_date_deadline):
+        '''
+        Filters out all users that should not be assign to this perticular reccurent task
+        and adds back in the user that should be assigned, if they weren't in the previous task.
+        This function is inherited by other modules.
+
+        :param data: the data used to creat this recurent task
+
+        :param original_task: the task that originaly started the recurrence
+
+        :param last_task: the previous task that trigger the creation of this task
+
+        :param next_task_date_begin: the begin datetime for this task
+
+        :param next_task_date_deadline: the end/deadline for this task
+
+        :return: the updated data with the correct users.
+        '''
+        return data
+
+    def _plan_task(self, field_data, original_task, recurrence):
+        '''
+        Plans the recuring task. This function is inherited by other modules.
+
+        :param field_data: the time dependent fields for the new task
+
+        :param original_task: the task that originaly started the recurrence
+
+        :return: returns a turple with the four following values:
+            - The datetime at which the task has been planned to start
+            - The datetime at which the task has been planned to end
+            - A dictionary containg all the planned time dependant fields in 'field_data'
+            - A dictionary containg all the unused time dependent fields in 'field_data'
+        '''
+        new_data = {}
+        has_start = hasattr(self.env['project.task'], 'planned_date_begin') and field_data['planned_date_begin']
+        new_data.update({
+                field: value and value + recurrence._get_recurrence_delta()
+                for field, value in field_data.items()
+            })
+        field_data = {}
+        date_deadline = new_data['date_deadline']
+        date_begin = new_data['planned_date_begin'] if has_start else False
+        return (date_begin, date_deadline, new_data, field_data)
+
+    def _get_original_task(self, recurrence):
+        return recurrence.task_ids[0]


addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+        :return: returns a turple with the four following values:
+            - The datetime at which the task has been planned to start
+            - The datetime at which the task has been planned to end
+            - A dictionary containg all the planned time dependant fields in 'field_data'
+            - A dictionary containg all the unused time dependent fields in 'field_data'
+        '''
+        new_data = {}
+        date_deadline = False
+        recurrence_cal = False
+        if (original_task.date_deadline.weekday() in [5, 6]):
```
A-Yehia19: I guess it is not a good idea to set weekends to always be Saturday and Sunday `[5, 6]`.
Some countries and companies have different weekends (eg., Friday, Saturday), or some can even have it as three days or only one day

this maybe handy `_works_on_date(date)` (to check 👀 )

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+            new_data.update({
+            field: value and value + recurrence._get_recurrence_delta()
+            for field, value in field_data.items()
+            })
```
A-Yehia19: another knit picking 😇 
```suggestion
            new_data.update({
                field: value and value + recurrence._get_recurrence_delta()
                for field, value in field_data.items()
            })
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+                users_to_add.add(user_id)
+
+        # filter by timeoff
+        for user_id in last_task.user_ids:
+            if (any(user_id.resource_calendar_id._leave_intervals_batch(next_task_date_deadline.replace(tzinfo=timezone.utc), next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1), user_id.resource_ids).values())):
+                users_to_remove.add(user_id.id)
+                continue
+
+            # filter by invalid contract
+        users_to_check = set(last_task.user_ids) | users_to_add
```
A-Yehia19: Some tiny tweak 🕵🏻 
```suggestion
        # filter by invalid contract
        users_to_check = set(last_task.user_ids) | users_to_add
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+                for employee in user_id.employee_ids:
+                    if (employee.contract_date_end):
+                        contract_found = True
+                        if (not employee.contract_date_end or datetime.combine(employee.contract_date_end, time(23, 59, 59)) >= next_task_date_deadline):
+                            contract_within_dates = True
+                            break
+                if (contract_found and not contract_within_dates):
+                    users_to_remove.add(user_id.id)
+
+        users_to_add = [user.id for user in users_to_add]
```
A-Yehia19: If `users_to_add` is already a set(), why change it to a list to return it to a set again 🤔 
```suggestion
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+        if (data['user_ids'] and data['user_ids'][0] and data['user_ids'][0][0] == Command.SET):
+            data['user_ids'][0] = Command.set(list((set(data['user_ids'][0][2]) - set(users_to_remove)) | set(users_to_add)))
+            return data
+        data['user_ids'] = [Command.set(list((set(data['user_ids'][0][2]) - set(users_to_remove)) | set(users_to_add)))]
+        return data
```
A-Yehia19: IMO, the if conditions with a specific positioning are a little fragile; I think it's better to make it more adaptable to whatever the command list may take. Also, it makes it easier to read
```suggestion
        data.setdefault('user_ids', [])
        data['user_ids'] += [Command.link(user_id) for user_id in users_to_add.ids]
        data['user_ids'] += [Command.unlink(user_id) for user_id in users_to_remove]
        return data
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+
+        :param next_task_date_begin: the begin datetime for this task
+
+        :param next_task_date_deadline: the end/deadline for this task
+
+        :return: the updated data with the correct users.
+        '''
+        # add back user out of vacation
+        users_to_add = set()
+        users_to_remove = set()
```
A-Yehia19: iMO maybe it's better to use recordsets instead of sets, as it adds a type for the variable so we can know what type they carry. What do you think 👀 
(You're storing the user record in `users_to_add`, while storing the user ids in `users_to_remove` which makes it trickier to maintain)
```suggestion
        users_to_add = self.env['res.users']
        users_to_remove = self.env['res.users']
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+        for user_id in original_task.user_ids - last_task.user_ids:
+            if (not any(user_id.resource_calendar_id._leave_intervals_batch(
+                    next_task_date_deadline.replace(tzinfo=timezone.utc),
+                    next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1),
+                    user_id.resource_ids).values())):
```
A-Yehia19: I got an error when trying to get the leaves, maybe it is missing this method `_get_resources_per_tz()`
```suggestion
            if (not any(user_id.resource_calendar_id._leave_intervals_batch(
                    next_task_date_deadline.replace(tzinfo=timezone.utc),
                    next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1),
                    user_id.resource_ids._get_resources_per_tz()).values())):
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+
+                users_to_add.add(user_id)
+
+        # filter by timeoff
+        for user_id in last_task.user_ids:
+            if (any(user_id.resource_calendar_id._leave_intervals_batch(next_task_date_deadline.replace(tzinfo=timezone.utc), next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1), user_id.resource_ids).values())):
```
A-Yehia19: same error is fired here also 🤔 
```suggestion
        # filter by timeoff
        for user_id in last_task.user_ids:
            if (any(user_id.resource_calendar_id._leave_intervals_batch(next_task_date_deadline.replace(tzinfo=timezone.utc), next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1), user_id.resource_ids._get_resources_per_tz()).values())):
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+                contract_found = False
+                contract_within_dates = False
+                for employee in user_id.employee_ids:
+                    if (employee.contract_date_end):
+                        contract_found = True
+                        if (not employee.contract_date_end or datetime.combine(employee.contract_date_end, time(23, 59, 59)) >= next_task_date_deadline):
+                            contract_within_dates = True
+                            break
+                if (contract_found and not contract_within_dates):
+                    users_to_remove.add(user_id.id)
```
A-Yehia19: in a first look i think its better to make a search query instead of double nested loops
(i will check that again)

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+            date_deadline = new_data['date_deadline']
+        else:
+            old_dead_line = field_data.pop('date_deadline')
+            if (last_task.user_ids.resource_calendar_id and len(last_task.user_ids.resource_calendar_id) == 1):
+                recurrence_cal = last_task.user_ids.resource_calendar_id
+            else:
+                recurrence_cal = self.env.user.company_id.resource_calendar_id
+
+            date_deadline = recurrence_cal.plan_hours(0, old_dead_line + recurrence._get_recurrence_delta(), True)
+            date_deadline.replace(tzinfo=timezone.utc)
```
A-Yehia19: This will make the deadline time the same as the working schedule time (eg., it's set to 8 AM, even if I change it to 11 PM). I don't think it's part of the spec to do so, so maybe it's better to keep the deadline time the same as the original task time

A-Yehia19: i just confimed with AJU that it should be he same as the original task

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+
+        :return: the updated data with the correct users.
+        '''
+        # add back user out of vacation
+        users_to_add = set()
+        users_to_remove = set()
+        for user_id in original_task.user_ids - last_task.user_ids:
+            if (not any(user_id.resource_calendar_id._leave_intervals_batch(
+                    next_task_date_deadline.replace(tzinfo=timezone.utc),
+                    next_task_date_deadline.replace(tzinfo=timezone.utc) + relativedelta(microseconds=1),
```
A-Yehia19: It's not a good idea to make the range only 1 microsecond because if the deadline is after working hours (eg, 10 PM) and the user has already taken that day off, it's not gonna detect this leave and will assign the user to it

addons/project/models/project_task_recurrence.py
A-Yehia19: Same comment here for the microsecond range

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+            if (last_task.user_ids.resource_calendar_id and len(last_task.user_ids.resource_calendar_id) == 1):
+                recurrence_cal = last_task.user_ids.resource_calendar_id
+            else:
+                recurrence_cal = self.env.user.company_id.resource_calendar_id
+
+            date_deadline = recurrence_cal.plan_hours(0, old_dead_line + recurrence._get_recurrence_delta(), True)
+            date_deadline.replace(tzinfo=timezone.utc)
+            new_data.update({
+                'date_deadline': date_deadline
+                })
```
A-Yehia19: Excuse my OCD 🙈 🧹 
```suggestion
            new_data.update({
                'date_deadline': date_deadline
            })
```

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+
+        return (None, date_deadline, new_data, field_data)
```
A-Yehia19: As mentioned in the enterprise PR, you can utilise the first return with the right deadline time shift to avoid doing an override and repeating the same code logic again

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +129,103 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+            for field, value in field_data.items()
+            })
+            field_data = {}
+            date_deadline = new_data['date_deadline']
+        else:
+            old_dead_line = field_data.pop('date_deadline')
+            if (last_task.user_ids.resource_calendar_id and len(last_task.user_ids.resource_calendar_id) == 1):
+                recurrence_cal = last_task.user_ids.resource_calendar_id
+            else:
+                recurrence_cal = self.env.user.company_id.resource_calendar_id
```
A-Yehia19: You didn't handle the case of when multiple assignees share the same working calendar here. It should pick up this calendar instead of the company calendar

addons/project/tests/test_project_recurrence.py
```diff
@@ -315,12 +316,13 @@ def test_next_occurrence_batch_call(self):
...
-        # Every date should be 1 week later
+        # Every date should be 1 week later if planned on weekend
         self.assertEqual(datetime(2023, 1, 8, 0, 0), tasks_copy[0].date_deadline)
         self.assertEqual(datetime(2023, 1, 9, 0, 0), tasks_copy[0].child_ids.date_deadline)
         self.assertEqual(datetime(2023, 1, 10, 0, 0), tasks_copy[0].child_ids.child_ids.date_deadline)
-        self.assertEqual(datetime(2023, 1, 11, 0, 0), tasks_copy[1].date_deadline)
-        self.assertEqual(datetime(2023, 1, 12, 0, 0), tasks_copy[1].child_ids.date_deadline)
+        # Every date should be 1 week later on a working day and hour if planned durring the week
+        self.assertEqual(datetime(2023, 1, 11, 8, 0), tasks_copy[1].date_deadline)
+        self.assertEqual(datetime(2023, 1, 12, 8, 0), tasks_copy[1].child_ids.date_deadline)
```
A-Yehia19: As mentioned, the created task should match the original task deadline time

addons/project/models/project_task_recurrence.py
```diff
@@ -108,10 +109,18 @@ def _create_next_occurrences_values(self, recurrence_by_task):
                 'stage_id': task.sudo().project_id.type_ids[0].id if task.sudo().project_id.type_ids else task.stage_id.id,
                 'child_ids': [Command.create(vals) for vals in self._create_next_occurrences_values({child: recurrence for child in task.child_ids})]
             }
+
```
A-Yehia19, line 111: Useless diff 👀 
```suggestion
```

addons/project/models/project_task_recurrence.py
VincentJanssen-Code: Thank you for the comment. I’m not sure this behavior is actually unwanted. The specification states: “When the user assigned is on time off, schedule the task but leave it unassigned.”

In the example you mention, the user’s time off ends at the end of the last working hour. After that point, the user is no longer on time off — they are simply outside their working hours. Since the deadline occurs after the time‑off period has ended, there is no reason to treat it as time off, and therefore no reason not to assign the user.

I agree that this logic is not the most intuitive, but I followed it because it matches the behavior in Planning. When a shift is outside working hours and recurs daily, no time‑off conflict warning appears on the last day of the time off for the same reason as above.

I also checked with DIAG, who wrote the specification (AJU is currently on vacation), and she confirmed that this behavior is indeed expected.

Please send me a message or add a comment if I missed something 😉

A-Yehia19: Hmm hmmm, if that's what the specs are, you are correct indeed👌🏻✨.
I just confirmed it with DIAG.
(You can ignore this comment 😌)

addons/project/models/project_task_recurrence.py
```diff
@@ -120,3 +126,51 @@ def _create_next_occurrences_values(self, recurrence_by_task):
...
+            - The datetime at which the task has been planned to end
+            - A dictionary containg all the planned time dependant fields in 'field_data'
+            - A dictionary containg all the unused time dependent fields in 'field_data'
+        '''
+        new_data = {}
+        has_start = hasattr(self.env['project.task'], 'planned_date_begin') and field_data['planned_date_begin']
+        new_data.update({
+                field: value and value + recurrence._get_recurrence_delta()
+                for field, value in field_data.items()
+            })
```
A-Yehia19, line 169: 🕵🏻 🧹 
```suggestion
        new_data.update({
            field: value and value + recurrence._get_recurrence_delta()
            for field, value in field_data.items()
        })
```

addons/project/tests/test_project_recurrence.py
```diff
@@ -92,10 +92,11 @@ def test_recurrent_tasks_fields(self):
...
 
-        self.assertEqual(
-            other_task.date_deadline, task.date_deadline + relativedelta(months=2),
-            "Next occurrence should have previous deadline + interval * unit",
-        )
+        # Due to recurring tasks appling to work day this assertion no longer applied
+        # self.assertEqual(
+        #     other_task.date_deadline, task.date_deadline + relativedelta(months=2),
+        #     "Next occurrence should have previous deadline + interval * unit",
+        # )
```
A-Yehia19: No need to remove this assertion since we are keeping the original behaviour in the `project` module untouched

PR #245521, opened 2026-01-23
[FIX] l10n_tr_nilvera_einvoice: get_param and set_param don't exist

gawa-odoo:
The methods don't exist anymore. They should be replaced by the specific types.

---


alialfie (approved):
Thanks for the PR! I'm stopping the FW ports as https://github.com/odoo/odoo/pull/244818 replaces the params with fields in master anyway
robodoo fw=no r+

PR #245534, opened 2026-01-23
[FIX] sale_loyalty: unlink unused coupon ids

mma1377:
### Issue:
Due to this issue, `coupon_ids` which are never used are still shown in loyalty coupons history which is confusing for users.

#### Steps to reproduce:
1- Create two promotions:
   - Promo A: 10% discount
   - Promo B: 20% discount

2- Create a SO and add a line. Then confirm the SO.
3- Click on `Reward` button.
4- A dialog should show both promos. Discard the dialog.
5- Navigate to `Discount and Loyalty`.
6- Open Promo A and B and you can see both have lines with
   `coupon_id.order_id` = `created SO` which is confusing. `Issue 1`
7- Navigate back to SO. Click on `Reward` and select Promo A, apply.
8- Navigate to Promo B. As you see the the `coupon_id` related to
   the SO is still there. `Issue 2`

### Cause and Fix:
This is due to unlinking `coupon_ids` only in SO confirm action, causing this issue on already confirmed SOs:
https://github.com/odoo/odoo/blob/604d07ab324caa5f3aa6f3baa9902c2137ea24db/addons/sale_loyalty/models/sale_order.py#L132-L136

We can fix this issue by `unlinking` the `coupons_ids` when a reward is applied or the dialog is discarded.

However there are 2 limitations:
1- The view needs to be updated. Otherwise, `coupon_ids` will not be deleted upon clicking on `discard` button. However the fix is still stable.
2- What if the dialog is closed with the cross button or the tab is closed without selecting a reward? In that case the issue will still persist.

To fix these issues, we can show only used `coupon_ids` in `loyalty_coupon_id` list view. However that cannot be done in stable as it requires `use_count` to be stored.
NB: With some complex filters this also could be fixed on stable but not worth it.

opw-5424753

### DIFF
--- a/addons/sale_loyalty/wizard/sale_loyalty_reward_wizard.py
+++ b/addons/sale_loyalty/wizard/sale_loyalty_reward_wizard.py
@@ -52,4 +52,18 @@ class SaleLoyaltyRewardWizard(models.TransientModel):
             raise ValidationError(_('Coupon not found while trying to add the following reward: %s', self.selected_reward_id.description))
         self.order_id._apply_program_reward(self.selected_reward_id, coupon, product=self.selected_product_id)
         self.order_id._update_programs_and_rewards()
+        self._unlink_unused_coupon_ids()
         return True
+
+    def action_cancel(self):
+        self.ensure_one()
+        self._unlink_unused_coupon_ids()
+
+    def _unlink_unused_coupon_ids(self):
+        reward_coupons = self.order_id.order_line.coupon_id
+        self.order_id.coupon_point_ids.filtered(
+            lambda points: (
+                points.coupon_id.program_id.applies_on == 'current' and
+                points.coupon_id not in reward_coupons
+            )
+        ).coupon_id.sudo().unlink()
--- a/addons/sale_loyalty/wizard/sale_loyalty_reward_wizard_views.xml
+++ b/addons/sale_loyalty/wizard/sale_loyalty_reward_wizard_views.xml
@@ -28,7 +28,7 @@
                     <!-- Has rewards -->
                     <button type="object" name="action_apply" string="Apply" class="btn btn-primary"
                         invisible="not reward_ids" data-hotkey="q"/>
-                    <button special="cancel" string="Discard" class="btn btn-secondary"
+                    <button type="object" name="action_cancel" string="Discard" class="btn btn-secondary"
                         invisible="not reward_ids" data-hotkey="x"/>
                     <!-- No rewards -->
                     <button special="cancel" string="Discard" class="btn btn-primary"


addons/sale_loyalty/wizard/sale_loyalty_reward_wizard.py
```diff
@@ -52,4 +52,15 @@ def action_apply(self):
...
         return True
+
+    def action_cancel(self):
+        self.ensure_one()
+        self.unlink_unused_coupon_ids()
+
+    def unlink_unused_coupon_ids(self):
+        reward_coupons = self.order_id.order_line.coupon_id
+        self.order_id.coupon_point_ids.filtered(
+            lambda pe: pe.coupon_id.program_id.applies_on == 'current' and pe.coupon_id not in reward_coupons
```
ltinel: Nit: why `pe`? Personally, I'd find `points` clearer :)
```suggestion
            lambda points: points.coupon_id.program_id.applies_on == 'current' and points.coupon_id not in reward_coupons
```

addons/sale_loyalty/wizard/sale_loyalty_reward_wizard.py
```diff
@@ -52,4 +52,15 @@ def action_apply(self):
...
         self.order_id._apply_program_reward(self.selected_reward_id, coupon, product=self.selected_product_id)
         self.order_id._update_programs_and_rewards()
+        self.unlink_unused_coupon_ids()
         return True
+
+    def action_cancel(self):
+        self.ensure_one()
+        self.unlink_unused_coupon_ids()
+
+    def unlink_unused_coupon_ids(self):
```
ltinel: I think this method can be private.

--count
ltinel, line 1: Added by mistake?

ltinel (changes requested):
Thanks for the fix :) From what I know, it LGTM (apart from a few nits)

@anko-odoo you probably have a bit more context here. Would you mind giving it a quick look as well?

--count
mma1377: @ltinel Oops! sorry 😅

addons/sale_loyalty/wizard/sale_loyalty_reward_wizard_views.xml
```diff
@@ -28,7 +28,7 @@
                     <!-- Has rewards -->
                     <button type="object" name="action_apply" string="Apply" class="btn btn-primary"
                         invisible="not reward_ids" data-hotkey="q"/>
-                    <button special="cancel" string="Discard" class="btn btn-secondary"
+                    <button type="object" name="action_cancel" string="Discard" class="btn btn-secondary"
```
ltinel, line 31: @Feyensv in theory, this could fail if the template is updated without updating the python model, but I don't think that can happen?

Feyensv: I think it can only happen if the db is in dev xml mode :thinking:

ltinel: And is that something we should handle? :sweat_smile: I guess not?

Feyensv: I don't think so :D

anko-odoo (approved):
LGTM

ltinel (approved):
@robodoo r+

PR #245535, opened 2026-01-23
[FIX] l10n_my_edi_pos: use tax-excluded amounts for discount calculation

Alpha9584:
The _add_consolidated_invoice_base_lines_vals method computed the gross subtotal using `price_unit * quantity`. When taxes are configured as "Included in Price", `price_unit` contains the tax-included amount, but `total_excluded` (used for the discounted amount) is tax-excluded.

This caused the tax amount to be incorrectly reported as an AllowanceCharge (discount) in the MyInvois XML, because:
  discount_amount = price_unit * qty - total_excluded
                  = tax_included - tax_excluded
                  = TAX AMOUNT (not a discount!)

Example: Product priced at 110 MYR with 10% tax included:
- price_unit = 110 (tax-included)
- total_excluded = 100 (tax-excluded: 110 / 1.10)
- discount_amount = 110 - 100 = 10 ← incorrectly reported as discount

refs:

The cac:AllowanceCharge element in UBL is specifically for discounts and surcharges, NOT for taxes. According to the Peppol Malaysia e-Invoice specification:

https://docs.peppol.eu/poac/my/pint-my-sb/bis/#_allowances_and_charges

https://docs.peppol.eu/poacc/billing/3.0/codelist/UNCL5189/ (For this case we are interested in code 95)

Steps to Reproduce:
1. Configure a tax as "Included in Price" with Malaysia tax type
2. Create a product with that tax.
3. Create POS orders without any discount
4. Generate consolidated invoice and XML
5. XML incorrectly shows <cac:AllowanceCharge> with tax amount as discount

The fix uses `raw_total_excluded / discount_factor` (always tax-excluded) instead of `price_unit * quantity` (may be tax-included), consistent with the parent method:
https://github.com/odoo/odoo/blob/d645361a95037ac580d55e80bcb61d1eeb293efd/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py#L1846-L1876

Ticket [link](https://www.odoo.com/odoo/project.task/5476526) 
opw-5476526

---

### DIFF
--- a/addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
+++ b/addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
@@ -375,6 +375,58 @@ class TestMyInvoisPoS(TestPoSCommon):
             # And the discount should be 100
             self._assert_node_values(xml_tree, "cac:InvoiceLine/cac:AllowanceCharge/cbc:Amount", '100.00')
 
+    @mute_logger('odoo.addons.point_of_sale.models.pos_order')
+    def test_consolidate_invoices_with_tax_included_in_price(self):
+        """ Test that price_include taxes don't incorrectly appear as discounts in XML. """
+        tax_included = self.env['account.tax'].create({
+            'name': "10% Included",
+            'amount_type': 'percent',
+            'amount': 10,
+            'price_include_override': 'tax_included',
+            'l10n_my_tax_type': '01',
+        })
+        product = self.create_product("Product", self.categ_basic, 110, tax_ids=tax_included.ids)
+
+        with freeze_time("2025-01-01"):
+            with self.with_pos_session():
+                order = self._create_order({'pos_order_lines_ui_args': [(product, 1.0)]})
+            wizard = self.env['myinvois.consolidate.invoice.wizard'].create({
+                'date_from': '2025-01-01',
+                'date_to': '2025-01-31',
+            })
+            wizard.button_consolidate_orders()
+            order.consolidated_invoice_ids.action_generate_xml_file()
+            root = etree.fromstring(order.consolidated_invoice_ids.myinvois_file_id.raw)
+            with file_open('l10n_my_edi_pos/tests/expected_xmls/consolidated_invoice_tax_included.xml', 'rb') as f:
+                expected_xml = etree.fromstring(f.read())
+            self.assertXmlTreeEqual(root, expected_xml)
+
+    @mute_logger('odoo.addons.point_of_sale.models.pos_order')
+    def test_consolidate_invoices_with_tax_included_in_price_and_discount(self):
+        """ Test that price_include taxes with actual discounts work correctly. """
+        tax_included = self.env['account.tax'].create({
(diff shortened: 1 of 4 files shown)


agr-odoo (approved):
Looks great, thanks

antoine162 (commented):
The fix seems good to me.

I'd suggest asserting a few more values in the test.

@vin-odoo FYI

addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
```diff
@@ -375,6 +375,30 @@ def test_consolidate_invoices_with_different_discounts(self):
...
+            with self.with_pos_session():
+                order = self._create_order({'pos_order_lines_ui_args': [(product, 1.0)]})
+            wizard = self.env['myinvois.consolidate.invoice.wizard'].create({
+                'date_from': '2025-01-01',
+                'date_to': '2025-01-31',
+            })
+            wizard.button_consolidate_orders()
+            order.consolidated_invoice_ids.action_generate_xml_file()
+            xml_tree = etree.fromstring(order.consolidated_invoice_ids.myinvois_file_id.raw)
+            self.assertFalse(xml_tree.xpath("cac:InvoiceLine/cac:AllowanceCharge", namespaces=NS_MAP))
```
antoine162: In this case, I'd suggest doing a few more asserts to make sure the generated XML is exactly as we expect:
- `InvoiceLine/Price/PriceAmount`
- `InvoiceLine/ItemPriceExtension/PriceAmount`
- `InvoiceLine/TaxTotal/TaxAmount`
- `InvoiceLine/TaxTotal/TaxSubtotal/TaxableAmount`
- `InvoiceLine/TaxTotal/TaxSubtotal/TaxAmount`
- `InvoiceLine/LineExtensionAmount`

as well as the
- `LegalMonetaryTotal/TaxExclusiveAmount`
- `LegalMonetaryTotal/TaxInclusiveAmount`
- `TaxTotal/TaxAmount`
- `TaxTotal/TaxSubtotal/TaxableAmount`
- `TaxTotal/TaxSubtotal/TaxAmount`

vin-odoo: I would also add a second test where we do use a discount, just to make sure it'll work in both cases.

vin-odoo (commented):
The fix itself seems ok to me; but we'll need to test it functionally to ensure it works as expected on MyInvois' platform.

(KIDO will take care of the functional validation)

addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
Alpha9584: I made it compare against a full expected XML file and added the second test with discount

vin-odoo (approved):
LGTM

cc @kidoodoo for final functional validation

vin-odoo (commented):
robodoo r+

Let's go then :) Thanks for the fix!

PR #245538, opened 2026-01-23
[IMP] setup: emergency controller for remote debug

loouis-t:
We add a separate service that hosts a simple http server with a controller to start Tailscale.

We make this controller available through a form on the "IoT Box is down" page, so that clients can easily grant access to the support to get help.

Task: 5871829

### DIFF
--- a/setup/iot_box_builder/overwrite_after_init/etc/sos-controller.py
+++ b/setup/iot_box_builder/overwrite_after_init/etc/sos-controller.py
@@ -1,0 +1,37 @@
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from urllib.parse import urlparse, parse_qs
+import subprocess
+
+
+class Handler(BaseHTTPRequestHandler):
+    def make_response(self, code, message):
+        self.send_response(code)
+        self.end_headers()
+        self.wfile.write(message.encode())
+
+    def do_GET(self):
+        qs = parse_qs(urlparse(self.path).query)
+        token = qs.get("token", [None])[0]
+
+        if not token or not token.startswith("tskey-"):
+            self.make_response(400, "Invalid or missing token")
+            return
+
+        try:
+            subprocess.run(
+                ["sudo", "tailscale", "up", f"--authkey={token}", "--hostname=emergency-access", "--timeout=10s"],
+                check=True,
+            )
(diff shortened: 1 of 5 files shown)


setup/iot_box_builder/overwrite_after_init/etc/sos-controller.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Handler(BaseHTTPRequestHandler):
+    def do_GET(self):
+        qs = parse_qs(urlparse(self.path).query)
+        token = qs.get("token", [None])[0]
+
+        if not token or not token.startswith("tskey-"):
+            self.send_response(400)
+            self.end_headers()
+            self.wfile.write(b"Invalid token")
```
MaxTheWhale: nitpick: we could reduce the repetition by making this a function on the class

```python
def make_response(self, code, message):
    self.send_response(code)
    self.end_headers()
    self.wfile.write(message.encode())
```

setup/iot_box_builder/overwrite_after_init/var/www/html/502.html
```diff
@@ -86,6 +86,11 @@ <h2>(502 Bad Gateway)</h2>
                     <a href="https://www.odoo.com/documentation/latest/applications/general/iot/iot_advanced/updating_iot.html#iot-updating-iot-image-code" target="_blank">documentation</a>
                 </li>
             </ol>
+            <p>Or, enter a Remote Debug token provided by <b>Odoo support ONLY</b>:</p>
+            <form method="GET" action="/sos" style="display: flex; gap: 10px;">
+                <input type="text" name="token" placeholder="tskey-auth-..." style="width: 100%; padding: 8px; border-radius: 4px; border: 1px solid #212529" required>
+                <button type="submit" style="background-color: #714b67; color: white; border: none; border-radius: 4px; cursor: pointer;">Enable</button>
```
MaxTheWhale: We could move the styles into the `style` block above to match the rest of the file

MaxTheWhale (approved):
Tested:
- Odoo service down, tskey can be entered and works as expected ✅ 
- Odoo service is up but has DB selector, tskey can be entered and works as expected ✅ 

Looks good to me!

PR #245543, opened 2026-01-23
[IMP] survey: add minimum width on columns to improve ux

jeromebelpaire:
The columns 'Skipped' and 'Correct' would by default show as 'S...' and 'C...'. By adding a minimum column width of 70 we ensure that the column names will be shown.

Task-5242008

### DIFF
--- a/addons/survey/views/survey_user_views.xml
+++ b/addons/survey/views/survey_user_views.xml
@@ -98,9 +98,9 @@
                                     <field name="page_id" optional="hidden"/>
                                     <field name="question_id"/>
                                     <field name="answer_type" optional="hidden"/>
-                                    <field name="skipped" hide="1"/>
+                                    <field name="skipped" width="[70]" hide="1"/>
                                     <field name="display_name" string="Answer"/>
-                                    <field name="answer_is_correct"/>
+                                    <field name="answer_is_correct" width="[70]"/>
                                     <field name="answer_score" sum="Score"/>
                                 </list>
                             </field>


pyduf (approved):
Hello @jeromebelpaire , 

Thanks for the PR. It looks good to me (and I didn't know that feature so I have learned something).

addons/survey/views/survey_user_views.xml
```diff
@@ -98,9 +98,9 @@
                                     <field name="page_id" optional="hidden"/>
                                     <field name="question_id"/>
                                     <field name="answer_type" optional="hidden"/>
-                                    <field name="skipped" hide="1"/>
+                                    <field name="skipped" width="[70]" hide="1"/>
```
awa-odoo, line 101: 👋 @jeromebelpaire 

Is it really specified as a table? Feels a bit strange.

Bonus: could target 19.0 if we'd like.

Otherwise LGTM ✅

jeromebelpaire: Not sure what you mean with _"specified as a table"_ ? 

And I can target 19.0* if you think it's better, I just thought we tried not to change views in stable? (this is more of visual improvement than a real bug)

* if I retarget i'll have to use `width="70"` (exactly 70px) instead of `width="[70]"` (minimum 70px) because this feature was only added in 19.1. But it doesn't really matter because the fields are booleans so width will be constant.

awa-odoo: Was not super clear sorry, I meant specified as a list `[70]` as opposed to a single number `70`.

But somehow your answer to my second question answered the first one so we're good :smile: 

Thanks for the details!

PR #245554, opened 2026-01-23
[FIX] account_edi_ubl_cii: set imported partner as company

MerlinGuillaume:
Importing an xml with an unknown partner sets the partner as individual when it should be set as a company

Steps to reproduce:
1. Install Invoicing
2. Open Invoicing and click on "Upload"
3. Select any invoice xml (e.g. https://docs.oasis-open.org/ubl/os-UBL-2.2/xml/UBL-Invoice-2.1-Example.xml)
4. On the draft invoice created, click on the internal link next to the customer
5. The new partner created is an individual, it should be a company

Solution:
Set new imported partners as company
backport of https://github.com/odoo/odoo/commit/aa9fd54ab321eb591666244fbd14c1127151de70

opw-5490363

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_common.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_common.py
@@ -381,7 +381,8 @@ class AccountEdiCommon(models.AbstractModel):
             .with_company(invoice.company_id) \
             ._retrieve_partner(name=name, phone=phone, mail=mail, vat=vat, domain=domain)
         if not invoice.partner_id and name and vat:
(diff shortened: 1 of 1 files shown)


kitan191 (approved):
LGTM

Not sure why the original PR was targeting 18.0.
Maybe we didn't want to change the behavior on the lowest version. :thinking:

antoine162 (approved):
@robodoo r+

The original task had the 18.0 and 19.0 tags on it, presumably because of concerns about doing bugfixes in 17.0.

But since we have lots of people using this feature via Peppol (including in 17.0), I think it makes sense to fix it there as well.

We even back-ported Peppol to 16.0 so we might need it there at some point too 🙈

PR #245577, opened 2026-01-23
[IMP] account_edi_ubl_cii: Generate PDF when missing on XML vendor bills

jugurtha-gaci:
Some vendor bills received via Peppol as XML do not include an attached PDF, preventing users from previewing the document and verifying the values parsed by Odoo.

This commit generates and attaches a PDF when the XML does not provide one.

task-5500236

---


addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,49 @@
...
 class AccountEdiUBL(models.AbstractModel):
     _name = "account.edi.ubl"
     _inherit = 'account.edi.common'
     _description = "Base helpers for UBL"
 
+    def _import_attachments(self, invoice, tree):
+        if additional_docs := super()._import_attachments(invoice, tree):
+            return additional_docs
+
+        if invoice.move_type != 'in_invoice':
```
clbr-odoo: We probably want to do it for refund too ? Use `is_purchase_document`.

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,49 @@
...
+            return []
+
+        # if an embedded PDF isn't present in the XML, we attempt to create one.
+        try:
+            pdf_raw, _ = self.env['ir.actions.report'].with_context(
+                            force_report_rendering=True,
```
clbr-odoo: Doesn't make sense to use this key outside of test environment, set that context directly in your test.

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,49 @@
...
     _description = "Base helpers for UBL"
 
+    def _import_attachments(self, invoice, tree):
+        if additional_docs := super()._import_attachments(invoice, tree):
+            return additional_docs
+
+        if invoice.move_type != 'in_invoice':
+            return []
+
+        # if an embedded PDF isn't present in the XML, we attempt to create one.
```
clbr-odoo: this comment would be better placed next to the if checking if there are already attachments :)

addons/account_edi_ubl_cii/views/report_invoice.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
clbr-odoo: Let's move this into `account` as its quite a generic change ...
Imo it's not worth it, this template will be created if we update the module only, so most people will still have their own template.
I'm discussing it with GUHEC. A very ugly trick would be to set the `external_report_layout_id` on the company, and change it back just after the render 🫠

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,49 @@
 from odoo import _, models
 from odoo.addons.account_edi_ubl_cii.models.account_edi_common import FloatFmt
-from odoo.tools import frozendict
+from odoo.tools import frozendict, pdf
+import base64
+import io
```
clbr-odoo: Imports are not following guidelines

addons/account_peppol/tests/test_peppol_messages.py
```diff
@@ -356,7 +357,7 @@ def test_send_invalid_edi_user(self):
 
     def test_receive_error_peppol(self):
         # an error peppol message should be created
-        with self._set_context({'error': True}):
+        with self._set_context({'error': True}), patch.object(AccountEdiUBL, '_import_attachments', return_value=[]):
```
clbr-odoo: Which assert fails if you don't do that ? 🤨 
Probably just removing the `force_report_rendering` from the business code will solve it.

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
clbr-odoo, line 1: You never check if your PDF was actually generated don't you ? 🤔

clbr-odoo (changes requested):

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,49 @@
...
+            return []
+
+        # if an embedded PDF isn't present in the XML, we attempt to create one.
+        try:
+            pdf_raw, _ = self.env['ir.actions.report'].with_context(
+                            force_report_rendering=True,
+                            report_layout='web.external_layout_standard'  # forcing the "light" layout
+                        )._render_qweb_pdf('account.account_invoices', res_ids=[invoice.id])
+
+            name = (invoice.display_name or 'Bill').replace('Draft', '') + ' - Generated by Odoo from the XML file.pdf'
```
clbr-odoo: Missing translation.

clbr-odoo (commented):
That's close to what we should do in master.
Let's talk about it Monday with Guhec, because I don't think there's a way to make this PR useful in stable.
People will never update `web`, nor `account` modules. That's not acceptable.
We should see if we can make use of `internal_layout` which would be more coherent conceptually.

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,51 @@
...
-from odoo.tools import frozendict
 
 
 class AccountEdiUBL(models.AbstractModel):
     _name = "account.edi.ubl"
     _inherit = 'account.edi.common'
     _description = "Base helpers for UBL"
 
+    def _import_attachments(self, invoice, tree):
+        """This is an overrided method, it ATTEMPTS to create a PDF attachment when the XML file doesn't provide one."""
```
clbr-odoo:
```suggestion
        """# EXTENDS 'account_edi_common': ATTEMPTS to create a PDF attachment when the XML file doesn't provide one."""
```

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,51 @@
...
+
+        if additional_docs := super()._import_attachments(invoice, tree):
+            return additional_docs
+
+        attachments = self.env['ir.attachment']
+        if not invoice.is_purchase_document():
+            return attachments
```
clbr-odoo:
```suggestion
        additional_docs = super()._import_attachments(invoice, tree)
        if additional_docs or not invoice.is_purchase_document():
            return additional_docs
```
Small simplification.

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,51 @@
...
+        try:
+            pdf_raw, _pdf_extension = self.env['ir.actions.report'] \
+                        ._render_qweb_pdf('account_edi_ubl_cii.action_report_generated_purchase_document', res_ids=[invoice.id])
```
clbr-odoo: This will break in databases that did not upgrade their `account_edi_ubl_cii` module, that thus don't have the new report/layout created.

We need to : 
```python
invoices_by_odoo_xmlid = 'account_edi_ubl_cii.action_report_generated_purchase_document'
report_xmlid = invoices_by_odoo_xmlid if self.env.ref(invoices_by_odoo_xmlid, raise_if_not_found=False) else 'account.account_invoices'
[...]._render_qweb_pdf(report_xmlid, [...])
```

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <template id="generated_purchase_document">
```
clbr-odoo:
```suggestion
    <template id="account_invoices_generated_by_odoo">
```
Let's stay as generic as possible, pretty sure we'll want to extend this function to self-billing, which are out-invoices later on.

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.xml
```diff
@@ -0,0 +1,26 @@
...
+        <t t-call="web.html_container">
+            <t t-foreach="docs" t-as="o">
+                <t t-set="lang" t-value="o.partner_id.lang"/>
+                <t t-call="account_edi_ubl_cii.invoice_document_with_watermark"/>
+            </t>
+        </t>
+    </template>
+
+    <template id="invoice_document_with_watermark" inherit_id="account.report_invoice_document" primary="True">
```
clbr-odoo:
```suggestion
    <template id="report_invoice_document_with_watermark" inherit_id="account.report_invoice_document" primary="True">
```

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.py
clbr-odoo, line 1: Why was this necessary ?

addons/account_edi_ubl_cii/tests/common.py
```diff
@@ -127,13 +127,13 @@ def _assert_invoice_ubl_file(self, invoice, filename):
     # -------------------------------------------------------------------------
 
     @classmethod
-    def _import_as_attachment_on(cls, file_path=None, attachment=None, journal=None):
```
clbr-odoo, line 149: Do we still need this override now ?

clbr-odoo (changes requested):

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -1,13 +1,50 @@
...
-from odoo.tools import frozendict
 
 
 class AccountEdiUBL(models.AbstractModel):
     _name = "account.edi.ubl"
     _inherit = 'account.edi.common'
     _description = "Base helpers for UBL"
 
+    def _import_attachments(self, invoice, tree):
+        """# EXTENDS 'account_edi_common': ATTEMPTS to create a PDF attachment when the XML file doesn't provide one."""
```
clbr-odoo:
```suggestion
        """ EXTENDS 'account_edi_common': ATTEMPTS to create a PDF attachment when the XML file doesn't provide one."""
```

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.py
```diff
@@ -0,0 +1,15 @@
...
+    _description = 'PDF invoice generated by Odoo from an XML file that does not contain one'
+
+    @api.model
+    def _get_report_values(self, docids, data=None):
+        return {
+            'doc_ids': docids,
+            'doc_model': self.env['account.move'],
+            'data': data or {},
+            'docs': self.env['account.move'].browse(docids),
+        }
```
smetl: This is useless. Use the `data` parameter of `_render_qweb_pdf` instead

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <data>
```
smetl:
```suggestion
```

addons/account_edi_ubl_cii/tests/common.py
```diff
@@ -127,13 +127,13 @@ def _assert_invoice_ubl_file(self, invoice, filename):
...
+    def _import_as_attachment_on(cls, file_path=None, attachment=None, journal=None, additional_context={}):
         assert file_path or attachment
         assert not file_path or not attachment
         journal = journal or cls.company_data["default_journal_purchase"]
         if file_path:
             attachment = cls._import_as_attachment(file_path)
-        return journal._create_document_from_attachment(attachment.id)
+        return journal.with_context(additional_context)._create_document_from_attachment(attachment.id)
```
smetl: Never shadow completely an odoo context
```suggestion
        return journal.with_context(**additional_context)._create_document_from_attachment(attachment.id)
```

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -663,3 +664,26 @@ def test_partner_name_in_xml(self):
         xml_tree = etree.fromstring(xml_content)
         partner_name = xml_tree.find('.//cac:AccountingCustomerParty/cac:Party/cac:PartyName/cbc:Name', self.ubl_namespaces)
         self.assertEqual(partner_name.text, 'partner_a')
+
+    def test_generate_pdf_when_xml_does_not_provide_one(self):
+        def _run_wkhtmltopdf(*args, **kwargs):
+            return file_open(f'{self.test_module}/tests/test_files/invoice_example.pdf', 'rb').read()
+
+        file_path = "bis3_bill_example_without_embedded_attachment.xml"
+        file_path = f"{self.test_module}/tests/test_files/{file_path}"
```
smetl:
```suggestion
        file_path = f"{self.test_module}/tests/test_files/bis3_bill_example_without_embedded_attachment.xml"
```

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -663,3 +664,26 @@ def test_partner_name_in_xml(self):
...
+                'mimetype': 'application/xml',
+                'name': 'test_invoice.xml',
+                'raw': file.read(),
+            })
+
+        # Import the document that doesn't contain an embedded PDF
+        with patch.object(self.env.registry['ir.actions.report'], '_run_wkhtmltopdf', _run_wkhtmltopdf):
+            bill = self._import_as_attachment_on(attachment=xml_attachment, additional_context={'force_report_rendering': True})
+
+        self.assertTrue(bill.is_purchase_document())
```
smetl: useless check ? 

```suggestion
        self.assertTrue(bill)
```
sounds better to me

addons/account_edi_ubl_cii/tests/common.py
smetl: However, the whole code here is useless. Just pass `cls.company_data["default_journal_purchase"].with_context(force_report_rendering=True)` as the `journal` parameter

addons/account_edi_ubl_cii/report/account_edi_ubl_cii_report_templates.xml
```diff
@@ -0,0 +1,36 @@
...
+                    <t t-call="account_edi_ubl_cii.report_invoice_document_with_watermark"/>
+                </t>
+            </t>
+        </template>
+
+        <template id="report_invoice_document_with_watermark" inherit_id="account.report_invoice_document" primary="True">
+            <xpath expr="//t[@t-call='web.external_layout']" position="attributes">
+                <attribute name="t-call">account_edi_ubl_cii.external_layout_with_context</attribute>
+            </xpath>
+        </template>
```
smetl: I don't understand why you are talking about a watermark... To me you are just inheriting `account.report_invoice_document` so...

```suggestion
        <template id="account_invoices_generated_by_odoo">
            <t t-call="web.html_container">
                <t t-foreach="docs" t-as="o">
                    <t t-set="lang" t-value="o.partner_id.lang"/>
                    <t t-call="account_edi_ubl_cii.report_invoice_document"/>
                </t>
            </t>
        </template>

        <template id="report_invoice_document" inherit_id="account.report_invoice_document" primary="True">
            <xpath expr="//t[@t-call='web.external_layout']" position="attributes">
                <attribute name="t-call">account_edi_ubl_cii.external_layout_with_context</attribute>
            </xpath>
        </template>
```

jugurtha-gaci: Because in the beginning, we created a CSS watermark in the background, but the PO changed his mind so I got rid of it (but i forgot to adapt the `id` )

addons/account_edi_ubl_cii/__init__.py
```diff
@@ -1,5 +1,6 @@
 from . import models
 from . import wizard
+from . import report
```
clbr-odoo:
```suggestion
```

clbr-odoo (changes requested):

clbr-odoo (approved):

clbr-odoo (approved):
@robodoo r+

PR #245662, opened 2026-01-25
[FIX] mrp_subcontracting_purchase: Prevent singleton error

kihy-odoo:
Related ticket: https://www.odoo.com/odoo/project.task/5416006 (The ticket was made for an unrelated perf issue).
Related PR: https://github.com/odoo/odoo/pull/243742

Reproducing this requires a database with
1. A product with `qty_available`, that is not `lot_valuated`, with a `cost_method` of average. 
2. The product has a stock move with a `production_id` where `is_in` is true.
3. The stock move has move than 1 `move_dest_ids` where `done` and `is_subcontract` is true.
Then:
4. Open an inventory valuation report from any date in the past. One can access inventory valuation reports from Accounting app --> Review menu --> Inventory --> Inventory Valuation. You will get a singleton error.

Note:
Per the related PR, it is no longer possible to give a stock move multiple subcontract move_dest_ids. So, one cannot create new stock moves that will trigger the bug. But also per the PR, this was possible in previous versions, and it is still valid.

Explanation:
In the `_get_value_from_account_move` method of stock.move, `last_subcontract_done_receipt` was changed in the related PR to accommodate the fact that a stock move could have multiple subcontract `move_dest_ids`, preventing a singleton error. However, a singleton error can still occur. It's possible that the `_get_value_from_account_move` method gets invoked on a recordset containing all the subcontract `move_dest_ids`. If there is an "at_date" passed in, it will ask for the date of the record set, causing a singleton error.

Solution:
Limit `last_subcontract_done_receipt` to 1 record.

opw-5416006
---

### DIFF
--- a/addons/mrp_subcontracting_purchase/models/stock_move.py
+++ b/addons/mrp_subcontracting_purchase/models/stock_move.py
@@ -15,7 +15,7 @@ class StockMove(models.Model):
         valuation_data = super()._get_value_from_account_move(quantity, at_date=at_date)
         last_subcontract_done_receipt = self.move_dest_ids.filtered(
             lambda m: m.state == 'done' and m.is_subcontract and m.purchase_line_id
-        )
+        ).sorted('create_date', reverse=True)[:1]
         if not self.production_id or not last_subcontract_done_receipt:
             return valuation_data
 


ticodoo (commented):
Fix itself seems ok, but can you clean up your commit message?

The title should be more specific so people already have an idea of what the commit is doing without reading the body, e.g.:

`[FIX] mrp_subcontracting_purchase: correctly select last subcontract done`

The body should also be more specific and not refer to specific line numbers since it can change over time (it's already incorrect):

```
Followup to previous fix: dae71a2ba5a51ad5a07c4294ecdd6192d8752b07

Because older subcontracting MOs may have multiple destination receipts,
a singleton error will occur if more there is more than one of these 
`last_subcontract_done_receipt` records. This is due to the recordset
calling `_get_value_from_account_move` on itself, which expects only 1
record and results in a singleton access error when it reaches:
https://github.com/odoo/odoo/blob/e7ebe07e1d60659f3d468255808a305e7423741b/addons/purchase_stock/models/stock_move.py#L184

Because we have no way of knowing which receipt move is the absolute last done receipt, we assume it is the last one created that is also 'done'.

{also copy your steps to reproduce from your PR description here}

opw-5416006
```

The github link will automagically show up prettified in the github PR and the commit message will also automatically make itself a link.

PR #245667, opened 2026-01-25
[IMP] orm: stored fields should not depend on context

kmagusiak:
Stored values cannot depend on the context because they have a single
stored value. We have a few exceptions:

- "lang" for translated fields
- "company" for company-dependent fields

task-2903826

Fix x2m related fields after the cache pollution fix odoo/odoo#254944: the related x2m fields need to be computed in sudo to fill the cache with all values.

---


rco-odoo (commented):
@robodoo rebase-ff r+

PR #245670, opened 2026-01-26
[IMP] payment_xendit: error message

nni-odoo:
The error message returned from Xendit when there is input error on phone number is not understandable. Customers often don't know which information has been input incorrectly.

task-5877596

---


livaios (changes requested):
Hello @nni-odoo 

There's a function you can override to modify the error message returned from the provider, here's an example 
https://github.com/odoo/odoo/blob/7436e0cfaf8d4b0c0e0390e8d6a3df404ba240f6/addons/payment_paymob/models/payment_provider.py#L260

PR #245672, opened 2026-01-26
[ADD] l10n_vn_edi_viettel_pos: SInvoice integration for PoS

jand-odoo:
Add Viettel SInvoice integration with the Point of Sale module to support PoS e-invoicing for the Vietnamese market.

task-4844619

### DIFF
--- a/addons/l10n_vn_edi_viettel/models/account_move.py
+++ b/addons/l10n_vn_edi_viettel/models/account_move.py
@@ -12,7 +12,7 @@ from datetime import datetime, timedelta
 import requests
 from requests import RequestException
 
-from odoo import _, api, fields, models
+from odoo import _, api, fields, models, SUPERUSER_ID
 from odoo.exceptions import UserError
 from odoo.tools import float_round, float_repr
 
@@ -306,6 +306,55 @@ class AccountMove(models.Model):
             'raw': file_bytes,
             'res_field': 'l10n_vn_edi_sinvoice_pdf_file',
         }, ""
+
+    def _l10n_vn_edi_fetch_invoice_files(self):
+        """
+        Fetches the SInvoice XML and PDF data from the SInvoice server if self is a sent invoice.
+        The files are saved in the l10n_vn_edi_sinvoice_pdf_file_id and l10n_vn_edi_sinvoice_xml_file_id.
+        """
+
+        if self.l10n_vn_edi_invoice_state != 'sent':
+            raise UserError(_("Please send the invoice to SInvoice before fetching the tax invoice files."))
+
+        xml_data, xml_error_message = self._l10n_vn_edi_fetch_invoice_xml_file_data()
+        pdf_data, pdf_error_message = self._l10n_vn_edi_fetch_invoice_pdf_file_data()
+
+        # Not using _link_invoice_documents for these because it depends on _need_invoice_document and I can't get it to work
+        # well while allowing users to download the files before sending.
+        attachments_data = []
+        for file, error in [(xml_data, xml_error_message), (pdf_data, pdf_error_message)]:
+            if error:
+                continue
+
+            attachments_data.append({
+                'name': file['name'],
+                'raw': file['raw'],
+                'mimetype': file['mimetype'],
+                'res_model': self._name,
+                'res_id': self.id,
+                'res_field': file['res_field'],  # Binary field
+            })
+
+        if attachments_data:
+            attachments = self.env['ir.attachment'].with_user(SUPERUSER_ID).create(attachments_data)
+            self.invalidate_recordset(fnames=[
+                'l10n_vn_edi_sinvoice_xml_file_id',
+                'l10n_vn_edi_sinvoice_xml_file',
+                'l10n_vn_edi_sinvoice_pdf_file_id',
+                'l10n_vn_edi_sinvoice_pdf_file',
+            ])
+
+            # Log the new attachment in the chatter for reference. Make sure to add the JSON file.
+            self.with_context(no_new_invoice=True).message_post(
+                body=_('Invoice sent to SInvoice'),
+                attachment_ids=attachments.ids + self.l10n_vn_edi_sinvoice_file_id.ids,
+            )
+
+        if xml_error_message or pdf_error_message:
+            return {
+                'error_title': _('Error when receiving SInvoice files.'),
+                'errors': [error_message for error_message in [xml_error_message, pdf_error_message] if error_message],
+            }
 
     def action_l10n_vn_edi_update_payment_status(self):
         """ Send a request to update the payment status of the invoice. """
--- a/addons/l10n_vn_edi_viettel/models/account_move.py
+++ b/addons/l10n_vn_edi_viettel/models/account_move.py
@@ -755,7 +755,7 @@ class AccountMove(models.Model):
                 'itemCode': line.product_id.code or '',
                 'itemName': textwrap.shorten(item_name, width=500, placeholder='...'),
                 'unitName': line.product_uom_id.name or 'Units',
-                'unitPrice': line.price_unit * sign,
+                'unitPrice': line.currency_id.round(line.price_unit * sign),
                 'quantity': line.quantity,
                 # This amount should be without discount applied.
                 'itemTotalAmountWithoutTax': line.currency_id.round(line.price_unit * line.quantity),
--- a/addons/l10n_vn_edi_viettel_pos/models/account_move.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/account_move.py
@@ -1,0 +1,31 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = 'account.move'
+
+    def _get_invoice_legal_documents(self, filetype, allow_fallback=False):
+        if filetype == 'pdf' and self.pos_order_ids and (sinvoice_attachment := self.l10n_vn_edi_sinvoice_pdf_file_id):
+            return {
+                'filename': sinvoice_attachment.name,
+                'filetype': 'pdf',
+                'content': sinvoice_attachment.raw,
+            }
+        return super()._get_invoice_legal_documents(filetype, allow_fallback=allow_fallback)
+
+    def _l10n_vn_edi_add_buyer_information(self, json_values):
+        super()._l10n_vn_edi_add_buyer_information(json_values)
+
+        # For Walk-In Customer, there is no address and buyerNotGetInvoice should be set to 1
+        if self.partner_id == self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer', raise_if_not_found=False):
+            del json_values['buyerInfo']['buyerAddressLine']
+            json_values['buyerInfo']['buyerNotGetInvoice'] = 1
+
+    def l10n_vn_edi_fetch_invoice_files(self):
+        """
+        Fetches the SInvoice XML and PDF data from the SInvoice server if self is a sent invoice.
+        The files are saved in the l10n_vn_edi_sinvoice_pdf_file_id and l10n_vn_edi_sinvoice_xml_file_id.
+        """
+        self._l10n_vn_edi_fetch_invoice_files()
--- a/addons/l10n_vn_edi_viettel_pos/models/pos_config.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/pos_config.py
@@ -1,0 +1,35 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, fields, api
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    l10n_vn_auto_send_to_sinvoice = fields.Boolean("Auto-send to SInvoice", default=True)
+    l10n_vn_pos_symbol = fields.Many2one(
+        comodel_name='l10n_vn_edi_viettel.sinvoice.symbol',
+        string='POS Symbol',
+        groups='base.group_system,point_of_sale.group_pos_manager',
+        help='This is the symbol that will be used on invoices issued from this POS.',
+        compute="_compute_l10n_vn_pos_symbol",
+        store=True,
+        readonly=False
+    )
+
+    @api.depends('company_id.l10n_vn_pos_default_symbol')
+    def _compute_l10n_vn_pos_symbol(self):
+        for pos_config in self:
+            if not pos_config.l10n_vn_pos_symbol and pos_config.company_id.l10n_vn_pos_default_symbol:
+                pos_config.l10n_vn_pos_symbol = pos_config.company_id.l10n_vn_pos_default_symbol
+
+    def _load_pos_data(self, data):
+        response = super()._load_pos_data(data)
+
+        company_id = self.env.company
+        if company_id.country_id.code != 'VN':
+            return response
+
+        response['data'][0]['_is_vn_edi_pos_applicable'] = bool(response['data'][0].get('l10n_vn_pos_symbol') or company_id.l10n_vn_pos_default_symbol)
+
+        return response
--- a/addons/l10n_vn_edi_viettel_pos/models/pos_order.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/pos_order.py
@@ -1,0 +1,46 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, fields, api
+
+
+class PosOrder(models.Model):
+    _inherit = 'pos.order'
+
+    l10n_vn_credit_note_reason = fields.Char(string='Credit Note Reason', copy=False)
+    l10n_vn_has_sinvoice_pdf = fields.Boolean(compute="_compute_sinvoice_has_pdf", string="SInvoice PDF Available")
+    l10n_vn_sinvoice_state = fields.Selection(related='account_move.l10n_vn_edi_invoice_state')
+
+    @api.depends("account_move.l10n_vn_edi_sinvoice_pdf_file")
+    def _compute_sinvoice_has_pdf(self):
+        for pos_order in self:
+            pos_order.l10n_vn_has_sinvoice_pdf = bool(pos_order.account_move.l10n_vn_edi_sinvoice_pdf_file)
+
+    def _prepare_invoice_vals(self):
+        vals = super()._prepare_invoice_vals()
+
+        if self.country_code != 'VN' or not self.config_id.l10n_vn_auto_send_to_sinvoice:
+            return vals
+
+        sinvoice_symbol = self.config_id.l10n_vn_pos_symbol or self.config_id.company_id.l10n_vn_pos_default_symbol
+        if sinvoice_symbol:
+            vals['l10n_vn_edi_invoice_symbol'] = sinvoice_symbol.id
+
+            # Refund Invoice (Credit Note)
+            if self.amount_total < 0:
+                vals['l10n_vn_edi_adjustment_type'] = '1'  # Money Adjustment
+
+                reason = self.l10n_vn_credit_note_reason
+                current_ref = vals.get('ref')
+                if current_ref and reason:
+                    vals['ref'] = f"{current_ref}, {reason}"
+
+        return vals
+
+    def _create_invoice(self, move_vals):
+        if self.country_code == 'VN' and self.config_id.l10n_vn_auto_send_to_sinvoice:
+            # When auto-sending to SInvoice, we want to skip fetching the SInvoice files
+            # right after sending the invoice to reduce the time spent in the POS checkout flow.
+            # The SInvoice files will be fetched by printing the invoice from the POS order page
+            # or fetched manually in the backend.
+            return super()._create_invoice(move_vals).with_context(skip_fetch_sinvoice_files=True)
+        return super()._create_invoice(move_vals)
--- a/addons/l10n_vn_edi_viettel_pos/models/pos_session.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/pos_session.py
@@ -1,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class PosSession(models.Model):
+    _inherit = 'pos.session'
+
+    def _load_pos_data(self, models_to_load):
+        data = super()._load_pos_data(models_to_load)
+
+        if self.company_id.country_id.code != 'VN':
+            return data
+
+        walk_in_customer_id = self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer', raise_if_not_found=False).id
+        data['data'][0]['_default_customer_id'] = walk_in_customer_id
+        return data
--- a/addons/l10n_vn_edi_viettel_pos/models/res_company.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/res_company.py
@@ -1,0 +1,12 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, fields
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    l10n_vn_pos_default_symbol = fields.Many2one(
+        comodel_name='l10n_vn_edi_viettel.sinvoice.symbol',
+        string='Default PoS Symbol',
+    )
--- a/addons/l10n_vn_edi_viettel_pos/models/res_partner.py
+++ b/addons/l10n_vn_edi_viettel_pos/models/res_partner.py
@@ -1,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, api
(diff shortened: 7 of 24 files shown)


addons/l10n_vn_edi_viettel_pos/models/account_move.py
```diff
@@ -0,0 +1,22 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = 'account.move'
+
+    def _get_invoice_legal_documents(self, filetype, allow_fallback=False):
+        if sinvoice_attachment := self.l10n_vn_edi_sinvoice_pdf_file_id:
```
vin-odoo:
```suggestion
        if filetype == 'pdf' and sinvoice_attachment := self.l10n_vn_edi_sinvoice_pdf_file_id:
```
It doesn't hurt and will avoid blocking getting other documents type later on if we need.

addons/l10n_vn_edi_viettel_pos/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+                'filetype': 'pdf',
+                'content': sinvoice_attachment.raw,
+            }
+        return super()._get_invoice_legal_documents(filetype, allow_fallback=allow_fallback)
+
+    def _l10n_vn_edi_add_buyer_information(self, json_values):
+        super()._l10n_vn_edi_add_buyer_information(json_values)
+        if self.partner_id == self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer'):
+            del json_values['buyerInfo']['buyerAddressLine']
+            json_values['buyerInfo']['buyerNotGetInvoice'] = 1
```
vin-odoo, line 24: Wouldn't hurt to leave a docstring to explain why this is done.

addons/l10n_vn_edi_viettel_pos/models/pos_config.py
```diff
@@ -0,0 +1,15 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, fields
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    l10n_vn_auto_send_to_sinvoice = fields.Boolean("Auto-send to SInvoice", default=True)
+    l10n_vn_pos_symbol = fields.Many2one(
```
vin-odoo, line 10: Wouldn't we want this to default to the value of `l10n_vn_pos_default_symbol`?

addons/l10n_vn_edi_viettel_pos/models/pos_config.py
```diff
@@ -0,0 +1,15 @@
...
+        comodel_name='l10n_vn_edi_viettel.sinvoice.symbol',
+        string='POS Symbol',
+        groups='base.group_system',
```
vin-odoo: I would say users who can manage the configuration of the PoS (journal, currency, etc) should also be able to choose the symbol for the same PoS no?

addons/l10n_vn_edi_viettel_pos/models/pos_session.py
```diff
@@ -0,0 +1,17 @@
...
+class PosSession(models.Model):
+    _inherit = 'pos.session'
+
+    def _load_pos_data(self, models_to_load):
+        data = super()._load_pos_data(models_to_load)
+
+        if self.company_id.country_id.code != 'VN':
+            return data
+
+        walk_in_customer_id = self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer').id
```
vin-odoo: Mind that users could delete the walk in customer by mistake, in which case this would break the PoS.
I would add a 'raise_if_not_found=False' here.

addons/l10n_vn_edi_viettel_pos/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+            return {
+                'filename': sinvoice_attachment.name,
+                'filetype': 'pdf',
+                'content': sinvoice_attachment.raw,
+            }
+        return super()._get_invoice_legal_documents(filetype, allow_fallback=allow_fallback)
+
+    def _l10n_vn_edi_add_buyer_information(self, json_values):
+        super()._l10n_vn_edi_add_buyer_information(json_values)
+        if self.partner_id == self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer'):
```
vin-odoo: Set raise_if_not_found=False. If the record was deleted, it wouldn't work anymore but at least wouldn't break.
Users could update the module to recreate the record I believe.

addons/l10n_vn_edi_viettel_pos/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.model
+    def _load_pos_data_domain(self, data):
+        # Make sure to always load the walk-in customer
+        domain = super()._load_pos_data_domain(data)
+        walk_in_customer_id = self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer').id
```
vin-odoo: Same here, if you don't find the ref don't raise

addons/l10n_vn_edi_viettel_pos/tests/test_l10n_vn_edi_pos.py
```diff
@@ -0,0 +1,128 @@
...
+        """ Test the refund flow of PoS order"""
+        order = self._create_simple_order()
+        move_vals = order._prepare_invoice_vals()
+        invoice = order._create_invoice(move_vals)
+        invoice.action_post()
+        self._send_invoice(invoice)
+
+        refund_order = order._refund()
+        refund_move_vals = refund_order._prepare_invoice_vals()
+        refund_move_vals.update({'l10n_vn_edi_adjustment_type': '1'})
```
vin-odoo: Would be better to set the values on the refund_order before calling _prepare_invoice_vals so that you test the code handling that

addons/l10n_vn_edi_viettel_pos/static/src/overrides/components/payment_screen/payment_screen.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+        super.toggleIsToInvoice(...arguments);
+    },
+    async _isOrderValid(isForceValidate) {
+        const result = await super._isOrderValid(...arguments);
+        if (
+            this.pos.isVietnamCompany() &&
+            this.pos.config.l10n_vn_auto_send_to_sinvoice &&
+            this.currentOrder.is_to_invoice() &&
+            (this.pos.config.l10n_vn_pos_symbol || this.pos.config.company_id.l10n_vn_pos_default_symbol)
```
vin-odoo: Are you adding support for the symbols etc in the PoS (and loading them) just for this check?
It's a lot of added complexity for a if; you could simply give to the PoS the info. (give a bool in load_data with the result of this condition)
This would lead to the same result but reduce a bit the amount of code needed for this.

vin-odoo (changes requested):
If I understood correctly this is mostly ready for review; so here is a first round :)
Once everything is finalized with TRTR, I'll do another round of review and will test the feature at the same time. 
Thanks for the work!

addons/l10n_vn_edi_viettel_pos/models/res_company.py
```diff
@@ -0,0 +1,16 @@
...
+    _inherit = 'res.company'
+
+    l10n_vn_pos_default_symbol = fields.Many2one(
+        comodel_name='l10n_vn_edi_viettel.sinvoice.symbol',
+        string='Default PoS Symbol',
+    )
+
+    @api.model
+    def _load_pos_data_fields(self, config_id):
+        return super()._load_pos_data_fields(config_id) + ['l10n_vn_pos_default_symbol']
```
vin-odoo: Maybe only load that for VN companies

addons/l10n_vn_edi_viettel_pos/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.model
+    def _load_pos_data_domain(self, data):
+        # Make sure to always load the walk-in customer
+        domain = super()._load_pos_data_domain(data)
+        walk_in_customer_id = self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer', raise_if_not_found=False).id
+        return expression.OR([domain, [('id', '=', walk_in_customer_id)]])
```
vin-odoo: Only for vn companies

addons/l10n_vn_edi_viettel_pos/models/pos_order.py
```diff
@@ -0,0 +1,42 @@
...
+                current_ref = vals.get('ref')
+                if current_ref and reason:
+                    vals['ref'] = f"{current_ref}, {reason}"
+
+        return vals
+
+    def _create_invoice(self, move_vals):
+        if self.country_code == 'VN' and self.config_id.l10n_vn_auto_send_to_sinvoice:
+            return super()._create_invoice(move_vals).with_context(skip_fetch_sinvoice_files=True)
+        return super()._create_invoice(move_vals)
```
vin-odoo, line 46: Add a docstring or comment to explain why this is needed.

addons/l10n_vn_edi_viettel_pos/models/pos_order.py
```diff
@@ -0,0 +1,42 @@
...
+    _inherit = 'pos.order'
+
+    credit_note_reason = fields.Char(string='Credit Note Reason', copy=False)
+    l10n_vn_has_sinvoice_pdf = fields.Boolean(compute="_compute_sinvoice_has_pdf", string="SInvoice PDF Available")
+    l10n_vn_sinvoice_state = fields.Selection(related='account_move.l10n_vn_edi_invoice_state')
+
+    @api.depends("account_move.l10n_vn_edi_sinvoice_pdf_file")
+    def _compute_sinvoice_has_pdf(self):
+        for record in self:
+            record['l10n_vn_has_sinvoice_pdf'] = bool(record.account_move.l10n_vn_edi_sinvoice_pdf_file)
```
vin-odoo:
```suggestion
            record.l10n_vn_has_sinvoice_pdf = bool(record.account_move.l10n_vn_edi_sinvoice_pdf_file)
```
Usually it's better to call the variable of the loop by the model 'name' and not 'record' (e.g. pos_order)

addons/l10n_vn_edi_viettel_pos/models/pos_config.py
```diff
@@ -0,0 +1,35 @@
...
+        compute="_compute_l10n_vn_pos_symbol",
+        store=True,
+        readonly=False
+    )
+
+    @api.depends('company_id.l10n_vn_pos_default_symbol')
+    def _compute_l10n_vn_pos_symbol(self):
+        for record in self:
+            if not record.l10n_vn_pos_symbol and record.company_id.l10n_vn_pos_default_symbol:
+                record['l10n_vn_pos_symbol'] = self.company_id.l10n_vn_pos_default_symbol
```
vin-odoo:
```suggestion
                record.l10n_vn_pos_symbol = self.company_id.l10n_vn_pos_default_symbol
```

addons/l10n_vn_edi_viettel_pos/models/account_move_send.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class AccountMoveSend(models.AbstractModel):
+    _inherit = 'account.move.send'
+
+    @api.model
+    def _is_vn_edi_applicable(self, move):
+        res = super()._is_vn_edi_applicable(move)
+
+        if not move.sudo().pos_order_ids:
```
vin-odoo: Why do you need to sudo here?

addons/l10n_vn_edi_viettel_pos/views/account_move_views.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="state">code</field>
+        <field name="code">
+if records.l10n_vn_edi_invoice_state != 'sent':
+    raise UserError("Please send the invoice to SInvoice before fetching the tax invoice files.")
+if (
+    records.l10n_vn_edi_sinvoice_pdf_file_id
+    and records.l10n_vn_edi_sinvoice_xml_file_id
+    and records.l10n_vn_edi_sinvoice_file_id
+):
+    raise UserError("The SInvoice files have already been generated for this invoice.")
```
vin-odoo: Why not do the checks in the l10n_vn_edi_fetch_invoice_files method 🤔

vin-odoo (commented):
I think as you have quite a bit of logic added in the frontend, you'll need to add some javascript tests for it.

Another partial review, will focus more on this once everything for the freeze is finalized

addons/l10n_vn_edi_viettel_pos/models/account_move_send.py
jand-odoo: I think previously I got an error about access right issue. But then i already modify a lot of things and forgot to remove it. i think without sudo works now :)

jand-odoo: I add sudo because there is a case where a user doesn't have access to pos.order model. And i want to make sure that the _is_vn_edi_applicable calculation is always succesful.

kidoodoo:
re-requested so that it lands on Nicolas's inbox

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -307,6 +307,60 @@ def _l10n_vn_edi_fetch_invoice_pdf_file_data(self):
             'res_field': 'l10n_vn_edi_sinvoice_pdf_file',
         }, ""
 
+    def l10n_vn_edi_fetch_invoice_files(self):
```
vin-odoo: You should 'backport' the method you made in the master task to align behaviors and avoid conflicts.

https://github.com/odoo/odoo/blob/master/addons/l10n_vn_edi_viettel/models/account_move.py#L309

As you need a public method to call from the PoS, you can simply add another public l10n_vn_edi_fetch_invoice_files that calls the private one from master (maybe adding the public one in the PoS module only)

vin-odoo: That said, be careful that you are not properly translating your errors here :)

addons/l10n_vn_edi_viettel_pos/models/account_move.py
```diff
@@ -0,0 +1,49 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = 'account.move'
+
+    def _get_invoice_legal_documents(self, filetype, allow_fallback=False):
```
vin-odoo, line 9: I would limit this change to only affect PoS flows.
If we want to apply to all flows, it should be done in master.

addons/l10n_vn_edi_viettel_pos/models/account_move.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _l10n_vn_edi_add_buyer_information(self, json_values):
+        super()._l10n_vn_edi_add_buyer_information(json_values)
+
+        # For Walk-In Customer, there is no address and buyerNotGetInvoice should be set to 1
+        if self.partner_id == self.env.ref('l10n_vn_edi_viettel_pos.partner_walk_in_customer', raise_if_not_found=False):
+            del json_values['buyerInfo']['buyerAddressLine']
+            json_values['buyerInfo']['buyerNotGetInvoice'] = 1
+
+    def _l10n_vn_edi_add_item_information(self, json_values):
```
vin-odoo: Todo: probably can be removed once the other task is merged.

addons/l10n_vn_edi_viettel_pos/models/res_company.py
```diff
@@ -0,0 +1,19 @@
...
+        comodel_name='l10n_vn_edi_viettel.sinvoice.symbol',
+        string='Default PoS Symbol',
+    )
+
+    @api.model
+    def _load_pos_data_fields(self, config_id):
+        result = super()._load_pos_data_fields(config_id)
+        if self.env.company.country_id.code == "VN":
+            return result + ['l10n_vn_pos_default_symbol']
+        return result
```
vin-odoo: I think you never use it in the PoS, so you don't need to send it.

addons/l10n_vn_edi_viettel_pos/static/src/overrides/models/pos_store.js
```diff
@@ -0,0 +1,10 @@
+/** @odoo-module */
+
+import { PosStore } from "@point_of_sale/app/store/pos_store";
+import { patch } from "@web/core/utils/patch";
+
+patch(PosStore.prototype, {
+    isVietnamCompany() {
+        return this.company.country_id?.code == "VN";
```
vin-odoo:
```suggestion
        return this.company.account_fiscal_country_id?.code === "VN";
```
Look at the fiscal country :)

addons/l10n_vn_edi_viettel_pos/static/src/overrides/models/pos_order.js
```diff
@@ -0,0 +1,24 @@
...
+            }
+
+            const defaultPartner = this.models["res.partner"].get(
+                this.session._default_customer_id
+            );
+            this.partner_id = this.partner_id || defaultPartner;
+        }
+    },
+    isVietnamCompany() {
+        return this.company.country_id?.code == "VN";
```
vin-odoo:
```suggestion
        return this.company.account_fiscal_country_id?.code == "VN";
```
Do we need to have this helper twice? 🤔 (Not sure if there are cases where the pos store country != the same as the order's country)

vin-odoo (changes requested):
While we're waiting to see if the other teams react to the discount task, here is a small review :)

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -307,6 +307,53 @@ def _l10n_vn_edi_fetch_invoice_pdf_file_data(self):
             'res_field': 'l10n_vn_edi_sinvoice_pdf_file',
         }, ""
 
+    def _l10n_vn_edi_fetch_invoice_files(self):
+
```
vin-odoo, line 315: A docstring could be beneficial here, to be clear that we don't return the files.
For example
```suggestion
    def _l10n_vn_edi_fetch_invoice_files(self):
    """
    Fetches the SInvoice XML and PDF data from the SInvoice server if self is a sent invoice.
    The files are saved in the l10n_vn_edi_sinvoice_pdf_file_id and l10n_vn_edi_sinvoice_xml_file_id.
    """
```

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -307,6 +307,53 @@ def _l10n_vn_edi_fetch_invoice_pdf_file_data(self):
...
+        if self.l10n_vn_edi_invoice_state != 'sent':
+            raise UserError(_("Please send the invoice to SInvoice before fetching the tax invoice files."))
+
+        # Download SInvoice documents in order to attach them to the email we sent to the customer.
+        # If the email is not being sent, we will still get the files and attach them to the invoice.
```
vin-odoo: I think this comment should be removed or be kept in account_move_send

addons/l10n_vn_edi_viettel_pos/static/src/overrides/models/pos_order.js
```diff
@@ -0,0 +1,21 @@
+/** @odoo-module */
+
+import { PosOrder } from "@point_of_sale/app/models/pos_order";
+import { patch } from "@web/core/utils/patch";
+
+patch(PosOrder.prototype, {
+    setup(vals) {
+        super.setup(...arguments);
+        if (this.company.account_fiscal_country_id?.code == "VN") {
+            // Set default to_invoce to true except for return order
```
vin-odoo:
```suggestion
            // Set to_invoice to true by default except for return order
```

vin-odoo (commented):
Few nitpicks.
The pr looks good to me at this stage. I suggest doing another round of testing with the PO after the changes.
I'll do a round of testing too on my side :)
Thanks!

addons/l10n_vn_edi_viettel_pos/views/account_move_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="l10n_vn_edi_pos_fetch_tax_invoice" model="ir.actions.server">
+        <field name="name">Fetch Tax Invoice</field>
+        <field name="model_id" ref="account.model_account_move"/>
+        <field name="binding_model_id" ref="account.model_account_move"/>
+        <field name="binding_view_types">form</field>
+        <field name="state">code</field>
+        <field name="code">
+records.l10n_vn_edi_fetch_invoice_files()
```
vin-odoo:
```suggestion
records._l10n_vn_edi_fetch_invoice_files()
```
To update this method and related calls as discussed on discord

addons/l10n_vn_edi_viettel_pos/static/src/overrides/components/invoice_button/invoice_button.js
```diff
@@ -0,0 +1,30 @@
...
+
+patch(InvoiceButton.prototype, {
+    async _downloadInvoice(orderId) {
+        const orderWithInvoice = await this.pos.data.read("pos.order", [orderId], [], {
+            load: false,
+        });
+        const order = orderWithInvoice[0];
+        if (
+            this.pos.isVietnamCompany() &&
+            order.l10n_vn_sinvoice_state == "sent" &&
```
vin-odoo:
```suggestion
            order.l10n_vn_sinvoice_state === "sent" &&
```

addons/l10n_vn_edi_viettel_pos/static/src/overrides/components/invoice_button/invoice_button.js
```diff
@@ -0,0 +1,30 @@
...
+            await this.pos.data.call("account.move", "l10n_vn_edi_fetch_invoice_files", [
+                order.raw.account_move,
+            ]);
+        }
+        return await super._downloadInvoice(...arguments);
+    },
+    get commandName() {
+        if (this.pos.isVietnamCompany() && this.pos.config.l10n_vn_auto_send_to_sinvoice) {
+            const name = super.commandName;
+            return name == _t("Reprint Invoice") ? _t("Print Tax Invoice") : name;
```
vin-odoo:
```suggestion
            return name === _t("Reprint Invoice") ? _t("Print Tax Invoice") : name;
```

addons/l10n_vn_edi_viettel_pos/static/src/overrides/models/pos_order.js
```diff
@@ -0,0 +1,21 @@
+/** @odoo-module */
+
+import { PosOrder } from "@point_of_sale/app/models/pos_order";
+import { patch } from "@web/core/utils/patch";
+
+patch(PosOrder.prototype, {
+    setup(vals) {
+        super.setup(...arguments);
+        if (this.company.account_fiscal_country_id?.code == "VN") {
```
vin-odoo:
```suggestion
        if (this.company.account_fiscal_country_id?.code === "VN") {
```

addons/l10n_vn_edi_viettel_pos/static/src/overrides/models/pos_store.js
```diff
@@ -0,0 +1,10 @@
+/** @odoo-module */
+
+import { PosStore } from "@point_of_sale/app/store/pos_store";
+import { patch } from "@web/core/utils/patch";
+
+patch(PosStore.prototype, {
+    isVietnamCompany() {
+        return this.company.account_fiscal_country_id?.code == "VN";
```
vin-odoo:
```suggestion
        return this.company.account_fiscal_country_id?.code === "VN";
```

vin-odoo (commented):
small nitpick on your equals in js.

addons/l10n_vn_edi_viettel_pos/models/pos_order.py
```diff
@@ -0,0 +1,46 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models, fields, api
+
+
+class PosOrder(models.Model):
+    _inherit = 'pos.order'
+
+    credit_note_reason = fields.Char(string='Credit Note Reason', copy=False)
```
vin-odoo:
```suggestion
    l10n_vn_credit_note_reason = fields.Char(string='Credit Note Reason', copy=False)
```

vin-odoo (commented):
Small change to do, otherwise LGTM.

vin-odoo (approved):
Thank you!

vin-odoo:
robodoo delegate+

I'm away next week, and the chain can't be ready today, so Jonathan will handle merging next week after the weekly update, once everything is ready.
Thanks!

PR #245675, opened 2026-01-26
[FIX] l10n_it: Wrong exoneration code on 0% EU S

Megaaaaaa:
In Italy, the code depends strictly on whether the transaction is for Goods or Services.

N3.2 is for Intra-community supply of GOODS (Cessioni Intracomunitarie di beni)
N2.1 is for Intra-community supply of SERVICES (Prestazioni di Servizi)

This commit fixes the exoneration code on the 0% EU S tax from N3.2 to N2.1.

task-5870894

---


epictete (commented):
Not sure we should still be merging fixes in `16.0` :thinking: (see [here](https://github.com/odoo/odoo/wiki/Contributing#against-which-version-should-i-submit-a-patch))

We might also need to write a local migration script to make sure the changes are also applied on existing databases, wdyt?

Megaaaaaa:
Hey @epictete :wave:

Imo that stable policy part of the wiki is only the bugfix team's point of view on the matter and as long as there is a running build of 16.0 on runbot and that we have clients that pay to get support in that version, we should keep fixing it. Even more if it concerns legal data. It spares the hassle of having and entire new process of dev-test-review-deployment for a fix that has already been approved if a client asks us to fix it in an earlier version.

As per the migration script, you are totally right, I've added a small script that updates the value after module upgrade. I've only overwritten `l10n_it_kind_exoneration` but it might be a good occasion to rewrite the three values and make sure they are correct, what would you prefer ?

```py
service_tax.sudo().write({
    'l10n_it_has_exoneration': True,
    'l10n_it_kind_exoneration': 'N2.1',
    'l10n_it_law_reference': 'Art. 7ter, DPR 633/1972',
})
```
maybe even on both 0% EU S and 0% EU G as one might argue that this restores the accuracy of both taxes.

addons/l10n_it/migrations/0.9/post-migrate.py
```diff
@@ -0,0 +1,13 @@
...
+    cr.execute("""
+        UPDATE account_tax
+           SET l10n_it_exempt_reason = CASE
+               WHEN amount = 0.0
+                AND type_tax_use = 'sale'
+                AND tax_scope = 'service'
+               THEN 'N2.1'
+               ELSE l10n_it_exempt_reason
+           END
+         WHERE l10n_it_exempt_reason IN ('N3.2')
```
epictete:
```suggestion
        UPDATE account_tax
        SET l10n_it_exempt_reason = 'N2.1'
        WHERE l10n_it_exempt_reason = 'N3.2'
            AND amount = 0.0
            AND type_tax_use = 'sale'
            AND tax_scope = 'service'
```
While limited, I think this would be better in terms of performance.
We don't really need/want to evaluate the `CASE` logic for every `N3.2` row.
Plus we can leverage indexes in the `WHERE` clause.
And there's no need to rewrite the existing value if it is unchanged :nerd_face: 
What do you think (I did not test it)?

Megaaaaaa: Tested it, works like a charm :+1: And yeah without the CASE, no need to rewrite the existing values; Thanks! (added the sequence too)

epictete (approved):
Thanks for the update :ok_hand:

@robodoo r+

PR #245678, opened 2026-01-26
[FIX] account_peppol: connection fails in demo mode

gawa-odoo:
There is one argument too much for the mock of _create_connection compared to the real one.

---


clbr-odoo (approved):
@robodoo r+

PR #245679, opened 2026-01-26
[FIX] sale{,_stock}: only consider non-available quantity as stock in transit

MaximeNoirhomme229:
**Issue**
It is possible to have unreserved quantity, outside the stock location, considered as 'stock in transit' in the forecast report.
This is an issue because that means some quantities, that are not meant to be moved to the stock location, are linked to a delivery.

**Steps to reproduce**
- Activate 3 step receipt on your warehouse 
- Create a new product 'Test' and adds 1 Unit in WH/Input and 1 Unit in WH/Quality 
- Create a sales order for 2 Units and check the forecast report of Test
-> 2 units are recorded as 'stock in transit' links to the so delivery while they should not.

**Cause**
The forecast report is computed here:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/stock/report/stock_forecasted.py#L17-L24
The lines are computed here:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/stock/report/stock_forecasted.py#L125

Currently, stock in transit is calculated by:

Aggregating quantities per product and location (including unreserved quantities):
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/stock/report/stock_forecasted.py#L327-L328

Removing the reserved stock, then summing per product:
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/stock/report/stock_forecasted.py#L353-L354

Assigning the result as 'stock in transit' (since free_stock is 0 here):
https://github.com/odoo/odoo/blob/f0196258ed41f56790db94747192545ae56c78b5/addons/stock/report/stock_forecasted.py#L361

This method incorrectly includes unreserved quantities from intermediate locations.

**Solution**
Since we only want non-available quantity to be considered as stock in transit, we use the reserved quantity instead.
No need to substract the free quantity anymore since free quantity are no reserved by definition.

opw-5865378

### DIFF
--- a/addons/sale_stock/tests/test_sale_stock_report.py
+++ b/addons/sale_stock/tests/test_sale_stock_report.py
@@ -135,6 +135,77 @@ class TestSaleStockReports(TestReportsCommon):
             (2.0, True, False)
         )
 
+    def test_report_forecast_in_transit_for_unreserved_qty(self):
+        """
+        Check that unreserve quantities are not tagged as "in transit" in forecasted report
+        """
+        warehouse = self.env.ref("stock.warehouse0")
+        warehouse.write({'reception_steps': 'three_steps'})
+
+        location_stock, location_quality, location_input = (
+            warehouse.lot_stock_id,
+            warehouse.wh_qc_stock_loc_id,
+            warehouse.wh_input_stock_loc_id,
+        )
+
+        self.env['stock.quant'].create([
+            {
+                'product_id': self.product.id,
+                'location_id': location.id,
+                'quantity': qty,
+            }
+            for location, qty in (
+                (location_stock, 5),
+                (location_quality, 1),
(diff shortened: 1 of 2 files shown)


lase-odoo:
Thanks for the proposition ! It is a really great commit message with super clear explanations. However, altering this part of the code is a risky change and the current behavior is IMO correct. The Input and quality locations are transit locations, the rules of the 3 steps receipts from these are even in pull & push in 17.0 and it makes a lot of sense to consider them as transit quantities. Hence, if a change is proposed, it is a functional design choice and it should not be considered for a fix for stability reasons. A feedback can be raised in master if a functional change is really desired.

PR #245680, opened 2026-01-26
[FIX] web_tour: avoid interacting with triggers behind modals

pipu-odoo:
Before this commit, the automatic tour engine checked whether the DOM
contained a `.modal` and whether the trigger element was inside it, in
order to avoid interacting with elements behind a modal. However, when
the DOM also contained a popover or a notification, this check was
skipped.

With this commit, this logic has been improved to also take other
overlay items into account.

This improvement involves some fixes and adaptations in a few tours.

### DIFF
--- a/addons/web_tour/static/src/js/tour_automatic/tour_step_automatic.js
+++ b/addons/web_tour/static/src/js/tour_automatic/tour_step_automatic.js
@@ -160,20 +160,51 @@ export class TourStepAutomatic extends TourStep {
             : true;
     }
 
+    /**
+     * When a modal is in the overlay and that the current step has an action,
+     * this method checks if the trigger element is in the more front overlay.
+     */
     get elementIsInModal() {
-        if (this.hasAction) {
-            const overlays = hoot.queryFirst(
-                ".popover, .o-we-command, .o-we-toolbar, .o_notification"
-            );
-            const modal = hoot.queryFirst(".modal:visible:not(.o_inactive_modal):last");
-            if (modal && !overlays && !this.trigger.startsWith("body")) {
-                return (
-                    modal.contains(hoot.getParentFrame(this.element)) ||
-                    modal.contains(this.element)
-                );
+        function isIn(element, parent) {
+            if (!parent) {
+                return false;
             }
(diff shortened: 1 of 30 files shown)


addons/web_tour/static/src/tour_utils.js
```diff
@@ -220,6 +220,23 @@ export const stepUtils = {
         };
     },
 
+    waitIfModalAndCloseIt({ title, timeout = 10000 } = {}) {
```
aab-odoo: remove

addons/web_tour/static/src/js/tour_automatic/tour_step_automatic.js
```diff
@@ -160,20 +160,51 @@ export class TourStepAutomatic extends TourStep {
...
+                return false;
+            }
+            return parent.contains(hoot.getParentFrame(element)) || parent.contains(element);
+        }
+
+        if (!this.hasAction) {
+            return true;
+        }
+        const modal = hoot.queryFirst(".modal:visible:not(.o_inactive_modal):last");
+        if (!modal || this.trigger.startsWith("body")) {
```
aab-odoo, line 179: remove startsWith("body')

addons/point_of_sale/static/tests/pos/tours/generic_tour.js
```diff
@@ -26,6 +26,19 @@ registry.category("web_tour.tours").add("generic_localization_tour", {
                 content: "feedback screen has finished the validation",
                 trigger: ".feedback-screen .button.validation:not([disabled])",
             },
+            {
+                content: "Close the modal that appears with l10n_sa_edi_pos",
```
aab-odoo: add this step from l10n_sa_edi_pos

addons/web_tour/static/src/js/tour_automatic/tour_step_automatic.js
aab-odoo: might break some tours that voluntarily use the body prefix to act on an element behind a modal => will be removed later, in another PR

aab-odoo (approved):
robodoo r+

PR #245682, opened 2026-01-26
[FIX] hr_holidays: fix accrual test for future leaves

SurryaT10:
Following the fix in PR: [odoo/odoo/pull/244202](https://github.com/odoo/odoo/pull/244202), _get_future_leaves_on failed to correctly compute the interval between dates, resulting in 0 accrued hours in specific environments.

This commit
fixes the test to be future-proof by removing reliance on demo data and fixing the date context.

Runbot build error: https://runbot.odoo.com/runbot/build/98933084
Task: 5875999

---


amah-odoo (approved):
Thanks for your work!

@robodoo r+

PR #227688, opened 2025-09-17
[FIX] payment_stripe: handle special currency decimal cases

dija-odoo:
Steps:
- Activate the 'UGX' currency.
- Make a sale order with amount 100 with 'UGX' currency
- Try to pay that order (100 USh) using card

Issue:
- stripe throws the following error - 

> 'The Checkout Session's total amount must convert to at least 50 cents. 1.00 USh converts to approximately €0.00.'

- Hence 100 USh sent was identified as 1 USh by stripe. This confirms issue with decimals and currency mapping.

Cause:
- 'UGX' is zero-decimal currency but stripe identify it as two-decimal.

Fix:
- Update mapping for such special currency cases for stripe that don't follow general rules

opw-5075707

PR #245689, opened 2026-01-26
[FIX] payment_stripe: use stripe special rounding in amount validation

mma1377:
### Issue:
Stripe payment is failing in some special currencies (e.g. ISK, UGX).

### Steps to reproduce:
1- Configure Stripe
2- Enable ISK currency
3- Create a SO with ISK currency and generate a payment.
4- Pay using Stripe.

The payment will fail due to the amount mismatch.

### Cause:
Commit https://github.com/odoo/odoo/commit/9493475273a40320228b92940be15ad8cb0643cb added support for special currency but did not adapt amount validation method in fw.

In stripe payload, we are using `const.CURRENCY_DECIMALS` as `arbitrary_decimal_number` to convert the amount to minor currency unit:
https://github.com/odoo/odoo/blob/6e2a29bbb23a8233132dc039d144d97be4feb3af/addons/payment_stripe/models/payment_transaction.py#L161-L166

However, we are not using the same constant in conversion back to major currency unit in `compare_notification_data`:
https://github.com/odoo/odoo/blob/6e2a29bbb23a8233132dc039d144d97be4feb3af/addons/payment_stripe/models/payment_transaction.py#L393-L395

opw-5871420

### DIFF
--- a/addons/payment_stripe/models/payment_transaction.py
+++ b/addons/payment_stripe/models/payment_transaction.py
@@ -391,7 +391,9 @@ class PaymentTransaction(models.Model):
         else:  # 'online_direct', 'online_token', 'offline'
             payment_data = notification_data['payment_intent']
         amount = payment_utils.to_major_currency_units(
-            payment_data.get('amount', 0), self.currency_id
+            payment_data.get('amount', 0),
+            self.currency_id,
+            arbitrary_decimal_number=const.CURRENCY_DECIMALS.get(self.currency_id.name),
         )
         currency_code = payment_data.get('currency', '').upper()
         precision_digits = CURRENCY_MINOR_UNITS.get(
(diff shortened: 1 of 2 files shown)


livaios (changes requested):
Hello @mma1377 
Thanks for investigating and creating the fix.
It seems like this issue exists from 17.0 https://github.com/odoo/odoo/pull/227688 can you just retarget your PR?

livaios (approved):
@vchu-odoo Oh you're right, then the fix should be good to go then 😄

vchu-odoo (commented):
Thank you for the fix!
LGTM :+1: 

Could you please add a test for that?
I know it is a small thing, but it is easy to forget about it, and a test will help a lot.

vchu-odoo (approved):
Thank you for the changes!

We need to adapt it in 19.0 since we refactored payment.

PR #238472, opened 2025-12-03
[IMP] im_livechat: persist info panel toggle across livechat sessions

nmak-odoo:
**Purpose of this PR:**

Previously, the livechat info panel would always open by default when switching between chats, regardless of the user's preference. This required users to manually close the panel repeatedly.

This commit persists the panel's toggle state, so the panel remains open or closed based on the user's last choice across all livechat sessions.

Task-5291258

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245697, opened 2026-01-26
[FIX] im_livechat: prevent local storage loop

tsm-odoo:
Since [1], the live chat info panel state is saved to the
local storage. Writing the livechat info panel state to local
storage on every field change (especially coming from the storage
event itself) caused a retroaction loop across tabs, leading to
potential freezes.

For example:
- Tab A writes OPEN to local storage.
- Tab B receives OPEN and updates its field.
- Tab A writes CLOSE, local storage updated.
- Tab B, based on stale state, writes OPEN back to local storage.
- Tab A receives OPEN, updates its field, writes CLOSE again.

This PR fixes the issue: the field is only written on direct user
action and the computeed field is invalidated on storage event,
effectively breaking the loop.

[1]: https://github.com/odoo/odoo/pull/238472


alexkuhn (approved):
@robodoo delegate+ r+

PR #229546, opened 2025-10-01
[FIX] mail: mention links should not be editable

zel-odoo:
This fixes the issue that when editing a message with mentions, the mentions were editable and could be partially deleted.

This is fixed by preprocessing the message body when loading it for edition, addding contenteditable="false" and draggable="false" to the mention links.

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245713, opened 2026-01-26
[REM] mail: remove redundant edit() override

maki-odoo:
The override of the message `edit()` in discuss became a pass through following changes in PR #229546. This change removes this dead code. Considering stable policies, the change is only made in master.


alexkuhn (approved):
@robodoo delegate+ r+

PR #245714, opened 2026-01-26
[FIX] purchase_stock: use manual invoice currency rate in valuation

plha-odoo:
**Problem:**
The value of a move does not take the correct currency rate if it 
was changed manually on the invoice.

**Steps to reproduce:**
- make sure dollar is your main currency 
- activate euro and set a rate of 1 USD -> 2 euros
- create an avco perpetual storable product with no tax
- confirm a purchase order for 1 product at 100 euros with no tax
- validate move
- create a bill for this purchase order
- set the currency to euro 
- below 'EUR' change the currency to 4 
(the bug also happens if you open the widget calendar and 
select a date with another rate than the invoice date) 
- in the journal items tab you should see the value going 
from 50 to 25 for the account move lines.
- confirm the invoice
- open stock and search your product 

**Current behavior:**
the unit price is 50 

**Expected behavior:**
it should be 25

**Cause of the issue:**
In _get_value_from_account_move when computing the value, 
we convert the value of the account move line using the _convert 
method on the currency at the date of the 
invoice, but if the rate was changed manually this is no longer correct. 
https://github.com/odoo/odoo/blob/4ad397e94b4661cd64515e137b3e650dc330269c/addons/purchase_stock/models/stock_move.py#L170

**fix**
When manually changing the rate on the widget, the account move lines 
are updated using currency_rate 
https://github.com/odoo/odoo/blob/8d7194128b26a09be75bf5bfe4e63c8bbfaac69f/addons/account/models/account_move_line.py#L1634
Using the same compution for the valuation of the move ensures
alignment.

currency_rate is computed based on invoice_currency_rate which 
is updated in case the rate was set manually.
https://github.com/odoo/odoo/blob/4ad397e94b4661cd64515e137b3e650dc330269c/addons/account/models/account_move_line.py#L726

opw-5416654

### DIFF
--- a/addons/purchase_stock/models/stock_move.py
+++ b/addons/purchase_stock/models/stock_move.py
@@ -167,10 +167,10 @@ class StockMove(models.Model):
             aml_ids.add(aml.id)
             if aml.move_type == 'in_invoice':
                 aml_quantity += aml.product_uom_id._compute_quantity(aml.quantity, self.product_id.uom_id)
-                value += aml.currency_id._convert(aml.price_subtotal, self.company_id.currency_id, date=aml.date)
+                value += aml.company_id.currency_id.round(aml.price_subtotal / aml.currency_rate)
             elif aml.move_type == 'in_refund':
                 aml_quantity -= aml.product_uom_id._compute_quantity(aml.quantity, self.product_id.uom_id)
-                value -= aml.currency_id._convert(aml.price_subtotal, self.company_id.currency_id, date=aml.date)
+                value -= aml.company_id.currency_id.round(aml.price_subtotal / aml.currency_rate)
 
         if aml_quantity <= 0:
             return valuation_data
--- a/addons/purchase_stock/tests/test_stockvaluation.py
+++ b/addons/purchase_stock/tests/test_stockvaluation.py
@@ -2078,3 +2078,49 @@ class TestStockValuationWithCOA(AccountTestInvoicingCommon):
             {'debit': 0.0, 'credit': 100.0, 'reconciled': True},
             {'debit': 100.0, 'credit': 0.0, 'reconciled': True},
         ])
+
+    def test_move_value_invoice_manual_rate(self):
+        """Check that if a rate is manually set on a bill, this rate
+        is used for the valuation of the move.
+        """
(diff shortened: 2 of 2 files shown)


addons/purchase_stock/models/stock_move.py
```diff
@@ -167,10 +167,11 @@ def _get_value_from_account_move(self, quantity, at_date=None):
             aml_ids.add(aml.id)
             if aml.move_type == 'in_invoice':
                 aml_quantity += aml.product_uom_id._compute_quantity(aml.quantity, self.product_id.uom_id)
-                value += aml.currency_id._convert(aml.price_subtotal, self.company_id.currency_id, date=aml.date)
+                # value += aml.currency_id._convert(aml.price_subtotal, self.company_id.currency_id, date=aml.date)
```
Sehnde:
```suggestion
```
don't forget to remove old lines

Sehnde (changes requested):
Even if the tests are disabled, could indicate which test it should trigger and adapt it if necessary, or add a test for this case if none is present, thanks.

addons/purchase_stock/tests/test_stockvaluation.py
Sehnde: Do you really need a specific date_planned ?If yes, I would suggest freezing it. A few millisecond could change the time from today to tomorrow and then your `bill.invoice_date` would be the next day.

Sehnde: Do you really need a specific date_planned ? If yes, I would suggest freezing it. A few millisecond could change the time from today to tomorrow and then your bill.invoice_date would be the next day.

Sehnde, line 2119:
```suggestion
        self.assertEqual(po.picking_ids.move_ids.value, 100)
```
To make sure the value is different before and after changing the `invoice_currency_rate`

Sehnde (commented):
Could you add a docstring to the test explaining the usecase, please ? Another nitpick, you're not adding a rate like you said in the commit message. It doesn't seem to impact the test buuuut... I like it when everything is consistent.

Sehnde (approved):

clesgow:
LGTM, thanks for the fix!
@robodoo r+

PR #242079, opened 2026-01-05
[IMP] web: enable sharing filtered view with URL

jeanschoenlaub:
Description of the issue/feature this PR addresses:

When teams collaborate, it is often useful to share a page by keeping the searches, (custom) filters and groups. Currently, when the user copies then pastes then URL somewhere, the colleague who clicks on it could not see the same things as the user who shared the link. This can be misleading.

Current behavior before PR:

The search model does not parse the url query string for search model parameters.

Desired behavior after PR is merged:

- Command to copy the current filters with the url into the clipboard (from the command pallette)
- - > ONLY for stored action, we prevent sharing dynamic actions on purpose.
- Activating individual filters / GroupBys from the url params if present.  

task#5391304

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245715, opened 2026-01-26
[FIX] web: fix undeterministic multi build url

jeanschoenlaub:
In the copy to clipboard with filter functionality,
we trigger different types of notifications depending
on the url length, resulting in underteministic behavior.
We fix this by using `browser.location.href` instead of
`window.location.href` which is mocked in tests.

ORIGIN:
In `shareViewUrl` we trigger different types of notifications
based on the copied url length. In the normal build the base
url is short but on multibuilds we include the id of targeted
modules in the test url:

`http://127.0.0.1:8069/web/tests?&headless&loglevel=2&preset=
desktop&timeout=15000&id=b21ec1ed&id=8d41322c&id=edfe3b06&....`

Where each `&id=...` represents a modules, rasing the total url
lenght over the 2000 char limit where we raise a warning.

FIX:
By using `browser.location.href` instead of `window.location.href`
the url is patched in the testing environment (see MockLocation)

So instead of having the long url displayed before, the base url
becomes `https://hoot-test.com`/, thereby dispaying the success
notification even in multibuilds where the base url is lenghty.

Breaking PR: https://github.com/odoo/odoo/pull/242079

task#391304

### DIFF
--- a/addons/web/static/src/search/search_bar_menu/search_bar_menu.js
+++ b/addons/web/static/src/search/search_bar_menu/search_bar_menu.js
@@ -5,6 +5,7 @@ import { DropdownItem } from "@web/core/dropdown/dropdown_item";
 import { registry } from "@web/core/registry";
 import { sortBy } from "@web/core/utils/arrays";
 import { useBus, useService } from "@web/core/utils/hooks";
+import { browser } from "@web/core/browser/browser";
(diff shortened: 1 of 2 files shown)


aab-odoo (approved):
robodoo r+

PR #229698, opened 2025-10-02
[IMP]: iot_box_image: improve and isolate the iot_box image builder

d-fence:
The iot_box_image module is a suite of shell script used to build iot_box raspberry pi images.

This commit remove the module in order to replace it by a python script and move the code outside of an odoo module.

The python script is intended to be used with a qemu/kvm virtual machine in order to isolate the build process.

python3-fabric debian package is required to communicate with the virtual machine through ssh.

PR #245716, opened 2026-01-26
[FIX] iot_drivers: pip/apt packages path for upgrade

loouis-t:
As we moved configuration folder in `setup/iot_box_builder`, we need to adapt the path to find them.

Related PR: https://github.com/odoo/odoo/pull/229698


Ysoroko (approved):
robodoo r+

PR #245717, opened 2026-01-26
[FIX] im_livechat: prevent local storage loops

tsm-odoo:
A bad pattern has been used for some time in discuss for fields stored in localStorage. The field updates via the `onUpdate` function in the current tab and writes to localStorage. Other tabs use the `storage` event to update their field.

This pattern can cause race conditions, leading to loops, high CPU usage, and freezes. When a tab receives a storage event, it may write back an outdated value, triggering further writes and conflicts across tabs.

Storage events should be treated as read-only. Only user actions should update the local storage.

This commit fixes the problematic fields.

Current behavior before PR:

---


alexkuhn (approved):
@robodoo delegate+

tsm-odoo (commented):
@robodoo r+

PR #224200, opened 2025-08-26
[FIX] iot_drivers: keep temporary disconnected printers

loouis-t:
Some printers (`lpd...PASSTHRU`s for example) tend to disappear when checking available printers list. They are often switching between one time `lpd...PASSTHRU` and the second time `socket...` protocols. As we get ip addresses for printers, we now check if the new protocol still correspond to the same printer, and if so, we keep the old one.

As the printer will still be in the cups queue list, it will still be able to print through it.

PR #245719, opened 2026-01-26
[FIX] iot_drivers: fix missing disconnect counter

MaxTheWhale:
This commit fixes an edge-case between the disconnect counter logic, and the logic added in odoo/odoo#224200 to prevent printers switching between `lpd` and `socket` protocols. In this case, a device is returned that only has the `identifier` key set. Because the driver has already been instantiated, this didn't break anything before, but now it also clears the `disconnect_counter` key, leading to a traceback on the next iteration of the interface when it tries to check the `disconnect_counter`.

The fix is simply to include the `disconnect_counter` as well.

---


loouis-t (approved):
robodoo r+

PR #245740, opened 2026-01-26
[FIX] hr_homeworking: send current day location field in get_views

kebeclibre:
Before this commit, the feature at commit odoo/odoo@b3be3af61cc08d0dea84969425d24957f215b26f worked by chance, because in most cases ALL fields where returned in the get views, since most of the time the search view is asked for as well, hence yielding all fields in the model

There were issues when triggering get_views from another place, namely studio when creating a many2many.

After this commit, we make sure the current day location field's description is sent

opw-5484321

Current behavior before PR:

---

### DIFF
--- a/addons/hr_homeworking/models/hr_employee.py
+++ b/addons/hr_homeworking/models/hr_employee.py
@@ -40,8 +40,9 @@ class HrEmployeeBase(models.AbstractModel):
         dayfield = self._get_current_day_location_field()
         if 'search' in res['views']:
             res['views']['search']['arch'] = res['views']['search']['arch'].replace('today_location_name', dayfield)
-        if 'list' in res['views']:
-            res['views']['list']['arch'] = res['views']['list']['arch'].replace('name_work_location_display', dayfield)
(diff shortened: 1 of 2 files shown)


addons/hr_homeworking/models/hr_employee.py
```diff
@@ -42,6 +42,7 @@ def get_views(self, views, options=None):
             res['views']['search']['arch'] = res['views']['search']['arch'].replace('today_location_name', dayfield)
         if 'list' in res['views']:
             res['views']['list']['arch'] = res['views']['list']['arch'].replace('name_work_location_display', dayfield)
+        res["models"][self._name].update(self.fields_get([dayfield]))
```
kebeclibre, line 45: FWPORT NOTE:
In 18 it will be res["models"][self._name]["fields"]

because of odoo/odoo@4251995c04056a71c60cd39dbd5aff22d308ecd1

mepe-odoo (approved):
LGTM

PR #245741, opened 2026-01-26
[FIX] planning, resource, hr_attendance_gantt: return no gantt unavailabilities for flexible calendar

mathcoutant:
### Issue:
When an employee has a flexible calendar, its gantt view in planning and other modules, like project or time off, shows incorrect grayed-out days. In general, employees with a flexible calendar should not have any grayed out days other than public holidays.

### Steps to reproduce:
- Create a new work schedule:
    - Flexible Hours
    - 24 hours per week
    - 8 average hours per day
- Create an employee with this work schedule
- Open Planning app
- Only the three first days of the week are not grayed out

### Cause:
With 24 hours per week, the worked time can be done in three days. Odoo considers that the first three days of the given time frame are worked, but the others aren't. This result is returned by `_work_intervals_batch()`. In `_unavailable_intervals_batch()` we then build the unavailabilities by taking the inverse of the work intervals. The only exception was for fully flexible resources (i.e. without a calendar).

### Solution:
The case we need to consider is for flexible resources: they are available anytime, all week long except on public holidays and leaves.

To do this we fix the method `_unavailable_intervals_batch()` which is then called by each module. If the resource is flexible, then we only return the interval of the public holidays overlapping with the time frame.

Works in `appointment`, `hr_holiday_gantt`and `project_enterprise`.

In `planning`, we need to skip the case where nothing is returned in `leaves_mapping` otherwise it will consider the `company_leaves` during weeks with no public holiday or leaves.

In `hr_attendances_gantt`, `_unavailable_intervals_batch()` is not called. So we change the way the unavailabilities were computed to call `_unavailable_intervals_batch()` instead.

opw-4879481

### DIFF
--- a/addons/resource/models/resource_calendar.py
+++ b/addons/resource/models/resource_calendar.py
@@ -559,7 +559,10 @@ class ResourceCalendar(models.Model):
         resources_work_intervals = self._work_intervals_batch(start_dt, end_dt, resources, domain, tz)
         result = {}
         for resource in resources_list:
-            if resource and resource._is_fully_flexible():
+            if resource and resource._is_flexible():
+                leaves = self._leave_intervals_batch(start_dt, end_dt, resource, domain, tz=tz)
+                if res_leaves := leaves.get(resource.id, []):
+                    result[resource.id] = [(i[0], i[1]) for i in res_leaves]
                 continue
             work_intervals = [(start, stop) for start, stop, meta in resources_work_intervals[resource.id]]
             # start + flatten(intervals) + end


addons/resource/models/resource_calendar.py
```diff
@@ -562,6 +562,17 @@ def _unavailable_intervals_batch(self, start_dt, end_dt, resources=None, domain=
...
+                public_holidays = self.env['resource.calendar.leaves'].search([
+                    ('resource_id', '=', False),
+                    ('date_from', '<', end_dt),
+                    ('date_to', '>', start_dt),
+                    ('calendar_id', 'in', [False, resource.calendar_id.id]),
+                    ('company_id', '=', resource.company_id.id)
+                ])
+                if public_holidays:
+                    result[resource.id] = [(ph.date_from.astimezone(utc), ph.date_to.astimezone(utc)) for ph in public_holidays]
+                continue
```
mepe-odoo, line 566: why only public holiday ? you can take a regular time off it will create a resource calendar leave for the employee's resource and in my opinion it should be added to the unavaibilities interval

mepe-odoo (changes requested):

addons/resource/models/resource_calendar.py
```diff
@@ -562,6 +562,17 @@ def _unavailable_intervals_batch(self, start_dt, end_dt, resources=None, domain=
...
+                leaves = self.env['resource.calendar.leaves'].search([
+                    ('resource_id', 'in', [False, resource.id]),
+                    ('date_from', '<', end_dt),
+                    ('date_to', '>', start_dt),
+                    ('calendar_id', 'in', [False, resource.calendar_id.id]),
+                    ('company_id', '=', resource.company_id.id)
+                ])
+                if leaves:
+                    result[resource.id] = [(ph.date_from.astimezone(utc), ph.date_to.astimezone(utc)) for ph in leaves]
+                continue
```
mepe-odoo, line 566: maybe you can check the changes done in master : https://github.com/odoo/odoo/blob/2dff65ab8b5a9db21d5b476065a72755cc4625be/addons/resource/models/resource_calendar.py#L455

I think you can do the same and it should works :D

mepe-odoo (changes requested):

addons/resource/models/resource_calendar.py
```diff
@@ -561,6 +562,11 @@ def _unavailable_intervals_batch(self, start_dt, end_dt, resources=None, domain=
         for resource in resources_list:
             if resource and resource._is_fully_flexible():
                 continue
+            if resource and resource._is_flexible():
+                leaves = self._leave_intervals_batch(start_dt, end_dt, resource, domain, tz=tz)
+                if res_leaves := leaves.get(resource.id, []):
+                    result[resource.id] = [(i[0], i[1]) for i in res_leaves]
+                continue
```
Bertrand2, line 566: I believe that we should do the same for fully flexible employees, their time off should be taken into account as well IMO.
```py
        for resource in resources_list:
            if resource and resource._is_flexible():
                leaves = self._leave_intervals_batch(start_dt, end_dt, resource, domain, tz=tz)
                if res_leaves := leaves.get(resource.id, []):
                    result[resource.id] = [(i[0], i[1]) for i in res_leaves]
                continue
```

Bertrand2 (changes requested):
Thanks for the fix!
The method `_unavailable_intervals_batch` is not only called for gantt views but also for planning and timesheet methods, so I'll add rd-services as a reviewer to ensure that it does not break expected behaviour on their side.
Depending on the expected behaviours, it could make sense to add extra tests to ensure that everything is correctly returned in the different used places (at least in `_gantt_unavailability` for example)
Thanks!

Bertrand2 (approved):
robodoo r+

PR #245743, opened 2026-01-26
[FIX] project_timesheet_holidays: create PH timesheets for right company

taqu-odoo:
**Steps to reproduce**
- Have 2 companies A and B
- Use a single working schedule (needs to have no company on it) for both companies and their employees
- Create a public holiday with company A, while having company B in the selected companies
- There's a timesheet for the public holiday created for employees of company B, even though the public holiday will not apply for them.

**Change**
Only generate the timesheets for employees belonging to the companies of the public holidays.

opw-5498462

### DIFF
--- a/addons/project_timesheet_holidays/models/resource_calendar_leaves.py
+++ b/addons/project_timesheet_holidays/models/resource_calendar_leaves.py
@@ -121,7 +121,7 @@ class ResourceCalendarLeaves(models.Model):
         resource_calendars = self._get_resource_calendars()
         work_hours_data = self._work_time_per_day(resource_calendars)
         employees_groups = self.env['hr.employee']._read_group(
-            [('resource_calendar_id', 'in', resource_calendars.ids), ('company_id', 'in', self.env.companies.ids)],
+            [('resource_calendar_id', 'in', resource_calendars.ids), ('company_id', 'in', self.company_id.ids if self.company_id else self.env.companies.ids)],
             ['resource_calendar_id'],
             ['id:recordset'])
         mapped_employee = {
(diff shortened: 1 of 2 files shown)


addons/project_timesheet_holidays/models/resource_calendar_leaves.py
```diff
@@ -121,7 +121,7 @@ def _timesheet_create_lines(self):
         resource_calendars = self._get_resource_calendars()
         work_hours_data = self._work_time_per_day(resource_calendars)
         employees_groups = self.env['hr.employee']._read_group(
-            [('resource_calendar_id', 'in', resource_calendars.ids), ('company_id', 'in', self.env.companies.ids)],
+            [('resource_calendar_id', 'in', resource_calendars.ids), ('company_id', 'in', self.company_id.ids)],
```
xavierbol: the company could be falsy for the `resource.calendar.leaves`, no? 🤔

xavierbol (commented):

addons/project_timesheet_holidays/models/resource_calendar_leaves.py
taqu-odoo: @xavierbol normally it always has one I think, there's a fallback on the active one:
https://github.com/odoo-dev/odoo/blob/29c503738cba8bcd2713433b7ec62f6f2295b37a/addons/resource/models/resource_calendar_leaves.py#L61
and it is not removable from the UI. But it doesn't cost anything to be safe :+1:

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #245745, opened 2026-01-26
[IMP] im_livechat,*: New tab for chatbot tests for easier navigation

KevGerard:
*: website_livechat

Currently, it is quite difficult to see the results of the test of the chatbot and then come back to the chatbot itself.

With this commit, we open this test in a new page to avoid disturbing the user experience.

The title of the page is also changed to have a better view of what chatbot is being tested.

task-5873723

---

### DIFF
--- a/addons/im_livechat/views/im_livechat_chatbot_templates.xml
+++ b/addons/im_livechat/views/im_livechat_chatbot_templates.xml
@@ -11,7 +11,7 @@
             <t t-call="web.frontend_layout"
                 no_livechat="True"
                 head="head"
-                title="chatbot_name">
+                title="title">
                 <t t-set="chatbot_name" t-value="chatbot_script.title" />
 
                 <div id="wrap">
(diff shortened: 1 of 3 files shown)


addons/im_livechat/views/im_livechat_chatbot_templates.xml
```diff
@@ -8,11 +8,11 @@
                         info="{ 'available': True, 'options': { 'chatbot_test_store': chatbot_test_store }, 'server_url': server_url }"/>
                 </script>
             </t>
+            <t t-set="chatbot_name" t-value="chatbot_script.title" />
             <t t-call="web.frontend_layout"
                 no_livechat="True"
                 head="head"
-                title="chatbot_name">
-                <t t-set="chatbot_name" t-value="chatbot_script.title" />
+                title="'Test ' + chatbot_name">
```
tsm-odoo: I don't think it plays well with translations. We should probably pass the translated title to the template when rendering it.

KevGerard: definitely not translated indeed...  code updated

tsm-odoo (approved):
@robodoo delegate+

Don't forget to mention `task-XXX` in your commit message :)


LGTM otherwise.

PR #245748, opened 2026-01-26
[IMP] web: detect correct closest elements

Syozik:
This commit improves the drag and drop functionality by detecting
the closest elements when dragging. The previous implementation
only calculated distance when the cursor was outside the
rectangle, returning 0 for any position inside. This caused
issues with nested dropdowns: when a small dropdown is inside a
larger one, the quadrance of the larger one would be 0, but since
for the smaller one it is actually harder to place it perfectly inside,
the distance would be non-zero, so the larger one would be preferred.

The new implementation:
- Handles rectangle-to-rectangle overlap detection using center
  distances
- For non overlapping rectangles, returns correct quadrance
- Returns negative normalized overlap ratio for overlapping
  elements, preferring smaller elements with similar overlap
  percentages
- Supports both point queries (x, y) and rectangle queries
  (x, y, width, height)

Steps to see the issue:
- Open Website and start editing
- Drop the `s_banner` snippet
- Try to drop a button (or any other inner element) next to the
existing button in the snippet
=> The experience is not the most user-friendly

Old implementation:

https://github.com/user-attachments/assets/507c0402-b855-46a7-938d-8f05d905dd3e

New implementation:

https://github.com/user-attachments/assets/34b352b4-e08a-4a7f-a2b5-2702ee8e63ef

It might not look like an improvement in the video, but it feels actually easier now when trying to drag and drop to a relatively small dropzone.

task-5794898

### DIFF
--- a/addons/web/static/src/core/utils/ui.js
+++ b/addons/web/static/src/core/utils/ui.js
@@ -6,7 +6,7 @@
 
 /**
  * @param {Iterable<HTMLElement>} elements
- * @param {Position} targetPos
+ * @param {Position | DOMRect} targetPos
  * @returns {HTMLElement | null}
  */
 export function closest(elements, targetPos) {
@@ -57,22 +57,63 @@ export function isVisible(el) {
 
 /**
  * @param {DOMRect} rect
- * @param {Position} pos
+ * @param {Position | DOMRect} pos
  * @returns {number}
  */
 export function getQuadrance(rect, pos) {
-    let q = 0;
-    if (pos.x < rect.x) {
-        q += (rect.x - pos.x) ** 2;
-    } else if (rect.x + rect.width < pos.x) {
-        q += (pos.x - (rect.x + rect.width)) ** 2;
+    const isPoint = !pos.width || !pos.height;
+
+    if (isPoint) {
+        // For a point, use standard point-to-rectangle distance
+        const dx = Math.max(rect.x - pos.x, 0, pos.x - (rect.x + rect.width));
+        const dy = Math.max(rect.y - pos.y, 0, pos.y - (rect.y + rect.height));
+        return dx ** 2 + dy ** 2;
     }
-    if (pos.y < rect.y) {
-        q += (rect.y - pos.y) ** 2;
-    } else if (rect.y + rect.height < pos.y) {
-        q += (pos.y - (rect.y + rect.height)) ** 2;
+
+    const rectCenter = {
+        x: rect.x + rect.width / 2,
+        y: rect.y + rect.height / 2,
+    };
+    const posCenter = {
+        x: pos.x + pos.width / 2,
+        y: pos.y + pos.height / 2,
+    };
+
+    // Compute the distance between the centers
+    const dx = Math.abs(rectCenter.x - posCenter.x);
+    const dy = Math.abs(rectCenter.y - posCenter.y);
+    // Gap is negative if rects are overlapping
+    const gapX = dx - (rect.width + pos.width) / 2;
+    const gapY = dy - (rect.height + pos.height) / 2;
+
+    // Rectangles don't overlap (at least one gap is positive)
+    if (gapX >= 0 || gapY >= 0) {
+        // Return squared Euclidean distance to nearest edge
+        return Math.max(0, gapX) ** 2 + Math.max(0, gapY) ** 2;
     }
-    return q;
+
+    // Rectangles overlap - calculate the overlap region
+    const overlapRect = {
+        left: Math.max(rect.x, pos.x),
+        right: Math.min(rect.x + rect.width, pos.x + pos.width),
+        top: Math.max(rect.y, pos.y),
+        bottom: Math.min(rect.y + rect.height, pos.y + pos.height),
+    };
+    const overlapArea =
+        (overlapRect.right - overlapRect.left) * (overlapRect.bottom - overlapRect.top);
+    const rectArea = rect.width * rect.height;
+    // Normalize the overlap area by rect size to prefer smaller elements when
+    // overlap area is similar (e.g., nested dropzones). Return negative
+    // value so larger overlap ratios are "closer".
+    // Example: two rects both have overlapArea = 15
+    //   Rect A: area = 100 → ratio = 15/100 = 0.15
+    //   Rect B: area = 30  → ratio = 15/30  = 0.50
+    //
+    // A higher ratio means we cover more of that element, so Rect B is the
+    // better target. But standard min-distance logic favors smaller values so
+    // we negate the ratio: -0.15 > -0.50, making Rect B "closer" and correctly
+    // selected as the nearest match.
+    return -(overlapArea / rectArea);
 }
 
 /**
--- a/addons/web/static/tests/core/utils/closest.test.js
+++ b/addons/web/static/tests/core/utils/closest.test.js
@@ -1,0 +1,38 @@
+import { expect, test } from "@odoo/hoot";
+import { closest } from "@web/core/utils/ui";
+
+test("Closest function works correctly with nested elements", () => {
+    const inner = {
+        getBoundingClientRect: () => ({ x: 50, y: 50, width: 50, height: 50 }),
+    };
+    const outer = {
+        getBoundingClientRect: () => ({ x: 0, y: 0, width: 200, height: 200 }),
+    };
+    const pos = { x: 60, y: 60 };
+    const result = closest([inner, outer], pos);
+    expect(result).toBe(inner);
+});
+
+test("Closest function finds the correct element", () => {
+    const inner = {
+        getBoundingClientRect: () => ({ x: 50, y: 50, width: 50, height: 50 }),
+    };
+    const outer = {
+        getBoundingClientRect: () => ({ x: 0, y: 0, width: 200, height: 200 }),
+    };
+    const pos = { x: 10, y: 10, width: 40, height: 40 };
+    const result = closest([inner, outer], pos);
+    expect(result).toBe(outer);
+});
+
+test("Closest function returns the right element when position is right next to it", () => {
+    const el1 = {
+        getBoundingClientRect: () => ({ x: 50, y: 50, width: 100, height: 100 }),
+    };
+    const el2 = {
+        getBoundingClientRect: () => ({ x: 0, y: 0, width: 300, height: 200 }),
+    };
+    const pos = { x: 49, y: 49, width: 20, height: 20 };
+    const result = closest([el1, el2], pos);
+    expect(result).toBe(el1);
+});


addons/web/static/tests/core/closest.test.js
```diff
@@ -0,0 +1,30 @@
...
+        getBoundingClientRect: () => ({ x: 50, y: 50, width: 50, height: 50 }),
+    };
+    const outer = {
+        getBoundingClientRect: () => ({ x: 0, y: 0, width: 200, height: 200 }),
+    };
+    const pos = { x: 60, y: 60 };
+
+    const result = closest([inner, outer], pos);
+
+    expect(result).toBe(inner);
```
sobo-odoo, line 13: You can remove the empty lines (same below), it takes space for nothing since the test is short :wink: 

In fact it is a general comment, try to not use too many empty lines if not really necessary :slightly_smiling_face:

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
+
+    // Calculate gap (negative if overlapping, positive if separated)
+    const gap_x = dx - combined_half_width;
+    const gap_y = dy - combined_half_height;
+
+    // If rectangles don't overlap (at least one gap is positive)
+    if (gap_x >= 0 || gap_y >= 0) {
+        // Return squared Euclidean distance to nearest edge
+        return Math.max(0, gap_x) ** 2 + Math.max(0, gap_y) ** 2;
     }
```
sobo-odoo: In JS we use camel case for variables, so rename them to `combinedHalfWidth|Height` and `gapX|Y` :slightly_smiling_face:

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
+        x: rect.x + rect.width / 2,
+        y: rect.y + rect.height / 2,
+    };
+    const pos_center = {
+        x: pos.x + pos.width / 2,
+        y: pos.y + pos.height / 2,
+    };
+
+    const combined_half_width = (rect.width + pos.width) / 2;
+    const combined_half_height = (rect.height + pos.height) / 2;
```
sobo-odoo: Maybe declare them just above where you use them, so near the gap variables :slightly_smiling_face: 

Also, maybe don't create intermediate variables and just directly do that in the gap computation, with a better comment like "Compute the gap between the two rects (...)" so it will be straighforward to understand :slightly_smiling_face: 

So 
```js
// Calculate the gap between the two rects (negative if overlapping,
// positive if separated)
const gapX = dx - (rect.width + pos.width) / 2;
const gapY = dy - (rect.height + pos.height) / 2;
```

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
+    // Rectangles overlap - calculate the overlap region
+    const leftPos = Math.max(rect.x, pos.x);
+    const rightPos = Math.min(rect.x + rect.width, pos.x + pos.width);
+    const topPos = Math.max(rect.y, pos.y);
+    const bottomPos = Math.min(rect.y + rect.height, pos.y + pos.height);
+
+    const width = rightPos - leftPos;
+    const height = bottomPos - topPos;
+    const overlapArea = width * height;
+    const rectArea = rect.width * rect.height;
```
sobo-odoo: Suggestion:
```js
const overlapRect = {
  left: Math.max(rect.x, pos.x),
  right: Math.min(rect.x + rect.width, pos.x + pos.width),
  top: Math.max(rect.y, pos.y),
  bottom: Math.min(rect.y + rect.height, pos.y + pos.height),
}
const overlapArea = (overlapRect.right - overlapRect.left) * (overlapRect.bottom - overlapRect.top);
const rectArea = rect.width * rect.height;
```

It prevents creating too many variables, and the fact that it is named `overlapRect` makes it easier to see what positions we are computing :slightly_smiling_face: (you can of course choose another name)
An alternative is to name your variables `overlapRectLeft|Right|...` :+1:

addons/web/static/tests/core/utils/closest.test.js
```diff
@@ -0,0 +1,30 @@
...
+        getBoundingClientRect: () => ({ x: 0, y: 0, width: 200, height: 200 }),
+    };
+    const pos = { x: 60, y: 60 };
+
+    const result = closest([inner, outer], pos);
+
+    expect(result).toBe(inner);
+});
+
+test("Closest function returns the right element when position is right next to it", () => {
```
sobo-odoo, line 28: Maybe add a third test for when we have two overlapping rect, with the pos not touching the inner one, which I think should return the outer one this time :slightly_smiling_face: 
Because the two tests are quite similar: pos fully inside inner, and pos outside but overlapping with inner, so it makes sense to have pos right next to inner but not touching it :slightly_smiling_face: 

Just to make the tests more robust/relevant :wink:

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
+
+    // Normalize the overlap area by rect size to prefer smaller elements
+    // when overlap percentage is similar (e.g., nested dropdowns)
+    // Return negative value so larger overlap ratios are "closer"
+    // (more negative = better, works with standard min-distance logic)
```
sobo-odoo: - The comments char limit is 80, so you can make your lines a bit longer here :slightly_smiling_face: (72 is only for the commit message :wink: )
- `nested dropdowns` -> dropzones :wink: 

- > // (more negative = better, works with standard min-distance logic)

  I don't really understand this part, maybe make it a bit more clear ?

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
-    if (pos.x < rect.x) {
-        q += (rect.x - pos.x) ** 2;
-    } else if (rect.x + rect.width < pos.x) {
-        q += (pos.x - (rect.x + rect.width)) ** 2;
+    const isPoint = !pos.width || !pos.height;
+
+    if (isPoint) {
+        // For a point, use standard point-to-rectangle distance
+        const dx = Math.max(rect.x - pos.x, 0, pos.x - (rect.x + rect.width));
+        const dy = Math.max(rect.y - pos.y, 0, pos.y - (rect.y + rect.height));
```
sobo-odoo, line 69: From a performance point of view, it would be more efficient to not use `Math.max` but use either if/else or a ternary operator (to check because I am not sure, from what I saw, maybe ternary is worse)
Because we can have a lot of dropzones on a page so doing `Math.max` for each one of them will slow down things :slightly_smiling_face: 

Well of course we run `closest` only on the ones returned by `touching` so there are not that many dropzones in reality, but if we can speed up things, it is better :slightly_smiling_face: 
I had the feeling that things were a bit too slow when testing, even if the experience is still way better than before, maybe that's why :thinking: 

(This comment applies to all the `Math.max` you added)

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
...
+    };
+    const pos_center = {
+        x: pos.x + pos.width / 2,
+        y: pos.y + pos.height / 2,
+    };
+
+    const combined_half_width = (rect.width + pos.width) / 2;
+    const combined_half_height = (rect.height + pos.height) / 2;
+
+    const dx = Math.abs(rect_center.x - pos_center.x);
```
sobo-odoo: I would add a comment saying "Compute the distance between the rects center"

addons/web/static/src/core/utils/ui.js
```diff
@@ -61,18 +61,57 @@ export function isVisible(el) {
  * @returns {number}
```
sobo-odoo, line 61: Commit message:
- Suggestion: maybe rename the title to `improve closest elements detection`
- `with nested dropdowns` -> dropzones :wink: (there is the mistake right after it too)

sobo-odoo (changes requested):
Nice work, it really makes the drag and drop experience way better ! :slightly_smiling_face:  I tested with the commit of BVR making the dropzones thinner and it works well too :wink:  

I have some comments and suggestions to improve the code and the tests and maybe to also improve the perfs :+1:

addons/web/static/src/core/utils/ui.js
Syozik: Yeah... Sorry, stupid mistake, I guess I woke up that day feeling python-ish :sweat_smile:

addons/web/static/src/core/utils/ui.js
Syozik: Hmm, I just ran a small test, and I don't see a big difference in perf, you can try it yourself:
```
const rect = { x: 10, y: 10, width: 100, height: 100 };
const pos = { x: 5, y: 5 };

function benchMax(iterations) {
    const start = performance.now();
    let dx;
    for (let i = 0; i < iterations; i++) {
        dx = Math.max(rect.x - pos.x, 0, pos.x - (rect.x + rect.width));
    }
    const end = performance.now();
    return end - start;
}

function benchIf(iterations) {
    const start = performance.now();
    let dx;
    for (let i = 0; i < iterations; i++) {
        if (pos.x < rect.x) {
            dx = rect.x - pos.x;
        } else if (pos.x > rect.x + rect.width) {
            dx = pos.x - (rect.x + rect.width);
        } else {
            dx = 0;
        }
    }
    const end = performance.now();
    return end - start;
}

let iterations = 1;
for (let i = 0; i <= 5; i++) {
    iterations *= 10
    let avgIfTime = 0;
    let avgMaxTime = 0;

    for (let i = 0; i < 10; i++) {
        avgIfTime += benchIf(iterations);
    }

    for (let i = 0; i < 10; i++) {
        avgMaxTime += benchMax(iterations);
    }

    console.log(`For ${iterations} iterations:`);
    console.log(`Avg if time (ms): ${avgIfTime / 10}`);
    console.log(`Avg max time (ms): ${avgMaxTime / 10}`);
}
```
Sometimes `Math.max` is faster
image
Sometimes not so much
image
However, `Math.max` is cleaner in this case than if we used lots of it/else, so I think we should stick to it.

sobo-odoo: Okay then, thanks for the benchmark :slightly_smiling_face:

sobo-odoo (approved):
It's ok with me :slightly_smiling_face: 
You just have to retarget the branch on 19.2 and we will be good :slightly_smiling_face: 

@jum-odoo It's okay for website, can you take a look ? :slightly_smiling_face:

addons/web/static/src/core/utils/ui.js
jum-odoo: Since you're changing the API of both `closest` & `getQuadrance`, you must also adapt docstring:
```suggestion
 * @param {Position | DOMRect} targetPos
```

jum-odoo: Here as well
```suggestion
 * @param {Position | DOMRect} pos
```

jum-odoo (commented):
Nice work 👍
Just a bit of nitpicking on docstrings and we're good to go

Syozik:
Thanks for the reviews @sobo-odoo , @jum-odoo, I pushed the requested changes

sobo-odoo (approved):
@robodoo r+

sobo-odoo:
@robodoo retry

PR #245753, opened 2026-01-26
[FIX] website_sale: make test demo-data-independent again

reth-odoo:
As part of the forward part for [1] the shop mail tour was updated because the new demo partner was earlier alphabetically than the partner generated for the test, and tests run with demo data for CI in this version.

The test should create a partner that will appear first in the m2o selection regardless of whether demo data is installed.

[1]: 549708965924b17403ef7c8e6a6d5bc43af460c3

runbot-238406

### DIFF
--- a/addons/website_sale/static/tests/tours/website_sale_shop_mail.js
+++ b/addons/website_sale/static/tests/tours/website_sale_shop_mail.js
@@ -29,17 +29,17 @@ registry.category("web_tour.tours").add('shop_mail', {
     {
         content: "Open recipients dropdown",
         trigger: '.o_field_many2many_tags_email[name=partner_ids] input',
-        run: 'click',
+        run: 'text Azure Interior (Test)',
     },
     {
         content: "Select azure interior",
-        trigger: '.ui-menu-item a:contains(olson28)',
+        trigger: '.ui-menu-item a:contains(Interior24)',
         in_modal: false,
     },
     {
         content: "click Send email",
         trigger: '.btn[name="action_send_mail"]',
-        extra_trigger: '.o_badge_text:contains("Acme")',
+        extra_trigger: '.o_badge_text:contains("Azure")',
     },
     {
         content: "wait mail to be sent, and go see it",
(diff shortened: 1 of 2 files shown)


reth-odoo (commented):
🤖 fumbled a test adaptation during fw
(breaks "no demo" nightly builds)

@awa-odoo

addons/website_sale/static/tests/tours/website_sale_shop_mail.js
```diff
@@ -33,7 +33,7 @@ registry.category("web_tour.tours").add('shop_mail', {
     },
```
ltinel, line 33: (Commenting on this line since the one above (line 32) is too far away from your changes and can't be commented on :sweat_smile: )

You don't need to rely on the ordering of the partners (which is a bit brittle). Instead, you can search for the desired partner in the dropdown:
```diff
    {
        content: "Open recipients dropdown",
        trigger: '.o_field_many2many_tags_email[name=partner_ids] input',
-       run: 'click',
+       run: 'edit olson28'
    },
```

ltinel (changes requested):
Hello, thanks for the fix :) I suggested a slightly more robust way to fix it instead (to avoid regressions)

addons/website_sale/static/tests/tours/website_sale_shop_mail.js
reth-odoo: @ltinel 

Indeed seems like a better approach ^^

ltinel: I'm guessing `text` is the version-17-equivalent of `edit`? Don't forget to update it in your FW ports (18 and up I think) 🙂

reth-odoo: @ltinel Yes, anyways it will fail when edit is introduced :)

Would you mind merging since you reviewed it? :eyes:

ltinel (approved):
@robodoo r+

PR #238396, opened 2025-12-02
[IMP] website_sale: archive guest contacts

csan-odoo:
Archive contacts associated with guest orders.
Unarchive them if they sign up and become users.

task-5345512

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245759, opened 2026-01-26
[FIX] website_sale: correctly encode URL

mwath:
Oversight of https://github.com/odoo/odoo/pull/238396
Calling `urlencode` with the result of `parse_qs` is incorrect, and requires `doseq=True` to work as expected.

opw-none

---


ltinel (approved):
@mwath I do agree it's safer to use `doseq=True`, but IIUC, it only has any effect if there are multiple query params with the same key. So I'm curious: are there any such query params and were you able to make the code fail?

Either way, your fix LGTM :)
@robodoo delegate+

PR #245761, opened 2026-01-26
[IMP] mail: better icon and title for native notification

Atovange:
This commit rehauls the icon and title of native notifications, giving them more useful context on the message received. 

For chatter messages:
- icon: module icon
- title: record name

For Discuss messages:
- icon: channel/group avatar or user avatar for DMs
- title: channel/group name or user name for DMs

For call notifications:
- icon: caller avatar or group avatar for group DMs
- title: channel name or caller name for DMs

This commit also removes the modal for microphone permissions when joining a call via the "Accept" action button of the call push notification.

task-5879912

### DIFF
--- a/addons/mail/models/discuss/discuss_channel.py
+++ b/addons/mail/models/discuss/discuss_channel.py
@@ -1311,22 +1311,31 @@ class DiscussChannel(models.Model):
             store.add(message.channel_id, ["message_count"])
         return rdata
 
+    def _notify_by_push_get_title(self, message, force_record_name=False):
+        author = message.author_id or message.author_guest_id
+        if self.channel_type == "chat":
+            # in a DM the conversation is the author
+            return author.name
+        record_name = force_record_name or message.record_name
+        if self.channel_type == "channel":
+            return f"#{record_name}"
+        return record_name or self.display_name
+
+    def _notify_by_push_get_body_prefix(self, message):
+        # in a DM the title is already the author's name
+        if self.channel_type == "chat":
+            return ""
+        return super()._notify_by_push_get_body_prefix(message)
+
+    def _notify_by_push_get_icon(self, message):
+        if self.channel_type in ["channel", "group"]:
+            return f"/web/image/discuss.channel/{self.id}/avatar_128"
+        author = message.author_id or message.author_guest_id
+        return f"/web/image/{author._name}/{author.id}/avatar_128"
+
     def _notify_by_web_push_prepare_payload(self, message, force_record_name=False):
         payload = super()._notify_by_web_push_prepare_payload(message, force_record_name=force_record_name)
         payload['options']['data']['action'] = 'mail.action_discuss'
-        record_name = force_record_name or message.record_name
-        author = message.author_id or message.author_guest_id
-        if self.channel_type == 'chat':
-            payload['title'] = author.name
-        elif self.channel_type == 'channel':
-            payload['title'] = "#%s - %s" % (record_name, author.name)
-        elif self.channel_type == 'group':
-            if not record_name:
-                member_names = self.channel_member_ids.mapped(lambda m: m.partner_id.name if m.partner_id else m.guest_id.name)
-                record_name = f"{', '.join(member_names[:-1])} and {member_names[-1]}" if len(member_names) > 1 else member_names[0] if member_names else ""
-            payload['title'] = "%s - %s" % (record_name, author.name)
-        else:
-            payload['title'] = "#%s" % (record_name)
         return payload
 
     def _notify_thread_by_web_push(self, message, recipients_data, **kwargs):
--- a/addons/mail/models/discuss/discuss_channel_member.py
+++ b/addons/mail/models/discuss/discuss_channel_member.py
@@ -637,20 +637,25 @@ class DiscussChannelMember(models.Model):
             )
             devices, private_key, public_key = self.channel_id._web_push_get_partners_parameters(members.partner_id.ids)
             if devices:
-                if self.channel_id.channel_type != 'chat':
+                user, guest = self.env["res.users"]._get_current_persona()
+                caller = guest or user.partner_id
+                icon = f"/web/image/{caller._name}/{caller.id}/avatar_128"
+                if self.channel_id.channel_type == "group":
                     icon = f"/web/image/discuss.channel/{self.channel_id.id}/avatar_128"
-                elif guest := self.env["mail.guest"]._get_guest_from_context():
-                    icon = f"/web/image/mail.guest/{guest.id}/avatar_128"
-                elif partner := self.env.user.partner_id:
-                    icon = f"/web/image/res.partner/{partner.id}/avatar_128"
                 languages = [partner.lang for partner in devices.partner_id]
                 payload_by_lang = {}
                 for lang in languages:
+                    if self.channel_id.channel_type == "chat":
+                        title = caller.name
+                    elif self.channel_id.channel_type == "channel":
+                        title = f"#{self.channel_id.display_name}"
+                    else:
+                        title = self.channel_id.with_context(lang=lang).display_name
                     env_lang = self.with_context(lang=lang).env
                     payload_by_lang[lang] = {
-                        "title": env_lang._("Incoming call"),
+                        "title": title,
                         "options": {
-                            "body": env_lang._("Conference: %s", self.channel_id.display_name),
+                            "body": env_lang._("Incoming call"),
                             "icon": icon,
                             "vibrate": [100, 50, 100],
                             "requireInteraction": True,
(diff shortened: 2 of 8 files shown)


zel-odoo:
runbot is not greened yet

addons/mail/models/mail_thread.py
```diff
@@ -4154,9 +4154,8 @@ def _notify_by_web_push_prepare_payload(self, message, force_record_name=False):
         if message.message_type == 'tracking':
             body = "\n%s\n%s" % (message.subtype_id.description, body)
 
-        if message.author_id:
-            title = "%s: %s" % (message.author_id.name, title)
-            icon = "/web/image/res.partner/%d/avatar_128" % message.author_id.id
+        if message.model:
+            icon = modules.module.get_module_icon(self.env[message.model]._original_module)
```
zel-odoo: `self` *is* the thread here, so `self.env[message.model]` is a detour.

```suggestion
        if module := self._original_module:
            icon = modules.module.get_module_icon(module)
```

addons/mail/models/mail_thread.py
```diff
@@ -4183,10 +4182,11 @@ def get_attachment_label(attachment):
                     count=total_attachments - 1,
                 )
 
+        author = message.author_id or message.author_guest_id
         return {
             'title': title,
             'options': {
-                'body': html2plaintext(body, include_references=False),
+                'body': f"{author.name}: {html2plaintext(body, include_references=False)}",
```
zel-odoo: `author` can be an empty recordset, and then `author.name` is `False`, so the body reads `"False: ..."`. Incoming mail from an unknown sender is exactly that case (`author_id = False`, only `email_from` set) — the old `if message.author_id:` branch covered it.

Suggest extracting the prefix, so `discuss_channel` can drop it for DMs instead of recomputing the body and losing the attachment fallback (see the comment there):

```suggestion
        author = message.author_id or message.author_guest_id
        body = html2plaintext(body, include_references=False)
        if prefix := self._notify_by_web_push_get_body_prefix(message):
            body = f"{prefix}{body}"
        return {
            'title': title,
            'options': {
                'body': body,
```

with:

    def _notify_by_web_push_get_body_prefix(self, message):
        """ Author prefix for the notification body, when the title does not already name them. """
        author = message.author_id or message.author_guest_id
        if name := author.name or message.email_from:
            return f"{name}: "
        return ""

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -1318,15 +1318,22 @@ def _notify_by_web_push_prepare_payload(self, message, force_record_name=False):
         author = message.author_id or message.author_guest_id
         if self.channel_type == 'chat':
             payload['title'] = author.name
+            payload["options"]["body"] = html2plaintext(message.body, include_references=False)
```
zel-odoo: Recomputing the body from `message.body` drops `super()`'s attachment fallback (`is_html_empty(body) and message.attachment_ids` → `"Voice Message"` / `"a.png and 2 other attachments"`). A voice message in a DM now pushes an **empty** body — and DMs are where voice messages happen.

```suggestion
        if self.channel_type == 'chat':
            payload['title'] = author.name
```

and override the prefix hook instead:

    def _notify_by_web_push_get_body_prefix(self, message):
        # in a DM the title is already the author's name
        if self.channel_type == "chat":
            return ""
        return super()._notify_by_web_push_get_body_prefix(message)

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -1318,15 +1318,22 @@ def _notify_by_web_push_prepare_payload(self, message, force_record_name=False):
...
+            payload["title"] = f"#{record_name}"
         elif self.channel_type == 'group':
             if not record_name:
                 member_names = self.channel_member_ids.mapped(lambda m: m.partner_id.name if m.partner_id else m.guest_id.name)
                 record_name = f"{', '.join(member_names[:-1])} and {member_names[-1]}" if len(member_names) > 1 else member_names[0] if member_names else ""
-            payload['title'] = "%s - %s" % (record_name, author.name)
+            payload["title"] = record_name
         else:
-            payload['title'] = "#%s" % (record_name)
+            payload["title"] = record_name
```
zel-odoo: Now that the author suffix is gone, `group` and `else` produce the same title, and the inline member-name join is a lossy copy of `display_name`: it misses the `member_count > 3` → `"1 other"` / `"%s others"` case and doesn't go through `format_list`.

```suggestion
        else:
            payload["title"] = record_name or self.display_name
```

addons/mail/models/discuss/discuss_channel_member.py
```diff
@@ -637,20 +637,22 @@ def _rtc_invite_members(self, member_ids=None):
...
-                    icon = f"/web/image/res.partner/{partner.id}/avatar_128"
+                elif self.channel_id.channel_type == "chat":
+                    title = caller.name
                 languages = [partner.lang for partner in devices.partner_id]
                 payload_by_lang = {}
                 for lang in languages:
                     env_lang = self.with_context(lang=lang).env
                     payload_by_lang[lang] = {
-                        "title": env_lang._("Incoming call"),
+                        "title": title,
```
zel-odoo, line 656: `title` is computed outside the per-lang loop, but `display_name` is translated for nameless groups (`format_list` + `_("1 other")` / `_("%s others")` in `_compute_display_name`), so every recipient gets it in the *caller's* language. Localising per recipient is the whole point of the loop.

```suggestion
                user, guest = self.env["res.users"]._get_current_persona()
                caller = guest or user.partner_id
                icon = f"/web/image/{caller._name}/{caller.id}/avatar_128"
                if self.channel_id.channel_type == "group":
                    icon = f"/web/image/discuss.channel/{self.channel_id.id}/avatar_128"
                languages = [partner.lang for partner in devices.partner_id]
                payload_by_lang = {}
                for lang in languages:
                    env_lang = self.with_context(lang=lang).env
                    title = (
                        caller.name
                        if self.channel_id.channel_type == "chat"
                        else self.channel_id.with_context(lang=lang).display_name
                    )
                    payload_by_lang[lang] = {
                        "title": title,
```

Also: a call in a `channel` gets a bare `display_name` here, while a message in that same channel gets `#name` from `_notify_by_web_push_prepare_payload`. Worth aligning.

addons/mail/static/src/core/common/out_of_focus_service.js
```diff
@@ -43,21 +43,25 @@ export class OutOfFocusService {
...
-                notificationTitle = _t("%(author name)s from %(channel name)s", {
-                    "author name": message.authorName,
-                    "channel name": message.channel_id.displayName,
-                });
-            } else {
-                notificationTitle = message.authorName;
-            }
+        } else if (message?.thread) {
+            icon = message.thread.module_icon;
         }
```
zel-odoo: Two problems here:

- `channel.name` is falsy for a nameless group (the member-name fallback lives in `computedDisplayName`), so **group DMs get an empty title** — the server does have that fallback.
- `message.thread.module_icon` overwrites the default with `undefined` when the store didn't send it (`predicate=lambda t: inbox_fields and t._original_module`).

`Thread.displayName` and `Thread.avatarUrl` already are this logic — for a `chat`, `channel.avatarUrl` returns the correspondent's avatar, i.e. the author of an incoming message:

```suggestion
        let title = message.thread?.displayName || _t("New message");
        if (message.thread?.channel?.channel_type === "channel") {
            title = `#${title}`;
        }
        const icon = message.thread?.avatarUrl ?? "/web/static/img/odoo-icon-192x192.png";
```

(`message` is dereferenced unconditionally above, so `message?.` was noise.)

zel-odoo:
And comments overall:

**OCN / native app isn’t covered.** This only affects web push (browser + PWA). Native Android/iOS uses `_notify_by_ocn_prepare_payload` in `enterprise/mail_mobile`, which is separate from `_notify_by_web_push_prepare_payload`. So native notifications keep the old format: `#channel - author` / author title, unprefixed body, no icon, and no `author_guest_id` fallback. Either add a companion enterprise PR or mention clearly that this is web-push only.

Better to take care of OCN here as well. Otherwise, users get confused why it is not unified.

**Android Accept button.** `delete notification.options.actions.accept` doesn’t work because `actions` is an array, so Android still shows Accept today. The `{audio: false}` change fixes that path. Either filter the action out properly or remove this block/comment now that Accept no longer needs a permission prompt.

addons/test_mail/tests/test_mail_push.py
```diff
@@ -563,14 +565,26 @@ def test_push_notification_regenerate_vapid_keys(self):
                 vapid_public_key=self.vapid_public_key,
             )
 
+    def _force_raw_body_payload(self, message, **kwargs):
```
tde-banana-odoo: Wait, what ? No

PR #245770, opened 2026-01-26
[FIX] mail: prevent local storage fields feedback loop

tsm-odoo:
Diccuss fields have a `localStorage` option. The field updates via the `onUpdate` function in the current tab and writes to the local storage. Other tabs use the `storage` event to update their field.

This pattern can cause race conditions, leading to loops, high CPU usage, and freezes. When a tab receives a storage event, it may write back an outdated value, triggering further writes and conflicts across tabs.

This commit ensures we don't trigger `storage` events recursively: upon the reception of a `storage` event, field is updated but local storage is untouched.

Current behavior before PR:

---

### DIFF
--- a/addons/mail/static/src/model/store_internal.js
+++ b/addons/mail/static/src/model/store_internal.js
@@ -44,23 +44,33 @@ export class StoreInternal extends RecordInternal {
         this.onStorage = this.onStorage.bind(this);
         browser.addEventListener("storage", this.onStorage);
     }
-
+    /**
+     * Indicates whether the current update cycle was triggered by a
+     * `storage` event. Used to prevent writing back to the local
+     * storage and creating a feedback loop.
+     */
+    isUpdatingFromStorageEvent = false;
     onStorage(ev) {
         const entryMap = this.localStorageKeyToRecordFields.get(ev.key);
         if (!entryMap) {
             return;
         }
-        for (const [record, fieldName] of entryMap.entries()) {
-            if (ev.newValue === null) {
-                record._proxy[fieldName] = record._.fieldsDefault.get(fieldName);
-            } else {
-                const parsed = parseRawValue(ev.newValue);
-                if (!parsed) {
+        this.isUpdatingFromStorageEvent = true;
+        try {
+            for (const [record, fieldName] of entryMap.entries()) {
+                if (ev.newValue === null) {
                     record._proxy[fieldName] = record._.fieldsDefault.get(fieldName);
                 } else {
-                    record._proxy[fieldName] = parsed.value;
+                    const parsed = parseRawValue(ev.newValue);
+                    if (!parsed) {
+                        record._proxy[fieldName] = record._.fieldsDefault.get(fieldName);
+                    } else {
+                        record._proxy[fieldName] = parsed.value;
+                    }
                 }
             }
+        } finally {
+            this.isUpdatingFromStorageEvent = false;
         }
     }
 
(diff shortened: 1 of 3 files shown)


addons/mail/static/src/model/store_internal.js
```diff
@@ -44,23 +44,36 @@ export class StoreInternal extends RecordInternal {
...
-
+    /**
+     * The local storage key currently being updated from a storage
+     * event. Used to prevent writing back to the local storage and
+     * creating a feedback loop. There is only one key being updated
+     * at a time since the update cycle is synchronous.
+     *
+     * @type {string|null}
+     */
+    _updatingFromStorageKey = null;
```
alexkuhn: No need to prefix `_`: this is already an object for internal stuffs, access through `store._`. Double `_` is not great.

addons/mail/static/tests/core/record.test.js
```diff
@@ -1498,6 +1499,38 @@ test("Fields with { localStorage: true } are restored from local storage", async
...
+    await expect.waitForSteps(["setItem 1"]);
+    const store = await start();
+    const message = store.Message.insert(1);
+    expect(message.body).toBe("1");
+    message.body = "2";
+    expect(message.body).toBe("2");
+    await expect.waitForSteps(["setItem 2"]);
+    browser.dispatchEvent(
+        new StorageEvent("storage", { key: bodyLocalId, newValue: JSON.stringify({ value: "3" }) })
+    );
```
alexkuhn, line 1528:
```suggestion
    browser.dispatchEvent(
        new StorageEvent("storage", { key: bodyLocalId, newValue: toRawValue("3") })
    );
```

As the shape of value could change (like versioning) and we won't need to change tests by hand if the helper handles versioning or other kinds of improvements soon.

addons/mail/static/src/model/store_internal.js
```diff
@@ -44,23 +44,36 @@ export class StoreInternal extends RecordInternal {
...
+                    const parsed = parseRawValue(ev.newValue);
+                    if (!parsed) {
+                        record._proxy[fieldName] = record._.fieldsDefault.get(fieldName);
+                    } else {
+                        record._proxy[fieldName] = parsed.value;
+                    }
                 }
             }
+        } finally {
+            this._updatingFromStorageKey = null;
```
alexkuhn: This implementation relies on the fact that there's no nested update cycle with `onStorage`.

This is not in `MAKE_UPDATE()` right now but this may happen in future... Or there's subtle case where this might happen, and this implementation would no longer work as the `onUpdate` are invoked after this code.

One implementation that would work in both cases, is to have `_updatingFromStorageKey` push keys and have it flushed at end of update cycle. Without a nested update cycle like now, this is behaviorally equivalent, but in a nested Update cycle this would work while this implementation in current diff doesn't.

Here's what I mean as an example, roughly:

- `MAKE_CYCLE`: `1`
  - `MAKE_UPDATE`: `2`
   - `onStorage`: `myStorageField = newValue`
   - `updatingFromStorageKey = null`
  - `MAKE_UPDATE_END`: `1`
- `MAKE_UPDATE_END`: `0`
- consume queues: compute, onUpdate, etc.

As you can see, onUpdate happens after `updatingFromStorageKey = null`, which is a problem because this would reintroduce the loop.

Thankfully this happens only once, as there's no `onStorage` in the meantime (flush of update cycle is synchronous, therefore `onStorage` cannot interleave while there's update cycle), so this is not super critical. We just potentially waste it once, like a reverb that converges to silence fast.

alexkuhn (approved):
LGTM.

I would prefer `_updatingFromStorageKey` to be `Set()` of storage keys and flush in QUEUE at end of update cycle for ensuing this works flawlessly regardless of nested update cycle.

@robodoo delegate+

addons/mail/static/src/model/store_internal.js
tsm-odoo: Yeah I thought anbout it as well but I decided to stick with the easy solution as we have the guarantee that update cycles are fully synchronous.

Nested update cycles are not an issue as long as they remain synchronous. In your example, the storage event occurs in the middle of an update cycle, but that can only happen if we yield back to the event loop. Since flushing an update cycle is synchronous, `onStorage` cannot interleave with `MAKE_UPDATE`.

Flushing the keys updating from storage at the end of the update cycle would not be accurate if we had async update cycles, because we would need to track which update cycle is linked to which key. 

e.g.

```
MAKE_UPDATE from storage for field "foo"
MAKE_UPDATE from another entry point
MAKE_UPDATE from storage ends. 
```

Technically, the second update cycle could update "foo", because it doesn't originate from the storage event. At the end of the storage update cycle, we would need to remove the "foo" key, but not other keys that could have been added in the meantime.  This clearly shows that async update cycles present other design problems, that should be addressed for other queues as well if ever implemented.

That's why I'd stick with the simple implementation in the end. If the storage event triggers nested update cycles, it will be in a sync fashion, which works well with the current implementation.

alexkuhn (approved):
@robodoo r+

PR #245771, opened 2026-01-26
[IMP] iot_drivers: enable custom handlers from db

loouis-t:
We now allow toggling the "custom handlers" option from the database.

see odoo/enterprise#105550
see odoo/upgrade#9352


Ysoroko (approved):
The feature is working as expected on the iot box

PR #245773, opened 2026-01-26
[FIX] mail: chat window fold/close buttons should have hover effect

alexkuhn:
These 2 buttons had no hover effect, compared to other buttons like the "start a call". The lack of hover effect makes it harder to click on these buttons.

This commit reduces slightly the opacity of button when not hovered, so that they are highlighted on hover from change to 100% opacity. The reduced opacity also makes these buttons slightly less visible, which is actually an improvement.

Before / After (hover on "x")
Screenshot 2026-01-26 at 17 25 10
Screenshot 2026-01-26 at 17 14 40


tsm-odoo (approved):
@robodoo r+

PR #245777, opened 2026-01-26
[FIX] payment_stripe: prevent crash on webhooks for unknown transactions

pedrambiria:
Prior to this commit, receiving a Stripe webhook notification for a transaction that does not exist in Odoo (e.g., Point of Sale "card_present" payments) would cause an Internal Server Error (500).

The issue occurred because the controller attempted to call `_verify_signature(tx_sudo)` even when `tx_sudo` was empty. Inside `_verify_signature`, accessing `tx_sudo.provider_id` on an empty record triggered a "ValueError: Expected singleton" crash.

Consequently, Stripe would interpret these 500 errors as a server failure and automatically disable the webhook endpoint, breaking online payments.

This commit fixes the issue by checking if `tx_sudo` exists before proceeding with signature verification. If no transaction is found, Odoo now logs a warning and returns a 200 OK status to gracefully acknowledge and ignore the extraneous event.

opw-5439528

---

### DIFF
--- a/addons/payment_stripe/controllers/main.py
+++ b/addons/payment_stripe/controllers/main.py
@@ -85,6 +85,10 @@ class StripeController(http.Controller):
                 tx_sudo = request.env['payment.transaction'].sudo()._search_by_reference(
                     'stripe', data
                 )
+
+                if not tx_sudo:
+                    return request.make_json_response('')
+
                 self._verify_signature(tx_sudo)
 
                 if event['type'].startswith('payment_intent'):  # Payment operation.
--- a/addons/payment_stripe/tests/test_stripe.py
+++ b/addons/payment_stripe/tests/test_stripe.py
@@ -147,6 +147,20 @@ class StripeTest(StripeCommon, PaymentHttpCommon):
             self.assertEqual(signature_check_mock.call_count, 1)
 
     @mute_logger('odoo.addons.payment_stripe.controllers.main')
+    @mute_logger('odoo.addons.payment_stripe.models.payment_transaction')
+    def test_webhook_notification_skips_signature_verification_for_missing_transactions(self):
+        """ Test that the webhook ignores signature verification for unknown transactions (e.g. POS). """
+        url = self._build_url(StripeController._webhook_url)
+        payload = dict(self.payment_data)
+        payload['data']['object']['description'] = None
+
+        with patch(
+            'odoo.addons.payment_stripe.controllers.main.StripeController._verify_signature'
+        ) as signature_check_mock:
+            self._make_json_request(url, data=payload)
+            self.assertEqual(signature_check_mock.call_count, 0)
+
+    @mute_logger('odoo.addons.payment_stripe.controllers.main')
     def test_return_from_tokenization_request(self):
         tx = self._create_transaction('direct', amount=0, operation='validation', tokenize=True)
         url = self._build_url(StripeController._return_url)


addons/payment_stripe/tests/test_stripe.py
```diff
@@ -233,3 +233,21 @@ def test_stripe_validate_amount_uses_payment_minor_unit(self):
         data['payment_intent'] = self._mock_payment_intent_request()
         tx._validate_amount(data)
         self.assertNotEqual(tx.state, 'error')
+
+    @mute_logger('odoo.addons.payment_stripe.controllers.main')
+    @mute_logger('odoo.addons.payment_stripe.models.payment_transaction')
+    def test_webhook_notification_skips_missing_transaction(self):
+        """ Test that the webhook ignores events for unknown transactions (e.g. POS). """
+        url = self._build_url(StripeController._webhook_url)
+        payload = dict(self.payment_data)
```
vchu-odoo: is casting to dict needed?

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -233,3 +233,21 @@ def test_stripe_validate_amount_uses_payment_minor_unit(self):
         data['payment_intent'] = self._mock_payment_intent_request()
         tx._validate_amount(data)
         self.assertNotEqual(tx.state, 'error')
+
+    @mute_logger('odoo.addons.payment_stripe.controllers.main')
+    @mute_logger('odoo.addons.payment_stripe.models.payment_transaction')
+    def test_webhook_notification_skips_missing_transaction(self):
```
vchu-odoo: could you please move the test after `test_webhook_notification_triggers_signature_check`  to keep webhook tests close to each other?

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -233,3 +233,21 @@ def test_stripe_validate_amount_uses_payment_minor_unit(self):
...
+        """ Test that the webhook ignores events for unknown transactions (e.g. POS). """
```
vchu-odoo:
```suggestion
        """ Test that the webhook ignores signature verification for unknown transactions (e.g. POS). """
```

addons/payment_stripe/tests/test_stripe.py
vchu-odoo:
```suggestion
    def test_webhook_notification_skips_signature_verification_for_missing_transactions(self):
```

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -233,3 +233,21 @@ def test_stripe_validate_amount_uses_payment_minor_unit(self):
...
+        payload['data']['object']['description'] = None
+
+        with patch(
+            'odoo.addons.payment_stripe.controllers.main.StripeController._verify_signature'
+        ) as signature_check_mock:
+            self._make_json_request(url, data=payload)
+
+            self.assertEqual(
+                signature_check_mock.call_count, 0,
+                msg="The signature verification should be skipped if no transaction is found."
```
vchu-odoo: no need for that as we mentioned it in the docstring already
```suggestion
```

addons/payment_stripe/controllers/main.py
```diff
@@ -85,6 +85,11 @@ def stripe_webhook(self):
                 tx_sudo = request.env['payment.transaction'].sudo()._search_by_reference(
                     'stripe', data
                 )
+
+                if not tx_sudo:
+                    _logger.warning("No transaction found for the received payment data; skipping")
```
vchu-odoo:
```suggestion
```
we already log missing txs here:
https://github.com/odoo/odoo/blob/fbdcb07f50c1c6ba3449afcf41f0144a08413a11/addons/payment_stripe/models/payment_transaction.py#L301-L302

vchu-odoo (commented):
Thank you for your fix!

LGTM, only some suggestions to apply.

addons/payment_stripe/tests/test_stripe.py
pedrambiria: I made a copy

addons/payment_stripe/tests/test_stripe.py
pedrambiria: The test checks if it ignores the event by checking if it doesn't call signature verification. So it's right.

addons/payment_stripe/tests/test_stripe.py
pedrambiria: The test checks if it ignores the event by checking if it doesn't call signature verification. So it's right.

addons/payment_stripe/tests/test_stripe.py
vchu-odoo: indeed, sorry, brain glitch

addons/payment_stripe/tests/test_stripe.py
vchu-odoo: I suggested it only to make it more precise about what we are testing, because events can happen before the signature verification and we do not cover them.

pedrambiria: I see. I modified it :)

vchu-odoo (approved):
Thank you for the changes!

@robodoo r+

PR #245782, opened 2026-01-26
[FIX] product-labels: Redundant HTML head in DOM output

gergoe:
Both product.report_productlabel and product.report_productlabel_dymo are called from a template which already includes the necessary DOM for the HTML (via web.basic_layout), having a web.html_container call in these two templates seems to be redundant and generates a malformed HTML output.

Current behavior before PR:
Duplicate/redundant DOM structure in reports generated when printing product labels

Correct DOM structure, valid HTML

---


addons/product/report/product_product_templates.xml
```diff
@@ -127,86 +127,82 @@
...
         <template id="report_productlabel_dymo">
-            <t t-call="web.html_container">
-                <t t-set="barcode_size" t-value="'width:45.5mm;height:7.5mm'"/>
-                <t t-set="table_style" t-value="'width:100%;height:32mm;'"/>
-                <t t-set="padding_page" t-value="'padding: 2mm'"/>
-                <t t-foreach="quantity.items()" t-as="barcode_and_qty_by_product">
-                    <t t-set="product" t-value="barcode_and_qty_by_product[0]"/>
-                    <t t-foreach="barcode_and_qty_by_product[1]" t-as="barcode_and_qty">
-                        <t t-set="barcode" t-value="barcode_and_qty[0]"/>
-                        <t t-foreach="range(barcode_and_qty[1])" t-as="qty">
```
rachid-ktk, line 204: - [ ]

PR #241571, opened 2025-12-29
[FIX] project_purchase : Check for account_id in distribution when multiple

yoba-odoo:
### Steps to reproduce:
	- Create a billable Project
	- Create a Purchase order and set the created project and a department in analytic distribution
	- Create a Vendor Bill with the same analytic distribution and match with the PO
	- Confirm the Vendor Bill
	- Check the project dashboard
	- Notice the amount is under To Bill not Billed

### Cause:
When checking the invoice lines we check if the AA in the distribution is the same as the one in the project. The problem comes when we have a Project and a Department set in the distribution as the keys in the JSON dict will be 'account_id,department_id' so the comma will trigger an incosistency so the invoice_line will be filtered out so we will consider this line as still waiting to be invoiced

### Fix:
Make sure we cover the case of the comma presence in one of the JSON keys.

opw-5350246

PR #245793, opened 2026-01-26
[FIX] project_purchase : Check for AA anywhere in the analytic dist

yoba-odoo:
### Steps to reproduce:
	- Create a billable Project
	- Navigate to Accounting > Configuration > Analytic Accounting > Analytic Plans
	- Change the order of the project plan
	- Create a Purchase order and set the created project and a department in analytic distribution
	- Create a Vendor Bill with the same analytic distribution and match with the PO
	- Confirm the Vendor Bill
	- Check the project dashboard
	- Notice the amount is under To Bill not Billed

### Cause:
In this commit https://github.com/odoo/odoo/pull/241571/changes/ef080f94609f1057c6d86af68bee605dcaeb287b we introduced a fix to search for the analytic account in purchase lines' analytic distribution when we have multiple accounts for the same purchase line if it is shown as the first number of the key but since it is not mandatory to have the project account id at the start of the key

### Fix:
We now search for the id in the whole not only the start of it.

opw-5350246

### DIFF
--- a/addons/project_purchase/models/project_project.py
+++ b/addons/project_purchase/models/project_project.py
@@ -165,8 +165,7 @@ class ProjectProject(models.Model):
                         l.parent_state != 'cancel'
                         and l.analytic_distribution
                         and any(
-                            key == str(self.account_id.id)
-                            or key.startswith(str(self.account_id.id) + ",")
+                            str(self.account_id.id) in key.split(',')
                             for key in l.analytic_distribution
                         )
                     )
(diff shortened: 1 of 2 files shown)


deneuvillem (commented):
Thanks for your fix, it looks good to me :)
Could you please add a test case to ensure it behaves as expected? Thanks!

deneuvillem:
robodoo r+

PR #245795, opened 2026-01-26
[FIX] l10n_gcc_pos: Add arabic text to receipts

mobar-odoo:
Problem:
When printing a receipt in arabic using the l10n_gcc_pos module, some of the text is in English.

Cause:
Translation is not enabled for the module and the text is written in English only in the receipts XML.

Solution:
Add the arabic translations of texts to the receipts XML and choose the display language based on the user's language (same behaviour in other receipts).

Steps to reproduce:
- Install l10n_gcc_pos module
- Activate and choose Arabic as the language
- Open Point of Sale and validate an order
- See how some text (specifically "Tax Invoice" and "Simplified Tax Invoice") are printed in English although the rest of the receipt is printed in Arabic.

opw-5501464


robinengels (approved):
Fix LGTM, just a bit of nitpicking but we usually put the `opw-` at the end of the commit message.

manv-afk (approved):

stva-odoo:
robodoo r+

PR #245796, opened 2026-01-26
[FIX] point_of_sale:include settled order payments

robinengels:
If you made a PoS order paid with the customer account payment method, and then you created a settle due order to settle the previous one. If you then create the invoice for the original order, the invoice would appear as unpaid, because the payments of the settle due order were not taken into account.

Steps to reproduce:
-------------------
* Create a PoS order and pay with the customer account payment method
* Settle the order that you just created with a settle due order
* Close the session
* Go on the original order and create the invoice
> Observation: The invoice appears as unpaid when it should be paid.

Why the fix:
------------
When creating the invoice we gather all the payments of the order to create the corresponding journal entries. But the payment of the settle due order were not included. So the order was considered as unpaid.

opw-5268042

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -1096,6 +1096,9 @@ class PosOrder(models.Model):
     def _get_invoice_post_context(self):
         return {"skip_invoice_sync": True}
 
+    def _get_payments(self):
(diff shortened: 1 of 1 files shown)


stva-odoo (commented):
robodoo r+

PR #245826, opened 2026-01-27
[FIX] web, project, payment: Fix extractProps and update props in CopyClipboardField

Mazen-Ghaleb:
The `CopyClipboardField` widget previously relied on the `string` attribute inside `attributes` to determine the text of copy button. In Odoo 17+, the `string` attribute is no longer inside `attributes` but defined outside, causing the widget to break. Additionally, using `string` for copy text caused confusion between the field label and the text to copy (`copyText`). The correct approach is to use widget options for defining `copyText`, as intended in the project share link. In the Payment Wizard button, the widget was inherited and `extractProps` was overridden to handle the `string` attribute manually, which is no longer necessary with the new approach.

- `CopyClipboardField.extractProps` reads the removed `string` attribute from `attributes`.  
- The copy text could be confused with the field label (`string`).  
- Project share link usage fails to read copy text correctly because the widget does not read options.  
- `PaymentWizardCopyClipboardButtonField` overrides `extractProps` unnecessarily to fix copy text handling.

- `CopyClipboardField` reads `copyText` from widget options, no longer relying on the removed `string` attribute.  
- Copy text is clearly separated from field labels.  
- Project share link now works correctly using the new `copyText` option.  
- `PaymentWizardCopyClipboardButtonField` no longer needs to override `extractProps` and copy text is set directly in field options  

---

### DIFF
--- a/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
+++ b/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
@@ -22,10 +22,6 @@ class PaymentWizardCopyClipboardButtonField extends CopyClipboardButtonField {
 const paymentWizardCopyClipboardButtonField = {
     ...copyClipboardButtonField,
     component: PaymentWizardCopyClipboardButtonField,
-    extractProps: (fieldInfo, dynamicInfo) => ({
-        ...copyClipboardButtonField.extractProps(fieldInfo, dynamicInfo),
-        copyText: fieldInfo.string,
-    }),
 };
 
 registry
--- a/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
+++ b/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
@@ -24,7 +24,7 @@ const paymentWizardCopyClipboardButtonField = {
     component: PaymentWizardCopyClipboardButtonField,
     extractProps: (fieldInfo, dynamicInfo) => ({
         ...copyClipboardButtonField.extractProps(fieldInfo, dynamicInfo),
-        string: fieldInfo.string,
+        copyText: fieldInfo.string,
     }),
 };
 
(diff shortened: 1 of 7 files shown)


Mazen-Ghaleb:
> This is not a bad idea though the options' values are not translated. I cannot merge this, though if you have a fix for the breaking case (which I fail to see directly here) that might be done

Good catch, thank you for pointing this out.

I spent some time investigating the issue and explored a few possible approaches before settling on a final solution. I’ll summarize them briefly and explain how the implemented approach addresses the breaking case you mentioned.

**1. Using the field string as copy text**
We could reuse the field’s `string` attribute as the copy text and define a separate tag for labels. However, this is not very intuitive, as the copy text could easily be confused with the field label itself.

**2. Returning translated text via a model method**
Another option was to define a method returning the translated copy text (using`@api.model`) and inject its name through options, attributes, or context. While functional, this would require an ORM call from the widget, adds unnecessary complexity, and does not feel very Odoo-idiomatic.

**3. Using a referenced field (similar to color in m2m fields or currency fields in monetary fields)**
In this approach, the widget reads a field defined on the model. I implemented the solution assuming the field would usually be **non-stored** to avoid storing of redundant data. Since the value is not stored, it is not present in the record data, so the widget evaluates it virtually and does not require any ORM calls.

For the project and payment references specifically, I used **non-stored fields** and defined the original message in the field’s default value with proper translation. This resolves the breaking case where option values were not translated, as the translation now happens at the Python model level rather than in the view options.

This approach worked well, so I also extended it to support a translated **success text**. I added unit tests for the copy-to-clipboard field (covering both copy text and success text) and updated the payment wizard copy widget tests accordingly.

I have squashed my previous commits, interactively rebased all changes into a single commit, and rebased the branch onto the latest v17. 

Please review, and reopen the PR if everything looks good. @kebeclibre @gdrius @xrg 

One final note: the translation files will need to be updated, since translations are now handled in the Python models rather than the XML views.

Thank you.

PR #245830, opened 2026-01-27
[FIX] pos_self_order: reset active category after order confirmation

parp-odoo:
The Product List page was keeping the previously selected category active after checkout in kiosk mode.

This commit resets the category selection on order confirmation.

Task-5877912


davidmonnom (approved):
@robodoo r+

PR #245832, opened 2026-01-27
[FIX] l10n_sa_edi: allow additional buyer ID for non-Saudi partners

hash-odoo:
This commit makes l10n_sa_additional_identification_number visible for non-Saudi individuals and company partners when the active company is in Saudi Arabia and keeps the identification scheme fixed to OTH, keeping it invisible.

When generating ZATCA XML for non-Saudi partners, VAT is used as the primary buyer ID if present, and fall back to the additional identification number when VAT is missing.

task-4525956

### DIFF
--- a/addons/l10n_sa_edi/views/res_partner_views.xml
+++ b/addons/l10n_sa_edi/views/res_partner_views.xml
@@ -9,7 +9,7 @@
             <field name="arch" type="xml">
                 <xpath expr="//div[hasclass('o_address_format')]"  position="after">
                     <field name="l10n_sa_additional_identification_scheme" invisible="country_code != 'SA'"/>
-                    <field name="l10n_sa_additional_identification_number" invisible="country_code != 'SA' or l10n_sa_additional_identification_scheme == 'TIN'"/>
+                    <field name="l10n_sa_additional_identification_number" invisible="'SA' not in fiscal_country_codes and country_code != 'SA' or l10n_sa_additional_identification_scheme == 'TIN'"/>
                 </xpath>
             </field>
         </record>
--- a/addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
+++ b/addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
@@ -104,7 +104,7 @@ class AccountEdiXmlUBL21Zatca(models.AbstractModel):
         """ Override to include/update values specific to ZATCA's UBL 2.1 specs """
         identification_number = partner.l10n_sa_additional_identification_number
         vat = re.sub(r'[^a-zA-Z0-9]', '', partner.vat or "")
-        if partner.country_code != "SA":
+        if partner.country_code != "SA" and vat:
             identification_number = vat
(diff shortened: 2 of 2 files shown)


addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -24,6 +24,16 @@ class ResPartner(models.Model):
     l10n_sa_additional_identification_number = fields.Char("Identification Number (SA)",
                                                            help="Additional Identification Number for Seller/Buyer")
 
+    l10n_sa_additional_identification_number_visibility = fields.Boolean(
+        compute="_compute_l10n_sa_additional_identification_number_visibility")
+
+    @api.depends_context('company')
+    @api.depends('company_id')
```
hamo-odoo: Not sure how does this depends on the `company_id`?

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -24,6 +24,16 @@ class ResPartner(models.Model):
...
+    def _compute_l10n_sa_additional_identification_number_visibility(self):
+        self.update({
+            'l10n_sa_additional_identification_number_visibility': self.env.company.country_code != 'SA'
+        })
```
hamo-odoo: Why use update if you can directly assign the value here to the field?

addons/l10n_sa_edi/views/res_partner_views.xml
```diff
@@ -9,7 +9,7 @@
             <field name="arch" type="xml">
                 <xpath expr="//div[hasclass('o_address_format')]"  position="after">
                     <field name="l10n_sa_additional_identification_scheme" invisible="country_code != 'SA'"/>
-                    <field name="l10n_sa_additional_identification_number" invisible="country_code != 'SA' or l10n_sa_additional_identification_scheme == 'TIN'"/>
+                    <field name="l10n_sa_additional_identification_number" invisible="l10n_sa_additional_identification_number_visibility or l10n_sa_additional_identification_scheme == 'TIN'"/>
```
hamo-odoo: Shouldn't ` or l10n_sa_additional_identification_scheme == 'TIN'` be a part of `l10n_sa_additional_identification_number_visibility` ?

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -24,6 +24,16 @@ class ResPartner(models.Model):
     l10n_sa_additional_identification_number = fields.Char("Identification Number (SA)",
                                                            help="Additional Identification Number for Seller/Buyer")
 
+    l10n_sa_additional_identification_number_visibility = fields.Boolean(
+        compute="_compute_l10n_sa_additional_identification_number_visibility")
+
+    @api.depends_context('company')
```
hamo-odoo: Better to use fiscal_country_codes may be?

hash-odoo: Agreed, that is already being done from 18.3. This will remove the need of computed field.

hamo-odoo (approved):
LGTM

jbw-odoo (approved):
Looks good, thx
robodoo r+

PR #245836, opened 2026-01-27
[IMP] l10n_sa_edi: remove l10n_sa_is_retention field

pssh-odoo:
Before this commit:
- The user needed to use the boolean field `l10n_sa_is_retention` on a negative tax for it to be considered a retention tax. This complicates the flow, as any negative tax is a retention tax by default, so the need for the checkbox is unnecessary.

After this commit:
- The boolean field `l10n_sa_is_retention` is removed.

Related: https://github.com/odoo/upgrade/pull/9351

Task [link](https://www.odoo.com/odoo/project.task/5865831)
task-5865831

### DIFF
--- a/addons/l10n_sa_edi/models/account_move.py
+++ b/addons/l10n_sa_edi/models/account_move.py
@@ -332,17 +332,6 @@ class AccountMove(models.Model):
 class AccountMoveLine(models.Model):
     _inherit = 'account.move.line'
 
-    def _apply_retention_tax_filter(self, tax_values):
-        return tax_values['tax_id'].amount < 0
-
-    def _is_global_discount_line(self):
-        """
-            Any line that has a negative amount and is not linked to a down-payment is considered as a
-            global discount line. These can be created either manually, or through a promotions program.
-        """
-        self.ensure_one()
-        return not self._get_downpayment_lines() and self.price_subtotal < 0
-
     @api.depends('price_subtotal', 'price_total')
     def _compute_tax_amount(self):
         super()._compute_tax_amount()
--- a/addons/l10n_sa_edi/models/account_move.py
+++ b/addons/l10n_sa_edi/models/account_move.py
@@ -333,7 +333,7 @@ class AccountMoveLine(models.Model):
     _inherit = 'account.move.line'
 
     def _apply_retention_tax_filter(self, tax_values):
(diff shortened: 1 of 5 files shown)


khva-odoo (approved):

addons/l10n_sa_edi/models/account_move.py
```diff
@@ -359,5 +359,5 @@ def _compute_tax_amount(self):
                 line.l10n_gcc_invoice_tax_amount = sum(
                     tax_data['tax_amount_currency']
                     for tax_data in base_line['tax_details']['taxes_data']
-                    if not tax_data['tax'].l10n_sa_is_retention
+                    if not (tax_data['tax'].amount < 0)
```
malb-odoo: Or you need to remove the not ? just by looking at the _apply_retention_tax_filter logic i suppose it should be that ?
```suggestion
                    if tax_data['tax'].amount > 0
```

malb-odoo (commented):
Just a question here ?

addons/l10n_sa_edi/models/account_move.py
pssh-odoo: Yeah, that makes more sense. 
I have made the changes, could you please check?

addons/l10n_sa_edi/models/account_move.py
```diff
@@ -333,7 +333,7 @@ class AccountMoveLine(models.Model):
     _inherit = 'account.move.line'
 
     def _apply_retention_tax_filter(self, tax_values):
-        return not tax_values['tax_id'].l10n_sa_is_retention
+        return tax_values['tax_id'].amount < 0
```
malb-odoo: Here too no ? If l10n_sa_is_retention ==> amount < 0 then there since it's not l10n_sa_is_retention it should be ==> amount > 0 ?

malb-odoo (commented):
Not sure what logic is the good one there ahah

addons/l10n_sa_edi/models/account_move.py
pssh-odoo: Upon further review, this function can be removed entirely as it is no longer used.
I have added a new commit to do so, could you please check?

malb-odoo (approved):
waiting on @ygu-odoo input theeen but technically seems ok now 😄

PR #245842, opened 2026-01-27
[IMP] hr_holidays: make refused allocation requests read-only

rame-odoo:
Purpose:
- Align allocation requests with time off requests by preventing any edits once an allocation is refused, ensuring a consistent and intuitive user experience.

This PR includes:
- Made all allocation fields read-only when the request is in Refused state.
- Applied the restriction consistently across regular and accrual allocations.
- Kept existing approval and validation flows unchanged.

task-5498398

### DIFF
--- a/addons/hr_holidays/views/hr_leave_allocation_views.xml
+++ b/addons/hr_holidays/views/hr_leave_allocation_views.xml
@@ -129,30 +129,32 @@
                             <field name="already_accrued" invisible="1"/>
                             <field name="holiday_status_id"
                                 context="{'employee_id':employee_id, 'default_date_from':current_date, 'request_type':'allocation'}"
-                                readonly="state == 'validate'"/>
+                                readonly="state in ['refuse', 'validate']"/>
                             <field name="allocation_type"
                                 widget="radio"
                                 invisible="1"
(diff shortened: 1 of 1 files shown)


addons/hr_holidays/views/hr_leave_allocation_views.xml
```diff
@@ -118,7 +118,7 @@
                     <div id="title" class="oe_title">
                         <h2><field name="name" class="w-100"
                             placeholder="e.g. Time Off type (From validity start to validity end / no limit)"
-                            invisible="state != 'confirm'" readonly="True" force_save="True"/>
```
amry-odoo, line 121: Why did we change this?

amry-odoo (changes requested):
Thanks for your work! Just a small question then would be good on my side

addons/hr_holidays/views/hr_leave_allocation_views.xml
rame-odoo: Yeah!! This is the base employee view. readonly="True" should stay here. Thanks for pointing out! :raised_hands:

amah-odoo (approved):
@robodoo r+

PR #245844, opened 2026-01-27
[IMP] orm: get now() when fetching sequences

kmagusiak:
We can get the time for free avoiding a separate query later. We usually start after checking sequences and we often use the date in the same transaction.

---


rco-odoo (approved):
@robodoo r+

PR #245847, opened 2026-01-27
[FIX] stock: fix quantities at past date for sub-locations

mohammadhossam:
Bug:
When querying product quantities at a past date using `to_date` context, the computed qty_available was incorrect for sub-locations. If a single stock.move had multiple stock.move.line records going to different sub-locations, the quantities would be attributed to the move's destination location (parent) rather than the actual sub-locations where the goods were placed via move lines.

Example: A receipt move with location_dest=Stock and product_qty=5, split into two move lines (3 units to Stock/Shelf-A, 2 units to Stock/Shelf-B), would show 5 units at Stock and 0 at each sub-location when querying past quantities, instead of 3 and 2 respectively.

Expected behavior:
Querying qty_available with `to_date` context for a specific location should return the actual quantity that was moved to that location, respecting the move line destinations.

Solution:
In `_compute_quantities_dict`, changed the past date calculation from aggregating `product_qty` on stock.move to aggregating `qty_done` on stock.move.line. This correctly accounts for the actual quantities moved per location.

### DIFF
--- a/addons/stock/models/product.py
+++ b/addons/stock/models/product.py
@@ -155,7 +155,7 @@ class ProductProduct(models.Model):
             product.with_context(skip_qty_available_update=True).update({key: val for key, val in res[product.id].items() if val})
 
     def _compute_quantities_dict(self, lot_id, owner_id, package_id, from_date=False, to_date=False):
-        domain_quant_loc, domain_move_in_loc, domain_move_out_loc = self._get_domain_locations()
+        domain_quant_loc, domain_move_in_loc, domain_move_out_loc, domain_move_in_loc_done, domain_move_out_loc_done = self._get_domain_locations()
         domain_quant = [('product_id', 'in', self.ids)] + domain_quant_loc
         dates_in_the_past = False
         # only to_date as to_date will correspond to qty_available
@@ -180,8 +180,12 @@ class ProductProduct(models.Model):
         if package_id is not None:
             domain_quant += [('package_id', '=', package_id)]
         if dates_in_the_past:
-            domain_move_in_done = list(domain_move_in)
-            domain_move_out_done = list(domain_move_out)
+            # Use the done location domains (simple location_dest_id) for move lines
+            domain_move_in_done = [('product_id', 'in', self.ids)] + domain_move_in_loc_done
+            domain_move_out_done = [('product_id', 'in', self.ids)] + domain_move_out_loc_done
+            if owner_id is not None:
+                domain_move_in_done += [('owner_id', '=', owner_id)]
+                domain_move_out_done += [('owner_id', '=', owner_id)]
         if from_date:
             date_date_expected_domain_from = [('date', '>=', from_date)]
             domain_move_in += date_date_expected_domain_from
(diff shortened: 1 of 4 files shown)


mohammadhossam:
To the future reviewer..
I'm not sure about the correct target version in this case, but I just targeted the first available version where the issue occurs.

addons/stock/models/product.py
```diff
@@ -180,8 +180,12 @@ def _compute_quantities_dict(self, lot_id, owner_id, package_id, from_date=False
         if package_id is not None:
             domain_quant += [('package_id', '=', package_id)]
         if dates_in_the_past:
-            domain_move_in_done = list(domain_move_in)
-            domain_move_out_done = list(domain_move_out)
+            # Use the done location domains (simple location_dest_id) for move lines
+            domain_move_in_done = [('product_id', 'in', self.ids)] + domain_move_in_loc_done
+            domain_move_out_done = [('product_id', 'in', self.ids)] + domain_move_out_loc_done
+            if owner_id is not None:
+                domain_move_in_done += [('move_id.restrict_partner_id', '=', owner_id)]
```
Whenrow: There is `owner_id` on stock move line and I may happen a stock move is composed of 2 stock move line. One having an owner while the second doesn't

addons/stock/models/product.py
```diff
@@ -206,14 +211,15 @@ def _compute_quantities_dict(self, lot_id, owner_id, package_id, from_date=False
         moves_out_res_past = defaultdict(float)
         if dates_in_the_past:
             # Calculate the moves that were done before now to calculate back in time (as most questions will be recent ones)
-            domain_move_in_done = [('state', '=', 'done'), ('date', '>', to_date)] + domain_move_in_done
-            domain_move_out_done = [('state', '=', 'done'), ('date', '>', to_date)] + domain_move_out_done
+            # Use move_id.date (not move line date) as the move line's date gets updated when quantities change
```
Whenrow: move line date is also updated at the validation and is never changed after (even if unlocked) so I would use the stock move line date field to avoid a potential additional join

addons/stock/models/product.py
```diff
@@ -206,14 +211,15 @@ def _compute_quantities_dict(self, lot_id, owner_id, package_id, from_date=False
...
+            domain_move_in_done = [('state', '=', 'done'), ('move_id.date', '>', to_date)] + domain_move_in_done
+            domain_move_out_done = [('state', '=', 'done'), ('move_id.date', '>', to_date)] + domain_move_out_done
 
             groupby = ['product_id', 'uom_id']
-            for product, uom, quantity in Move._read_group(domain_move_in_done, groupby, ['quantity:sum']):
+            for product, uom, quantity in MoveLine._read_group(domain_move_in_done, groupby, ['quantity:sum']):
```
Whenrow: The stock move lines have the field `quantity_product_uom` already computed in the product's uom. So you can get rid of the group on UoM

Whenrow (changes requested):

PR #245848, opened 2026-01-27
[FIX] l10n_be_pos_restaurant: fix duplicate taxes on branches restaurant

macocianradu:
When creating a restaurant with the POS scenario on a company with the Belgian accounting package, we automatically create a new 'Take out' fiscal position and a 'Take out' 6% tax. But when creating a branch for the company and using the same scenario, it will duplicate the 'Take out' fiscal position, and then fail to create because it will try to create a duplicate Tax with the same name and then odoo will complain.

After the fix the scenario checks whether the fiscal position and tax already exist on the company, and if they do, it doesn't recreate them. This should let branches use the scenario and then they can use the already existing taxes and FPs.

Task [link](https://www.odoo.com/odoo/project.task/5502948)
Task-5502948

### DIFF
--- a/addons/l10n_be_pos_restaurant/models/pos_config.py
+++ b/addons/l10n_be_pos_restaurant/models/pos_config.py
@@ -1,4 +1,5 @@
-from odoo import api, models, Command
+from odoo import Command, api, models
+
 
 class PosConfig(models.Model):
     _inherit = 'pos.config'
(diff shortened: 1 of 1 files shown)


lordkrandel (commented):
I don't like very much finding fiscal positions and taxes by `name`, I generally prefer when they have a specific `xml_id`, or some dominant characteristics.

Also, I also found the same code in `saas-18.3`, which is an active runbot target. Don't we wanna fix that too? This would probably cause fw-ports conflicts in the future.
https://github.com/odoo/odoo/blob/saas-18.3/addons/l10n_be_pos_restaurant/models/pos_config.py

The code is good, my remarks are made, I delegate the `r+` to the PoS team which has the domain.

loukwets (approved):
I don’t think finding fiscal positions and taxes by name is an issue in this case. They are scenario-specific, and since this is usually used for onboarding, it’s an acceptable consequence if someone changes the scenario.

Also discussed with Mobt to confirm that this should only be done in 19.0 since we might no longer onboard in 18.3.

macocianradu:
@loukwets @davidmonnom @lordkrandel 
I added a xmlid to the fiscal position and the tax and check on the xmlid on creation instead of the name

addons/l10n_be_pos_restaurant/models/pos_config.py
```diff
@@ -10,14 +11,28 @@ def _create_takeaway_fiscal_position(self, config):
...
-            tax_6.copy({
-                'name': f"{tax_6.name} Take out",
-                'fiscal_position_ids': [Command.set(fp.ids)],
-                'original_tax_ids': [Command.set((tax_12 | tax_21).ids)],
-            })
+            fp = ChartTemplate.ref('pos_takeout_fp', raise_if_not_found=False)
+            if not fp:
+                fp = self.env['account.fiscal.position'].create({
+                    'name': 'Take out',
+                })
```
lordkrandel, line 18:
```suggestion
            fp = (
                ChartTemplate.ref('pos_takeout_fp', raise_if_not_found=False)
                or self.env['account.fiscal.position'].create({'name': 'Take out'})
            )
```

addons/l10n_be_pos_restaurant/models/pos_config.py
```diff
@@ -10,14 +11,28 @@ def _create_takeaway_fiscal_position(self, config):
...
+                self.env['ir.model.data']._update_xmlids([{
+                    'xml_id': ChartTemplate.company_xmlid('pos_takeout_fp'),
+                    'record': fp,
+                    'noupdate': True,
+                }])
+            takeout_tax = ChartTemplate.ref('pos_takeout_tax', raise_if_not_found=False)
+            if not takeout_tax:
+                takeout_tax = tax_6.copy({
```
lordkrandel, line 26:
```suggestion
            takeout_tax = (
                ChartTemplate.ref('pos_takeout_tax', raise_if_not_found=False)
                or tax_6.copy({
```

addons/l10n_be_pos_restaurant/models/pos_config.py
macocianradu: Would that look like this?
```
            fp = (
                ChartTemplate.ref('pos_takeout_fp', raise_if_not_found=False)
                or self.env['account.fiscal.position'].create({'name': 'Take out'})
            )
            self.env['ir.model.data']._update_xmlids([{
                'xml_id': ChartTemplate.company_xmlid('pos_takeout_fp'),
                'record': fp,
                'noupdate': True,
            }])
```
In that case, I'm not sure it's better to remove the if here. 
If I remove it, I would update the xmlid every time (even when I just got the object using the same id)

lordkrandel: ... Right, sorry. ^^

lordkrandel:
Ping @davidmonnom

lordkrandel:
I asked for a quick rebase since the runbot was complaining.

@robodoo r+

PR #245850, opened 2026-01-27
Keep delayed translations with custom snippet

blse-odoo:
The feature "delayed translation" and "copy translation in custom snippet" have been worked on in parallel, but had buggy interactions.

This commit changes the way translations terms are copied in custom snippets to read and write the translated values as expected with delayed translations.

Steps to reproduce:
- With a page with some content on a website in 2 languages
- Open the website builder in the main language
- Drop a custom snippet
- Save
- Bug: The change is immediate on the version in the second language (instead of being only available after user translated it)

Delayed translations: 2d08f97c0778469b409fca23f2be5f5a98ce3df8
Copy translation in custom snippet: d3426b7714012e833caae10281cfb8433223299a
Re-enabling delayed translations: 03a85b13b2c46ef7174123d902e95d5103031c6c
task-5474184

### DIFF
--- a/addons/html_editor/models/ir_ui_view.py
+++ b/addons/html_editor/models/ir_ui_view.py
@@ -198,13 +198,23 @@ class IrUiView(models.Model):
         langs.add('en_US')
 
         # 2. Set translations
-        new_value = {
-            lang: field_to.translate(lambda term: translation_dictionary.get(term, {}).get(lang), record_to[name_field_to])
-            for lang in langs
-        }
-        record_to.env.cache.update_raw(record_to, field_to, [new_value], dirty=True)
+        # The orm layer currently does not offer a way to keep the delayed
+        # translations when changing translations, so we craft the desired
+        # dict of translations and directly write it to the cache
+        stored_translation = field_to._get_stored_translations(record_to)
+        for lang in langs:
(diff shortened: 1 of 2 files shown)


addons/html_editor/models/ir_ui_view.py
loco-odoo, line 189: For this step, should you also check the values for possible `_lang` ?

addons/website/static/tests/tours/delayed_translation_with_custom_snippet.js
```diff
@@ -0,0 +1,126 @@
...
+ * -> add the custom snippet into the page
+ * -> save
+ * -> go to second language
+ * -> check the last translated version is kept
+ * -> enter translate mode
+ * -> check the custom snippet got the translation from the user
+ */
+
+registerWebsitePreviewTour(
+    "delayed_translations_with_custom_snippet",
```
loco-odoo: Maybe part if not all of the test can be done in `py` (something similar to what was done in https://github.com/odoo/odoo/commit/863b31b74966ba226d29af8307def0203b31905b) (by doing as most as possible in `py`, we reduce the resources needed to run the test 👍 )

FrancoisGe:
@robodoo delegate=loco-odoo

loco-odoo:
Thanks for the PR Sebastien!

@robodoo r+

PR #245853, opened 2026-01-27
[IMP] account: backporting helpers for a fix

legi-odoo:
This PR backport helpers functions from 18.3 used to dispatch negatives lines (for EDI refusing negative lines).
It also backport a fix on _split_base_line().
opw-5382423


smetl (approved):
@robodoo r+

PR #245855, opened 2026-01-27
[FIX] base/ir.ui.view: keep tail when remove the field from access group

Gorash:
In a view, if a field is removed due to security access restrictions, the text that follows it is not preserved. This leads to inconsistencies in the view.

opw-5798852


kmagusiak (approved):
@robodoo r+

PR #245857, opened 2026-01-27
[IMP] im_livechat: remove chat hub options for public users

tsm-odoo:
Public users aren't supposed to have many chats active at a time.
For this reason, the chat hub options are overkill and more distracting
than helpful.

This commit removes the chat hub options for non-internal users.

At the same time, this commit fixes the chat hub drag feature that
was not working with `o-liftUp` since [1]. Once the chat hub has
been dragged, we should not move it programatically anymore.

part of task-5867464

[1]: https://github.com/odoo/odoo/pull/225757

### DIFF
--- a/addons/mail/static/src/core/common/chat_hub.scss
+++ b/addons/mail/static/src/core/common/chat_hub.scss
@@ -1,6 +1,9 @@
 .o-mail-ChatHub-bubbles {
     width: $o-mail-ChatHub-bubblesWidth;
     z-index: $o-mail-NavigableList-zIndex - 1;
+}
+
+.o-mail-ChatHub:not(.o-dragged) .o-mail-ChatHub-bubbles {
     bottom: calc(#{var(--mail-ChatHub-bubbles-bottom, 0px)} + #{var(--mail-ChatHub-bubbles-bottomLift, 0px)});
 
     &.o-liftUp {
--- a/addons/mail/static/src/core/common/chat_hub.xml
+++ b/addons/mail/static/src/core/common/chat_hub.xml
@@ -2,14 +2,14 @@
 <templates xml:space="preserve">
 
 <t t-name="mail.ChatHub">
-    <div class="o-mail-ChatHub d-print-none" t-ref="root">
+    <div class="o-mail-ChatHub d-print-none" t-att-class="{'o-dragged': position.dragged}" t-ref="root">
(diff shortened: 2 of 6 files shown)


addons/mail/static/src/core/common/chat_hub.scss
alexkuhn: `bottom` and `--mail-ChatHub-bubbles-bottomLift` should work in `:not(.o-dragged)`, but `width` and `z-index` should be applied even in `o-dragged`.

addons/mail/static/src/core/common/chat_hub.xml
```diff
@@ -2,14 +2,14 @@
...
             <ChatWindow chatWindow="cw" t-if="cw.canShow" right="chatHub.BUBBLE_START + chatHub.BUBBLE + (chatHub.BUBBLE_OUTER*2) + (visibleChatWindows.length - cw_index - 1) * (chatHub.WINDOW + chatHub.WINDOW_INBETWEEN * 2)"/>
         </t>
         <div class="o-mail-ChatHub-bubbles position-fixed d-flex flex-column align-content-start justify-content-end align-items-center" t-attf-style="top: {{position.top}}; left: {{position.left}}; right: {{position.right}}; --mail-ChatHub-bubbles-bottom: {{position.bottom}}" t-att-class="{ 'o-liftUp': busMonitoring.hasConnectionIssues, 'o-mobile': isMobileOS }" t-ref="bubbles">
             <div class="d-flex flex-column align-content-start justify-content-end align-items-center o-gap-1_5">
-                <Dropdown t-if="(chatHub.showConversations and !chatHub.compact) or position.dragged" state="options" position="'top-end'" menuClass="'m-0 mb-1 ' + store.discussDropdownMenuClass(this)">
+                <Dropdown t-if="store.self_user?.share === false and ((chatHub.showConversations and !chatHub.compact) or position.dragged)" state="options" position="'top-end'" menuClass="'m-0 mb-1 ' + store.discussDropdownMenuClass(this)">
```
alexkuhn: I think we can have `optionActions.length` as `t-if` and put condition on the "Hide all" and "Close all" options.

That way we don't show the options when there's no option to show, but we still allow drag and when the options are dragged it shows the options as there's the "Reset position".

alexkuhn (approved):
@robodoo delegate+

tsm-odoo (commented):
@robodoo r+

PR #245860, opened 2026-01-27
[FIX] hr_holidays: wrong allocations in report

MerlinGuillaume:
The balance report aggregates allocations regardless of the period in which they occur resulting in wrong information when an employee has two allocations in different periods and a filter is used in the report

Steps to reproduce:
1. Install Time Off app
2. Open Time Off app and go to Management > Allocations
3. Create a new allocation of 100 days for employee "Mitchell Admin" of time off type "Paid Time Off" with a validity period from 01/01/2027 to 31/12/2027 and approve it
4. Activate debug mode
5. Go to Time Off > Reporting > Balance
6. The number of days left for "Mitchell Admin" for "Paid Time Off" is 120 days even though the default filter is on the period of 2026

Solution:
Refactor the entire query: we order the allocations by date_from so we can subtract taken leaves from the allocations in that order

opw-5156142

### DIFF
--- a/addons/hr_holidays/report/hr_leave_employee_type_report.py
+++ b/addons/hr_holidays/report/hr_leave_employee_type_report.py
@@ -37,84 +37,141 @@ class LeaveReport(models.Model):
 
         self._cr.execute("""
             CREATE or REPLACE view hr_leave_employee_type_report as (
-                SELECT row_number() over(ORDER BY leaves.employee_id) as id,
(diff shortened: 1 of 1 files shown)


amah-odoo:
Hi 👋🏼 

Are you sure about the reproduction steps 👀 
image
image

Bertrand2 (approved):
Thanks for the fix!
robodoo r+

PR #245862, opened 2026-01-27
[FIX] account_edi_ubl_cii: Don't import null LineExtensionAmount lines

Megaaaaaa:
Zero LineExtensionAmount have no purpose after import so they can simply be ignored.

taks-5881008

---

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
@@ -716,6 +716,17 @@ class AccountEdiXmlUBL20(models.AbstractModel):
 
         invoice_line_tag = 'InvoiceLine' if invoice.move_type in ('in_invoice', 'out_invoice') or qty_factor == -1 else 'CreditNoteLine'
         for i, invl_el in enumerate(tree.findall('./{*}' + invoice_line_tag)):
+            # Avoid creating a line if its LineExtensionAmount is missing/empty/zero.
+            line_total_node = invl_el.find('./{*}LineExtensionAmount')
+            if line_total_node is None or not (line_total_node.text and line_total_node.text.strip()):
+                continue
+            try:
+                if float(line_total_node.text) == 0:
+                    continue
+            except (ValueError, TypeError):
+                # If the value is not a valid number, skip creating the line.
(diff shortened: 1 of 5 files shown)


epictete (commented):
LGTM, but could you please target `17.0` instead?

epictete (approved):
@robodoo r+ delegate+

clbr-odoo (commented):
Sorry to intervene like that :see_no_evil: 
I think this is worth backporting to 16.0, and the task explicitly is tagged 16.0.
Peppol unfortunately does not follow regular processes.
We backported it in 16.0 while it was already not supported, it's not to leave it broken just behind.
IMO either it's a nice to have, and should be in 19.0, either it's a must have, and then should be in 16.0.

Isn't it possible to check upfront the value of the invoice line, and don't create the `account.move.line` at all in this case, instead of creating it and unlinking it, which seems quite quick & dirty fixup, and keeps unnecessary computations ? 🤔

Megaaaaaa:
There's no change between 16.0 and 17.0, this would make peppol cleaner for a wider group of people so I agree. Let's go for 16.0. I'll try to change the `unlink()` thing.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -761,6 +772,10 @@ def _import_fill_invoice_line_form(self, journal, tree, invoice, invoice_line, q
 
         # Taxes
         inv_line_vals = self._import_fill_invoice_line_values(tree, xpath_dict, invoice_line, qty_factor)
+        # If the invoice line has no meaningful LineExtensionAmount (empty/null/zero),
+        # the parser returns None. In that case, skip the line.
+        if inv_line_vals is None:
+            return []
```
epictete: This is likely redundant/unnecessary with the code you added in `_import_fill_invoice_form()`, no?
We should no longer reach this code if `LineExtensionAmount` is falsy.

Megaaaaaa: Indeed, fixed. :+1:

epictete (approved):
LGTM (you still have the delegate 😉)

PR #245866, opened 2026-01-27
[FIX] sale: conditional display of sections containing combos in sale preview

nsirjacobs:
**Behavior:**
When a combo product is added under a section and the 'Hide Composition' option is selected the system will try to get a list of the prices grouped by different taxes, however since combo items usually don't cost anything and are not under any tax group, the quotation preview will try show the section's total prices under no tax which will likely amount to 0$ This results in a bloated preview.

Solution:
Only accept a grouping under a specific tax (be it no tax or a real tax) if the total price != 0$

**Steps to reproduce:**
- Create a combo product containing a product that is taxed
- Create a quote with a section
- Add the product under the section
- Check 'Hide Composition' in the section's options
- Preview the sale
- You'll notice the section duplicated with no tax and no price

opw-5481931

### DIFF
--- a/addons/sale/models/sale_order_line.py
+++ b/addons/sale/models/sale_order_line.py
@@ -1549,7 +1549,11 @@ class SaleOrderLine(models.Model):
         """
         self.ensure_one()
 
-        section_lines = self.order_id.order_line.filtered(self._is_line_in_section)
+        billable_lines = self.order_id.order_line.filtered(
+            lambda line:
+                line.product_type != 'combo'
+                and self._is_line_in_section(line)
+        )
 
         if display_taxes:
             res = [
@@ -1558,13 +1562,13 @@ class SaleOrderLine(models.Model):
                     'price_subtotal': sum(lines.mapped('price_subtotal')),
                     'price_total': sum(lines.mapped('price_total')),
                 }
-                for taxes, lines in section_lines.grouped('tax_ids').items()
+                for taxes, lines in billable_lines.grouped('tax_ids').items()
             ]
         else:
             res = [{
                 'tax_labels': [],
-                'price_subtotal': sum(section_lines.mapped('price_subtotal')),
-                'price_total': sum(section_lines.mapped('price_total')),
+                'price_subtotal': sum(billable_lines.mapped('price_subtotal')),
+                'price_total': sum(billable_lines.mapped('price_total')),
             }]
         return res or [{
             'tax_labels': [],
--- a/addons/sale/tests/test_sale_sections.py
+++ b/addons/sale/tests/test_sale_sections.py
@@ -121,6 +121,55 @@ class TestSaleSections(SaleCommon):
         self.assertEqual(subsection_summary_lines[0]['price_subtotal'], 600.00)
         self.assertEqual(subsection_summary_lines[1]['price_subtotal'], 100.00)
 
+    def test_sale_order_report_line_visibility_and_grouping_with_combo(self):
+        """
+        Verify that collapsed sections correctly aggregate on tax groups without splitting lines
+        for untaxed zero-price items such as combo items.
+        """
+        product_a = self._create_product(name="Beefy burger")
+        product_b = self._create_product(name="Belgian fries")
+        combos = self.env['product.combo'].create([{
+            'name': "Burger",
+            'combo_item_ids': [Command.create({'product_id': product_a.id})],
+        }, {
+            'name': "Side",
+            'combo_item_ids': [Command.create({'product_id': product_b.id})],
+        }])
+        product_combo = self._create_product(
+            name="Meal Menu",
+            list_price=10.0,
+            type='combo',
+            combo_ids=[Command.set(combos.ids)],
+        )
+        self.sections_sale_order.order_line = [
+            Command.clear(),
+            Command.create({
+                'name': 'Sec1',
+                'display_type': 'line_section',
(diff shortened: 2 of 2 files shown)


addons/sale/models/sale_order_line.py
```diff
@@ -1548,7 +1548,7 @@ def _get_grouped_section_summary(self, display_taxes=True):
                     'price_subtotal': sum(lines.mapped('price_subtotal')),
                     'price_total': sum(lines.mapped('price_total')),
                 }
-                for taxes, lines in section_lines.grouped('tax_ids').items()
+                for taxes, lines in section_lines.grouped('tax_ids').items() if sum(lines.mapped('price_total')) != 0.0
```
barracudapps: Avoid long lines for readability
```suggestion
                for taxes, lines in section_lines.grouped('tax_ids').items()
                if sum(lines.mapped('price_total')) != 0.0
```

addons/sale/tests/test_sale_sections.py
```diff
@@ -121,6 +121,63 @@ def test_sale_order_report_line_visibility_and_grouping(self):
...
+        self.sections_sale_order.order_line = [
+            Command.clear(),
+            Command.create({
+                'name': 'Sec1',
+                'display_type': 'line_section',
+                'collapse_composition': True,
+            }),
+            Command.create({
+                'product_id': product_combo.id,
+            }),
```
barracudapps: Readable on a single line
```suggestion
            Command.create({'product_id': product_combo.id}),
```

addons/sale/tests/test_sale_sections.py
```diff
@@ -121,6 +121,63 @@ def test_sale_order_report_line_visibility_and_grouping(self):
...
+                'linked_line_id': combo_line.id,
+            }),
+        ]
+
+        lines_to_report = self.sections_sale_order._get_order_lines_to_report()
+        self.assertEqual(len(lines_to_report), 1)
+        self.assertEqual(
+            lines_to_report.mapped('name'),
+            ['Sec1'],
+        )
```
barracudapps: I would remove this as already tested in `test_sale_order_report_line_visibility_and_grouping`

addons/sale/tests/test_sale_sections.py
```diff
@@ -121,6 +121,63 @@ def test_sale_order_report_line_visibility_and_grouping(self):
...
+
+        subsection_summary_lines = lines_to_report[0]._get_grouped_section_summary()
+        self.assertEqual(len(subsection_summary_lines), 1)
+
```
barracudapps:
```suggestion

```

barracudapps (approved):
Some nits but LGTM 👍 
Thank you for fixing this!
You can mark as ready once the comments applied

addons/sale/models/sale_order_line.py
```diff
@@ -1549,6 +1549,7 @@ def _get_grouped_section_summary(self, display_taxes=True):
                     'price_total': sum(lines.mapped('price_total')),
                 }
                 for taxes, lines in section_lines.grouped('tax_ids').items()
+                if sum(lines.mapped('price_total')) != 0.0
```
nipl-odoo: IMO not a good idea to always hide tax group with 0 price_total, we need lines with 0 `price_total` and different tax in a separate group. You should instead filter such lines in `section_lines` Maybe you can play with `_is_line_in_section` method and which directly skips `is_indirect_child` or `is_direct_child` checks if line.product_type == 'combo'

nipl-odoo (changes requested):
Discussed with Jean and looks like this solution ain't gonna cut it 👀

addons/sale/models/sale_order_line.py
nsirjacobs: Thank you for your proposition, it makes sense !
Is there any way that `_is_line_in_section()` would need to return the combo line if called in another context?
Currently it doesn't seem to be the case as the function is only called in functions that compute the sum of prices, or check for taxes which are not affected by combo lines:
https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/sale/models/sale_order_line.py#L1603-L1605
https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/sale/models/sale_order_line.py#L1581-L1585


If yes maybe I could instead remove combo lines among the returned lines after the function is called ?

nipl-odoo: Not an issue in our standard code, but if someone did some customization based on name of this method, it could be problematic since the name of function `_is_line_in_section` suggests to check whether or not a line is under (self) section or not 🤔 In that case maybe you could modify `section_lines` filter like
```py
groupable_section_lines = self.order_id.order_line.filtered(
  lambda line:
      line.product_type != 'combo'
      and self._is_line_in_section(line)
)
```
(Bad variable name I know maybe you can suggest something better)

But still I need @Feyensv's opinion on this, wdyt?

Feyensv: Sorry for the late reply here, I agree it kinda changes the meaning of `_is_line_in_section` to handle it directly inside the method.
It would indeed be safer to handle it directly inside `_get_grouped_section_summary`.

nipl-odoo (approved):
Thank you for the fix👍🏻

Feyensv (approved):
robodoo delegate+

Thanks :pray:

PR #245868, opened 2026-01-27
[FIX] web: introduce folded column mobile layout

ldau-odoo:
Before this PR, when a user pinned the browser window to one side (using another app side by side), any folded column would still be displayed empty and with a large size, even though the folding functionality is not available on mobile.

With this update, the folded column layout is fixed on mobile and the unfold button is now hidden, as this functionality is not available.

task-5391942

ent-PR: https://github.com/odoo/enterprise/pull/107144

| Before | After |
|--------|--------|
| Pipeline-02-11-2026_01_20_PM | Pipeline-02-11-2026_02_41_PM |

---

### DIFF
--- a/addons/web/static/src/views/kanban/kanban_header.xml
+++ b/addons/web/static/src/views/kanban/kanban_header.xml
@@ -4,7 +4,7 @@
     <t t-name="web.KanbanHeader">
         <div class="o_kanban_header position-sticky top-0 z-1" t-ref="root" t-attf-class="{{ !env.isSmall and group.isFolded ? 'd-print-none pt-2' : 'py-2 pt-print-0' }}">
             <div class="o_kanban_header_title position-relative d-flex lh-lg">
-                <div t-if="group.isFolded" class="o_column_title d-flex align-items-center pt-1 fs-4 lh-1 text-nowrap opacity-50 opacity-100-hover flex-grow-1 gap-1"
+                <div t-if="group.isFolded" class="o_column_title d-flex align-items-center pt-1 fs-4 lh-1 text-nowrap opacity-50 flex-grow-1 gap-1"
                      t-on-mouseenter="onTitleMouseEnter" t-on-mouseleave="onTitleMouseLeave">
                     <t t-esc="group.displayName"></t>
                     <span t-if="group.count > 0 and !props.list.model.useSampleModel" t-esc="'(' + group.count + ')'"/>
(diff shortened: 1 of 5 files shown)


addons/web/static/src/views/kanban/kanban_renderer.xml
```diff
@@ -52,7 +52,7 @@
                                         <i class="fa fa-plus fa-stack d-block small text-end text-700 opacity-50 opacity-100-hover cursor-pointer" role="img" aria-label="Quick add" title="Quick add" />
                                     </button>
                                 </t>
-                                <button t-else="" class="o_column_unfold btn btn-link d-block mt-0 mt-md-2 p-0 border-0 small text-center text-700 cursor-pointer">
+                                <button t-else="" class="o_column_unfold d-none d-md-block btn btn-link d-block mt-0 mt-md-2 p-0 border-0 small text-center text-700 cursor-pointer">
```
chgo-odoo:
```suggestion
                                <button t-else="" class="o_column_unfold btn btn-link d-none d-md-block mt-0 mt-md-2 p-0 border-0 small text-center text-700 cursor-pointer">
```

chgo-odoo (approved):
LGTM @ldau-odoo, thanks for your work !

chgo-odoo (commented):
Should probably be a [FIX] instead of an IMP as we intentionally prevented the folded column to exist on Mobile but in the end we're fixing this because it creates a flaw so I'd not use [IMP] here

addons/web/static/src/views/kanban/kanban_header.xml
```diff
@@ -2,9 +2,10 @@
 <templates xml:space="preserve">
 
     <t t-name="web.KanbanHeader">
-        <div class="o_kanban_header position-sticky top-0 z-1" t-ref="root" t-attf-class="{{ !env.isSmall and group.isFolded ? 'd-print-none pt-2' : 'py-2 pt-print-0' }}">
+        <div class="o_kanban_header position-sticky top-0 z-1" t-ref="root" t-attf-class="{{ !this.ui.isSmall and group.isFolded ? 'd-print-none pt-2' : 'py-2 pt-print-0' }}">
```
aab-odoo: as far as I understand, env.isSmall is just an indirection to ui.isSmall, so changing this should have no effect. Am I missing something?

ldau-odoo: The env.isSmall does not react to window resizing, unlike the other one.

aab-odoo: Ho indeed, right. Unfortunately that change isn't valid in stable versions. As you can see with the enterprise PR it can break stuff.

ldau-odoo: Hello @aab-odoo,
I tried a different approach, it should be OK now

aab-odoo: Hello @ldau-odoo 

Scrollbars are displayed on my laptop (brave) 
image

PR #245869, opened 2026-01-27
Master enhance partner page dtda

dtda-odoo:
# Enhance Partners Page Customisation and Map Configuration

This PR improves the partner/reseller pages across `website_partnership` and `website_crm_partner_assign` by enhancing map configuration, expanding card customisation options, clarifying visibility controls, and simplifying template rendering order.

### Map Builder Improvements

* Display the **Map View** option by default, even when no Google Maps API key is configured.
* Add a conditional API key input that becomes visible when the map option is enabled.

### Partner Card Design Options

* Add a new option to enable or disable **card shadows** on partner/reseller cards.
* Add a **border radius slider** to customize card roundness.

### Contact Details Visibility Refactor

* Split the previous **Contact Details** option into two distinct controls:

  * **State & Country**: Controls visibility of state and country information on the `/partners` listing page.
  * **Contact Details**: Controls visibility of address/contact information on individual `/partner/<id>` pages.
* Fix the issue where disabling contact details on the listing page unintentionally hid address information on partner detail pages.
* Remove the unused `partner_page` template from `website_partnership`, as it was never rendered.

### Template Rendering Order Improvements

* Replace the class-based inherited view positioning mechanism with a **priority-based rendering order**.
* Remove the obsolete `top_section_button_middle_pos` placeholder element.
* Add explicit view priorities for inherited content rendered within the `page_top_section`.

TASK - 5156311
---

### DIFF
--- a/addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
+++ b/addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
@@ -1,12 +1,13 @@
 import { useState } from "@web/owl2/utils";
 import { BaseOptionComponent } from "@html_builder/core/base_option_component";
-import { onWillStart } from "@odoo/owl";
+import { onMounted } from "@odoo/owl";
 import { useService } from "@web/core/utils/hooks";
 import { registry } from "@web/core/registry";
 
 export class WebsiteCRMPartnersPageOption extends BaseOptionComponent {
     static id = "website_crm_partners_page_option"
     static template = "website_crm_partner_assign.PartnersPageOption";
+    static hasApiKeyCache = null;
 
     setup() {
         super.setup();
@@ -15,8 +16,18 @@ export class WebsiteCRMPartnersPageOption extends BaseOptionComponent {
             has_google_maps_api_key: false,
         });
 
-        onWillStart(async () => {
-            this.state.has_google_maps_api_key = !!(await this.googleMaps.getGMapsAPIKey(false));
+        onMounted(async () => {
+            if (WebsiteCRMPartnersPageOption.hasApiKeyCache !== null) {
+                this.state.has_google_maps_api_key = WebsiteCRMPartnersPageOption.hasApiKeyCache;
+                return;
+            }
+
+            const hasKey = !!(await this.googleMaps.getGMapsAPIKey(true));
+            this.state.has_google_maps_api_key = hasKey;
+
+            if (hasKey) {
+                WebsiteCRMPartnersPageOption.hasApiKeyCache = true;
+            }
         });
     }
 }
--- a/addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
+++ b/addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
@@ -9,8 +9,13 @@
         </xpath>
     </t>
     <t t-name="website_crm_partner_assign.PartnersPageOption">
-        <BuilderRow t-if="this.state.has_google_maps_api_key" label.translate="World Map">
-            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
+        <BuilderRow label.translate="World Map">
+            <BuilderCheckbox id="'partners_world_map_opt'" action="'websiteConfig'"
+                actionParam="{views: ['website_crm_partner_assign.ref_country']}"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Google Map" level="1"
+            t-if="isActiveItem('partners_world_map_opt') and !this.state.has_google_maps_api_key">
+            <BuilderButton title.translate="Custom Key" action="'configureApiKey'" preview="false">Custom Key</BuilderButton>
         </BuilderRow>
     </t>
 
--- a/addons/website_crm_partner_assign/views/website_crm_partner_assign_templates.xml
+++ b/addons/website_crm_partner_assign/views/website_crm_partner_assign_templates.xml
@@ -140,32 +140,30 @@
     </xpath>
 </template>
 
-<template id="ref_country" inherit_id="website_partnership.search_setting" name="World Map">
-    <xpath expr="//div[hasclass('o_wcrm_search')]" position="inside">
-        <t t-if="google_maps_api_key">
-            <!-- modal for large map -->
-            <div role="dialog" class="modal fade partner_map_modal" tabindex="-1">
-              <div class="modal-dialog modal-lg">
-                <div class="modal-content">
-                    <header class="modal-header">
-                        <h4 class="modal-title">World Map</h4>
-                        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"/>
-                    </header>
-                    <iframe loading="lazy" t-attf-src="/google_map?height=485&amp;dom=website_crm_partner_assign.partners&amp;current_grade=#{ current_grade and current_grade.id }&amp;current_country=#{ current_country and current_country.id }&amp;partner_url=/partners/&amp;limit=1000"
-                    style="height:485px;"/>
-                </div>
-              </div>
+<template id="ref_country" inherit_id="website_crm_partner_assign.index_content" name="World Map" active="False" priority="20">
+    <xpath expr="//section[@id='page_top_section']" position="inside">
+        <!-- modal for large map -->
+        <div role="dialog" class="modal fade partner_map_modal">
+            <div class="modal-dialog modal-lg">
+            <div class="modal-content">
+                <header class="modal-header">
+                    <h4 class="modal-title">World Map</h4>
+                    <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"/>
+                </header>
+                <iframe loading="lazy" t-attf-src="/google_map?height=485&amp;dom=website_crm_partner_assign.partners&amp;current_grade=#{ current_grade and current_grade.id }&amp;current_country=#{ current_country and current_country.id }&amp;partner_url=/partners/&amp;limit=1000"
+                style="height:485px;"/>
             </div>
-            <!-- modal end -->
-            <div class="btn-group ms-2">
-                <button class="btn btn-light border-primary active">
-                    <i class="fa fa-th-large"/>
-                </button>
-                <button class="btn btn-light" data-bs-toggle="modal" data-bs-target=".partner_map_modal">
-                    <i class="fa fa-map-marker" role="img" aria-label="Open map" title="Open map"/>
-                </button>
             </div>
-        </t>
+        </div>
+        <!-- modal end -->
+        <div class="btn-group ms-2">
+            <button class="btn btn-light border-primary active">
+                <i class="fa fa-th-large"/>
+            </button>
+            <button class="btn btn-light" data-bs-toggle="modal" data-bs-target=".partner_map_modal">
+                <i class="fa fa-map-marker" role="img" aria-label="Open map" title="Open map"/>
+            </button>
+        </div>
     </xpath>
 </template>
 
--- a/addons/website_partnership/static/src/scss/primary_variable.scss
+++ b/addons/website_partnership/static/src/scss/primary_variable.scss
@@ -1,0 +1,6 @@
+$o-base-website-values-palette: map-merge(
+    $o-base-website-values-palette,
+    (
+        'website_partnership_card_roundness': 10% / 5%,
+    )
+);
--- a/addons/website_partnership/static/src/scss/website_partnership.scss
+++ b/addons/website_partnership/static/src/scss/website_partnership.scss
@@ -1,0 +1,5 @@
+$radius: o-website-value('website_partnership_card_roundness');
+
+.o_wpartners_card {
+  border-radius: #{$radius * 1%} / #{$radius * 0.5%};
+}
--- a/addons/website_partnership/static/src/website_builder/website_partnership_option.xml
+++ b/addons/website_partnership/static/src/website_builder/website_partnership_option.xml
@@ -23,6 +23,16 @@
         <BuilderRow label.translate="Categories" id="'categories_option'">
             <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_partnership.categories_setting']}" />
         </BuilderRow>
+        <hr/>
+        <BuilderRow label.translate="Cards Design">
+            Global options for all inner cards.
+        </BuilderRow>
+        <BuilderRow label.translate="Roundness" tooltip.translate="Applies to all cards" level="1">
+            <BuilderRange action="'customizeWebsiteVariable'" actionParam="'website_partnership_card_roundness'" default="10" max="50"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Shadows" tooltip.translate="Applies to all cards" level="1">
+            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_partnership.partners_card_shadow']}" />
+        </BuilderRow>
     </t>
 
     <t t-inherit="website.BuilderOptions" t-inherit-mode="extension">
--- a/addons/website_partnership/views/website_partnership_templates.xml
+++ b/addons/website_partnership/views/website_partnership_templates.xml
@@ -141,7 +141,7 @@
     </xpath>
 </template>
 
-<template id="search_setting" inherit_id="website_partnership.index" name="Search">
+<template id="search_setting" inherit_id="website_partnership.index" name="Search" priority="10">
     <xpath expr="//section[@id='page_top_section']" position="inside">
         <div class="o_wcrm_search d-flex w-100 w-lg-25">
             <form class="flex-grow-1" action="" method="get">
--- a/addons/website_partnership/views/website_partnership_templates.xml
+++ b/addons/website_partnership/views/website_partnership_templates.xml
@@ -81,10 +81,11 @@
                     <span t-out="partner.grade_id.name"/>
                     <t t-call="website.publish_management" object="partner.grade_id" action="'partnership.res_partner_grade_action'" publish_edit="True"/>
                 </h5>
+                <t t-set="partner_card_shadow" t-value="False"/>
                 <t t-set="last_grade" t-value="partner.grade_id.id"/>
             </t>
             <div class="col-md-4 col-xl-3 col-12 mb-4">
-                <div class="card h-100 text-decoration-none">
+                <div t-attf-class="o_wpartners_card card h-100 text-decoration-none #{partner_card_shadow and 'shadow-lg'}">
                     <a id="partner_link_card" class="text-decoration-none" aria-label="Go to partner">
                         <div t-field="partner.avatar_1920"
                             class="card-img-top border-bottom"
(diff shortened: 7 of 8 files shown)


addons/website_partnership/static/src/scss/website_partnership.scss
```diff
@@ -0,0 +1,5 @@
+$radius: o-website-value('website_partnership_card_roundness');
+
+.o_wpartners_card {
+  border-radius: #{$radius * 1%} / #{$radius * 0.5%} !important;
```
divy-odoo: 1. I didn't get the reason behind such calculation, why not simple `$radius*1rem`?
2. Do not use !important if it's not really required.

addons/website_partnership/static/src/website_builder/website_partnership_option.xml
```diff
@@ -23,6 +23,16 @@
...
+        <hr/>
+        <BuilderRow label.translate="Cards Design">
+            Global options for all inner cards.
+        </BuilderRow>
+        <BuilderRow label.translate="Roundness" tooltip.translate="Applies to all cards" level="1">
+            <BuilderRange action="'customizeWebsiteVariable'" actionParam="'website_partnership_card_roundness'" default="10" max="50"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Shadows" tooltip.translate="Applies to all cards" level="1">
+            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_partnership.partners_card_shadow']}" />
+        </BuilderRow>
```
divy-odoo, line 35: I don’t like this static approach. What’s the point of only having an option to enable the shadow, without allowing custom configuration? We should use the `<ShadowOption/>` component here to display shadow options, just like we do for other snippets.

Image

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
```diff
@@ -9,9 +9,17 @@
         </xpath>
     </t>
     <t t-name="website_crm_partner_assign.PartnersPageOption">
-        <BuilderRow t-if="this.state.has_google_maps_api_key" label.translate="World Map">
-            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
+        <BuilderRow label.translate="World Map">
+            <BuilderCheckbox id="'partners_world_map'" action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
```
divy-odoo:
```suggestion
            <BuilderCheckbox id="'partners_world_map_opt'" action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
```

`opt` prefix will help to identify it as builder option ( it's not required)

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
```diff
@@ -9,9 +9,17 @@
...
         </BuilderRow>
+        <BuilderContext t-if="isActiveItem('partners_world_map') and !this.state.has_google_maps_api_key">
```
divy-odoo: 1. `!this.state.has_google_maps_api_key` isn’t working as expected. The option is still displayed even though I have configured the API key.
2. Isn't this option should be nested to the `partners_world_map` ?

addons/website_partnership/views/website_partnership_templates.xml
```diff
@@ -141,7 +142,7 @@
     </xpath>
 </template>
 
-<template id="search_setting" inherit_id="website_partnership.index" name="Search">
+<template id="search_setting" inherit_id="website_partnership.index" name="Search" priority="10">
```
divy-odoo: Why do you need priority here?

divy-odoo (commented):
Hello @dtda-odoo, Here is my quick review. 
Also, please adapt the commit message as per the guideline. 

1. ~[IMP] *= improve map builder option in partners page~ = > `[IMP] website_*: improve map builder option in partners page`
2. ~[IMP] website_partnership= add card design options for partner's cards~ =>   `[IMP] website_partnership: add card design options for partner's cards`

addons/website_partnership/views/website_partnership_templates.xml
dtda-odoo: To maintain the order of components on the website. Currently, both have default priority, so whichever is toggled first append first.

addons/website_partnership/static/src/scss/website_partnership.scss
dtda-odoo: 1. Since size of card is changing based on `Image Format` option and can change in future, we should not use static variable as `rem` which depends in font size. 
Different card size will have different max values for border-radius and will affect differently for each card.
Image


Now using `$radius * 1%` result to non uniform borders which also should not be implemented. 
This way using ellipse radius makes more sense to me.

Case with `$radius * 1%`:
Image

2. 👍🏻

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
```diff
@@ -1,12 +1,13 @@
...
 import { BaseOptionComponent } from "@html_builder/core/base_option_component";
-import { onWillStart } from "@odoo/owl";
+import { onMounted } from "@odoo/owl";
 import { useService } from "@web/core/utils/hooks";
 import { registry } from "@web/core/registry";
 
 export class WebsiteCRMPartnersPageOption extends BaseOptionComponent {
     static id = "website_crm_partners_page_option"
     static template = "website_crm_partner_assign.PartnersPageOption";
+    static hasApiKeyCache = null;
```
divy-odoo: Why this static variable caching? In most of the time, component will get destroy so static cache will get lost. isn't the state works?

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
```diff
@@ -15,8 +16,18 @@ export class WebsiteCRMPartnersPageOption extends BaseOptionComponent {
...
 
-        onWillStart(async () => {
-            this.state.has_google_maps_api_key = !!(await this.googleMaps.getGMapsAPIKey(false));
+        onMounted(async () => {
+            if (WebsiteCRMPartnersPageOption.hasApiKeyCache !== null) {
+                this.state.has_google_maps_api_key = WebsiteCRMPartnersPageOption.hasApiKeyCache;
+                return;
+            }
+
+            const hasKey = !!(await this.googleMaps.getGMapsAPIKey(true));
```
divy-odoo: This will trigger an API call even if the `World Map` option is not enabled. Can we make the call only when `World Map` is enabled? We only need the `api_key` when that option is enabled.

addons/website_partnership/views/website_partnership_templates.xml
```diff
@@ -229,5 +230,11 @@
         <attribute name="t-options">{"widget": "image", "qweb_img_responsive": False, "class": "img img-fluid h-100 w-100 mw-100", "style": "max-height: 375px; min-height: 375px; object-fit: cover"}</attribute>
     </xpath>
 </template>
+<template name="Partner Card Shadow" id="website_partnership.partners_card_shadow"
+    inherit_id="website_partnership.index">
```
divy-odoo: Neatpick: Lint this in one line!

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
```diff
@@ -9,8 +9,13 @@
         </xpath>
     </t>
     <t t-name="website_crm_partner_assign.PartnersPageOption">
-        <BuilderRow t-if="this.state.has_google_maps_api_key" label.translate="World Map">
-            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
+        <BuilderRow label.translate="World Map">
+            <BuilderCheckbox id="'partners_world_map_opt'" action="'websiteConfig'"
+                actionParam="{views: ['website_crm_partner_assign.ref_country']}"/>
```
divy-odoo:
```suggestion
            <BuilderCheckbox id="'partners_world_map_opt'" action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}"/>
```
Neatpick:

addons/website_partnership/views/website_partnership_templates.xml
```diff
@@ -81,10 +81,11 @@
                     <span t-out="partner.grade_id.name"/>
                     <t t-call="website.publish_management" object="partner.grade_id" action="'partnership.res_partner_grade_action'" publish_edit="True"/>
                 </h5>
+                <t t-set="partner_card_shadow" t-value="False"/>
```
divy-odoo: Isn't this variable could be out side the parent/where it's being used to have a better scope? Even why do you need this while you can directly add the class using xPath?

addons/website_partnership/static/src/scss/primary_variable.scss
```diff
@@ -0,0 +1,6 @@
+$o-base-website-values-palette: map-merge(
+    $o-base-website-values-palette,
+    (
+        'website_partnership_card_roundness': 10% / 5%,
```
divy-odoo:
```suggestion
        'website-partnership-card-roundness': 10% / 5%,
```

addons/website_partner/static/src/website_builder/partner_page_option.xml
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
divy-odoo, line 1: why did we moved this file?

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
```diff
@@ -9,8 +9,13 @@
...
+        </BuilderRow>
+        <BuilderRow label.translate="Google Map" level="1"
+            t-if="isActiveItem('partners_world_map_opt') and !this.state.has_google_maps_api_key">
```
msh-odoo: Image
The issue occurs when clicking on Find a partner string

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
msh-odoo: Why this call in onMouted while onWillStart is the right hook for such calls, what is the benefit of doing all this, if you keep the old code then also call will be done each time and now also call is done each time.

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
dtda-odoo: When the component is destroyed, the instance of that class will be destroyed, not the class.
So anything under `new WebsiteCRMPartnersPageOption()` will be destroyed, the class will be present in JS runtime.

addons/website_partner/static/src/website_builder/partner_page_option.xml
dtda-odoo: Template `partner_page` is present in `website_partner`, and `contact_details` is called inside it. So controlling the call from another module does not make sense. Also, the template can be accessed via the route `/partners/<id>`, which is declared in `website_partner`. This file moved the option to website_partner (after discussion with PO - MEDA )

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
divy-odoo: Agree!! But is it really a good idea to use static cache? When & How does this cache will get invalidate? caching with no invalidation rule make no sense to me,  Instead I prefer not to cache the value if we don't have any other way. wdyt @msh-odoo ?

addons/website_crm_partner_assign/views/website_crm_partner_assign_templates.xml
```diff
@@ -140,49 +140,41 @@
...
-                <div class="modal-content">
-                    <header class="modal-header">
-                        <h4 class="modal-title">World Map</h4>
-                        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"/>
-                    </header>
-                    <iframe loading="lazy" t-attf-src="/google_map?height=485&amp;dom=website_crm_partner_assign.partners&amp;current_grade=#{ current_grade and current_grade.id }&amp;current_country=#{ current_country and current_country.id }&amp;partner_url=/partners/&amp;limit=1000"
-                    style="height:485px;"/>
-                </div>
-              </div>
+<template id="ref_country" inherit_id="website_crm_partner_assign.index_content" name="World Map" active="False" priority="20">
```
msh-odoo: Why priority added, is it because we are changing inherit_id and by default priority is 16 for search_setting? if this is the case then why change the inherit_id, will this diff make difference(in terms of technical or functional)?

addons/website_partnership/static/src/scss/website_partnership.scss
```diff
@@ -0,0 +1,5 @@
+$radius: o-website-value('website-partnership-card-roundness');
+
+.o_wpartners_card {
+  border-radius: #{$radius * 1%} / #{$radius * 0.5%};
```
msh-odoo: 4 space indentation please

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
dtda-odoo: Why use onMounted? After mount we can be sure the options are present and the value of `if (!this.isActiveItem("partners_world_map_opt"))` will be correct in each reload.

addons/website_crm_partner_assign/views/website_crm_partner_assign_templates.xml
dtda-odoo: Yes, to maintain the order of components on the website. Currently, both `search_setting` and this template have default priority, so whichever is toggled first append first.

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
divy-odoo: this a good approach to have such async call in onMount? This introduces UI flickering. On a slow network, the option is initially displayed and then suddenly disappears once the API key check completes successfully. This can be confusing for users.

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.xml
```diff
@@ -9,8 +9,12 @@
         </xpath>
     </t>
     <t t-name="website_crm_partner_assign.PartnersPageOption">
-        <BuilderRow t-if="this.state.has_google_maps_api_key" label.translate="World Map">
-            <BuilderCheckbox action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}" />
+        <BuilderRow label.translate="World Map">
+            <BuilderCheckbox id="'partners_world_map_opt'" action="'websiteConfig'" actionParam="{views: ['website_crm_partner_assign.ref_country']}"/>
+        </BuilderRow>
+        <BuilderRow label.translate="Google Map" level="1"
+            t-if="this.isActiveItem('partners_world_map_opt') and !this.state.has_google_maps_api_key">
```
divy-odoo:
```suggestion
        <BuilderRow label.translate="Google Map" level="1" t-if="this.isActiveItem('partners_world_map_opt') and !this.state.has_google_maps_api_key">
```

Nitpick:

addons/website_crm_partner_assign/views/website_crm_partner_assign_templates.xml
divy-odoo: In that case, I wonder why `div` with `top_section_button_middle_pos` class exists. I think template priority is much better that  `div.top_section_button_middle_pos`.  Can we add one more commit for that & move this priority changes into it? you can take reference of `blog_post_sidebar`.

addons/website_partnership/views/website_partnership_templates.xml
```diff
@@ -229,5 +229,10 @@
         <attribute name="t-options">{"widget": "image", "qweb_img_responsive": False, "class": "img img-fluid h-100 w-100 mw-100", "style": "max-height: 375px; min-height: 375px; object-fit: cover"}</attribute>
     </xpath>
 </template>
+<template name="Partner Card Shadow" id="website_partnership.partners_card_shadow" inherit_id="website_partnership.index">
+    <div class="o_wpartners_card card h-100 text-decoration-none" position="attributes">
```
divy-odoo: I'm against this option but anyway, what if someone will remove one of this class from the main template?

addons/website_partnership/views/website_partnership_templates.xml
```diff
@@ -112,23 +112,8 @@
     </div>
 </template>
 
-<template id="partner_page" name="Partner Page">
```
divy-odoo, line 115: Why we removed/moved this? Can you please add a reason in commit message, may be add a PR that made this template useless. same for the `website_crm_partner_assign.contact_details`.

addons/website_partner/views/website_partner_templates.xml
```diff
@@ -48,4 +48,11 @@
         </div>
     </div>
 </template>
+
+<template id="contact_details" inherit_id="website_partner.partner_page" name="Contact Details">
+    <xpath expr="//t[@t-call='website_partner.partner_detail']" position="before">
+        <t t-set="contact_details" t-value="True"/>
+    </xpath>
```
divy-odoo:
```suggestion
        <t t-call="website_partner.partner_detail" position="attributes">
            <attribute name="contact_details">True</attribute>
        </t>
```

May be this? I just a guess, didn't tried.

addons/website_crm_partner_assign/static/src/website_builder/website_crm_partner_assign_option.js
```diff
@@ -1,22 +1,36 @@
...
+            }
 
-        onWillStart(async () => {
-            this.state.has_google_maps_api_key = !!(await this.googleMaps.getGMapsAPIKey(false));
+            if (WebsiteCRMPartnersPageOption.hasApiKeyCache === true) {
+                this.state.has_google_maps_api_key = true;
+                return;
+            }
+
+            const hasKey = !!(await this.googleMaps.getGMapsAPIKey(true));
```
msh-odoo: Well, I checked this in detail and this will save some rpc, just an idea, what if we use RPCCache? what if we just add cache: true while calling ```/website/google_maps_api_key```

addons/website/static/src/builder/plugins/options/google_maps_option/google_maps_option_plugin.js
```diff
@@ -196,6 +197,7 @@ export class GoogleMapsOptionPlugin extends Plugin {
                         });
                         this.shouldRefetchApiKey = false;
                         isInvalidated = true;
+                        this.googleMaps.invalidateGMapsAPIKeyProm();
```
divy-odoo, line 200: Good catch!! I don't get why we don't have the function to write a `googleMapKey` inside google_maps, that way we can have robust cache management & abstract inside a single service. Anyway this is fine to me for now.

addons/website_partnership/views/website_partnership_templates.xml
divy-odoo: Still I didn't found much information about why it become useless. PR/commit you mentioned is something when this template is introduced. You may also need to add this commit from where it became useless. https://github.com/odoo/odoo/commit/b512df5165292e3d47af79498d7d0863f161b533#diff-0dbf2f5a91200a57c38b3561e8842e10ce2c62386c41e72e06c3c0b4bde4c17c

PR #245870, opened 2026-01-27
[IMP] pos_self_order: add back button on the attribute configuration

lcantraine:
The goal of this pr was to always have a Back button visible on the combo product configuration.

I also change method name to be more readable.

task: 5877219

---


davidmonnom (commented):
Overall, it's good, but you can't rename methods in the stable version. Can you restore the names?

davidmonnom (approved):
@robodoo r+

PR #245874, opened 2026-01-27
[FIX] pos_loyalty: allow scan of customer not preloaded

Nokoz:
When scanning a loyalty card in POS, the customer should be automatically selected. However, this fails when the customer is not in the initial preloaded customer list due to the limited_customer_count setting.

Partial backport of: https://github.com/odoo/odoo/commit/e2843355898d4cce953fba35504170c10910fc35

Steps to reproduce:
-------------------
* Open POS with limited_customer_count set to a low value (e.g., 5)
* Create a loyalty card for a customer that won't be in the top 5 preloaded
* Scan the loyalty card barcode in POS
> Observation:
Customer is not selected, shows "Invalid coupon code" error

Why the fix:
------------
The current implementation only searches for loyalty cards in the local POS cache. When a loyalty card's partner_id is not resolved (because the partner wasn't preloaded), the code fails to set the customer.

This fix backports the 19.0 solution which:
1. Uses server-side lookup via get_loyalty_card_partner_by_code() to find the partner ID from the loyalty card code
2. Explicitly loads the partner on-demand if not in the local cache
3. Works regardless of whether the customer was in the initial preload

The test didn't spot that previously because the default limited_customer_count was big enough.

opw-5778328

### DIFF
--- a/addons/pos_loyalty/static/src/overrides/models/pos_store.js
+++ b/addons/pos_loyalty/static/src/overrides/models/pos_store.js
@@ -283,14 +283,35 @@ patch(PosStore.prototype, {
         const rule = this.models["loyalty.rule"].find((rule) => {
             return rule.mode === "with_code" && (rule.promo_barcode === code || rule.code === code);
         });
-        const loyaltyCard = this.models["loyalty.card"].find(
+        let loyaltyCard = this.models["loyalty.card"].find(
             (card) => card.code === code && card.program_id?.program_type === "loyalty"
         );
+        let partner = loyaltyCard?.partner_id;
+        if (!loyaltyCard) {
+            loyaltyCard = await this.data.searchRead(
+                "loyalty.card",
+                [
+                    ["code", "=", code],
+                    ["program_type", "=", "loyalty"],
+                ],
+                this.data.fields["loyalty.card"],
+                {
+                    limit: 1,
+                }
+            );
+            loyaltyCard = loyaltyCard.at(-1);
+            partner = loyaltyCard?.partner_id;
+        } else if (!partner && loyaltyCard.raw.partner_id) {
+            partner = await this.data.read("res.partner", [loyaltyCard.raw.partner_id]);
+            partner = partner.at(-1);
+        }
         let claimableRewards = null;
         let coupon = null;
         // If the code belongs to a loyalty card we just set the partner
-        if (loyaltyCard && loyaltyCard.partner_id) {
-            order.set_partner(loyaltyCard.partner_id);
+        if (partner) {
+            order.set_partner(partner);
+            this.updateRewards();
+            return true;
         } else if (rule) {
             const date_order = DateTime.fromSQL(order.date_order);
             if (
--- a/addons/pos_loyalty/tests/test_frontend.py
+++ b/addons/pos_loyalty/tests/test_frontend.py
@@ -2940,6 +2940,7 @@ class TestUi(TestPointOfSaleHttpCommon):
         )
 
     def test_scan_loyalty_card_select_customer(self):
+        self.env['ir.config_parameter'].sudo().set_param('point_of_sale.limited_customer_count', 0)
         self.env['loyalty.program'].search([]).write({'active': False})
         self.test_partner = self.env['res.partner'].create({'name': 'A Test Partner'})
 


addons/pos_loyalty/tests/test_frontend.py
```diff
@@ -2938,6 +2938,7 @@ def test_buy_x_get_y_reward_qty(self):
         )
 
     def test_scan_loyalty_card_select_customer(self):
+        self.env['ir.config_parameter'].sudo().set_param('point_of_sale.limited_customer_count', 5)
```
sbel-odoo:
```suggestion
        self.env['ir.config_parameter'].sudo().set_param('point_of_sale.limited_customer_count', 0)
```
Otherwise the test passes fine for me without your fix.

addons/pos_loyalty/static/src/overrides/models/pos_store.js
```diff
@@ -280,17 +280,23 @@ patch(PosStore.prototype, {
...
         const order = this.get_order();
-        const rule = this.models["loyalty.rule"].find((rule) => {
-            return rule.mode === "with_code" && (rule.promo_barcode === code || rule.code === code);
-        });
-        const loyaltyCard = this.models["loyalty.card"].find(
-            (card) => card.code === code && card.program_id?.program_type === "loyalty"
+        const rule = this.models["loyalty.rule"].find(
+            (rule) =>
+                rule.mode === "with_code" && (rule.promo_barcode === code || rule.code === code)
         );
```
sbel-odoo:
```js
const rule = this.models["loyalty.rule"].find(
            (rule) =>
                rule.mode === "with_code" && (rule.promo_barcode === code || rule.code === code)
        );
```
and 
```js
const rule = this.models["loyalty.rule"].find((rule) => {
            return rule.mode === "with_code" && (rule.promo_barcode === code || rule.code === code);
        });
```
They do the exact same thing so it's better to avoid unnecessary diff and keep the previous version.

sbel-odoo (commented):
Left a few comments. otherwise looks good to me 👍 

Just fyi, it's best to put the link of the commit that you're backporting instead of the PR. With the PR link each time you push your commit it will put a notification on the PR, same with the forward ports commit when it applies. Commit link doesn't do that.

Here's the commit link: https://github.com/odoo/odoo/commit/e2843355898d4cce953fba35504170c10910fc35

addons/pos_loyalty/static/src/overrides/models/pos_store.js
Nokoz: I'm still fighting with the linter...

addons/pos_loyalty/static/src/overrides/models/pos_store.js
```diff
@@ -283,14 +283,19 @@ patch(PosStore.prototype, {
...
         // If the code belongs to a loyalty card we just set the partner
-        if (loyaltyCard && loyaltyCard.partner_id) {
-            order.set_partner(loyaltyCard.partner_id);
+        if (partnerId) {
+            if (!this.models["res.partner"].get(partnerId)) {
+                await this.data.read("res.partner", [partnerId]);
+            }
+            const partner = this.models["res.partner"].get(partnerId);
+            order.set_partner(partner);
+            await this.updateRewards();
```
adgu-odoo:
```suggestion
        let loyaltyCard = this.models["loyalty.card"].find(
            (card) => card.code === code && card.program_id?.program_type === "loyalty"
        );
        let partner = loyaltyCard?.partner_id;
        if (!loyaltyCard) {
            loyaltyCard = await this.data.searchRead(
                "loyalty.card",
                [
                    ["code", "=", code],
                    ["program_type", "=", "loyalty"],
                ],
                this.data.fields["loyalty.card"],
                {
                    limit: 1,
                }
            );
            loyaltyCard = loyaltyCard.at(-1);
            partner = loyaltyCard?.partner_id;
        } else if (!partner && loyaltyCard.raw.partner_id) {
            partner = await this.data.read("res.partner", [loyaltyCard.raw.partner_id]);
            partner = partner.at(-1);
        }
        let claimableRewards = null;
        let coupon = null;

        // If the code belongs to a loyalty card we just set the partner
        if (partner) {
            order.set_partner(partner);
            this.updateRewards();
            return true;
```
What do you think of this. No new backend method is needed. Either we do not have the loyaltyCard and we use `searchRead` with the barcode. It will load the loyalty card record and integrate it automatically to the related models service. If the partner is not loaded at the time of the `searchRead`, it will load it automatically too. In the another scenario where we already have the loyalty card loaded, we check if we have the partner loaded. If not, we do a `read` on the partner with its id using `loyaltyCard.raw.partner_id`

Nokoz: Yep it also works and isolate the fix!

adgu-odoo (commented):
robodoo r+

PR #245875, opened 2026-01-27
[FIX] iot_drivers: rename branch only when upgrade succeeds

loouis-t:
If the upgrade fails, the branch name can already be set to the target one. Making the IoT Box not upgrading on next start.

We now check `git` method return value, to ensure the command succeeded, before renaming the branch.

### DIFF
--- a/addons/iot_drivers/tools/upgrade.py
+++ b/addons/iot_drivers/tools/upgrade.py
@@ -153,37 +153,35 @@ def check_git_branch(server_url=None):
         _logger.exception('An error occurred while trying to update the code with git')
 
 
-def _ensure_production_remote(local_remote):
+def _ensure_production_remote():
     """Ensure that the remote repository is the production one
     (https://github.com/odoo/odoo.git).
-
-    :param local_remote: The name of the remote repository.
     """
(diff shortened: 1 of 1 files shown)


addons/iot_drivers/tools/upgrade.py
```diff
@@ -153,37 +153,33 @@ def check_git_branch(server_url=None):
...
 
-    _logger.warning("Checking out %s/%s", remote, branch)
-    git('remote', 'set-branches', remote, branch)
-    git('fetch', remote, branch, '--depth=1', '--prune')  # refs/remotes to avoid 'unknown revision'
-    git('reset', 'FETCH_HEAD', '--hard')
+    _logger.warning("Checking out origin/%s", branch)
+    if git("fetch", "origin", branch, "--depth=1", "--prune") is None:
+        return
+    if git("reset", "FETCH_HEAD", "--hard") is None:
+        return
```
MaxTheWhale, line 180: We could add a logger.error in these cases so we know which command failed

MaxTheWhale (approved):
Tested and checking out still works as expected when it succeeds, and the code for the other cases looks good :white_check_mark:

PR #245877, opened 2026-01-27
[FIX] *: round down amount to currency's minor unit in validate amount

mma1377:
*: payment{,_adyen,_mercado_pago,_stripe}

### Issue:
The payment in some payment providers are failing to due setting
currency decimals.

#### Steps to reproduce:
1- In USD currency form in developer mode, set the rounding factor
   to `0.001000`, so we have 3 decimal places.
2- From `Decimal Accuracy` set Price Unit to 4 digits.
3- Refresh the page.
4- Create a SO and add a line. Set the price to `200.7647`.
5- Refresh the page again and generate a payment link.
6- Pay through Adyen. As you see payments fail.

### Cause and Fix:
This issue is the same as https://github.com/odoo/odoo/pull/245059. That fix only focused on Stripe, however, the same issue is reproducible on other providers e.g. `Adyen`. In order to avoid the same issue, we can have a general fix to round down all payment transaction amounts to currency's minor unit in `_validate_amount_and_currency`.

---

There is another issue in Mercado Pago which the if currency is not one of these currencies then it will not be sent to provider rounded:
https://github.com/odoo/odoo/blob/035b701cbf7bd654f56f07244e606e4038c4115a/addons/payment_mercado_pago/const.py#L39-L43

Mercado Pago rounds **up** in these cases, which will cause mismatch in `_validate_amount_and_currency`.

opw-5871470

### DIFF
--- a/addons/payment/tests/test_payment_transaction.py
+++ b/addons/payment/tests/test_payment_transaction.py
@@ -294,3 +294,10 @@ class TestPaymentTransaction(PaymentCommon):
             'redirect', operation='validation', amount=0, reference='tx-validation',
         )
         validation_tx._validate_amount_and_currency(None, None)
+
+    def test_validate_amount_uses_payment_minor_unit(self):
+        self.currency_euro.rounding = 0.001
+        tx = self._create_transaction('direct', amount=123.452, currency_id=self.currency_euro.id)
+        self._assert_does_not_raise(
+            ValidationError, tx._validate_amount_and_currency, 123.45, self.currency_euro.name
+        )
--- a/addons/payment_adyen/models/payment_transaction.py
+++ b/addons/payment_adyen/models/payment_transaction.py
@@ -404,7 +404,11 @@ class PaymentTransaction(models.Model):
             arbitrary_decimal_number=const.CURRENCY_DECIMALS.get(self.currency_id.name),
         )
         currency_code = amount_data.get('currency')
-        self._validate_amount_and_currency(amount, currency_code)
+        self._validate_amount_and_currency(
+            amount,
+            currency_code,
+            precision_digits=const.CURRENCY_DECIMALS.get(self.currency_id.name),
+        )
 
     def _process_notification_data(self, notification_data):
         """ Override of payment to process the transaction based on Adyen data.
--- a/addons/payment_stripe/tests/test_stripe.py
+++ b/addons/payment_stripe/tests/test_stripe.py
@@ -202,21 +202,6 @@ class StripeTest(StripeCommon, PaymentHttpCommon):
             for payload_param in ('account', 'return_url', 'refresh_url', 'type'):
                 self.assertIn(payload_param, call_args)
 
-    def test_compare_notification_data_uses_payment_minor_unit(self):
-        """
-        Test that the payment's minor unit precision is used to validate the amount, not the custom
-        currency rounding that customers may have configured.
-        """
-        self.amount = 20076
-        self.currency.rounding = 0.001
-        tx = self._create_transaction(
-            'dummy', operation='online_direct', tokenize=True, amount=200.769
-        )
-        data = self.notification_data['data']
-        data['payment_intent'] = self._mock_payment_intent_request()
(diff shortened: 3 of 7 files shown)


vchu-odoo (commented):
After the discussion with @AntoineVDV it is concluded that for stable it is better to do less risky fix by applying the precision digits to every bugged provider in their override of `compare_notification_data` as we did for Stripe.

For master we created the technical task to round upon transaction creation.

mma1377:
@AntoineVDV Sorry, I now understood what you meant. Okay, I'll be on it.

vchu-odoo (commented):
Thank you for the changes!

Here are some suggestions to include in the commit:
- We can remove the previous fix from Stripe and pass only special currency precision digits in `_compare_notification_data` 
- in Adyen we need to pass the special currency precision digits in `_compare_notification_data` as well.
- `test_compare_notification_data_uses_payment_minor_unit` in Stripe can be removed since we moved the logic to `payment`

addons/payment_stripe/tests/test_stripe.py
```diff
@@ -202,21 +202,6 @@ def test_create_account_link_pass_required_parameters(self):
...
-        Test that the payment's minor unit precision is used to validate the amount, not the custom
-        currency rounding that customers may have configured.
-        """
-        self.amount = 20076
-        self.currency.rounding = 0.001
-        tx = self._create_transaction(
-            'dummy', operation='online_direct', tokenize=True, amount=200.769
-        )
-        data = self.notification_data['data']
-        data['payment_intent'] = self._mock_payment_intent_request()
```
vchu-odoo, line 216: Could you please remove `_mock_payment_intent_request()` since it is not used anymore?

addons/payment_adyen/models/payment_transaction.py
```diff
@@ -404,7 +404,11 @@ def _compare_notification_data(self, notification_data):
             arbitrary_decimal_number=const.CURRENCY_DECIMALS.get(self.currency_id.name),
         )
         currency_code = amount_data.get('currency')
-        self._validate_amount_and_currency(amount, currency_code)
+        self._validate_amount_and_currency(
+            amount,
+            currency_code,
+            const.CURRENCY_DECIMALS.get(self.currency_id.name),
```
vchu-odoo: `precision_digits` is kwargs
```suggestion
            precision_digits=const.CURRENCY_DECIMALS.get(self.currency_id.name),
```

addons/payment/tests/test_payment_transaction.py
```diff
@@ -294,3 +294,15 @@ def test_validate_amount_and_currency(self):
             'redirect', operation='validation', amount=0, reference='tx-validation',
         )
         validation_tx._validate_amount_and_currency(None, None)
+
+    def test_validate_amount_uses_payment_minor_unit(self):
+        self.currency = self.env.ref('base.EUR')
+        self.currency.rounding = 0.001
+        self.amount = 123.452
+        tx = self._create_transaction('direct', state='done', reference='tx-direct')
```
vchu-odoo:
```suggestion
        self.currency_euro.rounding = 0.001
        tx = self._create_transaction('direct', amount=123.452, currency_id=self.currency_euro.id)
```

addons/payment/tests/test_payment_transaction.py
```diff
@@ -294,3 +294,15 @@ def test_validate_amount_and_currency(self):
...
+        self._assert_does_not_raise(
+            ValidationError,
+            tx._validate_amount_and_currency,
+            123.45,
+            self.currency.name,
```
vchu-odoo: fits one line
```suggestion
            ValidationError, tx._validate_amount_and_currency, 123.45, self.currency_euro.name
```

vchu-odoo (commented):
Thank you for the changes, last comments!

vchu-odoo (approved):
LGTM 👍

AntoineVDV (approved):
@robodoo r+

PR #245879, opened 2026-01-27
[IMP] mail: disable chat renaming

jayp-odoo:
**Purpose of this PR:**

disabled rename of direct messages since its professional unethical.
- also we are removing `custom_channel_name` feature from everywhere since the feature was intended for "chat" specifically, having it in whatsapp and livechat is accidental.
there is task for whatsapp to have name shared among all agents, so it's removal of custom_channel_name for whatsapp.

related PR: https://github.com/odoo/upgrade/pull/9358

[`enterprise pr`](https://github.com/odoo/enterprise/pull/105963)

[`task`](https://www.odoo.com/odoo/project/1519/tasks/5873798)

---

### DIFF
--- a/addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
+++ b/addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
@@ -37,7 +37,7 @@ const discussChannelPatch = {
         return super._computeCanHide(...arguments);
     },
     get displayName() {
-        if (this.channel_type !== "livechat" || this.self_member_id?.custom_channel_name) {
+        if (this.channel_type !== "livechat") {
             return super.displayName;
         }
         const selfMemberType = this.isTransient
--- a/addons/im_livechat/static/tests/thread_model_patch.test.js
+++ b/addons/im_livechat/static/tests/thread_model_patch.test.js
@@ -5,7 +5,6 @@ import {
     openDiscuss,
     start,
     startServer,
-    triggerHotkey,
 } from "@mail/../tests/mail_test_helpers";
 import { describe, test, waitFor } from "@odoo/hoot";
 import { Command, serverState, withUser } from "@web/../tests/web_test_helpers";
@@ -44,26 +43,6 @@ test("Thread name unchanged when inviting new users", async () => {
     await contains(".o-discuss-ChannelInvitation", { count: 0 });
     await contains(".o-discuss-ChannelMember", { text: "James" });
     await contains(".o-mail-DiscussContent-threadName[title='Visitor #20']");
-});
-
-test("Can set a custom name to livechat conversation", async () => {
-    const pyEnv = await startServer();
-    const guestId = pyEnv["mail.guest"].create({ name: "Visitor #20" });
-    const channelId = pyEnv["discuss.channel"].create({
-        channel_member_ids: [
-            Command.create({ partner_id: serverState.partnerId, livechat_member_type: "agent" }),
-            Command.create({ guest_id: guestId, livechat_member_type: "visitor" }),
-        ],
-        channel_type: "livechat",
-    });
-    await start();
-    await openDiscuss(channelId);
-    await click(".o-mail-DiscussSidebar-item:contains('Visitor #20')");
-    await contains(".o-mail-DiscussContent-threadName[title='Visitor #20']");
-    await insertText(".o-mail-DiscussContent-threadName", "New Name", { replace: true });
-    await triggerHotkey("Enter");
-    await contains(".o-mail-DiscussContent-threadName[title='New Name']");
-    await contains(".o-mail-DiscussSidebar-item:contains('New Name')");
 });
 
 test("Display livechat custom username if defined", async () => {
--- a/addons/mail/static/src/core/common/thread_actions.js
+++ b/addons/mail/static/src/core/common/thread_actions.js
@@ -36,8 +36,8 @@ registerThreadAction("fold-chat-window", {
 registerThreadAction("rename-thread", {
     condition: ({ channel, owner, thread }) =>
         channel &&
+        channel.isAllowedToRename &&
         owner.props.chatWindow?.isOpen &&
-        (thread.is_editable || channel.channel_type === "chat") &&
         !owner.isDiscussSidebarChannelActions,
     icon: "fa fa-fw fa-pencil",
     name: _t("Rename Thread"),
--- a/addons/mail/static/src/core/common/thread_model.js
+++ b/addons/mail/static/src/core/common/thread_model.js
@@ -289,10 +289,6 @@ export class Thread extends Record {
         return persona?.displayName || persona?.name;
     }
 
-    get supportsCustomChannelName() {
-        return this.channel?.isChatChannel && this.channel?.channel_type !== "group";
-    }
-
     get displayName() {
         return this.channel?.displayName ?? this.display_name;
     }
--- a/addons/mail/static/src/core/public_web/discuss_content.xml
+++ b/addons/mail/static/src/core/public_web/discuss_content.xml
@@ -27,7 +27,7 @@
                         </div>
                         <AutoresizeInput
                             className="'o-mail-DiscussContent-threadName fw-bold flex-shrink-1 py-0 ' + (showsChatLocalDateTime ? 'fs-5' : 'fs-4')"
-                            enabled="thread.is_editable or thread.channel?.channel_type === 'chat'"
+                            enabled="thread.channel?.isAllowedToRename"
                             onValidate.bind="renameThread"
                             value="thread.displayName || ''"
                         />
--- a/addons/mail/static/src/discuss/core/common/discuss_channel_model.js
+++ b/addons/mail/static/src/discuss/core/common/discuss_channel_model.js
@@ -101,6 +101,14 @@ export class DiscussChannel extends Record {
     get allowedToLeaveChannelTypes() {
         return ["channel", "group"];
     }
+    get allowedToRenameChannelTypes() {
+        return ["channel", "group"];
+    }
+    get isAllowedToRename() {
+        return (
+            this.allowedToRenameChannelTypes.includes(this.channel_type) && this.thread.is_editable
+        );
+    }
     get areAllMembersLoaded() {
         return this.member_count === this.channel_member_ids.length;
     }
@@ -194,9 +202,6 @@ export class DiscussChannel extends Record {
         return this.channel_member_ids.filter(({ persona }) => persona?.notEq(this.store.self));
     }
     get displayName() {
-        if (this.supportsCustomChannelName && this.self_member_id?.custom_channel_name) {
-            return this.self_member_id.custom_channel_name;
-        }
         if (this.channel_type === "chat" && this.correspondent) {
             return this.correspondent.name;
         }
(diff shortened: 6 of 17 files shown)


addons/mail/static/src/core/common/thread_model.js
```diff
@@ -287,7 +287,7 @@ export class Thread extends Record {
     }
 
     get supportsCustomChannelName() {
-        return this.channel?.isChatChannel && this.channel?.channel_type !== "group";
+        return this.channel?.channel_type !== "group" && this.channel?.channel_type !== "chat";
```
tsm-odoo: I feel like the `isChatChannel` part is still relevant.  Sure, the `rename` function already handles it. But conceptually, channels do not support custom channel names.

Also, we can rewrite the condition to:

```js
return !["chat", "group"].includes(this.channel?.channel_type);
```

alexkuhn: Old condition with `isChatChannel && channel_type !== "group"` matched the following channel_type:

`["chat", "livechat", "whatsapp"]`

New condition here includes `"channel"`, which shouldn't have custom_channel_name.

We want to at least restrict to:
`["livechat", "whatsapp"]`

We have a task in pipeline to have rename of `"whatsapp"` to be seen by even agent:
https://www.odoo.com/odoo/project/1519/tasks/5475084

So this asks for removal of `custom_channel` too. There's no task for livechat, but I remember seeing internal discussions for also sharing the live chat name to all agents.

All this to say, the move is most likely removal of `custom_channel_name` entirely.

addons/mail/static/src/core/public_web/discuss_content.xml
```diff
@@ -28,7 +28,7 @@
                         </div>
                         <AutoresizeInput
                             className="'o-mail-DiscussContent-threadName fw-bold flex-shrink-1 py-0 ' + (showsChatLocalDateTime ? 'fs-5' : 'fs-4')"
-                            enabled="thread.is_editable or thread.channel?.channel_type === 'chat'"
+                            enabled="thread.is_editable and thread.channel?.channel_type !== 'chat'"
```
tsm-odoo: It looks wrong to just negate the condition instead of removing it.

alexkuhn: Could omit `channel_type`, `is_editable` should be enough to cover all but "chat".

jayp-odoo: if we remove and just put like `enabled="thread.is_editable"` then Input will be enabled.

alexkuhn: Ah no is_editable is just check write access, so this makes sense to have !== "chat" for admin. Ok to keep.

addons/im_livechat/static/src/core/common/thread_model_patch.js
```diff
@@ -37,4 +37,8 @@ patch(Thread.prototype, {
         }
         return super.getPersonaName(persona);
     },
+
+    get supportsChannelRename() {
+        return super.supportsChannelRename || this.channel?.channel_type === "livechat";
+    },
```
alexkuhn: Should do the same for `whatsapp`.

This type of conversation also is in the `supportsCustomChannelName` right now. With the removal of `custom_channel_name`, the whatsapp conversations need to keep renaming feature.

Without adding `whatsapp`, these branches add a regression by disabling rename of whatsapp conversations.

addons/im_livechat/static/tests/thread_model_patch.test.js
```diff
@@ -46,26 +45,6 @@ test("Thread name unchanged when inviting new users", async () => {
     await contains(".o-mail-DiscussContent-threadName[title='Visitor #20']");
 });
 
-test("Can set a custom name to livechat conversation", async () => {
```
seb-odoo, line 49: I don't think we should be removing the test, we still want to allow renaming livechat as before, but it applies to everyone.

Unless I'm missing something and there is another test already covering it?

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -286,8 +286,8 @@ export class Thread extends Record {
         return persona?.displayName || persona?.name;
     }
 
-    get supportsCustomChannelName() {
-        return this.channel?.isChatChannel && this.channel?.channel_type !== "group";
+    get supportsChannelRename() {
+        return ["channel", "group"].includes(this.channel?.channel_type);
```
seb-odoo: As commented on the other PR:
- if the condition only depends on the type, the getter should just return a list of types
- this should be moved to channel model instead of thread

addons/im_livechat/static/tests/thread_model_patch.test.js
alexkuhn: Pretty sure we don't have such test since we don't officially support rename that is broadcast to all agents before this PR. So yes we should keep the test and adapt to `"custom name"` => `"name"`.

addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
```diff
@@ -29,6 +29,9 @@ const discussChannelPatch = {
     get allowedToLeaveChannelTypes() {
         return [...super.allowedToLeaveChannelTypes, "livechat"];
     },
+    get supportsChannelRenameTypes() {
```
seb-odoo: If we follow the naming pattern of others: `allowedToRenameChannelTypes`

addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
```diff
@@ -37,9 +40,12 @@ const discussChannelPatch = {
         return super._computeCanHide(...arguments);
     },
     get displayName() {
-        if (this.channel_type !== "livechat" || this.self_member_id?.custom_channel_name) {
+        if (this.channel_type !== "livechat") {
             return super.displayName;
         }
+        if (this.name) {
```
alexkuhn: The problem with this branching is that all conversations have necessarily a `name` set:
https://github.com/odoo/odoo/blob/19.0/addons/mail/models/discuss/discuss_channel.py#L1390

And this is equivalent to `display_name`, so it reflects on the UI for all livechat conversations:
Image

I think that the LOC to set manually `name` on channel creation should be removed. The value is already passed as `display_name`.

@seb-odoo let me know what you think about it. I remember a long time ago we didn't like reliance on `display_name` but in the end we split `name` and `display_name` but have to keep `display_name`. I think this LOC is a residue from intention to separate from `display_name` that couldn't be 100% achieved.

If we remove this LOC, we probably have to compensate with fallback client-side on `display_name` (this is already passed to client-code, from what I can see in code), and we'd have to prepare upgrade scripts to remove the `name` on all affected livechat and whatsapp conversations at the very least.

seb-odoo: Situation before this PR:

`name` is mostly technical (dummy in `chat` and `livechat`, honestly could be left empty), actually empty in `group` (when we want to rely on the list of members to give them name).

Only for `channel` (and `group` when set) it is the actual name we want to display.

In most cases, even when the `name` is set, we want to display something different. In `livechat`, `whatsapp` and `chat` in particular, we prefer to contextually display the name of the correspondent(s).

So indeed, `name` should not be displayed to the user, except as a last resort when nothing else is available.

The only place where this fallback to `name` should be coded is inside `displayName` itself. All other places should use `displayName`.

Current diff uses `name` immediately rather than as a fallback, so yes it's likely incorrect.

seb-odoo: And yes, if the goal of the task is somehow to make `name` the priority (I'm ok with that if we want it), it then should not be automatically filled in code when creating channels, and it should be cleared in upgrade  to remove existing "dummy" occurrences (but it is pretty much impossible to detect whether they were dummy occurrences or values changed by the user).

seb-odoo: In any case, I would not change that as part of "disable chat renaming" task, but on its own task, if we really want it.

alexkuhn: Ok to sum up, we can proceed as follow:
- remove this `if (this.name)` branching for `livechat`.
- remove the rename feature of `livechat` completely (not that useful, needs `name` cleaning that is postponed for later)
- ok to keep new rename feature of `whatsapp` as the `name` is already clean.

addons/mail/static/src/core/public_web/discuss_content.xml
```diff
@@ -27,7 +27,7 @@
...
```
alexkuhn:
```suggestion
                            enabled="thread.is_editable and thread.channel and thread.channel.allowedToRenameChannelTypes.includes(thread.channel.channel_type)"
```

Current condition allows to edit the input of livechat even when this has no effect.
Maybe we can introduce a small getter for `channel.isAllowedToRename` that evaluates `thread.channel.allowedToRenameChannelTypes.includes(...)`, for more readable code here and in `Channel.rename` function.

addons/mail/static/tests/discuss_app/discuss.test.js
```diff
@@ -1862,21 +1847,6 @@ test("Thread avatar is not editable in DM chat", async () => {
...
-
-    await start();
-    await openDiscuss(channelId);
-    await insertText("input.o-mail-DiscussContent-threadName:enabled", "Mitchell Admin", {
-        replace: true,
-    });
-    triggerHotkey("Enter");
-    await expect.waitForSteps([]);
-});
-
```
alexkuhn, line 1879: Nitpick: this test can be kept but instead of rename of `chat` can make it rename of `channel`.

I presume the test was introduced because of mis-handling of `custom_channel_name` but we could have the same problem with `name` actually but there's no test coverage AFAIK.

addons/mail/tests/discuss/test_discuss_channel_access.py
```diff
@@ -563,8 +563,8 @@ def _execute_action_member(self, channel_id, user_key, target, operation, result
             self.assertEqual(len(member), 1, "should find the target member")
             if operation == "read":
                 self.assertEqual(len(ChannelMemberAsUser.search(domain)), 1 if result else 0)
-                member.read(["custom_channel_name"])
+                member.read(["custom_notifications"])
             elif operation == "write":
-                member.write({"custom_channel_name": "new name"})
+                member.write({"custom_notifications": "mentions"})
```
alexkuhn, line 568: Diff to keep original intended coverage there's no ACL problem, if I understand correctly?
Ok for me if that's what I understand :)

alexkuhn:
The thread action `"rename-thread"` hasn't been adapted to take the `isAllowedToRename` into account.
As a result, it's possible to fake "rename" in chat window:
Screenshot 2026-02-05 at 16 54 38

addons/mail/static/src/core/common/thread_actions.js
```diff
@@ -36,8 +36,9 @@ registerThreadAction("fold-chat-window", {
 registerThreadAction("rename-thread", {
     condition: ({ channel, owner, thread }) =>
         channel &&
+        channel.isAllowedToRename &&
         owner.props.chatWindow?.isOpen &&
-        (thread.is_editable || channel.channel_type === "chat") &&
+        thread.is_editable &&
```
seb-odoo: It seems `is_editable` is always checked when we check `isAllowedToRename` (rightfully so), so we might want to check `is_editable` directly inside `isAllowedToRename` instead to avoid the repetition.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -669,24 +672,14 @@ export class DiscussChannel extends Record {
             newName !== this.displayName &&
             ((newName && this.channel_type === "channel") || this.isChatChannel)
         ) {
-            if (["channel", "group"].includes(this.channel_type)) {
+            if (this.isAllowedToRename) {
```
seb-odoo: We usually avoid nested conditions when possible, this could we merged into the above one.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -669,24 +672,14 @@ export class DiscussChannel extends Record {
             newName !== this.displayName &&
             ((newName && this.channel_type === "channel") || this.isChatChannel)
```
seb-odoo: `|| this.isChatChannel` looks obsolete, it feels like `isAllowedToRename` is the correct thing to check here.

seb-odoo: And the rest of the condition is just there to ensure we always have a name for channel type `channel`, which could be written as `newName || this.channel_type !== "channel"`.
But I believe it is slightly wrong before this PR, and we should invert the condition, we only want to allow empty name for channel type `group` I think.
`newName || this.channel_type === "group"`

jayp-odoo:
can get a review please its updated :green_apple: thanks.

alexkuhn (approved):
@robodoo delegate+ r+

PR #245880, opened 2026-01-27
[FIX] mrp_subcontracting_purchase: fix invoice creation

Vancleeff:
Before this commit, creating an invoice for a purchase order with a subcontracted tracked product triggered a singleton error.

This commit fixes the issue by properly handling multiple finished move lines for the same product, which occurs with tracked products, some back order flows, and possibly upgrade scripts.

OPW-5712931

### DIFF
--- a/addons/mrp_subcontracting_purchase/models/account_move_line.py
+++ b/addons/mrp_subcontracting_purchase/models/account_move_line.py
@@ -26,5 +26,5 @@ class AccountMoveLine(models.Model):
         finished_moves = set()
         for m in moves:
             if mo := m._get_subcontract_production():
-                finished_moves.add(mo.move_finished_ids.filtered(lambda mf: mf.product_id == m.product_id).id)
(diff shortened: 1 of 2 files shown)


addons/mrp_subcontracting_purchase/models/account_move_line.py
```diff
@@ -23,8 +23,8 @@ def _get_price_unit_val_dif_and_relevant_qty(self):
 
     def _get_stock_moves(self):
         moves = super()._get_stock_moves()
-        finished_moves = set()
         for m in moves:
             if mo := m._get_subcontract_production():
-                finished_moves.add(mo.move_finished_ids.filtered(lambda mf: mf.product_id == m.product_id).id)
```
Whenrow, line 29: The `set` and `add()` is there for performance reason instead of `|=`.
So I would rather do 
```python
finished_moves |= set(mo.move_finished_ids.filtered(lambda mf: mf.product_id == m.product_id).ids)
```
and keep the browse at the end

PR #245881, opened 2026-01-27
[FIX] hr_holidays_attendance: compute employee_overtime on leave allocations

suma-odoo:
Issue :
-
After commit odoo/odoo@6d7f4012cfa06b35089e7557522fcbf658978b37, the `employee_overtime` field in `hr_holidays_attendance` was converted to a computed field on `hr.leave` model to use the new overtime deduction logic. However, the same change was not applied to `hr.leave.allocation` model.

Steps to Reproduce:
-
1. Create a Leave Allocation for an employee who has extra overtime hours.
2. Set the allocation less than the available extra hours.
3. In 19.0, when saving the allocation and creating a new allocation for the same employee:
   - The `employee_overtime` value did not update.
   - Even though the actual extra hours were reduced, the field still showed the old value.

Fix:
-
After this fix, the `employee_overtime` field in `hr.leave.allocation` is computed correctly, reflecting the actual available overtime hours.
This commit updates `hr.leave.allocation` by:
   - Converting `employee_overtime` to a computed field.
   - Computing its value using `_get_deductible_employee_overtime`.
This aligns the overtime computation between leave requests (`hr.leave`) and leave allocations (`hr.leave.allocation`).

Before Fix:
-
before_fix

After Fix:
-
after_fix

opw-5479200

Current behavior before PR:

---

### DIFF
--- a/addons/hr_holidays_attendance/models/hr_leave.py
+++ b/addons/hr_holidays_attendance/models/hr_leave.py
@@ -35,50 +35,18 @@ class HrLeave(models.Model):
 
     @api.model
     def _get_deductible_employee_overtime(self, employees):
-        # return dict {employee: number of hours}
-        diff_by_employee = defaultdict(lambda: 0)
-        for employee, hours in self.env['hr.attendance.overtime.line'].sudo()._read_group(
-            domain=[
-                ('compensable_as_leave', '=', True),
-                ('employee_id', 'in', employees.ids),
-                ('status', '=', 'approved'),
-            ],
-            groupby=['employee_id'],
-            aggregates=['manual_duration:sum'],
-        ):
-            diff_by_employee[employee] += hours
-        for employee, hours in self._read_group(
-            domain=[
-                ('holiday_status_id.overtime_deductible', '=', True),
-                ('holiday_status_id.requires_allocation', '=', False),
-                ('employee_id', 'in', employees.ids),
-                ('state', 'not in', ['refuse', 'cancel']),
-            ],
-            groupby=['employee_id'],
-            aggregates=['number_of_hours:sum'],
-        ):
-            diff_by_employee[employee] -= hours
-        for employee, hours in self.env['hr.leave.allocation']._read_group(
-            domain=[
-                ('holiday_status_id.overtime_deductible', '=', True),
-                ('employee_id', 'in', employees.ids),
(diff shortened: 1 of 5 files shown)


addons/hr_holidays_attendance/models/hr_leave_allocation.py
```diff
@@ -23,13 +23,19 @@ def default_get(self, fields):
...
     @api.depends('holiday_status_id')
     def _compute_overtime_deductible(self):
         for allocation in self:
             allocation.overtime_deductible = allocation.holiday_status_id.overtime_deductible
 
+    @api.depends('employee_id')
+    def _compute_employee_overtime(self):
+        diff_by_employee = self.env['hr.leave']._get_deductible_employee_overtime(self.employee_id)
+        for leave in self:
+            leave.employee_overtime = diff_by_employee[leave.employee_id]
```
amah-odoo:
```suggestion
        for allocation in self:
            allocation.employee_overtime = diff_by_employee[allocation.employee_id]
```

amah-odoo (commented):
Thanks for your work!

For this I think we should also move the `_get_deductible_employee_overtime` method to employee model we can update the allocation call for it in 19.0 to call from employee and drop it in the master-fw 🤔

amah-odoo (changes requested):
we should not drop a method from stable version as people could extend it in custom code. you should keep it in leave but call the employees one inside it and return res. Also, I do not see why it is static in the employee model I think it could normal use self as the employees recoredset

addons/hr_holidays_attendance/models/hr_leave_allocation.py
amah-odoo: I think we can go with this and we are ok to go
```suggestion
            allocation.employee_overtime = diff_by_employee.get(allocation.employee_id, 0)
```

amah-odoo (commented):
I would love to update the tests to call the employee method to ease the drop of the method on master 👀

amah-odoo (commented):
I searched the code base and found more calls for `_get_deductible_employee_overtime` can we update all of them 👀

addons/hr_holidays_attendance/models/hr_leave.py
amah-odoo, line 36: we should keeps this as it is

PR #245892, opened 2026-01-27
[IMP] im_livechat: improve live chat conversation transcript file name

nmak-odoo:

Before this commit, the live chat transcript file was named only with
the conversation id `(e.g, transcript_46.pdf)`, which made it difficult to
identify its content.

After this commit, the file name follows a more descriptive format:
`Live chat - <company/customer_name> - Jan 13, 10:00 AM.pdf`.

This commit also removes the unused company value previously passed to
the report rendering context.

task-5489913

---

### DIFF
--- a/addons/im_livechat/controllers/main.py
+++ b/addons/im_livechat/controllers/main.py
@@ -1,7 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from urllib.parse import urlsplit
-from zoneinfo import ZoneInfo
 
 from werkzeug.exceptions import NotFound
 
@@ -221,8 +220,20 @@ class LivechatController(http.Controller):
         channel = request.env["discuss.channel"].search([("id", "=", channel_id)])
         if not channel:
             raise NotFound()
-        user, guest = request.env["res.users"]._get_current_persona()
-        tz = ZoneInfo(user.tz or guest.timezone or "UTC")
+        tz = channel._get_livechat_customer_timezone()
+        # sudo: discuss.channel - access customer history to derive transcript name
+        conversation_name = next(
+            (
+                name
+                for customer in channel.sudo().livechat_customer_history_ids
+                if (
+                    name := customer.partner_id.commercial_company_name
+                    or customer.partner_id.name
+                    or customer.guest_id.name
+                )
+            ),
+            False,
+        )
         pdf, _type = (
             request.env["ir.actions.report"]
             .sudo()
@@ -230,7 +241,6 @@ class LivechatController(http.Controller):
                 "im_livechat.action_report_livechat_conversation",
                 channel.ids,
                 data={
-                    "company": request.env.company,
                     "tz": tz,
                     "correspondent_names": format_list(
                         request.env,
@@ -243,8 +253,14 @@ class LivechatController(http.Controller):
                 },
             )
         )
+        timestamp = channel.create_date.astimezone(tz).strftime("%b %d, %I:%M %p")
+        filename = (
+            "Live chat - %s - %s.pdf" % (conversation_name, timestamp)
+            if conversation_name
+            else "Live chat - %s.pdf" % timestamp
+        )
         headers = [
-            ("Content-Disposition", content_disposition(f"transcript_{channel.id}.pdf", "inline")),
+            ("Content-Disposition", content_disposition(filename, "inline")),
             ("Content-Length", len(pdf)),
             ("Content-Type", "application/pdf"),
         ]
--- a/addons/im_livechat/models/discuss_channel.py
+++ b/addons/im_livechat/models/discuss_channel.py
@@ -621,12 +621,23 @@ class DiscussChannel(models.Model):
                 subtype_xmlid="mail.mt_comment",
             )
 
+    def _get_livechat_customer_timezone(self):
+        """Return the customer's timezone"""
+        self.ensure_one()
+        # sudo: discuss.channel - access partner's/guest's timezone from customer history.
+        tz_name = next(
+            (
+                tz
+                for customer in self.sudo().livechat_customer_history_ids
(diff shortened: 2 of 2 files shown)


maad-odoo (approved):

addons/im_livechat/controllers/main.py
```diff
@@ -221,8 +221,8 @@ def download_livechat_transcript(self, channel_id):
         channel = request.env["discuss.channel"].search([("id", "=", channel_id)])
         if not channel:
             raise NotFound()
-        user, guest = request.env["res.users"]._get_current_persona()
-        tz = ZoneInfo(user.tz or guest.timezone or "UTC")
+        customer_tz = channel._get_livechat_customer_timezone()
+        tz = ZoneInfo(customer_tz)
```
tsm-odoo: Should be declared as close as possible to the usage (below report rendering).

addons/im_livechat/controllers/main.py
```diff
@@ -243,8 +243,9 @@ def download_livechat_transcript(self, channel_id):
                 },
             )
         )
+        timestamp = channel.create_date.astimezone(tz).strftime("%b %d, %I:%M %p")
         headers = [
-            ("Content-Disposition", content_disposition(f"transcript_{channel.id}.pdf", "inline")),
+            ("Content-Disposition", content_disposition(f"Live Chat - {request.env.company.name} - {timestamp}.pdf", "inline")),
```
tsm-odoo:
```suggestion
            ("Content-Disposition", content_disposition(f"Live chat - {request.env.company.name} - {timestamp}.pdf", "inline")),
```
⛏️

addons/im_livechat/models/discuss_channel.py
```diff
@@ -611,12 +611,22 @@ def _close_livechat_session(self, message):
                 subtype_xmlid='mail.mt_comment'
             )
 
+    def _get_livechat_customer_timezone(self):
+        """ Return the customer's timezone for livechat transcript rendering."""
```
tsm-odoo:
```suggestion
        """Return the customer's timezone."""
```

Or for anything else really, it's up to the caller :p

addons/im_livechat/controllers/main.py
tsm-odoo: As both call sites call ZoneInfo afterwards, the helper could directly return a zone info instance.

nmak-odoo: report rendering depends on `tz` in the data, so this placement is correct i think.

addons/im_livechat/controllers/main.py
```diff
@@ -243,8 +241,9 @@ def download_livechat_transcript(self, channel_id):
                 },
             )
         )
+        timestamp = channel.create_date.astimezone(tz).strftime("%b %d, %I:%M %p")
         headers = [
-            ("Content-Disposition", content_disposition(f"transcript_{channel.id}.pdf", "inline")),
+            ("Content-Disposition", content_disposition(f"Live chat - {request.env.company.name} - {timestamp}.pdf", "inline")),
```
tsm-odoo: Naming the file based on the company of the user downloading the transcript doesn't really make sense. We should probably use the customer's company instead?

nmak-odoo: I used `env.company`, since we already pass that same company to `_render_qweb_pdf()`. 
Wouldn’t using the customer’s company only for the filename make the filename and the transcript content differ?

tsm-odoo: We pass it, but I don't see any usage in `livechat_email_template`. And if it's used, we might want to question whether it really is what we want.

nmak-odoo: I remembered there being a company.name usage in the template, and since the email flow still uses the company through `mail_notification_layout` while the download side was also still passing company into `_render_qweb_pdf()`, I assumed we were still keeping that same company-based logic here too.

But the PDF does not actually go through `mail_notification_layout`, so yeah, its better to use the customer company in the filename and remove company from the data passed to the render for download, since it is no longer being used.

Thanks for pointing it out :)

addons/im_livechat/controllers/main.py
```diff
@@ -221,16 +220,25 @@ def download_livechat_transcript(self, channel_id):
...
+        company_name = next(
+            (
+                company_name
+                for company_name in channel.sudo().livechat_customer_history_ids.mapped(
+                    "partner_id.commercial_company_name"
+                )
+                if company_name
+            ),
+            request.env.company.name,
+        )
```
tsm-odoo: I wouldn't even fallback on the user's company name. That doesn't make sense (it's not even guaranteed that the user downloading the file took part in the convo). You are downloading a conversation, the name is supposed to reflect this conversation. Your own company name doesn't make sense in this context.

Also, this could be much simpler:

`channel.sudo().livechat_customer_history_ids.partner_id.filtered("commercial_company_name")[0].commercial_company_name`

nmak-odoo: Then I think it’s better to use the customer/guest name as a fallback when there's no company associated with the customer.

>channel.sudo().livechat_customer_history_ids.partner_id.filtered("commercial_company_name")[0].commercial_company_name

This would break when commercial_company_name is absent, so I kept a simpler fallback flow instead.

addons/im_livechat/controllers/main.py
```diff
@@ -221,16 +220,26 @@ def download_livechat_transcript(self, channel_id):
...
+            (
+                name
+                for customer in channel.sudo().livechat_customer_history_ids
+                if (
+                    name := customer.partner_id.commercial_company_name
+                    or customer.partner_id.name
+                    or customer.guest_id.name
+                )
+            )
+        )
```
tsm-odoo, line 236: We should still provide a fallback (False) to next.  And only add it to the name if we could find it. Looking good otherwise :)

PR #224800, opened 2025-08-29
[IMP] mail: editing html message

zel-odoo:
---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #245893, opened 2026-01-27
[FIX] mail: rpc call when editing an edited message with no change

maki-odoo:
Since #224800, we've been removing the `edited` element from the current message
body in order to compare it with the updated body. However, the `edited`
element may also be present in the updated body. This change removes the
`edited` element from the updated body as well, ensuring a reliable comparison
when editing an edited message.

Steps to reproduce:
- Send a message.
- Click `edit`, change the message body and click `save`.
- Click `edit` again, don't change the message body this time, then click `save` .
- An RPC is made to update the message.

### DIFF
--- a/addons/mail/static/src/core/common/message_model.js
+++ b/addons/mail/static/src/core/common/message_model.js
@@ -540,12 +540,11 @@ export class Message extends Record {
         attachments = [],
         { mentionedChannels = [], mentionedPartners = [], mentionedRoles = [] } = {}
     ) {
-        const bodyEl = createElementWithContent("div", this.body);
-        bodyEl.querySelector("span.o-mail-Message-edited")?.remove();
-        if (
-            createElementWithContent("div", body).innerHTML === bodyEl.innerHTML &&
-            attachments.length === 0
-        ) {
+        const messageBodyEl = createElementWithContent("div", this.body);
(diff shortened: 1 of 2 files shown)


addons/mail/static/src/core/common/message_model.js
```diff
@@ -540,12 +540,11 @@ export class Message extends Record {
...
-        bodyEl.querySelector("span.o-mail-Message-edited")?.remove();
-        if (
-            createElementWithContent("div", body).innerHTML === bodyEl.innerHTML &&
-            attachments.length === 0
-        ) {
+        const MessageBodyEl = createElementWithContent("div", this.body);
+        const updatedBodyEl = createElementWithContent("div", body);
+        MessageBodyEl.querySelector("span.o-mail-Message-edited")?.remove();
+        updatedBodyEl.querySelector("span.o-mail-Message-edited")?.remove();
+        if (updatedBodyEl.innerHTML === MessageBodyEl.innerHTML && attachments.length === 0) {
```
alexkuhn: `MessageBodyEl` => `messageBodyEl`
No capital letter, this is reserved for classes.

alexkuhn (approved):
LGTM otherwise.
@robodoo delegate+

PR #245898, opened 2026-01-27
[IMP] mail: use Bookmark as the string for starred

zel-odoo:
This commit updates the string used for starred messages in Discuss with Bookmark.

task-5873789
https://github.com/odoo/enterprise/pull/105757
https://github.com/odoo/upgrade/pull/9363

---


addons/mail/static/src/core/common/message.xml
```diff
@@ -223,8 +223,8 @@
 </t>
 
 <t t-name="mail.Message.sidebarStarred">
-    <small class="text-center lh-1 o-opacity-35" t-att-class="{ 'align-self-start ms-1': isAlignedRight, 'align-self-end me-1': !isAlignedRight, 'mt-2': props.squashed, 'mt-1': !props.squashed }" title="Starred message">
-        <i class="fa fa-star o-mail-starred"/>
+    <small class="text-center lh-1 o-opacity-35" t-att-class="{ 'align-self-start ms-1': isAlignedRight, 'align-self-end me-1': !isAlignedRight, 'mt-2': props.squashed, 'mt-1': !props.squashed }" title="Bookmark">
+        <i class="fa fa-bookmark o-mail-starred"/>
```
alexkuhn: `o-mail-starred` made sense with star icon. Now this is bookmark, we should play with only `bookmark` and `bookmark-o`.

addons/mail/static/src/core/common/message_model.js
```diff
@@ -801,14 +801,33 @@ export class Message extends Record {
         return false;
     }
 
-    async toggleStar() {
+    async toggleStar(isInBookmark) {
```
alexkuhn: Would rename `toggleStar` to `toggleBookmark`, so terminology in whole discuss code is updated with "bookmark" new naming.

I don't mind to do in follow-up PR, but that should be done imminently after merge.

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -69,7 +69,7 @@ const StorePatch = {
             model: "mail.box",
         };
         this.starred = {
-            display_name: _t("Starred messages"),
+            display_name: _t("Bookmarks"),
             id: "starred",
             model: "mail.box",
         };
```
alexkuhn: Same remark: would rename `starred` in whole code for bookmark.
I get this is lots of diff, so I wouldn't mind follow-up PR but as I said should be done imminently to make code look correct with new terminology.

addons/mail/static/src/core/web/thread_patch.xml
```diff
@@ -10,8 +10,13 @@
...
                 </t>
                 <t t-if="props.thread.id === 'starred'">
-                    <h4 class="mb-3 fw-bolder">No starred messages</h4>
-                    You can mark any message as 'starred', and it shows up in this mailbox.
+                    <div class="d-flex flex-column align-items-center text-center">
+                        <h4 class="mb-3 fw-bolder">Bookmark important notifications</h4>
+                        <span>Save messages here to easily keep track of them.</span>
+                        <button t-if="store.self_user?.notification_type === 'inbox'" class="btn btn-primary px-3 m-3" t-on-click="()=> this.store.inbox.setAsDiscussThread()">
+                            Try it
+                        </button>
```
alexkuhn: I see it in spec but this button doesn't make sense.
There are better way to incite people know the feature exists, like show / tell there's a "bookmark" message action.

As it is, the button is useless and the UX / UI would be better without this button.

addons/mail/static/src/core/common/message_model.js
zel-odoo: Ideally, we should use "bookmark" everywhere. However, I feel that starred makes more sense technically, while "Bookmark" is easier for users to understand. That is why I hesitated to change the entire codebase and only updated the UI strings. For example, `message.starred` feels logical, whereas `message.bookmark` or `message.bookmarked` feels less expressive. We should discuss whether we want to replace all star-related technical variables with "bookmark."

alexkuhn: This would be named `message.bookmarked`, as the message is a bookmark. `message.bookmark` would be misleading in that this would give the impression there's a model like `mail.bookmark`, which isn't the case.

I think `star` vs `bookmark` is more about being used to use either terminology. I think preferring `message.starred` is because we are not used to `message.bookmarked`... yet.

zel-odoo: that might be the reason...... ok i will update all the stars as well, i will push the code later

addons/mail/controllers/webclient.py
```diff
@@ -167,13 +167,10 @@ def _store_init_messaging_global_fields(cls, res: Store.FieldList, bus_last_id):
             },
         )
         res.attr(
-            "starred",
+            "bookmark",
             {
-                "counter": user.env["mail.message"].search_count(
-                    [("starred_partner_ids", "in", user.partner_id.ids)],
-                ),
```
seb-odoo, line 198: Removing this looks wrong, counter is not correct on page load now.

I wish we should remove this counter from the UI though, I'm not using this feature because the counter is annoying, I want a 0 counter sidebar.

I don't know why people keep using bookmark as todo for which they want to reach 0 rather as a way to permanently keep important information.

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -44,7 +44,7 @@ const StorePatch = {
             eager: true,
         });
         this.inbox = fields.One("mail.thread");
-        this.starred = fields.One("mail.thread");
+        this.bookmark = fields.One("mail.thread");
```
seb-odoo: I would name this `bookmarkBox` or similar, it would make it easier to find in code and more relevant to actual use. There is no such thing as a "bookmark" in store.

addons/mail/static/src/core/web/message_model_patch.js
```diff
@@ -1,9 +1,25 @@
...
+        this.is_bookmarked = fields.Attr(false, {
+            onUpdate() {
+                const bookmarks = this.store.bookmark;
+                if (this.is_bookmarked) {
+                    bookmarks.counter++;
+                    bookmarks.messages.add(this);
+                } else {
+                    bookmarks.counter--;
+                    bookmarks.messages.delete(this);
+                }
```
seb-odoo: This is too naive, we don't know if the message was already taken into account in the counter or not.

This cannot be done in business code, it depends on the handler.

Previous code had check on `counter_bus_id` which was already a little bit better. The best way would be to wait for PR from @tsm-odoo to properly resolve store insert issue.

But then we'd need to build something on top of it to handle counters properly.

Honestly the best way for now is probably just to re-send the full counter every time the user changes something. It's normally not going to change a lot or to be too complicated to compute, unlike the inbox counter (which notifies persons in batch) or the channel unread counter (which changes all the time for all users).

addons/hr_recruitment/data/hr_recruitment_demo.xml
```diff
@@ -750,7 +750,7 @@
         <field name="subtype_id" ref="mail.mt_comment"/>
         <field name="author_id" ref="base.partner_demo"/>
     </record>
-    <function model="mail.message" name="toggle_message_starred"
+    <function model="mail.message" name="toggle_message_bookmark"
```
seb-odoo: I'm still strongly against toggle method as they are non-deterministic by nature.

While we have diff on them, we should make one clear `add_bookmark` and one clear `remove_bookmark` method in python + the same in JS + separate actions.

addons/mail/static/src/core/common/message_model.js
```diff
@@ -808,14 +808,33 @@ export class Message extends Record {
         return false;
     }
 
-    async toggleStar() {
+    async toggleBookmark(isInBookmarkBox) {
```
seb-odoo: split in add and remove

addons/mail/static/src/core/common/message_actions.js
```diff
@@ -102,11 +102,11 @@ registerMessageAction("reply-to", {
     sequence: ({ message, store, thread }) =>
         thread?.eq(store.inbox) || message.isSelfAuthored ? 55 : 20,
 });
-registerMessageAction("toggle-star", {
-    condition: ({ message }) => message.canToggleStar,
-    icon: ({ message }) => (message.starred ? "fa fa-star o-mail-starred" : "fa fa-star-o"),
-    name: ({ message }) => (message.starred ? _t("Remove Star") : _t("Add Star")),
-    onSelected: ({ message }) => message.toggleStar(),
+registerMessageAction("toggle-bookmark", {
```
seb-odoo: split in add and remove

addons/mail/static/src/core/web/thread_patch.xml
```diff
@@ -9,9 +9,11 @@
                         New messages appear here.
                     </div>
                 </t>
-                <t t-if="props.thread.id === 'starred'">
-                    <h4 class="mb-3 fw-bolder">No starred messages</h4>
-                    You can mark any message as 'starred', and it shows up in this mailbox.
+                <t t-if="props.thread.id === 'bookmark'">
+                    <div class="d-flex flex-column align-items-center text-center">
+                        <h4 class="mb-3 fw-bolder">Bookmark important notifications</h4>
```
seb-odoo: This message is misleading, we don't bookmark notifications but messages.

addons/mail/static/src/js/tours/discuss_channel_tour.js
```diff
@@ -53,18 +53,18 @@ registry.category("web_tour.tours").add("discuss_channel_tour", {
             run: "press Enter",
         },
         {
-            trigger: ".o-mail-Message[data-persistent] [title='Add Star']:not(:visible)",
-            content: _t("Hover on your message and add a star"),
+            trigger: ".o-mail-Message[data-persistent] [title='Bookmark']:not(:visible)",
+            content: _t("Hover on your message bookmark it"),
```
seb-odoo: There is a problem in this phrase.

seb-odoo (changes requested):

addons/mail/static/src/core/web/message_model_patch.js
zel-odoo: Is the PR gonna be merged soon so that i freeze this Pr or I will just re-send the counter every time for now and merge this pr first?

seb-odoo: Don't wait for it, just revert for now I guess and focus on the rename.

zel-odoo: ok, will do

addons/hr_recruitment/data/hr_recruitment_demo.xml
```diff
@@ -750,7 +750,10 @@
         <field name="subtype_id" ref="mail.mt_comment"/>
         <field name="author_id" ref="base.partner_demo"/>
     </record>
-    <function model="mail.message" name="toggle_message_starred"
+    <function model="mail.message" name="add_bookmark"
+            eval="[ref('msg_case18_aplicant')]"
+    />
+    <function model="mail.message" name="remove_bookmark"
```
seb-odoo: I don't think this should be removed just after adding it.

addons/mail/controllers/webclient.py
```diff
@@ -167,13 +167,13 @@ def _store_init_messaging_global_fields(cls, res: Store.FieldList, bus_last_id):
...
-            "starred",
+            "bookmarkBox",
             {
                 "counter": user.env["mail.message"].search_count(
-                    [("starred_partner_ids", "in", user.partner_id.ids)],
+                    [("bookmarked_partner_ids", "in", user.partner_id.ids)],
                 ),
                 "counter_bus_id": bus_last_id,
-                "id": "starred",
+                "id": "bookmarkBox",
```
seb-odoo: We might want to keep just `bookmark` for the `id`, `bookmarkBox` suggestion was mainly for the field name itself.

addons/mail/static/src/core/web/mail_core_web_service.js
```diff
@@ -32,8 +32,8 @@ export class MailCoreWeb {
             if (message.needaction && notifId > this.store.inbox.counter_bus_id) {
                 this.store.inbox.counter--;
             }
-            if (message.starred && notifId > this.store.starred.counter_bus_id) {
-                this.store.starred.counter--;
+            if (message.is_bookmarked && notifId > this.store.bookmarkBox.counter_bus_id) {
+                this.store.bookmarkBox.counter--;
```
seb-odoo, line 36: If we're now sending `_update_bookmark_box` result in all flows where it changes, we could also send the value as payload of `mail.message/delete` to simply insert the new counter here instead of doing math.

seb-odoo: I'm pushing other changes, but this can be done as a separate PR.

seb-odoo (approved):
@robodoo r+

seb-odoo:
@robodoo r+

seb-odoo:
@robodoo r+ priority

seb-odoo (approved):
@robodoo r+

addons/mail/models/mail_message.py
```diff
@@ -225,12 +225,15 @@ def default_get(self, fields):
         'mail.notification', 'mail_message_id', 'Notifications',
         bypass_search_access=True, copy=False, depends=['notified_partner_ids'])
     # user interface
-    starred_partner_ids = fields.Many2many(
-        'res.partner', 'mail_message_res_partner_starred_rel', string='Favorited By')
+    bookmarked_partner_ids = fields.Many2many(
```
tde-banana-odoo, line 228: Useless renaming. UI != model, you could avoid migration and conflicts by keeping field names. Always think backward compatibility, conflicts, and ease of use through versions (as discussed notably in guru meetings, if you feel I am exagerating).

alexkuhn: Keeping older name "star" is confusing. Backward compatibility is nice when this is cheap, but at some point we need to ease working with current code, and that means saying goodbye to outdated names.

PR #245901, opened 2026-01-27
[FIX] account_peppol: ensure PDF is embedded in Peppol XML

bhra-odoo:
**Steps to reproduce:**
* Install `Accounting` and `accounting_peppol` modules.
* Create Belgium company.
* Go to accounting settings and activate PEPPOL under `PEPPOL Electronic Invoicing`.
* Create and post a customer invoice.
* Click **Send & Print**.
* Select only **Email** and click **Send** to generate the PDF.
* Click **Send & Print** again.
* Select only **Peppol** (partner configured for Peppol).
* Click **Send**.

**Observed behavior:**
* The invoice is sent via **Peppol**, but the generated UBL XML does **not** contain the embedded PDF.
* The receiver gets the XML data without the PDF representation.

**Cause:**
* When a PDF already exists on the invoice, the send flow skips the document post-processing step to avoid regenerating it.
* The PDF embedding into the UBL XML relies on this post-processing step.
* Sending via Peppol after emailing the invoice therefore bypasses the PDF embedding logic.

**Fix:**
* When sending via **Peppol**, detect if a PDF already exists on the invoice.
* If so, explicitly post-process the UBL XML to embed the existing PDF before sending.
* Ensures the PDF is always included in Peppol messages, regardless of when it was generated.

opw-5495465


addons/account_peppol/models/account_move_send.py
```diff
@@ -199,6 +199,14 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
                     invoice_data['error'] = _("Invoice %s is too big to send via peppol (64MB limit)", invoice.name)
                     continue
 
+                if (
+                    invoice.invoice_pdf_report_id
+                    and not invoice_data.get('error')
```
alsh-odoo:
```suggestion
```

alsh-odoo:
```suggestion
```

Thanks for the fix!

I don’t think this additional check is necessary. Since we already continue when there is an error in invoice_data, the remaining part of the loop is skipped. That means if invoice_data contains any error, this line would never be executed anyway.

So it should be safe to remove this line.

alsh-odoo (changes requested):

alsh-odoo (approved):

agr-odoo (approved):
Looks ok, thanks!

addons/account_peppol/models/account_move_send.py
```diff
@@ -204,6 +204,13 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
                     invoice_data['error'] = _("Invoice %s is too big to send via peppol (64MB limit)", invoice.name)
                     continue
 
+                if (
+                    invoice.invoice_pdf_report_id
+                    and 'ubl_cii_xml_options' in invoice_data
+                    and invoice_data['ubl_cii_xml_options']['ubl_cii_format'] != 'facturx'
+                ):
+                    self._postprocess_invoice_ubl_xml(invoice, invoice_data)
```
RicGR98: Also wondering if we should do this embedding before checking the size (just above) ?

addons/account_peppol/models/account_move_send.py
```diff
@@ -204,6 +204,13 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
                     invoice_data['error'] = _("Invoice %s is too big to send via peppol (64MB limit)", invoice.name)
                     continue
 
+                if (
+                    invoice.invoice_pdf_report_id
+                    and 'ubl_cii_xml_options' in invoice_data
+                    and invoice_data['ubl_cii_xml_options']['ubl_cii_format'] != 'facturx'
+                ):
```
RicGR98: I see that this condition is repeated here

https://github.com/odoo/odoo/blob/38876fa0449883de0a33f0f5ee54d02a69ca144f/addons/account_edi_ubl_cii/models/account_move_send.py#L149

Perhaps it would be worth having an helper method for this condition to avoid repeating it ? So that if this one or the other is modified at some point, then we don't forget to update the other

addons/account_peppol/models/account_move_send.py
RicGR98: Also, shouldn't this be done before the above check. Otherwise if the xml is 63MB, and then we embed a 10MB pdf in it, the sending might fail.

RicGR98 (commented):
Thanks for the PR :))

RicGR98 (approved):
Thanks!

@robodoo r+

PR #245903, opened 2026-01-27
[FIX] l10n_dk: allow CoA installation

mart-e:
When installing a chart of account, previous records are deleted to be replaced by the one of the installed chart.
Since odoo/odoo@4400130eb32, we need to explicitly check for force_delete flag.

Use the renamed method _read_group

### DIFF
--- a/addons/l10n_dk/models/account_account.py
+++ b/addons/l10n_dk/models/account_account.py
@@ -10,6 +10,8 @@ class AccountAccount(models.Model):
     
     @api.ondelete(at_uninstall=False)
     def _unlink_bank_cash_accounts(self):
+        if self.env.context.get('force_delete'):
+            return
         nb_account_to_delete_per_company = defaultdict(self.env['account.account'].browse)
         for account in self:
             for company in account.company_ids:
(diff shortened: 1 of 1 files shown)


william-andre (approved):
@robodoo r+

addons/l10n_dk/models/account_account.py
```diff
@@ -19,9 +21,9 @@ def _unlink_bank_cash_accounts(self):
         if not nb_account_to_delete_per_company:
             return
 
-        grouped_counts = self.read_group(
+        grouped_counts = self._read_group(
             domain=[('company_ids.account_fiscal_country_id.code', '=', 'DK'), ('account_type', '=', 'asset_cash')],
-            fields=['company_ids', 'id:count'],
+            aggregates=['company_ids', 'id:count'],
```
aamo-odoo, line 26: Hello @mart-e 

There is another error that will occur after your fix:

`ValueError: Aggregate method is mandatory for 'company_ids'`

This error occurs because we are using `_read_group()` with an aggregation on `company_ids`, but we haven't specified an aggregate function for it.

I already have a [PR](https://github.com/odoo/odoo/pull/245183) for this. If it looks fine to you, I’ll take care of the conflicts and we’ll be good to move forward.

Thanks

mart-e: @aamo-odoo the main intent of my PR was to add the context to avoid triggering this at the installation of a CoA. The change of `read_group` was trying to have a quick fix as well but apparently it was not enough. Feel free to proceed with your fix.

PR #245904, opened 2026-01-27
[FIX] website: translate links inline on media replacement

robinlej:
[FIX] website: translate links inline on media replacement
    
    On the website builder, files added through the `/file` command would go
    through the domPlugin `insert` method, which would call
    `before_insert_processors` and apply `.o_translate_inline` as expected.
    But there is another flow to add a document on the page: replace an
    image (or video, or icon), then select the "Documents" tab and upload
    a file. With this flow, `insert` is not called, so we have to add the
    class through some other resource.
    
[FIX] html_editor: target documents with right class
    
    The class `.o_image` was still associated with documents (in the context
    of the file selector) and the expected tag of a document was `A`, in
    spite of it not being true in `html_editor` since the introduction of
    the file box in [1].
    
    This has also been updated in the website builder since the introduction
    of the `html_builder` module in 18.4, which swapped uses of `web_editor`
    components for `html_editor`.
    
    As a side-effect, in website, a double click on a file did not open the
    media dialog on the "document" tab, unlike other media (images, videos,
    icons). In such a case, the file was also not shown as already selected
    in the media dialog (because it targetted the wrong tag name). Both of
    those behaviors were lost as we used the new file box design in 18.4.
    
    [1]: https://github.com/odoo/odoo/commit/7f9afa21dffba2f74f9fb8a68809db4af6c7c225
    
task-5876278

### DIFF
--- a/addons/html_editor/static/src/main/media/media_dialog/media_dialog.js
+++ b/addons/html_editor/static/src/main/media/media_dialog/media_dialog.js
@@ -91,9 +91,15 @@ export class MediaDialog extends Component {
             return this.props.activeTab;
         }
         if (this.props.media) {
-            const correspondingTab = Object.keys(this.tabs).find((id) =>
-                this.tabs[id].Component.tagNames.includes(this.props.media.tagName)
-            );
+            const correspondingTab =
+                Object.keys(this.tabs).find((id) =>
+                    this.tabs[id].Component.mediaSpecificClasses.some((cls) =>
+                        [...this.props.media.classList].includes(cls)
+                    )
+                ) ||
+                Object.keys(this.tabs).find((id) =>
+                    this.tabs[id].Component.tagNames.includes(this.props.media.tagName)
+                );
             if (correspondingTab) {
                 return correspondingTab;
             }
--- a/addons/html_editor/static/tests/delete/backward.test.js
+++ b/addons/html_editor/static/tests/delete/backward.test.js
@@ -407,21 +407,27 @@ describe("Selection collapsed", () => {
 
         test("should unwrap a block next to an inline unbreakable element", async () => {
             await testEditor({
-                contentBefore: `<div><p>abc</p><div class="o_image"></div><p>[]def</p></div>`,
+                contentBefore: `<div><p>abc</p><span class="oe_unbreakable"></span><p>[]def</p></div>`,
                 stepFunction: async (editor) => {
                     deleteBackward(editor);
                 },
-                contentAfter: `<div><p>abc</p><div class="o_image"></div>[]def</div>`,
+                // After the deleteBackward, the automatic normalization of
+                // setSelection puts the cursor at the deepest position, that is
+                // in the span, rather than after it. While it might look a bit
+                // weird in this case, it makes more sense if one imagine that
+                // it is a style tag like bold or italic.
+                contentAfter: `<div><p>abc</p><span class="oe_unbreakable">[]</span>def</div>`,
             });
         });
 
         test("should remove an inline unbreakable contenteditable='false' sibling element", async () => {
             await testEditor({
-                contentBefore: `<div><p>abc</p><div class="o_image"></div>[]def</div>`,
+                contentBefore: `<div><p>abc</p><span class="oe_unbreakable" contenteditable="false">d</span>[]efg</div>`,
+                contentBeforeEdit: `<div><p>abc</p><span class="oe_unbreakable" contenteditable="false">d</span>[]efg</div>`,
                 stepFunction: async (editor) => {
                     deleteBackward(editor);
                 },
-                contentAfter: `<div><p>abc</p>[]def</div>`,
+                contentAfter: `<div><p>abc</p>[]efg</div>`,
             });
         });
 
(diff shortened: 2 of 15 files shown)


addons/html_editor/static/src/main/media/media_dialog/media_dialog.js
```diff
@@ -90,9 +90,15 @@ export class MediaDialog extends Component {
...
-            );
+            const correspondingTab =
+                Object.keys(this.tabs).find((id) =>
+                    this.tabs[id].Component.mediaSpecificClasses.some((cls) =>
+                        [...this.props.media.classList].includes(cls)
+                    )
+                ) ||
+                Object.keys(this.tabs).find((id) =>
+                    this.tabs[id].Component.tagNames.includes(this.props.media.tagName)
+                );
```
FrancoisGe, line 102: We should add a test

addons/html_editor/static/tests/delete/backward.test.js
```diff
@@ -407,17 +407,17 @@ describe("Selection collapsed", () => {
...
                 },
-                contentAfter: `<div><p>abc</p><div class="o_image"></div>[]def</div>`,
+                contentAfter: `<div><p>abc</p><div class="o_file_box"></div>[]def</div>`,
             });
         });
 
         test("should remove an inline unbreakable contenteditable='false' sibling element", async () => {
             await testEditor({
-                contentBefore: `<div><p>abc</p><div class="o_image"></div>[]def</div>`,
+                contentBefore: `<div><p>abc</p><div class="o_file_box"></div>[]def</div>`,
```
dmo-odoo: I don't think these tests should have been changed as they do not seem to depict a file box. To be fair, they didn't seem to depict a media either. They are in fact quite strange. If they wanted an inline unbreakable they could just have done something like `<span class="oe_unbreakable"></span>`.

addons/html_editor/static/tests/delete/backward.test.js
```diff
@@ -495,32 +495,32 @@ describe("Selection collapsed", () => {
 
         test("should merge paragraph with previous one containing a media element", async () => {
             await testEditor({
-                contentBefore: `<p>abc</p><p style="margin-bottom: 0px;"><o-image class="o_image" contenteditable="false"></o-image></p><p>[]def</p>`,
+                contentBefore: `<p>abc</p><p style="margin-bottom: 0px;"><o-file-box class="o_file_box" contenteditable="false"></o-file-box></p><p>[]def</p>`,
                 stepFunction: deleteBackward,
-                contentAfterEdit: `<p>abc</p><p style="margin-bottom: 0px;"><o-image class="o_image" contenteditable="false"></o-image>[]def</p>`,
-                contentAfter: `<p>abc</p><p style="margin-bottom: 0px;"><o-image class="o_image"></o-image>[]def</p>`,
+                contentAfterEdit: `<p>abc</p><p style="margin-bottom: 0px;"><o-file-box class="o_file_box" contenteditable="false"></o-file-box>[]def</p>`,
+                contentAfter: `<p>abc</p><p style="margin-bottom: 0px;"><o-file-box class="o_file_box"></o-file-box>[]def</p>`,
```
dmo-odoo: Similar here and the rest of the file. It's a not a file box and was barely a media to begin with. I guess maybe the tests were written that way so that it was not necessary to have an actual src for the image? In any case, I think this test and the others below in this file should not have been changed.

addons/html_editor/static/tests/delete/forward.test.js
```diff
@@ -350,15 +350,15 @@ describe("Selection collapsed", () => {
 
         test("should remove a link to uploaded document", async () => {
             await testEditor({
-                contentBefore: `<p>[]<a href="#" title="document" data-mimetype="application/pdf" class="o_image" contenteditable="false"></a>abc</p>`,
+                contentBefore: `<p>[]<span class="o_file_box" contenteditable="false"><a href="#" title="document" data-mimetype="application/pdf"></a></span>abc</p>`,
```
dmo-odoo, line 353: Nor these two.

addons/html_editor/static/tests/initialization.test.js
```diff
@@ -98,12 +98,12 @@ describe("No orphan inline elements compatibility mode", () => {
         });
     });
 
-    test("should wrap a div.o_image direct child of the editable into a block", async () => {
+    test("should wrap a div.o_file_box direct child of the editable into a block", async () => {
         await testEditor({
-            contentBefore: '<p>abc</p><div class="o_image"></div><p>def</p>',
+            contentBefore: '<p>abc</p><div class="o_file_box"></div><p>def</p>',
```
dmo-odoo: This one should probably not have changed either.

dmo-odoo (commented):
Hello! Sorry for the late reply. I was on holiday and your message was drowned in my mailbox.

The filebox do have the \ufeff so that's expected. However, I think some of the tests should not have been changed because they do not seem to depict file boxes, I've marked them in separate comments. Hopefully `o_image` still means something or `o_media`? If not (I see you removed `o_image` from the selector constant), then those tests will need to be adapted to have actual images then, I guess. Or maybe icons? The point is having a media, so basically any media will do.

For the hint, it is meant to be displayed only in empty containers, so it should not be displayed when there is a file box.

Thanks!

FrancoisGe:
@robodoo r+ rebase-ff

PR #245905, opened 2026-01-27
[FIX] mrp_subcontracting: use correct location for return for exchange

qucol-odoo:
## Issue
When making a request for quotation for a subcontracted product and returning the delivery "for exchange", the new incoming delivery does not have the correct destination. Instead of having the stock of the user, the destination of the new incoming delivery is the same as its source: the subcontracting location.

## Steps to reproduce
1. Install MRP Subcontracting (`mrp_subcontracting`) and Purchase (`purchase`)
2. In Settings, enable *Subcontracting*
3. Create a Product P and a subcontracting BoM with Subcontractor S
4. Create a Request for Quotation
    - Vendor: Subcontractor S
    - Product: Product P (any quantity > 0)
5. Confirm the RFQ, receive the PO, validate the picking
6. On the validated picking, click *Return*, set the quantity of products to return, and click *Return for Exchange*
    - This creates two new pickings, one to return the product(s) we received, and one to receive new products
7. Validate the two new pickings
8. **In Inventory > Reporting > Moves History, the very last `stock.move.line` has the same location in the *From* (`location_id`) and the *To* (`location_dest_id`) columns**

## Cause
The `location_dest_id` of the new `stock.move` is updated in `StockReturnPickingLine._prepare_move_default_values`.

https://github.com/odoo/odoo/blob/fb534f1eadcb8ef74e2ee6fd5b68872dddb978e3/addons/mrp_subcontracting/wizard/stock_picking_return.py#L20-L25

The condition added by https://github.com/odoo/odoo/commit/5404b426aac9 sets the destination of all returned subcontracted moves to the subcontractor location. This is incorrect when using "return for exchange", as in this case, the return move is directed towards the user's stock. In fact, when using "return for exchange", the following pickings are created:

| id | name         | return_id |   |
|:--:|--------------|:---------:|---|
| 1  | WH/IN/00001  |           | Initial RFQ delivery |
| 2  | WH/OUT/00001 |     1     | Return of the initial RFQ delivery |
| 3  | WH/IN/00002  |     2     | New products delivery to replace the initial delivery. The stock.move.line of this stock.picking has a wrong `location_dest_id` |

## Fix
In the context of return for exchanges, the returned item must be directed to the *Subcontracting Location* while the new item must be directed to   the *Stock*. In the `_prepare_move_default_values`, we should only set   the `location_dest_it` to the subcontractor location for outgoing pickings.

opw-5479900

### DIFF
--- a/addons/mrp_subcontracting/tests/test_subcontracting.py
+++ b/addons/mrp_subcontracting/tests/test_subcontracting.py
@@ -1284,6 +1284,54 @@ class TestSubcontractingFlows(TestMrpSubcontractingCommon):
             {'product_id': self.comp1.id, 'location_id': stock_location.id, 'location_dest_id': supplier_location.id}
         ])
 
+    def test_subcontracting_order_returned_for_exchange(self):
+        """
+        Test the locations used when returning a receipt for exchange from a subcontractor
+        """
+        custom_subcontract_location = self.env['stock.location'].create({
+            'name': 'Custom Subcontractor Location',
+            'location_id': self.env.company.subcontracting_location_id.id,
+            'usage': 'internal',
+            'company_id': self.env.company.id,
+            'is_subcontracting_location': True,
+        })
+        self.subcontractor_partner1.property_stock_subcontractor = custom_subcontract_location.id
+        initial_receipt = self.env['stock.picking'].create({
+            'picking_type_id': self.ref('stock.picking_type_in'),
+            'partner_id': self.subcontractor_partner1.id,
+            'location_id': self.ref('stock.stock_location_suppliers'),
+            'location_dest_id': self.warehouse.lot_stock_id.id,
+            'move_ids': [
+                Command.create({
+                    'name': self.finished.name,
+                    'product_id': self.finished.id,
+                })
+            ]
+        })
+        initial_receipt.action_confirm()
+        initial_receipt.move_ids.quantity = 1
+        initial_receipt.button_validate()
+        return_wizard = Form(
+            self.env['stock.return.picking'].with_context(active_id=initial_receipt.id, active_model='stock.picking')
+        ).save()
+        return_wizard.product_return_moves.quantity = 1
+        return_picking = Form.from_action(self.env, return_wizard.action_create_exchanges()).save()
+        exchange_picking = return_picking.return_ids
+        self.assertRecordValues(initial_receipt.move_ids, [{
+            'location_id': self.subcontractor_partner1.property_stock_subcontractor.id,
+            'location_dest_id': self.warehouse.lot_stock_id.id,
+        }])
+        self.assertEqual(return_picking.return_id, initial_receipt)
+        self.assertRecordValues(return_picking.move_ids, [{
+            'location_id': self.warehouse.lot_stock_id.id,
+            'location_dest_id': self.subcontractor_partner1.property_stock_subcontractor.id,
+        }])
+        self.assertEqual(exchange_picking.return_id.return_id, initial_receipt)
+        self.assertRecordValues(exchange_picking.move_ids, [{
+            'location_id': self.subcontractor_partner1.property_stock_subcontractor.id,
+            'location_dest_id': self.warehouse.lot_stock_id.id,
+        }])
+
 
 @tagged('post_install', '-at_install')
 class TestSubcontractingTracking(TransactionCase):
--- a/addons/mrp_subcontracting/wizard/stock_picking_return.py
+++ b/addons/mrp_subcontracting/wizard/stock_picking_return.py
@@ -19,7 +19,7 @@ class StockReturnPickingLine(models.TransientModel):
 
     def _prepare_move_default_values(self, new_picking):
         vals = super()._prepare_move_default_values(new_picking)
-        if self.move_id.is_subcontract:
+        if self.move_id.is_subcontract and new_picking.picking_type_code == 'outgoing':
             vals['location_dest_id'] = new_picking.partner_id.with_company(new_picking.company_id).property_stock_subcontractor.id
         vals['is_subcontract'] = False
         return vals


addons/mrp_subcontracting/wizard/stock_picking_return.py
```diff
@@ -19,7 +19,7 @@ class StockReturnPickingLine(models.TransientModel):
 
     def _prepare_move_default_values(self, new_picking):
         vals = super()._prepare_move_default_values(new_picking)
-        if self.move_id.is_subcontract:
+        if self.move_id.is_subcontract and not new_picking.return_id.return_id:
```
lase-odoo: While I agree that we should not add the `location_dest_id` in the vals in your use case the fact that the return of your return is set is not  an accurate data to determine what part of the chain you are (the part that should or should not target the subcontracting location).  For instance returning for exchange once more the last receipt from your subcontractor should have the same behavior as the first time you did it, but, this time since the return change will have length 3 and 4, the `new_picking.return_id.return_id` will be set for both the receipt and the delivery. 

IMO, the fix of https://github.com/odoo/odoo/commit/5404b426aac9 looks correct, but the value of the `is_subcontract` field is not for the move of your flow. To be more precise, looking at the description of the field it should only be True on subcontracted receipt:
https://github.com/odoo/odoo/blob/c2cd08ab199202ea592199a691baf97b50bd6d4c/addons/mrp_subcontracting/models/stock_move.py#L15
IMO, the fix is rather to ensure that the `is_subcontract` field is only set to true for subcontracted receipt move.

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1284,6 +1284,47 @@ def test_subcontracted_product_return_locations(self):
             {'product_id': self.comp1.id, 'location_id': stock_location.id, 'location_dest_id': supplier_location.id}
         ])
 
+    def test_subcontracting_order_returned_for_exchange(self):
+        """
+        Test that the correct locations are used when returning a picking order for exchange
+        """
+        initial_picking = self.env['stock.picking'].create({
+            'picking_type_id': self.env.ref('stock.picking_type_in').id,
```
lase-odoo: You can use `self.ref` to return the `id` dirrectly.

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1284,6 +1284,47 @@ def test_subcontracted_product_return_locations(self):
             {'product_id': self.comp1.id, 'location_id': stock_location.id, 'location_dest_id': supplier_location.id}
         ])
 
+    def test_subcontracting_order_returned_for_exchange(self):
+        """
+        Test that the correct locations are used when returning a picking order for exchange
```
lase-odoo:
```suggestion
        Test the locations used when returning a receipt for exchange from a subcontractor
``

File: addons/mrp_subcontracting/tests/test_subcontracting.py [hunk 4]
```diff
@@ -1284,6 +1284,47 @@ def test_subcontracted_product_return_locations(self):
...
+        """
+        initial_picking = self.env['stock.picking'].create({
```
lase-odoo: I think it is a good idea for the name to indicate as much info as possible (a picking is a generic object that can go in every direction):
```        suggestion
        initial_receipt = self.env['stock.picking'].create({
```

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1284,6 +1284,47 @@ def test_subcontracted_product_return_locations(self):
...
+            ]
+        })
+        initial_picking.action_confirm()
+        initial_picking.move_ids.quantity = 1
+        initial_picking.button_validate()
+        return_wizard = Form(
+            self.env['stock.return.picking'].with_context(active_id=initial_picking.id, active_model='stock.picking')
+        ).save()
+        return_wizard.product_return_moves.quantity = 1
+        res = return_wizard.action_create_exchanges()
```
lase-odoo: Looking at the code, it looks like the action returned by the `action_create_exchanges` is a form view. I think it cleaner to return the record as is with a 
```suggestion
        return_picking = Form.from_action(self.env, return_wizard.action_create_exchanges()).save()
```

lase-odoo (commented):

While fix you propose would solve the current issue, I think it has some limitations and it looks like a patch of the consequences of the issue rather than its cause. I left a suggestion of an other fixing axis, let me know what you think !

addons/mrp_subcontracting/wizard/stock_picking_return.py
qucol-odoo: Hey @lase-odoo  :wave:

Thanks for the review, and sorry for the long delay :sweat:

I agree that the `.return_id.return_id` is not ideal, I am proposing a new solution now, using the `picking_type_code` of the `new_picking`.

Regarding your proposition to work around the `is_subcontract` field, I'm not sure it would be possible/make sense in this case. The `_prepare_move_default_values` is called twice when returning for exchange: once to send the product back to the subcontractor, and a second time to receive the new product from the subcontractor. In both cases, the `move_id` is the same, which is the one from the original PO. IMO, updating the `is_subcontract` field of that stock move would not make sense, as that move is still subcontracted.

lase-odoo: Indeed good catch ! I did not realize that that the move was the same on both trigger of the `_prepare_move_default_values` call (and was indeed a subcontracted receipt) but that the `new_picking` was on the incoming/outgoing one !

addons/mrp_subcontracting/wizard/stock_picking_return.py
```diff
@@ -19,7 +19,7 @@ class StockReturnPickingLine(models.TransientModel):
 
     def _prepare_move_default_values(self, new_picking):
         vals = super()._prepare_move_default_values(new_picking)
-        if self.move_id.is_subcontract:
+        if self.move_id.is_subcontract and new_picking.picking_type_code != 'incoming':
```
lase-odoo, line 22: I think that using the `picking_type_code` is a good call but I am wondering if it would not be best to only force the `location_dest_id` for `outgoing` rather than `outgoing` and `internal`:
https://github.com/odoo/odoo/blob/4252d2d37919714780a183a9dd932349f0d63e44/addons/stock/models/stock_picking.py#L636-L638
https://github.com/odoo/odoo/blob/4252d2d37919714780a183a9dd932349f0d63e44/addons/stock/models/stock_picking.py#L43
Note that I think it has some importance since technically the subcontracting locations are considered to have an `internal` usage:
https://github.com/odoo/odoo/blob/4252d2d37919714780a183a9dd932349f0d63e44/addons/mrp_subcontracting/models/res_company.py#L25-L27
So that I would not be surprised to use an internal transfer picking type to move products from one subcontracting location to the next and I do not think it would be correct to force the `location_dest_id` of this transfer either !

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1284,6 +1284,46 @@ def test_subcontracted_product_return_locations(self):
...
+        self.assertEqual(return_picking.return_id, initial_receipt)
+        self.assertRecordValues(return_picking.move_ids, [{
+            'location_id': self.warehouse.lot_stock_id.id,
+            'location_dest_id': self.subcontractor_partner1.property_stock_subcontractor.id,
+        }])
+        self.assertEqual(exchange_picking.return_id.return_id, initial_receipt)
+        self.assertRecordValues(exchange_picking.move_ids, [{
+            'location_id': self.subcontractor_partner1.property_stock_subcontractor.id,
+            'location_dest_id': self.warehouse.lot_stock_id.id,
+        }])
```
lase-odoo, line 1325: The test is almost perfect to me. I just have one small nitpick, currently, the "specific" `property_stock_subcontractor` location used for `subcontractor_partner1` is the default one that is the generic subcontracting location `self.env.company.subcontracting_location_id`. While your assert are true they would also be if the location dest we forced in case of subcontrating was the "generic" subcontracting location of the company. So that I think it would be valuable to create a `stock.location` that is a child location of `self.env.company.subcontracting_location_id` and to set it as "specific" `property_stock_subcontractor` for the subcontractor of your flow.

lase-odoo (commented):
Thanks a lot for your excellent work 🐝

I left some small improvement proposition.

qucol-odoo:
Hi @lase-odoo :wave:
I applied your suggestions:
- The location is now forced only for `outgoing` moves
- The test now uses a custom location instead of the generic subcontracting location of the company

lase-odoo:
Thanks a lot 🌊

robodoo r+

lase-odoo:
@robodoo delegate=@qucol-odoo

qucol-odoo:
@robodoo r+

PR #245906, opened 2026-01-27
[FIX] website_slides: add route to download slides documents

lrizzi1:
Currently if you have a cdn configured to the route '/web/content' and you try to download files in a e-learning course as a portal user, you won't be able to. You get a not found  error instead since you are not logged in in the cdn server, so no access to the resource.

Given that configuring a cdn to this route is in our documentation and that other contents on that route are expected to be public (so available trough cdn) this is unexpected behaviour. This adds a route to slides that calls the same method as the old one, so that you can cdn that route and this will still work

opw-4918546


Julien00859 (changes requested):
The bug is in the CDN, not Odoo

PR #245912, opened 2026-01-27
[FIX] account: payment widget title hover

hupo-odoo:
the forward port https://github.com/odoo/odoo/commit/2490d32f2a28cc05c7cd5d8450701232aede2c29 introduced some minor issues in the payment widget on the
account move form view, such as the title on hover. When
viewing a Bill, for example, with a possible match on a bank
statement, on hovering, we should see both the ref and the line name.

This commit also removes useless added classes in the template.

Finally, when viewing the reconciled widget (click on the
info popup beside the "Paid on ..."), the "Journal" section
was missing.

---

### DIFF
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -1476,7 +1476,7 @@ class AccountMove(models.Model):
                     'move_id': line.move_id.id,
                     'date': fields.Date.to_string(line.date),
                     'account_payment_id': line.payment_id.id,
-                    'move_ref': line.ref or "",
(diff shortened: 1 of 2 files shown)


oco-odoo (changes requested):
Please make the commit message a little more explicit

hupo-odoo:
> Please make the commit message a little more explicit

Indeed you were right, it was very clear. I made it a bit clearer :smiley:

oco-odoo (approved):
@robodoo r+

PR #245915, opened 2026-01-27
[IMP] account_peppol, account_edi_ubl_cii: Make the error message more accurate when sending via Peppol.

jugurtha-gaci:
When a user sends a move via Peppol to a customer that has a VAT number set but not a Peppol endpoint, we show the user a generic error ("no VAT").

This makes the user confused, as he already filled the VAT field of his customer, It's the Peppol VAT that is missing (it could be: Belgian Company Registry, France SIRET, ...etc, depending on the customer's country)

This PR makes the error message more accurate by showing exactly the missing required field.

task-5499707

---

### DIFF
--- a/addons/account_peppol/wizard/account_move_send_wizard.py
+++ b/addons/account_peppol/wizard/account_move_send_wizard.py
@@ -15,6 +15,9 @@ class AccountMoveSendWizard(models.TransientModel):
         """
         for wizard in self:
             peppol_partner = wizard.move_id.partner_id.commercial_partner_id.with_company(wizard.company_id)
+            if not peppol_partner.peppol_eas or not peppol_partner.peppol_endpoint:
+                peppol_partner._compute_peppol_endpoint()  # Try to recompute the Peppol credentials.
+            eas_label = dict(peppol_partner._fields['peppol_eas']._description_selection(self.env)).get(peppol_partner.peppol_eas)
             peppol_partner.button_account_peppol_check_partner_endpoint(company=wizard.company_id)
         super()._compute_sending_method_checkboxes()
         for wizard in self:
@@ -24,7 +27,16 @@ class AccountMoveSendWizard(models.TransientModel):
                 if peppol_partner.peppol_verification_state == 'not_valid':
                     addendum_disable_reason = _(' (Customer not on Peppol)')
                 elif peppol_partner.peppol_verification_state == 'not_verified':
-                    addendum_disable_reason = _(' (no VAT)')
+                    # The recomputation of the Peppol credentials did not manage to fill these fields.
+                    addendum_disable_reason = _(' (Customer not on Peppol)')
+                    if not peppol_partner.peppol_eas or not peppol_partner.peppol_endpoint:
+                        if not peppol_partner.vat:
+                            addendum_disable_reason = _(' (no VAT)')
+                        elif eas_label:
+                            addendum_disable_reason = _(
+                                ' (Missing %(eas)s)',
+                                eas=eas_label
+                            )
                 else:
                     addendum_disable_reason = ''
                 vals_not_valid = {'readonly': True, 'checked': False} if addendum_disable_reason else {}


addons/account_edi_ubl_cii/models/res_partner.py
```diff
@@ -244,6 +245,13 @@ def _compute_peppol_endpoint(self):
                 if field and value and not partner._build_error_peppol_endpoint(partner.peppol_eas, value):
                     partner.peppol_endpoint = value
 
+    @api.depends('peppol_eas')
+    def _compute_peppol_eas_label(self):
+        """Compute the readable label of the selected PEPPOL EAS"""
+        available_peppol_eas = dict(self._fields['peppol_eas'].selection)
```
dylankiss:
```suggestion
        available_peppol_eas = dict(self._fields['peppol_eas']._description_selection(self.env))
```
to get the translated values

addons/account_peppol/i18n/account_peppol.pot
```diff
@@ -36,7 +36,7 @@ msgstr ""
 #. module: account_peppol
 #. odoo-python
 #: code:addons/account_peppol/wizard/account_move_send_wizard.py:0
-msgid " (no VAT)"
+msgid " (no %(peppol_eas_label)s)"
```
dylankiss: Not needed, since we have an automatic export of all POT files every weekend.

addons/account_peppol/wizard/account_move_send_wizard.py
```diff
@@ -24,7 +24,7 @@ def _compute_sending_method_checkboxes(self):
                 if peppol_partner.peppol_verification_state == 'not_valid':
                     addendum_disable_reason = _(' (Customer not on Peppol)')
                 elif peppol_partner.peppol_verification_state == 'not_verified':
-                    addendum_disable_reason = _(' (no VAT)')
+                    addendum_disable_reason = _(' (no %(peppol_eas_label)s)', peppol_eas_label=peppol_partner.peppol_eas_label or "VAT on the Peppol settings")
```
dylankiss: Translations-wise this is not correct. You are including an English static string in a translation. But I also don't know if you want what you did here.

It will show e.g.:
`(no Belgian VAT)`
in case there is a `peppol_eas_label`, but
`(no VAT on the Peppol settings)`
if there is none.

Don't you want the part "on the Peppol settings" to always show? Not sure if the English is fully correct either.

addons/account_edi_ubl_cii/models/res_partner.py
jugurtha-gaci: Done ! Thank you

addons/account_peppol/i18n/account_peppol.pot
jugurtha-gaci: Understood ! Thank you

addons/account_peppol/wizard/account_move_send_wizard.py
jugurtha-gaci: I corrected it, thank you for your guidance.

Regarding your question : Don't you want the part "on the Peppol settings" to always show? 
No, it only shows as a fallback when the EAS is not filled.
And I changed the fallback message to to keep it clear and consistent.

addons/account_peppol/wizard/account_move_send_wizard.py
```diff
@@ -24,7 +26,20 @@ def _compute_sending_method_checkboxes(self):
...
+                        else:
+                            if peppol_partner.peppol_eas_label:
+                                addendum_disable_reason = _(
+                                    ' (Missing %(eas)s)',
+                                    eas=peppol_partner.peppol_eas_label
+                                )
+                            else:
+                                addendum_disable_reason = _(' (Customer not on Peppol)')
+                    else:
+                        addendum_disable_reason = _(' (Customer not on Peppol)')
```
clbr-odoo:
```suggestion
                    # The recomputation of the Peppol credentials did not manage to fill these fields.
                    addendum_disable_reason = _(' (Customer not on Peppol)')
                    if not peppol_partner.peppol_eas or not peppol_partner.peppol_endpoint:
                        if not peppol_partner.vat:
                            addendum_disable_reason = _(' (no VAT)')
                        elif peppol_partner.peppol_eas_label:
                            addendum_disable_reason = _(
                                ' (Missing %(eas)s)',
                                eas=peppol_partner.peppol_eas_label
                            )
```

PR #245916, opened 2026-01-27
[IMP] *: remove datas

kmagusiak:
Remove `ir.attachment.datas` field. We can use `raw.to_base64()` if needed.

Since the introduction of `BinaryValue`, the field `datas` which is the base64 computed version of `raw` is not necessary. We remove it as a cleanup of binary fields. This makes sure that developers don't do unnecessary conversions to base64.

Follow-up PR to odoo/odoo#244421.

odoo/enterprise#105663
odoo/upgrade#9354
task-4251301

---

### DIFF
--- a/addons/mail/views/mail_message_views.xml
+++ b/addons/mail/views/mail_message_views.xml
@@ -143,7 +143,7 @@
                     <field name="type"/>
                     <templates>
                         <t t-name="menu">
-                            <a t-attf-href="/web/content/ir.attachment/#{record.id.raw_value}/datas?download=true" download="" class="dropdown-item">Download</a>
+                            <a t-attf-href="/web/content/#{record.id.raw_value}?download=true" download="" class="dropdown-item">Download</a>
                             <a t-if="widget.deletable" role="menuitem" type="delete" class="dropdown-item">Delete</a>
                         </t>
                         <t t-name="card" class="o_kanban_attachment flex-row">
--- a/addons/web/tests/test_ir_qweb.py
+++ b/addons/web/tests/test_ir_qweb.py
@@ -1,8 +1,11 @@
 import base64
+
 from lxml import etree
 
-from odoo.tests.common import tagged, TransactionCase
+from odoo.tests.common import TransactionCase, tagged
 from odoo.tools.mimetypes import guess_mimetype
+
+from odoo.addons.base.tests.files import JPG_RAW
 
 
 @tagged('at_install', '-post_install')  # LEGACY at_install
@@ -61,7 +64,7 @@ class TestIrQweb(TransactionCase):
             "name": "webpcopy.jpg",
             "res_model": "ir.attachment",
             "res_id": attachment.id,
-            "datas": "iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAAF0lEQVR4nGJxKFrEwMDAxAAGgAAAAP//D+IBWx9K7TUAAAAASUVORK5CYII=",
+            "raw": JPG_RAW,
         })
 
         html = view.with_context(webp_as_jpg=False)._render_template(view.id, {"is_raw_image": True, "record": lang_record})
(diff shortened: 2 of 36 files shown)


addons/mail/views/mail_message_views.xml
```diff
@@ -143,7 +143,7 @@
                     <field name="type"/>
                     <templates>
                         <t t-name="menu">
-                            <a t-attf-href="/web/content/ir.attachment/#{record.id.raw_value}/datas?download=true" download="" class="dropdown-item">Download</a>
+                            <a t-attf-href="/web/content/ir.attachment/#{record.id.raw_value}/raw?download=true" download="" class="dropdown-item">Download</a>
```
Julien00859: ir.attachment and raw are already the default model and field
```suggestion
                            <a t-attf-href="/web/content/#{record.id.raw_value}?download=true" download="" class="dropdown-item">Download</a>
```

addons/web/tests/test_ir_qweb.py
```diff
@@ -61,7 +61,7 @@ def test_image_field_webp(self):
             "name": "webpcopy.jpg",
             "res_model": "ir.attachment",
             "res_id": attachment.id,
-            "datas": "iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAAF0lEQVR4nGJxKFrEwMDAxAAGgAAAAP//D+IBWx9K7TUAAAAASUVORK5CYII=",
+            "raw": "iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAAF0lEQVR4nGJxKFrEwMDAxAAGgAAAAP//D+IBWx9K7TUAAAAASUVORK5CYII=",
```
Julien00859: The filename says `.jpg` but the content is a PNG file.

simoerkinheimo:
Hi,

I'm the Product Manager of [ATR CustomTools](https://atrsoft.com/customtools), which comes with the world's leading [SOLIDWORKS - Odoo ERP integration](https://taskpane.customtools.info/en/2026/help/#t=root%2Fhtml%2FIntegrations%2FOdooV2.html).

Sorry for jumping to comment section of "random" issue, but it just so happens that this change causes some headache for us.

What's the suggested approach for External API integrations considering breaking changes like this? I've been following [this documentation](https://www.odoo.com/documentation/master/developer/reference/external_api.html) trying to stay ahead of upcoming changes, but obviously this one went under the radar.

From what I can tell after quick testing, **raw** doesn't work correctly at least in v19.0: Attachments are broken after upload. But it seems to work as expected in 19.3.

Since our customers use various versions of Odoo, we must also be able to support various versions. I can of couse fix this case by implementing version specific routing (**datas** by default, **raw** from 19.3 onwards), but the issue remains that I'd like to stay ahead of this type of changes before things break.

Please advice. Thank you!

Best regards,
Simo Erkiheimo
simo.erkinheimo@atrsoft.com

kmagusiak:
Hello @simoerkinheimo 

Indeed the `raw` field did not work correctly over RPC in previous versions. Since saas versions are not deployed outside of Odoo, we do all the changes before 20.0.

We are going to change the representation of binary fields in https://github.com/odoo/odoo/pull/266082 (to be merged soon) which is much more aligned with what is usually needed in the client.
Then, you should use `datas` in pre-20.0 versions and `raw` afterwards.

Best regards

simoerkinheimo:
Thanks for the reply @kmagusiak!

> Then, you should use `datas` in pre-20.0 versions and `raw` afterwards.

You meant *"use `datas` in **pre-19.3** versions and `raw` afterwards"*, right? 

Or, is there a difference in  `saas~19.3+e` and on-premises version of 19.3? Because this change is definitely already effective in `saas~19.3+e`, so I assumed it applies to also on-premises 19.3.

I'll check that #266082, thanks for the head's up.

-Simo

kmagusiak:
> You meant _"use `datas` in **pre-19.3** versions and `raw` afterwards"_, right?
> 
> Or, is there a difference in `saas~19.3+e` and on-premises version of 19.3? Because this change is definitely already effective in `saas~19.3+e`, so I assumed it applies to also on-premises 19.3.

Right. But on-premise saas versions are not supported [(as far as I know)](https://www.odoo.com/documentation/19.0/administration/standard_extended_support.html).

simoerkinheimo:
Ah, ok. I was a bit confused between different Odoo flavors.

So what's the correct approach for us, if we'd want to test against new saas version release before it goes public?

kmagusiak:
> So what's the correct approach for us, if we'd want to test against new saas version release before it goes public?

You can test against master.

PR #245919, opened 2026-01-27
[FIX] registry: lock registry while loading

kmagusiak:
This commit ensures that a registry can only be updated by one process at a time. It solves potential registry/cache inconsistency while a module is being installed.

Implementation detail in commit message.
TLDR: use a shared lock when loading registry and an exclusive lock when updating it.

---

### DIFF
--- a/odoo/addons/base/models/ir_module.py
+++ b/odoo/addons/base/models/ir_module.py
@@ -618,13 +618,14 @@ class IrModuleModule(models.Model):
                 "for help."
             )
 
-        # raise error if database is updating for module operations
-        if self.search_count([('state', 'in', ('to install', 'to upgrade', 'to remove'))], limit=1):
-            raise UserError(_("Odoo is currently processing another module operation.\n"
-                               "Please try again later or contact your system administrator."))
+        self.env.cr.execute("SET LOCAL lock_timeout = '3s'")
+
         try:
+            # raise error if database is updating for module operations
+            # acquire the shared-lock for the current transaction only
+            self.env.cr.execute("SELECT pg_advisory_xact_lock_shared(hashtext('registry_loading')) NOWAIT")
             # raise error if another transaction is trying to schedule module operations concurrently
-            self.env.cr.execute("LOCK ir_module_module IN EXCLUSIVE MODE NOWAIT")
+            self.env.cr.execute("LOCK ir_module_module IN EXCLUSIVE MODE")
         except psycopg2.OperationalError:
             raise UserError(_("Odoo is currently processing another module operation.\n"
                                "Please try again later or contact your system administrator."))
@@ -633,7 +634,7 @@ class IrModuleModule(models.Model):
             # This is done because the installation/uninstallation/upgrade can modify a currently
             # running cron job and prevent it from finishing, and since the ir_cron table is locked
             # during execution, the lock won't be released until timeout.
-            self.env.cr.execute("SELECT FROM ir_cron FOR UPDATE NOWAIT")
+            self.env.cr.execute("SELECT FROM ir_cron FOR UPDATE")
         except psycopg2.OperationalError:
             raise UserError(_("Odoo is currently processing a scheduled action.\n"
                               "Module operations are not possible at this time, "
--- a/odoo/modules/loading.py
+++ b/odoo/modules/loading.py
@@ -15,7 +15,6 @@ import traceback
 
 import odoo.sql_db
 import odoo.tools.sql
-import odoo.tools.translate
 from odoo import api, tools
 from odoo.tools import OrderedSet
 from odoo.tools.convert import convert_file, IdRef, ConvertMode as LoadMode
@@ -267,6 +266,7 @@ def load_module_graph(
             package.state = 'installed'
             module.env.flush_all()
             module.env.cr.commit()
+            registry.signal_changes()
 
         test_time = 0.0
         test_queries = 0
@@ -335,6 +335,7 @@ def _check_module_names(cr: BaseCursor, module_names: Iterable[str]) -> None:
 
 def load_modules(
     registry: Registry,
+    cr: BaseCursor,
     *,
     update_module: bool = False,
     upgrade_modules: Collection[str] = (),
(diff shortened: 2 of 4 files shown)


odoo/orm/registry.py
```diff
@@ -175,43 +178,50 @@ def new(
...
+                        if cr.rowcount:
+                            update_module = True
+                            # We are updating modules. We must release the shared lock in order to
+                            # acquire the exclusive lock. The upgrade marker (config parameter
+                            # above) is deleted while loading modules.
+                            cr.commit()
+
+                if update_module:
+                    # acquire the exclusive lock
+                    cr.execute("SELECT pg_advisory_xact_lock(hashtext('registry_loading'))")
```
HydrionBurst: This LOCK will be hold during the whole loading.
As a result `registry.new` during `registry.new` (the strategy for uninstall) will have dead look.

Maybe we can make the `cr` as a thread local or even global (because in multi-thread mode, we already have `@locked`) so that we can reuse it since `pg_advisory_xact_lock` is reentrant.

odoo/orm/registry.py
```diff
@@ -175,43 +178,50 @@ def new(
...
+                update_module = True
+
+            with ExitStack() as exit_stack:
+                # The transaction 'cr' defines a critical section for multi-worker concurrency
+                # control. It uses a shared lock to avoid registry updates while loading modules,
+                # and an exclusive lock when updating the registry.
+                cr = exit_stack.enter_context(registry.cursor())
+                if not update_module:
+                    # acquire the shared lock
+                    cr.execute("SELECT pg_advisory_xact_lock_shared(hashtext('registry_loading'))")
```
HydrionBurst: shall we `NOWAIT`?

Assume worker1 is installing slow modules for 5 minutes and has successfully installed one module with `signal_changes`. 
Also there are lots of requests for the same database to other workers.

Since those works detects the signal changes, they will try to `registry.new` and wait for the worker1.

I think it is unfair for other database users on saas 🤔

odoo/orm/registry.py
kmagusiak: I prefer to avoid thread locals or other kind of global variables when possible. I just passed it as a parameter now.

odoo/orm/registry.py
kmagusiak: Indeed, we should put some low timeout for the saas and raise some exception.
For instances with only a few daatabases, it makes sense to wait longer instead of failing.

I will test how it behaves.

odoo/orm/registry.py
```diff
@@ -175,43 +180,55 @@ def new(
...
+                        if cr.rowcount:
+                            update_module = True
+                            # We are updating modules. We must release the shared lock in order to
+                            # acquire the exclusive lock. The upgrade marker (config parameter
+                            # above) is deleted while loading modules.
+                            cr.commit()
+
+                if update_module:
+                    # acquire the exclusive lock
+                    cr.execute("SELECT pg_advisory_xact_lock(hashtext('registry_loading'))")
```
HydrionBurst: This one should always be `NO WAIT`
If there are 10 concurrent `registry.new`. We just need one of them to update database.

odoo/orm/registry.py
```diff
@@ -175,43 +180,55 @@ def new(
...
+                    cr.execute("SELECT pg_advisory_xact_lock_shared(hashtext('registry_loading'))" + wait_sql)
+                    if db.is_initialized(cr):
+                        # check whether module updates are pending
+                        cr.execute("SELECT FROM ir_config_parameter WHERE key='base.partially_updated_database'")
+                        if cr.rowcount:
+                            update_module = True
+                            # We are updating modules. We must release the shared lock in order to
+                            # acquire the exclusive lock. The upgrade marker (config parameter
+                            # above) is deleted while loading modules.
+                            cr.commit()
```
HydrionBurst: I can understand you move the DELETE to the load_modules to avoid the lock for update_module transaction blocks other requests.


But I don't get why you have to commit here

> We must release the shared lock in order to acquire the exclusive lock.

I think there is no need to commit in order to upgrade the pg_advisory_xact_lock_shared to pg_advisory_xact_lock
Or do you mean someting else?

On the other hand, I feel we need to `commit` immediately after `DELETE` to have the logic `try to upgrade remained modules once for partially upgraded database`
(NOT TESTED) Assume the partially upgraded customer's database on saas has `base.partially_updated_database` but the first module to upgrade will cause timeout or memory error. The first request will update modules and the worker will be killed without anything to be commited. Since the cursor for `DELETE` is not commited, and the `reset_modules_state` is not called. The second request will update modules and kill the worker again.

odoo/orm/registry.py
```diff
@@ -175,43 +180,55 @@ def new(
...
+                # The transaction 'cr' defines a critical section for multi-worker concurrency
+                # control. It uses a shared lock to avoid registry updates while loading modules,
+                # and an exclusive lock when updating the registry.
+                if lock_cr is None:
+                    cr = exit_stack.enter_context(registry.cursor())
+                else:
+                    cr = lock_cr
+                if not update_module:
+                    # acquire the shared lock
+                    wait_sql = "" if lock_wait else " NOWAIT"
```
HydrionBurst: NOTE for review.
Since we use `NOWAIT`, it will be more likely to get errors from 
`check_signaling` during update module. We need to check their error handling.

HydrionBurst (commented):
Another idea to be discussed.

What do you think about
`Registry.new_lock_crs :dict[str, BaseCursor]` we can pop it after use
So that
1. we don't have to pass it in args
2. we can group up all lock logic in `class Registry` and remove the code

https://github.com/odoo-dev/odoo/blob/beb66fbfdf834d9d9f4065b3f6039c1c0cb8a7cd/odoo/addons/base/models/ir_module.py#L622-L641

3. it may be easier to aovid the cron job being restarted during the whole loading process.

odoo/orm/registry.py
kmagusiak: Just a POC now, I will put the PR in draft.

odoo/orm/registry.py
kmagusiak: Trying to update a shared lock to an exclusive one will deadlock, if you have 2 servers with a shared lock and both try go get an exclusive lock.

We may store the number of times we had partially updated the database to detect timeouts. I will check your use case.

odoo/orm/registry.py
kmagusiak: We need to wait that shared locks are released.
If we have multiple updates, only the first one will successfully delete the upgrade marker. So they will just wait.

odoo/orm/registry.py
```diff
@@ -48,6 +49,7 @@
 
 _logger = logging.getLogger('odoo.registry')
 _schema = logging.getLogger('odoo.schema')
+_cr_registry_loading = contextvars.ContextVar("cr_registry_loading")
```
HydrionBurst: I think you use `contextvars` for gevent server right?
If so, could you leave a comment?

Aslo, what about using `Regsitry._loading_cr`

odoo/orm/registry.py
```diff
@@ -175,43 +181,70 @@ def new(
...
+                            _logger.debug("Database %s initialized, removing upgrade marker", cr.dbname)
+                            update_module = True
+                            # We are updating modules. We must release the shared lock in order to
+                            # acquire the exclusive lock. The upgrade marker (config parameter
+                            # above) is deleted while loading modules.
+
+                if update_module:
+                    # acquire the exclusive lock
+                    cr.execute("SELECT pg_advisory_unlock_shared(hashtext('registry_loading'))")
+                    cr.execute("SELECT pg_advisory_lock(hashtext('registry_loading'))")
```
HydrionBurst, line 211: Where and how do you promise the lock will be released?

kmagusiak: When you close the cursor (exiting the ExitStack), the connection is returned to the pool and reset. The reset of the connection releases session locks and session parameters. Otherwise lock timeouts would survive.

HydrionBurst: I get your idea, but when the `cr.close()` the `cnx` is not closed (because most of time we want `give_back(cnx, keep_in_pool=True)`, and the session will be kept for a long time I guess.

kmagusiak: I wrote a PR to remove the "leak" parameter which is unused. And in that PR, I am resetting a connection when giving it back to the pool (not when taking it). Here: #247204

kmagusiak:
@robodoo rebase-ff

odoo/addons/base/models/ir_module.py
```diff
@@ -619,13 +619,13 @@ def _button_immediate_function(self, function):
                 "for help."
             )
 
-        # raise error if database is updating for module operations
-        if self.search_count([('state', 'in', ('to install', 'to upgrade', 'to remove'))], limit=1):
-            raise UserError(_("Odoo is currently processing another module operation.\n"
-                               "Please try again later or contact your system administrator."))
+        self.env.cr.execute("SET LOCAL lock_timeout = '3s'")
```
nseinlet, line 621: Why 3? Why not 19?

odoo/orm/registry.py
```diff
@@ -175,40 +180,65 @@ def new(
...
+            if upgrade_modules or install_modules or reinit_modules:
+                update_module = True
+
+            with ExitStack() as exit_stack:
+                # The transaction 'cr' defines a critical section for multi-worker concurrency
+                # control. It uses a shared lock to avoid registry updates while loading modules,
+                # and an exclusive lock when updating the registry.
+                cr = exit_stack.enter_context(registry.cursor())
+                if lock_wait > 0:
+                    cr.execute(f"SET SESSION lock_timeout = '{int(lock_wait)}s'")
```
nseinlet: Why not 3?

odoo/orm/registry.py
```diff
@@ -175,40 +180,65 @@ def new(
...
+                if not update_module:
+                    # acquire the shared lock
+                    cr.execute("SELECT pg_advisory_lock_shared(hashtext('registry_loading'))")
+                    if db.is_initialized(cr):
+                        # check whether module updates are pending
+                        cr.execute("DELETE FROM ir_config_parameter WHERE key='base.partially_updated_database'")
```
nseinlet: Is this still necessary with the exclusive lock?

odoo/orm/registry.py
```diff
@@ -175,40 +180,65 @@ def new(
...
                     # load_modules multiple times in case there are modules to be uninstalled
-                    load_modules(
-                        registry,
-                        update_module=update_module,
-                        upgrade_modules=upgrade_modules,
-                        install_modules=install_modules,
-                        reinit_modules=reinit_modules,
-                        new_db_demo=new_db_demo,
-                    )
+                    cr.commit()  # start a new transaction
```
nseinlet: This will release the lock from line 208? Can this lead to 2 different workers reach this loop concurrently?

odoo/addons/base/models/ir_module.py
kmagusiak: This is in the GUI, I just took a small number so that if you click to install or update a module, you have the chance for a quickly running cron to terminate without letting the user hang there.

odoo/orm/registry.py
kmagusiak: The default 15 comes from the session lock timeout set in `load_modules` (loading).

odoo/orm/registry.py
kmagusiak: No, `pg_advisory_lock` is a lock at session-level.

Xavier-Do: Nevermind

odoo/orm/registry.py
kmagusiak: Indeed, I added a commit to remove it.
This is no longer needed because the exclusive lock is on the session-level.

odoo/addons/base/models/ir_module.py
```diff
@@ -619,13 +619,13 @@ def _button_immediate_function(self, function):
...
+
         try:
+            # raise error if database is updating for module operations
+            self.env.cr.execute("SELECT pg_advisory_xact_lock_shared(hashtext('registry_loading')) NOWAIT")
```
rco-odoo, line 626: Add explicitly in the comment that this is a transaction-level lock, which will be released  when the transaction ends.

odoo/orm/registry.py
```diff
@@ -175,40 +179,46 @@ def new(
...
-                    update_module = True
+            if upgrade_modules or install_modules or reinit_modules:
+                update_module = True
+
+            with ExitStack() as exit_stack:
+                # The transaction 'cr' defines a critical section for multi-worker concurrency
+                # control. It uses a shared lock to avoid registry updates while loading modules,
+                # and an exclusive lock when updating the registry.
+                cr = exit_stack.enter_context(registry.cursor())
+                if lock_wait > 0:
```
rco-odoo: Using a negative value feels like bad practice to me...

odoo/orm/registry.py
```diff
@@ -175,40 +179,46 @@ def new(
...
+                    cr.execute(f"SET SESSION lock_timeout = '{int(lock_wait)}s'")
+                if update_module:
+                    # acquire the exclusive lock
```
rco-odoo:
```suggestion
                    # acquire the exclusive lock at the session level; this
                    # enables to guard several transactions under the lock
```

odoo/orm/registry.py
```diff
@@ -175,40 +179,46 @@ def new(
...
+                    cr.execute("SELECT pg_advisory_lock(hashtext('registry_loading'))")
+                else:
+                    # acquire the shared lock
```
rco-odoo:
```suggestion
                    # acquire the shared lock (at the session level)
```

odoo/service/server.py
```diff
@@ -1595,7 +1595,14 @@ def preload_registries(dbnames):
...
                 update_module = config['init'] or config['update'] or config['reinit']
 
-                registry = Registry.new(dbname, update_module=update_module, install_modules=config['init'], upgrade_modules=config['update'], reinit_modules=config['reinit'])
+                registry = Registry.new(
+                    dbname,
+                    update_module=update_module,
+                    install_modules=config['init'],
+                    upgrade_modules=config['update'],
+                    reinit_modules=config['reinit'],
+                    lock_wait=-1,
```
rco-odoo: Why disabling the way here?

kmagusiak: I thought that we may want a different timeout value when starting a server, but we can add it only when we need it.

odoo/orm/registry.py
```diff
@@ -194,20 +193,10 @@ def new(
...
-                    if db.is_initialized(cr):
-                        # check whether module updates are pending
-                        cr.execute("DELETE FROM ir_config_parameter WHERE key='base.partially_updated_database'")
-                        if cr.rowcount:
-                            _logger.debug("Database %s initialized, removing upgrade marker", cr.dbname)
-                            # We are updating modules. We must release the shared lock in order to
-                            # acquire the exclusive lock. The upgrade marker (config parameter
-                            # above) is deleted while loading modules.
-                            cr.execute("SELECT pg_advisory_unlock_shared(hashtext('registry_loading'))")
-                            update_module = True
```
HydrionBurst: why can we remove the base.partially_updated_database?
`This flag was needed` because during upgrade we may not have codes for all modules which leaves many `to upgrade` ir.module.module.
We need some code/data to identify the situation and set `update_module=True` explicitly the next time `Registry.new(..., update_module=False)`

I expect maybe we can directly using sql to check if there is any `to xxx` module in database to update `update_module` without using the `ir_config_parameter` but I don't get why we can drop the whole logic.

odoo/orm/registry.py
```diff
@@ -191,23 +190,22 @@ def new(
                 assert lock_wait >= 0
                 cr.execute(f"SET SESSION lock_timeout = '{int(lock_wait)}s'")
 
+                # check if we have a partially upgraded database
+                # this can happen if a previous upgrade was done with some
+                # missing modules
+                if not update_module:
```
HydrionBurst:
```suggestion
                if not update_module and db.is_initialized(cr):
```
I hope this [error message](https://github.com/odoo/odoo/blob/320461ec15249e2c3506fbfda78a243f8f8e9435/odoo/modules/loading.py#L358) can be correctly logged

odoo/modules/loading.py
```diff
@@ -259,6 +258,7 @@ def load_module_graph(
             package.state = 'installed'
             module.env.flush_all()
             module.env.cr.commit()
+            registry.signal_changes()
```
HydrionBurst, line 269: Does this mean during a long installaiton (1 min), if there are many requests to the same database concurrently, all workers for those requests will be blocked for `lock_wait=15` seconds?

odoo/addons/base/models/ir_module.py
```diff
@@ -619,13 +619,14 @@ def _button_immediate_function(self, function):
...
-                               "Please try again later or contact your system administrator."))
+        self.env.cr.execute("SET LOCAL lock_timeout = '3s'")
+
         try:
+            # raise error if database is updating for module operations
+            # acquire the shared-lock for the current transaction only
+            self.env.cr.execute("SELECT pg_advisory_xact_lock_shared(hashtext('registry_loading')) NOWAIT")
             # raise error if another transaction is trying to schedule module operations concurrently
-            self.env.cr.execute("LOCK ir_module_module IN EXCLUSIVE MODE NOWAIT")
+            self.env.cr.execute("LOCK ir_module_module IN EXCLUSIVE MODE")
```
HydrionBurst, line 628: why do we remove the `NO WAIT`?

odoo/modules/loading.py
kmagusiak: Yes. And we don't want them connecting because otherwise they will query a database whose schema is being modified.
A possibility would be to have a maintenance page for Odoo. We don't have one yet.

odoo/addons/base/models/ir_module.py
kmagusiak: I added a timeout of 3s in this function.
In my experience some cron jobs were blocking the update buttons, therefore, I leave a few seconds for locking when the user clicks on upgrade.

HydrionBurst: Yes, I understand the ir_cron part, but this is for `ir_module_module`.

odoo/modules/loading.py
HydrionBurst: I am OK, as long as the side-effect is considered and accepted for now.

odoo/addons/base/models/ir_module.py
kmagusiak: Same logic. You can wait a little bit, but you don't want to wait indefinitely (which would be the case without a defined timeout).

odoo/orm/registry.py
```diff
@@ -191,23 +191,22 @@ def new(
...
                 cr.execute(f"SET SESSION lock_timeout = '{int(lock_wait)}s'")
 
+                # check if we have a partially upgraded database
+                # this can happen if a previous upgrade was done with some
+                # missing modules
+                if not update_module and db.modules_db.is_initialized(cr):
+                    cr.execute("SELECT FROM ir_module_module WHERE state IN ('to upgrade', 'to install', 'to remove') LIMIT 1")
+                    if cr.fetchone():
+                        _logger.info("Force update_module=True, some modules were partially updated")
+                        update_module = True
```
rco-odoo: We should acquire the shared lock _before_ this query, otherwise the following may happen:
```
        worker A                    worker B

   wait exclusive lock
            ⋮
 acquire exclusive lock
            |
     install module
            |
            |               detect modules to install
            |                  update_module = True
            |                           |
            |                  wait exclusive lock
            |                           ⋮
     install module                     ⋮
            |                           ⋮
            |                           ⋮
 release exclusive lock                 ⋮
                             acquire exclusive lock
                                but nothing to do!
```

odoo/addons/base/models/ir_module.py
HydrionBurst: I think the old `LOCK ir_module_module IN EXCLUSIVE MODE NO WAIT` was mainly for 2 transactions `button_immediate_install` at the same time. Even if it may not be perfect.

The current code, not sure when the error will be raise, but I think it is not because of `# raise error if another transaction is trying to schedule module operations concurrently`

Maybe we could even remove this line.

odoo/orm/registry.py
kmagusiak: It may happen only during the first boot after an upgrade done on another system. It will acquire an exclusive lock in that case, but will not fail.

odoo/addons/base/models/ir_module.py
kmagusiak: Given that we acquire the registry lock, this one seemed superfluous. I left it here because it was exclusive.

HydrionBurst: I think the current `self.env.cr.execute("LOCK ir_module_module IN EXCLUSIVE MODE")` is useless, we can remove it.

In the worst corner case, 2 transactions `button_immediate_install` different modules without conflict. We can still promise there is one worker update all modules. In the future, if we really want to prevent the corner case, we can write to   the same row of `ir_module_module` e.g. the `UDPATE state=state FROM ir_module_module WHERE name = 'base'`

And then you can move `self.env.cr.execute("SET LOCAL lock_timeout = '3s'")` close to `self.env.cr.execute("SELECT FROM ir_cron FOR UPDATE")`

odoo/orm/registry.py
HydrionBurst: No, it will happen everytime after you `button_immediate_install`

The main worker updating modules

Many other workers waiting for the exclusive lock with `update_module=True`
So, technically, after installation, these workers have to `load_modules(... update_module=True)` one by one because they need exclusive lock

Not sure if it is already a problem, I think we should avoid `update_module=True` in gevent server

odoo/orm/registry.py
```diff
@@ -191,23 +191,31 @@ def new(
...
+                    cr.execute("SELECT pg_advisory_lock(hashtext('registry_loading'))")
+                    cr.rollback()
+                    cr.execute("SELECT FROM ir_module_module WHERE state IN ('to upgrade', 'to install', 'to remove') LIMIT 1")
+                    update_partial = bool(cr.fetchone())
+                    if update_partial:
+                        _logger.info("Force update_module=True, some modules were partially updated")
+                        update_module = True
+                    else:
+                        cr.execute("SELECT pg_advisory_unlock(hashtext('registry_loading'))")
+                        cr.execute("SELECT pg_advisory_lock_shared(hashtext('registry_loading'))")
```
rco-odoo: It looks more complex than necessary to me.  You can actually ask for an exclusive lock while you already have the corresponding shared lock.  I therefore suggest the following structure:
```py
if not update_module:
    cr.execute("SELECT pg_advisory_lock_shared(hashtext('registry_loading'))")
    update_module = has_pending_module_updates()  # this checks module states
if update_module:
    cr.execute("SELECT pg_advisory_lock(hashtext('registry_loading'))")
```
**Case 1.**  A worker is currently doing module updates.  In that case, all other workers will have to wait for the shared lock before checking for pending module updates.

**Case 2.**  Two workers (or more) simultaneously enter the section between the shared lock and the exclusive lock.  They both go for module updates, and both try to acquire the exclusive lock.  The first one waits.  The second one will get an error, because PG detected a deadlock.  It therefore retries, implicitly releases the shared lock, and therefore enables the first worker to acquire the exclusive lock.  We are back to Case 1.
```
        worker A                    worker B

   acquire shared lock
            |                  acquire shared lock
  check pending updates                 |
            |                 check pending updates
   wait exclusive lock                  |
            ⋮                  wait exclusive lock  → error (deadlock detected)
            ⋮                                       → restart (release lock)
  acquire exclusive lock
```
I haven't checked the "retry" part, but for sure the second worker will release the shared lock.

kmagusiak: I didn't want to rely on deadlock detection. Indeed it is simpler.
Let's check the final solution during the week.

PR #245920, opened 2026-01-27
[FIX] project: Group project template tasks with template tasks

VincentJanssen-Code:
## Current behavior before PR:
- When a user viewed All Tasks or My Tasks, tasks related to project templates were visible.
- When a user specifically selected the Template Tasks filter, tasks associated with template projects were not visible.

## Desired behavior after PR is merged:
- When a user views My Tasks or All Tasks, tasks related to template projects are hidden.
- When a user selects the Template Tasks filter, both template tasks and tasks associated with template projects are shown.

## Additional note

A modification to `project_report.py` was required because it inherits from the project task view and overrides the view’s reference model.

Task-[5485658](https://www.odoo.com/odoo/project/4105/tasks/5485658)

---

### DIFF
--- a/addons/project/report/project_report.py
+++ b/addons/project/report/project_report.py
@@ -147,7 +147,6 @@ class ReportProjectTaskUser(models.Model):
     def _where(self):
         return """
                 t.project_id IS NOT NULL
-                AND p.is_template IS NOT TRUE
         """
 
     def init(self):
--- a/addons/project/report/project_report_views.xml
+++ b/addons/project/report/project_report_views.xml
@@ -75,7 +75,7 @@
             <field name="name">Tasks Analysis</field>
             <field name="res_model">report.project.task.user</field>
             <field name="path">tasks-analysis</field>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False), ('project_id.is_template', '=', False)]</field>
             <field name="view_mode">graph,pivot</field>
             <field name="search_view_id" ref="view_task_project_user_search"/>
             <field name="context">{'group_by': [], 'graph_measure': '__count__'}</field>
--- a/addons/project/static/src/views/project_task_model_mixin.js
+++ b/addons/project/static/src/views/project_task_model_mixin.js
@@ -12,10 +12,16 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends 
             ]).toList({});
         }
         if (this.env.searchModel.context?.render_task_templates) {
-            domain = Domain.and([
-                Domain.removeDomainLeaves(domain, ['has_template_ancestor']).toList(),
-                [['has_template_ancestor', '=', true]],
-            ]).toList({});
+            domain = Domain.removeDomainLeaves(domain, [
+                "has_template_ancestor",
+                "has_project_template",
+                "project_id.is_template",
+            ]);
+            const templateTaskDomain = Domain.or([[["has_template_ancestor", "=", true]],
+                "default_project_id" in this.env.searchModel.globalContext ?
+                        Domain.TRUE :
+                        [["project_id.is_template", "=", true]]]);
+            domain = Domain.and([domain, templateTaskDomain]).toList({});
         }
         return domain;
     }
--- a/addons/project/views/project_task_views.xml
+++ b/addons/project/views/project_task_views.xml
@@ -28,7 +28,7 @@
                         <filter name="create_date_last_365_days" string="Last 365 Days" domain="[('date_last_stage_update', '&gt;=', 'today -365d +1d')]"/>
                     </filter>
                     <separator/>
-                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="[('has_template_ancestor', '=', True)]"/>
+                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|', ('has_template_ancestor', '=', True), ('project_id.is_template', '=', True)]"/>
                     <group>
                         <filter string="Stage" name="stage" context="{'group_by': 'stage_id'}"/>
                         <filter string="Milestone" name="milestone" context="{'group_by': 'milestone_id'}" groups="project.group_project_milestone" invisible="not context.get('allow_milestones', True)"/>
@@ -1047,7 +1047,7 @@
                 'default_user_ids': [(4, uid)],
             }</field>
             <field name="search_view_id" ref="view_task_search_form"/>
-            <field name="domain">[('user_ids', 'in', uid), ('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('user_ids', 'in', uid), ('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
             <field name="help" type="html">
                 <p class="o_view_nocontent_smiling_face">
                     No tasks found. Let's create one!
@@ -1097,7 +1097,7 @@
             <field name="res_model">project.task</field>
             <field name="path">all-tasks</field>
             <field name="view_mode">list,kanban,form,calendar,activity,pivot,graph</field>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
             <field name="context">{'search_default_open_tasks': 1, 'default_user_ids': [(4, uid)]}</field>
             <field name="search_view_id" ref="view_task_search_form"/>
             <field name="help" type="html">
@@ -1198,7 +1198,7 @@
             <field name="res_model">project.task</field>
             <field name="view_mode">list,kanban,form,calendar,pivot,graph,activity</field>
             <field name="search_view_id" ref="view_task_search_form"/>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
             <field name="help" type="html">
                 <p class="o_view_nocontent_smiling_face">
                     No tasks found. Let's create one!
@@ -1238,7 +1238,7 @@
             <field name="name">Overpassed Tasks</field>
             <field name="res_model">project.task</field>
             <field name="view_mode">list,form,calendar,graph,kanban</field>
-            <field name="domain">[('is_closed', '=', False), ('date_deadline', '&lt;', 'today'), ('project_id', '!=', False), ('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('is_closed', '=', False), ('date_deadline', '&lt;', 'today'), ('project_id', '!=', False), ('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
             <field name="filter" eval="True"/>
             <field name="search_view_id" ref="view_task_search_form"/>
         </record>


addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -12,9 +12,9 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends
             ]).toList({});
         }
         if (this.env.searchModel.context?.render_task_templates) {
+            domain = Domain.removeDomainLeaves(domain,['has_template_ancestor','has_project_template']).toList()
```
xavierbol: We could let `,` breathe a bit. 😇 

```suggestion
            domain = Domain.removeDomainLeaves(domain,['has_template_ancestor', 'has_project_template']).toList()
```

addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -12,9 +12,9 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends
...
             domain = Domain.and([
-                Domain.removeDomainLeaves(domain, ['has_template_ancestor']).toList(),
-                [['has_template_ancestor', '=', true]],
+                ['|',['has_template_ancestor', '=', true],['has_project_template','=', true]],
```
xavierbol: same comment here. 😇 

```suggestion
                ['|', ['has_template_ancestor', '=', true],['has_project_template', '=', true]],
```

addons/project/views/project_task_views.xml
```diff
@@ -28,7 +28,7 @@
                         <filter name="create_date_last_365_days" string="Last 365 Days" domain="[('date_last_stage_update', '&gt;=', 'today -365d +1d')]"/>
                     </filter>
                     <separator/>
-                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="[('has_template_ancestor', '=', True)]"/>
+                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|',('has_template_ancestor', '=', True),('has_project_template','=',True)]"/>
```
xavierbol: same comment here
```suggestion
                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|',('has_template_ancestor', '=', True), ('has_project_template', '=', True)]"/>
```

addons/project/views/project_task_views.xml
```diff
@@ -209,7 +209,7 @@
             <field name="path">tasks</field>
             <field name="res_model">project.task</field>
             <field name="view_mode">kanban,list,form,calendar,pivot,graph,activity</field>
-            <field name="domain">[('project_id', '=', active_id), ('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('project_id', '=', active_id), ('has_template_ancestor', '=', False),('has_project_template','=',False)]</field>
```
xavierbol:
```suggestion
            <field name="domain">[('project_id', '=', active_id), ('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
```

addons/project/views/project_task_views.xml
```diff
@@ -1046,7 +1046,7 @@
                 'default_user_ids': [(4, uid)],
             }</field>
             <field name="search_view_id" ref="view_task_search_form"/>
-            <field name="domain">[('user_ids', 'in', uid), ('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('user_ids', 'in', uid), ('has_template_ancestor', '=', False),('has_project_template','=',False)]</field>
```
xavierbol:
```suggestion
            <field name="domain">[('user_ids', 'in', uid), ('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
```

addons/project/views/project_task_views.xml
```diff
@@ -1096,7 +1096,7 @@
             <field name="res_model">project.task</field>
             <field name="path">all-tasks</field>
             <field name="view_mode">list,kanban,form,calendar,activity,pivot,graph</field>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False),('has_project_template','=',False)]</field>
```
xavierbol:
```suggestion
            <field name="domain">[('has_template_ancestor', '=', False),('has_project_template', '=', False)]</field>
```

addons/project/views/project_task_views.xml
```diff
@@ -1197,7 +1197,7 @@
             <field name="res_model">project.task</field>
             <field name="view_mode">list,kanban,form,calendar,pivot,graph,activity</field>
             <field name="search_view_id" ref="view_task_search_form"/>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False),('has_project_template','=',False)]</field>
```
xavierbol:
```suggestion
            <field name="domain">[('has_template_ancestor', '=', False),('has_project_template', '=', False)]</field>
```

addons/project/views/project_task_views.xml
```diff
@@ -1237,7 +1237,7 @@
             <field name="name">Overpassed Tasks</field>
             <field name="res_model">project.task</field>
             <field name="view_mode">list,form,calendar,graph,kanban</field>
-            <field name="domain">[('is_closed', '=', False), ('date_deadline', '&lt;', 'today'), ('project_id', '!=', False), ('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('is_closed', '=', False), ('date_deadline', '&lt;', 'today'), ('project_id', '!=', False), ('has_template_ancestor', '=', False),('has_project_template','=',False)]</field>
```
xavierbol:
```suggestion
            <field name="domain">[('is_closed', '=', False), ('date_deadline', '&lt;', 'today'), ('project_id', '!=', False), ('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
```

xavierbol (commented):

addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -3,6 +3,7 @@ import { Domain } from "@web/core/domain";
 
 export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends T {
     _processSearchDomain(domain) {
+        console.log(this.env.searchModel.globalContext)
```
xavierbol: Should be removed. 😇

addons/project/report/project_report.py
```diff
@@ -159,3 +161,13 @@ def init(self):
...
+
+    # has_template_project is always assumed to have a value of false inside
+    # of the report as they are not stored in the report view
+    def _search_has_template_project(self, operator, value):
+        if (operator not in ('=', 'in')):
+            return NotImplemented
+        elif (not all(value)):
+            return Domain.TRUE
+        else:
+            return Domain.FALSE
```
xavierbol: I don't really understand that magic 🪄 

Why not storing that field and alter the query used for that report, it would be easier since `has_template_project` field is stored in `project.task`?

I am not sure we can do it in stable, I have to check but alternatively, we could use `task_id` in the search method instead of doing assumptions in the search method.

xavierbol (commented):

addons/project/views/project_task_views.xml
```diff
@@ -28,7 +28,7 @@
                         <filter name="create_date_last_365_days" string="Last 365 Days" domain="[('date_last_stage_update', '&gt;=', 'today -365d +1d')]"/>
                     </filter>
                     <separator/>
-                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="[('has_template_ancestor', '=', True)]"/>
+                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|', ('has_template_ancestor', '=', True), ('project_id.is_template', '=', True)]"/>
```
xavierbol:
```suggestion
                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|', ('has_template_ancestor', '=', True), ('has_project_template', '=', True)]"/>
```

addons/project/report/project_report_views.xml
```diff
@@ -75,7 +75,7 @@
             <field name="name">Tasks Analysis</field>
             <field name="res_model">report.project.task.user</field>
             <field name="path">tasks-analysis</field>
-            <field name="domain">[('has_template_ancestor', '=', False)]</field>
+            <field name="domain">[('has_template_ancestor', '=', False), ('project_id.is_template', '=', False)]</field>
```
xavierbol, line 78:
```suggestion
            <field name="domain">[('has_template_ancestor', '=', False), ('has_project_template', '=', False)]</field>
```

addons/project/report/project_report.py
xavierbol, line 1: Since it is a report and the SQL view built for that report is drop and create again when `init` is called, we can add `has_project_template` field inside that report. 🙂 

Let's do something like this: 

```diff
diff --git a/addons/project/report/project_report.py b/addons/project/report/project_report.py
index 4942d3ee27ce..9d4b1300a071 100644
--- a/addons/project/report/project_report.py
+++ b/addons/project/report/project_report.py
@@ -68,6 +68,7 @@ class ReportProjectTaskUser(models.Model):
     # We exclude template tasks, but we still need the field for the views
     is_template = fields.Boolean(readonly=True)
     has_template_ancestor = fields.Boolean(readonly=True)
+    has_project_template = fields.Boolean(readonly=True)
 
     def _select(self):
         return """
@@ -101,7 +102,8 @@ class ReportProjectTaskUser(models.Model):
                 (extract('epoch' from (t.date_deadline-(now() at time zone 'UTC'))))/(3600*24) as delay_endings_days,
                 COUNT(td.task_id) as dependent_ids_count,
                 t.is_template,
-                t.has_template_ancestor
+                t.has_template_ancestor,
+                p.is_template AS has_project_template
         """
 
     def _group_by(self):
```

addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -12,10 +12,12 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends
...
-            domain = Domain.and([
-                Domain.removeDomainLeaves(domain, ['has_template_ancestor']).toList(),
-                [['has_template_ancestor', '=', true]],
-            ]).toList({});
+            domain = Domain.removeDomainLeaves(domain, ['has_template_ancestor', 'has_project_template', 'project_id.is_template']).toList()
+            const templateTaskDomain = ['has_template_ancestor', '=', true]
+            const finaldomain = 'default_project_id' in this.env.searchModel.globalContext ? 
+                        [templateTaskDomain]:
+                        ['|', templateTaskDomain, ['project_id.is_template', '=', true]]
+            domain = Domain.and([domain, finaldomain]).toList({});
```
xavierbol: Some remarks 🙂 

1. no need to do `.toList()` at the first line of this diff because `Domain.and` done at the end of this diff support array and Domain object. 🙂 
2. missing `;` at line 16 (second line in that diff)
3. finalDomain can be replaced by using `Domain.or`. 😇 

```suggestion
            domain = Domain.removeDomainLeaves(domain, [
                "has_template_ancestor",
                "has_project_template",
                "project_id.is_template",
            ]);
            const templateTaskDomain = Domain.or([
                [["has_template_ancestor", "=", true]],
                "default_project_id" in this.env.searchModel.globalContext
                    ? Domain.TRUE
                    : [["has_project_template", "=", true]],
            ]);
            domain = Domain.and([domain, templateTaskDomain]).toList({});
```

xavierbol (commented):

I would prefer we add a field `has_project_template` field in `report.project.task.user` report because 
- this model is not a real table but a SQL view and since that SQL view is refreshed when we restart the server, there is no issue to add a field inside that report. 🙂 
- `project_id.is_template` could also be falsy if the current user has no access to that project (if the project is private and the user is a project user and he is not in the follower of that project for instance). In that case, the user could have access to some tasks inside that private but he would see it inside that report due to `('project_id.is_template', '=', False)` inside the action domain. 🙂

addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -12,10 +12,18 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends
             ]).toList({});
         }
         if (this.env.searchModel.context?.render_task_templates) {
-            domain = Domain.and([
-                Domain.removeDomainLeaves(domain, ['has_template_ancestor']).toList(),
-                [['has_template_ancestor', '=', true]],
-            ]).toList({});
+            console.log(domain);
```
xavierbol: has to be removed. 😇

addons/project/static/src/views/project_task_model_mixin.js
```diff
@@ -12,10 +12,18 @@ export const ProjectTaskModelMixin = (T) => class ProjectTaskModelMixin extends
...
+            domain = Domain.removeDomainLeaves(domain, [
+                "has_template_ancestor",
+                "has_project_template",
+                "project_id.is_template",
+            ]);
+            console.log(domain.toList());
```
xavierbol: same here

xavierbol (commented):
LGTM, thanks for your work! 🙂

addons/project/report/project_report.py
```diff
@@ -68,6 +68,7 @@ class ReportProjectTaskUser(models.Model):
     # We exclude template tasks, but we still need the field for the views
     is_template = fields.Boolean(readonly=True)
     has_template_ancestor = fields.Boolean(readonly=True)
+    has_project_template = fields.Boolean(readonly=True)
```
cawo-odoo: No, you cannot add a stored field in stable just like that. An upgrade exception cannot be added here. Either you find a way to avoid the new field, or you will need to add a new module just for this.

https://github.com/odoo/odoo/wiki/Contributing#what-does-stable-mean

xavierbol: TBH, I thought it is possible to do that in report since the fields defined in a report (model with sql view instead of being a real table) are not really stored in database, they are just populated with the sql view of that report.

xavierbol: Indeed, my bad, I wrongly test it

xavierbol (commented):
robodoo r+

PR #245922, opened 2026-01-27
remove o_translate_inline

blse-odoo:
design-themes PR: https://github.com/odoo/design-themes/pull/1223
migration script: https://github.com/odoo/upgrade/pull/10423

The class `o_translate_inline` was used to make `a` elements translatable inline (like `b` elements for example).

This PR adds `a` to the list of elements that can be translated inline and remove the `o_translatable_inline` class.


addons/website/static/src/builder/translate.edit.scss
Syozik, line 1: We also should update the rules for some `<a>` elements that do not have a role, right?

Syozik (commented):
Thanks for the PR!
There are some issues now that we are having `<a>` wrapped by a translation span. If we have a few buttons next to each other, the translation span will group them together and not wrap each `<a>` individually:
Image

Also, "Contact Us" and the phone number in the header miss the translation style.
And "Social Media" links aren't translatable currently, so they shouldn't be marked as translatable, right?

addons/website/static/src/builder/translate.edit.scss
blse-odoo: I agree, the `[role]` selector is an heuristic that aims to select the `a` elements for which there is currently another background color set.
- for the phone number, a rule is missing, it should also get a color (I'll add it)
- for the "normal" buttons (aka `a.btn`), the issue is already present: if you add a button with `/button` or with a snippet that contains a button, it should already be translated inline (thanks to the `o_translate_inline` class, it may be missing from some snippets but that is likely a bug). If we add a rule to change their background color, that would replace the purple that make up the button.
- for social media, I'm not sure what should be done. They are actually translatable, you can change them if you go (with debug mode) through "Settings" > "Technical" > "Views", find the view that contains the social snippet (for example search the key `footer_custom`), and click on "EN" just above the xml. But you cannot edit the translation from the builder. This is not the only case, there is also the "Généré par Odoo" in the bottom right of the footer, or the title of the social media snippet (if you activated it in the options from the main language)

Syozik: Ahh, okay, thanks for answering!
For the last point, I think, we could utilize the fact that they are not editable? Add something like `:not(.o_not_editable):not(.o_not_editable *)` ?

blse-odoo: I tried that, and it looks like it would work. But then I realized maybe we should actually be able to edit the title and the links of social media snippet in translate mode. In that case, that would probably be done by adding the social media plugin (or a modified version of it) in translate. In that case the `o_not_editable` would still be present, but they would be editable. I'll check with lebl on Wednesday if we want them to be editable in translate or not :)

Syozik: Thanks!

addons/website/views/website_templates.xml
```diff
@@ -243,7 +243,7 @@
     </xpath>
     <xpath expr="//t[@t-call='web.brand_promotion']/.." position="attributes">
         <attribute name="class" add="o_not_editable small" separator=" "/>
-        <attribute name="t-if">not editable</attribute>
+        <attribute name="t-if">not editable and not translatable</attribute>
```
loco-odoo, line 246: Random line for a comment: maybe you can use mardown link in all of your commit messages so that you do not have to do a line break that feels weird when reading them 👍

addons/html_builder/static/src/core/field_change_replication_plugin.js
```diff
@@ -30,8 +30,6 @@ export class FieldChangeReplicationPlugin extends Plugin {
                 closestElement(r.target, "[data-oe-model], [data-oe-translation-source-sha]")
             )
             .filter(Boolean)
```
loco-odoo, line 49: Random line: Do not forget `html_builder` in your commit message

odoo/addons/test_tools/tests/test_translate.py
```diff
@@ -285,7 +285,9 @@ def test_translate_xml_a(self):
         result = xml_translate(terms.append, source)
         self.assertEqual(result, source)
```
loco-odoo, line 231: Maybe you can add more explanation on the commit message? :pray:  In the commit you linked, it is explained that "The menu items are handled by bootstrap, which expects menu items to be `<a>` elements inside `<li>` inside `<ul>` elements". Why can we go over it now? Has BS becomes more robust or is it us that changed something?

addons/website/static/src/builder/plugins/translation/options/table_of_content_option_plugin_translate.js
```diff
@@ -0,0 +1,27 @@
+import { registry } from "@web/core/registry";
+import { TableOfContentOptionPlugin } from "../../options/table_of_content_option_plugin";
+
+export class TranslateTableOfContentOptionPlugin extends TableOfContentOptionPlugin {
+    getNavbarContainer(tableOfContentEl) {
+        const container = super.getNavbarContainer(tableOfContentEl);
```
loco-odoo: `containerEl` (same for all the other variables in this file ;) )

addons/website/static/src/builder/plugins/translation/options/table_of_content_option_plugin_translate.js
```diff
@@ -0,0 +1,27 @@
...
+        return container.querySelector(":scope > [data-oe-translation-state]") || container;
```
loco-odoo: Why the `||`? I am not sure to see the case in translate mode where the first part of the `||` is not enough

addons/website/static/src/builder/plugins/translation/options/table_of_content_option_plugin_translate.js
```diff
@@ -0,0 +1,27 @@
...
+    updateTableOfContentNavbar(tableOfContentMain) {
+        const tableOfContent = tableOfContentMain.closest(".s_table_of_content");
+        const tableOfContentNavbarContent = this.getNavbarContainer(tableOfContent);
+        const tableOfContentNavbarContentBefore = tableOfContentNavbarContent.innerHTML;
+        super.updateTableOfContentNavbar(tableOfContentMain);
+        const tableOfContentNavbarContentAfter = tableOfContentNavbarContent.innerHTML;
+        if (tableOfContentNavbarContent.matches("[data-oe-translation-state]")) {
+            tableOfContentNavbarContent.dataset.oeTranslationState = "none";
+            if (tableOfContentNavbarContentBefore !== tableOfContentNavbarContentAfter) {
+                tableOfContentNavbarContent.classList.add("o_dirty");
```
loco-odoo: Ideally, this plugin should be agnostic of the `o_dirty` concept

addons/website/static/src/snippets/s_share/000.scss
```diff
@@ -5,7 +5,7 @@
     // Inherit to icons so the option panel reads 40px initially
     --fa-icon-size: 40px;
     $-size: map-get($spacers, 5);
-    > * {
+    > *, > span[data-oe-translation-state] > * {
```
loco-odoo, line 8: We probably do not want to add a translation rule in an asset that is always loaded ;)

loco-odoo (commented):
Thanks for the PR @blse-odoo ! 🙏 
Just few comments ;)

addons/website/static/src/builder/plugins/translation/options/table_of_content_option_plugin_translate.js
blse-odoo: There is no translation span if the container is empty, which happens if there is no heading elements

addons/website/static/src/snippets/s_share/000.scss
blse-odoo: It is only a few rules, and the complexity to avoid it is not worth it

odoo/addons/test_tools/tests/test_translate.py
blse-odoo: I added a `patch` of bootstrap's dropdown in edit, and a change in the mega menu interaction (extracted a bit of code in the normal interaction, and overridden it in the edit interaction)

PR #245929, opened 2026-01-27
[FIX] mail: do not crop canned responses

tsm-odoo:
Canned responses have an arbitrary small width limit in
the navigable list panel. Due to this restriction, the
description of the canned response is barely readable.

This commit makes it take the whole available width.

task-5873810

|| Before | After|
| ------------- | ------------- | -----|
|Discuss|image |image|
|Chat window|image|image|


alexkuhn (approved):
@robodoo delegate+

PR #245930, opened 2026-01-27
[FIX] mrp_subcontracting_dropshipping: use correct account

Whenrow:
A component bought from a vendor and
dropshipped to a subcontractor should the valued into the stock valuation account and not the expense one. In case this component have its "Invoice Policy" set to "ordered quantity", the `_eligible_for_stock_account` method will test if the related stock move are dropshipped https://github.com/odoo/odoo/blob/7436e0cfaf8d4b0c0e0390e8d6a3df404ba240f6/addons/stock_account/models/account_move_line.py#L31-L35 The value `is_dropship` is only set at the validation of the stock move. Due to the invoice policy, the receipt is not validated yet.

close #243015
This commit will rather use the helper `_is_dropshipped()` that only rely on the location and not on the state.

This commit also clean and reenable the related tests.

Current behavior before PR:

---

### DIFF
--- a/addons/mrp_subcontracting_dropshipping/models/stock_move.py
+++ b/addons/mrp_subcontracting_dropshipping/models/stock_move.py
@@ -13,10 +13,25 @@ class StockMove(models.Model):
 
     def _is_dropshipped(self):
         res = super()._is_dropshipped()
-        return res or (
+        return (
+            res
+            # subcontractor -> customer
+            or (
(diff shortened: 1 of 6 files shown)


addons/mrp_subcontracting_dropshipping/models/stock_move.py
amoyaux, line 20: could in common

PR #245931, opened 2026-01-27
[IMP] peppol: Add missing invoice type codes

yagho-odoo:
Before this commit, when re-sending an invoice that was corrected the code for the invoice type code was incorrect, also if the invoice was a down payment and not a full invoice, the invoice type code was also incorrect.
When sending a debit note to a customer, a traceback occurred because the path was not set properly, and the code was also missing. task-5875689

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
@@ -1223,11 +1223,25 @@ class AccountEdiXmlUBL20(models.AbstractModel):
 
     def _add_invoice_header_nodes(self, document_node, vals):
         invoice = vals['invoice']
+        code = None
+        if 'process_type' in vals and vals['process_type'] == 'selfbilling':
+            code = 389
+        # if an invoice was sent before, it has a message uuid, so it's a 'corrected invoice'
+        elif 'peppol_message_uuid' in invoice._fields and invoice.peppol_message_uuid:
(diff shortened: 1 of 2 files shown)


addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py
```diff
@@ -1223,11 +1223,25 @@ def _add_invoice_monetary_totals_vals(self, vals):
 
     def _add_invoice_header_nodes(self, document_node, vals):
         invoice = vals['invoice']
+        code = None
+        if 'process_type' in vals and vals['process_type'] == 'selfbilling':
```
clbr-odoo, line 1227:
```suggestion
        if vals.get('process_type') == 'selfbilling':
```
This should be in an override in the module `account_peppol_selfbilling` in stable.

clbr-odoo (changes requested):
This change should not be part of the UBL(2.0) that is just a syntax definition.
It should be done in BIS3 files that is the semantic.

A change like that must always be backed up with tests.

I discussed with GUHEC. It's not enough to just add the type codes for Debit Note, Corrected Invoice and Downpayment. It's 3 features that are not really supported in the BIS3 at the moment (but they are in Odoo). We should do a whole task for it, support it with a complete test suite, decide what to do if we receive those documents, etc.

The only code that I think should be a "fix" is the selfbilling one, but it's not done at the right place.

PR #245933, opened 2026-01-27
[FIX] mail: avoid removing group when not inside this one

JKE-be:
Before this change, creating a user triggered a write that attempted to remove the group `mail.group_mail_notification_type_inbox` even if the user was not part of it. This unnecessary operation caused an extra invalidation of the stable cache.

### DIFF
--- a/addons/mail/models/res_users.py
+++ b/addons/mail/models/res_users.py
@@ -119,9 +119,11 @@ class ResUsers(models.Model):
 
     def _inverse_notification_type(self):
         inbox_group = self.env.ref('mail.group_mail_notification_type_inbox')
-        inbox_users = self.filtered(lambda user: user.notification_type == 'inbox')
-        inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
-        (self - inbox_users).write({"group_ids": [Command.unlink(inbox_group.id)]})
+        if inbox_users := self.filtered(lambda user: user.notification_type == 'inbox'):
+            inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
+        if notif_users := (self - inbox_users):
+            if to_update := notif_users.filtered(lambda u: inbox_group in u.group_ids):
+                to_update.write({"group_ids": [Command.unlink(inbox_group.id)]})
 
     @api.depends_context("uid")
     def _compute_can_edit_role(self):


addons/mail/models/res_users.py
```diff
@@ -119,9 +119,11 @@ def _compute_im_status(self):
...
     def _inverse_notification_type(self):
         inbox_group = self.env.ref('mail.group_mail_notification_type_inbox')
-        inbox_users = self.filtered(lambda user: user.notification_type == 'inbox')
-        inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
-        (self - inbox_users).write({"group_ids": [Command.unlink(inbox_group.id)]})
+        if inbox_users := self.filtered(lambda user: user.notification_type == 'inbox'):
+            inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
+        if notif_users := (self - inbox_users):
+            if to_update := notif_users.filtered(lambda u: inbox_group in u.group_ids):
+                to_update.write({"group_ids": [Command.unlink(inbox_group.id)]})
```
seb-odoo, line 126: Conditions can be grouped here
```suggestion
        if to_remove := (self - inbox_users).filtered(lambda u: inbox_group in u.group_ids):
            to_remove.write({"group_ids": [Command.unlink(inbox_group.id)]})
```

addons/mail/models/res_users.py
```diff
@@ -119,9 +119,11 @@ def _compute_im_status(self):
 
     def _inverse_notification_type(self):
         inbox_group = self.env.ref('mail.group_mail_notification_type_inbox')
-        inbox_users = self.filtered(lambda user: user.notification_type == 'inbox')
-        inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
-        (self - inbox_users).write({"group_ids": [Command.unlink(inbox_group.id)]})
+        if inbox_users := self.filtered(lambda user: user.notification_type == 'inbox'):
+            inbox_users.write({"group_ids": [Command.link(inbox_group.id)]})
```
seb-odoo, line 123: I'm not sure how the problem is manifesting exactly, but wouldn't we want to also not re-link those that are already correct?
```suggestion
    inbox_users = self.filtered(lambda u: u.notification_type == "inbox")
    if to_add := inbox_users.filtered(lambda u: inbox_group not in u.group_ids):
        to_add.write({"group_ids": [Command.link(inbox_group.id)]})
```

seb-odoo (approved):
Not tested but the fix makes sense.

@robodoo delegate+

PR #245934, opened 2026-01-27
[FIX] point_of_sale: ensure prices are aligned between back & front

davidmonnom:
Before this commit, prices were not always aligned between the backend and the frontend, when syncing an order in Self Order mode the price of combo products could differ between backend and frontend.

This commit ensures that combo products prices are always aligned between backend and frontend in Self Order mode.


stva-odoo (commented):
robodoo r+

PR #245937, opened 2026-01-27
[FIX] hr_recruitment: use template mail_from in refuse mail

mathcoutant:
### Issue:
When refusing an application, the system will always use the user's email to send the refusal, even if the template specifies an email to send from.

### Steps to reproduce:
- Recruitment > Configuration > Refuse Reasons
- Go to one of their template (i.e. "Recruitment: Refuse")
- in Settings add a "Send From" email
- Recruitment > Applications, click on an application
- Click "Refuse"
- Select the refuse reason with the template
- Refuse
- The mail is sent with the user email

### Cause:
`_prepare_mail_values()` only returns `self.env.user.email_formatted`.

### Solution:
If there is a template with `email_from`, we use this, otherwise `self.env.user.email_formatted`

opw-5356662

### DIFF
--- a/addons/hr_recruitment/tests/test_recruitment.py
+++ b/addons/hr_recruitment/tests/test_recruitment.py
@@ -339,6 +339,35 @@ class TestRecruitment(TransactionCase):
             app_2
         )
 
+    def test_applicant_refuse_mail_from_template(self):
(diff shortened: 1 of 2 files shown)


Einar-eian (approved):
Thank you for your work 😄 

@robodoo r+

PR #245939, opened 2026-01-27
[PERF] stock: Improve inventory product import

nguyenntph:
Problem:
When importing products in Inventory, _compute_quantities_dict fetches
data from all internal stock.location. For databases with many internal
locations (e.g., 37K), the query using LIKE ANY with a large array of
parent paths is very slow.

Solution:
- Replace LIKE ANY(%s) pattern matching with a semi-join subquery using
  EXISTS, allowing the query to terminate on the first match.
- Add an index on (parent_path, id) to convert sequential scans into
  index-only scans.

Benchmark:
|Number of rows| Before | After |
|--------------|--------|-------|
|37K           | 4.xxs  | 1.xxs |

Simplied plan (a few hundred records instead of 37K)
|Before|After|
|-|-|
|https://explain.dalibo.com/plan/8gb1257c6712age5|https://explain.dalibo.com/plan/5ef9c3g7aab3dfag|

Related ticket:
opw-5500169

---

### DIFF
--- a/addons/stock/models/product.py
+++ b/addons/stock/models/product.py
@@ -395,7 +395,16 @@ class ProductProduct(models.Model):
         elif locations:
             alias = locations._table + '_inner'
             paths_query = Query(locations.env, alias, SQL.identifier(locations._table))
-            paths_query.add_where(alias + '.parent_path LIKE ANY(%s)', [[loc.parent_path + '%' for loc in locations]])
+            paths_query.add_where(SQL(
+                """EXISTS (
+                    SELECT 1
+                      FROM stock_location parent
+                     WHERE parent.id IN %s
+                       AND %s LIKE parent.parent_path || '%%'
+                )""",
+                tuple(locations.ids),
+                SQL.identifier(alias, 'parent_path'),
+            ))
             loc_domain = Domain('location_id', 'in', paths_query)
             # The condition should be split for done and not-done moves as the final_dest_id only make sense
             # for the part of the move chain that is not done yet.
--- a/addons/stock/models/stock_location.py
+++ b/addons/stock/models/stock_location.py
@@ -98,6 +98,7 @@ class StockLocation(models.Model):
         'check(cyclic_inventory_frequency >= 0)',
         'The inventory frequency (days) for a location must be non-negative',
     )
+    _parent_path_id_idx = models.Index("(parent_path, id)")
 
     @api.depends('name', 'location_id.complete_name', 'usage')
     @api.depends_context('formatted_display_name')
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -191,10 +191,11 @@ class ProductProduct(models.Model):
     # -------------------------------------------------------------------------
 
     def _change_standard_price(self, old_price):
+        product_values = []
         for product in self:
             if product.cost_method == 'fifo' or product.standard_price == old_price.get(product):
                 continue
-            self.env['product.value'].sudo().create({
+            product_values.append({
                 'product_id': product.id,
                 'value': product.standard_price,
                 'company_id': product.company_id.id or self.env.company.id,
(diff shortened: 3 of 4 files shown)


addons/stock/models/product.py
```diff
@@ -395,7 +395,15 @@ def _get_domain_locations_new(self, location_ids) -> tuple[Domain, Domain, Domai
         elif locations:
             alias = locations._table + '_inner'
             paths_query = Query(locations.env, alias, SQL.identifier(locations._table))
-            paths_query.add_where(alias + '.parent_path LIKE ANY(%s)', [[loc.parent_path + '%' for loc in locations]])
+            paths_query.add_where(SQL(
+                """EXISTS (
+                    SELECT 1 FROM stock_location parent
+                    WHERE parent.id IN %s
+                    AND %s LIKE parent.parent_path || '%%'
+                )""",
```
pivi-odoo, line 404:
```suggestion
            paths_query.add_where(SQL(
                """EXISTS (
                    SELECT 1
                      FROM stock_location parent
                     WHERE parent.id IN %s
                       AND %s LIKE parent.parent_path || '%%'
                )""",
```
`FROM` on another line please :)

(nitpick): At Odoo, the SQL code normally follows the [Axis](https://www.sqlinform.com/river-style-vs-axis-style/#:~:text=Axis%20formatting%20is%20a%20style,compare%20elements%20within%20the%20query.) formatting style.
Although people also use Left aligned, but it's not the suggested way.
Axis style is also the format enforced for upgrade scripts.

pivi-odoo (commented):
LGTM, I agree with the approach.

But the explanation is kinda off.
The speed up comes from:
1) inserting in the query a list of `id` instead of a list of `parent_path`, less content to inject.
2) The `EXISTS` nodes, which are usually transformed into `Anti-join/Semi-join` which can end early their processing if they find the result they need.

The after-query still does a Seq.Scan on the locations 4 times, it's just that for many of the leafs, it's actually never executed, because it satisfied the upper node with the results coming from scanning the Pkey index of `stock_location` (thanks to your `parent.id` clause).

I guess, just to be sure, we should probably add a composite index for the `parent_path`, but adding the `id` to it.

```sql
CREATE INDEX stock_location__parent_path_id_idx ON stock_location(parent_path, id);
```

Could you test with this in your case do you have a *some* speed-up? I doubt, since 3/4 of the branches are never executed. But this index should transform the Seq.Scan (or Parallel Seq.Scan) into an Index Only Scan, which is significantly smaller and faster, even if not executed, in case it is, it should provide less contention on the postgres' buffers.

What do you think?

addons/stock/models/product.py
nguyenntph: Thanks. The style is now updated.

nguyenntph (commented):
Thanks @pivi-odoo, that's a good idea. In the [plan](https://explain.dalibo.com/plan/5ef9c3g7aab3dfag), with index, it became a Parallel Index Only Scan instead of a Parallel Seq Scan as you shared. Locally with the index, the difference is around ±0.2s, but I think it's worth a try to check with the customer staging database afterward.

pivi-odoo:
@nguyenntph it seems there is a check I wasn't aware of `_check_parent_path` that enforces to always have `index=True` on the field if it's a `parent_path`, which doesn't make sense if it's covered like our new definition... and there is no way to override this.
So I guess we can't change the definition, so both index are required. I will see in master to move this check in `test_lint` with some allow list.

Sorry for the back'n'forth :S

pivi-odoo (approved):
LGTM, thank you for your work @nguyenntph !

Hello @ajf-odoo, 
This [PERF] PR has already been pre-reviewed, and the CI is 🟢.
If you could take a look when you have the time :)

Thank you in advance.

amoyaux:
It seems okay. I would say at first that it's strange that we goes through this code with an import and it could be avoided. But generally speaking, it's call so often that a lot of features would benefits from it. If you are sure there is no drawback with it, it's okay for me.

PR #245940, opened 2026-01-27
[FIX] website_sale: allow eu vat fix in address submit

mma1377:
### Issue:
EU vat fix cause address submission to fail.

#### Steps to reproduce:
1- `base_vat` should be installed.
2- In portal contact, set the country to BE and VAT to `0477472701`.
3- Create an invoice for portal user.
4- Navigate to shop and add a product to cart, then checkout.
5- Edit address.
6- After filling, click on save address.

Even though the VAT is not modified in form, and it is not modifiable
the address fails to save with error:
`Changing VAT Number is not allowed.`

### Cause:
After address submission The vat is fixed in here:
https://github.com/odoo/odoo/blob/82a2c5305d65027ffb263f90c4e00f7e95f98b05/addons/website_sale/controllers/main.py#L1396-L1406

This converts given vat `0477472701` to `BE0477472701`.

Which makes `address_values['vat'] != partner_sudo.vat`:
https://github.com/odoo/odoo/blob/82a2c5305d65027ffb263f90c4e00f7e95f98b05/addons/website_sale/controllers/main.py#L1472-L1482

opw-5437586

### DIFF
--- a/addons/website_sale/controllers/main.py
+++ b/addons/website_sale/controllers/main.py
@@ -1393,15 +1393,8 @@ class WebsiteSale(payment_portal.PaymentPortal):
             elif value:  # The value cannot be saved on the `res.partner` model.
                 extra_form_data[key] = value
 
-        if (
-            hasattr(ResPartner, 'check_vat')  # The `base_vat` module is installed.
-            and address_values.get('vat')
-            and address_values.get('country_id')
-        ):
-            address_values['vat'] = ResPartner.fix_eu_vat_number(
-                address_values['country_id'],
-                address_values['vat'],
-            )
+        if address_values.get('vat'):
+            address_values['vat'] = self._fix_eu_vat_number(address_values['vat'], address_values.get('country_id'))
 
         return address_values, extra_form_data
 
@@ -1469,10 +1462,15 @@ class WebsiteSale(payment_portal.PaymentPortal):
                     " the account settings or contact your administrator."
                 ))
 
+            if vat := partner_sudo.vat:
+                vat = self._fix_eu_vat_number(
+                    vat,
+                    partner_sudo.country_id.id or address_values['country_id'],
+                )
             # Prevent changing the VAT number if invoices have been issued.
             if (
                 'vat' in address_values
-                and address_values['vat'] != partner_sudo.vat
+                and address_values['vat'] != vat
                 and not partner_sudo.can_edit_vat()
             ):
                 invalid_fields.add('vat')
@@ -2281,6 +2279,19 @@ class WebsiteSale(payment_portal.PaymentPortal):
                 domain += [('product_id.product_tmpl_id', '=', int(product_template_id))]
             request.env['website.track'].sudo().search(domain).unlink()
         return {}
+
+    @staticmethod
+    def _fix_eu_vat_number(vat, country_id):
+        ResPartner = request.env['res.partner']
+        if (
+            hasattr(ResPartner, 'check_vat')  # The `base_vat` module is installed.
+            and country_id
+        ):
+            vat = ResPartner.fix_eu_vat_number(
+                country_id,
+                vat,
+            )
+        return vat
 
     @staticmethod
     def _populate_currency_and_pricelist(kwargs):
--- a/addons/website_sale/tests/test_address.py
+++ b/addons/website_sale/tests/test_address.py
@@ -658,6 +658,51 @@ class TestCheckoutAddress(BaseUsersCommon, WebsiteSaleCommon):
             "Tax should no longer change after order confirmation",
         )
 
+    def test_13_shop_address_submit_eu_vat_number(self):
+        if not hasattr(self.env['res.partner'], 'check_vat'):
+            self.skipTest("base_vat is not installed")
+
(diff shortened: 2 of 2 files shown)


addons/website_sale/controllers/main.py
```diff
@@ -2282,6 +2282,18 @@ def products_recently_viewed_delete(self, product_id=None, product_template_id=N
...
+    def _fix_eu_vat_number(vat, country_id, ResPartner):
+        if (
+            hasattr(ResPartner, 'check_vat')  # The `base_vat` module is installed.
+            and country_id
+        ):
+            vat = ResPartner.fix_eu_vat_number(
+                country_id,
+                vat,
+            )
+        return vat
```
Feyensv:
```suggestion
    def _fix_eu_vat_number(vat, country_id):
        ResPartner = request.env['res.partner']
        if (
            hasattr(ResPartner, 'check_vat')  # The `base_vat` module is installed.
            and country_id
        ):
            vat = ResPartner.fix_eu_vat_number(
                country_id,
                vat,
            )
        return vat
```

no need to give the res partner class

addons/website_sale/controllers/main.py
```diff
@@ -1393,15 +1393,8 @@ def _parse_form_data(self, form_data):
...
-        if (
-            hasattr(ResPartner, 'check_vat')  # The `base_vat` module is installed.
-            and address_values.get('vat')
-            and address_values.get('country_id')
-        ):
-            address_values['vat'] = ResPartner.fix_eu_vat_number(
-                address_values['country_id'],
-                address_values['vat'],
-            )
+        if (address_values.get('vat')):
```
Feyensv:
```suggestion
        if address_values.get('vat'):
```

addons/website_sale/controllers/main.py
```diff
@@ -1469,10 +1462,17 @@ def _validate_address_values(
                     " the account settings or contact your administrator."
                 ))
 
+            vat = partner_sudo.vat
+            if (vat):
```
Feyensv:
```suggestion
            if vat := partner_sudo.vat:
```

Feyensv (commented):
little nitpickings to simplify the logic, but the fix LGTM 👌

addons/website_sale/tests/test_address.py
```diff
@@ -658,6 +658,52 @@ def test_12_recompute_taxes_on_address_change(self):
...
         )
 
+    def test_13_shop_address_submit_eu_vat_number(self):
+        if not hasattr(self.env['res.partner'], 'check_vat'):
+            self.skipTest("base_vat is not installed")
+
+        partner = self.env['res.partner'].create({
+            'name': 'test partner',
+            'vat': '0477472701',
+            'country_id': self.env['res.country'].search([('code', '=', 'BE')]).id,
```
Feyensv:
```suggestion
            'country_id': self.country_id
```

addons/website_sale/tests/test_address.py
```diff
@@ -658,6 +658,52 @@ def test_12_recompute_taxes_on_address_change(self):
...
+        })
+        user = self.env['res.users'].create({
+            'name': 'test user',
+            'login': 'test',
+            'email': 'test@test.com',
+            'partner_id': partner.id,
+        })
+
+        dumb_product = self.env['product.product'].create({'name': 'test'})
```
Feyensv: You can use `self.product` (from WebsiteSaleCommon -> ProductCommon)

addons/website_sale/tests/test_address.py
```diff
@@ -658,6 +658,52 @@ def test_12_recompute_taxes_on_address_change(self):
...
+        invoice.action_post()
+        self.assertFalse(partner.can_edit_vat())
+
+        so = self._create_so(partner_id=partner.id)
+        self.default_address_values.update({
+            'vat': partner.vat,
+            'name': partner.name,
+            'email': partner.email,
+        })
+        self.default_address_values['vat'] = '0477472701'
```
Feyensv:
```suggestion
        address_values = {
            **self.default_address_values,
            'vat': '0477472701',
            'name': partner.name,
            'email': partner.email,
        })
```
Kinda confusing to modify something on `self`, and to update the `vat` twice 🤔

addons/website_sale/tests/test_address.py
```diff
@@ -658,6 +658,52 @@ def test_12_recompute_taxes_on_address_change(self):
...
+        env = api.Environment(self.env.cr, user.id, {})
+        with MockRequest(env, website=self.website.with_user(user.id), sale_order_id=so.id) as req:
```
Feyensv:
```suggestion
        with MockRequest(env, website=self.website.with_env(env), sale_order_id=so.id) as req:
```

Feyensv (approved):
some last comments for the tests (should have taken a look earlier too, sorry).

It would have been better if we didn't touch the partner VAT at all in this case, but I guess it would require way more changes so we can keep this solution 👌 

@robodoo delegate+

Feyensv (approved):

mma1377:
robodoo r+

PR #245941, opened 2026-01-27
[FIX] point_of_sale: reconcile reversal

robinengels:
If you made a PoS order paid with the customer account payment method, then close the session. Then invoice the order, a reversal entry would be created but it would not be reconciled at all.
As no payment was made a this point it should be reconciled with the original Customer Account entry from the pos session, this way only the invoice remains open.

Steps to reproduce:
-------------------
* Create a PoS order and pay with the customer account payment method
* Close the session
* Go on the order and create the invoice
> Observation: The reversal move is not reconciled and the customer
  has two outstanding credits.

Why the fix:
------------
When creating the invoice we reconcile the reversal entry with the original customer account entry from the pos session. This way only the invoice remains open.

opw-5268042

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -1065,15 +1065,28 @@ class PosOrder(models.Model):
         })
         reversal_entry.action_post()
 
-        pos_account_receivable = self.company_id.account_default_pos_receivable_account_id
-        account_receivable = self.payment_ids.payment_method_id.receivable_account_id
-        reversal_entry_receivable = reversal_entry.line_ids.filtered(lambda l: l.account_id in (pos_account_receivable + account_receivable))
-        payment_receivable = payment_moves.line_ids.filtered(lambda l: l.account_id in (pos_account_receivable + account_receivable))
-        lines_to_reconcile = defaultdict(lambda: self.env['account.move.line'])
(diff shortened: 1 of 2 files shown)


addons/point_of_sale/models/pos_order.py
```diff
@@ -1065,15 +1065,26 @@ def _create_misc_reversal_move(self, payment_moves):
...
+        else:
+            candidate_lines |= self.session_move_id.line_ids.filtered(
+                lambda l: l.partner_id == partner and l.account_id == partner.property_account_receivable_id
+            )
+
+        lines_to_reconcile = candidate_lines.filtered(
+            lambda l: l.account_id in accounts and not l.reconciled
+        )
+        for account in lines_to_reconcile.account_id:
+            lines_to_reconcile.filtered(lambda l: l.account_id == account).reconcile()
```
stva-odoo: Minor:
```suggestion
        lines_by_account = {}
        for line in candidate_lines:
            if line.account_id in accounts and not line.reconciled:
                lines_by_account.setdefault(line.account_id, self.env['account.move.line'])
                lines_by_account[line.account_id] |= line        
        for lines in lines_by_account.values():
            lines.reconcile()
```

stva-odoo (commented):
robodoo r+

PR #245964, opened 2026-01-27
[FIX] microsoft_calendar: allow website_appointments to sync through Outlook cron

jelu-odoo:
Before this change, the "Outlook: synchronization" cron would not create calendar events on Outlook's side in _sync_odoo2microsoft due to a filter for calendar.events written to within 5 minutes of microsoft_last_sync_date, when _sync_data is not called when a calendar.event is created, such as through website.appointment. microsoft_last_sync_date was set to datetime.now() at the beginning of _sync_microsoft_calendar, which would skip a large period of time between the last sync and now, if the only syncs were triggered through cron, and not _sync_data (by opening the calendar app).

To reproduce, 

Default "Outlook: synchronization" is ran every 12 hours. 

1) Calendar event is synced through "Outlook: synchronization" cron at 00:00, setting microsoft_last_sync_date to 00:00
2) A website.appointment is created for a resource with Outlook calendar sync enabled any time between 00:01 - 11:54. 
3) "Outlook: synchronization" runs again at 12:00, which sets microsoft_last_sync_date to 12:00, and filters out calendar.events based on their write_dates in _extend_microsoft_domain that need syncing outside of 11:55 to 12:00. 

This change removes setting of microsoft_last_sync_date at the beginning of _sync_microsoft_calendar, where we need to use the old value before setting it at the end of _sync_microsoft_calendar.

opw-5212908

---

### DIFF
--- a/addons/microsoft_calendar/tests/test_create_events.py
+++ b/addons/microsoft_calendar/tests/test_create_events.py
@@ -396,6 +396,9 @@ class TestCreateEvents(TestCommon):
         mock_get_events.return_value = ([], None)
 
         # Synchronize local event with Outlook after updating it locally.
+        # Set user's microsoft_last_sync_date to now() to capture the event with the domain in
+        # _extend_microsoft_domain.
+        self.organizer_user.microsoft_last_sync_date = fields.datetime.now()
         self.organizer_user.with_user(self.organizer_user).sudo()._sync_microsoft_calendar()
         self.call_post_commit_hooks()
         event.invalidate_recordset()
@@ -780,6 +783,65 @@ class TestCreateEvents(TestCommon):
             event.invalidate_recordset()
             mock_insert.assert_called_once()
             self.assertEqual(mock_insert.call_args[0][0]['subject'], event.name)
+
+    @patch.object(MicrosoftCalendarService, 'get_events')
+    @patch.object(MicrosoftCalendarService, 'insert')
+    def test_sync_website_appointments_through_cron(self, mock_insert, mock_get_events):
+        """Check old events created after the last sync are synced."""
+        # mock setup
+        def _mock_calendar_token(user, *args, **kwargs):
+            return f'TOKEN_{user.id}' if user == self.organizer_user else False
+
+        event_uid = 1
+
+        def _mock_msft_insert(*args, **kwargs):
+            nonlocal event_uid  # ensure we have separate ids, for sanity
+            event_uid_string = f"msft_event_id_test_sync_website_appointments_through_cron_{event_uid}"
+            event_uid += 1
+            return (event_uid_string, event_uid_string)
+
+        mock_insert.side_effect = _mock_msft_insert
+        mock_get_events.return_value = ([], None)
+
+        t_now = datetime(2020, 5, 8, 8, 0, 0)
+        t_minus_12h = t_now - timedelta(hours=12)
+        t_minus_11h = t_now - timedelta(hours=11)
+
+        # test vals setup
+        CalendarEvent = self.env["calendar.event"].with_user(self.attendee_user)
+        vals_list = self.simple_event_values | {
+            'partner_id': self.organizer_user.partner_id.id,
+            'user_id': self.organizer_user.id,
+            'start': '2020-05-09 14:00',
+            'stop': '2020-05-09 15:00',
+        }
+
+        # set microsoft_last_sync_date for the user at t=-12h.
+        with freeze_time(t_minus_12h), patch.object(User, '_get_microsoft_calendar_token', _mock_calendar_token):
+            self.organizer_user.with_user(self.organizer_user).restart_microsoft_synchronization()
+        self.env.cr.postcommit.clear()
+
+        # users without sync create events at last sync, 1 hour later, and 12 hours later
+        # the organizer syncs those regularly but not always exactly at that time
+        # if the user syncs after 12 hours, the event created 1 hour after the previous sync should sync too
+        events = CalendarEvent
+        for time, should_sync, expected_insert_count in zip([t_minus_12h, t_minus_11h, t_now], [True, False, True], [1, 1, 3]):
+            with freeze_time(time), patch.object(User, '_get_microsoft_calendar_token', _mock_calendar_token):
+                event = CalendarEvent.create(vals_list)
+                events |= event
+                if should_sync:
+                    # sudo() is for consistency, even though the case passes without it here.
+                    self.organizer_user.with_user(self.organizer_user).sudo()._sync_microsoft_calendar()
+                else:
+                    self.assertFalse(event.microsoft_id)
+                self.env.cr.postcommit.run()  # run the actual sync synchronously
+                self.assertEqual(mock_insert.call_count, expected_insert_count)
+
+        events.invalidate_recordset()  # postcommit used a different cursor, invalidate to force re-fetch the values from DB
+        self.assertTrue(
+            all(events.mapped('microsoft_id')),
+            "This event should be synced. It was written to after the user's microsoft_last_sync_date."
+        )
 
 class TestSyncOdoo2MicrosoftMail(TestCommon, MailCommon):
     @classmethod
--- a/addons/microsoft_calendar/models/res_users.py
+++ b/addons/microsoft_calendar/models/res_users.py
@@ -79,7 +79,6 @@ class User(models.Model):
 
     def _sync_microsoft_calendar(self):
         self.ensure_one()
-        self.sudo().microsoft_last_sync_date = fields.datetime.now()
         if self._get_microsoft_sync_status() != "sync_active":
             return False
 


reth-odoo:
@jelu-odoo Just to confirm, does this mean this method effectively does nothing except in the rare case where you luckily create an event within 5 minutes of it being called?

addons/microsoft_calendar/tests/test_create_events.py
reth-odoo: Doesn't this test imply the previous behavior is the wanted behavior?
i.e. we don't think old events

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -396,6 +396,7 @@ def test_sync_create_update_single_event(self, mock_insert, mock_get_events):
         mock_get_events.return_value = ([], None)
 
         # Synchronize local event with Outlook after updating it locally.
+        self.organizer_user.microsoft_last_sync_date = fields.datetime.now()
```
reth-odoo, line 401: Can we do that after the freeze_time bit?

reth-odoo (commented):
I see so in 9132bde68867c3e54da5064eff115a7984985c6b we switched from setting this only if it's the first sync for this user to always doing it.

It seems like we originally did that to avoid syncing all events assigned to a user the first time they sync. https://github.com/odoo/odoo/commit/69d23262f847a173bde8d2c2efd95c3f8c429a43

Although it's not tested, I feel like that case is still broken. Could we cover it again?
Or at least make sure the test covers it?

addons/microsoft_calendar/tests/test_create_events.py
jelu-odoo: > @jelu-odoo Just to confirm, does this mean this method effectively does nothing except in the rare case where you luckily create an event within 5 minutes of it being called?

Hi @reth-odoo! Sorry I forgot to check back on this issue. Yes, if anything creates events where it doesn't go through the javascript crud method -- it appears that _sync_microsoft_calendar() only gets called in two ways: 
1) CRUD in microsoft_calendar_model.js in async syncMicrosoftCalendar()
2) through "Outlook: synchronization" cron

So, if website.appointments are made, it bypasses 1) and will rely on 2) to sync the event. However, the domain filters out the event if it wasn't updated within 5 minutes of the cron running.

This testcase was broken due to the with freeze_time(ten_minutes_after_creation): -- the simple_event's write date is t=0, but the user's microsoft_last_sync_date is t=+10min. When we go back to t=0 (exiting with: ... ) and run _sync_microsoft_calendar, _extend_microsoft_domain() filters it out since, we look at the calendar.events with write_dates >= (user.microsoft_last_sync_date - timedelta(5m)), which are events with a future write date >= t=+5min, which there are none. 

This line in the test case resets the microsoft_last_sync_date for that user to be t=0, so that we don't check for future write_dates. Alternatively, this could be 

self.organizer_user.with_user(self.organizer_user).sudo().restart_microsoft_synchronization()
self.organizer_user.with_user(self.organizer_user).sudo()._sync_microsoft_calendar()

I'm not too sure what you mean by " I feel like that case is still broken. " Could you specify which case?

reth-odoo: > could you specify the case

- have a bunch of events in odoo already, in the past and future
- sync your user with microsoft calendar for the very first time
-> Existing events should *not be synced (maybe syncing those from the last 5 minutes is fine)

It seems like that was originally the idea of setting the "last sync" to now from there: 69d23262f847a173bde8d2c2efd95c3f8c429a43

This test only seems to cover "restarting" but not "first sync". In restart, the restart method sets microsoft_last_sync_date, but for first thing I'm not sure we really ever do. If we don't, microsoft_last_sync_date will be False and we will effectively sync the whole calendar of people instead of just new events.

jelu-odoo: It seems microsoft_last_sync_date is set inside of the restart_microsoft_synchronization() method -- this method itself is both the start and the restart, in the javascript onMicrosoftSyncCalendar(). It should not be unset when we have begun syncing with Outlook. 

It seems I had to create a new class outside of the TestCreateEvents class, because it requires patching _get_microsoft_calendar_token to have different two different return values in the same test. The test also clears the postcommit._funcs deque, as call_post_commit_hooks() runs on a copy of postcommit._funcs and doesn't clear it, which is not what we expect. 

Let me know what you think, thanks!

addons/microsoft_calendar/tests/common.py
```diff
@@ -12,7 +12,6 @@
 
 from odoo.addons.microsoft_calendar.models.microsoft_sync import MicrosoftSync
 
-
```
reth-odoo, line 15: \-

addons/microsoft_calendar/tests/test_create_events.py
reth-odoo: Ok for me if we always call restart, thanks for looking into it :)

reth-odoo (commented):
@jelu-odoo 
Sorry to make you work for nothing but if it already works and we only sync via a call to restart, no need to add a specific test.

However can we add a step at the end of the existing test where we fetch records older than X minutes old? To cover the issue we're fixing.

After that lgtm ^^'

addons/microsoft_calendar/tests/common.py
jelu-odoo: @reth-odoo Hi, sorry for the delay -- I've made some changes and added a test case to capture events between the microsoft_user.last_sync_date until now. Let me know what you think, thanks!

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+        t_minus_12h = t_now - timedelta(hours=12)
+        t_minus_11h = t_now - timedelta(hours=11)
+
+        # Create event 12 hours ago and sync, setting microsoft_last_sync_date for the user.
+        # We call restart_microsoft_synchronization() to set a microsoft_last_sync_date for the user.
+        with freeze_time(t_minus_12h):
+            self.organizer_user.with_user(self.organizer_user).restart_microsoft_synchronization()
+            event_t_minus_12h = CalendarEvent.create(vals_list)
+            # clear() the postcommit._funcs deque to not call _microsoft_insert on the same event.
+            self.env.cr.postcommit._funcs.clear()
```
reth-odoo: Can just call `postcommit.clear` to also clear data etc..

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+            self.organizer_user.with_user(self.organizer_user).sudo()._sync_microsoft_calendar()
```
reth-odoo: Why sudo?

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+        self.call_post_commit_hooks()
+        event_t_minus_12h.invalidate_recordset()
+
+        # clear() the postcommit._funcs deque again, since it was added by _sync_microsoft_calendar().
+        self.env.cr.postcommit._funcs.clear()
```
reth-odoo: Can just call postcommit.run() which will also clear it

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+            self.env.cr.postcommit._funcs.clear()
+            self.call_post_commit_hooks()
+
+        # Again, we should have only called assert once here, ignoring the above event.
+        mock_insert.assert_called_once()
+
+        # Create event now to be synced, without clearing postcommit._funcs.
+        event_t_now = CalendarEvent.create(vals_list)
+
+        # We cannot call cron _sync_all_microsoft_calendar due to the method using self.env.cr.commit()
```
reth-odoo: Would remove this, we can always add `if not testing` in that method if we need to use it in a test later, it's a little irrelevant if the per-user sync is sufficient for this test

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+        # So instead, just call _sync_microsoft_calendar().
+        # This should capture both event_t_minus_11h and event_t_now and sync.
+        self.organizer_user.with_user(self.organizer_user).sudo()._sync_microsoft_calendar()
+        self.call_post_commit_hooks()
+        event_t_minus_11h.invalidate_recordset()
+        event_t_now.invalidate_recordset()
```
reth-odoo: Are all these invalidate_recordset really useful here?

addons/microsoft_calendar/tests/test_create_events.py
```diff
@@ -781,6 +784,72 @@ def test_old_db_odoo2microsoft_sync_previously_created_events(self, mock_insert,
...
+        # Set up syncing return values.
+        event_id = "123"
+        event_iCalUId = "456"
+        mock_insert.return_value = (event_id, event_iCalUId)
+        mock_get_events.return_value = ([], None)
+
+        # Setup t=0, t=-12h, t=-11h.
+        t_now = datetime.now()
+        t_minus_12h = t_now - timedelta(hours=12)
+        t_minus_11h = t_now - timedelta(hours=11)
```
reth-odoo: @jelu-odoo likewise, sorry for the delay. Feel free to ping me if needed too.

Could we create them in a loop like:

```py
self.organizer_user.with_user(self.organizer_user).restart_microsoft_synchronization()
self.env.postcommit.clear()

meetings = CalendarEvent
for time, expected_insert_count in zip([t_now, t_minus_12h, t_minus_11h], [1, 1, 2]):
    with freeze_time(time):
        meetings += CalendarEvent.create(vals_list)
        self.env.cr.postcommit.run()
        meetings.invalidate_recordset()  # necessary? Can we specify a field?
        mock_insert.assert_called(n_times=expected_insert_count)

t_now_event, t_minus_12h_event, t_minus_11h_event = meetings
# specific asserts here if relevant
```

It would be best to keep this relatively compact as quite specific and repeats a lot, helps readability.

reth-odoo (commented):
A few more comments on the new test

addons/microsoft_calendar/tests/test_create_events.py
jelu-odoo: I don't think this would work as what's intended in this test case; we specifically do not want to run the postcommit for t_minus_11h until t_now is inserted, which is why the order of clearing the postcommit._funcs is typed as such.

I wasn't able to find n_times as a valid parameter, or is this pseudocode? I would definitely like to have used a called n_times assertion, though but I wasn't able to find an example of this!

addons/microsoft_calendar/tests/test_create_events.py
jelu-odoo: Yes, if we don't call invalidate_recordset, the test case fails, when the event was truly inserted (microsoft_insert called). I followed the above test cases' pattern of calling 
call_post_commit_hooks() -> event.invalidate_recordset().

addons/microsoft_calendar/tests/test_create_events.py
jelu-odoo: All other cases of _sync_microsoft_calendar in this file is with sudo(), and in this particular test case the Outlook synchronization cron will usually be run by OdooBot, but you are right that this test cases passes without sudo(). Could you help explain what the proper use of sudo() is here?

addons/microsoft_calendar/tests/test_create_events.py
reth-odoo: > we specifically do not want to run the postcommit

I see
But that makes the test very artificial then, as it would normally always run. Can't we somehow mock something closer to the real case where there's no postcommit scheduled in the first place?

> no n_times

Ah yes my bad that's a different method from HttpCase, I guess you could assert based on `mock_insert.call_count`

addons/microsoft_calendar/tests/test_create_events.py
reth-odoo: Indeed :thinking: 
That seems insane to me to have sudo so high up in the chain. Either is fine for the then I'd say but if keeping sudo I'd add a comment to say it's only done for consistency.

addons/microsoft_calendar/tests/test_create_events.py
reth-odoo: Oh, I guess because postcommit uses a different cursor. Can you just add a comment for that then?

I know pre-existing tests do a lot of odd stuff but I'd prefer for new tests to be particular about non-obvious framework manipulation.

addons/microsoft_calendar/tests/test_create_events.py
jelu-odoo: I don't think it's artificial, since it captures what happens if an event is created through other means, like website.appointment, which is why I clear the postcommit for t_minus_11h and only run after we sync t_now through cron. 

I've reworked it to include the loop and added some comments :+1:

reth-odoo: In appointment isn't it because we modify the event with sudo/a user that does not sync?
postcommit always runs after a commit

jelu-odoo: Sort of -- any user (public, portal, internal) without a microsoft_calendar_token that creates the event for another user with sync enabled will not call insert. Since the creation user doesn't have a token, microsoft_service.insert is never called in the postcommit.

```py
@after_commit
    def _microsoft_insert(self, values, timeout=TIMEOUT):
        """
        Once the event has been really added in the Odoo database, add it in the Outlook calendar.

        Note that all self attributes to use in this method must be provided as method parameters because
        'self' may have been modified between the call of '_microsoft_insert' and its execution,
        due to @after_commit decorator.
        """
        if not values:
            return
        microsoft_service = self._get_microsoft_service()
        sender_user = self._get_event_user_m()
        with microsoft_calendar_token(sender_user.sudo()) as token:
            if token:
                self._ensure_attendees_have_email()
                event_id, uid = microsoft_service.insert(values, token=token, timeout=timeout)
                self.with_context(dont_notify=True).write({
                    'microsoft_id': event_id,
                    'ms_universal_event_id': uid,
                    'need_sync_m': False,
                })
```

reth-odoo: @jelu-odoo so would it be possible to rework the test slightly to use different users instead of manually managing post-commit hooks?

jelu-odoo: Yes, that's possible, but it seems to require a new class without the _get_microsoft_calendar_token patch, which I did in a previous amend. 

I wrote it as such for consistency, since there are tests in this class for the sync start (should be init to tokenless users). This class patch seems to take precedence over re-patching the same method for a test method (unless you know of another way to do this?), so to me, this would need to be in a new class without this patch. 

I'm manually managing them here due to the patch on _get_microsoft_calendar_token for this test class. Let me know if we should move this to a separate class and I can do that.

reth-odoo:
I made a few tweaks to the test so we can avoid some back and forth, but overall seemed fine thanks :)

reth-odoo:
@jeh-odoo :wave: quite urgent one that's been left aside for a while now, if you can have a look this week, would be neat :)

jeh-odoo:
LGTM! Once the style is resolved you can :rocket: 
@robodoo delegate=@reth-odoo

jelu-odoo:
Added a line to resolve ci/style check

reth-odoo:
danke

@robodoo r+

jelu-odoo:
@reth-odoo Hi, there seems to be an issue with the test case -- the entire test case passes without the changed code on this PR (and the forwardports as you noticed), and the forwardported merges will not have the fix. 

What's the protocol of fixing this error since this has been merged already? 

t_now being set to a time in the past fails to capture the write_date near "literal datetime.now()" when we come to _extend_microsoft_domain, since all events are written to after these dates. freeze_time does not push the write date back, unless we also patch the cursor's time, so mock_datetime_and_now should be used instead.

reth-odoo:
:/

good catch
@jelu-odoo I'll apply a fix in the remaining forward ports and we'll backport it here and in 18.2

PR #246032, opened 2026-01-28
[FIX] sale_management: fix indeterministic sign tour

nipl-odoo:
This commit fixes an indeterministic sign tour that could run either inside an iframe or directly in the main page. Since the execution context was unpredictable, both selectors were added to handle both cases reliably.

runbot error-238443

---


Feyensv (approved):
robodoo delegate+

PR #230712, opened 2025-10-09
[FIX] sale_pdf_quote_builder: prevent error when uploading invalid pdf file

utma-odoo:
Currently, an error occurs when uploading `encrypted or incomplete` 
PDF files (missing EOF marker) while creating a quotation document 
header or footer.

**Steps to reproduce:**
- Install the `sale_pdf_quote_builder` module.
- Navigate to: Sales > Configuration > Headers/Footers.
- Upload encrypted file [1], or incomplete file [2].

**Error:**
`PyPDF2.errors.DependencyError: PyCryptodome is required for AES algorithm`
`PyPDF2.errors.PdfReadError: EOF marker not found`

**Root cause:**
At [3], `_get_form_fields_from_pdf` and `_ensure_document_not_encrypted` directly 
call `pdf.PdfFileReader`, when it fails to read or decrypt the file, Python raises 
an error.

**Fix:**
This commit prevents errors when users upload unreadable or encrypted PDF files.

[1]:
https://drive.google.com/file/d/1moSlwXHkqcV6_7zHBNhLMLi-9Ye_xDGJ/view?usp=sharing

[2]:
https://drive.google.com/file/d/16O4LLH8dL0RWmbOx4HrcooFUyesWaVi-/view?usp=sharing

[3]:
https://github.com/odoo/odoo/blob/694f1d0fb03b56dd41a59eb676e56622634cc91b/addons/sale_pdf_quote_builder/utils.py#L11

sentry-6928220164
opw-5227601

PR #246034, opened 2026-01-28
[FIX] sale_pdf_quote_builder: prevent PyPDF warnings by using a real invalid pdf in tests

utma-odoo:
The `test_wrong_pdf` test fails on runbot in the `Nightly build` when executed under 
`Distro Builds > Testing distro PureTrixie without requirements for config Parallel testing`.

**Steps to reproduce:**
- Install `sale_pdf_quote_builder` module.
- Run `test_wrong_pdf` (Make sure `PyPDF==5.4.0` is installed).
- Observe `PyPDF` warnings in the logs.

**Warnings:**
```
WARNING	server	pypdf._reader:435 invalid pdf header: b'test' 
WARNING	server	pypdf._reader:435 EOF marker not found 
WARNING	server	pypdf._reader:435 EOF marker not found
```

**Root cause:**
Odoo officially supports the PyPDF version defined in requirements.txt [1]. 
The test passes with `PyPDF2`, but `PyPDF` is stricter and emits warnings when 
parsing arbitrary byte content. Because the test uses raw bytes instead of a 
structurally valid (but unsupported or encrypted) PDF, PyPDF logs warnings.

**Fix:**
This commit updates the test to use a real encrypted PDF instead of raw bytes, similar to [2].

This commit fixes test failures caused by [3].

**Note:**
Refer to [4] for generating the encrypted file.

[1]:
https://github.com/odoo/odoo/blob/a6af7df429f9cd3bfb60347e9b30cc48fb5bfaff/requirements.txt#L63-L65

[2]:
https://github.com/odoo/odoo/blob/0499dbd8944b4002883afab5d3ed76d8e55dab9c/addons/mail/tests/discuss/test_discuss_attachment_controller.py#L63-L75

[3]:
https://github.com/odoo/odoo/pull/230712

[4]:
https://colab.research.google.com/drive/1UfWSKDluObEkKxCg6eX-nHoe04_FGrEq#scrollTo=A5hO11Eh1qvV

runbot-238399


Julien00859:
@utma-odoo I would like to read in the commit message (at least) how the encrypted pdf file was generated. It could also be a (short, like 1 line) comment in the source code where you open it.

Julien00859 (approved):
@robodoo r+

PR #246035, opened 2026-01-28
[IMP] mail: clarify notification text and remove redundant notification

akap-odoo:
Purpose of this PR:

- Update notification text and type for "Copy text" and "Copy link"
actions so they reflect the accurate information.
- Remove redundant notification shown when an avatar is changed.

task-5873868

---

### DIFF
--- a/addons/mail/static/src/core/common/message_model.js
+++ b/addons/mail/static/src/core/common/message_model.js
@@ -581,12 +581,12 @@ export class Message extends Record {
     }
 
     async copyLink() {
-        let notification = _t("Message Link Copied!");
-        let type = "info";
+        let notification = _t("Message Link Copied");
+        let type = "success";
         try {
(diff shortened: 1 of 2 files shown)


addons/mail/static/src/core/common/message_model.js
alexkuhn: Would remove the `!` in the denied case too.

alexkuhn (approved):
LGTM

addons/mail/static/src/core/common/message_model.js
alexkuhn: And also the label of failed should be "Text Copy Failed" wording for consistency with change.

alexkuhn (approved):
@robodoo delegate+ r+

alexkuhn:
Rebase so hopefully green runbot

PR #246038, opened 2026-01-28
[FIX] stock_account: prevent traceback when user click produce all button in MO.

ddba-odoo:
Issue before this commit:
=========================
Clicking Produce All on a Manufacturing Order as a user with Manufacturing 
rights raised an Access Error.
The button tried to create account.move and account.move.line records, which the
user had no permission to create.

Steps to Reproduce:
=========================
1. Install account and mrp modules.
2. Create a product with:
   - Track Inventory enabled
   - Product uses real-time valuation
   - A cost of production exists on production/inventory location
3. Create a user with only Manufacturing user rights.
4. Log in as that user.
5. Create a Manufacturing Order, confirm it, and click Produce All.

Cause of the Issue:
=========================
Stock valuation during production attempted to create accounting entries for
specific locations or analytics, but users lacked access to
account.move and account.move.line, causing an access error.

With This Commit:
=========================
This commit allows users to produce Manufacturing Orders
without changing the stock valuation logic.

Steps to Reproduce: [Video Link](https://drive.google.com/file/d/1--k6m_UvoTyFpavPn1JUOyhsKydeMbVL/view?usp=sharing)

opw-5487383

### DIFF
--- a/addons/mrp_account/tests/test_mrp_account.py
+++ b/addons/mrp_account/tests/test_mrp_account.py
@@ -171,6 +171,11 @@ class TestMrpAccount(TestBomPriceCommon):
         overview_values = self.env['report.mrp.report_mo_overview'].get_report_values(mo.id)
         self.assertEqual(round(overview_values['data']['summary']['mo_cost'], 2), 677.08)
 
+    def test_mrp_user_without_account_permissions_can_create_bom(self):
+        mrp_user = new_test_user(self.env, 'temp_mrp_user', 'mrp.group_mrp_user')
+        mo_1 = self._create_mo(self.bom_1, 1)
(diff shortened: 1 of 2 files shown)


addons/mrp_account/tests/test_mrp_account.py
clesgow:
```suggestion
        mrp_user = new_test_user(self.env, 'temp_mrp_user', 'mrp.group_mrp_user')
```
Doesn't change anything test-wise, but it's confusing to call it a manager when we're especially testing a base mrp user.

ddba-odoo: Got it. I’ve updated the variable name to mrp_user to clearly reflect that we are testing a base user.
Thanks for the clarification.

PR #246039, opened 2026-01-28
[FIX] l10n_tw_edi_ecpay: Convert date time to TW time zone when

tau-odoo:
sending to ECPay

The date store in Odoo is in utc format, we need to convert it to tw time when sending the date to ECPay. The APIs are using the date to search for the invoices, if the date is not correct, it cannot find the invoices and return error.

task-5884616

Current behavior before PR:

---

### DIFF
--- a/addons/l10n_tw_edi_ecpay/utils.py
+++ b/addons/l10n_tw_edi_ecpay/utils.py
@@ -21,6 +21,22 @@ def transfer_time(time_before):
     return ecpay_time.astimezone(pytz.UTC).strftime("%Y-%m-%d %H:%M:%S")
 
 
+def convert_utc_time_to_tw_time(utc_datetime):
+    """
+        Converts UTC datetime object to a TW date string.
+
+        :param utc_datetime: datetime.datetime(2026, 1, 23, 18, 0, 0)
+        :return: "2026-01-24"
+    """
+    if utc_datetime.tzinfo is None:
+        utc_datetime = pytz.utc.localize(utc_datetime)
+
+    tw_tz = pytz.timezone('Asia/Taipei')
+    tw_time = utc_datetime.astimezone(tw_tz)
+
+    return tw_time.strftime("%Y-%m-%d")
+
+
 def encrypt(data, cipher):
     padder = padding.PKCS7(128).padder()
     padded_data = padder.update(data.encode("utf-8")) + padder.finalize()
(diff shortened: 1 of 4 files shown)


addons/l10n_tw_edi_ecpay/utils.py
```diff
@@ -21,6 +21,22 @@ def transfer_time(time_before):
     return ecpay_time.astimezone(pytz.UTC).strftime("%Y-%m-%d %H:%M:%S")
 
 
+def convert_to_tw_time(utc_datetime):
```
vin-odoo: You should be as clear as you can. By reading the method name, I have no idea what it converts from.

```suggestion
def convert_utc_time_to_tw_time(utc_datetime):
```

addons/l10n_tw_edi_ecpay/utils.py
vin-odoo, line 1: Wouldn't it be cleaner to use `fields.Date.context_today()` when setting l10n_tw_edi_invoice_create_date so that it stores the correct day?

Edit: I guess no, we can keep it this way to be 100% sure it would be tw time

vin-odoo (changes requested):
The title of the commit message is a bit too long.
I would instead write:

`[FIX] l10n_tw_edi_ecpay: Convert date to TW time zone
`

addons/l10n_tw_edi_ecpay/utils.py
```diff
@@ -21,6 +21,22 @@ def transfer_time(time_before):
...
+    """
+    if utc_datetime.tzinfo is None:
+        utc_datetime = pytz.utc.localize(utc_datetime)
+
+    tw_tz = pytz.timezone('Asia/Taipei')
+    tw_time = utc_datetime.astimezone(tw_tz)
+
+    return tw_time.strftime("%Y-%m-%d")
+
+
```
erle-odoo: Could consider this:

```py
from odoo.tools.date_utils import localized
import zoneinfo

return localized(dt).astimezone(zoneinfo.ZoneInfo("Asia/Taipei")).strftime("%Y-%m-%d")
```

tau-odoo: no `localized` function in version 18

vin-odoo (approved):
robodoo r+

Thank  you!

PR #246040, opened 2026-01-28
[FIX] mail: remove useless call to channel

phenix-factory:
`this` is already channel, the call to `.channel` isn't useful


tsm-odoo (approved):
@robodoo r+

PR #246054, opened 2026-01-28
[FIX] account_peppol: participant status fetch regression

antonrom1:
Fix regression of participant fetch cron introduced in forward port odoo/odoo#245038
Indeed self might be a record set in lots of different cases, which leads to users at least 1 existing edi proxy user in their company, all future registrations will fail


clbr-odoo (approved):
@robodoo r+

PR #246056, opened 2026-01-28
[FIX] html_editor: fix invalid document selection from safari in collab

Jinjiu96:
Before this commit: safari returns invalid document in collaboration, typically when a chrome user is sending history steps with undo. Reproduction steps:
1. In chrome, use an existing task with empty description or create a task in the project (first create the task title in the kanban view, then click edit), enter 4 lines of text
2. In one of the middle lines, delete one character --> undo --> add a new character
3. Save the task, open the task in Safari incognito, log in as demo (not admin), go to the task and click the description field
4. TraceBack: IndexSizeError: The index is not in the allowed range.

After this commit: we use the range of the DOM selection to set the
offsets of activeSelection. If the DOM selection is too wrong to be
corrected, e.g. the selection's anchor node isn't the same with range's
start container (or end container if direction is right to left), we do
not set new activeSelection but just return the previous activeSelection

task-5428788

---

### DIFF
--- a/addons/html_editor/static/src/core/selection_plugin.js
+++ b/addons/html_editor/static/src/core/selection_plugin.js
@@ -365,15 +365,30 @@ export class SelectionPlugin extends Plugin {
             if (anchorNode === focusNode && focusOffset < anchorOffset) {
                 direction = !direction;
             }
+
+            // Last resort: if the document selection doesn't even have the right
+            // anchorNode comparing to the range, we don't set the active selection.
+            const isSelectionUncorrectable = direction
+                ? anchorNode !== range.startContainer
+                : anchorNode !== range.endContainer;
+
             if (
                 this.activeSelection &&
-                (isProtecting(anchorNode) ||
+                (isSelectionUncorrectable ||
+                    isProtecting(anchorNode) ||
                     (isProtected(anchorNode) && !isUnprotecting(anchorNode)))
             ) {
                 // Keep the previous activeSelection in case of user interactions
                 // inside a protected zone.
                 return this.activeSelection;
             }
+
+            // For Safari, in edge cases in collaboration, the selection can be
+            // wrong (e.g. offsets are out of range) while the range is correct.
+            // We use range's offsets instead of selection's.
+            anchorOffset = direction ? range.startOffset : range.endOffset;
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/core/selection_plugin.js
```diff
@@ -497,7 +497,17 @@ export class SelectionPlugin extends Plugin {
...
+            // in collaboration when resetting from the peer's history and Safari
+            // can return a selection with invalid offsets.
+            const isDocumentSelectionValid =
+                selection.anchorNode.isConnected &&
+                nodeSize(selection.anchorNode) >= selection.anchorOffset &&
+                nodeSize(selection.focusNode) >= selection.focusOffset;
+
+            if (isDocumentSelectionValid) {
+                this.activeSelection = this.makeActiveSelection(selection);
+            }
```
dmo-odoo: I'm a bit confused as to the choice of fix here because this is not what I suggested to do in the task. This fix will only work if the previous value of `this.activeSelection` corresponds to the _current_ selection of the user. I would expect it to be the _previous_ selection instead, which would be incorrect. Am I wrong in my assumption?

It's hard to tell from the diff because the test would pass even if my assumption was true, since there is no previous selection in that case. I think to test against it, you'll need to move the selection in some place first, then move it in another place and have it returned wrong by the browser, then check the resulting editableSelection. If it has the first one then it's wrong because the selection of the user really has changed! The editor needs to read the correct current selection from the wrong data given by the browser. It's possible by reading the data from the range rather the selection, because the data of the range is correct, see technical specifications on the task.

Jinjiu96:
@dmo-odoo so I updated the commit, now the fix has two cases:
1. selection has correct anchornode/focusnode, but wrong offsets --> we set the offsets of active selection based on the range (direction based)
2. selection doesn't even have the correct anchornode/focusnode comparing to the range --> we don't change the activeSelection and return the existing one.

For the tests, they are testing the correct things. We didn't need to move the selection to somewhere else, cause when we have a document selection and call `getEditableSelection`, it triggers `makeActiveSelection` to align our editable selection with the document selection (which is different in the patched `getSelection`). So the function is properly tested 😇

PR #232973, opened 2025-10-24
[FIX] repair: allow serial number only after save

LouisGobert:
When creating a repair order, if you add a line and assign a serial
number before saving the line, the serial number is removed when saving.

Steps to reproduce:
- Create a repair order and save it.
- Add a repair line with a product tracked by serial number.
- Without saving the repair order, add a serial number on the line.
- Save the repair order.
- You will notice that the serial number is no longer assigned to the
repair line.

This PR fixes the issue by making the serial number field read-only
until the repair line is saved.

opw-5156267

PR #61287, opened 2020-11-03
[IMP] mrp,stock: improve duplicate SN warnings

ticodoo:
Add and improve onchange warnings when a duplicate SN is used in
following cases: inventory, picking (any type), manufacturing, and
scrap.

Improvement includes:
- include location where the existing SN is
- auto-correct source location when appropriate (e.g. trying to scrap a
  SN in the wrong location)

The goal of this is to prevent but not restrict duplicate SNs so users
can have some flexibility (especially if a dupe SN occurs because of an
error such as doing pick-pack-ship out of order.)

Task: 1924758

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246060, opened 2026-01-28
[IMP] stock, *: improve lot selection and reservation behavior

srap-odoo:
_* = mrp, repair, product_expiry

This improvement makes lot reservation and modification easier directly from the
picking view, removing the need to open Detailed Operations or check quants.
Users can now see available quantities while selecting lots and reserve them
in fewer steps, reducing manual effort and improving accuracy by avoiding
negative lot reservations.

* Rename the `Serial Numbers` column to `Tracking` to reflect support for both
  lots and serial numbers in stock moves.
* When a `lot is selected`:
  - The quantity is automatically set to the minimum of the remaining demand
    or the available stock (with lot or no lots defined); otherwise, it
    defaults to 1 if there is no availability or the demand is already fulfilled.
  - If a move line without a lot exists, it will be assigned to the newly
    selected lot, maintaining consistency with MRP lot assignment behavior.
* When a `lot is removed`, its corresponding move lines are deleted and the
  reserved quantity is recomputed accordingly, instead of setting all move lines
  to quantity 1 (as done for serial numbers).
* Display the `available quantity` of each lot in the move’s source location
  (internal usage) directly in the dropdown to help quick selection without
  checking quants.
* Clicking a lot tag opens the lot form view for quick access to details like
  on-hand quantity or expiration date.
* Enable editing of the `Tracking` column in Manufacturing and Repair pickings
  for consistent lot handling.
* Hide the Tracking column for Repair orders in draft state as it is not
  actionable before confirmation.
* Removed the warning for duplicate serial numbers when selecting a lot for a
  lot-tracked product with a different source location, because unlike serial
  numbers, a single lot can exist in multiple locations.
* Filtered the lot dropdown to show only available lots in the source location for
  internal operations, such as transfers and scrap flows, to prevent selection errors
  and minimize negative inventory reservations.

Enterprise PR: https://github.com/odoo/enterprise/pull/117240
task-5375336

### DIFF
--- a/addons/mrp/views/mrp_production_views.xml
+++ b/addons/mrp/views/mrp_production_views.xml
@@ -512,7 +512,6 @@
                                         options="{'on_tag_click': 'open_form'}"
                                         create="parent.use_create_components_lots"
                                         context="{'default_product_id': product_id, 'active_move_id': id}"
-                                        domain="[('product_id','=',product_id)]"
                                     />
                                     <button name="action_show_details" type="object" title="Show Details" string="Details" context="{'default_product_uom_qty': 0}" invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"/>
                                     <button class="o_optional_button btn" name="action_show_details" type="object" title="Show Details" string="Details" context="{'default_product_uom_qty': 0}" invisible="(has_tracking in ['lot', 'serial']) or not show_details_visible"/>
@@ -569,7 +568,7 @@
                                         options="{'on_tag_click': 'open_form'}"
                                         create="parent.use_create_components_lots"
                                         context="{'default_product_id': product_id}"
-                                        domain="[('product_id','=',product_id)]"/>
+                                    />
                                     <button name="action_show_details" type="object" title="Show Details" icon="fa-list" invisible="has_tracking not in ['lot', 'serial'] or not show_details_visible"/>
                                     <button class="o_optional_button btn btn-light" name="action_show_details" type="object" title="Show Details" icon="fa-list" invisible="has_tracking in ['lot', 'serial'] or not show_details_visible"/>
                                 </list>
--- a/addons/repair/views/repair_views.xml
+++ b/addons/repair/views/repair_views.xml
@@ -181,7 +181,6 @@
                                     invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
                                     optional="hide"
                                     context="{'default_product_id': product_id, 'active_move_id': id}"
-                                    domain="[('product_id','=',product_id)]"
                                 />
                                 <button name="action_show_details" type="object" string="Details"
                                         invisible="not show_details_visible" options='{"warn": true}'
--- a/addons/stock/models/stock_move.py
+++ b/addons/stock/models/stock_move.py
@@ -21,6 +21,19 @@ class StockMove(models.Model):
     _order = 'sequence, id'
     _rec_name = 'reference'
 
+    def _get_lot_domain(self):
+        return """
+            [('product_id', '=', product_id)] + (
+                [('id', 'in', context.get('lot_ids'))]
+                if context.get('lot_ids')
+                else (
+                    [('product_qty', '>', 0), ('quant_ids', 'any', [('location_id', 'child_of', location_id)])]
+                    if location_usage == 'internal'
+                    else []
+                )
+            )
+        """
+
     sequence = fields.Integer('Sequence', default=10)
     priority = fields.Selection(
         PROCUREMENT_PRIORITIES, 'Priority', default='0',
@@ -191,7 +204,7 @@ class StockMove(models.Model):
     orderpoint_id = fields.Many2one('stock.warehouse.orderpoint', 'Original Reordering Rule', index=True)
     forecast_availability = fields.Float('Forecast Availability', compute='_compute_forecast_information', digits='Product Unit', compute_sudo=True)
     forecast_expected_date = fields.Datetime('Forecasted Expected date', compute='_compute_forecast_information', compute_sudo=True)
-    lot_ids = fields.Many2many('stock.lot', compute='_compute_lot_ids', inverse='_set_lot_ids', domain="[('product_id', '=', product_id)]", string='Tracking')
+    lot_ids = fields.Many2many('stock.lot', compute='_compute_lot_ids', inverse='_set_lot_ids', domain=_get_lot_domain, string='Tracking')
     reservation_date = fields.Date('Date to Reserve', compute='_compute_reservation_date', store=True, help="Computes when a move should be reserved")
     packaging_uom_id = fields.Many2one('uom.uom', 'Packaging', help="Packaging unit from sale or purchase orders", compute='_compute_packaging_uom_id', precompute=True, store=True)
     packaging_uom_qty = fields.Float('Packaging Quantity', help="Quantity in the packaging unit", compute='_compute_packaging_uom_qty', store=True)
@@ -1440,6 +1453,7 @@ Please change the quantity done or the rounding precision in your settings.""",
         self.quantity = self.product_id.uom_id._compute_quantity(quantity, self.uom_id, round=False)
         base_location = self.location_id
         quants = self.env['stock.quant'].sudo().search([
+            ('tracking', '=', 'serial'),
             ('product_id', '=', self.product_id.id),
             ('lot_id', 'in', self.lot_ids.ids),
             ('quantity', '!=', 0),
--- a/addons/stock/views/stock_move_views.xml
+++ b/addons/stock/views/stock_move_views.xml
@@ -345,7 +345,7 @@
                                     <field name="quantity" string="Quantity" readonly="state in ('done', 'cancel')"/>
                                     <field name="uom_id" widget="many2one_uom" options="{'no_open': True, 'no_create': True}" groups="uom.group_uom" readonly="state in ('done', 'cancel')"/>
                                 </div>
-                                <field name="lot_ids" widget="many2many_tags" string="Lot/Serial Numbers" invisible="not product_id or has_tracking not in ['lot', 'serial']" readonly="state in ('done', 'cancel')" domain="[('product_id', '=', product_id), ('id', 'in', context.get('lot_ids'))] if context.get('lot_ids') else [('product_id', '=', product_id)]"/>
+                                <field name="lot_ids" widget="many2many_tags" string="Lot/Serial Numbers" invisible="not product_id or has_tracking not in ['lot', 'serial']" readonly="state in ('done', 'cancel')"/>
                                 <field name="scrap_reason_tag_ids" widget="many2many_tags" options="{'color_field': 'color'}" readonly="state == 'cancel'"/>
                             </group>
                             <group>
--- a/addons/stock/views/stock_picking_views.xml
+++ b/addons/stock/views/stock_picking_views.xml
@@ -318,7 +318,6 @@
                                         optional="hide"
                                         create="parent.use_create_lots == True"
                                         context="{'default_company_id': company_id, 'default_product_id': product_id, 'active_picking_id': parent.id, 'active_move_id': id}"
-                                        domain="[('product_id','=',product_id)]"
                                     />
                                     <button name="action_show_details" type="object" string="Details" title="Details"
                                             invisible="not show_details_visible"/>


addons/stock/tests/test_move.py
```diff
@@ -6742,3 +6742,26 @@ def test_newly_added_move_line_is_picked_if_move_is_picked(self):
         # All the move lines are also picked
         for move_line in delivery.move_ids.move_line_ids:
             self.assertTrue(move_line.picked)
+
+    def test_modifying_lots_updates_move_lines_and_quantity(self):
+        """ Ensure lot assignment changes correctly update move lines and quantity."""
```
dhha-odoo:
```suggestion
    def test_modifying_tracking_updates_move_lines_and_quantity(self):
        """Ensure that modifying the tracking (lot/serial) field on a move correctly
        updates its move lines and reserved quantities.
        """
```

[docstring](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)

addons/mrp/views/mrp_production_views.xml
```diff
@@ -496,14 +496,13 @@
...
-                                        string="Lot/Serial Numbers"
                                         help="Displays the consumed Lot/Serial Numbers."
                                         groups="stock.group_production_lot"
-                                        readonly="1"
+                                        options="{'on_tag_click': 'open_form'}"
                                         column_invisible="not parent.show_lot_ids or parent.state == 'draft'"
                                         invisible="not show_details_visible"
                                         create="parent.use_create_components_lots"
-                                        context="{'default_product_id': product_id}"
+                                        context="{'default_product_id': product_id, 'active_move_id': id}"
```
dhha-odoo, line 514: With this change, the **lot_ids field becomes editable even for non-tracked products**, which is incorrect and inconsistent with the behaviour in pickings.

For non-tracked products, allowing users to edit tracking information has no functional value and may confuse users. The same applies to by-products.

Please ensure the same functionality is present across all views (MRP, Stock, Repair) for consistency.

addons/repair/views/repair_views.xml
```diff
@@ -174,12 +174,12 @@
...
                                 <field name="lot_ids" widget="many2many_tags"
+                                    options="{'on_tag_click': 'open_form'}"
                                     groups="stock.group_production_lot"
-                                    invisible="not show_details_visible or has_tracking != 'serial'"
                                     optional="hide"
-                                    context="{'default_product_id': product_id}"
+                                    context="{'default_product_id': product_id, 'active_move_id': id}"
                                     domain="[('product_id','=',product_id)]"
-                                    readonly="not id"/>
+                                    readonly="not id or not show_details_visible or has_tracking == 'none'"/>
```
dhha-odoo: 1. Create a **storable product with lot tracking** and generate a lot for it.
2. Ensure **Multi-Step Routes, Storage Locations, and Consignment are disabled**.
3. Create a repair order and add a move using the same lot.
4. Result: The **user is not able to edit the lot_ids**, Why?.

User should be able to edit it, because in the same condition, the user can also change it in normal pickings.

addons/product_expiry/models/production_lot.py
```diff
@@ -24,19 +24,15 @@ class StockLot(models.Model):
...
-            super(StockLot, self.env['stock.lot'].browse(lots_to_process_ids))._compute_display_name()
+        super()._compute_display_name()
+        if not self.env.context.get('formatted_display_name'):
+            return
+        now = fields.Datetime.now()
+        for lot in self.filtered(lambda lot: lot.use_expiration_date and lot.expiration_date):
+            if now >= lot.expiration_date:
+                lot.display_name += self.env._("\t--Expired--")
+            elif lot.alert_date and now >= lot.alert_date:
+                lot.display_name += self.env._("\t--Expire on %(date)s--", date=fields.Datetime.to_string(lot.expiration_date))
```
dhha-odoo, line 35: Why is this change needed? It does not seem related to the specs.

Also, the **previous approach was better**. With this change, the **super() method is called for every record**. Previously, it was called in **batch only for records where use_expiration_date was not set**, which is more efficient.

So please revert this change.

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,25 @@ def _read_group_location_id(self, locations, domain):
...
+            available_qty_by_lot = self.env['stock.quant']._read_group(
+                domain=[
+                    ('product_id', '=', move_id.product_id.id),
+                    ('location_id', '=', move_id.location_id.id),
+                    ('lot_id', 'in', not_selected_lots.ids),
+                ], groupby=['lot_id'], aggregates=['available_quantity:sum'],
+            )
+            for lot, available_qty in available_qty_by_lot:
+                available_qty = move_id.product_id.uom_id._compute_quantity(available_qty, move_uom)
+                lot.display_name = f"{lot.name}\t--{available_qty or 0.0} {move_uom.name}--"
```
dhha-odoo:
```suggestion
                lot.display_name = f"{lot.name}\t--{available_qty} {move_uom.name}--"
```

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,25 @@ def _read_group_location_id(self, locations, domain):
         inverse='_set_single_location', domain="[('usage', '!=', 'view')]", group_expand='_read_group_location_id')
     is_scrap = fields.Boolean('Is Scrapped', compute='_compute_is_scrap')
 
+    @api.depends_context('active_move_id', 'formatted_display_name')
+    def _compute_display_name(self):
+        super()._compute_display_name()
+        ctx = self.env.context
+        move_id = self.env['stock.move'].browse(ctx.get('active_move_id'))
```
dhha-odoo: Avoid **calling browse() repeatedly**; it is currently executed even when **ctx.get('formatted_display_name') is not set**.

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,25 @@ def _read_group_location_id(self, locations, domain):
         inverse='_set_single_location', domain="[('usage', '!=', 'view')]", group_expand='_read_group_location_id')
     is_scrap = fields.Boolean('Is Scrapped', compute='_compute_is_scrap')
 
+    @api.depends_context('active_move_id', 'formatted_display_name')
+    def _compute_display_name(self):
+        super()._compute_display_name()
```
dhha-odoo, line 68: No need to call `super() for every record`. Call it only for records where the `condition is not satisfied for that record`.  [Like](https://github.com/odoo/odoo/pull/246060/changes#diff-9b0f4f883694c9b8a64fd6bc3d7c92b394626839543358be671440253291133bL27-L35)

addons/stock/models/stock_move.py
```diff
@@ -1360,9 +1369,42 @@ def _get_relevant_state_among_moves(self):
...
+        quantity = sum((mls_with_existing_lots | mls_without_lots).mapped('quantity_product_uom'))
+        newly_added_lots = self.lot_ids.filtered(lambda lot: lot._origin.id not in self._origin.lot_ids.ids)
+        if self.product_id.tracking == 'serial':
+            quantity = len(self.lot_ids)
+        elif newly_added_lots:
+            quants_by_lot = self.env['stock.quant']._read_group(
+                domain=[
+                    ('product_id', '=', self.product_id.id),
+                    ('location_id', '=', self.location_id.id),
+                    '|', ('lot_id', 'in', newly_added_lots._origin.mapped('id')), ('lot_id', '=', False),
```
dhha-odoo:
```suggestion
                    '|', ('lot_id', 'in', newly_added_lots.ids), ('lot_id', '=', False),
```

addons/stock/models/stock_move.py
```diff
@@ -1360,9 +1369,42 @@ def _get_relevant_state_among_moves(self):
 
     @api.onchange('lot_ids')
     def _onchange_lot_ids(self):
-        quantity = sum(ml.quantity_product_uom for ml in self.move_line_ids.filtered(lambda ml: not ml.lot_id and ml.lot_name))
-        quantity += self.product_id.uom_id._compute_quantity(len(self.lot_ids), self.uom_id)
-        self.update({'quantity': quantity})
+        mls_with_existing_lots = self.move_line_ids.filtered(lambda ml: ml.lot_id and ml.lot_id.id in self.lot_ids.ids)
+        mls_without_lots = self.move_line_ids.filtered(lambda ml: not ml.lot_id and not ml.lot_name)
+        quantity = sum((mls_with_existing_lots | mls_without_lots).mapped('quantity_product_uom'))
+        newly_added_lots = self.lot_ids.filtered(lambda lot: lot._origin.id not in self._origin.lot_ids.ids)
```
dhha-odoo: For **serial-tracked products**, the quantity is directly derived from the **length of self.lot_ids**.

The prior **computation of mls_with_existing_lots, mls_without_lots, newly_added_lots, and quantity is unnecessary** in this case. Then what is the purpose of computing these values when they are not used in the serial tracking case? 

Can they be skipped by handling the serial logic first? This would improve **readability and avoid redundant computation**.

addons/stock/tests/test_move.py
dhha-odoo:
```suggestion
    def test_modifying_tracking_updates_move_lines_and_quantity(self):
        """Ensure that modifying the tracking (lot/serial) field on a move correctly
        updates its move lines and reserved quantities.
        """
```

[docstring](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)

addons/stock/tests/test_move.py
```diff
@@ -6742,3 +6742,26 @@ def test_newly_added_move_line_is_picked_if_move_is_picked(self):
...
+        delivery = self.env['stock.picking'].create({'picking_type_id': self.picking_type_out.id})
+        move = self._create_move(product, stock_location, self.customer_location, product_uom_qty=20, picking_id=delivery.id)
+        delivery.action_confirm()
+        with Form(delivery) as delivery_form:
+            with delivery_form.move_ids.edit(0) as move_form:
+                move_form.lot_ids = lot3
+            delivery = delivery_form.save()
+        self.assertRecordValues(move.move_line_ids, [{'lot_id': lot3.id, 'quantity': 20}])
+        delivery.button_validate()
+        self.assertRecordValues(delivery.move_ids, [{'state': 'done', 'lot_ids': [lot3.id], 'quantity': 20}])
```
dhha-odoo:
```suggestion
        self.assertRecordValues(delivery.move_ids, [{'lot_ids': lot3.ids, 'quantity': 20}])
```

addons/stock/views/stock_move_views.xml
```diff
@@ -340,7 +340,11 @@
...
                                     <field name="uom_id" widget="many2one_uom" options="{'no_open': True, 'no_create': True}" groups="uom.group_uom" readonly="state in ('done', 'cancel')"/>
                                 </div>
-                                <field name="lot_ids" widget="many2many_tags" string="Lot/Serial Numbers" invisible="not product_id or has_tracking not in ['lot', 'serial']" readonly="state in ('done', 'cancel')" domain="[('product_id', '=', product_id), ('id', 'in', context.get('lot_ids'))] if context.get('lot_ids') else [('product_id', '=', product_id)]"/>
+                                <field name="lot_ids" widget="many2many_tags" string="Lot/Serial Numbers"
+                                    invisible="not product_id or has_tracking not in ['lot', 'serial']"
+                                    readonly="state in ('done', 'cancel')"
+                                    options="{'on_tag_click': 'open_form'}"
+                                    domain="[('product_id', '=', product_id), ('id', 'in', context.get('lot_ids'))] if context.get('lot_ids') else [('product_id', '=', product_id)]"/>
```
dhha-odoo: I don’t fully understand this change.

> Clicking a lot tag now opens the lot form view for quick access to details
such as on-hand quantity or expiration date in all related views where lot
details matter.


The commit message mentions opening the lot form view from lot tags for quick access to details. If this behaviour is important, shouldn’t it also be applied consistently across other relevant views (e.g., `quality_check_view_tree, quality_alert_view_tree`, etc.)?

This change feels a bit out of scope for the current task. While it's important, it's better handled in a separate task/PR. For now, I would suggest reverting these changes.

Also, could you confirm whether this was discussed or validated with the PO?

addons/stock/models/stock_move.py
dhha-odoo, line 1: 1: Create a lot-tracked product.
2: Create two warehouses (e.g., WH and WH1).
3: Add stock with lots in both warehouses:
- WH/Stock → LOT1 → Qty: 5
- WH1/Stock → LOT2 → Qty: 10
4: Create a delivery picking with source location: WH/Stock.
5: Check the tracking (lot) field on the picking.

Result: The tracking field **shows both LOT1 and LOT2**.

Expected: Only **LOT1 should be visible**, since **LOT2 is not available in the selected source location (WH/Stock)**.

Also, quantity is not shown next to the LOT2, which indicates incorrect behaviour.

**Additional Note:**
-------------------------------------------
This behaviour also exists in the master, so it is not related to your task or PR.

You may want to confirm with the PO whether this should be handled in this task or addressed in a separate task.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,48 @@ def _compute_lot_ids(self):
 
     def _set_lot_ids(self):
         for move in self:
-            if move.state == 'assigned' and all(ml.lot_id in move.lot_ids for ml in move.move_line_ids):
-                continue
+            # NOTE: this early return is removed because even after the demand is fulfilled we should allow adding the
+            # lots and assigning the ml a proper lot/serial to it. Otherwise when selecting the lot after demand is
```
dhha-odoo:
```suggestion
            # lots and assigning the ML a proper lot/serial to it. Otherwise, when selecting/creating the lot after demand is
```

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,48 @@ def _compute_lot_ids(self):
 
     def _set_lot_ids(self):
         for move in self:
-            if move.state == 'assigned' and all(ml.lot_id in move.lot_ids for ml in move.move_line_ids):
```
dhha-odoo, line 637: You can **modify the condition instead of removing** it, based on your requirement.

Currently, it executes even for moves in the **draft state without any purpose**, whereas before this change, those cases were not processed.

So, you should at least adapt the condition to skip draft moves, for example:

```
if move.state == 'draft': 
    continue
```

This will help avoid unnecessary computation(for draft moves).

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,48 @@ def _compute_lot_ids(self):
...
                         else:
                             move_line_vals = self._prepare_move_line_vals(quantity=0)
-                            move_line_vals['lot_id'] = lot.id
+                            move_line_vals['quantity'] = 1.0
+                        move_line_vals['lot_id'] = lot.id
                         move_line_vals['uom_id'] = move.product_id.uom_id.id
-                        move_line_vals['quantity'] = 1
                         move_lines_commands.append((0, 0, move_line_vals))
                 else:
                     move_line = move.move_line_ids.filtered(lambda line: line.lot_id.id == lot.id)
```
dhha-odoo: You’re not using the move_line variable anywhere after else, so what’s the purpose of keeping it?

addons/stock/models/stock_move.py
```diff
@@ -1360,9 +1369,42 @@ def _get_relevant_state_among_moves(self):
 
     @api.onchange('lot_ids')
     def _onchange_lot_ids(self):
-        quantity = sum(ml.quantity_product_uom for ml in self.move_line_ids.filtered(lambda ml: not ml.lot_id and ml.lot_name))
-        quantity += self.product_id.uom_id._compute_quantity(len(self.lot_ids), self.uom_id)
-        self.update({'quantity': quantity})
+        mls_with_existing_lots = self.move_line_ids.filtered(lambda ml: ml.lot_id and ml.lot_id.id in self.lot_ids.ids)
+        mls_without_lots = self.move_line_ids.filtered(lambda ml: not ml.lot_id and not ml.lot_name)
```
dhha-odoo:
```suggestion
        mls_without_lots = mls_with_existing_lots - self.move_line_ids._origin
```
🤔

dhha-odoo (changes requested):
Hello @srap-odoo 

Thanks for your work.

Here is my first pass; it still needs a deeper review of the core logic.
Meanwhile, I’ve added some comments; could you please have a look?

Thank you.

addons/repair/views/repair_views.xml
srap-odoo: This behaviour is due to the `not show_details_visible` condition, `lot_ids` are not editable while the repair is in draft or not yet confirmed. They become editable once the repair is confirmed.

addons/product_expiry/models/production_lot.py
srap-odoo: By this, I am trying to achieve the below functionality while still keeping the lot expiration logic intact.

> Display the `available quantity` of each lot in the move’s source location
> (when the usage is internal) directly in the dropdown, to help users quickly
> select the appropriate lots without having to check quants.

Regarding performance, you are right that the previous compute used batch processing for lots without expiration dates. However, while doing so, it ends up overwriting the available quantity returned by the `super()` call (from the Stock module).

I have also confirmed with the PO that the available quantity should be displayed first, followed by the expired tag or alert date.
image

addons/stock/models/stock_lot.py
srap-odoo: I initially introduced this to avoid displaying a `False` quantity when a lot has no available quantity [see](https://www.loom.com/share/ae95662bcefd45eb9da67d2e68af39ac)

After optimization of compute, I’ve updated the implementation to handle it as shown [here](https://github.com/odoo-dev/odoo/blob/287c75bc49a3acd179dffaa0e5da6ad6359ec9a1/addons/stock/models/stock_lot.py#L87)

addons/stock/models/stock_lot.py
srap-odoo: Hello @dhha-odoo,

I did try to use the batched `super()` call as you suggested, but here is the problem I am facing.

While doing it [like](https://codeshare.io/axjlkR), when there is no `formatted_display_name` and `self` (functionally, just selecting a random lot from the dropdown will trigger this condition) contains records of type `NewId`, the `browse` fails to fetch those records, as they are not present in the cache, and it raises the error:
`ValueError: Compute method failed to assign stock.lot(<NewId origin=141>, <NewId origin=139>).display_name`.

However, to ensure we avoid calling `super()` on every record:
[1] I assign `lot.name` when there is a `move_id` and `formatted_display_name`, but no available quantity for that lot. [See](https://github.com/odoo-dev/odoo/blob/287c75bc49a3acd179dffaa0e5da6ad6359ec9a1/addons/stock/models/stock_lot.py#L90-L91)
[2] When the [condition](https://github.com/odoo-dev/odoo/blob/287c75bc49a3acd179dffaa0e5da6ad6359ec9a1/addons/stock/models/stock_lot.py#L69-L73) is false, I bypass it using the `else` block, and then calling `super()` already handles the batched processing of lots, as self will contain all the lots.

Let me know if you see it differently :relaxed:

addons/stock/views/stock_move_views.xml
srap-odoo: Hello @dhha-odoo,

I have reverted these changes as they might create extra noise in the scope of this task.

The intention was to ensure consistency across related views. However, views like `quality_check_view_tree` and `quality_alert_view_tree` were excluded, as adding the `open_form` option there would have no effect, because those views already open the corresponding quality records on click.

Thank you for the suggestion!

addons/stock/models/stock_move.py
srap-odoo: Hello @dhha-odoo,
Thank you for pointing that out. I also noticed that behavior, and here is why we should keep it as is.

This is because we don’t block users from selecting lots from a different warehouse or even a different company.
I discussed adding a source location domain with the PO, and he suggested that we should not block the user. However, we already warn users when they choose lots from a different location.

See: [https://github.com/odoo/odoo/pull/61287](https://github.com/odoo/odoo/pull/61287)

addons/stock/models/stock_move.py
srap-odoo: Thank you for the suggestion. I have modified the condition as follows:

```python
if move.state == 'draft' and not move.lot_ids:
    continue
```

In cases like a newly created scrap order, the moves are in a draft state with lots already assigned to them. This ensures that we don’t clear `move.lot_ids` by skipping such moves.

addons/stock/models/stock_move.py
srap-odoo: By doing this, it won't allow me to assign the empty MLs to the newly selected/created lot.
So `mls_without_lots` is considering the MLs with quants that have no specific lot assigned to them.
For example,
image
1. With the above quants in the warehouse, create an SO with demand = 30 and confirm it.
The delivery will have 3 MLs.
image
2. Now, when the delivery operation type is set to create new lots, create a lot on the fly called lot3 and check what happens. The lot3 should have been assigned to the empty quant ML.

addons/mrp/views/mrp_production_views.xml
```diff
@@ -552,13 +551,12 @@
...
                                         optional="hide"
-                                        string="Lot/Serial Numbers"
                                         help="Displays the produced Lot/Serial Numbers."
                                         groups="stock.group_production_lot"
-                                        readonly="1"
                                         column_invisible="not parent.show_lot_ids or parent.state == 'draft'"
-                                        invisible="not show_details_visible"
+                                        invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
                                         create="parent.use_create_components_lots"
+                                        options="{'on_tag_click': 'open_form'}"
```
dhha-odoo, line 568: The field attribute order was already consistent before your changes, as it was handled in the same PR. 
Now that you’ve added options for both component and by-product, it would be good to maintain consistency here as well, either place `options` before `invisible` (as done for component) or follow the same order for both.

This is not a strict guideline, just a suggestion. Currently, it feels a bit inconsistent since in one place `options` is added before `invisible`, while in another it’s placed before `context`, even though both fields are the same.

addons/repair/views/repair_views.xml
```diff
@@ -174,10 +174,11 @@
...
                                 <field name="lot_ids" widget="many2many_tags"
+                                    options="{'on_tag_click': 'open_form'}"
                                     groups="stock.group_production_lot"
-                                    invisible="not show_details_visible or has_tracking != 'serial'"
+                                    invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
                                     optional="hide"
-                                    context="{'default_product_id': product_id}"
+                                    context="{'default_product_id': product_id, 'active_move_id': id}"
                                     domain="[('product_id','=',product_id)]"
                                     readonly="not id"/>
```
dhha-odoo: IMO it's redundant and can be removed(It's a small change). Could you please check this small [PR](https://github.com/odoo/odoo/pull/232973)?

addons/stock/tests/test_move.py
```diff
@@ -6742,3 +6742,28 @@ def test_newly_added_move_line_is_picked_if_move_is_picked(self):
...
+        updates its move lines and reserved quantities.
+        """
+        product = self.product_lot
+        stock_location = self.stock_location
+        Quant = self.env['stock.quant']
+        lot1, lot2, lot3 = self.env['stock.lot'].create([{
+            'name': 'lot%s' % str(i), 'product_id': product.id,
+        } for i in range(1, 4)])
+        Quant._update_available_quantity(product, stock_location, 10, lot_id=lot1)
+        Quant._update_available_quantity(product, stock_location, 10, lot_id=lot2)
```
dhha-odoo: 2 lots are sufficient

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,35 @@ def _read_group_location_id(self, locations, domain):
         inverse='_set_single_location', domain="[('usage', '!=', 'view')]", group_expand='_read_group_location_id')
     is_scrap = fields.Boolean('Is Scrapped', compute='_compute_is_scrap')
 
+    @api.depends_context('active_move_id', 'formatted_display_name')
+    def _compute_display_name(self):
+        ctx = self.env.context
+        if (
+            ctx.get('formatted_display_name')
+            and (move_id := self.env['stock.move'].browse(ctx.get('active_move_id')))
+            and move_id.product_id.tracking == 'lot' and move_id.location_id.usage == 'internal'
```
dhha-odoo:
```suggestion
            and move_id.has_tracking == 'lot' and move_id.location_usage == 'internal'
```

addons/stock/models/stock_move.py
```diff
@@ -1369,9 +1383,43 @@ def _get_relevant_state_among_moves(self):
 
     @api.onchange('lot_ids')
     def _onchange_lot_ids(self):
-        quantity = sum(ml.quantity_product_uom for ml in self.move_line_ids.filtered(lambda ml: not ml.lot_id and ml.lot_name))
-        quantity += self.product_id.uom_id._compute_quantity(len(self.lot_ids), self.uom_id)
-        self.update({'quantity': quantity})
+        if self.product_id.tracking == 'serial':
```
dhha-odoo:
```suggestion
        if self.has_tracking == 'serial':
```

addons/stock/models/stock_move.py
```diff
@@ -1369,9 +1383,43 @@ def _get_relevant_state_among_moves(self):
...
+            quantity = len(self.lot_ids)
+        else:
+            mls_with_existing_lots = self.move_line_ids.filtered(lambda ml: ml.lot_id and ml.lot_id.id in self.lot_ids.ids)
+            mls_without_lots = self.move_line_ids.filtered(lambda ml: not ml.lot_id and not ml.lot_name)
```
dhha-odoo:
```suggestion
            mls_with_existing_lots = self.move_line_ids.filtered(lambda ml: ml.lot_id.id in self.lot_ids.ids)
            mls_without_lots = self.move_line_ids - mls_with_existing_lots
```

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,35 @@ def _read_group_location_id(self, locations, domain):
...
+        ):
+            not_selected_lots = self - move_id.lot_ids
+            move_uom_name = move_id.uom_id.name
+            available_qty_by_lot = {
+                lot.id: available_qty
+                for lot, available_qty in self.env['stock.quant']._read_group(
+                    domain=[
+                        ('product_id', '=', move_id.product_id.id),
+                        ('location_id', '=', move_id.location_id.id),
```
dhha-odoo:
```suggestion
                        ('location_id', 'child_of', move_id.location_id.id),
```

Otherwise, it will avoid the child location and not count a quantity for a lot that is available at the child location

Image

dhha-odoo (changes requested):

addons/mrp/views/mrp_production_views.xml
```diff
@@ -554,12 +553,11 @@
                                     <field name="show_details_visible" column_invisible="True"/>
                                     <field name="lot_ids" widget="many2many_tags"
                                         optional="hide"
-                                        string="Lot/Serial Numbers"
                                         help="Displays the produced Lot/Serial Numbers."
                                         groups="stock.group_production_lot"
-                                        readonly="1"
```
dhha-odoo, line 568: Now that this field is editable, I noticed a functional issue.

When trying to assign a **lot to a by-product**, a warning is triggered (Unavailable Serial numbers, even though it's a lot number). However, since **by-products are part of the finished products being moved into stock**, this behaviour seems **inconsistent**. For lot tracking, using the same lot should be acceptable (this warning would make sense for serial tracking, but not for lots).

While the same issue already exists in the master for the receipt. I confirmed with the PO that it can be fixed as part of this task, but you may still want to double-check with them.

Image




For the solution, you may need to add a **('tracking', '=', 'serial')** domain in [quant selection here](https://github.com/odoo/odoo/blob/master/addons/stock/models/stock_move.py#L1376).

addons/repair/views/repair_views.xml
dhha-odoo: Hmm, makes sense, thanks for the clarification.

But, we could still add **column_invisible="parent.state == 'draft'"** to maintain consistency between Picking and MRP. In the draft stage, the user cannot edit this field anyway, so there’s no value in displaying it. We can hide it until the state changes from draft.

addons/repair/views/repair_views.xml
```diff
@@ -175,12 +175,12 @@
...
+                                    options="{'on_tag_click': 'open_form'}"
                                     groups="stock.group_production_lot"
-                                    invisible="not show_details_visible or has_tracking != 'serial'"
+                                    invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
                                     optional="hide"
-                                    context="{'default_product_id': product_id}"
-                                    domain="[('product_id','=',product_id)]"
-                                    readonly="not id"/>
+                                    context="{'default_product_id': product_id, 'active_move_id': id}"
+                                    domain="[('product_id','=',product_id)]"/>
```
dhha-odoo:
```suggestion
                                    domain="[('product_id', '=', product_id)]"/>
```

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,35 @@ def _read_group_location_id(self, locations, domain):
         inverse='_set_single_location', domain="[('usage', '!=', 'view')]", group_expand='_read_group_location_id')
     is_scrap = fields.Boolean('Is Scrapped', compute='_compute_is_scrap')
 
+    @api.depends_context('active_move_id', 'formatted_display_name')
+    def _compute_display_name(self):
+        ctx = self.env.context
+        if (
+            ctx.get('formatted_display_name')
+            and (move_id := self.env['stock.move'].browse(ctx.get('active_move_id')))
```
dhha-odoo:
```suggestion
        if (
            self.env.context.get('formatted_display_name')
            and (move_id := self.env['stock.move'].browse(self.env.context.get('active_move_id')))
```

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,35 @@ def _read_group_location_id(self, locations, domain):
...
+            available_qty_by_lot = {
+                lot.id: available_qty
+                for lot, available_qty in self.env['stock.quant']._read_group(
+                    domain=[
+                        ('product_id', '=', move_id.product_id.id),
+                        ('location_id', 'child_of', move_id.location_id.id),
+                        ('lot_id', 'in', not_selected_lots.ids),
+                    ], groupby=['lot_id'], aggregates=['available_quantity:sum'],
+                )
+            }
```
dhha-odoo:
```suggestion
            available_qty_by_lot = {}
            if not_selected_lots:
                available_qty_by_lot = {
                    lot.id: available_qty
                    for lot, available_qty in self.env['stock.quant']._read_group(
                        domain=[
                            ('product_id', '=', move_id.product_id.id),
                            ('location_id', 'child_of', move_id.location_id.id),
                            ('lot_id', 'in', not_selected_lots.ids),
                        ], groupby=['lot_id'], aggregates=['available_quantity:sum'],
                    )
                }
                move_uom_name = move_id.uom_id.name
```

Avoid calling **_read_group** if **not_selected_lots is empty**.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,53 @@ def _compute_lot_ids(self):
 
     def _set_lot_ids(self):
         for move in self:
-            if move.state == 'assigned' and all(ml.lot_id in move.lot_ids for ml in move.move_line_ids):
+            # NOTE: this early return is modified because even after the demand is fulfilled we should allow adding the
+            # lots and assigning the ML a proper lot/serial to it. Otherwise, when selecting/creating the lot after
+            # demand is fulfilled the quantity will add up +1 to it but the lot won't be assigned to the ML and thus
+            # wont't be added in the Tracking column too.
```
dhha-odoo:
```suggestion
            # NOTE: This early return is modified because even after the demand is fulfilled,
            # we still want to allow modifying the lots and update the related move lines accordingly.
```

Simple + No need to write specific case.

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,35 @@ def _read_group_location_id(self, locations, domain):
...
+    @api.depends_context('active_move_id', 'formatted_display_name')
+    def _compute_display_name(self):
+        ctx = self.env.context
+        if (
+            ctx.get('formatted_display_name')
+            and (move_id := self.env['stock.move'].browse(ctx.get('active_move_id')))
+            and move_id.has_tracking == 'lot' and move_id.location_usage == 'internal'
+        ):
+            not_selected_lots = self - move_id.lot_ids
+            move_uom_name = move_id.uom_id.name
```
dhha-odoo: It's incorrect to use the **move's unit of measure** here. We should use the **quant or product UoM** instead, as it becomes misleading.

For example, if a picking has a move with a **Pack of 6 UoM** and a **quant contains 5 units**, the picking would display **5 Pack of 6**, which incorrectly suggests a **total of 30 units(actually it's 5 units)**.

Image

addons/stock/models/stock_move.py
```diff
@@ -1369,9 +1383,43 @@ def _get_relevant_state_among_moves(self):
...
+                # only be assigned to the first lot that claimed it
+                no_lot_claimed = False
+                for lot in newly_added_lots:
+                    available_qty = available_qty_by_lot.get(lot._origin.id, 0)
+                    remaining_move_qty = demand - quantity
+                    if mls_without_lots and not no_lot_claimed:
+                        qty_to_add = 0
+                        no_lot_claimed = True
+                    elif remaining_move_qty > 0 and available_qty > 0:
+                        qty_to_add = min(available_qty, remaining_move_qty)
```
dhha-odoo: Here, you are taking a demand based on the unit, not on the UOM defined in the move, due to which the functional issue is occurring. [See the video]( https://www.awesomescreenshot.com/video/51268356?key=a581f4c64257bec355477ab58b98c164).

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,53 @@ def _compute_lot_ids(self):
...
+                            'quantity': 1 if is_serial_tracked else move_line.quantity,
                         }))
                         mls_without_lots -= move_line
                     else:  # No line without serial number, creates a new one.
-                        reserved_quants = self.env['stock.quant'].with_context(packaging_uom_id=move.packaging_uom_id)._get_reserve_quantity(move.product_id, move.location_id, 1.0, lot_id=lot)
-                        if reserved_quants and reserved_quants[0][0].lot_id:
+                        qty_to_reserve = 1 if is_serial_tracked or qty_to_assign <= 0 or qty_assigned >= demand \
+                            else min(qty_to_assign, demand - qty_assigned)
+                        reserved_quants = self.env['stock.quant'].with_context(packaging_uom_id=move.packaging_uom_id
+                            )._get_reserve_quantity(move.product_id, move.location_id, qty_to_reserve, lot_id=lot)
```
dhha-odoo: Improve indentation, please. It's really hard to read

addons/stock/models/stock_move.py
```diff
@@ -1369,9 +1383,43 @@ def _get_relevant_state_among_moves(self):
...
+                    domain=[
+                        ('product_id', '=', self.product_id.id),
+                        ('location_id', 'child_of', self.location_id.id),
+                        '|', ('lot_id', 'in', newly_added_lots.ids), ('lot_id', '=', False),
+                    ], groupby=['lot_id'], aggregates=['available_quantity:sum'],
+                )
+                available_qty_by_lot = {lot.id: available_qty for lot, available_qty in quants_by_lot}
+                demand = self.product_uom_qty
+                no_lot_available_qty = available_qty_by_lot.get(False, 0)
+                # in case we have mls without lots or quants without lots, it won't be reserved immediately and should
```
dhha-odoo:
```suggestion
                # in case we have mls/quants quants without lots, it won't be reserved immediately and should
```

dhha-odoo (changes requested):

addons/stock/models/stock_move.py
srap-odoo: Regarding the suggested change:

> `mls_without_lots = self.move_line_ids - mls_with_existing_lots`

While implementing this, I noticed that it would functionally break the [case](https://www.loom.com/share/f62218e5ffb342e7b9356bd007012a70), because `mls_with_existing_lots` does not include all move lines with lots, but only the move lines corresponding to the selected lots.

addons/stock/models/stock_move.py
srap-odoo: Removed this comment, as it was only added as a reviewer note.

addons/stock/views/stock_picking_views.xml
```diff
@@ -311,12 +311,13 @@
...
                                     <field name="lot_ids" widget="many2many_tags"
+                                        options="{'on_tag_click': 'open_form'}"
                                         column_invisible="parent.state == 'draft'"
                                         groups="stock.group_production_lot"
                                         invisible="not show_details_visible or has_tracking not in ['serial', 'lot']"
                                         optional="hide"
                                         create="parent.use_create_lots == True"
-                                        context="{'default_company_id': company_id, 'default_product_id': product_id, 'active_picking_id': parent.id}"
+                                        context="{'default_company_id': company_id, 'default_product_id': product_id, 'active_picking_id': parent.id, 'active_move_id': id}"
                                         domain="[('product_id','=',product_id)]"
```
dhha-odoo:
```suggestion
                                        domain="[('product_id', '=', product_id)]"
```

addons/stock/models/stock_move.py
```diff
@@ -1369,11 +1384,46 @@ def _get_relevant_state_among_moves(self):
...
+                        qty_to_add = min(no_lot_available_qty, remaining_move_qty)
+                        empty_quant_taken = True
+                    else:
+                        qty_to_add = 1
+                    quantity += qty_to_add
+
+        self.quantity = self.product_id.uom_id._compute_quantity(quantity, self.uom_id, round=False)
         base_location = self.location_id
         quants = self.env['stock.quant'].sudo().search([
+            ('tracking', '=', 'serial'),
```
dhha-odoo, line 1418: You should do it in a separate commit, as it's already an issue in master.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
+            qty_to_assign = move_uom._compute_quantity(move.quantity, product_uom) - qty_assigned
             for ml in mls_with_lots:
                 if ml.quantity and ml.lot_id not in move.lot_ids:
                     move_lines_commands.append((2, ml.id))
+                    released_qty = ml.uom_id._compute_quantity(ml.quantity, product_uom)
+                    qty_to_assign += released_qty
+                    qty_assigned -= released_qty
             ls = move.move_line_ids.lot_id
+            demand = move_uom._compute_quantity(move.product_uom_qty, product_uom)
+            is_serial_tracked = move.has_tracking == 'serial'
```
dhha-odoo: Define these variables along with the others at the top, so that all variables are declared in one place.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
+            product_uom = move.product_id.uom_id
+            qty_assigned = move_uom._compute_quantity(move._quantity_sml(), product_uom)
+            qty_to_assign = move_uom._compute_quantity(move.quantity, product_uom) - qty_assigned
             for ml in mls_with_lots:
                 if ml.quantity and ml.lot_id not in move.lot_ids:
                     move_lines_commands.append((2, ml.id))
+                    released_qty = ml.uom_id._compute_quantity(ml.quantity, product_uom)
+                    qty_to_assign += released_qty
+                    qty_assigned -= released_qty
             ls = move.move_line_ids.lot_id
```
dhha-odoo:
```suggestion
            ls = mls.lot_id
```
define above at the top

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
+            demand = move_uom._compute_quantity(move.product_uom_qty, product_uom)
+            is_serial_tracked = move.has_tracking == 'serial'
             for lot in move.lot_ids:
                 if lot not in ls:
                     if mls_without_lots[:1]:  # Updates an existing line without serial number.
                         move_line = mls_without_lots[:1]
```
dhha-odoo:
```suggestion
                    if move_line := mls_without_lots[:1]:
```

If you want to apply, as you have already improved this method.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
                         move_lines_commands.append(Command.update(move_line.id, {
                             'lot_id': lot.id,
-                            'uom_id': move.product_id.uom_id.id if move.product_id.tracking == 'serial' else move.uom_id.id,
-                            'quantity': 1 if move.product_id.tracking == 'serial' else move.quantity,
+                            'uom_id': product_uom.id if is_serial_tracked else move_line.uom_id.id,
+                            'quantity': 1 if is_serial_tracked else move_line.quantity,
```
dhha-odoo: This seems unnecessary.

You are updating the **same move line with its existing values for uom_id and quantity**, which does not introduce any actual change and results in redundant updates.

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
                     move_lines_commands.append((2, ml.id))
+                    released_qty = ml.uom_id._compute_quantity(ml.quantity, product_uom)
+                    qty_to_assign += released_qty
+                    qty_assigned -= released_qty
             ls = move.move_line_ids.lot_id
+            demand = move_uom._compute_quantity(move.product_uom_qty, product_uom)
+            is_serial_tracked = move.has_tracking == 'serial'
             for lot in move.lot_ids:
                 if lot not in ls:
                     if mls_without_lots[:1]:  # Updates an existing line without serial number.
```
dhha-odoo: Please update the comments (including a lot) with the new flow, same as above (if needed, elsewhere).

addons/stock/models/stock_move.py
```diff
@@ -609,39 +609,54 @@ def _compute_lot_ids(self):
...
+                            'quantity': 1 if is_serial_tracked else move_line.quantity,
                         }))
                         mls_without_lots -= move_line
                     else:  # No line without serial number, creates a new one.
-                        reserved_quants = self.env['stock.quant'].with_context(packaging_uom_id=move.packaging_uom_id)._get_reserve_quantity(move.product_id, move.location_id, 1.0, lot_id=lot)
-                        if reserved_quants and reserved_quants[0][0].lot_id:
+                        qty_to_reserve = (
+                            1.0 if is_serial_tracked or qty_to_assign <= 0 or qty_assigned >= demand
+                            else min(qty_to_assign, demand - qty_assigned)
+                        )
```
dhha-odoo:
```suggestion
                        qty_to_reserve = 1.0 if is_serial_tracked or qty_assigned >= demand else qty_to_assign
```

dhha-odoo (changes requested):
Hello @srap-odoo 

Here is my final review. Most of the comments are related to cleanup and minor improvements. Once it's done, it's ready from my side.

Thanks.

dhha-odoo:
Hello @srap-odoo 

I have applied the suggested changes. Since you are on leave and these were mostly cleanup/nitpick updates, I went ahead, as the task is ready for functional testing.

Before pushing to this branch, please rebase your branch on the latest version to avoid overriding these changes. Also, kindly review and verify the updates on your side.

Thank you.

addons/stock/models/stock_move.py
srap-odoo: Hello @dhha-odoo,

I’ve reverted this change as it could lead to issues when `qty_to_assign > demand`.
[Case](https://www.loom.com/share/636dcd41e8944c248a56b64f92cb9281)

Please let me know if you see this differently :relaxed:

addons/mrp/views/mrp_production_views.xml
```diff
@@ -506,15 +506,17 @@
...
+                                        invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
+                                        options="{'on_tag_click': 'open_form'}"
                                         create="parent.use_create_components_lots"
-                                        context="{'default_product_id': product_id}"
-                                        domain="[('product_id','=',product_id)]"/>
+                                        context="{'default_product_id': product_id, 'active_move_id': id}"
+                                        domain="
+                                            location_usage != 'internal' and [('product_id', '=', product_id)]
+                                            or [('product_id', '=', product_id), ('product_qty', '>', 0), ('quant_ids.location_id', 'child_of', location_id)]"
+                                    />
```
dhha-odoo, line 515: Instead of **defining the domain in multiple XML places**, we can move it directly to the **field definition**. This makes it easier to **maintain**, keeps the logic **centralised**, and avoids **duplicating** the same domain wherever lot_ids is used in views.

addons/stock/models/stock_move.py
```diff
@@ -189,7 +189,7 @@ class StockMove(models.Model):
     orderpoint_id = fields.Many2one('stock.warehouse.orderpoint', 'Original Reordering Rule', index=True)
     forecast_availability = fields.Float('Forecast Availability', compute='_compute_forecast_information', digits='Product Unit', compute_sudo=True)
     forecast_expected_date = fields.Datetime('Forecasted Expected date', compute='_compute_forecast_information', compute_sudo=True)
-    lot_ids = fields.Many2many('stock.lot', compute='_compute_lot_ids', inverse='_set_lot_ids', domain="[('product_id', '=', product_id)]", string='Serial Numbers', readonly=False)
+    lot_ids = fields.Many2many('stock.lot', compute='_compute_lot_ids', inverse='_set_lot_ids', domain="[('product_id', '=', product_id)]", string='Tracking', readonly=False)
```
dhha-odoo: You can also remove **readonly=False**, because [by default](https://github.com/odoo/odoo/blob/master/odoo/orm/fields.py#L473C20-L473C28) it's already false if the field is **computed and inverse**.

addons/stock/views/stock_picking_views.xml
```diff
@@ -311,14 +311,17 @@
...
                                         invisible="not show_details_visible or has_tracking not in ['serial', 'lot']"
                                         optional="hide"
                                         create="parent.use_create_lots == True"
-                                        context="{'default_company_id': company_id, 'default_product_id': product_id, 'active_picking_id': parent.id}"
-                                        domain="[('product_id','=',product_id)]"
-                                    />
+                                        context="{'default_company_id': company_id, 'default_product_id': product_id, 'active_picking_id': parent.id, 'active_move_id': id}"
+                                        domain="
+                                            location_usage != 'internal' and [('product_id', '=', product_id)]
+                                            or [('product_id', '=', product_id), ('product_qty', '&gt;', 0), ('quant_ids.location_id', 'child_of', location_id)]
```
dhha-odoo: Here you used `&gt;` and in the above domain you used `>`, both have the same meaning, so we can make it consistent across all.

addons/repair/views/repair_views.xml
```diff
@@ -175,12 +175,16 @@
...
+                                    column_invisible="parent.state == 'draft'"
+                                    invisible="not show_details_visible or has_tracking not in ['lot', 'serial']"
                                     optional="hide"
-                                    context="{'default_product_id': product_id}"
-                                    domain="[('product_id','=',product_id)]"
-                                    readonly="not id"/>
+                                    context="{'default_product_id': product_id, 'active_move_id': id}"
+                                    domain="
+                                        location_usage != 'internal' and [('product_id', '=', product_id)]
+                                        or [('product_id', '=', product_id), ('product_qty', '>', 0), ('quant_ids.location_id', 'child_of', parent.location_id)]"
```
dhha-odoo: **Why use parent.location_id?**

The **move’s location is linked to the repair location**, and in this flow, the move location is not intended to be changed. The **repair location itself defaults from the operation type**, which is also not editable. So, instead of relying on the **parent.location_id**, why not consistently use **location_id of move**?

This would help **maintain consistency across the module**.

addons/stock/tests/test_move.py
```diff
@@ -6827,3 +6827,28 @@ def test_scrap_unlink_behavior_on_discard_with_and_without_context(self):
...
+        Quant._update_available_quantity(product, stock_location, 20, lot_id=lot3)
+
+        delivery = self.env['stock.picking'].create({'picking_type_id': self.picking_type_out.id})
+        move = self._create_move(product, stock_location, self.customer_location, product_uom_qty=20, picking_id=delivery.id)
+        delivery.action_confirm()
```
dhha-odoo, line 6913: It would be good to add an assertion after the picking is confirmed to verify the expected lot assignment. This would ensure that by default only lot1 and lot2 are set on the move’s tracking field, and confirm that when we manually update it, it correctly changes to lot3. This will also strengthen the test coverage around the lot assignment behaviour.

addons/stock/models/stock_move.py
```diff
@@ -616,39 +616,49 @@ def _compute_lot_ids(self):
...
+                            move_line_vals['quantity'] = reserved_quants[0][1]
                         else:
                             move_line_vals = self._prepare_move_line_vals(quantity=0)
-                            move_line_vals['lot_id'] = lot.id
-                        move_line_vals['uom_id'] = move.product_id.uom_id.id
-                        move_line_vals['quantity'] = 1
+                            move_line_vals['quantity'] = 1.0
+                        move_line_vals['lot_id'] = lot.id
+                        move_line_vals['uom_id'] = product_uom.id
                         move_lines_commands.append((0, 0, move_line_vals))
```
dhha-odoo: I would improve this as well, using a command.create, if you want to. Same as above

addons/stock/models/stock_move.py
dhha-odoo: Hmm, that’s true, thanks for the video.

Still, could you try this please:
`
qty_to_reserve = 1.0 if is_serial_tracked or qty_assigned >= demand else min(qty_to_assign, demand - qty_assigned)
`
I believe **qty_to_assign <= 0** may be redundant given **qty_assigned >= demand(or vice versa)**, as I’m not seeing a scenario where they diverge.

addons/stock/models/stock_move.py
```diff
@@ -1425,7 +1425,6 @@ def _onchange_lot_ids(self):
         self.quantity = self.product_id.uom_id._compute_quantity(quantity, self.uom_id, round=False)
         base_location = self.location_id
         quants = self.env['stock.quant'].sudo().search([
-            ('tracking', '=', 'serial'),
```
dhha-odoo: It's wrong. In the first commit, you added this condition, and in this commit, you are removing it.
https://github.com/odoo/odoo/pull/246060#discussion_r3049548617

addons/stock/models/stock_move.py
```diff
@@ -1384,9 +1398,32 @@ def _get_relevant_state_among_moves(self):
...
+                            packaging_uom_id=self.packaging_uom_id
+                        )._get_reserve_quantity(self.product_id, self.location_id, remaining_move_qty, lot_id=lot._origin)
+                        qty_to_add = min(available_quants[0][1], remaining_move_qty) if available_quants else 1
+                    else:
+                        qty_to_add = 1
+                    quantity += qty_to_add
+
+        self.quantity = self.product_id.uom_id._compute_quantity(quantity, self.uom_id, round=False)
         base_location = self.location_id
         quants = self.env['stock.quant'].sudo().search([
```
dhha-odoo, line 1455: I think you forgot to apply this one: https://github.com/odoo/odoo/pull/246060#discussion_r3049548617

addons/stock/models/stock_move.py
```diff
@@ -21,6 +21,13 @@ class StockMove(models.Model):
     _order = 'sequence, id'
     _rec_name = 'reference'
 
+    def _get_lot_domain(self):
+        return """[('product_id', '=', product_id)] + (
+            [('id', 'in', context.get('lot_ids'))] if context.get('lot_ids')
+            else [('product_qty', '>', 0), ('quant_ids', 'any', [('location_id', 'child_of', location_id)])] if location_usage == 'internal'
+            else []
+        )"""
```
dhha-odoo:
```suggestion
    def _get_lot_domain(self):
        return """
            [('product_id', '=', product_id)] + (
                [('id', 'in', context.get('lot_ids'))]
                if context.get('lot_ids')
                else (
                    [('product_qty', '>', 0), ('quant_ids', 'any', [('location_id', 'child_of', location_id)])]
                    if location_usage == 'internal'
                    else []
                )
            )
        """
```

More readable.

srap-odoo: Hello @dhha-odoo,
As both versions look readable to me, I have applied the suggested change.
Thank you for the review!

addons/stock/models/stock_move.py
```diff
@@ -21,6 +21,19 @@ class StockMove(models.Model):
     _order = 'sequence, id'
     _rec_name = 'reference'
 
+    def _get_lot_domain(self):
+        return """
+            [('product_id', '=', product_id)] + (
```
amoyaux, line 26: try to use the Domain object now it exists

addons/stock/models/stock_lot.py
```diff
@@ -63,6 +63,36 @@ def _read_group_location_id(self, locations, domain):
...
+                    )
+                }
+            for lot in self:
+                available_qty = available_qty_by_lot.get(lot.id, False)
+                if available_qty:
+                    lot.display_name = f"{lot.name}\t--{available_qty} {move_uom.name}--"
+                else:
+                    lot.display_name = lot.name
+        else:
+            super()._compute_display_name()
```
amoyaux:
```suggestion
        super...
        if not self.env.context.get('formatted_display_name'):
            return
        move_id = self.env['stock.move'].browse(self.env.context.get('active_move_id')
        if not move_id or move_id.has_tracking != 'lot' or move_id.location_usage != 'internal':
            return
        not_selected_lots = self - move_id.lot_ids
        available_qty_by_lot = {}
        if not_selected_lots:
            move_uom = move_id.uom_id
            available_qty_by_lot = {
                lot.id: move_id.product_id.uom_id._compute_quantity(available_qty, move_uom)
                for lot, available_qty in self.env['stock.quant']._read_group(
                    domain=[
                        ('product_id', '=', move_id.product_id.id),
                        ('location_id', 'child_of', move_id.location_id.id),
                        ('lot_id', 'in', not_selected_lots.ids),
                    ], groupby=['lot_id'], aggregates=['available_quantity:sum'],
                )
            }
        for lot in self:
            available_qty = available_qty_by_lot.get(lot.id, False)
            if available_qty:
                lot.display_name = f"{lot.display_name}\t--{available_qty} {move_uom.name}--"
```

addons/stock/models/stock_move.py
srap-odoo: Hello @amoyaux ,

Thank you for the review!

I tried using a `Domain` object instead, but here's why it wasn't working:

When using `Domain('product_id', '=', product_id)`, it gets evaluated as `[('product_id', '=', 'stock.move.product_id')]` rather than being resolved to the actual `product_id` value of the record by the ORM.

Also, `self.product_id.id` would evaluate to `False` because it is evaluated at class level rather than record level.

So I kept it as a simple string domain instead of a `Domain` object, since this use case requires a dynamic domain.

Please let me know if I'm missing something :relaxed:

addons/stock/models/stock_lot.py
srap-odoo: I was trying to optimize the method by reducing super() calls based on the earlier review [discussion](https://github.com/odoo/odoo/pull/246060#discussion_r2972855055). For now, I'm reverting this change and keeping it simple.

dhha-odoo:
I am closing this PR because the task has been cancelled, as the purpose is already covered in this PR: https://github.com/odoo/odoo/pull/263037

PR #246064, opened 2026-01-28
[IMP] web,survey: display image filename

pyduf:
When an image was uploaded as an answer choice in a survey, the filename was visible as an extra column. However, upon saving the question, it was displaying the file size only (the extra column gets hidden), which is not the most informative. Here we remove the extra column for displaying the filename and make the column used to upload the image, displays the filename instead of the file size.

To fix this, we use the binary widget, and we add an else case in its template that always displays the filename if the fileNameField is provided (this case is triggered when editing a question choice on a new question after having uploaded a file and switched to the edition of the next choice).

Then we make the column "value_image_filename" invisible as the filename is now displayed in the value_image column (before the fix, the file size was displayed instead). We then delete the field has_image_only_suggested_answer as its only use was to hide/show the filename column which we now always hide.

Task-3580477

### DIFF
--- a/addons/survey/views/survey_question_views.xml
+++ b/addons/survey/views/survey_question_views.xml
@@ -15,7 +15,6 @@
                 <field name="page_id" invisible="1" required="False"/>
                 <field name="sequence" invisible="1"/>
                 <field name="scoring_type" invisible="1"/>
-                <field name="has_image_only_suggested_answer" invisible="1"/>
                 <sheet>
                     <field name="survey_id" invisible="not context.get('show_survey_field')" readonly="1"/>
                     <div class="float-end d-flex flex-column text-center" invisible="not is_page">
@@ -123,9 +122,9 @@
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
                                     <field name="answer_score"
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
-                                    <field name="value_image_filename" column_invisible="not parent.has_image_only_suggested_answer"/>
-                                    <field name="value_image" width="200px" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
-                                           column_invisible="parent.question_type in ['matrix', 'scale']" required="not value"/>
+                                    <field name="value_image_filename" column_invisible="True"/>
+                                    <field name="value_image" width="200px" widget="binary" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
+                                        column_invisible="parent.question_type in ['matrix', 'scale']" required="not value"/>
                                 </list>
                             </field>
 
--- a/addons/survey/models/survey_question.py
+++ b/addons/survey/models/survey_question.py
@@ -99,8 +99,6 @@ class SurveyQuestion(models.Model):
         'Scored', compute='_compute_is_scored_question',
         readonly=False, store=True, copy=True,
         help="Include this question as part of quiz scoring. Requires an answer and answer score to be taken into account.")
-    has_image_only_suggested_answer = fields.Boolean(
-        "Has image only suggested answer", compute='_compute_has_image_only_suggested_answer')
     # -- scoreable/answerable simple answer_types: numerical_box / date / datetime
     answer_numerical_box = fields.Float('Correct numerical answer', help="Correct number answer for this question.")
     answer_date = fields.Date('Correct date answer', help="Correct date answer for this question.")
@@ -244,13 +242,6 @@ class SurveyQuestion(models.Model):
     # -------------------------------------------------------------------------
     # COMPUTE METHODS
     # -------------------------------------------------------------------------
-
-    @api.depends('suggested_answer_ids', 'suggested_answer_ids.value')
-    def _compute_has_image_only_suggested_answer(self):
-        questions_with_image_only_answer = self.env['survey.question'].search(
-            [('id', 'in', self.ids), ('suggested_answer_ids.value', 'in', [False, ''])])
-        questions_with_image_only_answer.has_image_only_suggested_answer = True
-        (self - questions_with_image_only_answer).has_image_only_suggested_answer = False
 
     @api.depends('question_type')
     def _compute_question_placeholder(self):


addons/survey/views/survey_question_views.xml
```diff
@@ -123,9 +122,9 @@
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
                                     <field name="answer_score"
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
-                                    <field name="value_image_filename" column_invisible="not parent.has_image_only_suggested_answer"/>
-                                    <field name="value_image" width="200px" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
-                                           column_invisible="parent.question_type in ['matrix', 'scale']" required="not value"/>
+                                    <field name="value_image_filename" column_invisible="True"/>
+                                    <field name="value_image" width="200px" widget="list.binary" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
```
aab-odoo:
```suggestion
                                    <field name="value_image" width="200px" widget="binary" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
```

the "list" version of the binary widget will be used automatically

addons/web/static/src/views/fields/binary/binary_field.xml
```diff
@@ -63,6 +63,7 @@
                 <t t-if="fileName" t-esc="fileName" />
             </a>
         </t>
+        <t t-elif="props.fileNameField" t-out="fileName"/>
```
aab-odoo, line 61: A test would be nice :-)

aab-odoo (approved):
thanks

julien-banken (approved):

addons/web/static/src/views/fields/binary/binary_field.xml
awa-odoo, line 61: A little strange to check that we have a fileNameField here to output the fileName but the code right above this does not do the same, it just checks that we have a fileName at all.

awa-odoo, line 61: Few extra questions to make sure I understand the changes properly:

- This is only for the use-case where we just uploaded a file and it has not yet been saved on the record, otherwise it falls in the elif before this one, correct?
- What did it show before your changes? Just a blank cell?
- Does this work properly in other views as well (and not just on the list view where we are implementing it in survey)?

addons/survey/views/survey_question_views.xml
```diff
@@ -123,9 +122,9 @@
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
                                     <field name="answer_score"
                                         column_invisible="parent.scoring_type == 'no_scoring' or parent.question_type == 'matrix'"/>
-                                    <field name="value_image_filename" column_invisible="not parent.has_image_only_suggested_answer"/>
-                                    <field name="value_image" width="200px" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
-                                           column_invisible="parent.question_type in ['matrix', 'scale']" required="not value"/>
+                                    <field name="value_image_filename" column_invisible="True"/>
+                                    <field name="value_image" width="200px" widget="binary" filename="value_image_filename" options="{'accepted_file_extensions': 'image/*'}"
```
awa-odoo, line 126: Slightly counter-intuitive to force a binary field usage for an image field, but I'm guessing implementing this behavior in the image widget is not possible / much more complex? We can live with it, just trying to understand the reason :)

addons/web/static/tests/views/fields/binary_field.test.js
```diff
@@ -99,6 +99,48 @@ test("BinaryField is correctly rendered (readonly)", async () => {
...
+    expect(`.o_field_widget[name="document"]`).toHaveText(filenameDefaultValue, {
+        message: "the binary field should display the name of the file",
+    });
+    expect(`.o_field_widget[name="document"] a > .fa-download`).toHaveCount(0, {
+        message:
+            "the binary field should not be rendered as a downloadable link for a fresh record",
+    });
+    expect(`.o_field_char[name="foo"]`).toHaveText(filenameDefaultValue, {
+        message: "the filename field should have the file name as value",
+    });
```
awa-odoo: Does not seem related to our changes?

awa-odoo (commented):
Hello @pyduf 

A few comments / questions on this one.

Would also prefer if the web changes were isolated inside a separate commit if you don't mind.

Thanks & cheers!

addons/web/static/src/views/fields/binary/binary_field.xml
pyduf: Actually the fileName getter of the binary field returns the content of the file (the beginning in base64) if the “fileNameField” (or the record has no field of “fileNameField) is not provided. As I am altering what the widget displays I have added that condition to only display what is added when it has meaning (I suppose that displaying the beginning of the file in base64 has no use-case).

pyduf: Without the change (on master): when the line is selected, the filename is displayed and when it is not, the file size is displayed which is not user-friendly. The cause is that the ListRenderer doesn’t use the widget when the record is not in edition (line selected). This is determined by canUseFormatter which returns false when not in edition. In that case, the rendering is not performed by the widget but by getFormattedValue which ends up in formatBinary which compute the size of the file (I don’t think we can provide another formatter as the value is binary, we cannot extract the filename from it).

The solution is to force the widget even if the cell is not in edition (in that view there are not many lines so it shouldn’t be too costly). In that case the ListRenderer.canUseFormatter returns always false meaning the basic formatter is not used when the line is not selected but the widget instead.

The problem is that the widget in that case displays nothing (we are in the case “readonly” as not in edition). That’s why I am adding the extra “elif” to display the filename in that case if it is available (meaning if fileNameField is provided).

Other views if they do not force the widget on the field should still rely on the ListRenderer (no change). If they provide the widget (or in a form view), the only risk is that the filename is displayed instead of nothing.

addons/survey/views/survey_question_views.xml
pyduf: Before the change, it was already the widget binary used for edition. Here we also force it when not editing to display the filename. I actually proposed different solution on the task. One of then was using the widget image but it displays the image which is not ideal for the list view (make the rows very big).

addons/web/static/tests/views/fields/binary_field.test.js
pyduf: It is actually the field providing the filename to the field using the widget binary. I have removed that assertion and commented the form view instead. I’ve also made the field providing the filename invisible which is the common use-case as the filename is displayed by the binary widget.
```xml
<form edit="0">
    <!-- field holding the filename of the "document" file (see default value above) -->
    <field name="foo" invisible="1"/>
    <field name="document" filename="foo"/>
</form>
```

pyduf (commented):
Hello @awa-odoo , 

Thanks for the review and sorry for the delay. I have updated the code and answered your comments.

As suggested, I have also split the commit: one for the binary field changes and one for the implementation of the feature in survey. It makes indeed the thing simpler to understand.

awa-odoo:
@robodoo r+ rebase-merge

pyduf:
Hello @awa-odoo , 

For info, after having rebased (for the update), the first added binary field test was failing. To solve the problem, I have done the following change in the test:
```xml
<form edit="0">
  <!--...-->
  <field name="document" filename="foo"/>
</form>
```
->
```xml
<form>
  <!--...-->
  <field name="document" filename="foo" readonly="1"/>
</form>
```
The problem was that with edit=”0”, the field document was not in readonly anymore (I haven’t found the change that made that happens as edit=”0” is still used on forms in many other places). Here I've just forced it on the field itself which is equivalent (the test fails with the previous version of the binary field and not with the change) as we need to activate the added else part on the binary field template.

awa-odoo:
@robodoo r+

PR #246065, opened 2026-01-28
[FIX] point_of_sale: close session with SEPA first payment method

macocianradu:
SEPA expects a partner and a bank account. But the POS allows the user to sell to generic customers without a partner id or bank account. When SEPA direct debit is selected as the first incoming/outgoint payment method for the journal, closing the POS session will try to pos a SEPA payment which will fail when no partner is selected for the payment.

We've identified this before and excluded SEPA credit transfer from point of sales for issues with closing sessions, but the issue is still happening with SEPA direct debit. This fix will exclude the SEPA DD from the list of payment methods the POS is trying to use to post the transactions.

Task-[5358590](https://www.odoo.com/odoo/project/1737/tasks/5358590)

---


davidmonnom (approved):
@robodoo fw=no

PR #246069, opened 2026-01-28
[IMP] orm: deprecate Cache view of fields

kmagusiak:
The class ``Cache`` was an abstraction over the cache data that is located in the transaction. Since we refactored fields so that they manage the cache data on the transaction directly, ``transaction.cache`` is obsolete.

odoo/enterprise#105747

---

### DIFF
--- a/odoo/addons/base/tests/test_api.py
+++ b/odoo/addons/base/tests/test_api.py
@@ -1,9 +1,9 @@
-# -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
+from pprint import pformat
 
 from odoo import models, Command
 from odoo.addons.base.tests.common import SavepointCaseWithUserDemo
-from odoo.tools import mute_logger, unique, lazy
+from odoo.tools import SQL, mute_logger, unique, lazy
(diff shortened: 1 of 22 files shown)


rco-odoo (approved):
@robodoo r+

odoo/addons/base/tests/test_api.py
```diff
@@ -279,6 +279,57 @@ def test_60_cache(self):
...
+                        # the cache key 'company' actually comes from context
+                        # key 'allowed_company_ids' (see property env.company
+                        # and method env.cache_key())
+                        context['allowed_company_ids'] = [context.pop('company')]
+                    process(model.with_context(context), field, inner_cache)
+            else:
+                process(model, field, field_cache)
+
+        if invalids:
+            self.fail("Invalid cache: %s", pformat(invalids))
```
MiquelRForgeFlow, line 331: Image

MiquelRForgeFlow (commented):
odoolint fails.

odoo/addons/base/tests/test_api.py
kmagusiak: #246145

PR #246076, opened 2026-01-28
[FIX] web_editor: fix carousel preview height in Firefox

bvr-odoo:
Steps to reproduce (only with Firefox):

- Open the snippet dialog.
- Issue: The carousel snippets are too tall.

Bug introduced by this commit [1]

[1]: https://github.com/odoo/odoo/commit/78d33cf8b475f891dd95bec1b0c058446538824a

task-5156137


FrancoisGe:
@robodoo r+

PR #246080, opened 2026-01-28
[FIX] html_editor, web: ensure color picker appears above link popover

walidsahli:
Problem:
In Email Marketing, the color picker for custom links is hidden under the link popover, making it unusable.

Cause:
Overlays added in mass mailing use a default sequence of 1050, while the color picker uses the default sequence (50). As a result, the link popover overlays the color picker.

Solution:
Force the color picker sequence to 1051 (mass mailing default + 1) so it always appears above other overlays, including the link popover.

Steps to reproduce:
- Open Email Marketing.
- Add a link.
- Change its type to custom to enable text and fill color options.
- Open the color picker to change the text color.
- Observe that the color picker is hidden by the link popover.

opw-5866997

---

### DIFF
--- a/addons/html_editor/static/src/main/link/link_popover.js
+++ b/addons/html_editor/static/src/main/link/link_popover.js
@@ -210,6 +210,10 @@ export class LinkPopover extends Component {
                     },
                     {
                         env: this.__owl__.childEnv,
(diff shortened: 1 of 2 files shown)


dmo-odoo:
@robodoo r+

PR #246082, opened 2026-01-28
[REF] iot_drivers: simplify ctypes interfaces

loouis-t:
TIM and CTEP interfaces download their required libs from outside the Interface class, making it hard to read and understand (instantiation/download orders, ...). We simplified this inside simple methods.

In addition, CTEP interface was divided in two different interfaces: one for Windows IoT Boxes, the other for IoT Boxes. Both classes had very few differences, so we merged them, handling specific cases in the same file.


addons/iot_drivers/iot_handlers/interfaces/ctep_interface.py
```diff
@@ -0,0 +1,90 @@
...
+from odoo.addons.iot_drivers.tools.system import path_file, IS_RPI, IS_WINDOWS
+from odoo.addons.iot_drivers.iot_handlers.drivers.ctypes_terminal_driver import (
+    create_ctypes_string_buffer,
+    import_ctypes_library,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+def _configure_ctep_lib(lib_name: str) -> bool:
```
Ysoroko, line 19: is there a reason to have this method outside of the class if it's only called in its `__init__` method?

addons/iot_drivers/iot_handlers/interfaces/tim_interface.py
```diff
@@ -3,98 +3,103 @@
...
-# Make TIM SDK dependency libraries visible for the linker
-if IS_WINDOWS:
-    os.environ['PATH'] = file_path('iot_drivers/iot_handlers/drivers/tim') + os.pathsep + os.environ['PATH']
-else:
-    TIMAPI_DEPENDANCY_LIB = 'libtimapi.so.3'
-    TIMAPI_DEPENDANCY_LIB_V = f'{TIMAPI_DEPENDANCY_LIB}.38.0-5308'
-    DEP_LIB_PATH = file_path('iot_drivers/iot_handlers/drivers/tim')
-    USR_LIB_PATH = '/usr/lib'
+
+def _configure_tim_lib(lib_name: str) -> bool:
```
Ysoroko, line 16: Same question here

Ysoroko (approved):
Tested both on Windows and RPI
Worldline terminal is still detected and workind as expected ✅
Six terminal is still detected and working as expected ✅

The only comment here is to move methods `_configure_tim_lib` and `_configure_ctep_lib` inside the corresponding classes

Ysoroko:
robodoo r+

PR #246084, opened 2026-01-28
[FIX] account: prevent crash when selecting tax in journal item dialog

bhra-odoo:
**Steps to reproduce:**
* Install the **Accounting** module.
* Open a **Journal Entry** in mobile view.
* Add a **Journal Item**, opening it in a form dialog.
* Select a tax in the **Taxes Applied** field.
* Click on the **Save and close** button.

**Observed behavior:**
* The web client crashes with `TypeError: Cannot read properties of undefined (reading 'resId')`.

**Cause:**
* The `autosave_many2many_tags` widget triggers `model.root.save()` immediately when a tag is selected.
* When executing this save from within a transient dialog (common in mobile views), the client fails to correctly handle the record reload/synchronization leading to a crash when accessing `resId`.

**Fix:**
* Replace the autosave widget with the standard `many2many_tags` widget for **Journal Items** for form.
* Tax changes are now kept locally in the dialog and saved only when the user explicitly saves and closes it.

**Note:**
* The crash happens when `autosave_many2many_tags` calls
  `model.root.save()` on a **new** `account.move` from within a dialog.
* In QUnit tests, the mock server (`mockWebSave`) [1](https://github.com/odoo/odoo/blob/ca4d74c2a5d749ff8d41cd2fff80a73ba550a843/addons/web/static/tests/helpers/mock_server.js#L627). which creates records [2.](https://github.com/odoo/odoo/blob/ca4d74c2a5d749ff8d41cd2fff80a73ba550a843/addons/web/static/tests/helpers/mock_server.js#L1936)
  in-memory and does not trigger form reloads or component destruction.
* As a result, the crash cannot be reproduced in tests.

opw-5497342

### DIFF
--- a/addons/account/views/account_move_views.xml
+++ b/addons/account/views/account_move_views.xml
@@ -1299,7 +1299,7 @@
                                         <field name="debit" sum="Total Debit"/>
                                         <field name="credit" sum="Total Credit"/>
                                         <field name="balance" invisible="1"/>
-                                        <field name="tax_ids" string="Taxes Applied" widget="autosave_many2many_tags" options="{'no_create': True}"/>
+                                        <field name="tax_ids" string="Taxes Applied" widget="many2many_tags" options="{'no_create': True}"/>
                                         <field name="date_maturity" required="0" invisible="context.get('view_no_maturity', False)"/>
                                       </group>
                                     </form>


alsh-odoo (approved):

addons/account/views/account_move_views.xml
```diff
@@ -1299,7 +1299,7 @@
                                         <field name="debit" sum="Total Debit"/>
                                         <field name="credit" sum="Total Credit"/>
                                         <field name="balance" invisible="1"/>
-                                        <field name="tax_ids" string="Taxes Applied" widget="autosave_many2many_tags" options="{'no_create': True}"/>
+                                        <field name="tax_ids" string="Taxes Applied" widget="many2many_tags" options="{'no_create': True}"/>
```
epictete, line 1302: Aren't we losing the behavior that was introduced with the custom `autosave_many2many_tags` widget?
Isn't there a way to fix it without removing it?

bhra-odoo: Hello @epictete,
No, we're not losing any behavior in this specific case. Here's why:
The `autosave_many2many_tags widget` was designed for inline tree editing as per my understanding, where there's no explicit "Save" button. It auto-saves when:
1. A tax is added/removed
2. Balance, account, or partner changes (with taxes present)

However, in the dialog scenario (mobile view or non-editable tree):
* The dialog already has a "Save & Close" button
* Users expect to click that button to save their changes
* Auto-saving root record from within the dialog is not expected behavior, as the current record, which is open in dialog does not have resID until we save it, which causes the traceback

epictete: Thanks for the feedback :ok_hand: 
I agree the widget was designed for the tree view.
Just one small comment regarding:
> [...] for inline tree editing as per my understanding, where there's no explicit "Save" button [...]

There is a "Save manually" button:

image

Just FYI, it does not impact this PR :relaxed:

epictete (approved):
@robodoo r+

PR #246089, opened 2026-01-28
[FIX] account_edi_ubl_cii: fix EPD on credit notes

LouisGobert:
Submitting a credit note with an early payment discount fails with schematron errors `BR-CO-17` and `BR-S-09`.

The issue occurs because
`_get_early_payment_discount_grouped_by_tax_rate`
returns negative values for credit notes, which is not accepted by the Peppol validation rules since we will increase the `TaxableAmount` instead of decreasing them.

Steps to reproduce:
- Create a payment term with a 2% discount (always)
- Create a credit note using this payment term
- Submit the document to Peppol
- Observe the schematron validation error

opw-5357180
opw-5499252


LouisGobert:
@robodoo fw=no

clbr-odoo (approved):
LGTM, thanks !
The test reflects behavior in 18.0+.
@robodoo r+

PR #246095, opened 2026-01-28
[IMP] hr_*: remove sequence field

nikj-odoo:
This commit removes the sequence field, as contract types are already ordered by name by default, making the sequence field unnecessary.

Task:5868349

### DIFF
--- a/addons/hr/data/hr_data.xml
+++ b/addons/hr/data/hr_data.xml
@@ -74,62 +74,50 @@
 
         <record id="contract_type_permanent" model="hr.contract.type">
             <field name="name">Permanent</field>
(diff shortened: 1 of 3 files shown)


karygauss03 (approved):
Good job 💯 Please change the commit title to IMP instead of REM, we usually use REM when we remove a module.

karygauss03 (changes requested):
Good job 💯 Please change the commit title to IMP instead of REM, we usually use REM when we remove a module.

nikj-odoo:
Hello @karygauss03 :wave: 
I've changed the commit title.

Thanks :slightly_smiling_face:

PR #246097, opened 2026-01-28
[FIX] *: check ids in res_ids

kmagusiak:
`browse(x.mapped('res_id'))` may fail if there are res_ids that are equal to 0. Update existing code.

odoo/enterprise#105767

---

### DIFF
--- a/addons/account/controllers/download_docs.py
+++ b/addons/account/controllers/download_docs.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 import io
 import zipfile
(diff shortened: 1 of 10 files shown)


oomsveta (commented):
Hello, thank you for the fix!

That said, it seems a bit cumbersome to have to guard against falsy values every time you want to call `browse`. Maybe case 0 could be handled explicitly by the ORM instead? The current approach might not be ideal in terms of DX and maintainability: developers have to write extra code, and if they forget, the line becomes a liability in case of a 0 ID. The current filtering logic could also hide bugs since it doesn't explicitly check for 0 but rather filters out all falsy values. While 0 could arguably be considered an acceptable ID, this is not the case for other falsy values.

What do you think?

rco-odoo (approved):

rco-odoo:
@robodoo r+

tde-banana-odoo:
Have to say I agree writing if weird. Shouldn't we instead do a filtered('res_id').mapped('res_id'), would be a bit more explicit on "we filter on records that really link other reoocrds) ?

PR #246098, opened 2026-01-28
[IMP] point_of_sale: remove `Point of Sales` menu

paan-odoo:
In this commit:

- We are removing the menu of `Point of Sale List` view from the configuration Tab.

Task-5880979


davidmonnom (approved):
@robodoo r+

PR #246102, opened 2026-01-28
[FIX] pos_event: make tickets with seats_max=0(unlimited) available in pos

Nokoz:
We had a bug when selling event tickets in POS where tickets with seats_max set to 0 (which should mean unlimited) appeared as sold out when the event had seats_limited enabled.

Steps to reproduce:
-------------------
* Create an event with seats_limited=True and seats_max > 0
* Create a ticket for that event with seats_max=0 (unlimited)
* Open POS and view the event product

> Observation:
Ticket appears as "Sold out" instead of showing available seats

Why the fix:
------------
The frontend logic only treated tickets with seats_max=0 as unlimited when the event itself was not limited (!event.seats_limited). However, according to the backend logic, seats_max=0 means the ticket is unlimited regardless of the event's limitation status. When the event is limited, unlimited tickets should still respect the event's seats_available limit, not show as sold out.

We now check for seats_max=0 independently of event limitation. If the event is not limited, tickets with seats_max=0 are fully unlimited. If the event is limited, we use the event's seats_available value (since the ticket doesn't constrain availability but the event does), ensuring correct display of availability status.

opw-5475803


addons/pos_event/static/src/app/screens/product_screen/product_screen.js
```diff
@@ -125,8 +125,8 @@ patch(ProductScreen.prototype, {
             slotSelected = this.pos.models["event.slot"].get(slotResult.slotId);
         } else {
             avaibilityByTicket = tickets.reduce((acc, ticket) => {
-                if (ticket.seats_max === 0 && !event.seats_limited) {
-                    acc[ticket.id] = "unlimited";
+                if (ticket.seats_max === 0) {
+                    acc[ticket.id] = event.seats_limited ? event.seats_available : "unlimited";
```
robinengels, line 129: I don't understand why we modify this. If we have this scenario :
Event limit is 1000
Ticket 1 is unlimited
Ticket 2 is 200 max

The availability for ticket 1 would be 1000 ? But 200 of them should be reserved for Ticket 2 no ?

Nokoz: No, there is no repartition for tickets. The unlimited ticket is constrained only by the event cap, not by “reserving” some seats for other tickets.

stva-odoo (approved):
robodoo r+

PR #236109, opened 2025-11-17
[REF] html_editor, *: refactor resource functions

Zynton:
Enterprise PR: https://github.com/odoo/enterprise/pull/100343

\*: `html_builder`, `mail`, `mass_mailing`, `website`, `website_forum`, `website_mass_mailing`, `website_sale`.

This PR stems from an observation that predicate resources in the editor are used inconsistently: for one to return `true`, sometimes all resources of that name have to return `true`, sometimes it's enough for none of them to never return `false`, sometimes it's even for just one to return `true` (and any `false` will be ignored).

Similarly, other special resources in the editor which take functions tend to be used inconsistently in spite of their common suffix ("predicates", "handlers", "processors", "overrides"). The suffix makes us suppose a common behavior that actually isn't always true and we need to go check how the resource is used on a case by case basis.

This PR is a first step towards standardizing these resources and giving them util functions that define their usage more precisely.

In the process:
(1) resources were renamed which were misleadingly named before ;
(2) `processThrough` was introduced to handle "processor" resources (with the `_processors` suffix) ;
(3) "handler" resources were renamed so as to be prefixed by `on_` to emphasize their function as listeners, reacting to events;
(4) `dispatchTo` was replaced by `trigger` (which returns the return values as an array), and `triggerAsync` (which waits for each function call so as to make them sequentially);
(4) some minor refactors were applied ;
(5) `checkPredicates` was introduced to handle "predicate" resources (with the `_predicates` suffix), standardizing ternary predicate logic in the editor ;
(6) predicates formulated in the negative were changed to positives (`isX` rather than `isNotX`);
(7) predicates were renamed to be in the form of questions;
(8) notably, `clean_for_save_handlers` and `normalize_handlers` were turned into processors and renamed accordingly.

Standard function resources are now:

| **Utils**         | **Suffix**  | **Returns**                                                        | **Example**                                                                         |
|-------------------|-------------|--------------------------------------------------------------------|-------------------------------------------------------------------------------------|
| `trigger`         | _handlers   | `any[]`                                                            | `const values = this.trigger("on_my_event_handlers", arg1, arg2)`                   |
| `triggerAsync`    | _handlers   | `Promise<void>`                                                    | `await this.triggerAsync("on_my_sequential_async_handlers", arg1, arg2)`           |
| `delegateTo`      | _overrides  | `boolean`: `true` if an override was found to stop the propagation | `if (this.delegateTo("my_command_overrides", arg1, arg2)) return`                    |
[...]

PR #246103, opened 2026-01-28
[REF] html_editor: make areSimilarElements overridable by plugins

bso-odoo:
In [1] when `areSimilarElements` was adapted to recognize inline code, the conditions were embedded within the tool function.

This commit makes those conditions delivered through plugin resources.

[1]: https://github.com/odoo/odoo/commit/9539985287de6cf3ddd758550d13b01e29bea913

task-5487650

### DIFF
--- a/addons/html_editor/static/src/core/format_plugin.js
+++ b/addons/html_editor/static/src/core/format_plugin.js
@@ -58,9 +58,18 @@ function isFormatted(formatPlugin, format) {
  *      applyStyle: boolean,
  * }) => void | boolean)[]} format_selection_overrides
  * @typedef {(() => void)[]} on_all_formats_removed_handlers
+ * @typedef {((root: Node) => void)[]} on_will_merge_adjacent_siblings_handlers
+ * @typedef {((root: Node) => void)[]} on_merged_adjacent_siblings_handlers
  *
  * @typedef {((className: string) => boolean | undefined)[]} is_format_class_predicates
  * @typedef {((node: Node) => boolean | undefined)[]} has_format_predicates
+ *
+ * @typedef {(({
+ *      node: Node,
+ *      nodeStyle: CSSStyleProperties,
+ *      node2: Node,
+ *      node2Style: CSSStyleProperties,
+ * }) => void | true)[]} are_similar_elements_overrides
  */
 
 export class FormatPlugin extends Plugin {
@@ -709,6 +718,7 @@ export class FormatPlugin extends Plugin {
      * @param {boolean} [options.preserveSelection=true]
      */
     mergeAdjacentInlines(root, { preserveSelection = true } = {}) {
+        this.trigger("on_will_merge_adjacent_siblings_handlers", root);
         let selectionToRestore = null;
         for (const node of [root, ...descendants(root)].filter(isElement)) {
             if (this.shouldBeMergedWithPreviousSibling(node)) {
@@ -716,34 +726,19 @@ export class FormatPlugin extends Plugin {
                     selectionToRestore ??= this.dependencies.selection.preserveSelection();
                     selectionToRestore.update(callbacksForCursorUpdate.merge(node));
                 }
-                if (node.matches("code.o_inline_code")) {
-                    while (
-                        node.previousSibling?.nodeType === Node.TEXT_NODE &&
-                        /^\uFEFF*$/.test(node.previousSibling.nodeValue)
-                    ) {
-                        node.previousSibling.remove();
-                    }
-                }
                 node.previousSibling.append(...childNodes(node));
                 node.remove();
             }
         }
         selectionToRestore?.restore();
+        this.trigger("on_merged_adjacent_siblings_handlers", root);
     }
 
     shouldBeMergedWithPreviousSibling(node) {
         const isMergeable = (node) =>
             FORMATTABLE_TAGS.includes(node.nodeName) &&
             (this.checkPredicates("is_node_splittable_predicates", node) ?? true);
-        let previousSibling = node.previousSibling;
-        if (node.matches("code.o_inline_code")) {
-            while (
-                previousSibling?.nodeType === Node.TEXT_NODE &&
-                /^\uFEFF*$/.test(previousSibling.nodeValue)
-            ) {
-                previousSibling = previousSibling.previousSibling;
-            }
-        }
+        const previousSibling = node.previousSibling;
         return (
             !isSelfClosingElement(node) &&
             isMergeable(node) &&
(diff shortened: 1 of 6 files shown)


addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -672,7 +680,11 @@ export class FormatPlugin extends Plugin {
...
                 }
-                if (node.matches("code.o_inline_code")) {
+                if (
+                    this.getResource("format_mergeable_feff_predicates").some((predicate) =>
+                        predicate(node)
+                    )
+                ) {
                     while (
                         node.previousSibling?.nodeType === Node.TEXT_NODE &&
                         /^\uFEFF*$/.test(node.previousSibling.nodeValue)
```
dmo-odoo: I think it would make more sense if it was something like merge_adjacent_siblings_handlers so that the FeffPlugin could register one such handler which would call its removeFeffs() method. Feffs are meant to be removed and automatically re-added and it makes more sense if the FeffPlugin is the one to handle it. If we do this early enough in this function then we might not need to skip them in `shouldBeMergedWithPreviousSibling` either since they will have been removed already.

addons/html_editor/static/src/main/inline_code.js
```diff
@@ -19,6 +19,17 @@ export class InlineCodePlugin extends Plugin {
...
             ),
+        format_are_similar_overrides: ({ node, nodeStyle, node2Style }) => {
+            if (node.matches("code.o_inline_code")) {
+                if (
+                    nodeStyle.padding === node2Style.padding &&
+                    nodeStyle.margin === node2Style.margin
+                ) {
+                    return true;
+                }
+            }
```
dmo-odoo: I'm not sure I understand this. Is there a way to change the padding and margin of the inline code? The override could simply do "if node and node2 are both inline codes then return true".

addons/html_editor/static/src/main/inline_code.js
```diff
@@ -19,6 +19,17 @@ export class InlineCodePlugin extends Plugin {
             selectElements(root, ".o_inline_code").flatMap((code) =>
                 this.dependencies.feff.surroundWithFeffs(code, cursors)
             ),
+        format_are_similar_overrides: ({ node, nodeStyle, node2Style }) => {
```
dmo-odoo: Rename to `are_similar_elements_overrides`.

addons/html_editor/static/src/utils/dom_info.js
```diff
@@ -818,7 +818,7 @@ function hasPseudoElementContent(node, pseudoSelector) {
 
 const NOT_A_NUMBER = /[^\d]/g;
 
-export function areSimilarElements(node, node2) {
+export function areSimilarElements(node, node2, override) {
```
dmo-odoo: I don't like passing resources to utils because it means the function behaves differently when called as a util and when called from a plugin, assuming the plugin passes some overrides of course. I would rather move this to `DomPlugin` so that we don't pass resources to utils, if possible, depending on whether it's only called from plugins or not.

addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -669,21 +677,14 @@ export class FormatPlugin extends Plugin {
      * @param {boolean} [options.preserveSelection=true]
      */
     mergeAdjacentInlines(root, { preserveSelection = true } = {}) {
+        this.dispatchTo("merge_adjacent_siblings_handlers", root);
```
dmo-odoo: In light of https://github.com/odoo/odoo/pull/236109 I think this should be renamed to `on_will_merge_adjacent_siblings_handlers`. It's actually more correct in this case. The system says "I'm about to merge adjacent siblings" and the `FeffPlugin` knows it needs to remove its feffs for it to work properly. Sorry for the back and forth on naming these.

addons/html_editor/static/src/main/feff_plugin.js
```diff
@@ -53,6 +54,21 @@ export class FeffPlugin extends Plugin {
         }
     }
 
+    mergeAdjacentSiblingsHandler(root) {
+        for (const node of [root, ...descendants(root)].filter(isElement)) {
+            if (
+                this.getResource("mergeable_feff_predicates").some((predicate) => predicate(node))
```
dmo-odoo: In my original comment, I meant to _always_ remove the feffs before merging similar elements, regardless of whether they are mergeable or not. We remove and put back the feffs at multiple places already. Remove all feffs, merge similar elements, put back the feffs.

bso-odoo: @dmo-odoo Doing this (see third commit) leads to various side-effects: <https://runbot.odoo.com/runbot/build/102866348>
- merged `font` nodes, why not
- spaces that become `&nbsp;`
- disappearing links (zwnbsp-isolated ones - maybe this specific case needs to be handled)

dmo-odoo: I had a look at the failing tests:
- merged font nodes is actually better
- feffs inside an empty link where the cursor is currently sitting should _not_ be removed, otherwise it will eject the cursor, it's a bug in removeFeffs that is revealed by your change
- for the others, notably "should be able to remove color of an icon", "should insert a <br> after a format node (3)" and  "should insert ﻿ at the end of format node", it looks like it's not putting back the
feffs after the operation?

bso-odoo:
@dmo-odoo It took ages (it needed #260333) - but tests are finally green again after the latest changes.

addons/html_editor/static/tests/insert/line_break.test.js
```diff
@@ -298,15 +298,6 @@ describe("Selection collapsed", () => {
...
 
-        test("should insert \uFEFF at the end of format node", async () => {
-            await testEditor({
-                contentBefore: "<p><b>abc[]</b><br><br></p>",
-                stepFunction: insertLineBreak,
-                contentAfterEdit: `<p><b>abc<br>[]\uFEFF</b><br><br></p>`,
-                contentAfter: "<p><b>abc<br>[]</b><br><br></p>",
-            });
-        });
-
```
dmo-odoo, line 309: I was a bit wary of this one so I tested it and it seems it doesn't work properly after your changes.

Steps to reproduce:
- Create a new project task
- Press Shift+Enter
- Press ArrowUp to return to the first line
- Type abc
- Select abc using the keyboard (Shift+ArrowLeft)
- Press CTRL+B to bold it
- Press ArrowRight to be at the end of abc
- Press Shift+Enter
- Type something, notice you don't type in bold

bso-odoo: @dmo-odoo Restored tests and added missing call to `would_feff_be_legit_predicates`. Runbot is green again.
```js
    onWillMergeAdjacentSiblingsHandler(root) {
        const cursors = this.getCursors();
        this.removeFeffs(root, cursors, {
            exclude: (node) =>
                this.checkPredicates("would_feff_be_legit_predicates", node) ?? false,
        });
        cursors.restore();
    }
```

bso-odoo:
@dmo-odoo Finally green... I had to fight against some `Invalid selection. Node is not part of the editable: #text` warnings, which I ultimately solved with this:
<https://github.com/odoo/odoo/pull/246103/changes#diff-d172392dd0aa85e3ce8e1e308230594e076032e4bb81e044b515117d104b675fR147-R154>

dmo-odoo:
@robodoo r+

dmo-odoo:
@robodoo rebase-ff

PR #246107, opened 2026-01-28
[IMP] html_editor: map pasted table cells into target table 

adch-odoo:

- When pasting clipboard content that contains  table inside existing table, inserted the pasted table as a nested table element.

- Extends `addRow()` and `addColumn()` to accept `rowsToAdd/columnsToAdd` parameters (default 1), allowing multiple rows/columns to be inserted in single call.
- The target cells are filled cell-by-cell using the pasted table cells content.
- Single-cell paste preserves inline formatting (bold, links, lists, etc.).
- If the pasted table requires more cells than available from paste anchor position, target table is expanded by adding the required rows and columns to fit the incoming content.
- After paste, the cells whose content was updated are selected to highlight the affected cells.
- This improved behavior is applied only when:
   - Clipboard content contains a single table only, and
   - The paste selection/cursor is inside a single target table.

task-5499414

---

### DIFF
--- a/addons/html_editor/static/src/main/table/table_plugin.js
+++ b/addons/html_editor/static/src/main/table/table_plugin.js
@@ -1375,9 +1375,9 @@ export class TablePlugin extends Plugin {
      * @param {import("@html_editor/core/selection_plugin").EditorSelection} selection
      */
     processContentForClipboard(clonedContents, selection) {
-        if (clonedContents.firstChild.nodeName === "TR" || isTableCell(clonedContents.firstChild)) {
+        const table = closestElement(selection.commonAncestorContainer, "table.o_selected_table");
+        if (table) {
             // We enter this case only if selection is within single table.
-            const table = closestElement(selection.commonAncestorContainer, "table");
             const tableClone = table.cloneNode(true);
             // A table is considered fully selected if it is nested inside a
             // cell that is itself selected, or if all its own cells are
--- a/addons/html_editor/static/src/main/table/table_plugin.js
+++ b/addons/html_editor/static/src/main/table/table_plugin.js
@@ -269,97 +269,99 @@ export class TablePlugin extends Plugin {
     /**
      * @param {'before'|'after'} position
      * @param {HTMLTableCellElement} reference
+     * @param {number} [columnsToAdd=1]
      */
-    addColumn(position, reference) {
-        const columnIndex = getColumnIndex(reference);
+    addColumn(position, reference, columnsToAdd = 1) {
         const table = closestElement(reference, "table");
         const tableWidth = table.style.width && parseFloat(table.style.width);
-        const referenceColumn = table.querySelectorAll(
-            `tr :is(td, th):nth-of-type(${columnIndex + 1})`
-        );
-        const referenceCellWidth = reference.style.width
-            ? parseFloat(reference.style.width)
-            : reference.clientWidth;
-        // Temporarily set widths so proportions are respected.
-        const firstRow = table.querySelector("tr");
-        const firstRowCells = [...firstRow.children].filter(
-            (child) => child.nodeName === "TD" || child.nodeName === "TH"
-        );
-        let totalWidth = 0;
-        if (tableWidth) {
-            for (const cell of firstRowCells) {
-                const width = parseFloat(cell.style.width);
-                cell.style.width = width + "px";
-                // Spread the widths to preserve proportions.
-                // -1 for the width of the border of the new column.
-                const newWidth = Math.max(
-                    Math.round((width * tableWidth) / (tableWidth + referenceCellWidth - 1)),
-                    13
-                );
-                cell.style.width = newWidth + "px";
-                totalWidth += newWidth;
+        for (let columnOffset = 0; columnOffset < columnsToAdd; columnOffset++) {
+            const columnIndex = getColumnIndex(reference);
+            const referenceCellWidth = reference.style.width
+                ? parseFloat(reference.style.width)
+                : reference.clientWidth;
+
+            // Temporarily set widths so proportions are respected.
+            const firstRowCells = [...table.rows[0].cells];
+            let totalWidth = 0;
+            if (tableWidth) {
+                for (const cell of firstRowCells) {
+                    const width = parseFloat(cell.style.width);
+                    cell.style.width = width + "px";
+                    // Spread the widths to preserve proportions.
+                    // -1 for the width of the border of the new column.
+                    const newWidth = Math.max(
+                        Math.round((width * tableWidth) / (tableWidth + referenceCellWidth - 1)),
+                        13
+                    );
+                    cell.style.width = newWidth + "px";
+                    totalWidth += newWidth;
+                }
             }
-        }
-        referenceColumn.forEach((cell, rowIndex) => {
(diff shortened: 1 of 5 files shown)


addons/html_editor/static/src/main/table/table_plugin.js
```diff
@@ -252,97 +248,111 @@ export class TablePlugin extends Plugin {
...
+                        newCell.style.width = cell.style.width;
+                        totalWidth += parseFloat(cell.style.width);
+                    }
+                }
+
+                // When inserting multiple columns "after", update the reference so
+                // columns are appended in order.
+                if (position === "after" && cell === reference) {
+                    reference = newCell;
+                }
```
deso-odoo: Does it matter? We are going to append empty cells.

addons/html_editor/static/src/main/table/table_plugin.js
```diff
@@ -177,6 +179,81 @@ export class TablePlugin extends Plugin {
...
+        const sourceColumnCount = sourceRows[0]?.cells.length || 0;
+        if (!sourceRowCount || !sourceColumnCount) {
+            return false;
+        }
+
+        // Selection must be fully inside the same target table.
+        const targetTable = closestElement(selection.anchorNode, "table");
+        if (!targetTable || closestElement(selection.focusNode, "table") !== targetTable) {
+            return false;
+        }
```
deso-odoo, line 228: Can be combined.

addons/html_editor/static/src/main/table/table_plugin.js
```diff
@@ -177,6 +179,81 @@ export class TablePlugin extends Plugin {
...
+
+        const anchorCell =
+            targetTable.querySelector(".o_selected_td") ||
+            closestElement(selection.anchorNode, isTableCell);
+        const anchorRowIndex = getRowIndex(anchorCell);
+        const anchorColumnIndex = getColumnIndex(anchorCell);
```
deso-odoo, line 234: Might break after merge table cells PR, make sure to check with @dhba-odoo.

addons/html_editor/static/src/main/table/table_plugin.js
```diff
@@ -177,6 +179,81 @@ export class TablePlugin extends Plugin {
...
+            const targetCells = targetTable.rows[anchorRowIndex + sourceRowIndex].cells;
+            for (
+                let sourceColumnIndex = 0;
+                sourceColumnIndex < sourceColumnCount;
+                sourceColumnIndex++
+            ) {
+                const sourceCell = sourceCells[sourceColumnIndex];
+                const targetCell = targetCells[anchorColumnIndex + sourceColumnIndex];
+                const clonedCell = sourceCell.cloneNode(true);
+                targetCell.replaceChildren(...clonedCell.childNodes);
```
deso-odoo, line 263: replaceChildren? What if copied cell had background color?

addons/html_editor/static/tests/paste.test.js
```diff
@@ -4830,3 +4830,271 @@ describe("onDrop", () => {
...
+                        <tr>
+                            <td class="o_selected_td"><p>[x11</p></td>
+                            <td class="o_selected_td"><p>x12</p></td>
+                        </tr>
+                        <tr>
+                            <td class="o_selected_td"><p>x21</p></td>
+                            <td class="o_selected_td"><p>x22]</p></td>
+                        </tr>
+                    </tbody>
+                </table>
```
deso-odoo, line 5003: You selected a1 and a2 cells only where did my b1 b2 cells go....

addons/html_editor/static/tests/paste.test.js
```diff
@@ -4830,3 +4830,271 @@ describe("onDrop", () => {
...
+                        <tr>
+                            <td><p>b1</p></td>
+                            <td><p>b2</p></td>
+                        </tr>
+                    </tbody>
+                </table>
+                <p data-selection-placeholder="" style="margin: -9px 0px 8px;"><br></p>
+            `)
+        );
+    });
```
deso-odoo, line 5049: I see this test case as, previous table was completely selected and i'm pasting another table on top of it, shouldn't the new table be replace by the previous one? why keep b1 and b2 cell?

deso-odoo (commented):
Added some remarks. Some test cases feels off to me, I would suggest you to discuss those cases with SHRA.

addons/html_editor/static/src/main/table/table_plugin.js
adch-odoo: Agree they’re empty, but order still matters: without updating reference, every insert happens “after the same cell”, so 3 inserts become ref | c3 | c2 | c1. Updating it gives the expected ref | c1 | c2 | c3 (deterministic).

addons/html_editor/static/src/main/table/table_plugin.js
adch-odoo: Yes, can be combined, but I kept it separate so we only call closestElement after the quick source-size checks. It avoids an extra DOM walk when we already know we’ll return false.

addons/html_editor/static/src/main/table/table_plugin.js
adch-odoo: The task mentions “content only,” but I confirmed with Shra that it’s acceptable to also apply background color styles of pasted cells. I’ve made the changes accordingly.

addons/html_editor/static/tests/paste.test.js
adch-odoo: This was inspired by Google Docs, but their behavior is a bit odd. If the copied range is smaller than the selected target range, it repeats the content to fill the selection, if the copied range is larger, it only updates the selected cells and drops the rest. I chose a simpler approach where we paste everything once (no repetition and no dropping). I checked with SHRA and he’s fine with this.

addons/html_editor/static/tests/paste.test.js
adch-odoo: Same applies here as well: [https://github.com/odoo/odoo/pull/246107/changes#r2758027081](https://github.com/odoo/odoo/pull/246107/changes#r2758027081)

addons/html_editor/static/src/main/table/table_plugin.js
deso-odoo: > every insert happens “after the same cell”, so 3 inserts become ref | c3 | c2 | c1. Updating it gives the expected ref | c1 | c2 | c3 (deterministic).

I get that. But my question is does it matter? What issue does | c3 | c2 | c1 causes that | c1 | c2 | c3 don't?

addons/html_editor/static/src/main/table/table_plugin.js
deso-odoo: > but I confirmed with Shra and it’s ok to also apply cell styles.

If so use of `replaceChildren` is incorrect, isn't it?

> and it’s ok to also apply cell styles.

which cell styles? copied cell or the cell upon which you'll paste.

addons/html_editor/static/src/main/table/table_plugin.js
adch-odoo: Since the inserted cells are empty, the order doesn’t matter here. If they weren’t empty, it would matter. Removed this logic.

PR #246118, opened 2026-01-28
[IMP] base: allow programmatic creation of API keys

madprog:
Previously, API keys could only be created interactively: the method required an interactive session and a recently typed password.

This commit adds a new method on `res.users.apikeys` that allows creating an API key when an existing one is provided. Optionally, in the case of a renewal, the new key can replace the old one, and the provided one is unlinked.

### DIFF
--- a/odoo/addons/base/models/res_users.py
+++ b/odoo/addons/base/models/res_users.py
@@ -25,7 +25,7 @@ from odoo.api import SUPERUSER_ID
 from odoo.exceptions import AccessDenied, AccessError, UserError, ValidationError
 from odoo.fields import Command, Domain
 from odoo.http import request, DEFAULT_LANG
-from odoo.tools import email_domain_extract, is_html_empty, frozendict, reset_cached_properties, SQL
+from odoo.tools import email_domain_extract, is_html_empty, frozendict, reset_cached_properties, str2bool, SQL
 
 
 _logger = logging.getLogger(__name__)
@@ -1513,6 +1513,7 @@ KEY_CRYPT_CONTEXT = CryptContext(
     # attacks on API keys isn't much of a concern
     ['pbkdf2_sha512'], pbkdf2_sha512__rounds=6000,
 )
+DEFAULT_PROGRAMMATIC_API_KEYS_LIMIT = 10  # programmatic API key creation is refused if the user already has at least this amount of API keys
 
 
 class ResUsersApikeys(models.Model):
@@ -1566,6 +1567,7 @@ class ResUsersApikeys(models.Model):
             _logger.info("API key(s) removed: scope: <%s> for '%s' (#%s) from %s",
                self.mapped('scope'), self.env.user.login, self.env.uid, ip)
             self.sudo().unlink()
+            self.env.registry.clear_cache()
             return {'type': 'ir.actions.act_window_close'}
         raise AccessError(_("You can not remove API keys unless they're yours or you are a system user"))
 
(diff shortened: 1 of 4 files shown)


odoo/addons/base/models/res_users.py
```diff
@@ -1570,23 +1570,28 @@ def _remove(self):
...
+                    k.expiration_date >= now() at time zone 'utc'
                 )
         '''.format(self._table),
         [index, scope])
-        for user_id, current_key in self.env.cr.fetchall():
+        for key_id, user_id, current_key in self.env.cr.fetchall():
             if key and KEY_CRYPT_CONTEXT.verify(key, current_key):
-                return user_id
+                return key_id, user_id
+        return None, None, None
```
mwath: 👀 
```suggestion
        return None, None
```

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1633,31 @@ def _generate(self, scope, name, expiration_date):
...
+        """
+        logout_env = api.Environment(self.env.cr, None, {})
+        uid = logout_env['res.users.apikeys']._check_credentials(scope=scope or 'rpc', key=key)
+        if not uid or uid != self.env.uid:
+            raise UserError(_("The provided API key is invalid or does not belong to the current user"))
+        new_key = self._generate(scope, name, expiration_date)
+
+        old_index = key[:INDEX_SIZE]
+        new_index = new_key[:INDEX_SIZE]
+        ip = request.httprequest.environ['REMOTE_ADDR'] if request else 'n/a'
```
mwath: did you forget to use `ip` in the logging? 👀

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1635,31 @@ def _generate(self, scope, name, expiration_date):
...
         return k
 
+    @api.model
+    def generate(self, key, scope, name, expiration_date, renew=False):
+        """
+        Generate a new API key with an existing API key.
+        The provided `key` must be an existing API key that belongs to the current user.
+        Its scope must be compatible with `scope`.
+        The `expiration_date` must be allowed for the user's group.
+        If `renew` is true, the previous key will be disabled.
```
kmagusiak: Usually, the update of the key is done in steps:
1. Create new token
2. Update another system to use the new token
3. Wait that the system is updated (it may be using the old token)
4. Revoke the old token

So revoking directly seems just a bad idea. Let's say you execute it and fail to save the new token due to any error: you have no longer access.

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1635,31 @@ def _generate(self, scope, name, expiration_date):
 
         return k
 
+    @api.model
+    def generate(self, key, scope, name, expiration_date, renew=False):
+        """
+        Generate a new API key with an existing API key.
```
kmagusiak, line 1652: This seems like generating new keys to avoid expiration dates. If you need a key and can generate only the same scope, you must have a use case in mind. Could you explain it?

odoo/addons/base/models/res_users.py
odony: maybe we can use `with user.assert_can_auth()` to benefit from a basic rate limiting, by raising an `AccessDenied` if no key matched. Hiding the result of the revocation does not bring much added security (you can test the key before/after directly too)

odony: I think we should limit API key generation:
- based on a global ICP that the admin can enable (and we could default to False for now)
- to a maximum of e.g. 10 API keys per user (to prevent accidental loops and errors. Can be an ICP if we want.)

madprog: This would work for json2 calls, but xmlrpc borrow the request, so `assert_can_auth` will enter its `if not request` section and do nothing => that would be a virtual rate-limiting protection as it could be easily bypassed by using xmlrpc.
We will instead disable this route by default too, and only enable it if the global ICP `base.enable_programmatic_api_keys` is truthy.

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1630,87 @@ def _generate(self, scope, name, expiration_date):
...
+        The `expiration_date` must be allowed for the user's group.
+
+        To renew a key, generate the new one, store it, and then call `revoke` on the previous one.
+        """
+        self._ensure_can_manage_keys_programmatically()
+
+        if not isinstance(expiration_date, datetime.datetime):
+            expiration_date = fields.Datetime.from_string(expiration_date)
+
+        nb_keys = self.search_count([('user_id', '=', self.env.uid), ('expiration_date', '>=', self.env.cr.now())])
```
odony:
```suggestion
        nb_keys = self.search_count([('user_id', '=', self.env.uid), '|',('expiration_date', '=', False), ('expiration_date', '>=', self.env.cr.now())])
```

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1630,87 @@ def _generate(self, scope, name, expiration_date):
...
+            raise UserError(_('Limit of %s API keys is reached for programmatic creation', nb_keys_limit))
+
+        # Scope compatibility rules:
+        # - A global key can generate credentials for any scope (including global).
+        # - A scoped key can only generate credentials for its own scope.
+        #
+        # This is enforced in _check_credentials by validating scope usage,
+        # and the validated scope is then reused when calling _generate.
+
+        logout_env = api.Environment(self.env.cr, None, {})
```
odony: we can use self.env as discussed

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1630,87 @@ def _generate(self, scope, name, expiration_date):
...
+
+    @api.model
+    def revoke(self, key):
+        """
+        Revoke an existing API key.
+        If it exists, the `key` will be removed from the server.
+        """
+        self._ensure_can_manage_keys_programmatically()
+        assert key, "key required"
+        self.env.cr.execute(f'''
```
odony: we can use the `SQL` construct

odony (approved):
thanks, this will be welcome for /json2 RPC users who will be able to renew API keys automatically, in a safer manner than having long-life keys.

odoo/addons/base/models/res_users.py
Julien00859: I'm +1 to use `assert_can_auth`. xmlrpc is deprecated and will be removed at one point, so there is a point in the future where all that remains is json2 and we would benefit from that ratelimit.

odoo/addons/base/tests/test_res_users.py
```diff
@@ -668,3 +669,67 @@ def test_revoke_all_devices(self):
...
+        cls.tomorrow = datetime.now() + timedelta(days=1)
+        cls.unscoped_key = UsersApiKeys._generate(None, 'Key without a scope', cls.tomorrow)
+        cls.scoped_key = UsersApiKeys._generate('scope', 'Key with a scope', cls.tomorrow)
+
+    def test_generate_apikey_is_disabled_by_default(self):
+        self.env['ir.config_parameter'].set_param('base.enable_programmatic_api_keys', None)
+
+        # Attempting to create an key raises an error
+        with self.assertRaisesRegex(UserError, 'Programmatic API keys are not enabled'):
+            self.env['res.users.apikeys'].with_user(self.user_internal).generate(self.unscoped_key, None, 'Another key without a scope', self.tomorrow)
```
Julien00859:
```suggestion
        with self.assertRaises(UserError) as capture:
            self.env['res.users.apikeys'].with_user(self.user_internal).generate(self.unscoped_key, None, 'Another key without a scope', self.tomorrow)
        self.assertEqual(capture.exception.args, ("Programmatic API keys are not enabled",))
```
The second argument of `assertRaises` and `assertRaisesRegexp` is actually the *assert error message to emit would the assertion fails*, not the *error string to match*

odoo/addons/base/tests/test_res_users.py
```diff
@@ -668,3 +669,67 @@ def test_revoke_all_devices(self):
...
+
+        # Attempting to revoke a key raises an error
+        with self.assertRaisesRegex(UserError, 'Programmatic API keys are not enabled'):
+            self.env['res.users.apikeys'].with_user(self.user_internal).revoke(self.unscoped_key)
+
+    def test_generate_apikey_is_limited(self):
+        # create 8 new keys, which makes 10 keys in total for user_internal
+        for i in range(8):
+            self.env['res.users.apikeys'].with_user(self.user_internal).generate(self.unscoped_key, None, 'Another key without a scope', self.tomorrow)
```
Julien00859: we don't all have 4k screens...

odoo/addons/test_http/tests/test_webjson2.py
```diff
@@ -250,3 +250,69 @@ def test_webjson2_url_params_vs_body_params(self):
                 })
             url_search.assert_called()
             body_search.assert_not_called()
+
+    def test_renew_apikey(self):
+        self.env['ir.config_parameter'].set_param('base.enable_programmatic_api_keys', 1)
+        key = self.bearer_header['Authorization'][7:]
```
Julien00859:
```suggestion
        key = self.bearer_header['Authorization'].removeprefix('Bearer ')
```

odoo/addons/base/models/res_users.py
```diff
@@ -1628,6 +1630,87 @@ def _generate(self, scope, name, expiration_date):
...
+        # This is enforced in _check_credentials by validating scope usage,
+        # and the validated scope is then reused when calling _generate.
+
+        uid = self.env['res.users.apikeys']._check_credentials(scope=scope or 'rpc', key=key)
+        if not uid or uid != self.env.uid:
+            raise UserError(_("The provided API key is invalid or does not belong to the current user"))
+        new_key = self._generate(scope, name, expiration_date)
+        _logger.info("%s %r generated from %r", self._description, new_key[:INDEX_SIZE], key[:INDEX_SIZE])
+
+        return {'result': new_key}
```
Julien00859: Why `return {'result': new_key}` and not just `return new_key`?

odoo/addons/test_http/tests/test_webjson2.py
```diff
@@ -250,3 +250,69 @@ def test_webjson2_url_params_vs_body_params(self):
...
+        apikey = self.env['res.users.apikeys'].search([('user_id', '=', self.jackoneill.id)])
+
+        in_ten_minutes = (datetime.now() + timedelta(minutes=10)).replace(microsecond=0)
+        in_twenty_minutes = (datetime.now() + timedelta(minutes=20)).replace(microsecond=0)
+
+        # an API key can be used to create a new one
+        res = self.db_url_open(
```
Julien00859, line 263: add `res.raise_for_status()` immediately after the url-open you expect to work

odoo/addons/test_http/tests/test_webjson2.py
```diff
@@ -250,3 +250,69 @@ def test_webjson2_url_params_vs_body_params(self):
...
+        # the second key is now revoked
+        res = self.db_url_open(
+            '/json/2/res.users/context_get',
+            headers={
+                **CT_JSON,
+                'Authorization': f'Bearer {key2}',
+            },
+            data='{}',
+        )
+        self.assertEqual(res.status_code, 401)
```
Julien00859:
```suggestion
        self.assertEqual(res.status_code, HTTPStatus.UNAUTHORIZED)
```

odoo/addons/base/tests/test_res_users.py
madprog: The second argument of `assertRaises` is indeed the message to emit, but the second argument of `assertRaisesRegex` is actually a regex.

Julien00859: TIL thanks!

Julien00859 (approved):

odoo/addons/base/tests/test_res_users.py
```diff
@@ -668,3 +669,77 @@ def test_revoke_all_devices(self):
...
+
+        # Attempting to create a key raises an error
+        with self.assertRaisesRegex(UserError, 'Programmatic API keys are not enabled'):
+            self.env['res.users.apikeys'].with_user(self.user_internal).generate(
+                self.unscoped_key, None, 'Another key without a scope', self.tomorrow)
+
+        # Attempting to revoke a key raises an error
+        with self.assertRaisesRegex(UserError, 'Programmatic API keys are not enabled'):
+            self.env['res.users.apikeys'].with_user(self.user_internal).revoke(self.unscoped_key)
+
```
Levizar: naming

Levizar (approved):
nitpick done IRL

madprog:
@robodoo r+

PR #246121, opened 2026-01-28
[IMP] (test_)mail: improve a bit access / perf coverage

tde-banana-odoo:


tde-banana-odoo:
@robodoo r+ rebase-ff

PR #246122, opened 2026-01-28
[IMP] mail: add default static id for js models

phenix-factory:
task-4801145

PR enterprise: https://github.com/odoo/enterprise/pull/105785


seb-odoo (commented):
Ok for the default, but I would prefer if actual singleton explicitly defined themselves as such, rather than relying on a potentially undefined/non-existing id field, with all the problem that would happen if we did add this field on them by mistake.

I still want to also clean the localId of singleton records (especially the key in localstorage) to not have undefined on them, so we need to be able to detect them with certainty, and not just by their implicit lack of existing id field.

As we have for AND and OR, I'd suggest defining a symbol for the singleton case and using it explicitly as static id, rather than adding another property for that.

seb-odoo (approved):

PR #246125, opened 2026-01-28
[FIX] stock: consider MO for daily demand in forecasting

josv-odoo:
The daily demand in forecasting does not take into consideration the product that are used for production.

Steps to reproduce:
-------------------
* Install stock and mrp
* Create a tracked Product "Component" with RR
* Add some "on hand quantity" 
* Create Product "Product" with a BOM that use "Component"
* Create a manufacturing order for "Product" and confirm it
* Go back on the RR of "Component" and use "resplenishment information" button (small i) 
-> Daily demand does not reflect the "Component" used for manufacturing.

Observation:
-------------
When Opening the resplenishment information, the daily_demand is calculated: https://github.com/odoo/odoo/blob/c38346ccc3e44670b76bfb010795bdfa7004b532/addons/stock/wizard/stock_replenishment_info.py#L177 For that calculation quantity_out is used, obtained just before: https://github.com/odoo/odoo/blob/c38346ccc3e44670b76bfb010795bdfa7004b532/addons/stock/wizard/stock_replenishment_info.py#L164-L167 Since it only consider moves that goes directly to the client, it will not take into consideration products that are used for production.

opw-5497584

### DIFF
--- a/addons/stock/wizard/stock_replenishment_info.py
+++ b/addons/stock/wizard/stock_replenishment_info.py
@@ -162,7 +162,7 @@ class StockReplenishmentInfo(models.TransientModel):
                 [('company_id', '=', replenishment_report.orderpoint_id.company_id.id)],
             ])
             quantity_out = self.env['stock.move']._read_group(
-                Domain.AND([domain, [('location_dest_id.usage', '=', 'customer')]]),
+                Domain.AND([domain, [('location_dest_id.usage', 'in', ['customer', 'production'])]]),
                 aggregates=['product_qty:sum'],
             )[0][0] or 0.0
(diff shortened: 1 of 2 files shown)


addons/stock/wizard/stock_replenishment_info.py
adwid: I'm wondering whether we are considering the finished product here, while we should not
Can you check ?

josv-odoo: You are right!
Yes, we did consider finished product that were coming in (product coming from a manufacturing order)

Since we only want to know the "demand" of this product, we only want to know what is going out, minus the ones that were not necessary (customer returns).

I changed it!
Thanks :)

PR #246127, opened 2026-01-28
[IMP] survey: allow to set/unset mandatory question from list view

pyduf:
To change whether a survey question is mandatory or not, you had to open the question and modify it in the option section. Now with this change, we can modify it directly by switching the state in the question list views (you need first to show the "Mandatory Answer" column).

Task-5865119


julien-banken (approved):

tde-banana-odoo:
@robodoo r+

PR #246133, opened 2026-01-28
[FIX] account: UserError in linking of bills at import

gawa-odoo:
A UserError there can cause duplicates in peppol fetching.
It should never block an ack.
We can leave them unlinked
We also do the same for any unexpected error that occurs after we've created the move.

opw-5492092

---

### DIFF
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -4539,7 +4539,10 @@ class AccountMove(models.Model):
         if new_lines := (self.invoice_line_ids - existing_lines):
             new_lines.is_imported = True
             if not existing_lines:
-                self.with_context(default_move_type=self.move_type)._link_bill_origin_to_purchase_orders(timeout=4)
+                try:
+                    self.with_context(default_move_type=self.move_type)._link_bill_origin_to_purchase_orders(timeout=4)
+                except (UserError, ValueError):
+                    _logger.exception("Failed to link bill to purchase order")
 
         return res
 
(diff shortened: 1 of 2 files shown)


addons/account_peppol/models/account_edi_proxy_user.py
```diff
@@ -151,17 +151,20 @@ def _peppol_import_invoice(self, attachment, peppol_state, uuid, journal=None):
         })
         if 'is_in_extractable_state' in move._fields:
             move.is_in_extractable_state = False
+            try:
```
clbr-odoo: you "tabbed" everything under the `is_in_extractable_state` if

addons/account/models/account_move.py
```diff
@@ -4539,7 +4539,10 @@ def _extend_with_attachments(self, files_data, new=False):
         if new_lines := (self.invoice_line_ids - existing_lines):
             new_lines.is_imported = True
             if not existing_lines:
-                self.with_context(default_move_type=self.move_type)._link_bill_origin_to_purchase_orders(timeout=4)
+                try:
+                    self.with_context(default_move_type=self.move_type)._link_bill_origin_to_purchase_orders(timeout=4)
+                except (UserError, ValueError):
+                    _logger.exception("Failed to link bill to purchase order")
```
clbr-odoo, line 4545: What was the exact error that was raised here ?
It is worth doing a post on the invoice to inform the user ?

gawa-odoo: Not sure.
I don't have a clear way of knowing what exactly was the problem. In the ticket situation, when linking both, we arrive in https://github.com/odoo/odoo/blob/saas-18.4/addons/purchase/models/account_invoice.py#L209
Which causes a `Any journal item on a payable account must have a due date and vice versa`
I can guess there is a misconfig somewhere, but I don't think there is a single most likely answer on why it would raise there

clbr-odoo (approved):
@robodoo r+

PR #246146, opened 2026-01-28
[REF] point_of_sale, pos_glory_cash: generic support for cash machines

MaxTheWhale:
Before this commit, cash machines as a payment method were not supported at all in the base `point_of_sale` module, and this meant overrides were needed in order to support the Glory cash machine in `pos_glory_cash`.

After this commit, cash machines as a concept are supported and can be registered by modules in the same way as payment terminals. The desired module can be automatically installed from the payment method screen just like for payment terminals.

This paves the way for adding support for additional cash machines beyond Glory.

task-5888201

---


stva-odoo (approved):
Great work!
Sorry for the review delay, don’t hesitate to ping me in this case.

PR #246149, opened 2026-01-28
[IMP] website_sale: filter attribute values

livaios:
Change the current behaviour of the Attributes filters by displaying only the values that are relevant on the loaded /shop page. 

task-4423954

---

### DIFF
--- a/addons/website_sale/controllers/main.py
+++ b/addons/website_sale/controllers/main.py
@@ -494,21 +494,22 @@ class WebsiteSale(payment_portal.PaymentPortal):
         )
 
         ProductAttribute = request.env["product.attribute"]
+        pavs_per_attribute = {}
         if products:
-            # get all products without limit
-            attributes_grouped = request.env["product.template.attribute.line"]._read_group(
+            grouped_pavs = request.env["product.attribute.value"]._read_group(
                 domain=[
-                    ("product_tmpl_id", "in", search_product.ids),
+                    ("pav_attribute_line_ids.product_tmpl_id", "in", search_product.ids),
                     ("attribute_id.visibility", "=", "visible"),
                 ],
                 groupby=["attribute_id"],
                 order="attribute_id",
+                aggregates=["id:recordset"],
             )
-            attribute_ids = [attribute.id for (attribute,) in attributes_grouped]
-            attributes = ProductAttribute.browse(attribute_ids)
+            pavs_per_attribute = dict(grouped_pavs)
+            # Return attributes as recordset of `product.attribute`
+            attributes = ProductAttribute.union(pavs_per_attribute.keys())
         else:
             attributes = ProductAttribute.browse(attribute_ids).sorted()
-
         products_prices = products._get_sales_prices(website)
         product_query_params = self._get_product_query_params(**post)
 
@@ -553,6 +554,7 @@ class WebsiteSale(payment_portal.PaymentPortal):
             "previewed_attribute_values": lazy(
                 lambda: products._get_previewed_attribute_values(category, product_query_params)
             ),
+            "pavs_per_attribute": pavs_per_attribute,
         }
         if filter_by_price_enabled:
             values["min_price"] = min_price or available_min_price
--- a/addons/website_sale/templates/shop_page_templates.xml
+++ b/addons/website_sale/templates/shop_page_templates.xml
@@ -679,7 +679,7 @@
     <template id="website_sale.products_shop_title_align" active="False" name="Align Shop Title"/>
 
     <template id="website_sale.filter_color_attributes" name="Color attributes in Filter">
-        <t t-foreach="a.value_ids" t-as="v">
+        <t t-foreach="pavs_per_attribute[a]" t-as="v">
             <t t-set="img_style"
                t-value="'background:url(/web/image/product.attribute.value/%s/image); background-size:cover;' % v.id if v.image else ''"
             />
@@ -701,7 +701,7 @@
     </template>
 
     <template id="website_sale.filter_image_attributes" name="Image attributes in Filter">
-        <t t-foreach="a.value_ids" t-as="v">
+        <t t-foreach="pavs_per_attribute[a]" t-as="v">
             <label
                 t-attf-class="css_attribute_image size_small position-relative rounded-3 bg-body text-center cursor-pointer #{'active' if v.id in attrib_set else ''} #{'custom_value' if v.is_custom else ''} #{'transparent' if (not v.is_custom and not v.html_color) else ''}"
             >
@@ -724,7 +724,7 @@
     <template id="website_sale.filter_select_attributes" name="Select attributes in Filter">
         <select class="form-select css_attribute_select" name="attribute_value">
             <option value="" selected="true">All <t t-out="a.name"/></option>
-            <t t-foreach="a.value_ids" t-as="v">
+            <t t-foreach="pavs_per_attribute[a]" t-as="v">
                 <option
                     t-att-value="'%s-%s' % (a.id,v.id)"
                     t-out="v.name"
@@ -740,7 +740,7 @@
     >
         <t
             t-set="sorted_values"
-            t-value="sorted(a.value_ids, key=lambda v: v.id not in attrib_set)"
+            t-value="sorted(pavs_per_attribute[a], key=lambda v: v.id not in attrib_set)"
         />
         <t t-if="len(sorted_values) &lt;= 20">
             <div
@@ -814,7 +814,7 @@
                 <input
                     type="text"
                     class="o_wsale_attribute_search_bar form-control"
-                    t-attf-placeholder="Search {{a.name}} ({{len(a.value_ids)}})"
+                    t-attf-placeholder="Search {{a.name}} ({{len(sorted_values)}})"
                     t-att-data-container-id="'searchable-values-%s' % a.id"
                 />
             </div>
@@ -847,7 +847,7 @@
     </template>
 
     <template id="website_sale.filter_pills_attributes" name="Pills attributes in Filter">
-        <t t-foreach="a.value_ids" t-as="v">
+        <t t-foreach="pavs_per_attribute[a]" t-as="v">
             <input
                 type="checkbox"
                 name="attribute_value"
@@ -876,15 +876,22 @@
         >
             <t
                 t-set="visible_attributes"
-                t-value="len([a for a in attributes if a.value_ids and len(a.value_ids) &gt; 1])"
+                t-value="len([a for a in attributes if pavs_per_attribute[a] and len(pavs_per_attribute[a]) &gt; 1])"
             />
 
             <t t-foreach="attributes" t-as="a">
                 <t t-set="_status" t-value="'inactive'"/>
-                <t t-foreach="a.value_ids" t-as="v" t-if="v.id in attrib_set" t-set="_status" t-value="'active'"/>
+                <t t-set="current_attribute_pavs" t-value="pavs_per_attribute[a]" />
+                <t
+                    t-foreach="current_attribute_pavs"
+                    t-as="v"
+                    t-if="v.id in attrib_set"
+                    t-set="_status"
+                    t-value="'active'"
+                />
                 <t t-set="_expand_attributes" t-value="_status == 'active' or visible_attributes &lt; 4"/>
                 <div
-                    t-if="a.value_ids and len(a.value_ids) &gt; 1"
+                    t-if="current_attribute_pavs and len(current_attribute_pavs) &gt; 1"
                     t-attf-class="accordion-item {{('border-top-0 ' + ('order-1' if _status == 'active' else 'order-2')) if isMobile else 'nav-item mb-1 rounded-0'}}"
                 >
                     <h2
--- a/addons/test_website_modules/tests/test_performance.py
+++ b/addons/test_website_modules/tests/test_performance.py
@@ -295,7 +295,7 @@ class TestWebsiteAllPerformance(TestWebsitePerformanceCommon, TestWebsitePriceLi
         self.assertIn(f'<img src="/web/image/product.template/{self.productA.product_tmpl_id.id}/', html)
         self.assertIn(f'<img src="/web/image/product.image/{self.product_images.ids[1]}/', html)
 
-        query_count = 49  # To increase this number you must ask the permission to al
+        query_count = 48  # To increase this number you must ask the permission to al
         queries = {
             'orm_signaling_registry': 1,
             'website': 1,
@@ -305,7 +305,7 @@ class TestWebsiteAllPerformance(TestWebsitePerformanceCommon, TestWebsitePriceLi
             'product_tag': 1,
             'product_public_category': 5,
             'product_product': 1,
-            'product_template_attribute_line': 3,
+            'product_template_attribute_line': 2,
             'res_users': 1,
             'res_partner': 2,
             'product_category': 1,


addons/website_sale/controllers/main.py
```diff
@@ -473,21 +474,23 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
                 ],
                 groupby=['attribute_id'],
-                order='attribute_id'
+                order='attribute_id',
+                aggregates=['id:recordset'],
             )
-            attribute_ids = [attribute.id for attribute, in attributes_grouped]
-            attributes = ProductAttribute.browse(attribute_ids)
+            for attribute, pavs in grouped_pavs:
+                pavs_per_attribute[attribute] = pavs
```
lipiraux:
```suggestion
            pavs_per_attribute.update(grouped_pavs)
```

addons/website_sale/controllers/main.py
```diff
@@ -473,21 +474,23 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
+            attributes = pavs_per_attribute.keys()
```
lipiraux: Is it fine if its not the exact same type? I guess its only used to be iterated over 🤔

```suggestion
            attributes = ProductAttribute.concat(*pavs_per_attribute.keys())
```

addons/website_sale/controllers/main.py
```diff
@@ -473,21 +474,23 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
         product_variants = dict(zip(products, variants))
 
         ProductAttribute = request.env['product.attribute']
+        pavs_per_attribute = defaultdict(list)
```
lipiraux: Same suggestion about keeping the types consitent.
```suggestion
        pavs_per_attribute = defaultdict(request.env['product.attribute.value'].browse)
```

addons/website_sale/controllers/main.py
```diff
@@ -473,21 +474,23 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
-            attributes_grouped = request.env['product.template.attribute.line']._read_group(
+            grouped_pavs = request.env['product.attribute.value']._read_group(
                 domain=[
-                    ('product_tmpl_id', 'in', search_product.ids),
+                    ('pav_attribute_line_ids.product_tmpl_id.id', 'in', search_product.ids),
                     ('attribute_id.visibility', '=', 'visible'),
                 ],
                 groupby=['attribute_id'],
-                order='attribute_id'
+                order='attribute_id',
```
lipiraux: why do we sort by id, when the else block bellow uses the default order? Is it for performance sake?
Cause we could do it with the else block too and avoid a fetch of the sequence just to sort them, wdyt?

https://github.com/odoo-dev/odoo/blob/932cb48d712f47d468617373b4d0e070eb9fef16/addons/website_sale/controllers/main.py#L493-L493

https://github.com/odoo-dev/odoo/blob/932cb48d712f47d468617373b4d0e070eb9fef16/addons/product/models/product_attribute.py#L8-L12

lipiraux: Never mind, we need to fetch the name and the display_type either way 😞

addons/website_sale/controllers/main.py
lipiraux: Why do you need a defaultdict if its added to the render values only if it was populated?

addons/website_sale/controllers/main.py
```diff
@@ -526,6 +529,8 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
                 lambda: products._get_previewed_attribute_values(category, product_query_params),
             ),
         }
+        if pavs_per_attribute:
+            values['pavs_per_attribute'] = pavs_per_attribute
```
lipiraux: 💥 If no products are found => no pavs_per_attribute, but it doesn't mean they won't be accessed.

Image

<details>
<summary>Traceback</summary>
```
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/addons/base/models/ir_qweb.py", line 759, in _render_iterall
    for item in frame.iterator:
                ^^^^^^^^^^^^^^
  File "<2136>", line 497, in template_website_sale_product_attribute_filters_form_2136
  File "<2136>", line 36, in template_website_sale_product_attribute_filters_form_2136_content
KeyError: 'pavs_per_attribute'

The above exception was the direct cause of the following exception:

  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/requestlib.py", line 608, in _serve_db
    return retrying(serve_func, env=self.env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/retrying.py", line 45, in retrying
    result = func()
             ^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/requestlib.py", line 663, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/dispatcher.py", line 191, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/addons/base/models/ir_http.py", line 376, in _dispatch
    result.flatten()
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/tools/facade.py", line 83, in wrap_func
    func(self._wrapped__, *args, **kwargs)
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/response.py", line 91, in flatten
    self.response.append(self.render())
                         ^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/http/response.py", line 83, in render
    return request.env["ir.ui.view"]._render_template(self.template, self.qcontext)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/addons/website/models/ir_ui_view.py", line 459, in _render_template
    return super()._render_template(template, values=values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/addons/base/models/ir_ui_view.py", line 2530, in _render_template
    return self.env['ir.qweb']._render(template, values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/addons/base/models/ir_qweb.py", line 731, in _render
    return Markup(''.join(iterator))
                  ^^^^^^^^^^^^^^^^^
  File "/Users/lipiraux/dev/odoo/review/odoo/odoo/addons/base/models/ir_qweb.py", line 857, in _render_iterall
    raise QWebError(qweb_error_info) from error
odoo.addons.base.models.ir_qweb.QWebError: Error while rendering the template:
    KeyError: 'pavs_per_attribute'
    Template: website_sale.product_attribute_filters_form
    Reference: 2136
    Path: /t/form/t[1]
    Element: <t t-set="visible_attributes" t-value="len([a for a in attributes if pavs_per_attribute[a] and len(pavs_per_attribute[a]) &gt; 1])"/>
    From: (2121, '/t/t[2]', '<t t-call="website.layout" additional_title="category.name" opt_wsale_filter_tags="is_view_active(\'website_sale.filter_products_tags\')" website_sale_sortable="website_sale_sortable" isSortingBy="[sort for sort in website_sale_sortable if sort[0]==request.params.get(\'order\', \'\')]"/>')
          (2121, '/t/t[2]', '<t t-call="website.layout" additional_title="category.name" opt_wsale_filter_tags="is_view_active(\'website_sale.filter_products_tags\')" website_sale_sortable="website_sale_sortable" isSortingBy="[sort for sort in website_sale_sortable if sort[0]==request.params.get(\'order\', \'\')]"/>')
          (2121, '/t/t[2]/div[1]/div[2]/div/aside/div/div[3]/div/t', '<t t-call="website_sale.product_attribute_filters_form" isMobile="False"/>')
          (2136, '/t/form/t[1]', '<t t-set="visible_attributes" t-value="len([a for a in attributes if pavs_per_attribute[a] and len(pavs_per_attribute[a]) &gt; 1])"/>')
```
</details>

lipiraux (changes requested):
🦆

addons/website_sale/controllers/main.py
livaios: It's needed to avoid getting the traceback when there's no products

addons/website_sale/controllers/main.py
livaios: Here we don't sort with the `attribute_id.id` but we sort by the default order of the `product.attribute` 
I know its confusing when everything has the `_id` suffix 😅

addons/website_sale/controllers/main.py
```diff
@@ -494,21 +495,21 @@ def shop(self, page=0, category=None, search="", min_price=0.0, max_price=0.0, t
...
 
         ProductAttribute = request.env["product.attribute"]
+        pavs_per_attribute = defaultdict(lambda: request.env["product.attribute.value"])
         if products:
-            # get all products without limit
-            attributes_grouped = request.env["product.template.attribute.line"]._read_group(
+            grouped_pavs = request.env["product.attribute.value"]._read_group(
                 domain=[
-                    ("product_tmpl_id", "in", search_product.ids),
+                    ("pav_attribute_line_ids.product_tmpl_id.id", "in", search_product.ids),
```
lipiraux: I noticed the trailing `.id` in the condition and investigated if the ORM treated it differently. Did you know that the ORM will add another subquery to filter-out template not in the current company?

Since `search_product` is already filtered by the current company, I guess we don't need the extra overhead. wdyt?


To compare, here are both queries:

<details>

<summary>`("pav_attribute_line_ids.product_tmpl_id.id", "in", search_product.ids)`</summary>

```sql
SELECT "product_attribute_value"."attribute_id",
    ARRAY_AGG(
        "product_attribute_value"."id"
        ORDER BY "product_attribute_value"."id"
    )
FROM "product_attribute_value"
    JOIN "product_attribute" AS "product_attribute_value__attribute_id" ON (
        "product_attribute_value"."attribute_id" = "product_attribute_value__attribute_id"."id"
    )
WHERE (
        "product_attribute_value"."active" IS TRUE
        AND "product_attribute_value"."attribute_id" IN (
            SELECT "product_attribute"."id"
            FROM "product_attribute"
            WHERE "product_attribute"."visibility" IN ('visible')
        )
        AND EXISTS (
            SELECT 1
            FROM "product_attribute_value_product_template_attribute_line_rel" AS "product_attribute_value__pav_attribute_line_ids"
            WHERE "product_attribute_value__pav_attribute_line_ids"."product_attribute_value_id" = "product_attribute_value"."id"
                AND "product_attribute_value__pav_attribute_line_ids"."product_template_attribute_line_id" IN (
                    SELECT "product_template_attribute_line"."id"
                    FROM "product_template_attribute_line"
                    WHERE (
                            "product_template_attribute_line"."active" IS TRUE
                            AND "product_template_attribute_line"."product_tmpl_id" IN (
                                SELECT "product_template"."id"
                                FROM "product_template"
                                WHERE "product_template"."id" IN (< search_product.ids >)
                                    AND (
                                        "product_template"."company_id" IN (1)
                                        OR "product_template"."company_id" IS NULL
                                    )
                            )
                        )
                )
        )
    )
GROUP BY "product_attribute_value"."attribute_id",
    "product_attribute_value__attribute_id"."sequence",
    "product_attribute_value__attribute_id"."id"
ORDER BY "product_attribute_value__attribute_id"."sequence",
    "product_attribute_value__attribute_id"."id";
```

</details>

<details>

<summary>`("pav_attribute_line_ids.product_tmpl_id", "in", search_product.ids)`</summary>

```sql
SELECT "product_attribute_value"."attribute_id",
    ARRAY_AGG(
        "product_attribute_value"."id"
        ORDER BY "product_attribute_value"."id"
    )
FROM "product_attribute_value"
    JOIN "product_attribute" AS "product_attribute_value__attribute_id" ON (
        "product_attribute_value"."attribute_id" = "product_attribute_value__attribute_id"."id"
    )
WHERE (
        "product_attribute_value"."active" IS TRUE
        AND "product_attribute_value"."attribute_id" IN (
            SELECT "product_attribute"."id"
            FROM "product_attribute"
            WHERE "product_attribute"."visibility" IN ('visible')
        )
        AND EXISTS (
            SELECT 1
            FROM "product_attribute_value_product_template_attribute_line_rel" AS "product_attribute_value__pav_attribute_line_ids"
            WHERE "product_attribute_value__pav_attribute_line_ids"."product_attribute_value_id" = "product_attribute_value"."id"
                AND "product_attribute_value__pav_attribute_line_ids"."product_template_attribute_line_id" IN (
                    SELECT "product_template_attribute_line"."id"
                    FROM "product_template_attribute_line"
                    WHERE (
                            "product_template_attribute_line"."active" IS TRUE
                            AND "product_template_attribute_line"."product_tmpl_id" IN (< search_product.ids >)
                        )
                )
        )
    )
GROUP BY "product_attribute_value"."attribute_id",
    "product_attribute_value__attribute_id"."sequence",
    "product_attribute_value__attribute_id"."id"
ORDER BY "product_attribute_value__attribute_id"."sequence",
    "product_attribute_value__attribute_id"."id";
```

</details>

addons/website_sale/controllers/main.py
lipiraux: oooh 😲, then never mind. Thanks for the explaination.

addons/website_sale/controllers/main.py
livaios: Yeah I wouldn't have guessed that, nice catch. 
Indeed its best if we avoid any extra overhead especially in the `/shop` page

lipiraux: Last comment, then I swear I stop 🫣. The ORM doesn't like records in the conditions.

https://github.com/odoo/odoo/blob/1518e26a31aadd950355252d8db74164a2d87ce2/odoo/orm/domains.py#L855-L857

livaios: Done

lipiraux (approved):
Thanks @livaios 😊

all good for me 👍

addons/website_sale/controllers/main.py
```diff
@@ -494,21 +495,21 @@ def shop(self, page=0, category=None, search="", min_price=0.0, max_price=0.0, t
         )
 
         ProductAttribute = request.env["product.attribute"]
+        pavs_per_attribute = defaultdict(lambda: request.env["product.attribute.value"])
```
ltinel: Nit: can't we simply initialize this as an empty dict?
```suggestion
        pavs_per_attribute = {}
```

addons/website_sale/controllers/main.py
```diff
@@ -494,21 +495,21 @@ def shop(self, page=0, category=None, search="", min_price=0.0, max_price=0.0, t
...
+                    ("pav_attribute_line_ids.product_tmpl_id", "in", search_product.ids),
                     ("attribute_id.visibility", "=", "visible"),
                 ],
                 groupby=["attribute_id"],
                 order="attribute_id",
+                aggregates=["id:recordset"],
             )
-            attribute_ids = [attribute.id for (attribute,) in attributes_grouped]
-            attributes = ProductAttribute.browse(attribute_ids)
+            pavs_per_attribute.update(grouped_pavs)
```
ltinel: Nit: I find the `update` call a bit confusing (since it's not really an "update"). I guess it would be fine to use a simple assignment instead?
```suggestion
            pavs_per_attribute = dict(grouped_pavs)
```

addons/website_sale/templates/shop_page_templates.xml
```diff
@@ -810,7 +810,7 @@
                 <input
                     type="text"
                     class="o_wsale_attribute_search_bar form-control"
-                    t-attf-placeholder="Search {{a.name}} ({{len(a.value_ids)}})"
+                    t-attf-placeholder="Search {{a.name}} ({{len(pavs_per_attribute[a])}})"
```
ltinel: Nit: this is fine, but I guess `pavs_per_attribute[a]` has the same length as `sorted_values`, so I'd be tempted  to use `sorted_values` instead (since it's more concise):
```suggestion
                    t-attf-placeholder="Search {{a.name}} ({{len(sorted_values)}})"
```

addons/website_sale/templates/shop_page_templates.xml
```diff
@@ -872,15 +872,15 @@
...
             <t
                 t-set="visible_attributes"
-                t-value="len([a for a in attributes if a.value_ids and len(a.value_ids) &gt; 1])"
+                t-value="len([a for a in attributes if pavs_per_attribute[a] and len(pavs_per_attribute[a]) &gt; 1])"
             />
 
             <t t-foreach="attributes" t-as="a">
                 <t t-set="_status" t-value="'inactive'"/>
-                <t t-foreach="a.value_ids" t-as="v" t-if="v.id in attrib_set" t-set="_status" t-value="'active'"/>
+                <t t-foreach="pavs_per_attribute[a]" t-as="v" t-if="v.id in attrib_set" t-set="_status" t-value="'active'"/>
```
ltinel: I'm hesitating whether it's worth extracting `pavs_per_attribute[a]` into a variable. Probably not, but could save a bit of space 😅

ltinel (changes requested):
Just a few nits, but looks great, thanks! 😄

ltinel (approved):
LGTM, thanks!
@robodoo delegate+

addons/website_sale/controllers/main.py
livaios: @lipiraux FYI I got this warning on runbot
> "Since 20.0, use concat() with only one iterable"

Changed it to `ProductAttribute.union(pavs_per_attribute.keys())`

livaios:
robodoo r+

addons/website_sale/controllers/main.py
lipiraux: union will try to remove redundant ids, wheras you know they are already unique since `.keys()` is a set 👀

lipiraux: ah, well too late 😅 doesn't really matter much TBH

livaios: We shouldn't have redundant ids since we're grouping by the key anyways so the idea is to have all the values per one attribute

PR #234674, opened 2025-11-06
[IMP] mail: revamp discuss settings

nees-odoo:
**Description of the issue this PR addresses:**
Revamp Discuss Settings

**Desired behaviour after PR is merged:**

This PR includes:
- Added enable/disable options to allow users to turn features on or off as needed.
- Updated help strings across settings for consistent and user-friendly phrasing.
- Added descriptive placeholders showing the expected API key format, reassuring 
that the user's input matches the required pattern.
- Given an option as a call server to use SFU/ICE servers for video calls.

Enterprise PR: https://github.com/odoo/enterprise/pull/98977

part of - [5227387](https://www.odoo.com/odoo/project/1519/tasks/5227387)

**Before**:
image

**After**:
image

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246154, opened 2026-01-28
[FIX] discuss: prevent blocking sfu env setting

ThanhDodeurOdoo:
Before this commit, since https://github.com/odoo/odoo/pull/234674 the `use_call_server`/`use_sfu_server` settings were required to use the sfu url from the env.

This commit restores the previous behavior of allowing the use of a SFU regardless of the setting if it is provided from the env.


seb-odoo (approved):
@robodoo r+ delegate+

PR #246155, opened 2026-01-28
[FIX] l10n_es: 2025 update for mod390

agr-odoo:
Currently report line formulas for box 33 and 34 are wrong, as they are not according to the regulation.
https://sede.agenciatributaria.gob.es/static_files/Sede/Procedimiento_ayuda/G412/Instrucciones_modelo_390-2025.pdf 
Box 33 is missing the sum of 27, 29, 649 and 31
Box 34 is missing the sum of 28, 30, 650 and 32

Updating mod390 report for 2025

opw-5498345


agr-odoo:
Hello @chklop 
It would be best to have a functional check here, as something else may need to be updated

PR #246158, opened 2026-01-28
[FIX] sale: integrate customer addresses setting in quote pdfs

barracudapps:
## Versions
17.0+
16.0 too but template change is not worth risking stable policy.

## Issue
There is a behavior mismatch between invoices and quotes/orders addresses display. Disabling "Customer Addresses" setting should only display one address on generated documents. Yet, it has no effect on quotes and orders while it has on invoices.

## Steps to reproduce
- Go to Settings:
  - Look for "Customer Addresses";
  - Ensure the box is checked (and save).
- Go to Sales / Orders / Customers:
  - Create a new customer and add addresses under the "Contacts & Addresses" tab:
    - Add one "Delivery Address";
    - Add one "Invoice Address" (with different values).
- Click the "Sales" smart button to create a new quote for that customer:
  - Add any product;
  - From the cog, click Print > PDF Quote.
  - Confirm the order (deliver if needed);
  - Create regular invoice;
  - Confirm the invoice;
  - From the cog, click Print > Invoices.
  - Both documents display a shipping address.

Start again with "Customer Addresses" setting disabled and watch the shipping address disappear from the invoice while still present on the quote document.
*N.B.: Same behavior with the preview*

## Fix
Mimic invoices' behavior in https://github.com/odoo/odoo/blob/1bd2d63eb80020e6e213c889e83cca2950d4dcd4/addons/account/views/report_invoice.xml#L12

opw-5400368

### DIFF
--- a/addons/sale/views/sale_portal_templates.xml
+++ b/addons/sale/views/sale_portal_templates.xml
@@ -458,7 +458,8 @@
                 <!-- ======  Customer Information  ====== -->
                 <div id="customer_info" class="col-12 col-lg-6 mb-4">
                     <h4 class="mb-1">
-                        <t t-if="sale_order.partner_shipping_id == sale_order.partner_invoice_id">
+                        <t t-if="not env.user.has_group('account.group_delivery_invoice_address')
+                                 or sale_order.partner_shipping_id == sale_order.partner_invoice_id">
                             Invoicing and Shipping Address
                         </t>
                         <t t-else="">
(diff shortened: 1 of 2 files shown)


addons/sale/views/sale_portal_templates.xml
```diff
@@ -458,7 +458,8 @@
                 <!-- ======  Customer Information  ====== -->
                 <div id="customer_info" class="col-12 col-lg-6 mb-4">
                     <h4 class="mb-1">
-                        <t t-if="sale_order.partner_shipping_id == sale_order.partner_invoice_id">
+                        <t t-if="not env.user.has_group('account.group_delivery_invoice_address')
+                                 or sale_order.partner_shipping_id == sale_order.partner_invoice_id">
```
Feyensv, line 462: Feature groups should be used with care on pages accessible to public or portal users, as groups are only implied on internal users unless configured differently.

Here: 
> group_sale_delivery_address = fields.Boolean("Customer Addresses", implied_group='account.group_delivery_invoice_address')

will only impact internal users.
So this change means that even if the feature is enabled, this won't ever be shown to public/portal users

(unless website_sale is installed as it's adds the implication to portal/public users)
>     group_delivery_invoice_address = fields.Boolean(string="Shipping Address", implied_group='account.group_delivery_invoice_address', group='base.group_portal,base.group_user,base.group_public')

Feyensv: Also, I don't think this is worth fixing in stable at all, it's only shown if there are different shipping and billing addresses (which matters IMO).

PR #246163, opened 2026-01-28
[FIX] stock_account: Ensure product stock is accurate for a given to date

MaximeNoirhomme229:
**Issue**
While performing stock valuation, when a `to_date` is selected, the resulting valuation may be incorrect depending on the user's timezone.

**Steps to reproduce**
- Set the user timezone to UTC+1
- Create a storable product with:
  - quantity: 10 (created today)
  - unit cost: 5
  - valuation method: AVCO
- Go to Accounting > Review > Inventory > Inventory Valuation
- Select today's date
- Click on "Ending Stock"

-> The total value and quantity in stock are 0 instead of respectively 50 and 10.

**Cause**
When selecting a date (e.g. 12/04), the `to_date` is initially set at 00:00 in the user's timezone. An attempt is then made to convert it to 23:59 to avoid excluding quantities created during that day: https://github.com/odoo/odoo/blob/87e176ad76c9d7b87cd622ae38a8b9a62813b1cb/addons/stock_account/models/product.py#L147-L150

However, this conversion is performed on a naive UTC datetime. For a user in UTC+1, this results in the following situation:
- 12/04 00:00 (user timezone) -> 11/04 23:00 UTC
- Converted to 23:59 UTC -> 12/04 00:59 in user timezone

As a consequence, most of the quantities created on 12/04 are excluded from the valuation.

This issue impacts AVCO (and FIFO as well), as `at_date` is used during cost computation: https://github.com/odoo/odoo/blob/87e176ad76c9d7b87cd622ae38a8b9a62813b1cb/addons/stock_account/models/product.py#L147-L150

In addition, `at_date` is added to the context, while `qty_available` relies on `to_date` instead: https://github.com/odoo/odoo/blob/87e176ad76c9d7b87cd622ae38a8b9a62813b1cb/addons/stock/models/product.py#L151-L153

This inconsistency leads to incorrect quantities and valuation results.

opw-5491743

### DIFF
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -1,7 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 from bisect import bisect
 from collections import defaultdict
-from datetime import datetime
+from datetime import date, datetime, time
 
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
@@ -179,6 +179,18 @@ class ProductProduct(models.Model):
     def _compute_value(self):
         company_id = self.env.company
         self.company_currency_id = company_id.currency_id
+        products = self._with_valuation_context()
+
+        at_date = self.env.context.get('to_date')
+        original_value = at_date
+        at_date = fields.Datetime.to_datetime(at_date)
+        if (isinstance(original_value, date) and not isinstance(original_value, datetime)) or \
+            (isinstance(original_value, str) and len(original_value) == 10):
+            at_date = datetime.combine(at_date.date(), time.max)
+
+        if at_date:
+            products = products.with_context(at_date=at_date, to_date=at_date)
+
         # PERF: Pre-compute:the sum of 'total_value' of lots per product in go
         std_price_by_company_id = {}
         total_value_by_company_id = {}
@@ -189,11 +201,8 @@ class ProductProduct(models.Model):
 
             products = self.with_company(company.id).with_context(allowed_company_ids=company.ids)
             products = products._with_valuation_context()
-
-            at_date = fields.Datetime.to_datetime(self.env.context.get('to_date'))
             if at_date:
-                at_date = at_date.replace(hour=23, minute=59, second=59)
-                products = products.with_context(at_date=at_date)
+                products = products.with_context(at_date=at_date, to_date=at_date)
 
             env = products.env
 
--- a/addons/stock_account/tests/test_stockvaluation.py
+++ b/addons/stock_account/tests/test_stockvaluation.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from datetime import timedelta
+from datetime import date, timedelta
 from freezegun import freeze_time
 
 from odoo import Command
@@ -2213,6 +2213,58 @@ class TestStockValuation(TestStockValuationCommon):
         report_value_2 = report_for_company_2.get_report_values(docids=product.ids)
         self.assertEqual(report_value_1['docs']['value'], "U 50.00")
         self.assertEqual(report_value_2['docs']['value'], "48.00 DD")
+
+    @freeze_time("2024-01-10 10:00:00")
+    def test_product_qty_and_value_correct_at_to_date_with_timezone(self):
+        """
+        Ensure that qty_available, free_qty, avg_cost, and total_value are computed
+        correctly for products at a historical to_date, taking the user's timezone into account.
+        """
+        self.env.user.tz = 'Europe/Paris'
+        to_date = date(2024, 1, 10)
+
+        lot_product = self.env['product.product'].create([
+            {
+                'name': 'Product LOT',
+                'tracking': 'lot',
+                'categ_id': self.category_avco.id,
+                'is_storable': True,
+                'standard_price': 10.0,
+                'lot_valuated': True,
+            },
+        ])
+        lot = self.env['stock.lot'].create({
+            'name': 'lot',
+            'product_id': lot_product.id,
+        })
+
+        products = self.product_standard | self.product_avco | self.product_fifo | lot_product
+        for product in products:
+            self._make_in_move(
+                product=product,
+                quantity=10.0,
+                location_id=self.supplier_location.id,
+                location_dest_id=self.stock_location.id,
+                lot_ids=lot if product.tracking == 'lot' else self.env['stock.lot'],
+            )
+
+        expected_values = [
+            {
+                    'qty_available': 10.0,
+                    'free_qty': 10.0,
+                    'avg_cost': 10.0,
+                    'total_value': 100.0,
+            }
+            for _ in products
+        ]
+        self.assertRecordValues(
+            products.with_context(to_date=to_date),
+            expected_values,
+        )
+        self.assertRecordValues(
+            products.with_context(to_date="2024-01-10"),
+            expected_values,
+        )
 
     def test_stock_report_avco_warehouse_dependency(self):
         """ Create two warehouses and check that the total value and the on hand quantity
--- a/addons/stock/models/product.py
+++ b/addons/stock/models/product.py
@@ -4,8 +4,8 @@ import operator as py_operator
 from ast import literal_eval
 from collections import defaultdict
 from collections.abc import Iterable
+from datetime import date, datetime, time
 from dateutil.relativedelta import relativedelta
-from datetime import datetime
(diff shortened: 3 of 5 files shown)


addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -2263,6 +2263,55 @@ def test_forecast_report_value(self):
...
+                'valuation': valuation,
+                'is_storable': True,
+                'tracking': tracking,
+                'standard_price': 5.0,
+                'categ_id': categ.id,
+            } for n, categ, valuation, tracking in [
+                ('AVCO', self.category_avco, 'real_time', 'none'),
+                ('FIFO', self.category_fifo, 'real_time', 'none'),
+                ('LOT', self.category_avco, 'real_time', 'lot'),
+            ]
```
lase-odoo: I think that when you do things like that it improves readability to order the variables in the usage order. Also, the valuation field is computed from the product category, so it does not need to be set. Finally, to me `n` in is the kind of variable name you chose to iterate, `name` looks a nicer variable name.
```suggestion
        products = self.env['product.product'].create([
            {
                'name': f'Product {name}',
                'tracking': tracking,
                'categ_id': categ.id,
                'is_storable': True,
                'standard_price': 5.0,
            } for name, tracking, categ, in [
                ('AVCO', 'none', self.category_avco,),
                ('FIFO', 'none', self.category_fifo,),
                ('LOT', 'lot',  self.category_avco),
            ]
        ])
```
Also, I am wondering if it would not be even more relevant to set the product valuated by lot as `lot_valuated` ?
In which case imo, it is fine creating the products based solely on the categ and the name and to write only on the LOT product to set its tracking and `lot_valuated` field to improve readability even further (even thought it comes at the cost of an additional post creation write).

I think it is quite relevant to adapt the lot product to be lot valuated since the value and qtis of these products are computed by different lines in the `_compute_values` that should imo be tested aswell:
https://github.com/odoo/odoo/blob/c5c7b3bc5be7075516c18d1e4fd06795277be9d9/addons/stock_account/models/product.py#L150-L172
Particularly since the `total_value` field of `stock.lot` is also context dependant on the `to_date'`:
https://github.com/odoo/odoo/blob/c5c7b3bc5be7075516c18d1e4fd06795277be9d9/addons/stock_account/models/stock_lot.py#L21-L24

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -2263,6 +2263,55 @@ def test_forecast_report_value(self):
...
+        })
+        self.env['stock.quant'].create([
+            {
+                'product_id': product.id,
+                'location_id': self.env.ref('stock.stock_location_stock').id,
+                'quantity': 10.0,
+                'lot_id': lot.id if product.tracking == 'lot' else None,
+            }
+            for product in products
+        ])
```
lase-odoo: As a general rule of thumb, I do not think it is a good idea to create quant's by hand in tests unless you really need a particular quant creation I would rather rely on the `_update_available_quantity` which btw has a nice `in_date` parameter which I guess is quite relevant to your use case. Even if it does not allow a creation in batch it mimics the "on hand quantity update" better than a simple quant creation and handles on its own the creation parameters (also, it populates the cache with some product qtis and hence might be relevant to check the values we are looking at are "indeed" context dependent and recomputed even if already present for a different context in the cache):
https://github.com/odoo/odoo/blob/c5c7b3bc5be7075516c18d1e4fd06795277be9d9/addons/stock/models/stock_quant.py#L1027-L1095

addons/stock_account/models/product.py
```diff
@@ -173,7 +176,12 @@ def _compute_value(self):
 
         product_ids_grouped_by_cost_method = defaultdict(set)
         ratio_by_product_id = {}
-        for product in products:
+        for product, product_wo_context in zip(products, self):
+            if product_wo_context.mapped(lambda p: (p.qty_available, p.free_qty)) != product.mapped(lambda p: (p.qty_available, p.free_qty)):
+                product_wo_context.with_context(skip_qty_available_update=True).write({
+                    'qty_available': product.qty_available,
+                    'free_qty': product.free_qty,
+                })
```
lase-odoo: Houla, this code looks like forbidden dark magic.
Image

To begin, the present method is a the compute method of the `avg_cost`, `total_value` and `company_currency_id` fields not for any of the quantity fields. It should therefore not change the value of any of these other computed fields particularly since the compute of these quantity fields would provide a different value for the current context.

In addition on a technical standpoint standpoint there are a few things to say:
- The `qty_available` and `free_qty` fields are floats and should never be compared with `=` or `!=`.
- You should not call the `write` method of base models directly in such tricky use case unless you have a good reason to do so. The assignment:
```py
product_wo_context.with_context(skip_qty_available_update=True).qty_available = product.qty_available
```
will rely on the `write` of the field, but the write of fields does not always rely the `write` of the model see:
https://github.com/odoo/odoo/blob/c5c7b3bc5be7075516c18d1e4fd06795277be9d9/odoo/orm/fields.py#L1809-L1845
(you only write on the model if you end up in the `other_ids` use case.

lase-odoo (commented):
Thanks for the proposition and sorry for the delay on this review ! 🌊

The state of the art is actually quite tricky. One of the main issue being that changing the `at_date` value to use a different one that the `to_date` used in other computes such as the `qty_available` will create numerous miss-matches and lead to an lead to inconveniences in views where the qty_available is computed and displayed for the `to_date` value bu the `_compute_value` provides the value for the end of the day. 

But since all the problem comes form these lines:
https://github.com/odoo/odoo/blob/b071bc3cb9e1d91d9d0fbce7961c3c75f28ef874/addons/stock_account/models/product.py#L145-L148
I discussed with the stock dev team to understand the main purpose of these and, it was only to generate the accounting data's for the end of the day rather than the precise precise hour to which you create the invoice,...

In particular, while it is true that the the timezone conversion needs to be done, regarding the context discrepancy I think that the solution would just be to move the context change to 23:59:59 outside of the `_compute_value` and rather update the datetime to 23:59:59 when it is needed for accounting purposes in order to keep using a uniform `to_date` and `at_date` in views and simultaneous computations.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -2267,6 +2267,54 @@ def test_forecast_report_value(self):
...
+                ('FIFO', self.category_fifo, 'real_time', 'none'),
+                ('LOT', self.category_avco, 'real_time', 'lot'),
+            ]
+        ])
+        lot = self.env['stock.lot'].create({
+            'product_id': products[2].id,
+            'name': 'LOT-001',
+        })
+        for product in products:
+            self._make_in_move(product=product, quantity=10.0, location_id=self.supplier_location.id, location_dest_id=self.stock_location.id, lot_ids=[lot] if product.tracking == 'lot' else [])
```
lase-odoo:
```suggestion
            self._make_in_move(product=product, quantity=10.0, location_id=self.supplier_location.id, location_dest_id=self.stock_location.id, lot_ids=lot if product.tracking == 'lot' else self.env['stock.lot'])
```

lase-odoo (commented):
nitpick but work with record sets when possible not with unexpected iterables

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -2267,6 +2267,54 @@ def test_forecast_report_value(self):
...
+                'categ_id': categ.id,
+                'valuation': valuation,
+                'tracking': tracking,
+                'is_storable': True,
+                'standard_price': 5.0,
+            } for name, categ, valuation, tracking in [
+                ('AVCO', self.category_avco, 'real_time', 'none'),
+                ('FIFO', self.category_fifo, 'real_time', 'none'),
+                ('LOT', self.category_avco, 'real_time', 'lot'),
+            ]
```
Whenrow: no need to create new product. you can use 
self.product_standard
self.product_avco
self.product_fifo

I don't see the use of lot neither in this case 🤔

MaximeNoirhomme229: Thanks for the review @Whenrow, you are right, my bad! 
And for the lot, I don't remember to be honest and the code could have change since I did it. But anyway, it does not seem relevant (anymore) to test it so I remove it.

lase-odoo: @Whenrow @MaximeNoirhomme229 I suggested in my previous review to have a product valuated by lot eg. tracked by lot and `lot_valuated`, since the valuation of these products is quite different: https://github.com/odoo/odoo/pull/246163#discussion_r2851587665

PR #246181, opened 2026-01-28
[FIX] account_peppol: neutralize should be consistent

gawa-odoo:
Currently, the proxy_client_user is neutralized as demo. But the edi_mode is set to test.
It's confusing for the users, and we should be consistent.

---


clbr-odoo (approved):
@robodoo r+

PR #210189, opened 2025-05-15
[REF] mail, hr_{holidays,homeworking}: refactored im_status overrides

Atovange:
enterprise PR: https://github.com/odoo/enterprise/pull/85748

This commit improves how we apply overrides to the im_status without
mutating its original value.

Previously, customizing `im_status` required changing the status
string itself (for example, “online” would become “leave_online” or
“home_online”). This approach made it cumbersome for the front end to
interpret and display the correct status.

Now, instead of altering the `im_status` string, we keep it intact
and introduce a separate data payload to indicate any special
conditions (such as “on leave” or “working from home”). We register
these overrides in the `im-status-data` registry, which contains the
metadata needed for each condition like title, icon, and aria-label.
At render time, the front end checks this registry and applies the
appropriate override (for instance, switching the icon to fa-plane
when the user is on leave) without ever modifying the original status
string.

PR #246182, opened 2026-01-28
[IMP] mail, *: Proper mask of IM status on Avatars

alexkuhn:
*: hr_holidays, im_livechat

Before this commit, IM status and thread icon were used a lot as bottom-right floating icon on a conversation or user avatar. Each context that wanted this visual of avatar + floating icon had to craft by hand the avatar sizing and icon position, which has many drawbacks:

- image and icon sizing and positioning needs manual tweak in each context, requiring lots of effort.
- very-prone to inconsistencies due to high maintenance to make each context work. Current state had many inconsistencies.
- handling of mask of floating icon is poor, with `.bg-inherit` on parented chaining that is hard to enforce and in some cases, just don't work at all

This commit improves by introducing a new component `DiscussAvatar`, that accepts `channel`, `thread`, channel's `member` or `persona` to display avatar and conversation / person icon in the bottom-right corner. This new component has the following benefits:

- one component to handle all the cases of conversation / person avatar with floating icon / im status, drastically reducing maintenance effort and increasing significantly UI consistency.
- new technique to mask image with floating icon that works all the time thanks to the use of `SVG`'s mask.

Part of Task-5867464

https://github.com/odoo/enterprise/pull/105814

Before / After
Screenshot 2026-01-30 at 18 41 29
Screenshot 2026-01-30 at 18 49 51

Improvements to Discuss command:
Screenshot 2026-01-30 at 18 53 22

To user mentions:
Screenshot 2026-01-30 at 18 53 47

+ typing icon has been slightly improved.

### DIFF
--- a/addons/hr_holidays/static/src/im_status_patch.xml
+++ b/addons/hr_holidays/static/src/im_status_patch.xml
@@ -2,10 +2,10 @@
 <templates xml:space="preserve">
     <t t-inherit="mail.ImStatus" t-inherit-mode="extension">
         <xpath expr="//*[@name='icon']" position="replace">
-            <i t-if="persona.im_status === 'leave_online'" class="fa fa-plane text-success" title="On Leave (Online)" role="img" aria-label="User is on leave and online"/>
-            <i t-elif="persona.im_status === 'leave_away'" class="fa fa-plane o-yellow" title="On Leave (Idle)" role="img" aria-label="User is on leave and idle"/>
-            <i t-elif="persona.im_status === 'leave_busy'" class="fa fa-fw fa-plane text-danger" title="On Leave (Busy)" role="img" aria-label="User is on leave and busy"/>
-            <i t-elif="persona.im_status === 'leave_offline'" class="fa fa-plane text-500" title="On Leave" role="img" aria-label="User is on leave"/>
+            <i t-if="persona.im_status === 'leave_online'" t-att="attr" class="fa fa-plane text-success" title="On Leave (Online)" role="img" aria-label="User is on leave and online"/>
+            <i t-elif="persona.im_status === 'leave_away'" t-att="attr" class="fa fa-plane o-yellow" title="On Leave (Idle)" role="img" aria-label="User is on leave and idle"/>
+            <i t-elif="persona.im_status === 'leave_busy'" t-att="attr" class="fa fa-fw fa-plane text-danger" title="On Leave (Busy)" role="img" aria-label="User is on leave and busy"/>
+            <i t-elif="persona.im_status === 'leave_offline'" t-att="attr" class="fa fa-plane text-500" title="On Leave" role="img" aria-label="User is on leave"/>
             <t t-else="">$0</t>
         </xpath>
     </t>
--- a/addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
+++ b/addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
@@ -127,5 +127,8 @@ const discussChannelPatch = {
             await super.leaveChannel(...arguments);
         }
     },
+    showThreadIcon(...args) {
+        return this.channel_type === "livechat" || super.showThreadIcon(...args);
+    },
 };
 patch(DiscussChannel.prototype, discussChannelPatch);
--- a/addons/mail/static/src/core/common/chat_bubble.xml
+++ b/addons/mail/static/src/core/common/chat_bubble.xml
@@ -5,10 +5,9 @@
             <span class="o-mail-ChatBubble-unreadIndicator position-absolute" t-att-class="{ 'opacity-0': !channel.isUnread or channel.importantCounter }"><i class="fa fa-circle"/></span>
             <div t-if="channel.importantCounter > 0" class="o-mail-ChatBubble-counter position-absolute badge rounded-pill fw-bold o-discuss-badge shadow" t-out="channel.importantCounter"/>
             <button t-if="!env.embedLivechat and !isMobileOS" class="o-mail-ChatBubble-close position-absolute shadow rounded-circle fw-bold bg-view" title="Close Chat Bubble" t-on-click.stop="() => this.props.chatWindow.requestClose()"><i class="oi oi-close"/></button>
-            <ThreadIcon t-if="showThreadIcon" thread="channel.thread" size="'large'" className="'o-mail-ChatBubble-status position-absolute'"/>
             <CountryFlag t-if="channel.showCorrespondentCountry" country="channel.correspondentCountry" class="'o-mail-ChatBubble-country position-absolute bottom-0 border shadow-sm'"/>
             <button class="o-mail-ChatHub-bubbleBtn btn shadow">
-                <img class="o-mail-ChatBubble-avatar bg-view rounded-circle object-fit-cover" t-att-class="{ 'o-big': env.embedLivechat }" t-att-src="channel.avatarUrl" alt="Channel image" draggable="false"/>
+                <DiscussAvatar channel="channel" size="42" imgRoundedClass="'rounded-circle'"/>
             </button>
         </div>
     </t>
--- a/addons/mail/static/src/core/common/discuss_avatar.js
+++ b/addons/mail/static/src/core/common/discuss_avatar.js
@@ -1,0 +1,58 @@
+import { ImStatus } from "@mail/core/common/im_status";
+import { ThreadIcon } from "@mail/core/common/thread_icon";
+
+import { Component } from "@odoo/owl";
+
+let nextId = 0;
+
+export class DiscussAvatar extends Component {
+    static template = "mail.DiscussAvatar";
+    static props = [
+        "channel?",
+        "member?",
+        "persona?",
+        "thread?",
+        "size?",
+        "typing?",
+        "className?",
+        "imgRoundedClass?",
+    ];
+    static defaultProps = { className: "", size: 32, typing: true };
+    static components = { ImStatus, ThreadIcon };
+
+    setup() {
+        super.setup();
+        this.uniqueId = `mail.DiscussAvatar.${nextId++}`;
+    }
+
+    get channel() {
+        return this.props.channel ?? this.props.thread?.channel;
+    }
+
+    get thread() {
+        return this.props.thread ?? this.props.channel?.thread;
+    }
+
+    get isTyping() {
+        if (!this.props.typing) {
+            return false;
+        }
+        if (this.channel) {
+            return this.channel.hasOtherMembersTyping;
+        }
+        if (this.props.member) {
+            return this.props.member.isTyping;
+        }
+        return false;
+    }
+
+    get showIcon() {
+        if (this.channel) {
+            return this.channel.showThreadIcon({ ignoreTyping: !this.props.typing });
+        }
+        if (this.props.member || this.props.persona) {
+            return true;
+        }
+        return false;
+    }
+}
--- a/addons/mail/static/src/core/common/discuss_avatar.xml
+++ b/addons/mail/static/src/core/common/discuss_avatar.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+    <t t-name="mail.DiscussAvatar">
+        <svg t-att-width="props.size" t-att-height="props.size" viewBox="0 0 32 32" class="overflow-visible">
+            <mask t-att-id="uniqueId" width="32" height="32">
+                <rect x="0" y="0" height="32" width="32" fill="white"/>
+                <rect t-if="showIcon" color="black" t-att-x="isTyping ? 15 : 19" y="19" t-att-width="isTyping ? 24 : 16" t-att-height="isTyping ? 15 : 16" rx="8" ry="8"/>
+            </mask>
+            <foreignObject x="0" y="0" width="32" height="32" t-att-mask="'url(#' + uniqueId + ')'">
+                <t t-call="mail.DiscussAvatar.avatarImg"/> <!-- FIXME: owl doesn't like and load immediate <img> below <foreignObject>, hence t-call shenanigans -->
+            </foreignObject>
+            <g t-if="showIcon" t-att-transform="'translate(' + (isTyping ? 16 : 21) + ', 20)'">
+                <svg width="25" height="15" viewBox="0 0 25 15">
+                    <foreignObject x="0" y="0" width="25" height="15">
+                        <ThreadIcon t-if="thread" thread="thread" typing="props.typing"/>
+                        <ImStatus t-elif="props.member" member="props.member" className="'d-flex'" size="'md'" typing="props.typing"/>
+                        <ImStatus t-elif="props.persona" persona="props.persona" typing="props.typing"/>
+                    </foreignObject>
+                </svg>
+            </g>
+        </svg>
+    </t>
+
+    <t t-name="mail.DiscussAvatar.avatarImg">
+        <img t-if="thread" class="w-100 h-100 object-fit-cover" t-att-class="{ [props.imgRoundedClass]: props.imgRoundedClass, 'rounded-2': !props.imgRoundedClass }" t-att-src="thread.channel?.parent_channel_id?.avatarUrl ?? thread.channel?.avatarUrl ?? thread.avatarUrl" alt="Thread Image"/>
+        <img t-elif="props.member" alt="" class="w-100 h-100 object-fit-cover" t-att-class="{ [props.imgRoundedClass]: props.imgRoundedClass, 'rounded-3': !props.imgRoundedClass }" t-att-src="props.member.avatarUrl"/>
+        <img t-elif="props.persona" alt="" class="w-100 h-100 object-fit-cover" t-att-class="{ [props.imgRoundedClass]: props.imgRoundedClass, 'rounded': !props.imgRoundedClass }" t-att-src="props.persona.avatarUrl"/>
+    </t>
+</templates>
--- a/addons/mail/static/src/core/common/im_status.xml
+++ b/addons/mail/static/src/core/common/im_status.xml
@@ -2,34 +2,22 @@
 <templates xml:space="preserve">
 
     <t t-name="mail.ImStatus">
-        <div class="o-mail-ImStatus d-flex justify-content-center flex-shrink-0 align-items-center bg-white bg-opacity-100" t-attf-class="{{ props.className }}" t-att-style="props.style" t-att-class="{
-            'o-md': props.size === 'md' or props.size === 'medium',
-            'o-sm': props.size === 'sm' or props.size === 'small',
-            'rounded-circle': !showTypingIndicator,
-            'rounded-pill': showTypingIndicator,
-        }">
-            <span class="d-flex flex-column" t-att-class="{
-                'smaller': props.size === 'md' or props.size === 'medium',
-                'o-xsmaller': props.size === 'sm' or props.size === 'small',
-            }">
-                <t t-slot="pre_icon"/>
-                <t name="icon">
-                    <t t-if="(!props.member or !showTypingIndicator) and persona">
-                        <i t-if="persona.im_status === 'online'" class="fa fa-circle text-success" title="Online" role="img" aria-label="User is online"/>
-                        <i t-elif="persona.im_status === 'away'" class="fa fa-adjust o-yellow" title="Idle" role="img" aria-label="User is idle"/>
-                        <i t-elif="persona.im_status === 'busy'" class="fa fa-minus-circle text-danger" title="Busy" role="img" aria-label="User is busy"/>
-                        <i t-elif="persona.im_status === 'offline'" class="fa fa-circle-o text-700 opacity-75" title="Offline" role="img" aria-label="User is offline"/>
-                        <i t-elif="persona.im_status === 'bot'" class="fa fa-heart text-success" title="Bot" role="img" aria-label="User is a bot"/>
-                        <i t-else="" class="fa fa-question-circle opacity-75" title="No IM status available"/>
-                    </t>
-                    <div t-if="showTypingIndicator" class="rounded-pill" style="padding: 1px;" t-att-class="{ 'bg-300': env.inChatBubble, 'bg-inherit': !env.inChatBubble }">
-                        <div class="bg-success rounded-pill" style="padding: 3px;">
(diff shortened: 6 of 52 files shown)


addons/mail/static/src/core/common/discuss_avatar.js
```diff
@@ -0,0 +1,60 @@
...
+
+let nextId = 0;
+
+export class DiscussAvatar extends Component {
+    static template = "mail.DiscussAvatar";
+    static props = [
+        "channel?",
+        "member?",
+        "persona?",
+        "thread?",
```
seb-odoo, line 14: I'd like to get rid of the `thread` props, it's almost never used and confusing with the extra possibility of channel.

Some places even do `channel="props.thread.channel"` instead of passing `thread`.

I believe the only exception is the one in `DiscussContent-header` for mailbox. Mailboxes are a dummy case where we just need the mailbox icon and nothing else, so I'd rather see a separate component (or just the icon directly) for them.

seb-odoo (commented):
Nice to finally use `svg` to resolve this issue.

addons/mail/static/src/core/common/discuss_avatar.js
alexkuhn: Using the same component has the advantage to define the sizing once for all cases.
I can make a branching for mailboxes, but this can be weak against new threads, like ongoing PR that adds filters on "Inbox" by means of defining new special threads.

seb-odoo: If we keep it, I would either:
- remove channel and always use thread (so it's properly generic)
- rename thread to mailbox, keep channel (so it's always specific)

alexkuhn: I'd prefer keeping `channel` and `thread`. This allows to not think and just provide the business object we have at hand. Besides there's no mailbox model.

alexkuhn: There's lot of occurrences of code with `channel`, so this made sense to provide `channel`. `thread` is necessary for discuss content header.

addons/mail/static/src/discuss/core/public_web/discuss_command_palette.xml
```diff
@@ -5,9 +5,9 @@
         <div class="o-mail-DiscussCommand o_command_default d-flex align-items-center ps-3 pe-4" t-att-class="{ 'o-uiSmall': ui.isSmall }">
             <i t-if="props.channel" class="fa-fw opacity-75 text-muted me-2" t-att-class="props.channel.parent_channel_id ? 'fa fa-comments-o' : props.channel.discussAppCategory?.icon ?? 'oi oi-user opacity-0'"/>
             <i t-elif="props.persona" class="oi fa-fw oi-user opacity-75 text-muted me-2"/>
-            <img class="o-mail-DiscussCommand-img rounded-3 me-2 object-fit-cover shadow-sm" t-if="props.imgUrl" t-att-src="props.imgUrl" t-att-class="{ 'o-uiSmall': ui.isSmall }"/>
-            <ImStatus t-if="props.persona" className="'me-1'" persona="props.persona"/>
-            <span class="o-mail-DiscussCommand-nameContainer pe-1 text-ellipsis d-flex align-items-center fw-bold" t-att-class="{ 'o-action': props.action }">
+            <DiscussAvatar t-if="props.channel" channel="props.channel" size="ui.isSmall ? 32 : 26"/>
+            <DiscussAvatar t-if="props.persona" persona="props.persona" size="ui.isSmall ? 32 : 26"/>
```
seb-odoo: If they are mutually exclusive we don't need 2 lines, the non-set one will be always undefined.
Maybe just a t-if with `or` to avoid having it if both of them are undefined?

addons/mail/static/src/discuss/core/web/burger_menu_patch.xml
```diff
@@ -2,10 +2,7 @@
 <templates xml:space="preserve">  
     <t t-inherit="web.BurgerMenu" t-inherit-mode="extension">
         <xpath expr="//img[hasclass('o_burger_menu_avatar')]" position="replace">
-            <div class="position-relative d-flex flex-shrink-0 bg-inherit">
-                <t>$0</t>
-                <ImStatus className="'o-web-UserMenu-imStatus position-absolute'" persona="store.self" size="'md'"/>
-            </div>
+            <DiscussAvatar persona="store.self" size="26"/>
```
seb-odoo, line 5: This override is a bit strange (replacing the full parent just to "add" im status), it is a hint that we probably would like a generic `Avatar` in web instead, and `mail` to override it to support extra parameters (for thread) or add extra behavior (im status for user).

seb-odoo: Just to share the note, it doesn't necessarily call for a change right now.

addons/mail/static/src/discuss/core/public_web/discuss_command_palette.xml
alexkuhn: Turned into single line.

addons/hr_holidays/static/src/im_status_patch.xml
```diff
@@ -2,10 +2,10 @@
 <templates xml:space="preserve">
     <t t-inherit="mail.ImStatus" t-inherit-mode="extension">
         <xpath expr="//*[@name='icon']" position="replace">
-            <i t-if="persona.im_status === 'leave_online'" class="fa fa-plane text-success" title="On Leave (Online)" role="img" aria-label="User is on leave and online"/>
-            <i t-elif="persona.im_status === 'leave_away'" class="fa fa-plane o-yellow" title="On Leave (Idle)" role="img" aria-label="User is on leave and idle"/>
-            <i t-elif="persona.im_status === 'leave_busy'" class="fa fa-fw fa-plane text-danger" title="On Leave (Busy)" role="img" aria-label="User is on leave and busy"/>
-            <i t-elif="persona.im_status === 'leave_offline'" class="fa fa-plane text-500" title="On Leave" role="img" aria-label="User is on leave"/>
+            <i t-if="persona.im_status === 'leave_online'" t-att="attr" class="fa fa-plane text-success" title="On Leave (Online)" role="img" aria-label="User is on leave and online"/>
```
seb-odoo, line 5: I don't see where this `attr` is defined? Is it dead code?

If it's not, the name might be a bit too generic and make it hard to find too.

addons/mail/static/src/core/common/im_status.xml
```diff
@@ -2,34 +2,28 @@
...
+        <t name="icon">
+            <t t-if="(!props.member or !showTypingIndicator) and persona">
+                <i t-if="persona.im_status === 'online'" class="fa fa-circle text-success" t-att="attr" title="Online" role="img" aria-label="User is online"/>
+                <i t-elif="persona.im_status === 'away'" class="fa fa-adjust o-yellow" t-att="attr" title="Idle" role="img" aria-label="User is idle"/>
+                <i t-elif="persona.im_status === 'busy'" class="fa fa-minus-circle text-danger" t-att="attr" title="Busy" role="img" aria-label="User is busy"/>
+                <i t-elif="persona.im_status === 'offline'" class="fa fa-circle-o text-700 opacity-75" t-att="attr" title="Offline" role="img" aria-label="User is offline"/>
+                <i t-elif="persona.im_status === 'bot'" class="fa fa-heart text-success o-xsmaller o-pt-0_5" t-att="attr" title="Bot" role="img" aria-label="User is a bot"/>
+                <i t-else="" class="fa fa-question-circle opacity-75" t-att="attr" title="No IM status available"/>
+            </t>
+            <div t-if="showTypingIndicator" class="rounded-pill" t-att="attr" style="padding: 1px;">
```
seb-odoo: So the icon is now on a `div` in that case? Is there a reason? It's not a pattern we have often.

seb-odoo: style should be in css files

addons/mail/static/src/core/common/im_status.xml
```diff
@@ -2,34 +2,28 @@
...
+                <div class="bg-success rounded-pill" style="padding: 2px;">
```
seb-odoo: style should be in css files

addons/mail/static/src/discuss/typing/common/thread_icon_patch.xml
```diff
@@ -2,17 +2,17 @@
...
-            <t t-if="channel.hasOtherMembersTyping" t-call="mail.ThreadIcon.typing"/>
+            <t t-if="channel.hasOtherMembersTyping and props.typing" t-call="mail.ThreadIcon.typing"/>
             <t t-else="">$0</t>
         </xpath>
     </t>
     <t t-name="mail.ThreadIcon.typing">
-        <div t-if="channel.hasOtherMembersTyping" class="bg-inherit rounded-pill" style="padding: 1px;">
-            <div class="bg-success rounded-pill" style="padding: 3px;">
+        <div class="rounded-pill" t-att="attr" style="padding: 1px;">
+            <div class="bg-success rounded-pill" style="padding: 2px;">
```
seb-odoo: style should be in css files

addons/mail/static/src/discuss/core/public_web/discuss_command_palette.xml
```diff
@@ -5,9 +5,8 @@
         <div class="o-mail-DiscussCommand o_command_default d-flex align-items-center ps-3 pe-4" t-att-class="{ 'o-uiSmall': ui.isSmall }">
             <i t-if="props.channel" class="fa-fw opacity-75 text-muted me-2" t-att-class="props.channel.parent_channel_id ? 'fa fa-comments-o' : props.channel.discussAppCategory?.icon ?? 'oi oi-user opacity-0'"/>
             <i t-elif="props.persona" class="oi fa-fw oi-user opacity-75 text-muted me-2"/>
-            <img class="o-mail-DiscussCommand-img rounded-3 me-2 object-fit-cover shadow-sm" t-if="props.imgUrl" t-att-src="props.imgUrl" t-att-class="{ 'o-uiSmall': ui.isSmall }"/>
-            <ImStatus t-if="props.persona" className="'me-1'" persona="props.persona"/>
-            <span class="o-mail-DiscussCommand-nameContainer pe-1 text-ellipsis d-flex align-items-center fw-bold" t-att-class="{ 'o-action': props.action }">
+            <DiscussAvatar t-if="props.channel or props.persona" channel="props.channel" persona="props.persona" size="ui.isSmall ? 32 : 26"/>
+            <span class="o-mail-DiscussCommand-nameContainer ps-2 pe-1 text-ellipsis d-flex align-items-center fw-bold" t-att-class="{ 'o-action': props.action }">
```
seb-odoo: I don't think transforming the `me-2` on `ps-2` is correct, or at least it's not an equivalent change, as if there is no icon the spacing will now be there, and it wasn't before.

addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
```diff
@@ -127,5 +127,8 @@ const discussChannelPatch = {
             await super.leaveChannel(...arguments);
         }
     },
+    showThreadIcon(...args) {
+        return this.channel_type === "livechat" || super.showThreadIcon(...args);
+    },
```
seb-odoo, line 132: It's a rare occurrence where it feels like writing when it should be hidden instead would be much easier.

addons/mail/static/src/core/common/im_status.xml
```diff
@@ -2,34 +2,28 @@
...
-                    <div t-if="showTypingIndicator" class="rounded-pill" style="padding: 1px;" t-att-class="{ 'bg-300': env.inChatBubble, 'bg-inherit': !env.inChatBubble }">
-                        <div class="bg-success rounded-pill" style="padding: 3px;">
-                            <Typing member="props.member" size="['sm', 'smaller'].includes(props.size) ? 'sm' : 'md'" displayText="false" />
-                        </div>
-                    </div>
-                </t>
-            </span>
-        </div>
+        })"/>
+        <t t-set="attr" t-value="{ 'class': class, 'style': props.style }"/>
```
seb-odoo: Here it is, I would suggest to move these as getter on the component as they are used in other templates (overrides), as it becomes really hard to track otherwise. It feels like "local" variables should be considered really local.

addons/mail/static/src/core/common/chat_bubble.xml
```diff
@@ -5,10 +5,9 @@
...
             <div t-if="channel.importantCounter > 0" class="o-mail-ChatBubble-counter position-absolute badge rounded-pill fw-bold o-discuss-badge shadow" t-out="channel.importantCounter"/>
             <button t-if="!env.embedLivechat and !isMobileOS" class="o-mail-ChatBubble-close position-absolute shadow rounded-circle fw-bold bg-view" title="Close Chat Bubble" t-on-click.stop="() => this.props.chatWindow.requestClose()"><i class="oi oi-close"/></button>
-            <ThreadIcon t-if="showThreadIcon" thread="channel.thread" size="'large'" className="'o-mail-ChatBubble-status position-absolute'"/>
             <CountryFlag t-if="channel.showCorrespondentCountry" country="channel.correspondentCountry" class="'o-mail-ChatBubble-country position-absolute bottom-0 border shadow-sm'"/>
             <button class="o-mail-ChatHub-bubbleBtn btn shadow">
-                <img class="o-mail-ChatBubble-avatar bg-view rounded-circle object-fit-cover" t-att-class="{ 'o-big': env.embedLivechat }" t-att-src="channel.avatarUrl" alt="Channel image" draggable="false"/>
+                <DiscussAvatar channel="channel" size="42" imgRoundedClass="'rounded-circle'"/>
```
seb-odoo, line 10: `button` elements are limited to have [Phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#phrasing_content), I believe `DiscussAvatar` can generate `div` which is incorrect in this context.

addons/mail/static/src/core/common/discuss_avatar.js
seb-odoo: Ok for now but if we follow that principle `channel="props.thread.channel"` should pass `thread="props.thread"`

addons/mail/static/src/core/common/im_status.xml
```diff
@@ -2,34 +2,28 @@
...
+        <t t-slot="pre_icon"/>
+        <t name="icon">
+            <t t-if="(!props.member or !showTypingIndicator) and persona">
+                <i t-if="persona.im_status === 'online'" class="fa fa-circle text-success" t-att="attr" title="Online" role="img" aria-label="User is online"/>
+                <i t-elif="persona.im_status === 'away'" class="fa fa-adjust o-yellow" t-att="attr" title="Idle" role="img" aria-label="User is idle"/>
+                <i t-elif="persona.im_status === 'busy'" class="fa fa-minus-circle text-danger" t-att="attr" title="Busy" role="img" aria-label="User is busy"/>
+                <i t-elif="persona.im_status === 'offline'" class="fa fa-circle-o text-700 opacity-75" t-att="attr" title="Offline" role="img" aria-label="User is offline"/>
+                <i t-elif="persona.im_status === 'bot'" class="fa fa-heart text-success o-xsmaller o-pt-0_5" t-att="attr" title="Bot" role="img" aria-label="User is a bot"/>
+                <i t-else="" class="fa fa-question-circle opacity-75" t-att="attr" title="No IM status available"/>
```
seb-odoo, line 13: I'd like to collapse these eventually into a single line, providing variables for the parts that are actually different.

addons/hr_holidays/static/src/im_status_patch.xml
alexkuhn: This is defined in original template.

addons/mail/static/src/core/public_web/notification_item.xml
```diff
@@ -13,8 +13,10 @@
             'o-px-2_5 o-py-1_5 gap-2': !ui.isSmall,
             'o-active': props.isActive,
         }">
-            <div class="o-mail-NotificationItem-avatarContainer position-relative bg-inherit flex-shrink-0 align-self-start" t-att-class="{ 'o-small': ui.isSmall }">
-                <img t-if="props.iconSrc" class="o_avatar w-100 h-100" t-att-class="{ 'rounded-3': ui.isSmall, 'rounded': !ui.isSmall }" alt="Notification Item Image" t-att-src="props.iconSrc"/>
+            <div class="o-mail-NotificationItem-avatarContainer position-relative flex-shrink-0 align-self-start" t-att-class="{ 'o-small': ui.isSmall }">
```
seb-odoo, line 16: same remark about non allowed content inside button

addons/mail/static/src/core/common/im_status.xml
alexkuhn: This was already the case before this commit. The "is typing" icon is a component with 3 `span` that represent the dots of the "...".

addons/mail/static/src/core/common/discuss_avatar.js
```diff
@@ -0,0 +1,60 @@
...
+        if (this.channel) {
+            return this.channel.hasOtherMembersTyping;
+        }
+        if (this.props.member) {
+            return this.props.member.isTyping;
+        }
+        if (this.props.persona) {
+            return false;
+        }
+        return false;
```
seb-odoo: useless last case

addons/mail/static/src/core/common/discuss_avatar.xml
```diff
@@ -0,0 +1,29 @@
...
+            <mask t-att-id="uniqueId" width="32" height="32">
+                <rect x="0" y="0" height="32" width="32" fill="white"/>
+                <rect t-if="showIcon" color="black" t-att-x="isTyping ? 15 : 19" y="19" t-att-width="isTyping ? 24 : 16" t-att-height="isTyping ? 15 : 16" rx="8" ry="8"/>
+            </mask>
+            <foreignObject x="0" y="0" width="32" height="32" t-att-mask="'url(#' + uniqueId + ')'">
+                <t t-call="mail.DiscussAvatar.avatarImg"/> <!-- FIXME: owl doesn't like and load immediate <img> below <foreignObject>, hence t-call shenanigans -->
+            </foreignObject>
+            <g t-if="showIcon" t-att-transform="'translate(' + (isTyping ? 16 : 21) + ', 20)'">
+                <svg width="25" height="15" viewBox="0 0 25 15">
+                    <foreignObject x="0" y="0" width="25" height="15">
```
seb-odoo, line 14: We might want variables and doing math for the various numbers rather than hard-coding the values?
While faster and technically equivalent, it makes it harder to see which number are in relation to each other and why they have the value they have.

seb-odoo (approved):
When all comments are answered or resolved

@robodoo delegate+

addons/mail/static/src/core/common/chat_bubble.xml
alexkuhn: `DiscussAvatar` is `<svg>` without any `div`.

addons/mail/static/src/core/public_web/notification_item.xml
alexkuhn: This is unchanged with this PR, and replacing the `<div>` here is likely to affect style.
Better do this in another PR dedicated to fixing this.

addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
alexkuhn: I disagree: the component `DiscussAvatar` is easier to understand without the floating icon part.
It's easier to have condition for adding complexity with this floating icon, rather than the opposite.

Also this prevents double negation to tell positive assertion.

seb-odoo: I know we usually want to avoid double negation but in this case 99% time it is yes and we have to add overrides and complex conditions.
Whereas we only have one or 2 very specific cases where it should be hidden, which are easier to express I feel like.

addons/mail/static/src/core/common/im_status.xml
alexkuhn: Yes this is plan at some point. Here the issue of mask was functional improvement but this was a good excuse to introduce `DiscussAvatar`, which implies removing the technical debt around the many occurrences of `ImStatus` and `ThreadIcon`, and also this part of template that could become much simpler.

So in a sense, current PR makes it way easier to clean this code.

addons/mail/static/src/core/common/discuss_avatar.xml
alexkuhn: We could, but at the same time the initial working set of `32` is arbitrary (though somewhat related to avatar pixel dimension).

And I also think this is easier to read and tweak template with actual values rather than indirection with variables.

The typing resizing can be improved and more declarative, but this is much more work for almost no gain right now, therefore the tweaking of value is just very quick and just works.

addons/mail/static/src/discuss/core/web/burger_menu_patch.xml
alexkuhn: Yes this is something to improve. I don't feel this is a reason to overly engineer the code to tackle this issue now. I think it's good that this pushes for discussions around having some sort of `Avatar` in web, or a component that has `SVG` with masking in corners.

addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
alexkuhn: This is best to do in follow-up PR, because the code was already using the positive statement.
The diff is merely a move from component to model.

Doing the opposite means to state the case for which there's no icon. This could be simpler, but doing in a rush in such a PR is too prone to make accidental regression.

addons/hr_holidays/static/src/im_status_patch.xml
Bertrand2, line 5: What's the use of this? It's not really self explanatory

alexkuhn: The component has been unflatten for cleaner representation in DOM and making it more reusable.

This `attr` combines enrichments from attributes that all items expect to have, notably classnames and optional style received as props.

This could probably be better, but unless I can think of a better solution or someone suggests it, I picked this one.

seb-odoo: The better solution is refactoring of https://github.com/odoo/odoo/pull/210189 but it's out of scope

Bertrand2: Well maybe at least come up with a better name, because tbf this seems like a useless bit in code and nobody will understand why we would need that or not..

> This attr combines enrichments from attributes that all items expect to have

Being?

PR #246190, opened 2026-01-28
[FIX] web: align arrow direction of date picker in RTL

farahdawod:
### Steps to reproduce:
- Download Rental and eCommerce apps.
- Install an RTL language (e.g., Arabic) on the website.
- Create a rental and go to website.
- Try to pick a date for the rental.

### Issue:
When the website is viewed in an RTL language, the navigation arrows of the date picker are displayed in the wrong direction. This happens because the date picker is not inside `o_rtl` component, but inside `o-main-components-container` component. So, when `o_rtl` is called in css (for example):
https://github.com/odoo/odoo/blob/2264f330859b79010b227e3a9fda1075de8ed4e8/addons/web/static/lib/odoo_ui_icons/style.css#L67-L76
Since the arrows are not inside `o_rtl`, the transformation doesn't apply to them.

### Solution:
The `o_rtl` class has been appended to `o-main-components-container` class in case of a RTL language, so that had the css file contain rules for `o_rtl`, they would be applied automatically.

opw-5498615

### DIFF
--- a/addons/web/static/src/core/main_components_container.js
+++ b/addons/web/static/src/core/main_components_container.js
@@ -2,6 +2,7 @@ import { Component, xml } from "@odoo/owl";
 import { registry } from "@web/core/registry";
 import { useRegistry } from "@web/core/registry_hook";
 import { ErrorHandler } from "@web/core/utils/components";
+import { localization } from "@web/core/l10n/localization";
 
 const mainComponents = registry.category("main_components");
 
@@ -14,7 +15,7 @@ export class MainComponentsContainer extends Component {
     static components = { ErrorHandler };
     static props = {};
     static template = xml`
-    <div class="o-main-components-container">
+    <div class="o-main-components-container" t-att-class="{'o_rtl': this.isRTL}">
         <t t-foreach="Components.entries" t-as="C" t-key="C[0]">
             <ErrorHandler onError="error => this.handleComponentError(error, C)">
                 <t t-component="C[1].Component" t-props="C[1].props"/>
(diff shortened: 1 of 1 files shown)


hubvd (changes requested):
If the o_rtl class was applied to the body instead of the `#wrapwrap`, it would automatically work due to this: https://github.com/odoo/odoo/blob/23ffe210a3307db3d0dd642744a9e13cab732222/addons/web/static/lib/odoo_ui_icons/style.css#L94-L106 .
https://github.com/odoo/odoo/blob/d0bc42aede3019561a6f055599d6f68e815778e7/addons/portal/views/portal_templates.xml#L4-L5

Not sure why it's done this way, but I can't think of any downside of passing a props to MainComponentContainer inside createPublicRoot in order to also add the o_rtl class.
What do you think @nle-odoo ?

If we want to keep this fix though, the selector is way to generic.
Something like this would be better: 
```css
#wrapwrap.o_rtl ~ .o-main-components-container .o_datetime_picker_header {
  .oi-chevron-left,
  .oi-chevron-right {
    transform: rotate(180deg); 
  }
}
```

Don't forget to put the opw-id in the commit message

hubvd (commented):
After discussing with the JS team, we would rather add the o_rtl class in the MainComponentsContainer root [1]. This allows the fix to be more generic as other component can be renderer inside the MainComponentsContainer.

You can use localization.direction to conditionally add the class.


[1]: https://github.com/odoo/odoo/blob/2f6af6697e853acd60b9422be53021053cbb09bd/addons/web/static/src/core/main_components_container.js#L17

hubvd (commented):
Can you update the commit message to match the fix ? Thanks

PR #246208, opened 2026-01-28
[FIX] uom: _check_qty precision based on uom rounding

jowes-odoo:
When a product is part of a group with Reserve Only Full Packagings
enabled, another check occurs in _check_qty. This rounds with a precision
of 1.0 the down rounding method. In the case that the quantity was a
decimal, such as 22.4, this would be rounded to 22. In the next transfer
the quantity would then only be 22 instead of the expected 22.4. This
would also cause any packages to not be added as the quantity and demand
are not equal.

If the uom of the product and package are the same we dont need this
package check. Which also prevents the rounding issue.

This fix insures the _check_qty method does not round in cases it does
not need to. (self == uom_id)

How to reproduce:

    In Settings:
        Enable Units of Measure & Packagings
        Enable Multi-Step Routes
    In Warehouses
        Set Outgoing Shipments to Pick then Deliver (2 steps)
    In Inventory
        Create a new product
        Give the product a category
        Set category Reserve Packagings to Reserve Only Full Packagings
        Set on-hand amount

    Workflow:
    Create a sales order
        Create a new Company
        Add product with decimal quantity (1.3)
        Confirm SO
    Go to the sale order delivery
        Validate
    Go to the next transfer
        Quantity will now be rounded (1.0)

opw-5486932

### DIFF
--- a/addons/uom/models/uom_uom.py
+++ b/addons/uom/models/uom_uom.py
@@ -147,6 +147,8 @@ class UomUom(models.Model):
         """
         self.ensure_one()
         packaging_qty = self._compute_quantity(1, uom_id)
+        if self == uom_id:
+            return product_qty
         # We do not use the modulo operator to check if qty is a mltiple of q. Indeed the quantity
         # per package might be a float, leading to incorrect results. For example:
         # 8 % 1.6 = 1.5999999999999996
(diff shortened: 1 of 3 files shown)


addons/uom/models/uom_uom.py
```diff
@@ -147,12 +147,13 @@ def _check_qty(self, product_qty, uom_id, rounding_method="HALF-UP"):
         """
         self.ensure_one()
         packaging_qty = self._compute_quantity(1, uom_id)
+        uom_rounding = uom_id.rounding
```
Whenrow: I would either return `product_qty` in case `self == uom_id` at the beginning of the function or even not call `_check_quantity`

jowes-odoo: After further analysis, returning product_qty in case self == uom_id is the best solution.

PR #212726, opened 2025-06-03
[IMP] l10n_it_edi: remove many2one ir.attachment field

mairasalazar:
This is part of a general change for all the many2one ir.attachment fields to binary in order to avoid security issues as the former would allow access to all ir.attachment records.

In l10n_it_edi, we remove the l10n_it_edi_attachment_id and adapt the use of the existing l10n_it_edi_attachment_field, so only this binary field is used. The field l10n_it_edi_attachment_name is added, so it can be used in views.

task-4771720

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246220, opened 2026-01-28
[FIX] l10n_it_edi: Prevent base64 exceptions when importing TXT.

jacw-odoo:
When importing invoices, some text files encoded in `<Attachments>` cause a traceback and prevent the XML from importing. This error appears to be the direct result of decoding the text files twice, introduced in PR #212726.

Steps to reproduce:

1. Install `l10n_it_edi`.
2. Create a customer invoice for Biscotti Oslengi, then send the xml to the SDI. Download the XML file.
3. In a text editor, replace the base64-encoded PDF information inside `<Attachments>` with `VGhpcyBpcyBhIHRlc3Qgc3RyaW5nLg0K`. This is the base64 encoded string `"This is a test string.\r\n"`.
4. Change the `<FormatoAttachment>` from `PDF` to `TXT`.
5. Upload the edited invoice file to Accounting > Customers > Invoices.
6. The exception `Invalid base64-encoded string` is raised.

[opw-5388360](https://www.odoo.com/odoo/unassigned-tasks/5388360)

---

### DIFF
--- a/addons/l10n_it_edi/models/account_move.py
+++ b/addons/l10n_it_edi/models/account_move.py
@@ -1556,11 +1556,13 @@ class AccountMove(models.Model):
                     message_to_log += self._l10n_it_edi_import_line(element, move_line, extra_info)
 
             for element in tree.xpath('.//Allegati'):
-                self.l10n_it_edi_attachment_name = get_text(element, './/NomeAttachment')
-                self.l10n_it_edi_attachment_file = b64decode(get_text(element, './/Attachment'))
+                raw_name = get_text(element, './/NomeAttachment') or ''
+                raw_ext = get_text(element, './/FormatoAttachment') or ''
+                self.l10n_it_edi_attachment_name = f"{raw_name}.{raw_ext}" if raw_ext else raw_name
+                self.l10n_it_edi_attachment_file = get_text(element, './/Attachment')
                 self.sudo().message_post(
                     body=(_("Attachment from XML")),
-                    attachments=[(self.l10n_it_edi_attachment_name, self.l10n_it_edi_attachment_file)],
+                    attachments=[(self.l10n_it_edi_attachment_name, b64decode(self.l10n_it_edi_attachment_file))],
                 )
 
             global_enasarco_lines = []


addons/l10n_it_edi/models/account_move.py
```diff
@@ -1556,11 +1556,11 @@ def _l10n_it_edi_import_invoice(self, invoice, data, is_new):
                     message_to_log += self._l10n_it_edi_import_line(element, move_line, extra_info)
 
             for element in tree.xpath('.//Allegati'):
-                self.l10n_it_edi_attachment_name = get_text(element, './/NomeAttachment')
-                self.l10n_it_edi_attachment_file = b64decode(get_text(element, './/Attachment'))
+                self.l10n_it_edi_attachment_name = get_text(element, './/NomeAttachment') + b'.' + get_text(element, './/FormatoAttachment')
```
Megaaaaaa: This concatenates `str` values (what `get_text` normally returns) with `bytes` (`b'.'`), which will raise a `TypeError` at runtime, and it also doesn’t guard against missing/empty `FormatoAttachment` (you’ll end up with a trailing dot or `"None"` depending on `get_text`). What I would do instead is build a safe string filename using normal string concatenation and only append the extension when it’s present.
```suggestion
                raw_name = get_text(element, './/NomeAttachment') or ''
                raw_ext = get_text(element, './/FormatoAttachment') or ''
                attachment_name = f"{raw_name}.{raw_ext}" if raw_ext else raw_name
                self.l10n_it_edi_attachment_name = attachment_name
```

addons/l10n_it_edi/models/account_move.py
```diff
@@ -1556,11 +1556,11 @@ def _l10n_it_edi_import_invoice(self, invoice, data, is_new):
...
+                self.l10n_it_edi_attachment_file = get_text(element, './/Attachment')
                 self.sudo().message_post(
                     body=(_("Attachment from XML")),
-                    attachments=[(self.l10n_it_edi_attachment_name, self.l10n_it_edi_attachment_file)],
+                    attachments=[(self.l10n_it_edi_attachment_name, b64decode(self.l10n_it_edi_attachment_file))],
```
Megaaaaaa: `message_post(attachments=...)` expects attachment payloads in base64 (the value stored in `ir.attachment.datas`), so decoding here will push raw bytes and can lead to corrupted attachments or subtle encoding-dependent failures. 
```suggestion
                    attachments=[(self.l10n_it_edi_attachment_name, self.l10n_it_edi_attachment_file)],
```

Megaaaaaa (changes requested):
Hello hello :wave:

Thank you for the fix! Here's a very quick round of review to tell you that ci/l10n is red on your last build.

Please tell me if I missed something :+1:

addons/l10n_it_edi/models/account_move.py
jacw-odoo: Hey @Megaaaaaa  ! After making the changes, I thought about it some more. I decoded the base64 string because text files do not decode this information. I recorded a video to demonstrate: https://drive.google.com/file/d/1RMzg0VDom2ArRdGXfJ9aNckOo-_Fdr_J/view?usp=sharing

The string `'VGhpcyBpcyBhIHRlc3Qgc3RyaW5nLg0K'` is not decoded. From my initial testing, PDFs seemed to be rendered fine, but I probably did not run into the corruption or subtle encoding dependent failures that you mentioned.

Before PR #212726, the string was wrapped in `str.encode()` and saved directly to `IrAttachment.datas`. After, the decoded string was saved to the field `AccountMove.l10n_it_edi_attachment_file` and fed directly into the attachment during the `message_post`. 

Would it be appropriate to leave the decoding in this instance? Or should I add some logic to only decode when the file has the 'simple/text' mimetype? Or am I missing something else?

Megaaaaaa: Good catch, and thanks for digging into this (and for the video). You’re right, in this case it is appropriate to keep the decoding, and it shouldn’t be mimetype-specific. So yes, let’s keep the b64decode() here, no need for mimetypes exceptions, it's a global thing. :+1:

Megaaaaaa (approved):
Thank you for the fix, LGTM :+1:

robodoo r+

addons/l10n_it_edi/models/account_move.py
```diff
@@ -1556,11 +1556,13 @@ def _l10n_it_edi_import_invoice(self, invoice, data, is_new):
                     message_to_log += self._l10n_it_edi_import_line(element, move_line, extra_info)
 
             for element in tree.xpath('.//Allegati'):
-                self.l10n_it_edi_attachment_name = get_text(element, './/NomeAttachment')
-                self.l10n_it_edi_attachment_file = b64decode(get_text(element, './/Attachment'))
+                raw_name = get_text(element, './/NomeAttachment') or ''
+                raw_ext = get_text(element, './/FormatoAttachment') or ''
+                self.l10n_it_edi_attachment_name = f"{raw_name}.{raw_ext}" if raw_ext else raw_name
```
jacw-odoo, line 1561: should have used `endswith()` to check that the extension doesn't already exist in the file name.

```py
f"{raw_name}.{raw_ext}" if raw_ext and not raw_name.endswith(raw_ext) else raw_name
```

PR #246231, opened 2026-01-29
[IMP] website_sale_stock: validate product availability for stock notifications

elierwclik:
Currently, in the website_sale_stock module, there is no backend validation when subscribing to notifications for products without stock. This allows public users to potentially use emails that belong to registered accounts.

Current behavior before PR:
- Users could subscribe to stock notifications for products that don’t exist or cannot be added (no stock).
- Public users could use emails already associated with registered accounts, allowing them to subscribe on behalf of another user.
- No validation is enforced, leading to potential security issues.

- Adding a subscription for a non-existent or unavailable product raises a ValidationError.
- Public users trying to subscribe with an email that belongs to a registered user receive an AccessError prompting them to sign in first.
- Backend validation prevents misuse of registered user emails and improves security.

---

### DIFF
--- a/addons/website_sale_stock/controllers/main.py
+++ b/addons/website_sale_stock/controllers/main.py
@@ -5,28 +5,37 @@ from odoo.addons.website_sale.controllers import main as website_sale_controller
 from odoo.tools import email_re
 from odoo import http, _
 from odoo.http import request
-from werkzeug.exceptions import BadRequest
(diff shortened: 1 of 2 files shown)


elierwclik:
Note:
I think it would be good to add additional validation because, without proper backend checks, repeated requests could create multiple partner records if they don't exist.

PR #246236, opened 2026-01-29
[IMP] website_sale_stock: validate product availability for stock notifications

elierwclik:
Currently, in the website_sale_stock module, there is no backend validation when subscribing to notifications for products without stock. This allows public users to potentially use emails that belong to registered accounts.

Current behavior before PR:
- Users could subscribe to stock notifications for products that don’t exist or cannot be added (no stock).
- Public users could use emails already associated with registered accounts, allowing them to subscribe on behalf of another user.
- No validation is enforced, leading to potential security issues.

- Adding a subscription for a non-existent or unavailable product raises a ValidationError.
- Public users trying to subscribe with an email that belongs to a registered user receive an AccessError prompting them to sign in first.
- Backend validation prevents misuse of registered user emails and improves security.

---


addons/website_sale_stock/controllers/main.py
```diff
@@ -5,23 +5,32 @@
...
 
 
 class WebsiteSale(website_sale_controller.WebsiteSale):
     @http.route(['/shop/add/stock_notification'], type="json", auth="public", website=True)
     def add_stock_email_notification(self, email, product_id):
         if not email_re.match(email):
-            raise BadRequest(_("Invalid Email"))
+            raise ValidationError(_("Invalid Email"))
 
+        is_public_user = request.website.is_public_user()
```
ltinel: Could you move this line right above where it's first used?

addons/website_sale_stock/controllers/main.py
```diff
@@ -5,23 +5,32 @@
...
         product = request.env['product.product'].browse(int(product_id))
+
+        # check if product available
+        if not product.exists() or not product._is_add_to_cart_allowed():
```
ltinel: I'm not sure `_is_add_to_cart_allowed` is the best choice here. In some modules, this method checks whether the product is compatible with the cart (e.g. rental and subscription), which doesn't seem right.

@Feyensv do you know if we already have a method that could be used or if we should create a new one?

addons/website_sale_stock/static/src/js/website_sale.js
```diff
@@ -47,14 +50,16 @@ WebsiteSale.include({
...
-            this._displayEmailIncorrectMessage(stockNotificationEl);
+            let message = error instanceof RPCError ? error.data.message : defaultMessage;
+            this._displayEmailIncorrectMessage(message, stockNotificationEl);
         });
     },
 
-    _displayEmailIncorrectMessage(stockNotificationEl) {
+    _displayEmailIncorrectMessage(message, stockNotificationEl) {
         const incorrectIconEl = stockNotificationEl.querySelector('#stock_notification_input_incorrect');
+        incorrectIconEl.htmlContent = `<i class="fa fa-times text-danger"/> ${message}`;
```
ltinel, line 60: It's probably best to double-check this with the design team or the PO. Since we're now throwing errors from the backend, it might be easier let the errors propagate and let the framework handle them.

This will display an error dialog instead of displaying an error message below the input though. WDYT?

@Jean-boje FYI

ltinel (changes requested):
Thanks for the fix, it's indeed better to have some backend safety checks 😅

addons/website_sale_stock/controllers/main.py
elierwclik: Yes, I'll actually wait for the team's response before making all the changes at once. Thanks for the review.

PR #242867, opened 2026-01-08
[FIX] delivery: default shipping method on so

odoo-pda:
https://github.com/odoo/odoo/pull/203955, by wanting to ensure no unallowed carrier was set as default, removed the possibility to set one. 
This PR puts back the possibility of having a default shipping method if part of the available carriers.

opw-5263540

PR #203955, opened 2025-03-28
[IMP] {stock_}delivery: restrict delivery carrier selection in pickings

mohammadhossam:
This commit applies the same restrictions applied on sale orders when selecting a delivery carrier, on stock pickings. Therefore, consistency is ensured between both models.

Task-4555994

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246246, opened 2026-01-29
[FIX] delivery: ensure carrier is pre-selected

jand-odoo:
### Current behavior:
When updating shipping costs on a Sales Order, the delivery wizard opens with an empty carrier field. This happens because a recordset is passed to the context instead of an integer ID, causing the web client to fail the auto-population.

### Expected behavior:
The wizard should automatically select the Sales Order's current carrier as the default carrier_id

### Steps to reproduce:
- Add a shipping method to a Sales Order.
- Click "Update shipping cost".
- Observe the carrier field in the wizard is empty.

### Fix:
Pass `self.carrier_id.id `as an integer in the action context instead of the recordset object.

Caused by this commit: https://github.com/odoo/odoo/pull/203955/changes/651dd4313cf23105675249d569affee7b756cd3c

Backport of: https://github.com/odoo/odoo/pull/242867

opw-5890258

### DIFF
--- a/addons/delivery/models/sale_order.py
+++ b/addons/delivery/models/sale_order.py
@@ -128,8 +128,15 @@ class SaleOrder(models.Model):
         view_id = self.env.ref('delivery.choose_delivery_carrier_view_form').id
         if self.env.context.get('carrier_recompute'):
             name = _('Update shipping cost')
+            carrier = self.carrier_id
         else:
             name = _('Add a shipping method')
+            shipping_partner_id = self.with_company(self.company_id).partner_shipping_id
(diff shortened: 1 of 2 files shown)


vchu-odoo (commented):
Hello,

Thank you for the fix!
We already handled this issue for 19.0+ in https://github.com/odoo/odoo/pull/242867 and missed 18.4.
for consistency it is better to backport the commit, could you please do it in this PR?

vchu-odoo (approved):
Thank you for the changes!

@robodoo r+ fw=no

PR #246249, opened 2026-01-29
[FIX] website: show color option for image with shape

alap-odoo:
Steps to reproduce:
1. Go to the website and enter edit mode.
2. Drop `s_cta_mockups` or `s_closer_look` snippet.
3. Click on any image that has a shape.

Issue:
The color picker option is missing for images with shapes in these snippets.

Reason:
These snippet templates do not include the `shapeColors` dataset on the image elements.

task-5880905

### DIFF
--- a/addons/website/static/src/js/editor/snippets.options.js
+++ b/addons/website/static/src/js/editor/snippets.options.js
@@ -1418,6 +1418,32 @@ options.registry.ReplaceMedia.include({
 });
 
 options.registry.ImageTools.include({
+    /**
+     * Compute and set default shape colors for images that have a data-shape
+     * but are missing data-shape-colors (e.g. s_cta_mockups, s_closer_look).
+     *
+     * @override
+     */
+    async _applyOptions() {
+        const img = await this._super(...arguments);
+        // TODO remove in master: kept for stable.
+        if (!img) {
+            return img;
+        }
+        const { shape: shapePath, shapeColors } = img.dataset;
+        if (shapePath && (!shapeColors || shapeColors === ";;;;")) {
+            const shapeName = shapePath.split("/")[2];
+            const shape = this.shapeCache[shapeName];
+            if (shape) {
+                const palette = Object.values(weUtils.DEFAULT_PALETTE);
+                const defaultColors = palette.map((color) =>
+                    shape.includes(color) ? color : null
+                );
+                img.dataset.shapeColors = defaultColors.join(";");
+            }
+        }
+        return img;
+    },
     async _computeWidgetVisibility(widgetName, params) {
         if (params.optionsPossibleValues.selectStyle
                 && params.cssProperty === 'width'
--- a/addons/website/views/snippets/s_closer_look.xml
+++ b/addons/website/views/snippets/s_closer_look.xml
@@ -13,7 +13,7 @@
             <p>
                 <br/>
             </p>
-            <img class="img-fluid mx-auto" data-shape="web_editor/devices/macbook_front" data-file-name="s_closer_look.jpg" data-original-mimetype="image/jpeg" src="/web_editor/image_shape/website.s_closer_look_default_image/web_editor/devices/macbook_front.svg" alt=""/>
+            <img class="img-fluid mx-auto" data-shape="web_editor/devices/macbook_front" data-file-name="s_closer_look.jpg" data-original-mimetype="image/jpeg" src="/web_editor/image_shape/website.s_closer_look_default_image/web_editor/devices/macbook_front.svg" alt="" data-shape-colors=";;#F6F6F6;;"/>
         </div>
     </section>
 </template>
--- a/addons/website/static/tests/tours/cta_mockups_shape_color_option.js
+++ b/addons/website/static/tests/tours/cta_mockups_shape_color_option.js
@@ -1,0 +1,24 @@
+import { insertSnippet, registerWebsitePreviewTour } from "@website/js/tours/tour_utils";
+
+registerWebsitePreviewTour(
+    "cta_mockups_shape_color_option",
+    {
+        url: "/",
+        edition: true,
+    },
+    () => [
+        ...insertSnippet({
+            id: "s_cta_mockups",
+            groupName: "Content",
+        }),
+        {
+            content: "Click on the laptop shape",
+            trigger: ":iframe .s_cta_mockups img[data-shape$='macbook_front']",
+            run: "click",
+        },
+        {
+            content: "Check that the image shape color option appears in the sidebar",
+            trigger: "#oe_snippets .o_we_image_shape .o_we_color_preview",
+        },
+    ]
+);
--- a/addons/website/tests/test_ui.py
+++ b/addons/website/tests/test_ui.py
@@ -806,3 +806,6 @@ class TestUi(odoo.tests.HttpCase):
             snippet_key='s_button',
             template_key='website.snippets')
         self.start_tour('/', 'adapt_custom_button_on_drop', login='admin')
+
+    def test_cta_mockups_shape_color_option(self):
+        self.start_tour('/', 'cta_mockups_shape_color_option', login='admin')
--- a/addons/website/views/snippets/s_cta_mockups.xml
+++ b/addons/website/views/snippets/s_cta_mockups.xml
@@ -6,7 +6,7 @@
         <div class="container">
             <div class="row o_grid_mode" data-row-count="9">
                 <div class="o_grid_item o_grid_item_image g-height-9 g-col-lg-7 col-lg-7" style="grid-area: 1 / 6 / 10 / 13; z-index: 1;">
-                    <img src="/web_editor/image_shape/website.s_cta_mockups_default_image/web_editor/devices/macbook_front.svg" class="img img-fluid" data-shape="web_editor/devices/macbook_front" data-original-mimetype="image/jpeg" data-file-name="s_cta_mockups.jpg" alt=""/>
+                    <img src="/web_editor/image_shape/website.s_cta_mockups_default_image/web_editor/devices/macbook_front.svg" class="img img-fluid" data-shape="web_editor/devices/macbook_front" data-original-mimetype="image/jpeg" data-file-name="s_cta_mockups.jpg" alt="" data-shape-colors=";;#F6F6F6;;"/>
                 </div>
                 <div class="o_grid_item g-height-6 g-col-lg-4 col-lg-4" style="grid-area: 3 / 1 / 9 / 5; z-index: 2;">
                     <h3>50,000+ companies trust Odoo.</h3>
@@ -14,7 +14,7 @@
                     <a t-att-href="cta_btn_href" class="btn btn-primary btn-lg"><t t-out="cta_btn_text">Contact us</t></a>
                 </div>
                 <div class="o_grid_item o_grid_item_image o_snippet_mobile_invisible g-height-8 g-col-lg-2 col-lg-2 d-none d-lg-block" style="grid-area: 2 / 6 / 10 / 8; z-index: 3;">
-                    <img src="/web_editor/image_shape/website.s_cta_mockups_default_image_1/web_editor/devices/iphone_front_portrait.svg" class="img img-fluid" data-shape="web_editor/devices/iphone_front_portrait" data-original-mimetype="image/jpeg" data-file-name="s_cta_mockups_1.jpg" alt=""/>
+                    <img src="/web_editor/image_shape/website.s_cta_mockups_default_image_1/web_editor/devices/iphone_front_portrait.svg" class="img img-fluid" data-shape="web_editor/devices/iphone_front_portrait" data-original-mimetype="image/jpeg" data-file-name="s_cta_mockups_1.jpg" alt="" data-shape-colors=";;;;#383E45"/>
                 </div>
             </div>
         </div>


addons/website/static/src/snippets/s_cta_mockups/000.js
```diff
@@ -0,0 +1,15 @@
...
+import { addShapeColorAttribute } from "@website/js/utils";
+
+const CTAMockupsWidget = publicWidget.Widget.extend({
+    selector: ".s_cta_mockups",
+    disabledInEditableMode: false,
+    // Todo: Remove this logic before merging into the master branch.
+    start: function () {
+        const imgEls = this.el.querySelectorAll("img");
+        addShapeColorAttribute(imgEls);
+    },
```
shsa-odoo: Doesn't call this._super(...arguments), breaking the widget lifecycle. publicWidget.Widget.start() sets internal state and returns a Promise. This could cause subtle issue

addons/website/static/src/snippets/s_cta_mockups/000.js
```diff
@@ -0,0 +1,15 @@
+import publicWidget from "@web/legacy/js/public/public_widget";
+import { addShapeColorAttribute } from "@website/js/utils";
+
+const CTAMockupsWidget = publicWidget.Widget.extend({
+    selector: ".s_cta_mockups",
+    disabledInEditableMode: false,
+    // Todo: Remove this logic before merging into the master branch.
+    start: function () {
+        const imgEls = this.el.querySelectorAll("img");
+        addShapeColorAttribute(imgEls);
```
shsa-odoo: Say, I have added couple of more images to this snippet, which has nothing to do with shape, this line here will add the extra attribute "data-shape-colors".

addons/website/static/src/js/utils.js
```diff
@@ -513,6 +513,18 @@ export function checkAndNotifySEO(seo_data, OptimizeSEODialog, services) {
...
 
+/**
+ * Assigns a "data-shape-colors" attribute to each provided image element.
+ *
+ * @param {Array} imgEls - image elements to update.
+ */
+export function addShapeColorAttribute(imgEls) {
+    const shapeColors = [";;#F3F2F2;;", ";;;;#111827"];
+    imgEls.forEach((imgEl, index) => {
+        imgEl.setAttribute("data-shape-colors", shapeColors[index]);
```
shsa-odoo: No bounds checking , if there are 3+ images `shapeColors[index]` will set undefined .

addons/website/static/src/js/editor/snippets.editor.js
```diff
@@ -305,6 +306,11 @@ export class WebsiteSnippetsMenu extends weSnippetEditor.SnippetsMenu {
             cardRatioOptionEl.dataset.applyTo = ">.o_card_img_wrapper";
         }
 
+        // TODO remove in master: fix for missing shape color option
+        // initialization.
+        const imgEls = html.querySelectorAll(".s_cta_mockups img");
```
shsa-odoo: `s_closer_look` has the same issue. 😉

Instead of introducing a dedicated public widget and patching multiple files, could we handle this by overriding `_applyOptions` in `options.registry.ImageTools.include` (in `website/snippets.options.js`)?

This would keep the fix self-contained — no new widget, no extra utility function, and no snippet editor patch. Something like:

```js
/**
 * @override
 */
async _applyOptions() {
    const img = await this._super(...arguments);
    // TODO: remove in master (add data-shape-colors in XML template).
    // Compute and set default shape colors for images that have a
    // data-shape but are missing data-shape-colors (e.g. s_cta_mockups).
    if (img && img.dataset.shape && !img.dataset.shapeColors) {
        const shapeName = img.dataset.shape.split('/')[2];
        const shape = this.shapeCache[shapeName];
        if (shape) {
            const defaultColors = Object.values(weUtils.DEFAULT_PALETTE)
                .map((color, i) => shape.includes(color)
                    ? this._getCSSColorValue(`o-color-${i + 1}`)
                    : null
                );
            img.dataset.shapeColors = defaultColors.join(';');
        }
    }
    return img;
},
```

This runs at the right moment (when the editor loads the image options), has access to `shapeCache` and `_getCSSColorValue` already, and resolves colors against the current theme so it works generically for any image with a missing `data-shape-colors`, not just `s_cta_mockups`.

Could you give this approach a try and let me know how it goes?

shsa-odoo (commented):
Hello @alap-odoo :wave: Here is my review please have a look.

addons/website/static/src/js/editor/snippets.editor.js
alap-odoo: The solution introduces a bug and the steps to reproduce are:
1. Change the color 3 of the theme preset.
2. Drop the `s_cta_mockups` snippet.
3. Click on the image.

Issue:
The shape has the initial color while the color picker of that image option shows the updated color of the shape.
image

Am I doing something wrong ? :thinking: 

If not, I think that instead of `this._getCSSColorValue("o-color-${i + 1}")` we should simply return the color, so it will return the color which is applied on the svg. What's your views on it ? :slightly_smiling_face: 

```javascript
const defaultColors = Object.values(weUtils.DEFAULT_PALETTE).map((color, i) => shape.includes(color) ? color : null);
```

shsa-odoo: Mmm, may be try setting the colors, before the super call, I believe this bug will be resolved. :thinking: Have you tried ?
something like:
```js
const img = this._getImg();

// set your colors

return this._super(...arguments);

```

alap-odoo: Yes, I thought that way also but here also there is an issue :slightly_smiling_face: 
The issue is something like as we haven't call the super method, `this.shapeCache` will be empty and the color attribute will not apply at all. Moreover, calling `await this._loadShape(img.dataset.shape);` (which generates the shapeCache) is inefficient because when the super is called at the end of this function, it will be recalled by the parent function.

image

Reference of the parent `_applyOptions` function
image

So, I think assigning the color is simple solution. :thinking:

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,27 @@ options.registry.ReplaceMedia.include({
...
+    /**
+     * Compute and set default shape colors for images that have a data-shape
+     * but are missing data-shape-colors (e.g. s_cta_mockups, s_closer_look).
+     *
+     * @override
+     */
+    async _applyOptions() {
+        const img = await this._super(...arguments);
+        // TODO remove in master: kept for stable.
+        if (img && img.dataset.shape && !img.dataset.shapeColors) {
```
shsa-odoo: const {shape, shapeColor} = img.dataset.

then use then where neeeded.

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,27 @@ options.registry.ReplaceMedia.include({
...
+            const shapeName = img.dataset.shape.split("/")[2];
+            const shape = this.shapeCache[shapeName];
+            if (shape) {
+                const defaultColors = Object.values(weUtils.DEFAULT_PALETTE).map((color, i) =>
```
shsa-odoo: I know it will become verbose, however we can split  it into 2 parts, for better readability:

const palette = Object.values(weUtils.DEFAULT_PALETTE);
const defualtColors palette.map(...)

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,27 @@ options.registry.ReplaceMedia.include({
...
 options.registry.ImageTools.include({
+    /**
+     * Compute and set default shape colors for images that have a data-shape
+     * but are missing data-shape-colors (e.g. s_cta_mockups, s_closer_look).
+     *
+     * @override
+     */
+    async _applyOptions() {
+        const img = await this._super(...arguments);
+        // TODO remove in master: kept for stable.
```
shsa-odoo, line 1429: "// TODO master: remove fallback kept for stable compatibility" or something better you have in mind.

shsa-odoo (changes requested):
couple of comments to be resolved. :)

addons/website/static/src/js/editor/snippets.options.js
alap-odoo: I don’t think any changes are needed here, as the current wording is clear and understandable. The same phrasing is also used elsewhere, so keeping it consistent makes sense.

Image

shsa-odoo (approved):
In fw ports maybe we can add the hoot test instead of tour. Rest LGTM :)

bvr-odoo (changes requested):
👋🏼 Hello @alap-odoo In my opinion, we could have fixed this only from 18.4. I do not think the bug is important enough to bother fixing it both before and after the refactoring. But fine, let us keep it like this since it is already done 😉 

Could you also fix the XMLs? So new databases will have the correct snippets and they will not need the JS fix.

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,32 @@ options.registry.ReplaceMedia.include({
...
+        if (!img) {
+            return img;
+        }
+        const { shape: shapePath, shapeColors } = img.dataset;
+        if (shapePath && !shapeColors) {
```
shsa-odoo: After switching the theme(nano), this shape now has an attribute defined without any value (`;;;;`). Because of that, the condition fails, so the shape value doesn’t get applied, and as a result, the option isn’t shown.

image

alap-odoo: Hello @msh-odoo @Brieuc-brd @stefanorigano 
I have observed that in several design themes, the attribute `data-shape-colors=";;;;"` is passed in this format. I would like to understand the reasoning behind this approach. Specifically, why are no color values provided within this attribute?
image

Due to the absence of defined colors, the color picker option does not appear in the interface. It would be helpful to know whether this is intentional and what purpose this configuration serves.

Brieuc-brd: Hello @alap-odoo,

If I’m not mistaken, this was previously required to apply the default color of a shape. However, it seems that something may have broken in the latest editor updates. 🤔 

Hey @odoo/rd-website 👋  are you aware of this issue?

FrancoisGe: > data-shape-colors=";;;;

I'm not sure I see the issue. Which version does the bug occur in, and what code snippet can I use to reproduce it ?

shsa-odoo:
> Hello @shsa-odoo The issue is resolved. can you please have a look on it ?

@alap-odoo LGTM :) 

@bvr-odoo Please review from your end. Thanks!!

addons/website/views/snippets/s_closer_look.xml
bvr-odoo, line 1: You forgot to check the shapes in `design-themes`.
for instance:
- s_image_hexagonal (theme_kiddo)
- s_text_image (theme_graphene)
- s_text_image, s_image_text, s_showcase (theme_odoo_experts)

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,32 @@ options.registry.ReplaceMedia.include({
...
+        // TODO remove in master: kept for stable.
+        if (!img) {
+            return img;
+        }
+        const { shape: shapePath, shapeColors } = img.dataset;
+        if (shapePath && (!shapeColors || shapeColors === ";;;;")) {
+            const shapeName = shapePath.split("/")[2];
+            const shape = this.shapeCache[shapeName];
+            if (shape) {
+                const palette = Object.values(weUtils.DEFAULT_PALETTE);
```
bvr-odoo, line 1438: The colors added here are not the right ones. It works only when the palette has not been changed by the user.

New palette:
Image

For example, if you change the color palette and then drop the snippet, the colors will still be the ones from the default palette, not the ones from the current palette.

Options (tested with `Image Hexagonal` (kea):

Image

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,32 @@ options.registry.ReplaceMedia.include({
 });
 
 options.registry.ImageTools.include({
+    /**
+     * Compute and set default shape colors for images that have a data-shape
+     * but are missing data-shape-colors (e.g. s_cta_mockups, s_closer_look).
+     *
+     * @override
+     */
+    async _applyOptions() {
```
bvr-odoo, line 1427: Could you do this in a function that is more specific to shapes?

bvr-odoo (changes requested):

addons/website/static/src/js/editor/snippets.options.js
alap-odoo: I think it correctly displays the color which is present in the shape of the snippet.
image
image

I tried with multiple snippets and the options are showing the same color as that of snippet. Am I missing something?

bvr-odoo: > the options are showing the same color as that of snippet

Indeed but it is not the issue, this issue is that the colors of the shape are not the colors of the current palette. With the .js you get the default palette not the current palette

addons/website/views/snippets/s_closer_look.xml
alap-odoo: I’m a bit confused here. Why do we need to add this in XML?
I understand your point that defining it in XML helps new users, but if the JavaScript logic is already in place, why shouldn’t we rely on JS instead?
In this case, the colors are already being applied to the shape; the only thing missing is the attribute. Adding that attribute is not a complex operation and shouldn’t have any noticeable performance impact if handled in JavaScript. Also, based on our current JS implementation, every time the options are modified, it already checks whether the shape contains the shapeColor attribute. 
If the case was something like the rendered template did not bring the color to the shape and we have to rebuild the shape svg with the colors (making the call to server), then it's ok that for new user atleast we will have better performance.

bvr-odoo: > I’m a bit confused here. Why do we need to add this in XML? I understand your point that defining it in XML helps new users, but if the JavaScript logic is already in place, why shouldn’t we rely on JS instead? In this case, the colors are already being applied to the shape; the only thing missing is the attribute. Adding that attribute is not a complex operation and shouldn’t have any noticeable performance impact if handled in JavaScript. Also, based on our current JS implementation, every time the options are modified, it already checks whether the shape contains the shapeColor attribute. If the case was something like the rendered template did not bring the color to the shape and we have to rebuild the shape svg with the colors (making the call to server), then it's ok that for new user atleast we will have better performance.

I understand that you may feel this is not strictly necessary, but then why ask the question now, when we need to make a PR in design-themes, and not earlier when I asked to adapt the XMLs in community?

One of the goal of changing these XMLs in stable is to make sure we do not forget to do it in master, which we will also need to do in design-themes. It also helps keep the code correct in stable, which sets the right example if someone later needs to do something similar in stable and uses this as a reference.

Basically, it costs nothing to do it, so why not.

alap-odoo: > Then why raise the question now, when we need to make a PR in `design-themes`?

I had thought about this earlier as well, but at that time I was convinced since the change was small. Now, however, if we proceed with a PR in `design-themes`, it may impact more files. Given that scope, I need to think it’s worth reconsidering whether such change is actually necessary.

Moreover, if we add these attributes in XML, it could create confusion later. Someone might wonder: if the attribute is already defined in XML, why is there still logic in JavaScript to handle it?

> One of the goals of updating these XMLs in stable is to ensure we don’t forget to do it in master, which will also require changes in `design-themes`.

For that, we are already adding a comment in the JavaScript indicating that it should be removed in master and is only kept for stable.

> It also helps keep the code correct in stable, setting a good example for similar future changes.

In practice, developers tend to refer more to community or enterprise modules for examples. In `design-themes`, we usually just modify attributes (add/update/remove) using XPath based on theme requirements. So while adding this in XML makes sense for the community version as a reference, Sorry but I personally don’t think it is necessary for `design-themes`.

bvr-odoo: @alap-odoo Even if this is in stable, and current users will not get the XML changes, and even if the JS is there to fix existing databases, we still want correct snippets in stable, without missing parts or similar issues.

cc @sobo-odoo

alap-odoo: Okay

addons/website/static/src/js/editor/snippets.options.js
alap-odoo: Actually, before this PR also the shapes are getting the colors but my PR just brings the option to change it which was not present before this PR.

Moreover, in the options the current palette color will be displayed. see image
image
P.S: I have changed the `o-color-1` from the color palette from orange to blue.



> the colors of the shape are not the colors of the current palette

I agree that the color in the snippet is not changed, but I don't think it is under the scope of this task. Can we handle  it separately ?

bvr-odoo: Don't forget to disable your cache when testing shape colors.
Because old shapes (with previous applied colors) are kept in the cache so colors can be wrong 😉

addons/website/static/src/js/editor/snippets.options.js
alap-odoo: I am not sure to understand. But I think you are telling to handle the shape related logic in the different function like this :point_down: . Can you please confirm ? 
image

bvr-odoo: Hello @alap-odoo 

No, that is not what I meant 😉  My question was: is there not another function you could override, one that is more specific to the shapes, instead of `_applyOptions`? Because `_applyOptions` is applied to all images, including those without shapes.

alap-odoo:
Hello @shsa-odoo 

As per the suggestion, I have created the design-theme [PR](https://github.com/odoo/design-themes/pull/1268/changes), Can you please check it once ?

addons/website/static/src/js/editor/snippets.options.js
```diff
@@ -1418,6 +1418,32 @@ options.registry.ReplaceMedia.include({
 });
 
 options.registry.ImageTools.include({
+    /**
+     * Compute and set default shape colors for images that have a data-shape
+     * but are missing data-shape-colors (e.g. s_cta_mockups, s_closer_look).
```
bvr-odoo, line 1423: I noticed that on the 18.0 branch, without your fix, changing any image option is enough to make the color picker appear.

Couldn’t we simply trigger the same code that makes the color picker appear at that moment, instead of rewriting that code here?

Something like

Image

alap-odoo: During development, I also tried using this approach, but it appears to re-build the shape. In our case, this is unnecessary and may lead to performance issues.

bvr-odoo: > During development, I also tried using this approach, but it appears to re-build the shape. In our case, this is unnecessary and may lead to performance issues.

I don’t think performance is really a problem here.

`_reapplyCurrentShape` is already executed every time an image option containing a shape is changed, and even when it is only previewed. Running it once in this case should not cause any noticeable performance issue.

On the other hand, this is much more advantageous as a temporary fix, because it is only 3 very simple lines of code instead of around 20 lines.

So I don’t really understand the performance argument compared to the simplicity of my proposal.

alap-odoo: Hello @bvr-odoo :wave: 

> _reapplyCurrentShape is already executed every time an image option containing a shape is changed, and even when it is only previewed

I agree that it is applied on option change, but it will worth as we need to re-build the shape on option change and even on preview also. And my point is, in our case where we just want the dataset, this will be unnecessary computation (though it take very less time) as per my POV. Moreover, I found a small issue 

https://github.com/user-attachments/assets/3141be75-e8d1-469c-8dc5-33f357997c4e

Steps to reproduce:
1. In 18.0 version, switch of theme_kea.
2. Drop the `s_image_hexagon` snippet. -> It will be noticed that the default shape color defers from the theme palette.
3. Switch to this branch and apply the suggested solution instead of current one.
4. Enter the edit mode and the color will be changed.

Reason:
As per the suggested approach to rebuild the shape, which takes the color from the current palette instead of the default palette, it leads to change in color. Moreover, I agree that the new color is the correct one, but as user was expecting that old wrong color, I believe it should not be changed. Am I missing something, WDYT? :thinking: 

P.S: The current solution does not changes the color.

If we are okay with it, I am ready to proceed with the suggested approach :+1:

bvr-odoo: @alap-odoo I can see the bug you mentioned, but this is another bug that already exists without your branch:

Try this on 18.0 with `theme_kea`:

- Drop the s_image_hexagon snippet.
- For example, change the shape speed.
- Bug: the shape colors change.

So this bug already exists. It seems that the shape colors are not correct when the snippet is dropped, and only become correct after changing an option.

With your current solution, the colors no longer change, so they never become correct, even after changing an option. This hides the color bug even more.

We should investigate why it does not work as expected (check if it works correctly in master).

alap-odoo: > check if it works correctly in master

No, it does not work in master also.

But, in this PR, if we drop a new snippet then the color is correct. This happens because the color variables passed in the URL helps to build the shape with correct color.

bvr-odoo:
@robodoo remindme:master="remove js fix"

bvr-odoo:
@robodoo r+

PR #246252, opened 2026-01-29
[FIX] point_of_sale: do not change price on lot change

aksi-odoo:
Step to reproduce:
- have a lot tracked product, product 1 (price = 10)
- create a combo product with product 1, with price (100)
- start a pos, add combo product in order,
- notice total price is 100
- change lot number of product 1,
- notice order price reset to 10.

Cause:
- When the lot number is changed, `set_quantity_by_lot` is triggered.
- This calls `set_quantity`, which resets the price and loses the combo pricing.

Fix:
- use `keep_price` = true parameter when calling `set_quantity` if orderline
has combo_parent_id

opw-5495409

---


alsh-odoo (approved):

adgu-odoo (commented):
robodoo r+

PR #246258, opened 2026-01-29
[IMP] point_of_sale: refund button class based on quantity

paso-odoo:
Before this commit:
--------
- On the ticket screen, when no quantity was selected for refund, the Refund button was still shown as primary.

After this commit:
--------
- Show the Refund button as secondary when no quantity is selected.
- Switch the button to primary only when at least one item quantity is selected.

task-5490812

---


davidmonnom (approved):
@robodoo r+

PR #246261, opened 2026-01-29
[FIX] account_peppol: get_participant_status should archive proxy_user

gawa-odoo:
Currently, if the participant_status gets a client_gone, we call the _reset_peppol_configuration method.
But while we reset, we don't archive the proxy_user. 
We should do so, so the user can re-register

---


clbr-odoo (approved):
@robodoo r+

PR #246263, opened 2026-01-29
[FIX] mass_mailing: prevent blockquote content loss

walidsahli:
Problem:
When sending an email containing a blockquote, its content is missing in the received email.

Cause:
The `bootstrapToTable` conversion logic strictly filters the children of `.row` elements and only keeps nodes that have valid Bootstrap column classes.
In the `s_blockquote` snippet, the content was wrapped in a plain `<div>` placed directly inside a `.row`, so it was ignored during conversion.

Solution:
Add the `col-12` class to the inner `<div>` of the blockquote snippet so it is recognized as a valid Bootstrap column and preserved during conversion.

Steps to reproduce:
- Create a new email marketing.
- Add the "Blockquote" snippet.
- Send the email.
- Observe that the received email loses the blockquote content.

opw-5883624

---


abd-msyukyu-odoo:
@robodoo r+

PR #246267, opened 2026-01-29
[REM] barcodes: barcode_services: remove unused `target`

adr-odoo:
Before this commit, `target` was sent as an option of the `barcode_scanned` event but it seems that no one uses it anymore.

This commit avoids sending it.

Current behavior before PR:

---


svs-odoo (approved):
I see no use of `target` in implemented methods using the `barcode_scanned` event.
Also, I can't image usecase for it, and if we need it someday, I guess we can revert this commit but we usually don't keep useless code so IMO we can merge it 👍

PR #246269, opened 2026-01-29
[FIX] point_of_sale: show order time in order tab according to locale

Nokoz:
In POS, when using a 24h language, the time shown in the order tab (TicketScreen) was always in 12h/AM-PM format instead of 24h.

Steps to reproduce:
-------------------
Set the POS (or user) language to a 24h locale (e.g. French, Spanish) Open POS and create an order
Open the Orders tab and look at the time shown for the order

> Observation:
Time is displayed in 12h (e.g. 4:30) instead of 24h (e.g. 16:30).

Why the fix:
------------
The order tab time was formatted with a hardcoded 12h format ("hh:mm"), ignoring the user’s locale. It now uses the locale’s time format (e.g. localization.timeFormat) so the order tab shows time in 24h or 12h according to the language.

opw-5742177


sbel-odoo (approved):
LGTM 👍

adgu-odoo (commented):
robodoo r+

PR #246272, opened 2026-01-29
[FIX] point_of_sale: prevent infinite recursion in product models

pedrambiria:
Before this commit, if a getter was defined in both product.product and product.template models, and the getter returned a falsy value (like false or an empty string), it could lead to infinite recursion when accessing the property. This was due to the way the models fell back to each other to retrieve the property value.

opw-5886320

---


adgu-odoo (commented):
robodoo r+

PR #246274, opened 2026-01-29
[IMP] point_of_sale: display child categories in parent search

sjai-odoo:
In this commit:
------------------
- Introduced a new stored field `complete_name` to ensure proper search behavior
  based on the full hierarchical category name.
- Configured `complete_name` as the display name for `pos.category` by setting
  it as the model’s `_rec_name`, ensuring accurate results in POS category
  dropdown searches.
- We added a default increasing sequence because the view uses a `handle widget`
  which sequences records based on `sequence → id`. However, the backend
  `_order` follows `sequence → name`. Since these differ, the ordering was
  inconsistent in view and while we search somewhere. The default sequence
  ensures consistent ordering everywhere.

Task: 5882340

### DIFF
--- a/addons/point_of_sale/models/pos_category.py
+++ b/addons/point_of_sale/models/pos_category.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from typing import List, Tuple
 import random
 
 from odoo import api, fields, models, _
@@ -12,6 +11,7 @@ class PosCategory(models.Model):
     _name = 'pos.category'
     _description = "Point of Sale Category"
     _inherit = ['pos.load.mixin']
+    _rec_name = 'complete_name'
     _order = "sequence, name"
 
     @api.constrains('parent_id')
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/models/pos_category.py
```diff
@@ -48,15 +50,18 @@ def _load_pos_data_domain(self, data, config):
...
-        """ Returns a list representing the hierarchy of the categories. """
-        self.ensure_one()
-        return (self.parent_id._get_hierarchy() if self.parent_id else []) + [(self.name or '')]
+    @api.depends('name', 'parent_id.complete_name')
+    def _compute_complete_name(self):
+        for category in self:
+            if category.parent_id:
+                category.complete_name = '%s / %s' % (category.parent_id.complete_name, category.name)
+            else:
+                category.complete_name = category.name
```
paso-odoo: why removed? `_get_hierarchy`?

addons/point_of_sale/models/pos_category.py
```diff
@@ -48,15 +50,18 @@ def _load_pos_data_domain(self, data, config):
...
 
     @api.depends('parent_id')
     def _compute_display_name(self):
         for cat in self:
-            cat.display_name = " / ".join(cat._get_hierarchy())
+            cat.display_name = cat.complete_name
```
paso-odoo: Is is required? now?

sjai-odoo:
@robodoo retry

PR #246277, opened 2026-01-29
[IMP] stock: improve search filter order in stock move reports

srap-odoo:
When filtering stock moves, users generally search for the `Product` more often than the `Location`, but `Location` appeared as the default filter, which led to incorrect filtering. This change reorders the filters in both `Moves History`
and `Moves Analysis` reports to show `Product` before `Location`.

task-5481588

### DIFF
--- a/addons/stock/views/stock_move_line_views.xml
+++ b/addons/stock/views/stock_move_line_views.xml
@@ -127,8 +127,8 @@
         <field name="model">stock.move.line</field>
         <field name="arch" type="xml">
             <search string="Stock Moves">
-                <field name="location_id" string="Location" filter_domain="['|',('location_id', 'ilike', self),('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
                 <field name="product_id"/>
+                <field name="location_id" string="Location" filter_domain="['|', ('location_id', 'ilike', self), ('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
                 <field name="picking_id" string="Transfer"/>
                 <field name="reference" string="Reference"/>
                 <field name="product_category_id" string="Category"/>
--- a/addons/stock/views/stock_move_views.xml
+++ b/addons/stock/views/stock_move_views.xml
@@ -384,8 +384,8 @@
             <field eval="3" name="priority"/>
             <field name="arch" type="xml">
                 <search string="Stock Moves">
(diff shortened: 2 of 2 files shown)


addons/stock/views/stock_move_line_views.xml
```diff
@@ -127,8 +127,8 @@
         <field name="model">stock.move.line</field>
         <field name="arch" type="xml">
             <search string="Stock Moves">
-                <field name="location_id" string="Location" filter_domain="['|',('location_id', 'ilike', self),('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
                 <field name="product_id"/>
+                <field name="location_id" string="Location" filter_domain="['|',('location_id', 'ilike', self),('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
```
dhha-odoo:
```suggestion
                <field name="location_id" string="Location" filter_domain="['|', ('location_id', 'ilike', self), ('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
```

I know this change only adjusts the order of the filter, but this is a good opportunity to improve the indentation as well.

addons/stock/views/stock_move_views.xml
```diff
@@ -384,8 +384,8 @@
             <field eval="3" name="priority"/>
             <field name="arch" type="xml">
                 <search string="Stock Moves">
-                    <field name="reference" string="Location" filter_domain="['|',('location_id', 'ilike', self),('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
                     <field name="product_id"/>
+                    <field name="reference" string="Location" filter_domain="['|',('location_id', 'ilike', self),('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
```
dhha-odoo:
```suggestion
                    <field name="reference" string="Location" filter_domain="['|', ('location_id', 'ilike', self), ('location_dest_id', 'ilike', self)]" groups="stock.group_stock_multi_locations"/>
```

dhha-odoo (changes requested):
Hello @srap-odoo 

Please improve the commit message. If there is `only a single change`, there’s no need to list it point-wise; a simple one-line message is sufficient.

Thanks.

dhha-odoo (approved):

PR #246278, opened 2026-01-29
[FIX] hr_recruitment: Rename "Initial Qualification" to "Qualification"

SaraBriki:
This is a small follow-up PR to the original PR to simply rename a stage label.

See https://github.com/odoo/enterprise/pull/105278

Task-ID: 5454691


Einar-eian (approved):
Thank you for your work, LGTM 🚀 😄
@robodoo r+

PR #246279, opened 2026-01-29
[REF] im_livechat: remove rating dependency from discuss

KevGerard:
Before this PR, `rating.mixin` and `rating.parent.mixin` were used in `livechat.channel` and `discuss.channel` for inheritance. This means discuss channel is closely related to `rating.mixin` and therefore by inference `mail.thread`.

This prevents us to go forward with the split of discuss and thread.

In this PR, we are removing those inheritance, because the rating is too generic/difficult for a simple rating at the end of a livechat and the reasons stated above. The rating and feedback are now stored in a selection directly on the `discuss.channel`.

task-5437285

---


addons/im_livechat/static/src/embed/common/feedback_panel/feedback_panel.js
```diff
@@ -24,6 +24,11 @@ export class FeedbackPanel extends Component {
         THANKS: "thanks",
     });
     RATING = RATING;
+    RATING_URL = {
+        GOOD: url(`/rating/static/src/img/rating_5.png`),
```
KevGerard: Still not sure whether those files are hosted even if rating module is not installed

addons/im_livechat/views/discuss_channel_views.xml
```diff
@@ -16,10 +16,10 @@
...
                     <separator/>
-                    <filter name="filter_session_rating_happy" domain="[('rating_ids', '!=', False), ('rating_avg', '&gt;=', 3.66)]" string="Happy"/>
-                    <filter name="filter_session_rating_neutral" domain="[('rating_ids', '!=', False), ('rating_avg', '&gt;=', 2.33), ('rating_avg', '&lt;', 3.66)]" string="Neutral"/>
-                    <filter name="fiter_session_rating_unhappy" domain="[('rating_ids', '!=', False), ('rating_avg', '&lt;', 2.33) ]" string="Unhappy"/>
-                    <filter name="filter_session_unrated" domain="[('rating_ids', '=', False)]" string="Unrated"/>
+                    <filter name="filter_session_rating_happy" domain="[('livechat_rating', '!=', False), ('livechat_rating', '=', 3)]" string="Happy"/>
+                    <filter name="filter_session_rating_neutral" domain="[('livechat_rating', '!=', False), ('livechat_rating', '=', 2)]" string="Neutral"/>
+                    <filter name="fiter_session_rating_unhappy" domain="[('livechat_rating', '!=', False), ('livechat_rating', '=', 1) ]" string="Unhappy"/>
```
seb-odoo: These can be simplified, checking False becomes useless as we already check for a specific value.

addons/im_livechat/demo/im_livechat_channel/im_livechat_session_1.xml
```diff
@@ -116,16 +116,7 @@
...
-            <field name="res_id" ref="im_livechat.livechat_channel_session_1"/>
-            <field name="res_model_id" ref="mail.model_discuss_channel"/>
-            <field name="message_id" ref="livechat_channel_session_1_rating_message"/>
-            <field name="rated_partner_id" ref="base.partner_admin"/>
-            <field eval="False" name="partner_id"/>
-            <field eval="True" name="consumed"/>
-        </record>
         <function model="discuss.channel" name="_apply_livechat_feedback"
-            eval="([ref('im_livechat.livechat_channel_session_1')], 5, 'Good Job')"/>
+            eval="([ref('im_livechat.livechat_channel_session_1')], 3, 'Good Job')"/>
```
seb-odoo: I'm still not sure, what do we gain exactly by changing the numbering?

It would probably be easier to keep it like before.

But do we actually need numbers? If they are just sentinel values, we might as well use a selection field.

addons/im_livechat/models/digest.py
```diff
@@ -16,16 +17,27 @@ class DigestDigest(models.Model):
 
     def _compute_kpi_livechat_rating_value(self):
         self._raise_if_not_member_of('im_livechat.im_livechat_group_manager')
-        channels = self.env['discuss.channel'].search([('channel_type', '=', 'livechat')])
         start, end, __ = self._get_kpi_compute_parameters()
         domain = [
-            ('create_date', '>=', start),
-            ('create_date', '<', end),
+            ("channel_type", "=", "livechat"),
+            ("livechat_rating", ">=", RATING_UNHAPPY),
```
seb-odoo: Rather than using ordering comparison, doesn't this more simply a rating must be set? So != 0 (or != not_rated)?

addons/im_livechat/models/digest.py
```diff
@@ -16,16 +17,27 @@ class DigestDigest(models.Model):
...
+            domain,
+            ["livechat_rating"],
+            ["__count"],
+        )
+        total_count = 0
+        happy_count = 0
+        for rating, count in ratings_data:
+            if rating == RATING_HAPPY:
+                happy_count += count
+            total_count += count
```
seb-odoo: Not sure if more or less efficient (that's probably irrelevant here) but with `_read_group` we usually put the value in dict format for easier usage, like:
```suggestion
        counts = {r: c for r, c in ratings_data}
        happy_count = counts.get(RATING_HAPPY, 0)
        total_count = sum(counts.values())
```
Or basically where `counts` is the old `ratings` from result of `rating_get_grades` and the rest of the code can be unchanged with no extra variables.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -175,7 +183,15 @@ class DiscussChannel(models.Model):
         string="Live Chat Session Failure",
     )
     livechat_is_escalated = fields.Boolean("Is session escalated", compute="_compute_livechat_is_escalated", store=True)
-    rating_last_text = fields.Selection(store=True)
+    livechat_rating = fields.Integer(aggregator="avg")
```
seb-odoo: If we do actually use `avg` on this field (?), changing the numbering would have an impact on the stats and therefore should not be done.

addons/im_livechat/views/discuss_channel_views.xml
```diff
@@ -64,10 +64,10 @@
                     <field name="livechat_channel_id" optional="hide"/>
                     <field name="duration" widget="float_time" options="{'displaySeconds': True}" optional="show"/>
                     <field name="message_count" string="Messages" optional="show"/>
-                    <field name="rating_last_text" string="Rating" decoration-danger="rating_last_text == 'ko'"
-                        decoration-warning="rating_last_text == 'ok'" decoration-success="rating_last_text == 'top'"
+                    <field name="livechat_rating_text" string="Rating" decoration-danger="livechat_rating_text == 'unhappy'"
+                        decoration-warning="livechat_rating_text == 'neutral'" decoration-success="livechat_rating_text == 'happy'"
                         widget="badge" optional="show"/>
-                    <field name="rating_last_feedback" string="Comment" optional="hide"/>
+                    <field name="livechat_rating_feedback" string="Comment" optional="hide"/>
```
seb-odoo: Isn't this the sole place where we display the value of `livechat_rating_feedback`?
Honestly it feels like it is not worth keeping this field, especially if the default is to hide it.
If the user adds a message, we might as well just post it as an actual message in the discussion.

addons/im_livechat/views/discuss_channel_views.xml
```diff
@@ -96,7 +96,7 @@
                                         </span>
                                     </span>
                                 </div>
-                                <field name="rating_last_image" string="Rating" widget="image" options='{"size": [40, 40], "img_class": "bg-transparent"}' class="ms-auto" invisible="not rating_last_image"/>
+                                <field name="livechat_rating_image" string="Rating" widget="image" options='{"size": [40, 40], "img_class": "bg-transparent"}' class="ms-auto" invisible="not livechat_rating_image"/>
```
seb-odoo: I wish we could remove `livechat_rating_image` too. To replace it we basically just need a simple image tag/widget where the URL is computed from the rating value, so it shouldn't be that hard?

Also: returning image as base64 means it is individually returned and store for each channel, and never cached, as opposed to having url towards a static file.

addons/im_livechat/tests/test_message.py
```diff
@@ -118,112 +117,6 @@ def test_chatbot_message_format(self):
             ],
         )
 
-    @users('emp')
-    def test_store_add_message(self):
```
seb-odoo, line 123: I'm not sure if we want to move this, it's annoying to install all the modules when we just want to test the livechat stuff.

addons/rating/__manifest__.py
```diff
@@ -22,6 +22,7 @@
             "rating/static/src/core/web/**/*",
         ],
         'web.assets_frontend': [
+            "rating/static/src/core/common/**/*",
```
seb-odoo: What's the goal? Our `common` should never be added to `frontend`.
We need a frontend specific folder if necessary as this one is very sensitive to size/loading time.

addons/test_discuss_full/__manifest__.py
```diff
@@ -18,6 +18,7 @@
         "mail",
         "mail_bot",
         "project_todo",
+        "rating",
```
seb-odoo: What's the point? Not against explicit, but tt should come automatically with many of the other modules already in the list.

addons/test_discuss_full/tests/test_store.py
```diff
@@ -0,0 +1,153 @@
...
+                    "password": self.password,
+                },
+            ]
+        )
+        settings = self.env["res.users.settings"]._find_or_create_for_user(self.users[1])
+        settings.livechat_username = "chuck"
+        self.maxDiff = None
+
+    @users("emp")
+    def test_store_add_message(self):
```
seb-odoo: I would keep this test where it was.

addons/im_livechat/tests/test_im_livechat_report.py
```diff
@@ -74,10 +74,14 @@ def test_im_livechat_report_operator(self):
         self.assertEqual(result[0]["time_to_answer:avg"], 7800 / 3600)
         self.assertEqual(int(result[0]['duration:avg']), 195)
         channel = self.env["discuss.channel"].search([("livechat_channel_id", "=", self.livechat_channel.id)])
-        rated_channel = channel.copy({"rating_last_value": 5})
+        rated_channel = channel.copy()
+        rated_channel._apply_livechat_feedback(3)
+
         self._create_message(rated_channel, self.operator, "2023-03-18 11:00:00")
+        # flush channel, rating and message before custom sql query
+        self.env.flush_all()
```
seb-odoo: This should be done in business code never in tests. And flush only what is necessary if possible.

addons/im_livechat/static/src/embed/common/feedback_panel/feedback_panel.js
seb-odoo: Could you confirm it?

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -51,9 +51,10 @@ class ImLivechatChannelMemberHistory(models.Model):
         aggregator="avg",
         store=True,
     )
-    rating_id = fields.Many2one("rating.rating", compute="_compute_rating_id", store=True)
-    rating = fields.Float(related="rating_id.rating")
-    rating_text = fields.Selection(string="Rating text", related="rating_id.rating_text")
+    rating = fields.Integer("Rating", related="channel_id.livechat_rating", aggregator="avg")
+    rating_text = fields.Selection(
+        related="channel_id.livechat_rating_text"
+    )
```
seb-odoo: could probably be a single line

addons/im_livechat/models/discuss_channel.py
```diff
@@ -235,6 +264,50 @@ def _compute_livechat_is_escalated(self):
...
+
+    @api.depends("livechat_rating")
+    def _compute_livechat_rating_image(self):
+        for channel in self:
+            if not channel.livechat_rating:
+                channel.livechat_rating_image = False
+                continue
+            image = self._rating_image_filepath(
+                channel.livechat_rating
+            )
```
seb-odoo: single line

addons/im_livechat/models/discuss_channel.py
```diff
@@ -235,6 +264,50 @@ def _compute_livechat_is_escalated(self):
...
+            neutral=RATING_NEUTRAL,
+            unhappy=RATING_UNHAPPY,
+        )
+
+    @api.depends("livechat_rating")
+    def _compute_livechat_rating_image(self):
+        for channel in self:
+            if not channel.livechat_rating:
+                channel.livechat_rating_image = False
+                continue
```
seb-odoo: could pre-fitler self to simplify the loop

addons/im_livechat/models/discuss_channel.py
```diff
@@ -235,6 +264,50 @@ def _compute_livechat_is_escalated(self):
...
+            image = self._rating_image_filepath(
+                channel.livechat_rating
+            )
+            try:
+                with file_open(image, 'rb', filter_ext=('.png',)) as f:
+                    channel.livechat_rating_image = base64.b64encode(f.read())
+            except OSError:
```
seb-odoo: I don't think we except an error in this case.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -1009,3 +1063,15 @@ def _add_next_step_message_to_store(self, chatbot_script_step):
 
     def _allow_invite_by_email(self):
         return self.channel_type == "livechat" or super()._allow_invite_by_email()
+
+    @api.model
+    def _rating_image_filepath(self, rating):
```
seb-odoo: I'd rather name it url rather than path, and include the root in it.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -175,7 +183,15 @@ class DiscussChannel(models.Model):
...
-    rating_last_text = fields.Selection(store=True)
+    livechat_rating = fields.Integer(aggregator="avg")
+    livechat_rating_feedback = fields.Text(string="Livechat Rating Feedback")
+    livechat_rating_image = fields.Binary(string="Livechat Rating Image", compute="_compute_livechat_rating_image")
+    livechat_rating_text = fields.Selection(selection=[
+        ("no_rating", "Not Rated Yet"),
+        ("unhappy", "Unhappy"),
+        ("neutral", "Neutral"),
+        ("happy", "Happy"),
+    ], string="Livechat Rating Text", compute="_compute_livechat_rating_text", compute_sql="_compute_sql_livechat_rating_text", compute_sudo=True)
```
seb-odoo: If we don't actually need avg on `livechat_rating` we could simply make it a selection and merge it with this field.

seb-odoo: I don't think we need sudo here?

addons/im_livechat/models/discuss_channel.py
```diff
@@ -204,6 +224,15 @@ class DiscussChannel(models.Model):
...
 
     def write(self, vals):
+        if (
+            "livechat_rating" in vals or "livechat_rating_feedback" in vals
+        ) and not self.env.user.has_group("base.group_system"):
+            user, guest = self.env["res.users"]._get_current_persona()
+            if self.filtered(
+                lambda c: user not in c.livechat_customer_partner_ids.user_ids
+                and guest not in c.livechat_customer_guest_ids
+            ):
```
seb-odoo: could use self_member_id -> member_type

addons/im_livechat/models/discuss_channel.py
```diff
@@ -204,6 +224,15 @@ class DiscussChannel(models.Model):
     )
 
     def write(self, vals):
+        if (
+            "livechat_rating" in vals or "livechat_rating_feedback" in vals
+        ) and not self.env.user.has_group("base.group_system"):
```
seb-odoo, line 242: Only if the value is changed, no need to check when writing the existing value.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -235,6 +264,50 @@ def _compute_livechat_is_escalated(self):
...
+        for channel in self:
+            if channel.livechat_rating == RATING_UNHAPPY:
+                rating_text = "unhappy"
+            elif channel.livechat_rating == RATING_NEUTRAL:
+                rating_text = "neutral"
+            elif channel.livechat_rating == RATING_HAPPY:
+                rating_text = "happy"
+            else:
+                rating_text = "no_rating"
+            channel.livechat_rating_text = rating_text
```
seb-odoo: overkill variable, might as well directly write on the field

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +249,39 @@ def _compute_nbr_channel(self):
...
+                ],
+                ["livechat_channel_id", "livechat_rating"],
+                ["__count"],
+            )
+        )
+        channel_data = defaultdict(lambda: [0, 0])
+        for parent_id, rating, count in read_group_data:
+            if rating == RATING_HAPPY:
+                channel_data[parent_id.id][1] += count
+            channel_data[parent_id.id][0] += count
```
seb-odoo: we might be able to simplify here too

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +249,39 @@ def _compute_nbr_channel(self):
...
+        for channel in self:
+            rating_count, happy_count = channel_data.get(channel.id, (0, 0))
```
seb-odoo: get is not necessary with default dict, use `[]`

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +249,39 @@ def _compute_nbr_channel(self):
...
             record.nbr_channel = channel_count.get(record.id, 0)
 
+    @api.depends("channel_ids.livechat_rating")
+    def _compute_rating_percentage_satisfaction(self):
+        read_group_data = (
+            self.env["discuss.channel"]
+            ._read_group(
+                [
+                    Domain("livechat_channel_id", "in", self.ids)
+                    & Domain("livechat_rating", ">=", RATING_UNHAPPY)
```
seb-odoo: `!= 0 `?

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +249,39 @@ def _compute_nbr_channel(self):
...
+                    & (
+                        Domain(
+                            "livechat_end_dt",
+                            ">=",
+                            fields.Datetime.now() - timedelta(days=self._rating_satisfaction_days),
```
seb-odoo: We might as well hard-code the value here as it no longer really comes from the mixin.

addons/im_livechat/report/im_livechat_report_channel.py
```diff
@@ -119,12 +119,12 @@ def _select(self) -> SQL:
...
-                NULLIF(C.rating_last_value, 0) AS rating,
+                NULLIF(C.livechat_rating, 0) as rating,
                 CASE
-                    WHEN C.rating_last_value = 1 THEN 'Unhappy'
-                    WHEN C.rating_last_value = 5 THEN 'Happy'
-                    WHEN C.rating_last_value = 3 THEN 'Neutral'
-                    ELSE null
+                    WHEN C.livechat_rating = 1 THEN 'Unhappy'
+                    WHEN C.livechat_rating = 2 THEN 'Neutral'
+                    WHEN C.livechat_rating = 3 THEN 'Happy'
```
seb-odoo: Note: in some places you do substitute the numbers for constant but not here

seb-odoo (commented):
This looks promising.

addons/im_livechat/views/discuss_channel_views.xml
KevGerard: I think this is already done somewhere else, I will try to reuse it here.

addons/im_livechat/models/discuss_channel.py
KevGerard: The field is used in graph and pivot views I believe but always in conjunction with the `rating_percentage` which was also changed to take this new scale into account. So there shouldn't be any difference on the frontend side. I am not 100% sure this aggregator is useful though...

addons/im_livechat/demo/im_livechat_channel/im_livechat_session_1.xml
KevGerard: Apart from more diff, not a lot. The point was to be clear it is a 3 level rating and not a 5-star rating. I am honestly not against reverting this part.

On the numbers point, we do need those to calculate the % of satisfaction on the livechat_channel side (and I think it takes less space in the DB).

seb-odoo: I don't think it will take less space in db if it's stored as `integer`, it will take 4 bytes regardless.

`3` vs `5` would actually optimize space only if it was stored as `bit`, where `3` could fit in `bit 2` and `5` could fit in `bit 3`.

seb-odoo: I would say keep the old numbering for now, it just makes the PR much easier to review.

addons/im_livechat/models/discuss_channel.py
seb-odoo: Maybe @tsm-odoo would know/remember

tsm-odoo: It's needed. In the pivot view, the aggregator defaults to `:sum` otherwise. Widget has been adapted so it doesn't change anything regarding the pivot view.  TBH I don't feel like changing the numbering system adds much value but it's up to you to decide wether we want to do it or not (basically it just simplifies 3 domains and adds quite a lot of diff).

addons/im_livechat/static/src/embed/common/feedback_panel/feedback_panel.js
KevGerard: confirmed

addons/im_livechat/models/discuss_channel.py
KevGerard: It is required to be able to use `compute_sql`

addons/im_livechat/tests/test_message.py
KevGerard: Honestly I really don't see what this test is testing, it looked like it was testing a rating linked to a message and not to the livechat session. Reason why it did not make sense to keep it in `im_livechat` since we are removing the dependency on `rating` there. If interesting, we might split this test in two parts: one checking the addition of the message in the store (in im_livechat) and the other one here with the rating stuff.

I honestly might have missed the point of this test, the name is completely reflecting what the test is doing...

addons/test_discuss_full/__manifest__.py
KevGerard: With the `test_store_add_message` moved, we have an explicit call to rating therefore I thought it might be a good idea to show this hard dependency in the manifest.

addons/im_livechat/tests/test_im_livechat_report.py
KevGerard: I'm not sure I follow the reason why... The flush here is necessary between the setup of the test and the actual thing that we are testing, irl this flow does not exist and would be in 2 separate transactions anyway. On top of that `formatted_read_group` is an internal method of the base model, I am not sure adding  a flush in this method makes sense. I would agree with you in case we were using a higher level function containing the `formatted_read_group` itself.

addons/im_livechat/views/discuss_channel_views.xml
KevGerard: The message is already posted anyway, so indeed, I do not see a lot of reason to keep this field neither...

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -156,15 +155,6 @@ def _compute_help_status(self):
...
                 history.help_status = "provided"
 
-    @api.depends("channel_id.rating_ids")
-    def _compute_rating_id(self):
-        agent_histories = self.filtered(lambda h: h.livechat_member_type in ("agent", "bot"))
-        (self - agent_histories).rating_id = None
-        for history in agent_histories:
-            history.rating_id = history.channel_id.rating_ids.filtered(
-                lambda r: r.rated_partner_id == history.partner_id
-            )[:1]  # Live chats only allow one rating.
```
seb-odoo, line 166: Removing this feels wrong, we need to compute the rating per agent/bot, not the rating of channel regardless of member type.

visitors shouldn't appear as members being rated

addons/im_livechat/tests/test_message.py
seb-odoo: It's a rather low level test just testing the result of store add on message when livechat is installed.

The test should simply not be moved in the context of this PR.

There are other subtleties like `user_livechat_username` being returned instead of `name` and stuff like that.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -174,7 +176,18 @@ class DiscussChannel(models.Model):
         string="Live Chat Session Failure",
     )
     livechat_is_escalated = fields.Boolean("Is session escalated", compute="_compute_livechat_is_escalated", store=True)
-    rating_last_text = fields.Selection(store=True)
+    livechat_rating = fields.Selection(selection=[
+        ("0", "Not Rated Yet"),
+        ("1", "Unhappy"),
+        ("3", "Neutral"),
+        ("5", "Happy"),
+    ], string="Livechat Rating Text", default="0")
```
seb-odoo: Now that it is a (stored) selection field, it's better if there is simply no value (NULL) when it is not rated, which will be the case for most channels.

This also makes the upgrade far easier as we don't need to go and write "0" on all existing channels.

It's usually better when (new) selection fields don't have a default and use the NULL as the default case, especially when it does represent something empty.

The only drawback would be that if this selection is presented in the UI, it would not have a clean selection item for the 0 case, but that should be mostly irrelevant here.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -191,6 +204,10 @@ class DiscussChannel(models.Model):
         """,
         "Looking for help date should only be set if the channel is looking for help and must be empty otherwise.",
     )
+    _livechat_rating_constraint = models.Constraint(
+        "CHECK(channel_type != 'livechat' OR livechat_rating IS NOT NULL)",
+        "Livechat session should have a rating or '0' if not rated yet."
+    )
```
seb-odoo, line 222: constraint is missing the part where non-livechat channel should not have a rating.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -203,6 +220,15 @@ class DiscussChannel(models.Model):
...
     def write(self, vals):
+        if (
+            self.filtered(
+                lambda c: (
+                    "livechat_rating" in vals and c.livechat_rating != vals["livechat_rating"]
+                )
+                and c.self_member_id.livechat_member_type != "visitor"
+            )
+        ) and not self.env.user.has_group("base.group_system"):
+            raise UserError(self.env._("Only customers can rate a livechat session."))
```
seb-odoo: consistency: "live chat" (should always be 2 words now) and "session" should always be "conversation" now

addons/im_livechat/models/discuss_channel.py
```diff
@@ -174,7 +176,18 @@ class DiscussChannel(models.Model):
...
     livechat_is_escalated = fields.Boolean("Is session escalated", compute="_compute_livechat_is_escalated", store=True)
-    rating_last_text = fields.Selection(store=True)
+    livechat_rating = fields.Selection(selection=[
+        ("0", "Not Rated Yet"),
+        ("1", "Unhappy"),
+        ("3", "Neutral"),
+        ("5", "Happy"),
+    ], string="Livechat Rating Text", default="0")
+    livechat_rating_value = fields.Integer(
+        aggregator="avg",
```
seb-odoo, line 190: As discussed with @maki-odoo I think we need to remove the aggregator from the 0...5 value field, and actually make a new compute that holds the % satisfaction (with the math of the widget) so that it computes and displays properly in all cases.

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -51,9 +51,8 @@ class ImLivechatChannelMemberHistory(models.Model):
         aggregator="avg",
         store=True,
     )
-    rating_id = fields.Many2one("rating.rating", compute="_compute_rating_id", store=True)
-    rating = fields.Float(related="rating_id.rating")
-    rating_text = fields.Selection(string="Rating text", related="rating_id.rating_text")
+    rating = fields.Selection(related="channel_id.livechat_rating")
+    rating_value = fields.Integer(related="channel_id.livechat_rating_value", aggregator="avg")
```
seb-odoo: same 2 remarks:
- should be rating for the agent/bot, not rating of the channel (see comment on the compute below)
- should not have aggregator on dummy value but on actual % satisfaction (see other field)

addons/im_livechat/report/im_livechat_report_channel.py
```diff
@@ -40,9 +40,12 @@ class Im_LivechatReportChannel(models.Model):
...
     lang_id = fields.Many2one("res.lang", related="channel_id.livechat_lang_id", string="Language", readonly=True)
-    rating = fields.Integer('Rating', aggregator="avg", readonly=True)
-    # TODO DBE : Use Selection field - Need : Pie chart must show labels, not keys.
-    rating_text = fields.Char('Satisfaction Rate', readonly=True)
+    rating = fields.Selection(selection=[
+        ("1", "Unhappy"),
+        ("3", "Neutral"),
+        ("5", "Happy"),
+    ], string='Satisfaction Rate', required=True, readonly=True)
+    rating_value = fields.Integer('Rating', aggregator="avg", readonly=True)
```
seb-odoo: same remark about aggregator, should probably be on % value

addons/im_livechat/report/im_livechat_report_channel_views.xml
```diff
@@ -13,7 +13,7 @@
                     <field name="percentage_of_calls" type="measure" widget="percentage"/>
                     <field name="time_to_answer" string="Response Time" type="measure" widget="float_time"/>
                     <field name="duration" type="measure"/>
-                    <field name="rating" string="Rating (%)" type="measure" widget="im_livechat.rating_percentage"/>
+                    <field name="rating_value" string="Rating (%)" type="measure" widget="im_livechat.rating_percentage"/>
```
seb-odoo: same remark everywhere: remove this widget and have a field that actually contains the correct %

addons/im_livechat/report/im_livechat_report_channel_views.xml
```diff
@@ -93,7 +93,7 @@
                         <filter name="group_by_operator" string="Agent" domain="[]" context="{'group_by': 'partner_id'}"/>
                         <filter name="group_by_agent_requesting_help" domain="[]" context="{'group_by': 'agent_requesting_help_history'}"/>
                         <filter name="group_by_agent_providing_help" domain="[]" context="{'group_by': 'agent_providing_help_history'}"/>
-                        <filter name="group_by_rating" string="Rating" domain="[]" context="{'group_by':'rating_text'}"/>
+                        <filter name="group_by_rating" string="Rating" domain="[]" context="{'group_by':'rating'}"/>
```
seb-odoo, line 96: I believe there was a specific spec here where we wanted to see the actual name in the UI and not the magic numbers, so `rating_text` is probably still useful, unless that is now working fine with a selection field (I didn't test to verify)?

addons/im_livechat/views/discuss_channel_views.xml
```diff
@@ -31,7 +31,7 @@
                     <group>
                         <filter name="group_by_channel" string="Channel" domain="[]" context="{'group_by':'livechat_channel_id'}"/>
                         <filter name="group_by_agent" string="Agent" domain="[]" context="{'group_by':'livechat_agent_partner_ids'}"/>
-                        <filter name="group_by_rating" string="Rating" domain="[]" context="{'group_by':'rating_last_text'}"/>
+                        <filter name="group_by_rating" string="Rating" domain="[]" context="{'group_by':'livechat_rating'}"/>
```
seb-odoo, line 34: same remark about wanting to see the actual text (but maybe it is ok as a selection field?)

addons/im_livechat/views/discuss_channel_views.xml
```diff
@@ -97,7 +96,17 @@
...
+                                <field name="livechat_rating_value" invisible="1"/>
+                                <img
+                                    t-if="record.livechat_rating_value and record.livechat_rating_value.raw_value"
+                                    class="ms-auto bg-transparent"
+                                    loading="lazy"
+                                    height="40"
+                                    width="40"
+                                    style="max-width: 40px, max-height: 40px"
+                                    t-attf-src="/rating/static/src/img/rating_#{record.livechat_rating_value.raw_value}.png"
+                                    t-att-alt="record.livechat_rating_value.raw_value"
```
seb-odoo: can you remind me what was the reason to not use the image widget again?

addons/im_livechat/models/discuss_channel.py
KevGerard: Completely agreed, the only reason it is there is to show the text in the graph views. I might investigate another solution if we prefer not to have this option in the selection.

addons/im_livechat/report/im_livechat_report_channel_views.xml
KevGerard: It is the reason why I switched to the selection so that now it directly shows the correct label in all situation.

addons/im_livechat/views/discuss_channel_views.xml
KevGerard: It was related to the cache, with the widget the image had to be read by the backend for each rating (since it is related to the `rating.rating` model). With this solution, we make use of the cache since all similar ratings target the same url.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -663,38 +718,23 @@ def _apply_livechat_feedback(self, rate, reason=None):
...
-                "res_model_id": self.env["ir.model"]._get_id("discuss.channel"),
-                # sudo: res.partner - visitor can access the agent record to add a rating
-                "rated_partner_id": self.sudo().livechat_agent_partner_ids[:1].id,
-                "partner_id": user.partner_id.id,
-            }
-            rating_sudo = self.env["rating.rating"].sudo().create(rating_values)
+        # sudo: discuss.channel - visitor giving a rating to the session is allowed
+        self.sudo().write({
+            "livechat_rating": str(rate),
+        })
```
seb-odoo:
```suggestion
        self.sudo().livechat_rating = str(rate)
```
And do we actually need `str` cast?

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +247,37 @@ def _compute_nbr_channel(self):
...
+            self.env["discuss.channel"]
+            ._read_group(
+                [
+                    Domain("livechat_channel_id", "in", self.ids)
+                    & Domain("livechat_rating", "!=", False)
+                    & (
+                        Domain(
+                            "create_date",
+                            ">=",
+                            fields.Datetime.now() - timedelta(days=14),
```
seb-odoo: I believe this can use "-14d" syntax (note: a recent PR in progress moves it to 30d I believe).

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -247,6 +247,37 @@ def _compute_nbr_channel(self):
...
+        )
+        channel_data = defaultdict(dict)
+        for parent_id, rating, count in read_group_data:
+            channel_data[parent_id.id][rating] = count
+        for channel in self:
+            rating_data = channel_data[channel.id]
+            rating_count = sum(rating_data.values())
+            channel.rating_count = rating_count
+            channel.rating_percentage_satisfaction = (
+                rating_data.get("5", 0) * 100 / rating_count if rating_count else 0
```
seb-odoo: This doesn't match `_compute_livechat_rating_percentage`, should do -1 on the value as there is no 0 value.

Could it be possible to not duplicate the code, and simply call the percentage on all the channels?

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -51,9 +52,24 @@ class ImLivechatChannelMemberHistory(models.Model):
...
-    rating = fields.Float(related="rating_id.rating")
-    rating_text = fields.Selection(string="Rating text", related="rating_id.rating_text")
+    rating = fields.Selection(
+        [
+            ("1", "Unhappy"),
+            ("3", "Neutral"),
+            ("5", "Happy"),
+        ],
+        compute="_compute_rating",
+        search="_search_rating",
```
seb-odoo: I don't think stored compute need a search method.

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -124,6 +140,48 @@ def _compute_display_name(self):
...
+                channel_agent_history_ids |= channel_agent
+            else:
+                channel_agent_history_ids |= channel.livechat_bot_history_ids
+        return [("id", "in", channel_agent_history_ids.ids)]
+
+    @api.depends("rating")
+    def _compute_rating_percentage(self):
+        for member_history in self:
+            member_history.rating_percentage = (
+                (int(member_history.rating) - 1) / 4 * 100 if member_history.rating else False
```
seb-odoo: Repeated again, it would be nice to re-use code, maybe move it to utils function?

addons/im_livechat/report/im_livechat_report_channel.py
```diff
@@ -40,9 +40,18 @@ class Im_LivechatReportChannel(models.Model):
...
-    # TODO DBE : Use Selection field - Need : Pie chart must show labels, not keys.
-    rating_text = fields.Char('Satisfaction Rate', readonly=True)
+    rating = fields.Selection(
+        selection=[
+            ("1", "Unhappy"),
+            ("3", "Neutral"),
+            ("5", "Happy"),
+        ],
+        string="Satisfaction Rate",
+        required=True,
```
seb-odoo: Does it really do anything to set required on a reporting model?

addons/im_livechat/report/im_livechat_report_channel.py
```diff
@@ -119,13 +128,8 @@ def _select(self) -> SQL:
...
                 C.country_id,
-                NULLIF(C.rating_last_value, 0) AS rating,
-                CASE
-                    WHEN C.rating_last_value = 1 THEN 'Unhappy'
-                    WHEN C.rating_last_value = 5 THEN 'Happy'
-                    WHEN C.rating_last_value = 3 THEN 'Neutral'
-                    ELSE null
-                END as rating_text,
+                NULLIF(C.livechat_rating, '0') as rating,
+                (NULLIF(C.livechat_rating::double precision, 0.0) - 1.0) * 100.0 / 4.0 as rating_percentage,
```
seb-odoo: The query version of it is also repeated a lot

addons/im_livechat/static/tests/mock_server/livechat_mock_server.js
```diff
@@ -102,30 +102,15 @@ registerRoute("/im_livechat/feedback", feedback);
...
-    if (channel.rating_ids.length === 0) {
-        RatingRating.create(values);
-    } else {
-        RatingRating.write([channel.rating_ids[0]], values);
-    }
-    [channel] = DiscussChannel.search_read([["id", "=", channel_id]]);
-    return channel.rating_ids[0];
+    DiscussChannel.write([channel_id], {
+        livechat_rating: rate,
+    });
```
seb-odoo: could be single line

seb-odoo (commented):
This looks good enough to be marked a ready now.
I still have to test a little bit.

@tsm-odoo can make the final check

addons/im_livechat/report/im_livechat_report_channel_views.xml
```diff
@@ -13,7 +13,7 @@
                     <field name="percentage_of_calls" type="measure" widget="percentage"/>
                     <field name="time_to_answer" string="Response Time" type="measure" widget="float_time" options="{'show_seconds': true}"/>
                     <field name="duration" type="measure"/>
-                    <field name="rating" string="Rating (%)" type="measure" widget="im_livechat.rating_percentage"/>
+                    <field name="rating_percentage" type="measure"/>
```
maki-odoo, line 16: As `rating_percentage` is an aggregator field, I think it should be included in the measures automatically.
[See here](https://github.com/odoo/documentation/blob/b9a9c3c70608a0ac5199add919bfccde80a849cd/content/developer/reference/user_interface/view_architectures.rst?plain=1#L3087)

addons/im_livechat/models/discuss_channel.py
KevGerard: yes, I tried at some point but unfortunately it throws an error

addons/im_livechat/models/im_livechat_channel.py
KevGerard: For the sake of keeping it a REF PR I left it like this. 

Those percentages (on discuss channel and livechat channel) are not calculating the same thing.
The former is making an average of bad = 0%, neutral = 50% and good = 100%, while the latter is calculating how many good reviews on the number of reviews.

It definitely could be changed but would then change the behaviour.

@seb-odoo what do you think?

addons/im_livechat/models/im_livechat_channel_member_history.py
KevGerard: Bad cleaning indeed...

addons/im_livechat/report/im_livechat_report_channel_views.xml
KevGerard: You are right, though it was the case already before, this might have been there for more clarity?

maki-odoo: Might be but not in the previous case. If we use a widget or manipulate the original behaviour for an aggregated field in some way, for example making the field invisible, we need to put it in the report view.
But there is no harm if someone put it in the view even when it's not needed I guess.

addons/im_livechat/models/im_livechat_channel.py
seb-odoo: It doesn't make sense to me to compute it differently in different places cc @tsm-odoo

tsm-odoo: I agree, both are rating percentages, so I don't see why we should add such distinctions.

seb-odoo: Note that counting good / total basically considers middle the same as bad, which seems fundamentally misleading.

seb-odoo: If any case, we need the `-1` as the 0 value doesn't exist. Even if after that we do an average of the value rather than counting good.

addons/im_livechat/models/digest.py
```diff
@@ -16,16 +16,24 @@ class DigestDigest(models.Model):
...
         domain = [
-            ('create_date', '>=', start),
-            ('create_date', '<', end),
+            ("channel_type", "=", "livechat"),
+            ("livechat_rating", "!=", False),
+            ("create_date", ">=", start),
+            ("create_date", "<", end),
         ]
-        ratings = channels.rating_get_grades(domain)
+        ratings_data = dict(
```
tsm-odoo: ⛏️ 

count_by_rating?

addons/im_livechat/models/discuss_channel.py
```diff
@@ -191,6 +208,21 @@ class DiscussChannel(models.Model):
...
     )
+    _livechat_rating_constraint = models.Constraint(
+        """
+        CHECK (
+            CASE
+                WHEN channel_type = 'livechat'
+                THEN livechat_rating IN ('1', '3', '5') OR livechat_rating IS NULL
+                ELSE livechat_rating IS NULL
+            END
+        )
```
tsm-odoo, line 216:
```suggestion
        CHECK (
            livechat_rating IS NULL
            OR (channel_type = 'livechat' AND livechat_rating IN ('1', '3', '5'))
        )
```

I think this simplification is equivalent?

addons/im_livechat/models/discuss_channel.py
```diff
@@ -241,6 +282,38 @@ def _compute_livechat_is_escalated(self):
...
+            return False
+        return (float(rating) - 1.0) * 100.0 / 4.0
+
+    @api.model
+    def _rating_selection_to_percentage_sql(self, rating_field):
+        return SQL(
+            """
+            CASE WHEN %(rating_field)s IS NOT NULL
+            THEN ((%(rating_field)s::double precision) - 1) * 100.0 / 4.0
+            END
```
tsm-odoo: I don't know if we wan't to rely on this behavior, but in SQL, any arithmetic with NULL returns NULL. So we could just write it this way, and let SQL null handling do the rest.

```py
((%(rating_field)s::double precision) - 1) * 100.0 / 4.0
```

addons/im_livechat/models/discuss_channel.py
```diff
@@ -241,6 +282,38 @@ def _compute_livechat_is_escalated(self):
...
+        )
+
+    @api.depends("livechat_rating")
+    def _compute_livechat_rating_percentage(self):
+        for channel in self:
+            channel.livechat_rating_percentage = (
+                self._rating_selection_to_percentage(channel.livechat_rating)
+                if channel.livechat_rating
+                else False
+            )
```
tsm-odoo:
```suggestion
            channel.livechat_rating_percentage = self._rating_selection_to_percentage(channel.livechat_rating)
```
`rating_to_percentage` already handles falsy ratings gracefully.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -664,49 +732,35 @@ def _apply_livechat_feedback(self, rate, reason=None):
...
-                "res_model_id": self.env["ir.model"]._get_id("discuss.channel"),
-                # sudo: res.partner - visitor can access the agent record to add a rating
-                "rated_partner_id": self.sudo().livechat_agent_partner_ids[:1].id,
-                "partner_id": user.partner_id.id,
-            }
-            rating_sudo = self.env["rating.rating"].sudo().create(rating_values)
+        rated_partner = self.sudo().livechat_agent_partner_ids[:1]
+        # sudo: discuss.channel - visitor giving a rating to the session is allowed
+        self.sudo().write({"livechat_rating": str(rate)})
+        rating_url = f"/rating/static/src/img/rating_{rate}.png"
```
tsm-odoo, line 731: I guess that's ok-ish, but that's odd to remove the dependency to rating, only to keep using rating files. Might be worth it to duplicate the files and to put them in live chat? @seb-odoo

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -250,6 +250,42 @@ def _compute_nbr_channel(self):
...
+                [
+                    Domain("livechat_channel_id", "in", self.ids)
+                    & Domain("livechat_rating", "!=", False)
+                    & (
+                        Domain(
+                            "create_date",
+                            ">=",
+                            "-30d",
+                        )
+                    )
```
tsm-odoo: could probably be inlined?

addons/im_livechat/models/im_livechat_channel.py
KevGerard: Makes sense for me as well (it just changes compared to before), anyway it's updated (and also in the digest, so we are consistent everywhere)

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -250,6 +250,42 @@ def _compute_nbr_channel(self):
...
+                ["livechat_channel_id", "livechat_rating"],
+                ["__count"],
+            )
+        )
+        channel_data = defaultdict(dict)
+        for parent_id, rating, count in read_group_data:
+            channel_data[parent_id.id][rating] = count
+        rating_to_percentage = self.env["discuss.channel"]._rating_selection_to_percentage
+        for channel in self:
+            rating_data = channel_data[channel.id]
```
tsm-odoo: Would be nice to name the dict in a way that makes it easy to understand what's in it :)

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -124,6 +139,33 @@ def _compute_display_name(self):
...
+            history.rating = history.channel_id.livechat_rating
+
+    @api.depends("rating")
+    def _compute_rating_percentage(self):
+        for member_history in self:
+            member_history.rating_percentage = (
+                self.env["discuss.channel"]._rating_selection_to_percentage(member_history.rating)
+                if member_history.rating
+                else False
+            )
```
tsm-odoo, line 160: Guard can be removed.

addons/im_livechat/models/discuss_channel.py
seb-odoo: I don't think it's worth duplicating the files if it just works like this.

addons/im_livechat/models/im_livechat_channel.py
```diff
@@ -250,6 +250,37 @@ def _compute_nbr_channel(self):
...
+                    & Domain("create_date", ">=", "-30d")
+
+                ],
+                ["livechat_channel_id", "livechat_rating"],
+                ["__count"],
+            )
+        )
+        ratings_count_by_channel = defaultdict(dict)
+        for parent_id, rating, count in read_group_data:
+            ratings_count_by_channel[parent_id.id][rating] = count
```
tsm-odoo:
```suggestion
        for livechat_channel, rating, count in read_group_data:
            ratings_count_by_channel[livechat_channel][rating] = count
```
Easiet to follow when using more specific naming.

tsm-odoo (approved):
@robodoo delegate+

addons/im_livechat/models/im_livechat_channel_member_history.py
```diff
@@ -124,6 +139,31 @@ def _compute_display_name(self):
                 name = history.partner_id.display_name
             history.display_name = name or self.env._("Unknown")
 
+    @api.depends("channel_id.livechat_rating", "channel_id.livechat_agent_partner_ids")
+    def _compute_rating(self):
+        agent_histories = self.filtered(
+            lambda h: h.id in h.channel_id.livechat_agent_history_ids.ids
+            and h.livechat_member_type == "agent"
+            or not h.channel_id.livechat_agent_partner_ids
+            and h.livechat_member_type == "bot"
```
KangOl: Missing parentheses. Please be explicit about your condition.
What do you mean?
- `A and (B or C) and D`
- `(A and B) or (C and D)`
- `A and (B or (C and D))`

tsm-odoo: `(A and B) or (C and D)` which is correct but I agree that parantheses would help.

KevGerard: Indeed, second one, it is not like we have a choice. But if that helps, I will add the parenthesis... Better be explicit like they say.

addons/im_livechat/models/discuss_channel.py
```diff
@@ -241,6 +279,32 @@ def _compute_livechat_is_escalated(self):
         for channel in self:
             channel.livechat_is_escalated = len(channel.livechat_agent_history_ids) > 1
 
+    @api.model
+    def _rating_selection_to_percentage(self, rating):
+        if not rating:
+            return False
+        return (float(rating) - 1.0) * 100.0 / 4.0
```
seb-odoo, line 286: Does it actually make a difference to return False vs 0 on a float field?

Won't non-rated effectively contribute as being rated 0?

But I guess we can't really solve that problem if float fields don't allow NULL values.

KevGerard: From what I saw during development, the False values were filtered out during the aggregation calculations, reason why I put it explicit here.

PR #246287, opened 2026-01-29
[FIX] account_peppol: id error for branch company in demo mode

jedel-odoo:
Steps to reproduce in demo:
1.  Activate Peppol in the main company for both sending and receiving;
2. Create a branch company (with same VAT number as the main company);
3. Activate Peppol in the branch and select "Send from parent company";
4. Error message: "This id_client is already used on another user.".

In demo edi_mode, when trying to connect a branch company with the same VAT than the parent company, the demo function took the same id_client for every company, which violates the unique constraint on id_client.

task-5888203

---

### DIFF
--- a/addons/account_peppol/tools/demo_utils.py
+++ b/addons/account_peppol/tools/demo_utils.py
@@ -103,7 +103,7 @@ def _mock_check_peppol_participant_exists(func, self, *args, **kwargs):
 
 def _mock_create_connection(func, self, *args, **kwargs):
     peppol_identifier, _db_uuid, company = args
-    dummy_response = {'id_client': 'demo4peppol', 'refresh_token': 'demo', 'peppol_state': 'receiver'}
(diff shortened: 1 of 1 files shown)


addons/account_peppol/tools/demo_utils.py
```diff
@@ -104,6 +104,8 @@ def _mock_check_peppol_participant_exists(func, self, *args, **kwargs):
 def _mock_create_connection(func, self, *args, **kwargs):
     peppol_identifier, _db_uuid, company = args
     dummy_response = {'id_client': 'demo4peppol', 'refresh_token': 'demo', 'peppol_state': 'receiver'}
+    if self.use_parent_connection:
+        dummy_response['id_client'] = f'demo4peppol{company.id}'
```
clbr-odoo:
```suggestion
    dummy_response = {'id_client': f'demo4peppol_{company.id}', 'refresh_token': 'demo', 'peppol_state': 'receiver'}
```
I think you can just do it all the time tbh 🤔

clbr-odoo (approved):
@robodoo r+

PR #246292, opened 2026-01-29
[FIX] account: bank statement print action

agr-odoo:
In Bank Statement list view users may select a statement and download a pdf report. However, currently, the load of extra print options may raise an error.

Steps to reproduce:
- In Accounting Dashboard, from a bank journal card, 3 dots > Statements
- Select a line (it needs to have an id not present in `account.move`)
- Click 'Print' button

Issue:
Access error may occur, stating the record has been deleted or it is inaccessible. This happens because backend method `get_extra_print_items` is called on `account.move` with the id of the `account.bank.statement` On accessing the record fields we get the error

A solution is to avoid loading extra print item with loadExtraPrintItems if we are not in the `account.move` model

opw-5500506

### DIFF
--- a/addons/account/static/src/views/account_move_list/account_move_list_controller.js
+++ b/addons/account/static/src/views/account_move_list/account_move_list_controller.js
@@ -19,11 +19,14 @@ export class AccountMoveListController extends FileUploadListController {
     }
 
     get actionMenuProps() {
-        return {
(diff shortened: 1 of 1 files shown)


antoine162 (approved):
👍 
@robodoo r+

PR #246295, opened 2026-01-29
[FIX] base,web: prevent phone input collapse on mobile

pish-odoo:
Steps to reproduce:
1. Install `contacts`
2. Create an individual contact with a phone number
3. Try to edit the phone number on the mobile

Issue:
- Unable to edit in mobile view

Cause:
- On mobile devices, the utility buttons (Call, SMS, WhatsApp) inside the phone widget consume significant horizontal space. Due to the flex layout, this forces the actual phone input field to shrink too much, making it hard/impossible to view or edit the number.

Solution:
- Hide the text labels of the utility buttons to save space, showing only the icons
- make input's width 100% on mobile screens in partner form

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td align="center">
        Before
      </td>
      <td align="center">
       image
      </td>
    </tr>
  </tbody>
</table>

opw-5489021

### DIFF
--- a/addons/web/static/src/views/fields/phone/phone_field.scss
+++ b/addons/web/static/src/views/fields/phone/phone_field.scss
@@ -10,4 +10,9 @@ body:not(.o_touch_device) .o_field_phone {
 .o_phone_content small {
     overflow-wrap: normal;
     word-break: normal;
+
+    /* On small screens, hide the text labels inside the phone buttons */
+    @include media-breakpoint-down(md) {
+        display: none;
+    }
 }
--- a/addons/web/static/tests/views/fields/phone_field.test.js
+++ b/addons/web/static/tests/views/fields/phone_field.test.js
@@ -219,3 +219,21 @@ test("New record, fill in phone field, then click on call icon and save", async 
     expect(".o_field_widget[name=foo] input").toHaveValue("+12345678900");
     expect(`.o_form_status_indicator_buttons`).toHaveClass("invisible");
 });
+
+test.tags("mobile");
+test("PhoneField in form view shows only icon on mobile screens", async () => {
+    await mountView({
+        type: "form",
+        resModel: "partner",
+        arch: /* xml */ `
+            <form>
+                <sheet>
+                    <group>
+                        <field name="foo" widget="phone"/>
+                    </group>
+                </sheet>
+            </form>`,
+        resId: 1,
+    });
+    expect(".o_field_phone .o_phone_form_link small").not.toBeVisible();
+});
--- a/odoo/addons/base/views/res_partner_views.xml
+++ b/odoo/addons/base/views/res_partner_views.xml
@@ -138,11 +138,11 @@
                                     <field options="{'line_breaks': False}" widget="text" class="text-break d-block" name="name" default_focus="1" placeholder="e.g. Lumber Inc" invisible="not is_company" required="type == 'contact'"/>
                                     <field options="{'line_breaks': False}" widget="text" class="text-break d-block" name="name" default_focus="1" placeholder="e.g. Brandon Freeman" invisible="is_company" required="type == 'contact'"/>
                                 </h1>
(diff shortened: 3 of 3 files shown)


alsh-odoo (approved):

odoo/addons/base/views/res_partner_views.xml
```diff
@@ -138,11 +138,11 @@
                                     <field options="{'line_breaks': False}" widget="text" class="text-break d-block" name="name" default_focus="1" placeholder="e.g. Lumber Inc" invisible="not is_company" required="type == 'contact'"/>
                                     <field options="{'line_breaks': False}" widget="text" class="text-break d-block" name="name" default_focus="1" placeholder="e.g. Brandon Freeman" invisible="is_company" required="type == 'contact'"/>
                                 </h1>
-                                <div class="d-flex align-items-baseline w-50">
+                                <div class="d-flex align-items-baseline w-100 w-md-50">
```
rfr-odoo: Should works also without the `w-100` using only `w-md-50` for "desktop".
```suggestion
                                <div class="d-flex align-items-baseline w-md-50">
```

odoo/addons/base/views/res_partner_views.xml
```diff
@@ -138,11 +138,11 @@
...
                                     <i class="fa fa-fw me-1 fa-envelope text-primary" title="Email"/>
                                     <field name="email" class="w-100" widget="email" required="user_ids" placeholder="Email"/>
                                 </div>
-                                <div class="d-flex align-items-baseline w-50">
+                                <div class="d-flex align-items-baseline w-100 w-md-50">
```
rfr-odoo: Same as before
```suggestion
                                <div class="d-flex align-items-baseline w-md-50">
```

addons/web/static/src/views/fields/phone/phone_field.scss
```diff
@@ -11,3 +11,10 @@ body:not(.o_touch_device) .o_field_phone {
     overflow-wrap: normal;
     word-break: normal;
 }
+
+/* On small screens, hide the text labels inside the phone buttons */
+@include media-breakpoint-down(md) {
+    .o_phone_content a small {
+        display: none;
+    }
+}
```
rfr-odoo: Why not include this into the rules before (after the line 12) as it's targetting also `.o_phone_content small` ?

```suggestion
.o_phone_content small {
    overflow-wrap: normal;
    word-break: normal;

    /* On small screens, hide the text labels inside the phone buttons */
    @include media-breakpoint-down(md) {
        display: none;
    }
}
```

addons/web/static/tests/views/fields/phone_field.test.js
```diff
@@ -219,3 +219,24 @@ test("New record, fill in phone field, then click on call icon and save", async
...
+                    <group>
+                        <field name="foo" widget="phone"/>
+                    </group>
+                </sheet>
+            </form>`,
+        resId: 1,
+    });
+    expect(`input[type="tel"]`).toHaveCount(1);
+    expect(`input[type="tel"]`).toHaveValue("yop");
+    await click(`input[type="tel"]`);
```
rfr-odoo: These steps are not needed or I'm missing something 🤔  ?

pish-odoo: Without a click, it will never fail. I tested on a laptop without the solution and without the click. We can remove the above two expect. Let me know if I missed something

rfr-odoo: Your test target mobile so you don't need to click in it.
The click is needed only in desktop has when you click it will hover the input and so show all the buttons.
But in mobile the buttons are show directly without the focus into the field.

rfr-odoo (approved):
@robodoo r+

PR #213833, opened 2025-06-11
[FIX] mail: show jump present button more eagerly

alexkuhn:
Before this commit, the "jump present" button when scrolling to older messages was displayed when going quite far in older messages.

This happens because condition was: greater distance of either 3 times the view port client height and 10 times the height of most recent messages.

When most recent messages are very big, this can be long until to see the jump to present.
Also when there are lots of messages, 3 times the scroll client height can be a bit too much.

In the past the "jump to present" UI was a banner so we were tempted to show when user was likely to want to jump to present. This is a small button now so we can be much more eager to its showing.

This commit removes the condition to recent message height and just uses the height of scroll client height for the showing of jump to present.

Before / After
before after

PR #246299, opened 2026-01-29
[FIX] mail: keep Jump to Present button visible in chatter

akus-odoo:
----------------------------------------------
In 18.0, the Jump to Present button in chatter could fail to
appear when the user scrolled away from the latest messages.

----------------------------------------------
- The present detection threshold relied on both viewport height and the cumulative height of recent messages.
- In some chatter layouts, this threshold could exceed the visible area, preventing the Jump to Present button from appearing.

----------------------------------------------
- The present threshold is based only on the viewport height.
- The Jump to Present button appears consistently regardless of message size or chatter layout.

Task-5435586

----------------------------------------------
I confirm I have signed the CLA and read the PR guidelines at https://www.odoo.com/submit-pr

### DIFF
--- a/addons/mail/static/src/core/common/thread.js
+++ b/addons/mail/static/src/core/common/thread.js
@@ -24,8 +24,7 @@ import { Transition } from "@web/core/transition";
 import { useBus, useRefListener, useService } from "@web/core/utils/hooks";
 import { escape } from "@web/core/utils/strings";
 
-export const PRESENT_VIEWPORT_THRESHOLD = 3;
-const PRESENT_MESSAGE_THRESHOLD = 10;
+export const PRESENT_VIEWPORT_THRESHOLD = 1;
 
 /**
  * @typedef {Object} Props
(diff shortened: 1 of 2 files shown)


akus-odoo:
These changes should apply only to version 18.0 and saas-18.2, as from 18.3 onward it is already working correctly with the same logic.

addons/mail/static/src/core/common/thread.js
```diff
@@ -494,13 +493,7 @@ export class Thread extends Component {
     }
 
     get PRESENT_THRESHOLD() {
-        const viewportHeight = (this.getViewportEl?.clientHeight ?? 0) * PRESENT_VIEWPORT_THRESHOLD;
-        const messagesHeight = [...this.props.thread.nonEmptyMessages]
-            .reverse()
-            .slice(0, PRESENT_MESSAGE_THRESHOLD)
-            .map((message) => this.refByMessageId.get(message.id))
-            .reduce((totalHeight, message) => totalHeight + (message?.el?.clientHeight ?? 0), 0);
-        const threshold = Math.max(viewportHeight, messagesHeight);
```
alexkuhn, line 503: That's more an IMP than a FIX, but ok for partial backport of https://github.com/odoo/odoo/pull/213833. The strategy around amount of messages was not a good idea anyway.

alexkuhn (approved):
@robodoo delegate+ r+ up to saas-18.2
reason: partial backport of https://github.com/odoo/odoo/pull/213833 (18.3)

PR #246301, opened 2026-01-29
[FIX] website_forum: check state and mode

thle-odoo:
Add check for the "flagged" state and "questions" mode.

task-5891604


mwath (approved):

thle-odoo:
Fix in master: https://github.com/odoo/odoo/pull/253859

PR #246304, opened 2026-01-29
[FIX] purchase_stock,*: correct route domain for stock replenishment wizard

vandroogenbd:
*mrp,purchase_mrp,stock

Issue
-----
MTO gets suggested by default as the replenishment route for products.

Steps to reproduce
-----
- Enable routes
- Unarchive MTO
- Create a product with a vendor and MTO route
- Open the replenishment wizard

> The default/suggested replenishment route is MTO

Cause
-----
When creating the widget, we go through its' `default_get` to set its' default route

https://github.com/odoo/odoo/blob/310897cf07ebbf484ea6877fb329215e45c481d7/addons/stock/wizard/product_replenish.py#L76-L77

the domain is computed by `_get_route_domain`, which in turn calls `_get_allowed_route_domain`

https://github.com/odoo/odoo/blob/310897cf07ebbf484ea6877fb329215e45c481d7/addons/stock/wizard/product_replenish.py#L165-L170

However, since we are creating a new, it does not have a `warehouse_id` yet, so we don't go into the `if` condition.

https://github.com/odoo/odoo/blob/310897cf07ebbf484ea6877fb329215e45c481d7/addons/stock/models/stock_replenish_mixin.py#L26-L39

We only get routes from the product, but since buy is set on the warehouse and not the product anymore, it doesn't get included.

Solution
-----
There is an override of `_get_route_domain` in `purchase_stock`

https://github.com/odoo/odoo/blob/310897cf07ebbf484ea6877fb329215e45c481d7/addons/purchase_stock/wizard/product_replenish.py#L91-L96

it filters out the buy route if there is no `seller_ids` for the product, but since the buy route cannot be a part of the domain anymore, the condition should be changed to include buy if the product has a seller.

Note: `_get_route_domain` is only ever called in `default_get`.

Other changes
-----
The same change is also applied for the manufacture route, as the logic is the same.

It makes no sens to include route that do not lead to an internal location.

-----
Ticket:
opw-5388728

### DIFF
--- a/addons/mrp/wizard/product_replenish.py
+++ b/addons/mrp/wizard/product_replenish.py
@@ -2,6 +2,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo import api, fields, models
+from odoo.fields import Domain
 
 
 class ProductReplenish(models.TransientModel):
@@ -41,3 +42,11 @@ class ProductReplenish(models.TransientModel):
         if product_tmpl_id and product_tmpl_id.bom_ids:
             delay += product_tmpl_id.bom_ids[0].produce_delay + product_tmpl_id.bom_ids[0].days_to_prepare_mo
         return fields.Datetime.add(date, days=delay)
+
+    def _get_route_domain(self, product_tmpl_id):
+        domain = super()._get_route_domain(product_tmpl_id)
+        company = product_tmpl_id.company_id or self.env.company
+        manufacture_route = self.env['stock.rule'].search([('action', '=', 'manufacture'), ('company_id', '=', company.id)]).route_id
+        if manufacture_route and product_tmpl_id.bom_ids:
+            domain = Domain.OR([domain, Domain('id', '=', manufacture_route.id)])
+        return domain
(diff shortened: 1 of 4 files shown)


addons/mrp/wizard/product_replenish.py
```diff
@@ -41,3 +42,10 @@ def _get_date_planned(self, route_id, **kwargs):
         if product_tmpl_id and product_tmpl_id.bom_ids:
             delay += product_tmpl_id.bom_ids[0].produce_delay + product_tmpl_id.bom_ids[0].days_to_prepare_mo
         return fields.Datetime.add(date, days=delay)
+
+    def _get_route_domain(self, product_tmpl_id):
+        domain = super()._get_route_domain(product_tmpl_id)
+        manufacture_route = self.env.ref('mrp.route_warehouse0_manufacture', raise_if_not_found=False)
```
Whenrow:
```suggestion
        manufacture_route = self.env['stock.rule'].search('action', '=', 'manufacture').route_id
```
maybe add also the company in the search domain + do the same in purchase for `rule.action = buy`

PR #246308, opened 2026-01-29
[IMP] account_edi_ubl_cii: Identify the correct tax at import

jugurtha-gaci:
Context:
When importing an XML invoice or vendor bill, the tax prediction was based on the customer’s invoice history.

Example: if the imported invoice contains an item found in the history with two taxes (6% and 21%), the prediction would return both taxes (6% and 21%), even though only one tax is present in the XML file.

The actual tax data present in the imported XML was not taken into account.

After this commit, the prediction is more rigorous and correctly relies on the tax information provided in the XML (restricted search domain)

task-5503126

---


smetl (approved):
@robodoo r+

PR #246311, opened 2026-01-29
[FIX] l10n_sa: fix future invoicing bug

khom-odoo:
In odoo/odoo#236865 we decided to allow clients to backdate invoices by letting them use the `invoice_date` field for the invoice date and use the current time as the issue time because we are not supposed to use a dummy value for time.

This created an issue where if a user in a timezone before SA tries to invoice a document around midnight using the current date in SA the datetime created will be in the future which will lead to the invoice being rejected by ZATCA.

This commit makes sure we normalize the selected date wrt to the current datetime in saudi arabia so that we never accidentally invoice into the future.

task-5890423
opw-5373067
---

### DIFF
--- a/addons/l10n_sa/models/account_move.py
+++ b/addons/l10n_sa/models/account_move.py
@@ -2,7 +2,10 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 import base64
 from datetime import datetime
-from odoo import api, fields, models
+from pytz import timezone, utc
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
 from odoo.tools import float_repr, format_datetime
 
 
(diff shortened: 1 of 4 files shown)


addons/l10n_sa/models/account_move.py
```diff
@@ -79,13 +82,31 @@ def _get_l10n_sa_totals(self):
...
+    def _get_normalized_l10n_sa_confirmation_datetime(self, invoice_date, invoice_time=None):
+        """
+            Makes sure that the l10n_sa_confirmation_datetime is not in the future because of different user timezones
+            and how we store timezone unaware datetimes in the database.
+            e.g. The date/time now is 1/1/2026 22:00:00 UTC+0, a user in Belgium (date/time 1/1/2026 23:00:00 UTC+1)
+            sets the invoice date to the date now in Saudi Arabia (date/time 2/1/2026 01:00:00 UTC+3) because that's
+            the date when the invoice was created to the Saudi Govt.
+            A bug will arise because the date we store in the database will be 2/1/2026 22:00:00 (no tz) which is
+            in the future and will cause ZATCA to reject the invoice.
+        """
```
alialfie: While this is a very nice well detailed docstring, it could be much shorter and convey the same message like 
> Ensures the confirmation datetime does not exceed the current time in Asia/Riyadh to prevent ZATCA rejections.

This is up to you though, not blocking :)

addons/l10n_sa/tests/test_account_move.py
```diff
@@ -0,0 +1,44 @@
+from datetime import date, datetime, time
+
```
alialfie: no need for a new line here

alialfie (changes requested):
Thanks for your work!
Left two small nitpicks below.
Also, don't forget to mention all the modified modules in your commit message :)

alialfie (approved):
Thanks for the update!
robodoo r+

PR #246312, opened 2026-01-29
[IMP] hr_holidays: accrual plan level UX

kareemoo:
Purpose: inform the user about how many days per year the employee will get from the accrual level
- added a computed field for yearly gain to show how many days per year the employee will get
- cleaned the UI of `Set the employee accrual frequency :` to appear without too much gaps

task-id: 3143664

---


amah-odoo:
@robodoo r+

PR #246314, opened 2026-01-29
[IMP] crm: update tour

frbin-odoo:
Reorder the crm tour steps so that the user goes through all the stages.

Before:
New -> Proposition -> Won -> inspect lead (skipped qualified)
After:
New -> Qualified -> inspect lead -> Move to Proposition via the form -> Won

Changed the default country in the lead generation dialog, to be based on the active company instead of the user's default.

Task-5386684

### DIFF
--- a/addons/crm/static/src/js/tours/crm.js
+++ b/addons/crm/static/src/js/tours/crm.js
@@ -50,7 +50,7 @@ registry.category("web_tour.tours").add('crm_tour', {
     trigger: ".o_opportunity_kanban:not(:has(.o_view_sample_data)) .o_kanban_group .o_kanban_record:last-of-type",
     content: markup(_t("<b>Drag &amp; drop opportunities</b> between columns as you progress in your sales cycle.")),
     tooltipPosition: "right",
-    run: "drag_and_drop(.o_opportunity_kanban .o_kanban_group:eq(2))",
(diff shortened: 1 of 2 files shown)


addons/crm_iap_mine/models/crm_iap_lead_mining_request.py
```diff
@@ -31,7 +31,7 @@ def _default_lead_type(self):
             return 'opportunity'
 
     def _default_country_ids(self):
-        return self.env.user.company_id.country_id
+        return self.env.company.country_id
```
Odoonan, line 34: This seems unrelated, right? I would put everything that is not strictly the fix in one (or more) separate IMP commit(s) , grouped by theme :) (rewording and so on)

addons/crm/static/src/js/tours/crm.js
Odoonan, line 74: Feel free to detail which 'modal' in your commit message as it is quite implicit :)

Odoonan (commented):
Hi @frbin-odoo 

Small things, thanks!

Cheers

awa-odoo:
@frbin-odoo :wave: 

Nothing to say on this one, was ready to send but runbot seems mad, can you try a rebase?

Thanks!

frbin-odoo:
Hello :wave: @awa-odoo 
Runbot is happy once again :)

awa-odoo:
@robodoo r+ rebase-ff

PR #246318, opened 2026-01-29
[IMP] survey: prevent adding opacity on images

amdi-odoo:
Purpose
=======
Prevent adding opacity on the images in the question descriptions.

Specification
=============
The css rule "opacity" adds opacity on everything. But some elements like images, banners, buttons, file uploads, ... shouldn't have this opacity layer. Images and banners look dull. Buttons and file uploads look disabled.

Changing the css rule to use rgba with the opacity parameter. This makes sure the opacity is only applied on relevant elements: test, lists, quotes, headers, emojis, ...

Task-5468647

---


tde-banana-odoo:
@robodoo r+

PR #105873, opened 2022-11-16
[IMP] crm, phone_validation: improve duplicate lead detection performance

pyduf:
Accelerate duplicate lead detection by adding indexes on field involved in the search.

Technical note: PhoneMixin defines phone_mobile_search which is used in lead detection. The indexes that accelerate that search are defined in that mixin and activated for crm.lead with _accelerate_phone_mobile_search=True defined in Lead class.

Task-3007714

Description of the issue/feature this PR addresses:

Current behavior before PR:

Desired behavior after PR is merged:

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #9386, opened 2015-11-03
[FIX] Hr holidays: context isn't propagated at meeting deletion

adrienpeiffer:

PR #246321, opened 2026-01-29
[IMP] point_of_sale, pos_*: unify payment method integrations

macocianradu:
We currently support multiple payment platforms but trying to set them up is different for each platform in the POS payment method form screen.

This commit will standardize the setup process so they all behave the same even if the input they require is different

Task-[5887093](https://www.odoo.com/odoo/project/1737/tasks/5887093)
Enterprise PR-[#105873](https://github.com/odoo/enterprise/pull/105873)
Upgrade PR-[#9386](https://github.com/odoo/upgrade/pull/9386)

---

### DIFF
--- a/addons/point_of_sale/static/src/backend/pos_payment_provider_cards/pos_payment_provider_cards.js
+++ b/addons/point_of_sale/static/src/backend/pos_payment_provider_cards/pos_payment_provider_cards.js
@@ -21,22 +21,7 @@ export class PosPaymentProviderCards extends Component {
         });
 
         onWillStart(async () => {
-            const res = await this.orm.call("pos.payment.method", "get_provider_status", [
-                providers.map((p) => p[1]),
-            ]);
-
-            this.state.providers = providers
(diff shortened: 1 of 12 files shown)


addons/point_of_sale/static/src/backend/pos_payment_provider_cards/pos_payment_provider_cards.js
```diff
@@ -20,9 +20,8 @@ export class PosPaymentProviderCards extends Component {
         });
 
         onWillStart(async () => {
-            const res = await this.orm.call("pos.payment.method", "get_provider_status", [
-                providers.map((p) => p[1]),
-            ]);
+            const providers = await this.orm.call("pos.payment.method", "get_payment_providers");
+            const res = await this.orm.call("pos.payment.method", "get_provider_status");
```
davidmonnom: IMO, this can be merged.

davidmonnom (approved):
@robodoo r+

PR #246325, opened 2026-01-29
[FIX] orm: skip access on join of a delegated field

kmagusiak:
When building the SQL for a related field, we join the table for the comodel and may apply user access to that comodel. Using *inherits* already adds record rules for that field, so joining on the field on which "delegates" access to fields, can be done without additional permissions.

```py
class M1(Model): ...
class M2(Model):
  _inherits = {'m1': 'm1_id'}
  m1_id = fields.Many2one(...)
```

Consider the following, `Query(m2_record).m1_id` should join without additional access rules. This is fine because the query usually comes from a `_search` (which has added the necessary permissions).

---


rco-odoo (approved):
@robodoo r+

PR #246330, opened 2026-01-29
[FIX] mail,website,im_livechat: interpolation of custom property values

pparidans:
Older versions of LibSass and Ruby Sass parsed custom property declarations just like any other property declaration, allowing the full range of SassScript expressions as values. But this wasn’t compatible with CSS.

To provide maximum compatibility with plain CSS, since version 3.5.0, LibSass requires SassScript expressions in custom property values to be written within interpolation.

Reference:
https://sass-lang.com/documentation/breaking-changes/css-vars/


alexkuhn (approved):
@robodoo delegate+

adr-odoo (approved):
@robodoo r+

PR #246333, opened 2026-01-29
[FIX] orm: avoid groupby on models with custom `_select`

kmagusiak:
Add a context to know if we are grouppnig. This allows us to stop `_search` methods which need to search and filter manually all records.

For example, when asking if a field is groupable,
`_description_groupable` will run a dummy search for many2many fields, for a normal user searching on ir.attachment, it would search for all attachments and try to filter them in memory leading to OOM error.

---

### DIFF
--- a/odoo/addons/base/models/ir_attachment.py
+++ b/odoo/addons/base/models/ir_attachment.py
@@ -613,6 +613,8 @@ class IrAttachment(models.Model):
         domain = domain.optimize(self)
         if self.env.su or bypass_access or domain.is_false():
             return super()._search(domain, offset, limit, order, active_test=active_test, bypass_access=bypass_access)
+        if self.env.context.get('_read_groupby'):
+            raise ValueError("Cannot group by ir.attachment")
 
         # General access rules
         # - public == True are always accessible
--- a/odoo/orm/models.py
+++ b/odoo/orm/models.py
@@ -1996,6 +1996,7 @@ class BaseModel(metaclass=MetaModel):
         if not func:
             raise ValueError(f"Aggregate method is mandatory for {fname!r}")
 
+        table = table._with_model(table._model.with_context(_read_groupby=True))
         field = self._fields[fname]
         if func == 'sum_currency':
             if field.type != 'monetary':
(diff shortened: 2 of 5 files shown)


odoo/orm/models.py
```diff
@@ -1996,6 +1996,7 @@ def _read_group_select(self, table: TableSQL, aggregate_spec: str) -> SQL:
         if not func:
             raise ValueError(f"Aggregate method is mandatory for {fname!r}")
 
+        table = table._with_model(table._model.with_context(_read_group=True))
```
rco-odoo: Use the expected context key 😉 
```suggestion
        table = table._with_model(table._model.with_context(_read_groupby=True))
```

odoo/orm/models.py
```diff
@@ -2036,6 +2037,7 @@ def _read_group_groupby(self, table: TableSQL, groupby_spec: str) -> SQL:
         if fname not in self._fields:
             raise ValueError(f"Invalid field {fname!r} on model {self._name!r}")
         field = self._fields[fname]
+        table = table._with_model(table._model.with_context(read_groupby=True))
```
rco-odoo:
```suggestion
        table = table._with_model(table._model.with_context(_read_groupby=True))
```

odoo/addons/base/models/ir_attachment.py
```diff
@@ -613,6 +613,8 @@ def _search(self, domain, offset=0, limit=None, order=None, *, active_test=True,
         domain = domain.optimize(self)
         if self.env.su or bypass_access or domain.is_false():
             return super()._search(domain, offset, limit, order, active_test=active_test, bypass_access=bypass_access)
+        if self.env.context.get('_read_groupby'):
+            raise ValueError("Cannot group by ir.attachment")
```
rco-odoo, line 617: Even if the domain restricts `res_model` and lead to the optimization below?

kmagusiak: There is an optimization in ir.attachments, but no optimization for a similar case in mail (yet).

william-andre, line 617: Is it not only the `search_fetch` combined with `_as_query` that causes an issue?

Can't we just move this condition later, if `limit is None`?

kmagusiak: No, this function is also called in read group and in other places of the orm. Technically, the condition can be moved, but it results in some domains working and other not depending on the optimization (quick could change between versions). It's easier to say don't group on all attachments (what would be the use case?) 

(still a draft)

rco-odoo (approved):
Looks good to me.

odoo/addons/base/models/ir_attachment.py
kmagusiak: I won't move the condition later to keep things consistent.

PR #246335, opened 2026-01-29
[FIX] html_builder: handle unitless zero values in CSS comparison

YoussefM890:
**Problem:**
When using themes that define border-radius as unitless `0` (e.g., Anelusia theme) and attempting to modify the "Rounded Corners" value of any element through the website editor, the following error occurs: "Cannot convert 'px' units into '' units !"

**Steps to reproduce:**
1. Install and activate the Anelusia theme (or any theme with unitless border-radius: 0)
2. Open the website editor
3. Select any element (e.g., a table in the footer)
4. Try to change the "Rounded Corners" field to any non-zero value
5. Observe the error: "Uncaught Promise > Cannot convert 'px' units into '' units !"

**Current behavior:**
The website editor throws a JavaScript error and prevents changing rounded corners.

**Expected behavior:**
Users should be able to modify rounded corners values without errors, regardless of whether the theme uses unitless or unit-based zero values.

**Cause of the issue:**
Some themes define border-radius variables as unitless `0` (e.g., `$border-radius: 0`). When the website editor's areCssValuesEqual() function in utils_css.js compares the new value (e.g., "10px") with the existing computed value ("0" - unitless), it attempts to convert between units. The getNumericAndUnit() function extracts the unit from "0" as an empty string "", then convertValueToUnit() tries to convert "10px" to "" unit. Since there's no conversion defined for "px" to "" (empty unit), convertNumericToUnit() throws the error.

**Fix:**
Add a special case in areCssValuesEqual() to handle unitless zero values before attempting unit conversion. When the first value is unitless "0", we compare the numeric values directly using Number.EPSILON, avoiding the unit conversion entirely. This allows proper comparison between unitless "0" and values like "10px" or "0px" without errors, while maintaining correct equality checks (0 equals 0px, but 0 does not equal 10px).

opw-5412414

### DIFF
--- a/addons/html_builder/static/src/utils/utils_css.js
+++ b/addons/html_builder/static/src/utils/utils_css.js
@@ -203,7 +203,10 @@ export function areCssValuesEqual(value1, value2, cssProp, htmlStyle) {
         return false;
     }
     const numValue1 = data[0];
-    const numValue2 = convertValueToUnit(value2, data[1], htmlStyle);
+    // Zero values don't need unit conversion (0px === 0rem === 0em === 0)
+    const numValue2 = parseInt(numValue1) === 0
+        ? getNumericAndUnit(value2)[0]
+        : convertValueToUnit(value2, data[1], htmlStyle);
     return Math.abs(numValue1 - numValue2) < Number.EPSILON;
 }
 /**


addons/html_builder/static/src/utils/utils_css.js
```diff
@@ -203,6 +203,14 @@ export function areCssValuesEqual(value1, value2, cssProp, htmlStyle) {
...
     }
     const numValue1 = data[0];
+    const unitValue1 = data[1];
+    // If the first value is unitless 0 (e.g., from themes with $border-radius: 0),
+    // compare numeric values directly to avoid unit conversion errors. This handles
+    // cases like comparing "0" (unitless) with "10px" or "0px".
+    if (numValue1 === '0' && unitValue1 === '') {
+        const numValue2 = getNumericAndUnit(value2)[0];
+        return Math.abs(numValue1 - numValue2) < Number.EPSILON;
+    }
```
nle-odoo: this could probably be written more simply, for example not using `convertValueToUnit` if we know `numValue1 is < Number.EPSILON` (or 0)

nle-odoo (approved):
comparing with saas-18.2, in this function:

- in saas-18.2, value1 comes from `computedStyle.getPropertyValue('border-bottom-left-radius')` in `applyCSS` and we always receive `'0px'` (probably coerced since this is a meaningful CSS style attribute)

- in saas-18.2, value1 comes from `computedStyle.getPropertyValue('--box-border-bottom-left-radius')` in `applyNeededCss` and we always receive `0` (probably not coerced since this is a CSS variable)

This seems to be an intended change, so I think it makes sense to make a change after which we should handle value0 as '0' either as a '0px' or avoid the conversion.

I don't know the best way to implement this, and if only 0 that could have this behavior.

This could already probably be further reviewed by website team after my remark is applied. They are probably aware of the change from style attribute to style variable and more aware of the best way this could be solved.

addons/html_builder/static/src/utils/utils_css.js
```diff
@@ -203,7 +203,10 @@ export function areCssValuesEqual(value1, value2, cssProp, htmlStyle) {
         return false;
     }
     const numValue1 = data[0];
-    const numValue2 = convertValueToUnit(value2, data[1], htmlStyle);
+    // Zero values don't need unit conversion (0px === 0rem === 0em === 0)
+    const numValue2 = Math.abs(numValue1) < Number.EPSILON
```
robinlej: Why not simply `Math.abs(numValue1) === 0` or even `parseInt(numValue1) === 0`?
`Number.EPSILON` is useful to compare an operation, but here it's just a plain value, and we know exactly which one

robinlej (commented):
Could you add a test?
I think the best place would be in `website/static/tests/builder/options/border_configurator_option.test.js`. It could just be:

- setup the HTML Builder with an element with an inline style `border-radius: 0`
- click on it
- check that the input of the "Round corners" option shows "0"

(Currently on stable and master, this will wrongly show "6.4".)

addons/website/static/tests/builder/options/border_configurator_option.test.js
```diff
@@ -59,6 +59,21 @@ test("empty border input is treated as 0", async () => {
...
+        template: xml`<BorderConfigurator label="'Border'"/>`,
+    });
+    await setupWebsiteBuilder(
+        `<section class="test-options-target" style="border-radius: 0">Rounded element</section>`,
+        {
+            loadIframeBundles: true,
+        }
+    );
+    await contains(":iframe section").click();
+    await waitFor(".options-container [data-label='Round Corners']");
```
robinlej: **EDIT: I leave this comment FYI, but don't take it into account** (see other comment about the test)

Rather use `await waitSidebarUpdated();` instead of `waitFor` on an explicit element. We've had indeterministic errors before around that kind of actions in tests.

You can extract it from `setupHTMLBuilder`:
`const { waitSidebarUpdated } = await setupHTMLBuilder(....);`

addons/html_builder/static/src/utils/utils_css.js
```diff
@@ -203,7 +203,10 @@ export function areCssValuesEqual(value1, value2, cssProp, htmlStyle) {
         return false;
     }
     const numValue1 = data[0];
-    const numValue2 = convertValueToUnit(value2, data[1], htmlStyle);
+    // Zero values don't need unit conversion (0px === 0rem === 0em === 0)
+    const numValue2 = numValue1 === 0
```
robinlej: `getNumericAndUnit` returns 2 strings, so `numValue1` can never be `=== 0`. That's why I suggested to use `parseInt(numValue1)` ;)

addons/website/static/tests/builder/options/border_configurator_option.test.js
```diff
@@ -59,6 +59,21 @@ test("empty border input is treated as 0", async () => {
     await contains(borderOptionInputSelector).clear();
     expect.verifySteps(["hasBorder"]);
 });
+test("round corners input shows 0 for element with unitless border-radius: 0", async () => {
```
robinlej: I'm sorry for the extra work, especially given that I think you did a good work on that test... But I think I misunderstood both the issue and the behavior I saw in the builder when I asked you to write this test. It doesn't actually test anything fixed by your commit, and it already worked fine before.
A proper test would need to create a scss custom file to change the variable, load it through python, and run a tour. Given the fix, let's say we don't need it.

robinlej (changes requested):

robinlej (approved):
@robodoo r+

Thanks!

PR #246342, opened 2026-01-29
[FIX] base: partner bank should follow same company rules as partner

jco-odoo:

Current behavior before PR:

---

### DIFF
--- a/addons/account/models/res_partner_bank.py
+++ b/addons/account/models/res_partner_bank.py
@@ -16,6 +16,7 @@ class ResPartnerBank(models.Model):
 
     journal_id = fields.One2many(
         'account.journal', 'bank_account_id', domain=[('type', '=', 'bank')], string='Account Journal', readonly=True,
+        check_company=True,
         help="The accounting journal corresponding to this bank account.")
     has_iban_warning = fields.Boolean(
         compute='_compute_display_account_warning',
--- a/addons/account/models/res_partner_bank.py
+++ b/addons/account/models/res_partner_bank.py
@@ -16,7 +16,6 @@ class ResPartnerBank(models.Model):
 
     journal_id = fields.One2many(
         'account.journal', 'bank_account_id', domain=[('type', '=', 'bank')], string='Account Journal', readonly=True,
-        check_company=True,
         help="The accounting journal corresponding to this bank account.")
     has_iban_warning = fields.Boolean(
         compute='_compute_display_account_warning',
--- a/odoo/addons/base/models/res_bank.py
+++ b/odoo/addons/base/models/res_bank.py
@@ -74,6 +74,8 @@ class ResPartnerBank(models.Model):
     _rec_name = 'acc_number'
     _description = 'Bank Accounts'
     _order = 'sequence, id'
+    _check_company_auto = True
+    _check_company_domain = models.check_company_domain_parent_of
 
     @api.model
     def get_supported_account_types(self):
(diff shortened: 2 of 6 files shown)


addons/account/models/res_partner_bank.py
```diff
@@ -16,7 +16,6 @@ class ResPartnerBank(models.Model):
 
     journal_id = fields.One2many(
         'account.journal', 'bank_account_id', domain=[('type', '=', 'bank')], string='Account Journal', readonly=True,
-        check_company=True,
```
william-andre, line 19: Why? Commit doesn't explain either

jco-odoo: @william-andre Just a quick test because otherwise the test failed.  (bank_statement_extract) Basically because it does not allow bank to have company_id False and journal not. (if I put the _check_company rules)

william-andre: Then you need a different `_check_company_domain` allowing it

jco-odoo: Should ideally only be on the field level.

llabusch93:
Hi @jco-odoo 👋

Just a friendly nudge — we're an Odoo partner and this fix would really help one of our customers who is currently blocked by this issue (Sage50 credit note imports failing with "Incompatible companies"). We've had to pin their production instance to an older version as a workaround, which isn't ideal long-term.

Is there anything we can do to help move this forward? Happy to assist with testing or provide additional context if needed.

Thanks for your work on this! 🙏

jco-odoo:
@llabusch93 I am not sure this is the right fix in 19. @william-andre Did not you have another fix in mind or better to make this green?

odoo/addons/base/models/res_bank.py
```diff
@@ -74,6 +74,8 @@ class ResPartnerBank(models.Model):
     _rec_name = 'acc_number'
     _description = 'Bank Accounts'
     _order = 'sequence, id'
+    _check_company_auto = True
+    _check_company_domain = models.check_company_domain_parent_of
```
william-andre, line 79: ?
```suggestion
    _check_company_domain = models.check_company_domain_parent_of_false
```

odoo/orm/models.py
```diff
@@ -184,6 +184,23 @@ def check_company_domain_parent_of(self, companies):
         for parent in rec.parent_path.split('/')[:-1]
     ] + [False])]
 
+def check_company_domain_parent_of_false(self, companies):
+    """ A `_check_company_domain` function that lets a record be used if either:
+        - record.company_id = False (which implies that it is shared between all companies), or
+        - record.company_id is a parent of any of the given companies.
+    """
```
william-andre, line 191: Docstring seems to be missing copy pasted?

odoo/orm/models.py
```diff
@@ -184,6 +184,23 @@ def check_company_domain_parent_of(self, companies):
         for parent in rec.parent_path.split('/')[:-1]
     ] + [False])]
 
+def check_company_domain_parent_of_false(self, companies):
```
william-andre, line 187: Function name is not very explicit

william-andre (commented):
This is lacking context and I don't really know what I'm reviewing; we talked about it weeks ago but my memory doesn't retain information for that long.

It certainly deserves a test and a clearer commit message/PR description.

jco-odoo:
@llabusch93 Trying something simpler in https://runbot.odoo.com/runbot/bundle/190-fixbankparentoftry2-jco-462387

jco-odoo:
@llabusch93 We checked further again and the simpler fixes are the better ones: https://github.com/odoo/odoo/pull/262173/changes#r3219031853

PR #246346, opened 2026-01-29
[FIX] web: switchView after page reload

aurb9:
Steps to reproduce:
- Open a project
- Create a task and convert it into a template
- Open another task (task A)
- Reload the page
- Create a task from the newly created template

Refreshing the page causes `loadState` to rebuild the controller stack from the URL to represent the breadcrumb history. In this case, a virtual form controller is injected for the opened task A, due to the lack of context in the URL to reconstruct it fully.

When creating the task from the template, a `switchView` to the new task's form view is triggered. However, only the controller for this new form view is fully populated with the relevant metadata, as the preceding ones are virtual (due to the above).

This commit ensures that virtual controllers are excluded from the check on the `multiRecord` field, preventing an error since the `view` is undefined for virtual controllers.

task-5876607

### DIFF
--- a/addons/web/static/src/webclient/actions/action_service.js
+++ b/addons/web/static/src/webclient/actions/action_service.js
@@ -1692,7 +1692,8 @@ export function makeActionManager(env, router = _router) {
             // This case would mostly happen when loadState detects a change in the URL.
             // Also, I guess we may need it when we have other monoRecord views
             index = controllerStack.findIndex(
-                (ct) => ct.action.jsId === controller.action.jsId && !ct.view.multiRecord
+                (ct) =>
+                    ct.action.jsId === controller.action.jsId && !ct.virtual && !ct.view.multiRecord
             );
             index = index > -1 ? index : controllerStack.length;
         }
--- a/addons/web/static/tests/webclient/actions/load_state.test.js
+++ b/addons/web/static/tests/webclient/actions/load_state.test.js
@@ -1886,6 +1886,24 @@ describe(`new urls`, () => {
         // assert that we properly reload action and breacrumbs as lang changed
         expect.verifySteps(["/web/action/load_breadcrumbs", "/web/action/load"]);
     });
+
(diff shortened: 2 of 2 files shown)


addons/project/static/src/views/components/project_task_template_dropdown.js
```diff
@@ -64,15 +64,20 @@ export class ProjectTaskTemplateDropdown extends Component {
...
+            res_id: await this.orm.call(
                 "project.task",
                 "action_create_from_template",
                 [templateId],
                 {
                     context: context,
                 }
             ),
+            views: [[false, "form"]],
+            target: "current",
```
deneuvillem: I think it's the default value, so no need to mention it

addons/project/static/src/views/components/project_task_template_dropdown.js
```diff
@@ -64,15 +64,20 @@ export class ProjectTaskTemplateDropdown extends Component {
...
+            context: context,
             focusTitle: true,
```
deneuvillem, line 76: Does this still work as intended? I have some doubt 🙈

deneuvillem (commented):
Thanks for your work, I left some comments for you :)
(I rebased your branch as well)

addons/project/static/src/views/components/project_task_template_dropdown.js
aurb9: The target "current" is indeed the default, but the views are required

addons/project/static/src/views/components/project_task_template_dropdown.js
aurb9: Good gatch, thanks!

jeanschoenlaub:
Hey @aurb9 , thanks for the fix could you please add a unit test that fails before your fix and passes after it ? Thakns

aab-odoo (approved):
Thanks @aurb9 @jeanschoenlaub 👍 

robodoo r+

PR #246347, opened 2026-01-29
[FIX] base, web: use dynamic precision when using minimal precision

Arnaud-Sibille:
8c199f7783527735b35c9fbda334cbdcd55a004f introduced minimum precision
for float fields, but it has a display issue.

Steps to reproduce:
- create an invoice, set the price of an an invoice line to `9417.2`
-> the web client displays `'9,417.200000000001'`

Issue:
In case of a minimum precision, the precision was set to 12.
But it should be dynamic in regards to the number of integer digits.

Solution:
The precision is set to: 15 - number of integer digits

--

Also implementing the same logic in qweb, but by keeping a maximum
precision of 6.

To reproduce in qweb:
- create a an invoice with an invoice line of `300000000031.3`
- post and print
-> the price unit on the pdf shows "300,000,000,031.300232"

opw-5880951

### DIFF
--- a/addons/web/static/src/core/utils/numbers.js
+++ b/addons/web/static/src/core/utils/numbers.js
@@ -199,6 +199,7 @@ export function humanNumber(number, options = { decimals: 0, minDigits: 1 }) {
  * @param {Object} [options]
  * @param {number[]} [options.digits] the number of digits that should be used,
  *   instead of the default digits precision in the field.
+ * @param {nummber} [optinos.minDigits] the minimum number of decimal digits to display.
  * @param {boolean} [options.humanReadable] if true, large numbers are formatted
  *   to a human readable format.
  * @param {string} [options.decimalPoint] decimal separating character
@@ -215,9 +216,12 @@ export function formatFloat(value, options = {}) {
     if (options.digits && options.digits[1] !== undefined) {
         precision = options.digits[1];
     } else if (options.minDigits) {
-        // When no precision is set, 12 is chosen as the precision. It is high enough to keep it precise,
-        // but not too high in order to avoid rounding errors..
-        precision = 12;
+        const intDigitsCount = (value !== 0) ? Math.floor(Math.log10(Math.abs(value))) + 1 : 1;
+        // We estimate the maximum decimal digits we can display without showing rounding errors,
+        // by substracting the number of integer digits to 15, as floats have 15-16 digits precision.
+        const maxDecDigits = Math.max(15 - intDigitsCount, 0);
+        // We display maximum 6 digits or the number of significant digits (if it's lower)
+        precision = Math.min(6, maxDecDigits);
     } else {
         precision = 2;
     }
--- a/addons/web/static/tests/core/utils/numbers.test.js
+++ b/addons/web/static/tests/core/utils/numbers.test.js
@@ -252,6 +252,42 @@ test("floatIsZero", () => {
 });
 
 describe("formatFloat", () => {
+    test("precision", () => {
+        patchWithCleanup(localization, {
+            decimalPoint: ".",
+            grouping: [3, 0],
+            thousandsSep: ",",
+        });
+
+        let options = {};
+        expect(formatFloat(3, options)).toBe("3.00");
+        expect(formatFloat(3.1, options)).toBe("3.10");
+        expect(formatFloat(3.12, options)).toBe("3.12");
+        expect(formatFloat(3.129, options)).toBe("3.13");
+
+        options = { digits: [15, 3] };
+        expect(formatFloat(3, options)).toBe("3.000");
+        expect(formatFloat(3.1, options)).toBe("3.100");
+        expect(formatFloat(3.123, options)).toBe("3.123");
+        expect(formatFloat(3.1239, options)).toBe("3.124");
+
+        options = { minDigits: 3 };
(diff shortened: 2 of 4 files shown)


addons/web/static/src/core/utils/numbers.js
```diff
@@ -214,13 +214,19 @@ export function formatFloat(value, options = {}) {
...
+        const intDigitsCount = Math.floor(Math.log10(Math.abs(value))) + 1;
+        if (intDigitsCount >= 4) {
+            precision = Math.max(15 - intDigitsCount, options.minDigits);
+        } else {
+            precision = 12;
+        }
     } else {
         precision = 2;
     }
+
```
william-andre: Do we need these?

addons/web/static/src/core/utils/numbers.js
```diff
@@ -214,13 +214,15 @@ export function formatFloat(value, options = {}) {
...
         precision = options.digits[1];
-    } else if (options.minDigits) {
-        // When no precision is set, 12 is chosen as the precision. It is high enough to keep it precise,
-        // but not too high in order to avoid rounding errors..
-        precision = 12;
+    } else if (options.minDigits && value !== 0) {
+        // Floats only have ~15–16 reliable digits.  So, as the number of integer digits increase,
+        // the number of decimal digits we display decreases, to avoid showing rounding errors.
+        const intDigitsCount = Math.floor(Math.log10(Math.abs(value))) + 1;
+        precision = Math.max(15 - intDigitsCount, options.minDigits);
```
william-andre: What if `options.minDigits` is not defined?

william-andre: Uh, nevermind.

william-andre (approved):
@robodoo r+

aab-odoo:
Hello guys. You're adding props and options to core components/APIs of the web framework (without our approval). At some point, it would be nice to at least document (docstring at the very least) and unit test them... I didn't see a single test about the new minDigits thing.

william-andre:
@aab-odoo sorry about that. The previous PR was open for more than a week, and the bot pinged every team involved with the diff, including yours.
We got an approval from the ORM, the code in JS is only replicating the python code.

But we will add tests (@Arnaud-Sibille )

addons/web/static/src/core/utils/numbers.js
```diff
@@ -215,9 +215,10 @@ export function formatFloat(value, options = {}) {
     if (options.digits && options.digits[1] !== undefined) {
         precision = options.digits[1];
     } else if (options.minDigits) {
-        // When no precision is set, 12 is chosen as the precision. It is high enough to keep it precise,
-        // but not too high in order to avoid rounding errors..
-        precision = 12;
+        // Floats only have ~15–16 reliable digits.  So, as the number of integer digits increase,
+        // the number of decimal digits we display decreases, to avoid showing rounding errors.
```
aab-odoo: where does this justification come from? Do you have a pointer to an article or something that suggests that kind of solution? If so, it would be nice to link it here.

addons/web/static/src/core/utils/numbers.js
```diff
@@ -215,9 +215,10 @@ export function formatFloat(value, options = {}) {
     if (options.digits && options.digits[1] !== undefined) {
         precision = options.digits[1];
     } else if (options.minDigits) {
```
aab-odoo, line 218: the minDigits option must be added in the docstring here and in formatFloat in formatters.js

aab-odoo: also, the `min_display_digits` option that has been added on the float field widgets must be documented on the [official documentation](https://www.odoo.com/documentation/master/developer/reference/frontend/javascript_reference.html#non-relational-fields)

addons/web/static/tests/core/utils/numbers.test.js
```diff
@@ -252,6 +252,42 @@ test("floatIsZero", () => {
 });
 
 describe("formatFloat", () => {
+        test("precision", () => {
```
aab-odoo: could you format/lint your code? Can be done automatically if you enable the tooling (web/tooling/enable.sh)

addons/web/static/src/core/utils/numbers.js
william-andre: https://docs.python.org/3/tutorial/floatingpoint.html

JS uses the same representation for floats

Arnaud-Sibille: Also https://stackoverflow.com/questions/455612/limiting-floats-to-two-decimal-points

HydrionBurst (approved):

addons/web/static/src/core/utils/numbers.js
Arnaud-Sibille: added :)

addons/web/static/tests/core/utils/numbers.test.js
Arnaud-Sibille: should be ok now

aab-odoo (approved):

william-andre (approved):
@robodoo r+

william-andre:
@calkikhunt Hello, would you mind explaining what you consider an issue? This PR is already merged and shluld be in production.

PR #246350, opened 2026-01-29
[FIX] account{_peppol,_edi_ubl_cii,}: allow sending via Peppol after PDF or XML generation

coco2299:
### Issue:
In 17.0, the `Send via Peppol` option becomes unavailable after using `Send by email` on an invoice

### Cause:
The `Send via Peppol` action depends on the UBL-CII XML file
However, the XML generation was conditioned on the absence of a PDF document

After generating the PDF, the invoice was considered as already having all required documents, so no additional EDI documents (including the UBL-CII XML) were generated, which disabled the `Peppol` option

### Note:
If the XML was already generated, it should still be possible to `Send via Peppol`
However, the XML was not attached because it wasn't generated during Peppol sending
This issue is also fixed in this PR

### Steps to reproduce:
- Install `l10n_be` and switch to BE Company
- Enable Peppol and activate the Demo mode in Settings
- Create an invoice for the BE Company
- Add any product with a Tax
- Confirm -> Send & Print -> Email
- Open Send & Print again Before the fix, the `Send via Peppol` option is no longer available

opw-5490217

### DIFF
--- a/addons/account/wizard/account_move_send.py
+++ b/addons/account/wizard/account_move_send.py
@@ -461,6 +461,8 @@ class AccountMoveSend(models.TransientModel):
         """
         # create an attachment that will become 'invoice_pdf_report_file'
         # note: Binary is used for security reason
+        if 'pdf_attachment_values' not in invoice_data:
+            return
         invoice_sudo = invoice.sudo()
         invoice_sudo.message_main_attachment_id = self.sudo().env['ir.attachment'].create(invoice_data['pdf_attachment_values'])
         invoice_sudo.invalidate_recordset(fnames=['invoice_pdf_report_id', 'invoice_pdf_report_file'])
@@ -652,7 +654,7 @@ class AccountMoveSend(models.TransientModel):
         :param invoices_data:   The collected data for invoices so far.
         """
         for invoice, invoice_data in invoices_data.items():
-            if self._need_invoice_document(invoice):
+            if self.with_context(invoice_data=invoice_data)._need_invoice_document(invoice):
                 self._hook_invoice_document_before_pdf_report_render(invoice, invoice_data)
                 invoice_data['blocking_error'] = invoice_data.get('error') \
                                                  and not (allow_fallback_pdf and invoice_data.get('error_but_continue'))
@@ -672,7 +674,7 @@ class AccountMoveSend(models.TransientModel):
             if not invoice_data.get('error') or invoice_data.get('error_but_continue')
         }
         for invoice, invoice_data in invoices_data_pdf.items():
-            if self._need_invoice_document(invoice) and not invoice_data.get('error'):
+            if self.with_context(invoice_data=invoice_data)._need_invoice_document(invoice) and not invoice_data.get('error'):
                 self._prepare_invoice_pdf_report(invoice, invoice_data)
                 self._hook_invoice_document_after_pdf_report_render(invoice, invoice_data)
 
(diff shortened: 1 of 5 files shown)


coco2299:
@robodoo fw=no
The issue is only in 17.0

addons/account_edi_ubl_cii/wizard/account_move_send.py
```diff
@@ -68,6 +68,14 @@ def _compute_enable_ubl_cii_xml(self):
         for wizard in self:
             wizard.enable_ubl_cii_xml = any(m._need_ubl_cii_xml() for m in wizard.move_ids)
 
+    @api.model
+    def _need_invoice_document(self, invoice):
+        result = super()._need_invoice_document(invoice)
+        invoice_ams = self.env['account.move.send'].search([('move_ids', 'any', [('id', '=', invoice.id)])])
```
gawa-odoo: Sorry, can you explain what exactly you're trying to do here?
account.move.send is a TransientModel, so it's garbage collected regularly.
So if someone tries to come back the tomorrow, he won't find any existing invoice_ams

addons/account/wizard/account_move_send.py
```diff
@@ -461,6 +461,8 @@ def _link_invoice_documents(self, invoice, invoice_data):
         """
         # create an attachment that will become 'invoice_pdf_report_file'
         # note: Binary is used for security reason
+        if not 'pdf_attachment_values' in invoice_data:
```
gawa-odoo:
```suggestion
        if 'pdf_attachment_values' not in invoice_data:
```

addons/account_edi_ubl_cii/wizard/account_move_send.py
coco2299: In `_generate_invoice_documents()`, there are multiple places where a check is done to know whether documents must be generated via `_need_invoice_document()`.

So I extended `_need_invoice_document()` to take into account the `account.move.send` wizard: if `enable_ubl_cii_xml` and `checkbox_ubl_cii_xml` are set on the wizard, we want to generate the documents.

The wizard is created just before the document generation and is passed in `self`, but since `_need_invoice_document()` is an `@api.model`, relying on `self` felt wrong to me from a design point of view.

As I only need to read the parameters from the recently created wizard, I went with this approach instead. It seemed the most logical and least intrusive solution to access the wizard context without changing the method signature.

gawa-odoo: Yeah, but I would prefer you re-computing everything that base it on a Transient.
So re-compute `enable_ubl_cii_xml` and the like

addons/account_edi_ubl_cii/wizard/account_move_send.py
```diff
@@ -68,6 +68,14 @@ def _compute_enable_ubl_cii_xml(self):
         for wizard in self:
             wizard.enable_ubl_cii_xml = any(m._need_ubl_cii_xml() for m in wizard.move_ids)
 
+    @api.model
+    def _need_invoice_document(self, invoice):
+        result = super()._need_invoice_document(invoice)
+        invoice_data = self.env.context.get('invoice_data')
+        if invoice_data and invoice_data['ubl_cii_xml'] and invoice._need_ubl_cii_xml():
```
StefanRijnhart, line 75: This causes a KeyError
```
  File "/home/odoo/src/odoo/addons/account_edi_ubl_cii/wizard/account_move_send.py", line 75, in _need_invoice_document
    if invoice_data and invoice_data['ubl_cii_xml'] and invoice._need_ubl_cii_xml():
KeyError: 'ubl_cii_xml'
```
also reported in https://github.com/odoo/odoo/issues/250610

coco2299: I'll do a PR to fix that: https://github.com/odoo/odoo/pull/250648

PR #246354, opened 2026-01-29
[FIX] tools: never consider a node with directive as translatable

blse-odoo:
Nodes with directives must not be included inside a translatable span. But the function `translatable` missed the directive `groups` (without `t-`), and the class `o_translate_inline` should only override the predicate about the element's tag.

### DIFF
--- a/odoo/addons/base/tests/test_translate.py
+++ b/odoo/addons/base/tests/test_translate.py
@@ -153,6 +153,74 @@ class TranslationToolsTestCase(BaseCase):
         self.assertEqual(result, source)
         self.assertItemsEqual(terms, ['Form stuff'])
 
+    def test_translate_xml_o_translate_inline_on_block(self):
+        """ Test xml_translate() with non-inline elements with o_translate_inline. """
+        terms = []
+        source = """<div>
+                        <h1 class="o_translate_inline">Blah</h1>more text
+                        <h1 class="o_translate_inline" t-if="True">Other Blah</h1>even more text
+                    </div>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms,
+            ['<h1 class="o_translate_inline">Blah</h1>more text', 'Other Blah', 'even more text'])
+
+    def test_translate_xml_o_translate_inline_on_parent(self):
+        """ Test xml_translate() with non-inline elements inside o_translate_inline. """
+        terms = []
+        source = """<div>
+                        <span class="o_translate_inline">Blah<h1>more text</h1></span>
+                        <span class="o_translate_inline">Other Blah<h1 t-if="True">even more text</h1></span>
+                    </div>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms,
+            ['<span class="o_translate_inline">Blah<h1>more text</h1></span>', 'Other Blah', 'even more text'])
+
+    def test_translate_xml_highlight(self):
+        """ Test xml_translate() with highlight span (with o_translate_inline). """
+        terms = []
+        source = """<div>
+                        <span class="o_text_highlight o_translate_inline">
+                            <a>solo link</a>
+                        </span>
+                    </div>
+                    <div>
+                        <span class="o_text_highlight o_translate_inline">
+                            <span>Here is a <a>nested link</a> in highlight</span>
+                        </span>
+                    </div>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms, ["""<span class="o_text_highlight o_translate_inline">
+                            <a>solo link</a>
+                        </span>""", """<span class="o_text_highlight o_translate_inline">
+                            <span>Here is a <a>nested link</a> in highlight</span>
+                        </span>"""])
+
+    def test_translate_xml_o_translate_inline_with_groups(self):
+        """ Test xml_translate() with groups attribute and with o_translate_inline. """
+        terms = []
+        source = """<div>
+                        <a class="o_translate_inline" href="#" groups="anyone">Skip</a>
+                    </div>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms, ['Skip'])
+
+    def test_translate_xml_groups(self):
+        """ Test xml_translate() with groups attributes. """
+        terms = []
+        source = """<t t-name="stuff">
+                        stuff before
+                        <span groups="anyone"/>
+                        stuff after
+                    </t>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms,
+            ['stuff before', 'stuff after'])
+
     def test_translate_xml_t(self):
         """ Test xml_translate() with t-* attributes. """
         terms = []
--- a/odoo/tools/translate.py
+++ b/odoo/tools/translate.py
@@ -210,40 +210,50 @@ def translate_xml_node(node, callback, parse, serialize):
         """ Return whether ``text`` is a string with non-space characters. """
         return bool(text) and not space_pattern.fullmatch(text)
 
-    def translatable(node):
+    def is_force_inline(node):
+        """ Return whether ``node`` is marked as it should be translated as
+            one term.
+        """
+        return "o_translate_inline" in node.attrib.get("class", "").split()
+
+    def translatable(node, force_inline=False):
         """ Return whether the given node can be translated as a whole. """
+        # Some specific nodes (e.g., text highlights) have an auto-updated DOM
+        # structure that makes them impossible to translate.
+        # The introduction of a translation `<span>` in the middle of their
+        # hierarchy breaks their functionalities. We need to force them to be
+        # translated as a whole using the `o_translate_inline` class.
+        force_inline = force_inline or is_force_inline(node)
         return (
-            # Some specific nodes (e.g., text highlights) have an auto-updated
-            # DOM structure that makes them impossible to translate.
-            # The introduction of a translation `<span>` in the middle of their
-            # hierarchy breaks their functionalities. We need to force them to
-            # be translated as a whole using the `o_translate_inline` class.
-            "o_translate_inline" in node.attrib.get("class", "").split()
-            or node.tag in TRANSLATED_ELEMENTS
-            and not any(key.startswith("t-") for key in node.attrib)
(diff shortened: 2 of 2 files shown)


odoo/tools/translate.py
```diff
@@ -218,9 +218,9 @@ def translatable(node):
             # The introduction of a translation `<span>` in the middle of their
             # hierarchy breaks their functionalities. We need to force them to
             # be translated as a whole using the `o_translate_inline` class.
-            "o_translate_inline" in node.attrib.get("class", "").split()
-            or node.tag in TRANSLATED_ELEMENTS
-            and not any(key.startswith("t-") for key in node.attrib)
+            ("o_translate_inline" in node.attrib.get("class", "").split()
+            or node.tag in TRANSLATED_ELEMENTS)
```
HydrionBurst: what? why do you change it so much
It is the feature, we allow you to force translate the whole node

HydrionBurst (commented):
Since it may cause translated term mismatch, could you provide some real use cases that we must solve in stable version?

odoo/tools/translate.py
blse-odoo: While working on master, we noticed that some links annotated with `o_translate_inline` also have qweb directives ([example](https://github.com/odoo/odoo/blob/saas-18.3/addons/website_event_track/views/event_templates.xml#L11)).
And translating the page containing such a link from the website builder in translate mode would remove the directive and replace it with the result of the qweb engine. In the example:
```xml
... <a class="o_translate_inline" t-att-href="'/event/%s/track' % (slug(event))">register ...</a>
```
after saving a translation from website builder, would be:
```xml
... <a class="o_translate_inline" href="/event/reveal-8/track">inscrivez-vous ...</a>
```
Thus it would not reflect the correct event in that language, but always the one from which the translation mode was opened.

Based on this, we wanted to add the parentheses we added in this commit, but weren't sure if the fix would be desirable in stable as well.

We asked chm, who told us that this function is also used to control what is accepted as a translation term, and that the current implementation allows translators to insert any directives, which is an issue (I don't know how translations are sourced, thus I cannot judge that). And thus told us a fix preventing qweb directives from being inserted in terms would be desirable in stable. As `groups` is also a directive, but was missed with the check `startswith('t-')`, we added it as well.

About the terms mismatch, for the translations from website builder, as described above, it probably is better to break the translation to restore the directive. For other source of translations it would loose correct translations in these cases (`o_translate_inline` with qweb directives)

blse-odoo: I just realized that the suggested fix is stricter than necessary: it will break in several terms something like this
```xml
<span class="o_translate_inline">X<p>Y</p>Z<span>
```
But that is not necessary (for the goal here). We could consider translatable the elements with a tag not in `TRANSLATE_ELEMENTS` but with an ancestor that has the `o_translate_inline` class:
```suggestion
            (ancestor_has_o_translate_inline
            or "o_translate_inline" in node.attrib.get("class", "").split()
            or node.tag in TRANSLATED_ELEMENTS)
```

HydrionBurst: @xO-Tx 
Can you confirm it is what we need?
Now `o_translate_inline` actually means `o_translate_tag`

If it is accepted, I think we would assume we don't/won't have use cases like
`<a class="o_translate_inline text-muted"><div> xxx </div></a>`

Also I see we already have many codes in xml like
`<a class="o_translate_inline" t-att-href="`
my question is if we don't want to translate `<a t-att-href` why we added "o_translate_inline" explicitly in the code?

Gorash (commented):
LGFM
Please add a test

xO-Tx (changes requested):
> @xO-Tx
> Can you confirm it is what we need?
> Now o_translate_inline actually means o_translate_tag

@HydrionBurst Good catch! Actually, this fix will break what we did for text highlights and will again allow checking the children of a node to consider it translatable as a whole, (which breaks highlighted text translations, see: https://github.com/odoo/odoo/commit/e9659e55356e5e4de63f2324eb88a7b6f07cd835).
I’ll check with @blse-odoo to adapt the diff :+1:

HydrionBurst:
> @HydrionBurst Good catch! Actually, this fix will break what we did for text highlights and will again allow checking the children of a node to consider it translatable as a whole, (which breaks highlighted text translations, see: [e9659e5](https://github.com/odoo/odoo/commit/e9659e55356e5e4de63f2324eb88a7b6f07cd835)). I’ll check with @blse-odoo to adapt the diff 👍

BTW, a workaround would be `def translatable(node, check_tag=True)`

odoo/tools/translate.py
```diff
@@ -210,18 +210,19 @@ def nonspace(text):
...
             # The introduction of a translation `<span>` in the middle of their
             # hierarchy breaks their functionalities. We need to force them to
             # be translated as a whole using the `o_translate_inline` class.
-            "o_translate_inline" in node.attrib.get("class", "").split()
-            or node.tag in TRANSLATED_ELEMENTS
-            and not any(key.startswith("t-") for key in node.attrib)
-            and all(translatable(child) for child in node)
+            (marked_translate_inline
+            or node.tag in TRANSLATED_ELEMENTS)
+            and not any(key.startswith("t-") or key == 'groups' for key in node.attrib)
```
HydrionBurst, line 232: Could you add a comment to explain why we have to check `t-` and `gourps`?

odoo/tools/translate.py
```diff
@@ -210,18 +210,19 @@ def nonspace(text):
         """ Return whether ``text`` is a string with non-space characters. """
         return bool(text) and not space_pattern.fullmatch(text)
 
-    def translatable(node):
+    def translatable(node, marked_translate_inline=False):
         """ Return whether the given node can be translated as a whole. """
+        marked_translate_inline = marked_translate_inline or "o_translate_inline" in node.attrib.get("class", "").split()
         return (
             # Some specific nodes (e.g., text highlights) have an auto-updated
```
HydrionBurst: Nit: we could move this to before the `marked_translate_inline`

odoo/tools/translate.py
```diff
@@ -210,18 +210,19 @@ def nonspace(text):
         """ Return whether ``text`` is a string with non-space characters. """
         return bool(text) and not space_pattern.fullmatch(text)
 
-    def translatable(node):
+    def translatable(node, marked_translate_inline=False):
```
HydrionBurst: NIT: I feel `marked_translate_inline` is too specific 
what about `force_inline`/`translate_inline`?

HydrionBurst (approved):
Did the use case for text hightlights trigger the error for the old implementation of the PR?
If no, we should test them e.g. `<a class="o_translate_inline text-muted"><div> xxx </div></a>` should translate everything as a whole

Also, since there are some `class="o_translate_inline" t-xxx are in the xml, could you
1. check all their behaviors in website TRANSLATE mode after the change?
2. at least in master, remove the `o_translate_inline` since it doesn't work

odoo/tools/translate.py
```diff
@@ -210,40 +210,48 @@ def nonspace(text):
         """ Return whether ``text`` is a string with non-space characters. """
         return bool(text) and not space_pattern.fullmatch(text)
 
-    def translatable(node):
+    def is_force_inline(node):
+        """ Return whether ``node`` should be translated has one term """
```
HydrionBurst:
```suggestion
        """ Return whether ``node`` should be translated as one term """
```

odoo/addons/base/tests/test_translate.py
```diff
@@ -329,6 +384,46 @@ def test_translate_html_i(self):
...
         self.assertEqual(result, source)
 
+    def test_force_inline_translation(self):
+        """ Test xml_translate() with elements translated as a whole (using the
+            `o_translate_inline` class).
+        """
+        terms = []
+        source = """<form>
+                        <p>{}</p>
+                    </form>""".format
```
HydrionBurst: is the o_translate_inline designed to be used in the `<form>`?

I understand you want to avoid nested format like `a % (b % c)` 
But `source(content_v1('o_translate_inline'))` is still not intuitive.
Maybe you can try expression like
`make_form/xml(content_v1 % 'o_translate_inline')`

odoo/addons/base/tests/test_translate.py
```diff
@@ -153,6 +153,61 @@ def test_translate_xml_inline5(self):
         self.assertEqual(result, source)
         self.assertItemsEqual(terms, ['Form stuff'])
 
+    def test_translate_xml_inline6(self):
```
HydrionBurst: these 2 tests are for `o_translate_inline`
could you make their name similar to `test_force_inline_translation`
and put them together?

May be name them like
`test_translate_xml_o_translate_inline_xxx`

odoo/addons/base/tests/test_translate.py
```diff
@@ -153,6 +153,61 @@ def test_translate_xml_inline5(self):
...
+                    </div>"""
+        result = xml_translate(terms.append, source)
+        self.assertEqual(result, source)
+        self.assertItemsEqual(terms, ["""<span class="o_text_highlight o_text_highlight_underline o_translate_inline">
+                            <span class="o_text_highlight_item">
+                                Here is a <a>link</a> in highlight
+                            </span>
+                        </span>"""])
+
+    def test_translate_xml_groups(self):
```
HydrionBurst, line 211: Could you also add a small test when the xml has both
`o_translate_inline` and `group`?

HydrionBurst (commented):
LGTM, just some nitpickings

odoo/addons/base/tests/test_translate.py
blse-odoo: > is the o_translate_inline designed to be used in the `<form>`?

Nope, that's from copying the other tests :sweat_smile:

odoo/addons/base/tests/test_translate.py
```diff
@@ -329,6 +394,46 @@ def test_translate_html_i(self):
         result = html_translate(lambda term: term, source)
         self.assertEqual(result, source)
 
+    def test_force_inline_translation(self):
```
HydrionBurst: I think this test is covered by all previous tests and not needed?

HydrionBurst (approved):
@robodoo delegate+

odoo/tools/translate.py
```diff
@@ -210,40 +210,48 @@ def nonspace(text):
         """ Return whether ``text`` is a string with non-space characters. """
         return bool(text) and not space_pattern.fullmatch(text)
 
-    def translatable(node):
+    def is_force_inline(node):
+        """ Return whether ``node`` should be translated as one term """
```
HydrionBurst: maybe improve this comment

HydrionBurst (approved):
LGTM

PR #209103, opened 2025-05-08
[FIX] website_sale{_loyalty},payment_stripe: fix express checkout price

vchu-odoo:
Steps to reproduce:
1) Create a free shipping reward
2) Enable standard delivery and set price to 100
3) Configure stripe provider with express checkout
4) Go to /shop
5) Add a product to the cart with the price less than 100
6) Go to the checkout and apply reward
7) Go back to the cart and observe a traceback

Reason: delivery_amount was calculated without a discount and in express checkout form `_getOrderDetails` this amount was subtracted from the order total which resulted in negative value

Solution:
Add discount if any

opw-4648641

PR #246355, opened 2026-01-29
[FIX] payment: allow zero minor amount in express checkout form

mma1377:
### Issue:
Due to this issue, there is a traceback, in express stripe checkout when there is a 100% discount and shipping cost.

Steps to reproduce:
1- Setup `Stripe` as the only payment provider in the db.
2- Create a discount code with 100% discount on order.
3- Set a fixed price on the delivery method.
4- On shop, add a product to cart, and then enter the discount code.
5- Refresh the page.

There is a traceback.

#### Cause:
This issue is introduced after #209103, which set `minor_amount` to amount excluding delivery. In the case it's 0, it's not going to be rendered on express_checkout view as it is false.

Which cause a traceback here:
https://github.com/odoo/odoo/blob/8f7e3d588c6c1189e64442a33037666d5d70aae7/addons/payment_stripe/static/src/js/express_checkout_form.js#L73-L81

opw-5482646

### DIFF
--- a/addons/payment_stripe/static/src/js/express_checkout_form.js
+++ b/addons/payment_stripe/static/src/js/express_checkout_form.js
@@ -17,7 +17,7 @@ paymentExpressCheckoutForm.include({
      */
     _getOrderDetails(deliveryAmount, amountFreeShipping) {
         const pending = this.paymentContext['shippingInfoRequired'] && deliveryAmount === undefined;
-        let minorAmount = parseInt(this.paymentContext['minorAmount'])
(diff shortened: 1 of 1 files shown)


vchu-odoo (commented):
Thank you for the fix!
Will not it be better to fix it directly in js like `parseInt(this.paymentContext['minorAmount'] || 0);` ?
That way no template change is needed.

vchu-odoo (approved):
Thank you for the changes!

@robodoo r+

PR #246356, opened 2026-01-29
[PERF] account: speed up _get_total_amount_using_same_currency

mohannedah:
Before this commit, retrieving the total amount used nested loops—iterating over moves and filtering move lines from `batch_result['lines']` in $O(N \times M)$ complexity.
This caused significant performance bottlenecks during mass payment creation.

In this commit, I refactored the logic to iterate over batch_result['lines'] in a single pass, reducing complexity to $O(M)$.

The benchmark done below was on a database where the user tried to mass register a payment for **1200** moves and resulted in **12000** move_lines.

| Scenario | Time (seconds) |
|-----------|----------------|
| **Before** | **120s** |
| **After**  | **3s** |

opw-5440956

### DIFF
--- a/addons/account/wizard/account_payment_register.py
+++ b/addons/account/wizard/account_payment_register.py
@@ -505,15 +505,12 @@ class AccountPaymentRegister(models.TransientModel):
         self.ensure_one()
         amount = 0.0
         mode = False
-        moves = batch_result['lines'].mapped('move_id')
-        for move in moves:
+        for move, lines in batch_result['lines'].grouped('move_id').items():
             if early_payment_discount and move._is_eligible_for_early_payment_discount(move.currency_id, self.payment_date):
-                for aml in batch_result['lines'].filtered(lambda l: l.move_id.id == move.id):
-                    amount += aml.discount_amount_currency
                 mode = 'early_payment'
+                amount += sum(aml.discount_amount_currency for aml in lines)
             else:
-                for aml in batch_result['lines'].filtered(lambda l: l.move_id.id == move.id):
-                    amount += aml.amount_residual_currency
+                amount += sum(aml.amount_residual_currency for aml in lines)
         return abs(amount), mode
 
     def _get_total_amount_in_wizard_currency_to_full_reconcile(self, batch_result, early_payment_discount=True):


addons/account/wizard/account_payment_register.py
```diff
@@ -505,15 +505,19 @@ def _get_total_amount_using_same_currency(self, batch_result, early_payment_disc
...
+            if is_move_eligible_for_discount:
+                moves_eligible_for_discount.add(move)
+
+            if is_move_eligible_for_discount:
                 mode = 'early_payment'
+                amount += line.discount_amount_currency
             else:
-                for aml in batch_result['lines'].filtered(lambda l: l.move_id.id == move.id):
-                    amount += aml.amount_residual_currency
+                amount += line.amount_residual_currency
```
pivi-odoo:
```py
def _get_total_amount_using_same_currency(self, batch_result, early_payment_discount=True):
    self.ensure_one()
    amount = 0.0
    mode = False
    for move, lines in batch_result['lines'].grouped('move_id').items():
        if early_payment_discount and move._is_eligible_for_early_payment_discount(move.currency_id, self.payment_date):
            amount += sum(aml.discount_amount_currency for aml in lines)
            mode = 'early_payment'
        else:
            amount += sum(aml.amount_residual_currency for aml in lines)
    return abs(amount), mode
```

Why not just do this instead? Should fix the performance issue (could you double check), since we avoid the `filtered` that were being done, which was the slow part.

pivi-odoo (changes requested):
Thank you for your patch @mohannedah!

Left a comment, let me know if you have questions.

Remark (no hard feelings): Commit/PR description

Your commit msg is a literal "wall-of-text" with too much technical details and procedural explanations that requires quite the reading/re-reading to get an idea what was the issue, and how it was fixed.
Remember that the reader can understand the *what* by reading the diff, the important part that they can't always guess is the *why*.

In your case, the explanation is technically correct but it could've been resumed into something like:

> Previously, the method used nested loops to calculate the total amount: iterating over moves and then filtering **all** move lines for each move. This resulted in $O(N \times M)$ complexity, causing significant slowdowns during mass payment creation.
> 
> This commit flattens the logic to a single loop over move lines $O(M)$. It also introduces a local cache for move eligibility checks to ensure `_is_eligible_for_early_payment_discount` is only called once per distinct move.

With 2 sentences, it gives the gist of the idea, what was the issue, and how it was fixed.

Pro tip: *in doubt, LLMs are great at rephrasing/resuming text, it's good way to get some "3rd party" opinion*

addons/account/wizard/account_payment_register.py
mohannedah: Yes in fact it should, it is the same idea. And your version is better since I don't have to create a set to cache the result of the function call ` _is_eligible_for_early_payment_discount`. I will modify it.

pivi-odoo (approved):
LGTM now, thank you for your work @mohannedah 

Feel free to open the PR for RD review.

addons/account/wizard/account_payment_register.py
```diff
@@ -505,15 +505,12 @@ def _get_total_amount_using_same_currency(self, batch_result, early_payment_disc
         self.ensure_one()
         amount = 0.0
         mode = False
-        moves = batch_result['lines'].mapped('move_id')
-        for move in moves:
+        for move, lines in batch_result['lines'].grouped('move_id').items():
```
epictete, line 508: Couldn't we use `read_group()` instead of `grouped()`, so we can perform the aggregation up front (not sure if it would improve performance though)?

addons/account/wizard/account_payment_register.py
```diff
@@ -505,15 +505,12 @@ def _get_total_amount_using_same_currency(self, batch_result, early_payment_disc
...
             if early_payment_discount and move._is_eligible_for_early_payment_discount(move.currency_id, self.payment_date):
-                for aml in batch_result['lines'].filtered(lambda l: l.move_id.id == move.id):
-                    amount += aml.discount_amount_currency
                 mode = 'early_payment'
+                amount += sum(aml.discount_amount_currency for aml in lines)
```
epictete, line 511:
```suggestion
                amount += sum(aml.mapped('discount_amount_currency')
```
Purely cosmetic (feel free to ignore :smile:)

epictete (commented):
Thanks for the fix :+1:

addons/account/wizard/account_payment_register.py
mohannedah: Performing a `read_group` will query the database to sum up the whole amount. I believe we might benefit from the `read_group` if the fields are not in the cache at this stage of the processing. The key in the `read_group` will be `move_id.id` and a get operation on this field for the recordset `batch_result['lines']` will result into a cache-miss which will fetch all the fields anyways. And since, we have the fields we can directly sum it in the method instead of doing an extra query to the database.

What do you think?

addons/account/wizard/account_payment_register.py
mohannedah: I will modify it. I think it is simpler your way.

addons/account/wizard/account_payment_register.py
epictete: Not sure either about the potential gain (or loss) of the `read_group()`, so let's keep it as is for now :relaxed:

epictete (approved):
@robodoo r+

PR #239645, opened 2025-12-11
[IMP] website: translate search result names in search bar

subhadeepdev:
When using the website search bar on a multilingual site, the names of matched webpages were always displayed in the default language instead of the active language.

This commit adds the `name` field as a translatable field to the relevant model so that search results are shown in the active language.

Steps to reproduce:
1. Configure a multilingual website (e.g., English and French with English as the default language).
2. Switch the website language to French.
3. Use the search bar and search for “Contact us”.
4. Notice that the result shows the webpage name in English instead of French.

**Task:** [5063002](https://www.odoo.com/odoo/project/974/tasks/5063002)

PR #246357, opened 2026-01-29
[IMP] core: support write dict to translated fields

HydrionBurst:
[IMP] core: write dict to translated fields

Add support for assigning a dict to create/write translated fields:

    record.translated_field = {
        lang1: value1,
        lang2: value2,
        ...
    }

Field value update behavior:
* For model translated fields with ``field.translate is True``:
  only the specified language values are updated.
* For model_terms translated fields with ``callable(field.translate)``:
  - all values must have the same HTML/XML structure
  - if a language value equals the previous one, the user confirms the
    old value and all new terms are treated as new translated terms.
    Terms for unspecified languages won't be changed.
  - otherwise, the new term in value1 is used as fallback for
    unspecified language values

Computed fields:
* Non-stored computed translated fields can also be updated with a dict.
  All updated values are propagated to the original field.
* For computed fields depending on translated fields, the compute method
  is triggered only once.

Constraints:
* Constraint functions are triggered only once.

Technical note:
``class StoredTranslations`` is a dictionary-compatible value used for
transaction cache. It enforces logical constraints for translated field
values and maintains HTML/XML structure consistency for model_terms
translated fields. The ``BaseString.write`` code mainly focuses on cache
consistency.

https://github.com/odoo/enterprise/pull/110290

Current behavior before PR:

---


odoo/fields/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 from odoo.orm.fields_misc import Id, Json, Boolean
 from odoo.orm.fields_numeric import Integer, Float, Monetary
-from odoo.orm.fields_textual import Char, Text, Html
+from odoo.orm.fields_textual import Char, Text, Html, StoredTranslations
```
kmagusiak: Can you avoid exporting it?

odoo/orm/fields_textual.py
```diff
@@ -127,13 +128,16 @@ def _convert_db_column(self, model, column):
         else:
             sql.convert_column(model.env.cr, model._table, self.name, self.column_type[1])
 
-    def get_trans_terms(self, value):
-        """ Return the sequence of terms to translate found in `value`. """
-        if not callable(self.translate):
-            return [value] if value else []
+    def get_trans_terms(self, value, placeholder='x'):
```
kmagusiak: Nitpick: Would you mind adding typing info?

odoo/addons/test_orm/tests/test_related_translation.py
```diff
@@ -347,3 +348,92 @@ def test_new_records_html(self):
...
         self.assertEqual(record_en.html, '<p>Knife</p><p>Fork</p><p>Spoon</p>')
         self.assertEqual(record_fr.html, '<p>Couteau</p><p>Fourchette</p><p>Cuiller</p>')
+    
+    def test_write_translated_dict(self):
+        self.env['res.lang']._activate_lang('nl_NL')
+
+        test1_en = self.test1.with_context(lang='en_US')
+        test1_fr = self.test1.with_context(lang='fr_FR')
+        test1_nl = test1_fr.with_context(lang='nl_NL')
+        test1_en.name = {'fr_FR': 'Couteau 2'}
```
kmagusiak, line 436: I thought that writing a dict would replace all translations, not update the dict in cache. So I expect "Knife" to be gone. I think that after this you should have `StoredTranslation({"fr_fr": "Couteau 2", "en_US": "Couteau 2})` in cache.

odoo/orm/fields_textual.py
```diff
@@ -342,12 +348,58 @@ def _update_cache(self, records, cache_value, dirty=False):
                 return
         super()._update_cache(records, cache_value, dirty)
 
+    def _inverse_related(self, records: BaseModel) -> None:
+        if not self.translate:
+            return super()._inverse_related(records)
+        # store record values, otherwise they may be lost by cache invalidation!
```
kmagusiak, line 294: This is quite error prone if we change the implementation in Field. Can't you just call super with prefetch_langs=True?

odoo/orm/fields_textual.py
```diff
@@ -342,12 +348,58 @@ def _update_cache(self, records, cache_value, dirty=False):
...
     def write(self, records, value):
         if not self.translate or value is False or value is None:
             super().write(records, value)
             return
-        cache_value = self.convert_to_cache(value, records)
-        records = self._filter_not_equal(records, cache_value)
+
+        records.env.remove_to_compute(self, records)
+        lang = records.env.lang or 'en_US'
+        if isinstance(value, StoredTranslations):
```
kmagusiak: I think that dict and StoredTranslation should be handled the same way. (see my comment in tests)
I would expect a dict to be handled this way. Note that calls from RPC cannot set StoredTranslations directly and for python code, I think that devs should read all translations and update the dict as they prefer. This may simplify the implementation as well.

kmagusiak (commented):
First review. I did not go over HTML fields.

odoo/orm/fields_textual.py
HydrionBurst: Not that easy, the main problem is
1. we don't want record values to be dict
2. we want to populate written dict data from a related field to its original field

odoo/addons/test_orm/tests/test_related_translation.py
HydrionBurst: We can discuss the feature next Monday IRL

I was thinking about the future use case for tanslation importing.
Assume you have 20 languages in the database, and as a translator you may understand 2 or 3.
So you export those languaguas in csv with column title like `name@en_US`, `name@fr_FR`, `name@nl_NL` and translate.
When you import you only select `name@fr_FR`, `name@nl_NL`
the write value is expected to be `{"fr_FR": "fr", "nl_NL": "nl"}`
You don't expect translation for "zh_CN" to be lost and fallback to "fr" after importing.

HydrionBurst (commented):
@kmagusiak Thanks for the quick review for the POC. We can discuss those details on Monday next week IRL

odoo/orm/fields_textual.py
```diff
@@ -342,12 +348,58 @@ def _update_cache(self, records, cache_value, dirty=False):
...
+            # TODO: I expect the value is from ``StoredTranslations(field._get_stored_translations(record))``
```
kmagusiak: Note: this is a feature only for inversing related fields.
If only for a test, remove it.

odoo/modules/loading.py
```diff
@@ -595,9 +595,8 @@ def load_modules(
             cr.execute(
                 """
                 INSERT INTO ir_config_parameter(key, value)
-                SELECT 'base.partially_updated_database', '1'
-                WHERE EXISTS(SELECT FROM ir_module_module WHERE state IN ('to upgrade', 'to install', 'to remove'))
-                ON CONFLICT DO NOTHING
+                VALUES ('base.partially_updated_database', '1')
```
kmagusiak: Why do you change this here?

odoo/orm/fields_textual.py
```diff
@@ -87,6 +88,13 @@ def __contains__(self, key):
 
     def get(self, key, default=None):
         return self[key]
+    
+    def update_translations(self, translations: dict[str, str]):
+        for lang, value in translations.items():
+            assert not lang.startswith('_')
+            self.pop(f'_{lang}', None)
+            if self[lang] != value:
+                self[f'_{lang}'] = value
```
kmagusiak:
```suggestion
            if self[lang] != value:
                self[f'_{lang}'] = value
            else:
                self.pop(f'_{lang}', None)
```

odoo/orm/fields_textual.py
```diff
@@ -162,6 +189,11 @@ def convert_to_cache(self, value, record, validate=True):
         if value is None or value is False:
             return None
 
+        if self.translate and isinstance(value, dict):
+            # happen after INSERT SQL in _create
```
kmagusiak: Convert to cache may be used in other contexts and where this drops values.
If today it is only called there, it does not mean it is not going to change.

HydrionBurst: Agree, that's why I also add the logic for HTML fields which doesn't call `super()`

I am thinking how to make the comment more generic

Now for translated field I define
1. cache format: str | None
2. field data value: dict | None

odoo/addons/test_orm/tests/test_related_translation.py
```diff
@@ -347,3 +349,225 @@ def test_new_records_html(self):
...
+        for _ in range(2):
+            # first time check cache value
+            # second time check refetched/recomputed value
+            for field_name, expected_dict in expected.items():
+                if not expected_dict:
+                    self.assertEqual(record[field_name], False)
+                else:
+                    for lang, value in expected_dict.items():
+                        self.assertEqual(record.with_context(lang=lang)[field_name], value)
+            record.invalidate_recordset()
```
kmagusiak, line 373: Should this be indented?

odoo/addons/test_orm/models/test_orm.py
```diff
@@ -9,6 +9,8 @@
 from odoo.tools.translate import html_translate
 import itertools
 
+from odoo.fields import StoredTranslations
```
kmagusiak: Please remove it from exports, this is part of the cache implementation.

odoo/orm/fields_textual.py
```diff
@@ -51,7 +52,17 @@ class StoredTranslations(dict):
...
+    def fallback_langs(lang: str) -> tuple[str, ...]:
+        if lang == '_en_US':
+            return '_en_US', 'en_US'
+        if lang == 'en_US':
+            return ('en_US',)
+        if lang.startswith('_'):
+            return lang, lang[1:], '_en_US', 'en_US'
+        return lang, 'en_US'
+
+    def __getitem__(self, key: str) -> str:
```
kmagusiak: The signature is not compatible with super, same for some other overrides.
```suggestion
    def __getitem__(self, key) -> str:
```

odoo/orm/fields_textual.py
```diff
@@ -82,12 +93,201 @@ def __getitem__(self, key):
...
+
+        if callable(field.translate):
+            self.terms: list[str] = []
+
+            def translate_func(term):
+                self.terms.append(term)
+                return 'X'
+
+            self.structure: str = field.translate(translate_func, value)
+
```
kmagusiak: Always initialize terms and structure (avoids having to use hasattr) or failing in get_trans_terms. Unless this is wanted, then it would be better to raise an exception explicitely.

After reading the rest, it would be simpler to start the init with also `assert callable(field.translate)`.

odoo/orm/fields_textual.py
```diff
@@ -127,29 +327,33 @@ def _convert_db_column(self, model, column):
...
-    def get_text_content(self, term):
-        """ Return the textual content for the given term. """
-        func = getattr(self.translate, 'get_text_content', lambda term: term)
-        return func(term)
+    def get_trans_terms(self, value: str) -> list[str]:
+        """ Return the sequence of terms to translate found in `value`"""
+        return ParsedTranslation(value, self).terms
 
     def convert_to_column(self, value, record, values=None, validate=True):
+        assert not (self.translate and isinstance(value, dict))
```
kmagusiak, line 78: Can you explain why?

odoo/orm/fields_textual.py
```diff
@@ -127,29 +327,33 @@ def _convert_db_column(self, model, column):
...
+                if k in active_langs or k == 'en_US'
+            }
+            if not cache_value_dict:
+                _logger.warning("No translation in value %s for translated field %s", value, self.name)
                 return None
-            return PsycopgJson({'en_US': value, record.env.lang or 'en_US': value})
-        return super().convert_to_column_insert(value, record, values, validate)
+        cache_value_dict.setdefault('en_US', cache_value_dict.get(lang, next(iter(cache_value_dict.values()))))
+        StoredTranslations(cache_value_dict).validate(record.env, self, check_structure=callable(self.translate))
+        return PsycopgJson(cache_value_dict)
```
kmagusiak: Just wondering if `PsycopgJson(StoredTranslations(cache_value_dict))` would work.

odoo/orm/fields_textual.py
```diff
@@ -679,6 +902,9 @@ def convert_to_column(self, value, record, values=None, validate=True):
         return super().convert_to_column(value, record, values, validate=False)
 
     def convert_to_cache(self, value, record, validate=True):
+        if self.translate and isinstance(value, dict):
+            lang = record.env.lang or 'en_US'
+            value = value.get(lang, next(iter(value.values())))
```
kmagusiak: It is slightly different than in other places. I would expect.
`value.get(record.env.lang, value.get('en_US', next(iter(value.values()))))`

odoo/orm/fields_textual.py
kmagusiak: This still seems wrong as the function is just a copy-paste of the super method.
I think the best is still to read all translations of the related field and call super with prefetch_langs=True so that the dict is just copied.

odoo/orm/fields_textual.py
```diff
@@ -82,12 +93,201 @@ def __getitem__(self, key):
...
-    def get(self, key, default=None):
+    def copy(self) -> StoredTranslations:
+        return StoredTranslations(self)
+
+    def get(self, key: str, default=None) -> str:
         return self[key]
 
+    def validate(self, env: Environment, field: BaseString, *, check_structure: bool = False, auto_fix: bool = False):
+        """Validate stored translations against soft constraints for the translated field"""
+        translation_dict = dict(self)
```
kmagusiak: Nitpick: you can replace this variable by `self.keys()`

odoo/orm/fields_textual.py
```diff
@@ -342,122 +541,146 @@ def _update_cache(self, records, cache_value, dirty=False):
...
     def write(self, records, value):
         if not self.translate or value is False or value is None:
             super().write(records, value)
             return
-        cache_value = self.convert_to_cache(value, records)
-        records = self._filter_not_equal(records, cache_value)
+
+        records.env.remove_to_compute(self, records)
+        lang = records.env.lang or 'en_US'
+        if isinstance(value, StoredTranslations):
```
kmagusiak: If think you should add here:
```py
if records.env.context.get('prefetch_langs') and not isinstance(value, StoredTranslations):
  value = StoredTranslations(value)
  value.validate(...)
```

odoo/orm/fields_textual.py
HydrionBurst: If someone write a domain condition `('name', '=', {'en_US': 'value_en', 'fr_FR': 'value_fr'})`, I don't want to silently make it
work like `('name', '=',  'value_fr')` when `env.lang == 'fr_FR'`

However, you cannot really have domain condition `('name', '=', {'en_US': 'value_en', 'fr_FR': 'value_fr'})` for now, because [this](https://github.com/odoo/odoo/blob/d368f891f60653d1a81a9af0aaad9509ec6a4a79/odoo/orm/domains.py#L1306) ask the `value` to be hashable

odoo/orm/fields_textual.py
HydrionBurst: I hope all apis for `StoredTranslations` can work for both `field.translate is True` and `callable(field.translate)`. So it would be better to avoid `assert callable(field.translate)`

odoo/orm/fields_textual.py
HydrionBurst: We cannot, because
1. I will change `self` when autofix
2. I need the old value  for log and error message

odoo/orm/fields_textual.py
HydrionBurst: It will work and has already been the case in master
1. when write we save the `StoredTranslation` object to cache with dirty
2. when flush, the object is wrapped by `PsycopgJson` in `get_column_update`

odoo/orm/fields_textual.py
HydrionBurst: I drop the support for write `StoredTranslations` leave the feature to future

odoo/orm/fields_textual.py
```diff
@@ -342,122 +542,146 @@ def _update_cache(self, records, cache_value, dirty=False):
...
                 for record in records:
-                    super()._update_cache(record, dict(cache_value), dirty)
+                    super()._update_cache(record, cache_value.copy(), dirty)
                 return
         super()._update_cache(records, cache_value, dirty)
 
+    def _inverse_related(self, records):
+        if not self.translate:
+            return super()._inverse_related(records)
+        # store record values, otherwise they may be lost by cache invalidation!
```
HydrionBurst, line 294: The only difference between this method and its `super()`
the `super` uses the record value of the related field to write its original field
this method uses the value (dict | None) saved in cache to write its original field

Since we don't have stored related translated field.
Here the related field must not be stored, so all translations are only available in the cache.

odoo/addons/test_orm/tests/test_related_translation.py
HydrionBurst: Yes, as I commented before.
The first check is done immediately after create/write which checks cache value if there is. Then I invalidate the cache and check the value we refetched/recomputed.

odoo/addons/base/tests/test_translate.py
kmagusiak: I don't think we should allow doing such things.
Note that the regex checks start of the line, so you could just indent it if you really must import it.

odoo/orm/fields_textual.py
kmagusiak: You handle callable `translate` differently in write: you call `write_dict` for callable translations only.
I find it just confusing to use this class for non-callable translations. Also this class does not do anything useful for for translate=True.

odoo/orm/fields_textual.py
kmagusiak: Indeed, domains work for str for comparison. Could you add it as a comment in the code?

odoo/orm/fields_textual.py
HydrionBurst: Design: Decouple data integrity from cache management for translated fields.

 - BaseString: Ensures consistency between the database and the cache.
 - StoredTranslations: Ensures integrity of translated data (HTML/XML term sync).
   StoredTranslations.validate() documents and enforces the constraints.

 `StoredTranslations` API must support both `field.translate is True` and
 `callable(field.translate)`. ParsedTranslation is the value type for `translation_dict` in `StoredTranslations.write_dict()`.

 `BaseString` handles `field.translate is True` without `write_dict` as an optimization, avoiding StoredTranslations prefetch. ParsedTranslation is still implemented for `field.translate is True` to achieve full test coverage of StoredTranslations.

kmagusiak: As you say, since `BaseString` handles translate = True, `write_dict` is just there for callable translations. `StoredTranslations` is private, so it's just an implementation detail. Why do you want it to support something that is not used?

addons/website_sale/models/product_template.py
```diff
@@ -260,12 +260,14 @@ def _compute_variants_default_code(self):
...
-        if (
-            (description_ecommerce := vals.get('description_ecommerce'))
-            and is_html_empty(description_ecommerce)
-            and not ('media_iframe_video' in description_ecommerce or 'data-embedded' in description_ecommerce)  # don't remove "empty" video div
-        ):
-            vals['description_ecommerce'] = ''
+        if (description_ecommerce := vals.get('description_ecommerce')):
+            if isinstance(description_ecommerce, dict):
+                for lang, value in description_ecommerce.items():
+                    if is_html_empty(value) and not ('media_iframe_video' in value or 'data-embedded' in value):
```
ltinel: Nit (to avoid too long lines):
```suggestion
                    if (
                        is_html_empty(value)
                        and not ('media_iframe_video' in value or 'data-embedded' in value)
                    ):
```
Same below.
Note that you could also extract the duplicated logic, but it's probably not worth it.

ltinel (approved):
LGTM for `website_sale`, thanks :)

odoo/addons/test_tools/tests/test_translate.py
```diff
@@ -2054,3 +2608,56 @@ def test_pattern_conversion(self):
         ]
         for original_pattern, escaped_pattern, message in cases:
             self.assertEqual(sql.pattern_to_translated_trigram_pattern(original_pattern), escaped_pattern, message)
+
+
+class TestTranslatedFields(TransactionCase):
+    def test_write_override_translated_field(self):
```
HydrionBurst: A test to find `write` for translated fields.

This one is quite naive but useful. We may or may not need to merge it.

I use it with another technically wrong but useful test
```py
    def test_translated_fields(self):
        for model in self.env.values():
            for field_name, field in model._fields.items():
                with self.env.cr.savepoint() as sp:
                    try:
                        if field.translate:
                            n = model.new()                    
                            n.write({field_name: {'en_US': '<p>bbb</p>'}})
                            if n[field_name] != field.convert_to_cache('<p>bbb</p>', n):
                                pass
                    except Exception as e:
                        import traceback
                        traceback.print_exception(type(e), e, e.__traceback__)
                        pass
                    sp.rollback()
```

odoo/tools/translate.py
```diff
@@ -422,6 +424,266 @@ def is_text(term):
...
+    def __contains__(self, key) -> bool:
+        return True
+
+    def copy(self) -> StoredTranslations:
+        return StoredTranslations(self)
+
+    def get(self, key, default=None):
+        return self[key]
+
+    def validated(self, env: Environment, field: BaseString, *, check_structure: bool = False, auto_fix: bool = False) -> StoredTranslations | None:
```
rco-odoo: This is fundamentally two different APIs packed together:
- one for **checking** whether a `StoredTranslations` is correct;
- one for **fixing up** a `StoredTranslations` (possibly in-place).

I suggest the following APIs instead:
- `stored_translations.validate(env, field)` checks for soft constraints (does not modify, only raises exceptions);
- `stored_translations.fixup(env, field)` fixes up the dict to satisfy the soft constraints; you may get a fixed-up copy by combining `copy()` with that method.

Checking the structure should be automatic and depend on whether `field.translate` is callable.  For the use-case of `written`, you may add a flag to disable the check (like an explicit `check_structure=False` to enable the optimization).

odoo/tools/translate.py
```diff
@@ -422,6 +424,266 @@ def is_text(term):
...
+
+        if issues:
+            _logger.info(
+                "Auto fix StoredTranslations for field %s from %s to %s because:\n%s",
+                field, self, valid_self, ",\n".join(issues)
+            )
+
+        return valid_self
+
+    def written(self, env: Environment, field: BaseString, translation_dict: dict[str, ParsedTranslation], *, adapt_close_terms: bool = False, delay_translations: bool = False) -> StoredTranslations | None:
```
rco-odoo, line 595: What is `adapt_close_terms` for?
It looks like the boolean flags are there for factoring the implementation of several methods.  It's an implementation question, and should not show up in the caller API.

I also suggest to have a more "direct" API: call the method `write()` and make it work in-place.
If you need a copy, make a copy.  Be explicit.

odoo/tools/translate.py
```diff
@@ -422,6 +426,291 @@ def is_text(term):
...
+        self.terms: list[str] = []
+
+        def translate_func(term):
+            self.terms.append(term)
+            return 'X'
+
+        self.structure = field.translate(translate_func, value)
+
+
+def adapt_translated_field_val(
```
rco-odoo:
```suggestion
def adapt_translated_field_value(
```

rco-odoo (commented):
TBH the code is quite obscure (partly due to my lack of courage to read everything in detail), and the chosen API's don't help much.  That's why I suggested using different APIs for doing different things.  Besides that, it's a very nice refactoring of all the mechanisms behind translated fields!

rco-odoo (approved):
@robodoo delegate+

addons/website/models/website_page.py
```diff
@@ -189,35 +189,48 @@ def unlink(self):
...
+            url = '/' + self.env['ir.http']._slugify(url, max_length=1024, path=True)
+            vals_url = url
+
+        if 'visibility' in vals:
+            if vals['visibility'] != 'restricted_group':
+                vals['group_ids'] = False
 
-            # If name has changed, check for key uniqueness
-            if 'name' in vals and page.name != vals['name']:
-                vals['key'] = self.env['website'].with_context(website_id=website_id).get_unique_key(self.env['ir.http']._slugify(vals['name'] or ''))
```
HydrionBurst, line 214: @subhadeepdev @h4818 

Could you have a review for this commit? (you don't need to check the whole PR)

I have problems with [this line](https://github.com/odoo/odoo/blob/cca85df8590882470303696ddcf78a2829b7b2a1/addons/website/models/website_page.py#L214) when adapting code for [this PR](https://github.com/odoo/odoo/pull/246357) especially after a recent [IMP](https://github.com/odoo/odoo/pull/239645)

1. before the [IMP](https://github.com/odoo/odoo/pull/239645), the code doesn't make sense in batch, `page.name != vals['name']` is for one `page`, but the changed `vals['key']` is for all pages.

2. after the [IMP](https://github.com/odoo/odoo/pull/239645), do we really want to change `page.key` for the last updated language value of `page.name`?  In this commit, I just populate en_US value.

h4818: @HydrionBurst 

1. Indeed you're right! That line was not defining a unique key for each page that is being updated. So there is a bug. The original idea was to keep the key aligned to the view name.
2. Simple answer: no, we don't need to change the key for the last updated language. After the [IMP](https://github.com/odoo/odoo/pull/239645), the view name is kept in sync with the en_US page name (or default lang if there is no en_US). So we don't care to keep the view name aligned to any other translation, and in a similar way, we don't care about the translations for the view key either.

so basically - if the en_US (or default) page name changes, we should update the view and key too.
if a translated page name changes - there's no need to change neither the view name nor key.

h4818 (approved):
`website` changes LGTM. Thanks 👍 
Seems like this PR solves the bug in 1 and;
in 2. we only need to update the view `name` and `key` if the en_US value changed

PR #246363, opened 2026-01-29
[FIX] hr: add empty page for certification view

adak-ahmed:
[FIX] hr: add empty page for certification view

**Issue:**
-Creating a new Certification triggers a validation error if no Skill Type has "is_certification" checked.

**To reproduce:**
1-Navigate to Skill Types in Configuration Menu.
2-Disable "is_certificate" for all types.
3-Navigate to Certifications menu.
4-Try to create a new Certification for any Employee.

**Solution:**
-Empty page will show, with a button guiding to the Skill Types list view for adjustment / creation if no certifications available.

-**Task**: #5350163

### DIFF
--- a/addons/hr_skills/models/hr_employee_skill.py
+++ b/addons/hr_skills/models/hr_employee_skill.py
@@ -45,5 +45,27 @@ class HrEmployeeSkill(models.Model):
             'views': [(self.env.ref('hr_skills.employee_skill_view_inherit_certificate_form').id, 'form')],
         }
 
+    def action_hr_employee_skill_certification(self):
+        skill_type = self.env["hr.skill.type"].search(
+            [("is_certification", "=", True)], limit=1
+        )
+        show_certificate_button = bool(skill_type)
+        action = self.env["ir.actions.act_window"]._for_xml_id(
+            "hr_skills.action_hr_employee_skill_certification"
+        )
+        action["context"] = {"show_certificate": show_certificate_button}
+        if self.env.user.has_group("hr.group_hr_manager"):
+            action["help"] = (
+                self.env._("""<p class="o_view_nocontent_smiling_face">No Certifications available. Navigate to Skill types!</p>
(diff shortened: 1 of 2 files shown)


addons/hr_skills/models/hr_employee_skill.py
```diff
@@ -45,5 +45,32 @@ def open_hr_employee_skill_modal(self):
...
+                (self.env.ref('hr_skills.hr_employee_skill_view_list').id, 'list'),
+                (self.env.ref('hr_skills.employee_skill_view_inherit_certificate_form').id, 'form'),
+            ],
+            'help': self.env._("""
+                <p class="o_view_nocontent_smiling_face">No Certifications available. Navigate to Skill types!</p>
+                <a type="action" name="hr_skills.hr_skill_type_action" class="btn btn-primary">
+                Show Skill Types
+                </a>
+            """),
+        }
```
tivisse: We could keep the ir.actions.act_window, and do something like this for the dynamic parameters.

```py
    def action_show_versions(self):
        self.ensure_one()
        action = self.env['ir.actions.act_window']._for_xml_id('hr_attendance.hr_version_list_view')
        action['domain'] = self._get_current_versions_domain()
        action['context'] = {'default_ruleset_id': self.id}
        return action
```

addons/hr_skills/views/hr_views.xml
```diff
@@ -576,16 +576,13 @@
         </field>
     </record>
 
-    <record id="action_hr_employee_skill_certification" model="ir.actions.act_window">
+    <record id="action_server_hr_employee_skill_certification" model="ir.actions.server">
```
tivisse, line 579: You need to write access rights on the model to execute a server action.

Can you test it why the lowest access right to ensure you won't get an access error while clicking on the Certification menu item ?

adak-ahmed: Other than admins, only officers can access the menu items, and it is working fine for them.

PR #246365, opened 2026-01-29
[FIX] html_editor: skip srcset generation on hover effect images

guja-odoo:
Bug:
The on hover image animation in the website builder was not working properly. Once the image was saved, the animation would not play.

Steps to reproduce:
- in website, in edit mode
- Drop s_picture or add any image and set it's size to 1024 or more
- On the image, add on hover animation with any effect
- See that it works properly in edit mode
- Save
- Hover on the image, animation doesn't play

Why:
When adding an image to a certain size, a srcset is added for responsiveness. But because the hover interaction swaps the image `src` attribute, and not the srcset sources that are selected by the browser, the animation was not displayed even though it was correctly playing for the image `src`.

Fix:
If there is a `data-hover-effect` in the image attributes, then we skip the srcset generation. Doing so, there is only the image `src` to display and the animation plays correctly.

task-5891353

### DIFF
--- a/addons/html_editor/static/src/main/media/image_save_plugin.js
+++ b/addons/html_editor/static/src/main/media/image_save_plugin.js
@@ -181,7 +181,12 @@ export class ImageSavePlugin extends Plugin {
                 }
             }
         }
-        if (!isImageField && !isBackground && el.dataset.mimetype !== "image/gif") {
+        if (
+            !isImageField &&
+            !isBackground &&
+            el.dataset.mimetype !== "image/gif" &&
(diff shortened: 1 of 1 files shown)


sobo-odoo:
@robodoo delegate=bvr-odoo

bvr-odoo (approved):

guja-odoo:
> @guja-odoo LGTM, you can mark the PR as ready

thanks :pray:

bvr-odoo (approved):

PR #244061, opened 2026-01-15
[FIX] sale: prevent payment when so is expired

mma1377:
## Issue:
Payment link should expire if payment is expired.

#### Steps to reproduce:
1- Create a new quotation.
2- Set the expiry date in the past.
3- Open the action menu and generate a payment link.
4- Open the payment link and pay.

Expected result: The payment should fail if the so is expired.

opw-5478691

PR #246369, opened 2026-01-29
[FIX] sale: use single wizard.res_id in compute_warning

mma1377:
### Issue:
In #244061 changes, the expired so leads to warning message.
However, the batch `res_id` is used instead of single record which is an error.

opw-5478691


Feyensv (approved):
robodoo delegate+

PR #210299, opened 2025-05-16
[IMP] stock,*: remove scrap model

mohammadhossam:
This commit removes `stock.scrap` model without affecting its
functionality from the user perspective.

Previous to this commit, when a user wants to scrap some product,
a record of `stock.scrap` is created which is a scrap order
containing info about the product, quantity, scrap reason, etc.
When a scrap order is confirmed, it generates a `stock.move` and
confirms it so that the quantity move is reflected in the quants.
That's why `stock.scrap` model itself provides no intrinsic value
and is dispensable.

This commit removes the model completely and relies only on
`stock.move.line` and `stock.move` to achieve the same functionality.
The menu item "Scrap" now redirects to a list of stock.move.lines that
has the new field `is_scrap` = true. If the user creates a new record
from the list, a draft `stock.move.line` is created initially. When it's
confirmed, it generates its `stock.move` and its state is changed to be
done.

The only fields that are added to `stock.move.line` to fully replace
`stock.scrap` are `scrap_reason_tag_ids` and
`should_replenish_scrapped`.

This commit also adds a new action to scrap a lot/serial number from its
form view directly.

Task-4606260

PR #246370, opened 2026-01-29
[FIX] stock: avoid recompute to draft state on pickings

clesgow:
Steps to reproduce:
- Create a delivery with a reserved move
- Click on the "picked" checkbox on the move

Issue:
The picking is visibly reset to draft

This happens due to #210299, which used `self.move_ids` instead of the previously used `search()`. This means that rather than having proper ids to use, now the moves are `stock.move(<NewId origin=X>)` records, and their related picking as well.
This implies that the dict uses these `NewId` as keys as well.

However, below we use `picking.ids` which contains the "real" ids (i.e. int) and not the `NewId`. So no picking is found within the dict and the returned state is always `draft`.

---

### DIFF
--- a/addons/stock/models/stock_picking.py
+++ b/addons/stock/models/stock_picking.py
@@ -814,15 +814,15 @@ class StockPicking(models.Model):
         '''
         picking_moves_state_map = defaultdict(dict)
         picking_move_lines = defaultdict(set)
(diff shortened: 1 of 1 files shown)


Whenrow:
robodoo r+

PR #246371, opened 2026-01-29
[IMP] website_*: prepare records for AI previews

AbdelrahmanFawzy1:
### Description

This PR prepares website records to be reused by flows that are not running inside a normal website page request.

It contains two changes needed by the Enterprise AI preview cards work.

#### Decouple website sale pricing from `request`

Website sale pricing, combination, and stock helpers normally rely on the global `request` object to resolve the current website, pricelist, fiscal position, and cart.

That is fragile when website prices are computed later or outside the original frontend request, such as when Enterprise AI prepares product, appointment, or paid course previews. In that flow, the AI must show the same final website prices and discounts as the public site, not raw list prices.

This PR lets the relevant helpers receive explicit pricing context through optional pricelist and fiscal position arguments, and forwards those values through the affected website sale overrides.

#### Extract reusable website card templates

The PR also extracts standalone card templates for website records such as blog posts, events, and job positions.

The existing website pages keep the same rendering, but the card markup can now be reused by Enterprise AI preview cards instead of duplicating frontend templates.

---

task-id-5153868

### DIFF
--- a/addons/website_blog/views/website_blog_posts_loop.xml
+++ b/addons/website_blog/views/website_blog_posts_loop.xml
@@ -98,50 +98,7 @@ according to the enabled options.
             <t t-foreach="posts" t-as="blog_post">
                 <!-- Assign 'colWidth': 'col-12' is default for List-View and mobile -->
                 <div t-attf-class="pb-4 col-12 #{colWidth}">
-                    <article t-attf-class="o_wblog_post position-relative #{'card h-100' if opt_blog_cards_design else ''}" name="blog_post">
-                        <!-- List-View Design -->
-                        <t t-if="opt_blog_list_view">
-                            <div t-att-class="opt_blog_cards_design and 'card-body py-3'">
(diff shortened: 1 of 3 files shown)


vchu-odoo (approved):

Feyensv (changes requested):
Hello,
I'm quite concerned about the stability and reliability of this whole approach, even more considering it's not tested ...

If we have to make the base code more complex because of a obscure use-case that hijacks the base module logic and uses, then it probably means the approach is wrong.

I can see quite easily how this new logic would break (simply extending the template in any custom module with new required rendering values would fail your new logic).
Even if it was tested, frontend qweb templates are meant to be easily overridden and customized, and rendered through specific entry points, your code should be able to work with the current code without having to modify `website_sale`.

kmagusiak:
@robodoo override=ci/style

Moving existing `request` usage within model files.

Deimort:
@robodoo r+ rebase-ff

PR #246373, opened 2026-01-29
[FIX] mail: clear mail activities before testing KPIs

madprog:
Commit ff941e6555c12918803db4a72511c9381cc58c31 introduced a test that counts the amount of mail activities grouped by type.
However, the test doesn't take into consideration that the demo data might have created some activities.

With this commit, we ensure the tests run with a clean state by removing any pre-existing mail activities so that we have consistent and expected results.

Runbot-build-error-id: [234966](https://runbot.odoo.com/odoo/runbot.build.error/234966)


madprog:
@robodoo r+

Levizar (approved):
@robodoo r+

PR #246375, opened 2026-01-29
[FIX] base: fix issue with res.partner create on a non-empty recordset

delcourtfl:
**Issue:**
`ValueError: Expected singleton` when creating new partners on a non empty (no singleton) recordset due to the `_add_missing_default_values` call.

The default create method uses `self = self.browse()` to remove the records, but the `res.partner` override was still using its original `self`.

**Fix:**
Properly call the method on an empty recordset.

related: https://github.com/odoo/odoo/commit/79486ec3fc553845cac14fb135c16fe0b093e3b4

opw-4932114


tde-banana-odoo:
@robodoo r+

PR #246377, opened 2026-01-29
[FIX] account_edi_ubl_cii: Missing schemeID for Belgium Company registry

smetl:
1) This commit adds the new generic methods to build an UBL file.

The big objective is to remove the hierarchy of UBL later on and make BIS3 completely standalone. In this commit, BIS3 is not independent of UBL 2.1 & 2.0.

2) Missing schemeID for Belgium Company registry

Since https://github.com/odoo/odoo/commit/d40652c17b43b6a9fd4d19ce84b943b2a51fad9c, in Belgium, we set by default the company registry. However the schemeID is missing and should be part of the ISO 6523 ICD list: https://docs.peppol.eu/poacc/billing/3.0/codelist/ICD/

3) Fix PartyLegalEntity when the country is NL and there is no company registry/EAS/Endpoint

That case was raising a traceback before this commit.

---


addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
```diff
@@ -0,0 +1,143 @@
...
+			<cac:PostalAddress>
+				<cbc:StreetName>Rue des Bourlottes 9</cbc:StreetName>
+				<cbc:CityName>Ramillies</cbc:CityName>
+				<cbc:PostalZone>1367</cbc:PostalZone>
+				<cac:Country>
+					<cbc:IdentificationCode>BE</cbc:IdentificationCode>
+				</cac:Country>
+			</cac:PostalAddress>
+			<cac:PartyTaxScheme>
+				<cbc:CompanyID>0477472701</cbc:CompanyID>
```
clbr-odoo, line 65: Should be BE0477472701 ?

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
```diff
@@ -0,0 +1,143 @@
...
+				<cac:TaxScheme>
+					<cbc:ID>0208</cbc:ID>
```
clbr-odoo, line 67: That's incorrect. (should be "VAT")

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_vat.xml
```diff
@@ -0,0 +1,143 @@
...
+			</cac:PostalAddress>
+			<cac:PartyTaxScheme>
+				<cbc:CompanyID>BE0477472701</cbc:CompanyID>
+				<cac:TaxScheme>
+					<cbc:ID>VAT</cbc:ID>
+				</cac:TaxScheme>
+			</cac:PartyTaxScheme>
+			<cac:PartyLegalEntity>
+				<cbc:RegistrationName>partner_be</cbc:RegistrationName>
+				<cbc:CompanyID>BE0477472701</cbc:CompanyID>
```
clbr-odoo, line 72: Don't you think we should be able to recover from 0208 <-> 9925 for the Belgians ?

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_vat_and_ref.xml
```diff
@@ -0,0 +1,147 @@
...
+			<cac:Contact>
+				<cbc:Name>company_1_data</cbc:Name>
+			</cac:Contact>
+		</cac:Party>
+	</cac:AccountingSupplierParty>
+	<cac:AccountingCustomerParty>
+		<cac:Party>
+			<cbc:EndpointID schemeID="0208">0477472701</cbc:EndpointID>
+			<cac:PartyIdentification>
+				<cbc:ID>PARTNER_BE</cbc:ID>
```
clbr-odoo, line 55: To check: Is it only a fallback ? Imo again, for Belgian, we should be able to deduce a better value.
There can be only one PartyIdentification for Customer so let's make sure we give the better information available.

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_lu_company_registry.xml
```diff
@@ -0,0 +1,144 @@
...
+				<cbc:CompanyID schemeID="0208">0202239951</cbc:CompanyID>
+			</cac:PartyLegalEntity>
+			<cac:Contact>
+				<cbc:Name>company_1_data</cbc:Name>
+			</cac:Contact>
+		</cac:Party>
+	</cac:AccountingSupplierParty>
+	<cac:AccountingCustomerParty>
+		<cac:Party>
+			<cbc:EndpointID schemeID="9938">00005000041</cbc:EndpointID>
```
clbr-odoo, line 52: That looks SUS.
9938 is the LU VAT, and should most likely be prefixed by "LU" ?

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_lu_company_registry.xml
```diff
@@ -0,0 +1,144 @@
...
+				<cbc:CityName>Luxembourg</cbc:CityName>
+				<cbc:PostalZone>L-1528</cbc:PostalZone>
+				<cac:Country>
+					<cbc:IdentificationCode>LU</cbc:IdentificationCode>
+				</cac:Country>
+			</cac:PostalAddress>
+			<cac:PartyTaxScheme>
+				<cbc:CompanyID>00005000041</cbc:CompanyID>
+				<cac:TaxScheme>
+					<cbc:ID>9938</cbc:ID>
```
clbr-odoo, line 67: Same comment as before, this looks just incorrect to me.

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_only_vat.xml
```diff
@@ -0,0 +1,143 @@
...
+				<cbc:RegistrationName>company_1_data</cbc:RegistrationName>
+				<cbc:CompanyID schemeID="0208">0202239951</cbc:CompanyID>
+			</cac:PartyLegalEntity>
+			<cac:Contact>
+				<cbc:Name>company_1_data</cbc:Name>
+			</cac:Contact>
+		</cac:Party>
+	</cac:AccountingSupplierParty>
+	<cac:AccountingCustomerParty>
+		<cac:Party>
```
clbr-odoo, line 51: This one is missing the EndpointID ⚠️

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_only_vat.xml
```diff
@@ -0,0 +1,143 @@
...
+			</cac:PostalAddress>
+			<cac:PartyTaxScheme>
+				<cbc:CompanyID>NL000099998B57</cbc:CompanyID>
+				<cac:TaxScheme>
+					<cbc:ID>VAT</cbc:ID>
+				</cac:TaxScheme>
+			</cac:PartyTaxScheme>
+			<cac:PartyLegalEntity>
+				<cbc:RegistrationName>partner_nl</cbc:RegistrationName>
+				<cbc:CompanyID>NL000099998B57</cbc:CompanyID>
```
clbr-odoo: We could use the `schemeID="9944"`.

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_vat_eas_oin_company_registry.xml
```diff
@@ -0,0 +1,144 @@
...
+			</cac:PostalAddress>
+			<cac:PartyTaxScheme>
+				<cbc:CompanyID>NL000099998B57</cbc:CompanyID>
+				<cac:TaxScheme>
+					<cbc:ID>VAT</cbc:ID>
+				</cac:TaxScheme>
+			</cac:PartyTaxScheme>
+			<cac:PartyLegalEntity>
+				<cbc:RegistrationName>partner_nl</cbc:RegistrationName>
+				<cbc:CompanyID schemeID="0106">NL000099998B57</cbc:CompanyID>
```
clbr-odoo: this is incorrect: its the KVK scheme with the vat value, while the filename refers to oin ?
I've put some "official" example from the NL Peppol Authority on IAP `file_validator` test suite if you want to check.

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_vat_eas_kvk_company_registry.xml
```diff
@@ -0,0 +1,144 @@
...
+			</cac:PartyLegalEntity>
+			<cac:Contact>
+				<cbc:Name>company_1_data</cbc:Name>
+			</cac:Contact>
+		</cac:Party>
+	</cac:AccountingSupplierParty>
+	<cac:AccountingCustomerParty>
+		<cac:Party>
+			<cbc:EndpointID schemeID="9944">NL000099998B57</cbc:EndpointID>
+			<cac:PartyName>
```
clbr-odoo, line 53: No partyidentification ?

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_outside_eu.xml
```diff
@@ -0,0 +1,136 @@
...
+				<cbc:RegistrationName>company_1_data</cbc:RegistrationName>
+				<cbc:CompanyID schemeID="0208">0202239951</cbc:CompanyID>
+			</cac:PartyLegalEntity>
+			<cac:Contact>
+				<cbc:Name>company_1_data</cbc:Name>
+			</cac:Contact>
+		</cac:Party>
+	</cac:AccountingSupplierParty>
+	<cac:AccountingCustomerParty>
+		<cac:Party>
```
clbr-odoo, line 51: No EndpointID

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -215,6 +215,28 @@ def aggregate_function(target_base_line, base_line):
...
+
+    def _ubl_add_values_delivery(self, vals, delivery):
+        vals['delivery'] = delivery
+
+    def _ubl_add_values_partner_bank(self, vals, partner_bank):
+        vals['partner_bank'] = partner_bank
+
+    def _ubl_add_values_payment_term(self, vals, payment_term):
+        vals['payment_term'] = payment_term
+        vals['payment_term_node'] = html2plaintext(payment_term.note) if payment_term else None
```
clbr-odoo:
```suggestion
        vals['payment_term_note'] = html2plaintext(payment_term.note) if payment_term else None
```
maybe ? can't find where it's used

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -862,6 +884,166 @@ def _ubl_get_line_allowance_charge_discount_node(self, vals, discount_values):
...
+        return {
+            'cbc:StreetName': {'_text': partner.street},
+            'cbc:AdditionalStreetName': {'_text': partner.street2},
+            'cbc:CityName': {'_text': partner.city},
+            'cbc:PostalZone': {'_text': partner.zip},
+            'cbc:CountrySubentity': {'_text': partner.state_id.name},
+            'cbc:CountrySubentityCode': {'_text': partner.state_id.code},
+            'cac:Country': {
+                'cbc:IdentificationCode': {'_text': partner.country_id.code},
+                'cbc:Name': {'_text': partner.country_id.name},
```
clbr-odoo, line 1195: to check: was this ever set in the test files ? 🤔

addons/account_edi_ubl_cii/models/account_edi_ubl.py
```diff
@@ -862,6 +884,166 @@ def _ubl_get_line_allowance_charge_discount_node(self, vals, discount_values):
...
+            },
+        }
+
+        if self.module_installed('account_add_gln') and delivery_partner.global_location_number:
+            node['cac:DeliveryLocation']['cbc:ID']['schemeID'] = '0088'
+            node['cac:DeliveryLocation']['cbc:ID']['_text'] = delivery_partner.global_location_number
+
+        return node
+
+    def _ubl_add_invoice_delivery_nodes(self, vals):
```
clbr-odoo, line 1316: What's the added value to have two methods for the delivery ? 🤔 The side function seems to fill the whole <Delivery>.
`_ubl_add_invoice_delivery_nodes` + `_ubl_get_delivery_node_from_delivery_address`

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_a_nz.py
```diff
@@ -131,29 +131,57 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
-            party_node['cac:PartyTaxScheme'][0]['cbc:CompanyID']['_text'] = commercial_partner.company_registry
-            party_node['cac:PartyLegalEntity']['cbc:CompanyID'] = {
-                '_text': commercial_partner.company_registry,
-                'schemeID': '0088',
-            }
-
-        party_node['cac:PartyTaxScheme'][0]['cac:TaxScheme']['cbc:ID']['_text'] = 'GST'
-
-        return party_node
+            vals['party_node']['cbc:EndpointID']['_text'] = vat
```
clbr-odoo, line 140: looks incorrect, should be peppol eas/endpoint, which can or not be the VAT/company registry
but tbh the BIS3 AUNZ is deprecated, so we will never authorize it on Peppol... so doesn't matter much I guess...

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_a_nz.py
```diff
@@ -131,29 +131,57 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+                    '_text': vals['party_node']['cbc:EndpointID']['_text'],
+                    'schemeID': '0151',
+                },
+            }]
+        elif commercial_partner.country_code == 'NZ' and commercial_partner.company_registry:
+            vals['party_node']['cac:PartyLegalEntity'] = [{
+                'cbc:RegistrationName': {'_text': commercial_partner.name},
+                'cbc:CompanyID': {
+                    '_text': vals['party_node']['cbc:EndpointID']['_text'],
+                    'schemeID': '0088',
```
clbr-odoo, line 183: ... does NZ uses the GLN as their company registry ? super weird to mix `company_registry` and `global_location_number` even more now that we actually have a field for it, but again, hard to fix in stable, and most likely nobody uses this...

addons/l10n_ro_edi/tests/test_xml_ubl_ro.py
```diff
@@ -208,12 +208,12 @@ def test_export_no_vat_and_no_company_registry_raises_error_new(self):
         self.test_export_no_vat_and_no_company_registry_raises_error()
 
     def test_export_constraints(self):
-        self.company_data['company'].company_registry = False
         for required_field in ('city', 'street', 'state_id', 'vat'):
             prev_val = self.company_data["company"][required_field]
             self.company_data["company"][required_field] = False
             invoice = self.create_move("out_invoice", send=False)
             with self.assertRaisesRegex(UserError, "required"):
+                self.company_data['company'].company_registry = False
```
clbr-odoo, line 206: why this diff, aren't you just resetting in loop the same val now ?

addons/l10n_sg_ubl_pint/models/account_edi_xml_pint_sg.py
```diff
@@ -164,10 +164,26 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+
+        if (
+            commercial_partner.country_code == 'SG'
+            and commercial_partner.vat
+            and commercial_partner.vat != '/'
+        ):
+            vals['party_node']['cac:PartyTaxScheme'] = [{
+                'cbc:CompanyID': {
+                    '_text': commercial_partner.vat,
+                    'schemeID': None,
```
clbr-odoo:
```suggestion
```
that does not even exists on the node

addons/l10n_jp_ubl_pint/models/account_edi_xml_pint_jp.py
```diff
@@ -172,13 +172,14 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+        commercial_partner = partner.commercial_partner_id
 
-    def _get_party_node(self, vals):
-        party_node = super()._get_party_node(vals)
         # optional, if set: scheme_id should be taken from ISO/IEC 6523 list
-        party_node['cac:PartyLegalEntity']['cbc:CompanyID'] = None
-        return party_node
+        if commercial_partner.country_code == 'JP':
+            for node in nodes:
+                node['cbc:CompanyID'] = None
```
clbr-odoo, line 188: Also looks incorrect, the company ID can be set, but I just don't know if a `vat` content would make sense for JP. Looks like they have a Corporate Number, so probably more a company_registry for us.
https://docs.peppol.eu/poac/jp/pint-jp/trn-invoice/syntax/cac-AccountingSupplierParty/cac-Party/cac-PartyLegalEntity/
https://docs.peppol.eu/poac/jp/pint-jp/trn-invoice/syntax/cac-AccountingCustomerParty/cac-Party/cac-PartyLegalEntity/

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_nlcius.py
```diff
@@ -130,6 +130,38 @@ def _ubl_get_line_allowance_charge_discount_node(self, vals, discount_values):
...
+        commercial_partner = partner.commercial_partner_id
+
+        if (
+            commercial_partner.country_code == 'NL'
+            and commercial_partner.peppol_endpoint
+        ):
+            vals['party_node']['cac:PartyIdentification'] = [{
+                'cbc:ID': {
+                    '_text': commercial_partner.peppol_endpoint,
+                    'schemeID': None,
```
clbr-odoo: I guess we could put the EAS if its `in ('0106', '0190')`

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_xrechnung.py
```diff
@@ -79,16 +79,53 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+        if (
+            not nodes
+            and commercial_partner.peppol_eas
+        ):
+            nodes.append({
+                'cbc:CompanyID': {'_text': None},
+                'cac:TaxScheme': {
+                    'cbc:ID': {'_text': commercial_partner.peppol_eas},
+                },
+            })
```
clbr-odoo, line 116: Where does that comes from ? Same below PartyLegalEntity ?

addons/account_edi_ubl_cii/tests/test_ubl_export_bis3_be.py
```diff
@@ -476,16 +476,171 @@ def test_invoice_negative_discount_upsell(self):
         with the appropriate UNCL 7161 reason code (ADK) instead of an Allowance.
         """
         tax_21 = self.percent_tax(21.0)
-        product = self._create_product(lst_price=10.0, taxes_id=tax_21)
+        product = self._create_product(lst_price=5.76, taxes_id=tax_21)
```
clbr-odoo, line 564: ok, but why ? 😅

addons/account_edi_ubl_cii/tests/test_ubl_export_bis3_be.py
```diff
@@ -476,16 +476,171 @@ def test_invoice_negative_discount_upsell(self):
...
             discount=-1.09,
             partner_id=self.partner_be,
             post=True,
         )
 
         self._generate_invoice_ubl_file(invoice)
-
         self._assert_invoice_ubl_file(invoice, 'test_invoice_negative_discount_upsell')
+
+    def _test_invoice_partner_party_identifiers(self, partner, test_file):
```
clbr-odoo:
```suggestion
    def _assert_invoice_partner_party_identifiers(self, partner, test_file):
```
even with the private `_` prefix it's imo too close to regular test method

addons/account_edi_ubl_cii/tests/test_ubl_export_bis3_be.py
clbr-odoo, line 1: Probably some things to fix in NL setup, see my comments on the test files.

addons/l10n_anz_ubl_pint/models/account_edi_xml_pint_anz.py
```diff
@@ -223,18 +223,47 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
-        party_node['cac:PartyTaxScheme'][0]['cac:TaxScheme']['cbc:ID']['_text'] = 'GST'
+        if (
+            commercial_partner.country_code in ('AU', 'NZ')
+            and commercial_partner.vat
+            and commercial_partner.vat != '/'
+        ):
+            vals['party_node']['cac:PartyTaxScheme'] = [{
+                'cbc:CompanyID': {'_text': commercial_partner.vat},
+                'cac:TaxScheme': {
+                    'cbc:ID': {'_text': 'GST'},
```
clbr-odoo: If I understand correctly, this should actually be "TAX" in PINTs.
https://docs.peppol.eu/poac/aunz/pint-aunz/trn-invoice/syntax/cac-AccountingSupplierParty/cac-Party/cac-PartyTaxScheme-2/cac-TaxScheme/

addons/l10n_anz_ubl_pint/models/account_edi_xml_pint_anz.py
```diff
@@ -223,18 +223,47 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
         if commercial_partner.country_code in ('AU', 'NZ'):
-            party_node['cac:PartyLegalEntity']['cbc:CompanyID']['schemeID'] = commercial_partner.peppol_eas
-
-        return party_node
+            if commercial_partner.vat and commercial_partner.vat != '/':
+                vals['party_node']['cac:PartyLegalEntity'] = [{
+                    'cbc:RegistrationName': {'_text': commercial_partner.name},
+                    'cbc:CompanyID': {
+                        '_text': commercial_partner.vat,
+                        'schemeID': '0151' if commercial_partner.country_code == 'AU' else '0088',
```
clbr-odoo, line 244: same comment as in the bis3 version, that GLN usage is weirdo

addons/l10n_fr_facturx_chorus_pro/models/account_edi_xml_ubl_bis3.py
```diff
@@ -94,34 +94,43 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
+        # EXTENDS account.edi.ubl_bis3
+        super()._ubl_add_party_legal_entity_nodes(vals)
+        partner = vals['party_vals']['partner']
+        commercial_partner = partner.commercial_partner_id
+
+        if (
+            commercial_partner.country_code == 'FR'
+            and self.module_installed('l10n_fr')
+            and commercial_partner.siret
+            and self._is_customer_behind_chorus_pro(commercial_partner)
```
clbr-odoo: you included this `_is_customer_behind_chorus_pro` in the big if here but not in the one above, I don't care which solution, but be consistent

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -227,23 +218,34 @@ def _add_invoice_line_item_nodes(self, line_node, vals):
...
+                    },
+                }]
         elif commercial_partner.company_registry:
-            party_node['cac:PartyIdentification']['cbc:ID'].update({
-                '_text': '0088:' + commercial_partner.company_registry
-            })
-        return party_node
+            vals['party_node']['cac:PartyIdentification'] = [{
+                'cbc:ID': {
+                    '_text': f'0088:{commercial_partner.company_registry}',
```
clbr-odoo, line 269: Why do the guys have with that poor GLN.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+            node['cbc:ActualDeliveryDate']['_text'] = invoice.invoice_date
+        return node
+
+    def _add_invoice_delivery_nodes(self, document_node, vals):
+        # OVERRIDE
+        sub_vals = {
+            **vals,
+            'document_node': document_node,
+        }
+        self._ubl_add_invoice_delivery_nodes(sub_vals)
```
clbr-odoo, line 1155:
```suggestion
        super()._ubl_add_invoice_delivery_nodes(sub_vals)
```

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -585,145 +585,11 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
-            document_node['cac:Delivery']['cac:DeliveryLocation'].update({
-                'cbc:ID': {'schemeID': '0088', '_text': gln},
-            })
-
     def _add_invoice_payment_means_nodes(self, document_node, vals):
         super()._add_invoice_payment_means_nodes(document_node, vals)
         document_node['cac:PaymentMeans']['cbc:PaymentDueDate'] = None
         document_node['cac:PaymentMeans']['cbc:InstructionID'] = None
 
-    def _get_address_node(self, vals):
```
clbr-odoo, line 632: Just wondering if we should add an explicit comment on these methods in ubl20 like "will be deprecated in further version, please don't use it in new implems".

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+            vals['party_node']['cbc:EndpointID']['schemeID'] = commercial_partner.peppol_eas
+
+    def _ubl_add_party_identification_nodes(self, vals):
+        # EXTENDS account.edi.ubl
+        super()._ubl_add_party_identification_nodes(vals)
+        nodes = vals['party_node']['cac:PartyIdentification']
+        partner = vals['party_vals']['partner']
+        commercial_partner = partner.commercial_partner_id
+
+        if commercial_partner.country_code == 'BE' and commercial_partner.company_registry:
```
clbr-odoo, line 980: As discussed on the test files, we could imagine a belgian helper that retrieves both vat and company_registry as long as one of the two (three if we take the peppol eas/endpoint into consideration) is defined.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+                    'cbc:ID': {'_text': tax_scheme_id},
+                },
+            })
+        elif commercial_partner.peppol_endpoint and commercial_partner.peppol_eas:
+            # TaxScheme based on partner's EAS/Endpoint.
+            nodes.append({
+                'cbc:CompanyID': {'_text': commercial_partner.peppol_endpoint},
+                'cac:TaxScheme': {
+                    'cbc:ID': {'_text': commercial_partner.peppol_eas},
+                },
```
clbr-odoo, line 1021: Ok, I understand better the weird behavior with the EAS in the taxscheme, but I think it's not okay and misleading, even if in theory it's != "VAT" so it should pass the schematron. I'd rather put like NOT_VAT or whatever.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+            nl_id = commercial_partner.peppol_endpoint
+        else:
+            nl_id = commercial_partner.company_registry
+
+        if commercial_partner.country_code == 'NL' and nl_id:
+            # For NL, VAT can be used as a Peppol endpoint, but KVK/OIN has to be used as PartyLegalEntity/CompanyID
+            # To implement a workaround on stable, company_registry field is used without recording whether
+            # the number is a KVK or OIN, and the length of the number (8 = KVK, 20 = OIN) is used to determine the type
+            nodes.append({
+                'cbc:RegistrationName': {'_text': commercial_partner.name},
```
clbr-odoo, line 1041: the RegistrationName seems to be always the same for all countries listed here, so you can extract it outside of the if...else.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+        super()._ubl_add_accounting_supplier_party_tax_scheme_nodes(vals)
+        nodes = vals['party_node']['cac:PartyTaxScheme']
+        partner = vals['party_vals']['partner']
+        commercial_partner = partner.commercial_partner_id
+
+        if commercial_partner.country_code == 'NO':
+            nodes.append({
+                'cbc:CompanyID': {'_text': "Foretaksregisteret"},
+                'cac:TaxScheme': {
+                    'cbc:ID': {'_text': "TAX"},
```
clbr-odoo, line 1110: meh, suspect no ? I doubt this belongs to BIS3, or that it's correct
The only NO specific rule I see in the BIS3 is this: https://docs.peppol.eu/poacc/billing/3.0/rules/ubl-peppol/PEPPOL-COMMON-R041/

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
```diff
@@ -1048,6 +929,217 @@ def _add_invoice_line_price_nodes(self, line_node, vals):
...
+        invoice = vals['invoice']
+
+        if invoice.delivery_date:
+            node['cbc:ActualDeliveryDate']['_text'] = invoice.delivery_date
+
+        # Intracom delivery inside European area.
+        customer = vals['customer']
+        supplier = vals['supplier']
+        # TODO: missing countries in economic_area + what if the country of partner is not set but on commercial_partner
+        # TODO add a dedicated test for that part
```
clbr-odoo: should we fix that in this PR ?

clbr-odoo (changes requested):
I'm wondering if it wouldn't be nice to add a helper that takes as input a `res.partner` and a `country_code` to ouput a dict such as: 
```
{
"0208": "....",
"9925": "...",
"0088": "...",
...
}
```
That helper would search on all "known" places in stable: peppol eas/endpoint, company_registry, vat, ...

Add a test for this:
https://docs.peppol.eu/poacc/billing/3.0/rules/ubl-peppol/DE-R-015/

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
smetl: @clbr-odoo Currently there is a generic fallback to the EAS/Endpoint when there is nothing else to put here. I agree this is probably incorrect but I think this fallback in more for others country that have no country-specific override but a valid identifier for that part in EAS/Endpoint.
In Belgium, I think you could have a BCE but no VAT. In that case, Idk what should be put here

However, the fallback is not setting the SchemeID because it's not expecting the EAS but the ICD that is not always the same thing. So indeed, this part sucks and has to be improved

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
smetl: The whole test is about testing what is happening when there is only the EAS/Endpoint set on the partner

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_vat.xml
smetl: @clbr-odoo Yes. I plan to do that in part 2 with multi-ids

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_lu_company_registry.xml
smetl: Doc:

In Luxembourg the 9938 PEPPOL ID can be either

a VAT number (normally written with LU + 8 digits), or

the Luxembourg “matricule” / national identity number — an 11-digit company/organization registry number (Répertoire des Personnes Morales).

The 11-digit 00005000041 is not a VAT number — it’s a Luxembourg legal entity register number (matricule). That’s why it doesn’t have the LU prefix and doesn’t match the “LU + 8 digit VAT” format.

Then, 00005000041 corresponds to a specific administration (DIG) and then is not a VAT number but an organization registry. Consider this as an exception

addons/account_edi_ubl_cii/models/account_edi_ubl.py
smetl: `cac:Delivery` is 0...n and is a generic node in UBL standard.
For now, we compute only one based on the shipping address but there are multiple things that could be under this tag. That's why I have 2 methods:
`_ubl_add_invoice_delivery_nodes` adding *multiple nodes*
`_ubl_get_delivery_node_from_delivery_address` adding *the node specific to the shipping address*

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_a_nz.py
smetl: Yes that part sucks.... After multi-id`*`, I will remove all that code
`*` referring to a computed multi-id that will be backported in 18.0, not the stored json in master

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_a_nz.py
smetl: @clbr-odoo Yes I also bugged on that part..

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_vat_eas_oin_company_registry.xml
smetl: Indeed :+1:

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_vat_eas_kvk_company_registry.xml
smetl: Implemented only in NLCIUS => probably incorrect

addons/account_edi_ubl_cii/models/account_edi_ubl.py
smetl: According to BIS3, those values are not allowed and are removed by this override:
```py
    def _ubl_get_partner_address_node(self, vals, partner):
        # EXTENDS account.edi.ubl
        node = super()._ubl_get_partner_address_node(vals, partner)
        node['cbc:CountrySubentityCode'] = None
        node['cac:Country']['cbc:Name'] = None
        return node
```

addons/l10n_ro_edi/tests/test_xml_ubl_ro.py
smetl: I had a weird issue in this test. When we raise, `company_registry` is reset as it's default value. Idk why this was not raised before. Probably because I'm accessing to more fields than before triggering the recompute of the ORM

addons/l10n_jp_ubl_pint/models/account_edi_xml_pint_jp.py
smetl: To be fixed in part 2 ?

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_xrechnung.py
smetl: No idea 😆

addons/account_edi_ubl_cii/tests/test_ubl_export_bis3_be.py
smetl: @clbr-odoo This is useless to instanciate the product with a custom price_unit and put another one on the invoice. See diff below.

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
smetl: idk :cry:

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
smetl: This is not the same method ;)

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_xrechnung.py
smetl: Before my changes, those tags were a dict instead of a list. So the content was always there even if empty/incorrect/incomplete... I just mimic the old stupid behavior to keep the test working

addons/account_edi_ubl_cii/models/res_partner.py
```diff
@@ -7,6 +7,39 @@
 from odoo.addons.account_edi_ubl_cii.models.account_edi_common import EAS_MAPPING
 from odoo.addons.account.models.company import PEPPOL_DEFAULT_COUNTRIES
 
+GST_COUNTRY_CODES = {
```
clbr-odoo: EU looks correct, I'll trust you for the other 😅

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
clbr-odoo: If the output of the test is to generate incorrect file, we should fix it no ?

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_be_only_eas_endpoint.xml
clbr-odoo: I can leave with that for now.

addons/account_edi_ubl_cii/tests/test_files/export/bis3/invoice/be/test_invoice_customer_party_identifiers_partner_nl_only_vat.xml
clbr-odoo: Let's keep that for Multi ID I guess.

clbr-odoo (changes requested):
I resolved some, but there are still some pending comments.

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_bis3.py
clbr-odoo: NVM https://docs.peppol.eu/poacc/billing/3.0/rules/ubl-peppol/NO-R-002/

addons/l10n_my_ubl_pint/tests/expected_xmls/invoice_no_taxes_new.xml
```diff
@@ -33,7 +33,7 @@
       <cac:PartyTaxScheme>
         <cbc:CompanyID>C2584563200</cbc:CompanyID>
         <cac:TaxScheme>
-          <cbc:ID>GST</cbc:ID>
```
clbr-odoo, line 36: Forgot to revert this one.
https://docs.peppol.eu/poac/my/pint-my/bis/#_seller_accountingsupplierparty
Looks like it should also be VAT or GST...

clbr-odoo (changes requested):

clbr-odoo (approved):
@robodoo r+ delegate+ fw=skipmerge

clbr-odoo:
@robodoo r-

PR #246382, opened 2026-01-29
[FIX] account_edi_ubl_cii: Fix LineExtensionAmount computation [BR-CO…

smetl:
…-10]

According to [BR-CO-10], LineExtensionAmount should be: <quantity> * <price_unit_wo_tax> + charges - allowances

It was implemented as:
<quantity> * <price_unit_wo_tax> + charges - allowances + <delta_total_excluded>

<delta_total_excluded> is needed because it's the additional delta distributed by the global rounding of taxes accross the lines.
If you don't add it, you will change the untaxed and total amount of your document.

Instead, this commit adds 2 new values in the base_lines's tax_details: gross_total_excluded & discount_amount being the rounded versions of raw_gross_total_excluded & raw_discount but taking care of maintaining a global consistency regarding the global rounding.

---


clbr-odoo (approved):
Thanks !
@robodoo r+

PR #246411, opened 2026-01-29
[FIX] l10n_sa_edi: adapt CSR and invoice XML for VAT Group onboarding

hisi-odoo:
ZATCA requires VAT Group members (companies with 15-digit VAT where 11th digit = '1') to use their unique 10-digit TIN as `OrganizationUnitName` in CSR generation and `PartyIdentification` in invoice XML, instead of deriving it from `VAT[:10]`. Until now, we were still using the VAT-based value, which caused non-compliance during onboarding.

Changes made:
- Detects VAT Group members based on the VAT format
- Enforces TIN as the identification scheme for VAT Group members
- Validates that a proper 10-digit TIN is provided
- Uses the TIN for CSR `OrganizationUnitName` and invoice XML `PartyIdentification`
- Modified views to show/hide TIN field appropriately for VAT Group members

Existing behavior remains unchanged for regular (non-VAT-Group) companies, while ensuring VAT Group members fully comply with ZATCA requirements.

Also to clarify why the constraint is skipped during company creation:

Creating a `res.company` in Odoo is a bit special. Under the hood, Odoo first creates the delegated `res.partner` record and immediately runs partner-level constraints. At that moment, the company record does not exist yet and the VAT Group related fields (TIN scheme / number) are not fully set.

Because of this order:
(1) Partner is created first --> (2) Partner constraints are evaluated --> (3) Company record is created afterwards

Running the VAT Group validation on the partner at step (2) would always fail, even though the final company state is correct.

To avoid this false failure:
- The VAT Group validation is enforced on `res.company`, where the lifecycle is complete
- Partner-level validation is kept for standalone partner edits
- Partner constraints are skipped only during company create/write, to avoid validating incomplete delegated data

**Note: Test data and XML fixtures were updated to avoid triggering VAT Group validation where not intended.**

task-5886379

### DIFF
--- a/addons/l10n_sa_edi/models/res_company.py
+++ b/addons/l10n_sa_edi/models/res_company.py
@@ -1,6 +1,6 @@
 import re
-from odoo import models, fields, _
-from odoo.exceptions import UserError
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
@@ -43,6 +43,51 @@ class ResCompany(models.Model):
         related='partner_id.l10n_sa_additional_identification_scheme', readonly=False)
     l10n_sa_additional_identification_number = fields.Char(
         related='partner_id.l10n_sa_additional_identification_number', readonly=False)
+    l10n_sa_is_vat_group_member = fields.Boolean(
+        compute='_compute_l10n_sa_is_vat_group_member',
+    )
+
+    @api.depends('vat', 'country_id')
+    def _compute_l10n_sa_is_vat_group_member(self):
+        for company in self:
+            # VAT group if 15-digit VAT with 11th digit = '1'
+            vat = re.sub(r'\D', '', company.vat or '')
+            company.l10n_sa_is_vat_group_member = (
+                company.country_id.code == 'SA'
+                and len(vat) == 15
+                and vat[10] == '1'
+            )
+
+    @api.constrains('vat', 'l10n_sa_additional_identification_scheme', 'l10n_sa_additional_identification_number')
+    def _check_l10n_sa_vat_group_tin(self):
+        """
+        Validate that VAT Group members have proper TIN configuration
+        """
+        for company in self:
+            if not company.l10n_sa_is_vat_group_member:
+                continue
+            errors = []
+            if company.l10n_sa_additional_identification_scheme != 'TIN':
+                errors.append(_("Additional Identification Scheme as 'TIN'"))
+            tin = re.sub(r'\D', '', company.l10n_sa_additional_identification_number or '')
+            if len(tin) != 10:
+                errors.append(_("Additional Identification Number as the TIN Number(10 digits)"))
+            if errors:
+                raise ValidationError(_(
+                    "To comply with ZATCA VAT Group onboarding rules, please set %s.",
+                    " and ".join(errors)
+                ))
+
+    @api.onchange('vat', 'country_id')
+    def _onchange_l10n_sa_vat_group_scheme(self):
+        """Auto-set identification scheme to TIN for VAT Group members"""
+        if self.l10n_sa_is_vat_group_member:
+            self.l10n_sa_additional_identification_scheme = 'TIN'
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Skip VAT group validation during creation to avoid errors with incomplete partner data
+        return super(ResCompany, self.with_context(l10n_sa_skip_vat_group_validation=True)).create(vals_list)
 
     def write(self, vals):
         for company in self:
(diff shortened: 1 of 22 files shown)


khva-odoo (commented):
Hello,
I think we need to make changes to the code as we are not able to create the company and will get the validation error will be raised when the record is saved.

addons/l10n_sa_edi/models/res_company.py
```diff
@@ -1,10 +1,12 @@
 import re
-from odoo import models, fields, _
-from odoo.exceptions import UserError
```
khva-odoo, line 3: We do not encourage this type of diff in stable.

addons/l10n_sa_edi/models/res_company.py
```diff
@@ -43,6 +45,48 @@ def _l10n_sa_generate_private_key(self):
...
         related='partner_id.l10n_sa_additional_identification_number', readonly=False)
+    l10n_sa_is_vat_group_member = fields.Boolean(
+        compute='_compute_l10n_sa_is_vat_group_member',
+    )
+
+    @api.depends('vat', 'country_id')
+    def _compute_l10n_sa_is_vat_group_member(self):
+        for company in self:
+            # VAT group if 15-digit VAT with 11th digit = '1'
+            vat = re.sub(r'[^0-9]', '', company.vat or '')
```
khva-odoo:
```suggestion
            vat = re.sub(r'\D', '', company.vat or '')
```

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -24,6 +27,48 @@ class ResPartner(models.Model):
...
 
+    l10n_sa_is_vat_group_member = fields.Boolean(
+        compute='_compute_l10n_sa_is_vat_group_member'
+    )
+
+    @api.depends('vat', 'country_id')
+    def _compute_l10n_sa_is_vat_group_member(self):
+        for partner in self:
+            # VAT group if 15-digit VAT with 11th digit = '1'
+            vat = re.sub(r'[^0-9]', '', partner.vat or '')
```
khva-odoo: same here replace with `\D`. Also wherever applicable.

addons/l10n_sa_edi/models/res_company.py
```diff
@@ -43,6 +45,48 @@ def _l10n_sa_generate_private_key(self):
...
+            is_vat_group = company.country_id.code == 'SA' and len(vat) == 15 and vat[10] == '1'
+            company.l10n_sa_is_vat_group_member = is_vat_group
```
khva-odoo:
```suggestion
            company.l10n_sa_is_vat_group_member = (
                company.country_id.code == 'SA'
                and len(vat) == 15
                and vat[10] == '1'
            )
```
Please apply this to  compute method of `res.partner` also.

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -24,6 +27,48 @@ class ResPartner(models.Model):
...
+            is_vat_group = partner.country_id.code == 'SA' and len(vat) == 15 and vat[10] == '1'
+            partner.l10n_sa_is_vat_group_member = is_vat_group
```
khva-odoo:
```suggestion
             partner.l10n_sa_is_vat_group_member = (
                partner.country_id.code == 'SA'
                and len(vat) == 15
                and vat[10] == '1'
            )
```

addons/l10n_sa_edi/models/res_company.py
```diff
@@ -43,6 +45,48 @@ def _l10n_sa_generate_private_key(self):
...
+        """
+        Validate that VAT Group members have proper TIN configuration
+        """
+        for company in self:
+            if not company.l10n_sa_is_vat_group_member:
+                continue
+            if company.l10n_sa_additional_identification_scheme != 'TIN':
+                raise ValidationError(_(
+                    "To comply with ZATCA VAT Group onboarding rules, the Additional Identification Scheme "
+                    "must be set to 'TIN' for VAT Group members (VAT numbers with 11th digit = '1')."
```
khva-odoo:
```suggestion
                    "To comply with ZATCA VAT Group onboarding rules, please set "
                    "Additional Identification Scheme as 'TIN'."
```
Also change in `res_company`

addons/l10n_sa_edi/models/res_company.py
```diff
@@ -43,6 +45,48 @@ def _l10n_sa_generate_private_key(self):
...
+                ))
+            tin = re.sub(r'[^0-9]', '', company.l10n_sa_additional_identification_number or '')
+            if len(tin) != 10:
+                raise ValidationError(_(
+                    "To comply with ZATCA VAT Group onboarding rules, the Additional Identification Number (TIN) "
+                    "must be exactly 10 digits for VAT Group members."
```
khva-odoo:
```suggestion
                    "To comply with ZATCA VAT Group onboarding rules, please set "
                    "Additional Identification Number as the TIN Number(10 digits)."
```
Also change in `res_company`

PR #246434, opened 2026-01-30
[IMP] mail: show call permissions modal when microphone access is blocked

nmak-odoo:
**Purpose of this PR:**

Previously, if a user had blocked microphone access and attempted to record a voice message, the system only displayed a toast notification indicating that the site required microphone permissions.

This commit replaces the toast with the call permissions modal, which helps the user understand why recording is not working and how to enable microphone access.

task-5873815

---


maad-odoo (approved):

alexkuhn (approved):
@robodoo delegate+ r+

PR #246435, opened 2026-01-30
[IMP] pos*: adjust payment method sequence

rpka-odoo:
Purpose
-------
Validation can be done instantly when a manual payment method is listed first.
Having an online payment method as the first option breaks this flow by 
requiring a scan before validation.

In this commit
--------------
The payment method sequence is updated so that manual methods remain first,
restoring fast validation:

1) Cash
2) Card
3) Online payment
4) Customer account

Additionally:
- When a new payment method is added, it will automatically get the
  next sequence
- Correct domain when searching for the “Customer Account”
  payment method to avoid incorrect matching.

Task-5713304

### DIFF
--- a/addons/point_of_sale/models/pos_payment_method.py
+++ b/addons/point_of_sale/models/pos_payment_method.py
@@ -11,6 +11,9 @@ class PosPaymentMethod(models.Model):
     _order = "sequence, id"
     _inherit = ['pos.load.mixin']
 
+    def _default_sequence(self):
+        return (self.search([], order="sequence desc", limit=1).sequence or 0) + 1
+
     def _get_terminal_provider_selection(self):
         return []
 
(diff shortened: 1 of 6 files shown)


addons/point_of_sale/models/pos_payment_method.py
```diff
@@ -142,7 +142,11 @@ def _is_write_forbidden(self, fields):
 
     @api.model_create_multi
     def create(self, vals_list):
+        current_seq = self.search([], order='sequence desc', limit=1).sequence or 0
         for vals in vals_list:
+            if 'sequence' not in vals:
+                vals['sequence'] = current_seq + 1
+                current_seq += 1
```
jipr-odoo:
```suggestion
```
you can make a default sequence method for sequence field 
```py
sequence = fields.Integer(copy=False, default=_default_sequence)
```

addons/point_of_sale/models/pos_payment_method.py
```diff
@@ -142,7 +142,11 @@ def _is_write_forbidden(self, fields):
 
     @api.model_create_multi
     def create(self, vals_list):
+        current_seq = self.search([], order='sequence desc', limit=1).sequence or 0
```
jipr-odoo:
```suggestion
```

jipr-odoo (changes requested):

davidmonnom (approved):
@robodoo r+

PR #246353, opened 2026-01-29
[FIX] account: Adapt audit trail tests

abz-odoo:
Since this commit https://github.com/odoo/odoo/pull/242248/changes flushing inside `test_cant_unlink_message1`, `test_cant_unown_message` is now required.

https://runbot.odoo.com/odoo/runbot.build.error/237803
runbot/error-237803

PR #246436, opened 2026-01-30
[FIX][19.0] account: Ensure proper flushing of the database before audit trail checks

trungtuan88:
During the testing of the `account` module, I encountered errors in two tests: test_cant_unlink_message and test_cant_unown_message.

The cause is that the function def _search_account_audit_log_restricted uses _search, while the tests are still running within a transaction. As a result, when _search is executed, the database has not yet been updated with the latest values computed during the tests. Therefore, I added self.env.cr.flush() to flush the pending changes and update the database with the newest values.

---


davidtranhp:
This seemed to be the same as this https://github.com/odoo/odoo/pull/246353

PR #243967, opened 2026-01-15
[FIX] orm: filter 'pg_class' queries with current_schema

florentx:
Safety measure to prevent hidden bugs and data corruption which can happen when multiple schema are used in PostgreSQL.

Description of the issue/feature this PR addresses:
Related to:
- PR  #243833
- OPW [opw-5495025](https://www.odoo.com/my/tasks/5495025)
- issue reported earlier: #97891

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #243833, opened 2026-01-14
[FIX] orm: query current_schema instead of 'public'

florentx:
----
Description of the issue/feature this PR addresses:
See related OPW Ticket [opw-5495025](https://www.odoo.com/my/tasks/5495025)

Current behavior before PR:
Odoo seems to install correctly, and works normally.
However the database consistency is not ensured.  Foreign Keys are not created.
User receives `Record missing` errors after some time, when there was a Contact deleted, for example.

Issue is dangerous, because it can be latent and be undetected for weeks or months.

Desired behavior after PR is merged:

Odoo installs *with* the Foreign Keys, and works normally.

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

---
Ticket
opw-5495025

PR #246440, opened 2026-01-30
[FIX] orm: filter SQL queries with current_schema

florentx:
This is an addition to:
- [x] #243833
- [x] #243967

Current behavior before PR (use case):
- in database `odoodb` there's 2 PG schema `odoo_data` and `data_backup`
- schema `data_backup` contains tables from an old version of Odoo
- search_path is set to `odoo_data, "$user", public` (seen with `SHOW search_path;`)
- so the `data_backup` schema should be ignored

Then when you start Odoo, or install add-ons:
- you experience inconsistent behavior, and strange bugs
- this is due to SQL queries having access to the tables, and indexes in the other schema `data_backup`

- Odoo should ignore tables, indexes and objects which are not in the `current_schema`.

@kmagusiak 

---

### DIFF
--- a/odoo/cli/obfuscate.py
+++ b/odoo/cli/obfuscate.py
@@ -69,7 +69,7 @@ class Obfuscate(Command):
         return SQL("""CASE WHEN starts_with(%(field_name)s, 'odoo_cyph_') THEN pgp_sym_decrypt(decode(substring(%(field_name)s, 11)::text, 'base64'), %(pwd)s) ELSE %(field_name)s END""", field_name=sql_field, pwd=password)
 
     def check_field(self, table, field):
-        qry = "SELECT udt_name FROM information_schema.columns WHERE table_name=%s AND column_name=%s"
+        qry = "SELECT udt_name FROM information_schema.columns WHERE table_name=%s AND column_name=%s AND table_schema = current_schema"
         self.cr.execute(qry, [table, field])
         if self.cr.rowcount == 1:
             res = self.cr.fetchone()
@@ -81,7 +81,10 @@ class Obfuscate(Command):
         return False
 
     def get_all_fields(self):
-        qry = "SELECT table_name, column_name FROM information_schema.columns WHERE table_schema='public' AND udt_name IN ['text', 'varchar', 'jsonb'] AND NOT table_name LIKE 'ir_%' ORDER BY 1,2"
+        qry = (
+            "SELECT table_name, column_name FROM information_schema.columns"
+            " WHERE table_schema = current_schema AND udt_name IN ('text', 'varchar', 'jsonb') AND NOT table_name LIKE 'ir_%' ORDER BY 1,2"
+        )
         self.cr.execute(qry)
         return self.cr.fetchall()
 
--- a/odoo/tools/sql.py
+++ b/odoo/tools/sql.py
@@ -206,10 +206,9 @@ def existing_tables(cr, tablenames):
     cr.execute(SQL("""
         SELECT c.relname
           FROM pg_class c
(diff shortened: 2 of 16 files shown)


florentx:
Explanation.
This is a result of extensive search in the code base.

Here the typical changes which were applied:

1. Use helper from `odoo.tools.sql` whenever it's possible, to filter on the schema, like `index_exists(...)`,  `table_exists(...)` or `table_columns(...)`.

2. Then case of `pg_catalog` tables: `pg_class`, `pg_proc`, ...
```diff
 SELECT cl.relname FROM pg_class cl
  WHERE cl.relname = %s
+   AND cl.relnamespace = current_schema::regnamespace
```

3. Case of `pg_catalog` views: `pg_indexes`, ...
```diff
 SELECT indexname FROM pg_indexes
  WHERE indexname = %s AND tablename = %s
+   AND schemaname = current_schema
```

4. Case of `information_schema` views: `columns`, `tables`, ...
```diff
 SELECT column_name
   FROM information_schema.columns
  WHERE table_name = %s
+   AND table_schema = current_schema
```

It should cover all cases.

odoo/tools/sql.py
```diff
@@ -407,12 +407,11 @@ def get_depending_views(cr, table, column):
         JOIN pg_class as dependent ON pg_depend.refobjid = dependent.oid
         JOIN pg_attribute ON pg_depend.refobjid = pg_attribute.attrelid
             AND pg_depend.refobjsubid = pg_attribute.attnum
-        JOIN pg_namespace n ON dependee.relnamespace = n.oid
         WHERE dependent.relname = %s
+        AND dependent.relnamespace = current_schema::regnamespace
```
kmagusiak, line 411:
```suggestion
        AND dependee.relnamespace = current_schema::regnamespace
```

florentx: Both are ok.  since the check on previous line is `WHERE dependent.relname = %s`, it makes sense to query namespace on the same table `dependent`.
But if we query on `dependee`, result will be the same, since there is no dependencies across different schemas.

odoo/cli/obfuscate.py
```diff
@@ -81,7 +81,7 @@ def check_field(self, table, field):
         return False
 
     def get_all_fields(self):
-        qry = "SELECT table_name, column_name FROM information_schema.columns WHERE table_schema='public' AND udt_name IN ['text', 'varchar', 'jsonb'] AND NOT table_name LIKE 'ir_%' ORDER BY 1,2"
+        qry = "SELECT table_name, column_name FROM information_schema.columns WHERE table_schema = current_schema AND udt_name IN ['text', 'varchar', 'jsonb'] AND NOT table_name LIKE 'ir_%' ORDER BY 1,2"
```
Megaaaaaa: I know it's not part of your change but this is staring at me so bad I cannot not ask to make sure... Has this ever worked ?
```suggestion
        qry = "SELECT table_name, column_name FROM information_schema.columns WHERE table_schema = current_schema AND udt_name IN ('text', 'varchar', 'jsonb') AND NOT table_name LIKE 'ir_%' ORDER BY 1,2"
```

Megaaaaaa (commented):
Hey! Accounting part looks good, I just have something that tickles me in `[odoo/cli/obfuscate.py` 🤔

odoo/cli/obfuscate.py
kmagusiak: It probably was not.

florentx: I've fixed this syntax. Thanks.

kmagusiak (approved):
@robodoo r+

PR #246442, opened 2026-01-30
[FIX] mail: prevent traceback on free activities

mrsr-odoo:
Steps to reproduce
=================
1. Go to “View all activities”.
2. Switch to calendar view.
3. Click on an activity not linked to any model
(e.g. “Eat cookies”, “Send Email to Alfred”).
4. Click “View” in the popover.
=> Traceback

Reason
======
The commit [1] allow activities without a linked model and from commit [2]
such activities can be opened in the activity form view.

In the calendar view, the action is retrieved from the model and executed
using `doAction`. Since the action does not define `views`, and the `doAction`
depends on `action['views']`, an error occurs when the action service attempts
to copy it.

After this commit
==================
This commit fixes the issue by modifying the action returned from `action_open_document`
for non-linked models to include `views`, similar structure used when a model is present.

[1] https://github.com/odoo/odoo/commit/165b060473be8a5d33d62d311f0dc55ed6332d69
[2] https://github.com/odoo/odoo/commit/abeac135b9bb7aec4bcddd84fb0705743297f80d

Task-5857887

### DIFF
--- a/addons/mail/static/src/views/web/calendar/calendar_year/activity_calendar_year_popover.js
+++ b/addons/mail/static/src/views/web/calendar/calendar_year/activity_calendar_year_popover.js
@@ -12,6 +12,7 @@ export class ActivityCalendarYearPopover extends CalendarYearPopover {
         const action = await this.orm.call("mail.activity", "action_open_document", [
             record.rawRecord.id,
         ]);
-        this.actionService.doAction(action);
+        this.actionService.doAction(action, { onClose: () => this.props.model.load() });
+        this.props.close();
     }
 }
--- a/addons/mail/models/mail_activity.py
+++ b/addons/mail/models/mail_activity.py
@@ -609,12 +609,14 @@ class MailActivity(models.Model):
          access to the related record."""
         self.ensure_one()
         if not self.res_model:
+            view_id = self.env.ref('mail.mail_activity_view_form_popup').id
             return {
                 'res_id': self.id,
                 'type': 'ir.actions.act_window',
                 'view_mode': 'form',
                 'res_model': 'mail.activity',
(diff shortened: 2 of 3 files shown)


addons/mail/static/src/views/web/calendar/calendar_year/activity_calendar_year_popover.js
```diff
@@ -8,10 +8,14 @@ export class ActivityCalendarYearPopover extends CalendarYearPopover {
         this.actionService = useService("action");
     }
 
-    async onRecordClick(record) {
-        const action = await this.orm.call("mail.activity", "action_open_document", [
-            record.rawRecord.id,
-        ]);
-        this.actionService.doAction(action);
+    onRecordClick(record) {
```
tde-banana-odoo: Isn't it missing the sync / await statuses ?

mrsr-odoo: I guess its fine here
we don't need to wait for `doActionButton` to complete

awa-odoo:
Hello @mrsr-odoo 

Took me a little while to dig into the actual issue but I think I found it.

The difference between the previous/new call is that the previous one calls `/web/dataset/call_kw` that returns the plain result and the new one calls `/web/dataset/call_button` that in turn calls "clean_action", which contains the magic that inserts a "views" key into the result.

Instead of relying on that side-effect, isn't it better to make "action_open_document" return a valid action that does not need further cleaning?
Seems like adding the "views" key manually would also fix the issue: `'views': [self.env.ref('mail.mail_activity_view_form_popup').id, 'form']`.

And it would look similar to the structure returned when we have a model so it makes more sense to me, the method should return something you can always work with as opposed to "sometimes a valid action, sometimes not".

What do you think?

PR #246443, opened 2026-01-30
[FIX] cloud_storage, base: allow explicit content fetching

ushyme:
Some features hand a cloud storage download URL to an external service that may fetch it later. The default five-minute lifetime is too short for those flows.

### Steps to reproduce

1. Configure a cloud storage provider (e.g. cloud_storage_google).
2. Upload a large file from the web client, so it is stored in the cloud.
3. Generate a download URL for a consumer that may fetch it after five minutes.
4. The URL expires before the consumer fetches it.

### Cause

The Google and Azure providers always use the default download URL lifetime, so callers cannot request a longer-lived URL.

### Fix

Read an optional cloud_storage_download_url_time_to_expiry context value when generating a download URL. Keep the existing five-minute lifetime as the default for all current callers.

opw-5424132

Related Enterprise PR: odoo/enterprise#105967

### DIFF
--- a/addons/cloud_storage/models/ir_attachment.py
+++ b/addons/cloud_storage/models/ir_attachment.py
@@ -84,6 +84,12 @@ class CloudStorageAttachment(models.Model):
         """
         raise NotImplementedError()
 
+    def _get_cloud_storage_download_url_time_to_expiry(self):
+        return self.env.context.get(
+            'cloud_storage_download_url_time_to_expiry',
+            self._cloud_storage_download_url_time_to_expiry,
+        )
+
     def _generate_cloud_storage_download_info(self):
         """
         Generate the download info for the public client to directly download
--- a/addons/cloud_storage_google/models/ir_attachment.py
+++ b/addons/cloud_storage_google/models/ir_attachment.py
@@ -68,15 +68,16 @@ class IrAttachment(models.Model):
         if self.env['ir.config_parameter'].sudo().get_param('cloud_storage_provider') != 'google':
             return super()._generate_cloud_storage_download_info()
         info = self._get_cloud_storage_google_info()
+        time_to_expiry = self._get_cloud_storage_download_url_time_to_expiry()
         return {
             'url': self._generate_cloud_storage_google_signed_url(
                 info['bucket_name'], info['blob_name'],
                 method='GET',
-                expiration=self._cloud_storage_download_url_time_to_expiry,
+                expiration=time_to_expiry,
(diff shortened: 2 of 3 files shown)


addons/cloud_storage/models/ir_attachment.py
```diff
@@ -21,6 +21,12 @@ class CloudStorageAttachment(models.Model):
         ondelete={'cloud_storage': 'set url'}
     )
 
+    def _get_public_url(self):
+        self.ensure_one()
+        if self.type == 'cloud_storage':
+            return self._generate_cloud_storage_download_info().get('url')
```
HydrionBurst: Note: the url with access token will only be valid for 5 minutes. It cannot be cached.

Also, usually you don't need it.
You can always use the odoo url.
`web/content/id` (with the access token if the ir.attachment is not public)

when the user accesses the url
odoo will
1. check the access right
2. redirect user to the cloud storage url with cloud storage access token for 5 minutes

HydrionBurst:
@ushyme 
For the problem of whatsapp, does the whatsapp channel have an identifier in js code? If yes I propose
for stable versions we explicitly add a check in the [js code](https://github.com/odoo/odoo/blob/f5da5197620ad12302b40ff1702a9cda5007aaa3/addons/cloud_storage/static/src/core/common/attachment_upload_service_patch.js#L90) to prevent those attachments to be uploaded to the cloud storage.
And in master we can use ir.config_parameter to make the check configurable.

addons/cloud_storage_google/models/ir_attachment.py
```diff
@@ -68,15 +68,16 @@ def _generate_cloud_storage_download_info(self):
         if self.env['ir.config_parameter'].sudo().get_param('cloud_storage_provider') != 'google':
             return super()._generate_cloud_storage_download_info()
         info = self._get_cloud_storage_google_info()
+        time_to_expiry = self.env.context.get('cloud_storage_download_url_time_to_expiry', self._cloud_storage_download_url_time_to_expiry)
```
HydrionBurst: Could you create a method
`self._get_cloud_storage_download_url_time_to_expiry()`
in the `cloud_storage` and then we call that method to have consistent behavior everywhere

ushyme: Thanks, done !

HydrionBurst:
LGTM 

@robodoo delegate+

ushyme:
@robodoo r+

PR #246445, opened 2026-01-30
[FIX] test_http: disable device check by default during testing

thle-odoo:
In tests, when using `authenticate`, we create a session. When this session is retrieved (for example because we use `url_open`), we detect a new device and insert a log.
The consequence is that a query is performed in many tests and that is not necessary.
The fix consists of disabling the `res.device.log` feature by default in tests.

task-5894825


Julien00859 (approved):
@robodoo r+

PR #246446, opened 2026-01-30
[FIX] point_of_sale: use lna for printer test button

parp-odoo:
Ensure the printer and preparation printer Test buttons use LNA when triggered,
aligning behavior with `point_of_sale.use_lna` configuration.

Task-5886700

Related: https://github.com/odoo/enterprise/pull/106594

### DIFF
--- a/addons/point_of_sale/static/src/backend/test_epos/test_epos.js
+++ b/addons/point_of_sale/static/src/backend/test_epos/test_epos.js
@@ -1,8 +1,9 @@
-import { Component } from "@odoo/owl";
+import { Component, onWillStart } from "@odoo/owl";
 import { registry } from "@web/core/registry";
 import { standardWidgetProps } from "@web/views/widgets/standard_widget_props";
 import { _t } from "@web/core/l10n/translation";
 import { useService } from "@web/core/utils/hooks";
+import { initLNA } from "@point_of_sale/app/utils/init_lna";
 
 const EPSON_ERRORS = {
     DeviceNotFound: _t(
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/static/src/backend/test_epos/test_epos.js
```diff
@@ -26,9 +27,13 @@ export class TestEPos extends Component {
...
     };
 
-    setup() {
+    async setup() {
         super.setup();
         this.notification = useService("notification");
+        this.orm = useService("orm");
+        odoo.use_lna = Boolean(
+            await this.orm.call("ir.config_parameter", "get_param", ["point_of_sale.use_lna"])
+        );
```
davidmonnom: For this you can create a little method in `pos_printer.py` that will return the value of the parameter, otherwise you get an access error if the user isn't administrator

Image

```py
@api.model
def use_local_network_access(self):
      use_lna = <get params in sudo>
      return { 'use_lna': use_lna }
```

davidmonnom (changes requested):

addons/point_of_sale/static/src/backend/test_epos/test_epos.js
parp-odoo: Updated 🙂

addons/point_of_sale/static/src/backend/test_epos/test_epos.js
```diff
@@ -26,9 +27,11 @@ export class TestEPos extends Component {
         ...standardWidgetProps,
     };
 
-    setup() {
+    async setup() {
         super.setup();
         this.notification = useService("notification");
+        this.orm = useService("orm");
+        odoo.use_lna = (await this.orm.call("pos.printer", "use_local_network_access")).use_lna;
```
davidmonnom: onWillStart(async () => {

}

setup in component cannot be async, its better to use onWillStart instead

parp-odoo: Updated

davidmonnom (approved):
@robodoo r+

PR #246447, opened 2026-01-30
[IMP] hr_payroll,resource: Sync the work time rate

usseif97:
Current behavior before PR:
. The work time rate is currently computed separately in two different modules: hr_payroll (version model) and resource (resource_calendar model). Although they are calculated independently, both values represent the same concept and are functionally equivalent. This duplication can lead to inconsistencies and unnecessary complexity in maintenance

`resource/resource_calendar` model -> work_time_rate is not stored computed field with range=[0-100]
`hr_payroll/hr_version` model -> work_time_rate is stored computed field with range=[0-1]

. Synchronize the work time rate fields between the hr_payroll and resource modules to ensure they are related consistent
`resource/resource_calendar` model -> Modify work_time_rate range to be [0-1] and keep the same computation   
->  calendar.hours_per_week / calendar.full_time_required_hours

`hr_payroll/hr_version` model -> work_time_rate is `version.resource_calendar_id.work_time_rate` incase of Non-Flexible, 
While incase of Flexible versions it's 
->  version.hours_per_week / version.company_id.resource_calendar_id.hours_per_week

task-5429287

---

### DIFF
--- a/addons/resource/models/resource_calendar.py
+++ b/addons/resource/models/resource_calendar.py
@@ -80,11 +80,11 @@ class ResourceCalendar(models.Model):
     hours_per_week = fields.Float(
         string="Hours per Week",
         compute="_compute_hours_per_week", store=True, readonly=False, copy=False)
-    is_fulltime = fields.Boolean(compute='_compute_work_time_rate', string="Is Full Time")
+    is_fulltime = fields.Boolean(compute='_compute_is_fulltime', string="Is Full Time")
     two_weeks_calendar = fields.Boolean(string="Calendar in 2 weeks mode")
     two_weeks_explanation = fields.Char('Explanation', compute="_compute_two_weeks_explanation")
     work_resources_count = fields.Integer("Work Resources count", compute='_compute_work_resources_count')
-    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', search='_search_work_time_rate',
+    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', store=True,
         help='Work time rate versus full time working schedule, should be between 0 and 100 %.')
 
     # --------------------------------------------------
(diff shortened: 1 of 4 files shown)


kmagusiak (approved):

addons/resource/models/resource_calendar.py
```diff
@@ -77,7 +77,7 @@ def default_get(self, fields):
     is_fulltime = fields.Boolean(compute='_compute_work_time_rate', string="Is Full Time")
     work_resources_count = fields.Integer("Work Resources count", compute='_compute_work_resources_count')
     work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', search='_search_work_time_rate',
-        help='Work time rate versus full time working schedule, should be between 0 and 100 %.')
+        help='Work time rate versus full time working schedule, should be between 0 and 1')
```
romaincarlier4: Maybe we should store this field

addons/resource/models/resource_calendar.py
```diff
@@ -74,10 +74,10 @@ def default_get(self, fields):
     hours_per_week = fields.Float(
         string="Hours per Week",
         compute="_compute_hours_per_week", store=True, readonly=False, copy=False)
-    is_fulltime = fields.Boolean(compute='_compute_work_time_rate', string="Is Full Time")
+    is_fulltime = fields.Boolean(compute='_compute_is_fulltime', string="Is Full Time")
     work_resources_count = fields.Integer("Work Resources count", compute='_compute_work_resources_count')
-    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', search='_search_work_time_rate',
-        help='Work time rate versus full time working schedule, should be between 0 and 100 %.')
+    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', store=True, search='_search_work_time_rate',
+        help='Work time rate versus full time working schedule, should be between 0 and 1')
```
romaincarlier4: As you make this stored, you can remove the search method

usseif97: yes, exactly
I forgot this

addons/resource/models/resource_calendar.py
```diff
@@ -74,10 +74,10 @@ def default_get(self, fields):
     hours_per_week = fields.Float(
         string="Hours per Week",
         compute="_compute_hours_per_week", store=True, readonly=False, copy=False)
-    is_fulltime = fields.Boolean(compute='_compute_work_time_rate', string="Is Full Time")
+    is_fulltime = fields.Boolean(compute='_compute_is_fulltime', string="Is Full Time")
     work_resources_count = fields.Integer("Work Resources count", compute='_compute_work_resources_count')
-    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', search='_search_work_time_rate',
```
romaincarlier4, line 87: can you also remove the method ? it's not useful anymore if you remove it from here

romaincarlier4 (changes requested):

addons/resource/models/resource_calendar.py
usseif97: I think '_compute_work_time_rate' is needed here, it can't be removed

romaincarlier4: I meant the `_search_work_time_rate` method :)

usseif97: No, I remove it

usseif97: work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', store=True,
        help='Work time rate versus full time working schedule, should be between 0 and 1')

usseif97: you mean the method itself, got it

romaincarlier4 (approved):
lgtm, thanks

addons/resource/models/resource_calendar.py
```diff
@@ -80,12 +80,12 @@ def default_get(self, fields):
...
         compute="_compute_hours_per_week", store=True, readonly=False, copy=False)
-    is_fulltime = fields.Boolean(compute='_compute_work_time_rate', string="Is Full Time")
+    is_fulltime = fields.Boolean(compute='_compute_is_fulltime', string="Is Full Time")
     two_weeks_calendar = fields.Boolean(string="Calendar in 2 weeks mode")
     two_weeks_explanation = fields.Char('Explanation', compute="_compute_two_weeks_explanation")
     work_resources_count = fields.Integer("Work Resources count", compute='_compute_work_resources_count')
-    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', search='_search_work_time_rate',
-        help='Work time rate versus full time working schedule, should be between 0 and 100 %.')
+    work_time_rate = fields.Float(string='Work Time Rate', compute='_compute_work_time_rate', store=True,
+        help='Work time rate versus full time working schedule, should be between 0 and 1')
```
Bertrand2: Don't change the description, the widget is making the user express a value between 0 and 100% in the UX.

Bertrand2 (commented):
One comment otherwise LGTM 🙂

addons/resource/models/resource_calendar.py
usseif97: I assumed the description should reflect the stored field not what appears on the UI.

Bertrand2: The `help` attribute is meant for the end user, so it should always reflect on the UX :)

Bertrand2 (approved):
Thanks for the changes!
robodoo r+

PR #246449, opened 2026-01-30
[IMP] web: list column widths: wider columns in mobile

aab-odoo:
The list view column widths logic aims, among other things, at avoiding to have an horizontal scrollbar. To achieve it, it shrinks columns up to a mininum width depending on the column field type.

However, on small screens, the scrollbar is in general unavoidable, and we often end up with all columns being displaying with their minimal width, which implies a lot of truncated texts.

This commit introduces a slight change in the logic for small screens only: the minimal width of a column when it has to be shrinked is now 80% of the viewport. That means that if the content of the column is smaller than that, the width will be such that the content exactly fits the column, and if the content is larger, it will be truncated, but it will take 80% of the viewport.

Task~5249329

Current behavior before PR:

---


adr-odoo (approved):
@robodoo r+

PR #246450, opened 2026-01-30
[IMP] point_of_sale: printer management

shag-odoo:
Following this commit:
====
- Removed default selection of printer from backend, instead given option in
frontend which solves multiple devices issue also.
- Moved iface_cashdrawer field from pos.config to pos.printer
- Popup that is been added in frontend will ask user to select default printer
option from the list of printers that are configured.
 - User can change the options from menu setting from Printers Section.

task-5494313
Enterprise Related PR : https://github.com/odoo/enterprise/pull/105972
Upgrade Related PR : https://github.com/odoo/upgrade/pull/9381


addons/point_of_sale/views/pos_printer_view.xml
```diff
@@ -45,13 +46,26 @@
         <field name="model">pos.printer</field>
         <field name="arch" type="xml">
             <list string="Printers">
+            <field name="sequence" widget="handle"/>
                 <field name="name"/>
                 <field name="use_type"/>
```
paso-odoo:
```suggestion
                <field name="sequence" widget="handle"/>
                <field name="name"/>
                <field name="use_type"/>
```

paso-odoo (changes requested):

addons/point_of_sale/views/pos_printer_view.xml
shag-odoo: Updated!

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +518,46 @@ export class PosStore extends WithLazyGetterTrap {
...
+        const receiptPrinters = this.ticketPrinter.receiptPrinters;
+        const printer_id = Number(
+            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
+        );
+        if (!force && (printer_id || !receiptPrinters.length)) {
+            return;
+        }
+        if (receiptPrinters.length == 1) {
+            const relPrinter = Array.from(receiptPrinters)[0];
+            localStorage.setItem(
```
jipr-odoo: If there is only one printer then why do wee need to set in `localstorage`

addons/point_of_sale/models/pos_printer.py
```diff
@@ -53,6 +53,8 @@ class PosPrinter(models.Model):
         ),
     )
     use_lna = fields.Boolean(string="Use Local Network Access")
+    sequence = fields.Integer(default=1)
```
jipr-odoo: If you set the default 1 for sequence, then for all printers, the sequence will be the same = 1

I would suggest using 
```py
def _default_sequence(self):
        return (self.search([], order="sequence desc", limit=1).sequence or 0) + 1
```

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -245,6 +248,28 @@ export class PosStore extends WithLazyGetterTrap {
...
+        // If a default printer is set.
+        const storedPrinterId = Number(
+            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
+        );
+        if (storedPrinterId) {
+            const relPrinter = Array.from(this.ticketPrinter.printers).find(
+                (d) => d.id === storedPrinterId
+            );
+            if (relPrinter) {
+                localStorage.setItem(
```
jipr-odoo: why do we need to set again in localstorage if you get 
from here 

```js
const storedPrinterId = Number(
            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
        ); 
```

addons/point_of_sale/models/pos_printer.py
shag-odoo: What is the issue if every printer has the same sequence, although sequence value is not going to visible to user.

jipr-odoo: The issue isn’t visibility. With all sequences set to 1, ordering by sequence becomes meaningless and effectively falls back to the default ordering (usually by id).
If sequence is meant to control ordering, it should be unique. Otherwise, we might reconsider having a default value (in that case default value is also not needed).

shag-odoo: in most of the place in codebase itself default=1 is given
for reference check `pos_note.py`, `pos_restaurant.py` etc

addons/point_of_sale/static/src/app/services/pos_store.js
shag-odoo: Updated!

addons/point_of_sale/static/src/app/services/pos_store.js
shag-odoo: Updated!

addons/point_of_sale/models/pos_printer.py
jipr-odoo: Thanks for checking the codebase. You might have missed models like `pos.payment.method` and `pos.category`, where the sequence field is defined without a default value.

ok if we go with the existing code that you implemented.

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
...
+    async selectPrinter({ force = false } = {}) {
+        const receiptPrinters = this.ticketPrinter.receiptPrinters;
+        const printer_id = Number(
+            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
+        );
+        if (!force && (printer_id || !receiptPrinters.length)) {
+            return;
+        }
+        if (receiptPrinters.length == 1) {
+            this.ticketPrinter.activePrinter = Array.from(receiptPrinters)[0];
```
jipr-odoo:
```suggestion
            this.ticketPrinter.activePrinter =receiptPrinters[0];
```

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
         this.openCashbox(_t("Cash in / out"));
         return makeAwaitable(this.dialog, CashMovePopup);
     }
+    async selectPrinter({ force = false } = {}) {
```
jipr-odoo:
```suggestion
    async selectPrinter(force = false) {
```

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
...
+        const receiptPrinters = this.ticketPrinter.receiptPrinters;
```
jipr-odoo:
```js
 get receiptPrinters() {
        return [...this.config.receipt_printer_ids].sort((a, b) => a.sequence - b.sequence);
    }
```
this already done here so we dont need to do `Array.from(receiptPrinters)` everytime for 
receiptPrinters

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
...
+        );
+        if (!force && (printer_id || !receiptPrinters.length)) {
+            return;
+        }
+        if (receiptPrinters.length == 1) {
+            this.ticketPrinter.activePrinter = Array.from(receiptPrinters)[0];
+            return;
+        }
+        const defaultPrinter = await makeAwaitable(this.dialog, SelectDefaultPrinterPopup, {
+            receipt_printers: Array.from(receiptPrinters),
```
jipr-odoo:
```suggestion
            receipt_printers:receiptPrinters,
```

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
...
+            selectedId: printer_id,
+        });
+        if (defaultPrinter) {
+            const relPrinter = Array.from(receiptPrinters).find(
```
jipr-odoo:
```suggestion
            const relPrinter = receiptPrinters.find(
```

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -493,8 +514,41 @@ export class PosStore extends WithLazyGetterTrap {
         this.openCashbox(_t("Cash in / out"));
         return makeAwaitable(this.dialog, CashMovePopup);
     }
+    async selectPrinter({ force = false } = {}) {
+        const receiptPrinters = this.ticketPrinter.receiptPrinters;
+        const printer_id = Number(
+            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
+        );
```
jipr-odoo:
```js
        const printer_id = Number(
            localStorage.getItem(DEFAULT_PRINTER_STORAGE_KEY + this.config.id)
        );
```
Can we make a separate method to get localstorage printer id 
It will be easier to just call the method instead of doing the same code every time 

wdyt ?

jipr-odoo (changes requested):

addons/point_of_sale/static/src/app/services/pos_store.js
shag-odoo: yes we can make separate method for it.

addons/point_of_sale/models/pos_printer.py
paso-odoo:
```suggestion
    sequence = fields.Integer(copy=False)
```
There is no meaning for adding a default value in the sequence field. Seems like dead code.
Note: If someone has already done this, we should not do the same again without a valid reason.

addons/pos_self_order/static/src/app/services/self_order_service.js
paso-odoo, line 1: remove `default_receipt_printer_id` from `_load_pos_self_data_fields` in pos_self_order module

addons/point_of_sale/views/pos_printer_view.xml
```diff
@@ -19,6 +19,7 @@
                             <field name="printer_ip" invisible="printer_type != 'epson_epos'" placeholder="10.100.50.87" required="printer_type == 'epson_epos'"/>
                             <field name="use_lna"/>
                             <field name="product_categories_ids" invisible="use_type != 'preparation'" widget="many2many_tags" required="use_type == 'preparation'"/>
+                            <field name="iface_cashdrawer" invisible="use_type != 'receipt'"/>
```
paso-odoo: need 1 onchange or compute to make sure 
If you go from `receipt` to `preparation`, the `iface_cashdrawer` should be unchecked if it is checked in `receipt`

paso-odoo (changes requested):

addons/point_of_sale/models/pos_printer.py
```diff
@@ -62,13 +64,19 @@ def copy_data(self, default=None):
                 vals['name'] = _("%s (copy)", printer.name)
         return vals_list
 
+    @api.onchange('use_type')
+    def _onchange_use_type(self):
+        """Disable iface_cashdrawer when printer type is preparation"""
+        if self.use_type == "preparation" :
```
paso-odoo:
```suggestion
        if self.use_type == "preparation" and self.iface_cashdrawer:
```
also for loop for safe side

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -246,6 +248,23 @@ export class PosStore extends WithLazyGetterTrap {
...
+                (d) => d.id === storedPrinterId
+            );
+            if (relPrinter) {
+                this.ticketPrinter.activePrinter = relPrinter;
+                return;
+            }
+        }
+        if (["ProductScreen", "FloorScreen"].includes(this.router.state.current)) {
+            this.selectPrinter();
+        }
```
paso-odoo: When we show the popup on the first print, do we also need to show it when opening the register?

Reason for removal: This code becomes dead once the printer is stored in local storage.
May be you can also check with PO

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
```diff
@@ -97,11 +98,10 @@ export class PosTicketPrinterService {
     }
 
     async openCashbox() {
-        if (!this.config.default_receipt_printer_id?._instance?.openCashbox) {
+        if (!this.activePrinter._instance?.openCashbox) {
```
paso-odoo:
```suggestion
        if (!this.activePrinter?._instance?.openCashbox) {
```

paso-odoo (changes requested):

addons/point_of_sale/models/pos_printer.py
```diff
@@ -53,6 +53,7 @@ class PosPrinter(models.Model):
         ),
     )
     use_lna = fields.Boolean(string="Use Local Network Access")
+    iface_cashdrawer = fields.Boolean(string='Link Cashdrawer', help="Automatically open the cashdrawer.")
```
davidmonnom:
```suggestion
    use_cashdrawer = fields.Boolean(string='Link Cashdrawer', help="Automatically open the cashdrawer.")
```

addons/point_of_sale/models/pos_printer.py
```diff
@@ -62,13 +63,20 @@ def copy_data(self, default=None):
                 vals['name'] = _("%s (copy)", printer.name)
         return vals_list
 
+    @api.onchange('use_type')
+    def _onchange_use_type(self):
+        """Disable iface_cashdrawer when printer type is preparation"""
+        for rec in self:
+            if rec.use_type == "preparation" and rec.iface_cashdrawer:
+                rec.iface_cashdrawer = False
```
davidmonnom: I don't think its necessary, `use_cashdrawer` setting can keep its value even if type is preparation. It will don't have any side effect.

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
```diff
@@ -60,6 +64,10 @@ export class PosTicketPrinterService {
         return this.preparationPrinters.length > 0;
     }
 
+    get printerStorageKey() {
+        return `pos_default_printer_id_${this.config.id}`;
```
davidmonnom: Also use `odoo.info.db` in case of multi db

davidmonnom (commented):
Missing some tests, should do a hoot that check when we change the printer, when there is no default one etc...

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
shag-odoo: Thanks for the info, Updated!

addons/point_of_sale/models/pos_printer.py
shag-odoo: Got it but when an user sets a receipt printer and cashdrawer is enabled and later on then changes its type to preparation. cashdrawer still has the value (no issue at all) but if again user changes its type to receipt then cashdrawer will have the value and it will be enabled by default

addons/point_of_sale/static/tests/unit/data/pos_printer.data.js
```diff
@@ -4,7 +4,15 @@ export class PosPrinter extends models.ServerModel {
...
     _load_pos_data_fields() {
-        return ["id", "name", "product_categories_ids", "printer_type"];
+        return [
+            "id",
+            "name",
+            "product_categories_ids",
+            "printer_type",
+            "use_type",
+            "printer_ip",
+            "iface_cashdrawer",
```
adgu-odoo:
```suggestion
            "use_cashdrawer",
```

adgu-odoo (commented):
Good job, just a little change if you don't mind and this part will be good

addons/point_of_sale/static/tests/unit/data/pos_printer.data.js
shag-odoo: @adgu-odoo Updated!

adgu-odoo:
robodoo r+

adgu-odoo:
robodoo r+

PR #246464, opened 2026-01-30
[FIX] product: print product barcodes correctly

qucol-odoo:
## Issue
When printing the barcode for a product, only a blank page would be printed.

## Steps to reproduce
1. Install Barcode (`stock_barcode`)
2. Create a product and set a barcode
3. On the product page, click Actions (cog wheel) > Print Labels
4. Print with any of the following format:
    - Dymo
    - 2 x 7 with price
    - 4 x 7 with price
    - 4 x 12
    - 4 x 12 with price
5. **For the Dymo format, the barcode is not displayed properly. For the four other formats, a blank page is printed**

## Cause
[This commit](https://github.com/odoo/odoo/commit/b7ec60d68c6ee23f7684960e33e5bd6290c9d829) made the `t-call` tag parametric, replacing the `t-set` and `t-value` pairs in various contexts. In this case, the parameters were not added to the correct `t-call`. Without `columns` and `rows` defined, the `report_productlabel` template would not produce anything:

https://github.com/odoo/odoo/blob/b6598c11a3580cfc6ff1ffeb67b0d388c55383d1/addons/product/report/product_product_templates.xml#L129-L131

opw-5877421

### DIFF
--- a/addons/product/report/product_product_templates.xml
+++ b/addons/product/report/product_product_templates.xml
@@ -193,16 +193,16 @@
         </template>
 
         <template id="report_productlabel_dymo">
-            <t t-call="web.html_container"
-                barcode_size.f="width:45.5mm;height:7.5mm"
-                table_style.f="width:100%;height:32mm;"
-                padding_page.f="padding: 2mm">
+            <t t-call="web.html_container">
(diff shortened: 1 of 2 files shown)


barracudapps (approved):
LGTM 👍 
Than you for the fix!

PR #246467, opened 2026-01-30
[IMP] mail: proper store versioning

tsm-odoo:
\*: cloud_storage, im_livechat.

In the discuss app, store data is received from both RPC  and bus notifications. Because these updates can arrive out of order (e.g. a long-running RPC returning after a newer bus notification has already been processed), outdated data can overwrite newer data in the store. This leads to inconsistent UI states where messages or channel properties appear to roll back to previous values.

For now, bus notification often arrives ordered, but improving message dispatching would require to fetch notifications from different channels in parallel, which would only guarantee order *per channel*. This change would make this issue even more noticeable.

This commit introduces a decorator, `store_version` which return information about which transactions were visible to the client. With those metadata, store is now able to know whether an update represents a newer version of a field, thus solving the issue.

Another decorator is introduced, `mail_route` which applies `route`, `add_guest_to_context` and the `store_version` decorator, in the correct order.

task-4966085

Perfs, avg on 5 runs. 1k5 channels, 1k5 members:
- RPC
  - Initial page load rpc (`channels_as_member`, `init_messaging`, `failures`, `systray_get_activities`: Before ~3s , After ~3s.
  - Message post: Before ~80ms, After ~80ms.
- Store insert: Before ~9.5s, After ~9s.
Note: 9s is bad but unrelated to the PR, this needs to be fixed with OWL 3 / signals.

enterprise: https://github.com/odoo/enterprise/pull/106986

### DIFF
--- a/addons/im_livechat/controllers/attachment.py
+++ b/addons/im_livechat/controllers/attachment.py
@@ -3,15 +3,14 @@
 from werkzeug.exceptions import NotFound
 
 from odoo import _
-from odoo.http import route, request
+from odoo.http import request
 from odoo.exceptions import AccessError
 from odoo.addons.mail.controllers.attachment import AttachmentController
-from odoo.addons.mail.tools.discuss import add_guest_to_context
+from odoo.addons.mail.tools.discuss import mail_route
 
 
 class LivechatAttachmentController(AttachmentController):
-    @route()
-    @add_guest_to_context
+    @mail_route(type="http")
     def mail_attachment_upload(self, ufile, thread_id, thread_model, is_pending=False, **kwargs):
         thread = self._get_thread_with_access_for_post(thread_model, thread_id, **kwargs)
         if not thread:
--- a/addons/mail/controllers/poll.py
+++ b/addons/mail/controllers/poll.py
@@ -2,13 +2,12 @@ from datetime import timedelta
 
 from odoo import fields
 from odoo.addons.mail.controllers.thread import ThreadController
-from odoo.addons.mail.tools.discuss import add_guest_to_context, Store
+from odoo.addons.mail.tools.discuss import mail_route, Store
 from odoo.fields import Command, Domain
-from odoo.http import route
 
 
 class PollController(ThreadController):
-    @route("/mail/poll/create", type="jsonrpc", auth="user", methods=["POST"])
+    @mail_route("/mail/poll/create", type="jsonrpc", auth="user", methods=["POST"])
     def poll_create(
         self,
         duration,
@@ -40,7 +39,7 @@ class PollController(ThreadController):
         Store(**thread._store_target()).add(poll, "_store_poll_fields").bus_send()
         return poll.id
 
-    @route("/mail/poll/end", type="jsonrpc", auth="user", methods=["POST"])
+    @mail_route("/mail/poll/end", type="jsonrpc", auth="user", methods=["POST"])
     def poll_end(self, poll_id):
         poll_domain = Domain("id", "=", poll_id) & Domain("end_message_id", "=", False)
         if not self.env.user._is_admin():
@@ -48,7 +47,7 @@ class PollController(ThreadController):
         # sudo - mail.poll: creator of the poll or admins can end a poll prematurely.
         self.env["mail.poll"].sudo().search_fetch(poll_domain)._end_and_notify()
 
-    @route("/mail/poll/delete", type="jsonrpc", auth="user", methods=["POST"])
+    @mail_route("/mail/poll/delete", type="jsonrpc", auth="user", methods=["POST"])
     def poll_delete(self, poll_id):
         poll_domain = Domain("id", "=", poll_id)
         if not self.env.user._is_admin():
@@ -56,8 +55,7 @@ class PollController(ThreadController):
         # sudo - mail.poll: creator of the poll or admins can delete a poll.
         self.env["mail.poll"].sudo().search_fetch(poll_domain).unlink()
 
-    @route("/mail/poll/vote", type="jsonrpc", auth="public", methods=["POST"])
-    @add_guest_to_context
+    @mail_route("/mail/poll/vote", type="jsonrpc", auth="public", methods=["POST"])
     def poll_vote(self, poll_id, option_ids):
         options_domain = [("poll_id", "=", poll_id), ("id", "in", option_ids)]
         # sudo - mail.poll.option: can access poll options, vote access is validated
@@ -86,8 +84,7 @@ class PollController(ThreadController):
         store.add(vote, "_store_vote_fields")
         store.bus_send()
 
-    @route("/mail/poll/remove_vote", type="jsonrpc", auth="public", methods=["POST"])
-    @add_guest_to_context
+    @mail_route("/mail/poll/remove_vote", type="jsonrpc", auth="public", methods=["POST"])
     def poll_remove_vote(self, poll_id):
         votes_domain = [("option_id.poll_id", "=", poll_id), ("is_self_vote", "=", True)]
         # sudo - mail.poll.vote: removing/accessing self vote is allowed.
--- a/addons/mail/models/models.py
+++ b/addons/mail/models/models.py
@@ -24,6 +24,13 @@ class Base(models.AbstractModel):
     # ORM
     # ------------------------------------------------------------
 
+    def _flush(self):
+        if mail_store := self.env.context.get("mail_store"):
+            for field in self._fields.values():
+                if ids := self.env._field_dirty.get(field):
+                    mail_store.mark_field_as_written(field.model_name, ids, field.name)
+        return super()._flush()
+
     def _valid_field_parameter(self, field, name):
         # allow tracking on abstract models; see also 'mail.thread'
         return (
--- a/addons/mail/static/src/model/field_version.js
+++ b/addons/mail/static/src/model/field_version.js
@@ -1,0 +1,280 @@
+/**
+ * A nested mapping of fields, modified during a request. Organized by `model_name` ->
+ * `record_id` -> `field_names[]`.
+ * @typedef {{[model_name: string]: {[record_id: number]: string[]}}} WrittenFieldsByRecord
+ */
+
+/**
+ * Represents a PostgreSQL transaction snapshot. See:
+ * https://www.postgresql.org/docs/13/functions-info.html#FUNCTIONS-PG-SNAPSHOT-PARTS
+ *
+ * @typedef {Object} IPgSnapshot
+ * @property {string} xmin The lowest active transaction ID at the time this snapshot was
+ * taken. Lower transaction IDs are completed.
+ *
+ * @property {string} xmax The upper bound of transaction IDs for this snapshot. Greater
+ * or equals transaction IDs are invisible by this snapshot.
+ *
+ * @property {string} xip_bitmap A bitmap representing in progress transactions in the
+ * range [xmin, xmax). Each bit corresponds to a transaction ID within this range. If the
+ * bit is set, the transaction was in progress at the time this snapshot was taken.
+ *
+ * @property {string|null} current_xact_id The current transaction ID, assigned when a
+ * transaction modify the database.
+ *
+ * A versioned field value tied to a database snapshot. It tracks what this revision "saw"
+ * when this value was set, allowing the client to correctly order updates and ignore late
+ * arriving, stale data.
+ *
+ * @typedef {Object} FieldRevision
+ * @property {PgSnapshot} snapshot The transactional context defining what the current
+ * revision could "see" in the database.
+ * @property {boolean} isWrite Whether this revision modified the field.
+ */
+
+/**
+ * Version metadata sent by the server alongside insert data. Used to determine if
+ * incoming insert values should override stored ones.
+ * @typedef {{snapshot: IPgSnapshot, written_fields_by_record: WrittenFieldsByRecord}} StoreVersion
+ */
+
+/**
+ * VISUAL REPRESENTATION OF THE SNAPSHOT.
+ * Legend: √ committed, ~ pending, × future.
+ *
+ *                                +-------------------------+
+ *                                |     PENDING COUNT (4)   |
+ *                     +--------> +-------------------------+ <------+------+
+ *          XMIN 5     |                        ^                    |      |     XMAX 12
+ *               v     |                        |                    |      |     v
+ *               +--------+--------+--------+--------+--------+---------+---------+
+ *    √ √ √ √    | TX_5 ~ | TX_6 √ | TX_7 √ | TX_8 ~ | TX_9 √ | TX_10 ~ | TX_11 ~ |    × × × ×
+ *  <--------->  +--------+--------+--------+--------+--------+---------+---------+  <--------->
+ *       |                     |        |                |
+ *       |                     |        |                |
+ *       |                     v        v                v
+ *       +----------------> +-------------------------------------------------+
+ *                          |                  FINISHED COUNT                 |
+ *                          |                                                 |
+ *                          |    xmax - 1 - len(xip) = 12 - 1 - 4 = 7         |
+ *                          | OR xmin - 1 + len(not_in_xip) =  5 - 1 + 3 = 7  |
+ *                          +-------------------------------------------------+
+ */
+export class PgSnapshot {
+    /** @param {IPgSnapshot} */
+    constructor(params) {
+        this.current_xact_id = params.current_xact_id
+            ? BigInt(params.current_xact_id)
+            : params.current_xact_id;
+        this.xmin = BigInt(params.xmin);
+        this.xmax = BigInt(params.xmax);
+        const bitmapBinaryStr = atob(params.xip_bitmap);
+        // Bitmap [xmin, xmax) showing which xact_ids are in progress of the time of the snapshot.
+        this.xip_bitmap = new Uint8Array(bitmapBinaryStr.length).map((_, idx) =>
+            bitmapBinaryStr.charCodeAt(idx)
+        );
+        let pendingCount = 0;
+        for (let byte of this.xip_bitmap) {
+            while (byte > 0) {
+                byte &= byte - 1;
+                pendingCount++;
+            }
+        }
+        this.finishedCount = this.xmax - 1n - BigInt(pendingCount);
+    }
+
+    /**
+     * Determine whether the given transaction was visibile by this snapshot.
+     *
+     * @param {BigInt} txid
+     */
+    knowsTransaction(txid) {
+        // tx < xmin are completed. tx between xmin and xmax are completed if not in xip.
+        if (txid >= this.xmin && txid < this.xmax) {
+            const offset = Number(txid - this.xmin);
+            const byteIndex = Math.floor(offset / 8);
+            const bitIndex = offset % 8;
+            return !(this.xip_bitmap[byteIndex] & (1 << bitIndex));
+        }
+        return txid < this.xmin;
+    }
+
+    /**
+     * Determine whether the given revision comes from a newer snapshot than the current
+     * one (i.e. if the given revision knows more committed transactions than the current
+     * one).
+     *
+     * @param {PgSnapshot} other
+     */
+    isAfter(other) {
+        return this.finishedCount > other.finishedCount;
+    }
+
+    /**
+     * Checks whether this snapshot and another snapshot saw the same state of the database.
+     *
+     * @param {PgSnapshot} other
+     */
+    sawSameState(other) {
+        return this.finishedCount === other.finishedCount;
+    }
+}
+
+/**
+ * Determines whether a candidate field revision should replace the current one.
+ *
+ * @param {FieldRevision} candidate
+ * @param {FieldRevision} other
+ * @returns {Boolean} Whether the candidate revision can replace the other one.
+ */
+function shouldReplace(candidate, other) {
+    if (candidate.snapshot.sawSameState(other.snapshot)) {
+        return true;
+    }
+    // candidate is a read:
+    // - vs write: can replace if the candidate revision knows about the transaction the
+    //   write originates from.
+    // - vs read: can replace if the candidate revision comes from a newer snapshot (i.e.
+    //   this revision knows more committed transactions).
+    if (!candidate.isWrite) {
+        return other.isWrite
+            ? candidate.snapshot.knowsTransaction(other.snapshot.current_xact_id)
+            : candidate.snapshot.isAfter(other.snapshot);
+    }
+    // Candidate is a write: we can replace any revision that didn't know
+    // about this transaction.
+    return !other.snapshot.knowsTransaction(candidate.snapshot.current_xact_id);
+}
+
+export const SKIP_REVISION = Symbol("SKIP");
+
+/**
+ * Track a single value field's latest revision and allow to determine if a new value can
+ * be applied based on snapshot visibility.
+ */
+export class SingleFieldVersion {
+    lastRevision = {
+        snapshot: new PgSnapshot({ xmin: 0, xmax: 0, xip_bitmap: "" }),
+        isWrite: false,
+    };
+
+    /**
+     * Determine if the current revision can replace the given one.
+     *
+     * @template T
+     * @param {T} value
+     * @param {FieldRevision} incomingRevision
+     * @returns {typeof SKIP_REVISION|T} The skip symbol, or the value to update the
+     * field.
+     */
+    resolveApply(value, incomingRevision) {
+        if (shouldReplace(incomingRevision, this.lastRevision)) {
+            this.lastRevision = incomingRevision;
+            return value;
+        }
+        return SKIP_REVISION;
+    }
+}
+
+/**
+ * Track a multi value field's command history and determine which commands can be applied
+ * based on revision snapshots.
+ */
+export class ManyFieldVersion {
+    /** @type {import("@mail/model/record").Record} */
+    TargetModel;
+    /**
+     * Tracks the command history for this field, in chronological order. Each entry
+     * represents a single command along with the revision at which it was applied.
+     *
+     * @type {{cmd: [], revision: FieldRevision}[]}
+     */
+    history = [
+        {
+            cmd: ["REPLACE", []],
+            revision: {
+                snapshot: new PgSnapshot({ xmin: 0, xmax: 0, xip_bitmap: "" }),
+                isWrite: false,
+            },
+        },
+    ];
+
+    constructor(TargetModel) {
+        this.TargetModel = TargetModel;
+    }
+
+    /**
+     * Determine what commands should be applied.
+     *
+     * @param {Array[]} commands
+     * @param {FieldRevision} incomingRevision
+     * @returns {Array[]|typeof SKIP_REVISION} The skip symbol, or the commands to apply
+     * to update the field.
+     */
+    resolveApply(commands, incomingRevision) {
+        if (!shouldReplace(incomingRevision, this.history[0].revision)) {
+            return SKIP_REVISION;
+        }
+        const insertionIndex = this._findInsertionIndex(incomingRevision);
+        const insertAtTheEnd = insertionIndex === this.history.length;
+        this.history.splice(
+            insertionIndex,
+            0,
+            ...commands.map((cmd) => ({ cmd, revision: incomingRevision }))
+        );
+        const lastReplaceIndex = this.history.findLastIndex((entry) => entry.cmd[0] === "REPLACE");
+        if (lastReplaceIndex >= insertionIndex) {
+            this.history = this.history.slice(lastReplaceIndex);
+        }
+        if (insertAtTheEnd) {
+            return commands;
+        }
+        return this._generateReplaceFromHistory();
+    }
+
+    /**
+     * Returns the index of the first element strictly greater than the given revision.
+     * This ensures identical revisions are kept in their original arrival order.
+     *
+     * @param {FieldRevision} revision
+     */
+    _findInsertionIndex(revision) {
+        let start = 0;
+        let end = this.history.length;
+        while (start < end) {
+            const mid = Math.floor((start + end) / 2);
+            const midRevision = this.history[mid].revision;
+            if (shouldReplace(revision, midRevision)) {
+                start = mid + 1;
+            } else {
+                end = mid;
+            }
+        }
+        return start;
+    }
+
+    get lastRevision() {
+        return this.history.at(-1).revision;
+    }
+
+    /** Returns a replace command, equivalent to all the commands in history. */
+    _generateReplaceFromHistory() {
+        const positionByLocalId = {};
+        for (let idx = 0; idx < this.history.length; idx++) {
+            const [name, values] = this.history[idx].cmd;
+            for (let subIdx = 0; subIdx < values.length; subIdx++) {
+                const value = values[subIdx];
+                const localId = this.TargetModel.localId(value);
+                if (["REPLACE", "ADD", "ADD.noinv"].includes(name)) {
+                    positionByLocalId[localId] ??= { value, idx, subIdx };
+                } else {
+                    delete positionByLocalId[localId];
+                }
+            }
+        }
+        const sortedValues = Object.values(positionByLocalId)
+            .sort((a, b) => a.idx - b.idx || a.subIdx - b.subIdx)
+            .map((p) => p.value);
+        return [["REPLACE", sortedValues]];
+    }
+}
--- a/addons/mail/static/src/model/store.js
+++ b/addons/mail/static/src/model/store.js
@@ -1,4 +1,5 @@
 import { reactive } from "@web/owl2/utils";
+import { PgSnapshot } from "@mail/model/field_version";
 import { Record } from "./record";
 import { STORE_SYM, modelRegistry } from "./misc";
 import { toRaw } from "@odoo/owl";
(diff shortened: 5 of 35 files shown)


addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
...
+                    payload["__store_version_metadata__"] = store_version_metadata
+                    return True
+                for value in payload.values():
+                    if isinstance(value, dict):
+                        if insert_metadata(value):
+                            return True
+                return False
+
+            if pending_bus_send := store_versioning_ctx["pending_bus_send"]:
+                for target, notification_type, msg in pending_bus_send:
```
seb-odoo: `if` likely not necessary, iterating on empty list does the same

addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
...
+
+            def insert_metadata(payload):
+                if not isinstance(payload, dict):
+                    return False
+                if payload.pop("__add_version_meta__", False):
+                    payload["__store_version_metadata__"] = store_version_metadata
+                    return True
+                for value in payload.values():
+                    if isinstance(value, dict):
+                        if insert_metadata(value):
```
seb-odoo: Could merge into a single line

addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
...
+                    model: {record_id: list(fields) for record_id, fields in records.items()}
+                    for model, records in store_versioning_ctx["written_fields_by_record"].items()
+                },
+            }
+
+            def insert_metadata(payload):
+                if not isinstance(payload, dict):
+                    return False
+                if payload.pop("__add_version_meta__", False):
+                    payload["__store_version_metadata__"] = store_version_metadata
```
seb-odoo: `__store_version__` likely sufficient as key, maybe even just `__version__`?

addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
...
+                    return True
+                for value in payload.values():
+                    if isinstance(value, dict):
+                        if insert_metadata(value):
+                            return True
```
seb-odoo: This step is a bit strange/magic. What does it do exactly, if there's a dict inside the top-level dict, it adds into the first sub dict?
What makes this one the correct one?

addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
...
+            """)
+            snapshot = req.env.cr.fetchone()
+            store_version_metadata = {
+                "snapshot": {
+                    "xmin": int(snapshot[0]),
+                    "xmax": int(snapshot[1]),
+                    "xip": [int(xid) for xid in snapshot[2]],
+                    "current_xact_id": int(snapshot[3]) if snapshot[3] else None,
+                },
+                "written_fields_by_record": {
```
seb-odoo: technically `by_id_by_model` and not `by_record`, but I don't know if I want the name to be even longer

addons/mail/tools/discuss.py
```diff
@@ -32,12 +33,103 @@ def wrapper(self, *args, **kwargs):
         if guest:
             req.update_context(guest=guest)
             if isinstance(self, models.BaseModel):
-                self = self.with_context(guest=guest)
+                self = self.with_context(self.env.context, guest=guest)
```
seb-odoo: useless

addons/mail/static/src/core/common/message_model.js
```diff
@@ -643,7 +643,7 @@ export class Message extends Record {
             update_data: updateData,
             ...this.thread.rpcParams,
         });
-        this.store.insert(data);
+        this.store.insert(data.payload);
```
seb-odoo: Should revert, version should be part of store data

addons/mail/tools/discuss.py
tsm-odoo: Yes exactly, I feel like since it's only used at one place client side, and is very technichal its ok like this "record" implies that its linked to a model^^

addons/mail/tools/discuss.py
tsm-odoo: kep "__store_version__" for easier greps

addons/mail/controllers/poll.py
```diff
@@ -2,9 +2,8 @@
 
 from odoo import fields
 from odoo.addons.mail.controllers.thread import ThreadController
-from odoo.addons.mail.tools.discuss import add_guest_to_context, Store
+from odoo.addons.mail.tools.discuss import mail_route as route, Store
```
seb-odoo: Strange inconsistency here.
Wouldn't we want to import it as `route` in all files to simplify the syntax? And maybe export it as such too?
Or if we want to very well see these are `mail_route` specifically, I'd suggest to use the full name in this file too.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -10,7 +10,7 @@
 from odoo import _, api, fields, models, tools, Command
 from odoo.addons.base.models.avatar_mixin import get_random_ui_color_from_seed
 from odoo.addons.base.models.ir_mail_server import MailDeliveryException
-from odoo.addons.mail.tools.discuss import Store
+from odoo.addons.mail.tools.discuss import Store, store_version
```
seb-odoo: It's unused (for now?)

addons/mail/static/src/core/common/store_service.js
```diff
@@ -88,25 +92,7 @@ export class Store extends BaseStore {
...
-                browser.localStorage.setItem(
-                    "mail.user_setting.push_notification_dismissed",
-                    "true"
-                );
-            } else {
-                browser.localStorage.removeItem("mail.user_setting.push_notification_dismissed");
-            }
-        },
-    });
+    isNotificationPermissionDismissed = fields.Attr(false, { localStorage: true });
```
seb-odoo, line 97: rebase issue or unrelated change?

addons/mail/tools/discuss.py
```diff
@@ -202,6 +318,8 @@ def get_client_action(self):
     def get_result(self):
         """Gets resulting data built from adding all data together."""
         res = {}
+        if self._add_version_flag:
+            res["__add_version__"] = True
```
seb-odoo: I'm not sure if we really need this.
`_add_version_flag` seems overkill: either we are in the wrapper and we always want it, or we are not in the wrapper and it's going to do nothing anyway.

Detection of Store in result could be done with checking instance rather than presence of `__add_version__`.

addons/mail/tools/discuss.py
```diff
@@ -214,7 +332,18 @@ def bus_send(self, notification_type="mail.record/insert", /):
...
+                notification_type == "mail.record/insert"
+                and "mail_store_versioning" in self.target.channel.env.context
+            ):
+                self.target.channel.env.context["mail_store_versioning"]["pending_bus_send"].append(
+                    (self.target, notification_type, res)
+                )
+            else:
+                self.target.channel._bus_send(
+                    notification_type, res, subchannel=self.target.subchannel
+                )
```
seb-odoo: Note: this will change order of notifications, if other notifications are sent in the same request without the wrapper.

This might or might not be a problem depending on what is being sent.

If notif A contained obsoleted data (different part of the flow) and notif B was later sent to "correct it", changing their order would not be acceptable.

addons/mail/tools/discuss.py
seb-odoo: Edit: I guess not for the 2nd part, as we return the result and not the store unfortunately.

Now that we have the wrapper this could be changed though, I always wished we only returned the store in business and `get_result()` was only called by framework code when necessary (which was achieved for bus, not yet for rpc, but this PR can change it).

addons/mail/static/src/model/store_internal.js
```diff
@@ -227,12 +228,33 @@ export class StoreInternal extends RecordInternal {
...
+     * }} [options.versionMetadata]
      */
-    updateFields(record, vals) {
+    updateFields(record, vals, { versionMeta } = {}) {
         const fieldEntries = Object.entries(vals).concat(
             Object.getOwnPropertySymbols(vals).map((sym) => [sym, vals[sym]])
         );
+        let snapshot;
         for (const [fieldName, value] of fieldEntries) {
+            // Skipping many versioning for now, as it requires more thoughts.
```
seb-odoo: What problems with Many did you see?

seb-odoo: I'm pretty sure `replace` command can be applied with the same logic as the other fields.

For `add` and `delete` I guess we just need to check if their transaction saw (the last "replace" transaction AND the last "add" or "delete") that was applied?
If yes we can apply them, if not we have to ignore them?

addons/mail/tests/discuss/__init__.py
```diff
@@ -19,6 +19,7 @@
 from . import test_discuss_thread_controller
 from . import test_guest
 from . import test_message_controller
+from . import test_store_versioning
```
seb-odoo, line 23: alphabetical

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -6,8 +6,6 @@ import { _t } from "@web/core/l10n/translation";
 import { user } from "@web/core/user";
 import { Deferred } from "@web/core/utils/concurrency";
 
-/** @import { AwaitChatHubInit } from "@mail/core/common/chat_hub_model" */
```
seb-odoo: Also unrelated change?

addons/mail/static/src/model/record_field_version.js
```diff
@@ -0,0 +1,88 @@
+/** @typedef {{  current_xact_id: number|null, xmin: number, xmax: number, xip: Set<number> }} PGSnapshot */
+
+/** @type {PGSnapshot} */
+const INITIAL_FIELD_SNAPSHOT = { current_xact_id: null, xmin: 0, xmax: 0, xip: new Set() };
```
seb-odoo: This looks like pattern of retained memory in unit tests.
We might need to have a function that returns individual objects?

addons/mail/controllers/poll.py
tsm-odoo: I'd rather keeping `mail_route` everywhere to quickly identify which routes are/are not using the versioning.

addons/mail/tools/discuss.py
tsm-odoo: Used a type flag `StoreResult(dict)`, this way we can recognize it easily. The store change will be done as part of https://www.odoo.com/odoo/1519/tasks/5242369.

addons/mail/tools/discuss.py
tsm-odoo: Everything that impacts the store should go through `record/insert`.

addons/mail/static/src/model/store_internal.js
tsm-odoo: Makes sense. I just didn't think it through yet as I wanted to make it work for simple fields first.

addons/mail/tests/discuss/__init__.py
tsm-odoo: It's placed correctly. Everything up to this line is sorted properly. The lines below are a mess^^

addons/mail/tools/discuss.py
```diff
@@ -81,6 +194,14 @@ def get_sfu_key(env) -> str | None:
 
 NO_VALUE = object()
 
+
+class StoreResult(dict):
+    """Marker class for dictionaries returned by `Store.get_result()`.
+    Used to distinguish store results from arbitrary dicts so version
+    metadata can be added (see `store_version` decorator).
+    """
```
seb-odoo: All store sub-class are usually wrapped inside the main store class.

addons/mail/static/src/model/store_internal.js
seb-odoo: I think we can actually go further.
A more recent `replace` can override an older one, but if we already had `add` or `delete` that are newer than even the newer replace, we could re-apply them on top of the newer replace.

addons/mail/tools/discuss.py
tsm-odoo: This one is purely technical and there is no point in using it in the business code. Do you still want to add it to `Store`?

addons/mail/static/src/model/store_internal.js
tsm-odoo: Yeah, basically we have to keep a stack. Apply the `replace` then rewind the actions that are newer than this replace. We need to determine the order of operations and we need to determine when the stack should be cleaned up as we don't wanna end up storing every operation forever.

seb-odoo: Items in the stack are cleared whenever there is a `replace` newer than them.

We currently send `add` or `delete` in flows were we should send `replace` to avoid race conditions (mostly in store methods), but this PR allows to send the proper `replace` instead.

seb-odoo: And order of the stack uses the same comparison as when we compare 2 store insert to determine if one should override the other. The one that wins is applied afterwards.

tsm-odoo: I get that "we should" send replace instead of "add/delete" but I don't like that the stack can grow indefinitely in case we never receive any "replace" commands. Yeah the current logic allows us to determine the order  in which commands should be applied.

seb-odoo: It's nice to keep it in mind but it's probably a premature optimization to worry about that.
Also everything grows infinitely in store, as we don't do garbage collect anyway.

If we wanted we could collapse the stack into a single equivalent `replace` with all the known ids at a given time, either time based or number of items in the stack based. We just need to leave a decent window to let more recent values arrive.

addons/mail/tools/discuss.py
seb-odoo: They are strongly related and it's not really useful outside of Store specific code so I would say it belongs inside yes.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -1368,9 +1368,6 @@ def _lazy_load_members_channel_types(self):
     def channel_set_custom_name(self, name):
         self.ensure_one()
         self.self_member_id.custom_channel_name = name
-        store = Store(bus_channel=self.self_member_id._bus_channel())
-        store.add(self.self_member_id, ["custom_channel_name"])
-        store.bus_send()
```
tsm-odoo: Note, duplicate notification. We already send the update via the `bus.sync.mixin` mechanism.

seb-odoo: Also note: another pending PR is removing `custom_channel_name` in master

addons/mail/static/src/model/field_versions.js
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @param {FieldRevision} candidate
+ * @param {FieldRevision} other
+ * @returns {Boolean} Whether the incoming revision can override the
+ * current one.
+ */
+function canOverride(candidate, other) {
+    // candidate is a read:
+    // - vs write: can override if the other revision knows about the
+    //   transaction the write originates from.
```
seb-odoo: wrong word? can override if the _candidate_ revisions knows about the write?

addons/mail/static/src/model/field_versions.js
```diff
@@ -0,0 +1,177 @@
...
+    // - vs read: can override if the candidate revision comes from a newer
+    //   snapshot (i.e. this revision knows more committed
+    //   transactions).
+    if (!candidate.isWrite) {
+        return other.isWrite
+            ? candidate.snapshot.knowsTransaction(candidate.snapshot.current_xact_id)
```
seb-odoo: param should use `other` and not `candidate`?

addons/mail/tools/discuss.py
```diff
@@ -38,6 +40,120 @@ def wrapper(self, *args, **kwargs):
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            if "mail_store_versioning" in self.env.context:
+                return func(self, *args, **kwargs)
+            store_versioning_ctx = {
+                "pending_bus_send": [],
+                # { model_name: { record_id: set() } }
+                "written_fields_by_record": defaultdict(
+                    lambda: defaultdict(lambda: OrderedSet("write_uid", "write_date"))
```
seb-odoo: There is a syntax error here.

Also I would think this should only be added in write for models that have log field enabled.

Manually adding a field in the list (through manual query) should not have these values as the ORM won't be used.

addons/portal_rating/static/src/chatter/portal/chatter_patch.js
seb-odoo, line 1: several files incorrectly removed?

addons/mail/tools/discuss.py
```diff
@@ -38,6 +40,150 @@ def wrapper(self, *args, **kwargs):
...
+                model: {record_id: list(fnames) for record_id, fnames in recs.items()}
+                for model, recs in self.written_fields.items()
+            },
+        }
+
+    def _inject_version(self, payload):
+        """Add the version to the return value of `Store.get_result()`. Either the
+        payload itself, or at the first level of the payload.
+        """
+        metadata = self._get_formatted_version()
```
ThanhDodeurOdoo: can it be moved below so that there is no flush and snapshot collection if we early return?

addons/im_livechat/controllers/attachment.py
```diff
@@ -3,15 +3,14 @@
...
 from odoo.exceptions import AccessError
 from odoo.addons.mail.controllers.attachment import AttachmentController
-from odoo.addons.mail.tools.discuss import add_guest_to_context
+from odoo.addons.mail.tools.discuss import mail_route
 
 
 class LivechatAttachmentController(AttachmentController):
-    @route()
-    @add_guest_to_context
+    @mail_route()
```
ThanhDodeurOdoo: maybe I didnt follow the flow correctly, but could it add `mail_store` to the context, then cause an early return here ( https://github.com/odoo-dev/odoo/blob/master-discuss-order-tsm/addons/mail/tools/discuss.py#L154 , which would return a `Response`)  when calling super below, then early-returning in `_inject_version` without adding `__store_version__`?

addons/mail/static/tests/core/store_versioning.test.js
```diff
@@ -0,0 +1,480 @@
...
+                expected: { name: "v3" },
+                description: "V2 read's snapshot is older.",
+            },
+            {
+                values: { name: "v4" },
+                meta: {
+                    snapshot: {
+                        xmin: 65,
+                        xmax: 68,
+                        xip_bitmap: xipToBitmap(65, 68, [(65, 66)]),
```
ThanhDodeurOdoo: maybe a typo? `[(65, 66)]` === `[66]`

seb-odoo (approved):
Ok for me but I've seen it too much during dev so I might miss things.
It would be nice to have a double check for someone else.

@robodoo delegate+

addons/mail/tools/discuss.py
```diff
@@ -38,6 +40,157 @@ def wrapper(self, *args, **kwargs):
...
+        """Get the version metadata, used by the client to determine if an incoming
+        store insert is newer than what it already knows.
+        """
+        self.env.flush_all()  # Ensure xact_id is properly assigned before querying the snapshot.
+        self.env.cr.execute("SELECT pg_current_snapshot(), pg_current_xact_id_if_assigned()")
+        snapshot_str, current_xact_id = self.env.cr.fetchone()
+        xmin_str, xmax_str, xips_str = snapshot_str.split(":")
+        xmin = int(xmin_str)
+        xmax = int(xmax_str)
+        xips = {int(x) for x in xips_str.split(",") if x}
```
phenix-factory: What's wrong with `pg_snapshot_xip`, `pg_snapshot_xmax`, and `pg_snapshot_xmin`? Can't you fetch them directly from the query? 

https://www.postgresql.org/docs/current/functions-info.html

tsm-odoo: 4 selects instead of one. `pg_snapshot_xip(pg_current_snapshot()), pg_snapshot_xmin(pg_current_snapshot()), pg_snapshot_xmax(pg_current_snapshot())`. And we should probably cast`xip` to text array to be able to retrieve it as a list python side. So all in all, getting everything at once and parsing them afterwards is not that bad.

ThanhDodeurOdoo (approved):

addons/mail/static/src/model/field_versions.js
```diff
@@ -0,0 +1,223 @@
...
+        );
+        this.pendingTransactionCount = 0;
+        for (let byte of this.xipBitmap) {
+            while (byte > 0) {
+                byte &= byte - 1;
+                this.pendingTransactionCount++;
+            }
+        }
+        this.committedTransactionCount = this.xmax - 1n - BigInt(this.pendingTransactionCount);
+    }
```
alexkuhn, line 84: Documentation for params is missing.
`xmin` / `xmax` are related to pg snapshot, would be nice to explain again what they mean.
If we want to migrate to some other DB solution for Discuss, which could be a realistic option, we should make it easier to translate these concepts to another system.

I don't understand `xip_bitmap`, is it only to have a better format for range of `xmin` and `xmax`? The server seems to produce this listing, but tests seem to pass partial in listing which doesn't make sense with looking at python code. Again the lack of documentation is a big problem for such technical details that are very important for the good working of JS models.

tsm-odoo: Ah yes, documentation link was sitting in `knowsTransaction`. I moved it next to type definition (https://www.postgresql.org/docs/13/functions-info.html#FUNCTIONS-PG-SNAPSHOT-PARTS)

tsm-odoo: As for xip bitmap it's indeed a representation of xip_list, only smaller as sending a list of 64 bits represented as string might be huge. Tests build it as the server does so I'm not sure what you mean by partial listing.

alexkuhn: I would suggest to have docstring above to explain the core concepts:
- transaction (id)
- `xmin`
- `xmax`
- `xip_bitmap` (explain the subtleties around intermediate transactions not being "known")

I think that if this is explains, the code would be much easier to understand.

addons/mail/static/src/model/field_versions.js
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/**
+ * Determine if a candidate revision can override the other revision.
+ *
+ * @param {FieldRevision} candidate
+ * @param {FieldRevision} other
+ * @returns {Boolean} Whether the incoming revision can override the current one.
+ */
+function canOverride(candidate, other) {
```
alexkuhn: Explain a bit more what this means by "override". This docstring doesn't help much more than the function name. Could provide a brief example by what this means for a field revision to "can override".

addons/mail/static/src/model/field_version.js
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Determine if a snapshot has the same horizon than this snapshot. (i.e. they saw the
+     * same DB version).
+     *
+     * @param {PgSnapshot} other
+     */
+    hasSameHorizon(other) {
+        return this.committedTransactionCount === other.committedTransactionCount;
+    }
```
alexkuhn, line 120: The part in bracket is extremely important to understand the concept. The "horizon" terminology is very cryptic and requires to get used to the unintuitive jargon.

I don't mind to keep using `hasSameHorizon` function name, but I would avoid using "horizon" to define "horizon"... Same DB version is very clear and should be seen first and foremost, so that we learn without ambiguity that "horizon = DB version" of some sort.

addons/mail/static/src/model/field_versions.js
seb-odoo: Just for reference.

In sql `xmin` is redundant, it's the first item in `xip_list` (if not empty, otherwise it's equal `xmax`), it's only there for performance reasons (direct memory lookup instead of looking first element of list is faster when done million times).

With the compressing we do, in JS `xmin` is necessary to know where is the start of the bitmap.

There is a bitmap to significantly reduces the size of what we send. It's much cheaper to send a bit (yes/no) to tell for each item between xmin and xmap when the item is present in the list or not (size = xmax - xmin), rather than sending the ids in full (size = up to 64 bits for each item in the list + possibly even more as we have to send them as string).

addons/mail/static/src/model/field_version.js
```diff
@@ -0,0 +1,246 @@
...
+        this.current_xact_id = params.current_xact_id
+            ? BigInt(params.current_xact_id)
+            : params.current_xact_id;
+        this.xmin = BigInt(params.xmin);
+        this.xmax = BigInt(params.xmax);
+        const bitmapBinaryStr = atob(params.xip_bitmap);
+        // Bitmap [xmin, xmax) showing which xact_ids are in progress of the time of the snapshot.
+        this.xipBitmap = new Uint8Array(bitmapBinaryStr.length).map((_, idx) =>
+            bitmapBinaryStr.charCodeAt(idx)
+        );
```
alexkuhn, line 75: Nitpick: would name it `xip_bitmap` like the interface, for consistency.

addons/mail/static/src/model/field_version.js
```diff
@@ -0,0 +1,246 @@
...
+        this.pendingTransactionCount = 0;
+        for (let byte of this.xipBitmap) {
+            while (byte > 0) {
+                byte &= byte - 1;
+                this.pendingTransactionCount++;
+            }
+        }
+        this.committedTransactionCount = this.xmax - 1n - BigInt(this.pendingTransactionCount);
```
alexkuhn: I think that wording with `Count` for `commitedTransactionCount` is misleading: the `pendingTransactionCount` is fine because we understand this is the amount of pending transactions in the current snapshot, so 0 or 3, etc.

However here from my understanding this should be read more like an "effective x", if I understand it correctly?

I would personally suggest naming it `effectiveX` or something equivalent, to use equivalent wording as `xmin` / `xmax` and make it clearer what this represents. Comparing before / after / same in functions below would be stupidly easy to get, to the point we wonder if we should keep these functions.

tsm-odoo: As discussed, we cannot have an "effective x" as transactions commit out or order. So there's no way to say "Hey, I know about tx 10 so I must know tx 5".

addons/mail/static/src/model/field_version.js
```diff
@@ -0,0 +1,246 @@
...
+ * A versioned field value tied to a database snapshot. It tracks what this revision "saw"
+ * when this value was set, allowing the client to correctly order updates and ignore late
+ * arriving, stale data.
+ *
+ * @typedef {Object} FieldRevision
+ * @property {PgSnapshot} snapshot The transactional context defining what the current
+ * revision could "see" in the database.
+ * @property {boolean} isWrite Whether this revision modified the field.
+ */
+
```
alexkuhn, line 34: "Visible" is not clarified in docstring:
Image

IMO I would prefer to use wording "complete", but if we keep visible, at the very least there should be some comment to tell what "visible" means.

addons/mail/static/src/model/field_version.js
```diff
@@ -0,0 +1,266 @@
...
+/**
+ *
+ *                                             +-----------------------+
+ *                                         +-->|     PENDING COUNT (4) |
+ *                                         |   +-----------------------+
+ *          XMIN 5     +------------------->   ^            ^          ^  XMAX 12
+ *               ↓     |                       |            |          |  ↓
+ *               +------+--------+--------+------+--------+-------+-------+
+ *    √ √ √ √    | TX_5 | TX_6 √ | TX_7 √ | TX_8 | TX_9 √ | TX_10 | TX_11 |   × × × ×
```
seb-odoo: As we have × after xmax maybe we should also have it on the items inside to represent they are also pending?
And maybe add a legend to say × is pending and √ is finished

seb-odoo: Nice representation otherwise 👍️

seb-odoo: Also × looks a symbol more for cancelled than pending so if we can find something else it would be nice.

We might also just emoji instead of ascii, unless they mess up the alignment or something like that?

addons/mail/models/models.py
```diff
@@ -24,6 +24,13 @@ class Base(models.AbstractModel):
     # ORM
     # ------------------------------------------------------------
 
+    def write(self, vals):
+        result = super().write(vals)
+        if versioning := self.env.context.get("mail_store"):
+            for rec in self:
+                versioning["mark_fields_as_written"](rec, vals.keys())
```
rco-odoo: This doesn't catch all field updates: recomputed fields are not updates with a `write()`.

addons/mail/tools/discuss.py
```diff
@@ -38,6 +42,156 @@ def wrapper(self, *args, **kwargs):
...
+                self = self.with_context(mail_store=manager.context)
+            response = func(self, *args, **kwargs)
+            manager.do_send_bus_notifications(self.env)
+            manager.add_version_to_response(self.env, response)
+            if make_json_response:
+                req.make_json_response(response)
+            return response
+        return wrapper
+
+    return decorator(func) if func else decorator
```
kmagusiak: Provide just one API.
```suggestion
    return decorator
```

addons/mail/tools/discuss.py
```diff
@@ -38,6 +42,156 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+class StoreVersionInternal:
+    """Internal state used by the `@mail_store` decorator."""
+    def __init__(self):
+        # Maps model names to record IDs to the set of field names written.
+        self.written_fields_by_record = defaultdict(lambda: defaultdict(OrderedSet))
+        self.context = {
+            "manager": self,
```
kmagusiak: This sucks for the GC and since `context` is mutable, some behaviour is less explicit.
Could you make the `context` into a `@property`?

addons/mail/tools/discuss.py
```diff
@@ -38,6 +42,156 @@ def wrapper(self, *args, **kwargs):
...
+                ["write_date", "write_uid"],
+            )
+
+    def _get_formatted_version(self, env):
+        """Get the version metadata, used by the client to determine if an incoming
+        store insert is newer than what it already knows.
+        """
+        if self.snapshot_data is None:
+            env.flush_all()  # Ensure xact_id is properly assigned before querying the snapshot.
+            env.cr.execute("SELECT pg_current_snapshot(), pg_current_xact_id_if_assigned()")
```
kmagusiak: You should use SEQUENCE instead of internal functions.
`flush_all` does not guarantee that xact_id is assigned, you need a DML statement for that. If there is nothing to flush, you won't have an xact_id.

> used by the client to determine if an incoming store insert is newer than what it already knows

Does using snapshots mean that if I update something completely unrelated, the snapshot will change?
If so, you will detect no changes only on servers with very little traffic. Thus it seems to me that this kind of detection of changes is very limited.

tsm-odoo: > You should use SEQUENCE instead of internal functions.

Sequences don't work because visibility is directly linked to the isolation level. At least, I don't see how it could work. A sequence can be incremented, but that doesn't mean the changes have been committed nor does it tell wether they are visible to another transaction.

> If there is nothing to flush, you won’t have an xact_id

Indeed, xact_id is only expected for writes or deletes. If there is nothing to flush, then we didn't write or delete anything. The request only reads data. `xmin/xmax/xip_list` are enough to determine what was available to the snapshot.

> Does using snapshots mean that if I update something completely unrelated, the snapshot will change?

I don't get your point. When you receive a value for a field, you know both the current snapshot and the incoming one (for each field). You can decide whether to apply the update or not.

kmagusiak: Sequences are not linked to the transaction. If you start 2 transactions, you get different numbers in each of them.

I must have misunderstood the exact goal of getting this information. Forget it.

Julien00859 (approved):
Wearing my fw-python hat, in the name of the team, we approve the general idea of using postgres's transactions informations (tx-min/max/ip) and to forward them to the JS

Wearing my rd-http hat, the `mail_route` decorator is fine, it indeed makes for a better things that decorating a single controller 3 times (in the correct order). The implementation is fine, only that `make_json_response=True` could be changed by `if func.routing['type'] == 'http'`.

addons/mail/tools/discuss.py
```diff
@@ -38,6 +42,154 @@ def wrapper(self, *args, **kwargs):
...
+                req.update_context(mail_store=manager)
+            if isinstance(self, models.BaseModel):
+                self = self.with_context(mail_store=manager)
+            response = func(self, *args, **kwargs)
+            manager._add_version_to_response(self.env, response)
+            # Remove store from the context to stop enqueuing messages when calling `_sendone`.
+            if req:
+                req.update_context(mail_store=None)
+            if isinstance(self, models.BaseModel):
+                self = self.with_context(mail_store=None)
```
seb-odoo: We discussed having `if isinstance(self, models.BaseModel):` and `elif req:` but I don't remember if it was relevant to change it or not.

addons/mail/static/src/model/store_internal.js
```diff
@@ -32,6 +35,13 @@ export class StoreInternal extends RecordInternal {
     RHD_QUEUE = new Map(); // record-hard-deletes
     ERRORS = [];
     UPDATE = 0;
+    /**
+     * Current version context used in the current store insert operation.
+     *
+     * The version data is provided by the server. If no version is provided,
+     * each field falls back to its last known version.
+     */
+    currentInsertVersion = null;
```
alexkuhn, line 49: Would be nice to document shape of this attribute.

addons/mail/static/src/model/store.js
```diff
@@ -207,30 +208,44 @@ export class Store extends Record {
...
-                for (const vals of Array.isArray(data) ? data : [data]) {
-                    if (vals._DELETE) {
-                        delete vals._DELETE;
-                        recordsDataToDelete.push([modelName, vals]);
-                    } else {
-                        insertData.push(vals);
+        if ("__store_version__" in dataByModelName) {
+            const versionMeta = dataByModelName.__store_version__;
+            delete dataByModelName.__store_version__;
+            this._.currentInsertVersion = {
```
alexkuhn, line 216: Would also typedef `__store_version__`, just to give more confidence everything's good (or if some things are missing, this is made obvious with typedef).

addons/mail/static/tests/core/store_versioning.test.js
```diff
@@ -0,0 +1,480 @@
...
+                expected: { name: "WRITE" },
+                description: "Read from an older snapshot: do not override the write.",
+            },
+        ],
+    },
+];
+
+for (const testCase of SINGLE_FIELD_CASES) {
+    const testFn = testCase.only ? test.only : test;
+    testFn(`single store versionning - ${testCase.name}`, async () => {
```
alexkuhn: `versionning` => `versioning`

alexkuhn (approved):
LGTM for JS part, and quick glance at python part.

seb-odoo (approved):
@robodoo r+ delegate+

This will probably conflict with other PR in staging.

PR #246469, opened 2026-01-30
[IMP] website: lower sequence ir_actions_todo

ivkl-odoo:
The reason we lower the sequence of the ir_actions_todo is because only one can be executed after a module installation. That means that if another module has auto-install with website and also a todo, it will take precedence and skip the openning of the website_configurator.

In order to prevent this from happining (again), we lower the sequence to 8 so that it takes priority over the default which is 10. That way any new auto-install with a todo will not silently beak the flow from website.

Linked to: https://github.com/odoo/enterprise/pull/105919

---


lobaudoux (approved):
@robodoo r+

PR #232539, opened 2025-10-21
[REF] base: Update templates for new t-call syntax

Gorash:
The `t-call` syntax is updated to use the new semantic, which passes
values *as attributes/parameters* on the `<t>` (with `t-call`) tag itself,
instead of relying on nested `t-set` directive.

- Old (Deprecated): Used nested `<t t-set="var_name" t-value="x"/>` tags
inside the calling element to define variables.
- New: Variables are passed as attributes directly on the element where
the `t-call` is located (e.g., `<t t-call="module.template" var_name="x"/>`).

A warning is added to alert developers when using the old deprecated
syntax.

see: https://github.com/odoo/odoo/pull/197296
The script to automate the process has been added to: https://github.com/odoo/odoo/pull/235469

PR #246471, opened 2026-01-30
[FIX] website_forum: fix post comment button click

amdi-odoo:
Purpose
=======
Fix the comment button which doesn't react to clicks on post questions and post answers.

Specification
=============
Following odoo/odoo#232539 a script has been executed to convert every t-call to their new syntax.
Instead of being set inside the t-call with t-set, the variables are now passed as attributes directly on the t-call element.

This script has missed the post comment "_collapse_uid" variable as it was defined inside a <t t-if...> element.
Correctly rewriting the variable as a t-call attribute.

Task-5886055

---


julien-banken (commented):
Tested and approved 👍
Thanks for the fix @amdi-odoo

julien-banken (approved):

PR #246472, opened 2026-01-30
[FIX] website: use slugified url on "create page" in edit menu dialog

blse-odoo:
The "Create Page" button was added in edit menu dialog in commit 990b7c045bf27280c64433510d6e43fba5b3a4b0.

The button creates a page using the link in the menu for the url of the page, but the actual page creation may use a different url (as it slugifies it).

This commit uses the url returned by the server on page creation to update the url of the menu, and correctly redirect to the new page.

Steps to reproduce:
- In edit menu > menu item, create a menu with url `/abc,xyz`
- In edit menu, click "Create Page"
- The page is created with a url that is slugified
- Bug: but the menu does not use the slugified new url, and the url to which we redirect is not that one either

task-5895401

### DIFF
--- a/addons/website/static/src/components/dialog/edit_menu.js
+++ b/addons/website/static/src/components/dialog/edit_menu.js
@@ -405,8 +405,11 @@ export class EditMenuDialog extends Component {
         let url = menu.fields["url"];
         url = url.startsWith("/") ? url : "/" + url;
         this.dialogs.add(AddPageDialog, {
-            onAddPage: () => {
-                this.onClickSave(false, url);
+            onAddPage: ({ createdUrl }) => {
+                if (createdUrl) {
+                    menu.fields["url"] = createdUrl;
+                }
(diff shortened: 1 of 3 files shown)


addons/website/static/src/components/dialog/edit_menu.js
```diff
@@ -394,8 +394,11 @@ export class EditMenuDialog extends Component {
         let url = menu.fields["url"];
         url = url.startsWith("/") ? url : "/" + url;
         this.dialogs.add(AddPageDialog, {
-            onAddPage: () => {
-                this.onClickSave(false, url);
+            onAddPage: ({ createdUrl }) => {
+                if (createdUrl) {
+                    menu.fields["url"] = createdUrl;
+                }
+                this.onClickSave(false, createdUrl ?? url);
```
FrancoisGe, line 412: Could you add a test ?

PR #246475, opened 2026-01-30
[FIX] account: UserError in linking of bills at import

gawa-odoo:
Backport of https://github.com/odoo/odoo/commit/71de6902d73bec32f93e3be8f4de4ec89e9dc987

A UserError there can cause duplicates in peppol fetching. It should never block an ack.
We can leave them unlinked
We also do the same for any unexpected error that occurs after we've created the move.

opw-5492092

---


clbr-odoo (approved):
@robodoo r+

PR #246476, opened 2026-01-30
[FIX] point_of_sale: Fix indexedDB race condition on pos init

macocianradu:
There were 2 issues when `serverDateTime` was lower than `lastConfigChange` which triggered a reset on the indexedDB. The first issue was that we would try to await dbInstance.deleteDatabase request. But it returns a request object and then executes the delete asynchronously. This would cause a race condition on the init where we would trigger the init of the indexedDB at the same time as we were trying to delete it and it would hang for ~10 seconds before finally launching the POS.

The second more important issue is that after the reset there is a `localData = []`. This would cause the line in `synchronizeServerDataInIndexedDB` `JSON.parse(JSON.stringify(serverData));` to return an empty array so no new models would get created in the indexedDB and the POS would launch with an empty indexedDB.

This commit changes the `indexed_db.reset()` method to return a promise and awaits it before reinitialising the indexedDB. And resets the `localData` to `{}` which fixes the `synchronizeServerDataInIndexedDB`

---

### DIFF
--- a/addons/point_of_sale/static/src/app/models/utils/indexed_db.js
+++ b/addons/point_of_sale/static/src/app/models/utils/indexed_db.js
@@ -220,11 +220,51 @@ export default class IndexedDB {
     }
 
     reset() {
-        if (!this.dbInstance) {
-            return false;
-        }
-        this.dbInstance.deleteDatabase(this.dbName);
-        return true;
+        return new Promise((resolve) => {
+            if (this.db) {
+                this.db.close();
+            }
+
+            if (!this.dbInstance) {
+                return resolve(true);
+            }
+
+            const timeout = setTimeout(() => {
+                logPosMessage(
+                    "IndexedDB",
+                    "reset",
+                    "Timeout: Database reset took too long",
+                    CONSOLE_COLOR
+                );
+                resolve(false);
+            }, 3000);
+
+            const request = this.dbInstance.deleteDatabase(this.dbName);
+
+            request.onsuccess = () => {
+                logPosMessage("IndexedDB", "reset", "Database deleted successfully", CONSOLE_COLOR);
+                this.db = null;
+                clearTimeout(timeout);
+                resolve(true);
+            };
+
+            request.onerror = (event) => {
+                logPosMessage(
+                    "IndexedDB",
+                    "reset",
+                    `Error deleting DB: ${event.target.error}`,
+                    CONSOLE_COLOR
+                );
+                clearTimeout(timeout);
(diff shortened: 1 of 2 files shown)


addons/point_of_sale/static/src/app/models/utils/indexed_db.js
```diff
@@ -220,11 +220,51 @@ export default class IndexedDB {
...
+                resolve(false);
+            }, 3000);
+
+            const request = this.dbInstance.deleteDatabase(this.dbName);
+
+            request.onsuccess = () => {
+                logPosMessage("IndexedDB", "reset", "Database deleted successfully", CONSOLE_COLOR);
+                this.db = null;
+                clearTimeout(timeout);
+                resolve();
```
davidmonnom:
```suggestion
                resolve(true);
```

addons/point_of_sale/static/src/app/models/utils/indexed_db.js
```diff
@@ -220,11 +220,51 @@ export default class IndexedDB {
...
+
+            request.onerror = (event) => {
+                logPosMessage(
+                    "IndexedDB",
+                    "reset",
+                    `Error deleting DB: ${event.target.error}`,
+                    CONSOLE_COLOR
+                );
+                clearTimeout(timeout);
+                resolve(true);
```
davidmonnom:
```suggestion
                resolve(false);
```

addons/point_of_sale/static/src/app/models/utils/indexed_db.js
```diff
@@ -220,11 +220,51 @@ export default class IndexedDB {
...
+            };
+
+            request.onblocked = () => {
+                logPosMessage("IndexedDB", "reset", "Blocked deleting DB", CONSOLE_COLOR);
+                clearTimeout(timeout);
+                resolve(true);
```
davidmonnom:
```suggestion
                resolve(false);
```

addons/point_of_sale/static/src/app/services/data_service.js
```diff
@@ -330,7 +330,7 @@ export class PosData extends Reactive {
                     }
                 }
 
-                this.synchronizeServerDataInIndexedDB(localData);
+                await this.synchronizeServerDataInIndexedDB(localData);
```
davidmonnom: Do not await it here, if it blocks, the PoS will be stuck.

davidmonnom (approved):
@robodoo r/+

PR #246478, opened 2026-01-30
[FIX] hr_attendance : fix wrong reverse link between attendance and o…

mepe-odoo:
…vertime

STEP TO REPRODUCE: (hr_Work_entry need to be installed to be tested fromt he frontend) ------------------
1- Create an overtime rule with this configuration :
    quantity rule differs from hours defined in the contract
2- Create two attendances on the same day (should be a worked day and should not be today)
    one from 1AM-2PM
    second from 2PM-undefined
3- Approve the overtime of the first attendance

REASON:
------

If hr_work_entry is installed a traceback will be raised because the link between overtime to attendance is not correctly done Attendance -> overtime (only the first attendance have a link with overtime) overtime, attendance -> the two will be returned

Current behavior before PR:

---

### DIFF
--- a/addons/hr_attendance/models/hr_attendance_overtime.py
+++ b/addons/hr_attendance/models/hr_attendance_overtime.py
@@ -90,7 +90,7 @@ class HrAttendanceOvertimeLine(models.Model):
 
     def _linked_attendances(self):
         return self.env['hr.attendance'].search([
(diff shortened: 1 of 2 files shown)


romaincarlier4 (approved):
Thanks for the fix 
@robodoo r+

PR #246480, opened 2026-01-30
[FIX] account_peppol: avoid timeout in doc retrieval

LouisGobert:
Retrieving documents from the Peppol proxy could trigger a timeout for large databases (e.g. thousands of products or invoices with many thousands of lines), causing the cron to fail and preventing commits and ACK from being sent to the IAP server.

To mitigate this, the cron runtime is limited to
(limit_time_real or limit_time_real_cron) minus 60 seconds, leaving a 1-minute safety margin before the actual timeout.

opw-5868249

### DIFF
--- a/addons/account_peppol/models/account_edi_proxy_user.py
+++ b/addons/account_peppol/models/account_edi_proxy_user.py
@@ -2,6 +2,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 import logging
+import time
 
 from odoo import _, api, fields, models, tools
 from odoo.addons.account_edi_proxy_client.models.account_edi_proxy_user import AccountEdiProxyError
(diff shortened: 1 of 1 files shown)


addons/account_peppol/models/account_edi_proxy_user.py
```diff
@@ -131,6 +131,12 @@ def _peppol_get_new_documents(self):
                 'errors': False,
             }
         }
+        start_time_timestamp = fields.Datetime.now().timestamp()
```
clbr-odoo: No need for the field boilerplate, let's use `time.monotonic()` which is more appropriate here.

addons/account_peppol/models/account_edi_proxy_user.py
```diff
@@ -131,6 +131,12 @@ def _peppol_get_new_documents(self):
...
+        limit_time = tools.config['limit_time_real']
+        if tools.config.get('limit_time_real_cron', -1) > 0:
+            limit_time = tools.config['limit_time_real_cron']
+        limit_time -= 60  # Margin of 1 minute to avoid timeouts
```
clbr-odoo: I don't think that's a big enough margin... Looking at other example in Odoo, we limited between 3 to 5 minutes. 
Example: `limit_time = cron_limit_time if 0 < cron_limit_time < 180 else 180`
Also it will crash anyway if you put a ridiculous amount because there is some kind of hard limit, you can't increase this number as much as you wish.
We can probably try with 5 minutes, better have multiple retrigger than crashing ? 🤔

LouisGobert:
Closing for now, as other performance improvements appear to have already optimized the document retrieval time, and I haven't seen any recent reports of this issue.

PR #246482, opened 2026-01-30
[FIX] website_sale: prevent error when opening shop when variants are deleted

reka-odoo:
When all variants of a product template are deleted, opening the Shop page on the website raises a traceback.

Steps to reproduce the error:
- Install ``website_sale`` module
- Go to settings > Enable variants
- Create a product template > In Attributes & Variants Tab > add one attribute with two values > Save
- Click on the variants smart button > select all > delete
- Go to Website > Shop

Traceback:
```py
AssertionError: Invalid falsy real id
```

https://github.com/odoo/odoo/blob/b6598c11a3580cfc6ff1ffeb67b0d388c55383d1/addons/website_sale/controllers/main.py#L456 After [commit](https://github.com/odoo/odoo/commit/4290724a4c8c57fba4f4d3d688d38f65dadcc38f), falsy IDs are no longer allowed in ``browse()``.
When all variants of a product template are deleted, ``product._get_first_possible_variant_id()`` returns False.
This falsy value is later passed to ``browse()``,
which triggers the above traceback.

sentry-7201563878

---

### DIFF
--- a/addons/website_sale/controllers/main.py
+++ b/addons/website_sale/controllers/main.py
@@ -454,9 +454,12 @@ class WebsiteSale(payment_portal.PaymentPortal):
         products.fetch()
 
         # map each product to its variant, and prefetch the variants
-        variants = request.env['product.product'].sudo().browse(product._get_first_possible_variant_id() for product in products)
+        Product = request.env['product.product']
+        product_variant_ids = [product._get_first_possible_variant_id() for product in products]
+        variants = Product.sudo().browse(vid for vid in product_variant_ids if vid)
         variants.fetch()
-        product_variants = dict(zip(products, variants))
+        variant_by_id = {v.id: v for v in variants}
+        product_variants = dict(zip(products, (variant_by_id.get(vid, Product) for vid in product_variant_ids)))
 
         ProductAttribute = request.env['product.attribute']
         if products:
--- a/addons/website_sale/tests/test_product_attribute_value_config.py
+++ b/addons/website_sale/tests/test_product_attribute_value_config.py
@@ -1,7 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo.fields import Command
-from odoo.tests import tagged
+from odoo.tests import HttpCase, tagged
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 from odoo.addons.product.tests.test_product_attribute_value_config import (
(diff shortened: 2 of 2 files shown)


maan-odoo (approved):

addons/website_sale/controllers/main.py
```diff
@@ -453,9 +453,22 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
+        )
         variants.fetch()
-        product_variants = dict(zip(products, variants))
+        variant_by_id = {v.id: v for v in variants}
+        product_variants = dict(
+            zip(
+                products,
+                (variant_by_id.get(vid, Product) for vid in product_variant_ids)
+            )
+        )
```
ltinel: It's a bit annoying that this code is so much longer for a simple fix, but I'm not sure we can do much better (while keeping a single `browse` call) 🥲

addons/website_sale/controllers/main.py
```diff
@@ -453,9 +453,22 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
         # map each product to its variant, and prefetch the variants
-        variants = request.env['product.product'].sudo().browse(product._get_first_possible_variant_id() for product in products)
+        Product = request.env['product.product']
+        product_variant_ids = [
+            product._get_first_possible_variant_id()
+            for product in products
+        ]
+        variants = Product.sudo().browse(
+            vid for vid in product_variant_ids if vid
+        )
```
ltinel: Nit: could you put both of these instructions on single lines? I believe they'll still be below 100 chars wide (which is our team's max line length) 🙂

```suggestion
        product_variant_ids = [product._get_first_possible_variant_id() for product in products]
        variants = Product.sudo().browse(vid for vid in product_variant_ids if vid)
```

addons/website_sale/controllers/main.py
```diff
@@ -453,9 +453,22 @@ def shop(self, page=0, category=None, search='', min_price=0.0, max_price=0.0, t
...
         variants.fetch()
-        product_variants = dict(zip(products, variants))
+        variant_by_id = {v.id: v for v in variants}
+        product_variants = dict(
+            zip(
+                products,
+                (variant_by_id.get(vid, Product) for vid in product_variant_ids)
+            )
```
ltinel: Nit:
```suggestion
            zip(products, (variant_by_id.get(vid, Product) for vid in product_variant_ids))
```

ltinel (approved):
Thanks for the fix, LGTM 🙂 
@robodoo delegate+

PR #246484, opened 2026-01-30
[FIX] hr,hr_attendance: company-specific attendance setting and reliable module installation

ramezlahzy:
Steps to reproduce
1. Install hr_attendance and create employees in a company (e.g., US). The attendance pill appears as expected.Go to Employees > Configuration > Settings: 'Employees' and you will find 'Based on attendance' checked
2. Install l10n_be_hr_payroll, switch to the Belgian company, and create employees. The attendance pill is missing, even if 'Based on attendance' is checked in settings.

 Bug cause
- The field `module_hr_attendance` was used to control the attendance setting, but it only indicates if the hr_attendance module is installed, not the per-company setting.
- `hr_presence_control_attendance` is company-dependent, but `module_hr_attendance` is global. This caused the UI to show the setting as enabled in all companies if the module was installed, regardless of the actual per-company setting.
- The previous workaround set `hr_presence_control_attendance` to True for all companies at module install, but this fails if localizations (and thus new companies) are added after hr_attendance is installed.
 
 Solution
- The settings UI now displays and controls `hr_presence_control_attendance`, which is company-specific, instead of `module_hr_attendance`.
- In the uninstall hook, when hr_attendance is uninstalled, `hr_presence_control_attendance` is set to False for all companies.
- Override the write method in setting to check if `hr_presence_control_attendance` checked then install attendance module if doesn't installed

This ensures the attendance pill and settings are correctly shown per company, regardless of when localizations or companies are added.

Task Id: 5125126

### DIFF
--- a/addons/hr/models/res_company.py
+++ b/addons/hr/models/res_company.py
@@ -12,6 +12,10 @@ class ResCompany(models.Model):
     hr_presence_control_login = fields.Boolean(string="Based on user status in system", default=True)
     hr_presence_control_email = fields.Boolean(string="Based on number of emails sent")
     hr_presence_control_ip = fields.Boolean(string="Based on IP Address")
-    hr_presence_control_attendance = fields.Boolean(string="Based on attendances")
+    hr_presence_control_attendance = fields.Boolean(string="Based on attendances", default=lambda self: self._default_hr_presence_control_attendance())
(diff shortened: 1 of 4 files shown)


amah-odoo (approved):
LGTM 💯

PR #246488, opened 2026-01-30
[FIX] mail: handle x2many for inherits inverse delete cascade

reth-odoo:
Usually (always?) the "inverse" for an inherited field will be a one2many. Even if in practice it is a one2one.

The cascade deletion of the inverse should thus handle RecordList instead of assuming they're always single records.

related: 478647c4526f42a2455a599555748c844a6f20cf

task-5013894

### DIFF
--- a/addons/mail/static/tests/core/record.test.js
+++ b/addons/mail/static/tests/core/record.test.js
@@ -256,6 +256,43 @@ test("Assign & Delete on fields with inverses", async () => {
     expect(thread.messages).toBeEmpty();
 });
 
+test("Assign & Delete on _inherits fields", async () => {
+    (class Thread extends Record {
+        static id = "name";
+        name;
+        channel = fields.One("Channel", { inverse: "thread" });
+        pluriChannels = fields.Many("PluriChannel", { inverse: "thread" });
+    }).register(localRegistry);
+    (class Channel extends Record {
+        static id = "id";
+        static _inherits = { Thread: "thread" };
+        id;
+        thread = fields.One("Thread", { inverse: "channel" });
+    }).register(localRegistry);
+    (class PluriChannel extends Record {
+        static id = "id";
+        static _inherits = { Thread: "thread" };
+        id;
+        thread = fields.One("Thread", { inverse: "pluriChannels" });
+    }).register(localRegistry);
+    const store = await start();
+    const thread = store.Thread.insert("General");
+    const channel = store.Channel.insert({ id: 1, thread: thread });
+    const [pluriChannel1, pluriChannel2] = store.PluriChannel.insert([
+        { id: 1, thread },
+        { id: 2, thread },
+    ]);
+    expectRecord(channel).toEqual(thread.channel);
+    expectRecord(pluriChannel1).toBeIn(thread.pluriChannels);
+    expectRecord(pluriChannel2).toBeIn(thread.pluriChannels);
+    thread.delete();
+    expect(thread.pluriChannels).toBeEmpty();
+    expect(pluriChannel1.exists()).toBe(false);
+    expect(pluriChannel2.exists()).toBe(false);
+    expect(Boolean(thread.channel)).toBe(false);
+    expect(channel.exists()).toBe(false);
+});
+
 test("onAdd/onDelete hooks on relational with inverse", async () => {
     let logs = [];
     (class Thread extends Record {
--- a/addons/mail/static/src/model/record.js
+++ b/addons/mail/static/src/model/record.js
@@ -328,9 +328,16 @@ export class Record {
         return store.MAKE_UPDATE(function recordDelete() {
             // delete records inheriting the current record before deleting the current record
             for (const fieldName of record.Model._.inheritsInverseFields) {
-                const dependentRecordProxy = record._proxyInternal[fieldName];
-                if (dependentRecordProxy) {
-                    store._.ADD_QUEUE("delete", toRaw(dependentRecordProxy)._raw);
+                if (record.Model._.fieldsMany.get(fieldName)) {
+                    const dependentRecordListProxy = record._proxyInternal[fieldName];
+                    for (const dependentRecordProxy of dependentRecordListProxy) {
+                        store._.ADD_QUEUE("delete", toRaw(dependentRecordProxy)._raw);
+                    }
+                } else {
+                    const dependentRecordProxy = record._proxyInternal[fieldName];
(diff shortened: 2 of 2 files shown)


addons/mail/static/tests/core/record.test.js
```diff
@@ -216,10 +223,23 @@ test("Assign & Delete on fields with inverses", async () => {
...
         threads = fields.Many("Thread");
     }).register(localRegistry);
+    (class Channel extends Record {
+        static _inherits = { Thread: "thread" };
+        thread = fields.One("Thread", { inverse: "channel" });
+    }).register(localRegistry);
+    (class PluriChannel extends Record {
+        static _inherits = { Thread: "thread" };
+        thread = fields.One("Thread", { inverse: "pluriChannels" });
+    }).register(localRegistry);
```
reth-odoo: Couldn't really come up with an example that fully made sense here but for the real-life example where we indeed use it as a o2m (as opposed to it being semantically a o2o): ResParnter.user_ids

addons/mail/static/tests/core/record.test.js
```diff
@@ -21,6 +21,10 @@ const expectRecord = (record, not = false) => {
         expect(record?.in(reclist)).toBe(!not);
     };
 
+    const toBeDeleted = (isDeleted) => {
+        expect(record?.[IS_DELETED_SYM]).toBe(!not);
+    };
+
```
alexkuhn: I'd prefer to use `expect(record.exists()).toBe(true/false)` rather than define another `toBe()` flavour.

addons/mail/static/tests/core/record.test.js
```diff
@@ -216,10 +223,23 @@ test("Assign & Delete on fields with inverses", async () => {
...
+    }).register(localRegistry);
     const store = await start();
     const thread = store.Thread.insert("General");
     const [john, marc] = store.Member.insert(["John", "Marc"]);
     const [hello, world] = store.Message.insert(["hello", "world"]);
+    const channel = store.Channel.insert({ thread: thread });
+    const [pluriChannel1, pluriChannel2] = store.PluriChannel.insert([
+        { thread: thread },
+        { thread: thread },
+    ]);
```
alexkuhn: Records of a model are identified with `static id`. The `static id` tells how a record is identified for the given record.

For example:
```js
class Message extends Record {
   static id = "content";
}

class Chat extends Record {
   static id = "id";
}

class Thread extends Record {
   static id = AND("model", "id");
}

class DiscussApp extends Record {
   // no static id
}
```

Insert of message either create message if there's no similar content, otherwise if it matches an existing message with this content then it updates it.

- Same idea for `Chat`'s `id`.
- `Thread` here is when `model` and `id` both matches, so `res.partner,10` identifies a record and `discuss.channel,10` is another thread. If we insert on `res.partner,10` again, this updates the `res.partner,10` thread record.
- `DiscussApp` here has no `static id` defines. In 19.1, this is implicitly `static id = undefined` and therefore means this is a singleton record. 🚨 Just recently in `master`, this becomes implicitly `static id = "id"`: https://github.com/odoo/odoo/pull/246122 as most models map to server models which are identified with `"id"` field.

So `DiscussApp.insert()` creates new record if it doesn't exist, and any other `DiscussApp.insert()` necessarily updates the existing record.

----

All this to say, the `PluriChannel` model defined here has no explicit `static id` so this essentially works as a singleton record. The 2 inserts don't make 2 different records. I'm pretty sure this is not intentional for `pluriChannel1` to be the same as `pluriChannel2`.

Can be fixed with `static id = "someIdIdFieldName"` on the model and passing unique value to `someIdIdFieldName` when `PluriChannel.insert()`, like:
```js
class PluriChannel extends Record {
   static id = "id";
}
store.PluriChannel.insert([
   { thread, id: 1 },
   { thread, id: 2 },
]);
```

(Note that when `static id` is a single field, there's syntatic sugar to pass a primitive value in `.insert(value)` that is equivalent to `.insert({ [someIdFieldName]: value })`)

addons/mail/static/tests/core/record.test.js
alexkuhn: We don't judge (or we try to not judge) the functional relevancy of `record.test.js` tests. What matters is model definition to be sound.

I'd prefer to add another test for the "_inherits" version, because this test becomes a bit too long for my taste.

Could just use the same name and add `(_inherits)` as suffix, so we have coverage of "easy" cases as this current test and have a "_inherits" variant.

alexkuhn (commented):
Fix LGTM otherwise :)

addons/mail/static/tests/core/record.test.js
alexkuhn: Fun fact: could insert these PluriChannels with this simpler syntax:

```js
thread.pluriChannels = [1, 2];
```

This will `PluriChannel.insert([1, 2])` on the relation. This uses the syntatic sugar I hint in comment above, and also the `{ thread }` is implicitly provided thanks to an insert on the relational field (`Thread.pluriChannels` and `PluriChannel.thread` are linked together thanks to `inverse`, so that's why this is implicit in the insert).

All these explanations look complicated, but in practice this is much easier to use. Of course as long as we understand what ".insert()" means in relation to `static id`.

reth-odoo: @alexkuhn 
Oh I see, indeed I remember seeing something about adding a default static id recently. Thanks for the crash course ^^

Updated and moved to a new, similar, test

addons/mail/static/tests/core/record.test.js
```diff
@@ -256,6 +256,43 @@ test("Assign & Delete on fields with inverses", async () => {
...
+        id;
+        thread = fields.One("Thread", { inverse: "pluriChannels" });
+    }).register(localRegistry);
+    const store = await start();
+    const thread = store.Thread.insert("General");
+    const channel = store.Channel.insert({ id: 1, thread: thread });
+    const [pluriChannel1, pluriChannel2] = store.PluriChannel.insert([
+        { id: 1, thread: thread },
+        { id: 2, thread: thread },
+    ]);
```
alexkuhn, line 284:
```suggestion
    const channel = store.Channel.insert({ id: 1, thread });
    const [pluriChannel1, pluriChannel2] = store.PluriChannel.insert([
        { id: 1, thread },
        { id: 2, thread },
    ]);
```

alexkuhn (approved):
@robodoo delegate+

PR #246492, opened 2026-01-30
[IMP] mail: search messages panel improvements

Atovange:
This commit:
1. Adds the thread name to the input placeholder.
2. Capitalizes the search message panel title.

task-5873831

### DIFF
--- a/addons/mail/static/src/core/common/search_message_input.js
+++ b/addons/mail/static/src/core/common/search_message_input.js
@@ -98,4 +98,8 @@ export class SearchMessageInput extends Component {
             },
         ];
     }
+
+    get inputPlaceholder() {
+        return _t("Search %(threadName)s", { threadName: this.props.thread.displayName });
+    }
 }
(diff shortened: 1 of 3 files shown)


seb-odoo:
I don't see why we should hide reactions in search results, they are part of the message and indicate relevant information about it.

addons/mail/static/src/core/common/search_message_input.js
```diff
@@ -98,4 +98,8 @@ export class SearchMessageInput extends Component {
             },
         ];
     }
+
+    get inputPlaceholder() {
+        return _t("Search %(channelName)s", { channelName: this.props.thread.displayName });
```
seb-odoo: this is thread generic, the placeholder should not contain channel in it

alexkuhn:
I would make a different PR for disabling message reactions in search panel, as I think we'd want to debate this change, whereas the rest is good and could be merged already.

alexkuhn (approved):
@robodoo delegate+ r+

PR #246495, opened 2026-01-30
[FIX] account_peppol: fix the display of the itsme button in Peppol

clbr-odoo:
Add some space to the right of the logo.
Fix the colors in light mode.

task-none

**Before:**
image

**After:**
image
image


gawa-odoo (commented):
@robodoo r+

PR #246505, opened 2026-01-30
[FIX] mass_mailing: handle reviews wall crash

thjo-odoo:
This commit fixes an issue with the s_reviews_wall snippet which makes the builder crash if it selected.

The issue was that the BorderConfigurator used on the option's template wasn't specified in the option's components. Leading to a crash that renders the builder inoperable.

task-5380467

---


abd-msyukyu-odoo:
@robodoo r+

PR #246508, opened 2026-01-30
[FIX] html_editor: fix image size getting reduced

adsa-odoo:

Steps to reproduce:

- In website, drag and drop a `text - image` snippet.
- Click on image, click on Transform button.
- Try to rotate the image.
- You will notice that the size of the image is reduced a bit.

This issue happens because in `image_transformation.js`, `convertPixelWidthToPercentage` converts image width from `px` to percentage. In this case image's parentElement has padding, causing reduction in image's size.

This PR ensures that `paddingLeft` and `paddingRight` of image's parentElement is ignored from calculation so that image size doesn't get changed.

task-5884679

---

### DIFF
--- a/addons/html_builder/static/src/plugins/image/image_transform_option_plugin.js
+++ b/addons/html_builder/static/src/plugins/image/image_transform_option_plugin.js
@@ -26,7 +26,6 @@ class TransformImageAction extends BuilderAction {
         params: { isImageTransformationOpen, closeImageTransformation },
     }) {
         if (!isImageTransformationOpen()) {
-            let changed = false;
             const deferredTillMounted = new Deferred();
             registry.category("main_components").add("ImageTransformation", {
                 Component: ImageTransformation,
@@ -35,14 +34,8 @@ class TransformImageAction extends BuilderAction {
                     document: this.document,
                     editable: this.editable,
                     destroy: () => closeImageTransformation(),
-                    onChange: () => {
-                        changed = true;
-                    },
                     onApply: () => {
-                        if (changed) {
-                            changed = false;
-                            this.dependencies.history.addStep();
-                        }
+                        this.dependencies.history.addStep();
                     },
                     onComponentMounted: () => {
                         deferredTillMounted.resolve();
--- a/addons/html_builder/static/tests/editor.test.js
+++ b/addons/html_builder/static/tests/editor.test.js
@@ -1,6 +1,7 @@
-import { setupHTMLBuilder } from "@html_builder/../tests/helpers";
+import { dummyBase64Img, setupHTMLBuilder } from "@html_builder/../tests/helpers";
 import { setSelection } from "@html_editor/../tests/_helpers/selection";
 import { expandToolbar } from "@html_editor/../tests/_helpers/toolbar";
+import { expectElementCount } from "@html_editor/../tests/_helpers/ui_expectations";
 import { insertText, pasteHtml } from "@html_editor/../tests/_helpers/user_actions";
 import { FontPlugin } from "@html_editor/main/font/font_plugin";
 import { isTextNode } from "@html_editor/utils/dom_info";
@@ -14,7 +15,7 @@ import {
     waitForNone,
 } from "@odoo/hoot-dom";
 import { animationFrame } from "@odoo/hoot-mock";
-import { contains, patchWithCleanup } from "@web/../tests/web_test_helpers";
+import { contains, onRpc, patchWithCleanup } from "@web/../tests/web_test_helpers";
 
 describe.current.tags("desktop");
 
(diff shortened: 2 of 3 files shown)


addons/html_editor/static/src/main/media/image_transformation.js
```diff
@@ -215,14 +215,27 @@ export class ImageTransformation extends Component {
             this.image.parentElement,
             (node) => node.offsetWidth > currentPixelWidth
         );
-        const widthPercent = (currentPixelWidth / container.offsetWidth) * 100;
+        const containerStyles = window.getComputedStyle(container);
+        const widthPercent =
+            (currentPixelWidth /
+                (this.image.parentElement.offsetWidth -
```
deso-odoo: Shouldn't this be `container.offsetWidth`? 🤔

addons/html_editor/static/src/main/media/image_transformation.js
```diff
@@ -215,14 +215,27 @@ export class ImageTransformation extends Component {
...
     }
 
     mouseUp() {
+        if (!this.transfo.active) {
+            return;
+        }
         this.isCurrentlyTransforming = false;
+        if (this.transfo.active.type.length === 2) {
+            this.convertPixelWidthToPercentage();
+        }
```
deso-odoo, line 239: I don't get this change. What is `.length === 2` check? A comment would be helpful.

addons/html_editor/static/src/main/media/image_transformation.js
```diff
@@ -215,14 +215,27 @@ export class ImageTransformation extends Component {
...
+                    parseFloat(containerStyles.borderLeftWidth) -
+                    parseFloat(containerStyles.borderRightWidth))) *
+            100;
         this.image.style.width = Math.min(100, widthPercent).toFixed(2) + "%";
     }
 
     mouseUp() {
+        if (!this.transfo.active) {
+            return;
+        }
```
deso-odoo, line 233: `mouseUp` is on a `useExternalListener`, and if transfo is not active would mouseUp even be called? I don't get this change.

adsa-odoo: In website, tnasfo is not destroyed when you click on transformation button twice. In this case  `this.transfo.active` could be undefined if no transformation option is clicked.

addons/html_editor/static/src/main/media/image_transformation.js
adsa-odoo: Image width should be converted to percentage only when its width gets changed rather than for all the cases. I've added a comment.

addons/html_builder/static/tests/editor.test.js
```diff
@@ -154,6 +154,45 @@ test("should apply default table classes on paste", async () => {
...
+            <div class="o_not_editable">
+                <img style="width: 20%;" class="o_editable_media" src="${dummyBase64Img}"/>
+            </div>
+        </section>
+    `);
+    const editor = getEditor();
+    const img = editor.editable.querySelector("img");
+    await click(img);
+    await animationFrame();
+    expect("[data-action-id=transformImage]").toHaveCount(1);
```
deso-odoo: expectElementCount

addons/html_builder/static/tests/editor.test.js
```diff
@@ -154,6 +154,45 @@ test("should apply default table classes on paste", async () => {
...
+    await click("[data-action-id=transformImage]");
+    await animationFrame();
+    expect(".transfo-container").toHaveCount(1);
```
deso-odoo: expectElementCount

addons/html_builder/static/tests/editor.test.js
adsa-odoo: I've made the changes. However, a call to `animationFrame` is needed before clicking the `[data-action-id=transformImage]` button. The reason is that `selectionchange` event is called when image options are rendered and there is a `selectionchange` handler in `useImageTransform` hook which closes the the transformation immediately.

addons/html_editor/static/src/main/media/image_transformation.js
```diff
@@ -215,14 +215,29 @@ export class ImageTransformation extends Component {
             this.image.parentElement,
             (node) => node.offsetWidth > currentPixelWidth
         );
-        const widthPercent = (currentPixelWidth / container.offsetWidth) * 100;
```
dmo-odoo, line 218: @walidsahli If I'm not mistaken, you changed this multiple times. I'm a bit wary that the change here might introduce some regression with respect to previous bugfixes that you might have done. Could you have a look please?

addons/html_builder/static/src/plugins/image/image_transform_option_plugin.js
```diff
@@ -35,14 +34,8 @@ class TransformImageAction extends BuilderAction {
...
                     destroy: () => closeImageTransformation(),
-                    onChange: () => {
-                        changed = true;
-                    },
                     onApply: () => {
-                        if (changed) {
-                            changed = false;
-                            this.dependencies.history.addStep();
-                        }
+                        this.dependencies.history.addStep();
```
dmo-odoo, line 38: @bso-odoo The suggested changes here seem to revert your changes from https://github.com/odoo/odoo/pull/223311/changes/06a9fac776ca24a37466d686d960350bfee95dee. I fear there might be a regression. Could you have a look please?

addons/html_editor/static/src/main/media/image_transformation.js
walidsahli: @dmo-odoo this should be fine as a change, i didn't consider paddings/borders.

addons/html_builder/static/src/plugins/image/image_transform_option_plugin.js
bso-odoo: Behavior is still correct in both Chrome and Firefox.
`ImageTransformation` now only calls `onApply` on mouse up, making my old change redundant.

dmo-odoo:
@adsa-odoo Thanks for the fix!
@walidsahli @bso-odoo Thank you for your input!
@robodoo rebase-ff r+

PR #246515, opened 2026-01-30
[IMP] mrp: display and track end date on done MO

stan-odoo:
For traceability purposes, users need to know when a specific manufacturing
order was completed. In addition, exposing the manufacturing order end date can
support inventory backdating use cases, where the effective completion date of a
production order matters, helping users better understand and rely on the
recorded completion timestamp during operational reviews.

This change displays the manufacturing order end date on the production form
once the order is completed, while keeping it hidden during earlier stages. Any
modification to the end date after the MO is marked as done is logged in the
chatter to maintain a complete audit trail.

Task - 5490220

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -1024,7 +1024,17 @@ class MrpProduction(models.Model):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        initial_date_finished_mo_values = {}
+        if 'date_finished' in vals and 'state' not in vals:
+            date_finished_field_definition = self.fields_get(['date_finished'])
+            for production in self.filtered(lambda p: p.state == 'done'):
+                initial_date_finished_mo_values[production] = {'date_finished': production.date_finished}
+
         res = super(MrpProduction, self).write(vals)
+
+        for production, initial_value in initial_date_finished_mo_values.items():
+            if tracking_values := production._mail_track(date_finished_field_definition, initial_value)[1]:
+                production._message_log(tracking_value_ids=tracking_values)
 
         for production in self:
             if 'date_start' in vals and not self.env.context.get('force_date', False):
--- a/addons/mrp/views/mrp_production_views.xml
+++ b/addons/mrp/views/mrp_production_views.xml
@@ -245,7 +245,6 @@
                 </header>
                 <sheet>
                     <field name="reservation_state" invisible="1"/>
-                    <field name="date_finished" invisible="1"/>
                     <field name="is_locked" invisible="1"/>
                     <field name="qty_produced" invisible="1"/>
                     <field name="unreserve_visible" invisible="1"/>
@@ -390,6 +389,7 @@
                                 <field name="delay_alert_date" invisible="1"/>
                                 <field nolabel="1" name="json_popover" widget="stock_rescheduling_popover" invisible="not json_popover"/>
                             </div>
+                            <field name="date_finished" string="End Date" readonly="is_locked" invisible="state != 'done'" help="Date at which you finished production."/>
                             <field name="components_availability_state" invisible="1"/>
                             <field name="components_availability" invisible="state not in ['confirmed', 'progress']"
                                 decoration-success="reservation_state == 'assigned' or components_availability_state == 'available'"


dhha-odoo (approved):

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,17 @@ def write(self, vals):
...
         res = super(MrpProduction, self).write(vals)
-
+        if date_finished_changed:
+            self._message_log(
+                tracking_value_ids=[Command.create({
+                    'field_id': self.env['ir.model.fields']._get('mrp.production', 'date_finished').id,
+                    'old_value_datetime': old_date_finished,
+                    'new_value_datetime': self.date_finished,
+                })]
+            )
```
dhha-odoo: Now, we allow users to modify the `date_finished` field even when the Manufacturing Order (MO) is in the `Done stage`.

For example, if there are `100 MOs in the Done stage`, all of them being `backorders`, a user may want to update the `finished date for all backorders at once`. To do so, the user would use the `multi-edit feature from the list` view, after making the field editable (e.g., by setting `readonly="false"` in the view via Studio).

In this case, the update is applied to multiple records at once, which can `lead to a singleton issue here`.

See image:

Image

Yes, I agree that this edge case may occur very rarely, but we should still handle it properly. Since users have access to Studio, they can make the field editable and perform such actions if they want to. Therefore, we need to consider all possible cases and handle them safely, avoiding custom logic that could lead to a traceback.

dhha-odoo (changes requested):

addons/mrp/models/mrp_production.py
stan-odoo: Thank you for your suggestion.
I handled this case. 
Please let me know if anything is required.

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,24 @@ def write(self, vals):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        date_finished_changed = 'date_finished' in vals and 'state' not in vals
+
+        production_initial_values = defaultdict(dict)
+        if date_finished_changed:
```
tpma-odoo: Use the condition directly in the `if` statement, no need to create a separate variable.

addons/mrp/models/mrp_production.py
```diff
@@ -154,7 +154,7 @@ def _get_default_is_locked(self):
     date_finished = fields.Datetime(
         'End', copy=False, default=_get_default_date_finished,
         compute='_compute_date_finished', store=True,
-        help="Date you expect to finish production or actual date you finished production.")
+        help="Date at which you finished production.")
```
tpma-odoo: avoid changing the help text here, since this field is also used in the `list view` where it can represent both the `expected finish date` (when not finished yet) and the `actual finish date` once the MO is done.

tpma-odoo (changes requested):

tpma-odoo:
Hello @stan-odoo ,

Just one small change. The title could still be improved to better reflect the purpose, something like: `display and track end date on done MO`?

Also, the commit message could mention that `any modification to the end date after the MO is marked as done is logged in the chatter to maintain a complete audit trail.`

Thank You.

tpma-odoo (approved):

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,20 @@ def write(self, vals):
...
+            for production in self:
+                if production.state == 'done':
+                    production_initial_values[production] = {'date_finished': production.date_finished}
+
         res = super(MrpProduction, self).write(vals)
 
+        for production, initial_value in production_initial_values.items():
+            tracking_value_ids = production._mail_track(tracked_fields, initial_value)[1]
+            if tracking_value_ids:
+                production._message_log(tracking_value_ids=tracking_value_ids)
```
dhha-odoo:
```suggestion
            if tracking_values := production._mail_track(tracked_fields, initial_value)[1]:
                production._message_log(tracking_value_ids=tracking_values)
```

Avoids creating a temporary variable before the if using `assignment expression operator.`
[Improve the variable name](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions).

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,20 @@ def write(self, vals):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        production_initial_values = {}
+        if 'date_finished' in vals:
+            tracked_fields = self.fields_get(['date_finished'])
+            for production in self:
+                if production.state == 'done':
+                    production_initial_values[production] = {'date_finished': production.date_finished}
```
dhha-odoo:
```suggestion
            for production in self.filtered(lambda p: p.state == 'done'):
                production_initial_values[production] = {'date_finished': production.date_finished}
```

dhha-odoo (changes requested):

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,20 @@ def write(self, vals):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        production_initial_values = {}
```
dhha-odoo:
```suggestion
        initial_date_finished_mo_values = {}
```

dhha-odoo (changes requested):

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,18 @@ def write(self, vals):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        initial_date_finished_mo_values = {}
+        if 'date_finished' in vals:
+            tracked_fields = self.fields_get(['date_finished'])
```
dhha-odoo:
```suggestion
            date_finished_field_definition = self.fields_get(['date_finished'])
```

Try to use a meaningful variable name. `date_finished is not a tracking field`, and it is not `_fields(For just one field?)`.

`self.fields_get(['date_finished'])` actually returns the `field definition`, so the variable name should reflect that.

dhha-odoo (changes requested):

addons/mrp/models/mrp_production.py
```diff
@@ -1024,8 +1024,18 @@ def write(self, vals):
         if vals.get('state') == 'progress' and 'date_start' not in vals:
             vals['date_start'] = fields.Datetime.now()
 
+        initial_date_finished_mo_values = {}
+        if 'date_finished' in vals:
```
dhha-odoo:
```suggestion
        if 'date_finished' in vals and 'state' not in vals:
```

If you create an `MO and mark it as done`, a log note for `date_finished will be generated` even though the field was `not manually changed and was not even visible to the user at that point`. This may lead to unnecessary tracking messages.

dhha-odoo (changes requested):

dhha-odoo (approved):

addons/mrp/views/mrp_production_views.xml
```diff
@@ -390,6 +390,7 @@
                                 <field name="delay_alert_date" invisible="1"/>
                                 <field nolabel="1" name="json_popover" widget="stock_rescheduling_popover" invisible="not json_popover"/>
                             </div>
+                            <field name="date_finished" string="End Date" readonly="is_locked" invisible="state != 'done'" help="Date at which you finished production."/>
```
svs-odoo, line 392: The field is already in the view as an invisible field (line 248), you can remove it since it's useless to have two times this same field.

PR #246523, opened 2026-01-30
[FIX] web: avoid non-deterministic error in `barcode_dialog` test

rfr-odoo:
This commit ensure to await the view to be correctly re-rendered.

runbot-error-233551

---


aab-odoo (approved):
robodoo r+

PR #246526, opened 2026-01-30
[FIX] resource: Duplicated public holiday

3amo-magdy:
Problem:
If you duplicate the working schedule, it will also duplicate global leaves.

This commit prevents this undesired behavior.

However, a cleaner yet not stable-friendly solution would be to make the relation M2m to allow the same global leave record to be shared among multiple calendars.

Task-5895007


addons/resource/models/resource.py
```diff
@@ -255,6 +255,8 @@ def copy(self, default=None):
             default = {}
         if not default.get('name'):
             default.update(name=_('%s (copy)') % (self.name))
+        if not default.get('global_leave_ids'):
+            default.update(global_leave_ids=False)
```
romaincarlier4: Can you not just set the field as copy=False ?

PR #246529, opened 2026-01-30
[IMP] l10n_fr:  add SIRET format validation warning on partners

dvsh-odoo:
- Display a non-blocking warning on the partner form when a French company SIRET is invalid.

- The validation relies on stdnum.fr.siret to check the official SIRET rules (length, format, checksum, La Poste exceptions), and surfaces the library error message with contextual information.

TaskID-5882793

### DIFF
--- a/addons/l10n_fr/models/res_company.py
+++ b/addons/l10n_fr/models/res_company.py
@@ -8,6 +8,7 @@ class ResCompany(models.Model):
     _inherit = 'res.company'
 
     l10n_fr_closing_sequence_id = fields.Many2one('ir.sequence', 'Sequence to use to build sale closings', readonly=True)
+    l10n_fr_siret_warning = fields.Char(related='partner_id.l10n_fr_siret_warning')
     ape = fields.Char(string='APE')
     is_france_country = fields.Boolean(
         compute="_compute_is_france_country",
--- a/addons/l10n_fr/models/res_partner.py
+++ b/addons/l10n_fr/models/res_partner.py
@@ -1,3 +1,6 @@
+from stdnum.exceptions import ValidationError as StdnumValidationError
+from stdnum.fr import siret
+
 from odoo import api, fields, models
 
 
@@ -5,8 +8,29 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     l10n_fr_is_french = fields.Boolean(compute='_compute_l10n_fr_is_french')
+    l10n_fr_siret_warning = fields.Char(compute='_compute_l10n_fr_siret_warning')
 
     @api.depends('country_code')
     def _compute_l10n_fr_is_french(self):
         for partner in self:
             partner.l10n_fr_is_french = partner.country_code in self.env['res.company']._get_france_country_codes()
+
+    @api.depends('company_registry', 'l10n_fr_is_french')
+    def _compute_l10n_fr_siret_warning(self):
+        for partner in self:
+            partner.l10n_fr_siret_warning = False
+            if (not partner.company_registry or not partner.l10n_fr_is_french):
+                continue
+
+            try:
+                siret.validate(partner.company_registry)
+            except StdnumValidationError as e:
+                partner.l10n_fr_siret_warning = self.env._(
+                    "Invalid SIRET: %(error)s",
+                    error=str(e),
+                )
+
+    def _get_company_registry_labels(self):
+        labels = super()._get_company_registry_labels()
+        labels['FR'] = self.env._("SIRET")
+        return labels
--- a/addons/l10n_fr/views/res_company_views.xml
+++ b/addons/l10n_fr/views/res_company_views.xml
@@ -10,6 +10,22 @@
              <xpath expr="//field[@name='company_registry']" position="after">
                  <field name="ape" invisible="not is_france_country"/>
              </xpath>
+            <xpath expr="//sheet" position="before">
+                <div class="alert alert-warning mt-1 mb-1"
+                    role="alert"
+                    invisible="not is_france_country or not l10n_fr_siret_warning">
+                    <field name="l10n_fr_siret_warning"/>
+                </div>
+            </xpath>
+            <xpath expr="//field[@name='company_registry']" position="attributes">
+                <attribute name="invisible" add="is_france_country" separator=" or "/>
+            </xpath>
+            <xpath expr="//field[@name='company_registry']" position="after">
+                <field name="company_registry"
+                       string="SIRET"
+                       options="{'placeholder_field': 'company_registry_placeholder'}"
+                       invisible="not is_france_country"/>
+            </xpath>
         </data>
         </field>
     </record>
--- a/addons/l10n_fr/views/res_partner_views.xml
+++ b/addons/l10n_fr/views/res_partner_views.xml
@@ -6,8 +6,15 @@
         <field name="inherit_id" ref="base.view_partner_form"/>
         <field name="priority">14</field>
         <field name="arch" type="xml">
+            <xpath expr="//sheet" position="before">
+                <div class="alert alert-warning mt-1 mb-1"
+                    role="alert"
+                    invisible="not l10n_fr_is_french or not l10n_fr_siret_warning">
+                    <field name="l10n_fr_siret_warning"/>
+                </div>
+            </xpath>
             <field name="company_registry" position="attributes">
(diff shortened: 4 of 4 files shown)


addons/l10n_fr/views/res_company_views.xml
```diff
@@ -10,6 +10,22 @@
...
+                </div>
+            </xpath>
+            <xpath expr="//field[@name='company_registry']" position="attributes">
+                <attribute name="invisible" add="is_france_country" separator=" or "/>
+            </xpath>
+            <xpath expr="//field[@name='company_registry']" position="after">
+                <field name="company_registry"
+                       string="SIRET"
+                       options="{'placeholder_field': 'company_registry_placeholder'}"
+                       invisible="not is_france_country"/>
```
khva-odoo, line 27: We should also overwrite `_get_company_registry_labels` to include the label for FR.

addons/l10n_fr/views/res_company_views.xml
```diff
@@ -10,6 +10,22 @@
              <xpath expr="//field[@name='company_registry']" position="after">
                  <field name="ape" invisible="not is_france_country"/>
              </xpath>
+            <xpath expr="//sheet" position="before">
+                <div class="alert alert-warning mt-1 mb-1"
+                    role="alert"
+                    invisible="not is_france_country or not l10n_fr_siret_warning">
+                    <field name="l10n_fr_siret_warning" nolabel="1"/>
```
khva-odoo:
```suggestion
                    <field name="l10n_fr_siret_warning"/>
```

addons/l10n_fr/models/res_partner.py
```diff
@@ -1,12 +1,31 @@
...
 
     l10n_fr_is_french = fields.Boolean(compute='_compute_l10n_fr_is_french')
+    l10n_fr_siret_warning = fields.Char(compute="_compute_l10n_fr_siret_warning")
 
     @api.depends('country_code')
     def _compute_l10n_fr_is_french(self):
         for partner in self:
             partner.l10n_fr_is_french = partner.country_code in self.env['res.company']._get_france_country_codes()
+
+    @api.depends("company_registry", "is_company", "l10n_fr_is_french")
```
khva-odoo: Why `is_company` in the dependency??

khva-odoo (commented):
Hello,
Thanks for the PR!
Can you please have look into the comments?

addons/l10n_fr/models/res_company.py
```diff
@@ -8,6 +8,7 @@ class ResCompany(models.Model):
     _inherit = 'res.company'
 
     l10n_fr_closing_sequence_id = fields.Many2one('ir.sequence', 'Sequence to use to build sale closings', readonly=True)
+    l10n_fr_siret_warning = fields.Char(related="partner_id.l10n_fr_siret_warning", readonly=True)
```
khva-odoo:
```suggestion
    l10n_fr_siret_warning = fields.Char(related="partner_id.l10n_fr_siret_warning")
```

addons/l10n_fr/models/res_partner.py
```diff
@@ -1,12 +1,31 @@
-from odoo import api, fields, models
+from stdnum.fr import siret
+from stdnum.exceptions import ValidationError as StdnumValidationError
```
khva-odoo: order of import

khva-odoo (approved):

addons/l10n_fr/models/res_company.py
```diff
@@ -8,6 +8,7 @@ class ResCompany(models.Model):
     _inherit = 'res.company'
 
     l10n_fr_closing_sequence_id = fields.Many2one('ir.sequence', 'Sequence to use to build sale closings', readonly=True)
+    l10n_fr_siret_warning = fields.Char(related="partner_id.l10n_fr_siret_warning")
```
khva-odoo: Didn't get noticed, single quotes for the technical string.

addons/l10n_fr/models/res_partner.py
```diff
@@ -1,12 +1,36 @@
...
+from stdnum.fr import siret
+
+from odoo import _, api, fields, models
 
 
 class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     l10n_fr_is_french = fields.Boolean(compute='_compute_l10n_fr_is_french')
+    l10n_fr_siret_warning = fields.Char(compute="_compute_l10n_fr_siret_warning")
```
khva-odoo: same here

addons/l10n_fr/models/res_partner.py
```diff
@@ -1,12 +1,36 @@
...
 
     @api.depends('country_code')
     def _compute_l10n_fr_is_french(self):
         for partner in self:
             partner.l10n_fr_is_french = partner.country_code in self.env['res.company']._get_france_country_codes()
+
+    @api.depends("company_registry", "l10n_fr_is_french")
```
khva-odoo: same here

addons/l10n_fr/models/res_partner.py
```diff
@@ -1,12 +1,36 @@
...
+                siret.validate(partner.company_registry)
+            except StdnumValidationError as e:
+                partner.l10n_fr_siret_warning = _(
+                    "Invalid SIRET: %(error)s",
+                    error=str(e),
+                )
+
+    def _get_company_registry_labels(self):
+        labels = super()._get_company_registry_labels()
+        labels['FR'] = _("SIRET")
```
khva-odoo: Maybe we should use `self.env._` here instead.

khva-odoo (commented):
Just a formating changes, which should be the blocking point for the testing.

addons/l10n_fr/views/res_partner_views.xml
```diff
@@ -6,8 +6,15 @@
...
+            <xpath expr="//sheet" position="before">
+                <div class="alert alert-warning mt-1 mb-1"
+                    role="alert"
+                    invisible="not l10n_fr_is_french or not l10n_fr_siret_warning">
+                    <field name="l10n_fr_siret_warning"/>
+                </div>
+            </xpath>
             <field name="company_registry" position="attributes">
-                <attribute name="invisible">parent_id or not is_company or l10n_fr_is_french</attribute>
+                <attribute name="invisible">parent_id or (not is_company and l10n_fr_is_french)</attribute>
```
alialfie: What is this change fixing? Couldn't find it in the task or the commit message

alialfie (commented):
Thanks for the PR! Left one question below

addons/l10n_fr/views/res_partner_views.xml
dvsh-odoo: The base view has no **_invisible_** condition on `company_registry`, the field is always visible regardless of _**is_company**_. 

The original **"l10n_fr"** xpath uses position="attributes" which fully replaces that attribute, and the condition it sets  `parent_id or not is_company or l10n_fr_is_french`, introduces **not is_company** hiding that never existed in the base. This means after **l10n_fr** is installed, **individuals** on **any non-French localization** (e.g. l10n_sa) lose visibility of a field they should always see.

The fix changes it to `parent_id or (not is_company and l10n_fr_is_french)`, which scopes the individual-hiding to French partners only. Non-French partners remain completely unaffected, and French behavior is preserved, the generic field is still hidden for FR partners since the Siret field takes over.

alialfie: I understand, however, this new logic accidentally introduces a bug for French companies.

If we evaluate a French company, `parent_id or (not is_company and l10n_fr_is_french)` becomes `False or (not True and True)`, which evaluates to `False`. This means the generic field will now be visible alongside the custom French SIRET field, resulting in duplicates.

The original condition was definitely weird because it shouldn't be checking `is_company` at all. If you look at the code right below this, the custom SIRET field is shown for all French partners, regardless of whether they are individuals or companies.

Because the SIRET field completely takes over for France, I think the cleanest fix is simply:
`parent_id or l10n_fr_is_french`

This safely hides the generic field for all French contacts, while leaving it perfectly visible for individuals in other localizations.

dvsh-odoo: Ahh got it, thanks for pointing that out.

I got a bit confused between the two field nodes and ended up mixing the logic here. You’re right that this one is purely localization-driven, so `is_company` doesn’t belong.

I’ve fixed it to `parent_id or l10n_fr_is_french`. Please review when you get time 🙂

alialfie (approved):
thankss
robodoo r+

PR #246531, opened 2026-01-30
[FIX] mrp: Fix memory error

sagu-odoo:
Previously, ``self.order_ids`` fetched all work orders  all work centers, causing issue during fetching when there are multiple workorder per workcentre. Updated the logic to check only the count if order exist or not since that is all required.

opw-5472114
upg-3862178

---

### DIFF
--- a/addons/mrp/models/mrp_workcenter.py
+++ b/addons/mrp/models/mrp_workcenter.py
@@ -134,7 +134,8 @@ class MrpWorkcenter(models.Model):
     def _get_workcenter_load_per_week(self, week_range, date_start, date_stop):
         load_data = {rec: {} for rec in self}
         # demo data
-        if not self.order_ids:
+        has_workorder = self.env['mrp.workorder'].search_count([('workcenter_id', 'in', self.ids)], limit=1)
+        if not has_workorder:
             for wc in self:
                 load_limit = 40     # default max load per week is 40 hours on a new workcenter
                 load_data[wc] = {week_start: randint(0, int(load_limit * 2)) for week_start in week_range}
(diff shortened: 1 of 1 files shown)


addons/mrp/models/mrp_workcenter.py
```diff
@@ -153,7 +153,7 @@ def _prepare_graph_data(self, load_data, week_range):
         graph_data = {wid: [] for wid in self._ids}
         for workcenter in self:
             load_limit = sum(workcenter.resource_calendar_id.attendance_ids.mapped('duration_hours'))
-            wc_data = {'is_sample_data': not self.order_ids, 'labels': list(week_range.values())}
+            wc_data = {'is_sample_data': not self.env['mrp.workorder'].search([('workcenter_id', 'in', self.ids)], limit=1), 'labels': list(week_range.values())}
```
clesgow: ~~Don't do a `search()` inside a for loop.
Instead, do a `_read_group()` from outside the loop and just use a `id:count` as aggregator~~

EDIT: Nevermind, I misread. Instead, do the same query I suggested in the other answer and move it before the for loop. And just use the result in the loop, as we're looking at all workcenters.

addons/mrp/models/mrp_workcenter.py
```diff
@@ -134,7 +134,7 @@ def _get_week_range_and_first_last_days(self):
     def _get_workcenter_load_per_week(self, week_range, date_start, date_stop):
         load_data = {rec: {} for rec in self}
         # demo data
-        if not self.order_ids:
+        if not self.env['mrp.workorder'].search([('workcenter_id', 'in', self.ids)], limit=1):
```
clesgow:
```suggestion
        has_workorder = self.env['mrp.workorder'].search_count([('workcenter_id', 'in', self.ids)], limit=1)
        if not has_workorder:
```

addons/mrp/models/mrp_workcenter.py
sagu-odoo: Hello @clesgow Thanks for your review. One thing I’d like to point out is that it checks ``self.order_ids``, not ``workcenter.order_ids`` then we can directly do  [this](https://github.com/odoo/odoo/pull/246531/changes#r2768879226) here also.

clesgow: Yeah I edited the comment above.

clesgow:
Thanks for the fix!
@robodoo r+

PR #222186, opened 2025-08-07
[IMP] l10n_ie:  Added account tags for ireland CoA

unka-odoo:
Added dedicated account tags for ireland CoA. And set these tags to corresponding Irish and default accounts.
These tags are used for computation in new PL and BS tags reports.

Related Enterprise PR: https://github.com/odoo/enterprise/pull/91903

Task: 4978415

PR #186467, opened 2024-11-06
[FIX] account: ensure registry is loaded

aj-fuentes:
During a major upgrade the registry my be `loaded` but not `ready`. https://github.com/odoo/odoo/blob/9918e8f3d627f3c52238d6b04bcd15c05d34e40c/odoo/modules/registry.py#L159-L160

This causes unnecessary warnings during upgrades (blocking the CI).

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246541, opened 2026-01-30
[FIX] l10n_ie: Move the update of CoA in end-script

bda-odoo:
We faced a warning because this update doing a try_loading is called in `post-` scripts (after the module is loaded) and not in `end-` scripts (when all modules are loaded).
All l10n_* modules were changed in this PR: https://github.com/odoo/odoo/pull/183359

Regarding the changes in this PR related to Registry: https://github.com/odoo/odoo/pull/186467

see: https://github.com/odoo/odoo/pull/222186
upg-3828301

**Reference Image**

Current behavior before PR:

---


clbr-odoo (approved):
@robodoo r+

PR #246543, opened 2026-01-30
[FIX] im_livechat: no yellow bg in 'create lead' livechat action

alexkuhn:
Before this commit, when using the "Create Lead" thread action in discuss app on a live chat conversation, the background of input was yellow.

This happens because this is the default color of input of `<input>` in dark theme in a popover. This commit fixes by setting the `.form-control` classname on the `input`, which is intended to be used in almost all inputs to have proper web client style.

Part of Task-5867464

Before / After
Screenshot 2026-01-30 at 15 15 33
Screenshot 2026-01-30 at 15 14 46


seb-odoo (approved):
@robodoo r+ delegate+

PR #246554, opened 2026-01-30
[FIX] website_event: restore "Card Design" option

anso-odoo:
This commit restores the "Card design" option that was broken after the merge of this commit [*].

`opt_events_list_cards` was previously declared in `index` template, but it is only used inside `events_list`. This commit moves the variable declaration directly into `events_list`, making the card layout apply correctly and simplifying the parent template.

*: https://github.com/odoo/odoo/commit/bba2fc505f5d0b4770eacc6877155b1aeda6d772

task-5886440

---

### DIFF
--- a/addons/website_event/views/event_templates_list.xml
+++ b/addons/website_event/views/event_templates_list.xml
@@ -6,10 +6,7 @@
     <t t-set="head">
         <meta t-if="search_tags" name="robots" content="none"/>
     </t>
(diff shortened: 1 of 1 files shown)


reth-odoo (approved):
🤔 Indeed, no point in making this a "parameter", it's always going to be based on the same condition.

@Odoonan

PR #246556, opened 2026-01-30
[IMP] mail: rename 'Invite a User' to 'Invite People'

alexkuhn:
For consistency with other "Invite People" actions that result to the same action.

Part of Task-5867464

Before / After
Screenshot 2026-01-30 at 16 22 13 Screenshot 2026-01-30 at 16 21 37


seb-odoo (approved):
@robodoo r+ delegate+

PR #246557, opened 2026-01-30
[FIX] web: accept "True" as a value for the "nolabel" attribute

oomsveta:
## Before this commit
`nolabel="True"` sets nolabel to `false`

## After this commit
`nolabel="True"` sets nolabel to `true`

### DIFF
--- a/addons/web/static/src/views/form/form_compiler.js
+++ b/addons/web/static/src/views/form/form_compiler.js
@@ -341,7 +341,7 @@ export class FormCompiler extends ViewCompiler {
             let slotContent;
             if (getTag(child, true) === "field") {
                 const addLabel = child.hasAttribute("nolabel")
-                    ? child.getAttribute("nolabel") !== "1"
+                    ? !exprToBoolean(child.getAttribute("nolabel"))
                     : true;
                 slotContent = this.compileNode(child, { ...params, currentSlot: mainSlot }, false);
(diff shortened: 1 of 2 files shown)


addons/web/static/src/views/form/form_compiler.js
```diff
@@ -340,9 +340,11 @@ export class FormCompiler extends ViewCompiler {
 
             let slotContent;
             if (getTag(child, true) === "field") {
-                const addLabel = child.hasAttribute("nolabel")
-                    ? child.getAttribute("nolabel") !== "1"
-                    : true;
+                let addLabel = true;
+                if (child.hasAttribute("nolabel")) {
+                    const nolabel = child.getAttribute("nolabel");
+                    addLabel = nolabel !== "1" && nolabel !== "True";
```
aab-odoo: we typically use `exprToBoolean`, as it tolerates `true` as well. Actually, it tolerates everything that isn't falsy (`0`, `False`, `false`), so kind of the opposite of the logic here, but it would be consistent with other attributes/options.

aab-odoo (commented):
I would also change an arch or two in tests to use `true`/`True` instead of `1`

aab-odoo (approved):
robodoo delegate+

addons/web/static/src/views/form/form_compiler.js
oomsveta: Changes done :+1:

oomsveta (commented):
robodoo r+

PR #242863, opened 2026-01-08
[FIX] account: batch merges are not decided by the right key

martiyen:
When we register payments for a list of journal entries, the `account.payment.register` wizard computes batches and sometimes merge them together.

For instance, this allows to create a single payment if there is an outbound (a bill to pay) and an inbound (a refund to receive) payment to the same bank for the same partner. Instead of creating two payments of -1000 and +500, we only create one of -500.

Currently, this mechanism does not always work. That's because the `batch_key` used to decide whether to merge or not refers to a value that is not updated in the loop.

Related ticket: opw-5401372

PR #246558, opened 2026-01-30
[FIX] account: beneficiary account can be incorrect after payments merge

martiyen:
When we create new payments for bills and refunds, we offer the possibility to merge inbound and outbound payments together if they are from the same provider and the bills all reference the same recipient account.

Depending on the balance of the resulting payment, we assign an adequate inbound or outbound bank account as the recipient.

The `partner_bank_id` can be assigned through different processes:

- If the wizard has only one batch:

The wizard is editable and the user can select a bank account from the computed `available_partner_bank_ids`.

- If there are multiple batches:

Odoo assigns a `partner_bank_id` in `_create_payment_vals_from_batch()`. For an outbound payment, it uses the batch['payment_values']['partner_bank_id'].

The problem is that this value is not updated after a merge of payments. If the base line being merged on is a refund, but the result is an outbound payment, then the `partner_bank_id` should be changed accordingly.

I decided to include the changes of my previous PR targeting v18 and fixing the grouping of payments, even though it was deemed unnecessary for v17, because it felt weird not to considering how close these fixes are.

-Previous PR : [242863](https://github.com/odoo/odoo/pull/242863)

However, I can remove these changes or re-target this PR to v18. I am not sure what would be best here.

Ticket: opw-5401372

### DIFF
--- a/addons/account/wizard/account_payment_register.py
+++ b/addons/account/wizard/account_payment_register.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 
 from odoo import Command, models, fields, api, _
 from odoo.exceptions import UserError
-from odoo.tools import frozendict
+from odoo.tools import frozendict, OrderedSet
 
 
 class AccountPaymentRegister(models.TransientModel):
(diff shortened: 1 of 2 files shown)


addons/account/wizard/account_payment_register.py
william-andre: Use `OrderedSet` to ensure determinism in tests

william-andre (approved):
@robodoo r+

PR #246559, opened 2026-01-30
[FIX] mail: muted color for looking for help live chat description

alexkuhn:
Text was using `.text-warning`, which is too distracting.

Part of Task-5867464

Before / After
Screenshot 2026-01-30 at 16 27 00
Screenshot 2026-01-30 at 16 26 31

### DIFF
--- a/addons/im_livechat/static/src/core/web/discuss_content_patch.js
+++ b/addons/im_livechat/static/src/core/web/discuss_content_patch.js
@@ -13,7 +13,7 @@ patch(DiscussContent.prototype, {
     get threadDescriptionAttClass() {
         return {
             ...super.threadDescriptionAttClass,
-            "text-warning":
+            "text-muted":
                 this.thread.channel?.livechat_status === "need_help" && this.thread.description,
         };
     },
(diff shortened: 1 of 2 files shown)


seb-odoo (approved):
@robodoo r+ delegate+

addons/im_livechat/static/src/core/web/discuss_content_patch.js
```diff
@@ -13,7 +13,7 @@ patch(DiscussContent.prototype, {
     get threadDescriptionAttClass() {
         return {
             ...super.threadDescriptionAttClass,
-            "text-warning":
+            "text-muted":
```
seb-odoo, line 16: Maybe this one just remove and keep the standard description style?

alexkuhn: If we keep the standard style, the color is not the same, and this is unclear visually the 2 texts refer to the same feature:

Screenshot 2026-01-30 at 17 27 07

vs.

Screenshot 2026-01-30 at 17 27 41


For us this seems obvious, but the inconsistent color makes it unintuitive.

PR #246564, opened 2026-01-30
[IMP] website, *: refresh snippet thumbnails

Brieuc-brd:
*: website, website_blog, website_event, website_payment, mass_mailing

This PR updates the snippet, header, footer and mega menu thumbnails in the web editor with new
more modern designs. The updated thumbnails are more visually appealing
and provide a clearer representation of each snippet category.

task-5868664

Requires:
- https://github.com/odoo/enterprise/pull/106045

| Before | After |
|--------|--------|
| Capture d’écran 2026-04-13 à 08 33 38 | Capture d’écran 2026-04-13 à 08 32 24 |
| Screenshot 2026-05-26 at 16 15 48 | Screenshot 2026-05-26 at 16 16 09 |

---


addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -7,40 +7,47 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
     position: relative;
     z-index: $o-we-zindex;
     width: 77px;
-    background-color: $o-we-sidebar-blocks-content-snippet-bg;
+    background-color: $o-we-bg-lighter;
 
     &.o_draggable_dragging {
         transform: rotate(-3deg) scale(1.2);
+        border-radius: map-get($spacers, 1);
```
chgo-odoo: Let's use a `$border-radius` value instead

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -7,40 +7,47 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
     position: relative;
     z-index: $o-we-zindex;
     width: 77px;
-    background-color: $o-we-sidebar-blocks-content-snippet-bg;
+    background-color: $o-we-bg-lighter;
```
chgo-odoo: Since you replaced `$o-we-sidebar-blocks-content-snippet-bg` with `$o-we-bg-lighter`, maybe it's worth adapting the remaining occurences so we can delete `$o-we-sidebar-blocks-content-snippet-bg` entirely ?

But maybe we can keep it, from what I see it always matches `$o-we-bg-lighter` so maybe this intermediate variable is fine ?

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -7,40 +7,47 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
...
-    &:not(:hover):not(.o_disabled):not(.o_to_install) {
-        background-color: rgba($o-we-sidebar-blocks-content-snippet-bg, .9);
+    &:hover:not(.o_block_disabled):not(.o_to_install) {
+        background: $o-we-bg-light;
 
-        .o_snippet_thumbnail {
-            filter: saturate(.7);
-            opacity: .9;
+        .o_snippet_thumbnail_title {
+            color: #fff;
```
chgo-odoo: Nitpicking but maybe we can use `$o-we-fg-lighter` ?

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -106,21 +113,19 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
...
 
-            &.o_disabled, &.o_to_install {
+            &.o_block_disabled, &.o_to_install {
                 cursor: default;
                 background-color: rgba($o-we-sidebar-blocks-content-snippet-bg, .2);
 
                 .o_snippet_thumbnail_img {
-                    opacity: .4;
-                    filter: saturate(0) blur(1px);
+                    opacity: .5;
```
chgo-odoo: `$o-opacity-disabled` ?

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -53,42 +60,42 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
...
     .o_snippets_container_body {
-        display: flex; // Needed for too long snippet names
-        flex-wrap: wrap;
-        margin-left: -$o-we-sidebar-blocks-content-snippet-spacing;
-        z-index: 0;
         position: relative;
+        display: grid;
+        grid-template-columns: repeat(var(--SnippetsContainer-columns, 3), 1fr);
+        gap: 2px;
+        z-index: 0;
```
chgo-odoo: Can't we put part of this in the XML ?

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -53,42 +60,42 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
...
+                flex-direction: row;
+                justify-content: flex-start;
             }
         }
 
         .o_snippet {
-            flex: 0 0 auto;
-            width: percentage(1 / 3);
+            width: 100%;
+            border-radius: map-get($spacers, 1);
```
chgo-odoo: Same as above for the radius

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -7,40 +7,47 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
...
+            width: 30px;
+            padding-top: 45%;
             background-repeat: no-repeat;
-            background-size: cover;
+            background-size: contain;
             background-position: center;
-            overflow: hidden;
+
+            #snippet_content & {
+                margin: map-get($spacers, 2) 0;
```
chgo-odoo: Nitpicking ☝️🤓 (approved by MANO) but let's use `margin-block` instead so we avoid setting a margin for both axes

chgo-odoo (changes requested):
Hello @Brieuc-brd 👋 

Congrats on your efforts on this, the new thumbnails look way better 🤩 

Left some minor comments regarding the CSS, will recheck once it's pushed to ensure I didn't miss anything

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -106,21 +111,19 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
                 }
             }
 
-            &.o_disabled {
-                pointer-events: initial;
-
+            &.o_block_disabled {
```
chgo-odoo: Since you replaced `o_disabled` with `o_block_disabled` we lost the `opacity: .5` applied by the class. This is probably a nitpick and I'll let you decide but I personally prefer when the text doesn't stand out that much

I'd suggest to move [this line](https://github.com/odoo/odoo/pull/246564/changes?file-filters%5B%5D=.scss&file-filters%5B%5D=.xml&file-filters%5B%5D=.js#diff-a326fa2cee42a527a55df62c8050c14d73271cfde6726c58f792341c70a46101R125) to a higher element (but it'll affect the "!" as well so if you don't want that, maybe apply this rule on the label as well

Image

Image

chgo-odoo (approved):
LGTM beside this minor feedback

Thanks @Brieuc-brd !

Brieuc-brd:
Hey @odoo/rd-website 👋 
These COM and ENT PRs are ready for tech. review, could you have a look at them ?
Thanks in advance 👍

AntoineVDV (approved):

Brieuc-brd:
Hey @odoo/rd-website 👋 
Small reminder 🙂

FrancoisGe:
@robodoo delegate=bvr-odoo

addons/html_builder/static/src/sidebar/block_tab.scss
bvr-odoo, line 1: The `mass_mailing` sidebar is affected by these changes, but it still uses the old icons.

Image

Brieuc-brd:
Thanks @bvr-odoo 👍 

> Why didn't you remove the old images?

I've removed them 👍 

> It also seems that we forgot mass_mailing

It's now fixed ✅

bvr-odoo:
@abd-msyukyu-odoo Hello, ok for you the new icons in `mass_mailing` ?

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -140,10 +143,6 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
             min-width: 0; // Ensure text-overflow on flex children
         }
 
-        .o_snippet_thumbnail_title {
-            white-space: nowrap;
-        }
-
         .o_snippet_thumbnail_img {
             flex-shrink: 0;
             width: 41px;
```
sobo-odoo: I think you forgot to adapt them, but I find the icons too big on custom inner snippets, they are bigger than the inner snippets ones (they are supposed to be smaller) :open_mouth: 
Could you reduce their height a bit ? :slightly_smiling_face:

image

With 18px it looks fine and it doesn't make the rectangle bigger :slightly_smiling_face: 
image

addons/website/static/src/img/snippets_thumbs/s_progress_bar.svg
sobo-odoo, line 1: If I can give my opinion, I don't think it really looks like a progress bar, it is a bit too squarish => for me it should either be thinner or longer, and we should make the border smaller too :slightly_smiling_face: 
image

Or maybe just removing the border and making the loading look more like 75% is enough ? To see :thinking: 

Ideally, it should look more like the bottom line of the Text snippet icon :thinking: 
image

It's the only icon that bugs me, the other ones are really nice :slightly_smiling_face:

Brieuc-brd:
Thanks for your feedback @sobo-odoo 👍 .
I’ve reviewed the icon sizes and the progress bar icon as well ✅ 

Also, since [this change](https://github.com/odoo/odoo/pull/246564/changes#diff-a326fa2cee42a527a55df62c8050c14d73271cfde6726c58f792341c70a46101R30), which is necessary to get proper icon rendering, I’m getting [this weird error](https://runbot.odoo.com/runbot/batch/2553352/build/112326246) on Runbot.
I don’t understand why using a fixed height instead of padding would trigger this error. Do you have any idea what could be causing it? 🤔 

cc. @bvr-odoo

bvr-odoo:
> Thanks for your feedback @sobo-odoo 👍 . I’ve reviewed the icon sizes and the progress bar icon as well ✅
> 
> Also, since [this change](https://github.com/odoo/odoo/pull/246564/changes#diff-a326fa2cee42a527a55df62c8050c14d73271cfde6726c58f792341c70a46101R30), which is necessary to get proper icon rendering, I’m getting [this weird error](https://runbot.odoo.com/runbot/batch/2553352/build/112326246) on Runbot. I don’t understand why using a fixed height instead of padding would trigger this error. Do you have any idea what could be causing it? 🤔
> 
> cc. @bvr-odoo

Hello @Brieuc-brd Thanks for the changes

About the runbot issue, it's because, with your changes, the helper rect/offset changes, so the test now hits the dropzone after the existing button instead of the one before it. I guess you can just adapt the test 🤷🏼‍♂️

addons/html_builder/static/src/sidebar/block_tab.scss
```diff
@@ -11,36 +11,43 @@ $o-we-zindex: $o-we-overlay-zindex + 1 !default;
```
sobo-odoo, line 11: > About the runbot issue, it's because, with your changes, the helper rect/offset changes, so the test now hits the dropzone after the existing button instead of the one before it. I guess you can just adapt the test 🤷🏼‍♂️

@Brieuc-brd @bvr-odoo No it's not a test issue, it's more general :slightly_smiling_face: 
I don't know which change is responsible, but when we drag a snippet, the cursor is supposed to be centered on the thumbnail image, but now it is dragged from the left :arrow_down: 
| now | before |
|--------|--------|
| image | image | 

(the red crosses are the cursor position, I can't screenshot my cursor)

The helper offset computation is automatic: it is based on the width and height of the thumbnail and takes the middle, so maybe some style here changes the width on the way or something ? :thinking: 
image

sobo-odoo: Ahh I know !
Before, the thumbnail image was the same width as the thumbnail :arrow_down:, so taking the middle of it would center it correctly on drag and drop :slightly_smiling_face: 

image


But know the image is way smaller, so we can't use its width as a reference anymore :open_mouth: 
image

So taking only the middle of the image will not set it at the center of the whole snippet
=> We should use `draggedEl` directly now instead :slightly_smiling_face:  

EDIT: in fact maybe it is more complicated for the height :thinking:  So we should take into account the image height and width + padding (or gap or whatever that space around it is), or draggedEl width and image height + space or something
I let you check :slightly_smiling_face:

Brieuc-brd:
Thanks for the info @sobo-odoo & @bvr-odoo ! 🙏 
It should be fixed now ! 😄

bvr-odoo:
@robodoo r+

bvr-odoo:
@robodoo r+ rebase-ff

PR #246570, opened 2026-01-30
[FIX] mail: don't show timezone when same as self user in avatar card

alexkuhn:
Before this commit, the local time zone was shown on avatar card even when this is the same as current user.

The intent of showing of local timezone is to see when it differs. When this is the same, this just adds noise to the card.

This commit hides the showing of local timezone when this is the same as current user.

Part of Task-5867464

Before / After
Screenshot 2026-01-30 at 17 24 04 Screenshot 2026-01-30 at 17 23 50

### DIFF
--- a/addons/mail/static/src/discuss/web/avatar_card/avatar_card_popover.js
+++ b/addons/mail/static/src/discuss/web/avatar_card/avatar_card_popover.js
@@ -38,18 +38,21 @@ export class AvatarCardPopover extends Component {
             model: this.props.model,
         });
         useDynamicInterval(
-            (partnerTz, currentUserTz) => {
-                this.state.partnerLocalDateTimeFormatted = formatLocalDateTime(
-                    partnerTz,
-                    currentUserTz
-                );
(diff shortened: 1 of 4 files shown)


addons/mail/static/src/discuss/web/avatar_card/avatar_card_popover.js
```diff
@@ -38,20 +38,26 @@ export class AvatarCardPopover extends Component {
...
             () => [this.partner?.tz, this.store.self?.tz]
         );
     }
 
+    /**
+     * @param {string} partnerTz
+     * @param {string} currentUserTz
+     */
+    onChangeTz(partnerTz, currentUserTz) {
+        const partnerDtFormatted = formatLocalDateTime(partnerTz, currentUserTz);
```
seb-odoo: This should be in formatter, there are other callers.

alexkuhn: Done

seb-odoo (approved):
@robodoo r+ delegate+

PR #246581, opened 2026-01-30
[FIX] website,website_forum: forum search performance

h4818:
Improve the performance of the website forum search.

Issue:

There is a significant performance issue with a large number of forum posts impacting odoo.com.
This is due to the use of the `<%` operator on the content field of forum posts. It occurs when using the javascript search which autocompletes results in a dropdown on the search bar.

To deactivate the search on the `content` column, displayDescription needs to be set to false. but this was not possible due to the data attribute which contains text. setting it to 'false' was still truthy and therefore enabled the fuzzy search in the content column.

Fix:

- fix a js bug to properly cast the value to a boolean - this was done for other data attributes at the same time.
- set the display_description value to false to disable search on the content field. This is aligned with the python post search settings defined in: https://github.com/odoo/odoo/blob/7abd7ba2f38fdb1953c39fd3693f012c2ad1b497/addons/website_forum/controllers/website_forum.py#L95

### DIFF
--- a/addons/website/static/src/snippets/s_searchbar/search_bar.js
+++ b/addons/website/static/src/snippets/s_searchbar/search_bar.js
@@ -41,12 +41,12 @@ export class SearchBar extends Interaction {
         }
         const dataset = this.inputEl.dataset;
         this.options = {
-            displayImage: dataset.displayImage,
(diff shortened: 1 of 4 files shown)


FrancoisGe:
@robodoo r+

PR #246583, opened 2026-01-30
[FIX] account_edi_ubl_cii: Avoid importing existing bank belonging to different company

hsal-odoo:
Problem:
Assume there is in DB a bank account for partner P with account number A restricted to company C1.
If the user attempts to import a file with a bank account for partner P with account number A but for company C2
the search for the first bank account fails because it's restricted to a different company.
However, the creation of a new bank account also fails because the combination (P, A) already exists.

This was the case already with belgian demo data bank account and unit test files used in purchase_edi_ubl_bis3 module.

Solution:
This commit fixes this issue by:
1. Doing a sudo search for bank accounts to avoid duplicating (partner, account number) combination
2. avoid setting the bank account on the imported invoice if their company_id fields are not compatible.

build_error-237563

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_common.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_common.py
@@ -745,7 +745,7 @@ class AccountEdiCommon(models.AbstractModel):
         banks_to_create = []
         acc_number_partner_bank_dict = {
             bank.sanitized_account_number: bank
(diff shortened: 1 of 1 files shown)


smetl (approved):
@robodoo  r+

smetl (approved):
@robodoo r+

PR #246586, opened 2026-01-30
[FIX] mail: weird overflow of active action in dropdown

alexkuhn:
Webclient style has style to apply checkbox next to active dropdown item when it doesn't have the `.dropdown-item_active_noarrow`, through adding some content in `.dropdown-item:before`.

This checkbox is problematic because it overflows with `Dropdown`, and items in a discuss action list dropdown already have enough to show when they are active or not.

This commit cancels the webclient style of adding the checkbox by adding `dropdown-item_active_noarrow` on ActionList's dropdown items.

Part of Task-5867464

Before / After (see barely checkbox overflows on "Unmute" item):
Screenshot 2026-01-30 at 18 06 27
Screenshot 2026-01-30 at 18 17 42

### DIFF
--- a/addons/mail/static/src/core/common/action_list.xml
+++ b/addons/mail/static/src/core/common/action_list.xml
@@ -79,7 +79,7 @@
     })"/>
     <t t-set="attrs" t-value="{ 'aria-label': action.name, 'disabled': action.disabledCondition, 'name': action.id, 'data-sequence': action.sequence, 'data-sequence-group': action.sequenceGroup, 'data-sequence-quick': action.sequenceQuick }"/>
     <t t-if="action.component and action.componentCondition" t-component="action.component" t-props="action.componentProps"/>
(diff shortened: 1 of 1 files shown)


addons/mail/static/src/core/common/action_list.scss
```diff
@@ -11,6 +11,10 @@
         z-index: 1;
         --btn-disabled-opacity: 0.25;
 
+        &.active.dropdown-item:before {
+            content: "" !important; // cancel webclient style for :not(.dropdown-item_active_noarrow) 
+        }
```
tsm-odoo: Why not passing `dropdown-item_active_noarrow` as a dropdown item class props? This way we use the standard SCSS/item API instead of fighting it?

alexkuhn: Done.

tsm-odoo (approved):
@robodoo r+ delegate+

PR #236470, opened 2025-11-19
[FIX] event, website_event: allow outlook calendar to render HTML event descriptions

nsirjacobs:
****Behavior:****
**Current:**
When exporting an event for Outlook Calendar, the file returned is the same as for ICalendar(.ics), however Outlook cannot handle the HTML description of the event when passed through the 'DESCRIPTION:' field which causes the event description to display raw HTML tags.

**Solution:**
Outlook can use the 'X-ALT-DESC:' field with 'text/html' parameter to handle HTML. A new route was created to generate the event's ics file with the correct parameters for Outlook.

****Steps to reproduce:****
- Go to an event from Website
- Select the Outlook icon under 'Add to calendar'
- Import the file to Outlook Calendar
- You'll notice HTML tags in the description of the event.

opw-5116354

PR #246603, opened 2026-01-30
[FIX] event: render ical event description correctly

farahdawod:
### Steps to reproduce:
- Download Events app.
- Pick an event and click "Go to website".
- Download the Outlook Calendar.

### Issue:
- The HTML content within the event description is displayed as raw code and not converted properly
- This happens because outlook takes the description as text and displays it in the ics file without conversion

### Solution:
- Since we're working with *Outlook Calendar*, X Alt formatting was added to format HTML to text

opw-5237027

### DIFF
--- a/addons/event/models/event_event.py
+++ b/addons/event/models/event_event.py
@@ -762,7 +762,11 @@ class EventEvent(models.Model):
             cal_event.add('dtstart').value = event.date_begin.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
-            cal_event.add('description').value = event._get_external_description()
+            external_description = event._get_external_description()
+            cal_event.add('description').value = external_description
+            xalt = cal_event.add('X-ALT-DESC')
+            xalt.value = external_description
+            xalt.params['FMTTYPE'] = ['text/html']
             if event.address_id:
                 cal_event.add('location').value = event.address_inline
 
--- a/addons/website_event/tests/test_event_internals.py
+++ b/addons/website_event/tests/test_event_internals.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+import logging
+import unittest
 from datetime import datetime, timedelta
 
 from odoo.fields import Datetime as FieldsDatetime
@@ -9,6 +11,14 @@ from odoo.addons.website.tests.test_website_visitor import MockVisitor
 from odoo.addons.website.tools import MockRequest
 from odoo.addons.website_event.controllers.main import WebsiteEventController
 from odoo.addons.event.tests.common import EventCase
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import vobject
+except ImportError:
+    _logger.warning("`vobject` Python module not found, iCal file generation disabled. Consider installing this module if you want to generate iCal files")
+    vobject = None
 
 
 class TestEventData(EventCase, MockVisitor):
@@ -22,6 +32,38 @@ class TestEventData(EventCase, MockVisitor):
             'website_published': True,
         } for website_visibility in ['public', 'link', 'logged_users']])
         cls.events_visibility_test = cls.event_public | cls.event_link_only | cls.event_logged_users
+
+    @users('public_test')
+    def test_ics_file_html_description(self):
+        """Verify that _get_ics_file returns a valid .ics description
+        that will be rendered correctly for the user.
+        """
+        if not vobject:
+            raise unittest.SkipTest("Skip test when `vobject` Python module is not found.")
+
+        event = self.event_public
+        event.write({
+            'date_begin': FieldsDatetime.to_string(datetime(2022, 12, 31, 10, 0, 0)),
+            'date_end': FieldsDatetime.to_string(datetime(2022, 12, 31, 12, 0, 0)),
+            'description': '<p>This is <b>HTML</b> description</p>',
+            'event_register_url': 'https://www.example.com',
+        })
+
+        ics_map = event._get_ics_file()
+        ics_bytes = ics_map[event.id]
+        ics_str = ics_bytes.decode('utf-8')
+
+        cal = vobject.readOne(ics_str)
+        vevent = cal.vevent
+
+        external_description = event._get_external_description()
+        self.assertIn('This is HTML description', external_description)
+        self.assertEqual(vevent.description.value, external_description)
+
+        self.assertIn('x-alt-desc', vevent.contents)
+        xalt = vevent.contents['x-alt-desc'][0]
+        self.assertEqual(xalt.params.get('FMTTYPE'), ['text/html'])
+        self.assertEqual(xalt.value, external_description)
 
     def test_process_attendees_form(self):
         event = self.env['event.event'].create({


addons/event/models/event_event.py
```diff
@@ -762,7 +762,10 @@ def _get_ics_file(self):
             cal_event.add('dtstart').value = event.date_begin.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
-            cal_event.add('description').value = event._get_external_description()
+            cal_event.add('description').value = event._get_external_description(ical=True)
```
Ephoniz: I would maybe not have the same value in both the 'description' and the 'x-alt-desc', I would probably leave the plain text description within the `description, event._get_external_description(ical=False)` and the HTML description withing the `x-alt-dec,  event._get_external_description(ical=True)`

addons/website_event/models/event_event.py
```diff
@@ -461,10 +461,12 @@ def _track_subtype(self, init_values):
...
 
-    def _get_external_description(self):
+    def _get_external_description(self, ical=False):
         """ Adding the URL of the event into the description """
         self.ensure_one()
         event_url = f'<a href="{self.event_register_url}">{self.name}</a>'
+        if ical:
+            event_url = '<div style="white-space: pre-wrap;">' + event_url + '</div>'
         description = event_url + '\n' + super()._get_external_description()
         return description
```
Ephoniz:
```suggestion
base_description = super()._get_external_description(ical=ical)
    
    if ical:
        event_url = f'<div><a href="{self.event_register_url}">{self.name}</a></div><div style="white-space: pre-wrap;">{base_description}</div>'
    else:
        event_url = f'{self.name}: {self.event_register_url}\n'
    
    return event_url + base_description if not ical else event_url
```

Ephoniz (changes requested):
Hii, made a few suggestions 👍🏽 , did not checked if it works as intended, maybe it will be nice to have a small test for this. Also I'm not sure if we'd be able to add a new argument in stable.

addons/event/models/event_event.py
```diff
@@ -730,7 +730,7 @@ def _get_date_range_str(self, lang_code=False):
             return _('next month')
         return _('on %(date)s', date=format_date(self.env, self.date_begin, lang_code=lang_code, date_format='medium'))
 
-    def _get_external_description(self):
+    def _get_external_description(self, ical=False):
```
lase-odoo: Hey ! @Ephoniz, @farahdawod this change is not stable. We should never change the method signature to avoid breaking custo's.

addons/event/models/event_event.py
```diff
@@ -763,6 +763,9 @@ def _get_ics_file(self):
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
             cal_event.add('description').value = event._get_external_description()
+            xalt = cal_event.add('x-alt-desc')
+            xalt.value = f'<div style="white-space: pre-wrap;"><a href="{event.event_register_url}">{event.name}</a></div><div>{event.description}</div>'
```
Ephoniz: Here, we could be extra safe and excape the `event.name` using for example odoo.tools.html_escape, probably we don't see any error since the platform also handles this cases but just to be safer.

Ephoniz (approved):
Seems better now 👍🏽 Left a small comment.

Also, I've noticed we have the same issue in the Apple calendar, however I did not seem to find why does it happen, might be a limitation 🤔 just a note for the record.

addons/event/models/event_event.py
farahdawod: I applied your changes, but `event.name` has a `not-null` constraint in the database anyway.

Ephoniz: Hey, what I meant was not for guarding not having an event name (which I think it can't happen), we would like to use html_escape in some filed when introducing them on html, to avoid having issues if the even name contain special characters as <>, & or others that might conflict with the html and hide some content. 

But from my tests, it seems that the platforms handle this and we won't have an issue most of the time, but it's also a safe way to avoid XSS injection 👍🏽

Ephoniz (approved):

addons/event/models/event_event.py
```diff
@@ -763,6 +763,9 @@ def _get_ics_file(self):
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
             cal_event.add('description').value = event._get_external_description()
+            xalt = cal_event.add('x-alt-desc')
```
reth-odoo: Would add a comment to specify this is used by outlook specifically
(on google calendar they just support html in standard description, like normal people)

```suggestion
            xalt = cal_event.add('X-ALT-DESC')
```

addons/website_event/tests/test_event_internals.py
```diff
@@ -237,3 +238,31 @@ def test_website_visibility_public_user(self):
             self.assertIn(self.event_public, visible_events)
             self.assertIn(self.event_link_only, visible_events, "Should now be visible because visitor is participating")
             self.assertNotIn(self.event_logged_users, visible_events)
+
+    @users('public_test')
+    def test_ics_file_renders_correctly(self):
```
reth-odoo: 🔤 

Also not really a "render", maybe test_ics_file_generation or something and we can extend it later as needed as a generic "event ics test"

addons/website_event/tests/test_event_internals.py
```diff
@@ -237,3 +238,31 @@ def test_website_visibility_public_user(self):
...
+            'description': '<p>This is <b>HTML</b> description</p>',
+            'event_register_url': 'localhost:8069',
+        })
+
+        ics_map = event._get_ics_file()
+        self.assertIn(event.id, ics_map)
+        ics_bytes = ics_map[event.id]
+        self.assertIsInstance(ics_bytes, bytes)
+        ics_str = ics_bytes.decode('utf-8')
+        self.assertIn('BEGIN:VCALENDAR', ics_str)
```
reth-odoo: We can assume the ICS is valid, if not the test will still crash on readOne or other asserts.

reth-odoo (commented):
Indeed I suppose it's always fine to pass an `x-` parameter
https://icalendar.org/iCalendar-RFC-5545/3-8-8-2-non-standard-properties.html

It bloats the ics file a bit in cases where it's not required but it's otherwise more stable-friendly than
https://github.com/odoo/odoo/pull/236470

addons/event/models/event_event.py
```diff
@@ -763,6 +763,9 @@ def _get_ics_file(self):
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
             cal_event.add('description').value = event._get_external_description()
+            xalt = cal_event.add('X-ALT-DESC')  # Particularly used by outlook to support appropriate formatting from html to text
+            xalt.value = f'<div style="white-space: pre-wrap;"><a href="{event.event_register_url}">{html_escape(event.name)}</a></div><div>{event.description}</div>'
```
reth-odoo: Why not use the description again here actually? (_get_external_description)

We don't really want to support the description "as html" here anyways if that's a reason, we have a (slightly rotten..) task replacing it with a text field, as it being potentially very long or shortened arbitrarily is problematic.

addons/website_event/tests/test_event_internals.py
```diff
@@ -237,3 +238,28 @@ def test_website_visibility_public_user(self):
             self.assertIn(self.event_public, visible_events)
             self.assertIn(self.event_link_only, visible_events, "Should now be visible because visitor is participating")
             self.assertNotIn(self.event_logged_users, visible_events)
+
+    @users('public_test')
+    def test_ics_file_generation(self):
```
reth-odoo: Keep tests alphabetically ordered

addons/website_event/tests/test_event_internals.py
```diff
@@ -237,3 +238,28 @@ def test_website_visibility_public_user(self):
...
+        ics_bytes = ics_map[event.id]
+        ics_str = ics_bytes.decode('utf-8')
+
+        cal = vobject.readOne(ics_str)
+        vevent = cal.vevent
+
+        self.assertEqual(vevent.description.value, event._get_external_description())
+        self.assertIn('x-alt-desc', vevent.contents)
+        xalt = vevent.contents['x-alt-desc'][0]
+        self.assertEqual(xalt.params.get('FMTTYPE'), ['text/html'])
```
reth-odoo: Maybe assertIn some reasonable string to make sure we do include the description here.

addons/event/models/event_event.py
farahdawod: `_get_external_description()` already hooks the `event_register_url` to the `event.name` using `<a href>` , which don't work for ics files. To use `_get_external_description()` here, I'd have to append the appropriate ics link embedding in the title to the `_get_external_description()` after dismissing the `<a href>` from it, which is more costly than this option imo.

addons/website_event/tests/test_event_internals.py
farahdawod: Can you clarify what would count as reasonable string here?

reth-odoo: "This is \<b\>HTML\</b\> description" would be nice if possible in one line

Ephoniz: I would say anything within `<p>This is <b>HTML</b> description</p>` since it's the description we have for the test.

addons/event/models/event_event.py
reth-odoo: >  already hooks the event_register_url to the event.name using <a href> , which don't work for ics files.

But here we also use `event.name` and `<a href>` so not sure what part you're saying does not work in the original

addons/website_event/models/event_event.py
```diff
@@ -464,7 +464,7 @@ def _track_subtype(self, init_values):
     def _get_external_description(self):
         """ Adding the URL of the event into the description """
         self.ensure_one()
-        event_url = f'<a href="{self.event_register_url}">{self.name}</a>'
+        event_url = f'<div style="white-space: pre-wrap;"><a href="{self.event_register_url}">{self.name}</a><div>'
```
reth-odoo: Do we need pre-wrap? I think on apple-calendar they also show the raw html (no workaround there afaict) so it would be best to keep the additional html to a minimum.

addons/event/models/event_event.py
```diff
@@ -762,7 +762,11 @@ def _get_ics_file(self):
             cal_event.add('dtstart').value = event.date_begin.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('dtend').value = event.date_end.astimezone(pytz.timezone(event.date_tz))
             cal_event.add('summary').value = event.name
-            cal_event.add('description').value = event._get_external_description()
+            external_description = event._get_external_description()
+            cal_event.add('description').value = external_description
```
reth-odoo, line 766: I think it would make sense to html2plaintext this actually (but maybe that's a separate fix, for later versions, as it potentially impacts existing events.)

For google, we use a direct url instead of an ics so it's not relevant. Microsoft does not seem to use the description at all when x-alt-desc is set anyways. And apple does not seem to support html in any form (sadly I could not test that 🧱 🥀 🧱 )

reth-odoo (commented):
This seems to work fine on both gmail and microsoft. Sadly I don't think it will work on apple calendar from some investigation. Would say we probably should call html2plaintext if we want to be thorough, but maybe we only do that from 19 (or master) to be safe 🤔 

Pinging @Odoonan already as it otherwise seems fine, and for opinion 👀

addons/website_event/models/event_event.py
Odoonan: I agree, not sure what we are doing here tbh :) I feel like this is not needed? When looking in details on the outlook meeting description of a given event, we have a structure like this, not sure how it translates this way though.
```xml
<div class style="white-space:pre-wrap">
  <a> ... </a>
  <div> ... rest of the description ... </div>
</div>
```
and we end up with strange spacing inside the description. Could we try without it ? :)

addons/event/models/event_event.py
Odoonan: I cannot either, did you check this was just ignored in Apple Calendar ? :) e.g. not breaking anything?

addons/website_event/tests/test_event_internals.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+import vobject
```
Odoonan: Note that this is always in a try catch block from what I see in other files. Probably add one and a SkipTest in the test if not defined. Feel free to check existing occurences

Odoonan: Also, please test locally the test is correctly skipped when vobject is None

addons/website_event/models/event_event.py
Odoonan: Ah yes, you do not close the div actually! `</div>`
Is the div tag needed altogether btw? 🤔

addons/website_event/tests/test_event_internals.py
```diff
@@ -23,6 +24,32 @@ def setUpClass(cls):
         } for website_visibility in ['public', 'link', 'logged_users']])
         cls.events_visibility_test = cls.event_public | cls.event_link_only | cls.event_logged_users
 
+    @users('public_test')
+    def test_ics_file_generation(self):
```
Odoonan: `test_ics_file_html_description`

addons/website_event/tests/test_event_internals.py
```diff
@@ -23,6 +24,32 @@ def setUpClass(cls):
...
+        """Verify that _get_ics_file returns a valid .ics description
+        that will be rendered correctly for the user.
+        """
+        event = self.event_public
+        event.write({
+            'date_begin': FieldsDatetime.to_string(datetime(2022, 12, 31, 10, 0, 0)),
+            'date_end': FieldsDatetime.to_string(datetime(2022, 12, 31, 12, 0, 0)),
+            'description': '<p>This is <b>HTML</b> description</p>',
+            'event_register_url': 'localhost:8069',
```
Odoonan: https for good measure? :)

Odoonan (commented):
Hi @farahdawod 

Sorry for the delay! Here are a few comments and questions.

Cheers!

addons/website_event/models/event_event.py
farahdawod: > Ah yes, you do not close the div actually! `</div>` Is the div tag needed altogether btw? 🤔

I edited that locally a while ago and removed the `pre-wrap`, but, in general, yes we need the `<div></div>` to be able to render the link inside the event name. Please refer to the attached photo to see the resulting invitation.
ics_res

addons/event/models/event_event.py
farahdawod: apple_ics_screenshot
yes. this is how it looks in apple calendar after applying the changes

addons/website_event/models/event_event.py
```diff
@@ -464,7 +464,7 @@ def _track_subtype(self, init_values):
     def _get_external_description(self):
         """ Adding the URL of the event into the description """
         self.ensure_one()
-        event_url = f'<a href="{self.event_register_url}">{self.name}</a>'
+        event_url = f'<div><a href="{self.event_register_url}">{self.name}</a></div>'
```
Odoonan: Hi @farahdawod 

I would probably remove the div, unless I missed something:

First, because it adds an empty line in google. (and apple calendar?)
Second, because the link will be rendered correctly on microsoft, but will just be inlined, right? Good enough I would say.
Also, to keep it as an inline element and the diff minimal :) outlook seems to change the line break below to a white space so it should be fine?

Can we give it a try please? 🙏 Thanks for the updates!

Nitpicking on the commit message 🥸 
- Can you specify the fix is mainly for outlook calendar?
- Does it really happen _because HTML to text conversion for the part where the is link embedded inside the event title isn't compatible with iCal._ ? It seems to happen because outlook uses the description as text and displays it as is in the ics file by default? :) I would reword a bit.
- Do not forget in your commit message to have at most 80 / 100 characters per line
- Can you add an empty line after your title ? runbot has a strange commit titile preview 🤷 

Cheers

farahdawod: @Odoonan all done :)

addons/website_event/tests/test_event_internals.py
```diff
@@ -23,6 +33,35 @@ def setUpClass(cls):
...
+        ics_str = ics_bytes.decode('utf-8')
+
+        cal = vobject.readOne(ics_str)
+        vevent = cal.vevent
+
+        self.assertEqual(vevent.description.value, event._get_external_description())
+        self.assertIn('x-alt-desc', vevent.contents)
+        xalt = vevent.contents['x-alt-desc'][0]
+        self.assertIn('This is HTML description', xalt.value)
+        self.assertEqual(xalt.params.get('FMTTYPE'), ['text/html'])
```
Odoonan: As descriptions should be the same, small suggestion:

```py
external_description = event._get_external_description()
self.assertIn('This is HTML description', external_description)
self.assertEqual(vevent.description.value, external_description)

self.assertIn('x-alt-desc', vevent.contents)
xalt = vevent.contents['x-alt-desc'][0]
self.assertEqual(xalt.params.get('FMTTYPE'), ['text/html'])
self.assertEqual(xalt.value, external_description)
```

addons/website_event/tests/test_event_internals.py
```diff
@@ -10,6 +12,14 @@
 from odoo.addons.website_event.controllers.main import WebsiteEventController
 from odoo.addons.event.tests.common import EventCase
 
+_logger = logging.getLogger(__name__)
+
+try:
+    import vobject
+except ImportError:
+    _logger.warning("`vobject` Python module not found, vcard file generation disabled. Consider installing this module if you want to generate vcard files")
```
Odoonan: Hi @farahdawod 

Just a typo here, it should be iCal File Generation and not vcard file generation (same at the end of the warning msg)

Regarding the test, note that we do not strictly check that the tags are (not) displayed for the link, since it depends on the client (e.g. outlook) but check values instead. However it checks values for description (and its tags being removed) which I think is not covered yet, so let us keep it.

However, see my remark below. Then we are good to go, thx! 🚀

PR #221906, opened 2025-08-05
[IMP] mail: search messages and activities with access using SQL + remove fetch override

kmagusiak:
When searching for messages or activities, perform the access checks in SQL.
We can translate rules that have restriction on model values directly into a Domain.

We keep the fallback to search without access rights and filter afterwards for other cases. We will always limit the number of returned results from the query to avoid checking the whole table.

The goal is to remove the hack (sudo in fetch) and align access and permission checks in both `_search` and `_check_access`.

https://github.com/odoo/enterprise/pull/98774
task-5238007

PR #246609, opened 2026-01-30
[FIX] mail: filter inaccessible records

jjmaksoud:
Mail messages with the wrong access should be filtered early before reading mapped fields to check for the records existence.
Filtering existing records can be applied on the set of access filtered records.

How to reproduce:
Create a multi company env
Edit an account in one of the companies
Switch companies
Go to the Audit Trail view

The issue was detected in the upgrade-ci

---

### DIFF
--- a/addons/mail/models/mail_message.py
+++ b/addons/mail/models/mail_message.py
@@ -407,6 +407,7 @@ class MailMessage(models.Model):
             forbidden_doc_ids = set()
             try:
                 operation_result = records._check_access(record_operation)
+                records = records.filtered(lambda r: r.id not in operation_result[0].ids if operation_result else True)
                 # Check that records really exist;
                 # lengthy way of checking first in cache and avoiding exists() call.
                 # see test_record_unlinked_orphan_activities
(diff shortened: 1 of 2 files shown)


kmagusiak (changes requested):
What do you mean it is not used?
`forbidden_doc_ids |= set((operation_result ...`

Your description does not even provide what error you got and you have no tests to cover the issue. In any case, I plan to refactor this for the next saas (https://github.com/odoo/odoo/pull/221906).

addons/mail/models/mail_message.py
```diff
@@ -420,15 +421,8 @@ def _filter_records_for_message_operation(self, doc_model, doc_res_ids, operatio
...
-                existing = records.exists()
-                forbidden_doc_ids = set((records - existing).ids)
-                operation_result = existing._check_access(record_operation)
-            forbidden_doc_ids |= set((operation_result or [self.env[doc_model]])[0]._ids)
-            # keep actually returned records for the opration, that are not forbidden
-            allowed_ids += [
-                record.id for record in records
-                if record.id not in forbidden_doc_ids
-            ]
+                records = records.exists()
```
kmagusiak, line 426: This is wrong because if you have a missing error during `_check_access` above, then you just take all records and return existing ones without checking if you have access.
The overall logic was (and should be): try to check access and if records exist to find allowed ones, when some don't exist, filter them and check access.

tde-banana-odoo:
Made internal task for followup, as we are in freeze period, better avoid forgetting it ^^ .

addons/mail/models/mail_message.py
jjmaksoud: As far as I can see, `_check_access`[ will also filter ](https://github.com/odoo/odoo/blob/b71f81ef70e58a1d59fe07ed9d2ecd038049838e/odoo/orm/models.py#L3479-L3485) inexisting records if there is a domain to check.
cases:
1- There is a missing record and no access error: `_check_access` will be False before and after the call to `exists()`
2- There is an access error and no missing record: the results of the first `_check_access` will filter the bad records
3- There is a missing record and an access error: the results of the first `_check_access` will filter the inaccessible and the non-existing records

jjmaksoud: But if https://github.com/odoo/odoo/pull/221906 is for this freeze then this whole method is deleted

kmagusiak: `_check_access` does not filter inexisting records, it crashes when it tries to filter them using record rules.
If you have a MissingError during `_check_access`, then you did not check any access.

PR #246610, opened 2026-01-30
[IMP] account: Label at cash basis base journal items

leml-odoo:
To have a more explicit reference from which documents a cash basis base line was created, now we put the name of the two journal entries that triggered the caba.

target: 18.0 -> master
Task [link](https://www.odoo.com/odoo/project.task/5883344)
task-5883344

### DIFF
--- a/addons/account/models/account_partial_reconcile.py
+++ b/addons/account/models/account_partial_reconcile.py
@@ -223,7 +223,8 @@ class AccountPartialReconcile(models.Model):
             return {}
 
         for partial in self:
-            for move in {partial.debit_move_id.move_id, partial.credit_move_id.move_id}:
+            for field, counterpart_field in (('debit', 'credit'), ('credit', 'debit')):
+                move, counterpart_move = partial[f'{field}_move_id'].move_id, partial[f'{counterpart_field}_move_id'].move_id
 
                 # Collect data about cash basis.
                 if move.id in tax_cash_basis_values_per_move:
@@ -312,6 +313,7 @@ class AccountPartialReconcile(models.Model):
                     'partial': partial,
                     'percentage': percentage,
                     'payment_rate': payment_rate,
+                    'counterpart_move': counterpart_move,
                 }
 
                 # Add partials.
@@ -559,6 +561,8 @@ class AccountPartialReconcile(models.Model):
                         # Base line.
 
                         cb_line_vals = self._prepare_cash_basis_base_line_vals(line, balance, amount_currency)
+                        cb_line_vals['name'] = ' - '.join(filter(None, (line.move_id.name, partial_values['counterpart_move'].name)))
+
                         grouping_key = self._get_cash_basis_base_line_grouping_key_from_vals(cb_line_vals)
 
                     if grouping_key in partial_lines_to_create:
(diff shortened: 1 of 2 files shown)


Megaaaaaa (changes requested):
Hello hello, :wave:

Thank you for the PR, here's a review for you :wink:

As a generic comment for the PR, I noticed that the tests now assert exact name values on many lines that aren’t affected by this change, especially the repeated `'name': False`. That makes them more fragile, since unrelated changes to tax-line labels could break them. I would keep strict name checks only on the base lines, which are the actual scope of this change. So every line that is not on the cash basis base account would avoid asserting on `name` and only check the relevant accounting values.

Please do not hesitate to tell me if I missed something or if you don't agree with something :smile:

addons/account/models/account_partial_reconcile.py
```diff
@@ -559,6 +561,9 @@ def _create_tax_cash_basis_moves(self):
                         # Base line.
 
                         cb_line_vals = self._prepare_cash_basis_base_line_vals(line, balance, amount_currency)
+                        if partial_values['counterpart_move'].name:
+                            cb_line_vals['name'] += f" - {partial_values['counterpart_move'].name}"
```
Megaaaaaa: This assumes `_prepare_cash_basis_base_line_vals()` always sets a non-empty string in name, but it’s common for name to be `False`. In that case, this would crash when the counterpart move has a name. What I would do instead is build the label explicitly from the two move names so it doesn’t rely on the existing value of `cb_line_vals['name']`.

```suggestion
                            cb_line_vals['name'] = ' - '.join(filter(None, (line.move_id.name, partial_values['counterpart_move'].name)))
```

leml-odoo: Nice catch and thanks for the suggestion! I didn't consider that.

addons/account/models/account_partial_reconcile.py
```diff
@@ -559,6 +561,9 @@ def _create_tax_cash_basis_moves(self):
                         # Base line.
 
                         cb_line_vals = self._prepare_cash_basis_base_line_vals(line, balance, amount_currency)
+                        if partial_values['counterpart_move'].name:
+                            cb_line_vals['name'] = ' - '.join(filter(None, (line.move_id.name, partial_values['counterpart_move'].name)))
```
Megaaaaaa: This condition makes the final label depend on whether the counterpart move has a name, so in some cases we keep the previous value from `_prepare_cash_basis_base_line_vals()`. Wouldn't it be better to always compute the label from both move names and let `filter(None, ...)` handle missing values.
```suggestion
                        cb_line_vals['name'] = ' - '.join(filter(None, (line.move_id.name, partial_values['counterpart_move'].name)))
```

Megaaaaaa (commented):
Hello :wave:

Kind of both a change and a question at the same time, after that we're good to go :+1:

addons/account/models/account_partial_reconcile.py
leml-odoo: Originally it was to not affect the caba line name unless the counterpart move has a name, but with the `filter(None, ...)` like you said, we pretty much handle it, so I think your suggestion is better overall. I just pushed my changes, let me know if you want me to squash it.

Megaaaaaa: Yeah you can squash, it seems ready to me :+1: Thank you for the changes :smile:

leml-odoo: Done! Thanks for your review 😄!

Megaaaaaa (approved):
Thank you for the changes :+1:

robodoo r+

PR #246629, opened 2026-01-30
[FIX] website_event_sale: Prevent auto registration

zeche-odoo:
**Before this commit**
If a gift card balance fully covers a shopping cart containing multiple events, Odoo auto-confirms the order as soon as the last event is added, skipping the final checkout step.

**After this commit**
Sale orders will no longer be automatically confirmed when a customer registers for a paid event, even if an applied gift card brings the total balance to zero.

opw-5896626

### DIFF
--- a/addons/website_event_sale/controllers/main.py
+++ b/addons/website_event_sale/controllers/main.py
@@ -3,6 +3,7 @@
 
 from collections import defaultdict
 from odoo.http import request, route
+from odoo.tools import float_is_zero
 
 from odoo.addons.website_event.controllers.main import WebsiteEventController
 
@@ -75,12 +76,11 @@ class WebsiteEventSaleController(WebsiteEventController):
 
         # we have at least one registration linked to a ticket -> sale mode activate
         if any(info['event_ticket_id'] for info in registrations):
-            if order_sudo.amount_total:
-                request.session['sale_last_order_id'] = order_sudo.id
-                return request.redirect("/shop/checkout")
             # free tickets -> order with amount = 0: auto-confirm, no checkout
-            elif order_sudo:
+            if all(line.event_ticket_id and float_is_zero(line.event_ticket_id.price, precision_digits=2) for line in order_sudo.order_line):
                 order_sudo.action_confirm()  # tde notsure: email sending ?
                 request.website.sale_reset()
-
+            elif order_sudo:
+                request.session['sale_last_order_id'] = order_sudo.id
+                return request.redirect("/shop/checkout")
         return res
--- a/addons/website_event_sale/tests/test_website_event_sale.py
+++ b/addons/website_event_sale/tests/test_website_event_sale.py
@@ -61,3 +61,63 @@ class TestWebsiteEventSale(HttpCaseWithUserPortal, TestWebsiteEventSaleCommon):
             ('order_line.event_ticket_id', '=', self.ticket.id),
             ('order_line.event_ticket_id', '=', free_ticket.id)
         ]), "Sale order should be created for the free/paid tickets mix")
+
+    def test_website_event_sale_giftcard_covers_full_cost(self):
+        """Test saleorder is not auto confirmed if the gift card in the card can fully cover the cost"""
+        self.authenticate(None, None)
+        event_questions = self.event.question_ids
+        event_registration_count = len(self.event.registration_ids)
+        name_question = event_questions.filtered(lambda q: q.question_type == 'name')
+        email_question = event_questions.filtered(lambda q: q.question_type == 'email')
+        phone_question = event_questions.filtered(lambda q: q.question_type == 'phone')
+
+        paid_ticket_1 = self.env['event.event.ticket'].create({
+            'event_id': self.event.id,
+            'name': 'Paid_1',
+            'product_id': self.product_event.id,
+            'price': 50,
+        })
+
+        paid_ticket_2 = self.env['event.event.ticket'].create({
+            'event_id': self.event.id,
+            'name': 'Paid_2',
+            'product_id': self.product_event.id,
+            'price': 200,
+        })
+
+        self.pricelist.write({
+                    'item_ids': [(5, 0, 0), (0, 0, {
+                        'applied_on': '3_global',
+                        'compute_price': 'percentage',
+                        'percent_price': 100,
+                    })],
+                    'name': 'Full discount',
+        })
+        # There must be an event in the cart to apply the pricelist on it
+        self.url_open(f'/event/{self.event.id}/registration/confirm', data={
+            f'1-name-{name_question.id}': 'Bob',
+            f'1-email-{email_question.id}': 'bob@test.lan',
+            f'1-phone-{phone_question.id}': '8989898989',
+            '1-event_ticket_id': paid_ticket_1.id,
+            'csrf_token': http.Request.csrf_token(self),
+        })
+
+        self.url_open('/shop/change_pricelist', data={
+            'pricelist_id': self.pricelist.id,
+            'csrf_token': http.Request.csrf_token(self),
+        })
+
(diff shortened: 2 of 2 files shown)


addons/website_event_sale/tests/test_website_event_sale.py
```diff
@@ -61,3 +61,41 @@ def test_website_event_sale_free_paid_mix(self):
...
+        email_question = event_questions.filtered(lambda q: q.question_type == 'email')
+        phone_question = event_questions.filtered(lambda q: q.question_type == 'phone')
+
+        self.url_open(f'/event/{self.event.id}/registration/confirm', data={
+            f'1-name-{name_question.id}': 'Bob',
+            f'1-email-{email_question.id}': 'bob@test.lan',
+            f'1-phone-{phone_question.id}': '8989898989',
+            '1-event_ticket_id': self.ticket.id,
+            'csrf_token': http.Request.csrf_token(self),
+        })
```
reth-odoo, line 103: Seems useless to have the same registration twice (?)

addons/website_event_sale/controllers/main.py
```diff
@@ -78,9 +78,4 @@ def registration_confirm(self, event, **post):
             if order_sudo.amount_total:
                 request.session['sale_last_order_id'] = order_sudo.id
                 return request.redirect("/shop/checkout")
-            # free tickets -> order with amount = 0: auto-confirm, no checkout
-            elif order_sudo:
-                order_sudo.action_confirm()  # tde notsure: email sending ?
-                request.website.sale_reset()
```
reth-odoo, line 84: RIP auto-confirm

Shouldn't we check that there's an amount without discount instead?

tde-banana-odoo: Mmmh this seems to break a wanted flow, indeed.

addons/website_event_sale/tests/test_website_event_sale.py
zeche-odoo: The same registration shouldn't matter, it auto confirms as long as the gift card balance could cover them all. And yea, I don't think we want people to not stop by the cart when they have a paid event in it or they are adding a new paid event.

addons/website_event_sale/controllers/main.py
zeche-odoo: @reth-odoo @tde-banana-odoo 
The reason why this elif statement is not needed is because It doesn't have this if statement to check and auto confirm free events in 17. 
https://github.com/odoo/odoo/blob/97ae2ecbc0c903966f7444a8c69bfb599611f92f/addons/website_event_sale/controllers/main.py#L70-L72.

In 18+, it will handle the cases when the free events should be auto confirmed. Therefore, after this if statement, all the sale orders created should not be auto confirmed and should be redirect to the checkout page without checking the amount. 

Let me know if I'm wrong or there are edge cases.

reth-odoo: @zeche-odoo 
You're saying we don't need the branch because the registration state computation will handle it?

Unless there's a reason to remove it, I'd keep it just in case. You never know. Especially as we also call `sale_reset` which the compute definitely will not do.

zeche-odoo: @reth-odoo The reason why it needs to be removed because it's auto confirming and resetting a sale order that it's not supposed to be. Technically, we could check if there is amount with no discount but this change should achieve the same goal.

I've removed this line in my most recent push for this pr 

https://github.com/odoo/odoo/blob/2264f330859b79010b227e3a9fda1075de8ed4e8/addons/website_event_sale/controllers/main.py#L78

This line is not needed because registering a free event will not create a sale order unless there is a paid event or product in the cart while registering. Therefore, we should not confirm and reset any sale orders that are created after checking if the event is the only free event that is being registered at this point.

reth-odoo: > we could check if there is amount with no discount

So let's do that imo

18.0 is soon to be old stable, I'd rather not touch event sale flows at all unless needed. Maybe someone has a module where this behavior is relevant.

zeche-odoo: @reth-odoo Hi, I made a new push with a checking condition to check the amount with no discount. Let me know if there are any edge cases that might fail!

addons/website_event_sale/tests/test_website_event_sale.py
reth-odoo: @zeche-odoo 
No but I mean here you confirm this registration, then change the pricelist, then confirm a similar registration again and then check that we have two registration on a draft sale order.

Can't we skip the first registration? Is it relevant to the test to have 2?

addons/website_event_sale/controllers/main.py
```diff
@@ -75,12 +75,11 @@ def registration_confirm(self, event, **post):
 
         # we have at least one registration linked to a ticket -> sale mode activate
         if any(info['event_ticket_id'] for info in registrations):
-            if order_sudo.amount_total:
-                request.session['sale_last_order_id'] = order_sudo.id
-                return request.redirect("/shop/checkout")
             # free tickets -> order with amount = 0: auto-confirm, no checkout
-            elif order_sudo:
+            if order_sudo.amount_total - order_sudo.reward_amount == 0 and not order_sudo.pricelist_id:
```
reth-odoo: 🤔 seems like reward_amount only exists with sales_loyalty installed so this would not work

Maybe just say `all(line.event_ticket_id and line.price_total == 0 for line in order_sudo.line_ids)`
Or more conservatively `all(line.event_ticket_id and line.event_ticket_id.price == 0 for line in order_sudo.line_ids)`, but as we use price_total in other places that seems overly conservative.

addons/website_event_sale/tests/test_website_event_sale.py
zeche-odoo: @reth-odoo I see what you mean, we could skip the first registration in the test, the reason why I added it is because there is no way for the client to add a coupon in the website when their cart is empty so that it follows the actual workflow.

reth-odoo: @zeche-odoo 
Ah I see, then I would check that it isn't auto-confirmed there to be safe, and add a comment to say why we do it. And maybe make the registration values different just for "realism"

addons/website_event_sale/controllers/main.py
zeche-odoo: @reth-odoo The price_total of each line will show the cost after the discount. However, I found out that we can directly access the price of the event from their id so this should be conservative enough.

addons/website_event_sale/tests/test_website_event_sale.py
zeche-odoo: @reth-odoo I have made two different paid events with different prices and another check after applying the pricelist, this should be good.

addons/website_event_sale/controllers/main.py
```diff
@@ -75,12 +75,11 @@ def registration_confirm(self, event, **post):
 
         # we have at least one registration linked to a ticket -> sale mode activate
         if any(info['event_ticket_id'] for info in registrations):
-            if order_sudo.amount_total:
-                request.session['sale_last_order_id'] = order_sudo.id
-                return request.redirect("/shop/checkout")
             # free tickets -> order with amount = 0: auto-confirm, no checkout
-            elif order_sudo:
+            if all(line.event_ticket_id and line.event_ticket_id.price == 0 for line in order_sudo.order_line):
```
tde-banana-odoo: should probably use float comparisons tools

zeche-odoo: @tde-banana-odoo Fixed it

tde-banana-odoo:
@robodoo r+

tde-banana-odoo:
@robodoo r+

PR #246630, opened 2026-01-30
[FIX] pos_sale: fix payment provider creation in test

davidmonnom:
Depending on the modules installed, the tests sometimes use a different value for creating the `payment.provider` record. The code is sometimes set to custom if the `payment_custom` module is installed. This causes a runbot error.

To avoid this, we force the code `none` in our tests. This will have no impact and will fix the indeterminate error.

RunbotID: 237707


stva-odoo (commented):
robodoo r+

PR #246640, opened 2026-01-30
[IMP] orm: determine faster which fields are groupable

kmagusiak:
The default implementations have a a defined scope, they don't handle a lot of edge cases. If we have no overwrite, shortcut.

```
# master
In [2]: %timeit self.fields_get(attributes=['groupable'])
2.02 ms ± 122 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
In [3]: %timeit self.fields_get()
8.86 ms ± 141 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# this PR
In [4]: %timeit self.fields_get(attributes=['groupable'])
515 μs ± 5.33 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
In [5]: %timeit self.fields_get()
6.76 ms ± 67.6 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

---

### DIFF
--- a/odoo/orm/fields.py
+++ b/odoo/orm/fields.py
@@ -964,6 +964,10 @@ class Field[T]:
         if self.inherited_field and self.inherited_field._description_groupable(env):
             # avoid compuation for inherited field
             return True
+        from .models import BaseModel  # noqa: PLC0415
+        if self.type != 'many2many' and env.registry[self.model_name]._read_group_groupby is BaseModel._read_group_groupby:
+            # the default implementation has an edge case for many2many
+            return False
 
         model = env[self.model_name]
         groupby = self.name if self.type not in ('date', 'datetime') else f"{self.name}:month"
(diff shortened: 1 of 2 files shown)


odoo/orm/fields.py
```diff
@@ -979,6 +982,9 @@ def _description_aggregator(self, env: Environment):
         if self.inherited_field and self.inherited_field._description_aggregator(env):
             # avoid compuation for inherited field
             return self.inherited_field.aggregator
+        if env.registry[self.model_name]._read_group_select is env.registry['base']._read_group_select:
```
rco-odoo: I would use `BaseModel` instead of `env.registry['base']`, since the latter may be overridden.

rco-odoo (approved):
@robodoo r+

PR #243509, opened 2026-01-13
[IMP] l10n_latam_check: Allow multiple liquidity lines for checks in Payment JE

sham-odoo:
Till now, if the payment was made using checks for LATAM countries, the payment journal entry that was created consisted only one liquidity line for the JE. This was not much suitable in case if payment is sent / recieved via multiple checks.

This commit intends to solve this issue by splitting the Payment JE liquidity line into multiple liquidity lines (one for each check) when payment is made through multiple checks.

To improve the payment synchronization with its journal entry, the `payment_method_line_id` field is also added in
`_get_trigger_fields_to_synchronize()` method so if payment method is updated from / to Checks, the JE is updated accordingly.

task-5207671

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246675, opened 2026-02-01
[FIX] account: Fix and Improve payment move syncronization

sham-odoo:
This PR improves payment move synchronization and fixes an error that is caused while changing amount of a Payment that has multiple liquidity lines.

Steps to reproduce:
1. Create payment with x amount and validate it.
2. Open the payment journal entry.
3. Reset to draft and update the liquidity line amount from x to (x - y)
4. Create another liquidity line with amount y to balance entry and post it.
5. Draft the payment and try to update the amount.

A traceback will appear. `ValueError: Expected singleton`

Cause:
The lines for payment JE are prepared for the case assuming that there will be only 1 liquidity line, but since we have more than 1, we get a Singleton error.

Description of changes made:
While preparing values for move in `synchronize_to_moves()` check for multiple liquidity lines and append all values to write.
Further, the `_prepare_move_line_default_vals()` is also improved in order to manage different type of move lines individually.

---

### DIFF
--- a/addons/account/models/account_payment.py
+++ b/addons/account/models/account_payment.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from itertools import zip_longest
 from odoo import models, fields, api, _, Command
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.misc import format_date, formatLang
@@ -261,6 +262,7 @@ class AccountPayment(models.Model):
         return lines
 
     def _get_valid_liquidity_accounts(self):
+        self.ensure_one()
         return (
             self.journal_id.default_account_id |
             self.payment_method_line_id.payment_account_id |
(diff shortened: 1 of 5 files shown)


addons/account/tests/test_account_payment.py
```diff
@@ -878,3 +878,41 @@ def test_payment_amount_without_move(self):
...
+        move.write({
+            'line_ids': [
+                Command.update(liquidity_lines.id, {'amount_currency': 100}),
+                Command.create({
+                    'move_id': move.id,
+                    'account_id': liquidity_lines.account_id.id,
+                    'debit': 50.0,
+                    'credit': 0.0,
+                    'amount_currency': 50.0,
+                    'currency_id': payment.currency_id.id,
```
smetl:
```suggestion
                    'account_id': liquidity_lines.account_id.id,
                    'balance': 50.0,
                    'amount_currency': 50.0,
                    'currency_id': payment.currency_id.id,
```

addons/account/models/account_payment.py
```diff
@@ -998,13 +999,38 @@ def _synchronize_to_moves(self, changed_fields):
                     'balance': sum(writeoff_lines.mapped('balance')),
                 })
             line_vals_list = pay._prepare_move_line_default_vals(write_off_line_vals=write_off_line_vals)
```
smetl: Create a sub method
`def _prepare_move_line_values_per_type(self, counterpart_account_id=None):`
doing exactly like `_prepare_move_line_default_vals` but returning a dict:
`{'liquidity_lines': [...], 'counterpart_lines': [...], 'write_off_lines': [...]}`
Then change `_prepare_move_line_default_vals` to
```py
def _prepare_move_line_values_per_type(self, counterpart_account_id=None):
     line_values_per_type = self._prepare_move_line_values_per_type(counterpart_account_id=counterpart_account_id)
    line_values = []
    for sub_line_values in line_values_per_type.values():
        line_values += sub_line_values
    return line_values
```
=> You won't need anymore to use checks about accounts to deduce the number of lines in each category

addons/account/models/account_payment.py
```diff
@@ -998,13 +999,38 @@ def _synchronize_to_moves(self, changed_fields):
...
+                for val in line_vals_list
+                if val["account_id"] == self.outstanding_account_id.id
+            ]
+            for liquidity_line, newline_val in zip_longest(liquidity_lines, liquidity_lines_vals):
+                if liquidity_line and newline_val:
+                    line_ids_commands.append(Command.update(liquidity_line.id, newline_val))
+                elif not liquidity_line and newline_val:
+                    line_ids_commands.append(Command.create(newline_val))
+                elif liquidity_line and not newline_val:
+                    line_ids_commands.append(Command.delete(liquidity_line.id))
```
smetl, line 1068: I don't like regrouping lines when changing the amount. Shouldn't we avoid the user to edit the payment amount in that case?
I'm concerned about what will happen if you create multiple liquidity lines with != accounts.

addons/account/models/account_payment.py
```diff
@@ -335,37 +358,32 @@ def _prepare_move_line_default_vals(self, write_off_line_vals=None, force_balanc
...
-                'partner_id': self.partner_id.id,
-                'account_id': self.destination_account_id.id,
-            },
-        ]
-        return line_vals_list + write_off_line_vals_list
+        return {
+            'liquidity_lines': self._prepare_move_liquidity_line_vals(line_name, liquidity_amount_currency, liquidity_balance),
+            'counterpart_lines': self._prepare_move_counterpart_line_vals(line_name, counterpart_amount_currency, counterpart_balance),
+            'writeoff_lines': write_off_line_vals_list,
+        }
```
sham-odoo, line 405: hello @smetl 
I've created sub functions for this as well since, in our related task for `l10n_latam_check` (#243509), we want to split the liquidity lines per check, so this can be just overridden to reduce other complexity and diff.
please let me know your thoughts on this

smetl: This is exactly what I had in mind :)

addons/account/models/account_payment.py
```diff
@@ -292,17 +293,39 @@ def _get_aml_default_display_name_list(self):
...
 
-    def _prepare_move_line_default_vals(self, write_off_line_vals=None, force_balance=None):
-        ''' Prepare the dictionary to create the default account.move.lines for the current payment.
-        :param write_off_line_vals: Optional list of dictionaries to create a write-off account.move.line easily containing:
-            * amount:       The amount to be added to the counterpart amount.
-            * name:         The label to set on the line.
-            * account_id:   The account on which create the write-off.
-        :param force_balance: Optional balance.
-        :return: A list of python dictionary to be passed to the account.move.line's 'create' method.
+    def _prepare_move_liquidity_line_vals(self, line_name, amount, balance):
```
smetl: Prefer passing the parameters using a dictionary. That way, it's more flexible if we need to update such code in stable version ;)

addons/account/models/account_payment.py
smetl: What do you think about raising here if the amount is changed and if they are multiple liquidity lines?

addons/account/models/account_payment.py
```diff
@@ -292,17 +293,41 @@ def _get_aml_default_display_name_list(self):
...
-        :param force_balance: Optional balance.
-        :return: A list of python dictionary to be passed to the account.move.line's 'create' method.
+    def _prepare_move_liquidity_line_vals(self, line_vals):
+        return [{
+            'name': line_vals['name'],
+            'date_maturity': self.date,
+            'amount_currency': line_vals['amount'],
+            'currency_id': self.currency_id.id,
+            'debit': line_vals['balance'] if line_vals['balance'] > 0.0 else 0.0,
+            'credit': -line_vals['balance'] if line_vals['balance'] < 0.0 else 0.0,
```
smetl:
```suggestion
            'balance': line_vals['balance'],
```

addons/account/models/account_payment.py
```diff
@@ -985,6 +1007,10 @@ def _synchronize_to_moves(self, changed_fields):
             if pay.move_id.state == 'posted':
                 continue
             liquidity_lines, counterpart_lines, writeoff_lines = pay._seek_for_lines()
+
+            if 'amount' in changed_fields and len(liquidity_lines) > 1:
+                raise UserError(_("You cannot change the amount of a payment with multiple liquidity lines."))
+
```
maq-adhoc, line 1045: Hi @sham-odoo , 
The change proposed by this PR is great. But these lines are going to cause us problems in our layer. Is it really necessary to block the change in amount? Wouldn't it cause problems if I deleted the payment and tried to modify the checks? Would I have to modify the accounting entry first and then the checks?

smetl (approved):
@robodoo r+
@robodoo fw=skipmerge

PR #246679, opened 2026-02-02
[IMP] website_payment: make payment methods description more generic 

snrav-odoo:
Before the commit:
The website Payment Method description listed Google Pay and Apple Pay,
which are only available with Stripe in Express Checkout.

After the commit:
Updated the website Payment Methods description to use a more generic list,
including UPI, Netbanking, PayNow, and Klarna.

task-5896040

### DIFF
--- a/addons/website_payment/views/res_config_settings_views.xml
+++ b/addons/website_payment/views/res_config_settings_views.xml
@@ -78,7 +78,7 @@
                     <setting
                         string="Configure Payment Methods"
                         id="website_payment_setting"
-                        help="Visa, Mastercard, Maestro, Google Pay, Apple Pay, etc. as well as recurring charges."
+                        help="Visa, Mastercard, Maestro, UPI, Netbanking, PayNow, Klarna, etc., as well as recurring charges."
                         documentation="/applications/finance/payment_providers.html"
                     >
                         <div class="content-group">


shrm-odoo (approved):
LGTM !

kcv-odoo (approved):
🦆 

Commit msg is wrong it should be website_payment instead website_sale and it's not `payment providers` setting/description, it's related to website and payment methods

otherwise LGTM

addons/website_payment/views/res_config_settings_views.xml
```diff
@@ -78,7 +78,7 @@
                     <setting
                         string="Configure Payment Methods"
                         id="website_payment_setting"
-                        help="Visa, Mastercard, Maestro, Google Pay, Apple Pay, etc. as well as recurring charges."
+                        help="Visa, Mastercard, Maestro, UPI, Netbanking, PayNow, Klarna etc. as well as recurring charges."
```
AntoineVDV: Nitpicking, but the correct English form is "<something>, etc." 🙈 
```suggestion
                        help="Visa, Mastercard, Maestro, UPI, Netbanking, PayNow, Klarna, etc., as well as recurring charges."
```
Notice the two new commas, not just the first one.

AntoineVDV (changes requested):

AntoineVDV (approved):

AntoineVDV:
@robodoo r+

PR #246685, opened 2026-02-02
[FIX] tools, test_orm: handle bytearray in magic mimetype detection

ushyme:
The recent mimetype refactor reintroduced a crash where `python-magic` fails on `bytearray` inputs with a `TypeError` (ctypes buffer mismatch).

Steps to reproduce:
- Have python-magic library installed
- Products > Import records
- Upload a file containing an URL as image_1920
- Test upload

The upload fails with error: `Could not retrieve URL: 
[image_1920: L1]: argument 2: TypeError: wrong type`

This restores the fix from 7f36e380c0b175d0046a19f49e572944a5b307d9 by ensuring `bytearray` inputs are converted to `bytes` before detection.

opw-5893097

### DIFF
--- a/odoo/tools/mimetypes.py
+++ b/odoo/tools/mimetypes.py
@@ -42,7 +42,8 @@ def _odoo_guess_mimetype(bin_data, default='application/octet-stream'):
 
 
 def _magic_guess_mimetype(bin_data):
-    mimetype = magic.from_buffer(bin_data, mime=True)
+    magic_data = bytes(bin_data[:MIMETYPE_HEAD_SIZE]) if isinstance(bin_data, bytearray) else bin_data
+    mimetype = magic.from_buffer(magic_data, mime=True)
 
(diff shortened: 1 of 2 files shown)


rugo-odoo (approved):
@robodoo r+

odoo/tools/mimetypes.py
```diff
@@ -42,7 +42,8 @@ def _odoo_guess_mimetype(bin_data, default='application/octet-stream'):
 
 
 def _magic_guess_mimetype(bin_data):
-    mimetype = magic.from_buffer(bin_data, mime=True)
+    magic_data = bytes(bin_data[:MIMETYPE_HEAD_SIZE]) if isinstance(bin_data, bytearray) else bin_data
```
kmagusiak, line 45: @rugo-odoo For master, can you update this to use the following, so it works for all buffers?
```suggestion
    magic_data = bytes(bin_data)
```

PR #246687, opened 2026-02-02
[FIX] mrp: replace non-deterministic tour

vandroogenbd:
Issue
-----
The tour added in commit 6927265 was failing in nightly. The attempted fix in 246bf1d resolved the first problem but raised some new non-deterministic issues.

-----
Errors 237956, 238469 & 238470


vandroogenbd:
@robodoo up to 19.

PR #246688, opened 2026-02-02
[FIX] account_payment: duplicata of SEPA Direct Debit

jedel-odoo:
When activating SEPA Payment Provider and checking SEPA in your settings,
the name of the two (different) method was the same. Rename the method line
associated with the method that comes from the provider by 'Online SEPA' to
avoid confusion.

task-5875979

---

### DIFF
--- a/addons/account_payment/models/payment_provider.py
+++ b/addons/account_payment/models/payment_provider.py
@@ -70,6 +70,8 @@ class PaymentProvider(models.Model):
             )
             if pay_method_line_same_code:
                 create_values['payment_account_id'] = pay_method_line_same_code.payment_account_id.id
+            if self._get_code() == 'sepa_direct_debit':
+                create_values['name'] = "Online SEPA"
             self.env['account.payment.method.line'].create(create_values)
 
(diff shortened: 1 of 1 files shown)


AntoineVDV (dismissed):
Hello, I understand the problem you want to fix, but I don't believe that this is the right approach.

The thing is that you address an `account.payment.method(.line?)` *backend* issue by renaming a `payment.provider` and, most importantly, a `payment.method` record, which is customer-facing (it appears in the online payment form) and should retain its publicly known name "SEPA Direct Debit".

Another approach that would still comply with the current spec would be to tweak the `payment.provider` <-> `account.payment.method.line` mapping and give a different name to the latter record, so that it would remove the confusion for backend Accounting users, while preserving the proper name of the online payment method.

Feel free to reach out to me for more context or to challenge my suggestion.

Cheers!

malb-odoo (commented):
Seems ok to me thanks for the pr ! Maybe we can let @AntoineVDV Check the diff to be sure

malb-odoo (approved):
Good to go theeen
@robodoo r+

PR #246691, opened 2026-02-02
[IMP] data_recycle: allow user to manual refresh

pash-odoo:
Before:
- User was not able to refresh deduplication, recycle, and field cleaning data for all rules at once.
- To refresh data, user had to open each rule and update records manually.

After:
- Added a button in the cog menu to refresh deduplication, recycle, and field cleaning data for all rules at once.

task-5864554

Current behavior before PR:

---

### DIFF
--- a/addons/data_recycle/models/data_recycle_model.py
+++ b/addons/data_recycle/models/data_recycle_model.py
@@ -218,3 +218,21 @@ class Data_RecycleModel(models.Model):
         if self.recycle_mode == 'manual':
             return self.open_records()
         return
+
+    def refresh_recycle_records(self):
+        """
+        Refresh recycled records and reopen the Data Recycle view.
+        Preserves the selected search panel filter by passing
+        `recycle_model_id` into the action context.
+        """
+        self.search([])._recycle_records(batch_commits=True)
+        recycle_model_id = self.env.context.get('recycle_model_id')
+        action = self.env["ir.actions.actions"]._for_xml_id("data_recycle.action_data_recycle_record")
+        context = action.get('context', {})
+        if isinstance(context, str):
+            context = ast.literal_eval(context)
+        if recycle_model_id:
+            context['searchpanel_default_recycle_model_id'] = recycle_model_id
+        action['context'] = context
+        action['target'] = 'main'
+        return action
--- a/addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
+++ b/addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
@@ -1,0 +1,47 @@
+import { DropdownItem } from "@web/core/dropdown/dropdown_item";
+import { registry } from "@web/core/registry";
+import { Component } from "@odoo/owl";
+import { useService } from "@web/core/utils/hooks";
+
+const cogMenuRegistry = registry.category("cogMenu");
+
+export class DataRefreshCogMenu extends Component {
+    static template = "data_recycle.DataRefreshCogMenu";
+    static components = { DropdownItem };
+    static props = {};
+
+    setup() {
+        this.action = useService("action");
+    }
+
+    _getDomainValue(fieldName) {
+        // The searchpanel adds the selected rule to the domain as
+        // ['fieldName', '=', <id>]. We extract that id from the
+        // current domain and pass it in the action context so that,
+        // after refresh, the same rule remains selected in seachpanel.
+        const domain = this.env.searchModel.domain;
+        return domain.find((d) => Array.isArray(d) && d[0] === fieldName && d[1] === "=")?.[2];
+    }
+
+    async refreshData() {
+        const ruleId = this._getDomainValue("recycle_model_id");;
+        return await this.action.doActionButton({
+            type: "object",
+            resModel: "data_recycle.model",
+            name: "refresh_recycle_records",
+            context: {
+                recycle_model_id: ruleId,
+            }
+        });
+    }
+}
+
+export const DataRefreshCogMenuItem = {
+    Component: DataRefreshCogMenu,
+    isDisplayed: ({ searchModel }) => {
+        return searchModel.resModel === "data_recycle.record";
+    },
+    groupNumber: 50,
+};
+
+cogMenuRegistry.add("data_refresh_menu", DataRefreshCogMenuItem, { sequence: 10 });
--- a/addons/data_recycle/__manifest__.py
+++ b/addons/data_recycle/__manifest__.py
@@ -18,6 +18,8 @@
     'application': True,
     'assets': {
         'web.assets_backend': [
+            'data_recycle/static/src/cog_menu/data_refresh_cog_menu.js',
+            'data_recycle/static/src/cog_menu/data_refresh_cog_menu.xml',
             'data_recycle/static/src/views/*.js',
             'data_recycle/static/src/views/*.xml',
         ],
--- a/addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.xml
+++ b/addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+    <t t-name="data_recycle.DataRefreshCogMenu">
+        <DropdownItem onSelected.bind="refreshData">
+            <i class="fa fa-refresh me-1"/>
+            Refresh
+        </DropdownItem>
+    </t>
+</templates>


addons/data_recycle/models/data_recycle_model.py
```diff
@@ -207,3 +218,11 @@ def action_recycle_records(self):
         if self.recycle_mode == 'manual':
             return self.open_records()
         return
+
+    def refresh_recycle_records(self):
+        self.search([])._recycle_records(batch_commits=True)
+        recycle_model_id = self.env.context.get('recycle_model_id')
+        action = self.env["ir.actions.actions"]._for_xml_id("data_recycle.action_data_recycle_record")
+        action['context'] = dict(ast.literal_eval(action.get('context')), searchpanel_default_recycle_model_id=recycle_model_id)
```
arj-odoo: don't we have a better solution than ast.literal_eval ?

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    async refreshData() {
+        const model = this.action.currentController?.props?.resModel;
+        const searchModel = this.env.searchModel;
+
+        if (model === "data_recycle.record") {
+            const rule_id = searchModel.domain.filter(Array.isArray).find(
+                ([field, operator]) => field === "recycle_model_id" && operator === "="
+            )?.[2] ?? null;
```
arj-odoo: This seems quite fragile. No comment. it seems quite obscure.

arj-odoo (changes requested):
Thanks for your work. We are inheriting a new module, we should make sure that it remains clear when it was already the case and that we improve the readability when we add functionalities. It will help the appropriation of the code base. :-)

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
pash-odoo: I’ve added a comment to clarify this. We extract the `recycle_model_id` from the domain because the search panel stores the selected rule there, and we pass it to the action context to preserve the selection after refresh.

addons/data_recycle/models/data_recycle_model.py
```diff
@@ -128,10 +128,24 @@ def _recycle_records(self, batch_commits=False):
...
+            # and IDs of records already in the recycle records (existing_ids).
+            current_ids = set(records_to_recycle.ids)
+            existing_ids = set(mapped_existing_records[recycle_model])
+            # Remove recycle records for records that no longer match the recycle rule.
+            ids_to_remove = existing_ids - current_ids
+
+            if ids_to_remove:
+                self.env['data_recycle.record'].search([
+                    ('recycle_model_id', '=', recycle_model.id),
+                    ('res_id', 'in', list(ids_to_remove))
```
geflx: Doesn't need to be casted:
```suggestion
                    ('res_id', 'in', ids_to_remove)
```

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
```diff
@@ -0,0 +1,50 @@
...
+
+    setup() {
+        this.action = useService("action");
+    }
+
+    async refreshData() {
+        const model = this.action.currentController?.props?.resModel;
+        const searchModel = this.env.searchModel;
+
+        if (model === "data_recycle.record") {
```
geflx: No need to check model because isDisplayed is doing the job for us already (hiding it).

```suggestion
        const searchModel = this.env.searchModel;
```

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
```diff
@@ -0,0 +1,50 @@
...
+            // The searchpanel adds the selected rule to the domain as
+            // ['recycle_model_id', '=', <id>]. We extract that id from the
+            // current domain and pass it in the action context so that,
+            // after refresh, the same rule remains selected in seachpanel.
+            const rule_id = searchModel.domain.filter(Array.isArray).find(
+                ([field, operator]) => field === "recycle_model_id" && operator === "="
+            )?.[2] ?? null;
```
geflx: A bit faster:
```suggestion
            const ruleId = searchModel.domain.find(d) => Array.isArray(d) && d[0] === "recycle_model_id" && d[1] === "=")?.[2];
```

Also, in js prefer using `ruleId` instead of `rule_id`.

geflx (changes requested):
Thanks! You're in the right path 👏 
Some points added:

addons/data_recycle/models/data_recycle_model.py
```diff
@@ -207,3 +221,11 @@ def action_recycle_records(self):
         if self.recycle_mode == 'manual':
             return self.open_records()
         return
+
+    def refresh_recycle_records(self):
+        self.search([])._recycle_records(batch_commits=True)
+        recycle_model_id = self.env.context.get('recycle_model_id')
+        action = self.env["ir.actions.actions"]._for_xml_id("data_recycle.action_data_recycle_record")
+        action['context'] = dict(ast.literal_eval(action.get('context')), searchpanel_default_recycle_model_id=recycle_model_id)
```
geflx: Simpler and safer:
```suggestion
        action_context = {}
        if action.get('context') :
                safe_eval(action.get('context', '{}'))
        if recycle_model_id:
                action_context['searchpanel_default_recycle_model_id'] = recycle_model_id
        action['context'] = action_context
```

geflx (changes requested):

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
pash-odoo: I originally added this condition because we are patching this component in `data_cleaning` and calling `super.refreshData()` from there.
Now, I’ve added a return statement in the patch method when a certain condition is met. Because of that, for the `data_recycle.record model`, `the super.refreshData()` method will only be called if the model is not `data_recycle.record`.
So this extra condition is no longer needed and can be removed. Thanks.

addons/data_recycle/models/data_recycle_model.py
pash-odoo: Thanks for the suggestion.
I noticed that in this version, we are calling `safe_eval(action.get('context', '{}'))`, but we are not storing or merging its result into `action_context`. Because of that, the existing context gets ignored and action['context'] is always replaced with a new empty dictionary except for `searchpanel_default_recycle_model_id` when present.

To preserve the existing context and just update it, we could instead do something like:
```py
context = action.get('context', {})
if isinstance(context, str):
    context = safe_eval(context)
if recycle_model_id:
    context['searchpanel_default_recycle_model_id'] = recycle_model_id
action['context'] = context
```

geflx: I forgot in this one to add the context saving due the Ctrl+C Ctrl+V I did in the three different versions. Thanks for adding it :smile:

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    async refreshData() {
+        const searchModel = this.env.searchModel;
+
+        // The searchpanel adds the selected rule to the domain as
+        // ['recycle_model_id', '=', <id>]. We extract that id from the
+        // current domain and pass it in the action context so that,
+        // after refresh, the same rule remains selected in seachpanel.
+        const ruleId = searchModel.domain.find((d) => Array.isArray(d) && d[0] === "recycle_model_id" && d[1] === "=")?.[2];
```
geflx: Is it possible that this rule doesn't exist? Due to the `?`. Because in any way you're returning it as the `recycle_model_id` in the `context` key.
```suggestion
        const ruleId = searchModel.domain.find((d) => Array.isArray(d) && d[0] === "recycle_model_id" && d[1] === "=").[2];
```

geflx (commented):
I have a small question in this Community PR:

addons/data_recycle/static/src/cog_menu/data_refresh_cog_menu.js
pash-odoo: The rule may not always exist in the domain when 'All' option is selected in search panel. In that case `ruleId` will be `undefined`, and the backend method only adds `searchpanel_default_recycle_model_id` to the context if `recycle_model_id` is present. So passing undefined here is safe and simply results in no rule being preselected after refresh.

addons/data_recycle/models/data_recycle_model.py
```diff
@@ -207,3 +208,16 @@ def action_recycle_records(self):
         if self.recycle_mode == 'manual':
             return self.open_records()
         return
+
+    def refresh_recycle_records(self):
```
geflx, line 222: A small docstring would be welcome in this method.

geflx (approved):
robodoo r+
Thanks for the feature! 😉

PR #246692, opened 2026-02-02
[FIX] point_of_sale: fix combo last line price assignment

aksi-odoo:
Step to reproduce:
- create two products with 0 price, A and B
- create combo choice 'combo' with A and B, with max 2 items , free 2 items
- create a combo product "test combo" with this combo choice, set price=40
- start pos and select this product
- in combo product selector dialog, select same product 2 time and confirm

Observation:
- the total price is 80 (double then expected) in product screen

Cause:
- Remaining combo amount is assigned to the last line unit price.
- For zero-price items, remaining amount equals combo price.
- as `Total = qty × unit price`, as qty =2 it duplicates the combo price.

Fix:
- Since we assign unit price, divide remaining amount by combo qty.

opw-5868831

---

### DIFF
--- a/addons/point_of_sale/static/src/app/models/utils/compute_combo_items.js
+++ b/addons/point_of_sale/static/src/app/models/utils/compute_combo_items.js
@@ -27,8 +27,11 @@ export const computeComboItems = (
         let priceUnit = ProductPrice.round((combo.base_price * parentLstPrice) / originalTotal);
         remainingTotal -= priceUnit * conf.qty;
 
-        if (comboItem.id == childLineConf[childLineConf.length - 1].combo_item_id.id) {
-            priceUnit += remainingTotal;
+        if (
+            remainingTotal &&
+            comboItem.id == childLineConf[childLineConf.length - 1].combo_item_id.id
+        ) {
+            priceUnit += ProductPrice.round(remainingTotal / conf.qty);
         }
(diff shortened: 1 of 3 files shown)


alsh-odoo (approved):

addons/point_of_sale/static/src/app/models/utils/compute_combo_items.js
```diff
@@ -27,8 +27,11 @@ export const computeComboItems = (
         let priceUnit = ProductPrice.round((combo.base_price * parentLstPrice) / originalTotal);
         remainingTotal -= priceUnit * conf.qty;
 
-        if (comboItem.id == childLineConf[childLineConf.length - 1].combo_item_id.id) {
-            priceUnit += remainingTotal;
+        if (
+            remainingTotal &&
+            comboItem.id == childLineConf[childLineConf.length - 1].combo_item_id.id
+        ) {
+            priceUnit += ProductPrice.round(remainingTotal / conf.qty);
```
robinengels, line 34: Instead of redoing a computation and applying a second rounding, couldn't we just adapt the remainingTotal computation if we are on the last line ? I'm afraid we will have error with rounding in the future.

aksi-odoo:
issue fixed by https://github.com/odoo/odoo/commit/a7fd68d8cd0e897ad16effc2ef49ecc6a0b010bf

PR #246693, opened 2026-02-02
[FIX] website: TestWebsiteBlogUi.test_blog_access_rights tour

pipu-odoo:
With this commit, we fix the tour by adding a step that ensure the menu is well rendered before to destroy it.

runbot-error-id~238489

Current behavior before PR:

---

### DIFF
--- a/addons/website_blog/static/tests/tours/blog_manager_tour.js
+++ b/addons/website_blog/static/tests/tours/blog_manager_tour.js
@@ -40,6 +40,9 @@ registerWebsitePreviewTour(
             run: "editor Manager Blog Test Title",
         },
         {
+            trigger: ".o-snippets-menu:has(.o_customize_tab)",
+        },
+        {
             content: "Edit blog content.",
             trigger: ":iframe #o_wblog_post_content p",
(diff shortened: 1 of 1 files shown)


addons/website/static/src/js/tours/tour_utils.js
```diff
@@ -277,7 +277,7 @@ export function clickOnEditAndWaitEditMode(position = "bottom") {
         },
         {
             content: "Check that we are in edit mode",
-            trigger: ".o_builder_sidebar_open",
+            trigger: `body:not(:has([aria-label="Loading"])):has(.o-snippets-menu) :iframe body[is-ready='true']`,
```
FrancoisGe: I'm not a big fan of this change. It seems wrong. Before, we waited for the sidebar to open, but now we wait for the loading to finish and for is-ready to appear. is-ready means that the website page is ready (Interaction enabled, etc.), but that doesn't mean we're in edit mode.

pipu-odoo: Indeed, can be removed. Especially since that's not what fix the tour

PR #246695, opened 2026-02-02
[IMP] hr_expense: Remove cron used to send submitted email & improve mail template

seifgneedy:
* Improve the submitted email template to include a summary of expenses pending approval
* Remove the cron used to send the submitted email, as the cron is costly.
* Send an email when an expense is submitted, as long as we didn’t already send one recently.

task-5880374

---


addons/hr_expense/data/hr_expense_cron.xml
```diff
@@ -1,13 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<odoo>
-    <data>
-        <record id="ir_cron_send_submitted_expenses_mail" model="ir.cron">
```
oco-odoo, line 4: The removal of this cron should be in the migration script ; shouldn't it ?

addons/hr_expense/data/mail_templates.xml
```diff
@@ -36,26 +36,45 @@
             <div style="background:#F0F0F0;color:#515166;padding:10px 0px;font-family:Arial,Helvetica,sans-serif;font-size:14px;">
                 <table style="width:600px;margin:5px auto;background:white;border:1px solid #e1e1e1;">
                     <tbody><tr>
-                        <td style="padding:15px 20px 10px 20px;">
-                            <p style="font-size:20px;font-weight:bold;color:#515166;">Expenses approval</p>
+                        <td style="padding: 15px 20px 10px 20px;">
```
oco-odoo, line 39: Not sure there is a way, but this template is noupdate=1 (because users could modify it by hand), so even for people using the standard one, it will not be updated to the new version .... What do we do for those cases ? 🤔  Maybe we could check the write_uid at migration to decide whether or not we force-update it ? => to discuss PO, then with Upgrade if needed (check the helpers first)

addons/hr_expense/models/hr_expense.py
```diff
@@ -142,6 +143,7 @@ def _default_employee_id(self):
     approval_date = fields.Datetime(string="Approval Date", readonly=True)
     duplicate_expense_ids = fields.Many2many(comodel_name='hr.expense', compute='_compute_duplicate_expense_ids')  # Used to trigger warnings
     same_receipt_expense_ids = fields.Many2many(comodel_name='hr.expense', compute='_compute_same_receipt_expense_ids')  # Used to trigger warnings
+    submission_date = fields.Datetime(string="Submission Date", readonly=True, copy=False)
```
oco-odoo: The name is confusing IMO. "Submission" already refers to a step in the expense flow, with a matching "submitted" state. Here, it's the date where it got sent by mail to the manager. Something like last_notification_date ?

addons/hr_expense/views/hr_expense_views.xml
```diff
@@ -537,6 +537,7 @@
             <field name="res_model">hr.expense</field>
             <field name="view_mode">list,kanban,form,graph,pivot,activity</field>
             <field name="search_view_id" ref="hr_expense_view_search_with_panel"/>
+            <field name="domain">[('manager_id', '=?', context.get('active_id'))]</field>
```
oco-odoo, line 540: Looks sus 🧐 This action is called from a menuitem, so no context key is ever set then, right ? 🤔

addons/hr_expense/models/hr_expense.py
```diff
@@ -1055,17 +1073,35 @@ def _send_submitted_expenses_mail(self):
             for manager, expenses_submitted in expenses_submitted_per_company.grouped('manager_id').items():
                 if not manager:
                     continue
+                last_submission_date = other_expenses_map.get((company.id, manager.id), {}).get('max_submission_date')
+                if last_submission_date and last_submission_date > (fields.Datetime.now() - timedelta(hours=12)):
```
oco-odoo: fields.Datetime.now() - timedelta(hours=12) 
could be put in a variable before booth loops

addons/hr_expense/models/hr_expense.py
```diff
@@ -1055,17 +1073,35 @@ def _send_submitted_expenses_mail(self):
...
+                    continue
+
+                all_submitted_expenses = other_expenses_map.get((company.id, manager.id), {}).get('expenses', self.env['hr.expense'])
```
oco-odoo: You could keep  other_expenses_map.get((company.id, manager.id), {}) in a variable

addons/hr_expense/models/hr_expense.py
```diff
@@ -1149,6 +1185,7 @@ def action_submit(self):
         (self - expenses_autovalidated).approval_state = 'submitted'
         if expenses_autovalidated:  # Note, this will and should bypass the duplicate check. May be changed later
             expenses_autovalidated._do_approve(check=False)
+        self.sudo().submission_date = fields.Datetime().now()
```
oco-odoo: Shouldn't it rather be done (without the sudo()) at the end of _send_submitted_expenses_mail ?

oco-odoo (changes requested):

addons/hr_expense/views/hr_expense_views.xml
seifgneedy: The action is called from two flows:
1. menuitem: this doesn't have a context key(active_id) set, that's why I'm using `=?`, domain will be `[]` in that case.
2. Through the URL in the email notification we're sending, here we set the active_id to the manager of the expenses listed in the email, s.t when`View expenses` is clicked, only the expenses of the manager receiving the email are viewed.
```
'url': f'/odoo/{manager.id}/expenses-to-process',
```

addons/hr_expense/data/mail_templates.xml
seifgneedy: With the way mentioned in the upgrade comment: https://github.com/odoo/upgrade/pull/9395#discussion_r2769496719 . I think it will be ok only to update the unchanged mail template.

oco-odoo (approved):
@robodoo r+

PR #246696, opened 2026-02-02
[IMP] im_livechat, *: harmonize Live Chat copywriting

Atovange:
* = spreadsheet_dashboard_im_livechat, website, website_livechat

This commit harmonizes the copywriting of the Live Chat app name.

part of task-4461200
part of task-5867464


seb-odoo (approved):
@robodoo r+ delegate+

I've not checked if you didn't miss any as it's quite hard to grep with all the technical stings, but ok for those in the diff.

PR #246697, opened 2026-02-02
[FIX] l10n_ro_efactura: Allow reset to draft for e-bills

nape-odoo:
Adjusting the visibility check for "Reset to draft" button to allow Vendor Bills received from ANAF to be reset even when they have a EDI state.
Will require to be shifted to `l10n_ro_edi` in 18.0+ as the efactura module is merged into it.
task-5892651

---


jorenvo (commented):
Hey @nape-odoo! No expert in Romania, but I think we only handle `out_refund` and `out_invoice` in this module, so perhaps we should check for that instead?

jorenvo (approved):
Thanks @nape-odoo!

@robodoo r+

PR #246705, opened 2026-02-02
[IMP] website: add arrow styling option for dynamic snippet

alap-odoo:
This PR adds new option for arrow styling on the dynamic snippet carousel, including "**default**", "**boxed**", "**rounded**", and "**hidden**" states. Additionally, it adds a new option to **invert** the arrow colors.

task-5179779

### DIFF
--- a/addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
+++ b/addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
@@ -12,6 +12,20 @@
                 <BuilderSelectItem classAction="'o_arrows_bottom o_arrows_bottom_right'">Bottom Right</BuilderSelectItem>
             </BuilderSelect>
         </BuilderRow>
+        <BuilderRow label.translate="Style" level="1">
+            <BuilderSelect>
+                <BuilderSelectItem classAction="'o_default_arrows'">Default</BuilderSelectItem>
+                <BuilderSelectItem classAction="'o_boxed_arrows'">Boxed</BuilderSelectItem>
+                <BuilderSelectItem classAction="''">Rounded</BuilderSelectItem>
+                <BuilderSelectItem classAction="'o_hidden_arrows'">Hidden</BuilderSelectItem>
+            </BuilderSelect>
+        </BuilderRow>
+        <BuilderRow label.translate="Color" level="1">
+            <BuilderSelect>
+                <BuilderSelectItem classAction="'o_light_arrows'">Regular</BuilderSelectItem>
+                <BuilderSelectItem classAction="''">Inverted colors</BuilderSelectItem>
+            </BuilderSelect>
+        </BuilderRow>
 	    <BuilderRow label.translate="Slider Speed" t-if="numberOfRecords > 1">
 	        <BuilderNumberInput action="'setCarouselSliderSpeed'" default="1" unit="'s'" saveUnit="''" step="0.1" min="0" preview="false" id="'speed_opt'"/>
 	    </BuilderRow>
--- a/addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
+++ b/addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
@@ -24,6 +24,37 @@
         width: auto;
     }
 
+    // Rounded arrows (default selected)
+    .o_carousel_arrow {
+        border-radius: $border-radius-pill;
+        background-color: $carousel-dark-indicator-active-bg;
+        color: $carousel-control-color;
+    }
+
+    &.o_light_arrows, &.o_default_arrows {
+        .o_carousel_arrow {
+            filter: $carousel-dark-control-icon-filter;
+        }
+    }
+
+    &.o_default_arrows {
+        .o_carousel_arrow {
+            background: none;
+        }
+
+        &.o_light_arrows .o_carousel_arrow {
+            filter: none;
+        }
+    }
+
+    &.o_boxed_arrows .o_carousel_arrow {
+        border-radius: $border-radius;
+    }
+
+    &.o_hidden_arrows .o_carousel_arrow {
+        display: none;
+    }
+
     &:where(:not(.o_carousel_multi_items)) > :where(:not(.container-fluid)) {
         .s_dynamic_snippet_title_aside + .s_dynamic_snippet_content {
             .s_dynamic_snippet_arrows {
@@ -37,13 +68,21 @@
     }
 
     @include media-breakpoint-up(md) {
-        &:where(:not([class*="o_arrows_"])){
+        --o-DynamicCarousel__arrowPrev-offset: -50% /* rtl: 50% */;
+        --o-DynamicCarousel__arrowNext-offset: 50% /* rtl: -50% */;
+
+        &.o_default_arrows {
+            --o-DynamicCarousel__arrowPrev-offset: -100% /* rtl: 100% */;
+            --o-DynamicCarousel__arrowNext-offset: 100% /* rtl: -100% */;
+        }
+
+        &:where(:not([class*="o_arrows_"])) {
             .carousel-control-prev {
-                transform: translateX(-50%);
+                transform: translateX(var(--o-DynamicCarousel__arrowPrev-offset));
             }
 
             .carousel-control-next {
-                transform: translateX(50%);
+                transform: translateX(var(--o-DynamicCarousel__arrowNext-offset));
             }
         }
 
--- a/addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
+++ b/addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
@@ -3,13 +3,13 @@
     <t t-name="website.s_dynamic_snippet.carousel.arrows">
         <div class="o_carousel_arrow_wrap align-items-center pe-none">
             <a t-attf-href="##{unique_id}" class="carousel-control-prev position-relative pe-auto" data-bs-slide="prev" role="button" aria-label="Previous" title="Previous">
-                <span class="oi oi-chevron-left rounded-pill p-3 bg-700"/>
+                <span class="o_carousel_arrow oi oi-chevron-left p-3"/>
                 <span class="visually-hidden">Previous</span>
             </a>
         </div>
         <div class="o_carousel_arrow_wrap align-items-center justify-content-end pe-none">
             <a t-attf-href="##{unique_id}" class="carousel-control-next position-relative pe-auto" data-bs-slide="next" role="button" aria-label="Next" title="Next">
-                <span class="oi oi-chevron-right rounded-pill p-3 bg-700"/>
+                <span class="o_carousel_arrow oi oi-chevron-right p-3"/>
                 <span class="visually-hidden">Next</span>
             </a>
         </div>


addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
```diff
@@ -11,6 +11,14 @@
                 <BuilderSelectItem classAction="'o_arrows_bottom o_arrows_bottom_center'">Bottom Center</BuilderSelectItem>
                 <BuilderSelectItem classAction="'o_arrows_bottom o_arrows_bottom_right'">Bottom Right</BuilderSelectItem>
             </BuilderSelect>
+        </BuilderRow>
+        <BuilderRow label.translate="Style" level="1">
+            <BuilderSelect>
+                <BuilderSelectItem classAction="'arrows_default'">Default</BuilderSelectItem>
+                <BuilderSelectItem classAction="'arrows_boxed'">Boxed</BuilderSelectItem>
+                <BuilderSelectItem classAction="''">Rounded</BuilderSelectItem>
```
shsa-odoo, line 19: NIT: Let's follow the same convention as carousels give it 'arrow_rounded' class, instead of leaving it empty :smiling_face_with_tear:  and adapt the scss and xml accordingly as well. Let's keep the default option applied as well. like other (carousel) has default option is selected when snippet is dropped.

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -84,3 +84,36 @@
...
 }
+
+// Rounded arrows (default selected)
+.s_dynamic_snippet_arrows {
+    .carousel-control-prev,
+    .carousel-control-next {
+        width: $-o-carousel-controllers-size;
+        height: $-o-carousel-controllers-size;
+        background: $carousel-dark-indicator-active-bg;
+        border-radius: $border-radius-pill
```
shsa-odoo: nit : semicolon missing


Along with that When `arrows_default` is applied, it only sets `background: none` and `filter: invert(1)`, but the parent `.s_dynamic_snippet_arrows` selector still applies the rounded border, which has not UI impact.

shsa-odoo (commented):
@alap-odoo Few comments : )

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
alap-odoo: Discussed in IRL

As discussed with PO, by default `rounded` option should be selected. So, I found this approach more feasible.

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -84,3 +84,37 @@
...
+        border-radius: $border-radius;
+    }
+}
+
+.arrows_hidden {
+    .carousel-control-prev,
+    .carousel-control-next {
+        display: none;
+    }
+}
```
shsa-odoo: This should have been added [here](https://github.com/odoo/odoo/pull/246705/files#diff-1e9ce3b0f822b0986c402b461eea517085eb542d689b25aaf28706d63e5dcedfL23)

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
```diff
@@ -11,6 +11,14 @@
                 <BuilderSelectItem classAction="'o_arrows_bottom o_arrows_bottom_center'">Bottom Center</BuilderSelectItem>
```
sobo-odoo, line 11: Behavior:
- The snippet preview in the snippets dialog is weird: the arrows are too high and they have a weird coloring 
  image

  Note that in fact before your PR, the arrows were not visible in the preview

- You should also bring the "Invert Color" options, because then the arrow style doesn't look good depending on the background color, the products images, the cards template... we should be able to decide if they are dark or light :slightly_smiling_face:  
  | Dark | Light |
  | --- | --- |
  | image | image |

- For some reason, the arrows take the color of primary :open_mouth:

  image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -20,8 +20,33 @@
...
 
+    // Rounded arrows (default selected)
     .carousel-control-prev,	.carousel-control-next {
-        width: auto;
+        width: $-o-carousel-controllers-size;
+        height: $-o-carousel-controllers-size;
+        background: $carousel-dark-indicator-active-bg;
+        border-radius: $border-radius-pill;
+    }
+    &.arrows_default {
```
sobo-odoo: Add an empty line before that rule, to stay consistent :+1:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -20,8 +20,33 @@
...
+        .carousel-control-prev,
+        .carousel-control-next {
```
sobo-odoo: Put the selectors on one line, it takes space for nothing :slightly_smiling_face: (same for the rules below)

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -20,8 +20,33 @@
...
+            background: none;
+            filter: invert(1);
```
sobo-odoo: Shouldn't you use the `$carousel-dark-control-icon-filter` variable directly here ?

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -20,8 +20,33 @@
         display: contents;
     }
 
+    // Rounded arrows (default selected)
     .carousel-control-prev,	.carousel-control-next {
-        width: auto;
+        width: $-o-carousel-controllers-size;
+        height: $-o-carousel-controllers-size;
+        background: $carousel-dark-indicator-active-bg;
```
sobo-odoo: - -> background-color :+1: 
- Make sure to use important somewhere, or else older snippets that still have the `rounded-pill p-3 bg-700` classes will look like this :arrow_down: :grimacing: (+ also do something about the padding)
  
image
image image

Instead of

image 
image
image

sobo-odoo: You removed the `bg-700` class from the icon, but didn't set back a background-color on it, you did it on the `a` above it => that's why the color is primary I think
image


Make sure to distinguish well what should be applied on the icon, and what should be on the button, like in the carousel rules (`carousel-control-prev-icon` vs `carousel-control-prev`) :slightly_smiling_face: 

This comment applies to the whole file :+1:

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
sobo-odoo, line 11: Commit message:
- > It is acheived by adding "arrows_*" class using classAction and defining the related CSS rules.

  You don't need to describe the code, a commit message should describe the "why", not the "what" :wink: So you can remove that sentence :+1:

sobo-odoo (changes requested):

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
alap-odoo: I thought it would be helpful to have a quick overview of how the spec is achieved. However, I will remove it.

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: I thought the same as well, but it’s due to the formatter. Moreover, there are several places in the codebase where the same selectors are written this way, so I followed that approach. However, I will keep these selectors on a single line going forward.

image
image
image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: > You removed the bg-700 class from the icon, but didn't set back a background-color on it, you did it on the a above it

Yeah that's why the icons are taking the **link** color of the selected preset.

What if we apply the text-white class to the icons, as shown below?

```javascript
<span class="oi oi-chevron-right text-white"/>
``` 

Previously, the `bg-700` class was used, which indirectly applied the `color: #fff;` CSS property. 
image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: We can use it. The reason I didn’t use it earlier is that this variable applies the grayscale(100) property, which converts elements to black and white. Since our icons are already black and white, the grayscale filter is not necessary. However, it will not have a significant impact, so I will go ahead and use that variable for uniformity.

```scss
$carousel-dark-control-icon-filter:  invert(1) grayscale(100) !default;
```

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: I’m a little confused here. The `rounded-pill bg-700` classes were defined in `000.xml` inside the `/static/src` folder. After upgrading to master, wouldn’t these classes be removed automatically since they are rendered on the client side?

I tried to reproduce the behavior shown in the image by following these steps:
1. Checked out the `master` branch.
2. Dropped the snippet.
3. Checked out my branch again.
4. Ran it using the same database.
5. Changed the arrow styling.

However, it worked correctly for me. Am I missing something?

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
alap-odoo: > The snippet preview in the snippets dialog is weird: the arrows are too high and they have a weird coloring. Note that in fact before your PR, the arrows were not visible in the preview

I have implement the same behavior as master in this PR. Moreover, if we want to display the arrows in preview, then I can handle it in separate PR after this task get merged.

> You should also bring the "Invert Color" options

Yeah, that's true. I will add that option.

> For some reason, the arrows take the color of primary 

Discussed in below comment [here](https://github.com/odoo/odoo/pull/246705#discussion_r2813176317)

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
sobo-odoo: My bad, I forgot that the dynamic snippet is re-rendered by its interaction every time, so it will always be up to date with the template :+1: 
I tested by adding the classes manually, thinking it was really saved like that but it was wrong 

So you can ignore this comment :+1:

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
sobo-odoo: Just FYI: https://www.odoo.com/documentation/master/contributing/development/git_guidelines.html#commit-message-full-description :wink: 

=> We don't need to explain the what because we can directly read the code for that :wink: You should describe why the commit is needed and if you have to describe the code, it is to explain some design choices, issues encountered leading to that code, and so on :slightly_smiling_face:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
sobo-odoo: `text-white` seems to be a good solution :+1:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -24,6 +24,42 @@
         width: auto;
     }
 
+    // Rounded arrows (default selected)
+    .oi-chevron-left, .oi-chevron-right {
```
sobo-odoo: Hmm, I don't find these classes specific enough to just use them like that in the CSS => like this, any `oi-chevron-*` in the snippet will have the style, but we only want it on the carousel controls :thinking: 

Maybe you can add  the `carousel-previous|next-icon` classes in the xml and use them in the SCSS ? :slightly_smiling_face: 

Of course, we don't really have the case, but even without that, I find it easier to understand the rules if we have the classes, and we will have more the parallel with the carousels :slightly_smiling_face: 

However, from what I saw the `carousel-previous|next-icon` classes provide styles so I suggest the following:
- either add these classes in the template and adapt your rules to them,
- or create your own classes, add them in the template and only change the selector of your rules here :slightly_smiling_face: 

I prefer the first option so it will make all carousels consistent, but it depends on if it will be easy to adapt or not (since the default arrows are not the same) :slightly_smiling_face:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -24,6 +24,42 @@
...
+        background: $carousel-dark-indicator-active-bg;
```
sobo-odoo: `background` => `background-color` to be more specific since that variable only contains a color :+1:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -24,6 +24,42 @@
...
+        border-radius: $border-radius-pill;
+    }
+
+    &.arrows_light, &.arrows_default {
+        .oi-chevron-left, .oi-chevron-right {
+            filter: $carousel-dark-control-icon-filter;
+        }
+    }
+
+    &.arrows_default {
```
sobo-odoo: For the default arrows, maybe it could be nice to shift them a bit so they don't overlap with both the background and the images, or else it will complicated to always have them visible :slightly_smiling_face: (they should be over the background only)
| | | |
|--- | --- | --- |
| image | image | image |

The other arrows look nice because they have a background, but given that the default arrows have no effects, they will be hard to use :slightly_smiling_face:

sobo-odoo (changes requested):
Some other comments :slightly_smiling_face:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: I was thinking, instead of adding an additional class, would it be better to increase the CSS specificity? :thinking: 
```scss
.s_dynamic_snippet_arrows{
  .oi-chevron-left, .oi-chevron-right {
    background: $carousel-dark-indicator-active-bg;
    border-radius: $border-radius-pill;
  }
}
``` 
This would allow us to scope the styling more precisely without introducing an extra class. What's your views on it ?

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
alap-odoo: I think, what if we increase the value of `transform: translateX(XX%);` property to something like shown :thinking: 

**Before:**
```scss
&:where(:not([class*="o_arrows_"])){
  .carousel-control-prev {
    transform: translateX(-50%);
  }

  .carousel-control-next {
    transform: translateX(50%);
  }
}
``` 

**After:**
```scss
--arrow-offset: 50%;
&.arrows_default {
    --arrow-offset: 65%;
}
&:where(:not([class*="o_arrows_"])){
  .carousel-control-prev {
    transform: translateX(calc(-1 * var(--arrow-offset)));
  }

  .carousel-control-next {
    transform: translateX(var(--arrow-offset));
  }
}
``` 
As per this implementation, UI looks something like this:
image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
sobo-odoo: It can be a solution too indeed, I'm okay with it :+1:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -1,3 +1,4 @@
+$arrow-icons-selector: ".s_dynamic_snippet_arrows .oi-chevron-left, .s_dynamic_snippet_arrows .oi-chevron-right";
```
sobo-odoo: Move this right after `.s_dynamic` so that "variable" is scoped to that snippet only :slightly_smiling_face:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -37,13 +74,17 @@
     }
 
     @include media-breakpoint-up(md) {
+        --arrow-offset: 50%;
+        &.arrows_default {
+            --arrow-offset: 65%;
```
sobo-odoo: Maybe use a higher value, I find them too stuck to the images, some space around would be better => I tested with 100% and it looks pretty good  :arrow_down::slightly_smiling_face: 

image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -37,13 +74,17 @@
...
+        }
         &:where(:not([class*="o_arrows_"])){
             .carousel-control-prev {
-                transform: translateX(-50%);
+                transform: translateX(calc(-1 * var(--arrow-offset)));
             }
 
             .carousel-control-next {
-                transform: translateX(50%);
+                transform: translateX(var(--arrow-offset));
```
sobo-odoo: For some reason, the left arrow is shifted in RTL languages, it was not the case before your PR :open_mouth:

image

(I am adding the comment on these lines because there may be something more to do with the transform, but I am not sure if this is the issue)

sobo-odoo (changes requested):
Some last things to change :slightly_smiling_face:

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
```diff
@@ -11,6 +11,20 @@
...
+                <BuilderSelectItem classAction="'arrows_default'">Default</BuilderSelectItem>
+                <BuilderSelectItem classAction="'arrows_boxed'">Boxed</BuilderSelectItem>
+                <BuilderSelectItem classAction="''">Rounded</BuilderSelectItem>
+                <BuilderSelectItem classAction="'arrows_hidden'">Hidden</BuilderSelectItem>
+            </BuilderSelect>
+        </BuilderRow>
+        <BuilderRow label.translate="Color" level="1">
+            <BuilderSelect>
+                <BuilderSelectItem classAction="'arrows_light'">Regular</BuilderSelectItem>
+                <BuilderSelectItem classAction="''">Inverted colors</BuilderSelectItem>
```
Brieuc-brd:
```suggestion
                <BuilderSelectItem classAction="'o_arrows_boxed'">Boxed</BuilderSelectItem>
                <BuilderSelectItem classAction="''">Rounded</BuilderSelectItem>
                <BuilderSelectItem classAction="'o_arrows_hidden'">Hidden</BuilderSelectItem>
            </BuilderSelect>
        </BuilderRow>
        <BuilderRow label.translate="Color" level="1">
            <BuilderSelect>
                <BuilderSelectItem classAction="'o_arrows_light'">Regular</BuilderSelectItem>
                <BuilderSelectItem classAction="''">Inverted colors</BuilderSelectItem>
```

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
```diff
@@ -3,13 +3,13 @@
     <t t-name="website.s_dynamic_snippet.carousel.arrows">
         <div class="o_carousel_arrow_wrap align-items-center pe-none">
             <a t-attf-href="##{unique_id}" class="carousel-control-prev position-relative pe-auto" data-bs-slide="prev" role="button" aria-label="Previous" title="Previous">
-                <span class="oi oi-chevron-left rounded-pill p-3 bg-700"/>
+                <span class="oi oi-chevron-left p-3 text-white"/>
```
Brieuc-brd: Is `text-white` really necessary ? 🤔 
Same for the right arrow :)

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -1,4 +1,5 @@
 .s_dynamic {
+    $arrow-icons-selector: ".s_dynamic_snippet_arrows .oi-chevron-left, .s_dynamic_snippet_arrows .oi-chevron-right";
```
Brieuc-brd: I wouldn’t use icon classes as selectors, since this could break if we ever change the icon in the DOM.
Let’s simplify things by adding a custom class to both arrows, such as `o_carousel_arrow`.
That way, we won’t need to store the selector in a variable 👍 .

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -24,6 +25,42 @@
         width: auto;
     }
 
+    // Rounded arrows (default selected)
+    #{$arrow-icons-selector} {
```
Brieuc-brd:
```suggestion
    .o_carousel_arrow {
```
Same for the other occurrences below.

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -37,13 +74,18 @@
     }
 
     @include media-breakpoint-up(md) {
+        --arrow-offset: 50%;
```
Brieuc-brd: Let’s follow your [coding guidelines](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html?highlight=guidelines#css-variables) to rename this custom CSS variable.

Something like `--o-DynamicCarousel__arrow-offset` ? 🙂 

Also, add an empty space after the variable declaration 👍 ([example](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html?highlight=guidelines#css-variables:~:text=.o_element%20%7B%0A%20%20%20%24%2Dinner%2Dgap,filter%3A%20blur(2px)%3B%0A%7D))

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -37,13 +74,18 @@
...
+        &.arrows_default {
+            --arrow-offset: 100%;
+        }
```
Brieuc-brd, line 77:
```suggestion
        }
        
```

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -37,13 +74,18 @@
...
         &:where(:not([class*="o_arrows_"])){
             .carousel-control-prev {
-                transform: translateX(-50%);
+                transform: translateX(calc(-1 * var(--arrow-offset)));
             }
 
             .carousel-control-next {
-                transform: translateX(50%);
+                // "calc(1 * ...)" ensures value is correctly inverted in RTL.
```
Brieuc-brd: Are you sure about that?
I did a quick test and removed the `calc(1 * ...)`, and it seems to work.
Did I miss something? 🤔

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
```diff
@@ -3,13 +3,13 @@
...
                 <span class="visually-hidden">Previous</span>
             </a>
         </div>
         <div class="o_carousel_arrow_wrap align-items-center justify-content-end pe-none">
             <a t-attf-href="##{unique_id}" class="carousel-control-next position-relative pe-auto" data-bs-slide="next" role="button" aria-label="Next" title="Next">
-                <span class="oi oi-chevron-right rounded-pill p-3 bg-700"/>
+                <span class="oi oi-chevron-right p-3 text-white"/>
```
Brieuc-brd:
```suggestion
                <span class="o_carousel_arrow oi oi-chevron-right p-3 text-white"/>
```

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
Brieuc-brd:
```suggestion
                <span class="o_carousel_arrow oi oi-chevron-left p-3 text-white"/>
```

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
```diff
@@ -24,6 +25,42 @@
         width: auto;
     }
 
+    // Rounded arrows (default selected)
+    #{$arrow-icons-selector} {
+        background-color: $carousel-dark-indicator-active-bg;
+        border-radius: $border-radius-pill;
```
Brieuc-brd: Nitpicking 😅 
Following our [coding guidelines](https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html?highlight=guidelines#properties-order), `border-radius` should be defined first.

```suggestion
        border-radius: $border-radius-pill;
        background-color: $carousel-dark-indicator-active-bg;
```

Brieuc-brd (changes requested):
Thanks @alap-odoo, it looks nice 👍 
I've left a few comments below 🙂

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
alap-odoo: Yes, It prevents the arrows from inheriting the **link** color of the selected color preset.

Without `text-white`, arrows will inherit the link color of preset like shown below:
image

Brieuc-brd: Okay, I see 👍  the color preset overrides all link colors, including the arrows:
https://github.com/odoo/odoo/blob/master/addons/html_editor/static/src/scss/html_editor.common.scss#L653

I’m not sure that’s what we want. We should probably exclude these arrows from the preset color definition. What do you think? @sobo-odoo @stefanorigano 👋 

This is probably something that should be handled and tested in a separate PR 🙂
For now, I guess we can keep the class you added.

One small thing: could you define it on the `<a/>` element rather than on the icon? Thanks!

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
alap-odoo: I think suggested change was not implemented intentionally to prevent interferance with the CSS condition shown below.
image
Specifically, the condition ensures that the `transform: translateX(xx%)` property is applied only when the arrow position is set to **default**.

I think we can slightly tweak the name to something like `o_default_arrows`, `o_boxed_arrows` and `o_hidden_arrows`.

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
sobo-odoo: @Brieuc-brd I don't think we should go that far, I took a look at the normal carousels and we also have a rule forcing the color of the control to white (thanks to the `$carousel-control-color` variable), so doing the same and forcing it to white here doesn't seem to be an issue :slightly_smiling_face: 

Ideally, we should make the dynamic carousel work like a standard carousel so we can use the already defined carousel rules and extend the `%s_carousel_variants` :thinking: But given that all the carousel options are not in the dynamic one, I guess custom rules are fine for now :slightly_smiling_face: 

That said, maybe we should use `color: $carousel-control-color; ` in the SCSS file instead of the `text-white` class then :+1:

addons/website/static/src/builder/plugins/options/dynamic_snippet_carousel_option.xml
Brieuc-brd: > I think we can slightly tweak the name to something like o_default_arrows, o_boxed_arrows and o_hidden_arrows.

I'm ok with that 👍

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
sobo-odoo: We had [this issue](https://github.com/odoo/odoo/pull/246705#discussion_r2846582957) in RTL languages before that change -> one of the arrows was shifted :slightly_smiling_face: 
Don't you have it ? :open_mouth:

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.xml
alap-odoo: I think `color: $carousel-control-color;` property should be added on the icon instead of `<a/>` because the color preset's link color override the `color: $carousel-control-color;` property, when written in `<a/>`
image

addons/website/static/src/snippets/s_dynamic_snippet_carousel/000.scss
Brieuc-brd: I don't  have any position issue on RTL without `calc(1 * ...)`, am I missing something ? 🤔 

Image

sobo-odoo: If I do this in the SCSS file
image

I have well the issue 
image

Try do to do it locally and in the SCSS file, not manually with the inspector, because I think the RTL already has an effect on what you change :thinking: 
Or maybe save it like that and go back in LTR, you will surely have the opposite issue ?

Brieuc-brd: Ok now I see 👍 
We could simplify the implementation by defining clear values for both modes and avoiding unnecessary calculations 🙂
Could you test this?

```scss
@include media-breakpoint-up(md) {
        --o-DynamicCarousel__arrowPrev-offset: -50% /* rtl: 50% */;
        --o-DynamicCarousel__arrowNext-offset: 50% /* rtl: -50% */;
        
        &.o_default_arrows {
            --o-DynamicCarousel__arrowPrev-offset: -100% /* rtl: 100% */;
            --o-DynamicCarousel__arrowNext-offset: 100% /* rtl: -100% */;
        }

        &:where(:not([class*="o_arrows_"])) {
            .carousel-control-prev {
                transform: translateX(var(--o-DynamicCarousel__arrowPrev-offset));
            }

            .carousel-control-next {
                transform: translateX(var(--o-DynamicCarousel__arrowNext-offset));
            }
        }
 ```

alap-odoo: I think it does not solve the issue.
image

Am I doing something wrong? :thinking: 
image

sobo-odoo: @alap-odoo you also have to add the `/*rtl ... */` in the rules, maybe you don't know but they are commands for the CSS behavior in RTL, not just simple comments :wink: 

And so it works perfectly @Brieuc-brd :+1:

alap-odoo: Yeah, I thing it was missed by me. However, I have adapted the change.
Thanks @sobo-odoo @Brieuc-brd

sobo-odoo (approved):
It's okay with me, thanks for the work :+1: 
And thanks @Brieuc-brd for the review :slightly_smiling_face: 

@robodoo r+

PR #230757, opened 2025-10-09
[ADD] l10n_hr_edi: E-invoicing for Croatia with MojEracun

nape-odoo:
Adding e-invoice generation base functionality for Croatia. Implemented as an extension of account_edi_ubl_cii, similar to other localisations using such documents.
Adding proxy client for MojEracun sevice provider to send invoiced on the eRacun network. Proxy user is entirely separate from the standard Odoo IAP EDI proxy implementation.
Adding UBL 2.1 compliant node `cac:SellerContact` to existing node
`cac:AccountingSupplierParty` to be used in Croatian e-invoice XLM generation
with the new method chain.
XSD validation included in the tests on the client side as there is no corresponding Odoo IAP proxy.

task-4925745

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246722, opened 2026-02-02
[REM] l10n_hr_edi: duplicate method

xmo-odoo:
The duplicate method was introduced in #230757, it apparently was removed while forward porting as it is missing from #241691 but nobody went back to remove it from the original, and after updating pylint on the nightly distro builds they now complain.

https://runbot.odoo.com/odoo/runbot.build.error/237791


xmo-odoo:
@robodoo r+

PR #246723, opened 2026-02-02
[FIX] stock_account: add test to ensure avco valuation considers correct companies

nsirjacobs:
Stock value computation did not take companies into account correctly, this issue was fixed with these commits:
https://github.com/odoo/odoo/commit/701729df41a49843e5717675267a26c490050657
https://github.com/odoo/odoo/commit/87385888f4980b6ee827257238d5a6d1e0ce6a04

This commit adds a test to ensure correct behavior
    
 opw-5871009

### DIFF
--- a/addons/stock_account/tests/test_stockvaluation.py
+++ b/addons/stock_account/tests/test_stockvaluation.py
@@ -3521,3 +3521,46 @@ class TestStockValuation(TestStockValuationCommon):
         closing = self.branch.with_context(allowed_company_ids=[self.branch.id, self.company.id]).action_close_stock_valuation()
         closing_lines = self.env['account.move'].browse(closing['res_id']).line_ids
         self.assertEqual(closing_lines.move_id.company_id.id, self.branch.id)
+
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        product_avco = self.product_avco
+        now = Datetime.now()
+
+        c1 = self.env.company
+        c2 = self.env['res.company'].create({'name': 'Test Company'})
+        c1_stock_loc = self.warehouse.lot_stock_id
+        c2_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c2.id)], limit=1).lot_stock_id
+        product_company1 = product_avco.with_context(allowed_company_ids=[c1.id])
+        product_company2 = product_avco.with_context(allowed_company_ids=[c2.id])
+
+        # Change cost manually and create stock move for each company
+        with freeze_time(now + timedelta(minutes=1)):
+            product_company1.standard_price = 10
+        with freeze_time(now + timedelta(minutes=2)):
+            self._make_in_move(product_company1, 2.0, company=c1, location_id=self.supplier_location.id, location_dest_id=c1_stock_loc.id)
+        with freeze_time(now + timedelta(minutes=3)):
+            product_company2.standard_price = 50
+        with freeze_time(now + timedelta(minutes=4)):
+            self._make_in_move(product_company2, 1.0, company=c2, location_id=self.supplier_location.id, location_dest_id=c2_stock_loc.id)
+
+        self.assertRecordValues(product_company1, [{
+            'total_value': 20,
+            'qty_available': 2,
+            'avg_cost': 10,
+        }])
+        self.assertRecordValues(product_company2, [{
+            'total_value': 50,
+            'qty_available': 1,
+            'avg_cost': 50,
+        }])
+        product_companies = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        product_companies.invalidate_recordset(fnames=['total_value'])
+        self.assertRecordValues(product_companies, [{
+            'total_value': 70,  # sum of both total_values
+            'qty_available': 3,
+            'avg_cost': 23.33,
+        }])
--- a/addons/stock_account/tests/test_stockvaluation.py
+++ b/addons/stock_account/tests/test_stockvaluation.py
@@ -3122,3 +3122,45 @@ class TestStockValuation(TestStockValuationCommon):
         recs[-2:]._compute_cumulative_fields()
         self.assertEqual(recs[-1].total_quantity, 3)
         self.assertEqual(recs[-1].total_value, 30)
+
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
+        product_avco = self.product_avco
+        time = Datetime.now()
+
+        c1 = self.env.company
+        c2 = self.env['res.company'].create({
+            'name': 'Test Company',
+        })
+        c1_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c1.id)], limit=1).lot_stock_id
+        c2_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c2.id)], limit=1).lot_stock_id
+        c1_product = product_avco.with_context(allowed_company_ids=[c1.id])
+        c2_product = product_avco.with_context(allowed_company_ids=[c2.id])
+
+        # Change cost manually and create stock move for each company
+        # freeze_time is used because avco compares dates of stock moves against dates of new values
+        with freeze_time(time + timedelta(minutes=1)):
+            c1_product.standard_price = 10
+        with freeze_time(time + timedelta(minutes=2)):
+            self._make_in_move(c1_product, 2.0, company=c1, location_id=self.supplier_location.id, location_dest_id=c1_stock_loc.id)
+
+        with freeze_time(time + timedelta(minutes=3)):
+            c2_product.standard_price = 50
+        with freeze_time(time + timedelta(minutes=4)):
+            self._make_in_move(c2_product, 1.0, company=c2, location_id=self.supplier_location.id, location_dest_id=c2_stock_loc.id)
+
+        self.assertEqual(c1_product.total_value, 20)
+        self.assertEqual(c1_product.qty_available, 2)
+        self.assertEqual(c1_product.avg_cost, 10)
+        self.assertEqual(c2_product.total_value, 50)
+        self.assertEqual(c2_product.qty_available, 1)
+        self.assertEqual(c2_product.avg_cost, 50)
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        c3_product.invalidate_recordset(fnames=['total_value'])
+        self.assertEqual(c3_product.total_value, 70)  # sum of both total_values
+        self.assertEqual(c3_product.qty_available, 3)
+        self.assertEqual(c3_product.avg_cost, 23.33)


addons/stock_account/models/product.py
```diff
@@ -289,7 +290,7 @@ def _get_last_in(self, date=None):
         return last_in
 
     def _with_valuation_context(self):
-        valued_locations = self.env['stock.location'].search([('is_valued_internal', '=', True)])
+        valued_locations = self.env['stock.location'].search([('is_valued_internal', '=', True), ('company_id', 'in', self.env.companies.ids)])
```
lase-odoo: I like this change because it makes the total value of the stock based on the `allowed_company_ids` present in the context just as the `qty_available`:
https://github.com/odoo/odoo/blob/242afb9ca3a76e3628260ac81a9f5ddcd5d445dd/addons/stock/models/product.py#L146-L151

On the other hand, I was afraid that we should also consider internally valued locations without companies like maybe the inter-company transit location:
https://github.com/odoo/odoo/blob/52180c9ff3c3a3372a118fb40e133265bbaab22b/addons/stock/data/stock_data.xml#L34-L39
But `is_valued_internal` requires a company to be set on the location !
https://github.com/odoo/odoo/blob/52180c9ff3c3a3372a118fb40e133265bbaab22b/addons/stock_account/models/stock_location.py#L15
https://github.com/odoo/odoo/blob/52180c9ff3c3a3372a118fb40e133265bbaab22b/addons/stock_account/models/stock_location.py#L27-L34
https://github.com/odoo/odoo/blob/52180c9ff3c3a3372a118fb40e133265bbaab22b/addons/stock_account/models/stock_location.py#L36-L41
Hence, the filter looks right to me !

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+        c1_product = product_avco.with_context(allowed_company_ids=[c1.id])
+        c2_product = product_avco.with_context(allowed_company_ids=[c2.id])
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        c1_product.cost_method = 'average'
+        c2_product.cost_method = 'average'
+        c3_product.cost_method = 'average'
+
+        # Change cost manually and create stock move for each company
+        # freeze_time is used because avco compares dates of stock moves against dates of new values
+        with freeze_time(Datetime.now() + timedelta(minutes=1)):
```
lase-odoo: `Datetime.now()` will not be the same form one call to the next, hence for consistencies, it is best to set it once and for all in a variable at the start of the test `now = Datetime.now()` and add the time delta to this `now` each time you need a new timestamp.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+            'name': 'Test Company',
+        })
+        c1_stock_loc = self.env['stock.location'].search([('company_id', '=', c1.id), ('usage', '=', 'internal')], limit=1)
+        c2_stock_loc = self.env['stock.location'].search([('company_id', '=', c2.id), ('usage', '=', 'internal')], limit=1)
+        c1_product = product_avco.with_context(allowed_company_ids=[c1.id])
+        c2_product = product_avco.with_context(allowed_company_ids=[c2.id])
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        c1_product.cost_method = 'average'
+        c2_product.cost_method = 'average'
+        c3_product.cost_method = 'average'
```
lase-odoo: It is weird to set the `cost_method` field of a product directly as it is a  computed non-stored field without set methods. Also, that field is **not** company_dependent. By contrast the `property_cost_method` of the product category is if you really want. bu tin any case it does not and should not depend on the `allowed_company_ids` context.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+                'location_id': self.supplier_location.id,
+                'location_dest_id': c1_stock_loc.id,
+                'product_id': c1_product.id,
+                'product_uom': self.uom.id,
+                'product_uom_qty': 2.0,
+            })
+            move_1._action_confirm()
+            move_1.quantity = 2.0
+            move_1.picked = True
+            move_1._action_done()
```
lase-odoo: Use the `_make_in_move` util directly !

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+    @users('admin')
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
+        product_avco = self.product_avco.with_user(self.env.user)
+
+        #  Setup Companies, Locations, and Products
```
lase-odoo: No need to add such comments, it is quite common operations.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+        c1 = self.env.company
+        c2 = self.env['res.company'].create({
+            'name': 'Test Company',
+        })
+        c1_stock_loc = self.env['stock.location'].search([('company_id', '=', c1.id), ('usage', '=', 'internal')], limit=1)
+        c2_stock_loc = self.env['stock.location'].search([('company_id', '=', c2.id), ('usage', '=', 'internal')], limit=1)
```
lase-odoo: I would rather search on the `stock.warehouse` an use the `warehouse.lot_stock_id` because there are many internal locations that are not the stock location.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+        c1_product = product_avco.with_context(allowed_company_ids=[c1.id])
+        c2_product = product_avco.with_context(allowed_company_ids=[c2.id])
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
```
lase-odoo: I do not think that this assignment is useful by any mean since only the id of the product is used for the creation and move process. I would therefore work with a single product in the entire test and add the `with_context(allowed_company_ids=[c1.id])`,.. at the end of the test for the asserts.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+                'location_id': self.supplier_location.id,
+                'location_dest_id': c2_stock_loc.id,
+                'product_id': c2_product.id,
+                'product_uom': self.uom.id,
+                'product_uom_qty': 1.0,
+            })
+            move_2._action_confirm()
+            move_2.quantity = 1.0
+            move_2.picked = True
+            move_2._action_done()
```
lase-odoo: same as above, `_make_in_move`

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+
+        self.assertEqual(c1_product.total_value, 20)
+        self.env.invalidate_all()                      # ensure quantity is correctly recomputed during valuation
+        self.assertEqual(c2_product.total_value, 50)
+        self.assertEqual(c3_product.total_value, 150)  # last value is 50, and qty in both companies is 3 => 150
```
lase-odoo: You should also assert on `qty_available` (on hand quantity) and on `avg_cost`

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
         recs[-2:]._compute_cumulative_fields()
         self.assertEqual(recs[-1].total_quantity, 3)
         self.assertEqual(recs[-1].total_value, 30)
+
+    @users('admin')
```
lase-odoo: I do not really understand the purpose of this line Imo the the test runs fine with Odoobot and if we really want to run it with minimal access rights we should not use the admin user.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
+        product_avco = self.product_avco.with_user(self.env.user)
```
lase-odoo: `self.anything.with_user(self.env.user)` is equivalent to `self.anything`.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3135,3 +3135,63 @@ def test_avco_report_multiple_page(self):
...
+                'product_uom': self.uom.id,
+                'product_uom_qty': 1.0,
+            })
+            move_2._action_confirm()
+            move_2.quantity = 1.0
+            move_2.picked = True
+            move_2._action_done()
+
+        self.assertEqual(c1_product.total_value, 20)
+        self.env.invalidate_all()                      # ensure quantity is correctly recomputed during valuation
```
lase-odoo: You can just invalidate the recordset rather than the entire cache, but to be honest, the fact that you need to do so probably means that the api_depends_context of the `_compute_value` is not correct. And indeed while the `qty_available` is dependent on the `allowed_company_ids`:
https://github.com/odoo/odoo/blob/61135d75db85bb8b88ddfadb1e76abf5b9dac4cf/addons/stock/models/product.py#L146-L151
The `compute_value` was not prior to your change:
https://github.com/odoo/odoo/blob/61135d75db85bb8b88ddfadb1e76abf5b9dac4cf/addons/stock_account/models/product.py#L139-L141

addons/stock_account/models/product.py
```diff
@@ -265,6 +265,7 @@ def _get_last_product_value(self, date=None, lot=False):
         domain = Domain([
             ('product_id', 'in', self.ids),
             ('move_id', '=', False),
+            ('company_id', 'in', self.env.companies.ids),
```
lase-odoo: I do not really agree with this change, imo the method should return the last value of the current compagny:
```suggestion
            ('company_id', '=', self.env.company),
```
Just as the moves considered in `_run_average_batch`:
https://github.com/odoo/odoo/blob/242afb9ca3a76e3628260ac81a9f5ddcd5d445dd/addons/stock_account/models/product.py#L331-L335
Or in the fifo stack:
https://github.com/odoo/odoo/blob/242afb9ca3a76e3628260ac81a9f5ddcd5d445dd/addons/stock_account/models/product.py#L486-L489

PS: it seems that this part of the change has already been merged:
https://github.com/odoo/odoo/pull/251413

lase-odoo (commented):
Thanks for the proposition ! 🐝

The fix looks almost right to me but is actually already handled by other PR's:
Concerning the last `product.value` considered it was handled in https://github.com/odoo/odoo/pull/251413
And considering the total value contextual dependency on the `allowed_company_ids` it will seemingly be handled in https://github.com/odoo/odoo/pull/249790 see 790fa6858df28ad86de26c117ed208d6dfb8d93b

I think it would be valuable to merge the test thought as none of the other two PR's do test the behavior ! I left a few comments to improve it.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
         consigned_quant = quants.filtered(lambda q: q.location_id == self.stock_location and q.owner_id)
         self.assertEqual(regular_quant.value, 10)
         self.assertEqual(consigned_quant.value, 0)
+
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
```
lase-odoo: I do not understand this comment, is it really relevant to your test ?
```suggestion
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+        product_avco = self.product_avco
+        time = Datetime.now()
+
+        c1 = self.env.company
+        c2 = self.env['res.company'].create({
+            'name': 'Test Company',
+        })
+        c1_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c1.id)], limit=1).lot_stock_id
```
lase-odoo:
```suggestion
        c1_stock_loc = self.warehouse.lot_stock_id
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
         self.assertEqual(consigned_quant.value, 0)
+
+    def test_avco_valuation_multicompany(self):
+        """
+        Test that changing values manually for an AVCO product in different companies results
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
+        product_avco = self.product_avco
+        time = Datetime.now()
```
lase-odoo:
```suggestion
        now = Datetime.now()
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+            c2_product.standard_price = 50
+        with freeze_time(time + timedelta(minutes=4)):
+            self._make_in_move(c2_product, 1.0, company=c2, location_id=self.supplier_location.id, location_dest_id=c2_stock_loc.id)
+
+        self.assertEqual(c1_product.total_value, 20)
+        self.assertEqual(c1_product.qty_available, 2)
+        self.assertEqual(c1_product.avg_cost, 10)
+        self.assertEqual(c2_product.total_value, 50)
+        self.assertEqual(c2_product.qty_available, 1)
+        self.assertEqual(c2_product.avg_cost, 50)
```
lase-odoo: make 2 assertRecordValues one for each product.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        c3_product.invalidate_recordset(fnames=['total_value'])
+        self.assertEqual(c3_product.total_value, 70)  # sum of both total_values
+        self.assertEqual(c3_product.qty_available, 3)
+        self.assertEqual(c3_product.avg_cost, 23.33)
```
lase-odoo: Make a single assertRecordValues
```suggestion
        self.assertRecordValues(c3_product, [{'total_value': 70.0, 'qty_available': 3.0, 'avg_cost': 23.33}])
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+        c2 = self.env['res.company'].create({
+            'name': 'Test Company',
+        })
+        c1_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c1.id)], limit=1).lot_stock_id
+        c2_stock_loc = self.env['stock.warehouse'].search([('company_id', '=', c2.id)], limit=1).lot_stock_id
+        c1_product = product_avco.with_context(allowed_company_ids=[c1.id])
+        c2_product = product_avco.with_context(allowed_company_ids=[c2.id])
+
+        # Change cost manually and create stock move for each company
+        # freeze_time is used because avco compares dates of stock moves against dates of new values
```
lase-odoo:
```suggestion
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+        with freeze_time(time + timedelta(minutes=1)):
+            c1_product.standard_price = 10
+        with freeze_time(time + timedelta(minutes=2)):
+            self._make_in_move(c1_product, 2.0, company=c1, location_id=self.supplier_location.id, location_dest_id=c1_stock_loc.id)
+
```
lase-odoo:
```suggestion
```

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+            self._make_in_move(c2_product, 1.0, company=c2, location_id=self.supplier_location.id, location_dest_id=c2_stock_loc.id)
+
+        self.assertEqual(c1_product.total_value, 20)
+        self.assertEqual(c1_product.qty_available, 2)
+        self.assertEqual(c1_product.avg_cost, 10)
+        self.assertEqual(c2_product.total_value, 50)
+        self.assertEqual(c2_product.qty_available, 1)
+        self.assertEqual(c2_product.avg_cost, 50)
+        c3_product = product_avco.with_context(allowed_company_ids=[c1.id, c2.id])
+        c3_product.invalidate_recordset(fnames=['total_value'])
```
lase-odoo: I would call this product `product_companies`, and the others `product_company1` and `product_company2` starting with the company someone suggests it belongs to that company and c3 is not even a company in your flow.

addons/stock_account/tests/test_stockvaluation.py
```diff
@@ -3293,3 +3293,45 @@ def test_quants_values_partially_consigned(self):
...
+        in the correct total value when the product is viewed in a single or multi-company context.
+        """
+        # _compute_value() uses sudo(False) which doesnt work when the user is the Superuser
+        product_avco = self.product_avco
+        time = Datetime.now()
+
+        c1 = self.env.company
+        c2 = self.env['res.company'].create({
+            'name': 'Test Company',
+        })
```
lase-odoo:
```suggestion
        c2 = self.env['res.company'].create({'name': 'Test Company'})
```

PR #246724, opened 2026-02-02
[FIX] iap: make `account_token` editable

svfu-odoo:
Currently the IAP token (`account_token`) is readonly in the IAP account form view.
It was made readonly in this commit
https://github.com/odoo/odoo/commit/8a96f0fc3f18bd1ccfa96e654ef6f74b21593288#diff-f05503a4bccde75e6ef0c6ab5cd0dbd9fa0e32a30da505bdfcab8c7ed3a4a24cR33

Support needs to edit it when moving a DB from test to production though.

After this commit it is editable (again; like in lower versions).

task-None

### DIFF
--- a/addons/iap/views/iap_views.xml
+++ b/addons/iap/views/iap_views.xml
@@ -30,7 +30,7 @@
                             <field name="service_id" readonly="service_locked" options="{'no_open': True, 'no_create': True}"/>
                             <field name="description"/>
                             <field name="company_ids" widget="many2many_tags" domain="[('id', 'in', allowed_company_ids)]" />
-                            <field name="account_token" groups="base.group_no_one" readonly="1"/>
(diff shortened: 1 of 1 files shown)


clbr-odoo (approved):
Probably worth to check with @RicGR98  when he's back if it was indeed just a woopsie, since it looks super intended.
@robodoo r+

RicGR98:
It was indeed intentional however we planned a task to allow Odoo employees to  top up their credits free of charge (https://www.odoo.com/odoo/project/967/tasks/5449358). This fixes the complicated testing part.

But I agree that it won't help tech support. We should reconsider, I'll recheck with FHE

RicGR98 (approved):
Validated with FHE, let's revert that :)

Thanks!

@robodoo r+

PR #246725, opened 2026-02-02
[FIX] account: filter only active account on journal on_change

rapa-odoo:
### Steps to Reproduce:
1. Go to Accounting > Create a journal.
2. Archive the company’s default income/expense or cash diff. accounts.
3. Change the journal type.
3. Observe that an archived account is assigned by the onchange method.

### Issue:
- Prior to saas-18.3, no onchange logic existed for assigning default accounts. With the introduction of the [onchange]
(https://github.com/odoo/odoo/commit/334a8161cd2feb092cafedc35d94645900b40518) , archived income/expense or default cash diff. accounts can be assigned to journals.
- This creates an inconsistency, as the domain [defined] (https://github.com/odoo/odoo/blob/saas-18.3/addons/account/models/account_journal.py#L70-#L82) for `default_account_id` only returns active accounts.

### Solution:
- Since the domain for `default_account_id` returns only active accounts, this commit aims to update the onchange [logic](https://github.com/odoo/odoo/blob/saas-18.3/addons/account/models/account_journal.py#L540-#L551) in order to follow the same restriction, ensuring that only active accounts are assigned thereby preventing inconsistent behavior.

OPW-5501539

Current behavior before PR:

---

### DIFF
--- a/addons/account/models/account_journal.py
+++ b/addons/account/models/account_journal.py
@@ -542,13 +542,16 @@ class AccountJournal(models.Model):
             journal.default_account_id = False
             journal.profit_account_id = False
             journal.loss_account_id = False
-            if journal.type == 'sale':
-                journal.default_account_id = journal.company_id.income_account_id
-            elif journal.type == 'purchase':
-                journal.default_account_id = journal.company_id.expense_account_id
+            company = journal.company_id
+            if journal.type == 'sale' and company.income_account_id.active:
+                journal.default_account_id = company.income_account_id
+            elif journal.type == 'purchase' and company.expense_account_id.active:
+                journal.default_account_id = company.expense_account_id
             elif journal.type in ('cash', 'bank'):
-                journal.profit_account_id = journal.company_id.default_cash_difference_income_account_id
-                journal.loss_account_id = journal.company_id.default_cash_difference_expense_account_id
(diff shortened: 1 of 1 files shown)


alialfie (changes requested):
Thanks for the fix! I love the commit message, very detailed, but maaaybe a bit too detailed? The shell trace can be replaced with a sentence or two and it would convey the same meaning and make the commit message shorter overall. Other than that, LGTM!

rapa-odoo:
> Thanks for the fix! I love the commit message, very detailed, but maaaybe a bit too detailed? The shell trace can be replaced with a sentence or two and it would convey the same meaning and make the commit message shorter overall. Other than that, LGTM!

Thanks for the review! I have made suggested changes in the commit message.

alialfie (changes requested):
Not exactly what I meant. I see you shortened the shell trace, but I was actually asking to remove it altogether from the commit message and replace it with a sentence or two that conveys the same point. IF you really want to keep it in the commit message then maybe explain why it's important? :)

alialfie (changes requested):
Nice! Although I just noticed one small typo in the numbering of the steps. Could you fix it? I'll r+ after

alialfie (approved):
Thanksss
robodoo r+

PR #246728, opened 2026-02-02
[IMP] l10n_sa{_edi,_pos}: improve partner company/individual detection

ijja-odoo:
Before this commit:
- Partner identification relied only on the VAT field.

After this commit:
- Partner identification in Saudi Arabia also considers Identification Scheme and Identification Number.
- This improves detection of companies vs individuals when VAT is missing.

task-5866942

---

### DIFF
--- a/addons/l10n_sa_edi/models/res_partner.py
+++ b/addons/l10n_sa_edi/models/res_partner.py
@@ -1,5 +1,7 @@
 from odoo import fields, models, api
 
+COMPANY_SCHEMES = {'CRN', 'MOM', 'MLS', '700', 'SAG', 'OTH'}
+
 
 class ResPartner(models.Model):
     _inherit = 'res.partner'
@@ -22,6 +24,23 @@ class ResPartner(models.Model):
     ], default="OTH", string="Identification Scheme", help="Additional Identification Scheme for the Seller/Buyer")
 
     l10n_sa_edi_additional_identification_number = fields.Char("Identification Number (SA)", help="Additional Identification Number for the Seller/Buyer")
+
+    @api.depends('l10n_sa_edi_additional_identification_scheme', 'l10n_sa_edi_additional_identification_number')
+    def _compute_is_company(self):
+        """ Determines if a Saudi partner is a company or an individual based on VAT and
+        additional identification fields.
+        """
+        l10n_sa_commercial_partners = self.filtered(
+            lambda p: (
+                p.country_code == 'SA'
+                and p.commercial_partner_id == p
+                and p._is_vat_void(p.vat)
+                and p.l10n_sa_edi_additional_identification_number
+                and p.l10n_sa_edi_additional_identification_scheme in COMPANY_SCHEMES
+            )
+        )
+        l10n_sa_commercial_partners.is_company = True
+        super(ResPartner, self - l10n_sa_commercial_partners)._compute_is_company()
 
     @api.model
     def _commercial_fields(self):
--- a/addons/l10n_sa/models/account_move.py
+++ b/addons/l10n_sa/models/account_move.py
@@ -133,7 +133,6 @@ class AccountMove(models.Model):
     def _l10n_sa_is_simplified(self):
         """
             Returns True if the customer is an individual, i.e: The invoice is B2C
-        :return:
         """
         self.ensure_one()
-        return not self.partner_id.vat
+        return not self.commercial_partner_id.is_company
--- a/addons/l10n_sa_edi/tests/common.py
+++ b/addons/l10n_sa_edi/tests/common.py
@@ -117,8 +117,8 @@ class TestSaEdiCommon(AccountEdiTestCommon):
             'country_id': cls.saudi_arabia.id,
             'state_id': cls.riyadh.id,
             # Simplified invoices use different ID schemes
-            'l10n_sa_edi_additional_identification_scheme': 'MOM',  # Momra License
-            'l10n_sa_edi_additional_identification_number': '3123123213131',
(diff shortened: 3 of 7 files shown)


addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -23,6 +23,27 @@ class ResPartner(models.Model):
 
     l10n_sa_edi_additional_identification_number = fields.Char("Identification Number (SA)", help="Additional Identification Number for the Seller/Buyer")
 
+    @api.depends('l10n_sa_edi_additional_identification_scheme', 'l10n_sa_edi_additional_identification_number')
+    def _compute_is_company(self):
+        """
+        Determines if a Saudi partner is a company or an individual based on VAT and
+        additional identification fields.
+        """
+        l10n_sa_commercial_partners = self.filtered(lambda p: p.country_code == 'SA' and p.commercial_partner_id == p)
```
eapa-odoo:
```suggestion
        l10n_sa_commercial_partners = self.filtered(lambda p: p.country_code == 'SA' and p.commercial_partner_id == p and not p.vat)
```
You can also filter records where VAT is not set, which helps reduce the number of conditions below.

eapa-odoo (commented):
@ijja-odoo Thanks for the PR.
Since we're only updating 3 modules, it would be better to mention that in the commit title.
`[IMP] l10n_sa{_edi,_pos}: improve partner company/individual identification`

addons/l10n_sa_edi/models/res_partner.py
ijja-odoo: Instead of directly checking, `_is_vat_void` method is more suitable, here is what the method looks like:
https://github.com/odoo/odoo/blob/2b49b2182079020eb9bc605b03544f0570f4f39f/odoo/addons/base/models/res_partner.py#L916-L919

However, we still need to call that method again, either for filtering in `filtered()` method or for an explicit check. So I kept the `_is_vat_void` call as if condition.

Final call is up to you. However if we decided to filter invalid/empty `VATs` then we will have to wait for the YGU's reply as in the base method it uses `commercial_partner_id`.

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -23,6 +23,27 @@ class ResPartner(models.Model):
...
+        for partner in l10n_sa_commercial_partners:
```
eapa-odoo: I think we can skip this for loop, and we can make use for filter instead.

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -23,6 +23,26 @@ class ResPartner(models.Model):
 
     l10n_sa_edi_additional_identification_number = fields.Char("Identification Number (SA)", help="Additional Identification Number for the Seller/Buyer")
 
+    @api.depends('l10n_sa_edi_additional_identification_scheme', 'l10n_sa_edi_additional_identification_number')
+    def _compute_is_company(self):
+        """
+        Determines if a Saudi partner is a company or an individual based on VAT and
```
jbw-odoo:
```suggestion
        """ Determines if a Saudi partner is a company or an individual based on VAT and
```

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -23,6 +23,26 @@ class ResPartner(models.Model):
...
+                p.country_code == 'SA'
+                and p.commercial_partner_id == p
+                and p._is_vat_void(p.vat)
+                and p.l10n_sa_edi_additional_identification_number
+                and p.l10n_sa_edi_additional_identification_scheme in {'CRN', 'MOM', 'MLS', '700', 'SAG', 'OTH'}
+            )
+        )
+        l10n_sa_commercial_partners.write({
+            'is_company': True,
+        })
```
jbw-odoo:
```suggestion
        l10n_sa_commercial_partners.is_company = True
```

jbw-odoo (commented):
Hello,
Thanks for the PR. I have this comments :

addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -23,6 +23,26 @@ class ResPartner(models.Model):
...
+        Determines if a Saudi partner is a company or an individual based on VAT and
+        additional identification fields.
+        """
+        l10n_sa_commercial_partners = self.filtered(
+            lambda p: (
+                p.country_code == 'SA'
+                and p.commercial_partner_id == p
+                and p._is_vat_void(p.vat)
+                and p.l10n_sa_edi_additional_identification_number
+                and p.l10n_sa_edi_additional_identification_scheme in {'CRN', 'MOM', 'MLS', '700', 'SAG', 'OTH'}
```
jbw-odoo:
```suggestion
                and p.l10n_sa_edi_additional_identification_scheme in IS_COMPANY_SCHEMES
```

jbw-odoo (commented):
Idk why this was not included in my first review

addons/l10n_sa_edi/models/res_partner.py
ijja-odoo: But, `l10n_sa_commercial_partners` may contain more than one record.

jbw-odoo: Doesn't matter, it will still work.

ijja-odoo: I thought it would give `Singleton` error, but surprisingly it works. Changed it, thanks :+1:

jbw-odoo (approved):

PR #246729, opened 2026-02-02
[FIX] sale: merge payment wizard's duplicate _compute_warning_message

mma1377:
After #245772, the compute method `_compute_warning_message` is duplicated.

runbot-238757


Feyensv (approved):
robodoo r+

PR #246731, opened 2026-02-02
[FIX] point_of_sale,account_peppol: Fix concurrent update when sending invoices

clbr-odoo:
Steps to reproduce:
- Set up a test company with Peppol enabled
- Create a Peppol customer and ensure they can receive invoices
- In the POS settings, enable the “Self-service invoicing” option
- Create a POS order linked to the Peppol customer
- Log in as this customer
- Enter the POS order information to access the invoice download view (from /pos/ticket)
- Repeatedly click the “Request invoice” button
- On the Peppol side, you will notice that the same invoice is sent as many times as the button is clicked

Why ?
This error is not really Peppol related, it can actually happen with all EDI making API call: the transaction is rolled-back when Odoo tries to create multiple invoices with a serialization error but the API call happened so, in this cas, the invoice is sent over Peppol.

opw-5469467

### DIFF
--- a/addons/account_peppol/models/account_move_send.py
+++ b/addons/account_peppol/models/account_move_send.py
@@ -1,3 +1,5 @@
+import logging
+
 from base64 import b64encode
 from datetime import timedelta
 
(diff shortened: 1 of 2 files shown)


smetl (approved):
@robodoo r+

smetl (approved):
@robodoo r+

PR #246732, opened 2026-02-02
[FIX] website_slides: Show all courses properly on homepage

sben-odoo:
Steps to reproduce:
===================
1. Go to eLearning & select any course.
2. Update the course description with a link inside a paragraph.
3. Go to the website homepage where the course is displayed.
4. Observe that the course card link is broken.

Cause:
======
Since the entire course card is wrapped in an anchor tag, adding another anchor tag will break the HTML structure, leading to unexpected behavior.

Solution:
=========
Modify the course card template to ensure that only the footer section is wrapped in an anchor tag. This way, any links within the course description will function correctly without breaking the overall structure.

opw-5383136

---

### DIFF
--- a/addons/website_slides/views/website_slides_templates_homepage.xml
+++ b/addons/website_slides/views/website_slides_templates_homepage.xml
@@ -362,17 +362,23 @@
 </template>
 
 <template id='course_card' name="Course Card">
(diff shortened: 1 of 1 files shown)


Ephoniz (approved):
LGTM 👍🏽

awa-odoo (commented):
Hello @sben-odoo 

That is a bit too many changes to a stable template, could break people customisations.

Can we try to find something lighter that still works?

Thanks & cheers!

PR #246737, opened 2026-02-02
[FIX] mail: fix insertion of mixed up many commands

tsm-odoo:
Many fields can be updated in the discuss store in different modes: `ADD`, `DELETE`, `REPLACE`.

The `REPLACE` case is the most frequent one, so the server doesn't bother sending it wrapped up as a command but instead return the array that should replace the current field value.

However, the client doesn't handle well replace being mixed up with other commands.

This commit ensures many commands are normalized on the client side. This also allows to simplify the `updateRelationMany` function since we only have one command format.

Current behavior before PR:

---

### DIFF
--- a/addons/mail/static/tests/core/record.test.js
+++ b/addons/mail/static/tests/core/record.test.js
@@ -1589,3 +1589,22 @@ test("Record exists is reactive", async () => {
     thread.delete();
     await expect.waitForSteps(["thread does not exist"]);
 });
+
+test("Can insert mixed up many commands", async () => {
+    (class Message extends Record {
+        static id = "id";
+        id;
+    }).register(localRegistry);
+    (class Thread extends Record {
+        static id = "name";
+        name;
+        messages = fields.Many("Message");
+    }).register(localRegistry);
+    const store = await start();
+    const thread = store.Thread.insert("General");
+    store.Thread.insert({
+        name: "General",
+        messages: [{ id: 1 }, { id: 2 }, ["ADD", [{ id: 3 }, { id: 4 }]], ["DELETE", [{ id: 4 }]]],
+    });
+    expect(thread.messages.map((msg) => msg.id).sort()).toEqual([1, 2, 3]);
+});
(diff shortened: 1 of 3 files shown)


addons/mail/static/tests/core/record.test.js
```diff
@@ -1589,3 +1589,22 @@ test("Record exists is reactive", async () => {
...
+    (class Thread extends Record {
+        static id = "name";
+        name;
+        messages = fields.Many("Message");
+    }).register(localRegistry);
+    const store = await start();
+    const thread = store.Thread.insert("General");
+    store.Thread.insert({
+        name: "General",
+        messages: [{ id: 1 }, { id: 2 }, ["ADD", [{ id: 3 }, { id: 4 }]], ["DELETE", [{ id: 4 }]]],
```
seb-odoo, line 1607: I wonder if this is really the intended syntax or if Store is just bugged.

If `replace` and `add` are merged together, I'd rather expect the format to be `[[{...}, {...}], ["ADD", {}], ...]`

Single replace uses single list shortcut, but if we have multiple commands, it should contain a list of command (whether `replace` is explicit or not in the first command is up to debate), and not a mix of values and commands

tsm-odoo: `Many()` instances don't know about each other and I don't think it's worth adding an extra layer of logic to handle something that should be done anyway on the client side. Note that store doesn't merge `add` nor `delete` commands either.

seb-odoo: `_add_to_store` of `Many` does handle that, it's how we end up with this syntax. It's just doing it badly.

seb-odoo: We should have either:
- a single record (dict) (or empty value, considered as no record?)
- a list of records (list of dict) (empty list is considered empty list of record and not empty list of command?)
- a list of commands (list of list)

A command should be either:
- a list of records (list of dict), implied it is a replace
- a tuple in format (mode, list of records)

The format of this test with a list containing both records and commands doesn't really make sense. If anything, it should be considered as 2 replace, 1 add and 1 delete, where the 2nd object replaces the first.

tsm-odoo: My bad, the server is actually doing this correctly. i guess the normalize part can be done in the version PR then.

tsm-odoo: I already fixed it on my branch. The issue still exists, opening a new PR as it seems I can't re-open a PR once the branch has been force pushed.

PR #246743, opened 2026-02-02
[FIX] point_of_sale: adjust review button width on ticket screen

meji-odoo:
Before this commit:
---
- On the ticket screen in small UI, two btns are shown: Review and Load Order.
- The Load Order button is only visible when the selected order is not paid.
- When the order is paid, only the Review btn is shown, but it takes 50% width.

After this commit:
---
- When the order is paid, the Review button expands to take 100% width.

task-5892189

---


davidmonnom (approved):
@robodoo r+

PR #237876, opened 2025-11-28
[FIX] website_sale: prevent horizontal scroll in dynamic products

alap-odoo:
Steps to reproduce:
1. Drag and drop the dynamic products snippet.
2. Select it and change the content width to **max**.

Issue:
When the content width is set to **max**, an unnecessary horizontal scroll appears.

Reason:
The issue occurs because the `previous` and `next` navigation buttons were not properly positioned.

Fix:
For devices larger than "mobile", the `previous` and `next` buttons are re-positioned, horizontally inward by "**50%**" of their own width with the help of `transform` property. This keeps the controls within the visible area and prevents horizontal scrolling.

task-5090468

Before:
image

After:
image

PR #246744, opened 2026-02-02
[CLN] website: remove default arrow position option

alap-odoo:
This PR remove the "**default**" arrow position option and its related CSS from the dynamic product snippet and renames the "**inside**" option to "**default**". This change is based on the discussion here - https://github.com/odoo/odoo/pull/237876#issuecomment-3777902859


sobo-odoo (changes requested):
@alap-odoo @livaios  I don't think we should remove the default option, because the dynamic snippet carousel is not for eCommerce only => any record can be added in it (in debug mode) and removing the possibility to have the arrows partially outside may be problematic :open_mouth: 

For example, if we add blog posts:
- the default arrows look like this

  image
- and the inside ones like this

  image

=> We can barely see the inside arrows depending on the blog post cover :grimacing: 

Given that before the task adding the option to position the arrows, they were looking like this :arrow_down:, I think it is better to keep the option allowing them to be partially outside :slightly_smiling_face: 
  
image

PR #246748, opened 2026-02-02
[FIX] website, *: restrict gear icon in backend mode 

assk-odoo:
Steps to reproduce:
1.Install the Mass Mailing and Project modules.
2.Select any text to display the toolbar.
3.Select the Link option to open the link popover.
4.You will see a gear icon (in backend mode).

Before this commit:
The link popover displayed the gear icon in backend mode as well,
including in Mass Mailing and Project descriptions, where advanced link
options are not relevant.

After this commit
The gear icon is now shown only on the Website, ensuring that advanced
link options are available exclusively in the website context.
Additionally, the link attribute handling has been moved from
html_editor to website.

task-5112817


addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -511,60 +539,15 @@ export class LinkPlugin extends Plugin {
...
-                }
-            } else if (url) {
+                cursorsToRestore = this.applyLinkCallback(params, cursorsToRestore);
+            } else if (params.url) {
                 // prevent the link creation if the url field was empty
 
                 // create a new link with current selection as a content
-                if ((selectionTextContent && selectionTextContent === label) || isImage) {
-                    const link = this.createLink(url);
-                    if (relValue) {
```
divy-odoo, line 573: You removed this condition which will break the image link option. Try to apply advance option in image link, If you want to check.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -583,8 +566,9 @@ export class LinkPlugin extends Plugin {
                         link.append(content);
                         link.normalize();
                         cursorsToRestore = null;
-                        selection = this.dependencies.selection.getEditableSelection();
-                        const anchorClosestElement = closestElement(selection.anchorNode);
+                        const anchorClosestElement = closestElement(
+                            this.dependencies.selection.getEditableSelection().anchorNode
+                        );
```
divy-odoo: Why this changes?

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -474,6 +474,34 @@ export class LinkPlugin extends Plugin {
         );
     }
 
+    applyLinkCallback(params, cursorsToRestore) {
```
divy-odoo: May be you should choose more specific name for this method : )  Also, do you really need a `cursorToRestore` as params?

addons/html_editor/static/src/main/link/link_popover.js
```diff
@@ -303,32 +267,26 @@ export class LinkPopover extends Component {
...
-            this.state.attachmentId,
-            relValue
-        );
+        this.props.onApply({
+            url: this.state.url,
+            label: this.state.label,
+            classes: this.classes,
+            customStyle: this.customStyles,
+            attachmentId: this.state.attachmentId,
+            isImage: this.state.isImage,
```
divy-odoo: Why we need to pass the `isImage`? it's bit weird & confusing, it can be check in place where you need i guess.

addons/website/static/src/xml/html_editor.xml
```diff
@@ -49,6 +49,50 @@
     <xpath expr="//a[hasclass('o_we_url_link')]" position="attributes">
         <attribute name="t-on-click">onClickForcePreviewMode</attribute>
     </xpath>
+    <xpath expr="//div[hasclass('o_link_popover_container')]" position="replace">
```
divy-odoo: This will again breaks the image link popover. your Xpath will choose this element: https://github.com/odoo/odoo/pull/246748/changes?diff=unified#diff-48d30943cef52d00bf65cecd6c0ee6e5f35a29dd9ebc85c6ac94ef47fe39adc2R6 
While you want to replace the another element. may be this: https://github.com/odoo/odoo/pull/246748/changes?diff=unified#diff-48d30943cef52d00bf65cecd6c0ee6e5f35a29dd9ebc85c6ac94ef47fe39adc2L47

Can you please check again?

divy-odoo (commented):
Hello @assk-odoo, I've dropped few comments. Can you please have a look? Thanks : )

addons/html_editor/static/src/main/link/link_plugin.js
assk-odoo: We changed this because selection was defined as a const earlier, but in the current implementation it is no longer required, so it has been removed.

addons/html_editor/static/src/main/link/link_plugin.js
assk-odoo: May be we go to this name : applyLinkChanges.
Yes, it is needed. cursorToRestore represents the current cursor state and indicates whether the selection is still valid after DOM updates. Since it is not mutated globally, the cursor state may not be updated correctly in some cases. so updating cursorsToRestore explicitly avoids these issues and ensures correct cursor restoration.

addons/html_editor/static/src/main/link/link_popover.js
assk-odoo: Sorry ! It's by mistake add :smiling_face_with_tear:

addons/html_editor/static/src/main/link/link_plugin.js
assk-odoo: I want to verify that, for both image and non-image cases, the attribute is added correctly even without this condition.then this condition may not be needed here.
image

Previously, this condition may not have been needed because we were passing only six arguments in onchange function. However, in applyCallback we are now receiving seven parameters. At that time, the relValue was undefined.
For image links and first-time link creation, the code goes into the else if (url) condition, and at that moment relValue is still undefined.
But when we click the Apply button, the onClickApply function is called. At that point, the link already exists and relValue has been added for both images and first-time link creation.
Therefore, we do not need to handle relValue inside the else if condition.

Conclusion :
This condition was not needed previously and is also not needed now.

addons/website/static/src/js/editor/html_editor.js
```diff
@@ -74,6 +103,40 @@ patch(LinkPopover.prototype, {
...
+            .filter((key) => relOptions[key].isChecked)
+            .join(" ");
+        const originalOnApply = this.props.onApply;
+        this.props.onApply = (params) => {
+            params.linkTarget = this.linkState.linkTarget;
+            params.relValue = relValue;
+            originalOnApply(params);
+        };
+        super.onClickApply();
+    },
```
divy-odoo: I think, we should not touch the props value. here you changed the `this.props.onApply` but it never reverted back. so there might be a possibility of breaking the things in future. IMO the original value should be reassigned after the super call ( `this.props.onApply = originalOnApply` ). but instead what if we simply create new method `prepareParams()` which will return the params to be used while calling onApply? something like this,


html_editor:

```js
   prepareParams() {
       return {
            url: this.state.url,
            label: this.state.label,
            classes: this.classes,
            customStyle: this.customStyles,
            attachmentId: this.state.attachmentId,
        }
   }
   onClickApply() {
         const params = prepareParams();
         this.props.onApply(parmas);
    },
```
Website: 
```js
   prepareParams() {
        const relOptions = this.linkState.relAttributeOptions;
        const relValue = Object.keys(relOptions)
            .filter((key) => relOptions[key].isChecked)
            .join(" ");
       return {
            url: this.state.url,
            label: this.state.label,
            classes: this.classes,
            customStyle: this.customStyles,
            attachmentId: this.state.attachmentId,
            linkTarget = this.linkState.linkTarget,
            relValue = relValue,
        }
   }
```

divy-odoo (commented):
Hello @assk-odoo, Small comment and it's good to go from my end. 

Also, don't forgot to squash the commit & adapt the commit message accordingly. Thank : )

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -494,6 +496,16 @@ export class LinkPlugin extends Plugin {
         );
     }
 
+    applyAttributes(element, attributes) {
+        for (const [attr, value] of Object.entries(attributes || {})) {
+            if (value) {
+                element.setAttribute(attr, value);
+            } else if (value !== undefined) {
```
divy-odoo: I din't get in which case the value will be `undefined` & we don't won't remove attribute?

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -532,46 +544,25 @@ export class LinkPlugin extends Plugin {
...
-                    this.linkInDocument.removeAttribute("target");
-                }
+                this.applyAttributes(this.linkInDocument, params.attributes);
                 if (!isImage) {
-                    if (classes) {
-                        this.linkInDocument.className = classes;
-                    } else {
-                        this.linkInDocument.removeAttribute("class");
-                    }
+                    this.applyAttributes(this.linkInDocument, params.attributes);
```
divy-odoo: Do we need this call now? the same method will already been executed above if.

addons/html_editor/static/src/main/link/link_plugin.js
divy-odoo: Here where it'll fails: https://drive.google.com/file/d/1hVAae8e8YIknOmxX3Rdm6d_O4qJp7tWu/view?usp=drivesdk

addons/html_editor/static/src/main/link/link_popover.xml
```diff
@@ -15,29 +15,22 @@
...
-                                </div>
-                                <t t-if="state.linkTarget === '_blank'">
+                                <t t-if="Object.values(state.advancedAttributeOptions).find((opt) => opt.attribute === 'target')?.isChecked">
                                     <div class="o_seo_option_row o_seo_option_child d-flex justify-content-between align-items-center py-1 gap-2 ps-4">
                                         <span>noopener</span>
                                         <CheckBox
-                                            value="state.relAttributeOptions.noopener.isChecked"
-                                            onChange="toggleRelAttr.bind(this, 'noopener')"/>
+                                            value="state.advancedAttributeOptions.noopener.isChecked"
+                                            onChange="toggleAdvancedAttr.bind(this, 'noopener')"/>
```
divy-odoo: If you are still managing the `noopener` option manually, what's the mean of adding into plugin resource?

addons/website/static/src/builder/plugins/website_link_plugin.js
```diff
@@ -0,0 +1,56 @@
...
+                    "Prevents the new page from accessing the original window (security)"
+                ),
+                attribute: "rel",
+                value: "noopener",
+            },
+        ],
+        link_popovers: [
+            withSequence(20, {
+                PopoverClass: WebsiteLinkPopover,
+                isAvailable: () => true,
```
divy-odoo: Can you please describe this, why always true?

addons/website/static/src/builder/plugins/website_link_popover/website_link_popover.js
```diff
@@ -0,0 +1,55 @@
...
+                            ? currentTargetValue === option.value
+                            : false,
+                },
+            ])
+        );
+    }
+
+    toggleAdvancedOptions() {
+        this.state.showAdvancedOptions = !this.state.showAdvancedOptions;
+    }
```
divy-odoo: Move this method inside html_editor.

addons/website/static/src/builder/plugins/website_link_popover/website_link_popover.js
```diff
@@ -0,0 +1,55 @@
...
+
+    toggleAdvancedAttr(attr) {
```
divy-odoo: This method isn't overridden method & used inside html_editor. this should be inside html_editor.

addons/website/static/src/builder/plugins/website_link_plugin.js
assk-odoo: In our case we want to replace the default LinkPopover with WebsiteLinkPopover for all links because in website we adds the advanced attribute options (rel, target, etc.). Therefore isAvailable returns true so this popover is always applicable.
The editor selects the first popover whose isAvailable() returns true (getActivePopover()), and withSequence(20) ensures this popover is evaluated before the default one.

divy-odoo (commented):
LGTM, thanks : )

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -282,6 +282,8 @@ export class LinkPlugin extends Plugin {
             }),
         ],
 
+        advanced_popover_options: [],
```
FrancoisGe, line 286: You should add it in the doc. `community/addons/html_editor/static/src/@types/plugins.d.ts`

addons/website/static/src/builder/plugins/website_link_popover/website_link_popover.js
```diff
@@ -0,0 +1,38 @@
...
+                option.id,
+                {
+                    ...option,
+                    isChecked:
+                        (option.attribute === "rel" && currentRelValues.includes(option.value)) ||
+                        currentTargetValue === option.value,
+                },
+            ])
+        );
+    }
```
FrancoisGe: Why do you need to create a custom `WebsiteLinkPopover` ?
It doesn t look generic if you need to customize it in each editor

addons/html_editor/static/src/main/link/link_popover.js
```diff
@@ -160,37 +136,36 @@ export class LinkPopover extends Component {
...
+
+    prepareParams() {
+        return {
+            label: this.state.label,
+            attachmentId: this.state.attachmentId,
+            attributes: {
+                href: this.state.url,
+                class: this.classes,
+            },
+        };
```
FrancoisGe: Maybe a better name. When I read it I don t know whish params is it

addons/html_editor/static/src/main/link/link_popover.js
```diff
@@ -160,37 +136,36 @@ export class LinkPopover extends Component {
...
     }
 
-    toggleRelAttr(attr) {
-        const option = this.state.relAttributeOptions[attr];
+    toggleAdvancedAttr(attr) {
+        const option = this.state.advancedAttributeOptions[attr];
         option.isChecked = !option.isChecked;
+        if (option.attribute === "target" && !option.isChecked) {
+            this.state.advancedAttributeOptions.noopener.isChecked = false;
+        }
```
FrancoisGe: Why this changes ?

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -563,15 +557,12 @@ export class LinkPlugin extends Plugin {
...
+            } else if (attributes.href) {
                 // prevent the link creation if the url field was empty
 
                 // create a new link with current selection as a content
                 if ((selectionTextContent && selectionTextContent === label) || isImage) {
-                    const link = this.createLink(url);
-                    if (relValue) {
-                        link.setAttribute("rel", relValue);
-                    }
+                    const link = this.createLink(attributes.href);
```
FrancoisGe, line 566: Why we don t care the other attributes ?

addons/website/static/tests/builder/linkpopover_website.test.js
```diff
@@ -254,3 +252,129 @@ test("Focus should be on label when adding a new URL", async () => {
...
+        anchorNode: linkText.firstChild,
+        anchorOffset: 1,
+    });
+    await waitFor(".o-we-linkpopover");
+    await contains(".o_we_edit_link").click();
+    await waitFor(".o_we_href_input_link");
+    await contains(".o_we_href_input_link").click();
+    await contains(".o-we-linkpopover .fa-gear").click();
+    expect(".o_advance_option_panel").toHaveCount(1);
+    expect(".o_advance_option_panel .o_seo_option_row").toHaveCount(4);
```
FrancoisGe, line 273: You just move this list of tests, or you change something ?

assk-odoo: No changes. just moved the test case from html_editor to website.

addons/html_editor/static/src/main/link/link_plugin.js
assk-odoo: Previously, we only needed to pass a single URL attribute when calling createLink, so we can pass only `attributes.href` here. We may not need to pass any additional attributes. I didn’t find any cases where additional attributes are required.

image

addons/html_editor/static/src/main/link/link_popover.js
assk-odoo: The noopener option is dependent on the target="_blank" (Open in New Tab) option. 
when we unchecking "Open in New Tab" automatically uncheck noopener as well otherwise a that option is hidden but active noopener state would persist and produce incorrect link attributes on save.

addons/website/static/src/builder/plugins/website_link_popover/website_link_popover.js
assk-odoo: Actually, our option is only for the website side. So, if we think about  module perspective, our code is written in the website module, not in the editor. Therefore, we cannot handle this directly in the `link_popover`. Instead, we extend that component on the website side and manage the option there, so we created a custom `WebsiteLinkPopover`.

Creating a website-specific LinkPopover makes more sense IMO. Let me know if I missed anything 🙂.

FrancoisGe: My feeling is that this could easily be described with a generic pattern. Do you just want a second-order option with a visibility condition, or am I missing something?

addons/html_editor/static/src/main/link/link_popover.js
FrancoisGe: So we should fix it in the other version ?

assk-odoo: This is already working fine in the stable version. However, here the target="_blank" (Open in New Tab) option is handled on the resource side, so we need to modify our code accordingly. That is the reason for these changes. :slightly_smiling_face:

addons/website/static/src/builder/plugins/website_link_popover/website_link_popover.js
assk-odoo: >Do you just want a second-order option with a visibility condition, or am I missing something?

We handle the second-order option on the resource side here. We pass the `requires` key for the noopener option. https://github.com/odoo/odoo/pull/246748/changes#diff-525c524f2178934650b660b01e70a1cd704ec73afb1996bba0dc111c4daed9bdR41-R48

>My feeling is that this could easily be described with a generic pattern. 

Okay! In my opinion, we can pass an additional key for all rel attributes in `advanced_popover_options` on the resource side. Based on that, we can modify the setup and prepareLinkParams methods in the LinkPopover side  and remove the WebsiteLinkPopover component. 

Let me know if you have a better approach 🙂

addons/html_editor/static/src/main/link/link_popover.js
FrancoisGe: It should probably be generic. In that case, you could use your `required` to check if a required element is disabled, and if so, I'll disable myself.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -563,15 +557,12 @@ export class LinkPlugin extends Plugin {
...
                 }
-            } else if (url) {
+            } else if (attributes.href) {
                 // prevent the link creation if the url field was empty
 
                 // create a new link with current selection as a content
                 if ((selectionTextContent && selectionTextContent === label) || isImage) {
-                    const link = this.createLink(url);
-                    if (relValue) {
-                        link.setAttribute("rel", relValue);
```
dmo-odoo, line 574: Is the case reported by @divy-odoo solved? I don't see a reply.

assk-odoo: Yes ! it is already resolved.

dmo-odoo (approved):

PR #246749, opened 2026-02-02
[IMP] point_of_sale: prevent zero value for coins/bills

rhe-odoo:
- Add @api.constrains to block saving a pos.bill with value = 0
- Add decoration-danger in list view to highlight zero values in red

task-id: 5901790

---

### DIFF
--- a/addons/point_of_sale/views/pos_bill_view.xml
+++ b/addons/point_of_sale/views/pos_bill_view.xml
@@ -7,7 +7,7 @@
             <form string="Bills">
                 <sheet>
                     <group>
-                        <field name="name" />
+                        <field name="name" required="1"/>
                         <field name="value" />
                         <field name="pos_config_ids" widget="many2many_tags" />
                     </group>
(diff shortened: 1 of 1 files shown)


addons/point_of_sale/models/pos_bill.py
```diff
@@ -8,10 +8,16 @@ class PosBill(models.Model):
     _description = "Coins/Bills"
     _inherit = ["pos.load.mixin"]
 
-    name = fields.Char("Name")
+    name = fields.Char("Name", required=True)
```
davidmonnom: You cannot really add "required" in stable, because if the database already has some records that do not match it, the constrains will not be added in pg.

addons/point_of_sale/models/pos_bill.py
```diff
@@ -8,10 +8,16 @@ class PosBill(models.Model):
...
     value = fields.Float("Value", required=True, digits=(16, 4))
     pos_config_ids = fields.Many2many("pos.config", string="Point of Sales")
 
+    @api.constrains('value')
```
davidmonnom: Same for the api.constrains one.

davidmonnom (changes requested):
There modification are good to master with an upgrade script.

For now, you can add required=True directly in the view

PR #246752, opened 2026-02-02
[IMP] web,calendar: improve mobile display

amdi-odoo:
Improve the calendar mobile view:

- replace the "today" button from a calendar icon with the day to a simple "Today" button to be more explicit.
- in year view, make sure the months are taking the full width of the screen so that they are easier to interact with.
- in the quick create form, make sure the partner_ids field is taking the full modal width like the other fields. Also prevent the "all day" field from being squeezed on the right, instead if there's not enough space display it on a new line.

Task-5717052

---

### DIFF
--- a/addons/calendar/views/calendar_views.xml
+++ b/addons/calendar/views/calendar_views.xml
@@ -353,30 +353,33 @@
                     keep the last computed duration in the front-end model and send it with "start"
                     when it changes-->
                     <!--field is also used for custom behavior in default_get, check before removing-->
-                    <field name="duration" invisible="1" force_save="1"/>
-                    <field name="start" nolabel="1"
-                        widget="daterange" options="{'end_date_field': 'stop'}"
-                        placeholder="Dates" class="w-100 mb-0"
-                        invisible="allday"
-                    />
-                    <field name="start_date" nolabel="1"
-                        widget="daterange" options="{'end_date_field': 'stop_date'}"
-                        placeholder="Dates" class="w-100 mb-0"
-                        invisible="not allday"
-                    />
-                    <span class="ms-1 w-25">
-                        <!-- This is needed so that show_as can be set as 'free' when creating all day events -->
-                        <field name="show_as" invisible="1"/>
-                        <field name="allday" class="mb-0" title="All Day" nolabel="1"/>
-                        <span>All day</span>
-                    </span>
-                    <field name="stop" invisible="1"/>
+                    <div class="d-flex align-items-center flex-wrap gap-2 w-100">
+                        <field name="duration" invisible="1" force_save="1"/>
+                        <field name="start" nolabel="1"
+                            widget="daterange" options="{'end_date_field': 'stop'}"
+                            placeholder="Dates" class="text-nowrap flex-grow-1 mb-0"
+                            invisible="allday"
+                        />
+                        <field name="start_date" nolabel="1"
+                            widget="daterange" options="{'end_date_field': 'stop_date'}"
+                            placeholder="Dates" class="text-nowrap flex-grow-1 mb-0"
+                            invisible="not allday"
+                        />
+                        <span class="text-nowrap w-auto">
+                            <!-- This is needed so that show_as can be set as 'free' when creating all day events -->
+                            <field name="show_as" invisible="1"/>
+                            <field name="allday" class="mb-0" title="All Day" nolabel="1"/>
+                            <span>All day</span>
+                        </span>
+                        <field name="stop" invisible="1"/>
+                    </div>
                 </div>
                 <div class="d-flex align-items-center pt-1 pb-2" colspan="2">
                     <i class="fa fa-fw flex-shrink-0 fa-user text-600 me-2" title="Participants"/>
                     <field name="partner_ids" nolabel="1"
                         widget="many2many_tags"
                         placeholder="Participants"
+                        class="w-100 mb-0"
                     />
                 </div>
                 <div class="d-flex align-items-center pt-1 pb-2" colspan="2">
--- a/addons/web/static/src/views/calendar/calendar_controller.xml
+++ b/addons/web/static/src/views/calendar/calendar_controller.xml
@@ -35,10 +35,8 @@
                         <ViewScaleSelector scales="scales" currentScale="model.scale" isWeekendVisible="state.isWeekendVisible" setScale.bind="setScale" toggleWeekendVisibility.bind="toggleWeekendVisibility" dropdownClass="'d-print-none order-3 order-lg-0'"/>
                         <button
                             class="btn btn-secondary o_calendar_button_today d-print-none order-2 order-lg-0 ms-auto ms-lg-0"
-                            t-att-class="env.isSmall ? 'btn-sm btn-light' : 'btn-secondary'"
-                            t-on-click.stop="() => this.setDate('today')"
-                        >   <span t-if="env.isSmall" class="position-relative pt-1"><t t-out="today"/><i class="fa fa-calendar-o position-absolute top-50 start-50 translate-middle fs-1"></i></span>
-                            <t t-else="">Today</t>
+                            t-on-click.stop="() => this.setDate('today')">
+                            Today
                         </button>
                         <h5 class="d-inline-flex ps-lg-2 ps-print-0 mb-0">
                             <t t-if="model.meta.scale === 'year'">
--- a/addons/web/static/src/views/calendar/calendar_controller_mobile.scss
+++ b/addons/web/static/src/views/calendar/calendar_controller_mobile.scss
@@ -1,13 +1,4 @@
 @include media-breakpoint-down(md) {
-    .o_calendar_header {
-        .o_calendar_button_today {
-            line-height: 1.9em;
-            > span {
-                font-size: .65rem
-            }
-        }
-    }
-
     .o_calendar_container {
         .o_other_calendar_panel {
             grid-area: 2 / 1 / 2 / 4;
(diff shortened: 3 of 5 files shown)


addons/calendar/views/calendar_views.xml
```diff
@@ -375,7 +375,7 @@
                     <field name="partner_ids" nolabel="1"
```
Odoonan, line 379: Note on the 'all day' while we're at it: from the screenshots, it is all squeezed on the right side, should we not move it to next line for xs screens?

addons/web/static/src/views/calendar/calendar_controller.xml
```diff
@@ -35,10 +35,8 @@
                         <ViewScaleSelector scales="scales" currentScale="model.scale" isWeekendVisible="state.isWeekendVisible" setScale.bind="setScale" toggleWeekendVisibility.bind="toggleWeekendVisibility" dropdownClass="'d-print-none order-3 order-lg-0'"/>
                         <button
                             class="btn btn-secondary o_calendar_button_today d-print-none order-2 order-lg-0 ms-auto ms-lg-0"
-                            t-att-class="env.isSmall ? 'btn-sm btn-light' : 'btn-secondary'"
-                            t-on-click.stop="() => this.setDate('today')"
-                        >   <span t-if="env.isSmall" class="position-relative pt-1"><t t-esc="today"/><i class="fa fa-calendar-o position-absolute top-50 start-50 translate-middle fs-1"></i></span>
-                            <t t-else="">Today</t>
+                            t-on-click.stop="() => this.setDate('today')">
+                            Today
```
Odoonan, line 39: Funny to have this but to hide navigation? Not related to your changes though

addons/web/static/src/views/calendar/calendar_controller_mobile.scss
```diff
@@ -40,6 +31,7 @@
 
         .o_calendar_sidebar_container {
             height: 100%;
+            z-index: $zindex-popover;
```
Odoonan: Mmh, this will basically put the sidebar always at the highest. Is this change only meant to hide the 'share'? Not sure it will work with all existing behaviors. Not sure there is a better option, but we could hide it locally

Odoonan: If no better solution, we might ask review from js team (juca) as it has just been reworked

addons/web/static/src/views/calendar/calendar_renderer.scss
```diff
@@ -570,6 +570,18 @@
                     width: 21rem;
                     margin: auto;
 
+                    // On mobile view, adapt the month width to be as close as possible to
+                    // the screen width while still using a round number of pixels divisible per 7.
```
Odoonan: Could we not simplify a bit to `28,35,42`? or at least remove the `22.75`? Seems quite overfitting.

Also, I sometimes have the scrolling bar disappearing on certain pivot sizes, probably add the bar width to the `+1rem `below at all times, to be sure.

Also, did you see this kind of specificity in other places in the css ? :)

Odoonan (commented):
Hi @amdi-odoo 

Looking good! Small screens, small review!

Cheers!

addons/calendar/views/calendar_views.xml
amdi-odoo: Made it so that it goes on a new line as soon as it doesn't have enough space 🙂

addons/web/static/src/views/calendar/calendar_controller.xml
amdi-odoo: On mobile, navigation is done using a swipe left/right on the screen so no need for the navigation arrows :slightly_smiling_face:

addons/web/static/src/views/calendar/calendar_controller_mobile.scss
amdi-odoo: I added an enterprise PR to hide the button in mobile when the calendar sidepanel is open :slightly_smiling_face:

addons/web/static/src/views/calendar/calendar_renderer.scss
amdi-odoo: I changed the `rem` computations to use `px` instead so that I can rely on a fixed unit. :slightly_smiling_face: I also don't have any float anymore :+1: 

No I didn't see this kind of specificity anywhere else :thinking: I guess this calendar is kind of a special case that's why we used a fixed width of 21 rem here :smiling_face_with_tear: (divisible per 7 to make sure each day cell has the same width)
I tried a bunch of things but the width set here is per month, meaning I can't use something dynamic like 100% as it'll misalign the different months. I need a fixed width for each one.
image

amdi-odoo (commented):
Hello @Odoonan 🙂 
Thank you for validating this one as well ! 
Some answers 😄

addons/web/static/src/views/calendar/calendar_renderer.scss
```diff
@@ -571,6 +571,19 @@
...
                     margin: auto;
 
+                    // On mobile view, adapt the month width to be as close as possible to
+                    // the screen width while still keeping the same size for each day cells.
+                    @include media-breakpoint-down(md) {
+                        // Pivot widths (number of pixels divisible per 7 as calendar = 7 cells wide)
+                        // Between min width of 21 rem (336 px) and md screen width.
+                        $widths: 364, 392, 448, 504, 560, 616, 672;
+                        @each $w in $widths {
+                            @media (min-width: #{$w + 28}px) { // Keep some margins and space for scrollbar
```
Odoonan: Actually, I think rem is better from an accessibility perspective (even though I did not manage to make this fail). Would go to what you add, with 28, 35, 42 and + 2rem and that's it ^^ Would that work? It won't be perfect, but it should still fix the issue pretty much, no?

addons/calendar/views/calendar_views.xml
Odoonan: Looks good but when going to line, the ms-1 creates a weird jump. It can probably be scrapped, as gap-2 does the work :)

addons/web/static/src/views/calendar/calendar_renderer.scss
amdi-odoo: If we only use 28, 35 and 42 we still have huge margins like so and the months don't seem to take the whole width:

image image

It's still possible to gain space and make it easier for the user to click on the days so I'd keep the .5 widths. For the 22.75 is not mandatory but I think it's recommended as it'd fit a lot more accurately a larger range of phone screens

addons/web/static/src/views/calendar/calendar_renderer.scss
```diff
@@ -571,6 +571,19 @@
...
+                    @include media-breakpoint-down(md) {
+                        // Pivot widths (number of pixels divisible per 7 as calendar = 7 cells wide)
+                        // Between min width of 21 rem and md screen width.
+                        $widths: 22.75, 24.5, 28, 31.5, 35, 38.5, 42;
+                        @each $w in $widths {
+                            @media (min-width: #{$w + 2}rem) { // Keep some margins and space for scrollbar
+                                width: #{$w}rem;
+                            }
+                        }
+                    }
```
rfr-odoo, line 578: Why not using a rule like this:

```suggestion
                    @include media-breakpoint-down(md) {
                        width: calc(100% - #{map-get($spacers, 4)});
                    }
```
I'm maybe missing something 🤔 ?

adr-odoo: Maybe it would be because we might have a 1px difference between the columns?
I think we should keep it simple anyway. The browser should be able to handle this case.

amdi-odoo: Hello @rfr-odoo :slightly_smiling_face: 
I'm not entirely sure why but when I reduce the window size and reload the page, with your rule, in some cases, the columns don't align between the different `fc-month`:
image

With your rule all the `fc-month` seems to have the same width but the day cells widths are different.
Example:
For February, the 28th is 100.47 pixels wide, the 27th is 100.42 pixels.
For March, the 28th is 98.31 and the 27th 98.28.
As it's per day cells, the difference is adding up.

=> As it was set to a fixed width of 21rem, I tried to keep a fixed width for each month as well.
Maybe it has something to do with the fact that the month table layout is set to `fixed`?

adr-odoo: @amdi-odoo Hello, I tried to reproduce this on Chrome and Firefox and it seems to work well.
We tested on https://github.com/odoo-dev/odoo/compare/master-imp-cal-mobile-rfr
Do you still have an issue on this branch?

amdi-odoo: Weirdly no I don't seem to have issues :slightly_smiling_face: 
I'll use your suggestion then ! Thanks for having a look :smile: !

amdi-odoo: @adr-odoo Hello :slightly_smiling_face: 
Actually no sorry I still have the issue :sweat_smile: 
It appears when you load the page (not on resize)
image
I'm on Chrome. The February and March month columns are still not aligned.

amdi-odoo: Hello again @rfr-odoo @adr-odoo :slightly_smiling_face: 
This seems to be working (not `%` but `vw`). What do you think about it ? :
```scss
@include media-breakpoint-down(md) {
    width: calc(100vw - #{map-get($spacers, 5)});
}
```
Thanks!

rfr-odoo: Yes it's ok for me 👍

rfr-odoo (approved):

PR #246757, opened 2026-02-02
[FIX] hr_holidays,hr_holidays_attendance: use employee timezone for attendance-based accruals

ramezlahzy:
…ttendance-based accruals

Steps to reproduce:
- Create an accrual plan with frequency "hourly" and source "attendance"
- Create an employee with work entry source = attendance
- Set employee timezone to one with a big difference from UTC (e.g., Asia/Tokyo)
- Create an allocation with a time off type based on hours and validate it
- Create attendance records for this employee
- Check the employee's allocation on their dashboard

The allocated time is not accrued on the correct day based on the employee's local timezone. For example, an attendance from April 1, 22:00 UTC to April 2, 07:00 UTC is actually April 2, 07:00-16:00 in Tokyo, but the system incorrectly splits it across two UTC days.

Bug cause:
The accrual system used UTC dates for all calculations:
1. `_update_accrual()` only found allocations where nextcall <= today in UTC
2. `_process_accrual_plans()` used UTC date as the processing boundary
3. `_get_accrual_plan_level_work_entry_prorata()` interpreted date boundaries in UTC when searching attendance records

This caused incorrect hour attribution for employees in non-UTC timezones.

Solution:
- Add `_get_date_in_employee_timezone()` helper method to get today's date in the employee's configured timezone
- Modify `_process_accrual_plans()` to use employee timezone date for attendance-based allocations
- Extend `_update_accrual()` to also find attendance-based allocations that are ready based on employee timezone but not UTC
- Update `_get_accrual_plan_level_work_entry_prorata()` to convert date boundaries to employee timezone before searching attendance records

Task Id : 5890857

### DIFF
--- a/addons/hr_holidays/models/hr_leave_allocation.py
+++ b/addons/hr_holidays/models/hr_leave_allocation.py
@@ -6,6 +6,7 @@
 from datetime import datetime, date, time
 from dateutil.relativedelta import relativedelta
 from calendar import monthrange
+from pytz import timezone
 
 from odoo import api, fields, models, _
 from odoo.addons.resource.models.utils import HOURS_PER_DAY
@@ -477,7 +478,6 @@ class HolidaysAllocation(models.Model):
             leaves_taken = employee_days_per_allocation[origin.employee_id][origin.holiday_status_id][origin]['leaves_taken']
             return leaves_taken
 
-        date_to = date_to or fields.Date.today()
         already_accrued = {allocation.id: allocation.already_accrued or (allocation.number_of_days != 0 and allocation.accrual_plan_id.accrued_gain_time == 'start') for allocation in self}
         first_allocation = _("""This allocation have already ran once, any modification won't be effective to the days allocated to the employee. If you need to change the configuration of the allocation, delete and create a new one.""")
         for allocation in self:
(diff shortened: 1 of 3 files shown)


addons/hr_holidays_attendance/models/hr_leave_allocation.py
```diff
@@ -89,9 +90,11 @@ def _get_accrual_plan_level_work_entry_prorata(self, level, start_period, start_
         self.ensure_one()
         if level.frequency != 'hourly' or level.frequency_hourly_source != 'attendance':
             return super()._get_accrual_plan_level_work_entry_prorata(level, start_period, start_date, end_period, end_date)
+
+        employee_tz = timezone(self.employee_id.tz or 'UTC')
         datetime_min_time = datetime.min.time()
-        start_dt = datetime.combine(start_date, datetime_min_time)
-        end_dt = datetime.combine(end_date, datetime_min_time)
+        start_dt = employee_tz.localize(datetime.combine(start_date, datetime_min_time)).astimezone(UTC).replace(tzinfo=None)
+        end_dt = employee_tz.localize(datetime.combine(end_date, datetime_min_time)).astimezone(UTC).replace(tzinfo=None)
```
ziriraha, line 97:
```suggestion
        start_dt = datetime.combine(start_date, datetime_min_time, employee_tz).astimezone(UTC).replace(tzinfo=None)
        end_dt = datetime.combine(end_date, datetime_min_time, employee_tz).astimezone(UTC).replace(tzinfo=None)
```
nitpick, but just fyi, you can do this like this

addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -486,6 +486,11 @@ def _get_leaves_taken(allocation):
             level_ids = allocation.accrual_plan_id.level_ids.sorted('sequence')
             if not level_ids:
                 continue
+            if not date_to:
+                employee_tz = timezone(allocation.employee_id.tz or 'UTC')
+                employee_date_to = datetime.now(employee_tz).date()
+            else:
+                employee_date_to = date_to
```
ziriraha, line 493:
```suggestion
                date_to = datetime.now(employee_tz).date()
```
to keep the diff leaner i think its better to not rename the variable

ziriraha (changes requested):
thanks for your work, one small comment, apart from that LGTM

Also give it a little rebase

PR #234974, opened 2025-11-07
[IMP] *: Improve UX of bank account

hsal-odoo:
This commit improves the UX of bank accounts by:
1. removing res.bank model
2. removes localizations redundancy
3. some fields in localizations should be available to all companies
4. most countries had missing fields

task-5145618

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246758, opened 2026-02-02
[IMP] l10n_latam_check: Add bank field on checks

sham-odoo:
This commit adds partner bank field for Latam Checks.

After a recent improvement in saas~19.2, #234974 the `res.bank` was moved to `res.partner.bank`, resulting in removal of existing `bank_id` field from latam checks.
Hence this commit introduces a new field `bank_account_id` aiming to fullfill the same purpose.

task-5893461

---

### DIFF
--- a/addons/l10n_latam_check/models/account_payment.py
+++ b/addons/l10n_latam_check/models/account_payment.py
@@ -142,6 +142,7 @@ class AccountPayment(models.Model):
         'payment_method_line_id', 'state', 'date', 'amount', 'currency_id', 'company_id',
         'l10n_latam_move_check_ids.issuer_vat', 'l10n_latam_move_check_ids.payment_id.date',
         'l10n_latam_new_check_ids.amount', 'l10n_latam_new_check_ids.name',
+        'l10n_latam_new_check_ids.bank_account_id', 'l10n_latam_move_check_ids.bank_account_id',
     )
     def _compute_l10n_latam_check_warning_msg(self):
         """
@@ -156,11 +157,15 @@ class AccountPayment(models.Model):
             if rec.payment_method_code == 'new_third_party_checks':
                 same_checks = self.env['l10n_latam.check']
                 for check in rec.l10n_latam_new_check_ids.filtered(
-                        lambda x: x.name and x.payment_method_line_id.code == 'new_third_party_checks' and x.issuer_vat):
+                    lambda x: x.name
+                    and x.payment_method_line_id.code == "new_third_party_checks"
+                    and x.issuer_vat
+                ):
                     same_checks += same_checks.search([
                         ('company_id', '=', rec.company_id.id),
                         ('issuer_vat', '=', check.issuer_vat),
                         ('name', '=', check.name),
+                        ('bank_account_id', '=', check.bank_account_id.id),
                         ('payment_id.state', 'not in', ['draft', 'canceled']),
                         ('id', '!=', check._origin.id)], limit=1)
                 if same_checks:
--- a/addons/l10n_latam_check/models/l10n_latam_check.py
+++ b/addons/l10n_latam_check/models/l10n_latam_check.py
@@ -56,6 +56,7 @@ class L10n_LatamCheck(models.Model):
         related='payment_id.payment_method_line_id',
         store=True,
     )
+    bank_account_id = fields.Many2one('res.partner.bank', compute='_compute_bank_account_id', store=True, readonly=False)
 
     # issue_state is used to know that is an own check and also that is posted
     _unique = models.UniqueIndex("(name, payment_method_line_id) WHERE outstanding_line_id IS NOT NULL")
@@ -108,6 +109,13 @@ class L10n_LatamCheck(models.Model):
                     rec.issue_state = 'debited'
             else:
                 rec.issue_state = 'handed'
+
+    @api.depends('payment_method_line_id.code', 'partner_id')
+    def _compute_bank_account_id(self):
+        for rec in self:
+            rec.bank_account_id = False
+            if rec.payment_method_line_id.code == 'new_third_party_checks':
+                rec.bank_account_id = rec.partner_id.bank_ids[:1].id
 
     def action_void(self):
         for rec in self.filtered('outstanding_line_id'):
--- a/addons/l10n_latam_check/views/account_payment_view.xml
+++ b/addons/l10n_latam_check/views/account_payment_view.xml
@@ -14,6 +14,7 @@
                                     <field name="company_id" column_invisible="True"/>
                                     <field name="currency_id" column_invisible="True"/>
                                     <field name="name" />
+                                    <field name="bank_account_id" column_invisible="parent.payment_method_code == 'own_checks'" options="{'no_create': True}"/>
                                     <field name="issuer_vat" column_invisible="parent.payment_method_code == 'own_checks'"/>
                                     <field name="payment_date"/>
                                     <field name="amount" />
@@ -30,6 +31,7 @@
                                     <field name="company_id" column_invisible="True"/>
                                     <field name="currency_id" column_invisible="True"/>
                                     <field name="name" />
+                                    <field name="bank_account_id" optional="hide"/>
                                     <field name="issuer_vat" optional="hide"/>
                                     <field name="payment_date" optional="hide"/>
                                     <field name="amount"/>
--- a/addons/l10n_latam_check/views/l10n_latam_check_view.xml
+++ b/addons/l10n_latam_check/views/l10n_latam_check_view.xml
@@ -47,6 +47,7 @@
                     domain="[('current_journal_id.inbound_payment_method_line_ids.payment_method_id.code', '=', 'in_third_party_checks')]"/>
             </filter>
             <field name="original_journal_id" position="before">
+                <field name="bank_account_id"/>
                 <field name="issuer_vat"/>
                 <field name="current_journal_id"/>
             </field>
@@ -133,6 +134,7 @@
                         </group>
                         <group>
                             <field name="amount"/>
+                            <field name="bank_account_id" invisible="issue_state"/>
                             <field name="issuer_vat"  invisible="issue_state"/>
                             <field name="currency_id" invisible="1"/>
                             <field name="company_id" groups="base.group_multi_company"/>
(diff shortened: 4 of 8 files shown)


sham-odoo:
hello @jorenvo 
done. can you please have a look?
thanks!

addons/l10n_latam_check/models/l10n_latam_check.py
```diff
@@ -56,6 +56,7 @@ class L10n_LatamCheck(models.Model):
         related='payment_id.payment_method_line_id',
         store=True,
     )
+    bank_id = fields.Many2one('res.partner.bank', compute='_compute_bank_id', store=True, readonly=False)
```
jorenvo:
```suggestion
    bank_account_id = fields.Many2one('res.partner.bank', compute='_compute_bank_id', store=True, readonly=False)
```

You appear to bring back the old `bank_id` field and removed the upgrade script. But it's not the same, that was to `res.bank` which no longer exists.

I don't think we should change anything in upgrade. The original `bank_id` field is gone forever, and there's no way to populate `bank_account_id` on old checks in an upgrade script, no way for us to know what account they're linked to.

addons/l10n_latam_check/models/l10n_latam_check.py
```diff
@@ -109,6 +110,14 @@ def _compute_issue_state(self):
             else:
                 rec.issue_state = 'handed'
 
+    @api.depends('payment_method_line_id.code', 'partner_id')
+    def _compute_bank_id(self):
+        for rec in self:
+            if rec.payment_method_line_id.code != 'new_third_party_checks':
+                rec.bank_id = False
+                continue
+            rec.bank_id = rec.partner_id.bank_ids[:1].id
```
jorenvo: Why is there a `rec.bank_id = False` case? We should be able to figure out a good default for both our own checks and checks we receive from a customer.

addons/l10n_latam_check/views/account_payment_view.xml
```diff
@@ -14,6 +14,7 @@
                                     <field name="company_id" column_invisible="True"/>
                                     <field name="currency_id" column_invisible="True"/>
                                     <field name="name" />
+                                    <field name="bank_id" column_invisible="parent.payment_method_code == 'own_checks'"/>
```
jorenvo: @vbe-odoo you don't want to track what bank account our own checks belong to? Why not?

addons/l10n_latam_check/views/l10n_latam_check_view.xml
```diff
@@ -133,6 +134,7 @@
                         </group>
                         <group>
                             <field name="amount"/>
+                            <field name="bank_id" invisible="issue_state"/>
```
jorenvo: Why hide not show it if it has an `issue_state`? I'm not saying it's wrong necessarily, I just don't understand it.

addons/l10n_latam_check/wizards/l10n_latam_payment_register_check.py
```diff
@@ -22,6 +22,7 @@ class L10n_LatamPaymentRegisterCheck(models.TransientModel):
     )
     payment_date = fields.Date(readonly=False, required=True)
     amount = fields.Monetary()
+    bank_id = fields.Many2one('res.partner.bank', compute='_compute_bank_id', store=True, readonly=False)
```
jorenvo:
```suggestion
    bank_account_id = fields.Many2one('res.partner.bank', compute='_compute_bank_id', store=True, readonly=False)
```

jorenvo (commented):
Hello @sham-odoo, thanks for working on this!

It's confusing to me to refer to this as "restoring" the bank field. The old `res.bank` m2o is gone, and this is unrelated. It creates an entirely new `res.partner.bank` m2o on the check model.

addons/l10n_latam_check/views/account_payment_view.xml
vbe-odoo: @jorenvo, because own checks are created from a Bank type journal which already represents a bank.
Example: Own Check from Chase will always be from bank "Chase"; own check from "Wells Fargo" journal, will always be from Wells Fargo.

Moreover, there cannot be duplicates on check numbering when coming from the same bank.

On the other hand, on third-party checks, the journal use is fof type "Cash", as it represents cash to be collected in the future. In this case, a third-party check when created on a payment can be from different banks (1 from Wells Fargo, 1 from Chase, etc) but created using the same journal. In that case, is when we check if there's another check that already exists with the same number, bank, and issuer VAT (in own checks, the issuer vat is always the company VAT, but that's not the case for third-party checks).

I hope this helps!

jorenvo: Good point, thanks for explaining!

addons/l10n_latam_check/models/l10n_latam_check.py
sham-odoo: this question is addressed by VBE in [this](https://github.com/odoo/odoo/pull/246758#discussion_r2765820631) comment.
marking as resolved since that one also resolved.

addons/l10n_latam_check/models/l10n_latam_check.py
sham-odoo: I got your point. I'll make necessary changes.

addons/l10n_latam_check/wizards/l10n_latam_payment_register_check.py
```diff
@@ -41,3 +42,11 @@ def _clean_issuer_vat(self):
...
             if hasattr(stdnum_vat, 'compact'):
                 rec.issuer_vat = stdnum_vat.compact(rec.issuer_vat)
+
+    @api.depends('payment_register_id.payment_method_line_id.code', 'payment_register_id.partner_id')
+    def _compute_bank_id(self):
+        for rec in self:
+            if rec.payment_register_id.payment_method_line_id.code != 'new_third_party_checks':
+                rec.bank_account_id = False
+                continue
+            rec.bank_account_id = rec.payment_register_id.partner_id.bank_ids[:1].id
```
jorenvo:
```suggestion
            rec.bank_account_id = False
            if rec.payment_register_id.payment_method_line_id.code == 'new_third_party_checks':
                rec.bank_account_id = rec.payment_register_id.partner_id.bank_ids[:1].id
```

addons/l10n_latam_check/models/l10n_latam_check.py
```diff
@@ -109,6 +110,14 @@ def _compute_issue_state(self):
             else:
                 rec.issue_state = 'handed'
 
+    @api.depends('payment_method_line_id.code', 'partner_id')
+    def _compute_bank_account_id(self):
+        for rec in self:
+            if rec.payment_method_line_id.code != 'new_third_party_checks':
+                rec.bank_account_id = False
+                continue
+            rec.bank_account_id = rec.partner_id.bank_ids[:1].id
```
jorenvo: Same as below.

addons/l10n_latam_check/models/account_payment.py
```diff
@@ -218,11 +218,16 @@ def _compute_l10n_latam_check_warning_msg(self):
             if rec.payment_method_code == 'new_third_party_checks':
                 same_checks = self.env['l10n_latam.check']
                 for check in rec.l10n_latam_new_check_ids.filtered(
-                        lambda x: x.name and x.payment_method_line_id.code == 'new_third_party_checks' and x.issuer_vat):
+                    lambda x: x.name
+                    and x.payment_method_line_id.code == "new_third_party_checks"
+                    and x.bank_account_id
```
jorenvo:
```suggestion
```

If two checks have the same name and no bank account I think we can still flag them as a potential duplicate.

sham-odoo: I have a question here tho. should we flag them as a potential duplicate? because we don't know whether they are from the same bank.

hello @vbe-odoo what are your thoughts on this?

vbe-odoo: @sham-odoo @jorenvo I think so... we could flag even if no bank is added, at least that will already warn the user and lead them to check or add a bank to differentiate.

Let's add Joren's suggestion.

Thanks!

addons/l10n_latam_check/wizards/l10n_latam_payment_register_check.py
sham-odoo: hello @jorenvo 
this assigns value twice to `bank_account_id`, `False` everytime and the `bank` if condition satisfies.
can you please help me understand this as a better approach ?

addons/l10n_latam_check/models/account_payment.py
```diff
@@ -203,7 +203,7 @@ def _l10n_latam_check_unlink_split_move(self):
     @api.depends(
         'payment_method_line_id', 'state', 'date', 'amount', 'currency_id', 'company_id',
         'l10n_latam_move_check_ids.issuer_vat', 'l10n_latam_move_check_ids.payment_id.date',
-        'l10n_latam_new_check_ids.amount', 'l10n_latam_new_check_ids.name',
+        'l10n_latam_new_check_ids.amount', 'l10n_latam_new_check_ids.name', 'l10n_latam_move_check_ids.bank_account_id',
```
jorenvo:
```suggestion
        'l10n_latam_new_check_ids.amount', 'l10n_latam_new_check_ids.name',
        'l10n_latam_new_check_ids.bank_account_id', 'l10n_latam_move_check_ids.bank_account_id',
```
You use it `same_checks.search(...)`.

addons/l10n_latam_check/wizards/l10n_latam_payment_register_check.py
```diff
@@ -41,3 +42,10 @@ def _clean_issuer_vat(self):
             stdnum_vat = stdnum.util.get_cc_module(rec.company_id.country_id.code, 'vat')
             if hasattr(stdnum_vat, 'compact'):
                 rec.issuer_vat = stdnum_vat.compact(rec.issuer_vat)
+
+    @api.depends('payment_register_id.payment_method_line_id.code', 'payment_register_id.partner_id')
+    def _compute_bank_id(self):
```
jorenvo:
```suggestion
    def _compute_bank_account_id(self):
```

jorenvo (commented):
Hey @sham-odoo, just two small things remaining.

jorenvo (approved):
Thanks, @sham-odoo!

@robodoo r+

jorenvo (approved):
@robodoo r+

smetl (approved):
@robodoo r+

jorenvo (approved):
@robodoo r+

PR #246759, opened 2026-02-02
[FIX] l10n_fr_account: expense account default tax

legi-odoo:
**PROBLEM**
In the fr chart of account, default taxes for expense account are sale taxes while they should be purchase taxes.

opw-5891867


alialfie (approved):
PO approves, thank you!
robodoo r+

PR #246760, opened 2026-02-02
[IMP] point_of_sale: prevent archiving pos.config with active session

rhe-odoo:
Add 'active' to _get_forbidden_change_fields to block archiving a Point of Sale configuration while a session is still open.

---


davidmonnom (commented):
Good one, but missing tests :)

davidmonnom (approved):
@robodoo r+

PR #246762, opened 2026-02-02
[FIX] mrp{,_subcontracting}: use interpolated prefix/suffix for generating serial batch

qucol-odoo:
## Issue
When generating multiple serial numbers in a single manufacturing order, interpolations are not replaced. For example, if the sequence prefix is `%(month)s`, the placeholder is not be replaced by the current month, and the generated serial numbers contain `%(month)s` literally.

This issue also impacts the `mrp_subcontracting` serial numbers generation by causing a similar issue as the second bug described by https://github.com/odoo/odoo/commit/20158a1. When creating a duplicate of a subcontracted manufacturing order, the *First SN* field (`lot_name`) is initiated to `1` again, instead of starting from the last SN of the previous MO. This issue was supposedly fixed by https://github.com/odoo/odoo/commit/9cd0c4a, but the issue still occurs when using interpolated prefix.

## Steps to reproduce
1. Install *Manufacturing* (`mrp`)
2. In Settings, enable *Lots & Serial Numbers*
3. In Settings > Technical > Sequences, give a prefix using interpolation to the `stock.lot.serial` sequence. An example of such prefix could be `%(year)s-%(month)s-%(day)s-`
3. Create a product P
    - Tracked *By Unique Serial Number*
    - Create a BoM for the product P
4. Create and confirm a Manufacturing Order for product P
    - Quantity >= 2
5. Click *Generate Serial*
6. **The _First SN_ field uses the interpolation without replacing the placeholders. When clicking _Generate_, more serials are generated where the placeholders are not replaced. If we proceed, the produced serial numbers still contain the placeholders.**

If we generate a MO with a quantity of one, the serial number is correctly generated (the placeholders are replaced as expected). 

## Cause

The method responsible for replacing the interpolation placeholder is `IrSequence._get_prefix_suffix` which is called by `IrSequence.get_next_char`:

https://github.com/odoo/odoo/blob/bf7dee8069f203095c44381708153865d3e8f19e/odoo/addons/base/models/ir_sequence.py#L240-L242

Generating a single serial numbers work as expected because it calls `MrpProduction.action_generate_serial`, which uses `IrSequence.get_next_char` further down the call stack.

When generating **multiple** serial numbers in a single MO, the [`mrp.production.serials` wizard](https://github.com/odoo/odoo/blob/19.0/addons/mrp/wizard/mrp_production_serial_numbers.py) is used. In this flow, `get_next_char` is never called, because the serial number logic is reimplemented using fields (`serial_prefix_format` and `next_serial`) from the `product.product` model.

## Fix
Serial numbers are generated from the *First SN* (`lot_name`) field. Correctly applying the interpolation to that field ensures that all subsequently generated serial numbers also use the interpolated values. Because `mrp_subcontracting` inherits the behavior from `mrp` to generate the lot_name, fixing this issue in `mrp` also solves it in `mrp_subcontracting`.

When initiating the `lot_name` field, we need to take in account cases where the product does not have a `lot_sequence_id`. This can happen if the sequence of the product was deleted. The same check is not required when updating the `number_next_actual`, as if there's no `lot_sequence_id`, there's no `number_next_actual` to update.

The logic updating the `number_next_actual` has also been adjusted to ensure the field is incremented correctly. This was an issue corrected by https://github.com/odoo/odoo/commit/20158a115ef3dfb9fe2dd5103d3bed7e87e37940 and https://github.com/odoo/odoo/commit/3ff0f2e3ef4fbcbf0b0b0ae6db8ff6f52bfdf67a, but without taking interpolated prefixes into account.

opw-5882578

### DIFF
--- a/addons/mrp/wizard/mrp_production_serial_numbers.py
+++ b/addons/mrp/wizard/mrp_production_serial_numbers.py
@@ -19,12 +19,14 @@ class MrpProductionSerials(models.TransientModel):
     @api.depends('production_id')
     def _compute_lot_name(self):
         for wizard in self:
-            wizard.serial_numbers = '\n'.join(self.production_id.lot_producing_ids.mapped('name'))
+            wizard.serial_numbers = '\n'.join(wizard.production_id.lot_producing_ids.mapped('name'))
             if wizard.lot_name:
                 continue
-            wizard.lot_name = self.production_id.lot_producing_ids[:1].name
+            wizard.lot_name = wizard.production_id.lot_producing_ids[:1].name
             if not wizard.lot_name:
-                wizard.lot_name = self.production_id.product_id.serial_prefix_format + self.production_id.product_id.next_serial
+                sequence = wizard.production_id.product_id.lot_sequence_id
+                wizard.lot_name = sequence.get_next_char(sequence.number_next_actual) if sequence \
+                                 else wizard.production_id.product_id.serial_prefix_format + wizard.production_id.product_id.next_serial
 
     @api.depends('production_id')
     def _compute_lot_quantity(self):
@@ -56,13 +58,12 @@ class MrpProductionSerials(models.TransientModel):
         ])
         existing_lot_names = existing_lots.mapped('name')
         new_lots = []
+        sequence = self.production_id.product_id.lot_sequence_id
         for lot_name in sorted(lots):
             if lot_name in existing_lot_names:
                 continue
-
-            if lot_name == self.production_id.product_id.serial_prefix_format + self.production_id.product_id.next_serial:
-                if self.production_id.product_id.lot_sequence_id:
-                    self.production_id.product_id.lot_sequence_id.number_next_actual += 1
+            if sequence and lot_name == sequence.get_next_char(sequence.number_next_actual):
+                sequence.number_next_actual += 1
             new_lots.append({
                 'name': lot_name,
                 'product_id': self.production_id.product_id.id
--- a/addons/mrp_subcontracting/tests/test_subcontracting.py
+++ b/addons/mrp_subcontracting/tests/test_subcontracting.py
@@ -1469,6 +1469,26 @@ class TestSubcontractingSerialMassReceipt(TransactionCase):
             ]
         })
 
+    def generate_subcontracting_receipt_and_mo(self, product_qty, warehouse=None):
+        if not warehouse:
+            warehouse = self.env['stock.warehouse'].search([('company_id', '=', self.env.company.id)], limit=1)
+        receipt = self.env['stock.picking'].create({
+            'picking_type_id': warehouse.in_type_id.id,
+            'partner_id': self.subcontractor.id,
+            'location_id': self.ref('stock.stock_location_suppliers'),
+            'location_dest_id': warehouse.lot_stock_id.id,
+            'move_ids': [Command.create({
+                'product_id': self.finished.id,
+                'product_uom_qty': product_qty,
+                'product_uom': self.finished.uom_id.id,
+                'location_id': self.ref('stock.stock_location_suppliers'),
+                'location_dest_id': warehouse.lot_stock_id.id,
+            })]
+        })
+        receipt.action_confirm()
+        action = receipt.move_ids.action_show_subcontract_details()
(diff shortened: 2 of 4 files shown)


addons/mrp/wizard/mrp_production_serial_numbers.py
tpma-odoo, line 66: Hello @qucol-odoo ,

I’m also working on a similar [issue](https://www.odoo.com/odoo/project/966/tasks/5258667), and while checking this, I noticed that the `mrp_subcontracting` module overrides the `action_apply` method for subcontracted manufacturing orders. It seems that you may have missed that the `same change` needs to be applied there as well to keep the behavior consistent. Otherwise, the sequence will `not be incremented` when serials are generated for `subcontracted MO`.

You can check the [Video](https://www.awesomescreenshot.com/video/49024549?key=44019ccf4ebcf7d4d4e7eca2e1bda0cc) for the exact steps to reproduce the issue on subcontracted MOs.
[Runbot reference](https://runbot.odoo.com/runbot/batch/2347811)

Thanks! 😊

qucol-odoo: Hey @tpma-odoo,
Thank you for the reminder, I completely forgot about that! I will add the same changes in mrp_subcontracting 🙂

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1698,7 +1700,33 @@ def generate_subcontracting_receipt_and_mo(product_qty):
...
         ])
 
-        _third_receipt, third_mo = generate_subcontracting_receipt_and_mo(2)
+        _third_receipt, third_mo = self.generate_subcontracting_receipt_and_mo(2, warehouse)
         third_mo.action_confirm()
         third_serials_wizard = Form.from_action(self.env, third_mo.action_generate_serial())
         self.assertEqual(third_serials_wizard.lot_name, 'TEST0000006')
+
+    @freeze_time('2024-02-03')
+    def test_use_interpolated_prefix_in_subcontracting_productions(self):
```
lase-odoo, line 1709: Just to be picky, add a description to this test ^^

addons/mrp_subcontracting/tests/test_subcontracting.py
```diff
@@ -1698,7 +1700,33 @@ def generate_subcontracting_receipt_and_mo(product_qty):
...
+        self.finished.lot_sequence_id.number_next_actual = 1
+        serials_wizard = Form.from_action(self.env, mo.action_generate_serial())
+        self.assertEqual(serials_wizard.lot_name, '03-02-0000001')
+        serials_wizard.save().action_generate_serial_numbers()
+        serials_wizard.save().action_apply()
+        self.assertRecordValues(mo._get_subcontract_move().lot_ids.sorted('name'), [
+            {'name': '03-02-0000001'},
+            {'name': '03-02-0000002'},
+        ])
+        self.assertRecordValues(receipt.move_ids[0].lot_ids.sorted('name'), [
```
lase-odoo: There is a single move if I am not mistaken.
```suggestion
        self.assertRecordValues(receipt.move_ids.lot_ids.sorted('name'), [
```

addons/mrp/wizard/mrp_production_serial_numbers.py
lase-odoo: Unrelated to your change but, we are in a loop and there is a huge mistake here... Could you replace all the `self` in the loop by `wizard` before someone calls the compute method on a batch of `mrp.production.serials` with different `production_id.product_id`.

lase-odoo (approved):
Huge thanks for the fix and tests !! 🌊

Not much to say, the explanations are top notch, the fix is clean, the tests are surgical.

I just saw an other issue unrelated to your bug that we should also fix so I took the liberty of adding nitpick comments on the tests ;-)

addons/mrp/wizard/mrp_production_serial_numbers.py
lase-odoo: I am referring to these line:
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/mrp/wizard/mrp_batch_produce.py#L29-L36

PR #246768, opened 2026-02-02
[FIX] l10n_in: correct column shift section pdf

MAGREMENT:
How to reproduce :
- With l10n_in company
- Create Invoice
- Add section
- Add product under section
- Confirm invoice
- Print to pdf

The problem :
The value for the total amount of the section is in the wrong column (the one to the left)

Why :
The l10n_in module adds a column in the report for the HSN/SAC of the products. It correctly did for the products and the table title but not for the sections.

opw-5879828

---

### DIFF
--- a/addons/l10n_in/views/report_invoice.xml
+++ b/addons/l10n_in/views/report_invoice.xml
@@ -37,6 +37,10 @@
             </td>
         </xpath>
 
+        <xpath expr="//t[@name='account_invoice_line_accountable']/following-sibling::*[1]/td[1]" position="after">
+            <td t-if="o.company_id.account_fiscal_country_id.code == 'IN' and o.company_id.l10n_in_is_gst_registered"></td>
+        </xpath>
+
         <xpath expr="//div[@id='qrcode_info']" position="attributes">
             <attribute name="t-if" add="o.company_id.account_fiscal_country_id.code != 'IN'" separator="and"/>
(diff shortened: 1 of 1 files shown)


agr-odoo (approved):
Looks ok

**Before**

Image

**After**

Image

yosuanicolaus (approved):

addons/l10n_in/views/report_invoice.xml
```diff
@@ -37,6 +37,10 @@
             </td>
         </xpath>
 
+        <xpath expr="//t[@name='account_invoice_line_accountable']/following-sibling::*[1]/td[1]" position="after">
```
jco-odoo, line 40: I find that cumbersome that we have to do it that way, but I see no direct other option.  @hamo-odoo

hamo-odoo: Don't you think we should consider adding a name attribute to the sibling instead? (might be rare case) But at least better than being hacky? Not sure if we can add attributes to the xml in stable? (but anyways, the xml structure will not change due to stable policy so we can do something master)

hamo-odoo (approved):
For the stable fix seems okay to me because we don't change xml struture in stable but I would prefer having xml path with id or name something atleast in the master

jco-odoo:
@robodoo r+

PR #246776, opened 2026-02-02
[IMP] web: D&co apply decoration-muted to m2o avatar

PiaCOS:
This commit adds a grayscale filter on the m2o avatar when the field has
the `decoration-muted` attribute.

task-5469063

### DIFF
--- a/addons/web/static/src/views/fields/many2one_avatar/many2one_avatar_field.scss
+++ b/addons/web/static/src/views/fields/many2one_avatar/many2one_avatar_field.scss
@@ -3,6 +3,12 @@
         cursor: context-menu;
         outline: $border-width solid $o-action;
     }
+    &.text-muted {
+        img {
+            filter: grayscale(1);
+            opacity: 0.5;
+        }
+    }
 }
 
 .o_kanban_record  {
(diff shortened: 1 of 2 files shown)


addons/web/static/tests/views/form/form_view.test.js
```diff
@@ -698,6 +698,23 @@ test(`decoration works on widgets`, async () => {
     expect(`.o_field_widget[name="foo"]`).toHaveClass("text-danger");
 });
 
+test(`decoration-muted works on avatars`, async () => {
```
aab-odoo: nice test, but could you move it into many2one_avatar_field_test.js?

addons/web/static/src/views/fields/many2one_avatar/many2one_avatar_field.scss
```diff
@@ -16,6 +27,7 @@
             }
         }
     }
+    @include grayscale-if-muted;
```
aab-odoo: the rule applies on all img in o_kanban_record, not only those of m2o_avatar. I would scope it.

Also, not sure the mixin is really useful (it's only 2 rules, and the `&.text-muted`/`.text-muted &` is a bit confusing

aab-odoo (commented):
much easier 👍

addons/web/static/src/views/fields/many2one_avatar/many2one_avatar_field.scss
PiaCOS: I did a bit more testing and the rule inside `.o_kanban_record` does not seem to be needed

aab-odoo (approved):
Thanks @PiaCOS 

robodoo r+

PR #246777, opened 2026-02-02
[FIX] mail: fix insertion of mixed up many commands

tsm-odoo:
Many fields can be updated in the discuss store in different
modes: `ADD`, `DELETE`, `REPLACE`.

`REPLACE` commands are returned as a value list as it is the
default mode. However, when other commands are added, the
resulting value doesn't make sense as commands and raw values
are mixed up.

This commit fixes the `_add_to_store` method to properly format
the command list:
- When there is only a replace, the return value is kept as is
(i.e. a list of values).
- When there are other commands, the `REPLACE` command is wrapped
into a command.
- As a bonus, when a `REPLACE` command comes after other commands,
the list is cleared. It doesn't make sense to keep other commands
as they will be cleared by the `REPLACE` anyway.

This PR also normalize commands on the client side to ease their
processing: we don't need to guess the shape of the command anymore.

### DIFF
--- a/addons/mail/static/src/model/misc.js
+++ b/addons/mail/static/src/model/misc.js
@@ -24,9 +24,44 @@ export function OR(...args) {
     return [OR_SYM, ...args];
 }
 
-export function isCommand(data) {
-    return ["ADD", "DELETE", "ADD.noinv", "DELETE.noinv"].includes(data?.[0]?.[0]);
+export function isCommandList(data) {
+    return Array.isArray(data) && data.length > 0 && data.every((cmd) => isCommand(cmd));
 }
+
+export function isCommand(data) {
+    return ["ADD", "DELETE", "ADD.noinv", "DELETE.noinv", "REPLACE"].includes(data?.[0]);
+}
+
+/**
+ * Normalize a list of many commands.
+ *
+ * @param {Array|Object|null|false|undefined} command Many command. Can
+ * either be:
+ * - A falsy value or an empty array: interpreted as a "clear" command.
+ * - An object: interpreted as a replace.
+ * - A command.
+ * - An array of commands.
+ * - An array of raw values: interpreted as a replace.
+ * @returns {Array<[string, any[]]>} Normalized list of `[mode, value]` arrays.
+ */
+export function normalizeManyCommands(command) {
+    const ensureArrayValue = (cmd) => [cmd[0], Array.isArray(cmd[1]) ? cmd[1] : [cmd[1]]];
+    if (!command || (Array.isArray(command) && command.length === 0)) {
+        return [["REPLACE", []]];
+    }
+    if (isCommandList(command)) {
+        return command.map(ensureArrayValue);
+    }
+    if (isCommand(command)) {
+        return [ensureArrayValue(command)];
+    }
+    const replaceCmdList = ensureArrayValue(["REPLACE", command]);
+    if (replaceCmdList[1].some((val) => isCommand(val))) {
+        throw new Error("Many commands cannot mix raw values and commands");
+    }
+    return [replaceCmdList];
+}
+
 /**
  * @param {typeof import("./record").Record} Model
  * @param {string} fieldName
--- a/addons/mail/static/src/model/store_internal.js
+++ b/addons/mail/static/src/model/store_internal.js
@@ -4,7 +4,7 @@
 import { htmlEscape, markup, toRaw } from "@odoo/owl";
 import { RecordInternal } from "./record_internal";
 import { deserializeDate, deserializeDateTime } from "@web/core/l10n/dates";
-import { IS_DELETED_SYM, isCommand, isMany } from "./misc";
+import { IS_DELETED_SYM, isCommandList, isMany, normalizeManyCommands } from "./misc";
 import { browser } from "@web/core/browser/browser";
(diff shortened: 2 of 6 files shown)


addons/mail/tests/test_discuss_tools.py
```diff
@@ -250,3 +250,45 @@ def fields(user_res: Store.FieldList, user):
...
+        store.add(
+            general,
+            lambda res: res.many("messages", [], value=general.message_ids, mode="REPLACE"),
+        )
+        data = store.get_result()
+        self.assertEqual(data["discuss.channel"][0]["messages"], general.message_ids.ids)
+        store.add(
+            general,
+            lambda res: (
+                res.many("messages", [], value=general.message_ids, mode="REPLACE"),
```
seb-odoo, line 287: It's strange to re-do a replace on the same store when it was already the previous value.

addons/mail/static/src/model/store_internal.js
```diff
@@ -280,38 +283,32 @@ export class StoreInternal extends RecordInternal {
...
-                        } else if (cmd === "ADD.noinv") {
-                            recordList._.addNoinv(recordList, item);
-                        } else if (cmd === "DELETE.noinv") {
-                            recordList._.deleteNoinv(recordList, item);
-                        } else {
-                            recordList.delete(item);
-                        }
-                    }
+        for (const [cmd, cmdData] of value) {
+            const normalizedData = Array.isArray(cmdData) ? cmdData : [cmdData];
```
seb-odoo: Why is it not done as part of normalize?

seb-odoo: Isn't empty value incorrectly wrapped into array, leading to no longer doing clear?

addons/mail/static/src/model/misc.js
```diff
@@ -27,6 +27,45 @@ export function OR(...args) {
 export function isCommand(data) {
     return ["ADD", "DELETE", "ADD.noinv", "DELETE.noinv"].includes(data?.[0]?.[0]);
 }
+
+/**
+ * Normalize a list of many commands from the server. `REPLACE` commands
+ * are returned as a raw list, whereas `ADD` and `DELETE commands are
```
seb-odoo: In the context of this function: returned -> received

addons/mail/static/src/model/store_internal.js
tsm-odoo: `normalizeManyCommands` omit flasy values. This way, we only have one "clear" command: `[["REPLACE", []]]`

addons/mail/static/src/model/store_internal.js
```diff
@@ -280,38 +283,31 @@ export class StoreInternal extends RecordInternal {
...
-                        } else if (cmd === "DELETE.noinv") {
-                            recordList._.deleteNoinv(recordList, item);
-                        } else {
-                            recordList.delete(item);
-                        }
-                    }
+        for (const [cmd, cmdData] of value) {
+            if (cmd === "REPLACE") {
+                if (!cmdData.length) {
+                    recordList.clear();
```
seb-odoo: Do we even need a case for that? Assign empty list might work the same?

addons/mail/static/src/model/misc.js
```diff
@@ -27,6 +27,46 @@ export function OR(...args) {
...
+    cmdList = Array.isArray(cmdList) ? cmdList : [cmdList];
+    const formattedCmdList = [];
+    let currentReplaceValues;
+    for (const cmdOrValue of cmdList) {
+        if (!Array.isArray(cmdOrValue)) {
+            currentReplaceValues ??= [];
+            if (cmdOrValue) {
+                // Handle falsy values, which must be interpreted as "clear".
+                currentReplaceValues.push(cmdOrValue);
+            }
```
seb-odoo: I don't understand how this handles falsy values, why is it not an explicit empty list in that case?
If there is a pre-existing `currentReplaceValues` is should be cleared.

Also single value should replace everything, not append.

We have to pre-detect if we receive a list of object or a list of commands, mixing both should not be supported and make no sense.

addons/mail/static/src/model/store_internal.js
```diff
@@ -243,21 +243,24 @@ export class StoreInternal extends RecordInternal {
...
                 if (!this.isUpdatingFromStorageEvent) {
                     const lse = record._.fieldsLocalStorage.get(fieldName);
                     if (value === record._.fieldsDefault.get(fieldName)) {
                         lse.remove();
                     } else {
-                        lse.set(value);
+                        lse.set(valueNormalized);
                     }
                 }
             }
```
alexkuhn, line 256: We don't have `localStorage: true` fields that are `fields.Many` yet. I'm pretty sure there are small bugs to fix to make it work.

I'm pretty sure this change is not good, as setting list of commands in local storage field value seems wrong.

Is this a necessary change (= crash or other bug) or more a guess?

seb-odoo: It's not necessary, and indeed Many localstorage is wrong, we were discussing that and it makes no sense, we don't want command in localstorage but the full value of the field after the update is done.

seb-odoo (approved):
@robodoo r+ delegate+

tsm-odoo (commented):
@robodoo r+

PR #83737, opened 2022-02-01
[IMP] mail,portal,test_{mail_full|website}:add language support for portal link

pyduf:
Before this change, the portal link in the email always redirected to a page in the default language (for example while sending a quotation). Now, the portal link redirect to a page in the customer language if it is available.

We add methods in ir_http in portal to easily localize URL so that it can be used in other modules.

Technical notes: when "website" is installed, the portal language is restrained to the website languages. So having a partner with a given language doesn't guarantee that that language is available in the portal. That's why we check the availability of the language before prepending the language to the URL (otherwise it leads to a 404). For that, we call "_get_frontend" with request.is_frontend to True.

Tests have been added in test_mail_full to test the portal behavior alone and in test_website to test the portal behavior with the website installed. This is necessary because installing website restraint the available language of portal to the one of the main website.

In test_mail_full, we patch the method _get_frontend of res_lang to the "http_routing" behavior (i.e.: call _get_frontend with web_force_installed_langs=True context variable). The goal is to ensure that _get_frontend is not limited to the language of the main website when website module is installed and keep the behavior of the portal module without website module regarding the language. Indeed, when installing the module website, some language that were available in the portal become unavailable if not installed on the main website.

Task-2716028

PR #246778, opened 2026-02-02
[FIX] survey: use recipient lang for invitations in mixed-lang batches

ushyme:
When sending survey invitations to a group of recipients with different language preferences, some recipients would receive the invitation in the incorrect language.

### Steps to reproduce

1. Install the "Surveys" module and activate a second language (e.g., Dutch).
2. Create a survey and ensure its invitation template has translations for both languages.
3. Create two contacts: one with English as their language and another with Dutch.
4. On the survey, click "Share" and add both contacts as recipients.
5. Send the invitations.
6. The contact with Dutch preferred language receives the email in English.

### Cause

By default, the wizard uses a single language for every email in a batch. While it can switch this language if everyone in the group speaks the same tongue, it fails to do so for mixed-language groups. Adding compute_lang=True fixes this by telling the system to look up and use the correct language for each recipient one by one.

opw-5868581

### DIFF
--- a/addons/survey/tests/test_survey_invite.py
+++ b/addons/survey/tests/test_survey_invite.py
@@ -8,7 +8,7 @@ from lxml import etree
 from odoo import fields, Command
 from odoo.addons.survey.tests import common
 from odoo.addons.mail.tests.common import MailCommon
-from odoo.exceptions import UserError
+from odoo.exceptions import AccessError, UserError
 from odoo.tests import Form
 from odoo.tests.common import users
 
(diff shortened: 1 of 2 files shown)


pyduf (approved):
Hello @ushyme , 

Thanks for the fix, it works well and the test is clean.

Note that the link in the email starts the survey in English even if the mail is in French but it's not easy to fix I think (you can change it by editing the URL).  A fix in master is ongoing odoo/odoo#83737.

@std-odoo if you have time for this one, thanks.

addons/survey/tests/test_survey_invite.py
```diff
@@ -353,3 +353,57 @@ def test_survey_invite_with_template_attachment(self):
...
+            invite = Invite.create({
+                'survey_id': self.survey.id,
+                'partner_ids': [Command.set((partner_en + partner_nl).ids)],
+                'template_id': template.id,
+            })
+
+            with self.mock_mail_gateway():
+                invite.action_invite()
+
+        self.assertEqual(len(self._new_mails), 2)
```
std-odoo, line 398: can you add a test where `lang` contains forbidden code, different from the value on the template, with a user without the group "mail template editor", and check that an access error is raised?

then the same case, but this time the value is the same on the template and on the survey invite, and check that the rendering worked?

ushyme: BTW @std-odoo, I pushed the changes.

addons/survey/tests/test_survey_invite.py
```diff
@@ -353,3 +353,99 @@ def test_survey_invite_with_template_attachment(self):
...
+                'lang': '{{ object.access_token }}',
+            })
+            with self.assertRaises(AccessError):
+                invite.action_invite()
+
+            invite2 = self.env['survey.invite'].create({
+                'survey_id': self.survey.id,
+                'partner_ids': [Command.set([partner.id])],
+                'template_id': template.id,
+                'subject': 'Custom Subject',
```
std-odoo, line 447: can you check that the lang field is the same as the one on the template?

std-odoo (commented):
One last comment and we are good to go 🙂

addons/survey/tests/test_survey_invite.py
ushyme: Good catch. Done!

std-odoo (approved):
@ushyme Thanks 🙂

@robodoo r+

PR #246780, opened 2026-02-02
[FIX] account_edi_ubl_cii: separate domain and sql constraint domain …

jco-odoo:
…for import

Current behavior before PR:

---

### DIFF
--- a/addons/account_edi_ubl_cii/models/account_edi_common.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_common.py
@@ -500,11 +500,13 @@ class AccountEdiCommon(models.AbstractModel):
         bank_details = list(set(map(sanitize_account_number, bank_details)))
         partner = self.env.company.partner_id if invoice.is_inbound() else invoice.partner_id
         banks_to_create = []
+
         acc_number_partner_bank_dict = {
             bank.sanitized_acc_number: bank
-            for bank in ResPartnerBank.with_context(active_test=False).search(
-                [('partner_id', '=', partner.id), ('acc_number', 'in', bank_details)]
-            )
+            for bank in ResPartnerBank.with_context(active_test=False).search([
+                *self.env['res.partner.bank']._check_company_domain(self.env.company),
+                ('partner_id', '=', partner.id), ('acc_number', 'in', bank_details)
+            ])
         }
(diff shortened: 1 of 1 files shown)


addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -514,10 +516,15 @@ def _import_partner_bank(self, invoice, bank_details):
...
             elif not partner_bank and account_number:
-                banks_to_create.append({
-                    'acc_number': account_number,
-                    'partner_id': partner.id,
-                })
+                bank_accounts = ResPartnerBank.with_context(active_test=False).sudo().search([
+                    ('partner_id', '=', partner.id), ('acc_number', '=', account_number)
+                ])
+
+                if not bank_accounts: # Avoid breaking SQL constraint partner_id, acc_number
```
william-andre, line 523: What if it is found for another company?

jco-odoo: @william-andre Then it does not set it...

jco-odoo: The problem if it would set it is that the invoice will have the res_partner_bank with another account which will make it fail.  Because the res_partner_bank does not have the parent_of domain.

william-andre: Yes, that is why this model is broken 🥳

jco-odoo: Yes, so this is the almost ideal monkey patch.

william-andre: But shouldn't we remove the company_id on the bank?

Since if we have this case, we already have no company set on the partner. Or am I missing something?

jco-odoo: The company_id on the res.partner.bank is a related stored with the partner (but it does not have the same parent_of company domain defined). The problem here is that in this method, a partner can be found with a company_id that is a parent company, but the field on the invoice for the bank account does not allow to set the bank account related to that partner, because here it expects the bank account to have the same company_id as the invoice.

PR #246783, opened 2026-02-02
[IMP] mail: show message edited datetime on mouse-hover

alexkuhn:
Screenshot 2026-02-02 at 15 49 27

Task-4614516

### DIFF
--- a/addons/mail/static/src/core/common/message.xml
+++ b/addons/mail/static/src/core/common/message.xml
@@ -30,7 +30,7 @@
                         </t>
                         <t t-elif="message.isPending" t-call="mail.Message.pendingStatus"/>
                         <t t-elif="!message.is_transient">
-                            <small t-if="isActive and props.showDates" class="o-mail-Message-date o-xsmaller mt-2 text-center lh-1" t-att-title="message.datetimeShort">
+                            <small t-if="isActive and props.showDates" class="o-mail-Message-date o-xsmaller mt-2 text-center lh-1" t-att-title="message.datetimeMedium">
                                 <t t-esc="message.dateSimple"/>
                             </small>
                             <t t-elif="message.starred" t-call="mail.Message.sidebarStarred"/>
@@ -43,7 +43,7 @@
                             </span>
                             <t t-if="!isAlignedRight" t-call="mail.Message.notification"/>
                             <t t-if="isAlignedRight and !message.bubbleColor and !(props.asCard and isMobileOS)" t-call="mail.Message.actions"/>
-                            <small t-if="!message.is_transient" class="o-mail-Message-date o-xsmaller" t-att-title="message.datetimeShort">
+                            <small t-if="!message.is_transient" class="o-mail-Message-date o-xsmaller" t-att-title="message.datetimeMedium">
                                 <t t-if="message.isPending" t-call="mail.Message.pendingStatus"/>
                                 <t t-else="" t-out="message.dateSimpleWithDay"/>
                             </small>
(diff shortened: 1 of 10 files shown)


addons/mail/static/src/core/common/message_model.js
```diff
@@ -284,8 +296,8 @@ export class Message extends Record {
         return this.thread?.effectiveSelf ?? this.store.self;
     }
 
-    get datetimeShort() {
-        return this.datetime.toLocaleString(DateTime.DATETIME_SHORT_WITH_SECONDS);
+    get datetimeMedium() {
+        return this.datetime.toLocaleString({ ...DateTime.DATETIME_MED }, { locale: user.lang });
```
seb-odoo, line 299: Why change the format of this one?

addons/mail/static/src/core/common/message.xml
```diff
@@ -153,7 +153,7 @@
     </t>
 
 <t t-name="mail.Message.edited">
-    <span class="o-xsmaller opacity-50"> (edited)</span>
+    <span class="o-xsmaller fw-bold opacity-50" t-att-title="message.editedDatetimeMedium"> (edited)</span>
```
seb-odoo: Why change style?

addons/mail/static/src/core/common/message_model.js
```diff
@@ -74,6 +74,14 @@ export class Message extends Record {
             );
         },
     });
+    /** @type {boolean} */
```
seb-odoo: wrong type

addons/mail/static/src/core/common/message_model.js
```diff
@@ -274,6 +282,10 @@ export class Message extends Record {
         return this.date || DateTime.now();
     }
 
+    get editedDatetimeMedium() {
+        return this.editedDate.toLocaleString({ ...DateTime.DATETIME_MED }, { locale: user.lang });
```
seb-odoo: editedDate might be empty (what happens to datetime fields when we don't provide a value?)

seb-odoo (commented):
The task says it should be moved too.

addons/mail/static/src/core/common/message.xml
alexkuhn: This is barely visible, but this makes text slightly more readable, and the `fw-bold` is somewhat consistent with clickable / hoverable items like `btn` that have the implicit `fw-bold`.

Copy-paste from the PR of a year ago, I'm not sure this is as beneficial now.

addons/mail/static/src/core/common/message_model.js
alexkuhn: For consistency of the showing of datetime of message on mouse-hover, whether from post date and edited date.

The `DATETIME_MED` is a better format, like the `Feb 2, 2026 3:48 PM`, instead of `2/2/2026 3:48:37 PM` that we have right now, which hasn't been updated since a long time:

- format of datetime should be "Feb 2" instead of "2/2" for clarity
- "WITH_SECONDS" made sense when posting message was quite unreliable and we lacked confidence and UI hint like the spinner and opacity on temporary message, or warning on lack of real-time connection. This is no longer necessary, and short title is preferred. This also matched with competitors.

addons/mail/static/src/core/common/message_model.js
alexkuhn: Good catch, this crashes... I'm fixing it!

addons/mail/static/src/core/common/message.xml
alexkuhn: Looks better without, I've removed it.

alexkuhn:
> The task says it should be moved too.

There's a debate in chatter of task challenging this part of the task: this doesn't work for squashed messages, this increases significantly the size of the header of non-squashed messages, consistency of feature for all message is important concern.

seb-odoo (approved):
@robodoo r+ delegate+

PR #246785, opened 2026-02-02
[FIX] l10n_it_edi: propagate l10n_it on create_company

legi-odoo:
**STEP TO REPRODUCE**

1. Create a ecommerce order on a shop page of a italian company.
2. Goes to the checkout page, enter info (company_name, l10n_it_codice_fiscale, l10n_it_pa_index).
3. On the contact created, click on create company.
4. Notice l10n_it fields are not propagated to the company.

opw-5477372

### DIFF
--- a/addons/l10n_it_edi/models/res_partner.py
+++ b/addons/l10n_it_edi/models/res_partner.py
@@ -202,3 +202,10 @@ class ResPartner(models.Model):
     def _peppol_eas_endpoint_depends(self):
         # extends account_edi_ubl_cii
         return super()._peppol_eas_endpoint_depends() + ['l10n_it_codice_fiscale']
+
+    def create_company(self):
+        res = super().create_company()
+        if res:
+            it_values = self._update_fields_values(('l10n_it_codice_fiscale', 'l10n_it_pa_index'))
+            self.parent_id.update(it_values)
+        return res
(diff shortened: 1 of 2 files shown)


addons/l10n_it_edi/models/res_partner.py
```diff
@@ -202,3 +202,9 @@ def _deduce_country_code(self):
     def _peppol_eas_endpoint_depends(self):
         # extends account_edi_ubl_cii
         return super()._peppol_eas_endpoint_depends() + ['l10n_it_codice_fiscale']
+
+    def create_company(self):
+        if super().create_company():
+            it_values = {}
+            it_values.update(self._update_fields_values(('l10n_it_codice_fiscale', 'l10n_it_pa_index')))
+            self.parent_id.update(it_values)
```
guva-odoo: I think we should return the value from the super, something like that, wdyt?
```suggestion
        res = super().create_company()
        if res:
            it_values = self._update_fields_values(('l10n_it_codice_fiscale', 'l10n_it_pa_index'))
            self.parent_id.update(it_values)
        return res```

guva-odoo (approved):
Thanks! LGTM, ready for next review

epictete (approved):
Thanks for the fix 👌 

@robodoo r+

PR #246786, opened 2026-02-02
[FIX] hr_holidays: correct timezone for time off calendar event

taqu-odoo:
Forward port of https://github.com/odoo/odoo/pull/238941
which was not included by mistake here: https://github.com/odoo/odoo/pull/234703

Additionally, fix an issue where a full day leave was not displayed as a fullday event in the calendar if the leave type unit was half days.


amah-odoo (approved):
Thanks for your work 💯 

@robodoo  r+

PR #246791, opened 2026-02-02
[REM] website: remove unnecessary clean of highlight in snippet preview

blse-odoo:
Commit 21044308329e415621ef63e49709ebed22fe4723 added a cleanup of the highlight of the snippet preview.

This clean up is not necessary because the highlights are cleaned by the interaction on snippet drop.

This commit removes the extra cleanup.


xO-Tx (approved):
@robodoo r+

PR #246799, opened 2026-02-02
[FIX] point_of_sale: use_lna for iot printers

Ysoroko:
This PR https://github.com/odoo/enterprise/pull/102622 activates lna for the iot box requests if "use_lna" is active in the printer settings. However the field is invisible for the iot printers. This PR makes it visible for all printers, including iot to use lna with iot in saas-19.1


loouis-t (approved):
robodoo r+

PR #246807, opened 2026-02-02
[FIX] html_builder: use label of original snippet for custom snippets

blse-odoo:
The label was never set for the custom snippets.

This commit adds a pass in `computeSnippetTemplates` to copy the labels to custom snippets from their corresponding original snippet.

Steps to reproduce:
- Open website builder
- Drop a snippet with the label "Carousel"
- Save this snippet as a custom snippet
- Click on "Custom" category
- Bug: the custom carousel does not have the "Carousel" label

task-5902478

### DIFF
--- a/addons/html_builder/static/src/snippets/snippet_service.js
+++ b/addons/html_builder/static/src/snippets/snippet_service.js
@@ -183,6 +183,7 @@ export class SnippetModel extends Reactive {
                     imagePreviewSrc: snippetEl.dataset.oImagePreview,
                     isCustom: false,
                     label: this.getSnippetLabel(snippetEl),
+                    originalLabel: snippetEl.dataset.oLabel,
                     isDisabled: false,
                     forbidSanitize: false,
                 };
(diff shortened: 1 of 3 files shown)


addons/html_builder/static/src/snippets/snippet_service.js
```diff
@@ -238,6 +239,16 @@ export class SnippetModel extends Reactive {
...
+
+        // Apply to custom snippets the labels from their original snippets
+        for (const snippet of this.snippetsByCategory.snippet_custom) {
+            if (!snippet.label && snippet.name) {
+                const originalSnippet = this.getOriginalSnippet(snippet.name);
+                if (originalSnippet && originalSnippet.originalLabel) {
+                    snippet.label = originalSnippet.originalLabel;
+                }
+            }
+        }
```
FrancoisGe, line 251: Why we don t save this information when we save the custom snippet ?

blse-odoo: We could (we would still need `originalLabel`), and it would work the same for future custom snippet (if we do not change labels on the original ones).
Doing it here brings the following:
- correct label on already existing custom snippet
- update of the label on custom snippets following an update of the label on the original ones

PR #246816, opened 2026-02-02
[FIX] hr_expense,sale_expense: re-invoiced expense quantity

agr-odoo:
Currently, when re-invoicing an expense paid by company, the line added to the sale order will show a price unit equal to the whole expense amount.

Steps to reproduce:
- Create a new Expense Category with Re-Invoice Costs set to 'At cost'
- Create a new Expense, set the new category, and set Paid By to 'Company'
- Set Quantity to greater than 1, and Customer to Re-Invoice to any Sales Order
- Confirm and Submit Journal Entry on the Expense record

Issue:
- On the linked Sales Order you will see the total of the expense is used as the unit price

This occurs because we don't pass the quantity to the move line creation vals, which then default to 1. In turn, when the sale order line is added, the unit price will be based on the move line vals but the quantity will match the expense.

opw-5883290

### DIFF
--- a/addons/sale_expense/models/account_move_line.py
+++ b/addons/sale_expense/models/account_move_line.py
@@ -40,9 +40,12 @@ class AccountMoveLine(models.Model):
             res.update({
                 'name': self.name,
                 'expense_ids': [Command.set(self.expense_id.ids)],
-                'product_uom_qty': self.expense_id.quantity,
                 'analytic_distribution': self.analytic_distribution,
             })
+            if self.expense_id.product_id.expense_policy == 'sales_price' and self.expense_id.product_id.standard_price:
+                res.update({
+                    'product_uom_qty': self.expense_id.quantity,
+                })
         return res
 
     def _sale_create_reinvoice_sale_line(self):
--- a/addons/sale_expense/tests/test_reinvoice.py
+++ b/addons/sale_expense/tests/test_reinvoice.py
@@ -538,3 +538,26 @@ class TestReInvoice(TestExpenseCommon, TestSaleCommon):
                 'is_expense': True,
             },
         ])
+
+    def test_expenses_reinvoice_paid_by_company(self):
+        """ Test expense line paid by company is re-invoiced correctly """
+        expense = self.create_expenses({
+            'name': 'expense_1',
+            'date': '2016-01-01',
+            'product_id': self.company_data['service_order_cost_price'].id,
+            'quantity': 5,  # 235.28 * 5 = 1176.4
+            'employee_id': self.expense_employee.id,
+            'sale_order_id': self.expense_sale_order.id,
+            'payment_mode': 'company_account',
+        })
+
+        expense.action_submit()
+        expense._do_approve()
+        expense.action_post()
+
+        self.assertRecordValues(self.expense_sale_order.order_line, [
+            # Original SO line:
+            {'qty_delivered': 0.0, 'product_uom_qty': 3.0, 'price_unit': 235.0, 'price_total': 705.0, 'price_subtotal': 705.0, 'is_expense': False},
+            # Expense line:
+            {'qty_delivered': 1.0, 'product_uom_qty': 1.0, 'price_unit': 1045.7, 'price_total': 1176.4, 'price_subtotal': 1045.7, 'is_expense': True},
+        ])


addons/sale_expense/tests/test_reinvoice.py
```diff
@@ -538,3 +538,28 @@ def test_expense_reinvoice_tax_multine_line(self):
...
         ])
+
+    def test_expenses_reinvoice_paid_by_company(self):
+        """ Test expense line paid by company is re-invoiced correctly """
+        expense = self.create_expenses({
+            'name': 'expense_1',
+            'date': '2016-01-01',
+            'product_id': self.company_data['service_order_cost_price'].id,
+            'quantity': 5,
+            'price_unit': 50,
```
JulienAlardot: `price_unit` should never be declared directly and that isn't possible in the interface on purpose, remove it.

addons/sale_expense/tests/test_reinvoice.py
```diff
@@ -538,3 +538,28 @@ def test_expense_reinvoice_tax_multine_line(self):
...
+            'total_amount': 250,
+            'employee_id': self.expense_employee.id,
+            'sale_order_id': self.expense_sale_order.id,
+            'payment_mode': 'company_account',
+        })
+
+        expense.action_submit()
+        expense._do_approve()
+        self.post_expenses_with_wizard(expense)
```
JulienAlardot:
```suggestion
        expense.action_post()
```
Company expenses are not using the wizard

addons/sale_expense/tests/test_reinvoice.py
```diff
@@ -538,3 +538,28 @@ def test_expense_reinvoice_tax_multine_line(self):
...
+
+    def test_expenses_reinvoice_paid_by_company(self):
+        """ Test expense line paid by company is re-invoiced correctly """
+        expense = self.create_expenses({
+            'name': 'expense_1',
+            'date': '2016-01-01',
+            'product_id': self.company_data['service_order_cost_price'].id,
+            'quantity': 5,
+            'price_unit': 50,
+            'total_amount': 250,
```
JulienAlardot: `service_order_cost_price` has a `standard_price` so the expense is computed using that price * quantity
Please do not override the total_amount in that case

addons/sale_expense/tests/test_reinvoice.py
```diff
@@ -538,3 +538,28 @@ def test_expense_reinvoice_tax_multine_line(self):
...
+
+        expense.action_submit()
+        expense._do_approve()
+        self.post_expenses_with_wizard(expense)
+
+        self.assertRecordValues(self.expense_sale_order.order_line, [
+            # Original SO line:
+            {'qty_delivered': 0.0, 'product_uom_qty': 3.0, 'price_unit': 235.0, 'is_expense': False},
+            # Expense line:
+            {'qty_delivered': 5.0, 'product_uom_qty': 5.0, 'price_unit': 44.44, 'is_expense': True},
```
JulienAlardot: Should also test the total amounts with taxes included/excluded (and quantity should be 1)

JulienAlardot (changes requested):
Thanks for the fix, but IMO the fix goes the wrong way
on the payment entry we only want 1 as a quantity, period

The sale order quantity should be 1IMO
Else the taxes go tits up

addons/sale_expense/models/account_move_line.py
```diff
@@ -40,7 +40,6 @@ def _sale_prepare_sale_line_values(self, order, price):
             res.update({
                 'name': self.name,
                 'expense_ids': [Command.set(self.expense_id.ids)],
-                'product_uom_qty': self.expense_id.quantity,
```
JulienAlardot, line 43: conditional if product has cost and is in sale_price mode?

JulienAlardot (changes requested):

agr-odoo:
@JulienAlardot 
Sorry, I was not very clear in the last message.
I'm referring to the rounding I proposed to alter with the fixup commit here
Tests in /hr_expense and /sale_expense are passing but I don't know if there could be other side effects in margin analysis

addons/sale_expense/models/account_move_line.py
agr-odoo: Thanks for the note, I updated the PR
Here's the situation now
screenshot_040

agr-odoo: ping @JulienAlardot

agr-odoo: @JulienAlardot

agr-odoo:
@JulienAlardot 
Hello,
When you have the time please let me know if there is anything left to address here, so that we can move this forward. Thanks

JulienAlardot:
@robodoo r+ 
Sorry for the delay

agr-odoo:
@JulienAlardot 
I had to rebase
All :green_circle: now

JulienAlardot:
@robodoo r+

PR #246818, opened 2026-02-02
[IMP] mail: explicitly define singleton records

alexkuhn:
Before this commit, singleton records were implicit:
- Before [1] they relied on no definition of `static id`
- After [1] they either don't use default "id" field or had to hack their way to force not using `static id = "id"`

This commit improves definition of singleton through explicit definition `static singleton = true`. When this is set, the model enforces singleton, so all `.insert()` either creates record if none of this model exists or necessarily updates it.

Singleton records also have their local id format cleaned, with removal of ",undefined", which leaked the technical details of no `static id` means `undefined` is the id field, and singleton works thanks to no code attempting to insert data on this "fake" field.

Task-5893196

[1]: https://github.com/odoo/odoo/pull/246122

### DIFF
--- a/addons/mail/static/tests/discuss/call/call_settings_menu.test.js
+++ b/addons/mail/static/tests/discuss/call/call_settings_menu.test.js
@@ -124,11 +124,15 @@ test("local storage for call settings", async () => {
     localStorage.setItem(showOnlyVideoKey, toRawValue(true));
     const useBlurLocalStorageKey = makeRecordFieldLocalId(Settings.localId(), "useBlur");
     localStorage.setItem(useBlurLocalStorageKey, toRawValue(true));
+    const voiceActivationThresholdKey = makeRecordFieldLocalId(
+        Settings.localId(),
+        "voiceActivationThreshold"
+    );
     const callSettingsKeys = [
-        "Settings,undefined:backgroundBlurAmount",
-        "Settings,undefined:edgeBlurAmount",
-        "Settings,undefined:showOnlyVideo",
-        "Settings,undefined:voiceActivationThreshold",
+        backgroundBlurAmountKey,
+        edgeBlurAmountKey,
+        showOnlyVideoKey,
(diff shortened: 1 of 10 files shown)


addons/mail/static/tests/discuss/call/call_settings_menu.test.js
```diff
@@ -124,11 +124,15 @@ test("local storage for call settings", async () => {
     localStorage.setItem(showOnlyVideoKey, toRawValue(true));
     const useBlurLocalStorageKey = makeRecordFieldLocalId(Settings.localId(), "useBlur");
     localStorage.setItem(useBlurLocalStorageKey, toRawValue(true));
+    const voiceActivationThresholdKey = makeRecordFieldLocalId(
+        Settings.localId(),
+        "voiceActivationThreshold"
+    );
```
seb-odoo, line 130: Because the test needs to ensure it works whatever is computed now, not just for the hard-coded value that might become wrong if we change code?

seb-odoo (approved):
Seems alright, I didn't test yet.

Aren't we missing upgrade tests? I don't see a single test where we would assert the new key without undefined in it?

@robodoo delegate+

addons/mail/static/tests/discuss/call/call_settings_menu.test.js
alexkuhn: Yes, this is intentional for the test to stay true for recent values. If we want to define old values and have it work, this must be in upgrade tests.

alexkuhn:
> Aren't we missing upgrade tests? I don't see a single test where we would assert the new key without undefined in it?

The upgrade_19_1 tests do their work: they failed because the feature didn't work when upgrading to 99.9 version, so this includes the (lack of) 19.2 new scripts.

There's no need to tests every small steps.

Sure if there's new fields and features not covered by upgrade tests then we should add them, but in this case all features were already covered by upgrade tests.

PR #246820, opened 2026-02-02
[18.0][FIX] calendar: Display the description of an event using full width (similar to 16.0)

victoralmau:
Display the description of an event using full width (similar to 16.0 https://github.com/odoo/odoo/blob/1d4e5e796372190d9e8ee9030cd8f8d610c3437e/addons/calendar/views/calendar_views.xml#L181)

Same behavior as in v16.

**Before**
antes

**After**
despues

@Tecnativa TT60670

---

### DIFF
--- a/addons/calendar/views/calendar_views.xml
+++ b/addons/calendar/views/calendar_views.xml
@@ -211,7 +211,6 @@
                                         <field name="privacy" readonly="not user_can_edit" nolabel="1" placeholder="Default"/>
                                     </div>
                                     <field name="user_id" widget="many2one_avatar_user" readonly="not user_can_edit"/>
-                                    <field name="description" options="{'height': 100}" placeholder="Add description" readonly="not user_can_edit"/>
                         </group>
                         <group>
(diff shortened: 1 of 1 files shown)


addons/calendar/views/calendar_views.xml
```diff
@@ -242,6 +241,9 @@
                                 />
                             </div>
                         </group>
+                        <group colspan="2">
```
pedrobaeza, line 244: Why move this to the bottom? Can this be put next to where it was?

victoralmau: The same thing happens in v16: https://github.com/odoo/odoo/blob/d86dad61b62176bda2da652102c5ecabb852a487/addons/calendar/views/calendar_views.xml#L181

The important thing in this case is that the field has the full width, which is why the extra group is added at the end (as in v16).

pedrobaeza: OK, then please put this as `FIX` and remark that it's for keeping the same 16 layout and the problems the new one has.

PR #246831, opened 2026-02-02
[IMP] website: relocate popups after insertion

elgemmy:
Before this, popups were being inserted/moved inside other snippets, which caused confusion and styling/logic issues along with accidental deletion when the parent snippet was removed. Also, it could leak popup headings into structures like table of contents.

Now, on drop, page-specific popups are always relocated to a stable top-level savable container so we can avoid the previous issues.

---

### DIFF
--- a/addons/website/static/src/builder/plugins/options/popup_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/popup_option_plugin.js
@@ -1,6 +1,17 @@
-import { Plugin } from "@html_editor/plugin";
-import { registry } from "@web/core/registry";
 import { BuilderAction } from "@html_builder/core/builder_action";
+import { Plugin } from "@html_editor/plugin";
+import { withSequence } from "@html_editor/utils/resource";
+import { registry } from "@web/core/registry";
+
+function getPopupContainerFromSelectors(editable, selectors) {
+    for (const selector of selectors) {
+        const containerEl = editable.querySelector(selector);
+        if (containerEl) {
+            return containerEl;
+        }
+    }
+    return null;
+}
 
 export class PopupOptionPlugin extends Plugin {
     static id = "PopupOption";
@@ -31,9 +42,12 @@ export class PopupOptionPlugin extends Plugin {
             return popupModalChildrenEls.every((child) => child.matches(".s_popup_close"));
         },
         on_cloned_handlers: this.onCloned.bind(this),
-        on_snippet_dropped_handlers: this.onSnippetDropped.bind(this),
+        on_snippet_dropped_handlers: withSequence(0, this.onSnippetDropped.bind(this)),
+        // TODO remove when popup dragging from the page is disabled.
+        on_element_dropped_handlers: withSequence(0, this.onElementDropped.bind(this)),
         on_will_remove_handlers: this.onWillRemove.bind(this),
         no_parent_containers: ".s_popup",
+        popup_container_selectors: withSequence(10, "main .oe_structure.o_savable"),
     };
 
     onCloned({ cloneEl }) {
@@ -44,6 +58,7 @@ export class PopupOptionPlugin extends Plugin {
 
     onSnippetDropped({ snippetEl }) {
         if (snippetEl.matches(".s_popup")) {
+            this.relocatePopup(snippetEl);
             this.assignUniqueID(snippetEl);
             this.dependencies.history.addCustomMutation({
                 apply: () => {
@@ -71,6 +86,26 @@ export class PopupOptionPlugin extends Plugin {
     assignUniqueID(editingElement) {
         editingElement.closest(".s_popup").id = `sPopup${Date.now()}`;
     }
+
+    onElementDropped({ droppedEl }) {
+        if (droppedEl.matches(".s_popup")) {
+            this.relocatePopup(droppedEl);
+        }
+    }
+
+    relocatePopup(editingElement) {
+        const popupEl = editingElement.closest(".s_popup");
+        if (popupEl.closest("#o_shared_blocks")) {
+            return;
+        }
+        const containerEl = getPopupContainerFromSelectors(
+            this.editable,
+            this.getResource("popup_container_selectors")
+        );
+        if (containerEl) {
+            containerEl.insertAdjacentElement("afterbegin", popupEl);
+        }
+    }
 }
 
 // Moves the snippet in #o_shared_blocks to be common to all pages
@@ -84,10 +119,17 @@ export class MoveBlockAction extends BuilderAction {
             : value === "currentPage";
     }
     apply({ editingElement, value }) {
-        const selector = value === "allPages" ? "#o_shared_blocks" : "main .oe_structure.o_savable";
-        const whereEl = this.editable.querySelector(selector);
         const popupEl = editingElement.closest(".s_popup");
-        whereEl.insertAdjacentElement("afterbegin", popupEl);
+        const whereEl =
+            value === "allPages"
+                ? this.editable.querySelector("#o_shared_blocks")
+                : getPopupContainerFromSelectors(
+                      this.editable,
+                      this.getResource("popup_container_selectors")
+                  );
+        if (whereEl) {
+            whereEl.insertAdjacentElement("afterbegin", popupEl);
+        }
     }
 }
 export class SetBackdropAction extends BuilderAction {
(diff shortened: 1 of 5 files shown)


addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -21,6 +21,16 @@ export class PopupCookiesOption extends BaseOptionComponent {
     static applyTo = ".modal";
 }
 
+function getPopupContainerFromSelectors({ editable, getResource }) {
```
Syozik: I find making it a method of the plugin cleaner.

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -53,9 +63,12 @@ class PopupOptionPlugin extends Plugin {
             return popupModalChildrenEls.every((child) => child.matches(".s_popup_close"));
         },
         on_cloned_handlers: this.onCloned.bind(this),
-        on_snippet_dropped_handlers: this.onSnippetDropped.bind(this),
+        on_snippet_dropped_handlers: withSequence(0, this.onSnippetDropped.bind(this)),
+        // TODO remove when popup dragging from the page is disabled.
+        on_element_dropped_handlers: withSequence(0, this.onElementDropped.bind(this)),
         on_will_remove_handlers: this.onWillRemove.bind(this),
         no_parent_containers: ".s_popup",
+        popup_relocator_selectors: withSequence(10, "main .oe_structure.o_savable"),
```
Syozik: How about renaming it to `popup_container_selectors` ?

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -93,6 +107,24 @@ class PopupOptionPlugin extends Plugin {
...
+
+    onElementDropped({ droppedEl }) {
+        if (droppedEl.matches(".s_popup")) {
+            this.relocatePopup(droppedEl);
+        }
+    }
+
+    relocatePopup(editingElement) {
+        const popupEl = editingElement.closest(".s_popup");
+        if (popupEl.closest("#o_shared_blocks")) {
```
Syozik, line 98: Do we actually need to check for `#o_shared_blocks` here? We handle shared popups directly in the `MoveBlockAction`, right?

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -93,6 +107,24 @@ class PopupOptionPlugin extends Plugin {
...
+            return;
+        }
+        const containerEl = getPopupContainerFromSelectors(this);
+        if (!containerEl) {
+            return;
+        }
+        containerEl.insertAdjacentElement("afterbegin", popupEl);
```
Syozik: Nitpick, but instead of this, you may want to do this
```
if (containerEl) {
    containerEl.insertAdjacentElement("afterbegin", popupEl);
}
```
(or even `containerEl?.insertAdjacentElement("afterbegin", popupEl);`, but I personally like the other better)

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -106,9 +138,14 @@ export class MoveBlockAction extends BuilderAction {
...
-        const whereEl = this.editable.querySelector(selector);
         const popupEl = editingElement.closest(".s_popup");
+        const whereEl =
+            value === "allPages"
+                ? this.editable.querySelector("#o_shared_blocks")
+                : getPopupContainerFromSelectors(this);
+        if (!whereEl) {
+            return;
+        }
         whereEl.insertAdjacentElement("afterbegin", popupEl);
```
Syozik: Same as above.

Syozik (commented):
Nice work, thanks!
A few small concerns:
1. Perhaps it would be better to place a popup right before the end of the `containerEl`? 
So it would be
`containerEl.insertAdjacentElement("beforeend", popupEl);`
2. There's an issue when you will have a dropzone before an invisible popup, so then you can insert a snippet before the popup in the DOM

Image

And this will also create an unexpected extra dropzone:

Image

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
elgemmy: after double checking it, I believe it's more sensible to keep it as an external function as it's being used in `PopupOptionPlugin` and `MoveBlockAction`. 

moving the function inside would lead to unnecessary logic duplication. wdyt?

Syozik: You can make `getPopupContainerFromSelectors` shared in a plugin with `static shared`, so then you can add `PopupOption` as a dependency of `MoveBlockAction`, and in `MoveBlockAction` you can call `this.dependencies.PopupOption.getPopupContainerFromSelectors`

elgemmy: I see the benefit of moving it inside the plugin 👍🏻 

My concern is that it would make extra dependency purely for a helper utility, so I'm still slightly hesitant to move it. 

Would you agree that keeping it as a file-level helper is simpler for the current usecase?

Syozik: Alright, but then let's not pass `this` to the helper, but rather `this.editable` and selectors?

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
```diff
@@ -93,6 +107,23 @@ class PopupOptionPlugin extends Plugin {
...
+            this.relocatePopup(droppedEl);
+        }
+    }
+
+    relocatePopup(editingElement) {
+        const popupEl = editingElement.closest(".s_popup");
+        if (popupEl.closest("#o_shared_blocks")) {
+            return;
+        }
+        const containerEl = getPopupContainerFromSelectors(this);
```
Syozik:
```suggestion
        const selectors = this.getResource("popup_container_selectors");
        const containerEl = getPopupContainerFromSelectors(this.editable, selectors);
```
If we are keeping the helper outside of the plugin, I suggest making it like this, what do you think?

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
Syozik:
```suggestion
function getPopupContainerFromSelectors(editable, selectors }) {
    for (const selector of selectors) {
        const containerEl = editable.querySelector(selector);
        if (containerEl) {
            return containerEl;
        }
    }
    return null;
}
```
Then it will follow the name of the helper better

addons/website/static/tests/tours/conditional_visibility.js
```diff
@@ -52,6 +52,43 @@ function checkEyesIconAfterSave(footerIsHidden = true) {
...
+            }
+            for (let i = 1; i < indices.length; i++) {
+                if (indices[i - 1] >= indices[i]) {
+                    console.error(
+                        `error Unexpected Invisible Elements order. Expected: ${expectedNames.join(
+                            " < "
+                        )}`
+                    );
+                }
+            }
```
Syozik:
```suggestion
            const missing = [];
            let lastIndex = -1;
            
            for (const name of expectedNames) {
                const index = directPanelEntries.indexOf(name);
                
                if (index < 0) {
                    missing.push(name);
                } else if (index <= lastIndex) {
                    console.error(
                          `Unexpected Invisible Elements order. Expected: ${expectedNames.join(" < ")}`
                      );
                     return;
                } 
                 lastIndex = index;
            }
            
            if (missing.length) {
                console.error(
                    `Missing entries in Invisible Elements panel: ${missing.join(", ")}`
                );
            }
            }
```
Nitpick: I find single pass a bit more readable, and there's no need to mention `'error'` in `console.error()`.

addons/website/static/src/builder/plugins/options/popup_option_plugin.js
elgemmy: Agree, done!

addons/website/static/tests/tours/conditional_visibility.js
elgemmy: Agree on the first point 👍🏻 
for the error message, I think keeping it makes more sense for consistency reasons as it's already used throughout the file.

PR #246857, opened 2026-02-02
[FIX] Compute amount of loyalty using line currency

maq-adhoc:
When using the loyalty module with Chilean localization and a sale in UF or Dollars, an error occurs when calculating the promotion amount.

When the system evaluates whether a promotion is applicable, it appears to be taking the value of the sale line after applying the company currency rounding instead of the sale order currency rounding. In the case of Chile with the peso currency, where rounding is applied to the unit, if the value of a line is less than 0.50 UF, the system rounds it to 0. By interpreting the line value as zero, the promotion engine does not recognize the existence of valid products and the "Promotions/Reward" button does not display any available options, even though the program conditions are met.

The root cause of the problem is that the `compute_all` method of taxes is being called without specifying the currency, which causes it to use the company currency instead of the sale order currency.

Current behavior before PR:
- Create a Promotions program (Discount & Loyalty) that applies, for example, a 10% discount without complex minimum restrictions.
- Create a Sales Order and add a product line whose unit price is less than 0.50 (e.g. 0.34 UF).
- Click on the "reward" button. Result: No available promotions are shown or it does not apply if there is only one active promotion.

- Create a Promotions program (Discount & Loyalty) that applies, for example, a 10% discount without complex minimum restrictions.
- Create a Sales Order and add a product line whose unit price is less than 0.50 (e.g. 0.34 UF).
- Click on the "reward" button. Result: The button now detects the promotion and allows it to be applied.

---

### DIFF
--- a/addons/sale_loyalty/models/sale_order.py
+++ b/addons/sale_loyalty/models/sale_order.py
@@ -253,6 +253,7 @@ class SaleOrder(models.Model):
                 quantity=line.product_uom_qty,
                 product=line.product_id,
                 partner=line.order_partner_id,
+                currency=line.currency_id,
             )
             # To compute the discountable amount we get the subtotal and add
             # non-fixed tax totals. This way fixed taxes will not be discounted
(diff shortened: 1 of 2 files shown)


anko-odoo (changes requested):
Hello, the change itself looks fine, but please add a test.

masi-odoo:
Hello @maq-adhoc indeed this is reproducible in runbot.

@anko-odoo id did FUN test it is working as expected.

PR #246875, opened 2026-02-02
[IMP] website: introduce animated number snippet

romo-odoo:
This commit introduce a new snippet to allow users to add animated numbers (numbers that dynamically increase/decrease) on their website.

By changing the start and end values, the user can have an number that increase in an animation, to emphasize an improvement or a statistic. The snippet also enable a decreasing number.

Timing values (animation delay or duration) will be shared with onAppearance animations to avoid duplicating options, since in most cases the user will put the same value for both in they want to add an onAppearance animation.

task-5358340


addons/website/static/src/builder/plugins/options/animated_number_option.xml
```diff
@@ -0,0 +1,77 @@
...
+                classAction="'flex-column-reverse'"
+                actionValue="'bottom'"
+                iconImg="'/website/static/src/img/snippets_options/pos_bottom.svg'"
+                title.translate="Bottom"/>
+        </BuilderButtonGroup>
+    </BuilderRow>
+    <BuilderContext preview="false">
+        <BuilderRow 
+            label.translate="Start After" 
+            tooltip.translate="'On Appearance' animation will also use this value"
```
FrancoisGe: I m not sure to understand this tooltip ?

addons/website/static/src/interactions/base_animation.js
```diff
@@ -0,0 +1,107 @@
...
+
+export class BaseAnimation extends Interaction {
+    dynamicSelectors = {
+        ...this.dynamicSelectors,
+        _scrollingTarget: () => this.scrollingTarget,
+        _windowUnlessDropdown: () => this.windowUnlessDropdown,
+    };
+    dynamicContent = {
+        _window: { "t-on-resize": this.scrollWebsiteAnimate },
+        _windowUnlessDropdown: {
```
FrancoisGe, line 14: You just move this code or you change something ?

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,98 @@
...
+        if (!numberEl) {
+            return;
+        }
+        numberEl = firstLeaf(numberEl, (el) => el.childNodes.length != 1);
+        numberEl.textContent = this.startValue;
+        this.numberEl = numberEl;
+    }
+
+    start() {
+        if (this.el.closest(".dropdown")) {
```
FrancoisGe: Why ?

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,98 @@
...
+            return;
+        }
+        const { visible } = super.scrollWebsiteAnimate();
+        if (visible) {
+            this.startAnimation();
+        }
+    }
+
+    startAnimation() {
+        this.waiting = false;
```
FrancoisGe: Why `waiting` the name looks strange ? You want to check if the animation is already start or I miss something ?

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,98 @@
...
+
+        const numberEl = this.numberEl;
```
FrancoisGe: Why did you define it in the `setup` ?

FrancoisGe (commented):
Idea to discuss:
When you write something before/after the number => auto enable prefix/postfix
(It s looks strange to have this button prefix/postfix if we can add text) 


Strange:

Image 
This tooltip looks confusing in your situation

addons/website/static/src/interactions/base_animation.js
```diff
@@ -0,0 +1,107 @@
...
+    offsetMin = 10; // Minimum offset for small elements (in pixels)
+
+    setup() {
+        this.wrapwrapEl = document.querySelector("#wrapwrap");
+        this.windowUnlessDropdown = this.el.closest(".dropdown") ? [] : window;
+        this.scrollingElement = this.findScrollingElement();
+        this.scrollingTarget = isScrollableY(this.scrollingElement)
+            ? this.scrollingElement
+            : this.scrollingElement.ownerDocument.defaultView;
+        this.delay = undefined;
```
FrancoisGe, line 32: Not used

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,98 @@
...
+            numberEl.textContent = endValue;
+        } else {
+            this.waitForTimeout(() => {
+                const startTime = performance.now();
+
+                function animate() {
+                    if (this.forcedStop) {
+                        numberEl.textContent = endValue;
+                        return;
+                    }
```
FrancoisGe, line 78: It s not an arrow function, so the `this` shouldn't be correctly bind. (`this` will be the `this`of the function)

addons/website/static/src/interactions/base_animation.js
```diff
@@ -0,0 +1,107 @@
...
+        let scrollTop = this.scrollingElement.scrollTop;
+        if (closestModal && isVisible(closestModal)) {
+            scrollTop = closestModal.classList.contains("s_popup_no_backdrop")
+                ? closestModal.querySelector(".modal-content").scrollTop
+                : closestModal.scrollTop;
+        }
+        const elTop = this.getElementOffsetTop(el) - scrollTop;
+        let visible;
+        const footerEl = el.closest(".o_footer_slideout");
+        if (footerEl) {
```
FrancoisGe, line 91: Why did you remove `this.wrapwrapEl.classList.contains("o_footer_effect_enable")` ?

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,112 @@
...
+        } else if (!editingElement.querySelector(".s_animated_number_label")) {
+            const titleEl = document.createElement("div");
+            titleEl.classList.add(
+                "s_animated_number_label",
+                "d-flex",
+                "justify-content-center",
+                "align-items-center"
+            );
+            const h2El = document.createElement("h2");
+            h2El.textContent = "Clients";
```
FrancoisGe, line 62: You should use `_t()`

addons/website/static/src/builder/plugins/options/animated_number_option.xml
romo-odoo: To avoid setting the value twice (one for the number change and one for the onAppearance animation), the value if shared between them. The sub-option "Start After" below "onAppearance" is not shown.

addons/website/static/src/interactions/base_animation.js
romo-odoo: The code is moved to avoid duplication between `animation.js` and `animated_number.js`.
There shouldn't be any modification to the behavior of the `Animation` interaction.

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+import { _t } from "@web/core/l10n/translation";
+
+class AnimatedNumberOptionPlugin extends Plugin {
+    static id = "animatedNumberOption";
+    static dependencies = ["selection"];
+
+    /** @type {import("plugins").WebsiteResources} */
+    resources = {
+        so_content_addition_selectors: [".s_animated_number"],
+        is_unremovable_selectors: ".s_animated_number_display, .s_animated_number_value",
```
FrancoisGe: We maybe need the same feature in translate mode

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+            !selectionData.documentSelectionIsInEditable ||
+            selectionData.editableSelection.isCollapsed
+        ) {
+            return;
+        }
+
+        const selection = selectionData.editableSelection;
+        const numberEl = [...this.editable.querySelectorAll(".s_animated_number_value")].find(
+            (el) => selection.intersectsNode(el)
+        );
```
FrancoisGe: You don t want to use `findInSelection` ?
It will limit the querySelector to the commonAncestorContainer

It s not possible to have a selection on 2 counter ?

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+        if (!numberEl) {
+            return;
+        }
+
+        const numberLeaf = firstLeaf(numberEl);
+        if (!numberLeaf || numberLeaf.nodeType !== Node.TEXT_NODE) {
```
FrancoisGe: You should use `isTextNode`

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+            return;
+        }
+
+        const [startNode, startOffset, endNode, endOffset] = boundariesOut(numberEl);
+        const snapStart =
```
FrancoisGe: What do you mean by "snap" ?

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+                numberEl.classList.add("s_animated_number_value");
+                const fontSizeEl = this.document.createElement("span");
+                fontSizeEl.classList.add("h2-fs");
+                const fontWeightEl = this.document.createElement("strong");
+                fontSizeEl.append(fontWeightEl);
+                numberEl.append(fontSizeEl);
+                el.querySelector(".s_animated_number_display").replaceChildren(numberEl);
+                continue;
+            }
+            numberEl = firstLeaf(numberEl, (el) => el.childNodes.length != 1);
```
FrancoisGe: We don t have a way to make the node not editable ?
`is_unremovable_selectors` is not enough ?

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+            ".s_animated_number_prefix, .s_animated_number_postfix"
+        )) {
+            if (el.textContent == "") {
+                el.remove();
+            }
+        }
+    }
+
+    normalize(root) {
+        console.log("normalize");
```
FrancoisGe: console

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        const textContent = valueEl.textContent;
+        const valueTextNode = firstLeaf(valueEl, (el) => el.childNodes.length != 1);
+        const selection = this.dependencies.selection.getEditableSelection();
+        const digits = (textContent.match(/\d/g) || []).join("");
+        if (!digits) {
+            firstLeaf(valueEl).textContent = "0";
+            return;
+        }
```
FrancoisGe: Why ?

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,208 @@
...
+        const firstDigitIndex = textContent.search(/\d/);
```
FrancoisGe: It doesnt work if my fist char is a number. 

I feel like we're not taking the right approach here. For the “value” zone, there should be a special class that prevents us from modifying it. Or we could handle it in the `beforeinput` event: check if the character is in the “value” field, then add it to the prefix or create the prefix.

addons/website/static/src/snippets/s_animated_number/animated_number.js
FrancoisGe: you should fix it

addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -214,18 +215,24 @@ export class FormatPlugin extends Plugin {
...
      */
     removeFormats(formats, targetedNodes) {
-        const editableTargetedNodes = targetedNodes.filter(
-            this.dependencies.selection.isNodeEditable
-        );
         for (const format of formats) {
+            const formatTargetedNodes = targetedNodes.filter(
+                (node) =>
+                    this.dependencies.selection.isNodeEditable(node) ||
+                    this.isNonEditableFormatNode(node)
```
dmo-odoo: I don't understand. When I read `isNonEditableFormatNode` I get the sense that it's a format node that is not editable. Why `isNodeEditable` doesn't already return `false` ? This comment applies to every occurence of the same pattern.

addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -214,18 +215,24 @@ export class FormatPlugin extends Plugin {
         },
     };
 
+    isNonEditableFormatNode(node) {
+        return this.checkPredicates("is_node_format_only_predicates", node) ?? false;
```
dmo-odoo: What does it mean to be a "format only node" and what does it have to do with adding a new snippet?

addons/html_editor/static/src/main/font/color_plugin.js
```diff
@@ -98,6 +98,23 @@ export class ColorPlugin extends Plugin {
         }
     }
 
+    isFormatOnlyNode(node) {
+        return this.checkPredicates("is_node_format_only_predicates", node) ?? false;
+    }
+
+    isEditableOrFormatOnlyNode(node) {
+        return this.dependencies.selection.isNodeEditable(node) || this.isFormatOnlyNode(node);
+    }
```
dmo-odoo: This feels very ad-hoc.

addons/html_editor/static/src/main/toolbar/toolbar_plugin.js
```diff
@@ -370,6 +370,10 @@ export class ToolbarPlugin extends Plugin {
         }
     }
 
+    isNonEditableFormatNode(node) {
+        return this.checkPredicates("is_node_format_only_predicates", node) ?? false;
+    }
```
dmo-odoo: Wow, this concept of "format only node" is spilling in multiple plugins... cc @FrancoisGe 

If I understood the need correctly, maybe we should do something like "isFormattable" and make it possible that a node can be formattable but otherwise not editable? The predicates system usually supports that in the way of:
```js
isFormattable(node) {
    return this.checkPredicates("is_node_formattable", node) ?? this.checkPredicates("is_node_editable", node);
}
```
So we check the editability only if no predicate provided an information about formattability. What do you think?

romo-odoo:
@dmo-odoo

The goal is to have text that can be formatted but not edited. It is only for the snippet s_animated_number for now but we plan to use it in other cases.

I've updated the code and split the PR in two commits. Can you take another look ? (I'll check how to fix the remaining runbot issue in the meantime.

addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -219,13 +219,13 @@ export class FormatPlugin extends Plugin {
      * @param {Node[]} targetedNodes
      */
     removeFormats(formats, targetedNodes) {
-        const editableTargetedNodes = targetedNodes.filter(
-            this.dependencies.selection.isNodeEditable
+        const formatTargetedNodes = targetedNodes.filter(
+            this.dependencies.selection.isNodeFormattable
```
dmo-odoo: Whether a node is formattable or not is not the responsibility of the selection plugin. You can define the `isNodeFormattable` function in the format plugin instead.

addons/html_editor/static/src/main/font/color_plugin.js
```diff
@@ -130,10 +130,10 @@ export class ColorPlugin extends Plugin {
                 let max = 40;
                 const hasAnySelectedNodeColor = (mode) => {
                     const nodes = new Set();
-                    const editableTargetedNodes = this.dependencies.selection
+                    const formatTargetedNodes = this.dependencies.selection
```
dmo-odoo: I've seen this name used multiple times in this PR. I think that name is confusing because people might think it is format nodes that are targeted, like fonts or spans. Change all `formatTargetedNodes` to `formattableTargetedNodes`.

addons/html_editor/static/tests/format/bold.test.js
```diff
@@ -600,6 +601,29 @@ test("should not apply bold formatting for partial selection inside contentedita
...
+    const { editor, el } = await setupEditor(
+        `<p><span class="o-formattable" contenteditable="false">[Test]</span></p>`,
+        { config: { Plugins: [...MAIN_PLUGINS, TestPlugin] } }
+    );
+
+    bold(editor);
+
+    expect(getContent(el)).toBe(
+        `<p>[<span class="o-formattable" contenteditable="false"><strong>Test</strong></span>]</p>`
+    );
```
dmo-odoo: Functionally, I find that weird that the strong is not outside the span, but I guess that was precisely your goal? Why not put the format on the contenteditable="false" node instead? Because that one is actually editable, it's only its contents that are not. Though what happens if you select only partially the content? Can you format as well?

addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -375,26 +377,65 @@ export class FormatPlugin extends Plugin {
...
+        while (element && element !== this.editable) {
+            if (
+                !element.isContentEditable &&
+                this.isNodeFormattable(element) &&
+                this.dependencies.selection.areNodeContentsFullySelected(element)
+            ) {
+                return element;
+            }
+            element = element.parentElement;
+        }
```
deso-odoo: If the only goal is to allow styling of `contenteditable="false"` nodes, couldn't we follow the same approach as in `qweb_plugin` by using a combination of `formattable_node_providers` and `is_formattable_node_predicates`?

I'm not sure why all this additional code is needed here. My assumption is that the intent is to achieve the same behavior as `qweb_plugin`, but within `content_editable_plugin`.

addons/website/static/src/interactions/base_animation.js
```diff
@@ -0,0 +1,113 @@
...
+    offsetMin = 10; // Minimum offset for small elements (in pixels)
+
+    setup() {
+        this.wrapwrapEl = document.querySelector("#wrapwrap");
+        this.windowUnlessDropdown = this.el.closest(".dropdown") ? [] : window;
+        this.scrollingElement = this.findScrollingElement();
+        this.scrollingTarget = isScrollableY(this.scrollingElement)
+            ? this.scrollingElement
+            : this.scrollingElement.ownerDocument.defaultView;
+        this.targetEl = this.el;
```
FrancoisGe: `this.targetEl` is it used ?

addons/website/static/src/snippets/s_animated_number/animated_number.edit.js
```diff
@@ -0,0 +1,21 @@
+import { registry } from "@web/core/registry";
+import { AnimatedNumber } from "./animated_number";
+
+export const AnimatedNumberEdit = (I) =>
+    class extends I {
+        setup() {
+            super.setup();
+            this.stopAnimation = this._stopAnimation.bind(this);
+
+            this.el.addEventListener("click", this.stopAnimation, { once: true });
```
FrancoisGe: Why are you not using `this.addListener` ?

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,93 @@
...
+        const endValue = this.endValue;
+        const duration = this.duration;
+
+        if (window.matchMedia(`(prefers-reduced-motion: reduce)`).matches) {
+            numberEl.textContent = endValue;
+        } else {
+            this.waitForTimeout(() => {
+                const startTime = performance.now();
+
+                function animate() {
```
FrancoisGe: Did you check the issue about binding ?

addons/website/static/src/interactions/base_animation.js
```diff
@@ -0,0 +1,113 @@
...
+                return top;
+            }
+        } while (el);
+        return top;
+    }
+
+    scrollWebsiteAnimate() {
+        const el = this.targetEl;
+        if (el.classList.contains("o_animate_in_dropdown")) {
+            return { visible: false };
```
FrancoisGe, line 73: Is it a fix to return something now ?

FrancoisGe (commented):
I can create a link on the number, but it s removed by the interaction. we should disabled the create a link option in the sidebar.

Bug: 
Add a postfix 
Select counter + postfix and click on clear format => crash

addons/website/static/src/interactions/base_animation.js
romo-odoo: scrollWebsiteAnimate returns `{ visible, elTop, elHeight, windowsHeight }` as they are either used by `Animation` or `AnimatedNumber`

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,206 @@
...
+                .filter(Boolean)
+        );
+        for (const valueEl of valueElements) {
+            this.dependencies.color.colorElement(valueEl, color, mode);
+            [valueEl, ...descendants(valueEl)].forEach((node) => coloredNodes.add(node));
+        }
+    }
+
+    normalize(root) {
+        const displayEl = closestElement(root, ".s_animated_number_display");
```
FrancoisGe, line 55: What happens if `root` is a parent ? (root is the common ancestor)

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
```diff
@@ -0,0 +1,206 @@
...
+        if (!displayEl) {
+            return root;
```
FrancoisGe, line 57: Why did you return something ?

addons/website/static/src/builder/plugins/options/animated_number_option.xml
FrancoisGe: It s cleaner to move `styleAction="'animation-delay` to the `BuilderNumberInput`

FrancoisGe, line 18: I think you can just use a custom action for this `BuilderButton` and use `classAction`as a global action on the `BuilderButtonGroup`

addons/website/static/src/builder/plugins/options/animated_number_option_plugin.js
FrancoisGe: Do you really need a getter for one usecase ?

FrancoisGe: apply:  add d-none
clear: remove d-none

addons/website/static/src/snippets/s_animated_number/animated_number.edit.js
FrancoisGe: use arrow function

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,93 @@
...
+    setup() {
+        super.setup();
+
+        this.alreadyStarted = false;
+        this.forcedStop = false;
+
+        const numberEl = this.el.querySelector(".s_animated_number_value");
+        if (!numberEl) {
+            return;
+        }
```
FrancoisGe: Dead code, it s not possible because of `selectorHas`

addons/website/static/src/snippets/s_animated_number/animated_number.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    startAnimation() {
+        this.alreadyStarted = true;
+
+        const numberEl = this.numberEl;
+        if (!numberEl) {
+            return;
+        }
```
FrancoisGe: same dead code

addons/website/static/src/builder/plugins/options/animated_number_option.xml
```diff
@@ -0,0 +1,61 @@
...
+                iconImg="'/website/static/src/img/snippets_options/pos_top.svg'"
+                title.translate="Top"/>
+            <BuilderButton
+                classAction="'flex-row gap-2'"
+                iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"
+                title.translate="Left"/>
+            <BuilderButton
+                classAction="'flex-row-reverse gap-2'"
+                iconImg="'/website/static/src/img/snippets_options/pos_right.svg'"
+                title.translate="Right"/>
```
FrancoisGe, line 32: You should care about `ltrRtlMapping`

addons/html_editor/static/src/core/selection_plugin.js
```diff
@@ -1376,7 +1376,9 @@ export class SelectionPlugin extends Plugin {
         // Get up-to-date selection
         const { editableSelection } = this.getSelectionData();
         // Avoid setting the selection if it's not inside an uneditable element
-        const isInUneditable = (node) => !closestElement(node).isContentEditable;
+        const isInUneditable = (node) =>
+            !closestElement(node).isContentEditable &&
+            this.checkPredicates("is_node_editable_predicates", node) !== true;
```
FrancoisGe, line 1381: I think you can use `isNodeEditable` ?

romo-odoo: Here we use `!closestElement(node).isContentEditable;` while `isNodeEditable` uses `!!node.parentElement?.isContentEditable`.

There is a difference in our cases as the animated_number element has `contenteditable="false"`

FrancoisGe:
@robodoo r+ rebase-ff

addons/html_editor/static/src/core/selection_plugin.js
dmo-odoo: Wait, in your case `isNodeEditable` should return false for the animated number itself, correct? You have a predicate for this if I'm not mistaken. If it's a node _inside_ of it though, then that node is not editable since the animated number has `contenteditable="false"`, hence the `!!node.parentElement?.isContentEditable`.

I'm feeling like maybe we might not be talking about the same thing, though. What is the actual issue that you have if you remove those changes in `selection_plugin.js`? And what is the actual issue you have if you try:
```js
const isInUneditable = (node) => !isNodeEditable(node);
```

FrancoisGe:
@robodoo r+ rebase-ff

PR #246879, opened 2026-02-02
[IMP] website: add title element to countdown snippet

romo-odoo:
When the user dropped a countdown, there was no context to explain what
the countdown was for. Adding the countdown, not as an inner snippet,
gave no possibility to have the text next to the countdown.

Since we allow the user to drop the countdown on its own, we should
provide a way to easily add text to give context.

The option is similar to what is done for the s_social_media snippet
(with the option Title Position).

task-5449089

### DIFF
--- a/addons/website/static/src/builder/plugins/options/countdown_option.xml
+++ b/addons/website/static/src/builder/plugins/options/countdown_option.xml
@@ -2,6 +2,13 @@
 <templates xml:space="preserve">
 
 <t t-name="website.CountdownOption">
+    <BuilderRow label.translate="Title Position">
+        <BuilderButtonGroup action="'setCountdownTitlePosition'">
+            <BuilderButton actionParam="'s_countdown_title_top'" iconImg="'/website/static/src/img/snippets_options/pos_top.svg'"/>
+            <BuilderButton actionParam="'s_countdown_title_left'" iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"/>
+            <BuilderButton actionParam="''" icon="'fa-eye-slash'"/>
+        </BuilderButtonGroup>
+    </BuilderRow>
     <BuilderRow label.translate="Due Date">
         <BuilderDateTimePicker dataAttributeAction="'endTime'" acceptEmptyDate="false"/>
     </BuilderRow>
@@ -211,6 +218,16 @@
     <xpath expr="//t[@id='snippet_specific_options']" position="after">
         <countdown_option selector=".s_countdown"/>
     </xpath>
+</t>
+
+<t t-name="website.s_countdown.title">
+    <div class="s_countdown_title oe_unremovable">
+        <h2 style="text-align: center">
+            <span class="h3-fs">
+                <font class="text-o-color-1">Sales begins in:</font>
+            </span>
+        </h2>
+    </div>
 </t>
 
 </templates>
--- a/addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
@@ -4,6 +4,7 @@ import { getElementsWithOption } from "@html_builder/utils/utils";
 import { Plugin } from "@html_editor/plugin";
 import { registry } from "@web/core/registry";
 import { renderToElement } from "@web/core/utils/render";
+import { ClassAction } from "@html_builder/core/core_builder_action_plugin";
 
 export class CountdownOption extends BaseOptionComponent {
     static id = "countdown_option";
@@ -25,6 +26,7 @@ export class CountdownOptionPlugin extends Plugin {
             PreviewEndMessageAction,
             SetLayoutAction,
             SelectCountdownInlineTemplateAction,
+            SetCountdownTitlePositionAction,
         },
         on_cloned_handlers: ({ cloneEl }) => {
             const countdownEls = getElementsWithOption(cloneEl, ".s_countdown");
@@ -190,6 +192,23 @@ export class SelectCountdownInlineTemplateAction extends BuilderAction {
 
     clean(action) {
         return this.getAction("selectTemplate").clean(action);
+    }
+}
+
+class SetCountdownTitlePositionAction extends ClassAction {
+    static id = "setCountdownTitlePosition";
+    apply({ editingElement, params: { mainParam: className } }) {
+        super.apply(...arguments);
+        // Compatibility for old snippets (without a title element).
+        if (className && !editingElement.querySelector(".s_countdown_title")) {
+            const containerEl = editingElement.querySelector(
+                ".s_countdown_canvas_wrapper"
+            ).parentElement;
+            containerEl.insertAdjacentElement(
+                "afterbegin",
+                renderToElement("website.s_countdown.title")
+            );
+        }
     }
 }
 
--- a/addons/website/static/src/snippets/s_countdown/000.scss
+++ b/addons/website/static/src/snippets/s_countdown/000.scss
@@ -138,3 +138,24 @@
         }
     }
 }
+
+.s_countdown:not(.s_countdown_title_top, .s_countdown_title_left) {
+    .s_countdown_title {
+        display: none !important;
+    }
+}
+
+@include media-breakpoint-up(lg) {
+    .s_countdown_title_left {
+        > .container, > .container-fluid, > .o_container_small {
+            display: flex;
+            align-items: center;
+            justify-content: center;
+        }
+
+        .s_countdown_title {
+            width: 25%;
+            padding-right: 0.5rem;
+        }
+    }
+}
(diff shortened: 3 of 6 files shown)


addons/website/static/src/builder/plugins/options/countdown_option.xml
```diff
@@ -2,6 +2,13 @@
 <templates xml:space="preserve">
 
 <t t-name="website.CountdownOption">
+    <BuilderRow label.translate="Title Position">
+        <BuilderSelect action="'updateCountdownTitlePosition'">
+            <BuilderSelectItem actionParam="{ position: 'top' }">Top</BuilderSelectItem>
+            <BuilderSelectItem actionParam="{ position: 'left' }">Left</BuilderSelectItem>
+            <BuilderSelectItem actionParam="{}">None</BuilderSelectItem>
+        </BuilderSelect>
+    </BuilderRow>
```
bvr-odoo, line 11: I suggest buttons with icons instead of a dropdown. Wdyt ?

Image

addons/website/views/snippets/s_countdown.xml
```diff
@@ -9,6 +9,9 @@
         data-progress-bar-style="surrounded" data-progress-bar-weight="thin"
         data-text-color="o-color-1" data-layout-background-color="400" data-progress-bar-color="o-color-1">
         <div class="container">
+            <div class="s_countdown_title d-flex justify-content-center align-items-center" style="color: var(--o-color-1);">
```
bvr-odoo: This is not how the html editor adds text color. Look if you apply a color on a header

Image

FrancoisGe:
@robodoo delegate=bvr-odoo

addons/website/views/snippets/s_countdown.xml
```diff
@@ -9,6 +9,9 @@
...
+                <h2>Sales Begins in :</h2>
+            </div>
```
bvr-odoo, line 18: text alignment doesn't work 😢 

Image

bvr-odoo: The left-aligned text should not take up so much width (The space it takes on each line depends on the length of the sentence, which is very strange). Maybe wrap the title in an `oe_unremovable` element or something like that. Check the products snippet, it has the same option. We can talk about it togheter

Image

addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
```diff
@@ -196,4 +197,35 @@ export class SelectCountdownInlineTemplateAction extends BuilderAction {
     }
 }
 
+class UpdateCountdownTitlePositionAction extends BuilderAction {
+    static id = "updateCountdownTitlePosition";
+    apply({ editingElement, params: { position } }) {
+        const containerEl = editingElement.querySelector(":scope > .container");
+        let titleEl = editingElement.querySelector(".s_countdown_title");
+        if (!position) {
+            titleEl?.remove();
```
bvr-odoo: - Why not simply hide it with `d-none`, the same way we do for the "products" snippet? (This way, this JS might not even be needed anymore.)
- It would also prevent the text from being removed when the user only wants to change the option to see the result, or hide it briefly for some reason.
(Note that in that case, we will need to make sure the option is not available for old countdown snippets already dropped on the page)

addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
```diff
@@ -196,4 +197,35 @@ export class SelectCountdownInlineTemplateAction extends BuilderAction {
     }
 }
 
+class UpdateCountdownTitlePositionAction extends BuilderAction {
```
bvr-odoo: - Couldn’t you extend `ClassAction` instead of handling the classes in this builder action?

addons/website/static/src/snippets/s_countdown/000.scss
```diff
@@ -138,3 +138,7 @@
         }
     }
 }
+
+.s_countdown > .container.d-flex {
+    gap: 10px;
+}
```
bvr-odoo: There probably should not be a gap between everything, only to the right of the title. In any case, we should handle this by applying the right class to the element wrapping the title instead of doing this.

Image

addons/website/static/src/builder/plugins/options/countdown_option.xml
```diff
@@ -2,6 +2,13 @@
 <templates xml:space="preserve">
 
 <t t-name="website.CountdownOption">
+    <BuilderRow label.translate="Title Position">
+        <BuilderSelect action="'updateCountdownTitlePosition'">
+            <BuilderSelectItem actionParam="{ position: 'top' }">Top</BuilderSelectItem>
```
bvr-odoo: Why use an `object` when it only stores top or left?

addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
```diff
@@ -196,4 +197,35 @@ export class SelectCountdownInlineTemplateAction extends BuilderAction {
...
+    static id = "updateCountdownTitlePosition";
+    apply({ editingElement, params: { position } }) {
+        const containerEl = editingElement.querySelector(":scope > .container");
```
bvr-odoo: Be careful: this `.container` class can be replaced when the section’s `content width` option is changed. For example, if you change the `content width` option here and then play a bit with the `title position` option, you get a traceback.

bvr-odoo (changes requested):

addons/website/views/snippets/s_countdown.xml
```diff
@@ -2,13 +2,20 @@
...
 <template id="s_countdown" name="Countdown">
-    <section class="s_countdown pt48 pb48"
+    <section class="s_countdown s_countdown_title_top pt48 pb48"
         data-display="dhms" data-end-action="nothing" data-size="175"
         t-att-data-end-time="60 * round(datetime.datetime.now().timestamp() / 60) + 228300"
         data-layout="circle" data-layout-background="none"
         data-progress-bar-style="surrounded" data-progress-bar-weight="thin"
         data-text-color="o-color-1" data-layout-background-color="400" data-progress-bar-color="o-color-1">
         <div class="container">
+            <div class="s_countdown_title oe_unremovable">
```
bvr-odoo, line 12: why `oe_unremovable` here and not in `website.s_countdown.title`?

addons/website/views/snippets/s_countdown.xml
```diff
@@ -2,13 +2,20 @@
...
+                <h2 style="text-align: center">
+                    <span class="h3-fs">
```
bvr-odoo, line 14: why this `<span>` here and not in `website.s_countdown.title`?

addons/website/static/src/builder/plugins/options/countdown_option_plugin.js
```diff
@@ -193,4 +195,20 @@ export class SelectCountdownInlineTemplateAction extends BuilderAction {
     }
 }
 
+class SetCountdownTitlePositionAction extends ClassAction {
+    static id = "setCountdownTitlePosition";
+    apply({ editingElement, params: { mainParam: className } }) {
+        super.apply(...arguments);
+        if (className && !editingElement.querySelector(".s_countdown_title")) {
```
bvr-odoo, line 203: Add a comment to explain that this is for compatibility with old countdown snippets.

addons/website/static/src/snippets/s_countdown/000.scss
```diff
@@ -138,3 +138,24 @@
...
+
+.s_countdown:not(.s_countdown_title_top, .s_countdown_title_left) {
+    .s_countdown_title {
+        display: none !important;
+    }
+}
+
+@include media-breakpoint-up(lg) {
+    .s_countdown_title_left {
+        & > div {
```
bvr-odoo: the `&` is not necessary

bvr-odoo: This should only be applied to the container. Currently, the rule is also applied to divs with `o_we_bg_filter` or `o_we_shape` when you add a background shape or filter.


```suggestion
        > .container, > .container-fluid, > .o_container_small {
```

addons/website/static/src/snippets/s_countdown/000.scss
```diff
@@ -138,3 +138,24 @@
...
+            display: flex;
+            align-items: center;
+            justify-content: center;
+        }
+        
+        .s_countdown_title {
+            width: 25%;
+            padding-inline-end: 0.5rem;
```
bvr-odoo: I don't know this rule, why not use `padding-right`?

addons/website/views/snippets/s_countdown.xml
bvr-odoo: I still have the same question :)
But we'll take a look at it at the office because I spotted a bug with the alignment option.

romo-odoo: Sorry I was sure I added `oe_unremovable` before resolving your comment.
Since you did not reproduced the alignment issue, I think we can just keep it in mind and merge.

bvr-odoo:
@robodoo r+

PR #246906, opened 2026-02-03
[IMP] hr_expense: replace 'toutes taxes comprises' with 'ttc'

ppat-odoo:
In French, replace 'toutes taxes comprises' with 'ttc' for better UI alignment.

task-5712223

### DIFF
--- a/addons/hr_expense/i18n/fr.po
+++ b/addons/hr_expense/i18n/fr.po
@@ -101,7 +101,7 @@ msgstr ""
 #. module: hr_expense
 #: model_terms:ir.ui.view,arch_db:hr_expense.hr_expense_view_form
 msgid "(incl"
-msgstr "(toutes taxes comprises"
+msgstr "(ttc"
 
 #. module: hr_expense
 #. odoo-python


addons/hr_expense/i18n/fr.po
```diff
@@ -101,7 +101,7 @@ msgstr ""
 #. module: hr_expense
 #: model_terms:ir.ui.view,arch_db:hr_expense.hr_expense_view_form
 msgid "(incl"
-msgstr "(toutes taxes comprises"
+msgstr "(ttc"
```
zepa-odoo, line 104: Based on the discussion with the **translation team**, we should not change this translation to "ttc".
The string "incl" comes from the expression `(incl <tax_amount> tax)` and should not be interpreted as "toutes taxes comprises" or replaced with "ttc".
If any update is needed, it should be handled through [Weblate ](https://translate.odoo.com/), not directly in the code.

zepa-odoo (commented):
Hello @ppat-odoo 
Thanks for this pr!

ppat-odoo:
This will be handled by translation team.

PR #246911, opened 2026-02-03
[FIX] stock_account: multiple stock moves due to BOM breaks invoice

joswa-odoo:
Current behavior:
--
When an invoice that contains a kit product is created, it crashes at the _get_cogs_price_unit() function

Expected behavior:
---
An invoice should be able to be created even when containing kit products

Steps to reproduce:
---
1. Create a clean database
2. Turn on the locations setting
3. Create a product with a BOM with atleast 2 components and set it to kit
4. Create a product category and set the invoice period to perpetual
5. Create a sales order containing the product and confirm
6. Validate the Delivery
7. Create an invoice and confirm

Cause of the issue:
---
the code self.product_id only works
The code inside _get_cogs_price_unit() single records or a record set with the same product.

Fix:
---
Add an if statement before the _get_cogs_price_unit() call to handle kit products buy checking if the first stock_move is the same as the product used as the main product

opw-5879615
---------------------------------------------------------------

### DIFF
--- a/addons/sale_mrp/models/account_move.py
+++ b/addons/sale_mrp/models/account_move.py
@@ -7,8 +7,6 @@ class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     def _get_cogs_value(self):
-        price_unit = super()._get_cogs_value()
-
         so_line = self.sale_line_ids and self.sale_line_ids[-1] or False
         if so_line:
             # We give preference to the bom in the stock moves for the sale order lines
@@ -38,5 +36,5 @@ class AccountMoveLine(models.Model):
                     prod_moves = moves.filtered(lambda m: m.product_id == product)
                     product = product.with_company(self.company_id)
                     average_price_unit += factor * prod_moves._get_price_unit()
-                price_unit = average_price_unit / bom.product_qty or price_unit
-        return price_unit
+                return average_price_unit / bom.product_qty
+        return super()._get_cogs_value()


addons/stock_account/tests/test_account_move.py
```diff
@@ -338,3 +338,96 @@ def test_invoice_with_journal_item_without_label(self):
...
+                    'quantity': 1.0,
+                    'location_id': inv_loc.id,
+                    'location_dest_id': cust_loc.id,
+                    'picked': True
+                    }),
+                ],
+        }
+        ])
+
+        sale_order = self.env['sale.order'].create({
```
ande-odoo: Because you're using "sale.order" you should put your test in sale_stock instead. That's also where the _get_stock_moves override is.

addons/stock_account/models/account_move_line.py
```diff
@@ -65,6 +65,9 @@ def _get_cogs_value(self):
 
         cogs_qty = self._get_cogs_qty()
         if moves := self._get_stock_moves().filtered(lambda m: m.state == 'done'):
+            # if the moves have products that are not the same as the initial product, then its a kit
+            if moves[0].product_id != self.product_id:
+                return (sum(moves.mapped('value')) * cogs_qty - self._get_posted_cogs_value()) / self.quantity
```
Whenrow: Could you make an override of `_get_posted_cogs_value` in sale_mrp to handle kit. Also move your test to the corresponding module. Thank you

PR #246913, opened 2026-02-03
[FIX] web, project, payment: Fix extractProps and update props in CopyClipboardField

Mazen-Ghaleb:
The `CopyClipboardField` widget previously relied on the `string` attribute inside `attributes` to determine the text of copy button. In Odoo 17+, the `string` attribute is no longer inside `attributes` but defined outside, causing the widget to break. 

Additionally, using `string`for the copy text introduced ambiguity between the field label and the actual text to copy. The intended approach is to define the copy text through widget options (as shown in the project share link), but directly using `copyText`, in options previously caused translation issues, as discussed in PR #245826.

This PR resolves the translation issue by passing the copy text through a dedicated char field, similar to how color is passed in  fields and currency is passed in monetary fields. The widget handles if this field is not stored too by using virtual records.

In the Payment Wizard button, the widget was inherited and `extractProps` was overridden to handle the `string` attribute manually, which is no longer necessary with the new approach.

- `CopyClipboardField` reads the removed `string` attribute from `attributes`.  
- Copy text is ambiguous and can be confused with the field label.
- Project share link usage fails to read the copy text correctly.
- `PaymentWizardCopyClipboardButtonField` overrides `extractProps` unnecessarily to fix copy text handling.

- `CopyClipboardField` reads text from field in  `copy_text_field` defined in widget options.
- Copy text is clearly separated from field labels.  
- Project share link works correctly using widget options. 
- `PaymentWizardCopyClipboardButtonField` no longer needs to override `extractProps` and copy text is set directly via field options.
- Submit/copy text can be customized through field options.

---

### DIFF
--- a/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
+++ b/addons/payment/static/src/js/payment_wizard_copy_clipboard_field.js
@@ -22,10 +22,6 @@ class PaymentWizardCopyClipboardButtonField extends CopyClipboardButtonField {
 const paymentWizardCopyClipboardButtonField = {
     ...copyClipboardButtonField,
     component: PaymentWizardCopyClipboardButtonField,
-    extractProps: (fieldInfo, dynamicInfo) => ({
-        ...copyClipboardButtonField.extractProps(fieldInfo, dynamicInfo),
-        string: fieldInfo.string,
-    }),
 };
 
 registry
--- a/addons/payment/static/tests/payment_wizard_copy_clipboard_field_tests.js
+++ b/addons/payment/static/tests/payment_wizard_copy_clipboard_field_tests.js
@@ -16,6 +16,11 @@ QUnit.module("Payment", {
                     fields: {
                         amount: { string: "Amount", type: "float" },
                         link: { string: "Payment Link", type: "char" },
+                        link_copy_text: {
+                            string: "Custom Link Copy",
+                            type: "char",
+                            default: "Generate and Copy Payment Link",
+                        },
                     },
                     onchanges: {
                         amount(record) {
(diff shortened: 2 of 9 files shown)


AntoineVDV:
Hey @jeanschoenlaub, in this case, it's a community PR (not open by an Odoo employee), so there is no quadrigram to append :sweat_smile:

Mazen-Ghaleb-Upward:
Hey @jeanschoenlaub ,

Thanks for your comment!

As I briefly mentioned here and explained in more detail in the previous PR  #245826, the core issue lies within the widget itself.

In Odoo 17, the `string` attribute is no longer available in `attributes`. Additionally, relying on the `string` attribute creates ambiguity, as it becomes unclear whether it represents the field label or the specific copy text intended for the widget.

In my previous PR #245826, I attempted to solve this by passing `copyText` through the widget options. However, that approach did not properly handle translations.

In this PR, I resolved the issue by passing a char text field instead. This ensures proper translation handling and avoids ambiguity. I also implemented a same approach for submit text and added JavaScript unit tests to cover both cases in base widget and in payment widget.

# Usage of Custom Copy Text

Currently, there are two main places where this widget is used with custom copy text.

### Payment

Previously, a completely new widget was adjusting the copy text, and it was used in the payment wizard.  
With this PR, that workaround is no longer necessary in the payment copy widget.

### Project

The widget is used in the share project copy link wizard.

# Steps to Reproduce in Project

1. Go to any project form view.
2. Click on the action cogwheel.
3. Select "Share Project".
4. You will notice that the copy button text for the field labeled "Link" is shown as "Copy" instead of the intended "Copy Link" defined in XML.

I understand this may seem like a minor UI issue, but the real concern is the widget’s architecture and how it will impact future usages.

At the moment, in order to use custom copy text, one must build an entirely new widget, as done in Payment. This does not seem like the correct approach. The base widget itself should support this functionality properly and work as intended.

Another possible solution would be to rely on the `string` attribute and adjust the base widget JavaScript so it no longer reads `string` from attributes. However, this would still maintain the ambiguity between the field label and the copy text. The intent would not be explicit or clear in XML, and it would affect both the label and the copy text unless the label is defined separately using `label` tag and nolabel attribute on field tag. 

Thank you

PR #236931, opened 2025-11-21
[FIX] l10n_de: fix tax access error when changing product cost

yagho-odoo:
from v18.0 to v18.4, in a multi-company environment, when a product with no income/expense account had its cost changed, all the taxes for other companies were being fetched which caused access errors when trying to view it after a manual save.
This happens because the fetch is happening in a sudo environment because the stock valuation layer was being created as sudo, so all taxes were being fetched and probably because of cache pollution they were not being filtered properly, this is not happening in v19.0 because the stock valuation layer was removed, so everything is being called in a normal user environment, refer to this commit-08b62a4

task-5117882

PR #246920, opened 2026-02-03
[FIX] l10n_de,stock_account: prevent accesserror when confirming the order on eccomerce

krip-odoo:
**Steps to reproduce:**
- Install `l10n_de, website_sale and appointment` modules.
- Create a website for the `DE company` and make it default.
 (simply place it first in the sequence.)
- From settings enable `Automatic Invoice` and `Discounts, Loyalty & Gift Card`.
- Go to the appointment module and create a new appointment for the DE company.
- In the options tab, enable `Up-front payment` and publish it.
- Create a new gift card program and generate a gift card to test (should cover the entire cost of the appointment booking).
- Now go to an Incognito tab, go to the appointment, and book the currently created appointment. When confirming the order, use the gift card and then confirm the order.

**Issue:**
- When you confirm the order, an access error occurs.

**Root cause:**
- Since automatic invoicing is enabled, at [1] the method
  `get_product_accounts` is called for the DE company.
  Inside this method at [2], self.sudo(False) is used, which removes the
  elevated access rights. As a result, the `Public product template` record
  rule is triggered, and because the public user cannot access the product,
  an AccessError is raised.
- In this [commit], we can see that `sudo(False)` was added when coming from
  the stock flow because the stock valuation layer was being created as sudo.

**Solution:**
- Instead of adding `sudo(False)` in **l10n_de**, we can apply `sudo(False)`
  at the point where the **stock valuation layer** is created using `sudo()`.

[1]https://github.com/odoo/odoo/blob/9d5d5c08f950d32bf61c2186181e8eddad9036da/addons/account/models/account_move_line.py#L603-L604
[2]https://github.com/odoo/odoo/blob/9d5d5c08f950d32bf61c2186181e8eddad9036da/addons/l10n_de/models/datev.py#L20

[commit]: https://github.com/odoo/odoo/pull/236931/changes/0919774791bb998954b870798c88843f999f2de4

opw-5483472
---

### DIFF
--- a/addons/l10n_de/tests/test_account_move.py
+++ b/addons/l10n_de/tests/test_account_move.py
@@ -53,3 +53,19 @@ class TestAccountMoveDE(AccountTestInvoicingCommon):
         self.assertRecordValues(move.line_ids, expected_lines_vals)
         move.action_post()
         self.assertRecordValues(move.line_ids, expected_lines_vals)
+
+    def test_public_user_invoice_creation_with_product_record_rule(self):
+        public_user = self.env['res.users'].create({'name': 'Public User test', 'login': 'blablabla', 'is_public': True})
+        self.product_a.write({'property_account_income_id': False, 'property_account_expense_id': False})
+        model = self.env['ir.model']._get('product.template')
+        self.env['ir.rule'].create({
+            'name': "Only The Lone Wanderer allowed",
+            'model_id': model.id,
+            'domain_force': [('name', '=', 'coucou')],
+        })
+        move = self.env['account.move'].with_user(public_user).create({
+            'move_type': 'out_invoice',
+            'partner_id': self.partner_a.id,
+            'invoice_line_ids': [Command.create({'product_id': self.product_a.id})],
+        })
+        self.assertTrue(move)
--- a/addons/stock_account/tests/test_account_move.py
+++ b/addons/stock_account/tests/test_account_move.py
@@ -5,6 +5,7 @@ from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 from odoo.addons.stock_account.tests.test_stockvaluation import _create_accounting_data
 from odoo.tests import Form, tagged
 from odoo import fields, Command
+from unittest.mock import patch
 
 class TestAccountMoveStockCommon(AccountTestInvoicingCommon):
     @classmethod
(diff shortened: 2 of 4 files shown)


alsh-odoo (approved):

alsh-odoo (changes requested):
Thanks for the fix!

From what I can see, the issue was introduced by this commit:
https://github.com/odoo/odoo/commit/078b733fb00cea8b9e6b1aa139d109d2d7ef97e5
Would it be possible to apply sudo(False) at the point where this method is triggered for the above use case, instead of applying it directly here? 🤔
I'm not entirely sure though, as that might potentially introduce other side effects as well.
or you can pass a context key from where we are applying the sudo() for our usecase and handle it in l10n_de.

addons/l10n_de/tests/test_account_move.py
```diff
@@ -53,3 +53,20 @@ def test_out_invoice_custom_currency_rate_with_missing_delivery_date(self):
...
+            'categ_id': self.env.ref('product.product_category_all').id,
+            'taxes_id': [(6, 0, [self.company_data['default_tax_sale'].id])],
+        })
+        product.quantity_svl = 10.0
+        product.standard_price = 100
+        specification = {
+            'standard_price': {},
+            'taxes_id': {'fields': {'name': {}}},
+        }
+        product.web_read(specification)
```
guva-odoo: This test will fail if `stock_account` module is not installed, you need to:

1 - Move the test to `stock_account` and replicate the same kind of flow in it (preferred option)
2 - If option 1 not possible, you need to skip the test if `stock_account` is not installed

guva-odoo (changes requested):

addons/stock_account/tests/test_account_move.py
```diff
@@ -442,3 +443,34 @@ def test_invoice_with_journal_item_without_label(self):
...
+                    ('internal_group', '=', 'expense'),
+                    ('tax_ids', 'in', tax_expense.ids),
+                ], limit=1),
+            }
+        product = self.env['product.product'].create({
+            'name': 'Test Product',
+            'type': 'consu',
+            'is_storable': True,
+            'categ_id': self.env.ref('product.product_category_all').id,
+            'taxes_id': [(6, 0, [self.company_data['default_tax_sale'].id])],
```
guva-odoo:
```suggestion
            'taxes_id': [Command.set(self.company_data['default_tax_sale'].ids])],
```

addons/stock_account/tests/test_account_move.py
```diff
@@ -442,3 +443,34 @@ def test_invoice_with_journal_item_without_label(self):
...
+                    ('internal_group', '=', 'expense'),
+                    ('tax_ids', 'in', tax_expense.ids),
+                ], limit=1),
+            }
+        product = self.env['product.product'].create({
+            'name': 'Test Product',
+            'type': 'consu',
+            'is_storable': True,
+            'categ_id': self.env.ref('product.product_category_all').id,
+            'taxes_id': [Command.set([self.company_data['default_tax_sale'].id])],
```
guva-odoo:
```suggestion
            'taxes_id': [Command.set(self.company_data['default_tax_sale'].ids)],
```

guva-odoo (approved):
Thanks! LGTM
Just a little change, but we can move on to the next review

epictete (approved):
Thanks for the fix 👌 

@robodoo r+

PR #246921, opened 2026-02-03
[FIX] website: kept animation intensity visible on mode change

alap-odoo:
Steps to reproduce:
1. Go to the website and enter edit mode.
2. Click any text in the footer.
3. Change the animation mode from _none_ to _on appearance._
4. Select the _zoom out_ effect; the intensity slider is visible.
5. Change the animation mode to _on scroll_.

Issue:
The animation intensity slider is no longer visible. It only reappears after reselecting the same effect.

Reason:
The `apply` method of `SetAnimationModeAction` always adds the "**o_anim_fade_in**" class, even when an animation effect is already selected. When switching the animation mode, this additional class prevents the intensity slider from being displayed.

task-5896549

### DIFF
--- a/addons/website/static/src/builder/plugins/options/animate_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/animate_option_plugin.js
@@ -27,7 +27,12 @@ import { EmphasizeAnimatedText } from "./emphasize_animated_text";
 export class AnimateOptionPlugin extends Plugin {
     static id = "animateOption";
     static dependencies = ["history", "selection", "split"];
-    static shared = ["forceAnimation", "getDirectionsItems", "getEffectsItems"];
+    static shared = [
+        "forceAnimation",
+        "getDirectionsItems",
+        "getEffectsItems",
+        "hasAnimationEffect",
+    ];
     /** @type {import("plugins").WebsiteResources} */
     resources = {
         builder_options: [withSequence(ANIMATE, AnimateOption)],
@@ -106,6 +111,20 @@ export class AnimateOptionPlugin extends Plugin {
             { className: "o_anim_from_bottom_left", label: "From bottom left", check: isRotate },
         ];
     }
+
+    /**
+     * Checks whether the given element contains any animation class from the
+     * list returned by getEffectsItems().
+     *
+     * @param {HTMLElement} editingElement- The element to check
+     * @returns {boolean} True if at least one animation class is present
+     */
+    hasAnimationEffect(editingElement) {
+        return this.getEffectsItems().some(({ className }) =>
+            editingElement.classList.contains(className)
+        );
+    }
+
     async forceAnimation(editingElement) {
         editingElement.style.animationName = "dummy";
         if (editingElement.classList.contains("o_animate_on_scroll")) {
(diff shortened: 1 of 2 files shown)


addons/website/static/tests/builder/website_builder/animate_option.test.js
```diff
@@ -646,3 +646,29 @@ describe("animate text in toolbar", () => {
...
+
+    await contains(".options-container [data-label='Effect'] .dropdown-toggle").click();
+    await contains(".o-dropdown--menu [data-action-value='o_anim_zoom_out']").click();
+    await waitSidebarUpdated();
+    expect(".options-container [data-label='Intensity']").toHaveCount(1);
+
+    await contains(".options-container [data-label='Animation'] .dropdown-toggle").click();
+    await contains(".o-dropdown--menu [data-action-value='onScroll']").click();
+    await waitSidebarUpdated();
+    expect(".options-container [data-label='Intensity']").toHaveCount(1);
```
shsa-odoo: If aim is to simply test the intensity, it could have been done in continuation [here](https://github.com/odoo/odoo/pull/246921/files#diff-881fc49b1f5971664333eb225dd83ab1f5717a4185b24e6f89c4a3c9b1eb77b9R167)  IG: :slightly_smiling_face:

alap-odoo: From my perspective, adding a new test case helps explicitly define the specific functionality being validated. Additionally, this file already follows a pattern where each option combination is covered by an individual test, for example:

`visibility of animation animation=onAppearance`
`visibility of animation animation=onAppearance effect=slide`
`visibility of animation animation=onAppearance effect=bounce`

Given that structure, I followed this approach. However, will add the steps at requested position.

addons/website/static/src/builder/plugins/options/animate_option_plugin.js
```diff
@@ -392,7 +392,11 @@ export class SetAnimationModeAction extends BuilderAction {
     }
 
     async apply({ editingElement, value: effectName, params: { forceAnimation } }) {
-        if (this.animationWithFadein.includes(effectName)) {
+        // Prevent adding fade-in when another animation class is present.
+        if (
+            this.animationWithFadein.includes(effectName) &&
+            !/o_anim_/.test(editingElement.className)
```
shsa-odoo: If someone adds a class like `data_o_anim_something`, whether intentionally (customization, rarely a case but..) or by mistake it could produce a false match. As a result, when toggling to "onAppearance", the fade effect would not be applied.

why haven't we used `getEffectsItems` and `getDirectionsItems` which contains valid animation classes ?
This way it only matches actual, defined animation classes, not random strings that happen to start with "o_anim_" . :thinking:

shsa-odoo (changes requested):
@alap-odoo Concern to address, added a [comment](https://github.com/odoo/odoo/pull/246921#discussion_r2939443532). :) Thanks!!

addons/website/static/src/builder/plugins/options/animate_option_plugin.js
alap-odoo: Since it was a very rare case, I thought proceeding with my solution would not have much impact, and the current implementation was shorter and easier to understand. However, we can update it as per your suggestion.

shsa-odoo (approved):
LGTM :)

addons/website/static/src/builder/plugins/options/animate_option_plugin.js
```diff
@@ -392,7 +392,16 @@ export class SetAnimationModeAction extends BuilderAction {
...
-        if (this.animationWithFadein.includes(effectName)) {
+        const { getEffectsItems, getDirectionsItems } = this.dependencies.animateOption;
+        const hasKnownAnimationClass =
+            getEffectsItems(() => false).some(({ className }) =>
+                editingElement.classList.contains(className)
+            ) ||
+            getDirectionsItems().some(
+                ({ className }) => className && editingElement.classList.contains(className)
+            );
+        // Prevent adding fade-in when another animation class is present.
```
FrancoisGe: You can add a share function `hasAnimationEffect` or `getAnimationEffect` in `animateOption`. It will be more clear.
Why do you need to check the direction ? It s not enough to check the effect ?

PR #246924, opened 2026-02-03
[FIX] web, calendar: calendar popover: use "text-wrap" for the description field

victoralmau:
Changes done:
- [x] `calendar`: Add `class="text-wrap"` in the description field of the calendar view to use it in the popover
- [x] `web`: Define the appropriate class in the calendar popover field

**Before**
antes

**After**
despues

@Tecnativa TT60670

---


aab-odoo (approved):
Thanks everyone

robodoo r+

PR #246925, opened 2026-02-03
[FIX] pos_self_order: catch error when loading self order

macocianradu:
Inconsistent data in the localDB or local storage can cause the pos_self_rder to fail before it can get initialized. When this happens the page just stays white.

Now when an error happens while initializing an error message will show up and give the user the option to clear all local data and refresh the page. This should be especially useful for users on mobile where it's a pain to refresh the local data manually.

Task-[5420544](https://www.odoo.com/odoo/project/1737/tasks/5420544)

---


davidmonnom (approved):
@robodoo r+

PR #246927, opened 2026-02-03
[FIX] res.partner write() issues caused by custom mixin inheritance order

quoc-pn:

When `res.partner` is inherited together with a custom mixin that adds additional values to `vals` inside the `write()` method, the behavior depends on the inheritance order. If the mixin is inherited after `res.partner`, the values added by the mixin are not included in the data collected by `res.partner.write()` for later synchronization, which can lead to runtime errors.

Current behavior before PR:

If the inheritance order is:

```python
_inherit = ['res.partner', 'custom.mixin']
```

the custom mixin’s `write()` method is executed after `res.partner.write()`.
As a result, any values added by the mixin are missing from `pre_values_list`, which is built by `res.partner.write()` and later accessed by `_fields_sync()`, causing failures when those fields are expected to be present.

This issue does not occur when the inheritance order is:

```python
_inherit = ['custom.mixin', 'res.partner']
```

because the mixin modifies `vals` before `res.partner.write()` is executed.

Steps to reproduce:
1. Create a mixin model with a new field, then add this field into `vals` inside the `write()` method so it is included in the update flow.
```python
class CustomMixin(models.AbstractModel):
    _name = "custom.mixin"
    
    custom_field = fields.Char()
    
    def write(self, vals):
        vals['custom_field'] = "foo"
        return super().write(vals)
```
2. Inherit the mixin in `res.partner`
```python
class ResPartner(models.Model):
    _name = "res.partner"
    _inherit = ["res.partner", "custom.mixin"]
```
3. Update a `res.partner` record, an error will be raised
```bash
  ...
  File "/opt/odoo/code/projects/odoo/odoo/orm/fields.py", line 1845, in __set__
    records.write({self.name: write_value})
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/odoo/code/projects/odoo/odoo/addons/base/models/res_partner.py", line 907, in write
    updated = {fname: fvalue for fname, fvalue in vals.items() if partner[fname] != pre_values[fname]}
                                                                                    ~~~~~~~~~~^^^^^^^
KeyError: 'custom_field'
```

Values added to vals by a custom mixin during `write()` are consistently available to `res.partner` internal synchronization logic, regardless of the inheritance order.
`res.partner.write()` and `_fields_sync()` should behave correctly even when mixins extend vals and are inherited after `res.partner`.

---

### DIFF
--- a/odoo/addons/base/models/res_partner.py
+++ b/odoo/addons/base/models/res_partner.py
@@ -904,7 +904,7 @@ class ResPartner(models.Model):
         for partner, pre_values in zip(self, pre_values_list, strict=True):
             if internal_users := partner.user_ids.filtered(lambda u: u._is_internal() and u != self.env.user):
                 internal_users.check_access('write')
-            updated = {fname: fvalue for fname, fvalue in vals.items() if partner[fname] != pre_values[fname]}
+            updated = {fname: fvalue for fname, fvalue in vals.items() if partner[fname] != pre_values.get(fname)}
             if updated:
                 partner._fields_sync(updated)
         return result


quoc-pn:
My code only tries to add more values to `vals` in the `write()` method, but the native logic related to `_field_sync()` in the `write()` method raises an error. My suggested changes are intended to preserve the current behavior in case any additional values are present in `vals`.

I don’t face this issue in the community code because there is no similar inheritance pattern, especially on the `res.partner` model. I added steps to reproduce the issue in the description.

odoo/addons/base/models/res_partner.py
```diff
@@ -904,7 +904,7 @@ def write(self, vals):
         for partner, pre_values in zip(self, pre_values_list, strict=True):
             if internal_users := partner.user_ids.filtered(lambda u: u._is_internal() and u != self.env.user):
                 internal_users.check_access('write')
-            updated = {fname: fvalue for fname, fvalue in vals.items() if partner[fname] != pre_values[fname]}
+            updated = {fname: fvalue for fname, fvalue in vals.items() if not pre_values.get(fname) or partner[fname] != pre_values[fname]}
```
kmagusiak:
```suggestion
            updated = {fname: fvalue for fname, fvalue in vals.items() if partner[fname] != pre_values.get(fname)}
```

kmagusiak (commented):
Just a small change, and I can merge this.

cuongnmtm:
> `inherit` controls the mro of super() function calls (just like normal python multiple inheritance). You expect that the call to the method of mixin mutates vals and then it is synced with these mutated values... You will still go though your mixin for synced records.
> 
> Your solution calls `_fields_sync` even if the previous value for the newly added value by the mixin would remain the same.
> 
> The code is a little bit spaghetti here... (multiple inheritance + syncing of a hierarchy don't help). What you suggest is a workaround and inversing inherit is a workaround as well. Since we don't have a problem in community repo, I suggest you do the workaround on your side - otherwise if you find a bug in community code caused by this, add a test.

@kmagusiak FYI This issue does not occur in the community code, but in the OCA mail_tracking module. In v18.0, the module introduces an inheritance on res.partner, which leads to this issue during the migration to v19.0.

https://github.com/OCA/mail/blob/18.0/mail_tracking/models/res_partner.py#L9

odoo/addons/base/models/res_partner.py
cuongnmtm: This change seems simpler. However, it does not return the correct value for `updated` variable, in my opinion, although this is not an issue. 

@quoc-pn  could you please apply this suggestion? Thanks.

quoc-pn: I have applied your suggestion.

kmagusiak (approved):
@robodoo r+

kmagusiak:
@robodoo r+

PR #246928, opened 2026-02-03
[FIX] account: translate product names on invoice reprint

YoussefM890:
**Problem:**
When a customer's language is changed and an invoice is reprinted, all invoice fields translate correctly except for product names and descriptions, which remain in the original language.

**Steps to reproduce:**
1. Create an invoice for a customer with English language
2. Post the invoice (product names appear in English)
3. Change the customer's language to French
4. Reprint/download the invoice PDF
5. Observe that product names still appear in English instead of French

**Current behavior:**
Product names remain in the original language when reprinting invoices after changing the customer's language.

**Expected behavior:**
Product names and descriptions should dynamically translate based on the customer's current language setting when the invoice is reprinted.

**Cause of the issue:**
The invoice report template uses `line.name`, which is a stored text field frozen at invoice creation time. This field contains the product name in whatever language was active when the line was created and does not update when the partner's language changes. When rendering the PDF, this frozen text is displayed regardless of the current language context.

**Fix:**
Use `product_id.display_name` instead of `line.name` for product-linked invoice lines. The `display_name` field is computed dynamically and respects the current language context during PDF generation. For lines without a product (manual entries, notes, sections), fall back to `line.name` to preserve existing behavior. This allows product names to translate on reprint while maintaining compatibility with non-product lines.

opw-5887170

### DIFF
--- a/addons/account/views/report_invoice.xml
+++ b/addons/account/views/report_invoice.xml
@@ -140,7 +140,22 @@
 
                                     <tr t-att-class="'bg-200 fw-bold o_line_section' if line.display_type == 'line_section' else 'fst-italic o_line_note' if line.display_type == 'line_note' else ''">
                                         <t t-if="line.display_type == 'product'" name="account_invoice_line_accountable">
-                                            <td name="account_invoice_line_name"><span t-if="line.name" t-field="line.name" t-options="{'widget': 'text'}">Bacon Burger</span></td>
+                                            <td name="account_invoice_line_name">
+                                                <t t-if="line.product_id">
+                                                    <span t-field="line.product_id.display_name">Product Name</span>
+                                                    <t t-if="line.product_id.description_sale and line.journal_id.type == 'sale'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_sale"/>
+                                                    </t>
+                                                    <t t-if="line.product_id.description_purchase and line.journal_id.type == 'purchase'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_purchase"/>
+                                                    </t>
+                                                </t>
+                                                <span t-else="" t-if="line.name" t-field="line.name"
+                                                      t-options="{'widget': 'text'}">Bacon Burger
+                                                </span>
+                                            </td>
                                             <td name="td_quantity" class="text-end">
                                                 <span t-field="line.quantity">3.00</span>
                                                 <span t-field="line.product_uom_id"  groups="uom.group_uom">units</span>


addons/account/views/report_invoice.xml
```diff
@@ -140,7 +140,22 @@
...
+                                                    <t t-if="line.product_id.description_sale and line.journal_id.type == 'sale'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_sale"/>
+                                                    </t>
+                                                    <t t-if="line.product_id.description_purchase and line.journal_id.type == 'purchase'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_purchase"/>
+                                                    </t>
+                                                </t>
+                                                <span t-else="" t-if="line.name" t-field="line.name"
```
alialfie, line 155:
```suggestion
                                                <span t-elif="line.name" t-field="line.name"
```

alialfie (changes requested):
Thanks for the fix! One small comment then should be good to go

addons/account/views/report_invoice.xml
YoussefM890: would'nt that risk breaking a customization ?

YoussefM890: @alialfie

william-andre: What kind of customization are talking about?
There was no `t-else` before, so there couldn't have been any customization based on it. Am I missing something?

addons/account/views/report_invoice.xml
```diff
@@ -140,7 +140,22 @@
...
+                                                <t t-if="line.product_id">
+                                                    <span t-field="line.product_id.display_name">Product Name</span>
+                                                    <t t-if="line.product_id.description_sale and line.journal_id.type == 'sale'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_sale"/>
+                                                    </t>
+                                                    <t t-if="line.product_id.description_purchase and line.journal_id.type == 'purchase'">
+                                                        <br/>
+                                                        <span t-field="line.product_id.description_purchase"/>
+                                                    </t>
```
william-andre, line 153: The commit message explains the translation issue, what does this have to do with `description_sale` and `description_purchase`?

addons/account/views/report_invoice.xml
william-andre, line 155: What if `line.name` is different than `line.product_id`?
For instance, the product is just a generic service (for taxes and reporting); without a price but the name is "emergency plumbing of the kitchen"?

addons/account/views/report_invoice.xml
alialfie: line.name already includes those description values. Check how _compute_name works it's basically the same as in the template here

addons/account/views/report_invoice.xml
YoussefM890: @william-andre My concern was about XPath inherits targeting the original <span t-if="line.name">

YoussefM890: @william-andre  You're right, and I don't think there's a straightforward way to distinguish whether line.name was auto-computed from the product or manually edited by the user. Do you have an idea on how to approach this?

PR #246930, opened 2026-02-03
[FIX] point_of_sale: update serial numbers requirements

qucol-odoo:
## Issue
In the Point of Sale, when selling a product tracked by unique serial number with a quantity greater than one, no warning are displayed to the displayed as long as one serial number is given.

## Steps to reproduce
1. Install Point of Sale (`point_of_sale`)
2. In Settings, enable *Lots & Serial Numbers*
3. Create a PoS product P
    - Tracked *By Unique Serial Number*
    - Update its quantity and create a few serial numbers
4. In the Point of Sale, select product P and set a serial number, then update the quantity to 2.
5. **The _Valid product lot_ icon next to the product is still green. We can also go to the payment screen without any warning, even though only one serial number was selected for two products.**

## Cause
The color and name of the icon displayed next to the serial tracked product is decided here:

https://github.com/odoo/odoo/blob/3d787e1b7c2ab471694849a04bdf2d3a5ff7e515/addons/point_of_sale/static/src/app/screens/product_screen/order_summary/order_summary.xml#L9-L14

The `PosOrderLine.hasValidProductLot` is defined here:

https://github.com/odoo/odoo/blob/3d787e1b7c2ab471694849a04bdf2d3a5ff7e515/addons/point_of_sale/static/src/app/models/pos_order_line.js#L333-L341

The very first condition of this method makes the icon green as long as one serial number is selected.

opw-5486210

### DIFF
--- a/addons/point_of_sale/static/src/app/models/pos_order_line.js
+++ b/addons/point_of_sale/static/src/app/models/pos_order_line.js
@@ -331,10 +331,6 @@ export class PosOrderline extends Base {
     }
 
     hasValidProductLot() {
-        if (this.pack_lot_ids.length > 0) {
-            return true;
-        }
-
         const valid_product_lot = this.getValidLots();
         const lotsRequired = this.product_id.tracking == "serial" ? Math.abs(this.qty) : 1;
         return lotsRequired === valid_product_lot.length;
--- a/addons/point_of_sale/static/tests/unit/models/pos_order_line.test.js
+++ b/addons/point_of_sale/static/tests/unit/models/pos_order_line.test.js
@@ -256,3 +256,46 @@ test("Test taxes after fiscal position", async () => {
     const lineValues = orderLine.prepareBaseLineForTaxesComputationExtraValues();
     expect(lineValues.tax_ids.length).toBe(0);
 });
+
+test("Test serial number requirements", async () => {
+    const store = await setupPosEnv();
+    const order = await getFilledOrder(store);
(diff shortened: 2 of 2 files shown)


addons/point_of_sale/static/src/app/models/pos_order_line.js
```diff
@@ -331,10 +331,6 @@ export class PosOrderline extends Base {
     }
 
     hasValidProductLot() {
-        if (this.pack_lot_ids.length > 0) {
```
robinengels, line 334: I'm not sure of what was the purpose of this line, but the rest of the method looks like it's enough to check if the line has enough lot/serial number.

addons/point_of_sale/tests/test_frontend.py
```diff
@@ -3285,6 +3285,16 @@ def test_different_currencies_correct_convertion(self):
         self.main_pos_config.with_user(self.pos_user).open_ui()
         self.start_tour("/pos/ui?config_id=%d" % self.main_pos_config.id, 'test_different_currencies_correct_convertion', login="pos_user")
 
+    def test_serial_numbers_requirements_icon_display(self):
```
robinengels: I think you should be albe to write a unit instead of a tour here. If possible we should always try to to a python or unit test instead of a tour.

robinengels (changes requested):

robinengels (approved):

loukwets (approved):

PR #246933, opened 2026-02-03
[IMP] *: is_downpayment from field to display_type

Megaaaaaa:
*: account, delivery, l10n_hu_edi, l10n_my_edi, l10n_sa_edi, pos_sale, purchase, resource, sale, sale_management, sale_project, sale_stock, sale_timesheet, website_sale

---

Description of the issue this commit addresses:

Currently the detection of whether a line (account move, sale order, purchase order) or a move is a downpayement relies on a `is_downpayment` field one the lines models. This creates complexity as overridable methods are put on account models so that when other module with downpayments are installed, they get overriden and the downpayment behavior is implemented. This means that the account module always calls those helper just in case of if one of those modules is installed.

---

Desired behavior after this commit is merged:

This commit migrates the `is_downpayment` field to a new value for the `display_type` field so that checking for downpayment doesn't rely on whether some other module is installed.

---

Enterprise PR: https://github.com/odoo/enterprise/pull/106273
task-5871102

---


addons/sale/models/sale_order_line.py
Feyensv, line 1: I believe there is one thing you don't account for in your change.
The "Down payments" section is actually a line with `display_type` `line_section` AND `is_downpayment=True` to distinguish it from the other sections.

Is it really necessary to do that change on SOlines as well ?

Megaaaaaa: I used to have a version of this with an entire line_section_downpayment value for the display_type field as a temporary way to handle that but I'm trying to adapt all behaviors correctly without it now.

While this is in draft, don't overthink it, I might indeed go for a change without touching SO lines.

PR #246934, opened 2026-02-03
[FIX] website: prevent error when opening Website Page Properties Base form view

reka-odoo:
When a user opens the ``website.page.properties.base.form.view`` view, a traceback appears.

Steps to reproduce the error:
- Install ``website`` module
- Using Open View, Open ``website.page.properties.base.form.view`` view

Traceback:
```py
AttributeError: 'NoneType' object has no attribute '_name'
```

https://github.com/odoo/odoo/blob/33089a1f8a735a08aed95c8ce783da15500b3a3f/addons/website/models/website_page_properties.py#L106-L107
When the view is opened, the ``_compute_is_published`` method is triggered.
since no ``target_model_id`` is set, ``target`` will be None,
which leads to the above traceback.

sentry-7237615365

---

### DIFF
--- a/addons/website/models/website_page_properties.py
+++ b/addons/website/models/website_page_properties.py
@@ -86,6 +86,9 @@ class WebsitePagePropertiesBase(models.TransientModel):
     def _compute_can_publish(self):
         for record in self:
             target = record.target_model_id
+            if not target:
+                record.can_publish = False
+                continue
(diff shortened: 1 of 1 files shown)


maan-odoo (approved):
This is an edge case, as generally no users will open this form directly from view. But yes, it is good to fix.

PR #246935, opened 2026-02-03
[FIX] project: incorrect view change after deleting task stage

aurb9:
# Steps to reproduce
1. Create a project
2. Create a stage
4. Remove the stage

# Current behavior
Instead of remaining in the project tasks view, it switches to the tasks view filtered with the current project. Additionally, it displays archived tasks because no filter is selected, thereby discarding original ones.

This also applies to stage deletion in other views (e.g., My Tasks), where the search filters are completely discarded.

# Expected behavior
The dialog should be closed, the stage should be deleted, and the original view should remain active. This is done through a soft-reload of the page, ensuring the original view is kept, together with original breadcrumbs, and the stage is visually disappearing.

task-5498274


xavierbol:
robodoo r-

PR #246936, opened 2026-02-03
[IMP] account: include all attachments in email notifications

jugurtha-gaci:
Before this PR, when importing an invoice or vendor bill, the email notification only included the first imported attachment. As a result, embedded files (e.g. a PDF inside an XML) were missing from the email.

This commit ensures that all imported attachments are included in the notification email, improving document visibility for users.

Additionally, the field "Send Copy To" will no longer be computed, It'll be manually filled by the user.

task-5895328

---


gawa-odoo (commented):
@robodoo r+

PR #246940, opened 2026-02-03
[FIX] point_of_sale: fix error while loading module

adip-odoo:
Currently, an error is produced when loading the pos module when a referenced UoM record (e.g., Pack of 6) has been deleted by the user.

**Steps to Reproduce:**
- Install Sales Module.
- Enable **"Units of Measure & Packaging"**.
- Delete the **"Pack of 6"** Units.
- Install POS.

**Error:**
```
ParseError:
while parsing /home/odoo/src/odoo/saas-19.1/addons/point_of_sale/data/point_of_sale_data.xml:38, somewhere inside <record id="uom.product_uom_pack_6" model="uom.uom">
            <field name="is_pos_groupable">True</field>
        </record>

Exception:
Cannot update missing record 'uom.product_uom_pack_6'
```
The POS module overrides `uom.product_uom_*` records. If the original UoM record has been deleted, the override fails because the record no longer exists.

sentry-7237815280

### DIFF
--- a/addons/point_of_sale/data/point_of_sale_data.xml
+++ b/addons/point_of_sale/data/point_of_sale_data.xml
@@ -35,11 +35,11 @@
             <field name="is_pos_groupable">True</field>
         </record>
 
-        <record id="uom.product_uom_pack_6" model="uom.uom">
+        <record id="uom.product_uom_pack_6" model="uom.uom" forcecreate="0">
             <field name="is_pos_groupable">True</field>
         </record>
 
-        <record id="uom.product_uom_dozen" model="uom.uom">
+        <record id="uom.product_uom_dozen" model="uom.uom" forcecreate="0">
             <field name="is_pos_groupable">True</field>
         </record>
 


addons/point_of_sale/data/point_of_sale_data.xml
```diff
@@ -31,15 +31,15 @@
             <field name="taxes_id" eval="[(5,)]"/>
         </record>
 
-        <record id="uom.product_uom_unit" model="uom.uom">
+        <record id="uom.product_uom_unit" model="uom.uom" forcecreate="0">
```
maan-odoo: No need `forcecreate="0"` for `uom.product_uom_unit` since user can't delete `product_uom_unit`.

adip-odoo: Indeed! 🙌 
Actually, I added it as a defensive safeguard (it becomes deletable in the future) 🙂

maan-odoo (approved):

adgu-odoo (commented):
robodoo r+

PR #246948, opened 2026-02-03
[FIX] l10n_eg_edi_eta: prevent concurrent double processing of invoices

aaha-odoo:
Before this commit:
Steps
1) Create an invoice in an Egyptian company
2) The customer clicks the Sign Invoice button twice at the same moment

=> The invoice is sent to Egyptian EDI twice

This happens because the action_post_sign_invoices() method does not prevent concurrent processing of the same invoice. The initial filter relies on l10n_eg_submission_number, which may still be unset when two processes handle the same invoice simultaneously, allowing both to pass the filter and send the invoice.

After this commit:
Prevents concurrent processing of the same invoices by locking the records during the sending logic, ensuring that only one process can handle an invoice at a time.

opw-5411280

### DIFF
--- a/addons/l10n_eg_edi_eta/models/account_move.py
+++ b/addons/l10n_eg_edi_eta/models/account_move.py
@@ -67,6 +67,7 @@ class AccountMove(models.Model):
         return super().button_draft()
 
     def action_post_sign_invoices(self):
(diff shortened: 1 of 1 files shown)


aaha-odoo:
Closing as it's not reproducible now

PR #246949, opened 2026-02-03
[FIX] account: update invoice payment reference with name

clom-odoo:
Issue:
Updating the invoice name should update the payment reference if the invoice isn't already sent.

Step to reproduce:
- Create an invoice,
- Post it,
- Draft it,
- Change name,
- Post it again,

Current behavior:
only the invoice name change

Expected behavior:
- invoice name change
- payment_reference update
- linked account_move_line labeled payment_term are updated

opw-5428471

---

### DIFF
--- a/addons/account/tests/test_account_move_out_invoice.py
+++ b/addons/account/tests/test_account_move_out_invoice.py
@@ -2440,6 +2440,25 @@ class TestAccountMoveOutInvoiceOnchanges(AccountTestInvoicingCommon):
             'amount_total': 1430.0,
         })
 
+    def test_out_invoice_name_change(self):
+        ''' Check the invoice.name, payment_term and invoice_line.name.'''
+        # Create an invoice.
+        move = self._create_invoice(invoice_date='2016-01-01', post=True)
+        payment_term_line = move.line_ids.filtered(lambda m: m.display_type == 'payment_term')
+
+        self.assertRecordValues(move, [{'payment_reference': 'INV/2016/00001'}])
+        self.assertRecordValues(payment_term_line, [{'name': 'INV/2016/00001'}])
+
+        move.button_draft()
+        move.name = 'INV/2016/00002'
+        move.action_post()
+
+        self.assertRecordValues(move, [{
+            'name': 'INV/2016/00002',
+            'payment_reference': 'INV/2016/00002'
+        }])
+        self.assertRecordValues(payment_term_line, [{'name': 'INV/2016/00002'}])
+
     def test_out_invoice_switch_out_refund_1(self):
         # Test creating an account_move with an out_invoice_type and switch it in an out_refund.
         move = self.env['account.move'].create({
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -818,7 +818,8 @@ class AccountMove(models.Model):
         for move in self.filtered(lambda m: (
             m.state == 'posted'
             and m.move_type == 'out_invoice'
-            and not m.payment_reference
+            and not m.inalterable_hash
+            and not (m.is_move_sent or m.is_being_sent)
         )):
             move.payment_reference = move._get_invoice_computed_reference()
         self._inverse_payment_reference()


guva-odoo (changes requested):
I think we are on a wrong direction with this fix, you're basically bypassing what we do with compute methods

- `_inverse_name` already adds `payment_reference` to the compute method, what's wrong here is in the `_compute_payment_reference` where we exclude moves that already have a `payment_reference`
- then in the `_compute_name` of the payment term line, we get the payment reference from the move that has never been changed

I think we can check in the `_compute_payment_reference` if we can include the move that already have a payment reference if `move.inalterable_hash` is False or something like that

clom-odoo:
Payment reference is an editable field. We may not want to update it if it was manually updated.
Clean way to do it should use `_compute_payment_reference`. However, this method doesn't have access
to the previous name and can't assert whether the field was computed or not.

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+            'invoice_date': fields.Date.from_string('2016-01-01'),
+            'invoice_line_ids': [
+                Command.create({'product_id': self.product_line_vals_1['product_id']}),
+                Command.create({'product_id': self.product_line_vals_2['product_id']}),
+            ],
+        }])
+
+        # Invoice of journal based on odoo invoice
+        # post the invoice to compute the payment reference and payment_term label
+        move1.action_post()
```
guva-odoo:
```suggestion
        move = self._create_invoice(invoice_date='2016-01-01', post=True)
        payment_term_line = move1.line_ids.filtered(lambda m: m.display_type == 'payment_term')
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        self.assertRecordValues(move1, [{'payment_reference': 'INV/2016/00001'}])
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
```
guva-odoo:
```suggestion
            payment_term_line,
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+            [{'name': 'INV/2016/00001'}])
+
+        move1.button_draft()
+        with Form(move1) as move_form:
```
guva-odoo:
```suggestion
        move.button_draft()
        with Form(move) as move_form:
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+            move_form.name = 'INV/2016/00003'
+        move1 = move_form.save()
+        move1.action_post()
+
+        self.assertRecordValues(move1, [{'payment_reference': 'INV/2016/00003'}])
```
guva-odoo:
```suggestion
        self.assertRecordValues(move, [{'payment_reference': 'INV/2016/00003'}])
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
```
guva-odoo:
```suggestion
            payment_term_line
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        move1.action_post()
+        self.assertRecordValues(move1, [{'payment_reference': 'INV/2016/00001'}])
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
+            [{'name': 'INV/2016/00001'}])
+
+        move1.button_draft()
+        with Form(move1) as move_form:
+            move_form.name = 'INV/2016/00003'
+        move1 = move_form.save()
```
guva-odoo: Form is saved when exiting the context manager
```suggestion
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
+            [{'name': 'INV/2016/00003'}])
+
+        move1.button_draft()
+        with Form(move1) as move_form:
+            move_form.payment_reference = 'MyCustomer - 00001'
+            move1 = move_form.save()
+            move_form.name = 'INV/2016/00002'
+            move1 = move_form.save()
```
guva-odoo:
```suggestion
        with Form(move) as move_form:
            move_form.payment_reference = 'MyCustomer - 00001'
        with Form(move) as move_form:
            move_form.name = 'INV/2016/00002'
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        move1.action_post()
+
+        self.assertRecordValues(move1, [{
+            'name': 'INV/2016/00002',
+            'payment_reference': 'MyCustomer - 00001'}])
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
```
guva-odoo:
```suggestion
            payment_term_line
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+            move_form.name = 'INV/2016/00003'
+        move1 = move_form.save()
+        move1.action_post()
+
+        self.assertRecordValues(move1, [{'payment_reference': 'INV/2016/00003'}])
+        self.assertRecordValues(
+            move1.line_ids.filtered(lambda m: m.display_type == 'payment_term'),
+            [{'name': 'INV/2016/00003'}])
+
+        move1.button_draft()
```
guva-odoo:
```suggestion
        move.button_draft()
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2439,6 +2439,53 @@ def test_out_invoice_post_2(self):
...
+        with Form(move1) as move_form:
+            move_form.payment_reference = 'MyCustomer - 00001'
+            move1 = move_form.save()
+            move_form.name = 'INV/2016/00002'
+            move1 = move_form.save()
+        move1.action_post()
+
+        self.assertRecordValues(move1, [{
```
guva-odoo:
```suggestion
        self.assertRecordValues(move, [{
```

guva-odoo (changes requested):
Just some little change in the test.

About the fix, I don't have a strong opinion about what to do, my preference would go to either fix with the compute either doing nothing, but the proposed onchange also makes sense so I would be glad to have the opinion of another reviewer

guva-odoo (approved):

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2440,6 +2440,26 @@ def test_out_invoice_post_2(self):
...
         })
 
+    def test_out_invoice_name_change(self):
+        ''' Check the invoice.name, payment_term and invoice_line.name.'''
+        # Create an invoice.
+        move = self._create_invoice(invoice_date='2016-01-01', post=True)
+        payment_term_line = move.line_ids.filtered(lambda m: m.display_type == 'payment_term')
+
+        self.assertRecordValues(move, [{'payment_reference': 'INV/2016/00001'}])
+        self.assertRecordValues(payment_term_line,[{'name': 'INV/2016/00001'}])
```
william-andre: CI/style 😆 
```suggestion
        self.assertRecordValues(payment_term_line, [{'name': 'INV/2016/00001'}])
```

addons/account/tests/test_account_move_out_invoice.py
```diff
@@ -2440,6 +2440,26 @@ def test_out_invoice_post_2(self):
...
+
+        move.button_draft()
+        move.payment_reference = 'MyCustomer - 00001'
```
william-andre: This is not a behavior that we *want* to have, so no need to test it.
We just don't care about such a corner case for various reasons:
* if it is computed, the users are most likely not setting it manually afterwards
* it doesn't happen too often that the name is changed manually either
* even if they changed it and it is recomputed, it is not the worst thing ever. The payment matching will simply be a little more painful because not automated, but the tools should still match based on the partner and amount; and possibly even the old reference

```suggestion
```

PR #246950, opened 2026-02-03
[ADD] l10n_tr_nilvera_einvoice: support commercial invoices and bills

adns-odoo:
# Description of the issue/feature this PR addresses
Odoo's Nilvera integration (l10n_tr_nilvera_einvoice) currently supports only basic e-invoice scenarios.
Commercial invoices and commercial bills (TICARIFATURA), which are widely used in Turkish e-invoicing
workflows, were not supported. This PR adds full support for creating, sending, and responding to
commercial invoices and bills in alignment with Nilvera's API specifications.

# Current behavior before PR
- Only standard e-invoice profiles were supported.
- Commercial invoices and bills (ProfileID: TICARIFATURA) could not be created or processed.
- Incoming responses for commercial invoices were not handled.
- Sending responses for commercial bills was not possible.

# Desired behavior after PR is merged
- Commercial invoices and bills are fully supported via Nilvera.
- Odoo can create and process invoices/bills with:
    - ProfileID: TICARIFATURA
    - Move types: sale and purchase
- Both customer invoices and vendor bills are supported.
- Invoice responses can be retrieved from Nilvera and sent back for commercial bills.
- UI banner on the invoice form shows the current response status:
    - Pending: awaiting response from the counterpart
    - Accepted by the counterpart
    - Automatically accepted by Nilvera
    - Rejected (reason posted to chatter, invoice cancelled, updated PDF fetched)
- Commercial bills can be accepted or rejected directly from Odoo via a confirmation wizard.
- A scheduled job (12-hour interval) periodically syncs response status from Nilvera,
  processing invoices in order of priority to avoid repeatedly retrying failing records.

# Supported scenarios
- Profile ID
    - TICARIFATURA
- Invoice Type Codes
    - SATIS
    - TEVKIFAT
    - ISTISNA
    - IHRACKAYITLI
- Move types
    - Customer invoices
    - Vendor bills

# Behavior notes
- If InvoiceTypeCode is not one of the supported values, it falls back to SATIS to ensure
  bill response handling remains functional.
- Rejected commercial invoices can be reset to draft for correction and resubmission.
- The Nilvera PDF is automatically attached to outgoing emails for all terminal statuses
  (approved, auto-approved, rejected).

# Unsupported scenarios
- Credit notes (out of scope for this PR)

---
task-4741223
I confirm I have signed the CLA and read the PR guidelines at [www.odoo.com/submit-pr](http://www.odoo.com/submit-pr)

### DIFF
--- a/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
@@ -997,8 +997,24 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         # EXTENDS account.edi.xml.ubl_20
         logs = super()._import_fill_invoice(invoice, tree, qty_factor)
 
-        # ==== Nilvera UUID ====
-        if uuid_node := tree.findtext('./{*}UUID'):
-            invoice.l10n_tr_nilvera_uuid = uuid_node
+        scenario = self._find_value(".//cbc:ProfileID", tree)
+        if scenario not in {key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_scenario'].selection}:
+            scenario = False
+        invoice_type = self._find_value(".//cbc:InvoiceTypeCode", tree)
+        if invoice_type not in {key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_type'].selection}:
+            invoice_type = False
+
+        invoice.write({
+            'l10n_tr_nilvera_uuid': self._find_value('./cbc:UUID', tree),
+            'l10n_tr_gib_invoice_scenario': scenario,
+            'l10n_tr_gib_invoice_type': invoice_type,
+        })
+
+        if scenario == 'TICARIFATURA' and invoice.move_type in {'in_invoice', 'out_invoice'}:
+            try:
+                # Don't want to block the import in case status retrieval fails
+                invoice.l10n_tr_action_fetch_ticarifatura_response()
+            except UserError as e:
+                logs.append(self.env._("Failed to fetch TICARIFATURA response: %s", str(e)))
 
         return logs
--- a/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
@@ -983,8 +983,24 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         # EXTENDS account.edi.xml.ubl_20
         logs = super()._import_fill_invoice(invoice, tree, qty_factor)
 
-        # ==== Nilvera UUID ====
-        if uuid_node := tree.findtext('./{*}UUID'):
-            invoice.l10n_tr_nilvera_uuid = uuid_node
+        scenario = self._find_value(".//cbc:ProfileID", tree)
+        if scenario not in {key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_scenario'].selection}:
+            scenario = False
+        invoice_type = self._find_value(".//cbc:InvoiceTypeCode", tree)
+        if invoice_type not in {key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_type'].selection}:
+            invoice_type = False
+
+        invoice.write({
+            'l10n_tr_nilvera_uuid': self._find_value('./cbc:UUID', tree),
+            'l10n_tr_gib_invoice_scenario': scenario,
+            'l10n_tr_gib_invoice_type': invoice_type,
+        })
+
+        if scenario == 'TICARIFATURA' and invoice.move_type in {'in_invoice', 'out_invoice'}:
+            try:
+                # Don't want to block the import in case status retrieval fails
+                invoice.l10n_tr_action_fetch_ticarifatura_response()
+            except UserError as e:
+                logs.append(self.env._("Failed to fetch TICARIFATURA response: %s", str(e)))
 
         return logs
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move.py
@@ -1,11 +1,18 @@
 import base64
+import logging
 import uuid
-from markupsafe import Markup
+from datetime import timedelta
+from json import JSONDecodeError
 from urllib.parse import quote, urlencode, urlparse
 
-from odoo import Command, _, api, fields, models
-from odoo.exceptions import UserError
+from markupsafe import Markup
+
+from odoo import _, Command, api, fields, models
+from odoo.exceptions import UserError, ValidationError
+
 from odoo.addons.l10n_tr_nilvera.lib.nilvera_client import _get_nilvera_client
+
+_logger = logging.getLogger(__name__)
 
 MOVE_TYPE_CATEGORY_MAP = {
     "out_invoice": {
@@ -22,6 +29,21 @@ CATEGORY_MOVE_TYPE_MAP = {
     "sale": "out_invoice",
     "purchase": "in_invoice",
 }
+
+TICARIFATURA_ANSWER_TO_FIELD_VALUE_MAP = {
+    "approved": "commercial_approved",
+    "rejected": "commercial_rejected",
+    "documentAnsweredAutomatically": "commercial_answered_automatically",
+}
+
+UNSYNCED_COMMERCIAL_MOVE_DOMAIN = [
+    ("move_type", "in", ["out_invoice", "in_invoice"]),
+    ("l10n_tr_gib_invoice_scenario", "=", "TICARIFATURA"),
+    ("l10n_tr_nilvera_send_status", "=", "succeed"),
+    ("partner_id", "!=", False),  # Partner's status is needed to determine endpoint when fetching response
+    ("company_id.country_code", "=", "TR"),
+    ("company_id.l10n_tr_nilvera_api_key", "!=", False),
+]
 
 
 class AccountMove(models.Model):
@@ -44,6 +66,9 @@ class AccountMove(models.Model):
             ('succeed', "Successful"),
             ('waiting', "Waiting"),
             ('unknown', "Unknown"),
+            ('commercial_approved', "Approved (Commercial)"),
+            ('commercial_answered_automatically', "Approved Automatically (Commercial)"),
+            ('commercial_rejected', "Rejected (Commercial)"),
         ],
         string="Nilvera Status",
         readonly=True,
@@ -62,6 +87,7 @@ class AccountMove(models.Model):
         selection=[
             ('TEMELFATURA', "Basic"),
             ('KAMU', "Public Sector"),
+            ('TICARIFATURA', "Commercial"),
         ],
         default='TEMELFATURA',
         string="Invoice Scenario",
@@ -134,6 +160,9 @@ class AccountMove(models.Model):
         string="Partner Nilvera Status",
         related='partner_id.l10n_tr_nilvera_customer_status',
         help="Shows the Nilvera status of the customer. ",
+    )
+    l10n_tr_ticarifatura_last_checked_at = fields.Datetime(
+        string="Commercial Status Check Date", copy=False,
     )
     l10n_tr_nilvera_pdf_file = fields.Binary(
         attachment=True,
@@ -224,11 +253,11 @@ class AccountMove(models.Model):
 
     def button_draft(self):
         # EXTENDS account
-        for move in self.filtered(lambda move: move.l10n_tr_nilvera_uuid and move.move_type == 'out_invoice'):
+        for move in self.filtered(lambda move: move.l10n_tr_nilvera_uuid and move.move_type in {'out_invoice', 'in_invoice'}):
             if move.l10n_tr_nilvera_send_status == 'error':
                 move.message_post(body=_("To preserve accounting integrity and comply with legal requirements, invoices cannot be reused once an error occurs. Please create a new invoice to continue."))
-            elif move.l10n_tr_nilvera_send_status != 'not_sent':
-                raise UserError(_("You cannot reset to draft an entry that has been sent to Nilvera."))
+            elif move.l10n_tr_nilvera_send_status not in {"not_sent", "commercial_rejected"}:
+                raise UserError(_("You cannot reset to draft an entry that has been sent/received from Nilvera."))
         super().button_draft()
 
     def _l10n_tr_nilvera_einvoice_check_invalid_invoice_reference(self):
@@ -353,7 +382,6 @@ class AccountMove(models.Model):
 
                     nilvera_status = response.get('InvoiceStatus', {}).get('Code') or response.get('StatusCode')
                     if nilvera_status in dict(invoice._fields['l10n_tr_nilvera_send_status'].selection):
-                        invoice.l10n_tr_nilvera_send_status = nilvera_status
                         if nilvera_status == 'error':
                             invoice.message_post(
                                 body=Markup(
@@ -364,6 +392,13 @@ class AccountMove(models.Model):
                                     response.get('InvoiceStatus', {}).get('DetailDescription') or response.get('ReportStatus'),
                                 )
                             )
+                        elif invoice.l10n_tr_gib_invoice_scenario == "TICARIFATURA" and invoice.move_type in {'out_invoice', 'in_invoice'} and nilvera_status == 'succeed':
+                            if response['Answer'] and response['Answer'].get('AnswerCode') in {'approved', 'rejected', 'documentAnsweredAutomatically'}:
+                                invoice.l10n_tr_nilvera_send_status = TICARIFATURA_ANSWER_TO_FIELD_VALUE_MAP[response['Answer']['AnswerCode']]
+                                if response['Answer']['AnswerCode'] == 'rejected':
+                                    invoice._l10n_tr_action_process_rejected_ticarifatura(response['Answer'].get('Description', ''), client)
+                        else:
+                            invoice.l10n_tr_nilvera_send_status = nilvera_status
                     else:
                         invoice.message_post(body=_("The invoice status couldn't be retrieved from Nilvera."))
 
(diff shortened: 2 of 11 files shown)


adns-odoo:
Hi @FlorianGilbert 
I kept on resolving the merge conflicts for this PR again again :) 
I can solve it again, is there any we way can merge it on 19.2

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,232 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
             )
 
         return super()._reverse_moves(default_values_list, cancel=cancel)
+
+    def l10n_tr_handle_409_error_for_send_answer(self, response):
+        self.ensure_one()
+        for error in response["Errors"]:
+            if error.get("Code") in [1003, 1007, 1008, 1011]:
```
```suggestion
            if error.get("Code") in {1003, 1007, 1008, 1011}:
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,232 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                        ),
+                        "type": "warning",
+                        # Reload the page to show the updated response.
+                        "next": {
+                            "type": "ir.actions.act_window",
+                            "res_model": "account.move",
+                            "res_id": self.id,
+                            "view_mode": "form",
+                            "views": [[False, "form"]],
+                            "target": "current",
```
malb-odoo: the is an action reload instead of that 😄  {"type": "ir.actions.client", "tag": "reload"}

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,232 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+    def l10n_tr_handle_409_error_for_send_answer(self, response):
+        self.ensure_one()
+        for error in response["Errors"]:
+            if error.get("Code") in [1003, 1007, 1008, 1011]:
+                self.l10n_tr_action_fetch_ticafatura_response()
+                return {
+                    "type": "ir.actions.client",
+                    "tag": "display_notification",
+                    "params": {
+                        "message": _(
```
malb-odoo: use self.env._ for translation

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,232 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+        """
+        Send the Ticarifatura response to Nilvera.
+        Applicable only for Commercial bills.
+
+        :param str answer_code: 'approved' or 'rejected'
+        :param str rejection_note: Note for rejection, if applicable.
+        """
+        self.ensure_one()
+        if self.move_type != 'in_invoice' or self.l10n_tr_gib_invoice_scenario != 'TICARIFATURA':
+            raise UserError(_("This action is only available for Commercial bills."))
```
```suggestion
            raise UserError(_("This action is only available for commercial bills."))
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,232 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+    def l10n_tr_action_fetch_ticafatura_response(self):
+        """
+        Fetch the Ticarifatura response status from Nilvera for the given UUID.
+        Applicable only for Commercial Invoice.
+
+        :param str uuid: The NILVERA UUID of the Commercial Invoice.
+        """
+        self.ensure_one()
+
+        if self.move_type not in ['out_invoice', 'in_invoice'] or self.l10n_tr_gib_invoice_scenario != 'TICARIFATURA':
```
```suggestion
        if self.move_type not in {'out_invoice', 'in_invoice'} or self.l10n_tr_gib_invoice_scenario != 'TICARIFATURA':
```

addons/l10n_tr_nilvera_einvoice/wizards/l10n_tr_nilvera_einvoice_ticafatura_response_wizard.py
```diff
@@ -0,0 +1,22 @@
...
+class AccountMoveTicafaturaResponseWizard(models.TransientModel):
+    _name = 'l10n_tr.ticafatura.response.wizard'
+    _description = 'Response Wizard for Bills of type Ticafatura E-Invoice'
+
+    move_id = fields.Many2one('account.move', required=True)
+    response_code = fields.Selection([
+        ('approved', 'Approve'),
+        ('rejected', 'Reject'),
+    ], string='Response Code', required=True)
+    response_note = fields.Text(string='Response Note')
```
malb-odoo:
```suggestion
    response_code = fields.Selection([
        ('approved', "Approve"),
        ('rejected', "Reject"),
    ], string="Response Code", required=True)
    response_note = fields.Text(string="Response Note")
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -985,12 +985,25 @@ def _import_retrieve_partner_vals(self, tree, role):
...
         return partner_vals
 
-    def _import_fill_invoice_form(self, invoice, tree, qty_factor):
+    def _import_fill_invoice(self, invoice, tree, qty_factor):
         # EXTENDS account.edi.xml.ubl_20
-        logs = super()._import_fill_invoice_form(invoice, tree, qty_factor)
+        logs = super()._import_fill_invoice(invoice, tree, qty_factor)
+        if self._find_value(".//cbc:CustomizationID", tree) == "TR1.2":
+            scenario = self._find_value(".//cbc:ProfileID", tree)
+            if scenario not in [k for k, v in self.env['account.move']._fields['l10n_tr_gib_invoice_scenario'].selection]:
```
malb-odoo:
```suggestion
            if scenario not in [key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_scenario'].selection]:
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -985,12 +985,25 @@ def _import_retrieve_partner_vals(self, tree, role):
...
+                scenario = False
+            invoice_type = self._find_value(".//cbc:InvoiceTypeCode", tree)
+            if invoice_type not in [k for k, v in self.env['account.move']._fields['l10n_tr_gib_invoice_type'].selection]:
```
malb-odoo:
```suggestion
            if invoice_type not in [key for key, _ in self.env['account.move']._fields['l10n_tr_gib_invoice_type'].selection]:
```

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -985,12 +985,25 @@ def _import_retrieve_partner_vals(self, tree, role):
         })
         return partner_vals
 
-    def _import_fill_invoice_form(self, invoice, tree, qty_factor):
```
malb-odoo: for the next review this function doesn't exist anymore

malb-odoo (commented):
Just passing through with some nitpicks

addons/l10n_tr_nilvera_einvoice/wizards/l10n_tr_nilvera_einvoice_ticafatura_response_wizard.py
adns-odoo: Hi @malb-odoo 
I have updated to double quote the the values of the Key-Value here
But curious to know why this unless we are using a single-quote(') inside the string. 

Here is a field from account.move, initially I thought this is a convention for selection field
```
state = fields.Selection(
        selection=[
            ('draft', 'Draft'),
            ('posted', 'Posted'),
            ('cancel', 'Cancelled'),
        ],
        string='Status',
        required=True,
        readonly=True,
        copy=False,
        tracking=True,
        default='draft',
    )
```

malb-odoo: It's not really a convention but more a rule of thumb that i like to apply so won't block if you don't want to do it, double quote for thing the user will see and single otherwise :smile:

addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
adns-odoo: Hi @malb-odoo 
I did not get what you meant

malb-odoo: It's a message for the next reviewer, since it's flg that was pinged i don't want to steal the review ahah

addons/l10n_tr_nilvera_einvoice/models/account_move.py
FlorianGilbert: Ideally, if the list/set is fixed, define it before the loop

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                    raise UserError(self.env._("An error occurred. Try again later."))
+                return self._l10n_tr_handle_409_error_for_send_answer(decoded_response)
+            return True
+
+    def l10n_tr_action_fetch_ticafatura_response(self):
+        """
+        Fetch the Ticarifatura response status from Nilvera for the given UUID.
+        Applicable only for Commercial Invoice.
+
+        :param str uuid: The NILVERA UUID of the Commercial Invoice.
```
FlorianGilbert: The param doesn't exist.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                except JSONDecodeError:
+                    _logger.exception("Invalid JSON response: %s", response.text)
+                    raise UserError(self.env._("An error occurred. Try again later."))
+                return self._l10n_tr_handle_409_error_for_send_answer(decoded_response)
+            return True
+
+    def l10n_tr_action_fetch_ticafatura_response(self):
+        """
+        Fetch the Ticarifatura response status from Nilvera for the given UUID.
+        Applicable only for Commercial Invoice.
```
FlorianGilbert: The docstring is not true then.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -135,6 +142,27 @@ class AccountMove(models.Model):
...
+            ('approved', "Approved"),
+            ('documentAnsweredAutomatically', "Approved Automatically"),
+            ('rejected', "Rejected"),
+        ],
+        string="Commercial Response",
+        readonly=True,
+        copy=False,
+        tracking=True,
+    )
+    l10n_tr_ticarifatura_response_note = fields.Text(
```
FlorianGilbert: This field couldn't be in the chatter directly?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -135,6 +142,27 @@ class AccountMove(models.Model):
...
+            ('pending', "Waiting Response"),
+            ('approved', "Approved"),
+            ('documentAnsweredAutomatically', "Approved Automatically"),
+            ('rejected', "Rejected"),
+        ],
+        string="Commercial Response",
+        readonly=True,
+        copy=False,
+        tracking=True,
+    )
```
FlorianGilbert: I don't get why we need another field to handle other status? Isn't it possible to have something generic with the `l10n_tr_nilvera_send_status`?
My concern is about how many fields we are adding on a big model like `account_move`. We should keep it minimal as possible.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+
+        :param tuple ratio: A tuple of two integers representing the ratio of never checked and previously checked moves.
+        :param int batch_size: The number of invoices to process in each batch.
+        """
+        _logger.info("Nilvera TICARIFATURA response sync started.")
+        BASE_DOMAIN = [
+                    ('move_type', 'in', ['out_invoice', 'in_invoice']),
+                    ('l10n_tr_gib_invoice_scenario', '=', 'TICARIFATURA'),
+                    ('l10n_tr_ticarifatura_status', '=', 'pending'),
+                ]
```
FlorianGilbert: Indentation is wrong + Please move it outside the function scope. It doesn't make sense to have a constant variable in the function 😄

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+        invoices_count = dict(
```
FlorianGilbert: The variable name could be more meaningful 😅 Based on what?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            self.env['account.move']._read_group(
+                domain=BASE_DOMAIN,
+                groupby=['l10n_tr_ticarifatura_status_check_priority'],
+                aggregates=['id:count'],
+            ),
+        )
+        never_checked_moves_count = invoices_count.get(0, 0)
+        checked_moves_count = sum(count for priority, count in invoices_count.items() if priority > 0)
```
FlorianGilbert:
```suggestion
        invoices_count_per_priority = dict(
            self.env['account.move']._read_group(
                domain=BASE_DOMAIN,
                groupby=['l10n_tr_ticarifatura_status_check_priority'],
                aggregates=['id:count'],
            ),
        )
        never_checked_moves_count = invoices_count_per_priority.pop(0, 0)
        checked_moves_count = sum(invoices_count_per_priority.values())
```
What about this?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+
+        if never_checked_moves_count + checked_moves_count == 0:
+            _logger.info("No Commercial Invoices found for TICARIFATURA response sync.")
+            return
+
+        if ratio and sum(ratio) != 100:
+            _logger.error("Invalid ratio value: %s. The sum of the two values must be 100.", ratio)
```
FlorianGilbert: How the user could know this issue without reading the logs?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+        if not ratio:
+            # Calculation the ratio to process data based on the count of checked and never checked moves.
+            ratio = [0, 0]
+            total = never_checked_moves_count + checked_moves_count
+            ratio[0] = int((never_checked_moves_count * 100) / total)
+            ratio[1] = int((checked_moves_count * 100) / total)
+
+            diff = 100 - sum(ratio)
+            if diff != 0:
+                ratio[0] += diff
```
FlorianGilbert: Correct me if I am wrong but what about this:

```suggestion
        if not ratio:
            # Calculation the ratio to process data based on the count of checked and never checked moves.
            ratio = [0, 0]
            total = never_checked_moves_count + checked_moves_count
            ratio[0] = math.ceil(never_checked_moves_count * 100 / total)
            ratio[1] = 100 - ratio[0]
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+
+        # Calculate the total records to process of each type
+        # Fetch the record and process
+        # Avoiding fetching all records at once to prevent memory issues
+        while never_checked_moves_count or checked_moves_count:
+            # get invoices to check based on ratio and batch size
+            never_checked_move_to_process_count = min(((batch_size * ratio[0]) // 100), never_checked_moves_count)
+            checked_move_to_process_count = min(((batch_size * ratio[1]) // 100), checked_moves_count)
```
FlorianGilbert: Same comment, no need to compute both, we can compute the first one and deduct the second one from the first one.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            if never_checked_move_to_process_count + checked_move_to_process_count < batch_size:
+                remaining = batch_size - (never_checked_move_to_process_count + checked_move_to_process_count)
+
+                max_take_from_checked = min(remaining, checked_moves_count - checked_move_to_process_count)
+                checked_move_to_process_count += max_take_from_checked
+                remaining -= max_take_from_checked
+
+                # if still remaining, take from recently added moves
+                max_take_from_unchecked = min(remaining, never_checked_moves_count - never_checked_move_to_process_count)
+                never_checked_move_to_process_count += max_take_from_unchecked
```
FlorianGilbert: It's really difficult for nothing. Please improve it. I won't rewrite every part of the function 😅

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+
+            # Fetch and process new records first
+            if never_checked_move_to_process_count:
+                records = self.env['account.move'].search(
+                    domain=BASE_DOMAIN + [('l10n_tr_ticarifatura_status_check_priority', '=', 0)],
```
FlorianGilbert: Nice place to use [Domain class](https://github.com/odoo/odoo/blob/99859c013363d8e96e212b52c1734b7dc1eafdec/odoo/orm/domains.py#L191)

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                    order="create_date desc",
+                    limit=never_checked_move_to_process_count,
+                )
+                for record in records:
+                    record.l10n_tr_action_fetch_ticafatura_response()
+                    record.write({"l10n_tr_ticarifatura_status_check_priority": record.l10n_tr_ticarifatura_status_check_priority + 1})
```
FlorianGilbert: I am starting to understand why we need this field and I am not convinced, it's answering any legal obligation or it's just to "optimize" things?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                never_checked_move_to_process_count += max_take_from_unchecked
+
+            # Fetch and process new records first
+            if never_checked_move_to_process_count:
+                records = self.env['account.move'].search(
+                    domain=BASE_DOMAIN + [('l10n_tr_ticarifatura_status_check_priority', '=', 0)],
+                    order="create_date desc",
+                    limit=never_checked_move_to_process_count,
+                )
+                for record in records:
```
FlorianGilbert: Nit, but I personally prefer when we name things by their real name. In this case: move 😄

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            },
+        }
+
+    def _l10n_tr_nilvera_company_sync_ticarifatura_response(self, ratio=None, batch_size=20):
+        for company in self.env.companies:
+            if company.country_code != "TR" or not company.l10n_tr_nilvera_api_key:
+                continue
+            self.with_company(company)._cron_l10n_tr_nilvera_sync_ticarifatura_response(ratio, batch_size)
+
+    def _cron_l10n_tr_nilvera_sync_ticarifatura_response(self, ratio=None, batch_size=20):
```
FlorianGilbert: Genuine question, why don't we just get the number of moves according to the batch size and ordered by the priority? So if we have a batch size of 100. Let's imagine that we have 97 not checked and 15 checked, we'll have the 97 unchecked and 3 checked. 
Why do we have to complexify more than that?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            'target': 'new',
+            'context': {
+                'default_move_id': self.id,
+                'default_response_code': 'approved',
+            },
+        }
+
+    def l10n_tr_action_reject_ticarifatura(self):
+        self.ensure_one()
+        return {
```
FlorianGilbert: Use `_get_records_action`

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                self.l10n_tr_ticarifatura_status = response["Answer"]["AnswerCode"]
+                if response["Answer"].get("AnswerCode") == "rejected":
+                    self.l10n_tr_ticarifatura_response_note = response["Answer"].get("Description")
+                    self.l10n_tr_nilvera_get_pdf()
+                    self.button_draft()
+                    self.button_cancel()
+
+    def l10n_tr_action_approve_ticarifatura(self):
+        self.ensure_one()
+        return {
```
FlorianGilbert: Same here

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                    "type": "ir.actions.client",
+                    "tag": "display_notification",
+                    "params": {
+                        "message": self.env._(
+                            "Nilvera has already received a response for this invoice."
+                            "\nThe latest response has been fetched and updated on the invoice.",
+                        ),
+                        "type": "warning",
+                        # Reload the page to show the updated response.
+                        "next": {"type": "ir.actions.client", "tag": "reload"},
```
FlorianGilbert: Isn't `soft_reload` sufficient?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -622,3 +656,225 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
             )
 
         return super()._reverse_moves(default_values_list, cancel=cancel)
+
+    def _l10n_tr_handle_409_error_for_send_answer(self, response):
+        self.ensure_one()
+        for error in response["Errors"]:
+            if error.get("Code") in {1003, 1007, 1008, 1011}:
```
FlorianGilbert: As said, move the set outside of the loop.

addons/l10n_tr_nilvera_einvoice/wizards/l10n_tr_nilvera_einvoice_ticafatura_response_wizard.py
```diff
@@ -0,0 +1,22 @@
...
+        ('rejected', "Reject"),
+    ], string="Response Code", required=True)
+    response_note = fields.Text(string="Response Note")
+
+    def action_proceed(self):
+        self.ensure_one()
+        if self.response_code == 'approved':
+            self.move_id.l10n_tr_action_send_ticarifatura_response('approved')
+        else:
+            self.move_id.l10n_tr_action_send_ticarifatura_response('rejected', self.response_note)
```
FlorianGilbert: Couldn't we simplify? 🤓

addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
```diff
@@ -62,6 +101,20 @@
...
                        readonly="state in ['cancel', 'posted']"/>
+                <div class="o_td_label"
+                     invisible="l10n_tr_gib_invoice_scenario != 'TICARIFATURA' or l10n_tr_nilvera_send_status != 'succeed'">
+                    <label for="l10n_tr_ticarifatura_status"/>
+                </div>
+                <div class="d-flex"
+                     invisible="l10n_tr_gib_invoice_scenario != 'TICARIFATURA' or l10n_tr_nilvera_send_status != 'succeed'">
+                    <field name="l10n_tr_ticarifatura_status" invisible="l10n_tr_gib_invoice_scenario != 'TICARIFATURA' or l10n_tr_nilvera_send_status != 'succeed'"/>
+                    <button name="l10n_tr_action_fetch_ticafatura_response" type="object"
+                            class="btn btn-link text-info ms-auto d-flex align-items-center gap-1"
```
FlorianGilbert: Why are we using a `text-info` class with a `btn btn-link` one? Especially when you are using that class in both children elements.

FlorianGilbert (changes requested):
The code looks like it's working but please: simplify it, apply the KISS principle and challenge code you wrote. Thank you!

addons/l10n_tr_nilvera_einvoice/models/account_move.py
adns-odoo: The field **l10n_tr_nilvera_send_status** and **l10n_tr_ticarifatura_status** represents 2 different things. 

l10n_tr_nilvera_send_status as selection 'Unknown'
This 2 fields might overlap. 

But I agree with your concern of having so many fields. 
I will refactor it.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
adns-odoo: Just to evenly process invoices

adns-odoo:
Hi @FlorianGilbert 
I have refactored the code and tried to apply KISS :laughing:

adns-odoo:
Hi @malb-odoo 
I have updated the PR.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -684,3 +717,187 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+        if not pending_move_ids:
+            _logger.info("No commercial moves found for response sync.")
+            return
+
+        total = len(pending_move_ids)
+        total_batches = -(-total // batch_size)  # ceiling division
+        _logger.info("Found %d commercial move(s) to process in %d batch(es).", total, total_batches)
+
+        for offset in range(0, total, batch_size):
+            batch_ids = pending_move_ids[offset : offset + batch_size]
```
FlorianGilbert:
```suggestion
            batch_ids = pending_move_ids[offset:offset + batch_size]
```

[See](https://www.flake8rules.com/rules/E203.html)

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -684,3 +717,187 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            batch_num = offset // batch_size + 1
+            _logger.info("Processing batch %d/%d — move ids: %s", batch_num, total_batches, batch_ids)
+
+            for move in self.env["account.move"].browse(batch_ids):
```
FlorianGilbert: That's not really how batch should work..
[See the doc here](https://www.odoo.com/documentation/master/developer/reference/backend/actions.html#writing-cron-functions).

The idea is that you'll process the 20 first ones from your batch size and the cron needs to know the remaining to relaunch a call to the cron.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -684,3 +717,187 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+            """)
+            % (
+                responder.name,
+                rejection_note,
+            ),
+            message_type="notification",
+            subtype_xmlid="mail.mt_note",
+        )
+        if self.move_type == 'out_invoice':
+            self._l10n_tr_nilvera_add_pdf_to_invoice(
```
FlorianGilbert, line 829: I know it's not your code but passing self as invoices and not using the self that calls the function is not normal. Don't change it but we'll probably have to clean that.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -684,3 +717,187 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+                        "type": "warning",
+                        "next": {"type": "ir.actions.client", "tag": "soft_reload"},
+                    },
+                }
+        errors = [f"{error.get('Code')}: {error.get('Description')}" for error in response["Errors"]]
+        raise ValidationError(self.env._("Error sending request:\n%s", "\n".join(errors)))
+
+    def _l10n_tr_action_send_ticarifatura_response(self, answer_code="approved", rejection_note=""):
+        self.ensure_one()
+        if (self.move_type != "in_invoice" or self.l10n_tr_gib_invoice_scenario != "TICARIFATURA"):
```
FlorianGilbert: Remove the redundant parentheses 😄

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -135,6 +161,9 @@ class AccountMove(models.Model):
         related='partner_id.l10n_tr_nilvera_customer_status',
         help="Shows the Nilvera status of the customer. ",
     )
+    l10n_tr_ticarifatura_last_checked_at = fields.Datetime(
+        string="Commercial Status Check Date", copy=False
```
FlorianGilbert:
```suggestion
        string="Commercial Status Check Date", copy=False,
```
Trailing comma

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -698,3 +733,196 @@ def _reverse_moves(self, default_values_list=None, cancel=False):
...
+
+            # Always update checked date, to avoid checking the same record on the same cron
+            move.write({"l10n_tr_ticarifatura_last_checked_at": fields.Datetime.now()})
+        remaining = (
+            0
+            if len(move_ids_to_process) < limit
+            else self.env["account.move"].search_count(domain)
+        )
+        self.env["ir.cron"]._commit_progress(
+            len(move_ids_to_process), remaining=remaining
```
FlorianGilbert:
```suggestion
            len(move_ids_to_process), remaining=remaining,
```
trailing comma

adns-odoo:
Hi @FlorianGilbert 
I have updated them both with trailing comma

PR #246952, opened 2026-02-03
[FIX] website_project: Avoid duplicated description in project form

sben-odoo:
Steps to reproduce:
===================
1. Go to website > contact us page
2. Submit a new task with description.
3. View the Description in the project app.

-> The description field is already shown in the form by default. You will find another time it is repeated.

Cause:
======
The default field content is always appended to the form content.

Solution:
=========
Avoid adding description to the description field content.

opw-5868382

---

### DIFF
--- a/addons/website_project/controllers/main.py
+++ b/addons/website_project/controllers/main.py
@@ -27,7 +27,9 @@ class WebsiteForm(form.WebsiteForm):
         custom_label = nl2br_enclose(_("Other Information"), 'h4')  # Title for custom fields
         default_field = model.website_form_default_field_id
         default_field_data = values.get(default_field.name, '')
-        default_field_content = nl2br_enclose(default_field.name.capitalize(), 'h4') + nl2br_enclose(html2plaintext(default_field_data), 'p')
+        default_field_content = nl2br_enclose(html2plaintext(default_field_data), 'p')
+        if default_field.name and default_field.name != 'description':
+            default_field_content = nl2br_enclose(default_field.name.capitalize(), 'h4') + default_field_content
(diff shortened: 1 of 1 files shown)


addons/website_project/controllers/main.py
```diff
@@ -27,7 +27,9 @@ def insert_record(self, request, model, values, custom, meta=None):
         custom_label = nl2br_enclose(_("Other Information"), 'h4')  # Title for custom fields
         default_field = model.website_form_default_field_id
         default_field_data = values.get(default_field.name, '')
-        default_field_content = nl2br_enclose(default_field.name.capitalize(), 'h4') + nl2br_enclose(html2plaintext(default_field_data), 'p')
+        if default_field.name and default_field.name != 'description':
+            default_field_content = nl2br_enclose(default_field.name.capitalize(), 'h4')
+        default_field_content = default_field_content + nl2br_enclose(html2plaintext(default_field_data), 'p')
```
nle-odoo: I would expect an error if default_field_content is undefined (if the default field name is not description), you could inverse the line for example

```suggestion
        default_field_content = nl2br_enclose(html2plaintext(default_field_data), 'p')
        if default_field.name and default_field.name != 'description':
            default_field_content = nl2br_enclose(default_field.name.capitalize(), 'h4') + default_field_content
```

nle-odoo (approved):
could be good to me

note: the steps are missing that you need to install website_project and change the form to "Create a task"

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #239499, opened 2025-12-10
[PERF] website_sale: Batch has_published field on category

Beekan:
Previously the function was fetching all the product template ids and looping over them for each product template. This triggered the prefetch_ids and prefetch fields for these products which would cause memory issues due to the bloat of the cache from the prefetcher.

The current way is a read_group over the categories and check if category id exists or not.

Description of the issue/feature this PR addresses:

Current behavior before PR:

Desired behavior after PR is merged:

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #246955, opened 2026-02-03
[PERF] website_sale: Batch has_published field on category

Beekan:
Previously the function was fetching all the product template ids and looping over them for each product template. This triggered the prefetch_ids and prefetch fields for these products which would cause memory issues due to the bloat of the cache from the prefetcher.

The current way is a read_group over the categories and check if category id exists or not.

Backport of odoo/odoo#239499

---


AntoineVDV (approved):
@robodoo r+ fw=no

PR #246956, opened 2026-02-03
[FIX] orm: resolve ANY on non-stored x2many fields into ID sets

apan-odoo:
In `Odoo 19`, the domain engine keeps the `any / not any` operators as relational semantics and translates them to SQL `EXISTS` subqueries. This assumes that `x2many` fields are backed by real relational tables.

However, some `x2many` fields are virtual bridges implemented via a custom `search` method and non-relational storage (e.g. `JSON` fields). A concrete example is `distribution_analytic_account_ids` on `account.move.line`, which maps analytic accounts stored inside the `analytic_distribution` `JSON` field.

Domains such as:
```py
('distribution_analytic_account_ids.plan_id', 'in', [2])
```
are normalized into:
```py
('distribution_analytic_account_ids', 'any', [('plan_id', 'in', [2])])
```
During optimization in `v19`, this hits the relational ANY logic and may raise:
```py
    if not field.relational:
        condition._raise("Cannot use 'any' with non-relational fields")
```
or produce SQL EXISTS clauses that cannot be satisfied by the field’s custom search implementation.

**Steps to reproduce:**
- Created a `v19` db and install Accounting app
- Navigate to `journal items` menu
- Add the custom filter: `[("distribution_analytic_account_ids.plan_id", "in", [2])]`

Screenshot from 2026-02-03 15-15-34

- In logs you will see:
```py
ValueError: Cannot use 'any' with non-relational fields in condition ('analytic_distribution', 'any', [('plan_id', 'in', [2])])
```
- In UI it will pop up domain is invalid.

In `Odoo 18`, similar cases were handled implicitly later during SQL generation, where x2many `any` conditions were resolved by first searching the comodel and then converting the condition into an in operator on the resulting record IDs. This effectively allowed virtual x2many bridges to behave like relational fields at the domain level. [ref](https://github.com/odoo/odoo/blob/18.0/odoo/osv/expression.py#L1173-L1177)

This commit restores that behavior in a way aligned with the `v19` domain architecture by extending `_optimize_any_domain_at_level`:

- When the field is:
  - `one2many` or `many2many`
  - has a custom `search` method
  - is not stored (`store=False`)

- The `any` / `not any` condition is resolved early by:
  - Optimizing the subdomain on the comodel
  - Executing a search on the comodel
  - Rewriting the condition into: `(field, 'in' / 'not in', [ids])`

This keeps the domain engine semantics correct:
- Real relational x2many fields still use SQL EXISTS
- Virtual `x2many` fields using Python search methods receive a standard `in` / `not in` operator with concrete IDs, which their search methods already support

As a result, domains traversing JSON-backed or otherwise virtual `x2many` fields work again in Odoo 19 without weakening validation for genuine non-relational fields.

opw-5875337
upg-3855669

---

### DIFF
--- a/odoo/orm/domains.py
+++ b/odoo/orm/domains.py
@@ -1365,6 +1365,21 @@ def _optimize_any_domain_at_level(level: OptimizationLevel, condition, model):
         comodel = model.env[field.comodel_name]
     except KeyError:
         condition._raise("Cannot determine the comodel relation")
+
+    # Virtual relational field (JSON bridge, computed x2many, etc.)
+    # ANY cannot be expressed as SQL EXISTS → resolve to ids
+    if (
+        field.type in ('one2many', 'many2many')
+        and field.search
+        and not field.store
+        and not field.related
+    ):
+        subdomain = domain._optimize(comodel, level)
+        query = comodel._search(subdomain)
+        right_ids = query.get_result_ids()
+        new_operator = 'in' if condition.operator in ('any', 'any!') else 'not in'
+        return DomainCondition(condition.field_expr, new_operator, right_ids)
+
     domain = domain._optimize(comodel, level)
     # const if the domain is empty, the result is a constant
     # if the domain is True, we keep it as is


kmagusiak (changes requested):
This is not an issue in `domains.py`, the implementation of `_search_distribution_analytic_account_ids` needs to handle the "any" operator because it is a relational field.

Something along the lines (I did not do "not any" here):
```py
        if operator == 'any':
            operator = 'in'
            value = self.env['account.analytic.account'].search(value).ids
        return [('analytic_distribution', operator, value)]
```

PR #246957, opened 2026-02-03
[IMP] survey: show submit confirmation only for certification surveys

yagp-odoo:
Previously, a confirmation dialog could appear when finishing a survey
outside of a live session, even when the survey was not a certification.

This commit restricts the submit confirmation dialog to certification
surveys only. For non-certification surveys, the submit button label
is considered explicit enough, and no extra confirmation is required.

Task-5241498

### DIFF
--- a/addons/survey/static/src/interactions/survey_form.js
+++ b/addons/survey/static/src/interactions/survey_form.js
@@ -70,6 +70,7 @@ export class SurveyForm extends Interaction {
             scoringType: optionsData.scoringType,
             answerToken: optionsData.answerToken,
             surveyToken: optionsData.surveyToken,
+            isCertification: optionsData.certification,
             usersCanGoBack: !!optionsData.usersCanGoBack,
             sessionInProgress: !!optionsData.sessionInProgress,
             isStartScreen: !!optionsData.isStartScreen,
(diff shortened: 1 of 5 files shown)


addons/survey/static/src/interactions/survey_form.js
```diff
@@ -413,8 +414,8 @@ export class SurveyForm extends Interaction {
             this.submitForm({ previousPageId: parseInt(targetEl.dataset.previousPageId) });
         } else if (targetEl.value === "next_post_submit") {
             this.submitForm({ showNextPostSubmitPage: true });
-        } else if (targetEl.value === "finish" && !this.options.sessionInProgress) {
-            // Adding pop-up before the survey is submitted when not in live session
+        } else if (targetEl.value === "finish" && !this.options.sessionInProgress && this.options.isCertification) {
```
mrsr-odoo: Hello @yagp-odoo :wave: ,

below else if condition is removed
could you rebase and adapt it accordingly?

Thanks

addons/survey/static/src/interactions/survey_form.js
```diff
@@ -414,7 +415,10 @@ export class SurveyForm extends Interaction {
         } else if (targetEl.value === "next_post_submit") {
             this.submitForm({ showNextPostSubmitPage: true });
         } else if (targetEl.value === "finish" && !this.options.sessionInProgress) {
-            // Adding pop-up before the survey is submitted when not in live session
+            if (!this.options.isCertification) {
```
awa-odoo: Would do a regular if/else here and not a shortcut, simpler and also consistent with other cases that don't return the result of "submitForm".

awa-odoo (commented):
Hello @yagp-odoo 

One minor comment, otherwise LGTM :heavy_check_mark: 

Thanks & cheers!

PR #246961, opened 2026-02-03
[IMP] account_edi_ubl_cii: Prevent BIS3 without EndpointID to be send

smetl:
If EndpointID is not set, the generated file is invalid due to the 2 following rules:
[PEPPOL-EN16931-R010] Buyer electronic address MUST be provided. [PEPPOL-EN16931-R020] Seller electronic address MUST be provided.

Since this is a configuration issue, there is no point of sending such files to be rejected right away. Instead, let's add those 2 contraints ODOO-side.

task-5890887

---


clbr-odoo (approved):
@robodoo r+

PR #246962, opened 2026-02-03
[FIX] sale: allow access to cart so with expired date

mma1377:
### Issue:
Due to this issue public and portal users cannot access their carts once the so is expired.

#### Steps to reproduce:
1- Using a public user, add a product to cart.
2- Using the admin user, go to the sale order created for the cart.
3- Change the date to an earlier date.
4- Back to public user, try to access the cart.

You will see the error message:
`The sale order has expired.`

### Cause:
This regression is due to #245772.
`_get_payment_values` is not the right place to check if the so is expired, as it is also used inside `_get_express_shop_payment_values`.

opw-5903033

### DIFF
--- a/addons/sale/controllers/portal.py
+++ b/addons/sale/controllers/portal.py
@@ -179,7 +179,7 @@ class CustomerPortal(payment_portal.PaymentPortal):
         }
 
         # Payment values
-        if order_sudo._has_to_be_paid() or payment_amount:
+        if order_sudo._has_to_be_paid() or (payment_amount and not order_sudo.is_expired):
             values.update(self._get_payment_values(
                 order_sudo,
                 is_down_payment=self._determine_is_down_payment(
@@ -187,6 +187,8 @@ class CustomerPortal(payment_portal.PaymentPortal):
                 ),
                 payment_amount=payment_amount,
             ))
+        else:
+            values['payment_amount'] = None
 
         if order_sudo.state in ('draft', 'sent', 'cancel'):
             history_session_key = 'my_quotations_history'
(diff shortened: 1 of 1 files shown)


addons/sale/controllers/portal.py
```diff
@@ -178,6 +178,9 @@ def portal_order_page(
             'payment_amount': payment_amount,
         }
 
+        if order_sudo.is_expired:
+            raise ValidationError(_("The sale order has expired."))
```
Feyensv: This doesn't seem the good way to do it either 🤔 
You should be able to see an expired SO (and it will be displayed that it's expired).

Maybe it should be handled in the if below, to avoid calling `_get_payment_values` when the order is expired

addons/sale/controllers/portal.py
```diff
@@ -180,6 +180,9 @@ def portal_order_page(
 
         # Payment values
         if order_sudo._has_to_be_paid() or payment_amount:
+            if order_sudo.is_expired:
+                raise ValidationError(_("The sale order has expired."))
```
Feyensv:
```suggestion
if order_sudo._has_to_be_paid() or (payment_amount and not order_sudo.is_expired):
```

addons/sale/controllers/portal.py
```diff
@@ -179,14 +179,16 @@ def portal_order_page(
...
+        if order_sudo._has_to_be_paid() or (payment_amount and not order_sudo.is_expired):
             values.update(self._get_payment_values(
                 order_sudo,
                 is_down_payment=self._determine_is_down_payment(
                     order_sudo, amount_selection, payment_amount
                 ),
                 payment_amount=payment_amount,
             ))
+        else:
+            values.update(payment_amount=None)
```
Feyensv:
```suggestion
            values['payment_amount'] = None
```

why did you need to reset the payment amount, did it break something in the view ?

mma1377: Hi @Feyensv , yes that would break views. I was investigating to find the reason before pinging you. It seems it is because `payment_methods_sudo` is not present, but I'm not sure.
However if I reset the payment amount it will be fixed.

Feyensv (approved):
@robodoo r+ priority

PR #246975, opened 2026-02-03
[FIX] pos_restaurant: create new table using current floor max number

vpan-odoo:
Before this commit:
===================
When multiple floors exist, creating a new table assigns the next table number based on the maximum table number across all floors.

After this commit:
==================
When multiple floors exist, creating a new table assigns the next table number based on the maximum table number of the current floor.

Task: 5888449

---


davidmonnom (approved):
@robodoo r+

PR #246986, opened 2026-02-03
[FIX] pos_glory_cash: various fixes and improvements

MaxTheWhale:
- Cash machine UI in the POS has been cleaned up
  - Reset button and Download logs button have been moved to the
    backend.
  - Warnings about low cash now take up less space.
  - The cash inserted amount is now shown inside the payment line status
    box.
- Shortcut to the Glory admin page has been added to the payment method
  form.
- Error detail text is now shown when provided by the cash machine.
- Bugfixes
  - There should no longer be multiple disconnected popups appear if the
    tab is left inactive for a while.
  - The session/occupy settings of the connected Glory machine are now
    respected, this reduces the number of requests sent when they are
    not enabled.
- Some tests have been added for the Glory service.

In order to add Glory functionality in the backend, the Glory service was split into two components. The `glory_service.js` file handles all the communication with the Glory machine, but it doesn't handle the POS integration at all. The POS side is now handled by `payment_glory.js`. This means that in the backend we can still use `glory_service.js` to interact with the machine without being in the POS.

task-5891688

---


addons/pos_glory_cash/__manifest__.py
loouis-t, line 1: these updates will require an update of the module, but maybe it's not that critical as not many people have it yet

addons/pos_glory_cash/static/src/backend/glory_admin_buttons/glory_admin_buttons.js
```diff
@@ -0,0 +1,92 @@
...
+            this.notification.add("Please configure the IP address before opening the admin page", {
+                type: "warning",
+            });
+            return;
+        }
+        const protocol = window.location.protocol;
+        const port = protocol === "http:" ? 3000 : 3001;
+        browser.open(
+            `${protocol}//${this.props.record.data.glory_websocket_address}:${port}/control`
+        );
```
loouis-t, line 85: http://127.0.0.1:3000/control returns `Cannot GET /control`, on the Glory VM (due to VM network configuration?)

loouis-t (approved):
download logs ✅
reset cash machine ✅
open admin page ❓ 
pay order using cash machine still works as before ✅

addons/pos_glory_cash/static/src/backend/glory_admin_buttons/glory_admin_buttons.js
MaxTheWhale: Yes it's a setting that has to be enabled on the Glory web interface, I'll have to mention this when I update the docs. If they don't have this enabled the only way they can manage their cash machine is with the connected touchscreen.

addons/pos_glory_cash/__manifest__.py
MaxTheWhale: I've just double checked by installing in 19.0 and then checking out my branch, the only effect is that the buttons don't appear in the payment method form, but everything else is working fine. I think this is probably okay since the buttons weren't there before?

loouis-t (approved):
robodoo r+

PR #246989, opened 2026-02-03
[IMP] mail: redesign message highlight effect for better focus

pmah-odoo:
When a message is highlighted it visually `jumps` due to a vertical translation (`translateY`). This animation creates a layout shift that is distracting and makes the text difficult to read while the transition is occurring.

The highlighted message now remains static, eliminating the visual jump. Instead, the focus is achieved by dimming and applying a blur effect to all surrounding messages and elements in the thread. This creates a stable `spotlight` effect that improves readability and focus on highlighted message.

Before
before

After
after

task-[4543530](https://www.odoo.com/odoo/project/1519/tasks/4543530)

---

### DIFF
--- a/addons/mail/static/src/core/common/message.scss
+++ b/addons/mail/static/src/core/common/message.scss
@@ -10,11 +10,6 @@
         --border-opacity: .2;
     }
 
-    &.o-highlighted {
-        transform: translateY(-#{map-get($spacers, 3)});
-        z-index: 2;
-    }
-
     &.o-actionMenuMobileOpen {
         background-color: rgba($o-action, .075);
         outline: 1px solid rgba($o-action, .125);
@@ -177,4 +172,15 @@
 
 .o-mail-Message-translated {
     color: $o-enterprise-action-color;
+}
+
+.o-mail-Thread-content {
+    > * {
+        transition: opacity .3s ease-in-out, filter .3s ease-in-out;
+    }
+
+    &:has(.o-highlighted) > *:not(.o-highlighted) {
+        opacity: .4;
+        filter: blur(1px);
+    }
 }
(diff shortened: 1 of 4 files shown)


addons/mail/static/src/core/common/message.scss
```diff
@@ -176,3 +171,13 @@
...
+
+.o-mail-Thread:has(.o-highlighted) {
+    > div {
+        > *:not(.o-highlighted) {
+            transition: opacity .3s ease-in-out, filter .3s ease-in-out;
+            opacity: .4;
+            filter: blur(3px);
+        }
+    }
+}
```
ramh-odoo: We should follow the BEM pattern here and avoid relying on nested or structural selectors.

addons/mail/static/src/core/common/message.scss
```diff
@@ -178,3 +173,14 @@
...
 }
+
+.o-mail-Thread-content {
+    > * {
+        transition: opacity .3s ease-in-out, filter .3s ease-in-out;
+    }
+
+    &:has(.o-highlighted) > *:not(.o-highlighted) {
+        opacity: .4;
+        filter: blur(3px);
```
alexkuhn: I would `blur(1px)`, too much blur is exhausting to many people.

For me it doesn't bother but I think this is similar to motion blur in some games, lots of people hate it, therefore I think tuning down to `blur(1px)` keeps it visible but gives less motion sickness.

And some blur is fine, especially with the auto-scroll. It works well with motion, so that's why the `1px` blur is good.

alexkuhn (commented):
I'd also take the opportunity to reduce the highlight effect from 2 seconds to 1.5 seconds:
https://github.com/odoo/odoo/blob/19.0/addons/mail/static/src/utils/common/hooks.js#L328

2 seconds is slightly too much, especially when there's barely any scroll.
1.5 seconds is long enough in case we have to jump very far. The shortest value to meet both is the best.

alexkuhn (approved):
@robodoo delegate+ r+

PR #246991, opened 2026-02-03
[FIX] account: Clear message_main_attachment_id field at unattaching

hsal-odoo:
Problem:
when a user uploads an XML file containing an embedded PDF to create a bill,
the bill is created correctly and the PDF is set as the main attachment.
However, when the same XML is used to create a bill via email aliases,
the PDF is not set as the bill’s main attachment.

Root cause:
1. When an email is received, it's first posted as a message on the bill chatter
2. Posting the email sets its XML as the main attachment of the bill
3. _fix_attachments_on_record from account.document.import.mixin then removes res_id and res_model fields from the XML
4. The PDF is then attached as the main attachment of the bill
5. In documents_account module, when the PDF is set as the main attachment,
it mistakenly assigns it as an attachment to the XML document because the XML was assigned as the bill main attachment,
resulting in the bill having no attachment actually linked to it.

Solution:
This commit fixes this issue by clearing the field message_main_attachment_id whenever the attachment in it
gets unattached in _fix_attachment_on_record.

task-5900088

### DIFF
--- a/addons/account/models/account_document_import_mixin.py
+++ b/addons/account/models/account_document_import_mixin.py
@@ -379,6 +379,10 @@ class AccountDocumentImportMixin(models.AbstractModel):
     # Helpers to consistently attach/unattach attachments to records
     # --------------------------------------------------------------
 
+    def _attachment_fields_to_clear(self):
+        """ Return a list of fields that should be cleared when an attachment is unattached from the record. """
+        return []
+
     def _fix_attachments_on_record(self, attachments):
         """ Ensure that only attachments of certain types appear in `self`'s attachments.
 
@@ -396,6 +400,8 @@ class AccountDocumentImportMixin(models.AbstractModel):
             })
         attachments_to_unattach = (attachments - attachments_to_attach).filtered(lambda a: a.res_model == self._name and not a.res_field)
         if attachments_to_unattach:
+            for fname in self._attachment_fields_to_clear():
+                self[fname] -= attachments_to_unattach
             attachments_to_unattach.write({
                 'res_model': False,
                 'res_id': 0,
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -6897,6 +6897,9 @@ class AccountMove(models.Model):
 
         return move
 
+    def _attachment_fields_to_clear(self):
+        return super()._attachment_fields_to_clear() + ['message_main_attachment_id']
+
     def _message_post_after_hook(self, new_message, message_values):
         """ This method processes the attachments of a new mail.message. It handles the 3 following situations:
             (1) receiving an e-mail from a mail alias. In that case, we potentially want to split the attachments into several invoices.
--- a/addons/account/tests/test_account_incoming_supplier_invoice.py
+++ b/addons/account/tests/test_account_incoming_supplier_invoice.py
@@ -8,6 +8,7 @@ from unittest.mock import patch
 from odoo import Command
 from odoo.exceptions import ValidationError
 from odoo.tests import tagged, RecordCapturer
+from odoo.tools import file_open
 from odoo.tools.misc import mute_logger
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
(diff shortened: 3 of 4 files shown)


reth-odoo (commented):
I suppose it makes sense insofar as modifying these means you potentially don't have access to the attachment anymore. But I'm not sure we want to do that in stable.

We should be careful to check this doesn't impact other flows as it already seems to break some tests. And for example it can be valid for it to be linked to a MailMessage, or simply to nothing, so it's not clear that we necessarily want to reset it in all cases.

Seems more of a specific problem for your flow if I understand well, so maybe just fix it there to be safe?

reth-odoo (commented):
Why do we now prevent setting it's res_model/res_id compared to previously forcing it to not be the main attachment?

From your description if the behavior is different on receiving a message to an alias, shouldn't the fix be in the `if self.env.context.get('from_alias'):` branch of `AccountMove._message_post_after_hook`? As I understand it should be calling `_fix_attachments_on_record` on the invoices that are created, and then their "main_attachment_id" should be computed automatically (?)

A test may be useful here either way 👀

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -550,7 +550,8 @@ def _import_attachments(self, invoice, tree):
                         invoice.message_main_attachment_id.name.endswith('.xml') and \
                         'pdf' not in invoice.message_main_attachment_id.mimetype and \
                         mimetype == 'application/pdf':
-                    invoice._message_set_main_attachment_id(attachment, force=True, filter_xml=False)
+                    invoice.message_main_attachment_id = False
+                    invoice._message_set_main_attachment_id(attachment, filter_xml=False)
```
aboo-odoo: This is the same as `force=True`. When looking into `def _message_set_main_attachment_id`, we can notice that if there is no main attachment or if force then we set the main attachment

Your test proves that since it does not fail without your change :/

hsal-odoo: it actually fails without my change if `documents_account` is installed in the DB
The issue is, `documents_account` creates a document for an attachment, and when we attempt to update `message_main_attachment_id` on the invoice (from XML to PDF), it adds the PDF as a new version of the document instead of setting it on the `message_main_attachment_id`. It's a pretty interleaved scenario, but I am not sure what is the best solution for it. What I do now is that I clear the message_main_attachment_id (instead of updating it) to avoid the versioning link problem.

addons/account/models/account_document_import_mixin.py
```diff
@@ -396,6 +396,9 @@ def _fix_attachments_on_record(self, attachments):
             })
         attachments_to_unattach = (attachments - attachments_to_attach).filtered(lambda a: a.res_model == self._name and not a.res_field)
         if attachments_to_unattach:
+            for fname, field in self._fields.items():
+                if field.comodel_name == 'ir.attachment' and field.store:
+                    self[fname] -= attachments_to_unattach
```
aboo-odoo: I'm sorry I don't understand your fix.. Didn't we agree on not deleting attachment that are set as main attachment ? + It might be a good idea to add a comment

aboo-odoo: I'm wondering if it wouldn't be better to add a method here that will be overriden in relevant models. The overrides would actually limit the removal of the attachments from the wanted fields instead of all the fields. Avoiding removing the attachments from fields we don't want

```suggestion
            fields = self._get_fields_...()
            
            for f in fields:
                self[f] -= attachments_to_unattach
```

In the mixin
```
def _get_fields_...(self):
    return []
```

In account.move
```
def _get_fields_...(self):
    return super()._get_fields...() + ['message_main_attachment_id']
```

What do you think ?

addons/account/models/account_document_import_mixin.py
```diff
@@ -396,6 +400,10 @@ def _fix_attachments_on_record(self, attachments):
             })
         attachments_to_unattach = (attachments - attachments_to_attach).filtered(lambda a: a.res_model == self._name and not a.res_field)
         if attachments_to_unattach:
+            for fname in self._attachment_fields_to_clear():
+                for attachment_to_unattach in attachments_to_unattach:
+                    if attachment_to_unattach in self[fname]:
+                        self[fname] -= attachment_to_unattach
```
aboo-odoo: This should also work 🤔 
```suggestion
                self[fname] -= attachments_to_unattach
```

addons/account/models/account_move.py
```diff
@@ -6900,6 +6897,9 @@ def is_right_company(partner):
 
         return move
 
+    def _attachment_fields_to_clear(self):
+        return ['message_main_attachment_id']
```
aboo-odoo:
```suggestion
        return super()._attachment_fields_to_clear + ['message_main_attachment_id']
```
No ?

addons/account/models/account_document_import_mixin.py
hsal-odoo: but the if-condition avoids unnecessary call to write()

aboo-odoo: There won't be a write if the cache does not change, so if the value in the record is not part of the recordset, it will remain unchanged and thus not written

aboo-odoo (approved):
Thank you ! 😄 

@robodoo r+

PR #246993, opened 2026-02-03
[IMP] website: clarify default website language behavior

alap-odoo:
1. Clarify default website language behavior
    Previously, the purpose of a website's default language was unclear to many users. It was commonly assumed that the website would always open in the selected default language.
    
    In practice, if a visitor's browser language is available on website, it takes precedence over the default language. This PR adds clear help messages to explain the same.

2. Restrict default languages to installed ones
    
    Before this PR, users could select a default language that was not installed on the website.

    Steps to reproduce:
    i. Add new languages in Settings > General.
    ii. Add a language in Settings > Website > Languages.
    iii. Go to Settings > Website > Languages > Default.

    The default language dropdown included languages which are not installed on the website.
    
    After this PR, only languages installed on the website are available for default language selection.

task-5901757

**After this PR**
| If single language is installed | If multiple languages are installed |
|-----------------------------|---------------------------------|
| image | image |

### DIFF
--- a/addons/website/models/res_config_settings.py
+++ b/addons/website/models/res_config_settings.py
@@ -206,6 +206,7 @@ class ResConfigSettings(models.TransientModel):
             self.website_default_lang_id = False
         elif self.website_default_lang_id not in language_ids:
             self.website_default_lang_id = language_ids[0]
+        self.website_language_count = len(language_ids)
 
     def action_website_create_new(self):
         return {
--- a/addons/website/views/res_config_settings_views.xml
+++ b/addons/website/views/res_config_settings_views.xml
@@ -50,7 +50,7 @@
                             <field name="language_ids" widget="many2many_tags" options="{'no_create': True, 'no_open': True}" required="website_id"/>
                             <div class="row mt8" invisible="website_language_count &lt; 2">
                                 <label class="col-lg-3" string="Default" for="website_default_lang_id"/>
-                                <field name="website_default_lang_id" options="{'no_open': True, 'no_create': True}" required="website_id"/>
+                                <field name="website_default_lang_id" options="{'no_open': True, 'no_create': True}" required="website_id" domain="[('id', 'in', language_ids)]"/>
                             </div>
                             <div class="mt8">
                                 <button type="action" name="%(base.action_view_base_language_install)d" string="Install new languages" class="btn-link" icon="oi-arrow-right"/>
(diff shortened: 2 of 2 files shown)


addons/website/views/res_config_settings_views.xml
```diff
@@ -43,6 +43,10 @@
                         <setting string="Languages"
                                  title="Languages available on your website"
                                  documentation="/applications/websites/website/configuration/translate.html">
+                            <div class="text-muted mt-n3 mb16">
```
alap-odoo, line 46: Here, `mt-n3 mb16` translates to the following CSS:

```scss
margin-top: -16px;
margin-bottom: 16px;
```
This was added to support a conditional help message based on the number of installed languages.

The `help` prop in `<setting>` expects a string value (see image 1), so it cannot handle conditional logic. To work around this limitation, I introduced an additional div and rendered the message conditionally inside it.

However, according to the implementation in "**setting.xml**", any div placed inside `<setting>` automatically receives a top margin of "16px" (see image 2). Since the original help text is rendered above the content without that extra spacing, adding a new div caused an unintended layout shift.

To preserve the original spacing and visual behavior, I applied `mt-n3` (-16px top margin) and `mb16` (+16px bottom margin). This offsets the automatic margin and keeps the layout consistent with the standard help rendering.

Image

Image

addons/website/models/res_config_settings.py
```diff
@@ -206,6 +206,7 @@ def _onchange_language_ids(self):
             self.website_default_lang_id = False
         elif self.website_default_lang_id not in language_ids:
             self.website_default_lang_id = language_ids[0]
+        self.website_language_count = len(language_ids)
```
alap-odoo, line 209: It updates the value of `website_language_count` when the language selection changes. This ensures that the help message is updated immediately as soon as the user adds another language.

FrancoisGe, line 209: We should only propose the language used by the website

Image

FrancoisGe: @meda-odoo will create another task for this issue, it s not link to your PR

FrancoisGe:
@robodoo r+

FrancoisGe:
@robodoo r-

addons/website/models/res_config_settings.py
FrancoisGe: @alap-odoo Could you do it in your task ?
You just need to add `domain="[('id', 'in', language_ids)]"` on `<field name="website_default_lang_id">` 
in a another commit :) 
You can ping me when it s done

alap-odoo: Hello @shsa-odoo 
The suggested work is done. Can you please check it once ?

FrancoisGe:
@robodoo r+ rebase-ff

PR #246994, opened 2026-02-03
[FIX] project: working time holiday with multicompany

maedekey:
__
Fix after revert bf9729e997b15777a981ac258b0088107804fbe6 
__
Let's say we have 2 companies: company A and company B. We create a public time off of a few days starting before today and ending 2 days later in company A, then switch back to company B. In company B, we create a project and a task, and assign this task. The working time to assign will stay at 0.

1. Switch to company A and create a timeoff starting before today and ending later.
2. Switch back to company B. Create a project, a stage and a task.
3. Enable the debugger.
4. The field Working Time to Assign is invisible by default, so open studio, click on View, and check Show Invisible Elements.
5. Click on the tab Extra info and on the block Working time to assign. Uncheck Invisible.
6. Close studio and assign someone to the task. Make sure that you do this operation at a different time than the one recorded for the last stage change.

The hours under Working Time to Assign should compute the difference between the last time the task got its stage changed and the time of assignation

Nothing happens

When computing the working time to assign, we also take into consideration leaves: if this happened during public holidays, we consider that it took no working time to get assigned. However, when a holiday is set in another company, the Working Time to Assign duration will be impacted, as the domain to retrieve the corresponding leaves is the following: https://github.com/odoo/odoo/blob/c7e965a61b7ce856c2daa8e2574cf4c60caf7a20/addons/resource/models/resource_calendar.py#L537-L546

The company isn't taken into account in the domain, applying the holiday for every company.
 __
opw-5222883

### DIFF
--- a/addons/project/models/project_task.py
+++ b/addons/project/models/project_task.py
@@ -583,10 +583,10 @@ class Task(models.Model):
         )
         for task in task_linked_to_calendar:
             dt_create_date = fields.Datetime.from_string(task.create_date)
-
+            domain = [('company_id', 'in', task.project_id.company_id.ids), ('time_type', '=', 'leave')]
             if task.date_assign:
                 dt_date_assign = fields.Datetime.from_string(task.date_assign)
-                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_assign, compute_leaves=True)
+                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_assign, compute_leaves=True, domain=domain)
                 task.working_hours_open = duration_data['hours']
                 task.working_days_open = duration_data['days']
             else:
@@ -595,7 +595,7 @@ class Task(models.Model):
 
             if task.date_end:
                 dt_date_end = fields.Datetime.from_string(task.date_end)
-                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_end, compute_leaves=True)
+                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_end, compute_leaves=True, domain=domain)
                 task.working_hours_close = duration_data['hours']
                 task.working_days_close = duration_data['days']
             else:
--- a/addons/project/tests/test_multicompany.py
+++ b/addons/project/tests/test_multicompany.py
@@ -2,8 +2,11 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from contextlib import contextmanager
+from datetime import datetime
+from freezegun import freeze_time
 from lxml import etree
 
+from odoo import Command, fields
 from odoo.tests import Form, TransactionCase
 from odoo.exceptions import AccessError, UserError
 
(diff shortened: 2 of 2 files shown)


addons/project/models/project_task.py
```diff
@@ -586,7 +586,7 @@ def _compute_elapsed(self):
 
             if task.date_assign:
                 dt_date_assign = fields.Datetime.from_string(task.date_assign)
-                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_assign, compute_leaves=True)
+                duration_data = task.project_id.resource_calendar_id.get_work_duration_data(dt_create_date, dt_date_assign, compute_leaves=True, domain=[('company_id', 'in', self.project_id.company_id.mapped('id')), ('time_type', '=', 'leave')])
```
taqu-odoo: Should be `task.project_id.company_id.id` instead of `self.project_id.company_id.mapped('id')`.
Style-wise, since the same domain is copied below, we could also consider creating it once before the two `if`s to avoid duplication

addons/project/tests/test_project_recurrence.py
```diff
@@ -278,3 +278,33 @@ def test_recurrence_copy_task_dependency(self):
 
         self.assertEqual(len(side_task1.depend_on_ids), 0)
         self.assertCountEqual(side_task2.depend_on_ids.ids, [node3.id, parent_copy_node3.id], 'SideTask2 - Node3 and SideTask2 - Node3copy relations should be present')
+
+    @freeze_time("2019-5-28 08:00:00")
+    def test_date_to_assign_project(self):
```
taqu-odoo: Wouldn't it make more sense to have the test in `test_multicompany` instead of `test_project_recurrence`?

addons/project/tests/test_multicompany.py
```diff
@@ -3,9 +3,12 @@
 
 from contextlib import contextmanager
 from lxml import etree
+from datetime import datetime
+from freezegun import freeze_time
 
 from odoo.tests import Form, TransactionCase
 from odoo.exceptions import AccessError, UserError
+from odoo import Command, fields
```
taqu-odoo:
```suggestion
from contextlib import contextmanager
from datetime import datetime
from freezegun import freeze_time
from lxml import etree

from odoo import Command, fields
from odoo.tests import Form, TransactionCase
from odoo.exceptions import AccessError, UserError
```

taqu-odoo (approved):
Left one comment, appart from that LGTM

addons/project/models/project_task.py
```diff
@@ -583,10 +583,10 @@ def _compute_elapsed(self):
         )
         for task in task_linked_to_calendar:
             dt_create_date = fields.Datetime.from_string(task.create_date)
-
+            domain = [('company_id', 'in', task.project_id.company_id.mapped('id')), ('time_type', '=', 'leave')]
```
xavierbol: equivalent to:
```suggestion
            domain = [('company_id', 'in', task.project_id.company_id.ids), ('time_type', '=', 'leave')]
```

xavierbol (commented):

robodoo delegate+

you can r+ once it is done. 🙂

PR #246995, opened 2026-02-03
[FIX] mrp: prevent duplicated workorders on BoM change

amahm-odoo:
Steps to reproduce:
1. Create a product and two BoMs: BoM A (with operations) and BoM B (empty).
2. Create a Manufacturing Order (MO) for the product selecting BoM A.
3. Switch BoM A to BoM B, then switch back to BoM A.
4. Modify the production quantity field. -> New operation lines are appended every time the quantity is changed.

The issue occurred because _compute_workorder_ids used 'wo.ids' to filter existing workorders. In the "Draft" state (UI/onchange), records exist as "virtual records" (NewIds). For these records, .ids returns an empty list [], which evaluates to False in Python.
Consequently, the existing virtual workorders were filtered out of the dictionary used to map operations to existing lines. The logic assumed the lines didn't exist and used Command.create() instead of Command.update(), causing duplication. Similar issues existed where 'NewIds' were ignored during BoM swaps, leaving "phantom" records in the cache.

Solution:
Removing the '.ids' check and using '.mapped('id')' ensures the computation remains "virtual-aware" and stable across sequential onchanges.

TECHNICAL JUSTIFICATION:
In Odoo 18.0, the ORM explicitly supports using Command.update and Command.delete with virtual records (NewIds) without an origin. This is handled by the 'write_new' method in relational fields:
- Virtual browse wraps IDs in NewId: https://github.com/odoo/odoo/blob/f688c6b66310438fa3e36a207770a63d0d8fffa5/odoo/fields.py#L4826-L4855

opw-5489862

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -586,7 +586,7 @@ class MrpProduction(models.Model):
             workorders_list = [Command.link(wo.id) for wo in production.workorder_ids.filtered(lambda wo: wo.ids)]
             relevant_boms = [exploded_boms[0] for exploded_boms in production.bom_id.explode(production.product_id, 1.0, picking_type=production.bom_id.picking_type_id)[0]]
             # we don't delete wo's that are not bom related nor related to a subom
-            deleted_workorders_ids = production.workorder_ids.filtered(lambda wo: wo.operation_id and wo.operation_id.bom_id not in relevant_boms).ids
+            deleted_workorders_ids = production.workorder_ids.filtered(lambda wo: wo.operation_id and wo.operation_id.bom_id not in relevant_boms).mapped('id')
             workorders_list += [Command.delete(wo_id) for wo_id in deleted_workorders_ids]
             if not production.bom_id and not production._origin.product_id:
                 production.workorder_ids = workorders_list
@@ -622,7 +622,7 @@ class MrpProduction(models.Model):
                             'state': 'pending',
                         }]
                 workorders_dict = {wo.operation_id.id: wo for wo in production.workorder_ids.filtered(
-                    lambda wo: wo.operation_id and wo.ids and wo.id not in deleted_workorders_ids)}
+                    lambda wo: wo.operation_id and wo.id not in deleted_workorders_ids)}
                 for workorder_values in workorders_values:
                     if workorder_values['operation_id'] in workorders_dict:
                         # update existing entries
@@ -632,7 +632,7 @@ class MrpProduction(models.Model):
                         workorders_list += [Command.create(workorder_values)]
                 production.workorder_ids = workorders_list
             else:
-                production.workorder_ids = [Command.delete(wo.id) for wo in production.workorder_ids.filtered(lambda wo: wo.ids and wo.operation_id)]
+                production.workorder_ids = [Command.delete(wo_id) for wo_id in production.workorder_ids.filtered(lambda wo: wo.operation_id).mapped('id')]
 
     @api.depends('state', 'move_raw_ids.state')
     def _compute_reservation_state(self):
--- a/addons/mrp/tests/test_order.py
+++ b/addons/mrp/tests/test_order.py
@@ -5375,6 +5375,46 @@ class TestMrpOrder(TestMrpCommon):
         mo = mo_form.save()
         self.assertEqual(len(mo.workorder_ids), 1)
 
+    def test_workorder_update_based_on_bom_and_qty_onchanges(self):
+        """Test that work orders are updated rather than duplicated when
+        switching BoMs and then modifying the quantity in the draft form.
+        """
+        product = self.env['product.product'].create({
+            'name': 'Test Finished Product',
+            'is_storable': True,
+        })
+        workcenter = self.env['mrp.workcenter'].create({'name': 'Assembly Line'})
+        operation_vals = [Command.create({
+            'name': 'Manual Assembly',
+            'workcenter_id': workcenter.id,
+            'time_cycle': 60,
+            'sequence': 1,
+        })]
+        bom_empty, bom_with_ops = self.env['mrp.bom'].create([{
+            'product_tmpl_id': product.product_tmpl_id.id,
+            'product_qty': 1.0,
+            'operation_ids': vals,
+        } for vals in ([], operation_vals)])
+
+        mo = self.env["mrp.production"].create({
+            "product_id": product.id,
+            "bom_id": bom_empty.id,
+            "product_uom_id": product.uom_id.id,
+            "product_qty": 1.0,
+        })
+
+        with Form(mo) as mo_form:
+            mo_form.bom_id = bom_with_ops
+            self.assertEqual(len(mo_form.workorder_ids), 1)
+            mo_form.product_qty = 10.0
+            self.assertEqual(len(mo_form.workorder_ids), 1)
+            mo_form.bom_id = bom_empty
+            self.assertEqual(len(mo_form.workorder_ids), 0)
+            mo_form.bom_id = bom_with_ops
+            self.assertEqual(len(mo_form.workorder_ids), 1)
+            mo_form.bom_id = self.env['mrp.bom']
+            self.assertEqual(len(mo_form.workorder_ids), 0)
+
 
 @tagged('-at_install', 'post_install')
 class TestTourMrpOrder(HttpCase):


addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,63 @@ def test_change_bom_and_quantity_together(self):
         mo = mo_form.save()
         self.assertEqual(len(mo.workorder_ids), 1)
 
+    def test_workorder_duplication_on_bom_change_and_qty_edit(self):
```
lase-odoo: The name of a test should be about the functional flow you are testing or about in the worst cases about the expected behavior. The "duplication" of the workorders is an unexpected behavior, it should not be part of the title. A more appropriate name would be something alike:
```suggestion
    def test_workorder_update_based_on_bom_and_qty_onchanges(self):
```

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,63 @@ def test_change_bom_and_quantity_together(self):
...
+        """ Test that work orders are updated rather than duplicated when
+            switching BoMs and then modifying the quantity in the draft form.
+        """
```
lase-odoo: Same as for the title, you should only describe the flow/expected behaviors. Also, I would not use the word `draft` as it is usually used in a different context than the one you expect. It is common for form's to present different values than the one saved in DB. We usually refer to these Form changes as "onchange" mechanisms (this is btw the mechanism at the origin of your New ids in the compute).

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,63 @@ def test_change_bom_and_quantity_together(self):
...
+        # 1. Setup Data using TestMrpCommon's workcenter_1
```
lase-odoo: There is no need to comment the test that much. You should only comment lines that are unexpected or hide subtleties that you want to underline, the flow should be clear from the code, asserts and test description. I think you can safely remove every of these comments.

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,63 @@ def test_change_bom_and_quantity_together(self):
...
+                    'time_cycle': 60
+                }),
+            ]
+        })
+
+        # BoM B: Empty (No operations)
+        bom_empty = self.env['mrp.bom'].create({
+            'product_tmpl_id': product.product_tmpl_id.id,
+            'product_qty': 1.0,
+        })
```
lase-odoo: Records should be created in batches when possible. This approach rely on the orm for optimizations.
```suggestion
       bom_empty, bom_with_ops = self.env['mrp.bom'].create([
           {
                'product_tmpl_id': product.product_tmpl_id.id,
                'product_qty': 1.0,
           },
           {
                'product_tmpl_id': product.product_tmpl_id.id,
                'product_qty': 1.0,
                'operation_ids': [Command.create({
                    'name': 'Manual Assembly',
                    'workcenter_id': self.workcenter_1.id,
                    'time_cycle': 60
                })]
            },
        ])
```
In the present case it would probably be even cleaner with a list comprehension alike:
```suggestion
        operation_vals = [Command.create({
            'name': 'Manual Assembly',
            'workcenter_id': self.workcenter_1.id,
            'time_cycle': 60
        })]
        bom_empty, bom_with_ops = self.env['mrp.bom'].create([{
            'product_tmpl_id': product.product_tmpl_id.id,
            'product_qty': 1.0,
            'operation_ids': vals,
        } for vals in ([], operation_vals)])
```

lase-odoo (changes requested):

Concerning the test, it is not failing without your change and hence it is not testing the use case you change properly. It is important that it does. Could you try to modify the test for it to fail without the fix ? Note that due to the nature of the issue it might be possible that the Form class does not allow you to reproduce the use case in which case a tour will be necessary.

IMO, the fix is going into the right direction. However, it requires much more explanations. The `.ids' you remove was purposefully added by this commit https://github.com/odoo/odoo/commit/b3580e84fb3fcb6f00d1c733a790ff1099d538b5
precisely to not consider `NewIds`. At first glimpse your fix seems to work as the Command.update looks like it can be provided Newrecords in the values but for me to be convinced would you mind citing the code proving that this use case is indeed supported by the ORM in 18.0 ?

Also, the issue you are currently addressing is not the only issue introduced by https://github.com/odoo/odoo/commit/b3580e84fb3fcb6f00d1c733a790ff1099d538b5
There are a lot of `ids` that should porbably be removed or replaced by `mapped('id')`.
For instance, if you start with the empty bom swap to the bom with an operation and then back to the empty bom, the new operation will not be removed because this `.ids` is not a `mapped('id')`:
https://github.com/odoo/odoo/blob/5b39cdceb84758f767b83b741be927c1a7cfac7a/addons/mrp/models/mrp_production.py#L589-L590
These other issues should also be addressed, properly tested and the correct fix should be defended with arguments by citing the code as well. Can we add command.delete on NewIds without origins, are these records properly handled by the orm when performing multiple onchanges,...

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,40 @@ def test_change_bom_and_quantity_together(self):
         mo = mo_form.save()
         self.assertEqual(len(mo.workorder_ids), 1)
 
+    def test_workorder_duplication_on_bom_change_and_qty_edit(self):
```
lase-odoo: If you chose to not apply a comment, could you argue why yo do not ? That way I do not need to make the same review twice: https://github.com/odoo/odoo/pull/246995#discussion_r2763839024

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,40 @@ def test_change_bom_and_quantity_together(self):
...
+            'is_storable': True,
+        })
+        workcenter = self.env['mrp.workcenter'].create({'name': 'Assembly Line'})
+        operation_vals = [Command.create({
+            'name': 'Manual Assembly',
+            'workcenter_id': workcenter.id,
+            'time_cycle': 60,
+            'sequence': 1,
+        })]
+
```
lase-odoo: uncessary space

addons/mrp/tests/test_order.py
lase-odoo: The test should have a description describing the flow you protected. Also, as suggested here https://github.com/odoo/odoo/pull/246995#discussion_r2763896860 it should refer to the "onchange" machanism.

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,40 @@ def test_change_bom_and_quantity_together(self):
...
+        bom_empty, bom_with_ops = self.env['mrp.bom'].create([{
+            'product_tmpl_id': product.product_tmpl_id.id,
+            'product_qty': 1.0,
+            'operation_ids': vals,
+        } for vals in ([], operation_vals)])
+
+        mo = Form(self.env['mrp.production'])
+        mo.product_id = product
+        mo.bom_id = bom_empty
+        mo = mo.save()
```
lase-odoo: Use create when you do not need to use a form. These steps only create an `mrp.production` and can be performed directly.

addons/mrp/models/mrp_production.py
```diff
@@ -632,7 +632,7 @@ def _compute_workorder_ids(self):
                         workorders_list += [Command.create(workorder_values)]
                 production.workorder_ids = workorders_list
             else:
-                production.workorder_ids = [Command.delete(wo.id) for wo in production.workorder_ids.filtered(lambda wo: wo.ids and wo.operation_id)]
+                production.workorder_ids = [Command.delete(wo_id) for wo_id in production.workorder_ids.filtered(lambda wo: wo.operation_id).mapped('id')]
```
lase-odoo, line 635: The test is not failing without this change. It is therefore either unnecessary or untested. In the present case it is just untested and could be easily by adding these lines at the end of the test:
```py
            mo_form.bom_id = bom_with_ops
            self.assertEqual(len(mo_form.workorder_ids), 1)
            mo_form.product_qty = 0
            self.assertEqual(len(mo_form.workorder_ids), 0)
```

addons/mrp/tests/test_order.py
```diff
@@ -5375,6 +5375,40 @@ def test_change_bom_and_quantity_together(self):
...
+
+        with Form(mo) as mo_form:
+            mo_form.bom_id = bom_with_ops
+            self.assertEqual(len(mo_form.workorder_ids), 1, "Should have 1 workorder after selecting valid BoM")
+
+            mo_form.product_qty = 10.0
+            self.assertEqual(len(mo_form.workorder_ids), 1, "Workorder count duplicated after qty change! Got %s" % len(mo_form.workorder_ids))
```
lase-odoo: The message of an assert should always be about the expected behavior. Imo, it is not necessary in these 3 asserts but if you really want to write something, you should write: `"The mo should have a singe operation."`

addons/mrp/models/mrp_production.py
lase-odoo, line 596: According to the comment prior to the condition, the condition should be triggered in case of a second onchange of the bom, and in that case the `workorder_ids` are all new ids so that I agree that this ids should stay untouched to only check that their are already existing wo linked to the mo.

lase-odoo (changes requested):
Thanks for the update ! 🌊

The fix LGTM now. At least we seem to address all the onchange use cases I can think of the `compute_workorder_ids`.  That being said, you did update the PR title bu not the commit title. You did not apply some of my previous comments without explaining why you did not and one of your change is still untested and can be. I left a few comments. 

Also, concerning the explanation about the fact that the orm handles writing virtual values in commands, it is due to these lines(the first two explicitly accepting new and vritual ids): https://github.com/odoo/odoo/blob/6b232a2dc96a995eaf3714f3f077a9205dfa7ca8/odoo/fields.py#L4826-L4855
And it would be nice to update the PR message in order to render this part of the code nicely !

PR #246996, opened 2026-02-03
[IMP] l10n_tr_nilvera_einvoice: add new field for Nilvera PDFs

alialfie:
Before this commit, the PDF generated by Nilvera was stored as a generic attachment (message_main_attachment_id) on the invoice because a dedicated field could not be added in the stable version.

To find out which invoices still needed to download their Nilvera generatedPDF, a workaround was used in stable to avoid adding a field.

This commit adds a new dedicated field for the Nilvera PDF to replace that workaround.

Upgrade PR: https://github.com/odoo/upgrade/pull/9408

task-5904554

### DIFF
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move.py
@@ -5,7 +5,6 @@ from urllib.parse import quote, urlencode, urlparse
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import SQL
 from odoo.addons.l10n_tr_nilvera.const import NILVERA_ERROR_CODE_MESSAGES
 from odoo.addons.l10n_tr_nilvera.lib.nilvera_client import _get_nilvera_client
 
@@ -102,6 +101,17 @@ class AccountMove(models.Model):
     l10n_tr_nilvera_customer_status = fields.Selection(
         string="Partner Nilvera Status",
         related='partner_id.l10n_tr_nilvera_customer_status',
+    )
+    l10n_tr_nilvera_pdf_file = fields.Binary(
+        attachment=True,
+        string="Nilvera PDF File",
+        copy=False,
+    )
+    l10n_tr_nilvera_pdf_id = fields.Many2one(
+        comodel_name='ir.attachment',
+        string="Nilvera PDF Attachment",
+        compute=lambda self: self._compute_linked_attachment_id('l10n_tr_nilvera_pdf_id', 'l10n_tr_nilvera_pdf_file'),
+        depends=['l10n_tr_nilvera_pdf_file'],
     )
 
     @api.depends("l10n_tr_gib_invoice_scenario", "l10n_tr_gib_invoice_type", "l10n_tr_is_export_invoice")
@@ -411,11 +421,13 @@ class AccountMove(models.Model):
         attachment = self.env['ir.attachment'].create({
             'name': filename,
             'res_id': invoice.id,
+            'res_field': 'l10n_tr_nilvera_pdf_file',
             'res_model': 'account.move',
             'raw': base64.b64decode(response),
             'type': 'binary',
             'mimetype': 'application/pdf',
         })
+        self.invalidate_recordset(fnames=["l10n_tr_nilvera_pdf_id", "l10n_tr_nilvera_pdf_file"])
         # The created attachement coming form Nilvera should be the main attachment
         invoice.message_main_attachment_id = attachment
         invoice.with_context(no_new_invoice=True).message_post(attachment_ids=attachment.ids)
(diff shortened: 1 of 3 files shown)


addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -103,6 +103,18 @@ class AccountMove(models.Model):
...
+    l10n_tr_nilvera_pdf_file = fields.Binary(
+        attachment=True,
+        string="Nilvera PDF File",
+        copy=False,
+    )
+    l10n_tr_nilvera_pdf_id = fields.Many2one(
+        comodel_name='ir.attachment',
+        string="Nilvera PDF Attachment",
+        compute=lambda self: self._compute_linked_attachment_id('l10n_tr_nilvera_pdf_id', 'l10n_tr_nilvera_pdf_file'),
+        depends=['l10n_tr_nilvera_pdf_file']
```
```suggestion
        depends=['l10n_tr_nilvera_pdf_file'],
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -411,11 +423,13 @@ def _l10n_tr_nilvera_add_pdf_to_invoice(self, client, invoice, document_uuid, do
         attachment = self.env['ir.attachment'].create({
             'name': filename,
             'res_id': invoice.id,
+            "res_field": "l10n_tr_nilvera_pdf_file",
```
malb-odoo: Just to be consistent
```suggestion
            'res_field': 'l10n_tr_nilvera_pdf_file',
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -502,28 +516,12 @@ def _cron_nilvera_get_invoice_status(self):
...
-           LIMIT %s
-        """, batch_size)
-        move_ids = [row[0] for row in self.env.execute_query(sql)]
-        invoice_to_fetch_pdf = self.env['account.move'].browse(move_ids)
-        for company, invoices in invoice_to_fetch_pdf.grouped("company_id").items():
+        invoices_to_fetch_pdf = self.env['account.move'].search([
+            ('l10n_tr_nilvera_send_status', '=', 'succeed'),
+            ('move_type', '=', 'out_invoice'),
+            ('l10n_tr_nilvera_pdf_file', '=', False),
+        ], limit=batch_size)
```
malb-odoo, line 521: Just wondering if we should be scared of the perf of using a search on account move ? I suppose you need the whole record set so search_fetch is a no go, maybe a read group ? or keep the sql ?

malb-odoo (commented):
Thanks for the pr !

addons/l10n_tr_nilvera_einvoice/models/account_move.py
alialfie: Thanks for the review!

Performance wise, this is identical to how it was before so I don't think keeping the SQL is the right thing to do. Using the ORM is better and it will end up doing the same thing: writing a select query, joining the ir attachment table. This is one of the benefits of adding the field as well.
The browse before would've fetched the whole recordset as well so no change in performance there.
Not sure I see the point of read group though?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -411,11 +421,13 @@ def _l10n_tr_nilvera_add_pdf_to_invoice(self, client, invoice, document_uuid, do
         attachment = self.env['ir.attachment'].create({
             'name': filename,
             'res_id': invoice.id,
+            "res_field": 'l10n_tr_nilvera_pdf_file',
```
malb-odoo: Forgot to change :eyes: 
```suggestion
            'res_field': 'l10n_tr_nilvera_pdf_file',
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
malb-odoo: let's keep the search and we will see if a perf issue arise later on

malb-odoo (commented):
one small nitpick

malb-odoo (commented):
@robodoo r+

malb-odoo:
@robodoo r+

malb-odoo:
@alialfie News on this or we wait after the freeze ?

addons/l10n_tr_nilvera_einvoice/tests/test_tr_nilvera_mocked_requests.py
```diff
@@ -279,4 +279,5 @@ def test_fetching_einvoices(self, mocked_request):
 
             invoice = self.env['account.move'].search([('l10n_tr_nilvera_uuid', '=', 'invoice_uuid')])
             self.assertEqual(len(invoice), 1)
-            self.assertListEqual(sorted(invoice.attachment_ids.mapped('mimetype')), ['application/pdf', 'application/xml'])
+            self.assertListEqual(sorted(invoice.attachment_ids.mapped('mimetype')), ['application/xml'])
+            self.assertTrue(invoice.l10n_tr_nilvera_pdf_id)
```
alialfie, line 283: This was merged after I made the PR and before I rebased so it was failing. I adapted it.

alialfie (commented):
@malb-odoo Sorry, got distracted with other stuff 😬 Should be ready now!

malb-odoo:
@robodoo r+

PR #246998, opened 2026-02-03
[IMP] mass_mailing: Promen-Adductor

waterflow80:
## Description
This PR adds new features to the `mass_mailing` app.

## Added Features
The features are:
- [x] adding mailing Templates:
- [x] upgrade the `Mailing List` view (UI/UX)
- [x] disable link tracker on demand
- [x] `Start From Scratch` template: the `Drag a block...` button should open the block library
- [x] `View in Browser` should be a toggle 
- [x] support more fonts (implement a fallback mechanism)
- [x] upgrade `mailing.lists`: create `mailing.contacts` from `res_partners` (`Contacts` app).
- [ ] create `dynamic lists` 

---


addons/mass_mailing/models/link_tracker.py
```diff
@@ -15,6 +15,7 @@ class LinkTrackerClick(models.Model):
 
     mailing_trace_id = fields.Many2one('mailing.trace', string='Mail Statistics', index='btree_not_null')
     mass_mailing_id = fields.Many2one('mailing.mailing', string='Mass Mailing', index='btree_not_null')
+    email = fields.Char(string="Email", compute="_compute_clicker_email", readonly=True, store=True, help="The email address of the entity that clicked the link")
```
thjo-odoo: Couldn't this be a related field then? Or is there another specific reason you needed a compute on this one

addons/mass_mailing/models/mail_render_mixin.py
```diff
@@ -4,6 +4,43 @@
...
+    """Check if link tracking is enabled for the given link_node
+
+    :param lxml.etree._Element link_node: The link node of which to check the link tracking status
+
+    :rtype: bool
+    """
+    prop = link_node.attrib if link_node is not None else None
+    if prop and prop.get('data-no-tracking'):
+        return prop.get('data-no-tracking') != '1'
+    return True
```
thjo-odoo: Any specific reason those needed to be external to the class? Do you plan on using them elsewhere?

thjo-odoo: In my opinion you could even define them inside the new method you added to the model. So that they can only be used inside it.

addons/mass_mailing/models/mail_render_mixin.py
```diff
@@ -21,3 +58,10 @@ def _render_template_postprocess(self, model, rendered):
                     blacklist=['/unsubscribe_from_list', '/view', '/cards']
                 )
         return rendered
+
+    def _filter_urls_and_labels(self, link_nodes, urls_and_labels):
+        """Return only the links in which link tracking is enabled"""
+        tracking_enabled_status_list = list(map(is_link_tracking_enabled, link_nodes))  # --> [True, False, True, True, ...]
+        tracked_link_nodes, tracked_urls_and_labels = retrieve_tracked_links_with_urls_and_labels(link_nodes, urls_and_labels, tracking_enabled_status_list)
```
thjo-odoo: I even think that `retrieve_tracked_links_with_urls_and_labels` don't really need to exist and you could directly apply it here. As it's a pretty simple function in itself.

addons/mass_mailing/models/mailing.py
```diff
@@ -598,6 +599,36 @@ def copy_data(self, default=None):
...
 
+    def action_use_template(self):
+        """
+        <p>
+        Create a copy of the template and set it as a mailing (no longer a template) ready to be used.
+        </p>
+        <p>
+        The template is removed from Favorites to avoid duplicates in the Templates view, however the user
+        can set it back to Favorites so that it appears in the Templates view.
+        </p>"""
```
thjo-odoo: That's the 1st time I see a function's doc show what's the template returned. Was it more for you to easily visualize what would be returned when working, or is there another reason?

addons/mass_mailing/models/mailing.py
```diff
@@ -598,6 +599,36 @@ def copy_data(self, default=None):
...
+        if mass_mailing_copy := self._create_mailing_from_template(self):
+            res_context = dict(self.env.context)
+            res_context['default_is_template'] = 0
+            res_context['default_favorite'] = 0
```
thjo-odoo:
```suggestion
            res_context = dict(**self.env.context, default_is_template=0, default_is_favorite=0)
```

addons/mass_mailing/models/mailing.py
```diff
@@ -598,6 +599,36 @@ def copy_data(self, default=None):
...
+                'view_mode': 'form',
+                'res_model': 'mailing.mailing',
+                'res_id': mass_mailing_copy.id,
+                'context': res_context,
+                }
+        return False
+
+    def _create_mailing_from_template(self, mass_mailing_template):
+        mass_mailing_template.ensure_one()
+        if mass_mailing := mass_mailing_template.copy():
```
thjo-odoo: You can pass a default argument to copy so you could have
```suggestion
        if mass_mailing := mass_mailing_template.copy(default={"favorite": False, "is_template": False}):
```
And remove the value assignations inside the condition.

addons/mass_mailing/models/mailing.py
```diff
@@ -598,6 +599,36 @@ def copy_data(self, default=None):
...
+            return {
+                'type': 'ir.actions.act_window',
+                'view_mode': 'form',
+                'res_model': 'mailing.mailing',
+                'res_id': mass_mailing_copy.id,
+                'context': res_context,
+                }
+        return False
+
+    def _create_mailing_from_template(self, mass_mailing_template):
```
thjo-odoo, line 628: Don't think you need to pass it an argument, why not simply use self?

addons/mass_mailing/models/mailing.py
```diff
@@ -610,6 +641,27 @@ def action_duplicate(self):
...
+                'res_id': mass_mailing_copy.id,
+                'context': dict(self.env.context),
+            }
+        return False
+
+    def create_mailing_copy(self):
+        if mass_mailing_copy := self.copy():
+            mass_mailing_copy.subject += " - Copy"
+            if self.is_template:
+                mass_mailing_copy.favorite = True
```
thjo-odoo: I am kinda wondering: aren't templates always favorite then? Or is there something I don't understand.
Can you be a template that isn't a favorite mailing?

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+
+        model_name = self.env['ir.model']._get('mailing.trace').display_name
+        res_model = 'mailing.trace'
+        domain = None
+        view_mode = 'graph'
         helper_header = None
         helper_message = None
+        ctx = dict(self.env.context, create=False)
+        ctx["graph_mode"] = "line"
+        ctx["stacked"] = True
```
thjo-odoo:
```suggestion
        ctx = dict(self.env.context, create=False, graph_mode='line', stacked=True)
```

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+                context["search_default_groupby_email"] = True
+                context["graph_mode"] = "pie"
+                context["stacked"] = False
+            elif view_filter == 'open':
+                context.update({'search_default_filter_opened': False})
+                context["search_default_group_open_date"] = True
+            elif view_filter == 'delivered':
+                context.update({'search_default_filter_delivered': False})
+            elif view_filter == 'sent':
+                context.update({'search_default_filter_sent': False})
```
thjo-odoo: could be a match...case on filter. Also you use `filter` for the 1st condition then switch to view_filter any specific reason?

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+            'view_mode': view_mode,
+            'res_model': res_model,
+            'domain': domain,
+            'context': ctx,
         }
         if helper_header and helper_message:
             action['help'] = Markup('<p class="o_view_nocontent_smiling_face">%s</p><p>%s</p>') % (
                 helper_header, helper_message,
-            ),
+            )
```
thjo-odoo: Be aware that it could be an unwanted change here.

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+                context.update({'search_default_filter_opened': False})
+                context["search_default_group_open_date"] = True
+            elif view_filter == 'delivered':
+                context.update({'search_default_filter_delivered': False})
+            elif view_filter == 'sent':
+                context.update({'search_default_filter_sent': False})
+
+            return context
+
+        model_name = self.env['ir.model']._get('mailing.trace').display_name
```
thjo-odoo: And you could even not use the intermediary variable.
```suggestion
        model_name = self.env['ir.model']._get('mailing.trace' if view_filter != 'clicked' else 'link.tracker.click').display_name
```

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+        res_model = 'mailing.trace'
+        domain = None
+        view_mode = 'graph'
```
thjo-odoo: Same as before, you could also directly use it in the action dict lower
```suggestion
        view_mode = f'{"" if view_filter != "cliked" else "list,"}graph'
```

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
         helper_header = None
         helper_message = None
+        ctx = dict(self.env.context, create=False)
+        ctx["graph_mode"] = "line"
+        ctx["stacked"] = True
         if view_filter == 'reply':
-            res_ids = _fetch_trace_res_ids(Domain('trace_status', '=', 'reply'))
+            trace_ids = _fetch_trace_ids(Domain('trace_status', '=', 'reply'))
+            ctx = _update_context_for_filter(ctx, 'reply')
```
thjo-odoo: could be done once at the start depending on the view_filter.

thjo-odoo: We could even imagine something like this
```python
ctx = dict(self.env.context, 
    create=False, 
    graph_mode='line', 
    stacked=True, 
    _get_context_for_filter(view_filter))
```

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
+            ctx = _update_context_for_filter(ctx, 'bounce')
             helper_header = _("No %s address bounced yet!", model_name)
             helper_message = _("Bounce happens when a mailing cannot be delivered (fake address, "
                                "server issues, ...). Check each record to see what went wrong.")
         elif view_filter == 'clicked':
-            res_ids = _fetch_trace_res_ids(Domain('links_click_ids', '!=', False))
+            res_model = 'link.tracker.click'
+            model_name = self.env['ir.model']._get('link.tracker.click').display_name
+            trace_ids = _fetch_trace_ids(Domain('links_click_ids', '!=', False))
+            link_tracker_ids = _fetch_link_tracker_ids(Domain('mailing_trace_id', 'in', trace_ids))
```
thjo-odoo: Or simply
```suggestion
            trace_ids = _fetch_link_tracker_ids(Domain('mailing_trace_id', 'in', trace_ids))
```
and you always set the domain to [('id', 'in', trace_ids)] after the if/else tree.

addons/mass_mailing/models/mailing.py
```diff
@@ -713,68 +765,112 @@ def _action_view_traces_filtered(self, view_filter):
...
         elif view_filter == 'sent':
-            res_ids = _fetch_trace_res_ids(Domain('sent_datetime', '!=', False))
+            trace_ids = _fetch_trace_ids(Domain('sent_datetime', '!=', False))
+            ctx = _update_context_for_filter(ctx, 'sent')
         else:
-            res_ids = []
+            trace_ids = []
+
+        if not domain:
+            domain = [('id', 'in', trace_ids)]
```
thjo-odoo: As said earlier you could always set domain to `[('id', 'in', trace_ids)]` and update trace_ids even when you are handling 'clicked'

addons/mass_mailing/static/src/builder/plugins/empty_mailing_plugin.js
```diff
@@ -10,35 +9,26 @@ export class EmptyMailingPlugin extends Plugin {
...
         "history",
         "selection",
+        "blockTab",
     ];
 
-    resources = {
-        on_beforeinput_handlers: withSequence(1, this.ensureTextBlock.bind(this)),
-    };
+    setup() {
+        this.addDomListener(this.editable, "click", this.onMailWrapperClick);
```
thjo-odoo: Don't forget to bind the function, otherwise you might have unwanted behaviors.

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mailing_link_popover.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<templates xml:space="preserve">
+    <t t-name="mass_mailing.MailingLinkPopover"
+       t-inherit="html_editor.linkPopover"
+       t-inherit-mode="primary">
+        <xpath expr="//div[contains(@class, 'o_advance_options_grid')]" position="inside">
```
thjo-odoo: We usually use the `hasclass` function when you want to xpath on an element that has a specific class.

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mass_mailing_link_popover.js
```diff
@@ -0,0 +1,53 @@
...
+export class MassMailingLinkPopover extends LinkPopover {
+    static template = "mass_mailing.MailingLinkPopover";
+    static props = {
+        ...LinkPopover.props,
+    };
+    setup() {
+        super.setup();
+        this.linkElement = this.props.linkElement;
+        this.notrackingStatus = this.linkElement.dataset.noTracking || "0";
+        this.state.notracking = {
```
thjo-odoo: So that the camelCase is respected.
```suggestion
        this.noTrackingStatus = this.linkElement.dataset.noTracking || "0";
        this.state.noTracking = {
```

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mass_mailing_link_popover.js
```diff
@@ -0,0 +1,53 @@
...
+            label: "Disable Link Tracking",
+            description: _t("Send the orignal url instead of wraping it into a tracking url"),
+            isChecked: this.notrackingStatus == "1",
+        };
+    }
+
+    toggleDisableLinkTracking() {
+        this.notrackingStatus = this.notrackingStatus == "1" ? "0" : "1";
+        this.state.notracking.isChecked = !this.state.notracking.isChecked;
```
thjo-odoo: I won't say for all of them but don't forget to update all the noTracking in the file.

addons/mass_mailing/static/src/editor/plugins/mass_mailing_link_plugin.js
```diff
@@ -0,0 +1,34 @@
+import { withSequence } from "@html_editor/utils/resource";
+import { registry } from "@web/core/registry";
+import { MassMailingLinkPopover } from "../../components/mass_mailing_link_popover/mass_mailing_link_popover";
+import { LinkPlugin } from "@html_editor/main/link/link_plugin";
+
+export class MassMailingLinkPlugin extends LinkPlugin {
+    static id = "link";
```
thjo-odoo: I don't think that you need to reassign the id as you extend the original plugin.

addons/mass_mailing/static/src/views/fields/mailing_percentage_field.js
```diff
@@ -0,0 +1,35 @@
...
+            getValue: () =>
+                formatMailingPercentage(this.props.record.data[this.props.name], {
+                    digits: this.props.digits,
+                    noSymbol: true,
+                    field: this.props.record.fields[this.props.name],
+                }),
+            refName: "numpadDecimal",
+            parse: (v) => parsePercentage(v),
+        });
+    }
```
thjo-odoo, line 20: The call to `useInputField` seems to be the same as in the original class. If that is the case no need for an override.

thjo-odoo (commented):
@waterflow80 
Hello, 
Thanks for your work on your 1st big task! 
As I didn't have time to finish I'll already send you some changes today and I'll finish this tomorrow.
Either way you will have some things to already do for it.
If something is unclear, let me know and I'll answer your questions.
Cheers!

addons/mass_mailing/models/link_tracker.py
waterflow80: You're right, `computed=` would be a better fit here.

addons/mass_mailing/models/mail_render_mixin.py
waterflow80: Correct!

addons/mass_mailing/models/mailing.py
waterflow80: I just wanted to give explanation for the reader about what the `Use this` button do (must do). But, I agree that it's unnecessary.
Removed!

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mass_mailing_link_popover.js
waterflow80: (Also done for the xml template)

addons/mass_mailing/static/src/views/fields/mailing_percentage_field.js
waterflow80: We actually changed `formatPercentage(...)` to `formatMailingPercentage(...)` to introduce our formatting logic.

addons/mass_mailing/models/mail_render_mixin.py
```diff
@@ -21,3 +21,28 @@ def _render_template_postprocess(self, model, rendered):
...
+            """Check if link tracking is enabled for the given link_node
+
+            :param lxml.etree._Element link_node: The link node of which to check the link tracking status
+
+            :rtype: bool
+            """
+            prop = link_node.attrib if link_node is not None else None
+            if prop and prop.get('data-no-tracking'):
+                return prop.get('data-no-tracking') != '1'
+            return True
```
thjo-odoo, line 38: Is the case `link_node is None` possible in the array you give to `_filter_urls_and_labels`?

addons/mass_mailing/views/mailing_templates_views.xml
```diff
@@ -0,0 +1,155 @@
...
+                <chatter/>
+            </form>
+        </field>
+    </record>
+
+    <record id="mailing_templates_action" model="ir.actions.act_window">
+        <field name="name">Templates</field>
+        <field name="res_model">mailing.mailing</field>  
+        <field name="view_mode">kanban,list,form</field>
+        <field name="view_ids" eval="[(5, 0, 0),
```
thjo-odoo, line 142: ~Why the (5, 0, 0)? You are technically creating a new record so no real need for it.~

Edit: it seems everyone is doing this so you can pass this one.

addons/mass_mailing/views/mailing_templates_views.xml
```diff
@@ -0,0 +1,155 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Templates -->
+
```
thjo-odoo: Not necessary.

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mailing_link_popover.xml
```diff
@@ -0,0 +1,18 @@
...
+    <t t-name="mass_mailing.MailingLinkPopover"
+       t-inherit="html_editor.linkPopover"
+       t-inherit-mode="primary">
+        <xpath expr="//div[hasclass('o_advance_options_grid')]" position="inside">
+            <t t-if="shouldDisplayNoTrackingOption()">
+                <div class="o_seo_option_row d-flex justify-content-between align-items-center py-1 gap-2">
+                    <span t-att-title="state.noTracking.description"><t t-out="state.noTracking.label"/></span>
+                    <CheckBox
+                        value="state.noTracking.isChecked"
+                        onChange="toggleDisableLinkTracking.bind(this)"/>
```
thjo-odoo: In OWL templates there is already a mechanism that allows you to do such a thing.
```suggestion
                        onChange.bind="toggleDisableLinkTracking"/>
```

addons/mass_mailing/static/src/editor/plugins/mass_mailing_link_plugin.js
```diff
@@ -0,0 +1,33 @@
...
+                isAvailable: (linkEl) => !linkEl || !this.isLinkImmutable(linkEl),
+                getProps: (props) => props,
+            }),
+        ],
+    };
+
+    getProps(originalArgs) {
+        const { linkElement, applyCallback } = originalArgs;
+        const applyCallbackExtended = (...args) => {
+            const [url, label, classes, linkTarget, attachmentId, relValue, notrackingVal] = args;
```
thjo-odoo:
```suggestion
            const [url, label, classes, linkTarget, attachmentId, relValue, noTrackingVal] = args;
```

thjo-odoo: And if you want you could arrange your arguments such that noTrackingVal is the 1st value in the array.
This way you can extract the other arguments like such.
```js
const [noTrackingVal, ...otherArgs] = args;
applyCallback(...otherArgs);
```

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mass_mailing_link_popover.js
```diff
@@ -0,0 +1,53 @@
...
+
+export class MassMailingLinkPopover extends LinkPopover {
+    static template = "mass_mailing.MailingLinkPopover";
+    static props = {
+        ...LinkPopover.props,
+    };
+    setup() {
+        super.setup();
+        this.linkElement = this.props.linkElement;
+        this.noTrackingStatus = this.linkElement.dataset.noTracking || "0";
```
thjo-odoo: Wouldn't it be simpler to store a boolean instead of a string? Then in the apply function based on its value then you set the string.

addons/mass_mailing/static/src/components/mass_mailing_link_popover/mass_mailing_link_popover.js
```diff
@@ -0,0 +1,53 @@
+import { LinkPopover } from "@html_editor/main/link/link_popover";
+import { _t } from "@web/core/l10n/translation";
+
+/**
+ * List of paths the URL of which will not have any effect when using the `notracking` option.
+ *
+ * @constant {string[]}
+ */
+export const NON_TRACKABLE_URLS = ["/unsubscribe_from_list", "/view", "/cards"];
```
thjo-odoo: You don't seem to use it anywhere else. Do you plan to update this one in the future? If you're not sure IMO it's best to not export it right now and do it when it becomes necessary.

addons/mass_mailing/static/src/views/fields/mailing_percentage_field.js
thjo-odoo: Indeed I didn't see that.

addons/mass_mailing/tests/test_mail_render_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+            <div>
+                <a href="https://example.com/page1" data-no-tracking='1'>Link 1</a>
+                <a href="https://example.com/page2" data-no-tracking='1'>Link 2</a>
+            </div>
+        """
+        link_tracker_vals = {
+            'mass_mailing_id': 1,
+            'utm_reference': 'mailing.mailing,1',
+            'source_id': 2,
+            'medium_id': 4,
```
thjo-odoo: Would prefer if those weren't static values => you could store the mailing created before into variables and use those here. 
Same thing for the source_id and medium_id, if those id are referenced somewhere it's better to use the ref or the record.

addons/mass_mailing/tests/test_mail_render_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        # Execute
+        result_html_tracked = self.env["mailing.mailing"]._shorten_links(html_links_tracked, link_tracker_vals)
+        result_html_untracked = self.env["mailing.mailing"]._shorten_links(html_links_not_tracked, link_tracker_vals)
```
thjo-odoo: It's better to use the actual model where the method is defined.
```suggestion
        result_html_tracked = self.env["mail.render.mixin"]._shorten_links(html_links_tracked, link_tracker_vals)
        result_html_untracked = self.env["mail.render.mixin"]._shorten_links(html_links_not_tracked, link_tracker_vals)
```

addons/mass_mailing/tests/test_mail_render_mixin.py
```diff
@@ -0,0 +1,63 @@
+import lxml.html
```
thjo-odoo: And change their usage.
```suggestion
from lxml import html
```

addons/mass_mailing/tests/test_mail_render_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+
+from odoo.addons.mass_mailing.tests.common import MassMailCommon
+from odoo.tests import tagged
+
+
+@tagged('at_install', '-post_install')  # LEGACY at_install
+class TestMailRenderMixin(MassMailCommon):
+
+    def setUp(self):
+        super().setUp()
```
thjo-odoo: Define `setUpClass` instead as it's what's used in other test classes.
Don't forget to add the `@classmethod` decorator

addons/mass_mailing/tests/test_mailing_templates.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Execute
+        res_action = mailing_template.action_use_template()
+        created_mailing_id = res_action['res_id']
+
+        # Assert
+        self.assertEqual("attribute value", res_action['context']['some_attribute'])
+        self.assertEqual(False, res_action['context']['default_is_template'])
+        self.assertEqual(False, res_action['context']['default_favorite'])
+        self.assertEqual(2, self.env['mailing.mailing'].search_count([]))
```
thjo-odoo: Don't think you really need this assert

addons/mass_mailing/tests/test_mailing_templates.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertEqual(False, self.env['mailing.mailing'].search([('id', '=', created_mailing_id)]).favorite)
+        self.assertEqual(False, self.env['mailing.mailing'].search([('id', '=', created_mailing_id)]).is_template)
```
thjo-odoo: You could use browse instead since you know the ID and don't need to check that you have access to it

addons/mass_mailing/models/mailing.py
```diff
@@ -610,6 +631,27 @@ def action_duplicate(self):
...
+                'type': 'ir.actions.act_window',
+                'view_mode': 'form',
+                'res_model': 'mailing.mailing',
+                'views': [(form_view_ref, 'form')],
+                'res_id': mass_mailing_copy.id,
+                'context': dict(self.env.context),
+            }
+        return False
+
+    def create_mailing_copy(self):
```
thjo-odoo: If you intend to only use this method within the context of specific actions let's set it as a private method
```suggestion
    def _create_mailing_copy(self):
```
This way it's not possible to call the method via RPC

addons/mass_mailing/models/mailing.py
```diff
@@ -610,6 +631,27 @@ def action_duplicate(self):
...
+        if mass_mailing_copy := self.copy():
+            mass_mailing_copy.subject += " - Copy"
+            if self.is_template:
+                mass_mailing_copy.favorite = True
```
thjo-odoo: Same as before, you can set default so that you don't need to afterwards set the value you wanted on the record.
```suggestion
        if mass_mailing_copy := self.copy(default={"subject": f"{self.subject} - Copy"}, "favorite": self.is_template):
```

addons/mass_mailing/models/mailing.py
```diff
@@ -598,6 +599,26 @@ def copy_data(self, default=None):
...
+    def action_use_template(self):
+        if mass_mailing_copy := self._create_mailing_from_template(self):
+            res_context = dict(**self.env.context, default_is_template=0, default_favorite=0)
+            return {
+                'type': 'ir.actions.act_window',
+                'view_mode': 'form',
+                'res_model': 'mailing.mailing',
+                'res_id': mass_mailing_copy.id,
+                'context': res_context,
+                }
```
thjo-odoo:
```suggestion
            }
```

addons/mass_mailing/tests/test_mail_render_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+        for link in links:
+            self.assertIn('/r/', link.get("href"))
+
+    def _assert_all_untracked(self, links):
+        link_1_href = links[0].get("href")
+        link_2_href = links[1].get("href")
+        self.assertNotIn('/r/', link_1_href)
+        self.assertNotIn('/r/', link_2_href)
+        self.assertEqual('https://example.com/page1', link_1_href)
+        self.assertEqual('https://example.com/page2', link_2_href)
```
thjo-odoo: Why handle the asserts 2 different way? Seems you could've fused both of them and have 2 arguments (tracked_links, untracked_links) and the method checks the corresponding value for each array.

addons/mass_mailing/tests/test_mailing_internals.py
```diff
@@ -837,5 +837,89 @@ def test_mailing_action_open(self):
...
+        ])
+        link_1 = self.env["link.tracker"].create({
+            "url": "https://example.com/1",
+        })
+        link_2 = self.env["link.tracker"].create({
+            "url": "https://example.com/2",
+        })
+        link_3 = self.env["link.tracker"].create({
+            "url": "https://example.com/3",
+        })
```
thjo-odoo:
```suggestion
        link_1, link_2, link_3 = self.env["link.tracker"].create([
            {"url": f"https://example.com/{i}"} for i in range(1,4)
        ])
```

addons/mass_mailing/models/mailing.py
```diff
@@ -610,6 +631,27 @@ def action_duplicate(self):
...
+                'view_mode': 'form',
+                'res_model': 'mailing.mailing',
+                'views': [(form_view_ref, 'form')],
+                'res_id': mass_mailing_copy.id,
+                'context': dict(self.env.context),
+            }
+        return False
+
+    def create_mailing_copy(self):
+        if mass_mailing_copy := self.copy():
```
thjo-odoo: I am starting to wonder: is it a good idea to have an action that directly creates a copy of the mailing? Wouldn't it be better to have the action open a form view with all the same field values but let the user create the record or not?
I am concerned that you could end up with users that will create hundreds of records too easily.

thjo-odoo (changes requested):
@waterflow80 
Hello,
Here's some of the second part of the validation for this. I stopped it a bit early as I had some concerns with your actions. You are directly creating records when using the template or duplicating a template. 
That may be a concern if a user calls this action multiple times in a row => RPC calls without using the action.
 Is it something that was wanted in the spec?

addons/mass_mailing/models/mailing.py
waterflow80: I'm actually using this method from `marketing_automation.marketing_activity` too. Maybe we can move it to a common place?

addons/mass_mailing/models/mailing.py
waterflow80: You're correct, however the `favorite` field has `copy=False` attribute, so when copying a template (which is a mailing) the `copy` field is not copied.

addons/mass_mailing/models/mailing.py
thjo-odoo: Discussed this one with Luc and it seems that he's aware and okay with the current implementation. 
Since it's simpler let's keep it like this.

addons/mass_mailing/models/mailing.py
waterflow80: Ohh nice catch, that's a mistake. It should be `filter`. I also think that we can get rid of the `filter` parameter and only use the `view_filter` since it's in its scope.

addons/mass_mailing/views/mailing_mailing_views.xml
```diff
@@ -184,7 +184,7 @@
                                 id="button_view_delivered"
                                 type="object"
                                 context="{'search_default_filter_delivered': True}"
-                                invisible="state in ('draft', 'test')"
+                                invisible="1"
```
thjo-odoo, line 187: Couldn't you remove this button then? Since you are in master you can remove buttons from the views.

addons/mass_mailing/models/mailing.py
waterflow80: I actually changed this into
```py
{**self.env.context, 'default_is_template': 0, 'default_favorite': 0}
```
because `dict(...)` expects no duplicate values, and with `{...}`, the last value present will override the previous ones.

addons/mass_mailing/models/mailing.py
waterflow80: Used `{..}` instead of `dict(..)` to able to override existing values of the context.

addons/mass_mailing/models/mail_render_mixin.py
waterflow80: Normally not, but I wanted to make the access to the `link_node._some_attribute` safe, to avoid any potential `AttributeError` in the future.

addons/mass_mailing/views/mailing_mailing_views.xml
waterflow80: Actually this button is referenced by an `xpath` call from `mass_mailing_crm.mailing_mailing_views.xml`:
```xml
<xpath expr="//button[@id='button_view_delivered']" position="before">
```

PR #247001, opened 2026-02-03
[FIX] project: portal user should be able to close recurrent tasks

abla001:
to reproduce:
=============
- create a project with recurrent task and share it with a portal user
- log in with the portal user and open the recurrent task
- change the state to 'Done' and save -> the state is not changed to 'Done' and the task is not closed

problem:
========
the method responsible for creating the next occurrence of a recurrent task executes some operations that portal users are not allowed to do, causing the whole operation to fail and preventing the task from being updated.

solution:
=========
use sudo() to bypass access rights checks when creating the next occurrence of the recurrent task. This is safe because the operations being performed do not involve any sensitive data or actions that could compromise security.

opw-5442919

---


xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #247003, opened 2026-02-03
[IMP] http: move session methods

thle-odoo:
The session is a critical object.
The goal of this commit is to limit its exposure.

Developers should use the session as a dictionary (`MutableMapping`).
Developers should (generally) not use internal logic that processes the
session.

To achieve this, this commit moves the methods from the `Session` class
to make them module-level functions.

Task-5882018

### DIFF
--- a/odoo/http/router.py
+++ b/odoo/http/router.py
@@ -254,7 +254,7 @@ class Application:
                         _logger.warning("Database or registry unusable, trying without", exc_info=e.__cause__)
                         # TODO: move those bits in a dedicated function
                         request.db = None
-                        request.session.logout()
+                        logout(request.session)
                         if (httprequest.path.startswith('/odoo/')
                             or httprequest.path in (
                                 '/odoo', '/web', '/web/login', '/test_http/ensure_db',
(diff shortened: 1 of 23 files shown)


thle-odoo:
Enterprise PR: https://github.com/odoo/enterprise/pull/106330

odoo/http/router.py
```diff
@@ -307,4 +307,4 @@ def fake_start_response(status, headers):  # noqa: E301, E306
     request,
 )
 from .routing_map import ROUTING_KEYS, _generate_routing_rules
-from .session import SessionExpiredException, SessionStore
+from .session import logout, SessionExpiredException, SessionStore
```
Julien00859: https://docs.astral.sh/ruff/rules/unsorted-imports/
```suggestion
from .session import SessionExpiredException, SessionStore, logout
```

odoo/http/session.py
```diff
@@ -1,10 +1,13 @@
...
+
 import glob
 import json
 import logging
 import os
 import re
 import secrets
 import tempfile
+import typing
 import time
```
Julien00859: https://docs.astral.sh/ruff/rules/unsorted-imports/
```suggestion
import time
import typing
```

Julien00859 (approved):
@robodoo delegate+

Julien00859:
@rco-odoo can we have a ci/style override plz?

PR #247010, opened 2026-02-03
[IMP] point_of_sale, product, website_sale: migrate base unit model

lugr-odoo:
### Part 2: odoo/odoo#250891

We move up the base unit model from `website_sale` to `product` proper so that other modules like `point_of_sale` can use it without having `website_sale` installed.

Task ID: [4277792](https://www.odoo.com/odoo/my-tasks/4277792)


addons/point_of_sale/models/res_config_settings.py
```diff
@@ -131,6 +131,7 @@ def _default_pos_config(self):
     pos_basic_receipt = fields.Boolean(related='pos_config_id.basic_receipt', readonly=False)
     pos_fallback_nomenclature_id = fields.Many2one(related='pos_config_id.fallback_nomenclature_id', domain="[('id', '!=', barcode_nomenclature_id)]", readonly=False)
     group_pos_preset = fields.Boolean(string="Presets", implied_group="point_of_sale.group_pos_preset", help="Hide or show the Presets menu in the Point of Sale configuration.")
+    group_show_uom_price = fields.Boolean(string="Base Unit Price", default=False, implied_group="product.group_show_uom_price")
```
Feyensv: Why do you need to redefine the settings field here?
Isn't it already available as `point_of_sale` depends on `product` ?

addons/website_sale/security/ir.model.access.csv
```diff
@@ -59,10 +59,6 @@ access_ecom_extra_fields_public_public,access_ecom_extra_field public,model_webs
 access_ecom_extra_fields_public_portal,access_ecom_extra_field public,model_website_sale_extra_field,base.group_portal,1,0,0,0
 access_ecom_extra_fields_public_employee,access_ecom_extra_field public,model_website_sale_extra_field,base.group_user,1,0,0,0
 access_ecom_extra_fields_restricted_editor,access_ecom_extra_field restricted_editor,model_website_sale_extra_field,website.group_website_restricted_editor,1,1,1,1
-access_website_base_unit_public_public,website.base.unit public,model_website_base_unit,base.group_public,1,0,0,0
-access_website_base_unit_public_portal,website.base.unit public,model_website_base_unit,base.group_portal,1,0,0,0
-access_website_base_unit_public_employee,website.base.unit public,model_website_base_unit,base.group_user,1,0,0,0
-access_website_base_unit_sale_manager,website.base.unit sale manager,model_website_base_unit,sales_team.group_sale_manager,1,1,1,1
```
Feyensv, line 65: I believe only the `base.group_user` rule should be moved to `product`.
The base unit logic has no impact on the `sale` module, there is no need to give access to sale manager in the sale module.
Same for portal & public rules, unless you really need those for pos, I think they should be kept here.

addons/product/models/res_config_settings.py
```diff
@@ -7,6 +7,8 @@ class ResConfigSettings(models.TransientModel):
     _inherit = 'res.config.settings'
 
     group_uom = fields.Boolean("Units of Measure & Packagings", implied_group='uom.group_uom')
+    group_show_uom_price = fields.Boolean(string="Base Unit Price", default=False,
+                                          group='base.group_user', implied_group="product.group_show_uom_price")
```
Feyensv:
```suggestion
    group_show_uom_price = fields.Boolean(string="Base Unit Price", implied_group="product.group_show_uom_price")
```
The default is meaningless on a settings field computed to match a feature group, there is no need to keep it here ;).

Feyensv (commented):
As the PR is only about moving stuff around, I only commented what seemed unexpected/unnecessary, not about improving the existing logic itself ;).

addons/website_sale/security/ir.model.access.csv
lugr-odoo: I put them back (but still renamed `*_{website,product}_base_unit_*`).

addons/point_of_sale/models/res_config_settings.py
lugr-odoo: Taken out. 🫡

addons/product/models/res_config_settings.py
lugr-odoo: Done!

addons/product/security/ir.model.access.csv
```diff
@@ -35,3 +35,5 @@ access_update_product_attribute_value_manager,update.product.attribute.value,mod
 access_product_uom_manager,product.uom manager,model_product_uom,group_product_manager,1,1,1,1
 access_product_uom_user,product.uom user,model_product_uom,base.group_user,1,0,0,0
 access_uom_uom_product_manager,uom.uom.product.manager,model_uom_uom,group_product_manager,1,1,1,1
+access_product_base_unit_public_public,product.base.unit public,model_product_base_unit,base.group_public,1,0,0,0
```
Feyensv: Not for internal users here (`base.group_user`) ?
Are there public users used in pos flows ?

lugr-odoo: I moved back the wrong one. 🤦

Thank you for catching it.

Feyensv (approved):

PR #247011, opened 2026-02-03
[FIX] point_of_sale: correctly post COGS

robinengels:
When the product category valuation is set to `real_time` but the
company is set to `periodic`, the POS session closing was not posting
COGS entries as expected. If the product category valuation is set to
`real_time`, it should always prevail over the company setting. And if
no valuation is set on the product category, then the company setting
should be used.

Steps to reproduce:
-------------------
* Create a product category with `Inventory Valuation` set to
  `real_time` (Perpetual).
* Create a product in this category and make sure it is storable and has
  a cost price.
* Set the company `Inventory Valuation` to `periodic` (Periodic).
* Create a POS order with this product and pay it.
* Close the POS session.
> Observation: In the session no COGS entries are created for the sold
  product.

Why the fix:
------------
We adapt the `_search_valuation` to correctly fallback on the company
setting only if the product category valuation is not set. It was not
working before because in some cases the product had no company set
(it means that it is visible to all companies) and the domain was not
matching. To fix that we check that the current company matches the
search value, and if it does we also match all the products without
company set.

We also adapt the PoS code to use the `product_id.valuation` field to
filter all the stock_moves that should create COGS entries when
closing the session.

opw-5885960

### DIFF
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -46,6 +46,9 @@ class ProductTemplate(models.Model):
             raise UserError(self.env._("Only the value 'periodic' and 'real_time' are accepted to search on valuation field."))
         domain_categ = Domain([('categ_id.property_valuation', operator, value)])
         domain_company = Domain(['|', ('categ_id.property_valuation', '=', False), ('categ_id', '=', False), ('company_id.inventory_valuation', operator, value)])
+
+        if self.env.company.inventory_valuation and self.env.company.inventory_valuation == value:
+            domain_company = Domain(['|', ('categ_id.property_valuation', '=', False), ('categ_id', '=', False), '|', ('company_id.inventory_valuation', operator, value), ('company_id', '=', False)])
         return domain_company | domain_categ
 
     @api.depends('tracking')
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -1001,7 +1001,7 @@ class PosOrder(models.Model):
                         'display_type': 'rounding',
                     })
         # Stock.
-        if self.company_id.inventory_valuation == 'real_time' and self.picking_ids.ids:
+        if self.picking_ids.ids:
             stock_moves = self.env['stock.move'].sudo().search([
                 ('picking_id', 'in', self.picking_ids.ids),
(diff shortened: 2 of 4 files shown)


addons/stock_account/models/product.py
robinengels: Hello @ticodoo, could I have a review for this part of the code ? Thanks !

robinengels, line 1: Hello @odoo/rd-logistics could one of you have a look at this part of the fix ? Thanks !

addons/stock_account/models/product.py
```diff
@@ -46,7 +46,11 @@ def _search_valuation(self, operator, value):
             raise UserError(self.env._("Only the value 'periodic' and 'real_time' are accepted to search on valuation field."))
         domain_categ = Domain([('categ_id.property_valuation', operator, value)])
         domain_company = Domain(['|', ('categ_id.property_valuation', '=', False), ('categ_id', '=', False), ('company_id.inventory_valuation', operator, value)])
-        return domain_company | domain_categ
+
+        res_domain = domain_company | domain_categ
+        if self.env.company.inventory_valuation and self.env.company.inventory_valuation == value:
+            res_domain |= Domain(['|', ('categ_id', '=', False), ('categ_id.property_valuation', '=', False), ('company_id', '=', False)])
+        return res_domain
```
svs-odoo: Hello @robinengels 
Seems good to me, that said I would rather redefine `domain_company` so we don't have a domain with duplicate part
```suggestion
        if self.env.company.inventory_valuation and self.env.company.inventory_valuation == value:
            domain_company = Domain(['|', ('categ_id.property_valuation', '=', False), ('categ_id', '=', False), '|', ('company_id.inventory_valuation', operator, value), ('company_id', '=', False)])
        return domain_company | domain_categ
```

robinengels: @svs-odoo I applied the changes you suggested. Should I ask a review from the PoS team ? It's actually only stock related changes

svs-odoo (approved):
@robinengels The done change seems logical and right to me but while being thigh to stock related business (picking, stock move), it's `point_of_sale` code, so yep, you can ask a review from PoS team, I approve the change but won't merge this PR.

adgu-odoo (approved):

PR #247015, opened 2026-02-03
[FIX] point_of_sale: archival of journals

vlst-odoo:
Currently, in pos, we only prevent archiving `account.journal` records when a payment in an active pos session is linked to that specific journal.

This can lead to issues if a journal is archived while being linked to a `pos.payment.method`.

In this commit we prevent that from happening, by only allowing the archival of `account.journal` records when they are not linked to `pos.payment.method`.

Steps to reproduce:
- in payment method configuration set the journal bank on your card payment method
- Link this payment method to your POS
- Archive the bank journal in the backend
- Open POS
- Make a sale
- Close register
- You're stuck because the POS cannot input the journal item into the archived journal

Task 5896123

---


davidmonnom (approved):
Nice

PR #247025, opened 2026-02-03
[FIX] account_edi_ubl_cii: Add commodity codes from intrastat/unspsc/cpv

smetl:
task-5890887

---


clbr-odoo (approved):
Flawless, thanks ! 
@robodoo r+

PR #247028, opened 2026-02-03
[FIX] point_of_sale:  print combo choice items to preparation printers

stva-odoo:
Previously, combo choice items with categories assigned to a preparation printer were skipped when their category differed from the combo parent product’s category. This commit ensures that both the combo parent product’s category and the item product’s category are taken into account to correctly print items to the preparation printer.

Task: 5902389

---


davidmonnom (approved):
@robodoo r+

PR #247029, opened 2026-02-03
[FIX] point_of_sale: prevent setting non available pricelist on order

pedrambiria:
When changing the customer on a POS order, if the customer's pricelist is not in the list of available pricelists for the POS, but the pricelist was loaded due to loading a paid order, the POS would still set that pricelist on the order.

opw-5461556

---


adgu-odoo (commented):
robodoo r+

PR #247030, opened 2026-02-03
[FIX] mail: no composer local storage save on empty

alexkuhn:
Before this commit, composer content was saved in local storage even when the content is empty.

This happens because while there's code to clear local storage entry on message post, the content was necessarily saved from debounced or when triggering an action that unmounts the composer of thread, like a change of active thread in discuss.

This commit fixes the issue by checking that composer has some content, to determine whether the content must be saved in local storage or explicitly removed.

Task-5905834

### DIFF
--- a/addons/mail/static/src/core/common/composer.js
+++ b/addons/mail/static/src/core/common/composer.js
@@ -918,14 +918,20 @@ export class Composer extends Component {
             emailAddSignature,
             replyToMessageId,
         }) => {
-            browser.localStorage.setItem(
-                composer.localId,
-                JSON.stringify({
-                    emailAddSignature,
-                    replyToMessageId,
-                    composerHtml: isMarkup(composerHtml) ? ["markup", composerHtml] : composerHtml,
-                })
-            );
+            if (isHtmlEmpty(composerHtml)) {
+                browser.localStorage.removeItem(composer.localId);
+            } else {
+                browser.localStorage.setItem(
+                    composer.localId,
+                    JSON.stringify({
+                        emailAddSignature,
+                        replyToMessageId,
+                        composerHtml: isMarkup(composerHtml)
+                            ? ["markup", composerHtml]
+                            : composerHtml,
+                    })
+                );
+            }
         };
         if (this.state.isFullComposerOpen) {
             this.fullComposerBus.trigger("SAVE_CONTENT", {
--- a/addons/mail/static/tests/composer/composer.test.js
+++ b/addons/mail/static/tests/composer/composer.test.js
@@ -1470,6 +1470,7 @@ test("composer reply-to message is restored on thread change", async () => {
     await click(".o-mail-Message [title='Expand']");
     await click(".o-dropdown-item:contains('Reply')");
     await contains(".o-mail-Composer:contains('Replying to')");
+    await insertText(".o-mail-Composer-input", "Hello World!");
     await click(".o-mail-DiscussSidebar-item:contains('Inbox')");
     await contains(".o-mail-Message", { count: 0 });
     await click(".o-mail-DiscussSidebar-item:contains('General')");
@@ -1481,8 +1482,25 @@ test("composer reply-to message is restored on thread change", async () => {
             store.Thread.get({ model: "discuss.channel", id: channelId }).composer.localId
         )
     ).toBe(
-        '{"emailAddSignature":true,"replyToMessageId":1,"composerHtml":["markup","<div class=\'o-paragraph\'><br></div>"]}'
+        '{"emailAddSignature":true,"replyToMessageId":1,"composerHtml":["markup","Hello World!"]}'
     );
+    // check local storage emptied on message post
+    await click(".o-mail-Composer button:enabled[aria-label='Send']");
+    await click(".o-mail-Message[data-persistent]:contains(Hello World!)");
+    expect(
+        browser.localStorage.getItem(
+            store.Thread.get({ model: "discuss.channel", id: channelId }).composer.localId
+        )
+    ).toBe(null);
+    // check local storage empty, change thread to force save in local storage if needed (debounced otherwise)
+    await click(".o-mail-DiscussSidebar-item:contains('Inbox')");
+    await contains(".o-mail-Message", { count: 0 });
+    await click(".o-mail-DiscussSidebar-item:contains('General')");
+    expect(
+        browser.localStorage.getItem(
+            store.Thread.get({ model: "discuss.channel", id: channelId }).composer.localId
+        )
+    ).toBe(null);
 });
 
 test("composer reply-to message is restored page reload", async () => {
@@ -1518,7 +1536,7 @@ test("composer reply-to message is restored page reload", async () => {
     // not taking last message as to not fetch last message data prematurely
     browser.localStorage.setItem(
         `Composer,(Thread,discuss.channel AND ${channelId}) OR (undefined)`,
-        `{"replyToMessageId":${messageId_1},"text":""}`
+        `{"replyToMessageId":${messageId_1},"text":"some text"}`
     );
     await start();
     await openDiscuss(channelId);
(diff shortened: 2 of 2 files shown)


addons/mail/static/src/core/common/composer.js
```diff
@@ -918,14 +918,20 @@ export class Composer extends Component {
...
+                    composer.localId,
+                    JSON.stringify({
+                        emailAddSignature,
+                        replyToMessageId,
+                        composerHtml: isMarkup(composerHtml)
+                            ? ["markup", composerHtml]
+                            : composerHtml,
+                    })
+                );
+            }
```
seb-odoo: I don't see how `emailAddSignature` or `replyToMessageId` are relevant to call `removeItem`.

These are only additional info to save together with the composer, but the composer state should fully determine what has to be done.

alexkuhn: I assume when they differ from default value we may want to save in local storage. `emailAddSignature` I don't know the flow, reply-to sure we can save only when there's content.

alexkuhn: Change done, to just rely on `isHtmlEmpty(composerHtml)`.

seb-odoo: For `emailAddSignature` we need to keep track if the signature was automatically added inside the body or not, as we automatically add it sometimes, to avoid adding it again a 2nd time after restore.

addons/mail/static/tests/composer/composer.test.js
```diff
@@ -1481,8 +1482,25 @@ test("composer reply-to message is restored on thread change", async () => {
...
+    ).toBe(null);
+    // check local storage empty, change thread to force save in local storage if needed (debounced otherwise)
+    await click(".o-mail-DiscussSidebar-item:contains('Inbox')");
+    await contains(".o-mail-Message", { count: 0 });
+    await click(".o-mail-DiscussSidebar-item:contains('General')");
+    expect(
+        browser.localStorage.getItem(
+            store.Thread.get({ model: "discuss.channel", id: channelId }).composer.localId
+        )
+    ).toBe(null);
```
seb-odoo, line 1503: I don't see the point of this part, this is essentially a no-op as it was already empty before and we just close/open channel again?

Relevant new assert would be emptying the composer manually (no post) and checking emptiness (after it was filled again, or in another test).

seb-odoo (approved):
@robodoo delegate+

It seems to be working fine.

addons/mail/static/tests/composer/composer.test.js
alexkuhn: This is `null` after message post but there's some debounce code to save composer content. The change of conversation unmounts the composer and forces the saveContent.

So when looking at the test these steps looks useless, but this is actually the part that shows the fix is relevant for not re-adding the composer in local storage from component being unmounted.

addons/mail/static/src/core/common/composer.js
alexkuhn: So from my understanding we'd need save truthy of `emailAddSignature`.
This is a problem... Because all the polluted local storage entries had this value to `true` which is the default. The current code is too eagerly setting it to `true`, so if I just let it pass with truthy value we'd keep the current problem of empty composer in local storage.

addons/mail/static/src/core/common/composer.js
```diff
@@ -918,14 +918,20 @@ export class Composer extends Component {
...
-            browser.localStorage.setItem(
-                composer.localId,
-                JSON.stringify({
-                    emailAddSignature,
-                    replyToMessageId,
-                    composerHtml: isMarkup(composerHtml) ? ["markup", composerHtml] : composerHtml,
-                })
-            );
+            if (isHtmlEmpty(composerHtml)) {
+                browser.localStorage.removeItem(composer.localId);
```
phenix-factory, line 922: What about all the ones that are already present? Maybe an upgrade script? Since we now have this tech

alexkuhn: I thought about it, the upgrade scripts are available starting at 19.1, not 19.0.

I'm not entirely sure how I'd implement it, but I felt that I could look at this problem in its own PR. The related task is marked as "urgent" so I felt removing the proliferation of the disease is already a good enough fix that could be merged imminently.

seb-odoo: Old remains actually get cleaned up with this PR when the same composer is visited again.
Those never visited? Well too bad for them I guess. It would be nice to clean at some point but it's less a priority compared no longer creating them.

addons/mail/static/src/views/web/fields/html_composer_message_field/html_composer_message_field.js
```diff
@@ -20,7 +20,7 @@ export class HtmlComposerMessageField extends HtmlMailField {
                 const emailAddSignature = Boolean(
                     this.editor.editable.querySelector(".o-signature-container")
                 );
-                const composerHtml = markup(this.getNoSignatureElContent().innerHTML);
+                const composerHtml = markup(this.editor.getElContent().innerHTML);
```
seb-odoo: What's the change?

addons/mail/static/src/core/common/composer.js
alexkuhn: `emailAddSignature` is badly defined: right now this means both "should add signature" and "signature has been added". These 2 definitions at once are problematic because we can't know for sure if the signature has already been added. This "worked" because on reliance on adding the actual signature from opening the full composer, but because of omitting the signature part in `composerHtml` sent by full composer, this forces wasting local storage space.

I've fixed the problem around `emailAddSignature` to become "should add signature", which should have been its rightful definition.

This implies default value is `undefined`, so that it defaults to `true` and adds the signature. When the signature is effectively added, this flag is `false`, meaning it should not add the signature again. The Full composer trigger to save in local storage was removing the signature from `composerHtml`, which is a mistake: the content is not empty, thus it should show the signature. I fixed by providing the actual `composerHtml`.

All this makes it so that the save in local storage (or removal) is solely reliant on whether `composerHtml` is empty or not. The signature part, when it has been added, is necessarily in `composerHtml`. The local storage saves `emailAddSignature` in the case the composer is empty and there's no signature. I'm no expert of `.o-signature-container`, but I think this works as intended with the changes I made.

addons/mail/static/src/views/web/fields/html_composer_message_field/html_composer_message_field.js
alexkuhn: https://github.com/odoo/odoo/pull/247030#discussion_r2763565364
Basically the full composer pretends the `composerHtml` is empty when there's only the signature.
This is a problem because we can't rely on simple condition of whether the html is empty or not.

This code assumed that the `emailAddSignature` flag was redundant, but the problematic usage made it conceptually wrong: the flag meant both both "should add signature" and "has signature added". This cannot work with `true` value and expect efficient local storage space.

So this change makes it so that it provides the actual `composerHtml`, including signature being added. That way we can rely solely on `composerHtml` being empty or not.

alexkuhn: Hm actually the `emailAddSignature` has been designed to mean "this message has a signature".

So in the end I picked this definition, and tweaked the code to use another field to mean "should add the email signature".

The save in local storage is around this new field "should add the email signature", rather than the `emailAddSignature`.

All this made it match the intended functional behavior, while efficiently save in local storage only when the `composerHtml` value is non-empty.

alexkuhn: `isHtmlEmpty(body)` should be more than enough to save in local storage, otherwise remove it, without any diff around `emailAddSignature`.

addons/mail/static/src/core/common/composer.js
alexkuhn: https://github.com/odoo/odoo/pull/247030#discussion_r2773935416
Just `isHtmlEmpty(composerHtml)` in condition is enough, without change to `emailAddSignature`.

PR #247032, opened 2026-02-03
[IMP] iot_drivers: simplify linux printer interface

loouis-t:
we now deduplicate printers before adding them to cups, and name `dssnd` printers using their mac address as we already do for lpd and socket ones: this prevents the "Not Found" page when clicking on a `dnssd` printer as the name contained forbidden caracters to provide to a url.

### DIFF
--- a/addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
+++ b/addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
@@ -33,36 +33,21 @@ class PrinterInterface(Interface):
         self.PPDs = self.conn.getPPDs()
 
     def get_devices(self):
-        discovered_devices = {}
-        printers = self.conn.getPrinters()
-        devices = self.conn.getDevices()
+        # get printers already configured in Cups
+        discovered_devices = {
+            name: {"identifier": name, "already-configured": True, **printer}
+            for name, printer in self.conn.getPrinters().items()
+        }
 
-        # get and adjust configuration of printers already added in cups
-        for printer_name, printer in printers.items():
-            path = printer.get('device-uri')
-            if path and printer_name != self.get_identifier(path):
-                device_class = 'direct' if 'usb' in path else 'network'
-                printer.update({
-                    'already-configured': True,
-                    'device-class': device_class,
-                    'device-make-and-model': printer_name,  # give name set in Cups
-                    'device-id': '',
-                })
-                devices.update({printer_name: printer})
+        # Check if new printers available, and configure them if so
+        for name, printer in self.conn.getDevices().items():
+            identifier, printer = self.process_device(name, printer)
 
-        # filter devices (both added and not added in cups) to show as detected by the IoT Box
-        for path, device in devices.items():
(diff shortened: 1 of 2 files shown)


addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
```diff
@@ -94,37 +79,27 @@ def process_device(self, path, device):
...
+        e.g. ``uuid=1234-5678-90ab-cdef`` -> ``1234-5678-90ab-cdef``
         """
         ip = self.get_ip(path)
         if ip:
             mac_address = self.get_mac_from_ip(ip)
             if mac_address:
-                path = path.replace(ip, mac_address)
-        return re.sub(r'[:\/\.\\ ]|(uuid=)|(serial=)', '', path)[:127]
+                path = f"dnssd{mac_address}" if "dnssd" in path else path.replace(ip, mac_address)
+        return re.sub(r'[:/.?@\\ ]|(uuid=)|(serial=)]', '', unquote(path))[:127]
```
MaxTheWhale:
```suggestion
        return re.sub(r'[:/.?@\\ ]|(uuid=)|(serial=)', '', unquote(path))[:127]
```
An extra `]` was causing the `serial=` replacement to not work

addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
```diff
@@ -94,37 +79,27 @@ def process_device(self, path, device):
             'url': path,
             'disconnect_counter': 0,
         })
-        if device['device-class'] == 'direct':
+        if "usb" in path:
```
MaxTheWhale:
```suggestion
        if path.startswith("usb"):
```

addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
```diff
@@ -94,37 +79,27 @@ def process_device(self, path, device):
...
+        e.g. ``ipp://printers/printer1:1234/abcd`` -> ``ippprintersprinter11234abcd``
+        e.g. ``uuid=1234-5678-90ab-cdef`` -> ``1234-5678-90ab-cdef``
         """
         ip = self.get_ip(path)
         if ip:
             mac_address = self.get_mac_from_ip(ip)
             if mac_address:
-                path = path.replace(ip, mac_address)
-        return re.sub(r'[:\/\.\\ ]|(uuid=)|(serial=)', '', path)[:127]
+                path = f"dnssd{mac_address}" if "dnssd" in path else path.replace(ip, mac_address)
```
MaxTheWhale:
```suggestion
                path = f"dnssd{mac_address}" if path.startswith("dnssd") else path.replace(ip, mac_address)
```

MaxTheWhale (approved):
Tested:
- DNSSD printers use the new format and no longer give a Not Found page in CUPS ✅ 
- Printers are de-duplicated before adding to CUPS ✅ 
- USB printers can still be hotplugged ✅ 

Looks good to me! Just a few minor comments

PR #247034, opened 2026-02-03
[FIX] pos_self_order: pay after meal only shows the line changes price

hadielyakhni:
Steps to reproduce
------------------
If we set pay after to "meal", order one product A for price 10, then on
the same order, adds again product A and go to cart page, we see
"product A: 1" as expected, but its price set to "20".

-> Mismatch between the shown quantity and the total price of that line.

Why it's happening
------------------
When displaying the line qty, we just display the diff (1), however,
when displaying its price, we take the price of the whole line (having
qty 2).

Fix
---
Backport the line price display amount change made in saas-18.2 in
https://github.com/odoo/odoo/commit/98f0534450520ce037039ffe0be17cf1dc3b91ba.

We also fix "Your Order" to show the price of only the products of the
newly added lines.

opw-5467612


loukwets:
Hello @hadielyakhni,

Shouldn't this be fixed too ?

image

stva-odoo (approved):

PR #247035, opened 2026-02-03
[IMP] account_edi_ubl_cii: adding new formats

baje-odoo:
Add French variations for UBL, CII, and Factur-X formats to support France's mandatory e-invoicing requirements.

Add support for French Peppol EAS `0225` (FR:CTC) in EAS mappings/selections.

These profiles are required for compliance with French B2B e-invoicing regulations:

- UBL/CII: EN16931 (standard) and EXTENDED-CTC-FR (extended French)
- Factur-X: BASIC WL, EN16931, and EXTENDED

The different profiles allow handling varying levels of invoice complexity as required by French tax authorities.

Task : 4603737


addons/account_edi_ubl_cii/models/__init__.py
```diff
@@ -1,9 +1,16 @@
 from . import account_edi_common
 from . import account_edi_ubl
 from . import account_edi_xml_cii_facturx
+from . import account_edi_xml_cii_fr
+from . import account_edi_xml_cii_fr_extended
+from . import account_edi_xml_facturx_fr_basicwl
+from . import account_edi_xml_facturx_fr_en16931
+from . import account_edi_xml_facturx_fr_extended
```
gawa-odoo, line 8: Should this be moved in the new pdp module? I know it did not exist when you did open that PR (and would need a rebase on his).
It would avoid the checking on fields presence etc...

addons/account_edi_ubl_cii/models/account_edi_xml_cii_fr.py
```diff
@@ -0,0 +1,1102 @@
...
+
+# Namespace map for element creation
+NSMAP = {
+    'ram': CII_NAMESPACES['ram'],
+    'udt': CII_NAMESPACES['udt'],
+    'qdt': CII_NAMESPACES['qdt'],
+}
+
+
+def _make_elem(tag, text=None, attrib=None, nsmap=None):
```
gawa-odoo, line 22: In Bis3, we have defined the `_export_invoice_new`, which has as a goal the fact that we don't need to deal with xml, and so, don't need to force people to update their module. I suppose this has a similar goal, but why re-define another way of doing something similar?
We work everything from dictionaries in Python, and only at the end, would we convert it to an xml.

It would avoid having to generate your xml, then with elem.find do things.

I know it would need to adapt the `account.edi.xml.cii` to the Python from the templates

addons/account_edi_ubl_cii/models/account_edi_xml_cii_fr.py
```diff
@@ -0,0 +1,1102 @@
...
+            'line_vals': self._get_line_french_vals(invoice),
+        }
+
+    def _get_french_notes(self, invoice):
+        """Get French-specific notes with subject codes."""
+        notes = []
+
+        # Mapping of field names to subject codes
+        note_fields = {
+            'l10n_fr_note_aab': 'AAB',  # Cash discount terms
```
gawa-odoo, line 133: Not sure what these all are
real fields or not?

addons/account_edi_ubl_cii/models/account_edi_xml_cii_fr.py
```diff
@@ -0,0 +1,1102 @@
...
+
+    def _get_party_french_vals(self, partner):
+        """Get French-specific values for a trade party."""
+        if not partner:
+            return {}
+
+        vals = {}
+
+        # Street3 / LineThree (BT-162/163/164/165)
+        if 'street3' in partner._fields and partner.street3:
```
gawa-odoo, line 165: seems facultative, no?
(like we can probably fit all infos on first 2 street fields)

addons/account_edi_ubl_cii/models/account_edi_xml_cii_fr.py
```diff
@@ -0,0 +1,1102 @@
...
+            vals['street3'] = partner.street3
+
+        # Country Subdivision Name (BT-39/54/68/79)
+        if partner.state_id:
+            vals['country_subdivision_name'] = partner.state_id.name
+
+        # Trading Business Name (BT-28/45)
+        if 'l10n_fr_trading_name' in partner._fields and partner.l10n_fr_trading_name:
+            vals['trading_business_name'] = partner.l10n_fr_trading_name
```
gawa-odoo, line 174: Seems like that the name of the partner no? Do we really want him to change it just here?

gawa-odoo: And same for the rest of the fields, but I know you will check it with OLMA

addons/account_edi_ubl_cii/models/account_edi_xml_cii_fr.py
```diff
@@ -0,0 +1,1102 @@
...
+
+    def _get_partner_global_ids(self, partner):
+        """Get global identifiers for a partner."""
+        global_ids = []
+
+        # SIRET
+        if 'siret' in partner._fields and partner.siret:
+            global_ids.append({'id': partner.siret, 'scheme_id': '0002'})
+        elif partner.company_registry:
+            global_ids.append({'id': partner.company_registry, 'scheme_id': '0002'})
```
gawa-odoo, line 203: with sven's PR, will probably be the pdp identifier?

addons/account_edi_ubl_cii/models/account_edi_xml_ubl_fr.py
```diff
@@ -0,0 +1,550 @@
...
+
+        # Override CustomizationID for EN16931
+        document_node['cbc:CustomizationID'] = {'_text': self._get_customization_ids()['ubl_fr']}
+
+        # [BR-FR-08] Override ProfileID with French billing mode
+        billing_mode = self._get_french_billing_mode(invoice)
+        document_node['cbc:ProfileID'] = {'_text': billing_mode}
+
+        # BT-7: TaxPointDate
+        if 'l10n_fr_tax_point_date' in invoice._fields and invoice.l10n_fr_tax_point_date:
```
gawa-odoo, line 90: We added studio fields in Peppol.
To see if all these additional fields can be done the same way (if they're optional)

addons/account_edi_ubl_cii/models/account_move.py
```diff
@@ -74,6 +74,20 @@ def get_extra_print_items(self):
     def _get_ubl_cii_builder_from_xml_tree(self, tree):
         customization_id = tree.find('{*}CustomizationID')
         if tree.tag == '{urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100}CrossIndustryInvoice':
+            # Route CII imports to FR builders when the payload is clearly French.
+            guideline_id = (
+                tree.findtext('.//{*}GuidelineSpecifiedDocumentContextParameter/{*}ID') or ''
+            ).strip()
+            is_french_cii = self.env['account.edi.xml.cii_fr']._is_french_invoice(tree)
+            if guideline_id == 'urn:factur-x.eu:1p0:basicwl' and is_french_cii:
```
gawa-odoo, line 82: Not sure why we check `is_french_cii`
If we receive a specific format and we have a parser for that one precisely, shouldn't we use it?

addons/account_edi_ubl_cii/models/account_move_send.py
```diff
@@ -145,20 +147,26 @@ def _hook_invoice_document_after_pdf_report_render(self, invoice, invoice_data):
...
             self._postprocess_invoice_ubl_xml(invoice, invoice_data)
 
         # Always silently generate a Factur-X and embed it inside the PDF for inter-portability
-        if invoice_data.get('ubl_cii_xml_options', {}).get('ubl_cii_format') == 'facturx':
-            xml_facturx = invoice_data['ubl_cii_xml_attachment_values']['raw']
+        if ubl_cii_format in self._FACTURX_FORMATS:
+            xml_attachment_values = invoice_data['ubl_cii_xml_attachment_values']
+            xml_facturx = xml_attachment_values['raw']
+            # Factur-X expects a stable embedded filename.
+            xml_facturx_filename = 'factur-x.xml'
```
gawa-odoo, line 161: Could be done outside of the if to avoid declaring it twice :D

addons/account_peppol/models/res_company.py
```diff
@@ -311,10 +311,6 @@ def _peppol_modules_document_types(self):
                     "Peppol BIS Billing UBL Invoice V3",
                 "urn:oasis:names:specification:ubl:schema:xsd:CreditNote-2::CreditNote##urn:cen.eu:en16931:2017#compliant#urn:fdc:peppol.eu:2017:poacc:billing:3.0::2.1":
                     "Peppol BIS Billing UBL CreditNote V3",
-                "urn:oasis:names:specification:ubl:schema:xsd:Invoice-2::Invoice##urn:cen.eu:en16931:2017#compliant#urn:fdc:nen.nl:nlcius:v1.0::2.1":
```
gawa-odoo, line 314: Care to explain?

addons/l10n_sa_edi/models/account_edi_format.py
gawa-odoo, line 1: What did you change here and is it on purpose?

PR #247039, opened 2026-02-03
[FIX] account: access error auto post with only account

guva-odoo:
When a user with `group_account_manager` only tries to confirm an
invoice with auto post set, he gets an Access Error

Steps:

- Install only account
- Create an invoice, set auto post to 'monthly' for example
- Confirm
-> AccessError

Fix:

In the `_check_user_access` method, we align with what we do in the
`write` method to set the `is_user_able_to_review` variable

opw-5886542


antoine162 (approved):
@robodoo r+

PR #247040, opened 2026-02-03
[ADD] l10n_gr_edi{_delivery_note}: Add dispatch note (type 9.3)

vicire1:
To comply with Greek tax regulations on the mandatory electronic
reporting of physical goods movement, this commit introduces the
possibility to send Dispatch/Delivery Note (Document Type 9.3).

Under Greek tax legislation, entities delivering goods to customers
or transferring stock between internal locations must report these
physical movements electronically to the AADE platform.

- Implemented support for Dispatch Notes adding required fields on
the stock picking view
- Added the linking of those delivery notes to the invoice when those
are created with the same sale order
- Added prior sending to AADE validation to ensure that evrythin required
is filled in

task-5485365


addons/l10n_gr_edi/__manifest__.py
```diff
@@ -10,7 +11,7 @@
         to digitize business tax and accounting information declaration.
     """,
     'countries': ['gr'],
-    'depends': ['account_edi_ubl_cii', 'l10n_gr'],
+    'depends': ['account_edi_ubl_cii', 'l10n_gr', 'stock', 'sale', 'sale_stock'],
```
alialfie: Shouldn't this be done in a new module? Something like `l10n_gr_edi_sale_stock` or two new modules for sale and stock? Forcing a dependency on the stock and sale app for the l10n_gr module when users might just want to have the accounting app is not great. Could you double check this with the PO please?

addons/l10n_gr_edi/models/stock_move.py
```diff
@@ -0,0 +1,44 @@
...
+        string="myDATA Unit of Measure",
+        compute='_compute_l10n_gr_edi_measurement_unit',
+        store='True',
+    )
+
+    @api.depends('picking_id.is_greek_company', 'uom_id')
+    def _compute_l10n_gr_edi_measurement_unit(self):
+        uom_map = {name: code for code, name in MEASUREMENT_UNIT_SELECTION}
+        for move in self:
+            if self.company_id.account_fiscal_country_id.code == 'GR' and move.uom_id:
```
alialfie:
```suggestion
            if move.company_id.account_fiscal_country_id.code == 'GR' and move.uom_id:
```
Or is it intended?

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,392 @@
...
+        if not (self.company_id.l10n_gr_edi_aade_id and self.company_id.l10n_gr_edi_aade_key):
+            errors['l10n_gr_edi_company_no_cred'] = {
+                'message': _("myDATA: You need to set AADE ID and Key in the company settings."),
+                **error_action_gr_settings,
+            }
+        if not (self.company_id.zip and self.company_id.city):
+            errors['l10n_gr_edi_company_no_zip_city'] = {
+                'message': _("myDATA: Missing city and/or ZIP code on partner %s.", self.partner_id.name),
+                **error_action_partner,
+            }
```
alialfie, line 352: Checking company_id fields but error blames the partner and opens the partner form.

addons/l10n_gr_edi/models/l10n_gr_edi_document.py
```diff
@@ -94,6 +98,112 @@ class GreeceEDIDocument(models.Model):
...
+    def _l10n_gr_edi_handle_send_result(self, record, result, xml_vals):
+        """
+        Handle the result object received from sending xml to myDATA.
+        Create the related error/sent document with the necessary values.
+        """
+        xml_map = {}  # Dictionary mapping of ``move_id, picking_id`` -> ``xml_content``.
+        for vals in xml_vals['invoice_values_list']:
+            single_xml_vals = {'invoice_values_list': [vals]}
+            move = vals['__move__']
+            if record._name == 'account.move' and record.is_purchase_document(include_receipts=True):
```
alialfie: In this new implementation, can't `record` be a recordset with multiple bills/invoices? The previous implementation handled each move separately.

addons/l10n_gr_edi/models/l10n_gr_edi_document.py
```diff
@@ -94,6 +98,112 @@ class GreeceEDIDocument(models.Model):
...
+            xml_content = self._l10n_gr_edi_generate_xml_content(xml_template, single_xml_vals)
+            xml_map[move] = xml_content
+
+        move_ids = list(xml_map.keys())
+
+        if 'error' in result:
+            # If the request failed at this stage, it is probably caused by connection/credentials issues.
+            # In such case, we don't need to attach the xml here as it won't be helpful for the user.
+            for move in move_ids:
+                self._l10n_gr_edi_create_error_document(record, result)
```
alialfie: Same comment. Looping over a single record `move` but passing a multi recordset `record`. Is this intended?

addons/l10n_gr_edi/models/l10n_gr_edi_document.py
```diff
@@ -94,6 +98,112 @@ class GreeceEDIDocument(models.Model):
...
+                xml_content = xml_map[move]
+                document_values = {**result_dict, 'xml_content': xml_content}
+                # Delete previous error documents
+                move.l10n_gr_edi_document_ids.filtered(lambda d: d.state in ('invoice_error', 'bill_error', 'delivery_note_error')).unlink()
+                if 'error' in result_dict:
+                    # In this stage, the sending process has succeeded, and any error we receive is generated from the myDATA API.
+                    # Previous error(s) without attachments (generated from pre-compute) are now useless and can be unlinked.
+                    self._l10n_gr_edi_create_error_document(record, document_values)
+                else:
+                    self._l10n_gr_edi_create_sent_document(record, document_values)
```
alialfie: Same

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,392 @@
...
+        comodel_name='ir.attachment',
+        compute='_compute_from_l10n_gr_edi_document_ids',
+        store=True,
+    )
+    l10n_gr_edi_alerts = fields.Json(compute='_compute_l10n_gr_edi_alerts')
+    l10n_gr_edi_move_purpose = fields.Selection(
+        selection=MOVE_PURPOSE_SELECTION,
+        compute='_compute_l10n_gr_edi_move_purpose',
+        string="Move Purpose",
+        store=True,
```
alialfie, line 67: According to [this](https://www.odoo.com/documentation/19.0/developer/reference/backend/orm.html#computed-fields), computed fields are read only by default, yet in the view, you're presenting the field as an editable one, with `readonly="l10n_gr_edi_state == 'delivery_note_sent'"` as a condition. You might wanna add `readonly=False,` here

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -551,14 +501,23 @@ def _l10n_gr_edi_get_invoices_xml_vals(self):
...
                 })
 
+            correlated_marks = {p.l10n_gr_edi_mark for p in move.l10n_gr_edi_correlated_picking_ids}
+            if move.l10n_gr_edi_correlation_id.l10n_gr_edi_mark:
+                correlated_marks.add(move.l10n_gr_edi_correlation_id.l10n_gr_edi_mark)
+
+            # correlated_marks = {move.l10n_gr_edi_correlation_id.l10n_gr_edi_mark} or {}
+            # # In case of a return we don't want to include the correlated delivery notes
+            # if move.move.l10n_gr_edi_inv_type != '':
+            #     correlated_mark.add({p.l10n_gr_edi_mark for p in move.l10n_gr_edi_correlated_picking_ids})
```
alialfie: 🤨

addons/l10n_gr_edi/utils.py
```diff
@@ -0,0 +1,29 @@
...
+            'street_number': match.group(2).strip(),
+        }
+
+    # number at the end
+    match = re.match(r'^(\d+[-/]?\d*\s*\w?)\s+(.+)$', street)
+    if match:
+        return {
+            'street_name': match.group(2).strip(),
+            'street_number': match.group(1).strip(),
+        }
```
alialfie: The comments are reversed?

addons/l10n_gr_edi/models/l10n_gr_edi_document.py
```diff
@@ -94,6 +98,112 @@ class GreeceEDIDocument(models.Model):
...
+            'mydata_cls_mark': <optional/str>,
+            'mydata_url': <str>,
+            'xml_content': <str>,
+        }
+        """
+        if record._name == 'account.move':
+            state = 'invoice_sent' if record.is_sale_document(include_receipts=True) else 'bill_sent'
+        elif record._name == 'stock.picking':
+            state = 'delivery_note_sent'
+        document = self.env['l10n_gr_edi.document'].create({
```
alialfie:
```suggestion
        document = self.create({
```
Why not this

addons/l10n_gr_edi/models/stock_picking.py
```diff
@@ -0,0 +1,392 @@
...
+    l10n_gr_edi_loading_address_zip = fields.Char()
+    l10n_gr_edi_loading_address_city = fields.Char()
+    l10n_gr_edi_delivery_address_street = fields.Char()
+    l10n_gr_edi_delivery_address_number = fields.Char()
+    l10n_gr_edi_delivery_address_zip = fields.Char()
+    l10n_gr_edi_delivery_address_city = fields.Char()
+    l10n_gr_edi_vehicle_number = fields.Char(string="Vehicule number")
+
+    @api.depends('company_id', 'company_id.account_fiscal_country_id', 'company_id.account_fiscal_country_id.code')
+    def compute_is_greek_company(self):
```
alialfie:
```suggestion
    def _compute_is_greek_company(self):
```

addons/l10n_gr_edi_delivery_note/tests/test_mydata_delivery_note.py
```diff
@@ -0,0 +1,140 @@
...
+        mock_result = {
+            0: {
+                'error': '[223] Unsupported invoice type',
+            }
+        }
+        xml_vals = picking._l10n_gr_edi_get_pickings_xml_vals()
+
+        with self.assertRaises(UserError):
+            self.env['l10n_gr_edi.document']._l10n_gr_edi_handle_send_result(picking, mock_result, xml_vals)
+            picking._l10n_gr_edi_send_delivery_note()
```
alialfie, line 154: This should be two different tests/asserts. Which one do you expect to raise an error?

addons/l10n_gr_edi/__manifest__.py
vicire1: ~I'll confirm this with NASTI next week but I think it will be done this way because it is the edi module and delivery notes are now mandatory in Greece but have to confirm 😁~
Discussed with NASTI and will do a separate module

addons/l10n_gr_edi_delivery_note/models/account_move.py
```diff
@@ -0,0 +1,34 @@
...
+        for move in self:
+            if move.move_type == 'out_invoice':  # There will be delivery notes only in cases of out_invoices
+                pickings = move.invoice_line_ids.mapped('sale_line_ids.move_ids.picking_id')
+                move.l10n_gr_edi_correlated_picking_ids = pickings.filtered(lambda p: p.l10n_gr_edi_mark and p.picking_type_code == 'outgoing')
+            else:
+                move.l10n_gr_edi_correlated_picking_ids = False
+
+    @api.model
+    def _l10n_gr_edi_get_invoices_xml_vals(self):
+        # EXTENDS
```
alialfie: Extends what? 😆

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
alialfie: Did you check this? I see it in the new diff :raised_eyebrow: Please don't mark comments as resolved if they're not

addons/l10n_gr_edi/data/template.xml
```diff
@@ -33,33 +37,65 @@
...
+                            <deliveryAddress>
+                                <street t-out="invoice_vals['delivery_street']"/>
+                                <number t-out="invoice_vals['delivery_number']"/>
+                                <postalCode t-out="invoice_vals['delivery_postal_code']"/>
+                                <city t-out="invoice_vals['delivery_city']"/>
+                            </deliveryAddress>
+                        </otherDeliveryNoteHeader>
+                        <otherMovePurposeTitle t-out="invoice_vals['header_other_move_purpose_title']"/>
+                    </t>
+                    <multipleConnectedMarks t-foreach="invoice_vals['connected_marks']" t-as="picking" t-out="picking"/>
```
alialfie: `connected_marks` key is only added in the other module, and if it's not installed, you'll get an error here when you try to render this.
To do this cleanly, you should inherit the template in the other module and add this node there. Another option would be to use `.get()` here instead.

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -692,71 +631,27 @@ def _l10n_gr_edi_get_pre_error_dict(self):
...
-            self._l10n_gr_edi_handle_send_result(result, xml_vals)
+            self.env['l10n_gr_edi.document']._l10n_gr_edi_handle_send_result(result, xml_vals)
 
     def l10n_gr_edi_try_send_invoices(self):
         moves_to_send = self.env['account.move']
         for move in self:
-            if error := move._l10n_gr_edi_get_pre_error_string():
-                move._l10n_gr_edi_create_error_document({'error': error})
+            if error := move._l10n_gr_edi_get_pre_error_dict():
+                self.env['l10n_gr_edi.document']._l10n_gr_edi_create_error_document(self, {'error': utils.get_pre_error_string(error)})
```
alialfie:
```suggestion
                self.env['l10n_gr_edi.document']._l10n_gr_edi_create_error_document(move, {'error': utils.get_pre_error_string(error)})
```

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -767,8 +662,8 @@ def l10n_gr_edi_try_send_invoices(self):
     def l10n_gr_edi_try_send_expense_classification(self):
         moves_to_send = self.env['account.move']
         for move in self:
-            if error_message := move._l10n_gr_edi_get_pre_error_string():
-                move._l10n_gr_edi_create_error_document({'error': error_message})
+            if error_message := move._l10n_gr_edi_get_pre_error_dict():
+                self.env['l10n_gr_edi.document']._l10n_gr_edi_create_error_document(self, {'error': utils.get_pre_error_string(error_message)})
```
alialfie:
```suggestion
                self.env['l10n_gr_edi.document']._l10n_gr_edi_create_error_document(move, {'error': utils.get_pre_error_string(error_message)})
```

addons/l10n_gr_edi_delivery_note/models/stock_move.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    l10n_gr_edi_measurement_unit = fields.Selection(
+        selection=MEASUREMENT_UNIT_SELECTION,
+        string="myDATA Unit of Measure",
+        compute='_compute_l10n_gr_edi_measurement_unit',
+        store='True',
```
alialfie: That's just a string. Use `True` instead

addons/l10n_gr_edi_delivery_note/models/stock_move.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    @api.depends('picking_id.is_greek_company', 'uom_id')
+    def _compute_l10n_gr_edi_measurement_unit(self):
+        uom_map = {name: code for code, name in MEASUREMENT_UNIT_SELECTION}
+        for move in self:
+            if move.company_id.account_fiscal_country_id.code == 'GR' and move.uom_id:
```
alialfie: If you have a condition on the fiscal country, then you should depend on it so that the compute triggers when it changes.

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,412 @@
...
+from odoo.addons.l10n_gr_edi import utils
+from odoo.addons.l10n_gr_edi.models.l10n_gr_edi_document import _make_mydata_request
+from odoo.exceptions import UserError, ValidationError
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def _default_l10n_gr_edi_move_purpose(self):
+        return '1' if self.env.company.country_id.code == 'GR' else None
```
alialfie: Use account_fiscal_country_id instead, you're already using it everywhere else I think. It makes more sense for edi

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,412 @@
...
+            if picking.l10n_gr_edi_document_ids and not picking.l10n_gr_edi_state:
+                picking.l10n_gr_edi_state = 'delivery_note_error'
+
+    @api.depends('country_code', 'state')
+    def _compute_l10n_gr_edi_alerts(self):
+        for picking in self:
+            # Warnings are only calculated when the picking state is done.
+            # We use `._origin` to make sure the validated picking have all the needed data for validation.
+            if picking.is_greek_company and picking.state == 'done':
+                picking.l10n_gr_edi_alerts = picking._origin._l10n_gr_edi_get_pre_error_dict()
```
alialfie, line 125: `_l10n_gr_edi_get_pre_error_dict` checks a lot of fields that are not depended on. Same issue.

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
vicire1: Sorry about this, I thought I changed it... :facepalm:

addons/l10n_gr_edi/data/template.xml
vicire1: I preffered the .get() option so that the template is more 'complete', if you think inheritage is still a better option, tell me and I'll change it :smile:

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,431 @@
...
+            if not line.l10n_gr_edi_measurement_unit:
+                errors[f'l10n_gr_edi_{line_no}_missing_uom'] = {
+                    'message': self.env._("myDATA: unit of measure on line %s is invalid, please select between 'Units', 'kg', 'L', 'm', 'm²' or 'm³'", line_no),
+                }
+        return errors
+
+    def _l10n_gr_edi_attach_report_picking(self):
+        """Render delivery note PDF report and post log entry with attachment."""
+        self.ensure_one()
+        report_picking = self.env['ir.actions.report']._render_qweb_pdf('stock.action_report_delivery', self.id)[0]
```
alialfie:
```suggestion
        report_picking = self.env['ir.actions.report']._render_qweb_pdf('stock.action_report_delivery', self.ids)[0]
```
nitpick but the method expects res_ids. It will still handle an int anyway so it's okay

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
```diff
@@ -0,0 +1,431 @@
...
+        'partner_id.street',
+        'move_ids',
+        'move_ids.l10n_gr_edi_measurement_unit',
+    )
+    def _compute_l10n_gr_edi_alerts(self):
+        for picking in self:
+            # Warnings are only calculated when the picking state is done.
+            # We use `._origin` to make sure the validated picking have all the needed data for validation.
+            if picking.is_greek_company and picking.state == 'done':
+                picking.l10n_gr_edi_alerts = picking._origin._l10n_gr_edi_get_pre_error_dict()
```
alialfie, line 125: I'm not sure why _origin is necessary tbh. Could you please explain the usecase?

alialfie (commented):
Almost good to go, just one nitpick and a question 😄 
Could you also handle the conflict please?

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -1,9 +1,8 @@
-from lxml import etree
 from urllib.parse import urlencode
 
 from odoo import api, fields, models, _
+from odoo.addons.l10n_gr_edi import utils
```
alialfie: Should be the last one instead https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports

addons/l10n_gr_edi_delivery_note/models/stock_picking.py
vicire1: I assume it is to have all the datas really stored atm, tbh I did the same things that YONI did for the account_move

alialfie:
runbot is not happy :grimacing:

kmagusiak:
@robodoo override=ci/style
l10n commit

alialfie (approved):
robodoo r+

PR #247042, opened 2026-02-03
[FIX] web, portal: allow `confirmatioDialog` size change

stefanorigano:
### Context:

A confirmation dialog must be small and minimal for several reasons:

- Users are being interrupted mid-task. A wall of text forces them to context-switch and parse information when they just need to make a quick decision
- The goal is a binary choice (confirm/discard). Too much content complicates what should be a simple yes/no moment
- ...

In the backend, our dialogs `SM` size is set to a **forgiving** `460px`. This value allows dialogs to be relatively small, while still providing enough tolerance to showcase more content if necessary.

In the frontend, instead, the default value in use is BS default, thus `300px`. A website default font-size is also much bigger, `16px` against `13px` in the backend.

As a result, the dialog may be too small when used in scenarios that are not "confirmation dialogs" strictly speaking.

### The issue:

In some cases, enforcing SM size to any confirmation dialog gives bad results in the frontend.

| backend | fronted |
|--------|--------|
| image | image | 

Enforcing a higher `SM` value for the frontend (eg. `500px`) is not a solution because the default BS value is technically correct and users can customize it anyway by overriding this setting.

### The solution:

Allow the ConfirmationDialog component to use sizes different from `SM`. This commit sets some portal dialogs to MD size.

| saas-19.1 | this pr |
|--------|--------|
| image | image | 

task-5906425

note: Several design improvements are necessary and will be handled in forward-port targeting master 

---

### DIFF
--- a/addons/portal/static/src/interactions/portal_security.js
+++ b/addons/portal/static/src/interactions/portal_security.js
@@ -72,6 +72,7 @@ export class PortalSecurity extends Interaction {
                 duration_selection: duration.selection.filter((option) => option[0] !== "-1"),
             }),
             confirmLabel: _t("Confirm"),
+            size: 'md',
             confirm: async ({ inputEl }) => {
                 const formData = Object.fromEntries(new FormData(inputEl.closest("form")));
                 const wizardId = await this.services.orm.create("res.users.apikeys.description", [{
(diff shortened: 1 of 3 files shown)


aab-odoo (approved):
robodoo r+

PR #247043, opened 2026-02-03
[FIX] pos_hr: prevent loading all employees

pedrambiria:
Before this commit, when some employee was assigned to advanced or minimal employee, all of the employees were loaded in the POS session, because there was no basic employee assigned to the POS config.

opw-5898068

---


davidmonnom (approved):
@robodoo r+

PR #247054, opened 2026-02-03
[FIX] website_event_sale: ensure address form redirection for public user payment

delcourtfl:
**Steps to reproduce:**
- Install Website/Event/Sales apps
- Create a new event and set a ticket price
- Create a fiscal position (with specific tax) for a country with `auto_apply` enabled
- Publish the event
- Register to the event as a anonymous user
- The process bypasses the address form and goes directly to payment
- The resulting sale order will have no fiscal position
- Prices won't be impacted by taxes related to the user billing address

**Issue:**
Address form is skipped before payment for event registration of a public user as `_needs_customer_address` is not overwritten properly in some module. This is probably due to a refactoring that changed how the required information is evaluated in the payment flow (see related commit).

**Fix:**
Set `_needs_customer_address` to `True` by default to avoid such issues in
dependant modules. Might need to remove this feature in master as the workarounds are
not that clean (geo_ip, check on fiscal position enabled, overwrite everywhere, others ?).

Similar fix is done for appointments with payment enabled.

related: https://github.com/odoo/odoo/commit/d43f0423667835512e16c3fd3474328da63a948d
original-task: https://www.odoo.com/odoo/project/49/tasks/4307281

opw-5143124

### DIFF
--- a/addons/website_sale/models/sale_order.py
+++ b/addons/website_sale/models/sale_order.py
@@ -258,10 +258,11 @@ class SaleOrder(models.Model):
     def _needs_customer_address(self):
         """Return whether we need the address details of the customer (country, street, ...).
 
-        If an order only has services, unless the customer wants an invoice, their checkout can
-        be sped up by allowing them to only provide their name, email and phone numbers.
+        Make it true by default as it's required before payment for taxes based on fiscal position.
         """
-        return not self.only_services
+        # TODO: should probably be removed in master, as we cannot skip the address form
+        # when the taxes applied on product/service depend on the fiscal position.
+        return True
 
     def _update_address(self, partner_id, fnames=None):
         if not fnames:
--- a/addons/website_event_sale/static/tests/tours/website_event_sale_last_ticket.js
+++ b/addons/website_event_sale/static/tests/tours/website_event_sale_last_ticket.js
@@ -56,6 +56,17 @@ registry.category("web_tour.tours").add('event_buy_last_ticket', {
         run: "click",
         expectUnloadPage: true,
     },
+    ...wsTourUtils.fillAdressForm(
+        {
+            name: "John Doe",
+            phone: "123456789",
+            email: "johndoe@gmail.com",
+            street: "1 rue de la paix",
+            city: "Paris",
+            zip: "75000",
+        },
+        true,
+    ),
     ...wsTourUtils.payWithTransfer(true),
     ],
 });


ushyme (approved):
LGTM, thanks for your work.

reth-odoo:
Isn't it a broader issue? I see the docstring for `_needs_customer_address` says we skip it because we don't need it for delivery, but indeed we probably do always need it for taxes even if all products are services

addons/website_sale/models/sale_order.py
```diff
@@ -261,7 +261,7 @@ def _needs_customer_address(self):
         If an order only has services, unless the customer wants an invoice, their checkout can
         be sped up by allowing them to only provide their name, email and phone numbers.
         """
-        return not self.only_services
+        return True
```
livaios, line 265: Don't change the main function, you can override the `def _needs_customer_address(self):` function inside the relevant modules like the override in [`website_sale_subscription`](https://github.com/odoo/enterprise/blob/7bc3a7f692b3955fc369819f912b8b8370bd421a/website_sale_subscription/models/sale_order.py#L51)

livaios, line 265: Don't change the main function, you can override the `def _needs_customer_address(self):` function inside the relevant modules like the override in [`website_sale_subscription`](https://github.com/odoo/enterprise/blob/7bc3a7f692b3955fc369819f912b8b8370bd421a/website_sale_subscription/models/sale_order.py#L51)

Make sure it only applies to event orders

livaios (changes requested):

addons/website_sale/models/sale_order.py
delcourtfl: That's what I did first on both sides (appointment/event) https://github.com/odoo/odoo/compare/ff772c77b3f07673fc5e3d971e71cc090879bede..eedb32db5172965a11150666efe09f2f86495d89
I tried finding conditions where it should still be used, but I feel like the fiscal position issue is still present on `website_sale` directly. It should be possible to apply localized taxes on services as well right ?

Feyensv: Yes, they came to ask in the office directly and I told them that it was the whole point of the task and basically doesn't make sense anymore since taxes needs to be considered.
We change the base method to return True in stable, and we'll probably remove it someday in master.

livaios (changes requested):
Hello @delcourtfl 
Can you just add a comment explaining why its set to true and maybe add a TODO for master so that we don't miss why its kept as `True`.

livaios (approved):
Thanks for spotting this!
It's okay for the website_sale changes 🙂

reth-odoo (approved):
Still skips it for free tickets/appointments which is good 👍 
Otherwise makes sense to me thanks :)

jeh-odoo (approved):
@robodoo r+

PR #239998, opened 2025-12-15
[FIX] {l10n_ar_,}website_sale: handle missing l10n fields on checkout

lvsz:
Versions
--------
- 18.0

`l10n_ar_website_sale` was removed in later versions via 5a93da8e9220

Steps
-----
1. Create a partner with an address;
2. set up a company with Argentinian localization;
3. assign website to Argentinian company;
4. as partner, go to website & add something to your cart;
5. go to checkout.

Issue
-----
Cannot get past the address form.

Cause
-----
The `l10n_latam_identification_type_id` and `l10n_ar_afip_responsibility_type_id` fields are required, but not shown. Because they're not editable, the `website_sale` controller doesn't consider them "missing", making the `l10n_ar_website_sale` override not handle the missing fields.

Going to "My Account" to edit details also doesn't allow you to add them, as the form lacks the `can_edit_vat` value.

Solution
--------
Instead of relying on the `missing_fields` value from the base method call, have the `_validate_address_values` override explicitly check the presence of those fields in either the address values or the current partner record (adding them to address values if it's the latter).

If the values are missing, show an error message telling the client to add the missing info via "My Account".

In the portal controller, provide the `can_edit_vat` value to make the fields editable.

opw-5376149

PR #247084, opened 2026-02-03
[FIX] l10n_ar_website_sale: handle missing l10n fields on checkout

mma1377:
### Issue:

Cannot get passed the address form in `ar` localization.

#### Steps to reproduce:
1- Set up a company with Argentinian localization.
2- Set the portal user to this company.
3- Add an address and set country to `Argentina`.
4- Assign website to Argentinian company;
5- As the portal user, add something to cart.
6- go to checkout, and save the address.

You will get the error that some fields are missing.

### Cause:

This issue is partially fixed in https://github.com/odoo/odoo/pull/239998. However, there are three issues left:

1. That fix doesn't show the error regarding 'Identification and AFIP Resposibility Types' when we have `use_delivery_as_billing`. As a result a generic error `some fields are missing` is shown.

2. `can_edit_vat` is used in:
https://github.com/odoo/odoo/blob/5b39cdceb84758f767b83b741be927c1a7cfac7a/addons/l10n_ar_website_sale/views/templates.xml#L7-L9
https://github.com/odoo/odoo/blob/5b39cdceb84758f767b83b741be927c1a7cfac7a/addons/l10n_ar_website_sale/views/templates.xml#L32-L35
This needs to also consider `l10n_ar_afip_responsibility_type_id` and `l10n_latam_identification_type_id` being set.

3. Also we need to make sure `Identification Number` which is `vat` is set if we want to prevent it from being modified.

opw-5376149

### DIFF
--- a/addons/l10n_ar_website_sale/controllers/main.py
+++ b/addons/l10n_ar_website_sale/controllers/main.py
@@ -41,18 +41,21 @@ class L10nARWebsiteSale(WebsiteSale):
             fnames.add('l10n_latam_identification_type_id')
         return fnames
 
-    def _validate_address_values(self, address_values, partner_sudo, address_type, *args, **kwargs):
+    def _validate_address_values(self, address_values, partner_sudo, address_type, use_delivery_as_billing=None, *args, **kwargs):
         """ We extend the method to add a new validation. If AFIP Resposibility is:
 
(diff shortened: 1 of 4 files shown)


antoine162 (approved):
👍 Thanks so much!
@robodoo r+

PR #247091, opened 2026-02-03
[FIX] hr_holidays : Automatically calculate the allocation duration when different mode

yoba-odoo:
### Steps to reproduce:
	- Create an accrual plan of one level to give 20 days at the start of the year
	- Create an allocation with different mode than 'By Employee'
	- Set the accrual plan for the allocation and date from 1st Jan
	- Notice the Allocation number of days doesn't get automatically calculated

### Cause:
This is happening because when trying to process the accrual plan we won't have any records in the field employee_id

https://github.com/odoo/odoo/blob/bcdd12d13d73915e565fd2c8478b936a16efb9f4/addons/hr_holidays/models/hr_leave_allocation.py#L892-L893

And since employee_id is computed field when computing it we don't handle the case of any other mode other than 'By Employee'.

https://github.com/odoo/odoo/blob/bcdd12d13d73915e565fd2c8478b936a16efb9f4/addons/hr_holidays/models/hr_leave_allocation.py#L259-L270

### Fix:
If we have different mode in the allocation we fetch the employees in this mode (Department, Company, Employee Tag) and set them as the allocation employee_ids so when computing the employee_id we will have a record in the field and it won't be null

P.S. In the forward port we will have to introduce another fix for the multi allocation wizard

opw-5888023

### DIFF
--- a/addons/hr_holidays/models/hr_leave_allocation.py
+++ b/addons/hr_holidays/models/hr_leave_allocation.py
@@ -832,10 +832,14 @@ class HolidaysAllocation(models.Model):
 
                 allocation_vals += allocation._prepare_holiday_values(employees)
         if allocation_vals:
-            self.env['hr.leave.allocation'].with_context(
+            allocations = self.env['hr.leave.allocation'].with_context(
                 mail_notify_force_send=False,
                 mail_activity_automation_skip=True
             ).create(allocation_vals)
+            accrual_allocations = allocations.filtered(lambda a: a.allocation_type == 'accrual')
+            for date_to, allocation in accrual_allocations.grouped('date_to').items():
+                date_to = min(date_to, date.today()) if date_to else False
+                allocation._process_accrual_plans(date_to)
         self.linked_request_ids.filtered(lambda c: c.state != 'validate').action_validate()
 
     def action_refuse(self):
--- a/addons/hr_holidays/tests/test_accrual_allocations.py
+++ b/addons/hr_holidays/tests/test_accrual_allocations.py
@@ -16,6 +16,9 @@ class TestAccrualAllocations(TestHrHolidaysCommon):
     @classmethod
     def setUpClass(cls):
         super(TestAccrualAllocations, cls).setUpClass()
+        cls.department = cls.env['hr.department'].create({
+            'name': 'Test Department',
+        })
         cls.leave_type = cls.env['hr.leave.type'].create({
             'name': 'Paid Time Off',
             'time_type': 'leave',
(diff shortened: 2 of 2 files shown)


addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -269,7 +269,7 @@ def _compute_from_employee_ids(self):
                 allocation.employee_id = False
             allocation.multi_employee = len(employees) > 1
 
-    @api.depends('holiday_type')
+    @api.depends('holiday_type', 'category_id', 'department_id', 'mode_company_id')
```
Bertrand2: feels wrong to have in the depends the fields that are modified by the compute

Bertrand2 (commented):
Thanks for the fix!
However, I have some issues with the functional content of it.
The fact that the amount given by the accrual isn't computed is actually intended and considered a limitation in the case of an allocation with multiple employees; simply because the accrued amount might vary depending on the working schedule, already taken leaves etc., bringing every selected employee to a different amount.
As mentioned in my comment, I also feel like it's weird to add dependencies to fields that are modified by the compute. Though it can be more tedious, in the case of accrual allocations I advice doing it one by one

addons/hr_holidays/models/hr_leave_allocation.py
```diff
@@ -832,10 +832,13 @@ def _action_validate_create_childs(self):
...
                 allocation_vals += allocation._prepare_holiday_values(employees)
         if allocation_vals:
-            self.env['hr.leave.allocation'].with_context(
+            allocations = self.env['hr.leave.allocation'].with_context(
                 mail_notify_force_send=False,
                 mail_activity_automation_skip=True
             ).create(allocation_vals)
+            if self.allocation_type == 'accrual':
+                date_to = min(self.date_to, date.today()) if self.date_to else False
+                allocations._process_accrual_plans(date_to)
```
Bertrand2: If you validate multiple-employee allocations at the same time, self will not be a singleton
```suggestion
            accrual_allocations = allocations.filtered(lambda a: a.allocation_type == 'accrual')
            for date_to, allocations in accrual_allocations.grouped('date_to').items():
                date_to = min(date_to, date.today()) if date_to else False
                allocations._process_accrual_plans(date_to)
```

Bertrand2 (changes requested):
Thanks for the changes! I still have one comment :)

Bertrand2 (approved):
Thanks for the changes! LGTM
robodoo r+

PR #247104, opened 2026-02-03
[18.0][FIX] auth_signup: render new device alert email in user's language

pilarvargas-tecnativa:

The “new device” security alert email sent by auth_signup renders its QWeb body without any lang in the rendering context. When env.context does not carry a language (which is the case in this code path), QWeb falls back to the source language (en_US), even if the recipient user has a different language configured and translations are available.
This results in mixed-language emails (translated subject via _(), but body still in English).

Configure a user language different from en_US (e.g. es_ES).

Login with that user from a new browser/device (or after clearing cookies) to trigger the “new device” alert.

The email is sent, but the body is rendered in English because the template is rendered without lang in context (fallback to en_US), even when QWeb translations exist.

The email body is rendered using the recipient user’s language (res.users.lang / partner.lang) by explicitly setting lang in the rendering context before rendering the QWeb template.

As a result, when translations exist, the whole email (subject and body) is consistently rendered in the user’s language instead of falling back to en_US.

**Steps to reproduce:**

Set a user language to es_ES (or any non-en_US language).

Ensure translations exist for auth_signup.alert_login_new_device.

Login from a new browser/device (or clear cookies) to trigger the alert.

Observe the received email: body is in English before this PR, and in the user’s language after.

@Tecnativa TT60646
@victoralmau @pedrobaeza please review

---

### DIFF
--- a/addons/auth_signup/models/res_users.py
+++ b/addons/auth_signup/models/res_users.py
@@ -278,6 +278,7 @@ class ResUsers(models.Model):
     def _alert_new_device(self):
         self.ensure_one()
         if self.email:
+            lang = self.lang or "en_US"
             email_values = {
                 'email_cc': False,
                 'auto_delete': True,
@@ -288,7 +289,7 @@ class ResUsers(models.Model):
                 'email_to': self.email
             }
 
-            body = self.env['mail.render.mixin']._render_template(
+            body = self.env['mail.render.mixin'].with_context(lang=lang)._render_template(
                     'auth_signup.alert_login_new_device',
                     model='res.users', res_ids=self.ids,
                     engine='qweb_view', options={'post_process': True},


addons/auth_signup/models/res_users.py
```diff
@@ -278,6 +278,7 @@ def send_unregistered_user_reminder(self, *, after_days=5, batch_size=100):
     def _alert_new_device(self):
         self.ensure_one()
         if self.email:
+            lang = self.lang or self.partner_id.lang
```
pedrobaeza: The field is the same in the user or the partner (related inherited by delegation), so:

```suggestion
            lang = self.lang or "en_US"
```

pilarvargas-tecnativa: Right, thanks.

pilarvargas-tecnativa:
Thanks! It looks like this issue has now been addressed by #261468, which has already been merged.

This PR predates that one, but since the fix is now available upstream, there is no need to keep this one open.

Closing as superseded.

PR #247105, opened 2026-02-03
[PERF] stock: optimize purchase order computation on lots

spco-odoo:

We optimize the retrieval of stock move lines by resolving a classic N+1 query bottleneck. In the previous implementation, the system performed a database search for every individual lot in the recordset. This refactor backports the batch processing pattern to fetch all necessary data in a single query, significantly improving performance for large data batches. 

This PR goes beyond just backporting the V18/19 logic by using python set logic instead of recordsets to ensure that that grouping logic remains O(n)

### Current behavior before PR:

The system uses a nested search pattern within a loop. A separate database request is triggered for every lot in the recordset. Performance degrades linearly as the number of lots increases, leading to slow processing times for large batches. 

Logic is refactored into a single batch search on stock.move.line. Database overhead is reduced to a constant number of queries regardless of the recordset size. A defaultdict of python sets is utilized to map findings back to their respective lots. 

### Benchmarks

Profiling creation of purchase orders. Database has ~4200 stock.lot records.

| Before | After |
|---|---|
| 27sec | 7sec |  

### References

opw-5453842

Backported from: 99b39b72c7e65e85af6f06dcb6b02867623f3f69

---

### DIFF
--- a/addons/purchase_stock/models/stock.py
+++ b/addons/purchase_stock/models/stock.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
-
+from collections import defaultdict
 from odoo import api, fields, models, _
 from odoo.osv.expression import AND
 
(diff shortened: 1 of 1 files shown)


pivi-odoo (commented):
Hello @spco-odoo 👋🏼,

Since this commit is a backport, can you note it in the commit message and PR description with a reference with the commit hash.

An additional improvement can be done to be honest.
The `|=` in a loop is $O(n^2)$, since it re-iterate over the recordset and re-browse on each call.
For faster result, using a `set` of `ids` and 1 browse at the last assignation is preferred. But since this PR is a backport, not sure if we want to adjust this.

If you are just doing the backport without modifications, you can disable the forward-port with `@robodoo fw=no` on a new isolated line, to merge only in 17.
If you plan on adapting with a change, then you can forward port normally with the new changes.

addons/purchase_stock/models/stock.py
```diff
@@ -225,14 +225,18 @@ class StockLot(models.Model):
 
     @api.depends('name')
     def _compute_purchase_order_ids(self):
+        purchase_order_ids_map = defaultdict(set)
+        move_lines = self.env['stock.move.line'].search([
+            ('lot_id', 'in', self.ids),
+            ('state', '=', 'done')
```
pivi-odoo: [formating] [missing trailling-comma](https://docs.astral.sh/ruff/rules/missing-trailing-comma/)

pivi-odoo (approved):
LGTM, thank you for your work.

Just some formatting.

Hello @ajf-odoo, this [PERF] PR is just a slightly modified backport, already pre-reviewed. Feel free to take a look when you have the time.
Cheers.

addons/purchase_stock/models/stock.py
spco-odoo: fixed

ajf-odoo (approved):

spco-odoo:
Hi @pivi-odoo, trailing comma fixed! Ready for merge. Cheers!

addons/purchase_stock/models/stock.py
```diff
@@ -225,14 +225,18 @@ class StockLot(models.Model):
 
     @api.depends('name')
     def _compute_purchase_order_ids(self):
+        purchase_order_ids_map = defaultdict(set)
+        move_lines = self.env['stock.move.line'].search([
```
amoyaux, line 229: Isn't it quicker to directly add the bellow condition in this domain?
('location_dest_id.usage', '=', 'supplier') and ('move_id.purchase_line_id', '!=', False)

spco-odoo: Changed!

PR #247106, opened 2026-02-03
[REF] sql_db: Cursor should not have clear or reset

kmagusiak:
Both of these functions are nearly unused and their descrpition is unclear. They are shortcuts to existing functions on the Transaction object.

---

### DIFF
--- a/odoo/addons/base/tests/test_db_cursor.py
+++ b/odoo/addons/base/tests/test_db_cursor.py
@@ -392,7 +392,7 @@ class TestCursorHooksTransactionCaseCleanup(common.TransactionCase):
 
     def test_4_isolation(self):
         self.assertHookData()
-        self.env.cr.clear()
+        self.env.transaction.clear()
 
     def test_5_isolation(self):
         self.assertHookData()
--- a/odoo/sql_db.py
+++ b/odoo/sql_db.py
@@ -142,8 +142,10 @@ class _FlushingSavepoint(Savepoint):
         super().__init__(cr)
 
     def rollback(self):
-        assert isinstance(self._cr, BaseCursor)
-        self._cr.clear()
+        cr = self._cr
+        assert isinstance(cr, BaseCursor)
+        if cr.transaction is not None:
+            cr.transaction.clear()
         super().rollback()
 
     def _close(self, rollback: bool):
@@ -192,20 +194,6 @@ class BaseCursor(_CursorProtocol):
             self.precommit.run()
         else:
             _logger.warning("Too many iterations for flushing the cursor!")
-
-    def clear(self) -> None:
-        """ Clear the current transaction, and clear precommit hooks. """
-        if self.transaction is not None:
-            self.transaction.clear()
-        self.precommit.clear()
-
-    def reset(self) -> None:
-        """ Reset the current transaction (this invalidates more than `clear()`).
-            This method should be called only right after commit() or rollback().
-        """
-        if self.transaction is not None:
-            self.transaction.default_env = None  # break the cyclic reference
-            self.transaction.reset()
 
     def execute(self, query, params=None, log_exceptions: bool = True) -> None:
         """ Execute a query inside the current transaction.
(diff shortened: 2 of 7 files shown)


odoo/tests/common.py
```diff
@@ -551,7 +551,8 @@ def _raisesContext(self, method, expected_exception, *args, **kwargs):
                     # When checking for an `AccessError`, the cache is cleared
                     # before executing the code. This avoids cache pollution issues and
                     # ensures that access are re-evaluated correctly.
-                    self.env.cr.clear()
+                    if transaction := self.env.cr.transaction:
+                        transaction.clear()
```
HydrionBurst: why not `self.env.transaction.clear()`?

odoo/sql_db.py
```diff
@@ -193,20 +195,6 @@ def flush(self) -> None:
         else:
             _logger.warning("Too many iterations for flushing the cursor!")
 
-    def clear(self) -> None:
-        """ Clear the current transaction, and clear precommit hooks. """
-        if self.transaction is not None:
-            self.transaction.clear()
-        self.precommit.clear()
```
HydrionBurst, line 200: I don't fully understand why we had to and only had to clean `precommit` before.

odoo/addons/base/tests/test_db_cursor.py
```diff
@@ -392,7 +392,7 @@ def test_3_isolation(self):
 
     def test_4_isolation(self):
         self.assertHookData()
-        self.env.cr.clear()
+        self.env.cr.transaction.clear()
```
HydrionBurst:
```suggestion
        self.env.transaction.clear()
```

HydrionBurst (approved):
LGTM

I have a question for the design, which one is lower-level? `Transaction` or `Cursor`

odoo/sql_db.py
kmagusiak: Probably a historical reason. Now it is just confusing. For example, why not clear prerollbacks?

rco-odoo: More calls to `clear()` have been added since it was introduced.  The initial semantics of `clear()` was something like "clear all the stuff related to the current SQL transaction".  As the `precommit` hooks should be run before committing, they contain stuff that will be part of the SQL transaction.  That's why it had to be cleared.

On the other hand, `rollback()` invokes `clear()` before running the `prerollback` hooks.  If you clear the latter, there's nothing left to run...

odoo/sql_db.py
```diff
@@ -552,18 +542,21 @@ def commit(self) -> None:
...
 
     def rollback(self) -> None:
         """ Perform an SQL `ROLLBACK` """
-        self.clear()
+        self.precommit.clear()
         self.postcommit.clear()
         self.prerollback.run()
         self._cnx.rollback()
+        if self.transaction is not None:
+            self.transaction.clear()
```
rco-odoo, line 559: A potential issue with moving `self.transaction.clear()` _after_ the actual rollback is that pre-rollback hooks might use an inconsistent cache and/or leftover recomputations.  But I guess that it is reasonable to assume that pre-rollback hooks will not use stuff from the current transaction, since the transaction must be discarded anyway.  Overall, it's probably the best place to clear the transaction object.

kmagusiak: Yes, I expect to still be inside the transaction during prerollback. If I don't care, then I can use postrollback since the transaction is probably in a bad state anyways.

rco-odoo (approved):
Looks correct to me.

rco-odoo:
@robodoo delegate+

HydrionBurst:
> Yes, there is.  You may have a `Cursor` without a `Transaction`, but you cannot have a `Transaction` without a `Cursor`, since the `Transaction` is related to a set of `Environment`, which must have a `Cursor`.  

If `Transaction` is higher than the `Cursor` I still cannot strictly define what is `transaction.clear()`
but for `transaction.reset()` I expected its result should be as clean as `api.Environment(Registry[dbname].cursor(), api.SUPERUSER_ID, {}).transaction`. Which means I expect a `cr.rollback()` to make the cr is always in a new sql transaction.

PR #247115, opened 2026-02-03
[FIX] calendar: use proper icon syntax for mobile view

andg-odoo:
The proper syntax for buttons with icons is to use the icon="" keyword instead of manually defining the font within a class attribute. When it is implemented via the class attribute any string attribute on the button appears in some random sarif font instead of the standard odoo text font.

Before / After
image image


andg-odoo:
@odoo/rd-subscription-sign-calendar Would it be possible to get a review on this small PR before the freeze? The enterprise has been approved by the other teams just waiting on this to get an approval and ideally R+ across both.

PR #247119, opened 2026-02-04
[IMP] mrp, *: correct typo "Theorical" to "Theoretical"

srap-odoo:
_* = purchase_mrp, sale_mrp

- Fix the misspelling `Theorical time` to `Theoretical time` in the Operation
  Cost Mode selection field to avoid unnecessary confusion.
- Update the tooltip from `Based on Estimated time` to `Based on Theoretical
  time`,
  as the user-facing label refers to theoretical time rather than estimated time.
- Rename the related occurrences in the codebase to maintain consistent terminology
  and help prevent similar issues in the future.

task-5137717

### DIFF
--- a/addons/mrp/models/mrp_routing.py
+++ b/addons/mrp/models/mrp_routing.py
@@ -57,11 +57,11 @@ class MrpRoutingWorkcenter(models.Model):
     cycle_number = fields.Integer("Repetitions", compute="_compute_time_cycle")
     time_total = fields.Float('Total Duration', compute="_compute_time_cycle")
     show_time_total = fields.Boolean('Show Total Duration?', compute="_compute_time_cycle")
-    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theorical time')],
+    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theoretical time')],
                                  string='Cost based on', default='actual', tracking=True,
                                  help="Determines the way Odoo calculates the cost of the operation:\n"
                                  "- Based on Actual time: the cost will be calculated based on tracked time and real employee costs.\n"
-                                 "- Based on Estimated time: the cost will be calculated based on estimated time and costs.")
+                                 "- Based on Theoretical time: the cost will be calculated based on estimated time and costs.")
     cost = fields.Float('Cost', compute="_compute_cost")
 
     @api.depends('time_mode', 'time_mode_batch')
--- a/addons/mrp/models/mrp_workorder.py
+++ b/addons/mrp/models/mrp_workorder.py
@@ -922,7 +922,7 @@ class MrpWorkorder(models.Model):
     def _compute_current_operation_cost(self):
         return (self.get_duration() / 60.0) * (self.costs_hour or self.workcenter_id.costs_hour)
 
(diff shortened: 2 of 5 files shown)


dhha-odoo (approved):

addons/mrp/models/mrp_routing.py
```diff
@@ -57,7 +57,7 @@ class MrpRoutingWorkcenter(models.Model):
     cycle_number = fields.Integer("Repetitions", compute="_compute_time_cycle")
     time_total = fields.Float('Total Duration', compute="_compute_time_cycle")
     show_time_total = fields.Boolean('Show Total Duration?', compute="_compute_time_cycle")
-    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theorical time')],
+    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theoretical time')],
```
ticodoo: Not in stable. It's understandable in English and you'll wipe the existing translations + existing dbs will still see the old string anyways (and won't have a translation anymore if they reload/add a new language)

Also, the help also needs to be updated as well.

Please tell your team to stop trying to fix typos that are still understandable in stable as it not worth it. Only target master.

addons/mrp/models/mrp_routing.py
```diff
@@ -57,11 +57,11 @@ class MrpRoutingWorkcenter(models.Model):
     cycle_number = fields.Integer("Repetitions", compute="_compute_time_cycle")
     time_total = fields.Float('Total Duration', compute="_compute_time_cycle")
     show_time_total = fields.Boolean('Show Total Duration?', compute="_compute_time_cycle")
-    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theorical time')],
+    cost_mode = fields.Selection([('actual', "Actual time"), ('estimated', "Theoretical time")],
```
ticodoo:
```suggestion
    cost_mode = fields.Selection([('actual', 'Actual time'), ('estimated', 'Theoretical time')],
```

Keep it consistent with the rest of logistics code

ticodoo (commented):
For future reference, only user facing typos really matter. Code comments and method names don't

addons/mrp/models/mrp_routing.py
srap-odoo: Hello @ticodoo ,

Thanks for the suggestion! I’ve aligned it with the surrounding logistics code :relaxed: 

I agree that staying consistent within the module is important. Initially, I followed the general convention of using `""` for user-facing strings and `''` for technical ones. While adjusting this, I noticed there are still a few variations in logistics (for example in the following files: [stock_quant.py](https://github.com/odoo/odoo/blob/19.0/addons/stock_account/models/stock_quant.py#L20-L22) and [stock_replenishment_info.py](https://github.com/odoo/odoo/blob/19.0/addons/stock/wizard/stock_replenishment_info.py#L31-L38)), so the convention isn’t entirely uniform yet.

My intention was simply to stay aligned with the coding guidelines. For now, keeping it consistent locally within the module makes more sense to me. It would be great to standardize this more thoroughly at some point, for example by consistently using either `""` or `''` in `Selection` fields throughout the modules.

PR #247127, opened 2026-02-04
[FIX] pos_loyalty: show next-order coupons on receipt after refresh

djip-odoo:
Step to reproduce:
=
- Complete order with Next order coupon 
- Open Order and print receipt (This time coupon is visible)
- Refresh  page Or Reload Data 
- Same order > Print receipt

Issue:
=
- Next time on-wards next order coupon will not shown on the receipt.

Reason:
=
- Loyalty program data is stored in the uistate which destroyed on refresh page OR reload data.

Fix:
=
- Fetched data from the server when uistate doent have any loyalty program data for that order while receipt printing only.

task-5890998
related pr: https://github.com/odoo/enterprise/pull/128730

### DIFF
--- a/addons/pos_loyalty/models/pos_order.py
+++ b/addons/pos_loyalty/models/pos_order.py
@@ -199,6 +199,29 @@ class PosOrder(models.Model):
             'coupon_report': coupon_per_report,
         }
 
+    def get_ticket_screen_order_data(self):
+        data = super().get_ticket_screen_order_data()
+
+        histories = self.env['loyalty.history'].search([
+            ('order_model', '=', 'pos.order'),
+            ('card_id.source_pos_order_id', 'in', self.ids),
+            ('card_id.program_id.program_type', '=', 'next_order_coupons'),
+        ])
+        if not histories:
+            return data
+
+        coupon_data = defaultdict(list)
+        for history in histories:
+            coupon_data[history.card_id.source_pos_order_id.id].append({
+                'program_name': history.card_id.program_id.name,
+                'expiration_date': history.card_id.expiration_date,
+                'code': history.card_id.code,
+            })
+        for order in data["pos.order"]:
+            order["new_coupon_info"] = coupon_data.get(order["id"], [])
+
+        return data
+
     def _check_existing_loyalty_cards(self, coupon_data):
         coupon_key_to_modify = []
         for coupon_id, coupon_vals in coupon_data.items():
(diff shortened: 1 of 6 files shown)


addons/pos_loyalty/models/pos_order.py
```diff
@@ -194,6 +194,24 @@ def confirm_coupon_programs(self, coupon_data):
             'coupon_report': coupon_per_report,
         }
 
+    def get_next_order_coupon_data(self):
+        data = {}
```
paso-odoo:
```suggestion
    def get_next_order_coupon_data(self):
        self.ensure_one()
        data = {}
```

addons/pos_loyalty/models/pos_order.py
```diff
@@ -194,6 +194,24 @@ def confirm_coupon_programs(self, coupon_data):
...
+        histories = self.env['loyalty.history'].search([
+            ('order_id', '=', self.id),
+            ('order_model', '=', 'pos.order'),
+            ('card_id.program_id.program_type', '=', 'next_order_coupons'),
+        ])
+
+        if len(histories) > 0:
```
paso-odoo:
```suggestion
        if histories := self.env['loyalty.history'].search([
            ('order_id', '=', self.id),
            ('order_model', '=', 'pos.order'),
            ('card_id.program_id.program_type', '=', 'next_order_coupons'),
        ]):
```

paso-odoo (changes requested):

paso-odoo (changes requested):
If you are doing RPC calls, offline print doesn't work so, I would suggest to find another way

djip-odoo:
@adgu-odoo Could you please review this PR ? Thanks.

addons/pos_loyalty/static/src/overrides/components/ticket_screen/ticket_screen.js
```diff
@@ -54,4 +54,14 @@ patch(TicketScreen.prototype, {
...
+    async _fetchNotInCachedOrOutdatedOrders(idsNotInCacheOrOutdated) {
+        const orders = await super._fetchNotInCachedOrOutdatedOrders(...arguments);
+        const couponData = await this.pos.data.call("pos.order", "get_next_order_coupon_data", [
+            orders.map(({ id }) => id),
+        ]);
+        for (const order of orders) {
+            order.new_coupon_info = couponData[order.id] ?? [];
+        }
+        return orders;
+    },
```
davidmonnom: You should override the previous request to make only one :)

djip-odoo:
@davidmonnom  Could you please review this PR ? Thanks.

davidmonnom (approved):
@robodoo r+

PR #247128, opened 2026-02-04
[FIX] website: adapt the selector for livechat composer text

maad-odoo:
Purpose of this commit,
To update the selector for asserting the presence of the livechat
chat window in open state.

runbot error -  https://runbot.odoo.com/odoo/error/232912


seb-odoo (approved):
@robodoo r+ delegate+

PR #240263, opened 2025-12-16
19.0 hr attendance negative overtime zirah

ziriraha:
Description of the issue/feature this PR addresses:

Current behavior before PR:

Desired behavior after PR is merged:

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247133, opened 2026-02-04
[REF] hr_attendance: improve readability of unused function

ziriraha:
In odoo/odoo#240263, these functions were duplicated as changing the signature was not possible. In this commit we want to improve the readability of this file by calling the new function inside the duplicated function.


Bertrand2 (approved):
robodoo up to saas-19.1 r+

PR #247134, opened 2026-02-04
[REF] *: remove synchronous assertions in tours

pipu-odoo:
How tours works ?
=================

For each animationFrame, engine searchs the trigger in the DOM until it has been found, then the action (run) is direclty launched.

Why it could be not work properly ?
===================================

Most of time, assertions done in `run` are synchronous. And if the trigger was too generic or clearly present in the DOM (like body, .o_content,...) or was simply the trigger element is already in DOM, then the synchronous assertions can failed... because elements are not present in DOM yet.
Then, either the trigger of this step must be more precise, either you use trigger and HOOT selector to make assertions.
This can cause undeterministic errors.

What we do ?
============

With this commit, we replace synchronous assertions in `run` step function by asynchronous assertions in the `trigger`.
This PR is implemented with the aim of removing the undeterministicTour_doNotCopy key to have tours without delay  between each step.

---

### DIFF
--- a/addons/auth_totp/static/tests/totp_flow.js
+++ b/addons/auth_totp/static/tests/totp_flow.js
@@ -4,20 +4,6 @@ import { registry } from "@web/core/registry";
 import { stepUtils } from "@web_tour/tour_utils";
 import { whenReady } from "@odoo/owl";
 
-function openRoot() {
-    return [{
-        content: "return to client root to avoid race condition",
-        trigger: 'body',
-        run() {
-            document.querySelector("body").classList.add("wait");
-            window.location = '/odoo';
-        },
-        expectUnloadPage: true,
-    }, {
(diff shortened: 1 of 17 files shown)


aab-odoo (approved):

aab-odoo (commented):
robodoo r+

This PR removes error prone usecases of the `run` attribute in tours where assertions on the current state of the DOM were done, synchronously (i.e. as soon as the step is executed). This is error prone because if the trigger isn't correctly defined (which is often the case), those assertions only work thanks to the 50ms delay that the tour system currently waits between each step. That delay is the main source of non determinism in tours and we're about to remove it. Instead of doing assertions in the `run`, this PR defines steps and rely on the tour system itself to wait for the asserted element to be in the DOM before moving on.

PR #229683, opened 2025-10-02
[REF] point_of_sale, pos_*: improve tax computation in PoS

davidmonnom:
*: l10n_es_pos, l10n_sa_pos, point_of_sale, pos_discount,
  pos_loyalty, pos_razorpay, pos_restaurant, pos_self_order

Before this commit taxes computation was a mess in PoS, with differents
implementations in different places, and some of them not taking into
account all the complexity of the tax system (taxes included in price,
taxes on taxes, etc.).

This commit aims to unify all the tax computation logic in a single
place, and to make it more robust and easier to understand.

---

What's changes:

Rounding methods available in PoS are now:
1) Rounding applied only on cash payments
    In this case the remaining due is rounded only if there is at least
    one cash payment line and the remaining due is less than the
    rounding tolerance.

2) Rounding applied on all payment methods
    In this case the remaining due is always rounded even if a card
    payment method is used. The remaining due is rounded if it is less
    than the rounding tolerance. No payment method is rounded in this
    case, the whole order is rounded instead.

Taxes computation is now done globally on the order, and not on each
line. This way we ensure that the total tax amount is always correct,
even if there is some rounding issues on the lines.

---

Changes in tests:

`test_cash_rounding_down_add_invoice_line_not_only_round_cash_method_with_residual_rounding`
`test_cash_rounding_up_add_invoice_line_not_only_round_cash_method`

Are removed because the tested behavior is not longer present. Now when
rounding is enabled with only_round_cash_method=False, the whole order
is always rounded.

Accounting tests are now principally tested with Hoot instead of tours.

---

Developer note:

No prices should be calculated manually in the code base. Getters have
been created for this purpose in the following files:
- `product_template_accounting.js`
- `pos_order_accounting.js`
- `pos_order_line_accounting.js`

In the future, all PRs containing tax calculations must justify them.

taskId: 5143758

PR #247136, opened 2026-02-04
[FIX] point_of_sale: fix runbot error #234810

Lud0do1202:
TASK: [#5897377](https://www.odoo.com/odoo/project/1737/tasks/5897377)

---

The test `point_of_sale:TestPointOfSaleFlow.test_pos_order_refund_ship_delay_totalcost` was failing with the following error:

> You cannot use a credit_note document type with an invoice.

This issue occurred because the refund order was not marked as a refund. As a result, the `account.move` `move_type` was set to `out_invoice` instead of `out_refund`.

Since 19.0, following the change introduced in [odoo/229683](https://github.com/odoo/odoo/pull/229683/files#diff-29cbaebb5b63b539ab173d9340b2aec87b9ad63cd322eec2347879c5412bd50bR850), the `move_type` is no longer determined based on the `pos.order` `amount_total`, but on its `is_refund` field. This field was missing in the test, causing the incorrect behavior.

### DIFF
--- a/addons/point_of_sale/tests/test_point_of_sale_flow.py
+++ b/addons/point_of_sale/tests/test_point_of_sale_flow.py
@@ -1405,6 +1405,7 @@ class TestPointOfSaleFlow(CommonPosTest):
                 'name': fields.Datetime.now(),
                 'payment_method_id': self.cash_payment_method.id
             }]],
+            'is_refund': True,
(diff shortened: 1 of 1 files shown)


davidmonnom (approved):
@robodoo r+

PR #247141, opened 2026-02-04
[IMP] mail, *: unify model and field definitions across bundles

maki-odoo:
The JS models field definitions are applied once when the page loads. If the first bundle is loaded and a second bundle is loaded later, the field definitions in the second bundle will be missed. For example, this could happen when the live chat is loaded first and the portal chatter is loaded later.

This change ensures that all model and field definitions are loaded together by including the `common` folder in all related bundles, regardless of the module in which they are defined.

task-5895454

[Enterprise PR](https://github.com/odoo/enterprise/pull/106436)


seb-odoo:
We might want a little bit more information in the commit and PR message.

addons/rating/static/src/core/common/thread_model_patch.js
```diff
@@ -0,0 +1,11 @@
+import { Thread } from "@mail/core/common/thread_model";
+
+import { patch } from "@web/core/utils/patch";
+
+patch(Thread.prototype, {
+    setup() {
+        super.setup();
+        /** @type {{ avg: number, total: number, percent: Object<number, number>}}*/
+        this.rating_stats;
```
seb-odoo, line 10: This should probably be moved to `rating` module as this is where it is defined/sent in python.

addons/website_livechat/__manifest__.py
```diff
@@ -20,7 +20,7 @@
     ],
     'assets': {
         "im_livechat.assets_embed_core": [
-            "website/static/src/**/common/**/*",
+            "website/static/src/mail/**/common/**/*",
```
seb-odoo: website files should be added from website manifest

addons/website_livechat/__manifest__.py
```diff
@@ -34,8 +34,12 @@
         "mail.assets_public": [
             "website_livechat/static/src/**/common/**/*",
         ],
+        "portal.assets_chatter_helpers": [
+            "website/static/src/mail/**/common/**/*.js",
```
seb-odoo: same for all of them

addons/rating/__manifest__.py
```diff
@@ -30,7 +30,10 @@
         "mail.assets_public": [
             "rating/static/src/core/common/**/*",
         ],
-        "portal.assets_chatter": [
+        "portal.assets_chatter_helpers": [
+            "rating/static/src/core/common/**/*",
+        ],
+        "im_livechat.assets_embed_core": [
```
seb-odoo: I would place them alphabetically (as best as possible, at least among themselves, don't change the web lines).

addons/mail/static/src/chatter/web/scheduled_message_model_patch.js
```diff
@@ -1,60 +1,49 @@
-import { fields, Record } from "@mail/model/export";
+import { ScheduledMessage } from "@mail/chatter/common/scheduled_message_model";
+import { fields } from "@mail/model/export";
 import { htmlToTextContentInline } from "@mail/utils/common/format";
+
 import { _t } from "@web/core/l10n/translation";
+import { patch } from "@web/core/utils/patch";
+
+const ScheduledMessagePatch = {
```
seb-odoo, line 9: It would be nice to type it.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -338,6 +341,11 @@ export class DiscussChannel extends Record {
             }
         },
     });
+    isLocallyPinned = fields.Attr(false, {
+        onUpdate() {
+            this.onPinStateUpdated();
+        },
+    });
```
seb-odoo: I'm really not sure for this one, it only really makes sense in `public_web`.
As a simple boolean field, it won't make issue if the python sends it while it is not defined.

addons/mail/static/src/chatter/common/scheduled_message_model.js
```diff
@@ -0,0 +1,25 @@
+import { fields, Record } from "@mail/model/export";
+
+export class ScheduledMessage extends Record {
+    static _name = "mail.scheduled.message";
+    static id = "id";
```
alexkuhn: `static id = "id"` is the default, no need to define it anymore.

addons/mail/static/src/chatter/common/scheduled_message_model.js
```diff
@@ -0,0 +1,25 @@
...
+    /** @type {Object.<number, import("models").ScheduledMessage>} */
+    static records = {};
+    /** @returns {import("models").ScheduledMessage} */
+    static get(data) {
+        return super.get(data);
+    }
```
alexkuhn: Not need to define `static records` and `static get` for typing, this works automatically for all models thanks to some logic in `models.d.ts` on `Record`.

This is just a code move, but can take the opportunity to clean this part.

addons/website/static/tests/builder/chat_hub.test.js
```diff
@@ -10,7 +10,7 @@ import { CHAT_HUB_DEFAULT_BUBBLE_START } from "@mail/core/common/chat_hub_model"
 import { animationFrame, describe, expect, queryFirst, test } from "@odoo/hoot";
 import { onRendered } from "@odoo/owl";
 import { getService, patchWithCleanup } from "@web/../tests/web_test_helpers";
-import { CHAT_HUB_WE_SIDEBAR_WIDTH } from "@website/common/chat_hub_model_patch";
+import { CHAT_HUB_WE_SIDEBAR_WIDTH } from "@website/mail/core/common/chat_hub_model_patch";
```
alexkuhn, line 13: Nitpick: I don't think the `core/` is worth it. `mail/` is already relating to a feature folder.

The `core/` we use in `@mail` is for some arbitrary split with other folders, but when no sub-folder I would not artificially introduce one now and postpone to when we actually need sub-folders.

addons/im_livechat/__manifest__.py
```diff
@@ -183,6 +183,9 @@
             "im_livechat/static/src/core/common/**/*",
             "im_livechat/static/src/core/public_web/**/*",
         ],
+        "portal.assets_chatter_helpers": [
+            "im_livechat/static/src/core/common/**/*.js",
+        ],
```
alexkuhn, line 189: Why precise `.js`? If we can allow all files in this folder that would be better.
Same remark in other manifests related to `portal.assets_chatter_helpers` too.

addons/calendar/static/src/core/web/activity_model_patch.js
seb-odoo, line 1: `rescheduleMeeting` uses action service so it must say in `web`

addons/hr/static/src/core/common/store_service_patch.js
seb-odoo, line 1: For me this should be common as `getChat` is `common` in `mail` too

addons/hr_holidays/static/src/core/web/avatar_card_popover_patch.xml
seb-odoo, line 1: Note: ok for now, but avatar card should be moved to `common` in `mail` eventually, there is no reason to limit that feature to `web`.

addons/hr_holidays/static/src/core/web/im_status_patch.xml
seb-odoo, line 1: This file has recently been removed in master (otherwise should be common)

addons/hr_holidays/static/src/core/common/res_partner_model_patch.js
seb-odoo, line 1: Should be `common`, no reason to limit `outOfOfficeDateEndText` to `web` context

addons/im_livechat/__manifest__.py
maki-odoo: There may be some CSS variables in the liveChat common bundle that we don't use in the Portal, such as `o-mail-NavigableList-zIndex`. Since the aim of the clean up is to integrate the model and fields, I excluded  the shared part to the js files for now.

addons/hr_holidays/static/src/core/web/store_service_patch.js
seb-odoo, line 1: should be removed recently https://github.com/odoo/odoo/pull/210189

addons/hr_holidays/static/src/core/common/thread_patch.xml
seb-odoo, line 1: should be `common`

addons/hr_holidays/static/src/core/web/store_service_patch.js
seb-odoo: new code about im status introduced in that PR should all be moved to common (hr holidays, hr homeworking, ...)

addons/im_livechat/static/src/core/common/thread_model_patch.js
```diff
@@ -1,13 +1,27 @@
...
 import { Thread } from "@mail/core/common/thread_model";
+import { convertBrToLineBreak } from "@mail/utils/common/format";
 
 import { patch } from "@web/core/utils/patch";
 import { url } from "@web/core/utils/urls";
 
 patch(Thread.prototype, {
     setup() {
         super.setup();
+        this.country_id = fields.One("res.country");
```
seb-odoo, line 11: Ok as a separate PR as it might have more implications but all of these fields should be moved from thread to channel model to match python.

addons/rating/static/src/core/common/thread_model_patch.js
```diff
@@ -0,0 +1,11 @@
+import { Thread } from "@mail/core/common/thread_model";
+
+import { patch } from "@web/core/utils/patch";
+
+patch(Thread.prototype, {
```
seb-odoo: Missing type import (not a big deal here, just for consistency)

addons/website/__manifest__.py
```diff
@@ -227,8 +227,14 @@
     'post_init_hook': 'post_init_hook',
     'uninstall_hook': 'uninstall_hook',
     'assets': {
+        "im_livechat.assets_embed_core": [
+            "website/static/src/mail/**/common/**/*",
```
seb-odoo: Note: when using `**` for feature, we need to always explicitly list `core` above to ensure `core` is loaded before feature-specific.

If there is only `core`, don't use `**` in the feature slot.

addons/website/static/tests/builder/chat_hub.test.js
seb-odoo: I would not make the suggested change from Alex, it's much better if patches follow the same structure as the files they are patching. (So current diff is good for me)

seb-odoo (approved):
Looking ok except comments

addons/im_livechat/__manifest__.py
alexkuhn: It's not used but by limiting to `*.js` now, this means any `*.css` / `*.xml` new file would necessarily have to adapt this manifest. If this is not a problem to have these files imported, I think it's best to pick all file types of the given folder.

maki-odoo: I need to define such variables for chatter shadow dom if I import those files which I thought it's not in the spec of this change?

addons/website_livechat/__manifest__.py
```diff
@@ -32,20 +31,20 @@
             "website_livechat/static/tests/mock_server/**/*",
         ],
         "mail.assets_public": [
-            "website_livechat/static/src/**/common/**/*",
+            "website_livechat/static/src/core/common/**/*",
         ],
-        'website.assets_wysiwyg': [
-            "website/static/src/**/common/**/*",
+        "portal.assets_chatter_helpers": [
+            "website_livechat/static/src/core/common/**/*.*",
```
seb-odoo: `*.*` is on purpose? we never have that pattern anywhere else, probably to remove

seb-odoo (approved):
@robodoo delegate+

There are conflicts to resolve again.

PR #55391, opened 2020-08-04
[FIX] survey: 403 Forbidden when printing survey

simongoffin:
Steps to reproduce the bug:

- Let's consider a survey S
- Set the scoring option of S to Scoring without answers at the end
- print S

Bug:

A 403 error was raised.

opw:2307495

PR #247143, opened 2026-02-04
[IMP] survey: always shows answers when printing participant inputs

pko-odoo:
Shows answers on printing report whether the survey_type. Before the answers were not shown when survey_type was scoring_without_answers.

Task-5407701

---

### DIFF
--- a/addons/survey/controllers/main.py
+++ b/addons/survey/controllers/main.py
@@ -671,7 +671,7 @@ class Survey(http.Controller):
             'is_html_empty': is_html_empty,
             'review': review,
             'survey': survey_sudo,
-            'answer': answer_sudo if survey_sudo.scoring_type != 'scoring_without_answers' else answer_sudo.browse(),
+            'answer': answer_sudo,
             'questions_to_display': answer_sudo._get_print_questions(),
             'scoring_display_correction': survey_sudo.scoring_type in ['scoring_with_answers', 'scoring_with_answers_after_page'] and answer_sudo,
             'format_datetime': lambda dt: format_datetime(request.env, dt, dt_format=False),


addons/survey/controllers/main.py
```diff
@@ -671,7 +671,7 @@ def survey_print(self, survey_token, review=False, answer_token=None, **post):
             'is_html_empty': is_html_empty,
             'review': review,
             'survey': survey_sudo,
-            'answer': answer_sudo if survey_sudo.scoring_type != 'scoring_without_answers' else answer_sudo.browse(),
+            'answer': answer_sudo,
```
julien-banken, line 674: Hello @pko-odoo 👋

Thank you for your PR. Just to be sure: If we now print the survey with the answers than, is it still relevant to have the `scoring_without_answers` option? What will be the difference between `scoring_without_answers` and `scoring_with_answers`?

**If the option is still relevant, then the change looks good to me** :+1: I didn't achieve to reproduce the error described in the commit message adding the condition we are removing (https://github.com/odoo/odoo/pull/55391).

For the PR/commit message, we could clarify the commit message a little bit:

"Previously, printing surveys of type `scoring_without_answers` omitted participants' answers, resulting in incomplete and inconsistent reports. This commit ensures answers are always included, regardless of survey type."

We could also target a stable version as we do not change the function's API?

pko-odoo: `scoring_without_answers` is still useful, we want to display the answers only when printing the survey.

julien-banken:
These changes have been migrated to a new PR targeting an earlier version.
see: https://github.com/odoo/odoo/pull/258616

PR #247145, opened 2026-02-04
[FIX] portal, website_slides: boot portal chatter with an Interaction

Dj0ulo:
### Before commit
Tour `test_fullscreen_slide_text_highlights` sometimes shows the following warning:
```
should not have any "characterData", "remove" or "add" mutations in
current step when you update the selection
```

### Cause
`PortalChatterService` adds an element in the DOM while the selection is updated in the `selectText` tour step.

### Fix
Replace the `boot_service` of the chatter by an Interaction inside which the `#chatterRoot` element is created to leverage the protection that the Interaction mechanism provides against this kind of issue.

Replace the deprecated Deferred `portalChatterReady` by a `Promise.withResolvers` and fix the `portal_chatter_bundle` tour to properly wait for it to resolve.

runbot-229803


addons/portal/static/src/chatter/boot/portal_chatter_boot.js
```diff
@@ -0,0 +1,21 @@
+import { Interaction } from "@web/public/interaction";
+import { registry } from "@web/core/registry";
+import { loadBundle } from "@web/core/assets";
+import { memoize } from "@web/core/utils/functions";
+
+odoo.portalChatterReady = Promise.withResolvers();
```
seb-odoo, line 6: Change this API doesn't look necessary at all in stable.

addons/portal/static/src/chatter/frontend/portal_chatter_service.js
```diff
@@ -34,12 +34,10 @@ export class PortalChatterService {
             twoColumns: chatterEl.getAttribute("data-two_columns") === "true" ? true : false,
             displayRating: chatterEl.getAttribute("data-display_rating") === "True" ? true : false,
         };
-        const root = document.createElement("div");
-        root.setAttribute("id", "chatterRoot");
+        const root = chatterEl.querySelector("#chatterRoot");
```
seb-odoo, line 37: Is this move necessary?

seb-odoo (changes requested):
Hello, thank you for the PR.

I had in mind we should clean it up in master, I'm not really sure if it's safe for stable.

I would definitely not remove the boot service code in stable.
Making it inoperative might be acceptable.

addons/portal/static/src/chatter/boot/portal_chatter_boot.js
Dj0ulo: Indeed, I can revert this back

addons/portal/static/src/chatter/frontend/portal_chatter_service.js
Dj0ulo: This is actually the point of the PR. Adding the element using the service can cause an nondeterministic bug with the tour `test_fullscreen_slide_text_highlights` when the website builder is open.

Moving that logic inside an interaction prevents this bug. But if you feel like it is unsafe in stable I can find another workaround :wink:

seb-odoo: What would be the workaround? It'd be nice to compare PRs if you have it.

Dj0ulo: I can just add a step in the tour to wait for `chatterRoot` just to fix the runbot error. This should do the trick (not tried yet). And so I let you guys clean the chatter as you wish in master :+1:

seb-odoo: Yes please, note that you can just target this PR to master as you already have done the work.

And make the simpler one for fixing runbot in stable.

seb-odoo (commented):
I'm wondering if the `createShadow` and maybe other parts of the service shouldn't also be moved to the interaction?

seb-odoo (approved):
We can do it step by step, if this already works it can be done now.

@robodoo delegate=maki-odoo

maki-odoo:
@robodoo r+

PR #247149, opened 2026-02-04
[FIX] api_doc: list parameter are not parsed fully

PiaCOS:
In python 3.10;
When the type of a method parameter is a container (eg: `list[T]`), `stringify_signature` was not getting the full type but only the origin of it (eg: `list` instead of `list[int]`).

This commit fixes the issue by stringifiying the container types separately.

runbot-237782


Julien00859 (approved):

kmagusiak (approved):
LGTM if this passes runbot.
This can be forwarded further to keep the test in newer versions.

Julien00859 (approved):
@robodoo r+

PR #247150, opened 2026-02-04
[FIX] stock: prevent invisible serial number when assigning generic stock

amahm-odoo:
Steps to reproduce:
1. Create a storable product with tracking set to 'By Quantity'.
2. Update the Quantity on Hand (e.g., 100 units).
3. Change the product tracking to 'By Serial Number'.
4. Create a Repair Order for this product.
5. Add a line, select a specific Serial Number, and click Save.
6. Observe that the serial number disappears.

Cause:
When reserving stock that was originally created as 'Generic' (no serial), the `_prepare_move_line_vals` method returns `lot_id=False`. The repair view uses `_compute_lot_ids` to display selected lots, which filters out any move lines where `lot_id` is False. This causes the new line to be effectively invisible to the UI immediately after creation.

Solution:
In the `_set_lot_ids` inverse method, explicitly force the `lot_id` into the create values dictionary (`move_line_vals`). This ensures that even if Odoo reserves generic stock, the resulting move line is born with the correct Serial Number identity, keeping it visible and valid.

opw-5156267

### DIFF
--- a/addons/stock/tests/test_move_lines.py
+++ b/addons/stock/tests/test_move_lines.py
@@ -4,6 +4,7 @@
 import datetime
 from freezegun import freeze_time
 
+from odoo import Command
 from odoo.addons.stock.tests.common import TestStockCommon
 from odoo.tests import Form
 from odoo.exceptions import UserError
@@ -236,3 +237,37 @@ class StockMoveLine(TestStockCommon):
             })
             # 36 units > 24 units
             self.assertTrue(ml.date > update_date_3, "Quantity change check for date should take into account UoM conversion")
+
+    def test_lot_creation_from_move_line_with_generic_stock(self):
+        """
+        Test that if the product already have quantities and after that tracking is set to serial,
+        we can create a lot and assign it to the move.
+        """
+        self.productA.tracking = "none"
+
+        self.env["stock.quant"]._update_available_quantity(self.productA, self.env["stock.location"].browse(self.stock_location), 100)
+
+        self.productA.tracking = "serial"
+
+        serial_lot = self.env["stock.lot"].create(
+            {
+                "name": "SN001",
+                "product_id": self.productA.id,
+                "company_id": self.env.company.id,
+            }
+        )
+
+        move = self.env["stock.move"].create(
+            {
+                "name": "Test Repair Move",
+                "product_id": self.productA.id,
+                "product_uom": self.productA.uom_id.id,
+                "product_uom_qty": 1.0,
+                "location_id": self.stock_location,
+                "location_dest_id": self.customer_location,
+                "lot_ids": [Command.link(serial_lot.id)],
+            }
+        )
+
+        line = move.move_line_ids[0]
+        self.assertEqual(line.lot_id, serial_lot)
--- a/addons/stock/models/stock_move.py
+++ b/addons/stock/models/stock_move.py
@@ -597,7 +597,7 @@ Please change the quantity done or the rounding precision of your unit of measur
                         mls_without_lots -= move_line
                     else:  # No line without serial number, creates a new one.
                         reserved_quants = self.env['stock.quant']._get_reserve_quantity(move.product_id, move.location_id, 1.0, lot_id=lot)
-                        if reserved_quants:
+                        if reserved_quants and reserved_quants[0][0].lot_id:
                             move_line_vals = self._prepare_move_line_vals(quantity=0, reserved_quant=reserved_quants[0][0])
                         else:
                             move_line_vals = self._prepare_move_line_vals(quantity=0)


adwid (changes requested):
I understand your idea but basing some SML values (such as its source location) on an incorrect quant is not the best idea. We should rather fine-tune the if-condition so we don't enter if the quant doesn't have what we want, aka the lot

About the description, I think it's possible to explain better why we have a falsy `lot_id`. Also, "_The repair view uses \_compute\_lot\_ids_"  is not very correct: the view simply displays a SM, one of its field is `lot_ids` and the latter is a computed one. So, when displaying the record, it will trigger the compute method. Anyway, it could give something like

> When defining a lot, it triggers the inverse method of the field in order to create the related SML. When preparing the SML values, we first look for a quant. However, the lot condition allows its value to be `False`, so we find the quant without any defined lot. As a result, the SML values (i.e., the result `_prepare_move_line_vals`) doesn't have any value for `lot_id`. And this explains why the lot "disappears": the line is actually a SM and its lots are the lots of the SML. So, when computed, the field will become empty, hence the "disappearance"

No need to explain what the commit does, based on the explanations of the issue, it will be straightforward to understand your fix 👌 

Last but not least:
Can you add a test ? This is really important in the bugfix process, don't forget the test section of our [documention](https://www.odoo.com/odoo/knowledge/10288) 😄

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
             })
             # 36 units > 24 units
             self.assertTrue(ml.date > update_date_3, "Quantity change check for date should take into account UoM conversion")
+
+    def test_lot_creation_from_move_line_with_generic_stock(self):
```
adwid, line 241: Can you add a docstring that broadly explains what is the protected case here

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+        product = self.env["product.product"].create(
```
adwid: Better to use existing data whenever it's possible, even if you have to touch it a bit to match your expectations

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+            {
+                "name": "Test Serial Product",
+                "is_storable": True,
+                "tracking": "none",
+            }
+        )
+
+        self.env["stock.quant"]._update_available_quantity(
+            product, self.env["stock.location"].browse(self.stock_location), 100
+        )
```
adwid: You can write it in oneline, it's not that long

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+
+        product.write({"tracking": "serial"})
```
adwid:
```suggestion
        product.tracking = "serial"
```

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+                "name": "Test Repair Move",
+                "product_id": product.id,
+                "product_uom": product.uom_id.id,
+                "product_uom_qty": 1.0,
+                "location_id": self.stock_location,
+                "location_dest_id": self.customer_location,
+            }
+        )
+
+        move.lot_ids = [serial_lot.id]
```
adwid: You can define `lot_ids` during the above SM creation
Also, use `Command` when dealing with x2m fields

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+
+        self.assertEqual(len(move.move_line_ids), 1, "A move line should have been created.")
```
adwid: This assert is implicitly implied by the next one, which makes it useless

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +236,44 @@ def test_move_line_date(self):
...
+
+        line = move.move_line_ids[0]
+        self.assertEqual(line.lot_id, serial_lot, "The Move Line must have the Serial Number assigned.")
```
adwid: Most of the time the assert message is useless
```
self.assertEqual(my_variable, 4, "My variable should be equal to 4"
```
This kind of message doesn't bring any additional/relevant information
Actually the asserts are very often clear enough and don't need any msg
Write it only when you need to explain the expected value

addons/stock/tests/test_move_lines.py
```diff
@@ -236,3 +237,37 @@ def test_move_line_date(self):
             })
             # 36 units > 24 units
             self.assertTrue(ml.date > update_date_3, "Quantity change check for date should take into account UoM conversion")
+
+    def test_lot_creation_from_move_line_with_generic_stock(self):
+        """
+        Test that if the product already have quantities and after that tracking is set to serial,
+        we can create a lot and assign it to the move line.
```
adwid:
```suggestion
        we can create a lot and assign it to the move
```
Sorry, very small details, but in stock SM is not the same as SML
Here you're dealing with SM

PR #247151, opened 2026-02-04
[ADD] core: checker for methods being set on models during tests

xmo-odoo:
Setting methods on models during tests and not removing them afterwards can cause misbehaviour of other tests. Setting mock or test-specific methods is not an issue but it should be done via patchers.

https://runbot.odoo.com/odoo/error/237983

### DIFF
--- a/odoo/tests/common.py
+++ b/odoo/tests/common.py
@@ -29,6 +29,7 @@ import tempfile
 import threading
 import time
 import traceback
+import types
 import unittest
 import warnings
 from collections import defaultdict, deque
@@ -419,16 +420,12 @@ class BaseCase(case.TestCase, metaclass=MetaCase):
 
     def patch(self, obj, key, val):
         """ Do the patch ``setattr(obj, key, val)``, and prepare cleanup. """
-        patcher = patch.object(obj, key, val)   # this is unittest.mock.patch
-        patcher.start()
-        self.addCleanup(patcher.stop)
+        self.startPatcher(patch.object(obj, key, val))   # this is unittest.mock.patch
 
     @classmethod
     def classPatch(cls, obj, key, val):
         """ Do the patch ``setattr(obj, key, val)``, and prepare cleanup. """
-        patcher = patch.object(obj, key, val)   # this is unittest.mock.patch
-        patcher.start()
-        cls.addClassCleanup(patcher.stop)
+        cls.startClassPatcher(patch.object(obj, key, val))   # this is unittest.mock.patch
 
     def startPatcher(self, patcher):
         mock = patcher.start()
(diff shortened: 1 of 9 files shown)


AntoineVDV (approved):
FYI the `website_sale_comparison` module has been merged into `website_sale` in master (19.2).

carlos-lopez-tecnativa:
@xmo-odoo I investigated this more in depth and noticed that, with your changes, you are generating a dictionary with the `class` of each model as the key: https://github.com/odoo/odoo/blob/ecb55de1b7c896fdccdc15e861e6612e70cc209f/odoo/tests/common.py#L1022-L1034
. In the `odoo-test-helper` library, we alter the registry to load new (fake) models, so when it tries to retrieve the original model values https://github.com/odoo/odoo/blob/ecb55de1b7c896fdccdc15e861e6612e70cc209f/odoo/tests/common.py#L1133C35-L1133C47, it does not get the original model because it has been changed. Therefore, the class ID or hash is different, and the attributes of the old class are compared with those of the new one.

Is it possible to use the `model name` as a `string` key instead of the `class` itself?

I tried to apply a patch where we only alter the new models and not the existing ones, and it works partially, but in some OCA repositories existing models are patched, and since they are not reloaded in the registry, it fails.
https://github.com/OCA/odoo-test-helper/pull/39/changes/e7419bb624e941e91cdd1aab6bfb31122ee84c7e

odoo/tests/common.py
```diff
@@ -1012,6 +1069,8 @@ def _crypt_context(self):  # noqa: ARG001
 
     def setUp(self):
         super().setUp()
+        self.setattrs.clear()
+        self.addCleanup(self.check_attrs)
```
rousseldenis, line 1073: @xmo-odoo So, this means we cannot do a setup (e.g.: introducing a fake model) at test class level but at instance only. This will affect a lot of tests in terms of performances (as creating tests setup is done before each test instance instead of once).

rousseldenis: @pedrobaeza @carlos-lopez-tecnativa @lmignon @sbidoul @rco-odoo

xmo-odoo: > So, this means we cannot do a setup (e.g.: introducing a fake model) at test class level but at instance only.

No, it does not. It does mean that you have to either update the attribute sets or override `check_attrs` to neuter it.

rousseldenis: @carlos-lopez-tecnativa Could you try to implement that in odoo-test-helper in order to avoid changing all our tests flows ? Thanks

carlos-lopez-tecnativa: > @carlos-lopez-tecnativa Could you try to implement that in odoo-test-helper in order to avoid changing all our tests flows ? Thanks

@rousseldenis At Tecnativa, we invested time in analyzing the issue and fixing the repositories with failing CI. The problem has now been resolved, and we understand the performance concerns you mentioned.

However, at this point, we’re unable to allocate additional time to further improvements in this area. Please feel free to propose any changes you believe could help enhance performance, and we’ll be happy to review them.

rousseldenis: It's more a workaround than a real fix on top of Odoo change as we need to change all tests flows that use that logic.

Nevertheless, we will find a solution for that.

Thanks

PR #247153, opened 2026-02-04
[FIX] html_builder, *: disable image filters when WebGL missing

robinlej:
*: html_editor

This commit backports [3817cdd], which was properly adapted in 19.0. In 18.4 ([c943a88]), it was kept by mistake as in previous versions, prior to the html_builder refactoring. As a result, the fix didn't work in this specific version.

Original explanation:
On recent versions of Chrome for Linux (v140+), the old SwiftShader software fallback for WebGL has been removed. As a result, new window.WebGLImageFilter() now throws if no GPU context is available, typically when WebGL is disabled or unsupported.

Since the application cannot enable WebGL from JavaScript, this commit improves the user experience by detecting the absence of a WebGL context early and disabling image filters in edit mode. Instead of raising a traceback, the editor now skips the filter feature and can optionally display a friendly message explaining that WebGL is required to use image filters.

This avoids runtime errors and ensures a more robust behavior on platforms where WebGL is unavailable.

[3817cdd]: https://github.com/odoo/odoo/commit/3817cdd3bc096e62dbd017a85356a8032d5217bd
[c943a88]: https://github.com/odoo/odoo/commit/c943a882624900743f02719560b70a0a683f095a

task-5117584

### DIFF
--- a/addons/html_builder/static/src/core/building_blocks/builder_row.js
+++ b/addons/html_builder/static/src/core/building_blocks/builder_row.js
@@ -22,6 +22,7 @@ export class BuilderRow extends Component {
         extraLabelClass: { type: String, optional: true },
         observeCollapseContent: { type: Boolean, optional: true },
         fullRowToggler: { type: Boolean, optional: true },
+        disabled: { type: Boolean, optional: true },
     };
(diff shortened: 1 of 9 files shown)


robinlej:
@robodoo up to saas-18.4

PR #247157, opened 2026-02-04
[IMP] l10n_in: Rounding TDS amount

jva-odoo:
Normally TDS is round up and also TDS challan not support float amount And when user use case rounding for simplify payment amount but if they deduct tds it's not work

task - 4908458

---

### DIFF
--- a/addons/l10n_in/models/account_tax.py
+++ b/addons/l10n_in/models/account_tax.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
 
 from odoo import api, fields, models
-from odoo.tools import frozendict
+from odoo.tools import frozendict, float_round
 
 
 class AccountTax(models.Model):
@@ -141,3 +141,9 @@ class AccountTax(models.Model):
                 for key, values in items_map.items()
             ],
         }
+
+    def _eval_tax_amount_price_excluded(self, batch, raw_base, evaluation_context):
+        res = super()._eval_tax_amount_price_excluded(batch, raw_base, evaluation_context)
+        if self.l10n_in_tax_type in ('tds_sale', 'tds_purchase'):
+            res = float_round(res, precision_digits=0, rounding_method='UP')
+        return res
--- a/addons/l10n_in/tests/__init__.py
+++ b/addons/l10n_in/tests/__init__.py
@@ -6,3 +6,4 @@ from . import test_check_status
 from . import test_tds_tcs_alert
 from . import test_gstr_section
 from . import test_invoice_label
+from . import test_tds_calculation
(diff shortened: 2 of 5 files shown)


addons/l10n_in/models/account_tax.py
```diff
@@ -141,3 +141,9 @@ def grouping_function(base_line, tax_data):
                 for key, values in items_map.items()
             ],
         }
+
+    def _eval_tax_amount_price_excluded(self, batch, raw_base, evaluation_context):
```
jco-odoo, line 145: The problem of this method is that it changes the core of Odoo tax calculation, so even if you create an entry with this TDS manually or you put a TDS on an invoice, ... it will do this rounding with possible side effects. Also, in theory, the PoS should have it in the js.  

However, another option is to do the rounding in the wizard and then override the manual amount when generating the accounting entry.  In the base line dict calculation, you have the manual_tax_mounts dict where you can set tax_id as key and the amount you are overriding.  Then you can use fix_base_lines_tax_details_on_manual_tax_amount to set the extra_tax_data Json field on the account.move.line and that way you do not have to worry about that your values would be overwritten by the sync_dynamic_lines.  (need to check a bit deeper still myself)

jva-odoo: > The problem of this method is that it changes the core of Odoo tax calculation, so even if you create an entry with this TDS manually or you put a TDS on an invoice

That we need even if they create TDS entry manually or modify TDS entry same is applied.
Put on invoice is suggested and preferable becosue they can't see that in partner ledger as separate entry 
Use need always rounded value of this tax.


> it will do this rounding with possible side effects. Also, in theory, the PoS should have it in the js.

There is no use case to use this tax in POS.

> However, another option is to do the rounding in the wizard and then override the manual amount when generating the accounting entry. In the base line dict calculation, you have the manual_tax_mounts dict where you can set tax_id as key and the amount you are overriding. Then you can use fix_base_lines_tax_details_on_manual_tax_amount to set the extra_tax_data Json field on the account.move.line and that way you do not have to worry about that your values would be overwritten by the sync_dynamic_lines. (need to check a bit deeper still myself)

This is very complex method do it like that and also drawback is that if user modify entry then it's not consider rounding.
**So if you like to go with generic methods then we need add support of case rounding where type is entry.
or even we can do it only in `l10n_in` where type is entry and `l10n_in_is_withholding` is true.**

but from this all solution what I do it's very simple way to do it.
I know you feed this is more hacky way but also very simple way :wink:

jco-odoo: @jva-odoo Is there a real use case where people need to do it manually?  Because they can always just cancel the entry and create a new one. I tried it out a bit and it seems to work, but it is cumbersome with the automatic balancing, ...

Image

jva-odoo: that is not big use case so you can ignore manually edit thing.

but what you think about allowed case rounding in this TDS entry.

PR #247158, opened 2026-02-04
[FIX] point_of_sale: disallow multiple order by refund

davidmonnom:
Before this validation, it was possible to refund multiple orders per refund, but this resulted in a traceback if some orders were invoiced.

This validation blocks the possibility of having multiple orders per refund, as in the following versions.

opw-4791231


stva-odoo:
robodoo r+

PR #247159, opened 2026-02-04
[IMP] core: colorize PID in logs

Julien00859:
The logs use the following format:

{date} {time} {pid} {level} {database} {module}: {message}

Before the commit, only the `{level}` part of the message, and the request line in case of of a HTTP request, was colored.

In this work we also color the `{pid}` so it is possible at a glance to determine the worker/thread that emitted the log record.

There are two reasons for this change:

1. To be able to follow the logs of a specific server, when multiple servers are emitting logs at a same time.
2. To be able to quickly determine what were the log messages emitted during the processing of a specific HTTP request.

We noted that many many developers think that the HTTP request line is logged *upon receiving the request* but it actually is logged *after* the response is sent. It means that all the logs of a specific request actually appear *before* the logged request line. We hope that this new colored PID adds a welcome cue so that devs read the lines *above* the HTTP request line they care about, and not *bellow*.

Current behavior before PR:

---

### DIFF
--- a/odoo/netsvc.py
+++ b/odoo/netsvc.py
@@ -102,6 +102,7 @@ RESET_SEQ = "\033[0m"
 COLOR_SEQ = "\033[1;%dm"
 BOLD_SEQ = "\033[1m"
 COLOR_PATTERN = f"{COLOR_SEQ}{COLOR_SEQ}%s{RESET_SEQ}"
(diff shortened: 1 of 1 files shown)


kmagusiak (approved):

Julien00859:
image

kmagusiak (approved):
@robodoo r+

PR #247162, opened 2026-02-04
[IMP] core: stop using werkzeug's dev server in production

Julien00859:
We are using werkzeug's development server as our production HTTP server, both in multi-thread and multi-process (worker mode). Werkzeug's documentation recommends **strongly against** using that server in production.

> Do not use the development server when deploying to production. It is intended for use only during local development. It is not designed to be particularly secure, stable, or efficient.[^1]

> The development server is not intended to be used on production systems. It was designed especially for development purposes and performs poorly under high load.[^2]

[^1]: https://werkzeug.palletsprojects.com/en/stable/deployment/
[^2]: https://werkzeug.palletsprojects.com/en/stable/serving/

When Antony developped the Odoo server, the focus was put on a scalable multithread/multiworker architecture and he used werkzeug as an off-the-shelve free/simple HTTP server. The goal of this PR is to leverable the pretty good multithread/multiworker architecture we have, and to only replace the HTTP server.

H11 is a sans-io implementation of the HTTP/1.0 and 1.1 protocol stack for Python. That it is "sans-io" means that the library perform no socket operation such as `recv` or `send` on its own, the user of the lib is responsible to bring the bytes to and from H11. That it implements the protocol stack means that the lib understand HTTP and will make sure all messages from/to the lib strictly follow the protocol, and it does a damn good job at handling all the HTTP quirks. It offers the few datastructures we need such as "Request" and "Response", but those datastructures are pretty low-level, e.g. headers are `list[tuple[bytes, bytes]]`.

We decided to continue using WSGI, so we can keep the high level Request and Response of Werkzeug and need to change neither the Odoo HTTP stack: http/ module, ir.http model are controllers are mostly untouched. For WSGI we had to implement the server-side portion of the protocol (namely `environ`, `start_response` and file-like body).

[benchmark-werkzeug-h11-2026-03.zip](https://github.com/user-attachments/files/26309861/benchmark-werkzeug-h11-2026-03.zip)

Benchmark showing that h11 is better than werkzeug across configurations

Task-6050990

### DIFF
--- a/odoo/http/router.py
+++ b/odoo/http/router.py
@@ -37,7 +37,7 @@ from odoo.modules.module import (
     initialize_sys_path,
 )
 from odoo.modules.registry import Registry
-from odoo.tools import config, file_open, file_path, profiler, real_time
+from odoo.tools import config, file_open, file_path, profiler
 from odoo.tools.misc import submap
 
 if typing.TYPE_CHECKING:
@@ -233,20 +233,9 @@ class Application:
             server that this application must call in order to send the
             HTTP response status line and the response headers.
         """
-        current_thread = threading.current_thread()
-        current_thread.query_count = 0
-        current_thread.query_time = 0
-        current_thread.perf_t0 = real_time()
-        current_thread.cursor_mode = None
-        if hasattr(current_thread, 'dbname'):
-            del current_thread.dbname
-        if hasattr(current_thread, 'uid'):
-            del current_thread.uid
-        if hasattr(current_thread, 'sess_id'):
-            del current_thread.sess_id
-        current_thread.rpc_model_method = ''
-
         if config['proxy_mode'] and environ.get("HTTP_X_FORWARDED_HOST"):
+            # It is done in our wsgi server already, but people might be
+            # using another wsgi server.
             # The ProxyFix middleware has a side effect of updating the
             # environ, see https://github.com/pallets/werkzeug/pull/2184
             def fake_app(environ, start_response):
@@ -261,7 +250,7 @@ class Application:
 
             try:
                 _set_session_and_dbname(request)
-                current_thread.url = httprequest.url
+                threading.current_thread().url = httprequest.url
 
                 if self.get_static_file(httprequest.path):
                     response = serve_static(request)
--- a/odoo/http/server.py
+++ b/odoo/http/server.py
@@ -1,0 +1,406 @@
+import io
+import logging
+import sys
+import time
+import typing
+from contextlib import suppress
+from http import HTTPStatus
+from urllib.parse import unquote
+from wsgiref.handlers import format_date_time
+from wsgiref.simple_server import sys_version
+
+import h11
+from werkzeug.exceptions import RequestTimeout
+from werkzeug.exceptions import default_exceptions as werkzeug_exceptions
+from werkzeug.middleware.proxy_fix import ProxyFix
+
+import odoo
+import odoo.release
+from odoo.tools import config
+
+from .router import root
+from .server_log import _logger, http_log, reset_thread_info
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Buffer
+
+# Buffer size for recv() calls.
+# 16kiB to match nginx's default client_body_buffer_size.
+RECV_SIZE = 16384
+
+# h11 bufferizes what has been recv()-ed inside its connection object.
+# It has not buffer size limit: receive_data() never raise an exception.
+# Instead, when calling next_event(), it does the following:
+#
+#   event = _extract_next_receive_event()
+#   if event is NEED_DATA:
+#       if len(_receive_buffer) > MAX_INCOMPLETE_EVENT_SIZE:
+#           raise RemoteProtocolError(...)
+#   return event
+#
+# As we always do the following pattern:
+#
+#   while conn.next_event() is h11.NEED_DATA:
+#       conn.receive_data(sock.recv(RECV_SIZE))
+#
+# The buffer can grow bigger than MAX_INCOMPLETE_EVENT_SIZE, but in our
+# case it never grows bigger than MAX_INCOMPLETE_EVENT_SIZE + RECV_SIZE.
+MAX_INCOMPLETE_EVENT_SIZE = 16384
+
+# No need to conceal the server agent as Odoo MUST run behind a web
+# server (apache/nginx/...), which is gonna replace it with its own.
+SERVER_SOFTWARE = ' '.join((
+    f'odoo/{odoo.release.series}',
+    h11.PRODUCT_ID,
+    sys_version,
+))
+SERVER_AGENT = SERVER_SOFTWARE.encode()
+
+
+def _verify_target(target: bytes):
+    # RFC9112 says in #3.2:
+    # 1. Clients MUST use origin-form when talking to origin servers.
+    # 2. Proxies MUST convert absolute-form to origin-from + Host.
+    # 3. Servers MUST convert absolute-form to origin-from + Host
+    # We ignore this third point as Odoo MUST be deployed behind a
+    # proxy and both the clients and proxies MUST NOT send
+    # absolute-form request targets. If we receive an absolute-form
+    # it likely is a rogue client.
+    if not target.startswith(b'/') and target != b'*':
+        e = "target is not origin-form"
+        raise h11.RemoteProtocolError(e)
+
+
+def _verify_headers(headers):
+    # RFC9112 says in #field.transfer-encoding:
+    # Transfert-Encoding and Content-Length can be both present in the
+    # headers, in which case Transfert-Encoding takes precedent over
+    # Content-Length, but it can make it vulnerable to request smuggling
+    # or splitting. "A server MAY reject a request that contains both
+    # Content-Length and Transfer-Encoding".
+    if (
+        any(h == b'transfert-encoding' for h, _ in headers) and
+        any(h == b'content-length' for h, _ in headers)
+    ):
+        e = "can't use both Content-Length and Transfert-Encoding"
+        raise h11.RemoteProtocolError(e)
+
+
+class HTTPSocket:
+    __slots__ = (
+        'addr',
+        'conn',
+        'h11request',
+        'h11response',
+        'ip',
+        'sock',
+    )
+
+    def __init__(self, client_sock, client_addr, *, prelude=b''):
+        self.sock = client_sock
+        self.addr = client_addr
+        self.ip = client_addr[0]  # until we find the X-Forwarded-For header
+
+        self.conn = h11.Connection(
+            h11.SERVER,
+            max_incomplete_event_size=MAX_INCOMPLETE_EVENT_SIZE,
+        )
+        if prelude:
+            self.conn.receive_data(prelude)
+
+        self.h11request = None
+        self.h11response = None
+
+    def _log(self, level, msg, *args, req=None, res=None, **kwargs):
+        kwargs.setdefault('extra', {}).setdefault('remote_addr', self.ip)
+        http_log(
+            level,
+            msg,
+            *args,
+            req=req or self.h11request,
+            res=res or self.h11response,
+            **kwargs,
+        )
+
+    def _make_environ(self):
+        # h11 made sure the target, http_version, header names and
+        # content-length only contain valid ascii characters. It didn't
+        # verify any other header value as HTTP (RFC9110) states they
+        # can be "opaque" data. Decoding them as latin-1 gives the
+        # correct string for US-ASCII and ISO-8859-1 which are the two
+        # commonly used charset for header values. Using latin-1 doesn't
+        # break RFC2047/RFC5987/RFC8187-encoding (base64 or %-encoding)
+        # and just leaves the string unparsed. HTTP headers that use
+        # other (non-standard) charsets are passed to the application as
+        # latin-1 string and must be re-encoded by the application. This
+        # is in line with the WSGI spec.
+        assert self.h11request
+        request_uri = self.h11request.target.decode('ascii')
+        path_quoted, _, query = request_uri.partition('?')
+
+        environ = {
+            'REQUEST_METHOD': self.h11request.method.decode('ascii'),
+            'SCRIPT_NAME': '',
+            'PATH_INFO': unquote(path_quoted, 'latin-1'),
+            'QUERY_STRING': query,
+            'REQUEST_URI': request_uri,
+            'RAW_URI': request_uri,
+            'REMOTE_ADDR': self.addr[0],
+            'REMOTE_PORT': self.addr[1],
+            'SERVER_NAME': config['http_interface'],
+            'SERVER_PORT': config['http_port'],
+            'SERVER_PROTOCOL': 'HTTP/' + self.h11request.http_version.decode('ascii'),
+            'SERVER_SOFTWARE': SERVER_SOFTWARE,
+            'wsgi.version': (1, 0),
+            'wsgi.url_scheme': 'http',
+            'wsgi.input': HTTPBodyReader(self.sock, self.conn),
+            'wsgi.errors': sys.stderr,
+            'wsgi.multithread': not config['workers'] and not odoo.evented,
+            'wsgi.multiprocess': config['workers'] and not odoo.evented,
+            'wsgi.run_once': False,
+        }
+        if environ['wsgi.multithread']:
+            # cannot use websocket in multiworker and gevent uses
+            # another http library than this one
+            environ['socket'] = self.sock
+            environ['odoo.trailing_data'] = lambda: self.conn.trailing_data
+
+        # do like gunicorn: ignore headers with _ and merge those with
+        # an identic name
+        headers = {}
+        for header, value in self.h11request.headers:
+            if b'_' in header:
+                continue
+            header = 'HTTP_' + header.upper().replace(b'-', b'_').decode('ascii')
+            if header in headers:
+                headers[header] += ',' + value.decode('latin-1')
+            else:
+                headers[header] = value.decode('latin-1')
+        if content_type := headers.pop('HTTP_CONTENT_TYPE', ''):
+            environ['CONTENT_TYPE'] = content_type
+        if content_length := headers.pop('HTTP_CONTENT_LENGTH', ''):
+            environ['CONTENT_LENGTH'] = content_length
+        environ.update(headers)
+
+        if config['proxy_mode'] and environ.get('HTTP_X_FORWARDED_HOST'):
+            pf = ProxyFix(lambda e, sr: (), x_for=1, x_proto=1, x_host=1)
+            pf(environ, lambda status, headers: None)  # it updates environ
+            self.ip = environ['REMOTE_ADDR']
+
+        return environ
+
+    def process_request(self):
+        reset_thread_info()
+
+        # Receive the HTTP request
+        try:
+            event = self.conn.next_event()
+            while event is h11.NEED_DATA:
+                try:
+                    self.conn.receive_data(self.sock.recv(RECV_SIZE))
+                except ConnectionError:
+                    return
+                except TimeoutError as exc:
+                    e = "Timed out"
+                    raise h11.RemoteProtocolError(e, HTTPStatus.REQUEST_TIMEOUT) from exc
+                event = self.conn.next_event()
+            if type(event) is h11.ConnectionClosed:
+                return
+            assert type(event) is h11.Request, event
+            self.h11request = event
+            _verify_target(event.target)
+            _verify_headers(event.headers)
+        except h11.RemoteProtocolError as exc:
+            # There is an error in the HTTP request. Fail with a 4xx and
+            # close the socket. Do not bother with a reason or a body.
+            self.h11response = h11.Response(
+                status_code=exc.error_status_hint,
+                headers=[
+                    (b'date', format_date_time(time.time()).encode()),
+                    (b'server', SERVER_AGENT),
+                    (b'connection', b'close'),
+                    (b'content-length', b'0'),
+                ],
+            )
+            with suppress(ConnectionError):
+                self.sock.sendall(self.conn.send(self.h11response))
+            self._log(logging.INFO, '', exc_info=_logger.isEnabledFor(logging.DEBUG))
+            eof = self.conn.send(h11.EndOfMessage())
+            assert eof == b'', eof
+            assert self.conn.our_state is h11.MUST_CLOSE, self.conn.our_state
+            return
+        else:
+            # We got our request.
+            self._log(logging.DEBUG, '[REQ] ')  # request line + headers
+
+        # If the request has header "Expect: 100-continue". Send a 100
+        # Continue response now: we're just before forwarding the
+        # request to the WSGI application, and we've read all headers.
+        if self.conn.client_is_waiting_for_100_continue:
+            res_continue = h11.InformationalResponse(
+                status_code=HTTPStatus.CONTINUE,
+                headers=[
+                    (b'date', format_date_time(time.time()).encode()),
+                    (b'server', SERVER_AGENT),
+                ],
+            )
+            self.sock.sendall(self.conn.send(res_continue))
+            self._log(logging.DEBUG, '[100] ', res=res_continue)
+
+        # Pass the request into the WSGI application. It'll call our
+        # start_response() with the http response line and headers, then
+        # return an iterable object containing the body.
+        wsgi_environ = self._make_environ()
+        wsgi_response = root(wsgi_environ, self.start_response)
+        assert self.h11response
+
+        if self.conn.our_state is h11.SWITCHED_PROTOCOL:
+            # The WSGI application called start_response() with
+            # > HTTP/1.1 101 Switching Protocols
+            # > Connection: upgrade
+            # > Upgrade: websocket
+            # HTTP-wise there is nothing left to do.
+            if hasattr(wsgi_response, 'close'):
+                wsgi_response.close()  # werkzeug call_on_close()
+            self._log(logging.INFO, '')
+            return
+
+        # Iter over the iterable object the WSGI application gave us. It
+        # contains the body that we are yet to send on the network.
+        bytes_sent = 0
+        exc_info = None
+        try:
+            for chunk in wsgi_response:
+                if data := self.conn.send(h11.Data(chunk)):
+                    self.sock.sendall(data)
+                bytes_sent += len(chunk)
+            if data := self.conn.send(h11.EndOfMessage()):
+                self.sock.sendall(data)
+        except ConnectionError:
+            # The client is no longer connected. Ignore the exception.
+            self.conn.send_failed()
+        except Exception as exc:  # noqa: BLE001
+            # A fatal error occured while sending the body, mark the
+            # connection failed and close it.
+            self.conn.send_failed()
+            exc_info = exc
+        finally:
+            if hasattr(wsgi_response, 'close'):
+                wsgi_response.close()  # werkzeug call_on_close()
+
+        self._log(  # response body
+            logging.ERROR if exc_info else logging.INFO,
+            '',
+            exc_info=exc_info,
+            extra={
+                'http_response_body': bytes_sent,
+                'http_headers': (),  # already printed at [RES]
+            },
+        )
+
+    def start_response(
+        self,
+        status: str,
+        response_headers: list[tuple[str, str]],
+        exc_info=None,
+    ):
+        status_code, _, reason = status.partition(' ')
+        status_code = HTTPStatus(int(status_code))
+
+        # Do the wsgi-encoding dance back, see _make_environ()
+        response_headers: list[tuple[bytes, bytes]] = [
+            (header.encode('ascii').lower(), value.encode('latin-1'))
+            for header, value in response_headers
+        ]
+        get_header = dict(response_headers).get
+
+        # Add the mandatory HTTP headers
+        if (cnx := get_header(b'connection')) is not None:
+            if cnx.lower() == b'upgrade':
+                if not get_header(b'upgrade'):
+                    e = "the Upgrade header is mandatory with Connection: upgrade"
+                    raise ValueError(e)
+            elif cnx.lower() != b'close':
+                e = f"invalid Connection header: {cnx}"
+                raise ValueError(e)
+        else:
+            response_headers.append((b'connection', b'close'))
+        if get_header(b'server') is None:
+            response_headers.append((b'server', SERVER_AGENT))
+        if get_header(b'date') is None:
+            date = format_date_time(time.time()).encode()
+            response_headers.insert(0, (b'date', date))
+
+        # Send and log the response
+        h11response_cls = (
+            h11.InformationalResponse
+            if status_code.is_informational else
+            h11.Response
+        )
+        self.h11response = h11response_cls(
+            status_code=status_code,
+            reason=reason,
+            headers=response_headers,
+        )
+        self.sock.sendall(self.conn.send(self.h11response))
+        self._log(logging.DEBUG, '[RES] ')  # response line + headers
+
+
+class HTTPBodyReader(io.RawIOBase):
+    def __init__(self, sock, conn):
+        self._sock = sock
+        self._conn = conn
+        self._buffered: Buffer | None = b''
+
+    def readable(self) -> bool:
+        return True
+
+    def read(self, size: int = -1):
+        if size == -1:
+            buff = memoryview(bytearray(RECV_SIZE))
+            result = bytearray()
+            while count := self.readinto(buff):
+                result.extend(buff[:count])
+            return bytes(result)
+
+        if self._buffered:
+            result, self._buffered = self._buffered[:size], self._buffered[size:]
+            return bytes(result)
+
+        buff = memoryview(bytearray(size))
+        count = self.readinto(buff)
+        return bytes(buff[:count])
+
+    def readinto(self, buff):
+        if self._buffered is None:  # eof
+            return 0
+
+        if self._buffered:
+            size = min(len(buff), len(self._buffered))
+            buff[:size], self._buffered = self._buffered[:size], self._buffered[size:]
+            return size
+
+        try:
+            event = self._conn.next_event()
+            while event is h11.NEED_DATA:
+                try:
+                    self._conn.receive_data(self._sock.recv(max(len(buff), RECV_SIZE)))
+                except ConnectionError as exc:
+                    exc.loglevel = logging.WARNING
+                    raise
+                except TimeoutError as exc:
+                    raise RequestTimeout from exc
+                event = self._conn.next_event()
+        except h11.RemoteProtocolError as exc:
+            raise werkzeug_exceptions[exc.error_status_hint] from exc
+
+        if type(event) is h11.EndOfMessage:
+            self._buffered = None  # eof
+            return 0
+
+        assert type(event) is h11.Data, event
+        size = min(len(buff), len(event.data))
+        data = memoryview(event.data)
+        buff[:size] = data[:size]
+        self._buffered = data[size:]
+        return size
--- a/odoo/http/server_log.py
+++ b/odoo/http/server_log.py
@@ -1,0 +1,254 @@
+import functools
+import logging
+import pprint
+import threading
+import time
+import typing
+from http import HTTPStatus
+from wsgiref.handlers import format_date_time
+
+import h11
+
+from odoo.netsvc import (
+    BOLD_SEQ,
+    COLOR_PATTERN,
+    CYAN,
+    DEFAULT,
+    GREEN,
+    MAGENTA,
+    PID_COLORS,
+    RED,
+    RESET_SEQ,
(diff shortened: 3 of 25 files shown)


odoo/http/router.py
```diff
@@ -242,10 +242,8 @@ def __call__(self, environ: WSGIEnvironment, start_response: StartResponse) -> I
         current_thread.perf_t0 = real_time()
         current_thread.cursor_mode = None
         if hasattr(current_thread, 'dbname'):
-            del current_thread.dbname
```
kmagusiak, line 242: In wrong commit?

odoo/http/server_log.py
```diff
@@ -0,0 +1,106 @@
+import functools
+import logging
+import typing
+from http import HTTPStatus
```
kmagusiak, line 7: Why do you use http module here instead of writing the codes like we did in exceptions.py?

odoo/http/log.py
```diff
@@ -0,0 +1,106 @@
...
+@functools.lru_cache(1)
+def _has_color():
+    return any(
+        isinstance(handler.formatter, ColoredFormatter)
+        for handler
+        in logging.root.handlers
+    )
+
+
+def http_log(logger, level, msg, *args, **kwargs):
```
kmagusiak: Isn't it the role of the formatter?
You call `http_log(_logger, logging.INFO, "toto", extra={})`, but you could just configure the logger and call `_logger.info("toto", extra={})`.

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+                http_log(_logger, logging.INFO, '', extra={
+                    'remote_addr': self.ip,
+                    'http_request_line': '"- - HTTP/?"',
+                    'http_response_status': exc.status_code_hint,
+                    'http_response_body': 0,
+                    'query_count': 0,
+                    'query_time': 0,
+                    'remaining_time': real_time() - t0 - current_thread.query_time,
+                }, exc_info=_logger.isEnabledFor(logging.DEBUG))
+                assert not self.conn.send(h11.EndOfMessage())
```
kmagusiak: Action should not be in an assert.

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+            h11request.target,
+            h11request.http_version,
+        )).decode()
+
+        if self.conn.client_is_waiting_for_100_continue:
+            # The request has header "Expect: 100-continue". Send a
+            # 100 Continue response now as we are just before forwarding
+            # the request to the WSGI application.
+            # Maybe we should instead let the WSGI application send this
+            # 100-Continue info response, and we would only does it
```
kmagusiak:
```suggestion
            # 100-Continue info response, and we would only do it
```

If we do it, it makes sense to send it when starting to read the body. However, to me it is an implementation detail of the server and should not be handled by WSGI layer.

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+        # Iter over the iterable object the WSGI application gave us. It
+        # contains the body that we are yet to send on the network.
+        bytes_sent = 0
+        response_iter = iter(wsgi_response)
+        try:
+            while True:
+                try:
+                    chunk = next(response_iter)
+                except StopIteration:
+                    # There are no more data to send. Close the response
```
kmagusiak: Also below
```suggestion
                    # There is no more data to send. Close the response
```

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+            _logger.debug("%s\n%s", status, pprint.pformat(response_headers))
+        status_code, _, reason = status.partition(' ')
+        status_code = HTTPStatus(int(status_code))
+
+        # Do the wsgi-encoding dance back, see _make_environ()
+        response_headers: list[tuple[bytes, bytes]] = [
+            (header.encode('ascii').lower(), value.encode('latin-1'))
+            for header, value in response_headers
+        ]
+        find_header_ = partial(find_header, response_headers)
```
kmagusiak: Just make a dict

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+        self.sock.sendall(self.conn.send(h11_response))
+
+        current_thread = threading.current_thread()
+        http_log(_logger, logging.INFO, '', extra={
+            'remote_addr': self.ip,
+            'http_request_line': self._request_line,
+            'http_response_status': status_code,
+            'http_response_body': (
+                    cl if (cl := find_header_(b'content-length')) is not None
+                else 0 if status_code in _NO_BODY_STATUS
```
kmagusiak: You should check for "transfer-encoding: chunked" for "stream"; otherwise, it should be 0.

Nitpick: strange indentation

odoo/http/server.py
```diff
@@ -0,0 +1,406 @@
...
+            return b''
+
+        assert type(event) is h11.Data, event
+        self._buffered[:] = event.data[size if size != -1 else None:]
+        return event.data[:size if size != -1 else None]
+
+    def readinto(self, buff):
+        # /!\ Keep me sync with read() /!\
+
+        if self._eof_received:
```
kmagusiak: I will suggest another implementation for the reader. I also think some edge cases are not correctly implemented here (ex: if you have a buffer and eof, you should still read the buffer; `read(-1)` means read until EOF).

kmagusiak: I pushed another implementation

odoo/http/log.py
Julien00859: can't do, cause formatter is linked to an handler, not a logger, so I cannot have a formatter only for the http server without copying all the handlers from the root logger.

odoo/http/server.py
Julien00859: Indeed, `there is no` for uncountable
https://grammarhow.com/there-is-no-or-there-are-no/

odoo/http/server_log.py
Julien00859: Cause in exceptions.py there were no import, so I wanted to leave it that way. In this file I'm already importing a lot of things, and it only is impor

odoo/service/server.py
kmagusiak:
```suggestion
                    self.port = config['http_port'] = server.getsockname()[1]
```

kmagusiak: Shouldn't this do `client.close()`?

requirements.txt
```diff
@@ -14,6 +14,7 @@ gevent==24.2.1 ; sys_platform != 'win32' and python_version >= '3.12' and python
 gevent==24.11.1 ; sys_platform != 'win32' and python_version >= '3.13'  # (Trixie)
 greenlet==3.0.3 ; sys_platform != 'win32' and python_version >= '3.12' and python_version < '3.13' # (Noble)
 greenlet==3.1.1 ; sys_platform != 'win32' and python_version >= '3.13'  # (Trixie)
+h11==0.14.0
```
KangOl: Include the fix for [CVE-2025-43859](https://github.com/python-hyper/h11/security/advisories/GHSA-vqfr-h8mv-ghfj)[^1].
```suggestion
h11==0.16.0
```

[^1]: It has been backported in the ubuntu package: https://changelogs.ubuntu.com/changelogs/pool/universe/p/python-h11/python-h11_0.14.0-1ubuntu0.25.04.1/changelog

kmagusiak: @odoo/rd-runbot As far as I know, we pin the same version as in the supported distribution.
We could pin `h11>=0.14.0,h11<=0.16.0` so that pip keeps the system version, but if it installs, it uses the latest version. Maybe overkill; what do you think?

odoo/http/server_log.py
```diff
@@ -0,0 +1,114 @@
...
+from .requestlib import DEFAULT_MAX_CONTENT_LENGTH, MAX_FORM_SIZE
+
+__all__ = ('http_log',)
+
+_HTTP_EXTRA = {
+    'remote_addr': '-',
+    'http_request_line': '"- - -"',
+    'http_response_status': '-',
+    'http_response_body': '-',
+    'query_count': 0,
```
kmagusiak, line 48: Since you moved the implementation here, shouldn't you remove `PerfFilter` from netsvc?

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+        # commonly used charset for header values. Using latin-1 doesn't
+        # break RFC2047/RFC5987/RFC8187-encoding (base64 or %-encoding)
+        # and just leaves the string unparsed. HTTP headers that use
+        # other (non-standard) charsets are passed to the application as
+        # latin-1 string and must be re-encoded by the application. This
+        # is in line with the WSGI spec.
+        path_info = unquote(url.path, 'latin-1')
+        request_uri = f'{path_info}?{url.query}' if url.query else path_info
+        environ = {
+            'REQUEST_METHOD': h11request.method.decode('ascii'),
```
kmagusiak: Could you split the standard WSGI variables and the non-standard ones?
https://peps.python.org/pep-3333/#environ-variables

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+        if config['proxy_mode'] and environ.get('HTTP_X_FORWARDED_HOST'):
+            pf = ProxyFix(lambda environ, start_response: (), x_for=1, x_proto=1, x_host=1)
+            pf(environ, lambda status, headers: None)  # it updates environ
+            self.ip = environ['REMOTE_ADDR']
+
+        return environ
+
+    def process_request(self):
+        t0 = real_time()
+        current_thread = threading.current_thread()
```
kmagusiak: Could you move this initialization block to `server_log.py` since it is used for logging?
`def http_log_start_process()`

odoo/service/server.py
```diff
@@ -867,21 +711,23 @@ def get_environ(self):
...
-
-        # socket initiation copied from gevent's WSGIServer with added SO_REUSEPORT if possible
-        sock = socket.socket(family, socket.SOCK_STREAM)
-        if sys.platform != 'win32':
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-            if hasattr(socket, 'SO_REUSEPORT'):
-                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
-        sock.bind((self.interface, self.port))
-        sock.listen(128)
+        sock = socket.create_server(
```
kmagusiak, line 713: Can you refactor: make a function `create_server(...) -> (socket, host, port)` and use it in both threaded and multi-process servers?

odoo/service/server.py
```diff
@@ -1461,28 +1309,21 @@ def process_request(self, client, addr):
...
-            self.process_request(client, addr)
-        except OSError as e:
-            if e.errno not in (errno.EAGAIN, errno.ECONNABORTED):
-                raise
-
-    def start(self):
-        Worker.start(self)
-        self.server = BaseWSGIServerNoBind(self.multi.app)
+        except BlockingIOError:
+            pass
```
kmagusiak: Isn't this a programming error? (because the socket should be non-blocking)

odoo/http/server_log.py
```diff
@@ -0,0 +1,114 @@
...
+        for handler
+        in logging.root.handlers
+    )
+
+
+def http_log(level, msg, *args, **kwargs):
+    if not _logger.isEnabledFor(level):
+        return
+
+    extra = kwargs['extra'] = _HTTP_EXTRA | kwargs.get('extra', {})
```
kmagusiak: You could compute the performance counters here instead of always passing them in attributes.
You can have code that looks like:
```py
current_thread = ...
extra = kwargs.get('extra', {})
extra.setdefault('query_count', current_thread.query_count)
# ... etc
extra = kwargs['extra'] = _HTTP_EXTRA | extra
```

It will remove repeated lined from the server.

Another possibility would be to have
`def http_log(level, msg: str, *args, query_count=0, ..., **kwargs)`

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+    def __init__(self, client_sock, client_addr, *, prelude=b''):
+        self.sock = client_sock
+        self.addr = client_addr
+        self.ip = client_addr[0]
+
+        self.conn = h11.Connection(h11.SERVER, max_incomplete_event_size=MAX_BUFFER_SIZE)
+        if prelude:
+            self.conn.receive_data(prelude)
+
+        self._request_line = None
```
kmagusiak: No reason to have only one attribute with '_'
```suggestion
        self.request_line = None
```

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+                except StopIteration:
+                    # There is no more data to send. Close the response
+                    # and http stream.
+                    if hasattr(wsgi_response, 'close'):
+                        wsgi_response.close()  # werkzeug call_on_close()
+                    if data := self.conn.send(h11.EndOfMessage()):
+                        try:
+                            self.sock.sendall(data)
+                        except ConnectionError:
+                            break
```
kmagusiak: Can you add a comment that we ignore the error because the client may already be gone?

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+        response_iter = iter(wsgi_response)
+        exc_info = None
+        try:
+            while True:
+                try:
+                    chunk = next(response_iter)
+                except StopIteration:
+                    # There is no more data to send. Close the response
+                    # and http stream.
+                    if hasattr(wsgi_response, 'close'):
```
kmagusiak: `break` and move this code outside of the loop.

odoo/http/server.py
```diff
@@ -0,0 +1,425 @@
...
+            for header, value in response_headers
+        ]
+        get_header = dict(response_headers).get
+
+        # https://github.com/pallets/werkzeug/issues/3127
+        if get_header(b'www-authenticate'):
+            for index, (header, value) in enumerate(response_headers):
+                if header == b'www-authenticate':
+                    break
+            response_headers[index] = (header, value.rstrip())
```
kmagusiak: We produce the response, can't we strip it when creating the response header (I mean in business code)?
According to the [RFC](https://httpwg.org/specs/rfc9110.html#fields.values) "A field value does not include leading or trailing whitespace." - so it's not only for that specific header.

odoo/http/router.py
```diff
@@ -243,17 +243,6 @@ def __call__(self, environ: WSGIEnvironment, start_response: StartResponse) -> I
...
-        current_thread.query_time = 0
-        current_thread.perf_t0 = real_time()
-        current_thread.cursor_mode = None
-        if hasattr(current_thread, 'dbname'):
-            del current_thread.dbname
-        if hasattr(current_thread, 'uid'):
-            del current_thread.uid
-        current_thread.rpc_model_method = ''
-
         if config['proxy_mode'] and environ.get("HTTP_X_FORWARDED_HOST"):
```
kmagusiak, line 236: Can you add a comment that this only for wsgi deployments?
Because the server already handles it.

odoo/service/server.py
```diff
@@ -1461,28 +1309,21 @@ def process_request(self, client, addr):
...
-        self.server.socket = client
-        # tolerate broken pipe when the http client closes the socket before
-        # receiving the full reply
-        try:
-            self.server.process_request(client, addr)
-        except OSError as e:
-            if e.errno != errno.EPIPE:
-                raise
+        http_socket = HTTPSocket(client, addr)
+        http_socket.process_request()
```
kmagusiak, line 1316: Run in a new context, so that it works well with contextvars:
```suggestion
        # if you merge process_request at the end of the __init__, you could do:
        contextvars.Context().run(HTTPSocket, client, addr, process=True)
```

Same for the threaded server.

odoo/http/server.py
Julien00859: There's no easy way to do that, the bug is in werkzeug.

odoo/service/server.py
Julien00859: We'll do that when we introduce contextvars.

odoo/http/server.py
kmagusiak: Let's monkeypatch `WWWAuthenticate.to_header` so that it rstrips the result.

odoo/service/server.py
Julien00859: All servers (including gevent) are *just different enough* to prevent further refactors

kmagusiak: Threaded and multi-process servers should create the server socket in exactly the same way.

odoo/http/server.py
Julien00859: That's what I meant by "there is no easy way to do that" :D

odoo/service/server.py
Julien00859: Agreed, but outside the scope of this branch

kmagusiak: Why? You just wrote that code, so you can factor the same code that is in both servers.

odoo/http/server.py
kmagusiak: I think a monkey patch is cleaner because it fixes the issue at the root.

odoo/service/server.py
Julien00859: We agreed IRL to work on this later

requirements.txt
Julien00859: Decided with runbot: let's put 16 in the requirements.txt with a comment explaining it's for a CVE

Julien00859: There are no API change between 0.14.0 and 0.16.0. The only two commits that are not style/typing noise are:

* python-hyper/h11@60782ad107e538b9312aac7e1c119c8358bf797c "Reject Content-Length longer 1 billion TB", a mitigation for [CVE-2020-10735](https://security-tracker.debian.org/tracker/CVE-2020-10735). Fixed in Python 3.12 (the minimal python version in master) so we don't need it.
* python-hyper/h11@dff7cc397a26ed4acdedd92d1bda6c8f18a6ed9f "Validate Chunked-Encoding chunk footer" the patch for [CVE-2025-43859](https://security-tracker.debian.org/tracker/CVE-2025-43859), we need it.

kmagusiak (approved):
Great work!

odoo/http/server.py
```diff
@@ -0,0 +1,383 @@
...
+                h11request = self.conn.next_event()
+            if type(h11request) is h11.ConnectionClosed:
+                return
+            assert type(h11request) is h11.Request, h11request
+            try:
+                url = parse_url(h11request.target.decode('ascii'))
+                if url.scheme or url.auth or url.host or url.port:
+                    raise ValueError  # noqa: TRY301
+            except ValueError as exc:
+                e = "bad target: %s" % h11.request.target
```
antonrom1:
```suggestion
                e = "bad target: %s" % h11request.target
```

odoo/service/server.py
```diff
@@ -629,10 +411,74 @@ def cron_spawn(self):
...
+            current_thread = threading.current_thread()
+            _logger.critical("Thread %s (%s) Exception occurred, quitting...",
+                current_thread.name, current_thread.ident, exc_info=True)
+        finally:
+            client.close()
+
+    def http_server_thread(self, stop_event):
+        try:
+            piscine = ThreadPoolExecutor(
+                max_workers=config.max_http_threads or None,  # None: no limit
```
antonrom1: None does not mean no limit, it just uses `min(32, (os.process_cpu_count() or 1) + 4)`

odoo/http/server.py
```diff
@@ -0,0 +1,383 @@
...
+        # contains the body that we are yet to send on the network.
+        bytes_sent = 0
+        exc_info = None
+        try:
+            for chunk in wsgi_response:
+                if data := self.conn.send(h11.Data(chunk)):
+                    self.sock.sendall(data)
+                bytes_sent += len(chunk)
+            if hasattr(wsgi_response, 'close'):
+                wsgi_response.close()  # werkzeug call_on_close()
```
antonrom1: Shouldn't that be in finally block?
 
https://peps.python.org/pep-0333/:

> If the iterable returned by the application has a close() method, the server or gateway must call that method upon completion of the current request, whether the request was completed normally, or terminated early due to an application error during iteration or an early disconnect of the browser. (The close() method requirement is to support resource release by the application. This protocol is intended to complement [PEP 342](https://peps.python.org/pep-0342/)’s generator support, and other common iterables with close() methods.)

odoo/http/server.py
```diff
@@ -0,0 +1,383 @@
...
+                try:
+                    self._conn.receive_data(self._sock.recv(max(len(buff), RECV_SIZE)))
+                except ConnectionError as exc:
+                    exc.loglevel = logging.WARNING
+                    raise
+                except TimeoutError as exc:
+                    raise RequestTimeout from exc
+                event = self._conn.next_event()
+        except h11.RemoteProtocolError as exc:
+            raise werkzeug_exceptions[exc.status_code_hint] from exc
```
antonrom1: it's not error_status_hint?
```py
try: raise h11.RemoteProtocolError("teapot kaboomed", 418)
except h11.RemoteProtocolError as exc:
    assert not hasattr(exc, 'status_code_hint')
    print(exc.error_status_hint)
```

odoo/http/server.py
```diff
@@ -0,0 +1,383 @@
...
+                    e = "Timed out"
+                    raise h11.RemoteProtocolError(e, HTTPStatus.REQUEST_TIMEOUT) from exc
+                h11request = self.conn.next_event()
+            if type(h11request) is h11.ConnectionClosed:
+                return
+            assert type(h11request) is h11.Request, h11request
+            try:
+                url = parse_url(h11request.target.decode('ascii'))
+                if url.scheme or url.auth or url.host or url.port:
+                    raise ValueError  # noqa: TRY301
```
antonrom1: why not accept absolute form urls?
```
GET http://example.com/odoo HTTP/1.1
Host: example.com
```
this is valid, isn't it?

odoo/http/server.py
```diff
@@ -0,0 +1,383 @@
...
+            'wsgi.run_once': False,
+        }
+        if environ['wsgi.multithread']:
+            # cannot use websocket in multiworker and gevent uses
+            # another http library than this one
+            environ['socket'] = self.sock
+            environ['odoo.trailing_data'] = lambda: self.conn.trailing_data
+
+        environ.update({
+            'HTTP_' + header.upper().replace(b'-', b'_').decode('ascii'): value.decode('latin-1')
```
antonrom1: it looks a bit sketchy to collapse _ into -, if the reverse proxy writes a trusted header, an attacker might be able to overwrite it?

X-Real-IP: 8.8.8.8
X_Real-IP: 1.1.1.1



looked into it, gunicorn has an option for it https://gunicorn.org/reference/settings/#header_map

odoo/http/server.py
kmagusiak: Indeed :(
I knew it was for proxy servers, but "A server MUST accept the absolute-form in requests even though most HTTP/1.1 clients will only send the absolute-form to a proxy."
https://datatracker.ietf.org/doc/html/rfc9112#name-absolute-form

Julien00859: I've added an length comment to explain why I don't support it, even if the RFC says "MUST"

odoo/http/server_log.py
```diff
@@ -0,0 +1,179 @@
...
+    'http_response_body': '-',
+    'query_count': 0,
+    'query_time': 0.0,
+    'remaining_time': 0.0,  # total time - query time
+    'cursor_mode': '-',
+    'http_headers': (),
+})
+
+
+def http_log(level, msg, *args, extra=_HTTP_EXTRA, **kwargs):
```
kmagusiak:
```suggestion
def http_log(level, msg, *args, extra=frozendict(), **kwargs):
```

odoo/http/server.py
```diff
@@ -0,0 +1,462 @@
...
+        current_thread = threading.current_thread()
+        status_code, _, reason = status.partition(' ')
+        status_code = HTTPStatus(int(status_code))
+
+        # Some controllers (xmlrpc/jsonrpc/call-kw) let users call any
+        # model-method. Add the model and method as fake fragment.
+        if current_thread.rpc_model_method:
+            self.h11request = h11.Request(
+                method=self.h11request.method,
+                target=self.h11request.target + b'#' + current_thread.rpc_model_method.encode(),
```
kmagusiak: Can you move this to `_log` method and avoid using `current_thread` here?

odoo/http/server_log.py
```diff
@@ -0,0 +1,179 @@
...
+    ip address, request line, response status code and response body
+    length) and the perf meters (query count, query time, remaining
+    time, cursor mode).
+    """
+    if not _logger.isEnabledFor(level):
+        return
+
+    extra = _HTTP_EXTRA | extra
+
+    extra_log = extra.copy()  # keep "extra" color free
```
kmagusiak: The function could be made private and we ensure that we have all keys defined in `_log` method of the connection. This makes some copies here not required as the caller can ensure all is filled.

Julien00859:
@kmagusiak some more thoughts on `http_log`...

odoo/http/server_log.py
```diff
@@ -43,44 +52,114 @@
...
+):
     if not _logger.isEnabledFor(level):
         return
 
-    extra = _HTTP_EXTRA | extra
+    now = time.time()
+    real_now = real_time()
+
+    ct = threading.current_thread()
+    thread_extra = {
```
kmagusiak: Nitpick : just use one dict for all extra.
Removes chain map and other dicts that are just overhead here.

odoo/http/server_log.py
```diff
@@ -43,44 +52,114 @@
...
+        'query_count': ct.query_count,
+        'query_time': ct.query_time,
+        'remaining_time': real_now - ct.perf_t0 - ct.query_time,
+        'date': format_date_time(now),
```
kmagusiak: I would leave this empty here and always just use the value from res.

odoo/http/server.py
```diff
@@ -0,0 +1,398 @@
...
+
+        if config['proxy_mode'] and environ.get('HTTP_X_FORWARDED_HOST'):
+            pf = ProxyFix(lambda e, sr: (), x_for=1, x_proto=1, x_host=1)
+            pf(environ, lambda status, headers: None)  # it updates environ
+            self.ip = environ['REMOTE_ADDR']
+
+        return environ
+
+    def process_request(self):
+        t0 = real_time()
```
kmagusiak: See comment below (move this code).

Julien00859:
Hello @odoo/rd-security, we are ready. The various `type(event)` are only ever on data we manage, and it only is for a bunch of sanity assertions, I find that they help read the code but I can get rid of them if you prefer. The two `os.getenv` and `os.environ` were present in service/server.py, me moved them to config.py. Usage of `sys.stderr` is required by the WSGI spec.

odoo/http/server.py
antonrom1:
```diff
        # content-length only contain valid ascii characters. It didn't
-       # verified any other header value as HTTP (RFC9110) states they
+       # verify any other header value as HTTP (RFC9110) states they
```

setup.py
antonrom1: nitpick
```diff
-        "h11",
+        'h11',
```

odoo/http/server.py
xmo-odoo: Invoking the urllib3 parser doesn't seem necessary when at the end of the day what we want to check is whether the `target` is `*` or starts with a `/` and only has whitelisted bytes and valid hex escapes:

    origin-form    = absolute-path [ "?" query ]
    absolute-path = 1*( "/" segment )
    segment       = *pchar
    pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
    unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
    pct-encoded = "%" HEXDIG HEXDIG
    sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

In fact urllib3's parse seems straight up incorrect for the purpose of validating the target:

    >>> parse_url(b"/\x00".decode('ascii'))
    Url(scheme=None, auth=None, host=None, port=None, path='/%00', query=None, fragment=None)

C0 control codes should be rejected.

xmo-odoo: AFAIK the entire reason REQUEST_URI and RAW_URI exist is that WSGI specifies a lossy behaviour for PATH_INFO, giving them the same lossy behaviour is useless: etianen/aiohttp-wsgi#17 https://github.com/pallets/werkzeug/blob/main/docs/wsgi.rst#raw-request-uri-and-path-encoding

xmo-odoo: The use of urllib3's parse seems unnecessary since we've already validated we're getting an origin-form.

xmo-odoo: This doesn't seem to handle the case where the headers contain duplicate keys.

xmo-odoo:
```suggestion
            # The WSGI application called start_response() with
```

xmo-odoo: Does this need to be a buffer? As far as I can tell nothing ever gets written into that object, it just gets assigned to and split (sliced), so we create memoryviews for nothing.

odoo/service/server.py
xmo-odoo: ?

xmo-odoo: ...

odoo/tools/config.py
xmo-odoo, line 1073: Why are you forcing this on people who never asked for it?

odoo/http/server.py
Julien00859: C0 controls are rejected by h11 already
https://github.com/python-hyper/h11/blob/62c5068c971579d61fa1b55373390e12f25fd856/h11/_abnf.py#L85

Julien00859: We tried to copy as few data as possible and decided to make extensive use of bytearray and memoryview to achieve that. Our position is that allocating a few memoryviews too much is better than copying `RECV_SIZE` bytes by accident. When we'll be compatible with py3.15 we will be able to use [`bytearray.take_bytes`](https://docs.python.org/3.15/library/stdtypes.html#bytearray.take_bytes) where possible. h11's lack of a documented way to do `sock.recvinto(conn.buffer)` is annoying me too.

So yes it'll stay typed as a `Buffer`, but I agree that this `memoryview(b'')` is useless, I'm changing it to just `b''`. Will see for the other ones with @kmagusiak (we co-authored that code). Do you have suggestions?

odoo/tools/config.py
Julien00859: python's http server happily spawn a new thread per socket connection, which IMO is a terrible thing to do, both on a perf and security POV. We're instead using a thread-pool. For the thread-pool we need a default number of threads to spawn: `2 * cpu + 1` is the recommended value for pre-fork servers, I didn't had much thoughts about GIL and websockets and I decided to use the same. Maybe you have a better default value in mind?

By the way, that `ODOO_MAX_HTTP_THREADS` pre-dates me, it is used by OdooSH. We now validate this environment variable *upon server startup*, this is better than the old code that silently ignored `ValueError` and used a bizarre fallback computation.

I rather have that configuration thing next to all the other configuration things: in config.py, than lost inside service/server.py.

Code removed:
```py
class ThreadedWSGIServerReloadable(LoggingBaseWSGIServerMixIn, werkzeug.serving.ThreadedWSGIServer):
    """ werkzeug Threaded WSGI Server patched to allow reusing a listen socket
    given by the environment, this is used by autoreload to keep the listen
    socket open when a reload happens.
    """
    def __init__(self, host, port, app):
        # The ODOO_MAX_HTTP_THREADS environment variable allows to limit the amount of concurrent
        # socket connections accepted by a threaded server, implicitly limiting the amount of
        # concurrent threads running for http requests handling.
        self.max_http_threads = os.environ.get("ODOO_MAX_HTTP_THREADS")
        if self.max_http_threads:
            try:
                self.max_http_threads = int(self.max_http_threads)
            except ValueError:
                # If the value can't be parsed to an integer then it's computed in an automated way to
                # half the size of db_maxconn because while most requests won't borrow cursors concurrently
                # there are some exceptions where some controllers might allocate two or more cursors.
                self.max_http_threads = max((config['db_maxconn'] - config['max_cron_threads']) // 2, 1)
            self.http_threads_sem = threading.Semaphore(self.max_http_threads)
```

odoo/service/server.py
Julien00859: Useless indeed. Think I tried to save a bit of memory, but since we're not using the prelude atm it is useless (that prelude thing is a request from OdooSH, another PR is gonna build on top of that).

odoo/http/server.py
Julien00859: Simplified to `not target.startswith(b'/') or target != b'*'`. Thanks!

Julien00859: Did that out of simplicity to only import a single `parse_url` utility. Changed for `urllib.parse.urlsplit`. Would doing `target.partition(b'?')` be safe/ok?

Julien00859: Did the job myself, went and read both h11 and urlsplit (should had done it in the first place, sorry), since we guarantee it is origin-form it is safe to `target.partition(b'?')`.

Julien00859: Thought the web server would do something about those, like merging the values, or ignoring the latter/former ones. It actually forward the headers as-is. gunicorn is doing that:

```http
X-Foo: 1
X-Foo: 2
```
```py
{"HTTP_X_FOO": "1,2"}
```
decided to do the same.

Sorry.

Julien00859: Indeed my bad. REQUEST_URI and RAW_URI now have the %-encoded (raw) value.

odoo/tools/config.py
xmo-odoo: > python's http server happily spawn a new thread per socket connection, which IMO is a terrible thing to do, both on a perf and security POV. 

Neither really being relevant when using a threaded server for development.

> We're instead using a thread-pool. For the thread-pool we need a default number of threads to spawn

The parameter is optional so you don't.

Although the default is apparently only `cpu_count + 4`, so on most CPUs this should be higher. Eh.

> By the way, that ODOO_MAX_HTTP_THREADS pre-dates me

Yes, but the old version would just not do anything, this one sets a default.

odoo/http/server.py
kmagusiak: You can just assign `b''` here.

Using memory views is much faster here as it avoids moving bytes around when slicing.
`take_bytes` of 3.15 is not needed.

odoo/tools/config.py
```diff
@@ -1061,6 +1068,14 @@ def http_socket_activation(self):
             and os.getenv('LISTEN_PID') == str(os.getpid())
         )
 
+    @property
+    def max_http_threads(self):
+        mht = os.getenv('ODOO_MAX_HTTP_THREADS', str(2 * os.cpu_count() + 1))
+        if not mht.isdigit():
```
xmo-odoo: This is the wrong check:

> Digits include decimal characters and digits that need special handling, such as the compatibility superscript digits. This covers digits which cannot be used to form numbers in base 10, like the Kharosthi numbers.

```pycon
>>> n = '𐩃'
>>> n.isdigit()
True
>>> int(n)
  File "<python-input-6>", line 1, in <module>
    int(n)
    ~~~^^^
ValueError: invalid literal for int() with base 10: '𐩃'
```
the check should be isdecimal, or even isdecimal and isascii.

odoo/tools/config.py
```diff
@@ -1061,6 +1068,14 @@ def http_socket_activation(self):
...
+            e = f"{os.environ['ODOO_MAX_HTTP_THREADS']=} is not a positive integer"
```
xmo-odoo: Seems worse than just

```suggestion
            e = f"ODOO_MAX_HTTP_THREADS={mht} is not a positive integer"
```

kmagusiak:
@robodoo r+
codeowner: h11 added in setup.py

Julien00859:
@robodoo r-

waiting for runbot to add h11 in their dependencies

hoangtrann:
@Julien00859 should this be considered for backport to other versions too, or maybe at least the LTS ones?

Julien00859:
> @Julien00859 should this be considered for backport to other versions too, or maybe at least the LTS ones?

Hello @hoangtrann thank you for your interest in this work. As per our stable policy, the risk/gain ratio isn't good enough to be backported to previous versions, not even 19.0. Still, if you decide to do it for your own organization, and you make it open-source: I can offer to give feedback / review.

kmagusiak:
@robodoo r+

omerahmed1994:
Hi @Julien00859, thank you for this work.

Since the new HTTP server is built around h11 and a sans-I/O design, is there any long-term consideration for providing an optional asyncio/ASGI escape hatch in Odoo?

I do not mean converting the ORM or the existing WSGI stack to async. I am thinking of a limited extension point for use cases such as WebSockets, Server-Sent Events, or other long-lived I/O operations, without requiring them to depend directly on gevent.

Is this something the new server architecture could support in the future, or is continuing with gevent for evented workloads an intentional architectural direction?

If this would be better discussed elsewhere, please let me know and I’ll move the question there.

Julien00859:
Hello @omerahmed1994, our current architecture for everything async-related is indeed intentionally based on gevent. Our team would like to work on an asyncio server and to eventually phase out gevent, but a lot of people at work are weary about the "async virus" (coloring functions) and are not convinced that asyncio is the way forward. So for the foreseeable future, gevent is gonna stay, and a "proper" async/await server is not planned. Hopefully this answers your question :)

omerahmed1994:
Thank you for the clear explanation. This answers my question and helps me better understand the current architectural direction.

PR #247168, opened 2026-02-04
[IMP] mail: meeting channel revamp

maad-odoo:
Purpose of this commit:

For channels created using the New Meeting Button:
- The channel should have a default name in the format 'Meeting-{date}'

For channels with default_display_mode == 'video_full_screen':
- In the meeting view, the chat panel of the these channels should not contains any 'call' type notification message
- Rename meeting view's chat panel's name from 'Chat' ->  'In Call Messages'

task-5460982

### DIFF
--- a/addons/mail/models/discuss/discuss_channel.py
+++ b/addons/mail/models/discuss/discuss_channel.py
@@ -1227,6 +1227,7 @@ class DiscussChannel(models.Model):
         res.one("discuss_category_id", "_store_category_fields", sudo=True)
         res.attr("channel_type")
         res.attr("create_uid")
+        res.attr("create_date", predicate=lambda channel: channel.default_display_mode == "video_full_screen")
         res.many(
             "channel_member_ids",
             "_store_member_fields",
--- a/addons/mail/static/src/discuss/call/common/message_model_patch.js
+++ b/addons/mail/static/src/discuss/call/common/message_model_patch.js
@@ -1,0 +1,19 @@
+import { Message } from "@mail/core/common/message_model";
+
+import { patch } from "@web/core/utils/patch";
+
+/** @type {import("models").Message} */
+const messagePatch = {
+    get notificationHidden() {
+        if (
+            this.store.rtc.channel?.eq(this.channel_id) &&
+            this.channel_id.default_display_mode === "video_full_screen" &&
+            this.store?.meetingViewOpened &&
+            this.notificationType === "call"
+        ) {
+            return true;
+        }
+        return super.notificationHidden;
+    },
+};
+patch(Message.prototype, messagePatch);
--- a/addons/mail/static/tests/tours/discuss_channel_meeting_view_tour.js
+++ b/addons/mail/static/tests/tours/discuss_channel_meeting_view_tour.js
@@ -28,7 +28,7 @@ function getMeetingViewTourSteps({ inWelcomePage = false } = {}) {
         },
         {
             trigger:
-                ".o-mail-Meeting .o-mail-ActionPanel .o-mail-Thread:contains('john (base.group_user), bob (base.group_user), and Guest')",
+                ".o-mail-Meeting .o-mail-ActionPanel .o-mail-Thread:contains('Meeting - Jan 1, 2026')",
         },
         {
             trigger: ".o-mail-Message[data-persistent]:contains('Hello everyone!')",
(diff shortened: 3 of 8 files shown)


addons/mail/models/discuss/discuss_channel.py
```diff
@@ -76,6 +76,7 @@ def _generate_random_token(self):
         string='Channel Type', required=True, default='channel', readonly=True, help="Chat is private and unique between 2 persons. Group is private among invited persons. Channel can be freely joined (depending on its configuration).")
     is_editable = fields.Boolean('Is Editable', compute='_compute_is_editable')
     default_display_mode = fields.Selection(string="Default Display Mode", selection=[('video_full_screen', "Full screen video")], help="Determines how the channel will be displayed by default when opening it from its invitation link. No value means display text (no voice/video).")
+    is_meeting_channel = fields.Boolean(string="Meeting Channel", help="Determines whether the display name is computed from the meeting date when the channel name is empty.")
```
maad-odoo: Field was added so that the old group channels that are already present in the DB with `default_display_mode` set to `video_full_screen`, with no `name`, is not affected. Only the newly created channels using the New Meeting button will have a custom name which is computed on the client w.r.t. user's timezone.

alexkuhn: I would not bother with that, "new meeting" conversations are intended to be short-lived, so it's not a big deal for older "meeting" conversations to be affected.

The extra field here forces us to have the PR merged in master necessarily. I haven't look at the whole diff yet, but considering the freeze is soon and this task would be nice for 19.2, it would be sad to have it postponed only due to the introduction of this field.

Also we think that instead of a dedicated meeting field, we'd probably want `channel_type = "meeting"`. What makes it harder to change now is that ACL are based on `channel_type`, but are planning technical improvements to have ACL outside of `channel_type`, so that this is easier to add new `channel_type` to functional needs.

addons/mail/static/src/discuss/call/common/message_model_patch.js
```diff
@@ -0,0 +1,19 @@
...
+import { patch } from "@web/core/utils/patch";
+
+/** @type {import("models").Message} */
+const messagePatch = {
+    get notificationHidden() {
+        if (
+            this.store.discuss?.thread?.eq(this.thread) &&
+            this.thread.channel.default_display_mode === "video_full_screen" &&
+            this.store?.meetingViewOpened &&
+            this.notificationType === "call"
```
alexkuhn: Nitpick: there's `this.channel_id` to access the channel record. This is preferred over `this.thread.channel` for readability.

alexkuhn: `this.store.discuss?.thread?.eq(this.thread)` should be replaced by `this.store.rtc.channel?.eq(this.thread)`, as the conversation is not necessarily open from the discuss app: it could be open from a chat window.

alexkuhn (approved):
LGTM otherwise.

addons/mail/static/tests/tours/discuss_channel_meeting_view_tour.js
```diff
@@ -27,8 +27,20 @@ function getMeetingViewTourSteps({ inWelcomePage = false } = {}) {
...
+                })}`;
+                const currentThreadName = document.querySelector(
+                    ".o-mail-Thread div[role='heading']"
+                );
+                if (currentThreadName.textContent.trim() !== expectedThreadName) {
+                    throw new Error(
+                        `Unexpected thread name. Expected "${expectedThreadName}", got "${currentThreadName.textContent}"`
+                    );
+                }
+            },
```
alexkuhn: Dynamic values based on date are hard to cover by tests without `freezetime` / `mockDate()`.

I would personally not check the date part, this is too prone to:
- race condition around midnight (nightly builds)
- high maintenance (`new Date().toLocaleDateString()` is not exactly the same as the luxon equivalent)

I'd personally be content with `"Meeting -"` prefix for the tour. The exact test coverage with date from unit test is more than enough.

maad-odoo: I tried using the freeze_time while the group creation, but since the create_date uses the record creation transaction time stamp, that will not be the same as the freeze time date. 

See:
https://github.com/odoo/odoo/blob/9f31815b4514317be2aa126583e477673db54adc/odoo/sql_db.py#L254-L261

tsm-odoo: we have `freeze_all_time` :)

maad-odoo: It's exactly what is needed in this case, thankyou @tsm-odoo for the suggestion :smile:

alexkuhn (approved):
@robodoo delegate+ r+

PR #247169, opened 2026-02-04
[FIX] l10n_it_pos: it fiscal printer traceback

loukwets:
Issue:
When printing with the italian fiscal printer since the sync_from_ui was not awaited before printing,
the generation of the ticket was trying to access
the currency from the order that wasn't set.

Fix:
Print the ticket after the order is synced.

Current behavior before PR:

---

### DIFF
--- a/addons/point_of_sale/static/src/app/screens/feedback_screen/feedback_screen.js
+++ b/addons/point_of_sale/static/src/app/screens/feedback_screen/feedback_screen.js
@@ -45,12 +45,7 @@ export class FeedbackScreen extends Component {
                             await this.props.waitFor;
                         }
                     } finally {
-                        this.state.loading = false;
-                        if (this.isAutoSkip && !this.ignoreTimeout) {
-                            this.state.timeout = setTimeout(() => {
-                                this.pos.orderDone(this.currentOrder);
-                            }, this.pos.feedbackScreenAutoSkipDelay);
-                        }
+                        await this._afterWaitFinished();
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


addons/point_of_sale/static/src/app/screens/feedback_screen/feedback_screen.js
```diff
@@ -45,12 +45,7 @@ export class FeedbackScreen extends Component {
                             await this.props.waitFor;
                         }
                     } finally {
-                        this.state.loading = false;
-                        if (this.isAutoSkip && !this.ignoreTimeout) {
-                            this.state.timeout = setTimeout(() => {
-                                this.pos.orderDone(this.currentOrder);
-                            }, this.pos.feedbackScreenAutoSkipDelay);
-                        }
+                        this._afterWaitFinished();
```
Megaaaaaa: Calling this without awaiting means any async override will run concurrently with the rest of the flow which will reintroduce the same race issues this extraction is trying to avoid. Await the hook so both the base implementation and any async overrides are completed before the `finally` block fully resolves.
```suggestion
                        await this._afterWaitFinished();
```

addons/point_of_sale/static/src/app/screens/feedback_screen/feedback_screen.js
```diff
@@ -64,6 +59,16 @@ export class FeedbackScreen extends Component {
         });
     }
 
+    _afterWaitFinished() {
```
Megaaaaaa: So this should be async


```suggestion
    async _afterWaitFinished() {
```

Megaaaaaa (changes requested):
Goes along with my odoo/enterprise review :+1:

Megaaaaaa (approved):
Thank you for the changes, LGTM :+1:

robodoo r+

PR #247170, opened 2026-02-04
[FIX] stock: prevent new move lines from disappearing in moves view

YoussefM890:
**Problem:**
When creating a new stock.move.line in the moves view (accessed via smart button from a picking), the newly created line disappears after any refresh action (manual refresh or triggering "Put in Pack").

**Steps to reproduce:**
1. Open a receipt/picking operation
2. Click on the "Moves" smart button to open the detailed operations view
3. Create a new stock.move.line record
4. Click "Put in Pack" or manually refresh the page
5. Observe that the newly created line disappears

**Current behavior:**
The newly created stock.move.line disappears from the view after refresh, and only reappears if you navigate back to the picking and then return to the moves view.

**Expected behavior:**
The newly created stock.move.line should remain visible in the view after refresh or any action that triggers a view reload.

**Cause of the issue:**
  The action_detailed_operations method uses a static domain [('id', 'in', self.move_line_ids.ids)] that captures a snapshot of move line IDs at the moment the action is opened.
  https://github.com/odoo/odoo/blob/22ac818970f104a732cc7d24afc440cf0e6d74bd/addons/stock/models/stock_picking.py#L1204-L1212 When a new stock.move.line is created in this view, its ID is not included in the original static list. Any refresh (manual or triggered by operations like "Put in Pack") re-applies this static domain, filtering out the newly created lines because their IDs weren't captured in the initial list.

**Fix:**
Using a dynamic domain based on picking_id ensures all move lines belonging to the picking are always visible, regardless of when they were created. This aligns with the expected behavior of showing "all move lines for this picking" rather than "only the move lines that existed when the view was opened". The relational lookup [('picking_id', '=', self.id)] is re-evaluated on each refresh, automatically including any newly created lines that have the correct picking_id set.

opw-5398620

### DIFF
--- a/addons/stock/tests/test_move_lines.py
+++ b/addons/stock/tests/test_move_lines.py
@@ -271,3 +271,35 @@ class StockMoveLine(TestStockCommon):
 
         line = move.move_line_ids[0]
         self.assertEqual(line.lot_id, serial_lot)
+
+    def test_action_detailed_operations_domain_includes_new_lines(self):
+        """Test that newly created move lines remain visible in detailed operations view."""
+        receipt = self.env['stock.picking'].create({
+            'picking_type_id': self.picking_type_in,
+            'location_id': self.supplier_location,
+            'location_dest_id': self.stock_location,
+            'move_ids': [Command.create({
+                'name': 'Test Move',
+                'product_id': self.productA.id,
+                'product_uom_qty': 10.0,
+                'product_uom': self.productA.uom_id.id,
+                'location_id': self.supplier_location,
+                'location_dest_id': self.stock_location,
+            })],
+        })
+        receipt.action_confirm()
+        initial_move_line = receipt.move_line_ids
+        self.assertRecordValues(initial_move_line, [{'quantity': 10.0}])
+        action = receipt.action_detailed_operations()
+        self.assertEqual(initial_move_line, self.env['stock.move.line'].search(action['domain']))
+
+        new_move_line = self.env['stock.move.line'].create({
+            'picking_id': receipt.id,
+            'product_id': self.productA.id,
+            'product_uom_id': self.productA.uom_id.id,
+            'location_id': self.supplier_location,
+            'location_dest_id': self.stock_location,
+            'move_id': receipt.move_ids.id,
+            'quantity': 5.0,
+        })
+        self.assertEqual(initial_move_line | new_move_line, self.env['stock.move.line'].search(action['domain']))
--- a/addons/stock/models/stock_picking.py
+++ b/addons/stock/models/stock_picking.py
@@ -1209,7 +1209,7 @@ class Picking(models.Model):
             'type': 'ir.actions.act_window',
             'res_model': 'stock.move.line',
             'views': [(view_id, 'list')],
-            'domain': [('id', 'in', self.move_line_ids.ids)],
+            'domain': [('picking_id', '=', self.id)],
             'context': {
                 'sml_specific_default': True,
                 'default_picking_id': self.id,


addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.stock_location = cls.env.ref('stock.stock_location_stock')
+        cls.customer_location = cls.env.ref('stock.stock_location_customers')
+        cls.supplier_location = cls.env.ref('stock.stock_location_suppliers')
+        cls.picking_type_in = cls.env.ref('stock.picking_type_in')
+        cls.product = cls.env['product.product'].create({
+            'name': 'Test Product',
+            'is_storable': True,
+        })
```
lase-odoo: There is no need to create a new test class for this test, imo the test could be in the `StockMoveLine` test class in `/odoo/addons/stock/tests/test_move_lines.py`. Or in the second move test file.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_action_detailed_operations_domain_includes_new_lines(self):
+        """Test that newly created move lines remain visible in detailed operations view.
+
+        This test verifies the fix for the issue where creating a new stock.move.line
+        in the detailed operations view would cause it to disappear after refresh or
+        when using the "Put in Pack" button.
```
lase-odoo: This second part of the description isn't really necessary. The first part of the description is crystal clear and since the issue is well referenced by your commit message, any developer will be able to know the context of your fix by "blaming" the code.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        # Create a receipt picking
```
lase-odoo: No need to comment such standard steps. Actually you can imo remove all the comments in this test.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+        # Add a move to the picking
+        move = self.env['stock.move'].create({
+            'name': 'Test Move',
+            'product_id': self.product.id,
+            'product_uom_qty': 10.0,
+            'product_uom': self.product.uom_id.id,
+            'picking_id': picking.id,
+            'location_id': self.supplier_location.id,
+            'location_dest_id': self.stock_location.id,
+        })
```
lase-odoo: You can and should create these records in a single create. This helps rely on the orm optimisations in order to minimize the amount of sql transactions. If this is not yet imported, in the test file you can 
```py
from odoo.fields import Command
```
Then:
```suggestion
        receipt = self.env['stock.picking'].create({
            'picking_type_id': self.picking_type_in.id,
            'location_id': self.supplier_location.id,
            'location_dest_id': self.stock_location.id,
            'move_ids': [Command.create({
                'name': 'Test Move',
                'product_id': self.product.id,
                'product_uom_qty': 10.0,
                'product_uom': self.product.uom_id.id,
                'location_id': self.supplier_location.id,
                'location_dest_id': self.stock_location.id,
            })],
        })
```

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+
```
lase-odoo: It is also best not to put empty lines unless you want to improve the readability at precise plcaes for instance between the set up, the start of the flow and the asserts.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+        # Get the action for detailed operations
+        action = picking.action_detailed_operations()
+
+        # Verify the domain uses picking_id instead of static ID list
+        self.assertEqual(
+            action['domain'],
+            [('picking_id', '=', picking.id)],
+            "Domain should filter by picking_id to include newly created lines"
+        )
```
lase-odoo: I would not make this assert, It is fine if the domain is not the same, it is only important that the relevant records appear and that the un-relevant one do not.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+            'product_id': self.product.id,
+            'product_uom_qty': 10.0,
+            'product_uom': self.product.uom_id.id,
+            'picking_id': picking.id,
+            'location_id': self.supplier_location.id,
+            'location_dest_id': self.stock_location.id,
+        })
+
+        picking.action_confirm()
+        picking.action_assign()
```
lase-odoo: This step is not necessary, the assignment is automatically performed during move confirmation. What is important in general is to have enough units at the picking location to be assigned before confirmation. BUT, since your flow is done with a receipt, the picking will bypass the reservation process anyway and it will automatically be reserved at confirmation !

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_action_detailed_operations_domain_includes_new_lines(self):
+        """Test that newly created move lines remain visible in detailed operations view.
+
+        This test verifies the fix for the issue where creating a new stock.move.line
+        in the detailed operations view would cause it to disappear after refresh or
+        when using the "Put in Pack" button.
+        """
+        # Create a receipt picking
+        picking = self.env['stock.picking'].create({
```
lase-odoo: It is nicer to use precise names in case it is relevant, for instance, here the picking is a receipt, it is nicer to know it than reading the `picking_type_id` to determine the flow.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+            filtered_lines.ids,
+            "Newly created move line should be visible after domain re-evaluation"
+        )
+
+        for original_line_id in initial_move_line_ids:
+            self.assertIn(
+                original_line_id,
+                filtered_lines.ids,
+                "Original move lines should still be visible"
+            )
```
lase-odoo: To resume here is my suggestion:
```suggestion
    def test_action_detailed_operations_domain_includes_new_lines(self):
        """Test that newly created move lines remain visible in detailed operations view.
        """
        receipt = self.env['stock.picking'].create({
            'picking_type_id': self.picking_type_in.id,
            'location_id': self.supplier_location.id,
            'location_dest_id': self.stock_location.id,
            'move_ids': [Command.create({
                'name': 'Test Move',
                'product_id': self.product.id,
                'product_uom_qty': 10.0,
                'product_uom': self.product.uom_id.id,
                'location_id': self.supplier_location.id,
                'location_dest_id': self.stock_location.id,
            })],
        })
        receipt.action_confirm()
        initial_move_line = receipt.move_line_ids
        self.assertRecordValues(initial_move_line, [{'quantity': 10.0}])
        action_detailed_operations = receipt.action_detailed_operations()
        self.assertEqual(initial_move_line, self.env['stock.move.line'].search(action_detailed_operations['domain']))

        new_move_line = self.env['stock.move.line'].create({
            'picking_id': receipt.id,
            'product_id': self.product.id,
            'product_uom_id': self.product.uom_id.id,
            'location_id': self.supplier_location.id,
            'location_dest_id': self.stock_location.id,
            'move_id': receipt.move_ids.id,
            'quantity': 5.0,
        })
        self.assertEqual(initial_move_line | new_move_line, self.env['stock.move.line'].search(action_detailed_operations['domain']))
```

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Verify the domain uses picking_id instead of static ID list
+        self.assertEqual(
+            action['domain'],
+            [('picking_id', '=', picking.id)],
+            "Domain should filter by picking_id to include newly created lines"
+        )
+
+        # Get initial move line IDs
+        initial_move_line_ids = picking.move_line_ids.ids
```
lase-odoo: it is usually not relevant to assert on ids, recorset equality is performed through ids automatically.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Verify that both the original and new move lines are included
+        self.assertIn(
+            new_move_line.id,
+            filtered_lines.ids,
+            "Newly created move line should be visible after domain re-evaluation"
+        )
+
+        for original_line_id in initial_move_line_ids:
+            self.assertIn(
```
lase-odoo: it is imo better to assert on the record set directly than to assert on all possible inclusions. Also, your test is not testing that there is no extra records in the result.

addons/stock/tests/test_detailed_operations.py
```diff
@@ -0,0 +1,90 @@
...
+        action = picking.action_detailed_operations()
+
+        # Verify the domain uses picking_id instead of static ID list
+        self.assertEqual(
+            action['domain'],
+            [('picking_id', '=', picking.id)],
+            "Domain should filter by picking_id to include newly created lines"
+        )
+
+        # Get initial move line IDs
```
lase-odoo: As a general rule of thumb I would not phrase `ID` in capital letters.

lase-odoo (commented):
Thanks a lot for the propositon ! 🌊

Great explanations, amazing fix ! 

I left a review on the test ! 

Also, I think it would be nice to also test the put in pack behavior this can be done in the same test by packing the move before creating the extra move :)

addons/stock/tests/test_move_lines.py
```diff
@@ -271,3 +271,35 @@ def test_lot_creation_from_move_line_with_generic_stock(self):
...
+                'location_id': self.supplier_location,
+                'location_dest_id': self.stock_location,
+            })],
+        })
+        receipt.action_confirm()
+        initial_move_line = receipt.move_line_ids
+        self.assertRecordValues(initial_move_line, [{'quantity': 10.0}])
+        action_detailed_operations = receipt.action_detailed_operations()
+        self.assertEqual(initial_move_line, self.env['stock.move.line'].search(action_detailed_operations['domain']))
+        receipt.action_put_in_pack()
```
lase-odoo: Careful, in the flow you describe the action that is porcessed is the `action_put_in_pack` of the `stock.move.line` model and this action rely on the context of the current action: 
https://github.com/odoo/odoo/blob/de056cc784a3bbe2575fd3c9e81ca62e73c362d4/addons/stock/models/stock_picking.py#L1845-L1848
So if we had to test your flow porperly, we would need to do something like:
```py
     initial_move_line.with_context(action_detailed_operations['context']).action_detailed_operations()
```
But, after testing your flow I realised that only the non-preexisting move lines disappear from the view when you put in pack (precisely because we proceed with a `web_search_read` on the domain of the action, just as if you refreshed the page). I thought that in that case the issue was that a new move line was created referring to the package. So that it is probably not relevant to test the `put_in_pack behavior` here
```suggestion
```
Sorry for the suggestion 😅

PR #247176, opened 2026-02-04
[FIX] account,sale: ensure product column visibility in tour

epictete:
Commit 01848b5 missed a case causing the `test_add_section_from_product_catalog_on_invoice_tour` tour to fail.

To resolve this, `showProductColumn()` is now called automatically within `addSectionFromProductCatalog()`. This ensures the product column is visible before content verification, removing the need for manual calls in other tours.

Additionally, the `showProductColumn()` trigger is updated to support both `product_id` and `product_template_id`, ensuring compatibility with the `purchase_product_matrix` module.

runbot-234872

---

### DIFF
--- a/addons/account/static/src/js/tours/tour_utils.js
+++ b/addons/account/static/src/js/tours/tour_utils.js
@@ -8,7 +8,7 @@ export function showProductColumn() {
         },
         {
             content: "Show product column",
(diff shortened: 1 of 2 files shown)


FlorianGilbert (approved):
@robodoo r+

PR #247179, opened 2026-02-04
[IMP] account: include draft vendor bills in 'Unpaid' filter

devp-odoo:
In this commit:
- Updated the `Unpaid` filter to show draft bills in addition to posted bills. The filter now includes all non-cancelled bills with payment status `Not Paid` or `Partially Paid`.
- Backported the logic from 18.0 to ensure journal entries are filtered out by checking that type is not equal to `journal_entry`.

task-5900283


khva-odoo (approved):

gawa-odoo (commented):
@robodoo r+

PR #247182, opened 2026-02-04
[REF] spreadsheet: reuse default operator and adapt editor filter values

rmbh-odoo:
## Description of the issue/feature this PR addresses:

Current behavior before PR:
- Default operator logic was embedded in getDefaultValue().
- Text filter allowed values were always computed from the global filter id.

- Extracted default operator logic into getDefaultOperator() for reuse.
- Editor panel now computes text filter allowed values directly from ranges.

Task: [5850422](https://www.odoo.com/odoo/project/2328/tasks/5850422)

---

### DIFF
--- a/addons/spreadsheet/static/src/global_filters/components/filter_value/filter_value.js
+++ b/addons/spreadsheet/static/src/global_filters/components/filter_value/filter_value.js
@@ -38,6 +38,7 @@ export class FilterValue extends Component {
         globalFilterValue: { optional: true },
         showTitle: { type: Boolean, optional: true },
         showClear: { type: Boolean, optional: true },
+        rangesOfAllowedValues: { type: Array, optional: true },
     };
 
     setup() {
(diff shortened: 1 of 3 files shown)


addons/spreadsheet/static/src/global_filters/components/filter_value/filter_value.js
```diff
@@ -38,6 +38,8 @@ export class FilterValue extends Component {
         globalFilterValue: { optional: true },
         showTitle: { type: Boolean, optional: true },
         showClear: { type: Boolean, optional: true },
+        isEditorPanel: { type: Boolean, optional: true },
+        rangesOfAllowedValues: { type: Array, optional: true },
```
hokolomopo: not sure about the name `isEditorPanel`, the component should not be aware of its parent. `shouldLimitValuesToRange` or something ? 

But is this props even needed ? Can't we just make the condition on if the props `rangesOfAllowedValues` is defined or not ?

LucasLefevre:
robodoo r+

PR #247184, opened 2026-02-04
[FIX] pos_self_order: fix qr for pos_self_order

aksi-odoo:
Step to reproduce:
- setup Restaurant and do the following config
- set "self-ordering" -> "Qr Menu + Ordering"
- set "service At" -> "pickup zone"
- save and print the qr-code ( option is right there below these configs)

Observation:
- in downloaded pdf, we see a wrong url, which includes table_id, which shouldn't be.
- QRs do not have table_id

Cause:
- there is issue in `generate_qr_codes_page` method, which uses table_id even if ordering mode is "pickup zone"

Fix:
- we only use table_id when ordering_mode = 'table'.

**Before:** 
image

**After**
image

opw-5095607

---

### DIFF
--- a/addons/pos_self_order/models/res_config_settings.py
+++ b/addons/pos_self_order/models/res_config_settings.py
@@ -145,17 +145,16 @@ class ResConfigSettings(models.TransientModel):
         """
         Generate the data needed to print the QR codes page
         """
+        name = ""
+        url = url_unquote(self.pos_config_id._get_self_order_url())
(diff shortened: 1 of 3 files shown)


alsh-odoo (approved):
Thanks for the fix!!

Quick question: should we still require at least one table when the service mode is not table (e.g., pickup)?

If yes, the current implementation looks good 👍
If not, we could include `self.pos_self_ordering_service_mode == 'table'` in the main `if` condition so the validation applies only in table mode.

davidmonnom (approved):
@robodoo r+

PR #238513, opened 2025-12-03
[IMP] orm: binary stored as bytes (not base64) 

kmagusiak:
Since we are going to move to a raw bytes representation in the cache, let's store binary fields that are not attachments as raw bytes in the database instead of storing them in base64-encoded raw bytes (which just use more space).

task-4251301
https://github.com/odoo/upgrade/pull/8995

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247189, opened 2026-02-04
[FIX] base_import: field must work with base64 data

kmagusiak:
Since #238513, all Binary fields work the same way. The data is stored in base64 format in cache (no exceptions). Here aligning `file` field of that module did not work this way.

Current behavior before PR:
`file = fields.Binary(...)` expected to store raw bytes.

This is not supported as all Binary fields work the same now (base64).

---

### DIFF
--- a/addons/base_import/controllers/main.py
+++ b/addons/base_import/controllers/main.py
@@ -1,22 +1,20 @@
-# -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+import base64
 import json
 
 from odoo import http
 from odoo.http import request
-from odoo.tools import misc
 
 
 class ImportController(http.Controller):
 
     @http.route('/base_import/set_file', methods=['POST'])
     # pylint: disable=redefined-builtin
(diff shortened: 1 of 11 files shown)


addons/base_import/controllers/main.py
```diff
@@ -1,22 +1,20 @@
...
 
 class ImportController(http.Controller):
 
     @http.route('/base_import/set_file', methods=['POST'])
     # pylint: disable=redefined-builtin
-    def set_file(self, id):
-        file = request.httprequest.files.getlist('ufile')[0]
-
+    def set_file(self, id, ufile, model=None):
+        file = ufile
```
Julien00859, line 15: "ufile" is just a weird convention used in some controllers, go ahead and have the parameter be `file` right away

rco-odoo (approved):
LGTM

@robodoo delegate+

Julien00859 (approved):

addons/base_import/controllers/main.py
kmagusiak: Then I need to change all js files too...

PR #226909, opened 2025-09-12
[FIX] calendar: compute stop date when updating start in quick-edit

reth-odoo:
`CalendarEvent._compute_stop` relies on "duration" being somehow available when "start" is changed. In quick-create, where the duration is not stored we need to "store" it in the front-end between onchange calls.

For this reason "duration" needs to be force_save and invisible on the view so that it is not recomputed each time. Which otherwise defeats the purpose of the feature.

In appointment duration is also used in default_get hence fetching a default value for it by having it in view actually changes the behavior of the default values. Though this can be mitigated in other, more reliable ways, this fix is sufficient for that case too.

The field was removed during view refactoring in
odoo/enterprise@c8eb4fe4ba9938b1c7e07034e6ca7c08a0d6f215
and 81e51985f2e01ee3ff115477ab324e2ad75a1b77

task-5081903

PR #247193, opened 2026-02-04
[FIX] calendar: preserve duration when editing quick-created events

utma-odoo:
Currently, an incorrect duration is displayed when the start or end time 
is `manually changed` during event creation.

**Steps to reproduce:**
- Install the `Calendar` module and open the app.
- Drag on the calendar to create a `2-hour` time slot (quick-create popup opens).
- Manually adjust the start or end time to make the event `3 hours` long.
- Click `Save & Close`.
- Click on the event: it correctly displays (3 hours).
- Click `Edit` and observe the `Duration` value.

**Observation:**
The duration field shows 2 hours instead of 3 hours.

**Root cause:**
- The `duration` field is not available (and therefore not stored) in the 
 `quick-create view` at [1].
- When the `stop` time is set manually, the `duration` is computed at [2].
- When the `start` time is changed, the `stop` time is computed based on the 
  previously computed `duration` at [3].

**Fix:**
This commit adds the `duration` field to the `quick-create` view as `invisible` 
(preventing it from being recomputed on each onchange) and `force_save`. 
This aligns the behavior with `19.0` by preserving the last computed 
duration in the front-end model, as implemented in PR [4].

[1]:
https://github.com/odoo/odoo/blob/84571c03ff38ee768ed135bef3afa98511e3ab7b/addons/calendar/views/calendar_views.xml#L294-L339

[2]:
https://github.com/odoo/odoo/blob/84571c03ff38ee768ed135bef3afa98511e3ab7b/addons/calendar/models/calendar_event.py#L353-L356

[3]:
https://github.com/odoo/odoo/blob/84571c03ff38ee768ed135bef3afa98511e3ab7b/addons/calendar/models/calendar_event.py#L358-L374

[4]:
https://github.com/odoo/odoo/pull/226909

opw-5867946

### DIFF
--- a/addons/calendar/views/calendar_views.xml
+++ b/addons/calendar/views/calendar_views.xml
@@ -305,6 +305,11 @@
                         <h1 class="w-100"><field name="name" nolabel="1" placeholder="Add title" colspan="2"/></h1>
                     </div>
                     <group>
+                        <!--when "stop" is set manually, "duration" is computed. When "start" is changed
+                        "stop" is computed based on the previously-computed "duration". Hence we need to
+                        keep the last computed duration in the front-end model and send it with "start"
+                        when it changes-->
+                        <field name="duration" invisible="1" force_save="1"/>
                         <field name="start_date" widget="daterange" options="{'end_date_field': 'stop_date'}" invisible="not allday" required="allday"/>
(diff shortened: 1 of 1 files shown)


utma-odoo:
@robodoo up to saas-18.4

alsh-odoo (approved):
Backport of https://github.com/odoo/odoo/pull/226909

abla001 (approved):
LGTM

PR #121757, opened 2023-05-19
[FIX] base: prevent traceback when Portal User Template is not found

sauc-odoo:
When user deletes 'Portal User Template' record from 'res.users' model and when 'action_open_template_user' function is executed at that time traceback is generated on the user side as well as in the log.

See this traceback: 
```
ValueError: Invalid template user. It seems it has been deleted.
  File "odoo/http.py", line 2115, in __call__
    response = request._serve_db()
  File "odoo/http.py", line 1698, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "odoo/service/model.py", line 134, in retrying
    result = func()
  File "odoo/http.py", line 1725, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "odoo/http.py", line 1922, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "addons/website/models/ir_http.py", line 234, in _dispatch
    response = super()._dispatch(endpoint)
  File "odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "odoo/http.py", line 715, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "addons/web/controllers/dataset.py", line 32, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "addons/web/controllers/dataset.py", line 24, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "odoo/api.py", line 448, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "odoo/addons/base/models/res_config.py", line 763, in action_open_template_user
    raise ValueError(_('Invalid template user. It seems it has been deleted.'))
```

Steps to reproduce:
    1. Install Website module.
    2. Go to settings > Users&Companies > Users.
    3. Remove Internal Users from default filter.
    4. Add Inactive Users in filter and delete Portal User Template.
    5. Go to Website Settings > Privacy > Customer Account.
    6. Click on Free sign up and then on Save button.
    7. Now in Website Settings click on Default Access Rights button.
    8. The error will occur.

Applying this commit will fix this issue.

sentry-4184456381

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #224952, opened 2025-09-01
[IMP] stock: improve UX when splitting picking

dhha-odoo:
When splitting a picking, users often need to configure the backorder,
such as adjusting the quantity to keep, setting a scheduled date, or
preparing it for further splits. Currently, after the split, the system
remains on the original picking. In practice, however, users primarily
need to continue working on the backorder, not the original picking.

This forces users to find the backorder or follow the chatter link manually,
which is not smooth and wastes time.

Additionally, the reservation of the picking now follows based on the operation’s reservation method
for both the original picking and the backorder created from it (already follow it).

This commit improves usability by directly redirecting to the backorder and adding a log note
of the original picking, reduces unnecessary clicks, and makes it easier to track and manage backorders.

Task Id:- 5003505

PR #247195, opened 2026-02-04
[IMP] stock, product: enforce barcode uniqueness across stock models

jodh-odoo:
Barcodes are expected to uniquely identify a single record in the system.
While uniqueness was enforced between products and packagings, other
stock models were not included in this constraint.

This allowed the same barcode to be assigned across multiple entities
such as products, lots, packagings, locations, operation types, packages
or package types, leading to ambiguous resolution in the Barcode app
and blocking warnings during scanning like

- **Barcode scan is ambiguous with several model: product.product
  and stock.package.type. Use the most likely.**

Users were unable to perform reliable operations when such conflicts existed.

This improvement ensures barcode uniqueness also applies between
- Products,
- Packagings,
- Locations,
- Operation Types,
- Package Types,
- Packages,
- Lot/Serial,

preventing ambiguity and making barcode scanning more predictable
and reliable for users.

Ent PR - [#121757](https://github.com/odoo/enterprise/pull/121757)
Task-5417767

### DIFF
--- a/addons/stock/models/barcode_uniqueness_mixin.py
+++ b/addons/stock/models/barcode_uniqueness_mixin.py
@@ -1,0 +1,50 @@
+from odoo import api, models
+from odoo.exceptions import ValidationError
+from odoo.fields import Domain
+from odoo.tools import groupby
+
+BARCODE_UNIQUENESS_MAP = {
+    'stock.location': {'field': 'barcode', 'label': 'Location'},
+    'product.product': {'field': 'barcode', 'label': 'Product', 'skip_model': 'product.uom'},
+    'product.uom': {'field': 'barcode', 'label': 'Packaging', 'skip_model': 'product.product'},
+    'stock.package.type': {'field': 'barcode', 'label': 'Package Type'},
+    'stock.package': {'field': 'name', 'label': 'Package'},
+    'stock.lot': {'field': 'name', 'label': 'Lot/Serial'},
+    'stock.picking.type': {'field': 'barcode', 'label': 'Operation Type'},
+}
+
+
+class BarcodeUniquenessMixin(models.AbstractModel):
+    _name = 'barcode.uniqueness.mixin'
+    _description = "Barcode Uniqueness Mixin"
+    _explanation = "Provides a reusable mechanism to ensure barcodes remain unique across different inventory objects, prevents assigning the same barcode to conflicting records."
+
+    @api.constrains(lambda self: self._get_barcode_field_name())
+    def _check_barcode_uniqueness(self):
+        for company_id, barcodes_within_company in self._group_barcodes_by_company():
+            if barcodes_within_company:
+                self._check_duplicated_barcodes(barcodes_within_company, company_id)
+
+    def _check_duplicated_barcodes(self, barcodes_within_company, company_id):
+        for model_name, config in BARCODE_UNIQUENESS_MAP.items():
+            if self._name in (model_name, config.get('skip_model')):
+                continue
+            domain = self.env[model_name]._get_duplicate_barcode_domain(config['field'], barcodes_within_company, company_id)
+            if self.env[model_name].search_count(domain, limit=1):
+                raise ValidationError(self.env._("The barcode is already assigned to a %(label)s.", label=config['label']))
+
+    def _get_barcode_field_name(self):
+        return [BARCODE_UNIQUENESS_MAP[self._name]['field']]
+
+    def _get_duplicate_barcode_domain(self, field_name, barcodes, company_id):
+        domain = Domain(field_name, 'in', barcodes)
+        if company_id:
+            domain = Domain.AND([domain, Domain('company_id', 'in', (False, company_id))])
+        return domain
+
+    def _group_barcodes_by_company(self):
+        field_name = self._get_barcode_field_name()[0]
+        return [
+            (company_id, [r[field_name] for r in records if r[field_name]])
+            for company_id, records in groupby(self, lambda r: r.company_id.id)
+        ]
--- a/addons/stock/models/product.py
+++ b/addons/stock/models/product.py
@@ -43,7 +43,8 @@ SERIAL_PREFIX_FORMAT_HELP_TEXT = """
 
 
 class ProductProduct(models.Model):
-    _inherit = "product.product"
+    _name = 'product.product'
+    _inherit = ['product.product', 'barcode.uniqueness.mixin']
 
     stock_quant_ids = fields.One2many('stock.quant', 'product_id') # used to compute quantities
     stock_move_ids = fields.One2many('stock.move', 'product_id') # used to compute quantities
@@ -515,6 +516,12 @@ class ProductProduct(models.Model):
             product.nbr_reordering_rules = count
             product.reordering_min_qty = product_min_qty_sum
             product.reordering_max_qty = product_max_qty_sum
+
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        for company_id, barcodes_within_company in self._get_barcodes_by_company():
+            self._check_duplicated_barcodes(barcodes_within_company, company_id)
 
     @api.onchange('tracking')
     def _onchange_tracking(self):
@@ -1326,3 +1333,14 @@ class UomUom(models.Model):
         else:
             computed_qty = self._compute_quantity(qty, procurement_uom, rounding_method='HALF-UP')
         return (computed_qty, procurement_uom)
+
+
+class ProductUom(models.Model):
+    _name = 'product.uom'
+    _inherit = ['product.uom', 'barcode.uniqueness.mixin']
+
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        for company_id, barcodes_within_company in self._group_barcodes_by_company():
+            self._check_duplicated_barcodes(barcodes_within_company, company_id)
--- a/addons/stock/models/stock_lot.py
+++ b/addons/stock/models/stock_lot.py
@@ -23,7 +23,7 @@ PY_OPERATORS = {
 
 class StockLot(models.Model):
     _name = 'stock.lot'
-    _inherit = ['mail.thread', 'mail.activity.mixin']
+    _inherit = ['mail.thread', 'mail.activity.mixin', 'barcode.uniqueness.mixin']
     _description = 'Lot/Serial'
     _check_company_auto = True
     _order = 'name, id'
@@ -106,9 +106,6 @@ class StockLot(models.Model):
         domain = [('product_id', 'in', self.product_id.ids),
                   ('name', 'in', self.mapped('name'))]
         groupby = ['company_id', 'product_id', 'name']
-        if any(not lot.company_id for lot in self):
-            # We need to check across other companies to not have duplicates between 'no-company' and a company.
-            self = self.sudo()
         records = self.with_context(skip_preprocess_gs1=True)._read_group(domain, groupby, ['__count'], order='company_id DESC')
         error_message_lines = set()
         cross_lots = {}
--- a/addons/stock/models/stock_package_type.py
+++ b/addons/stock/models/stock_package_type.py
@@ -6,6 +6,7 @@ from odoo import api, fields, models
 
 class StockPackageType(models.Model):
     _name = 'stock.package.type'
+    _inherit = ['barcode.uniqueness.mixin']
     _description = "Stock package type"
     _order = "sequence, id"
 
@@ -124,6 +125,9 @@ class StockPackageType(models.Model):
             if seq_to_todo_ids:
                 self.env['ir.sequence'].browse(list(seq_to_todo_ids)).sudo().write(seq_vals)
         return super().write(vals)
+
+    def _group_barcodes_by_company(self):
+        return [(False, list(filter(None, self.mapped('barcode'))))]
 
     def _get_next_name_by_sequence(self):
         if len(self) == 1 and self.sequence_id:
--- a/addons/stock/models/stock_picking.py
+++ b/addons/stock/models/stock_picking.py
@@ -17,6 +17,7 @@ from odoo.tools.misc import clean_context
 
 class StockPickingType(models.Model):
     _name = 'stock.picking.type'
+    _inherit = ['barcode.uniqueness.mixin']
     _description = "Picking Type"
     _explanation = "Defines the type of stock operation (e.g., Receipts, Deliveries, Internal Transfers) and contains configuration for how these operations should behave in the warehouse."
     _order = 'is_favorite desc, sequence, id'
@@ -159,6 +160,11 @@ class StockPickingType(models.Model):
         ('direct', 'As soon as possible, with back orders'), ('one', 'When all products are ready')], 'Shipping Policy',
         default=lambda self: self.env.company.picking_policy, required=True,
         help="It specifies goods to be transferred partially or all at once")
+
+    _check_picking_type_barcode_company_unique = models.Constraint(
+        'unique (barcode,company_id)',
+        'The barcode for an operation type must be unique per company!',
+    )
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/addons/product/models/product_product.py
+++ b/addons/product/models/product_product.py
@@ -329,12 +329,12 @@ class ProductProduct(models.Model):
 
     def _check_duplicated_product_barcodes(self, barcodes_within_company, company_id):
         domain = self._get_barcode_search_domain(barcodes_within_company, company_id)
-        products_by_barcode = self.sudo()._read_group(
+        products_by_barcode = self._read_group(
             domain, ['barcode'], ['id:recordset'], having=[('__count', '>', 1)],
         )
 
         duplicates_as_str = "\n".join(
-            self._build_duplicate_barcode_error_string(barcode, duplicate_products._filtered_access('read'))
+            self._build_duplicate_barcode_error_string(barcode, duplicate_products.sudo(False)._filtered_access('read'))
             for barcode, duplicate_products in products_by_barcode
         )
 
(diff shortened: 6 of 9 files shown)


addons/stock/models/product.py
```diff
@@ -119,6 +119,18 @@ class ProductProduct(models.Model):
...
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        self_ctx = self.with_context(skip_preprocess_gs1=True)
+        for company_id, barcodes_within_company in self_ctx._get_barcodes_by_company():
+            self_ctx._check_duplicated_package_type_barcodes(barcodes_within_company, company_id)
+
+    def _check_duplicated_package_type_barcodes(self, barcodes_within_company, company_id):
+        package_domain = self._get_barcode_search_domain(barcodes_within_company, company_id)
+        if self.env['stock.package.type'].sudo().search_count(package_domain, limit=1):
```
dhha-odoo: Why is `sudo()` used here?
Is there any specific case where it’s required?

From what I see, this code only reads `package type records (used for counting)`, and normal users already have `read access`. So this shouldn’t raise an `access error`.

Do you see any scenario where `sudo()` is actually needed here?

addons/stock/models/product.py
```diff
@@ -119,6 +119,18 @@ class ProductProduct(models.Model):
...
+            raise ValidationError(self.env._("A package type already uses the barcode."))
```
dhha-odoo:
```suggestion
            raise ValidationError(self.env._("This barcode is already assigned to a package type."))
```
More understandable? Still, you can confirm with PO.
Same for all

addons/stock/models/product.py
```diff
@@ -119,6 +119,18 @@ class ProductProduct(models.Model):
...
 
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        self_ctx = self.with_context(skip_preprocess_gs1=True)
+        for company_id, barcodes_within_company in self_ctx._get_barcodes_by_company():
+            self_ctx._check_duplicated_package_type_barcodes(barcodes_within_company, company_id)
+
+    def _check_duplicated_package_type_barcodes(self, barcodes_within_company, company_id):
+        package_domain = self._get_barcode_search_domain(barcodes_within_company, company_id)
```
dhha-odoo:
```suggestion
        domain = self._get_barcode_search_domain(barcodes_within_company, company_id)
```
`_get_barcode_search_domain` doesn’t return a `package-specific domain`.
It only builds a generic barcode domain, e.g. `[('barcode', 'in', ['dhha1'])]`.

addons/stock/models/product.py
```diff
@@ -1248,3 +1260,14 @@ def _adjust_uom_quantities(self, qty, quant_uom):
...
         return (computed_qty, procurement_uom)
+
+
+class ProductUom(models.Model):
+    _inherit = 'product.uom'
+
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        domain = [('barcode', 'in', [b for b in self.mapped('barcode') if b])]
```
dhha-odoo:
```suggestion
        domain = [('barcode', 'in', self.mapped('barcode'))]
```

The `barcode field is required`, so a `record without a barcode cannot be saved` and will already raise a `missing required field` error.
Because of this, self will never contain `records without a barcode`, so there’s no need to check for `empty barcodes` here.

Also, even when multiple records are updated at once, we don’t need a loop.
Using `self.mapped('barcode')` is enough; it returns a list and works correctly for `multi-record updates`. I verified this by enabling `multi_edit="1"` in the tree view and `updating barcodes for multiple UoMs at the same time`.

addons/stock/models/stock_package_type.py
```diff
@@ -59,6 +60,14 @@ def _get_default_weight_uom(self):
         'Max Weight must be positive',
     )
 
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        domain = [('barcode', 'in', [b for b in self.mapped('barcode') if b])]
+        product = self.env['product.product']
+        packaging = self.env['product.uom']
+        if product.search_count(domain, limit=1) or packaging.search_count(domain, limit=1):
```
dhha-odoo:
```suggestion
        domain = [('barcode', 'in', self.mapped('barcode'))]
        if self.env['product.product'].search_count(domain, limit=1) or self.env['product.uom'].search_count(domain, limit=1):
```

addons/stock/models/stock_package_type.py
dhha-odoo, line 1: IMO, this check should be `company-dependent`.
If a `product exists in another company with the same barcode`, adding that `barcode to a package type in a different company` would still raise a `validation error`, even though the `product doesn’t exist in the current company`.

Shouldn’t the `barcode uniqueness be enforced per company` instead of globally?

dhha-odoo (changes requested):
Hello @jodh-odoo 

Since this PR targets the master branch and is tagged as a bug improvement, I think it would be more appropriate to use the IMP format, as the changes are mainly improvements.

Could you please adapt it accordingly?

Thanks

addons/stock/models/stock_package_type.py
jodh-odoo: You’ve got a point: with the global check, a validation error would be raised for a product the user can’t even see due to access rights.
However, if we switch to company-dependent checks, we reintroduce the ambiguity issue at scan time, as package types can still be scanned across companies, which is exactly what this task aims to solve.
(i've attached a video showing the exact case in company-dependent check)
We could revisit making the checks company-wise once package types have strict company-level access (as planned in this task: [Task](https://www.odoo.com/odoo/project/966/tasks/3717250)).
I’d be interested to hear your thoughts on this 😊

https://github.com/user-attachments/assets/29f7b427-3ffc-45c7-8243-c117d9ac4535

addons/stock/models/product.py
jodh-odoo: You're right, However there is a scenario where a user had access to create products but no access to package types. Even in that case, the constraint still works correctly without sudo(), since it is already executed in sudo mode by the ORM. So the explicit sudo() isn’t necessary, I’ll remove it.

addons/stock/models/stock_package_type.py
dhha-odoo: > We could revisit making the checks company-wise once package types have strict company-level access (as planned in this task: [Task](https://www.odoo.com/odoo/project/966/tasks/3717250)).

In that case, it would be better to handle this after the that task, so we don’t block this one.

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,61 @@
...
+class BarcodeUniquenessMixin(models.AbstractModel):
+    _name = 'barcode.uniqueness.mixin'
+    _description = 'Barcode Uniqueness Mixin'
+
+    def _get_barcode_fname(self):
+        """ Return the name of the field containing the barcode to check for uniqueness. """
+        return ['barcode']
+
+    def _get_models_to_skip(self):
+        """ Return the list of models to skip when checking for duplicate barcodes. """
```
dhha-odoo: [Docstring](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,61 @@
...
+}
+
+
+class BarcodeUniquenessMixin(models.AbstractModel):
+    _name = 'barcode.uniqueness.mixin'
+    _description = 'Barcode Uniqueness Mixin'
+
+    def _get_barcode_fname(self):
+        """ Return the name of the field containing the barcode to check for uniqueness. """
+        return ['barcode']
```
dhha-odoo: No need to **inherit this method** in a different model (**lot and package**); check **self._name in BARCODE_CHECK_MAP** and, based on that, return the **field name** from it.

addons/stock/models/stock_lot.py
```diff
@@ -101,14 +101,17 @@ def _get_next_serial(self, company, product):
                 return self.env['stock.lot'].generate_lot_names(last_serial.name, 2)[1]['lot_name']
         return False
 
+    def _get_barcode_fname(self):
+        return ['name']
+
+    def _get_models_to_skip(self):
+        return ['product.product']
```
dhha-odoo: Image

This still allows adding the **same barcode product as a lot/serial number**, but we also want to constrain it.

addons/stock/models/stock_package_type.py
```diff
@@ -125,6 +126,13 @@ def write(self, vals):
                 self.env['ir.sequence'].browse(list(seq_to_todo_ids)).sudo().write(seq_vals)
         return super().write(vals)
 
+    def _group_barcodes_by_company(self):
+        return [(False, [b for b in self.mapped('barcode') if b])]
+
+    def _get_duplicate_barcode_domain(self, field_name, barcodes, company_id):
+        domain = [(field_name, 'in', barcodes)]
+        return domain
```
dhha-odoo:
```suggestion
        return [(field_name, 'in', barcodes)]
```

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,63 @@
+from odoo import api, models
+from odoo.exceptions import ValidationError
+from odoo.tools import groupby
+
+BARCODE_CHECK_MAP = {
```
dhha-odoo:
```suggestion
BARCODE_UNIQUENESS_MAP = {
```
IMO, better name.

addons/stock/models/barcode_uniqueness_mixin.py
dhha-odoo, line 1: [Order of methods](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions)

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+        :param int|bool company_id: Company ID to check the barcodes against.
+        :raises ValidationError: If a duplicate is found.
+        """
+        models_to_skip = self._get_models_to_skip()
+        for model_name, config in BARCODE_CHECK_MAP.items():
+            if self._name == model_name or model_name in models_to_skip:
+                continue
+            domain = self.env[model_name]._get_duplicate_barcode_domain(config['field'], barcodes_within_company, company_id)
+            if self.env[model_name].search_count(domain, limit=1):
+                raise ValidationError(self.env._("The barcode is already assigned to a %s.", config['label']))
```
dhha-odoo: Add a safeguard at **_check_barcode_uniqueness** to **avoid calling this logic/methods/search_count repeatedly** for every model in **BARCODE_CHECK_MAP**.

For example, when creating a Package Type without a barcode (since the barcode field is not required there), this method gets called for all **7 models in BARCODE_CHECK_MAP**. In such cases, **barcodes_within_company will always be empty/false**, so there is no need to perform barcode uniqueness checks.

So we can **avoid unnecessary computation and reduce the number of queries**.

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+    def _check_duplicated_barcodes(self, barcodes_within_company, company_id):
+        """Ensure barcodes are not already used in other models.
+
+        :param list barcodes_within_company: Barcodes to check for duplicates within the same company.
+        :param int|bool company_id: Company ID to check the barcodes against.
+        :raises ValidationError: If a duplicate is found.
+        """
+        models_to_skip = self._get_models_to_skip()
+        for model_name, config in BARCODE_CHECK_MAP.items():
+            if self._name == model_name or model_name in models_to_skip:
```
dhha-odoo: Add a **skip_model key directly in BARCODE_CHECK_MAP to define the models** that should be skipped, instead of introducing a separate hook for returning skipped models.

This way, **_get_models_to_skip** becomes unnecessary and can be removed.

addons/stock/models/stock_package_type.py
```diff
@@ -125,6 +126,12 @@ def write(self, vals):
                 self.env['ir.sequence'].browse(list(seq_to_todo_ids)).sudo().write(seq_vals)
         return super().write(vals)
 
+    def _group_barcodes_by_company(self):
+        return [(False, [b for b in self.mapped('barcode') if b])]
+
+    def _get_duplicate_barcode_domain(self, field_name, barcodes, company_id):
+        return [(field_name, 'in', barcodes)]
```
dhha-odoo: Remove this; it's unnecessary.

In the main method, you already added a condition to **avoid adding the company domain when company_id is not present**. Since **_group_barcodes_by_company** is already inherited to avoid **grouping barcodes by **company_id****, company_id will not be returned there, so this domain will never be added.

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+            (company_id, [r[field_name] for r in records if r[field_name]])
+            for company_id, records in groupby(self, lambda r: r.company_id.id)
+        ]
+
+    def _get_duplicate_barcode_domain(self, field_name, barcodes, company_id):
+        """Return the domain to check for duplicate barcodes.
+        """
+        domain = [(field_name, 'in', barcodes)]
+        if company_id:
+            domain.append(('company_id', 'in', (False, company_id)))
```
dhha-odoo: Use Domain.

addons/stock/models/stock_package_type.py
```diff
@@ -125,6 +126,12 @@ def write(self, vals):
                 self.env['ir.sequence'].browse(list(seq_to_todo_ids)).sudo().write(seq_vals)
         return super().write(vals)
 
+    def _group_barcodes_by_company(self):
+        return [(False, [b for b in self.mapped('barcode') if b])]
```
dhha-odoo:
```suggestion
        return [(False, list(filter(None, self.mapped('barcode'))))]
```
No for loop, no need to check using an if condition.

addons/stock/models/product.py
```diff
@@ -1325,3 +1335,17 @@ def _adjust_uom_quantities(self, qty, quant_uom):
...
+
+class ProductUom(models.Model):
+    _name = 'product.uom'
+    _inherit = ['product.uom', 'barcode.uniqueness.mixin']
+
+    @api.constrains('barcode')
+    def _check_barcode_uniqueness(self):
+        super()._check_barcode_uniqueness()
+        for company_id, barcodes_within_company in self._group_barcodes_by_company():
+            self._check_duplicated_barcodes(barcodes_within_company, company_id)
```
dhha-odoo, line 1346: I'm not sure why this method is inherited. I checked all use cases, and everything seems to work fine.
Could you double-check, please?

jodh-odoo: This method is inherited to extend the barcode uniqueness validation from the product module while preserving its existing behavior.

The super method handles product-specific barcode checks (products and packagings), and this override adds the generic `barcode_uniqueness_mixin` validation to detect duplicates with the other models defined in `BARCODE_UNIQUENESS_MAP`.

For example, without this extension, using the same barcode for a `package type` and a `product packaging` would be allowed. With this method, such duplicates are detected and prevented.

addons/stock/models/stock_picking.py
```diff
@@ -160,6 +161,11 @@ class StockPickingType(models.Model):
         default=lambda self: self.env.company.picking_policy, required=True,
         help="It specifies goods to be transferred partially or all at once")
 
+    _barcode_company_uniq = models.Constraint(
```
dhha-odoo: Could we follow the naming convention from the [coding guidelines for the constraint method](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions)? According to the documentation, I'd prefer something like **_check_picking_type_barcode_company_unique**. WDYT?

addons/stock/models/product.py
```diff
@@ -674,6 +681,9 @@ def write(self, vals):
             })
         return super().write(vals)
 
+    def _get_models_to_skip(self):
+        return ['product.uom']
+
```
dhha-odoo: Remove dead code

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,55 @@
...
+        :param list barcodes_within_company: Barcodes to check for duplicates within the same company.
+        :param int|bool company_id: Company ID to check the barcodes against.
+        :raises ValidationError: If a duplicate is found.
+        """
+        for model_name, config in BARCODE_UNIQUENESS_MAP.items():
+            if self._name == model_name or self._name == config.get("skip_model"):
+                continue
+            domain = self.env[model_name]._get_duplicate_barcode_domain(config['field'], barcodes_within_company, company_id)
+            if self.env[model_name].search_count(domain, limit=1):
+                raise ValidationError(self.env._("The barcode is already assigned to a %s.", config['label']))
```
dhha-odoo:
```suggestion
                raise ValidationError(self.env._("The barcode is already assigned to a %(label)s.", label=config['label']))
```
[Read](https://github.com/odoo/odoo/pull/224952#discussion_r2581336443)

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def _get_barcode_field_name(self):
+        return [BARCODE_UNIQUENESS_MAP[self._name]['field']]
+
+    def _get_duplicate_barcode_domain(self, field_name, barcodes, company_id):
+        domain = Domain(field_name, 'in', barcodes)
+        if company_id:
+            domain &= Domain('company_id', 'in', (False, company_id))
```
dhha-odoo:
```suggestion
            domain = Domain.AND([domain, Domain('company_id', 'in', (False, company_id))])
```
https://github.com/odoo/odoo/pull/262473#discussion_r3356148386
It's more explicit and readable.

addons/stock/models/barcode_uniqueness_mixin.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def _check_duplicated_barcodes(self, barcodes_within_company, company_id):
+        """Ensure barcodes are not already used in other models.
+
+        :param list barcodes_within_company: Barcodes to check for duplicates within the same company.
+        :param int|bool company_id: Company ID to check the barcodes against.
+        :raises ValidationError: If a duplicate is found.
+        """
+        for model_name, config in BARCODE_UNIQUENESS_MAP.items():
+            if self._name == model_name or self._name == config.get("skip_model"):
```
dhha-odoo:
```suggestion
            if self._name in (model_name, config.get('skip_model')):
```

PR #247198, opened 2026-02-04
[FIX] html_builder: transfer builder range props to builder number input

romo-odoo:
The props "withNumberInput" for builder range added a number input next to the slider, to fine tune the value. However, the min / max / step props were not given. Therefore if the user pressed arrow up in the number input, the value would increase by 1, instead of the value of step given to the builder range input.

This commit fixes the issue by giving the correct props to the builder number input.


FrancoisGe:
@robodoo r+

PR #247201, opened 2026-02-04
Master mrp workorder revamp ajf

ajf-odoo:
1) Adapt planning related views
2) Allow to plan a single workorder
3) Add Planning per Employee

Note that now a production is considered as planned when all of its workorders are planned.

task: 5259535

---

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -491,7 +491,7 @@ class MrpProduction(models.Model):
     def _compute_is_planned(self):
         for production in self:
             if production.workorder_ids:
-                production.is_planned = any(wo.date_start and wo.date_finished for wo in production.workorder_ids)
+                production.is_planned = all(wo.date_start and wo.date_finished for wo in production.workorder_ids if wo.state != 'cancel')
             else:
                 production.is_planned = False
 
@@ -931,6 +931,8 @@ class MrpProduction(models.Model):
     def _change_producing(self):
         if self.state in ['draft', 'cancel'] or (self.state == 'done' and self.is_locked):
             return False
+        if self.state not in ('progress', 'done'):
+            self.state = 'progress'
         if self.product_tracking == 'serial' and self.lot_producing_ids:
             self.qty_producing = len(self.lot_producing_ids)
         productions_bypass_qty_producting = self.filtered(lambda p: p.lot_producing_ids and p.product_tracking == 'lot' and p._origin and p._origin.qty_producing == p.qty_producing)
@@ -948,6 +950,11 @@ class MrpProduction(models.Model):
             res = self._can_produce_serial_numbers()
             if res is not True:
                 return res
+
+    @api.onchange('state')
+    def _onchange_date_start(self):
+        if self.state == 'progress' and self._origin.state == 'confirmed':
+            self.date_start = fields.Datetime.now()
 
     def _can_produce_serial_numbers(self, sns=None):
         self.ensure_one()
@@ -1021,6 +1028,9 @@ class MrpProduction(models.Model):
                 if picking_type != production.picking_type_id:
                     production.name = picking_type.sequence_id.next_by_id()
                     moves_to_reassign |= production.move_raw_ids
+
+        if vals.get('state') == 'progress' and 'date_start' not in vals:
+            vals['date_start'] = fields.Datetime.now()
 
         res = super(MrpProduction, self).write(vals)
 
@@ -1704,7 +1714,7 @@ class MrpProduction(models.Model):
         for workorder in final_workorders:
             workorder._plan_workorder(replan)
 
-        workorders = self.workorder_ids.filtered(lambda w: w.state not in ['done', 'cancel'])
+        workorders = self.workorder_ids.filtered(lambda w: w.state not in ['cancel'])
         if not workorders:
             return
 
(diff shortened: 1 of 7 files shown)


addons/mrp/models/mrp_production.py
```diff
@@ -490,7 +490,7 @@ def _compute_duration(self):
     def _compute_is_planned(self):
         for production in self:
             if production.workorder_ids:
-                production.is_planned = any(wo.date_start and wo.date_finished for wo in production.workorder_ids)
+                production.is_planned = all(wo.date_start and wo.date_finished and wo.state != 'cancel' for wo in production.workorder_ids)
```
amoyaux: I.M.O. We should just skip the cancel workorder instead

addons/mrp/models/mrp_production.py
```diff
@@ -920,6 +920,8 @@ def _compute_serial_numbers_count(self):
     def _change_producing(self):
         if self.state in ['draft', 'cancel'] or (self.state == 'done' and self.is_locked):
             return False
+        if self.state == 'progress' and self._origin.state == 'confirmed':
+            self.date_start = fields.Datetime.now()
```
amoyaux: Very strange. I don't get why you should do that here. I would keep it, in a specific code like write and go for a simple logic -> If you write the progress state, you write the date start

amoyaux: I would do a compute on date_start base on state: (computed, stored, readonly false)
```python
@api.depends('state')
def _compute_date_start(self):
    for production in self:
        if production.state == 'progress' and production._origin.state == 'confirmed':
            production.date_start = fields.Datetime.now()
```

addons/mrp/models/mrp_production.py
```diff
@@ -3028,7 +3030,10 @@ def action_open_label_type(self):
     def action_start(self):
         self.ensure_one()
         if self.state == "confirmed":
-            self.state = "progress"
+            self.write({
+                'state': 'progress',
+                'date_start': fields.Datetime.now(),
+            })
```
amoyaux: I would move this section in the write of the production itself

addons/mrp/models/mrp_workorder.py
```diff
@@ -885,6 +889,31 @@ def action_mark_as_done(self):
...
+            if self.env.context.get('workcenter_to_plan_on'):
+                workcenter_to_plan_on = self.env['mrp.workcenter'].browse(self.env.context.get('workcenter_to_plan_on'))
+            workorder._internal_plan_workorder(date_to_plan_on, workcenter_to_plan_on)
+            if not workorder.blocked_by_workorder_ids:
+                initial_workorders = workorder.production_id.workorder_ids.filtered(lambda wo: not wo.blocked_by_workorder_ids)
+                workorder.production_id.with_context(force_date=True).write({
+                    'date_start': min((wo.leave_id.date_from for wo in initial_workorders if wo.leave_id), default=None),
+                })
+
+    def action_unplan(self):
```
amoyaux, line 913: You should call this function in `_unplan_workorders`

addons/mrp/models/mrp_workorder.py
```diff
@@ -573,8 +573,12 @@ def _plan_workorder(self, replan=False):
         # Plan only suitable workorders
         if not replan:
             return
+        self._internal_plan_workorder(date_start, self.workcenter_id | self.workcenter_id.alternative_workcenter_ids)
+
+    def _internal_plan_workorder(self, date_start, workcenters):
```
amoyaux, line 581: Rename + optional params + multi
```python
def _action_plan(self, forced_date_start=None, forced_workcenters=None, replan_predecessors=False):
    ...
    predecessors = self.blocked_by_workorder_ids
    predecessors._action_plan(forced_date, forced_workcenters, True)
    for workorder in predecessors:
        ...adapt the start date of workorder in self...

    self.action_unplan()
    for workorder in self:
        date_start = forced_date_start or predecessor date or max(self.production_id.date_start, datetime.now())
        workcenters = forced_workcenters or (workorder.production.workcenter_id | workorder.production.workcenter_id.alternative_workcenter_ids)

    ... the code of _internal_plan_workorder
    ... update the production date start if needed
```

addons/mrp/models/mrp_workorder.py
amoyaux, line 561: remove and use _action_plan

addons/mrp/models/mrp_workorder.py
amoyaux: new commit

addons/mrp/models/mrp_workorder.py
```diff
@@ -885,6 +889,31 @@ def action_mark_as_done(self):
...
+                date_to_plan_on = fields.Datetime.from_string(self.env.context.get('date_to_plan_on'))
+            workcenter_to_plan_on = workorder.workcenter_id
+            if self.env.context.get('workcenter_to_plan_on'):
+                workcenter_to_plan_on = self.env['mrp.workcenter'].browse(self.env.context.get('workcenter_to_plan_on'))
+            workorder._internal_plan_workorder(date_to_plan_on, workcenter_to_plan_on)
+            if not workorder.blocked_by_workorder_ids:
+                initial_workorders = workorder.production_id.workorder_ids.filtered(lambda wo: not wo.blocked_by_workorder_ids)
+                workorder.production_id.with_context(force_date=True).write({
+                    'date_start': min((wo.leave_id.date_from for wo in initial_workorders if wo.leave_id), default=None),
+                })
```
amoyaux, line 911: call _action_plan with adapted kwargs

addons/mrp/models/mrp_production.py
Whenrow: I would rather remove explcit write on state and adapt the compute_state instead

```suggestion
            self.date_start = fields.Datetime.now()
```
and set 'progress' in the compute is `date_start != False`

PR #247203, opened 2026-02-04
[FIX] website_slides: wait for portal chatter in tours

Dj0ulo:
### Before commit
Some tours did not wait for the portal chatter to load when needed:
- `portal_chatter_bundle`: the `run` function creates a promise but the tour finishes before it resolves.
- `fullscreen_slide_text_highlights`: `PortalChatterService` adds an element in the DOM while the selection is updated in the `selectText` tour step. This caused the following warning to show:
```
should not have any "characterData", "remove" or "add" mutations in
current step when you update the selection
```

### Fix
Properly wait for `portalChatterReady` to resolve before proceeding with the rest of the tour steps.

(Also fixed the typo in the filename `slide_portal_chatter_bundle.js`)

runbot-229803


seb-odoo (approved):
@robodoo delegate+

PR #247204, opened 2026-02-04
[IMP] core: small clean of sql_db pool

kmagusiak:
- Remove "leak" parameter
- Do not put back closed connections to the pool
- Reset the connection when putting it back
- SQL pool implementation with 2 sets

---

### DIFF
--- a/odoo/sql_db.py
+++ b/odoo/sql_db.py
@@ -646,12 +646,6 @@ class ConnectionPool:
 
         for i, cnx in enumerate(self._connections):
             if not cnx._pool_in_use and self._dsn_equals(cnx.dsn, connection_info):
-                try:
-                    cnx.reset()
-                except psycopg2.OperationalError:
-                    self._debug('Cannot reset connection at index %d: %r', i, cnx.dsn)
-                    cnx.close()
-                    continue
                 cnx._pool_in_use = True
                 self._debug('Borrow existing connection to %r at index %d', cnx.dsn, i)
 
@@ -693,15 +687,20 @@ class ConnectionPool:
         except ValueError:
             raise PoolError('This connection does not belong to the pool')
 
-        if keep_in_pool:
+        if keep_in_pool and not connection.closed:
             # Release the connection and record the last time used
-            connection._pool_in_use = False
-            connection._pool_last_used = time.time()
             self._debug('Put connection to %r in pool', connection.dsn)
-        else:
-            cnx = self._connections.pop(index)
-            self._debug('Forgot connection to %r', cnx.dsn)
-            cnx.close()
+            try:
+                connection.reset()
+            except psycopg2.OperationalError as e:
+                self._debug('Cannot reset connection: %r (%s)', connection.dsn, e)
+            else:
+                connection._pool_in_use = False
+                connection._pool_last_used = time.time()
+                return
+        self._connections.pop(index)
+        self._debug('Forget connection to %r', connection.dsn)
+        connection.close()
 
     @locked
     def close_all(self, dsn: dict | str | None = None):
--- a/odoo/sql_db.py
+++ b/odoo/sql_db.py
@@ -533,8 +533,13 @@ class Cursor(BaseCursor):
             del self._obj
 
             # Put the connection back to the pool
-            chosen_template = tools.config['db_template']
-            keep_in_pool = self.dbname not in ('template0', 'template1', config['db_system'], chosen_template)
+            # Forget already closed connections and system-related databases
+            keep_in_pool = not self._cnx.closed and self.dbname not in (
+                'template0', 'template1',
+                # keep open if one of preloaded databases
+                config['db_system'] if config['db_system'] not in config['db_name'] else '',
+                config['db_template'],
+            )
             self.__pool.give_back(self._cnx, keep_in_pool=keep_in_pool)
 
     def commit(self) -> None:
--- a/odoo/sql_db.py
+++ b/odoo/sql_db.py
@@ -1,6 +1,5 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-
 """
 The PostgreSQL connector is a connectivity layer between the OpenERP code and
 the database, *not* a database abstraction toolkit. Database abstraction is what
@@ -21,7 +20,7 @@ from datetime import datetime, timedelta
 from inspect import currentframe
 
 import psycopg2
-import psycopg2.errorcodes
+import psycopg2.errorcodes  # noqa: F401
 import psycopg2.errors
 import psycopg2.extensions
 import psycopg2.extras
@@ -39,8 +38,6 @@ from .tools.func import frame_codeinfo, locked
 from .tools.misc import Callbacks, real_time
 
 if typing.TYPE_CHECKING:
-    from collections.abc import Iterable, Iterator
-
     from odoo.orm.environments import Transaction
 
     # when type checking, the BaseCursor exposes methods of the psycopg cursor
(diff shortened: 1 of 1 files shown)


odoo/sql_db.py
```diff
@@ -542,8 +542,13 @@ def _close(self) -> None:
 
         self._closed = True
 
-        chosen_template = tools.config['db_template']
-        keep_in_pool = self.dbname not in ('template0', 'template1', config['db_system'], chosen_template)
+        # Forget already closed connections an system-related databases
```
HydrionBurst:
```suggestion
        # Forget already closed connections and system-related databases
```

HydrionBurst (commented):
As for the commit [PERF] core: ConnectionPool with 2 sets
Theoretically, using 2 sets can be faster. But I am not sure if the complexity is worthy.
I don't expect Cursor borrow will become the bottleneck of any code. Using simple and stupid one flag is not so bad to me. 🤔

odoo/sql_db.py
```diff
@@ -390,7 +399,8 @@ def __del__(self):
             else:
                 msg += "Please enable sql debugging to trace the caller."
             _logger.warning(msg)
-            self._close(True)
+            self._cnx.close()
+            self._close()
```
HydrionBurst: Not really sure, I feel we can remove the `self._close()` and only keep `self._cnx.close()` 🤔 
Because
1. it can promise the closed cnx will be garbage collected in next borrow
2. most resources are already recursively recycled by garbage collection
3. strictly speaking, `__del__` can be called any time, but `self._close` may not be a safe function to be called any time.
For example, what if it during iterating the `connection_pool._free_connections` the garbage colelction happens and `append` the connection to it.

kmagusiak: Indeed. Since `__del__` is called, there is no references to the cursor (the pool is already or being GC'ed too).

odoo/sql_db.py
HydrionBurst: NIT 
```suggestion
        if not self.closed:
```

odoo/sql_db.py
```diff
@@ -626,49 +628,41 @@ def borrow(self, connection_info: dict) -> PsycoConnection:
...
+        now = time.time()
+        check_all = self._check_free_at < now
+        self._check_free_at = now + MAX_IDLE_TIMEOUT / 10
+        close_used_before = now - MAX_IDLE_TIMEOUT
+        selected_cnx = None
+        for i, cnx in tools.reverse_enumerate(self._free_connections):
+            if not cnx.closed and cnx._pool_last_used < close_used_before:
                 self._debug('Close connection at index %d: %r', i, cnx.dsn)
                 cnx.close()
             if cnx.closed:
```
HydrionBurst, line 646: As a side effect of our [change](https://github.com/odoo/odoo/pull/247204#discussion_r2774759220) to `__del__`, unlike the old code, we popped all closed `self._connections`, here we only check `_free_connections`. So the cnx closed by `__del__` may still be remained in the `_used_connections`.
Also if a cnx is closed because of a database problem, the `Cursor.close`will bypass `_close`(`ConnectionPool.give_back`) and its closed cnx will be remained in the `_used_connections` forever

kmagusiak: Fixed, instead of closing it in `__del__`, I give it back to the pool so it can close the connection.

HydrionBurst: did you check 

> Also if a cnx is closed because of a database problem, the Cursor.closewill bypass _close(ConnectionPool.give_back) and its closed cnx will be remained in the _used_connections forever

I think if the connection is closed by the database side, e.g. (readonly server disconnected). In the next query the `cr._cnx.closed` is truthy and then `cr.close()` will not call `cr._close()` and `ConnectionPool.give_back`. And `__del__` will also skip all logics sine `cr.closed` is Falsy. As a result, the cursor will be remained in the `_used_connections` forever, and we have to restart the server.

kmagusiak: Added in the first commit. To give back `_cnx`, we should not look at its state.

odoo/sql_db.py
```diff
@@ -390,7 +387,9 @@ def __del__(self):
             else:
                 msg += "Please enable sql debugging to trace the caller."
             _logger.warning(msg)
-            self._close(True)
+            # Just close the raw connection as other all environments are
+            # (being) collected at this time.
+            self.__pool.give_back(self._cnx, keep_in_pool=False)
```
HydrionBurst, line 392: Maybe I am overthinking, but theoretically, I think calling `give_back` which modify the shared `_used_conections` in the `__del__` is risky. Because it might be called anytime. For this PR, 
I think it is safe, so I am OK with it. Because we never iterate through the `_used_conections`

For me, I still prefer we set a flag attribute in `__del__` which is simple, local and 100% no side effect. And then we always garbage collect (including dropping them from `_used_connections`) them during `borrow`.

odoo/sql_db.py
```diff
@@ -503,9 +502,9 @@ def _enable_logging(self):
 
     def close(self) -> None:
         if not self.closed:
```
HydrionBurst: Just my personal opinion.
Maybe we can use `self._closed` here?

So, the `__del__` only handles coding error when a cursor is garbage collected without explicit `close` before.
And if the `_cnx` is closed because of the external reason, we still handle the logic in the `_close` explicitly

HydrionBurst (approved):

odoo/sql_db.py
kmagusiak: The change here is simply: deleting the Cursor means we need to return the referenced connection to the pool.

Setting a flag means that we don't know the state of the pool, which we do since the cursor has a reference to the pool so it cannot be collected before. Having a flag means that we always need to check it.
What could go wrong here? If we return the connection twice without closing it... but returning it is only done in the `_close` method.

odoo/sql_db.py
kmagusiak: `_close` needs an open connection to execute ROLLBACK. I tried to refactor so that it is cleaner.

HydrionBurst: > _close needs an open connection to execute ROLLBACK

That's another thing I feel theoretically risky.
According to my observations, the `cr._cnx.closed` is lazily checked. Which means it is possible, when the connection has been closed, but the `cr._cnx.closed == 0`, the value will be updated in the next sql call. As a result, the `cr.rollback()` may raise `psycopg2.OperationalError: server closed the connection unexpectedly` and the `give_back` is not executed.
The `BaseCursor.__exit__` only catch exception for `self.commit()`, so if there is an error raised, and the connection is closed by accident, the `give_back` won't be executed 😢 
But I think it is only a theoretical problem since in most business code, we always `rollback` explicitly when needed.

kmagusiak: I've put the code in a finally block. Then whatever happens, we give it back.

odoo/sql_db.py
```diff
@@ -390,7 +390,9 @@ def __del__(self):
             else:
                 msg += "Please enable sql debugging to trace the caller."
             _logger.warning(msg)
```
rco-odoo: Only log a warning when the connection is not closed.
(Connections are closed when exiting the process.)

odoo/sql_db.py
```diff
@@ -604,17 +600,19 @@ class ConnectionPool:
         The connections are *not* automatically closed. Only a close_db()
         can trigger that.
     """
-    _connections: list[PsycoConnection]
 
     def __init__(self, maxconn: int = 64, readonly: bool = False):
-        self._connections = []
+        # most recently used connections are at the end of the queue
+        self._free_connections: list[PsycoConnection] = []
+        self._used_connections: dict[PsycoConnection, None] = {}
```
rco-odoo: Maybe use `OrderedSet` instead.

rco-odoo (approved):
@robodoo rebase-ff r+

PR #247210, opened 2026-02-04
[FIX] iot_drivers: IPP Everywhere when no ppd available

loouis-t:
When no `ppd` is available for the cups discovered printer, we now use the IPP Everywhere ppd.

We only set it for dnssd printers as printer that only implement socket/lpd could be too old and not compatible with IPP Everywhere.

### DIFF
--- a/addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
+++ b/addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py
@@ -53,15 +53,12 @@ class PrinterInterface(Interface):
         for path, device in devices.items():
             identifier, device = self.process_device(path, device)
 
-            url_is_supported = any(protocol in device["url"] for protocol in ['dnssd', 'lpd', 'socket'])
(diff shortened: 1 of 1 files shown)


Ysoroko (changes requested):
Tested in the office, the result is printer detection now takes a long time + every detected printer prints a receipt.
For a lot of printers found in the office the following error is shown:
```
2026-02-05 06:38:10,671 5733 ERROR ? odoo.addons.iot_drivers.iot_handlers.interfaces.printer_interface_L: Failed to add printer 'dnssdEPSON%20TM-m30III-A1E3A0_pdl-datastream_tcplocal' 
  File "/home/pi/odoo/addons/iot_drivers/iot_handlers/interfaces/printer_interface_L.py", line 248, in set_up_printer_in_cups
    self.conn.addPrinter(name=device['identifier'], device=url, **ppdname_argument)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cups.IPPError: (1280, 'Success')
```

Ysoroko (approved):
robodoo r+

PR #247216, opened 2026-02-04
[IMP] l10n_in: remove restriction for accessing tax returns

unka-odoo:
BEFORE:
- Before this commit, when we disable the gst e-filing feature from the configuration and try to access the tax return view, we are getting blocked by the redirect warning, which suggests enabling the gst e-filing feature from the configuration.
- Which is not desirable, as there might be some returns that are not related to gst e-filing, which should be accessible by the user.

AFTER:
- After this commit, removed the RedirectWarning when accessing the tax return view with gst e-filing feature disabled. So now the user can access tax returns without enabling gst e-filing feature.
- At the time of setting the fiscal year(generating/refreshing returns automatically), the GSTR returns will not be created.
- And at the time of manual GSTR return creation, we are raising UserError to instruct the user about enabling the gst e-filing feature.

Related Ent PR: https://github.com/odoo/enterprise/pull/105983

Task-5486586

### DIFF
--- a/addons/account/models/company.py
+++ b/addons/account/models/company.py
@@ -1138,6 +1138,7 @@ class ResCompany(models.Model):
             self.env['ir.default'].set('product.category', 'property_account_expense_categ_id', company.expense_account_id.id, company_id=company.id)
             self.env['ir.default'].set('product.category', 'property_account_income_categ_id', company.income_account_id.id, company_id=company.id)
 
+    # Deprecated, removed in master.
     def _check_tax_return_configuration(self):
         """
         To override in localizations to check if the company is properly configured for tax returns.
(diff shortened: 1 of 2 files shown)


jco-odoo:
@unka-odoo  Do not remove it in stable, but as l10n_in is the only one, it might make sense to do it in master.  It seems indeed more logical at the creation of the tax return itself.

jco-odoo:
@unka-odoo  Do not remove it in stable, but as l10n_in is the only one, it might make sense to do it in master.  It seems indeed more logical at the creation of the tax return itself.

jco-odoo:
@robodoo r+

PR #247220, opened 2026-02-04
[FIX] website_sale: show contact us button for zero-priced variants

farahdawod:
### Steps to reproduce:
- Download eCommerce app.
- Create a product priced at zero.
- Add an attribute and make the FIRST value of the attribute priced at more than 0.
- Check Configuration > Settings > Prevent Sale of Zero Priced Product.
- Open the created product on the website, choose the zero-priced variant.

### Issue:
The 'Contact Us' button doesn't appear when you view the zero-priced variant. This happens because when you initially load the product that has a positive price, the contact us component isn't loaded because it has an if condition that checks if the price is above 0.
https://github.com/odoo/odoo/blob/cda011dc8590773f6c3a26f4ae9d5242a3147024/addons/website_sale/views/templates.xml#L2147-L2154
Consequently, when https://github.com/odoo/odoo/blob/7a39185f83d0daca207c8007512f4700537c7e88/addons/website_sale/static/src/js/variant_mixin.js#L302-L315 tries to load the 'Contact Us' button via the query selector, it gets null as a return value.
The test made for this issue loaded the variant that has the zero-price first, so the 'Contact Us' button was loaded normally.

### Note about the test:
Edit the test so that the positive priced variant is loaded first, which will fail in case that the fix isn't applied, as the 'contact-us-wrapper' component won't be loaded at the start of the test.

opw-5491935

### DIFF
--- a/addons/website_sale/static/tests/tours/website_sale_zero_priced_product.js
+++ b/addons/website_sale/static/tests/tours/website_sale_zero_priced_product.js
@@ -7,6 +7,23 @@ registry.category('web_tour.tours').add('website_sale_contact_us_button', {
             trigger: '.js_attribute_value:has([checked]):contains(red)',
         },
         {
+            content: "Product price should be 20",
+            trigger: '.product_price .oe_currency_value:contains(20.00)',
+        },
+        {
+            content: '"Add to Cart" button should be visibile',
+            trigger: '#add_to_cart_wrap',
+        },
+        {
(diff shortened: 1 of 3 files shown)


lase-odoo (approved):
The fix LGTM!

barracudapps (approved):
LGTM 👍 
Thank you for fixing this!

vchu-odoo (approved):
Thank you for the fix!

@robodoo r+

PR #247223, opened 2026-02-04
[FIX] point_of_sale: prevent rounding invoices when not needed

pedrambiria:
Before this commit, when a rounding method was set in the PoS, but the cash rounding was not enabled, or only enabled for cash payments and the order did not contain any cash payment, the generated invoice would still be linked to the rounding method. This could lead to unwanted rounding on the invoice lines.

opw-5864642

---


adgu-odoo (commented):
robodoo r+

PR #247224, opened 2026-02-04
[FIX] hr: use action_archive in departure wizard

HashemKhaled:
When archiving an employee from the departure wizard, use the action_archive method instead of toggle_active.

Related to odoo/enterprise#100437

task: 5354002


amry-odoo (approved):
LGTM 🚀 🦝

PR #230341, opened 2025-10-07
[IMP] stock{,_delivery,_picking_batch,_fleet}:  improve weight and volume for packages

msho-odoo:
This PR  adds volume and shipping_volume in packages.
It also refactors and fixes some calculations in shipping_weight ans shipping_volume for `stock_picking` and `stock_picking_batch` and `report_picking_cmr`.

This commit:
1- adds computed `volume` and `shipping_volume` to packages.

2- fixes the computation of the picking shipping_volume to be the shipping_volume of the outermost package if it's set, otherwise, it uses the package's type dimensions if set, otherwise, it uses the volumes of the internals of that package and so on recursively.

3- fixes the computation of picking shipping_weight to be accurate by making it recursive as well when you have a chain of packages inside each other. It opts out to the shipping_weight of the package if exists, otherwise it uses the weight of the package's type if exists in addition to the weight of the internals of that package ans so on recursively.

4- fixes the issue of calculating shipping_weight and shipping_volume when pickings are done in both `stock_picking_batch` and `report_picking_cmr` using `stock_package_history` to result in the same shipping_weight and shipping_volume calculations.

5- It also moves weight related fields on `stock_package`, `stock_move` and `stock_picking` from `stock_delivery` module to `stock` module.

Related upgrade PR: https://github.com/odoo/upgrade/pull/8749
Task-5056474

PR #247227, opened 2026-02-04
[FIX] stock: include product in package for shipping weight

josv-odoo:
On delivery slip, the total weight of a package doesn't include the weight of the product in the outermost package.

Steps to reproduce:
-------------------
* Create a Package "Pa" with a package type that has a weight
* Create a product "Po" tracked by quantity
* Add two units of the product Po to the package Pa
* Create Delivery with two units of Po.
* Confirm the Delivery and print the delivery slip -> The total weight is the weight of the package type without including the products inside.

Observation:
-------------
When computing the shipping_weight for the delivery, it will fallback on the package_weight to calculate it weight, https://github.com/odoo/odoo/blob/ca14f1aa21a75398919c1453be19011522bb3b5c/addons/stock/models/stock_picking.py#L893-L906 
It retrieve the weight calculated by _get_weight
https://github.com/odoo/odoo/blob/93fa6d9fff63534cfa9251e21fc82797d8b83468/addons/stock/models/stock_package.py#L435
Except for the outermost package,  where it incorrectly considers only the package type weight and does not include the products weight

opw-5499770

### DIFF
--- a/addons/stock/models/stock_picking.py
+++ b/addons/stock/models/stock_picking.py
@@ -893,18 +893,14 @@ class StockPicking(models.Model):
     def _compute_shipping_weight(self):
         for picking in self:
             # if shipping weight is not assigned => default to calculated product weight
-            packages_weight = picking.move_line_ids.result_package_id.sudo()._get_weight(picking.id)
-
             shipping_weight = picking.weight_bulk
             relevant_packages = picking.move_line_ids.result_package_id.outermost_package_id
-            children_packages_by_pack = relevant_packages._get_all_children_package_dest_ids()[0]
+            packages_weight = relevant_packages.sudo()._get_weight(picking.id)
             for package in relevant_packages:
                 if package.shipping_weight:
                     shipping_weight += package.shipping_weight
                 else:
-                    shipping_weight += package.package_type_id.base_weight
-                    shipping_weight += sum(packages_weight.get(pack, 0) for pack in self.env['stock.package'].browse(children_packages_by_pack.get(package)))
-
+                    shipping_weight += packages_weight.get(package, 0)
             picking.shipping_weight = shipping_weight
 
     def _compute_shipping_volume(self):
--- a/addons/stock/tests/test_packing.py
+++ b/addons/stock/tests/test_packing.py
@@ -2034,6 +2034,48 @@ class TestPackagePropagation(TestPackingCommon):
         self.assertFalse((pallet | container).package_dest_id)
         self.assertFalse((pallet | container).picking_ids)
 
+    def test_package_shipping_weight(self):
+        """
+        Create a 10 kg package containing:
+         - five products weighing 2 kg each
+         - a 5 kg box containing three product weighing 2 kg each.
+        Create a 10 kg package containing:
+         - a 5 kg box containing one product weighing 2 kg.
+        Check that the packages total shipping weight is 31 kg and 17kg respectively.
+        """
+        self.productA.weight = 2
+        self.pack_type_box.base_weight = 5
+        self.pack_type_pallet.base_weight = 10
+        box, box2, pallet, pallet2 = self.env['stock.package'].create([{
+            'package_type_id': pack_type.id,
+        } for pack_type in [self.pack_type_box, self.pack_type_box, self.pack_type_pallet, self.pack_type_pallet]])
+        self.env['stock.quant']._update_available_quantity(self.productA, self.stock_location, 5, package_id=pallet)
+        self.env['stock.quant']._update_available_quantity(self.productA, self.stock_location, 3, package_id=box)
+        self.env['stock.quant']._update_available_quantity(self.productA, self.stock_location, 1, package_id=box2)
+        box.parent_package_id = pallet
+        box2.parent_package_id = pallet2
+        delivery = self.env['stock.picking'].create({
+            'picking_type_id': self.warehouse.out_type_id.id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.customer_location.id,
+        })
+        delivery2 = self.env['stock.picking'].create({
+            'picking_type_id': self.warehouse.out_type_id.id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.customer_location.id,
+        })
+        delivery.action_add_entire_packs(pallet.id)
+        delivery.action_confirm()
+        delivery2.action_add_entire_packs(pallet2.id)
+        delivery2.action_confirm()
+
+        self.assertEqual(delivery.move_line_ids.result_package_id, pallet | box)
+        self.assertEqual(delivery2.move_line_ids.result_package_id, box2)
+        self.assertEqual(delivery.move_line_ids.outermost_result_package_id, pallet)
+        self.assertEqual(delivery2.move_line_ids.outermost_result_package_id, pallet2)
+        self.assertEqual(delivery.shipping_weight, 31)
+        self.assertEqual(delivery2.shipping_weight, 17)
+
     def test_package_removal(self):
         """ Checks that the button 'Remove' in the package view in pickings behaves as expected:
             - Only removes related move/move lines from the picking if it was only added through an entire pack


addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,43 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
...
+            'base_weight': 5,
+        }])
+        boxA = self.env['stock.package'].create([{
+            'name': "box for product A",
+            'package_type_id': sbox_type.id,
+        }])
+
+        # Content of package:
+        # boxA              (base  5kg) -> 5 + 10 = 15kg
+        # └ 5x Product A      (5*2 =  10kg)
```
lase-odoo: I do not really like the usage of special characters in stings specially if they are not improving readability and this display is not necessary imo. It would be much cleaner to keep a simple test description at a start of the test like:
```py
    """
    Create a 5 kg package containing five products weighing 2 kg each.
    Check that the package total shipping weight is 15 kg.
    """
```
And to not say anything in the test.

addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,43 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
         self.assertTrue(move_line_1.result_package_id, 'A package should have been created for the selected move line')
         self.assertFalse(move_line_2.result_package_id, 'The other move line should not be packed')
 
+    def test_package_shipping_weight(self):
```
lase-odoo: Add a description to your tests.

addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,43 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
...
+                    'location_id': self.stock_location.id,
+                    'location_dest_id': self.customer_location.id,
+                    'product_id': self.productA.id,
+                    'product_uom_qty': 5,
+                }),
+            ],
+        })
+        delivery.action_confirm()
+
+        self.assertEqual(delivery.move_line_ids.package_id.name, "box for product A")
```
lase-odoo: It is usually more relevant to assert on the record themself.
```suggestion
        self.assertEqual(delivery.move_line_ids.package_id, boxA)
```

lase-odoo (commented):

Taking a quick look on the change, it looks fine but could you modify the use case of your test for the packA to contain both products and sub-packs with product themself to ensure that the weight of sub-packages is not counted twice ? I t looks like the `_get_weight` method called to generate the `packages_weight` used in your loop is already taking sub-packages into account
https://github.com/odoo/odoo/blob/8ef673d27481d17b83f93892e78b09b085738ca7/addons/stock/models/stock_picking.py#L895-L906
https://github.com/odoo/odoo/blob/8ef673d27481d17b83f93892e78b09b085738ca7/addons/stock/models/stock_package.py#L435-L470
And I am therefore wondering if the line following your change will not add that same sub-pack weight a second time with your change.
https://github.com/odoo/odoo/blob/8ef673d27481d17b83f93892e78b09b085738ca7/addons/stock/models/stock_picking.py#L906

addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,52 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
...
+                    'location_id': self.stock_location.id,
+                    'location_dest_id': self.customer_location.id,
+                    'product_id': self.productA.id,
+                    'product_uom_qty': 8,
+                }),
+            ],
+        })
+        delivery.action_confirm()
+
+        self.assertEqual(delivery.move_line_ids.package_id.ids, [pack.id for pack in (box, sbox)])
```
lase-odoo: Assert on records directly.
```suggestion
        self.assertEqual(delivery.move_line_ids.package_id, box | sbox])
```

addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,52 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
...
 
+    def test_package_shipping_weight(self):
+        """
+        Create a 10 kg package containing:
+         - five products weighing 2 kg each
+         - a 5 kg box containing three product weighing 2 kg each.
+
+        Check that the package total shipping weight is 31 kg.
+        """
+        self.warehouse.delivery_steps = 'ship_only'
```
lase-odoo: why is this relevant ? Why not simply make a ship ?

addons/stock_delivery/tests/test_packing_delivery.py
```diff
@@ -393,6 +393,52 @@ def test_put_in_pack_applies_only_to_selected_move_line(self):
         self.assertTrue(move_line_1.result_package_id, 'A package should have been created for the selected move line')
         self.assertFalse(move_line_2.result_package_id, 'The other move line should not be packed')
 
+    def test_package_shipping_weight(self):
```
lase-odoo: why is this test in `stock_delivery` ? A priori the `shipping_weight` is defined in `stock` and you only cite code from the stock module in your commit explanations.
https://github.com/odoo/odoo/blob/00b042c264c25d8e2ee4b019915a3f76fc04dc39/addons/stock/models/stock_picking.py#L666-L670

addons/stock/tests/test_packing.py
```diff
@@ -2034,6 +2034,50 @@ def test_pack_in_pack_already_packed(self):
...
+        Create a 10 kg package containing:
+         - five products weighing 2 kg each
+         - a 5 kg box containing three product weighing 2 kg each.
+        Check that the package total shipping weight is 31 kg.
+        """
+        self.productA.weight = 2
+        sbox_type, box_type = self.env['stock.package.type'].create([{
+            'name': name,
+            'base_weight': weight,
+        } for (name, weight) in [('Small box"', 5), ('Box', 10)]])
```
clesgow:
```suggestion
        self.pack_type_box.base_weight = 5
        self.pack_type_pallet.base_weight = 10
```
Use records already used in the `setUpClass` when possible

addons/stock/tests/test_packing.py
```diff
@@ -2034,6 +2034,50 @@ def test_pack_in_pack_already_packed(self):
...
+                    'location_id': self.stock_location.id,
+                    'location_dest_id': self.customer_location.id,
+                    'product_id': self.productA.id,
+                    'product_uom_qty': 8,
+                }),
+            ],
+        })
+        delivery.action_confirm()
+
+        self.assertEqual(delivery.move_line_ids.package_id, box | sbox)
```
clesgow:
```suggestion
        self.assertEqual(delivery.move_line_ids.result_package_id, box | sbox)
```
You should rather check the `result_package_id` as we only consider the destination package for shipping weight.

addons/stock/tests/test_packing.py
```diff
@@ -2034,6 +2034,50 @@ def test_pack_in_pack_already_packed(self):
...
+            'move_ids': [
+                Command.create({
+                    'location_id': self.stock_location.id,
+                    'location_dest_id': self.customer_location.id,
+                    'product_id': self.productA.id,
+                    'product_uom_qty': 8,
+                }),
+            ],
+        })
+        delivery.action_confirm()
```
clesgow, line 2068:
```suggestion
        delivery = self.env['stock.picking'].create({
            'picking_type_id': self.warehouse.out_type_id.id,
            'location_id': self.stock_location.id,
            'location_dest_id': self.customer_location.id,
        })
        delivery.action_add_entire_packs(box.id)
```
Nitpick, but for future reference if you want to add the whole content of a pack, you can directly add the package to the picking.

addons/stock/tests/test_packing.py
```diff
@@ -2034,6 +2034,50 @@ def test_pack_in_pack_already_packed(self):
...
+        self.productA.weight = 2
+        sbox_type, box_type = self.env['stock.package.type'].create([{
+            'name': name,
+            'base_weight': weight,
+        } for (name, weight) in [('Small box"', 5), ('Box', 10)]])
+        sbox, box = self.env['stock.package'].create([{
+            'package_type_id': pack_type.id,
+        } for pack_type in [sbox_type, box_type]])
+
+        self.assertEqual(box.weight, 10)
```
clesgow: `weight` is defined on packages in `stock_delivery`, not merely `stock`.
This should change in `master` through #230341, but it's not the case in `19.0`.

To avoid issues like this, I'd advise you to always try to run the test once in a "minimal setup" (e.g. only stock installed) to see if it can properly run without issues.

addons/stock/models/stock_picking.py
```diff
@@ -897,14 +897,11 @@ def _compute_shipping_weight(self):
...
             relevant_packages = picking.move_line_ids.result_package_id.outermost_package_id
-            children_packages_by_pack = relevant_packages._get_all_children_package_dest_ids()[0]
             for package in relevant_packages:
                 if package.shipping_weight:
                     shipping_weight += package.shipping_weight
                 else:
-                    shipping_weight += package.package_type_id.base_weight
-                    shipping_weight += sum(packages_weight.get(pack, 0) for pack in self.env['stock.package'].browse(children_packages_by_pack.get(package)))
-
+                    shipping_weight += packages_weight.get(package, 0)
```
clesgow, line 903: This won't work if your `outermost_package_id` only contains packages and no products (e.g. a pallet only containing boxes on top).

I guess you'd need to use `_get_weight()` on the `relevant_packages` instead, since it *should* handle the contained packages correctly.

addons/stock/models/stock_picking.py
```diff
@@ -893,18 +893,15 @@ def _compute_bulk_weight(self):
     def _compute_shipping_weight(self):
         for picking in self:
             # if shipping weight is not assigned => default to calculated product weight
-            packages_weight = picking.move_line_ids.result_package_id.sudo()._get_weight(picking.id)
+            packages_weight = picking.move_line_ids.result_package_id.outermost_package_id.sudo()._get_weight(picking.id)
 
             shipping_weight = picking.weight_bulk
             relevant_packages = picking.move_line_ids.result_package_id.outermost_package_id
```
clesgow:
```suggestion
            relevant_packages = picking.move_line_ids.result_package_id.outermost_package_id
            packages_weight = relevant_packages.sudo()._get_weight(picking.id)
            shipping_weight = picking.weight_bulk
```

PR #247228, opened 2026-02-04
[FIX] resource : fix leave duration display

adak-ahmed:
**To reproduce:**
1- Adjust working schedule of an employee to have no work on Wednesday afternoon.
2- Create a time off type that allows half day.
3-Take one week time off, the shown leave duration should be 4.5 days, not 5.

**FIX:**
- Removing unnecessary conditions that leads to wrong calculations.

Task: #5909431


Mahmoudk3m (approved):
Thanks for your work, LGTM

PR #247232, opened 2026-02-04
[FIX] base: match server actions visual sort with execution

jeromebelpaire:
Currently, the server actions Kanban view implicitly sorts records by sequence and then by ID. This behavior is triggered automatically by the web client when a field with the 'handle' widget is present and no default order is specified in the view.

However, the backend execution logic for multi-actions (ir.actions.server) processes actions based on 'sequence' and then 'name' (alphabetical).

This creates a discrepancy when multiple actions share the same sequence number: the user sees them sorted by creation order (ID) in the interface, but they are executed alphabetically. This leads to confusion as the visual hierarchy does not accurately represent the flow of operations.

This commit ensures the view aligns with the backend model's sorting logic, so that the visual order in the frontend matches the actual execution order.

Task-5075610


std-odoo (approved):
@robodoo r+

PR #247235, opened 2026-02-04
[ADD] l10n_latam_sale: Show LATAM identification types in sale reports

mobar-odoo:
Problem:
Latin American (LATAM) countries have many identification types (Tax ID, Passport, etc.). Each LATAM partners can be identified by any of the types. However, when printing a sales report (a PDF Quote of a sales order), the VAT label of the company is showing instead of the chosen identification type of the customer. For example, CUIT (VAT label for Argentina) would show instead of Passport for a customer who is identified by their passport number.

Cause:
The generic sales reports do not take into consideration that LATAM partners can be identified by types other than their tax ID.

Solution:
Create a new app, specifically for LATAM partners, to override the sales report generated. It overrides the part with the customer's details to show the correct identification type for LATAM partners.

Steps to reproduce:
- Choose (or create) a Latin American Company from your companies.
- Create a new partner with passprt identification type.
- Create a new sales quotation for that customer.
- Confirm the quotation.
- Print a PDF quote of the sales order
- See how the partner is identified by the country's VAT label and not the passport. (Although the number shown is the correct passport number, but the label is incorrect)

opw-5879261

### DIFF
--- a/.weblate.json
+++ b/.weblate.json
@@ -2550,6 +2550,12 @@
                 "language_regex": "^(es_419|pt_BR)$"
             },
             {
+                "name": "l10n_latam_sale",
+                "filemask": "addons/l10n_latam_sale/i18n/*.po",
+                "new_base": "addons/l10n_latam_sale/i18n/l10n_latam_sale.pot",
+                "language_regex": "^(es_419|pt_BR)$"
+            },
+            {
                 "name": "l10n_lb_account",
                 "filemask": "addons/l10n_lb_account/i18n/*.po",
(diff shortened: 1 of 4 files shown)


agr-odoo (changes requested):
Hello,
Adding a new module for 3 lines of view is not great, auto-install would not even work unless the user trigger the install process.

In `account` report counterpart I see we have the same problem with the vat label, but it has been addressed in each localization. https://github.com/odoo/odoo/blob/5d2e4b0f3fa8cdf2a977db05c7c2565b06ca2513/addons/l10n_ar/views/report_invoice.xml#L204

We might need to fix this case by case

PR #247241, opened 2026-02-04
[FIX] website_event_track: avoid falsy ids

reth-odoo:
[1] added asserts preventing the use of falsy ids when browsing.

To compute "wishlist_visitor_ids" we did not take into account that some "track visitors" don't have a visitor set. Leading to a traceback when opening some tracks.

[1]: 4290724a4c8c57fba4f4d3d688d38f65dadcc38f

task-5911184

### DIFF
--- a/addons/website_event_track/models/event_track.py
+++ b/addons/website_event_track/models/event_track.py
@@ -365,15 +365,20 @@ class EventTrack(models.Model):
 
     @api.depends('event_track_visitor_ids.visitor_id', 'event_track_visitor_ids.is_wishlisted')
     def _compute_wishlist_visitor_ids(self):
+        """Compute the number of track visitors, and the associated website visitors.
+
+        Note that there may be more track visitors than visitors, as some may be linked to partners.
+        """
         results = self.env['event.track.visitor']._read_group(
             [('track_id', 'in', self.ids), ('is_wishlisted', '=', True)],
             ['track_id'],
-            ['visitor_id:array_agg'],
+            ['visitor_id:recordset', '__count'],
         )
-        visitor_ids_map = {track.id: visitor_ids for track, visitor_ids in results}
+        track_visitor_data = {track.id: (visitors, visitor_count) for track, visitors, visitor_count in results}
         for track in self:
-            track.wishlist_visitor_ids = visitor_ids_map.get(track.id, [])
-            track.wishlist_visitor_count = len(visitor_ids_map.get(track.id, []))
+            visitors, visitor_count = track_visitor_data.get(track.id, (self.env['website.visitor'], 0))
+            track.wishlist_visitor_ids = visitors
+            track.wishlist_visitor_count = visitor_count
 
     def _search_wishlist_visitor_ids(self, operator, operand):
         if operator in ('not in', 'not any'):


addons/website_event_track/models/event_track.py
```diff
@@ -365,15 +365,17 @@ def _compute_is_reminder_on(self):
...
         results = self.env['event.track.visitor']._read_group(
             [('track_id', 'in', self.ids), ('is_wishlisted', '=', True)],
             ['track_id'],
-            ['visitor_id:array_agg'],
+            ['visitor_id:recordset', '__count'],
         )
-        visitor_ids_map = {track.id: visitor_ids for track, visitor_ids in results}
-        for track in self:
-            track.wishlist_visitor_ids = visitor_ids_map.get(track.id, [])
-            track.wishlist_visitor_count = len(visitor_ids_map.get(track.id, []))
```
Odoonan, line 376: Hi @reth-odoo ,

So basically, we may have some arrays like `[ v1, v2, False, False ]` in results? And try setting Falsy ids to the field ? :)

Does the `__count` account for the visitorless entries? 🤔 

Thanks

reth-odoo: @Odoonan :wave: 

Yes exactly, as we just aggregate all rows and some of these rows can have a NULL "visitor_id" column, we also aggregate "False" values.

The :recordset aggregator is written with that in mind, so we don't need to implement the logic for "valid ids only" everywhere ourselves.

__count will count all matched rows, to not count those columns we would need to use ~~having=~~`domain=[...('visitor_id', '!=', False)]`. I opted to keep the previous behavior of counting "all people who wishlisted this track", even if they don't have a visitor, as it seems usually more relevant to have a (more) accurate count of how many people wishlisted it rather than exactly how many visitors you will see in the list (anyways probably very few people ever check the list, it requires admin access and is only useful for broad trends).

addons/website_event_track/models/event_track.py
```diff
@@ -365,15 +365,17 @@ def _compute_is_reminder_on(self):
 
     @api.depends('event_track_visitor_ids.visitor_id', 'event_track_visitor_ids.is_wishlisted')
     def _compute_wishlist_visitor_ids(self):
+        self.wishlist_visitor_ids = []
+        self.wishlist_visitor_count = 0
```
Odoonan: On implementation: it felt fine seeing this initialization at first sight but searching in the code, it seems we seldom do that in computes? (maybe I missed it) Could this trigger operations and we want a single allocation? 

suggestion: we can maybe keep something closer to what we had?
```python
track_visitor_data = {track.id: (visitors, visitor_count) for track, visitors, visitor_count in results}
```

so that we can keep:
```python
for track in self:
    visitors, visitor_count = track_visitor_data.get(track.id, ([], 0))
    track.wishlist_visitor_ids = visitors
    track.wishlist_visitor_count = visitor_count
```

reth-odoo: Normally writing to the computed field inside a compute does not trigger anything, it just writes to cache and then they check if it was modified after the compute (this is done via the env.protecting thingy I believe)

I pretty much always use this pattern for _read_group computes. Not against doing it that way though, there's very little performance gain either way.

Odoonan: Ok! Leaving the comment open if reviewer wants to compare options :+1:

tde-banana-odoo: Indeed for computes, nothing is sent ... so seems good to me. I suppose. Anyway don't push on friday they said.

PR #247245, opened 2026-02-04
[FIX] mail: restore formatting of full composer content

alexkuhn:
Before this commit, when using the full composer and accidentally closing it with some content, the composer content was recovered on the basic composer and formatting was lost.

This happens because restore of content only works in the basic composer, and basic composer does not support rich HTML like the full composer.

This commit adds a new UX/UI to restore formatting when accidentally leaving the full composer:

When some content has been restored from full composer, opening the chatter composer momentarily disables everything but the full composer button, in addition to show a popover suggesting the user to decide to either continue with Full Composer and restore formatting, or restore content in the small composer without the formatting.

Most of the time people want to restore formatting from the full composer, but at the same time the full composer can be a frustrating experience that may incite to just continue with the more reliable small composer. This new popover support both use-cases.

Task-5910961

Screenshot 2026-02-26 at 17 28 06


addons/mail/static/src/core/common/action.js
```diff
@@ -7,6 +7,7 @@ import { Reactive } from "@web/core/utils/reactive";
 export const ACTION_TAGS = Object.freeze({
     DANGER: "DANGER",
     SUCCESS: "SUCCESS",
+    PRIMARY: "PRIMARY",
```
alexkuhn, line 10: Backport of https://github.com/odoo/odoo/pull/247648

addons/mail/static/tests/tours/mail_composer_test_tour.js
```diff
@@ -232,6 +235,7 @@ registry.category("web_tour.tours").add("mail/static/tests/tours/mail_composer_t
                 if ((bodyContent.match(/--\nErnest/g) || []).length !== 0) {
                     console.error("The composer should not contain the user's signature.");
                 }
+                window.aku = queryFirst;
```
ThanhDodeurOdoo:
```suggestion

```

addons/mail/static/src/core/common/composer_actions.js
```diff
@@ -131,11 +131,19 @@ registerComposerAction("open-full-composer", {
         composer.targetThread &&
         composer.targetThread.model !== "discuss.channel" &&
         !owner.env.inFrontendPortalChatter,
```
ThanhDodeurOdoo, line 133: would it be possible that a backup gets saved  in the localstorage with the `FromFullComposer` flag, then someone opens that same chatter from a portal view and sees a disabled text area, but not this button because of the condition? which would make the ui stuck

addons/mail/static/tests/tours/mail_composer_test_tour.js
alexkuhn: Oops

addons/mail/static/src/core/common/composer_actions.js
alexkuhn: In theory this could happen. In practice I'm not sure we can get this problem.

To fix this, I would need to backport `actions` in action params, so that `_disabledCondition` can be based on availability of the `open-full-composer` feature.

ThanhDodeurOdoo (approved):

PR #247255, opened 2026-02-04
[FIX] base: ensure an uniform environment between action variables

Levizar:
### [FIX] base: ensure an uniform environment between action variables

The aim of this commit is to make the environment and context uniform
across the variables that can be accessed in an action.

Context:
The mail override of `<ir.actions.server>._get_eval_context` that
modifies the `env` wasn't reflected on the other variables which lead to
a difference of context.

Before this commit:
This difference resulted in a concurrent update error caused by a direct
mail send in method like `_cron_try_auto_reconcile_statement_lines`.

After this commit:
The environment and its context are uniform as expected and cron don't
trigger a mail send directly, removing the concurrent error issue.

task-id: None

The issue was brought in the internal channel, temporarily fixed by
adding the context in the server action and seen again in odoo.com logs
through another cron.

### DIFF
--- a/addons/mail/models/ir_actions_server.py
+++ b/addons/mail/models/ir_actions_server.py
@@ -391,7 +391,4 @@ class IrActionsServer(models.Model):
         key set to False in the context. This way all notification emails linked
         to the currently executed action will be set in the queue instead of
         sent directly. This will avoid possible break in transactions. """
-        eval_context = super()._get_eval_context(action=action)
-        env = eval_context['env']
-        eval_context['env'] = env(context={**env.context, 'mail_notify_force_send': False})
-        return eval_context
+        return super(IrActionsServer, self.with_context(mail_notify_force_send=False))._get_eval_context(action=action)


odoo/addons/base/models/ir_actions.py
```diff
@@ -1195,9 +1202,10 @@ def _run(self, records, eval_context):
                 if not active_id:  # onchange on new record
                     res = runner(self, eval_context=eval_context)
             active_ids = self.env.context.get('active_ids', [active_id] if active_id else [])
+            context = {**eval_context['env'].context}
```
kmagusiak: Why do you make a copy here?

odoo/addons/base/models/ir_actions.py
```diff
@@ -1195,9 +1202,10 @@ def _run(self, records, eval_context):
...
             for active_id in active_ids:
                 # run context dedicated to a particular active_id
-                run_self = self.with_context(active_ids=[active_id], active_id=active_id)
+                run_self = self.with_context(**context, active_ids=[active_id], active_id=active_id)
```
kmagusiak: There are some gotches in python: creating a dict with conflicting keys keep the last one, but calling methods with conflicting keys raises errors. Here if context contains "active_id", it will crash.

odoo/addons/base/models/ir_actions.py
```diff
@@ -1171,6 +1171,13 @@ def run(self):
...
             eval_context = self._get_eval_context(action)
+            # Ensure environment and context consistency accross all the accessible variable
+            eval_context['model'] = eval_context['model'].with_env(eval_context['env'])
+            if eval_context['record']:
+                eval_context['record'] = eval_context['record'].with_env(eval_context['env'])
+            if eval_context['records']:
+                eval_context['records'] = eval_context['records'].with_env(eval_context['env'])
+
             records = eval_context.get('record') or eval_context['model']
             records |= eval_context.get('records') or eval_context['model']
```
kmagusiak: Just some small changes. I am not sure records and records are always in the context.
```suggestion
            records = eval_context['model'] = eval_context['model'].with_env(eval_context['env'])
            if 'record' in eval_context:
                records = eval_context['record'] = eval_context['record'] .with_env(records.env)
            records |= eval_context.get('records') or eval_context['model']
            if 'records' in eval_context:
                records = eval_context['records'] = records | eval_context['records']
```

kmagusiak (changes requested):

odoo/addons/base/models/ir_actions.py
Levizar: It did in the tests indeed :smile:

kmagusiak (approved):
@robodoo r+

PR #247270, opened 2026-02-04
[FIX] web: properly handle null element passed to scrollTo

iacio-odoo:
Steps to reproduce:
1. Ensure you have a good amount of contacts set up in your database
2. Ensure that you have an IAP Account setup for the service "Partner Autocomplete"
3. Zoom in on your browser to at least 150% AND/OR make your browser window incredibly short
4. Go to the CRM app
5. Go into the form view of an opportunity
6. Type 3 characters or more into the 'Contact' field
7. Observe the error

An Unhandled Promise error could occur if, after typing more than 3 characters in an autocomplete field, the dropdown is scrollable, and Partner Autocomplete finishes loading before the normal autocomplete does. In this scenario, normal autocomplete is still loading and has not rendered it's options. This causes a `null` element to be passed [`scrollTo()`](https://github.com/odoo/odoo/blob/555df96ac87a51405767c32f0396f762d458fd29/addons/web/static/src/core/utils/scrolling.js#L80), which causes an error when it tries to access `element.parentElement`. This change ensures that this scenario is properly handled and does not cause the database to crash.

opw-5491270

### DIFF
--- a/addons/web/static/tests/core/autocomplete.test.js
+++ b/addons/web/static/tests/core/autocomplete.test.js
@@ -890,3 +890,32 @@ test("items are selected only when the mouse moves, not just on enter", async ()
         "ui-state-active"
     );
 });
+
+test("do not attempt to scroll if element is null", async () => {
+    const def = new Deferred();
+    class Parent extends Component {
+        static template = xml`<AutoComplete value="''" sources="sources" />`;
+        static components = { AutoComplete };
+        static props = [];
+
+        sources = [
+            buildSources(async () => {
+                await def;
+                return [item("delayed one"), item("delayed two"), item("delayed three")];
+            }),
+            buildSources(Array.from(Array(20)).map((_, index) => item(`item ${index}`))),
+        ].flat();
+    }
+
+    await mountWithCleanup(Parent);
+    queryOne(`.o-autocomplete input`).focus();
+    queryOne(`.o-autocomplete input`).click();
+    await animationFrame();
+    expect(".o-autocomplete .dropdown-menu").toHaveCount(1);
+    expect(".o-autocomplete .dropdown-item").toHaveCount(21);
+    expect(".o-autocomplete .dropdown-item:eq(0)").toHaveClass("o_loading");
+
+    def.resolve();
+    await animationFrame();
+    expect(".o-autocomplete .dropdown-item").toHaveCount(23); // + 3 items - loading
+});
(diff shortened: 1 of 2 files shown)


addons/web/static/src/core/utils/scrolling.js
```diff
@@ -77,7 +77,7 @@ export function closestScrollableY(el) {
  */
 export function scrollTo(element, options = {}) {
```
aab-odoo, line 78: that seems wrong to call scrollTo with `null` as element. The `element` attribute is mandatory in its API, and is expected to be an HTMLElement. It seems to me that the issue rather comes from  [here](https://github.com/odoo/odoo/blob/19.0/addons/web/static/src/core/autocomplete/autocomplete.js#L492), which could instead be
```diff
diff --git a/addons/web/static/src/core/autocomplete/autocomplete.js b/addons/web/static/src/core/autocomplete/autocomplete.js
index a9dd5ddc45fe..d58889181a99 100644
--- a/addons/web/static/src/core/autocomplete/autocomplete.js
+++ b/addons/web/static/src/core/autocomplete/autocomplete.js
@@ -497,7 +497,10 @@ export class AutoComplete extends Component {
             return;
         }
         if (isScrollableY(this.listRef.el)) {
-            scrollTo(this.listRef.el.querySelector(`#${this.activeSourceOptionId}`));
+            const element = this.listRef.el.querySelector(`#${this.activeSourceOptionId}`);
+            if (element) {
+                scrollTo(element);
+            }
         }
     }
 }
```

Also, a test would be nice (in autocomplete.test.js then)

iacio-odoo: this is true, thank you for the feedback! i will revise this and write a test👍🏻

addons/web/static/tests/core/autocomplete.test.js
```diff
@@ -890,3 +890,45 @@ test("items are selected only when the mouse moves, not just on enter", async ()
...
+    }
+
+    await mountWithCleanup(Parent);
+    const input = queryFirst(`.o-autocomplete input`);
+    input.focus();
+    input.click();
+    await animationFrame();
+    const element = await queryFirst('#autocomplete_0_0');
+    await expect(element).not.toBe(null);
+    document.querySelector('#autocomplete_0_0').remove();
```
aab-odoo: you're no supposed to manipulate the DOM like this. You should instead try to reproduce the scenario that was breaking

PR #197286, opened 2025-02-11
[IMP] product: product administrator group

R1D1CUL0US:
Remove all write/create/delete permission on product related models from the different 'manager' roles in different modules. Instead, we use a new product administrator group for giving these permissions.

task-4578961

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247276, opened 2026-02-04
[FIX] sale: allow sale user create product.attribute.custom.value

mma1377:
### Issue:
Due to this issue, sale group cannot create a sale order with a product with custom value attribute.

#### Steps to reproduce:
1- Create a product using admin with a custom value attribute.
2- Using demo user with sale access group, create a so.
3- Add the created product, and fill the custom value.

The sale order cannot be saved due to access error.

### Cause:
This is due to #197286. However that shouldn't have been applied to `product.attribute.custom` as that shouldn't be only accessed by people who can manage product like the rest of deleted accesses, but also it's needed by sale groups creating SOs.

opw-5498719


Feyensv (approved):
robodoo delegate+

PR #247278, opened 2026-02-04
[FIX] stock_account: remove store from avg_cost in multi-company

cmal-odoo:
Field avg_cost was set to store=True in 42d3e34, but the compute method _compute_avg_cost() is company-context-dependent (calls _run_fifo which filters by env.company). Storing a single value causes incorrect valuations in multi-company databases with FIFO + lot valuation.

In multi-company setups with shared products (company_id=False), the lot's avg_cost would compute in one company's context and store that value globally, causing all other companies to see the wrong cost.

Fix: Remove store=True to compute dynamically per company context. Cannot use company_dependent=True as it requires JSONB migration.

opw-5446941

---

### DIFF
--- a/addons/stock_account/models/stock_lot.py
+++ b/addons/stock_account/models/stock_lot.py
@@ -7,7 +7,7 @@ class StockLot(models.Model):
     _inherit = 'stock.lot'
 
     lot_valuated = fields.Boolean(related='product_id.lot_valuated', readonly=True, store=False)
-    avg_cost = fields.Monetary(string="Average Cost", compute='_compute_value', compute_sudo=True, readonly=True, currency_field='company_currency_id')
+    avg_cost = fields.Monetary(string="Average Cost", compute='_compute_value', compute_sudo=True, store=False, readonly=True, currency_field='company_currency_id')
     total_value = fields.Monetary(string="Total Value", compute='_compute_value', compute_sudo=True, currency_field='company_currency_id')
     company_currency_id = fields.Many2one('res.currency', 'Valuation Currency', compute='_compute_value', compute_sudo=True)
     standard_price = fields.Float(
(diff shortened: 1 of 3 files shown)


addons/stock_account/models/stock_lot.py
```diff
@@ -7,12 +7,12 @@ class StockLot(models.Model):
     _inherit = 'stock.lot'
 
     lot_valuated = fields.Boolean(related='product_id.lot_valuated', readonly=True, store=False)
-    avg_cost = fields.Monetary(string="Average Cost", compute='_compute_avg_cost', store=True, readonly=True, currency_field='company_currency_id')
```
svs-odoo: There is no point to keep to use `_compute_avg_cost` if the field is not computed anymore, you can compute `avg_cost` in `_compute_value` too.
Thus, it will need to be in compute sudo too and you can update `_compute_avg_cost` to not compute anything (remove the depends, set the value at 0, add a comment to explain why this method shouldn't be used)

svs-odoo (commented):
Hello @cmal-odoo 
A test to ensure the described case is fixed is needed

addons/stock_account/models/stock_lot.py
cmal-odoo: Added test, moved avg cost computation to _compute_value, deprecated _compute_avg_cost method in latest commit as requested.

addons/stock_account/models/stock_lot.py
```diff
@@ -25,43 +25,48 @@ def _compute_value(self):
...
-                total_value = lot.product_id._run_fifo(qty_available, at_date=at_date, lot=lot)
-            lot.avg_cost = total_value / qty_available if qty_available else 0.0
+        The computation was moved to _compute_value() which already handles company
+        context correctly and doesn't store the result.
+
+        This stub method is kept to avoid breaking any code that might call it directly,
+        but it intentionally does nothing.
+        """
+        for lot in self:
+            lot.avg_cost = 0.0
```
svs-odoo: You can't read a basic field directly from a recordset, but you can assign it a value from a recordset (the value will be set for every record in the set)
```suggestion
        self.avg_cost = 0.0
```

addons/stock_account/models/stock_lot.py
```diff
@@ -25,43 +25,48 @@ def _compute_value(self):
...
-                total_value = lot.product_id._run_avco(at_date=at_date, lot=lot)[1]
-            else:
-                total_value = lot.product_id._run_fifo(qty_available, at_date=at_date, lot=lot)
-            lot.avg_cost = total_value / qty_available if qty_available else 0.0
+        The computation was moved to _compute_value() which already handles company
+        context correctly and doesn't store the result.
+
+        This stub method is kept to avoid breaking any code that might call it directly,
+        but it intentionally does nothing.
+        """
```
svs-odoo: Very good explanation but keep that for the commit message 😅 
You can keep a minimal comment to avoid someone re-adding code in it by error, but prefer the commit message to give historical and context explanation
```suggestion
        # DEPRECATED: This method is no longer used.
```

addons/stock_account/tests/test_multicompany_lot_valuation.py
```diff
@@ -0,0 +1,79 @@
...
+            'tracking': 'serial',
+            'company_id': False,
+            'lot_valuated': True,
+        })
+
+        cls.warehouse_b = cls.env['stock.warehouse'].with_company(cls.company_b).create({
+            'name': 'Warehouse B',
+            'code': 'WHB',
+            'company_id': cls.company_b.id,
+        })
```
svs-odoo: A warehouse is automatically created alongside the company's creation
```suggestion
        cls.warehouse_b = cls.env['stock.warehouse'].search([('company_id', '=', cls.company_b.id)])
```

addons/stock_account/tests/test_multicompany_lot_valuation.py
```diff
@@ -0,0 +1,79 @@
...
+
+        cls.supplier_location = cls.env.ref('stock.stock_location_suppliers')
+
+    def test_lot_avg_cost_multicompany_fifo(self):
```
svs-odoo, line 41: It feels like this test should be in `TestLotValuation` instead of having its own class, but for now, `TestLotValuation` is skipped 😬

PR #247284, opened 2026-02-04
[FIX] sale: robustness in _create_invoices sequence handling

Vyas106:
This commit acts as a robustness fix for _create_invoices. It explicitly handles the 'sequence' argument by assigning it to the result dictionary after the _prepare_invoice_line call. This prevents TypeErrors when third-party modules override _prepare_invoice_line without accepting **kwargs, ensuring greater compatibility and stability.

Current behavior before PR:

---


Feyensv:
It is the responsibility of third-party modules to correctly override methods as they were defined in the base module.
Some overrides might add other values to the kwargs, and your code shouldn't break because it doesn't handle those ...

PR #247288, opened 2026-02-04
[FIX] stock_account: prevent standard_price update on empty stock

Vyas106:
Revaluing an inventory move (e.g., Landed Costs or manual adjustment) for a product that is currently out of stock (quantity <= 0) incorrectly recalculates and updates the product's `standard_price`. This causes "ghost" cost fluctuations for future stock based on past, fully consumed inventory.

Current behavior before PR:
When `qty_available` is 0, `_update_standard_price` still applies a new cost based on the revaluation of old moves.

`_update_standard_price` now skips the update if `qty_available <= 0`, treating the revaluation purely as an accounting variance without affecting the current unit cost of the product.

---


adr-odoo (changes requested):
--

PR #247310, opened 2026-02-05
[FIX] website: ensure vertical and menu sale 3 headers align navbar correctly

krip-odoo:
Steps to produce:
---
- Install website module.
- Go to Website > Edit.
- Select the header and choose Vertical or Menu Sale 3 template.

Issue:
---
- In both Vertical and Menu Sale 3 templates, the navbar is aligned to the `start`.
- This happens because `justify-content-start` is applied to the navbar container.

Solution:
---
- For **Vertical** template, it should be `justify-content-center`.
- For **Menu sale** 3 template, it should be `justify-content-end`.

Before:
---
- `Vertical`        
image

- `Menu sale - 3`
image

After:
---
- `Vertical`
image

- `Menu sale - 3`
image

Note:
---
- As this is a small and simple case, the tour has not been added to avoid unnecessary overhead.

opw-5908224
---

### DIFF
--- a/addons/website/views/website_templates.xml
+++ b/addons/website/views/website_templates.xml
@@ -833,7 +833,7 @@
                         <t t-call="website.placeholder_header_call_to_action"/>
                     </ul>
                 </div>
-                <t t-call="website.navbar_nav_wrapper" _wrapper_class.f="justify-content-start">
+                <t t-call="website.navbar_nav_wrapper" _wrapper_class.f="justify-content-center">
                     <!-- Navbar -->
                     <t t-call="website.navbar_nav"
                         _nav_class.f="pb-0">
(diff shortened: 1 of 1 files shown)


alsh-odoo (approved):

nle-odoo (commented):
hello,

in 19.0 there is a new option "Desktop menu" that allows you to set the menu to left / center / right.

image

The template is defined with left alignment, and for example for vertical menu there is `template_header_vertical_align_center` that align to the center, and `template_header_vertical_align_right` that align to the right.

This is defined in addons/website/static/src/builder/plugins/options/header/header_navigation_option.xml:

https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/website/static/src/builder/plugins/options/header/header_navigation_option.xml#L9-L38

With the current change, this is breaking this option and we can't align the menu to the left.

When you have an alignement setting on a menu, it is kept when switching to another menu so it might also be intended to not set a forced "alignment" when switching the menu (in Sales 3 and Vertical).

PR #247311, opened 2026-02-05
[IMP] payment_custom: make pending message computation generic

dija-odoo:
Change the way pending payment instructions works for `wire_transfer`.

- Introduce generic template for instructions to be displayed after the
  payment completion(status and confirmation pages)
- Choose a specific bank account to show payment instructions.
- Allow easy extension for other custom modes (e.g. SEPA)

task-4975272

### DIFF
--- a/addons/account_payment/models/payment_provider.py
+++ b/addons/account_payment/models/payment_provider.py
@@ -16,6 +16,17 @@ REPORT_REASONS_MAPPING = {
 class PaymentProvider(models.Model):
     _inherit = 'payment.provider'
 
+    bank_account_id = fields.Many2one(
+        string="Bank Account",
+        comodel_name='res.partner.bank',
+        domain='[("partner_id", "=", company_partner_id)]',
+        compute='_compute_bank_account_id',
+        store=True,
+        readonly=False,
+        copy=False,
+        check_company=True,
+    )
+    company_partner_id = fields.Many2one(comodel_name='res.partner', related='company_id.partner_id')
     journal_id = fields.Many2one(
         string="Payment Journal",
         help="The journal in which the successful transactions are posted.",
@@ -33,6 +44,11 @@ class PaymentProvider(models.Model):
     )
 
     #=== COMPUTE METHODS ===#
+
+    @api.depends('company_id')
+    def _compute_bank_account_id(self):
+        for provider in self:
+            provider.bank_account_id = provider.company_id.partner_id.bank_ids[:1]
 
     def _ensure_payment_method_line(self, allow_create=True):
         self.ensure_one()
--- a/addons/account_payment/views/payment_provider_views.xml
+++ b/addons/account_payment/views/payment_provider_views.xml
@@ -17,6 +17,15 @@
                     required="state != 'disabled' and code not in ['none', 'custom']"/>
                 <field name="support_refund" groups="base.group_no_one"/>
             </group>
+            <group name="payment_followup_bank_details" position="inside">
+                <field
+                    name="bank_account_id"
+                    context="{
+                        'default_company_id': company_id,
+                        'default_partner_id': company_partner_id,
+                    }"
+                />
+            </group>
             <field name="available_country_ids" position="after">
                 <field
                     name="available_pricelist_ids"
--- a/addons/payment/models/payment_provider.py
+++ b/addons/payment/models/payment_provider.py
@@ -713,6 +713,7 @@ class PaymentProvider(models.Model):
         :return: The pending message.
         :rtype: str
         """
+        self.ensure_one()
         return self.pending_msg
 
     def _is_tokenization_required(self, **_kwargs):
--- a/addons/payment/models/payment_provider.py
+++ b/addons/payment/models/payment_provider.py
@@ -707,6 +707,14 @@ class PaymentProvider(models.Model):
 
         return providers
 
+    def _get_pending_msg(self, **_kwargs):
+        """Return the message when the payment is pending after processing.
+
+        :return: The pending message.
+        :rtype: str
+        """
+        return self.pending_msg
+
     def _is_tokenization_required(self, **_kwargs):
         """Return whether tokenizing the transaction is required given its context.
 
--- a/addons/payment/views/payment_provider_views.xml
+++ b/addons/payment/views/payment_provider_views.xml
@@ -99,6 +99,11 @@
                                            options="{'no_create': True}"/>
                                 </group>
                                 <group string="Payment Followup" name="payment_followup" invisible="1"/>
+                                <group
+                                    string="Payment Followup"
+                                    name="payment_followup_bank_details"
+                                    invisible="1"
+                                />
                             </group>
                         </page>
                         <page string="Messages"
--- a/addons/payment_custom/data/payment_provider_data.xml
+++ b/addons/payment_custom/data/payment_provider_data.xml
@@ -4,18 +4,15 @@
     <record id="payment.payment_provider_transfer" model="payment.provider">
         <field name="code">custom</field>
         <field name="redirect_form_view_id" ref="payment.generic_redirect_form"/>
-        <!-- Clear the default value before recomputing the pending_msg -->
-        <field name="pending_msg" eval="False"/>
         <field name="custom_mode">wire_transfer</field>
         <field name="payment_method_ids"
                eval="[Command.set([
                          ref('payment_custom.payment_method_wire_transfer'),
                      ])]"
         />
+        <field name="pending_msg" type="html">
+            <p>Please find the payment details below.</p>
+        </field>
     </record>
-
-    <function model="payment.provider"
-              name="_transfer_ensure_pending_msg_is_set"
-              eval="[[ref('payment.payment_provider_transfer')]]"/>
 
 </odoo>
--- a/addons/payment_custom/data/payment_provider_data.xml
+++ b/addons/payment_custom/data/payment_provider_data.xml
@@ -17,12 +17,9 @@
         <field name="custom_mode">wire_transfer</field>
         <field name="is_live">True</field>
         <field name="redirect_form_view_id" ref="payment.generic_redirect_form"/>
-        <!-- Clear the default value before recomputing the pending_msg -->
-        <field name="pending_msg" eval="False"/>
+        <field name="pending_msg" type="html">
+            <p>Please find the payment details below.</p>
+        </field>
     </record>
-
-    <function model="payment.provider"
-              name="_transfer_ensure_pending_msg_is_set"
-              eval="[[ref('payment.payment_provider_transfer')]]"/>
 
 </odoo>
--- a/addons/payment_custom/models/payment_provider.py
+++ b/addons/payment_custom/models/payment_provider.py
@@ -25,14 +25,28 @@ class PaymentProvider(models.Model):
     qr_code = fields.Boolean(
         string="Enable QR Codes", help="Enable the use of QR-codes when paying by wire transfer."
     )
+    company_partner_id = fields.Many2one(
+        comodel_name="res.partner", related="company_id.partner_id"
+    )
+    partner_bank_id = fields.Many2one(
+        string="Bank Account",
+        comodel_name="res.partner.bank",
+        domain='[("partner_id", "=", company_partner_id)]',
+        compute="_compute_partner_bank_id",
+        store=True,
+        readonly=False,
+        copy=False,
+        check_company=True,
+    )
+
+    # === COMPUTE METHODS === #
+
+    @api.depends("company_id")
+    def _compute_partner_bank_id(self):
+        for provider in self:
+            provider.partner_bank_id = provider.company_id.partner_id.bank_ids[:1]
 
     # === CRUD METHODS ===#
-
-    @api.model_create_multi
-    def create(self, vals_list):
-        providers = super().create(vals_list)
-        providers.filtered(lambda p: p.custom_mode == "wire_transfer").pending_msg = None
-        return providers
 
     def copy_data(self, default=None):
         vals_list = super().copy_data(default=default)
@@ -50,40 +64,17 @@ class PaymentProvider(models.Model):
             return super()._get_default_payment_method_codes()
         return const.DEFAULT_PAYMENT_METHOD_CODES
 
-    # === ACTION METHODS ===#
+    # === BUSINESS METHODS === #
 
-    def action_recompute_pending_msg(self):
-        """Recompute the pending message to include the existing bank accounts."""
-        account_payment_module = self.env["ir.module.module"]._get("account_payment")
-        if account_payment_module.state == "installed":
-            for provider in self.filtered(lambda p: p.custom_mode == "wire_transfer"):
-                company_id = provider.company_id.id
-                accounts = (
-                    self
-                    .env["account.journal"]
-                    .search([
-                        *self.env["account.journal"]._check_company_domain(company_id),
-                        ("type", "=", "bank"),
-                    ])
-                    .bank_account_id
-                )
-                account_names = "".join(
-                    f"<li><pre>{account.display_name}</pre></li>" for account in accounts
-                )
-                bank_account_label = (
-                    provider.env._("Bank Account")
-                    if len(accounts) == 1
-                    else provider.env._("Bank Accounts")
-                )
-                provider.pending_msg = (
-                    f"<div>"
-                    f"<h5>{provider.env._('Please use the following transfer details')}</h5>"
-                    f"<p><br></p>"
-                    f"<h6>{bank_account_label}</h6>"
-                    f"<ul>{account_names}</ul>"
-                    f"<p><br></p>"
-                    f"</div>"
-                )
+    def _get_custom_bank_account(self):
+        """Return the bank account to display in the pending payment instructions template.
+
+        :return: The bank account of the provider.
+        :rtype: res.partner.bank
+        """
+        if self.custom_mode == "wire_transfer":
+            return self.partner_bank_id
+        return self.env["res.partner.bank"]
 
     # === SETUP METHODS === #
 
@@ -100,10 +91,3 @@ class PaymentProvider(models.Model):
         res = super()._get_removal_values()
         res["custom_mode"] = None
         return res
-
-    def _transfer_ensure_pending_msg_is_set(self):
-        transfer_providers_without_msg = self.filtered(
-            lambda p: p.custom_mode == "wire_transfer" and not p.pending_msg
-        )
-        if transfer_providers_without_msg:
-            transfer_providers_without_msg.action_recompute_pending_msg()
--- a/addons/payment_custom/models/payment_provider.py
+++ b/addons/payment_custom/models/payment_provider.py
@@ -1,6 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from odoo import _, api, fields, models
+from odoo import api, fields, models
+from odoo.exceptions import ValidationError
 from odoo.fields import Domain
 
 from odoo.addons.payment_custom import const
@@ -28,11 +29,22 @@ class PaymentProvider(models.Model):
 
     # === CRUD METHODS ===#
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        providers = super().create(vals_list)
-        providers.filtered(lambda p: p.custom_mode == "wire_transfer").pending_msg = None
-        return providers
+    def _check_required_if_provider(self):
+        """Check that `bank_account_id` have been filled for wire transfer provider."""
+        if "bank_account_id" not in self._fields:
+            return super()._check_required_if_provider()
+
+        wire_transfer_without_bank = self.filtered(
+            lambda provider: (
+                provider.custom_mode == "wire_transfer"
+                and provider.state != "disabled"
+                and not provider.bank_account_id
+            )
+        )
+        if wire_transfer_without_bank:
+            raise ValidationError(self.env._("Bank account must be filled for Wire Transfer."))
+
+        super()._check_required_if_provider()
 
     def _get_default_payment_method_codes(self):
         """Override of `payment` to return the default payment method codes."""
@@ -41,35 +53,23 @@ class PaymentProvider(models.Model):
             return super()._get_default_payment_method_codes()
         return const.DEFAULT_PAYMENT_METHOD_CODES
 
-    # === ACTION METHODS ===#
+    # === BUSINESS METHODS === #
 
-    def action_recompute_pending_msg(self):
-        """Recompute the pending message to include the existing bank accounts."""
-        account_payment_module = self.env["ir.module.module"]._get("account_payment")
-        if account_payment_module.state == "installed":
-            for provider in self.filtered(lambda p: p.custom_mode == "wire_transfer"):
-                company_id = provider.company_id.id
-                accounts = (
-                    self
-                    .env["account.journal"]
-                    .search([
-                        *self.env["account.journal"]._check_company_domain(company_id),
-                        ("type", "=", "bank"),
-                    ])
-                    .bank_account_id
-                )
-                account_names = "".join(
-                    f"<li><pre>{account.display_name}</pre></li>" for account in accounts
-                )
-                provider.pending_msg = (
-                    f"<div>"
-                    f"<h5>{_('Please use the following transfer details')}</h5>"
-                    f"<p><br></p>"
-                    f"<h6>{_('Bank Account') if len(accounts) == 1 else _('Bank Accounts')}</h6>"
-                    f"<ul>{account_names}</ul>"
-                    f"<p><br></p>"
-                    f"</div>"
-                )
+    def _get_custom_provider_bank_details(self):
+        """Return bank details configured on the provider.
+
+        :return: A dictionary containing the beneficiary name and bank account number.
+        :rtype: dict
+        """
+        if self.custom_mode != "wire_transfer" or "bank_account_id" not in self._fields:
+            return {}
+
+        bank_account = self.bank_account_id
+        return {"beneficiary": bank_account.holder_name, "bank_account": bank_account.display_name}
+
+    def _get_custom_bank_related_modes(self):
+        """Return custom modes that rely on bank details for pending payment instructions."""
+        return ["wire_transfer"]
 
     # === SETUP METHODS === #
 
@@ -86,10 +86,3 @@ class PaymentProvider(models.Model):
         res = super()._get_removal_values()
         res["custom_mode"] = None
         return res
-
-    def _transfer_ensure_pending_msg_is_set(self):
-        transfer_providers_without_msg = self.filtered(
-            lambda p: p.custom_mode == "wire_transfer" and not p.pending_msg
-        )
-        if transfer_providers_without_msg:
-            transfer_providers_without_msg.action_recompute_pending_msg()
--- a/addons/payment_custom/models/payment_provider.py
+++ b/addons/payment_custom/models/payment_provider.py
@@ -61,11 +61,24 @@ class PaymentProvider(models.Model):
         :return: A dictionary containing the beneficiary name and bank account number.
         :rtype: dict
         """
-        if self.custom_mode != "wire_transfer" or "bank_account_id" not in self._fields:
-            return {}
+        if bank_account := self._get_custom_bank_account():
+            return {
+                "beneficiary": bank_account.holder_name,
+                "bank_account": bank_account.display_name,
+            }
+        return {}
 
-        bank_account = self.bank_account_id
-        return {"beneficiary": bank_account.holder_name, "bank_account": bank_account.display_name}
+    def _get_custom_bank_account(self):
+        """Return the bank account configured on the custom provider.
+
+        :return: The bank account of the provider.
+        :rtype: record of `res.partner.bank` or None
+        """
+        if (
+            self.custom_mode in self._get_custom_bank_related_modes()
+            and "bank_account_id" in self._fields
+        ):
+            return self.bank_account_id
 
     @api.model
     def _get_custom_bank_related_modes(self):
--- a/addons/payment_custom/models/payment_provider.py
+++ b/addons/payment_custom/models/payment_provider.py
@@ -67,8 +67,9 @@ class PaymentProvider(models.Model):
         bank_account = self.bank_account_id
         return {"beneficiary": bank_account.holder_name, "bank_account": bank_account.display_name}
 
+    @api.model
     def _get_custom_bank_related_modes(self):
-        """Return custom modes that rely on bank details for pending payment instructions."""
+        """Return custom modes that rely on bank details."""
         return ["wire_transfer"]
 
     # === SETUP METHODS === #
--- a/addons/payment_custom/models/payment_provider.py
+++ b/addons/payment_custom/models/payment_provider.py
@@ -69,7 +69,11 @@ class PaymentProvider(models.Model):
 
     @api.model
     def _get_custom_bank_related_modes(self):
-        """Return custom modes that rely on bank details."""
+        """Return custom modes that rely on bank details.
+
+        :return: The list of custom modes.
+        :rtype: list
+        """
         return ["wire_transfer"]
 
     # === SETUP METHODS === #
--- a/addons/payment_custom/models/payment_transaction.py
+++ b/addons/payment_custom/models/payment_transaction.py
@@ -108,3 +108,17 @@ class PaymentTransaction(models.Model):
         elif hasattr(self, "sale_order_ids") and self.sale_order_ids:
             communication = self.sale_order_ids[0].reference
         return communication or self.reference
+
+    def _get_custom_qr_code(self):
+        """Return the QR code to scan to pay the custom provider's bank account, if any.
+
+        :return: The base64 QR code.
+        :rtype: str
+        """
+        self.ensure_one()
+        provider = self.provider_id
+        if provider.qr_code and (bank_account := provider._get_custom_bank_account()):
+            return bank_account.build_qr_code_base64(
+                self.amount, self._get_communication(), None, self.currency_id, self.partner_id
+            )
+        return ""
--- a/addons/payment_custom/views/payment_custom_templates.xml
+++ b/addons/payment_custom/views/payment_custom_templates.xml
@@ -14,27 +14,6 @@
     </template>
 
     <template id="custom_state_header" inherit_id="payment.state_header">
-        <xpath expr="//div[@name='o_payment_status_alert']" position="inside">
-            <t t-set="qr_code"
-               t-value="tx.provider_id.sudo().qr_code and tx.company_id.sudo().partner_id.bank_ids[:1].build_qr_code_base64(tx.amount, tx._get_communication(), None, tx.currency_id, tx.partner_id)"
-            />
-            <t t-if="tx._requires_payment_instructions() and qr_code">
-                <div class="position-relative order-2 d-flex flex-md-column justify-content-center align-items-center align-items-md-stretch w-100 w-md-auto">
-                    <hr class="d-inline d-md-none w-100"/>
-                    <div class="vr d-none d-md-block h-100 mx-auto"/>
-                    <h6 class="my-1 my-md-3 px-3 px-md-0 text-muted">OR</h6>
-                    <hr class="d-inline d-md-none w-100"/>
-                    <div class="vr d-none d-md-block h-100 mx-auto"/>
-                </div>
-                <div class="o_qr_code_card card order-1 order-md-3 bg-info flex-shrink-0">
-                    <div class="card-body d-flex flex-column align-items-center justify-content-center pb-3">
-                        <img class="mb-2 border border-dark rounded" t-att-src="qr_code"/>
-                        <small class="text-center text-wrap lh-sm">Scan me in your banking app</small>
-                    </div>
-                </div>
-            </t>
-        </xpath>
-
         <xpath expr="//t[@t-set='o_payment_status_alert_class']" position="replace">
             <t t-if="tx._requires_payment_instructions()">
                <t t-set="o_payment_status_alert_class"
@@ -49,22 +28,67 @@
         </xpath>
 
         <xpath expr="//div[@id='o_payment_status_message']" position="replace">
-            <div t-if="tx._requires_payment_instructions()"
-                 id="o_payment_status_message"
-                 class="order-3 order-md-1 flex-grow-1"
+            <div
+                t-if="tx._requires_payment_instructions()"
+                id="o_payment_status_message"
+                class="w-100"
             >
-                <div class="card flex-grow-1">
-                    <div id="o_payment_status_message_details" class="card-body">
-                        <t>$0</t>
-                        <hr class="w-100"/>
-                        <strong class="mt-auto">Communication: </strong>
-                        <span t-out="tx._get_communication()"/>
-                    </div>
-                </div>
+                <t t-call="payment_custom.custom_payment_transfer_details" tx_sudo="tx"/>
             </div>
             <t t-else="">$0</t>
         </xpath>
 
+    </template>
+
+    <template id="custom_payment_transfer_details">
+        <t
+            t-set="bank_account"
+            t-value="tx_sudo.provider_id._get_custom_bank_account()"
+        />
+        <t t-set="communication" t-value="tx_sudo._get_communication()"/>
+        <t t-set="qr_code" t-value="tx_sudo._get_custom_qr_code()"/>
+        <div
+            t-attf-class="d-flex flex-column align-items-stretch gap-2 pt-1 #{'flex-md-row gap-md-3' if qr_code else ''}"
+        >
+            <div class="card flex-grow-1 order-3 order-md-1 bg-info">
+                <div class="card-body d-flex flex-column gap-1">
+                    <t t-if="status_message" t-out="status_message"/>
+                    <t t-if="bank_account">
+                        <div class="d-flex align-items-center gap-2">
+                            <b>Beneficiary:</b>
+                            <pre t-out="bank_account.holder_name" class="user-select-all mb-0"/>
+                        </div>
+                        <div class="d-flex align-items-center gap-2">
+                            <b>Bank Account:</b>
+                            <pre t-out="bank_account.display_name" class="user-select-all mb-0"/>
+                        </div>
+                    </t>
+                    <div class="d-flex align-items-center gap-2">
+                        <b>Amount:</b>
+                        <pre class="user-select-all mb-0" t-field="tx_sudo.amount"/>
+                    </div>
+                    <t t-if="communication">
+                        <div class="d-flex align-items-center gap-2">
+                            <b>Communication:</b>
+                            <pre t-out="communication" class="user-select-all mb-0"/>
+                        </div>
+                    </t>
+                </div>
+            </div>
+            <t t-if="qr_code">
+                <div class="position-relative order-2 d-flex flex-md-column justify-content-center align-items-center align-items-md-stretch w-100 w-md-auto">
+                    <h6 class="my-1 my-md-3 px-3 px-md-0 text-muted">OR</h6>
+                </div>
+                <div class="o_qr_code_card card order-1 order-md-3 flex-shrink-0">
+                    <div class="card-body d-flex flex-column align-items-center justify-content-center pb-3">
+                        <img class="mb-2 border border-dark rounded" t-att-src="qr_code"/>
+                        <small class="text-center text-wrap lh-sm">
+                            Scan me in your banking app
+                        </small>
+                    </div>
+                </div>
+            </t>
+        </div>
     </template>
 
 </odoo>
--- a/addons/payment_custom/views/payment_custom_templates.xml
+++ b/addons/payment_custom/views/payment_custom_templates.xml
@@ -83,15 +83,15 @@
                 <div>
                     <b>Beneficiary: </b>
                     <pre
-                        class="user-select-all d-inline"
                         t-out="bank_details['beneficiary']"
+                        class="user-select-all d-inline"
                     />
                 </div>
                 <div>
                     <b>Bank Account: </b>
                     <pre
-                        class="user-select-all d-inline"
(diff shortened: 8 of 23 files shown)


addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -70,10 +70,19 @@
                 <div class="card flex-grow-1">
                     <div id="o_payment_status_message_details" class="card-body">
                         <t>$0</t>
+                        <div t-if="tx.amount" class="mt-3">
+                            <b>Amount: </b>
+                            <span class="o_payment_instructions_box">
+                                <t t-out="tx.amount"
+                                   t-options="{'widget': 'monetary', 'display_currency': tx.currency_id}"/>
```
shrm-odoo:
```suggestion
                                <t
                                    t-out="tx.amount"
                                    t-options="{'widget': 'monetary', 'display_currency': tx.currency_id}"
                                 />
```

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -70,10 +70,19 @@
...
+                            </span>
+                        </div>
                         <t t-if="tx._get_communication()">
-                            <hr class="w-100"/>
-                            <strong class="mt-auto">Communication: </strong>
-                            <span t-out="tx._get_communication()"/>
+                            <div class="mt-3">
+                                <b>Communication: </b>
+                                <span class="o_payment_instructions_box"
+                                      t-out="tx._get_communication()"/>
```
shrm-odoo:
```suggestion
                                <span
                                    class="o_payment_instructions_box"
                                    t-out="tx._get_communication()"
                                />
```

addons/payment_custom/views/payment_provider_views.xml
```diff
@@ -26,20 +26,21 @@
             <group name="payment_followup" position="attributes">
                 <attribute name="invisible">code == 'custom'</attribute>
             </group>
+            <group name="payment_followup" position="after">
+                <group
+                    string="Payment Followup"
+                    name="payment_followup_custom"
+                    invisible="not (code == 'custom' and custom_mode == 'wire_transfer')">
```
shrm-odoo:
```suggestion
                    invisible="not (code == 'custom' and custom_mode == 'wire_transfer')"
                >
```

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -156,8 +156,19 @@
...
                 </div>
                 <t t-if="tx_sudo.provider_code == 'custom'">
-                    <div id="order_reference" t-if="order.reference" class="mt-2">
-                        <b>Communication: </b><span t-out='order.reference'/>
+                    <div id="payment_instructions">
+                        <div t-if="tx_sudo.amount" class="mt-3">
+                            <b>Amount: </b>
+                            <span class="o_payment_instructions_box">
+                                <t t-out="tx_sudo.amount"
+                                   t-options="{'widget': 'monetary', 'display_currency': tx_sudo.currency_id}"/>
```
shrm-odoo:
```suggestion
                                <t
                                    t-out="tx_sudo.amount"
                                    t-options="{'widget': 'monetary', 'display_currency': tx_sudo.currency_id}"
                                />
```

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -156,8 +156,19 @@
...
+                            </span>
+                        </div>
+                        <div t-if="tx_sudo._get_communication()" class="mt-3">
+                            <b>Communication: </b>
+                            <span class="o_payment_instructions_box"
+                                  t-out="tx_sudo._get_communication()"/>
```
shrm-odoo:
```suggestion
                            <span
                                class="o_payment_instructions_box"
                                t-out="tx_sudo._get_communication()"
                            />
```

addons/payment_custom/models/payment_provider.py
```diff
@@ -33,33 +44,51 @@ def create(self, vals_list):
...
-                    f'<h5>{_("Please use the following transfer details")}</h5>' \
-                    f'<p><br></p>' \
-                    f'<h6>{_("Bank Account") if len(accounts) == 1 else _("Bank Accounts")}</h6>' \
-                    f'<ul>{account_names}</ul>'\
-                    f'<p><br></p>' \
-                    f'</div>'
+    # === BUSINESS METHODS ===#
+
+    def _get_pending_msg_bank_account(self):
+        return self.bank_account_id
```
shrm-odoo: Place this method after the method where it is called

addons/payment_custom/models/payment_provider.py
```diff
@@ -24,6 +24,17 @@ class PaymentProvider(models.Model):
...
+        'res.partner',
+        related='company_id.partner_id',
+        store=False,
+    )
+    bank_account_id = fields.Many2one(
+        'res.partner.bank',
+        string="Bank Account",
+        domain="[('partner_id', '=', company_partner_id)]",
+        copy=False,
+    )
```
shrm-odoo: Fields should be in alphabetical order, with `string` as the first attribute.

shrm-odoo (changes requested):
Hello @dija-odoo 
Thanks for your work!
I just have a few small comments.
Please resolve the conflicts also.

addons/payment_custom/models/payment_provider.py
```diff
@@ -28,48 +37,28 @@ class PaymentProvider(models.Model):
...
+        :return: A dictionary containing the beneficiary name and bank account details.
+        :rtype: dict
+        """
+        if self.custom_mode == "wire_transfer":
+            bank_account = self.bank_account_id
+            return {
+                "beneficiary": bank_account.holder_name,
+                "bank_account": bank_account.display_name,
+            }
+        return {}
```
shrm-odoo:
```suggestion
        if self.custom_mode != "wire_transfer":
            return {}
        return {
            "beneficiary": self.bank_account_id.holder_name,
            "bank_account": self.bank_account_id.display_name,
        }
```

It feels a bit more straightforward, but happy to go with whatever you prefer.

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +73,42 @@
...
+                    <pre
+                        class="user-select-all d-inline"
+                        t-out="pending_payment_instructions.get('bank_account')"
+                    />
+                </div>
+            </t>
+            <div>
+                <b>Amount: </b>
+                <pre class="user-select-all d-inline"><t t-out="payment_tx.amount"
+                        t-options="{'widget': 'monetary', 'display_currency': payment_tx.currency_id}"/></pre>
```
shrm-odoo:
```suggestion
                <pre
                    class="user-select-all d-inline"
                    t-out="payment_tx.amount"
                    t-options="{'widget': 'monetary', 'display_currency': payment_tx.currency_id}"
                />
```

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +73,42 @@
...
     </template>
 
+    <template id="custom_pending_payment_instructions">
+        <div class="d-flex flex-column gap-3 pt-2">
+            <t t-set="payment_tx" t-value="tx or tx_sudo"/>
+            <t
+                t-set="pending_payment_instructions"
+                t-value="payment_tx.provider_id.sudo()._get_custom_pending_payment_instructions()"
+            />
+            <t t-set="communication" t-value="payment_tx._get_communication()"/>
```
shrm-odoo: You could move this to where it’s actually used.

addons/payment_custom/views/payment_provider_views.xml
```diff
@@ -26,20 +26,22 @@
             <group name="payment_followup" position="attributes">
                 <attribute name="invisible">code == 'custom'</attribute>
             </group>
+            <group name="payment_followup" position="after">
+                <group
+                    string="Payment Followup"
+                    name="payment_followup_custom"
```
shrm-odoo:
```suggestion
                    name="payment_followup_wire_transfer"
```

addons/payment_custom/models/payment_provider.py
```diff
@@ -28,48 +37,28 @@ class PaymentProvider(models.Model):
...
-                    f"<p><br></p>"
-                    f"<h6>{_('Bank Account') if len(accounts) == 1 else _('Bank Accounts')}</h6>"
-                    f"<ul>{account_names}</ul>"
-                    f"<p><br></p>"
-                    f"</div>"
-                )
+    def _get_custom_pending_payment_instructions(self):
+        """Return bank details configured on the provider.
+
+        :return: A dictionary containing the beneficiary name and bank account details.
```
shrm-odoo:
```suggestion
        :return: A dictionary containing the beneficiary name and bank account name.
```

shrm-odoo (changes requested):
Thank you for the changes.
It’s looking good now 👍
I’ve added a few suggestions below.

shrm-odoo (approved):
Thank you for your work
LGTM 🦆

addons/payment_custom/data/payment_provider_data.xml
kcv-odoo, line 22: Don't we need upgrade script for this? As in this case we probably want to update pending_msg for custom_providers or not 🤔

addons/payment_custom/models/payment_provider.py
kcv-odoo:
```suggestion
    custom_bank_account_id = fields.Many2one(
```
- Since currently it is specific to custom providers
- Also we could make it compute to set default bank from company and it'll reset if company changed like Journal on other providers

addons/payment_custom/data/payment_provider_data.xml
kcv-odoo: Since this is not specific to order, it can also display in invoice and even on confirmed order, isn't it wrong? @Jean-boje

addons/website_sale/templates/checkout/confirmation_templates.xml
kcv-odoo: Why? It'll also impact other state messages

see:
Image

addons/payment_custom/models/payment_provider.py
kcv-odoo:
```suggestion
        comodel_name="res.partner", related="company_id.partner_id"
```
Related are non store by default

kcv-odoo: Since this method does not return details specific to pending payment or there are no instruction name seems too generic
`_get_custom_provider_bank_details` seems better, wdyt?

addons/payment_custom/views/payment_custom_templates.xml
kcv-odoo: IMO It would be better to update variable name when call this template instead this

like <t t-call="xyx"><t t-set="tx" t-value="tx_sudo"/></t>

this we we don't need to re-assign variable into 3rd variable, wdyt?

kcv-odoo: No need for .get since we always pass those keys

kcv-odoo: Does it make sense to not display, bank details but Amount and Communication? would that be useful, And since bank_account_id is required and we always share bank_details for custom if guess we don't need to check condition whether _get_custom_pending_payment_instructions return values or not. wdyt?

addons/website_sale/templates/checkout/confirmation_templates.xml
kcv-odoo: Also didn't we wanted to remove this icon?

Image

kcv-odoo (changes requested):
Thank you for your work! Added some small comments

Also I was thinking can we make back_account_id related to `journal_id` so it will get and set values from `journal`, with inverse to make it writable and set first journal related to bank account, this way we don't need to store bank field and it'll have consistent value with `journal`, wdyt?
(Not that it make much more difference functionally)

addons/payment_custom/views/payment_custom_templates.xml
dija-odoo: For example in `cash on delivery,` `_get_custom_pending_payment_instructions`(previously) returns nothing so bank details are hidden, but Amount and Communication are still shown which is useful. So I think the check is required.

addons/payment_custom/models/payment_provider.py
dija-odoo: For now, I used a default instead of a computed field to keep the behavior simple. Can be revisited later if needed (with compute field), after the primary approach is decided.

addons/payment_custom/data/payment_provider_data.xml
dija-odoo: I have made an upgrade PR already, maybe I missed to add it in the pad 🙂

Jean-boje: Hello @kcv-odoo , indeed, can't we do a specific pending message for the eCommerce then? WDYT? If so we can have something like "Please find the payment details below."

addons/account_payment/views/payment_provider_views.xml
```diff
@@ -17,6 +17,13 @@
                     required="state != 'disabled' and code not in ['none', 'custom']"/>
                 <field name="support_refund" groups="base.group_no_one"/>
             </group>
+            <group name="payment_followup_bank_details" position="inside">
+                <field
+                    name="bank_account_id"
+                    context="{'default_partner_id': company_partner_id}"
+                    required="state != 'disabled'"
```
kcv-odoo: Add another condition like `journal_id`

addons/account_payment/models/payment_provider.py
```diff
@@ -31,6 +31,14 @@ class PaymentProvider(models.Model):
         help="Only allow this payment provider when the partner's pricelist matches one of these.",
         comodel_name='product.pricelist',
     )
+    bank_account_id = fields.Many2one(
+        string="Bank Account",
+        comodel_name="res.partner.bank",
+        domain="[('partner_id', '=', company_partner_id)]",
+        default=lambda self: self.env.company.partner_id.bank_ids[:1],
```
kcv-odoo: Make it compute as provider can have different company then `self.env.company`, wdyt?

addons/payment/views/payment_provider_views.xml
```diff
@@ -99,6 +99,7 @@
                                            options="{'no_create': True}"/>
                                 </group>
                                 <group string="Payment Followup" name="payment_followup" invisible="1"/>
+                                <group name="payment_followup_bank_details" invisible="1"/>
```
kcv-odoo: I guess we could add string here as there is no need to add string in override

addons/payment_custom/models/payment_provider.py
```diff
@@ -28,48 +28,25 @@ class PaymentProvider(models.Model):
...
-                    f"<ul>{account_names}</ul>"
-                    f"<p><br></p>"
-                    f"</div>"
-                )
+        :return: A dictionary containing the beneficiary name and bank account number.
+        :rtype: dict
+        """
+        if self.custom_mode != "wire_transfer":
+            return {}
+        bank_account = self.bank_account_id
```
kcv-odoo: Need to check if `account_payment` is installed or not, or self has `bank_account_id` field

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -64,11 +64,10 @@
...
                         <t>$0</t>
-                        <t t-if="tx._get_communication()">
-                            <hr class="w-100"/>
-                            <strong class="mt-auto">Communication: </strong>
-                            <span t-out="tx._get_communication()"/>
-                        </t>
+                        <t
+                            t-call="payment_custom.custom_pending_payment_instructions"
+                            payment_tx="tx"
+                        />
```
kcv-odoo: IMO only tx is better so no need to pass payment_tx from here also `custom_payment_instuctions` as tx can be draft or pending in this case, wdyt?

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +76,46 @@
...
+    <template id="custom_pending_payment_instructions">
+        <div class="d-flex flex-column gap-3 pt-2 pb-3">
+            <span
+                id="pending_payment_confirmation_msg"
+                t-if="is_confirmation and payment_tx.provider_id.sudo().custom_mode == 'wire_transfer'"
+            >
+                Your order will be confirmed after payment is received.
+            </span>
+            <t
+                t-set="pending_payment_instructions"
```
kcv-odoo:
```suggestion
                t-set="bank_details"
```

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +76,46 @@
...
     </template>
 
+    <template id="custom_pending_payment_instructions">
+        <div class="d-flex flex-column gap-3 pt-2 pb-3">
+            <span
+                id="pending_payment_confirmation_msg"
+                t-if="is_confirmation and payment_tx.provider_id.sudo().custom_mode == 'wire_transfer'"
+            >
+                Your order will be confirmed after payment is received.
+            </span>
```
kcv-odoo: Not good idea, can't we update main template to get pending method from a function then override here and override in website_sale to updated message as per our need, like _get_pending_msg which will return self.pending_msg in payment and here in our case if provider is custom and code is `wire_transfer` update it, wdyt?

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -122,17 +122,6 @@
...
-                    title="Edit"
-                    class="btn btn-sm btn-link text-dark float-end"
-                    role="button"
-                    groups="base.group_system"
-                    target="_blank"
-                    aria-label="Edit"
-                    t-attf-href="/odoo/action-payment.action_payment_provider/{{tx_sudo.provider_id.id}}"
-                >
-                    <i class="fa fa-pencil"/>
-                </a>
```
kcv-odoo: Do we need to remove this for all the providers and all the state of tx? can you check with boje?

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -156,8 +145,12 @@
                     </t>
                 </div>
                 <t t-if="tx_sudo.provider_code == 'custom'">
-                    <div id="order_reference" t-if="order.reference" class="mt-2">
-                        <b>Communication: </b><span t-out='order.reference'/>
+                    <div id="custom_payment_instructions">
+                        <t
+                            t-call="payment_custom.custom_pending_payment_instructions"
```
kcv-odoo: Will this template exist when payment_custom is not installed?

kcv-odoo (changes requested):

addons/website_sale/templates/checkout/confirmation_templates.xml
dija-odoo: Yes, we need to remove it completely

addons/payment/views/payment_provider_views.xml
dija-odoo: Wouldn’t it look weird to have two groups with the same label? I was thinking it might be cleaner to override it and assign a string there.

addons/payment_custom/models/payment_provider.py
dija-odoo: In most cases wouldn't `account_payment` already be installed? The payment flow via all (ecommerce, sales, account) somehow gets `account_payment` installed.

I do agree that adding a guard here makes sense! As `payment_custom` doesn't directly depend on `account_payment`, but I am just wondering what flow am I missing?

addons/website_sale/templates/checkout/confirmation_templates.xml
dija-odoo: `website_sale` needs `payment_custom` so this will work 🤔

addons/payment/views/payment_provider_views.xml
kcv-odoo: Nope! 
I wouldn't want it to add same string in every override in each module with same string while I can directly add string on parent node

addons/payment_custom/models/payment_provider.py
kcv-odoo: Yes, but since payment_custom can be installed without `account` then we should ensure it can be used without account, for modularity

addons/website_sale/templates/checkout/confirmation_templates.xml
kcv-odoo: I am not able to see it's dependency :thinking: or I am missing something?

dija-odoo: `website_sale` > `delivery` > `payment_custom`. 
So it should be available

addons/account_payment/models/payment_provider.py
```diff
@@ -31,9 +31,38 @@ class PaymentProvider(models.Model):
...
+        string="Bank Account",
+        comodel_name='res.partner.bank',
+        domain='[("partner_id", "=", company_partner_id)]',
+        compute='_compute_bank_account_id',
+        store=True,
+        readonly=False,
+        copy=False,
+        check_company=True,
+    )
+    company_partner_id = fields.Many2one(comodel_name='res.partner', related='company_id.partner_id')
```
kcv-odoo: Move it above

addons/account_payment/models/payment_provider.py
```diff
@@ -31,9 +31,38 @@ class PaymentProvider(models.Model):
...
+            if provider.state == 'disabled' or provider.bank_account_id:
+                continue
+            journal = self.env['account.journal'].search(
+                [
+                    ('bank_account_id', '!=', False),
+                    ('company_id', '=', provider.company_id.id),
+                    ('type', '=', 'bank'),
+                ],
+                limit=1,
+            )
```
kcv-odoo: Seems not useful if we are not getting it from journal of the provider

I think it would be better to get journal from `'account.payment.method.line'` to get proper journal which is most related to provider journal

Or if it is not that useful then just set `provider.company_id.partner_id.bank_ids[:1]` as before but in compute if there is not better way to get journal as it is non store

Or just make journal store if it simplifies things

addons/account_payment/models/payment_provider.py
```diff
@@ -31,9 +31,38 @@ class PaymentProvider(models.Model):
...
+        check_company=True,
+    )
+    company_partner_id = fields.Many2one(comodel_name='res.partner', related='company_id.partner_id')
 
     #=== COMPUTE METHODS ===#
 
+    @api.depends('company_id', 'state')
+    def _compute_bank_account_id(self):
+        for provider in self:
+            if provider.state == 'disabled' or provider.bank_account_id:
```
kcv-odoo: I would not add `bank_account_id` here is it will not update bank account if company get changed, and no need for `state` check

and since it's compute stored it'll only re-compute when company of provider changes

addons/payment_custom/models/payment_provider.py
```diff
@@ -41,35 +56,29 @@ def _get_default_payment_method_codes(self):
...
+
+    def _get_pending_msg(self):
+        """Override of `payment` to return specifc pending message for wire transfer."""
+        if self.code == "custom" and self.custom_mode == "wire_transfer":
+            return self._get_custom_pending_msg()
+
+        return super()._get_pending_msg()
+
+    def _get_custom_pending_msg(self):
+        return self.env._("Your order will be confirmed after payment is received.")
```
kcv-odoo: No, we want this msg for only E-commerce orders so it should have something that ensure in only display orders

Instead add **kw in `_get_pending_msg` add `from_website` in website_sale override and pass it from website_sale template, wdyt?

kcv-odoo (changes requested):

addons/account_payment/models/payment_provider.py
dija-odoo: I have attempted to achieve the behavior very similar to `journal_id` field.
1. Only assign defaults when provider is set to test/enabled ( active states )
2. In any state if company is changed and if the value is set, _company crossover_ error is thrown.
3. Forcing user to unset the field first and then only allow to change the company of the provider.

kcv-odoo: I don't think error is good when it can be auto computed when company changed, as we are setting this regardless of provider of user won't see field to reset if is it is not wire transfer

addons/account_payment/models/payment_provider.py
dija-odoo: The previous implementation was also relying on `account.journal `to gather bank accounts for the pending message. since wire transfer providers don’t have a linked `journal_id`, I have simplified by using the company’s partner bank account directly.

addons/payment_custom/models/payment_provider.py
```diff
@@ -41,35 +56,34 @@ def _get_default_payment_method_codes(self):
...
+    def _get_pending_msg(self, **kwargs):
+        """Override to return a specific pending message for website orders."""
+        if (
+            kwargs.get("from_website")
+            and self.code == "custom"
+            and self.custom_mode in self._get_custom_bank_related_modes()
+        ):
+            return self.env._("Your order will be confirmed after payment is received.")
+
+        return super()._get_pending_msg(**kwargs)
```
kcv-odoo: This should be in website_sale

addons/payment_custom/models/payment_provider.py
```diff
@@ -28,11 +29,25 @@ class PaymentProvider(models.Model):
 
     # === CRUD METHODS ===#
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        providers = super().create(vals_list)
-        providers.filtered(lambda p: p.custom_mode == "wire_transfer").pending_msg = None
-        return providers
+    def _check_required_if_provider(self):
+        """Validate bank account is set for wire transfer when in enabled/test mode."""
```
kcv-odoo:
```suggestion
        """Check that `bank_account_id` have been filled for wire transfer provider."""
```

addons/payment_custom/models/payment_provider.py
```diff
@@ -28,11 +29,25 @@ class PaymentProvider(models.Model):
...
+                provider.code == "custom"
+                and provider.custom_mode == "wire_transfer"
+                and provider.state != "disabled"
+                and not provider.bank_account_id
+            ):
+                raise ValidationError(
+                    self.env._(
+                        "Bank Account must be configured for Wire Transfer payment providers."
+                    )
+                )
```
kcv-odoo:
```suggestion
        if "bank_account_id" not in self._fields:
            return super()._check_required_if_provider()
        
        wire_transfer_without_bank = self.filtered(lambda provider:
            and provider.custom_mode == "wire_transfer"
            and provider.state != "disabled"
            and not provider.bank_account_id
        )
        if wire_transfer_without_bank:
            raise ValidationError(
                self.env._(
                    "Bank account must be filled for Wire Transfer."
                )
            )
                
        super()._check_required_if_provider()
```

- Call super at last so we don't check other validation if there is wire transfer and bank account is not set
- Since just need wire transfer to check bank is filled or not, used filtered instead for bit quicker

Variable name could improve
(Not tested)

addons/payment/models/payment_provider.py
```diff
@@ -707,6 +707,14 @@ def _get_compatible_providers(
 
         return providers
 
+    def _get_pending_msg(self, **_kwargs):
+        """Return the message to be displayed when the transaction is in pending state.
```
kcv-odoo:
```suggestion
        """Return the message when the payment is pending after processing.
```
Nit

addons/payment_custom/views/payment_provider_views.xml
```diff
@@ -26,20 +26,14 @@
             <group name="payment_followup" position="attributes">
                 <attribute name="invisible">code == 'custom'</attribute>
             </group>
+            <group name="payment_followup_bank_details" position="attributes">
+                <attribute name="invisible">
+                    not (code == 'custom' and custom_mode == 'wire_transfer')
+                </attribute>
```
kcv-odoo: Nit: Use separator like it done in `pre_msg` below

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +73,40 @@
 
     </template>
 
+    <template id="custom_payment_instructions">
```
kcv-odoo:
```suggestion
    <template id="custom_payment_transfer_details">
```
Nit

kcv-odoo (changes requested):
Almost there

addons/website_sale/models/payment_provider.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models
+
+
+class PaymentProvider(models.Model):
+    _inherit = "payment.provider"
+
+    # === BUSINESS METHODS === #
+
+    def _get_pending_msg(self, **kwargs):
```
kcv-odoo:
```suggestion
    def _get_pending_msg(self, from_website=False, **kwargs):
```
extract kwargs here instead in method, and don't forget to add extracted kwargs in super call

addons/website_sale/models/payment_provider.py
```diff
@@ -0,0 +1,20 @@
...
+        """Override to return a specific pending message for website orders."""
+        if (
+            kwargs.get("from_website")
+            and self.code == "custom"
```
kcv-odoo: No need to check this

addons/payment_custom/views/payment_custom_templates.xml
```diff
@@ -77,4 +73,40 @@
...
+            <t
+                t-set="bank_details"
+                t-value="tx.provider_id.sudo()._get_custom_provider_bank_details()"
+            />
+            <t t-if="bank_details">
+                <div>
+                    <b>Beneficiary: </b>
+                    <pre
+                        class="user-select-all d-inline"
+                        t-out="bank_details['beneficiary']"
```
kcv-odoo: Nit: swap as t-out is more useful then class, same for other places

addons/payment_custom/models/payment_provider.py
```diff
@@ -41,35 +53,23 @@ def _get_default_payment_method_codes(self):
...
+        :rtype: dict
+        """
+        if self.custom_mode != "wire_transfer" or "bank_account_id" not in self._fields:
+            return {}
+
+        bank_account = self.bank_account_id
+        return {"beneficiary": bank_account.holder_name, "bank_account": bank_account.display_name}
+
+    def _get_custom_bank_related_modes(self):
+        """Return custom modes that rely on bank details for pending payment instructions."""
```
kcv-odoo:
```suggestion
        """Return custom modes that rely on bank details."""
```
No need to specify how it going to use as it can be use for something else

addons/payment_custom/models/payment_provider.py
```diff
@@ -41,35 +53,23 @@ def _get_default_payment_method_codes(self):
...
+        :return: A dictionary containing the beneficiary name and bank account number.
+        :rtype: dict
+        """
+        if self.custom_mode != "wire_transfer" or "bank_account_id" not in self._fields:
+            return {}
+
+        bank_account = self.bank_account_id
+        return {"beneficiary": bank_account.holder_name, "bank_account": bank_account.display_name}
+
+    def _get_custom_bank_related_modes(self):
```
kcv-odoo, line 69: api.models or move it in utils 🤔 I guess it's fine here for now

addons/payment/models/payment_provider.py
```diff
@@ -707,6 +707,14 @@ def _get_compatible_providers(
 
         return providers
 
+    def _get_pending_msg(self, **_kwargs):
+        """Return the message when the payment is pending after processing.
+
```
kcv-odoo, line 712: self.ensure_one()

kcv-odoo (approved):
LGTM! just few non blocking comments

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -127,20 +127,9 @@
...
-                    target="_blank"
-                    aria-label="Edit"
-                    t-attf-href="/odoo/action-payment.action_payment_provider/{{tx_sudo.provider_id.id}}"
-                >
-                    <i class="fa fa-pencil"/>
-                </a>
                 <div>
                     <t t-if="tx_sudo.state == 'pending'">
-                        <t t-out="tx_sudo.provider_id.sudo().pending_msg"/>
+                        <t t-out="tx_sudo.provider_id.sudo()._get_pending_msg(from_website=True, order=order)"/>
```
kcv-odoo, line 132: Only order kwarg is enough as it ensure this call is coming from order 🤔

kcv-odoo:
Can you rebase this branch? seems lots of things change since last changes

addons/payment_custom/models/payment_transaction.py
```diff
@@ -73,17 +70,6 @@ def _get_sent_message(self):
             )
         return message
 
-    def _get_status_message(self, **kwargs):
```
vchu-odoo, line 76: would not it break the default msg for portal or for other custom providers or we are removing the default msg for all custom ones?

addons/payment_custom/models/payment_provider.py
```diff
@@ -50,40 +44,32 @@ def _get_default_payment_method_codes(self):
...
-                    f"<h5>{provider.env._('Please use the following transfer details')}</h5>"
-                    f"<p><br></p>"
-                    f"<h6>{bank_account_label}</h6>"
-                    f"<ul>{account_names}</ul>"
-                    f"<p><br></p>"
-                    f"</div>"
-                )
+    # === BUSINESS METHODS === #
+
+    def _get_custom_provider_bank_details(self):
```
vchu-odoo: i do not think we need this wrapper, we can directly call _get_custom_bank_account and set the values there

addons/payment_custom/models/payment_provider.py
```diff
@@ -50,40 +44,32 @@ def _get_default_payment_method_codes(self):
...
+
+    def _get_custom_bank_account(self):
+        """Return the bank account configured on the custom provider.
+
+        :return: The bank account of the provider.
+        :rtype: record of `res.partner.bank`
+        """
+        return self.env["res.partner.bank"]
+
+    def _get_custom_qr_bank_account(self):
```
vchu-odoo: is it needed? wdy do we have even this fallback? it could be inconsistent with the bank account set on the provider.

addons/account_payment_custom/models/payment_provider.py
```diff
@@ -1,11 +1,30 @@
...
+        string="Bank Account",
+        comodel_name="res.partner.bank",
+        domain='[("partner_id", "=", company_partner_id)]',
+        compute="_compute_bank_account_id",
+        store=True,
+        readonly=False,
+        copy=False,
+        check_company=True,
+    )
+    company_partner_id = fields.Many2one(comodel_name="res.partner", related="company_id.partner_id")
```
vchu-odoo: \> 100

addons/account_payment_custom/models/payment_provider.py
```diff
@@ -1,11 +1,30 @@
...
+
+    # === COMPUTE METHODS === #
+
+    @api.depends("company_id")
```
vchu-odoo: should not we depend on all fields in the chain?
```suggestion
    @api.depends("company_id.partner_id.bank_ids")
```

addons/payment_custom/models/payment_provider.py
```diff
@@ -50,40 +44,32 @@ def _get_default_payment_method_codes(self):
...
+        """
+        if bank_account := self._get_custom_bank_account():
+            return {
+                "beneficiary": bank_account.holder_name,
+                "bank_account": bank_account.display_name,
+            }
+        return {}
+
+    def _get_custom_bank_account(self):
+        """Return the bank account configured on the custom provider.
```
vchu-odoo: to be more precise
```suggestion
        """Return the bank account to display in the pending payment instructions template.
```

addons/payment_custom/models/payment_provider.py
```diff
@@ -50,40 +44,32 @@ def _get_default_payment_method_codes(self):
...
+
+        :return: The bank account of the provider.
+        :rtype: record of `res.partner.bank`
```
vchu-odoo:
```suggestion
        :rtype: res.partner.bank
```

addons/account_payment_custom/tests/common.py
```diff
@@ -12,8 +12,14 @@ class AccountPaymentCustomCommon(AccountPaymentCommon, PaymentCustomCommon):
...
 
+        cls.bank_account = cls.env["res.partner.bank"].create({
+            "account_number": "BANK123456789",
+            "partner_id": cls.env.company.partner_id.id,
+        })
         cls.wire_transfer_provider = cls._prepare_provider(
-            code="custom", custom_mode="wire_transfer"
+            code="custom",
+            custom_mode="wire_transfer",
+            update_values={"bank_account_id": cls.bank_account.id},
```
vchu-odoo: do we use it in the tests?

addons/website_event_sale/tests/test_frontend_buy_tickets.py
```diff
@@ -30,6 +30,7 @@ def setUpClass(cls):
         cls.env.ref('payment.payment_provider_transfer').write({
             'is_live': True,
             'is_published': True,
+            'bank_account_id': cls.bank_account.id,
```
vchu-odoo: not necessary for the test as you noticed :)

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -129,11 +129,16 @@
...
                     </div>
-                    <div t-if="tx_sudo.provider_id.sudo().qr_code">
-                        <t t-set="qr_code" t-value="tx_sudo.company_id.partner_id.bank_ids[:1].build_qr_code_base64(order.amount_total,tx_sudo.reference, None, tx_sudo.currency_id, tx_sudo.partner_id)"/>
+                    <t t-set="provider_sudo" t-value="tx_sudo.provider_id.sudo()"/>
+                    <div t-if="provider_sudo.qr_code">
+                        <t t-set="qr_bank_account" t-value="provider_sudo._get_custom_qr_bank_account()"/>
+                        <t
+                            t-set="qr_code"
+                            t-value="qr_bank_account and qr_bank_account.build_qr_code_base64(order.amount_total, tx_sudo.reference, None, tx_sudo.currency_id, tx_sudo.partner_id)"
+                        />
```
vchu-odoo: seems like we can extract this into the method `tx._get_custom_qr_code(amount, reference)` since there is too much logic for the template and we need to repeat it at least twice, wdyt?

addons/account_payment_custom/models/payment_provider.py
dija-odoo: What I think is even if we extend the depends to the full `bank_ids` chain, the compute still just picks the first bank account so wouldn't that overwrite a bank account someone picked manually. also when i added this compute it was meant to set default for a new provider or when company is changed on provider, wdyt ?

addons/payment_custom/models/payment_transaction.py
dija-odoo: I just realized that a user can even empty the whole `pending_msg` field and then it should fallback to `Finalize your payment` like before. Thanks!

addons/payment_custom/models/payment_provider.py
dija-odoo: Yes, I think keeping it in-sync with set bank account is better. Thanks!

addons/website_sale/templates/checkout/confirmation_templates.xml
dija-odoo: Yes, updated 👍🏻

addons/account_payment_custom/models/payment_provider.py
```diff
@@ -1,11 +1,33 @@
...
+        domain='[("partner_id", "=", company_partner_id)]',
+        compute="_compute_bank_account_id",
+        store=True,
+        readonly=False,
+        copy=False,
+        check_company=True,
+    )
+    company_partner_id = fields.Many2one(
+        comodel_name="res.partner",
+        related="company_id.partner_id",
```
vchu-odoo:
```suggestion
        comodel_name="res.partner", related="company_id.partner_id"
```

addons/account_payment_custom/models/payment_provider.py
vchu-odoo: agree, let us leave only company :)

addons/account_payment_custom/views/payment_provider_views.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
```
vchu-odoo: when we merge the fix for the other tab we should remember to adapt this.

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -127,19 +127,21 @@
...
-                    <div t-if="tx_sudo.provider_id.sudo().qr_code">
-                        <t t-set="qr_code" t-value="tx_sudo.company_id.partner_id.bank_ids[:1].build_qr_code_base64(order.amount_total,tx_sudo.reference, None, tx_sudo.currency_id, tx_sudo.partner_id)"/>
-                        <div class="mt-2" t-if="qr_code">
-                            <h3>Or scan me with your banking app.</h3>
-                            <img class="border border-dark rounded" t-att-src="qr_code"/>
-                        </div>
+                    <t t-set="qr_code" t-value="tx_sudo._get_custom_qr_code(order.amount_total, tx_sudo.reference)"/>
+                    <div class="d-flex flex-column align-items-center flex-shrink-0 pe-3 gap-2" t-if="qr_code">
+                        <img class="border border-dark rounded mb-1" t-att-src="qr_code" style="width: 128px; height: 128px;"/>
+                        <small class="text-center"><b>Scan me with<br/>your Banking App</b></small>
```
vchu-odoo: should not it match `custom_state_header`?

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -127,19 +127,21 @@
...
+                            </div>
+                        </t>
                     </div>
-                    <div t-if="tx_sudo.provider_id.sudo().qr_code">
-                        <t t-set="qr_code" t-value="tx_sudo.company_id.partner_id.bank_ids[:1].build_qr_code_base64(order.amount_total,tx_sudo.reference, None, tx_sudo.currency_id, tx_sudo.partner_id)"/>
-                        <div class="mt-2" t-if="qr_code">
-                            <h3>Or scan me with your banking app.</h3>
-                            <img class="border border-dark rounded" t-att-src="qr_code"/>
-                        </div>
+                    <t t-set="qr_code" t-value="tx_sudo._get_custom_qr_code(order.amount_total, tx_sudo.reference)"/>
```
vchu-odoo: should not we always use tx_sudo._get_communication()? it will already resolve in order reference, so we can drop it and call it directly in the method

addons/account_payment_custom/tests/common.py
dija-odoo: I think there is no need of this anymore 🙂

addons/account_payment_custom/models/payment_provider.py
```diff
@@ -9,6 +9,27 @@
 class PaymentProvider(models.Model):
     _inherit = "payment.provider"
 
+    bank_account_id = fields.Many2one(
```
vchu-odoo: we can define it in payment_custom, there is no dependency on accounting

addons/payment_custom/models/payment_transaction.py
```diff
@@ -108,3 +108,16 @@ def _get_communication(self):
...
+
+        :return: The base64 QR code.
+        :rtype: str
+        """
+        self.ensure_one()
+        provider_sudo = self.provider_id.sudo()
+        if provider_sudo.qr_code and (bank_account := provider_sudo._get_custom_bank_account()):
+            return bank_account.build_qr_code_base64(
+                self.amount, self._get_communication(), None, self.currency_id, self.partner_id
+            )
```
vchu-odoo, line 123:
```suggestion
            return bank_account.build_qr_code_base64(
                self.amount, self._get_communication(), None, self.currency_id, self.partner_id
            )
        return ""
```

addons/payment_custom/models/payment_transaction.py
```diff
@@ -108,3 +108,16 @@ def _get_communication(self):
...
         return communication or self.reference
+
+    def _get_custom_qr_code(self):
+        """Return the QR code to scan to pay the custom provider's bank account, if any.
+
+        :return: The base64 QR code.
+        :rtype: str
+        """
+        self.ensure_one()
+        provider_sudo = self.provider_id.sudo()
```
vchu-odoo: why do you need `sudo` here?
in `custom_payment_transfer_details` we can directly accept only tx_sudo

addons/website_sale/models/payment_transaction.py
```diff
@@ -20,20 +20,19 @@ def _process(self, payment_data):
...
+        if is_website_order and self.state == "done" and order.amount_total != self.amount:
             return Markup("<p>%s</p>") % self.env._(
                 "Unfortunately your order can not be confirmed as the amount of your payment"
                 " does not match the amount of your cart. Please contact the responsible of"
                 " the shop for more information."
             )
+        if is_website_order and self.state == "pending" and self._requires_payment_instructions():
+            return Markup("<p>%s</p>") % self.env._(
+                "Your order will be confirmed after payment is received."
+            )
```
vchu-odoo: to make conditions more visible
```suggestion
        if  order and order.website_id:
            if self.state == "done" and order.amount_total != self.amount:
                return Markup("<p>%s</p>") % self.env._(
                     "Unfortunately your order can not be confirmed as the amount of your payment"
                     " does not match the amount of your cart. Please contact the responsible of"
                     " the shop for more information."
                 )
            if self.state == "pending" and self._requires_payment_instructions():
                return Markup("<p>%s</p>") % self.env._(
                    "Your order will be confirmed after payment is received."
                )
```

addons/payment_custom/views/payment_provider_views.xml
```diff
@@ -33,23 +33,28 @@
             <group name="provider_config_others" position="attributes">
                 <attribute name="invisible" separator="or" add="code == 'custom'"/>
             </group>
+            <group name="provider_config_others" position="after">
```
vchu-odoo: we can merge this conditions. Instead of hiding the whole `provider_config_others`, we can hide fields included there, except for company id (to fix the bug, could be done is separate commit). And extend `provider_config_others` with the new field for wire transfer.

PR #247313, opened 2026-02-05
[FIX] mail: show thread display name in notification dialog

ramh-odoo:
Before this PR, when a user opened a group chat in Discuss and accessed
the notification settings from the thread actions, the thread name was missing
in the dialog.

This commit fixes the issue by correctly displaying the thread name in
the notification settings dialog.

Part of task-5910210.

---


alexkuhn (approved):
@robodoo delegate+ r+

PR #247314, opened 2026-02-05
[IMP] pos_sale: mention down payment details in invoice

shag-odoo:
Following this commit:
=====
- Down payment details are been mentioned in invoice, if down payment is done on percentage basis then % value is displayed other Down Payment (POS) is mentioned.

task-5896695


davidmonnom (changes requested):
Can you add tests ?

davidmonnom (approved):
@robodoo r+

PR #247320, opened 2026-02-05
[FIX] mail: prevent horizontal scrollbar in messaging menu

nmak-odoo:
**Purpose of this PR:**

Hovering over a muted notification in the messaging menu caused a horizontal scrollbar because the `'Mark as Read'` button retained a margin that overflowed the container when the unread counter was hidden.

This commit makes the margin conditional on the counter's presence, preventing the layout overflow.

task-[5904526](https://www.odoo.com/odoo/project.task/5904526)

**Before/After:**
Before

After


alexkuhn (approved):
@robodoo delegate+ r+

PR #247321, opened 2026-02-05
[FIX] account_peppol: prevent invoice reset to skipped

LouisGobert:
This fix addresses two issues related to resending invoices via Peppol.

It prevents users from accidentally resending an invoice and having its status incorrectly set to “skipped”. It also prevents from resending an invoice via Peppol when it is already in a “processing”, then "skipped" state.

Steps to reproduce:
- Create and send a customer invoice to Peppol
- Try to send it again, Odoo sets the status to “skipped”
- Try to send it again, Odoo resends the invoice via Peppol

This fix is a light adaptation of the `_is_applicable_to_move` method introduced in version 18.

After the fix:
Trying to resend to Peppol an invoice already in "processing" or "done" state is prevented.

opw-5491341

### DIFF
--- a/addons/account_peppol/tests/test_peppol_messages.py
+++ b/addons/account_peppol/tests/test_peppol_messages.py
@@ -300,14 +300,15 @@ class TestPeppolMessage(TestAccountMoveSendCommon):
             checkbox_send_peppol=True,
         )
 
-        wizard.action_send_and_print()
-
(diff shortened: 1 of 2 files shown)


clbr-odoo (approved):
LGTM thanks 👍 
But the fix should probably be targeted to 16.0 🥴

PR #247322, opened 2026-02-05
[FIX] purchase: po user should able to delete bill lines

daiduongnguyen-odoo:
* Problem: when purchase user create bill from PO they only to record bill for some product not all so some move line need to be deleted, but no deleted button appear

* Reason: Purchase User only have read, create, write access to journal items

Current behavior before PR:

---


Whenrow:
@daiduongnguyen-odoo Hello. This is something intended, and we left the database admin tweak the access on his/her side

PR #247330, opened 2026-02-05
[FIX] purchase : fix POL Vendor Code/Name in desc

MAGREMENT:
How to reproduce :
- Create a product
- Create an attribute with a value set to "free text"
- Add 2 different vendors to the purchase tab of the product with different Vendor Code and/or Vendor Name
- Put Dropship as the route for the product to create a PO on a SO confirmation
- Create a SO with that product
- Confirm the SO and go to the PO
- Change the vendor in the SO

The problem :
The Vendor Code and/or Vendor Name does not change correctly in the product description

Why :
The way the description generation for a change in a purchase order line works as follows : Create a collection of default descriptions based on the product and the different vendors. If the collection contains the current description, it means the description was not changed by the user and it can be modified. This is done to prevent resetting a custom description made by an user.

This logic is pretty broken because the code that generate the description when going from a PO to a SO and the code that generate the description when changing the PO partner are different. This makes it so the current description is almost never found in the list of default descriptions.

It looks tempting to try to "resynchronize" those part of the code so the description generation is the same in both cases. This would be done by overriding the _get_product_purchase_description method in the purchase_stock module and using the product_description_variants field. Sadly, this does not work because it conflicts with the purchase_product_matrix module and its own override of that method.

The fix that I implemented checks that the Vendor Code and Name are the right one every time the description is computed following a change.

opw-5888233

---

### DIFF
--- a/addons/purchase/models/purchase_order_line.py
+++ b/addons/purchase/models/purchase_order_line.py
@@ -325,6 +325,30 @@ class PurchaseOrderLine(models.Model):
             if seller or not line.date_planned:
                 line.date_planned = line._get_date_planned(seller).strftime(DEFAULT_SERVER_DATETIME_FORMAT)
 
+            # record product names to avoid resetting custom descriptions
+            default_names = []
+            display_names = []
+            vendors = line.product_id._prepare_sellers(params=params)
+            product_ctx = {'seller_id': None, 'partner_id': None, 'lang': get_lang(line.env, line.partner_id.lang).code}
+            line_without_seller = line.product_id.with_context(product_ctx)
+            default_names.append(line._get_product_purchase_description(line_without_seller))
+            for vendor in vendors:
+                product_ctx = {'seller_id': vendor.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
+                default_names.append(line._get_product_purchase_description(line.product_id.with_context(product_ctx)))
+                display_names.append(line.product_id.with_context(product_ctx).display_name)
+            if not line.name or line.name in default_names:
+                product_ctx = {'seller_id': seller.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
+                line.name = line._get_product_purchase_description(line.product_id.with_context(product_ctx))
+            else:
+                # Checks that the product vendor and vendor name are correct
+                for vendor, display_name in zip(vendors, display_names):
+                    if line.name.startswith(display_name):
+                        if not seller:
+                            line.name = line_without_seller.display_name + line.name[len(display_name):]
+                        elif vendor.id != seller.id:
+                            line.name = display_names[vendors.ids.index(seller.id)] + line.name[len(display_name):]
+                        break
+
             # If not seller, use the standard price. It needs a proper currency conversion.
             if not seller:
                 line.discount = 0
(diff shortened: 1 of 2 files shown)


ajf-odoo (changes requested):
Your test is not sufficient,
I would make the po on self.self.partner with no name on the line
then change to self.partner_a then self.partner_b then back to self.partner
This way, your (adapted) solution will cover all the cases

addons/purchase/models/purchase_order_line.py
```diff
@@ -369,6 +369,17 @@ def _compute_price_unit_and_date_planned_and_name(self):
...
+            for vendor in vendors:
+                product_ctx = {'seller_id': vendor.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
+                line_with_vendor_context = line.product_id.with_context(product_ctx)
+                if line.name.startswith(line_with_vendor_context.display_name):
+                    if vendor.id != seller.id:
+                        product_ctx = {'seller_id': seller.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
+                        line_with_seller_context = line.product_id.with_context(product_ctx)
+                        line.name = line_with_seller_context.display_name + line.name[len(line_with_vendor_context.display_name):]
+                    break
+
```
ajf-odoo: starting at line 361, why not something like
```
            default_names = []
            display_names = []
            vendors = line.product_id._prepare_sellers(params=params)
            product_ctx = {'seller_id': None, 'partner_id': None, 'lang': get_lang(line.env, line.partner_id.lang).code}
            product_id_with_ctx = line.product_id.with_context(product_ctx)
            default_names.append(line._get_product_purchase_description(product_id_with_ctx))
            display_names.append(product_id_with_ctx.display_name)
            for vendor in vendors:
                product_ctx = {'seller_id': vendor.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
                product_id_with_ctx = line.product_id.with_context(product_ctx)
                default_names.append(line._get_product_purchase_description(product_id_with_ctx))
                display_names.append(product_id_with_ctx.display_name)
            if not line.name or line.name in default_names:
                line.name = default_names[vendors.ids.index(seller.id)+1]
            else:
                # Checks that the product vendor and vendor name are correct
                for index, vendor in enumerate(vendors):
                    if line.name.startswith(display_names[index+1]):
                        if vendor.id != seller.id:
                            line.name = display_names[vendors.ids.index(seller.id)+1] + line.name[len(display_names[index+1]):]
                        break

```
seems simpler:
all data built in 1st loop
take advantage of indexing
clearer line.name updating
to be checked with @adwid

addons/purchase/models/purchase_order_line.py
```diff
@@ -359,15 +359,25 @@ def _compute_price_unit_and_date_planned_and_name(self):
...
                 product_ctx = {'seller_id': seller.id, 'lang': get_lang(line.env, line.partner_id.lang).code}
                 line.name = line._get_product_purchase_description(line.product_id.with_context(product_ctx))
+            else:
+                # Checks that the product vendor and vendor name are correct
+                for index, vendor in enumerate(vendors):
+                    display_name = display_names[index]
+                    if line.name.startswith(display_name):
+                        if vendor.id != seller.id:
+                            line.name = display_names[vendors.ids.index(seller.id)] + line.name[len(display_name):]
+                        break
```
Whenrow:
```suggestion
                for vendor, display_name in zip(vendors, display_names):
                    if line.name.startswith(display_name):
                        if vendor.id != seller.id:
                            line.name = display_name + line.name[len(display_name):]
                        break
```

MAGREMENT: This does not work because I need the display_name of the current seller, not of the current vendor in the loop
```py
display_names[vendors.ids.index(seller.id)]
```

Whenrow: yes sorry I made this suggestion a bit to fast but we can keep the zip to avoid messing with indexes too much
```python
for vendor, display_name in zip(vendors, display_names):
    if line.name.startswith(display_name):
        if vendor.id != seller.id:
            line.name = display_names[vendors.ids.index(seller.id)] + line.name[len(display_name):]
        break
```

addons/purchase/tests/test_purchase.py
```diff
@@ -1177,3 +1177,67 @@ def test_product_price_on_purchase_order_view_catalog(self):
...
+                **supplierinfo_vals,
+                'price': 20,
+                'product_name': 'Name 2',
+                'product_code': 'Code 2',
+                'partner_id': self.partner_b.id,
+            },
+            {
+                **supplierinfo_vals,
+                'price': 15,
+                'partner_id': self.partner.id,
```
ajf-odoo, line 1207: test fails if vendor not in sellers...

PR #247334, opened 2026-02-05
[IMP] l10n_my_edi_pos: Add EDI fields to PoS

jand-odoo:
Add EDI fields (Identification and Industrial Classification) to the partner form in Point of Sale to support Malaysian EDI requirements.

task-5912243


addons/l10n_my_edi_pos/views/res_partner_view.xml
hupo-odoo, line 1: I'm pretty sure you should make it invisible if the company' fiscal country code isn't `MY`.

jand-odoo: Thanks for the catch! I have updated the code :))

hupo-odoo (approved):
@robodoo r+

PR #247339, opened 2026-02-05
[IMP] l10n_hr_edi: Improvements to XML generation, test coverage, API

nape-odoo:
- Improving the ubl_hr method chain to include less overrides in favor of extensions
- Adding proper handling of multiple tax total and tax category nodes for hrextac
- Refining docstrings and comments for ubl_hr
- Adding a test with multiple tax types
- Adding a test for account_edi_ubl_cii_tax_extension integration
- Adjusting MER methods to correspond to API changes: fiscalization queries are now only available for own documents
task-none

---

### DIFF
--- a/addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
+++ b/addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
@@ -168,27 +168,45 @@ class AccountEdiXmlUBLHR(models.AbstractModel):
             })
 
     def _add_hr_extension_node(self, document_node):
-        tax_subtotals = document_node['cac:TaxTotal'][0]['cac:TaxSubtotal']
-        hr_tax_subtotals = []
+        """
+        This function constructs hrextac node from existing data within the document.
+        The structure mostly follows that of 'cac:TaxTotal' node of a UBL 2.1/BIS 3 document,
+        but requires additional data compared to the totals/subtotals nodes in UBL HR format.
+        To avoid making additional queries and possible desyncs, we calculate all the data
+        we need while assembling normal subtotals, then trim out the extra bits.
+        """
         cash_basis_line = False
-        for i in range(len(tax_subtotals)):
-            # Removing the HR-specific node from the normal subtotal where we calculate it
-            hr_tax_name = tax_subtotals[i]['cac:TaxCategory'][0].pop('cbc:Name')
-            cash_basis_flag = tax_subtotals[i]['cac:TaxCategory'][0].pop('hrextac:HRObracunPDVPoNaplati')  # Ensure pop() always runs
-            cash_basis_line = cash_basis_line or cash_basis_flag
-            new_item = {
-                'cbc:TaxableAmount': tax_subtotals[i]['cbc:TaxableAmount'],
-                'cbc:TaxAmount': tax_subtotals[i]['cbc:TaxAmount'],
-                'hrextac:HRTaxCategory': {
-                    'cbc:ID': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:ID'],
-                    'cbc:Name': hr_tax_name,
-                    'cbc:Percent': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:Percent'],
-                    'cbc:TaxExemptionReasonCode': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReasonCode'],
-                    'cbc:TaxExemptionReason': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReason'],
-                    'hrextac:HRTaxScheme': tax_subtotals[i]['cac:TaxCategory'][0]['cac:TaxScheme'] if hr_tax_name['_text'] != "HR:POVNAK" else {'_text': "OTH"},
-                }
-            }
-            hr_tax_subtotals.append(new_item)
+        tax_totals = document_node['cac:TaxTotal']
+        hr_tax_totals = []
+        for total in tax_totals:
+            tax_subtotals = total['cac:TaxSubtotal']
+            hr_tax_subtotals = []
+            for subtotal in tax_subtotals:
+                tax_categories = subtotal['cac:TaxCategory']
+                hr_tax_categories = []
+                for category in tax_categories:
+                    # Cash basis is document-wide, so we do not need to keep it for each category
+                    cash_basis_flag = category.pop('hrextac:HRObracunPDVPoNaplati')  # Ensure pop() always runs
+                    cash_basis_line = cash_basis_line or cash_basis_flag
+                    # Removing the HR-specific node from the normal subtotal where we calculate it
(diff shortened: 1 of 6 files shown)


antoine162 (approved):
Hi @nape-odoo this all seems good, just there seems to be a merge conflict that needs to be resolved 😄

nape-odoo:
@antoine162 Don't r+ this yet if you happen to look at it, I'm adding another fix to this PR.

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -168,27 +168,45 @@ def _add_invoice_header_nodes(self, document_node, vals):
...
-                    'cbc:Percent': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:Percent'],
-                    'cbc:TaxExemptionReasonCode': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReasonCode'],
-                    'cbc:TaxExemptionReason': tax_subtotals[i]['cac:TaxCategory'][0]['cbc:TaxExemptionReason'],
-                    'hrextac:HRTaxScheme': tax_subtotals[i]['cac:TaxCategory'][0]['cac:TaxScheme'] if hr_tax_name['_text'] != "HR:POVNAK" else {'_text': "OTH"},
-                }
-            }
-            hr_tax_subtotals.append(new_item)
+        tax_totals = document_node['cac:TaxTotal']
+        hr_tax_totals = []
+        for i_totals in range(len(tax_totals)):
```
antoine162: Using indices here (`i_totals`, `i_subtotals`, and `i_categories`) seems unnecessary to me.

Since you're only using them to iterate over lists, you might as well just iterate over the lists directly.

So your for loop becomes `for totals_node in tax_totals` and you can replace `tax_totals[i_totals]` with `totals_node`.

Same for the other things you are iterating on.

addons/l10n_hr_edi/models/account_move.py
```diff
@@ -206,21 +206,29 @@ def l10n_hr_edi_mer_action_fetch_status(self):
         self.ensure_one()
         if self.is_sale_document():
             response_mer = _mer_api_query_document_process_status_outbox(self.company_id, electronic_id=self.l10n_hr_mer_document_eid)[0]
-            response_fisc = _mer_api_check_fiscalization_status_outbox(self.company_id, electronic_id=self.l10n_hr_mer_document_eid)[0]
+            response_fisc = _mer_api_check_fiscalization_status_outbox(self.company_id, electronic_id=self.l10n_hr_mer_document_eid)
+            if isinstance(response_fisc, list):
```
antoine162, line 210: All I can say when I see this kind of API variation is ... 🙈 

Just can you confirm for sure that the API sometimes returns a (potentially empty or singleton) list and sometimes a dict? And that both still happen and it's not that the API has switched completely to returning a list?

Maybe you could send an e-mail to MER as well to confirm what happened (though I won't delay merging to wait for them) 😄

I also think that if there is a lot of variation in responses, it could be worth adding extra mocked tests with those responses. But that could be done in another PR if we need to first merge this to unblock users quickly.

addons/l10n_hr_edi/models/account_edi_xml_ubl_hr.py
```diff
@@ -214,10 +229,12 @@ def _add_hr_extension_node(self, document_node):
         })
 
     def _add_invoice_line_item_nodes(self, line_node, vals):
+        """
+        HR-BR-25: Each item MUST have an item classification identifier from the Classification of Products
+        by Activities scheme: KPD (CPA) - listID "CG", except in the case of advance payment invoices.
+        """
```
antoine162: Honestly, personally I wouldn't making these "cosmetic" changes - it doesn't change much whether this info is in a comment or a docstring. Same for the indent changes or modifying dict key assignment to `dict.update()` - for these things where both versions are correct there isn't really any point in changing from one to the other IMO 😄

antoine162 (commented):
Thanks so much for those improvements 😄 
I just had a few style comments + I'd appreciate if you can confirm the API change 😄

addons/l10n_hr_edi/models/account_move.py
nape-odoo: All the API info is here: https://manual.moj-eracun.hr/documentation/getting-started/api-specification-2/
There are no detailed response format explanations, only some examples. For fiscalization outbox specifically, the example makes it appear that it should always be a list of dicts (or a list of a single dict), the rules of when a user is allowed to request a document's fiscalization status have changed - now user can only request fisc info for actions initiated by them - so I can't always expect it to return a non-empty response. This is also the reason why I changed fiscalization checking for importing documents.

As for asking MER directly, I think this is PO's job rather than mine, and any significant decisions I may have to make in relation to how the module works would have to be confirmed with someone else anyway, so it won't make things any faster either.

antoine162: I guess my question is more about why we need to handle both the dict and list of dicts case?
If MER always returns a list of dicts, can't we remove the `if isinstance(response_fisc, list)`?

As for knowing the API we're working with, that's definitely our job as application developers rather than the PO's. :smile: 
If the API doc is clear and there's no reason to believe that there's any deviation from it, then there's no reason to double check with MER. But in that case does that mean that the previous code never worked? I just want to be sure I understand what we're trying to fix.

nape-odoo: Current API related changes are based on these two tracebacks: https://www.prodaja.eu/documents/content/RG5Tbr37RgOUYXbxRbYJ7Qo32a?download=0 https://karambol.odoo.com/documents/content/gES2uyxtQSOC2fRbreLLfAo1e?download=0

There was a separate error that appears to correspond to MER returning a dict (that isn't interpreted as an error dict by the caller function) instead of a list of dicts for a different call (statusInbox), which I am unable to reproduce on the demo server for a non-single-document request, but I suspect this is what is happening when there's only a single document there, as the example shown is a single document dict not wrapped in a list. I suspect that similar things might happen with fisc requests as well but those are even harder to check on the demo server since now they depend on what user does what with the document and on the demo server we send invoices to ourselves so our credentials match both ends. Overall, this check is a precaution rather than a necessity.

Previous code worked as far as 2p-iskaro and I were able to tell, for how fiscalization API worked in December and January. Since then, MER API underwent a number of changes that I've been adjusting the code for; unfortunately they don't document the changes made very well - sometimes there were "Updated [date]: done such and such" notes in the text, but now all of these are gone from the page and instead we have a "Changelog" button that doesn't actually provide any details when interacted with.

antoine162: Okay super, thanks so much for the explanation. 
I understand that most of the difficulty comes from them just not providing a stable API.
I'll write to them on Monday (I'm out today) to tell them about this problem and ask them to preferably not make changes to the API but if they need to, to let us know ahead of time.

antoine162 (approved):
Thanks so much!
@robodoo r+

PR #247340, opened 2026-02-05
[IMP] fleet: add multi-company support to vehicle models

ramezlahzy:
Add a company_id field to fleet.vehicle.model to allow restricting vehicle models to specific companies. When empty, the model is available to all companies (default for existing records).

Changes:
- Add company_id field to fleet.vehicle.model
- Add multi-company record rule
- Update form, list, and search views with company field

Task id: 5906191


YassinWalid (approved):
LGTM 🚀

PR #240603, opened 2025-12-18
[FIX] base: rename inherited custom field

kmagusiak:
When renaming a custom field that is inherited, first remove the generated inherited fields from the database.

task-5172546

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247344, opened 2026-02-05
[FIX] base: rename inherited custom field - backport

kmagusiak:
When renaming a custom field that is inherited, first remove the generated inherited fields from the database.

task-5172546
backport of #240603

---


kmagusiak:
@robodoo up to saas-19.1

Julien00859 (approved):
@robodoo r+

PR #247346, opened 2026-02-05
[FIX] mail: link attachments to record in mass mail

delcourtfl:
**Steps to reproduce:**
- Install Event app
- Log as a user with Events/Administrator rights
- Ensure its Administration rights are not set to Settings
- Go to an open Event
- Add a new attendee with mail
- Ticket is automatically sent, and logged in chatter
- User doesn't have the rights to download the file from the chatter

**Issue:**
User has no 'read' access to `ir.attachment` due to a missing related record (res_model
and res_id are set to "mail.message" and 0 instead of the `event.registration` record).

It can be bypassed when the user also has system rights (or is the sender) and
it works properly when sending manually the mail.

**Fix:**
Reapply the fix used in 19.0+.
(We could also hide the attachment in chatter when it is not linked properly)

related: https://github.com/odoo/odoo/commit/19c78b2585cec9473ad2d2ab9101235c45a03314

opw-5349820

### DIFF
--- a/addons/mail/wizard/mail_compose_message.py
+++ b/addons/mail/wizard/mail_compose_message.py
@@ -1099,7 +1099,10 @@ class MailComposer(models.TransientModel):
                 mail_values['attachment_ids'] = process_record._process_attachments_for_post(
                     decoded_attachments,
                     attachment_ids,
-                    {'model': 'mail.message', 'res_id': 0}
+                    {'model': 'mail.message', 'res_id': 0} if (
+                        not hasattr(record, "_process_attachments_for_post")
+                        or (self.auto_delete and not self.auto_delete_keep_log)
+                    ) else {}  # link to record if kept in chatter, this ensures users can download it
                 )['attachment_ids']
             # comment mode: prepare attachments as a list of IDs, to be processed by MailThread
             else:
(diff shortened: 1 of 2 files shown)


Ephoniz (approved):
LGTM 👍🏽

addons/mail/wizard/mail_compose_message.py
```diff
@@ -1099,7 +1099,10 @@ def _prepare_mail_values_dynamic(self, res_ids):
                 mail_values['attachment_ids'] = process_record._process_attachments_for_post(
                     decoded_attachments,
                     attachment_ids,
-                    {'model': 'mail.message', 'res_id': 0}
+                    {'model': 'mail.message', 'res_id': 0} if (
+                        not hasattr(record, "_process_attachments_for_post")
+                        or (self.auto_delete and not self.auto_delete_keep_log)
+                    ) else {}  # link to record if kept in chatter, for ease of access
```
reth-odoo: Maybe update to also say it's necessary for access at all then

reth-odoo (commented):
I guess this is a relatively big change in behavior then if none of the reports were available to regular users 🤔 
Although it makes sense, I'm not totally sure we actually want to change that in stable.. Maybe the move is to say it's fixed in 19. We'll see what tde says I guess, small comment

tde-banana-odoo:
@robodoo r+

Indeed, wondering potential functional changes this may imply ... but anyway we should avoid access issues for generated attachments, notably like the one reproduced here.

Note that we could link the attachments to the actual mail.message, which means access would depend on mail.message, but this would imply a post processing (creating the message, then setting the id to the message ID). And it also rely on related record anyway, so ... seems good to me.

PR #247347, opened 2026-02-05
[FIX] point_of_sale: prevent combo price from being doubled

hadielyakhni:
Combo product prices were doubled in the `comboTotalPrice` and
`comboTotalPriceWithoutTax` getters, as we were summing the
`displayPrice` of all the combo lines, including the parent line, which
already had its `displayPrice` as the sum of its children. So now, we
filter out the parent line in those getters before summing.


davidmonnom (approved):
@robodoo r+

PR #247352, opened 2026-02-05
[IMP] mail: improve notification settings thread action

ramh-odoo:
Before this PR, opening the notification settings from the thread action (In Discuss App) displayed them in a dialog, which made the UI less consistent with other thread actions.

After this commit, opening the notification settings from the thread action displays them in a dropdown instead of a dialog, providing a more consistent and streamlined user experience.

Task-5936685

---

### DIFF
--- a/addons/mail/static/src/discuss/core/common/notification_settings.xml
+++ b/addons/mail/static/src/discuss/core/common/notification_settings.xml
@@ -2,7 +2,7 @@
 <templates xml:space="preserve">
 
     <t t-name="discuss.NotificationSettings">
-        <ActionPanel t-if="ui.isSmall" title.translate="Notification Settings" resizable="!!env.inMeetingView" initialWidth="400" icon="'fa fa-bell'">
+        <ActionPanel t-if="!env[DROPDOWN_NESTING] and env.inChatWindow" title.translate="Notification Settings" resizable="!!env.inMeetingView" initialWidth="400" icon="'fa fa-bell'">
             <t t-call="discuss.NotificationSettings.content"/>
         </ActionPanel>
         <t t-else="" t-call="discuss.NotificationSettings.content"/>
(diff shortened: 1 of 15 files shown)


addons/mail/static/src/discuss/core/common/thread_actions.js
```diff
@@ -94,21 +94,11 @@ registerThreadAction("notification-settings", {
...
-                channel,
-            });
-        }
-    },
-    actionPanelOuterClass: "bg-100 border border-secondary",
+    dropdown: true,
+    dropdownTemplate: "discuss.NotificationSettings",
+    dropdownComponent: NotificationSettings,
+    dropdownComponentProps: ({ channel }) => ({ channel }),
+    dropdownPosition: ({ owner }) => owner.isDiscussContent ? "bottom-end" : undefined,
```
alexkuhn: I would remove the title and icon. "Notifications Settings" is already clear from the 1st item.
Image

For non-channels, there's only the "Mute" feature. I would rename the item to "Mute conversation" in the sidebar action for them and immediately show the dropdown selection of the timings.

addons/mail/static/src/discuss/core/common/thread_actions.js
```diff
@@ -94,21 +94,11 @@ registerThreadAction("notification-settings", {
...
-                hasSizeConstraints: true,
-                channel,
-            });
-        }
-    },
-    actionPanelOuterClass: "bg-100 border border-secondary",
+    dropdown: true,
+    dropdownTemplate: "discuss.NotificationSettings",
+    dropdownComponent: NotificationSettings,
+    dropdownComponentProps: ({ channel }) => ({ channel }),
```
alexkuhn: `dropdownTemplate` and `dropdownComponent` are mutually exclusive. `dropdownComponent` has precedence over `dropdownTemplate`, so `dropdownTemplate` here is useless.

These are just 2 ways to provide content of dropdown, whether as a template or as a component.

addons/mail/static/src/discuss/core/common/notification_settings.xml
```diff
@@ -18,7 +18,7 @@
                             <div class="d-flex flex-grow-1 align-items-center px-0 py-1 w-100 rounded">
                                 <span class="text-wrap text-start text-break">Mute Conversation</span>
                                 <div class="o-discuss-NotificationSettings-separator flex-grow-1"/>
-                                <i class="oi oi-arrow-right ms-2 me-1"/>
+                                <i t-if="ui.isSmall" class="oi oi-arrow-right ms-2 me-1"/>
```
alexkuhn: Could use `t-if="!env[DROPDOWN_NESTING]"`, for exact condition to not being redundant with the caret from nested dropdown.

I would take the opportunity to change icon to `fa fa-caret-right fa-lg mx-2` for consistent visual in both cases.

addons/mail/static/src/discuss/core/common/thread_actions.js
alexkuhn: To do so, I would split the "content" part of `NotificationSettings` in its own component, so that we don't have to deal with `<ActionPanel>` with title / icon / extra padding and so on.

alexkuhn (approved):
@robodoo delegate+ r+ rebase-ff

PR #247353, opened 2026-02-05
[FIX] loyalty: remove duplicated gift card

lkerroum:
Current behavior before PR: When creating a database, two gift card products appear in eCommerce>Products. This happens because the link between the reward record and a product was missing, which caused Odoo to auto-generate a duplicate, visible product for the reward.

In this commit, we explicitly linked the gift card reward to the gift card product, preventing the creation of an unwanted "ghost" product in the shop.

task-5865210


livaios (approved):
Hello @lkerroum,

Thanks for investigating and creating the fix
LGTM 👍

PR #247355, opened 2026-02-05
[IMP] account: show that invoice_currency_rate is rounded if rounded

Arnaud-Sibille:
task-5231229


Arnaud-Sibille:
Task canceled

PR #247368, opened 2026-02-05
[FIX] website: prevent badge option merging issues

sben-odoo:
Steps to reproduce:
- Go to Website > Edit a page
- Add a badge using editor
- Duplicate that badge more than once
- Save

-> badges intersect each other

Cause:
======
When saving, `cleanForSave` will be triggered which then will call `mergeAdjacentInlines`, since badges are inline elements, they will be merged into one single badge causing that issue.

Solution:
=========
We add a predicate to `unsplittable_node_predicates` to prevent `mergeAdjacentInlines` from merging badges.

opw-5790891

---


nle-odoo (approved):
make sense to me

note: a similar `shouldBeMergedWithPreviousSibling` exists in 18.0 but it doesn't seem to be run (if I reproduce the situation of 18.4 copy by modifying the DOM). I guess this is a difference between 18.4 and 18.0 version.

PR #247371, opened 2026-02-05
[IMP] mail: suggest unmuting when speaking while muted in calls

pmah-odoo:
When a user speaks while their microphone is muted in a Discuss call or meeting, no visual indication or suggestion is shown to inform them that they are muted.

When a user tries to speak while muted during a Discuss call or meeting, a contextual muted suggestion appears near the mute button, prompting them to unmute before continuing.

In addition, these changes introduces a call suggestion registry and a shared suggestion tooltip to centralize call-related suggestions and reuse the same implementation for existing microphone warnings.

image

task-[4931425](https://www.odoo.com/odoo/project/1519/tasks/4931425)

---


addons/mail/static/src/discuss/call/common/call_mute_suggestion.js
```diff
@@ -0,0 +1,97 @@
...
+            timers.display = null;
+            timers.cooldown = setTimeout(() => {
+                timers.cooldown = null;
+            }, MUTE_SUGGESTION_CONFIG.cooldownDuration);
+        }, MUTE_SUGGESTION_CONFIG.activeDuration);
+    }
+
+    function showSuggestion() {
+        timers.arming = null;
+        const target = document.querySelector(
```
maad-odoo: why can't we use `t-custom-ref` to narrow down the search of the mute button ?

addons/mail/static/src/discuss/call/common/call_suggestion_tooltip.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+    <t t-name="mail.CallSuggestionTooltip">
+        <div class="d-flex align-items-center py-1 px-2 gap-2" style="max-width: 210px;">
+            <span t-out="props.text"/>
+            <i class="fa fa-times cursor-pointer opacity-50 opacity-100-hover flex-shrink-0" aria-label="Dismiss" t-on-click="props.onDismiss"/>
```
maad-odoo: wrap this in button instead

pmah-odoo: wrapping this in button takes some extra space which makes UI look bad IMO and to remove that space we need to sprinkle some extra classes to achieve same result at the end so personally i'd prefer to keep it as it is.

addons/mail/static/src/discuss/call/common/call_mute_suggestion.js
```diff
@@ -0,0 +1,97 @@
...
+            ".o-discuss-CallActionList .o-discuss-callAction-Mute"
+        );
```
ramh-odoo: Using `document.querySelector` to parse the whole document for showing the suggestion doesn’t look ideal to me. It would be cleaner and more reliable to use a custom ref instead of querying the DOM globally.

With the current approach, we depend on a CSS selector that could easily break during refactors or template changes. It also makes the logic less explicit, as the connection between the component and the target element is indirect. Using a ref would keep the relation localized, improve maintainability, and better align with Owl/component-based patterns.

pmah-odoo: Though it's not ideal and I'm not denying that, considering the feature and the extra code complexity that targeting the element with a `ref` would add, IMO I'd stick with the current `querySelector` approach. especially given that this element needs to be targeted not so frequently.

ramh-odoo: We can now use resources for multiple references in Owl 3, so accessing the target element through refs should be straightforward and shouldn't add much complexity. 
see: [Multiple References with Resources](https://github.com/odoo/owl/blob/master/doc/v3/owl/reference/refs.md#multiple-references-with-resources)

pmah-odoo: Indeed, now we can, thanks to Owl 3. I'll adapt the changes accordingly.

addons/mail/static/src/discuss/call/common/call_suggestion_tooltip.js
```diff
@@ -0,0 +1,13 @@
...
+
+export class CallSuggestionTooltip extends Component {
+    static template = "discuss.CallSuggestionTooltip";
+    props = props({
+        id: types.string(),
+        iconClass: types.string(),
+        headerText: types.string(),
+        bodyText: types.string(),
+        onDismiss: types.function([]),
+        close: types.function([]),
```
ramh-odoo: Why do we need to pass both? Can't `onDismiss` be handled inside `close` instead?

addons/mail/static/src/discuss/call/common/call_suggestions.js
```diff
@@ -0,0 +1,192 @@
...
+ * @property {import("@mail/discuss/call/common/rtc_service").CallSuggestion} suggestion - The suggestion object.
+ */
+
+/**
+ * @param {string} id - Unique identifier for the suggestion.
+ * @param {CallSuggestionDefinition} definition
+ */
+export function registerCallSuggestion(id, definition) {
+    callSuggestionsRegistry.add(id, definition);
+}
```
ramh-odoo, line 42: I'm wondering what the motivation was for moving the call suggestions to a registry, since they seem to be used only in calls.

addons/mail/static/src/discuss/call/common/call_actions.js
```diff
@@ -53,17 +53,17 @@ export const muteAction = {
...
             });
             useEffect(() => {
-                const hasWarning =
-                    store.rtc.showMicrophonePermissionWarning ||
-                    store.rtc.showMicrophoneSilentWarning;
-                if (!action.popover.isOpen && action.actionRef() && hasWarning) {
-                    action.popover.open(action.actionRef(), {});
+                const suggestion =
+                    store.rtc.visibleCallSuggestion?.targetActionId === action.id &&
+                    store.rtc.visibleCallSuggestion;
```
ramh-odoo: This pattern is a bit hard to read because it relies on `&&` returning either the suggestion object or `false`. Wouldn't a ternary make the intent clearer?

addons/mail/static/src/discuss/call/common/call_suggestion_tooltip.js
pmah-odoo: If you suggest to put `onDismiss` in the `onClose` callback, then it would change the behavior.

`onClose` is called whenever the popover is being closed, but closing the popover does not necessarily mean the user wants to dismiss the suggestion. Some suggestions keep a dismissed state, and for that we expect the user to explicitly click the `X`.

So I would keep both separate.

addons/mail/static/src/discuss/call/common/call_suggestions.js
pmah-odoo: The previous two suggestions were already split between `rtc` and `call_actions`, and adding this new one there would make that code messy and harder to maintain. Since they are all call suggestions, I thought it was cleaner to keep them in one place and generalize the logic so new suggestions are easier to add later.

addons/mail/static/src/discuss/call/common/call_suggestions.js
```diff
@@ -0,0 +1,192 @@
...
+                rtc.microphonePermission === "granted" &&
+                session.is_muted &&
+                session.isTalking &&
+                !timers.arming
+            ) {
+                timers.arming = setTimeout(showSuggestion, MUTE_SUGGESTION_CONFIG.armingDelay);
+            } else {
+                clearTimeout(timers.arming);
+                timers.arming = null;
+            }
```
alexkuhn, line 113: The heuristic is too relax, and would show the popup even for false positives.
The cooldown of 60 seconds is quite long, so you could have it triggered accidentally while coming back to your room and sitting on your chair, putting the headphone on your head, and then start talking, only to not have the popout showing it.

I think that we should have a greater window of someone continuously talking, let's say 5 seconds, and make an average that about 80% of the time the person was talking (to take into account some blanks while speaking). Don't know how feasible but it would be nice to detect whether frequency is around the human voice to dismiss some noises, and also take proximity of the voice from the microphone into through volume, to dismiss chatting with someone else IRL in the same room but with distance from the microphone.

And when we do all of this, I think the cooldown of popup to appear again should be extremely short and the delay to automatically hide it should be much greater than 6 seconds, should be like 20 seconds or so, as the main ways to remove the popup is user solves with unmute or manual click on "x".

Also I think it would be nice to have a button "Don't show me again", as I see quite a lot of people complaining about this feature on Google Meet.

addons/mail/static/src/discuss/call/common/call_suggestions.js
alexkuhn, line 1: I don't know how you make this code, but I suspect AI to copy the shape of `registerAction()` we have.

This feels like copying the code for the sake of it: the reason we have a `registerAction()` architecture is because we have lots of actions in Discuss with their own subtleties but we want to harmonize the visual and UX as much as possible.

Here with the `CallSuggestion` this is extremely niche for the mute button and the showing of the popup. This looks very extreme to define a `registerCallSuggestion()` with lots of params as if this would scale as much as `registerAction()`.

My quick glance of code is that this is good that popover is plugged to mute button and it links to the component, but this could only be a plain state machine with an enum and tooltip's content is based around that. The `register()` function with the many typedef and props is very extreme and just makes code harder than it needs + not flexible enough for future needs.

pmah-odoo: Just to clarify, this wasn't AI, It was me! I was inspired by the existing registerAction() pattern and thought a similar registry approach would keep all call suggestions in one place :)

That said, I do take your point that the use case here is niche enough that a full registerCallSuggestion might be an overkill. I'll rework the approach. Thanks for the feedback! :smiley_cat:

alexkuhn: Ah oops sorry for the wrong assumption! 🥲

I'm not against something like that at some point if we feel that this is a frequent piece of code with minor logic and we want to harmonize the UI / UX for all of them, but I prefer to not make it now if not motivated enough, and if we eventually do something like this I would prefer to harmonize further with `registerAction()` API:
I'm trying my best to avoid additional features in `registerAction()`, we have already too many props, and improvements is making each prop more powerful but still in a way to ease UI / UX consistency.

TBH I'm still not happy with `registerAction()` right now. I'm happy that the UI / UX is more consistent thanks to them, but this needs more cleaning to be more natural and flexible to use + much less cumbersome stuffs and hacks to fight against them (like popovers with action panels, this definitely could be better).

Anyway, for your PR I think better keep code very specific for the need of this PR, as I don't think we'd have much more `CallSuggestion` in the future.

PR #247378, opened 2026-02-05
[IMP] sale_gelato: allow using PDF file as Gelato print file

xavieralt:
When Gelato print a Mug or a Tshirt, using an image (PNG or JPEG) as source "print file" could introduce slight difference of colors.

This commit allow using a PDF as "print file" for better color accuracy.

Task-Id: 4637331

---

### DIFF
--- a/addons/sale_gelato/views/product_document_views.xml
+++ b/addons/sale_gelato/views/product_document_views.xml
@@ -8,7 +8,8 @@
         <field name="arch" type="xml">
             <form>
                 <sheet>
-                    <field name="datas" widget="image" nolabel="True"/>
+                    <field name="datas" widget="image" nolabel="True"
+                           options='{"accepted_file_extensions": "application/pdf,image/png,image/tiff,image/svg+xml,image/jpeg"}'/>
                 </sheet>
             </form>
         </field>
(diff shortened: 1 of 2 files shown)


addons/sale_gelato/views/product_document_views.xml
```diff
@@ -8,7 +8,8 @@
         <field name="arch" type="xml">
             <form>
                 <sheet>
-                    <field name="datas" widget="image" nolabel="True"/>
+                    <field name="datas" widget="image" nolabel="True"
+                           options='{"accepted_file_extensions": "image/*,application/pdf"}'/>
```
anko-odoo: I am not sure if we should allow all image types, Gelato only supports  PNG, TIFF, SVG, JPEG image types and PDFs.

anko-odoo (changes requested):
Hello, sorry for the long wait, the changes seem to be working as expected, just tiny change for the image types.

addons/sale_gelato/views/product_document_views.xml
xavieralt: @anko-odoo was using `image/*` as it's the default behavior of the `image` widget, but you're right so I've limited the image formats to only those supported by Gelato.

anko-odoo (approved):
@robodoo r+

PR #247380, opened 2026-02-05
[FIX] loyalty: add (copy) on program duplication

lkerroum:
Currently, we don't have any visual indicator when duplicating a loyalty program. 
After this commit, when a loyalty program is duplicated "(copy)" it's added to its name.

task-5887481

### DIFF
--- a/addons/loyalty/models/loyalty_program.py
+++ b/addons/loyalty/models/loyalty_program.py
@@ -678,3 +678,11 @@ class LoyaltyProgram(models.Model):
                 del vals['trigger_product_ids']
 
         return super().create(vals_list)
+
+    def copy_data(self, default=None):
+        default = dict(default or {})
+        vals_list = super().copy_data(default=default)
+        if 'name' not in default:
+            for program, vals in zip(self, vals_list):
+                vals['name'] = _("%s (copy)", program.name)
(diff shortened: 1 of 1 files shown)


addons/loyalty/models/loyalty_program.py
```diff
@@ -678,3 +678,11 @@ def create(self, vals_list):
...
 
         return super().create(vals_list)
+
+    def copy_data(self, default=None):
+        default = dict(default or {})
+        vals_list = super().copy_data(default=default)
+        for program, vals in zip(self, vals_list):
+            if 'name' not in default:
+                vals['name'] = _("%s (copy)", program.name)
+        return vals_list
```
livaios: Better to check the if first then loop on the values
```suggestion
        if 'name' not in default:
            for program, vals in zip(self, vals_list):
                vals['name'] = _("%s (copy)", program.name)
        return vals_list
```

livaios (changes requested):
Good work on this one!
Just one small suggestion 🙂

livaios (approved):
LGTM 👍
Thanks!

ltinel (approved):
LGTM, thanks! 🙂 
@robodoo r+

PR #247381, opened 2026-02-05
[IMP] sale, website_sale: improve configuator performance

csan-odoo:
Refactor to improve performance when opening the product configurator dialog.
- preloading configurator data to avoid an extra RPC roundtrip when opening the configurator

Enterprise PR: https://github.com/odoo/enterprise/pull/121341

task-3891049

### DIFF
--- a/addons/sale/controllers/product_configurator.py
+++ b/addons/sale/controllers/product_configurator.py
@@ -44,30 +44,32 @@ class SaleProductConfiguratorController(Controller):
             request.update_context(allowed_company_ids=[company_id])
         product_template = self._get_product_template(product_template_id)
 
-        combination = self.env["product.template.attribute.value"]
-        if ptav_ids:
-            combination = (
-                self
-                .env["product.template.attribute.value"]
-                .browse(ptav_ids)
-                .filtered(lambda ptav: ptav.product_tmpl_id.id == product_template_id)
-            )
-            # Set missing attributes (unsaved no_variant attributes, or new attribute on existing
-            # product)
-            unconfigured_ptals = (
-                product_template.attribute_line_ids - combination.attribute_line_id
-            ).filtered(lambda ptal: ptal.attribute_id.display_type != "multi")
-            combination += unconfigured_ptals.mapped(
-                lambda ptal: ptal.product_template_value_ids._only_active()[:1]
-            )
-        if not combination:
-            combination = product_template._get_first_possible_combination()
-        currency = self.env["res.currency"].browse(currency_id)
-        pricelist = self.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = product_template.get_single_product_variant()
 
-        return {
-            "products": [
+        if self.should_show_product_configurator(res, product_template, **kwargs):
+            combination = self.env["product.template.attribute.value"]
+            if ptav_ids:
+                combination = (
+                    self
+                    .env["product.template.attribute.value"]
+                    .browse(ptav_ids)
+                    .filtered(lambda ptav: ptav.product_tmpl_id.id == product_template_id)
+                )
+                # Set missing attributes (unsaved no_variant attributes, or new attribute on
+                # existing product)
+                unconfigured_ptals = (
+                    product_template.attribute_line_ids - combination.attribute_line_id
+                ).filtered(lambda ptal: ptal.attribute_id.display_type != "multi")
+                combination += unconfigured_ptals.mapped(
+                    lambda ptal: ptal.product_template_value_ids._only_active()[:1]
+                )
+            if not combination:
+                combination = product_template._get_first_possible_combination()
+            currency = self.env["res.currency"].browse(currency_id)
+            pricelist = self.env["product.pricelist"].browse(pricelist_id)
+            so_date = datetime.fromisoformat(so_date)
+
+            res["products"] = [
                 dict(
                     **self._get_product_information(
                         product_template,
@@ -80,26 +82,38 @@ class SaleProductConfiguratorController(Controller):
                         **kwargs,
                     )
                 )
-            ],
-            "optional_products": [
-                dict(
-                    **self._get_product_information(
-                        optional_product_template,
-                        optional_product_template._get_first_possible_combination(),
-                        currency,
-                        pricelist,
-                        so_date,
-                        **kwargs,
-                    ),
-                    parent_product_tmpl_id=product_template.id,
-                )
-                for optional_product_template in product_template.optional_product_ids
-                if self._should_show_product(optional_product_template)
             ]
-            if not only_main_product
-            else [],
-            "currency_id": currency_id,
-        }
+
+            res["optional_products"] = (
+                [
+                    dict(
+                        **self._get_product_information(
+                            optional_product_template,
+                            optional_product_template._get_first_possible_combination(),
+                            currency,
+                            pricelist,
+                            so_date,
+                            **kwargs,
+                        ),
+                        parent_product_tmpl_id=product_template.id,
+                    )
+                    for optional_product_template in product_template.optional_product_ids
+                    if self._should_show_product(optional_product_template)
+                ]
+                if not only_main_product
+                else []
+            )
+
+            res["currency_id"] = currency_id
+
+        return res
+
+    def should_show_product_configurator(
+        self, single_product_variant_dict, product_template, **kwargs
+    ):
+        return not single_product_variant_dict.get("product_id") or single_product_variant_dict.get(
+            "has_optional_products"
+        )
 
     @route(
         route="/sale/product_configurator/create_product",
--- a/addons/sale/controllers/product_configurator.py
+++ b/addons/sale/controllers/product_configurator.py
@@ -44,30 +44,35 @@ class SaleProductConfiguratorController(Controller):
             request.update_context(allowed_company_ids=[company_id])
         product_template = self._get_product_template(product_template_id)
 
-        combination = request.env["product.template.attribute.value"]
-        if ptav_ids:
-            combination = (
-                request
-                .env["product.template.attribute.value"]
-                .browse(ptav_ids)
-                .filtered(lambda ptav: ptav.product_tmpl_id.id == product_template_id)
-            )
-            # Set missing attributes (unsaved no_variant attributes, or new attribute on existing
-            # product)
-            unconfigured_ptals = (
-                product_template.attribute_line_ids - combination.attribute_line_id
-            ).filtered(lambda ptal: ptal.attribute_id.display_type != "multi")
-            combination += unconfigured_ptals.mapped(
-                lambda ptal: ptal.product_template_value_ids._only_active()[:1]
-            )
-        if not combination:
-            combination = product_template._get_first_possible_combination()
-        currency = request.env["res.currency"].browse(currency_id)
-        pricelist = request.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = (
+            request.env["product.template"].browse(product_template_id).get_single_product_variant()
+        )
 
-        return {
-            "products": [
+        is_template_configurable = not res.get("product_id") or res.get("has_optional_products")
+        if is_template_configurable:
+            combination = request.env["product.template.attribute.value"]
+            if ptav_ids:
+                combination = (
+                    request
+                    .env["product.template.attribute.value"]
+                    .browse(ptav_ids)
+                    .filtered(lambda ptav: ptav.product_tmpl_id.id == product_template_id)
+                )
+                # Set missing attributes (unsaved no_variant attributes, or new attribute on
+                # existing product)
+                unconfigured_ptals = (
+                    product_template.attribute_line_ids - combination.attribute_line_id
+                ).filtered(lambda ptal: ptal.attribute_id.display_type != "multi")
+                combination += unconfigured_ptals.mapped(
+                    lambda ptal: ptal.product_template_value_ids._only_active()[:1]
+                )
+            if not combination:
+                combination = product_template._get_first_possible_combination()
+            currency = request.env["res.currency"].browse(currency_id)
+            pricelist = request.env["product.pricelist"].browse(pricelist_id)
+            so_date = datetime.fromisoformat(so_date)
+
+            res["products"] = [
                 dict(
                     **self._get_product_information(
                         product_template,
@@ -80,26 +85,31 @@ class SaleProductConfiguratorController(Controller):
                         **kwargs,
                     )
                 )
-            ],
-            "optional_products": [
-                dict(
-                    **self._get_product_information(
-                        optional_product_template,
-                        optional_product_template._get_first_possible_combination(),
-                        currency,
-                        pricelist,
-                        so_date,
-                        **kwargs,
-                    ),
-                    parent_product_tmpl_id=product_template.id,
-                )
-                for optional_product_template in product_template.optional_product_ids
-                if self._should_show_product(optional_product_template)
             ]
-            if not only_main_product
-            else [],
-            "currency_id": currency_id,
-        }
+
+            res["optional_products"] = (
+                [
+                    dict(
+                        **self._get_product_information(
+                            optional_product_template,
+                            optional_product_template._get_first_possible_combination(),
+                            currency,
+                            pricelist,
+                            so_date,
+                            **kwargs,
+                        ),
+                        parent_product_tmpl_id=product_template.id,
+                    )
+                    for optional_product_template in product_template.optional_product_ids
+                    if self._should_show_product(optional_product_template)
+                ]
+                if not only_main_product
+                else []
+            )
+
+            res["currency_id"] = currency_id
+
+        return res
 
     @route(
         route="/sale/product_configurator/create_product",
--- a/addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
+++ b/addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
@@ -55,7 +55,8 @@ export class ComboConfiguratorDialog extends Component {
         });
         this._initSelectedComboItems();
         this.getPriceUrl = '/sale/combo_configurator/get_price';
-        useSubEnv({ currency: { id: this.props.currency_id } });
+        this.getValuesUrl = '/sale/product_configurator/get_values';
+        useSubEnv({ currencyId: this.props.currency_id });
 
         this.unconfigurableCombos = this.props.combos.filter(combo => !combo.isConfigurable);
         this.configurableCombos = this.props.combos.filter(combo => combo.isConfigurable);
@@ -76,11 +77,25 @@ export class ComboConfiguratorDialog extends Component {
         comboItem = this.getSelectedOrProvidedComboItem(comboId, comboItem);
         let product = comboItem.product;
         if (comboItem.is_configurable) {
+            const productConfiguratorData = await rpc(this.getValuesUrl,
+                {
+                    product_template_id: product.product_tmpl_id,
+                    quantity: 1,
+                    currency_id: this.props.currency_id,
+                    so_date: this.props.date,
+                    company_id: this.props.company_id,
+                    pricelist_id: this.props.pricelist_id,
+                    ptav_ids:  product.selectedPtavIds,
+                    only_main_product: true,
+                    show_packaging: false,
+                    ...this._getAdditionalRpcParams(),
+                });
+            const { products } = productConfiguratorData;
             this.dialog.add(ProductConfiguratorDialog, {
                 productTemplateId: product.product_tmpl_id,
-                ptavIds: product.selectedPtavIds,
+                products: products,
+                optionalProducts: [],
                 customPtavs: product.selectedCustomPtavs,
-                quantity: 1,
                 companyId: this.props.company_id,
                 pricelistId: this.props.pricelist_id,
                 currencyId: this.props.currency_id,
@@ -90,7 +105,6 @@ export class ComboConfiguratorDialog extends Component {
                     canChangeVariant: false,
                     showQuantity: false,
                     showPrice: false,
-                    showPackaging: false,
                 },
                 size: "md",
                 save: async configuredProduct => {
--- a/addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
+++ b/addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
@@ -1,5 +1,5 @@
 import { useSubEnv } from "@web/owl2/utils";
-import { Component, onMounted, onWillStart, onWillUnmount, props, proxy, t } from "@odoo/owl";
+import { Component, onMounted, onWillUnmount, props, proxy, t } from "@odoo/owl";
 import { Dialog } from '@web/core/dialog/dialog';
 import { _t } from "@web/core/l10n/translation";
 import { rpc } from "@web/core/network/rpc";
@@ -11,20 +11,18 @@ export const productConfiguratorDialogOptionsShape = {
     canChangeVariant: t.boolean().optional(),
     showQuantity: t.boolean().optional(),
     showPrice: t.boolean().optional(),
-    showPackaging: t.boolean().optional(),
 };
 
 export const productConfiguratorDialogProps = {
     productTemplateId: t.number(),
-    ptavIds: t.array(t.number()),
+    products: t.array(),
+    optionalProducts: t.array(),
     customPtavs: t.array(
         t.object({
             id: t.number(),
             value: t.string(),
         })
     ),
-    quantity: t.number(),
-    productUOMId: t.number().optional(),
     companyId: t.number().optional(),
     pricelistId: t.number().optional(),
     currencyId: t.number().optional(),
@@ -57,10 +55,6 @@ export class ProductConfiguratorDialog extends Component {
             products: [],
             optionalProducts: [],
         });
-        // Nest the currency id in an object so that it stays up to date in the `env`, even if we
-        // modify it in `onWillStart` afterwards.
-        this.currency = { id: this.props.currencyId };
-        this.getValuesUrl = '/sale/product_configurator/get_values';
         this.createProductUrl = '/sale/product_configurator/create_product';
         this.updateCombinationUrl = '/sale/product_configurator/update_combination';
         this.getOptionalProductsUrl = '/sale/product_configurator/get_optional_products';
@@ -69,10 +63,9 @@ export class ProductConfiguratorDialog extends Component {
 
         useSubEnv({
             mainProductTmplId: this.props.productTemplateId,
-            currency: this.currency,
+            currencyId: this.props.currency_id,
             canChangeVariant: this.props.options?.canChangeVariant ?? true,
             showQuantity: this.props.options?.showQuantity ?? true,
-            showPackaging: this.props.options?.showPackaging ?? true,
             showPrice: this.props.options?.showPrice ?? true,
             addProduct: this._addProduct.bind(this),
             removeProduct: this._removeProduct.bind(this),
@@ -83,32 +76,23 @@ export class ProductConfiguratorDialog extends Component {
             isPossibleCombination: this._isPossibleCombination,
         });
 
-        onWillStart(async () => {
-            const {
-                products,
-                optional_products,
-                currency_id,
-            } = await this._loadData(this.props.edit);
+        // If the product configurator is opened after the combo configurator (which happens if
+        // a combo product has optional products), props.products will contain a single product
+        // (i.e. the combo product), which should be linked to the previously selected combo
+        // items.
 
-            // If the product configurator is opened after the combo configurator (which happens if
-            // a combo product has optional products), `_loadData` will return a single product
-            // (i.e. the combo product), which should be linked to the previously selected combo
-            // items.
-            products[0].selectedComboItems = this.props.selectedComboItems || [];
+        this.props.products[0].selectedComboItems = this.props.selectedComboItems || [];
 
-            this.state.products = products;
-            this.state.optionalProducts = optional_products;
-            for (const customPtav of this.props.customPtavs) {
-                this._updatePTAVCustomValue(
-                    this.env.mainProductTmplId,
-                    customPtav.id,
-                    customPtav.value
-                );
-            }
-            this._checkExclusions(this.state.products[0]);
-            // Use the currency id retrieved from the server if none was provided in the props.
-            this.currency.id ??= currency_id;
-        });
+        this.state.products = this.props.products;
+        this.state.optionalProducts = this.props.optionalProducts;
+        for (const customPtav of this.props.customPtavs) {
+            this._updatePTAVCustomValue(
+                this.env.mainProductTmplId,
+                customPtav.id,
+                customPtav.value
+            );
+        }
+        this._checkExclusions(this.state.products[0]);
 
         onMounted(() => this.env.bus.trigger("FORM-CONTROLLER:FORM-IN-DIALOG:ADD"));
         onWillUnmount(() => this.env.bus.trigger("FORM-CONTROLLER:FORM-IN-DIALOG:REMOVE"));
@@ -128,28 +112,12 @@ export class ProductConfiguratorDialog extends Component {
             (sum, product) => sum + product.price * product.quantity,
             0
         );
-        return formatCurrency(total, this.currency.id);
+        return formatCurrency(total, this.props.currency_id);
     }
 
     //--------------------------------------------------------------------------
     // Data Exchanges
     //--------------------------------------------------------------------------
-
-    async _loadData(onlyMainProduct) {
-        return rpc(this.getValuesUrl, {
-            product_template_id: this.props.productTemplateId,
-            quantity: this.props.quantity,
-            currency_id: this.currency.id,
-            so_date: this.props.soDate,
-            product_uom_id: this.props.productUOMId,
-            company_id: this.props.companyId,
-            pricelist_id: this.props.pricelistId,
-            ptav_ids: this.props.ptavIds,
-            only_main_product: onlyMainProduct,
-            show_packaging: this.env.showPackaging,
-            ...this._getAdditionalRpcParams(),
-        });
-    }
 
     async _createProduct(product) {
         return rpc(this.createProductUrl, {
@@ -162,7 +130,7 @@ export class ProductConfiguratorDialog extends Component {
         return rpc(this.updateCombinationUrl, {
             product_template_id: product.product_tmpl_id,
             ptav_ids: this._getCombination(product),
-            currency_id: this.currency.id,
+            currency_id: this.props.currencyId,
             so_date: this.props.soDate,
             quantity: quantity,
             product_uom_id: uomId,
@@ -175,7 +143,7 @@ export class ProductConfiguratorDialog extends Component {
     async _getOptionalProducts(product) {
         return rpc(this.getOptionalProductsUrl, {
             product_template_id: product.product_tmpl_id,
-            currency_id: this.currency.id,
+            currency_id: this.props.currencyId,
             so_date: this.props.soDate,
             company_id: this.props.companyId,
             pricelist_id: this.props.pricelistId,
--- a/addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
+++ b/addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
@@ -12,6 +12,10 @@ export class ProductConfiguratorDialog extends Component {
     static props = {
         productTemplateId: Number,
         ptavIds: { type: Array, element: Number },
+        preloadedData: {
+            type: Object,
+            optional: true,
+        },
         customPtavs: {
             type: Array,
             element: Object,
@@ -94,7 +98,7 @@ export class ProductConfiguratorDialog extends Component {
                 products,
                 optional_products,
                 currency_id,
-            } = await this._loadData(this.props.edit);
+            } = this.props.preloadedData ?? await this._loadData(this.props.edit);
 
             // If the product configurator is opened after the combo configurator (which happens if
             // a combo product has optional products), `_loadData` will return a single product
--- a/addons/sale/static/src/js/sale_product_field.js
+++ b/addons/sale/static/src/js/sale_product_field.js
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
-    async _onProductTemplateUpdate() {
-        const result = await this.orm.call(
-            'product.template',
-            'get_single_product_variant',
-            [this.props.record.data.product_template_id.id],
+    async _getPreloadedConfigData() {
+        const saleOrderRecord = this.props.record.model.root;
+        const saleOrderLine = this.props.record.data;
+        const ptavIds = this._getVariantPtavIds(saleOrderLine);
+        return rpc('/sale/product_configurator/get_values',
             {
-                context: this.props.context,
-            }
-        );
+                product_template_id: saleOrderLine.product_template_id.id,
+                quantity: saleOrderLine.product_uom_qty,
+                currency_id: saleOrderLine.currency_id.id,
+                so_date: serializeDateTime(saleOrderRecord.data.date_order),
+                product_uom_id: saleOrderLine.product_uom_id.id,
+                company_id: saleOrderRecord.data.company_id.id,
+                pricelist_id: saleOrderRecord.data.pricelist_id.id,
+                ptav_ids: ptavIds,
+                ...this._getAdditionalRpcParams(),
+            });
+    }
+
+    async _onProductTemplateUpdate() {
+        const result = await this._getPreloadedConfigData();
         if (result && result.product_id) {
             if (this.props.record.data.product_id != result.product_id.id) {
                 if (result.is_combo) {
                     await this.props.record.update({
                         product_id: { id: result.product_id, display_name: result.product_name },
                     });
-                    this._openComboConfigurator(false, result.has_optional_products);
+                    this._openComboConfigurator(false, result.has_optional_products, result);
                 } else if (result.has_optional_products) {
-                    this._openProductConfigurator();
+                    this._openProductConfigurator({ preloadedData: result });
                 } else {
                     await this.props.record.update({
                         product_id: { id: result.product_id, display_name: result.product_name },
@@ -212,7 +223,7 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
                 }
             }
         } else if (!result.mode || result.mode === 'configurator') {
-            this._openProductConfigurator();
+            this._openProductConfigurator({ preloadedData: result });
         } else {
             // only triggered when sale_product_matrix is installed.
             this._openGridConfigurator();
@@ -227,11 +238,15 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         if (this.isCombo) {
             this._openComboConfigurator(true);
         } else if (this.isConfigurableTemplate) {
-            this._openProductConfigurator(true);
+            this._openProductConfigurator({edit: true});
         }
     }
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator({
+        edit = false,
+        selectedComboItems = [],
+        preloadedData,
+    } = {}) {
         const saleOrderRecord = this.props.record.model.root;
         const saleOrderLine = this.props.record.data;
         const ptavIds = this._getVariantPtavIds(saleOrderLine);
@@ -249,6 +264,7 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         this.dialog.add(ProductConfiguratorDialog, {
             productTemplateId: saleOrderLine.product_template_id.id,
             ptavIds: ptavIds,
+            preloadedData: preloadedData,
             customPtavs: customPtavs,
             quantity: saleOrderLine.product_uom_qty,
             productUOMId: saleOrderLine.product_uom_id.id,
@@ -292,7 +308,7 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         });
     }
 
-    async _openComboConfigurator(edit = false, hasOptionalProducts = false) {
+    async _openComboConfigurator(edit = false, hasOptionalProducts = false, preloadedData) {
         const saleOrder = this.props.record.model.root.data;
         const comboLineRecord = this.props.record;
         const comboItemLineRecords = getLinkedSaleOrderLines(comboLineRecord).filter(record => !!record.data.combo_item_id);
@@ -321,7 +337,8 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
                 { 'quantity' : remainingData.quantity },
                 preselectedComboItems,
                 edit,
-                hasOptionalProducts
+                hasOptionalProducts,
+                preloadedData
             );
         }
         this.dialog.add(ComboConfiguratorDialog, {
@@ -344,7 +361,13 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         });
     }
 
-    async handleComboSave(comboProductData, selectedComboItems, edit, hasOptionalProducts) {
+    async handleComboSave(
+        comboProductData,
+        selectedComboItems,
+        edit,
+        hasOptionalProducts,
+        preloadedData,
+    ) {
         const saleOrder = this.props.record.model.root.data;
         const comboLineRecord = this.props.record;
         saleOrder.order_line.leaveEditMode();
@@ -365,7 +388,10 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
             const selectedComboProducts = selectedComboItems.map(
                 item => ({ name: item.product.display_name })
             );
-            await this._openProductConfigurator(false, selectedComboProducts);
+            await this._openProductConfigurator({
+                selectedComboItems: selectedComboProducts,
+                preloadedData: preloadedData,
+            });
         }
     }
 
--- a/addons/sale/static/src/js/sale_product_mixin.js
+++ b/addons/sale/static/src/js/sale_product_mixin.js
@@ -75,34 +75,50 @@ export const saleProductMixin = () => ({
         return _t("Edit Configuration");
     },
 
+    async _getProductConfiguratorData(edit = false) {
+        const saleOrder = this.props.record.model.root.data;
+        const saleOrderLine = this.props.record.data;
+        const ptavIds = [...this._getVariantPtavIds(saleOrderLine)];
+        if (edit) {
+            // no_variant attributes don't need to be given to the configurator for new products.
+            ptavIds.push(...this._getNoVariantPtavIds(saleOrderLine));
+        }
+        return rpc('/sale/product_configurator/get_values',
+            {
+                product_template_id: saleOrderLine.product_template_id.id,
+                quantity: saleOrderLine.product_uom_qty,
+                currency_id: saleOrderLine.currency_id.id,
+                so_date: serializeDateTime(saleOrder.date_order),
+                product_uom_id: saleOrderLine.product_uom_id.id,
+                company_id: saleOrder.company_id.id,
+                pricelist_id: saleOrder.pricelist_id.id,
+                ptav_ids: ptavIds,
+                only_main_product: edit,
+                ...this._getAdditionalRpcParams(),
+            });
+    },
+
     async _onProductTemplateUpdate() {
         super._onProductTemplateUpdate();
-        const result = await this.orm.call(
-            "product.template",
-            "get_single_product_variant",
-            [this.props.record.data.product_template_id.id],
-            {
-                context: this.props.context,
-            }
-        );
-        if (result && result.product_id) {
-            if (this.props.record.data.product_id != result.product_id.id) {
-                if (result.is_combo) {
+        const data = await this._getProductConfiguratorData();
+        if (data && data.product_id) {
+            if (this.props.record.data.product_id != data.product_id.id) {
+                if (data.is_combo) {
                     await this.props.record.update({
-                        product_id: { id: result.product_id, display_name: result.product_name },
+                        product_id: { id: data.product_id, display_name: data.product_name },
                     });
-                    this._openComboConfigurator(false, result.has_optional_products);
-                } else if (result.has_optional_products) {
-                    this._openProductConfigurator();
+                    this._openComboConfigurator({edit: false, data: data});
+                } else if (data.has_optional_products) {
+                    this._openProductConfigurator({ data: data });
                 } else {
                     await this.props.record.update({
-                        product_id: { id: result.product_id, display_name: result.product_name },
+                        product_id: { id: data.product_id, display_name: data.product_name },
                     });
                     this._onProductUpdate();
                 }
             }
-        } else if (!result.mode || result.mode === "configurator") {
-            this._openProductConfigurator();
+        } else if (!data.mode || data.mode === 'configurator') {
+            this._openProductConfigurator({ data: data });
         } else {
             // only triggered when sale_product_matrix is installed.
             this._openGridConfigurator();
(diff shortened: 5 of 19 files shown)


lipiraux:
Hello @csan-odoo , could you rebase and apply ruff before I take a look? thx!

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,15 +186,29 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
+    async _getPreloadedConfigData() {
+        const saleOrderRecord = this.props.record.model.root;
+        const saleOrderLine = this.props.record.data;
+        const ptavIds = this._getVariantPtavIds(saleOrderLine);
+        const result = rpc('/sale/product_configurator/get_values',
+            {   
+                product_template_id: this.props.record.data.product_template_id.id,
```
lipiraux:
```suggestion
                product_template_id: saleOrderLine.product_template_id.id,
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,15 +186,29 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
+    async _getPreloadedConfigData() {
+        const saleOrderRecord = this.props.record.model.root;
+        const saleOrderLine = this.props.record.data;
```
lipiraux: What is the difference between `this.props.record.data` and `this.props.record.model.root.data`?

Im not really familiar with the JS record objects 😅

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,15 +186,29 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
+                product_template_id: this.props.record.data.product_template_id.id,
+                quantity: saleOrderLine.product_uom_qty || 1,
+                currency_id: saleOrderLine.currency_id?.id,
+                so_date: saleOrderRecord.data.date_order ?
+                    serializeDateTime(saleOrderRecord.data.date_order)
+                    : undefined,
+                product_uom_id: saleOrderLine.product_uom_id?.id,
+                company_id: saleOrderRecord.data.company_id?.id,
+                pricelist_id: saleOrderRecord.data.pricelist_id?.id,
+                ptav_ids: ptavIds,
```
lipiraux: I see that you add a lot of fallback in case their is no data. But I also see that it wasn't [before](https://github.com/odoo-dev/odoo/blob/master-improve_configurator_dialog_efficiency-csan/addons/sale/static/src/js/sale_product_field.js#L273-L282).
What is different? Is it possible that their is no SOL data when you preload the configurator values?

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -365,7 +388,10 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
             const selectedComboProducts = selectedComboItems.map(
                 item => ({ name: item.product.display_name })
             );
-            await this._openProductConfigurator(false, selectedComboProducts);
+            await this._openProductConfigurator({
+                edit: false,
```
lipiraux: since its the default
```suggestion
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -203,7 +217,9 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
                     });
                     this._openComboConfigurator(false, result.has_optional_products);
                 } else if (result.has_optional_products) {
-                    this._openProductConfigurator();
+                    this._openProductConfigurator({
+                        preloadedData: result
+                    });
```
lipiraux: nit: that can be applied throughout this file
```suggestion
                    this._openProductConfigurator({ preloadedData: result });
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -203,7 +217,9 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
                     });
                     this._openComboConfigurator(false, result.has_optional_products);
```
lipiraux: Another idea of improvement would be to pass the preloaded data to _openComboConfigurator such that
it can pass it down to _handleComboSave which will open the product configurator if their is
optional products to choose. This way you remove another duplicate rpc.

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -227,11 +245,15 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         if (this.isCombo) {
             this._openComboConfigurator(true);
         } else if (this.isConfigurableTemplate) {
-            this._openProductConfigurator(true);
+            this._openProductConfigurator({edit: true});
         }
     }
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator({
```
lipiraux: You forgot to update the definition in the other modules.
For example
https://github.com/odoo-dev/odoo/blob/master-improve_configurator_dialog_efficiency-csan/addons/sale_product_matrix/static/src/js/sale_product_field.js#L15

addons/sale/static/tests/sale_product_field.test.js
```diff
@@ -67,14 +67,23 @@ test("pressing tab with incomplete text will create a product", async () => {
     await contains(".o_field_x2many_list .o_field_x2many_list_row_add button").click();
     await contains("[name='product_template_id'] input").edit("new product");
     await press("tab");
+    onRpc("/sale/product_configurator/get_values", () => {
+        expect.step("get_values");
+        return {
+            product_id: 42,
+            product_name: "Test Product",
+            has_optional_producyts: false,
```
lipiraux: if it works without 😅
```suggestion
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,15 +186,29 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
+                so_date: saleOrderRecord.data.date_order ?
+                    serializeDateTime(saleOrderRecord.data.date_order)
+                    : undefined,
+                product_uom_id: saleOrderLine.product_uom_id?.id,
+                company_id: saleOrderRecord.data.company_id?.id,
+                pricelist_id: saleOrderRecord.data.pricelist_id?.id,
+                ptav_ids: ptavIds,
+                ...this._getAdditionalRpcParams(),
+            });
+        return result;
```
lipiraux:
```suggestion
        return rpc('/sale/product_configurator/get_values', {   
            product_template_id: this.props.record.data.product_template_id.id,
            quantity: saleOrderLine.product_uom_qty || 1,
            currency_id: saleOrderLine.currency_id?.id,
            so_date: saleOrderRecord.data.date_order ?
                serializeDateTime(saleOrderRecord.data.date_order)
                : undefined,
            product_uom_id: saleOrderLine.product_uom_id?.id,
            company_id: saleOrderRecord.data.company_id?.id,
            pricelist_id: saleOrderRecord.data.pricelist_id?.id,
            ptav_ids: ptavIds,
            ...this._getAdditionalRpcParams(),
        });
```

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,32 @@ def sale_product_configurator_get_values(
...
+        if not res.get("product_id") or res.get("has_optional_products"):
+            combination = request.env["product.template.attribute.value"]
+            if ptav_ids:
+                combination = (
+                    request
+                    .env["product.template.attribute.value"]
+                    .browse(ptav_ids)
+                    .filtered(lambda ptav: ptav.product_tmpl_id.id == product_template_id)
+                )
+                unconfigured_ptals = (
```
lipiraux, line 60: Shouldn't we keep this comment?
```suggestion
                )
                # Set missing attributes (unsaved no_variant attributes, or new attribute on
                # existing product)
                unconfigured_ptals = (
```

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,32 @@ def sale_product_configurator_get_values(
...
-        currency = request.env["res.currency"].browse(currency_id)
-        pricelist = request.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = (
+            request.env["product.template"].browse(product_template_id).get_single_product_variant()
+        )
 
-        return {
-            "products": [
+        if not res.get("product_id") or res.get("has_optional_products"):
```
lipiraux: nit: took me some time to understand what it meant 😅
```suggestion
        is_template_configurable = not res.get("product_id") or res.get("has_optional_products")
        if is_template_configurable:
```

lipiraux (changes requested):

addons/sale/static/src/js/sale_product_field.js
csan-odoo: this.props.record is the proxy record representing one sale order line. 
- .data gives access to the actiual field values of the sale order line. 
- .model.root gives access to the parent record(Sale order record) which is the same thing as ._parentRecord. Again .data allows access to the actual fields.

addons/sale/controllers/product_configurator.py
csan-odoo: you're right, left it out by mistake 👍

addons/sale/static/src/js/sale_product_field.js
csan-odoo: I included these fallbacks in earlier iterations of the code, but they’re indeed no longer necessary 👍

addons/sale/static/src/js/sale_product_field.js
lipiraux: oooh, ok ok. nice thanks. totally makes sense given the name of the variables 🤦‍♂️

addons/sale/static/tests/sale_product_field.test.js
```diff
@@ -63,6 +64,16 @@ test("pressing tab with incomplete text will create a product", async () => {
...
 
+    SaleOrderLineProductField.prototype._getPreloadedConfigData = async function() {
+        expect.step("_getPreloadedConfigData");
+        return {
+            product_id: { id: 42, display_name: "Test Product" },
+            product_name: "Test Product",
+            has_optional_products: false,
+            is_combo: false,
+        };
+    };
```
lipiraux: Im not familiar enough with how hoot works and should be written, but isn't this going to remain
patched even after the test? 😅

I saw in others tests this function `patchWithCleanup` which seems to be what you want.

addons/sale_product_matrix/static/src/js/sale_product_field.js
```diff
@@ -12,7 +12,7 @@ patch(SaleOrderLineProductField.prototype, {
         return this.matrixConfigurator.open(this.props.record, edit);
     },
 
-    async _openProductConfigurator(edit=false, selectedComboItems=[]) {
+    async _openProductConfigurator(edit=false, selectedComboItems=[], preloadedData = {},) {
```
lipiraux: mmmh 🤔, I think I realise why it was working fine without updating the definition.

JS function don't seem to care about receiving more arguments than expected. And further down the
`...arguments` will unpack whatever was given, unexpected args included.

Not sure if it was necessary in the end. wdyt? To me we should enumerate all the allowed/expected
arguments even if not used. But that might be a python way of thinking 🤷‍♂️

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -344,7 +361,13 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         });
     }
 
-    async handleComboSave(comboProductData, selectedComboItems, edit, hasOptionalProducts) {
+    async handleComboSave(
+        comboProductData,
+        selectedComboItems,
+        edit,
+        hasOptionalProducts,
+        preloadedData
```
lipiraux:
```suggestion
        preloadedData,
```

addons/sale_product_matrix/static/src/js/sale_product_field.js
lipiraux:
```suggestion
    async _openProductConfigurator(edit=false, selectedComboItems=[], preloadedData = {}) {
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -90,11 +94,18 @@ export class ProductConfiguratorDialog extends Component {
...
+            }
+            else {
+                data = await this._loadData(this.props.edit);
+            }
             const {
                 products,
                 optional_products,
                 currency_id,
-            } = await this._loadData(this.props.edit);
+            } = data;
```
lipiraux: Its a little thing, but if `preloadedData` is the same as what would be returned by `_loadData`, but
there wasn't any `products` in `preloadedData`, then why would there be in the result of
`_loadData`? 

To me you should just check if `preloadedData` was given and let the code crash further down if
`products` wasn't in it. If it wasn't, then there is already something fishy going on, no need to go
talk to the server before realizing 😅

Also, just a suggestion to reduce the diff 
```suggestion
            const {
                products,
                optional_products,
                currency_id,
            } = this.props.preloadedData ?? await this._loadData(this.props.edit);
```

lipiraux (changes requested):
Thx for the changes. Last couple suggestions 🚀

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
csan-odoo: This.props.preloadedData was sometimes an empty object ({}), which caused it to pass the conditional check unexpectedly. Using this.props.preloadedData?.products ensured the check behaved correctly in that case. I made some adjustments and it seems to be working fine now. Thanks for the suggestion!

addons/sale_product_matrix/static/src/js/sale_product_field.js
csan-odoo: I agree. It improves readability, especially since we’re forwarding the remaining arguments to the parent function.

addons/sale/static/tests/sale_product_field.test.js
csan-odoo: that's right!

lipiraux (approved):

csan-odoo:
Hello @ltinel,

Thanks for the feedback :)

I removed the `/website_sale/should_show_product_configurator` RPC, so eCommerce now also relies exclusively on one call to open the configurator.

I confirm that my approach does not slow down either combo products or regular products. In fact, load times are marginally improved. My assumption is that the direct RPC call:
```python
return rpc('/sale/product_configurator/get_values', {
    product_template_id: saleOrderLine.product_template_id.id,
    quantity: saleOrderLine.product_uom_qty,
    currency_id: saleOrderLine.currency_id.id,
    so_date: serializeDateTime(saleOrderRecord.data.date_order),
    product_uom_id: saleOrderLine.product_uom_id.id,
    company_id: saleOrderRecord.data.company_id.id,
    pricelist_id: saleOrderRecord.data.pricelist_id.id,
    ptav_ids: ptavIds,
    ...this._getAdditionalRpcParams(),
});
```
is more efficient than going through the ORM with:
```python
const result = await this.orm.call(
    'product.template',
    'get_single_product_variant',
    [this.props.record.data.product_template_id.id],
    {
        context: this.props.context,
    }
);
```

I also managed to compute `has_optional_product` correctly without requiring an override.

Let me know what you think 🙂

addons/website_sale/controllers/product_configurator.py
```diff
@@ -9,12 +9,27 @@
...
+        should_show = self.website_sale_should_show_product_configurator(
+            product_template_id, is_product_configured
+        )
+
+        if not should_show:
+            return False
+
+        return super().sale_product_configurator_get_values(*args, **kwargs)
+
     def website_sale_should_show_product_configurator(
```
ltinel: Ideally, I'd like to get rid of this method :) One way to do this is to add an override of `get_single_product_variant` in the `website_sale` module, and move this logic in there. More precisely, in the override, add a `has_multiple_uoms` entry to the result, and update the `has_optional_products` entry to only consider "valid" optional products (i.e. products that can be added to the cart and belong to the current website).

This way, all the necessary info can be fetched via `get_single_product_variant` (except for `is_product_configured`, but this info comes directly from the frontend, so we can handle it there).

The idea is that the computation of whether we should show the configurator would happen in `addons/website_sale/static/src/js/cart_service.js`, kind of like we do in `addons/sale/static/src/js/sale_product_field.js`.

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -94,7 +98,7 @@ export class ProductConfiguratorDialog extends Component {
                 products,
                 optional_products,
                 currency_id,
-            } = await this._loadData(this.props.edit);
+            } = this.props.preloadedData ?? await this._loadData(this.props.edit);
```
ltinel: IMO, we should _always_ compute the data before opening the product configurator (like we do for the combo configurator), which will allow us to drop `_loadData`. I believe the product configurator is opened in 3 places (which are the places to update):
- When adding a configurable product to an SO in the backend (`sale`),
- When adding a configurable product to the cart in the frontend (`website_sale`),
- When selection a configurable product in the combo configurator.

Additionally, I believe a few of the `ProductConfiguratorDialog` props were only used by `_loadData`, so they can be dropped as well.

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -12,6 +12,10 @@ export class ProductConfiguratorDialog extends Component {
     static props = {
         productTemplateId: Number,
         ptavIds: { type: Array, element: Number },
+        preloadedData: {
```
ltinel: I'm not a big fan of putting the data inside a `preloadedData` prop. Could you instead extract each piece of data into its own prop?

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
-    async _onProductTemplateUpdate() {
-        const result = await this.orm.call(
-            'product.template',
-            'get_single_product_variant',
-            [this.props.record.data.product_template_id.id],
+    async _getPreloadedConfigData() {
```
ltinel: For the time being, I'd still keep the `get_single_product_variant` ORM call separate from the `/sale/product_configurator/get_values` RPC.

Conceptually, `get_single_product_variant` computes the minimum amount of data needed to know whether to open the product configurator, the combo configurator, the grid configurator, or no configurator at all. Moving the `get_single_product_variant` call into the `/sale/product_configurator/get_values` RPC couples it to the product configurator, which is conceptually debatable, and makes the code slightly more confusing IMO.

What you did seems indeed a bit more time-efficient, but I'm not convinced that the perf gain outweighs the loss of clarity (especially since this code isn't really a perf bottleneck and doesn't really need to be more time-efficient).

ltinel: Discussed offline. What you did is fine, but maybe rename this into something like `_getProductConfiguratorData`

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
+                ptav_ids: ptavIds,
+                ...this._getAdditionalRpcParams(),
+            });
+    }
+
+    async _onProductTemplateUpdate() {
+        const result = await this._getPreloadedConfigData();
         if (result && result.product_id) {
             if (this.props.record.data.product_id != result.product_id.id) {
                 if (result.is_combo) {
```
ltinel: As discussed offline, let's remove `is_combo` from `get_single_product_variant` (and thus also from the `/sale/product_configurator/get_values` RPC), and instead retrieve it from the view (something like `this.props.record.data.product_template_id.type === 'combo'`)

addons/website_sale/controllers/product_configurator.py
ltinel: Discussed offline. What you did is fine, but I would specify `product_template_id` and `is_product_configured` explicitly in the the method params.

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,35 @@ def sale_product_configurator_get_values(
...
-            combination += unconfigured_ptals.mapped(
-                lambda ptal: ptal.product_template_value_ids._only_active()[:1]
-            )
-        if not combination:
-            combination = product_template._get_first_possible_combination()
-        currency = self.env["res.currency"].browse(currency_id)
-        pricelist = self.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = (
+            request.env["product.template"].browse(product_template_id).get_single_product_variant()
```
ltinel:
```suggestion
            product_template.get_single_product_variant()
```

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,35 @@ def sale_product_configurator_get_values(
...
+        )
 
-        return {
-            "products": [
+        is_template_configurable = not res.get("product_id") or res.get("has_optional_products")
+        if is_template_configurable:
+            combination = request.env["product.template.attribute.value"]
```
ltinel: Same elsewhere :)
```suggestion
            combination = self.env["product.template.attribute.value"]
```

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,35 @@ def sale_product_configurator_get_values(
...
-        currency = self.env["res.currency"].browse(currency_id)
-        pricelist = self.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = (
+            request.env["product.template"].browse(product_template_id).get_single_product_variant()
+        )
 
-        return {
-            "products": [
+        is_template_configurable = not res.get("product_id") or res.get("has_optional_products")
```
ltinel: This computation will vary depending on the module. Could you move it into a hook? And rename it to something like `should_show_product_configurator`.

That way, you'll be able to include `mode` in the computation (and exclude it in `website_sale`).

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
         comboItem = this.getSelectedOrProvidedComboItem(comboId, comboItem);
         let product = comboItem.product;
         if (comboItem.is_configurable) {
+            const preloadedData = await rpc(this.getValuesUrl,
```
ltinel:
```suggestion
            const productConfiguratorData = await rpc(this.getValuesUrl,
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+                    ...this._getAdditionalDialogProps(),
+                });
+            const {
+                products,
+                optional_products,
+                currency_id,
+            } = preloadedData;
             this.dialog.add(ProductConfiguratorDialog, {
                 productTemplateId: product.product_tmpl_id,
                 ptavIds: product.selectedPtavIds,
```
ltinel: I don't think this is needed anymore.
```suggestion
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+                products: products,
+                optional_products: optional_products,
+                currency_id: currency_id,
                 customPtavs: product.selectedCustomPtavs,
                 quantity: 1,
```
ltinel: Also not needed anymore.
```suggestion
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -13,6 +13,8 @@ export class ProductConfiguratorDialog extends Component {
     static props = {
         productTemplateId: Number,
         ptavIds: { type: Array, element: Number },
```
ltinel: Not needed anymore.
```suggestion
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
ltinel: Not needed anymore.
```suggestion
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -116,7 +112,7 @@ export class ProductConfiguratorDialog extends Component {
             }
             this._checkExclusions(this.state.products[0]);
             // Use the currency id retrieved from the server if none was provided in the props.
-            this.currency.id ??= currency_id;
+            this.currency.id ??= this.props.currency_id;
```
ltinel: No need for this anymore.
```suggestion
```

In fact, could you:
- Replace `this.currency.id` with `this.props.currency_id` in this file.
- Replace `currency: this.currency` with `currencyId: this.props.currency_id` in the subEnv in this file.
- Remove `this.currency` in this file.
- Replace `currency: { id: this.props.currency_id }` with `currencyId: this.props.currency_id` in the subEnv in `combo_configurator_dialog.js`.
- Replace `this.env.currency.id` with `this.env.currencyId` everywhere.

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -93,20 +95,14 @@ export class ProductConfiguratorDialog extends Component {
         });
 
         onWillStart(async () => {
```
ltinel: Since there's no async code in the `onWillStart` hook anymore, could you extract its content and drop the hook? :)

See https://odoo.github.io/owl/documentation/v3/owl/reference/component.html#willstart

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+                {
+                    product_template_id: product.product_tmpl_id,
+                    quantity: 1,
+                    currency_id: this.props.currency_id,
+                    so_date: this.props.date,
+                    product_uom_id: null,
+                    company_id: this.props.company_id,
+                    pricelist_id: this.props.pricelist_id,
+                    ptav_ids:  product.selectedPtavIds,
+                    ...this._getAdditionalDialogProps(),
```
ltinel:
```suggestion
                    ...this._getAdditionalRpcParams(),
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -144,22 +140,6 @@ export class ProductConfiguratorDialog extends Component {
     // Data Exchanges
     //--------------------------------------------------------------------------
 
-    async _loadData(onlyMainProduct) {
-        return rpc(this.getValuesUrl, {
```
ltinel, line 139: Could you also remove `this.getValuesUrl` and its overrides?

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
         comboItem = this.getSelectedOrProvidedComboItem(comboId, comboItem);
         let product = comboItem.product;
         if (comboItem.is_configurable) {
+            const preloadedData = await rpc(this.getValuesUrl,
+                {
+                    product_template_id: product.product_tmpl_id,
+                    quantity: 1,
+                    currency_id: this.props.currency_id,
+                    so_date: this.props.date,
+                    product_uom_id: null,
```
ltinel: No need :)
```suggestion
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+            const {
+                products,
+                optional_products,
+                currency_id,
+            } = preloadedData;
             this.dialog.add(ProductConfiguratorDialog, {
                 productTemplateId: product.product_tmpl_id,
                 ptavIds: product.selectedPtavIds,
+                products: products,
+                optional_products: optional_products,
```
ltinel: We shouldn't pass optional products.
```suggestion
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+            const preloadedData = await rpc(this.getValuesUrl,
+                {
+                    product_template_id: product.product_tmpl_id,
+                    quantity: 1,
+                    currency_id: this.props.currency_id,
+                    so_date: this.props.date,
+                    product_uom_id: null,
+                    company_id: this.props.company_id,
+                    pricelist_id: this.props.pricelist_id,
+                    ptav_ids:  product.selectedPtavIds,
```
ltinel, line 88:
```suggestion
                    ptav_ids:  product.selectedPtavIds,
                    only_main_product: true,
                    show_packaging: false,
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
ltinel: Not needed
```suggestion
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
ltinel: You can drop this now (and remove it from the subEnv).
```suggestion
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
-    async _onProductTemplateUpdate() {
-        const result = await this.orm.call(
-            'product.template',
-            'get_single_product_variant',
-            [this.props.record.data.product_template_id.id],
+    async _getProductConfiguratorData() {
+        const saleOrderRecord = this.props.record.model.root;
```
ltinel:
```suggestion
        const saleOrder = this.props.record.model.root.data;
```
That way you don't need to systematically do `.data` below.

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
-            }
-        );
+                product_template_id: saleOrderLine.product_template_id.id,
+                quantity: saleOrderLine.product_uom_qty,
+                currency_id: saleOrderLine.currency_id.id,
+                so_date: serializeDateTime(saleOrderRecord.data.date_order),
+                product_uom_id: saleOrderLine.product_uom_id.id,
+                company_id: saleOrderRecord.data.company_id.id,
+                pricelist_id: saleOrderRecord.data.pricelist_id.id,
+                ptav_ids: ptavIds,
```
ltinel, line 95:
```suggestion
                ptav_ids: ptavIds,
                only_main_product: edit,
```

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         return this.props.record.data[this.props.name];
     }
 
-    async _onProductTemplateUpdate() {
-        const result = await this.orm.call(
-            'product.template',
-            'get_single_product_variant',
-            [this.props.record.data.product_template_id.id],
+    async _getProductConfiguratorData() {
```
ltinel: You need to add an `edit = false` param to this method, and set the param to `true` when a line is being edited (as opposed to created).

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -186,24 +186,35 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
 
-    async _onProductTemplateUpdate() {
-        const result = await this.orm.call(
-            'product.template',
-            'get_single_product_variant',
-            [this.props.record.data.product_template_id.id],
+    async _getProductConfiguratorData() {
+        const saleOrderRecord = this.props.record.model.root;
+        const saleOrderLine = this.props.record.data;
+        const ptavIds = this._getVariantPtavIds(saleOrderLine);
```
ltinel: This only takes into account the ptavs that create a variant. You need to append the `no_variant` ptavs if `edit` is true.

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -223,15 +234,19 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
-            this._openProductConfigurator(true);
+            this._openProductConfigurator({edit: true, preloadedData: await this._getProductConfiguratorData()});
         }
     }
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator({
+        edit = false,
+        selectedComboItems = [],
+        preloadedData,
```
ltinel:
```suggestion
        data,
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -245,10 +260,17 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
             ptavIds.push(...this._getNoVariantPtavIds(saleOrderLine));
             customPtavs = await this._getCustomPtavs(saleOrderLine);
         }
-
+        const {
+            products,
+            optional_products,
+            currency_id,
+        } = preloadedData;
         this.dialog.add(ProductConfiguratorDialog, {
```
ltinel, line 152: The `ptavIds`, `quantity`, and `productUOMId` props should be removed from the product configurator dialog, so they can also be removed here (and you can drop the computation of `ptavIds` above).

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -291,7 +313,7 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
         });
     }
 
-    async _openComboConfigurator(edit = false, hasOptionalProducts = false) {
+    async _openComboConfigurator(edit = false, hasOptionalProducts = false, preloadedData) {
```
ltinel: Now that we pass the product configurator data in here, we might be able to drop `hasOptionalProducts` :) I believe `hasOptionalProducts` will be true iff the product configurator data contains optional products (to double check).

addons/sale_product_matrix/static/src/js/sale_product_field.js
```diff
@@ -12,7 +12,7 @@ patch(SaleOrderLineProductField.prototype, {
         return this.matrixConfigurator.open(this.props.record, edit);
     },
 
-    async _openProductConfigurator(edit=false, selectedComboItems=[]) {
+    async _openProductConfigurator(edit=false, selectedComboItems=[], preloadedData) {
```
ltinel: I think you'll have to update the rental override as well :)

addons/website_sale/controllers/product_configurator.py
```diff
@@ -8,12 +8,28 @@
...
         website=True,
         readonly=True,
     )
+    def website_sale_product_configurator_get_values(
+        self, product_template_id=None, is_product_configured=False, *args, **kwargs
+    ):
+        self._populate_currency_and_pricelist(kwargs)
+        should_show = self.website_sale_should_show_product_configurator(
+            product_template_id, is_product_configured
+        )
```
ltinel: If you extract the `should_show_product_configurator` computation in a hook (see https://github.com/odoo/odoo/pull/247381/changes#r3373484343), you'll be able to simply override the hook, while leaving `website_sale_product_configurator_get_values` unchanged :)

addons/sale/controllers/product_configurator.py
ltinel: IMO, we should override `get_single_product_variant` in the `website_sale` module, because this method will now be called in `website_sale` as well (via the `/website_sale/product_configurator/get_values` route), and the `has_optional_products` computation isn't correct for eCommerce products at the moment.

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -203,29 +203,36 @@ export class CartService {
...
                 quantity,
                 uomId,
-                ptavs.concat(noVariantAttributeValues),
+                ptavIds,
                 productCustomAttributeValues,
                 {
                     isBuyNow: isBuyNow,
                     isMainProductConfigurable: !isConfigured,
                     showQuantity: showQuantity,
+                    preloadedData: result,
```
ltinel: Could you add the product configurator data (`products`, `optional_products`, and `currency_id`) as parameters to `_openProductConfigurator`?  The `options` dict is meant to only store "real" options.

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -342,9 +349,17 @@ export class CartService {
         additionalData
     ) {
         return await new Promise((resolve) => {
+            const {
+                products,
+                optional_products,
+                currency_id,
+            } = options.preloadedData;
             this.dialog.add(ProductConfiguratorDialog, {
```
ltinel, line 351: Same as https://github.com/odoo/odoo/pull/247381/changes#r3374532530

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -203,29 +203,36 @@ export class CartService {
...
-        );
-        if (shouldShowProductConfigurator) {
+        const ptavIds = ptavs.concat(noVariantAttributeValues);
+        const configDataParams = {
+            product_template_id: productTemplateId,
+            quantity: quantity,
+            ptav_ids: ptavIds,
+            product_uom_id: uomId,
+            so_date: serializeDateTime(DateTime.now()),
+            pricelist_id: odoo.default_pricelist_id,
```
ltinel: I'm not sure we should do this.

addons/sale/static/src/js/sale_product_field.js
```diff
@@ -245,10 +260,17 @@ export class SaleOrderLineProductField extends ProductLabelSectionAndNoteField {
...
+            products,
+            optional_products,
+            currency_id,
+        } = preloadedData;
         this.dialog.add(ProductConfiguratorDialog, {
             productTemplateId: saleOrderLine.product_template_id.id,
             ptavIds: ptavIds,
+            products: products,
+            optional_products: optional_products,
+            currency_id: currency_id,
```
ltinel:
```suggestion
            optionalProducts: optional_products,
            currencyId: currency_id,
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,9 +77,29 @@ export class ComboConfiguratorDialog extends Component {
...
+                products,
+                optional_products,
+                currency_id,
+            } = preloadedData;
             this.dialog.add(ProductConfiguratorDialog, {
                 productTemplateId: product.product_tmpl_id,
                 ptavIds: product.selectedPtavIds,
+                products: products,
+                optional_products: optional_products,
+                currency_id: currency_id,
```
ltinel:
```suggestion
                currencyId: currency_id,
```

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -342,9 +349,17 @@ export class CartService {
...
+                products,
+                optional_products,
+                currency_id,
+            } = options.preloadedData;
             this.dialog.add(ProductConfiguratorDialog, {
                 productTemplateId: productTemplateId,
                 ptavIds: combination,
+                products: products,
+                optional_products: optional_products,
+                currency_id: currency_id,
```
ltinel:
```suggestion
                optionalProducts: optional_products,
                currencyId: currency_id,
```

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -203,29 +203,36 @@ export class CartService {
...
+        const ptavIds = ptavs.concat(noVariantAttributeValues);
+        const configDataParams = {
+            product_template_id: productTemplateId,
+            quantity: quantity,
+            ptav_ids: ptavIds,
+            product_uom_id: uomId,
+            so_date: serializeDateTime(DateTime.now()),
+            pricelist_id: odoo.default_pricelist_id,
+            is_product_configured: isConfigured,
+            ...this._getAdditionalRpcParams?.() || {},
```
ltinel: This method doesn't exist, and I don't think you need it either. The additional params should already be included in `rest` (to double check).

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,32 @@ def sale_product_configurator_get_values(
...
-        currency = self.env["res.currency"].browse(currency_id)
-        pricelist = self.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        res = product_template.get_single_product_variant()
+
+        if self.should_show_product_configurator(res, product_template, **kwargs):
+            combination = self.env["product.template.attribute.value"]
+            if ptav_ids:
+                combination = (
+                    request
```
ltinel:
```suggestion
                    self
```

addons/sale/static/src/js/product_configurator_dialog/product_configurator_dialog.js
```diff
@@ -83,32 +76,18 @@ export class ProductConfiguratorDialog extends Component {
...
-            const {
-                products,
-                optional_products,
-                currency_id,
-            } = await this._loadData(this.props.edit);
-
-            // If the product configurator is opened after the combo configurator (which happens if
-            // a combo product has optional products), `_loadData` will return a single product
-            // (i.e. the combo product), which should be linked to the previously selected combo
-            // items.
```
ltinel, line 96: I think we should keep this comment, but reformulate it to match the code

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -75,34 +75,52 @@ export const saleProductMixin = () => ({
...
-    async _onProductTemplateUpdate() {
-        super._onProductTemplateUpdate();
-        const result = await this.orm.call(
-            "product.template",
-            "get_single_product_variant",
-            [this.props.record.data.product_template_id.id],
+    async _getProductConfiguratorData(edit = false) {
+        const saleOrder = this.props.record.model.root.data;
+        const saleOrderLine = this.props.record.data;
+        const ptavIds = this._getVariantPtavIds(saleOrderLine);
```
ltinel: We need to duplicate the list (to prevent some strange bug)
```suggestion
        const ptavIds = [...this._getVariantPtavIds(saleOrderLine)];
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -75,34 +75,52 @@ export const saleProductMixin = () => ({
...
+        if (edit) {
+            /**
+             * no_variant don't need to be given to the configurator for new
+             * products.
+             */
```
ltinel:
```suggestion
            // no_variant attributes don't need to be given to the configurator for new products.
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
 
     async _onProductUpdate() {}, // event_booth_sale, event_sale, sale_renting
 
-    onEditConfiguration() {
-        super.onEditConfiguration();
+    async onEditConfiguration() {
```
ltinel, line 132: Could you update the method signature for all overrides (i.e. prefix them with `async`)? Unless that's not needed for the override to work properly?

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -75,34 +75,52 @@ export const saleProductMixin = () => ({
...
+                company_id: saleOrder.company_id.id,
+                pricelist_id: saleOrder.pricelist_id.id,
+                ptav_ids: ptavIds,
+                only_main_product: edit,
+                ...this._getAdditionalRpcParams(),
+            });
+    },
+
+    async _onProductTemplateUpdate() {
+        const result = await this._getProductConfiguratorData();
```
ltinel: Nit:
```suggestion
        const data = await this._getProductConfiguratorData();
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
...
         const saleOrderLine = this.props.record.data;
-        const ptavIds = [...this._getVariantPtavIds(saleOrderLine)];
         let customPtavs = [];
 
         if (edit) {
             /**
-             * no_variant and custom attribute don't need to be given to the configurator for new
+             * custom attribute don't need to be given to the configurator for new
              * products.
              */
```
ltinel:
```suggestion
            // custom attributes don't need to be given to the configurator for new products.
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
...
         }
     },
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator({
+        edit = false,
+        selectedComboItems = [],
+        data,
+    } = {}) {
         const saleOrderRecord = this.props.record.model.root;
```
ltinel: Since we always access `data` on this variable, could you include it here? (and remove it wherever the variable is used)
```suggestion
        const saleOrder = this.props.record.model.root.data;
```

addons/sale/static/src/js/combo_configurator_dialog/combo_configurator_dialog.js
```diff
@@ -76,11 +77,27 @@ export class ComboConfiguratorDialog extends Component {
...
+                    company_id: this.props.company_id,
+                    pricelist_id: this.props.pricelist_id,
+                    ptav_ids:  product.selectedPtavIds,
+                    only_main_product: true,
+                    show_packaging: false,
+                    ...this._getAdditionalRpcParams(),
+                });
+            const {
+                products,
+            } = productConfiguratorData;
```
ltinel: Nit:
```suggestion
            const { products } = productConfiguratorData;
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
...
              * products.
              */
-            ptavIds.push(...this._getNoVariantPtavIds(saleOrderLine));
             customPtavs = await this._getCustomPtavs(saleOrderLine);
         }
-
+        const {
+            products,
+            optional_products,
+        } = data;
```
ltinel: Nit:
```suggestion
        const { products, optional_products } = data;
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
...
+            this._openProductConfigurator({edit: true, data: data});
         }
     },
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator({
+        edit = false,
+        selectedComboItems = [],
+        data,
+    } = {}) {
```
ltinel: Nit:
```suggestion
    async _openProductConfigurator({ edit = false, selectedComboItems = [], data } = {}) {
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -182,7 +204,7 @@ export const saleProductMixin = () => ({
         });
     },
 
-    async _openComboConfigurator(edit = false, hasOptionalProducts = false) {
+    async _openComboConfigurator(edit = false, data) {
```
ltinel: Could you do the same as for `_openProductConfigurator`?
```suggestion
    async _openComboConfigurator({ edit = false, data } = {}) {
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -226,15 +248,20 @@ export const saleProductMixin = () => ({
...
         });
     },
 
-    async handleComboSave(comboProductData, selectedComboItems, edit, hasOptionalProducts) {
+    async handleComboSave(
+        comboProductData,
+        selectedComboItems,
+        edit,
+        data,
+    ) {
```
ltinel: Nit:
```suggestion
    async handleComboSave(comboProductData, selectedComboItems, edit, data) {
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -249,11 +276,14 @@ export const saleProductMixin = () => ({
         // Ensure that the order lines are sorted according to their sequence.
         await saleOrder.order_line._sort();
 
-        if (hasOptionalProducts && !edit) {
+        if (data.has_optional_products && !edit) {
```
ltinel: Could you invert this condition (so that we can leave out `data` if `edit` is true):
```suggestion
        if (!edit && data.has_optional_products) {
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,36 +131,40 @@ export const saleProductMixin = () => ({
 
     async _onProductUpdate() {}, // event_booth_sale, event_sale, sale_renting
 
-    onEditConfiguration() {
-        super.onEditConfiguration();
+    async onEditConfiguration() {
+        const data = await this._getProductConfiguratorData(true)
         if (this.isCombo) {
-            this._openComboConfigurator(true);
+            this._openComboConfigurator(true, data);
```
ltinel: If edit is true, we don't need `data`, so could you remove it from here and move the `data` assignment into the `else if` below? (see also https://github.com/odoo/odoo/pull/247381/changes#r3442406743)

addons/sale_product_matrix/static/src/js/sale_product_field.js
ltinel, line 1: I'm not sure you need these changes, since they're already in `addons/sale_product_matrix/static/src/js/sale_product_mixin.js`.

Except for the method signature change for `_openProductConfigurator` (and maybe `onEditConfiguration`) which probably need to be applied in the file mentioned above.

addons/website_sale/controllers/product_configurator.py
```diff
@@ -24,7 +28,11 @@ def website_sale_should_show_product_configurator(
         :rtype: bool
         :return: Whether the product configurator dialog should be shown.
         """
-        product_template = self.env["product.template"].browse(product_template_id)
+        if not request.is_frontend:
+            return super().should_show_product_configurator(
+                single_product_variant_dict, product_template
```
ltinel:
```suggestion
                single_product_variant_dict, product_template, **kwargs
```

addons/website_sale/controllers/product_configurator.py
```diff
@@ -24,7 +28,11 @@ def website_sale_should_show_product_configurator(
...
+            )
+        is_product_configured = kwargs.get("is_product_configured")
```
ltinel: Could you specify it explicitly in the method signature to avoid having to extract it? (and also pass it to super of course)

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -203,19 +203,29 @@ export class CartService {
...
+            quantity: quantity,
+            ptav_ids: ptavIds,
+            product_uom_id: uomId,
+            so_date: serializeDateTime(DateTime.now()),
+            is_product_configured: isConfigured,
+            ...rest
+        };
+
+        const result = await this.rpc('/website_sale/product_configurator/get_values', configDataParams);
+        if (result.products || result.has_optional_products) {
```
ltinel: Why `|| result.has_optional_products`? Isn't checking `result.products` sufficient?

addons/website_sale/static/src/js/cart_service.js
```diff
@@ -203,19 +203,29 @@ export class CartService {
...
+            const {
+                products,
+                optional_products,
+                currency_id,
+            } = result;
```
ltinel: Nit:
```suggestion
            const { products, optional_products, currency_id } = result;
```

addons/website_sale/tests/test_product_configurator.py
```diff
@@ -312,7 +328,7 @@ def test_product_configurator_sale_not_ok(self):
             )
 
         self.assertFalse(show_configurator)
-        self.assertListEqual(configurator_values["optional_products"], [])
+        self.assertFalse(configurator_values["has_optional_products"])
```
ltinel, line 331: Could you also check that `products` and `optional_products` aren't set?
```py
self.assertNotIn("products", configurator_values)
self.assertNotIn("optional_products", configurator_values)
```

addons/sale_product_matrix/static/src/js/sale_product_field.js
ltinel: Could you create an enterprise PR and link it in the task and this PR? :)

addons/sale/controllers/product_configurator.py
ltinel: I guess that we can skip this for now, since we never read `has_optional_products` in eCommerce :)

addons/sale/controllers/product_configurator.py
```diff
@@ -44,30 +44,34 @@ def sale_product_configurator_get_values(
...
-            ).filtered(lambda ptal: ptal.attribute_id.display_type != "multi")
-            combination += unconfigured_ptals.mapped(
-                lambda ptal: ptal.product_template_value_ids._only_active()[:1]
-            )
-        if not combination:
-            combination = product_template._get_first_possible_combination()
-        currency = self.env["res.currency"].browse(currency_id)
-        pricelist = self.env["product.pricelist"].browse(pricelist_id)
-        so_date = datetime.fromisoformat(so_date)
+        if product_template._is_donation():
```
ltinel: This method is defined in `website_sale`, so you can't call it here (you'll get a traceback if `website_sale` isn't installed)

addons/sale_product_matrix/static/src/js/sale_product_mixin.js
```diff
@@ -19,7 +19,7 @@ const saleMatrixProductMixin = () => ({
         return this.matrixConfigurator.open(this.props.record, edit);
     },
 
-    async _openProductConfigurator(edit = false, selectedComboItems = []) {
+    async _openProductConfigurator(edit = false, selectedComboItems = [], preloadedData) {
```
ltinel:
```suggestion
    async _openProductConfigurator({ edit = false, selectedComboItems = [], data } = {}) {
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -75,34 +75,49 @@ export const saleProductMixin = () => ({
...
+            if (this.props.record.data.product_id != data.product_id.id) {
+                if (data.is_combo) {
                     await this.props.record.update({
-                        product_id: { id: result.product_id, display_name: result.product_name },
+                        product_id: { id: data.product_id, display_name: data.product_name },
                     });
-                    this._openComboConfigurator(false, result.has_optional_products);
-                } else if (result.has_optional_products) {
-                    this._openProductConfigurator();
+                    this._openComboConfigurator({edit: false, data});
```
ltinel: Nit:
```suggestion
                    this._openComboConfigurator({edit: false, data: data});
```

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -75,34 +75,49 @@ export const saleProductMixin = () => ({
         return _t("Edit Configuration");
     },
 
-    async _onProductTemplateUpdate() {
-        super._onProductTemplateUpdate();
```
ltinel, line 79: I guess we should keep this `super` call?

addons/website_sale/controllers/product_configurator.py
```diff
@@ -8,32 +8,40 @@
...
+        return super().sale_product_configurator_get_values(*args, **kwargs)
+
+    def should_show_product_configurator(
+        self, single_product_variant_dict, product_template, is_product_configured=False, **kwargs
     ):
         """Return whether the product configurator dialog should be shown.
 
-        :param int product_template_id: The product being checked, as a `product.template` id.
+        :param dict single_product_variant_dict: Information about the single product variant.
+        :param recordset product_template: The product.template record being checked.
```
ltinel: Nit:
```suggestion
        :param product.template product_template: The product.template record being checked.
```

addons/website_sale/controllers/product_configurator.py
```diff
@@ -8,32 +8,40 @@
...
         :param bool is_product_configured: Whether the product is already configured.
         :rtype: bool
         :return: Whether the product configurator dialog should be shown.
         """
-        product_template = self.env["product.template"].browse(product_template_id)
-        if product_template._is_donation():
-            return False
```
ltinel, line 29: I think we still need this.

addons/sale/static/src/js/sale_product_mixin.js
```diff
@@ -113,40 +128,34 @@ export const saleProductMixin = () => ({
 
     async _onProductUpdate() {}, // event_booth_sale, event_sale, sale_renting
 
-    onEditConfiguration() {
-        super.onEditConfiguration();
```
ltinel, line 117: Same here: we should probably keep the `super` call

addons/website_sale/controllers/product_configurator.py
csan-odoo: After the changes in this PR, the only path to this method is through the following code in the sale module:
```python
if company_id:
    request.update_context(allowed_company_ids=[company_id])
product_template = self._get_product_template(product_template_id)

res = product_template.get_single_product_variant()

if self.should_show_product_configurator(res, product_template, **kwargs):
```
Before should_show_product_configurator is called, the code already needs to fetch the product_template. However, for donation products, the template is not accessible to a public user. As a result, the call to get_single_product_variant() will fail because it needs to access the template internally.

Because of that, I think it would make more sense to perform the donation check before calling should_show_product_configurator(). That way, whenever we reach this method, we already know we're dealing with a non-donation product.

ltinel: Sure, in that case you can add the check at the start of the `website_sale_product_configurator_get_values` method :)

ltinel:
@robodoo squash

ltinel:
@robodoo r+

PR #247387, opened 2026-02-05
 [IMP] pos_self_order: compute delivery distance

vlst-odoo:
In the pos_self_order app, customers have the ability to choose the
delivery option when ordering. The app then asks for the customer's
address, but it does not perform any checks on this address.

In this commit we add logic that computes the distance between the
restaurant and the customer and restricts ordering if this distance
exceeds a certain limit.

The restaurant owner can now configure the max distance and pricing
rules for delivery.

In order to make it simpler for the customer to add their
address, we replace the old address fields ( street, zip, city, country
) with the `AddressAutoComplete` odoo field.

In order to compute the distance, we rely on the `base_geolocalize`
module to get the lat and long of the address and we then compute the
straigh line distance to the restaurant's address.

Task: 5913205

---


addons/pos_self_order/controllers/orders.py
```diff
@@ -7,6 +12,37 @@
 from odoo.tools import consteq
 
 
+def _haversine_distance(lat1, long1, lat2, long2):
+    """Compute the straight-line distance in km between two lat/long points."""
+    R = 6371
```
adgu-odoo: Can you put a comment to why this R is 6371 (something like `# earth's radius in km`)

addons/pos_self_order/controllers/orders.py
```diff
@@ -7,6 +12,37 @@
...
+        referrer = request.httprequest.referrer or ''
+        if isinstance(referrer, bytes):
+            referrer = referrer.decode('utf-8', errors='replace')
+        parsed_url = urlparse(referrer)
+        if not parsed_url.path.startswith('/pos-self'):
+            return super()._get_api_key(use_employees_key)
+        access_token = parse_qs(parsed_url.query).get('access_token', [None])[0]
+        pos_config_id = int(parsed_url.path.split('/')[2])
+        pos_config_sudo = request.env['pos.config'].sudo().browse(pos_config_id)
+        if secrets.compare_digest(pos_config_sudo.access_token, access_token):
```
adgu-odoo: This looks like a `_verify_pos_config`, isn't it possible to use the method `_verify_pos_config`. Something like creating an endpoint in the `PosSelfOrderController` like `/pos-self/autocomplete/address` that would take an `access_token`. You could then verify the access_token and then call `AutoCompleController._perform_place_search` with the api_key directly. What do you think ? We can talk about it if you want

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+        if not delivery_product_tmpl:
+            return
+        delivery_variant = delivery_product_tmpl.product_variant_ids[:1]
+        if not delivery_variant:
+            return
+
+        non_delivery_lines = order.lines.filtered(
+            lambda l: l.product_id.product_tmpl_id != delivery_product_tmpl
+        )
+        non_delivery_total = sum(l.price_subtotal_incl for l in non_delivery_lines)
```
adgu-odoo: if they work tax excluded, maybe we should consider the total of line.price_subtotal instead of price_subtotal_incl

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+            existing_delivery_lines.unlink()
+        elif not existing_delivery_lines:
+            price = preset.delivery_product_price
+            order.env['pos.order.line'].sudo().create({
+                'order_id': order.id,
+                'product_id': delivery_variant.id,
+                'qty': 1,
+                'price_unit': price,
+                'price_subtotal': price,
+                'price_subtotal_incl': price,
```
adgu-odoo: This does not take into account a tax that could be set on the product. In several countries, tax should be applied on the delivery fee too

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+                'full_product_name': 'Delivery fee',
```
adgu-odoo: You could take the name of the delivery product IMO

addons/pos_self_order/controllers/orders.py
adgu-odoo: You could add the order line without price_subtotal and price_subtotal incl set and then call `_compute_line_price` with an extra arguments `price` for example. The method would become
```python
    def _compute_line_price(self, line, price=False):
        ...
        price = price or pricelist._get_product_price(product, 1.0, currency=self.currency_id)
        ...
```

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
-            'zip': zip,
-            'city': city,
-            'country_id': country_id.id,
-            'state_id': state_id.id if state_id else False,
-            'company_id': pos_config.company_id.id,
-        })
+        preset = pos_config.env['pos.preset'].browse(int(preset_id)) if preset_id else False
+        partner = pos_config.env['res.partner'].sudo().browse(int(partner_id)) if partner_id else False
+
+        if not partner:
```
adgu-odoo:
```suggestion
        if not partner or not partner.exists():
```

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+            if error:
+                return {'error': error}
 
         return {
-            'res.partner': self.env['res.partner']._load_pos_self_data_read(partner_sudo, pos_config),
+            'res.partner': self.env['res.partner']._load_pos_self_data_read(partner, pos_config),
         }
 
+    def _check_delivery_address_for_partner(self, preset, partner):
+        partner.geo_localize()
```
adgu-odoo:
```suggestion
        if not parner.partner_latitude and not partner.partner_longitude:
            partner.geo_localize()
```

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+        if partner.partner_latitude is False or partner.partner_longitude is False:
```
adgu-odoo: I think that those fields are Float and they default to 0, moreover, when changing the address of the partner, they are reset to 0.0 and if they cannot be localized, they will not be set again. Last thing, let's accept it if one of the two coordinate is 0.0 but not if both are 0.0
```suggestion
        if not partner.partner_latitude and not partner.partner_longitude:
```

addons/pos_self_order/controllers/orders.py
```diff
@@ -48,34 +91,90 @@ def _generate_return_values(self, order, config):
...
+                'type': 'address',
+                'message': self.env._("We couldn't locate this address. Please enter a complete address with a street number."),
+            }
+        if not preset.delivery_from_address:
+            return None
+        distance_km = _haversine_distance(
+            partner.partner_latitude, partner.partner_longitude,
+            preset.delivery_from_latitude, preset.delivery_from_longitude,
+        )
+        max_distance = preset.delivery_max_distance_km or 0
```
adgu-odoo: I think float default to 0 so it is not necessary to put `or 0`

addons/pos_self_order/data/preset_data.xml
```diff
@@ -1,5 +1,14 @@
 <odoo>
     <data noupdate="1">
+        <record id="product_delivery_template" model="product.template">
+            <field name="name">Delivery</field>
```
adgu-odoo: There is already a delivery product in the data of the delivery module, what do you think about making it explicit that we created this product for delivery fee of the self order. I'm not sure this is a good idea but we could talk about it with mobt I think
```suggestion
            <field name="name">Delivery Fee (Self-order)</field>
```

addons/pos_self_order/models/pos_preset.py
```diff
@@ -11,13 +17,130 @@ class PosPreset(models.Model):
...
+        string="State",
+    )
+    delivery_from_latitude = fields.Float(string="Latitude", digits=(10, 7))
+    delivery_from_longitude = fields.Float(string="Longitude", digits=(10, 7))
+    delivery_max_distance_km = fields.Float(
+        string="Max radius",
+        default=10.0,
+        help="Maximum delivery distance, in the unit shown next to the field.",
+    )
+    delivery_distance_unit = fields.Char(
```
adgu-odoo: This could be a selection field with km and miles and by default put the unit depending on IMPERIAL_COUNTRY_CODE but the user could chose in the form view what unit he wants to use. I discussed about it with mobt and he likes the idea

addons/pos_self_order/models/pos_preset.py
```diff
@@ -11,13 +17,130 @@ class PosPreset(models.Model):
...
+        for preset in self:
+            country_code = preset.delivery_from_country_id.code or company_country_code
+            preset.delivery_distance_unit = 'mi' if country_code in _IMPERIAL_COUNTRY_CODES else 'km'
+
+    @api.model
+    def default_get(self, fields):
+        defaults = super().default_get(fields)
+        if defaults.get('delivery_from_address'):
+            return defaults
+        address_fields = {'delivery_from_address', 'delivery_from_city', 'delivery_from_zip',
```
adgu-odoo: You can use ADDRESS_FIELDS

addons/pos_self_order/models/pos_preset.py
```diff
@@ -11,13 +17,130 @@ class PosPreset(models.Model):
...
+                    street=company.street,
+                    zip=company.zip or '',
+                    city=company.city or '',
+                    state=company.state_id.name if company.state_id else '',
+                    country=company.country_id.name if company.country_id else '',
+                ),
+                force_country=company.country_id.name if company.country_id else '',
+            )
+        except Exception:
+            return defaults
```
adgu-odoo, line 105: Instead of calling everytime you create a new record, you could check if any other record in delivery has an address and if yes, put this one, else do a geo_find. This would reduce the number of api request

addons/pos_self_order/models/pos_preset.py
```diff
@@ -26,13 +149,14 @@ def _load_pos_self_data_domain(self, data, config):
...
-        params.extend(['service_at', 'mail_template_id'])
+        params.extend(['service_at', 'mail_template_id', 'free_delivery_min_amount', 'delivery_product_id', 'delivery_product_price'])
         return params
     
+
     @api.model
     def _load_pos_data_fields(self, config):
         params = super()._load_pos_data_fields(config)
-        params.extend(['mail_template_id'])
+        params.extend(['mail_template_id', 'free_delivery_min_amount'])
```
adgu-odoo: Is this really needed, I don't think you use this field in the pos

addons/pos_self_order/models/pos_preset.py
```diff
@@ -11,13 +17,130 @@ class PosPreset(models.Model):
...
+        string="Max radius",
+        default=10.0,
+        help="Maximum delivery distance, in the unit shown next to the field.",
+    )
+    delivery_distance_unit = fields.Char(
+        string="Distance unit",
+        compute="_compute_delivery_distance_unit",
+    )
+    delivery_product_id = fields.Many2one(
+        comodel_name="product.template",
```
adgu-odoo: Why is this not a `product.product` ? It would make more sense to put a record and not a template IMO

adgu-odoo: It would also make a lot of usage of this more direct I think

addons/pos_self_order/models/product_template.py
```diff
@@ -0,0 +1,27 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class ProductTemplate(models.Model):
```
adgu-odoo: There is already an override of product_template in the file product_product. It is true that we should create a file product_template, can you move the inherit of the class product_template from product_product to this file too ?

addons/pos_self_order/models/product_template.py
```diff
@@ -0,0 +1,27 @@
...
+    _inherit = 'product.template'
+
+    def _load_pos_self_data_search_read(self, data, config):
+        result = super()._load_pos_self_data_search_read(data, config)
```
adgu-odoo: I get the idea of adding the product in the list if it was not already loaded but I don't like that you have to recall code that is already done in `_load_pos_self_data_read` in order to do that. What about you just add the ids that you need in the `_load_pos_self_data_domain` method such that everything that is needed to do after that is automatically done and you do not need to recall everything

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
-import { useState } from "@web/owl2/utils";
-import { Component, onWillStart } from "@odoo/owl";
+import { Component, onWillStart, useState, EventBus } from "@odoo/owl";
```
adgu-odoo: keep the import useState as it was `import { useState } from "@web/owl2/utils";`

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
...
+    }
 
-            const partner = this.selfOrder.models.connectNewData(result);
+    setPartnerAndOrderName(partner) {
+        if (this.preset.needsPartner) {
             this.selfOrder.currentOrder.floating_order_name = `${this.preset.name} - ${partner["res.partner"][0].name}`;
-            this.selfOrder.currentOrder.partner_id = partner["res.partner"][0];
         } else {
-            this.selfOrder.currentOrder.floating_order_name = this.state.name;
+            this.selfOrder.currentOrder.floating_order_name = partner["res.partner"][0].name;
```
adgu-odoo: since you are not using anything else than `"res.partner"` from `partner`, pass directly the `partner["res.partner"][0]` value to this function

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
...
+                    }
+                }
+                this.state.validationError = null;
+            },
+        };
+    }
+
+    get availablePresets() {
+        const currentPresetId = this.selfOrder.currentOrder?.preset_id?.id;
+        return this.selfOrder.models["pos.preset"].getAll().filter((p) => p.id !== currentPresetId);
```
adgu-odoo:
```suggestion
        return this.selfOrder.models["pos.preset"].filter((p) => p.id !== currentPresetId);
```

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
...
+                    } else {
+                        this.state[key] = value || "";
+                    }
+                }
+                this.state.validationError = null;
+            },
+        };
+    }
+
+    get availablePresets() {
```
adgu-odoo: Can you find a name that represent what this method does a bit better ? Something like `otherPresets` or something like that ?

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
...
+        const currentPreset = this.selfOrder.currentOrder?.preset_id;
+        // Remove delivery lines when switching away from delivery
+        if (currentPreset?.service_at === "delivery") {
+            const deliveryTmplId = currentPreset.delivery_product_id?.id;
+            if (deliveryTmplId) {
+                const toDelete = this.selfOrder.currentOrder.lines.filter(
+                    (l) => l.product_id?.product_tmpl_id?.id === deliveryTmplId
+                );
+                for (const line of toDelete) {
+                    line.delete();
```
adgu-odoo:
```suggestion
                    this.selfOrder.removeLine(line);
```

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -1,82 +1,279 @@
...
+                Object.keys(options.categories[dateId].subCategories).length === 0 ||
+                Object.values(options.categories[dateId].subCategories).every(
+                    (subCateg) => subCateg.options.length === 0
+                )
+            ) {
+                delete options.categories[dateId];
+            }
+        }
+        return options;
+    }
```
adgu-odoo: this looks like a copy paste from cart_page for a big part, can you extract a part in self_order_service and call a method to do this ?

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -101,27 +298,42 @@ export class PresetInfoPopup extends Component {
...
-        return this.selfOrder.models["res.country"].getAll();
+    flagEmoji(code) {
+        return [...code.toUpperCase()]
+            .map((c) => String.fromCodePoint(0x1f1e6 + c.charCodeAt(0) - 65))
+            .join("");
+    }
+
+    get allCountries() {
+        return this.selfOrder.models["res.country"]
+            .getAll()
```
adgu-odoo:
```suggestion
```

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.xml
```diff
@@ -2,24 +2,24 @@
...
             title="this.props.title">
             <div class="select_popup_preset_info w-100 d-flex flex-column h-100 overflow-hidden">
                 <div class="d-flex justify-content-between align-items-center px-2 mt-2 mb-4">
                     <h2 class="mb-0">We need more info</h2>
                     <button class="btn-close" aria-label="close" t-on-click="this.close"/>
                 </div>
-                <div class="d-flex flex-column gap-3 overflow-y-auto overflow-x-hidden flex-grow-1">
-                    <select t-if="this.preset.needsPartner and this.existingPartners.length"
+                <div class="d-flex flex-column gap-3 overflow-y-auto overflow-x-hidden flex-grow-1 pb-2">
+                    <select t-if="preset.needsPartner and existingPartners.length"
```
adgu-odoo: why do you remove the `this` ?

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.xml
```diff
@@ -2,24 +2,24 @@
...
                         class="partner-select form-select form-select-lg"
                         t-on-change="this.selectExistingPartner"
                         t-custom-model="state.selectedPartnerId">
                         <option value="">
                             Select your address
                         </option>
-                        <t t-foreach="this.existingPartners" t-as="partner" t-key="partner.id">
-                            <option t-att-value="partner.id" t-out="partner.name + ' ' + partner.address" />
+                        <t t-foreach="existingPartners" t-as="partner" t-key="partner.id">
```
adgu-odoo: same

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.xml
```diff
@@ -33,55 +33,53 @@
...
-                                    <t t-if="this.partnerIsSelected">*************</t>
-                                    <t t-else="">Select a state</t>
-                                </option>
-                                <t t-foreach="availableState" t-as="state" t-key="state.id">
-                                    <option t-att-value="state.id" t-out="state.name" />
+                        t-att-disabled="partnerIsSelected" />
+                    <div t-if="preset.needsName || preset.needsPartner" class="input-group input-group-lg">
+                        <!-- Country prefix picker: transparent select overlays a styled display div -->
+                        <div class="position-relative" style="max-width: 25%; min-width: 90px">
+                            <div class="form-select h-100 d-flex align-items-center gap-1 pe-none user-select-none">
```
adgu-odoo: with `form-select` you have a `background-image` that display a caret down icon. In mobile, this caret down is not displayed in a good way, I think in this case it is not needed and you can remove it with `background-image: none`

addons/pos_self_order/static/src/app/services/self_order_service.js
adgu-odoo, line 1: I think you should also override `verifyCart` to accept the delivery product even if it is not self_order_available

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -1,11 +1,12 @@
 import { useRef, useState } from "@web/owl2/utils";
-import { Component } from "@odoo/owl";
+import { Component, useEffect } from "@odoo/owl";
```
adgu-odoo: use `useLayoutEffect`, it does the same thing but will help a bit the js framework team for owl 3

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -23,11 +24,21 @@ export class CartPage extends Component {
...
-            showOrderNote: this.orderNote,
             orderNoteValue: "",
         });
 
         this.scrollShadow = useScrollShadow(useRef("scrollContainer"));
+        useEffect(
+            () => this.selfOrder.ensureDeliveryLine(),
+            () => {
+                const order = this.selfOrder.currentOrder;
+                const nonDeliveryId = order?.preset_id?.delivery_product_id?.id;
```
adgu-odoo: order should never be undefined. currentOrder getter is creating an order when there is no existing order

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -23,11 +24,21 @@ export class CartPage extends Component {
...
+                const nonDeliveryTotal = order?.lines
+                    ?.filter((l) => l.product_id?.product_tmpl_id?.id !== nonDeliveryId)
+                    .reduce((sum, l) => sum + (l.qty || 0) * (l.price_unit || 0), 0);
+                return [order?.preset_id?.id, nonDeliveryTotal];
```
adgu-odoo: I think some of those `?` are not necessary
```suggestion
                const nonDeliveryTotal = order.lines
                    .filter((l) => l.product_id.product_tmpl_id.id !== nonDeliveryId)
                    .reduce((sum, l) => sum + (l.qty || 0) * (l.price_unit || 0), 0);
                return [order.preset_id?.id, nonDeliveryTotal];
```

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -375,6 +395,11 @@ export class CartPage extends Component {
         return !this.selfOrder.isTaxesIncludedInPrice();
     }
 
+    isDeliveryLine(line) {
+        const deliveryTmplId = this.selfOrder.currentOrder?.preset_id?.delivery_product_id?.id;
```
adgu-odoo:
```suggestion
        const deliveryTmplId = this.selfOrder.currentOrder.preset_id?.delivery_product_id?.id;
```

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -375,6 +395,11 @@ export class CartPage extends Component {
...
+        return deliveryTmplId && line.product_id?.product_tmpl_id?.id === deliveryTmplId;
```
adgu-odoo:
```suggestion
        return deliveryTmplId && line.product_id.product_tmpl_id.id === deliveryTmplId;
```

addons/pos_self_order/static/src/app/services/self_order_service.js
```diff
@@ -233,6 +233,50 @@ export class SelfOrder extends Reactive {
...
+        }
+        const deliveryTemplate = preset.delivery_product_id;
+        if (!deliveryTemplate) {
+            return;
+        }
+        const existingLine = this.currentOrder.lines.find(
+            (line) => line.product_id?.product_tmpl_id?.id === deliveryTemplate.id
+        );
+        if (this._shouldDeliveryBeFree()) {
+            existingLine?.delete();
```
adgu-odoo:
```suggestion
            this.removeOrderline(existingLine);
```

addons/pos_self_order/views/pos_preset_view.xml
```diff
@@ -11,6 +11,25 @@
                         <group>
                             <field name="available_in_self" widget="boolean_toggle"/>
                             <field name="service_at" invisible="not available_in_self" />
+                            <field
```
adgu-odoo, line 14: you can do `<t t-set="not_delivery_preset" t-value="not available_in_self or service_at != 'delivery'"+>` and use it everytime, it will be more readable imo

addons/pos_self_order/__manifest__.py
```diff
@@ -59,6 +59,22 @@
...
+            "web/static/src/views/fields/char/*",
+            "web/static/src/views/fields/translation_button.*",
+            "web/static/src/views/fields/translation_dialog.*",
+            "web/static/src/views/fields/dynamic_placeholder_hook.js",
+            "web/static/src/views/fields/dynamic_placeholder_popover.*",
+            "web/static/src/views/fields/formatters.js",
+            "web/static/src/views/fields/standard_field_props.js",
+            "web/static/src/views/fields/input_field_hook.js",
+            "web/static/src/core/model_field_selector/model_field_selector_popover.js",
+            "web/static/src/core/model_field_selector/model_field_selector_popover.xml",
```
adgu-odoo, line 76: I didn't check all those imports but I trust you on those

adgu-odoo (commented):
Good work, I have a few comments after this first pass, don't hesitate to ask me if you don't understand some comments

addons/pos_self_order/controllers/orders.py
```diff
@@ -185,6 +259,22 @@ def _get_order_prices(self, lines):
...
+    @http.route('/pos-self/autocomplete/address', methods=['POST'], type='jsonrpc', auth='public', website=True)
+    def pos_self_order_autocomplete_address(self, access_token, partial_address, session_id=None, **kwargs):
+        self._verify_pos_config(access_token)
+        api_key = request.env['ir.config_parameter'].sudo().get_str('google_address_autocomplete.google_places_api_key') or None
+        if not api_key:
+            return {'results': [], 'session_id': session_id}
+        return AutoCompleteController._perform_place_search(self, partial_address, api_key=api_key, session_id=session_id)
+
+    @http.route('/pos-self/autocomplete/address_full', methods=['POST'], type='jsonrpc', auth='public', website=True)
+    def pos_self_order_autocomplete_address_full(self, access_token, address, google_place_id=None, session_id=None, **kwargs):
```
adgu-odoo, line 271: We never use session_id if I'm not wrong so I think so we can remove it from the args and from the call to `_perform_complete_place_search`

adgu-odoo: Same on the method above

addons/pos_self_order/controllers/orders.py
```diff
@@ -47,34 +70,85 @@ def _generate_return_values(self, order, config):
...
+    def _ensure_delivery_fee(self, order, preset):
+        """Add or remove the delivery fee line based on the order total and preset configuration."""
+        delivery_product = preset.delivery_product_id
+        if not delivery_product:
+            return
+
+        non_delivery_lines = order.lines.filtered(
+            lambda l: l.product_id != delivery_product
+        )
+        non_delivery_total = sum(l.price_subtotal for l in non_delivery_lines)
```
adgu-odoo, line 82: I might not have been clear on my previous comment, I think we should sum the line `price_subtotal` when the company is working tax excluded and `price_subtotal_incl` when the company is working tax included

addons/pos_self_order/models/pos_preset.py
adgu-odoo: Did you check if it was needed ?

addons/pos_self_order/controllers/orders.py
```diff
@@ -49,34 +72,89 @@ def _generate_return_values(self, order, config):
...
+        non_delivery_total = sum(non_delivery_lines.mapped(total_field))
+
+        free_min = preset.free_delivery_min_amount
+        delivery_is_free = bool(free_min) and non_delivery_total >= free_min
+
+        existing_delivery_lines = order.lines.filtered(
+            lambda l: l.product_id == delivery_product
+        )
+        if delivery_is_free:
+            existing_delivery_lines.unlink()
```
adgu-odoo, line 93: In the future, we could check to add the line but with a 100% to show that the delivery fee is free

addons/pos_self_order/controllers/orders.py
```diff
@@ -49,34 +72,89 @@ def _generate_return_values(self, order, config):
...
+                'phone': phone,
+                'street': street,
+                'zip': zip,
+                'city': city,
+                'country_id': country_id,
+                'state_id': state_id,
+                'company_id': pos_config.company_id.id,
+            })
+
+        if preset and preset.service_at == 'delivery':
```
adgu-odoo:
```suggestion
        if preset and preset.exists() and preset.service_at == 'delivery':
```

addons/pos_self_order/models/pos_preset.py
```diff
@@ -11,13 +17,132 @@ class PosPreset(models.Model):
...
         comodel_name='mail.template',
         domain="[('model', '=', 'pos.order')]",
     )
 
+    @api.model
+    def default_get(self, fields):
+        defaults = super().default_get(fields)
+        if not set(ADDRESS_FIELDS).intersection(fields):
+            return defaults
+        existing = self.search([('service_at', '=', 'delivery'), ('delivery_from_address', '!=', False)], limit=1)
```
adgu-odoo: I'm not sur but I think you have to do this to not get other companies preset and AccessError later
```suggestion
        existing = self.search([*self._check_company_domain(self.env.company), ('service_at', '=', 'delivery'), ('delivery_from_address', '!=', False)], limit=1)
```

addons/pos_self_order/models/product_product.py
```diff
@@ -1,93 +1,24 @@
...
-        if field_name in ["image_512", "image_128"] and self.sudo().self_order_available:
-            return True
-        return super()._can_return_content(field_name, access_token)
-
-
-class ProductProduct(models.Model):
-    _inherit = "product.product"
+        # Also include delivery product variants whose templates are being loaded
+        loaded_tmpl_ids = {t['id'] for t in data.get('product.template', [])}
+        delivery_variants = self.env['pos.preset'].sudo().search([
```
adgu-odoo: is this sudo necessary ?

adgu-odoo: If you add the delivery product template id in `_load_pos_self_data_domain` of product_template.py, is it necessary to add the ids of the variants here as their template will already be loaded and I think we load every product variant which product template is loaded. Can you check ?

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -36,50 +118,202 @@ export class PresetInfoPopup extends Component {
...
+                this.setPartnerAndOrderName(connectedData["res.partner"][0]);
+            } else {
+                this.selfOrder.currentOrder.floating_order_name = this.state.name;
+            }
+
+            if (this.preset.needsSlot && this.state.selectedSlot) {
+                this.selfOrder.currentOrder.preset_time = DateTime.fromSQL(this.state.selectedSlot)
+                    .toUTC()
+                    .toFormat("yyyy-MM-dd HH:mm:ss");
+            }
```
adgu-odoo: Why do you need to do this ? Isn't it done in selectPreset ?

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -36,50 +118,202 @@ export class PresetInfoPopup extends Component {
...
+                country_id: "countryId",
+                state_id: "stateId",
+            },
+        };
+    }
 
-            const partnerId = result?.["res.partner"]?.[0]?.id;
-            if (!partnerId) {
-                return;
+    selectExistingPartner(event) {
```
adgu-odoo: This has been removed in master, I don't think it's a good idea to put it back https://github.com/odoo/odoo/pull/256158

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.xml
```diff
@@ -2,15 +2,26 @@
...
+            contentClass="'w-100 m-0 position-absolute self_order_pills_selection_popup_dialog'"
             bodyClass="'p-4 pb-3 m-0'"
             title="this.props.title">
             <div class="select_popup_preset_info w-100 d-flex flex-column h-100 overflow-hidden">
                 <div class="d-flex justify-content-between align-items-center px-2 mt-2 mb-4">
                     <h2 class="mb-0">We need more info</h2>
                     <button class="btn-close" aria-label="close" t-on-click="this.close"/>
                 </div>
-                <div class="d-flex flex-column gap-3 overflow-y-auto overflow-x-hidden flex-grow-1">
+                <div class="d-flex flex-column gap-3 overflow-y-auto overflow-x-hidden flex-grow-1 pb-2">
```
adgu-odoo: This has been removed in master, I don't think it's a good idea to put it back https://github.com/odoo/odoo/pull/256158

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -36,50 +118,202 @@ export class PresetInfoPopup extends Component {
...
         }
-        this.props.getPayload({ ...this.state, phone: this.getFullPhone() });
-        this.props.close();
+    }
+
+    get existingPartners() {
+        return this.selfOrder.models["res.partner"].getAll();
+    }
+
+    get partnerIsSelected() {
```
adgu-odoo: I think this can be removed too

addons/pos_self_order/static/src/app/components/preset_info_popup/preset_info_popup.js
```diff
@@ -36,50 +118,202 @@ export class PresetInfoPopup extends Component {
...
+            }
+        }
+    }
+
+    get addressAutoCompleteProps() {
+        return {
+            record: this.addressRecord,
+            name: "street",
+            placeholder: "Address",
+            readonly: Boolean(this.partnerIsSelected),
```
adgu-odoo: This will never be readonly if we remove the selectedPartner

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
```diff
@@ -28,6 +28,17 @@ export class CartPage extends Component {
         });
 
         this.scrollShadow = useScrollShadow(useRef("scrollContainer"));
+        useEffect(
```
adgu-odoo: Can you use `useLayoutEffect`

addons/pos_self_order/static/src/app/pages/cart_page/cart_page.js
adgu-odoo, line 89: You can rebase to remove this

addons/pos_self_order/static/src/app/services/self_order_service.js
```diff
@@ -268,6 +268,98 @@ export class SelfOrder extends Reactive {
...
+        const freeDeliveryMin = preset?.free_delivery_min_amount || 0;
+        if (!freeDeliveryMin) {
+            return false;
+        }
+        const deliveryProductId = preset?.delivery_product_id?.id;
+        const nonDeliveryLines = this.currentOrder.lines.filter(
+            (line) => line.product_id?.id !== deliveryProductId
+        );
+        const totalData = this.currentOrder.getPriceWithOptions({ lines: nonDeliveryLines });
+        const orderTotal = this.currency.round(totalData.taxDetails.total_amount_no_rounding);
```
adgu-odoo: Does this check the iface_tax_included ? If yes you can keep it like this, otherwise, check total_included or total_excluded depending on iface_tax_included like in python

addons/pos_self_order/static/src/app/services/self_order_service.js
```diff
@@ -842,10 +934,14 @@ export class SelfOrder extends Reactive {
         let result = true;
         const unavailableProducts = new Set();
 
+        const deliveryProductId = this.currentOrder?.preset_id?.delivery_product_id?.id;
```
adgu-odoo:
```suggestion
        const deliveryProductId = this.currentOrder.preset_id?.delivery_product_id?.id;
```

addons/pos_self_order/views/pos_preset_view.xml
```diff
@@ -11,6 +11,26 @@
...
+                                name="delivery_from_address"
+                                widget="google_address_autocomplete"
+                                invisible="not available_in_self or service_at != 'delivery'"
+                                options="{'street': 'delivery_from_street', 'city': 'delivery_from_city', 'zip': 'delivery_from_zip', 'state_id': 'delivery_from_state_id', 'country_id': 'delivery_from_country_id'}"
+                            />
+                            <field name="delivery_from_street" invisible="1" />
+                            <field name="delivery_from_address" invisible="1" />
```
adgu-odoo, line 21: not sure this line is needed, should be tested withtou

adgu-odoo (approved):

adgu-odoo:
robodoo delegate+

vlst-odoo:
robodoo r+

PR #247390, opened 2026-02-05
[FIX] hr_timesheet: make sure we search favorite project when needed.

xavierbol:
Before this commit, each time the default_get method is called for `account.analytic.line` model, the favorite project is computed even if we don't expect to search the default value for `project_id` field.

This commit makes sure we will only compute the default project when it is really needed.


xavierbol:
robodoo r+

PR #247393, opened 2026-02-05
[FIX] urls: fix typing error on _contains_dot_segments

antonrom1:
_contains_dot_segments returns a boolean, not str

opw-wama-raleur-du-typing

### DIFF
--- a/odoo/tools/urls.py
+++ b/odoo/tools/urls.py
@@ -4,7 +4,7 @@ import urllib.parse
 __all__ = ['urljoin']
 
 
-def _contains_dot_segments(path: str) -> str:
+def _contains_dot_segments(path: str | bytes) -> bool:
     # most servers decode url before doing dot segment resolutions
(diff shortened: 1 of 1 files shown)


odoo/tools/urls.py
```diff
@@ -4,7 +4,7 @@
 __all__ = ['urljoin']
 
 
-def _contains_dot_segments(path: str) -> str:
+def _contains_dot_segments(path: str) -> bool:
```
mwath: Fake news according to the [documentation](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote):
> string may be either a str or a bytes object.


```suggestion
def _contains_dot_segments(path: str | bytes) -> bool:
```

antonrom1: happy now?

mwath: Yes :clown_face:

mwath (approved):
Aweomse tyipng, thakns!

PR #247395, opened 2026-02-05
[FIX] website_sale: remove unwanted eCommerce query param

ltinel:
An `extra_domain` query param was wrongly added to the shop page's query params [1], polluting eCommerce URLs. This param was only ever used to tweak the search options, so it can simply be added directly to the search options instead of the query params.

[1] https://github.com/odoo/odoo/commit/07ea9d2d716770fec0d0097d211047e32216090d


Feyensv (approved):
robodoo delegate+

PR #247397, opened 2026-02-05
[IMP] web: display raw value for unknown selection options

HydrionBurst:
When a selection field contains a value not found in the current options list, display the raw value instead of leaving it blank. This behavior aligns with the backend handling for record values and exported data.

Current behavior before PR:

---


aab-odoo (approved):
robodoo r+

PR #210142, opened 2025-05-15
[PERF] stock: compute the child locations once per move in putaway

Beekan:
Previously when validating a movement with a location that is used in a rule, move_dest_ids was set which led to it going to check the putaway strategy. The bottleneck was the computation of the computed field child_internal_location_ids which is the same across all smls.

This PR utilizes the fact that the smls in the same stock_move would have the same destination location and by proxy the same child_internal_location_ids

Speed up:

1000 stock move lines:
Before:  2 mins
After: 1 mins

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247403, opened 2026-02-05
[FIX] stock: make putaway strategy default only to child locations

vandroogenbd:
Steps to reproduce
-----
- Enable locations & by products
- Create 4 locations
	- View A of type view (parent: Stock)
	- View B of type view (parent: Stock)
	- Storage A of type internal location (parent: View A)
	- Storage B of type internal location (parent: View B)
- Create a "Manufacture to A" route
	- Rule 1: Manufacture, Stock -> Stock
	- Rule 2: Push To, Internal Transfer, Stock -> View A
- Create a "By Products to B" route
	- Rule 1: Push To, Internal Transfer, Stock -> View B
- Create a "Bonus" product with route "By Products to B"
- Create a "Finished" product with route "Manufacture to A"
	- BoM with "Bonus" as by product
- Create a MO for Finished, and produce it
- Open the linked transfers' list view
- Open the "Bonus" transfer

> "Destination Location" states View B

- Open the line's details (hamburger button)

> "Store To" states View A/Storage A

Cause
-----
When retrieving the location in

https://github.com/odoo/odoo/blob/26dbbdaf460a91ef4b33392c27a28951d5de2c57/addons/stock/models/stock_move_line.py#L280-L282

we pass `locations` as a context key. The problem is that `locations` contains the childs of **all** of the SMLs' locations

https://github.com/odoo/odoo/blob/26dbbdaf460a91ef4b33392c27a28951d5de2c57/addons/stock/models/stock_move_line.py#L260

This means that, in `_get_putaway_strategy`, when no `putaway_location` is found, we end up defaulting to the first element of the list

https://github.com/odoo/odoo/blob/26dbbdaf460a91ef4b33392c27a28951d5de2c57/addons/stock/models/stock_location.py#L370-L371

which might not be a child of the location.

Solution
-----
By removing the context key, locations get populated as such

https://github.com/odoo/odoo/blob/26dbbdaf460a91ef4b33392c27a28951d5de2c57/addons/stock/models/stock_location.py#L328-L330

This is ok because we know the call to `_get_putaway_strategy` is made on a single record (`sml.move_id.location_dest_id`) so the default will correctly be a child of said location.

-----
Ticket:
opw-5359945

### DIFF
--- a/addons/stock/tests/test_stock_flow.py
+++ b/addons/stock/tests/test_stock_flow.py
@@ -4,7 +4,7 @@ from odoo.addons.stock.tests.common import TestStockCommon
 from odoo.exceptions import ValidationError
 from odoo.tests import Form, tagged
 from odoo.tools import mute_logger, float_round
-from odoo import fields
+from odoo import Command, fields
 
 
 class TestStockFlow(TestStockCommon):
@@ -2598,6 +2598,58 @@ class TestStockFlow(TestStockCommon):
         backorder_wizard.process()
         bo = self.env['stock.picking'].search([('backorder_id', '=', picking.id)])
         self.assertEqual(bo.state, 'assigned')
+
+    def test_multiple_moves_with_different_destinations_putaway_strategy(self):
+        '''
+        Ensure that, when assigning a batch of moves with different destinations,
+        putaway strategy correctly defaults to child locations.
+        '''
+        view_a, view_b = self.env['stock.location'].create([{
+            'name': 'View A',
+            'usage': 'view',
+            'location_id': self.stock_location,
+        }, {
+            'name': 'View B',
+            'usage': 'view',
+            'location_id': self.stock_location,
+        }])
+        child_a, child_b = self.env['stock.location'].create([{
+            'name': 'Child A',
+            'usage': 'internal',
+            'location_id': view_a.id,
+        }, {
+            'name': 'Child B',
+            'usage': 'internal',
+            'location_id': view_b.id,
+        }])
+        picking_1, picking_2 = self.env['stock.picking'].create([{
+            'location_id': self.supplier_location,
+            'location_dest_id': view_a.id,
+            'picking_type_id': self.picking_type_in,
+            'move_ids': [Command.create({
+                'name': 'SML should end up in Child A',
+                'location_id': self.supplier_location,
+                'location_dest_id': view_a.id,
+                'product_id': self.productA.id,
+                'product_uom_qty': 1.0,
+            })],
+        }, {
+            'location_id': self.supplier_location,
+            'location_dest_id': view_b.id,
+            'picking_type_id': self.picking_type_in,
+            'state': 'draft',
+            'move_ids': [Command.create({
+                'name': 'SML should end up in Child B',
+                'location_id': self.supplier_location,
+                'location_dest_id': view_b.id,
+                'product_id': self.productB.id,
+                'product_uom_qty': 1.0,
+            })],
+        }])
+        (picking_1 | picking_2).action_confirm()
+        self.assertEqual(picking_1.move_ids.move_line_ids.location_dest_id, child_a)
+        self.assertEqual(picking_2.move_ids.move_line_ids.location_dest_id, child_b)
+
 
 @tagged('-at_install', 'post_install')
 class TestStockFlowPostInstall(TestStockCommon):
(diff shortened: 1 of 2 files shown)


addons/stock/tests/test_stock_flow.py
```diff
@@ -2599,6 +2599,58 @@ def test_picking_mixed_tracking_with_backorder(self):
...
 
+    def test_multiple_moves_with_different_destinations_putaway_strategy(self):
+        '''
+        Ensure that, when assigning a batch of moves with different destinations,
+        putaway strategy correctly defaults to child locations.
+        '''
+        view_a, view_b = self.env['stock.location'].create([{
+            'name': 'View A',
+            'usage': 'view',
+            'location_id': self.warehouse_1.lot_stock_id.id,
```
clesgow:
```suggestion
            'location_id': self.stock_location.id,
```
Nitpick: Use what is defined in the commons, makes things more readable 🙂 
Same thing for the supplier location, picking type, ...

addons/stock/tests/test_stock_flow.py
```diff
@@ -2599,6 +2599,58 @@ def test_picking_mixed_tracking_with_backorder(self):
...
+        child_a, child_b = self.env['stock.location'].create([{
+            'name': 'Child A',
+            'usage': 'internal',
+            'location_id': view_a.id,
+        }, {
+            'name': 'Child B',
+            'usage': 'internal',
+            'location_id': view_b.id,
+        }])
+        batch = self.env['stock.picking'].create([{
```
clesgow:
```suggestion
        picking1, picking2 = self.env['stock.picking'].create([{
```
Nitpick: `batch` has a specific meaning in stock modules, and since we're not talking about `stock.picking.batch` here, better avoid it.

addons/stock/tests/test_stock_flow.py
```diff
@@ -2599,6 +2599,58 @@ def test_picking_mixed_tracking_with_backorder(self):
...
+            'location_id': self.partner.property_stock_supplier.id,
+            'location_dest_id': view_a.id,
+            'picking_type_id': self.warehouse_1.in_type_id.id,
+            'state': 'draft',
```
clesgow:
```suggestion
```
Nitpick: No need to explicitly set default values.

addons/stock/tests/test_stock_flow.py
```diff
@@ -2599,6 +2599,58 @@ def test_picking_mixed_tracking_with_backorder(self):
...
+            'move_ids': [Command.create({
+                'name': 'SML should end up in Child B',
+                'location_id': self.partner.property_stock_supplier.id,
+                'location_dest_id': view_b.id,
+                'product_id': self.productB.id,
+                'product_uom_qty': 1.0,
+            })],
+        }])
+        batch.action_confirm()
+        self.assertEqual(batch.move_ids.move_line_ids.location_dest_id.ids, (child_a | child_b).ids)
```
clesgow: Imho, you'd be better do the following check:
```suggestion
        self.assertEqual(picking1.move_line_ids.location_dest_id, child_a)
        self.assertEqual(picking2.move_line_ids.location_dest_id, child_b)
```
So we make sure that what was expected for each move is correct

addons/stock/models/stock_location.py
```diff
@@ -368,7 +368,7 @@ def _get_putaway_strategy(self, product, quantity=0, package=None, packaging=Non
             putaway_location = putaway_rules._get_putaway_location(product, quantity, package, packaging, qty_by_location)
 
         if not putaway_location:
-            putaway_location = locations[0] if locations and self.usage == 'view' else self
+            putaway_location = locations.filtered(lambda l: l._child_of(self))[0] if locations and self.usage == 'view' else self
```
clesgow: Imho, this isn't the right fix. The issue stems this call:
https://github.com/odoo/odoo/blob/26dbbdaf460a91ef4b33392c27a28951d5de2c57/addons/stock/models/stock_move_line.py#L280-L282

In this case, we have several move lines that are grouped up since they don't have any `result_package_id`, and have their destination locations taken as a whole which doesn't make sense. It's a side-effect that I didn't notice in #210142, but I think for that call you should just remove the `locations` context.

This passes your test as well.

addons/stock/tests/test_stock_flow.py
vandroogenbd: Fair enough, I missed that the information was already there in the form of ids while I looked for records :sweat_smile:

addons/stock/models/stock_location.py
vandroogenbd: You're right this is better, I was fixing the symptom not the cause :sweat_smile:

PR #247404, opened 2026-02-05
[FIX] web: retrigger clicks after interactions started

robinlej:
[IMP] web, website: retrigger clicks after interactions started

Commit [1] made sure events (especially clicks) done on buttons (or
button-like elements) before the lazy-loaded JS had time to load would
be caught, delayed and triggered again after all the code is loaded.
Indeed, as long as it is not the case, a lot of events won't have any
effect.

With the introduction of Interactions in [2], that behavior stopped
working (in the context of interactions):
- the old system (publicWidgets) attached events on the DOM before the
publicWidget had started and retriggered the events after awaiting the
publicWidget's start (which was asynchronous).
- the new system (interactions) only attaches the events at the
interaction's start, and the `start` method is now synchronous. Both of
these aspects were done on purpose to avoid other pitfalls. The
lazyloader would also need to be notified that the interactions have
been loaded by pushing a promise into `retriggeringWaitingProms`.
It means that the `lazyloader` mechanism retriggers the events before
they have been attached on the DOM, meaning they are never played.

With this commit, we make sure to wait for the interaction service to
start (just after Colibri setup) before retriggering events.
In the service, click events are buffered (max one click per element,
captured) until all interactions that may have an impact on it are
started.
Interactions are considered to have a potential impact if either the
element is selected by one of the dynamicContent keys (including dynamic
selectors), or the element is a descendant of the interaction's root
element.
After these interactions have started, buffered clicks are retriggered
once more (and only once per element).

⋅ Why do we only wait for impacted interactions instead of waiting for all
interactions?

Some interactions may take much more time than the rest of the page, but
have absolutely no impact around them. That is for instance the case of
dynamic snippets, in website. Not waiting for them allows the page to be
a bit more reactive.
It is still possible (but discouraged) to have an interaction impacting
an element outside of its scope. It should be avoided and, otherwise,
rare enough.

⋅ Why do we wait for impacted interactions to be started instead of
calling the handlers one by one?

Waiting for the last one let us be sure that the DOM is in the expected
state: all `willStart` have run and cannot remove or add a class, the
dynamicContent of all impactful interactions has been evaluated once and
potential classes / styles updates have been processed.
Imagine a button that appears in the dynamicContent of 2 different
interactions: the 1st one handles a click event, but the 2nd one sets
the button to `disabled` by default. The right behavior is that the
click shouldn't do anything, as the button should be disabled. If we
called the handlers, the 1st interaction would run its handler no matter
what. This pitfall is avoided by triggering an actual click event.

[1]: https://github.com/odoo/odoo/commit/0f9388cee6120f23ba4636501cae4cd9eb148461
[2]: https://github.com/odoo/odoo/commit/dd13994674d4ef4683f5a4d46a1f604650cfb92b

task-5221179

### DIFF
--- a/addons/web/static/tests/public/interaction.test.js
+++ b/addons/web/static/tests/public/interaction.test.js
@@ -461,7 +461,7 @@ describe("adding listeners", () => {
             await click("button");
             await animationFrame();
             expect("button").toHaveClass("pe-none");
-            expect("button span").toHaveCount(1);
+            expect("button > span").toHaveCount(1);
             def.resolve();
             expect.verifySteps([]);
             await animationFrame();
@@ -2827,7 +2827,13 @@ describe("locked", () => {
 
             const observer = new MutationObserver((mutations) => {
                 for (const m of mutations) {
-                    if ([...m.addedNodes].some((node) => node.tagName === "SPAN")) {
+                    if (
+                        [...m.addedNodes].some(
+                            (node) =>
+                                node.tagName === "SPAN" &&
+                                node.classList.contains("o_btn_loading_spinner")
+                        )
+                    ) {
                         expect.step("loading added");
                     }
                 }
@@ -2836,12 +2842,12 @@ describe("locked", () => {
         });
 
         test("should be added when the handler takes more than 400ms", async () => {
-            expect("span.oi-spin").toHaveCount(0);
+            expect(".o_btn_loading_spinner").toHaveCount(0);
             await click("button");
             // Advance time more than the debounce delay of makeButtonHandler
             // (400ms) but less than the handler duration.
             await advanceTime(500);
-            expect("span.oi-spin").toHaveCount(1);
+            expect(".o_btn_loading_spinner").toHaveCount(1);
             expect.verifySteps(["loading added"]);
             await advanceTime(handlerDuration);
             expect.verifySteps(["handler done"]);
@@ -2849,14 +2855,14 @@ describe("locked", () => {
 
         test("should never be added when the handler takes less than 400ms", async () => {
             handlerDuration = 100;
-            expect("span.oi-spin").toHaveCount(0);
+            expect(".o_btn_loading_spinner").toHaveCount(0);
             await click("button");
             // Advance time more than the handler duration but less than the
             // debounce delay of makeButtonHandler (400ms).
             await advanceTime(200);
             expect.verifySteps(["handler done"]);
             await advanceTime(1000);
-            expect("span.oi-spin").toHaveCount(0);
+            expect(".o_btn_loading_spinner").toHaveCount(0);
             expect.verifySteps([], {
                 message:
                     "Loading effect should never be added in the DOM for handlers shorter than 400ms",
--- a/addons/web/static/src/core/utils/ui.js
+++ b/addons/web/static/src/core/utils/ui.js
@@ -229,18 +229,33 @@ export function getPreviousTabableElement(container = document.body) {
  *         initial state
  */
 export function addLoadingEffect(btnEl) {
-    // Note that pe-none is used alongside "disabled" so that the behavior is
-    // the same on links not using the "btn" class -> pointer-events disabled.
-    btnEl.classList.add("o_btn_loading", "disabled", "pe-none");
-    btnEl.disabled = true;
-    const loaderEl = document.createElement("span");
-    loaderEl.classList.add("oi", "oi-spin", "me-2");
-    loaderEl.dataset.icon = "autorenew";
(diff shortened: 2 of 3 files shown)


addons/web/static/tests/public/interaction.test.js
```diff
@@ -209,6 +209,34 @@ describe("adding listeners", () => {
         expect(clicked).toBe(1);
         expect("span").toHaveAttribute("x", "1");
     });
+
+    test("events triggered before interaction is ready are buffered and replayed", async () => {
```
robinlej: NB: I don't think it's possible to test the lazyloader, so this is a minimal test at least for colibri

addons/web/static/src/public/colibri.js
```diff
@@ -1,3 +1,5 @@
+import { BUTTON_HANDLER_SELECTOR, makeButtonHandler } from "./utils";
```
robinlej: Not sure about importing things into Colibri. From a design perspective, maybe we want to keep this file agnostic of other files?

addons/web/static/src/public/colibri.js
```diff
@@ -40,21 +43,47 @@ export class Colibri {
...
 
     async start() {
-        await this.interaction.willStart();
+        const content = this.interaction.dynamicContent;
+        let willStartProm;
+        const bufferHandler = makeButtonHandler((ev) => {
+            const target = ev.target;
+            target._interactionsToReplay = ++target._interactionsToReplay || 1;
+            this.bufferedClicks.push({ ev, target });
+            return willStartProm;
```
robinlej: I return the promise so that `makeButtonHandler` properly adds the loading effect

addons/web/static/src/public/colibri.js
```diff
@@ -40,21 +43,47 @@ export class Colibri {
...
+            }
+        }
         this.hasStarted = true;
     }
 
     async start() {
-        await this.interaction.willStart();
+        const content = this.interaction.dynamicContent;
+        let willStartProm;
+        const bufferHandler = makeButtonHandler((ev) => {
```
robinlej: One issue with this approach: if for some reason the setup of the interaction service already took a long time, the 400ms timer starts again from 0.

What I mean is:
- Page takes a long time to load
- User clicks on a button
- The lazyloader buffers the click, and after 400ms will start showing the loading spinner
- The interaction is set up => the lazyloader retriggers the click, the spinner is removed
- The interaction buffers the new click
  - either `willStart` takes > 400ms and the spinner reappears then,
  - or it takes < 400ms and the spinner doesn't reappear (but the click is finally triggered after `willStart` anyway)
  => in both cases, the UX is weird (click, wait, loader, no loader but still no action, wait, (new loader), actual action happens)

It should be limited though, as I don't expect the interaction service to take that much time to start... except on bad connections. 😕

robinlej: Note: I tried keeping track of the click timestamp across the lazyloader, utils and colibri so that the 400 ms are not fixed but dependent on the delta between `Date.now()` and the tracked timestamp.
It works, but there's still a flash between the moment the lazyloader removes the loader, and the moment the click is retriggered by colibri (and the loader reappears).

robinlej:
A note about the diff:
image


338 of those lines are for tests, this PR is really not as big as it seems

addons/web/static/src/public/lazyloader.js
```diff
@@ -171,4 +175,5 @@ export default {
     loadScripts: _loadScripts,
     allScriptsLoaded: _allScriptsLoaded,
     registerPageReadinessDelay: retriggeringWaitingProms.push.bind(retriggeringWaitingProms),
+    bufferedClicks,
```
FrancoisGe: It s not imported in another file ?

addons/web/static/src/public/interaction_service.js
```diff
@@ -211,6 +231,110 @@ class InteractionService {
         const proms = this.proms.slice();
         return Promise.all(proms);
     }
+
+    /**************************** BUFFERING CLICKS ****************************/
+    /**
+     * Buffers clicks made before interactions are started to be replayed after.
+     *
+     * @param {object} interaction - Colibri instance
```
FrancoisGe: object => Object

FrancoisGe:
@robodoo r+

FrancoisGe:
@robodoo rebase-ff

FrancoisGe:
@robodoo r+

seb-odoo:
Hello, HOOT tests can no longer run without `website` installed since this PR, could you have a look?

https://github.com/odoo/odoo/pull/281411

PR #247407, opened 2026-02-05
[FIX] stock_account: fix AVCO valuation with backdated stock moves

MaximeNoirhomme229:
**Issue**:
If the date of some stock moves is anterior to the creation date of the product
in the database, the associated valuation is replaced by the initial standard
price of the product.

**Steps to reproduce**:
- Create a new product with a standard price of 0 and AVCO cost method
- Create a PO for that product with a unit cost of 1,000,000, confirm it and
  validate the receipt
- Go to Accounting > Review > Inventory > Inventory Valuation
  -> Observe that the valuation correctly takes the purchase into account
- Go back to the receipt, unlock it and change the effective date to one week
  in the past
- Go back to Inventory Valuation
  -> Observe that the valuation no longer takes the purchase into account
- Change the valuation date to yesterday
  -> Observe that the valuation takes it into account again

**Cause**:
When a product is created, a `product.value` record is instantiated with
today’s date:
https://github.com/odoo/odoo/blob/bbaf38aa99143be4679cf951c5c0f1a1c8ecf716/addons/stock_account/models/product.py#L174
https://github.com/odoo/odoo/blob/bbaf38aa99143be4679cf951c5c0f1a1c8ecf716/addons/stock_account/models/product.py#L202

In the AVCO computation, a manually set product value (`product.value`) takes
precedence over move values when it is anterior, either here:
https://github.com/odoo/odoo/blob/bbaf38aa99143be4679cf951c5c0f1a1c8ecf716/addons/stock_account/models/product.py#L309-L312
or here:
https://github.com/odoo/odoo/blob/bbaf38aa99143be4679cf951c5c0f1a1c8ecf716/addons/stock_account/models/product.py#L334-L338

Since the stock move date is set one week in the past, the initial product value
(0.0) takes precedence over the move valuation. When the valuation date is moved
forward to yesterday, this initial product value is ignored and the move value
is correctly applied again.

**Solution**:
Setting the initial `product.value` date to the product creation date is
arbitrary, as it makes inventory valuation depend on when the product was
encoded rather than on real stock history. Instead, set the date of the first
`product.value` to the earliest possible epoch, ensuring that any real stock
move always takes precedence in AVCO valuation.

opw-5882080

### DIFF
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -1,6 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 from bisect import bisect
 from collections import defaultdict
+from datetime import datetime
 
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
@@ -216,7 +217,7 @@ class ProductProduct(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         products = super().create(vals_list)
-        products._change_standard_price({product: 0 for product in products if product.standard_price})
+        products.with_context(valuation_date=datetime.min)._change_standard_price({product: 0 for product in products if product.standard_price})
         return products
 
     def write(self, vals):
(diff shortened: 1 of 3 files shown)


amoyaux:
You could also remove this line (NTH, if it's not working I will check in a second time) https://github.com/odoo/odoo/blob/19.0/addons/stock_account/tests/common.py#L415

amoyaux:
robodoo r+

PR #247438, opened 2026-02-05
[IMP] website, *: js tracker

h4818:
*: website_blog, website_event, website_event_track, website_forum, website_hr_recruitment, website_livechat, website_profile, website_sale, website_slides

There are currently some issues with the existing tracker (extension of `ir.http`'s `_post_dispatch`)
1. Writing a tracking event to the DB prevents the readonly capability of odoo.
2. Number of bots crawling odoo pages consume resources.

By using a JS based tracker, we can
1. improve the performance of browsing website pages as we don't need a write to the DB
2. enable readonly routes on the website - allowing for spreading the db load.
3. potentially reduce the number of tracking events caused by bots (which don't normally execute js)

Additionally, tracking is activated by default on all website templates/views.

Performance Analysis

Using a sample of 115 loads of a blank website home page, we seem to save approximately 10ms on page load time.

| Description       | _register_website_track (old) | js track (this pr) |
|-------------------|-------------------------------|--------------------|
| Average load time*|    0.0197652 s                |    0.0093054 s     |
| Tracking request  |    0 s                        |    0.0154167 s **  |

&ast; This average includes the response time for the first uncached response of approx 500 ms
** Due to the js loading delay, the tracking request only executed on approximately 10% of the page loads.

Task-4939052

### DIFF
--- a/addons/website/controllers/main.py
+++ b/addons/website/controllers/main.py
@@ -241,7 +241,7 @@ class Website(Home):
             params = dict(kwargs, res_model=res_model, res_id=res_id, url=url)
             extra_tracking_vals = request.env[res_model].browse(res_id).sudo()._get_extra_tracking_values(**params)
 
-        request.env['website.visitor']._get_visitor_from_request(
+        request.env['ir.http']._get_visitor_from_request(
             force_create=True,
             force_track_values={
                 'url': url,
--- a/addons/website/controllers/main.py
+++ b/addons/website/controllers/main.py
@@ -226,6 +226,31 @@ class Website(Home):
         return super().web_login(*args, **kw)
 
     # ------------------------------------------------------
+    # Tracking
+    # ------------------------------------------------------
+
+    @http.route('/website/odoo_track', type='jsonrpc', auth='public', methods=['POST'], website=True, csrf=True)
+    def track(self, res_model, res_id, **kwargs):
+        url = request.httprequest.referrer
+        extra_tracking_vals = {}
+        if (
+            res_model and res_id
+            and res_model in request.env
+            and hasattr(request.env[res_model], '_get_extra_tracking_values')
+        ):
+            params = dict(kwargs, res_model=res_model, res_id=res_id, url=url)
+            extra_tracking_vals = request.env[res_model].browse(res_id).sudo()._get_extra_tracking_values(**params)
+
+        request.env['website.visitor']._get_visitor_from_request(
+            force_create=True,
+            force_track_values={
+                'url': url,
+                **extra_tracking_vals,
+            },
+        )
+        return {}
+
+    # ------------------------------------------------------
     # Business
     # ------------------------------------------------------
 
--- a/addons/website/models/ir_http.py
+++ b/addons/website/models/ir_http.py
@@ -195,27 +195,6 @@ class IrHttp(models.AbstractModel):
             super()._auth_method_public()
 
     @classmethod
-    def _register_website_track(cls, response):
-        if request.env['ir.http'].is_a_bot():
-            return False
-        if getattr(response, 'status_code', 0) != 200 or request.httprequest.headers.get('X-Disable-Tracking') == '1':
-            return False
-        template = False
-        if hasattr(response, '_cached_page'):
-            website_page, template = response._cached_page, response._cached_view_id
-        elif hasattr(response, 'qcontext'):  # classic response
-            main_object = response.qcontext.get('main_object')
-            website_page = getattr(main_object, '_name', False) == 'website.page' and main_object
-            template = response.qcontext.get('response_template')
-            if isinstance(template, str) and '.' not in template:
-                template = 'website.%s' % template
-
-        if template and not request.env.cr.readonly and request.env['ir.ui.view']._get_cached_template_info(template)['track']:
-            request.env['website.visitor']._handle_webpage_dispatch(website_page)
-
-        return False
-
-    @classmethod
     def _match(cls, path):
         if not hasattr(request, 'website_routing'):
             website = request.env['website'].with_context(lang=None).get_current_website()
@@ -282,7 +261,6 @@ class IrHttp(models.AbstractModel):
     @classmethod
     def _post_dispatch(cls, response):
         super()._post_dispatch(response)
-        cls._register_website_track(response)
 
     @api.model
     def get_nearest_lang(self, lang_code):
--- a/addons/website/models/ir_http.py
+++ b/addons/website/models/ir_http.py
@@ -434,6 +434,45 @@ class IrHttp(models.AbstractModel):
         # is not restricted by the website module.
         return result
 
+    def _get_visitor_from_request(self, force_create=False, force_track_values=None):
+        """ Return the visitor as sudo from the request.
+
+        :param force_create: force a visitor creation if no visitor exists
+        :param force_track_values: an optional dict to create a track at the
+            same time.
+        :return: the website visitor if exists or forced, empty recordset
+            otherwise.
+        """
+
+        # This function can be called in json with mobile app.
+        # In case of mobile app, no uid is set on the jsonRequest env.
+        # In case of multi db, _env is None on request, and request.env unbound.
+        if not (request and request.env and request.env.uid):
+            return None
+
+        access_token = self.env['website.visitor']._get_access_token()
+
+        if force_create:
+            force_track_values = force_track_values or {}
+            visitor_id, _ = self.env['website.visitor']._upsert_visitor(
+                token_or_partner_id=access_token,
+                website_id=request.website.id,
+                lang_id=request.lang.id,
+                country_code=request.geoip.country_code,  # GEOIP might return a country code unknown to odoo
+                timezone=self.env['website.visitor']._get_visitor_timezone(),
+                **force_track_values
+            )
+            return self.env['website.visitor'].sudo().browse(visitor_id)
+
+        visitor = self.env['website.visitor'].sudo().search_fetch([('access_token', '=', access_token)])
+
+        if not force_create and not self.env.cr.readonly and visitor and not visitor.timezone:
+            tz = self.env['website.visitor']._get_visitor_timezone()
+            if tz:
+                visitor._update_visitor_timezone(tz)
+
+        return visitor
+
 
 class ModelConverter(ir_http.ModelConverter):
 
--- a/addons/website/models/website.py
+++ b/addons/website/models/website.py
@@ -1269,6 +1269,11 @@ class Website(models.CachedModel):
             page_temp = page_url + (inc and "-%s" % inc or "")
         return page_temp
 
+    def _is_tracking_enabled(self, main_object):
+        if main_object._name in ['ir.ui.view', 'website.page']:
+            return main_object.track
+        return True
+
     def _get_plausible_script_url(self):
         return self.env['ir.config_parameter'].sudo().get_str(
             'website.plausible_script') or 'https://plausible.io/js/plausible.js'
--- a/addons/website/models/website_visitor.py
+++ b/addons/website/models/website_visitor.py
@@ -16,7 +16,7 @@ from odoo.http import request
 class WebsiteTrack(models.Model):
     _name = 'website.track'
     _description = 'Visited Page'
-    _order = 'visit_datetime DESC'
+    _order = 'visit_datetime DESC, id DESC'
     _log_access = False
 
     visitor_id = fields.Many2one('website.visitor', ondelete="cascade", index=True, required=True, readonly=True)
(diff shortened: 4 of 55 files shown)


addons/website/controllers/main.py
```diff
@@ -221,6 +221,49 @@ def _login_redirect(self, uid, redirect=None):
...
+        request.lang = request.env['res.lang']._get_data(code=(context_lang))
+        # Improvements:
+        # - Instead of searching for the page based on URL, we could send that data in the httprequest - done
+        # - prevent arbitrary calls on this route (bot spam)
+        # - Clean up/ split the _get_visitor_from_request method (see lang, website_id, etc)
+        # - Handle other types of events? eg. time on page (plausible tracks: Appointment Confirmations, Lead Generated, 404, Order Confirmed - but these are records, except 404) 
+        # - Time: Since the event is sent 2 seconds after the script is loaded, the time of the event is not correct
+        # - use an interaction and rpc to avoid the manual beacon/ fetch call - to manage the lang and website etc. - consider including it in assets_frontend_minimal
+        #   also might provide the ability to prevent the call if the component/page is destroyed before the tracking is sent.
+        required_keys = {"url", "path", "website_id", "res_model", "res_id"}
```
Gorash: define this arguments in the controller method

addons/website/controllers/main.py
```diff
@@ -221,6 +221,49 @@ def _login_redirect(self, uid, redirect=None):
...
+            # and page.track
+        ):
+            website_track_values = {'url': data['url']}
+            website_track_values['website_id'] = int(data['website_id'])
+            res_id = int(data['res_id'])
+            if data['res_model'] == 'website.page':
+                if not data['track']:
+                    return resp
+                website_track_values['page_id'] = res_id
+            request.env['website.visitor']._get_visitor_from_request(force_create=True, force_track_values=website_track_values)
```
Gorash: Change the api of this method. It is divided into methods, one of which should not retrieve information from the request but receives it as a parameter. The request should not be modified.

addons/website/views/website_templates.xml
```diff
@@ -272,6 +272,66 @@
...
+        <script id="odoo_tracking">
+            /**
+            * POC Odoo Analytics Tracker
+            * This is a crude proof of concept tracker for odoo.
+            * It could/should be moved to a separate file, maybe a standard interaction.
+            * To explore pros and cons - whether the delay in loading is acceptable.
+            */
+            (function() {
+                // TODO: try to filter bots based on webdriver, _phantom, __nightmare, Cypress etc. or see ["ir.http"].is_a_bot
+
```
Gorash:
```xml
if <t t-out="0 if your_check_method_for_tracking(main_object) else 1"/>:
    return;
```

addons/website/views/website_templates.xml
```diff
@@ -272,6 +272,66 @@
         <t t-if="website.plausible_shared_key and not editable">
             <script id="plausible_script" name="plausible" defer="defer" t-att-data-domain="website.plausible_site" t-att-src="website._get_plausible_script_url()"></script>
         </t>
+        <script id="odoo_tracking">
```
Gorash: idea ?
```xml
<script if="your_check_method_for_tracking(main_object)"
    defer="defer"
    t-attf-src="/website/track?website_id={{website.id}}"></script>
```
or
```xml
<t t-out="script_to_log(main_object)"/>  => this method return the script tag
```

h4818: I like these ideas
but, getting the script with `t-attf-src` is not ideal as it means an additional request

so I went with the second idea - it's cool because we can change it in python without a module update, but python returning js code is a little weird. not sure we really need it.

addons/website/views/website_templates.xml
h4818: this is now checked in the python function `_get_odoo_tracking_script`

addons/website_forum/controllers/website_forum.py
```diff
@@ -63,6 +63,18 @@ def _prepare_mark_as_offensive_values(self, post, **kwargs):
         })
         return values
 
+    # Tracking
```
h4818: Add this to a separate py controller file that extends the website controller instead

addons/website_profile/controllers/main.py
```diff
@@ -16,10 +16,12 @@
 from odoo.http import request
 from odoo.tools.translate import LazyTranslate
 
+from odoo.addons.website.controllers.main import Website
+
 _lt = LazyTranslate(__name__)
 
 
-class WebsiteProfile(http.Controller):
+class WebsiteProfile(Website):
```
h4818: Unnecessary shortcut

addons/website/models/website_visitor.py
```diff
@@ -238,24 +237,33 @@ def _upsert_visitor(self, access_token, force_track_values=None):
...
+                vals_extra=vals_extra,
             )
 
         [result] = self.env.execute_query(query)
         return result
 
+    def _get_additional_track_query_parts(self, **kwargs):
+        page_id = kwargs.get('page_id')
+        if page_id:
+            return SQL(", %(page_id)s AS page_id", page_id=page_id), SQL(", page_id"), SQL(", page_id::integer")
```
h4818: 🤮

addons/website/models/website_visitor.py
```diff
@@ -238,24 +237,33 @@ def _upsert_visitor(self, access_token, force_track_values=None):
...
+        return SQL(), SQL(), SQL()
+
     def _get_visitor_from_request(self, force_create=False, force_track_values=None):
```
h4818: move this method to a controller - models should not use the request object

addons/website/models/website.py
```diff
@@ -1297,6 +1299,45 @@ def get_unique_path(self, page_url):
...
         return page_temp
 
+    def _get_odoo_tracking_script(self, main_object):
+        if (
+            main_object._name in {'ir.ui.view', 'website.page'}
+            and not main_object.track
+            # or self.env['ir.http'].is_a_bot  # removed as this uses the request lib
+        ):
+            return ""
+        return markupsafe.Markup(f"""
```
h4818: if possible, put this into a static js file / interaction
(consider the main_object model and id and also other params like url, product_id etc )

addons/website/models/website_visitor.py
```diff
@@ -190,32 +190,31 @@ def action_send_mail(self):
...
         related visitor if exists, through a single raw SQL UPSERT Query.
 
         It will also create a tracking record if requested, in the same query.
 
-        :param access_token: token to be used to upsert the visitor
-        :param force_track_values: an optional dict to create a track at the
-            same time.
+        :param token_or_partner_id: token (or partner id) to be used to upsert the visitor
+        :param
+        :param url: optional url to create a track record at the same time
```
h4818, line 204: update docstring

addons/website_livechat/tests/test_website_visitor.py
```diff
@@ -67,7 +67,14 @@ def test_visitor_id_continuity_across_sessions(self):
         self.env["mail.presence"]._update_presence(operator)
 
         # Anonymous user
-        self.url_open(self.tracked_page.url)  # visitor created
+        self.make_jsonrpc_request(
```
h4818, line 70: what if the user only visits untracked pages before starting a session?
maybe website_livechat need to force the visitor creation when starting a chat session?

addons/website_sale/controllers/main.py
```diff
@@ -1940,6 +1928,8 @@ def order_2_return_dict(self, order):
     # --------------------------------------------------------------------------
     @route('/shop/products/recently_viewed_update', type='jsonrpc', auth='public', website=True)
     def products_recently_viewed_update(self, product_id, **kwargs):
+        # Consider using the standard tracking route with the product_id field
+        # but there shouldn't (?) be a URL
```
h4818: do it - as an example

addons/website/models/website.py
h4818: Done (in 2nd commit - to be squashed with 1st)
note: The url should always be captured, it triggers the insert of `website.track` record.

addons/website/models/website_visitor.py
h4818: Done in last commit

addons/website/models/website_visitor.py
h4818: cleaned up somewhat...
the data is injected only into track query and not the visitor query
this means one of SQL statements can be removed.

Now we iterate all kwargs (extra values for the website.track record) and prepare the sql for all fields in the website.track model.
So, the extension in website_sale can be removed, as the the product_id column and value will be handled in the base code that looks at `_fields`

note: lost the cast `::integer`

addons/payment/views/portal_templates.xml
```diff
@@ -104,7 +104,7 @@
     </template>
 
     <!-- Display of /payment/status -->
-    <template id="payment.payment_status" name="Payment Status">
+    <template id="payment.payment_status" name="Payment Status" track="1">
```
AntoineVDV: Why is this needed here? 🤔

h4818: I have since removed the diff, but the tracking is still enabled.
but to answer your question anyway, we want to track all website pages now. It should paint a complete picture of all urls visited so  it even includes visits to non-existent pages
Is there a reason you wouldn't want tracking there?

AntoineVDV: No reason in particular, but I found this page so irrelevant to tracking that it made me wonder :smile:

addons/website/controllers/main.py
```diff
@@ -225,6 +225,26 @@ def _login_redirect(self, uid, redirect=None):
...
+    def track(self, res_model, res_id, url, **kwargs):
+        extra_tracking_vals = {}
+        if res_model and res_id:
+            params = dict(kwargs, res_model=res_model, res_id=res_id, url=url)
+            extra_tracking_vals = request.env[res_model].browse(res_id).sudo()._get_extra_tracking_values(**params)
+
+        request.env['ir.http']._get_visitor_from_request(
+            force_create=True,
+            force_track_values={
+                'url': url,
```
Gorash, line 243: Wouldn't it be interesting to look at the referrer of request.httprequest?

h4818: It works, and seems fine, but I think it gives us less control.
Previously we could use this method to create the visitor record only (no tracking) by not providing a url,
now most calls will automatically create the track record.

It's ok though. we can simply use `_get_visitor_from_request` to create a visitor without tracking.

addons/website/views/website_templates.xml
h4818: So eventually, I use an interaction, so we can change the js script without a module update.
I simply include the `tracking-enabled` attribute in the `<html>` element where the main object is specified.

h4818:
@odoo/rd-security just moving a method from `website.visitor` to `ir.http` - I'm also not sure why a domain would have the same issue.

addons/website_sale/models/product_product.py
```diff
@@ -229,3 +229,12 @@ def _get_image_1024_url(self):
         """
         self.ensure_one()
         return self.env["website"].image_url(self, "image_1024")
+
+    def _get_extra_tracking_values(self, **kwargs):
+        etv = {}
```
livaios, line 234: Hello @h4818,
Very nice feature overall, code is much cleaner 😉 
Just one small comment, can you be more descriptive in the name?
Took me a sec to understand what it was

```suggestion
        extra_tracking_values = {}
```

addons/website/models/ir_http.py
```diff
@@ -456,6 +434,45 @@ def _is_allowed_cookie(cls, cookie_type):
...
+                token_or_partner_id=access_token,
+                website_id=request.website.id,
+                lang_id=request.lang.id,
+                country_code=request.geoip.country_code,  # GEOIP might return a country code unknown to odoo
+                timezone=self.env['website.visitor']._get_visitor_timezone(),
+                **force_track_values
+            )
+            return self.env['website.visitor'].sudo().browse(visitor_id)
+
+        visitor = self.env['website.visitor'].sudo().search_fetch([('access_token', '=', access_token)])
```
antonrom1, line 467: I am not sure it's a good idea to search `access_token` like this, it's in a btree index, thus potentially vulnerable to timing attack. IMO you should have 2 identifiers: a public and a private one. e.g. visitor_id & access_token.

You would search on the public one (`search_fetch([('visitor_id', '=', visitor_id)])`), and then compare the `access_token` with timing-resistant comparison

I didn't really check what would be the implications of leaking this `access_token` though

addons/website/controllers/main.py
```diff
@@ -225,6 +225,27 @@ def _login_redirect(self, uid, redirect=None):
...
+    # Tracking
+    # ------------------------------------------------------
+
+    @http.route('/website/odoo_track', type='jsonrpc', auth='public', methods=['POST'], website=True, csrf=True)
+    def track(self, res_model, res_id, **kwargs):
+        url = request.httprequest.referrer
+        extra_tracking_vals = {}
+        if res_model and res_id:
+            params = dict(kwargs, res_model=res_model, res_id=res_id, url=url)
+            extra_tracking_vals = request.env[res_model].browse(res_id).sudo()._get_extra_tracking_values(**params)
```
antonrom1, line 242: Isn't that a new way of checking if a model is installed on a db or not on a public controller? I am not sure we want to expose this information

antonrom1, line 242: Also you add no-op `_get_extra_tracking_values` on product.product just so the generic route doesn't raise an attribute error.
IMO that smells a little bit, it's easy to add tracking to a new model and miss adding this method, no? Maybe a small mixin that you would need to inherit from to be able to track?

h4818: What do you mean by expose? This controller doesn't return anything
it simply collects additional values for the `website.track` record
`page_id` if you're visiting a `website.page`
`product_id` if you're visiting a `product.product`
and soon many others.

antonrom1: you bubble up KeyError on `request.env[res_model]`, so you basically can deduce if e.g. `'account.move'` is installed by simply observing if you got a key error (means not installed) or an attribute error (means installed)

addons/website/models/ir_http.py
h4818: @antonrom1
1. this method is simply moved from `website.visitor`
2. This is not always an access token - when a user is logged in, it is actually a partner id (int) - I can change the name if you prefer, like I did on `_upsert_visitor`
3. I'm not sure the information in `website.visitor` or `website.track` is something an attacker is going to target - at the very most they would get which products were recently viewed :upside_down_face: 
4. I'm not sure we should radically change the website.visitor model for a potential timing attack on a web application, wouldn't network latency significantly outweigh the time for indexing?

Should we really build a vault door for data that we delete in a cron job?

addons/website/controllers/main.py
h4818: Indeed, you're right.
I have added the no op functions to the published mixin - most published models will inherit from there - I didn't think it was worth it to add yet another mixin. And we can't guarantee the use of the mixin anyway.
(eg there are a few exceptions like `product` and `user` which don't use the published mixin)

(I think you were referring to the no-op on published mixin and other models that don't use the mixin - the one on product.product does return a value)

addons/website/models/ir_http.py
h4818: > I didn't really check what would be the implications of leaking this access_token though

It's just a fingerprinting identifier for a user: 
when they're logged in - the id, 
for public users it's the first 32 characters of a hash of their ip, user agent, and session id

as mentioned, this data gets deleted, so I don't think there are major implications :man_shrugging:

addons/website/controllers/main.py
h4818: Nice
solving with:
```py
res_model in request.env and hasattr(request.env[res_model], '_get_extra_tracking_values')
```

h4818: mixin is available and `hasattr` used

h4818:
@robodoo rebase-ff

antonrom1:
robodoo override=ci/security
`website.visitor` access_token is a fingerprinting identifier that doesn't give access to sensitive information

FrancoisGe:
@robodoo r+

h4818:
@robodoo r+

sobo-odoo:
@robodoo delegate+

FrancoisGe:
@robodoo r+

h4818 (commented):
@robodoo r+

PR #247465, opened 2026-02-05
[PERF] mrp: speed up _bom_find

mohannedah:
Before this commit, finding a bom for a recordset of `products` involved looping over all the boms and it will loop over all the `product_variant_ids` of `bom.product_tmpl_id` if the bom's `product_id` is NULL. This approach might loop over variants which we are not trying to find a bom for. In additon to that, due to the fact that multiple boms might have the same `product_tmpl_id`, this approach might consider the same variants in the inner loop redundantly even though we matched the variant with a bom in a previous itration.

Worst case, this might result in a time complexity of $O(N * M)$ where N is the number of boms and M is the number of variants.

To improve the performance, I only considered the variants given in the paramater `products` and in addition to that, I created a new dictionary mapping a `product_tmpl_id` to its bom if the bom doesn't have a variant set. By doing this, I can loop over the `products` given and if it doesn't have a bom set then it will be set to the one its template had taken from the previos loop.

In a method call with the following constraints
    - **2** products the method was finding a bom for
    - The 2 products had the same template and the template contained **550** active variants
    - The boms were only related to the template rather than the variants themselves.

| Input Size | Before | After |
| :--- | :--- | :--- |
| 100 | 0.78s | 0.03s |
| 1000 | 8.53s | 0.11s |
| 10000 | 80.99s | 0.73s |

---

### DIFF
--- a/addons/mrp/models/mrp_bom.py
+++ b/addons/mrp/models/mrp_bom.py
@@ -380,13 +380,16 @@ class MrpBom(models.Model):
             return bom_by_product
 
         boms = self.search(domain, order='sequence, product_id, id')
-
-        products_ids = set(products.ids)
+        bom_by_product_tmpl = defaultdict(lambda: self.env['mrp.bom'])
         for bom in boms:
-            products_implies = bom.product_id or bom.product_tmpl_id.product_variant_ids
-            for product in products_implies:
-                if product.id in products_ids and product not in bom_by_product:
-                    bom_by_product[product] = bom
+            if bom.product_id and (bom.product_id.product_tmpl_id not in bom_by_product_tmpl) and (bom.product_id not in bom_by_product):
+                bom_by_product[bom.product_id] = bom
+            elif not bom.product_id and bom.product_tmpl_id not in bom_by_product_tmpl:
+                bom_by_product_tmpl[bom.product_tmpl_id] = bom
+
+        for product in products:
+            if product.product_tmpl_id in bom_by_product_tmpl and product not in bom_by_product:
+                bom_by_product[product] = bom_by_product_tmpl[product.product_tmpl_id]
 
         return bom_by_product
 


addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
-            products_implies = bom.product_id or bom.product_tmpl_id.product_variant_ids
-            for product in products_implies:
-                if product.id in products_ids and product not in bom_by_product:
-                    bom_by_product[product] = bom
+            if bom.product_id and (bom.product_id.product_tmpl_id not in bom_by_product_tmpl) and (bom.product_id not in bom_by_product):
+                bom_by_product[bom.product_id] = bom
+            elif not bom.product_id and bom.product_tmpl_id and (bom.product_tmpl_id not in bom_by_product_tmpl):
+                bom_by_product_tmpl[bom.product_tmpl_id] = bom
+
+        for product in products.filtered(lambda product: product not in bom_by_product and product.active):
```
Sehnde:
```suggestion
        for product in products:
            if product in bom_by_product or not product.active:
                continue
```
instead of using `filtered`, you can simply add a condition inside the loop. That way, we don't loop twice on `products`

addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
+            if product.product_tmpl_id in bom_by_product_tmpl:
+                bom_by_product[product] = bom_by_product_tmpl[product.product_tmpl_id]
```
Sehnde:
```suggestion
        for product in products:
            if product.product_tmpl_id in bom_by_product_tmpl and product not in bom_by_product and product.active:
                bom_by_product[product] = bom_by_product_tmpl[product.product_tmpl_id]
```
instead of using `filtered`, you can simply use all the conditions inside the loop. That way, we don't loop twice on `products`

Sehnde (changes requested):
Please add more benchmarks: how much time is gained for 100, 1000, 10000 BOMs ?

addons/mrp/models/mrp_bom.py
mohannedah: Hello @Sehnde, Thank you for looking into this. I have applied the modification and added some benchmarks as well.

addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
-        products_ids = set(products.ids)
+        bom_by_product_tmpl = defaultdict(lambda: self.env['mrp.bom'])
         for bom in boms:
-            products_implies = bom.product_id or bom.product_tmpl_id.product_variant_ids
-            for product in products_implies:
-                if product.id in products_ids and product not in bom_by_product:
-                    bom_by_product[product] = bom
+            if bom.product_id and (bom.product_id.product_tmpl_id not in bom_by_product_tmpl) and (bom.product_id not in bom_by_product):
+                bom_by_product[bom.product_id] = bom
+            elif not bom.product_id and bom.product_tmpl_id and (bom.product_tmpl_id not in bom_by_product_tmpl):
```
Sehnde:
```suggestion
            elif not bom.product_id and (bom.product_tmpl_id not in bom_by_product_tmpl):
```
`bom.product_tmpl_id` is required, no need to check if there's one

addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
+                bom_by_product_tmpl[bom.product_tmpl_id] = bom
+
+        for product in products:
+            if product.product_tmpl_id in bom_by_product_tmpl and product not in bom_by_product and product.active:
```
Sehnde:
```suggestion
            if product.product_tmpl_id in bom_by_product_tmpl and product not in bom_by_product:
```
It is possible to do a `_bom_find` on an archived product by going into the bom overview of a parent bom. Now, when you archive a product, it automatically archives the linked boms but you can unarchive them.
This is a weird usecase but I still woudn't add a check on active in stable.

mohannedah: Actually the reason I only considered checking for active products was due to a failed test. `product_tmpl_id.product_variant_ids` only returns the active variants bypassing the archived products. I will try to remove it again and see the test case that fails for it.

mohannedah: Nevermind, it worked 😂

addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
         boms = self.search(domain, order='sequence, product_id, id')
-
-        products_ids = set(products.ids)
+        bom_by_product_tmpl = defaultdict(lambda: self.env['mrp.bom'])
         for bom in boms:
-            products_implies = bom.product_id or bom.product_tmpl_id.product_variant_ids
-            for product in products_implies:
-                if product.id in products_ids and product not in bom_by_product:
-                    bom_by_product[product] = bom
+            if bom.product_id and (bom.product_id.product_tmpl_id not in bom_by_product_tmpl) and (bom.product_id not in bom_by_product):
```
Sehnde, line 385:
```suggestion
            if bom.product_id and bom.product_id not in bom_by_product:
```
If the boms are reorganized, we could end up in a case where the template bom is given for a product has its own bom. E.g. :
- 1 template -> 2 variants A & B
- A has its own bom, the template has a BOM, B doesn't have a bom
- reorganize the boms so that the template bom is at the top -> lower sequence
- _bom_find returns the template for A instead of its own bom

addons/mrp/models/mrp_bom.py
```diff
@@ -380,13 +380,16 @@ def _bom_find(self, products, picking_type=None, company_id=False, bom_type=Fals
...
+                bom_by_product[bom.product_id] = bom
+            elif not bom.product_id and (bom.product_tmpl_id not in bom_by_product_tmpl):
```
Sehnde:
```suggestion
            elif not bom.product_id and bom.product_tmpl_id not in bom_by_product_tmpl:
```
you can remove the parentheses

addons/mrp/models/mrp_bom.py
mohannedah: https://github.com/odoo/odoo/blob/cb9bb5907e3e4e3629380ca61020f780e7fc5153/addons/mrp/models/mrp_bom.py#L411 

Hello. Wasn't it in the previous flow? In the code above the line `product not in bom_by_product` makes sure that the products that had a `bom` linked to the same template and was in an earlier iteration of the loop to skip the the product. Let me know if I 'm missing something.

Sehnde: Indeed, it's the current behavior, my mistake. I find it meh, imo the specific bom should be prioritized over the generic, but we can't change that in stable. You can leave it like that (but you can still remove the parentheses).

Sehnde (approved):

Sehnde:
@robodoo r+

Sehnde:
@mohannedah could you rebase your branch please ?

Sehnde:
robodoo r+

PR #247496, opened 2026-02-06
 [IMP] point_of_sale: allow configuring printer paper sizes

tong-odoo:
This update natively guarantees correct formatting by adding support for
standard 80mm and 58mm rolls, along with specific dimensions for IoT devices
like iMin and various Epson series (TM-U220, TM-U330, TM-P60, and TM-L100).

To prevent misaligned receipts across different hardware, a new setting now
dynamically adjusts POS ticket sizes based on the printer type and installed
modules. This guarantees correct formatting by supporting standard 80mm and
58mm rolls, along with specific dimensions for IoT devices like iMin and
various Epson series (TM-U220, TM-U330, TM-P60, and TM-L100).

task-5244833

### DIFF
--- a/addons/point_of_sale/models/pos_printer.py
+++ b/addons/point_of_sale/models/pos_printer.py
@@ -2,6 +2,7 @@
 
 from base64 import b32encode
 from hashlib import sha256
+
 from odoo import api, fields, models, _
 from odoo.exceptions import ValidationError
 
@@ -23,6 +24,17 @@ def format_epson_certified_domain(serial_number):
     sha256_hash = sha256(serial_number.encode()).digest()
     base32_text = b32encode(sha256_hash).decode().rstrip("=")
     return f"{base32_text.lower()}.{epson_domain}"
+
+
+EPSON_MODELS = [
+    ('tm_u22_76', 'TM-U220 series (76mm)'),
+    ('tm_u22_70', 'TM-U220 series (70mm)'),
+    ('tm_u22_58', 'TM-U220 series (58mm)'),
+    ('tm_u33_76', 'TM-U330 series (76mm)'),
+    ('tm_u33_70', 'TM-U330 series (70mm)'),
+    ('tm_p60_60', 'TM-P60 series (60mm)'),
+    ('tm_l100_40', 'TM-L100 series (40mm)'),
+]
 
 
 class PosPrinter(models.Model):
@@ -53,6 +65,13 @@ class PosPrinter(models.Model):
         ),
     )
     use_lna = fields.Boolean(string="Use Local Network Access")
+    paper_size = fields.Selection(string="Paper Size", selection=[
+        ('80', 'Standard 80mm'),
+        ('58', 'Standard 58mm'),
+        *EPSON_MODELS,
+    ], required=True, default='80')
+    paper_size_keys = fields.Char(compute='_compute_paper_size_keys')
+    timeout = fields.Integer(string="Connection Timeout (ms)", default=3000, help="Time in milliseconds before considering that the printer is not responding.")
 
     def copy_data(self, default=None):
         default = dict(default or {}, pos_config_ids=[(5, 0, 0)], printer_ip="0.0.0.0")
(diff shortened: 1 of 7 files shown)


addons/point_of_sale/models/pos_printer.py
```diff
@@ -25,6 +26,53 @@ def format_epson_certified_domain(serial_number):
     return f"{base32_text.lower()}.{epson_domain}"
 
 
+EPSON_MODELS = [
```
mobt-odoo, line 29: @tong-odoo Why do you make a list of all the devices? 
I think format and timeout is enough

tong-odoo: @fja-odoo wants users to select the printer version then will have the best fit of receipt size. Also there are special cases like u220 series, which paper size doesn't reflect the correct width should be used.

mobt-odoo: @tong-odoo as discussed with @fja-odoo 

put the epson list model in debug
and on the list, put the format this way. 

Epson model: 

- 80 mm
- 58 mm 
- tm_u22_76
- tm_u22_70
- tm_u22_58
- tm_u33_76
- tm_u33_70
- tm_p60_60
- tm_l100_40

The default being 80 mm 

thanks

tong-odoo: Okay, put the list model in debug means hide the paper size field and available with debug mode? or only debug mode will show epson model option?

mobt-odoo: @tong-odoo 
I think both settings should be in visible in debug mode only
Only a minority of use cases would require the users to actually care about it. 
image

Thanks

tong-odoo: Okay thanks!

qle-odoo:
@tong-odoo 
please write commit message.
what about community don't have access to Odoo ??
https://www.odoo.com/documentation/master/contributing/development/git_guidelines.html#commit-message-structure

tong-odoo:
@qle-odoo Thanks for the review, I have updated the commit message.

> what about community don't have access to Odoo ??

I am not quite sure what you mean by this. Could you elaborate more on this question?

addons/point_of_sale/static/src/app/utils/printer/epson_printer.js
```diff
@@ -129,7 +129,21 @@ export class EpsonPrinter extends BasePrinter {
...
         const protocol = this.use_lna ? "http:" : "https:";
-        return `${protocol}//${this.printer_ip}/cgi-bin/epos/service.cgi?devid=local_printer&timeout=3000`;
+        return `${protocol}//${this.printer_ip}/cgi-bin/epos/service.cgi?devid=local_printer&timeout=${this.timeout}`;
+    }
+
+    get STYLE_MAPPING() {
+        const base = super.STYLE_MAPPING;
+        return {
+            ...base,
+            tm_u22_76: { maxWidth: 200, fontSize: 12 },
```
adgu-odoo, line 139: Where does those maxWidth come from ?

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
```diff
@@ -330,6 +332,39 @@ export class PosTicketPrinterService {
...
+            `;
+        };
+
+        const cssRules = `
+            #pos-receipt { width: ${maxWidth}px !important; font-size: ${fontSize}px !important; }
+            /** Text classes **/
+            #pos-receipt .text-small { ${getFontRules(0.8)} }
+            #pos-receipt .text-normal { ${getFontRules(1.0)} }
+            #pos-receipt .text-large { ${getFontRules(1.2)} }
+            #pos-receipt .text-insane { ${getFontRules(2.2)} }
```
adgu-odoo, line 363: How did you decide those multipliers ?

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
```diff
@@ -330,6 +332,39 @@ export class PosTicketPrinterService {
...
+    async setIframeSizeFromPrinter(iframe, printer) {
+        const doc = iframe.contentDocument || iframe.contentWindow.document;
+        const iframeEl = doc.getElementById("pos-receipt");
+        const iframeHead = iframeEl.querySelector("head");
+        const { maxWidth, fontSize } = printer._instance.getStyle();
+
+        const style = document.createElement("style");
+        style.id = "printer-receipt-style";
+
+        const LINE_HEIGHT_RATIO = 1.4;
```
adgu-odoo, line 345: Where does this constant come from?

adgu-odoo (commented):
Very good PR, I just have some questions. Also I tested several sizes and I get some weird printing for some sizes. For example, the 40mm seems to show the images without adjusting their sizes:

Image

the u220 looks like its doing the same thing

Image

on the standard 58mm, it looks like the images are also not adjusted and the url for the invoice is not well displayed (some part are printed outside of the paper so we don't see them)

Image

I didn't check all size but it might be good to test every size with a real printer if possible

addons/point_of_sale/static/src/app/utils/printer/epson_printer.js
tong-odoo: image
The maxWidth is from the emulator attached from the task.

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
tong-odoo: Same as the font ratio. It is from `pos_receipt_common.xml`

addons/point_of_sale/static/src/app/services/pos_ticket_printer_service.js
tong-odoo: I used the font-size in `pos_receipt_common.xml` to calculate the ratio.

tong-odoo:
@adgu-odoo I updated the image max-width to ensure it display on the ticket, however if the paper size is too thin, the QR code might become too small for phone scanning.

adgu-odoo (commented):
robodoo r+

adgu-odoo (commented):
robodoo r+

PR #247504, opened 2026-02-06
[IMP] hr_attendance: restrict schedules and ruleset countries by wizard

tshe-odoo:
- Limit the working schedules to those belonging to the companies selected in the wizard, as well as common schedules not linked to any specific company.
- Restrict the country field on the overtime ruleset to only allow countries corresponding to the companies selected in the wizard.

task-5912498

### DIFF
--- a/addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
+++ b/addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
@@ -14,6 +14,7 @@ class HrAttendanceOvertimeRuleset(models.Model):
     country_id = fields.Many2one(
         'res.country',
         default=lambda self: self.env.company.country_id,
+        domain=lambda self: [('id', 'in', self.env.companies.country_id.ids)],
     )
     rate_combination_mode = fields.Selection([
             ('max', "Maximum Rate"),
--- a/addons/hr_attendance/views/hr_attendance_overtime_rule_views.xml
+++ b/addons/hr_attendance/views/hr_attendance_overtime_rule_views.xml
@@ -49,7 +49,7 @@
                                     <label for="expected_hours_from_contract" string="and"/>
                                     <field nolabel="1" name="timing_stop" widget="float_time" style="width:10ch"/>
                                 </span>
-                                <field name="resource_calendar_id" string="Schedule" invisible="base_off != 'timing' or timing_type != 'schedule'"/>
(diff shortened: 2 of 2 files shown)


addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
```diff
@@ -1,6 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from odoo import models, fields
+from odoo import models, fields, api
```
ravb-odoo: Should be in alphabetical order

addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
```diff
@@ -82,3 +83,9 @@ def action_show_versions(self):
     def copy_data(self, default=None):
         vals_list = super().copy_data(default=default)
         return [dict(vals, name=self.env._("%s (copy)", ruleset.name)) for ruleset, vals in zip(self, vals_list)]
+
+    @api.depends_context('allowed_company_ids')
+    def _compute_allowed_country_ids(self):
+        for record in self:
+            allowed_companies = self.env.companies
+            record.allowed_country_ids = allowed_companies.mapped('country_id')
```
ravb-odoo:
```suggestion
       self.allowed_country_ids = self.env.companies.mapped('country_id')
```
Since the value is identical for all records

addons/hr_attendance/models/hr_attendance_overtime_ruleset.py
```diff
@@ -15,6 +15,7 @@ class HrAttendanceOvertimeRuleset(models.Model):
         'res.country',
         default=lambda self: self.env.company.country_id,
     )
+    allowed_country_ids = fields.Many2many('res.country', compute='_compute_allowed_country_ids')
```
ravb-odoo: Instead of creating new field, directly add domain in country_id like 
```
country_id = fields.Many2one(
        'res.country',
        default=lambda self: self.env.company.country_id,
        domain=lambda self: [('id', 'in', self.env.companies.country_id.ids)]
    )
```

ravb-odoo (changes requested):
Thanks for your work. Here are some comments.

amah-odoo (approved):
Thanks for your work 🚀 

@robodoo r+

PR #247521, opened 2026-02-06
[FIX] iot_drivers: fix pos_id not being sent to SE blackbox

MaxTheWhale:
When using a v1 CleanCash blackbox, the command being sent to the blackbox was mistakenly sending a POS ID of " ". It just so happened this worked correctly when testing with our blackbox because it had " " registered as a POS ID. The POS ID is now sent correctly.

task-5077448

---


loouis-t (approved):
robodoo r+

PR #247530, opened 2026-02-06
[FIX] spreadsheet_dashboard: mark model as raw

LucasLefevre:
Since we split o-spreadsheet in two:
- core engine, independent from owl and DOM
- UI

The `Model` doesn't mark itself as raw for owl's reactivity.

It must be done manually at every call site where it's needed.

Here the entire `DashboardLoader` is made reactive, including its properties.

We cannot let the model be reactive as it would hurt performance very badly.

From a DX POV, the current situation is a bad design because the developer doesn't know the model needs to be marked as raw, and even if the developer knows it, it's very easy to overlook (as we did and this commit proves)

It would be probably better with a private constructor and factory methods with clear names.

Task-5916564

Current behavior before PR:

---


rrahir:
robodoo r+

PR #217277, opened 2025-07-03
[IMP] core: many2many access check

kmagusiak:
Description of the issue/feature this PR addresses:

- many2many.auto_join implemented to bypass access rights (just like on one2many)
- when writing on a many2many field, check read access on the corecords - to avoid linking to records that are not accessible by the user
- remove the `BaseModel._apply_ir_rules` function and use `_search` instead

Current behavior before PR:
Model access in not checked on many2many fields when reading them, only access rules are applied.

Desired behavior after PR is merged:
Access checks are performed when reading and read access is checked when writing on many2many fields.

task-4067945
odoo/enterprise#89371

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247531, opened 2026-02-06
[FIX] *: make basic sale/stock/purchase flows with minimal access rights

lase-odoo:
*: account,sale,stock_delivery

Since [19.0](https://github.com/odoo/odoo/pull/217277#issue-3198442339), read access rights are checks on comodels when trying to read the value of a many2many fields you have read acccess to. This change highlight numerous access right issues in basic flows for users with minimal access.

Here is a list of examples (each performed with every other access rights disabled):
- With a `stock user`, open the delivery list or form view
 #### > Access error
- With a `purchase user` create and confirm a PO > Upload Bill
####  > the Bill will be created but an access error will prevent the draft bill from opening.
- With a `sale user` create and confrim an SO > Create invoice
### Use case: Open an invoice (`acount.move`) linked to one of your SO/PO with a basic `sale`/`purchase` user:

1) For basic `sale` and `purchase` users to be able to open the `account.move` Form on which they have read, update, create, delete access rights, it is necessary for the `payment_count` to be compute sudo since it is used in the view:
https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/account/views/account_move_views.xml#L859 And these users do not have the read access of the `account.payment` model.

Similarily the `_compute_asset_ids` needs to be compute sudo because it relies on the related `asset_ids` of `account.move.line`s or on values of these `account.asset`s for which the users shoud not have read access: https://github.com/odoo/enterprise/blob/06cf3f2b6663c61f1fcc158678dfb1fa43ece4e1/account_asset/models/account_move.py#L27-L30 https://github.com/odoo/enterprise/blob/06cf3f2b6663c61f1fcc158678dfb1fa43ece4e1/account_asset/models/account_move.py#L317-L323 and the `asset_ids`, `count_asset`, `asset_id_display_name` and `draft_asset_exists` are all used in the view.

#### Note for master:

IMO, the `asset_ids` field of the `account.move` model should probably be in a separate compute to not be computed in sudo and removed from the views as it is currently used only to determine if there is or not `asset_ids`. An information that is provided by the `count_asset`. E.G. here:
https://github.com/odoo/enterprise/blob/06cf3f2b6663c61f1fcc158678dfb1fa43ece4e1/account_asset/views/account_move_views.xml#L10 https://github.com/odoo/enterprise/blob/06cf3f2b6663c61f1fcc158678dfb1fa43ece4e1/account_asset/views/account_move_views.xml#L35-L40 https://github.com/odoo/enterprise/blob/06cf3f2b6663c61f1fcc158678dfb1fa43ece4e1/account_asset/views/account_move_views.xml#L44-L52

2) For basic `purchase` users to open the invoice linked to one of their PO, it is necessary that the `sale_order_count` is computed in sudo as they do not have access to the related `sale_line_ids` field and the field is used in the `account.move` form:
https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/sale/models/account_move.py#L46-L49 https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/sale/views/account_views.xml#L53

### Use case: Open a `stock.picking` views as basic stock user:

3) The basic `stock` users have a read access on the `delivery.carrier` model and should also on the related `delivery.zip.prefix` and `delivery.price.rule` models. First as it make sense functionally but also as it currently blocks them on basic flows. For instance basic stock users can not open the `stock.picking` list or form view as the `carrier_id` is part of these view:
https://github.com/odoo/odoo/blob/4ea42f8b16a8619cced4255f5bba8de6427345b7/addons/stock_delivery/views/delivery_view.xml#L125-L132 This is problematic as this field has a domain relying on the related `allowed_carrier_ids` field:
https://github.com/odoo/odoo/blob/93fa6d9fff63534cfa9251e21fc82797d8b83468/addons/stock_delivery/models/stock_picking.py#L23-L24 As such, when the view is opened, the related field needs to be read. However, the `_compute_allowed_carrier_ids` fails if you do not have read access rights on the `delivery.zip.prefix` model: https://github.com/odoo/odoo/blob/93fa6d9fff63534cfa9251e21fc82797d8b83468/addons/delivery/models/delivery_carrier.py#L198 https://github.com/odoo/odoo/blob/93fa6d9fff63534cfa9251e21fc82797d8b83468/addons/delivery/models/delivery_carrier.py#L70-L72 Similarily `delivery.price.rule` model should be readable for `stock` users in order to be able to get be able to rely on the `price_rule_ids` when necessary such as here:
https://github.com/odoo/odoo/blob/93fa6d9fff63534cfa9251e21fc82797d8b83468/addons/delivery/models/delivery_carrier.py#L492-L496

---

opw-5461135
opw-5417749

### DIFF
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -223,7 +223,7 @@ class AccountMove(models.Model):
         search='_search_reconciled_payment_ids',
         help='Payments that have been reconciled with this invoice.'
     )
-    payment_count = fields.Integer(compute='_compute_payment_count')
+    payment_count = fields.Integer(compute='_compute_payment_count', compute_sudo=True)
 
     # === Statement fields === #
     statement_line_id = fields.Many2one(
--- a/addons/purchase/tests/test_purchase_flow.py
+++ b/addons/purchase/tests/test_purchase_flow.py
@@ -1,0 +1,32 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import Command
+from odoo.tests import HttpCase, tagged
+
+from odoo.addons.account.tests.common import AccountTestInvoicingCommon
+
+
+@tagged('-at_install', 'post_install')
+class TestPurchaseFlowTourPostInstall(AccountTestInvoicingCommon, HttpCase):
+
+    def test_basic_purchase_flow_with_minimal_access_rights(self):
+        """
+        Test that a purchase user with minimal access rights can open both the list and form view,
+        create and process a purchase order, upload and open the associated vendor bill.
+        """
+        purchase_user = self.env['res.users'].create({
+            'name': 'Super Purchase Woman',
+            'login': 'SuperPurchaseWoman',
+            'group_ids': [Command.set([self.ref('purchase.group_purchase_user')])],
+        })
+        # create and confirm a PO to populate the list view
+        purchase_order = self.env['purchase.order'].with_user(purchase_user.id).create({
+            'partner_id': self.partner_a.id,
+            'order_line': [Command.create({
+                'name': self.product.name,
+                'product_id': self.product.id,
+                'product_uom_qty': 1,
+            })],
+        })
+        purchase_order.button_approve()
+        self.start_tour('/odoo', 'test_basic_purchase_flow_with_minimal_access_rights', login='SuperPurchaseWoman')
--- a/addons/account/views/account_move_views.xml
+++ b/addons/account/views/account_move_views.xml
@@ -853,6 +853,7 @@
                                     </div>
                             </button>
                             <button name="open_payments"
+                                    groups="account.group_account_invoice,account.group_account_readonly"
                                     class="oe_stat_button"
                                     icon="fa-bars"
                                     type="object"
--- a/addons/purchase/static/tests/tours/purchase_flow_tour.js
+++ b/addons/purchase/static/tests/tours/purchase_flow_tour.js
@@ -1,0 +1,72 @@
+import { inputFiles } from "@web/../tests/utils";
+import { registry } from "@web/core/registry";
+
+registry.category("web_tour.tours").add("test_basic_purchase_flow_with_minimal_access_rights", {
+    steps: () => [
+        {
+            trigger: ".o_menuitem[href='/odoo/purchase']",
+            run: "click",
+        },
+        {
+            content: "Check that at least one RFQ is present in the view",
+            trigger: ".o_purchase_dashboard_list_view .o_data_row",
+        },
+        {
+            trigger: ".o_list_button_add",
+            run: "click",
+        },
+        {
+            trigger: ".o_input[id=partner_id_0]",
+            run: "edit partner_a",
+        },
+        {
+            trigger: ".dropdown-item:contains(partner_a)",
+            run: "click",
+        },
+        {
+            trigger: ".o_field_x2many_list_row_add > a",
+            run: "click",
+        },
+        {
+            trigger: ".o_data_row .o_input",
+            run: "edit Test Product",
+        },
+        {
+            trigger: ".dropdown-item:contains('Test Product')",
+            run: "click",
+        },
+        {
+            trigger: ".o_data_cell[name=price_unit]",
+            run: "click",
+        },
+        {
+            trigger: ".o_data_cell[name=price_unit] .o_input",
+            run: "edit 3",
+        },
+        {
+            trigger: "button[name=button_confirm]",
+            run: "click",
+        },
+        {
+            trigger: ".o_statusbar_status .o_arrow_button_current:contains(Purchase Order)",
+        },
+        {
+            content: "Upload the vendor bill",
+            trigger: ".o_widget_purchase_file_uploader",
+            run: async () => {
+                const testFile = new File(["Vendor, Bill"], "my_vendor_bill.png", {
+                    type: "image/*",
+                });
+                await inputFiles(".o_widget_purchase_file_uploader input", [testFile]);
+            },
+        },
+        {
+            content: "Check that we are in the invoice form view",
+            trigger: ".o_statusbar_status:contains(Posted) .o_arrow_button_current:contains(Draft)",
+        },
+        {
+            content: "Check that the invoice is linked to the sale order",
+            trigger: "button[name=action_view_source_purchase_orders]",
+        },
+    ],
+});
--- a/addons/purchase/tests/__init__.py
+++ b/addons/purchase/tests/__init__.py
@@ -3,6 +3,7 @@
 
 from . import test_purchase
 from . import test_purchase_downpayment
+from . import test_purchase_flow
 from . import test_purchase_order_product_catalog
 from . import test_purchase_order_report
 from . import test_purchase_invoice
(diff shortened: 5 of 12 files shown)


rco-odoo (approved):
It's actually very nice to finally test those simple business flows with a _regular_ user.  Kudos for finding minimal fixes to make those flows work as expected!

addons/purchase/tests/test_purchase_flow.py
lordkrandel: indent

addons/sale/tests/test_sale_flow.py
lordkrandel: indent

addons/account/models/account_move.py
```diff
@@ -223,7 +222,7 @@ def _sequence_year_range_monthly_regex(self):
         search='_search_reconciled_payment_ids',
         help='Payments that have been reconciled with this invoice.'
     )
-    payment_count = fields.Integer(compute='_compute_payment_count')
+    payment_count = fields.Integer(compute='_compute_payment_count', compute_sudo=True)
```
lordkrandel, line 226: You are putting `compute_sudo` to enable visibility to this button, that if you click, you get an access error anyway.

Image
Image

You're not allowed to see payments from invoices if you don't have any accounting rights, at the moment. This was also true in 18.0.

If this is the only reason we are adding `compute_sudo`, please add `groups` on the button instead, to avoid security issues.

lordkrandel (changes requested):
I like the idea, I like really much the tests. 🎉 
I don't know about `compute_sudo` 🤔

addons/purchase/tests/test_purchase_flow.py
lordkrandel: I don't know why Github is misbehaving with this comment, sorry.

```py
                'product_uom_qty': 1,
            })],
```

addons/account/models/account_move.py
lase-odoo: Thanks for the review @lordkrandel ! 🌊

IMO, computed counting fields that do not provide sensitive information should be `compute_sudo`. First because the read access of the field itself is protected by the ACL of the model (only users with the read access of the account.move model will be able to read it) and since they should provide a consistent result no matter the user. Note that there are already places in the code where we remove record sets you do not have access to rather than raising an error:
https://github.com/odoo/odoo/blob/e2239cd67fe57ed8158f1c230543e035c41819f9/odoo/orm/fields_relational.py#L1392-L1404
Hence, if we fall in such cases, we could endup using an erroneous value on the count because of the user that read it and this value might be used by other computational logic leading to other unexpected result for no valid reasons.


I am not against adding a group on the button on the view, but to me it presents a few drawbacks:
1) It is not a good stable fix, since users that do not update the view will still be blocked by access errors.
2) It does not look good with respect to scalability. For instance, if I add a the `account.group_account_readonly` on the button because we are not expecting other users to open that record set in the account module and an other module alter provide the read access of this record set to an other user group, say pos managers, these pos managers will not see the button because they are not account users even thought they are expected to be able to read its content.
The same holds if the compute method is overridden in a module and that override uses record set to which your account users do not have access to, your `group_account_readonly` user will not be able to read the count simply because it depends on a model they do not have access to. Here is an interesting example from loyalty that could suffer from it if the `pos_order_count` and `order_count` depends on models your basic loyalty user do not have access to:
https://github.com/odoo/odoo/blob/1a62c4333de61a4e1358ef8e229b2e0f5a3e9826/addons/loyalty/models/loyalty_program.py#L207-L208
https://github.com/odoo/odoo/blob/1a62c4333de61a4e1358ef8e229b2e0f5a3e9826/addons/pos_loyalty/models/loyalty_program.py#L99-L102
https://github.com/odoo/odoo/blob/1a62c4333de61a4e1358ef8e229b2e0f5a3e9826/addons/sale_loyalty/models/loyalty_program.py#L23-L26

Let mo know what you think 🌻

lordkrandel: I don't want to block the PR, I'll just make some comments about what I think 
If any other reviewer wants to merge after considering them, feel free.

- > It is not a good stable fix, since users that do not update the view will still be blocked by access errors.

  True, strong point, let's consider then to remove `compute_sudo` in `master`?
  I fear that `compute_sudo` spreads to every view that has a `compute` on the `comodel`.
  I don't like `sudo` as a default answer to this problem.
  If the ORM check was changed, I guess it was for a reason.

- > scalability

  This is a matter of how we handle security in general, not a problem of `payment_count`.
  The more the code grows, the more ACL combinations will grow.

> IMO, computed counting fields that do not provide sensitive information should be `compute_sudo`.

The number of payments is kinda sensitive to me 😅  (Insider trading...?)

> they should provide a consistent result no matter the user. 

If you have access to 2 moves out of 6 total - you should see 2 as count. (edit: or use sudo() explicitly if you need strange computations)

> Note that there are already places in the code where we remove record sets you do not have access to rather than raising an error

The rule allows for access to the model, but not to all records of that model. Doesn't mean you can count the records you don't have access to.

>  If I add a the `account.group_account_readonly` on the button 
> [...] PoS managers will not see the button

You can add `compute_sudo` if it makes sense, but here it's taken as a default.
It's no good that a sales manager sees how many payments _and_ assets _and_... whatever is linked to a move, if they don't have access to the related models 🤔 To me it's just a bug.
We introduced the check, I guess for some good reasons.

> `pos_order_count` and `order_count` depends on models your basic loyalty user do not have access to.

So we inherit computed fields on models we have no access to? ... wow! 😊 
I guess in those cases a `compute_sudo` wouldn't hurt, but I would take that as an exception.

If we just think that recordsets should be counted regardless of access, at least in views, then we should be able to do it programmatically in the framework, like
```xml
    payment_count = fields.Integer(count="payment_ids")
```
which has `compute_sudo()` and `api.depends` automatically set

Feyensv (approved):
LGTM for `sale`, thanks for covering those flows with clean tests :pray:

FYI there are some tour utils in web & sale that could be used to simplify some steps of your tour (it makes it easier for us when we change stuff to only update the utils and not each tour separately).
Also, wouldn't it also make sense to update the `test_main_flows` module tests to try running main flows with the least rights possible, instead of `admin` ones ?

addons/account/models/account_move.py
lase-odoo: @lordkrandel I'll update the views as well then so that the count should not be visible. And I'll remove the `compute_sudo` in master then.

lase-odoo:
@lordkrandel I have updated the views regarding `payment_count` and `asset` related fields.

@Feyensv I have updated the sale test to rely on utils. I also removed the change in the "lazada" module as the csv file was already updated in https://github.com/odoo/enterprise/commit/892c181f7adebdd39659b214b760e732ebf32632 Regarding the `test_main_flows` to rely on minimal access related to the flow I think it is a good idea but due to the size of the test it will take some time and I think it would be best to do it on a separate PR to merge the current changes asap. Let me know what you think !

lordkrandel (approved):
Ok to me, for the accounting part. 
`compute_sudo` will be dealt with in `master`, as per discussion.

Feyensv (approved):
@robodoo delegate+

Indeed, was hinting at improving `test_main_flows`, but it certainly doesn't have to be done right here, it's already very good to merge those fixes (and tests) 👍

PR #247537, opened 2026-02-06
[FIX] l10n_it_edi: fix total in company currency

gteboul:
With an `l10n_it` company:
- Create an invoice in a foreign currency and export the XML.

In the XML, the field `ImportoTotaleDocumento` is expressed in the invoice currency, while `Divisa` is set to the company currency.
`ImportoTotaleDocumento` must be expressed in EUR, as stated in the following document:
https://www.agenziaentrate.gov.it/portale/documents/20143/0/10FAQ+pubblicate+il+19+luglio+2019+(aggiornate+il+1+luglio+2021).pdf/6e8acb34-b7c5-730e-6a39-79c0a296e02b

> L'art. 21, comma 2, lettera l) del d.P.R. n. 633/72 specifica che
> “aliquota, ammontare dell'imposta e dell'imponibile con arrotondamento al centesimo di euro”.
> Conseguentemente, se la fattura è emessa da soggetti residenti o stabiliti il codice da inserire nel campo `<Divisa>`
> deve essere obbligatoriamente “EUR”.

> Article 21, paragraph 2, letter (l) of Presidential Decree No. 633/72 specifies that
> “the tax rate, the amount of the tax, and the taxable amount must be stated with rounding to the euro cent.”
> Consequently, if the invoice is issued by resident or established taxpayers, the code to be entered in the `<Divisa>`
> field must mandatorily be “EUR”.

As a result, we now use the **company currency amount** for the `ImportoTotaleDocumento` field
(assuming the company currency is always EUR for `l10n_it`).

Ticket [link](https://www.odoo.com/odoo/project.task/5913088)
opw-5913088

### DIFF
--- a/addons/l10n_it_edi/models/account_move.py
+++ b/addons/l10n_it_edi/models/account_move.py
@@ -370,7 +370,7 @@ class AccountMove(models.Model):
 
         # Self-invoices are technically -100%/+100% repartitioned
         # but functionally need to be exported as 100%
-        document_total = self.amount_total
+        document_total = self.amount_total_signed if convert_to_euros else self.amount_total
         if is_self_invoice:
             document_total += sum([abs(v['tax_amount_currency']) for k, v in tax_details['tax_details'].items()])
             if reverse_charge_refund:
--- a/addons/l10n_it_edi/tests/export_xmls/export_foreign_currency_global_discount.xml
+++ b/addons/l10n_it_edi/tests/export_xmls/export_foreign_currency_global_discount.xml
@@ -58,7 +58,7 @@
         <Divisa>EUR</Divisa>
         <Data>2024-08-07</Data>
         <Numero>INV/2024/00003</Numero>
-        <ImportoTotaleDocumento>512.07</ImportoTotaleDocumento>
+        <ImportoTotaleDocumento>474.63</ImportoTotaleDocumento>
       </DatiGeneraliDocumento>
     </DatiGenerali>
     <DatiBeniServizi>
--- a/addons/l10n_it_edi/tests/export_xmls/export_foreign_currency_global_simple_discount.xml
+++ b/addons/l10n_it_edi/tests/export_xmls/export_foreign_currency_global_simple_discount.xml
@@ -59,7 +59,7 @@
         <Divisa>EUR</Divisa>
         <Data>2024-08-06</Data>
         <Numero>INV/2024/00002</Numero>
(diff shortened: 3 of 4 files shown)


lordkrandel (changes requested):
Hmmm... there's a bit of confusion here

- Can you please verify whether this fix is due also in 17.0?
  It would lead to incorrect computations for the Tax Agency.
  Look for `document_total` instead of `importo_totale_documento`, there is some _(pseudocodely)_->`move.amount_total - sum(tax_amount_currency)` for the reverse charge case.
- The reference to the document is actually inexact.
  In the commit message you say that the doc states something about `ImportoTotaleDocumento`, which is not mentioned, but:
  - `fex-app.com` tells you that `ImportoTotaleDocumento` should be `ImponibileImporto (base) + Imposta (tax) + ScontoMaggiorazione (discount)` 
  - the document tells you that these values must in EUR.
  So it's a bit more indirect.
- I... totally do not understand how the `currency` field evolved in our code 🤣  Looks to me like a stupid result from clever iterations. Can you verify what I'm saying? Maybe I'm missing something. If it's correct, it might be something that we clarify in `master`.

```xml
<Divisa t-out="currency.name"/>
```
and while determining the values:
```py
convert_to_euros = self.currency_id.name != 'EUR'
```
```py
'currency': self.currency_id or self.company_currency_id if not convert_to_euros else self.env.ref('base.EUR'),
```

So, if `not convert_to_euros --> self.currency_id.name == 'EUR'`
and this means:
- `self.currency_id or self.company_currency_id if not convert_to_euros` is `EUR`
- `else self.env.ref('base.EUR')` is also `EUR`

This tells us that your premise (assuming the company currency is `EUR`) actually means that we are justifiedly arbitrarily always use `EUR`.

... so couldn't we just write `EUR` in the `XML`? 🤔

addons/l10n_it_edi/models/account_move.py
```diff
@@ -409,7 +409,7 @@ def _l10n_it_edi_get_values(self, pdf_values=None):
             'origin_document_date': self.l10n_it_origin_document_date,
             'cig': self.l10n_it_cig,
             'cup': self.l10n_it_cup,
-            'currency': self.currency_id or self.company_currency_id if not convert_to_euros else self.env.ref('base.EUR'),
+            'currency': self.env.ref('base.EUR'),
```
lordkrandel: This is correct, and I know I pointed that out, sorry, but let's leave it out for another PR, please, as it isn't part of the fix.

PR #247538, opened 2026-02-06
[FIX] account: fix computation of residual_at_date with old reconcile

xavieralt:
Following odoo/odoo@de450d20c239, old journal entries that have both debit and credit matching entries (partial reconcile) were not considered anymore as fully reconciled when asking of `recon_date` further than the last payment; which greatly inpact the `Aged Receivable` and `Aged Payable` reports.

This commit fix it by grouping together matching debit / credit for the same currency, this to match the computation of the `amount_residual` field.

no-task (from feedback pad)

---

### DIFF
--- a/addons/account/tests/test_account_move_reconcile.py
+++ b/addons/account/tests/test_account_move_reconcile.py
@@ -2807,6 +2807,56 @@ class TestAccountMoveReconcile(AccountTestInvoicingCommon):
             ]
         )
 
+    def test_residual_at_date_manual_exchange_diff_entry(self):
+        company_currency = self.company_data['currency']
+        foreign_currency = self.other_currency
+
+        line_1 = self.create_line_for_reconciliation(92.11, 99.34, foreign_currency, '2016-01-01')
+        line_2 = self.create_line_for_reconciliation(-92.11, -92.11, company_currency, '2016-02-01')
+        line_3 = self.create_line_for_reconciliation(-0, 0.76, foreign_currency, '2016-02-01')
+        all_lines = line_1 | line_2 | line_3
+
+        # Create the partials as it should be created in previous version.
+        self.env['account.partial.reconcile'].create([
+            {
+                'amount': 92.11,
+                'debit_move_id': line_1.id,
+                'debit_amount_currency': 100.1,
+                'debit_currency_id': foreign_currency.id,
+                'credit_move_id': line_2.id,
+                'credit_amount_currency': 92.11,
+                'credit_currency_id': company_currency.id,
+            },
+            {
+                'amount': 0.0,
+                'debit_move_id': line_3.id,
+                'debit_amount_currency': 0.76,
+                'debit_currency_id': foreign_currency.id,
+                'credit_move_id': line_1.id,
+                'credit_amount_currency': 0.76,
+                'credit_currency_id': foreign_currency.id,
+            },
+        ])
+        self.assertTrue(all(line.reconcile for line in all_lines))
+
+        # "line_1" both have matching credits and debits, ensure we correctly
+        # compute residual at date
+        self.assertRecordValues(all_lines, [
+            {'amount_residual': 0.0, 'residual_at_date': 0.0},
+            {'amount_residual': 0.0, 'residual_at_date': 0.0},
+            {'amount_residual': 0.0, 'residual_at_date': 0.0},
+        ])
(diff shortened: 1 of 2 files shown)


xavieralt:
@h4818 @william-andre you feedback will be much appreciated on this - we got some major diff from this while testing our upgrade to `saas~19.1` :pray:

addons/account/tests/test_account_move_reconcile.py
```diff
@@ -2807,6 +2807,56 @@ def test_residual_at_date(self):
...
+                'credit_move_id': line_2.id,
+                'credit_amount_currency': 92.11,
+                'credit_currency_id': company_currency.id,
+            },
+            {
+                'amount': 0.0,
+                'debit_move_id': line_3.id,
+                'debit_amount_currency': 0.76,
+                'debit_currency_id': foreign_currency.id,
+                'credit_move_id': line_1.id,
```
william-andre, line 2835: It doesn't make sense to have one line both on `debit_move_id` and `credit_move_id` 🤯 

Either it is a debit line, or a credit line...
I'm sure other parts of the code are failing too if these are inverted somehow.

xavieralt: @william-andre we can't change the history, also this align the algo. with the computation of the `amount_residual` with is doing `amount_residual = balance - matched_debit_amount - matched_credit_amount` (see [^1])

[^1]: https://github.com/odoo/odoo/blob/a4fae96d53d65f5ef07bff4525e39a9b4ca12030/addons/account/models/account_move_line.py#L932

william-andre: Actually, yes we can rewrite the history. :shrug: 
There are other places where we look at the side of the balance:
* https://github.com/odoo/enterprise/blob/9bae1afecf4b7064f2cb3295282727e9b48ba5d0/account_reports/models/account_cash_flow_report.py#L422
* https://github.com/odoo/enterprise/blob/9bae1afecf4b7064f2cb3295282727e9b48ba5d0/account_reports/models/account_multicurrency_revaluation_report.py#L266
* https://github.com/odoo/enterprise/blob/9bae1afecf4b7064f2cb3295282727e9b48ba5d0/account_reports/models/account_partner_ledger.py#L30

In many cases, even if we take the balance and consider a credit of `-x` instead of a debit of `x`, it is wrong in the context of double entry bookkeeping

Your fix looks like just a band-aid to cover up the real issue

xavieralt: Ok, then you need to write an upgrade script to rewrite those entries.

qdp-odoo: the thing is: indeed if the `account.partial.reconcile` between `line_1` and the exchange diff `line_3` would have been created with `debit_move_id` and `credit_move_id` inverted, the reconciliation would be considered full and both the `amount_residual` and `residual_at_date` would be 0, as expected. 

Which means: it's easily fixable.

```python
self.env['account.partial.reconcile'].create([
    {
        'amount': 92.11,
        'debit_move_id': line_1.id, 
        'debit_amount_currency': 100.1, 
        'debit_currency_id': foreign_currency.id, 
        'credit_move_id':line_2.id, 
        'credit_amount_currency': 92.11,
        'credit_currency_id': company_currency.id,
    },
    {
        'amount': 0.0,  
        'credit_move_id': line_3.id,    
        'credit_amount_currency': 0.76,        
        'credit_currency_id': foreign_currency.id,        
        'debit_move_id': line_1.id,        
        'debit_amount_currency': 0.76,        
        'debit_currency_id': foreign_currency.id,
    },
])
```

So before a line wit 0 debit/credit and some amount_currency could been used anyway to fix , but now apparently it's not the case anymore... but from purely theoretical point of view, the line_3 is a debit (balance 0, but amount_currency > 0), and should be used in debit_move_id. 

I don't find that more correct, TBH, @william-andre

krge-odoo:
Hello @qdp-odoo , @william-andre and @xavieralt ,

I have similar issue on my ticket (https://www.odoo.com/odoo/my-tasks/6193686).

cyro-odoo:
Hello @xavieralt , @william-andre , @qdp-odoo ,
Is there any update on this PR? Just wondering what I should communicate to the customer in terms of when a resolution may be merged.

Thanks!

PR #247540, opened 2026-02-06
[FIX] account: allow translation reload without all companies

madprog:
When a user doesn't have access to all companies, he couldn't reload the translation terms.

However, the exception occurs in the call to _get_chart_template_data, which doesn't especially require privileges, because it is static data.

With this commit, we bypass the lack of company access to retrieve this data, and let the user continue the language reloading process.

Task-id: [5916490](https://www.odoo.com/odoo/project.task/5916490)


Levizar (approved):
@robodoo r+

testing it correctly is a bit cumbersome and this kind of issue isn't supposed to happen a lot so we prefer to merge it without the test. (The one we had needs several l10n modules to be installed)

PR #217853, opened 2025-07-08
[FIX] hr_homeworking: added necessary values to avoid unwanted errors

bhna-odoo:
Currently an error occurs when we open Employees app with a non admin account.

Steps to replicate:

- Install `hr_homeworking` with demo data.
- Login with admin and open `Employee > Mitchell Admin` form view.
- On the Work tab you'll find the seven days of the week, set today's field to 'Office'. (For Ex: If Today is monday so fill in Office for monday).
- Now Logout and login through some other account like Marc Demo. Open Employees and you will get the error.

Error:
`ValueError: Wrong value for hr.employee.public.hr_icon_display: 'presence_office'`

This error occurs after a recent [refactor](https://github.com/odoo/odoo/pull/202869) in the code and that time they forgot to add the other values for `hr_icon_display` field.

In saas-18.3 [1], the code  directly  inherited from the  base model `hr.employee.base`. However, in saas-18.4 [2], the refactor split the code into `hr.employee` and `hr.employee.public`, only `hr.employee` includes the new selection items, while `hr.employee.public` does not. This caused the wrong value error.

[1] - https://github.com/odoo/odoo/blob/aa8b09ca61dfd405ebebbeff4e45a08bf7ad84fd/addons/hr_homeworking/models/hr_employee.py#L8-L24

[2] - https://github.com/odoo/odoo/blob/850e575b71b04597d31140cec53a596456370325/addons/hr_homeworking/models/hr_employee.py#L8-L24

This commit solves this issue by adding the necessary values to the `hr_icon_display` field.

sentry-6731346730
---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247542, opened 2026-02-06
[IMP] core: support selection_add display without database records

HydrionBurst:
Previously, when developers use `selection_add` in stable versions, the new selection values could be written to the database (since `field._selection` contains them), but they would not appear in the UI due to missing `ir.model.fields.selection` records.

This commit allows newly added selections to be displayed in the field description without requiring database updates.

Note: the new selections still cannot be translated without database updates.

mainly for enterprise https://github.com/odoo/enterprise/pull/106612
may be also helpful for https://github.com/odoo/odoo/pull/217853

Current behavior before PR:

---

### DIFF
--- a/odoo/orm/fields_selection.py
+++ b/odoo/orm/fields_selection.py
@@ -206,11 +206,8 @@ class Selection(Field[str | typing.Literal[False]]):
             # force all values to be strings (check _get_year_selection)
             return [(str(key), str(label)) for key, label in selection]
 
-        # translate selection labels
-        if env.lang:
-            return env['ir.model.fields'].get_field_selection(self.model_name, self.name)
-        else:
-            return selection
+        translations = dict(env['ir.model.fields'].get_field_selection(self.model_name, self.name))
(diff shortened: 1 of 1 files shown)


odoo/orm/fields_selection.py
```diff
@@ -207,10 +207,12 @@ def _description_selection(self, env):
...
-        if env.lang:
-            return env['ir.model.fields'].get_field_selection(self.model_name, self.name)
-        else:
-            return selection
+        selection_dict = {key: (key, label) for key, label in selection}
+        selection_dict.update(
+            {key: (key, label)
+            for key, label in env['ir.model.fields'].get_field_selection(self.model_name, self.name)
+        })
+        return list(selection_dict.values())
```
kmagusiak: You can use the label instead of tuple[key, label] in the selection and `return list(selection_dict.items())`

kmagusiak (approved):
@robodoo r+

PR #247546, opened 2026-02-06
[FIX] web: use fields inside the records instead of the ones in filters

xavierbol:
Before this commit, when the field of the model is not the same then the
writable field for the filters, the user cannot correctly load the
calendar view since the view does not find the write_field in the record to
get the value.

This commit fixes the issue to correctly use the right field depending on
which case we are.

task-5259036


deneuvillem (commented):
LGTM, thanks for your work!

xavierbol:
robodoo r+

PR #247549, opened 2026-02-06
[FIX] account: remove comments while parsing XMLs

clom-odoo:
Issue:
Comments in node with data as text may cause issue while reading data

Step to reproduce:
- In European company create an invoice to a company in the same country
- Send it with peppol
- Download the XML of the invoice
- Add a comment right before the value in TaxExclusiveAmount like this <cbc:TaxExclusiveAmount>0.00</cbc:TaxExclusiveAmount>
- Go to accounting Dashboard
- Click on "Upload" in the vendor dashboard
- Select the previously modified XML

Current behavior:
- an empty invoice is created as the invoice builder failed

Expected behavior:
- Invoice should be recognized

Cause:
lxml treat data like this:
<parentNode>parent_text<childNode>child_text</childNode>child_tail</parentNode>parent_tail So `tax_exclusive_amount_node.text` return '' instead of `0.00`

opw-5462664

---

### DIFF
--- a/addons/account_edi/models/account_edi_format.py
+++ b/addons/account_edi/models/account_edi_format.py
@@ -226,7 +226,7 @@ class AccountEdiFormat(models.Model):
         """
         to_process = []
         try:
-            xml_tree = etree.fromstring(content)
+            xml_tree = etree.fromstring(content, parser=etree.XMLParser(remove_comments=True, resolve_entities=False))
         except Exception as e:
             _logger.exception("Error when converting the xml content to etree: %s" % e)
             return to_process
--- a/addons/account_edi_ubl_cii/tests/test_ubl_cii.py
+++ b/addons/account_edi_ubl_cii/tests/test_ubl_cii.py
@@ -182,6 +182,39 @@ class TestAccountEdiUblCii(AccountTestInvoicingCommon):
             'zip': '8010',
         }])
 
+    def test_import_bill(self):
+        partner = self.env['res.partner'].create({
+            'name': "My Belgian Partner",
+            'vat': "BE0477472701",
+            'email': "mypartner@email.com",
+        })
+        invoice = self.env['account.move'].create({
+            'partner_id': partner.id,
+            'move_type': 'out_invoice',
+            'invoice_line_ids': [Command.create({'product_id': self.product_a.id})]
+        })
+        invoice.action_post()
+        my_invoice_raw = self.env['account.edi.xml.ubl_bis3']._export_invoice(invoice)[0]
+        my_invoice_root = etree.fromstring(my_invoice_raw)
+        modifying_xpath = """
+            <xpath expr="(//*[local-name()='PaymentMeans']/*[local-name()='PaymentID'])" position="after">
+                <PayeeFinancialAccount><ID>Test account</ID></PayeeFinancialAccount>
+            </xpath>
+            <xpath expr="(//*[local-name()='LegalMonetaryTotal']/*[local-name()='TaxExclusiveAmount'])" position="replace">
+                <TaxExclusiveAmount currencyID="EUR"><!--Some valid XML
+                comment-->1000.0</TaxExclusiveAmount>
+            </xpath>"""
+        xml_attachment = self.env['ir.attachment'].create({
+            'raw': etree.tostring(self.with_applied_xpath(my_invoice_root, modifying_xpath)),
+            'name': 'test_invoice.xml',
+        })
+        imported_invoice = self.env['account.journal']\
+                .with_context(default_journal_id=self.company_data['default_journal_purchase'].id)\
+                ._create_document_from_attachment(xml_attachment.id)
+        self.assertRecordValues(imported_invoice.invoice_line_ids, [{
+            'amount_currency': 1000.00,
+            'quantity': 1.0}])
+
     def test_import_bill_without_tax(self):
         """ Test that no tax is set (even the default one) when importing a bill without tax."""
         file_path = "bis3_bill_without_tax.xml"


clbr-odoo (commented):
LGTM thanks, can this be re-targetted to 16.0 promptly so that we merge it this week.
Did you tried to apply the security flag recommendation and pass `resolve_entities=False` to the parser instantiation ?

addons/account/models/ir_attachment.py
```diff
@@ -32,7 +32,7 @@ def _decode_edi_xml(self, filename, content):
         :returns:           A list with a dictionary.
         """
         try:
-            xml_tree = etree.fromstring(content)
+            xml_tree = etree.fromstring(content, parser=etree.XMLParser(remove_comments=True))
```
Megaaaaaa: Stripping comments at parse time fixes the immediate invoice issue, but it permanently removes information from the XML document and changes its structure for all future uses. Even if comments are not needed today, another part of the system may rely on them being preserved. I think it'd be better to keep the XML intact and make the invoice import logic robust to comments when reading numeric values, rather than deleting comments globally.

```suggestion
            xml_tree = etree.fromstring(content)
```

```py
# addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py (_get_import_document_amount_sign): read full string value to handle leading comments (value may be in comment tail), then strip before float()
amount_node = tree.find('.//{*}LegalMonetaryTotal/{*}TaxExclusiveAmount')
amount_text = (amount_node.xpath("string(.)").strip() if amount_node is not None else "")
if amount_text and float(amount_text) < 0:
    return 'refund', -1
```

This would fix the issue solely for `TaxExclusiveAmount` though. If that is not acceptable, the very minimal change you need to do to keep it safe would be to add `resolve_entities=False` to it but that should be required in this case.

```suggestion
            xml_tree = etree.fromstring(content, parser=etree.XMLParser(remove_comments=True, resolve_entities=False))
```

Megaaaaaa (changes requested):
Hello, :wave:

Thank you for the fix !

Related to the failing ci/security, I don't think you need a custom parser at all here. And actually, it's better not to have it at all. Plus you would avoid the ci/security. Let me know what you think :smile:

addons/account/models/ir_attachment.py
clbr-odoo: We'd need to do that for each numeric node which is not acceptable.
I don't think any EDI relies on comments.

Here is a real-case example: 
```xml
  <cac:TaxTotal>
    <cbc:TaxAmount currencyID="EUR">2.33</cbc:TaxAmount>
    <cac:TaxSubtotal>
      <cbc:TaxableAmount currencyID="EUR"><!--Raw text: 11.10, currency: EUR, decimals: 2, round: true, negate: false, abs: false-->11.10</cbc:TaxableAmount>
      <cbc:TaxAmount currencyID="EUR"><!--Raw text: 2.33, currency: EUR, decimals: 2, round: true, negate: false, abs: false-->2.33</cbc:TaxAmount>
      <cac:TaxCategory>
        <cbc:ID>S</cbc:ID>
        <cbc:Percent>21</cbc:Percent>
        <cac:TaxScheme>
          <cbc:ID>VAT</cbc:ID>
        </cac:TaxScheme>
      </cac:TaxCategory>
    </cac:TaxSubtotal>
  </cac:TaxTotal>
  <cac:LegalMonetaryTotal>
    <cbc:LineExtensionAmount currencyID="EUR"><!--Raw text: 11.10, currency: EUR, decimals: 2, round: true, negate: false, abs: true-->11.10</cbc:LineExtensionAmount>
    <cbc:TaxExclusiveAmount currencyID="EUR"><!--Raw text: 11.10, currency: EUR, decimals: 2, round: true, negate: false, abs: true-->11.10</cbc:TaxExclusiveAmount>
    <cbc:TaxInclusiveAmount currencyID="EUR"><!--Raw text: 13.43, currency: EUR, decimals: 2, round: true, negate: false, abs: true-->13.43</cbc:TaxInclusiveAmount>
    <cbc:PrepaidAmount currencyID="EUR"><!--Raw text: 13.43, currency: EUR, decimals: 2, round: true, negate: false, abs: true-->13.43</cbc:PrepaidAmount>
    <cbc:PayableAmount currencyID="EUR"><!--Raw text: 0.00, currency: EUR, decimals: 2, round: true, negate: false, abs: false-->0.00</cbc:PayableAmount>
  </cac:LegalMonetaryTotal>
  <cac:InvoiceLine>
    <cbc:ID>1</cbc:ID>
    <cbc:InvoicedQuantity unitCode="EA"><!--Raw text: 0.6000, currency: , decimals: 2, round: false, negate: false, abs: false-->0.6</cbc:InvoicedQuantity>
    <cbc:LineExtensionAmount currencyID="EUR"><!--Raw text: 11.10, currency: EUR, decimals: 2, round: true, negate: false, abs: false-->11.10</cbc:LineExtensionAmount>
    <cac:Item>
      <cbc:Description>anonymized</cbc:Description>
      <cbc:Name>anonymized</cbc:Name>
      <cac:SellersItemIdentification>
        <cbc:ID>anonymized</cbc:ID>
      </cac:SellersItemIdentification>
      <cac:ClassifiedTaxCategory>
        <cbc:ID>S</cbc:ID>
        <cbc:Percent>21</cbc:Percent>
        <cac:TaxScheme>
          <cbc:ID>VAT</cbc:ID>
        </cac:TaxScheme>
      </cac:ClassifiedTaxCategory>
    </cac:Item>
    <cac:Price>
      <cbc:PriceAmount currencyID="EUR"><!--Raw text: 18.5000, currency: EUR, decimals: 4, round: true, negate: false, abs: false-->18.5000</cbc:PriceAmount>
      <cbc:BaseQuantity unitCode="EA">1</cbc:BaseQuantity>
    </cac:Price>
  </cac:InvoiceLine>
  ```

Comment (Megaaaaaa) [reply addons/account/models/ir_attachment.py]: Yeah with that amount of comments, it becomes a pretty annoying change. Might be worth thinking about a helper to read numeric nodes in the future. In the meantime, I suppose the `resolve_entities=False` and an override will be enough :+1:

```suggestion
            xml_tree = etree.fromstring(content, parser=etree.XMLParser(remove_comments=True, resolve_entities=False))
```

Megaaaaaa (approved):
LGTM for security :+1:

robodoo override=ci/security

reason:
- `resolve_entities=False` blocks entity resolution, preventing XXE/data exfiltration.

I keep in mind a change at least for master to remove this and have a more robust way to read numeric nodes though (mostly for comments, not against the parser) :smile:

clbr-odoo (approved):
Thanks. Skip-merging to get following overrides quickly 🙂 
Let's try to merge this before next Monday.
@robodoo r+ fw=skipmerge

PR #247552, opened 2026-02-06
[FIX] html_editor: allow re-uploading file in link popover

rahg-odoo:
Specification:
- When a user uploads a file in the link popover, user can not re-upload another
file without manually removing previous URL.

image
image

After this commit:
- This commit allows user to re-upload a file without removing previously
generated URL first.
- TL;DR : upload button won't disappear.

task-5470111

---

### DIFF
--- a/addons/html_editor/static/tests/link/popover.test.js
+++ b/addons/html_editor/static/tests/link/popover.test.js
@@ -1533,7 +1533,7 @@ describe("link in contenteditable=false", () => {
 });
 
 describe("upload file via link popover", () => {
-    test("should display upload button when url input is empty", async () => {
+    test("should display upload button whether url input is empty or filled.", async () => {
         const { editor } = await setupEditor("<p>[]<br></p>");
         execCommand(editor, "openLinkTools");
         await waitFor(".o-we-linkpopover");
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/tests/link/popover.test.js
```diff
@@ -1533,22 +1533,6 @@ describe("link in contenteditable=false", () => {
 });
 
 describe("upload file via link popover", () => {
```
Jinjiu96, line 1535: Could you add related tests? For example: upload the same file twice and check the url (shouldn't be uploaded twice and should be the same url); also upload 2 different files and check the url changes.

Jinjiu96 (changes requested):
@rahg-odoo Helloo, thanks for the change 🙏 just nit-pick about the tests

addons/html_editor/static/tests/link/popover.test.js
rahg-odoo: Sure, thanks for the suggestion.
On it,

rahg-odoo: @Jinjiu96 

Regarding the suggested test cases:

For uploading the same file twice and checking the URL: since we are patching the upload method, this scenario will consistently return the same URL. Additionally, our current implementation already allows uploading the same file again with the same URL, so this behavior is expected.
For uploading two different files and verifying that the URL changes: a similar test case is already covered. Given that the upload method is patched, the URLs will behave as expected, and this scenario has effectively been validated.

I also explored other possible cases, but at the moment, I don’t see any additional scenarios that would require separate tests.

do you have any specific flow in mind?

Jinjiu96: @rahg-odoo Thanks for the checking and explanation 🙏 

Then maybe a simple test that checking the uploading button is still there after uploading a file? Just to include the behavior in the spec if there isn't any yet :) 

You can just modify the test you were removing and check the upload button's count is still 1 after the url isn't empty.

I will r+ if when it's done 😄

Jinjiu96 (approved):
remember to mark it ready, I will r+ it asap the runbot is green 😉

PR #247562, opened 2026-02-06
[FIX] mail: fix style of mention suggestion

seb-odoo:
Use the same style in full composer than in small composer.

Tweak style to account for small device, better handle overflows

task-5916878

Before / After (small composer)

image

image

Before / After (full)

image

image

### DIFF
--- a/addons/mail/static/src/core/common/composer.scss
+++ b/addons/mail/static/src/core/common/composer.scss
@@ -67,3 +67,7 @@
     padding-bottom: 10px;
     line-height: 1.42857143 !important; // so that input is rounded to 20px = 14px (base font) * 1.42857143 (line-height)
 }
+
+.o-mail-Composer-suggestionPartner-name.o-has-email {
+    max-width: 70%;
+}
--- a/addons/mail/static/src/core/common/composer.xml
+++ b/addons/mail/static/src/core/common/composer.xml
@@ -123,10 +123,12 @@
     <t t-name="mail.Composer.suggestionPartner">
         <t t-set="partner" t-value="option.partner"/>
         <ImStatus t-if="partner" persona="partner"/>
-        <strong class="px-2 py-1 align-self-center flex-shrink-0 text-truncate">
+        <t t-set="title" t-value="partner.name"/>
+        <t t-if="partner.email" t-set="title" t-value="`${title} (${partner.email})`"/>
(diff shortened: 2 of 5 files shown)


addons/mail/static/src/core/common/composer.scss
```diff
@@ -67,3 +67,7 @@
     padding-bottom: 10px;
     line-height: 1.42857143 !important; // so that input is rounded to 20px = 14px (base font) * 1.42857143 (line-height)
 }
+
+.o-mail-Composer-suggestionPartner-name.o-has-email {
+    max-width: 70%;
+}
```
seb-odoo, line 73: Note: give enough space to display at least part of the email so we can access its `title`

addons/mail/static/src/core/common/composer.xml
```diff
@@ -123,10 +123,10 @@
     <t t-name="mail.Composer.suggestionPartner">
         <t t-set="partner" t-value="option.partner"/>
         <ImStatus t-if="partner" persona="partner"/>
-        <strong class="px-2 py-1 align-self-center flex-shrink-0 text-truncate">
+        <strong class="o-mail-Composer-suggestionPartner-name px-2 py-1 align-self-center flex-shrink-0 text-truncate mw-100" t-att-class="{'o-has-email': partner.email}" t-att-title="partner.name">
```
seb-odoo: note: `mw-100` to prevent overflow when there is no email

addons/mail/static/src/core/web/mention_list.scss
```diff
@@ -0,0 +1,3 @@
+.o-mail-MentionList {
+    width: Min(80vw, 450px);
```
seb-odoo, line 2: Note: increase base width otherwise it is too small to display everything. Cap to 80vw to handle small devices.

addons/mail/static/src/core/common/composer.xml
```diff
@@ -123,10 +123,10 @@
...
             <t t-esc="partner.name"/>
         </strong>
-        <span t-if="partner.email" class="text-600 text-truncate align-self-center">(<t t-esc="partner.email"/>)</span>
+        <span t-if="partner.email" class="text-600 text-truncate align-self-center" t-att-title="partner.email">(<t t-esc="partner.email"/>)</span>
```
alexkuhn: What about having title of both being the same, i.e. display the name and email in full?

That would effectively make it so that we can peek the full email address even with just mouse-hover on the name part.

alexkuhn (approved):
@robodoo delegate+ r+

PR #247568, opened 2026-02-06
[FIX] web_editor, html_editor: remove checklist indent

bso-odoo:
The checklist has different indents than bullet list and numbered list. It should not be the case.

This commit removes the extra indent from checklist entries.

Steps to reproduce:
- Go to a "To do" note
- Create a checklist with indented items
- Create a bullet list with indented items

=> Both list were not aligned

task-5916723


dmo-odoo:
@robodoo r+

PR #247569, opened 2026-02-06
[FIX] mail: hydrate needaction counter on threads when fetching inbox messages

nees-odoo:
Thread unread counters should reflect the actual backend state 
as soon as Inbox messages are loaded, not only after a new bus 
notification is received.

**Steps to Reproduce:**
- Log in as User A and User B.
- Ensure User A preferences set to Handle in Odoo.
- From User B, mention User A in the chatter of a record.
- From User A, Open the messaging menu, counter for that related thread 
is correct.
- Refresh the page.
- Open the messaging menu again, the thread now shows a grey badge 
instead of the expected unread counter.

- When loading Inbox messages, related `mail.thread` records 
(such as project.task) did not include unread counters in the store payload. 
As a result, threads that already had unread messages appeared as (grey badge) 
after a refresh, leading to an incorrect counter state.

- Additionally, frontend-only counter adjustments could sometimes lead to 
inconsistent or even negative unread values due to missing initial sync.

- Threads with existing unread messages now receive the correct counter state, 
when Inbox messages are fetched.
- This also prevents inconsistent or negative counter values caused by 
frontend-only updates.

task-[5474143](https://www.odoo.com/odoo/project/1519/tasks/5474143)

Before/After (**on refresh**):
image

image

---

### DIFF
--- a/addons/mail/controllers/mailbox.py
+++ b/addons/mail/controllers/mailbox.py
@@ -11,9 +11,21 @@ class MailboxController(http.Controller):
         domain = [("needaction", "=", True)]
         res = request.env["mail.message"]._message_fetch(domain, **(fetch_params or {}))
         messages = res.pop("messages")
+        # sudo: bus.bus: reading non-sensitive last id
+        bus_last_id = request.env["bus.bus"].sudo()._bus_last_id()
+        store = Store().add(
+            messages,
+            extra_fields=[
+                Store.One("thread", [
+                    Store.Attr("message_needaction_counter"),
+                    Store.Attr("message_needaction_counter_bus_id", bus_last_id)
+                ], as_thread=True)
+            ],
+            add_followers=True
+        )
         return {
             **res,
-            "data": Store().add(messages, add_followers=True).get_result(),
+            "data": store.get_result(),
             "messages": messages.ids,
         }
 
--- a/addons/mail/static/src/core/web/mail_core_web_service.js
+++ b/addons/mail/static/src/core/web/mail_core_web_service.js
@@ -72,7 +72,8 @@ export class MailCoreWeb {
                 if (
                     thread &&
                     message.needaction &&
-                    notifId > thread.message_needaction_counter_bus_id
+                    notifId > thread.message_needaction_counter_bus_id &&
+                    thread.message_needaction_counter > 0
                 ) {
                     thread.message_needaction_counter--;
                 }
--- a/addons/test_discuss_full/tests/test_performance_inbox.py
+++ b/addons/test_discuss_full/tests/test_performance_inbox.py
@@ -78,5 +78,5 @@ class TestInboxPerformance(HttpCase, MailCommon):
                 rating_value="4",
             )
         self.authenticate(self.user_employee.login, self.user_employee.password)
-        with self.assertQueryCount(42):
+        with self.assertQueryCount(43):
             self.make_jsonrpc_request("/mail/inbox/messages")


addons/mail/models/mail_message.py
```diff
@@ -1283,6 +1283,20 @@ def _get_store_attachment_fields(self, target):
...
+                    Store.Attr("message_needaction_counter"),
+                    Store.Attr("message_needaction_counter_bus_id", bus_last_id),
+                ],
+                as_thread=True,
+            ),
+        ]
+
     def _get_store_linked_messages_fields(self):
         """Add the messages that are referenced by the current message's body to the given store.
         This method should only return message data that are not sensitive to be broadcasted to
```
maad-odoo, line 1290: Are you sure that the ID returned by `_bus_last_id()` here, while preparing the payload actually corresponds to the `mail.message/inbox` bus event we’re targeting, and not some unrelated record?

From what I understand, the method simply returns the ID of the last record in the table, rather than specifically the last `mail.message/inbox` entry.

nees-odoo: `_bus_last_id()` is intentionally used as a global synchronization marker. The value is not meant to correspond to a specific `mail.message/inbox` bus event, but to represent the latest known bus state at the time of payload generation.

The frontend only compares bus ids monotonically (**`notifId > current_bus_id`**) to ensure correct ordering. Therefore, using the last global bus id is sufficient and consistent with existing sync logic.

maad-odoo: Okay, but if we send a bus_last_id that isn’t actually related to mail.message/inbox as message_needaction_counter_bus_id, wouldn’t that essentially mean we’re sending incorrect or misleading information to the client?

If the purpose of this ID is only to perform a comparison and if we need it, then perhaps we should simply send it as bus_last_id instead of labeling it as something action-specific. Right now, on the client side, the received bus_last_id is treated as message_needaction_counter_bus_id, which gives the impression that it’s tied to that specific event. What do think ?

nees-odoo: If we rename the field, the frontend will no longer recognize `message_needaction_counter_bus_id`, and it will default to `0`. The counter will still look correct immediately after refresh (since it’s _hydrated_ from the **backend**), but the sync boundary is lost.

On the next bus event, the client does:

```javascript
if (message.thread && notifId > message.thread.message_needaction_counter_bus_id) {
    message.thread.message_needaction_counter++;
}
``` 
Since `message_needaction_counter_bus_id` would default to 0 as the backend sends diff labeling, the condition becomes true for every new notification, potentially leading to incorrect increments.

The field is not tied to a specific event, but serves as a global sync marker for that counter. Keeping the existing naming preserves the backend–frontend contract and ensures correct update ordering.

addons/mail/controllers/mailbox.py
```diff
@@ -13,7 +13,7 @@ def discuss_inbox_messages(self, fetch_params=None):
         messages = res.pop("messages")
         return {
             **res,
-            "data": Store().add(messages, add_followers=True).get_result(),
+            "data": Store().add(messages, extra_fields=request.env["mail.message"]._get_store_inbox_fields(), add_followers=True).get_result(),
```
alexkuhn: Could just `.add(messages)` and `.add(thread)` (with needaction fields and `as_thread`) separately.
This is easier to read than indirection through `_get_store_inbox_fields` method.

alexkuhn (approved):
@robodoo delegate+ r+

PR #247574, opened 2026-02-06
[IMP] pos_restaurant: correct changes count

vlst-odoo:
In pos_restaurant, when ordering, we show the number of items in each
    category that will be sent to preparation. For ex, if we add one pizza
    and one drink to the order, we expect to see something of the type
    `Food: 1, Drinks: 1` ( depending on the way the user configured their
    pos ).
    
    When using `combos`, the combo item itself, which is not a physical
    product, was counted as an item to be sent to preparation.
    
    Ex:
    1. Create a combo consisting of a pizza and a drink.
    2. Place the combo in the `food` category
    3. Add the combo to the order
    4. Observe that the pos shows `Food: 2. Drinks: 1`
    
    In this commit we prevent the pos from counting combos as products to be
    sent to preparation.
    
    In the same spirit, on the corner of each table in the floor plan we see
    a number indicating the number of items that were added on that table.
    This number also included combos in the count, which, as described, is
    wrong. We address the issue in this commit.
    
    Task: 5906779

---

### DIFF
--- a/addons/point_of_sale/static/src/app/models/utils/order_change.js
+++ b/addons/point_of_sale/static/src/app/models/utils/order_change.js
@@ -162,7 +162,9 @@ export const getOrderChanges = (order, orderPreparationCategories) => {
                 // if note update with qty add
                 changes[lineKey] = lineDetails;
                 changesCount += quantityDiff;
-                changeAbsCount += Math.abs(quantityDiff);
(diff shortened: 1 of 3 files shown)


davidmonnom (approved):
@robodoo r+

PR #247578, opened 2026-02-06
Master odoo list rar

rrahir:

Current behavior before PR:

---


addons/spreadsheet/static/src/list/list_data_source.js
```diff
@@ -84,13 +110,18 @@ export class ListDataSource extends OdooViewsDataSource {
 
     async _load() {
         await super._load();
+        // TODORAR not sure it's worth deleting valid values... the field path cannot change nature,
```
rrahir: @pro-odoo  wdyt?

addons/spreadsheet/models/spreadsheet_mixin.py
```diff
@@ -135,12 +135,13 @@ def _empty_spreadsheet_data(self):
...
                     "id": "sheet1",
                     "name": _("Sheet1"),
-                }
+                },
             ],
             "settings": {
                 "locale": locale,
             },
             "revisionId": "START_REVISION",
+            "version": "19.1.0",
```
pro-odoo, line 137: Why is it necessary ?

addons/spreadsheet/static/src/helpers/odoo_functions_helpers.js
```diff
@@ -1,5 +1,13 @@
 // @ts-check
 
+/**
+ * @typedef {import("@odoo/o-spreadsheet").AST} AST
+ *
+ * @typedef {Object} OdooFunctionDescription
+ * @property {string} functionName Name of the function
+ * @property {Array<AST>} args Arguments of the function
+ */
+
```
pro-odoo, line 10: Seems useless

pro-odoo (commented):
scroll scroll review, will have a deeper look after testing

addons/spreadsheet/static/src/helpers/odoo_functions_helpers.js
rrahir: this was removed on the past when we moved the pivot code to o-spreadsheet and the typing wans 'undefined' which bothered me, we can do without

addons/spreadsheet/models/spreadsheet_mixin.py
rrahir: Without version, no migration. It had an impact in spreadsheet_sale_management in which we add a list to the definition. I feel it's better to have a migrable version of the data rather than something that is assumed to be in the latest version. The current situation is probably problematic already  because in the case of 'version history', we assume all commands and data have been migrated to the latest version. If your data has no version, it's not migrated, therefore can have an unexpected structure during import.

addons/spreadsheet/static/src/list/list_functions.js
```diff
@@ -48,11 +48,27 @@ const ODOO_LIST_HEADER = {
...
-        const translatedDisplayName = this.getters.getListHeaderValue(id, field);
-        return displayName || translatedDisplayName;
+        return displayName || this.getters.getListHeaderValue(id, field);
     },
     returns: ["NUMBER", "STRING"],
 };
 
-functionRegistry.add("ODOO.LIST", ODOO_LIST);
+const ODOO_LIST = {
+    description: _t("Get a spillable odoo list function."),
```
LucasLefevre:
```suggestion
    description: _t("Get a dynamic odoo list function."),
```
?

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,31 +20,129 @@ export class ListUIPlugin extends OdooUIPlugin {
...
+                this.dispatch("DUPLICATE_ODOO_LIST", {
+                    listId: cmd.listId,
+                    newListId: cmd.newListId,
+                    duplicatedListName,
+                });
+
+                const sheetIdFrom = this.getters.getActiveSheetId();
+                const sheetId = uuidGenerator.smallUuid();
+
+                console.log(
```
LucasLefevre: console.log 👀

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,31 +20,129 @@ export class ListUIPlugin extends OdooUIPlugin {
...
+                    this.dispatch("CREATE_SHEET", {
+                        sheetId,
+                        position: this.getters.getSheetIds().length,
+                        name: this.getters.getSheetIdByName(duplicatedListName)
+                            ? undefined
+                            : duplicatedListName,
+                    })
+                );
+                this.dispatch("ACTIVATE_SHEET", { sheetIdFrom, sheetIdTo: sheetId });
+                // sheetId, col, row,   listId, linesNumber, columns, mode
```
LucasLefevre: ?

addons/spreadsheet/static/src/list/list_data_source.js
```diff
@@ -84,13 +118,16 @@ export class ListDataSource extends OdooViewsDataSource {
 
     async _load() {
         await super._load();
+        this.fieldPathsToFieldMap = {};
+        const { domain, orderBy, context } = this._searchParams;
+        this._recordsCount = await this._orm.searchCount(this._metaData.resModel, domain, {
```
LucasLefevre: like in odoo list views, it should have a limit of 10k to avoid scanning the entire table of 14M records for no reason (most of the time).
Display 10000+ and clickable to get the real number

addons/spreadsheet/static/src/@types/commands.d.ts
```diff
@@ -145,6 +154,27 @@ declare module "@spreadsheet" {
...
+        listId: string;
+        definition: OdooListCoreDefinition;
+        sheetId: string;
+        col: number;
+        row: number;
+        linesNumber: number;
+        mode: "static" | "dynamic";
+    }
+
+    export interface reInsertListCommand {
```
LucasLefevre:
```suggestion
    export interface ReInsertListCommand {
```

addons/spreadsheet/static/src/list/plugins/list_core_view_plugin.js
```diff
@@ -383,4 +403,34 @@ export class ListCoreViewPlugin extends OdooCoreViewPlugin {
             !this.getters.isDataSourceLinkedToChart("list", listId)
         );
     }
+
+    getListValuesAndFormats(listId, rowCount = undefined) {
+        const dataSource = this.getters.getListDataSource(listId);
+
+        if (rowCount === undefined) {
+            rowCount = dataSource.getRecordsCount();
```
LucasLefevre: We should spare this RPC here. If no limit is specified, just leave the limit undefined and avoid a `search_count`.

Also it looks like the `search_count` is done twice

Image

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,31 +20,129 @@ export class ListUIPlugin extends OdooUIPlugin {
...
+                    _t("%(list_name)s (List #%(list_id)s)", {
+                        list_name: cmd.name,
+                        list_id: cmd.listId,
+                    })
+                );
+                if (cmd.insertInNewSheet) {
+                    const sheetIdFrom = this.getters.getActiveSheetId();
+                    const sheetId = uuidGenerator.smallUuid();
+                    if (this.getters.getSheetIdByName(sheetName)) {
+                        sheetName = undefined;
```
LucasLefevre: Shouldn't you use `getters.getNextSheetName` to generate the sheet name here?

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,31 +20,129 @@ export class ListUIPlugin extends OdooUIPlugin {
...
+                });
+
+                const sheetIdFrom = this.getters.getActiveSheetId();
+                const sheetId = uuidGenerator.smallUuid();
+
+                console.log(
+                    this.dispatch("CREATE_SHEET", {
+                        sheetId,
+                        position: this.getters.getSheetIds().length,
+                        name: this.getters.getSheetIdByName(duplicatedListName)
```
LucasLefevre: same here -> `getNextSheetName` ?

addons/spreadsheet/static/src/list/list_data_source.js
```diff
@@ -36,17 +36,18 @@ export class ListDataSource extends OdooViewsDataSource {
...
      * @param {ListSearchParams} params.searchParams
-     * @param {number} params.limit
      */
     constructor(services, params) {
         super(services, params);
         this.maxPosition = 0;
         this.maxPositionFetched = 0;
         this.data = [];
-        this.fieldPathsToFetch = new Set(["id"]);
+        this.fieldPathsToFetch = new Set(params.columns.map((col) => col.name).concat(["id"]));
```
LucasLefevre: I don't think we want that. If only `ODOO.LIST.VALUE` formulas are used and only with some fields, we don't want to fetch all fields for nothing

addons/spreadsheet/static/src/list/list_functions.js
```diff
@@ -48,11 +48,27 @@ const ODOO_LIST_HEADER = {
         const field = toString(fieldName);
         assertListsExists(id, this.getters);
         const displayName = toString(fieldDisplayName);
-        const translatedDisplayName = this.getters.getListHeaderValue(id, field);
-        return displayName || translatedDisplayName;
+        return displayName || this.getters.getListHeaderValue(id, field);
+    },
+    returns: ["NUMBER", "STRING"],
```
LucasLefevre: `returns` has been removed a long long time ago 😄

addons/spreadsheet/static/tests/helpers/data.js
```diff
@@ -91,29 +103,25 @@ export function getBasicServerData() {
  */
 export function generateListDefinition(model, columns, actionXmlId, orderBy = []) {
     const cols = [];
-    for (const name of columns) {
-        const fieldName = name.split(".")[0]; // in case of property field (eg. partner_properties.my_property)
-        const PyModel = Object.values(SpreadsheetModels).find((m) => m._name === model);
+    for (let { name, string } of columns) {
+        if (!string) {
+            // const fieldName = name.split(".")[0]; // in case of property field (eg. partner_properties.my_property)
```
LucasLefevre: 👀

addons/spreadsheet/static/src/o_spreadsheet/migration.js
```diff
@@ -208,6 +208,16 @@ migrationStepRegistry.add("19.1.2", {
     },
 });
 
+migrationStepRegistry.add("19.2.1", {
+    migrate(data) {
+        for (const list of Object.values(data.lists || {})) {
+            list.columns = list.columns?.map((col) => ({ name: col, string: col })) || [];
```
LucasLefevre, line 214: The technical field name as the display string isn't very top 🤔 
Should it be optional ?

addons/spreadsheet/static/src/o_spreadsheet/migration.js
```diff
@@ -275,12 +285,30 @@ function renameFunctions(data, map) {
     return data;
 }
 
+function renameFunctions(data, map) {
+    for (const sheet of data.sheets || []) {
+        for (const xc in sheet.cells || []) {
+            const cell = sheet.cells[xc];
+            if (cell?.startsWith("=")) {
```
LucasLefevre: this crashes with squished cells

addons/spreadsheet/static/tests/global_filters/global_filters_model.test.js
```diff
@@ -435,7 +438,7 @@ test("Can import/export filters", async function () {
         lists: {
             1: {
                 id: 1,
-                columns: ["foo", "contact_name"],
+                columns: [{ name: "foo" }, { name: "contact_name" }],
```
LucasLefevre: ah, ici it looks like it's optional

addons/spreadsheet/static/src/list/list_functions.js
rrahir: well it wasn't adapted :p

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
rrahir: there is a fallback to the command handler that will fill up sheetName already. I can try to do the same as pivots though

LucasLefevre: The fallback was left for backward compatibility, but it shouldn't be used. We should document it or even trigger a warning when used

addons/spreadsheet/static/src/list/plugins/list_core_plugin.js
```diff
@@ -78,34 +88,34 @@ export class ListCorePlugin extends OdooCorePlugin {
...
                 const { listId, newListId, duplicatedListName } = cmd;
-                const duplicatedList = deepCopy(this.lists[listId].definition);
+                const duplicatedList = deepCopy(this.lists[listId]);
                 duplicatedList.name = duplicatedListName ?? duplicatedList.name + " (copy)";
                 this._addList(newListId, duplicatedList);
                 break;
             }
             case "RE_INSERT_ODOO_LIST": {
-                const { sheetId, col, row, id, linesNumber, columns } = cmd;
+                const { sheetId, col, row, listId, linesNumber, columns, mode } = cmd;
```
anhe-odoo, line 112: `columns` isn't in the interface for RE_INSERT_ODOO_LIST, but here you take `columns` from `cmd`. Am I missing something ?

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,34 +20,132 @@ export class ListUIPlugin extends OdooUIPlugin {
...
+                const sheetIdFrom = this.getters.getActiveSheetId();
+                const sheetId = uuidGenerator.smallUuid();
+
+                this.dispatch("CREATE_SHEET", {
+                    sheetId,
+                    position: this.getters.getSheetIds().length,
+                    name: this._computeValidSheetName(duplicatedListName),
+                });
+                this.dispatch("ACTIVATE_SHEET", { sheetIdFrom, sheetIdTo: sheetId });
+                const columns = this.getters.getListDefinition(cmd.listId).columns;
```
rrahir:
```suggestion
                const columns = this.getters.getListDefinition(cmd.newListId).columns;
```
if the columns actually make sense in. the command to begin with, then it should target the new definition ,not the old one ...

rrahir: Need to determine if columns really makes sense in the command

addons/spreadsheet/static/src/list/list_data_source.js
rrahir: i'll split this in 2 sets, one for the path definition and the other with the actual fields to fetch

addons/spreadsheet/static/src/list/list_actions.js
```diff
@@ -13,30 +13,36 @@ const { isMatrix } = helpers;
  * @returns {Promise<void>}
  */
 export const SEE_RECORD_LIST = async (position, env, newWindow) => {
+    // TODORAR make it work with dynamic lists as well
```
anhe-odoo: Still needed in this PR, or is it for another task ?

addons/spreadsheet/static/src/@types/commands.d.ts
```diff
@@ -155,7 +186,9 @@ declare module "@spreadsheet" {
         | AddGlobalFilterCommand
         | EditGlobalFilterCommand
         | RemoveGlobalFilterCommand
-        | MoveGlobalFilterCommand;
+        | MoveGlobalFilterCommand
+        | InsertListCommand
+        | reInsertListCommand;
```
LucasLefevre: `ReinsertListCommand` above vs `reInsertListCommand` here

addons/spreadsheet/static/src/list/plugins/list_core_plugin.js
```diff
@@ -209,10 +198,7 @@ export class ListCorePlugin extends OdooCorePlugin {
 
     _addList(id, definition) {
         const lists = { ...this.lists };
-        lists[id] = {
-            id,
-            definition,
-        };
+        lists[id] = definition;
         this.history.update("lists", lists);
```
LucasLefevre: can be simplified to this while we are at it
`this.history.update("lists", id, definition);`

addons/spreadsheet/static/src/list/plugins/list_core_view_plugin.js
```diff
@@ -383,4 +428,42 @@ export class ListCoreViewPlugin extends OdooCoreViewPlugin {
             !this.getters.isDataSourceLinkedToChart("list", listId)
         );
     }
+
+    getListValuesAndFormats(listId, rowCount) {
```
LucasLefevre, line 432: somehow, there are 2 `web_search_read` done. One fetching only the ids, the second one with all the required fields

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
LucasLefevre: up

addons/spreadsheet/static/src/list/list_functions.js
```diff
@@ -48,11 +49,24 @@ const ODOO_LIST_HEADER = {
...
+const ODOO_LIST = {
+    description: _t("Get a dynamic Odoo list function."),
+    args: [
+        arg("list_id (string)", _t("ID of the list.")),
+        arg(`row_count (number, default=${MAX_LIMIT})`, _t("number of rows to display")),
+    ],
+    category: "Odoo",
+    compute: function (listId, rowCount = { value: MAX_LIMIT }) {
+        const id = toString(listId);
+        assertListsExists(id, this.getters);
```
LucasLefevre: It should result in an error if `rowCount` is not specified and the result is larger than `MAX_LIMIT`.
To let the user know he's missing some data

addons/spreadsheet/static/src/o_spreadsheet/migration.js
LucasLefevre, line 214: `string` must also be added to `def _dispatch_insert_list_revision`

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
```diff
@@ -16,34 +20,132 @@ export class ListUIPlugin extends OdooUIPlugin {
...
         this.dispatch("CREATE_TABLE", {
-            tableType: "static",
+            tableType: mode,
             sheetId,
             ranges: [this.getters.getRangeDataFromZone(sheetId, zone)],
             config: { ...PIVOT_STATIC_TABLE_CONFIG, firstColumn: false },
         });
     }
+
+    _computeValidSheetName(name) {
```
LucasLefevre, line 153: `getters.getNextSheetName` instead ?
There's a slight functional difference (` (2)` postfix instead of just `2`, but do we care ?
And `getNextSheetName` should use `sanitizeSheetName`

addons/spreadsheet/static/src/list/list_actions.js
```diff
@@ -64,13 +69,25 @@ export const SEE_RECORD_LIST_VISIBLE = (position, getters) => {
...
+    const { functionName } = getFirstListFunction(cell.compiledFormula, getters);
+    if (functionName === "ODOO.LIST") {
+        const mainPosition = getters.getCellPosition(cell.id);
+        const rowOffset = position.row - mainPosition.row;
+        return rowOffset >= 1;
+    } else if (functionName === "ODOO.LIST.VALUE") {
+        return true;
+    } else {
+        return false;
+    }
```
LucasLefevre:
```suggestion
    } else {
        return functionName === "ODOO.LIST.VALUE";
    }
```

addons/spreadsheet/static/src/list/list_data_source.js
```diff
@@ -29,24 +31,29 @@ const { DEFAULT_LOCALE } = spreadsheet.constants;
...
      * @param {ListSearchParams} params.searchParams
-     * @param {number} params.limit
      */
     constructor(services, params) {
         super(services, params);
         this.maxPosition = 0;
         this.maxPositionFetched = 0;
         this.data = [];
         this.fieldPathsToFetch = new Set(["id"]);
+        this.fieldPathDefinitionsToFetch = new Set(params.columns.map((col) => col.name));
```
LucasLefevre, line 49: is `fieldPathDefinitionsToFetch` needed ?
AFAICT, the fields are added by `getListCellValue` in `fieldPathsToFetch` already

addons/spreadsheet/static/src/list/plugins/list_core_view_plugin.js
```diff
@@ -383,4 +428,42 @@ export class ListCoreViewPlugin extends OdooCoreViewPlugin {
...
+        const dataSource = this.getListDataSource(listId);
+        const definition = this.getters.getListDefinition(listId);
+        const columns = definition.columns.filter((col) => !col.hidden);
+
+        if (columns.length === 0) {
+            return { value: this.getters.getListDisplayName(listId) };
+        }
+
+        // triggers the fetch of the list values up to `rowCount` to fill the datasource cache (if not already done)
+        if (columns.length > 0) {
```
LucasLefevre: `columns.length > 0` is useless, we know there's at least one column thanks to `if (columns.length === 0)` above

addons/spreadsheet/static/src/list/list_data_source.js
```diff
@@ -70,6 +103,21 @@ export class ListDataSource extends OdooViewsDataSource {
         }
     }
 
+    async getRecordsCount({ fullCount } = { fullCount: false }) {
```
LucasLefevre, line 105: when opening the side panel, there are 2 identical `search_count` done

addons/spreadsheet/static/src/list/list_data_source.js
rrahir: which means it's not fetched until a formula is added in the spreadsheet : pretty much the same issue with the pivots:  if we don"t evaluate it, it's not updated  BUT we need the data to display something relevant in the sidepanel

addons/spreadsheet/static/src/list/plugins/list_ui_plugin.js
rrahir: as we discussed, `getters.getNextSheetName` will always append a number so it doesn't fit the need. Probably why we came up with uniqueText in the first place

addons/spreadsheet/__manifest__.py
```diff
@@ -75,7 +75,8 @@
             'spreadsheet/static/src/o_spreadsheet/o_spreadsheet_extended.scss',
             'spreadsheet/static/src/o_spreadsheet/migration.js',
             'spreadsheet/static/src/helpers/odoo_functions_helpers.js',
-            'spreadsheet/static/src/pivot/pivot_helpers.js',
+            # 'spreadsheet/static/src/data_sources/data_source.js',
+            # 'spreadsheet/static/src/pivot/pivot_helpers.js',
```
LucasLefevre: 👀

addons/spreadsheet/static/src/list/list_functions.js
```diff
@@ -48,11 +49,39 @@ const ODOO_LIST_HEADER = {
...
+        const _rowCount = rowCount ? toNumber(rowCount, this.locale) : undefined;
+        if (_rowCount !== undefined && _rowCount <= 0) {
+            return new EvaluationError(
+                _t("The number of rows parameter should be a positive number.")
+            );
+        }
+        const result = this.getters.getListValuesAndFormats(id, _rowCount ?? MAX_LIMIT);
+        if (result[0]?.length > MAX_LIMIT && _rowCount) {
```
LucasLefevre:
```suggestion
        if (result[0]?.length > MAX_LIMIT && _rowCount === undefined) {
```

LucasLefevre (commented):
👌 

small condition to fix
We just need to wait for the next release of o-spreadsheet now...

pro-odoo:
robodoo r+

PR #191511, opened 2024-12-23
[IMP] mail, web, *: `useService()` with auto-`useState()`

alexkuhn:
Before this commit, when a component uses service "mail.store", it has to wrap in a `useState()` to make sure any change to store is observed. This is the behaviour that essentially all `useService("mail.store")` want. If they omit it, it can fail to re-render store changed, which is painful to debug.

This commit improves by automatically wrapping `useState()` on services that are stateful. Stateful services return a reactive. Using stateful services becomes the same as non-stateful one: just put `useService(serviceName)`.

https://github.com/odoo/enterprise/pull/76058

PR #247581, opened 2026-02-06
[FIX] mail: properly sort mention suggestions in the full composer

seb-odoo:
Fetch/sort suggestion in the full composer don't receive the thread param which leads to follower not being sorted at the top.

task-5917226

### DIFF
--- a/addons/mail/static/src/core/web/wysiwyg.js
+++ b/addons/mail/static/src/core/web/wysiwyg.js
@@ -1,10 +1,12 @@
 /** @odoo-module **/
+import { useState } from "@odoo/owl";
 
 import { renderToElement } from "@web/core/utils/render";
 import { Wysiwyg } from "@web_editor/js/wysiwyg/wysiwyg";
 import { closestBlock, setCursorEnd } from "@web_editor/js/editor/odoo-editor/src/OdooEditor";
 import { patch } from "@web/core/utils/patch";
 import { MentionList } from "@mail/core/web/mention_list";
+import { useService } from "@web/core/utils/hooks";
 import { url } from "@web/core/utils/urls";
 import { usePopover } from "@web/core/popover/popover_hook";
 import { isEventHandled } from "@web/core/utils/misc";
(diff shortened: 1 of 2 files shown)


addons/mail/static/src/core/web/wysiwyg.js
```diff
@@ -18,10 +20,14 @@ patch(Wysiwyg.prototype, {
                 onClose: () => this.focus(),
             });
             this.triggerMentionList = this.triggerMentionList.bind(this);
+            this.store = useState(useService("mail.store"));
```
alexkuhn: I don't know the whole scope of wysiwyg overrides, but to play safe we can name it `mailStore`, like what we do in other non-discuss big components like `FormController` and so on.

addons/mail/static/src/core/web/wysiwyg.js
```diff
@@ -18,10 +20,14 @@ patch(Wysiwyg.prototype, {
...
         }
     },
     get inDiscuss() {
-        return this.props.options.recordInfo?.res_model === "mail.compose.message";
+        return (
+            this.props.options.recordInfo?.res_model === "mail.compose.message" &&
+            this.env.services["mail.store"]
```
alexkuhn, line 29:
```suggestion
            this.mailStore
```

alexkuhn (approved):
@robodoo delegate+

addons/mail/static/src/core/web/wysiwyg.js
seb-odoo: This crashes in some tests if I use that, the goal is to not call `useService` if it's not there.

addons/mail/static/src/core/web/wysiwyg.js
seb-odoo: Ok to rename

addons/mail/static/src/core/web/wysiwyg.js
alexkuhn: Ah ok, so ignore this comment

alexkuhn: Oh yeah I'm stupid, this is the condition to useService() 🙈

addons/mail/static/src/core/web/wysiwyg.js
```diff
@@ -18,10 +20,14 @@ patch(Wysiwyg.prototype, {
                 onClose: () => this.focus(),
             });
             this.triggerMentionList = this.triggerMentionList.bind(this);
+            this.mailStore = useState(useService("mail.store"));
```
alexkuhn, line 23: Reminder could remove `useState` in 18.1: https://github.com/odoo/odoo/pull/191511

alexkuhn (approved):
@robodoo r+

PR #247582, opened 2026-02-06
[FIX] html_editor: show toolbar when selecting list text with icons

walidsahli:
Problem:
When triple-clicking on text inside a list item that contains an icon, the toolbar is not shown.

Cause:
The toolbar is not displayed when the selection anchor node is inside an element with `contenteditable="false"`.
This happens in cases like:

`<li><span class="…" contenteditable="false"></span>abc</li>`

When triple-clicking on "abc", the selection anchorNode becomes the `span` and the offset is `0` and offset is "abc" offset 3. In that case, `selection.toString()` returns an empty string.

While this may also be a browser related issue as the selection includes text but `toString()` still returns empty string, the root cause here is that normalization does not wrap the icon with `feff` because the corresponding selector is missing.

Solution:
Include icons under `li` elements when wrapping nodes with `feff`, so the selection is normalized correctly and the toolbar can be shown.

Steps to reproduce:
- Open Email Marketing.
- Drop the "Comparison" snippet.
- Triple-click on one of the texts with check icons.
- Observe that the toolbar is not shown.

task-5913202

---

### DIFF
--- a/addons/html_editor/static/src/main/media/media_plugin.js
+++ b/addons/html_editor/static/src/main/media/media_plugin.js
@@ -103,7 +103,7 @@ export class MediaPlugin extends Plugin {
         selectors_for_feff_providers: () =>
             `:is(${paragraphRelatedElementsSelector}, ${FORMATTABLE_TAGS.join(
                 ", "
-            )}, A) > :is(${ICON_SELECTOR})`,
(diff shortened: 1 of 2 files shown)


walidsahli:
In 19.0+ i need to cherry-pick [this commit](https://github.com/odoo/odoo/commit/859f14b20341aea967359167a8d62a9d746c60ce)

PR #189049, opened 2024-11-28
[IMP] website_sale: add product information as subject in contactus

odoo-pda:
Use the contactus url defined in the settings and add a subject
parameter containing the information of the product such as the
name and the variants chosen.
The form in the contact page is then pre-filled with product info
for the recipient of the email.

task-4331152

PR #247587, opened 2026-02-06
[FIX] website_sale: Fix the urls redirection for anchors

syna-odoo:
Issue:
-------
When a zero price product is created and the contact us button on the product page is intended to redirect to some snippets created through drag and drop then the button doesn't work as intended meaning it doesn't redirects to the desired snippet even after putting the correct anchor.
for ex: `#snippet-anchor` in the `Button URL` field in the settings.

Cause:
--------
This works fine for the pages having '/contactus' or '/'. Issues raise only when we try to redirect to a snippet. Now, if the we try to redirect to any snippet on click of the button(Contact Us) by placing the corresponding anchor, it will not redirect/work as intended. This is because of the appending`?subject=product_name` that took place.

Solution:
------------
To concatenate the `subject=product_name` conditionally if the url has '#' in it If yes, we just use the `url` in the URL so that it redirects as intended else concatenate the subject & so on.
This is because for redirecting to snippets we use anchors such as '#Let's-Connect'. So, In an anchor the '#' will definitely reside.

Steps to reproduce:
-----------------------
1. Create a db in version 18.3 with website_sale installed.
2. Enable the `Prevent Sale of Zero Priced Product` checkbox in the settings.
3. Create a zero price product and few snippets under it and copy the anchor of one of the snippets to redirect when clicked on the 'Contact Us' button.
4. Use the Anchor(for ex: '#Let's-Connect') in the 'Button URL' field of settings.
5. Navigate to the created product and click on the 'Contact Us' button. Nothing happens & no intended redirection to the desired snippet.

Ref PR:
----------
https://github.com/odoo/odoo/pull/189049/changes#diff-39e02d03a8b765b4e3afc68627aeb33f11b587163638fedfb92ed5657c3336e7R398-R399

Attachments:
-----------------
**Before Fix:**
[vokoscreenNG-2026-02-06_17-36-37.webm](https://github.com/user-attachments/assets/a09101d4-13df-415d-a902-420a28aedef0)

**After Fix**:
[vokoscreenNG-2026-02-06_17-38-37.webm](https://github.com/user-attachments/assets/a6256d0f-d8cb-4146-b95e-33452a0a79c5)

- OPW - [5494517](https://www.odoo.com/odoo/project/70/tasks/5494517)

---

### DIFF
--- a/addons/website_sale/static/src/js/sale_variant_mixin.js
+++ b/addons/website_sale/static/src/js/sale_variant_mixin.js
@@ -397,8 +397,9 @@ var VariantMixin = {
             contactUsButton.removeClass('d-flex').addClass('d-none');
             product_unavailable.removeClass('d-flex').addClass('d-none');
         }
-        const url = contactUsButton.find('a').attr('data-url');
-        contactUsButton.find('a').attr('href', `${url}?subject=${combination.display_name}`);
+        const url = new URL(contactUsButton.find('a').attr('data-url'), window.location.origin);
+        url.searchParams.set('subject', combination.display_name);
+        contactUsButton.find('a').attr('href', url.toString());
 
         const self = this;
         const $price = $parent.find(".oe_price:first .oe_currency_value");


addons/website_sale/static/src/js/sale_variant_mixin.js
```diff
@@ -398,7 +398,11 @@ var VariantMixin = {
             product_unavailable.removeClass('d-flex').addClass('d-none');
         }
         const url = contactUsButton.find('a').attr('data-url');
-        contactUsButton.find('a').attr('href', `${url}?subject=${combination.display_name}`);
+        if (url.includes("#")){
+            contactUsButton.find('a').attr('href', `${url}`);
+        } else {
+            contactUsButton.find('a').attr('href', `${url}?subject=${combination.display_name}`);
+        }
```
ltinel: I don't think we should get rid of the `subject` query param if there's a hash in the URL. I guess people may still want to have it, even in snippets. Instead, I think we simply need to properly build the URL. Something like:
```suggestion
        const url = new URL(contactUsButton.find('a').attr('data-url'), window.location.origin);
        url.searchParams.set('subject', combination.display_name);
        contactUsButton.find('a').attr('href', url.toString());
```

ltinel (changes requested):
Thanks for the fix :)

addons/website_sale/static/src/js/sale_variant_mixin.js
syna-odoo: Hello @ltinel 
Thanks for your valuable feedback and suggestion.
I've applied the changes. Please do have a look :)

ltinel (approved):
Thanks for the fix 🙂 LGTM
@robodoo r+

PR #247589, opened 2026-02-06
[FIX] im_livechat, *: fix traceback and displayName bug in chatbot

aych-odoo:
*=test_discuss_full

Before this PR:
- When starting a session with a chatbot, the name of conversation is not visible on livechat channel. Because the name is computed from agent/customer history and not from bot history.
- Using a chatbot while the website editor is active could trigger a TypeError traceback.

This PR:
- computes the name properly from bot history.
- adds a guard that closes the transient channel

Steps to reproduce traceback:
 - go to localhost:8069/@/contactus
 - start chatbot and select some answer
 - receive traceback

task-[5881970](https://www.odoo.com/odoo/project/1519/tasks/5881970)

### DIFF
--- a/addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
+++ b/addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
@@ -14,6 +14,9 @@ const discussChannelPatch = {
         this.livechat_agent_history_ids = fields.Many("im_livechat.channel.member.history", {
             inverse: "channelAsAgentHistory",
         });
+        this.livechat_bot_history_ids = fields.Many("im_livechat.channel.member.history", {
+            inverse: "channelAsBotHistory",
+        });
         this.livechat_channel_id = fields.One("im_livechat.channel", { inverse: "channel_ids" });
         this.livechat_channel_member_history_ids = fields.Many(
             "im_livechat.channel.member.history",
@@ -86,10 +89,12 @@ const discussChannelPatch = {
             })
             .map((m) => m.name);
         if (!memberNames.length) {
-            const histories =
-                selfMemberType === "visitor"
+            let histories = this.livechat_customer_history_ids;
+            if (selfMemberType === "visitor") {
+                histories = this.livechat_agent_history_ids.length
                     ? this.livechat_agent_history_ids
-                    : this.livechat_customer_history_ids;
+                    : this.livechat_bot_history_ids;
+            }
             memberNames = histories
                 .map((h) => this.getPersonaName(h.partner_id || h.guest_id))
                 .filter(Boolean);
--- a/addons/im_livechat/static/src/embed/common/livechat_service.js
+++ b/addons/im_livechat/static/src/embed/common/livechat_service.js
@@ -84,7 +84,7 @@ export class LivechatService {
             // flickering, we do not want another load that would result in the
             // same issue.
             savedChannel.scrollUnread = false;
-            temporaryThread.channel.chatWindow?.close();
+            temporaryThread.channel?.chatWindow?.close();
             savedChannel.openChatWindow({ focus: true });
         });
         return savedChannel;
--- a/addons/im_livechat/static/src/core/common/@types/models.d.ts
+++ b/addons/im_livechat/static/src/core/common/@types/models.d.ts
@@ -32,6 +32,7 @@ declare module "models" {
     export interface DiscussChannel {
         chatbot: Chatbot;
         livechat_agent_history_ids: LivechatChannelMemberHistory[];
+        livechat_bot_history_ids: LivechatChannelMemberHistory[];
         livechat_channel_id: LivechatChannel;
         livechat_channel_member_history_ids: LivechatChannelMemberHistory[];
         livechat_customer_history_ids: LivechatChannelMemberHistory[];
--- a/addons/im_livechat/static/src/core/common/livechat_channel_member_history_model.js
+++ b/addons/im_livechat/static/src/core/common/livechat_channel_member_history_model.js
@@ -13,6 +13,15 @@ export class LivechatChannelMemberHistory extends Record {
             return false;
         },
     });
+    channelAsBotHistory = fields.One("discuss.channel", {
+        inverse: "livechat_bot_history_ids",
+        compute() {
+            if (this.livechat_member_type === "bot") {
+                return this.channel_id;
+            }
+            return false;
+        },
+    });
     channelAsCustomerHistory = fields.One("discuss.channel", {
         inverse: "livechat_customer_history_ids",
         compute() {
--- a/addons/test_discuss_full/static/tests/tours/chatbot_redirect_to_portal_tour.js
+++ b/addons/test_discuss_full/static/tests/tours/chatbot_redirect_to_portal_tour.js
@@ -8,6 +8,9 @@ registry.category("web_tour.tours").add("chatbot_redirect_to_portal", {
             run: "click",
         },
         {
+            trigger: ".o-livechat-root:shadow .o-mail-ChatWindow-header:contains(Redirection Bot)",
+        },
+        {
             trigger:
                 ".o-livechat-root:shadow .o-mail-Message:contains(Hello, were do you want to go?)",
             run: "click",


maad-odoo (commented):
Can we please have a test case covering this case ?

addons/im_livechat/static/src/core/common/discuss_channel_model_patch.js
```diff
@@ -72,10 +75,13 @@ const discussChannelPatch = {
...
             .map((m) => m.name);
         if (!memberNames.length) {
-            const histories =
+            let histories =
                 selfMemberType === "visitor"
                     ? this.livechat_agent_history_ids
                     : this.livechat_customer_history_ids;
+            if (!histories.length) {
+                histories = this.livechat_bot_history_ids;
+            }
```
tsm-odoo: I think it would be more explicit written like this. This way, we keep the clear link between member type and histories chosen from the display name.

```js
            let histories = this.livechat_customer_history_ids;
            if (selfMemberType === "visitor") {
                histories = this.livechat_agent_history_ids.length
                    ? this.livechat_agent_history_ids
                    : this.livechat_bot_history_ids;
            }
```

addons/im_livechat/static/src/embed/common/livechat_service.js
```diff
@@ -84,7 +84,7 @@ export class LivechatService {
             // flickering, we do not want another load that would result in the
             // same issue.
             savedChannel.scrollUnread = false;
-            temporaryThread.channel.chatWindow?.close();
+            temporaryThread.channel?.chatWindow?.close();
```
tsm-odoo: Your explanation is only partial, and so is the fix. The issue indeed comes from the `public_web` bundle being loaded. Specifically, `autoOpenChatWindowOnNewMessage` opens the new channel when the first message is received. 

However, this fix only addresses the traceback; `readyToSwapDeferred` is still not awaited. It was introduced to avoid flickers, so we need to ensure everything is ready before swapping the two channels. 

With this fix, the channels are swapped when the new message notification is received, without any guarantee that everything is fully ready yet.

tsm-odoo (changes requested):
We should enrich an existing tour to ensure the chat window header properly displays the chat bot name.

addons/im_livechat/static/src/embed/common/livechat_service.js
aych-odoo: In the website editor, the livechat embed bundle is loaded inside an iframe, while the web/public_web bundle is already loaded in the parent frame.

When a new message arrives, the PARENT frame attempts to open the chat window because `autoOpenChatWindowOnNewMessage` is enabled. To do this, the PARENT frame updates localStorage to mark the chat hub as opened and folded.

This localStorage event is received by the IFRAME, which then closes the transient thread and opens the persistent one, as `ChatHub` is not aware of `readyToSwapDeferred` promise.

This bug also existed in previous versions, but it went unnoticed because it did not trigger any traceback.

tsm-odoo: I understand what happens but still, not awaiting `readyToSwapDeferred` can lead to flickers. We should find a way to respect it.

addons/im_livechat/static/src/embed/common/thread_model_patch.js
```diff
@@ -40,9 +39,9 @@ patch(Thread.prototype, {
          * with its temporary counterpart (i.e. when the actions following the
          * persist call are done to avoid flickering).
          *
-         * @type {Deferred}
+         * @type {PromiseWithResolvers}
```
tsm-odoo:
```suggestion
         * @type {PromiseWithResolvers<void>}
```

tsm-odoo: Actually, we shouldn't change it in stable, I'd keep deferred until we reach master.

addons/im_livechat/static/src/embed/common/thread_model_patch.js
```diff
@@ -40,9 +39,9 @@ patch(Thread.prototype, {
...
          */
-        this.readyToSwapDeferred = new Deferred();
+        this.readyToSwapDeferred = null;
```
tsm-odoo: I'd add a different flag rather than hacking `readyToSwapDeferred`. Its not obvious that the absence of this deferred means that the chat was created from another env.

addons/mail/static/src/core/common/chat_hub_model.js
```diff
@@ -119,10 +119,10 @@ export class ChatHub extends Record {
         /** @param {import("models").Channel[]} channels */
         const insertChatWindows = (channels) =>
             channels
-                .filter((channel) => channel?.exists())
+                .filter((channel) => this._shouldInsertChatWindow(channel))
                 .map((channel) => this.store.ChatWindow.insert({ channel }));
         const toFold = insertChatWindows(foldChannels);
-        const toOpen = insertChatWindows(openChannels);
+        const toOpen = this._prepareChatWindowsToOpen(insertChatWindows(openChannels));
```
tsm-odoo: It looks very hacky and quite hard to follow to add an override point (`_prepareChatWindowsToOpen`) just to remove the transient live chat, which will prevent closing the transient thread later on (`#127`). I feel like the correct approach would be to prevent closing transient chat from the `_load` method: the user opened it deliberately, of course, `getOrFetch` cannot return it because the server doesn't know about it, so it only make sense to preserve it.

addons/im_livechat/static/src/embed/common/thread_model_patch.js
tsm-odoo: Actually, relying on the env instead of modelization is quite a bad pattern we're trying to get rid of. The issue comes from `autoOpenChatWindowOnNewMessage`. This should only occur for agents (as opening is handled by the live chat service on the visitor side in order to cleanly swap the two threads), so a cleaner fix would be to change `autoOpenChatWindowOnNewMessage` to disable it for visitors?

addons/im_livechat/static/src/embed/common/thread_model_patch.js
seb-odoo: 19.2 is not exactly stable yet, it's fine to change here

addons/im_livechat/static/src/embed/common/thread_model_patch.js
aych-odoo: We can add a condition to only auto open for visitors, but that would create a folded chatbubble for a brief amount of time while the thread is not ready to be swapped.

<video src="https://github.com/user-attachments/assets/1bb7af74-4718-4ca6-831d-addc34c4781f" ></video>
(i added a delay in `Chatbot@processAnswer`)

tsm-odoo: You might have a look at `inChathubOnNewMessage` then.

addons/mail/static/src/core/common/chat_hub_model.js
```diff
@@ -116,13 +116,15 @@ export class ChatHub extends Record {
...
-        /** @param {import("models").Channel[]} channels */
+        /** @param {import("models").DiscussChannel[]} channels */
         const insertChatWindows = (channels) =>
             channels
                 .filter((channel) => channel?.exists())
                 .map((channel) => this.store.ChatWindow.insert({ channel }));
         const toFold = insertChatWindows(foldChannels);
         const toOpen = insertChatWindows(openChannels);
+        // Ensure cross-tab updates don't close transient channels' chatWindow as they may not exist in another tab
+        toOpen.unshift(...this.opened.filter((cw) => cw.channel.isTransient && cw.notIn(toOpen)));
```
tsm-odoo: Why not avoiding the closing of the chat window below? It looks cleaner that adding them back.

aych-odoo: https://github.com/odoo/odoo/blob/78705345712cc06e630e5594fee8e88f8da23f03/addons/mail/static/src/core/common/chat_hub_model.js#L125-L134

We require the transient chatWindows to be there in `toOpen` else the chatwindow will be closed because of `hubAsCanShowOpened`

tsm-odoo (approved):
Thanks for the update!

@robodoo r+

PR #247591, opened 2026-02-06
[IMP] point_of_sale: currency symbol on cdis

Ysoroko:
This PR adds the currency symbol on the customer display for point of sale for the total amount + change. (formatted based on the currency, e.g. $ before, eur after the amount) This is required by the scale certification 

Before:
image

After:
image

---


stva-odoo:
robodoo r+

PR #247596, opened 2026-02-06
[FIX] point_of_sale: persist frozen time in tours across reloads

ushyme:
Mocking time in tours via `freezeDateTime` is lost when the page reloads , causing date-filtered data to disappear and tests to fail.

Introduce `withTimeFreeze(millis, steps)`, which persists the mock timestamp in `sessionStorage`. POS now checks this storage on module load to automatically re-apply the freeze, ensuring the clock survives refreshes while handling cleanup after the steps finish.

runbot-232601

Related Enterprise PR: odoo/enterprise#106724


robinengels:
@robodoo r+

PR #247600, opened 2026-02-06
[FIX] payment: allow unarchiving express payment tokens

mma1377:
### Issue:
Due to this issue payment tokens with an express checkout payment, once archived cannot be unarchived.

#### Steps to reproduce:
1- Enable an express checkout method. We can use Demo.
2- On website, add a payment_subscription product to the cart.
3- Navigate to the cart. Pay using express checkout.
4- Navigate to payment tokens, and archive the created token.
5- Add archived filter to the list and unarchive the token.

You will get an error:
`You can't unarchive tokens linked to inactive payment methods.`

### Cause:
`payment_method_id` of express payments is set to `payment_method_unknown`. However, this payment method is not active, raising this user error.

opw-5886048

### DIFF
--- a/addons/payment/models/payment_token.py
+++ b/addons/payment/models/payment_token.py
@@ -83,7 +83,10 @@ class PaymentToken(models.Model):
         if 'active' in vals:
             if vals['active']:
                 if any(
-                    not token.payment_method_id.active
+                    (
+                        token.payment_method_id != self.env.ref('payment.payment_method_unknown')
+                        and not token.payment_method_id.active
+                    )
                     or token.provider_id.state == 'disabled'
                     for token in self
                 ):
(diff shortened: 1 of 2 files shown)


AntoineVDV (commented):
Hello, I discussed this with the PO because I'm not really fond of making an exception for a specific payment method in the `write`. The 'unknown' payment method is used for more than just express checkout transactions, and I wouldn't want to introduce unwanted behaviors because of this change.

The steps to reproduce this problem are already quite convoluted and rely on mechanisms that were not really intended in the first place (auto-tokenization of transactions when paying for subscription products through express checkout), so I prefer not to attempt to fix it. The workaround would be to either create a new token or to assign the 'unknown' PM to a provider, then unarchive the desired tokens.

PR #247604, opened 2026-02-06
[IMP] tools: add `safe_eval` runtime analysis

thle-odoo:
The execution of arbitrary code in Odoo via `safe_eval` already takes place in a restricted environment, in particular via prior opcode, `co_names` static analysis and specific builtins in the evaluation context.

The purpose of this commit is to introduce the ability to dynamically analyse code executed via the `safe_eval` tool.

The dangerous situation we want to avoid is transferring logic to code that is considered untrusted. In other words, we want to avoid calls that use objects that are not in the whitelist.

The behavior to adopt when detecting an unsafe object depends on the `--unsafe-policy` configuration parameter:
- `disable`: Disable the feature
- `log`: Log a warning (default)
- `raise`: Raise an exception
- `terminate`: Terminate the process

Note:
Python 3.12+ is required since the [sys.monitoring](https://docs.python.org/3.12/library/sys.monitoring.html) API is used.
The presence of the JSON C extension library is required (`_json`) for performance reasons.

_This is not a sandbox in general, but it helps prevent dangerous behavior in the context of `safe_eval`._

Task-4658036

### DIFF
--- a/odoo/addons/test_tools/tests/test_safe_eval.py
+++ b/odoo/addons/test_tools/tests/test_safe_eval.py
@@ -1,11 +1,18 @@
 import ast
 
 from textwrap import dedent
+from unittest.mock import patch
 
 from odoo import Command
-from odoo.tests.common import tagged, BaseCase
+from odoo.tests.common import tagged, BaseCase, TransactionCase
 from odoo.tools import mute_logger
-from odoo.tools.safe_eval import safe_eval, const_eval, expr_eval
+from odoo.tools.safe_eval import (
+    const_eval,
+    expr_eval,
+    safe_eval,
+    UnsafeObjectError,
+    UnsafePolicy,
+)
 
 
 @tagged('at_install', '-post_install')
@@ -185,3 +192,147 @@ class TestSafeEval(BaseCase):
                 c: int = a + b
                 return c
         """), mode="exec")
+
+
+@tagged('at_install', '-post_install')
+class TestSafeEvalRuntime(TransactionCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.startClassPatcher(patch(
+            'odoo.tools.safe_eval.unsafe_policy',
+            lambda: UnsafePolicy.RAISE
+        ))
+
+        class UnsafeClass:
+            __module__ = 'odoo.unsafe_module'
+
+            def __init__(self, *args, **kwargs): ...
+
+        cls.unsafe_context = {'UnsafeClass': UnsafeClass}
+
+    def test_transform_decorators(self):
+        steps = []
+
+        def make_deco(x):
+            steps.append(f'make_deco_{x}')
+
+            def deco(func):
+                steps.append(f'deco_{x}')
+
+                def wrapper(*args, **kwargs):
+                    steps.append(f'wrapper_{x}')
+                    return func(*args, **kwargs)
+
+                return wrapper
+
+            return deco
+
+        expr = """
+            @make_deco('A')
+            @make_deco('B')
+            def func(*args, **kwargs):
+                return
+
+            func()
+        """
+        safe_eval(dedent(expr), {'make_deco': make_deco}, mode='exec')
+        self.assertListEqual(steps, [
+            'make_deco_A', 'make_deco_B',  # Evaluation order
+            'deco_B', 'deco_A',  # Wrapping order (reversed)
+            'wrapper_A', 'wrapper_B',  # Evaluation order
+        ])
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_callee(self):
+        expr = """
+            UnsafeClass()
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_args(self):
+        expr = """
+            callee = lambda *args, **kwargs: ...
+            callee(UnsafeClass)
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_kwargs(self):
+        expr = """
+            callee = lambda *args, **kwargs: ...
+            callee(kw=UnsafeClass)
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_structure(self):
+        expr = """
+            callee = lambda *args, **kwargs: ...
+            struct = {'a': {'b': {'c': UnsafeClass}}}
+            callee(struct)
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+        expr = """
+            callee = lambda *args, **kwargs: ...
+            struct = ['a', 'b', 'c', UnsafeClass]
+            callee(struct)
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_builtin_callee(self):
+        expr = """
+            map(UnsafeClass, ['foo'])
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+        expr = """
+            filter(UnsafeClass, ['foo'])
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+        expr = """
+            sorted(['foo'], key=UnsafeClass)
+        """
+        with self.assertRaisesRegex(ValueError, '^UnsafeObjectError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+
+    @mute_logger('odoo.tools.safe_eval.runtime')
+    def test_check_callee_qweb(self):
+        arch = '''
+            <t t-name="template">
+                <div>
+                    <t t-out="UnsafeClass()"/>
+                </div>
+            </t>
+        '''
+        view = self.env['ir.ui.view'].create({
+            'name': 'test',
+            'type': 'qweb',
+            'arch_db': arch,
+        })
+        with self.assertRaises(UnsafeObjectError):
+            self.env['ir.qweb']._render(view.id, self.unsafe_context)
+
+    def test_override_call(self):
+        expr = """
+            # Override in locals
+            _save_eval_call = lambda callee, *args, **kwargs: callee(*args, **kwargs)
+            UnsafeClass()
+        """
+        with self.assertRaisesRegex(ValueError, '^AssertionError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+        # Note that without the assert protection, we get a `RecursionError`,
+        # because after transformer, the code becomes:
+        # `_save_eval_call = lambda callee, *args, **kwargs: _save_eval_call(callee, *args, **kwargs)`
--- a/odoo/orm/model_classes.py
+++ b/odoo/orm/model_classes.py
@@ -19,6 +19,7 @@ from odoo.tools import (
     frozendict,
     sql,
 )
+from odoo.tools.safe_eval import safe_checker
 from odoo.tools.translate import FIELD_TRANSLATE
 
 if typing.TYPE_CHECKING:
@@ -227,6 +228,7 @@ def add_to_registry(registry: Registry, model_def: type[BaseModel]) -> type[Base
     for model_name in registry.descendants([name], '_inherit', '_inherits'):
         registry[model_name]._setup_done__ = False
 
+    safe_checker.add_hook(model_cls, None)  # Optimization to serialize recordset(s) faster
     return model_cls
 
 
--- a/odoo/orm/models.py
+++ b/odoo/orm/models.py
@@ -53,7 +53,7 @@ from odoo.tools.constants import PREFETCH_MAX
 from odoo.tools.func import deprecated
 from odoo.tools.lru import LRU
 from odoo.tools.misc import ReversedIterable, exception_to_unicode, unquote
-from odoo.tools.safe_eval import _UNSAFE_ATTRIBUTES
+from odoo.tools.safe_eval import _UNSAFE_ATTRIBUTES, safe_checker
 from odoo.tools.translate import _, LazyTranslate
 
 from . import decorators as api
@@ -309,6 +309,8 @@ class MetaModel(type):
                 add_default('write_date', Datetime(
                     string='Last Updated on', readonly=True))
 
+
+safe_checker.add_hook(MetaModel, None)  # Optimization to serialize model class(es) faster
 
 # special columns automatically created by the ORM
 LOG_ACCESS_COLUMNS = ['create_uid', 'create_date', 'write_uid', 'write_date']
--- a/odoo/tools/config.py
+++ b/odoo/tools/config.py
@@ -260,6 +260,14 @@ class configmanager:
                          help="Comma-separated list of server-wide modules.")
         group.add_option("-D", "--data-dir", dest="data_dir", type='path',  # sensitive default set in _load_default_options
                          help="Directory where to store Odoo data")
+        group.add_option("--unsafe-policy", dest='unsafe_policy', type='choice', my_default='log',
+                         choices=['disable', 'log', 'raise', 'terminate'],
+                         help="Policy if an unsafe object is detected during          "
+                              "arbitrary code execution                               "
+                              "- disable: No action taken                             "
+                              "- log: Log a warning with unsafe object information    "
+                              "- raise: Raise an exception (BaseException)            "
+                              "- terminate: Terminate the current worker process      ")
         parser.add_option_group(group)
 
         # HTTP
--- a/odoo/tools/safe_eval.py
+++ b/odoo/tools/safe_eval.py
@@ -14,24 +14,71 @@ condition/math builtins.
 #  - http://code.activestate.com/recipes/286134/
 #  - safe_eval in lp:~xrg/openobject-server/optimize-5.0
 #  - safe_eval in tryton http://hg.tryton.org/hgwebdir.cgi/trytond/rev/bbb5f73319ad
+import ast
+import contextvars
 import dis
 import functools
+import gc
 import logging
+import os
+import re
 import sys
 import types
 import typing
 import zoneinfo
+from collections import defaultdict, OrderedDict
+from enum import auto, IntEnum
+from json.encoder import c_make_encoder
 from opcode import opmap, opname
 from types import CodeType
+from weakref import WeakKeyDictionary
 
 import werkzeug
 from psycopg2 import OperationalError
 
 import odoo.exceptions
+from .config import config
+from .func import lazy
+from .misc import OrderedSet
 
 unsafe_eval = eval
 
 __all__ = ['const_eval', 'safe_eval']
+
+_logger = logging.getLogger(__name__)
+_logger_runtime = logging.getLogger(f'{__name__}.runtime')
+
+
+class UnsafePolicy(IntEnum):
+    DISABLE = 0
+    LOG = auto()
+    RAISE = auto()
+    TERMINATE = auto()
+
+
+@functools.cache
+def unsafe_policy():
+    policy = UnsafePolicy[config['unsafe_policy'].upper()]
+
+    if policy and c_make_encoder is None:
+        _logger_runtime.warning(
+            "Unsafe object policy '%s' is configured, but it is not supported on this runtime "
+            "(missing native encoder). The policy will be ignored.",
+            policy.name,
+        )
+        policy = UnsafePolicy.DISABLE
+
+    return policy
+
+
+class UnsafeObjectError(BaseException):
+
+    def __init__(self, obj):
+        obj_path = safe_whitelist.get_full_path(obj)
+        super().__init__(
+            f'Unsafe object access (type {type(obj)!r}): {obj!r} (path: {obj_path})',
+        )
+
 
 # The time module is usually already provided in the safe_eval environment
 # but some code, e.g. datetime.datetime.now() (Windows/Python 2.5.2, bug
@@ -195,8 +242,6 @@ _SAFE_OPCODES = _EXPR_OPCODES.union(to_opcodes([
 ])) - _BLACKLIST
 
 
-_logger = logging.getLogger(__name__)
-
 def assert_no_dunder_name(code_obj, expr):
     """ assert_no_dunder_name(code_obj, expr) -> None
 
@@ -264,12 +309,20 @@ def compile_codeobj(expr: str, /, filename: str = '<unknown>', mode: typing.Lite
     try:
         if mode == 'eval':
             expr = expr.strip()  # eval() does not like leading/trailing whitespace
-        code_obj = compile(expr, filename or '', mode)
+
+        if unsafe_policy():
+            tree = ast.parse(expr, mode=mode)
+            tree = safe_transform(tree)
+            code_obj = compile(tree, filename or '', mode)
+            add_monitoring(code_obj)
+            return code_obj
+
+        return compile(expr, filename or '', mode)
+
     except (SyntaxError, TypeError, ValueError):
         raise
     except Exception as e:
         raise ValueError('%r while compiling\n%r' % (e, expr))
-    return code_obj
 
 
 def const_eval(expr):
@@ -409,7 +462,7 @@ def safe_eval(expr, /, context=None, *, mode="eval", filename=None):
     except _BUBBLEUP_EXCEPTIONS:
         raise
 
-    except Exception as e:
+    except (Exception, UnsafeObjectError) as e:  # noqa: BLE001
         raise ValueError('%r while evaluating\n%r' % (e, expr))
 
     finally:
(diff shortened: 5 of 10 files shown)


odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,55 @@
...
 
+# Enable runtime analysis for unsafe object usage.
+# The configured level determines the action taken:
+#   - 0: Disabled
+#   - 1: Log a warning (default)
+#   - 2: Raise an exception
+#   - 3: Terminate the process (prevent exception escape)
+PROTECTED = sys.version_info > (3, 12) \
+            and c_make_encoder \
+            and int(os.getenv('PROTECTED', '1'))
```
kmagusiak: 1. You should drop > 3.12 -- I think you meant >= 3.12 and that's already the minimum version.
2. If `c_make_encoder` is not available, do you want to deactivate it all or just run slower?
3. The name PROTECTED as an environment variable is too generic. What about "ODOO_SAFE_EVAL_SECURITY"?
  Maybe instead of int values, use names "log", "raise", "terminate" and `os.getenv('...', 'log').lower()`.

In any case, if you have the env variable set and you don't enable the protection, log a warning.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+            ast.copy_location(call, node)
+            ast.copy_location(call.func, node)
+            ast.copy_location(call.args[1], node)
+            ast.copy_location(assign, node)
+            ast.copy_location(assign.targets[0], node)
+            stmts.append(assign)
+        return stmts
+
+
+Transformer = _Transformer()
```
kmagusiak: I don't recall exactly why you wanted to do this. maybe you forgot `del _Transformer` after this line so that you have the singleton instantiated once.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+            args=[node.func, *node.args],
+            keywords=node.keywords)
+        ast.copy_location(call, node)
+        ast.copy_location(call.func, node)
+        return call
+
+    def visit_FunctionDef(self, node):
+        """
+        Transforms:
+            @decorator
```
kmagusiak: Nitpick: Provide an example with 2 decorators.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+        SafeObject.add_function(('urllib', 'parse', 'quote'))
+        SafeObject.add_function(('base64', ('b64decode', 'b64encode', 'urlsafe_b64decode', 'urlsafe_b64encode')))
+        # Werkzeug
+        SafeObject.add_class(('werkzeug', 'datastructures', 'structures', ('ImmutableMultiDict', 'MultiDict')))
+        SafeObject.add_class(('werkzeug', 'exceptions', ('NotFound', 'BadRequest', 'Forbidden', 'Unauthorized', 'RequestEntityTooLarge')))
+        SafeObject.add_instance(('werkzeug', 'local', 'LocalProxy'))
+        SafeObject.add_instance(('werkzeug', 'datastructures', 'file_storage', 'FileStorage'))
+        SafeObject.add_function(('werkzeug', 'datastructures', 'structures', 'TypeConversionDict', '*'))
+        # Test
+        if not config['test_enable']:
```
kmagusiak: When importing the module, the config may not be parsed yet. The code below should be moved to `odoo.tests`.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+        return cls._instance
+
+    __SEQUENCE_TYPES = frozenset((list, tuple, set, frozenset, OrderedSet))
+    __MAPPING_TYPES = frozenset((dict, defaultdict, OrderedDict, types.MappingProxyType))
+    __TRUSTED_TYPES = frozenset((
+        *__SEQUENCE_TYPES, *__MAPPING_TYPES,
+        object, bool, int, float, str, bytes, bytearray, memoryview,
+        types.NoneType, types.MappingProxyType,
+        filter, map, enumerate, range, zip, reversed,
+        Exception, ValueError, TypeError, AttributeError, KeyError, ZeroDivisionError, UnboundLocalError,
```
kmagusiak: Nitpick: `UnboundLocalError` seems strange, it is a programming error.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+        SafeObject.add_instance(('pytz', ('tzfile', 'UTC')))
+        SafeObject.add_function((('datetime', 'date', 'time'), '*'))
+        SafeObject.add_function(('pytz', 'tzinfo', 'DstTzInfo', 'localize'))
+        SafeObject.add_function(('pytz', 'UTC', 'localize'))
+        # Miscellaneous
+        SafeObject.add_class(('markupsafe', '*'))
+        SafeObject.add_class(('collections', ('defaultdict', 'OrderedDict')))
+        SafeObject.add_class(('collections', 'abc', ('Sized', 'Mapping', 'ValuesView')))
+        SafeObject.add_class(('itertools', 'islice'))
+        SafeObject.add_instance(('uuid', 'UUID'))
```
kmagusiak: Nitpick: can you sort within each group?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+    __TRUSTED_TYPES = frozenset((
+        *__SEQUENCE_TYPES, *__MAPPING_TYPES,
+        object, bool, int, float, str, bytes, bytearray, memoryview,
+        types.NoneType, types.MappingProxyType,
+        filter, map, enumerate, range, zip, reversed,
+        Exception, ValueError, TypeError, AttributeError, KeyError, ZeroDivisionError, UnboundLocalError,
+    ))
+    __TRUSTED_BUILTIN_FUNCTIONS = frozenset((
+        min, max, sum, abs, sorted, round, len, repr, all, any, ord, chr, divmod,
+        isinstance, hasattr, functools.reduce,
```
kmagusiak: Nitpick: `reduce` is not a built-in function.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+        self[types.ModuleType] = self.__check_module
+        self[types.BuiltinFunctionType] = self.__check_builtin_function
+        self[types.BuiltinMethodType] = self.__check_builtin_function
+        self[types.FunctionType] = self.__check_function
+        self[types.MethodType] = self.__check_function
+        self[types.MethodDescriptorType] = self.__check_descriptor
+        self[functools.partial] = self.__check_partial
+        self[lazy] = self.__check_lazy
+        self[type(reversed([]))] = self.__check_simple_iterator
+        self[types.GeneratorType] = None  # TODO
```
kmagusiak: What TODO?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+
+        paths = []
+        for path in trusted:
+            elts = [[sub_path] if not isinstance(sub_path, tuple) else sub_path for sub_path in path]
+            paths.extend(itertools.product(*elts))
+
+        re_paths = []
+        for path in paths:
+            parts = ['.*' if part == '*' else re.escape(part) for part in path]
+            re_paths.append(r'.'.join(parts))
```
kmagusiak: I think you want that "abc.*.def" matches only "abc.a.def", not multiple parts. If you want multiple parts, did you check the performance with `r'.*?'`?
```suggestion
            parts = [r'[^.]*' if part == '*' else re.escape(part) for part in path]
            re_paths.append(r'\.'.join(parts))
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+
+    add_class = functools.partialmethod(__add, 'class')
+    add_instance = functools.partialmethod(__add, 'instance')
+    add_function = functools.partialmethod(__add, 'function')
+    __re_class = functools.cached_property(lambda self: self.__re_compile(self.__class))
+    __re_instance = functools.cached_property(lambda self: self.__re_compile(self.__instance))
+    __re_function = functools.cached_property(lambda self: self.__re_compile(self.__function))
+
+    def handle_error(self, error):
+        _logger.warning('[SAFE_EVAL_RUNTIME] UnsafeObjectError: %s', error)
```
kmagusiak: Nitpick: instead of prefixing with "[SAFE_EVAL_RUNTIME]", you can create a second logger `_logger_runtime = logging.getLogger(f'{__name__}.runtime')` or similar.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +522,398 @@ def __repr__(self):
...
+        try:
+            encoder = self.__encoder_ctx.get()
+        except LookupError:
+            # markers, default, encoder, indent, key_separator, item_separator, sort_keys, skipkeys, allow_nan
+            encoder = c_make_encoder({}, self.__check, lambda x: '', None, '', '', False, False, False)
+            self.__encoder_ctx.set(encoder)
+        encoder.markers.clear()
+        return encoder
+
+    def __getitem__(self, obj) -> typing.Callable | None:
```
kmagusiak: Overwriting `__getitem__` might be confusing. Maybe you could just name this method.
```py
In [1]: so = odoo.tools.safe_eval.SafeObject

In [2]: so.get(int)

In [3]: so[int]
Out[3]: <bound method _SafeObject.__check_class of <_SafeObject at 0x7fe66e5baa50>>
```

odoo/tools/safe_eval.py
thle-odoo: This is to avoid instantiating the transformer in `compile_codeobj` each time.

> maybe you forgot del _Transformer after this line so that you have the singleton instantiated once.

I will add it. I will also apply this comment to the `SafeObject` singleton.

odoo/tools/safe_eval.py
thle-odoo: It is true that this can be confusing. I opted for `get_handler`.

odoo/tools/safe_eval.py
thle-odoo: I added a brief description (task to be completed during the second iteration).

odoo/tools/safe_eval.py
thle-odoo: I will sort for each subgroup in groups as well (`class`|`instance`|`function`)

odoo/tools/safe_eval.py
thle-odoo: I adapted the code with your comments to improve the management of this feature.

> If c_make_encoder is not available, do you want to deactivate it all or just run slower?

I prefer to disable it to not impact performance.

Note:
This requirement is generally met (a specific build must be performed to ensure that it is not present).

odoo/tools/safe_eval.py
thle-odoo: It is a programming error, however, it is harmless. Nothing prevents someone from writing the following code to retrieve it:
```py
def foo():
    a = b
    b = 1

try:
    foo()
except Exception as e:
    unbound_local_error = e
```
This exception class remains safe (it is therefore not the role of the feature to block it).
But I understand that it may seem strange.

odoo/tools/safe_eval.py
thle-odoo: I completely agree.
However, functions that are implemented in C (in CPython) are marked as such. These functions are placed in modules for clarity.
```
>>> import functools
>>> type(functools.reduce)
<class 'builtin_function_or_method'>
```

odoo/tools/safe_eval.py
thle-odoo: > I think you want that "abc.*.def" matches only "abc.a.def", not multiple parts.

In this case, yes.
However, the use of `'*'` in tuples that define trusted objects should only be at the end.
```
path = ('abc', '*')
'abc.foo' -> ok
'abc.foo.bar' -> ok
```
So I think I will keep using `.*`, but I will make sure it is at the end.

In terms of performance, when using `fullmatch`, the greedy quantifier is faster. This is because `fullmatch` must validate the entire string. As a result, the lazy quantifier will test all possible expansions until it has consumed the entire string (lazy backtracking). Meanwhile, the greedy quantifier will consume all characters at once.

Please feel free to correct me if necessary.

```py
import re, timeit

pattern_greedy = re.compile(r'foo.*')
def test_greedy(s):
    return pattern_greedy.fullmatch(s)

pattern_lazy = re.compile(r'foo.*?')
def test_lazy(s):
    return pattern_lazy.fullmatch(s)

s = 'foo.' + '.'.join(['x']*1000)

print('Greedy:', timeit.timeit('test_greedy(s)', globals=globals(), number=1000))
print('Lazy:', timeit.timeit('test_lazy(s)', globals=globals(), number=1000))
```
```
Greedy: 0.0007410540001728805
Lazy:   0.016903445000025386
```

odoo/tools/safe_eval.py
kmagusiak: Ok, this is confusing :)
Why not put all functions from `_initialize_safe_object` from "C functions" here or move this one there?

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,72 @@
...
+    def __bool__(self):
+        return self is not UnsafePolicy.DISABLE
+
+
+try:
+    UNSAFE_POLICY = UnsafePolicy(os.getenv('ODOO_UNSAFE_POLICY', UnsafePolicy.LOG).lower())
+except ValueError:
+    UNSAFE_POLICY = UnsafePolicy.LOG
+
+if UNSAFE_POLICY and (sys.version_info < (3, 12) or c_make_encoder is None):
```
kmagusiak: Nitpick: you cannot have version <3.12 - see `odoo.release.MIN_PY_VERSION`.

odoo/tools/safe_eval.py
thle-odoo: We can move everything to one side or the other.
The `__TRUSTED_BUILTIN_FUNCTIONS` is an internal optimization for the check (which I prefer to keep).
Putting everything in this `frozenset` would be "heavy" but feasible (I plan to improve the check of builtins/descriptors in the future).

As for `functools.reduce`, it will be accepted via: `SafeObject.add_function(('_functools', 'reduce'))` in the `Miscellaneous` category. I am removing it from the `frozenset` to avoid ambiguity.

kmagusiak: As you have a plan, that's the important part. :)

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,72 @@
...
+    def __bool__(self):
+        return self is not UnsafePolicy.DISABLE
+
+
+try:
+    UNSAFE_POLICY = UnsafePolicy(os.getenv('ODOO_UNSAFE_POLICY', UnsafePolicy.LOG).lower())
+except ValueError:
+    UNSAFE_POLICY = UnsafePolicy.LOG
+
+if UNSAFE_POLICY and (sys.version_info < MIN_PY_VERSION or c_make_encoder is None):
```
kmagusiak: You cannot be below the min version.
```suggestion
if UNSAFE_POLICY and c_make_encoder is None:
```

kmagusiak (approved):
Nice work 👍

rugo-odoo (commented):
Seems like a big diff for it to not have any specific test 😬 Shouldn't we cover it ?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,402 @@ def __repr__(self):
...
 dateutil.tz.gettz = zoneinfo.ZoneInfo
+
+
+class _Transformer(ast.NodeTransformer):
+
+    def __new__(cls):
+        if hasattr(cls, '_instance'):
+            raise AssertionError(f"{cls} is a singleton")
+        cls._instance = super().__new__(cls)
+        return cls._instance
```
rco-odoo: What's wrong with having several class instances?  Would it make the code incorrect?  Let's simply use a single instance, and call it a day.  Unless I am mistaken, you really don't need defensive code for that.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,402 @@ def __repr__(self):
...
+    __re_function = functools.cached_property(lambda self: self.__re_compile(self.__function))
+
+    def handle_error(self, error):
+        _logger_runtime.warning(error)
+        if UNSAFE_POLICY == UnsafePolicy.RAISE:
+            raise error
+        if UNSAFE_POLICY == UnsafePolicy.TERMINATE:
+            os.kill(os.getpid(), signal.SIGKILL)
+
+    __encoder_ctx = contextvars.ContextVar('__encoder_ctx')
```
rco-odoo: Please explain why you need a context var for the encoder.  Is it not thread-safe, maybe?

rco-odoo (commented):
I find the code inside `safe_eval.py` quite obscure, which makes the review hard to say the least.  And possibly makes its maintenance hard, too...

There's a bunch of things in this PR that I consider bad practice (variable names, inheritance instead of composition, etc).  I believe we could make the code simpler and more readable.  I can provide some help, if you want to.

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,76 @@
...
+    RAISE = 'raise'
+    TERMINATE = 'terminate'
+
+    def __bool__(self):
+        return self is not UnsafePolicy.DISABLE
+
+
+try:
+    UNSAFE_POLICY = UnsafePolicy(os.getenv('ODOO_UNSAFE_POLICY', UnsafePolicy.LOG).lower())
+except ValueError:
```
xmo-odoo: Nit: given there's already a fallback I'd go with

```suggestion
    UNSAFE_POLICY = UnsafePolicy[os.environ['ODOO_UNSAFE_POLICY'].lower()]
except KeyError:
```
and probably drop the Str off of the enum unless it's necessary for something else? I think stringly-typing enums is mostly for backwards compatibility purposes.

(and `EnumType.__getitem__` is a lot more straightforward than `EnumType.__call__` if you don't need the complexity and fallbacks, which think this does not)

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,76 @@
...
+if UNSAFE_POLICY and c_make_encoder is None:
+    _logger_runtime.warning(
+        "Unsafe object policy '%s' is configured, but it is not supported on this runtime "
+        "(missing native encoder). The policy will be ignored.",
+        UNSAFE_POLICY,
+    )
+    UNSAFE_POLICY = UnsafePolicy.DISABLE
+
+
+class UnsafeObjectError(BaseException):
```
xmo-odoo, line 74: Is punching through `Exception` necessary?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+    return callee(*args, **kwargs)
+
+
+class _Singleton:
+
+    def __new__(cls):
+        if hasattr(cls, '_instance'):
+            raise AssertionError(f"{cls} is a singleton")
+        cls._instance = super().__new__(cls)
+        return cls._instance
```
xmo-odoo:
```suggestion
```
Don't see the point.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+    sys.monitoring.register_callback(TOOL_ID, EVENTS.CALL, monitoring_call)
+
+    def monitoring_codeobj(code):
+        codes = [code]
+        while codes:
+            code = codes.pop()
+            sys.monitoring.set_local_events(TOOL_ID, code, EVENTS.CALL)
+            codes.extend(const for const in code.co_consts if isinstance(const, types.CodeType))
+
+    @functools.lru_cache(maxsize=1)
```
xmo-odoo:
```suggestion
    @functools.cache
```
Having a maxsize for a cache with a static number of entries (especially 1) is a deoptimisation as the cache needs to add the LRU linked list and update it on every call. `cache()` (which is an alias for `lru_cache(maxsize=None)` is just a dictionary, with the only overhead being updating the statistics.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+
+    def monitoring_call(code, instruction_offset, callee, arg0):
+        if callee is __CALL__:
+            return
+        frame = sys._getframe(1)
+        assert (call_func := frame.f_locals.get(CALL_ID)) is None or call_func is __CALL__
+        assert (call_func := frame.f_globals.get(CALL_ID)) is None or call_func is __CALL__
+        assert frame.f_builtins[CALL_ID] is __CALL__
+
+    sys.monitoring.free_tool_id(TOOL_ID)
```
xmo-odoo: That seems like a misuse of the API. `free_tool_id` is used to unregister your tool, not to smash over an existing one.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+        return stmts
+
+
+class _SafeEncoder(_Singleton):
+
+    MAPPINGS = frozenset((dict, defaultdict, OrderedDict, types.MappingProxyType))
+    SEQUENCES = frozenset((list, tuple, set, frozenset, OrderedSet))
+
+    def __init__(self):
+        self.encoder_ctx = contextvars.ContextVar('encoder_ctx')  # Encoder must be thread safe (because of the markers structure)
```
xmo-odoo: Probably less relevant than in most contexts but

> Context Variables should be created at the top module level

contextvars are essentially always globals as they're strongly bound to the Context.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+    EVENTS = sys.monitoring.events
+    TOOL_ID = 4  # Not prevent other tools from using "official IDs"
+
+    def monitoring_call(code, instruction_offset, callee, arg0):
+        if callee is __CALL__:
+            return
+        frame = sys._getframe(1)
+        assert (call_func := frame.f_locals.get(CALL_ID)) is None or call_func is __CALL__
+        assert (call_func := frame.f_globals.get(CALL_ID)) is None or call_func is __CALL__
+        assert frame.f_builtins[CALL_ID] is __CALL__
```
xmo-odoo: So this is only here to try and flag call sites we might have failed to instrument?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+                # Serialisation failure
+                _logger_runtime.warning('_json.c_make_encoder: %s %r', e, o)
+
+    def _default(self, obj):
+        """ Dispatch to the default check-serialisation function of `obj`. """
+        t_obj = type(obj)
+        try:
+            default_ = self.defaults[t_obj]
+        except KeyError:
+            if issubclass(t_obj, type):
```
xmo-odoo: This seems wrong. Classes are not subclasses of `type`, they're instances of `type`. *Metaclasses* are subclasses of `type`.

Furthermore calling `issubclass` with a non-class object will fail with a `TypeError`.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+
+
+class _SafeEncoder(_Singleton):
+
+    MAPPINGS = frozenset((dict, defaultdict, OrderedDict, types.MappingProxyType))
+    SEQUENCES = frozenset((list, tuple, set, frozenset, OrderedSet))
+
+    def __init__(self):
+        self.encoder_ctx = contextvars.ContextVar('encoder_ctx')  # Encoder must be thread safe (because of the markers structure)
+        self.defaults: WeakKeyDictionary[type, typing.Callable | None] = WeakKeyDictionary()
```
xmo-odoo: I understand that it is the phraseology of the json module, but using `defaults` is very confusing when we're not actually trying to default anything, these are check hooks.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+        if (module := getattr(cls_obj, '__module__', None)) and module in sys.modules:
+            return f"{module}.{qualname}".strip('.')
+        if type(cls_obj) is types.ModuleType:
+            return cls_obj.__name__
+        return qualname
+
+    def __init__(self):
+        self._classes = OrderedSet()
+        self._instances = OrderedSet()
+        self._functions = OrderedSet()
```
xmo-odoo: Why ordered?

odoo/addons/base/tests/test_base.py
```diff
@@ -134,6 +136,92 @@ def test_05_safe_eval_forbiddon(self):
             safe_eval("self.__name__", {'self': self}, mode="exec")
 
 
+@tagged('at_install', '-post_install')  # LEGACY at_install
+class TestSafeEvalRuntime(BaseCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        patch_stack = ExitStack()
```
xmo-odoo: An exitstack is useful if you're going to either add CMs / callbacks dynamically, or through the execution of a block. This does neither, it's completely useless, just start your patch and add `patch.stop` to the class cleanup. Or use `startClassPatcher`.

odoo/tools/safe_eval.py
mwath: I don't think it's a good idea to free the tool. In Python 3.12, it does nothing, but in 3.14, it also unregisters all callbacks and events associated with the tool.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+            elts = [[sub_path] if not isinstance(sub_path, tuple) else sub_path for sub_path in path]
+            paths.extend(itertools.product(*elts))
+
+        re_paths = []
+        for path in paths:
+            assert '*' not in path or path.index('*') == len(path) - 1, "'*' can appear at most once at the end of the path"
+            parts = ['.*' if part == '*' else re.escape(part) for part in path]
+            re_paths.append(r'\.'.join(parts))
+
+        return re.compile('|'.join(re_paths)) if re_paths else self.RE_NOTHING
```
xmo-odoo: That might actually be undesirable, python's regex engine is basically a bytecode interpreter, and while it does do some optimisations like common prefix extraction that's easily defeated, so e.g. `odoo.tools|odoo.http` will compile to something like

    if not input.startswith('odoo.'):
        fail
    if input.startwith('tools', 5):
        success
    if input.startswith('tools', 5):
        success

as soon as you introduce a non-matching prefix this falls over e.g. `odoo.tools|odoo.http|psycopg2` becomes something like

    if input.startswith('odoo.tools'):
        success
    if input.startswith('odoo.http'):
        success
    if input.startswith('psycopg2'):
        success

Since the input is pretty much already grouped by prefix and those tend to overlap massively, it would probably be a good idea to build a trie, and then use that to generate nested alternations, because `(odoo.tools|odoo.http)|psycopg2` does compile to something along the lines of

    if input.startswith('odoo.'):
        if input.startswith('tools', 5):
            success
        if input.startswith('http', 5):
            success
        failure
    if input.startswith('psycopg2'):
        success

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+    def check_module(self, obj):
+        raise UnsafeObjectError(obj)
+
+
+SafeTransformer = _SafeTransformer()
+SafeEncoder = _SafeEncoder()
+SafeWhitelist = _SafeWhitelist()
+
+
+if UNSAFE_POLICY:
```
xmo-odoo: Is there a reason for `SafeWhitelist` to just be straight up broken if `UNSAFE_POLICY` is not enabled at load time?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+                pass
+            func = CALL_ID(decorator_B, func)
+            func = CALL_ID(decorator_A, func)
+        """
+        self.generic_visit(node)
+        stmts = [node]
+        while node.decorator_list and (decorator := node.decorator_list.pop()):
+            call = ast.Call(
+                func=ast.Name(id=CALL_ID, ctx=ast.Load()),
+                args=[decorator, ast.Name(id=node.name, ctx=ast.Load())],
```
xmo-odoo: The decorator nodes should probably be visited as well, as they can be calls, with parameters which are themselves calls, e.g.

    @foo.bar(baz(), qux())
    def func():
        ...

odoo/tools/safe_eval.py
```diff
@@ -484,3 +541,419 @@ def __repr__(self):
...
+# - SafeTransformer:
+#       modifies the code to use the `__CALL__` function for each
+#       function call
+# - SafeEncoder:
+#       attempts to serialize the objects used during a call
+# - SafeWhitelist:
+#       used during serialization if the objects need to be verified
+
+
+def __CALL__(callee, /, *args, **kwargs):
```
xmo-odoo:
```suggestion
def __CALL__(callee, *args, **kwargs):
```

xmo-odoo (commented):
>  The presence of the JSON C extension library is required (_json) for performance reasons.

Given this is an optimisation and the feature can be disabled from the env (should probably also be a config item I'd say) forcefully disabling the feature seems unnecessary?

odoo/tools/safe_eval.py
thle-odoo: Yes, we don't want this exception to be catched with a blind except (to be certain that we can bubble it up to `safe_eval`):
```py
try:
	... unsafe code ...
except Exception as e:
	pass
... continue code ...
```
Therefore, using a `BaseException` allows us to avoid this.

Note:
It is currently possible to use bare excepts in `safe_eval` (bad practice), something that will be improved in the next iteration. (I add a TODO for this)

odoo/tools/safe_eval.py
thle-odoo: The goal is to work with a single instance (singleton). This class is not necessary. I change it to:
```py
@functools.lru_cache(1)  # @functools.cache
def safe_encoder():
    return _SafeEncoder()

...
```

Note:
We use this pattern to create a singleton with the `session_store` for example.

odoo/tools/safe_eval.py
thle-odoo: Yes, thank you for pointing that out. I remove it. If we end up using an existing TOOL, we will get an error (and we don't ignore anything).

odoo/tools/safe_eval.py
thle-odoo: _Not really_

This monitoring is called for each "call" (which also includes instructions other than the CALL opcode, such as CALL_FUNCTION_EX).

At this point, the purpose of this monitoring function is to ensure that our call function (initially placed in the evaluation context builtins) has not been overridden[^1]. This ensures that we will always use our verification call function.

> So this is only here to try and flag call sites we might have failed to instrument?

Even if we did not succeed in instrumenting a call, the condition `callee is __CALL__` will fail. But we still make sure that the function is present in the evaluation context.

[^1]: for example, in the case of nested `safe_eval`

odoo/tools/safe_eval.py
thle-odoo: I am going to rename this.

odoo/tools/safe_eval.py
thle-odoo: The `/` is necessary to pass a keyword argument whose name is `callee`.
```py
>>> def func(callee, *args, **kwargs):
...     print(callee, args, kwargs)
... 
>>> func('foo', callee='bar')
  File "<stdin>", line 1, in <module>
TypeError: func() got multiple values for argument 'callee'
```
```py
>>> def func(callee, /, *args, **kwargs):
...     print(callee, args, kwargs)
... 
>>> func('foo', callee='bar')
foo () {'callee': 'bar'}
```

odoo/tools/safe_eval.py
thle-odoo: That is not necessary. It is just useful for debugging the whitelist. I will change that to `set`.

odoo/tools/safe_eval.py
thle-odoo: > Classes are not subclasses of type

Correct, but the condition uses the type of the class and not the class itself.

---

The purpose of this condition is to determine whether an object is a class or an instance based on the type of the object (we will never have a `TypeError`).

With:
```
>>> class MetaC(type): ...
>>> class C(metaclass=MetaC): ...
>>> c = C()
```
```
>>> issubclass(type(c), type)  # detect an instance (`False`)
False
>>> issubclass(type(C), type)  # detect a class (`True`)
True
>>> issubclass(type(MetaC), type)  # detect a class (`True`)
True
```

That said, as I write this comment, I realize that we can simply use the original object and test whether it is an instance of `type`, since classes are instances of `type`.
```
>>> isinstance(c, type)
False
>>> isinstance(C, type)
True
>>> isinstance(MetaC, type)
True
>>> isinstance(type, type)
True
```
This will allow the condition to be better understood.

The condition will become:
```py
if isinstance(obj, type):
    default_ = self._check_class
else:
    default_ = self._check_instance
```

odoo/tools/safe_eval.py
thle-odoo: > it would probably be a good idea to build a trie

Thanks for the explanation and the proposal for improvement.
I modified the `_re_compile` function so that it generates a trie and then builds the regex based on it.

```py
def _re_compile(self, trusted: set) -> typing.Pattern:
    if not trusted:
        self.RE_NOTHING

    # Make the trie
    trie = {}
    for path in trusted:
        nodes = [trie]
        for part in path:
            part = part if isinstance(part, tuple) else (part,)
            next_nodes = []
            for node in nodes:
                for p in part:
                    child = node.setdefault(p, {})
                    next_nodes.append(child)
            nodes = next_nodes

    # Convert the trie to regex
    def _trie_to_regex(trie):
        if not trie:
            return ''

        re_parts = []
        for key, subtree in trie.items():
            subpattern = '.+' if key == '*' else re.escape(key)
            if child_pattern := _trie_to_regex(subtree):
                re_parts.append(subpattern + r'\.' + child_pattern)
            else:
                re_parts.append(subpattern)

        # non-capturing group only if multiple alternatives
        if len(re_parts) == 1:
            return re_parts[0]
        return '(?:' + '|'.join(re_parts) + ')'

    return re.compile(_trie_to_regex(trie))
```

odoo/tools/safe_eval.py
thle-odoo: `UNSAFE_POLICY` took into account the Python version (in the history) because `sys.monitoring` is only available from version 3.12. It was therefore not necessary to define the elements under this condition.

This is no longer the case, as the minimum version is currently 3.12. I am therefore removing this condition.

odoo/tools/translate.py
```diff
@@ -573,6 +563,8 @@ def _get_translation_source(stack_level: int, module: str = '', lang: str = '',
         frame = inspect.currentframe()
         for _index in range(stack_level + 1):
             frame = frame.f_back
+        if frame.f_code is odoo.tools.safe_eval.safe_call.__code__:
```
kmagusiak: Should it be inside the for loop?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,449 @@ def __repr__(self):
...
+def safe_whitelist():
+    whitelist = _SafeWhitelist()
+    _initialize_safe_whitelist(whitelist)
+    return whitelist
+
+
+def safe_call(callee, /, *args, **kwargs):
+    """ Ensure objects used for the call are safe """
+    try:
+        safe_encoder().encode((callee, args, kwargs))
```
kmagusiak: Using calls here instead of a global is around twice slower and this is in hot paths. I  understand that you need some public API, but you should manage the variable as a global.
```
import functools

@functools.cache
def x():
    return 5
def tx():
    a = x() + 1

y = 5
def ty():
    global y
    if y is None:
        y = 5
    a = y + 1

In [7]: %timeit tx()
38.3 ns ± 0.546 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit ty()
22.1 ns ± 0.334 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

odoo/tools/translate.py
thle-odoo: Indeed, if we want to be sure to cover all cases (I am moving this into the loop).

odoo/tools/safe_eval.py
thle-odoo: I will create global variables for use within the `safe_eval.py` module.
```py
_safe_transformer = safe_transformer()
_safe_encoder = safe_encoder()
_safe_whitelist = safe_whitelist()
```

odoo/tools/safe_eval.py
xmo-odoo: > > Classes are not subclasses of type
> 
> Correct, but the condition uses the type of the class and not the class itself.

Ah right, sorry about that.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,456 @@ def __repr__(self):
...
+
+
+@functools.cache
+def safe_whitelist():
+    whitelist = _SafeWhitelist()
+    return whitelist
+
+
+# Make global objects for this module for performance
+_safe_transformer = safe_transformer()
```
kmagusiak: Since you instantiate here, what's the value of having the functions with `functools.cache`?

thle-odoo: None, I just kept the function as an API. So I removed those functions and kept the instance that is global as suggested.

odoo/tools/safe_eval.py
```diff
@@ -345,6 +403,7 @@ def expr_eval(expr):
     'zip': zip,
     'Exception': Exception,
 }
+_BUILTINS_QWEB = dict(_BUILTINS)
```
kmagusiak: Where is it used?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +540,472 @@ def __repr__(self):
...
+
+        # If the safe expression generated is evaluated using `safe_eval`,
+        # there will be double wrapping: `CALL_ID(CALL_ID(...))`.
+        # Example: `_guess_qweb_variables.qweb_like_eval`
+
+        tree = ast.parse(expr.strip(), mode='eval')
+        tree = self.visit(tree)
+        return ast.unparse(tree)
+
+    # TODO: Prevent the use of bare excepts by converting them to blind excepts
```
kmagusiak: Nitpick: Remove the note, it will be another PR and not necessarily changed in this exact place.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +540,472 @@ def __repr__(self):
...
+class _SafeTransformer(ast.NodeTransformer):
+
+    # In Qweb, we use the transformer to modify the python expression.
+    # The ast will not be compiled.
+    # The expression will be used later and therefore re-parsed.
+    # It is necessary for the identifiers to be valid.
+
+    CALL_ID = '_CALL'  # Must be a valid identifier
+
+    def make_safe_expr(self, expr: str) -> str:
```
kmagusiak: This is for a specific case, move this code to ir_qweb (avoids overhead of jumping around).

odoo/tools/safe_eval.py
thle-odoo: I forgot to remove that. It is historical, because before QWEB used a different identifier. Now it is consistent (`_CALL`).

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,74 @@
...
+
+
+class UnsafeObjectError(BaseException):
+
+    def __init__(self, obj):
+        super().__init__(
+            'Unsafe object access (type %r): %r (path: %s)' % (
+                type(obj), obj, safe_whitelist.get_full_path(obj),
+            ),
+        )
```
rco-odoo, line 80: f-strings are cleaner and faster
```suggestion
        path = safe_whitelist.get_full_path(obj)
        super().__init__(
            f"Unsafe object access (type {type(obj)!r}): {obj!r} (path: {path})"
        )
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+# - safe_whitelist:
+#       used during serialization if the objects need to be verified
+
+
+class _SafeTransformer(ast.NodeTransformer):
+
+    # In Qweb, we use the transformer to modify the python expression.
+    # The ast will not be compiled.
+    # The expression will be used later and therefore re-parsed.
+    # It is necessary for the identifiers to be valid.
```
rco-odoo:
```suggestion
    # In Qweb, we use the transformer to modify the python expression.
    # The ast will not be compiled, but instead "unparsed" in the transpiled template.
    # The expression will be used later and therefore re-parsed.
    # It is therefore necessary for the identifiers to be valid.
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+
+    CALL_ID = '_CALL'  # Must be a valid identifier
```
rco-odoo: I would be more specific
```suggestion
    CALL_ID = '_SAFE_CALL'  # Must be a valid identifier
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+
+    def visit_Call(self, node):
+        """
+        Transforms:
+            func(*args, **kwargs)
+
+        Into:
+            CALL_ID(func, *args, **kwargs)
```
rco-odoo: Nitpicking: use the expected identifier given by `CALL_ID`
```suggestion
            _CALL(func, *args, **kwargs)
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+
+        Into:
+            def func():
+                pass
+            func = CALL_ID(decorator_B, func)
+            func = CALL_ID(decorator_A, func)
+        """
+        self.generic_visit(node)
+        stmts = [node]
+        while node.decorator_list and (decorator := node.decorator_list.pop()):
```
rco-odoo: Can `decorator` be falsy?  This would be more idiomatic to me:
```suggestion
        while node.decorator_list:
            decorator = node.decorator_list.pop()
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+                        return f'odoo.upgrade.{qualname}'.strip('.')
+
+        if type(cls_obj) is types.ModuleType:
+            return cls_obj.__name__
+        return qualname
+
+    def __init__(self):
+        self._classes = set()
+        self._instances = set()
+        self._functions = set()
```
rco-odoo: Those three collections are only used to generate a regex by _iterating_ on them.
I suggest to use lists instead, because
* set iteration makes the process nondeterministic with no benefits;
* potential duplicates are merged when generating the regex.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+
+    def add_class(self, value: tuple) -> None:
```
rco-odoo: This is clearly an implementation-oriented API, and not a caller-oriented API.  I see plenty of calls like `safe_whitelist.add_class(('odoo', 'orm', 'commands', 'Command'))` where the double parentheses don't have any purpose, except for the implementation.

Why not passing one string like `safe_whitelist.add_class('odoo.orm.commands.Command')` ?  This would make the calls way more readable, and those calls are not performance-critical IMHO.  It would also simplify the trie construction, by the way.

The only feature you lose is the ability to match multiple items in a path, like in `('odoo', 'tools', 'json', ('_ScriptSafe', 'JSON'))`.  I personally don't mind making it more explicit like `'odoo.tools.json._ScriptSafe'` and `'odoo.tools.json.JSON'`.  I can write a `for` loop to make multiple calls if necessary.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+
+_BUILTINS[safe_transformer.CALL_ID] = safe_call
+
+EVENTS = sys.monitoring.events
+TOOL_ID = 4  # Not prevent other tools from using "official IDs"
+sys.monitoring.use_tool_id(TOOL_ID, 'ODOO_UNSAFE_POLICY_TOOL')
+sys.monitoring.register_callback(TOOL_ID, EVENTS.CALL, monitoring_call)
+
+
+def monitoring_codeobj(code):
```
rco-odoo:
```suggestion
def add_monitoring(code):
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +539,458 @@ def __repr__(self):
...
+        # Convert the trie to regex
+        def _trie_to_regex(trie):
+            if not trie:
+                return ''
+
+            re_parts = []
+            for key, subtree in trie.items():
+                subpattern = '.+' if key == '*' else re.escape(key)
+                if child_pattern := _trie_to_regex(subtree):
+                    re_parts.append(subpattern + r'\.' + child_pattern)
```
rco-odoo:
```suggestion
                    re_parts.append(f"{subpattern}.{child_pattern}")
```

rco-odoo (commented):
This is way much better than last time I reviewed it!  I only have a few remarks left, but nothing critical.

odoo/tools/safe_eval.py
thle-odoo: I applied the suggestion to pass a string as an argument rather than an iterable. This is easier to read and simplifies the construction of the trie, as you mentioned.

However, I adapted the `_initialize_safe_whitelist` function so that it is not too "difficult" to read. Almost the entire whitelist is managed within this function.

For example, I find this more readable:
```py
...
'odoo.exception': {
      'classes': ['AccessDenied', 'AccessError', 'MissingError', 'UserError', 'ValidationError'],
  },
...
```
than:
```py
...
exception_path = 'odoo.exception'
for obj in ('AccessDenied', 'AccessError', 'MissingError', 'UserError', 'ValidationError'):
    safe_whitelist.add_class(f'{exception_path}.{obj}')
...
```

Note:
This choice was made after adapting the entire list and realizing that it was more complicated to identify the groups. For example, answering the question: "Which objects from module X are allowed ?"

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,506 @@ def __repr__(self):
...
+                node = node.setdefault(part, {})
+
+        # Convert the trie to regex
+        def _trie_to_regex(trie):
+            if not trie:
+                return ''
+
+            re_parts = []
+            for key, subtree in trie.items():
+                subpattern = '.+' if key == '*' else re.escape(key)
```
antonrom1, line 879: Technically, if subtree is non-empty shouldn't you do `[^.+]` instead of `.+`?
But it doesn't seem useful for now, so how about just banning non-empty subtrees under wildcards?

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,506 @@ def __repr__(self):
...
+def safe_call(callee, /, *args, **kwargs):
+    """ Ensure objects used for the call are safe """
+    try:
+        safe_encoder.encode((callee, args, kwargs))
+    except UnsafeObjectError as e:
+        _logger_runtime.warning(e)
+        if unsafe_policy() is UnsafePolicy.RAISE:
+            raise
+        if unsafe_policy() is UnsafePolicy.TERMINATE:
+            os.kill(os.getpid(), signal.SIGKILL)
```
antonrom1: [`os._exit(1)`](https://docs.python.org/3/library/os.html#os._exit) is more portable for hard termination

odoo/tools/safe_eval.py
```diff
@@ -14,25 +14,73 @@
...
+    policy = UnsafePolicy[config['unsafe_policy'].upper()]
+
+    if policy and c_make_encoder is None:
+        _logger_runtime.warning(
+            "Unsafe object policy '%s' is configured, but it is not supported on this runtime "
+            "(missing native encoder). The policy will be ignored.",
+            policy,
+        )
+        policy = UnsafePolicy.DISABLE
+
```
antonrom1, line 70: It makes no sense to log IntEnum, you can just log the configuration name directly no?
```suggestion
    policy_name = config['unsafe_policy']
    policy = UnsafePolicy[policy_name.upper()]

    if policy and c_make_encoder is None:
        _logger_runtime.warning(
            "Unsafe object policy '%s' is configured, but it is not supported on this runtime "
            "(missing native encoder). The policy will be ignored.",
            policy_name,
        )
        policy = UnsafePolicy.DISABLE

```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,506 @@ def __repr__(self):
...
+            @decorator_B
+            def func():
+                pass
+
+        Into:
+            def func():
+                pass
+            func = _SAFE_CALL(decorator_B, func)
+            func = _SAFE_CALL(decorator_A, func)
+        """
```
antonrom1, line 601: Your implementation preserves application order (f = A(B(f))), but it reverses evaluation order of decorators

In python, the evaluation should happen top to bottom. A small example:

```
In [1]: def dec(n=1):
   ...:     print(f"decorator {n}")
   ...:     return lambda fun: fun
   ...:

In [2]: @dec(1)
   ...: @dec(2)
   ...: def bob():
   ...:     pass
   ...:
decorator 1
decorator 2
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,506 @@ def __repr__(self):
...
+
+    def add_class(self, value: str) -> None:
+        self._classes.append(value)
+        self.add_instance(value)
+        vars(self).pop('_re_class', None)
+
+    def add_instance(self, value: str) -> None:
+        self._instances.append(value)
+        # Trust function of the class/instance implicitly
+        value = value.rstrip('.*') + '.*'
```
antonrom1: I think you meant this. rstrip removes any combination of . and *

```suggestion
        if not value.endswith('.*'):
          value = value + '.*'    
```

But if it's to just trust the methods of the class, couldn't you do:

```suggestion
        if not value.endswith('.*'):
          value = value + '(\.[^.]+)?'    
```

odoo/tools/safe_eval.py
thle-odoo: Very good point, thanks!

To match Python's behavior, we need to evaluate decorators top-to-bottom and apply wrapping bottom-to-up.

This must be done in two steps.

Before the fix the code you provide:
```py
def dec(n=1):
    _SAFE_CALL(print, f'decorator {n}')
    return lambda fun: fun

def bob():
    pass
bob = _SAFE_CALL(_SAFE_CALL(dec, 2), bob)
bob = _SAFE_CALL(_SAFE_CALL(dec, 1), bob)
```
After the fix:
```py
def dec(n=1):
    _SAFE_CALL(print, f'decorator {n}')
    return lambda fun: fun
_deco_bob_0 = _SAFE_CALL(dec, 1)
_deco_bob_1 = _SAFE_CALL(dec, 2)

def bob():
    pass
bob = _SAFE_CALL(_deco_bob_1, bob)
bob = _SAFE_CALL(_deco_bob_0, bob)
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,539 @@ def __repr__(self):
...
+
+
+# Runtime verification is performed using three objects:
+# - safe_transformer:
+#       modifies the code to use the `safe_call` function for each
+#       function call
+# - safe_encoder:
+#       attempts to serialize the objects used during a call
+# - safe_whitelist:
+#       used during serialization if the objects need to be verified
```
kmagusiak: Serialization is an implementation detail.
```suggestion
#       verifies the objects used during a call
# - safe_whitelist:
#       used during verifications
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +538,539 @@ def __repr__(self):
...
+
+
+class _SafeTransformer(ast.NodeTransformer):
+
+    # In Qweb, we use the transformer to modify the python expression.
+    # The ast will not be compiled, but instead "unparsed" in the transpiled template.
+    # The expression will be used later and therefore re-parsed.
+    # It is therefore necessary for the identifiers to be valid.
+
+    CALL_ID = '_SAFE_CALL'
```
kmagusiak: Since these need to be valid identifiers, shouldn't they be more unique/namespaced, for example, by prefixing them with a common string. "_seval_call"/"_seval_deco_%s_%s"?

odoo/tools/safe_eval.py
thle-odoo: I agree with your first suggestion.

The second one cannot be used because at this point in the code, we are not building the regex but using the `add_instance(self, value: str)` API to whitelist one or more objects (example: 'a.b.c.*').

odoo/tools/safe_eval.py
thle-odoo: > shouldn't they be more unique/namespaced

That is a good idea (although not necessary).

I opted for:
```py
CALL_ID = '_seval_call'
DECORATOR_ID = '_seval_deco_%s_%s'  # Removed
```

odoo/tools/safe_eval.py
thle-odoo: We can do better:
```py
def A(x): ...
def B(y): ...

@A(x)
@B(y)
def func: ...
```
can become:
```py
func = A(x)(B(y)(func))
```

Your example will becomes:
```py
def dec(n=1):
    _seval_call(print, f'decorator {n}')
    return lambda fun: fun

def bob():
    pass
bob = _seval_call(_seval_call(dec, 1), _seval_call(_seval_call(dec, 2), bob))
```

This preserves the evaluation order and wrapping order.

_`visit_FunctionDef` updated_

odoo/tools/safe_eval.py
thle-odoo: > how about just banning non-empty subtrees under wildcards?

I add an assert for this case.

> if subtree is non-empty shouldn't you do [^.+] instead of .+?

I want to be able to validate all paths (`aaa.bbb`, but also `aaa.bbb.ccc`)

_(This can be improved in a future iteration if necessary.)_

odoo/tools/safe_eval.py
thle-odoo: I use `fr'{subpattern}\.{child_pattern}'` to conserve the match on the literal dot character.

antonrom1 (approved):

odoo/addons/test_tools/tests/test_safe_eval.py
```diff
@@ -129,3 +136,115 @@ def test_05_safe_eval_forbiddon(self):
...
+
+    def test_override_call(self):
+        expr = """
+            # Override in locals
+            _seval_call = lambda callee, *args, **kwargs: callee(*args, **kwargs)
+            UnsafeClass()
+        """
+        with self.assertRaisesRegex(ValueError, '^AssertionError'):
+            safe_eval(dedent(expr), self.unsafe_context, mode='exec')
+        # Note without the assert protection, we get  `RecursionError`,
```
rco-odoo:
```suggestion
        # Note that without the assert protection, we get a `RecursionError`,
```

odoo/tools/safe_eval.py
rco-odoo: @thle-odoo please add a test case to verify that decorators are applied in the expected order.  All you need is to `safe_eval` something like
```py
log = []

def outer(func):
    log.append('outer decorate')
    return lambda *args, **kwargs: log.append('outer call') or func(*args, **kwargs)

def inner(func):
    log.append('inner decorate')
    return lambda *args, **kwargs: log.append('inner call') or func(*args, **kwargs)

@outer
@inner
def func(arg):
    return arg

func(42)
```
and check that `log` equals `['inner decorate', 'outer decorate', 'outer call', 'inner call']`.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+
+
+# Runtime verification is performed using three objects:
+# - safe_transformer: modifies the code to use the `safe_call` function
+#                     for each function call
+# - safe_encoder: verifies the objects used during a call
+# - safe_whitelist: used during verifications
+
+
+class _SafeTransformer(ast.NodeTransformer):
```
rco-odoo, line 558:
```suggestion
class _SafeTransformer(ast.NodeTransformer):
    """
    Code transformer that wraps calls with :func:`_seval_call` in order to check
    that the arguments are "safe".
    """
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+        ast.copy_location(assign, node)
+        ast.copy_location(assign.targets[0], node)
+
+        return [node, assign]
+
+
+encoder_ctx = contextvars.ContextVar('encoder_ctx')
+
+
+class _SafeEncoder:
```
rco-odoo:
```suggestion
class _SafeChecker:
    """
    Callable object that checks whether a value is considered "safe".
    It does so by using on a JSON encoder, for which one can add type-specific hooks.
    An empty hook simply considers the given type as completely safe,
    while other hooks may check a value and return values to be recursively checked.
    """
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+
+    MAPPINGS = frozenset((dict, defaultdict, OrderedDict, types.MappingProxyType))
+    SEQUENCES = frozenset((list, tuple, set, frozenset, OrderedSet))
+
+    def __init__(self):
+        self.hooks: WeakKeyDictionary[type, typing.Callable | None] = WeakKeyDictionary()
```
rco-odoo: This attribute should be private.  Adding hooks should be done with a method, like `safe_checker.add_hook(type, hook)`, where `hook` defaults to `None`.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+            # markers, default, encoder, indent, key_separator, item_separator, sort_keys, skipkeys, allow_nan
+            encoder = c_make_encoder({}, self._default, lambda x: '', None, '', '', False, False, False)
+            encoder_ctx.set(encoder)
+        encoder.markers.clear()
+        return encoder
+
+    def encode(self, obj) -> None:
+        """
+        Perform a safety encoding on the given object to determine if it is
+        considered "safe" according to the rules defined in `SafeWhitelist`.
```
rco-odoo: This _checks_ a value by _encoding_ it.  The purpose is to check the value, and the encoding is a implementation detail.  The name of the method should reflect the purpose.
```suggestion
    def check(self, obj) -> None:
        """
        Check whether the given object is considered "safe" according to the
        rules defined in `SafeWhitelist`.
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+        if type(cls_obj) is types.ModuleType:
+            return cls_obj.__name__
+        return qualname
+
+    def __init__(self):
+        self._classes = list()
+        self._instances = list()
+        self._functions = list()
+
+    def add_class(self, value: str) -> None:
```
rco-odoo: Please add a docstring on the class to describe the three methods' argument, which is a fully qualified name, like `odoo.orm.environment.Environment` (which may contain wildcard `*`).
```suggestion
    def add_class(self, qual_name: str) -> None:
```

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+        'odoo.addons': {
+            'classes': ['*'],  # TODO: Restrict addons
+        },
+        'odoo.upgrade': {
+            'classes': ['*'],
+        },
+        'odoo.orm': {
+            'classes': ['domains.*', 'fields.*', 'fields_selection.*', 'models.*', 'commands.Command'],
+            'instances': ['environments.Environment', 'identifiers.NewId', 'registry.Registry'],
+        },
```
rco-odoo: This dict makes it harder to read than necessary.  What's wrong with explicit code _with no complexity_ like follows?
```py
safe_whitelist.add_class('odoo.orm.domains.*')
safe_whitelist.add_class('odoo.orm.fields.*')
safe_whitelist.add_class('odoo.orm.fields_selection.*')
safe_whitelist.add_class('odoo.orm.models.*')
safe_whitelist.add_class('odoo.orm.commands.Command')
safe_whitelist.add_instance('odoo.orm.environments.Environment')
safe_whitelist.add_instance('odoo.orm.identifiers.NewId')
safe_whitelist.add_instance('odoo.orm.registry.Registry')
```

odoo/tools/safe_eval.py
```diff
@@ -255,12 +300,20 @@ def compile_codeobj(expr: str, /, filename: str = '<unknown>', mode: typing.Lite
     try:
         if mode == 'eval':
             expr = expr.strip()  # eval() does not like leading/trailing whitespace
-        code_obj = compile(expr, filename or '', mode)
+
+        if unsafe_policy():
+            tree = ast.parse(expr, mode=mode)
+            tree = safe_transformer.visit(tree)
```
rco-odoo: `visit` is an implementation detail of the transformer.  I suggest to define
```py
safe_transform = safe_transformer.visit
```
to have the more intuitive
```suggestion
            tree = safe_transform(tree)
```

odoo/orm/model_classes.py
```diff
@@ -227,6 +228,7 @@ def add_to_registry(registry: Registry, model_def: type[BaseModel]) -> type[Base
     for model_name in registry.descendants([name], '_inherit', '_inherits'):
         registry[model_name]._setup_done__ = False
 
+    safe_encoder.hooks[model_cls] = None
```
rco-odoo: Couldn't you add it in the whitelist instead?

odoo/orm/models.py
```diff
@@ -310,6 +310,8 @@ def add_default(name, field):
                     string='Last Updated on', readonly=True))
 
 
+safe_encoder.hooks[MetaModel] = None
```
rco-odoo: Same question here.

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,513 @@ def __repr__(self):
...
+
+
+class _SafeTransformer(ast.NodeTransformer):
+
+    # In Qweb, we use the transformer to modify the python expression.
+    # The ast will not be compiled, but instead "unparsed" in the transpiled template.
+    # The expression will be used later and therefore re-parsed.
+    # It is therefore necessary for the identifiers to be valid.
+
+    CALL_ID = '_seval_call'
```
rco-odoo: There is no tax on characters, they're cheap.  Just make it straightforward.
```suggestion
    CALL_ID = '_safe_eval_call'
```

odoo/orm/model_classes.py
thle-odoo: Personally, I would leave these hooks in this place.
Adding the `None` hook directly means that the regex check is not used.
This is beneficial given the frequency with which recordsets appear in the evaluation context

odoo/orm/models.py
thle-odoo: See https://github.com/odoo/odoo/pull/247604#discussion_r2889564854

odoo/tools/safe_eval.py
thle-odoo: I understand that it is not necessary to structure the whitelist using a dictionary.
I modified it to use the API directly as suggested.

odoo/tools/safe_eval.py
thle-odoo: I added a test that guarantees the execution order and wrapping order of decorators (test name: `test_transform_decorators`).

odoo/tools/config.py
```diff
@@ -260,6 +260,9 @@ def _build_cli(self):
                          help="Comma-separated list of server-wide modules.")
         group.add_option("-D", "--data-dir", dest="data_dir", type='path',  # sensitive default set in _load_default_options
                          help="Directory where to store Odoo data")
+        group.add_option("--unsafe-policy", dest='unsafe_policy', type='choice', my_default='log',
+                         choices=['disable', 'log', 'raise', 'terminate'],
+                         help="Determine policy if an unsafe object is detected during arbitrary code execution.")
```
xmo-odoo:
```suggestion
                         help="policy if an unsafe object is detected during arbitrary code execution.")
```
Would also be a good idea to explain what the policies are, because they're not necessarily *entirely* self explanatory (notably `terminate` immediately killing the current worker).

odoo/tools/safe_eval.py
```diff
@@ -484,3 +537,538 @@ def __repr__(self):
...
+    while other hooks may check a value and return values to be recursively checked.
+    """
+
+    MAPPINGS = frozenset((dict, defaultdict, OrderedDict, types.MappingProxyType))
+    SEQUENCES = frozenset((list, tuple, set, frozenset, OrderedSet))
+
+    def __init__(self):
+        self.__hooks: WeakKeyDictionary[type, typing.Callable | None] = WeakKeyDictionary()
+        # Add hooks
+        [self.add_hook(t, None) for t in _SafeWhitelist.TRUSTED_TYPES]  # Optimization to save time when serializing these types
```
xmo-odoo: Really not a fan of using listcomps for side-effects, especially when you can write
```suggestion
        for t in _SafeWhitelist.TRUSTED_TYPES: self.add_hook(t, None)  # Optimization to save time when serializing these types
```
(possibly with a nofmt/noqa if a formatter is being annoying)

xmo-odoo (commented):
@robodoo override=ci/security

kmagusiak (approved):
@robodoo r+

PR #247606, opened 2026-02-06
[FIX] account: don't block edition of opening date if an opening move is posted

oco-odoo:
To reproduce the issue (in enterprise):
- Go to accounting > configuration
- click on "Review Manually" (under Accounting Import)
- add a balance to 1 or 2 accounts
- Search for the created draft opening move in the journal entries, and post it
- On the dashboard, click on "Tax returns" ====> The opening date is not set in the wizard, and still, it's not editable. You're stuck

The opening date was originally introduced in the setup bar flows, years ago. Today, this field isn't really used anymore, except for creating the returns, and it then does not especially always correspond to the date of the opening move. To solve the issue, we just always allow editing it in the wizard. To keep things comfortable for users who would be used to the previous way it worked, we still modify the date of the opening move when doing so if it's still in draft.


hupo-odoo (approved):
@robodoo r+

PR #247607, opened 2026-02-06
[FIX] peppol_proxy: Peppol - don't recommend technical contact for su…

clbr-odoo:
…pport

Our TechnicalContactUrl is peppol.iap@odoo.com, it should not be exposed to the client as the support contact.
Let's just expose the name of the service like we do above 18.0.

task-none


gawa-odoo (commented):
@robodoo r+

clbr-odoo:
@robodoo fw up to 18.0

Issue #820, opened 2014-06-28
review the category field in ir.module.module


GoodERPJeff:
IMHO, openerp use a strange design on this field, this is a many2one in database, but char in source code (**openerp**.py). this make things un-control. when we list the module group by category , you can easily found there are so many module with wrong category what made the list  hard to understand.

given we have this mass in official package, things got more complex when community modules involved, no clear enough document guide the developer what they should input(should be select ) in **openerp**.py category field. I do like the category list on apps.openerp.com:

Accounting
Human Resources
Localization
Manufacturing
Project
Purchases
Sales
Extra Tools
Warehouse

With clear category, end user and consultant will find the features they need easily. and I guess odoo s.a look modules in same category as a application(used to charge opw users 12 EUR ?) . also, in future version development, odoo s.a and OCA can plan project scope by category first.

But I don't know why modules in openerp repo not follow this category, here I list some module, the category of it,  and my comments, further more than category, I also list some modules I think shouldn't include in official package.    

Category/Name   Technical Name  Comments
Accounting  account_bank_statement_extensions   why this new category name?shouldn't be "Accounting & Finance"
Accounting  account_check_writing   why this new category name?shouldn't be "Accounting & Finance"
Accounting & Finance    account_test    who this module for, when the End user should install this?
Accounting & Finance    l10n_be_coda    this is only for be, why include this in official branch? Also the category should be "Localization",agree?
Accounting & Finance    report_intrastat    this is only for Europ, why include this in official branch? Also the category should be "Localization",agree?
Association membership  we have only 1 module in this category, why not go into "Specific Industry Applications"?
Customer Relationship Management    crm_partner_assign  assign leads to others is a strange logic for me, don't know who use it other then openerp s.a self, is it worth been include in package ?
Dependency  l10n_fr_rib this is for Franch only, why it's included ? Also the category should be "Localization",agree?
Extra Tools lunch   so much joke in community for this, why not move it out?
Hardware Drivers    hw_escpos   this category should be  "Point of Sale",yes ?
Hardware Drivers    hw_posbox_homepage  this category should be  "Point of Sale",yes ?
Hardware Drivers    hw_posbox_upgrade   this category should be  "Point of Sale",yes ?
Hardware Drivers    hw_scale    this category should be  "Point of Sale",yes ?
Hardware Drivers    hw_scanner  this category should be  "Point of Sale",yes ?
Inventory Control   product_extended    why we have a new categoty here, shouldn't be "Warehouse"?
[...]

PR #247615, opened 2026-02-06
[IMP] point_of_sale, pos_safaricom: Add mpesa payment methods

lcantraine:
This pr introduces MPesa payment methods using Safaricom APIs for Kenya.

There are two types of payment methods :

- Mpesa Express: This is implemented with the M-Pesa Express API. It's the Merchant Initiated payment method. The merchant create a payment method M-Pesa Express with Safaricom payment provider. The merchant is allowed to ask for payment through the API with an USSD prompt on the customer phone number. Customer enters the PIN and the API call the callback of the POS to reconcile transaction with order. (https://daraja.safaricom.co.ke/dashboard/apis?api=MpesaExpressSimulate)

- Lipa Na Mpesa: This is implemented with the Customer To Business (C2B) API. This payment method is different from the first one. It is a Merchant-Initiated payment method. The POS registers callback URLs through the register URL. The POS listen on its callbacks. Customer sends a payment through his Safaricom or MPesa application with some informations of the business. When a transaction is catch in the callback, POS displays it. The cashier needs, manually, to reconcile it with the current order. To reconcile, cashier see a popup with every transaction done on this business account number. (https://daraja.safaricom.co.ke/dashboard/apis?api=CustomerToBusiness)

Dynamic QR Code API is also implemented to allow customers to scan the QR Code in the official Safaricom application and send the payment easily without errors of amount or till number. (https://daraja.safaricom.co.ke/dashboard/apis?api=DynamicQRCode)

Note : The problem with this payment method is that there is two steps. The first step is a "Verification" step, which is optional. Businesses needs to manually sends an email to Safaricom to allows them to add this step. This step is useful because it is used to accept or reject a transaction before charging the customer. But, in reality, not all of the businesses ask Safaricom to add this step. So, we don't configure it for the moment. The second step is the "Confirmation" step, before this step, the customer is already charged. The API sends transaction data to the POS, we display it in the POS to allow cashier to reconcile the current order with a transaction.

Requirements : If you want to test this API, https is needed and a valid kenyan (+254) phone number is also needed. But, you can simulate requests in the Daraja Portal. (https://daraja.safaricom.co.ke/dashboard/apis)

task: 5886047

---


addons/pos_safaricom/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+import json
+
+from odoo import http
+from odoo.http import request
+from odoo.tools import verify_hash_signed
+
+
+class SafaricomController(http.Controller):
+
+    @http.route('/pos_safaricom/callback', type='http', auth='public', methods=['POST'], csrf=True)
```
xmo-odoo: Why are you using `csrf=True`? It's the default. Also it seems somewhat dubious as I don't see how the caller could be sending a CSRF token?

addons/pos_safaricom/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+    def safaricom_callback(self, payload):
+        """
+        Handle M-Pesa STK Push callback
+        """
+
+        try:
+            raw_data = request.httprequest.data.decode('utf-8') if request.httprequest.data else ''
```
xmo-odoo: 1. decoding by hand even though `json.loads` is perfectly happy with bytes
2. and `Request.get_data(text=True)` can decode to text
3. falling back to an empty string even though that's already the "worst case" for `data`
4. manually json-decoding even though *both* `odoo.http.Request` and `werkzeug.wrappers.Request` have JSON helpers (respectively `get_json_data()` and `json`/`get_json()`)

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,333 @@
...
+    def create(self, vals_list):
+        """Override create to automatically register URLs for Lipa na M-PESA payment methods"""
+        payment_methods = super().create(vals_list)
+
+        for payment_method in payment_methods:
+            if (payment_method.use_payment_terminal == 'safaricom' and
+                payment_method.safaricom_payment_type == 'lipa_na_mpesa' and
+                payment_method.consumer_key and
+                payment_method.consumer_secret and
+                payment_method.business_short_code):
```
davidmonnom: If `lipa_na_mpesa_register_urls` needs all of these fiedls, the check shouldn't be inside that method ?

addons/pos_safaricom/static/src/app/components/popups/mpesa_transaction_popup.js
```diff
@@ -0,0 +1,80 @@
...
+                    amount: r.amount,
+                    received_at: r.received_at,
+                }))
+                .reverse();
+            this.state.allTransactions = transactions;
+            this.state.transactions = transactions;
+            // Reapply search filter if there's a search query
+            if (this.state.searchQuery) {
+                this.searchInTransactions(this.state.searchQuery);
+            }
```
davidmonnom: You can do a getter

```js
    get transactions() {
        if (!this.state.searchQuery) {
            return this.state.transactions;
        }

        const search = this.state.searchQuery;
        return this.state.transactions.filter(
            (transaction) =>
                transaction.name.toLowerCase().includes(search.toLowerCase()) ||
                transaction.phone.toLowerCase().includes(search.toLowerCase())
        );
    }
```

addons/pos_safaricom/static/src/app/components/popups/mpesa_transaction_popup.xml
```diff
@@ -0,0 +1,68 @@
...
+                        <tr>
+                            <th>Name</th>
+                            <th>Phone</th>
+                            <th>Amount</th>
+                            <th>Arrived at</th>
+                            <th></th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <t t-if="state.transactions.length > 0" t-foreach="state.transactions" t-as="tx" t-key="tx.id">
```
davidmonnom:
```suggestion
                        <t t-if="this.transactions.length > 0" t-foreach="this.transactions" t-as="tx" t-key="tx.id">
```

addons/pos_safaricom/static/src/app/components/popups/mpesa_transaction_popup.js
```diff
@@ -0,0 +1,80 @@
...
+            const transactions = records
+                .map((r) => ({
+                    id: r.id,
+                    name: r.name,
+                    phone: r.number,
+                    amount: r.amount,
+                    received_at: r.received_at,
+                }))
+                .reverse();
+            this.state.allTransactions = transactions;
```
davidmonnom:
```suggestion
```

addons/pos_safaricom/static/src/app/payment_safaricom.js
```diff
@@ -0,0 +1,214 @@
...
+        const order = this.pos.getOrder();
+        const line = order.payment_ids.find((paymentLine) => paymentLine.uuid === uuid);
+
+        if (!this._validatePaymentLine(line)) {
+            return Promise.resolve(false);
+        }
+
+        line.setPaymentStatus("waitingCard");
+
+        const qr_data = {
```
davidmonnom:
```suggestion
        const qrData = {
```

davidmonnom (commented):
A few comments, but otherwise very good!

davidmonnom (approved):

addons/pos_safaricom/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+                if stk_callback:
+                    result_code = stk_callback.get('ResultCode')
+                    result_desc = stk_callback.get('ResultDesc')
+                    checkout_request_id = stk_callback.get('CheckoutRequestID')
+                    merchant_request_id = stk_callback.get('MerchantRequestID')
+                    callback_metadata = stk_callback.get('CallbackMetadata', {})
+
+                    # Find the specific payment method using the signed payload
+                    payment_method = request.env['pos.payment.method'].sudo().search([
+                        ('id', '=', decoded_payload.get('payment_method_id')),
```
xmo-odoo: Why are all these (even the payment method id) optional?

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,327 @@
...
+
+            return {
+                'error': result.get('errorMessage') or result.get('CustomerMessage', 'Payment request failed'),
+            }
+
+        except (requests.exceptions.RequestException, ValueError) as e:
+            return {'error': e}
+
+    def retrieve_payment_status(self, merchant_request_id, checkout_request_id, result_code, result_desc, callback_metadata=None):
+        """Process payment status from callback and notify POS sessions"""
```
xmo-odoo: So this is less retrieving and more formatting, and notifying?

Also why is this callable over RPC? And why is it formatting one way then formatting another, and returning the first, when the sole callsite completely ignores the return value?

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,327 @@
...
+        active_sessions = self.env['pos.session'].search([
+            ('state', '=', 'opened'),
+            ('config_id.payment_method_ids', 'in', self.ids),
+        ])
+
+        for pos_session in active_sessions:
+            pos_session.config_id._notify('SAFARICOM_LATEST_RESPONSE', {
+                'merchant_request_id': payment_data.get('merchant_request_id'),
+                'checkout_request_id': payment_data.get('checkout_request_id'),
+                'success': payment_data.get('success', False),
```
xmo-odoo: We just set these 30 lines above, why where would they have disappeared to that they require a conditional fetch & default value?

Also why are we not just sending the `payment_data` directly? This seems to do a pretty arbitrary filtering e.g. send the transaction id but not the date, send the result desc but not the code, ... And per the above since the `payment_data` is not being used anyway why are we bothering with values nobody cares about?

addons/pos_safaricom/controllers/main.py
```diff
@@ -0,0 +1,114 @@
...
+                            checkout_request_id,
+                            result_code,
+                            result_desc,
+                            callback_metadata,
+                        )
+
+            return request.make_response(json.dumps({
+                "ResultCode": "0",
+                "ResultDesc": "Accepted",
+            }), [('Content-Type', 'application/json')])
```
xmo-odoo: If we didn't find a callback info or a payment id then accepted?

And why not use `make_json_response`?

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,327 @@
...
+        if self.safaricom_test_mode:
+            return 'https://sandbox.safaricom.co.ke/mpesa/c2b/v2/registerurl'
+        return 'https://api.safaricom.co.ke/mpesa/c2b/v2/registerurl'
+
+    def _get_qr_code_endpoint(self):
+        if self.safaricom_test_mode:
+            return 'https://sandbox.safaricom.co.ke/mpesa/qrcode/v1/generate'
+        return 'https://api.safaricom.co.ke/mpesa/qrcode/v1/generate'
+
+    def bearer_token(self):
```
xmo-odoo: I would question why most of these methods are on a model, but beyond that for what god-forsaken purpose is this callable over RPC?

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,327 @@
...
+        return payment_data
+
+    def lipa_na_mpesa_register_urls(self):
+        """
+        Register C2B URLs for Lipa na M-PESA
+        The ValidationURL is the URL that will be called to validate the payment before charges the customer if business has activated it.
+        The ConfirmationURL is the URL that will be called when the payment is successful or unsuccessful.
+        The ResponseType is set to Completed to charge the customer even if the ValidationURL returns an error or is unreachable.
+
+        This is a one-time API call. URLs should only be registered once unless force_register is True.
```
xmo-odoo, line 207: Where is the `force_register` this talks about? And if this should only be called once, why does it not check if it's already been called? And what happens if it's called twice?

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,327 @@
...
+            )
+            result = response.json()
+
+            if result.get('ResponseCode') != '00000000':
+                raise UserError(_("Failed to register URLs"))
+
+        except (requests.exceptions.RequestException, ValueError):
+            raise UserError(_("Failed to register URLs. Check your credentials and try again."))
+
+    def create_payment_transaction(self, trans_id, trans_amount, msisdn, name):
```
xmo-odoo: Once again, why is this callable over RPC. Repeat for every method which is callable over RPC for seemingly no reason.

addons/pos_safaricom/controllers/main.py
lcantraine: You're right. I don't need it.

addons/pos_safaricom/controllers/main.py
lcantraine: I updated with get_json_data().

addons/pos_safaricom/controllers/main.py
lcantraine: You're right. It doesn't supposed to be optional. I refactored a bit.

addons/pos_safaricom/models/pos_payment_method.py
lcantraine: I updated the method name with _notify_stk_callback() and did some changes.

addons/pos_safaricom/models/pos_payment_method.py
lcantraine: Yes, that doesn't make much sense. I've modified it to send only the values required on the front end.

addons/pos_safaricom/controllers/main.py
lcantraine: Actually, I didn't know that existed. I've modified.

addons/pos_safaricom/models/pos_payment_method.py
lcantraine: I forgot to update my comment the force_register does not exist anymore.

But fyi, URLs can be registered once (when creating the payment method). However, users can delete callback URLs from the Safaricom portal and register them again from the payment method form view. 

If it's called twice, nothing happened. Unless he deleted them, it will re-register them.

addons/pos_safaricom/models/pos_payment_method.py
lcantraine: I updated all the methods which are normally not callable over RPC.

xmo-odoo (commented):
@robodoo override=ci/security

addons/point_of_sale/models/res_config_settings.py
```diff
@@ -38,6 +38,7 @@ def _default_pos_config(self):
     module_pos_pine_labs = fields.Boolean(string="Pine Labs Payment Terminal", help="The transactions are processed by Pine Labs. Set your Pine Labs credentials on the related payment method.")
     module_pos_qfpay = fields.Boolean(string="QFPay Payment Terminal", help="The transactions are processed by QFPay. Set your QFPay credentials on the related payment method.")
     module_pos_pricer = fields.Boolean(string="Pricer electronic price tags", help="Display the price of your products through electronic price tags")
+    module_pos_safaricom = fields.Boolean(string="Safaricom Payment Terminal", help="The transactions are processed by Safaricom. Set your Safaricom credentials on the related payment method.")
```
KangOl: This is not stable.
Won't work without a `-u`
```suggestion
```

lcantraine: you're right. thanks.

C3POdoo:
Upgrade exception [#820](https://runbot.odoo.com/odoo/runbot.upgrade.exception/820?menu_id=405) added.

Waiting the forward-port of this PR, the exception will be applied on all builds, and create an inconsistent state for migrations.

> [!IMPORTANT]
> **Please forward port this PR ASAP up to master without change, preferably before the end of the week.**

PRs adding fields/models should be merged at the beginning of the week and all the forward ports should be merged in the same week.

If you need to apply any change before it reaches master, please notify runbot team.

Details:
```
module:pos_safaricom
```
cc @odoo/upgrade

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,310 @@
...
+            'number': msisdn,
+            'amount': int(float(trans_amount)),
+            'received_at': datetime.now(),
+        })
+
+        for pos_config in self.config_ids:
+            pos_config._notify('NEW_LIPA_NA_MPESA_TRANSACTION', {})
+
+    def mark_transaction_used(self, transaction_id):
+        """Mark a transaction as used by deleting it or updating its status"""
```
davidmonnom: The comment is wrong, we don't update its status.

addons/pos_safaricom/models/pos_payment_method.py
```diff
@@ -0,0 +1,310 @@
...
+    def _get_bearer_token(self):
+        """Get OAuth access token"""
+        self.ensure_one()
+
+        if not self.consumer_key or not self.consumer_secret:
+            raise UserError(_("Consumer Key and Consumer Secret are required for Safaricom M-Pesa"))
+
+        try:
+            consumer_key = self.consumer_key.strip() if self.consumer_key else ''
+            consumer_secret = self.consumer_secret.strip() if self.consumer_secret else ''
```
davidmonnom: Useless condition, is checked above.

davidmonnom (approved):

davidmonnom (approved):

davidmonnom:
@robodoo fw=skipmerge

davidmonnom (approved):
@robodoo r+

PR #247616, opened 2026-02-06
[IMP] account,analytic: analytic confidence

jedel-odoo:
Reordering the analytic distribution field.

task-5887978

---


william-andre (commented):
Very limited benefits for a small risk of breaking stable things.
Not worth IMO.

PR #239813, opened 2025-12-12
[FIX] point_of_sale: incorrect tax names shown in info popup

meji-odoo:
Before this commit:
---
- The POS product info popup displayed only the first tax from `tax_details`, which meant multiple applied taxes were not visible.
- The logic fetched a single tax name instead of all tax applied to the product.

After this commit:
---
- The POS store sends a full list of tax names instead of a single name.
- The popup template displays all tax names, comma-separated.

task-5406826

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247618, opened 2026-02-06
[FIX] point_of_sale: correct VAT name in Info popup

MAGREMENT:
How to reproduce;
- Use a company from a country with a different vat_label (Mexico for example)
- Go to the frontend of POS
- Select a product > ... > Info

The problem:
Under the "Financials" section, "RFC" is displayed instead of "VAT" or "IVA". RFC is the tax identification number, not the tax name

Why:
This commit (https://github.com/odoo/odoo/commit/4fa9f9b849016f312efcb73f9a76b223e429aec0) introduced the usage of vat_label to display the tax name in a dynamic way. However, the vat_label field is used to store the tax identification number, not the tax name

opw-5915087

---

### DIFF
--- a/addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.js
+++ b/addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.js
@@ -35,6 +35,9 @@ export class ProductInfoPopup extends Component {
         });
     }
     get vatLabel() {
-        return this.pos.company.country_id.vat_label || _t("VAT");
+        return _t("VAT:");
+    }
+    get totalVatLabel() {
+        return _t("Total VAT:");
     }
 }
--- a/addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.xml
+++ b/addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.xml
@@ -76,7 +76,7 @@
                     <div class="grid gap-2">
                         <label class="fw-bold">Price excl. Tax:</label>
                         <span><t t-esc="env.utils.formatCurrency(props.info.productInfo.all_prices.price_without_tax)"/></span>
-                        <label class="fw-bold vat-label"><t t-out="vatLabel"/>:</label>
+                        <label class="fw-bold vat-label"><t t-esc="vatLabel"/></label>
                         <span class="vat-value">
                             <t t-esc="props.info.taxAmount"/>
                             (<t t-esc="props.info.taxName"/>)
@@ -102,7 +102,7 @@
                     <div class="grid gap-2">
                         <label class="fw-bold">Total Price excl. Tax:</label>
                         <span t-esc="props.info.orderPriceWithoutTaxCurrency"/>
-                        <label class="fw-bold">Total VAT:</label>
+                        <label class="fw-bold"><t t-esc="totalVatLabel"/></label>
                         <span t-esc="props.info.orderTaxTotalCurrency"/>
                         <label class="fw-bold">Total Price incl. Tax:</label>
                         <span t-esc="props.info.orderPriceWithTaxCurrency"/>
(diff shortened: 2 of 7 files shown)


addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.js
```diff
@@ -34,7 +33,4 @@ export class ProductInfoPopup extends Component {
             is_favorite: !this.props.productTemplate.is_favorite,
         });
     }
-    get vatLabel() {
-        return this.pos.company.country_id.vat_label || _t("VAT");
-    }
```
sbel-odoo, line 39:
```suggestion
    get vatLabel(){
        return _t("VAT");
    }
```

addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.xml
```diff
@@ -76,7 +76,7 @@
                     <div class="grid gap-2">
                         <label class="fw-bold">Price excl. Tax:</label>
                         <span><t t-esc="env.utils.formatCurrency(props.info.productInfo.all_prices.price_without_tax)"/></span>
-                        <label class="fw-bold vat-label"><t t-out="vatLabel"/>:</label>
+                        <label class="fw-bold vat-label">VAT:</label>
```
sbel-odoo: Keep previous code.

sbel-odoo (commented):
Hello @MAGREMENT 

I'm good with the end result of the fix 👍 I won't ask for a test here since it seems very trivial but in case the next reviewer asks for one, you could imagine changing this test `test_product_long_press` and change the vat label of the company's country. That test would the fail without your fix and pass with.

However I think we should modify the fix slightly to keep in mind the original task where it was changed (see review comments). And make an override in the `l10n_in_pos` where it checks if `this.pos.country_id` is IN and either return GST or the super.

Also don't forget to add opw- at the end of the commit message 😉 

I'll leave a message for the next reviewer to avoid them having the same confusion I had here.

At first I had doubts regarding this change because it seemed like the behavior was intentional. See this pr: https://github.com/odoo/odoo/pull/239813 and the associated task: 5406826

It was changed on the associated task with the explanation that in India VAT is actually called GST. GST is indeed the equivalent of VAT in India but the field vat_label on the country seems to be more used in Odoo for partner vat labels. The way I see it would be that the vat_label for a be company would represent "Numéro d'identification TVA" and not the words "TVA" that we see for products. And in the case of mexico the vat_label is RFC which can be seen as our "Numéro d'identification TVA".

sbel-odoo (approved):
Looks good to me 👍

addons/point_of_sale/static/src/app/components/popups/product_info_popup/product_info_popup.js
```diff
@@ -35,6 +35,9 @@ export class ProductInfoPopup extends Component {
         });
     }
     get vatLabel() {
-        return this.pos.company.country_id.vat_label || _t("VAT");
+        return _t("VAT:");
```
davidmonnom, line 38: `this.pos.company.country_id.vat_label` isn't used ?

MAGREMENT: No, because vat_label is used to store the label for the tax identification number, which is called "VAT number" in the EU, "Numéro d'identification TVA" in Belgium like @sbel-odoo said, or "RFC" in Mexico. In this case, we want the translation for the word "VAT", which is the name of the tax

davidmonnom: Ok it should be nice to have some tests for this.

Also in localization

MAGREMENT: I edited the test added by the other commit and added a test for the India localization

davidmonnom (approved):
@robodoo r+

MAGREMENT:
@robodoo r+

ushyme:
@robodoo delegate=MAGREMENT

PR #247619, opened 2026-02-06
[FIX] account_edi_ubl_cii: do not import info on vendor refund

agr-odoo:
opw-5870516
opw-5898302
opw-5917394
opw-5892671


alialfie (changes requested):
Hello, thanks for the fix

I'm still checking the fix and the source of this, but in the meantime can you please write a test case?

alialfie (approved):
Since all checks passed and this is very urgent, skipping the test for now
robodoo r+

PR #247620, opened 2026-02-06
[IMP] project: creation of project templates

aurb9:
Currently, it is not really easy to a lambda user to discover the project template and also discover all the features behind that project template (project role, assign tasks based on project role, etc) because there is no way to create project templates from scratch. The user has first create a project, since the project creation in the main view of project app redirects to the kanban view of its tasks, the user will create stages and tasks and afterwards he has to go back to the main view and go into the settings of that project to be able to convert it into a template. However, by doing that, he will miss the power of project roles since that field is only available in the template tasks linked to a project template, which means if the user never goes back to the form view of a task inside that project template, he will surely never discover project roles.

Therfore, in this commit we add a "Project Templates" menu to the configurations. This enables users to manage, add and delete templates from the same place, without the complicated worflow described above. In addition, a project can now easily be converted to a template within the project's tasks views, using the action gear.

task-5249020

### DIFF
--- a/addons/project/views/project_task_views.xml
+++ b/addons/project/views/project_task_views.xml
@@ -28,7 +28,7 @@
                         <filter name="create_date_last_365_days" string="Last 365 Days" domain="[('date_last_stage_update', '&gt;=', 'today -365d'), ('date_last_stage_update', '&lt;', 'today')]"/>
                     </filter>
                     <separator/>
-                    <filter string="Templates" context="{'render_task_templates': True}" name="templates" domain="['|', ('has_template_ancestor', '=', True), ('has_project_template', '=', True)]"/>
+                    <filter string="Templates" context="{'render_task_templates': True, 'default_is_template': True}" name="templates" domain="['|', ('has_template_ancestor', '=', True), ('has_project_template', '=', True)]"/>
                     <group>
                         <filter string="Stage" name="stage" context="{'group_by': 'stage_id'}"/>
                         <filter string="Milestone" name="milestone" context="{'group_by': 'milestone_id'}" groups="project.group_project_milestone" invisible="not context.get('allow_milestones', True)"/>


addons/project/static/src/components/project_convert_project_to_template_cog_menu/project_convert_project_to_template_cog_menu.js
xavierbol, line 1: I know you loved JS 🤩 BUT we could maybe just to a server action, no? 😇 

Something like this: <https://github.com/odoo/enterprise/blob/master/planning/views/planning_views.xml#L1057-L1065>

addons/project/views/project_project_views.xml
```diff
@@ -427,6 +430,15 @@
             <field name="context">{"default_allow_billable": 0}</field>
         </record>
 
+        <record id="open_create_project_template" model="ir.actions.act_window">
+            <field name="name">Create a Project Template</field>
+            <field name="res_model">project.project</field>
+            <field name="view_mode">form</field>
+            <field name="view_id" ref="project_project_view_form_simplified_footer"/>
+            <field name="target">new</field>
+            <field name="context">{"default_allow_billable": 0}</field>
```
xavierbol, line 442: funny to have that context in project since allow_billable is only defined in sale_project :o

addons/project/views/project_project_views.xml
```diff
@@ -799,11 +811,25 @@
...
+            <field name="name">Project Templates</field>
+            <field name="res_model">project.project</field>
+            <field name="path">project-template</field>
+            <field name="domain">[('is_template', '=', True)]</field>
+            <field name="view_mode">list,kanban,form</field>
+            <field name="context">{'default_is_template': True}</field>
+            <field name="view_ids" eval="[(5, 0, 0),
+                (0, 0, {'view_mode': 'kanban', 'view_id': ref('project_templates_view_kanban')}),
+                (0, 0, {'view_mode': 'list', 'view_id': ref('project_templates_view_list')}),
+                (0, 0, {'view_mode': 'form', 'view_id': ref('project_templates_view_form')})]"/>
```
xavierbol: I would prefer creating `ir.actions.act_window.view` records instead of doing that because it is easier to alter it. 😇

xavierbol (commented):

addons/project/static/src/components/project_convert_project_to_template_cog_menu/project_convert_project_to_template_cog_menu.js
aurb9: Actually no, it does not seem to work here as we want it on the list and kanban views of project.task but the action should be called on project.project. It seems like we have to do that in JS

addons/project/views/project_project_views.xml
aurb9: Indeed, but that would require to override / patch the action in sale_project if not done here right?

xavierbol: yes, exactly, but you can let as it is, it is not really a big deal

addons/project/static/src/components/project_convert_project_to_template_cog_menu/project_convert_project_to_template_cog_menu.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+    <t t-name="project.ConvertProjectToTemplateCogMenu">
+        <DropdownItem onSelected.bind="toggleProjectTemplateMode">
```
deneuvillem:
```suggestion
        <DropdownItem onSelected.bind="this.toggleProjectTemplateMode">
```

addons/project/static/src/components/project_convert_project_to_template_cog_menu/project_convert_project_to_template_cog_menu.xml
```diff
@@ -0,0 +1,10 @@
...
+            Convert to Template
```
deneuvillem:
```suggestion
            Convert Project to Template
```

deneuvillem:
Thanks for your work!

robodoo r+

deneuvillem:
robodoo rebase-ff

deneuvillem:
robodoo r-

deneuvillem:
TODO: change the `ConvertProjectToTemplateCogMenu` by using a project.task controller and override `getStaticActionMenuItems()`. Also add a check to only display it for project managers :)

deneuvillem:
robodoo r+

PR #247624, opened 2026-02-06
[18.0][FIX] product: Return product record with a clean context

ypapouin:

Creating a new product and then copying it should works correctly.

If you create a new product
```
    item_a = self.env["product.product"].create({"name": "ItemA"})
```
And then copy it
```
    item_b = item_a.copy({"name": "ItemB"})
    print(item_b.name) -> False
```

You will have the surprise to see that "item_b" is en empty recordset >  `item_b == product.product()`

Why ? Simply because a context key is set by odoo (`create_product_product=False`) when creating a `product.product`

So to avoid this issue, re-browse your newly created record to have a clean context:
```
    item_a = self.env["product.product"].create({"name": "ItemA"})
    item_a = self.env["product.product"].browse(item_a.id)
    item_b = item_a.copy({"name": "ItemB"})
```

_Or fix Odoo by returning a clean context: a good rule should be that any context key set in `create`/`copy`  be unset before returning the recordset._

```
    item_a = self.env["product.product"].create({"name": "ItemA"})
    item_b = item_a.copy({"name": "ItemB"})
    print(item_b.name) -> "ItemB"
```

---

### DIFF
--- a/addons/product/models/product_product.py
+++ b/addons/product/models/product_product.py
@@ -721,7 +721,8 @@ class ProductProduct(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        return super(ProductProduct, self.with_context(create_product_product=False)).create(vals_list)
+        products = super(ProductProduct, self.with_context(create_product_product=False)).create(vals_list)
+        return products.with_env(self.env)
 
     def action_archive(self):
         records = self.filtered('active')
--- a/addons/product/tests/test_variants.py
+++ b/addons/product/tests/test_variants.py
@@ -248,6 +248,12 @@ class TestVariants(ProductVariantsCommon):
         self.assertEqual(variant_copy.name, 'Test Copy (copy) (copy)')
         self.assertEqual(len(variant_copy.product_variant_ids), 2)
 
+        # test copy of variant from variant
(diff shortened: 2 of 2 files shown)


addons/product/models/product_product.py
```diff
@@ -376,6 +376,10 @@ def create(self, vals_list):
         products = super(ProductProduct, self.with_context(create_product_product=False)).create(vals_list)
         # `_get_variant_id_for_combination` depends on existing variants
         self.env.registry.clear_cache()
+        if 'create_product_product' in products._context:
+            new_context = dict(products._context)
+            new_context.pop('create_product_product')
+            return products.with_context(new_context)
         return products
```
Feyensv:
```suggestion
        return products.with_env(self.env)
```
might be simpler (or `product.with_context(self.env.context)`) if we indeed stick to the principle that created returns must have the same context (& user)

Feyensv (commented):
Hello @ypapouin,

Thanks for the fix but I don't think it's worth merging in a stable version (it's been there for years and only impacts custom code).
Could you please target `master` instead ?

addons/product/models/product_product.py
ypapouin: Thank you for the tip, I applied it while re-targeting master

PR #247625, opened 2026-02-06
19.0 perf valo mmha arm

amoyaux:

Current behavior before PR:

---

### DIFF
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -3,10 +3,7 @@
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError
 from odoo.fields import Domain
-from odoo.tools import float_is_zero, float_repr, float_round, float_compare
 from odoo.exceptions import ValidationError
-from collections import defaultdict
-from datetime import datetime
 
 
 class ProductTemplate(models.Model):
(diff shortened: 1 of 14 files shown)


addons/stock_account/models/product.py
```diff
@@ -143,6 +140,28 @@ def _compute_value(self):
...
         self.company_currency_id = company_id.currency_id
 
+        # PERF: Pre-compute:the sum of 'total_value' of lots per product in go
+        total_value_by_product_id = {}
+        lot_valuated_products = self.filtered('lot_valuated')
+        if lot_valuated_products:
+            lots_by_product = self.env['stock.lot']._read_group(
+                domain=[
+                    ('product_id', 'in', lot_valuated_products.ids),
+                    ('product_qty', '!=', 0),
```
Whenrow: This field is not warehouse dependant which can be problematic if the value is computed on a single warehouse

addons/stock_account/models/product.py
```diff
@@ -158,34 +166,52 @@ def _compute_value(self):
                     for _, lots in lots_by_product
                     for lot in lots
             ).mapped('total_value')
-
             for product, lots in lots_by_product:
-                total_value_by_product_id[product.id] = sum(lots.mapped('total_value'))
+                value = sum(lots.mapped('total_value'))
+                std_price_by_product_id[product.id] = value / product.qty_available if product.qty_available else product.standard_price
```
Whenrow, line 171: Not sure the `at_date` param is respected here

addons/stock_account/models/product.py
```diff
@@ -280,17 +304,31 @@ def _get_remaining_moves(self):
...
+        if at_date:
+            product_value_by_product = self._get_last_product_value(at_date, lot=lot)
+            std_price_by_product_id = {
+                product.id: product_value_by_product[product].value if product in product_value_by_product else product.standard_price
+                for product in self
+            }
+        value_by_product_id = {p.id: p.qty_available * std_price_by_product_id.get(p.id, 0) for p in self}
+        return std_price_by_product_id, value_by_product_id
+
+    def _run_average_batch(self, at_date=None, lot=None, force_recompute=False):
```
Whenrow, line 318: I don't see a call without `force_recompute=True`

addons/stock_account/models/product.py
amoyaux: It call total_value after that uses correctly the warehouse in the context. Otherwise it will select the lot but returns zero as value

addons/stock_account/models/product.py
amoyaux: Yep but it was zero before. Could be improve in the future

addons/stock_account/models/product.py
amoyaux: In _update_standard_price

PR #247641, opened 2026-02-06
Fold options (with extra button to select)

blse-odoo:
Enterprise PR: https://github.com/odoo/enterprise/pull/107029

### [IMP] html_builder, website: show overlay buttons of carousel from slide

When the user clicks in a "Carousel" snippet, the target of the click
is the slide in the carousel, not the carousel itself, thus the overlay
buttons of the carousel do not appear.

To make it easier to move the carousel around, this commit shows the
overlay buttons of the carousel when he user clicks inside the slide
(because there is no overlay buttons for the slides).

task-5470590

### [FIX] html_builder: add a fallback rule to name blocks "Form"

Some forms snippet have the attribute `data-name=Form` on their form,
but some do not have it. For those that don't, the name of the option
group is the default fallback "Block".

This commit adds a fallback rule to name the snippets `s_website_form`
as "Form".

task-5470590

### [IMP] html_builder, website, *: fold groups of options but the last

*: test_website, website_mass_mailing, website_payment

To have a quicker access to the options related to the element the user
clicked, the groups above the last one are folded.

The user can fold/unfold a group by clicking on its header (with the
name of the element).

Clicking on the header used to select just the corresponding element,
this action is sill possible on `section` elements (as they are the ones
for which this action might be useful)

task-5470590

### [FIX] website_sale: always show header of product page options

Commit e8019e39f1e8dd7cdea2d6cb3070de4c073ab94d, added a css rule to
hide the header of the product page group (it reproduced what was done
in 328c241fa58976bd5029dc526789113a88a333f2). With the changes from the
previous commit, this leads to the very weird behavior when clicking on
a block inside the page: the header for the options of the page appears
as folded, and if unfolded the header disappears.

This commit remove the rule that make the header disappear.

task-5470590

### [IMP] html_builder, website: show temporary overlay on mouse move

To help the user click on a specific block, this commit adds an overlay
on the first element with options on hover.

task-5470590

### DIFF
--- a/addons/html_builder/static/src/core/builder_options_plugin.js
+++ b/addons/html_builder/static/src/core/builder_options_plugin.js
@@ -6,6 +6,7 @@ import { isClonable } from "./clone_plugin";
 import { getElementsWithOption, isElementInViewport } from "@html_builder/utils/utils";
 import { OptionsContainer } from "@html_builder/sidebar/option_container";
 import { shouldEditableMediaBeEditable } from "@html_builder/utils/utils_css";
+import { _t } from "@web/core/l10n/translation";
 
 /** @typedef {import("@html_builder/core/utils").BaseOptionComponent} BaseOptionComponent */
 /** @typedef {import("@odoo/owl").Component} Component */
@@ -140,6 +141,17 @@ export class BuilderOptionsPlugin extends Plugin {
                 const el = this.editable.querySelector(this.config.initialTarget);
                 this.updateContainers(el);
             }
+        },
+        get_options_container_top_buttons: (el) => {
+            const buttons = [];
+            if (el.matches("section")) {
+                buttons.push({
+                    class: "fa fa-fw fa-crosshairs btn o-hb-btn btn-accent-color-hover",
+                    title: _t("Select only this block"),
+                    handler: (el) => this.updateContainers(el),
+                });
+            }
+            return buttons;
         },
     };
 
@@ -343,12 +355,18 @@ export class BuilderOptionsPlugin extends Plugin {
             element = element.parentElement;
         }
 
-        const previousElementToIdMap = new Map(this.lastContainers.map((c) => [c.element, c.id]));
+        const previousElementToIdAndStateMap = new Map(
+            this.lastContainers.map((c) => [c.element, { id: c.id, folded: c.folded }])
+        );
+        const keepUnfolded = this.lastContainers.some((c) => c.element === element);
         let containers = reactive(
             [...elementToOptions]
                 .sort(([a], [b]) => (b.contains(a) ? 1 : -1))
                 .map(([element, Options]) => ({
-                    id: previousElementToIdMap.get(element) || uniqueId(),
+                    id: previousElementToIdAndStateMap.get(element)?.id || uniqueId(),
+                    folded: keepUnfolded
+                        ? previousElementToIdAndStateMap.get(element)?.folded ?? true
+                        : true,
                     element,
                     options: Options,
                     optionTitleComponents: elementToOptionTitleComponents.get(element) || [],
@@ -370,6 +388,10 @@ export class BuilderOptionsPlugin extends Plugin {
         );
         if (lastValidContainerIdx > 0) {
             containers = containers.slice(lastValidContainerIdx);
+        }
+        const lastContainerWthOptions = containers.findLast((c) => c.options.length);
+        if (lastContainerWthOptions) {
+            lastContainerWthOptions.folded = false;
         }
         return containers;
     }
--- a/addons/html_builder/static/src/core/builder_options_plugin.js
+++ b/addons/html_builder/static/src/core/builder_options_plugin.js
@@ -7,6 +7,7 @@ import { getElementsWithOption, isElementInViewport } from "@html_builder/utils/
 import { OptionsContainer } from "@html_builder/sidebar/option_container";
 import { shouldEditableMediaBeEditable } from "@html_builder/utils/utils_css";
 import { _t } from "@web/core/l10n/translation";
+import { closestElement } from "@html_editor/utils/dom_traversal";
 
 /** @typedef {import("@html_builder/core/utils").BaseOptionComponent} BaseOptionComponent */
 /** @typedef {import("@odoo/owl").Component} Component */
@@ -44,6 +45,7 @@ import { _t } from "@web/core/l10n/translation";
 /**
  * @typedef { Object } BuilderOptionsShared
  * @property { BuilderOptionsPlugin['checkElement'] } checkElement
+ * @property { BuilderOptionsPlugin['closestWithOption'] } closestWithOption
  * @property { BuilderOptionsPlugin['computeContainers'] } computeContainers
  * @property { BuilderOptionsPlugin['findOption'] } findOption
  * @property { BuilderOptionsPlugin['getContainers'] } getContainers
@@ -116,6 +118,7 @@ export class BuilderOptionsPlugin extends Plugin {
     static dependencies = ["operation", "history"];
     static shared = [
         "checkElement",
+        "closestWithOption",
         "computeContainers",
         "findOption",
         "getContainers",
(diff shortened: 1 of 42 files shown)


addons/html_builder/static/src/core/builder_options_plugin.js
```diff
@@ -141,6 +142,17 @@ export class BuilderOptionsPlugin extends Plugin {
...
+            const buttons = [];
+            if (el.matches("section")) {
+                buttons.push({
+                    class: "fa fa-fw fa-location-arrow btn o-hb-btn btn-accent-color-hover",
+                    title: _t("Select only this block"),
+                    handler: (el) => this.updateContainers(el),
+                });
+            }
+            return buttons;
+        },
```
AlessandroLupo, line 158: Two questions:
1. Should we hide the "Select only this block" button if it appears on the last option? Clicking on it would have no effect, right?
2. Maybe we should add a test for this?

addons/website_sale/static/src/website_builder/website_sale.editor.scss
```diff
@@ -1,9 +1,5 @@
 .options-container:has(.o_wsale_products_list_page_options),
 .options-container:has(.o_wsale_product_page_options) {
-    > div:first-child { // Title
-        display: none !important;
-    }
-
```
AlessandroLupo, line 6: Idk if you already had a chat with a PO, but another possibility here may be to allow for non-foldable options. In that case, the "product page" option could be always unfolded with its tilte hidden

AlessandroLupo (commented):
Great work! The builder feels good with the new overlays and with the folded options. I left just a couple of optional comments/questions. 👍

addons/html_builder/static/src/core/builder_options_plugin.js
```diff
@@ -371,6 +391,12 @@ export class BuilderOptionsPlugin extends Plugin {
         if (lastValidContainerIdx > 0) {
             containers = containers.slice(lastValidContainerIdx);
         }
+        for (let i = containers.length - 1; i >= 0; i--) {
+            containers[i].foldedState.folded = false;
+            if (containers[i].options.length) {
+                break;
+            }
+        }
```
FrancoisGe: Why do you need this code ?

addons/html_builder/static/src/core/builder_options_plugin.js
```diff
@@ -343,12 +355,20 @@ export class BuilderOptionsPlugin extends Plugin {
...
             [...elementToOptions]
                 .sort(([a], [b]) => (b.contains(a) ? 1 : -1))
                 .map(([element, Options]) => ({
-                    id: previousElementToIdMap.get(element) || uniqueId(),
+                    id: previousElementToIdAndStateMap.get(element)?.id || uniqueId(),
+                    foldedState: {
+                        folded: keepUnfolded
+                            ? previousElementToIdAndStateMap.get(element)?.folded ?? true
+                            : true,
+                    },
```
FrancoisGe: It s to keep the last container open ?

addons/html_builder/static/src/sidebar/option_container.js
```diff
@@ -92,8 +93,10 @@ export class OptionsContainer extends BaseOptionComponent {
         return (title || getSnippetName(this.env.getEditingElement())) + titleExtraInfo;
     }
 
-    selectElement() {
-        this.dependencies.builderOptions.updateContainers(this.props.editingElement);
+    toggleFold() {
+        if (this.props.foldedState) {
+            this.props.foldedState.folded = !this.props.foldedState.folded;
+        }
```
FrancoisGe: It s not a good practice in owl 2 to edit something from the props. But in owl 3, it will be ok so we can probably keep it like this

addons/html_builder/static/src/sidebar/option_container.xml
```diff
@@ -7,19 +7,20 @@
          t-on-pointerenter="onPointerEnter" t-on-pointerleave="onPointerLeave">
         <div class="options-container-header d-flex align-items-center justify-content-between">
             <t t-set="isActionable" t-value="Object.keys(props.snippetModel).length > 0"></t>
-            <span
-                class="options-container-label ps-2 py-2"
+            <div
+                class="options-container-label py-2 flex-grow-1 d-flex align-items-center"
```
FrancoisGe, line 11: Why do you need this changes ?

addons/html_builder/static/src/sidebar/option_container.xml
```diff
@@ -59,7 +60,7 @@
                 </t>
             </div>
         </div>
-        <div class="we-bg-options-container pb-3" t-ref="content">
+        <div t-if="!this.props.foldedState?.folded" class="we-bg-options-container pb-3" t-ref="content">
```
FrancoisGe: Did you find any scenarios where the container is empty when you unfold it?

addons/html_builder/static/tests/helpers.js
```diff
@@ -551,3 +551,9 @@ export async function editBuilderRangeValue(selector, newValue) {
     input.dispatchEvent(new Event("change"));
     await delay();
 }
+
+export async function unfoldAllOptionsGroups() {
+    for (const i of queryAll(".options-container-header i.fa-caret-right")) {
+        await contains(i).click();
+    }
```
FrancoisGe, line 558: It s not possible to click and just wait one annimationFrame at the end ?

addons/website/static/tests/builder/custom_tab/container_buttons.test.js
```diff
@@ -385,3 +388,14 @@ test("applying option container button should wait for actions in progress", asy
     undo(editor);
     expect(editable).toHaveInnerHTML(`<p class="test-options-target">plop</p>`);
 });
+
+test("Use the sidebar 'target' button", async () => {
+    await setupWebsiteBuilderWithSnippet("s_banner");
+    await contains(":iframe h1").click();
+    await animationFrame();
```
FrancoisGe: Why do you need to await 2 `animationFrame` (1 from `contains.click` and 1 from `animationFrame`)

addons/website/static/tests/builder/website_builder/image_gallery.test.js
```diff
@@ -135,6 +138,8 @@ test("Change gallery restore the container to the cloned equivalent image", asyn
     };
 
     await contains(":iframe img[data-index='1']").click();
+    await builder.waitSidebarUpdated();
+    await unfoldAllOptionsGroups();
```
FrancoisGe, line 142: Maybe we want to await `waitSidebarUpdated` always when you call `unfoldAllOptionsGroups`

addons/html_builder/static/src/core/builder_overlay/builder_overlay_plugin.js
```diff
@@ -156,17 +175,42 @@ export class BuilderOverlayPlugin extends Plugin {
...
+            hasOverlayOptions: false,
+            next: this.dependencies.operation.next,
+            isMobileView: this.config.isMobileView,
+            mobileBreakpoint: this.config.mobileBreakpoint,
+            isRtl: this.config.isEditableRTL,
+            isHoverOverlay: true,
+        });
+        this.hoverOverlay = overlay;
+        this.overlayContainer.append(overlay.overlayElement);
+        this.resizeObserver.observe(overlay.overlayTarget, { box: "border-box" });
```
FrancoisGe, line 211: Finally, did we choose to display the overlay for all elements that have a container option or just those that have the blue square ?

addons/html_builder/static/src/core/builder_options_plugin.js
blse-odoo: Some containers only have "overlay options", but zero options in the sidebar (thus do not appear in the sidebar), we need to skip them to unfold the last one in the sidebar

addons/html_builder/static/src/core/builder_options_plugin.js
blse-odoo: It is to keep the containers that were open just before, in case the `computeContainers` is triggered by a step (due to an option change) or the user select a element which was already in the sidebar (like with the "Select only this block" button)

addons/html_builder/static/src/sidebar/option_container.xml
blse-odoo: To add the icon indicating the folded/unfolded state and grab the click on the icon + the name + the blank space after the name

addons/html_builder/static/src/sidebar/option_container.xml
blse-odoo: Not yet

addons/html_builder/static/src/core/builder_overlay/builder_overlay_plugin.js
blse-odoo: We choose to show it on elements that have options

FrancoisGe:
@robodoo r+

PR #247693, opened 2026-02-07
[FIX] account_edi_ubl_cii: log the account numbers

gawa-odoo:
---


odony:
@robodoo r+ rebase-ff fw=skipmerge

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -351,41 +351,9 @@ def _import_retrieve_and_fill_partner(self, invoice, name, phone, mail, vat, cou
                 invoice.partner_id.vat = vat
 
     def _import_retrieve_and_fill_partner_bank_details(self, invoice, bank_details):
-        """ Retrieve the bank account, if no matching bank account is found, create it
```
eapa-odoo: Hello @gawa-odoo 
Just want to understand - why aren’t we creating a bank account? what's the reason behind that?
I can't find any task related to that, so..

Thanks

PR #247708, opened 2026-02-07
[IMP] mail, im_livechat:  replace MessageConfirmDialog with dedicated dialogs

mear-odoo:
This PR removes the generic MessageConfirmDialog component and replaces it
with dedicated dialogs such as MessagePinUnpinDialog, LivechatLeaveDialog, and
MessageDeleteDialog.

Using dedicated dialogs simplifies the code, avoids unnecessary customization
of a generic component, and makes each dialog easier to reason about by keeping
its content static and scoped to its specific use case.

<table> <tr> <th>Before</th> <th>After</th> </tr> <tr> <td>  </td> <td> image </td> </tr> <tr> <td>  </td> <td> image </td> </tr> </table>

enterprise: https://github.com/odoo/enterprise/pull/108806
task-5873845

---


jayp-odoo (approved):
LGTM

alexkuhn (commented):
- I would put 2nd sentence in its own line and muted, so that the dialog looks easy to understand and this extra text looks optional
- I would add icons on the buttons, for me icons makes it much easier to pick the right action without thinking.

With these changes, I would turn the `MessageConfirmDialog` props into slots, to enable richer content on the text and buttons.

I would expect a look like this:

Image

addons/mail/static/src/core/common/message_delete_dialog.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+<t t-name="mail.MessageDeleteDialog">
+    <Dialog size="'xl'" title.translate="Send this message to the great trash can in the sky?" contentClass="'o-bg-body'">
+        <p class="mx-3">Are you sure you want to bid farewell to this message forever?</p>
+        <blockquote class="mx-3 fst-normal" style="pointer-events:none;">
+            <t t-component="messageComponent" message="props.message" isReadOnly="true" hasActions="false"/>
```
alexkuhn:
```suggestion
        <blockquote class="mx-3 fst-normal pe-none">
            <t t-component="messageComponent" message="props.message" isReadOnly="true" hasActions="false"/>
```

Can change to all of them.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -641,33 +641,18 @@ export class DiscussChannel extends Record {
...
             message,
-            prompt: _t("You sure want this message pinned to %(conversation)s forever and ever?", {
-                conversation: this.prefix + this.displayName,
-            }),
-            size: "md",
-            title: _t("Pin It"),
-            onConfirm: () => {
-                this.setMessagePin(message, true);
-            },
+            conversation: this.prefix + this.displayName,
```
alexkuhn: The conversation can be deduced from `message`, so the component only needs `message` as prop. `prefix` + `message.channel_id.displayName` can be retrieved by the component itself.

addons/mail/static/src/core/common/message_pin_unpin_dialog.js
```diff
@@ -0,0 +1,50 @@
...
+export class MessagePinUnpinDialog extends Component {
+    static components = { Dialog };
+    static props = {
+        close: Function,
+        message: Object,
+        onConfirm: Function,
+        conversation: { type: String, optional: true },
+        isUnpin: { type: Boolean, optional: true },
+    };
+    static template = "mail.MessagePinUnpinDialog";
```
alexkuhn: I would just name it `MessagePinDialog`. I think it's reasonable to assume the unpin is made with same dialog.

`MessagePinDialog` keeps the name more readable ;)

addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
```diff
@@ -90,16 +90,10 @@ const discussChannelPatch = {
     },
     async leaveChannel() {
         if (this.livechatShouldAskLeaveConfirmation) {
-            this.store.env.services.dialog.add(MessageConfirmDialog, {
+            this.store.env.services.dialog.add(LivechatLeaveDialog, {
+                channelName: this.displayName,
                 message: this.newestPersistentOfAllMessage,
```
alexkuhn: You could pass `channel: this` and have `displayName` and `channel.newestPersistentOfAllMessage` deduced by the `LivechatLeaveDialog` internal code / template.

This would reduce the amount of props, and having a richer prop `channel` is also easier to understand source of data of the component.

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -641,33 +641,18 @@ export class DiscussChannel extends Record {
...
-            prompt: _t(
-                "Well, nothing lasts forever, but are you sure you want to unpin this message?"
-            ),
-            size: "md",
-            title: _t("Unpin Message"),
-            onConfirm: () => {
-                this.setMessagePin(message, false);
-            },
+            isUnpin: true,
+            onConfirm: () => this.setMessagePin(message, false),
```
alexkuhn: `onConfirm` can be removed from the props: the internal handler of the button of the `MessagePinDialog` can deduce the method to call on `props.message` solely base on `message` and `props.isUnpin`.

addons/im_livechat/static/src/core/public_web/discuss_channel_model_patch.js
```diff
@@ -90,16 +90,10 @@ const discussChannelPatch = {
...
-                confirmText: _t("Leave Conversation"),
                 onConfirm: async () => await super.leaveChannel(...arguments),
```
alexkuhn, line 95: Would be best not have to pass `onConfirm` and have it deduced from `props.channel` of the component, but since this is implemented here with a `super.leaveChannel`, we need to keep it otherwise we'd have to clean the code further. We are already doing enough cleaning in this PR, so ok to keep `onConfirm` for `LivechatLeaveDialog` for now!

addons/mail/static/src/core/common/message_model.js
```diff
@@ -737,14 +737,10 @@ export class Message extends Record {
      */
     showDeleteConfirm(owner) {
         this.store.env.services.dialog.add(
-            discussComponentRegistry.get("MessageConfirmDialog"),
+            discussComponentRegistry.get("MessageDeleteDialog"),
             {
                 message: this,
-                confirmText: _t("Delete"),
                 onConfirm: () => this.onShowDeleteConfirm(owner),
```
alexkuhn: Here also the `onShowDeleteConfirm` can be deduced by the `MessageDeleteDialog` internal code, thanks to `props.message`.

The less props the better. This keeps complexity of component small.

addons/im_livechat/static/src/core/public_web/livechat_leave_dialog.xml
```diff
@@ -0,0 +1,23 @@
...
+        <blockquote t-if="props.message" class="mx-3 fst-normal" style="pointer-events:none;">
+            <t t-component="messageComponent" message="props.message" isReadOnly="true" hasActions="false"/>
+        </blockquote>
+        <t t-set-slot="footer">
+            <button class="btn btn-danger me-2" t-on-click="onClickConfirm">
+                <i class="fa fa-sign-out me-1"/>
+                Leave Conversation
+            </button>
+            <button class="btn btn-secondary me-2" t-on-click="props.close">
+                <i class="fa fa-times me-1"/>
```
alexkuhn: Prefer `oi oi-close` over `fa fa-times`: thinner and better ^^
For all the "Discard" in this diff.

addons/im_livechat/static/src/core/public_web/livechat_leave_dialog.xml
```diff
@@ -0,0 +1,23 @@
...
+                Discard
+            </button>
```
alexkuhn, line 18: All the icons in button, I would put `opacity-50`. This would reduce their visibility to not give a bad impression the UI is bloated, while still have these icons for the many people that think a UI is more intuitive with presence of this icons (like myself).

See the difference before / after:
Image
Image

alexkuhn (commented):
Lots of comments but all of them are quite minor.

Looks pretty good. The extra diff is justified with wanting for richer dialogs and not making the component too generic in a way that feels wrong. So all's good for me.

addons/mail/static/src/core/common/message_delete_dialog.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+<t t-name="mail.MessageDeleteDialog">
+    <Dialog size="'xl'" title.translate="Send this message to the great trash can in the sky?" contentClass="'o-bg-body'">
```
alexkuhn:
```suggestion
    <Dialog size="'md'" title.translate="Send this message to the great trash can in the sky?" contentClass="'o-bg-body'">
```

`"xl"` was there before, but this is too much. And also I didn't realize, but sm and md don't have visible separators on header and footer, whereas the other do. The `MessageDeleteDialog` feels bloated because of these visible separators.

Changing to `'md'` fixes that, and also makes it consistent with the Pin Dialog :)

addons/im_livechat/static/src/core/public_web/livechat_leave_dialog.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+<t t-name="im_livechat.LivechatLeaveDialog">
+    <Dialog size="'xl'" title="title" contentClass="'o-bg-body'">
```
alexkuhn:
```suggestion
    <Dialog size="'md'" title="title" contentClass="'o-bg-body'">
```

Same, would also reduce to `md` for more pleasant dialog and consistent size :)

alexkuhn (approved):
LGTM, thank you!
@robodoo delegate+

PR #247715, opened 2026-02-09
[IMP] mail,hr: improve auto-subscribe mechanism of discuss channels

shmn-odoo:
**Current behavior before PR,**
We could only auto-subscribe users based on their groups or a department of
their related employee. this was often too broad, making the feature less
effective.

**Desired behavior after PR is merged,**
It is now possible to subscribe users by company, which also works in
combination with groups and departments.

we also track users added via auto-subscribe to ensure they are not re-added if
they have already left the channel.

additionally, we added an 'Auto-Subscribe' checkbox under the 'Channel Visibility'
section which is disabled by default, related fields only appear when
'Auto-Subscribe' is enabled.

when active, all authorized users are subscribed by default.

enterprise: https://github.com/odoo/enterprise/pull/117048
upgrade: odoo/upgrade#9436
upgrade-util: odoo/upgrade-util#416

task-4804991

---

### DIFF
--- a/addons/hr/models/discuss_channel.py
+++ b/addons/hr/models/discuss_channel.py
@@ -3,6 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.fields import Domain
 
 
 class DiscussChannel(models.Model):
@@ -10,7 +11,7 @@ class DiscussChannel(models.Model):
 
     subscription_department_ids = fields.Many2many(
         'hr.department', string='HR Departments',
-        help='Automatically subscribe members of those departments to the channel.')
+        help="Restrict auto-subscribe users based on their related employees' departments.")
 
     @api.constrains('subscription_department_ids')
     def _constraint_subscription_department_ids_channel(self):
@@ -18,15 +19,13 @@ class DiscussChannel(models.Model):
         if failing_channels:
             raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the department auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
 
-    def _subscribe_users_automatically_get_members(self):
-        """ Auto-subscribe members of a department to a channel """
-        new_members = super()._subscribe_users_automatically_get_members()
-        for channel in self:
-            new_members[channel.id] = list(
-                set(new_members[channel.id]) |
-                set((channel.subscription_department_ids.member_ids.user_id.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids)
-            )
-        return new_members
+    def _get_auto_subscribe_domain(self):
+        domain = super()._get_auto_subscribe_domain()
+        # sudo: hr.department - ensure all departments and their members are accessible
+        departments_sudo = self.sudo().subscription_department_ids
+        if departments_sudo:
+            domain &= Domain('id', 'in', departments_sudo.member_ids.user_id.ids)
+        return domain
 
     def write(self, vals):
         res = super().write(vals)
--- a/addons/hr/models/discuss_channel.py
+++ b/addons/hr/models/discuss_channel.py
@@ -3,6 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.fields import Domain
 
 
 class DiscussChannel(models.Model):
@@ -18,15 +19,19 @@ class DiscussChannel(models.Model):
         if failing_channels:
             raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the department auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
 
-    def _subscribe_users_automatically_get_members(self):
-        """ Auto-subscribe members of a department to a channel """
-        new_members = super()._subscribe_users_automatically_get_members()
-        for channel in self:
-            new_members[channel.id] = list(
-                set(new_members[channel.id]) |
-                set((channel.subscription_department_ids.member_ids.user_id.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids)
+    def _get_extra_domain(self):
+        domain = super()._get_extra_domain()
+        # sudo: hr.department - ensure all departments and its members are accesible
+        departments_sudo = self.sudo().subscription_department_ids
+        if departments_sudo:
+            department_domain = Domain(
+                'id', 'in', departments_sudo.member_ids.user_id.ids,
             )
-        return new_members
+            if self.group_ids:
+                domain |= department_domain
+            else:
+                domain &= department_domain
+        return domain
 
     def write(self, vals):
         res = super().write(vals)
--- a/addons/hr/tests/test_channel.py
+++ b/addons/hr/tests/test_channel.py
@@ -27,6 +27,7 @@ class TestChannel(TestHrCommon):
         self.assertEqual(self.department.member_ids, self.emp0)
 
         self.channel.write({
+            'auto_subscribe': True,
             'subscription_department_ids': [(4, self.department.id)]
         })
 
@@ -34,6 +35,7 @@ class TestChannel(TestHrCommon):
 
     def test_auto_subscribe_when_updating_employee_department(self):
         self.channel.write({
+            'auto_subscribe': True,
             'subscription_department_ids': [(4, self.department.id)],
         })
         self.assertEqual(self.channel.channel_partner_ids, self.env['res.partner'])
@@ -41,4 +43,14 @@ class TestChannel(TestHrCommon):
 
         self.emp0.write({'department_id': self.department.id})
 
+        self.assertEqual(self.channel.channel_partner_ids, self.emp0.user_id.partner_id)
+
+    def test_auto_subscribe_group_department(self):
+        self.department.invalidate_recordset(['member_ids'])
+        self.emp0.write({'department_id': self.department.id})
+        self.channel.write({
+            'auto_subscribe': True,
+            'group_ids': [(4, self.env.ref("base.group_user").id)],
+            'subscription_department_ids': [(4, self.department.id)],
+        })
         self.assertEqual(self.channel.channel_partner_ids, self.emp0.user_id.partner_id)
--- a/addons/hr/tests/test_channel.py
+++ b/addons/hr/tests/test_channel.py
@@ -13,19 +13,31 @@ class TestChannel(TestHrCommon):
 
         cls.channel = cls.env['discuss.channel'].create({'name': 'Test'})
 
-        emp0 = cls.env['hr.employee'].create({
+        cls.emp0 = cls.env['hr.employee'].create({
             'user_id': cls.res_users_hr_officer.id,
         })
         cls.department = cls.env['hr.department'].create({
             'name': 'Test Department',
-            'member_ids': [(4, emp0.id)],
+            'member_ids': [(4, cls.emp0.id)],
         })
 
-    def test_auto_subscribe_department(self):
+    def test_auto_join_department(self):
         self.assertEqual(self.channel.channel_partner_ids, self.env['res.partner'])
 
         self.channel.write({
+            'auto_join': True,
             'subscription_department_ids': [(4, self.department.id)]
         })
 
         self.assertEqual(self.channel.channel_partner_ids, self.department.mapped('member_ids.user_id.partner_id'))
+
+    def test_auto_join_group_department(self):
+        self.department.invalidate_recordset(['member_ids'])
+        channel = self.env['discuss.channel'].create({
+            'name': 'Test group and department',
+            'auto_join': True,
+            'group_ids': [(4, self.env.ref("base.group_system").id)],
+            'subscription_department_ids': [(4, self.department.id)],
+        })
+        both_partners = self.env.ref('base.user_admin').partner_id | self.emp0.user_id.partner_id
+        self.assertEqual(channel.channel_partner_ids, both_partners)
--- a/addons/hr/views/discuss_channel_views.xml
+++ b/addons/hr/views/discuss_channel_views.xml
@@ -5,10 +5,9 @@
         <field name="model">discuss.channel</field>
         <field name="inherit_id" ref="mail.discuss_channel_view_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='group_ids']" position="after">
+            <xpath expr="//field[@name='auto_subscribe_company_ids']" position="after">
                 <field name="subscription_department_ids" widget="many2many_tags"
-                    invisible="channel_type != 'channel'"
-                    string="Auto Subscribe Departments"/>
+                    placeholder="All departments" string="Departments" invisible="not auto_subscribe"/>
             </xpath>
         </field>
     </record>
--- a/addons/hr/views/discuss_channel_views.xml
+++ b/addons/hr/views/discuss_channel_views.xml
@@ -5,10 +5,9 @@
         <field name="model">discuss.channel</field>
         <field name="inherit_id" ref="mail.discuss_channel_view_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='group_ids']" position="after">
+            <xpath expr="//field[@name='company_ids']" position="after">
                 <field name="subscription_department_ids" widget="many2many_tags"
-                    invisible="channel_type != 'channel'"
-                    string="Auto Subscribe Departments"/>
+                    placeholder="All departments" string="Departments" invisible="not auto_join"/>
             </xpath>
         </field>
     </record>
--- a/addons/mail/models/discuss/discuss_channel.py
+++ b/addons/mail/models/discuss/discuss_channel.py
@@ -129,11 +129,19 @@ class DiscussChannel(models.Model):
         index=True,
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_subscribe = fields.Boolean(
+        string="Auto Subscribe", default=False,
+        help="Auto-subscribe authorized users to this channel. "
+             "Users can leave anytime and won't be re-added automatically.")
     group_ids = fields.Many2many(
         'res.groups', string='Auto Subscription',
-        help="Members of those groups will automatically added as followers. "
-             "Note that they will be able to manage their subscription manually "
-             "if necessary.")
+        help="Restrict auto-subscribe users to the selected groups.")
+    auto_subscribe_company_ids = fields.Many2many(
+        "res.company", string="Companies",
+        help="Restrict auto-subscribe users to the selected companies.")
+    auto_subscribed_partner_ids = fields.Many2many(
+        "res.partner", "auto_subscribed_partners", export_string_translation=False,
+    )
     # access
     uuid = fields.Char('UUID', size=50, default=_generate_random_token, copy=False)
     group_public_id = fields.Many2one('res.groups', string='Authorized Group', compute='_compute_group_public_id', recursive=True, readonly=False, store=True)
@@ -203,12 +211,20 @@ class DiscussChannel(models.Model):
             if len(ch.channel_member_ids) > 2:
                 raise ValidationError(_("A channel of type 'chat' cannot have more than two users."))
 
-    @api.constrains('group_public_id', 'group_ids')
+    @api.constrains("auto_subscribe", "group_public_id", "group_ids")
     def _constraint_group_id_channel(self):
         # sudo: discuss.channel - skipping ACL for constraint, more performant and no sensitive information is leaked
-        failing_channels = self.sudo().filtered(lambda channel: channel.channel_type != 'channel' and (channel.group_public_id or channel.group_ids))
+        failing_channels = self.sudo().filtered(
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_subscribe or channel.group_ids
+            ),
+        )
         if failing_channels:
-            raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the group-based authorization or group auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
+            raise ValidationError(_(
+                "For %(channels)s, channel_type should be 'channel' to have the group-based "
+                "authorization or auto-subscription",
+                channels=[ch.name for ch in failing_channels],
+            ))
 
     # COMPUTE / INVERSE
 
@@ -594,7 +610,9 @@ class DiscussChannel(models.Model):
                         )
                     )
         result = super().write(vals)
-        if vals.get('group_ids'):
+        if {
+            "auto_subscribe", "auto_subscribe_company_ids", "group_ids", "group_public_id",
+        } & set(vals):
             self._subscribe_users_automatically()
         return result
 
@@ -734,7 +752,10 @@ class DiscussChannel(models.Model):
     # ------------------------------------------------------------
 
     def _subscribe_users_automatically(self):
-        if not (new_members_to_create := self._subscribe_users_automatically_get_members()):
+        channels = self.filtered(lambda c: c.auto_subscribe)
+        if not channels:
+            return
+        if not (new_members_to_create := channels._subscribe_users_automatically_get_members()):
             return
         to_create = [
             {"channel_id": channel_id, "partner_id": partner_id}
@@ -743,7 +764,9 @@ class DiscussChannel(models.Model):
         ]
         # sudo: discuss.channel.member - adding member of other users based on channel auto-subscribe
         new_members = self.env["discuss.channel.member"].sudo().create(to_create)
+        auto_subscribed_members = []
         for member, store in new_members._get_member_store_list():
+            auto_subscribed_members.append((member.channel_id.id, member.partner_id.id))
             store.add(member.channel_id, "_store_channel_fields").add(
                 member,
                 lambda res: (
@@ -751,14 +774,23 @@ class DiscussChannel(models.Model):
                     res.attr("unpin_dt"),
                 ),
             )
+        self.env.cr.execute_values("""
+            INSERT INTO auto_subscribed_partners (discuss_channel_id, res_partner_id)
+                 VALUES %s
+            ON CONFLICT DO NOTHING
+            """, auto_subscribed_members,
+        )
 
     def _subscribe_users_automatically_get_members(self):
         """ Return new members per channel ID """
-        return dict(
-            (channel.id,
-             ((channel.group_ids.all_user_ids.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids))
-                for channel in self
-            )
+        partners_by_channel = {}
+        users = self.env["res.users"].search_fetch([])
+        for channel in self:
+            filtered_users = users.filtered_domain(channel._get_auto_subscribe_domain())
+            if not filtered_users:
+                continue
+            partners_by_channel[channel.id] = filtered_users.partner_id.filtered("active").ids
+        return partners_by_channel
 
     def action_unfollow(self):
         if self.channel_type in self._types_allowing_unfollow():
@@ -904,6 +936,22 @@ class DiscussChannel(models.Model):
                     # sudo: discuss.channel.rtc.session - current user can invite new members in call
                     channel.self_member_id.sudo()._rtc_invite_members(member_ids=new_members.ids)
         return all_new_members
+
+    def _get_auto_subscribe_domain(self):
+        joined_partners = self.channel_member_ids.partner_id | self.auto_subscribed_partner_ids
+        user_domain = Domain([
+            ("active", "=", True),
+            ("partner_id", "not in", joined_partners.ids),
+        ])
+        # sudo: res_company - can access all companies for auto-subscribing
+        company_ids = self.sudo().auto_subscribe_company_ids.ids
+        if self.group_public_id:
+            user_domain &= Domain("all_group_ids", "in", self.group_public_id.ids)
+        if company_ids:
+            user_domain &= Domain("company_ids", "in", company_ids)
+        if self.group_ids:
+            user_domain &= Domain("all_group_ids", "in", self.group_ids.ids)
+        return user_domain
 
     def _get_member_join_notification(self, member):
         if member.is_self:
--- a/addons/mail/models/discuss/discuss_channel.py
+++ b/addons/mail/models/discuss/discuss_channel.py
@@ -112,11 +112,14 @@ class DiscussChannel(models.Model):
         index=True,
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_join = fields.Boolean(string="Auto Join", default=False)
     group_ids = fields.Many2many(
         'res.groups', string='Auto Subscription',
         help="Members of those groups will automatically added as followers. "
              "Note that they will be able to manage their subscription manually "
              "if necessary.")
+    company_ids = fields.Many2many("res.company", string="Companies")
+    auto_joined_partner_ids = fields.Many2many("res.partner", "auto_joined_partners")
     # access
     uuid = fields.Char('UUID', size=50, default=_generate_random_token, copy=False)
     group_public_id = fields.Many2one('res.groups', string='Authorized Group', compute='_compute_group_public_id', recursive=True, readonly=False, store=True)
@@ -186,12 +189,21 @@ class DiscussChannel(models.Model):
             if len(ch.channel_member_ids) > 2:
                 raise ValidationError(_("A channel of type 'chat' cannot have more than two users."))
 
-    @api.constrains('group_public_id', 'group_ids')
+    @api.constrains("group_public_id", "auto_join", "group_ids")
     def _constraint_group_id_channel(self):
         # sudo: discuss.channel - skipping ACL for constraint, more performant and no sensitive information is leaked
-        failing_channels = self.sudo().filtered(lambda channel: channel.channel_type != 'channel' and (channel.group_public_id or channel.group_ids))
+        failing_channels = self.sudo().filtered(
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_join or channel.group_ids
+            ),
+        )
         if failing_channels:
-            raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the group-based authorization or group auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
+            raise ValidationError(
+                _(
+                    "For %(channels)s, channel_type should be 'channel' to have the group-based authorization or auto-join or groups",
+                    channels=', '.join([ch.name for ch in failing_channels]),
+                ),
+            )
 
     # COMPUTE / INVERSE
 
@@ -547,7 +559,7 @@ class DiscussChannel(models.Model):
                         )
                     )
         result = super().write(vals)
-        if vals.get('group_ids'):
+        if {"auto_join", "company_ids", "group_ids", "group_public_id"} & set(vals):
             self._subscribe_users_automatically()
         return result
 
@@ -569,7 +581,10 @@ class DiscussChannel(models.Model):
     # ------------------------------------------------------------
 
     def _subscribe_users_automatically(self):
-        if not (new_members_to_create := self._subscribe_users_automatically_get_members()):
+        channels = self.filtered(lambda c: c.auto_join)
+        if not channels:
+            return
+        if not (new_members_to_create := channels._subscribe_users_automatically_get_members()):
             return
         to_create = [
             {"channel_id": channel_id, "partner_id": partner_id}
@@ -579,7 +594,9 @@ class DiscussChannel(models.Model):
         # sudo: discuss.channel.member - adding member of other users based on channel auto-subscribe
         new_members = self.env["discuss.channel.member"].sudo().create(to_create)
         stores = Store.Stores()
+        auto_joined_members = []
         for member, store in new_members._get_member_store_list(stores):
+            auto_joined_members.append((member.channel_id.id, member.partner_id.id))
             store.add(member.channel_id, "_store_channel_fields").add(
                 member,
                 lambda res: (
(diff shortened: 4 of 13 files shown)


addons/hr/models/discuss_channel.py
```diff
@@ -18,15 +19,19 @@ def _constraint_subscription_department_ids_channel(self):
...
-    def _subscribe_users_automatically_get_members(self):
-        """ Auto-subscribe members of a department to a channel """
-        new_members = super()._subscribe_users_automatically_get_members()
-        for channel in self:
-            new_members[channel.id] = list(
-                set(new_members[channel.id]) |
-                set((channel.subscription_department_ids.member_ids.user_id.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids)
+    def _get_extra_domain(self):
+        domain = super()._get_extra_domain()
+        # sudo: hr.department - ensure all departments and its members are accesible
```
oomsveta, line 24:
```suggestion
        # sudo: hr.department - ensure all departments and their members are accessible
```

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -186,12 +189,21 @@ def _constraint_partners_chat(self):
...
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_join or channel.group_ids
+            ),
+        )
         if failing_channels:
-            raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the group-based authorization or group auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
+            raise ValidationError(
+                _(
+                    "For %(channels)s, channel_type should be 'channel' to have the group-based authorization or auto-join or groups",
+                    channels=', '.join([ch.name for ch in failing_channels]),
```
oomsveta, line 204: gettext automatically joins lists using the separator relevant to the current locale
```suggestion
                    channels=[ch.name for ch in failing_channels],
```

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -186,12 +189,21 @@ def _constraint_partners_chat(self):
...
+        failing_channels = self.sudo().filtered(
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_join or channel.group_ids
+            ),
+        )
         if failing_channels:
-            raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the group-based authorization or group auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
+            raise ValidationError(
+                _(
+                    "For %(channels)s, channel_type should be 'channel' to have the group-based authorization or auto-join or groups",
```
oomsveta, line 203:
```suggestion
                    "For %(channels)s, channel_type should be 'channel' to have the group-based authorization or auto-subscription",
```

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -112,11 +112,14 @@ def _generate_random_token(self):
...
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_join = fields.Boolean(string="Auto Join", default=False)
     group_ids = fields.Many2many(
         'res.groups', string='Auto Subscription',
         help="Members of those groups will automatically added as followers. "
              "Note that they will be able to manage their subscription manually "
              "if necessary.")
+    company_ids = fields.Many2many("res.company", string="Companies")
+    auto_joined_partner_ids = fields.Many2many("res.partner", "auto_joined_partners")
```
oomsveta, line 122: Could you add a `help` attribute to explain the purpose of these fields? Thanks!

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -588,14 +605,23 @@ def _subscribe_users_automatically(self):
...
 
     def _subscribe_users_automatically_get_members(self):
         """ Return new members per channel ID """
-        return dict(
-            (channel.id,
-             ((channel.group_ids.all_user_ids.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids))
-                for channel in self
-            )
+        partners_by_channel = {}
+        users = self.env["res.users"].search_fetch([], ["partner_id"])
```
oomsveta, line 618: I may be wrong, but I feel like not fetching all the fields (in this case, only fetching the `partner_id`) could easily backfire. If the rest of the code were to access any other field, it'd have to run another query, which would negate the benefit of the optimization

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -731,6 +757,27 @@ def _add_members(
...
+        if company_ids:
+            user_domain &= Domain("company_ids", "in", company_ids)
+        user_domain &= self._get_extra_domain()
+        return user_domain
+
+    def _get_extra_domain(self):
+        domain = Domain(True)
+        if self.group_ids:
+            domain &= Domain("all_group_ids", "in", self.group_ids.ids)
+        return domain
```
oomsveta, line 779: Can you explain why this function is separate from `_get_auto_subscribe_domain`?

addons/mail/models/discuss/res_users.py
```diff
@@ -25,7 +27,15 @@ def write(self, vals):
             user_group_ids = [command[1] for command in vals["group_ids"] if command[0] == 4]
             user_group_ids += [id for command in vals["group_ids"] if command[0] == 6 for id in command[2]]
             user_group_ids = self.env['res.groups'].browse(user_group_ids).all_implied_ids._ids
-            self.env["discuss.channel"].search([("group_ids", "in", user_group_ids)])._subscribe_users_automatically()
+            self.env["discuss.channel"].search([
+                ("auto_join", "=", True),
+                "|", ("group_public_id", "in", user_group_ids),
```
oomsveta, line 32: Could you explain why `group_public_id` wasn't part of the domain before, yet it is now?

addons/mail/models/discuss/discuss_channel.py
oomsveta, line 122: If this field is not meant to be displayed to the end user, you should add `export_string_translation=False` as to filter it out of the .po(t) files

addons/hr/models/discuss_channel.py
```diff
@@ -18,15 +19,19 @@ def _constraint_subscription_department_ids_channel(self):
...
+        departments_sudo = self.sudo().subscription_department_ids
+        if departments_sudo:
+            department_domain = Domain(
+                'id', 'in', departments_sudo.member_ids.user_id.ids,
             )
-        return new_members
+            if self.group_ids:
+                domain |= department_domain
+            else:
+                domain &= department_domain
```
oomsveta, line 33: Could you explain the rationale for ORing vs ANDing the domain?

addons/mail/views/discuss_channel_views.xml
```diff
@@ -103,12 +103,17 @@
...
+                                    <field name="group_public_id" placeholder="Public - anyone with the link can access"
+                                        readonly="not is_editable"/>
+                                </group>
+                                <group string="Auto-Join Rules">
+                                        <field name="auto_join" string="Auto-Join channel"/>
+                                        <field name="company_ids" widget="many2many_tags" placeholder="All companies"
+                                                string="Companies" readonly="not is_editable" invisible="not auto_join"/>
+                                        <field name="group_ids" widget="many2many_tags" placeholder="All groups"
+                                                string="Groups" domain="[('all_implied_ids', 'in', group_public_id)]"
+                                                readonly="not is_editable" invisible="not auto_join"/>
```
oomsveta, line 116: too many indents

oomsveta (commented):
Hello, thank you for the PR 🙂

addons/mail/models/discuss/discuss_channel.py
shmn-odoo: On large tables like `res.users`, there are many fields. Prefetching all of them for every record will really slow things down. Plus, we don't actually need most of those fields, so it's better not to prefetch them all.

Because of that, I think we should just explicitly prefetch the fields we need.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -112,11 +112,14 @@ def _generate_random_token(self):
         index=True,
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_join = fields.Boolean(string="Auto Join", default=False)
     group_ids = fields.Many2many(
         'res.groups', string='Auto Subscription',
         help="Members of those groups will automatically added as followers. "
              "Note that they will be able to manage their subscription manually "
              "if necessary.")
+    company_ids = fields.Many2many("res.company", string="Companies")
```
seb-odoo, line 121: Note: the names `group_ids` and `group_public_id` have been confusing for years.

I'd take this opportunity to introduce a more explicit name at least for the new field.

Something like `auto_join_company_ids` or `auto_subscribe_company_ids`

Note: I'm not a fan of the `auto_join` name/string in general, as joining implies to be an action from the current user, while auto is specifically not an action from the current user.

I'd challenge the spec to rename this field/string in all places, to something such as auto subscribe or similar (where "subscribing" has to be understood as "add member").

Note that we usually use the term "invite" (or just "add" member) when the action is coming from another person. But in this case as it is automatic, I'm not sure if invite is a good fit either.

addons/hr/models/discuss_channel.py
shmn-odoo: When `group_ids` is set, the domain returned by the parent method contains group domain, which needs an `OR` with department domain (`_get_auto_subscribe_domain` `AND` (`group domain` `OR` `department domain`)).

When `group_ids` is not set, the department domain needs to be combined with the other domain (`_get_auto_subscribe_domain` `AND` `department domain`).

addons/mail/models/discuss/res_users.py
shmn-odoo: We are adding a separate boolean field to enable or disable auto-subscription for a channel. When enabled, all authorized users (those in the `group_public_id`) will join the channel, even if other fields are not set. Previously, either the groups or departments field had to be set for this to work.

For example, if a channel has an authorized group and auto-subscription is enabled, a user now qualifies for auto-subscribe just by being authorized, even if the other auto-subscribe fields are empty.

This means channels with a matching `group_public_id` and auto-subscribe enabled now also qualify as a potential match.

addons/mail/models/discuss/discuss_channel.py
shmn-odoo: We want to subscribe users who are either part of the group or the department. To do this, we need to add an `OR` to the `_get_auto_subscribe_domain` method in the hr override.

However, that doesn't work correctly because it mixes up the logic by applying the `OR` to the whole domain instead of just the group part.

I could have rewritten the entire domain by hand, but that would mean taking out the `group_ids` and putting them back in with the new department domain. That looked very messy.

So, I created this function to focus specifically on the group and department domains.

addons/mail/models/discuss/discuss_channel.py
shmn-odoo: > Note: the names group_ids and group_public_id have been confusing for years.

It doesn’t necessarily have to be part of this task/PR, but `authorized_group_id` or `allowed_group_id` seem like better alternatives to `group_public_id` at first glance.

The `group_ids` should share the same prefix as the other auto-subscribing fields, so we might want to update that in this PR.

> Something like auto_join_company_ids or auto_subscribe_company_ids

I don’t think `auto_subscribe` is a good prefix, as it would make the field names longer and the code harder to read. That said, I don’t have any better alternatives either :(

shmn-odoo: How about using `subscription` as the prefix, similar to the [department](https://github.com/odoo/odoo/blob/master/addons/hr/models/discuss_channel.py#L11) field? 

I find it to be a good compromise between clear naming and readability.

seb-odoo: I find `auto_subscribe` more explicit than `subscription`, but both are better than no prefix for sure.

shmn-odoo: > The `group_ids` should share the same prefix as the other auto-subscribing fields, so we might want to update that in this PR.

Actually, we’re better off handling this in a dedicated PR:

1. Let’s merge the current PR first.
2. A cleanup PR (we are currently sending the `group_ids` field to the client to prevent users from leaving auto-subscribing channels, however, this should no longer be needed as we now keep this history).
3. Then, rename the remaining occurrences in this [PR](https://github.com/odoo/odoo/pull/262504).

seb-odoo: Good for me!

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -129,11 +129,19 @@ def _generate_random_token(self):
         index=True,
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_subscribe = fields.Boolean(
+        string="Auto Subscribe", default=False,
+        help="Auto-subscribe authorized users to this channel. "
+             "Users can leave anytime and won't be re-added automatically.")
     group_ids = fields.Many2many(
```
seb-odoo, line 136: Note: I think it was already posted but we should rename it eventually to be consistent with the new one with `auto_subscribe_`. Ok to do as a follow up PR.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -129,11 +129,19 @@ def _generate_random_token(self):
...
         'res.groups', string='Auto Subscription',
-        help="Members of those groups will automatically added as followers. "
-             "Note that they will be able to manage their subscription manually "
-             "if necessary.")
+        help="Restrict auto-subscribe users to the selected groups.")
+    auto_subscribe_company_ids = fields.Many2many(
+        "res.company", string="Companies",
+        help="Restrict auto-subscribe users to the selected companies.")
```
seb-odoo, line 141: formatter/linter: this should have trailing commas and have one param per line (all of them)

addons/hr/tests/test_channel.py
```diff
@@ -42,3 +44,13 @@ def test_auto_subscribe_when_updating_employee_department(self):
...
+
+    def test_auto_subscribe_group_department(self):
+        self.department.invalidate_recordset(['member_ids'])
+        self.emp0.write({'department_id': self.department.id})
+        self.channel.write({
+            'auto_subscribe': True,
+            'group_ids': [(4, self.env.ref("base.group_user").id)],
+            'subscription_department_ids': [(4, self.department.id)],
+        })
+        self.assertEqual(self.channel.channel_partner_ids, self.emp0.user_id.partner_id)
```
seb-odoo, line 56: It's not obvious that the test explicitly excludes users that are not part of either the group of the department. So it can actually pass just because there are no other users available, filters or not.

Either it should have more users explicitly, or probably better we should remove the group and then the department as last steps of the test and see we progressively get more users.

addons/hr/tests/test_channel.py
```diff
@@ -42,3 +44,13 @@ def test_auto_subscribe_when_updating_employee_department(self):
         self.emp0.write({'department_id': self.department.id})
 
         self.assertEqual(self.channel.channel_partner_ids, self.emp0.user_id.partner_id)
+
+    def test_auto_subscribe_group_department(self):
+        self.department.invalidate_recordset(['member_ids'])
+        self.emp0.write({'department_id': self.department.id})
+        self.channel.write({
```
seb-odoo, line 51: nitpick, less important as not in our module, but this doesn't follow the ruff format

addons/hr/tests/test_channel.py
```diff
@@ -42,3 +44,13 @@ def test_auto_subscribe_when_updating_employee_department(self):
         self.emp0.write({'department_id': self.department.id})
 
         self.assertEqual(self.channel.channel_partner_ids, self.emp0.user_id.partner_id)
+
+    def test_auto_subscribe_group_department(self):
+        self.department.invalidate_recordset(['member_ids'])
```
seb-odoo, line 49: This kind of invalidate in a test is never a good sign. If it's due to a manual query, the invalidate should actually be in the business code before the query.
In either case, if we keep it it's worth a comment to explain it.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -129,11 +129,19 @@ def _generate_random_token(self):
         index=True,
         help="Contains the date and time of the last interesting event that happened in this channel. This updates itself when new message posted.",
     )
+    auto_subscribe = fields.Boolean(
```
seb-odoo, line 132: To be checked:
as we search on this, it should probably be indexed (only when True, low cardinality, indexing False is useless as it's almost all channels)

But if all the queries include the other fields in them, maybe the planner will always prioritize the relation indexes, I don't know.

But maybe we can just add it regardless, index when True will be cheap and can't really hurt.

addons/hr/models/discuss_channel.py
```diff
@@ -3,30 +3,29 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.fields import Domain
 
 
 class DiscussChannel(models.Model):
     _inherit = 'discuss.channel'
 
     subscription_department_ids = fields.Many2many(
```
seb-odoo, line 12: Note: same for this field, we can eventually rename to match others (not in this PR)

addons/hr/models/discuss_channel.py
```diff
@@ -3,30 +3,29 @@
...
-                set(new_members[channel.id]) |
-                set((channel.subscription_department_ids.member_ids.user_id.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids)
-            )
-        return new_members
+    def _get_auto_subscribe_domain(self):
+        domain = super()._get_auto_subscribe_domain()
+        # sudo: hr.department - ensure all departments and their members are accessible
+        departments_sudo = self.sudo().subscription_department_ids
+        if departments_sudo:
+            domain &= Domain('id', 'in', departments_sudo.member_ids.user_id.ids)
```
seb-odoo, line 27: Are we sure about the removal of the active flag filtering?
What is the reasoning?

addons/hr/models/discuss_channel.py
```diff
@@ -3,30 +3,29 @@
...
-            new_members[channel.id] = list(
-                set(new_members[channel.id]) |
-                set((channel.subscription_department_ids.member_ids.user_id.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids)
-            )
-        return new_members
+    def _get_auto_subscribe_domain(self):
+        domain = super()._get_auto_subscribe_domain()
+        # sudo: hr.department - ensure all departments and their members are accessible
+        departments_sudo = self.sudo().subscription_department_ids
+        if departments_sudo:
```
seb-odoo, line 26:
```suggestion
        if departments_sudo := self.sudo().subscription_department_ids:
```

addons/hr/views/discuss_channel_views.xml
```diff
@@ -5,10 +5,9 @@
         <field name="model">discuss.channel</field>
         <field name="inherit_id" ref="mail.discuss_channel_view_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='group_ids']" position="after">
+            <xpath expr="//field[@name='auto_subscribe_company_ids']" position="after">
                 <field name="subscription_department_ids" widget="many2many_tags"
-                    invisible="channel_type != 'channel'"
-                    string="Auto Subscribe Departments"/>
+                    placeholder="All departments" string="Departments" invisible="not auto_subscribe"/>
```
seb-odoo, line 10: Wouldn't we want to update the field string in the model to be consistent rather than defining a specific string in this view?

addons/mail/models/discuss/res_groups.py
```diff
@@ -9,5 +9,9 @@ class ResGroups(models.Model):
     def write(self, vals):
         res = super().write(vals)
         if vals.get("user_ids"):
-            self.env["discuss.channel"].search([("group_ids", "in", self.all_implied_ids._ids)])._subscribe_users_automatically()
+            self.env["discuss.channel"].search_fetch([
+                ("auto_subscribe", "=", True),
+                "|", ("group_ids", "in", self.all_implied_ids._ids),
```
seb-odoo, line 14: I prefer Domain operations now rather than the old syntax, especially when OR are involved.

addons/mail/models/discuss/res_groups.py
```diff
@@ -9,5 +9,9 @@ class ResGroups(models.Model):
     def write(self, vals):
         res = super().write(vals)
         if vals.get("user_ids"):
-            self.env["discuss.channel"].search([("group_ids", "in", self.all_implied_ids._ids)])._subscribe_users_automatically()
+            self.env["discuss.channel"].search_fetch([
```
seb-odoo, line 12: for the whole diff: apply formatter on changed lines

addons/mail/models/discuss/res_users.py
```diff
@@ -13,19 +13,29 @@ class ResUsers(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         users = super().create(vals_list)
-        self.env["discuss.channel"].search([("group_ids", "in", users.all_group_ids.ids)])._subscribe_users_automatically()
+        self.env["discuss.channel"].search_fetch([
+            ("auto_subscribe", "=", True),
```
seb-odoo, line 17: we definitely need the index for this one, as there are no other criteria

also this is a potential scaling problem, but ok for now, we can hope database will stay low on amount of auto subscribe channels

addons/mail/models/discuss/res_users.py
```diff
@@ -13,19 +13,29 @@ class ResUsers(models.Model):
...
+        if {"company_ids", "group_ids"} & vals.keys():
+            domain = Domain("auto_subscribe", "=", True)
+            if vals.get("company_ids"):
+                domain &= Domain("auto_subscribe_company_ids", "in", self.company_ids.ids)
+            if vals.get("group_ids"):
+                # form: {'group_ids': [(3, 10), (3, 3), (4, 10), (4, 3)]} or {'group_ids': [(6, 0, [ids]}
+                user_group_ids = [command[1] for command in vals["group_ids"] if command[0] == 4]
+                user_group_ids += [id for command in vals["group_ids"] if command[0] == 6 for id in command[2]]
+                user_group_ids = self.env['res.groups'].browse(user_group_ids).all_implied_ids._ids
+                domain &= Domain.OR([
```
seb-odoo, line 34: Domain.OR is only useful for dynamic lists, this could just `|` the 2 manual Domain below.

addons/mail/models/discuss/res_users.py
```diff
@@ -13,19 +13,29 @@ class ResUsers(models.Model):
...
-        if vals.get("group_ids"):
-            # form: {'group_ids': [(3, 10), (3, 3), (4, 10), (4, 3)]} or {'group_ids': [(6, 0, [ids]}
-            user_group_ids = [command[1] for command in vals["group_ids"] if command[0] == 4]
-            user_group_ids += [id for command in vals["group_ids"] if command[0] == 6 for id in command[2]]
-            user_group_ids = self.env['res.groups'].browse(user_group_ids).all_implied_ids._ids
-            self.env["discuss.channel"].search([("group_ids", "in", user_group_ids)])._subscribe_users_automatically()
+        if {"company_ids", "group_ids"} & vals.keys():
+            domain = Domain("auto_subscribe", "=", True)
+            if vals.get("company_ids"):
+                domain &= Domain("auto_subscribe_company_ids", "in", self.company_ids.ids)
```
seb-odoo, line 28: Why for company can we only check ids simply, and for groups we have to parse the commands?

The lack of consistency is strange, is one optimized and not the other?

Isn't this lack of optimization why we have more queries in `odoo/addons/base/tests/test_res_users.py`? It would be nice to:

1. reduce the queries of possible
2. list the extra one we must have with a justification (at least in the commit)

addons/mail/tests/discuss/test_discuss_channel_auto_subscribe.py
```diff
@@ -0,0 +1,93 @@
...
+from odoo.tests import TransactionCase
+
+
+class TestChannelAutoSubscribe(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.group_user = cls.env.ref("base.group_user")
+        cls.group_admin = cls.env.ref("base.group_system")
+        cls.user_admin = cls.env.ref('base.user_admin')
```
seb-odoo, line 12: Most of those can be found in mail common, it would be better to extend it rather than bare transaction case.

addons/mail/tests/discuss/test_discuss_channel_auto_subscribe.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+        cls.both_partners = cls.user_a.partner_id | cls.user_b.partner_id
+        cls.all_partners = cls.user_admin.partner_id | cls.both_partners
+
+    def test_channel_auto_subscribe_all_users(self):
+        channel = self.env["discuss.channel"].create({
+            "name": "Test all auto subscribe",
+            "auto_subscribe": True,
+        })
+        self.assertEqual(channel.channel_partner_ids, self.all_partners)
```
seb-odoo, line 36: This is going to fail as soon as there is demo data or unexpected users, we need to assert it differently.

addons/mail/tests/discuss/test_discuss_channel_auto_subscribe.py
```diff
@@ -0,0 +1,93 @@
...
+        channel = self.env["discuss.channel"].create({
+            "name": "Test company auto subscribe",
+            "auto_subscribe": True,
+            "auto_subscribe_company_ids": [Command.link(self.company_a.id)],
+        })
+        self.assertEqual(channel.channel_partner_ids, self.user_a.partner_id)
+
+        channel.write({"auto_subscribe_company_ids": [Command.link(self.company_b.id)]})
+        self.assertEqual(
+            channel.channel_partner_ids, self.both_partners,
```
seb-odoo, line 48: If using company shared by mail common, this kind of assert might fail too if other unexpected users share the company too.

Currently this one is fine as the company is created in the test, so its users is a controlled set.

addons/mail/tests/discuss/test_discuss_channel_auto_subscribe.py
```diff
@@ -0,0 +1,93 @@
...
+        channel = self.env["discuss.channel"].create({
+            "name": "Test group auto subscribe",
+            "auto_subscribe": True,
+            "group_ids": [Command.link(self.group_admin.id)],
+        })
+        self.assertEqual(
+            channel.channel_partner_ids, self.user_admin.partner_id | self.user_a.partner_id,
+        )
+        channel.write({"group_ids": [Command.link(self.group_user.id)]})
+        self.assertEqual(channel.channel_partner_ids, self.all_partners)
```
seb-odoo, line 61: same remark for all tests basically

addons/mail/tests/discuss/test_discuss_channel_auto_subscribe.py
```diff
@@ -0,0 +1,93 @@
...
+        # 'User A' leaves the channel
+        channel.with_user(self.user_a).action_unfollow()
+        self.assertEqual(
+            channel.channel_partner_ids, self.user_admin.partner_id,
+        )
+        channel.write({"group_ids": [Command.link(self.group_user.id)]})
+        self.assertEqual(
+            channel.channel_partner_ids, self.user_admin.partner_id | self.user_b.partner_id,
+        )
+        self.assertNotIn(self.user_a.partner_id, channel.channel_partner_ids)
```
seb-odoo, line 93: It's not clear from the test if this works thanks to `auto_subscribed_partner_ids` or just because the `channel.write({"group_ids": [Command.link(self.group_user.id)]})` line has no effect.

Maybe it would be best to then as confirmation remove the user from `auto_subscribed_partner_ids`, re-do the same kind of change, and see the user is added back.

addons/mail/views/discuss_channel_views.xml
```diff
@@ -103,12 +103,16 @@
...
                             <page string="Privacy" name="privacy" invisible="channel_type != 'channel'">
-                                <group class="o_label_nowrap">
-                                    <field name="group_public_id"
-                                        invisible="channel_type != 'channel' or parent_channel_id" readonly="not is_editable"/>
-                                    <field name="group_ids" widget="many2many_tags"
-                                        invisible="channel_type != 'channel'"
-                                        string="Auto Subscribe Groups" readonly="not is_editable"/>
+                                <group string="Channel Visibility" invisible="parent_channel_id">
+                                    <field name="group_public_id" placeholder="Public - anyone with the link can access"
+                                        readonly="not is_editable"/>
```
seb-odoo, line 108: while we have diff on them, I would keep a single line here, it feels off to just have one attribute out when it's not too long (and if it's too long, it's usually better to have one attribute per line)

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -203,12 +211,20 @@ def _constraint_partners_chat(self):
...
                 raise ValidationError(_("A channel of type 'chat' cannot have more than two users."))
 
-    @api.constrains('group_public_id', 'group_ids')
+    @api.constrains("auto_subscribe", "group_public_id", "group_ids")
     def _constraint_group_id_channel(self):
         # sudo: discuss.channel - skipping ACL for constraint, more performant and no sensitive information is leaked
-        failing_channels = self.sudo().filtered(lambda channel: channel.channel_type != 'channel' and (channel.group_public_id or channel.group_ids))
+        failing_channels = self.sudo().filtered(
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_subscribe or channel.group_ids
```
seb-odoo, line 219: This missing a check for company (+ dependency). We don't want auto subscribe of any kind.

It would have no effect as auto_subscribe would be false, but better keep it clean.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -594,7 +610,9 @@ def write(self, vals):
                         )
                     )
         result = super().write(vals)
-        if vals.get('group_ids'):
+        if {
+            "auto_subscribe", "auto_subscribe_company_ids", "group_ids", "group_public_id",
+        } & set(vals):
```
seb-odoo, line 615: strange to write `set(vals)`, isn't it just `vals.keys()`? if so, we usually write the later for clarity (from my memory keys() has a set-like API), as otherwise we have to guess what casting to set does

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -743,22 +764,33 @@ def _subscribe_users_automatically(self):
         ]
         # sudo: discuss.channel.member - adding member of other users based on channel auto-subscribe
         new_members = self.env["discuss.channel.member"].sudo().create(to_create)
+        auto_subscribed_members = []
```
seb-odoo, line 767: This list is very close to `new_members_to_create`, do we really need to build a new one?

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -743,22 +764,33 @@ def _subscribe_users_automatically(self):
...
         for member, store in new_members._get_member_store_list():
+            auto_subscribed_members.append((member.channel_id.id, member.partner_id.id))
             store.add(member.channel_id, "_store_channel_fields").add(
                 member,
                 lambda res: (
                     res.from_method("_store_persona_default_fields"),
                     res.attr("unpin_dt"),
                 ),
             )
+        self.env.cr.execute_values("""
```
seb-odoo, line 777: Manual query should have invalidate before them (cf comment in test)

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -743,22 +764,33 @@ def _subscribe_users_automatically(self):
...
+            INSERT INTO auto_subscribed_partners (discuss_channel_id, res_partner_id)
+                 VALUES %s
+            ON CONFLICT DO NOTHING
```
seb-odoo, line 780: Just to confirm, the query is just for this line right?

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -743,22 +764,33 @@ def _subscribe_users_automatically(self):
...
 
     def _subscribe_users_automatically_get_members(self):
         """ Return new members per channel ID """
-        return dict(
-            (channel.id,
-             ((channel.group_ids.all_user_ids.partner_id.filtered(lambda p: p.active) - channel.channel_partner_ids).ids))
-                for channel in self
-            )
+        partners_by_channel = {}
+        users = self.env["res.users"].search_fetch([])
```
seb-odoo, line 787: This is definitely not ok. This will just fetch all users of the DB which can be a lot.

As I posted on the chatter of the task, we probably need a cron to batch (100 to 1000) delay it.

That is, in the case where there is no post-filtering from other criteria.

But when we have post-filtering, we should build that into the initial search instead of filtering in python.
Even after we do that, we can still have too many users though, so we probably need a cron regardless.

\+ as suggested also on the chatter, add an indicator in the view to say how many users the current domain will add (maybe with a search count limit 1000+, as more results than that can take a long time even just to count, and we don't need to display the exact number, just that it is very high)

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -743,22 +764,33 @@ def _subscribe_users_automatically(self):
...
                 ),
             )
+        self.env.cr.execute_values("""
+            INSERT INTO auto_subscribed_partners (discuss_channel_id, res_partner_id)
+                 VALUES %s
+            ON CONFLICT DO NOTHING
+            """, auto_subscribed_members,
+        )
 
     def _subscribe_users_automatically_get_members(self):
```
seb-odoo, line 784: It's already an issue before this PR but the name of this method is confusing: it doesn't return members, but a dict of partners by channel.

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -905,6 +937,22 @@ def _add_members(
                     channel.self_member_id.sudo()._rtc_invite_members(member_ids=new_members.ids)
         return all_new_members
 
+    def _get_auto_subscribe_domain(self):
+        joined_partners = self.channel_member_ids.partner_id | self.auto_subscribed_partner_ids
```
seb-odoo, line 941: This is also bad for performance: it will force a fetch of all existing members and all auto_subscribed_partner_ids partners, which can both be a lot.
This should be part of a domain (or a manual query), and not done in python.

addons/mail/models/discuss/res_users.py
```diff
@@ -13,19 +13,29 @@ class ResUsers(models.Model):
...
+            if vals.get("group_ids"):
+                # form: {'group_ids': [(3, 10), (3, 3), (4, 10), (4, 3)]} or {'group_ids': [(6, 0, [ids]}
+                user_group_ids = [command[1] for command in vals["group_ids"] if command[0] == 4]
+                user_group_ids += [id for command in vals["group_ids"] if command[0] == 6 for id in command[2]]
+                user_group_ids = self.env['res.groups'].browse(user_group_ids).all_implied_ids._ids
+                domain &= Domain.OR([
+                    [("group_public_id", "in", user_group_ids)],
+                    [("group_ids", "in", user_group_ids)],
+                ])
+            self.env["discuss.channel"].search_fetch(domain)._subscribe_users_automatically()
```
seb-odoo, line 38: company and group conditions are `&=`; so a write carrying both misses channels that match only one

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -203,12 +211,20 @@ def _constraint_partners_chat(self):
...
         # sudo: discuss.channel - skipping ACL for constraint, more performant and no sensitive information is leaked
-        failing_channels = self.sudo().filtered(lambda channel: channel.channel_type != 'channel' and (channel.group_public_id or channel.group_ids))
+        failing_channels = self.sudo().filtered(
+            lambda channel: channel.channel_type != "channel" and (
+                channel.group_public_id or channel.auto_subscribe or channel.group_ids
+            ),
+        )
         if failing_channels:
-            raise ValidationError(_("For %(channels)s, channel_type should be 'channel' to have the group-based authorization or group auto-subscription.", channels=', '.join([ch.name for ch in failing_channels])))
+            raise ValidationError(_(
```
seb-odoo, line 223: translation can be read from `self.env._` now

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -203,12 +211,20 @@ def _constraint_partners_chat(self):
...
+                "For %(channels)s, channel_type should be 'channel' to have the group-based "
+                "authorization or auto-subscription",
+                channels=[ch.name for ch in failing_channels],
```
seb-odoo, line 226: I guess this is the definition of mapped:
```suggestion
                channels=failing_channels.mapped("name"),
```

addons/mail/models/discuss/discuss_channel.py
```diff
@@ -905,6 +937,22 @@ def _add_members(
                     channel.self_member_id.sudo()._rtc_invite_members(member_ids=new_members.ids)
         return all_new_members
 
+    def _get_auto_subscribe_domain(self):
```
seb-odoo, line 940: This looks like it builds a domain for a specific channel, so we might want to ensure_one to guarantee it. (or batch if that's not intended)

addons/mail/views/discuss_channel_views.xml
```diff
@@ -103,12 +103,16 @@
...
+                                </group>
+                                <group string="Auto-Subscribe Rules">
+                                    <field name="auto_subscribe" string="Auto-Subscribe to channel"/>
```
seb-odoo, line 111: missing `readonly="not is_editable` to be consistent with the rest

addons/hr/views/discuss_channel_views.xml
seb-odoo, line 10: missing `readonly="not is_editable` to be consistent with the rest

addons/mail/views/discuss_channel_views.xml
seb-odoo: Also it looks like standard users can edit those fields and still save the view on a read-only channel. Normally other fields are prevented? To double check but can be fixed separately.

PR #247718, opened 2026-02-09
[FIX] purchase_requisition: correct company for blanket order vendor pricelist

derick-david:
In a multi-company setup, when confirming a Blanket Order (Purchase Requisition), the resulting Vendor Pricelist (product.supplierinfo) entry was created in the current environment company instead of the requestion's company.

Current behavior before PR:
The 'product.supplierinfo' record is created with 'company_id' equal to the current user's active company, regardless of the 'company_id' on the Purchase Requisition.

The 'product.supplierinfo' record is created with 'company_id' explicitly set from the Purchase Requisition's 'company_id'.

---

### DIFF
--- a/doc/cla/individual/derick-david.md
+++ b/doc/cla/individual/derick-david.md
@@ -1,0 +1,11 @@
+Dubai, 2026-02-06
+
+I hereby agree to the terms of the Odoo Individual Contributor License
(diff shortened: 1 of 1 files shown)


derick-david:
Any chance this might get merged ?

PR #247720, opened 2026-02-09
[FIX] l10n_sa_edi: handle negative quantity as discount in XML

matd-odoo:
Before this commit, using a negative quantity for a discount line caused ZATCA
validation warnings (BR-S-08, BR-CO-10 and BR-CO-13), as it was not handled
like a negative price.

This commit fixes the XML generation to treat negative quantities identically to
negative unit prices, ensuring correct totals.

task-5883377

### DIFF
--- a/addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
+++ b/addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
@@ -337,12 +337,13 @@ class AccountEdiXmlUBL21Zatca(models.AbstractModel):
 
     def _l10n_sa_postprocess_line_vals(self, vals):
         """
-            Postprocess vals to remove negative line amounts, as those will be used to compute
-            document level allowances (global discounts)
+            Postprocess vals to remove negative lines (treated as Global Discounts),
+            unless flagged by 'is_line_retained' (e.g., Down Payments) to be retained
+            as Prepayments in the XML.
         """
         final_line_vals = []
         for line_vals in vals['vals']['line_vals']:
-            if line_vals['price_vals']['price_amount'] >= 0:
+            if line_vals['price_vals']['price_amount'] >= 0 and line_vals['is_line_retained']:
                 final_line_vals.append(line_vals)
         vals['vals'][('line_vals')] = final_line_vals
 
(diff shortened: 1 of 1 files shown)


addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
```diff
@@ -342,7 +342,7 @@ def _l10n_sa_postprocess_line_vals(self, vals):
         """
         final_line_vals = []
         for line_vals in vals['vals']['line_vals']:
-            if line_vals['price_vals']['price_amount'] >= 0:
+            if line_vals['price_vals']['price_amount'] >= 0 and line_vals['line_selection']:
```
khva-odoo: I think we should modify the doc string in the method according to the logic

addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
```diff
@@ -420,15 +420,18 @@ def grouping_key_generator(base_line, tax_values):
...
             line_vals['prepayment_vals'] = self._l10n_sa_get_line_prepayment_vals(line, taxes_vals)
         else:
             # - BR-KSA-80: only down-payment lines should have a tax subtotal breakdown, as that is
             # used during computation of prepaid amount as ZATCA sums up tax amount/taxable amount of all lines
             # irrespective of whether they are down-payment lines.
+            line_selection = False
             line_vals['tax_total_vals'][0].pop('tax_subtotal_vals', None)
         line_vals['tax_total_vals'][0]['total_amount_sa'] = total_amount_sa
+        line_vals['line_selection'] = line_selection or line_vals['line_quantity'] >= 0
         line_vals['line_quantity'] = abs(line_vals['line_quantity'])
```
khva-odoo, line 436:
```suggestion
        line_vals['line_quantity'] = line_vals['line_quantity']
```
We can make the line quantity abs in the `_l10n_sa_postprocess_line_vals`. By doing so we eliminate the need for have the new variable.

khva-odoo (commented):
Hello,
Thanks for the PR!

addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
matd-odoo: I’ve already tested this approach, but the [down payment](https://github.com/odoo/odoo/blob/17.0/addons/l10n_sa_edi/tests/test_edi_zatca.py#L239) test case fails. The issue is that a down payment line can also have -ve quantity, while it is still considered a regular invoice line rather than a discount.

For this reason, we likely need a dedicated variable to clearly distinguish between down payment lines and actual discount lines with -ve quantity.

addons/l10n_sa_edi/models/account_edi_xml_ubl_21_zatca.py
```diff
@@ -420,15 +421,18 @@ def grouping_key_generator(base_line, tax_values):
         extension_amount = abs(line_vals['line_extension_amount'])
         if not line.move_id._is_downpayment() and line._get_downpayment_lines():
             total_amount_sa = extension_amount = 0
+            line_selection = True
```
khva-odoo: We should rename the variable to something more meaningful so the purpose of hte variable can be clearer.

PR #247731, opened 2026-02-09
[FIX] account: Add Filter to support selecting multiple journals

leclerc-leo:
In accounting reports, we want to be able to filter on multiples journals if needed. This filter will allow us to do this.

Using `context['search_default_journal_id'] = journal_ids` doesn't work as only the first id is passed to the filter.

---


oco-odoo (approved):
@robodoo r+

PR #247738, opened 2026-02-09
[IMP] web, portal, auth_*: improve portal security page and components

stefanorigano:
auth_*: auth_passkey_portal, auth_totp, auth_totp_portal.

#### This PR main goal is to provide an overall visual enhancement for the portal security page.

At the framework level (`web`), the PR addresses a problem caused by the excessive size of dialogs headings on the frontend.
This issue is due to the `fs-4` class, which provides a correct font size in the backend but not in the portal/website where the default font size is sensibly larger.
The solution introduces a new `titleClasses` prop for the `Dialog` component, which defaults to `fs-4` but allows for the use of different classes when necessary.

In portal specifically, the PR improves a functional but somewhat "cluttered" security page into a cleaner interface that's easier to understand and use, especially for non-technical users.

task-5925007

Documentation PR:
- https://github.com/odoo/documentation/pull/16344

----

**Introduced changes:**

- **Meaningful dialog titles**: eg, from "Passkeys" to "Rename Passkey"  
- **Added descriptive subtitles** explaining what each section does
- **Responsive tables** for API keys and Passkeys 
- **Improve tables actions clarity** using text links rather than icons  ("Rename" / "Delete" instead of pencil/trash icons)
- **Hide tables when they have no entries**
-  **Contextual colors**: ensure proper use of warning and danger classes regardless by theme customizations (see "Two-factor authentication not enabled" alert) 
- **Responsive form** grid layout
- **Streamline 2FA modal**:  Clarify that's a 3 steps procedure using bullets. Don't hide the manual code inside a collapse -> not necessary, increases friction for users struggling with the flow.  
- **Add confirmation dialogs** for irreversible and potentially dangerous actions, eg. key removal or forced log-out

**Accessibility:**
- Icons should use `aria-hidden="true"` to be ignored by screen-readers 
- Headings should match the page hierarchy  
- Use the `<time>` tag and `datatime` attributes for time related data
- Added meaningful `title` attributes: eg. in the "Edit Password" form, the "Discard" button now uses a "_Keep my current password_" title
- Form labels properly associated with inputs via for/id attributes
- Make use of `role` and `aria` attributes

**Varius:**
- Removed unnecessary SCSS file

## Examples

| Header |
|--------|
| image|
|image|

### Create Passkey
| master | this PR |
|--------|--------|
| image | image | 

### Create API Key

| master | this PR |
|--------|--------|
| image | image | 

### 2FA 
| master | this PR |
|--------|--------|
| image |  image | 

### NEW Confirmation Dialogs

| Disable 2FA | Delete Passkey | Revoke Sessions |
|--------|--------|--------|
| image | image | image | 

---


addons/portal/static/src/interactions/portal_security.js
```diff
@@ -67,6 +67,7 @@ export class PortalSecurity extends Interaction {
 
         this.services.dialog.add(InputConfirmationDialog, {
             title: _t("New API Key"),
+            titleClasses: "fs-4",
```
cth-odoo: Hello @stefanorigano 

Is it really necessary to introduce a new prop just to format the title? Why not always using `fs-4` since `fs-3` is too big in some cases?

CC: @aab-odoo

stefanorigano: Hi 👋

> is it really necessary to introduce a new prop just to format the title?

Yes. `fs-3` is perfect fro the backedn,  while `fs-4` would be a bit too small.
`fs-3` is  too large for the frontend only, that's the reason of the introduction of the new prop.

Many classes can be used in both environments, but font sizes are a bit tricky because they depend on the environment. On their own, `fs-4` doesn’t mean anything; it’s the environment that ultimately determines how big `fs-4` will be.

A “one size fits all” approach isn’t really possible here because, even if we find a compromise value that works in both environments, we might redesign the website (or the backend) tomorrow, which would invalidate this logic.

stefanorigano: @aab-odoo , could I have a feedback on https://github.com/odoo/odoo/pull/247738/changes/dd76fb829109df1b13b8c9f54e6eed6ba4080035 ?

The alternative would be a CSS override. Thanks!

aab-odoo: I would also prefer avoiding this new props on Dialog, because every new props slightly increases the complexity, but mostly because I don't think we'd want people to start customizing the title of their dialogs. In your case, I assume it's fine because you want a different font-size in the backend and in the frontend. But inside a given application (e.g. the backend), you want consistency, so giving the ability to everyone to easily customize the title is probably not a good idea. I would then go with the css solution if it's easy.

stefanorigano: https://github.com/odoo/odoo/pull/247738/changes/116750ef373776d29ae8dbbb8d1df271b1df3835 adapted, thanks 👍

PR #228358, opened 2025-09-24
 [IMP] loyalty: prevent setting expiration date in the past

njor-odoo:
Before this commit, it was possible to create or update loyalty cards
with an expiration date in the past, potentially leading to invalid or
unusable coupons.

After this commit, a ValidationError is raised when attempting to set
expiration_date to a past date, ensuring data integrity for loyalty cards.

task-5075177

PR #247740, opened 2026-02-09
[FIX] sale_loyalty: prevent claiming expired ewallet points

mma1377:
### Issue:
Due to this issue, ewallet points are claimable after expiry.

#### Steps to reproduce:
1- Create a ewallet program, and generate an ewallet for a partner.
2- Set the expiration date in the past.
3- Create a SO with the same partner.
4- Click on reward.

Expected: The ewallet shouldn't be claimable.
Current outcome: The ewallet is claimable.

opw-5476686


ltinel (approved):
Thanks for the fix :)
@robodoo delegate+

ltinel:
> > Actually, it would be nice to also prevent selecting an expiration date in the past. WDYT?
> 
> There is even already a PR for that, but for some reason it's still not merged.
> 
> https://github.com/odoo/odoo/pull/228358/changes/42b2c8e3fc1d485058cec4f776a328a49916677a..123c7056e95a87fdd9d67ad4bde74b98d2c07543

I reviewed it but kinda forgot about it :sweat_smile: It could be backported, but it's not a "real" bugfix so probably not worth it

PR #230334, opened 2025-10-07
[IMP] point_of_sale: refund flow UX

macocianradu:
- Added a buttonClass property on the action pad which is applied on the action button
- Disable the Refund button when there is no refund to process
- Added a check on hasItemsToRefund to check whether any of the lines to be refunded is already being processed in another opened order to avoid being able to refund the same item multiple times. Before it was possible to create a refund order, keep the payment page open, go back to the refund list, and create another refund for the same item. Then if you closed both orders, they would both appear as refunds and you could refund more than what was sold
- Removed the refund button from the action menu.
- Removed 'refunding in other order' labels from the order line display
- Moved the 'to refund' label on the same line as the orderline before the total
- Changed the clickRefund() test function to use the Orders menu with
  the 'paid' filter instead of the now removed control button

Task-[5145135](https://www.odoo.com/odoo/project/1737/tasks/5145135)
Enterprise PR-[#101657](https://github.com/odoo/enterprise/pull/101657)

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #247741, opened 2026-02-09
[IMP] point_of_sale, pos_*: revamp control buttons in Actions modal

edi-odoo:
In this PR we've redesigned the control buttons inside the Actions modal for more clarity.

|Before|After|
|---|---|
| Screenshot 2026-02-09 at 10 08 24 | Screenshot 2026-02-09 at 10 07 36 |

task-5187160

---

### DIFF
--- a/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
+++ b/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
@@ -1,11 +1,75 @@
-.control-buttons-modal {
-    grid-template-columns: repeat(3, 1fr);
-
+.modal-dialog:has(.control-button:nth-child(10)){
     @include media-breakpoint-down(lg) {
-        grid-template-columns: repeat(2, 1fr);
+        @media (orientation: landscape) {
+            --modal-width: 80vw;
+        }
+    }
+}
+
+.control-buttons-modal {
+    grid-template-columns: repeat(var(--grid-columns, 1), 1fr);
+    
+    .control-button {
+        aspect-ratio: var(--controlButton-aspectRatio, auto);
+        
+        &:last-child {
+            --controlButton-aspectRatio: auto;
+        }
+
+        .o_guests_number {
+            
+            @include media-breakpoint-down(md) {
+                font-size: map-get($font-sizes, 6);
+            }
+        }
     }
 
-    @include media-breakpoint-down(sm) {
-        grid-template-columns: 1fr;
+    // Handle button layout and size based on the number of buttons and screen orientation
+    @include media-breakpoint-up(md) {
+        --controlButton-aspectRatio: 1.1;
+        
+        @media (orientation: portrait) {
+            --grid-columns: 3;
+        }
+
+        @media (orientation: landscape) {
+            --grid-columns: 4;  
+
+            @media (max-height: 940px) {
+                --controlButton-aspectRatio: 1.5;
+            }
+
+            &:has(.control-button:nth-child(10)):not(:has(.control-button:nth-child(11))){
+                --controlButton-aspectRatio: 1.5;
+            }
+        }
+        
+        &:has(.control-button:nth-child(9)):not(:has(.control-button:nth-child(11))) {
+                --grid-columns: 3;
+                
+                @media (max-height: 940px) {
+                    --controlButton-aspectRatio: 1.5;
+                }
+                
+                @media (max-height: 600px) {
+                    --controlButton-aspectRatio: 2;
+                }
+        }
+        
+        // When there are 10 or 13 buttons, the last button is alone on the last row (and should always be the Cancel button)
+        // so we make it take up the whole width of the grid, adapt the aspect ratio and place the icon next to the text
+        &:has(.control-button:nth-child(10)):not(:has(.control-button:nth-child(11))), 
+        &:has(.control-button:nth-child(13)):not(:has(.control-button:nth-child(14))) {
+            .control-button:last-child {
+                --controlButton-aspectRatio: auto;
+                --oi-font-size: 1.5rem;
+                
+                grid-column: 1 / -1;
+                flex-direction: row !important;
+                align-items: center !important;
+                gap: map-get($spacers, 2);
+                height: 5rem;
+            }
+        }
     }
 }
--- a/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
+++ b/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
@@ -1,63 +1,104 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
+    <t t-name="point_of_sale.controlButton">
+        <button 
+        t-attf-class="control-button btn btn-lg {{btnClasses ? btnClasses : ''}} d-inline-flex flex-md-column justify-content-md-center gap-1 overflow-hidden"
+        t-on-click="onClick"
+        t-att-disabled="disabled ? 'disabled' : ''">
+            <i t-if="iconClasses" t-attf-class="{{iconClasses}} {{!this.ui.isSmall ? ' fa-2x fa-fw' : ''}} align-self-center opacity-50" role="img" aria-hidden="true"/>
+            <t t-if="badge" t-out="badge"/>
+            <span t-attf-class="d-flex flex-md-column flex-grow-1 flex-md-grow-0 align-items-center align-items-md-stretch min-w-0 {{this.ui.isSmall ? ' lh-sm' : ''}} text-start text-md-center">
+                <span class="px-1" t-esc="btnText"/>
+                <t t-if="subText" t-out="subText"/>
+            </span>
+        </button>
+    </t>
     <t t-name="point_of_sale.ControlButtons">
         <!-- All buttons always displayed -->
-        <SelectPartnerButton partner="this.partner" t-if="!this.props.showRemainingButtons"/>
-        <t t-if="!this.props.showRemainingButtons || (this.ui.isSmall and this.props.showRemainingButtons)">
-            <InternalNoteButton label.translate="Note" class="this.buttonClass"/>
+        <t t-if="!this.props.showRemainingButtons">
+            <SelectPartnerButton partner="this.partner"/>
+            <InternalNoteButton label.translate="Note" class="'btn-secondary '"/>
+            <button 
+                t-if="this.pos.config.use_presets"
+                class="btn btn-secondary btn-lg flex-shrink-0 border-0 lh-lg text-truncate"
+                t-attf-class="{{this.currentOrder.preset_id?.color and `o_colorlist_item_color_${this.currentOrder.preset_id?.color}`}}"
+                t-on-click="() => this.pos.selectPreset()">
+                <span t-if="this.currentOrder.preset_id?.name" t-out="this.currentOrder.preset_id?.name" />
+                <span t-else="">Preset</span>
+            </button>
+            <button t-if="!this.ui.isSmall and this.pos.showSaveOrderButton" t-att-disabled="this.pos.getOrder().isEmpty()" t-attf-class="{{ buttonClass }} ms-auto" t-on-click="() => this.pos.clickSaveOrder()">
+                <i class="fa fa-upload" aria-hidden="true"/>
+            </button>
+            <button t-if="!this.ui.isSmall and this.props.onClickMore" class="more-btn btn btn-secondary btn-lg flex-shrink-0 ms-auto lh-lg" t-on-click="this.props.onClickMore">
+                <i class="oi oi-fw oi-ellipsis-v" aria-hidden="true" />
+            </button>
         </t>
-        <button t-if="this.pos.config.use_presets and !this.props.showRemainingButtons"
-            class="btn btn-secondary btn-lg border-0 text-truncate"
-            t-attf-class="{{this.currentOrder.preset_id?.color and `o_colorlist_item_color_${this.currentOrder.preset_id?.color}`}}"
-            t-on-click="() => this.pos.selectPreset()">
-            <span t-if="this.currentOrder.preset_id?.name" t-out="this.currentOrder.preset_id?.name" />
-            <span t-else="">Preset</span>
-        </button>
-        <button t-if="!this.props.showRemainingButtons and !this.ui.isSmall and this.pos.showSaveOrderButton" t-att-disabled="this.pos.getOrder().isEmpty()" t-attf-class="{{ this.buttonClass }} ms-auto" t-on-click="() => this.pos.clickSaveOrder()">
-            <i class="fa fa-upload"/>
-        </button>
-        <button class="btn btn-secondary btn-lg flex-shrink-0 more-btn" t-att-class="{'ms-auto': !this.pos.showSaveOrderButton}" t-if="!this.props.showRemainingButtons and !this.ui.isSmall and this.props.onClickMore" t-on-click="this.props.onClickMore">
-            <i class="oi oi-fw oi-ellipsis-v" aria-hidden="true" />
-        </button>
         <!-- All these buttons will only be displayed in a dialog -->
         <t t-if="this.props.showRemainingButtons">
-            <div class="control-buttons control-buttons-modal d-grid gap-2 mt-2">
+            <div class="control-buttons control-buttons-modal d-grid grid-auto-flow-columns gap-2">
+                <t t-if="(this.ui.isSmall and this.props.showRemainingButtons)">
+                    <InternalNoteButton label.translate="Note" class="'btn-secondary '"/>
+                </t>
                 <NoteButton
                     label.translate="Customer Note"
-                    class="this.buttonClass"
+                    class="'btn-secondary '"
                     icon="'fa fa-sticky-note'"/>
-                <button t-if="this.pos.config.iface_printbill" t-att-class="this.buttonClass"
-                    t-att-disabled="!this.pos.getOrder()?.getOrderlines()?.length"
-                    t-on-click="this.clickPrintBill">
-                    <i class="fa fa-print me-1"></i>Bill
-                </button>
-                <button t-if="this.pos.cashier._role != 'minimal'" class="o_pricelist_button" t-att-class="this.buttonClass" t-on-click="() => this.clickPricelist()">
-                    <i class="fa fa-th-list me-2" role="img" aria-label="Price list" title="Price list" />
-                    <t t-if="this.currentOrder?.pricelist_id" t-out="this.currentOrder.pricelist_id.display_name" />
-                    <t t-else="">Pricelist</t>
-                </button>
-                <button t-if="this.pos.config.fiscal_position_ids.length and this.pos.cashier._role != 'minimal'"
-                    class="control-button o_fiscal_position_button"
-                    t-att-class="this.buttonClass"
-                    t-on-click="() => this.clickFiscalPosition()">
-                    <i class="fa fa-book me-1" role="img" aria-label="Set fiscal position"
-                    title="Set fiscal position" />
-                    <t t-if="this.currentOrder?.fiscal_position_id" t-out="this.currentOrder.fiscal_position_id.display_name" />
-                    <t t-else="">Tax</t>
-                </button>
-                <button t-if="this.displayProductInfoBtn()" t-att-class="this.buttonClass" t-on-click="() => this.pos.onProductInfoClick(this.currentOrder.getSelectedOrderline().product_id.product_tmpl_id, this.currentOrder.getSelectedOrderline().product_id)">
-                    <i class="fa fa-info-circle me-1" role="img" aria-label="Product Info" title="Product Info" /> Info
-                </button>
-                <button t-if="this.pos.isSelectedLineCombo" class="break-combo-button" t-att-class="this.buttonClass" t-on-click="() => this.breakSelectedCombo()">
-                    <i class="fa fa-outdent" role="img" aria-label="Break combo" title="Brean combo" /> Break combo
-                </button>
-                <button
-                    t-if="this.pos.cashier._role != 'minimal'"
-                    t-att-class="this.buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())"
-                    t-att-disabled="!this.pos.getOrder()?.getOrderlines()?.length"
-                >
-                    <i class="fa fa-trash me-1" role="img" /> Cancel Order
-                </button>
+                <!-- Bill Button -->
+                <t t-if="this.pos.config.iface_printbill"> 
+                    <t t-call="point_of_sale.controlButton">
+                        <t t-set="iconClasses" t-value="'fa fa-print'"/>
+                        <t t-set="btnText">Bill</t>
+                        <t t-set="btnClasses" t-value="'btn-secondary '"/>
+                        <t t-set="disabled" t-value="!this.pos.getOrder()?.getOrderlines()?.length"/>
+                        <t t-set="onClick" t-value="this.clickPrintBill"/>
+                    </t>
+                </t>
+                <!-- Pricelist Button -->
+                <t t-if="this.pos.cashier._role != 'minimal'" t-call="point_of_sale.controlButton" name="o_pricelist_button">
+                    <t t-set="iconClasses" t-value="'fa fa-th-list'"/>
+                    <t t-set="btnText">Pricelist</t>
+                    <t t-set="btnClasses" t-value="'o_pricelist_button btn-secondary '"/>
+                    <t t-set="onClick" t-value="() => this.clickPricelist()"/>
+                    <t t-set="subText">
+                        <t t-if="this.currentOrder?.pricelist_id">
+                            <small t-out="this.currentOrder.pricelist_id.display_name" class="ms-auto ms-md-0 text-truncate text-muted text-end text-md-center"/>  
+                        </t>
+                    </t>
+                </t>
+                <!-- Tax Button -->
+                <t t-if="this.pos.config.fiscal_position_ids.length and this.pos.cashier._role != 'minimal'" t-call="point_of_sale.controlButton">
+                    <t t-set="iconClasses" t-value="'fa fa-book'"/>
+                    <t t-set="btnText">Tax</t>
+                    <t t-set="btnClasses" t-value="'o_fiscal_position_button btn-secondary '"/>
+                    <t t-set="onClick" t-value="() => this.clickFiscalPosition()"/>
+                    <t t-set="subText">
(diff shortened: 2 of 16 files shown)


addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.js
```diff
@@ -11,6 +11,7 @@ export class NoteButton extends Component {
         icon: { type: String, optional: true },
         label: { type: String, optional: false },
         class: { type: String, optional: true },
+        inDialog: { type: Boolean, optional: true },
```
stefanorigano: `inDialog` wording: I think we can improve the prop name since it's not really representative of what it does.
By using this name it somehow implies that we should always set to `true` when the button is inside a modal BUT, for example,  we may want buttons to be styled normally for other dialogs... OR we may want to achieve this design in a context that's not a a dialog one (eg. a future sidebar)

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
```diff
@@ -5,9 +5,11 @@
             t-attf-class="{{this.props.class}} flex-shrink-0"
             t-att-disabled="pos.getOrder()?.isEmpty()" t-on-click="onClick">
             <t t-if="this.props.icon">
-                <i t-attf-class="{{this.props.icon}} me-1"/>
+                <span class="d-lg-flex flex-column justify-content-end h-lg-50">
+                    <i t-attf-class="{{this.props.icon}} {{this.env.isSmall ? 'me-1' : 'fs-1'}} text-500"/>
```
stefanorigano: Missing `aria-hidden="true"` attribute. We don't want screenreaders to "read" icons ;)

addons/pos_restaurant/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -10,30 +10,50 @@
...
-                    <i class="oi oi-arrow-down me-1" />Transfer course
-                </button>
-                <button t-if="!pos.getOrder()?.table_id" t-att-class="buttonClass" t-on-click="() => this.pos.editFloatingOrderName(this.pos.getOrder())">
-                    <i class="fa fa-pencil-square-o me-1" />Edit Order Name
-                </button>
+                <t t-call="point_of_sale.controlButton">
+                    <t t-set="notIcon">
+                        <span t-esc="currentOrder?.getCustomerCount() || 0" t-attf-class="o_guests_number rounded-5 text-bg-500 d-lg-block m-lg-auto my-lg-1 {{ui.isSmall ? 'px-1 me-1 small' : ''}}"/>
+                    </t>
+                    <t t-set="btnText" t-value="'Guest' + (currentOrder?.getCustomerCount() > 1 ? 's' : '')"/>
```
stefanorigano: Two issues:

1: Strings **concatenations** are not translations friendly.
Eg, in Italian the singular form of "Guest" is "Ospit**e**", and the plural form is "Ospit**i**"... by using your implementation we'll get "Ospite**s**".

2: Strings inside `t-value` are not translatable.



```suggestion
                    <t t-set="btnText">Guest</t>
                    <t t-set="btnText" t-if="'currentOrder?.getCustomerCount() > 1">Guests</t>
```

addons/pos_loyalty/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -2,46 +2,44 @@
...
+                <t t-set="disabled" t-value="!pos.getOrder().isProgramsResettable()"/>
+                <t t-set="onClick" t-value="() => this.pos.resetPrograms()"/>
             </t>
         </xpath>
         <xpath expr="//button[hasclass('more-btn')]" position="attributes">
             <attribute name="t-attf-class">{{ state.nbrRewards ? 'active text-action' : '' }}</attribute>
         </xpath>
     </t>
-</templates>
+</templates>
```
stefanorigano: Missing new line at the end of the file.

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,10 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
+    <t t-name="point_of_sale.controlButton">
+        <button t-attf-class="control-button btn {{btnClasses}} {{ui.isSmall ? 'btn-md text-start' : 'btn-lg'}}"
+                t-on-click="onClick"
+                t-att-disabled="disabled ? 'disabled' : ''">
+            <span class="d-lg-flex flex-column justify-content-end h-lg-50">
+            <t t-if="notIcon" t-out="notIcon"/>
+                <t t-else="">
+                    <i t-if="iconClasses" t-attf-class="{{iconClasses}} {{ui.isSmall ? 'me-1' : 'fs-1'}} {{btnText == 'Cancel Order' ? '' : 'text-500' }}" role="img" t-att-aria-label="iconTitle ? iconTitle : btnText" t-att-title="iconTitle ? iconTitle : btnText"/>
```
stefanorigano: Icons should use `aria-hidden="true"` and being ignored by screenreaders. 

`aria-label` can be removed: The element that may need an `aria-label` is the `<button>` itself, but only when we need to override the accessible name that's normally retrieved from the text content (eg. when we have an icon-only button and we need to enforce a label ), but it's not the case here since the text is readable.

`title` can be removed as well. We may set it to the button if we need to say "something more" that what the button normally says, eg "Click to apply a discount"... but it's not the case here neither ;)

addons/pos_restaurant/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -10,30 +10,50 @@
...
+                    <t t-set="btnClasses" t-value="'o_guests_button ' + buttonClass"/>
+                    <t t-set="onClick" t-value="clickTableGuests"/>
+                </t>
+                <!-- Split Button -->
+                <t t-call="point_of_sale.controlButton">
+                    <t t-set="iconClasses" t-value="'fa fa-files-o'"/>
+                    <t t-set="btnText" t-value="'Split'"/>
```
stefanorigano: Here again, not translatable. Please, check all other occurrences 

```suggestion
                    <t t-set="btnText">Split</t>
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,10 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
+    <t t-name="point_of_sale.controlButton">
+        <button t-attf-class="control-button btn {{btnClasses}} {{ui.isSmall ? 'btn-md text-start' : 'btn-lg'}}"
+                t-on-click="onClick"
+                t-att-disabled="disabled ? 'disabled' : ''">
+            <span class="d-lg-flex flex-column justify-content-end h-lg-50">
+            <t t-if="notIcon" t-out="notIcon"/>
```
stefanorigano: Oops, wrong indentation ;)

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,10 +1,26 @@
...
+                <t t-else="">
```
stefanorigano: I’m not a big fan of this approach. The slot name `notIcon` isn’t very clear, and the correlation with the following icon block is also unclear (for example, why can’t I have both?).

I propose to have one single slot and use to to render whatever we need. eg:

```xml
<t t-name="point_of_sale.controlButton">
    <t t-if="pictogram" t-out="pictogram"/> 
    [...]
</t>

<t t-call="point_of_sale.controlButton">
    <t t-set="pictogram">
         <i class="fa fa-whatever">
    </t>
    [...]
</t>

<t t-call="point_of_sale.controlButton">
    <t t-set="pictogram">
         <span class="badge>20</span>
    </t>
    [...]
</t>
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.js
```diff
@@ -129,8 +129,8 @@ export class ControlButtons extends Component {
     get buttonClass() {
         return this.props.showRemainingButtons
             ? this.ui.isSmall
-                ? "btn bg-100 btn-md py-2 text-start"
-                : "btn btn-secondary btn-lg py-5"
+                ? "btn bg-100 btn-md text-start"
```
stefanorigano: Why enforcing `bg-100` rather than using `btn-secondary`, like on desktop? 
Also, we can keep `btn-lg`on mobile as well, it makes perfectly sense for these buttons on mobile, they must be easy to "tap".
I'd remove disabled button **on mobile**... @mobt-odoo are you ok with these changes? 

| this PR | proposal |
|--------| --------|
| Image |  Image |



note: I think that `btn-md` doesn't exist and should be removed...  I know that it was already there, but since you're touching the line....

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
```diff
@@ -1,11 +1,21 @@
...
+    @include media-breakpoint-up(sm) {
+        --grid-columns: 2;
     }
 
-    @include media-breakpoint-down(sm) {
-        grid-template-columns: 1fr;
+    @include media-breakpoint-up(lg) {
+        --grid-columns: 4;
+
+        .btn-lg {
```
stefanorigano: Let's target the `.control-button` class instead.
We should avoid targeting framework base components classes for custom CSS

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
```diff
@@ -1,11 +1,21 @@
...
+            min-width: 200px;
+            min-height: 195px;
```
stefanorigano: Can we replace these rules with `aspect-ratio: 1.5` or smthg?
Not an huge fan of arbitrary rules 😅

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -18,43 +34,74 @@
...
+                    <!-- Bill Button -->
+                <t t-if="pos.config.iface_printbill" t-call="point_of_sale.controlButton">
+                    <t t-set="iconClasses" t-value="'fa fa-print d-lg-block text-500'"/>
+                    <t t-set="btnText" t-value="'Bill'"/>
+                    <t t-set="btnClasses" t-value="buttonClass"/>
+                    <t t-set="disabled" t-value="!pos.getOrder()?.getOrderlines()?.length"/>
+                    <t t-set="onClick" t-value="clickPrintBill"/>
+                </t>
+                <!-- Pricelist Button -->
+                <t t-if="this.pos.cashier._role != 'minimal'" t-call="point_of_sale.controlButton" class="o_pricelist_button">
```
stefanorigano: `class="o_pricelist_button"` => I don't think that's doing something 😉

stefanorigano (changes requested):
Hello @edi-odoo, well done, thanks :)
Some remarks for you in the code

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
edi-odoo: I'd keep it because it's used for xpaths, unless you have another suggestion? 
Also used in tests.

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.js
edi-odoo: I just noticed `this.env.inDialog` already exists, so I can use that instead :)

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
stefanorigano: Are you referring to this [xpath](https://github.com/odoo/odoo/pull/247741/changes#diff-e8e3e0fedfd9fdb8b4fe593b269d7faaab18066b414a9afa60ba5fe2eec8621cL5)?
```diff
- expr="//t[@t-if='props.showRemainingButtons']/div/button[hasclass('o_pricelist_button')]"
+ expr="//t[@t-if='props.showRemainingButtons']/div/t[hasclass('o_pricelist_button')]"
```

Before it was targeting the button, that's element that can have a `class` attribute indeed.
Using class on a `<t>` is misleading instead because it will never be used as a "DOM class" strictly speaking. You could use the `name` attribute instead and adapt the xpath accordingly.

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.js
stefanorigano: Are we sure that's always the design we want to trigger each time the button is rendered inside a modal ?

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.js
edi-odoo: Didn't apply this because `isSmall` is triggered at the MD breakpoint. Buttons are still "large/squared" at the MD breakpoint and only become listed above at the SM breakpoint. 
If we disable the buttons on `isSmall`, the user using their tablet in landscape and turning it to portrait would see buttons disappear before their eyes.

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
```diff
@@ -1,11 +1,68 @@
...
+        
+        &:has(.control-button:nth-child(9)):not(:has(.control-button:nth-child(11))) {
+                --grid-columns: 3;
+                
+                @media (max-height: 940px) {
+                    --btn-aspect-ratio: 1.5;
+                }
+        }
+
+        &:has(.control-button:nth-child(10)):not(:has(.control-button:nth-child(11))), &:has(.control-button:nth-child(13)):not(:has(.control-button:nth-child(14))) {
```
stefanorigano: Nitpicking: more readable. Also add a comment explaining what the code does and why is needed.


```suggestion
        &:has(.control-button:nth-child(10)):not(:has(.control-button:nth-child(11))),
        &:has(.control-button:nth-child(13)):not(:has(.control-button:nth-child(14))) {
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
```diff
@@ -2,12 +2,14 @@
 <templates id="template" xml:space="preserve">
     <t t-name="point_of_sale.NoteButton">
         <button
-            t-attf-class="{{this.props.class}} flex-shrink-0"
+            t-attf-class="control-button btn btn-lg {{this.props.class}} flex-shrink-0 text-start text-md-center lh-lg"
             t-att-disabled="pos.getOrder()?.isEmpty()" t-on-click="onClick">
             <t t-if="this.props.icon">
-                <i t-attf-class="{{this.props.icon}} me-1"/>
+                <span t-attf-class="d-md-flex flex-md-column justify-content-end {{this.props.icon ? 'h-md-50' : ''}}">
+                    <i t-attf-class="{{this.props.icon ? this.props.icon + ' me-2 me-md-0' : ''}} text-500" aria-hidden="true"/>
```
stefanorigano: I propose to use opacity rather than an enforced text color. 

#### `text-500`

---

Image

Image

Image

#### `opacity-75`

---

Image

Image

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
```diff
@@ -2,12 +2,14 @@
...
+                </span>
             </t>
-            <span t-esc="this.props.label"/>
+            <span t-attf-class="{{this.props.icon ? 'd-md-flex flex-md-column align-items-center h-md-50 pt-md-2 lh-sm' : ''}}" t-esc="this.props.label"/>
```
stefanorigano: `h-md-50`: why? looks unnecessary.
could be simplified by using `d-inline-flex flex-column justify-content-center`

Image

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.scss
```diff
@@ -1,11 +1,68 @@
 .control-buttons-modal {
-    grid-template-columns: repeat(3, 1fr);
+    grid-template-columns: repeat(var(--grid-columns, 1), 1fr);
+    
+    .control-button {
+        aspect-ratio: var(--btn-aspect-ratio, auto);
```
stefanorigano: CSS variable naming: let's follow our [conventions](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#css-variables). 

I propose `--controlButton-aspectRatio`.

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
+            <InternalNoteButton label.translate="Note" class="'btn-secondary '"/>
+            <button t-if="pos.config.use_presets"
+                class="btn btn-secondary btn-lg flex-shrink-0 border-0 lh-lg"
+                t-attf-class="{{currentOrder.preset_id?.color and `o_colorlist_item_color_${currentOrder.preset_id?.color}`}}"
+                t-on-click="() => this.pos.selectPreset()">
+                <span t-if="currentOrder.preset_id?.name" t-esc="currentOrder.preset_id?.name" />
+                <span t-else="">Preset</span>
+            </button>
+            <button t-if="!props.showRemainingButtons and !ui.isSmall and this.pos.showSaveOrderButton" t-att-disabled="this.pos.getOrder().isEmpty()" t-attf-class="{{ buttonClass }} ms-auto" t-on-click="() => this.pos.clickSaveOrder()">
+                <i class="fa fa-upload"/>
```
stefanorigano: Nit
```suggestion
                <i class="fa fa-upload" aria-hidden="true"/>
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
-        <SelectPartnerButton partner="partner" t-if="!props.showRemainingButtons"/>
-        <t t-if="!props.showRemainingButtons || (ui.isSmall and props.showRemainingButtons)">
-            <InternalNoteButton label.translate="Note" class="buttonClass"/>
+        <t t-if="!props.showRemainingButtons">
+            <SelectPartnerButton partner="partner"/>
+            <InternalNoteButton label.translate="Note" class="'btn-secondary '"/>
+            <button t-if="pos.config.use_presets"
+                class="btn btn-secondary btn-lg flex-shrink-0 border-0 lh-lg"
+                t-attf-class="{{currentOrder.preset_id?.color and `o_colorlist_item_color_${currentOrder.preset_id?.color}`}}"
+                t-on-click="() => this.pos.selectPreset()">
```
stefanorigano: Nit indentation: either one line or indent each attribute.

```suggestion
            <button 
                t-if="pos.config.use_presets"
                class="btn btn-secondary btn-lg flex-shrink-0 border-0 lh-lg"
                t-attf-class="{{currentOrder.preset_id?.color and `o_colorlist_item_color_${currentOrder.preset_id?.color}`}}"
                t-on-click="() => this.pos.selectPreset()"
              >
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
+                <span t-if="currentOrder.preset_id?.name" t-esc="currentOrder.preset_id?.name" />
+                <span t-else="">Preset</span>
+            </button>
+            <button t-if="!props.showRemainingButtons and !ui.isSmall and this.pos.showSaveOrderButton" t-att-disabled="this.pos.getOrder().isEmpty()" t-attf-class="{{ buttonClass }} ms-auto" t-on-click="() => this.pos.clickSaveOrder()">
+                <i class="fa fa-upload"/>
+            </button>
+            <button class="more-btn btn btn-secondary btn-lg flex-shrink-0 ms-auto lh-lg" t-if="!ui.isSmall and props.onClickMore" t-on-click="props.onClickMore">
```
stefanorigano: Convention: `t-if` attribute before everything else

```suggestion
            <button t-if="!ui.isSmall and props.onClickMore" class="more-btn btn btn-secondary btn-lg flex-shrink-0 ms-auto lh-lg" t-on-click="props.onClickMore">
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
-                </button>
-                <button
-                    t-if="this.pos.cashier._role != 'minimal'"
-                    t-att-class="buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())"
-                    t-att-disabled="!pos.getOrder()?.getOrderlines()?.length"
-                >
-                    <i class="fa fa-trash me-1" role="img" /> Cancel Order
-                </button>
+                    <!-- Bill Button -->
+                <t t-if="pos.config.iface_printbill">
```
stefanorigano: Nit
```suggestion
                <!-- Bill Button -->
                <t t-if="pos.config.iface_printbill"> 
```

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
-                <button
-                    t-if="this.pos.cashier._role != 'minimal'"
-                    t-att-class="buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())"
-                    t-att-disabled="!pos.getOrder()?.getOrderlines()?.length"
-                >
-                    <i class="fa fa-trash me-1" role="img" /> Cancel Order
-                </button>
+                    <!-- Bill Button -->
+                <t t-if="pos.config.iface_printbill"> 
+                    <t t-set="btnCount" t-value="btnCount + 1"/>
```
stefanorigano: `btnCount` seems not used... leftover?

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,116 @@
...
+                    <t t-set="onClick" t-value="() => this.pos.onProductInfoClick(currentOrder.getSelectedOrderline().product_id.product_tmpl_id, currentOrder.getSelectedOrderline().product_id)"/>
+                </t>
+                <!-- Break Combo Button -->
+                <t t-if="this.pos.isSelectedLineCombo" t-call="point_of_sale.controlButton">
+                    <t t-set="iconClasses" t-value="'fa fa-outdent'"/>
+                    <t t-set="btnText">Break Combo</t>
+                    <t t-set="btnClasses" t-value="'break-combo-button btn-secondary '"/>
+                    <t t-set="onClick" t-value="() => this.breakSelectedCombo()"/>
+                </t>
+                <t t-call="point_of_sale.controlButton">
```
stefanorigano: Nit: missing comment

```suggestion
                <!-- Cancel Order Button -->
                <t t-call="point_of_sale.controlButton">
```

stefanorigano (changes requested):
Very good @edi-odoo ,
Some minor remark on my side, for the rest just nitpicking comments.
Well done  👍 

---
Sidenote: I can get the following result on small tablets, could you have a look? 

Image

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
edi-odoo: This was done so that all the icons and texts are aligned with the neighbouring buttons' icons and texts

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
edi-odoo: Not sure how you got that result with `opacity-75` 😅

Image

Image

I think you meant `opacity-50` (I've changed it to that)

Image
Image

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/orderline_note_button/orderline_note_button.xml
stefanorigano: @edi-odoo , yep, but it doesn't always work. I'd suggest the flex approach
image

edi-odoo: Just noticed this button didn't have the `h-50` on it, so indeed it didn't work ;) 
It was only on the "Note/Customer Note" buttons (leftover).

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -1,63 +1,111 @@
...
+                        </t>
+                    </t>
+                </t>
+                <!-- Refund Button -->
+                <t t-if="this.pos.cashier._role != 'minimal'" t-call="point_of_sale.controlButton">
+                    <t t-set="iconClasses" t-value="'fa fa-undo'"/>
+                    <t t-set="btnText">Refund</t>
+                    <t t-set="btnClasses" t-value="'btn-secondary '"/>
+                    <t t-set="onClick" t-value="() => this.clickRefund()"/>
+                </t>
```
stva-odoo: Hello @edi-odoo, 
The Refund button  has been removed by this PR: https://github.com/odoo/odoo/pull/230334.
Could you remove it as well?

stva-odoo:
robodoo r+

PR #247743, opened 2026-02-09
[FIX] sale_{loyalty, timesheet}: apply loyalty reward with non-unit uom

mma1377:
### Issue:
Due to this issue, applying a different uom than unit one, will not apply the reward regarding to quantities.

#### Steps to reproduce:
1- Create a program: buy 12 get 6 free.
2- Create a SO, add a dozen of product to SOL.
3- Click on reward.

Expect: 6 free unit is added.
Current outcome: Nothing is added. Unless you add 12 dozens which is going to add 6 units.

### Cause:
In checking rules, `product_uom_qty` is directly used without conversion to quantity.

Note:
`test_different_uom_to_hours_on_sale_order_confirmation` is failing due to this fix, because the uom_id unit/dozens and hours/days are not compatible. As this is not possible in UI, IMO we can delete the breaking SOL in that test.

opw-5913638

### DIFF
--- a/addons/sale_loyalty/models/sale_order.py
+++ b/addons/sale_loyalty/models/sale_order.py
@@ -1168,7 +1168,10 @@ class SaleOrder(models.Model):
         products = order_lines.product_id
         products_qties = dict.fromkeys(products, 0)
         for line in order_lines:
-            products_qties[line.product_id] += line.product_uom_qty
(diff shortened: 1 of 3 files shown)


ltinel (approved):
LGTM, thanks for the fix :)
@robodoo r+

PR #247745, opened 2026-02-09
[FIX] website_sale: fix shop page filters

ltinel:
Partially undo [1] which wrongly converted some `t-set`/`t-value` assignments into `t-call` parameters.

[1] https://github.com/odoo/odoo/commit/bba2fc505f5d0b4770eacc6877155b1aeda6d772


vchu-odoo (approved):
Thank you for the fix!

@robodoo delegate+

PR #247746, opened 2026-02-09
[IMP] html_builder, website[_sale]: ui improvements

romo-odoo:
The VerticalJustifyOption is moved to the start of the option containers where the layout options are.
The scroll down button "spacing" option was changed to a slider to simplify edition.
Finally, the classList of some buttons was updated for consistency.

### DIFF
--- a/addons/website/static/src/builder/plugins/options/chart_option.xml
+++ b/addons/website/static/src/builder/plugins/options/chart_option.xml
@@ -54,7 +54,7 @@
                         action="'addColumn'"
                         className="'add_column fa fa-fw fa-plus text-success d-inline-block'"
                         title.translate="Add Dataset"
-                        type="' '"
+                        type="''"
                         preview="false"/>
                 </th>
             </tr>
@@ -82,7 +82,7 @@
                         actionParam="label_index"
                         className="'o_builder_matrix_remove_row visually-hidden-focusable text-danger fa fa-fw fa-minus'"
                         title.translate="Remove Row"
-                        type="' '"
+                        type="''"
                         preview="false"/>
                 </td>
             </tr>
@@ -93,7 +93,7 @@
                         action="'addRow'"
                         className="'add_row fa fa-fw fa-plus text-success d-inline-block'"
                         title.translate="Add Row"
-                        type="' '"
+                        type="''"
                         preview="false"/>
                 </th>
                 <t t-foreach="domState.data.datasets" t-as="dataset" t-key="dataset.key">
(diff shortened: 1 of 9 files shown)


addons/website/static/src/builder/plugins/options/scroll_button_option.js
```diff
@@ -15,6 +15,7 @@ export class ScrollButtonOption extends BaseOptionComponent {
                     ? _t("Min-Height")
                     : _t("Height"),
             heightFieldEnabled: editingElement.dataset.snippet === "s_image_gallery",
+            scrollDownButtonDisabled: editingElement.matches("footer section"),
```
bvr-odoo, line 18: Why not use the `exclude` property?

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -54,7 +54,7 @@
                         action="'addColumn'"
                         className="'add_column fa fa-fw fa-plus text-success d-inline-block'"
                         title.translate="Add Dataset"
-                        type="' '"
+                        type="''"
```
bvr-odoo, line 57: is it useful for something?

addons/website/static/src/builder/plugins/options/scroll_button_option.xml
```diff
@@ -39,14 +39,7 @@
...
         <BuilderRow label.translate="Spacing" level="1" applyTo="':scope > .o_scroll_button'">
-            <BuilderSelect>
-                <BuilderSelectItem classAction="''">None</BuilderSelectItem>
-                <BuilderSelectItem classAction="'mb-1'">Extra-Small</BuilderSelectItem>
-                <BuilderSelectItem classAction="'mb-2'">Small</BuilderSelectItem>
-                <BuilderSelectItem classAction="'mb-3'">Medium</BuilderSelectItem>
-                <BuilderSelectItem classAction="'mb-4'">Large</BuilderSelectItem>
-                <BuilderSelectItem classAction="'mb-5'">Extra-Large</BuilderSelectItem>
-            </BuilderSelect>
+            <BuilderRange action="'setClassRange'" actionParam="['', 'mb-1', 'mb-2', 'mb-3', 'mb-4', 'mb-5']" max="5"/>
```
bvr-odoo: Why not use `mb-0` instead of leaving it empty for none? That way, it won’t be necessary to modify the `apply` method in `SetClassRangeAction`.

bvr-odoo (changes requested):

addons/website/static/src/builder/plugins/options/chart_option.xml
romo-odoo: `type` is used to add the class `btn-{type}`
Here it adds `btn- ` in the classList

addons/website/static/src/builder/plugins/options/scroll_button_option.js
romo-odoo: Using the `exclude` property will remove all options, including the "Height" option.

image

bvr-odoo (approved):
ദ്ദി (˶˃ ᵕ ˂˶)

PR #247747, opened 2026-02-09
[MERGE] web,*: upgrade owl to version 3

ged-odoo:
This PR updates owl to version 3 and adapts the codebase accordingly. A compatibility layer
has been introduced to ease the transition. However, for some breaking changes, such a compatibility
layer wasn't possible.

Migration guide: https://odoo.github.io/owl/documentation/v3/owl/migration_owl2_to_owl3.html

OWL3 documentation: https://odoo.github.io/owl/documentation/v3/owl/

Enterprise PR: https://github.com/odoo/enterprise/pull/114326

Co-authored-by: Aaron Bohy <aab@odoo.com>
Co-authored-by: Achraf <abz@odoo.com>
Co-authored-by: Alex Kühn <aku@odoo.com>
Co-authored-by: Géry Debongnie <ged@odoo.com>
Co-authored-by: Bastien Pierre <ipb@odoo.com>
Co-authored-by: Jean Schoenlaub <jesc@odoo.com>
Co-authored-by: Julien Mougenot <jum@odoo.com>
Co-authored-by: Lucas Perais <lpe@odoo.com>
Co-authored-by: Michaël Mattiello <mcm@odoo.com>
Co-authored-by: Nicolas Bayet <nby@odoo.com>
Co-authored-by: Pierre Rousseau <pro@odoo.com>
Co-authored-by: Stephane Vanmeerhaeghe <stva@odoo.com>


addons/mail/static/src/core/common/composer_model.js
```diff
@@ -17,7 +17,11 @@ export class Composer extends Record {
         this.attachments.length = 0;
         this.replyToMessage = undefined;
         this.restoredFromFullComposer = false;
-        this.composerHtml = markup("<div class='o-paragraph'><br></div>");
+        if (this.updateFrom === "html") {
+            this.composerHtml = markup("<div class='o-paragraph'><br></div>");
```
seb-odoo, line 21: Note: should use markup` instead to avoid security flag.

addons/mail/static/src/core/common/thread.js
```diff
@@ -354,7 +359,7 @@ export class Thread extends Component {
             }
         });
         onWillDestroy(() => stopOnChange());
-        onWillPatch(() => {
+        onRendered(() => {
```
seb-odoo, line 362: What's the goal of the change?
In OWL 2, rendered might have been too early here, as the rendering was not synchronous with the patching and the goal was to capture the state of the DOM as late as possible before swapping it.
Has something changed?

addons/mail/static/src/core/common/navigable_list.js
```diff
@@ -54,7 +53,7 @@ export class NavigableList extends Component {
             () => {
                 this.open();
             },
-            () => [this.props]
+            () => [this.props?.options]
```
seb-odoo, line 56: Note: unrelated to the current PR, but this `useLayoutEffect` implementation is weak as chain of methods read values from `this` rather than from dependency parameter, which means we're never really sure we observe what we need.

It doesn't even feel like it should be a layout effect, as it seems to be only doing state management, it should probably be a standard effect instead.

addons/mail/static/src/discuss/gif_picker/common/gif_picker.xml
```diff
@@ -11,10 +11,9 @@
                 <div class="o-EmojiPicker-search d-flex align-items-center m-2 rounded">
                     <i t-if="!this.state.showCategories" aria-label="back" t-on-click="this.openCategories" class="oi oi-arrow-left cursor-pointer p-2 fs-5"/>
                     <span class="d-flex mx-1 w-100 rounded o-active">
-                        <t t-call="discuss.GifPicker.searchInput">
-                            <t t-if="this.props.state" t-set="localState" t-value="this.props.state"/>
-                            <t t-else="" t-set="localState" t-value="this.state"/>
-                        </t>
+                        <t t-if="this.props.state" t-set="localState" t-value="this.props.state"/>
+                        <t t-else="" t-set="localState" t-value="this.state"/>
+                        <t t-call="discuss.GifPicker.searchInput"/>
```
seb-odoo, line 16: `localState` doesn't need to be passed as parameter explicitly to the `t-call`?


```suggestion
                        <t t-call="discuss.GifPicker.searchInput" localState="this.props.state || this.state"/>
```

addons/mail/static/src/views/web/activity/activity_controller.js
```diff
@@ -24,7 +23,7 @@ export class ActivityController extends Component {
     static template = "mail.ActivityController";
 
     setup() {
-        this.model = useState(useModel(this.props.Model, this.modelParams));
+        this.model = proxy(useModel(this.props.Model, this.modelParams));
```
seb-odoo, line 26: Shouldn't this be built inside `useModel` instead?

addons/im_livechat/static/src/core/public_web/discuss_app/discuss_app_model_patch.js
```diff
@@ -1,48 +1,53 @@
...
+import { effectWithCleanup } from "@mail/utils/common/misc";
 
 // Looking for help subscription is triggered when the sidebar category is
 // opened, and when the discuss app is active. To avoid unsubscribing right away
 // when the user closes the sidebar or switches to another app, wait for 5
 // minutes before unsubscribing.
 export const LFH_UNSUBSCRIBE_DELAY = 5 * 60 * 1000;
 
+const DISPOSE_EFFECT_TIMEOUT_SYM = Symbol("DISPOSE_EFFECT_TIMEOUT");
+const DISPOSE_EFFECT_SYM = Symbol("DISPOSE_EFFECT");
```
seb-odoo, line 15: It's nice that we don't have to handle the reactive/dependencies with a function anymore, but wouldn't we want to keep `effectWithDebouncedCleanup` to keep the timeout implementation details internally rather than leaking these in the model/business code?

addons/mail/static/src/discuss/call/common/call_preview.js
```diff
@@ -60,46 +60,62 @@ export class CallPreview extends Component {
...
-            });
-            onChange(this.store.settings, ["edgeBlurAmount", "backgroundBlurAmount"], () => {
-                if (this.state.blurManager) {
-                    this.state.blurManager.edgeBlur = this.store.settings.edgeBlurAmount;
-                    this.state.blurManager.backgroundBlur =
-                        this.store.settings.backgroundBlurAmount;
-                }
-            });
+            const disposeFns = [];
+            disposeFns.push(
```
seb-odoo, line 64: Can't we initialize the array correctly immediately rather than doing many push in sequence?

addons/mail/static/src/core/common/im_status_mixin.js
```diff
@@ -27,41 +31,36 @@ export class ImStatusMixin extends Record {
...
+                busService.addChannel(presenceChannel);
+                return () => busService.deleteChannel(presenceChannel);
+            }
         });
         return record;
     }
+    delete(...args) {
+        this[DISPOSE_EFFECT_SYM_1]();
+        this[DISPOSE_EFFECT_SYM_2]();
+        super.delete(...args);
```
seb-odoo, line 62: Shouldn't these be done with `_registerDisposeFn` instead?

addons/im_livechat/static/src/core/public_web/discuss_app/discuss_app_model_patch.js
```diff
@@ -84,6 +89,10 @@ patch(DiscussApp.prototype, {
         this.livechats = fields.Many("discuss.channel", { inverse: "appAsLivechats" });
         this.isLivechatInfoPanelOpenByDefault = fields.Attr(true, { localStorage: true });
     },
+    delete(...args) {
+        this[DISPOSE_EFFECT_SYM]();
+        super.delete(...args);
```
seb-odoo, line 94: Not done with `_registerDisposeFn`?

addons/mail/static/src/core/common/composer.js
```diff
@@ -341,22 +308,24 @@ export class Composer extends Component {
...
-            if (!this.editor?.editable) {
-                return;
-            }
-            setElementContent(this.editor.editable, composerHtml);
-            this.setEditorCursorEnd();
-            this.editor.shared.history.addStep();
-        });
-        void composerProxy.composerHtml; // start observing
+        const composerProxy = reactive(this.props.composer);
+        onWillDestroy(
```
seb-odoo, line 354: Why is this moved to `onWillDestroy` only?

The reactive callback used to run whenever there was a change.

addons/mail/static/src/core/common/composer.js
```diff
@@ -538,26 +507,25 @@ export class Composer extends Component {
     }
 
     get navigableListProps() {
+        const { loading, searchTerm, results } = this.suggestion.search;
```
seb-odoo, line 552: Was this actually necessary diff? If so, what is it fixing?

addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -554,6 +552,7 @@ export class DiscussChannel extends Record {
     }
 
     delete() {
+        this[DISPOSE_EFFECT_SYM]();
```
seb-odoo, line 558: register dispose?

addons/mail/static/src/model/record_internal.js
```diff
@@ -11,12 +9,20 @@ import {
...
 import { RecordList } from "./record_list";
-import { toRaw } from "@odoo/owl";
+import { immediateEffect, toRaw, untrack } from "@odoo/owl";
 import { RecordUses } from "./record_uses";
 import { LocalStorageEntry } from "@mail/utils/common/local_storage";
 
 export class RecordInternal {
     [IS_RECORD_SYM] = true;
+    /**
+     * All dispose functions for this records.
```
seb-odoo, line 19:
```suggestion
     * All dispose functions for this record.
```

addons/mail/static/src/model/record_internal.js
```diff
@@ -58,12 +64,27 @@ export class RecordInternal {
...
      */
     fieldsComputeOnNeed = new Map();
-    /** @type {Map<string, () => void>} */
-    fieldsOnUpdateObserves = new Map();
-    /** @type {Map<string, Record>} */
-    fieldsSortProxy2 = new Map();
-    /** @type {Map<string, Record>} */
-    fieldsComputeProxy2 = new Map();
+    /**
+     * Fields that have an `compute` defined. Key is fieldName, Value is function of ongoing `immediateEffect` that let it stop.
```
seb-odoo, line 68:
```suggestion
     * Fields that have a `compute` defined. Key is fieldName, Value is function of ongoing `immediateEffect` that let it stop.
```
Stop is the same as dispose right? We might want to keep a consistent naming.

addons/mail/static/src/model/record_internal.js
```diff
@@ -58,12 +64,27 @@ export class RecordInternal {
...
+     * Useful to prevent any ongoing `immediateEffect` and restart if need be.
```
seb-odoo, line 69: Comment is not clear enough without prior knowledge.

addons/mail/static/src/model/record_internal.js
```diff
@@ -58,12 +64,27 @@ export class RecordInternal {
...
+     *
+     * @type {Map<string, Function>}
+     */
+    fieldsComputeStop = new Map();
+    /**
+     * Fields that have an `onUpdate` defined. Key is fieldName, Value is function of ongoing `onChange` that allow to stop it.
+     * Useful to prevent any ongoing onChange and restart if need be.
```
seb-odoo, line 76: same remark as above

addons/mail/static/src/model/record_internal.js
```diff
@@ -11,12 +9,20 @@ import {
...
 import { LocalStorageEntry } from "@mail/utils/common/local_storage";
 
 export class RecordInternal {
     [IS_RECORD_SYM] = true;
+    /**
+     * All dispose functions for this records.
+     * For the store, this stores the dispose functions of all records.
+     * Useful to automatically call the dispose functions when the record is deleted or in-between each tests.
+     *
+     * @type {Set<Function>}
```
seb-odoo, line 23: I believe `function` should be lower case (same in all docs).

addons/mail/static/src/utils/common/format.js
```diff
@@ -33,20 +33,24 @@ const messageUrlRegExp = new RegExp(`^${escapeRegExp(getOrigin())}/mail/message/
...
  * @returns {Promise<string|ReturnType<markup>>}
  */
-export function prettifyMessageText(rawBody, { validMentions = {}, thread } = {}) {
+export function prettifyMessageText(rawBody, { validMentions = {}, thread, trim = true } = {}) {
     if (rawBody instanceof markup().constructor) {
         // markup is already "pretty"
         return rawBody;
     }
-    let body = htmlTrim(rawBody);
+    let body = trim ? htmlTrim(rawBody) : htmlJoin([rawBody]);
```
seb-odoo, line 55: I don't see the point of doing dummy join.

Normally nothing should be done if trim is not necessary.

If we want to force markup, which I don't think is a good idea, the most canonical way would be explicit htmlEscape.

addons/mail/static/src/utils/common/format.js
```diff
@@ -33,20 +33,24 @@ const messageUrlRegExp = new RegExp(`^${escapeRegExp(getOrigin())}/mail/message/
...
     body = htmlReplace(body, /(\r|\n){2,}/g, () => markup`<br/><br/>`);
     body = htmlReplace(body, /(\r|\n)/g, () => markup`<br/>`);
     body = htmlReplace(body, /&nbsp;/g, () => " ");
-    body = htmlTrim(body);
+    body = trim ? htmlTrim(body) : htmlJoin([body]);
```
seb-odoo, line 59: same remark

addons/mail/static/src/utils/common/format.js
```diff
@@ -402,9 +380,13 @@ export function htmlToTextContentInline(htmlString) {
...
-export function convertBrToLineBreak(str) {
-    str = htmlReplace(str, /<br\s*\/?>/gi, () => "\n");
-    return createDocumentFragmentFromContent(str).body.textContent;
+export function convertBrToLineBreak(str, { trim = true } = {}) {
+    const regex = trim ? /<br\s*\/?>/gi : /<br\/?>/gi;
+    str = htmlReplace(str, regex, () => "\n");
+    if (!trim) {
+        str = htmlReplace(str, /\s/g, () => markup`&nbsp;`);
+    }
+    return createDocumentFragmentFromContent(str).body.textContent.replaceAll(" ", " ");
```
seb-odoo, line 421: This change and API is really confusing.

The param is named `trim` but it doesn't even do trimming in the usual meaning of the word.

addons/mail/static/src/utils/common/format.js
```diff
@@ -402,9 +380,13 @@ export function htmlToTextContentInline(htmlString) {
...
 }
 
-export function convertBrToLineBreak(str) {
-    str = htmlReplace(str, /<br\s*\/?>/gi, () => "\n");
-    return createDocumentFragmentFromContent(str).body.textContent;
+export function convertBrToLineBreak(str, { trim = true } = {}) {
+    const regex = trim ? /<br\s*\/?>/gi : /<br\/?>/gi;
+    str = htmlReplace(str, regex, () => "\n");
+    if (!trim) {
+        str = htmlReplace(str, /\s/g, () => markup`&nbsp;`);
```
seb-odoo, line 419: `&nbsp;` looks completely unrelated to br to line break

addons/mail/static/src/utils/common/format.js
```diff
@@ -402,9 +380,13 @@ export function htmlToTextContentInline(htmlString) {
         .replace(/\s\s+/g, " ");
 }
 
-export function convertBrToLineBreak(str) {
-    str = htmlReplace(str, /<br\s*\/?>/gi, () => "\n");
-    return createDocumentFragmentFromContent(str).body.textContent;
+export function convertBrToLineBreak(str, { trim = true } = {}) {
+    const regex = trim ? /<br\s*\/?>/gi : /<br\/?>/gi;
```
seb-odoo, line 416: Why would it matter to check or ignore space inside the br tag when converting to line break, and how is that related to trimming?

addons/mail/static/src/discuss/core/common/discuss_category_model.js
```diff
@@ -1,33 +1,39 @@
...
+            if (busService && busChannel) {
+                busService.addChannel(busChannel);
+                return () => busService.deleteChannel(busChannel);
+            }
         });
         return category;
     }
 
+    delete(...args) {
+        this[DISPOSE_EFFECT_SYM]();
```
seb-odoo, line 24: register dispose?

addons/mail/static/src/discuss/core/common/discuss_category_model.js
```diff
@@ -1,33 +1,39 @@
...
+        super.delete(...args);
+    }
+
+    busChannelSubscribed = false;
```
seb-odoo, line 28: dead variable?

addons/mail/static/src/model/record_list.js
```diff
@@ -49,12 +51,12 @@ function isSortOnNeed(reclist) {
 
 /** @param {RecordList} reclist */
 function computeField(reclist) {
-    reclist._.owner._.compute(reclist._.owner, reclist._.name);
+    reclist._.owner._.compute(reclist._.owner, reclist._.name, { fromInNeed: true });
```
seb-odoo, line 54: It's confusing to set `fromInNeed` true here, as there is no check about in need anywhere in this part of the code.

addons/mail/static/src/model/record_list.js
```diff
@@ -49,12 +51,12 @@ function isSortOnNeed(reclist) {
...
 }
 
 /** @param {RecordList} reclist */
 function sortField(reclist) {
-    reclist._.owner._.sort(reclist._.owner, reclist._.name);
+    reclist._.owner._.sort(reclist._.owner, reclist._.name, { fromInNeed: true });
```
seb-odoo, line 59: It's confusing to set `fromInNeed` true here, as there is no check about in need anywhere in this part of the code.

addons/mail/static/src/model/store.js
```diff
@@ -294,21 +297,31 @@ export class Store extends Record {
             }
         }
         if (Array.isArray(key)) {
+            /** @type {Function[]} */
+            const arrayDisposeFns = [];
+            arrayDisposeFns.forEach((f) => f());
+            arrayDisposeFns.length = 0;
```
seb-odoo, line 303: List is initialized empty just above, `forEach` and `length=0` are useless?

addons/mail/static/src/utils/common/misc.js
```diff
@@ -95,17 +96,30 @@ export function onChange(target, key, callback) {
         }
     }
     if (Array.isArray(key)) {
+        /** @type {Function[]} */
+        const arrayDisposeFns = [];
+        arrayDisposeFns.forEach((f) => f());
+        arrayDisposeFns.length = 0;
```
seb-odoo, line 102: same remark

addons/mail/static/tests/core/record.test.js
```diff
@@ -692,26 +674,26 @@ test("lazy compute should re-compute while they are observed", async () => {
     }).register(localRegistry);
     const store = await start();
     const channel = store.Channel.insert(1);
-    let observe = true;
+    const reactiveChannel = reactive(channel);
```
seb-odoo, line 701: What's the point of wrapping in reactive? Aren't records always reactive?

addons/mail/static/tests/tours/discuss_sub_channel_search_tour.js
```diff
@@ -28,18 +27,12 @@ registry.category("web_tour.tours").add("test_discuss_sub_channel_search", {
...
-                                if (status(this) === "destroyed") {
-                                    return;
-                                }
-                                if (!state.isVisible) {
-                                    resolveLoadMoreDisappeared?.();
-                                }
-                            },
-                            [this.loadMoreState]
-                        );
+                        useEffect(() => {
```
seb-odoo, line 30: no need to dispose this one?

addons/mail/static/tests/tours/discuss_channel_public_tour.js
```diff
@@ -94,16 +94,19 @@ registry.category("web_tour.tours").add("discuss_channel_public_tour.js", {
                         throw new Error(`Attachment was not found from src: ${src}`);
                     }
                     if (!attachment.raw_access_token) {
+                        let disposeEffect;
                         await new Promise((resolve) => {
-                            const proxy = reactive(attachment, () => {
+                            const proxifiedAttachment = proxy(attachment);
```
seb-odoo, line 99: why proxy?

addons/mail/static/tests/tours/discuss_channel_public_tour.js
```diff
@@ -94,16 +94,19 @@ registry.category("web_tour.tours").add("discuss_channel_public_tour.js", {
...
+                            disposeEffect = effect(() => {
                                 if (attachment.raw_access_token) {
                                     resolve();
                                 } else {
-                                    void proxy.raw_access_token; // keep observing until a value is received
+                                    void proxifiedAttachment.raw_access_token; // keep observing until a value is received
```
seb-odoo, line 104: Can't we read from `attachment` directly in owl 3, with no  need for these tricks?

addons/mail/static/tests/discuss_app/discuss.test.js
```diff
@@ -2429,6 +2429,7 @@ test("Correct breadcrumb when open discuss from chat window then see settings as
     await contains("[title='Open Actions Menu']");
     await click("[title='Open Actions Menu']");
     await click(".o-dropdown-item:text('Open in Discuss')");
+    await contains(".o-mail-DiscussSidebarChannel:has(:text('General'))");
```
seb-odoo, line 2432: This is strange, as it is already asserted on the next line.
Is this an artificial way to add a tick?

addons/mail/static/tests/discuss/core/discuss_performance.test.js
```diff
@@ -142,9 +140,10 @@ test("right-click message selection should only render relevant part", async ()
     await contains(".o-mail-Message", { count: 10 });
     const stopObserve = observeRenders();
     rightClicking = true;
+
```
seb-odoo, line 143: no empty lines

addons/mail/static/tests/discuss/core/discuss_performance.test.js
```diff
@@ -60,14 +60,14 @@ test("posting new message should only render relevant part", async () => {
     await insertText(".o-mail-Composer-input", "Test");
     const result1 = stopObserve1();
     // LessThan because renders could be batched
-    expect(result1.get(Message)).toBeLessThan(11); // 10: all messages initially
+    expect(result1.get(Message)).toBeLessThan(21); // 10: all messages initially
```
seb-odoo, line 63: Why such an increase?

addons/mail/static/src/core/common/navigable_list.js
alexkuhn: The effects on `props` are poor. This is the minimal diff to make it work for now.

addons/im_livechat/static/src/core/public_web/discuss_app/discuss_app_model_patch.js
alexkuhn: The function `effectWithDebouncedCleanup` was too cumbersome to keep. The combination of `predicate` and `deps` required user of function to understand implementation details.

Sure the code could be improved, but this felt easier to just kill the util especially when this is only used here.

addons/mail/static/src/discuss/call/common/call_preview.js
alexkuhn: This doesn't change much, using explicit `disposeFn` in component is weird already, so better cleanup code more in depth at some point in master after this PR is merged.

addons/mail/static/src/core/common/im_status_mixin.js
alexkuhn: This diff was made quite early when `_registerDisposeFn` didn't exist (was added quite recently). Yes this can probably use `_registerDisposeFn`, though not a big deal for now. I think we can do this code cleaning after PR is merged.

addons/im_livechat/static/src/core/public_web/discuss_app/discuss_app_model_patch.js
alexkuhn: Same remark: was added before `_registerDisposeFn` was introduced. Can be cleaned at some point to use `_registerDisposeFn` indeed.

addons/mail/static/src/model/record_internal.js
alexkuhn: I plan to do some work in `/model/` of discuss. Again changes here have been made as quick-fix.

addons/mail/static/src/utils/common/format.js
alexkuhn: I used `htmlJoin()` because there's no other function in `web/utils/html` that minimally `markup(htmlEscape())`. This is the simplest function already implemented there.

aab-odoo:
robodoo rebase-merge

addons/mail/static/src/utils/common/format.js
alexkuhn: The trim is implicit from HTML text node that does the trim. Putting `nbsp` is a workaround to have textnode preserve the whitespaces, so that we can use the same trick with `createDocumentFragmentFromContent.body.textContent`.

There are a few drawbacks for the technique here, but the code of syncing `composerText` / `composerHtml` was wrong and the `owl3` branch revealed the problem. This is a fix to have improved syncing of the 2 fields while keeping mostly the implementation details unchanged (except for the `!trim`, which is the behavior we definitely want for correct syncing of the 2 fields).

addons/mail/static/src/utils/common/format.js
alexkuhn: The code here is replace whitespace by `nbsp`, so that the "whitespace" are kept in text node, for the `createDocumentFragmentFromContent.body.textContent` below. This is important for good syncing of `composerText` / `composerHtml` that is buggy in master actually.

addons/mail/static/src/discuss/core/common/discuss_category_model.js
alexkuhn: Oops indeed

addons/mail/static/src/model/record_list.js
alexkuhn: This was done to keep internal functions of `RecordList` the same, with `reclist` in param.
This function is only used when making a read on a computed field of a record list so this is fine.
But I get that this assumes the function is used for that purpose.

addons/mail/static/src/model/store.js
alexkuhn: Hm... yes the `.foreach()` and `length = 0` are useless here. Residual code from before I've put this in the `return` that acts as a `disposeFn` 🙈

d-fence:
robodoo alone

addons/mail/static/tests/core/record.test.js
alexkuhn: Yes the `reactive()` is probably not needed. Prior code was `reactive(channel, render)`, the wrap of `reactive()` was only kept to play safe.

addons/mail/static/tests/tours/discuss_sub_channel_search_tour.js
alexkuhn: This is used in a tour anyway, and the tour controls with `resolveLoadMoreDisappeared`, so dispose shouldn't be necessary other than adding more LOCs.

addons/mail/static/tests/tours/discuss_channel_public_tour.js
alexkuhn: `reactive()` is an alias for `proxy()`. I suppose this was kept because previous code was `reactive(attachment, callback)` so to play safe.

addons/mail/static/tests/discuss/core/discuss_performance.test.js
alexkuhn: Code in master has `onMounted()` and `onPatched()`, at some point this was changed to `onMounted()` and `onRendered()`, and now this is only `onRendered()`. I presume the increase from `onMounted()` that was counted twice with `onRendered()`, but the counter hasn't been adapted back to `11`.

So yes I think this should be kept to `11`.

seb-odoo:
For some reason I'm not able to find where `immediateEffect` is defined. I see it on the lib/owl file, but nowhere on the owl repo. Could anyone point me in the right direction?

aab-odoo:
Hello @xmo-odoo 

May I ask you to check the ci/security alerts triggered on the owl3 migration PR?

> [addons/web/static/lib/owl/owl.js:1775](https://github.com/odoo/odoo/blob/8e413c3aaf2342403c0c44ab2dca5d3916f48f07/addons/web/static/lib/owl/owl.js#L1775-L1775)
> [JS004 🔗](https://github.com/odoo/enterprise/wiki/security#JS004): markup(String(str))
> [addons/web/static/lib/owl/owl.js:1787](https://github.com/odoo/odoo/blob/8e413c3aaf2342403c0c44ab2dca5d3916f48f07/addons/web/static/lib/owl/owl.js#L1787-L1787)
> [JS004 🔗](https://github.com/odoo/enterprise/wiki/security#JS004): markup(str)

These two markup calls come from the htmlEscape function of owl and are done either on a safe string or on a string that has been escaped.

> [addons/web/static/lib/owl/owl.js:2792](https://github.com/odoo/odoo/blob/8e413c3aaf2342403c0c44ab2dca5d3916f48f07/addons/web/static/lib/owl/owl.js#L2792-L2792)
> [JS003 🔗](https://github.com/odoo/enterprise/wiki/security#JS003): template.innerHTML = this.html
> [addons/web/static/lib/owl/owl.js:2822](https://github.com/odoo/odoo/blob/8e413c3aaf2342403c0c44ab2dca5d3916f48f07/addons/web/static/lib/owl/owl.js#L2822-L2822)
> [JS003 🔗](https://github.com/odoo/enterprise/wiki/security#JS003): template.innerHTML = html2

are both in owl, in the `mount` and `patch` functions which obviously need to manipulate the DOM.

Thanks

xmo-odoo:
@robodoo override=ci/security

addons/html_builder/static/src/core/building_blocks/builder_list.js
```diff
@@ -42,50 +38,69 @@ export function useAutoFocusNewItem(refName) {
...
+
+    setup() {
+        this.props.setupLayoutEffect();
+    }
+}
+
 export class BuilderList extends Component {
     static template = "html_builder.BuilderList";
-    static props = {
-        ...basicContainerBuilderComponentProps,
```
FrancoisGe, line 48: Why did you remove `basicContainerBuilderComponentProps` ?

addons/html_builder/static/src/core/building_blocks/builder_list.js
```diff
@@ -42,50 +38,69 @@ export function useAutoFocusNewItem(refName) {
...
+    static template = xml`<t t-call-slot="default"/>`;
+
+    props = props({
+        setupLayoutEffect: t.function(),
+    });
+
+    setup() {
+        this.props.setupLayoutEffect();
+    }
+}
```
FrancoisGe, line 51: Why do you need to add `SortableContainer` ?

addons/html_builder/static/src/core/building_blocks/builder_number_input_base.js
```diff
@@ -25,7 +25,8 @@ export class BuilderNumberInputBase extends BuilderInputBase {
             values.forEach((value, i) => {
                 values[i] = this.props.clampValue(value + (e.key === "ArrowUp" ? step : -step));
             });
-            e.target.value = values.join(" ");
+            this.state.value = values.join(" ");
+            e.target.value = this.state.value;
```
FrancoisGe, line 29: Why did you need this changes ? The state should be manage by the parent.

addons/html_editor/static/src/main/font/font_size_selector.js
```diff
@@ -100,6 +100,11 @@ export class FontSizeSelector extends Component {
                 this.fontSizeInput.addEventListener("click", () => {
                     if (!this.dropdown.isOpen) {
                         this.dropdown.open();
+                        requestAnimationFrame(() => {
+                            if (this.menuRef.el?.closest(".o_bottom_sheet")) {
+                                this.props.onBlur?.();
+                            }
+                        });
```
FrancoisGe, line 107: Why do you need this change ?

addons/website/static/src/builder/plugins/options/social_media_links.js
```diff
@@ -93,6 +93,9 @@ export class SocialMediaLinks extends BaseOptionComponent {
      * @returns { SocialMediaLinkItem[] }
      */
     computeItems() {
+        // subscribe to signal to update render
+        void this.reorderTriggered.trigger;
```
FrancoisGe, line 97: It looks like an hack ?

addons/website/static/src/client_actions/website_preview/website_builder_action.js
```diff
@@ -88,19 +96,19 @@ export class WebsiteBuilderClientAction extends Component {
...
-                    }
-                    this.toggleIsMobile(websiteContext.isMobile);
-                },
-                [this.websiteContext]
-            );
+            disposeToggleMobileEffect = immediateEffect(() => {
+                this.websiteContext.isMobile; // consume signal
+                if (status(this.component) !== "mounted") {
+                    return;
+                }
```
FrancoisGe, line 107: Why we don t just check `=== "destroyed"` like before ?

addons/website/static/src/client_actions/website_preview/website_builder_action.js
```diff
@@ -88,19 +96,19 @@ export class WebsiteBuilderClientAction extends Component {
...
-            effect(
-                (websiteContext) => {
-                    if (status(this.component) === "destroyed") {
-                        return;
-                    }
-                    this.toggleIsMobile(websiteContext.isMobile);
-                },
-                [this.websiteContext]
-            );
+            disposeToggleMobileEffect = immediateEffect(() => {
```
FrancoisGe, line 103: Why do you need `immediateEffect` vs `useEffect` ?

addons/mail/static/tests/discuss/core/discuss_performance.test.js
jeanschoenlaub: For the record, I was the one working on this part, but lost the commit in squash, so I don't fully remember, but I will look into this again once merged. But I do believe what Alex said is right and we could revert the check `toBeLessThan(11)`.

aab-odoo:
robodoo override=ci/style

mcm-odoo:
> For some reason I'm not able to find where `immediateEffect` is defined. I see it on the lib/owl file, but nowhere on the owl repo. Could anyone point me in the right direction?

Hello @seb-odoo 
The commit for `immediateEffect` is on [this branch](https://github.com/odoo/owl/commits/master-owl3-migration/) on odoo/owl repo. You can see there is a few other commits.
These additionnal commits only exist to help us during the migration and should be removed from odoo in the future.

addons/website/static/src/client_actions/website_preview/website_builder_action.js
mcm-odoo: `immediateEffect` is trigger synchronously when a subscribed signal changes and `(use)effect` waits a micro tick (`Promise.resolve()`) before it triggers.
Here, waiting a micro tick was maybe too long and caused issues.

addons/website/static/src/client_actions/website_preview/website_builder_action.js
mcm-odoo: There were certainly multiple steps to this. I might removed the `onMounted` to only have the effect and changed the condition. Noticed later that it didn't work and reintroduced the `onMounted` without changing the effect behaviour.

addons/website/static/src/builder/plugins/options/social_media_links.js
mcm-odoo: Yes, we needed to read the state to subscribe the effect (re-render)

addons/html_builder/static/src/core/building_blocks/builder_number_input_base.js
mcm-odoo: Yep, the state should be managed by the parent. I can't remember the reasons but this was necessary, we made multiple fix for this component.

addons/html_builder/static/src/core/building_blocks/builder_list.js
mcm-odoo: It's a bit complicated to explain but when there is a render only `SortableContainer` triggers `onPatched`. Not the `BuilderList` nor `BuilderComponent`.

addons/html_builder/static/src/core/building_blocks/builder_list.js
mcm-odoo: There is a test that check the props validation but `static props = {...}` does not work anymore now with owl3.
To make this test pass, we converted the props of component only. Of course, props defined by `basicContainerBuilderComponentProps` are kept in this component.

aab-odoo:
robodoo r+

aab-odoo:
Due to the heavy workload of maintaining these branches in // of master and the other JS teams waiting for them to be merged to unlock their own developments, we'll tackle the comments post merge (in agreement with those having posted them, as none are really blocking). As well as the few non deterministic hoot tests that were skipped.

addons/html_editor/static/src/main/font/font_size_selector.js
jeanschoenlaub: We had a bug on hoot: `should focus the editable area after selecting a font size item on mobile`


Better explanation of the problem here:
https://github.com/odoo/odoo/commit/f10c5da44f02fa6e2a07468c2a760c39643bc1f7
==> But that commit broke some stuff on desktop so I tried multiple approach and went for this:
https://github.com/odoo/odoo/commit/41d332e1ad22b38e12c9bf0af9987ecc614f194c


TLDR: is with OWL2 the bottom sheet appeared before and triggering select on the hidden input didn't focus it, with OWL3 it does.

addons/mail/static/tests/discuss/core/discuss_performance.test.js
jeanschoenlaub: I just reran the test on master on we do have 20 renders which does not seem normal. I will check with Alex on who is going to fix this.

PR #247748, opened 2026-02-09
[FIX] pos_sale,sale: set tax on down payment

robinengels:
In certain conditions all lines in invoice require a tax. When making
a down payment from an order containing products using fixed price tax,
the corresponding invoice line was created without tax. The issue appear
both when making the down payment from the sale order and from the PoS.

Steps to reproduce:
-------------------
* Create a fixed price tax of 10€
* Create a product with this tax
* Create a sale order with this product and make a downpayment of 10%
> Observation: The down payment line has no tax set.
* Open PoS and make a down payment of 10% for the same order
* Pay and invoice the order
> Observation: The down payment line has no tax set.

Why the fix:
------------
If the tax is required on every invoice line we manually add a 0% tax to
the down payment line to ensure that the invoice is compliant.

At the moment we only add the tax when peppol is activated on the
current company. But the `_require_tax_ids_on_invoice_lines` method can
be overriden by other modules if downpayment lines also require tax.

opw-5853070

### DIFF
--- a/addons/account/models/account_tax.py
+++ b/addons/account/models/account_tax.py
@@ -4348,6 +4348,25 @@ class AccountTax(models.Model):
             return ''
         return html2plaintext(self.description)
 
+    def _get_zero_tax(self):
+        """ Return the 0% tax for the current company, creating it if it does not exist yet.
+        """
+        zero_tax = self.search([
+            ('amount_type', '=', 'percent'),
+            ('amount', '=', 0),
+            ('type_tax_use', 'in', ['sale']),
+            *self.env['account.tax']._check_company_domain(self.env.company),
+        ], limit=1)
+        if not zero_tax:
+            zero_tax = self.create({
+                'name': '0% Tax',
+                'amount_type': 'percent',
+                'amount': 0,
+                'type_tax_use': 'sale',
+                'company_id': self.env.company.id,
+            })
+        return zero_tax
+
 
 class AccountTaxRepartitionLine(models.Model):
     _name = "account.tax.repartition.line"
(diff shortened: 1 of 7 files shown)


robinengels:
Hello @gawa-odoo Could I have a first review from the accounting POV for this PR ? 

@odoo/rd-sales I'd also need a review from your side

addons/account/models/account_tax.py
```diff
@@ -4348,6 +4348,25 @@ def _get_description_plaintext(self):
...
         return html2plaintext(self.description)
 
+    def _get_zero_tax(self):
+        """ Return the 0% tax for the current company, creating it if it does not exist yet.
+        """
+        zero_tax = self.search([
+            ('amount_type', '=', 'percent'),
+            ('amount', '=', 0),
+            ('type_tax_use', 'in', ['sale']),
+            ('company_id', '=', self.env.company.id),
```
gawa-odoo: The only question is if it works well with branches. Aren't there cases where we have the CoA on the main company and not on the branches? (I don't remember well)

addons/account/models/account_tax.py
```diff
@@ -4348,6 +4348,25 @@ def _get_description_plaintext(self):
...
         return html2plaintext(self.description)
 
+    def _get_zero_tax(self):
+        """ Return the 0% tax for the current company, creating it if it does not exist yet.
+        """
+        zero_tax = self.search([
+            ('amount_type', '=', 'percent'),
+            ('amount', '=', 0),
+            ('type_tax_use', 'in', ['sale']),
+            ('company_id', 'parent_of', self.env.company.id),
```
gawa-odoo: I think the "proper" way would be to do `*self.env['account.tax']._check_company_domain(self.env.company)` (so if we change the behaviour of branches and co, it should still work)

gawa-odoo:
@robodoo r-
I suppose POS needs to check?

But LGTM from accounting POV

davidmonnom (approved):
@robodoo r+

PR #247754, opened 2026-02-09
[FIX] web: fix copy button submitting form

sdegueldre:
Previously, the type of the button in the template of the CopyButton utility component was left unspecified. Because the default type for buttons is "submit", the copy button will not work if it is placed within a `<form>` element, and will instead submit the form (see [1]).

This commit just forces the type of the button to "button" which has no default behavior, meaning it can be used even inside of `<form>` elements without issues.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type


aab-odoo (approved):
robodoo r+

PR #247757, opened 2026-02-09
[IMP] stock: zpl product label layout

loouis-t:
To allow displaying more information on the small zpl product label, we reduce the font size and add an ellipsis on the title.

task-4100420


Ysoroko (approved):
Image

PR #247759, opened 2026-02-09
[FIX] web_editor: detect drop outside of dropzone in a RTL language

Syozik:
When dropping outside a dropzone but still on the page, the code checks
if the drop happened well outside of the sidebar (so on its left).
However, in RTL languages, the sidebar is positioned on the left, so we
need to check if the drop is on the right side of it instead.

The fix checks if the sidebar is at the left edge (the body of the
document should have the `o_rtl` class) and verifies the drop position
is on the right of the sidebar.

Steps to reproduce:
- Set your profile to Arabic
- Drag and drop a snippet outside of a dropzone

=> It's not dropped, but it should, as it would with an LTR language.

task-5484936

### DIFF
--- a/addons/web_editor/static/src/js/editor/snippets.editor.js
+++ b/addons/web_editor/static/src/js/editor/snippets.editor.js
@@ -3803,7 +3803,12 @@ class SnippetsMenu extends Component {
                         dropped = true;
                     }
                 }
-                if (!dropped && y > 3 && x + helper.getBoundingClientRect().height < this.el.getBoundingClientRect().left) {
+                const sidebarRect = this.el.getBoundingClientRect();
+                const isRTL = document.body.classList.contains("o_rtl");
+                const isOutOfSidebar = isRTL
+                    ? sidebarRect.left + sidebarRect.width < x - helper.getBoundingClientRect().width / 2
+                    : x + helper.getBoundingClientRect().width / 2 < sidebarRect.left;
+                if (!dropped && y > 3 && isOutOfSidebar) {
                     const point = { x, y };
                     let droppedOnNotNearest = touching(doc.body.querySelectorAll('.oe_structure_not_nearest'), point);
                     // If dropped outside of a dropzone with class oe_structure_not_nearest,


Dj0ulo (changes requested):
Thanks for the fix!

Maybe extract the part you added in a variable because it's a bit hard to read.

Also either add a comment or replace `elRect.x === 0` by something that checks if the page is in RTL (there is `o_rtl` class on the body for example) because it's not clear why we are doing that without checking the commit message.

Dj0ulo (approved):
LGTM (hoot test to be added in saas-18.4)

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
                         dropped = true;
```
sobo-odoo, line 3803: Commit message:
- > if the drop is not on a sidebar by verifying it's on the left side of the sidebar

  Nitpick: I would reformulate like this `if the drop happened well outside of the sidebar (so on its left)` to be more clear :wink: 

- > the drop position is to the right of the sidebar's width.

  Nitpick: technically, you can't be on the right of a width :stuck_out_tongue:, I would simply say this instead: `is on the right of the sidebar` :slightly_smiling_face: 

- > as in would with a LTR language

  -> as it would

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
...
                     }
                 }
-                if (!dropped && y > 3 && x + helper.getBoundingClientRect().height < this.el.getBoundingClientRect().left) {
+                const elRect = this.el.getBoundingClientRect();
+                // Sidebar can be positioned on either side of the page:
+                // - Right side (typical LTR layout): elRect.x > 0
+                // - Left side (RTL layout): elRect.x = 0
+                // Check if drop position is outside sidebar bounds based on its position.
+                const isOutsideOfSidebar =
+                    elRect.x === 0
```
sobo-odoo: You should use `.closest(".o_rtl")` to see if we are in RTL, instead of checking that the sidebar x position is 0, because checking the x position is not robust, as we cannot guarantee the sidebar is well at 0 :slightly_smiling_face: 

For example, given that this file is in web_editor, it means that it is also used in mass_mailing, and as you can see here, the sidebar is not really at 0 because we are in a form view :arrow_down: 
image

Therefore, your fix doesn't work in mass_mailing :+1: 

There is always the `o_rtl` class in the body when we are in RTL, so it is more robust to use that information instead :slightly_smiling_face:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
...
+                        ? elRect.width < x
```
sobo-odoo: As you can see when we are in LTR, the dragged thumbnail really has to be outside of the sidebar to be considered as dropped, but with your fix, in RTL, it is already considered as dropped when only half of the thumbnail is on the page

This is a nitpick because it is better at the half than never, but could you make the behavior consistent with LTR ? :slightly_smiling_face:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
...
+                        : x + helper.getBoundingClientRect().width < elRect.x;
```
sobo-odoo: Thanks for fixing `height` to `width` (I had it in mind since a long time haha)
However, I preferred `left` to `x`, it is easier to visualize when reading the code :slightly_smiling_face:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
                         dropped = true;
                     }
                 }
-                if (!dropped && y > 3 && x + helper.getBoundingClientRect().height < this.el.getBoundingClientRect().left) {
+                const elRect = this.el.getBoundingClientRect();
+                // Sidebar can be positioned on either side of the page:
+                // - Right side (typical LTR layout): elRect.x > 0
+                // - Left side (RTL layout): elRect.x = 0
+                // Check if drop position is outside sidebar bounds based on its position.
```
sobo-odoo: You can remove these comments if you declare an `isRTL` variable to use in your condition :+1: It will be straightforward to understand with the `isRTL` and `isOutsideOfSidebar` variables :slightly_smiling_face:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
...
+                const isOutsideOfSidebar =
```
sobo-odoo: Suggestion: for a shorter variable name, I propose `isOutOfSidebar` :+1:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,16 @@ class SnippetsMenu extends Component {
                         dropped = true;
                     }
                 }
-                if (!dropped && y > 3 && x + helper.getBoundingClientRect().height < this.el.getBoundingClientRect().left) {
+                const elRect = this.el.getBoundingClientRect();
```
sobo-odoo: This is technically the best name since it is the rect of `this.el`, but in that context, maybe call it `sidebarRect` ? :thinking:

sobo-odoo (changes requested):

addons/web_editor/static/src/js/editor/snippets.editor.js
Syozik: Really good catch, thanks :pray: 
Having thought about it, there's an opposite problem in LTR if the distance between the thumbnail and the sidebar is less than half of the sidebar (the thumbnail is on the left of the sidebar), it won't be dropped:

https://github.com/user-attachments/assets/35ae4b89-61f2-4297-93e8-325912045096


I pushed a small change that fixes this, too. `x` is the center of the thumbnail, not the left corner, so we should calculate half of the thumbnail's width to make it work correctly.

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,13 @@ class SnippetsMenu extends Component {
                         dropped = true;
```
sobo-odoo, line 3803: Commit message:
- > checks if the sidebar is at the left edge (x === 0) and

  Since you now check `o_rtl`, you have to change this too :wink:

addons/web_editor/static/src/js/editor/snippets.editor.js
```diff
@@ -3803,7 +3803,13 @@ class SnippetsMenu extends Component {
...
                     }
                 }
-                if (!dropped && y > 3 && x + helper.getBoundingClientRect().height < this.el.getBoundingClientRect().left) {
+                const sidebarRect = this.el.getBoundingClientRect();
+                const isRTL = document.body.classList.contains("o_rtl");
+                const isOutOfSidebar = isRTL
+                    ? sidebarRect.left + sidebarRect.width <
+                      x - helper.getBoundingClientRect().width / 2
+                    : x + helper.getBoundingClientRect().width / 2 < sidebarRect.left;
```
sobo-odoo: Nitpick: in order to make the first computation fit in one line and to not repeat `helper.getBoundingClientRect()`, maybe you can create a `helperRect` variable (so next to `sidebarRect`) ? :slightly_smiling_face: 

Note that we didn't have a linter in 18.0, so you can clearly already make it on one line without the variable too :wink: (maybe it's better to do this, so we don't create a variable for "nothing" :+1: )

sobo-odoo (changes requested):
Last details and we will be good :slightly_smiling_face:

addons/web_editor/static/src/js/editor/snippets.editor.js
sobo-odoo: Not sure if you forgot, but you didn't adapt the message :slightly_smiling_face: We are not checking if it is at the left with x = 0 anymore but checking if the backend is in rtl :slightly_smiling_face:

Syozik: My bad, somehow I didn't push after I amended the message, sorry.

sobo-odoo (approved):
Thanks for the fix :slightly_smiling_face: 

@robodoo r+

PR #247760, opened 2026-02-09
[FIX] point_of_sale: change payment_type for refund

aksi-odoo:
Step to reproduce:
- Install point_of_sale
- Enable Identify Customer on the Card payment method
- Create an order with a customer and refund it
- Use Card as the payment method
- Close the POS session
- Go to Invoicing → Customers → Payments

Observation:
- Two payment records are created
- Both payments have payment_type = inbound
- The refund payment should be outbound

Cause:
- When Identify Customer is enabled, `_create_split_account_payment` is used to create payment records
- The method does not adjust payment_type for refund transactions

Fix:
- Add helpers to swap destination and outstanding accounts
- Set `force_outstanding_account_id` instead of `outstanding_account_id`, as the former has priority
- Ensure refund payments are created as `outbound`

few related fix: https://github.com/odoo/odoo/commit/303a9061da85048f14a3ca7b1e13df0ab34da99e
https://github.com/odoo/odoo/commit/718fac6832ecd343bf26d41fa5ae5b1ab74f4228
https://github.com/odoo/odoo/commit/684415b9ff2e151506da561016dbfa991bfa8dc8

opw-5437456

---

### DIFF
--- a/addons/point_of_sale/models/pos_session.py
+++ b/addons/point_of_sale/models/pos_session.py
@@ -1081,6 +1081,18 @@ class PosSession(models.Model):
         data['pay_later_move_lines'] = MoveLine.create(vals)
         return data
 
+    def _ensure_payment_outstanding_account(self, payment, payment_amount):
+        # In community the outstanding account is computed on the creation of account.payment records
+        if not payment.outstanding_account_id and self.env['account.move']._get_invoice_in_payment_state() == 'in_payment':
+            payment.outstanding_account_id = payment._get_outstanding_account(payment.payment_type)
+
+        if float_compare(payment_amount, 0, precision_rounding=self.currency_id.rounding) < 0:
+            payment.write({
+                'force_outstanding_account_id': payment.destination_account_id,
+                'destination_account_id': payment.outstanding_account_id,
+                'payment_type': 'outbound',
+            })
+
     def _create_combine_account_payment(self, payment_method, amounts, diff_amount):
         outstanding_account = payment_method.outstanding_account_id
         destination_account = self._get_receivable_account(payment_method)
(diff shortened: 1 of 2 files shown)


alsh-odoo (approved):

addons/point_of_sale/models/pos_session.py
```diff
@@ -1080,6 +1080,19 @@ def _create_pay_later_receivable_lines(self, data):
...
+            payment.outstanding_account_id = payment._get_outstanding_account(payment.payment_type)
+
+    def _swap_accounts_if_negative_amount(self, payment, payment_amount):
+        if float_compare(payment_amount, 0, precision_rounding=self.currency_id.rounding) < 0:
+            payment.write({
+                'force_outstanding_account_id': payment.destination_account_id,
+                'destination_account_id': payment.outstanding_account_id,
+                'payment_type': 'outbound',
+            })
+
```
stva-odoo: Wouldn’t it be simpler to just have a single function, since both functions are called in each case?


```python
    def _ensure_payment_outstanding_account(self, payment, payment_amount):
        # In community the outstanding account is computed on the creation of account.payment records
        if not payment.outstanding_account_id and self.env['account.move']._get_invoice_in_payment_state() == 'in_payment':
            payment.outstanding_account_id = payment._get_outstanding_account(payment.payment_type)

        if float_compare(payment_amount, 0, precision_rounding=self.currency_id.rounding) < 0:
            payment.write({
                'force_outstanding_account_id': payment.destination_account_id,
                'destination_account_id': payment.outstanding_account_id,
                'payment_type': 'outbound',
            })


  ....
 self._ensure_payment_outstanding_account(account_payment, amounts['amount'])
   

```

aksi-odoo: Hello @stva-odoo  thanks for review, i wrote them differently as their intent were different, 
However, it's not a problem. they are merged now :)

aksi-odoo: oops. messed up bad 

edit: fixed it

dc-oerp:
Hi @aksi-odoo @stva-odoo 

I am facing the same issue as @niyasraphy and this is reproducible on Runbot.
Could you please fix this

Thank you

aksi-odoo:
Hello @dc-oerp @niyasraphy   i think this will solve that https://github.com/odoo/odoo/pull/256258

niyasraphy:
Hi @aksi-odoo,

The PR has been ready for over three weeks, and there has still been no movement to merge it. Meanwhile, the system continues to generate incorrect data, which is becoming increasingly concerning.

I’m honestly quite dissatisfied with the current handling of this issue. When a known fix is already available, allowing the problem to persist for this long without action or a clear plan to correct existing incorrect entries is not acceptable from a data integrity perspective.

I would also like to understand what the actual plan is to fix the already-affected records.

@dc-oerp @kripal754

PR #247761, opened 2026-02-09
[FIX] portal: fix error 500 when info edit with no country

MAGREMENT:
How to reproduce:
- Create a company with no country
- Set a contact's company to that company
- Grant portal access to that contact
- Login as that contact
- Go to the edit information tab
- Leave some of the required fields (Phone, Street & City) empty
- Click on save

The problem:
The page displays an error 500

Why:
When you try to submit the form with some required field missing, the page will try to evaluate this expression : 'int(country_id)''. But since this commit (https://github.com/odoo/odoo/commit/d6d6bee087fe2d3dc17974054353430c2662aecf), the post variable country_id is set to "False" if the partner has no country. 'int("False")' will then raise an error.

opw-5867975

---


Ephoniz (approved):
LGTM 👍🏽

PR #247765, opened 2026-02-09
[PERF] mail: avoid sending inbox only-fields in chatter

seb-odoo:
These fields are only necessary in inbox/systray, not in chatter, so they don't need to be fetched/computed when not necessary.


phenix-factory (approved):
LGTM
@robodoo delegate+ r+

PR #247769, opened 2026-02-09
[FIX] point_of_sale: ensure NumberBuffer uses live localization separators

amahm-odoo:
Steps to reproduce:
1. Open the Point of Sale.
2. In Odoo Settings, change the Decimal Separator (e.g., from '.' to ',').
3. Go back to the PoS and add a Tip.
4. Observe that the tip math is wrong.

Cause:
The 'NumberBuffer' service is a singleton initialized at PoS startup. During initial setup, it caches the decimal point from 'services.localization.decimalPoint' into 'this.defaultDecimalPoint'. Because the service is a singleton, this value remains stale if settings are changed without a server refresh. When the 'NumberPopup' is opened, it uses the cached stale separator, causing parsing issues in methods like 'addTip'.

Solution:
Modify the 'NumberBuffer' service to fetch the decimal separator directly from the 'localization' service during the '_setUp' process.

opw-5895622

### DIFF
--- a/addons/point_of_sale/static/tests/pos/tours/payment_screen_tour.js
+++ b/addons/point_of_sale/static/tests/pos/tours/payment_screen_tour.js
@@ -1,3 +1,5 @@
+/* global posmodel */
+
 import * as Chrome from "@point_of_sale/../tests/pos/tours/utils/chrome_util";
 import * as Dialog from "@point_of_sale/../tests/generic_helpers/dialog_util";
 import * as ProductScreen from "@point_of_sale/../tests/pos/tours/utils/product_screen_util";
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/static/tests/pos/tours/payment_screen_tour.js
```diff
@@ -225,3 +226,47 @@ registry.category("web_tour.tours").add("test_add_money_button_with_different_de
...
+            {
+                content: "Switch localization to comma",
+                trigger: "body",
+                run: () => {
+                    const env = odoo.__WOWL_DEBUG__.root.env;
+                    env.services.localization.decimalPoint = ",";
+                    env.services.localization.thousandsSep = ".";
+                    env.services.number_buffer._setUp();
+                },
+            },
```
sbel-odoo, line 248: You can use posmodel for this. Put this line as the first line of the test file:
```
/* global posmodel */
```

```suggestion
            {
                content: "Switch localization to comma",
                trigger: "body",
                run: () => {
                    posmodel.numberBuffer.localization.decimalPoint = ",";
                    posmodel.numberBuffer.localization.thousandsSep = ".";
                    posmodel.numberBuffer._setUp();
            },
```
Same comment applies below as well.

sbel-odoo (commented):
I'm good with this fix good catch. I've left a comment regarding your tour.

amahm-odoo:
@stva-odoo could you take a look on this PR

stva-odoo (commented):
robodoo r+

amahm-odoo:
@robodoo r+

stva-odoo:
robodoo r+

PR #247770, opened 2026-02-09
[FIX] pos*: ensure invoice selection before order validation in tour

parp-odoo:
pos*: point_of_sale, pos_sale, l10n_sa_edi_pos, l10n_es_pos,l10n_be_pos_sale

When invoice selection takes longer, and the order validation button is clicked immediately after, the tour may serialize data before the invoice field has settled. This can cause invoice generation to be skipped during order validation

Since the delay between tour steps was removed, this commit adds an explicit step to ensure the invoice is selected before validating the order.

Additionally, removed unused tour `PosSettleAndInvoiceOrder`

Task-5897375
Err-237600, 238502

Related-https://github.com/odoo/enterprise/pull/106863

### DIFF
--- a/addons/point_of_sale/static/tests/pos/tours/utils/payment_screen_util.js
+++ b/addons/point_of_sale/static/tests/pos/tours/utils/payment_screen_util.js
@@ -107,14 +107,18 @@ export function clickPaymentline(name, amount) {
         },
     ];
 }
-export function clickInvoiceButton() {
-    return [
+export function clickInvoiceButton(isHighlighted = true) {
+    const steps = [
         {
             content: "click invoice button",
             trigger: ".payment-buttons .js_invoice",
             run: "click",
         },
     ];
+    if (isHighlighted) {
+        steps.push(...isInvoiceButtonChecked());
+    }
+    return steps;
 }
 export function clickValidate() {
     return [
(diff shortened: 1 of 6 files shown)


addons/point_of_sale/static/tests/pos/tours/utils/payment_screen_util.js
```diff
@@ -107,14 +107,18 @@ export function clickPaymentline(name, amount) {
...
+export function clickInvoiceButton(shouldVerifyHighlighted = true) {
+    const steps = [
         {
             content: "click invoice button",
             trigger: ".payment-buttons .js_invoice",
             run: "click",
         },
     ];
+    if (shouldVerifyHighlighted) {
+        steps.push(...isInvoiceButtonChecked());
```
jipr-odoo, line 119:
```
        isInvoiceButtonChecked() and isInvoiceOptionSelected()
```

Both utilities are doing the same thing. Can we remove one and keep only a single?

parp-odoo: Removed `isInvoiceOptionSelected`

davidmonnom:
@robodoo r+

davidmonnom (approved):
@robodoo r+

PR #247777, opened 2026-02-09
[FIX] analytic, account: Allow (set/not set) filter for analytic distribution search

HazemWaheed2050:
**Problem:** 
When filtering by "Analytic Distribution" in views, using "is set" or "is not set" filters (or searching for False) causes a "search domain not valid" error or returns incorrect results.

**Steps to reproduce:**
1) Go to Accounting > Journal Entries.
2) Apply a filter for Invoice lines > Distribution Analytic Account.
3) Select "is set" or "is not set".
4) Check records.

Issue produces error "search domain not valid". Also reproduceable on Purchase Orders.

**Cause:** 
The `_condition_to_sql` method did not handle the `('=', '!=')` operators early enough, so it always went to the "operation not supported" error.

**Solution:**
- We need to move the handling of the operators '=', '!=' early enough in the `_condition_to_sql` method.

opw-5478688

---

### DIFF
--- a/addons/account/tests/test_account_move_entry.py
+++ b/addons/account/tests/test_account_move_entry.py
@@ -1351,3 +1351,18 @@ class TestAccountMove(AccountTestInvoicingCommon):
         }])
         move.line_ids.account_id = shared_account
         move.action_post()
+
+    def test_journal_entry_analytic_distribution_search_is_set(self):
+        """ Verify searching on analytic_distribution with 'is set', 'is not set'."""
+        analytic_plan = self.env['account.analytic.plan'].create({'name': 'Plan'})
+        analytic_account = self.env['account.analytic.account'].create({
+            'name': "Test Account",
+            'plan_id': analytic_plan.id,
+        })
+        self.assertFalse(self.env['account.move.line'].search([('analytic_distribution', '!=', False)]))
+        self.test_move.line_ids[0]['analytic_distribution'] = {str(analytic_account.id): 100}
+        self.assertEqual(
+            self.env['account.move.line'].search([('analytic_distribution', '!=', False)]),
+            self.test_move.line_ids[0],
+            ""
+        )
--- a/addons/analytic/models/analytic_mixin.py
+++ b/addons/analytic/models/analytic_mixin.py
@@ -74,7 +74,7 @@ class AnalyticMixin(models.AbstractModel):
         # Don't use this override when account_report_analytic_groupby is truly in the context
         # Indeed, when account_report_analytic_groupby is in the context it means that `analytic_distribution`
         # doesn't have the same format and the table is a temporary one, see _prepare_lines_for_analytic_groupby
-        if field_expr != 'analytic_distribution' or self.env.context.get('account_report_analytic_groupby'):
+        if field_expr != 'analytic_distribution' or self.env.context.get('account_report_analytic_groupby') or (isinstance(value, bool) and operator in ('=', '!=')):
             return super()._condition_to_sql(alias, field_expr, operator, value, query)
 
         def search_value(value: str, exact: bool):
@@ -98,11 +98,6 @@ class AnalyticMixin(models.AbstractModel):
         if not ids:
             # not ids found, just call super with an empty list
             return super()._condition_to_sql(alias, field_expr, operator, ids, query)
-
-        if isinstance(value, int) and operator in ('=', '!='):
-            value = [value]
-            operator = 'in' if operator == '=' else 'not in'
-
         # keys can be comma-separated ids, we will split those into an array and then make an array comparison with the list of ids to check
         analytic_accounts_query = self._query_analytic_accounts()
         ids = [str(id_) for id_ in ids if id_]  # list of ids -> list of string


addons/account/tests/test_account_move_entry.py
```diff
@@ -1351,3 +1351,73 @@ def test_post_invoice_fails_with_account_and_journal_company_inconsistency(self)
         }])
         move.line_ids.account_id = shared_account
         move.action_post()
+
+    def test_journal_entry_analytic_distribution_search_is_set(self):
```
yolannsabaux, line 1355: You are testing on a journal entry but create a whole setup for an account move. I think your test could be way shorter. 

```
        analytic_plan = self.env['account.analytic.plan'].create({'name': 'Plan'})
        analytic_account = self.env['account.analytic.account'].create({
            'name': "Test Account",
            'plan_id': analytic_plan.id,
        })
        self.assertFalse(self.env['account.move.line'].search([('analytic_distribution', '!=', False)]))
        self.test_move.line_ids[0].write({'analytic_distribution': {str(analytic_account.id): 100}})
        self.assertEqual(
            self.env['account.move.line'].search([('analytic_distribution', '!=', False)]),
            self.test_move.line_ids[0],
            ""
        )
```

addons/analytic/models/analytic_mixin.py
```diff
@@ -82,6 +82,17 @@ def search_value(value: str, exact: bool):
                 [('display_name', ('=' if exact else 'ilike'), value)]
             ))
 
+        if isinstance(value, int) and operator in ('=', '!='):
```
yolannsabaux: Would this not be enough ? 
```diff
                 [('display_name', ('=' if exact else 'ilike'), value)]
             ))
 
+        if isinstance(value, bool) and operator in ('=', '!='):
+            return super()._condition_to_sql(alias, field_expr, operator, value, query)
+
         # reformulate the condition as <field> in/not in <ids>
         if operator in ('in', 'not in'):
             ids = [
```

addons/purchase/tests/test_purchase.py
```diff
@@ -1101,3 +1101,63 @@ def test_purchase_order_line_without_uom(self):
             uom_test.unlink()
 
         self.assertEqual(po.order_line[0].product_uom_id, uom_test)
+
+    def test_purchase_order_analytic_distribution_search_is_set(self):
```
yolannsabaux: same as for the journal entry comment

yolannsabaux (changes requested):
Hello,

Thanks for the fix!

Tests are appreciated, although I don't think they need to be such overkill (see comment).

For the fix per se, I think that my proposal may be a bit more relevant, as we return early if it's an 'is (not) set'.

Let me know

HazemWaheed2050:
@yosa-odoo Done the requested changes :+1:

yolannsabaux (approved):
LGTM

Wondering if the test in purchase is really necessary tough

addons/analytic/models/analytic_mixin.py
```diff
@@ -82,6 +82,9 @@ def search_value(value: str, exact: bool):
                 [('display_name', ('=' if exact else 'ilike'), value)]
             ))
 
+        if isinstance(value, bool) and operator in ('=', '!='):
+            return super()._condition_to_sql(alias, field_expr, operator, value, query)
```
epictete: Then let's group all the early return conditions together at the top ☺️

addons/account/tests/test_account_move_entry.py
```diff
@@ -1351,3 +1351,20 @@ def test_post_invoice_fails_with_account_and_journal_company_inconsistency(self)
...
+        """
+            Verify searching on analytic_distribution with 'is set', 'is not set'.
+        """
```
epictete:
```suggestion
        """Verify searching on analytic_distribution with 'is set', 'is not set'."""
```
See [here](https://peps.python.org/pep-0257/#one-line-docstrings) 🤓

addons/account/tests/test_account_move_entry.py
```diff
@@ -1351,3 +1351,20 @@ def test_post_invoice_fails_with_account_and_journal_company_inconsistency(self)
...
+        analytic_plan = self.env['account.analytic.plan'].create({'name': 'Plan'})
+        analytic_account = self.env['account.analytic.account'].create({
+            'name': "Test Account",
+            'plan_id': analytic_plan.id,
+        })
+        self.assertFalse(self.env['account.move.line'].search([('analytic_distribution', '!=', False)]))
+        self.test_move.line_ids[0].write({'analytic_distribution': {str(analytic_account.id): 100}})
```
epictete:
```suggestion
        self.test_move.line_ids[0]['analytic_distribution'] = {str(analytic_account.id): 100}
```
Nitpick 😉

addons/purchase/tests/test_accrued_purchase_orders.py
epictete, line 1: I agree with @yosa-odoo, there's no value in having an additional test for purchase orders 🤷 
Let's keep it simple.

epictete (commented):
Thanks for the fix! 👌 

Just a few cosmetic details.

epictete (approved):
Thanks for the quick update!

@robodoo r+

PR #247778, opened 2026-02-09
[IMP] mail: show when no suggestions are found

Atovange:
Before this commit, when showing suggestions in the composer (e.g. mentions, canned responses, commands) and no suggestions are found nothing would show.

This commit improves the composer suggestions by showing "No partner found" (appropriate label for each suggestion type) when the search does not return any results.

task-5873751

### DIFF
--- a/addons/mail/static/src/core/common/composer.js
+++ b/addons/mail/static/src/core/common/composer.js
@@ -57,6 +57,13 @@ const EDIT_CLICK_TYPE = {
     SAVE: "save",
 };
 export const MENTION_AMOUNT_WARNING = 50;
+const NO_SUGGESTIONS_LABELS = {
+    Partner: _t("No contacts found"),
+    "discuss.channel": _t("No channels found"),
+    ChannelCommand: _t("No commands found"),
+    "mail.canned.response": _t("No canned responses found"),
+    emoji: _t("No emojis found"),
+};
 
 class FullComposerRecoveryPopover extends Component {
     static props = ["composer", "onClickFullRecover", "onClickTextRecover", "close?"];
@@ -490,10 +497,11 @@ export class Composer extends Component {
     }
 
     get hasSuggestions() {
-        return Boolean(this.suggestion?.state.items);
+        return Boolean(this.suggestion?.state.suggestions?.length);
     }
 
     get navigableListProps() {
+        const isLoading = Boolean(this.suggestion.search.term) && this.suggestion.state.isFetching;
         const props = {
             anchorRef: this.inputContainerRef.el,
             position: this.env.inChatter ? "bottom-fit" : "top-fit",
@@ -501,19 +509,19 @@ export class Composer extends Component {
                 this.suggestion.insert(option);
                 markEventHandled(ev, "composer.selectSuggestion");
             },
-            isLoading: !!this.suggestion.search.term && this.suggestion.state.isFetching,
+            isLoading,
             options: [],
         };
+        const { type, suggestions } = this.suggestion.state;
         if (!this.hasSuggestions) {
+            if (type && !isLoading && !this.suggestion.search.isOutOfBounds) {
+                props.options = [{ label: NO_SUGGESTIONS_LABELS[type], unselectable: true }];
+            }
             return props;
         }
         return {
             ...props,
-            ...mapSuggestionsToOptions(
-                this.suggestion.state.items.type,
-                this.suggestion.state.items.suggestions,
-                { thread: this.thread }
-            ),
+            ...mapSuggestionsToOptions(type, suggestions, { thread: this.thread }),
         };
     }
 
--- a/addons/mail/static/src/core/common/suggestion_hook.js
+++ b/addons/mail/static/src/core/common/suggestion_hook.js
@@ -84,13 +84,15 @@ export class UseSuggestion {
     suggestionService = useService("mail.suggestion");
     state = useState({
         count: 0,
-        items: undefined,
+        type: undefined,
+        suggestions: undefined,
         isFetching: false,
     });
     search = {
         delimiter: undefined,
         position: undefined,
         term: "",
+        isOutOfBounds: undefined,
     };
     lastFetchedSearch;
     get isSearchMoreSpecificThanLastFetch() {
@@ -114,7 +116,7 @@ export class UseSuggestion {
             position: undefined,
             term: "",
         });
-        this.state.items = undefined;
+        Object.assign(this.state, { type: undefined, suggestions: undefined });
     }
     detect() {
         let start = 0;
@@ -146,6 +148,7 @@ export class UseSuggestion {
         const candidatePositions = [];
         // consider the chars before the current cursor position
         let numberOfSpaces = 0;
+        this.considerationBoundary = -1;
         for (let index = start - 1; index >= 0; --index) {
             if (/\s/.test(text[index])) {
                 numberOfSpaces++;
@@ -154,6 +157,7 @@ export class UseSuggestion {
                     // a majority of partners have a two-word name. This
                     // removes the need to check for mentions following a
                     // delimiter used earlier in the content.
+                    this.considerationBoundary = index;
                     break;
                 }
             }
@@ -200,6 +204,7 @@ export class UseSuggestion {
                 delimiter: candidateDelimiter,
                 position: candidatePosition,
                 term: text.substring(candidatePosition + candidateDelimiter.length, start),
+                isOutOfBounds: this.considerationBoundary > candidatePosition,
             });
             this.state.count++;
             return;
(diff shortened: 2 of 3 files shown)


addons/mail/static/src/core/common/composer.js
```diff
@@ -55,6 +55,13 @@ const EDIT_CLICK_TYPE = {
     SAVE: "save",
 };
 export const MENTION_AMOUNT_WARNING = 50;
+export const NO_SUGGESTIONS_LABELS = {
+    Partner: _t("No partner found"),
+    "discuss.channel": _t("No channel found"),
+    ChannelCommand: _t("No command found"),
+    "mail.canned.response": _t("No canned response found"),
+    emoji: _t("No emoji found"),
+};
```
phenix-factory, line 67: It doesn't need to be exported (yet).
It's quite specific to the suggestion list, I don't think we could reuse this list for anything else

addons/mail/static/tests/discuss_app/discuss.test.js
```diff
@@ -449,7 +449,7 @@ test("reply to message from inbox (message linked to document)", async () => {
     await contains(".o-mail-Composer");
     await contains(".o-mail-Composer-coreHeader:has(:text('on: Mitchell Admin'))");
     await insertText(".o-mail-Composer-input:focus", "Hello");
-    await press("Enter");
+    await click(".o-mail-Composer button[title='Send']:enabled");
     await contains(".o-mail-Composer", { count: 0 });
```
phenix-factory, line 455: Why did you need to change this? Enter should still work no?

Atovange: The composer at that point in the test contains `@Refactoring Hello` which makes the "no suggestions found" item still present. Pressing Enter dismisses it but does not send the message. By clicking the button I avoid having to do Enter twice.

addons/mail/static/src/core/common/composer.js
```diff
@@ -57,6 +57,13 @@ const EDIT_CLICK_TYPE = {
     SAVE: "save",
 };
 export const MENTION_AMOUNT_WARNING = 50;
+const NO_SUGGESTIONS_LABELS = {
+    Partner: _t("No partner found"),
+    "discuss.channel": _t("No channel found"),
+    ChannelCommand: _t("No command found"),
+    "mail.canned.response": _t("No canned response found"),
+    emoji: _t("No emoji found"),
```
oomsveta: Note that in English, 0 is usually plural:
> Although the everyday meaning of _plural_ is "more than one", the grammatical term has a slightly different technical meaning. In the English system of grammatical number, singular means "one (or minus one)", and plural means "not singular". In other words, plural means not just "more than one" but also "less than one (except minus one)". This less-than aspect can be seen in cases like _the temperature is zero degrees_ (not *_zero degree_) and _0.5 children per woman_ (not *_0.5 child per woman_).
> 
> https://en.wikipedia.org/wiki/English_plurals#Meaning

So it'd look more natural to write "_No channel**s** found_", "_No command**s** found_", etc.

addons/mail/static/src/core/common/composer.js
```diff
@@ -57,6 +57,13 @@ const EDIT_CLICK_TYPE = {
     SAVE: "save",
 };
 export const MENTION_AMOUNT_WARNING = 50;
+const NO_SUGGESTIONS_LABELS = {
+    Partner: _t("No partner found"),
```
oomsveta: "Partner" is the name of the technical model, but it shouldn't be used in user-facing strings, unless it specifically refers to [Odoo Partners](https://www.odoo.com/partners). The best term depends on the context, but for general use, "Contact" is generally proper.

addons/mail/static/src/core/common/composer.js
```diff
@@ -490,30 +497,31 @@ export class Composer extends Component {
...
                 this.suggestion.insert(option);
                 markEventHandled(ev, "composer.selectSuggestion");
             },
-            isLoading: !!this.suggestion.search.term && this.suggestion.state.isFetching,
+            isLoading,
             options: [],
         };
+        const { type, suggestions } = this.suggestion.state;
         if (!this.hasSuggestions) {
+            if (type && !isLoading && !this.suggestion.search.isOutOfBounds) {
```
oomsveta, line 517: Could you clarify why testing for `type` is needed here?

addons/mail/static/src/core/common/suggestion_hook.js
```diff
@@ -200,6 +204,7 @@ export class UseSuggestion {
                 delimiter: candidateDelimiter,
                 position: candidatePosition,
                 term: text.substring(candidatePosition + candidateDelimiter.length, start),
+                isOutOfBounds: this.considerationBoundary > candidatePosition,
```
oomsveta, line 207: Could explain the rationale for `isOutOfBounds`? What problem does it solve?

addons/mail/static/src/core/common/composer.js
Atovange: `type` is undefined when no known delimiter has been used (check `searchSuggestions@suggestion_service.js`).

It's not strictly necessary thanks to the `hasSuggestions` guard, it was probably put there before this piece of code got moved around.

That said the relation "suggestions is truthy <-> type is truthy" has no strict guarantee, so I would keep it especially since `type` is used to access an object property.

addons/mail/static/src/core/common/suggestion_hook.js
Atovange: It's solving the problem of deciding whether or not the user has stopped searching for a suggestion.
It's mostly an issue of spaces: we stop looking for suggestion after two spaces.
To illustrate:
`@Joh` -> "No contacts found"
`@John Sm` -> "No contacts found"
`@John Smith please send the report` -> don't show anything (the user stopped looking for a contact).

oomsveta: Instead of flagging the suggestion as "out of bound", do you think you could just not push it to the suggestion list at all?

Atovange: But we could have the situation of a contact's name containing more than 2 spaces, say for instance in Odoo we have: "John Smith (josm)".
In that case we want to show the contact suggestion if the user typed out the whole string.
Another case would be long spanish names ("Maria Teresa García Ramírez de Arroyo")

addons/mail/static/src/core/common/composer.js
```diff
@@ -57,6 +57,13 @@ const EDIT_CLICK_TYPE = {
     SAVE: "save",
 };
 export const MENTION_AMOUNT_WARNING = 50;
+const NO_SUGGESTIONS_LABELS = {
+    Partner: _t("No contacts found"),
+    "discuss.channel": _t("No channels found"),
+    ChannelCommand: _t("No commands found"),
+    "mail.canned.response": _t("No canned responses found"),
+    emoji: _t("No emojis found"),
+};
```
alexkuhn, line 66: Would define a typedef for `type` of suggestion, and document here and other uses of `type` this refers to this typedef.

That way the expected value is made very clear in code.

_(and I plan to change some `type` in follow-up PR, like `Partner` should become `res.partner`)_

addons/mail/static/src/core/common/suggestion_hook.js
alexkuhn: I feel that we should stop supporting the whitespace altogether.

This was a lesser problem before this PR, but with the showing of "No X found", this is a problem because the showing of this UI element is far too persistent, like if you type an email address it's already not great to display the suggestion for all characters after `@`, but this is made worse that a single whitespace is not enough to remove it.

If you want to search names with lots of word, I would rely on fuzzy search and not consider whitespace at all. For example if you search for "Maria Teresa García Ramírez de Arroyo" you can type ""MariaArroyo" and it should match.

Atovange: > If you want to search names with lots of word, I would rely on fuzzy search and not consider whitespace at all. For example if you search for "Maria Teresa García Ramírez de Arroyo" you can type ""MariaArroyo" and it should match.

I am not sure how intuitive that would be from a user perspective.
Perhaps we should restrict the "no x found" feature  to only specific types? After all the task only talks about Canned Responses, I expanded it to all suggestion types, but I do agree it's a bit too sticky.

alexkuhn: I think generalizing to all suggestions is a good move, though if we don't find a good consensus then we could limit to `::` for now as this is rarely typed outside of genuine attempt to insert a canned response.

> I am not sure how intuitive that would be from a user perspective.

I think this would be very intuitive, as thanks to the "No X found" this shows when mention suggestion is showing. If people add a whitespace, this is made very clear that this doesn't trigger the suggestion list. People are pretty used to type with fuzzy search, for which no whitespace is quite frequent.

BTW the whitespace being strict stop matches behavior of competitors like Discord.

PR #247780, opened 2026-02-09
[ADD] account_peppol_selfbilling: Backport self-billing to 17.0

antoine162:
Backport of 8c590efd369b to 17.0.

In existing modules, added ability to receive via Peppol and decode BIS3 self-billing invoices as sale invoices.

In new module account_peppol_selfbilling, added ability to export self-billed invoices and send them via Peppol.

Vendor bills are considered to be self-billed invoices if they are created in a journal where the `is_self_billing` field is True.

task-5865317


gawa-odoo:
As we forgot in the fw-port, let's atleast do it here.
You need to add the selfbilling to the weblate

gawa-odoo (commented):
@robodoo r+

PR #247781, opened 2026-02-09
[FIX] website_sale: close global search modal when entering edit mode

paru-odoo:
Steps to reproduce:
- Go to the Shop and enable the floating toolbar.
- Click the search icon in the header.
- When the search modal opens, click 'Edit' to enable website editing.
- Try to add a snippet from the floating toolbar.
=> The same issue also occurs on individual product pages.

Observed behavior:
Snippets cannot be added while the global search modal remains open.

Expected behavior:
Snippets should be draggable and added normally in edit mode.

This PR ensures that the global search modal is closed when entering edit mode, preventing it from blocking add snippet.

task-5905963

### DIFF
--- a/addons/website_sale/static/src/js/website_sale.js
+++ b/addons/website_sale/static/src/js/website_sale.js
@@ -669,7 +669,7 @@ export const WebsiteSale = publicWidget.Widget.extend(VariantMixin, {
 publicWidget.registry.WebsiteSale = WebsiteSale
 
 publicWidget.registry.WebsiteSaleSearchModal = publicWidget.Widget.extend({
-    selector: '#o_wsale_search_modal',
+    selector: "#o_wsale_search_modal, #o_wsale_product_search_modal",
     disabledInEditableMode: true,
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


addons/website_sale/static/src/js/website_sale.js
```diff
@@ -669,7 +669,7 @@ export const WebsiteSale = publicWidget.Widget.extend(VariantMixin, {
 publicWidget.registry.WebsiteSale = WebsiteSale
 
 publicWidget.registry.WebsiteSaleSearchModal = publicWidget.Widget.extend({
-    selector: '#o_wsale_search_modal',
+    selector: "#o_wsale_search_modal, #o_wsale_product_search_modal",
```
msh-odoo, line 672: I was thinking to use ```css_editable_mode_hidden``` in PublicWidget as selector and handle all ```css_editable_mode_hidden``` there but if there is only 1-2 occurrences where we have to explicitly do this then I am ok with this solution, did you check all the places where we have used ```css_editable_mode_hidden``` ?

msh-odoo: @paru-odoo I tested all the places where we have used ```css_editable_mode_hidden```, all the cases working fine except this modal

paru-odoo:
> Hello @paru-odoo
> 
> I was unable to reproduce the issue on runbot on 18.4 can you provide a video?

Steps are already mentioned in the commit message, and here is a video reference: https://www.loom.com/share/3b350fb8dbec42dbafbb39bde26af12c.

livaios (approved):
LGTM 👍 
Thanks for the fix and for the explanation!

ltinel (approved):
LGTM, thanks :)
@robodoo r+

PR #247782, opened 2026-02-09
[FIX] point_of_sale: correctly group products by category

davidmonnom:
Before this commit when activating the "Group by category" option in the POS, some products were missing from the product list.

Now we use the `getAllBy` generic model method to group products by category, which ensures that all products are correctly grouped and displayed in the POS.

### DIFF
--- a/addons/point_of_sale/static/src/app/services/pos_store.js
+++ b/addons/point_of_sale/static/src/app/services/pos_store.js
@@ -2678,6 +2678,21 @@ export class PosStore extends WithLazyGetterTrap {
         );
     }
 
+    orderProductBySequenceAndFav(products) {
+        const searchWord = this.searchProductWord.trim();
+        const isSearchByWord = searchWord !== "";
+        return isSearchByWord
+            ? products.sort((a, b) => b.is_favorite - a.is_favorite)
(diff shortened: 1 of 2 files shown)


addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -2756,58 +2756,58 @@ export class PosStore extends WithLazyGetterTrap {
...
+                : products;
+
+            if (filtered.length) {
+                // Its advised to not use group by categ with too much products in differents
+                // categories, but in case of we end up with too much products, we slice them in
+                // group of 100 to avoid freezing the browser tab.
+                // We cannot just slice the products to display and keep the same category, because
+                // we want to avoid having categories with only few products displayed and others
+                // with a lot of products not displayed.
+                results.push([catId, filtered.splice(0, 100)]);
```
adgu-odoo: you could sort the filtered produts like we do at the end of `get productsToDisplay` to show favorite first and if there is no searchword, sort by sequence and then by name comparaison. You should do this before slicing

adgu-odoo (commented):
robodoo delegate+

PR #247788, opened 2026-02-09
[IMP] html_editor: support separator command within list

adsa-odoo:
Currently when a user attempts to use the separator command within a list, no action is triggered. This PR ensures that separator command is supported within a list. If the separator is inserted on a list item, list is split into two and separator is inserted after first list.

task-5912576

---

### DIFF
--- a/addons/html_editor/static/src/main/list/list_plugin.js
+++ b/addons/html_editor/static/src/main/list/list_plugin.js
@@ -8,6 +8,7 @@ import {
     wrapInlinesInBlocks,
 } from "@html_editor/utils/dom";
 import {
+    getDeepestEditablePosition,
     getDeepestPosition,
     isElement,
     isEmptyBlock,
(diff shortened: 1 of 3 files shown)


addons/html_editor/static/src/main/list/list_plugin.js
```diff
@@ -1281,6 +1282,32 @@ export class ListPlugin extends Plugin {
         }
     }
 
+    /**
+     * Exit the list from given list item by splitting
+     * the list.
+     *
+     * @param {HTMLElement} blockEl
+     */
+    exitList(blockEl) {
```
deso-odoo: Can't you use `splitList` or `liToBlocks`?

addons/html_editor/static/src/main/list/list_plugin.js
```diff
@@ -1281,6 +1282,32 @@ export class ListPlugin extends Plugin {
...
+            li = blockEl.parentElement;
+        }
+        if (li) {
+            // Helper li to split the list
+            const [, after] = this.dependencies.split.splitElementBlock({
+                targetNode: blockEl,
+                targetOffset: nodeSize(blockEl),
+                blockToSplit: li,
+            });
+            this.dependencies.selection.setCursorStart(firstLeaf(after));
```
deso-odoo: I'd personally prefer to use `getDeepstEditablePosition` over `firstLeaf` for such case.

adsa-odoo:
@dmo-odoo Conflicts are resolved, thanks!

PR #247792, opened 2026-02-09
[FIX] hr: avoid access error editing bank account

taqu-odoo:
**Steps to reproduce**
- have `hr` and `account` installed
- add a bank account to a contact
- switch to a user with accounting rights but no HR rights
- on the contact form, click on the tag of the bank account in the "Bank accounts" field
- Access Error due to missing rights related to employee model

**Cause**
The `_compute_employee_id` method requires read access on the `hr.employee` model.

**Solution**
Hide the field for non-HR users.

opw-5217072


amah-odoo (approved):
Thanks for your work 💯 

@robodoo r+

PR #247795, opened 2026-02-09
[FIX] portal, website_slides: empty messages should not be fetched in portal

maki-odoo:
PR #246783 added a new `data-o-datetime` metadata attribute to the `edited` tag.
 The fetch domain in portal uses the same tag to filter empty messages and
 prevent them from being fetched. This change adapts the fetch domain in portal
 to take into account the recent change to the `edited` tag.

### DIFF
--- a/addons/portal/controllers/thread.py
+++ b/addons/portal/controllers/thread.py
@@ -93,8 +93,13 @@ class PortalThreadController(ThreadController):
         }
 
     def _get_non_empty_message_domain(self):
-        return Domain(
-            "body", "not in", [False, '<span class="o-mail-Message-edited"></span>']
+        return (
+            Domain("body", "!=", False)
+            & Domain(
+                "body",
+                "not like",
+                '<span class="o-mail-Message-edited" data-o-datetime="%"></span>',
+            )
         ) | Domain("attachment_ids", "!=", False)
 
     def _setup_portal_message_fetch_extra_domain(self, data) -> Domain:
--- a/addons/website_slides/static/tests/tours/slides_course_review_modification.js
+++ b/addons/website_slides/static/tests/tours/slides_course_review_modification.js
@@ -99,6 +99,10 @@ registry.category("web_tour.tours").add("course_review_modification", {
             trigger:
                 "#chatterRoot:shadow .o-mail-Message:contains(Second review) .o_website_rating_static[title='3 stars on 5']",
         },
+        // If it fails here, it means that empty messages are also being fetched.
+        {
+            trigger: "#chatterRoot:shadow .o-mail-Chatter .o-mail-Message:count(1)",
+        },
         {
             trigger: "span:contains(Edit Review)",
             run: "click",


addons/portal/controllers/thread.py
```diff
@@ -93,9 +93,11 @@ def portal_message_fetch(self, thread_model, thread_id, fetch_params=None, **kw)
         }
 
     def _get_non_empty_message_domain(self):
-        return Domain(
-            "body", "not in", [False, '<span class="o-mail-Message-edited"></span>']
-        ) | Domain("attachment_ids", "!=", False)
+        return (
+            Domain("body", "!=", False)
+            & Domain("body", "not like", '<span class="o-mail-Message-edited"%')
```
seb-odoo: Doesn't this prevent from fetching all edited messages, whether or not they are empty?

maki-odoo: No, sql treats `%` as `starts with` so if the message body starts with anything other than `<span class="o-mail-Message-edited"`, it would be included in the result.

seb-odoo: We should if possible not enforce a specific position. The feature should be working the same whether the span is at the start or at the end.
We want to be checking specifically for only having this span, not its position.
Basically the placeholder should only be there to catch the potential date, but the rest of the span should be checked like before.

addons/portal/controllers/thread.py
```diff
@@ -93,9 +93,12 @@ def portal_message_fetch(self, thread_model, thread_id, fetch_params=None, **kw)
...
     def _get_non_empty_message_domain(self):
-        return Domain(
-            "body", "not in", [False, '<span class="o-mail-Message-edited"></span>']
-        ) | Domain("attachment_ids", "!=", False)
+        return (
+            Domain("body", "!=", False)
+            & Domain("body", "not like",
+                     '<span class="o-mail-Message-edited" data-o-datetime=%></span>')
+            | Domain("attachment_ids", "!=", False)
+        )
```
seb-odoo: I never know about priority of operators, I prefer to put explicit parenthesis when there is a mix of `&` and `|` to avoid confusion.

addons/portal/controllers/thread.py
```diff
@@ -93,9 +93,12 @@ def portal_message_fetch(self, thread_model, thread_id, fetch_params=None, **kw)
         }
 
     def _get_non_empty_message_domain(self):
-        return Domain(
-            "body", "not in", [False, '<span class="o-mail-Message-edited"></span>']
-        ) | Domain("attachment_ids", "!=", False)
+        return (
+            Domain("body", "!=", False)
+            & Domain("body", "not like",
+                     '<span class="o-mail-Message-edited" data-o-datetime=%></span>')
```
seb-odoo: Strange formatting, I'd expect the formatter to have it like
```
            & Domain(
                "body",
                "not like",
                '<span class="o-mail-Message-edited" data-o-datetime=%></span>',
            )
```

seb-odoo: Is there a particular reason to not also include the quotes around the datetime in the string?

seb-odoo (approved):
@robodoo delegate+

PR #247796, opened 2026-02-09
[FIX] account_peppol: Move self-billing import to account_peppol

antoine162:
At the moment, the Peppol AP accepts incoming self-billing invoices for all 18.0 users, but if the `account_peppol_selfbilling` module is not installed on the database, incoming self-billing invoices are not decoded correctly, and end up as as vendor bills rather than sales documents.

This is because the import logic for self-billing invoices is in `account_peppol_selfbilling` at the moment.

To solve this issue, we move the import logic to `account_peppol`, as well as related tests.

task-none


addons/account_peppol/models/account_edi_proxy_user.py
```diff
@@ -157,10 +179,6 @@ def _peppol_import_invoice(self, attachment, partner_endpoint, peppol_state, uui
         attachment.write({'res_model': 'account.move', 'res_id': move.id})
         return True
 
-    def _peppol_get_import_journal_and_move_type(self, attachment):
```
gawa-odoo, line 160: I know we have few reason to keep it, but we should avoid deleting method that somehow someone could override
Maybe just put a deprecated instead

I know it's quite recent, but I think it's not a huge bloat

gawa-odoo (commented):
Just a mini question about the removal of this method. 
Except that, LGTM

addons/account_peppol/models/account_edi_proxy_user.py
antoine162: Ah that's right, so I've just kept the method. :smile:

PR #179354, opened 2024-09-05
Rename groups_id into group_ids and all_group_ids

Gorash:
The  `groups_id` from `res.groups` fields have been removed and replaced by `group_ids` and `all_group_ids`.
The `users` from `res.groups` fields have been removed and replaced by `user_ids` and `all_user_ids`.

The changes reduce also the number of queries when modifying a user's groups (base: 16 -> 8, all: 28 -> 21).

task: 4341594 & 4166361

Models:
```
<res.groups>:
  implied_ids        = Many2many <res.groups> Users of this group are also implicitly part of these inherited groups (db table 'res_groups_implied_rel')
  all_implied_ids    = Many2many <res.groups> Transitive and recursive closure
  implied_by_ids     = Many2many <res.groups> Automatically added from (db table 'res_groups_implied_rel')
  all_implied_by_ids = Many2many <res.groups> Reversed transitive and recursive closure

  user_ids           = Many2many <res.users>  Users with this group (db table: 'res_groups_users_rel')
  all_user_ids       = Many2many <res.users>  Users and implied users

<res.users>:
  group_ids          = Many2many <res.groups> Group added directly to this user (db table: 'res_groups_users_rel')
  all_group_ids      = Many2many <res.groups> Groups and implied groups
```

Concept:
```
The groups involved are to be interpreted as sets.
Thus we can define groups that we will call for example N, Z... such as mathematical sets.
        ┌──────────────────────────────────────────┐
        │ C  ┌──────────────────────────┐          │
        │    │ R  ┌───────────────────┐ │ ┌──────┐ |   "C"
        │    │    │ Q  ┌────────────┐ │ │ │ I    | |   "I" implied "C"
        │    │    │    │ Z  ┌─────┐ │ │ │ │      | |   "R" implied "C"
        │    │    │    │    │ N   │ │ │ │ │      │ │   "Q" implied "R"
        │    │    │    │    └─────┘ │ │ │ │      │ │   "R\Q" implied "R"
        │    │    │    └────────────┘ │ │ │      │ │   "Z" implied "Q"
        │    │    └───────────────────┘ │ │      │ │   "N" implied "Z"
        │    │      ┌───────────────┐   │ │      │ │
        │    │      │ R\Q           │   │ │      │ │
        │    │      └───────────────┘   │ └──────┘ │
        │    └──────────────────────────┘          │
        └──────────────────────────────────────────┘
    For example:
    * A manager group will imply a user group: all managers are users (like Z imply C);
[...]

PR #247802, opened 2026-02-09
[FIX] web: do not allow edition of user group field in readonly

aab-odoo:
Before this commit, the res_user_group_ids field introduced in [1] didn't care about the `readonly` props. As a consequence, when the field (or the whole view, via `edit="0"`) was readonly, the widget still rendered editable SelectMenu. Obviously, editing it and then triggering a save would raise a validation error, so it was only an UI issue.

This commit fixes it by properly setting the field in readonly if its props states it.

[1] https://github.com/odoo/odoo/pull/179354

task~5922282

Current behavior before PR:

---


juliusc2066 (approved):
@robodoo rebase-ff delegate+ r+

PR #247804, opened 2026-02-09
[IMP] sale_project: Allow projects on draft SO

VincentJanssen-Code:
## Behavior Before Commit:
- When a user selected "Create Project" on an SO that was not confirmed or already had a project, Odoo did not create the project and showed a warning.

## Expected Behavior After Commit:
- Users can create multiple projects for the same SO.
- Users can create a project regardless of the SO state.
- Project SOLs remain empty when the SO is in draft.
- Project SOLs are auto-filled when the SO is confirmed.
- SO are no longer auto confirmed once the are connected to a project to make it possible for a draft SO tro be attached to a project.

## Task
[5872486](https://www.odoo.com/odoo/project/4105/tasks/5872486)

---

### DIFF
--- a/addons/sale_project/models/sale_order.py
+++ b/addons/sale_project/models/sale_order.py
@@ -52,18 +52,9 @@ class SaleOrder(models.Model):
 
     def _compute_show_project_and_task_button(self):
         is_project_manager = self.env.user.has_group('project.group_project_manager')
-        show_button_ids = self.env['sale.order.line']._read_group([
-            ('order_id', 'in', self.ids),
-            ('order_id.state', 'not in', ['draft', 'sent']),
-        ], aggregates=['order_id:array_agg'])[0][0]
         for order in self:
-            state = order.state not in ['draft', 'sent']
-            order.show_project_button = state and order.project_count
-            order.show_create_project_button = (
-                is_project_manager
-                and order.id in show_button_ids
-                and not order.project_count
-            )
+            order.show_project_button = order.project_count
+            order.show_create_project_button = is_project_manager
 
     @api.model
     def _search_tasks_ids(self, operator, value):
@@ -145,15 +136,22 @@ class SaleOrder(models.Model):
             for order in self:
                 order.order_line.sudo().with_company(order.company_id)._timesheet_service_generation()
 
-        # If the order has exactly one project and that project comes from a template, set the company of the template
-        # on the project.
         for order in self.sudo(): # Salesman may not have access to projects
+            sale_line_to_assign = next((sol for sol in order.order_line if sol.is_service), False)
+            if not sale_line_to_assign:
+                continue
+            for project in order.project_ids.sudo():
+                if not project.sale_line_id and project.reinvoiced_sale_order_id:
+                    project.sale_line_id = sale_line_to_assign
+            # If the order has exactly one project and that project comes from a template, set the company of the template
+            # on the project.
             if len(order.project_ids) == 1:
                 project = order.project_ids[0]
                 for sol in order.order_line:
                     if project == sol.project_id and (project_template := sol.product_template_id.project_template_id):
                         project.sudo().company_id = project_template.sudo().company_id
                         break
+
         return super()._action_confirm()
 
     def _tasks_ids_domain(self):
@@ -167,10 +165,9 @@ class SaleOrder(models.Model):
                 'tag': 'display_notification',
                 'params': {
                     'type': 'danger',
-                    'message': self.env._("The project couldn't be created as the Sales Order must be confirmed or is already linked to a project."),
+                    'message': self.env._("The project couldn't be created because you don't have the right to create a project"),
                 }
             }
-
         sorted_line = self.order_line.sorted('sequence')
         default_sale_line = next((
             sol for sol in sorted_line
--- a/addons/sale_project/models/sale_order_line.py
+++ b/addons/sale_project/models/sale_order_line.py
@@ -175,16 +175,18 @@ class SaleOrderLine(models.Model):
     def _timesheet_create_project_prepare_values(self):
         """Generate project values"""
         # create the project or duplicate one
-        return {
+        values = {
             'name': '%s - %s' % (self.order_id.client_order_ref, self.order_id.name) if self.order_id.client_order_ref else self.order_id.name,
             'account_id': self.env.context.get('project_account_id') or self.order_id.project_account_id.id or self.env['account.analytic.account'].create(self.order_id._prepare_analytic_account_data()).id,
             'partner_id': self.order_id.partner_id.id,
-            'sale_line_id': self.id,
             'active': True,
             'company_id': self.company_id.id,
             'allow_billable': True,
             'user_id': self.product_id.project_template_id.user_id.id,
         }
+        if self.order_id.state != 'draft':
+            values['sale_line_id'] = self.id
+        return values
 
     def _timesheet_create_project(self):
         """ Generate project for the given so line, and link it.
@@ -200,15 +202,16 @@ class SaleOrderLine(models.Model):
                 project = project_template.action_create_from_template(values)
             else:
                 project = project_template.copy(values)
-            project.tasks.write({
-                'sale_line_id': self.id,
-                'partner_id': self.order_id.partner_id.id,
-            })
-            # duplicating a project doesn't set the SO on sub-tasks
-            project.tasks.filtered('parent_id').write({
-                'sale_line_id': self.id,
-                'sale_order_id': self.order_id.id,
-            })
+            if (self.order_id.state != 'draft'):
+                project.tasks.write({
+                    'sale_line_id': self.id,
+                    'partner_id': self.order_id.partner_id.id,
+                })
+                # duplicating a project doesn't set the SO on sub-tasks
+                project.tasks.filtered('parent_id').write({
+                    'sale_line_id': self.id,
+                    'sale_order_id': self.order_id.id,
+                })
         else:
             project_only_sol_count = self.env['sale.order.line'].search_count([
                 ('order_id', '=', self.order_id.id),
--- a/addons/sale_project/tests/test_sale_project.py
+++ b/addons/sale_project/tests/test_sale_project.py
@@ -387,7 +387,6 @@ class TestSaleProject(TestSaleProjectCommon):
         # add a project to the SO
         line_delivered_milestone.project_id = self.project_global
         sale_order_1._compute_show_project_and_task_button()
-        self.assertFalse(sale_order_1.show_create_project_button, "There is a product service with the service_policy set on 'delivered on milestone' and a project on the sale order, the button should be hidden")
         self.assertTrue(sale_order_1.show_project_button, "There is a product service with the service_policy set on 'delivered on milestone' and a project on the sale order, the button should be displayed")
 
         # add an ordered_prepaid service product
@@ -1097,7 +1096,7 @@ class TestSaleProject(TestSaleProjectCommon):
             copy.order_line.project_id._get_analytic_distribution(),
         )
 
-    def test_confirm_sale_order_on_task_save(self):
+    def test_non_confirm_sale_order_on_task_save(self):
         sale_order = self.env['sale.order'].create({
             'name': 'Sale Order',
             'partner_id': self.partner.id,
@@ -1113,9 +1112,9 @@ class TestSaleProject(TestSaleProjectCommon):
             'project_id': self.project_global.id,
         })
         task.write({'sale_line_id': sale_order_line.id})
-        self.assertEqual(sale_order.state, 'sale')
+        self.assertEqual(sale_order.state, 'draft')
 
-    def test_confirm_sale_order_on_project_creation(self):
+    def test_non_confirm_sale_order_on_project_creation(self):
         sale_order = self.env['sale.order'].create({
             'name': 'Sale Order',
             'partner_id': self.partner.id,
@@ -1130,7 +1129,7 @@ class TestSaleProject(TestSaleProjectCommon):
             'name': 'Project',
             'sale_line_id': sale_order_line.id,
         })
-        self.assertEqual(sale_order.state, 'sale')
+        self.assertEqual(sale_order.state, 'draft')
 
     def test_create_project_on_fly(self):
         """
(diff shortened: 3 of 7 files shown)


addons/sale_project/models/sale_order.py
```diff
@@ -133,10 +124,20 @@ def _compute_project_ids(self):
             order.project_count = len(projects.filtered('active'))
 
     def _action_confirm(self):
+        """ On SO confirmation, some lines should generate a task or a project. """
+        self._generate_template()
+        for record in self:
+            for lines in record.order_line:
+                if (not lines.project_id.sale_line_id):
+                    lines.project_id.sale_line_id = lines
```
deneuvillem: I don't get why we need to do that. Isn't it already handled here? https://github.com/odoo/odoo/blob/master/addons/sale_project/models/sale_order_line.py#L180
-> The link is done when creating a new project from a SOL (when confirming the SO)

addons/sale_project/models/sale_order.py
```diff
@@ -154,7 +155,7 @@ def _action_confirm(self):
                     if project == sol.project_id and (project_template := sol.product_template_id.project_template_id):
                         project.sudo().company_id = project_template.sudo().company_id
                         break
-        return super()._action_confirm()
+        return len(self.sudo().project_ids) > project_count
```
deneuvillem: You want to compare the number of projects initially vs this number after the generation process, why so?

deneuvillem: If you really need to achieve that, let's use variables to be cleaner (e.g. `initial_project_count`, `final_project_count`)

addons/sale_project/models/sale_order.py
```diff
@@ -133,10 +124,20 @@ def _compute_project_ids(self):
...
+            for lines in record.order_line:
+                if (not lines.project_id.sale_line_id):
+                    lines.project_id.sale_line_id = lines
+        return super()._action_confirm()
+
+    def _generate_template(self):
         """ On SO confirmation, some lines should generate a task or a project. """
         if self.env.context.get('disable_project_task_generation'):
-            return super()._action_confirm()
+            return False
```
deneuvillem: You should keep that logic in `_action_confirm`, as we still need to call the super() if the condition matches

addons/sale_project/models/sale_order.py
```diff
@@ -167,9 +168,12 @@ def action_create_project(self):
                 'tag': 'display_notification',
                 'params': {
                     'type': 'danger',
-                    'message': self.env._("The project couldn't be created as the Sales Order must be confirmed or is already linked to a project."),
+                    'message': self.env._("The project couldn't be created because you don't have the right to creat a project"),
```
deneuvillem:
```suggestion
                    'message': self.env._("The project couldn't be created because you don't have the right to create a project"),
```

addons/sale_project/models/sale_order.py
```diff
@@ -167,9 +168,12 @@ def action_create_project(self):
...
                 }
             }
+        # generates projects based on the pre-existing SOLs befor prompting for template selection
+        if (self._generate_template()):
+            return
```
deneuvillem: I don't get why adding a new project from that server action should trigger the process to generate all projects/tasks/etc linked to all SOLs; Also why do we return nothing?

deneuvillem (commented):
Hello @VincentJanssen-Code and thanks for your work :)

I just have a concern about the use of `_generate_template` in `action_create_project`, I don't really understand why we do that.

Also, I think we can handle the logic of  this point of the spec: 

> The SOL should be set automatically once the related SO is confirmed

in the `write()` method of `sale.order` (overridden in `sale_project` module), which would check that the state has changed from `'draft'` to `'sale'`, and then do your logic. Another idea would be to move that logic to the `_compute_sale_line_id()` of `project.project` directly, but this may be less flexible.

Also, you have a failing test on the runbot :0

addons/sale_project/models/sale_order.py
VincentJanssen-Code: This line is necessary because the link between the project and its SOL is not always created at the same time as the project itself. The spec states the following on this point:

"if the SO is still a quote, set the SO field on the project but leave the SOL empty. The SOL should be set automatically once the related SO is confirmed"

Because of this, projects may be created without an SOL, and they need to have their SOL assigned once the SO is confirmed.

Unfortunately, I attempted to move the code that assigns the SOL of empty projects from the `_action_confirm` function into the `write` function, and this change introduced several issues. The `write` call that updates the status from draft to sold occurs in the middle of the `_action_confirm` execution. At that moment, the projects that should be generated from the SOLs are not yet fully created. As a result, both the projects and the tasks generated from the SOL end up malformed.

Here is a concrete example:

1. A user confirms an SO containing an SOL that should generate only a task.
2. A global project is created to hold the task generated from the SOL.
3. The `write` function is triggered, and we assign the SOL to the global project because it does not yet have one.
4. When the SOL’s task is generated, its name follows this logic:
   - If the project already has an SOL, the task name should be the first line of the SOL description.
   - If the project does not have an SOL, the task name should be the SO number followed by the SOL description.

Previously, the SOL-generated task would include the SO number in its name. After moving the code, the task can no longer include the SO number, because the SOL is prematurely assigned to the project. While this may seem like a minor side effect, it breaks several Odoo tests, including:

- TestSaleProject.test_sale_order_with_project_task (line 165)
- TestSaleService.test_sale_service (line 39)

Keeping the code at the end of `_action_confirm` inside the SO model prevents this naming issue and ensures that all tests pass.

addons/sale_project/models/sale_order.py
```diff
@@ -145,15 +137,23 @@ def _action_confirm(self):
...
-        # If the order has exactly one project and that project comes from a template, set the company of the template
-        # on the project.
         for order in self.sudo(): # Salesman may not have access to projects
+            try:
+                sale_line_to_assign = order.get_first_service_line()
+            except (UserError):
+                sale_line_to_assign = False
+            project_with_sale_line = order.project_ids.filtered(lambda p: p.sale_line_id)
+            for project_id in order.project_ids.filtered(lambda p: p.sale_order_id) - project_with_sale_line:
+                project_id.sudo().sale_line_id = sale_line_to_assign
```
deneuvillem:
```suggestion
             sale_line_to_assign = next((sol for sol in self.order_line if sol.is_service), False)
             if not sale_line_to_assign:
                continue
            for project in order.project_ids.sudo():
                if project.sale_order_id != order or project.sale_line_id:
                    continue
                project.sale_line_id = sale_line_to_assign
```
This way, we avoid doing a `try-catch` (which is not the best practice IMO but understandable), and reduce the number of loops that we do (instead of using `filtered()` multiple times, we do the check in a `for` loop).

addons/sale_project/models/sale_order.py
```diff
@@ -284,9 +283,10 @@ def create(self, vals_list):
 
     def write(self, vals):
         res = super().write(vals)
-        if 'state' in vals and vals['state'] == 'cancel':
-            # Remove sale line field reference from all projects
-            self.env['project.project'].sudo().search([('sale_line_id.order_id', 'in', self.ids)]).sale_line_id = False
+        if 'state' in vals:
+            if vals['state'] == 'cancel':
+                # Remove sale line field reference from all projects
+                self.env['project.project'].sudo().search([('sale_line_id.order_id', 'in', self.ids)]).sale_line_id = False
```
deneuvillem: Isn't it the same logic as the previous code?

addons/sale_project/models/sale_order_line.py
```diff
@@ -175,16 +175,18 @@ def _convert_qty_company_hours(self, dest_company):
     def _timesheet_create_project_prepare_values(self):
         """Generate project values"""
         # create the project or duplicate one
-        return {
+        ret = {
```
deneuvillem:
```suggestion
        values = {
```

addons/sale_project/models/sale_order.py
deneuvillem: Note that a search may be more efficient instead of looping on the `project_ids`, but let's keep it like that for the moment

addons/sale_project/models/sale_order_line.py
```diff
@@ -200,15 +202,16 @@ def _timesheet_create_project(self):
                 project = project_template.action_create_from_template(values)
             else:
                 project = project_template.copy(values)
-            project.tasks.write({
-                'sale_line_id': self.id,
-                'partner_id': self.order_id.partner_id.id,
-            })
+            if (self.order_id.state == 'sent'):
```
deneuvillem: Shouldn't it be `state != 'draft'` as you did before ?

addons/sale_project/tests/test_sale_project.py
```diff
@@ -387,7 +387,8 @@ def test_compute_project_and_task_button(self):
         # add a project to the SO
         line_delivered_milestone.project_id = self.project_global
         sale_order_1._compute_show_project_and_task_button()
-        self.assertFalse(sale_order_1.show_create_project_button, "There is a product service with the service_policy set on 'delivered on milestone' and a project on the sale order, the button should be hidden")
+        # A project can always be made reguardless of the state of the sale order as long as the user has the acess rigths to project
+        # self.assertFalse(sale_order_1.show_create_project_button, "There is a product service with the service_policy set on 'delivered on milestone' and a project on the sale order, the button should be hidden")
```
deneuvillem: I guess we can safely remove that check then (and the two lines aboves)

addons/sale_project/tests/test_sale_project.py
```diff
@@ -1641,8 +1642,8 @@ def test_sale_order_project_task_smartbutton(self):
             - Create a project and link it to the sale order.
             - Create a task and link it to the project.
             - Verify that the project is linked to the sale order.
-            - Verify that the task without SO item is not linked to the sale order.
-            - Verify that the project & task smart button is hidden.
+            - Verify that the tasks is linked to the sale order.
```
deneuvillem:
```suggestion
            - Verify that the task is linked to the sale order.
```

deneuvillem (commented):
Thanks for the changes! I left some more comments for you 👀

addons/sale_project/models/sale_order.py
deneuvillem: Oki, make sense! Thanks for your detailed explanation :)

addons/sale_project/models/sale_order.py
```diff
@@ -145,15 +137,22 @@ def _action_confirm(self):
...
 
-        # If the order has exactly one project and that project comes from a template, set the company of the template
-        # on the project.
         for order in self.sudo(): # Salesman may not have access to projects
+            sale_line_to_assign = next((sol for sol in self.order_line if sol.is_service), False)
+            if not sale_line_to_assign:
+                continue
+            for project in order.project_ids.sudo():
+                if not project.sale_line_id and project.sale_order_id:
+                    project.sudo().sale_line_id = sale_line_to_assign
```
deneuvillem: Do we need to `sudo()` ? As you already do it above

addons/sale_project/models/sale_order.py
deneuvillem: Let's keep the previous code to avoid useless diff :)

deneuvillem (commented):
Left 2 remarks for you :)

addons/sale_project/models/sale_order.py
```diff
@@ -145,15 +137,23 @@ def _action_confirm(self):
             for order in self:
                 order.order_line.sudo().with_company(order.company_id)._timesheet_service_generation()
 
-        # If the order has exactly one project and that project comes from a template, set the company of the template
-        # on the project.
         for order in self.sudo(): # Salesman may not have access to projects
+            sale_line_to_assign = next((sol for sol in self.order_line if sol.is_service), False)
```
xavierbol:
```suggestion
            sale_line_to_assign = next((sol for sol in order.order_line if sol.is_service), False)
```
? 🤔

addons/sale_project/models/sale_order.py
```diff
@@ -145,15 +137,23 @@ def _action_confirm(self):
...
+            if not sale_line_to_assign:
+                continue
+            for project in order.project_ids.sudo():
+                if not project.sale_line_id and project.reinvoiced_sale_order_id:
+                    project.sale_line_id = sale_line_to_assign
+            # If the order has exactly one project and that project comes from a template, set the company of the template
+            # on the project.
+        for order in self.sudo():
```
xavierbol: why not doing everything in the same for loop? 🧐

addons/sale_project/models/sale_order_line.py
```diff
@@ -200,15 +202,16 @@ def _timesheet_create_project(self):
...
-            project.tasks.write({
-                'sale_line_id': self.id,
-                'partner_id': self.order_id.partner_id.id,
-            })
+            if (self.order_id.state != 'draft'):
+                project.tasks.write({
+                    'sale_line_id': self.id,
+                    'partner_id': self.order_id.partner_id.id,
+                })
             # duplicating a project doesn't set the SO on sub-tasks
```
xavierbol: you can also add a tabulation for that comment. 😇

xavierbol (commented):

addons/sale_project/models/sale_order.py
```diff
@@ -133,10 +124,20 @@ def _compute_project_ids(self):
             order.project_count = len(projects.filtered('active'))
 
     def _action_confirm(self):
+        """ On SO confirmation, some lines should generate a task or a project. """
+        self._generate_template()
+        for record in self:
+            for lines in record.order_line:
+                if (not lines.project_id.sale_line_id):
```
xavierbol: We should first check if project exists before checking if the project has SOL.

addons/sale_project/models/sale_order.py
```diff
@@ -133,10 +124,20 @@ def _compute_project_ids(self):
...
+                    lines.project_id.sale_line_id = lines
+        return super()._action_confirm()
+
+    def _generate_template(self):
```
xavierbol: I don't understand why you create a new method. Also, we will not really generate a template, we will generate a project/task based on the settings of the products linked to SOLs. 🤔 

So, IMO, no need to create a new method everything can be done in `_action_confirm`. 🙂

addons/sale_project/models/sale_order_line.py
xavierbol:
```suggestion
        action = {
```
🤷🏼‍♂️

addons/sale_project/models/sale_order_line.py
xavierbol: why should we only do that when the SO is sent? 🤔 

Before your changes, we always update the tasks of the project. 🤔

xavierbol (commented):
Thanks for your work, I left some comments!

xavierbol:
Ah zut, I will redo my review and I will just check all commits together instead of doing the review commit per commit

xavierbol:
LGTM, thanks for your work!

robodoo r+

PR #247805, opened 2026-02-09
[FIX] mass_mailing: unlink mail.message of test mails

delcourtfl:
**Steps to reproduce:**
- Go to Email Marketing app
- Create a mailing campaign
- Set its recipients to Contact
- Click on the test button to send a test mail to any mail
- Go to the first contact record
- Chatter will show `This message has been removed` message

**Issue:**
Previously, message created for testing were ignored by the Chatter as they were empty. As we now keep empty messages visible but with removed content display, they shows up on related records.
```py
record = self.env[mailing.mailing_model_real].search([], limit=1)
```

**Fix:**
Ensure the related messages are unlinked at the same time as the test mail in `send_mail_test` by setting `is_notification` to `False` to trigger the `unlink` logic.
```py
def unlink(self):
    # cascade-delete the parent message for all mails that are not created for a notification
    mail_msg_cascade_ids = [mail.mail_message_id.id for mail in self if not mail.is_notification]
    res = super(MailMail, self).unlink()
    if mail_msg_cascade_ids:
        self.env['mail.message'].browse(mail_msg_cascade_ids).unlink()
    return res
```

related: https://github.com/odoo/odoo/commit/21f92550f83cbd38df2c223c65c61bd16dc8e2b0

opw-5502787

### DIFF
--- a/addons/test_mass_mailing/tests/test_mailing_test.py
+++ b/addons/test_mass_mailing/tests/test_mailing_test.py
@@ -160,6 +160,21 @@ class TestMailingTest(TestMassMailCommon):
             "Should use the value of the previous record's email_to as default",
         )
 
+        # Also test that related messages were properly deleted
+        mailing.subject = 'Dummy Subject'
+
+        with self.mock_mail_gateway(mail_unlink_sent=True):
+            mailing_test.send_mail_test()
+
+        test_subject = '[TEST] %s' % mailing.subject
+        self.assertSentEmail(
+            self.env.user.partner_id,
+            ['test@test.com'],
+            subject=test_subject,
+        )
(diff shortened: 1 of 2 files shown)


Ephoniz (approved):
LGTM 👍🏽

addons/test_mass_mailing/tests/test_mailing_test.py
```diff
@@ -160,6 +160,26 @@ def test_mailing_test_equals_reality(self):
             "Should use the value of the previous record's email_to as default",
         )
 
+    def test_mailing_test_remove_related_messages(self):
```
tde-banana-odoo: Can't it be included in a previous test so that we don't have to define yet another setup and unit test ?

PR #247806, opened 2026-02-09
[FIX] digest: Count users based on allowed companies

dido-odoo:
**Problem:**
Currently, the digest KPI for connected users checks the "company_id" field (as with all other models), but this field corresponds to "Default Company" on res.users, meaning a user can only be considered for one company when computing the digest KPI. This can cause misleading digest KPIs if users work in multiple companies, or mainly in a company that isn't their default company.

**Solution:**
Instead of always using the "company_id" field, we use the "company_ids" field if present on the model.

opw-5404940

### DIFF
--- a/addons/digest/models/digest.py
+++ b/addons/digest/models/digest.py
@@ -410,8 +410,10 @@ class Digest(models.Model):
         """
         start, end, companies = self._get_kpi_compute_parameters()
 
+        company_field = self._get_company_field(model)
+
         base_domain = [
-            ('company_id', 'in', companies.ids),
+            (company_field, 'in', companies.ids),
             (date_field, '>=', start),
             (date_field, '<', end),
         ]
@@ -421,7 +423,7 @@ class Digest(models.Model):
 
         values = self.env[model]._read_group(
             domain=base_domain,
-            groupby=['company_id'],
+            groupby=[company_field],
             aggregates=[f'{sum_field}:sum'] if sum_field else ['__count'],
         )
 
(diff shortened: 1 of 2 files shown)


addons/digest/models/digest.py
```diff
@@ -421,7 +423,7 @@ def _calculate_company_based_kpi(self, model, digest_kpi_field, date_field='crea
 
         values = self.env[model]._read_group(
             domain=base_domain,
-            groupby=['company_id'],
+            groupby=[company_field],
```
reth-odoo, line 426: If you group by company_ids it will make groups like {[1,2,3]: record} right? So you probably want to unwrap that in values_per_company

I guess the test might pass currently because it's only keeping the last value in the dict, which happens to be company 2

addons/digest/models/digest.py
```diff
@@ -410,8 +410,10 @@ def _calculate_company_based_kpi(self, model, digest_kpi_field, date_field='crea
         """
         start, end, companies = self._get_kpi_compute_parameters()
 
+        company_field = 'company_ids' if 'company_ids' in self.env[model]._fields else 'company_id'
```
reth-odoo: Maybe in some external modules it does not make sense to start using company_ids over company_id. Change should be more targeted imo, since this targets old stable. Adding a field getter or such.

addons/digest/models/digest.py
dido-odoo: Checking with a debugger, it looks like the keys are individual companies even if a user belongs to multiple companies:
`[(res.company(1,), 2), (res.company(2,), 1), (res.company(3,), 1), (res.company(4,), 1)]`

So this should be fine (at least in 17.0), if it's done differently in later versions the tests will break, I can fix the forward ports in that case.

addons/digest/models/digest.py
dido-odoo: I updated the commit so that only res.users will use the company_ids field, let me know if the fix looks good.

PR #247807, opened 2026-02-09
[REF] html_builder: use resource to tell which snippets connect shape

blse-odoo:
Commit e591ea77f4a800bf79cb19f80529eed0d8ded927 added the recomputation of shape colors based on background of sibling snippets.

This computation needs to ignore (among other things) the snippets that can be invisible. And used a new shared function of the visibility plugin.

This commit instead uses a resource, as we are planning to move the visibility plugin to the website module; and implements the loop to search for the sibling only once.

task-5149984

### DIFF
--- a/addons/html_builder/static/src/plugins/background_option/background_shape_option_plugin.js
+++ b/addons/html_builder/static/src/plugins/background_option/background_shape_option_plugin.js
@@ -26,11 +26,11 @@ import { selectElements } from "@html_editor/utils/dom_traversal";
  * }>} BackgroundShapeGroups
  * @typedef {((shapeGroups: BackgroundShapeGroups) => BackgroundShapeGroups | void)[]} background_shape_groups_providers
  * @typedef {((editingElement: HTMLElement) => HTMLElement)[]} background_shape_target_providers
+ * @typedef {((el: HTMLElement) => boolean)[]} is_element_in_invisible_panel_predicates
  */
 
 export class BackgroundShapeOptionPlugin extends Plugin {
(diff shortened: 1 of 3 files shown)


addons/html_builder/static/src/plugins/background_option/background_shape_option_plugin.js
```diff
@@ -133,7 +133,10 @@ export class BackgroundShapeOptionPlugin extends Plugin {
             return;
         }
         const { originNextEl, originPreviousEl } = dragState;
-        const neighborShapeEls = this.getNeighborShapeEls(originPreviousEl, originNextEl);
+        const neighborShapeEls = this.getNeighborShapeEls({
```
loco-odoo: Here you give an object as parameter for the `getNeighborShapeEls` method but the next call is giving an html element

addons/html_builder/static/src/plugins/background_option/background_shape_option_plugin.js
```diff
@@ -159,7 +162,10 @@ export class BackgroundShapeOptionPlugin extends Plugin {
      */
     updateConnectionShapeColor(editingEl) {
         const { shape, colors } = this.getShapeData(editingEl);
-        if (!this.isShapeEligibleForComputation(shape, editingEl)) {
+        if (
+            !this.isShapeEligibleForComputation(shape, editingEl) ||
+            !editingEl.querySelector(":scope > .o_we_shape")
```
loco-odoo: Why? Isn't it already taken into account in `isShapeEligibleForComputation` ?

PR #247808, opened 2026-02-09
[FIX] mass_mailing: make background color responsive to picker

Alpha9584:
The `o_mail_wrapper_td` element in the `ThemeWrapper` template had a hardcoded `bg-white` class. The intent was for the wrapper background to default to white, but doing it this way meant the color didn't participate in the CSS variable system. Whenever a user set a custom Content Background color and a border-radius on a text block, the corners would reveal white underneath instead of the actual Content Background color.

Removed the `bg-white` class from the template and set white as the proper default for the `--wrapper-background-color` variable instead, so the wrapper is still white by default but correctly responds to the Content Background color picker.

Steps to reproduce:
1. Create a new Mailing
2. Set the Content Background color to something other than white
3. Add a Text block
4. Set a border-radius value on the block

=> white shows behind the rounded corners instead of the Content Background

Ticket [link](https://www.odoo.com/odoo/project.task/5868955) 
opw-5868955

---

### DIFF
--- a/addons/mass_mailing/static/src/themes/theme_wrapper.xml
+++ b/addons/mass_mailing/static/src/themes/theme_wrapper.xml
@@ -9,7 +9,7 @@
     <t t-name="mass_mailing.ThemeWrapper">
         <div class="container o_mail_wrapper o_mail_regular oe_unremovable">
             <div class="row mw-100 mx-0">
-                <div class="col o_mail_no_options o_mail_wrapper_td bg-white oe_structure" t-out="html"/>
+                <div class="col o_mail_no_options o_mail_wrapper_td oe_structure" t-out="html"/>
             </div>
         </div>
     </t>


addons/mass_mailing/static/src/builder/plugins/customize_mailing_variables.js
```diff
@@ -104,7 +104,7 @@ export const CUSTOMIZE_MAILING_VARIABLES = Object.assign(
 
 export const CUSTOMIZE_MAILING_VARIABLES_DEFAULTS = {
     "--wrapper-background-color": {
-        "background-color": "#FFFFFF",
+        "background-color": "rgb(255, 255, 255)",
```
ushyme, line 107: I'm not sure I understand what this change does, why do we need to switch from hex to rgb ?

ushyme (commented):
Thanks for your work and sorry for the long delay. I just have one small question.

addons/mass_mailing/static/src/builder/plugins/customize_mailing_variables.js
Alpha9584: Hello, I did not change it, background-color was empty when I made this PR. The hex was put by another change. 
You can see here the original diff: https://github.com/odoo/odoo/commit/db5f67a2da5fa2af1ae3aa842d1dae839c8de9e2

Tho tbh I think we should change it to rgb regardless since, the whole file already uses that instead of hex

ushyme: Thanks for the quick reply.
I would leave it in hex format, if only to preserve history and avoid unnecessary changes

Alpha9584: Fair enough, I undid the change.

ushyme (commented):
LGTM, thanks!

abd-msyukyu-odoo:
@robodoo r+

PR #247812, opened 2026-02-09
[IMP] l10n_pl_edi: KSeF Download Bills

abmn-odoo:
If a company sell me something, they can send me an invoice. With KSeF, they do not have to send an invoice to me by post or mail - they can send it directly to KSeF. And it is legally delivered to me.

Implement fetching bills from KSeF

task-5405257

---

### DIFF
--- a/addons/l10n_pl_edi/data/ir_cron_data.xml
+++ b/addons/l10n_pl_edi/data/ir_cron_data.xml
@@ -12,5 +12,15 @@
             <field name="interval_type">weeks</field>
             <field name="nextcall" eval="(DateTime.now() + timedelta(days=1)).strftime('%Y-%m-%d 22:00:00')"/>
         </record>
+
+        <record id="cron_l10n_pl_edi_ksef_download_bills" model="ir.cron">
+            <field name="name">Polish eInvoice: Download vendor bills from KSeF</field>
+            <field name="model_id" ref="l10n_pl_edi.model_account_move"/>
+            <field name="state">code</field>
+            <field name="code">model._cron_l10n_pl_edi_download_bills()</field>
+            <field name="user_id" ref="base.user_root"/>
+            <field name="interval_number">3</field>
+            <field name="interval_type">hours</field>
+        </record>
     </data>
 </odoo>
--- a/addons/l10n_pl_edi/models/account_move.py
+++ b/addons/l10n_pl_edi/models/account_move.py
@@ -2,9 +2,11 @@ import logging
 import re
 from xml.dom.minidom import parseString
 
+from dateutil.relativedelta import relativedelta
+from lxml import etree
 from stdnum.pl.nip import compact
 
-from odoo import api, fields, models
+from odoo import Command, api, fields, models
 from odoo.exceptions import UserError
 from odoo.tools import float_compare, float_is_zero, float_repr, OrderedSet
 
@@ -44,6 +46,10 @@ class AccountMove(models.Model):
         compute=lambda self: self._compute_linked_attachment_id('l10n_pl_edi_upo_id', 'l10n_pl_edi_upo_file'),
         depends=['l10n_pl_edi_upo_file'],
     )
+
+    _sql_constraints = [
+        ('unique_l10n_pl_edi_number', 'UNIQUE(l10n_pl_edi_number)', 'The KSeF number must be unique'),
+    ]
 
     def _l10n_pl_edi_check_mandatory_fields(self):
         errors = {}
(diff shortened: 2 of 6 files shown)


addons/l10n_pl_edi/data/ir_cron_data.xml
```diff
@@ -12,5 +12,16 @@
             <field name="interval_type">weeks</field>
             <field name="nextcall" eval="(DateTime.now() + timedelta(days=1)).strftime('%Y-%m-%d 22:00:00')"/>
         </record>
+
+        <record id="cron_l10n_pl_edi_ksef_download_bills" model="ir.cron">
+            <field name="name">Polish eInvoice: Download vendor bills from KSeF</field>
+            <field name="model_id" ref="l10n_pl_edi.model_account_move"/>
+            <field name="state">code</field>
+            <field name="code">model._cron_l10n_pl_edi_download_bills()</field>
+            <field name="active" eval="True"/>
```
Megaaaaaa: True is the default value for active :)

```suggestion
```

addons/l10n_pl_edi/tests/test_l10n_pl_edi.py
```diff
@@ -486,3 +489,135 @@ def send_invoice_raise(xml_content):
...
+
+        bill_1 = self.env['account.move'].search([('l10n_pl_edi_number', '=', 'KSEF-BILL-001')])
+        self.assertTrue(bill_1)
+
+        bill_2 = self.env['account.move'].search([('l10n_pl_edi_number', '=', 'KSEF-BILL-002')])
+        self.assertFalse(bill_2)
+
+        self.assertEqual(len(capt.records), cron_runs_before + 1)
+        self.assertGreaterEqual(capt.records.call_at, fields.Datetime.now() + timedelta(seconds=120))
+        self.assertLessEqual(capt.records.call_at, fields.Datetime.now() + timedelta(seconds=240))
```
Megaaaaaa: These two break as soon as `capt.records` contain multiple records, no ?
```suggestion
        self.assertGreaterEqual(capt.records[-1].call_at, fields.Datetime.now() + timedelta(seconds=120))
        self.assertLessEqual(capt.records[-1].call_at, fields.Datetime.now() + timedelta(seconds=240))
```

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -436,3 +442,200 @@ def _compute_show_reset_to_draft_button(self):
...
+                'invoice_date_due': invoice_date_due,
+                'invoice_number': invoice_number,
+                'currency_code': currency_code,
+                'lines': lines,
+            }
+
+        def create_ksef_bill(data):
+            partner = self.env['res.partner'].search(
+                [
+                    ('vat', 'like', data['vendor_nip']),
```
Megaaaaaa: I guess you use `like` to handle both when the country code prefix is present and when it's not but you risk linking the bill to the wrong partner imo, it'd be better to check for exact value and list the values you can expect.

```suggestion
                    ('vat', 'in', [data['vendor_nip'], (data['vendor_country'] or '') + data['vendor_nip']]),
```

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -436,3 +442,200 @@ def _compute_show_reset_to_draft_button(self):
...
+                        Command.set(
+                            self.env['account.tax'].search(
+                                [
+                                    ('name', '=', tax_name),
+                                    ('type_tax_use', '=', 'purchase'),
+                                    *self.env['account.tax']._check_company_domain(self.env.company),
+                                ], limit=1,
+                            ).ids,
+                        ),
+                    )
```
Megaaaaaa: I think I wouldn't rely on the tax being found here because you're relying on the name to find taxes which is not the safest field to search for. Not that you have much choice here 😅 
```suggestion
                    tax = self.env['account.tax'].search(
                        [
                            ('name', '=', tax_name),
                            ('type_tax_use', '=', 'purchase'),
                            *self.env['account.tax']._check_company_domain(self.env.company),
                        ],
                    )
                    if not tax:
                        raise UserError(self.env._("Missing purchase tax '%s' required for KSeF import.", tax_name))
                    tax_ids = [Command.set(tax.ids)]
```

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -436,3 +442,200 @@ def _compute_show_reset_to_draft_button(self):
...
+    def _cron_l10n_pl_edi_download_bills(self):
+        for company in self.env['res.company'].search([('l10n_pl_edi_access_token', '!=', False)]):
+            self.with_company(company)._l10n_pl_edi_download_bills_from_ksef()
+
+    @api.model
+    def _l10n_pl_edi_download_bills_from_ksef(self):
+
+        def handle_download_bills_from_ksef_error(error):
+            cron = self.env.ref('l10n_pl_edi.cron_l10n_pl_edi_ksef_download_bills')
+            if error.get('retry_after'):
```
Megaaaaaa: Raising a `UserError` from cron will mark the job as failed and can generate noisy logs/notifications, even though rate limiting is an expected operational condition. What I would do instead is raise a regular exception only when it’s truly fatal, and for expected throttling return cleanly after rescheduling so the cron stays “green” under normal rate-limit behavior.
```suggestion
            if not delay:= error.get('retry_after'):
                raise UserError(error.get('message'))
            cron._trigger(at=fields.Datetime.now() + relativedelta(seconds=delay))
            cron._add_progress()
```

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -436,3 +442,200 @@ def _compute_show_reset_to_draft_button(self):
...
+            to_process = [invoice for invoice in invoices if invoice['ksefNumber'] not in already_processed]
+
+            for invoice in to_process:
+                response = service.get_invoice_by_ksef_number(invoice['ksefNumber'])
+                if response.get('error'):
+                    handle_download_bills_from_ksef_error(response['error'])
+                bill = self._l10n_pl_edi_create_ksef_bill_from_xml(response['xml_content'])
+                bill.l10n_pl_edi_number = invoice['ksefNumber']
+
+            page_offset += 1
```
Megaaaaaa: You should try to get the `.search()` out of the `while` loop. Here's my quick take on it, haven't tried running it though :smile:


```suggestion
        invoices = []
        while has_more:

            response = service.query_invoice_metadata(query, page_size, page_offset)

            if response.get('error'):
                handle_download_bills_from_ksef_error(response['error'])

            invoices += response['invoices']
            has_more = response['hasMore']
            
            page_offset += 1

        already_processed = set(self.env['account.move'].search([
            ('l10n_pl_edi_number', 'in', [invoice['ksefNumber'] for invoice in invoices])
        ]).mapped('l10n_pl_edi_number'))

        to_process = [invoice for invoice in invoices if invoice['ksefNumber'] not in already_processed]

        for invoice in to_process:
            response = service.get_invoice_by_ksef_number(invoice['ksefNumber'])
            if response.get('error'):
                handle_download_bills_from_ksef_error(response['error'])
            bill = self._l10n_pl_edi_create_ksef_bill_from_xml(response['xml_content'])
            bill.l10n_pl_edi_number = invoice['ksefNumber']

```

Megaaaaaa (commented):
Hello :wave:

Saw the task was high priority and went ahead to do a review before testing :+1:

This looks really good honestly. Just a very few comments on my part.

addons/l10n_pl_edi/models/account_move.py
abmn-odoo: I agree it's not great, but couldn't come up with something better tbh.

addons/l10n_pl_edi/models/account_move.py
abmn-odoo: I understand your idea, but…

Since we process invoices page by page (i.e., we will never process an invoice from page x before processing invoices from page x − 1), and the page size is greater than or equal to the maximum rate limit value, it is impossible for this code to process more than two pages per cron run.

Therefore, your suggested approach would cause us to run the `search` query on significantly more invoices than `get_invoice_by_ksef_number` can handle anyway, which would result in worse overall performance.

abmn-odoo: As discussed offline, we'll go with @Megaaaaaa  suggestion. While the current rate limits make it impossible for the query to run more than twice, we should rely only on the `hasMore` flag and not assume that their rate limits are accurate or that they won’t change.

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -436,3 +442,211 @@ def _compute_show_reset_to_draft_button(self):
...
+
+                tax_ids = []
+                if tax_name := p12_to_tax_map.get(line['tax_name']):
+                    tax = self.env['account.tax'].search(
+                        [
+                            ('name', '=', tax_name),
+                            ('type_tax_use', '=', 'purchase'),
+                            *self.env['account.tax']._check_company_domain(self.env.company),
+                        ], limit=1,
+                    )
```
Megaaaaaa: Okay, sorry for the back and forth but I've actually found a better solution that works even if the tax has been renamed. As you know precisely which tax you want from the value of the node, you can have a mapping of `{<value_in_the_node>: <xml_id_of_the_tax>}` and then get the tax with 

```suggestion
                    tax = self.env['account.chart.template'].with_company(self.env.company).ref(tax_xmlid)
```

As the xmlid is not editable and is unique, you will always get exactly one tax after this (if value in the node is a key of the mapping) so you can drop the raise. So this entire block becomes something like

```py
                if tax_xmlid := p12_to_tax_xmlid_map.get(line['tax_name']):
                    tax_ids.appends(
                        Command.set(
                            self.env['account.chart.template'].with_company(self.env.company).ref(tax_xmlid).ids
                        )
                    )
```

Looking more like what you had before (sorry for that 😅 )

And needing to change the mapping for:

```py
        p12_to_tax_xmlid_map = {
            '23': 'vz_kraj_23',
            '8': 'vz_kraj_8',
            '5': 'vz_kraj_5',
            '0 KR': 'vz_kraj_0',
            'zw': 'vz_kraj_zw',
            'oo': 'vz_stal',
            '0 WDT': 'vz_unia',
            'np II': 'vz_nabu',
            '0 EX': 'vz_kraj_zw',
            'np I': 'vz_impu',
        }
```

You might want to check the mapping I did but it feels right to me. `vz_kraj_zw` is a duplicate tho, i don't know if that's wanted ?

addons/l10n_pl_edi/models/account_move.py
```diff
@@ -2,9 +2,11 @@
 import re
 from xml.dom.minidom import parseString
 
+from dateutil.relativedelta import relativedelta
 from stdnum.pl.nip import compact
+from lxml import etree
```
Megaaaaaa: The nitpickest nitpick :smile:

```suggestion
from lxml import etree
from stdnum.pl.nip import compact
```

Megaaaaaa (commented):
Tought some more about what we dsicussed and found a nice alternative. Tell me what you think. (Also I haven't tested it locally) :+1:

addons/l10n_pl_edi/models/account_move.py
abmn-odoo: Great :+1:

Megaaaaaa (approved):
Well then, I think we're good to go :+1:

robodoo r+

PR #247815, opened 2026-02-09
[FIX] web: middle clicking url field

pparidans:
This commit fixes the obtrusive opening of the record's modal when middle clicking on a x2many's url field, in a form view. It applies the same logic as the one used for regular click (letting the url be opened, but not the modal).

Note: originally spotted in the Runbot build error's form view (Builds page or Error content...).

_Note to reviewers : doesn't look to apply to small screens, but not sure why... Feedback requested :pray:_

### DIFF
--- a/addons/web/static/tests/views/fields/url_field.test.js
+++ b/addons/web/static/tests/views/fields/url_field.test.js
@@ -1,5 +1,6 @@
 import { expect, getFixture, test } from "@odoo/hoot";
-import { queryAllAttributes, queryAllTexts, queryFirst } from "@odoo/hoot-dom";
+import { queryAllAttributes, queryAllTexts, queryFirst, click, middleClick } from "@odoo/hoot-dom";
+import { animationFrame } from "@odoo/hoot-mock";
 import {
     contains,
     defineModels,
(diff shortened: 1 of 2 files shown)


addons/web/static/tests/views/fields/url_field.test.js
```diff
@@ -193,3 +194,29 @@ test("with non falsy, but non url value", async () => {
...
+        resId: 1,
+        arch: `<form><field name="p" widget="one2many"/></form>`,
+    });
+    await click(".o_field_widget[name=url] a");
+    await animationFrame();
+    expect(".modal.o_technical_modal").toHaveCount(0);
+
+    await middleClick(".o_field_widget[name=url] a");
+    await animationFrame();
+    expect(".modal.o_technical_modal").toHaveCount(0);
```
aab-odoo, line 221: it could not open a modal as there's no form view defined, right? Without the fix, I assume it crashes? I would define a form view, it would help debugging if there's a regression someday.

Also, I assume that you can't assert that the click on the link succeeded?

aab-odoo (approved):
robodoo delegate+

Iucapad:
Thanks for the fix @pparidans 
Do you know if we can have other <a> with an href inside a list view, or except with a url field, it is not the case?

If so, maybe we could instead prevent onCellClicked in the list renderer if the target contains an href? I guess it may be more safe to explicitely stop click and auxclick as you did. In case the field is being used in similar situations

addons/web/static/tests/views/fields/url_field.test.js
pparidans: In my test, there is a form view as the one2many is against the same model :wink: 
So the behavior of the test matches what happens in "real life" with the form opening in a modal on click/middleClick.

image

> Also, I assume that you can't assert that the click on the link succeeded?

About testing that the click on a the link actually succeed... don't really know how I could assert that, indeed :grimacing:

PR #247817, opened 2026-02-09
[FIX] hr_org_chart: Access organization chart multicompany

maedekey:
When an employee A tries to access to the page of another employee B, if that employee has a manager belonging to a company to which employee A doesn't have access, the employee A will see an AccessError, even if employee A can have access to the page of employee B.

## Reproduction Steps
1. Create a second company to your database. Let's call these companies Company A and Company B.
2. Select Company A as your main company, but make sure Company B is also selected.
3. Create an Employee A.
4. Select Company B as tour main company; but make sure Company A is also selected.
5. Create an Employee B who has A as manager.
6. Create Employee C who has B as manager.
7. With a user only having access to Company B, log in and try to access Employee C's page.

### Expected behavior
Employee's C page loads correctly.

### Unexpected behavior
An access error occurs.

## Origin of the issue
A sudo() was missing.
__
opw-5910036

---


romaincarlier4 (approved):
LGTM, thanks for the fix
@robodoo r+

PR #247827, opened 2026-02-09
[IMP] html_editor: Blocking browser plugin translation on editable content.

OmarKhalil2001:
Browser translation plugins were altering editable content by replacing the
original content with translated versions, which caused issues when paired 
with auto-save. To prevent this behavior, the attribute `translate="no"` has 
been added to editable fields.

task-5485078


abd-msyukyu-odoo:
@robodoo r+

PR #247829, opened 2026-02-09
[FIX] iot_handlers: ensure custom handlers re-download

loouis-t:
The IoT Box will, most of the time, update before the database does.

In such event, the custom handlers will be removed from the IoT Box, but the etag of the files downloaded from the server will be kept, making the db return a 304 (not modified) when trying to download the handlers again.

We now ensure that this etag is removed, so that the database sends the files again.


Ysoroko (approved):
robodoo r+

PR #247830, opened 2026-02-09
[FIX] account: right number of payments skipped in send wizard

roto-odoo:
[FIX] account: right number of payments skipped in send wizard

Steps to reproduce:
- install modules account_sepa_direct_debit, account_iso20022
- create 2 vendor bills with payment terms so that there are 2
installments per bill, and post them
- from the list view, select both bills and click pay
- select SEPA as a payment method, a warning message is displayed
mentionning 4 payments

We want the warning to display a number of 2 payments because we're
paying only the first installment of each bill

This commit also fixes the visibility of the "group payment" button:
when two bills from different suppliers were selected with one
having installments, the button was visible

task-5917803


william-andre (approved):
God this code is complicated

@robodoo r+

PR #247831, opened 2026-02-09
[FIX] l10n_ar_withholding: remove deactivated tax from invoice

MAGREMENT:
# How to reproduce
- Install the l10n_ar_withholding module
- Pick a Vendor and add a Purchase Withholding tax in the accounting tab
- Go into configuration and disable that tax
- Create a Vendor Bill with that Vendor
- Confirm the bill and create a payment

# The problem
The deactivated tax is still computed in the payment

# Why
The function that fetches the withholding taxes does not check if the taxes are active or not

# The fix
I though of 2 solution : 
- Add an override to remove the taxes from the Vendor when they are deactivated
- Add a filter to the fetch function to check if the taxes are active

I chose the 2nd solution as it is the least invasive one.

opw-5917257


MAGREMENT:
@pablohmontenegro Okay I see, thank you for the explanation. It seems the active_test context is being set here https://github.com/odoo/odoo/blob/18.0/addons/l10n_account_withholding_tax/wizards/account_payment_register.py#L215. It's in a shared localization package for withholding taxes and the commit that added this feature does not explain why they decided to fetch the non-active taxes. What do you think we should do ?

PR #247833, opened 2026-02-09
[FIX] pos_mrp: correctly compute the cost of a kit

pedrambiria:
Before this commit, if a kit had multiple components, the cost of the line was counted as zero if the product cost method was FIFO or average.

opw-5911338

---

### DIFF
--- a/addons/pos_mrp/models/pos_order.py
+++ b/addons/pos_mrp/models/pos_order.py
@@ -16,7 +16,7 @@ class PosOrderLine(models.Model):
         #Get a flat list of all bom_line_ids
         bom_line_ids = [item.id for x in boms for item in x[0].bom_line_ids if set(item.bom_product_template_attribute_value_ids.ids).issubset(product.product_template_variant_value_ids.ids)]
         ml_product_to_consider = (product.bom_ids and [comp[0].product_id.id for comp in components]) or [product.id]
-        return stock_moves.filtered(lambda ml: ml.product_id.id in ml_product_to_consider and (ml.bom_line_id.id in bom_line_ids))
+        return stock_moves.filtered(lambda ml: ml.product_id.id in ml_product_to_consider and (ml.bom_line_id.id in bom_line_ids)).with_context(pos_order_line_id=self.id, bom_id=bom.id)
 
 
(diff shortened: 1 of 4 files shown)


addons/pos_mrp/models/pos_order.py
```diff
@@ -16,7 +16,7 @@ def _get_stock_moves_to_consider(self, stock_moves, product):
         #Get a flat list of all bom_line_ids
         bom_line_ids = [item.id for x in boms for item in x[0].bom_line_ids if set(item.bom_product_template_attribute_value_ids.ids).issubset(product.product_template_variant_value_ids.ids)]
         ml_product_to_consider = (product.bom_ids and [comp[0].product_id.id for comp in components]) or [product.id]
-        return stock_moves.filtered(lambda ml: ml.product_id.id in ml_product_to_consider and (ml.bom_line_id.id in bom_line_ids))
+        return stock_moves.filtered(lambda ml: ml.product_id.id in ml_product_to_consider and (ml.bom_line_id.id in bom_line_ids)).with_context(pos_order_line_id=self.id, bom_id=bom.id)
```
davidmonnom, line 19: `with_context` shouldn't be added in `pos_mrp` too ? 

```py
    def _get_stock_moves_to_consider(self, stock_moves, product):
        self.ensure_one()
        bom = product.env['mrp.bom']._bom_find(product, company_id=stock_moves.company_id.id, bom_type='phantom').get(product)
        if not bom:
            return super()._get_stock_moves_to_consider(stock_moves, product)
        boms, components = bom.explode(product, self.qty)
        #Get a flat list of all bom_line_ids
        bom_line_ids = [item.id for x in boms for item in x[0].bom_line_ids if set(item.bom_product_template_attribute_value_ids.ids).issubset(product.product_template_variant_value_ids.ids)]
        ml_product_to_consider = (product.bom_ids and [comp[0].product_id.id for comp in components]) or [product.id]
        return stock_moves.filtered(lambda ml: ml.product_id.id in ml_product_to_consider and (ml.bom_line_id.id in bom_line_ids))
```

davidmonnom (approved):
@robodoo r+

PR #247834, opened 2026-02-09
[FIX] website: clean popup on save snippet

blse-odoo:
There was a special handling for `s_popup` snippet when saving snippets: if the snippet was popup, it would not be cleaned directly, but instead its child would be cleaned. The goal was to avoid saving a hidden popup (like it normally is when saving a page), because it would show as an empty custom snippet.

This commit instead do not close the popup if it is the root of what is saved. And do not make special cases inside the snippet service.

Steps to reproduce:
- Open website builder
- Drop a popup
- Click on the eye to hide the popup
- Click on the eye to show the popup
- Save the snippet as a custom snippet
- Click to add a custom snippet
- Bug: with the inspector, you can see that the `o_draggable` class has not been cleaned from the popup

task-5149984

### DIFF
--- a/addons/website/static/src/builder/plugins/popup_visibility_plugin.js
+++ b/addons/website/static/src/builder/plugins/popup_visibility_plugin.js
@@ -71,6 +71,11 @@ export class PopupVisibilityPlugin extends Plugin {
     }
 
     cleanForSave(rootEl) {
+        // Do not hide a popup if it is saved as a custom snippet, otherwise it
+        // appears empty in the snippet dialog.
+        if (rootEl.matches(".s_popup")) {
+            return;
+        }
         // Hide the popups manually, as we cannot rely on the `onTargetHide`
         // flow since the cleaned popup is a clone and is not in the DOM.
         for (const modalEl of rootEl.querySelectorAll(".s_popup .modal.show")) {
--- a/addons/website/static/src/builder/snippet_model.js
+++ b/addons/website/static/src/builder/snippet_model.js
@@ -53,13 +53,6 @@ export class WebsiteSnippetModel extends SnippetModel {
         return label;
     }
 
-    cleanSnippetForSave(snippetCopyEl, cleanForSaveProcessors) {
-        const rootEl = snippetCopyEl.matches(".s_popup")
-            ? snippetCopyEl.firstElementChild
-            : snippetCopyEl;
-        super.cleanSnippetForSave(rootEl, cleanForSaveProcessors);
-    }
-
     getContext(snippetEl) {
         const context = super.getContext(...arguments);
         const editableParentEl = snippetEl.closest("[data-oe-model][data-oe-field][data-oe-id]");


addons/website/static/src/builder/plugins/popup_visibility_plugin.js
```diff
@@ -73,7 +73,7 @@ export class PopupVisibilityPlugin extends Plugin {
     cleanForSave(rootEl) {
```
sobo-odoo, line 73: Commit message:
- commit title => I would reformulate like this: `[FIX] website: clean the popup when saving it as a custom snippet` it is more precise as we mention the custom snippet part (so easier to find later) and you still have space so let's use it fully :slightly_smiling_face: 
- a special handling for `s_popup` snippet => nitpick: for the `s_popup` snippet
- if the snippet was popup => was a popup
- This commit instead do not close => does not
- > is saved. And do not make special cases inside the snippet service.

  Reformulate like this to be more clear/correct: `is saved, which allows to remove the special cases inside the snippet service.` :slightly_smiling_face: 

- > \- Drop a popup
    \- Click on the eye to hide the popup
    \- Click on the eye to show the popup

  You can replace these steps by this one: `Drop a popup snippet and click on it` (no need to hide and show, it only needs to be selected) :slightly_smiling_face:

addons/website/static/src/builder/plugins/popup_visibility_plugin.js
```diff
@@ -73,7 +73,7 @@ export class PopupVisibilityPlugin extends Plugin {
...
         // Hide the popups manually, as we cannot rely on the `onTargetHide`
         // flow since the cleaned popup is a clone and is not in the DOM.
-        for (const modalEl of rootEl.querySelectorAll(".s_popup .modal.show")) {
+        for (const modalEl of rootEl.querySelectorAll(".s_popup:not(:scope) > .modal.show")) {
```
sobo-odoo: Instead of using `:scope`, which I find not straightforward to understand here, I prefer that you return if the root is a popup, with a comment explaining why :+1:  So something like this just above :slightly_smiling_face: :arrow_down: 
```js
// Do not clean the popup if it is saved as a custom snippet, or else it
// will not be visible in the snippet dialog.
if (rootEl.matches(".s_popup")) {
    return;
}
```

It is easier to understand what is happening like this :slightly_smiling_face:

sobo-odoo: You should technically do something equivalent in the clean for save of the visibility plugin, because since we hide the invisible elements on save, we end up with the `data-invisible="1"` added on the popup, despite it being visible, which is inconsistent/incorrect :+1: 
But note that we can't mention `s_popup` in that plugin, since it is in html_builder

It is correctly removed on snippet drop though, but to me it is worse than the `o_draggable` class, which is correctly removed when saving the editor anyway, because here the visibility status is wrong :+1:

sobo-odoo (changes requested):
There is a problematic case to fix + some comments on the commit message :+1:

addons/website/static/src/builder/plugins/popup_visibility_plugin.js
blse-odoo: Indeed, having this `data-invisible="1"` is quite wrong. I did not realized that when extracting this commit from the rest of the invisible PR :sweat_smile: 
I think the easiest is to move back this commit to that PR, after the commit that removes `data-invisible`. Unless you see a very easy change to fix this here?

I'm a bit confused by your note about avoiding mentioning `s_popup` in the visibility plugin. If I understood correctly, that whole plugin should be moved to the `website` module, right? So it would not be too bad to add one more thing of `website` in that plugin of `website` (that is currently in `html_builder` by accident), no?

(I guess the `modalEl.parentElement.dataset.invisible = "1";` in the loop is useless :see_no_evil: )

PR #247835, opened 2026-02-09
[FIX] html_builder, website: use resize observer to detect mobile view switch

blse-odoo:
For elements that are invisible based on the size of the window, the "Invisible Elements" panel did not update its state when the size of the window changed (but instead it only did when the "Mobile Preview" button was toggled).

This commit observes the size of the window to trigger the update when the mobile breakpoint size is crossed.

All the implementations of `on_mobile_preview_clicked_handlers` need to know if the builder is currently in mobile view, so this commit updates them all to use the new resource `on_mobile_view_switched_handlers` which is triggered by the observer of the window size (and remove the old resource as it is now unused).

Steps to reproduce:
- Open website builder
- Drop a section snippet and click on it
- Set the "Visibility" option to make it invisible on mobile
- Toggle the "Mobile Preview" button
- Observe the snippet becomes hidden and is listed in the "Invisible Elements" panel
- Toggle the "Mobile Preview" button back to desktop view
- Reduce the size of the window
- Bug: the snippet becomes hidden, but is not listed in the "Invisible Elements" panel, and it is still the target of the builder

task-5149984

### DIFF
--- a/addons/html_builder/static/src/@types/plugins.d.ts
+++ b/addons/html_builder/static/src/@types/plugins.d.ts
@@ -19,7 +19,7 @@ declare module "plugins" {
     import { default_shape_providers, image_shape_groups_providers, on_shape_computed_handlers } from "@html_builder/plugins/image/image_shape_option_plugin";
     import { background_filter_target_providers, target_element_providers, on_bg_image_hidden_handlers } from "@html_builder/plugins/background_option/background_image_option_plugin";
     import { is_draggable_predicates, on_element_dragged_handlers, on_element_dropped_handlers, on_element_dropped_near_handlers, on_element_dropped_over_handlers, on_element_move_handlers, on_element_out_dropzone_handlers, on_element_over_dropzone_handlers, on_prepare_drag_handlers } from "@html_builder/core/drag_and_drop_plugin";
-    import { lower_panel_entries, on_mobile_preview_clicked_handlers, on_dom_updated_handlers } from "@html_builder/builder";
+    import { lower_panel_entries, on_dom_updated_handlers, on_mobile_view_switched_handlers } from "@html_builder/builder";
     import { on_target_revealed_handlers } from "@html_builder/sidebar/invisible_elements_panel";
     import { on_snippet_dragged_handlers, on_snippet_dropped_handlers, on_snippet_dropped_near_handlers, on_snippet_dropped_over_handlers, on_snippet_move_handlers, on_snippet_out_dropzone_handlers, on_snippet_over_dropzone_handlers } from "@html_builder/sidebar/block_tab";
     import { snippet_preview_dialog_bundles, snippet_preview_dialog_stylesheets_processors } from "@html_builder/snippets/add_snippet_dialog";
@@ -68,6 +68,7 @@ declare module "plugins" {
         on_bg_image_hidden_handlers: on_bg_image_hidden_handlers;
         on_cloned_handlers: on_cloned_handlers;
         on_current_options_containers_changed_handlers: on_current_options_containers_changed_handlers;
+        on_mobile_view_switched_handlers: on_mobile_view_switched_handlers;
         on_dom_updated_handlers: on_dom_updated_handlers;
         on_element_arrow_moved_handlers: on_element_arrow_moved_handlers;
         on_element_dragged_handlers: on_element_dragged_handlers;
@@ -77,7 +78,6 @@ declare module "plugins" {
         on_element_move_handlers: on_element_move_handlers;
         on_element_out_dropzone_handlers: on_element_out_dropzone_handlers;
         on_element_over_dropzone_handlers: on_element_over_dropzone_handlers;
-        on_mobile_preview_clicked_handlers: on_mobile_preview_clicked_handlers;
         on_prepare_drag_handlers: on_prepare_drag_handlers;
         on_removed_handlers: on_removed_handlers;
         on_replicated_handlers: on_replicated_handlers;
--- a/addons/html_builder/static/src/builder.js
+++ b/addons/html_builder/static/src/builder.js
@@ -7,7 +7,6 @@ import {
     onWillDestroy,
     onWillStart,
     onWillUnmount,
-    onWillUpdateProps,
     status,
 } from "@odoo/owl";
 import { useHotkey } from "@web/core/hotkeys/hotkey_hook";
@@ -34,7 +33,8 @@ const ONLY_ALLOW_INLINE_TAGS = new Set([
 ]);
 
 /**
- * @typedef {(() => void)[]} on_mobile_preview_clicked_handlers
+ * @typedef {((args: {isMobileView: boolean}) => ())[]} on_mobile_view_switched_handlers
+ * called when the screen size switches between mobile and desktop view
  * @typedef {(() => void)[]} on_dom_updated_handlers
  * @typedef {{ Component: Component; props: object; }[]} lower_panel_entries
  */
@@ -82,7 +82,7 @@ export class Builder extends Component {
         });
         this.invisibleElementsPanelState = useState({
             invisibleEls: [],
-            invisibleSelector: this.getInvisibleSelector(),
+            invisibleSelector: "",
         });
         useHotkey("control+z", () => this.undo());
         useHotkey("control+y", () => this.redo());
@@ -140,8 +140,9 @@ export class Builder extends Component {
                     on_dom_updated_handlers: () => {
                         this.triggerDomUpdated();
                     },
-                    on_mobile_preview_clicked_handlers: withSequence(20, () => {
+                    on_mobile_view_switched_handlers: withSequence(20, () => {
                         this.triggerDomUpdated();
+                        this.updateInvisibleEls();
                     }),
                     on_will_save_handlers: () => {
                         const snippetMenuEl = this.builder_sidebarRef.el;
@@ -245,6 +246,19 @@ export class Builder extends Component {
                     ev.stopPropagation();
                 }
             };
+
+            // Use a resize observer to trigger `on_mobile_view_switched_handlers`
+            // when the view size switches between desktop and mobile view
+            let isMobileView = this.editor.config.isMobileView(this.editableEl);
+            this.resizeObserver = new ResizeObserver(() => {
+                const wasMobileView = isMobileView;
+                isMobileView = this.editor.config.isMobileView(this.editableEl);
+                if (wasMobileView !== isMobileView) {
+                    this.editor.trigger("on_mobile_view_switched_handlers", { isMobileView });
+                }
+            });
+            this.resizeObserver.observe(this.editableEl);
+
             this.editor.attachTo(this.editableEl);
         });
 
@@ -256,6 +270,7 @@ export class Builder extends Component {
             editShadow: this.editShadow.bind(this),
         });
         onWillDestroy(() => {
+            this.resizeObserver.disconnect();
             this.editor.destroy();
         });
 
@@ -269,14 +284,6 @@ export class Builder extends Component {
         onWillUnmount(() => {
             this.editableEl.removeEventListener("dragstart", this.onDragStart);
         });
-        onWillUpdateProps((nextProps) => {
-            if (nextProps.isMobile !== this.props.isMobile) {
-                this.updateInvisibleEls(nextProps.isMobile);
-                this.invisibleElementsPanelState.invisibleSelector = this.getInvisibleSelector(
-                    nextProps.isMobile
-                );
-            }
-        });
     }
     async triggerDomUpdated() {
         this.lastTrigerUpdateId++;
@@ -287,12 +294,6 @@ export class Builder extends Component {
         await Promise.allSettled(getStatePromises);
         const isLastTriggerId = this.lastTrigerUpdateId === currentTriggerId;
         resolve(isLastTriggerId);
-    }
-
-    getInvisibleSelector(isMobile = this.props.isMobile) {
-        return `.o_snippet_invisible, ${
-            isMobile ? ".o_snippet_mobile_invisible" : ".o_snippet_desktop_invisible"
-        }`;
     }
 
     /**
@@ -337,12 +338,16 @@ export class Builder extends Component {
 
     onMobilePreviewClick() {
         this.props.toggleMobile();
-        this.editor.resources["on_mobile_preview_clicked_handlers"].forEach((handler) => handler());
     }
 
-    updateInvisibleEls(isMobile = this.props.isMobile) {
+    updateInvisibleEls() {
+        const isMobile = this.editor.config.isMobileView(this.editor.editable);
+        const invisibleSelector = `.o_snippet_invisible, ${
+            isMobile ? ".o_snippet_mobile_invisible" : ".o_snippet_desktop_invisible"
+        }`;
+        this.invisibleElementsPanelState.invisibleSelector = invisibleSelector;
         this.invisibleElementsPanelState.invisibleEls = [
-            ...this.editor.editable?.querySelectorAll(this.getInvisibleSelector(isMobile)) || [],
+            ...this.editor.editable.querySelectorAll(invisibleSelector),
         ];
     }
 
--- a/addons/html_builder/static/src/core/disable_snippets_plugin.js
+++ b/addons/html_builder/static/src/core/disable_snippets_plugin.js
@@ -17,7 +17,7 @@ export class DisableSnippetsPlugin extends Plugin {
         on_removed_handlers: this.disableUndroppableSnippets.bind(this),
         on_undone_handlers: this.disableUndroppableSnippets.bind(this),
         on_redone_handlers: this.disableUndroppableSnippets.bind(this),
-        on_mobile_preview_clicked_handlers: withSequence(
+        on_mobile_view_switched_handlers: withSequence(
             20,
             this.disableUndroppableSnippets.bind(this)
         ),
--- a/addons/html_builder/static/src/core/visibility_plugin.js
+++ b/addons/html_builder/static/src/core/visibility_plugin.js
@@ -1,6 +1,5 @@
 import { Plugin } from "@html_editor/plugin";
 import { getElementsWithOption } from "@html_builder/utils/utils";
-import { withSequence } from "@html_editor/utils/resource";
 
 /**
  * @typedef { Object } VisibilityShared
@@ -30,10 +29,7 @@ export class VisibilityPlugin extends Plugin {
     ];
     /** @type {import("plugins").BuilderResources} */
     resources = {
-        on_mobile_preview_clicked_handlers: withSequence(
-            10,
-            this.onMobilePreviewClicked.bind(this)
-        ),
+        on_mobile_view_switched_handlers: this.onMobileViewSwitched.bind(this),
         system_attributes: ["data-invisible"],
         system_classes: ["o_snippet_override_invisible"],
         clean_for_save_processors: this.cleanForSaveVisibility.bind(this),
@@ -128,7 +124,7 @@ export class VisibilityPlugin extends Plugin {
         }
     }
 
-    onMobilePreviewClicked() {
+    onMobileViewSwitched() {
         const deviceInvisibleEls = this.editable.querySelectorAll(deviceInvisibleSelector);
         const currentContainerTargetEl = this.dependencies["builderOptions"].getTarget();
         for (const deviceInvisibleEl of [...deviceInvisibleEls]) {
--- a/addons/website/static/tests/builder/invisibility_options.test.js
+++ b/addons/website/static/tests/builder/invisibility_options.test.js
@@ -5,6 +5,7 @@ import {
     defineWebsiteModels,
     setupWebsiteBuilder,
     setupWebsiteBuilderWithSnippet,
+    toggleMobilePreview,
 } from "@website/../tests/builder/website_helpers";
 import { insertText, undo } from "@html_editor/../tests/_helpers/user_actions";
 import { setSelection } from "@html_editor/../tests/_helpers/selection";
@@ -103,7 +104,7 @@ test("check invisible element after save", async () => {
 test("click on 'Show/hide on mobile' in mobile view", async () => {
     await setupWebsiteBuilder(websiteContent);
     await contains(":iframe .col-lg-3").click();
-    await contains("button[data-action='mobile']").click();
+    await toggleMobilePreview();
 
     await contains("button[data-action-id='toggleDeviceVisibility']:last").click();
     expect(".o-snippets-tabs button:contains('Blocks')").toHaveClass("active");
@@ -122,7 +123,7 @@ test("click on 'Show/hide on mobile' in desktop view", async () => {
 test("click on 'Show/hide on desktop' in mobile view", async () => {
     await setupWebsiteBuilder(websiteContent);
     await contains(":iframe .col-lg-3").click();
-    await contains("button[data-action='mobile']").click();
+    await toggleMobilePreview();
 
     await contains(
         "[data-container-title='Column']  button[data-action-id='toggleDeviceVisibility']"
@@ -138,13 +139,13 @@ test("hide on mobile and toggle mobile view", async () => {
     await contains(":iframe .col-lg-3").click();
 
     await contains("button[data-action-id='toggleDeviceVisibility']:last").click();
-    await contains("button[data-action='mobile']").click();
+    await toggleMobilePreview();
     expect(":iframe .col-lg-3").not.toHaveClass("o_snippet_override_invisible");
     expect(queryOne(".o_we_invisible_el_panel .o_we_invisible_entry i")).toHaveClass(
         "fa-eye-slash"
     );
 
-    await contains("button[data-action='mobile']").click();
+    await toggleMobilePreview();
     expect(".o_we_invisible_el_panel").not.toHaveCount();
 });
 
--- a/addons/html_builder/static/src/core/overlay_buttons/overlay_buttons_plugin.js
+++ b/addons/html_builder/static/src/core/overlay_buttons/overlay_buttons_plugin.js
@@ -35,7 +35,7 @@ export class OverlayButtonsPlugin extends Plugin {
         on_selection_leave_handlers: this.showOverlayButtonsUi.bind(this),
         on_step_added_handlers: this.refreshButtons.bind(this),
         on_current_options_containers_changed_handlers: this.addOverlayButtons.bind(this),
-        on_mobile_preview_clicked_handlers: withSequence(20, this.refreshButtons.bind(this)),
+        on_mobile_view_switched_handlers: withSequence(20, this.refreshButtons.bind(this)),
     };
 
     setup() {
--- a/addons/website/static/tests/builder/builder_action.test.js
+++ b/addons/website/static/tests/builder/builder_action.test.js
@@ -19,7 +19,12 @@ import {
     onRpc,
     patchWithCleanup,
 } from "@web/../tests/web_test_helpers";
-import { addPlugin, defineWebsiteModels, setupWebsiteBuilder } from "./website_helpers";
+import {
+    addPlugin,
+    defineWebsiteModels,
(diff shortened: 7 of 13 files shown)


addons/html_builder/static/src/@types/plugins.d.ts
```diff
@@ -10,6 +10,7 @@ declare module "plugins" {
     import { dropzone_selector, DropZoneShared, is_valid_for_sibling_dropzone_predicates } from "@html_builder/core/drop_zone_plugin";
```
sobo-odoo: Commit message:
- Title: I would reformulate it to something like this :arrow_down:  
  ```
  [IMP] html_builder, *: use resize observer to update the mobile view
  
  *: website
  ```

  => to me, it is more an IMP than a FIX, it is a resize observer, and we use it to update things, not to know (nitpick but to know, we use `isMobileView` :stuck_out_tongue:) 

- > but instead when the mobile preview button changed

  => I would reformulate like this: `it only did when the mobile preview was toggled` :slightly_smiling_face: 

- > This commit replaces the resource `on_mobile_preview_clicked` by
`device_view_switched_handlers` that is triggered by a plugin that
observes the size of the window.

  Here you are describing your changes/the code, not explaining the fix :+1: You should say something like this :arrow_down: :

   Instead of refreshing the visibility status only when the mobile preview was toggled (thanks to the button), we now observe the window size changes and refresh when the mobile breakpoint is crossed. Note that the `on_mobile_preview_clicked` was therefore renamed into `device_view_switched_handlers`.

- Add a section => Drop a section snippet and click on it (to be more precise)
- Make it invisible on mobile => Use the "Visibility" option to hide it on mobile.
- > - Reduce the size of the window

  I would add the following steps before this one, to make the steps more relevant :slightly_smiling_face: :
  ```
  - Toggle the mobile preview.
    => the snippet is invisible and listed in the "Invisible Elements" panel.
  - Toggle back the desktop view.
  ```

addons/html_builder/static/src/@types/plugins.d.ts
```diff
@@ -10,6 +10,7 @@ declare module "plugins" {
...
     import { on_replicated_handlers } from "@html_builder/core/field_change_replication_plugin";
     import { MediaWebsiteShared } from "@html_builder/core/media_website_plugin";
+    import { on_device_view_switched_handlers } from "@html_builder/core/mobile_preview_plugin";
```
sobo-odoo: I prefer that you call it `on_mobile_preview_toggled_handlers` instead, since it is easier to remember "mobile preview" as we always used it, and never used "device view"; and "toggled" can be used to describe what happens when we resize (as you can toggle on and off) :slightly_smiling_face: 

`on_mobile_view_toggled_handlers` could be used too (to not mention "preview"), but "mobile preview" would match with how you called the plugin :+1:

addons/html_builder/static/src/builder.js
```diff
@@ -272,7 +265,7 @@ export class Builder extends Component {
         resolve(isLastTriggerId);
     }
 
-    getInvisibleSelector(isMobile = this.props.isMobile) {
+    getInvisibleSelector(isMobile = this.editor.config.isMobileView(this.editor.editable)) {
```
sobo-odoo: Before, we were reading a value that was used as default parameter, but now you are computing one so it should not be considered as a default value/parameter anymore (it is not really correct/a good practice)
You should compute it in the function directly :+1: 
```js
getInvisibleSelector(isMobile) {
  if (isMobile === undefined) {
      isMobile = this.editor.config.isMobileView(this.editor.editable);
   }
 ...
```

=> same below in `updateInvisibleEls` :slightly_smiling_face:

addons/html_builder/static/src/core/mobile_preview_plugin.js
```diff
@@ -0,0 +1,27 @@
...
+                this.trigger("on_device_view_switched_handlers", { isMobileView });
+            }
+        });
+        this.observer.observe(this.editable);
+    }
+
+    destroy() {
+        super.destroy();
+        this.observer.disconnect();
+    }
```
sobo-odoo: Do this in the cleanup instead of the destroy (that's what we prefer to do in general) :slightly_smiling_face: 

So at the end of the `setup`, do this
```js
this._cleanups.push(() => {
    this.observer.disconnect();
});
```

addons/html_builder/static/src/builder.js
```diff
@@ -320,10 +313,10 @@ export class Builder extends Component {
 
     onMobilePreviewClick() {
         this.props.toggleMobile();
-        this.editor.resources["on_mobile_preview_clicked_handlers"].forEach((handler) => handler());
+        this.triggerDomUpdated();
```
sobo-odoo: Why do you need to `triggerDomupdated` here, if the resource already triggers it above in the file ? :thinking: It seems redundant :+1:

addons/html_builder/static/src/builder.js
```diff
@@ -131,9 +129,12 @@ export class Builder extends Component {
                     on_dom_updated_handlers: () => {
                         this.triggerDomUpdated();
                     },
-                    on_mobile_preview_clicked_handlers: withSequence(20, () => {
+                    on_device_view_switched_handlers: () => {
```
sobo-odoo: These sequences were deliberately high (20 seemed enough for that use case), because they are the "system" resources which must always be the last ones to run to keep the builder consistent :+1: 

Let's suppose that a snippet/plugin needs to do something when the view is changed (for example: change a state of a value used for the overlay buttons, or hide an element which will imply that no inner snippet can be dropped anymore)

Without the sequence, it could happen that we update all the builder, the disabled snippets, the overlay buttons,... and only then we will run that plugin code 
=> the overlay buttons or the disabled snippets would not be correct, because we updated them too soon, and so the builder is inconsistent :+1: 

Let's restore the sequence, so we don't have to worry when using this resource :slightly_smiling_face: 

We could increase the sequence to be more obvious, but at that time, 20 seemed enough for what we do; and we could also add comments to explain that they should always be the last ones :slightly_smiling_face:

addons/html_builder/static/src/builder.js
```diff
@@ -72,7 +70,7 @@ export class Builder extends Component {
         });
         this.invisibleElementsPanelState = useState({
             invisibleEls: [],
-            invisibleSelector: this.getInvisibleSelector(),
+            invisibleSelector: this.getInvisibleSelector(false),
```
sobo-odoo: Technically, `false` is wrong, because you could start the builder in the mobile preview or with a small screen, which would not give the right value for the selector :+1: 
=> You should leave it empty so it is undefined and so the function will compute it :slightly_smiling_face: 

We therefore have an issue: 
- Drop Text-Image
- Set a column as invisible on mobile, and then the section on invisible on mobile too
- Save
- Activate the mobile preview, or resize the screen so when the builder will start, it will be in mobile view
- Go in edit mode
- => The hierarchy of the blocks is broken in the invisible elements panel, because the invisibleSelector is the wrong one

  image

addons/html_builder/static/src/builder.js
```diff
@@ -131,9 +129,12 @@ export class Builder extends Component {
                     on_dom_updated_handlers: () => {
                         this.triggerDomUpdated();
                     },
-                    on_mobile_preview_clicked_handlers: withSequence(20, () => {
+                    on_device_view_switched_handlers: () => {
                         this.triggerDomUpdated();
-                    }),
+                        this.updateInvisibleEls();
+                        this.invisibleElementsPanelState.invisibleSelector =
+                            this.getInvisibleSelector();
```
sobo-odoo: Question: Why don't we update `invisibleSelector` inside `updateInvisibleEls` directly ? :slightly_smiling_face: It would make more sense to update everything related `invisibleElementsPanelState` in one go, and we still call `getInvisibleSelector` in it anyway so at best, it will remove this redundant call here, and at worst we will update something for nothing but it won't cost anything since the value was already computed anyway :slightly_smiling_face: 

It would look like this (following my other suggestion on the default value too):
```js
updateInvisibleEls(isMobile) {
    if (isMobile === undefined) {
        isMobile = this.editor.config.isMobileView(this.editor.editable);
    }
    const invisibleSelector = this.getInvisibleSelector(isMobile);
    this.invisibleElementsPanelState.invisibleEls = [
        ...(this.editor.editable?.querySelectorAll(invisibleSelector) || []),
    ];
    this.invisibleElementsPanelState.invisibleSelector = invisibleSelector;
}
```

addons/html_builder/static/src/core/disable_snippets_plugin.js
```diff
@@ -17,10 +16,7 @@ export class DisableSnippetsPlugin extends Plugin {
         on_removed_handlers: this.disableUndroppableSnippets.bind(this),
         on_undone_handlers: this.disableUndroppableSnippets.bind(this),
         on_redone_handlers: this.disableUndroppableSnippets.bind(this),
-        on_mobile_preview_clicked_handlers: withSequence(
-            20,
-            this.disableUndroppableSnippets.bind(this)
-        ),
+        on_device_view_switched_handlers: this.disableUndroppableSnippets.bind(this),
```
sobo-odoo: Repeating my comment about the sequence, we should keep it as guarantee to be run last :+1: 
Same for the other plugins :+1:

addons/html_builder/static/src/core/visibility_plugin.js
```diff
@@ -30,10 +30,7 @@ export class VisibilityPlugin extends Plugin {
     ];
     /** @type {import("plugins").BuilderResources} */
     resources = {
-        on_mobile_preview_clicked_handlers: withSequence(
-            10,
-            this.onMobilePreviewClicked.bind(this)
-        ),
+        on_device_view_switched_handlers: withSequence(5, this.onMobilePreviewClicked.bind(this)),
```
sobo-odoo: You should rename the handler, to be consistent with the resource, since it is not on click anymore :wink:

addons/html_builder/static/src/core/mobile_preview_plugin.js
```diff
@@ -0,0 +1,27 @@
...
+        this.isMobileView = this.config.isMobileView(this.editable);
+        this.observer = new ResizeObserver(() => {
+            const isMobileView = this.config.isMobileView(this.editable);
+            if (this.isMobileView !== isMobileView) {
+                this.isMobileView = isMobileView;
+                this.trigger("on_device_view_switched_handlers", { isMobileView });
+            }
+        });
+        this.observer.observe(this.editable);
+    }
```
sobo-odoo: I would have preferred that you do this directly in the `builder.js` file instead of creating a plugin, since everything mobile related is in that file, I find it a bit "treasure hunt" like this :thinking: (the `isMobileView` utils is in the `config` for the same reason, it is a more general builder/system concept to me)
There is surely a way to add a resize observer without needing a plugin, as long as you know what the editable is :slightly_smiling_face: 
I let you investigate and if you estimate that a plugin is better, then we could go with it :+1:

addons/html_builder/static/src/core/mobile_preview_plugin.js
```diff
@@ -0,0 +1,27 @@
+import { Plugin } from "@html_editor/plugin";
+
+/**
+ * @typedef {((args: {isMobileView: boolean}) => ())[]} on_device_view_switched_handlers
+ * called on switch between previewing a small screen and a large screen
+ */
+
+export class MobilePreviewPlugin extends Plugin {
```
sobo-odoo: Maybe call it `MobileViewPlugin` instead, given that the preview is more when the iframe is wrapped in the phone when we click on the button :slightly_smiling_face: 
Maybe `MobileViewManagerPlugin` could be nice too :thinking:

addons/html_builder/static/src/core/mobile_preview_plugin.js
```diff
@@ -0,0 +1,27 @@
...
+    static id = "mobilePreview";
+
+    setup() {
+        this.isMobileView = this.config.isMobileView(this.editable);
+        this.observer = new ResizeObserver(() => {
```
sobo-odoo: Nitpick: call it `this.resizeObserver` to be more precise :slightly_smiling_face:

addons/website/static/tests/builder/invisibility_options.test.js
```diff
@@ -139,12 +139,14 @@ test("hide on mobile and toggle mobile view", async () => {
...
     await contains("button[data-action-id='toggleDeviceVisibility']:last").click();
     await contains("button[data-action='mobile']").click();
+    await animationFrame();
     expect(":iframe .col-lg-3").not.toHaveClass("o_snippet_override_invisible");
     expect(queryOne(".o_we_invisible_el_panel .o_we_invisible_entry i")).toHaveClass(
         "fa-eye-slash"
     );
 
     await contains("button[data-action='mobile']").click();
+    await animationFrame();
```
sobo-odoo: It's strange that you suddenly need to wait an animation frame everywhere, given that there is already an implicit one done inside `click` :thinking: 
Do you really need it ?

sobo-odoo (changes requested):

addons/html_builder/static/src/@types/plugins.d.ts
blse-odoo: > Title: `use resize observer to update the mobile view`

The current title is a compromise to fit in one line, and it is not the best :sweat_smile: I agree with your reason to change it, but I'm also not convinced by your suggestion. To me it sounds like the resize observer will be used to update if we are in mobile view or not. But it is more about triggering the handlers that need to know if we are in mobile view or not, and the change about whether we are in mobile view comes from elsewhere.
What do you think of the following?
```
use resize observer to detect mobile view switch
```
> more an IMP than a FIX

Could you explain a bit more the reasoning behind this? To me it is not an imp because it does not bring any new feature or behavior. But it is a fix (to me) because the current implementation reads the state of things with `isMobileView` but does not always read it again when that value changed (and thus may sometimes present an inconsistent interface: wrong overlay buttons, invisible elements list not matching what is displayed), and the commit is made so that it _always_ reads the state again when it changed (well, if the commit is correct :sweat_smile:)

> Here you are describing your changes/the code, not explaining the fix

I agree I was describing the code change, but you also do that in your suggestion ^^ And I was explaining the fix with "a plugin that observes the size of the window" (maybe that is a bit too short, even with the explanation from the previous paragraph). I think your suggestion is better because it leads with the explanation of the fix (instead of the naming detail), I'm not sure about re-summarizing what the current code does, as it was done in previous paragraph. I'll reformulate

(for the other points, I'll apply your suggestions :slightly_smiling_face: )

addons/html_builder/static/src/@types/plugins.d.ts
blse-odoo: I used "device" and "switch" because we can switch from viewing one device to another. But I agree that we are probably better calling it "mobile" anyway. I'm not sure about "toggled" (we can switch on and off just like we can toggle on and off, right?). And the plugin has indeed not the best name, I'll rename it to "MobileViewPlugin" (to match "isMobileView")

addons/html_builder/static/src/core/mobile_preview_plugin.js
blse-odoo: I went with `MobileViewPlugin`

addons/html_builder/static/src/@types/plugins.d.ts
sobo-odoo: > To me it sounds like the resize observer will be used to update if we are in mobile view or not. But it is more about triggering the handlers that need to know if we are in mobile view or not, and the change about whether we are in mobile view comes from elsewhere
What do you think of the following?
`use resize observer to detect mobile view switch`

Yes I was trying to find a compromise too but it may not be the best :wink: => to me `update if we are in mobile view` implies that we will call/update everything related to the mobile view, so the handlers and all were included
Your suggestion is okay with me :slightly_smiling_face: 

> Could you explain a bit more the reasoning behind this? To me it is not an imp because it does not bring any new feature or behavior.

It is an IMP because we deliberately chose to not update the view when the screen size changed before, it would have been complicated with how the old builder worked (to refresh the options and all)
With the new builder, it is much easier to do it, so to me you improve the old limited behavior thanks to the new system :slightly_smiling_face: A limitation is not a bug, it is just the best we could achieve at a given time with a given technology, and when these limitation disappears, we are improving :slightly_smiling_face: 

> I agree I was describing the code change, but you also do that in your suggestion

Of course, when you explain you will describe a bit :+1: What bugged me in your version is that you put the accent on the resource changes, so to me it was really a description of the code, and not of the behavior. For example, the accent should be put on the fact that a new plugin observing the resize was added, and which will now be the coordinator of the mobile view (instead of just the mobile preview button), so you have a bit more context
And then you can describe some notable changes, like a resource renaming :slightly_smiling_face: 

=> so as a summary, in a commit message we have to understand what it does, what was the behavior before, why it is needed, some choice explanations,... so it is a complement to the code :slightly_smiling_face:

addons/html_builder/static/src/@types/plugins.d.ts
sobo-odoo: As long as we have "mobile" and "view", I'm fine with it :slightly_smiling_face:

addons/html_builder/static/src/builder.js
blse-odoo: Indeed, it is! (I'm not sure how I ended up with both ^^)

addons/html_builder/static/src/builder.js
blse-odoo: > You should leave it empty so it is undefined and so the function will compute it

That does not work, because we don't have the `editor` 😬. I set it to `false` because I thought (by mistake, thank you for catching it! 🙂) that the first value would not be relevant, just like the empty array in `invisibleEls`.

We could either compute it manually at this point (we could move the definition of `isMobileView` out from the config, and use the iframe instead of the editable element (because it is not loaded yet)), or we could compute it in `onMounted` like we do for `updateInvisibleEls` (which discards the empty array initially passed to `invisibleEls`, and we could initialize it with an even more obviously wrong value like `""`).
I think the latter makes more sense. And that would be implicitly done by moving the `getInvisibleSelector` into `updateInvisibleEls` (like you suggest in another comment)

addons/html_builder/static/src/builder.js
blse-odoo: I agree that this should only result in a win 👍 
From looking at the history, the `invisibleSelector` is stored in the state since [this commit](https://github.com/odoo/odoo/commit/faf57afddfaf4863da4417d0a3a40f9caae3133e) which splits the invisible panel from the builder (as it is adding other things next to the panel). It looks like it missed the opportunity to do what you are suggesting now (but that was already weird before that the two were not computed at the same time ^^).
So I'll do as you suggest! 🙂

addons/website/static/tests/builder/invisibility_options.test.js
blse-odoo: Yes, I added it in tests failing without it. From what I understood, clicking the button causes the builder to change the size of the iframe (after the first animation frame), which makes the editable smaller and triggers the resize observer's callback, which changes the content of the panel (after the second animation frame).
I did not add that animation frame delay everywhere, but only in the failing tests (maybe I should add it everywhere we toggle that button?)

addons/html_builder/static/src/core/mobile_preview_plugin.js
blse-odoo: I did not realize that, but yes, having the function `isMobileView` defined in `builder.js` and moving the caller of `on_device_view_switched_handlers` in this plugin is not ideal.
What do you think of moving the definition of `isMobileView` in this plugin as well (and make it a `shared` so it's callable from elsewhere) ? Future uses of the shared would more easily spot the resource that they should probably implement as well.
The other mobile-related things are things related to invisible just as well (they are in `builder.js` mostly because we currently store the state of the invisible panel there, I think it should move to another plugin as well, but probably in another PR)

addons/html_builder/static/src/builder.js
sobo-odoo: Indeed, if it is correctly recomputed in `onMounted` (once you make the change), then keeping `false` is fine :slightly_smiling_face: Maybe add a comment saying that it will be recomputed once everything is ready, and that they are dummy values :slightly_smiling_face:

addons/website/static/tests/builder/invisibility_options.test.js
sobo-odoo: In that case, I prefer that you create an utils function (like `toggleMobilePreview`) and that you use it everywhere we use that button :slightly_smiling_face: 
It will prevent having tests needing it and other that don't so everything will be consistent (it's strange that some tests need it and others don't)
And in the future, if we change the system again, we will have only one place to modify :slightly_smiling_face:

addons/html_builder/static/src/core/mobile_preview_plugin.js
sobo-odoo: For me, we should not move `isMobileView` out of the config, because it is really a builder/system concept and we should not depend on a plugin to check the mode of the builder :slightly_smiling_face: 
The mobile breakpoint used to decide if we are in mobile is passed as props to the builder and saved in the config too, this was done like this for mass mailing, because their breakpoint is `md` :+1: 
And the button to toggle the mobile preview is also defined in `builder.js`, because it is a part of the component so even if we move the function to the plugin, the preview will be toggled in another file anyway, and we will still need to track what is happening

That's why I prefer that everything mobile related stays in `builder.js` :slightly_smiling_face: 
(You could define the observer in the `onMounted`, with the initial state computed manually)

About the invisible elements panel, it should not move to a plugin either because it is part of the `Builder` component too :+1: The code could be extracted better to the `InvisibleElementsPanel` component, but given that we are in the OWL universe, I don't think adding a plugin to communicate between them is good, when you can just store in the parent component and communicate with props, it is a useless intermediary to me  :+1: (But I agree that we could improve the way we do it)

That said, I have a task where I will need to change the mobile view in the middle of an operation (to make the grid mode responsive), so I will surely need a way to stop the resize observer (to not update everything when the screen will be resized) and a shared function seems logical for that when I think about it :thinking: 

So, if you think that the function to stop the observer can be in the config too (it doesn't really bother me), do everything in builder.js :+1: 
But if you think that a shared function is better, then keep your current code without moving anything else (you can add a comment in `toggleMobile` saying that everything will be managed by the mobileView plugin) :slightly_smiling_face:

blse-odoo: I'm sorry I don't agree with or understand a few things here :thinking: I'll try to pick the core points that I believe would help me the most understand your position if they were clearer to me:
> it is really a builder/system concept

I wonder why do you think it is different from a lot of other things that are mostly (or completely) not handled in `builder.js`, which I feel like they would deserve the same rank of builder/system concept. For example, why do you think the list of options should not be moved into `builder.js`, or the mutex of the `operation` plugin?

> everything mobile related stays in `builder.js`

To me, it is already not the case: the "Mobile Preview" button forwards its toggle events to the parent component (outside `builder.js`) which takes care of maintaining the state of whether the button is active or not, as well as showing the iframe in a smaller container or not.
And for the function `isMobileView`, it is not looking up the value of `isMobile` (the "active" state of the button) but it is mimicking rules defined elsewhere (outside `builder.js` and the parent components) that determines whether the document appears for a small or a large screen.

> the invisible elements panel should not move to a plugin

From what I understood, we want to move the invisible-related stuff out of `html_builder` and into `website` module. Moving the panel to a plugin seems to be a rather easy way to be able to move it between module. How do you imagine to move that otherwise?

sobo-odoo: > For example, why do you think the list of options should not be moved into builder.js, or the mutex of the operation plugin

By definition, the plugins are well scoped functionalities that can be extended, added and excluded depending on the context, and which can be used by other plugins by adding dependencies to them (=> as the name says it, they can be plugged in if we want their functionality) :+1:
For example, we could imagine a module using the builder but that would not need the mutex, because everything in it would be synchronous.  We could also imagine a builder where the options would not be used (it was the case for the translate mode until recently, we had a builderOption plugin that was lighter for it, but since we are making more complex options in translate mode, I think it is now used by both modes). We also exclude many html_editor plugin because they are not needed in website :+1: 

The fact that we are in mobile view or not is different to me, because it is not a functionality but a status. Like the fact that you can add some CSS using a media breakpoint anytime, you should also be able to know if you are above the breakpoint in JS anytime too. So it's just an information that should be available universally, without needing to depend on a plugin to know it :slightly_smiling_face: 
In fact, it was an utils function before, so we were just importing it when needed (and since we had way fewer but bigger files, we didn't import it a lot). But now that it needs the breakpoint of the config for mass_mailing, it cannot be imported anymore.
So the easiest and shortest way to always have access to it is to put it in the config, in which it is also guaranteed to always exist :+1: 

Also, for me, we created plugins for too many things that should not have required them at all if we had done things differently (like the plugins for the different options, we should have done it another way, directly in the option components) 

Anyway, just to explain my vision, as asked (sorry for the long text) :slightly_smiling_face: 

> To me, it is already not the case: the "Mobile Preview" button forwards its toggle events to the parent component (outside builder.js)

When I say in `builder.js`, I don't mean that file exactly, but the OWL environment the Builder component belongs to => So here it is forwarded to the parent component of the builder, which uses that information as it needs to :+1: But all that happens inside `WebsiteBuilderClientAction` for website, and the equivalent for mass_mailing
So everything is in the same "OWL component" in the end :+1:

> From what I understood, we want to move the invisible-related stuff out of html_builder and into website module. 

What needs to be moved is the invisible stuff related to website, not the invisibility concept itself :slightly_smiling_face: 
For example, the ´invisibleSelector´ uses classes that are website-only, so we should in fact give these selectors from the website builder, as props or in the config :+1: And same for the device/conditional visibility options/plugins, they are website options :+1: 
Mass mailing could decide one day to add their own visibility options, to show blocks depending on the size of the mailing app screen, and in that case, it would also provide its own selectors :slightly_smiling_face: 
So the panel should stay in builder.js, but what it uses to compute and display the entries should be scoped :slightly_smiling_face:

addons/html_builder/static/src/core/visibility_plugin.js
```diff
@@ -30,10 +30,7 @@ export class VisibilityPlugin extends Plugin {
     ];
     /** @type {import("plugins").BuilderResources} */
     resources = {
-        on_mobile_preview_clicked_handlers: withSequence(
-            10,
-            this.onMobilePreviewClicked.bind(this)
-        ),
+        on_mobile_view_switched_handlers: withSequence(5, this.onMobileViewSwitched.bind(this)),
```
sobo-odoo: I think you forgot to change back the 5 to a 10 :slightly_smiling_face:

addons/html_builder/static/src/builder.js
```diff
@@ -25,7 +24,8 @@ import { getHtmlStyle } from "@html_editor/utils/formatting";
 import { isVisible } from "@html_builder/utils/utils";
 
 /**
- * @typedef {(() => void)[]} on_mobile_preview_clicked_handlers
+ * @typedef {((args: {isMobileView: boolean}) => ())[]} on_mobile_view_switched_handlers
+ * called on switch between viewing a small screen and a large screen
```
sobo-odoo: I would reformulate to something like this: `called when the screen size switches between mobile and desktop view` :slightly_smiling_face: 
It's more precise, as small and large screen doesn't mean much :+1:

addons/html_builder/static/src/builder.js
```diff
@@ -233,6 +234,17 @@ export class Builder extends Component {
                     ev.stopPropagation();
                 }
             };
+
+            let isMobileView = this.editor.config.isMobileView(this.editableEl);
```
sobo-odoo, line 252: Add a comment saying that we are using a resize observer to detect the switch between desktop and mobile view :slightly_smiling_face:

sobo-odoo (changes requested):
Last small comments :slightly_smiling_face:

addons/html_builder/static/src/core/visibility_plugin.js
blse-odoo: 10 is the default, I just removed the `withSequence`

sobo-odoo (approved):

@robodoo r+

PR #247836, opened 2026-02-09
[IMP] account: description field

malb-odoo:
This commit will add the description field in optional hide in the view of account_account.

task-5493859

---

### DIFF
--- a/addons/account/views/account_account_views.xml
+++ b/addons/account/views/account_account_views.xml
@@ -109,7 +109,8 @@
                     <field name="tax_ids" optional="hide" widget="many2many_tax_tags"/>
                     <field name="tag_ids" domain="[('applicability', '=', 'accounts')]" optional="hide" widget="many2many_tags"/>
                     <field name="currency_id" options="{'no_create': True}" groups="base.group_multi_currency"/>
-                    <field name="company_ids" widget="many2many_tags" readonly="True" groups="base.group_multi_company"/>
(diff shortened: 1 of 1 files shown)


addons/account/views/account_account_views.xml
```diff
@@ -109,7 +109,8 @@
                     <field name="tax_ids" optional="hide" widget="many2many_tax_tags"/>
                     <field name="tag_ids" domain="[('applicability', '=', 'accounts')]" optional="hide" widget="many2many_tags"/>
                     <field name="currency_id" options="{'no_create': True}" groups="base.group_multi_currency"/>
-                    <field name="company_ids" widget="many2many_tags" readonly="True" groups="base.group_multi_company"/>
+                    <field name="company_ids" optional="hide" widget="many2many_tags" readonly="True" groups="base.group_multi_company"/>
+                    <field name="description" optional="hide"/>
```
FlorianGilbert, line 113: This change means that we will have it everywhere, do we want this? I have the feeling in the task that it should be only present for invoice/bills.

FlorianGilbert (changes requested):

addons/account/views/account_account_views.xml
malb-odoo: Talked to olma about it and he seems ok with it but we can double check @Olivier-olma

FlorianGilbert: I'm not against it, I'm more questioning the fact that we are doing this in stable :smile:

PR #247838, opened 2026-02-09
[IMP] mail: standardize inconsistent Dropdown design

chgo-odoo:
This PR fixes inconsistent `Dropdown` that were used to select the input and output devices when starting a call within the `mail` module.

As this custom design was creating inconsistencies with the rest of the UI as well as accessibility flaws, we remove all the custom styling to rely on the default `Dropdown` component.

task-5921830

---

### DIFF
--- a/addons/mail/static/src/discuss/call/common/call_preview.xml
+++ b/addons/mail/static/src/discuss/call/common/call_preview.xml
@@ -28,13 +28,13 @@
         </div>
         <div t-if="props.hasSettingsAtBottom" class="d-flex flex-row justify-content-between mt-2 w-100 mw-100 flex-nowrap px-3">
             <div class="px-2" style="width: 30%">
-                <DeviceSelect icon="'fa fa-fw fa-microphone'" kind="'audioinput'" roundedType="'pill'"/>
+                <DeviceSelect icon="'fa fa-fw fa-microphone'" kind="'audioinput'"/>
             </div>
             <div class="px-2" style="width: 30%">
-                <DeviceSelect icon="'fa fa-fw fa-volume-up'" kind="'audiooutput'" roundedType="'pill'"/>
+                <DeviceSelect icon="'fa fa-fw fa-volume-up'" kind="'audiooutput'"/>
             </div>
             <div class="px-2" style="width: 30%">
-                <DeviceSelect icon="'fa fa-fw fa-video-camera'" kind="'videoinput'" roundedType="'pill'"/>
+                <DeviceSelect icon="'fa fa-fw fa-video-camera'" kind="'videoinput'"/>
             </div>
             <div class="d-flex align-items-center flex-grow-0 flex-shrink-0 px-2 text-muted">
                 <a role="button" t-on-click="onClickSettings" title="Advanced Settings">
--- a/addons/mail/static/src/discuss/call/common/device_select.xml
+++ b/addons/mail/static/src/discuss/call/common/device_select.xml
@@ -3,48 +3,42 @@
 
     <t t-name="discuss.CallDeviceSelect">
         <Dropdown
-            onStateChanged.bind="openSelect"
             disabled="!isPermissionGranted(props.kind)"
             openOnMouseEnter="false"
         >
             <button
                 t-if="isPermissionGranted(props.kind)"
-                class="o-mail-DeviceSelect-button btn cursor-pointer border-1 border-secondary align-items-center d-flex w-100 mw-100 shadow-sm o-dropdown--no-caret overflow-hidden"
+                t-attf-class="o-mail-DeviceSelect-button btn btn-secondary o-dropdown-caret {{ env.inDialog ? 'w-50' : 'w-100' }}"
                 t-att-data-kind="props.kind"
-                t-attf-class="{{ props.roundedType === 'pill' ? 'rounded-pill' : 'rounded' }}"
              >
                 <i t-if="props.icon" t-att-class="props.icon"/>
                 <span
-                    class="pe-2 d-inline-block text-truncate flex-grow-1 min-w-0"
+                    class="pe-2 text-start text-truncate"
                     t-attf-class="{{ props.icon ? 'ps-2' : '' }}"
                     t-esc="selectLabel || BROWSER_DEFAULT"/>
-                <span t-attf-class="oi smaller {{state.isSelectOpen ? 'oi-chevron-up': 'oi-chevron-down'}}"/>
             </button>
             <button
                 t-else=""
-                class="o-mail-DeviceSelect-button btn cursor-pointer border-1 border-secondary align-items-center d-flex w-100 mw-100 shadow-sm o-dropdown--no-caret"
+                t-attf-class="o-mail-DeviceSelect-button btn btn-secondary {{ env.inDialog ? 'w-50' : 'w-100' }}"
                 t-att-data-kind="props.kind"
-                t-attf-class="{{ props.roundedType === 'pill' ? 'rounded-pill' : 'rounded' }}"
                 t-on-pointerdown.prevent="() => this.showPermissionDialog(props.kind)"
                 t-on-keydown.prevent="() => this.showPermissionDialog(props.kind)"
             >
                 <i t-if="props.icon" t-att-class="props.icon"/>
                 <span
-                    class="d-inline-block text-truncate flex-grow-1 min-w-0"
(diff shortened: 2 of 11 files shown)


addons/mail/static/tests/discuss/call/call_settings_menu.test.js
```diff
@@ -64,12 +64,12 @@ test("Renders the call settings", async () => {
     rtc.microphonePermission = "granted";
     const browserDefaultLabel = isBrowserChrome() ? "Default" : "Browser Default";
     await click(".o-mail-DeviceSelect-button[data-kind='audioinput']:has(:text('Default'))");
-    await contains(".o-dropdown-item:has(:text('mockAudioDeviceLabel'))");
-    await contains(`.o-dropdown-item:has(:text(${browserDefaultLabel}))`);
+    await contains(".o-dropdown-item:contains('mockAudioDeviceLabel')");
+    await contains(`.o-dropdown-item:contains(${browserDefaultLabel})`);
```
seb-odoo: Why this change? We prefer to use exact match unless there is a strong reason.

seb-odoo (commented):
Hello, thank you for the clean up.

I think the idea is good, as AKU is not there this week if it can't wait feel free to move forward with it.

addons/mail/static/tests/discuss/call/call_settings_menu.test.js
seb-odoo: Probably just remove `:has` if the value is now directly on the node.

chgo-odoo: I'm not super familiar with these so I just used contains since the value is directly on the node indeed, I'll adapt 👌

chgo-odoo:
Hello @seb-odoo 👋 

I'd rather move it forward yes, as this this about accessibility 🙏  I kept the diff as minimal as possible especially since AKU is not there so we should be good !

Thanks !

chgo-odoo:
Hello @alexkuhn 👋 

May I ask for your review on this one, please ?

Thanks !

addons/mail/static/src/discuss/call/common/call_preview.xml
```diff
@@ -26,21 +26,13 @@
...
-                <DeviceSelect icon="'fa fa-fw fa-volume-up'" kind="'audiooutput'" roundedType="'pill'"/>
-            </div>
-            <div class="px-2" style="width: 30%">
-                <DeviceSelect icon="'fa fa-fw fa-video-camera'" kind="'videoinput'" roundedType="'pill'"/>
-            </div>
-            <div class="d-flex align-items-center flex-grow-0 flex-shrink-0 px-2 text-muted">
-                <a role="button" t-on-click="onClickSettings" title="Advanced Settings">
-                    <i class="fa fa-fw fa-gear"/>
-                </a>
-            </div>
```
alexkuhn, line 43: The `width: 30%`, while not perfect, had the benefit to have constant sizing of the dropdown.

Their removal make width change when changing selection, which is a regression:

Image
Image

Whereas before:

Image
Image

alexkuhn: Also in mobile we knew the preview of items was not great, but at least the footprint of these buttons were small enough to be fine.

With the removal, again, the sizing and wrapping change whenever an item is selected, and this is really too much for a mobile experience:

Image

Compared to previously:

Image

alexkuhn: By the way, the rounded design was to take inspiration from Google Meet, which FP is a fan and really expects Discuss meetings to look like Google Meet from the recent specs. While I understand the desire to have consistent UI, please note that the consistency of squarish button goes against following trends that are somewhat desired :/ Not a huge deal for me, but just wanted to let you know.

alexkuhn (commented):
LGTM otherwise.

addons/mail/static/src/discuss/call/common/call_preview.xml
chgo-odoo: > The width: 30%, while not perfect, had the benefit to have constant sizing of the dropdown.
> Their removal make width change when changing selection, which is a regression:

Now that the container is bigger we could probably restore it, but I don't think there is an issue regarding the mobile. The only other option would be to truncate the label and we should not, but as I'm mentioning below, this work is temporary, it only aims to temporarily "patch" the style

> By the way, the rounded design was to take inspiration from Google Meet, which FP is a fan and really expects Discuss meetings to look like Google Meet from the recent specs. While I understand the desire to have consistent UI, please note that the consistency of squarish button goes against following trends that are somewhat desired :/ Not a huge deal for me, but just wanted to let you know.

I do understand, I reverted it for now as I do prefer to "stick to the system", especially since this screen will be fine-tuned in task-5223747 so I do prefer to keep it standard (for now)

addons/mail/static/src/discuss/call/common/call.scss
```diff
@@ -104,3 +104,12 @@
...
 }
+
+.o-mail-WelcomePage .o-mail-DeviceSelect-button-large {
+    @include media-breakpoint-down(sm) {
+        --o-mail-CallPreview-deviceSelectWidth: 75%;
+    }
+
+    width: var(--o-mail-CallPreview-deviceSelectWidth, 30%);
+
+}
```
alexkuhn: While I understand the desire to make the input selection better in mobile, this solution bloats the welcome screen, when not many people would want to change these settings in mobile.

We knew the 3 buttons next to each other and very strong truncate was poor, but at least the welcome page is kept looking nice, and at least these button are easy to access when needed with minimal footprint if we don't need them:
Image

Now the new design makes the screen look more complex than it needs, for a mobile screen that acts as a Welcome page that is the 1st thing many people outside of Odoo would see, as Guest page, I think this is really not a good change:
Image

Again, I understand the desire to fix the text truncate, but I insist that this is not the good solution, and overall this is a worse screen. I'm all for fixing this, but if you want the other improvements for 19.2, I strongly advice to not rush making this change and instead make a dedicated task to improve this part.

chgo-odoo: I do think it is not okay to have a dropdown with one single letter visible and honestly I don't think we should sacrifice this for the sake of keeping the screen sleek but anyway I won't push too much for it, especially knowing there is another task to fix this screen, I'll adapt 👍

alexkuhn: I agree but at the same time the feature to select audio input / output / camera is useless 99% of the time in mobile. In fact the selection of input / output is most likely at OS level. The layout in mobile is more an after-thought than anything else, and to play safe we preferred to have it present even though the text is not visible.

The menu selection was made in 19.1 very close to the freeze, hence the desire to not take this much time in mobile, but at the same time keep it look good as this is a welcome page and it's important to make it look good... And also we were unsure of 1% of friction to not being able to select the devices, hence the compromise we made.

Yes this screen can be improved in its own task, specifically for mobile. We could think of collapsible container, single button to access all call settings... Or any other ideas, but please let's try to keep it welcoming as intended for a Welcome Page.

addons/mail/static/src/discuss/call/common/device_select.xml
```diff
@@ -3,48 +3,42 @@
...
     <t t-name="discuss.CallDeviceSelect">
         <Dropdown
-            onStateChanged.bind="openSelect"
             disabled="!isPermissionGranted(props.kind)"
             openOnMouseEnter="false"
         >
             <button
                 t-if="isPermissionGranted(props.kind)"
-                class="o-mail-DeviceSelect-button btn cursor-pointer border-1 border-secondary align-items-center d-flex w-100 mw-100 shadow-sm o-dropdown--no-caret overflow-hidden"
+                t-attf-class="o-mail-DeviceSelect-button btn btn-secondary o-dropdown-caret {{ props.isInDialog ? 'w-50' : 'w-100' }}"
```
alexkuhn: `props.isInDialog` can be replaced by `env.inDialog`: no need to define a new prop and make sure to pass it when the `env` works for any dialog ;)

alexkuhn (approved):
LGTM otherwise
@robodoo delegate+

addons/mail/static/src/discuss/call/common/device_select.xml
chgo-odoo: Thanks 😵‍💫

PR #247839, opened 2026-02-09
[FIX] mrp: avoid recursion error while producing large quantities

MaximeNoirhomme229:
**Issue**
When producing a large number of serial-tracked products, a RecursionError can occur.

**Steps to reproduce**
- Create three products tracked by serial number (ensure MTO and Manufacture routes are enabled).
- Create a BoM for product A containing product B.
- Create a BoM for product B containing product C.
- Create a BoM for product C containing another product.
- Create a manufacturing order of 100 units for product A and confirm it.
- Go to the MO C and split into 100 mo
- Go to the MO B and split into 100 mo
-> RecursionError: maximum recursion depth exceeded.

**Cause**
While splitting, this method is called:
https://github.com/odoo/odoo/blob/cda011dc8590773f6c3a26f4ae9d5242a3147024/addons/mrp/models/mrp_production.py#L2031 which ultimately calls:
https://github.com/odoo/odoo/blob/cda011dc8590773f6c3a26f4ae9d5242a3147024/addons/stock/models/stock_move.py#L658-L661 This retriggers `_compute_packaging_uom_id` for all moves in `move_orig_ids` or `move_dest_ids`, and accessing the full recordsets causes recursive recomputation leading to a RecursionError.

opw-5265424

### DIFF
--- a/addons/stock/models/stock_move.py
+++ b/addons/stock/models/stock_move.py
@@ -191,7 +191,7 @@ class StockMove(models.Model):
     forecast_expected_date = fields.Datetime('Forecasted Expected date', compute='_compute_forecast_information', compute_sudo=True)
     lot_ids = fields.Many2many('stock.lot', compute='_compute_lot_ids', inverse='_set_lot_ids', string='Serial Numbers', readonly=False)
     reservation_date = fields.Date('Date to Reserve', compute='_compute_reservation_date', store=True, help="Computes when a move should be reserved")
-    packaging_uom_id = fields.Many2one('uom.uom', 'Packaging', help="Packaging unit from sale or purchase orders", compute='_compute_packaging_uom_id', precompute=True, recursive=True, store=True)
+    packaging_uom_id = fields.Many2one('uom.uom', 'Packaging', help="Packaging unit from sale or purchase orders", compute='_compute_packaging_uom_id', precompute=True, store=True)
     packaging_uom_qty = fields.Float('Packaging Quantity', help="Quantity in the packaging unit", compute='_compute_packaging_uom_qty', store=True)
     show_quant = fields.Boolean("Show Quant", compute="_compute_show_info")
     show_lots_m2o = fields.Boolean("Show lot_id", compute="_compute_show_info")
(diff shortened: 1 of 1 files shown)


lase-odoo (commented):
Thanks a lot for the proposition ! 🦐

While the initial fix proposition of modifying `_compute_packaging_uom_id` to only look at the `move.move_orig_ids[0].packaging_uom_id` and `move.move_dest_ids[0].packaging_uom_id` would avoid the recursion depth error in your case, if we are careful the compute method only ever assign a value via the else condition and it only hapends when we ended up trying to get the value of `packaging_uom_id` of a move that we tried to compute above in the stack (because all the moves in `move_orig_ids` have a set of `move_dest_ids` (containing yourself)). And, in that case we will assign the `move.product_uom` of the first move that ended up in that situation, this looks quite order dependent and a bit random. I would simply remove this recursion dependence since the packaging is already correctly propagated via procurement at move creations and via the sale and purchase overrides when the `packaging_uom_id` really matters:
https://github.com/odoo/odoo/blob/f5da5197620ad12302b40ff1702a9cda5007aaa3/addons/stock/models/stock_move.py#L1710
https://github.com/odoo/odoo/blob/8ef673d27481d17b83f93892e78b09b085738ca7/addons/sale_stock/models/stock.py#L19-L24
https://github.com/odoo/odoo/blob/8ef673d27481d17b83f93892e78b09b085738ca7/addons/purchase_stock/models/stock_move.py#L32-L37
I therefore think it is best to remove all the recursive dependencies and simplify the compute as much as possible.

Note that this suggestion even improve the use case of receipt 2 in `test_scan_packaging_on_picking_with_mixed_uom` as it is currently losing the information of the `packaging_uom_id` set to a pack of 6 at validation since the `move_dest_ids` is set to `[]` during the `push_apply` triggering the dependencies of the compute for no valid reasons:
https://github.com/odoo/enterprise/blob/ac186aa71cd7e1b80b307ea12c7eaca246afd649/stock_barcode/tests/test_barcode_client_action_picking.py#L3940-L3956

MaximeNoirhomme229:
@lase-odoo Thanks for the review!

PR #247846, opened 2026-02-09
[FIX] pos_sale: prevent subscription_discount traceback

artn-odoo:
**Steps to reproduce:**
- Make a subscription product
- Make a quotation with it, confirm it, then invoice it
- Go back to the sale order and upsell it
- Add another product
- Go to the PoS to settle the order
- A traceback appears

**Why the fix:**
Why tried to treat the informative line that says that this is a discount as a normal pos order line. We then tried to access the line's template, which caused a crash as the line's template was undefined.

We now treat the line as we do a note, meaning to add it to the previous line in the order.

We create a function to check if the line is a note and we override it in the **pos_sale_subscription** module.
Enterprise PR: https://github.com/odoo/enterprise/pull/107002
opw-5582448

### DIFF
--- a/addons/pos_sale/static/src/app/services/pos_store.js
+++ b/addons/pos_sale/static/src/app/services/pos_store.js
@@ -98,7 +98,7 @@ patch(PosStore.prototype, {
         ]);
 
         for (const line of sale_order.order_line) {
-            if (line.display_type === "line_note") {
+            if (this.isSaleOrderLineNote(line)) {
                 if (previousProductLine) {
                     const previousNote = previousProductLine.customer_note;
                     previousProductLine.customer_note = previousNote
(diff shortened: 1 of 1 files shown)


addons/pos_sale/static/src/app/services/pos_store.js
```diff
@@ -98,7 +98,7 @@ patch(PosStore.prototype, {
         ]);
 
         for (const line of sale_order.order_line) {
-            if (line.display_type === "line_note") {
+            if (["line_note", "subscription_discount"].includes(line.display_type)) {
```
stva-odoo: I suggest adding a method line.isNote(), or alternatively a method in pos_store such as isSaleOrderLineNote(orderLine).
Then override it in the pos_sale_subscription module.

stva-odoo:
robodoo r+

PR #247874, opened 2026-02-09
[FIX] product: remove redundant HTML head in product labels

gergoe:
Both product.report_productlabel and product.report_productlabel_dymo are called from a template which already includes the necessary DOM for the HTML (via web.basic_layout), having a web.html_container call in these two templates seems to be redundant and generates a malformed HTML output.

Current behavior before PR:
Duplicate/redundant DOM structure in reports generated when printing product labels

Correct DOM structure, valid HTML

---

### DIFF
--- a/addons/product/report/product_product_templates.xml
+++ b/addons/product/report/product_product_templates.xml
@@ -127,83 +127,79 @@
         </template>
 
         <template id="report_productlabel">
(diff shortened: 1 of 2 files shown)


Sehnde (approved):
LGTM

PR #247918, opened 2026-02-10
[FIX] im_livechat: do not show leave warning on closed chatbot conversation

akth-odoo:
Before this commit:
When a user finishes a chatbot script and the conversation is already ended, clicking on close / continue still triggers the leave conversation warning.

After this commit:
The leave conversation warning is no longer shown when the chatbot conversation is already closed or ended. The warning is only shown for active conversations.

[Task-5882084](https://www.odoo.com/odoo/project/1519/tasks/5882084)
---

### DIFF
--- a/addons/im_livechat/controllers/chatbot.py
+++ b/addons/im_livechat/controllers/chatbot.py
@@ -62,7 +62,23 @@ class LivechatChatbotScriptController(http.Controller):
             # sudo: visitor cannot write on channel otherwise. Just writing a
             # boolean is safe
             discuss_channel.sudo().livechat_active = False
-            return None
+            step_message = next(
+                # sudo - chatbot.message.id: visitor can access chat bot messages.
+                m.mail_message_id for m in discuss_channel.sudo().chatbot_message_ids
+                if m.script_step_id == current_step
+                and m.mail_message_id.author_id == chatbot.operator_partner_id
+            )
+            store = Store(discuss_channel)
+            store.add(
+                "ChatbotStep",
+                {
+                    "id": (current_step.id, step_message.id),
+                    "scriptStep": current_step.id,
+                    "message": step_message.id,
+                    "isLast": True,
+                }
+            )
+            return store.get_result()
 
         posted_message = next_step._process_step(discuss_channel)
         store = Store(posted_message, for_current_user=True)
@@ -71,7 +87,6 @@ class LivechatChatbotScriptController(http.Controller):
             "ChatbotStep",
             {
                 "id": (next_step.id, posted_message.id),
-                "isLast": next_step._is_last_step(discuss_channel),
                 "message": Store.one(posted_message, only_id=True),
                 "operatorFound": next_step.step_type == "forward_operator"
                 and len(discuss_channel.channel_member_ids) > 2,
--- a/addons/im_livechat/static/src/embed/common/chat_window_patch.js
+++ b/addons/im_livechat/static/src/embed/common/chat_window_patch.js
@@ -20,7 +20,11 @@ patch(ChatWindow.prototype, {
 
     async close() {
         const chatWindow = toRaw(this.props.chatWindow);
-        if (chatWindow.thread.id > 0 && !this.livechatState.showCloseConfirmation) {
+        if (
+            chatWindow.thread.id > 0 &&
+            chatWindow.thread.livechat_active &&
+            !this.livechatState.showCloseConfirmation
+        ) {
             this.state.actionsDisabled = true;
             this.livechatState.showCloseConfirmation = true;
         } else {
--- a/addons/im_livechat/static/src/embed/common/chatbot/chatbot_model.js
+++ b/addons/im_livechat/static/src/embed/common/chatbot/chatbot_model.js
@@ -83,8 +83,7 @@ export class Chatbot extends Record {
 
     get completed() {
         return (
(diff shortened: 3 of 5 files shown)


addons/im_livechat/controllers/chatbot.py
```diff
@@ -62,7 +62,8 @@ def chatbot_trigger_step(self, channel_id, chatbot_script_id=None):
             # sudo: visitor cannot write on channel otherwise. Just writing a
             # boolean is safe
             discuss_channel.sudo().livechat_active = False
-            return None
+            store = Store(discuss_channel, {"livechat_active": False})
+            return store.get_result()
```
ramh-odoo: It could be done inline without creating a new variable.

addons/im_livechat/controllers/chatbot.py
```diff
@@ -62,7 +62,7 @@ def chatbot_trigger_step(self, channel_id, chatbot_script_id=None):
             # sudo: visitor cannot write on channel otherwise. Just writing a
             # boolean is safe
             discuss_channel.sudo().livechat_active = False
-            return None
+            return Store(discuss_channel, {"livechat_active": False}).get_result()
```
ramh-odoo: The value of `livechat_active` has already been updated above. Why are we updating it again in the store here?
Maybe we could simply return the `get_result` from the store with the discuss channel and only pass this field if it’s actually needed?

addons/im_livechat/static/src/embed/common/chatbot/chatbot_model.js
```diff
@@ -94,19 +93,20 @@ export class Chatbot extends Record {
...
                 chatbot_script_id: this.script.id,
             });
-            if (!storeData) {
+            const data = this.store.insert(storeData);
+            if (!storeData.ChatbotStep) {
                 this.currentStep.isLast = true;
                 return;
             }
-            const { ChatbotStep: steps } = this.store.insert(storeData, { html: true });
+            const { ChatbotStep: steps } = data;
```
ramh-odoo: Why do we need extra variable here even if it is used once?

akth-odoo: We need that variable because we can’t initialize `steps` when there is no `ChatbotStep` in the response.

ramh-odoo (approved):

addons/im_livechat/static/src/embed/common/chat_window_patch.js
```diff
@@ -20,7 +20,11 @@ patch(ChatWindow.prototype, {
 
     async close() {
         const chatWindow = toRaw(this.props.chatWindow);
-        if (chatWindow.thread.id > 0 && !this.livechatState.showCloseConfirmation) {
+        if (
+            chatWindow.thread.id > 0 &&
+            !this.livechatState.showCloseConfirmation &&
+            chatWindow.thread.livechat_active
+        ) {
```
alexkuhn, line 27: Nitpick: I would swap the position of `showCloseConfirmation` and `livechat_active`. This would be more natural to read the expression in that order.

This would read as:
- a persistent conversation
- that is active
- without an ongoing show close confirmation

addons/im_livechat/static/src/embed/common/chatbot/chatbot_model.js
```diff
@@ -94,19 +93,19 @@ export class Chatbot extends Record {
...
         }
         if (this.steps.at(-1)?.eq(this.currentStep)) {
             const storeData = await rpc("/chatbot/step/trigger", {
                 channel_id: this.thread.id,
                 chatbot_script_id: this.script.id,
             });
-            if (!storeData) {
+            const { ChatbotStep: steps } = this.store.insert(storeData, { html: true });
+            if (!storeData.ChatbotStep) {
                 this.currentStep.isLast = true;
```
alexkuhn: Instead of using heuristics that there's no chat bot step in the insert of last step, prefer to pass the chat bot step data that has the `isLast`.
This is more robust to rely on presence of data with strong semantics rather than the absence of data. Absence of data is best reserved for performance, i.e. being efficient to provide only relevant data.

addons/im_livechat/static/src/embed/common/chatbot/chatbot_model.js
```diff
@@ -94,19 +93,18 @@ export class Chatbot extends Record {
...
         }
         if (this.steps.at(-1)?.eq(this.currentStep)) {
             const storeData = await rpc("/chatbot/step/trigger", {
                 channel_id: this.thread.id,
                 chatbot_script_id: this.script.id,
             });
-            if (!storeData) {
-                this.currentStep.isLast = true;
+            const { ChatbotStep: steps } = this.store.insert(storeData, { html: true });
+            if (storeData.ChatbotStep.isLast) {
```
alexkuhn: `storeData.ChatBotStep` are raw data sent by the server. We should not assert on them: always insert the data and then have logic around inserted data, as working with records is easier that these raw data.

Here I think it should be
```js
if (this.currentStep.isLast) {
    return;
}
```
As insertion of data of the steps would flag the current step as `isLast`.

alexkuhn (approved):
@robodoo delegate+ r+ override=ci/security
`html: true` to insert safe message body html content. In 18.3+ this is improved to not require `html: true` anyway.

PR #247920, opened 2026-02-10
[FIX] website: fix domain allocation on employee documents smart button

pajo-odoo:
Steps to reproduce:
---------------------------------
1. Install `documents_hr` and `website_documents` modules
2. Go to website > configuration > websites
3. In My Website set any arbitary domain (e.g. https://test.com)
4. Open any employee record
5. Click on Documents smart button

Observation:
---------------------------------
It will try to open employee's documents with the website's domain, e.g. `https://test.com/odoo/documents/xyz`

Issue:
---------------------------------
After the following commit:
odoo@46c43c1

the smart button redirects to the document folder via an access token. The `access_url` is computed using `get_base_url()`, which is overridden by the website module to return the website domain instead of the system base URL.
https://github.com/odoo/odoo/blob/f6cf0d067e5f30e2b22ea513071cd7c5e3d9f44c/addons/website/models/ir_model.py#L10-L36

Solution:
---------------------------------
Added a context-based check to `get_base_url()`.
When the context key `use_config_parameter_domain` is set, and the record has a `website_id` field, the system base URL from the configuration parameters is used instead of the website domain. This allows any model to explicitly rely on the configured base URL when required

NOTE: No module installs `hr`, `documents` and `website`, so test case is not possible without bridge module of all three

Related Enterprise PR: https://github.com/odoo/enterprise/pull/106478
opw-5471683

### DIFF
--- a/addons/website/models/ir_model.py
+++ b/addons/website/models/ir_model.py
@@ -26,6 +26,11 @@ class Base(models.AbstractModel):
             return super().get_base_url()
         self.ensure_one()
 
+        # When a model has a `website_id`, calling `get_base_url()` always returns
+        # the website domain. This context key acts as a gateway to bypass that
+        # behavior and use the web.base.url instead.
+        if self.env.context.get('use_config_parameter_domain'):
+            return super().get_base_url()
         if self._name == 'website':
             # Note that website_1.company_id.website_id might not be website_1
             return self.domain or super().get_base_url()


bhra-odoo (approved):

nle-odoo (commented):
maybe this could be good if there was other use case where we want to skip website get_base_url

but since it seems to be the only case where we currently wanted that maybe rather than adding complexity to `get_base_url` it might be ok in this case to user a context key rather in `documents_document()._compute_access_url` and based on this context prefix a domain or not (and just return "/path" as URL without domain)

because the ir.config_parameter could be different than the one of the current website too (it's just the URL of the last administrator that logged in the database, unless there is a `web.base.url.freeze` parameter to keep it from changing) and so the advantage of this PR (not needing to reconnect if we were connected on another domain) could not exist in this case

side note:

on the ticket, the two comments about verifying the bug were doing misconfiguration things in my opinion:

- putting a wrong domain on the website => the domain should be pointing to the odoo instance

- putting a wrong domain with query parameter like "https://www.youtube.com/watch?v=nbso3NVz3p8" => this is not a domain but a full URL that is not hosting odoo, of course this will not work and is a misconfiguration

but even with good configuration, it would probably make sense to not make it needed to change domain (and reconnect) for this button

> After the following commit:
> `46c43c1`

the link does not exist (because it is a non-merged commit in the enterprise repository), you should refer to the merged commit (it's better to ensure the link works forever) and refer to enterprise:

```
odoo/enterprise@954d3ef8009e8d79d9b8c5b52c297de2c8ab794a
```

PR #247923, opened 2026-02-10
[CLN] html_editor, website: remove duplicate height utility classes

rahg-odoo:
Remove the redundant 'o_full_screen_height' and related height utility
classes from the 'website' SCSS, as they are already defined in the
'html_editor' SCSS and do not need to be duplicated.

task-[5123275](https://www.odoo.com/odoo/project.task/5123275)

---

### DIFF
--- a/addons/html_editor/static/src/scss/html_editor.common.scss
+++ b/addons/html_editor/static/src/scss/html_editor.common.scss
@@ -722,14 +722,17 @@ pre {
     background-image: none !important;
 }
 
-.o_full_screen_height {
+@mixin o_full_screen_height {
     display: flex;
     flex-direction: column;
     justify-content: space-around;
     min-height: 100vh !important;
 }
+.o_full_screen_height {
+    @include o_full_screen_height;
+}
 .o_half_screen_height {
-    @extend .o_full_screen_height;
+    @include o_full_screen_height;
     min-height: 55vh !important;
 }
 
--- a/addons/website/static/src/scss/website.scss
+++ b/addons/website/static/src/scss/website.scss
@@ -782,13 +782,6 @@ font[class*='bg-'] {
         box-sizing: content-box;
     }
 }
-// Smaller container
-.o_container_small {
-    @extend .container;
-    @include media-breakpoint-up(lg) {
-        max-width: map-get($container-max-widths, md);
-    }
-}
 
 // Buttons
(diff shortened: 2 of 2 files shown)


addons/website/static/src/scss/website.scss
```diff
@@ -942,26 +942,10 @@ table.table_desc tr td {
...
-    min-height: 55vh !important;
-}
-
-// TODO remove cover_full and cover_mid classes (kept for compatibility for now)
-.cover_full {
-    @extend .o_full_screen_height;
-}
-.cover_mid {
-    @extend .o_half_screen_height;
-}
```
aans-odoo, line 990: `.o_container_small` is also duplicated.

addons/html_editor/static/src/scss/html_editor.common.scss
```diff
@@ -722,20 +722,23 @@ pre {
...
 .o_half_screen_height {
-    @extend .o_full_screen_height;
+    @include o_full_screen_height;
     min-height: 55vh !important;
 }
 
 // TODO remove cover_full and cover_mid classes (kept for compatibility for now)
 .cover_full {
-    @extend .o_full_screen_height;
+    @include o_full_screen_height;
```
sobo-odoo: It should stay `@extend .o_full_screen_height`, because otherwise it only uses the mixin with the `min-height: 100vh`, but we want it to use `100svh` when we are in website, so it should still extend it :slightly_smiling_face:

addons/html_editor/static/src/scss/html_editor.common.scss
```diff
@@ -722,20 +722,23 @@ pre {
     background-image: none !important;
```
sobo-odoo, line 722: Commit message:
- you didn't add `html_editor` as a module in the commit title :+1: 
- The char limit is 72 chars per line, but you have to fill them as much as possible to reach 72 (without cutting words of course) :slightly_smiling_face: (here it seems that you limited them at 70)
- Nitpicks:
  - There is only one spec so no need for a bullet
  - You can remove the "specification :", the text after it is enough :+1:

sobo-odoo (changes requested):

sobo-odoo (approved):
Thanks for the cleaning :slightly_smiling_face: 

@robodoo r+

PR #247924, opened 2026-02-10
[IMP] mail, web: add delete and add to documents actions in file viewer

akth-odoo:
Users can currently delete attachments or add them to Documents from the chatter. However, when opening an attachment in the file viewer, these actions are not available.

This commit adds Delete and Add to Documents buttons in the file viewer, allowing users to perform the same actions without leaving the viewer.

Related Enterprise PR: https://github.com/odoo/enterprise/pull/107467

image

Task-[5408185](https://www.odoo.com/odoo/project/1519/tasks/5408185) 
---

### DIFF
--- a/addons/mail/static/src/core/common/attachment_list.js
+++ b/addons/mail/static/src/core/common/attachment_list.js
@@ -96,26 +96,36 @@ export class AttachmentList extends Component {
      */
     onClickUnlink(attachment) {
         if (this.env.inComposer) {
-            return this.props.unlinkAttachment(attachment);
+            this.props.unlinkAttachment(attachment);
+            return true;
         }
         if (this.hasUnlinkConfirmation(attachment)) {
-            this.dialog.add(ConfirmationDialog, {
-                title: _t("Delete Attachment"),
-                body: _t(
-                    'Are you sure you want to delete "%s"?\nThis action cannot be undone.',
-                    attachment.name
-                ),
-                confirmLabel: _t("Delete Attachment"),
-                cancel: () => {},
-                confirm: () => this.onConfirmUnlink(attachment),
+            return new Promise((resolve) => {
+                this.dialog.add(ConfirmationDialog, {
+                    title: _t("Delete Attachment"),
+                    body: _t(
+                        'Are you sure you want to delete "%s"?\nThis action cannot be undone.',
+                        attachment.name
+                    ),
+                    confirmLabel: _t("Delete Attachment"),
+                    cancel: () => resolve(false),
+                    confirm: () => {
+                        this.onConfirmUnlink(attachment);
+                        resolve(true);
+                    },
+                });
             });
         } else {
             this.onConfirmUnlink(attachment);
+            return true;
         }
     }
 
     onClickAttachment(attachment) {
-        this.fileViewer.open(attachment, this.props.attachments);
+        this.fileViewer.open(attachment, this.props.attachments, {
+            onUnlink: this.onClickUnlink.bind(this),
+            canUnlink: (file) => this.showDelete(file),
+        });
     }
 
     /**
--- a/addons/web/static/src/core/file_viewer/file_viewer.js
+++ b/addons/web/static/src/core/file_viewer/file_viewer.js
@@ -34,6 +34,8 @@ export class FileViewer extends Component {
         startIndex: t.any(),
         close: t.any().optional(),
         modal: t.any().optional(true),
+        onUnlink: t.function().optional(),
+        canUnlink: t.function().optional(() => () => false),
     });
 
     autofocusRef = signal(null);
@@ -320,6 +322,10 @@ export class FileViewer extends Component {
         this.zoomerRef().style = "transform: " + `translate(${tx}px, ${ty}px)`;
     }
 
+    get canUnlink() {
+        return this.props.onUnlink && this.props.canUnlink(this.state.file);
+    }
+
     get imageStyle() {
         let style =
             "transform: " +
@@ -333,6 +339,12 @@ export class FileViewer extends Component {
         }
         style += `background: repeating-conic-gradient(#ccc 0deg 90deg, #fff 90deg 180deg) 50% / 20px 20px;`;
         return style;
+    }
+
+    async onClickUnlink() {
+        if (await this.props.onUnlink(this.state.file)) {
+            this.close();
+        }
     }
 
     onClickPrint() {
--- a/addons/web/static/src/core/file_viewer/file_viewer.xml
+++ b/addons/web/static/src/core/file_viewer/file_viewer.xml
@@ -21,6 +21,9 @@
                             <span>Download</span>
                         </div>
                     </div>
+                    <div t-if="this.canUnlink" class="o-FileViewer-headerButton d-flex align-items-center px-3 cursor-pointer" role="button" title="Remove" aria-label="Remove" t-on-click.stop="this.onClickUnlink">
+                        <i class="fa fa-fw fa-trash text-danger" role="img"/>
+                    </div>
 
                     <div t-on-click.stop="this.close" class="o-FileViewer-headerButton d-flex align-items-center mb-0 px-3 h4 text-reset cursor-pointer" role="button" title="Close (Esc)" aria-label="Close">
                         <i class="fa fa-fw fa-times" role="img"/>
--- a/addons/web/static/src/core/file_viewer/file_viewer_hook.js
+++ b/addons/web/static/src/core/file_viewer/file_viewer_hook.js
@@ -3,6 +3,8 @@ import { registry } from "@web/core/registry";
 import { useService } from "../utils/hooks";
 import { FileViewer } from "./file_viewer";
 
+/** @typedef {import("@web/core/file_viewer/file_viewer").File} File */
+
 const fileViewerService = {
     dependencies: ["overlay"],
     /**
@@ -19,10 +21,13 @@ const fileViewerService = {
             }
 
             /**
-             * @param {import("@web/core/file_viewer/file_viewer").File} file
-             * @param {import("@web/core/file_viewer/file_viewer").File[]} files
+             * @param {File} file the file to open in the viewer
+             * @param {File[]} files
+             * @param {Object} [options]
+             * @param {(File) => boolean} [options.canUnlink]
+             * @param {(File) => boolean} [options.onUnlink] called when the user requests to unlink the file. * Returns `true` on success, allowing the file viewer to close.
              */
-            function open(file, files = [file]) {
+            function open(file, files = [file], { canUnlink, onUnlink } = {}) {
                 close();
                 if (!file.isViewable) {
                     return;
@@ -35,6 +40,8 @@ const fileViewerService = {
                         {
                             files: viewableFiles,
                             startIndex: index,
+                            canUnlink,
+                            onUnlink,
                             close,
                         },
                         { rootId: ref?.()?.getRootNode()?.host?.id }
--- a/addons/web/static/tests/core/file_viewer.test.js
+++ b/addons/web/static/tests/core/file_viewer.test.js
@@ -1,0 +1,28 @@
+import { expect, test, waitFor } from "@odoo/hoot";
+import { click } from "@odoo/hoot-dom";
+import { mountWithCleanup } from "@web/../tests/web_test_helpers";
+
+import { FileModel } from "@web/core/file_viewer/file_model";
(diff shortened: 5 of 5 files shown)


ramh-odoo (changes requested):
I haven’t reviewed the PR yet, but functionally it’s not behaving as expected. Currently, it allows all channel members (button visible for guests as well) to delete attachments from other users’ messages, which shouldn’t be allowed.

Image

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -350,4 +353,17 @@ export class FileViewer extends Component {
...
+    onClickUnlink() {
+        this.dialog.add(ConfirmationDialog, {
+            title: _t("Delete Attachment"),
+            body: _t(
+                'Are you sure you want to delete "%s"?\nThis action cannot be undone.',
+                this.state.file.name
+            ),
+            confirmLabel: _t("Delete Attachment"),
+            cancel: () => {},
+            confirm: () => this.state.file.remove(),
```
ramh-odoo: This doesn’t work when opening file preview from the Documents app.

Image

addons/web/static/src/core/file_viewer/file_viewer.xml
```diff
@@ -56,6 +56,9 @@
                     <div class="o-FileViewer-toolbarButton p-3 rounded-0" t-on-click.stop="onClickPrint" title="Print" role="button">
                         <i class="fa fa-fw fa-print" role="img"/>
                     </div>
+                    <div t-if="state.file.isDeletable" class="o-FileViewer-toolbarButton p-3 rounded-0" t-on-click.stop="onClickUnlink" title="Remove" role="button">
```
ramh-odoo: same here. Don’t forget to add `this.` for the preparation of Owl 3.

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -29,7 +31,7 @@ const IMAGE_BUFFER_PADDING = 20;
 export class FileViewer extends Component {
     static template = "web.FileViewer";
     static components = {};
-    static props = ["files", "startIndex", "close?", "modal?"];
+    static props = ["files", "startIndex", "close?", "modal?", "onUnlink?", "inComposer?"];
```
ramh-odoo: I would take this opportunity to sort them alphabetically.

addons/web/static/src/core/file_viewer/file_viewer_hook.js
```diff
@@ -12,8 +12,10 @@ const fileViewerService = {
             /**
              * @param {import("@web/core/file_viewer/file_viewer").FileViewer.props.files[]} file
              * @param {import("@web/core/file_viewer/file_viewer").FileViewer.props.files} files
+             * @param {{onUnlink?: boolean}} [options={}]
+             * @param {{inComposer?: boolean}} [options={}]
              */
-            function open(file, files = [file]) {
+            function open(file, files = [file], { onUnlink, inComposer } = {}) {
```
ramh-odoo: Since `inComposer` only exists in mail, wouldn't it make more sense to patch it there if needed?

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -325,6 +325,13 @@ export class FileViewer extends Component {
         return style;
     }
 
+    async onClickUnlink() {
+        const confirmed = await this.props.onUnlink(this.state.file);
```
ramh-odoo: `onUnlink` is optional props.

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -335,6 +336,13 @@ export class FileViewer extends Component {
         return style;
     }
 
+    async onClickUnlink() {
+        const confirmed = await this.props?.onUnlink(this.state.file);
```
jeanschoenlaub: `this.props?.onUnlink`  => `this.props.onUnlink`, no need for `?` as props always exists

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -335,6 +336,13 @@ export class FileViewer extends Component {
...
+        if (confirmed) {
+            this.close();
+        }
+    }
```
jeanschoenlaub, line 348: I would do:
```
async onClickUnlink() {
    if (await this.props.onUnlink(this.state.file)) {
        this.close();
    }
}
```

addons/web/static/src/core/file_viewer/file_viewer.xml
```diff
@@ -21,6 +21,9 @@
                             <span>Download</span>
                         </a>
                     </div>
+                    <div t-if="this.props.onUnlink" class="o-FileViewer-headerButton d-flex align-items-center px-3 cursor-pointer" role="button" title="Remove" aria-label="Remove" t-on-click.stop="this.onClickUnlink">
+                        <i class="fa fa-fw fa-trash" role="img"/>
+                    </div>
```
jeanschoenlaub: There is a potential for bug in this code, and your current implementation has a bug I believe:

--> If you open the file viewer on a file you can delete (`this.props.onUnlink` is true) and then navigate to another file you can't delete, the delete button will still appear (the file viewer props have not changed, its `this.state.file` which changed on navigation)

I would suggest managing the `showDelete` logic per file. Something like this:
```
// attachment_list.js
onClickAttachment(attachment) {
    this.fileViewer.open(attachment, this.props.attachments, {
        onUnlink: this.onClickUnlink.bind(this),
        canUnlink: (file) => this.showDelete(file),
    });
}

// file_viewer.js
get canUnlink() {
    return this.props.onUnlink && (this.props.canUnlink?.(this.state.file) ?? false);
}

<!-- file_viewer.xml -->
<div t-if="this.canUnlink" ...>
```

Note: This is not super scalable: a better approach would be to use a registry of actions (see for eg. `userMenuRegistry`) but I don't think this PR would be the place to do it

jeanschoenlaub:
Hey Akshay,

I did a quick first pass on the web part *only*, and I am not sure about those changes. Could you double check the bug I think exist is real and if yes implement a fix for the web part.

--> In any cases once everything is final please tag `aab-odoo` for a final review of the web part please. Thanks :)

akth-odoo:
@aab-odoo Could you please review this PR?

addons/web/static/src/core/file_viewer/file_viewer.js
```diff
@@ -320,6 +322,10 @@ export class FileViewer extends Component {
         this.zoomerRef.el.style = "transform: " + `translate(${tx}px, ${ty}px)`;
     }
 
+    get canUnlink() {
+        return this.props.onUnlink && (this.props.canUnlink?.(this.state.file) ?? false);
```
aab-odoo: I would define a defaultProps for `canUnlink`: `() => false` which would then allow to simplify this:

```suggestion
        return this.props.onUnlink && this.props.canUnlink(this.state.file);
```

aab-odoo: a unit test case for this new feature would be nice. There doesn't seem to be a file_viewer.test.js file yet, but you can create it :-)

alexkuhn: > There doesn't seem to be a file_viewer.test.js file yet, but you can create it :-)

There are few tests in `@mail` on the FileViewer from when this was in `@mail`, but feel free to make tests focused very specifically on the file viewer itself ;)

addons/web/static/src/core/file_viewer/file_viewer_hook.js
```diff
@@ -21,8 +21,11 @@ const fileViewerService = {
             /**
              * @param {import("@web/core/file_viewer/file_viewer").File} file
              * @param {import("@web/core/file_viewer/file_viewer").File[]} files
+             * @param {Object} [options]
+             * @param {Function} [options.canUnlink]
+             * @param {Function} [options.onUnlink]
```
alexkuhn: Would document the importance of return value, especially on `onUnlink`.
I can imagine mistakes where `undefined` is implicitly passed and this prevents the good closing of the file viewer.

Also document the `File` param could be quite useful too.

addons/web/static/src/core/file_viewer/file_viewer.xml
```diff
@@ -21,6 +21,9 @@
                             <span>Download</span>
                         </div>
                     </div>
+                    <div t-if="this.canUnlink" class="o-FileViewer-headerButton d-flex align-items-center px-3 cursor-pointer" role="button" title="Remove" aria-label="Remove" t-on-click.stop="this.onClickUnlink">
+                        <i class="fa fa-fw fa-trash" role="img"/>
```
alexkuhn: Nitpick: I feel that `text-danger` on icon shows better this is a destructive action, and overall improves the readability of actions instantaneously.

addons/web/static/src/core/file_viewer/file_viewer.js
alexkuhn, line 28: If this was an object, we could have put it in the props validation.
This could be considered at some point, as this is quite nice to be helped by props validation / typing rather than generic `Function` that we have to guess what's accepted / what are the expected returned value types.

addons/mail/static/src/core/common/attachment_list.js
```diff
@@ -92,22 +92,30 @@ export class AttachmentList extends Component {
      */
     onClickUnlink(attachment) {
         if (this.env.inComposer) {
-            return this.props.unlinkAttachment(attachment);
+            return this.props.unlinkAttachment(attachment).then(() => true);
         }
```
alexkuhn, line 101: Nitpick: I slightly prefer turning this function `async` and use `await X; return true;`.
Returned value is made more explicit that way.

alexkuhn:
@robodoo delegate+

addons/web/static/src/core/file_viewer/file_viewer_hook.js
```diff
@@ -19,10 +19,13 @@ const fileViewerService = {
             }
 
             /**
-             * @param {import("@web/core/file_viewer/file_viewer").File} file
+             * @param {import("@web/core/file_viewer/file_viewer").File} file the file to open in the viewer
              * @param {import("@web/core/file_viewer/file_viewer").File[]} files
+             * @param {Object} [options]
+             * @param {Function} [options.canUnlink]
+             * @param {Function} [options.onUnlink] called when the user requests to unlink the file. Returns * (or resolves to) `true` on success, allowing the file viewer to close.
```
alexkuhn:
```suggestion
             * @param {(import("@web/core/file_viewer/file_viewer").File) => boolean} [options.canUnlink]
             * @param {(import("@web/core/file_viewer/file_viewer").File} => boolean [options.onUnlink] called when the user requests to unlink the file. Returns `true` on success, allowing the file viewer to close.
```

We can consider define `File` as a typedef:

```js
/** @typedef {import("@web/core/file_viewer/file_viewer").File} File */
```

So that this is more readable:


```suggestion
             * @param {(File) => boolean} [options.canUnlink]
             * @param {(File) => boolean} [options.onUnlink] called when the user requests to unlink the file. Returns `true` on success, allowing the file viewer to close.
```

alexkuhn:
@robodoo r+

akth-odoo:
@robodoo r+

akth-odoo:
@robodoo r+

mbaragiola (commented):
No encontré problemas materiales en el SHA actual. La acción reutiliza `showDelete` por archivo, conserva la confirmación y solo cierra el visor tras un borrado exitoso. La prueba HOOT focalizada cubre la nueva acción y todos los checks visibles, incluidos Runbot, seguridad y CLA, están en verde; no ejecuté pruebas locales. GitHub no permite a esta cuenta emitir APPROVE sin acceso explícito al repositorio, por lo que dejo este resultado como COMMENT.

Reviewed by Hermes Agent using Codex gpt-5.6-sol.

addons/mail/static/src/core/common/attachment_list.js
akth-odoo: I kept the function synchronous and simply returned `true` instead of making it `async` and awaiting the operation.

The reason is that, in the code below, `onConfirmUnlink` is called without `await`. If I also `await` `onConfirmUnlink` for consistency, the FileViewer takes longer to close. During that delay, the `get documentIcon()` getter tries to read the attachment ID, which has already been cleared, resulting in a traceback. So, I opted to simply return `true` instead of awaiting the operation.

akth-odoo:
@robodoo r+

PR #247927, opened 2026-02-10
[FIX] website_sale: fix wishlist Add to Cart button rendering

khaj-odoo:
Before:
The wishlist page `Add to Cart` and `Contact Us` buttons use bare text and me-2 margin on the icon, not matching the standard shop_product_buttons template structure. The missing <span class=`o_label`> wrapper prevents CSS display variables from controlling button label visibility across different tile layouts and sizes.

After:
Button text is wrapped in <span class=`o_label small ms-1`> and icon margin changed from me-2 to match the shop page button structure, allowing proper responsive rendering across all tile designs.

Steps to reproduce:
1.Log in and add products to the wishlist
2.Go to the wishlist page — buttons display correctly (inline)
3.Open the website editor and switch the wishlist design from
  Thumbnails to Cards, save. Switch back to Thumbnails, save.
4.The Add to Cart and Compare buttons are not rendering properly their icon with
  the correct size.

opw-5924433
image

---

### DIFF
--- a/addons/website_sale/templates/wishlist_templates.xml
+++ b/addons/website_sale/templates/wishlist_templates.xml
@@ -74,8 +74,15 @@
                                                 t-att-href="website._get_contact_us_url(combination_info['display_name'])"
                                                 class="o_wsale_product_btn_primary btn btn-primary"
                                             >
-                                                <i class="fa fa-fw fa-envelope o_not-animable me-2" role="presentation"/>
-                                                Contact Us
+                                                <span class="d-inline-flex align-items-baseline">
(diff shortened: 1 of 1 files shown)


ltinel (changes requested):
Thanks for the fix! :) It looks good, the only thing that annoys me is that the button label isn't centered anymore with the "chips" layout (and some other layouts). Do you know if that's fixable while keeping the other layouts working?

ltinel (approved):
LGTM, thanks :)
@robodoo r+

PR #247928, opened 2026-02-10
[IMP] html_editor: allow switching banner type via powerbox

dhba-odoo:
feature this PR addresses:

- Previously, it was not possible to change the banner type from inside the banner. To change the banner type, the existing banner had to be removed and a new one added.
- Now, the banner type can be switched directly from within the banner using the powerbox.
- Also, when opening the powerbox inside a banner, the command for the current banner type is not shown.

task-5895288

### DIFF
--- a/addons/html_editor/static/src/main/banner_plugin.js
+++ b/addons/html_editor/static/src/main/banner_plugin.js
@@ -28,8 +28,7 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert an info banner"),
                 icon: "fa-info-circle",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ??
-                    true,
+                    this.checkPredicates("is_banner_command_available_predicates", selection, "info") ?? true,
                 run: () => {
                     this.insertBanner(_t("Banner Info"), "💡", "info");
                 },
@@ -40,8 +39,7 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert a success banner"),
                 icon: "fa-check-circle",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ??
-                    true,
+                    this.checkPredicates("is_banner_command_available_predicates", selection, "success") ?? true,
                 run: () => {
                     this.insertBanner(_t("Banner Success"), "✅", "success");
                 },
@@ -52,8 +50,7 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert a warning banner"),
                 icon: "fa-exclamation-triangle",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ??
-                    true,
+                    this.checkPredicates("is_banner_command_available_predicates", selection, "warning") ?? true,
                 run: () => {
                     this.insertBanner(_t("Banner Warning"), "⚠️", "warning");
                 },
@@ -64,8 +61,7 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert a danger banner"),
                 icon: "fa-exclamation-circle",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ??
-                    true,
+                    this.checkPredicates("is_banner_command_available_predicates", selection, "danger") ?? true,
                 run: () => {
                     this.insertBanner(_t("Banner Danger"), "❌", "danger");
                 },
@@ -76,8 +72,7 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert a monospace banner"),
                 icon: "fa-laptop",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ??
-                    true,
+                    this.checkPredicates("is_banner_command_available_predicates", selection, "secondary") ?? true,
                 run: () => {
                     this.insertBanner(
                         _t("Monospace Banner"),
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -9,11 +9,11 @@ import { closestBlock } from "@html_editor/utils/blocks";
...
 
-function isAvailable(selection) {
-    return (
-        isHtmlContentSupported(selection) &&
-        !closestElement(selection.anchorNode, ".o_editor_banner")
-    );
+function isAvailable(alertClass) {
+    return (selection) => {
+        const banner = closestElement(selection.anchorNode, `.alert-${alertClass}`);
+        return isHtmlContentSupported(selection) && !banner;
```
deso-odoo:
```suggestion
        return isHtmlContentSupported(selection) && !closestElement(selection.anchorNode, `.alert-${alertClass}`);
```

addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -34,7 +34,7 @@ export class BannerPlugin extends Plugin {
                 title: _t("Banner Info"),
                 description: _t("Insert an info banner"),
                 icon: "fa-info-circle",
-                isAvailable,
+                isAvailable: isAvailable("info"),
```
deso-odoo: isAvailable("info") is now confusing. I would rather rename it to something like `canInsertBanner("info")`

addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -83,10 +98,10 @@ export class BannerPlugin extends Plugin {
                 },
             },
         ],
-        is_banner_command_available_predicates: (selection) => {
+        is_banner_command_available_predicates: ({ selection, alertClass }) => {
             if (
                 !isHtmlContentSupported(selection) ||
-                closestElement(selection.anchorNode, ".o_editor_banner")
+                closestElement(selection.anchorNode, `.alert-${alertClass}`)
```
deso-odoo:
```suggestion
                closestElement(selection.anchorNode, `.o_editor_banner .alert-${bannerType}`)
```

addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -28,7 +28,10 @@ export class BannerPlugin extends Plugin {
                 description: _t("Insert an info banner"),
                 icon: "fa-info-circle",
                 isAvailable: (selection) =>
-                    this.checkPredicates("is_banner_command_available_predicates", selection) ?? true,
+                    this.checkPredicates("is_banner_command_available_predicates", {
+                        selection,
+                        alertClass: "info",
+                    }) ?? true,
```
deso-odoo:
```suggestion
                    this.checkPredicates("is_banner_command_available_predicates", selection, "info") ?? true,
```

addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -83,10 +98,10 @@ export class BannerPlugin extends Plugin {
                 },
             },
         ],
-        is_banner_command_available_predicates: (selection) => {
+        is_banner_command_available_predicates: ({ selection, alertClass }) => {
```
deso-odoo:
```suggestion
        is_banner_command_available_predicates: (selection, bannerType) => {
```

addons/html_editor/static/tests/banner.test.js
```diff
@@ -331,3 +323,104 @@ test("Monospace banner should have no emoji", async () => {
...
+                    <i class="o_editor_banner_icon mb-3 fst-normal" data-oe-aria-label="Banner Info" aria-label="Banner Info">💡</i>
+                    <div class="o_editor_banner_content o-contenteditable-true w-100 px-3" contenteditable="true">
+                        <p>Test[]</p>
+                    </div>
+                </div><p data-selection-placeholder="" style="margin: -9px 0px 8px;"><br></p>`
+        )
+    );
+    await insertText(editor, "/banner");
+    await animationFrame();
+    expect(".active .o-we-command-name").not.toHaveText("Banner Info");
```
deso-odoo: Will this check for all `.o-we-command-name`?

dhba-odoo: It doesn’t check all commands because the `active` class is applied to only one command. When using `/banner`, the active class is applied to the first command by default. Since Banner Info is the first command, the test would still fail without my fix.
Still, I removed the active class from the expect so it now checks all available banner commands, not just the one that has the active class.

addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -139,7 +139,31 @@ export class BannerPlugin extends Plugin {
...
         contentClass = contentClass ? `${contentClass} ` : "";
 
+        const bannerClasses = `${containerClass}o_editor_banner user-select-none o-contenteditable-false ${
+            emoji ? "lh-1 " : ""
+        }d-flex align-items-center alert alert-${alertClass} pb-0 pt-3`;
+        const bannerContentClasses = `${contentClass}o_editor_banner_content o-contenteditable-true w-100 px-3`;
+        const emojiHtml = emoji
+            ? `<i class="o_editor_banner_icon mb-3 fst-normal" data-oe-aria-label="${htmlEscape(
+                  title
+              )}">${emoji}</i>`
```
dmo-odoo: I think the security ci is right in this case. The function could be called with an harmful value for `emoji` like `</i><script>altert('this is executed!')</script><i>`. Couldn't we call `htmlEscape` on `emoji` as well?

dhba-odoo: @dmo-odoo It was already defined the same way, i only changed its position. However, it’s better to call htmlEscape on emoji. I have made the change.

dmo-odoo:
@robodoo override=ci/security r+
Reason: html is escaped

PR #247929, opened 2026-02-10
[IMP] calendar: limit recurrent event creation to 15 years

yagp-odoo:
`forever` recurrent events could currently be created for 720 years
into the future, leading to excessive and unnecessary event generation.

This change limits the creation of forver recurrent events to a fixed
horizon of 15 years (by default) in the future, ensuring predictable
behavior and preventing unbounded recurrence.

The maximum horizon can be adjusted through the system parameter
`calendar.max_recurrence_years` for events that require a longer
recurrence window.

Task-5714633

### DIFF
--- a/addons/calendar/models/calendar_recurrence.py
+++ b/addons/calendar/models/calendar_recurrence.py
@@ -594,10 +594,17 @@ class CalendarRecurrence(models.Model):
             rrule_params['byweekday'] = weekdays
             rrule_params['wkst'] = self._get_lang_week_start()
 
+        # Limit the number of years to avoid creating recurrent events for up to hundreds of years
+        limit_years = int(self.env['ir.config_parameter'].sudo().get_param('calendar.max_recurrence_years', 15))
         if self.end_type == 'count':  # e.g. stop after X occurence
             rrule_params['count'] = min(self.count, MAX_RECURRENT_EVENT)
         elif self.end_type == 'forever':
-            rrule_params['count'] = MAX_RECURRENT_EVENT
+            if freq == 'yearly':
+                rrule_params['count'] = min(limit_years, MAX_RECURRENT_EVENT)
+            elif freq == 'monthly':
+                rrule_params['count'] = min(limit_years * 12, MAX_RECURRENT_EVENT)
+            else:
+                rrule_params['count'] = MAX_RECURRENT_EVENT
         elif self.end_type == 'end_date':  # e.g. stop after 12/10/2020
             rrule_params['until'] = datetime.combine(self.until, time.max)
         return rrule.rrule(
--- a/addons/calendar/tests/test_calendar_recurrent_event_case2.py
+++ b/addons/calendar/tests/test_calendar_recurrent_event_case2.py
@@ -52,3 +52,45 @@ class TestRecurrentEvent(common.TransactionCase):
             ('start', '>=', '2011-03-13'), ('stop', '<=', '2011-05-13')
         ])
         self.assertEqual(meetings_count, 10, 'Recurrent weekly meetings are not created!')
+
+    def test_recurrent_meeting3(self):
+        """
+        Test that 'forever' recurrences are limited to a certain number of years (default 15).
+        The default limit can be change by the system parameter `calendar.max_recurrence_years`.
+        """
+        values = {
+            'duration': 1.0,
+            'end_type': 'forever',
+            'start': '2026-04-01 05:00:00',
+            'stop': '2026-04-01 06:00:00',
+            'recurrency': True,
+        }
+        for rrule_type, name, expected_count in (
+            ('daily', 'Daily Meeting', 720),
+            ('monthly', 'Monthly Meeting', 180),
+            ('yearly', 'Yearly Meeting', 15),
+        ):
+            with self.subTest(rrule_type=rrule_type):
+                self.CalendarEvent.create(dict(values, name=name, rrule_type=rrule_type))
+                meetings_count = self.CalendarEvent.search_count([('name', '=', name)])
+                self.assertEqual(
+                    meetings_count,
+                    expected_count,
+                    f'Recurrent {rrule_type} meetings should be created and not exceed {expected_count}!',
+                )
+
+        # Edit the max recurrence years
+        self.env['ir.config_parameter'].sudo().set_param('calendar.max_recurrence_years', 5)
+        for rrule_type, name, expected_count in (
+            ('daily', 'Custom Daily Meeting', 720),
+            ('monthly', 'Custom Monthly Meeting', 60),
+            ('yearly', 'Custom Yearly Meeting', 5),
+        ):
+            with self.subTest(rrule_type=rrule_type):
+                self.CalendarEvent.create(dict(values, name=name, rrule_type=rrule_type))
+                meetings_count = self.CalendarEvent.search_count([('name', '=', name)])
+                self.assertEqual(
+                    meetings_count,
+                    expected_count,
+                    f'Recurrent {rrule_type} meetings should be created and not exceed {expected_count}!',
+                )


addons/calendar/models/calendar_recurrence.py
```diff
@@ -520,15 +520,23 @@ def _get_occurrences(self, dtstart):
         """
         self.ensure_one()
         dtstart = self._get_start_of_period(dtstart)
+
+        # Limit the number of years to avoid creating recurrent events for up to hundreds of years
+        limit_years = int(self.env['ir.config_parameter'].sudo().get_param('calendar.max_recurrence_years', 5))
```
Odoonan: Note: from 19.1 we use the new config param tools, e.g. here `get_int`() (it allows a default)

addons/calendar/models/calendar_recurrence.py
```diff
@@ -520,15 +520,23 @@ def _get_occurrences(self, dtstart):
...
+        limit_date = dtstart + relativedelta(years=limit_years)
+
         if self._is_allday():
-            return self._get_rrule(dtstart=dtstart)
+            rule = self._get_rrule(dtstart=dtstart)
+            return rule.between(dtstart, limit_date, inc=True)
```
Odoonan: Can you explain why we have `inc=True` in both places? thanks!

addons/calendar/tests/test_calendar_recurrent_event_case2.py
```diff
@@ -52,3 +52,17 @@ def test_recurrent_meeting2(self):
             ('start', '>=', '2011-03-13'), ('stop', '<=', '2011-05-13')
         ])
         self.assertEqual(meetings_count, 10, 'Recurrent weekly meetings are not created!')
+
+    def test_recurrent_meeting3(self):
+        # Test that the recurrence is limited to a certain number of years (default 5).
```
Odoonan: Could we put the creation values in a dict, and use it twice, once with default value and one with a custom value of the new parameter you added? to cover more ground? Also, please update the comment accordingly and include the parameter name `calendar.max_recurrence_years`, it will be easier to grep

addons/calendar/tests/test_calendar_recurrent_event_case2.py
```diff
@@ -52,3 +52,17 @@ def test_recurrent_meeting2(self):
...
+        self.CalendarEvent.create({
+            'count': 50,
```
Odoonan: Seems a bit strange not to let the count operate. I've asked BEJO, we'll see.

addons/calendar/models/calendar_recurrence.py
Odoonan, line 515: Is this exclusively used at creation ? Nothing will happen to existing events with longer recurrences? (thinking of the calendar sync here)

Odoonan (commented):
Hi @yagp-odoo ,

Thanks for handling this one. Here is a first pass.

Cheers!

addons/calendar/models/calendar_recurrence.py
yagp-odoo: So that we can include the start and end dates in the occurrences. else it will exclude start and end boundaries.

addons/calendar/models/calendar_recurrence.py
yagp-odoo: It will affect calendar sync, too. I mean, we set a max limit of 720 events recurrence, so it will apply to calendar sync as well.

Odoonan, line 515: Hi @yagp-odoo 

Sorry for the delay. Thanks for the updates, looking better. Just a question, as we need to be extra careful with recurrence (and synchronizations) in stable.

What would happen to _existing_ recurrences extending to more than 15 years (so, month and year forever recurrences if I count correctly) ? On edition notably ( in odoo / in external calendar, google or microsoft ) of the meeting for instance? (of all / future events) both when synced and not synced

We want to avoid side effects as much as possible.

Thanks & cheers!

yagp-odoo: Hello, @Odoonan , after further testing, I came to know it will not work with sync (just tried with google_calendar).

Here is the reason:
We create reccurent event, suppose `rrule_type` is yearly, and `end_type` is forever.
Now in `_get_rrule` if `end_type` if forever we set count = 720 (MAX)
And during `_rrule_parse` we set `end_type` to `count` if we get it :arrow_up: 
So if only the calendar is installed, it will work because the values will change after, but once google_calendar is installed, it will get `end_type` = `count` and `count` = 720.

So in short, if you create a recurrent event, you will see that its `end_type` is changed to count, and `count` is 720. Hence, on edition of the meeting, nothing happens as our condition is not met

Log when google_calendar is installed:
```cmd
2026-03-11 06:20:23,069 96617 INFO base19 werkzeug: 127.0.0.1 - - [11/Mar/2026 06:20:23] "POST /web/dataset/call_kw/calendar.event/onchange#calendar.event.onchange HTTP/1.1" 200 - 9 0.003 0.012
count changed here. before:  1  after:  720
CALENDAR RECURRENCE GET RRULE Called. Return params:  {'dtstart': None, 'interval': 1, 'count': 720}
end_type changed here. before:  forever  after:  count
CALENDAR RECURRENCE PARSE Called. Return data:  {'rrule_type': 'yearly', 'count': 720, 'interval': 1, 'until': None, 'end_type': 'count'}
count changed here. before:  720  after:  720
CALENDAR RECURRENCE GET RRULE Called. Return params:  {'dtstart': datetime.datetime(2026, 3, 11, 12, 0), 'interval': 1, 'count': 720}

```

The possible small fix for this I can think of is to modify the condition in `_rrule_parse`:
```python
elif data.get('count') and data['count'] != MAX_RECURRENT_EVENT:
    data['end_type'] = 'count'
```

Or we can skip the condition of `end_type == 'forever'` and just limit the years regardless of count.

What do you think?

Odoonan: Hello @yagp-odoo 

Thanks for having a look! I'll discuss it with BEJO and see what we could do. I'll come back to you. 

Cheers!

addons/calendar/tests/test_calendar_recurrent_event_case2.py
```diff
@@ -52,3 +52,27 @@ def test_recurrent_meeting2(self):
...
+            'end_type': 'forever',
+            'name': 'Yearly Meeting',
+            'start': '2026-04-20 05:00:00',
+            'stop': '2026-04-20 06:00:00',
+            'recurrency': True,
+            'rrule_type': 'yearly',
+        }
+        self.CalendarEvent.create(values)
+        meetings_count = self.CalendarEvent.search_count([('name', '=', 'Yearly Meeting')])
+        self.assertEqual(meetings_count, 15, 'Recurrent yearly meetings should be created and not exceed 15!')
```
Odoonan: We could probably add a case for monthly in both parts of the test then since your latest changes :)

addons/calendar/models/calendar_recurrence.py
Odoonan: I've played a bit with recurrences when google sync is on and here are a few things I noted, using daily recurrence, and might be useful: (already before your changes)

1) When creating a forever recurrence from Odoo, it will be changed to COUNT + 720 on save. On google, it will show indeed "Daily, 720 times"

2) When creating a forever recurrence from Google, it will create 730 events (2*365 = 2 years), not 720.
-> In Odoo, the meeting will be created as forever, but only have 720 events
-> When editing a synchronized field (e.g. calendar description), it will synchronize to Google and change to a COUNT + 720 as well. The extra 10 meetings will be removed from Google.

So basically it seems sync from Odoo will always set the recurrence to COUNT with MAX.

As it was like that before, I would not change it, but indeed maybe just update the max amounts as you did. Note that tailing events will be lost next time we sync I believe, so I would probably say an IMP is fine instead of a FIX, even if it is unlikely to have written on month / year events after 15 years.

Does your change do anything to the outlook sync?

Odoonan (commented):
Hello @yagp-odoo 

Thanks for the update. I've added a few details on google sync and recurrences, a question and a suggestion for the test.

I will probably green the validation stage then, as it has been there a while now, I have been quite busy 🥸 

Cheers!

addons/calendar/models/calendar_recurrence.py
yagp-odoo: > Does your change do anything to the outlook sync?

I don't think the current code will affect Outlook sync, as that module has its own way to handle recurrent events.

`addons/microsoft_calendar/models/calendar.py` -> `def _microsoft_values`

Limiting years there should done the job, tho mostly have to test, probably by bejo (i dont have creds :smiling_face_with_tear: )

I'll update the code :+1: 

PS: `_microsoft_values` is used for Odoo <> Microsoft sync, but we can only create a recurrent event from Microsoft Calendar.

yagp-odoo: Update!

There is already a system parameter for Microsoft Sync: `microsoft_calendar.sync.range_days`

By default, it's set to 2 years, not sure why, but probably having sync [timeout error](https://github.com/odoo/odoo/commit/dac590d64a02c9b6ea3f6a75a0bdce6c51312376) with a large amount of events

Also, tested Odoo & Google Sync, works perfectly fine :v:

addons/calendar/tests/test_calendar_recurrent_event_case2.py
```diff
@@ -52,3 +52,41 @@ def test_recurrent_meeting2(self):
...
+        values = {
+            'duration': 1.0,
+            'end_type': 'forever',
+            'start': '2026-04-01 05:00:00',
+            'stop': '2026-04-01 06:00:00',
+            'recurrency': True,
+        }
+        for rrule_type, name, expected_count in (
+            ('yearly', 'Yearly Meeting', 15),
+            ('monthly', 'Monthly Meeting', 180),
```
awa-odoo, line 70: Could also test daily while we're at it.

addons/calendar/tests/test_calendar_recurrent_event_case2.py
```diff
@@ -52,3 +52,41 @@ def test_recurrent_meeting2(self):
...
+        ):
+            self.CalendarEvent.create(dict(values, name=name, rrule_type=rrule_type))
```
awa-odoo: Seems like a perfect candidate for a subTest usage with "rrule_type" as input.

awa-odoo:
@robodoo r+

@Odoonan I see you had concerns about merging in 19.0. I think it's a pretty safe change and behavior-wise I can hardly see people complaining about us "only creating 15 years of events". Worst case it's a pretty easy revert. Let me know if you think otherwise 🙂

PR #247930, opened 2026-02-10
[FIX] html_editor: copy full table selection in clipboard block

walidsahli:
Problem:
In Knowledge, when adding a clipboard block and inserting a table inside it, clicking on the copy button only copies the text inside the table instead of the full table structure.

Cause:
In `html_viewer`, the copy logic clones only the deepest selected node. In contrast, `html_editor` (via `clipboard_plugin`) copies the entire selection range. This difference causes inconsistent behavior between editable and locked content.

Solution:
Align the behavior by copying the full selection in `html_viewer`, ensuring tables and other complex structures are copied entirely and consistently.

Steps to reproduce:
- Open Knowledge.
- Add a clipboard block.
- Insert a table inside the block.
- Lock the content.
- Click the copy button.
- Paste into any editable field.
- Observe that only the text (not the table) is pasted.

opw-5476320

---

### DIFF
--- a/addons/html_editor/static/src/components/html_viewer/html_viewer.js
+++ b/addons/html_editor/static/src/components/html_viewer/html_viewer.js
@@ -15,7 +15,6 @@ import { fillHtmlTransferData } from "@html_editor/utils/clipboard";
 import { fixInvalidHTML, instanceofMarkup } from "@html_editor/utils/sanitize";
 import { HtmlUpgradeManager } from "@html_editor/html_migrations/html_upgrade_manager";
 import { TableOfContentManager } from "@html_editor/others/embedded_components/core/table_of_content/table_of_content_manager";
-import { getDeepestPosition } from "@html_editor/utils/dom_info";
 import { scrollAndHighlightHeading } from "@html_editor/utils/url";
 
 export class HtmlViewer extends Component {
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/components/html_viewer/html_viewer.js
```diff
@@ -152,19 +151,7 @@ export class HtmlViewer extends Component {
...
         ev.preventDefault();
         const selection = ev.target.ownerDocument.defaultView.getSelection();
-        const [deepAnchorNode, deepAnchorOffset] = getDeepestPosition(
-            selection.anchorNode,
-            selection.anchorOffset
-        );
-        const [deepFocusNode, deepFocusOffset] = getDeepestPosition(
-            selection.focusNode,
-            selection.focusOffset
-        );
```
dmo-odoo, line 162: @bso-odoo Those lines seem to come from you so I fear we might introduce an issue by removing it. Do you remember why they were needed, by any chance?

bso-odoo: @dmo-odoo I believe this might have been a mistake: when trying to replicate `getEditableSelection` which was called in the existing `onCopy` inside the editor, I also replicated the code of `handleEmptySelection`. :grimacing:

PR #247932, opened 2026-02-10
[FIX] website, html_builder: show invisible snippets in preview dialog

blse-odoo:
The (saved custom) snippets which were invisible on desktop or conditionally invisible, appeared as empty in the dialog to preview snippet.

This commit changes the css rules so that
- conditionally visible elements are always visible in preview
- snippet that are device invisible are show with the overlay (and inner elements that are device invisible are not shown)

Steps to reproduce:
- Open website builder
- Drop a section
- Change the "Visibility" to hide on desktop
- Click on the eye to show it
- Save the snippet as a custom snippet
- Open the dialog to add a snippet
- Bug: the newly saved snippet is shown as empty

Same for "Visibility" to "Conditionally". And Same for "Visibility" to hide on mobile, if the window size is reduced when showing the dialog.

task-5149984

### DIFF
--- a/addons/html_builder/static/src/snippets/snippet_viewer.scss
+++ b/addons/html_builder/static/src/snippets/snippet_viewer.scss
@@ -75,6 +75,65 @@
                         }
                     }
                 }
+
+                // Force visibility on root snippets if desktop/mobile invisible
+                &.o_snippet_desktop_invisible {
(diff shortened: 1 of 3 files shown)


blse-odoo:
in stable: https://github.com/odoo/odoo/pull/250395

PR #247934, opened 2026-02-10
[FIX] html_editor: skip syntax highlighting for very large pre blocks

adch-odoo:
### Purpose of this PR:

- Large `<pre>` blocks that use syntax highlighting can severely impact performance when opening a task or when applying transformations (e.g. converting to paragraph). In extreme cases, this can even crash the browser.
- To avoid this, `<pre>` elements whose text content exceeds the syntax highlighting size limit are no longer converted into syntax-highlighted embedded code components (with textarea and Prism). Instead, they are kept as plain `<pre>` elements.

task-5907330

---

### DIFF
--- a/addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
+++ b/addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
@@ -23,6 +23,9 @@ export class SyntaxHighlightingPlugin extends Plugin {
         "protectedNode",
         "embeddedComponents",
     ];
+    static defaultConfig = {
+        syntaxHighlightingTextLimit: 1_000_000,
+    };
     /** @type {import("plugins").EditorResources} */
     resources = {
         // Ensure focus can be preserved within the textarea:
@@ -127,6 +130,12 @@ export class SyntaxHighlightingPlugin extends Plugin {
             (pre) => !pre.closest(CODE_BLOCK_SELECTOR)
         );
         for (const pre of nonEmbeddedPres) {
+            // Do not convert large <pre> elements into syntax-highlighted
+            // components. If the content exceeds the defined limit, original
+            // <pre> is preserved to avoid performance issues.
+            if (pre.textContent.length > this.config.syntaxHighlightingTextLimit) {
+                continue;
+            }
             const isPreInSelection = !targetedNodes.some((node) => !pre.contains(node));
             const embeddedProps = JSON.stringify(
                 Object.assign(
--- a/addons/html_editor/static/tests/syntax_highlighting.test.js
+++ b/addons/html_editor/static/tests/syntax_highlighting.test.js
@@ -52,10 +52,26 @@ const testEditorWithHighlightedContent = async (config) =>
     await testEditor({
         ...config,
         compareFunction: compareHighlightedContent,
-        config: configWithEmbeddings,
+        config: {
+            ...configWithEmbeddings,
+            // Reduce the syntax highlighting limit for tests.
+            // Real limit (e.g. 1 million characters) is too large to use
+            // in unit test, so use smaller value to reliably test behavior.
+            syntaxHighlightingTextLimit: 50,
+        },
     });
 
 beforeEach(patchPrism);
+
+test("should not syntax-highlight when content exceeds the highlighting limit", async () => {
+    await testEditorWithHighlightedContent({
+        contentBefore: "<p>[]Lorem ipsum dolor sit amet consectetur adipiscing elit</p>", // exceeds limit
+        stepFunction: async (editor) => {
+            await insertPre(editor);
+        },
+        contentAfter: "<pre>[]Lorem ipsum dolor sit amet consectetur adipiscing elit</pre>",
+    });
+});
 
 test("starting edition with a pre activates syntax highlighting", async () => {
     await testEditorWithHighlightedContent({


addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
```diff
@@ -128,6 +142,15 @@ export class SyntaxHighlightingPlugin extends Plugin {
...
                 },
+                convertSyntaxHighlightingToPre: ({ target }) => {
+                    const component = target.closest(`[data-embedded='${name}']`);
+                    const embeddedProps = getEmbeddedProps(component);
+                    const pre = this.document.createElement("pre");
+                    pre.textContent = embeddedProps.value;
+                    component.replaceWith(pre);
+                    newlinesToLineBreaks(pre);
+                    this.dependencies.selection.setCursorEnd(pre);
+                },
```
deso-odoo: `convertSyntaxHighlightingToPre` is almost doing the same thing as `convertToParagraph` can't you create a new function like `convertToElement` and pass type?

addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
```diff
@@ -7,9 +7,13 @@ import {
     newlinesToLineBreaks,
 } from "../../core/syntax_highlighting/syntax_highlighting_utils";
 import { removeInvisibleWhitespace } from "@html_editor/utils/dom";
+import { closestElement } from "@html_editor/utils/dom_traversal";
 
 const CODE_BLOCK_CLASS = "o_syntax_highlighting";
 const CODE_BLOCK_SELECTOR = `div.${CODE_BLOCK_CLASS}`;
+export const SYNTAX_HIGHLIGHTING_LIMITS = {
+    maxTextLength: 1_000_000,
+};
```
deso-odoo: `SYNTAX_HIGHLIGHTING_LIMITS` is object because you patch it for test cases. I was thinking about keeping it as a static property of `SyntaxHighlightingPlugin`, but better if final reviewer decides on this.

addons/html_editor/static/src/others/embedded_component_plugin.js
```diff
@@ -31,7 +31,7 @@ export class EmbeddedComponentPlugin extends Plugin {
         restore_savepoint_handlers: () => this.handleComponents(this.editable),
         history_reset_handlers: () => this.handleComponents(this.editable),
         history_reset_from_steps_handlers: () => this.handleComponents(this.editable),
-        step_added_handlers: ({ stepCommonAncestor }) => this.handleComponents(stepCommonAncestor),
+        step_added_handlers: () => this.handleComponents(this.editable),
```
deso-odoo: why?

addons/html_editor/static/tests/syntax_highlighting.test.js
```diff
@@ -55,7 +56,85 @@ const testEditorWithHighlightedContent = async (config) =>
...
+                            value: "Lorem ipsum dolor sit amet consectetur",
+                            textareaRange: 38,
+                        }) +
+                        '<p data-selection-placeholder="" style="margin: -9px 0px 8px;"><br></p>',
+                    "syntax highlighting applied within limit",
+                    editor
+                );
+            },
+            contentAfter:
+                '<pre data-embedded="readonlySyntaxHighlighting" data-language-id="plaintext">Lorem ipsum dolor sit amet consectetur</pre>[]',
```
deso-odoo: selection `</pre>[]`?

addons/html_editor/static/src/others/embedded_component_plugin.js
adch-odoo: I used this.editable here because if we have a normal `<pre>` like
`<pre>abcd<u>ef[]gh</u>ijkl</pre>` and we press backspace so that the text length drops below the syntax-highlighting limit, the `<pre>` needs to be converted into an embedded syntax-highlighted block.

When addStep is called, the stepCommonAncestor becomes `<u>` since the edit happens there. But syntax highlighting is scoped to `<pre>`, so if we rely only on the common ancestor, the `<pre>` would not be converted.

addons/html_editor/static/tests/syntax_highlighting.test.js
adch-odoo: readonlySyntaxHighlighting is meant to save only the `<pre>` element along with the information required to rebuild the embedded component later, so the persisted DOM remains independent of this plugin. Because of that, selection handling does not matter here. We already follow the same behavior in other tests as well.

addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
adch-odoo: I don’t think these two should be merged. convertSyntaxHighlightingToPre is a live conversion, so it doesn’t require a history step and simply sets the cursor at the end.

addons/html_editor/static/src/others/embedded_component_plugin.js
deso-odoo: `closestElement(stepcommonAncestor, "pre, .odoo-editor-editable")` or `closestElement(stepcommonAncestor, "[data-embedded], .odoo-editor-editable")`?

adch-odoo: `handleComponents(elem)` only works if `elem` is `[data-embedded]` or contains `[data-embedded]`, since it only does `elem.querySelectorAll("[data-embedded]")` and (`elem.matches`).

In our case:

```html
<pre>abcd<u>ef[]gh</u>ijkl</pre>
```

The edit happens inside `<u>`, so `stepCommonAncestor` is `<u>`.

* `closestElement(stepCommonAncestor, "pre, .odoo-editor-editable")` does not help because `handleComponents` does not look for `<pre>` at all, only `[data-embedded]`. Passing a normal `<pre>` is a no-op.
* `closestElement(stepCommonAncestor, "[data-embedded], .odoo-editor-editable")` also fails because `<u>` is inside a regular `<pre>` and there is no `[data-embedded]` in its ancestor chain.

Therefore, we must pass **`this.editable`**, which is a stable root guaranteed to contain embedded hosts, allowing `handleComponents` to find and mount them correctly.

addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
dmo-odoo: Why not a config of the plugin instead of an hardcoded limit?

adch-odoo: Sorry, I didn’t understand what you meant by “plugin config.” Are you referring to the editor config?

dmo-odoo: Yes, editor config but used by the plugin.

adch-odoo: @dmo-odoo I have made the changes. Please have a look. Thanks!!!

addons/html_editor/static/src/fields/html_field.js
```diff
@@ -358,6 +358,7 @@ export const htmlField = {
             mediaModalParams: {
                 useMediaLibrary: true,
             },
+            syntaxHighlightingTextLimit: 1_000_000,
```
dmo-odoo: I would have expected the default value to live in the plugin itself, because as of now there is no default value anymore if you use the editor by another mean than the html field. 

The only place where we override it in practice is the tests, but someone might override it to have a bigger (or smaller) value as needed depending on the expected performances.

adch-odoo: You're right, the default should be defined in the plugin. I've moved it to defaultConfig and removed it from HtmlField, so the plugin works independently while still allowing overrides via editor config.

addons/html_editor/static/src/others/embedded_component_plugin.js
dmo-odoo: Remind me again why we want to do it live instead of when opening the record? The way I see it, it's not a big deal if a pre does not automatically convert into a syntax highlighted pre, but we would need a way to convert it manually.

adch-odoo: Since the conversion happens during normalize (addCodeBlocks), we cannot rely only on a check like pre.textContent > limit.
So a `<pre>` that exceeded the limit (and was intentionally kept as a normal `<pre>`) can later drop below the limit (backspace), and normalize would incorrectly convert it into an embedded component.
Because of this, a simple pre.textContent > limit condition is not sufficient in normalize. To make this reliable, we need to introduce an explicit attribute/flag to tell normalize that this `<pre>` should not be converted, even if its content length goes below the limit.
This is why I think live conversion is the simpler approach.  Otherwise, normalize needs extra state/attribute to avoid unintended conversions.

dmo-odoo: I see, thanks for the reminder.

I'm a bit confused by your example with:
```html
<pre>abcd<u>ef[]gh</u>ijkl</pre>
```
You say that  `closestElement` with `"pre"` would not work because the PRE element does not have `[data-embedded]` but, in that case, where is `[data-embedded]` then? On the editable itself I suppose since it works if you pass `this.editable`? How comes it works right now but not with the other changes?

adch-odoo: If we have content:
```html
<pre>abcd<u>ef[]gh</u>ijkl</pre>
```
and suppose on backspace it gets converted into a syntax highlighting `<pre>`, check the DOM inside handleComponents method of embedded_component_plugin (see screenshot). You’ll notice the normal `<pre>` is no longer in the DOM because syntax highlighting is being set up, so using closestElement('pre') doesn’t make sense.

data-embedded is applied on the highlighting `<div>`, not on the `<pre>`. In forEachEmbeddedComponentHost, it looks for [data-embedded] inside the given elem using querySelectorAll, and also checks if elem itself matches.

Since we now pass the editable as elem, the selector runs on the whole editor. The syntax highlight blueprint `<div>` is already in the DOM and has data-embedded, so it gets picked as a target. Then forEachEmbeddedComponentHost calls the callback, and handleComponents mounts the component for that host.
Please let me know if anything is unclear.
image

dmo-odoo: > If we have content:
> 
> ```
> <pre>abcd<u>ef[]gh</u>ijkl</pre>
> ```
> 
> and suppose on backspace it gets converted into a syntax highlighting `\<pre\>`, check the DOM inside handleComponents method of embedded_component_plugin (see screenshot). You’ll notice the normal `\<pre\>` is no longer in the DOM because syntax highlighting is being set up, so using closestElement('pre') doesn’t make sense.
> 
> data-embedded is applied on the highlighting `<div>`, not on the `\<pre\>`. In forEachEmbeddedComponentHost, it looks for [data-embedded] inside the given elem using querySelectorAll, and also checks if elem itself matches.

Ok, but then why `closestElement(stepCommonAncestor, "[data-embedded], .odoo-editor-editable")` wouldn't work?

adch-odoo: stepCommonAncestor (`<u>ef[]gh</u>`) is not connected to the DOM, then closestElement will return null. You can refer to the comment above screenshot, there’s only a syntaxhighlighting blueprint present there.

dmo-odoo: @adch-odoo I see what you mean now, thanks.

@abd-msyukyu-odoo Would there be any issue with running `handleComponents` on `this.editable` at every step added? On one hand, I feel like stopping at `stepCommonAncestor` was more of an early optimization than a real functional requirement. On the other hand though, it has been sufficient until this PR, so I feel like the situation here might be a bit fishy and there might be something wrong with how we do things here. What do you think?

abd-msyukyu-odoo: I'm off today so I'll answer tomorrow, but this whole thing seems to be a lot of chipotage 😅

abd-msyukyu-odoo: Should be fine to do it on the editable, but it feels weird to opt for an optimization regression (even if negligible) to solve an extremely rare use case.

But my question is: do you really want to dynamically switch between states while the user is writing? I feel like we could just make that decision to use prism or not when loading the article, and stick to that state until reload. Else it will interrupt the flow of writing, no? If I understood properly, we are talking about a 1000000 character code block... A Odoo HTML field does not seem to be the place for such a volume imo
(we are talking about a book of ~400 pages here, if I'm not wrong in my estimations)

dmo-odoo: I asked the same question and the answer was:

> Since the conversion happens during normalize (addCodeBlocks), we cannot rely only on a check like pre.textContent > limit. So a `\<pre\>` that exceeded the limit (and was intentionally kept as a normal `\<pre\>`) can later drop below the limit (backspace), and normalize would incorrectly convert it into an embedded component. Because of this, a simple pre.textContent > limit condition is not sufficient in normalize. To make this reliable, we need to introduce an explicit attribute/flag to tell normalize that this `\<pre\>` should not be converted, even if its content length goes below the limit. This is why I think live conversion is the simpler approach. Otherwise, normalize needs extra state/attribute to avoid unintended conversions.

But actually I think you're right. The case is niche enough that we can just do the simplest thing, that is do the check on load and that's it. It's not a big deal if `addCodeBlocks` transforms it later on when it becomes short enough. And even then, can't we just do the check in `addCodeBlocks`? That way we will never create a code block for a pre that is too long, and we're done. In pratice, the issue happens because the record is created automatically from the content of an email and it is not possible to open the form view because the syntax highlighting freezes the browser. It's the only real case that matters.

adch-odoo: @dmo-odoo Done with the suggested changes. Please have a look. Thanks!!!

addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
```diff
@@ -87,6 +90,19 @@ export class SyntaxHighlightingPlugin extends Plugin {
...
+            // Do not convert large <pre> elements into syntax-highlighted
+            // components. If the content exceeds the defined limit, original
+            // <pre> is preserved to avoid performance issues. Any leftover
+            // embedded attributes are removed.
+            if (pre.textContent.length > this.config.syntaxHighlightingTextLimit) {
+                if (pre.dataset.embedded === "readonlySyntaxHighlighting") {
+                    delete pre.dataset.embedded;
+                    delete pre.dataset.languageId;
+                    pre.removeAttribute("data-code-wrap");
+                }
```
dmo-odoo: Why bother removing those attributes though? It takes more lines of code for no real benefit, and we would lose useful data if we ever want to dynamically re-enable the syntax highlighting in the future. Also, even if we wanted to remove those attributes, why only do it for the readonly version and not the other? Anyway, I suggest you remove those lines and adapt the comment accordingly, thanks!

addons/html_editor/static/src/others/embedded_component_plugin.js
dmo-odoo: @adch-odoo I think it's better this way, thanks!

addons/html_editor/static/src/others/embedded_components/plugins/syntax_highlighting_plugin/syntax_highlighting_plugin.js
adch-odoo: When we save, the syntax highlighting block gets converted into something like:
```html
<pre data-embedded="readonlySyntaxHighlighting" data-language-id="plaintext" data-oe-version="2.0">aaaaaa</pre>
```
These attributes are added by the cleanForSave method in the syntax highlighting plugin.
Later, when we open the page again, we read this data and convert it back into a syntax-highlighted block.
Now, if the content is larger than the allowed limit, we skip converting it into syntax highlighting. So the `<pre>` stays as it is in the DOM with those attributes.
But these attributes are not meant for a normal `<pre>` element. That’s why I removed them.
Btw keeping them doesn’t cause any issue. I’ve removed the code that was deleting those attributes.

dmo-odoo: I understand why you removed them, indeed they don't make sense on a PRE. But my point was that if, later on, the PRE becomes a syntax highlighted code block again, it will have been good that we keep those attributes, most notably the language metadata. For example if the performance of the library is better in the future and we can lower the limit so that some long PRE would still be highlighted again. Thanks for the change.

PR #247935, opened 2026-02-10
[IMP] html_builder, website: replace 50% height option with 75%

Syozik:
Users rarely want to define a 50% height, either they want Auto, or
100% or something that's close enough to show the visitor there's still
content below the snippet. We replace all occurrences of 50% height
snippets with 3/4 height, except for the cover (e.g. Blog Cover) as I
don't find replacing Half Screen Size with 75% Size useful there.

task-5921163

### DIFF
--- a/addons/website/static/src/builder/plugins/options/scroll_button_option.xml
+++ b/addons/website/static/src/builder/plugins/options/scroll_button_option.xml
@@ -5,8 +5,10 @@
     <BuilderRow label="this.state.heightLabel">
         <BuilderButtonGroup action="'scrollButtonSectionHeightClass'">
             <BuilderButton id="'minheight_auto_opt'" actionParam="''" title.translate="Fit content">Auto</BuilderButton>
-            <BuilderButton actionParam="'o_half_screen_height'" title.translate="Half screen">50%</BuilderButton>
+            <BuilderButton actionParam="'o_three_quarter_height'" title.translate="75% of the screen">75%</BuilderButton>
             <BuilderButton id="'full_height_opt'" actionParam="'o_full_screen_height'" title.translate="Full screen">100%</BuilderButton>
+            <!-- A snippet can have .o_half_screen_height, but we don't want to offer it as an option, this is kept for compatibility -->
+            <BuilderButton actionParam="'o_half_screen_height'" className="'d-none'">50%</BuilderButton>
         </BuilderButtonGroup>
     </BuilderRow>
 
--- a/addons/website/static/src/scss/website.scss
+++ b/addons/website/static/src/scss/website.scss
@@ -941,7 +941,7 @@ table.table_desc tr td {
     border-radius: 0;
 }
 
-.o_full_screen_height {
+@mixin o_full_screen_height {
     display: flex;
     flex-direction: column;
     justify-content: space-around;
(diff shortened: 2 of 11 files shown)


addons/website/static/src/scss/website.scss
```diff
@@ -933,14 +933,24 @@ table.table_desc tr td {
     // Arc browser does not support this correctly at the moment.
     min-height: 100svh !important;
 }
+
+.o_full_screen_height {
+    @include o_full_screen_height;
+}
+
+.o_three_quarter_height {
+    @include o_full_screen_height;
```
Syozik, line 959: First, I simply extended .o_full_screen_height, but the CI style check didn't pass:
> @extends should be avoided as much as possible. Check with js team for more info

So I had to make it a mixin

AlessandroLupo (commented):
Nice that you added the test! 👍 The code looks good to me.
I believe you should check with a PO what to do with the snippets that currently have `o_half_screen_height`. Leave them like that or replace with `o_three_quarter_height`? Apart from that, it looks fine 👌

addons/website/static/src/builder/plugins/options/scroll_button_option.xml
```diff
@@ -5,8 +5,10 @@
     <BuilderRow label="this.state.heightLabel">
         <BuilderButtonGroup action="'scrollButtonSectionHeightClass'">
             <BuilderButton id="'minheight_auto_opt'" actionParam="''" title.translate="Fit content">Auto</BuilderButton>
-            <BuilderButton actionParam="'o_half_screen_height'" title.translate="Half screen">50%</BuilderButton>
+            <BuilderButton actionParam="'o_three_quarter_height'" title.translate="75% of the screen">75%</BuilderButton>
             <BuilderButton id="'full_height_opt'" actionParam="'o_full_screen_height'" title.translate="Full screen">100%</BuilderButton>
+            <!-- A snippet can have .o_half_screen_height, but we don't want to offer it as an option -->
```
bvr-odoo: Explain that it's kept for compatiblity

bvr-odoo (changes requested):
Hello @Syozik ,

- When modifying the CSS of a snippet (e.g. `s_color_blocks_2`), do not forget that older versions of the snippet still exist on users’ websites. When they update their site, they would lose all the rules applied to `o_half_screen_height`, even though this class is still present in their pages. Therefore, the class must be kept and a comment should be added explaining that it remains for backward compatibility with older versions. The same issue also applies to the snippet viewer, which may still contain this class in custom snippets, and to the “Add New Page” dialog where saved pages may also include this class.

- Please also add the Enterprise and Design-theme PRs links to the task.

addons/website/static/src/builder/plugins/options/scroll_button_option.xml
```diff
@@ -5,8 +5,10 @@
     <BuilderRow label="this.state.heightLabel">
         <BuilderButtonGroup action="'scrollButtonSectionHeightClass'">
             <BuilderButton id="'minheight_auto_opt'" actionParam="''" title.translate="Fit content">Auto</BuilderButton>
-            <BuilderButton actionParam="'o_half_screen_height'" title.translate="Half screen">50%</BuilderButton>
+            <BuilderButton actionParam="'o_three_quarter_height'" title.translate="75% of the screen">75%</BuilderButton>
             <BuilderButton id="'full_height_opt'" actionParam="'o_full_screen_height'" title.translate="Full screen">100%</BuilderButton>
+            <!-- A snippet can have .o_half_screen_height, but we don't want to offer it as an option, this is kept for compatibility -->
+            <BuilderButton actionParam="'o_half_screen_height'" className="'d-none'">50%</BuilderButton>
```
bvr-odoo, line 11: It seems cleaner to remove the class in `ScrollButtonSectionHeightClassAction` (if changed) and add a comment there. No need to pollute the option XML with this 😉

bvr-odoo (changes requested):
last detail 😉

addons/website/static/src/builder/plugins/options/scroll_button_option.xml
Syozik: That's what I did originally, but then actually it was a bit cleaner to do it in the option XML.
To move it back to the action, I'd need to add a check in `isApplied` whether the editingElement has the half screen class, in order not to mark "Auto" as selected, and then a line in `apply` to remove the class. The hack with `d-none` is used in the codebase for similar cases already. 

But if you still think I should move it to the action, I'll do it.

bvr-odoo: > Member

Ok, let's keep it like that 👍🏼 
Do you think it's necessary to rebase ?

bvr-odoo (approved):

PR #247937, opened 2026-02-10
[IMP] l10n_my: is_company as per vat

soza-odoo:
In this commit:
- For Malaysia, TIN is distinguishable between an individual and a company by prefix
- if vat starts with 'IG', 'is_company' should remain False

Task [link](https://www.odoo.com/odoo/project.task/5912436)
task-5912436

### DIFF
--- a/addons/l10n_my/models/res_partner.py
+++ b/addons/l10n_my/models/res_partner.py
@@ -1,0 +1,11 @@
+from odoo import models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def _compute_is_company(self):
+        super()._compute_is_company()
+        for partner in self:
+            if partner.country_code == 'MY' and partner.is_company and partner.vat.upper().startswith("IG"):
+                partner.is_company = False
--- a/addons/l10n_my/models/__init__.py
+++ b/addons/l10n_my/models/__init__.py
@@ -1,2 +1,4 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
-from . import product_template, template_my
+from . import product_template
+from . import res_partner
+from . import template_my
--- a/addons/l10n_my/tests/__init__.py
+++ b/addons/l10n_my/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_res_partner
--- a/addons/l10n_my/tests/test_res_partner.py
+++ b/addons/l10n_my/tests/test_res_partner.py
@@ -1,0 +1,43 @@
+from odoo.tests import tagged
+from odoo.addons.account.tests.common import AccountTestInvoicingCommon
+
+
+@tagged('post_install_l10n', 'post_install', '-at_install')
+class L10nMYTest(AccountTestInvoicingCommon):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.partner_a.write({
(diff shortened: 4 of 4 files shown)


addons/l10n_my/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def _compute_is_company(self):
+        super()._compute_is_company()
+        for partner in self:
+            if partner.country_code == 'MY' and not partner._is_vat_void(partner.vat) and partner.vat.startswith("IG"):
```
zepa-odoo:
```suggestion
            if partner.country_code == 'MY' and not partner._is_vat_void(partner.vat) and partner.vat.upper().startswith("IG"):
```
makes the check case-insensitive and prevents issues with lowercase or mixed-case VAT values.

zepa-odoo (changes requested):
Hello @soza-odoo ,
Thanks for this PR!

zepa-odoo (approved):

addons/l10n_my/models/res_partner.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def _compute_is_company(self):
+        super()._compute_is_company()
+        for partner in self:
+            if partner.country_code == 'MY' and not partner._is_vat_void(partner.vat) and partner.vat.upper().startswith("IG"):
+                partner.is_company = False
```
vin-odoo, line 11: I think we can make the if shorter in two ways:
- Filter self at the loop level (I use a generator in the example, but filtered works too)
- You can in theory skip the is_vat_void check by checking is_company instead (would have been set in the super)

```suggestion
        for partner in (p for p in self if p.country_code == 'MY'):
            if partner.is_company and partner.vat.upper().startswith("IG"):
                partner.is_company = False
```
I believe this will give you the correct result in a clearer (IMO of course) way:
'If the super set this MY partner as company, but the VAT starts with IG, we revert that'

addons/l10n_my/models/res_partner.py
vin-odoo, line 1: It would never hurt to add a simple test for this 😄

addons/l10n_my/models/res_partner.py
soza-odoo: IMO to make the if shorter, we should not use two loops here (one for filtering the MY partner and another for travers throught that filtered partner), by this we are increasing time complexity.
So I keep the previous code with only one loop and remove that '_is_vat_void' check from the if.
Let me know what you think about it.

soza-odoo:
Hello, @vin-odoo 
I have updated the code and added the test cases.
Please have a look.
Thank you.

addons/l10n_my/models/res_partner.py
vin-odoo: my example above is using a generator expression `(p for p in self if p.country_code == 'MY')`.
So it would yield the partners as it filters them, doing it in a single loop. (unlike using filtered, which would loop to filter, then return the resulting recordset to loop on it again)

In practice, you would rarely compute is_company on so many partners at once that it would really matter, though.

I'm fine with the if in the loop, that was just a suggestion anyway :)

vin-odoo (approved):
robodoo r+

LGTM thanks

PR #247938, opened 2026-02-10
[FIX] account_edi_ubl_cii: enable discount for null price product with charges

clom-odoo:
Allowances for Product with price as 0.00 aren't applied

Step to reproduce:
- import vendor bill from an XML having a product:
	- price: 0.00
	- charge: any positive amount
	- allowance: any positive amount

Current behavior:
- allowance isn't apply resulting in a difference between the XML total and Odoo total

Cause of the issue:
Before this commit the discount was applied as a percent of price only. Having a price as 0 prevent doing so.

opw-5499525

---

### DIFF
--- a/addons/account_edi_ubl_cii/tests/test_files/import/bis3/invoice/be/test_import_invoice_discount_on_price_zero.xml
+++ b/addons/account_edi_ubl_cii/tests/test_files/import/bis3/invoice/be/test_import_invoice_discount_on_price_zero.xml
@@ -1,0 +1,127 @@
+<Invoice xmlns="urn:oasis:names:specification:ubl:schema:xsd:Invoice-2" xmlns:cac="urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2" xmlns:cbc="urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2" xmlns:ext="urn:oasis:names:specification:ubl:schema:xsd:CommonExtensionComponents-2">
+    <cbc:CustomizationID>urn:cen.eu:en16931:2017#compliant#urn:fdc:peppol.eu:2017:poacc:billing:3.0</cbc:CustomizationID>
+    <cbc:ProfileID>urn:fdc:peppol.eu:2017:poacc:billing:01:1.0</cbc:ProfileID>
+    <cbc:ID>INV/2026/00001</cbc:ID>
+    <cbc:IssueDate>2026-04-15</cbc:IssueDate>
+    <cbc:DueDate>2026-04-15</cbc:DueDate>
+    <cbc:InvoiceTypeCode>380</cbc:InvoiceTypeCode>
+    <cbc:DocumentCurrencyCode>EUR</cbc:DocumentCurrencyCode>
+    <cac:OrderReference>
+        <cbc:ID>INV/2026/00001</cbc:ID>
+    </cac:OrderReference>
+    <cac:AccountingSupplierParty>
+        <cac:Party>
+            <cbc:EndpointID schemeID="0208">0202239951</cbc:EndpointID>
+            <cac:PartyName>
+                <cbc:Name>company_1_data</cbc:Name>
+            </cac:PartyName>
+            <cac:PostalAddress>
+                <cbc:StreetName>Chaussée de Namur 40</cbc:StreetName>
+                <cbc:CityName>Ramillies</cbc:CityName>
+                <cbc:PostalZone>1367</cbc:PostalZone>
+                <cac:Country>
+                    <cbc:IdentificationCode>BE</cbc:IdentificationCode>
+                </cac:Country>
+            </cac:PostalAddress>
+            <cac:PartyTaxScheme>
+                <cbc:CompanyID>BE0202239951</cbc:CompanyID>
+                <cac:TaxScheme>
+                    <cbc:ID>VAT</cbc:ID>
+                </cac:TaxScheme>
+            </cac:PartyTaxScheme>
+            <cac:PartyLegalEntity>
+                <cbc:RegistrationName>company_1_data</cbc:RegistrationName>
+                <cbc:CompanyID>BE0202239951</cbc:CompanyID>
+            </cac:PartyLegalEntity>
+            <cac:Contact>
+                <cbc:Name>company_1_data</cbc:Name>
+            </cac:Contact>
+        </cac:Party>
+    </cac:AccountingSupplierParty>
+    <cac:AccountingCustomerParty>
+        <cac:Party>
+            <cbc:EndpointID schemeID="0208">0477472701</cbc:EndpointID>
+            <cac:PartyName>
+                <cbc:Name>partner_a</cbc:Name>
+            </cac:PartyName>
+            <cac:PostalAddress>
+                <cbc:StreetName>Rue des Bourlottes 9</cbc:StreetName>
+                <cbc:CityName>Ramillies</cbc:CityName>
+                <cbc:PostalZone>1367</cbc:PostalZone>
+                <cac:Country>
+                    <cbc:IdentificationCode>BE</cbc:IdentificationCode>
+                </cac:Country>
+            </cac:PostalAddress>
+            <cac:PartyTaxScheme>
+                <cbc:CompanyID>BE0477472701</cbc:CompanyID>
+                <cac:TaxScheme>
+                    <cbc:ID>VAT</cbc:ID>
+                </cac:TaxScheme>
+            </cac:PartyTaxScheme>
+            <cac:PartyLegalEntity>
+                <cbc:RegistrationName>partner_a</cbc:RegistrationName>
+                <cbc:CompanyID>BE0477472701</cbc:CompanyID>
+            </cac:PartyLegalEntity>
+            <cac:Contact>
+                <cbc:Name>partner_a</cbc:Name>
+            </cac:Contact>
+        </cac:Party>
+    </cac:AccountingCustomerParty>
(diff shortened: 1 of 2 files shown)


addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -444,6 +444,52 @@ def test_import_discount(self):
...
+        })
+
+        imported_invoice = self._import_as_attachment_on(attachment=xml_attachment, journal=self.company_data["default_journal_sale"])
+        self.assertEqual(len(imported_invoice.invoice_line_ids), 3)
+        self.assertRecordValues(imported_invoice, [{'amount_total': 0.50}])
+        self.assertRecordValues(imported_invoice.invoice_line_ids, [
+            {'name': self.product_a.name, 'price_subtotal': 0.00},
+            {'name': 'Discount', 'price_subtotal': -0.50},
+            {'name': ' FREIGHT', 'price_subtotal': 1.0},
+            ])
```
agr-odoo: Indentation

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -444,6 +444,52 @@ def test_import_discount(self):
...
+      <ChargeIndicator>true</ChargeIndicator>
+      <AllowanceChargeReason>FREIGHT</AllowanceChargeReason>
+      <Amount currencyID="EUR">1.00</Amount>
+    </AllowanceCharge>
+    <AllowanceCharge>
+      <ChargeIndicator>false</ChargeIndicator>
+      <AllowanceChargeReason>DISCOUNTED FUEL SURCHARGE</AllowanceChargeReason>
+      <Amount currencyID="EUR">0.50</Amount>
+    </AllowanceCharge>
+</xpath>"""
```
agr-odoo: same comment as in the other PR :p

agr-odoo (commented):
Looks good to me, beside minor nitpicks

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -444,6 +444,52 @@ def test_import_discount(self):
...
+            'raw': etree.tostring(self.with_applied_xpath(my_invoice_root, modifying_xpath)),
+            'name': 'test_invoice.xml',
+        })
+
+        imported_invoice = self._import_as_attachment_on(attachment=xml_attachment, journal=self.company_data["default_journal_sale"])
+        self.assertEqual(len(imported_invoice.invoice_line_ids), 3)
+        self.assertRecordValues(imported_invoice, [{'amount_total': 0.50}])
+        self.assertRecordValues(imported_invoice.invoice_line_ids, [
+            {'name': self.product_a.name, 'price_subtotal': 0.00},
+            {'name': 'Discount', 'price_subtotal': -0.50},
```
william-andre: Why don't we keep `DISCOUNTED FUEL SURCHARGE`?

clom-odoo: Default behavior is to set discount as percent of product price. While implementing this mechanism, Odoo discard the information of discount reason in [`_retrieve_charge_allowance_vals`](https://github.com/odoo/odoo/blob/436921c24a531eba6bf57ffe3f7c3b4978139d83/addons/account_edi_ubl_cii/models/account_edi_common.py#L688-L705).

This commit introduces a fallback mechanism to set the discount in a new line instead of percent when the price is null.

The information was not displayable in normal behavior, so discarded. The fallback offers a place to display the information but doesn't have it.

william-andre: But why do we set the name of the line as "Discount" instead of "DISCOUNTED FUEL SURCHARGE"?

clom-odoo: The discount reason (DISCOUNTED FUEL SURCHARGE) was discarded in _retrieve_charge_allowance_vals which return the total amount of discounts/allowances, but no other information on them. As the normal behavior doesn't keep the reason and only applies one discount aggregating all the discounts/allowances, I thought it was fine to aggregate the discounts/allowances in one line named "Discount".

If you think it's better in this case to create one line per discount/allowance and keep their name, it's possible. I see two options for doing this. 
Either change the behavior behind _retrieve_charge_allowance_vals to also return allowances, and then filter between allowances and charges. I don't really like this option as it's not that stable. 
Or going back in the XML to find allowance reasons again. I don't really like this option as it's doing twice the same thing.

william-andre: You have the `reason` in `charges`, from `_retrieve_charge_allowance_vals` already

clom-odoo: I don't have the `reason` in `charges`.

DISCOUNTED FUEL SURCHARGE has its `charge_indicator` to `False`. Its amount is added to the `discount_amount` and nothing about it is added in `charges`.

I'm not really a fan of adding allowances to the list of returned values as it would change the method goal.

clom-odoo: Hello @william-andre
Could you take a look at this PR ?

william-andre: @clom-odoo What do you think of my fixup?

clom-odoo: I think it would have been the right fix, however as this https://github.com/odoo/odoo/pull/254111 PR remove the charges invoice lines it doesn't fit well anymore. I reworked it to adapt to the new charge system.

Not only that, but I also added a commit before these changes to comply with these (https://github.com/odoo/odoo/pull/245862 https://github.com/odoo/odoo/pull/251463) and avoid having the line discarded before applying the allowances.

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -708,23 +708,20 @@ def _retrieve_rebate_val(self, tree, xpath_dict, quantity):
...
-            else:
-                discount_amount += amount
-        return discount_amount, charges
+            charge_allowance.append({
+                'amount': amount if charge_indicator.lower() == 'true' else -amount,
+                'line_quantity': quantity,
+                'reason': reason,
+                'reason_code': reason_code,
+            })
+        return _discount_amount, charge_allowance  # TODO deprecated _discount_amount
```
william-andre: To make the linter happy (and no need to still declare it of course)
```suggestion
        return 0, charge_allowance  # TODO deprecated _discount_amount
```

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -946,11 +937,12 @@ def _retrieve_line_charges(self, record, line_values, taxes):
         """
         Handle the charges on the document line at import.
 
-        For each charge on the line, it creates a new aml.
+        Each charge on the line, is applied to the line.
         Special case: if the ReasonCode == 'AEO', there is a high chance the xml was produced by Odoo and the
         corresponding line had a fixed tax, so it first tries to find a matching fixed tax to apply to the current aml.
         """
         charges_vals = []
+        discount_amount = line_values['discount'] * line_values['price_unit'] * line_values['quantity']
```
william-andre: Isn't `discount` always 0?

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -965,11 +957,18 @@ def _retrieve_line_charges(self, record, line_values, taxes):
...
 
-            price_subtotal_before = line_values['price_unit'] * charge['line_quantity'] * (1.0 - line_values['discount'] / 100.0)
-            price_subtotal_after = price_subtotal_before + charge['amount']
-            line_values['price_unit'] += charge['amount'] / charge['line_quantity']
-            new_price_subtotal_before_discount = line_values['price_unit'] * charge['line_quantity']
-            line_values['discount'] = (1 - (price_subtotal_after / new_price_subtotal_before_discount)) * 100.0
+            if charge['amount'] > 0:
+                line_values['price_unit'] += charge['amount'] / charge['line_quantity']
+            else:
+                discount_amount -= charge['amount']
```
william-andre: Why is there a difference with positive or negative regarding the quantity handling?

william-andre (commented):
I'll let @smetl finish the review as he still has some work/fixes to do on this module

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -839,6 +839,14 @@ def _retrieve_line_vals(self, tree, document_type=False, qty_factor=1):
         if not float_is_zero(delivered_qty * price_unit, currency.decimal_places) and price_subtotal is not None:
             inferred_discount = 100 * (1 - (price_subtotal - charge_amount) / currency.round(delivered_qty * price_unit))
             discount = inferred_discount if not float_is_zero(inferred_discount, currency.decimal_places) else 0.0
+        elif discount_amount:
```
smetl, line 842: You have to rebase. This entire code is no longer used.

addons/account_edi_ubl_cii/tests/test_ubl_cii.py
```diff
@@ -414,6 +414,56 @@ def test_import_discount(self):
         for line in imported_invoice.invoice_line_ids:
             self.assertFalse(line.discount, "A discount on the imported lines signals a rounding error in the discount computation")
 
+    def test_import_discount_2(self):
```
smetl, line 417: You need to add this test in the new file dedicated to the extraction of the amounts for a single line after the rebase

smetl (changes requested):

PR #247943, opened 2026-02-10
[FIX] website: snippet popup & animations tour

pparidans:
Specific steps in the snippet_popup_and_animations tour work around the fact that the three columns snippet is visible or not while depending on the scroll position to hide/show its third column and test the related animation.

This introduces some indeterminism around the actual timing of the animated column when no scroll is actually performed (and the element actually in the viewport).

This commit forces the scroll to be performed and the snippet to actually be visible.

Note: this issue appears more frequently since Chrome 144+ and depending on the CPU load.

runbot-238402

### DIFF
--- a/addons/website/static/tests/tours/snippet_popup_and_animations.js
+++ b/addons/website/static/tests/tours/snippet_popup_and_animations.js
@@ -140,6 +140,10 @@ registerWebsitePreviewTour("snippet_popup_and_animations", {
         content: "Check that the popup has been closed",
         trigger: ".o_we_invisible_el_panel .o_we_invisible_entry:contains('Popup') i.fa-eye-slash",
     },
+    {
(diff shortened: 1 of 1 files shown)


FrancoisGe:
> > Not sure if it would be better to backport those or keep your fix, but at least here is the reason why it doesn't break in 19.0
> 
> The same for me. My only goal is that it doesn't break anymore 😁 @FrancoisGe your opinion ?

I agree with JULA, I ask to @shsa-odoo to backport. He is off, he will comeback tomorrow. If you want @Dj0ulo can already do the backport ?

shsa-odoo:
Hello :wave: https://github.com/odoo/odoo/pull/249505/files Here is the awaited backport :sweat_smile: . Please have a look. Thanks!!

PR #247945, opened 2026-02-10
[IMP] point_of_sale: mark invoices as reviewed

vlst-odoo:
Commit af368c872fcbdd952dffde47abcdc7d1fd02f298
introduced the `review_state` field on invoices.

With this, invoices from pos were marked with the default `To review` state. This does not make sense in pos, as pos only creates invoices after customer payment.

In this commit we ensure that pos invoices are marked as 'reviewed'.

Task: 5917658

---


davidmonnom:
@robodoo r+

PR #247948, opened 2026-02-10
[FIX] point_of_sale: detect refund+eWallet orders as refunds for invoice signs

Nokoz:
[FIX] point_of_sale: detect refund+eWallet orders as refunds for invoice signs
Refund orders paid through eWallet top-up can have a net total of 0, which
made refund detection based only on negative totals inconsistent. As a result,
some refund flows were treated as normal invoices and refund tax/invoice signs
were incorrect.

Steps to reproduce:
-------------------
* Configure an eWallet program in POS.
* Create and pay a POS order for one product, then invoice it.
* Refund that order and choose eWallet as refund payment method (refund + top-up).
* Validate and inspect the generated accounting document.

> Observation:
The refund flow may not be consistently treated as a refund when the order’s
net amount is 0, causing incorrect invoice move type/sign handling and wrong
tax booking behavior.

Why the fix:
------------
Refund detection now also relies on `refunded_order_id` in key paths:
* `_compute_prices`: apply refund factor when order is linked to a refunded order
  (or already negative), so totals/taxes keep refund semantics.
* `_prepare_invoice_vals`: create `out_refund` when the order is linked to a
  refunded order (or has negative total), ensuring a credit note is produced.
* `_prepare_base_line_for_taxes_computation`: consider refund context with
  `is_refund` or negative total for tax base sign consistency.

This keeps existing negative-total refund behavior while correctly handling
refund+eWallet cases where the arithmetic total can be 0.

opw-5426818

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -524,7 +524,7 @@ class PosOrder(models.Model):
                 company=order.company_id,
                 cash_rounding=cash_rounding,
             )
-            refund_factor = -1 if (order.amount_total < 0.0) else 1
+            refund_factor = -1 if (order.is_refund or order.amount_total < 0.0) else 1
             order.amount_tax = refund_factor * tax_totals['tax_amount_currency']
             order.amount_total = refund_factor * tax_totals['total_amount_currency']
             order.amount_difference = order.amount_paid - order.amount_total
@@ -898,7 +898,7 @@ class PosOrder(models.Model):
         fiscal_position = self.fiscal_position_id
         pos_config = self.config_id
         move_type = 'out_invoice' if not any(
-            order.is_refund or order.amount_total < 0 for order in self
+            order.is_refund or order.amount_total < 0.0 for order in self
         ) else 'out_refund'
         invoice_payment_term_id = (
             self.partner_id.property_payment_term_id.id
@@ -1888,7 +1888,7 @@ class PosOrderLine(models.Model):
         if fiscal_position:
             account = fiscal_position.map_account(account)
 
-        is_refund_order = line.order_id.amount_total < 0.0
+        is_refund_order = line.order_id.is_refund or line.order_id.amount_total < 0.0
         is_refund_line = line.qty * line.price_unit < 0
 
         lang = line.order_id.partner_id.lang or self.env.user.lang
--- a/addons/point_of_sale/tests/test_frontend.py
+++ b/addons/point_of_sale/tests/test_frontend.py
@@ -3159,15 +3159,16 @@ class TestUi(TestPointOfSaleHttpCommon):
             ('session_id', '=', self.main_pos_config.current_session_id.id)
         ])
         self.assertEqual(len(orders), 2, "Expected two orders: original and refund.")
-        order, refund_order = orders[0], orders[1]
+        original_order = next(o for o in orders if o.amount_total > 0)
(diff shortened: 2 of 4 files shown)


sbel-odoo (commented):
Fix looks good to me, I don't think there's a problem relying on refunded_order_id since in 19.0 you cannot modify an order you're currently refunding, as in you cannot add a product to the order you're refunding. Obviously ewallet is a special case. 

I'm also ok with the test, although I would've preferred it to be in loyalty originally. It was weird at first cause originally we cannot have another product in the refund order but if we wanted to simulate that flow without a tour in loyatly the test would have been the same. Since I don't mind it too much I'll leave it to the next person.

👍

addons/point_of_sale/models/pos_order.py
```diff
@@ -1884,8 +1878,14 @@ def _prepare_base_line_for_taxes_computation(self):
         if fiscal_position:
             account = fiscal_position.map_account(account)
 
-        is_refund_order = line.order_id.amount_total < 0.0
+        if self.env.context.get('invoicing'):
+            is_refund_order = bool(line.order_id.refunded_order_id or line.order_id.amount_total < 0.0)
+        else:
+            is_refund_order = bool(line.order_id.refunded_order_id)
```
sbel-odoo: Coming back after you've made a few changes to the original fix. I just want to make sure I understand why we don't consider the amount_total < 0 as being a refund. What scenario leads to amount_total < 0 and not being a refund?

sbel-odoo (commented):
I've had a look since the PR is slightly different than originally. I left a comment for clarity but I think I'm still good with the new version of this fix.
FYI: I did not test again your test w/ and w/o your fix. I'm assuming you haven't changed those and they were okay before.

addons/point_of_sale/models/pos_order.py
Nokoz: amount_total < 0 can happen without the Refund flow (no refunded_order_id), e.g. when the cashier enters a negative qty line (manual return/correction).
In that case, if the order is invoiced, we must create a credit note (out_refund); otherwise posting would fail because accounting forbids validating an invoice with a negative total.

addons/point_of_sale/models/pos_order.py
```diff
@@ -898,7 +898,7 @@ def _prepare_invoice_vals(self):
         fiscal_position = self.fiscal_position_id
         pos_config = self.config_id
         move_type = 'out_invoice' if not any(
-            order.is_refund or order.amount_total < 0 for order in self
+            order.refunded_order_id or order.amount_total < 0.0 for order in self
```
stva-odoo:
```suggestion
            order.is_refund or order.amount_total < 0 for order in self
```

addons/point_of_sale/models/pos_order.py
```diff
@@ -524,7 +524,7 @@ def _compute_prices(self):
                 company=order.company_id,
                 cash_rounding=cash_rounding,
             )
-            refund_factor = -1 if (order.amount_total < 0.0) else 1
+            refund_factor = -1 if (order.refunded_order_id or order.amount_total < 0.0) else 1
```
stva-odoo:
```suggestion
            refund_factor = -1 if (order.is_refund or order.amount_total < 0.0) else 1
```

stva-odoo (commented):
robodoo r+

PR #247950, opened 2026-02-10
[IMP] point_of_sale: add bg color to button

vlst-odoo:
The `Review` button accidentally had the class `secondary` instead of the class `btn-secondary`. It thus has not discernable border or bg color, making it hard unclear that it was a button.

This commit fixes the issue.

Task : 5910329

Before             |  After
:-------------------------:|:-------------------------:
 image |  image

---


adgu-odoo (commented):
robodoo r+

PR #247961, opened 2026-02-10
[FIX] l10n_tw_edi_ecpay_website_sale: fix paper invoice option for guest checkout

avar-odoo:
In Taiwan e-invoicing, B2C customers can request a paper copy of their invoice. When selected, the "Invoicing Info" step—which collects data like donation codes or carriers—should be skipped as it is not applicable to physical copies.

Previously, this logic failed during guest checkouts because the partner initially associated with the order is an archived public user. The persistent partner is only created/assigned after the address form is submitted.

This commit:
- Overrides `_create_or_update_address` instead of `_handle_extra_form_data` to ensure the paper format preference is saved on the correct, newly generated partner.
- Updates `_prepare_address_form_values` to correctly load existing preferences from the partner for registered users.
- Ensures the `l10n_tw_edi_is_print` flag on the Sales Order stays in sync with the partner's preference.

Task-5912985

---

### DIFF
--- a/addons/l10n_tw_edi_ecpay_website_sale/controllers/main.py
+++ b/addons/l10n_tw_edi_ecpay_website_sale/controllers/main.py
@@ -167,8 +167,9 @@ class WebsiteSaleL10nTW(WebsiteSale):
         **kwargs
     ):
         rendering_values = super()._prepare_address_form_values(*args, callback=callback, order_sudo=order_sudo, **kwargs)
-        if order_sudo:
-            rendering_values["l10n_tw_edi_require_paper_format"] = order_sudo.partner_id.l10n_tw_edi_require_paper_format
+        partner_sudo = kwargs.get('partner_sudo') or (args[0] if args else None)
+        if partner_sudo and request.website.sudo().company_id.country_id.code == 'TW' and request.website.sudo().company_id._is_ecpay_enabled():
+            rendering_values["l10n_tw_edi_require_paper_format"] = partner_sudo.l10n_tw_edi_require_paper_format
         return rendering_values
 
     def _validate_address_values(self, address_values, partner_sudo, address_type, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


addons/l10n_tw_edi_ecpay_website_sale/controllers/main.py
```diff
@@ -159,16 +159,10 @@ def check_love_code(self, sale_order_id, **kwargs):
...
-        *args,
-        callback='',
-        order_sudo=False,
-        **kwargs
-    ):
-        rendering_values = super()._prepare_address_form_values(*args, callback=callback, order_sudo=order_sudo, **kwargs)
-        if order_sudo:
-            rendering_values["l10n_tw_edi_require_paper_format"] = order_sudo.partner_id.l10n_tw_edi_require_paper_format
+    def _prepare_address_form_values(self, partner_sudo, *args, **kwargs):
+        rendering_values = super()._prepare_address_form_values(partner_sudo, *args, **kwargs)
```
vin-odoo: This is confusing, why not keep the same signature as the parent class?

addons/l10n_tw_edi_ecpay_website_sale/controllers/main.py
```diff
@@ -159,16 +159,10 @@ def check_love_code(self, sale_order_id, **kwargs):
...
+        if request.website.sudo().company_id.country_id.code == 'TW' and request.website.sudo().company_id._is_ecpay_enabled():
```
vin-odoo: You probably should check the `account_fiscal_country_id` code instead, here and in the other checks.
While it is usually going to match, it is the one used in the setting to show/hide ecpay. (I didn't pay attention last time, but let's correct it)

vin-odoo (approved):
robodoo r+

Thank you!

vin-odoo:
robodoo rebase-ff

PR #247965, opened 2026-02-10
[IMP] upgrade_code: add xml rendering context migration script

ged-odoo:
In owl3 all template variables which target component variable
must prefix `this.` to the variable. (as opposed to template
variables set with `t-set`, `t-for-each`... which do not need
to be prefixed with `this,`)

WHY: Allowing users to avoid writing `this` when accessing
component variables from the template is a nice shorthand but
add umbiguity illustrated by the complexity of this script.
Indeed the migration process highlighted many cases where
this ambiguity allowed incorrect code, see:

https://github.com/odoo/odoo/pull/249841, https://github.com/odoo/odoo/pull/249069, https://github.com/odoo/odoo/pull/238845, https://github.com/odoo/odoo/pull/248894

OWL2: The compiler would first look at the template context then
the component context when resolving variables.
OWL3: The compiler will throw if we don't put a `.this` template
variables.

Note: This script has already been run multiple times on the entire repo 
(first module by module,  and thereafter on all modules). You can find 
a mapping on the first module by module PRs for this script here: 
https://docs.google.com/spreadsheets/d/1kaZn9SgdRu8CuKRuvApctiz2qEsucb84XWMKYZANWJM/edit?gid=980192073#gid=980192073

---


odoo/upgrade_code/tools_etree.py
```diff
@@ -339,7 +339,7 @@ def _test():
                         test
                     </t>
                     <t t-if="len(resources)">
-                    
+
```
Gorash: The changing spaces implies that the code made unwanted modifications.

ged-odoo:
robodoo rebase-ff

ged-odoo:
robodoo r+

mcm-odoo:
@robodoo delegate+ r+

PR #236998, opened 2025-11-21
[REF] mail, *: introduce Store.FieldList

seb-odoo:
The main goal and the power of `Store` is to be able to use it declaratively by giving it a list of fields to add and letting `Store` handle the complexity of generating the result.

Many flows therefore simply have to handle lists of field name (or more complex field description). This is currently done manually in each flow which is cumbersome.

This commit introduces `Store.FieldList`, a small helper class to handle lists of field (name/description) compatible with `Store`, and allowing to more easily add new entries (`attr`, `one`, and `many`).

Usage of `Store.Attr`, `Store.One` and `Store.Many` have been removed thanks to this commit, as well as the `extra_fields` and `kwargs` of `store.add`, the magic call of `_to_store_defaults`, and finally `_to_store` is almost removed too.

Support of passing a single field (without list) is removed to avoid confusion and to keep a single syntax.

This commit also resolves the inconsistencies between the `env` of the records and the `env` used to call the field functions. Now the function is always bound to the records, which also allows easier syntax (simply giving the function name).

https://github.com/odoo/enterprise/pull/100237

task-4676427
task-4676467
task-4676469
task-4775128

PR #247973, opened 2026-02-10
[IMP] mail: access to notifications for portal users

maki-odoo:
Since #236998, a sudo access to mail notifications has been removed, which was not intended. Without sudo, if a portal user tries to read some messages containing notifications for other users, it crashes with an access error to `mail.notification`. This change restores that access.

Steps to reproduce:
- Mention a non portal user in a public channel.
- Open that channel as a portal user.
- The channel does not load due to an access error to `mail.notification`.


seb-odoo (approved):
@robodoo r+ delegate+

PR #247974, opened 2026-02-10
[FIX] tools: relax PDF mimetype check using startswith

phenix-factory:
The mimetype check was too strict and rejected values like `application/pdf;base64`.

This change fixes the issue by only verifying that the mimetype starts with `application/pdf`, ignoring any additional parameters.

runbot-231598


seb-odoo (approved):
@robodoo r+ delegate+

PR #247984, opened 2026-02-10
[FIX] crm: fix meeting display date with timezone

MAGREMENT:
# How to reproduce
- Use a browser extension to manage your browser's timezone
- Set your browser's timezone to a timezone with quite a big delay (like "America/Grand_Turk" if you live in Europe)
- Go to the form view of an opportunity
- Click on the smart button for meetings (Should be "No Meeting" if it is a new Opportunity)
- In the calendar view, add a new meeting for very late in the day (Example : 2026-02-10 22:00:00 => 23:00:00)
- Go back to the opportunity for view

# The problem
The date displayed is a day after the meeting that was just set up. Taking back our example, the date displayed would be 2026-02-11

# Why
The calendar view uses the browser's timezone to manage the dates. The smart button does not.

It is not possible to make the smart button use the browser's timezone, atleast in a clean way. That is because the smart button's data is managed by a python template, which does not have access to the browser's data. Trying to change the data displayed by the framework would be clunky as the html would need to be edited directly.

The fix that I implemented follows what the hr_appraisal module does for it's smart button with a date: use the timezone set in the user's preferences.

opw-5898520

---

### DIFF
--- a/addons/crm/models/crm_lead.py
+++ b/addons/crm/models/crm_lead.py
@@ -537,10 +537,10 @@ class Lead(models.Model):
                 lead.meeting_display_date = False
                 lead.meeting_display_label = _('No Meeting')
             elif lead_meeting_info['next_meeting_date']:
-                lead.meeting_display_date = lead_meeting_info['next_meeting_date']
+                lead.meeting_display_date = fields.Datetime.context_timestamp(lead, lead_meeting_info['next_meeting_date'])
                 lead.meeting_display_label = _('Next Meeting')
             else:
-                lead.meeting_display_date = lead_meeting_info['last_meeting_date']
+                lead.meeting_display_date = fields.Datetime.context_timestamp(lead, lead_meeting_info['last_meeting_date'])
                 lead.meeting_display_label = _('Last Meeting')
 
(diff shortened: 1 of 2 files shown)


ushyme (approved):
LGTM, thanks for your work!

addons/crm/models/crm_lead.py
```diff
@@ -537,10 +537,10 @@ def _compute_meeting_display(self):
                 lead.meeting_display_date = False
                 lead.meeting_display_label = _('No Meeting')
             elif lead_meeting_info['next_meeting_date']:
-                lead.meeting_display_date = lead_meeting_info['next_meeting_date']
+                lead.meeting_display_date = fields.Datetime.context_timestamp(self, lead_meeting_info['next_meeting_date'])
```
reth-odoo: Doesn't change anything currently, but logically a single record is expected
```suggestion
                lead.meeting_display_date = fields.Datetime.context_timestamp(lead, lead_meeting_info['next_meeting_date'])
```

reth-odoo (commented):
Just one note otherwise lgtm

@jeh-odoo coucou

jeh-odoo:
@robodoo r+

PR #247985, opened 2026-02-10
[FIX] web: kanban: ribbon misplaced when `ALT` is pressed

rfr-odoo:
The CSS rules of `.o_record_selection_available` are used to add a selection overlay on kanban cards.
These rules relied on `filter: brightness()` to slightly dim all child elements when pressing `ALT`.

However, using filter creates a new stacking context. As a result, when a ribbon is present, it no longer sticks to the card border and becomes misplaced.

This commit updates the rules to avoid that behavior and keep the ribbon correctly positioned.

---


Iucapad (approved):
@robodoo r+

PR #247986, opened 2026-02-10
[FIX] l10n_es_pos: fix singleton error on refund

aksi-odoo:
Step to reproduce:
- install l10n_es_pos
- create a pos, open its setting and set its `Simplified Invoice` with a journal
- start pos, create a order and refund it

Observation:
- we receive a singleton error for account.move

Cause:
- when calling `get_invoice_name` method, we pass `order_server_ids` which contains order and refund order id, hence two ids are passed

Fix:
- instead of using `order_server_ids` we use 'order.id' i.e. current order

opw-5870707

---


alsh-odoo (approved):

robinengels (approved):

jorenvo (approved):
Thank you, @aksi-odoo!

@robodoo r+

PR #247987, opened 2026-02-10
[IMP] hr_holidays: correct balance report

dhbr-odoo:
Previously, the balance report was filtering allocations and leaves per year.
Due to this when allocation exceeds the period, it will shows wrong data.

Example:
- Allocation of 20 days from the start of 2025 to the end of 2026
- 10 days taken in 2025 and 5 days taken in 2026
- Viewing the report for 2026 showed 15 days available and 5 days taken,
while it should show 5 days available and 15 days taken.
- Also, currently it doesn't track changes in days for accrual allocations.

This improvement redesigns the balance report to display the current (today’s)
balance only, fully aligned with the employee Time Off dashboard.

- Display only 'Allocated' and 'Left' balances
- Removed period/year-based filtering
- Add default 'Time Left' filter and a 'Partial Workers' filter
(employees with a work time rate below 100%)
- Add group-by options for Department, Manager, Job Position, and
Working Schedule

Approach:-
The report is redesigned as a current balance snapshot per employee and leave
type. The actual allocated/left figures are computed in Python via the
same allocation data method the employee dashboard itself relies on,
rather than being independently re-derived in SQL.

This guarantees the report's numbers match the dashboard by construction,
since both read from the same source of truth.

task-[4915926](https://www.odoo.com/odoo/project/1251/tasks/4915926)
---


addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,131 +1,161 @@
...
-                FROM hr_leave_allocation as allocation
-                INNER JOIN hr_employee as employee ON (allocation.employee_id = employee.id)
-                LEFT JOIN hr_version v ON v.id = employee.current_version_id
+    def _generate_balance_query(self):
+        today = fields.Date.today()
+        if self.env.context.get('active_ids'):
+            employees = self.env['hr.employee'].search([
+                ('id', 'in', self.env.context['active_ids']),
+                ('company_id', 'in', self.env.companies.ids),
+            ])
```
ravb-odoo: Store active_ids in a variable and use it in both places instead of fetching it from the context every time.

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,131 +1,161 @@
...
+        else:
+            employees = self.env['hr.employee'].search([
+                ('company_id', 'in', self.env.companies.ids),
+            ])
```
ravb-odoo:
```suggestion
        active_ids = self.env.context.get('active_ids')
        domain = [
             ('company_id', 'in', self.env.companies.ids),
         ]
         if active_ids:
             domain.append(('id', 'in', active_ids))
             
         employees = self.env['hr.employee'].search(domain)
```

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,131 +1,161 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from odoo import api, fields, models, tools, _
+from odoo import api, fields, models, _
```
ravb-odoo: Replace all occurrences of _ with self.env._ in entire file.

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,131 +1,161 @@
...
+        if not allocations:
+            return self._empty_query()
 
-                /* Obtain the minimum id for a given employee and type of leave */
-                LEFT JOIN
-                    (SELECT employee_id, work_entry_type_id, min(id) as min_id
-                    FROM hr_leave_allocation GROUP BY employee_id, work_entry_type_id) min_allocation_id
-                on (allocation.employee_id=min_allocation_id.employee_id and allocation.work_entry_type_id=min_allocation_id.work_entry_type_id)
+        work_entry_types = allocations.work_entry_type_id
+        leaves_taken = employees._get_consumed_leaves(work_entry_types)[0]
```
ravb-odoo:
```suggestion
        consumed_leaves_by_employee = employees._get_consumed_leaves(work_entry_types)[0]
```
more suitable

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,131 +1,161 @@
...
+                    row_id,
+                    employee.id,
+                    employee.active,
+                    max(0.0, remaining_days),
+                    max(0.0, remaining_hours),
+                    department_id,
+                    work_entry_type.id,
+                    'left',
+                    company_id,
+                ))
```
ravb-odoo:
```suggestion
                records.extend([
                (
                    row_id,
                    employee_id,
                    employee_active,
                    allocated_days,
                    allocated_hours,
                    department_id,
                    work_entry_type_id,
                    'allocated',
                    company_id,
                ),
                (
                    row_id + 1,
                    employee_id,
                    employee_active,
                    max(0.0, remaining_days),
                    max(0.0, remaining_hours),
                    department_id,
                    work_entry_type_id,
                    'left',
                    company_id,
                ),
            ])
            row_id += 2

```

ravb-odoo (changes requested):
Thanks for your work. Here are some comments.
Would it be possible to make the report using SQL instead of Python? With large datasets, the Python implementation might be slightly slower.

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,239 +1,163 @@
...
-                            OR
-                            vl.date_to >= oa.date_from
-                        )
-                    GROUP BY oa.allocation_id
+                        l.employee_id AS employee_id,
+                        l.work_entry_type_id AS work_entry_type_id,
+                        l.number_of_days AS number_of_days,
+                        l.number_of_hours AS number_of_hours
+                    FROM hr_leave l
+                    WHERE l.state IN ('confirm', 'validate1', 'validate')
```
kais-odoo: This deducts all leaves for the employee regardless of whether the leave overlaps a currently valid allocation.

This can make today’s balance wrong when a leave was consumed by an expired allocation, or belongs to a future allocation period. `_get_consumed_leaves()` only consumes leaves against allocations whose validity period overlaps the leave, so the SQL report should apply the same overlap condition before aggregating deducted leaves.

addons/hr_holidays/report/hr_leave_employee_type_report.py
```diff
@@ -1,239 +1,163 @@
...
-                    LEFT JOIN hr_version v ON v.id = employee.current_version_id
-                    WHERE request.state IN ('confirm', 'validate', 'validate1')
-                ) leaves
+                    UNION ALL
+
+                    /* Left: remaining balance = allocated - consumed (floored at 0) */
+                    SELECT
+                        at.employee_id AS employee_id,
+                        at.active_employee AS active_employee,
+                        GREATEST(at.total_allocated_days - COALESCE(lt.total_taken_days, 0), 0) AS number_of_days,
```
kais-odoo: The remaining balance is always floored with `GREATEST(..., 0)`, but leave types with `allows_negative` can legitimately have a negative remaining balance on the dashboard.

Since this report should align with dashboard behavior, the floor should only apply when the leave type does not allow negative balances. For `allows_negative` types, the report should show the actual allocated minus consumed value.

addons/hr_holidays/report/hr_leave_employee_type_report.xml
```diff
@@ -6,10 +6,22 @@
...
                 <field name="employee_id"/>
-                <field name="date_from"/>
-                <filter name="year" date="date_from" default_period="year" string="Period"/>
+                <filter name="left_balance" string="Time Left" domain="[('holiday_status', '=', 'left')]"/>
+                <separator/>
+                <filter 
+                    name="partial_workers" 
+                    string="Partial Workers" 
+                    domain="[('working_schedule.work_time_rate', '&lt;', 1.0)]"
+                />
```
kais-odoo:
```suggestion
                <filter name="partial_workers"  string="Partial Workers" 
                    domain="[('working_schedule.work_time_rate', '&lt;', 1.0)]"/>
```
white space

addons/hr_holidays/report/hr_leave_employee_type_report.xml
```diff
@@ -6,10 +6,22 @@
...
                 <filter string="Company" name="company" context="{'group_by':'company_id'}" groups="base.group_multi_company"/>
                 <filter string="Employee" name="employee" context="{'group_by':'employee_id'}"/>
+                <filter string="Department" name="department_id" context="{'group_by':'department_id'}"/>
+                <filter string="Manager" name="manager_id" context="{'group_by':'manager_id'}"/>
+                <filter string="Job Position" name="job_id" context="{'group_by':'job_id'}"/>
+                <filter
```
kais-odoo: here too

addons/hr_holidays/report/hr_leave_employee_type_report.py
dhbr-odoo: Thanks for the feedback!
I have applied the changes

addons/hr_holidays/report/hr_leave_employee_type_report.py
dhbr-odoo: Thanks for the feedback!
I have applied the changes.

PR #247990, opened 2026-02-10
[IMP] mail: improve message search ui 

phenix-factory:
* Use the primary color for the search button when the input field contains something.
* trigger the search while typing after a small debounce. 

This PR implements task-5867464 point 38

### DIFF
--- a/addons/mail/static/src/core/common/search_message_input.js
+++ b/addons/mail/static/src/core/common/search_message_input.js
@@ -27,7 +27,7 @@ export class SearchMessageInput extends Component {
 
     setup() {
         super.setup();
-        this.state = useState({ searchTerm: "", searchedTerm: "" });
+        this.state = useState({ searchTerm: "" });
         useAutofocus();
         useExternalListener(
             browser,
@@ -44,31 +44,30 @@ export class SearchMessageInput extends Component {
     search() {
         this.props.messageSearch.searchTerm = this.state.searchTerm;
         this.props.messageSearch.search();
-        this.state.searchedTerm = this.state.searchTerm;
     }
 
     clear() {
         this.state.searchTerm = "";
-        this.state.searchedTerm = this.state.searchTerm;
         this.props.messageSearch.clear();
+    }
+
+    onClickClose() {
+        this.clear();
         this.props.closeSearch?.();
     }
 
-    onClickClearSearch() {
-        this.state.searchTerm = "";
-        this.state.searchedTerm = "";
-        this.props.messageSearch.clear();
-    }
-
-    onKeydownSearch(ev) {
-        if (ev.key !== "Enter") {
+    onInputSearch(ev) {
+        if (!this.state.searchTerm) {
+            return this.clear();
+        }
+        if (
+            this.state.searchTerm.startsWith(this.props.messageSearch.searchTerm) &&
+            this.props.messageSearch.searched &&
+            this.props.messageSearch.count === 0
+        ) {
             return;
         }
-        if (!this.state.searchTerm) {
-            this.clear();
-        } else {
-            this.search();
-        }
+        this.search();
     }
 
     /** @param {SearchFilter} searchFilter */
--- a/addons/mail/static/src/core/common/search_message_input.xml
+++ b/addons/mail/static/src/core/common/search_message_input.xml
@@ -5,14 +5,13 @@
             <div class="input-group">
                 <div class="o_searchview form-control d-flex align-items-center bg-view p-0" role="search" aria-autocomplete="list">
                     <div class="o_searchview_input_container d-flex align-items-center flex-grow-1 flex-wrap gap-1 h-100">
-                        <input type="text" class="o_searchview_input flex-grow-1 w-auto border-0 rounded-start ps-2 bg-view" accesskey="Q" t-att-placeholder="inputPlaceholder" t-custom-model="state.searchTerm" t-on-keydown="onKeydownSearch" t-custom-ref="autofocus" role="searchbox"/>
-                        <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover" t-att-class="{ 'invisible': !state.searchedTerm }" role="img" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.onClickClearSearch()"/>
+                        <div class="input-group-text bg-transparent border-0 pe-0">
+                            <i class="oi oi-search opacity-25"/>
+                        </div>
+                        <input type="text" class="o_searchview_input form-control flex-grow-1 w-auto border-0 rounded-start ps-2 bg-view" accesskey="Q" t-att-placeholder="inputPlaceholder" t-custom-model="state.searchTerm" t-on-input="onInputSearch" t-custom-ref="autofocus" role="searchbox"/>
+                        <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover cursor-pointer" t-att-class="{ 'invisible': !state.searchTerm }" role="button" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.clear()"/>
                     </div>
                 </div>
-                <button class="btn" t-att-class="state.searchedTerm === state.searchTerm ? 'btn-outline-secondary' : 'btn-secondary'" t-on-click="() => this.search()" aria-label="Search button">
-                    <i t-if="!props.messageSearch.searching" class="o_searchview_icon oi oi-search" role="img" aria-label="Search Messages" title="Search Messages"/>
-                    <i t-else="" class="fa fa-circle-o-notch fa-spin" aria-label="Search in progress" title="Search in progress"/>
-                </button>
             </div>
             <Dropdown t-if="env.inChatter">
                 <button class="btn btn-outline-secondary ms-2" aria-label="Filter Messages" title="Filter Messages">
@@ -27,7 +26,7 @@
                     </t>
                 </t>
             </Dropdown>
-            <button t-if="env.inChatter" class="btn btn-outline-secondary ms-2" t-on-click="() => this.clear()" aria-label="Close button">
+            <button t-if="env.inChatter" class="btn btn-outline-secondary ms-2" t-on-click="() => this.onClickClose()" aria-label="Close button">
                 <i class="o_searchview_icon oi oi-close cursor-pointer" role="img" aria-label="Close search" title="Close"/>
             </button>
         </div>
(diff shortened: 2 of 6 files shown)


addons/mail/static/src/core/common/search_message_input.js
```diff
@@ -38,6 +39,7 @@ export class SearchMessageInput extends Component {
             },
             { capture: true }
         );
+        this.debouncedSearch = useDebounced(this.search, 250);
```
alexkuhn: I would not define a debounce at all:
`useMessageSearch` is already using `useSequential()`, which awaits resolution of next search until the previous one has finished (and silently drop the intermediate that should not initiate an RPC)

So this can only "flood" the server when the server responds fast... but that means the server is chilling, so no reason to slow down.

Intermediate search can be useful. The extra latency can add confusion, especially when the UI is not handling it properly (= more LOCs). For example, with the current logic of adding a debounce, many other UI elements are not aware of a debounced search being queued, such as:

- there's no spinner on the search icon (that occurs when there's RPC ongoing, but not with debounce in queue).
- search results show intermediate values, but will be shortly replaced by delayed RPC, again that's frustrating because the UI feels slow.


So really, don't bother with `useDebounced()`, because this puts more harm to UX. We ensure to not flood the server thanks to `useSequential()`, which has been designed to give the best tradeoff: immediately do RPC when fresh, and when not then await until ongoing RPC has finished. The `useMessageSearch` puts `searching = true / false` around the `sequential`, so it ensures there's no visible flicker of 250ms of no spinner for example.

addons/mail/static/src/core/common/search_message_input.js
```diff
@@ -46,11 +48,13 @@ export class SearchMessageInput extends Component {
         this.state.searchedTerm = this.state.searchTerm;
     }
 
-    clear() {
+    clear(options = { close: true }) {
```
alexkuhn: Only the clear from clicking on the dedicated button is doing that.
Rather than having a parameter, better define a `onClickClose` handler that does the clear and calls `props.closeSearch?.()`.

addons/mail/static/src/core/common/search_message_input.xml
```diff
@@ -5,11 +5,11 @@
...
+                        <input type="text" class="o_searchview_input flex-grow-1 w-auto border-0 rounded-start ps-2 bg-view" accesskey="Q" t-att-placeholder="inputPlaceholder" t-model="state.searchTerm" t-on-input="onInputSearch" t-ref="autofocus" role="searchbox"/>
                         <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover" t-att-class="{ 'invisible': !state.searchedTerm }" role="img" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.onClickClearSearch()"/>
                     </div>
                 </div>
-                <button class="btn" t-att-class="state.searchedTerm === state.searchTerm ? 'btn-outline-secondary' : 'btn-secondary'" t-on-click="() => this.search()" aria-label="Search button">
+                <button class="btn" t-att-class="state.searchedTerm === state.searchTerm ? 'btn-outline-secondary' : 'btn-primary'" t-on-click="() => this.search()" aria-label="Search button">
```
alexkuhn: I would revert the `btn-primary`. The specs say either auto-trigger the search when inputing search terms or keep the manual trigger but with `btn-primary` to incite clicking to register the search.

Since the auto-trigger of search input has been implemented here (the preferred choice), then no `btn-primary` change as no one will bother to click on this button. The change of style just adds undesirable visual changes of
- 0ms: `btn-secondary` => `btn-primary`
- 250ms: `btn-primary` => `fa fa-circle-o-notch fa-spin`
- 250ms + RPC return: `fa fa-circle-o-notch fa-spin` => `btn-secondary`

When this could simply be:
- 0ms: `btn-secondary` => `fa fa-circle-o-notch fa-spin`
- RPC return: `fa fa-circle-o-notch fa-spin` => `btn-secondary`

Which is immediately working with just triggering `.search()` from the input.

alexkuhn (commented):
Don't forget to adapt tests to remove the useless interaction to validate the search with `triggerHotkey("Enter")`.

LGTM otherwise!

addons/mail/static/src/core/common/search_message_input.js
phenix-factory: On locale, the spinner looks more like a flicker than anything. I'll look into a better solution

addons/mail/static/src/core/common/search_message_input.xml
```diff
@@ -5,11 +5,11 @@
...
+                        <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover cursor-pointer" t-att-class="{ 'invisible': !state.searchTerm }" role="img" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.clear()"/>
                     </div>
                 </div>
-                <button class="btn" t-att-class="state.searchedTerm === state.searchTerm ? 'btn-outline-secondary' : 'btn-secondary'" t-on-click="() => this.search()" aria-label="Search button">
+                <button class="btn btn-outline-secondary" t-on-click="() => this.search()" aria-label="Search button">
                     <i t-if="!props.messageSearch.searching" class="o_searchview_icon oi oi-search" role="img" aria-label="Search Messages" title="Search Messages"/>
                     <i t-else="" class="fa fa-circle-o-notch fa-spin" aria-label="Search in progress" title="Search in progress"/>
                 </button>
```
alexkuhn: Since typing text in input is enough to trigger the search, there's no point to have click on button.
We can have the following look:
Image

So the search icon in prefix is not clickable, just there to make it look more as an input to search (like the "Search conversations" in discuss sidebar), and it dynamically changes to spinner like now.

alexkuhn (commented):
LGTM otherwise.

addons/mail/static/src/core/common/search_message_input.xml
phenix-factory: Button could be useful to "retry" in case something went wrong?

alexkuhn: I would still remove it, as if this doesn't work this is a bug to fix, and when it works (which should be 100%), having the button just make UI bloated / confusing.

phenix-factory: Button removed

addons/mail/static/src/core/common/search_message_input.js
```diff
@@ -44,31 +44,23 @@ export class SearchMessageInput extends Component {
...
-        }
+    onInputSearch(ev) {
         if (!this.state.searchTerm) {
-            this.clear();
-        } else {
-            this.search();
+            return this.clear();
         }
+        this.search();
     }
```
alexkuhn, line 71: If search term had no results and you type more text, it should not RPC again like in suggestion hook:
https://github.com/odoo/odoo/blob/master/addons/mail/static/src/core/common/suggestion_hook.js#L35

addons/mail/static/src/core/common/search_message_input.xml
```diff
@@ -5,14 +5,10 @@
             <div class="input-group">
                 <div class="o_searchview form-control d-flex align-items-center bg-view p-0" role="search" aria-autocomplete="list">
                     <div class="o_searchview_input_container d-flex align-items-center flex-grow-1 flex-wrap gap-1 h-100">
-                        <input type="text" class="o_searchview_input flex-grow-1 w-auto border-0 rounded-start ps-2 bg-view" accesskey="Q" t-att-placeholder="inputPlaceholder" t-custom-model="state.searchTerm" t-on-keydown="onKeydownSearch" t-custom-ref="autofocus" role="searchbox"/>
-                        <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover" t-att-class="{ 'invisible': !state.searchedTerm }" role="img" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.onClickClearSearch()"/>
+                        <input type="text" class="o_searchview_input flex-grow-1 w-auto border-0 rounded-start ps-2 bg-view" accesskey="Q" t-att-placeholder="inputPlaceholder" t-model="state.searchTerm" t-on-input="onInputSearch" t-custom-ref="autofocus" role="searchbox"/>
```
alexkuhn: Search input is too thin. This is fixed by adding `.form-control` on `input` that is expected on almost all inputs.

alexkuhn: Should keep `t-custom-model`. This happens very often for your PRs to silently revert code when there are conflicts.

When you have conflicts, please make sure to follow this process:
- keep previous code
- then redo the work you did on top of it

Don't dismiss old code, otherwise you get almost all the time the conflicts resolved by dismissing recent changes.

addons/mail/static/src/core/common/search_message_input.xml
```diff
@@ -5,14 +5,10 @@
...
+                        <i class="fa fa-fw fa-times-circle me-2 opacity-50 opacity-100-hover cursor-pointer" t-att-class="{ 'invisible': !state.searchTerm }" role="img" aria-label="Clear Search" title="Clear Search" t-on-click="() => this.clear()"/>
```
alexkuhn: Already there before this PR, but could change `role="img"` to `role="button"`.

alexkuhn: I would put `oi oi-search` as prefix in input.

While this looks redundant with the icon of the search panel, this icon helps identify purpose of the input as a search.

This is notably important in chatter, where the input is quite confusing visually without this icon:

Image

Compared to, which is better:

Image

In search panel you can make it look like this:
Image

So having icon opacity to 25%. It keeps it visible and makes UI intuitive, while being shy enough to not bloat UI even when next to another `oi-search` like in panel title.

addons/mail/static/src/core/common/search_message_input.xml
phenix-factory: I just missed that change

phenix-factory: It doesn't seem to change the height of the input, but sure I'll add form-control

addons/mail/static/src/core/common/search_message_input.js
phenix-factory: Not sure it's worth it, but I have updated the code to do that.

addons/mail/static/src/core/common/search_message_input.xml
alexkuhn: For me it does. Before:
Screenshot 2026-03-09 at 15 34 02

After:
Screenshot 2026-03-09 at 15 34 13

So the `form-control` on `input` is good ;)

alexkuhn (approved):
@robodoo delegate+ r+

PR #248000, opened 2026-02-10
[FIX] pos_loyalty: keep reward when loading an order from TicketScreen

Nokoz:
Reward disappeared from an order when it was loaded from the Orders tab (TicketScreen) in restaurant mode after working on other tables.

Steps to reproduce:
-------------------
* Configure a Buy X Get Y (or similar) loyalty program and open a POS restaurant session.
* On table A, create an order that triggers the program and confirm the reward line is applied.
* Leave table A, create or edit another order on a different table (so another order becomes current).
* Go to the Orders tab (TicketScreen), select the order from table A and click "Load Order".

> Observation:
The order from table A is loaded without its reward line, even though it was present when the order was first created.

Why the fix:
------------
TicketScreen’s "Load Order" flow was only switching the current order without refreshing loyalty state, so after changing tables and coming back, later loyalty recomputations could drop the existing reward lines.

opw-5909899

### DIFF
--- a/addons/pos_loyalty/static/src/app/screens/ticket_screen/ticket_screen.js
+++ b/addons/pos_loyalty/static/src/app/screens/ticket_screen/ticket_screen.js
@@ -11,6 +11,12 @@ patch(TicketScreen.prototype, {
         super.setup(...arguments);
         this.notification = useService("notification");
     },
+    async setOrder(order) {
+        await super.setOrder(...arguments);
+        if (order && this.pos.models["loyalty.program"]?.length) {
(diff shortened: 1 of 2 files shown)


robinengels:
@Nokoz Where exactly are we losing the reward line ? Because when we leave the table the lines should be saved in the backend and we shouldn't be losing the reward line. How is it different than just opening the table ?

manv-afk (approved):

adgu-odoo (commented):
robodoo r+

PR #248003, opened 2026-02-10
[FIX] core: ignore AssetsLoadingError after tour termination

pparidans:
Similarly to commit https://github.com/odoo/odoo/commit/493bab4f460dd4069d5cb6805933b8088067ff17
hiding "failed to fetch" errors, this commit adds AssetsLoadingError as
those represents "just" another category of failed assets request (i.e.
lazy loaded) after tour termination.

runbot-233826

### DIFF
--- a/odoo/tests/common.py
+++ b/odoo/tests/common.py
@@ -112,6 +112,7 @@ TEST_CURSOR_COOKIE_NAME = 'test_request_key'
 IGNORED_MSGS = re.compile(r"""
     failed\ to\ fetch  # base error
   | connectionlosterror:  # conversion by offlineFailToFetchErrorHandler
+  | assetsloadingerror: # lazy loaded bundle
 """, flags=re.VERBOSE | re.IGNORECASE).search
 
 def get_db_name():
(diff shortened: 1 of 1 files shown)


odoo/tests/common.py
```diff
@@ -109,6 +109,13 @@ def __getattr__(name):
 BROWSER_WAIT = CHECK_BROWSER_SLEEP * CHECK_BROWSER_ITERATIONS # seconds
 TEST_CURSOR_COOKIE_NAME = 'test_request_key'
 
+IGNORED_MSGS = re.compile(r"""
+    (?: failed\ to\ fetch  # base error
```
xmo-odoo: The group is not needed, since there's *only* an alternation.

xmo-odoo (commented):
Commit message review

> Similarly to commit [^1] hiding "failed to fetch" errors, this commit
also adds AssetsLoadingError as those represents "just" another category
of failed assets request (i.e. lazy loaded) after tour termination.
>
> [^1]: odoo/odoo@493bab4

 `[^anchor]` is used for footnotes, or ancillary notes, but here it's used for the main text, I don't think it's useful, it just makes reading the commit message more difficult. And the "also" in the commit message makes no sense, adding AssetsLoadingError to the ignored causes is *the only thing* the commit does.

PR #248005, opened 2026-02-10
[FIX] account_edi_ubl_cii: Fix SB tests if account_peppol not installed

antoine162:
Commit 0163627275 reintroduced the tests for exporting vendor bills in the BIS3 Self Billing format. However, at the moment, these tests fail if `account_peppol` is not installed.

This is because the Send & Print framework prevents sending vendor bills by default and is overridden only in `account_peppol`.

To test the BIS3 Self Billing export in `account_peppol`, we need to patch the `AccountMoveSend._get_move_constraints` method in order to export vendor bills in the test.

runbot-238802
runbot-238803


gawa-odoo (commented):
@robodoo r+

PR #248010, opened 2026-02-10
18.0 opw 5471809 fix agr

agr-odoo:
**Issue 1**

Steps to reproduce:
- Create a new invoice
- Add an invoice line, with just a description
- Save
- On the same line, add any product
- Print

Issue:
Only the custom description will be visible on the line

opw-5471809

---

**Issue 2**

Steps to reproduce:
- Import factur-x e-invoice
- Print pdf

Issue:
Only product name will be printed, even if the invoice record include
the complete description

opw-5471809


RicGR98 (approved):
LGTM, thanks !


@robodoo r+

PR #248014, opened 2026-02-10
[FIX] base,website,test_mass_mailing: pregenerate binary bundles

pparidans:
The commit [^1] introducing binary asset bundle support overlooked the pregeneration of said bundles for the tests runs. This leads to hot-regeneration of those bundles during tests runs on the runbot (multiple hundreds of times) instead of only once and reusing them.

This commit adds support for those binary bundles during pregeneration.

[^1]: odoo/odoo@a5c02da5c24bfc85b3bbb7d1410d489d3c7185b8

### DIFF
--- a/odoo/addons/base/models/ir_qweb.py
+++ b/odoo/addons/base/models/ir_qweb.py
@@ -2962,13 +2962,19 @@ class IrQweb(models.AbstractModel):
         """
         _logger.runbot('Pregenerating assets bundles')
 
-        js_bundles, css_bundles = self._get_bundles_to_pregenarate()
+        js_bundles, css_bundles, bin_bundles = self._get_bundles_to_pregenarate()
 
         links = []
         start = time.time()
         for bundle in sorted(js_bundles):
             links += self._get_asset_bundle(bundle, css=False, js=True).js()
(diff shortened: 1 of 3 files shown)


odoo/addons/base/models/ir_qweb.py
```diff
@@ -2962,14 +2962,20 @@ def _pregenerate_assets_bundles(self):
...
         links = []
         start = time.time()
         for bundle in sorted(js_bundles):
             links += self._get_asset_bundle(bundle, css=False, js=True).js()
         _logger.info('JS Assets bundles generated in %s seconds', time.time()-start)
         start = time.time()
+        for bundle in sorted(bin_bundles):
+            binary_bundle = self._get_asset_bundle(bundle, binary=True)
+            for bin in binary_bundle.binaries:
+                links += binary_bundle.bin(bin.extension)
```
xmo-odoo:
```suggestion
                links.extend(binary_bundle.bin(bin.extension))
```
I dislike `list.__iadd__` as a rule, but here it's even more confusing than usual as the bundle methods return recordsets, and recordsets have their own version of (i)add.

PR #248016, opened 2026-02-10
[FIX] website_sale: products configurator "Add to cart" text-wrap

estoj:
This commit fixes some text wrapping issues for the buttons inside the product configurator:

- Change "Add to Cart" text visibility from d-md-inline to d-lg-inline for better mobile layout and prevent text wrapping
- Add text-nowrap class to configurator's footer primary button to prevent text wrapping

task-5896776

| Before | After |
|--------|--------|
| image | image | 

---

### DIFF
--- a/addons/website_sale/static/src/js/product/product.xml
+++ b/addons/website_sale/static/src/js/product/product.xml
@@ -9,7 +9,7 @@
         </QuantityButtons>
         <button name="sale_product_configurator_add_button" position="attributes">
             <attribute name="t-if">this.props.can_be_sold</attribute>
-            <attribute name="class" remove="btn-secondary" add="btn-outline-primary" separator=" "/>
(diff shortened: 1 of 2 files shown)


addons/website_sale/static/src/js/product_configurator_dialog/product_configurator_dialog.scss
anso-odoo, line 1: I think we can use `.text-nowrap` directly on the button instead, so we don't need to add extra SCSS

anso-odoo (changes requested):
Hello @estoj 

Good job, just a small suggestion

Thanks !

anso-odoo (approved):

ltinel (approved):
LGTM, thanks :)
@robodoo delegate+

PR #248017, opened 2026-02-10
[IMP] orm: allow upercase letters on regex_alphanumeric matching

hazemayman1:
Relax the condition on regex_alphanumeric to allow for uppercase letters A-Z
along with already allowed lowercase letters and underscores.

Currently it is only used by field properties to verify property names

---


tivisse:
@robodoo r+

PR #248022, opened 2026-02-10
[FIX] html_builder: remove wrong tooltip for installable snippet groups

bvr-odoo:
Since this commit [1], installable snippet groups like "Blogs" or
"Events" are shown as “not droppable” on the page when they are not
installed, which is incorrect.

Installable snippet groups should never be marked as “not droppable”
unless all groups are disabled (e.g. in the mega menu).
This commit fixes that.

[1]: https://github.com/odoo/odoo/commit/ae4824640665fc639e03a13c341f18e73060349e

### DIFF
--- a/addons/html_builder/static/src/core/disable_snippets_plugin.js
+++ b/addons/html_builder/static/src/core/disable_snippets_plugin.js
@@ -104,6 +104,12 @@ export class DisableSnippetsPlugin extends Plugin {
         // Disable the groups containing only disabled snippets.
         if (!areGroupsDisabled) {
             snippetGroups.forEach((snippetGroup) => {
+                // Do not mark uninstalled groups as disabled (unless all
+                // groups are disabled).
+                if (snippetGroup.isInstallable) {
+                    snippetGroup.isDisabled = false;
+                    return;
+                }
                 if (snippetGroup.groupName !== "custom") {
                     snippetGroup.isDisabled = !snippets.find(
                         (snippet) =>


addons/html_builder/static/src/core/disable_snippets_plugin.js
```diff
@@ -103,19 +103,22 @@ export class DisableSnippetsPlugin extends Plugin {
...
-                } else {
-                    const customSnippets = this.snippetModel.snippetsByCategory["snippet_custom"];
-                    snippetGroup.isDisabled = !customSnippets.find(
-                        (snippet) => !snippet.isDisabled
-                    );
-                }
-            });
+            snippetGroups
+                .filter((snippetGroup) => !snippetGroup.isInstallable)
+                .forEach((snippetGroup) => {
```
sobo-odoo: Given that this code is run a lot, in order to prevent having two loops and so to be more efficient, I would remove the filter and simply return inside the `forEach` if the snippet group is not installable :slightly_smiling_face: 
(It will also remove a level of indentation :slightly_smiling_face:)

That's the fix I made (but didn't commit...) + I also planned to add that comment if you want to add it (to explain a bit) :slightly_smiling_face: 
```js
snippetGroups.forEach((snippetGroup) => {
  // Do not mark uninstalled groups as disabled (unless all
  // groups are disabled).
  if (snippetGroup.isInstallable) {
       return;
   }
   ...
```

addons/html_builder/static/src/core/disable_snippets_plugin.js
```diff
@@ -103,19 +103,22 @@ export class DisableSnippetsPlugin extends Plugin {
```
sobo-odoo, line 103: Commit message:
> Installable snippet groups should never be marked as “not droppable”.

Technically, this is wrong, because if all groups are disabled (because there is nowhere to drop e.g. in the mega menu) even the installable should be marked as undroppable in that case :slightly_smiling_face: You should add "unless all groups are disabled" to be correct :slightly_smiling_face:

sobo-odoo (changes requested):
Here is my review :slightly_smiling_face: Small change to be more efficient :+1:

sobo-odoo (approved):

@robodoo r+

PR #248023, opened 2026-02-10
[FIX] account_peppol: prevent singleton error when Peppol is disabled

iacio-odoo:
Steps to reproduce:
**Note: This does not occur in Peppol demo mode**
1. Install the Peppol module and enable it in Settings
2. Create a portal user, then go to the 'Accounting' tab in their contact record
3. Under Customer Invoices, set the following:
a. Invoice Sending -> By Peppol
b. eInvoice Format -> Peppol Bis 3.0
c. Belgian VAT -> A valid VAT
4. Click 'Verify' under 'Peppol Status' and this should change `peppol_verification_status` to `valid`
5. Disable Peppol in Settings
6. Log in as the created user and make a purchase on the web shop
7. Observe that no sale order confirmation email is sent, and a singleton error is logged.

When Peppol is installed, but disabled, the `account_peppol_edi_user` field is not set (or gets unset) on the company, which can lead to a singleton error when trying to send an invoice via Peppol. This can occur if Peppol was previously enabled and there were `res.partner` records with Peppol information that was validated. The database checks this after E-Commerce transactions and, if there is no `invoice_sending_method` set on the `res.partner` record, it defaults to Peppol, even if it is disabled, which can cause the singleton error. This change will ensure that we check to see if the company has registered with Peppol before using it to send invoices.

[Ticket link](https://www.odoo.com/odoo/project.task/5890740)
opw-5890740

### DIFF
--- a/addons/account_peppol/models/account_move_send.py
+++ b/addons/account_peppol/models/account_move_send.py
@@ -123,7 +123,7 @@ class AccountMoveSend(models.AbstractModel):
     def _is_applicable_to_company(self, method, company):
         # EXTENDS 'account'
         if method == 'peppol':
-            return company.country_code in PEPPOL_LIST and company.account_peppol_proxy_state != 'rejected'
+            return company.country_code in PEPPOL_LIST and company.account_peppol_proxy_state not in ['rejected', 'not_registered']
         else:
             return super()._is_applicable_to_company(method, company)
 
@@ -135,7 +135,7 @@ class AccountMoveSend(models.AbstractModel):
             return all([
                 self._is_applicable_to_company(method, move.company_id),
                 self.env['res.partner']._get_peppol_verification_state(partner.peppol_endpoint, partner.peppol_eas, invoice_edi_format) == 'valid',
-                move.company_id.account_peppol_proxy_state != 'rejected',
+                move.company_id.account_peppol_proxy_state not in ['rejected', 'not_registered'],
                 move._need_ubl_cii_xml(invoice_edi_format)
                 or move.ubl_cii_xml_id and move.peppol_move_state not in ('processing', 'done'),
             ])


addons/account_peppol/models/account_move_send.py
```diff
@@ -218,6 +218,9 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
 
         edi_user = next(iter(invoices_data)).company_id.account_peppol_edi_user
 
+        if not edi_user:
```
william-andre: This can come from multiple different companies, seems like there are more issues than just a traceback 👀

iacio-odoo: this is one of two spots where i debated on making the fix. the other spot would be [here](https://github.com/odoo/odoo/blob/636465d5eb0aaa78eeb9a47fd0b09be3b97b5072/addons/account_peppol/models/account_move_send.py#L14). it seems to be overriding the default sending method set on `res.partner` if they dont have an invoice sending method set. this can put it into a state where it tries to send with peppol even when it's disabled, since it never directly checks if it's actually enabled or not. this could be another problem spot, i will give it another look over though and revise!🫡

addons/account_peppol/models/account_move_send.py
```diff
@@ -123,7 +123,7 @@ def _do_peppol_pre_send(self, moves):
     def _is_applicable_to_company(self, method, company):
         # EXTENDS 'account'
         if method == 'peppol':
-            return company.country_code in PEPPOL_LIST and company.account_peppol_proxy_state != 'rejected'
+            return company.country_code in PEPPOL_LIST and company.account_peppol_proxy_state not in ['rejected', 'not_registered']
```
william-andre, line 126: This fix seems to makes sense[^1], but it still doesn't explain why it is possible to have multiple companies related to `invoices_data` in `_call_web_service_after_invoice_pdf_render`, especially when coming from the CRON `_cron_account_move_send`

Isn't there still an issue there?

[^1]: I'm not a Peppol expert, still waiting for a review from them

iacio-odoo: i'm also a bit confused on this. i haven't seen issues related to that on my ticket specifically, but it could be related to [this commit](https://github.com/odoo/odoo/commit/4bfe16cd45828a864a159b566d7983246e7e03a5) which seems to deal with multi company workflows. this is the only spot i can see being related, though, and even then i'm not entirely sure

iacio-odoo:
@clbr-odoo @william-andre came back to this to write a test, but after some searching, it looks like this issue has been resolved by another PR that was recently merged. i tried reproducing the issue with the new code and was unable to. i think we should be able to close my PR, but i just want to double check before i do so

https://github.com/odoo/odoo/pull/254671

PR #248025, opened 2026-02-10
[IMP] crm: add team switcher

amdi-odoo:
Purpose
=======
Ease the CRM teams navigation by adding a CRM team switcher next to the search bar.

Specification
=============
The team switcher is a dropdown with all the CRM teams the user has read access to.
Selecting a team in the dropdown will:
- apply a filter on the team to display every related "crm.lead" records
- display the selected team stages
- use the selected team as default on "crm.lead" records creation,
but also when generating leads from crm iap mine, or when creating
leads from a business card picture (cf enterprise PR).

The selected team will be saved in local storage and be restored in every views
which are using the switcher. The first team acts as default.
An "All" option is also available in the switcher displaying every
available records and team stages.

The team switcher will be displayed if:
- the "show_team_switcher" context key is set on the action/view
- the "team_id" field is defined in the search view
- there's at least one accessible team

Removing the "Teams" menu which is now obsolete.
Teams settings are still available from the "Sales Teams" configuration menu.

Task-5887454

---

### DIFF
--- a/addons/crm/models/crm_lead.py
+++ b/addons/crm/models/crm_lead.py
@@ -984,19 +984,18 @@ class CrmLead(models.Model):
 
     @api.model
     def _read_group_stage_ids(self, stages, domain):
-        # retrieve team_id from the context and write the domain
-        # - ('id', 'in', stages.ids): add columns that should be present
-        # - OR ('fold', '=', False): add default columns that are not folded
-        # - OR ('team_ids', '=', team_id), ('fold', '=', False) if team_id: add team columns that are not folded
-        team_id = self.env.context.get('default_team_id')
-        team_ids = self.env.user.crm_team_ids._ids if self.env.context.get('show_user_team_stages') else ()
-        team_ids += (team_id,) if team_id else ()
-        search_domain = ['|', ('id', 'in', stages.ids), ('team_ids', '=', False)]
-        if team_ids:
-            search_domain = ['|', ('id', 'in', stages.ids), '|', ('team_ids', '=', False), ('team_ids', 'in', team_ids)]
+        default_team_id = self.env.context.get('default_team_id')
+        if self.env.context.get('show_team_switcher') and default_team_id:
+            team_ids = [default_team_id]
+        else:
+            team_ids = self.env.user.crm_team_ids.ids if self.env.context.get('show_user_team_stages') else []
+            team_ids += [default_team_id] if default_team_id else []
 
         # perform search
-        stage_ids = stages.sudo()._search(search_domain, order=stages._order)
+        stage_ids = stages.sudo()._search(
+            Domain('id', 'in', stages.ids) | self.env['crm.stage']._get_visible_stages_domain(team_ids),
+            order=stages._order
+        )
         return stages.browse(stage_ids)
 
     def _stage_find(self, team_id=False, domain=None, order='sequence, id', limit=1):
--- a/addons/crm/models/crm_stage.py
+++ b/addons/crm/models/crm_stage.py
@@ -1,6 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo import _, api, fields, models
+from odoo.fields import Domain
 from odoo.tools import sql
 
 AVAILABLE_PRIORITIES = [
@@ -34,6 +35,11 @@ class CrmStage(models.Model):
         help='This stage is folded in the kanban view when there are no records in that stage to display.')
     # This field for interface only
     team_count = fields.Integer('team_count', compute='_compute_team_count')
+
+    @api.model
+    def _get_visible_stages_domain(self, team_ids):
+        """ Returns the "crm.stage" domain to know which stages are visible for the given team ids. """
+        return Domain("team_ids", "in", team_ids) | Domain("team_ids", "=", False)
 
     def _auto_init(self):
         # TODO stop hardcoding ids in tests and remove this
--- a/addons/crm/models/crm_team.py
+++ b/addons/crm/models/crm_team.py
@@ -785,3 +785,41 @@ class CrmTeam(models.Model):
         if self.use_opportunities:
             return self.action_open_opportunities()
         return super().action_primary_channel_button()
+
+    # ------------------------------------------------------------
+    # UTILS
+    # ------------------------------------------------------------
+
+    @api.model
+    def get_team_switcher_teams_data(self):
+        """ Retrieve the teams and their data for the team switcher.
+        Only considering the teams using opportunities.
+        Limited to own teams if the user doesn't have "All Leads" access.
+        """
+        domain = Domain("use_opportunities", "=", True)
+        if not self.env.user.has_group("sales_team.group_sale_salesman_all_leads"):
+            domain &= Domain("id", "in", self.env.user.crm_team_ids.ids)
+        teams = self.env["crm.team"].search(domain, order="sequence, id")
+
+        team_stages_mapping = self.env["crm.stage"]._read_group(
+            self.env["crm.stage"]._get_visible_stages_domain(teams.ids),
+            groupby=["team_ids"],
+            aggregates=["id:recordset"],
+        )
+        stage_ids_by_team = {team.id: stages.ids for team, stages in team_stages_mapping}
+        shared_stage_ids = stage_ids_by_team.get(False, [])
+        return [
+            {
+                'id': team.id,
+                'name': team.name,
+                # Domain applied when this team will be selected in the team switcher.
+                # Showing all "crm.lead" records assigned to the team + the unassigned ones which
+                # happens to be part of the team visible stages.
+                'switcher_domain': list(
+                    Domain("team_id", "=", team.id) | (
+                        Domain("team_id", "=", False) &
+                        Domain("stage_id", "in", stage_ids_by_team.get(team.id, []) + shared_stage_ids)
+                    )
+                ),
+            } for team in teams
+        ]
--- a/addons/crm/models/ir_http.py
+++ b/addons/crm/models/ir_http.py
@@ -1,0 +1,11 @@
+from odoo import models
+
+
+class IrHttp(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    def session_info(self):
+        result = super().session_info()
+        if self.env.user._is_internal():
+            result['sales_team_membership_multi'] = self.env['ir.config_parameter'].sudo().get_bool('sales_team.membership_multi')
+        return result
--- a/addons/crm/static/src/components/breadcrumbs/crm_breadcrumbs.xml
+++ b/addons/crm/static/src/components/breadcrumbs/crm_breadcrumbs.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+    <t t-name="crm.Breadcrumbs">
+        <t t-if="this.showTeamSwitcher">
+            <TeamSwitcher/>
+            <t t-call="web.Breadcrumb.Actions"/>
+            <t t-call-slot="breadcrumb-status-indicator"/>
+        </t>
+        <t t-else="" t-call="web.Breadcrumbs"/>
+    </t>
+
+</templates>
--- a/addons/crm/static/src/components/team_switcher/team_switcher.xml
+++ b/addons/crm/static/src/components/team_switcher/team_switcher.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="crm.team_switcher">
+        <Dropdown menuClass="'crm_team_switcher_dropdown d-flex flex-column'">
+            <button class="o_cp_team_switcher text-truncate fs-4 btn btn-link text-reset fw-normal"
+                t-attf-class="{{this.hasDropdown ? 'o-dropdown-caret' : 'pe-none'}}" t-att-data-tooltip="this.currentLabel">
+                <span class="text-truncate" t-out="this.currentLabel"/>
+            </button>
+            <t t-set-slot="content">
+                <div class="overflow-auto flex-grow-1">
+                    <t t-foreach="this.teams" t-as="team" t-key="team_index">
+                        <DropdownItem onSelected="() => this.onSelect(team.id)" class="'text-truncate ' + (this.selectedTeamId === team.id ? 'selected' : '')">
+                            <t t-out="team.name"/>
+                        </DropdownItem>
+                    </t>
+                </div>
+                <t t-if="this.isSaleManager">
+                    <div class="dropdown-divider flex-shrink-0"/>
+                    <DropdownItem onSelected="() => this.onClickManageTeams()" class="'text-truncate flex-shrink-0'">
+                        Manage Teams
+                    </DropdownItem>
+                </t>
+            </t>
+        </Dropdown>
+    </t>
+
+</templates>
--- a/addons/crm/static/src/views/crm_activity/crm_activity_view.js
+++ b/addons/crm/static/src/views/crm_activity/crm_activity_view.js
@@ -1,0 +1,12 @@
+import { activityView } from "@mail/views/web/activity/activity_view";
+import { CrmControlPanel } from "@crm/views/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { registry } from "@web/core/registry";
+
+export const crmActivityView = {
+    ...activityView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+};
+
+registry.category("views").add("crm_activity", crmActivityView);
--- a/addons/crm/static/src/views/crm_graph/crm_graph_view.js
+++ b/addons/crm/static/src/views/crm_graph/crm_graph_view.js
@@ -1,0 +1,12 @@
+import { CrmControlPanel } from "@crm/views/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { graphView } from "@web/views/graph/graph_view";
+import { registry } from "@web/core/registry";
+
+export const crmGraphView = {
+    ...graphView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+};
+
+registry.category("views").add("crm_graph", crmGraphView);
--- a/addons/crm/static/src/views/crm_pivot/crm_pivot_view.js
+++ b/addons/crm/static/src/views/crm_pivot/crm_pivot_view.js
@@ -1,0 +1,12 @@
+import { CrmControlPanel } from "@crm/views/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { pivotView } from "@web/views/pivot/pivot_view";
+import { registry } from "@web/core/registry";
+
+export const crmPivotView = {
+    ...pivotView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+};
+
+registry.category("views").add("crm_pivot", crmPivotView);
--- a/addons/crm/static/src/views/crm_search_model.js
+++ b/addons/crm/static/src/views/crm_search_model.js
@@ -1,0 +1,217 @@
+import { browser } from "@web/core/browser/browser";
+import { Domain } from "@web/core/domain";
+import { _t } from "@web/core/l10n/translation";
+import { SearchModel } from "@web/search/search_model";
+import { session } from "@web/session";
+
+import { computed, proxy } from "@odoo/owl";
+
+export class CrmSearchModel extends SearchModel {
+
+    switcherTeam = computed(() => this.state.switcherTeams.find((t) => t.id === this.state.switcherTeamId));
+
+    /**
+     * @override
+     */
+    setup() {
+        super.setup(...arguments);
+        this.state = proxy({
+            switcherTeams: [],
+            switcherTeamId: undefined,
+        });
+    }
+
+    get isTeamSwitcherEnabled() {
+        return session.sales_team_membership_multi && this._actionContext.show_team_switcher;
+    }
+
+    /**
+     * @override
+     * Init the switcher teams and selection.
+     */
+    async load(config) {
+        // Keep a reference to the original action context so "_updateActionContext" can
+        // later patch it to affect the context of the "New" button form view.
+        this._actionContext = config.context;
+        await this._initSwitcher(config);
+        await super.load(config);
+    }
+
+    /**
+     * @override
+     * Export team switcher state with the config state.
+     */
+    exportState() {
+        const state = super.exportState();
+        state.teamSwitcherState = {
+            teams: this.state.switcherTeams,
+            teamId: this.state.switcherTeamId,
+        };
+        return state;
+    }
+
+    /**
+     * @override
+     * Restore team switcher state when the config state is imported (i.e. on switch view).
+     */
+    _importState(state) {
+        super._importState(...arguments);
+        if (state.teamSwitcherState) {
+            this.state.switcherTeams = state.teamSwitcherState.teams;
+            this.state.switcherTeamId = state.teamSwitcherState.teamId;
+        }
+    }
+
+    /**
+     * @override
+     * Update the search context so the selected team is used as the default when
+     * creating crm.lead records via quick create, business card, ... and creating crm.stage records.
+     * Also ensure that only stages related to the selected team are displayed (see _read_group_stage_ids).
+     */
+    _getContext() {
+        const context = super._getContext();
+        if (!this.state.switcherTeamId) {
+            return context;
+        }
+        return {
+            ...context,
+            default_team_id: this.state.switcherTeamId,
+            default_team_ids: [this.state.switcherTeamId], // for stages
+        };
+    }
+
+    /**
+     * @override
+     * Update search domain depending on the team switcher selection.
+     * Showing all "crm.lead" records assigned to the team + the unassigned ones which
+     * happens to be part of the team visible stages.
+     */
+    _getDomain(params = {}) {
+        const domain = super._getDomain({ ...params, raw: true }); // Force raw to simplify
+        const team = this.switcherTeam();
+        if (!team) {
+            return params.raw ? domain : domain.toList(this.domainEvalContext);
+        }
+        const switcherDomain = Domain.and([domain, new Domain(team.switcher_domain)]);
+        return params.raw ? switcherDomain : switcherDomain.toList(this.domainEvalContext);
+    }
+
+    /**
+     * Initialize the team switcher by:
+     * - retrieving the list of crm teams to display
+     * - restoring the previously selected team
+     *
+     * The initialization is skipped if the team switcher state is present
+     * in the config state as it'll be restored with "_importState".
+     */
+    async _initSwitcher(config) {
+        if (!this.isTeamSwitcherEnabled || config.state?.teamSwitcherState) {
+            return;
+        }
+        // Retrieve teams to display in team switcher
+        this.state.switcherTeams = await this.orm
+            .cache({
+                type: "disk",
+                update: "always",
+                callback: (result, hasChanged) => {
+                    if (hasChanged) {
+                        this.state.switcherTeams = result;
+                        this._initSwitcherSelection(true);
+                    }
+                },
+            })
+            .call("crm.team", "get_team_switcher_teams_data");
+        this._initSwitcherSelection();
+    }
+
+    /**
+     * Init the switcher selected team by retrieving it from
+     * the local storage or fallback on first team.
+     */
+    _initSwitcherSelection(loaded=false) {
+        let teamId = JSON.parse(browser.localStorage.getItem("crm.switcher_team_id"));
+        const isValid = this.state.switcherTeams.find((t) => t.id === teamId);
+        if (!isValid) {
+            // Fallback on first team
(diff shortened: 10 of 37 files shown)


addons/crm/static/src/views/crm_control_panel/crm_control_panel.js
```diff
@@ -0,0 +1,68 @@
...
+    setup() {
+        super.setup();
+        this.selectedTeamIdKey = "selectedTeamId";
+        this.state.selectedTeamId = JSON.parse(browser.localStorage.getItem(this.selectedTeamIdKey));
+
+        onWillStart(async () => {
+            this.accessibleTeams = await this.orm
+                .cache({
+                    type: "disk",
+                    update: "always",
```
aab-odoo: following my discussion with lna, we agreed to update only once (i.e. we do the rpc once, and we never do it again until we reload the page), so `update="once", which is the default

addons/crm/static/src/views/crm_calendar/crm_calendar_model.js
```diff
@@ -0,0 +1,10 @@
+import { CalendarModel } from "@web/views/calendar/calendar_model";
+import { CrmTeamSwitcherModelMixin } from "@crm/views/crm_control_panel/crm_team_switcher_model_mixin";
+
+export class CrmCalendarModel extends CrmTeamSwitcherModelMixin(CalendarModel) {
+    async load(params = {}) {
+        const domain = params.domain || this.meta.domain;
+        params.domain = this._processSearchDomain(params, domain);
```
aab-odoo: This pattern exposes an issue with the spreadsheet integration: when the user click on "Add to spreadsheet", they need to retrieve the current domain (non evaluated) to be able to generate the correct data inside the spreadsheet. As they need the not yet evaluated domain, they retrieve it from the search model, so before you insert the team_id part. As a consequence, the team filter won't be applied in the spreadsheet.

They already have the issue with the project sub tasks, and we couldn't find a proper way to fix it...

aab-odoo (commented):
Modulo my remarks, that's indeed what we discussed with lna

addons/crm/static/src/views/crm_control_panel/crm_control_panel.js
amdi-odoo: Hello @aab-odoo :slightly_smiling_face: 

About this :thinking:
When update "always" was set, if the `selectedTeamId` wasn't in the `accessibleTeams`, the selected team id was removed from the local storage and the search was re-triggered to fallback on the "All" selection (cf the cache code here)

But with update "once", if I have a `selectedTeamId` set to `3` from a past DB, and the new DB doesn't have the team of id 3, the search domain is updated to include 3 in my mixin (even though the team doesn't exist), and here the update isn't triggered meaning it doesn't fallback on "All".

Also, second point, I'm not sure re-triggering the `.search()` of the search model in the `onWillStart` is a good idea as we could have a slight view flicker when it happens.

I think we need to fallback on "All" in the mixin when the `selectedTeamId` isn't in the accessible teams to prevent re-triggering the search and be able to update the local storage before the control panel processes the value.
But retrieving the accessible teams in the mixin doesn't seem a good idea either:
- we will need the accessible teams in 2 places, 2 rpcs?
- I still have the crash when awaiting in the load method of kanban and list views (cf my discussion in the js channel on discord).

Do you have any idea on how to make it work ? 

Thanks!

amdi-odoo: Maybe if I retrieve the accessibleTeams in the `onWillStart` of the `searchModel` ?
And then just access it in the mixin and the control panel using `this.env.searchModel.accessibleTeams` :thinking: 

EDIT: I updated like so

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,148 @@
...
+            switcherTeamId: undefined,
+        });
+    }
+
+    /**
+     * @override
+     */
+    async load(config) {
+        await this._initSwitcher(config);
+        super.load(config);
```
tde-banana-odoo: Shouldn't you *await* for super ? Otherwise you might end up with some race condition, isn't it ?

addons/crm/views/crm_menu_views.xml
```diff
@@ -32,13 +32,6 @@
         groups="sales_team.group_sale_salesman"
         action="crm.crm_lead_action_my_activities"
         sequence="2"/>
-
-    <menuitem
```
tde-banana-odoo, line 36: I saw something in spec about getting "Stages" menu out of debug mode, is it possible ? Was it discarded ?

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,148 @@
...
+        const savedTeamId = JSON.parse(browser.localStorage.getItem("crm.switcher_team_id"));
+        const isValid = savedTeamId === false || this.state.switcherTeams.some((t) => t.id === savedTeamId);
+        this.state.switcherTeamId = isValid ? savedTeamId : this.state.switcherTeams[0]?.id ?? false;
+    }
+
+    /**
+     * Update the team switcher selected team.
+     * @param {Number | false} teamId Id of the new selected team ("All" = false).
+     */
+    async _updateSwitcherSelection(teamId) {
```
tde-banana-odoo: Do wee need the async here ?

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,148 @@
...
+        this.state.switcherTeamId = teamId;
+        browser.localStorage.setItem("crm.switcher_team_id", teamId);
```
tde-banana-odoo: Shouldn't we be symmetric to the setter and use json.stringify ? Would probably be more explicit with potential false handling, ...

addons/crm_iap_mine/static/src/components/lead_generation_dropdown/crm_lead_gen_element.js
```diff
@@ -19,6 +19,7 @@ patch(LeadGenerationDropdown.prototype, {
 
     async openLeadGenerationForm() {
         const action = await this.orm.call("crm.lead", "action_generate_leads", []);
+        Object.assign(action.context, this.env.searchModel.context);
```
tde-banana-odoo: Shouldn't we only fetch default_team_id from the searchModel context ? Isn't it going to propagate potentially unwanted keys ?

addons/crm/static/src/views/crm_activity/crm_activity_view.js
```diff
@@ -0,0 +1,12 @@
+import { activityView } from "@mail/views/web/activity/activity_view";
+import { CrmControlPanel } from "@crm/views/crm_control_panel/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { registry } from "@web/core/registry";
+
+export const crmActivityView = {
+    ...activityView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+}
```
tde-banana-odoo: :socks:
```suggestion
};
```

addons/crm/static/src/views/crm_control_panel/crm_control_panel.js
```diff
@@ -0,0 +1,10 @@
+import { ControlPanel } from "@web/search/control_panel/control_panel";
+import { TeamSwitcher } from "@crm/components/team_switcher/team_switcher";
+
+export class CrmControlPanel extends ControlPanel {
+    static template = "crm.ControlPanel";
+    static components = {
+        ...ControlPanel.components,
+        TeamSwitcher,
+    };
+}
```
tde-banana-odoo: :socks: 
```suggestion
};
```

addons/crm/static/src/views/crm_graph/crm_graph_view.js
```diff
@@ -0,0 +1,12 @@
+import { CrmControlPanel } from "@crm/views/crm_control_panel/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { graphView } from "@web/views/graph/graph_view";
+import { registry } from "@web/core/registry";
+
+export const crmGraphView = {
+    ...graphView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+}
```
tde-banana-odoo: :socks: 
```suggestion
};
```

addons/crm/static/src/views/crm_pivot/crm_pivot_view.js
```diff
@@ -0,0 +1,12 @@
+import { CrmControlPanel } from "@crm/views/crm_control_panel/crm_control_panel";
+import { CrmSearchModel } from "@crm/views/crm_search_model";
+import { pivotView } from "@web/views/pivot/pivot_view";
+import { registry } from "@web/core/registry";
+
+export const crmPivotView = {
+    ...pivotView,
+    ControlPanel: CrmControlPanel,
+    SearchModel: CrmSearchModel,
+}
```
tde-banana-odoo: :socks: 
```suggestion
};
```

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,148 @@
...
+        const teamIds = this.state.switcherTeamId
+            ? [this.state.switcherTeamId] // Specific team
+            : this.state.switcherTeams.map((t) => t.id); // All teams
+        const switcherDomain = new Domain([["team_id", "in", teamIds]]);
+        if (params.raw) {
+            return Domain.and([domain, switcherDomain]);
+        }
+        return Domain.and([new Domain(domain), switcherDomain]).toList(this.domainEvalContext);
+    }
+}
```
tde-banana-odoo: :socks: 
```suggestion
};
```

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,148 @@
...
+            ...switcherContext,
+        };
+    }
+
+    /**
+     * @override
+     * Update search domain depending on the team switcher selection.
+     */
+    _getDomain(params = {}) {
+        const domain = super._getDomain(params);
```
tde-banana-odoo: Note that I think we could simplify a bit raw / list management if we call super with force raw=True, allowing to always have domains, then simply returning a toList 'd version of the result if original raw was true.

addons/crm/models/crm_team.py
```diff
@@ -785,3 +785,19 @@ def action_primary_channel_button(self):
...
+
+    @api.model
+    def get_team_switcher_teams(self):
+        """ Retrieve the current user team ids and names for the team switcher.
+        Only considering the teams using opportunities.
+        Limited to own teams if the user doesn't have "All Leads" access.
+        """
+        domain = [("use_opportunities", "=", True)]
+        if not self.env.user.has_group("sales_team.group_sale_salesman_all_leads"):
+            domain = Domain.AND([domain, [("id", "in", self.env.user.crm_team_ids._ids)]])
```
reth-odoo:
```suggestion
            domain = Domain.AND([domain, [("id", "in", self.env.user.crm_team_ids.ids)]])
```

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,218 @@
...
+            switcherTeamId: undefined,
+        });
+    }
+
+    /**
+     * @override
+     * Init the switcher teams and selection.
+     */
+    async load(config) {
+        this._actionContext = config.context;
```
reth-odoo, line 35: Seems strange to "steal" the context like this, as we can then try to update it after the load which probably would break things.

As we pass the config anyways, can't we just directly access that and set it there?

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,218 @@
...
+     */
+    _updateSwitcherSelection(teamId) {
+        this.state.switcherTeamId = teamId;
+        if (teamId) {
+            browser.localStorage.setItem("crm.switcher_team_id", JSON.stringify(teamId));
+        } else {
+            browser.localStorage.removeItem("crm.switcher_team_id");
+        }
+        this._updateActionContext(teamId);
+        this._notify();
```
reth-odoo: Can't we call `reload` with a new context?
Especially as it seems we want to rely on the context for the search, but as we don't update the domain here it doesn't seem like we'll necessarily reload

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,218 @@
...
+        if (teamId === this.state.switcherTeamId) {
+            // Already current one, nothing to do
+            return;
+        }
+        if (this._loaded) {
+            // Already loaded: trigger a manual update to save the team,
+            // and recompute the search context and domain.
+            this._updateSwitcherSelection(teamId);
+            return;
+        }
```
reth-odoo: When do we reach this actually? Only in the callback from caching?

Isn't it possible that `_loaded` is not set, but calling `_updateActionContext` is not sufficient, if `load` is already past the point of reading its context?

```
                                               load
         ------------------------------------------------------------------------
                      0 |                                                4 |
------------------------------------------------                       data fetched
|     get_team_switcher_teams                  |                         6 |
------------------------------------------------                       load finished
  1 | (cached [1,2,3])                  3 | (new [2,3])
_initSwitcherSelection         _initSwitcherSelection
  2 |                                   5 |
_updateActionContext           _updateActionContext
```

data is fetched using the original (now wrong) context at 4, then context is updated at 5 (too late) and we don't reload

Would say once you enter `super.load`, we already shouldn't be trying to update the context anymore since we don't necessarily know what's happening in there. Although it does mean reloading every time the cache was stale (which is rarely, I should assume)


------------------------------------

But maybe it's fine and I'm overthinking it. Didn't dive much into `load`.

addons/crm/static/src/views/crm_search_model.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * @override
+     * In offline mode, the team switcher dropdown is disabled to reduce complexity.
+     * Instead, showing the current search selected team in the search facet.
+     */
+    getCurrentSearch() {
+        const search = super.getCurrentSearch();
+        const team = this.state.switcherTeams.find((t) => t.id === this.state.switcherTeamId);
```
reth-odoo: As we do this a few times, might be worth having the new `computed()` values that are kept up to date with signals/proxy

addons/crm/static/src/views/crm_search_model.js
amdi-odoo: Not sure I understand 🤔 
We are indeed updating the `config.context` after the load using `_updateActionContext` and then `notify` to take it into account 😄 
That's the only way I found to actually modify the action context (`config.context` is actually derived from the action context)
We have `globalContext` but it's retrieved from the `config.context` everytime, meaning it won't work I need to update the context from the config directly 

Did you have any idea in mind?

addons/crm/static/src/views/crm_search_model.js
amdi-odoo: The `notify` triggers an update, a re-fetch, which a `reload` doesn't do.
The reload also re-fetch the `search_default` from the context which could reset the current user search if they removed the filters at some point 🙂 
Globally using reload could lead to greater issue (like variable and data overwrites) comparing to a simple notify :smile:

addons/crm/static/src/views/crm_search_model.js
amdi-odoo: Sorry I'm a bit lost your table renders a bit messy on my side 😄 
But yes this part of the code will only execute following the cache callback

reth-odoo: Ah, well it was meant to be a graph of sorts, with the numbers on the left representing the call order.. no? Well I tried 😆 

Basically my point was we're using `_loaded` as a mutex, but it's not a proper async mutex. So you could imagine that at some point the callback will fire, the load step will have started with the previous context, but we won't have `_loaded = true` yet. In that case we would display stale data.

Hence would propose instead just always reloading if the callback fires, so we don't have to deal with that.

addons/crm/static/src/views/crm_search_model.js
reth-odoo: Oh I see now, indeed I thought SearchModel handled model fetching directly somehow.
Indeed we don't care about the inner context of SearchModel here I suppose so it's pointless to use reload anyways.

addons/crm/static/src/views/crm_search_model.js
reth-odoo: :broom: After digging a bit actually, you don't need to update the context of the action directly at all I think
The model gets its context indirectly from the controller prop `context="this.searchModel.context"`

So we can just override the getter with:
```js
get context() {
     return {...super.context,  default_team_id: this.state.switcherTeamId}
}
```

-------------------------


My main worry was that it seemed like there was a lot of effort put into making shallow copies of the config/context in other places specifically to prevent updating the action context.

amdi-odoo: Hello @reth-odoo  :slightly_smiling_face: Thanks for giving this task a look it helps :smile: 
For the context there's indeed a getter which is defined like so in the base search model:
```js
    get context() {
        if (!this._context) {
            this._context = makeContext([this.globalContext, this._getContext()]);
        }
        return deepCopy(this._context);
    }
```
It uses the `_getContext` method which is also called a bit everywhere manually.
That's why I overwrite the `_getContext` method from my crm search model to add the context I want :slightly_smiling_face: 
=> The context works fine within the search domain, it's the one used when creating a record from a quick create etc...

The only reason why I need to update the **action** context is because, when creating a record by clicking on the "New" button in the list view for example, it opens the form view via the action service. The context used to render the form view is the [action context](https://github.com/odoo/odoo/blob/master/addons/web/static/src/webclient/actions/action_service.js#L714), not the search model context. :smile: So I also need the default team in the action context directly.

And it happens that the `config.context` is a shallow copy of this action context :slightly_smiling_face: 
In the action service, the action context is passed as props to the `View` component and then to the `withSearch` via the [props](https://github.com/odoo/odoo/blob/master/addons/web/static/src/views/view.js#L441). The withSearch then save this action context in the `config` object and [pass it to the search model](https://github.com/odoo/odoo/blob/master/addons/web/static/src/search/with_search/with_search.js#L86-L96).
=> The `config.context` is a shallow copy of the `action.context` which is used to render the form view that's why I need to update it as well

Hope it helps! :smile:

reth-odoo: Oh I see ok, my bad. Thanks for the explanation :)

So because `createRecord` is made directly in the `View` class just patching the context passed to the controller is not sufficient 🤔 

It looks like we could pass a new createRecord prop to the slot in with_search.xml to override the value given in the view, but it doesn't seem there's a clean way to get the original createRecord to augment it (technically we can dissect `this.props.slots` but yeah..)

--------------------------

Ok then indeed I don't see a nice way to do this without patching the view which is more trouble than it's going to be worth.

Can you just add a comment along the lines of "We need to patch the original context passed from the action to affect the context of the New button"

addons/crm/models/crm_lead.py
```diff
@@ -984,16 +984,15 @@ def unlink(self):
...
-        team_ids = self.env.user.crm_team_ids._ids if self.env.context.get('show_user_team_stages') else ()
-        team_ids += (team_id,) if team_id else ()
-        search_domain = ['|', ('id', 'in', stages.ids), ('team_ids', '=', False)]
-        if team_ids:
-            search_domain = ['|', ('id', 'in', stages.ids), '|', ('team_ids', '=', False), ('team_ids', 'in', team_ids)]
+        default_team_id = self.env.context.get('default_team_id')
+        if self.env.context.get('show_team_switcher') and default_team_id:
+            team_ids = [default_team_id]
+        else:
+            team_ids = self.env.user.crm_team_ids._ids if self.env.context.get('show_user_team_stages') else []
```
reth-odoo:
```suggestion
            team_ids = self.env.user.crm_team_ids.ids if self.env.context.get('show_user_team_stages') else []
```

addons/crm/static/src/components/breadcrumbs/crm_breadcrumbs.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+
+    <t t-name="crm.Breadcrumbs">
+        <t t-if="this.hasTeamSwitcher">
+            <TeamSwitcher/>
+            <t t-call="web.Breadcrumb.Actions"/>
+            <t t-call-slot="breadcrumb-status-indicator"/>
+        </t>
```
reth-odoo: So just to make sure, we're removing breadcumbs in this context right?

I suppose it's ok as you wouldn't really land here from anywhere but menu items

addons/crm/static/src/components/team_switcher/team_switcher.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="crm.team_switcher">
+        <Dropdown t-if="this.isActive" menuClass="'crm_team_switcher_dropdown d-flex flex-column'">
+            <button class="o_cp_team_switcher o-dropdown-caret text-truncate fs-4 btn btn-link text-reset fw-normal" t-att-data-tooltip="this.currentLabel">
```
reth-odoo: Does it stand for "currently picked"? Also if it's only used for the tests maybe we can just do `.crm_team_switcher_dropdown > button`?
```suggestion
            <button class="o_team_switcher_label o-dropdown-caret text-truncate fs-4 btn btn-link text-reset fw-normal" t-att-data-tooltip="this.currentLabel">
```

addons/crm/static/tests/crm_team_switcher.test.js
```diff
@@ -0,0 +1,217 @@
...
+import { WebClient } from "@web/webclient/webclient";
+
+class Users extends models.Model {
+    name = fields.Char();
+
+    _records = [
+        { id: 1, name: "Mario" },
+        { id: 2, name: "Luigi" },
+        { id: 3, name: "Link" },
+        { id: 4, name: "Zelda" },
```
reth-odoo, line 25: 🍄

addons/crm/static/src/components/team_switcher/team_switcher.xml
amdi-odoo: It stands for `control panel` like the `o_cp_switch_buttons` or the `o_cp_action_menus` 🙂 
I think it would be great to keep the naming, so that it's aligned with the rest and we can keep an explicit reference to the switcher if needed 🙂

addons/crm/static/src/components/breadcrumbs/crm_breadcrumbs.xml
amdi-odoo: Yes we're removing the breadcrumbs
Adding a new button for the switcher was clogging the interface 😄

reth-odoo:
@aab-odoo if you have a moment to give this one a second look for the approach (I'm not sure if it changed much from the version in May, conceptually). I guess mostly around the searchmodel/controlpanel overrides then

addons/crm/models/crm_team.py
```diff
@@ -785,3 +785,42 @@ def action_primary_channel_button(self):
...
+
+    @api.model
+    def get_team_switcher_teams_data(self):
+        """ Retrieve the teams and their data for the team switcher.
+        Only considering the teams using opportunities.
+        Limited to own teams if the user doesn't have "All Leads" access.
+        """
+        domain = [("use_opportunities", "=", True)]
+        if not self.env.user.has_group("sales_team.group_sale_salesman_all_leads"):
+            domain = Domain.AND([domain, [("id", "in", self.env.user.crm_team_ids.ids)]])
```
tde-banana-odoo: Can't you make a ``&=`` ?

addons/crm/models/crm_team.py
```diff
@@ -785,3 +785,42 @@ def action_primary_channel_button(self):
...
+                'name': team.name,
+                # Domain applied when this team will be selected in the team switcher.
+                # Showing all "crm.lead" records assigned to the team + the unassigned ones which
+                # happens to be part of the team visible stages.
+                'switcher_domain': list(Domain.OR([
+                    [("team_id", "=", team.id)],
+                    Domain.AND([
+                        [("team_id", "=", False)],
+                        [("stage_id", "in", stage_ids_by_team.get(team.id, []) + shared_stage_ids)],
+                    ]),
```
tde-banana-odoo: Can't we write something with less lists like
```suggestion
                'switcher_domain': list(
                    # assigned to team
                    Domain("team_id", "=", team.id) | \
                    # unassigned in team visible stages
                    (
                        Domain("team_id", "=", False) &
                        Domain("stage_id", "in", stage_ids_by_team.get(team.id, []) + shared_stage_ids),
                    )
                ),
```
Seems easier to read to me

addons/crm/models/crm_team.py
tde-banana-odoo: Can't you make a &= ?
```suggestion
            domain &= Domain("id", "in", self.env.user.crm_team_ids.ids)
```

addons/crm/models/crm_team.py
```diff
@@ -785,3 +785,42 @@ def action_primary_channel_button(self):
...
+    # UTILS
+    # ------------------------------------------------------------
+
+    @api.model
+    def get_team_switcher_teams_data(self):
+        """ Retrieve the teams and their data for the team switcher.
+        Only considering the teams using opportunities.
+        Limited to own teams if the user doesn't have "All Leads" access.
+        """
+        domain = [("use_opportunities", "=", True)]
```
tde-banana-odoo:
```suggestion
        domain = Domain("use_opportunities", "=", True)
```

addons/crm/models/crm_stage.py
```diff
@@ -35,6 +35,11 @@ class CrmStage(models.Model):
     # This field for interface only
     team_count = fields.Integer('team_count', compute='_compute_team_count')
 
+    @api.model
+    def _get_visible_stages_domain(self, team_ids):
+        """ Returns the "crm.stage" domain to know which stages are visible for the given team ids. """
+        return ["|", ("team_ids", "in", team_ids), ("team_ids", "=", False)]
```
tde-banana-odoo: Could return a 'Domain' directly
```suggestion
        return Domain("team_ids", "in", team_ids) | Domain("team_ids", "=", False)
```

addons/crm/models/ir_http.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models
+
+
+class IrHttp(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    def session_info(self):
+        result = super().session_info()
+        if self.env.user._is_internal():
+            result['is_membership_multi'] = self.env['ir.config_parameter'].sudo().get_bool('sales_team.membership_multi')
```
tde-banana-odoo: Can't we keep the "sales_team" namespacing to avoid any conflict ?
```suggestion
            result['sales_team_membership_multi'] = self.env['ir.config_parameter'].sudo().get_bool('sales_team.membership_multi')
```

addons/mass_mailing_crm/models/mailing_mailing.py
```diff
@@ -25,7 +25,12 @@ def _compute_crm_lead_count(self):
 
     @api.model
     def action_create_mailing_template_with_leads(self):
-        reply_to = self.env.user.sale_team_id.alias_email or False
+        team_id = self.env.context.get('default_team_id', False)
```
tde-banana-odoo: As this is called from lead generation brol which, I hope, is linked to crm rights, shouldn't we call CrmTeam._get_default_team_id() ?

addons/crm/models/crm_team.py
```diff
@@ -785,3 +785,42 @@ def action_primary_channel_button(self):
         if self.use_opportunities:
             return self.action_open_opportunities()
         return super().action_primary_channel_button()
+
+    # ------------------------------------------------------------
+    # UTILS
+    # ------------------------------------------------------------
+
+    @api.model
+    def get_team_switcher_teams_data(self):
```
tde-banana-odoo, line 794: Shouldn't ``_get_default_team_id`` method on CrmTeam be impacted by team switcher ?

tde-banana-odoo (commented):
Some comments on py

addons/mass_mailing_crm/models/mailing_mailing.py
reth-odoo: Note that unlike `sale_team_id`, `_get_default_team_id` doesn't care what your oldest membership is so that's a change

But typically we use that one regardless so it shouldn't be too confusing

addons/crm/models/crm_team.py
reth-odoo: It already is indirectly since the team switcher works by setting `default_team_id` (which is what `_get_default_team_id` uses if it's valid and using default parameters), and won't let you pick a team that would be invalid in `_get_default_team_id`

Not sure if you'd expect more than that

tde-banana-odoo: Ok then, seems correct :)

tde-banana-odoo:
@robodoo r+

reth-odoo:
rebased

tde-banana-odoo:
@robodoo r+

reth-odoo:
And updated session_info getter query-count. It's unclear what made it increase more than previously. Perhaps the user group is no longer in cache when the crm override is called. Both queries are in long caches anyways so it only impacts cold starts.

tde-banana-odoo:
@robodoo priority

Reason: startmaster / FP demo

tde-banana-odoo:
@robodoo r+

PR #248026, opened 2026-02-10
[FIX] sale_timesheet: allow salesperson to open SO

abla001:
to reproduce:
=============
- create a user with only user access to timesheet and no access to project
- create and confirm a sale order with a timesheet-delivered service
product assigned to that salesperson
- record some hours on that SO with another user
- open the sale order as that salesperson

-> AccessError

problem:
========
`_compute_show_hours_recorded_button` reads `timesheet_count` and
`project_count` on the sale order, which internally query
`account.analytic.line` and `project.project` records. a salesperson
without all timesheet or project groups lacks read access to those
models, causing an AccessError during the compute.

solution:
=========
use `sudo()` since the result is only used to determine
button visibility and does not expose sensitive data to the user.

opw-5893324

### DIFF
--- a/addons/sale_timesheet/tests/test_sale_service.py
+++ b/addons/sale_timesheet/tests/test_sale_service.py
@@ -1,5 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+from odoo.addons.mail.tests.common import mail_new_test_user
 from odoo.addons.sale_timesheet.tests.common import TestCommonSaleTimesheet
 from odoo.exceptions import UserError, ValidationError
 from odoo.tests import tagged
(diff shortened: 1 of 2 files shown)


abla001 (commented):
@odoo/rd-services can someone take a look at this please ?

addons/sale_timesheet/tests/test_sale_service.py
```diff
@@ -812,6 +813,32 @@ def test_compute_show_timesheet_button(self):
...
+            'order_id': sale_order.id,
+            'product_id': self.product_delivery_timesheet2.id,
+            'product_uom_qty': 5,
+        })
+        sale_order.action_confirm()
+        # Computing show_hours_recorded_button as the restricted salesperson
+        # should not raise an AccessError when reading timesheet_count and
+        # project_count internally.
+        sale_order_as_salesperson = sale_order.with_user(salesperson)
+        sale_order_as_salesperson._compute_show_hours_recorded_button()
```
xavierbol: we could take the opportunity to check if `show_hours_recorded_button` contains the expected value. 😇

xavierbol (commented):

robodoo delegate+

you can r+ once it is done

abla001 (commented):
@xavierbol  Great catch suggesting to check the value of `show_hours_recorded_button` ! While adding the check, I ran into an access error and discovered that a user without the User Timesheet group shouldn't be able to read that field in the first place. Since the view already has a group attribute ensuring this, that setup avoids the AccessError entirely.
The correct steps to reproduce the error are: a salesperson with the User Timesheet group tries to open their SO that has hours recorded by another user, that's what triggers the access error.
With the `sudo` fix, the salesperson will now be able to see the recorded hours smart button, but clicking it will lead to an empty list.
Thanks for the delegate, that said, I'd prefer to wait for a second review from your side before merging. 🫶

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #248034, opened 2026-02-10
[IMP] l10n_tr_nilvera_edispatch: send XML with API 

alialfie:
Adds automatic sending to Nilvera.

task-5134297

### DIFF
--- a/addons/l10n_tr_nilvera_edispatch/models/res_partner.py
+++ b/addons/l10n_tr_nilvera_edispatch/models/res_partner.py
@@ -1,4 +1,4 @@
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 
 class ResPartner(models.Model):
@@ -9,6 +9,21 @@ class ResPartner(models.Model):
         help="The postal code of the customs office used to ship to the destination country.",
         size=5,
     )
+    l10n_tr_nilvera_edispatch_alias_id = fields.Many2one(
+        comodel_name='l10n_tr.nilvera.alias',
+        string="eDispatch Alias",
+        compute='_compute_l10n_tr_nilvera_edispatch_alias_id',
+        domain="[('partner_id', '=', id)]",
+        copy=False,
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends('l10n_tr_nilvera_customer_alias_ids', 'l10n_tr_nilvera_customer_status')
+    def _compute_l10n_tr_nilvera_edispatch_alias_id(self):
+        for record in self:
+            if not record.l10n_tr_nilvera_edispatch_alias_id:
+                record.l10n_tr_nilvera_edispatch_alias_id = record.l10n_tr_nilvera_customer_alias_ids[:1]
 
     def _l10n_tr_nilvera_validate_partner_details(self, tax_office_required=False):
         error_messages = {}
--- a/addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
+++ b/addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
@@ -327,7 +327,7 @@ class StockPicking(models.Model):
                             indicates a client error (4xx), or if a server error occurs (500).
         :return: None
         """
-        with _get_nilvera_client(self.env.company) as client:
+        with _get_nilvera_client(self.env._, self.company) as client:
             response = client.request(
                 "POST",
                 endpoint='/edespatch/Send/Xml',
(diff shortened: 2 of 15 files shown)


addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
```diff
@@ -8,11 +10,19 @@
...
 
 class StockPicking(models.Model):
     _inherit = 'stock.picking'
 
+    l10n_tr_nilvera_uuid = fields.Char(
+        string="Nilvera Document UUID",
+        copy=False,
+        readonly=True,
+        default=lambda self: str(uuid.uuid4()),
+        help="Universally unique identifier of the stock operation",
```
jbw-odoo:
```suggestion
```
If the field does not appear in a view, help has no use. Add a comment instead if needed

addons/l10n_tr_nilvera_edispatch/models/res_partner.py
```diff
@@ -9,6 +9,21 @@ class ResPartner(models.Model):
         help="The postal code of the customs office used to ship to the destination country.",
         size=5,
     )
+    l10n_tr_nilvera_dispatch_alias_id = fields.Many2one(
```
jbw-odoo:
```suggestion
    l10n_tr_nilvera_edispatch_alias_id = fields.Many2one(
```
Why not go with "edispatch" (here and elsewhere) ? It's clearer imo.

addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
```diff
@@ -72,13 +82,49 @@ class StockPicking(models.Model):
...
+    l10n_tr_nilvera_edispatch_xml_id = fields.Many2one(
+        "ir.attachment",
+        string="Nilvera E-Despatch XML",
+        compute=lambda self: self._compute_linked_attachment_id(
+            "l10n_tr_nilvera_edispatch_xml_id", "l10n_tr_nilvera_edispatch_xml_file"
+        ),
+        depends=["l10n_tr_nilvera_edispatch_xml_file"],
+    )
+
+    def _compute_linked_attachment_id(self, attachment_field, binary_field):
```
jbw-odoo: I get that this is how it's done in other places. But why not a simple compute method? This is only used in the l10n_tr_nilvera_edispatch_xml_id compute. And the method name is not namespaced. I  see that l10n_uy_edi also extends stockpicking and define a method with the same name, with no check on country or super()

addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
```diff
@@ -263,29 +307,122 @@ def _l10n_tr_generate_edispatch_xml(self):
...
+            },
+        )
+
+    def _l10n_tr_nilvera_get_submitted_document_status(self):
+        for company, stock_pickings in self.grouped("company_id").items():
+            with _get_nilvera_client(company) as client:
+                for stock_picking in stock_pickings:
+                    response = client.request(
+                        "GET",
+                         endpoint=f"/edespatch/Sale/{stock_picking.l10n_tr_nilvera_uuid}/Status",
```
jbw-odoo:
```suggestion
                        endpoint=f"/edespatch/Sale/{stock_picking.l10n_tr_nilvera_uuid}/Status",
```

jbw-odoo (commented):
Hello, started reviewing before realizing the task what not tested by the PO. Here are my comments so far.

addons/l10n_tr_nilvera_edispatch/models/stock_picking.py
```diff
@@ -263,29 +304,122 @@ def _l10n_tr_generate_edispatch_xml(self):
...
+        if self.l10n_tr_nilvera_edispatch_warnings:
+            raise UserError(_("Cannot send the XML when there are warnings."))
+        self._l10n_tr_generate_edispatch_xml()
+        xml_file = BytesIO(self.l10n_tr_nilvera_edispatch_xml_id.raw or b'')
+        xml_file.name = self.l10n_tr_nilvera_edispatch_xml_id.name or ''
+        self._l10n_tr_nilvera_submit_document(xml_file)
+
     def action_generate_l10n_tr_edispatch_xml(self, is_list=False):
-        errors = []
+        invalid_picking_names = []
```
alialfie, line 408: PS: this change is irrelevant to the scope of the PR, but didn't want to make a whole commit just to fix this. If you mind it I can remove it though

jbw-odoo: No no, it's ok.

jbw-odoo (approved):
Looks good, thanks !
robodoo r+

jbw-odoo:
robodoo r+

PR #248036, opened 2026-02-10
[IMP] base: remove permissions to create/unlink languages

kmagusiak:
Languages are technical and their codes must be recognized by babel. Otherwise, we will get stack traces for unknown languages.

Avoiding things like `UnknownLocaleError` (in babel).

---


rco-odoo (approved):
@robodoo r+

PR #248038, opened 2026-02-10
[IMP] web: core: expression editor supports contains for char fields

kebeclibre:
Before this commit the `contains` operator (alias `ilike`) was not supported
by the expression editor, so the use of char fields in a field's invisible
expression was limited.

After this commit, we implement limited support for this.

task-5189106

### DIFF
--- a/addons/web/static/src/core/tree_editor/condition_tree.js
+++ b/addons/web/static/src/core/tree_editor/condition_tree.js
@@ -1,3 +1,4 @@
+import { deepEqual } from "@web/core/utils/objects";
 import { Domain } from "@web/core/domain";
 import { formatAST, parseExpr } from "@web/core/py_js/py";
 import { toPyValue } from "@web/core/py_js/py_utils";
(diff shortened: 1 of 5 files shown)


addons/web/static/src/core/py_js/py.js
```diff
@@ -40,7 +40,9 @@ export function evaluateExpr(expr, context = {}) {
     try {
         return evaluate(ast, context);
     } catch (error) {
-        throw new EvalError(`Can not evaluate python expression: (${expr})\nError: ${error.message}`);
+        throw new EvalError(
```
Polymorphe57: remove diff in this file

addons/web/static/src/core/tree_editor/construct_expression_from_tree.js
```diff
@@ -71,6 +71,12 @@ function _constructExpressionFromTree(tree, options, isRoot = false) {
     if (path instanceof Expression && operator === "=" && value === 1) {
         return path.toString();
     }
+    if (
```
Polymorphe57: add comment
// we can assume that negate = false here: "ilike" and "not ilike" have negation defined

addons/web/static/src/core/tree_editor/condition_tree.js
```diff
@@ -346,3 +347,48 @@ export function rewriteNConsecutiveChildren(transformation, N = 2) {
...
+        const path = rightExpr.value;
+        const dummy = parseExpr(
+            expressionContainsString.toString(path, leftString.value, ast.op === "not in")
+        );
+        const areEqual = deepEqual(ast, dummy);
+        if (areEqual) {
+            return {
+                path,
+                operator: "ilike",
+                negate: ast.op === "not in",
```
Polymorphe57: why not have negate false and operator be ilike or not ilike?

Polymorphe57: you get not contains in the select for operators -> not stable

Polymorphe57:
@robodoo r+

PR #248042, opened 2026-02-10
[FIX] website: clean up event listeners on iframe reload

sben-odoo:
Steps to reproduce:
===================

1. In Website > Shop, open any product.
2. Click "Edit" on the product.
3. Make a change (for example, toggle "Zoom on click") 2 times.

-> In the second time a traceback will appear `(Safari, Epiphany)`.

Cause:
======

The website builder uses two contexts:
- The **iframe** runs services (e.g. websiteEditService)
- The **parent window** runs editor plugins (e.g. EditInteractionPlugin)

They communicate via custom events on window.parent.document. The service (in the iframe) registers a listener for
"edit_interaction_plugin_loaded" on the parent document, and responds with a "transfer_website_edit_service" event carrying its service object.

When the iframe reloads after a save, the old listener was never removed from window.parent.document. The new iframe added a second listener. This meant two listeners existed on the parent document:

  1. Old listener → responds with OLD websiteEditService (stale)
  2. New listener → responds with NEW websiteEditService (correct)

Both fire when the plugin dispatches "edit_interaction_plugin_loaded". But the plugin only accepts the first response ({ once: true }), and since listeners fire in registration order, it always received the OLD stale service.

With the stale service:
- installPatches() early-returned (patches.length > 0 from session 1)
- destroy() called uninstallPatches() on the old service, removing stopInteractionByName from the old prototype
- stopInteraction() then tried to call the now-deleted method → crash: TypeError: publicInteractions.stopInteractionByName is not a function

Solution:
=========

Extract the anonymous event handlers into named functions and register a "beforeunload" listener on the iframe's window that removes them from window.parent.document. This ensures that when the iframe reloads, the old listeners are cleaned up before the new iframe registers its own, so the plugin always receives the current service.

opw-5864798

---

### DIFF
--- a/addons/website/static/src/core/website_edit_service.js
+++ b/addons/website/static/src/core/website_edit_service.js
@@ -293,12 +293,12 @@ export const websiteEditService = {
             callShared,
         };
 
-        window.parent.document.addEventListener("edit_page", (ev) => {
+        const handleEditPage = (ev) => {
             stop(ev.detail.iframeDocument);
-        });
+        };
 
         // Transfer the iframe website_edit service to the EditInteractionPlugin
-        window.parent.document.addEventListener("edit_interaction_plugin_loaded", (ev) => {
+        const handlePluginLoaded = (ev) => {
             ev.currentTarget.dispatchEvent(
                 new CustomEvent("transfer_website_edit_service", {
                     detail: {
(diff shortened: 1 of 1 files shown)


nle-odoo (approved):
I don't know what's the cleanest way to fix this

- this
- unregister on next transfer_website_edit_service event but then we still need to keep references to the method)
- gemini tells me there is AbortController (example https://gist.github.com/nle-odoo/2f9b23779f2179be334dfdf7e5ee0336 but I had never heard of it and it ends up being a little)

but current fix seems good to me

side note: since this issue causes "step into" to not work in debugger (that we think happen since we step into a dead iframe code) this is infuriating to debug and test in safari / epiphany

PR #248044, opened 2026-02-10
[FIX] html_editor: allow cropping images not linked to any attachment

julien-banken:
Currently, when a user edits an image using the image cropper, the script transforms the image, converts it to base 64, and stores it in the image's `src` attribute.

The script then adds the `o_modified_image_to_save` class. When the document is saved, the script retrieves all images with this class, creates a copy of the original attachment containing the modified image, and replaces the image src with the URL of the new attachment.

The problem occurs when the image is not linked to any attachment (for example, when the image uses an external URL). In this case, the method crashes and the document cannot be saved.

Steps to reproduce:

1. Open Mass Mailing
2. Insert a snippet containing an image (with an external url)
3. Click on the image
4. Click on the "Style" tab
5. Click on the crop icon
6. Resize the image
7. Click on "Apply" to save the transformation
8. Save the record

=> An error appears in the console when calling:
   `/html_editor/modify_image/undefined`

When the image does not have a `data-original-id` attribute, use the CSS class `o_b64_image_to_save` instead of `o_modified_image_to_save`. This ensures that the script automatically creates an attachment for the modified image when no original attachment exists.

Task-6128841

---

### DIFF
--- a/addons/website/static/tests/builder/images.test.js
+++ b/addons/website/static/tests/builder/images.test.js
@@ -62,7 +62,7 @@ test("Double click on image and replace it", async () => {
     await waitSidebarUpdated();
     await waitForNone(".o_select_media_dialog");
     expect(".o_select_media_dialog").toHaveCount(0);
-    expect(":iframe img").toHaveClass("o_modified_image_to_save");
+    expect(":iframe img").toHaveClass("o_b64_image_to_save");
     expect(".options-container[data-container-title='Image']").toHaveCount(1);
 });
 
--- a/addons/html_editor/static/src/main/media/image_crop.js
+++ b/addons/html_editor/static/src/main/media/image_crop.js
@@ -72,10 +72,7 @@ export class ImageCrop extends Component {
         );
 
         onMounted(() => {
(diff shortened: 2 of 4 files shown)


addons/website/static/tests/builder/images.test.js
```diff
@@ -62,7 +62,7 @@ test("Double click on image and replace it", async () => {
     await waitSidebarUpdated();
     await waitForNone(".o_select_media_dialog");
     expect(".o_select_media_dialog").toHaveCount(0);
-    expect(":iframe img").toHaveClass("o_modified_image_to_save");
+    expect(":iframe img").toHaveClass("o_b64_image_to_save");
```
julien-banken, line 65: For the test involving the media dialog, I don't understand why it doesn't use the attachment url link. Instead, it re-encodes the selected image in base 64 and injects that in the image's src...

The issue is that it has the effect of duplicating the attachment upon saving. I can reproduce the issue on master. Maybe, it's to ensure that people can have access to the attachment, so it's normal that we duplicate it? 🤔

PR #241541, opened 2025-12-29
[FIX] website_sale: align carousel arrow direction in RTL

sben-odoo:
Steps to reproduce:
===================
- Install an RTL language (e.g., Arabic) on the website.
- Go to website & Edit mode
- Drop product snippet `s_dynamic_snippet_products`.

->When the website is viewed in an RTL language, the navigation arrows
in the snippet are displayed in the wrong direction.

Cause:
======
A previous commit [1] applied a specific `transform` to arrows to adjust their visual style,
By defining a new transformation for the snippet, the default generic transform for RTL language was overwritten.
https://github.com/odoo/odoo/blob/e0a5e68e861363378ecd184711bc9356fe0376ff/addons/web/static/lib/odoo_ui_icons/style.css#L105 Consequently, the arrows lost their RTL-specific orientation.

Solution:
=========
The CSS has been updated to explicitly handle the RTL context.

[1]: https://github.com/odoo/odoo/commit/d9ea6d1ebabf64dac434e0d2a7c2535bbfff5c71

opw-5440273

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248046, opened 2026-02-10
[FIX] website_sale: fix products carousel arrow size

YoussefM890:
**Problem:**
When a "Products" block on the website is set to the full-width (extended)
content option, the previous/next carousel navigation arrows render at an
abnormally large size instead of their intended compact appearance.

**Steps to reproduce:**
1. Go to Website > Edit a page
2. Add a "Products" block
3. Set the content width to the rightmost (full-width) option
4. Observe the carousel navigation arrows

**Current behavior:**
The carousel navigation arrows appear oversized and stretched.

**Expected behavior:**
The carousel navigation arrows should display at their normal size.

**Cause of the issue:**
The carousel control buttons have a `transform: scaleX(0.2)` to counteract
the inner chevron icon's `transform: scaleX(5)`, keeping
the visible arrow at its intended size. In full-width mode, a positional
`transform: translateX(...)` is applied to the same parent element. Since
`transform` is a single CSS property, this completely overrides the
`scaleX(0.2)`, leaving the chevron's `scaleX(5)` unopposed and causing the
arrows to render at 500% of their expected width.

**Fix:**
Both the positional offset and the scale compensation must coexist in the same
`transform` declaration on the parent. Extracting the scale value into a CSS
variable allows the full-width rule to reference it directly, combining
`translateX()` and `scaleX()` in one `transform` without duplicating the magic
value. This also preserves RTL compatibility, as `transform: translateX()` is
correctly mirrored by rtlcss, unlike the standalone `translate` property.

opw-5917549

### DIFF
--- a/addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss
+++ b/addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss
@@ -11,6 +11,7 @@
         .carousel-control-prev {
             aspect-ratio: var(--o-wsale-card-thumb-aspect-ratio, 1/1);
             width: var(--o-wsale-dynamic-snippet-col-size) !important;
+            --o-wsale-dynamic-snippet-control-scale: 0.2;
             transform: scaleX(.2);
             bottom: auto;
             isolation: isolate;
@@ -77,14 +78,14 @@
         }
     }
 }
-.s_dynamic_snippet_products .container-fluid .s_dynamic_snippet_content {
+.s_dynamic_snippet_products .container-fluid .s_dynamic_snippet_title_aside + .s_dynamic_snippet_content {
     @include media-breakpoint-up(md) {
         // Prevented horizontal overflow of control buttons.
         .carousel-control-prev {
-            transform: translateX(50%); 
+            transform: translateX(50%) scaleX(var(--o-wsale-dynamic-snippet-control-scale, 1));
         }
         .carousel-control-next {
-            transform: translateX(-50%); 
+            transform: translateX(-50%) scaleX(var(--o-wsale-dynamic-snippet-control-scale, 1));
         }
     }
 }


addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss
```diff
@@ -81,10 +81,10 @@
     @include media-breakpoint-up(md) {
         // Prevented horizontal overflow of control buttons.
         .carousel-control-prev {
-            transform: translateX(50%); 
+            translate: 50% 0%;
```
sben-odoo: Thanks for the fix. However, it will behave unexpectedly when switching to Arabic (RTL languages).

nle-odoo (approved):
> The inner chevron icon has a `transform: scaleX(5) rotate(...)`. The parent  .carousel-control-prev/next` was using
`transform: translateX(...)` for positioning. Since both the parent's
positioning and the child's scale relied on the `transform` property, the
parent's transform overrode the intended scale compensation, leaving the
chevron's `scaleX(5)` unopposed and causing the arrows to appear massively
stretched in full-width mode.

The issue is rather than on:

https://github.com/odoo/odoo/blob/00b042c264c25d8e2ee4b019915a3f76fc04dc39/addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss#L10-L14

The scaleX is set 0.2, so it will be 100% with the scaleX(5) on the icon.

The `transform: translateX(...)` is overriding the scaleX(0.2) so we have a horizontal width of 500% of expectation (in the case where the icon is scaled 5 times).

For the fix it seems good to me to avoid the conflict 👍 

Edit: as a side note, the module in the title should be the modified module, so "website_sale" in this case.

addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss
nle-odoo: good catch, it seems translate is not rtlcss'ed while "transform:translateX" is rtlcss'ed (rtlcss is the program that is mirroring all CSS statement for right-to-left language)

I guess then the cleanest solution would be to have a variable in:

https://github.com/odoo/odoo/blob/7c7c6663e28974834d1569b27605f6ce400c7b16/addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss#L10-L14

like:

```css
--o-wsale-dynamic-snippet-control-scale: 0.2;
transform: scaleX(var(--o-wsale-dynamic-snippet-control-scale));
````
then in the part you were modifying, do:

``` css
.carousel-control-prev {
    transform: translateX(50%) scaleX(var(--o-wsale-dynamic-snippet-control-scale, 1));
}
.carousel-control-next {
    transform: translateX(-50%) scaleX(var(--o-wsale-dynamic-snippet-control-scale, 1));
}
```

YoussefM890: How to reproduce this issue because it looks fine to me ?  @nle-odoo @sben-odoo 
image

nle-odoo: there is a backend language (that you have in Arabic as shown in the editor panel) and frontend (=website) language, to change the language of frontend you can just use the button next to "Mitchell Admin":

image

if Arabic doesn't appear in available language, you need to change the website configuration to add the language as available.

YoussefM890: @nle-odoo I reproduced the issue, Thanks for the help 🙏

nle-odoo (approved):

livaios:
Hello @YoussefM890 
Thanks for the fix, it LGTM 👍 
However I noticed another bug when I switched from ltr to rtl, the arrows are reversed.
image

This is caused by this PR https://github.com/odoo/odoo/pull/241541 and I guess it was fixed elsewhere and the fix was no longer needed, so can you just revert this fix in your PR?

just noting that for both bugs, they only occurs for 19.0 and wont be needing a fw port.

livaios:
Hello @Brieuc-brd @Jean-boje 

Can you check if this is the intended placement of arrows when the snippet is full width?
Screenshot 2026-02-13 at 09 06 35

livaios:
Hello @YoussefM890 
Indeed I can't reproduce it as well maybe my environment was polluted somehow.
Can you just implement the changes requested by Brieuc here

> Hey @livaios @YoussefM890 👋 This placement is not intended. I’ve taken a look and this should fix the issue :
> 
> ```diff
> // addons/website_sale/static/src/snippets/s_dynamic_snippet_products/000.scss‎  -  line 81
> 
> - .s_dynamic_snippet_products .container-fluid .s_dynamic_snippet_content {
> + .s_dynamic_snippet_products .container-fluid .s_dynamic_snippet_title_aside + .s_dynamic_snippet_content {
> ```
> 
> Could you please double-check by testing the following features with this change ? : “Content width”, “Scrolling Mode”, and “Section title” ?
> 
> Thanks !

livaios (approved):
Hello @YoussefM890 
LGTM 👍 
Can you just fix runbot? 
Not sure what's the issue but maybe a rebase will be enough 🤷‍♀️

YoussefM890:
@livaios Rebase done 👍

PR #248045, opened 2026-02-10
[PERF] hr_holidays: batch first working interval

kmagusiak:
Given a recordset, we do O(n) queries. Rewrite
`_get_first_working_interval` to use batching of the underlying functions.

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248047, opened 2026-02-10
[PERF] hr: presence state refactor

kmagusiak:
Use batch computation for `_compute_dates`. Getting all values of `hr_presence_state` need that field.
Implement `hr_presence_state` using *compute_sql* to have one implementation and add a limit for big databases because "Ugly hack".

Related task that should be batched: #248045

---

### DIFF
--- a/addons/hr/models/hr_employee.py
+++ b/addons/hr/models/hr_employee.py
@@ -109,7 +109,7 @@ class HrEmployee(models.Model):
         ('present', 'Present'),
         ('absent', 'Absent'),
         ('archive', 'Archived'),
-        ('out_of_working_hour', 'Off-Hours')], compute='_compute_presence_state', search='_search_presence_state', default='out_of_working_hour')
+        ('out_of_working_hour', 'Off-Hours')], compute='_compute_presence_state', compute_sql='_compute_sql_presence_state', compute_sudo=False, default='out_of_working_hour')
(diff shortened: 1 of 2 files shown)


kmagusiak:
@robodoo rebase-ff

addons/hr/models/hr_employee.py
```diff
@@ -986,27 +986,16 @@ def _compute_presence_state(self):
                 state = 'archive'
             employee.hr_presence_state = state
 
-    def _search_presence_state(self, operator, value):
```
Einar-eian, line 989: I wonder if removing this method in stable could cause issues for someone? As in someone having their own external module that calls this method. But then again it has warnings etc. so maybe the chances of someone building something using this method is pretty low.

Einar-eian (commented):
Thank you for your work 😄 I have one question but otherwise it's looking good to me!

addons/hr/models/hr_employee.py
kmagusiak: You should not call this method directly, it is used for searching only. Since this is an intermediary saas version, no modules outside of Odoo are supported to use it.

If you prefer, I can bring it back.

Einar-eian: Yes in that case we should just remove it and everything looks good to me! :smile:

rco-odoo (approved):
@robodoo r+

PR #248056, opened 2026-02-10
[FIX] mrp: correctly handle unbuild more than manufactured

MAGREMENT:
# How to reproduce
- Create a BOM for a product
- Create a MO for a set quantity (Exemple: 1) for that product
- Unbuild that MO and ask to unbuild more than what was manufactured (Exemple: 3)
- Confirm the unbuild
- Go to the stock moves of that unbuild via the smart button

# The problem
3 stock move lines are created, 2 in the 'Done' state and 1 in the 'Available' state. This last move line is stuck and cannot be validated

# Why
The cause of this issue is due to a discrepency between the quantity set for the move lines and the quantity set for their respective moves.

When creating the move lines for the produce move, we use the original move of the MO (this is done to keep Lots consistent). If the quantity of product to unbuild is more than the quantity of product built by the MO, the quantity of the move lines will be less than expected.

This will then create a backorder when the produce move is set to done. This backorder will then be unvalidatable because the unbuild it is linked to will be set to 'Done'.

opw-5915981

opw-5449109

---

### DIFF
--- a/addons/mrp/tests/test_unbuild.py
+++ b/addons/mrp/tests/test_unbuild.py
@@ -1122,3 +1122,45 @@ class TestUnbuild(TestMrpCommon):
                     {'product_id': self.product_2.id, 'quantity': 24, 'state': 'done'},  # Wood
                     {'product_id': self.product_1.id, 'quantity': 48, 'state': 'done'},  # Courage
                 ])
+
+    def test_unbuild_more_than_manufactured(self):
+        mo, bom, p_final, p1, p2 = self.generate_mo()
+
+        self.env['stock.quant']._update_available_quantity(p1, self.stock_location, 100)
+        self.env['stock.quant']._update_available_quantity(p2, self.stock_location, 5)
+        mo.action_assign()
+
+        mo_form = Form(mo)
+        mo_form.qty_producing = 5.0
+        mo = mo_form.save()
+        mo.button_mark_done()
+        self.assertEqual(mo.state, 'done', "Production order should be in done state.")
+
+        # Check quantity in stock before unbuild.
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p_final, self.stock_location), 5, 'You should have the 5 final product in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p1, self.stock_location), 80, 'You should have 80 products in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p2, self.stock_location), 0, 'You should have consumed all the 5 product in stock')
+
+        # ---------------------------------------------------
+        #       unbuild
+        # ---------------------------------------------------
+
+        x = Form(self.env['mrp.unbuild'])
+        x.product_id = p_final
+        x.bom_id = bom
+        unbuild = x.save()
+
+        unbuild.mo_id = mo.id
+        unbuild.product_qty = 7
+        unbuild.action_unbuild()
+
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p_final, self.stock_location, allow_negative=True), -2, 'You should have 2 less final product in stock than what you started with')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p1, self.stock_location), 108, 'You should have gained back 8 more products than what you started with')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p2, self.stock_location), 7, 'You should have gained back 2 more products than what you started with')
+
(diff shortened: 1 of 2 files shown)


addons/mrp/tests/test_unbuild.py
```diff
@@ -1122,3 +1122,46 @@ def test_unbuild_non_storable_product(self):
...
+        x.bom_id = bom
+        unbuild = x.save()
+
+        unbuild.mo_id = mo.id
+        unbuild.product_qty = 7
+        unbuild.action_unbuild()
+
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p_final, self.stock_location, allow_negative=True), -2, 'You should have consumed 3 final product in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p1, self.stock_location), 108, 'You should have 80 products in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p2, self.stock_location), 7, 'You should have consumed all the 5 product in stock')
```
svs-odoo: Update the strings

addons/mrp/tests/test_unbuild.py
```diff
@@ -1122,3 +1122,46 @@ def test_unbuild_non_storable_product(self):
...
+
+        moves = self.env['stock.move.line'].search([
+            ('reference', '=', unbuild.name)
+        ])
+
+        for move in moves:
+            self.assertEqual(move.state, 'done')
```
svs-odoo: You can find the unbuild's move lines through the field `produce_line_ids`.
You can also use `assertRecordValues` to check multiple records' values at once.
Finally, if you handle move lines, don't name the variable `moves` since they aren't moves but move lines (it's quite confusing otherwise)
```suggestion
        move_lines = unbuild.produce_line_ids.move_line_ids.sorted(lambda ml: ml.product_id.id)
        self.assertRecordValues(move_lines, [
            {'reference': unbuild.name, 'quantity': 7, 'product_id': p_final.id, 'state': 'done'},
            {'reference': unbuild.name, 'quantity': 28, 'product_id': p1.id, 'state': 'done'},
            {'reference': unbuild.name, 'quantity': 7, 'product_id': p2.id, 'state': 'done'},
        ])
```

addons/mrp/tests/test_unbuild.py
```diff
@@ -1122,3 +1122,46 @@ def test_unbuild_non_storable_product(self):
                     {'product_id': self.product_2.id, 'quantity': 24, 'state': 'done'},  # Wood
                     {'product_id': self.product_1.id, 'quantity': 48, 'state': 'done'},  # Courage
                 ])
+
+    def test_unbuild_more_than_manufactured(self):
+        mo, bom, p_final, p1, p2 = self.generate_mo()
+        self.assertEqual(len(mo), 1, 'MO should have been created')
```
svs-odoo: Not sure to understand the purpose of this assert ?
```suggestion
```

svs-odoo (approved):
Note that by doing this change, it won't be possible to unbuild more than available in stock if some of the components are tracked by serial numbers, but since previously the exceeding quantity's move was not done, the fixed version is better anyway.

addons/mrp/tests/test_unbuild.py
```diff
@@ -1122,3 +1122,45 @@ def test_unbuild_non_storable_product(self):
...
+        x.bom_id = bom
+        unbuild = x.save()
+
+        unbuild.mo_id = mo.id
+        unbuild.product_qty = 7
+        unbuild.action_unbuild()
+
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p_final, self.stock_location, allow_negative=True), -2, 'You should have consumed 2 more than the 3 final product in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p1, self.stock_location), 108, 'You should have 108 products in stock')
+        self.assertEqual(self.env['stock.quant']._get_available_quantity(p2, self.stock_location), 7, 'You should have consumed 7 products')
```
svs-odoo: String is still wrong, it's a unbuild order, not a MO, it's wrong to say component was consumed when they were added in stock

MAGREMENT: I copied what was done here https://github.com/odoo-dev/odoo/blob/81d38784f72d232c032c187b6eb9544e6d6f2bdd/addons/mrp/tests/test_unbuild.py#L54 but, sure, I'll change it

svs-odoo: @MAGREMENT Copying an error doesn't make it right, you can also see that that string was wrongly copy-pasted at multiple places.
A string in the assert must refer to what is tested, your current message is much better

svs-odoo:
robodoo r+

PR #248061, opened 2026-02-10
[IMP] mail, *: replace `-webkit-line-clamp` with `.o_line_clamp`

chgo-odoo:
*: account, gamification, html_builder, html_editor, point_of_sale,
pos_restaurant, pos_self_order, website_crm_partner_assign, website_customer,
website_hr_recruitment, website_partnership, website_profile, website_sale_slides, website_slides

- requires https://github.com/odoo/enterprise/pull/107027
-------------

In Commit[^1], we replaced the `o-line-clamp` mixin in favor of a class. The main goal was to avoid using multiple `@include` across files.

In Commit[^2], we faced a use-case for this utility class in the back-end and decided to move the definition of the class in this bundle, which makes it available for both back and front-end.

To clamp a certain amount of line, you can now apply the `.o_line_clamp` class to your element, and assign a value to the `--lines-clamp` CSS variable, which will be used to clamp that number of line. If you don't define this variable on your element, the default value will be used, which is `2`.

##### OLD WAY:
```css
   overflow: hidden;
   text-overflow: ellipsis;
   display: -webkit-box;
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
```

##### NOW:
```xml
<!-- If you don't define `--lines-clamp` it'll clamp `2` lines by default !-->
<myElement class="o_line_clamp" style="--lines-clamp: 3;"/>
```

This commit clears up the `-webkit-line-clamp` handled in CSS and replace them with the `.o_line_clamp` CSS class.

[^1]: https://github.com/odoo/odoo/commit/89779db8c103f9d72c52db07db436b08dcaeb914
[^2]: https://github.com/odoo/odoo/commit/f8497b16ce292947a4ede8de052e32130e092c60

task-5905780

---

### DIFF
--- a/addons/mail/static/src/core/common/link_preview.xml
+++ b/addons/mail/static/src/core/common/link_preview.xml
@@ -9,10 +9,10 @@
 
     <t t-name="mail.LinkPreviewCard.content">
         <span t-if="this.linkPreview.og_site_name" class="text-muted smaller" t-out="this.linkPreview.og_site_name"/>
-        <p class="o-mail-LinkPreviewCard-title mb-0" t-att-class="{ 'me-4': !this.linkPreview.og_site_name }">
+        <p class="o-mail-LinkPreviewCard-title o_line_clamp mb-0" t-att-class="{ 'me-4': !this.linkPreview.og_site_name }">
             <a t-att-href="this.linkPreview.source_url" target="_blank" t-out="this.linkPreview.og_title"/>
         </p>
-        <p t-if="this.linkPreview.og_description" class="o-mail-LinkPreviewCard-description mt-1 mb-0 text-muted smaller" t-out="this.linkPreview.og_description"/>
+        <p t-if="this.linkPreview.og_description" class="o-mail-LinkPreviewCard-description o_line_clamp o_line_clamp_3 mt-1 mb-0 text-muted smaller" t-out="this.linkPreview.og_description"/>
     </t>
     <t t-name="mail.LinkPreviewCard">
         <div class="o-mail-LinkPreviewCard card mb-1 rounded bg-view shadow-sm overflow-hidden" t-att-class="{ 'o-mail-LinkPreviewCard-noImage': !this.linkPreview.og_image, 'w-100': this.ui.isSmall or this.env.inChatWindow }" t-attf-class="{{ this.className }}">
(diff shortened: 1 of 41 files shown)


chgo-odoo:
cc @adr-odoo @rfr-odoo

addons/mail/static/src/core/common/link_preview.xml
```diff
@@ -9,10 +9,10 @@
 
     <t t-name="mail.LinkPreviewCard.content">
         <span t-if="this.linkPreview.og_site_name" class="text-muted smaller" t-out="this.linkPreview.og_site_name"/>
-        <p class="o-mail-LinkPreviewCard-title mb-0" t-att-class="{ 'me-4': !this.linkPreview.og_site_name }">
+        <p class="o-mail-LinkPreviewCard-title o_line_clamp mb-0" t-att-class="{ 'me-4': !this.linkPreview.og_site_name }">
             <a t-att-href="this.linkPreview.source_url" target="_blank" t-out="this.linkPreview.og_title"/>
         </p>
-        <p t-if="this.linkPreview.og_description" class="o-mail-LinkPreviewCard-description mt-1 mb-0 text-muted smaller" t-out="this.linkPreview.og_description"/>
+        <p t-if="this.linkPreview.og_description" class="o-mail-LinkPreviewCard-description o_line_clamp mt-1 mb-0 text-muted smaller" style="--lines-clamp: 3;" t-out="this.linkPreview.og_description"/>
```
seb-odoo: Is it really necessary/an improvement to have inline style instead?

chgo-odoo: I do believe it's an improvement, it'll prevent us from creating specific selector only to handle line clamp + this inline-style is only needed IF your value is different from the default one and you don't manipulate the value depending on the breakpoint. We're just parameterising a reusable class, in the end.

You can and should still apply it in CSS if you need different values of course

adr-odoo: What if we generate or add some extra classes. 
`.o_line_clamp_1` -> _10 
or 
only the useful ones like 1, 2 (default), 3, 5, 7 and 10. I let you to choose.
I'm not against replace all `.o_line_clamp` by `.o_line_clamp_2`.

addons/web/static/src/scss/utilities_custom.scss
```diff
@@ -207,10 +207,19 @@ $utilities: map-merge(
...
 }
 
-// Multiline text truncation utility : limits visible text to N lines (2 by default)
+// Multiline text truncation utility: limits visible text to N lines.
+// Use .o_line_clamp_N for a predefined count (1, 2, 3, 5, 7, 10),
+// or .o_line_clamp with a --lines-clamp CSS variable for other values.
 .o_line_clamp {
     display: -webkit-box;
-    -webkit-line-clamp: var(--lines-clamp, 2);
+    -webkit-line-clamp: var(--lines-clamp);
```
adr-odoo: Keep 2 as default.

addons/web/static/src/scss/utilities_custom.scss
```diff
@@ -207,10 +207,19 @@ $utilities: map-merge(
...
     -webkit-box-orient: vertical;
     overflow: hidden;
 }
+
+@each $lines in 1, 2, 3, 5, 7, 10 {
+    .o_line_clamp_#{$lines} {
+        @extend .o_line_clamp;
```
adr-odoo: remove this extend. It's ok to have `..o_line_clamp .o_line_clamp_3` in xml files.

addons/mail/static/src/core/common/link_preview.xml
chgo-odoo: @adr-odoo implemented this approach [here](https://github.com/odoo/odoo/compare/4e2daa9bc749359d3e99b1def9f4d0ab93e943f4..6bf4f2bc6f2d1917d59fd22a3c3defdbbbb9acd6)

Could you check [the part in utility_custom.scss](https://github.com/odoo/odoo/compare/4e2daa9bc749359d3e99b1def9f4d0ab93e943f4..6bf4f2bc6f2d1917d59fd22a3c3defdbbbb9acd6#diff-dd64a5af6d918fc16ee98120c43c69c89290eba058311fe0026e9a1916e86dbbR213) to see if it matches what you had in mind?

adr-odoo:
@robodoo r-
Waiting OWL3

adr-odoo:
@chgo-odoo Can I ask you to rebase?

adr-odoo:
@robodoo r+

PR #248080, opened 2026-02-10
[FIX] website: prevent PageBreadcrumb crash without header

FrancoisGe:
When the header is disabled globally via the Theme tab in edit mode, navigating to a page containing a breadcrumb caused a crash.

The PageBreadcrumb interaction did not handle the case where no header was present on the page.

This commit updates the interaction to safely handle pages without a header.

Task-ID: 5927177

Current behavior before PR:

---

### DIFF
--- a/addons/website/static/src/interactions/page_breadcrumb.js
+++ b/addons/website/static/src/interactions/page_breadcrumb.js
@@ -11,12 +11,11 @@ export class PageBreadcrumb extends Interaction {
         const updatePageBreadcrumbOnResize = this.protectSyncAfterAsync(
             this.updatePageBreadcrumbOnResize.bind(this)
         );
-        this.headerObserver = new ResizeObserver(updatePageBreadcrumbOnResize);
-        this.headerObserver.observe(this.headerEl);
-    }
-
-    destroy() {
-        this.headerObserver.disconnect();
+        if (this.headerEl) {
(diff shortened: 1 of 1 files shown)


Syozik (approved):
Thanks for the fix 👋 
While testing it, I noticed another breadcrumb issue present, but not related to your PR.
If we change the header template to be 'Sidebar' and set the breadcrumb position to 'Over The Content`, the breadcrumb will be at the bottom of the page after the footer. instead of on the top. Also, its width should be changed. Here's the screenshot for reference.

Image

This is not related to your PR, but can be fixed together as the same interaction changes the breadcrumb's `style.top`. Or maybe you would prefer to create a new task for that?

addons/website/static/src/interactions/page_breadcrumb.js
```diff
@@ -11,12 +11,14 @@ export class PageBreadcrumb extends Interaction {
...
-        this.headerObserver.observe(this.headerEl);
+        if (this.headerEl) {
+            this.headerObserver = new ResizeObserver(updatePageBreadcrumbOnResize);
+            this.headerObserver.observe(this.headerEl);
+        }
     }
 
     destroy() {
-        this.headerObserver.disconnect();
+        this.headerObserver?.disconnect();
```
aans-odoo: We could also use `registerCleanup` inside the `if` block to disconnect the observer instead of doing it in `destroy`.

PR #248123, opened 2026-02-11
[FIX] stock_account: Prevent memory error

kihy-odoo:
Related ticket: https://www.odoo.com/odoo/project/49/tasks/5416006

Issue:
Opening the inventory valuation report can trigger a memory error under these conditions:
1. There are millions of stock.move records.
2. Many of the products use the average cost method.

Explanation:
When the inventory valuation report is opened, the total_value of every product is computed. If the product uses the average cost method and there is quantity available, all of the product's stock moves and stock move lines are queried. This can take up too much memory.

Solution:
To prevent this, we periodically clear the cache for the stock.move and stock.move.line models.

opw-5416006

---

### DIFF
--- a/addons/stock_account/models/product.py
+++ b/addons/stock_account/models/product.py
@@ -3,7 +3,7 @@
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError
 from odoo.fields import Domain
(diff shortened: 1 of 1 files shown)


kihy-odoo:
The _compute_value method has been recently changed extensively so that this fix is no longer possible.

PR #248126, opened 2026-02-11
[FIX] l10n_fr_fec: use correct company when downloading FEC file

mograby3500:
The FEC generation spans two HTTP requests: generate_fec (RPC call)
and the download controller (plain GET). RPC calls include
allowed_company_ids in the context, so self.env.company resolves
correctly. But the plain GET to /download/fec_file/<id> carries no
context, so self.env.company falls back to user.company_id (the user's
default company), which may differ from the one selected in the UI.

This commit aims to fix the issue by passing
the company_id from generate_fec through the download URL and restore it in the controller via with_company(), with an access check to ensure the user belongs to that company.

task-5926528

### DIFF
--- a/addons/l10n_fr_fec/controllers/main.py
+++ b/addons/l10n_fr_fec/controllers/main.py
@@ -1,11 +1,15 @@
 from odoo import http
+from odoo.exceptions import AccessDenied
 from odoo.http import request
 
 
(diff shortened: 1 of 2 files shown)


FlorianGilbert (approved):
@robodoo r+

PR #248130, opened 2026-02-11
[ADD] l10n_gr_edi: implement CIUS for Greece

smjo-odoo:
In this PR:

- Implemented Greek CIUS (core invoice usage specification based on BIS-3) which is mandatory for all B2G e-invoices.
- Related PR: https://github.com/odoo/upgrade/pull/9465

task-4981749

### DIFF
--- a/addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
+++ b/addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
@@ -1,0 +1,156 @@
+from stdnum.gr import vat
+
+from odoo import models
+
+
+class AccountEdiXmlUblGr(models.AbstractModel):
+    _name = 'account.edi.xml.ubl_gr'
+    _inherit = 'account.edi.xml.ubl_bis3'
+    _description = "CIUS GR"
+
+    def _export_invoice_filename(self, invoice):
+        return f"{invoice.name.replace('/', '_')}_gr_peppol_cius.xml"
+
+    def _format_greek_invoice_number(self, invoice):
+        # GR-R-001-2: First segment - VAT without EL prefix
+        company_vat = vat.compact(invoice.company_id.vat or '')
+        invoice_name = invoice.name.split('/')
+        # GR-R-001-3: Second segment - Issue date DD/MM/YYYY
+        issue_date = invoice.invoice_date.strftime('%d/%m/%Y')
+        # GR-R-001-4: Third segment - Installation serial(branch number)
+        installation_sn = str(invoice.company_id.l10n_gr_edi_branch_number)
+        # GR-R-001-5: Fourth segment - Valid Greek document type
+        invoice_type = invoice.l10n_gr_edi_inv_type
+        # GR-R-001-6: Fifth segment - Series
+        series = invoice_name[0]
+        # GR-R-001-7: Sixth segment - Serial number
+        serial_number = invoice_name[-1]
+        return f"{company_vat}|{issue_date}|{installation_sn}|{invoice_type}|{series}|{serial_number}"
+
+    def _add_invoice_header_nodes(self, document_node, vals):
+        super()._add_invoice_header_nodes(document_node, vals)
+        invoice = vals['invoice']
+        document_node.update({
+            # BT-1: Invoice number with format: "VAT|Issue date|Installation sn.|Invoice Type|Series|eINV Issue sn."
+            'cbc:ID': {'_text': self._format_greek_invoice_number(invoice)},
+        })
+        # BT-10: Buyer reference - Contracting authority name
+        buyer = invoice.partner_id.commercial_partner_id
+        if buyer.l10n_gr_edi_contracting_authority_name:
+            document_node['cbc:BuyerReference'] = {'_text': buyer.l10n_gr_edi_contracting_authority_name}
+
+    # -------------------------------------------------------------------------
+    # EXPORT VALUES
+    # -------------------------------------------------------------------------
+
+    def _get_invoice_node(self, vals):
+        document_node = super()._get_invoice_node(vals)
+        invoice = vals['invoice']
+        additional_document_references = document_node.setdefault('cac:AdditionalDocumentReference', [])
+        if not isinstance(additional_document_references, list):
+            additional_document_references = [additional_document_references]
+            document_node['cac:AdditionalDocumentReference'] = additional_document_references
+        additional_document_references.append({
+            'cbc:ID': {'_text': str(invoice.l10n_gr_edi_mark)},
+            'cbc:DocumentDescription': {'_text': '##M.AR.K##'},
+        })
+
+        if vals['document_type'] != 'credit_note':
+            # BT-11: ProjectReference
+            document_node['cac:ProjectReference'] = {
+                'cbc:ID': {'_text': f"{invoice.l10n_gr_edi_budget_type}|{invoice.l10n_gr_edi_project_reference}"}
+            }
+            # BT-12: ContractDocumentReference
+            document_node['cac:ContractDocumentReference'] = {
+                'cbc:ID': {'_text': invoice.l10n_gr_edi_contract_reference}
+            }
+        # BT-25 Billing reference (preceding invoice number) for credit notes
+        elif vals['document_type'] == 'credit_note':
+            reversed_entry = invoice.reversed_entry_id
+            document_node['cac:BillingReference'] = {
+                'cac:InvoiceDocumentReference': {
+                    'cbc:ID': {'_text': self._format_greek_invoice_number(reversed_entry) if reversed_entry else ''},
+                }
+            }
+        return document_node
+
+    # -------------------------------------------------------------------------
+    # Party Nodes (Supplier & Customer)
+    # -------------------------------------------------------------------------
+
+    def _ubl_add_party_endpoint_id_node(self, vals):
+        super()._ubl_add_party_endpoint_id_node(vals)
+        party_node = vals['party_node']
+        endpoint_node = party_node.get('cbc:EndpointID')
+        if endpoint_node and endpoint_node.get('schemeID') == '9933':
+            endpoint_node['_text'] = vat.compact(endpoint_node.get('_text') or '')
+
+    def _ubl_add_party_identification_nodes(self, vals):
+        super()._ubl_add_party_identification_nodes(vals)
+        partner = vals['party_vals']['partner']
+        buyer = vals['customer'].commercial_partner_id
+        # BT-46: Contracting authority code (HT code for Greek contracting authorities)
+        if partner.id == vals['customer'].id and buyer.l10n_gr_edi_contracting_authority_code:
+            nodes = vals['party_node'].setdefault('cac:PartyIdentification', [])
+            nodes.append({
+                'cbc:ID': {'_text': buyer.l10n_gr_edi_contracting_authority_code}
+            })
+
+    # -------------------------------------------------------------------------
+    # Line Item Nodes
+    # -------------------------------------------------------------------------
+
+    def _add_invoice_line_item_nodes(self, line_node, vals):
+        super()._add_invoice_line_item_nodes(line_node, vals)
+        line = vals['base_line']['record']
+        # BT-158 CPV (Common Procurement Vocabulary) code
+        if not line or not line.l10n_gr_edi_cpv_code:
+            return
+        item = line_node.setdefault('cac:Item', {})
+        commodity_node = item.setdefault('cac:CommodityClassification', [])
+        commodity_node.append({
+            'cbc:ItemClassificationCode': {
+                '_text': line.l10n_gr_edi_cpv_code,
+                'listID': 'STI'
+            }
+        })
+
+    # -------------------------------------------------------------------------
+    # CONSTRAINTS
+    # -------------------------------------------------------------------------
+
+    def _export_invoice_constraints(self, invoice, vals):
+        constraints = super()._export_invoice_constraints(invoice, vals) or {}
+        constraints.update(self._validate_greek_business_rules(invoice))
+        return constraints
+
+    def _validate_greek_business_rules(self, invoice):
+        """Validation of all Greek PEPPOL CIUS business rules"""
+        constraints = {}
+        supplier_vat = invoice.company_id.vat or ''
+        buyer = invoice.partner_id.commercial_partner_id
+        buyer_vat = buyer.vat or ''
+        _ = self.env._
+        # check format : (code, condition, error message)
+        checks = [
+            ('gr_r_001_5', not invoice.l10n_gr_edi_inv_type, _("Missing Greek Invoice type")),
+            ('gr_r_003', not supplier_vat.upper().startswith('EL'), _("Supplier VAT must start with 'EL'")),
+            ('gr_r_004_1', not invoice.l10n_gr_edi_mark, _("M.AR.K number is required for Suppliers")),
+            ('gr_r_006', not buyer_vat or not buyer_vat.upper().startswith('EL'), _("Buyer VAT must start with prefix 'EL'")),
+            ('gr_bt_10', not buyer.l10n_gr_edi_contracting_authority_name, _("Contracting authority name is required for Greek buyer")),
+            ('gr_bt_46', not buyer.l10n_gr_edi_contracting_authority_code, _("Contracting authority code is required for Greek buyer")),
+            ('gr_r_007', invoice.move_type != 'out_refund' and not (invoice.l10n_gr_edi_budget_type and invoice.l10n_gr_edi_project_reference), _("Budget Type and Project reference is required for B2G invoicing")),
+            ('gr_bt_25', invoice.move_type == 'out_refund' and not invoice.reversed_entry_id, _("A Greek CIUS credit note must reference the original invoice.")),
+        ]
+        for code, condition, message in checks:
+            if condition:
+                constraints[code] = message
+        # BT-158: CPV code (KED mandatory fields)
+        for line in invoice.invoice_line_ids:
+            if line.product_id and not line.l10n_gr_edi_cpv_code:
+                constraints['gr_bt_158'] = _(
+                    "Each invoice line with a product must have a CPV code for Greek B2G invoicing."
+                    "Set CPV code on the product or on the line."
+                )
+                break
+        return constraints
--- a/addons/l10n_gr_edi/models/account_move.py
+++ b/addons/l10n_gr_edi/models/account_move.py
@@ -86,6 +86,22 @@ class AccountMove(models.Model):
         compute='_compute_from_l10n_gr_edi_document_ids',
         store=True,
     )
+    l10n_gr_edi_budget_type = fields.Selection([
+        ('1', '1 - Ordinary Budget'),
+        ('2', '2 - PDE (Public Investment Budget)'),
+        ('3', '3 - Other Budget')
+        ], string="Budget Type"
+    )
+    l10n_gr_edi_project_reference = fields.Char(
+        string='Project reference number',
+        help='ADA or Enaritmos number',
+    )
+    l10n_gr_edi_contract_reference = fields.Char(
+        string='Contract ADAM',
+        help='Internet contract posting number(ADAM) of the Central Electronic Register of Public Procurement (KIMDIS)',
+        default='0',
+        required=True,
+    )
 
     def _auto_init(self):
         """
@@ -98,6 +114,9 @@ class AccountMove(models.Model):
             ('l10n_gr_edi_inv_type', 'varchar'),
             ('l10n_gr_edi_payment_method', 'varchar'),
             ('l10n_gr_edi_attachment_id', 'int4'),
+            ('l10n_gr_edi_budget_type', 'varchar'),
+            ('l10n_gr_edi_project_reference', 'varchar'),
+            ('l10n_gr_edi_contract_reference', 'varchar'),
         ):
             if not column_exists(self.env.cr, 'account_move', column_name):
                 create_column(self.env.cr, 'account_move', column_name, column_type)
@@ -786,3 +805,19 @@ class AccountMove(models.Model):
             raise UserError(_("Some of the selected moves does not meet the requirements to be sent to myDATA."))
 
         self.l10n_gr_edi_try_send_expense_classification()
+
+    ################################################################################
+    # Peppol/UBL helpers
+    ################################################################################
+
+    def _need_ubl_cii_xml(self, ubl_cii_format):
+        """
+        Override the generic check to postpone the creation of the UBL file
+        for Greek B2G invoices until we actually receive the `l10n_gr_edi_mark`
+        from myDATA.
+        """
+        if not super()._need_ubl_cii_xml(ubl_cii_format):
+            return False
+        if ubl_cii_format == 'ubl_gr':
+            return bool(self.l10n_gr_edi_mark)
+        return True
--- a/addons/l10n_gr_edi/models/account_move_send.py
+++ b/addons/l10n_gr_edi/models/account_move_send.py
@@ -8,6 +8,19 @@ class AccountMoveSend(models.AbstractModel):
     def _is_gr_edi_applicable(self, move):
         return move.l10n_gr_edi_enable_send_invoices
 
+    def _is_applicable_to_move(self, method, move, **move_data):
+        # EXTENDS 'account'
+        # For Greek invoices with Peppol in ubl_gr format, prevent sending via Peppol until myDATA mark is received
+        res = super()._is_applicable_to_move(method, move, **move_data)
+        if not res or method != 'peppol' or move.country_code != 'GR':
+            return res
+        partner = move.partner_id.commercial_partner_id.with_company(move.company_id)
+        invoice_edi_format = move_data.get('invoice_edi_format') or partner._get_peppol_edi_format()
+        # Block Peppol for Greek CIUS format if no mark yet
+        if invoice_edi_format == 'ubl_gr' and not move.l10n_gr_edi_mark:
+            return False
+        return res
+
     def _get_all_extra_edis(self) -> dict:
         # EXTENDS 'account'
         res = super()._get_all_extra_edis()
@@ -33,6 +46,18 @@ class AccountMoveSend(models.AbstractModel):
                 'action': invoices_with_alert._get_records_action(name=_("Check Invoice(s)")),
             }
 
+        # Alert for Greek invoices with Peppol that haven't been sent to myDATA yet
+        move_error = moves.filtered(lambda move: (move.country_code == 'GR'
+            and 'peppol' in moves_data[move]['sending_methods']
+            and moves_data[move]['invoice_edi_format'] == 'ubl_gr'
+            and not move.l10n_gr_edi_mark
+        ))
+        if move_error:
+            alerts['l10n_gr_edi_peppol_requires_mydata'] = {
+                'message': self.env._("Invoice(s) are not yet sent to myDATA."
+                             " First enable myDATA sending, then you can send via Peppol for B2G invoicing."),
+                'level': 'warning',
+            }
         return alerts
 
     # -------------------------------------------------------------------------
--- a/addons/l10n_gr_edi/models/res_partner.py
+++ b/addons/l10n_gr_edi/models/res_partner.py
@@ -1,5 +1,10 @@
+import re
+
 from odoo import api, fields, models
+from odoo.exceptions import ValidationError
 from odoo.tools.sql import column_exists, create_column
+
+_CONTRACTING_AUTHORITY_REGEX = re.compile(r'^\d+\.\d+\.\d+$')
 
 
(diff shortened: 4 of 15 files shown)


addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+        # GR-R-001-3: Second segment - Issue date DD/MM/YYYY
+        issue_date = invoice.invoice_date.strftime('%d/%m/%Y')
+        # GR-R-001-4: Third segment - Installation serial(branch number)
+        installation_sn = str(invoice.company_id.l10n_gr_edi_branch_number)
+        # GR-R-001-5: Fourth segment - Valid Greek document type
+        invoice_type = invoice.l10n_gr_edi_inv_type
+        # GR-R-001-6: Fifth segment - Series
+        series = invoice.name.split('/')[0]
+        # GR-R-001-7: Sixth segment - Serial number
+        serial_number = invoice.name.split('/')[-1]
```
zepa-odoo: `invoice.name.split('/')` is called twice here. You could store the result in a variable and reuse it to improve readability and avoid repeating the same operation

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+            # BT-11: ProjectReference
+            if invoice.l10n_gr_edi_budget_type and invoice.l10n_gr_edi_project_reference:
+                document_node['cac:ProjectReference'] = {
+                    'cbc:ID': {'_text': f"{invoice.l10n_gr_edi_budget_type}|{invoice.l10n_gr_edi_project_reference}"}
+                }
+            # BT-12: ContractDocumentReference
+            if invoice.l10n_gr_edi_contract_reference:
+                document_node['cac:ContractDocumentReference'] = {
+                    'cbc:ID': {'_text': invoice.l10n_gr_edi_contract_reference}
+                }
```
zepa-odoo: According to `BT-12`, `cac:ContractDocumentReference` must always be present. If there is no contract signed, the value `0`  should be provided. Since `l10n_gr_edi_contract_reference` already has a `default value of 0`, it might be better to mark this field as required to ensure the value is always set and to prevent missing data before generating the EDI document.

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+    # -------------------------------------------------------------------------
+    # Party Nodes (Supplier & Customer)
+    # -------------------------------------------------------------------------
+
+    def _ubl_add_party_identification_nodes(self, vals):
+        super()._ubl_add_party_identification_nodes(vals)
+        partner = vals['party_vals']['partner']
+        # BT-46: Contracting authority code (HT code for Greek contracting authorities)
+        if partner.id == vals['customer'].id and partner.l10n_gr_edi_contracting_authority_code:
+            nodes = vals['party_node']['cac:PartyIdentification']
```
zepa-odoo: `party_node` usually already contains `cac:PartyIdentification` as a list. However, if there is any chance that the key is missing or not initialized as a list, it would be safer to initialize it using `setdefault()` before appending to it

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+        super()._add_invoice_line_item_nodes(line_node, vals)
+        line = vals['base_line']['record']
+        # BT-158 CPV (Common Procurement Vocabulary) code
+        if line and line.product_id:
+            cpv_code = line.product_id.l10n_gr_edi_cpv_code
+            if cpv_code:
+                if 'cac:Item' not in line_node:
+                    line_node['cac:Item'] = {}
+                if 'cac:CommodityClassification' not in line_node['cac:Item']:
+                    line_node['cac:Item']['cac:CommodityClassification'] = []
```
zepa-odoo: The node initialization of `cac:Item` and `cac:CommodityClassification` could be simplified using `setdefault()` instead of manual checks.

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+    # Line Item Nodes
+    # -------------------------------------------------------------------------
+
+    def _add_invoice_line_item_nodes(self, line_node, vals):
+        super()._add_invoice_line_item_nodes(line_node, vals)
+        line = vals['base_line']['record']
+        # BT-158 CPV (Common Procurement Vocabulary) code
+        if line and line.product_id:
+            cpv_code = line.product_id.l10n_gr_edi_cpv_code
+            if cpv_code:
```
zepa-odoo: You could simplify the implementation by returning early when `product_id` or the `CPV code` is missing. This would reduce the nesting level.

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,153 @@
...
+        return constraints
+
+    def _validate_greek_business_rules(self, invoice):
+        """Comprehensive validation of all Greek PEPPOL CIUS business rules"""
+        constraints = {}
+        vat = invoice.company_id.vat or ''
+        buyer_vat = invoice.partner_id.commercial_partner_id.vat
+        # GR-R-001-5: Fourth segment - Valid Greek document type
+        if not invoice.l10n_gr_edi_inv_type:
+            constraints['gr_r_001_5'] = self.env._("Missing Greek Invoice type")
```
zepa-odoo: We could probably avoid repeating multiple if blocks by defining a mapping of `error_code -> (condition, message)` and `iterating over it`. This would make the validations more compact and easier to extend if new Greek business rules are added later.

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -786,3 +801,19 @@ def _l10n_gr_edi_try_send_batch(self):
...
+        """
+        Override the generic check to postpone the creation of the UBL file
+        for Greek B2G invoices until we actually receive the ``l10n_gr_edi_mark``
+        from myDATA.
+        """
+        if not super()._need_ubl_cii_xml(ubl_cii_format):
+            return False
+        return (
+            bool(self.l10n_gr_edi_mark) if ubl_cii_format == 'ubl_gr' else True
+        )
```
zepa-odoo:
```suggestion
        if ubl_cii_format == 'ubl_gr':
            return bool(self.l10n_gr_edi_mark)
        return True
```

Nitpick: could be made more readable by avoiding the inline conditional in the return.

addons/l10n_gr_edi/models/account_move_send.py
```diff
@@ -8,6 +8,18 @@ class AccountMoveSend(models.AbstractModel):
     def _is_gr_edi_applicable(self, move):
         return move.l10n_gr_edi_enable_send_invoices
 
+    def _is_applicable_to_move(self, method, move, **move_data):
+        # EXTENDS 'account'
+        # For Greek invoices with Peppol in ubl_gr format, prevent sending via Peppol until myDATA mark is received
+        res = super()._is_applicable_to_move(method, move, **move_data)
+        if method == 'peppol' and move.country_code == 'GR':
```
zepa-odoo: You could add an early return `(if not res or method != 'peppol' or move.country_code != 'GR': return res)` to avoid nesting and make the logic easier to follow.

addons/l10n_gr_edi/models/account_move_send.py
```diff
@@ -32,7 +44,22 @@ def _get_alerts(self, moves, moves_data):
...
+        greek_moves = moves.filtered(lambda m: m.country_code == 'GR')
+        for move in greek_moves:
+            move_data = moves_data[move]
+            edi_format = move_data.get('invoice_edi_format')
+            # Check if Peppol is a sending method but no mark yet
+            if (
+                'peppol' in move_data.get('sending_methods', {})
+                and edi_format == 'ubl_gr'
+                and not move.l10n_gr_edi_mark
+            ):
```
zepa-odoo: You could filter the moves directly with `filtered(...)` instead of looping and checking the condition inside the loop.

addons/l10n_gr_edi/models/res_partner.py
```diff
@@ -25,3 +37,20 @@ def _compute_l10n_gr_edi_branch_number(self):
                 partner.l10n_gr_edi_branch_number = partner.l10n_gr_edi_branch_number or 0
             else:
                 partner.l10n_gr_edi_branch_number = False
+
+    @api.constrains('l10n_gr_edi_contracting_authority_code')
+    def _check_l10n_gr_edi_contracting_authority_format(self):
+        for partner in self:
+            if partner.country_id.code == 'GR' and partner.l10n_gr_edi_contracting_authority_code:
+                if not re.match(r'^\d+\.\d+\.\d+$', partner.l10n_gr_edi_contracting_authority_code):
+                    raise ValidationError(self.env._("Greek contracting authority code is in invalid format"))
```
zepa-odoo:
```suggestion
                    raise ValidationError(self.env._("The Greek contracting authority code format is invalid."))
```

addons/l10n_gr_edi/models/res_partner.py
```diff
@@ -25,3 +37,20 @@ def _compute_l10n_gr_edi_branch_number(self):
                 partner.l10n_gr_edi_branch_number = partner.l10n_gr_edi_branch_number or 0
             else:
                 partner.l10n_gr_edi_branch_number = False
+
+    @api.constrains('l10n_gr_edi_contracting_authority_code')
+    def _check_l10n_gr_edi_contracting_authority_format(self):
+        for partner in self:
+            if partner.country_id.code == 'GR' and partner.l10n_gr_edi_contracting_authority_code:
+                if not re.match(r'^\d+\.\d+\.\d+$', partner.l10n_gr_edi_contracting_authority_code):
```
zepa-odoo: You could compile the regex once (e.g., as a module-level constant) instead of compiling it at each iteration.

addons/l10n_gr_edi/models/res_partner.py
```diff
@@ -25,3 +37,20 @@ def _compute_l10n_gr_edi_branch_number(self):
                 partner.l10n_gr_edi_branch_number = partner.l10n_gr_edi_branch_number or 0
             else:
                 partner.l10n_gr_edi_branch_number = False
+
+    @api.constrains('l10n_gr_edi_contracting_authority_code')
+    def _check_l10n_gr_edi_contracting_authority_format(self):
+        for partner in self:
+            if partner.country_id.code == 'GR' and partner.l10n_gr_edi_contracting_authority_code:
```
zepa-odoo: You could filter the recordset first (`self.filtered(...)`) to avoid nested conditions

addons/l10n_gr_edi/views/product_template_views.xml
```diff
@@ -30,4 +30,14 @@
         </field>
     </record>
 
+    <record id="product_template_form_view_inherit_l10n_gr_edi_cpv_code" model="ir.ui.view">
+        <field name="name">product.template.form.inherit.l10n.gr.edi.cpv.code</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <field name="categ_id" position="after">
+                <field name="l10n_gr_edi_cpv_code"/>
```
zepa-odoo: missing country condition it should not visible in other countries

addons/l10n_gr_edi/views/account_move_views.xml
```diff
@@ -81,6 +82,16 @@
                        column_invisible="parent.country_code != 'GR'"
                        readonly="not l10n_gr_edi_need_exemption_category"/>
             </xpath>
+            <xpath expr="//notebook/page[@name='other_info']//field[@name='delivery_date']" position="after">
+                <field name="l10n_gr_edi_budget_type"
+                       string="Budget Type"/>
```
zepa-odoo: missing country condition

zepa-odoo: same for `l10n_gr_edi_project_reference` and `l10n_gr_edi_contract_reference`

addons/l10n_gr_edi/tests/test_xml_ubl_gr.py
```diff
@@ -0,0 +1,141 @@
...
+                                           tax_ids=self.company_data['default_tax_sale']),
+                self._prepare_invoice_line(product_id=self.product_b, quantity=1.0, price_unit=5.1,
+                                           tax_ids=self.company_data['default_tax_sale']),
+            ],
+            **b2g_defaults,
+        )
+
+    def _mark_as_sent(self, invoice, mark):
+        invoice.l10n_gr_edi_state = 'invoice_sent'
+        invoice.l10n_gr_edi_mark = mark
```
zepa-odoo, line 75: is there any reason we need different values for `l10n_gr_edi_mark`? 
Since the actual value may not matter for the test logic, could we just set a **static value for all documents** instead of passing it dynamically? This might simplify the test setup.

addons/l10n_gr_edi/tests/test_xml_ubl_gr.py
```diff
@@ -0,0 +1,141 @@
...
+        refund.reversed_entry_id = original_invoice.id
+        self._mark_as_sent(refund, mark=123456789)
+
+        self._assert_xml_fixture_equal(refund, 'grcius_out_refund')
+
+    # -------------------------------------------------------------------------
+    # CREDIT NOTE TESTS
+    # -------------------------------------------------------------------------
+
+    def test_credit_note_preceding_invoice_reference(self):
```
zepa-odoo: This test looks very similar to `test_export_greek_credit_note` and shares the same setup. It might be possible to combine them to avoid duplication while still validating both behaviors.

addons/l10n_gr_edi/tests/test_xml_ubl_gr.py
smjo-odoo: Yes, we need a test case with 2 different MARK values in case we create a credit note, we would require the previous invoice's MARK number.

Yes, I will make the values static and simplify the code.

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,144 @@
...
+        return constraints
+
+    def _validate_greek_business_rules(self, invoice):
+        """Comprehensive validation of all Greek PEPPOL CIUS business rules"""
+        constraints = {}
+        vat = invoice.company_id.vat or ''
+        buyer_vat = invoice.partner_id.commercial_partner_id.vat
+
+        checks = [
+            ('gr_r_001_5', not invoice.l10n_gr_edi_inv_type, self.env._("Missing Greek Invoice type")),
```
zepa-odoo: Add a comment for tuple order (code, condition, message) for clarity.

addons/l10n_gr_edi/views/product_template_views.xml
zepa-odoo: missing country condition

addons/l10n_gr_edi/views/product_template_views.xml
```diff
@@ -30,4 +30,14 @@
...
+    <record id="product_template_form_view_inherit_l10n_gr_edi_cpv_code" model="ir.ui.view">
+        <field name="name">product.template.form.inherit.l10n.gr.edi.cpv.code</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <field name="categ_id" position="after">
+                <field name="l10n_gr_edi_cpv_code" invisible="country_code != 'GR'"/>
+            </field>
+        </field>
+    </record>
```
zepa-odoo: No need to create a new view record, the CPV code can be added by extending `product_template_form_view_inherit_l10n_gr_edi` with an additional xpath.

addons/l10n_gr_edi/models/account_move.py
```diff
@@ -786,3 +802,19 @@ def _l10n_gr_edi_try_send_batch(self):
...
         self.l10n_gr_edi_try_send_expense_classification()
+
+    ################################################################################
+    # Peppol/UBL helpers
+    ################################################################################
+
+    def _need_ubl_cii_xml(self, ubl_cii_format):
+        """
+        Override the generic check to postpone the creation of the UBL file
+        for Greek B2G invoices until we actually receive the ``l10n_gr_edi_mark``
```
zepa-odoo:
```suggestion
        for Greek B2G invoices until we actually receive the `l10n_gr_edi_mark`
```

addons/l10n_gr_edi/models/account_move_send.py
```diff
@@ -32,7 +45,18 @@ def _get_alerts(self, moves, moves_data):
                 'action_text': _("View Invoice(s)"),
                 'action': invoices_with_alert._get_records_action(name=_("Check Invoice(s)")),
             }
-
+        # Alert for Greek invoices with Peppol that haven't been sent to myDATA yet
+        move_error = moves.filtered(lambda move: (move.country_code == 'GR'
+                and 'peppol' in moves_data[move]['sending_methods']
+                and moves_data[move]['invoice_edi_format'] == 'ubl_gr'
+                and not move.l10n_gr_edi_mark
+            ))
```
zepa-odoo:
```suggestion
        move_error = moves.filtered(lambda move: (move.country_code == 'GR'
            and 'peppol' in moves_data[move]['sending_methods']
            and moves_data[move]['invoice_edi_format'] == 'ubl_gr'
            and not move.l10n_gr_edi_mark
        ))
```

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,140 @@
...
+        return constraints
+
+    def _validate_greek_business_rules(self, invoice):
+        """Validation of all Greek PEPPOL CIUS business rules"""
+        constraints = {}
+        vat = invoice.company_id.vat or ''
+        buyer_vat = invoice.partner_id.commercial_partner_id.vat
+        # check format : (code, condition, error message)
+        checks = [
+            ('gr_r_001_5', not invoice.l10n_gr_edi_inv_type, self.env._("Missing Greek Invoice type")),
```
zepa-odoo: Consider assigning `self.env._` to `_` locally to avoid repetition and improve readability.

addons/l10n_gr_edi/models/account_move_send.py
```diff
@@ -32,7 +45,18 @@ def _get_alerts(self, moves, moves_data):
...
+        if move_error:
+            alerts['l10n_gr_edi_peppol_requires_mydata'] = {
+                'message': _("Invoice(s) are not yet sent to myDATA."
```
zepa-odoo:
```suggestion
                'message': self.env._("Invoice(s) are not yet sent to myDATA."
```

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,152 @@
...
+    def _add_invoice_header_nodes(self, document_node, vals):
+        super()._add_invoice_header_nodes(document_node, vals)
+        invoice = vals['invoice']
+        document_node.update({
+            # BT-1: Invoice number with format: "VAT|Issue date|Installation sn.|Invoice Type|Series|eINV Issue sn."
+            'cbc:ID': {'_text': self._format_greek_invoice_number(invoice)},
+        })
+        # BT-10: Buyer reference - Contracting authority name
+        if invoice.partner_id.l10n_gr_edi_contracting_authority_name:
+            document_node['cbc:BuyerReference'] = {'_text': invoice.partner_id.l10n_gr_edi_contracting_authority_name}
```
Megaaaaaa: This reads the contracting-authority name from the invoice contact while VAT validation uses the commercial partner, is it intended ?
```suggestion
        if invoice.partner_id.commercial_partner_id.l10n_gr_edi_contracting_authority_name:
            document_node['cbc:BuyerReference'] = {'_text': invoice.partner_id.commercial_partner_id.l10n_gr_edi_contracting_authority_name}
```

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,152 @@
...
+            }
+            # BT-12: ContractDocumentReference
+            document_node['cac:ContractDocumentReference'] = {
+                'cbc:ID': {'_text': invoice.l10n_gr_edi_contract_reference}
+            }
+        # BT-25 Billing reference (preceding invoice number) for credit notes
+        elif vals['document_type'] == 'credit_note':
+            document_node['cac:BillingReference'] = {
+                'cac:InvoiceDocumentReference': {
+                    'cbc:ID': {'_text': self._format_greek_invoice_number(invoice.reversed_entry_id)},
```
Megaaaaaa: Manually created credit notes can exist without `reversed_entry_id` and this will fail during export. Or do you expect them all to have a `reversed_entry_id` here ?
```suggestion
                    'cbc:ID': {'_text': self._format_greek_invoice_number(invoice.reversed_entry_id) if invoice.reversed_entry_id else ''},
```
And add something like `('gr_bt_25', invoice.move_type == 'out_refund' and not invoice.reversed_entry_id, _("A Greek CIUS credit note must reference the original invoice.")` to `def _validate_greek_business_rules():`

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
```diff
@@ -0,0 +1,152 @@
...
+        _ = self.env._
+        # check format : (code, condition, error message)
+        checks = [
+            ('gr_r_001_5', not invoice.l10n_gr_edi_inv_type, _("Missing Greek Invoice type")),
+            ('gr_r_003', not supplier_vat.upper().startswith('EL'), _("Supplier VAT must start with 'EL'")),
+            ('gr_r_004_1', not invoice.l10n_gr_edi_mark, _("M.AR.K number is required for Suppliers")),
+            ('gr_r_006', not buyer_vat or not buyer_vat.upper().startswith('EL'), _("Buyer VAT must start with prefix 'EL'")),
+            ('gr_bt_10', not invoice.partner_id.l10n_gr_edi_contracting_authority_name, _("Contracting authority name is required for Greek buyer")),
+            ('gr_bt_46', not invoice.partner_id.l10n_gr_edi_contracting_authority_code, _("Contracting authority code is required for Greek buyer")),
+            ('gr_r_007', not (invoice.l10n_gr_edi_budget_type and invoice.l10n_gr_edi_project_reference), _("Budget Type and Project reference is required for B2G invoicing")),
```
Megaaaaaa: You have a mismatch here between emission and validation. Should we not restrict this validation to documents that emit `ProjectReference` only ?
```suggestion
            ('gr_r_007', invoice.move_type != 'out_refund' and not (invoice.l10n_gr_edi_budget_type and invoice.l10n_gr_edi_project_reference), _("Budget Type and Project reference is required for B2G invoicing")),
```

addons/l10n_gr_edi/models/account_move_send.py
```diff
@@ -33,6 +46,18 @@ def _get_alerts(self, moves, moves_data):
...
+        # Alert for Greek invoices with Peppol that haven't been sent to myDATA yet
+        move_error = moves.filtered(lambda move: (move.country_code == 'GR'
+            and 'peppol' in moves_data[move]['sending_methods']
+            and moves_data[move]['invoice_edi_format'] == 'ubl_gr'
+            and not move.l10n_gr_edi_mark
+        ))
+        if move_error:
+            alerts['l10n_gr_edi_peppol_requires_mydata'] = {
+                'message': self.env._("Invoice(s) are not yet sent to myDATA."
+                             "First enable myDATA sending, then you can send via Peppol for B2G invoicing."),
```
Megaaaaaa: These two strings will be concatenated without a space.
```suggestion
                             " First enable myDATA sending, then you can send via Peppol for B2G invoicing."),
```

addons/l10n_gr_edi/models/res_partner.py
```diff
@@ -25,3 +44,27 @@ def _compute_l10n_gr_edi_branch_number(self):
...
+            raise ValidationError(self.env._("The Greek contracting authority code format is invalid"))
+
+    def _get_edi_builder(self, invoice_edi_format):
+        if invoice_edi_format == 'ubl_gr':
+            return self.env['account.edi.xml.ubl_gr']
+        return super()._get_edi_builder(invoice_edi_format)
+
+    def _get_suggested_invoice_edi_format(self):
+        if self.country_code == 'GR':
+            return 'ubl_gr'
```
Megaaaaaa: This makes Greek CIUS the suggested format for every Greek customer, while the scope is specifically Greek B2G Peppol invoicing. Is it intentionnal ? It will also steer non-public-sector Greek customers toward fields and validations that only make sense for contracting authorities. We could suggest `ubl_gr` only when the partner has B2G contracting-authority data, no ?
```suggestion
            return 'ubl_gr' if self.l10n_gr_edi_contracting_authority_code or self.l10n_gr_edi_contracting_authority_name else super()._get_suggested_invoice_edi_format()
```

smjo-odoo: Right, we should suggest `ubl_gr` only when the partner has B2G contracting-authority data :saluting_face:

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
smjo-odoo: Yes, we should add this restriction

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
smjo-odoo: It was unintended; it was missed out. I have corrected it now :)

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
smjo-odoo: The documentation states: 

> "In case of more than one previous debit invoice, all these debit Notes must carry the same Contract (ADAM, BT-12)
> and the same budget (BT-11). In the credit note, BT-11 and BT-12 are left empty, and therefore the software of KE.D of GSIS PA traces back to the first related debit invoice, to find them. It is recommended, for reasons of sound
> management of eINV, that an eINV credit be associated uniquely with a previous (equivalent) debit eINV." 

So, it makes sense to add a gr_bt_25 validation check :saluting_face:

addons/l10n_gr_edi/models/account_edi_xml_ubl_gr.py
smjo-odoo: @Megaaaaaa Should I do the same for methods `_ubl_add_party_identification_nodes` and `_validate_greek_business_rules`  as well ?
Example:
 ```
   def _ubl_add_party_identification_nodes(self, vals):
        super()._ubl_add_party_identification_nodes(vals)
        partner = vals['party_vals']['partner']
        buyer = partner.commercial_partner_id
        # BT-46: Contracting authority code (HT code for Greek contracting authorities)
        if partner.id == vals['customer'].id and buyer.l10n_gr_edi_contracting_authority_code:
            nodes = vals['party_node'].setdefault('cac:PartyIdentification', [])
            nodes.append({
                'cbc:ID': {'_text': buyer.l10n_gr_edi_contracting_authority_code}
            })
```

Megaaaaaa: Yes, we should make this consistent imo. The contracting-authority fields describe the legal authority, so they should be read from the commercial partner, just like the buyer VAT in `_validate_greek_business_rules`.

For `_ubl_add_party_identification_nodes`, I would use the commercial partner only for the customer party, for example `buyer = vals['customer'].commercial_partner_id` and  `buyer.l10n_gr_edi_contracting_authority_code`. For `_validate_greek_business_rules`, the `gr_bt_10` and `gr_bt_46` checks should also use buyer instead of `invoice.partner_id`, otherwise invoices addressed to a child contact can fail validation or miss the XML value even though the company-level customer has the required B2G data.

Do you agree with that ? Don't hesitate if I omitted anything :+1:

smjo-odoo: Yes, I agree. I updated BT-10/BT-46 export and validation to use the commercial buyer, and scoped the BT-46 node to the accounting customer party

smjo-odoo:
Hi @Megaaaaaa , could you please check the changes?

smjo-odoo:
@Megaaaaaa , I have re-applied the diff and pushed the corrected code again

smjo-odoo:
Hello @Megaaaaaa , the upgrade PR is approved. Can you please approve the community PR again ? Thanks :)

Megaaaaaa:
robodoo r+

PR #172329, opened 2024-07-08
[IMP] core: empty char fields are ""

kmagusiak:
Make char fields (text and html too) return always a value of type str.

task-4013641

odoo/enterprise#66267

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248132, opened 2026-02-11
[FIX] *: improve falsy vat values check

hamo-odoo:
Before this commit-
we used to check only if the
vat has a value or not. But in odoo we also
use `/`, `NA`, `na` as False values for the vat

After this commit-
We add a compute field `has_vat`,
to check if the vat has any default falsy value
or not

Related PR- https://github.com/odoo/enterprise/pull/111396
task-5935566

---

### DIFF
--- a/odoo/addons/base/models/res_partner.py
+++ b/odoo/addons/base/models/res_partner.py
@@ -316,7 +316,7 @@ class ResPartner(models.Model):
         precompute=True,  # avoid queries post-create
         readonly=False, store=True,
         help='The internal user in charge of this contact.')
-    vat = fields.Char(string='Tax ID', index=True, inverse='_inverse_vat', help="You can use '/' to indicate that the customer has no Tax ID.")
+    vat = fields.Char(string='Tax ID', index='btree_not_null', inverse='_inverse_vat', help="You can use '/' to indicate that the customer has no Tax ID.")
     has_vat = fields.Boolean(
         string="Has Tax ID",
         compute="_compute_has_vat",


hamo-odoo:
I haven't checked other l10ns, but possibly we can do a thorough run for them as well

hamo-odoo:
I know it's stupid but the orm.....
```py
In [1]: env.company.vat
Out[1]: False

In [2]: env.company.vat = ''

In [3]: env.company.vat
Out[3]: ''
```

addons/l10n_in/models/l10n_in_pan_entity.py
```diff
@@ -29,7 +29,7 @@ class L10nInPanEntity(models.Model):
         comodel_name='res.partner',
         inverse_name='l10n_in_pan_entity_id',
         string="Partners",
-        domain="[('l10n_in_pan_entity_id', '=', False), '|', ('vat', '=', False), ('vat', 'like', name)]"
+        domain="[('l10n_in_pan_entity_id', '=', False), '|', ('vat', 'in', [False, '', '/', 'NA', 'na']), ('vat', 'like', name)]"
```
jco-odoo: This one seems wrong.  Because it is meant to add new partners (it is a bit of a hack this thing anyways...)

jva-odoo: > This one seems wrong

How? 
it's normal that partner have PAN but not have GST number.
like Individual and Proprietor.

addons/l10n_in/views/report_invoice.xml
```diff
@@ -2,7 +2,7 @@
 <odoo>
     <template id="l10n_in_report_invoice_document_inherit" inherit_id="account.report_invoice_document" primary="True">
         <xpath expr="//div[@name='shipping_address_block']" position="inside">
-            <div t-if="o.company_id.account_fiscal_country_id.code == 'IN' and o.company_id.l10n_in_is_gst_registered and o.partner_shipping_id.vat">
+            <div t-if="o.company_id.account_fiscal_country_id.code == 'IN' and o.company_id.l10n_in_is_gst_registered and o.partner_shipping_id.vat not in [False, '', '/', 'NA', 'na']">
```
jva-odoo: I don't this is needed.
@ajjo-odoo can you plz confirm that we need to show shipping partner GST number in Invoice PDF.

jva-odoo (commented):
just have question for invoice PDF else LGTM

addons/l10n_in/models/l10n_in_pan_entity.py
jco-odoo: Indeed, I was thinking that those would automatically have a GST and that is not necessarily the case.

addons/l10n_in/views/report_invoice.xml
hamo-odoo: I remember discussing this with @ajjo-odoo in the past, and he agreed upon it. Better to reconfirm

addons/l10n_in/models/company.py
```diff
@@ -125,7 +125,7 @@ def _activate_l10n_in_taxes(self, group_refs, company):
     @api.depends('vat')
     def _compute_l10n_in_hsn_code_digit(self):
         for record in self:
-            if record.country_code == "IN" and record.vat:
+            if record.country_code == "IN" and record.vat not in [False, '', '/', 'NA', 'na']:
                 record.l10n_in_hsn_code_digit = "4"
             else:
                 record.l10n_in_hsn_code_digit = False
```
hamo-odoo, line 144: @jva-odoo @ajjo-odoo don't you think, we should not override the hsn_code_digit, if someone has already configured it?

odoo/addons/base/models/res_partner.py
```diff
@@ -914,7 +914,12 @@ def create(self, vals_list):
             partner._fields_sync(vals)
         return partners
 
-    def _is_vat_void(self, vat):
+    def _is_vat_void(self, vat=None):
+        if not self and vat is None:
+            raise TypeError("The 'vat' argument must be provided.")
+        if self and vat is None:
+            self.ensure_one()
+            vat = self.vat
```
jva-odoo:
```suggestion
    def is_vat_void(self):
        self.ensure_one()
        return self._is_vat_void(self.vat)

    def _is_vat_void(self, vat):
```

addons/l10n_in/views/report_invoice.xml
ajjo-odoo: > I don't this is needed. @ajjo-odoo can you plz confirm that we need to show shipping partner GST number in Invoice PDF.

It's kinda 'nice to have' but **not mandatory** by Rule 46 of CGST Act.
We also one task for the same, Print Invoice as Per GST requirement.
https://link.excalidraw.com/l/65VNwvy7c4X/AK8Jryi4w96

image

addons/l10n_in/models/company.py
ajjo-odoo: > @jva-odoo @ajjo-odoo don't you think, we should not override the hsn_code_digit, if someone has already configured it?

let's make this possible with another task with "disable gst" and inactive all warning validation belongs to gst

odoo/addons/base/models/res_partner.py
rco-odoo: Splitting the method makes sense if you expect overrides of the internal method.  In such a case, the first method does the plumbing to get the argument, and the second one does the real thing.  However, in this case, splitting it looks like applying "Clean Code" principles to the letter with no actual benefit (not even in terms of API).

rco-odoo: Note that this is a stable branch.  Don't add a new public method unless there is a good reason for it.  (Public methods can be part of RPC attacks, so every new method is a new risk.)  @hamo-odoo your proposal is fine to me.

jco-odoo: @rco-odoo The point is that doing a method that either takes the parameter or the field on the partner is bringing unncessary confusion.  The inner method, we probably need it only once elsewhere and maybe we can better redo the logic but put a global variable with the list of  FALSY_VATS.  

And these refactoring changes should indeed be done in master and not in saas-19.1.

hamo-odoo:
I have changed the target to master for a safer end and to move this PR forward. @jco-odoo

odoo/addons/base/models/res_partner.py
```diff
@@ -948,7 +948,12 @@ def create(self, vals_list):
             partner._fields_sync(vals)
         return partners
 
-    def _is_vat_void(self, vat):
+    def _is_vat_void(self, vat=None):
+        if not self and vat is None:
+            raise TypeError("The 'vat' argument must be provided.")
```
jco-odoo:
```suggestion
    def _is_vat_void(self):
        self.ensure_one()
```

hamo-odoo: @jco-odoo I still like to go with my current proposal because we do still have instances where the object is not necessary
https://github.com/odoo/odoo/blob/e967ad2f7876ecc020e98a63d00dadcac81c92b1/addons/base_vat/models/res_partner.py#L106-L111

https://github.com/odoo/odoo/blob/e967ad2f7876ecc020e98a63d00dadcac81c92b1/addons/l10n_pl/models/bank_account_verification.py#L157

`partner_vat` is a stored computed field

kmagusiak: In any case, this is an API change and should not be done in stable.

hamo-odoo: @kmagusiak, the PR is targeting master (though previously it was targeting stable, but it didn't seem a good decision anyway)

But still, I prefer to keep the method callable with and without the object because it is more handy and likely to change one method in case of a major change of Falsy vat values rather than checking and changing all checkable condition where the `vat` doesn't seems to have a partner as I have referenced above

kmagusiak: Sorry, I saw the 19.0 in the branch name.

jco-odoo:
@hamo-odoo I asked someone else in the team to see if I was not crazy and he said: "It is better to replicate the logic than to hide logic and split it there inside the method while it has 2 behaviours. "  I don't know why you are trying to go around.

hamo-odoo:
> @hamo-odoo I asked someone else in the team to see if I was not crazy and he said: "It is better to replicate the logic than to hide logic and split it there inside the method while it has 2 behaviours. " I don't know why you are trying to go around.

@jco-odoo But then there is also this opinion https://github.com/odoo/odoo/pull/248132/changes#r2960252194 🤷🏻‍♂️ 

It's not always necessary to apply the "Clean Code", if there is no specific reason for it. And obvious developer will be more confused about which method to call out of the two. Keeping one is most certain. Maybe it's just a difference of POV, but I'm not convinced by the comment until there is a real advantage to it

jco-odoo:
@hamo-odoo Actually, there is an even better way. We could add the method with param in tools/business_data.py in base and then add a method on partner as well calling that method. @jva-odoo was looking in the right direction, except that it should not be public.

odoo/tools/business_data.py
```diff
@@ -7,6 +7,11 @@
 _REGIONAL_INDICATOR_OFFSET = ord('\N{Regional Indicator Symbol Letter A}') - ord('A')
 
 
+def is_vat_void(vat: str) -> bool:
+    if not vat:
+        return True
+    return vat in ['/', 'na', 'NA']
```
jco-odoo:
```suggestion
    return not vat or vat in ['/', 'na', 'NA']
```

odoo/addons/base/models/res_partner.py
```diff
@@ -948,10 +948,9 @@ def create(self, vals_list):
             partner._fields_sync(vals)
         return partners
 
-    def _is_vat_void(self, vat):
-        if not vat:
-            return True
-        return vat in ['/', 'na', 'NA']
+    def _is_vat_void(self):
```
kmagusiak: I think that this function makes no sense and should be removed because void means empty. Don't just move the definition to tools.  Moreover there is odoo/tools/partner_identifiers that contains `is_identifier_void` - which is wrong for the same reasons as above. The function `split_vat` should be moved to partner_identifiers as well.

"/" or "na" or "NA" are probably used for UI, because they make no sense from a modelling perspective. First of all, why not add "n/a" or "not applicable" or any other variant in any language. Then the help attribute of `res.partner.vat` indicates that "use '/' to indicate that customer has no Tax ID", no other values. In any case, that makes no sense: a partner that has no tax ID is `not partner.vat`.

You could make a boolean field that checks if the vat field has a correct format without validating it using VIES or more complex country-dependent methods. You could even show a message to the user when they are editing the vat, that it is an incorrect value even before saving.

addons/l10n_in/models/l10n_in_pan_entity.py
```diff
@@ -27,7 +27,7 @@ class L10nInPanEntity(models.Model):
...
```
kmagusiak: You cannot maintain this or other such places. See my other comment and make a boolean field if you need such things.

kmagusiak:
> I know it's stupid but the orm.....
> 
> ```python
> In [1]: env.company.vat
> Out[1]: False
> 
> In [2]: env.company.vat = ''
> 
> In [3]: env.company.vat
> Out[3]: ''
> ```

There are people relying on this odd behaviour which I tried unsuccessfully to remove in #172329. Maybe I'll rebase it someday. Note that the domain ('vat', '=', False) matches all falsy values.

odoo/addons/base/models/res_partner.py
hamo-odoo: Sorry for the delay, but I had discussed this issue with the PO; they have a different opinion. See https://discord.com/channels/678381219515465750/1099994955830796348/1527620983136190515

But yes can move it to `odoo/tools/partner_identifiers` but for compute boolean, not sure if @jco-odoo agrees with that? But yes, it would be a helpful field for domains. @william-andre what do you think?

kmagusiak: It's still a display problem, but if you ask most people to write a domain to find "without VAT", they will put "VAT is not set". If you want a value that says, "we know there is no VAT", choose 1, not 3 different values. Having a boolean field which is computed can be a visual cue as well.

william-andre: How would the boolean field work?
A field that is invisible if `vat` is set, but that you can check when it is not set in order to say that we know there is nothing to set? Wouldn't that clutter the UI for no reason?

It does make sense to put `/` in something we don't want to fill. It has been the way of filling forms since forms have been invented. Maybe not the cleanest from a relational database schema perspective, but I don't think our clients care about that.

However, I agree that `na`, etc are getting out of hands.

kmagusiak: > It does make sense to put / in something we don't want to fill. It has been the way of filling forms since forms have been invented. 

This is what happens when you have a required field. Depending on where you come from, but I saw "/", "-", "non", "na" and tens of other variants.
Usually, the best thing from data quality point of view is to validate at least the format when saving. Meaning you don't have a VAT number, don't put a value. I am not against having a value for people non-subject to VAT - but one value and not many.

The boolean field does not have to be displayed. It's mainly to indicate that business rule and simplify searching. See l10n_in domain which had `('vat', '=', False)` and tries to hardcode more stuff now.

william-andre: Ok, I get your point and I agree.

A new field `is_vat_void` on `res.partner`, that we can use to replace domains like `('vat', '=', False)` by `('is_vat_void', '=', True)`, and same in the business code everywhere `_is_vat_void` is called.

For places where we do the check on something else than a `res.partner`, we can also do something like `env['res.partner'].new({'vat': vat}).is_vat_void` if we want to avoid adding a function in the tooling.

kmagusiak: Better to use positives to be less confusing: `has_vat` instead of `is_vat_void`.

hamo-odoo: Alas we have a conclusion. Thanks for your input @william-andre I was a bit confused about this. But just need to confirm whether we are keeping the `na`, etc. values, or just `/`? though doesn't make sense to me as well, but as per PO questionnaire thread, they seem to keep the values for `na`. Let me know

kmagusiak: `na` is language-specific, please choose something that is not.

william-andre: That would require an upgrade script btw, but don't change the behavior without PO input

hamo-odoo:
@william-andre @kmagusiak Ping for a re-review. I have made the changes

odoo/addons/base/models/res_partner.py
```diff
@@ -921,7 +928,16 @@ def _clean_website(self, website):
...
 
-    @api.depends('vat', 'commercial_partner_id')
+    @api.depends('vat')
+    def _compute_has_vat(self):
+        for partner in self:
+            vat = partner.vat
+            partner.has_vat = vat and vat not in ('/', 'na', 'NA')
+
+    def _compute_sql_has_vat(self, table):
+        return SQL("( %s IS NOT NULL AND %s NOT IN ('/', 'na', 'NA', ''))", table.vat, table.vat)
```
kmagusiak: Nitpick: for any condition to be true, the value must be not null.
```suggestion
        return SQL("( %s NOT IN ('/', 'na', 'NA', ''))", table.vat)
```

kmagusiak (commented):
Much better. I still don't like the fact of having multiple values as exclusions.
And you can see from the fixed code that if you say "give me partners without VAT", people will write `not partner.vat`. It's not to be pedantic about modelling, but you will have code like that reappear after some time.

kmagusiak (approved):

odoo/addons/base/models/res_partner.py
hamo-odoo: I have made the changes, but still I don't get it how the value must be not null in this case for _compute_sql I don't see ORM handling it?
```py
In [11]: env['res.partner'].search([('vat', '=', False)])
2026-09-01 12:14:35,969 289755 DEBUG master odoo.sql_db: [0.440 ms] query: SELECT "res_partner"."id" FROM "res_partner" WHERE ("res_partner"."active" IS TRUE AND ("res_partner"."vat" IN ('') OR "res_partner"."vat" IS NULL)) ORDER BY "res_partner"."complete_name" ASC , "res_partner"."id" DESC 
Out[11]: res.partner(8, 14, 44, 6, 3, 7, 43, 13, 1, 41, 42)

In [12]: env['res.partner'].search([('has_vat', '=', False)])
2026-09-01 12:14:42,401 289755 DEBUG master odoo.sql_db: [0.472 ms] query: SELECT "res_partner"."id" FROM "res_partner" WHERE ("res_partner"."active" IS TRUE AND "res_partner"."vat" NOT IN ('/', 'na', 'NA', '') IS NOT TRUE) ORDER BY "res_partner"."complete_name" ASC , "res_partner"."id" DESC 
Out[12]: res.partner(8, 14, 44, 6, 3, 7, 43, 13, 1, 41, 42)
```
Why not add the NOT NULL?

kmagusiak: SQL: `select c from (select 1 as c union all select null) where c not in (2);`
`c not in ...` already means c is not null.

odoo/addons/base/models/res_partner.py
```diff
@@ -317,6 +317,14 @@ def default_get(self, fields):
         readonly=False, store=True,
         help='The internal user in charge of this contact.')
     vat = fields.Char(string='Tax ID', index=True, inverse='_inverse_vat', help="You can use '/' to indicate that the customer has no Tax ID.")
```
william-andre: Not related directly, but does it make sense to index `NULL` and falsy values (including `/`, `na`, etc)?

For instance on our prod, 20M out of the 21M partners don't have a vat set (NULL)

hamo-odoo: Doesn't makes sense to me as well, but we can add `btree_not_null`? Not sure how to handle Falsy values. @kmagusiak might know better?

kmagusiak: Probably not as they are not very selective. Does it really hurt since the field is filled most of the time anyways?
As all perf questions, it depends on user data :)

william-andre: > since the field is filled most of the time anyways

As soon as there is a website, I expect most records to have an empty `vat`

hamo-odoo: If we consider POS and E-commerce users, there will be chances of B2C more in such a case but does if hurt to add btree_not_null ?

william-andre: I believe that it hurts even less now that we are using `has_tax`, which is kind of a barrier to using the index AFAIK

william-andre (approved):
@robodoo r+

william-andre:
@robodoo rebase-ff

PR #248135, opened 2026-02-11
[FIX] base: remove aggregation in device models

thle-odoo:
Invariants:
- For a given session identifier, the user will always be the same (otherwise the session is no longer valid).
- For a given device determined by a session identifier, an IP address, and a user agent, the country and city will remain the same, as these are determined when the device is first detected.

It is therefore possible to no longer aggregate these values and place them directly in the `GROUP BY` clause in order to improve the performance[^1] of `res.device` and `res.session` models.

[^1]: Aggregate computations (`MIN`/`MAX`) force the database to perform redundant comparisons for every row in the group. By placing these columns directly in the `GROUP BY` clause, the query reduces per-row processing. This lead to a more efficient execution plan (especially on large datasets).

Task-5928301

### DIFF
--- a/odoo/addons/base/models/res_device.py
+++ b/odoo/addons/base/models/res_device.py
@@ -187,11 +187,11 @@ class ResDevice(models.Model):
             SELECT
                 MAX(L.id) as id,
                 L.session_identifier as session_identifier,
-                MIN(L.user_id) as user_id,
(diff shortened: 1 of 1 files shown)


kmagusiak (approved):
@robodoo r+

PR #248138, opened 2026-02-11
[FIX] l10n_sa_edi: correct visibility of additional identification fields

hash-odoo:
This commit fixes the visibility logic of `l10n_sa_edi_additional_identification_*` fields.

- `l10n_sa_edi_additional_identification_scheme` is now visible only when both the contact and the current company is Saudi.
- `l10n_sa_edi_additional_identification_number` is visible for all contacts, but only when the current company is Saudi.

This ensures the fields follow the intended localization scope and prevents them from appearing in non-Saudi company contexts.

task-5881754

### DIFF
--- a/addons/l10n_sa_edi/models/res_partner.py
+++ b/addons/l10n_sa_edi/models/res_partner.py
@@ -22,6 +22,7 @@ class ResPartner(models.Model):
     ], default="OTH", string="Identification Scheme", help="Additional Identification scheme for Seller/Buyer")
 
     l10n_sa_edi_additional_identification_number = fields.Char("Identification Number (SA)", help="Additional Identification Number for Seller/Buyer")
+    l10n_sa_edi_additional_identification_visibility = fields.Boolean(compute='_compute_l10n_sa_edi_additional_identification_visibility')
 
     @api.model
     def _commercial_fields(self):
@@ -33,3 +34,7 @@ class ResPartner(models.Model):
     def _address_fields(self):
         return super()._address_fields() + ['l10n_sa_edi_building_number',
                                             'l10n_sa_edi_plot_identification']
+
+    @api.depends_context('company')
+    def _compute_l10n_sa_edi_additional_identification_visibility(self):
+        self.l10n_sa_edi_additional_identification_visibility = self.env.company.country_code == 'SA'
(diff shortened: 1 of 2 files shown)


addons/l10n_sa_edi/models/res_partner.py
```diff
@@ -33,3 +34,7 @@ def _commercial_fields(self):
     def _address_fields(self):
         return super()._address_fields() + ['l10n_sa_edi_building_number',
                                             'l10n_sa_edi_plot_identification']
+
+    @api.depends_context('company')
+    def _compute_l10n_sa_edi_additional_identification_visibility(self):
+        self.l10n_sa_edi_additional_identification_visibility = self.env.company.country_code == 'SA'
```
hamo-odoo, line 40: Not sure if this makes sense? Even if it does, then should it be done in `account`?

jco-odoo: I think what was there before was good or almost good.  No need to add compute fields...

addons/l10n_sa_edi/views/res_partner_views.xml
```diff
@@ -8,8 +8,8 @@
             <field name="inherit_id" ref="base.view_partner_form"/>
             <field name="arch" type="xml">
                 <xpath expr="//div[hasclass('o_address_format')]"  position="after">
-                    <field name="l10n_sa_edi_additional_identification_scheme" invisible="'SA' not in fiscal_country_codes or country_code != 'SA'"/>
-                    <field name="l10n_sa_edi_additional_identification_number" invisible="'SA' not in fiscal_country_codes and country_code != 'SA' or l10n_sa_edi_additional_identification_scheme == 'TIN'"/>
```
jco-odoo, line 12: I wonder if it is not just the and that needs to change in or to be consistent with the field above

addons/l10n_sa_edi/models/res_partner.py
hamo-odoo: I have the same opinion, and I don't see anywhere in the l10ns we have done something like this

cc- @ygu-odoo

hash-odoo:
Task Cancelled

PR #248142, opened 2026-02-11
[FIX] pos_sale: do not change pos.order status

aksi-odoo:
Step to reproduce:
- install spreadsheet_dashboard_sale and pos_sale
- create a order in pos , invoice it too
- open sale dashboard

Observation:
- the order fulfilled in pos, does not reflect in dashboard

Cause:
- sale has 4 status i.e ["draft", "sent", "sale", "cancel"]
- when pos_sale is installed, new status oders are added i.e ['paid', 'invoiced', 'done']
- sale dashboard pivot relies on sale defined status only, which so not consider orders that have status in  ['paid', 'invoiced', 'done']

Fix:
- fix the domain of pivots such that, it will now accept other orders too

**Before:**

image

image
<br/>

**After**
<br/>

image

image

opw-5487654

---

### DIFF
--- a/addons/pos_sale/report/sale_report.py
+++ b/addons/pos_sale/report/sale_report.py
@@ -57,7 +57,7 @@ class SaleReport(models.Model):
             count(*) AS nbr,
             pos.name AS name,
             pos.date_order AS date,
-            (CASE WHEN pos.state = 'done' THEN 'sale' ELSE pos.state END) AS state,
+            pos.state AS state,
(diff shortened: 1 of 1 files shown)


alsh-odoo (approved):
Not an expert in this part, so I’ll defer the detailed review there to other reviewers.
Tested functionally and it does the job 👍

ushyme (approved):
LGTM 👍🏾 , thanks for your work !

PR #248149, opened 2026-02-11
[FIX] l10n_it_edi: fix import of cross-company products

LouisGobert:
In a multi-company environment, importing an EDI invoice could fail if a line contains a product restricted to another company. The error "Incompatible company on records" was raised because company consistency rules were applied.

Example steps to reproduce:
- Create two Italian companies, A and B
- In company A, create a product with default code or barcode "TEST"
    - This can also occur using the barcode and the “Vendor Product code” from the Purchase supplier.
- Restrict this product to company A
- Create an EDI invoice in company A with this product for company B as customer
- Import the invoice in company B, which would fail

This fix adds a search domain to correctly find products during import

opw-5496912

### DIFF
--- a/addons/l10n_it_edi/tests/test_edi_import.py
+++ b/addons/l10n_it_edi/tests/test_edi_import.py
@@ -325,3 +325,78 @@ class TestItEdiImport(TestItEdi):
         user = new_test_user(self.env, login='jag', groups='account.group_account_invoice')
         move = self.env['account.move'].create({'move_type': 'in_invoice'})
         move.with_user(user).read(['l10n_it_edi_is_self_invoice'])  # should not raise
+
+    def test_cron_import_bill_from_another_company_without_conflicts(self):
+        """
+        Ensure that in a multi-company environment, importing a bill containing products
+        restricted to another company does not fail due to company inconsistencies.
+        """
+        test_product = self.env['product.product'].create({
+            'name': 'Test Product',
+            'default_code': 'TEST',
+            'barcode': 'TEST',
+            'standard_price': 75.0,
+            'company_id': self.company_data['company'].id,
+        })
+        self.env['product.supplierinfo'].create({
+            'product_id': test_product.id,
+            'product_code': 'TEST',
+            'partner_id': self.company_data_2["company"].partner_id.id,
+        })
+
+        applied_xml = """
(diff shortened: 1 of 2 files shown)


addons/l10n_it_edi/tests/test_edi_import.py
```diff
@@ -325,3 +325,78 @@ def test_invoice_user_can_compute_is_self_invoice(self):
...
+                        <CodiceValore>TEST</CodiceValore>
+                    </CodiceArticolo>
+                    <Descrizione>[TEST] Test Product</Descrizione>
+                    <Quantita>1.00</Quantita>
+                    <PrezzoUnitario>0.00</PrezzoUnitario>
+                    <PrezzoTotale>0.00</PrezzoTotale>
+                    <AliquotaIVA>22.00</AliquotaIVA>
+                </DettaglioLinee>
+                <DettaglioLinee>
+                    <NumeroLinea>2</NumeroLinea>
```
lordkrandel:
```suggestion
                    <NumeroLinea>3</NumeroLinea>
```
it's called `LineNumber` 🤣

addons/l10n_it_edi/tests/test_edi_import.py
```diff
@@ -325,3 +325,78 @@ def test_invoice_user_can_compute_is_self_invoice(self):
         user = new_test_user(self.env, login='jag', groups='account.group_account_invoice')
         move = self.env['account.move'].create({'move_type': 'in_invoice'})
         move.with_user(user).read(['l10n_it_edi_is_self_invoice'])  # should not raise
+
+    def test_cron_import_bill_from_another_company_without_conflicts(self):
+        """
```
lordkrandel:
```suggestion
        """
```

lordkrandel (changes requested):
Some nitpicks (really). 
- The trailing whitespace is CI/style blocking.
- the other - you never know if people will copypaste those records, better put things in order.
- I don't know about the test prices being 0.0 with a 22% VAT tax 🤷🏻 
  Can't you put like, 4.0?

Rest is neat, very well written, consistent with the testing style, clean.
Thanks for the PR!

lordkrandel:
@robodoo r+

🤝

lordkrandel:
@robodoo r+

PR #248154, opened 2026-02-11
[FIX] base: add hard gc for `res.device.log`

thle-odoo:
The res.device.log table contains a huge amount of data. It makes sense to provide a way for administrators to garbage collect these logs according to their retention policy.

The `ir.config_parameter` to determine the retention policy is `base.res_device_log_retention_days`.

With the hard GC configured, a log will be deleted if its last activity has exceeded the number of days specified in the system parameter. This can remove all traces of a session identifier. The parameter must be chosen judiciously according to the database backups.

Note:
Put more information in the logs, as this data may be deleted.

Task-5928301

### DIFF
--- a/odoo/addons/base/models/res_device.py
+++ b/odoo/addons/base/models/res_device.py
@@ -63,27 +63,29 @@ class ResDeviceLog(models.Model):
         else:
             cursor = nullcontext(self.env.cr)
         with cursor as cr:
+            info = {
+                'session_identifier': session_identifier,
+                'user_id': user_id,
+                'ip_address': device['ip_address'],
+                'user_agent': device['user_agent'][:1024],  # Truncate very long user-agent
+                'country': device.get('country'),  # TODO (v20): remove backward compatibility by forcing the key
+                'city': device.get('city'),  # TODO (v20): remove backward compatibility by forcing the key
+                'first_activity': datetime.fromtimestamp(device['first_activity']),
+                'last_activity': datetime.fromtimestamp(device['last_activity']),
+                'revoked': False,
+            }
             cr.execute(SQL("""
                 INSERT INTO res_device_log (session_identifier, user_id, ip_address, user_agent, country, city, first_activity, last_activity, revoked)
                 VALUES (%(session_identifier)s, %(user_id)s, %(ip_address)s, %(user_agent)s, %(country)s, %(city)s, %(first_activity)s, %(last_activity)s, %(revoked)s)
-            """,
-                session_identifier=session_identifier,
-                user_id=user_id,
-                ip_address=device['ip_address'],
-                user_agent=device['user_agent'][:1024],  # Truncate very long user-agent
-                country=device.get('country'),  # TODO (v20): remove backward compatibility by forcing the key
-                city=device.get('city'),  # TODO (v20): remove backward compatibility by forcing the key
-                first_activity=datetime.fromtimestamp(device['first_activity']),
-                last_activity=datetime.fromtimestamp(device['last_activity']),
-                revoked=False,
-            ))
-        _logger.info('User %d inserts device log (%s)', user_id, session_identifier)
+            """, **info))
+        _logger.info('User %(user_id)d inserts device log: %(session_identifier)s - %(ip_address)s - %(user_agent)s', info)
 
     @api.autovacuum
     def _gc_device_log(self):
-        # Keep the last device log
-        # (even if the session file no longer exists on the filesystem)
-        self.env.cr.execute("""
+        # Soft GC:
+        # Keep the last device log even if the session file
+        # no longer exists on the filesystem
+        query = SQL("""
             DELETE FROM res_device_log log1
             WHERE EXISTS (
                 SELECT 1 FROM res_device_log log2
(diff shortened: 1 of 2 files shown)


odoo/addons/test_http/tests/test_device.py
```diff
@@ -423,6 +430,57 @@ def test_filesystem_reflexion(self):
...
+        ICP.set_int('base.res_device_log_retention_days', 10)
+
+        with freeze_time(now(days=11)):
+            self.DeviceLog._gc_device_log()
+        *_, logs = self.get_devices(self.user_admin)
+        self.assertEqual(len(logs), 1, 'Because the session still exists on the filesystem')
+
+        # Revoke the session and mark it revoked in database
+        odoo.http.root.session_store.store.clear()
+        with freeze_time(now(days=(SESSION_LIFETIME / 86400) + 1)), patch.object(self.cr, 'commit', lambda: ...):
```
kmagusiak: Commits are forbidden when testing. Why do you need this?

odoo/addons/test_http/tests/test_device.py
```diff
@@ -423,6 +430,57 @@ def test_filesystem_reflexion(self):
...
+            self.DeviceLog.sudo()._ResDeviceLog__update_revoked()
+        self.DeviceLog.flush_model()
+        self.Session.invalidate_model()
```
kmagusiak: Nitpick: After a freeze_time, you could just do `self.env.invalidate_all()`.

odoo/addons/base/models/res_device.py
```diff
@@ -93,7 +94,23 @@ def _gc_device_log(self):
...
-        """)
+        """
+        params = ()
+
+        # Hard GC:
+        # Delete device logs if the last activity has been exceeded by a defined
+        # number of days (ensure the session file no longer exists on the filesystem)
+        if retention_days := self.env['ir.config_parameter'].get_int('base.res_device_log_retention_days'):
+            max_last_activity = self.env.cr.now() - timedelta(days=retention_days)
+            query += """
```
kmagusiak: You should use `SQL` object instead of concat.

kmagusiak (changes requested):

odoo/addons/test_http/tests/test_device.py
thle-odoo: This is necessary because in this context manager, the `__update_revoked` method is called. This method is an `autovacuum` that performs batch "cleaning" with a commit to ensure progress continues if it takes too long. As you mention, since commits are prohibited in tests, I make sure that it does nothing.

Note:
This is already done in the `test_filesystem_reflexion` test.

odoo/addons/test_http/tests/test_device.py
thle-odoo: Yes, but that was to be explicit about the models that must be invalidated.

(I think I will leave it as it is if that is okay, to be consistent with the other tests.)

After thinking about it, I changed it.

odoo/addons/test_http/tests/test_device.py
kmagusiak: Instead, you should enter registry test mode the code may commit.

Julien00859: goddamnit
image

odoo/addons/test_http/tests/test_device.py
```diff
@@ -21,6 +23,11 @@
 @tagged('at_install', '-post_install')  # LEGACY at_install
 class TestDevice(TestHttpBase):
 
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.classPatch(BaseCursor, 'now', lambda cr: datetime.now())
```
kmagusiak: This should not be here and is not correct. When we create a cursor, we set the current date on it (for test cursors).

odoo/addons/test_http/tests/test_device.py
kmagusiak: Maybe you wanted to do:
```py
with registry.cursor() as cr:
   self.DeviceLog.with_env(cr=cr)....
```

odoo/addons/test_http/tests/test_device.py
```diff
@@ -415,14 +423,71 @@ def test_filesystem_reflexion(self):
         odoo.http.root.session_store.store.clear()
 
         # Update all device logs
-        with freeze_time('2025-02-01 08:00:00'), patch.object(self.cr, 'commit', lambda: ...):
-            self.DeviceLog.sudo()._ResDeviceLog__update_revoked()
-        self.DeviceLog.flush_model()  # Because write on ``res.device.log`` and so we have new values in cache
-        self.Session.invalidate_model()  # Because it depends on the ``res.device.log`` model (updated in database)
+        with freeze_time('2025-02-01 08:00:00'):
```
kmagusiak: The time of the cursor is frozen during its creation, so you should create the cursor after freeze_time.

kmagusiak (approved):

beledouxdenis (approved):

kmagusiak (approved):
@robodoo r+

PR #248158, opened 2026-02-11
[FIX] web: fix unscheduled event text truncation

juliusc2066:
Previously, the `text-truncate` class was applied only to the text span of unscheduled events in the calendar side panel. This failed to account for the width of the drag handle, potentially causing layout issues.

This commit moves the truncation class to the parent container, ensuring the text is properly truncated while respecting the space occupied by the drag handle.


aab-odoo (approved):
robodoo r+

PR #248166, opened 2026-02-11
[FIX] web: image_field: avoid some nondeterministic error

rfr-odoo:
Encoding the data to base64 can take some times.
Before this commit we used this code:
```js
await runAllTimers();
await animationFrame();
```
Now we explicitly wait for the change to happen as awaiting an animation frame can't be enough.

runbot-error-237568

---


aab-odoo (approved):
robodoo r+

PR #248167, opened 2026-02-11
[FIX] project: ensure archived subtasks are not copied during task duplication

utma-odoo:
Currently, when duplicating a task that contains `archived subtasks`, 
the archived subtasks are also duplicated.

**Steps to reproduce:**
- Install the `project` module.
- Open any `project` and create a task with a subtask.
- `Archive` the subtask.
- `Duplicate` the parent task.

**Observation:**
The duplicated task contains a copy of the archived subtask, 
even though it is inactive.

**Root Cause:**
At [1], subtasks are duplicated without checking their active status. As a 
result, archived (`active=False`) subtasks are also copied during duplication.

**Fix:**
This commit ensures that archived subtasks are not copied when duplicating a task.

[1]:
https://github.com/odoo/odoo/blob/531b887aec92c2fbf57495992be9fbc32d9ea20e/addons/project/models/project_task.py#L822

opw-5926009

### DIFF
--- a/addons/project/models/project_task.py
+++ b/addons/project/models/project_task.py
@@ -819,7 +819,7 @@ class Task(models.Model):
                     'dependent_ids': False,
                     'parent_id': False,
                 }
-                vals['child_ids'] = [Command.create(child_id.copy_data(default)[0]) for child_id in task.child_ids]
+                vals['child_ids'] = [Command.create(child_id.copy_data(default)[0]) for child_id in task.child_ids.filtered(lambda c: c.active)]
         return vals_list
 
     def _create_task_mapping(self, copied_tasks):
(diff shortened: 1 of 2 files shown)


alsh-odoo (approved):

abla001 (approved):
LGTM

utma-odoo:
Hello @xavierbol :wave:,

>That's strange, normally since active_test is not in the context, child_ids should only have active subtasks. 🤔

When duplicating a task, the `copy()` method is called at [1], 
where `active_test` is not in the context.

However, in the base implementation at [2], `copy()` internally calls:
`vals_list = self.with_context(active_test=False).copy_data(default)`

This forces `active_test=False` during `copy_data()`, which causes `child_ids` to include 
`archived (active=False)` subtasks.

[1]:
https://github.com/odoo/odoo/blob/d2966d5f3624a64bf9740126b2854628b204d2a9/addons/project/models/project_task.py#L849-L875

[2]:
https://github.com/odoo/odoo/blob/d2966d5f3624a64bf9740126b2854628b204d2a9/odoo/models.py#L5961-L5976

xavierbol:
Okay, I missed that, thanks. 🙂

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #248168, opened 2026-02-11
[IMP] mail, *: remove im_status from partner

Atovange:
Enterprise PR: https://github.com/odoo/enterprise/pull/107104
Upgrade PR: https://github.com/odoo/upgrade/pull/9861

* = hr, hr_{holidays,homeworking}, im_livechat, project

This commit removes the im_status field from python's partner model. The concept of IM status belongs logically to the user model since it's about users's online/offline presence.
This commit also removes the "bot" IM status.


addons/im_livechat/tests/test_chatbot_internals.py
```diff
@@ -317,16 +317,17 @@ def notifications():
                     {
                         "res.partner": self._filter_partners_fields(
                             {
+                                "agent_ids": [],
                                 "email": "e.e@example.com",
                                 "id": self.partner_employee.id,
-                                "im_status": "offline",
-                                "im_status_access_token": self.partner_employee._get_im_status_access_token(),
-                                "main_user_id": self.user_employee.id,
+                                "user_ids": [self.user_employee.id],
```
seb-odoo: Small simplification
```suggestion
                                "user_ids": self.user_employee.ids,
```
same in all places

addons/mail/models/ir_websocket.py
```diff
@@ -86,19 +86,13 @@ def _prepare_subscribe_data(self, channels, last):
...
         # There is a gap between a subscription client side (which is debounced)
         # and the actual subcription thus presences can be missed. Send a
         # notification to avoid missing presences during a subscription.
         presence_domain = Domain("last_poll", ">", datetime.now() - timedelta(seconds=2)) & (
-            Domain(
-                "user_id",
-                "in",
-                allowed_partners.with_context(active_test=False).sudo().user_ids.ids,
-            )
+            Domain("user_id", "in", allowed_users.with_context(active_test=False).ids)
```
seb-odoo: `active_test` is useless when reading `ids`, it was there for reading `user_ids`

addons/mail/models/mail_presence.py
```diff
@@ -102,17 +102,9 @@ def _send_presence(self, im_status=None, bus_target=None):
         for presence in self:
             persona = presence.guest_id or presence.user_id
             target = bus_target or (persona, "presence")
-            stores[target].add(
-                presence.guest_id or presence.user_id.partner_id,
-                {"im_status": im_status or persona.im_status},
-            )
+            stores[target].add(persona, {"im_status": im_status or persona.im_status})
```
seb-odoo, line 105: Can't we remove the `im_status` param yet in master?

addons/mail/models/res_partner.py
```diff
@@ -252,9 +213,8 @@ def _store_avatar_fields(self, res: Store.FieldList):
         res.attr("write_date")
 
     def _store_im_status_fields(self, res: Store.FieldList):
-        res.attr("im_status")
-        res.attr("im_status_access_token", lambda p: p._get_im_status_access_token())
-        res.one("main_user_id", "_store_im_status_fields")
+        # sudo: res.users - can access IM status of accessible users
```
seb-odoo: accessible _partners_ in this case

addons/mail/static/src/core/common/im_status_dropdown.js
```diff
@@ -23,12 +23,12 @@ export class ImStatusDropdown extends Component {
...
 
     setManualImStatus(status) {
-        this.store.self.forceImStatus(status);
+        this.store.self_user.forceImStatus(status);
         rpc("/mail/set_manual_im_status", { status });
     }
 
     get readableImStatus() {
-        const imStatus = this.store.self.imStatusUI || "offline";
+        const imStatus = this.store.self_user.imStatusUI || "offline";
```
seb-odoo, line 31: This will exclude guests?

addons/mail/static/src/core/common/im_status_dropdown.xml
```diff
@@ -6,14 +6,14 @@
                 <ImStatus user="this.store.self_user"/> <span t-out="this.readableImStatus"/> <i t-if="this.env.isSmall" class="oi oi-chevron-right ms-auto mt-1"/>
             </a>
             <t t-set-slot="content">
-                <DropdownItem class="{ 'active': this.store.self.imStatusUI === 'online' }" onSelected="() => this.setManualImStatus('online')"><i class="fa fa-circle text-success me-1" title="Online" role="img"/> Online</DropdownItem>
+                <DropdownItem class="{ 'active': this.store.self_user.imStatusUI === 'online' }" onSelected="() => this.setManualImStatus('online')"><i class="fa fa-circle text-success me-1" title="Online" role="img"/> Online</DropdownItem>
```
seb-odoo, line 9: Same here, why exclude guests?

addons/mail/static/src/core/common/res_partner_model.js
```diff
@@ -60,18 +80,16 @@ export class ResPartner extends ImStatusMixin {
         return this.name || this.display_name;
     }
 
-    _computeMonitorPresence() {
-        return (
-            super._computeMonitorPresence() && this.im_status !== "im_partner" && !this.is_public
```
seb-odoo, line 65: Check on `is_public` is removed on this PR?

addons/mail/static/src/core/web/avatar_card/avatar_card.xml
```diff
@@ -6,7 +6,7 @@
                 <div class="d-flex align-items-start gap-2 bg-inherit">
                     <span t-if="this.displayAvatar" class="o_avatar pt-1 position-relative o_card_avatar bg-inherit">
                         <img t-att-src="this.avatarUrl" class="rounded"/>
-                        <span t-if="this.user and this.partner?.im_status and this.partner.im_status !== 'im_partner'" name="icon" class="o_card_avatar_im_status position-absolute d-flex align-items-center justify-content-center bg-inherit">
+                        <span t-if="this.user?.imStatusUI or this.partner?.imStatusUI" name="icon" class="o_card_avatar_im_status position-absolute d-flex align-items-center justify-content-center bg-inherit">
```
seb-odoo, line 9: If we only pass user and not persona to ImStatus component I don't think the condition on partner is useful.
But maybe we should pass persona too when there is no user?

addons/mail/static/src/core/web/mail_core_web_service.js
```diff
@@ -42,7 +42,7 @@ export class MailCoreWeb {
             /** @type {import("models").Message} */
             const message = this.store["mail.message"].get(messageId);
             this.addMessageToInbox(message, notifId);
-            if (!this.store.self?.im_status?.includes("busy")) {
+            if (this.store.self_user?.im_status !== "busy") {
```
seb-odoo, line 45: Also excluding guests?

addons/mail/static/src/discuss/core/common/channel_member_model.js
```diff
@@ -183,7 +183,7 @@ export class ChannelMember extends Record {
     }
 
     get isOnline() {
-        return this.store.onlineMemberStatuses.includes(this.imStatusUI);
+        return ["away", "busy", "online"].includes(this.imStatusUI);
```
seb-odoo: Won't this make bot appear offline instead of online?

addons/mail/static/src/discuss/core/public_web/discuss_channel_model_patch.js
```diff
@@ -162,7 +162,7 @@ const discussChannelPatch = {
         if (
             (!this.self_member_id?.mute_until_dt ||
                 message.partner_ids?.includes(this.store.self)) &&
-            this.store.self.im_status !== "busy" &&
+            this.store.self_user?.im_status !== "busy" &&
```
seb-odoo, line 165: guest (all places)

addons/mail/models/mail_presence.py
Atovange: No, because it's used by `unlink`.

addons/mail/static/src/core/common/im_status_dropdown.js
Atovange: Guests can't set their IM status manually, so I'd rather use the `im_status` on `store.self` directly rather than go through the partner compute.

addons/mail/static/src/core/common/res_partner_model.js
Atovange: I don't think we ever end up with the public user in the store, do we?

seb-odoo: Nothing prevents it, it's as easy as having public partner as author of a message or as channel member.
This was the case in old live chat channels, and it's still possible to do it manually from the UI.

addons/mail/static/src/core/common/im_status_dropdown.js
seb-odoo: Ok for manual status flows, but it seems the PR changed other unrelated places?

addons/mail/static/src/core/web/avatar_card/avatar_card.xml
Atovange: > But maybe we should pass persona too when there is no user?

Perhaps yes, since the `im_status` info does not come from `main_user_id`, so `imStatusUI` could be set while `main_user_id` is undefined.

addons/mail/static/src/discuss/core/common/channel_member_model.js
Atovange: We now set the `im_status` of `isBot` partners to "online" JS side.

addons/mail/static/src/core/web/avatar_card/avatar_card.xml
seb-odoo: But it's a regression to recent improvements that made it use user instead, which is better since im status _is_ on user.
Persona is only necessary to handle bots?

Atovange: > Persona is only necessary to handle bots?

We want to display the IM status of members without user record in those cases:
1. OdooBot (even though it has a `res.users` record but it's archived, so not "found" by `res.many("user_ids", ...`).
2. AI Agents.
3. Live Chat bots (currently it shows the "?" icon so it should be fixed).
4. WhatsApp users.

addons/mail/static/src/core/common/im_status_dropdown.js
Atovange: The other cases are when checking for `"busy"` status, which guests can not have.

addons/mail/models/mail_message.py
```diff
@@ -1138,11 +1138,7 @@ def _store_message_fields(
         # sudo: mail.message: access to author_id is allowed
         res.one(
             "author_id",
-            lambda res: (
-                res.attr("is_company"),
-                res.one("main_user_id", ["partner_id", "share"]),
-                res.from_method("_store_avatar_fields"),
-            ),
+            "_store_author_fields",
```
seb-odoo: I'd expect `_store_author_fields` to be defined in message not in partner.
(where `self` is message, `res.records` is partner)

seb-odoo (approved):
@robodoo delegate+

addons/mail/tests/discuss/test_rtc.py
```diff
@@ -1217,9 +1217,8 @@ def _res_for_partner(self, partner, common=True, internal=False):
             res["name"] = partner.name
             res["write_date"] = fields.Datetime.to_string(partner.write_date)
         if internal:
-            res["im_status"] = partner.im_status
-            res["im_status_access_token"] = partner._get_im_status_access_token()
-            res["main_user_id"] = partner.main_user_id.id
+            res["agent_ids"] = []
+            res["user_ids"] = [partner.main_user_id.id]
```
seb-odoo:
```suggestion
            res["user_ids"] = partner.main_user_id.ids
```

seb-odoo: or actually:
```suggestion
            res["user_ids"] = partner.user_ids.ids
```

addons/im_livechat/tests/test_message.py
```diff
@@ -194,6 +194,7 @@ def test_store_add_message(self):
                         "avatar_128_access_token": self.users[
                             1
                         ].partner_id._get_avatar_128_access_token(),
+                        "agent_ids": [],
```
seb-odoo: alphabetical: should be before `avatar_128_access_token` in all tests

seb-odoo (commented):
Note: to fix the query count increase in most flows, I think it would be acceptable to return the `agent_ids` field only for messages that belongs to channels.

Atovange:
Everything should be good now, I'm waiting for https://github.com/odoo/odoo/pull/256635 to merge (to remove the "im_partner" in unknownStatusMembers) and the upgrade review.

addons/mail/tests/discuss/test_discuss_channel.py
```diff
@@ -205,17 +205,18 @@ def notifications():
...
-                                "im_status_access_token": self.test_partner._get_im_status_access_token(),
-                                "main_user_id": self.test_user.id,
                                 "tz": False,
+                                "user_ids": self.test_user.ids,
                             },
                         ),
                         "res.users": self._filter_users_fields(
                             {
                                 "employee_ids": [],
+                                "im_status": "offline",
```
seb-odoo, line 219: `im_status` after `id`

Atovange:
@robodoo r+

PR #248180, opened 2026-02-11
[FIX] html_builder: restore image src when discarding image transformation

julien-banken:
When a user crops an image and reset the transformation using the right panel, the system does not restore the original image's src (from `data-image-src`). Instead, it replaces it with a base 64 encoded version of the original image without any transformation.

Later, when the user saves the document, the system converts this base 64 image into a new attachment, even though the original attachment already exists or the image is loaded from an external url.

Steps to reproduce:
1. Open Mass Mailing
2. Drag and drop a snippet containing a croppable image
3. Click on the image
4. Crop the image
5. Click on the "Reset" button from the right panel

=> The image is encoded in base 64 instead of reusing the original image src.

To fix this issue, we will modify the `resetCropAction` so that it restores the image's src (stored in `data-original-src`) instead of calling the method to process the image and convert it to base 64, etc.

Task-6140227

### DIFF
--- a/addons/html_builder/static/src/plugins/image/image_tool_option_plugin.js
+++ b/addons/html_builder/static/src/plugins/image/image_tool_option_plugin.js
@@ -159,6 +159,15 @@ export class ResetCropAction extends BuilderAction {
         const newDataset = Object.fromEntries(
             cropperDataFieldsWithAspectRatio.map((field) => [field, undefined])
         );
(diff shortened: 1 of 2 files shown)


julien-banken:
After investigating with the website team, we cannot simply restore the `data-original-src` attribute. Doing so would discard any filters applied to the image prior to cropping. Canceling the task 😊

PR #248186, opened 2026-02-11
[FIX] point_of_sale: prevent invoice from being credit note with ewallet

artn-odoo:
**Steps to reproduce:**
- Make a $10 product, the tax must be tax included
- Set up an ewallet
- Open the PoS, order the created product and add the ewallet
- Ask for an invoice and complete the payment
- The generated invoice is a Credit Note

**Why the fix:**
When chosing the invoice's move_type (invoice or credit note), we check if the total amount is greater or equal to 0, and if it is, we make an invoice, else we make a credit note.

This did not work correctly as with the tax included product there was a tiny bit of difference between the product's price and what the ewallet covered.
This means that the total was just a tiny bit under zero, and was treated as a refund, so a credit note was then created.

The same problem was observed when preparing the invoice lines, the order was treated as a refund, so the lines' qty were inverted.

We now use float_compare to avoid such rounding issues.

opw-5490407

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -793,7 +793,7 @@ class PosOrder(models.Model):
             'pos_refunded_invoice_ids': pos_refunded_invoice_ids,
             'pos_order_ids': self.ids,
             'journal_id': self.session_id.config_id.invoice_journal_id.id,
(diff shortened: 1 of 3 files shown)


davidmonnom (approved):
@robodoo r+

PR #248187, opened 2026-02-11
[REF] website: generate HeaderTemplateOption from resource

duau-odoo:
This commit refactors `HeaderTemplateOption` to use a provider-based resource instead of a hardcoded list in the template.

Previously, adding new header templates required using XPath to inject elements into the 'website.HeaderTemplateOption' template. This was inconsistent with the footer's implementation, which uses a resource provider.

This commit introduces the 'header_templates_providers' resource. The header option now iterates over provided templates, aligning the API with 'footer_templates_providers' and offering a cleaner extension point for customisation.

No functional changes are introduced.

task-5928790

### DIFF
--- a/addons/website/static/src/builder/plugins/options/header/header_option_plugin.js
+++ b/addons/website/static/src/builder/plugins/options/header/header_option_plugin.js
@@ -5,12 +5,27 @@ import {
 } from "@html_builder/utils/option_sequence";
 import { Plugin } from "@html_editor/plugin";
 import { withSequence } from "@html_editor/utils/resource";
+import { _t } from "@web/core/l10n/translation";
 import { registry } from "@web/core/registry";
 import { HeaderElementsOption } from "./header_elements_option";
 import { HeaderFontOption } from "./header_font_option";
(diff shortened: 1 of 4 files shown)


guja-odoo:
@duau-odoo LGTM :+1:

addons/website/static/src/builder/plugins/options/header/header_option_plugin.js
```diff
@@ -63,7 +80,63 @@ export class HeaderOptionPlugin extends Plugin {
...
+
+        // we don't wait for all promises to resolve and show the ones available
+        // as soon as they are (and keep them in the order of the providers)
+        const templatesByProvider = this.getResource("header_templates_providers").map((p) => {
+            const provided = [];
+            Promise.resolve(p()).then((t) => {
+                provided.push(...t);
+                templates.splice(0, Infinity, ...templatesByProvider.flat());
+            });
+            return provided;
```
FrancoisGe: I don t understand why we need to do this strange code (I know footer has the same )

PR #248189, opened 2026-02-11
[IMP] l10n_tr_nilvera_einvoice: Separate tax ID for public invoices

alny-odoo:
Adds support for Public Spending Units (PSU) in public sector invoices and enforces partner address and VAT validation.

Current behavior before PR:
- No field to select Public Spending Units.
- PSU not included in the BuyerCustomerParty node.
- No strict validation for partner address and VAT.

- Users can select Public Spending Units for public sector invoices.
- Selected PSUs are added to the BuyerCustomerParty node in the XML.
- Partner address and VAT are validated before invoice processing.

task-5407085

---

### DIFF
--- a/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
@@ -162,9 +162,12 @@ class AccountEdiXmlUblTr(models.AbstractModel):
         For export invoices, a buyer party node is created based on the customer,
         and the PartyIdentification values are updated to indicate an export customer.
 
+        For public sector invoices, a buyer party node is created based on the public sector partner.
+
         :param document_node: dict representing the invoice XML structure to modify.
         :param vals: dict containing invoice-related values, including 'invoice'.
         """
+        buyer_party_node = False
         if vals['invoice'].l10n_tr_is_export_invoice:
             buyer_party_node = self._get_party_node({**vals, 'partner': vals['customer'], 'role': 'buyer'})
             buyer_party_node['cac:PartyIdentification'] = [{'cbc:ID': {
@@ -172,7 +175,11 @@ class AccountEdiXmlUblTr(models.AbstractModel):
                 'schemeID': 'PARTYTYPE',
                 },
             }]
-            document_node['cac:BuyerCustomerParty'] = {'cac:Party': buyer_party_node}
+
+        elif vals['invoice'].l10n_tr_gib_invoice_scenario == 'KAMU' and vals['invoice'].l10n_tr_public_spending_unit_id:
+            buyer_party_node = self._get_party_node({**vals, 'partner': vals['invoice'].l10n_tr_public_spending_unit_id, 'role': 'buyer'})
+
+        document_node['cac:BuyerCustomerParty'] = {'cac:Party': buyer_party_node} if buyer_party_node else None
 
     def _add_invoice_delivery_nodes(self, document_node, vals):
         super()._add_invoice_delivery_nodes(document_node, vals)
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move.py
@@ -147,6 +147,20 @@ class AccountMove(models.Model):
         depends=['l10n_tr_nilvera_pdf_file'],
     )
     l10n_tr_original_invoice_date = fields.Date(string="Original Invoice Date")
+    l10n_tr_public_spending_unit_id = fields.Many2one(comodel_name='res.partner',
+        string='Public Spending Unit',
+        store=True,
+        readonly=False,
+        compute='_compute_l10n_tr_public_spending_unit_id',
+    )
+
+    @api.depends('move_type', 'l10n_tr_gib_invoice_scenario')
+    def _compute_l10n_tr_public_spending_unit_id(self):
+        for record in self:
+            if record.move_type == 'out_invoice' and record.l10n_tr_gib_invoice_scenario == 'KAMU':  # Public Scenario Invoices
+                record.l10n_tr_public_spending_unit_id = record.l10n_tr_public_spending_unit_id or record.partner_id
+            else:
+                record.l10n_tr_public_spending_unit_id = False
 
     @api.depends('invoice_line_ids.tax_ids')
     def _compute_l10n_tr_l10n_tr_zero_vat_warning(self):
--- a/addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
+++ b/addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
@@ -139,6 +139,20 @@ class AccountMoveSend(models.AbstractModel):
                 'action': moves_with_invalid_name._get_records_action(name=_("Check name on Invoice(s)")),
             }
 
+        if public_spending_units_missing_vat := tr_nilvera_moves.l10n_tr_public_spending_unit_id.filtered(lambda rec:
+            not rec.vat
+            or not rec.street
+            or not rec.city
+            or not rec.state_id
+            or not rec.country_id,
+        ):
+            alerts['tr_public_spending_units_missing_vat'] = {
+                'level': 'danger',
+                'message': self.env._("For Public Sector e-Invoices, the Public Spending Unit must have a valid address and Tax ID defined."),
+                'action_text': self.env._("View Partner(s)"),
+                'action': public_spending_units_missing_vat._get_records_action(name=self.env._("Check VAT on Partner(s)")),
+            }
+
         exemption_702 = self.env['account.chart.template'].ref('l10n_tr_nilvera_einvoice.account_tax_code_702')
         # Warning alert if a line has no product and is missing CTSP Number
         tr_export_moves = tr_nilvera_moves.filtered(
--- a/addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
+++ b/addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
@@ -93,6 +93,13 @@
             <xpath expr="//group[@name='sale_info_group']/field[@name='ref']" position="after">
                 <field name="l10n_tr_original_invoice_date" string="Original Invoice Date" invisible="country_code != 'TR' or move_type != 'out_refund' or reversed_entry_id" required="country_code == 'TR' and not reversed_entry_id and move_type == 'out_refund'"/>
             </xpath>
+            <group id="header_left_group" position="inside">
+                <field name="l10n_tr_public_spending_unit_id"
+                       widget="res_partner_many2one"
+                       readonly="state != 'draft'"
+                       invisible="move_type != 'out_invoice' or l10n_tr_gib_invoice_scenario != 'KAMU'"
+                       required="move_type == 'out_invoice' and l10n_tr_gib_invoice_scenario == 'KAMU'"/>
+            </group>
             <xpath expr="//sheet" position="before">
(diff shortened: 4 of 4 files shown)


addons/l10n_tr_nilvera_einvoice/models/account_edi_xml_ubl_tr.py
```diff
@@ -137,13 +137,22 @@ def _add_invoice_buyer_customer_party_nodes(self, document_node, vals):
...
                 },
             }]
+        elif vals['invoice'].l10n_tr_gib_invoice_scenario == 'KAMU':
+            partner = vals['invoice'].l10n_tr_public_spending_unit_id
+
+        if partner:
+            buyer_party_node = self._get_party_node({**vals, 'partner': partner, 'role': 'buyer'})
+            if party_type_identification:
+                buyer_party_node['cac:PartyIdentification'] = party_type_identification
             document_node['cac:BuyerCustomerParty'] = {'cac:Party': buyer_party_node}
```
khfa-odoo: I would say just add the new condition without  abstracting the field

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -103,6 +103,18 @@ class AccountMove(models.Model):
...
+                                                   store=True,
+                                                   compute='_compute_l10n_tr_public_spending_unit_id')
+
+    @api.depends('move_type', 'l10n_tr_gib_invoice_scenario')
+    def _compute_l10n_tr_public_spending_unit_id(self):
+        public_scenario_invoices = self.filtered(lambda r: r.move_type == 'out_invoice'
+                                                 and r.l10n_tr_gib_invoice_scenario == 'KAMU')
+        (self - public_scenario_invoices).l10n_tr_public_spending_unit_id = False
+        for record in public_scenario_invoices.filtered(lambda r: not r.l10n_tr_public_spending_unit_id):
+            record.l10n_tr_public_spending_unit_id = record.partner_id
```
khfa-odoo: this field is not editable, so i will either be partner_id or False,  is it really required?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -103,6 +103,18 @@ class AccountMove(models.Model):
...
+                                                   string='Public Spending Unit',
+                                                   store=True,
+                                                   compute='_compute_l10n_tr_public_spending_unit_id')
+
+    @api.depends('move_type', 'l10n_tr_gib_invoice_scenario')
+    def _compute_l10n_tr_public_spending_unit_id(self):
+        public_scenario_invoices = self.filtered(lambda r: r.move_type == 'out_invoice'
+                                                 and r.l10n_tr_gib_invoice_scenario == 'KAMU')
+        (self - public_scenario_invoices).l10n_tr_public_spending_unit_id = False
+        for record in public_scenario_invoices.filtered(lambda r: not r.l10n_tr_public_spending_unit_id):
```
khfa-odoo: this feels redundant, just include it in the loop and assign False if needed

you are skipping for records with l10n_tr_public_spending_unit_id filled, but if the partner_id changes, user has no way to modify the field

khfa-odoo (changes requested):

addons/l10n_tr_nilvera_einvoice/views/account_move_views.xml
```diff
@@ -72,6 +72,13 @@
                         type="object"
                         invisible="move_type != 'out_invoice' or l10n_tr_nilvera_send_status != 'succeed'"/>
             </xpath>
+            <group id="header_left_group" position="inside">
+                <field name="l10n_tr_public_spending_unit_id"
+                       widget="res_partner_many2one"
+                       readonly="state != 'draft'"
```
khfa-odoo, line 99: this is bad design, the field is not editable at python level. Why not also make it readonly=False in python?

addons/l10n_tr_nilvera_einvoice/models/account_move.py
alny-odoo: the field is editable. partner_id is just acting as a default value for the field.

khfa-odoo (approved):

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -147,6 +147,19 @@ class AccountMove(models.Model):
         depends=['l10n_tr_nilvera_pdf_file'],
     )
     l10n_tr_original_invoice_date = fields.Date(string="Original Invoice Date")
+    l10n_tr_public_spending_unit_id = fields.Many2one(comodel_name='res.partner',
+                                                      string='Public Spending Unit',
+                                                      store=True,
+                                                      readonly=False,
+                                                      compute='_compute_l10n_tr_public_spending_unit_id')
```
```suggestion
    l10n_tr_public_spending_unit_id = fields.Many2one(
        comodel_name='res.partner',
        string="Public Spending Unit",
        store=True,
        readonly=False,
        compute='_compute_l10n_tr_public_spending_unit_id',
    )
```

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -147,6 +147,19 @@ class AccountMove(models.Model):
...
+
+    @api.depends('move_type', 'l10n_tr_gib_invoice_scenario')
+    def _compute_l10n_tr_public_spending_unit_id(self):
+        for record in self:
+            if record.move_type == 'out_invoice' and record.l10n_tr_gib_invoice_scenario == 'KAMU':  # Public Scenario Invoices
+                record.l10n_tr_public_spending_unit_id = record.l10n_tr_public_spending_unit_id or record.partner_id
+            else:
+                record.l10n_tr_public_spending_unit_id = False
```
malb-odoo, line 163: FYI it will be false by default since it's stored, so unless you want to change the value of the field along the way the else is not necessary

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -139,6 +139,20 @@ def _is_valid_nilvera_name(move):
...
+        if public_spending_units_missing_vat := tr_nilvera_moves.l10n_tr_public_spending_unit_id.filtered(lambda rec:
+            not rec.vat
+            or not rec.street
+            or not rec.city
+            or not rec.state_id
+            or not rec.country_id,
+        ):
+            alerts['tr_public_spending_units_missing_vat'] = {
+                'level': 'danger',
+                'message': _("For Public Sector e-Invoices, the Public Spending Unit must have a valid address and Tax ID defined."),
```
malb-odoo:
```suggestion
                'message': self.env._("For Public Sector e-Invoices, the Public Spending Unit must have a valid address and Tax ID defined."),
```

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -139,6 +139,20 @@ def _is_valid_nilvera_name(move):
...
+                'action_text': _("View Partner(s)"),
```
malb-odoo:
```suggestion
                'action_text': self.env._("View Partner(s)"),
```

addons/l10n_tr_nilvera_einvoice/models/account_move_send.py
```diff
@@ -139,6 +139,20 @@ def _is_valid_nilvera_name(move):
...
+                'action': public_spending_units_missing_vat._get_records_action(name=_("Check VAT on Partner(s)")),
```
malb-odoo:
```suggestion
                'action': public_spending_units_missing_vat._get_records_action(name=self.env._("Check VAT on Partner(s)")),
```

malb-odoo (commented):
Just passing through quickly to assign a reviewer on the task 👀

addons/l10n_tr_nilvera_einvoice/models/account_move.py
```diff
@@ -147,6 +147,19 @@ class AccountMove(models.Model):
...
+                                                      string='Public Spending Unit',
+                                                      store=True,
+                                                      readonly=False,
+                                                      compute='_compute_l10n_tr_public_spending_unit_id')
+
+    @api.depends('move_type', 'l10n_tr_gib_invoice_scenario')
+    def _compute_l10n_tr_public_spending_unit_id(self):
+        for record in self:
+            if record.move_type == 'out_invoice' and record.l10n_tr_gib_invoice_scenario == 'KAMU':  # Public Scenario Invoices
+                record.l10n_tr_public_spending_unit_id = record.l10n_tr_public_spending_unit_id or record.partner_id
```
FlorianGilbert, line 161: In which scenario, it's something different than the partner_id? 
Do we need this field, isn't it sufficient to use the partner? And if needed specify on the partner that is a public sector partner?

FlorianGilbert (changes requested):
I don't get why we need another field. Please be aware that adding field on account.move is never a good idea and should be avoid when possible. It's already a huge table.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
alny-odoo: hi @FlorianGilbert it's usually different from the partner_id here partner_id is just acting like a default. the user can edit this field on the view.

> In Türkiye, Public Sector e-Invoices are invoices issued to governmental institutions where two different Tax Identifications must be declared on the same invoice:
The recipient Public Institution (main customer), and
The Public Spending Unit, which represents the budgeting and expenditure authority.

FlorianGilbert: Hello @alny-odoo 
And couldn't we have the second vat on the partner id?

FlorianGilbert (approved):
Except one comment (I already sent) LGTM.

addons/l10n_tr_nilvera_einvoice/models/account_move.py
alny-odoo: > Hello @alny-odoo And couldn't we have the second vat on the partner id?

Hi @FlorianGilbert check:
```python
        elif vals['invoice'].l10n_tr_gib_invoice_scenario == 'KAMU' and vals['invoice'].l10n_tr_public_spending_unit_id:
            buyer_party_node = self._get_party_node({**vals, 'partner': vals['invoice'].l10n_tr_public_spending_unit_id, 'role': 'buyer'})
```

_get_party_node here doesnt only get vat from the partner but also other fields such as the address and name.

FlorianGilbert: I am sorry @alny-odoo 
I am not sure to understand what you are saying here, my question is about puttin the field `l10n_tr_public_spending_unit_id` on the partner instead of the move. I am not sure to understand what I have to check on your comment :sweat_smile: Sorry!

alny-odoo: > I am sorry [@alny-odoo](https://github.com/alny-odoo) I am not sure to understand what you are saying here, my question is about puttin the field `l10n_tr_public_spending_unit_id` on the partner instead of the move. I am not sure to understand what I have to check on your comment 😅 Sorry!

ah I understand what you mean now, I'll check with the PO regarding this but I believe that the same partner could do an invoice with a public spending unit and one invoice without.

alny-odoo: hi @FlorianGilbert I checked with the PO. and there are cases where the l10n_tr_public_spending_unit_id is different for the same partner_id

for example:
Customer: Dubai municipality
Spending unit: Internal affairs

Customer: Dubai municipality
Spending unit: External affairs

FlorianGilbert: Ok make sense! @alny-odoo Thank you for checking.

FlorianGilbert:
@robodoo r+
@robodoo delegate+

alny-odoo:
@robodoo r+

PR #248193, opened 2026-02-11
[FIX] l10n_gcc_pos: fix test generic localization for gcc modules

vpan-odoo:
- Fixed `TestGenericGCC.test_generic_localization` which was failing
  because some information was not rendered on the order receipt.
- Added rounding configuration to the POS config so that the assertion
  for `Rounding` does not fail.
- Added steps for `Discount` and `Change` in `generic_localization_tour`
  so that the assertions for `Discount` and `Change` do not fail.

Error-237988
Task-5897376

---


addons/l10n_gcc_pos/tests/test_gcc_pos.py
```diff
@@ -30,12 +31,12 @@ def setUpClass(cls):
 
     def test_generic_localization(self):
         self.main_pos_config.l10n_gcc_dual_language_receipt = True
+        self.main_pos_config.cash_rounding = True
+        self.main_pos_config.rounding_method = self.account_cash_rounding_half
```
adgu-odoo: If you inherit from CommonPosTest only for the cash_rounding, I prefer that you do
```suggestion
        self.main_pos_config.rounding_method = self.env['account.cash.rounding'].create({
            'name': 'Test rounding',
            'rounding': 0.05,
        })
```

addons/l10n_gcc_pos/tests/test_gcc_pos.py
```diff
@@ -30,12 +31,12 @@ def setUpClass(cls):
...
         _, html = super().test_generic_localization()
         self.assertTrue("Served by / خدم بواسطة" in html)
         self.assertTrue("Ticket / تذكرة" in html)
         self.assertTrue(" / الفاتورة الضريبية" in html)
         self.assertTrue("Subtotal / الإجمالي الفرعي" in html)
         self.assertTrue("Rounding / التقريب" in html)
         self.assertTrue("Total / اﻹجمالي" in html)
-        self.assertTrue("Discount / الخصومات" in html)
```
adgu-odoo, line 40: I don't get why you remove this ? I think you can keep it

vpan-odoo: > I don't get why you remove this ? I think you can keep it

Actually, this test case runs the [`generic_localization_tour`](https://github.com/odoo/odoo/blob/saas-19.2/addons/point_of_sale/static/tests/pos/tours/generic_tour.js#L10), which is common for all localizations' `test_generic_localization`. Since we are asserting `Change` and `Discount` only in this test case, keeping these assertions would require adding the corresponding steps to the common tour, which would then run unnecessarily for all other localizations.

vpan-odoo: @adgu-odoo added the assertions for `Change` and `Discount`

adgu-odoo:
robodoo r+

adgu-odoo:
robodoo r+

PR #248194, opened 2026-02-11
[IMP] account: add a country-specific check for structured reference validity

SaddemAmine:
Currently, when initiating a payment, we check if the reference is a
structured one by using `is_valid_structured_reference` which checks
the validity of the structure accross all supported countries.

This can lead to issues when it matches formats accepted by other
countries but not the one of the bank account.

With this commit, we replace this check by a call to a new function that
checks the structure validity according to the country of the bank account,
with a fallback to the generic check (ISO 11649) if the country is not supported.

opw-5387269

### DIFF
--- a/addons/account/tools/structured_reference.py
+++ b/addons/account/tools/structured_reference.py
@@ -168,3 +168,25 @@ def is_valid_structured_reference(reference):
         is_valid_structured_reference_si(reference) or
         is_valid_structured_reference_iso(reference)
     )
+
+
+def is_valid_structured_reference_for_country(reference, country_code=''):
+    """Check the validity of the reference's structure for a specific country or ISO 11649 as a fallback.
+
+    :param reference: the reference to check
+    :param country_code: the country code to check against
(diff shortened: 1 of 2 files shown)


addons/account/tools/structured_reference.py
```diff
@@ -167,4 +167,26 @@ def is_valid_structured_reference(reference):
...
+
+
+def is_valid_structured_reference_for_country(reference, country_code):
+    """Check the validity of the reference's structure for a specific country or ISO 11649 as a fallback.
+
+    :param reference: the reference to check
+    :param country_code: the country code to check against
+    :return: True if reference is a structured reference for the given country or ISO 11649, False otherwise
+    """
+    check_per_country = {
```
SaddemAmine, line 180: The mapping isn't a top level constant because at import time the functions would be undefined.

malb-odoo (commented):
Could you add a small test for this 😄 Diff looks ok but let's bullet proof it

malb-odoo (approved):
Nice jobbb
@robodoo r+

PR #229675, opened 2025-10-02
[IMP] core: bump minimal python version to 3.12

kmagusiak:
[IMP] core: bump minimal python version to 3.12
    
    Select the minimum python version for Odoo 20.0 release.
    We are looking at Debian 13 and Ubuntu 24.04 and 26.04 (which will be
    released before Odoo 20.0). The minimum version of these systems is
    python 3.12.

    [IMP] core: bump minimal postgresql version to 15
    
    Once Odoo 20.0 will be out postgresql 14 will be around a month from
    being out of support. We set the minimal supported version to the next
    available version.

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248197, opened 2026-02-11
PyPDF2 2.1 compatibility removal

xmo-odoo:
PyPDF2 1.x compatibility was dropped by #229675 but the compatibility shims were left in:

- `odoo.tools.pdf._pypdf2_1`
- use of the 1.x style API (camelCase) rather than the modern api (snake_case), requiring extensive shimming in `odoo.tools._pypdf`


odoo/tools/pdf/_pypdf.py
```diff
@@ -13,70 +16,100 @@
 ]
 
 
-pypdf.PageObject.mergePage = lambda self, page2: self.merge_page(page2)
-pypdf.PageObject.mediaBox = property(lambda self: self.mediabox)
+def deprecate(fn):
```
kmagusiak: For all deprecation warnings, maybe we can move `odoo.orm.decorators.deprecated` to tools (misc?) and use it here an in other files?

xmo-odoo: I'd certainly have used it if it was in `odoo.tools` instead of rolling my own.

`func` would probably be a better location unless there's more warnings-related stuff to coalesce into their own module.

kmagusiak: That function is a backport of `warnings.deprecated` coming in py3.13. Could you move it to odoo.tools.func in this PR?

PR #248206, opened 2026-02-11
[FIX] html_editor: handle column insertion inside list items

walidsahli:
Problem:
When trying to add a column under a list item, a traceback occurs.

Cause:
The list item is removed during the operation, but the selection that is restored still references the removed list item. As a result, the selection is restored on a disconnected element, causing a traceback.

Solution:
Handle column insertion inside lists separately. In this case, according to the specifications, we split the list and insert the columns between the resulting lists (or after the list if the cursor is in the last list item). This avoids restoring the selection on a removed node.

Steps to reproduce:
- Add any list.
- Run `/column` while the cursor is inside a list item.
- Observe the traceback.

task-5916246

---

### DIFF
--- a/addons/html_editor/static/src/main/column_plugin.js
+++ b/addons/html_editor/static/src/main/column_plugin.js
@@ -98,9 +98,47 @@ export class ColumnPlugin extends Plugin {
                 this.removeColumns(anchor);
             }
         } else if (numberOfColumns) {
+            const li = closestElement(anchor, "li");
+            if (li) {
+                this.createColumnsFromList(anchor, li, numberOfColumns);
+                return;
+            }
             this.createColumns(anchor, numberOfColumns, addParagraphAfter);
         }
+
         this.dependencies.selection.setSelection(selectionToRestore);
+        this.dependencies.history.addStep();
+    }
+
+    createColumnsFromList(anchor, li, numberOfColumns) {
+        const currentList = li.closest("ol, ul");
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/main/list/list_plugin.js
```diff
@@ -577,6 +579,44 @@ export class ListPlugin extends Plugin {
         return null;
     }
 
+    handleColumnize(selection, numberOfColumns) {
```
dmo-odoo: I think having the list plugin handle columnize in the case of lists is conceptually wrong. Lists exist as a concept in pure html whereas our columns from columnize do not. Therefore, I think it makes more sense that the column plugin be te one to handle the sepcial case of inserting its columns stuff into an html list. As a bonus, it will allow to remove the util.

walidsahli: Is it okay then to check if we are in a list in the column plugin ? `closestElement(node, "li")` in column plugin ?

dmo-odoo: Yes. Since `li` is a valid html tag, it is better to reference that in `ColumnPlugin` than to reference our custom columns in the `ListPlugin`.

walidsahli: @dmo-odoo updated.

PR #182334, opened 2024-10-01
[IMP] mail, portal, *: replace the portal chatter fetch route

maki-odoo:
This change replaces `/mail/chatter_fetch` with `/mail/thread/messages` in order
to reuse the base mail code in portal as much as possible.

part of task-3599659

[Enterprise PR](https://github.com/odoo/enterprise/pull/121606)

PR #248208, opened 2026-02-11
[IMP] mail, portal: support extra params in message fetch methods

maki-odoo:
This is a preparation change for replacing the portal chatter fetch route with
the mail fetch route.
Fetch methods can now receive extra fetch params from the caller components if
needed, making it possible to inform the server when a request originates from
the portal. This allows the server to apply specific limitations required for
portal chatter.

The change also involves a minor refactoring of the `useMessageScrolling` hook,
which makes the thread built-in and removes the need to pass it as a parameter
in every call. Extra fetch params could also be passed through another callback
in the same way.

[Enterprise PR](https://github.com/odoo/enterprise/pull/107640)

### DIFF
--- a/addons/mail/static/src/chatter/web_portal_project/@types/models.d.ts
+++ b/addons/mail/static/src/chatter/web_portal_project/@types/models.d.ts
@@ -1,5 +1,5 @@
 declare module "models" {
     export interface Thread {
-        fetchThreadData: (requestList: string[]) => Promise<void>;
+        fetchThreadData: (requestList: string[], param0: { messageFetchRouteParams: MessageRouteParams }) => Promise<void>;
     }
 }
--- a/addons/mail/static/src/chatter/web_portal_project/chatter.js
+++ b/addons/mail/static/src/chatter/web_portal_project/chatter.js
@@ -1,4 +1,4 @@
-import { useChildSubEnv, useRef, useState } from "@web/owl2/utils";
+import { useChildSubEnv, useRef, useState, useSubEnv } from "@web/owl2/utils";
 import { Composer } from "@mail/core/common/composer";
 import { Thread } from "@mail/core/common/thread";
 import { useMessageScrolling } from "@mail/utils/common/hooks";
@@ -29,11 +29,15 @@ export class Chatter extends Component {
             aside: false,
             disabled: !this.props.threadId,
         });
-        this.messageHighlight = useMessageScrolling();
+        this.messageHighlight = useMessageScrolling({
+            thread: () => this.state.thread,
+            messageFetchRouteParams: () => this.messageFetchRouteParams,
+        });
         this.highlightMessage = router.current.highlight_message_id;
         this.rootRef = useRef("root");
         this.onScrollDebounced = useThrottleForAnimation(this.onScroll);
         useChildSubEnv(this.childSubEnv);
+        useSubEnv(this.subEnv);
 
         onMounted(this._onMounted);
         onWillUpdateProps((nextProps) => {
@@ -64,12 +68,24 @@ export class Chatter extends Component {
         };
     }
 
+    get extraMessageFetchRouteParams() {
+        return {};
+    }
+
+    get messageFetchRouteParams() {
+        return this.env.messageFetchRouteParams;
+    }
+
     get onCloseFullComposerRequestList() {
         return ["messages"];
     }
 
     get requestList() {
         return [];
+    }
+
+    get subEnv() {
+        return { messageFetchRouteParams: this.extraMessageFetchRouteParams };
     }
 
     changeThread(threadModel, threadId) {
@@ -106,7 +122,9 @@ export class Chatter extends Component {
         if (!thread.id || !this.state.thread?.eq(thread)) {
             return;
         }
-        await thread.fetchThreadData(requestList);
+        await thread.fetchThreadData(requestList, {
+            messageFetchRouteParams: this.messageFetchRouteParams,
+        });
     }
 
     onCloseFullComposerCallback() {
--- a/addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
+++ b/addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
@@ -2,10 +2,14 @@ import { Thread } from "@mail/core/common/thread_model";
 import { patch } from "@web/core/utils/patch";
 
 patch(Thread.prototype, {
-    /** @param {string[]} requestList */
-    async fetchThreadData(requestList) {
+    /**
+     * @param {string[]} requestList
+     * @param {Object} [options]
+     * @param {MessageRouteParams} [options.messageFetchRouteParams]
+     */
+    async fetchThreadData(requestList, { messageFetchRouteParams = {} } = {}) {
         if (requestList.includes("messages")) {
-            this.fetchNewMessages();
+            this.fetchNewMessages({ routeParams: messageFetchRouteParams });
         }
         await this.store.fetchStoreData("mail.thread", {
             access_params: this.rpcParams,
--- a/addons/mail/static/src/core/common/thread.js
+++ b/addons/mail/static/src/core/common/thread.js
@@ -117,7 +117,9 @@ export class Thread extends Component {
                     this.smoothScrollingDeferred,
                 ]);
                 if (this.loadOlderState.isVisible) {
-                    toRaw(this.props.thread).fetchMoreMessages();
+                    toRaw(this.props.thread).fetchMoreMessages({
+                        routeParams: this.messageFetchRouteParams,
+                    });
                 }
             },
             { ready: false }
@@ -130,7 +132,10 @@ export class Thread extends Component {
                     this.smoothScrollingDeferred,
                 ]);
                 if (this.loadNewerState.isVisible) {
-                    toRaw(this.props.thread).fetchMoreMessages("newer");
+                    toRaw(this.props.thread).fetchMoreMessages({
+                        epoch: "newer",
+                        routeParams: this.messageFetchRouteParams,
+                    });
                 }
             },
             { ready: false }
@@ -169,10 +174,7 @@ export class Thread extends Component {
         useLayoutEffect(
             () => {
                 if (this.props.thread.highlightMessage && this.state.mountedAndLoaded) {
-                    this.messageHighlight?.highlightMessage(
-                        this.props.thread.highlightMessage,
-                        this.props.thread
-                    );
+                    this.messageHighlight?.highlightMessage(this.props.thread.highlightMessage);
                     this.props.thread.highlightMessage = null;
                 }
             },
@@ -489,8 +491,12 @@ export class Thread extends Component {
         return true;
     }
 
+    get messageFetchRouteParams() {
+        return this.env.messageFetchRouteParams;
+    }
+
     fetchInitialMessages() {
-        toRaw(this.props.thread).fetchNewMessages();
+        toRaw(this.props.thread).fetchNewMessages({ routeParams: this.messageFetchRouteParams });
     }
 
     get viewportEl() {
@@ -515,7 +521,7 @@ export class Thread extends Component {
     }
 
     onClickLoadOlder() {
-        this.props.thread.fetchMoreMessages();
+        this.props.thread.fetchMoreMessages({ routeParams: this.messageFetchRouteParams });
     }
 
     async onClickPreferences() {
@@ -547,7 +553,7 @@ export class Thread extends Component {
     async jumpToPresent({ immediate = false } = {}) {
         this.messageHighlight?.clear();
         if (!immediate || this.props.thread.loadNewer) {
-            await this.props.thread.loadAround();
+            await this.props.thread.loadAround({ routeParams: this.messageFetchRouteParams });
             this.props.thread.loadNewer = false;
             this.state.showJumpPresent = false;
         }
--- a/addons/mail/static/src/core/common/thread_model.js
+++ b/addons/mail/static/src/core/common/thread_model.js
@@ -15,6 +15,17 @@ import { Deferred } from "@web/core/utils/concurrency";
  * @property {string} reason
  */
 
+/**
+ * @typedef {Object} MessageFetchParams
+ * @property {number} [after]
+ * @property {number} [around]
+ * @property {number} [before]
+ */
+
+/**
+ * @typedef {Object} MessageRouteParams Specific overrides for the RPC route metadata usually provided by the caller component.
+ */
+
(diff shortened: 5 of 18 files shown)


addons/mail/static/src/utils/common/hooks.js
```diff
@@ -395,17 +400,19 @@ export function useMessageScrolling(duration = 2000) {
...
          * @param {import("models").Message} message
-         * @param {import("models").Thread} thread
+         * @param {Object} [options]
+         * @param {import("@web/env").OdooEnv} [options.env]
          */
-        async highlightMessage(message, thread) {
-            if (thread.model !== "mail.box" && thread.notEq(message.thread)) {
+        async highlightMessage(message, { env } = {}) {
+            const thread = getThread();
+            if (!thread || (thread.model !== "mail.box" && thread.notEq(message.thread))) {
```
seb-odoo: Can't we get env like we get thread?

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -389,16 +389,22 @@ export class Thread extends Record {
...
-    /** @param {{after: Number, before: Number}} */
-    async fetchMessages({ after, around, before } = {}) {
+    /**
+     * @param {Object} [options]
+     * @param {number} [options.after]
+     * @param {number} [options.around]
+     * @param {number} [options.before]
+     * @param {import("@web/env").OdooEnv} [options.env]
+     */
+    async fetchMessages({ after, around, before, env } = {}) {
```
seb-odoo: As discussed, I would prefer to forward the specific fetch param that we need rather than the whole env.
Gives more controller to callers, rather than expecting them to "fill" env to fit their needs.

addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
```diff
@@ -2,10 +2,13 @@ import { Thread } from "@mail/core/common/thread_model";
 import { patch } from "@web/core/utils/patch";
 
 patch(Thread.prototype, {
-    /** @param {string[]} requestList */
-    async fetchThreadData(requestList) {
+    /**
+     * @param {string[]} requestList
+     * @param {Object} [fetchParams]
+     */
+    async fetchThreadData(requestList, fetchParams = {}) {
```
seb-odoo: Inside all "fetch" methods, if the param will go into the route, I would name it `routeParams` instead, as fetch is both redundant with the method name and actually incorrect.

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -415,11 +417,19 @@ export class Thread extends Record {
...
+    /**
+     * @param {Object} [fetchParams]
+     * @param {number} [fetchParams.after]
+     * @param {number} [fetchParams.around]
+     * @param {number} [fetchParams.before]
+     * @param {Object} [fetchParams.messageFetchRouteParams] Specific overrides for the RPC route metadata provided by the caller component.
+     */
+    async fetchMessagesData(fetchParams = {}) {
         // ordered messages received: newest to oldest
+        const { after, around, before, ...messageFetchRouteParams } = fetchParams;
```
seb-odoo: I would prefer to have 2 different parameters, one for the `fetch_params` (-> `_message_fetch`) that can say as `fetchParams`, and the other one which would be `routeParams`, so we don't need to magically deconstruct based on name.

All (sub-)methods should provide both parameters separately.

addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
```diff
@@ -2,10 +2,13 @@ import { Thread } from "@mail/core/common/thread_model";
 import { patch } from "@web/core/utils/patch";
 
 patch(Thread.prototype, {
-    /** @param {string[]} requestList */
-    async fetchThreadData(requestList) {
+    /**
+     * @param {string[]} requestList
+     * @param {Object} [routeParams]
+     */
+    async fetchThreadData(requestList, routeParams = {}) {
```
seb-odoo: In `fetchThreadData` it needs to be `messageFetchRouteParams` too, as these params are given to the fetch route but not to the data route.

addons/mail/static/src/utils/common/hooks.js
```diff
@@ -381,9 +381,15 @@ export function useVisible(refName, cb, { ready = true } = {}) {
...
+ */
+
+/**
+ * @param {function(): import("models").Thread|null} getThread
+ * @param {function(): Object} [getFetchRouteParams=() => ({})]
+ * @param {number} [duration=2000]
  * @returns {MessageScrolling}
  */
-export function useMessageScrolling(duration = 2000) {
+export function useMessageScrolling(getThread, getFetchRouteParams = () => ({}), duration = 2000) {
```
seb-odoo: For consistency with `messageFetchRouteParams` in other places, I would use this term here too (with `get` prefix.)

addons/mail/static/src/chatter/web_portal_project/@types/models.d.ts
```diff
@@ -1,5 +1,5 @@
 declare module "models" {
     export interface Thread {
-        fetchThreadData: (requestList: string[]) => Promise<void>;
+        fetchThreadData: (requestList: string[], routeParams: Object) => Promise<void>;
```
seb-odoo: This needs to be updated after last change (good job on updating it in the first place though).

addons/mail/static/src/core/common/thread.js
```diff
@@ -465,8 +467,12 @@ export class Thread extends Component {
         }
     }
 
+    get messageFetchRouteParams() {
+        return {};
+    }
```
seb-odoo, line 496: This is defined in both chatter and thread. Won't this lead to having to override it in both places in portal?
I'd rather have a single override.
chatter -> thread should likely be a props, or something in (sub)env

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -420,15 +424,21 @@ export class Thread extends Record {
...
+     * @param {Object} [fetchParams]
+     * @param {number} [fetchParams.after]
+     * @param {number} [fetchParams.around]
+     * @param {number} [fetchParams.before]
+     * @param {Object} [routeParams] Specific overrides for the RPC route metadata provided by the caller component.
      * @returns {Promise<{messages: number[]}>}
      */
-    async fetchMessagesData({ after, around, before } = {}) {
+    async fetchMessagesData(fetchParams = {}, routeParams = {}) {
+        const { after, around, before } = fetchParams;
```
seb-odoo: They could directly be spread into the param like before.

addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
```diff
@@ -2,10 +2,13 @@ import { Thread } from "@mail/core/common/thread_model";
 import { patch } from "@web/core/utils/patch";
 
 patch(Thread.prototype, {
-    /** @param {string[]} requestList */
-    async fetchThreadData(requestList) {
+    /**
+     * @param {string[]} requestList
+     * @param {Object} [routeParams]
```
seb-odoo: rename

addons/mail/static/src/core/common/thread.js
```diff
@@ -465,8 +467,12 @@ export class Thread extends Component {
...
+
     fetchInitialMessages() {
-        toRaw(this.props.thread).fetchNewMessages();
+        toRaw(this.props.thread).fetchNewMessages(this.messageFetchRouteParams);
```
seb-odoo: It's kind of strange/inconsistent that some methods (like `fetchNewMessages` and `fetchMessagesData`) have `routeParams` as a positional argument, and others (like `fetchMoreMessages` and `fetchMessages`) have it as a named argument inside an object.

maki-odoo: @seb-odoo 
I made all signatures consistant as:
- `fetchMessages({ fetchParams = {}, routeParams = {} } = {})`
- `fetchMessagesData({ fetchParams: { after, around, before } = {}, routeParams = {} } = {})`
- `fetchMoreMessages({ epoch = "older", routeParams = {} } = {})`
- `fetchNewMessages({ routeParams = {} } = {})`
- `fetchThreadData(requestList, { messageFetchRouteParams = {} } = {})`
- `loadAround({ messageId, routeParams = {} } = {})`

If it's ok with you, I'll merge the commits.

addons/mail/static/src/chatter/web_portal_project/thread_model_patch.js
```diff
@@ -2,10 +2,13 @@ import { Thread } from "@mail/core/common/thread_model";
 import { patch } from "@web/core/utils/patch";
 
 patch(Thread.prototype, {
-    /** @param {string[]} requestList */
-    async fetchThreadData(requestList) {
+    /**
+     * @param {string[]} requestList
+     * @param {Object} [messageFetchRouteParams]
```
seb-odoo: top level object should be `param0`/`options`
and then `options.messageFetchRouteParams` for that one

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -397,15 +397,19 @@ export class Thread extends Record {
         return this.hasReadAccess;
     }
 
-    /** @param {{after: Number, before: Number}} */
-    async fetchMessages({ after, around, before } = {}) {
+    /**
+     * @param {Object} [options]
+     * @param {Object} [options.fetchParams]
+     * @param {Object} [options.routeParams]
```
seb-odoo: As these are used in several places, it would be nice to typedef what each of them can contain in a central place so we get proper auto-complete.

addons/mail/static/src/core/common/thread.js
seb-odoo: Yes this is good, feel free to squash commits.

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -420,15 +424,24 @@ export class Thread extends Record {
     }
 
     /**
-     * @param {{after: Number, before: Number}}
+     * @param {Object} [options]
+     * @param {Object} [options.fetchParams]
+     * @param {number} [options.fetchParams.after]
+     * @param {number} [options.fetchParams.around]
+     * @param {number} [options.fetchParams.before]
+     * @param {Object} [options.routeParams] Specific overrides for the RPC route metadata provided by the caller component.
```
seb-odoo: It's not necessarily provided by the caller component, any caller could provide it. Maybe specify "usually"?

addons/mail/static/src/core/common/thread.js
```diff
@@ -465,8 +467,12 @@ export class Thread extends Component {
         }
     }
 
+    get messageFetchRouteParams() {
+        return this.env.messageFetchRouteParams;
```
seb-odoo, line 495: I've not seen where this is added in env, shouldn't chatter do it at least?

maki-odoo: It will be added to the env when the fetch route is replaced. There is currently no use for it.

seb-odoo: Well there is currently no use of anything this PR is doing, but as a prep the goal is to get everything ready.

addons/mail/static/src/chatter/web_portal_project/chatter.js
```diff
@@ -70,6 +75,14 @@ export class Chatter extends Component {
         };
     }
 
+    get extraMessageFetchRouteParams() {
+        return {};
+    }
```
maki-odoo, line 73: I'm not sure about the name, but since we already have `get messageFetchRouteParams()`, I couldn't think of a better one.

addons/mail/static/src/utils/common/hooks.js
```diff
@@ -381,9 +381,19 @@ export function useVisible(refName, cb, { ready = true } = {}) {
...
+ * @param {function(): Object} [getMessageFetchRouteParams=() => ({})]
+ * @param {number} [duration=1500]
  * @returns {MessageScrolling}
  */
-export function useMessageScrolling(duration = 1500) {
+export function useMessageScrolling(
+    getThread,
+    getMessageFetchRouteParams = () => ({}),
+    duration = 1500
+) {
```
alexkuhn:
```suggestion
export function useMessageScrolling({
    thread,
    messageFetchRouteParams = () => ({}),
    duration = 1500
}) {
```

I prefer named params, and also plain names `thread`, to match with other important hooks like `useThreadActions`.

addons/project/static/src/project_sharing/chatter/chatter_patch.js
```diff
@@ -18,6 +18,13 @@ patch(Chatter.prototype, {
         });
     },
 
+    get extraMessageFetchRouteParams() {
+        return {
+            ...super.extraMessageFetchRouteParams,
+            ...(this.props.projectSharingId ? { only_portal: true } : {}),
```
alexkuhn: `only_portal` looks unused

maki-odoo: Yes, this is actually the extra parameter that we need for the next PR ([replacing the fetch route](https://github.com/odoo/odoo/pull/182334)). It is the parameter that the request from the portal will send to the fetch route.

[See also](https://github.com/odoo/odoo/pull/248208#discussion_r2816130566 )

addons/mail/static/src/utils/common/hooks.js
maki-odoo: I think the plain names are slightly harder to read here. Why use a plain name when `getThread` is clear enough? Do we want it to be an object and not a function in some cases?
I suggest that we use the named params but not the plain names for this case:
```js
export function useMessageScrolling({
    getThread,
    getMessageFetchRouteParams = () => ({}),
    duration = 1500,
})
```

addons/project/static/src/project_sharing/chatter/chatter_patch.js
alexkuhn: Let's add when we actually use it, having it now is confusing and even next PR that needs it, it feels weird to use the param for the 1st time :)

addons/mail/static/src/utils/common/hooks.js
alexkuhn: `getThread` is harder to read than `thread`, and being a function is more a shenanigans from OWL 2 lack of reactivity.

With OWL3 this would be way easier, and I think that having easy to read names is better than having prefix `get` because this is technically a function we provide.

alexkuhn (approved):
@robodoo delegate+ r+

maki-odoo:
@robodoo r+

PR #191132, opened 2024-12-19
[FIX] website_sale_loyalty: prevent reload cycle on multi tax discounts

lvsz:
Versions
--------
- saas-17.4+

Steps
-----
1. Create a deliverable product with sales tax;
2. create another product without sales tax;
3. publish both to eCommerce;
4. create a discount program;
5. add both products to shopping cart;
6. apply discount;
7. go to delivery selection step.

Issue
-----
Endless browser reload cycle.

Cause
-----
The reload is supposed to sync the discount lines in the back-end with the discount lines displayed during checkout. If the number of lines don't match, a reload is triggered.

The issue is that on the back-end, discounts are counted per line, while in eCommerce, the discounts are displayed per reward. This difference matters when a discount is applied on products with different sales taxes, creating a separate discount line for every sales tax.

Solution
--------
When iterating over discount lines in the controller, group them per `reward_id`, and sum up their subtotals.

opw-4343959

PR #248215, opened 2026-02-11
[FIX] website_sale_loyalty: prevent reload cycle on multi gift card

mma1377:
### Issue:
Applying multi gift card from the same program results in an infinite reload cycle.

#### Steps to reproduce:
1- Create a gift card program and generate two gift cards.
2- In eCommerce add a deliverable product to cart.
3- Open cart and add both gift card codes.
4- Click on checkout to navigate to delivery step.

The page is going to keep refreshing.

### Cause:
The reload is supposed to sync the discount lines in the back-end with the discount lines displayed during checkout. If the number of lines don't match, a reload is triggered.:
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/website_sale_loyalty/static/src/js/checkout.js#L22-L24

After the fix #191132, discount lines are grouped by `reward_id`. This is done because in the back-end, by applying a single percent discount, separate lines are created per product lines with different taxes, while in front-end there is still single line.

However, that fix doesn't consider that a SO might have multiple lines of the same reward: ewallet, gift_card

### Fix:
This can be fixed by separating logic of groupable lines which are the lines with `percent` mode, with `per-point` mode lines.

Discount is the only mode that might be added to multiple lines separately, keeping #191132 fix, and `per-point` is the only mode that might be applied multiple times per reward (in ewallet and gift_card).

opw-5485912

### DIFF
--- a/addons/website_sale_loyalty/controllers/delivery.py
+++ b/addons/website_sale_loyalty/controllers/delivery.py
@@ -34,9 +34,14 @@ class WebsiteSaleLoyaltyDelivery(Delivery):
             res['delivery_discount_minor_amount'] = payment_utils.to_minor_currency_units(
                 shipping_discount, order.currency_id
             )
+        discount_lines = order.order_line.filtered(
+            lambda line: line.reward_id.reward_type == 'discount'
+        )
+        groupable_lines = discount_lines.filtered(
+            lambda line: line.reward_id.discount_mode == 'percent'
+        )
         res['discount_reward_amounts'] = [
             to_html(sum(lines.mapped('price_subtotal')))
-            for reward, lines in order.order_line.grouped('reward_id').items()
-            if reward.reward_type == 'discount'
-        ]
+            for lines in groupable_lines.grouped('reward_id').values()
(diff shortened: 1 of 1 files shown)


addons/website_sale_loyalty/controllers/delivery.py
```diff
@@ -34,9 +34,14 @@ def _order_summary_values(self, order, **post):
...
+        )
+        groupable_lines = discount_lines.filtered(
+            lambda line: line.reward_id.discount_mode == 'percent'
+        )
         res['discount_reward_amounts'] = [
             to_html(sum(lines.mapped('price_subtotal')))
-            for reward, lines in order.order_line.grouped('reward_id').items()
-            if reward.reward_type == 'discount'
-        ]
+            for _, lines in groupable_lines.grouped('reward_id').items()
```
ltinel: Nit:
```suggestion
            for lines in groupable_lines.grouped('reward_id').values()
```

ltinel (changes requested):
Thanks for the fix :)

ltinel (approved):
(will already delegate, since my only comment is a nit :)
@robodoo delegate+

PR #248216, opened 2026-02-11
[FIX] im_livechat: define chatbot field in core bundle to avoid traceback

aych-odoo:
Using a chatbot while the website editor is active triggers a traceback. The issue occurs because bus notifications containing `Chatbot` model data are also received by the backend bundle.

In the backend bundle, the `Thread` model did not define the `chatbot` field, so the inverse thread relation on the `Chatbot` model could not be resolved, leading to a crash.

This commit adds the chatbot field to the core bundle so it is available in all contexts and prevents the traceback.

Steps to reproduce:
- Goto localhost:8069/@/contactus
- Start chatbot
- Answer any question -> traceback

task-5924295


alexkuhn (approved):
@robodoo delegate+ r+

PR #248218, opened 2026-02-11
[FIX] pos_self_order: prevent already sent items in Pay-After-Meal kitchen print

rpka-odoo:
Steps to Reproduce:
-------------------
1. Place a self-order (Mobile / Pay-After-Meal).
2. Add a second order.
3. Kitchen print shows old + new items instead of only the new ones.

Issue:
-------------
Pay-After-Meal flow, kitchen prints included previously sent items. 

Cause:
-----------
all orderlines were sent to the kitchen instead of only new changes.

Fix:
------------
Send only newly added/updated lines to the kitchen

Task-5929555
Related PR: https://github.com/odoo/enterprise/pull/107129

### DIFF
--- a/addons/pos_self_order/static/src/app/services/self_order_service.js
+++ b/addons/pos_self_order/static/src/app/services/self_order_service.js
@@ -499,9 +499,17 @@ export class SelfOrder extends Reactive {
                 order,
                 Object.values(printer.config.product_categories_ids)
             );
+            const lastsentlines = order.uiState.lastSentPreparationChange?.lines || {};
             if (orderlines.length > 0) {
                 const printingChanges = {
-                    new: orderlines,
+                    new: orderlines
+                        .map((line) => ({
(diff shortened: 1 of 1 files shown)


addons/pos_self_order/static/src/app/services/self_order_service.js
```diff
@@ -649,6 +659,9 @@ export class SelfOrder extends Reactive {
             const tableIdentifier = this.router.getTableIdentifier([]);
             let uuid = this.selectedOrderUuid;
             this.currentOrder.recomputeOrderData();
+            this.lastSentPreparationChange = JSON.parse(
```
jipr-odoo: creating a new `lastSentPreparationChange` in `self_order_service` is not a good idea 
i would suggest a to store it in `order uistate`

wdyt ?

jipr-odoo (changes requested):

rpka-odoo:
Close this PR as discussed with moda task- 5929555

PR #248222, opened 2026-02-11
[FIX] im_livechat: prevent traceback on tab press in chatter

shmn-odoo:
**Steps to Reproduce:**
Install 'Live Chat'
Open any record with a chatter
Click Send Message to focus the composer
Press the 'Tab' key on the keyboard

A traceback occurs because the code attempts to read 'channel_type' from a 
channel field that is undefined in standard threads.

A guard is added to the channel field. Pressing Tab no longer causes an error.

task-5925393

---


tsm-odoo (approved):
@robodoo r+

PR #248224, opened 2026-02-11
[FIX] project: update follower preferences when moving tasks between projects.

maap-odoo:
Steps to reproduce:
-
 1. Create projects A and B.
 2. Add a user as a follower of project B and select specific notification subtypes (e.g., 'Stage Changed').
 3. Create a task in project A and add the same user as a follower(defaulting to 'Discussions').
 4. Move the task from project A to project B.

Issue:
-
The follower's subscription preferences on the task do not reflect their project-level settings after the move. In the example above, the user remains subscribed only to 'Discussions' and misses 'Stage Changed' updates.

Cause:
-
The default auto-subscription logic skips existing followers. When moving a task, this prevents the system from adding the new project's notification preferences to users who were already following the task.

Fix:
-
Override `_message_auto_subscribe` in project.task to the `update` policy when the `project_id` is changed.

task-5877507

### DIFF
--- a/addons/project/models/project_task.py
+++ b/addons/project/models/project_task.py
@@ -1335,6 +1335,11 @@ class Task(models.Model):
                     mail_auto_delete=False,
                 )
 
+    def _message_auto_subscribe(self, updated_values, followers_existing_policy='skip'):
+        if updated_values.get('project_id'):
+            followers_existing_policy = 'update'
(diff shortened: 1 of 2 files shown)


maap-odoo:
Hello @MissingNoShiny :wave:,

Could you please r+ it again?

MissingNoShiny:
Oops, I didn’t see it was still a draft 🙈 
robodoo r+

PR #248227, opened 2026-02-11
[FIX] html_editor: turn tabs to four spaces inside monospace banner

bso-odoo:
Tabs are aligned on 40px boundaries. This leads to misalignment of characters when using the monospace banner.

This commit fixes this by replacing all tabs inside monospace banners by four spaces.

Steps to reproduce:
- Go to a "To do" note
- Insert a monospace banner
- Type some text on several lines
- Type tab at the begin of some lines

=> Character columns were misaligned.

task-5916747

### DIFF
--- a/addons/html_editor/static/src/main/banner_plugin.js
+++ b/addons/html_editor/static/src/main/banner_plugin.js
@@ -1,6 +1,6 @@
 import { Plugin } from "@html_editor/plugin";
 import { fillEmpty, fillShrunkPhrasingParent } from "@html_editor/utils/dom";
-import { closestElement, selectElements } from "@html_editor/utils/dom_traversal";
+import { closestElement, descendants, selectElements } from "@html_editor/utils/dom_traversal";
 import { parseHTML } from "@html_editor/utils/html";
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/main/banner_plugin.js
```diff
@@ -191,4 +195,11 @@ export class BannerPlugin extends Plugin {
         this.dependencies.selection.setCursorStart(baseContainer);
         return true;
     }
+
+    handle_monospace_tab_to_spaces(root) {
+        for (const el of selectElements(root, ".font-monospace.o_editor_banner .oe-tabs")) {
+            const spacesElement = document.createTextNode("\u00A0\u00A0\u00A0\u00A0");
+            el.replaceWith(spacesElement);
+        }
+    }
```
dmo-odoo, line 204: I was going to say that this is converting tabs to four spaces while I would have expected an override to change what happens when the user presses tab, but in fact this has the benefit of handling the case of converting from a paragrah with existing tabs to a banner so I think it's in fact better than the override I had in mind.

PR #248231, opened 2026-02-11
[FIX] stock{,_landed_costs},mrp_subcontracting: create aml correct va…

plha-odoo:
…lues landed cost sbc

**Problem:**
account move line created from a landed cost on a  subcontracted 
receipt do not take into account already out quantity.

**Steps to reproduce:**
- create a tracked product with avco auto category
- create a subcontracted bom for this product with no comp
- create and confirm a PO for 10 unit of this product 
with the same partner as the subcontractor of the bom
- validate the receipt
- create and validate a delivery for 4 unit of your product
- navigate to inventory/operations/adjustments/landed costs
- create a new landed cost
- select the receipt from the PO
- add a landed cost of 10$ and validate
- select the valuation smart button
- a 6$ svl was created (which is correct because 6 out 10 products
of the receipt are still in stock)
- click on the book widget to open the account move view

**Current behavior:**
Only two account move lines were created both with a value 
of 10
One crediting sotck interim received
On debiting stock valuation

**Expected behavior:**
4 extra account move lines (all with a value of 4) should have been 
created to compensate the out quantity like it is the case for 
non subcontracted product.
One debiting stock interim delivered
One crediting stock valuation 
One debiting expenses
One crediting stock interim delivered

**Cause of the issue:**
_is_in() will return false for the move of a subcontracted receipt
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/stock_landed_costs/models/stock_landed_cost.py#L185
This is wanted and happens because _should_be_valued()
will return true when called on the subcontracted location
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/stock_account/models/stock_move.py#L129

As a consequence, qty_out stays 0
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/stock_landed_costs/models/stock_landed_cost.py#L185-L186
and we do not append the values for the extra amls 
inside _create_account_move_line()
https://github.com/odoo/odoo/blob/7462ec423e8b106a5175a71ac009176d16cdf225/addons/stock_landed_costs/models/stock_landed_cost.py#L465-L466

**fix:**
if we make sure the the adjustment line is linked to the move 
of the MO instead of the move of the receipt, this problem does 
not happen because _is_in() returns true for the move of the MO. 
Also in this case we don't need _get_stock_valuation_layer_ids()
which was introduced by this PR https://github.com/odoo/odoo/pull/166107
to solve the same issue. 
That is because the move used in button_validate is the move linked
to the adjustment line, which will, after this fix, be the one of 
the MO, so we can directly take its stock valuation layers.

opw-5723126

### DIFF
--- a/addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
+++ b/addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
@@ -1,0 +1,18 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+from odoo.tools import OrderedSet
+
+
+class StockLandedCost(models.Model):
+    _inherit = 'stock.landed.cost'
+
+    def _get_targeted_move_ids(self):
+        res = super()._get_targeted_move_ids()
+        target_moves_ids = OrderedSet()
+        for move in res:
+            if move.is_subcontract:
+                target_moves_ids.update(move.move_orig_ids.ids)
+            else:
+                target_moves_ids.add(move.id)
+        return self.env['stock.move'].browse(target_moves_ids)
--- a/addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
+++ b/addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
@@ -20,7 +20,7 @@ class TestSubcontractingLandedCosts(TestMrpSubcontractingCommon):
             'order_line': [(0, 0, {
                 'name': self.finished.name,
                 'product_id': self.finished.id,
-                'product_uom_qty': 10,
+                'product_qty': 10,
                 'product_uom': self.finished.uom_id.id,
                 'price_unit': 10,
             })],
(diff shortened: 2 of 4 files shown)


addons/stock_landed_costs/models/stock_landed_cost.py
```diff
@@ -123,9 +123,9 @@ def button_validate(self):
             cost_to_add_byproduct = defaultdict(lambda: 0.0)
             cost_to_add_bylot = defaultdict(lambda: defaultdict(float))
             for line in cost.valuation_adjustment_lines.filtered(lambda line: line.move_id):
-                line_svls = line.move_id._get_stock_valuation_layer_ids()
```
ticodoo, line 126: Mm, you might break customizations by removing this method call. It will probably be safer to update the method in `mrp_subontracting_landed_costs`  to just return the `super` value (since it's the same result, although I don't know why it has a `ensure_one()` in it)

addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = 'stock.landed.cost'
+
+    def _get_targeted_move_ids(self):
+        res = super()._get_targeted_move_ids()
+        target_moves = self.env['stock.move']
+        for move in res:
+            if move.is_subcontract:
+                target_moves |= move.move_orig_ids
+            else:
+                target_moves |= move
```
ticodoo: same `OrderedSet` optimization comment

addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
```diff
@@ -130,3 +135,70 @@ def test_subcontracting_landed_cost_receipts_flow(self):
...
+        # validate move
+        receipt = po.picking_ids[0]
+        receipt.button_validate()
+
+        # simulate only partial quantity remaining
+        mo = receipt._get_subcontract_production()
+        mo.move_finished_ids.stock_valuation_layer_ids.remaining_qty = 7
+
+        # create a landed cost for the incoming picking
+        default_vals = self.env['stock.landed.cost'].default_get(list(self.env['stock.landed.cost'].fields_get()))
```
ticodoo: Same `default_get` comment. If you feel like it, you can fix it in the other test in the file since you're already modifying it + so hopefully people are less like to copy it in the future.

plha-odoo:
@robodoo up to saas-18.4
the issue is fixed by this PR https://github.com/odoo/odoo/pull/248469 from version 19 onward

PR #248233, opened 2026-02-11
[IMP] base_import, *: comprehensive UX cleanup and modernization

juliusc2066:
*: account, crm, hr, hr_timesheet, mass_mailing, mrp, product, project, purchase, sale, stock, web, base

This commit introduces a wide range of UI/UX improvements to the import records feature:

**UI/Layout:**
* **Revamped Initial View:** The import view prior to file selection has been redesigned for clarity (includes shortening the template import button text).
* **Breadcrumbs:** The file name now appears in the action breadcrumb instead of the side panel.
* **Input Styling:** Updated input fields to use rounded borders, matching the standard design.
* **Sticky Headers:** Column list headers are now sticky to improve usability on long lists.
* **Tooltips:** Moved several help tooltips to `<sup>` elements and updated their content.

**Workflow & Logic:**
* **Simplification:** Shortened overly verbose button names.
* **Sheet Selection:** Automatically hides the sheet selection input if the file contains only one sheet.
* **Advanced Options:** Removed the "Advanced" side panel section. The "Track History" option has been renamed and moved to the main section.
* **Separators:** Default float separators are now determined by the user's active language locale. Added "Other" option for CSV separators to customize the value beyond the default select options using an
additional input.
* **Batch options:** Batch customization has been reduced to the batch
size, got moved to the main section and is only accessible in debug mode.

**Field Mapping:**
* **Required Fields:** Required fields are now grouped at the top of the field selection dropdown.
* **External ID Warning:** Added a warning when the External ID column contains simple integers.
* **Error Messages:** Simplified the error message for missing required values.

task-5242240

### DIFF
--- a/addons/base_import/static/src/import_model.js
+++ b/addons/base_import/static/src/import_model.js
@@ -1,11 +1,12 @@
 import { useState } from "@web/owl2/utils";
+import { localization } from "@web/core/l10n/localization";
 import { _t } from "@web/core/l10n/translation";
 import { registry } from "@web/core/registry";
+import { groupBy, sortBy } from "@web/core/utils/arrays";
 import { checkFileSize, DEFAULT_MAX_FILE_SIZE } from "@web/core/utils/files";
+import { memoize } from "@web/core/utils/functions";
 import { useService } from "@web/core/utils/hooks";
 import { pick } from "@web/core/utils/objects";
-import { groupBy, sortBy } from "@web/core/utils/arrays";
-import { memoize } from "@web/core/utils/functions";
 import { session } from "@web/session";
 import { ImportBlockUI } from "./import_block_ui";
 import { BinaryFileManager } from "./binary_file_manager";
@@ -98,7 +99,6 @@ export class BaseImportModel {
         this.importOptionsValues = {
             ...this.formattingOptionsValues,
             advanced: {
-                reloadParse: true,
                 value: true,
             },
             has_headers: {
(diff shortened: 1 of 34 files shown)


odoo/orm/models.py
```diff
@@ -2529,10 +2529,10 @@ def _sql_error_to_message(self, exc: psycopg2.Error) -> str:
             # get the message from the object
             if message := cons.get_error_message(self, exc.diag):
                 return message
-        return self._sql_error_to_message_generic(exc)
+        return self._sql_error_to_message_generic(exc, from_import)
 
     @api.model
-    def _sql_error_to_message_generic(self, exc: psycopg2.Error) -> str:
+    def _sql_error_to_message_generic(self, exc: psycopg2.Error, from_import=False) -> str:
```
kmagusiak: Boolean flags are bad and I don't think we should have different messages. If the existing message is not helpful, rephrase it instead of adding another message for another context.

addons/base_import/static/src/import_model.js
```diff
@@ -485,7 +482,7 @@ export class BaseImportModel {
                 dryrun,
                 context: {
                     ...this.context,
-                    tracking_disable: this.importOptions.tracking_disable,
+                    tracking_disable: true,
```
aab-odoo: c'est bizarre

aab-odoo (approved):
robodoo r+

PR #193012, opened 2025-01-09
[IMP] stock, mrp, repair: forecast widget + MO form UX

R1D1CUL0US:
UX revamp of the MO form view, in particular the components tab.
UX revamp of the forecast widget.

task-4441085

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248234, opened 2026-02-11
[IMP] stock: Defensively specify element classes in tours.

amh-mw:
#### Description of the issue/feature this PR addresses:
This avoids unit test failures when the overly generic `.fa-list` matches an unexpected element instead of the "Open Move" button.

#### Current behavior before PR:
Tests fail.

#### Desired behavior after PR is merged:
Tests pass.

---

### DIFF
--- a/addons/stock/static/tests/tours/stock_picking_tour.js
+++ b/addons/stock/static/tests/tours/stock_picking_tour.js
@@ -19,7 +19,7 @@ registry.category("web_tour.tours").add('test_detailed_op_no_save_1', {  steps: 
         run: "click",
     },
     {
-        trigger: ".fa-list",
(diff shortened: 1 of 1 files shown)


Sehnde (approved):
As you said, it's unreasonable to expect us to test with 3rd party modules to make sure our tests still pass.
In this case however, the changes are minor, limited to 18.0 and 18.2 and make the tests more precise, so it should be ok.
Please make sure to adapt the forward-port in 18.2 since there are only 2 occurrences of `trigger: "fa-list"` left there.

Sehnde:
@robodoo up to saas-18.2

clesgow:
Note: Button was reworked for `18.3` in #193012, so this doesn't need to apply after that.

@robodoo r+ up to saas-18.2

PR #248235, opened 2026-02-11
[FIX] project: preserve privacy_visibility during share flow

mjvi-odoo:
Steps to reproduce:
- Create/open any project
- Set visibility to `Invited internal users and portal users`.
- Share project with a portal user
- Visibility changes to `All internal users and invited portal users`.

Cause:
- Previously, this behavior was implemented specifically for sharing the project from the cog menu. However, the `Share Project` option has now been removed from the cog menu.

- After introducing the new `invited_users` visibility option, the existing share logic was not updated to support this new value; as a result, the project's existing visibility is unintentionally overwritten.

Solution:
- Remove the hardcoded update of `privacy_visibility` in the `action_send_mail` method.

task-5924243

---

### DIFF
--- a/addons/project/wizard/project_share_wizard.py
+++ b/addons/project/wizard/project_share_wizard.py
@@ -154,7 +154,6 @@ class ProjectShareWizard(models.TransientModel):
         }
 
     def action_send_mail(self):
-        self.env['project.project'].browse(self.res_id).privacy_visibility = 'portal'
         result = {
             'type': 'ir.actions.client',
             'tag': 'display_notification',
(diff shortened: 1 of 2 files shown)


addons/project/wizard/project_share_wizard.py
```diff
@@ -154,7 +154,6 @@ def action_share_record(self):
         }
 
     def action_send_mail(self):
-        self.env['project.project'].browse(self.res_id).privacy_visibility = 'portal'
```
ppr-odoo, line 157: In the helpdesk_timesheet module, we are currently checking the project’s privacy visibility only when it is set to portal. For example, if a user is not following the project, but is following the ticket, the timesheet will not be visible to them. i have asked about this on discord to get others’ opinions. Let’s see what the feedback is.

mjvi-odoo: This issue is not related to this change, but it is an existing bug in `helpdesk_timesheet` regarding `privacy_visibility` checks.

We can fix it separately by creating an Enterprise PR, let’s first see what others suggest.

PR #248236, opened 2026-02-11
[FIX] purchase : correct analytic account PO count

MAGREMENT:
# How to reproduce
- Enable the analytic accounting in the settings
- Create a PO
- Add a PO line
- Set the Analytic Distribution of that PO line to an Analytic Account of your choice
- Confirm the PO
- Create a Vendor Bill from that PO and confirm the VB
- Go to the Analytic Account chosen before
- Change the Plan of that Analytic Account

# The problem
When the Plan is not set to the "Project Plan", the Purchase Orders smart button disappears

# Why
The Purchases Orders smart button is invisible if the variable purchase_order_count is equal to 0. That field is computed by a function that does a search with the following domain :

```py
[('order_line.invoice_lines.analytic_line_ids.account_id', '=', account.id)]
```

When we change the Plan of the Analytic Account, analytic_line_ids.account_id is set to NULL, so the search return nothing.

Why is that field set to NULL ?

Well, to reference its plan, an Analytic Line does not use a python-defined field. In fact, each time a new Analytic Plan is added to the database, a new column is added to the Analytic Line model. That column's name is x_plan{plan.id}_id or, for the specific case of the "Project Plan", it is account_id

When the Plan of an Analytic Account is changed, it takes every Analytic Line associated with that Plan and switch which column containing the id of the Analytic Account.

Take for exemple the following Analytic Line :

```
(account_id = NULL, x_plan2_id = NULL, x_plan3_id = 1)
```

When the associated Analytic Account's Plan is changed to the "Project Plan", it becomes :

```
(account_id = 1, x_plan2_id = NULL, x_plan3_id = NULL)
```

So, we need to adapt to search so that it uses the right plan's name.

opw-5897037

---

### DIFF
--- a/addons/purchase/models/analytic_account.py
+++ b/addons/purchase/models/analytic_account.py
@@ -13,13 +13,13 @@ class AccountAnalyticAccount(models.Model):
     def _compute_purchase_order_count(self):
         for account in self:
             account.purchase_order_count = self.env['purchase.order'].search_count([
-                ('order_line.invoice_lines.analytic_line_ids.account_id', '=', account.id)
+                ('order_line.invoice_lines.analytic_line_ids.' + account.plan_id._column_name(), '=', account.id)
             ])
 
     def action_view_purchase_orders(self):
         self.ensure_one()
         purchase_orders = self.env['purchase.order'].search([
-            ('order_line.invoice_lines.analytic_line_ids.account_id', '=', self.id)
+            ('order_line.invoice_lines.analytic_line_ids.' + self.plan_id._column_name(), '=', self.id)
         ])
         result = {
             "type": "ir.actions.act_window",
--- a/addons/purchase/tests/test_purchase_invoice.py
+++ b/addons/purchase/tests/test_purchase_invoice.py
@@ -770,6 +770,26 @@ class TestPurchaseToInvoice(TestPurchaseToInvoiceCommon):
         self.assertEqual(inv.invoice_line_ids[2].name, f"{pol_prod_product_in_name.name}", "When description contains the product name, the invoice line name should only be the description")
         self.assertEqual(inv.invoice_line_ids[3].name, f"{pol_prod_name_in_product.product_id.display_name}\n{pol_prod_name_in_product.name}", "When the product name contains the description, the invoice line name should be the product name and the description")
 
+    def test_compute_po_count_with_different_plan(self):
+        analytic_plan_1, analytic_plan_2 = self.env['account.analytic.plan'].create([
+            {'name': 'Plan 1'}, {'name': 'Plan 2'}
+        ])
+        analytic_account = self.env['account.analytic.account'].create({'name': 'Account', 'plan_id': analytic_plan_1.id})
+        purchase_order = self.init_purchase(partner=self.partner_a, products=[self.product_a])
+        purchase_order.order_line[0].analytic_distribution = {analytic_account.id: 100}
+        purchase_order.button_confirm()
+        purchase_order.order_line.qty_received = 1
+
+        bill = self.env['account.move'].browse(purchase_order.action_create_invoice()['res_id'])
+        bill.invoice_date = fields.Date.today()
+        bill.action_post()
+
+        self.assertEqual(analytic_account.purchase_order_count, 1)
+        analytic_account.plan_id = analytic_plan_2.id
+        analytic_account.invalidate_recordset(['purchase_order_count'])
+        self.assertEqual(analytic_account.purchase_order_count, 1)
+        self.assertEqual(analytic_account.action_view_purchase_orders()['domain'], [['id', 'in', purchase_order.ids]])
+
 
 @tagged('post_install', '-at_install')
 class TestInvoicePurchaseMatch(TestPurchaseToInvoiceCommon):


DjamelTouati (changes requested):
Hello,

For example, when you change the plan of the analytic account following your steps (from Project to Department), an error is triggered: “Whoa there! Making this change would wipe out your current data. Let’s avoid that, shall we?”. This is raised by the _update_accounts_in_analytic_lines function.

addons/purchase/models/analytic_account.py
```diff
@@ -11,16 +11,15 @@ class AccountAnalyticAccount(models.Model):
 
     @api.depends('line_ids')
     def _compute_purchase_order_count(self):
+        plan_column_names = self._get_all_plan_column_names()
+
         for account in self:
-            account.purchase_order_count = self.env['purchase.order'].search_count([
-                ('order_line.invoice_lines.analytic_line_ids.account_id', '=', account.id)
-            ])
+            account.purchase_order_count = self.env['purchase.order'].search_count(self._get_domains_for_po_search(plan_column_names, account.id))
```
DjamelTouati:
```suggestion
account.purchase_order_count = len(account.line_ids.move_line_id.purchase_line_id.order_id)
```

DjamelTouati (changes requested):
When a database is created, only the default “Project” analytic plan is mapped to the `account_id` field.  
When you create additional analytic plans, they are actually implemented as Studio custom fields through a workaround (hack).

Example:
- Create a new analytic account “AA - test” under a new plan “Plan - test”
- Create a storable product “P1”
- Create a purchase order:
  - Product: “P1”
  - Analytic distribution: AA - test 100%
  - Confirm the PO, create the bill, and post it

- Go to Accounting > Analytic Items:
  - You will see that a new analytic line is created, but the custom field `x_plan_1` (Plan - test) is filled with “AA - test”
  - The standard field `account_id` is empty

Therefore, when you open the analytic account, it is normal that no lines are found.  
The current logic checks invoice lines linked to the purchase order lines and expects the analytic line to have `account_id` equal to the analytic account. In this case, it is not true, because the value is stored in `x_plan_1` instead.

We could theoretically check `x_plan_1`, but that would not be scalable. If there are 100 analytic plans, we would need a domain with 100 different fields, and performing searches in loops is inefficient.

Instead, we can simplify the logic by relying on the existing relational links:
```diff
diff --git a/addons/purchase/models/analytic_account.py b/addons/purchase/models/analytic_account.py
index 951c29b2dc12..223fd3477e4c 100644
--- a/addons/purchase/models/analytic_account.py
+++ b/addons/purchase/models/analytic_account.py
@@ -12,9 +12,7 @@ class AccountAnalyticAccount(models.Model):
     @api.depends('line_ids')
     def _compute_purchase_order_count(self):
         for account in self:
-            account.purchase_order_count = self.env['purchase.order'].search_count([
-                ('order_line.invoice_lines.analytic_line_ids.account_id', 'in', account.ids)
-            ])
+            account.purchase_order_count = len(account.line_ids.move_line_id.purchase_line_id.order_id)

```

The same adaptation should be applied in `action_view_purchase_orders`.

addons/purchase/models/analytic_account.py
```diff
@@ -13,13 +13,13 @@ class AccountAnalyticAccount(models.Model):
     def _compute_purchase_order_count(self):
         for account in self:
             account.purchase_order_count = self.env['purchase.order'].search_count([
-                ('order_line.invoice_lines.analytic_line_ids.account_id', '=', account.id)
+                ('order_line.invoice_lines.analytic_line_ids.' + self.plan_id._column_name(), '=', account.id)
```
Sehnde:
```suggestion
                ('order_line.invoice_lines.analytic_line_ids.' + account.plan_id._column_name(), '=', account.id)
```
otherwise you risk a singleton error if for example, a user adds the purchase_order_count to the list view

addons/purchase/tests/test_analytic_account.py
```diff
@@ -0,0 +1,32 @@
...
+            ]
+        })
+        purchase_order.button_confirm()
+        purchase_order.order_line.qty_received = 1
+
+        bill = self.env['account.move'].browse(purchase_order.action_create_invoice()['res_id'])
+        bill.invoice_date = datetime.today()
+        bill.action_post()
+
+        self.assertEqual(analytic_account.with_context(analytic_plan_id=analytic_plan_1.id).purchase_order_count, 1)
```
Sehnde: Not sure you need the added context here, the test works without it, and if you use my example of adding the purchase_order_count to the list view, there's no context anyway. The only place I see it used is in `_compute_auto_account` which is not relevant here.

However, you need to invalidate analytic_account with `invalidate_recordset` to force a recompute, or just call the compute function again. Otherwise, the test will simply recheck the same value of 1 for plan_1, which in this is case is a false positive. If you create a new PO with a new bill between both `assertEqual`, you'll see that `purchase_order_count` will still be at 1 instead of 2.

Sehnde (changes requested):
As ARM said, please use an existing file for your test.
I think test_purchase_invoice.py would be appropriate. You can simply add your test to `TestPurchaseToInvoice`

addons/purchase/tests/test_purchase_invoice.py
```diff
@@ -770,6 +770,31 @@ def test_invoice_line_name_has_product_name(self):
...
+            ]
+        })
+        purchase_order.button_confirm()
+        purchase_order.order_line.qty_received = 1
+
+        bill = self.env['account.move'].browse(purchase_order.action_create_invoice()['res_id'])
+        bill.invoice_date = datetime.today()
+        bill.action_post()
+
+        self.assertEqual(analytic_account.purchase_order_count, 1)
```
Sehnde, line 787:
```suggestion
        self.assertEqual(analytic_account.purchase_order_count, 1)
        self.assertEqual(analytic_account.action_view_purchase_orders()['domain'], [['id', 'in', purchase_order.ids]])
```
You can also test the action, to make sure it correctly includes the PO in its domain.
Not sure if we usually check on action domains but since your fix also touches on that

Sehnde (commented):
Just a small nitpick, otherwise LGTM

addons/purchase/tests/test_purchase_invoice.py
```diff
@@ -770,6 +770,32 @@ def test_invoice_line_name_has_product_name(self):
...
+        analytic_account = self.env['account.analytic.account'].create({'name': 'Account', 'plan_id': analytic_plan_1.id})
+        purchase_order = self.env['purchase.order'].create({
+            'partner_id': self.partner_a.id,
+            'order_line': [
+                Command.create({
+                    'product_id': self.product_a.id,
+                    'analytic_distribution': {analytic_account.id: 100},
+                }),
+            ]
+        })
```
Sehnde: just realized, you can replace all that with `init_purchase` since we're now in TestPurchaseToInvoice.
Well, not the the analyic distribution but that's still only 2 lines vs 9

Sehnde (approved):

addons/purchase/tests/test_purchase_invoice.py
```diff
@@ -770,6 +770,26 @@ def test_invoice_line_name_has_product_name(self):
...
+            {'name': 'Plan 1'}, {'name': 'Plan 2'}
+        ])
+        analytic_account = self.env['account.analytic.account'].create({'name': 'Account', 'plan_id': analytic_plan_1.id})
+        purchase_order = self.init_purchase(partner=self.partner_a, products=[self.product_a])
+        purchase_order.order_line[0].analytic_distribution = {analytic_account.id: 100}
+        purchase_order.button_confirm()
+        purchase_order.order_line.qty_received = 1
+
+        bill = self.env['account.move'].browse(purchase_order.action_create_invoice()['res_id'])
+        bill.invoice_date = datetime.today()
```
clesgow:
```suggestion
        bill.invoice_date = fields.Date.today()
```
No need for import, and `account_move.invoice_date` is a `Date`, not a `Datetime`.

clesgow:
Thanks for the fix!
@robodoo r+

PR #248237, opened 2026-02-11
[FIX] point_of_sale: Refunds should have negative amount in the report

pedrambiria:
Before this commit, the refund orders had positive amount in the report, which is not correct.

opw-5821547

---


adgu-odoo (commented):
robodoo r+

PR #248241, opened 2026-02-11
[IMP] mail: improve threads (subchannel) pannel

phenix-factory:
image

Task-5867464 point 46

### DIFF
--- a/addons/mail/static/src/discuss/core/common/discuss_channel_model.js
+++ b/addons/mail/static/src/discuss/core/common/discuss_channel_model.js
@@ -13,6 +13,8 @@ import {
 import { formatList } from "@web/core/l10n/utils";
 import { imageUrl } from "@web/core/utils/urls";
 
+const { DateTime } = luxon;
+
 /** @import { AwaitChatHubInit } from "@mail/core/common/chat_hub_model" */
 
 export class DiscussChannel extends Record {
@@ -202,6 +204,11 @@ export class DiscussChannel extends Record {
         }
         return this.channel_member_ids.filter(({ persona }) => persona?.notEq(this.store.self));
     }
+    get createDateSimple() {
+        return this.create_date
+            ?.toLocaleString(DateTime.TIME_SIMPLE, { locale: user.lang })
+            .replace(" ", " "); // so that AM/PM are properly wrapped
+    }
     discuss_category_id = fields.One("discuss.category", {
         inverse: "channel_ids",
     });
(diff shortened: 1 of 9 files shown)


addons/mail/static/src/discuss/core/common/discuss_channel_model.js
```diff
@@ -202,6 +204,13 @@ export class DiscussChannel extends Record {
         }
         return this.channel_member_ids.filter(({ persona }) => persona?.notEq(this.store.self));
     }
+    get createDateSimple() {
+        return this.create_date
+            ?.toLocaleString(DateTime.TIME_SIMPLE, {
+                locale: user.lang,
+            })
+            .replace(" ", " "); // so that AM/PM are properly wrapped
+    }
```
tsm-odoo, line 211:
```suggestion
    get createDateSimple() {
        return this.create_date
            ?.toLocaleString(DateTime.TIME_SIMPLE, { locale: user.lang })
            .replace(" ", " "); // so that AM/PM are properly wrapped
    }
```

tsm-odoo (commented):
What about the sub-sections part?

tsm-odoo:
> It might be useful to create a task for it as it's a new feature, and take opinions before

I agree, Im not saying we should or shouldn't do it but just making sure the note on the side of the 100 point long excalidraw doesnt get lost in oblivion^^

tsm-odoo (approved):
@robodoo r+ delegate+

addons/mail/static/src/discuss/core/public_web/sub_channel_preview.xml
```diff
@@ -5,25 +5,48 @@
         <div class="o-mail-SubChannelPreview btn btn-light-subtle d-flex border border-secondary mx-1 px-2 py-1 rounded-2 align-items-center" role="button" t-att-class="{'o-pe-1_5': env.inMessage}" t-attf-class="{{ props.class }}" t-on-click="onClick">
             <div class="d-flex flex-column w-100 bg-inherit">
                 <div class="d-flex w-100 align-items-baseline">
-                    <span class="o-fw-600 text-truncate small" t-esc="props.channel.displayName"/>
+                    <span class="o-mail-SubChannel-name o-fw-600 text-truncate small" t-esc="props.channel.displayName"/>
```
alexkuhn, line 8: Should be named `o-mail-SubChannelPreview-name`.

The naming of "SubChannelPreview" is important to match component name, but also because this can be misleading with the discuss sidebar item for channel's thread.

phenix-factory: https://github.com/odoo/odoo/pull/248971

PR #248242, opened 2026-02-11
[FIX] web: prevent useless loading effect mutations

Dj0ulo:
__Before commit__
When using `makeButtonHandler`, a loading effect is applied to the
button if the handler takes more than 400ms to execute. However, if
the handler finishes sooner, the effect is still briefly added and
removed, triggering a DOM mutation.

If a button is clicked and the website builder opens immediately
after, this mutation may be recorded in the current history step
despite being unnecessary.

This causes the `design-themes` tour `theme_menu_hierarchies` to fail
in a non-deterministic way with the following warning:
"should not have any 'characterData', 'remove' or 'add' mutations in
current step when you update the selection"

__Fix__
Ensure the loading effect is never added if the handler execution
time is shorter than the debounce duration.

This fix is applied to both `makeButtonHandler` functions in the
codebase. The test case is only added for the interaction use case,
as the other instance resides in legacy code.

runbot-229803

### DIFF
--- a/addons/web/static/src/legacy/js/public/minimal_dom.js
+++ b/addons/web/static/src/legacy/js/public/minimal_dom.js
@@ -80,11 +80,11 @@ export function makeButtonHandler(fct, preventDefault, stopPropagation, stopImme
     fct = makeAsyncHandler(fct, preventDefault, stopPropagation, stopImmediatePropagation);
 
     return function (ev) {
-        const result = fct.apply(this, arguments);
+        const handlerResult = fct.apply(this, arguments);
 
         const buttonEl = ev.target.closest(BUTTON_HANDLER_SELECTOR);
         if (!(buttonEl instanceof HTMLElement)) {
-            return result;
+            return handlerResult;
         }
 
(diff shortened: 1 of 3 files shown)


addons/web/static/src/legacy/js/public/minimal_dom.js
```diff
@@ -80,24 +80,32 @@ export function makeButtonHandler(fct, preventDefault, stopPropagation, stopImme
...
         }
 
         // Disable the button for the duration of the handler's action
         // or at least for the duration of the click debounce. This makes
         // a 'real' debounce creation useless. Also, during the debouncing
         // part, the button is disabled without any visual effect.
         buttonEl.classList.add("pe-none");
-        new Promise(resolve => setTimeout(resolve, DEBOUNCE)).then(() => {
+        // Pending flag as an object to be able to compare promises by reference
+        const pending = {};
```
FrancoisGe: let flag = false;
setTimeout(() => flag = true; resolve(pending)

PR #248244, opened 2026-02-11
[MOV] http: move session-related functions from request

thle-odoo:
[x] move session related-functions out of request
[x] move router related-functions out of request

Task-6031124

### DIFF
--- a/odoo/http/retrying.py
+++ b/odoo/http/retrying.py
@@ -64,7 +64,10 @@ def retrying[T](func: Callable[[], T], env: Environment) -> T:
                 env.transaction.reset()
                 env.registry.reset_changes()
                 if request:
-                    request.session = request._get_session_and_dbname()[0]
+                    # We need to reset the `session` attribute of `request`
+                    # which may have been modified during the transaction.
+                    # The `dbname` remains the same (consistent with the session).
+                    _set_session_and_dbname(request)
                     # Rewind files in case of failure
                     for filename, file in request.httprequest.files.items():
                         if hasattr(file, "seekable") and file.seekable():
@@ -113,4 +116,5 @@ def retrying[T](func: Callable[[], T], env: Environment) -> T:
     return result
 
 
-from .requestlib import request  # noqa: E402
+# ruff: noqa: E402
+from .router import _set_session_and_dbname, request
--- a/odoo/http/router.py
+++ b/odoo/http/router.py
@@ -31,7 +31,6 @@ from odoo.modules.module import (
     initialize_sys_path,
 )
 from odoo.modules.registry import Registry
-from odoo.service.server import thread_local
 from odoo.tools import config, file_path, real_time
 from odoo.tools.misc import submap
 
@@ -245,7 +244,7 @@ class Application:
             del current_thread.dbname
         if hasattr(current_thread, 'uid'):
             del current_thread.uid
-        thread_local.rpc_model_method = ''
+        current_thread.rpc_model_method = ''
 
         if config['proxy_mode'] and environ.get("HTTP_X_FORWARDED_HOST"):
             # The ProxyFix middleware has a side effect of updating the
--- a/odoo/http/router.py
+++ b/odoo/http/router.py
@@ -1,17 +1,24 @@
 from __future__ import annotations
 
+import datetime as dt
 import functools
 import logging
 import re
 import threading
 import typing
+from contextlib import nullcontext
 from os.path import join as opj
 from urllib.parse import urlparse
 
 import psycopg2
 import werkzeug.routing
 from psycopg2.errors import OperationalError, ReadOnlySqlTransaction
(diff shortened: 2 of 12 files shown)


mwath (approved):
👍

odoo/http/session.py
```diff
@@ -601,6 +602,78 @@ def session_store():
     return SessionStore(path=config.session_dir)
 
 
+def set_session_and_dbname(request: Request) -> None:
```
Julien00859: have this one private, and inside `requestlib`

Julien00859:
@thle-odoo up

odoo/http/session.py
```diff
@@ -608,6 +609,42 @@ def session_store():
     return SessionStore(path=config.session_dir)
 
 
+def save_session(request: Request, env: Environment | None = None) -> None:
```
kmagusiak, line 612: Why not put it on the session? `def save(self, env: Environment)`
This avoids adding more weird imports.

odoo/http/router.py
```diff
@@ -447,6 +454,100 @@ def serve_db(request: Request) -> Response:
...
+                request.session['profile_session'] = None
+
+    return nullcontext()
+
+
+def _set_session_and_dbname(request: Request) -> None:
+    sid = request.httprequest.cookies.get('session_id', '')
+    session = session_store().get(sid, keep_sid=True)
+
+    for key, val in get_default_session().items():
```
kmagusiak, line 495: `get_default_session` could be removed and the contents could be moved inside `SessionStore.new` (or `Session.__init__`).

odoo/http/retrying.py
```diff
@@ -64,7 +64,7 @@ def retrying[T](func: Callable[[], T], env: Environment) -> T:
                 env.transaction.reset()
                 env.registry.reset_changes()
                 if request:
-                    request.session = request._get_session_and_dbname()[0]
+                    _set_session_and_dbname(request)
```
kmagusiak, line 70: Could you explain this one? (comments in the code)
The session and the database should already be set on the request.

odoo/http/session.py
thle-odoo: The session is an exposed object. We prefer to provide access only to the `MutableMapping` API (and a few useful properties). The rest is handled in standalone functions. This decision was made to meet specific requirements.

odoo/http/router.py
thle-odoo: The purpose of this PR is to move certain functions. If we want to perform this initialization directly in `__new__` or `__init__`, we need to take precautions for cases where a session without these `setdefault` is expected (which I think should be tested in a separate PR).

odoo/http/session.py
kmagusiak: Then `def save__` and we are fine, aren't we?

odoo/http/router.py
kmagusiak: I agree that can be done in master. In all cases, I found we just create a new session and put these values into it. If we need these values, we can create a session object which won't be saved.

odoo/http/session.py
kmagusiak: As discussed, we can provide a function on the session store to avoid additional imports.

Julien00859: The goal of this PR is to only expose stuff that developers typically need, and to make it difficult to use any low-level stuff. They don't need to access `save()` as the http-stack is responsible of saving the session. For the rare cases where controllers need to manually save the session, they can afford to make an extra import. There's no even 10 cases where we need that import.

odoo/http/retrying.py
Julien00859: The session is a mutable object, we need to reload it from the filesystem to make sure no mutation persists in between retries.

odoo/http/router.py
Julien00859: For another PR

Julien00859 (approved):
@robodoo r+

odoo/http/session.py
kmagusiak: So the merge settled because of circular imports... I think there is enough cases to put the function on the store.

thle-odoo: A circular import error occurs when using the `test_module_operations.py` script as the main module.

Placing imports at the end of the file makes it easy to manage imports between "two" modules, but this becomes more complex when dealing with external packages.

As the traceback shows, the problematic situation is:
- router module uses service package
- service package uses retrying module
- retrying module uses router module

Wouldn't it make sense to handle this circular import directly by importing `odoo.service` in the `__init__` of the `odoo/http` package ? (To keep these functions standalone as originally planned)

kmagusiak: Yes, there are 2 possible solutions.
- import all sub-modules in __init__
- move the save function to the session store

The second one is cleaner because it does not introduce additional complexity in imports which you did not spot.
Please move the function into the session store.

Julien00859:
@robodoo r-

PR #248255, opened 2026-02-11
fix oddball logs on install

xmo-odoo:


xmo-odoo:
@robodoo r+

addons/mail/__manifest__.py
```diff
@@ -41,12 +41,13 @@
...
-----------------------------------------------------------------------------------------------------------
-    * CRM Leads/Opportunities
-    * CRM Claims
-    * Project Issues
-    * Project Tasks
-    * Human Resource Recruitment (Applicants)
+
+* CRM Leads/Opportunities
+* CRM Claims
+* Project Issues
```
tde-banana-odoo, line 47: Glad we fixed the display of modules removed before we even had Farm 2 :disguised_face: :D

PR #248256, opened 2026-02-11
[FIX] core: pregenerate lazy loaded bundles

pparidans:
tl;dr: look for lazy loaded bundles and add them to the bundles' list to pregenerate before running tests to avoid generation on the fly.

### DIFF
--- a/odoo/addons/base/models/ir_qweb.py
+++ b/odoo/addons/base/models/ir_qweb.py
@@ -368,6 +368,7 @@ structure.
 import base64
 import contextlib
 import fnmatch
+import glob
 import io
 import logging
 import math
@@ -385,16 +386,17 @@ from collections.abc import Sized, Mapping
 from itertools import count, chain
 from lxml import etree
 from dateutil.relativedelta import relativedelta
+from os.path import join as opj
 from psycopg2.extensions import TransactionRollbackError
 
 from odoo import api, models, tools
-from odoo.modules import registry
+from odoo.modules import get_module_path, registry
 from odoo.tools import config, safe_eval, pycompat
 from odoo.tools.constants import SUPPORTED_DEBUGGER, EXTERNAL_ASSET
 from odoo.tools.safe_eval import assert_valid_codeobj, _BUILTINS, to_opcodes, _EXPR_OPCODES, _BLACKLIST
 from odoo.tools.json import scriptsafe
 from odoo.tools.lru import LRU
-from odoo.tools.misc import str2bool
+from odoo.tools.misc import file_open, str2bool
 from odoo.tools.image import image_data_uri, FILETYPE_BASE64_MAGICWORD
 from odoo.http import request
 from odoo.tools.profiler import QwebTracker
@@ -2662,7 +2664,11 @@ class IrQWeb(models.AbstractModel):
         """
         Returns the list of bundles to pregenerate.
         """
+        js_views_bundles, css_views_bundles = self._get_bundles_from_views()
+        lazy_bundles = self._get_lazy_bundles_from_js()
+        return (js_views_bundles | lazy_bundles, css_views_bundles | lazy_bundles)
 
+    def _get_bundles_from_views(self):
         views = self.env['ir.ui.view'].search([('type', '=', 'qweb'), ('arch_db', 'like', 't-call-assets')])
         js_bundles = set()
         css_bundles = set()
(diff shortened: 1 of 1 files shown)


odoo/addons/base/models/ir_qweb.py
```diff
@@ -2677,6 +2683,20 @@ def _get_bundles_to_pregenarate(self):
...
+        lazy_bundle_regex = re.compile(r'loadBundle\([\"\'](.+)[\"\']\)')
+        bundles = set()
+        for modroot in map(get_module_path, modules):
+            for root, _, fnames in os.walk(modroot):
+                fnames = fnmatch.filter([opj(root, n) for n in fnames], '**/static/src/**/*.js')
+                for fname in fnames:
+                    fcontent = file_open(fname).read()
+                    if match := lazy_bundle_regex.search(fcontent):
+                        bundle_name, = match.groups()
+                        bundles.add(bundle_name)
```
xmo-odoo:
```suggestion
                        bundles.add(match[1])
```

odoo/addons/base/models/ir_qweb.py
```diff
@@ -2677,6 +2683,20 @@ def _get_bundles_to_pregenarate(self):
...
         return (js_bundles, css_bundles)
 
+    def _get_lazy_bundles_from_js(self):
+        modules = self.env['ir.module.module'].search([('state', '=', 'installed')]).mapped('name')
+        lazy_bundle_regex = re.compile(r'loadBundle\([\"\'](.+)[\"\']\)')
+        bundles = set()
+        for modroot in map(get_module_path, modules):
+            for root, _, fnames in os.walk(modroot):
+                fnames = fnmatch.filter([opj(root, n) for n in fnames], '**/static/src/**/*.js')
+                for fname in fnames:
```
xmo-odoo: I'm pretty sure that's a hand-rolled version of something along the lines of
```suggestion
            for fname in glob.iglob('**/static/src/**/*.js', root_dir=modroot, recursive=True):
```

odoo/addons/base/models/ir_qweb.py
```diff
@@ -2677,6 +2683,20 @@ def _get_bundles_to_pregenarate(self):
...
+                    fcontent = file_open(fname).read()
```
xmo-odoo:
```suggestion
                    with file_open(fname) as f:
                        fcontent = f.read()
```

odoo/addons/base/models/ir_qweb.py
kmagusiak: Nitpick: `glob.glob('**/static/src/**/*.js', root_dir=modroot, recursive=True)`

odoo/addons/base/models/ir_qweb.py
```diff
@@ -386,15 +387,16 @@
 from lxml import etree
 from dateutil.relativedelta import relativedelta
 from psycopg2.extensions import TransactionRollbackError
+from os.path import join as opj
 
 from odoo import api, models, tools
-from odoo.modules import registry
+from odoo.modules import registry, get_module_path
```
kmagusiak: Nitpick: sort the imports

odoo/addons/base/models/ir_qweb.py
```diff
@@ -2677,6 +2683,20 @@ def _get_bundles_to_pregenarate(self):
                     css_bundles.add(asset)
         return (js_bundles, css_bundles)
 
+    def _get_lazy_bundles_from_js(self):
+        modules = self.env['ir.module.module'].search([('state', '=', 'installed')]).mapped('name')
+        lazy_bundle_regex = re.compile(r'loadBundle\([\"\'](.+)[\"\']\)')
```
xmo-odoo: The regex has a bunch of weaknesses:

1. doesn't account for template strings which could be used instead of single or double quoted strings
2. in backtracking engines, `.` is greedy, so on something like `loadBundle("foobar").then(_ => something("xyz"))` things the bundle name is `foobar").then(_ => something("xyz`
3. in fact because it doesn't check the quotes against one another, `loadBundle("foobar").then(_ => something('xyz'))` would also match very oddly
4. it does not seem to leave room for the options parameter so it's going to miss even otherwise static `loadBundle` (missing dynamic ones is probably fair), using semgrep for a quick check matches 68 loadBundle calls in the codebase, of which 17 have an options.

Also it's not necessary (and is possibly incorrect) to escape dquotes inside a single-quoted string.

pparidans: Would it be better with something like ?
```py
r'loadBundle\(["\'`]([\w\.-]+)["\'`]\)'
```
(1. add support for template strings, 2. no `.`, 3. still doesn't check for matching quotes, 4. only matches purely static `loadBundle(...)` calls)

pparidans: up ? @xmo-odoo

xmo-odoo: Seems a lot better. You can implement matching quotes via backrefs e.g. 

    r'loadBundle\((["\'`])([\w\.-]+)\1\)'

Might also make sense to set `re.ASCII` to avoid fancy non-ascii bundle names?

pparidans: Good idea!
Even maybe with :
```
r'\bloadBundle\((["\'`])([\w\.-]+)\1\)'
```
to avoid also matching `preloadBundle(...)` calls ?

xmo-odoo: sure

PR #248273, opened 2026-02-11
[IMP] point_of_sale, *: Improve button consistency in navbar

xlu-odoo:
*: pos_hr, pos_restaurant

Before this commit, the POS navbar had inconsistent button styling where navigation buttons (Register, Orders, Tables, Booking) used different color schemes and border treatments than order management buttons. The secondary buttons had borders added in a recent PR to make the button more visible in light-challenging situations, but the main navigation buttons remained borderless as btn-light. Also, the order overflow indicator showed as a simple caret, providing no information about how many hidden orders existed.

This commit fixes these inconsistencies with a better button hierarchy.

Lastly, the overflow indicator now shows "+N" instead of a caret, clearly indicating how many orders are hidden. The layout calculation was improved to ensure all navigation elements remain visible and accessible.

requires: https://github.com/odoo/enterprise/pull/107166

task-5905481

---

### DIFF
--- a/addons/point_of_sale/static/src/app/components/list_container/list_container.js
+++ b/addons/point_of_sale/static/src/app/components/list_container/list_container.js
@@ -37,17 +37,22 @@ export class ListContainer extends Component {
         class: "",
     };
     static template = xml`
-        <div class="d-flex flex-grow-1" t-attf-class="{{props.class}}" t-att-class="{'overflow-hidden': !isUiSmall}">
-            <button t-if="props.onClickPlus" class="list-plus-btn btn btn-secondary btn-lg me-1 my-2" t-on-click="props.onClickPlus">
+        <div class="d-flex gap-1 align-items-center flex-grow-1" t-attf-class="{{props.class}}" t-att-class="{'overflow-hidden': !isUiSmall}">
+            <button t-if="props.onClickPlus" class="list-plus-btn btn btn-secondary btn-lg flex-shrink-0 lh-lg" t-on-click="props.onClickPlus">
                 <i class="fa fa-fw fa-plus-circle" aria-hidden="true"/>
             </button>
-            <button t-if="this.sizing.isLarger or props.forceSmall" t-on-click="toggle"
-                class="btn btn-secondary mx-1 fa fa-caret-down my-2" />
-            <div class="overflow-hidden w-100 position-relative py-2">
-                <div t-ref="container" class="list-container-items d-flex w-100">
-                    <div t-if="!props.forceSmall" t-foreach="props.items" t-as="item" t-key="item_index" t-att-class="{'invisible': shouldBeInvisible(item_index)}">
+            <span t-if="props.onClickPlus" class="navbar-separator mx-1"/>
+            <div class="overflow-hidden flex-grow-1">
+                <div t-ref="container" class="list-container-items d-flex align-items-center gap-1">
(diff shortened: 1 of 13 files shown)


addons/point_of_sale/static/src/app/components/list_container/list_container.js
```diff
@@ -37,17 +37,22 @@ export class ListContainer extends Component {
...
+            <button t-if="props.onClickPlus" class="list-plus-btn btn btn-secondary btn-lg flex-shrink-0 rounded-circle lh-lg" t-on-click="props.onClickPlus">
                 <i class="fa fa-fw fa-plus-circle" aria-hidden="true"/>
             </button>
-            <button t-if="this.sizing.isLarger or props.forceSmall" t-on-click="toggle"
-                class="btn btn-secondary mx-1 fa fa-caret-down my-2" />
-            <div class="overflow-hidden w-100 position-relative py-2">
-                <div t-ref="container" class="list-container-items d-flex w-100">
-                    <div t-if="!props.forceSmall" t-foreach="props.items" t-as="item" t-key="item_index" t-att-class="{'invisible': shouldBeInvisible(item_index)}">
+            <span class="navbar-separator mx-1"/>
+            <div class="overflow-hidden">
```
adgu-odoo: By removing `flex-grow-1` on the parent div and `w-100` on this div, the div is never big enough to display items (at least never during my tests), it always display only the expend button. 

Image

addons/pos_restaurant/static/src/app/components/popup/edit_order_name_popup/edit_order_name_popup.xml
```diff
@@ -13,7 +13,7 @@
                     <t t-set="order" t-value="scope.item" />
                     <div class="position-relative">
                         <button t-esc="order.getName()"
-                            class="btn btn-lg btn-secondary text-truncate me-1"
+                            class="btn btn-lg btn-secondary me-1 rounded-pill px-3 lh-lg text-truncate"
```
adgu-odoo: is the rounded-pill wanted here, when in the navbar, they are rectangle so I don't know if we wanna make them rounded here but ok if it's wanted

addons/point_of_sale/static/src/app/components/list_container/list_container.js
```diff
@@ -37,17 +37,22 @@ export class ListContainer extends Component {
...
+        <div class="d-flex gap-1 align-items-center" t-attf-class="{{props.class}}" t-att-class="{'overflow-hidden': !isUiSmall}">
+            <button t-if="props.onClickPlus" class="list-plus-btn btn btn-secondary btn-lg flex-shrink-0 rounded-circle lh-lg" t-on-click="props.onClickPlus">
                 <i class="fa fa-fw fa-plus-circle" aria-hidden="true"/>
             </button>
-            <button t-if="this.sizing.isLarger or props.forceSmall" t-on-click="toggle"
-                class="btn btn-secondary mx-1 fa fa-caret-down my-2" />
-            <div class="overflow-hidden w-100 position-relative py-2">
-                <div t-ref="container" class="list-container-items d-flex w-100">
-                    <div t-if="!props.forceSmall" t-foreach="props.items" t-as="item" t-key="item_index" t-att-class="{'invisible': shouldBeInvisible(item_index)}">
+            <span class="navbar-separator mx-1"/>
```
adgu-odoo:
```suggestion
            <span t-if="props.onClickPlus" class="navbar-separator mx-1"/>
```
Otherwise you always show it and when the plus button is not shown, it makes something strange where there is a vertical bar that might not make sense

PR #248282, opened 2026-02-11
[IMP] project_timesheet_holidays: make public holiday timesheets non-editable

aurb9:
Timesheet entries generated from public holidays are currently editable by users with minimal rights. Although these entries appear greyed out, they can still be modified, which can lead to inconsistencies in reporting and payroll calculations. This behavior is inconsistent with other time-off–related entries, which are non-editable by default.

Hence, this commit makes global time off timesheet entries non-editable in all views, except the grid, which instead displays a user error if they try to edit the timesheet amount.

task-5360222

### DIFF
--- a/addons/project_timesheet_holidays/models/account_analytic.py
+++ b/addons/project_timesheet_holidays/models/account_analytic.py
@@ -14,6 +14,9 @@ class AccountAnalyticLine(models.Model):
 
     _timeoff_timesheet_idx = models.Index('(task_id) WHERE (global_leave_id IS NOT NULL OR holiday_id IS NOT NULL) AND project_id IS NOT NULL')
 
+    def _is_readonly(self):
+        return super()._is_readonly() or self.holiday_id or self.global_leave_id
+
     def _get_redirect_action(self):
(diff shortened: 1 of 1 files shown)


addons/project_timesheet_holidays/models/account_analytic.py
```diff
@@ -31,7 +34,7 @@ def _get_redirect_action(self):
     @api.ondelete(at_uninstall=False)
     def _unlink_except_linked_leave(self):
         if any(line.global_leave_id for line in self):
-            raise UserError(_('You cannot delete timesheets that are linked to global time off.'))
+            raise UserError(_('Timesheets linked to public holidays cannot be deleted.'))
```
deneuvillem:
```suggestion
            raise UserError(self.env._('Timesheets linked to public holidays cannot be deleted.'))
```
We can use this approach now :)

deneuvillem: Same comment for the other error messages you modified

deneuvillem (commented):
Thanks for your imp! Just left a small comment for you 🎖️

PR #248283, opened 2026-02-11
[FIX] website: allow copyright bg color when footer has no bg color

romo-odoo:
Before this commit, a css error would happen when the user tried to change the copyright background color if the footer had no background color.

This was due to $-footer-color not having a fallback value when neither o-color('footer-custom') nor o-color('footer') was defined.

This commit adds a fallback value to fix the issue.

task-5452457


bvr-odoo (approved):
@romo-odoo LGTM, could you rebase on 17.0 + mark the PR as ready and then ping me
Thanks 👍🏼

PR #248289, opened 2026-02-11
[FIX] api_doc: Clean tooltips

PiaCOS:
In api_doc, the method parameters tooltips were broken.

This commit fixes them by:
- aligning the text to the left
- improving the hideTooltip behavior
- preventing `undefined` tooltip content from being displayed

task-5246075

### DIFF
--- a/addons/api_doc/static/src/components/doc_method.js
+++ b/addons/api_doc/static/src/components/doc_method.js
@@ -27,7 +27,7 @@ export class DocMethod extends Component {
                     value: "annotation" in options ? options.annotation : "-",
                 },
                 { type: TABLE_TYPES.Code, value: this.getDefaultValue(options) },
-                { type: TABLE_TYPES.Tooltip, value: markup(options.doc) || "" },
+                { type: TABLE_TYPES.Tooltip, value: options.doc ? markup(options.doc) : "" },
             ]),
         };
     }
--- a/addons/api_doc/static/src/components/doc_table.js
+++ b/addons/api_doc/static/src/components/doc_table.js
@@ -17,16 +17,16 @@ export class DocTable extends Component {
     setup() {
         this.subTableRef = useRef("subTableRef");
         this.tooltipRef = useRef("tooltipRef");
-        this.hideTooltipTimeout = null;
         this.state = useState({
             sortBy: 0,
             sortOrder: "desc",
             subTable: undefined,
             tooltipContent: "",
             tooltipStyle: "",
-            lastTooltipLeft: 0,
-            lastTooltipTop: 0,
         });
+        this.isHovering = false;
+        this.hideTimeout = null;
+        this.requestAnim = null;
 
         onWillRender(() => {
             this.items = this.computeItems();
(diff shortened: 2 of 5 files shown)


addons/api_doc/static/src/components/doc_method.js
```diff
@@ -27,7 +27,7 @@ export class DocMethod extends Component {
                     value: "annotation" in options ? options.annotation : "-",
                 },
                 { type: TABLE_TYPES.Code, value: this.getDefaultValue(options) },
-                { type: TABLE_TYPES.Tooltip, value: markup(options.doc) || "" },
+                { type: TABLE_TYPES.Tooltip, value: !!options.doc ? markup(options.doc) : "" },
```
kebeclibre: no need for the conversion to bool

addons/api_doc/static/src/components/doc_table.xml
```diff
@@ -42,13 +42,13 @@
                         t-on-click="(ev) => this.showSubTable(ev, row.subData)"
                     />
 
-                    <t t-if="row and row.type == 'tooltip'">
+                    <t t-if="row and row.type == 'tooltip' and !!row.value">
```
kebeclibre: no need for conversion to bool

addons/api_doc/static/src/components/doc_table.js
```diff
@@ -46,52 +46,47 @@ export class DocTable extends Component {
...
-            this.state.lastTooltipTop = top;
-            this.state.lastTooltipLeft = left;
         });
     }
 
-    hideDynamicTooltip() {
+    scheduleHide() {
+        this.hideTooltipTimeout = setTimeout(() => {
+            this.state.tooltipStyle = `
+                top: ${this.tooltipRef.el ? this.tooltipRef.el.style.top : 0};
```
kebeclibre: I suppose this function covers two flows:
- closing the tooltip when getting out of a tooltipable element
- prevent the closing if the target requesting the tooltip has changed

Would not that be easier if the target requesting the tooltip was saved and compared against the current tooltip ?

PiaCOS: Good idea! I made the change, thanks

addons/api_doc/static/src/components/doc_table.xml
```diff
@@ -87,9 +87,11 @@
         <DocTable data="state.subTable"/>
     </div>
     <div
-        class="o-doc-dynamic-tooltip"
+        class="o-doc-dynamic-tooltip p-3 border overflow-auto"
         t-att-style="state.tooltipStyle"
         t-ref="tooltipRef"
+        t-on-mouseenter="(ev) => this.activeHoverTarget = ev.target"
```
kebeclibre: So hovering the tooltip itself prevents its closing.
What I would do with those handlers is to set a flag on the tooltip itself (or in some variable of the component) that scheduleHide would use and eventually reset when actually hiding

addons/api_doc/static/src/components/doc_table.js
```diff
@@ -46,52 +45,45 @@ export class DocTable extends Component {
...
+        setTimeout(() => {
+            if (this.activeHoverTarget === currentTarget) {
+                this.state.tooltipStyle = `
+                    top: ${this.tooltipRef.el ? this.tooltipRef.el.style.top : 0};
+                    left: ${this.tooltipRef.el ? this.tooltipRef.el.style.left : 0};
+                    opacity: 0;
+                    pointer-events: none;
+                `;
+                setTimeout(() => {
+                    if (this.state.tooltipStyle.includes('opacity: 0')) {
```
kebeclibre: the check on opacity here is the one I was targetting with my comment (mainly).
if opacity is 0 ; then the code just above had the time to finish without being overriden by showDynamicTootip.

suggestion
```js
setTimeout(() => {
    if (currentTarget === this.activeHoverTarget) {  this.state.tooltipContent = "" }

})
```

addons/api_doc/static/src/components/doc_table.js
```diff
@@ -46,52 +45,45 @@ export class DocTable extends Component {
     }
 
     showDynamicTooltip(event, content) {
-        if (this.hideTooltipTimeout) {
-            clearTimeout(this.hideTooltipTimeout);
-            this.hideTooltipTimeout = null;
-        }
+        this.activeHoverTarget = event.target;
```
kebeclibre, line 52: it is still worth cancelling the setTimeouts eventhough their callbacks would not do nothing.

addons/api_doc/static/src/components/doc_table.js
```diff
@@ -46,52 +45,45 @@ export class DocTable extends Component {
...
+
         this.state.tooltipContent = content;
         const triggerRect = event.target.getBoundingClientRect();
-        this.state.tooltipStyle = `
-            opacity: 0;
-            pointer-events: none;
-        `;
+
         requestAnimationFrame(() => {
```
kebeclibre: I suppose the same concurrency thing may happen here:
enter an element then another in the same frame ; both callbacks would execute.
So I would cancel that request (cancelAnimationFrame) when entering the method

As a side note: we could in master consider importing the web framework's tooltip subset, those are hard to implement

addons/api_doc/static/src/components/doc_table.xml
```diff
@@ -87,10 +87,20 @@
...
     <div
-        class="o-doc-dynamic-tooltip"
+        class="o-doc-dynamic-tooltip p-3 border overflow-auto"
         t-att-style="state.tooltipStyle"
         t-ref="tooltipRef"
+        t-on-mouseenter="() => this.keepAlive()"
+        t-on-mouseleave="(ev) => this.scheduleHide(ev)"
     >
+
+    <!-- <div -->
```
kebeclibre: comment to delete,
otherwise lgtm, thanks

@robodoo delegate+

xmo-odoo:
@robodoo override=ci/security

PR #248311, opened 2026-02-11
[REF] mail, im_livechat: move /messages into /mail/data

seb-odoo:
Allow grouping this RPC with others, which implies less processing all around: less requests, less queries, less duplicate entries in results.

task-4685447

https://github.com/odoo/enterprise/pull/107182

Example: open messaging menu: grouped channels as member and inbox messages

image

Example: chatter: often grouping thread/data and messages (depends on form view/chatter rendering speed)

image

### DIFF
--- a/addons/mail/controllers/mailbox.py
+++ b/addons/mail/controllers/mailbox.py
@@ -1,32 +1,27 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from odoo import http
+from odoo.fields import Domain
 from odoo.http import request
+from odoo.addons.mail.controllers.webclient import WebclientController
 from odoo.addons.mail.tools.discuss import Store
 
 
-class MailboxController(http.Controller):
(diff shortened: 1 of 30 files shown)


addons/mail/controllers/mailbox.py
```diff
@@ -1,32 +1,27 @@
...
+        if name == "/mail/history/messages":
+            message_fetch_domain = Domain("needaction", "=", False)
+        if name == "/mail/starred/messages":
+            message_fetch_domain = Domain("starred_partner_ids", "in", [request.env.user.partner_id.id])
+        if message_fetch_domain:
+            self._resolve_messages(
+                store,
+                domain=message_fetch_domain,
+                fetch_params=params and params.get("fetch_params"),
+            )
```
tsm-odoo, line 27: ⛏️ process loop is a good candidate for using `match/case` instead of checking every single if. not a big deal though.

tsm-odoo (commented):
@robodoo r+ delegate+

PR #248357, opened 2026-02-11
[FIX] mail,account: Fix unathorized error in audit trail

hsal-odoo:
When there is a message on an account that the user has no access to, and the user attempts to access
the audit trail, they are faced with authorization error.

This commit solves this issue by excluding messages of records that the user has no access to.


kmagusiak:
Test cherry-picked to https://github.com/odoo/odoo/pull/221906

PR #248377, opened 2026-02-12
[IMP] l10n_vn_edi_viettel: flow improvements

jand-odoo:
In this commit, we improve and streamline the SInvoice process for Viettel. Key changes include:
- Add a fetch button to automatically retrieve available symbols and templates.
- Symbols and Templates are now company dependent.
- Simplify Vietnam E-Invoice section in account.move form view.
- Add a feature to fetch tax code info in invoice.
- Add default symbol setting at journal level.
- Move symbol setting from partner to company level for better management.
- Uniform all "Sinvoice" naming to "SInvoice" for consistency.

task-5357151


addons/l10n_vn_edi_viettel/views/sinvoice_views.xml
```diff
@@ -1,17 +1,47 @@
...
+                <field name="invoice_template_code"/>
+            </search>
+        </field>
+    </record>
+
+    <record id="view_l10n_vn_edi_viettel_sinvoice_symbol_list" model="ir.ui.view" >
+        <field name="name">l10n_vn_edi_viettel.sinvoice.symbol.list</field>
+        <field name="model">l10n_vn_edi_viettel.sinvoice.symbol</field>
+        <field name="arch" type="xml">
+            <list create="0">
```
vin-odoo, line 20: Are we sure we want to block creation?
@TRTR-Trang wouldn't it be safer to still allow it?

What if SInvoice server is offline for any reason, but user want to be able to setup their new symbol and still prepare invoices to send later/...

Also our logic to fetch the symbol could fail (They update the api and we are not aware, ...).
By keeping the manual creation even if the automatic system fails they are not stuck, but here they would be.

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -80,19 +33,27 @@ class L10n_Vn_Edi_ViettelSinvoiceSymbol(models.Model):
     name = fields.Char(
         string='Symbol',
         required=True,
+        readonly=True,
```
vin-odoo, line 36: Same as previous comment, it's good to automate things but let's be careful to not be too strict and end up bothering users

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +76,59 @@ def _constrains_changes(self):
...
+                'invoiceType': "all",
+            },
+            headers={
+                'Content-Type': 'application/json',
+            },
+            cookies={'access_token': access_token},
+        )
+        return symbols_data, error_message
+
+    def action_fetch_symbols(self):
```
vin-odoo, line 109: There can be multiple companies selected in the company switcher.
With the current logic, this wouldn't work well (you would only fetch for the main selected company)

It's ok! But wouldn't it be better to support that fully?

As you would do one api call per company you could wrap most of this inside a for loop as such.
```for company in self.env.companies:```

File: addons/l10n_vn_edi_viettel/models/sinvoice.py [hunk 4]
```diff
@@ -115,8 +76,59 @@ def _constrains_changes(self):
...
+        symbols_data, error = self._l10n_vn_edi_lookup_symbols()
+
+        if error:
+            raise UserError(error)
+
+        if not symbols_data:
+            raise UserError(_('No symbols found. Please check your configuration and try again.'))
+
+        # Get existing symbols in one query, use tuple of (name, template_name) as key
+        existing_symbols = {(symbol.name, symbol.invoice_template_code): symbol for symbol in self.search([])}
```
vin-odoo: Add the company in the domain to only get relevant symbols.
Otherwise you may update the symbol of another company

```suggestion
        existing_symbols = self.search([('company_id', '=', self.env.company)]).grouped(lambda s: (s.name, s.invoice_template_code))
```
Or even better, use read_group!
Something like this (didn't test)

```suggestion
        existing_symbols = self._read_group([('company_id', '=', self.env.company)], ['name', 'invoice_template_code'], ['id:recordset'])
```

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +76,59 @@ def _constrains_changes(self):
...
+
+        for symbol_data in symbols_data.get('template', []):
+            symbol_code = symbol_data.get('invoiceSeri')
+            template_name = symbol_data.get('templateCode')
+
+            values = {'name': symbol_code, 'invoice_template_code': template_name}
+            if (symbol_code, template_name) in existing_symbols:
+                existing_symbols[symbol_code, template_name].write(values)
+            else:
+                self.create(values)
```
vin-odoo: Alway better to batch create, you should accumulate the values in a list and create all at once at the end of all loops before reloading.

addons/l10n_vn_edi_viettel/security/sinvoice_security.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+<data noupdate="1">
+    <record id="l10n_vn_edi_viettel_sinvoice_symbol_rule" model="ir.rule">
+        <field name="name">SInvoice Symbol</field>
+        <field name="model_id" ref="model_l10n_vn_edi_viettel_sinvoice_symbol"/>
+        <field name="global" eval="True"/>
```
vin-odoo:
```suggestion
```
It should be set automatically as there is no group on the rule.

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -574,8 +580,7 @@ def _l10n_vn_edi_add_general_invoice_information(self, json_values):
         self.ensure_one()
         invoice_data = {
             'transactionUuid': str(uuid.uuid4()),
-            'invoiceType': self.l10n_vn_edi_invoice_symbol.invoice_template_id.template_invoice_type,
```
vin-odoo, line 577: Do we just not send this anymore? Will this not cause any issue?

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -52,7 +52,7 @@ class AccountMove(models.Model):
         copy=False,
         compute='_compute_l10n_vn_edi_invoice_state',
         store=True,
-        readonly=False,
+        readonly=True,
```
vin-odoo, line 54: I don't mind but we did this on purpose for cases when our system may fail, we don't block users 🙈

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -769,52 +774,28 @@ def _l10n_vn_edi_lookup_invoice(self):
         )
         return invoice_data, error_message
 
-    def _l10n_vn_edi_get_access_token(self):
-        """ Return an access token to be used to contact the API. Either take a valid stored one or get a new one. """
+    def l10n_vn_edi_fetch_tax_code(self):
```
vin-odoo: Can you explain a bit more in the docstring?
I suppose it is not returned by SInvoice at invoice creation. Why not fetch it automatically after the creation is done?
Is it because there is a delay before receiving it?

What is the purpose of having this field in Odoo?

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -769,52 +774,28 @@ def _l10n_vn_edi_lookup_invoice(self):
...
 
-        return self.company_id.sudo().parent_ids.filtered(
-            lambda c: c.l10n_vn_edi_username and c.l10n_vn_edi_password
-        )[-1:]
+        if 'result' in invoice_lookup:
+            tax_code = invoice_lookup['result'][0]['codeOfTax']
+            if not tax_code:
+                raise UserError(_("The Code Of Tax is not available for this invoice. Please make sure it has been processed on SInvoice side."))
+            self.l10n_vn_edi_invoice_tax_code = tax_code
+            return self.l10n_vn_edi_invoice_tax_code
```
vin-odoo: Is there a flow where we want to receive the code in response?
As this method is public, it's best not to expose data unless required.

addons/l10n_vn_edi_viettel/views/sinvoice_views.xml
TRTR-Trang: Initially, we want to block it to avoid manual error and duplication. 
Also, some users even don't know how to add the correct template code and symbol or where to get it from SInvoice. 

I think the offline server shouldn't be the issue. Most of operating companies will have their symbols ready from the set up and testing phase. The chance they add more symbols along the way is there but not that common I suppose.

addons/l10n_vn_edi_viettel/models/sinvoice.py
TRTR-Trang: Yeah follow the fetching logic, we don't want them to edit the symbols fetching from SInvoice because it is not necessary

addons/l10n_vn_edi_viettel/models/account_move.py
TRTR-Trang: In the templatecode already include invoice type. For example 1/001. 1 means type 1

addons/l10n_vn_edi_viettel/models/account_move.py
TRTR-Trang: Yeah but allow editing can cause incorrect info in case of human errors. VN loves strict rules dont worry haha

addons/l10n_vn_edi_viettel/models/account_move.py
TRTR-Trang: Yeah the flow for Invoice which uses symbol starts with letter C is: Odoo --> SInvoice --> Tax Authority--> Approval--> Grant taxcode. 
By the time we fetch the SInvoic Odoo back to Odoo, maybe taxcode is not granted yet. 
To be honest, they can go to Sinvoice and get the final Invoice with taxcode. But since we have field in Odoo to store some info, I think it is better to get it to Odoo. 

Maybe, we should also redownload the Final invoice with taxcode also if taxcode is available.

addons/l10n_vn_edi_viettel/models/account_move.py
vin-odoo: If it's enough for it to work as before, then that's a nice change :+1:

addons/l10n_vn_edi_viettel/models/account_move.py
vin-odoo: Well I'll trust you on that, we can revert the change if we get negative feedbacks.
But in theory, this would lead to less issues indeed

addons/l10n_vn_edi_viettel/views/sinvoice_views.xml
vin-odoo: I get it, I'm usually pushing for more freedom for users, as we've seen many time that trying to be too strict 'because it's better' ends up annoying users and break their flow/...

I'll trust your decision :)

addons/l10n_vn_edi_viettel/models/account_move.py
vin-odoo: I'm more wondering about what is the benefit of having the code inside Odoo.
We don't use it for any flows I believe? And we don't display it in the pdf. So is it just one extra info there in case of need?

Is it common to need to know this info? Isn't it enough to find it on SInvoice?

TRTR-Trang: hmm..
It did show up in the Sinvoice pdf file after it is granted (at the bottom) and it is a must to have this taxcode for Invoice using symbol starts with C for it to be legally used. 
But yes, we may not need this field in Odoo though. Because they will not use it to look for Invoices in Sinvoice. Only need to use Invoice number. 

But do you think having an option of download the complete Invoice after tax code is granted to Odoo is nice? So we fetch both taxcode field and Sinvoices files again so that users don't need to go to SInvoice portal again to get it.

addons/l10n_vn_edi_viettel/views/res_config_settings_views.xml
vin-odoo, line 1: Image

If we want to disable manual creation, you should set options="{'no_create':True}" on the field to hide the create option.
Also, it would be good to either:
- Have a link to the list view of your symbols here so that we can easily open it or;
- Maybe show a button/... to fetch the symbols right from the setting screen so that it's clear that it should be done.

cc @TRTR-Trang as we discussed

addons/l10n_vn_edi_viettel/models/sinvoice.py
vin-odoo, line 109: Image

Just in case, could you add a check that the VAT is set and if not raise a proper user error? (Or better, a redirectwarning)

vin-odoo, line 109: Shouldn't we compare the existing symbols and the fetched one, and archive or unlink the symbols that we no longer get from sinvoice?

addons/l10n_vn_edi_viettel/models/account_move.py
vin-odoo: You need to update the depends; it should not longer depend on the partner but it should depend on the journal

addons/l10n_vn_edi_viettel/views/account_move_views.xml
```diff
@@ -5,37 +5,35 @@
...
-            </xpath>
-            <xpath expr="//page[@id='other_tab_entry']" position="after">
-                <page id="l10n_vn_edi"
-                      string="SInvoice"
-                      invisible="move_type not in ['out_invoice', 'out_refund'] or country_code != 'VN'">
+            </header>
+            <xpath expr="//page[@id='l10n_vn']/group" position="replace"/>
+            <page id="l10n_vn" position="inside">
+                <group>
                     <group>
```
vin-odoo, line 18: Don't we want a name/title on the groups ? (Just checking as the spec says 'like the example below' which has them)

addons/l10n_vn_edi_viettel/models/account_move.py
vin-odoo: Image

Image

I got this on my invoice; not sure what it means. (Maybe it's expected)

vin-odoo (commented):
Few additional comments based on a quick functional test.
Thanks!

addons/l10n_vn_edi_viettel/views/res_config_settings_views.xml
TRTR-Trang: Thanks Nicolas. Added this to specs

addons/l10n_vn_edi_viettel/views/account_move_views.xml
jand-odoo: After discussion with TRTR, we decided to keep it without the title. This is because the e-invoice section and the s-invoice section will never displayed at the same time. E-invoice will only be displayed if the l10n_vn_edi_viettel is not installed. Meanwhile, when its installed the e-invoice fields will be hidden and Sinvoice fields will be displayed.

addons/l10n_vn_edi_viettel/models/res_config_settings.py
```diff
@@ -15,25 +15,19 @@ class ResConfigSettings(models.TransientModel):
...
-                company_id=setting.company_id.id
-            )
+    def action_fetch_symbols(self):
+        self.env['l10n_vn_edi_viettel.sinvoice.symbol'].action_fetch_symbols()
+        return {
+            'name': _('Symbols'),
+            'type': 'ir.actions.act_window',
+            'res_model': 'l10n_vn_edi_viettel.sinvoice.symbol',
+            'view_mode': 'list',
+        }
```
vin-odoo, line 31: I wouldn't return the action; the user would need to stay on the setting screen to set the default symbol and they also have the other button to open if needed.

addons/l10n_vn_edi_viettel/views/account_move_views.xml
```diff
@@ -79,4 +76,16 @@
                 action = records.action_l10n_vn_edi_update_payment_status()
         </field>
     </record>
+
+    <record id="l10n_vn_edi_fetch_tax_code" model="ir.actions.server">
+        <field name="name">Fetch SInvoice with Taxcode</field>
```
vin-odoo: Hm for the PoS task we are adding a similar action to allow fetching the files later on manually no?
Shouldn't we make this one a bit more generic to support both need?

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +77,98 @@ def _constrains_changes(self):
...
+                continue
+
+            existing_symbols = {
+                (name, invoice_template_code): symbol
+                for name, invoice_template_code, symbol in self.with_context(active_test=False)._read_group(
+                    domain=[('company_id', '=', company.id)],
+                    groupby=['name', 'invoice_template_code'],
+                    aggregates=['id:recordset'],
+                )
+            }
```
vin-odoo: Always do search/reads outside of the loops if possible to avoid going to look in the db multiple times.
You can remove the domain and add some logic inside the loop to filter out the symbols of other companies

addons/l10n_vn_edi_viettel/tests/test_edi.py
vin-odoo, line 1: You should add a few tests for the symbol fetching system 😉
- Fetching symbols
- multicompany
- fetching multiple time => no duplicate
- archving of existing symbols when fetching if they're no longer returned by sinvoice
...

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +77,98 @@ def _constrains_changes(self):
...
+                    symbols_to_create.append({
+                        'name': symbol_code,
+                        'invoice_template_code': template_name,
+                        'company_id': company.id,
+                    })
+
+                seen_keys.add(key)
+
+            if symbols_to_create:
+                self.create(symbols_to_create)
```
vin-odoo: you can create the symbols of all companies at once outside the loop 👍

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +77,98 @@ def _constrains_changes(self):
...
+
+            symbols_to_archive = [symbol for key, symbol in existing_symbols.items() if key not in seen_keys]
+            if symbols_to_archive:
+                self.browse([s.id for s in symbols_to_archive]).write({'active': False})
```
vin-odoo: Same here, you can do it for all companies at once outside of the loop

addons/l10n_vn_edi_viettel/models/account_move_send.py
```diff
@@ -135,46 +135,10 @@ def _call_web_service_after_invoice_pdf_render(self, invoices_data):
...
-                    'error_title': _('Error when receiving SInvoice files.'),
-                    'errors': [error_message for error_message in [xml_error_message, pdf_error_message] if error_message],
-                }
-
-            # Not using _link_invoice_documents for these because it depends on _need_invoice_document and I can't get it to work
-            # well while allowing users to download the files before sending.
-            attachments_data = []
-            for file, error in [(xml_data, xml_error_message), (pdf_data, pdf_error_message)]:
+            if not self.env.context.get('skip_fetch_sinvoice_files'):
+                error = invoice.l10n_vn_edi_fetch_invoice_files()
```
vin-odoo: I think you can remove the context key check. We don't need it until the PoS module is here, so we can add it at that time.

vin-odoo (changes requested):
Smaller review 👍 Mostly good to go, just a few smaller change and some tests to add.

Thanks!

addons/l10n_vn_edi_viettel/models/res_config_settings.py
jand-odoo: I see, I think if that the case it will be better to reload the view so the many2one selection list will be updated

addons/l10n_vn_edi_viettel/models/account_move.py
```diff
@@ -307,6 +306,53 @@ def _l10n_vn_edi_fetch_invoice_pdf_file_data(self):
             'res_field': 'l10n_vn_edi_sinvoice_pdf_file',
         }, ""
 
+    def l10n_vn_edi_fetch_invoice_files(self):
```
vin-odoo: Missed this, but for now I don't think this is called by the frontend at all so it shouldn't be public (add _)

addons/l10n_vn_edi_viettel/models/sinvoice.py
```diff
@@ -115,8 +77,100 @@ def _constrains_changes(self):
...
+        existing_symbols = {
+            (name, invoice_template_code, company_id): symbol
+            for name, invoice_template_code, company_id, symbol in self.with_context(active_test=False)._read_group(
+                domain=[('company_id', 'in', vn_companies.ids)],
+                groupby=['name', 'invoice_template_code', 'company_id'],
+                aggregates=['id:recordset'],
+            )
+        }
+
+        _logger.warning("key present: %s", ('C23TSA', '1/0201147203', self.env['res.company'].browse(2)) in existing_symbols)
```
vin-odoo: 🧐

addons/l10n_vn_edi_viettel/tests/test_sinvoice.py
```diff
@@ -0,0 +1,118 @@
...
+    def test_fetch_symbols_creates_new_symbols(self):
+        """Fetching symbols from the API should create new symbol records."""
+        request_response = self._mock_api_fetch_symbols([('C23TSA', '1/001'), ('C23TSB', '1/002')])
+        with patch('odoo.addons.l10n_vn_edi_viettel.models.sinvoice.L10n_Vn_Edi_ViettelSinvoiceSymbol._l10n_vn_edi_lookup_symbols', return_value=request_response):
+            self.env['l10n_vn_edi_viettel.sinvoice.symbol'].with_company(self.company_vn_1).action_fetch_symbols()
+
+        symbols = self.env['l10n_vn_edi_viettel.sinvoice.symbol'].search([('company_id', '=', self.company_vn_1.id)])
+        self.assertEqual(len(symbols), 2)
+        self.assertIn('C23TSA', symbols.mapped('name'))
+        self.assertIn('C23TSB', symbols.mapped('name'))
```
vin-odoo: You could use assertRecordValues

addons/l10n_vn_edi_viettel/tests/test_sinvoice.py
```diff
@@ -0,0 +1,118 @@
...
+        active_symbols = self.env['l10n_vn_edi_viettel.sinvoice.symbol'].search([('company_id', '=', self.company_vn_1.id)])
+        archived_symbols = self.env['l10n_vn_edi_viettel.sinvoice.symbol'].with_context(active_test=False).search([
+            ('company_id', '=', self.company_vn_1.id),
+            ('active', '=', False),
+        ])
+
+        self.assertEqual(len(active_symbols), 1)
+        self.assertEqual(active_symbols.name, 'C23TSA')
+        self.assertEqual(len(archived_symbols), 1)
+        self.assertEqual(archived_symbols.name, 'C23TSB')
```
vin-odoo: I suggest you do only one search with the active_test=False

Then use assertRecordValues to assert the name and 'active' of both symbols

vin-odoo (changes requested):
Nearly there 😄
I'll recheck more in details the logic to update symbols after the logger has been corrected.
Thanks!

addons/l10n_vn_edi_viettel/models/sinvoice.py
jand-odoo: Forgot to remove it

addons/l10n_vn_edi_viettel/tests/test_sinvoice.py
```diff
@@ -0,0 +1,115 @@
+from unittest.mock import patch
+from odoo.tests import TransactionCase
+
+
+class TestSInvoiceSymbol(TransactionCase):
```
vin-odoo, line 6: You need to add the test tags :)
```suggestion
@tagged('post_install_l10n', 'post_install', '-at_install')
class TestSInvoiceSymbol(TransactionCase):
```

vin-odoo (approved):
Besides the test tags missing, LGTM

Nice work! Thank you!

addons/l10n_vn_edi_viettel/tests/test_sinvoice.py
jand-odoo: Oh right, missed that 😅

PR #248380, opened 2026-02-12
 [IMP] pos_imin: adapt pos.printer updates 

tong-odoo:
Update the Imin POS driver to ensure compatibility with recent changes
in the `pos.printer` base class. Adjusted method calls and signatures
to match the new interface.


davidmonnom (approved):

tong-odoo:
@davidmonnom it is ready

davidmonnom (approved):
@robodoo r+

PR #205154, opened 2025-04-08
[FIX] mrp: calculate workorder workcenter costs as intervals

ethanrobv:
**Current behavior**
Having multiple employees on a workorder with overlapping work
intervals will result in an inaccurate valuation value.

**Expected behavior:**
Accurate valuation.

**Steps to reproduce:**
1. Create a BOM with an operation (workorder) for a finished
product which is valuated

2. Create an MO for that BOM, assign multiple employees on the
workorder with various (overlapping) working intervals

3. Process the MO -> look at the valuation generated -> its
value is too high (because it counted the overlapping work
intervals)

**Cause of the issue**
A workorder's resource intervals are currently just naively
summed.

**Fix**
Use the `Intervals` class to find the union of all resources in
a workorder.

opw-4430375

PR #248381, opened 2026-02-12
[FIX] mrp: calculate real duration as intervals

daiduongnguyen-odoo:
[[FIX] mrp: calculate real duration excluding non-productive intervals](https://github.com/odoo/odoo/pull/248381/changes/fac6b2540a32a015f56085c2c27ba4281eb659cf)
and deduplicating overlaps

* Current Situation: Currently real duration is total duration of each
time tracking which
is not consistent with the time that use to calculate the cost for
valuation , see https://github.com/odoo/odoo/pull/205154 .The real
duration of a work order was incorrectly summing all time tracking
entries regardless of their loss type, and using simple addition which
double-counts overlapping intervals.

* Solution:
- Filter time entries to only 'productive' and 'performance' loss types,
  excluding 'availability' and 'quality' as they represent
downtime/blocking
  time, not actual work duration.
- Pool all productive and performance entries into a single Intervals
call
  so that overlaps across both types are merged in one pass.

Note: the enterprise17 implementation groups time entries by loss_type
into
separate buckets before calling Intervals, which means overlaps between
'productive' and 'performance' entries are not merged and get
double-counted.
Pooling both types together before the Intervals call avoids this.

* This also fix:
- Fix _set_duration to ensure newly created time entries start after the
  latest existing entry's end date. Without this, calling _set_duration
  twice in quick succession (e.g. in tests) produces two entries with
  overlapping timestamps, which Intervals correctly merges into one,
  causing the computed duration to be half the expected value.

---

### DIFF
--- a/addons/mrp/models/mrp_workorder.py
+++ b/addons/mrp/models/mrp_workorder.py
@@ -308,10 +308,10 @@ class MrpWorkorder(models.Model):
                 or (workorder._origin != workorder and workorder._origin.qty_producing and workorder.qty_producing != workorder._origin.qty_producing)):
                 workorder.duration_expected = workorder._get_duration_expected()
 
-    @api.depends('time_ids.duration', 'qty_produced')
+    @api.depends('time_ids.duration', 'time_ids.loss_type', 'qty_produced')
     def _compute_duration(self):
         for order in self:
-            order.duration = sum(order.time_ids.mapped('duration'))
+            order.duration = order.get_duration()
             order.duration_unit = round(order.duration / max(order.qty_produced, 1), 2)  # rounding 2 because it is a time
             if order.duration_expected:
                 order.duration_percent = max(-2147483648, min(2147483647, 100 * (order.duration_expected - order.duration) / order.duration_expected))
@@ -326,7 +326,7 @@ class MrpWorkorder(models.Model):
             return minutes * 60 + seconds
 
         for order in self:
-            old_order_duration = sum(order.time_ids.mapped('duration'))
+            old_order_duration = order.get_duration()
             new_order_duration = order.duration
             if new_order_duration == old_order_duration:
                 continue
@@ -336,6 +336,14 @@ class MrpWorkorder(models.Model):
             if delta_duration > 0:
                 enddate = datetime.now()
                 date_start = enddate - timedelta(seconds=_float_duration_to_second(delta_duration))
+                # If existing entries would overlap with the new one, push the new entry
+                # to start exactly where the latest existing entry ends.
+                end_dates = order.time_ids.filtered('date_end').mapped('date_end')
+                if end_dates:
+                    latest_end = max(end_dates)
+                    if latest_end > date_start:
+                        date_start = latest_end
+                        enddate = latest_end + timedelta(seconds=_float_duration_to_second(delta_duration))
                 if order.duration_expected >= new_order_duration or old_order_duration >= order.duration_expected:
                     # either only productive or only performance (i.e. reduced speed) time respectively
                     self.env['mrp.workcenter.productivity'].create(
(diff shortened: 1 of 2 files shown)


daiduongnguyen-odoo:
> @daiduongnguyen-odoo Hello, we group by loss_type because we don't want to have time entries that are not related to productive work in the duration. (see
> 
> https://github.com/odoo/odoo/blob/09a56b25429f296a85761b6518647a5797ba7735/addons/mrp/models/mrp_workcenter.py#L489-L502
> 
> )
> but you are right the duration computation should be harmonized with the Intervals computation.

i think for `productive` and `performance` should be same group right, For now if we have that 2 group the time also consider interval but because separeted in 2 group it will increase the real duration much bigger than the one for valuation
see this image, with this time tracking the real duration will be `333:31` minutes , but Valuation cost use just 3.5 hours (~200 minutes)
image

daiduongnguyen-odoo:
image


image



Hello @Whenrow i did like you said, also i notice this override in Odoo Enterprise of `mrp_workorder`,  the `wo_ids_without_employees` never populate , also second image, not normal in compute field we try to assiged another field and that field not have same compute method, i dont think this is normal  at all ?

daiduongnguyen-odoo:
@Whenrow you should adapt in enterprise for me too right, something like this
```
def get_duration(self):
    self.ensure_one()
    now = fields.Datetime.now()
    # productive + performance share the same duration formula and can overlap each other
    # — merge them in one Intervals call to avoid double-counting
    direct_times = self.env['mrp.workcenter.productivity']
    indirect_times = defaultdict(lambda: self.env['mrp.workcenter.productivity'])
    for time in self.time_ids:
        if time.loss_id.loss_type in ('productive', 'performance'):
            direct_times |= time
        else:
            indirect_times[time.loss_id.loss_type] |= time
    duration = self._intervals_duration([(t.date_start, t.date_end or now, t) for t in direct_times])
    for dummy, times in indirect_times.items():
        duration += self._intervals_duration([(t.date_start, t.date_end or now, t) for t in times])
    return duration
```

Whenrow:
> Hello @Whenrow i did like you said, also i notice this override in Odoo Enterprise of `mrp_workorder`, the `wo_ids_without_employees` never populate , 
indeed the super is not needed 

> also second image, not normal in compute field we try to assiged another field and that field not have same compute method, i dont think this is normal at all ?
it's not a good practice But it does not seems to be buggy right now so I won't touch it


 I can make the enterprise override. 
 It seems there is a test failing now.
thank you

daiduongnguyen-odoo:
> > Hello @Whenrow i did like you said, also i notice this override in Odoo Enterprise of `mrp_workorder`, the `wo_ids_without_employees` never populate ,
> > indeed the super is not needed
> 
> > also second image, not normal in compute field we try to assiged another field and that field not have same compute method, i dont think this is normal at all ?
> > it's not a good practice But it does not seems to be buggy right now so I won't touch it
> 
> I can make the enterprise override. It seems there is a test failing now. thank you

Ok done, can you check

daiduongnguyen-odoo:
@Whenrow Hi , are you plan to keep doing this

Whenrow:
@daiduongnguyen-odoo Hello, it's still on my todo list. We are pretty busy with the next release comming

addons/mrp/models/mrp_workorder.py
Whenrow: This should also be `get_duration()`

Whenrow (approved):
only a small request. The rest is good. Thank you

addons/mrp/models/mrp_workorder.py
daiduongnguyen-odoo: ok  Done

Whenrow:
robodoo r+
thank you

PR #248383, opened 2026-02-12
[IMP] mrp: allow quantity-based quality checks in MO

sako-odoo:
In Enterprise, we are introducing the quantity type quality checks creation for
manufacturing. To do so, we need to make some adjustments in the community
to avoid code duplication between Community and Enterprise, and a separate
method has been introduced and overridden in Enterprise to handle the
specific behaviour cleanly.

task: 5888514

Enterprise PR: odoo/enterprise#107209

### DIFF
--- a/addons/mrp/models/mrp_production.py
+++ b/addons/mrp/models/mrp_production.py
@@ -1969,6 +1969,15 @@ class MrpProduction(models.Model):
         self.ensure_one()
         return True
 
+    def _set_lots_to_finished_moves(self):
+        self.ensure_one()
+        finish_moves = self.move_finished_ids.filtered(lambda m: m.product_id == self.product_id and m.state not in ('done', 'cancel'))
+        # the finish move can already be completed by the workorder.
+        for move in finish_moves:
+            if move.product_id.tracking in ['lot', 'serial'] and not move.lot_ids:
+                move.lot_ids = self.lot_producing_ids.ids
+            move.quantity = self.uom_id.round(self.qty_producing - self.qty_produced, rounding_method='HALF-UP')
+
     def _post_inventory(self, cancel_backorder=False):
         moves_to_do, moves_not_to_do, moves_to_cancel = set(), set(), set()
         for move in self.move_raw_ids:
@@ -1988,15 +1997,7 @@ class MrpProduction(models.Model):
             for key, values in tools_groupby(moves_to_do, key=lambda m: m.raw_material_production_id.id)
         ])
         for order in self:
-            finish_moves = order.move_finished_ids.filtered(lambda m: m.product_id == order.product_id and m.state not in ('done', 'cancel'))
-            # the finish move can already be completed by the workorder.
-            for move in finish_moves:
-                if move.product_id.tracking in ['lot', 'serial'] and not move.lot_ids:
-                    move.lot_ids = order.lot_producing_ids.ids
-                move.quantity = order.uom_id.round(order.qty_producing - order.qty_produced, rounding_method='HALF-UP')
-                extra_vals = order._prepare_finished_extra_vals()
-                if extra_vals:
-                    move.move_line_ids.write(extra_vals)
+            order._set_lots_to_finished_moves()
             # workorder duration need to be set to calculate the price of the product
             for workorder in order.workorder_ids:
                 if workorder.state not in ('done', 'cancel'):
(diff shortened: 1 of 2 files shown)


addons/mrp/models/mrp_production.py
```diff
@@ -1913,15 +1922,7 @@ def _post_inventory(self, cancel_backorder=False):
             for key, values in tools_groupby(moves_to_do, key=lambda m: m.raw_material_production_id.id)
         ])
         for order in self:
-            finish_moves = order.move_finished_ids.filtered(lambda m: m.product_id == order.product_id and m.state not in ('done', 'cancel'))
-            # the finish move can already be completed by the workorder.
-            for move in finish_moves:
-                if move.product_id.tracking in ['lot', 'serial'] and not move.lot_ids:
-                    move.lot_ids = order.lot_producing_ids.ids
-                move.quantity = order.uom_id.round(order.qty_producing - order.qty_produced, rounding_method='HALF-UP')
-                extra_vals = order._prepare_finished_extra_vals()
```
dhha-odoo, line 2096: The **_prepare_finished_extra_vals** method is not used anywhere, so it is dead code and can be removed.

dhha-odoo (changes requested):

addons/mrp_account/models/mrp_production.py
```diff
@@ -61,36 +61,36 @@ def _cal_price(self, consumed_moves):
...
+            for finished_move in finished_moves:
+                for work_order in self.workorder_ids:
+                    work_center_cost += work_order._cal_cost()
+                quantity = finished_move.uom_id._compute_quantity(
+                    finished_move.quantity, finished_move.product_id.uom_id)
+                extra_cost = self.extra_cost * quantity
+                total_cost = sum(move.value for move in consumed_moves) + work_center_cost + extra_cost
+
+                byproduct_moves = self.move_byproduct_ids.filtered(
+                    lambda m: m.state not in ('done', 'cancel')
```
tpma-odoo:
```suggestion
        for finished_move in finished_moves:
            for work_order in self.workorder_ids:
                work_center_cost += work_order._cal_cost()
            quantity = finished_move.uom_id._compute_quantity(
                finished_move.quantity, finished_move.product_id.uom_id)
            extra_cost = self.extra_cost * quantity
            total_cost = sum(move.value for move in consumed_moves) + work_center_cost + extra_cost

            byproduct_moves = self.move_byproduct_ids.filtered(
                lambda m: m.state not in ('done', 'cancel')
```
The `if finished_moves` is `unnecessary` because iterating over an empty recordset already does nothing. Directly iterate over `finished_moves` here to avoid shifting the entire existing code block and keep the diff smaller.

addons/mrp/models/mrp_production.py
```diff
@@ -1462,10 +1462,11 @@ def _set_qty_producing(self, mark_moves_picked=True):
             is_byproduct = move in self.move_byproduct_ids
             # Never update already picked moves.
             # sudo needed for portal users
-            if move.picked or move.sudo()._should_bypass_set_qty_producing():
+            if move.picked or move.sudo()._should_bypass_set_qty_producing() or move.additional:
                 continue
 
-            new_qty = move.uom_id.round((self.qty_producing - self.qty_produced) * move.unit_factor)
+            unit_factor, additiona_qty = move._get_production_move_qty_data((self.product_qty - self.qty_produced) or 1)
+            new_qty = move.uom_id.round((self.qty_producing - self.qty_produced) * unit_factor - additiona_qty)
```
tpma-odoo:
```suggestion
            unit_factor, additional_qty = move._get_production_move_qty_data((self.product_qty - self.qty_produced) or 1)
            new_qty = move.uom_id.round((self.qty_producing - self.qty_produced) * unit_factor - additional_qty)
```
😄

addons/mrp/models/stock_move.py
```diff
@@ -703,3 +703,7 @@ def _prepare_scrap_move_vals(self):
         else:
             res['raw_material_production_id'] = self.production_id.id
         return res
+
+    def _get_production_move_qty_data(self, initial_qty):
+        unit_factor = self.product_uom_qty / initial_qty
+        return unit_factor, 0
```
tpma-odoo:
```suggestion
        return self.product_uom_qty / initial_qty, 0
```

addons/mrp/models/mrp_production.py
tpma-odoo: There is `still an issue` when the `whole move is failed`. In this case, no split or additional move is created; instead, the existing move is directly moved to the failed location `without` being marked as `additional`, causing the `newly produced` quantity to be added to the already `failed move`.

- Create a Quantity Type QCP for the mrp Operation.
- Create an MO with a byproduct and set the initial quantity to produce.
- Fail the QC for the entire initial byproduct quantity.
- Increase the qty_producing and pass the QC for the newly produced byproduct quantity. Produce the MO.

Observe that the `entire byproduct quantity` is moved to the `failed location`, instead of only the initially failed quantity. The same issue is also present for the main product.

sako-odoo: Hey, I confirmed this with the PO, and he said there is no issue with this flow.

Even in the receipt flow, we can have a failed move and then increase its quantity. In that case, the increased quantity is also moved to the failed location.

So, we are going to keep the current flow and not change it. If we wanted to change this behavior, we would need to create a new move for every quantity increase, which would make the flow more complex.

Therefore, we are going with the current flow.

PR #248384, opened 2026-02-12
[IMP] pos_restaurant: make Pay button secondary in mobile restaurant POS

parp-odoo:
On the mobile product screen, the `Pay` button was highlighted in the same way as the `Cart` button, which could lead to user confusion.

This change updates the `Pay` button to a secondary style, clearly differentiating it from the `Cart` button.
Additionally, the `Pay` button now consistently appears as a proper button even when the cart is empty.

When             | Before             |  After
:-------------------------:|:-------------------------:|:-------------------------:
With products in cart and no preparation printer/display |  | 
Empty cart |  | 

Task-5896161


davidmonnom (approved):
@robodoo r+

PR #248386, opened 2026-02-12
[FIX] pos_restaurant: display global discount correctly on split screen

meji-odoo:
Steps:
---
- Configure Restaurant with a global discount.
- Open a session and add products.
- Apply a global discount.
- Open the control buttons and click Split.

Issue:
---
- The split screen displays the text `With [Object Object]% discount`.

Cause:
---
- The discount object was rendered directly instead of its value.

Fix:
---
- Display the correct discount value based on its type.

task-5905885

---


davidmonnom (approved):
@robodoo r+

PR #248389, opened 2026-02-12
[FIX] account: inherit product category accounts from parent hierarchy

bhra-odoo:
**Steps to reproduce:**
* Install the **Accounting** module.
* Create a product category hierarchy (grandparent → parent → child).
* Set an **income account** on the grandparent category.
* Set an **expense account** on the child category.
* Leave the parent category without any accounts.
* Create a product assigned to the child category.
* Create a customer invoice using this product.

**Observed behavior:**
* The invoice line uses the **default income account** from settings instead of the grandparent category’s income account.
* The category hierarchy is not checked to retrieve parent accounts.

**Expected behavior:**
* If no account is defined on a category, Odoo should traverse the **parent categories** to find one.
* Only if no account exists in the hierarchy should it fall back to journal or default accounts.

**Cause:**
* Account lookup only checked the **immediate category**.
* Parent categories were not considered.

**Fix:**
* Traverse the product category hierarchy when searching for accounts.
* Use the first account found in the parent chain.

opw-5869592

### DIFF
--- a/addons/account/tests/test_product.py
+++ b/addons/account/tests/test_product.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
+from odoo import Command
 from odoo.tests import Form, tagged
 from odoo.tests.common import new_test_user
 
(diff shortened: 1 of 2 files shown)


addons/account/tests/test_product.py
```diff
@@ -90,3 +90,52 @@ def test_product_tax_with_company_and_branch(self):
...
+            'property_account_expense_categ_id': child_expense.id,
+        })
+
+        product = self.env['product.product'].create({
+            'name': 'Test Product',
+            'categ_id': child_categ.id,
+        })
+
+        # Verify customer invoice uses grandparent's income account
+        invoice_form = Form(self.env['account.move'].with_context(default_move_type='out_invoice'))
```
alsh-odoo: We should avoid using `Form` when possible.

alsh-odoo (changes requested):

alsh-odoo (approved):

agr-odoo (approved):
Looks ok.
At first, it doesn't seems a bugfix but a change in behavior. However I see on the ticket chatter it was confirmed by PO, so ok for me.

Megaaaaaa (approved):
Hello hello. Thank you for the fix!

LGTM :+1: I'm just surprised a method like `_get_product_accounts` doesn't have an `ensure_one()` but it's been like that ten years and this is a stable version fix so let's not change anything.

robodoo r+

PR #248390, opened 2026-02-12
[FIX] mail:  show seen indicator when last message is a notification

mear-odoo:
Before this PR:
In channels with seen indicators, the indicator is hidden if the user's last message is a notification. Since indicators are not displayed on notifications and the logic does not fall back to the previous message, the user is left with no visible 'seen' status.

After this PR:
The seen indicator now skips notification-type messages and is correctly displayed on the last message sent by the user.

task-5921911

---

### DIFF
--- a/addons/mail/static/src/discuss/core/common/thread_model_patch.js
+++ b/addons/mail/static/src/discuss/core/common/thread_model_patch.js
@@ -181,10 +181,11 @@ const threadPatch = {
                 // starts from most recent persistent messages to find early
                 for (let i = this.persistentMessages.length - 1; i >= 0; i--) {
                     const message = this.persistentMessages[i];
-                    if (!message.isSelfAuthored) {
-                        continue;
-                    }
-                    if (message.id > this.lastMessageSeenByAllId) {
+                    if (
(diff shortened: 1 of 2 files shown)


jayp-odoo (approved):
LGTM :fireworks:

addons/mail/static/src/discuss/core/common/thread_model_patch.js
```diff
@@ -184,6 +184,9 @@ const threadPatch = {
                     if (!message.isSelfAuthored) {
                         continue;
                     }
+                    if (message.isNotification) {
+                        continue;
+                    }
                     if (message.id > this.lastMessageSeenByAllId) {
                         continue;
                     }
```
alexkuhn: I would take the opportunity to `||` the 3 expressions together and `continue` once.
Although the expressions will be on each line with the linter, this becomes 7 LOCs instead of 9 LOCs, we gain 2 LOCs. This ain't much, but seeing more LOCs in viewport is good!

alexkuhn (approved):
@robodoo delegate+

mear-odoo:
@robodoo r+

PR #248391, opened 2026-02-12
[FIX] html_editor: preserve code block newlines when container is hidden

adch-odoo:
### Steps to Reproduce:

- Create a toggle list and add content.
- Expand the toggle list and insert a code block (e.g.: /code).
- Add multiple lines inside the code block using Enter.
- Refresh the page with the toggle open, content is preserved.
- Close the toggle and refresh then content collapses into one line.

- When the toggle list is closed, the code block is inside `display: none` container. In this state, `innerText` depends on rendered layout and does not preserve newline characters `\n`. As a result, multiline code content is extracted as single line.

- Read html and normalize it into plain text by:
  - Converting `<br>` tags into newline characters.
  - Stripping remaining HTML tags.
  - Decoding HTML entities back to their literal characters.
  - Removing the extra newline introduced by a trailing `<br>`
  - Cleaning up invisible zero-width characters.

task-5909034

---

### DIFF
--- a/addons/html_editor/static/src/others/embedded_components/core/syntax_highlighting/syntax_highlighting_utils.js
+++ b/addons/html_editor/static/src/others/embedded_components/core/syntax_highlighting/syntax_highlighting_utils.js
@@ -46,10 +46,20 @@ export const newlinesToLineBreaks = (element, doc = element.ownerDocument || doc
  */
 export const getPreValue = (pre) => {
     // Trailing br gives \n in innerText but should not be visible.
-    const trailingBrs = pre.innerHTML.match(/(<br>)+$/)?.length || 0;
-    return pre.innerText
-        .slice(0, pre.innerText.length - (trailingBrs > 1 ? trailingBrs - 1 : trailingBrs))
-        .replace(/[\u200B\uFEFF]/g, "");
+    const html = pre.innerHTML;
+    const hasTrailingBr = /<br\s*\/?>$/i.test(html);
+    let text = html
+        .replace(/<br\s*\/?>/gi, "\n")
+        .replace(/<[^>]+>|[\u200B\uFEFF]/g, "")
+        .replace(
+            /&(amp|lt|gt|#x27|quot|#x60);/g,
+            (_, entity) =>
+                ({ amp: "&", lt: "<", gt: ">", "#x27": "'", quot: '"', "#x60": "`" }[entity])
+        );
+    if (hasTrailingBr && text.endsWith("\n")) {
+        text = text.slice(0, -1);
+    }
+    return text;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


addons/html_editor/static/src/others/embedded_components/core/syntax_highlighting/syntax_highlighting_utils.js
```diff
@@ -46,10 +46,20 @@ export const newlinesToLineBreaks = (element, doc = element.ownerDocument || doc
  */
 export const getPreValue = (pre) => {
     // Trailing br gives \n in innerText but should not be visible.
-    const trailingBrs = pre.innerHTML.match(/(<br>)+$/)?.length || 0;
-    return pre.innerText
```
deso-odoo, line 50: For context `innerText` is meant to return text with \n.
But it does not work in case of d-none. 🤯

Image

dmo-odoo: I suppose it's some sort of edge case. The purpose of `innerText` is to return the text as it is rendered on the screen. With `d-none` in play, its behavior is not intuitive.

Should it return the text with newlines because that's how it would be rendered if it was displayed? But it's _not_ displayed, so it's _not_ rendered. Should it return an empty string then, since _nothing_ is rendered? Or should it still return the inner text, as its name imply, but without taking rendering into account since the text is not rendered? It seems the last option has been chosen ultimately, but a case could be made for any of the three in my opinion.

Note that the [MDN documentation page](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) is explicit on the matter:
> If the element itself is not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) (for example, is detached from the document or is hidden from view), the returned value is the same as the [Node.textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) property.

And the [HTML specifications](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute) explicitely admits that this can lead to surprising results: 
> This step can produce surprising results, as when the [innerText](https://html.spec.whatwg.org/multipage/dom.html#dom-innertext) getter is invoked on an element not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered), its text contents are returned, but when accessed on an element that is [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered), all of its children that are not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) have their text contents ignored.

Funny catch indeed! 😄

adch-odoo: In our case, the embedded syntax highlighting computes its state during plugin setup.
Since `innerText` depends on rendering (e.g. when `d-none` is applied), relying on it can initialize the embedded component with an incorrect state that cannot be recovered later.
For this reason, we should avoid using `innerText` here and rely on a rendering-independent text source for state initialization.

PR #248393, opened 2026-02-12
[MOV] http: move serving outside request

thle-odoo:
Now that the http module has been split, it makes sense to move the
serving logic to the `router.py` module.

These functions do not need to be exposed from a `Request` instance.

Task-5926433


Julien00859 (approved):

kmagusiak (approved):

Julien00859:
> https://runbot.odoo.com/runbot/batch/2395787/build/102723945

Known false-positives that have been reported to the odoo-ls team. Overriding.

@robodoo r+

PR #248395, opened 2026-02-12
[IMP] spreadsheet_*: add printing wizard in o_spreadsheet

hokolomopo:
### [IMP] spreadsheet_*: add printing wizard in o_spreadsheet

A real printing wizard was introduced in o_spreadsheet, we can now
remove the `useSpreadsheetPrint` hook in odoo as well as the print
asset bundle.

Task: [5891329](https://www.odoo.com/web#id=5891329&cids=1&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

---


addons/spreadsheet_dashboard/static/src/print_assets/print_dashboard_action.scss
```diff
@@ -1,8 +1,20 @@
...
+            height: 100% !important;
+            .o_renderer {
+                overflow: visible !important;
+            }
+        }
+    }
+}
+
+.o_spreadsheet_dashboard_action:has(.o-spreadsheet-print) {
+    .o_spreadsheet_dashboard_search_panel, .o_control_panel {
```
rrahir: seems useless, the search panel is either already afected by a d-print-none, or it is wrapped in a parent o-control-panel class which means the rule in line 5 of this file applies.

shouldn't a d-print-none on the full action fix the situation altogether?

hokolomopo: It's indeed useless for the `o_spreadsheet_dashboard_search_panel`, it has a `d-print-none`. It's however needed for the `o_control_panel`, it doesn't have a `d-print-none` and I can't add one because it's a web component. 

I don't get your last point, a `d-print-none` on the full action would make it print nothing at all ? How would this work ?

PR #248399, opened 2026-02-12
[FIX] point_of_sale: Calculate the correct cost of products in POS when using AVCO

mobar-odoo:
Problem:
When using the AVCO method in calculating costs of products in POS
orders, the product cost is calculated incorrectly (sometimes shown as
0)

Cause:
When using AVCO with a product that's configured as a kit, the cost is
calculated incorrectly because the code treats the materials as the
actual product. So, if there is more than 1 material in the BOM, the
cost actually appears as 0 because there is a check that makes sure only
the cost of one product is being calculated. If one material is in the
BOM, the product cost is shown to be the unit cost of the material
(which is incorrect, since the product can use many units of that
material).

Solution:
Override the method that calculates the cost of the product in POS to
take into account the AVCO costing method and the possibility of the
product being a kit.

Steps to reproduce:
- Choose (or create a new product)
- Choose the product's category or (create a new category) that uses
AVCO costing method
- Go to Point of Sale
- Create a new POS order that contains the product.
- Validate the order
- Check the order from the POS backend orders
- See how the cost is 0 or incorrect from the product's actual cost

Based on commit 5a44ae1

opw-5401820

### DIFF
--- a/addons/pos_mrp/models/stock_move.py
+++ b/addons/pos_mrp/models/stock_move.py
@@ -1,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _get_price_unit(self):
+        """ Returns the unit price to value this stock move """
+        line = self.env.context.get('pos_order_line')
+        if line and any(move.product_id != line.product_id for move in self):
+            product = line.product_id.with_company(self.company_id)
+            bom = product.env['mrp.bom']._bom_find(product, company_id=self.company_id.id, bom_type='phantom')[product]
+            if bom:
+                return self._get_kit_price_unit(product, bom, line.qty)
+        return super()._get_price_unit()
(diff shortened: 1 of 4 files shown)


addons/pos_mrp/models/stock_move.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _get_price_unit(self):
+        """ Returns the unit price to value this stock move """
+        line = self.env.context.get('pos_order_line')
+        if line:
```
sbel-odoo: The other PR has this condition:
```
if order_line and all(move.sale_line_id == order_line for move in self) and any(move.product_id != order_line.product_id for move in self):
```
not much we can do about the first part `all(move.sale_line_id == order_line for move in self)`. It was meant to avoid unnecessary computation but since `_get_price_unit()` is called on a single pos order line we don't really need it is this case.

However, is there any reason to remove the last part `any(move.product_id != order_line.product_id for move in self)`. I feel like we could re-use this part, as a safeguard. 

Let me know what you think.

sbel-odoo (commented):
Hello @mobar-odoo, 

When you base your fix on another commit don't hesitate to mention it for reference: https://github.com/odoo/odoo/commit/5a44ae1ec71573c57a5d60903f9b81e842fc582b

I left a few comments.

addons/pos_mrp/models/stock_move.py
mobar-odoo: @sbel-odoo You're right. I just neglected the line when I saw it started with sale_line_id. We can use the condition as a safeguard.

PR #248400, opened 2026-02-12
[FIX] point_of_sale: add close option on customer display popup for small device

jipr-odoo:
Before this commit:
===================
On small devices, the customer display popup had no way to be closed, which made it impossible to exit the QR code screen.

After this commit:
==================
Added a header back button and a close action to allow users to dismiss the customer display popup on small devices as well.

Task-5926747


davidmonnom (approved):
@robodoo r+

PR #248401, opened 2026-02-12
[FIX] account,product,purchase,sale: remove product name in invoice line

WaleedElgamal:
**Steps to reproduce:
1. Install invoicing app
2. Add French language in the settings
3. Create a customer with language set as French
4. Create a product and define french translations of the name and the description in Sales tab
5. Create an invoice for that customer and choose the product you created
6. You will find the translated product name and description under the product name
7. Edit the description, save and preview the invoice
8. The invoice line will contain [Product Name EN] [Product Name FR] [Product Description FR]

**Description of the issue:
When creating an invoice for a customer whose language differs from the user's account language, manually editing the product description on an invoice line causes the original product name to be prepended to the description. The same issue happens in a RFQ in Purchase.

**Expected behaviour:
User can edit the product description in the invoice line and the output in the invoice should only be the translated name and description, without the original product name.

**Why this happens?
1. When the product is selected in the invoice line, the label is loaded from _compute_name method in account_move_line, which holds the translated name and description.
2. After editing the description and escaping the field (clicking outside it), the parseLabel method is called, which prepends the original name to the label, making the invoice output as [original name] [translated name] [translated desc.]

**The fix
Added translated_product_name computed field to follow a similar flow as the sale quotation which is already working as intended. Fix is done in Invoice and Purchase.

opw-5480494

### DIFF
--- a/addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.js
+++ b/addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.js
@@ -42,15 +42,24 @@ export class ProductLabelSectionAndNoteField extends ProductNameAndDescriptionFi
         return this.isSection || this.isSubSection;
     }
 
+    get translatedProductName() {
+        return this.props.record.data.translated_product_name;
+    }
+
+    get label() {
+        let label = this.props.record.data[this.descriptionColumn];
+        if (this.translatedProductName && label.startsWith(this.translatedProductName)) {
+            label = label.slice(this.translatedProductName.length + 1);
+        }
+        else if (this.productName && label.startsWith(this.productName)) {
+            label = label.slice(this.productName.length + 1);
+        }
+        return label;
+    }
+
     isNote(record = null) {
         record = record || this.props.record;
         return record.data.display_type === "line_note";
-    }
-
-    parseLabel(value) {
-        return (this.productName && value && this.productName.concat("\n", value))
-            || (this.productName && !value && this.productName)
-            || (value || "");
     }
 
     shouldShowWarning() {
@@ -60,6 +69,16 @@ export class ProductLabelSectionAndNoteField extends ProductNameAndDescriptionFi
             !this.isSectionOrSubSection &&
             !this.isNote()
         );
+    }
+
+    parseLabel(value) {
+        if(!value){
+            return this.productName;
+        }
+        if(this.translatedProductName){
+            return this.translatedProductName.concat("\n", value);
+        }
+        return this.productName.concat("\n", value);
     }
 }
 
@@ -80,6 +99,9 @@ export const productLabelSectionAndNoteField = {
         props.show_label_warning = options.show_label_warning;
         return props;
     },
+    fieldDependencies: [
+        { name: 'translated_product_name', type: 'char' },
+    ],
 };
 registry
     .category("fields")
--- a/addons/product/static/src/product_name_and_description/product_name_and_description.js
+++ b/addons/product/static/src/product_name_and_description/product_name_and_description.js
@@ -104,14 +104,6 @@ export class ProductNameAndDescriptionField extends Component {
         return this.props.record.data[this.props.name].display_name || "";
     }
 
-    get label() {
-        let label = this.props.record.data[this.descriptionColumn];
-        if (label.includes(this.productName)) {
-            label = label.replace(this.productName, "");
-        }
-        return label.trim();
-    }
-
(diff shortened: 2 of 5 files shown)


addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.js
```diff
@@ -42,17 +42,15 @@ export class ProductLabelSectionAndNoteField extends ProductNameAndDescriptionFi
         return this.isSection || this.isSubSection;
     }
 
+    get translatedProductName() {
+        return this.props.record.data.translated_product_name;
+    }
```
hupo-odoo, line 47: since this function is the same as defined in the sales module, is the one in sales still necessary since it overrides the component ? 🤔

hupo-odoo (approved):
Other than the small comment I made, it looks good to me. Once approved by rd-logistic we can move on 😃

addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.js
WaleedElgamal: You're right.
I was trying to minimize the alterations in the code. But in that sense, i believe this method, get label() and parseLabel() are redundant now since the base class handles everything.

addons/product/static/src/product_name_and_description/product_name_and_description.js
```diff
@@ -105,11 +105,14 @@ export class ProductNameAndDescriptionField extends Component {
     }
 
     get label() {
-        let label = this.props.record.data[this.descriptionColumn];
```
ticodoo, line 108: `this.descriptionColumn` is not always `data.name`, but I don't understand how/where this `label` is used. Maybe @mohammadhossam can confirm that this change won't brake something for stock pickings/barcode.

ticodoo (commented):
Fix feels hacky to me, but afaik the `product_name_and_description` is full of string parsing hacks that have caused lots of bugs already so 🤷

If Mohammad says the change looks ok to him then I guess it's fine based on my limited testing in Purchase and Inventory and followup bugfixes can fix any bugs that come out of these changes.

addons/product/static/src/product_name_and_description/product_name_and_description.js
mohammadhossam: I didn't go through the whole diff here, but the label is used in this [hook above](https://github.com/odoo/odoo/pull/248401/changes/40c73704ba58b08b666ad7bb4eefd479e3b74980#diff-02c6f274ca9cfa19362710363af5f19e34e263ca3ec59011a3b15ebcb078c206L62). It's indeed not always `data.name` but it's handled in the override here:
https://github.com/odoo/odoo/blob/aeaace7c70b7ac3db68f188c9c517f1ff849e55d/addons/stock/static/src/views/picking_form/stock_move_product_label.js#L9-L17
I also tested the widget in pickings on the runbot of this PR and it's not broken indeed.

mohammadhossam: That being said, I don't see why the change in this line is needed, because the value of `descriptionColumn` is set correctly in the override here: https://github.com/odoo/odoo/pull/248401/changes/40c73704ba58b08b666ad7bb4eefd479e3b74980#diff-b8ade1bf6ad092228d83b87e2a4d11f4df086bfd184356e2d134530bf6d7ac4cR13.

ticodoo (approved):
Alright, since the changes don't seem to affect logistics flows and this widget is sloppily overridden anyways, it's fine as is for logistics

mohammadhossam (approved):
Nothing concerning for logistics side.
I have just a small note which is that `translatedProductName` is referenced in `product` module, while it's defined in `account` module. Even if there is currently no flow that leads to showing the widget with `product` installed and `account` not installed, you still should better reference the field only in `account` to avoid any future crashes.

addons/product/static/src/product_name_and_description/product_name_and_description.js
Feyensv, line 1: are you sure you can remove the methods here?
they have to be overridden in `account` for sure, but aren't they needed for the base widget logic ?

WaleedElgamal: I think it's fine because that file is not registered as a widget. `product_label_section_and_note_field` is the one registered as a widget and used in the xml files.

hupo-odoo (approved):
@robodoo r+

PR #248402, opened 2026-02-12
[FIX] l10n_my_edi_pos: include all orders for invoice consolidation

aksi-odoo:
Step to Reproduce:
* Install l10n_my_edi_pos and switch to that company
* Configure two POS sessions (POS 1 and POS 2)
* Create orders as follows:

POS 1:
* 2 uninvoiced orders
* 1 invoiced order
* Close POS 1

POS 2:
* 1 uninvoiced order
* 1 invoiced order
* Close POS 2

Reopen POS 1:
* 1 uninvoiced order
* Close POS 1

Reopen POS 2:
* 1 uninvoiced order

Close POS 2
* Go to Orders → Consolidated Invoice
* Create a consolidated invoice using the correct date range

Observation:
* Two documents are created
* No. of orders linked to each document are incorrect

Cause:
* `_split_pos_orders_in_lines` overwrites entries in the `lines_per_config` dictionary
* Previous orders for the same config are lost

Fix:
* Extend the existing recordset in `lines_per_config`  instead of overwriting it

Before
order consolidation bug

After
consolidated order fixed

opw-5904420

---

### DIFF
--- a/addons/l10n_my_edi_pos/models/myinvois_document_pos.py
+++ b/addons/l10n_my_edi_pos/models/myinvois_document_pos.py
@@ -240,7 +240,7 @@ class MyInvoisDocumentPoS(models.Model):
             if continuous_orders:
                 config_lines.append(self.env["pos.order"].browse(continuous_orders))
                 continuous_orders = []
-            lines_per_config[config] = config_lines
+            lines_per_config.setdefault(config, []).extend(config_lines)
 
         return lines_per_config
 
--- a/addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
+++ b/addons/l10n_my_edi_pos/tests/test_myinvois_pos.py
@@ -170,21 +170,33 @@ class TestMyInvoisPoS(TestPoSCommon):
     @mute_logger('odoo.addons.point_of_sale.models.pos_order')
     def test_consolidate_invoices_from_multiple_configs(self):
         """ When consolidating from multiple configs at once, we expect one Consolidated Invoice per config. """
+        orders = self.env['pos.order']
         with freeze_time("2025-01-01"):
             with self.with_pos_session():
-                first_order = self._create_order({'pos_order_lines_ui_args': [(self.product_one, 1.0)]})
+                orders |= self._create_order({'pos_order_lines_ui_args': [(self.product_one, 1.0)]})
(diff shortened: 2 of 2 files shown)


alsh-odoo (approved):

guva-odoo (approved):
Thanks! LGTM -> ready for next review

loukwets (approved):

addons/l10n_my_edi_pos/models/myinvois_document_pos.py
```diff
@@ -240,7 +240,7 @@ def _split_pos_orders_in_lines(self, pos_order_ids):
             if continuous_orders:
                 config_lines.append(self.env["pos.order"].browse(continuous_orders))
                 continuous_orders = []
-            lines_per_config[config] = config_lines
+            lines_per_config.setdefault(config, []).extend(config_lines)
```
adgu-odoo, line 243: It works but the issue comes first because the groupby is not used well IMO. It should be used on a recordset that is sorted by the same key that we want to group by at first. Here, it is sorted by id I guess but we could just change the line 230 to make it like `sorted_orders_to_consolidated.session_id.order_ids.sorted("config_id", reverse=True)` and we would have order that are really grouped by `config_id` like we want in the first place. I feel like it would be more understandable this way. What do you think ?

aksi-odoo: hello  @adgu-odoo 

Thanks for review, but what you suggested cannot be done, as we not only want to orders by config, but we also want continuity between the orders, so if we just sort/groupby the config_id, we will have wrong thing in our hand. 
here's a example of what i am saying, lets say we have following order in following config: 

- config 1: order 1,2,4  
- config 2: order 3,4

what we want is : 

`{pos.config(1,): [pos.order(1,2), pos.order(4,)], pos.config(2,): [pos.order(3), pos.order(4,)]}`

but your suggestion will lead us to:
`{pos.config(1,): [pos.order(1,2,4),], pos.config(2,): [pos.order(3,4),)]}`

i had this thought too, and discussed the whats and hows with the author of this feature. conclusion was We should get only records of continuous order in one line. 

i tried my best to explain this XD , but feel free to add your thought...

adgu-odoo (approved):
robodoo r+

PR #248405, opened 2026-02-12
[REM] l10n_it_edi: remove unused request parameter

LouisGobert:
Just remove the `recipient_codice_fiscale` parameter for the request to the SdI, as it is not used in the API.


lordkrandel:
@robodoo r+

PR #248406, opened 2026-02-12
[FIX] mail: prevent error on closing mail composer

pish-odoo:
Steps to reproduce:
1. Install `crm`
2. Create leads more than 500.
3. Select all and try to send email
4. Not close the wizard by "X"

Issue:
- Traceback occurs: `Uncaught Promise > Unexpected end of JSON input`

Cause:
- res_ids is not set on the composer when more than 500 records are selected. This is expected, as the compute method `_compute_res_ids()` does not write `res_ids` when the number of `active_ids` exceeds 500 (to avoid storing large payloads on the field). Because of this, the code trying to JSON.parse(res_ids) fails while dismissing the wizard at `onCloseWizardModal`

Solution:
- Fallback to context.active_ids when res_ids is not available

opw-5891862

### DIFF
--- a/addons/mail/static/tests/composer/composer.test.js
+++ b/addons/mail/static/tests/composer/composer.test.js
@@ -25,6 +25,7 @@ import { Deferred, animationFrame, tick } from "@odoo/hoot-mock";
 import {
     Command,
     getService,
+    makeDialogMockEnv,
     onRpc,
     patchWithCleanup,
     serverState,
@@ -797,6 +798,35 @@ test("remove an uploading attachment", async () => {
     await contains(".o-mail-AttachmentCard.o-isUploading:contains(text.txt)");
     await click(".o-mail-AttachmentCard-unlink");
     await contains(".o-mail-Composer .o-mail-AttachmentCard", { count: 0 });
+});
+
+test("Can dismiss mail composer with 500+ active_ids", async () => {
+    // When there are more than 500 active_ids, _compute_res_ids
+    // short-circuits and leaves res_ids empty for performance reasons.
+    // In that case, the code must rely on active_ids and dismissing
+    // the dialog must not crash.
+    const pyEnv = await startServer();
+    const env = await makeDialogMockEnv();
+    const partnerId = pyEnv["res.partner"].create({ name: "Partner" });
+    registerArchs({
+        "mail.compose.message,false,form": `
+            <form string="Compose Email" js_class="mail_composer_form">
+                <field name="model"/>
+                <field name="partner_ids"/>
+            </form>`,
+    });
+    await start();
+    const composerId = pyEnv["mail.compose.message"].create({
+        model: "res.partner",
+        res_ids: "", // simulate >500 active_ids case
+        partner_ids: [],
+    });
+    await openFormView("mail.compose.message", composerId, {
+        context: { active_ids: [partnerId] },
+    });
+    expect(env.dialogData).not.toBeEmpty()
+    // Dialog is closed without errors
+    await env.dialogData.dismiss()
 });
 
 test("Uploading multiple files in the composer create multiple temporary attachments", async () => {
--- a/addons/mail/static/src/chatter/web/mail_composer_form.js
+++ b/addons/mail/static/src/chatter/web/mail_composer_form.js
@@ -44,14 +44,19 @@ export class MailComposerFormRenderer extends formView.Renderer {
             () => [this.props.record.isInEdition, this.root.el, this.props.record.resId]
         );
 
-        const getActiveMailThreads = () =>
-            JSON.parse(this.props.record.data.res_ids).map((resId) => {
+        const getActiveMailThreads = () => {
+            // composer does not store res_ids past a certain limit, assume active_ids is used
+            const resIds = this.props.record.data.res_ids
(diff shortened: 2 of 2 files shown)


alsh-odoo (approved):

Ephoniz (approved):
LGTM 👍🏽

seb-odoo (commented):
Hello, is it possible to have a test to cover it? Especially if it has to be forward-ported all the way through master.

addons/mail/static/tests/composer/composer.test.js
```diff
@@ -799,6 +799,40 @@ test("remove an uploading attachment", async () => {
...
+        model: "res.partner",
+        res_ids: "",
+        partner_ids: [],
+    });
+    await openFormView("mail.compose.message", composerId, {
+        context: { active_ids: [partnerId] },
+    });
+    await dialogData.dismiss();
+    // Ensure active_ids fallback created the thread on dismiss.
+    expect(!!getService("mail.store").Thread.get({ model: "res.partner", id: partnerId })).toBe(true);
```
alexkuhn: This is very weird to assert that. Having the local knowledge of a thread is not relevant unless you use it.

Here this is like asserting some payload has been received. Great... But what's the purpose of it?

From my understanding of the motivation of PR, it's to fix a traceback. So I presume the `dialogData.dismiss()` here would crash without the fix? If so, just having the action and assertion something more relevant as a consequence on `dialogData.dismiss()` (like some UI changes?) is more than enough, as any crash would make the test not pass.

This is better to have implicit coverage of crash than artificially assert something very technical like `Thread.get({ model: "res.partner", id: partnerId })` that has no functional purpose here.

addons/mail/static/tests/composer/composer.test.js
```diff
@@ -799,6 +799,40 @@ test("remove an uploading attachment", async () => {
     await contains(".o-mail-Composer .o-mail-AttachmentCard", { count: 0 });
 });
 
+test("on dismiss mail composer form falls back to active_ids when res_ids is empty", async () => {
+    let dialogData;
+    patchWithCleanup(MailComposerFormController.prototype, {
```
alexkuhn: I would add some comments to give more context on the test.
Many unit tests are functionally very relevant just by test name.
Here this is technically related to `composer.res_ids` computations relying on `context.active_ids` that has a hard-limit of 500 to ignore setting on `res_ids` and code has to rely on `active_ids` instead as a fall-back.

It would have been best to have a test that shows a feature should work when reaching this 500+ records... or simulated, to have coverage of code using `active_ids`, and have functional actions that execute codes with this `active_ids` instead of `res_ids`.

If we don't feel doing the functionally relevant test, then at the very least the test should be made very clear this is technical / simulating technical aspects, through comments as I suggest. For example:

```js
test("Can dismiss mail composer with 500+ active_ids", async () => {
   // When there's more than 500 active_ids, _compute_res_ids is short-circuit to be empty for performance reasons. Code should rely on active_ids.
   // ...
   const composerId = pyEnv["mail.compose.message"].create({
        model: "res.partner",
        res_ids: "", // simulate 500+ active_ids
        partner_ids: [],
    });
}
```

These comments are clue to understand the context of test being introduced, understand this is very technical and these are related to some code in python that this test make behavioral assumptions, like providing an empty `res_ids:` here.

It's good to know all that and made it explicit in test, because if the test fails from

addons/mail/static/tests/composer/composer.test.js
```diff
@@ -799,6 +799,45 @@ test("remove an uploading attachment", async () => {
...
+        res_ids: "", // simulate >500 active_ids case
+        partner_ids: [],
+    });
+    await openFormView("mail.compose.message", composerId, {
+        context: { active_ids: [partnerId] },
+    });
+    
+    // Dialog is closed without errors
+    await dialogData.dismiss();
+    expect(".o_error_dialog").toHaveCount(0);
```
alexkuhn: No need to assert this: when there's an error the test is necessarily failing, unless it has `expect().toThrow()` or some other means to catch error before it bubbles and let HOOT decide to fail the test.

addons/mail/static/tests/composer/composer.test.js
```diff
@@ -799,6 +799,45 @@ test("remove an uploading attachment", async () => {
...
+    let dialogData;
+    patchWithCleanup(MailComposerFormController.prototype, {
+        setup() {
+            if (!this.env.dialogData) {
+                useSubEnv({ dialogData: {} });
+            }
+            dialogData = this.env.dialogData;
+            super.setup();
+        },
+    });
```
alexkuhn: I'm not entirely sure, but I think the `dialogData` patching stuff can be replaced by `makeDialogMockEnv()` entirely, which seems to be the official way to do it.

From the look of it, `makeDialogMockEnv()` has to create the client-side `env`, so this needs to be before `await start()`. The `start()` helper is designed to either create or reuse existing `env`, so it should work fine to let `makeDialogMockEnv` create the `env` and `start()` should automatically reuse this `env` ;)

alexkuhn (commented):
Test is much better! Thank you :)

alexkuhn (approved):
LGTM otherwise
@robodoo delegate+

PR #248413, opened 2026-02-12
[FIX] hw_drivers: 1 second delay when reading scale

MaxTheWhale:
Before this commit, the `_get_raw_response` method would wait until the read timed out (1 second) before returning the response.

After this commit, for the Toledo scale we instead read until we reach a `\r` character, which is at the end of the response. This eliminates the 1 second delay.

Enterprise: https://github.com/odoo/enterprise/pull/107241

---


Ysoroko (approved):
robodoo r+

PR #248415, opened 2026-02-12
[FIX] website_sale: prevent double delivery amount in express checkout

mma1377:
Versions:
---
- 17.0+

Issue:
---
Delivery amount is doubled in stripe express payment checkout.

Steps to reproduce:
---
1- Setup a stripe payment express checkout. (In my client case it is apple pay express checkout but we also reproduced it in google pay)
2- Set fixed price on standard delivery: 10
3- Navigate to website: add a product to cart. Use express checkout.

Outcome: The delivery price is calculated twice which is wrong.

Cause:
---
This regression is due to #238574. Once `/compute_taxes` is called, there is delivery lines included in SO. As a result returning `amount_total` will cause `this.paymentContext['minorAmount']` to include delivery amount.
Then we add delivery carrier `minorAmount`:
https://github.com/odoo/odoo/blob/469220cde82ecebf4a59451cc1325add6bfc3e69/addons/payment_stripe/static/src/js/express_checkout_form.js#L188-L196 Which cause delivery amount be added twice.

Fix:
---
We could exclude delivery amount from total amount in `express_checkout_shipping_address_compute_taxes`, which is going to calculate the tax for lines excluding delivery lines.

opw-5424398

### DIFF
--- a/addons/website_sale/controllers/main.py
+++ b/addons/website_sale/controllers/main.py
@@ -1787,8 +1787,11 @@ class WebsiteSale(payment_portal.PaymentPortal):
     def express_checkout_shipping_address_compute_taxes(self):
         order_sudo = request.website.sale_get_order()
         order_sudo._recompute_taxes()
+        amount_without_delivery = order_sudo._compute_amount_total_without_delivery()
 
-        return payment_utils.to_minor_currency_units(order_sudo.amount_total, order_sudo.currency_id)
+        return payment_utils.to_minor_currency_units(
+            amount_without_delivery, order_sudo.currency_id
+        )
 
     def _get_shop_payment_errors(self, order):
         """ Check that there is no error that should block the payment.
--- a/addons/website_sale/tests/test_delivery_express_checkout_flows.py
+++ b/addons/website_sale/tests/test_delivery_express_checkout_flows.py
@@ -10,6 +10,7 @@ from odoo.http import root
 from odoo.tests import tagged
 
 from odoo.addons.base.tests.common import HttpCaseWithUserDemo
+from odoo.addons.website.tools import MockRequest
 from odoo.addons.website_sale.controllers.delivery import WebsiteSaleDelivery as WebsiteSaleDeliveryController
 from odoo.addons.website_sale.controllers.main import WebsiteSale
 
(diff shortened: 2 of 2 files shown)


addons/website_sale/controllers/main.py
```diff
@@ -1787,8 +1787,13 @@ def _get_shop_payment_values(self, order, **kwargs):
     def express_checkout_shipping_address_compute_taxes(self):
         order_sudo = request.website.sale_get_order()
         order_sudo._recompute_taxes()
+        amount_delivery = sum(
+            order_sudo.order_line.filtered(lambda line: line.is_delivery).mapped('price_total')
+        )
```
vchu-odoo: there is a method `_compute_amount_total_without_delivery` that does the same

vchu-odoo (commented):
Thank you for the fix!

Could you please add a small test for that?

addons/website_sale/tests/test_delivery_express_checkout_flows.py
```diff
@@ -360,3 +360,52 @@ def test_express_checkout_registered_user_with_shipping_option_new_address(self)
...
+
+            self.make_jsonrpc_request(
+                urls.url_join(self.base_url(), WebsiteSale._express_checkout_route),
+                params={
+                    'billing_address': dict(self.express_checkout_billing_values),
+                    'shipping_address': dict(self.express_checkout_demo_shipping_values),
+                    'shipping_option': shipping_options[0],
+                },
+            )
+
```
vchu-odoo: is it really needed?
maybe we can already create the order with all necessary fields and then call `compute_taxes`

addons/website_sale/tests/test_delivery_express_checkout_flows.py
```diff
@@ -360,3 +360,52 @@ def test_express_checkout_registered_user_with_shipping_option_new_address(self)
...
+                    + "/compute_taxes",
+                ),
+                params={},
+            )
+
+            self.assertEqual(
+                minor_amount,
+                10000,
+                "The amount returned by the compute_taxes route should be the amount without delivery",
+            )
```
vchu-odoo: it is already described in the docstring
```suggestion
            self.assertEqual(minor_amount, 10000)
```

vchu-odoo (commented):
Thank you for the changes!

I would only suggest simplifying the test.

addons/website_sale/tests/test_delivery_express_checkout_flows.py
mma1377: @vchu-odoo 
I was actually changing to this just now! :smile:

addons/website_sale/tests/test_delivery_express_checkout_flows.py
```diff
@@ -360,3 +361,28 @@ def test_express_checkout_registered_user_with_shipping_option_new_address(self)
...
+            'name': 'Normal Delivery Charges',
+            'fixed_price': 10,
+            'delivery_type': 'fixed',
+            'product_id': delivery_product.id,
+        })
+        self.sale_order._create_delivery_line(delivery, 10)
+        expected_amount = self.sale_order._compute_amount_total_without_delivery()
+
+        with MockRequest(self.env, website=self.website, sale_order_id=self.sale_order.id) as req:
+            req.httprequest.method = "POST"
```
vchu-odoo: is it needed?

addons/website_sale/tests/test_delivery_express_checkout_flows.py
```diff
@@ -360,3 +361,28 @@ def test_express_checkout_registered_user_with_shipping_option_new_address(self)
...
+        """
+        Test that the /express_checkout/compute_taxes route returns the minor amount
+        without delivery costs.
+        """
+        websiteSaleDeliveryController = WebsiteSaleDeliveryController()
+
+        delivery_product = self.env['product.product'].create({
+            'name': 'Delivery',
+            'type': 'service',
+        })
```
vchu-odoo:
```suggestion
        delivery_product = self.env['product.product'].create({ 'name': 'Delivery'})
```

vchu-odoo (approved):
Thank you for your changes!

Small nits and good to go.

@robodoo delegate+

PR #248421, opened 2026-02-12
[IMP] account: improve duplicate detection

devp-odoo:
In this commit for bills:
- Highlight the reference in red when it is identical and in yellow when references are different. Apply the same logic to the duplicate document warning in the form view.
- In the form view duplicate warning, use the Bill Reference if available, then the Move Name, then "Draft".
- Yellow warnings are restricted to Draft moves; only red warnings remain visible once Posted.

task-5916255

### DIFF
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -751,7 +751,9 @@ class AccountMove(models.Model):
         groups="account.group_account_invoice,account.group_account_readonly",
     )
     duplicated_ref_ids = fields.Many2many(comodel_name='account.move', compute='_compute_duplicated_ref_ids')
+    # used to check if any moves in duplicated_ref_ids are in the 'Draft' state.
     is_draft_duplicated_ref_ids = fields.Boolean(compute="_compute_is_draft_duplicated_ref_ids")
+    is_exact_move_duplicate = fields.Boolean(compute='_compute_is_draft_duplicated_ref_ids')
     need_cancel_request = fields.Boolean(compute='_compute_need_cancel_request')
 
     show_update_fpos = fields.Boolean(string="Has Fiscal Position Changed", store=False)  # True if the fiscal position was changed
@@ -2054,7 +2056,7 @@ class AccountMove(models.Model):
         for move in self:
             move.quick_encoding_vals = move._get_quick_edit_suggestions()
 
-    @api.depends('ref', 'move_type', 'partner_id', 'invoice_date', 'tax_totals')
+    @api.depends('ref', 'move_type', 'partner_id', 'invoice_date', 'tax_totals', 'currency_id')
     def _compute_duplicated_ref_ids(self):
         move_to_duplicate_move = self._fetch_duplicate_reference()
         for move in self:
@@ -2067,7 +2069,7 @@ class AccountMove(models.Model):
         if not moves:
             return {}
 
-        used_fields = ("company_id", "partner_id", "commercial_partner_id", "ref", "move_type", "invoice_date", "state", "amount_total")
+        used_fields = ("company_id", "partner_id", "commercial_partner_id", "ref", "move_type", "invoice_date", "state", "amount_total", "currency_id")
 
         self.env["account.move"].flush_model(used_fields)
 
@@ -2144,6 +2146,7 @@ class AccountMove(models.Model):
                    AND move.id != duplicate_move.id
                    AND duplicate_move.state IN %(matching_states)s
                    AND move.move_type = duplicate_move.move_type
+                   AND move.currency_id = duplicate_move.currency_id
                    AND (
                            move.commercial_partner_id = duplicate_move.commercial_partner_id
                            OR (move.commercial_partner_id IS NULL AND duplicate_move.state = 'draft')
@@ -2166,6 +2169,15 @@ class AccountMove(models.Model):
     def _compute_is_draft_duplicated_ref_ids(self):
         for move in self:
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
+            move.is_exact_move_duplicate = any(
+                move.ref and move.ref == dup.ref
+                and move.move_type == dup.move_type
+                and move.partner_id == dup.partner_id
+                and move.invoice_date == dup.invoice_date
+                and move.amount_total == dup.amount_total
+                and move.is_purchase_document()
+                for dup in move.duplicated_ref_ids
+            )
 
     @api.depends('company_id')
     def _compute_display_qr_code(self):
@@ -5245,13 +5257,16 @@ class AccountMove(models.Model):
         self.ensure_one()
         if self.env.context.get('name_as_amount_total'):
             currency_amount = self.currency_id.format(self.amount_total)
-            if self.state == 'posted':
+            if self.is_sale_document(include_receipts=True) and self.state == "posted":
                 ref = f" - {self.ref}" if self.ref else ""
                 return _("%(name)s%(ref)s at %(currency_amount)s", name=(self.name), ref=ref, currency_amount=currency_amount)
-            if self.name:
-                return _("%(name)s - Draft at (%(currency_amount)s)", name=(self.name), currency_amount=currency_amount)
-            else:
-                return _("Draft (%(currency_amount)s)", currency_amount=currency_amount)
+            label = (self.ref or self.name or "") if self.is_purchase_document(include_receipts=True) else (self.name or "")
+            if label:
+                if self.state == 'draft':
+                    return _("%(label)s at %(currency_amount)s (Draft)", label=label, currency_amount=currency_amount)
+                return _("%(label)s at %(currency_amount)s", label=label, currency_amount=currency_amount)
+            return _("Draft (%(currency_amount)s)", currency_amount=currency_amount)
+
         name = ''
         if self.state == 'draft':
             name += {
--- a/addons/account/static/src/views/account_move_list/account_move_list_renderer.js
+++ b/addons/account/static/src/views/account_move_list/account_move_list_renderer.js
@@ -12,7 +12,11 @@ export class AccountMoveListRenderer extends FileUploadListRenderer {
     getCellClass(column, record) {
         const classNames = super.getCellClass(column, record);
         if (column.name === 'ref' && record.data.duplicated_ref_ids && record.data.duplicated_ref_ids.count !== 0) {
-            return `${classNames} table-warning`;
+            if (record.data.is_exact_move_duplicate) {
+                return `${classNames} table-danger`;
+            } else if (record.data.state === 'draft') {
+                return `${classNames} table-warning`;
+            }
         }
         return classNames;
     }
--- a/addons/account/views/account_move_views.xml
+++ b/addons/account/views/account_move_views.xml
@@ -525,6 +525,7 @@
                     </header>
                     <field name="made_sequence_gap" column_invisible="True"/>
                     <field name="duplicated_ref_ids" column_invisible="True"/>  <!-- Needed for custome view account_tree-->
+                    <field name="is_exact_move_duplicate" column_invisible="True"/>  <!-- Needed for custome view account_tree-->
                     <field name="name" decoration-bf="1" decoration-danger="made_sequence_gap and state == 'posted'" widget="char_with_placeholder_field_to_check" placeholder="/"/>
                     <field name="checked" column_invisible="True" groups="account.group_account_user"/> <!-- To be used for Reviewed badge display, see char_with_placeholder_field_to_check -->
                     <field name="invoice_partner_display_name" column_invisible="context.get('default_move_type') not in ('in_invoice', 'in_refund', 'in_receipt')" groups="base.group_user" string="Vendor" />
@@ -799,7 +800,19 @@
                         <field name="alerts" class="o_field_html" widget="actionable_errors"/>
                     </div>
                     <!--Duplicate check-->
-                    <div class="d-flex alert alert-warning w-100 d-flex align-items-center gap-1" invisible="not duplicated_ref_ids"  role="alert">
+                    <div class="d-flex alert alert-danger w-100 d-flex align-items-center gap-1" invisible="not is_exact_move_duplicate"  role="alert">
+                        <span>This document might be a duplicate of</span>
+                        <field name="duplicated_ref_ids" widget="x2many_buttons" nb_records_shown="1" string="Duplicated Documents" context="{'name_as_amount_total': True}"/>
+                        <button name="action_delete_duplicates"
+                                type="object"
+                                class="btn btn-link text-danger ms-auto d-flex align-items-center gap-1"
+                                invisible="not is_draft_duplicated_ref_ids">
+                            <i class="fa fa-trash text-danger"/>
+                            <span class="text-danger" invisible="duplicated_ref_ids.length == 1">Delete all duplicates</span>
+                            <span class="text-danger" invisible="duplicated_ref_ids.length &gt; 1">Delete duplicate</span>
+                        </button>
+                    </div>
+                    <div class="d-flex alert alert-warning w-100 d-flex align-items-center gap-1" invisible="not duplicated_ref_ids or is_exact_move_duplicate or state != 'draft'"  role="alert">
                         <span>This document might be a duplicate of</span>
                         <field name="duplicated_ref_ids" widget="x2many_buttons" nb_records_shown="1" string="Duplicated Documents" context="{'name_as_amount_total': True}"/>
                         <button name="action_delete_duplicates"


addons/account/models/account_move.py
```diff
@@ -2167,6 +2168,17 @@ def _compute_is_draft_duplicated_ref_ids(self):
         for move in self:
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
 
+    @api.depends('duplicated_ref_ids')
+    def _compute_is_exact_move_duplicate(self):
+        fields_to_check = ["ref", "move_type", "partner_id", "invoice_date", "tax_totals"]
```
khva-odoo: Why are you checking entire `tax_totals`?? `tax_totals` is a Binary field!!

addons/account/models/account_move.py
```diff
@@ -2167,6 +2168,17 @@ def _compute_is_draft_duplicated_ref_ids(self):
...
+        for move in self:
+            move.is_exact_move_duplicate = False
+            if move.move_type in ['in_invoice', 'in_refund']:
+                move.is_exact_move_duplicate = any(
+                    all(move[field] and move[field] == dup[field] for field in fields_to_check)
+                    for dup in move.duplicated_ref_ids
+                )
```
khva-odoo: This compute method can be simplified by directly accessing the fields and then checking the value.

khva-odoo (commented):
Hello :wave:
Can you please have a look at the comments?

addons/account/models/account_move.py
```diff
@@ -2167,6 +2168,22 @@ def _compute_is_draft_duplicated_ref_ids(self):
         for move in self:
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
 
+    @api.depends('duplicated_ref_ids')
+    def _compute_is_exact_move_duplicate(self):
+        for move in self:
+            move.is_exact_move_duplicate = False
+            if move.move_type in ['in_invoice', 'in_refund']:
```
khva-odoo:
```suggestion
            if move.is_purchase_document():
```

khva-odoo (commented):
Helllo,
Just a small comment.
Also please include the modification as per discussion with CHKL.

khva-odoo (approved):
LGTM, Thanks

addons/account/models/account_move.py
```diff
@@ -2167,6 +2169,22 @@ def _compute_is_draft_duplicated_ref_ids(self):
         for move in self:
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
 
+    @api.depends('duplicated_ref_ids')
+    def _compute_is_exact_move_duplicate(self):
+        for move in self:
+            move.is_exact_move_duplicate = False
+            if move.is_purchase_document():
```
aboo-odoo: Why just for the purchase documents ?

addons/account/models/account_move.py
```diff
@@ -2167,6 +2169,22 @@ def _compute_is_draft_duplicated_ref_ids(self):
         for move in self:
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
 
+    @api.depends('duplicated_ref_ids')
+    def _compute_is_exact_move_duplicate(self):
```
aboo-odoo: Is there a way to merge the logic of both compute and have a single one ? I mean, it can either be one of 3 cases no ?

- No duplicate found - `duplicated_ref_ids` is empty
- Full duplicate - match as done below - print in red
- Partial match - the rest

This idea is further reinforce by the fact that the JS code does not actually use `is_draft_duplicated_ref_ids` and only:  is the duplicate in `duplicated_ref_ids` and is there full match ?

aboo-odoo (commented):
Hey! 

I'd like to have your opinion on the _compute merge 😄 Let me know!

addons/account/models/account_move.py
devp-odoo: Actually, CHKL currently only wants this check for purchase documents.

addons/account/models/account_move.py
devp-odoo: Yes, you are right. Since both depend on the same field, it makes sense to merge the logic. As this is a stable version, I will add my logic into `_compute_is_draft_duplicated_ref_ids`. Does that work for you?

aboo-odoo: Yes no problem :+1: but maybe you can rename the compute into something more meaningful ? It works in stable I believe

devp-odoo: I agree a better name would be more meaningful. However, since this is a stable version, I am concerned that renaming it would technically 'delete' the original method and break any custom modules that override it. We can keep it as is for safety and rename it in forward port of master.

What do you think? :smile:

aboo-odoo: You're right my bad, forgot about custom :smile:

addons/account/static/src/views/account_move_list/account_move_list_renderer.js
```diff
@@ -12,7 +12,11 @@ export class AccountMoveListRenderer extends FileUploadListRenderer {
     getCellClass(column, record) {
         const classNames = super.getCellClass(column, record);
         if (column.name === 'ref' && record.data.duplicated_ref_ids && record.data.duplicated_ref_ids.count !== 0) {
-            return `${classNames} table-warning`;
+            if (record.data.is_exact_move_duplicate) {
+                return `${classNames} table-danger`;
+            } else if (record.data.state === 'draft') {
```
aboo-odoo, line 17: Why not use `is_draft_duplicated_ref_ids` ?
```suggestion
            } else if (record.data.is_draft_duplicated_ref_ids) {
```

This would allow to only have the yellow box on the duplicated item making it consistent with CHKL's comment about only showing the warning an the Draft moves.

Image

addons/account/views/account_move_views.xml
```diff
@@ -799,7 +800,19 @@
                         <field name="alerts" class="o_field_html" widget="actionable_errors"/>
                     </div>
                     <!--Duplicate check-->
-                    <div class="d-flex alert alert-warning w-100 d-flex align-items-center gap-1" invisible="not duplicated_ref_ids"  role="alert">
+                    <div class="d-flex alert alert-danger w-100 d-flex align-items-center gap-1" invisible="not is_exact_move_duplicate"  role="alert">
```
aboo-odoo, line 803: Again, this is oddly similar to the button just below, why not play with the invisible part to declare it only once ?

addons/account/models/account_move.py
```diff
@@ -2165,7 +2167,19 @@ def _fetch_duplicate_reference(self, matching_states=('draft', 'posted')):
...
+                for dup in move.duplicated_ref_ids:
+                    if (
+                        move.ref and move.ref == dup.ref
+                        and move.move_type == dup.move_type
+                        and move.partner_id == dup.partner_id
+                        and move.invoice_date == dup.invoice_date
+                        and move.tax_totals['total_amount_currency'] == dup.tax_totals['total_amount_currency']
+                    ):
+                        move.is_exact_move_duplicate = True
+                        break
```
aboo-odoo: How about this ?
```suggestion
@api.depends('duplicated_ref_ids')
    def _compute_is_draft_duplicated_ref_ids(self):
        self.is_draft_duplicated_ref_ids = False
        self.is_exact_move_duplicate = False

        full_match_bills_duplicates_dict = {
            move: [
                (dup.ref, dup.partner_id, dup.invoice_date, dup.tax_totals['total_amount_currency'])
                for dup in move.duplicated_ref_ids
            ]
            for move in self
            if move.is_purchase_document()
        }

        for move in self:
            # full match case (only for bills)
            if (move.ref, move.partner_id, move.invoice_date, move.tax_totals['total_amount_currency']) in full_match_bills_duplicates_dict.get(move, []):
                move.is_exact_move_duplicate = True
            # match case (cf. SQL query for duplicated_ref_ids)
            elif move.state == 'draft' and move.duplicated_ref_ids:
                move.is_draft_duplicated_ref_ids = True
```

aboo-odoo (commented):
Several additional feedback needs some caring after further discussing the POs: 
1 - Don't show the banner on non-draft moves

Image

Image

2 - Don't try to delete posted moves

Image

Image

addons/account/static/src/views/account_move_list/account_move_list_renderer.js
devp-odoo: Actually, my code is doing the same thing. We already achieve the same result (regarding not showing the yellow warning in posted moves) with my code. But yes, we can use `is_draft_duplicated_ref_ids` instead of checking the `state`.

addons/account/models/account_move.py
devp-odoo: I think there are few points we should consider regarding this suggestion: :thinking: 
```
self.is_draft_duplicated_ref_ids = False
self.is_exact_move_duplicate = False
```
There may be multiple moves at once, which will cause an issue if we use `self` directly. Am i right?
```
elif move.state == 'draft' and move.duplicated_ref_ids:
     move.is_draft_duplicated_ref_ids = True
```
Additionally, we need to check if there are any draft moves within `duplicated_ref_ids`. We don't need to check whether the current move is in draft or not, because this field is used for the `Delete duplicate` button in the warning, as shown in the image.

Regarding the use of a dictionary, I think it could cause performance issues if `duplicated_ref_ids` is too large, as we would be storing those move values temporarily.
Please correct me if i am wrong :smiley: 
image

aboo-odoo: > There may be multiple moves at once, which will cause an issue if we use `self` directly. Am i right?

You can write on multiple records the same value no problem
image

> Additionally, we need to check if there are any draft moves within `duplicated_ref_ids`. We don't need to check whether the current move is in draft or not, because this field is used for the `Delete duplicate` button in the warning, as shown in the image.

Why though ?

The wanted behavior is that a posted move **cannot** have partial duplicates. If the user posted the bill knowing it was potentially a duplicate, it is not a duplicate.

The opposite is not True, a 'Draft' move may be a duplicate of posted and/or draft moves. It made no sense to me as to why the condition was done in that way filtering on draft move only initially.

**edit**: I've just understood the goal of the button, it took me a while... if its goal is just to show the button when there is at least 1 draft invoice I have a comment about is: The naming could be much more explicit, for example, `has_draft_duplicates` or `show_delete_duplicate_button`. Because when reading `is_draft_duplicate`, I expect it to be true when the bills is a draft and is a duplicate of another bill.

Forget my suggestions about it and continue using it like before :+1: However, in master, we'll rename it so that it makes and sense (I probably not going to be the last one to question that field) :smile: 

We still need to maintain that a posted bill should not have the yellow warning shown.

> Regarding the use of a dictionary, I think it could cause performance issues if `duplicated_ref_ids` is too large, as we would be storing those move values temporarily. Please correct me if i am wrong 😃 

You are right, it's probably best to build a set of tuple on a per/move basis (in the loop) :+1:

addons/account/static/src/views/account_move_list/account_move_list_renderer.js
aboo-odoo: True sorry. However, I still think it's better to use the field we compute rather than redoing conditional statement. It is better for consistency and easier to maintain.

devp-odoo: Oh, wait my mistake! I just realized we can't use `is_draft_duplicated_ref_ids` here.
As we discussed, this field checks the state of the duplicated moves, not the current move. If we used it in the JS, a Posted move with Draft duplicates would incorrectly show the yellow warning box.

(I agree, this name is really confusing! 😄)

addons/account/views/account_move_views.xml
devp-odoo: The two blocks are very similar, but they serve different visual purposes: one is a danger (red) alert for exact matches, and the other is a warning (yellow) for partial matches. Since Odoo XML doesn't allow us to dynamically toggle the alert-danger/alert-warning classes on a single div easily, I kept them separate to maintain that visual distinction. Do you have a preferred way to merge the CSS classes?

aboo-odoo: Can you not use something like this ? 
image

devp-odoo:
Hello @aboo-odoo,
Can we move forward with this, or is there anything else you would like me to change?
Thank you for the review :smile:

aboo-odoo (approved):
Let's go! Thank you

@robodoo r+

addons/account/models/account_move.py
william-andre: That's not an answer though.

addons/account/models/account_move.py
```diff
@@ -5240,11 +5255,10 @@ def _get_move_display_name(self, show_ref=False):
         self.ensure_one()
         if self.env.context.get('name_as_amount_total'):
             currency_amount = self.currency_id.format(self.amount_total)
-            if self.state == 'posted':
-                ref = f" - {self.ref}" if self.ref else ""
-                return _("%(name)s%(ref)s at %(currency_amount)s", name=(self.name), ref=ref, currency_amount=currency_amount)
-            if self.name:
-                return _("%(name)s - Draft at (%(currency_amount)s)", name=(self.name), currency_amount=currency_amount)
+            if self.ref:
+                return _("%(ref)s at %(currency_amount)s", ref=self.ref, currency_amount=currency_amount)
```
william-andre: Isn't this valid only for purchase documents?
For sales docs, we should use the `name`; and why don't we display both ?

addons/account/models/account_move.py
```diff
@@ -5240,11 +5255,10 @@ def _get_move_display_name(self, show_ref=False):
         self.ensure_one()
         if self.env.context.get('name_as_amount_total'):
             currency_amount = self.currency_id.format(self.amount_total)
-            if self.state == 'posted':
```
william-andre, line 5248: Isn't it important to know the state of the other documents if we need to decide which one to keep/delete?

If the current document is draft but the duplicate is posted, don't we want to delete the current one?
And if the current is posted but the other one is draft, don't we want to delete the other one?

addons/account/models/account_move.py
```diff
@@ -2165,7 +2168,19 @@ def _fetch_duplicate_reference(self, matching_states=('draft', 'posted')):
...
+            move.is_exact_move_duplicate = False
             move.is_draft_duplicated_ref_ids = any(duplicate_move.state == 'draft' for duplicate_move in move.duplicated_ref_ids)
+            if move.is_purchase_document():
+                for dup in move.duplicated_ref_ids:
+                    if (
+                        move.ref and move.ref == dup.ref
+                        and move.move_type == dup.move_type
+                        and move.partner_id == dup.partner_id
+                        and move.invoice_date == dup.invoice_date
+                        and move.tax_totals['total_amount_currency'] == dup.tax_totals['total_amount_currency']
```
william-andre: We shouldn't use the `tax_totals` amount, it should be used only for display.
There are other regular `Monetary` fields for this.

addons/account/models/account_move.py
```diff
@@ -2165,7 +2168,19 @@ def _fetch_duplicate_reference(self, matching_states=('draft', 'posted')):
...
+                    ):
+                        move.is_exact_move_duplicate = True
+                        break
```
william-andre: Instead of
```python
record.field = False
for related in record.related_ids:
    if related.condition:
        record.field = True
        break
```

it would be more readable and easy to follow something like this
```python
record.field = any(
    related.condition
    for related in record.related_ids
)
```

addons/account/models/account_move.py
devp-odoo: You are right about that. I think it is better to show the ref/name when the move is posted so the user can understand, and if the move is not posted, we should always show 'Draft'.

However, this goes against CHKL's spec, so we need to discuss it with him. Another solution would be to add the state in the name.

chklop: Having the reference is always useful, as it's the true name of the Bill. But we can add (Draft) at the end if the move is not posted, to differentiate them.

addons/account/models/account_move.py
```diff
@@ -5245,13 +5257,15 @@ def _get_move_display_name(self, show_ref=False):
         self.ensure_one()
         if self.env.context.get('name_as_amount_total'):
             currency_amount = self.currency_id.format(self.amount_total)
-            if self.state == 'posted':
+            state = " (Draft)" if self.state == 'draft' else ""
```
william-andre: Not translated, and not a translatable of doing things.

Since you only use this once, it is better to do something like this
```python
if self.is_sale_document(include_receipts=True) and self.state == "posted":
    ref = f" - {self.ref}" if self.ref else ""
    return _("%(name)s%(ref)s at %(currency_amount)s", name=(self.name), ref=ref, currency_amount=currency_amount)
label = (self.ref or self.name or "") if self.is_purchase_document(include_receipts=True) else (self.name or "")
if label:
    if self.state == 'draft':
        return return _("%(label)s at %(currency_amount)s%(state)s (Draft)", label=label, currency_amount=currency_amount)
    return return _("%(label)s at %(currency_amount)s%(state)s (Draft)", label=label, currency_amount=currency_amount)
return _("Draft (%(currency_amount)s)", currency_amount=currency_amount)
```

william-andre: However, I don't understand why we have a different "label" for posted or draft sales documents.
Wouldn't something like this make more sense?

```python
if self.ref and self.name:
    if self.is_purchase_document(include_receipts=True):
        label = _("%(ref)s - %(name)s", name=self.name, ref=self.ref)
    else:
        label = _("%(name)s - %(ref)s", name=self.name, ref=self.ref)
    # possibly have a different case for journal entries if it makes sense?
else:
    label = self.name or self.ref or ""

if state == 'draft':
    return _("%(label)s at %(currency_amount)s (Draft)", label=label, currency_amount=currency_amount)
return _("%(label)s at %(currency_amount)s", label=label, currency_amount=currency_amount)
```

devp-odoo: Thanks for the feedback! :smile: You’re right, but according to the CHKL's specs for purchase docs, we need to prioritize the Bill Reference if available, then the move name, and finally 'Draft' as a fallback.
I think it's ok since Bill reference is more useful for identifying duplicates than our internal name.
Does this look okay to you?

devp-odoo:
Hello @william-andre,
I’ve fixed the forward-port chain. However, I had to create enterprise PRs for saas-19.2 and master due to a failing test case.
You can find the PRs here: [saas-19.2](https://github.com/odoo/enterprise/pull/113619) and [master](https://github.com/odoo/enterprise/pull/113620). Could you please take a look at these as well?

Thank you for the review :smile:

william-andre:
@robodoo r+

PR #248423, opened 2026-02-12
[FIX] l10n_pl_edi: enable KSeF auth with encrypted private keys

AhmedElemary57:
In Odoo 19.0 (Master), the handling of KSeF certificates introduced an issue where uploading an encrypted private key caused a server crash, blocking the configuration of KSeF. This functionality worked correctly in v18 (where keys were often unencrypted).

The `XadesSigner` class was not designed to accept a password argument. The `serialization.load_pem_private_key` method would fail with a `TypeError` because the password was not passed to the underlying

Updated the `XadesSigner.__init__` method to accept a `private_key_password` argument. Passed this password correctly to `serialization.load_pem_private_key`.

Current behavior before PR:

---

### DIFF
--- a/addons/l10n_pl_edi/models/res_config_settings.py
+++ b/addons/l10n_pl_edi/models/res_config_settings.py
@@ -85,8 +85,9 @@ class ResConfigSettings(models.TransientModel):
 
         key_bytes = base64.b64decode(self.l10n_pl_edi_certificate.private_key_id.pem_key)
         cert_bytes = base64.b64decode(self.l10n_pl_edi_certificate.pem_certificate)
+        private_key_password = (self.l10n_pl_edi_certificate.private_key_id.password or "").encode("utf-8") or None
         if key_bytes and cert_bytes:
-            signer = XadesSigner(key_bytes, cert_bytes)
+            signer = XadesSigner(key_bytes, cert_bytes, private_key_password)
         else:
             raise UserError(self.env._("KSeF certificate and private key are not set."))
 
--- a/addons/l10n_pl_edi/tools/xades_signer.py
+++ b/addons/l10n_pl_edi/tools/xades_signer.py
@@ -13,8 +13,8 @@ class XadesSigner:
     """
     Handles the creation of XAdES-BES signatures for KSeF authentication challenges.
     """
-    def __init__(self, key_pem, cert_pem):
-        self.private_key = serialization.load_pem_private_key(key_pem, password=None)
+    def __init__(self, key_pem, cert_pem, private_key_password=None):
+        self.private_key = serialization.load_pem_private_key(key_pem, password=private_key_password)
         self.cert = x509.load_pem_x509_certificate(cert_pem)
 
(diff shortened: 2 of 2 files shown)


addons/l10n_pl_edi/models/res_config_settings.py
```diff
@@ -85,8 +85,10 @@ def _l10n_pl_edi_ksef_authenticate(self):
 
         key_bytes = base64.b64decode(self.l10n_pl_edi_certificate.private_key_id.pem_key)
         cert_bytes = base64.b64decode(self.l10n_pl_edi_certificate.pem_certificate)
+        password_str = self.l10n_pl_edi_certificate.private_key_id.password
+        private_key_password = password_str.encode('utf-8') if password_str else None
```
Megaaaaaa: I'm gonna be annoying but as it is a matter of passwords, you could remove a tiny additionnal risk by removing the intermediate variable. As is, it is harmless but you risk future leaks of the variable value via logging tools for pretty much nothing :smile:

```suggestion
        private_key_password = (self.l10n_pl_edi_certificate.private_key_id.password or "").encode("utf-8") or None
```

addons/l10n_pl_edi/tools/xades_signer.py
```diff
@@ -13,8 +13,8 @@ class XadesSigner:
     """
     Handles the creation of XAdES-BES signatures for KSeF authentication challenges.
     """
-    def __init__(self, key_pem, cert_pem):
-        self.private_key = serialization.load_pem_private_key(key_pem, password=None)
+    def __init__(self, key_pem, cert_pem, private_key_password):
```
Megaaaaaa: I know it's ridiculously improbable that anyone has overriden your model already but just for safety, I would add a default value.

```suggestion
    def __init__(self, key_pem, cert_pem, private_key_password=None):
```

Megaaaaaa (commented):
Hey :wave:

So I digged into the CryptographyDeprecationWarning when setting up the certificate and the difference with 18.x is indeed due to the Python version.
\< 19.0 runs Python 3.10 with cryptography 3.4.8, which accepted this certificate silently.
\>= 19.0 runs Python 3.12, which requires a much newer cryptography version (>= 41, e.g. 42.0.8), where X.509 parsing became stricter.

The provided test certificate encodes its signature AlgorithmIdentifier with an explicit ASN.1 NULL parameter. For certain algorithms, the X.509 specification requires the parameters to be absent, not NULL. Newer versions of cryptography still accept it for compatibility but emit a deprecation warning and indicate it may be rejected in the future.

This warning is therefore caused by the certificate encoding itself. If a new test certificate with compliant encoding is isued, the warning will disappear without any change on our side.

As for the fix itself here, just two small comments, It's very minimal but it's details that can't hurt :)

Thank you again! :+1:

Megaaaaaa (approved):
LGTM, all green, let's get this merged :+1:

robodoo r+

PR #248428, opened 2026-02-12
[IMP] hr: add personal info to My Preferences

tskv:
In versions after 18.0, the Personal Info tab on the employee form has been removed for public user, and only a limited set of details is now available under the My Preferences menu. While this keeps the interface minimal, employees should still be able to review and update certain personal details if they are incorrect.

For example, information such as Marital Status or the preference to show birthdays to other employees should remain accessible to the employee.

This commit adds the sections "Family" and "Personal Information" in My Preferences.

task: 5920455

---


amah-odoo (approved):
Thanks for your work 💯

PR #232112, opened 2025-10-17
[IMP] website_sale: allow payment retry on cancel/error

livaios:
Design:
1. Improve the design of the payment processing screen.
2. Remove banner in Post Processing

task-4951911

Technical:
3. Redirect directly to the payment step for cancel and error payment status.
4. Get status message from a single entry point.
5. Websocket instead of polling for checking if the transaction is done processing
and trigger postprocessing.

task-4951436

See Also: https://github.com/odoo/enterprise/pull/103048

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248429, opened 2026-02-12
[IMP] (website_)sale(_collect), stock(_sms): smooth pick up in store flow

lkerroum:
The Pick up in store has some friction because everything was thought for delivery flows. In this commit, we do some adaptations to smoothen the flow. 

For the templates customization, several approaches were considered. In order to avoid template duplication, we edit the existing templates after the `website_sale_collect` module is installed by adding logic if none exists yet.

task-5896806

### DIFF
--- a/addons/sale/data/mail_template_data.xml
+++ b/addons/sale/data/mail_template_data.xml
@@ -110,7 +110,25 @@
                         <t t-set="tx_sudo" t-value="object.get_portal_last_transaction()"/>
                         Your order <span style="font-weight:bold;" t-out="object.name or ''">S00049</span>
                         <t t-if="object.state == 'sale' or (tx_sudo and tx_sudo.state in ('done', 'authorized'))">
-                            has been confirmed.<br/>
+                            <t
+                                t-if="hasattr(object, 'carrier_id') and object.carrier_id.delivery_type == 'in_store'"
+                            >
+                                <t
+                                    t-if="object.commitment_date and object.commitment_date.date() != object.date_order.date()"
+                                >
+                                    has been confirmed and will be ready on
+                                    <span
+                                        t-field="object.commitment_date"
+                                        t-options='{"widget": "date"}'
+                                    />.<br/>
+                                </t>
+                                <t t-else="">
+                                    has been confirmed and will be ready soon.<br/>
+                                </t>
+                            </t>
+                            <t t-else="">
+                                has been confirmed.<br/>
+                            </t>
                             Thank you for your trust.
                         </t>
                         <t t-elif="tx_sudo and tx_sudo.state == 'pending'">
@@ -311,7 +329,6 @@
                                     <td style="width: 20%; padding-top: 4px; border-top: 1px solid #343a40; opacity: 0.9;" align="left"><strong>Total</strong></td>
                                     <td style="width: 80%; padding-top: 4px; border-top: 1px solid #343a40; font-weight: bold; " align="right">
                                         <span style="display: inline-block; margin-right: 4px; padding: 4px 8px; border: 1px solid #dee2e6; border-radius: 4px;">
-                                            <span style="opacity: 0.9;">Paid with</span>
                                             <span t-attf-style="color: #{color}; font-weight: bold;">
                                                 <t t-out="tx_sudo.provider_id.sudo().name or ''"></t>
                                             </span>
@@ -324,7 +341,11 @@
                         <table t-if="object.partner_shipping_id and not object.only_services" width="295" style="vertical-align: top; font-size: 14px; margin-bottom: 12px;" align="left" role="presentation">
                             <tr>
                                 <td style="padding-bottom: 4px; font-weight: bold;">
-                                    <span style="opacity: 0.9;">Delivery with</span>
+                                    <t
+                                        t-if="not (hasattr(object, 'carrier_id') and object.carrier_id.delivery_type == 'in_store')"
+                                    >
+                                        <span style="opacity: 0.9;">Delivery with</span>
+                                    </t>
                                     <span t-attf-style="color: {{color}}; opacity: 1;"><t t-out="object.carrier_id.name or ''"></t></span>
                                 </td>
                             </tr>
--- a/addons/stock/data/mail_template_data.xml
+++ b/addons/stock/data/mail_template_data.xml
@@ -12,7 +12,12 @@
 <div style="margin: 0px; padding: 0px;">
     <p style="margin: 0px; padding: 0px; font-size: 13px;">
         Hello <t t-out="object.partner_id.name or ''">Brandon Freeman</t>,<br/><br/>
-        We are glad to inform you that your order has been shipped.
+        <t t-if="hasattr(object, 'delivery_type') and object.delivery_type == 'in_store'">
+            Your order has been collected.
+        </t>
+        <t t-else="">
+            We are glad to inform you that your order has been shipped.
+        </t>
         <t t-if="hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref">
             Your tracking reference is
             <strong>
--- a/addons/stock_sms/data/sms_data.xml
+++ b/addons/stock_sms/data/sms_data.xml
@@ -5,7 +5,7 @@
             <field name="name">Delivery: Send by SMS Text Message</field>
             <field name="model_id" ref="stock.model_stock_picking"/>
             <field name="body">
-                {{ (object.company_id.name + ': We are glad to inform you that your order n° ' + object.origin + ' has been shipped.' if object.origin else object.company_id.name + ': We are glad to inform you that your order has been shipped.') + (' Your tracking reference is ' + (object.carrier_tracking_ref) + '.' if hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref else '') }}
+                {{ (object.company_id.name + ': Your order n° ' + object.origin + ' has been collected.' if object.origin else object.company_id.name + ': Your order has been collected.') if hasattr(object, 'delivery_type') and object.delivery_type == 'in_store' else (object.company_id.name + ': We are glad to inform you that your order n° ' + object.origin + ' has been shipped.' if object.origin else object.company_id.name + ': We are glad to inform you that your order has been shipped.') + (' Your tracking reference is ' + (object.carrier_tracking_ref) + '.' if hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref else '') }}
             </field>
         </record>
     </data>
--- a/addons/website_sale/templates/checkout/confirmation_templates.xml
+++ b/addons/website_sale/templates/checkout/confirmation_templates.xml
@@ -112,14 +112,19 @@
             t-att-data-order-tracking-info="json.dumps(order_tracking_info)"
         >
             <t t-set="tx_sudo" t-value="order.get_portal_last_transaction()"/>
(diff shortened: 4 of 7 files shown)


addons/stock/data/mail_template_data.xml
```diff
@@ -12,7 +12,12 @@
 <div style="margin: 0px; padding: 0px;">
     <p style="margin: 0px; padding: 0px; font-size: 13px;">
         Hello <t t-out="object.partner_id.name or ''">Brandon Freeman</t>,<br/><br/>
-        We are glad to inform you that your order has been shipped.
+        <t t-if="hasattr(object, 'delivery_type') and object.delivery_type == 'in_store'">
+                Your order has been collected.
+            </t>
+            <t t-else="">
+                We are glad to inform you that your order has been shipped.
+            </t>
```
livaios: Indentation issue
```suggestion
            Your order has been collected.
        </t>
        <t t-else="">
            We are glad to inform you that your order has been shipped.
        </t>
```

addons/stock_sms/data/sms_data.xml
```diff
@@ -5,7 +5,7 @@
...
             <field name="model_id" ref="stock.model_stock_picking"/>
             <field name="body">
-                {{ (object.company_id.name + ': We are glad to inform you that your order n° ' + object.origin + ' has been shipped.' if object.origin else object.company_id.name + ': We are glad to inform you that your order has been shipped.') + (' Your tracking reference is ' + (object.carrier_tracking_ref) + '.' if hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref else '') }}
+                {{ (object.company_id.name + ': Your order n° ' + object.origin + ' has been collected.' if object.origin else object.company_id.name + ': Your order has been collected.') if hasattr(object, 'delivery_type') and object.delivery_type == 'in_store' else (object.company_id.name + ': We are glad to inform you that your order n° ' + object.origin + ' has been shipped.' if object.origin else object.company_id.name + ': We are glad to inform you that your order has been shipped.') + (' Your tracking reference is ' + (object.carrier_tracking_ref) + '.' if hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref else '') }}
```
livaios, line 8: I know the existing formatting of the mail templates aren't great but we could format it so its easily readible.
```suggestion
                {{(
                    object.company_id.name +
                    ': Your order n° ' + object.origin +
                    ' has been collected.'
                    if object.origin
                    else object.company_id.name + ': Your order has been collected.'
                ) if hasattr(object, 'delivery_type') and object.delivery_type == 'in_store'
                else (
                    object.company_id.name +
                    ': We are glad to inform you that your order n° ' +
                    object.origin +
                    ' has been shipped.'
                    if object.origin
                    else
                        object.company_id.name +
                        ': We are glad to inform you that your order has been shipped.'
                    ) +
                (
                    ' Your tracking reference is ' +
                    object.carrier_tracking_ref +
                    '.' if hasattr(object, 'carrier_tracking_ref') and object.carrier_tracking_ref 
                    else ''
                )}}
```

Your current logic doesn't add `carrier_tracking_ref` if the delivery method is PUIS, if its not the intended behavior your should add parenthesis around the first if/else

addons/sale/data/mail_template_data.xml
```diff
@@ -110,7 +110,25 @@
                         <t t-set="tx_sudo" t-value="object.get_portal_last_transaction()"/>
                         Your order <span style="font-weight:bold;" t-out="object.name or ''">S00049</span>
                         <t t-if="object.state == 'sale' or (tx_sudo and tx_sudo.state in ('done', 'authorized'))">
-                            has been confirmed.<br/>
+                            <t t-if="hasattr(object, 'carrier_id') 
+                                and object.carrier_id.delivery_type == 'in_store'"
```
livaios: Sometimes formatting rules in xml makes it less clear because of indentation so if splitting over 2 lines makes it not as readable we ignore the >100 chars rule
```suggestion
                            <t
                                t-if="hasattr(object, 'carrier_id') and object.carrier_id.delivery_type == 'in_store'"
```

addons/sale/data/mail_template_data.xml
```diff
@@ -110,7 +110,25 @@
...
+                            >
+                                <t t-if="object.commitment_date
+                                    and object.commitment_date.date() != datetime.date.today()"
+                                >
```
livaios:
```suggestion
                                <t 
                                    t-if="object.commitment_date and object.commitment_date.date() != datetime.date.today()"
                                >
```

addons/website_sale/templates/checkout/confirmation_templates.xml
```diff
@@ -113,6 +113,7 @@
         >
             <t t-set="tx_sudo" t-value="order.get_portal_last_transaction()"/>
             <div
+                id="order_confirmation_banner"
```
livaios: Instead of copying the class all over again to add a specific condition we can extract the alert class from this div into a specific value and override that value in website_sale_collect like this:

```xml
            <t 
                t-set="alert_class" 
                t-value="
                    (tx_sudo.state == 'pending' and 'alert-info') or
                    (tx_sudo.state == 'done' and order.amount_total == tx_sudo.amount and 'alert-success') or
                    (tx_sudo.state == 'done' and order.amount_total != tx_sudo.amount and 'alert-warning') or
                    (tx_sudo.state == 'authorized' and 'alert-success') or
                    'alert-danger'" 
            />
            <div
                t-if="tx_sudo"
                t-attf-class="alert px-3 pb-0 pt-3 #{alert_class}"
                role="alert"
            >
```

And in`addons/website_sale_collect/views/templates.xml`
You can override the value like this:
```xml
<t
    t-if="tx_sudo.state == 'pending' and tx_sudo.provider_id.custom_mode == 'on_site'"
    t-set="alert_class"
    t-value="'alert-success'"
/>
```

addons/website_sale_collect/views/templates.xml
```diff
@@ -4,14 +4,32 @@
...
+        <div id="order_confirmation_banner" position="attributes">
+            <attribute name="t-attf-class" add="alert px-3 pt-3 #{
+                (tx_sudo.state == 'pending' and tx_sudo.provider_id.custom_mode == 'on_site' and 'alert-success') or
+                (tx_sudo.state == 'pending' and 'alert-info') or
+                (tx_sudo.state == 'done' and order.amount_total == tx_sudo.amount and 'alert-success') or
+                (tx_sudo.state == 'done' and order.amount_total != tx_sudo.amount and 'alert-warning') or
+                (tx_sudo.state == 'authorized' and 'alert-success') or
+                'alert-danger'}"/>
+        </div>
+        <xpath expr="//t[@t-out='tx_sudo.provider_id.sudo().pending_msg']" position="replace" >
```
livaios: You can inherit the template like this as well its clearer
```suggestion
        <t t-out="tx_sudo.provider_id.sudo().pending_msg" position="replace" >
```

In another task, we are unifying the entry point to the status message of a transaction by getting the message from a function, so if this [PR](https://github.com/odoo/odoo/pull/232112/changes#diff-8b91acdcd4759865cf7b042ea950e36db670fe848226dcecbd2f41e91f3fc5f8) is merged first you can simply override the function instead of the template but nothing to do for now

addons/sale/data/mail_template_data.xml
```diff
@@ -324,7 +341,11 @@
                         <table t-if="object.partner_shipping_id and not object.only_services" width="295" style="vertical-align: top; font-size: 14px; margin-bottom: 12px;" align="left" role="presentation">
                             <tr>
                                 <td style="padding-bottom: 4px; font-weight: bold;">
-                                    <span style="opacity: 0.9;">Delivery with</span>
+                                    <t t-if="not (hasattr(object, 'carrier_id')
+                                        and object.carrier_id.delivery_type == 'in_store')"
```
livaios:
```suggestion
                                    <t 
                                        t-if="not (hasattr(object, 'carrier_id') and object.carrier_id.delivery_type == 'in_store')"
```

livaios (changes requested):
Thanks for your work on this!
I just have a few comments 🙂

addons/stock_sms/data/sms_data.xml
lkerroum: Thanks for the review! I've tried formatting like this but then the sms sent is the full template and not the formatted corresponding message. 

Also, I thought that as when the delivery method is PUIS the message is sent when the order is collected by the customer, we don't need to add the tracking number. Wdyt?

livaios: Maybe the tracking number is how they identify the PUIS order, we never know how customers utilise existing functionalities so we should keep that in mind.

@Jean-boje what do you think, do we add the tracking reference to the message if the order is PUIS or is it not relevant?

Jean-boje: Hello @livaios as this email is sent after the order is picked up, I don't think it's necessary to keep it :)
Thanks !

addons/website_sale_collect/views/templates.xml
```diff
@@ -5,13 +5,29 @@
         id="payment_confirmation_status"
         inherit_id="website_sale.payment_confirmation_status"
     >
+        <xpath expr="//t[@t-set='alert_class']" position="after">
```
livaios: You can also define an override by matching the signature of the element and specifying the `position` of the override.
```suggestion
        <t t-set="alert_class" position="after">
```

livaios (changes requested):
Just one small nitpicking comment 🙂

livaios (approved):
Thanks for making the changes!
LGTM 👍

addons/sale/data/mail_template_data.xml
ticodoo, line 1: As a heads up, all the templates in this file are `noupdate="1"` which means upgraded databases won't have any of the changes you have here + existing translations will be unchanged, but newly loaded languages will match the new template. We've had issues with this in the past, but mainly only when code is changed (i.e. your changes shouldn't cause any stacktraces, just something to keep in mind that upgraded dbs won't see these changes unless you force update it in an upgrade script).

addons/stock/data/mail_template_data.xml
ticodoo, line 1: same noupdate="1" comment

ticodoo (approved):
Stock changes are fine, didn't test.

addons/sale/data/mail_template_data.xml
```diff
@@ -110,7 +110,25 @@
                         <t t-set="tx_sudo" t-value="object.get_portal_last_transaction()"/>
                         Your order <span style="font-weight:bold;" t-out="object.name or ''">S00049</span>
                         <t t-if="object.state == 'sale' or (tx_sudo and tx_sudo.state in ('done', 'authorized'))">
-                            has been confirmed.<br/>
+                            <t
+                                t-if="hasattr(object, 'carrier_id') and object.carrier_id.delivery_type == 'in_store'"
+                            >
+                                <t
+                                    t-if="object.commitment_date and object.commitment_date.date() != datetime.date.today()"
```
vchu-odoo: Can not we use date_order to compare in the same UTC?
```suggestion
                                    t-if="object.commitment_date and object.commitment_date.date() != date_order.date()"
```

vchu-odoo (commented):
Great work!

Just one small comment :)

vchu-odoo (approved):
@robodoo r+

vchu-odoo:
@robodoo r+

PR #248430, opened 2026-02-12
[FIX] pos_online_payment: only delete online payments

pedrambiria:
Before this commit, if an order had a terminal payment and an online payment was added accidentally, and then get removed, the terminal payment would get removed from server, while it won't be removed from the UI, resulting in a mismatch between the server and the UI, and the order would no be synced correctly.

opw-5911699

---


davidmonnom (approved):
@robodoo r+

PR #248431, opened 2026-02-12
[FIX] hr_work_entry_attendance : fix synchronization issue with work entries

Xiran-Kali:
Steps to reproduce:
- On an attendance with overtime, click the "Add a line" button and add a new overtime line
- Refresh the page
- The newly added line has disappeared and navigating to work entries causes a traceback

How it was fixed:
Disabled the ability to add a new overtime line.

Task ID: 5899657


mepe-odoo:
@robodoo r+

mepe-odoo:
@robodoo r+

PR #248432, opened 2026-02-12
[IMP] mail: add category group in channel search view

maad-odoo:
Purpose of this commit:
To add category based group-by in channel search view.

task-5264898


seb-odoo (approved):
@robodoo r+ delegate+

PR #248433, opened 2026-02-12
[FIX] purchase_stock: avoid extra product return during backorder

josv-odoo:
When creating a backorder because of missing product, if there is an extra product received it will create a return also create a return for that product.

### Steps to reproduce:

* Create and confirm a purchase ordre for 2 products (A & B)
* Add a product C to the picking order and reduce the quantity of product B received
* Validate the picking ordre
* Create a backorder
* Go back on the PO and open the linked pickings 
-> There is delivery linked for product C

### Observation:

When processing the backorder, it will sync the delivery and the PO with _action_synch_order where it will add the new product to the PO: https://github.com/odoo/odoo/blob/1a62c4333de61a4e1358ef8e229b2e0f5a3e9826/addons/purchase_stock/models/stock_move.py#L88-L89 
The create that is started by the creation of the new POL will trigger _create_or_update_picking that will create a stock move and a picking: https://github.com/odoo/odoo/blob/2855f0f2b0ffbc340e2af4785dde5bf465579ee8/addons/purchase_stock/models/purchase_order_line.py#L97 https://github.com/odoo/odoo/blob/2855f0f2b0ffbc340e2af4785dde5bf465579ee8/addons/purchase_stock/models/purchase_order_line.py#L197-L198

opw-5868172

### DIFF
--- a/addons/purchase_stock/tests/test_create_picking.py
+++ b/addons/purchase_stock/tests/test_create_picking.py
@@ -39,6 +39,9 @@ class TestCreatePicking(ProductVariantsCommon):
         }
 
     def test_00_create_picking(self):
+        """
+        Check that purchase orders and receipts are properly synchronized.
+        """
 
         # Draft purchase order created
         self.po = self.env['purchase.order'].create(self.po_vals)
@@ -54,9 +57,9 @@ class TestCreatePicking(ProductVariantsCommon):
         self.assertEqual(len(self.po.order_line.move_ids), 1, 'The two moves should be merged in one')
 
         # Validate first shipment
-        self.picking = self.po.picking_ids[0]
-        self.picking.move_ids.picked = True
-        self.picking._action_done()
+        receipt_1 = self.po.picking_ids[0]
+        receipt_1.move_ids.picked = True
+        receipt_1._action_done()
         self.assertEqual(self.po.order_line.mapped('qty_received'), [7.0], 'Purchase: all products should be received')
 
         # create new order line
@@ -71,6 +74,24 @@ class TestCreatePicking(ProductVariantsCommon):
         self.assertEqual(self.po.incoming_picking_count, 2, 'New picking should be created')
         moves = self.po.order_line.mapped('move_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
         self.assertEqual(len(moves), 1, 'One moves should have been created')
+        # In second receipt, add product which is not in the PO.
+        receipt_2 = self.po.picking_ids - receipt_1
+        receipt_2.move_ids.quantity = 1
+        receipt_2_form = Form(receipt_2)
+        with receipt_2_form.move_ids.new() as move:
+            move.product_id = self.product_sofa_blue
+            move.quantity = 1.0
+        receipt_2 = receipt_2_form.save()
+        Form.from_action(self.env, receipt_2.button_validate()).save().process()
+        backorder = receipt_2.backorder_ids
+        receipt_type_id = self.ref('stock.picking_type_in')
+
+        self.assertEqual(self.po.picking_ids, receipt_1 | receipt_2 | backorder)
+        self.assertRecordValues(receipt_2.move_line_ids, [
+            {'product_id': self.product_id_2.id, 'quantity': 1, 'picking_type_id': receipt_type_id},
+            {'product_id': self.product_sofa_blue.id, 'quantity': 1, 'picking_type_id': receipt_type_id}])
+        self.assertRecordValues(backorder.move_line_ids, [
+            {'product_id': self.product_id_2.id, 'quantity': 4, 'picking_type_id': receipt_type_id}])
 
     def test_01_check_double_validation(self):
 
--- a/addons/purchase_stock/models/purchase_order_line.py
+++ b/addons/purchase_stock/models/purchase_order_line.py
@@ -94,7 +94,8 @@ class PurchaseOrderLine(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         lines = super().create(vals_list)
-        lines.filtered(lambda l: l.order_id.state == 'purchase')._create_or_update_picking()
+        if not self.env.context.get('bypass_move_update'):
+            lines.filtered(lambda l: l.order_id.state == 'purchase')._create_or_update_picking()
         return lines
 
     def write(self, vals):
--- a/addons/purchase_stock/models/stock_move.py
+++ b/addons/purchase_stock/models/stock_move.py
@@ -89,7 +89,7 @@ class StockMove(models.Model):
                 po_line_vals['price_unit'] = 0
             purchase_order_lines_vals.append(po_line_vals)
         if purchase_order_lines_vals:
-            self.env['purchase.order.line'].create(purchase_order_lines_vals)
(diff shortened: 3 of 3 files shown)


addons/purchase_stock/tests/test_create_picking.py
```diff
@@ -71,6 +74,16 @@ def test_00_create_picking(self):
         self.assertEqual(self.po.incoming_picking_count, 2, 'New picking should be created')
         moves = self.po.order_line.mapped('move_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
         self.assertEqual(len(moves), 1, 'One moves should have been created')
+        self.po.picking_ids[-1].move_line_ids.quantity = 1
+        self.po.picking_ids[-1].write({'move_line_ids': [
+            Command.create({
+                'product_id': self.product_sofa_blue.id,
+                'quantity': 1.0,
+                })]})
```
lase-odoo: You are not expected to write on the `stock.move.line` directly, you can update the move quantities if you really want to force quantity. Concerning the new move addition, it is best to do it with by editing the `stock.picking` Form of the receipt.

addons/purchase_stock/tests/test_create_picking.py
```diff
@@ -71,6 +74,16 @@ def test_00_create_picking(self):
...
+        Form.from_action(self.env, self.po.picking_ids[-1].button_validate()).save().process()
+        self.assertEqual(len(self.po.picking_ids), 3)
+        self.assertEqual(len(self.po.picking_ids[-1].move_line_ids), 1)
+        self.assertRecordValues(self.po.picking_ids[-1].move_line_ids[0], [{'product_id': self.product_id_2.id, 'quantity': 4, 'picking_code': 'incoming'}])
```
lase-odoo: These asserts are quite opaque to me. At this point of the test, you want to check that a backorder has been created for the validation of the second receipt which was created because of the POL update. If you want to do so in that same test what you could do is just to name things explicitly. Eg, the line:
```py
self.picking = self.po.picking_ids[0]
```
should just be replaced by something 
```py
receipt_1 = self.po.picking_ids
```
Then you can just define:
```suggestion
        receipt_2 = self.po.picking_ids - receipt_1
        receipt_2.move_ids.quantity = 1
        ...
        Form.from_action(self.env, second_receipt.button_validate()).save().process()
        backorder = receipt_2.backorder_ids
        self.assertEqual(self.po.picking_ids, receipt_1 | receipt_2 | backorder)
        self.assertRecordValues(backorder.move_line_ids, [{'product_id': self.product_id_2.id, 'quantity': 4, 'picking_type_id': self.}])
```

addons/purchase_stock/tests/test_create_picking.py
```diff
@@ -39,6 +39,9 @@ def setUpClass(cls):
         }
 
     def test_00_create_picking(self):
+        """
+        Verify PO and Picking synchronization during modifications and backorder creation.
```
lase-odoo:
```suggestion
        Check that purchase orders and receipts are properly synchronized. 
```

lase-odoo (commented):

I do not think the fix is at the right place. Looking at the stack of the outgoing move creation we can see that the move outgoing move for product C is created during the creation of the POL for 0 units that is created because the move for C has been validated for 1 unit (and hence need to update the delivered quantity on some pol on the PO). In particular, we see that the `_create_or_update_picking` is triggered during the `_action_synch_order` of the move for that very same product in particular, if the pol was created with the correct vals there should and need to be nothing to be update on the receipt. Imo the fix is to add a context key `bypass_move_update` in the in the pol creation :
```py
        if purchase_order_lines_vals:
            self.env['purchase.order.line'].with_context(bypass_move_update=True).create(purchase_order_lines_vals)
```
here :
https://github.com/odoo/odoo/blob/a220fb71c036c93fa1e75d4d37127e5eda0118f9/addons/purchase_stock/models/stock_move.py#L88-L89
And to do this:
```py
        if not self.env.context.get('bypass_move_update'):
            lines.filtered(lambda l: l.order_id.state == 'purchase')._create_or_update_picking()
```
here:
https://github.com/odoo/odoo/blob/a220fb71c036c93fa1e75d4d37127e5eda0118f9/addons/purchase_stock/models/purchase_order_line.py#L94-L98
The advantage of this proposition being that we will never update the moves because of move updates.

I propose this because it is really hard to see waht would be the impact of your change. It is functionally quite opaque at least to me to understand in which use case we would endup at the line of your change and satisfy the if condition or not. 

Let me know what you think :)

josv-odoo:
@lase-odoo Thank you for the review! 
The context solution is indeed safer ! :military_helmet:

addons/purchase_stock/tests/test_create_picking.py
```diff
@@ -71,6 +74,21 @@ def test_00_create_picking(self):
...
+            move.quantity = 1.0
+        receipt_2 = receipt_2_form.save()
+        Form.from_action(self.env, receipt_2.button_validate()).save().process()
+        backorder = receipt_2.backorder_ids
+        self.assertEqual(self.po.picking_ids, receipt_1 | receipt_2 | backorder)
+        self.assertRecordValues(receipt_2.move_line_ids, [
+            {'product_id': self.product_id_2.id, 'quantity': 1, 'picking_type_id': self.env.ref('stock.picking_type_in').id},
+            {'product_id': self.product_sofa_blue.id, 'quantity': 1, 'picking_type_id': self.env.ref('stock.picking_type_in').id}])
+        self.assertRecordValues(backorder.move_line_ids, [
+            {'product_id': self.product_id_2.id, 'quantity': 4, 'picking_type_id': self.env.ref('stock.picking_type_in').id}])
```
svs-odoo: Define a variable for the picking type, it will be shorter, easier to read and you don't need to call `ref` multiple times
```python
receipt_type_id = self.ref('stock.picking_type_in')
```
```suggestion
        self.assertRecordValues(receipt_2.move_line_ids, [
            {'product_id': self.product_id_2.id, 'quantity': 1, 'picking_type_id': receipt_type_id},
            {'product_id': self.product_sofa_blue.id, 'quantity': 1, 'picking_type_id': receipt_type_id}])
        self.assertRecordValues(backorder.move_line_ids, [
            {'product_id': self.product_id_2.id, 'quantity': 4, 'picking_type_id': receipt_type_id}])
```

addons/purchase_stock/tests/test_create_picking.py
```diff
@@ -71,6 +74,21 @@ def test_00_create_picking(self):
         self.assertEqual(self.po.incoming_picking_count, 2, 'New picking should be created')
         moves = self.po.order_line.mapped('move_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
         self.assertEqual(len(moves), 1, 'One moves should have been created')
+        receipt_2 = self.po.picking_ids - receipt_1
```
svs-odoo, line 78: I think it's a good idea to precise what is the tested case when multiple cases are tested in the same test, that way, it's easier to understand what was the point when looking back at the test in the future without need to try to guess it only by reading the code
```suggestion
        # In second receipt, add product which is not in the PO.
        receipt_2 = self.po.picking_ids - receipt_1
```

svs-odoo (approved):

PR #248437, opened 2026-02-12
[FIX] html_builder, website: prevent invalid fields and show notification instead of tracebacks

AlessandroLupo:
This PR improves the handling of invalid field values in the website editor, particularly for event and product names.

Before this PR, deleting a name in pages such as “/event”, “/product”, “/blog”, or entering invalid values (e.g., wrong prices in “/shop”) could lead to inconsistent behavior.
1. Due to `DeletePlugin.fillEmptyInlines`, deleting a title left a zero-width space in the DOM. The backend considers a single zero-width space as a valid string, allowing apparently empty names to be saved without validation errors.
2. When actual validation errors occurred, saving triggered a traceback popup. After closing it, the editor remained open, valid changes were already saved, and “discard” did not properly restore the previous state.

This PR introduces two improvements:

1. Embedded fields consisting of a single zero-width space are converted to empty strings before save.
2. Validation errors during save now display a sticky warning notification instead of a traceback. The editor closes correctly, valid changes are preserved and invalid fields are restored to their original values.

task-5387282

### DIFF
--- a/addons/html_builder/static/src/core/save_plugin.js
+++ b/addons/html_builder/static/src/core/save_plugin.js
@@ -3,6 +3,7 @@ import { withSequence } from "@html_editor/utils/resource";
 import { groupBy } from "@web/core/utils/arrays";
 import { uniqueId } from "@web/core/utils/functions";
 import { isZWS } from "@html_editor/utils/dom_info";
+import { _t } from "@web/core/l10n/translation";
 
 /** @typedef {import("plugins").CSSSelector} CSSSelector */
 /**
@@ -95,6 +96,15 @@ export class SavePlugin extends Plugin {
             );
             await this._save(groupedElements);
             skipAfterSaveHandlers = await shouldSkipAfterSaveHandlers();
+        } catch (error) {
+            if (error.exceptionName === "odoo.exceptions.ValidationError") {
+                this.services.notification.add(_t("Previous values restored."), {
+                    title: _t("One or more fields were not valid"),
+                    type: "warning",
+                });
+            } else {
+                throw error;
+            }
         } finally {
             if (!skipAfterSaveHandlers) {
                 this.trigger("on_saved_handlers");
--- a/addons/html_builder/static/src/core/save_plugin.js
+++ b/addons/html_builder/static/src/core/save_plugin.js
@@ -2,6 +2,7 @@ import { Plugin } from "@html_editor/plugin";
 import { withSequence } from "@html_editor/utils/resource";
 import { groupBy } from "@web/core/utils/arrays";
 import { uniqueId } from "@web/core/utils/functions";
+import { isZWS } from "@html_editor/utils/dom_info";
 
 /** @typedef {import("plugins").CSSSelector} CSSSelector */
 /**
@@ -47,7 +48,19 @@ export class SavePlugin extends Plugin {
         // Do not change the sequence of this resource, it must stay the first
         // one to avoid marking dirty when not needed during the drag and drop.
         on_prepare_drag_handlers: withSequence(0, this.ignoreDirty.bind(this)),
+        on_will_save_handlers: this.removeZWSPFromEmbeddedFields.bind(this),
     };
+
+    removeZWSPFromEmbeddedFields(editableEl) {
+        // Remove zero-width spaces left by DeletePlugin.fillEmptyInlines on
+        // embedded fields to prevent saving blank model fields.
+        const selector = '[data-oe-model]:not([data-oe-model="ir.ui.view"])';
+        for (const el of editableEl.querySelectorAll(selector)) {
+            if (isZWS(el)) {
+                el.innerHTML = "";
+            }
+        }
+    }
 
     setup() {
         this.canObserve = false;
--- a/addons/website/static/tests/builder/save.test.js
+++ b/addons/website/static/tests/builder/save.test.js
@@ -9,6 +9,7 @@ import {
     models,
     onRpc,
     patchWithCleanup,
+    makeServerError,
 } from "@web/../tests/web_test_helpers";
 import { WebsiteBuilderClientAction } from "@website/client_actions/website_preview/website_builder_action";
 import {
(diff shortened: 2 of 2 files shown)


addons/html_editor/models/ir_qweb_fields.py
```diff
@@ -85,7 +85,7 @@ def attributes(self, record, field_name, options, values=None):
         return attrs
 
     def value_from_string(self, value):
-        return value
+        return value if value != "\u200B" else None
```
elgemmy: would we ever end up with a sequence of zero-width elements? If so, would it make sense to generalize this check to reject any invisible/zero-width value?

addons/html_builder/static/src/core/save_plugin.js
```diff
@@ -58,6 +59,16 @@ export class SavePlugin extends Plugin {
             await Promise.all(this.getResource("before_save_handlers").map((handler) => handler()));
             await this._save();
             skipAfterSaveHandlers = await shouldSkipAfterSaveHandlers();
+        } catch (error) {
+            if (error.exceptionName?.includes("ValidationError")) {
+                this.services.notification.add(error.data.message, {
```
elgemmy: While testing locally and on runbot I found the message in the toast still technical as shown in the screenshot. 

My guess is that the technical message comes from `error.data.message`, doesn't it? 
Additionally, the spec asks for a user friendly message (Title cannot be blank. Restoring previous value). Could we show that fixed message instead for this case?

Image

addons/html_builder/static/src/core/save_plugin.js
```diff
@@ -58,6 +59,16 @@ export class SavePlugin extends Plugin {
             await Promise.all(this.getResource("before_save_handlers").map((handler) => handler()));
             await this._save();
             skipAfterSaveHandlers = await shouldSkipAfterSaveHandlers();
+        } catch (error) {
+            if (error.exceptionName?.includes("ValidationError")) {
```
elgemmy: I think this catch is a bit broad and would catch **every** `ValidationError`, not only empty titles?

elgemmy (commented):
Nice work on this PR 👍🏻 

I’ve left a few clarification/suggestion comments for consideration.
Regarding the tests, they seem to test the current logic, so I didn’t add specific comments there. If we decide to adjust the validation or error messages, we can update the tests accordingly as part of that change.

addons/html_builder/static/src/core/save_plugin.js
AlessandroLupo: You are right. This morning I chatted with LEBL and we agreed that it is OK to extend the change to every `ValidationError`. For example, it would catch badly formatted prices in the ecommerce product page.

addons/html_editor/models/ir_qweb_fields.py
AlessandroLupo: Good point. I was considering using a more complex logic to cover for other kinds of invisible texts (including, as you say, multiple zero-width spaces). In the end, I decided to stick with this simple version because, after all, the purpose of the commit is to fix the problems that could arise from [this code](https://github.com/odoo/odoo/blob/9c11065c65ed6c8972e8397b38a9a4de312bfa09/addons/html_editor/static/src/core/delete_plugin.js#L535), which only adds a single "\u200B" character. :)

addons/html_builder/static/src/core/save_plugin.js
AlessandroLupo: Right. `error.data.message` usually contains good text, but this particular message may look a bit overwhelming. The problem is that the exception that we catch in the builder contains only this text, and at the moment it doesn't have other properties that would allow us to build a custom message (the exception is raised  [here](https://github.com/odoo/odoo/blob/3656171994450d11151565efdb4b9dd0468cefa8/odoo/http/retrying.py#L82-L85)). I may keep looking for an easy solution to be able to build our own message, but I am not sure it is possible without changing something in the `orm` module. What's your opinion?

elgemmy: I don't fully get the reason behind using `error.data.message` here in that case if it doesn't have relevant info about the problem.
Wouldn't replacing this with a simple message work? e.g:
image

addons/html_builder/static/src/core/save_plugin.js
elgemmy: I see. Honestly, I still prefer more specific conditions for better UX. At least to give an indication to the user about what went wrong. wdyt?

addons/html_builder/static/src/core/save_plugin.js
AlessandroLupo: The point is that `error.data.message`  brings some info about what went wrong, after all. I agree that the message could be more clear, but in the end it is still telling the user that the problem occurrend on the name of a product. In case of an error on a product price, the message is actually clearer. 
But your proposal is maybe better, because it also gives us the possibility to control the message. I will do this. :+1:

addons/html_builder/static/src/core/save_plugin.js
AlessandroLupo: At the moment the `ValidationError` does not bring any info about what went wrong, apart form the textual error message in `error.data.message`. I will go with your proposal above, and write a custom message that will clearly state that an invalid field was entered and that previous value has been restored. Is it better, right?

addons/html_editor/models/ir_qweb_fields.py
FrancoisGe: Don't you think this should be handled by the frontend (the JS)? It seems strange to me that the backend has to check for “\u200B”. Are there other cases in the backend?

addons/html_builder/static/src/core/save_plugin.js
```diff
@@ -82,6 +83,15 @@ export class SavePlugin extends Plugin {
             );
             await this._save(groupedElements);
             skipAfterSaveHandlers = await shouldSkipAfterSaveHandlers();
+        } catch (error) {
+            if (error.exceptionName?.includes("ValidationError")) {
```
FrancoisGe: Isn't there a better way to check the type of error?
Or why we need to use includes and not "===" ?

AlessandroLupo: Here we can't just use `error instanceof RPCError` because we want to catch only validation errors and not every possible RPC error. I agree that `===` is better than `includes`, I changed that

addons/website/static/tests/builder/save.test.js
```diff
@@ -598,3 +599,54 @@ test("Modifying an element inside '.o_not_editable' should not mark this element
...
+
+test("Validation errors are caught on save", async () => {
+    onRpc("ir.ui.view", "save", () => {
+        const error = new RPCError();
+        error.exceptionName = "odoo.exceptions.ValidationError";
+        error.message = "ValidationError";
+        error.data = {
+            name: "odoo.exceptions.ValidationError",
+            message: "Validation Error",
+        };
```
FrancoisGe: You should use `makeServerError`

addons/website/static/tests/builder/save.test.js
```diff
@@ -598,3 +599,54 @@ test("Modifying an element inside '.o_not_editable' should not mark this element
...
+    expect.errors(1);
+    onRpc("ir.ui.view", "save", () => {
+        const error = new RPCError();
+        error.exceptionName = "odoo.exceptions.AnotherError";
+        error.message = "Another Error";
+        error.data = {
+            name: "AnotherError",
+            message: "Another Error",
+        };
+        return Promise.reject(error);
```
FrancoisGe: same `makeServerError`

PR #248441, opened 2026-02-12
[FIX] l10n_{be,ro,dk} : remove company_registry when VAT is emptied

MAGREMENT:
# How to reproduce:
- Create a company from Belgium, Romania or Denmark
- Add a VAT to the company
- Set the Company of a Contact to the one created before and save
- Remove the company and the VAT from the contact and save

# The problem:
The company_registry field (which is visible using Studio) is still set to the company's VAT. In those countries, company_registry is equal to the VAT without the country's prefix. So if the VAT is empty, company_registry should be too. This sometimes caused problems with double detection (cf. Task 5921966)

# Why:
There was no logic put in place to remove the value for the company_registry when removing the VAT

# Note
Some tests in l10n_ro_edi checked that exporting an invoice of a partner with a company_registry but no VAT produced the same output as a partner with a company_registry and a VAT. We removed those tests because a partner with no VAT but a company_registry should never happen since in Romania, VAT = "RO" + CUI (https://static.anaf.ro/static/10/Anaf/legislatie/OrdineMFP/901.htm).

Also, the file "ciusro_out_invoice_no_prefix_company_registry.xml" was the exact same as "ciusro_out_invoice_no_prefix_vat.xml"

opw-5926665

---


guva-odoo (approved):
LGTM thanks!, can you just update your commit message by using "we" instead of "I"?

addons/l10n_be/models/res_partner.py
```diff
@@ -14,7 +14,10 @@ def _compute_company_registry(self):
         # OVERRIDE
         # If a belgian company has a VAT number then its company registry is its VAT Number (without country code).
         super()._compute_company_registry()
-        for partner in self.filtered(lambda p: p._deduce_country_code() == 'BE' and p.vat):
-            vat_country, vat_number = self._split_vat(partner.vat)
-            if vat_country in ('BE', '') and self._check_vat_number('BE', vat_number):
-                partner.company_registry = vat_number
+        for partner in self.filtered(lambda p: p._deduce_country_code() == 'BE'):
+            if partner.vat:
```
malb-odoo, line 18: I would add a comment to explain the case, it's not obvious that it's when you remove the vat that we should reset the company registry. (Did you check functionally with a po though ?)

malb-odoo (commented):
Thanks for the pr, one small nitpick that could be done is adding a comment to explain why we do that specifically since company registry is stored, the base value of company_registry is false when not set so i was mind fucked at first ahah

malb-odoo:
> Thank you for the review @malb-odoo. I did not check the functionality with a PO. Do you think I should contact the PO of every localization for this PR (since this applies to 3 localization) ?

i think you can send a message on the thread "question for a po-naire" :smile: https://discord.com/channels/678381219515465750/1099994955830796348

And you have some conflict :eyes:

PR #248443, opened 2026-02-12
[FIX] account: portal bill report no proforma

guva-odoo:
When accessing to a bill via the portal, we display a proforma vendor
bill, same when we download from the portal, it should not be the case

Steps:

- Create and confirm a vendor bill
- Via the menu action, click on "Share" and copy the link
- Open the link
-> a proforma vendor bill is displayed
- Download the bill
-> we get a proforma vendor bill

Fix:

Allow proforma only for invoices/credit notes

opw-5882732


william-andre (approved):
@robodoo r+

PR #248451, opened 2026-02-12
[FIX] web: x2many: prevent crash with many create

hubvd:
Steps to reproduce
==================

- Edit the Vendor Bill form view to set a limit of 2 on the invoice_line_ids field.
- Create 2 purchase orders with the same vendor and 4 products
- Create a vendor bill
- Set the same vendor
- Using the autocomplete field, select the first purchase order
- Save the form
- Select the other purchase order in the autocomplete field
- Delete the record before the last one

=> Got duplicate key in t-foreach: datapoint_12

Cause of the issue
==================

Before inserting the last 4 lines:
`this._currentIds = [1, 2, 3, 4]`

After the second insert, we have
`this._currentIds = [1, 2, virtual_1, 3, virtual 2, virtual_3, virtual_4, 4]`

Only the first record is inserted at the correct place, following ones are off by one.

opw-5264594

### DIFF
--- a/addons/web/static/src/model/relational_model/static_list.js
+++ b/addons/web/static/src/model/relational_model/static_list.js
@@ -477,7 +477,7 @@ export class StaticList extends DataPoint {
                     const record = this._createRecordDatapoint(command[2], { virtualId });
                     this.records.push(record);
                     addOwnCommand([CREATE, virtualId]);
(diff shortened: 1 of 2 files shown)


aab-odoo (approved):
robodoo r+

PR #248454, opened 2026-02-12
[FIX] pos_event: use pricelist for event tickets in pos

macocianradu:
Event tickets in POS would have their price locked to the price defined in the event itself. They would be filtered out of any price recalculation inside the POS to keep the POS from recalculating the price based on the `product_template` and to keep the price defined in the event itself.

This PR will add event tickets back into price recalculation. It will set the price to the price defined inside the event if no pricelist is applicable, or use the pricelist to calculate the price if there is one applicable.

Task-[5092613](https://www.odoo.com/odoo/project/1737/tasks/5092613)

---

### DIFF
--- a/addons/pos_event/static/src/app/screens/product_screen/product_screen.js
+++ b/addons/pos_event/static/src/app/screens/product_screen/product_screen.js
@@ -188,10 +188,19 @@ patch(ProductScreen.prototype, {
 
         for (const [ticketId, data] of Object.entries(result.byRegistration)) {
             const ticket = this.pos.models["event.event.ticket"].get(parseInt(ticketId));
+            const priceExtra = ticket.price - ticket.product_id.lst_price;
+            const priceUnit = ticket.product_id.getPrice(
+                this.pos.getOrder().pricelist_id,
+                1,
+                priceExtra,
+                false,
+                ticket.product_id
+            );
             const line = await this.pos.addLineToCurrentOrder({
                 product_id: ticket.product_id,
                 product_tmpl_id: ticket.product_id.product_tmpl_id,
-                price_unit: ticket.price,
+                price_unit: priceUnit,
+                price_extra: priceExtra,
                 price_type: "original",
                 qty: data.length,
                 event_ticket_id: ticket,
--- a/addons/pos_event/static/src/app/models/pos_order.js
+++ b/addons/pos_event/static/src/app/models/pos_order.js
@@ -6,8 +6,4 @@ patch(PosOrder.prototype, {
     get eventRegistrations() {
         return this.lines.flatMap((line) => line.event_registration_ids);
     },
-    getLinesToCompute() {
-        // Override to ensure orderline price of event tickets are not recomputed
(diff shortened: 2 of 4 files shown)


loukwets (changes requested):
I'm not sure this is the right fix, by taking a quick look, the issue of calling getPrice to calculate the ticket price is that is takes the `variant.lst_price` to compute the price and not the actual "event.event.ticket" ticket price. So your solution will still compute the wrong price for the ticket, you need to manage to compute the price based on the ticket price. 

Also when finished could you write tests ?

addons/pos_event/static/src/app/screens/product_screen/product_screen.js
loukwets: What do you think about:
```suggestion
                price_extra: ticket.price - ticket.product_id.lst_price,
```
I think this could replace all your changes, since all the rest of the computations will be done with price extra

loukwets (commented):
Thanks for the test !

addons/pos_event/static/src/app/screens/product_screen/product_screen.js
macocianradu:
```
if (line.event_ticket_id) {
   line.product_id.lst_price = line.event_ticket_id.price;
 }
```
That's kind of what I do. I just assign the ticket price directly to the lst_price instead of the extra price. It feels more accurate to modify the list price instead of treat it as extra.

loukwets: I discussed it with moda, you cannot modify the product linked to the ticket, you should only modify the "dummy product" passed in argument of the method and created here: `pos_event/static/src/app/services/pos_store.js` you should then adapt the addLineToOrder to give that product to the line. Or you can make my suggestion and this single change is sufficient to resolve the issue.

loukwets (approved):

addons/pos_event/static/src/app/screens/product_screen/product_screen.js
stva-odoo, line 199: I noticed that when you add a ticket while a pricelist is applied, the original price is used instead of the price from the applied pricelist.

I suggest adding this to your code:

```javascript
            const priceExtra = ticket.price - ticket.product_id.lst_price;
            const priceUnit = ticket.product_id.getPrice(
                this.pos.getOrder().pricelist_id,
                1,
                priceExtra,
                false,
                ticket.product_id
            );
            const line = await this.pos.addLineToCurrentOrder({
                product_id: ticket.product_id,
                product_tmpl_id: ticket.product_id.product_tmpl_id,
                price_unit: priceUnit,
                price_extra: priceExtra,
                price_type: "original",
                qty: data.length,
                event_ticket_id: ticket,
                event_slot_id: slotSelected,
            });

```

Could you also add a test to verify that when adding a product, the applied pricelist is correctly used?

stva-odoo (changes requested):

stva-odoo (approved):
robodoo r+

PR #248466, opened 2026-02-12
[FIX] account: disallow untrusted accounts on is_inbound() invoices + log bank account

gawa-odoo:
Revert and fix in one commit, to avoid issues with empty node

Fixed some tests and remove the computation logic from account_move_reversal wizard, to rely on existing compute method.
Log them instead of trying to unarchive/create them.

task-5923591


gawa-odoo:
@robodoo fw=no

clbr-odoo (approved):
@robodoo r+

PR #248469, opened 2026-02-12
[FIX] mrp_subcontracting_landed_costs: compute correct landed cost su…

plha-odoo:
…bcontracted

**Problem:**
Landed cost added on the receipt of a subcontracted product
do not increase the valuation of the product and do not create 
account move lines.

**Steps to reproduce:**
- create a tracked product with avco perpetual category
- create a subcontracted bom for this product with no comp
- create and confirm a PO for 10 unit of this product at a unit price 
of 1$ with the same partner as the subcontractor of the bom
- validate the receipt
- navigate to inventory/operations/adjustments/landed costs
- create a new landed cost
- select the receipt from the PO
- add a landed cost of 10$ and validate
- navigate to inventory/reporting/stock
- search for your product and click on the unit cost

**Current behavior:**
1) the valuation of the product was not increased by the value 
of the landed cost
2) open journal items : no account move lines were created for the 
landed cost

**Expected behavior:**
1) the valuation of the product should have been increased:
in the unit cost view, the SBC move should have gone from a 
value of 10 to 20
2) account move lines should have been created with a value 
of 10

**Cause of the issue:**
both issues come from the fact that when creating the 
stock valuation adjustment line, the move linked is the receipt move
when it should be the move of the subcontracted MO linked to the 
receipt.

**fix:** 
if we create the adjustement line with move_id as the move of the MO
(instead of the move of the receipt as it is the case currently) : 
when button_validate is called on the landed cost : 
- when using the remaining quantity, it will be the correct one 
(in our case 10, instead of 0 for the move of the receipt because it's 
actually an internal move) so the account move line are going to 
be created
https://github.com/odoo/odoo/blob/9c85d7265d7b1ca0b212f46c20aa1e8119c33a22/addons/stock_landed_costs/models/stock_landed_cost.py#L129-L130
https://github.com/odoo/odoo/blob/9c85d7265d7b1ca0b212f46c20aa1e8119c33a22/addons/stock_landed_costs/models/stock_landed_cost.py#L372-L373
which solves problem 2) 

-  when calling _set_value on the move (which will be the move of 
the MO thanks to this fix), 
https://github.com/odoo/odoo/blob/064407d32f998ceb08601f9e0a6356c94ad10347/addons/stock_landed_costs/models/stock_landed_cost.py#L152
get_value_data will call _get_value_from_extra,
https://github.com/odoo/odoo/blob/9c85d7265d7b1ca0b212f46c20aa1e8119c33a22/addons/stock_account/models/stock_move.py#L392
which uses _get_landed_cost to fetch the landed cost
https://github.com/odoo/odoo/blob/9c85d7265d7b1ca0b212f46c20aa1e8119c33a22/addons/stock_landed_costs/models/stock_move.py#L18
before this fix the landed cost created from the receipt were linked to 
the receipt move so they were not fetched inside _get_landed_cost
which caused problem 1) but now the move_id of the adjustment lines 
is the move of the MO so they are fetched inside _get_landed_cost
https://github.com/odoo/odoo/blob/9c85d7265d7b1ca0b212f46c20aa1e8119c33a22/addons/stock_landed_costs/models/stock_move.py#L7-L12
So now the adjustment lines do impact the valuation of the move of 
the MO which solves problem 1)

opw-5723126

### DIFF
--- a/addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
+++ b/addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
@@ -1,0 +1,18 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+from odoo.tools import OrderedSet
+
+
+class StockLandedCost(models.Model):
+    _inherit = 'stock.landed.cost'
+
+    def _get_targeted_move_ids(self):
+        res = super()._get_targeted_move_ids()
+        target_moves_ids = OrderedSet()
+        for move in res:
+            if move.is_subcontract:
+                target_moves_ids.update(move.move_orig_ids.ids)
+            else:
+                target_moves_ids.add(move.id)
+        return self.env['stock.move'].browse(target_moves_ids)
--- a/addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
+++ b/addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
@@ -2,14 +2,15 @@ from unittest import skip
 
 from odoo.exceptions import ValidationError
 from odoo.tests import Form, tagged
+from odoo import Command
 
 from odoo.addons.mrp_subcontracting.tests.common import TestMrpSubcontractingCommon
 
 
 @tagged('post_install', '-at_install')
-@skip('Temporary to fast merge new valuation')
 class TestSubcontractingLandedCosts(TestMrpSubcontractingCommon):
 
+    @skip('Temporary to fast merge new valuation')
     def test_subcontracting_landed_cost_receipts_flow(self):
         """
             This test verifies that landed costs can be applied to subcontracting receipts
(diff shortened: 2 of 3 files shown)


addons/mrp_subcontracting_landed_costs/models/stock_landed_cost.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = 'stock.landed.cost'
+
+    def _get_targeted_move_ids(self):
+        res = super()._get_targeted_move_ids()
+        target_moves = self.env['stock.move']
+        for move in res:
+            if move.is_subcontract:
+                target_moves |= move.move_orig_ids
+            else:
+                target_moves |= move
```
ticodoo: Instead of re-creating the recordset one record at a time (i.e. inefficient/slow), you should probably collect the ids in an `OrderedSet` and then return the `browse` result

addons/mrp_subcontracting_landed_costs/tests/test_subcontracting_landed_costs.py
```diff
@@ -134,3 +135,74 @@ def test_subcontracting_landed_cost_receipts_flow(self):
...
+            'move_line_ids': [Command.create({
+                'quantity': 3,
+                'product_id': self.finished.id,
+            })]
+        })
+        move.picked = True
+        move._action_done()
+
+        # create a landed cost for the incoming picking and validate it
+        default_vals = self.env['stock.landed.cost'].default_get(list(self.env['stock.landed.cost'].fields_get()))
```
ticodoo: What? why? Just make the record. I think the code you maybe copied this from in the same file is garbage that was copied from 8 year old code.

ticodoo (commented):
Fix seems fine, just needs some polishing

PR #222156, opened 2025-08-07
[PERF] core: Environment kept longer

kmagusiak:
Description of the issue/feature this PR addresses:

Keep environments using hard references instead of weak references so that they can be reused later.

On the performance side, lookup of an environment costs ~30% of the time it takes to look it up and create a new one. This is not a huge gain, however we keep all other lazy properties that were already computed: like the user, accessed fields, etc.

Since there is quite some code that would just create an environment for a single operation (think of `active_test=False`), it is not uncommon that there will soon be another similar operation.

In order to avoid keeping too many Environments, we add a method `_dereference_envs` which removes environments that are no longer accessible through the transaction. It is called during cleanup of the transaction or when a lot of environments exist. Most simple transactions use less than 8 environments, but can double when using multiple languages.

<details><summary>Performance</summary>
<p>

```py
env = Environment(Registry("odoo").cursor(), 1, {})
uenv = env(user=55)

%timeit env(user=1)  # no change
%timeit env(user=4)
%timeit uenv(su=True)
%timeit uenv(su=True, context={'hello': 5})
%timeit env(su=True, context={'active_test': False})

%timeit for a in "hello": env(context={'x': a})
%timeit for a in "hello world ABCDEF": env(context={'y': a})
%timeit for a in range(1000): env(user=a)

%timeit uenv(context={'active_test': False}).user
%timeit uenv(context={'active_test': False}).companies

tmp = [env(user=a) for a in range(10)]
%timeit [env(user=a) for a in range(10)]
%timeit env(user=6)
tmp = [env(user=a) for a in range(1000)]
%timeit [env(user=a) for a in range(1000)]
del tmp
```

```
OLD:

1.41 μs ± 15.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
2.82 μs ± 54 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
2.97 μs ± 75.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
2.92 μs ± 54.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
2.85 μs ± 33.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

14.3 μs ± 84.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
51.2 μs ± 359 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[...]

PR #248470, opened 2026-02-12
[IMP] base: add python 3.14 support for safe_eval

mwath:
This PR adds support for Python 3.14 new opcodes, as well as improving test coverage for safe_eval.
Python 3.14 will be the default version of Ubuntu 26.04, as mentioned in the release milestones: https://discourse.ubuntu.com/t/resolute-raccoon-release-schedule/47198


odoo/tools/safe_eval.py
```diff
@@ -104,6 +106,8 @@ def to_opcodes(opnames, _opmap=opmap):
     'RETURN_CONST',
     # 3.13
     'TO_BOOL',
+    # 3.14 https://docs.python.org/3/whatsnew/3.14.html#cpython-bytecode-changes
+    'NOT_TAKEN', 'LOAD_SMALL_INT',
```
antonrom1: NOT_TAKEN for const opcodes? It's a branching opcode, do we need branching for building literal values?

antonrom1 (commented):
The opcodes whitelist looks good to me, can you also take a look at new frame attributes? There is `f_generator`, I think we can safely ban it in `_UNSAFE_ATTRIBUTES`

odoo/tools/safe_eval.py
```diff
@@ -183,6 +187,15 @@ def to_opcodes(opnames, _opmap=opmap):
     'STORE_FAST_STORE_FAST', 'STORE_FAST_LOAD_FAST',
     'CONVERT_VALUE', 'FORMAT_SIMPLE', 'FORMAT_WITH_SPEC',
     'SET_FUNCTION_ATTRIBUTE',
+    # Added in 3.14, used to load a variable that is the compiler can prove
+    # that the reference in the frame outlives the reference loaded onto the
+    # stack. E.g. args & kwargs.
```
xmo-odoo:
```suggestion
    # 3.14
```

odoo/tools/safe_eval.py
```diff
@@ -183,6 +187,15 @@ def to_opcodes(opnames, _opmap=opmap):
...
+    'LOAD_FAST_BORROW', 'LOAD_FAST_BORROW_LOAD_FAST_BORROW',
```
xmo-odoo:
```suggestion
    'LOAD_FAST_BORROW', 'LOAD_FAST_BORROW_LOAD_FAST_BORROW',  # LOAD_FAST optimizations
```

odoo/tools/safe_eval.py
```diff
@@ -84,6 +84,8 @@ def to_opcodes(opnames, _opmap=opmap):
     'STORE_ATTR', 'DELETE_ATTR',
     # no reason to allow this
     'STORE_GLOBAL', 'DELETE_GLOBAL',
+    # Deals with dunder methods
+    'LOAD_SPECIAL',
```
xmo-odoo:
```suggestion
```
According to the changelog this is used for the `with` statement, which we do not support in safe_eval (no `SETUP_WITH`, `WITH_EXCEPT_START`, or `BEFORE_WITH`). 

Is this used for any statement we actually support? And if so which ones?

odoo/tools/safe_eval.py
```diff
@@ -183,6 +187,15 @@ def to_opcodes(opnames, _opmap=opmap):
     'STORE_FAST_STORE_FAST', 'STORE_FAST_LOAD_FAST',
     'CONVERT_VALUE', 'FORMAT_SIMPLE', 'FORMAT_WITH_SPEC',
     'SET_FUNCTION_ATTRIBUTE',
+    # Added in 3.14, used to load a variable that is the compiler can prove
+    # that the reference in the frame outlives the reference loaded onto the
+    # stack. E.g. args & kwargs.
+    'LOAD_FAST_BORROW', 'LOAD_FAST_BORROW_LOAD_FAST_BORROW',
+    'POP_ITER',  # comprehensions
```
xmo-odoo: It's for iteration in general, not comprehensions (python/cpython#128445)

odoo/tools/misc.py
```diff
@@ -1026,6 +1026,9 @@ def __repr__(self):
     def intersection(self, *others):
         return reduce(OrderedSet.__and__, others, self)
 
+    def copy(self):
+        return OrderedSet(self._map)
```
xmo-odoo: Why? The commit message is unhelpful, it mentions weaksets but the only weakset directly used in Odoo is in websocket, and stores websocket instances.

xmo-odoo (changes requested):

odoo/tools/safe_eval.py
mwath: > Is this used for any statement we actually support? And if so which ones?

No, but given the description of the instruction, I thought it would be a good idea to add it to the blacklist as a safeguard.
The implementation loads a dunder attribute onto the stack: https://github.com/python/cpython/blob/b1946887032da34eff4acebed9c03cf74abf4373/Python/bytecodes.c#L3520
And while the list of attributes is [hardcoded](https://github.com/python/cpython/blob/1564e231aae7afad5b9b19a277d1efff2b840ad2/Python/ceval.c#L360), we have no guarantee that the opcode will be limited to `with` statements in future versions.

Let me know if you think this is unnecessary

odoo/tools/misc.py
mwath: There was one before: https://github.com/odoo/odoo/pull/222156 ([here](https://github.com/odoo/odoo/pull/222156/changes/4ad4b95b6e1dadd31c558b8a531524f6b600e1de#diff-f1d221d99a41c57054ddd9967e3dec5926bbadb2595323a674b01f0fb4dc91feL573-L574))
It's no longer needed, I'll remove the commit.

xmo-odoo: :+1:

odoo/tools/safe_eval.py
xmo-odoo: I think it should be left out, if the feature gets expanded to other attributes we can revisit it at that latter point, but currently (and officially) this is just support for `with` / `async with`.

xmo-odoo (approved):
WFM, opcodes list seems fine

@robodoo override=ci/security

PR #248471, opened 2026-02-12
[FIX] hr_attendance: fix access issue

Bertrand2:
The commit https://github.com/odoo/odoo/commit/d115cca1296c0bd20a2875bdaadf2a9136c33dda introduced the opportunity for attendance managers to edit their subordinates attendances, but the condition was lacking parenthesis, leading to an unexpected behaviour.

This commit changes the condition so that it is possible both for a manager and an officer to do the edition.
It also rewrite the tests in order to be more thorough.

No related task


amah-odoo (approved):
Thanks for your work 🚀 
@robodoo r+

PR #189370, opened 2024-12-02
[IMP] profiler: add a wizard to control the rendering of speedscope

Beekan:
This pr aims to improve the profiling, and mainly speedscope rendering experience.

Right now, visualizing a speedscope result:
- Gives a static result. The  makes the usage of some options like constant time impossible without editing the code. It also generate to much different types of output at the same time making it slow and leads faster to memory error.
- Needs to enable the profiling on the database to be able to visualize the result.  This can be painful mainly when trying to access result after some time, or just after profiling in python.

This pr removes the need to enable the profiling to visualize the results. 

This pr adds a wizard to enable or disable some options.

As a bonus it is now possible to combine multiple profile in the same view (experimental)

PR #248472, opened 2026-02-12
[IMP] profiler: add a wizard to control the rendering of speedscope

Beekan:
This pr aims to improve the profiling, and mainly speedscope rendering experience.

Right now, visualizing a speedscope result:

Gives a static result. The makes the usage of some options like constant time impossible without editing the code. It also generate to much different types of output at the same time making it slow and leads faster to memory error.
Needs to enable the profiling on the database to be able to visualize the result. This can be painful mainly when trying to access result after some time, or just after profiling in python.
This pr removes the need to enable the profiling to visualize the results.

This pr adds a wizard to enable or disable some options.

As a bonus it is now possible to combine multiple profile in the same view (experimental)

Backport of https://github.com/odoo/odoo/pull/189370

---

### DIFF
--- a/addons/web/controllers/profiling.py
+++ b/addons/web/controllers/profiling.py
@@ -28,15 +28,16 @@ class Profiling(Controller):
     ], type='http', sitemap=False, auth='user', readonly=True)
     def speedscope(self, profile=None, action=False, **kwargs):
         if not profile:
-            return
+            raise request.not_found()
         profile_str = profile
-        profile = request.env['ir.profile'].browse(int(profile_str))
+        profiles = request.env['ir.profile'].browse(int(p) for p in profile.split(','))
         if not kwargs and not action:
             context = {
-                'profile': profile,
+                'profile_str': profile_str,
+                'profiles': profiles,
             }
             return request.render('web.config_speedscope_index', context)
-        speedscope_result = profile._generate_speedscope(profile._parse_params(kwargs))
+        speedscope_result = profiles._generate_speedscope(profiles._parse_params(kwargs))
         if action == 'download_json':
             headers = [
                 ('Content-Type', 'application/json'),
@@ -46,7 +47,7 @@ class Profiling(Controller):
             return request.make_response(speedscope_result, headers)
         icp = request.env['ir.config_parameter']
         context = {
-            'profile': profile,
+            'profiles': profiles,
             'speedscope_base64': base64.b64encode(speedscope_result).decode('utf-8'),
             'url_root': request.httprequest.url_root,
             'cdn': icp.sudo().get_param('speedscope_cdn', "https://cdn.jsdelivr.net/npm/speedscope@1.13.0/dist/release/")
--- a/addons/web/tests/test_profiler.py
+++ b/addons/web/tests/test_profiler.py
@@ -55,10 +55,10 @@ class TestProfilingWeb(ProfilingHttpCase):
         self.assertTrue(res['result']['session'])
         self.assertEqual(last_profile, self.env['ir.profile'].search([], limit=1, order='id desc'), "profiling route shouldn't have been profiled")
         # Profile a page
-        res = self.url_open('/web/speedscope')  # profile a light route
+        res = self.url_open('/web/login')  # profile a light route
         new_profile = self.env['ir.profile'].search([], limit=1, order='id desc')
         self.assertNotEqual(last_profile, new_profile, "A new profile should have been created")
-        self.assertEqual(new_profile.name, '/web/speedscope?')
+        self.assertEqual(new_profile.name, '/web/login?')
 
     def test_profile_test_tool(self):
         with self.profile():
--- a/odoo/addons/base/models/ir_profile.py
+++ b/odoo/addons/base/models/ir_profile.py
@@ -58,27 +58,47 @@ class IrProfile(models.Model):
 
     def _parse_params(self, params):
         return {
-            'constant_time' : str2bool(params.get('constant_time', False)),
-            'aggregate_sql' : str2bool(params.get('aggregate_sql', False)),
-            'use_context' : str2bool(params.get('use_execution_context', False)),
-            'combined_profile' : str2bool(params.get('combined_profile', False)),
-            'sql_no_gap_profile' : str2bool(params.get('sql_no_gap_profile', False)),
-            'sql_density_profile' : str2bool(params.get('sql_density_profile', False)),
-            'frames_profile' : str2bool(params.get('frames_profile', False)),
+            'constant_time': str2bool(params.get('constant_time', False)),
+            'aggregate_sql': str2bool(params.get('aggregate_sql', False)),
+            'use_context': str2bool(params.get('use_execution_context', False)),
+            'combined_profile': str2bool(params.get('combined_profile', False)),
+            'sql_no_gap_profile': str2bool(params.get('sql_no_gap_profile', False)),
+            'sql_density_profile': str2bool(params.get('sql_density_profile', False)),
+            'frames_profile': str2bool(params.get('frames_profile', False)),
             'profile_aggregation_mode': params.get('profile_aggregation_mode', 'tabs'),
         }
 
     def _generate_speedscope(self, params):
-        sp = Speedscope(init_stack_trace=json.loads(self.init_stack_trace))
-        if self.sql:
-            sp.add('sql', json.loads(self.sql), params['aggregate_sql'])
-        if self.traces_async:
-            sp.add('frames', json.loads(self.traces_async))
-        if self.traces_sync:
-            sp.add('settrace', json.loads(self.traces_sync))
+        init_stack_trace = self[0].init_stack_trace
+        for record in self:
+            if record.init_stack_trace != init_stack_trace:
+                raise UserError(_('All profiles must have the same initial stack trace to be displayed together.'))
+        sp = Speedscope(init_stack_trace=json.loads(init_stack_trace))
+        for profile in self:
+            if (params['sql_no_gap_profile'] or params['sql_density_profile'] or params['combined_profile']) and profile.sql:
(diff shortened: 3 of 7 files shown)


addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
-            return request.not_found()
+    @route([
+        '/web/speedscope',
+        '/web/speedscope/<profile>',
+    ], type='http', sitemap=False, auth='user', readonly=True)
+    def speedscope(self, profile=None, action=False, **kwargs):
+        if not profile:
+            raise request.not_found()
+        profile_str = profile
+        profiles = request.env['ir.profile'].browse(int(p) for p in profile.split(','))
```
Julien00859, line 33: The "normal" may to provide multiple arguments would be `/web/speedscope?profile=1&profile=2`. What does that URL do? (sorry no time here and now to test it myself)

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
-    @route(['/web/speedscope', '/web/speedscope/<model("ir.profile"):profile>'], type='http', sitemap=False, auth='user')
-    def speedscope(self, profile=None):
-        # don't server speedscope index if profiling is not enabled
-        if not request.env['ir.profile']._enabled_until():
-            return request.not_found()
+    @route([
+        '/web/speedscope',
+        '/web/speedscope/<profile>',
+    ], type='http', sitemap=False, auth='user', readonly=True)
+    def speedscope(self, profile=None, action=False, **kwargs):
```
Julien00859, line 29:
```suggestion
    def speedscope(self, profile='', action='', **kwargs):
```
I'm always uneasy when the default values in controllers are non-string, cause it always is possible to call the function passing query-string parameters, that will always be parsed as strings by the system. e.g. `/web/speedscope?profile=&action=`. Using a falsy empty string makes for a better default value.

Julien00859, line 29: I don't like `**kwargs`, please list all the expected arguments and their default values. If it is not possible, please at least add a docstring to list them.

addons/web/controllers/profiling.py
Julien00859: FWIW the `split(',')` pattern never was a common way to deal with URL arguments. It's pretty much only done at Odoo, and IMO it's bad.

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
+                'profile_str': profile_str,
+                'profiles': profiles,
+            }
+            return request.render('web.config_speedscope_index', context)
+        speedscope_result = profiles._generate_speedscope(profiles._parse_params(kwargs))
+        if action == 'download_json':
+            headers = [
+                ('Content-Type', 'application/json'),
+                ('X-Content-Type-Options', 'nosniff'),
+                ('Content-Disposition', content_disposition(f'profile_{profile_str}.json')),
```
Julien00859, line 45: It is a pretty bad idea to use a user-defined name as filename. By the way it is likely to fail because `profile_str` can contain commas, but commas are explicitly disallowed by `content_disposition`.

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
+            ]
+            return request.make_response(speedscope_result, headers)
```
Julien00859, line 47: There's `make_json_response`.

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
+            context = {
+                'profile_str': profile_str,
+                'profiles': profiles,
+            }
+            return request.render('web.config_speedscope_index', context)
+        speedscope_result = profiles._generate_speedscope(profiles._parse_params(kwargs))
+        if action == 'download_json':
+            headers = [
+                ('Content-Type', 'application/json'),
+                ('X-Content-Type-Options', 'nosniff'),
```
Julien00859, line 44: This header is already added by the framework. see `set_csp`.

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
+        profile_str = profile
+        profiles = request.env['ir.profile'].browse(int(p) for p in profile.split(','))
+        if not kwargs and not action:
+            context = {
+                'profile_str': profile_str,
+                'profiles': profiles,
+            }
+            return request.render('web.config_speedscope_index', context)
+        speedscope_result = profiles._generate_speedscope(profiles._parse_params(kwargs))
+        if action == 'download_json':
```
Julien00859, line 41: So it is gonna be `/web/speedscope/1?action=download_json` and it gonna return a file named `profile_1.json`. Why not have the URL be `/web/speedscope/1.json`?

```py
@route(["/web/speedscope/<profile>.<ext>", ...], ...)
def ...(...):
    if ext not in ('', 'html', 'json'):
        raise request.not_found()
```

addons/web/controllers/profiling.py
```diff
@@ -20,15 +22,39 @@ def profile(self, profile=None, collectors=None, **params):
...
+            'speedscope_base64': base64.b64encode(speedscope_result).decode('utf-8'),
             'url_root': request.httprequest.url_root,
             'cdn': icp.sudo().get_param('speedscope_cdn', "https://cdn.jsdelivr.net/npm/speedscope@1.13.0/dist/release/")
         }
-        return request.render('web.view_speedscope_index', context)
+        response = request.render('web.view_speedscope_index', context)
+        if action == 'download_html':
+            response.headers['Content-Disposition'] = content_disposition(f'profile_{profile_str}.html')
+            response.headers['X-Content-Type-Options'] = 'nosniff'
+            response.headers['Content-Type'] = 'text/html'
```
Julien00859, line 59: I don't get it why you always return a html document, but only care to add those headers when the user added `?action=download_html`.
```suggestion
        response.headers['Content-Disposition'] = content_disposition(f'profile_{profile_str}.html')
        response.headers['X-Content-Type-Options'] = 'nosniff'
        response.headers['Content-Type'] = 'text/html'
```

addons/web/tests/test_profiler.py
```diff
@@ -55,10 +55,10 @@ def test_profiling_enabled(self):
         self.assertTrue(res['result']['session'])
         self.assertEqual(last_profile, self.env['ir.profile'].search([], limit=1, order='id desc'), "profiling route shouldn't have been profiled")
         # Profile a page
-        res = self.url_open('/web/speedscope')  # profile a light route
+        res = self.url_open('/web/login')  # profile a light route
```
Julien00859, line 58: /web/health would be a light route

addons/web/views/speedscope_template.xml
```diff
@@ -20,6 +20,12 @@
                     crossorigin="anonymous"
                     integrity="sha256-CvDqAOMjq0Sv/D59O5JSbzzXoClZ3rptt6ts8D/6CWw="
                 ></script>
+                <script>
+                const b64 = '<t t-esc="speedscope_base64"/>';
+                document.addEventListener("DOMContentLoaded", () => {
+                    speedscope.loadFileFromBase64('Profile', b64);
+                });
```
Julien00859, line 27: Must the HTML document be standalone? If not could we find a better solution than inlining the entire blob... as base64? Best would be to download a cachable URL.

odoo/addons/base/models/ir_profile.py
```diff
@@ -49,17 +50,55 @@ def _gc_profile(self):
         return self.sudo().search(domain).unlink()
 
     def _compute_speedscope(self):
+        # The params variable is done to control input from the user
+        # When expanding this, it should be select from an enum to input only the correct values
+        params = self._parse_params(self.env.context)
```
Julien00859, line 55: So we're passing about 8 function arguments, as string, via the context?! Crazy!

addons/web/views/speedscope_template.xml
Xavier-Do: Making it standalone is very helpful to backup/share profiles that can be open via a single click.

addons/web/controllers/profiling.py
Beekan: Because the flow is not always to download the file. Adding the header content-disposition would download while I would just want to view the file. To have a common flow would mean to pass either attachment or inline with the content disposition depending on the flow which would essentially be the if statement again

addons/web/controllers/profiling.py
Beekan: Yeah i didnt know that i'll remove it

addons/web/views/speedscope_template.xml
Julien00859: In that case, could we leverage compression so the final html document is lighter? It is likely that the speedscope data is gonna compress very well.

Seems possible in js: https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API (use gzip)

pseudo-code:
```
const b64 = base64.encode(gzip.compress(speedscope_blob))
document.addEventListener("DOMContentLoaded", () => {
    speedscope.loadFileFromBase64('Profile', gzip.decompress(b64));
});
```

odoo/addons/base/models/ir_profile.py
Xavier-Do: I cannot tell without going in depth again in the code on why it was needed but a constraint was to have a computed field  to serve it using web/content, not to create route, and to keep the existing behaviour. We tried to emancipate from some of those constraints but this is the current state in higher stable versions and most of the changes were thought to be "stable friendly"

PR #248473, opened 2026-02-12
[FIX] l10n_latam_check: prevent premature "Paid" status for multiple checks

amahm-odoo:
Steps to reproduce:
1. Create a Vendor Bill for 1000 ARS.
2. Register a payment using "Own Checks".
3. Add two checks (e.g., 500 ARS each).
4. Create and Post the payment.
5. Observe that the Bill ribbon immediately shows "Paid".

Cause:
When processing multiple own checks, Odoo creates a "Split Move" entry to handle individual check maturity dates. To balance the books, it reconciles the original payment liquidity line with a counterpart line in this split entry. Because Odoo's core payment state logic monitors this original liquidity line, it sees the reconciliation as a "clearing" event and incorrectly flips the bill status to "Paid" before the checks have actually cleared the bank.

Solution:
Override _seek_for_lines to redirect Odoo's "status sensor." By pointing the liquidity line target to the individual check lines in the split move instead of the original payment line, the payment state now correctly depends on whether those specific checks have been reconciled with a bank statement. A context guard (skip_check_search) is used during the split move creation to allow the builder to access the original payment line without interference.

opw-5482766

### DIFF
--- a/addons/l10n_latam_check/models/account_payment.py
+++ b/addons/l10n_latam_check/models/account_payment.py
@@ -131,8 +131,9 @@ class AccountPayment(models.Model):
 
     def _l10n_latam_check_split_move(self):
         for payment in self.filtered(lambda x: x.payment_method_code == 'own_checks' and x.payment_type == 'outbound'):
+            pay_with_signal = payment.with_context(skip_check_search=True)
             if len(payment.l10n_latam_new_check_ids) == 1:
-                liquidity_line = payment._seek_for_lines()[0]
+                liquidity_line = pay_with_signal._seek_for_lines()[0]
                 payment.l10n_latam_new_check_ids.outstanding_line_id = liquidity_line.id
                 continue
 
@@ -141,7 +142,7 @@ class AccountPayment(models.Model):
                 'move_type': 'entry',
                 'line_ids': [],
             }
-            payment_liquidity_line = payment._seek_for_lines()[0]
+            payment_liquidity_line = pay_with_signal._seek_for_lines()[0]
 
             # One line per check
             checks_total = sum(payment.l10n_latam_new_check_ids.mapped('amount'))
@@ -199,6 +200,21 @@ class AccountPayment(models.Model):
             check.outstanding_line_id.move_id.button_draft()
             check.outstanding_line_id.move_id.unlink()
 
+    def _seek_for_lines(self):
+        """
+        Overridden to return the check lines when using own checks so the payment status
+        depends on the check status and not on the payment line (which is reconciled against the split move)
+        """
+        # EXTENDS account
+        res = super()._seek_for_lines()
+        if self.env.context.get('skip_check_search'):
+            return res
+
+        if self.payment_method_code == 'own_checks' and self.payment_type == 'outbound' and self.l10n_latam_new_check_ids and len(self.l10n_latam_new_check_ids) > 1:
+            res[0] = self.l10n_latam_new_check_ids.outstanding_line_id
+
+        return res
+
     @api.depends(
         'payment_method_line_id', 'state', 'date', 'amount', 'currency_id', 'company_id',
         'l10n_latam_move_check_ids.issuer_vat', 'l10n_latam_move_check_ids.bank_id', 'l10n_latam_move_check_ids.payment_id.date',
@@ -238,6 +254,11 @@ class AccountPayment(models.Model):
     def _get_trigger_fields_to_synchronize(self):
         res = super()._get_trigger_fields_to_synchronize()
         return res + ('l10n_latam_new_check_ids',)
+
+    @api.depends("l10n_latam_new_check_ids.outstanding_line_id.amount_residual")
+    def _compute_reconciliation_status(self):
+        # EXTENDS account
+        super()._compute_reconciliation_status()
 
     def _prepare_move_line_default_vals(self, write_off_line_vals=None, force_balance=None):
         """ Add check name and operation on liquidity line """
--- a/addons/l10n_latam_check/tests/test_own_checks.py
+++ b/addons/l10n_latam_check/tests/test_own_checks.py
@@ -8,6 +8,21 @@ from odoo import Command, fields
 @tagged('post_install_l10n', 'post_install', '-at_install')
 class TestOwnChecks(L10nLatamCheckTest):
 
+    def _reconcile_check_with_statement(self, check, journal):
+        """Create a bank statement line and reconcile it against the check's outstanding line."""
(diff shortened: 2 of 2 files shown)


malb-odoo (commented):
Technically LGTM but since i don't really know the checks feature maybe @smetl can take a quick look if anything is fishy here 😄

addons/l10n_latam_check/models/account_payment.py
```diff
@@ -199,6 +200,20 @@ def _l10n_latam_check_unlink_split_move(self):
             check.outstanding_line_id.move_id.button_draft()
             check.outstanding_line_id.move_id.unlink()
 
+    def _seek_for_lines(self):
+        """
+        Overridden to return the check lines when using own checks so the payment status
+        depends on the check status and not on the payment line (which is reconciled against the split move)
+        """
+        res = super()._seek_for_lines()
```
AllePilli, line 209:
```suggestion
        """
        # EXTENDS account
        res = super()._seek_for_lines()
```

addons/l10n_latam_check/models/account_payment.py
```diff
@@ -239,6 +254,10 @@ def _get_trigger_fields_to_synchronize(self):
         res = super()._get_trigger_fields_to_synchronize()
         return res + ('l10n_latam_new_check_ids',)
 
+    @api.depends("l10n_latam_new_check_ids.outstanding_line_id.amount_residual")
+    def _compute_reconciliation_status(self):
+        super()._compute_reconciliation_status()
```
AllePilli, line 261:
```suggestion
    def _compute_reconciliation_status(self):
        # EXTENDS account
        super()._compute_reconciliation_status()
```

addons/l10n_latam_check/tests/test_own_checks.py
```diff
@@ -92,3 +92,149 @@ def test_post_own_check_with_3_lines(self):
...
+                        {
+                            "name": "Test Product",
+                            "price_unit": 500.0,
+                            "quantity": 1,
+                        }
+                    )
+                ],
+            }
+        )
+        invoice.action_post()
```
AllePilli: Same as above

addons/l10n_latam_check/tests/test_own_checks.py
```diff
@@ -92,3 +92,149 @@ def test_post_own_check_with_3_lines(self):
...
+
+            with wizard_form.l10n_latam_new_check_ids.new() as check:
+                check.name = "Check002"
+                check.amount = 250.0
+                check.payment_date = fields.Date.today()
+
+        wizard = wizard_form.save()
+        action = wizard.action_create_payments()
+
+        payment = self.env["account.payment"].browse(action["res_id"])
```
AllePilli: same as above

addons/l10n_latam_check/tests/test_own_checks.py
```diff
@@ -92,3 +92,149 @@ def test_post_own_check_with_3_lines(self):
...
+            "l10n_latam_document_type_id": 1,
+            "invoice_line_ids": [
+                Command.create({
+                    "name": "Single Product Test",
+                    "price_unit": 1000.0,
+                    "quantity": 1,
+                })
+            ],
+        })
+        invoice.action_post()
```
AllePilli:
```suggestion
        invoice = self._create_invoice_one_line(
            move_type='in_invoice',
            company_id=self.bank_journal.company_id.id,
            l10n_latam_document_number="00001-00000002",
            l10n_latam_document_type_id=1,
            name="Single Product Test",
            price_unit=1000.0,
            post=True,
        )
```

addons/l10n_latam_check/tests/test_own_checks.py
```diff
@@ -92,3 +92,149 @@ def test_post_own_check_with_3_lines(self):
...
+            wizard_form.payment_method_line_id = own_checks_method
+
+            with wizard_form.l10n_latam_new_check_ids.new() as check:
+                check.name = "CHK-SINGLE-001"
+                check.amount = 1000.0
+                check.payment_date = fields.Date.today()
+
+        wizard = wizard_form.save()
+        action = wizard.action_create_payments()
+        payment = self.env["account.payment"].browse(action["res_id"])
```
AllePilli: Is it possible to use the `AccountTestInvoicingCommon._register_payment()` instead?

addons/l10n_latam_check/tests/test_own_checks.py
```diff
@@ -92,3 +92,149 @@ def test_post_own_check_with_3_lines(self):
...
+                "amount": -check_1.amount,
+                "date": check_1.payment_date,
+                "partner_id": self.partner_a.id,
+            }
+        )
+        _st_liq_1, st_suspense_1, _st_other_1 = st_line_1.with_context(skip_account_move_synchronization=True)._seek_for_lines()
+        st_suspense_1.account_id = check_1.outstanding_line_id.account_id
+        (st_suspense_1 + check_1.outstanding_line_id).reconcile()
+
+        invoice.invalidate_recordset(["payment_state"])
```
AllePilli: Extract this to a separate function (similar to `AccountTestInvoicingCommon.payment_with_statement_line()`) because you're doing the same thing multiple times in both test cases

AllePilli (changes requested):
In accounting we use following convention for hardcoded strings:
- use single quotes only for strings that don't appear in the UI
- use double quotes only for strings that do appear in the UI

ex: `self.env['account.move'].search([('name', '=', "INV001")])`
please adapt your code to follow this rule.

AllePilli (approved):

amahm-odoo:
should we merge this PR ?

addons/l10n_latam_check/models/account_payment.py
```diff
@@ -199,6 +200,21 @@ def _l10n_latam_check_unlink_split_move(self):
             check.outstanding_line_id.move_id.button_draft()
             check.outstanding_line_id.move_id.unlink()
 
+    def _seek_for_lines(self):
```
jco-odoo, line 203: - Is it such a bad thing that the status is different? In any case, people have the state on the check? 
- This is a not an ideal inheritance (but maybe no easy alternative), changing a quite core method.

jco-odoo: @vbe-odoo @jorenvo

amahm-odoo:
@jco-odoo should we merge this PR ? or backport what was implemented here 
https://www.odoo.com/odoo/project.task/5207671

how should we move from here ?

PR #248474, opened 2026-02-12
[FIX] core: handle term count mismatch with context edit_translations

HydrionBurst:
When the number of translated terms differs between the base language and its translation, fall back to the base value instead of attempting to map terms. This avoids potential errors and incorrect translation mappings.

Current behavior before PR:

---


kmagusiak (approved):
LGTM.
Even if I find these delayed translation hard to follow.

kmagusiak (approved):
@robodoo r+

PR #248475, opened 2026-02-12
[IMP] website: rework s_chart options and use sliding panel

romo-odoo:
Since the user will not update table values a lot, the table to edit the values was moved in a sliding panel.

Moreover, to increase readibility when there are multiples columns, the table is now scrollable (with the first column fixed) to increase the size of each inputs. The user also has the possibility to change the table size (rows/columns) with a new option that adds/removes column in batch, instead of adding or removing them one by one.

Finally, to simplify the edition, the user can copy/paste values directly in the cells.

task-5104732

### DIFF
--- a/addons/html_builder/static/src/builder.scss
+++ b/addons/html_builder/static/src/builder.scss
@@ -207,3 +207,10 @@
     overflow-y: scroll;
     scroll-behavior: smooth;
 }
+
+// Increase the size of the icon within a button to improve readability
+// (e.g., "fa-table" is hard to understand when the font-size is too small)
+.larger_button_icon {
+    --btn-padding-y: 0.05rem !important;
+    font-size: 14px;
+}
--- a/addons/website/static/src/builder/plugins/options/chart_option.js
+++ b/addons/website/static/src/builder/plugins/options/chart_option.js
@@ -2,25 +2,14 @@ import { useState } from "@web/owl2/utils";
 import { BaseOptionComponent } from "@html_builder/core/base_option_component";
 import { useDomState } from "@html_builder/core/utils";
 import { registry } from "@web/core/registry";
-import { getCSSVariableValue } from "@html_editor/utils/formatting";
 import { _t } from "@web/core/l10n/translation";
-import { isCSSColor } from "@web/core/utils/colors";
-
-export const DATASET_KEY_PREFIX = "chart_dataset_";
-
-export function getColor(color, win, doc) {
-    if (!color) {
-        return "";
-    }
-    return isCSSColor(color)
-        ? color
-        : getCSSVariableValue(color, win.getComputedStyle(doc.documentElement));
-}
+import { DATASET_KEY_PREFIX, addChartColumn, addChartRow, getColor } from "./chart_option_utils";
+import { getActiveHotkey } from "@web/core/hotkeys/hotkey_service";
 
 export class ChartOption extends BaseOptionComponent {
     static id = "chart_option";
     static template = "website.ChartOption";
-    static dependencies = ["chartOptionPlugin"];
+    static dependencies = ["chartOptionPlugin", "operation", "history"];
 
     setup() {
         super.setup();
@@ -96,6 +85,7 @@ export class ChartOption extends BaseOptionComponent {
     getColor(color) {
         return getColor(color, this.window, this.document);
     }
+
     /**
      * Retrieve the colors already in use in the chart.
      *
@@ -193,6 +183,8 @@ export class ChartOption extends BaseOptionComponent {
         // click on a table inner cell
         else if (datasetIndex !== -1 && datasetIndex !== cellRowEl.children.length - 2) {
             this.updateCurrentCell({ datasetIndex, dataIndex });
+        } else if (datasetIndex === -1) {
+            this.updateCurrentCell({ datasetIndex: null, dataIndex: null });
         }
     }
     /**
@@ -293,6 +285,112 @@ export class ChartOption extends BaseOptionComponent {
                 .querySelector(".o_builder_matrix_remove_row")
                 .classList.remove("visually-hidden-focusable");
         }
+    }
+
+    onCellKeydown(ev) {
+        const hotkey = getActiveHotkey(ev);
+        if (hotkey === "control+z") {
+            this.dependencies.operation.next(() => this.dependencies.history.undo());
+        } else if (hotkey === "control+shift+z" || hotkey === "control+y") {
+            this.dependencies.operation.next(() => this.dependencies.history.redo());
+        }
+    }
+
+    /**
+     * Handles paste events for a table of inputs.
+     *
+     * Reads tab-separated clipboard data (e.g. copied from Excel), then
+     * distributes the values into the table starting from the targeted cell.
+     * Label cells use raw text (same behavior as BuilderTextInput), while
+     * dataset value cells parse numbers.
+     *
+     * @param {ClipboardEvent} ev
+     */
+    onPasteValues(ev) {
+        ev.preventDefault();
+
+        const clipText = ev.clipboardData.getData("text/plain");
+        if (!clipText) {
+            return;
+        }
+        const clipRows = clipText
+            .replace(/\r/g, "")
+            .split("\n")
+            .map((row) => row.split("\t"));
+        if (!clipRows.length) {
+            return;
+        }
+
+        const { cellEl, cellSectionEl, datasetIndex, dataIndex } = this.getCellInfo(ev);
+        if (!cellEl) {
+            return;
+        }
+
+        const isDatasetLabelCell = cellSectionEl.tagName === "THEAD" && datasetIndex >= 0;
+
+        const startRow = isDatasetLabelCell ? 0 : dataIndex + 1;
+        const startCol = datasetIndex + 1;
+
+        const maxPastedColumnCount = Math.max(...clipRows.map((row) => row.length));
+        if (!maxPastedColumnCount) {
+            return;
+        }
+
+        this.dependencies.operation.next(async () => {
+            const editingElement = this.env.getEditingElement();
+            const data = this.prepareData(editingElement);
+            const isPieChart = this.isPieChart(editingElement);
+
+            const maxTargetRow = startRow + clipRows.length - 1;
+            const maxTargetCol = startCol + maxPastedColumnCount - 1;
+            const rowsToAdd = Math.max(maxTargetRow - data.labels.length, 0);
+            const colsToAdd = Math.max(maxTargetCol - data.datasets.length, 0);
+            const timestamp = Date.now();
+
+            for (let i = 0; i < rowsToAdd; i++) {
+                addChartRow(data, isPieChart);
+            }
+            for (let i = 0; i < colsToAdd; i++) {
+                addChartColumn(data, isPieChart, `${DATASET_KEY_PREFIX}${timestamp}${i}`);
+            }
+
+            let hasChanged = !!(rowsToAdd || colsToAdd);
+            for (let i = 0; i < clipRows.length; i++) {
+                for (let j = 0; j < clipRows[i].length; j++) {
+                    const targetRow = startRow + i;
+                    const targetCol = startCol + j;
+                    const pastedValue = clipRows[i][j];
+
+                    if (targetRow === 0 && targetCol > 0) {
+                        const currentValue = data.datasets[targetCol - 1].label || "";
+                        if (currentValue !== pastedValue) {
+                            data.datasets[targetCol - 1].label = pastedValue;
+                            hasChanged = true;
+                        }
+                    } else if (targetCol === 0 && targetRow > 0) {
+                        const currentValue = data.labels[targetRow - 1] || "";
+                        if (currentValue !== pastedValue) {
+                            data.labels[targetRow - 1] = pastedValue;
+                            hasChanged = true;
+                        }
+                    } else if (targetRow > 0 && targetCol > 0) {
+                        const parsedValue = Number.parseFloat(pastedValue);
+                        const value = Number.isNaN(parsedValue) ? 0 : parsedValue;
+                        const currentValue = data.datasets[targetCol - 1].data[targetRow - 1];
+                        if (currentValue !== value) {
+                            data.datasets[targetCol - 1].data[targetRow - 1] = value;
+                            hasChanged = true;
+                        }
+                    }
+                }
+            }
+
+            if (!hasChanged) {
+                return;
+            }
+            editingElement.dataset.data = JSON.stringify(data);
+            this.dependencies.history.addStep({ batchable: true });
+        });
     }
 }
 
--- a/addons/website/static/src/builder/plugins/options/chart_option.scss
+++ b/addons/website/static/src/builder/plugins/options/chart_option.scss
@@ -1,36 +1,71 @@
-table.o_builder_matrix {
-    table-layout: fixed;
-    max-width: 100%;
+.o_builder_matrix_wrapper {
+    overflow-x: auto;
+    scrollbar-width: thin;
+    max-height: 500px;
 
-    td, th {
-        text-align: center;
+    table.o_builder_matrix {
+        table-layout: fixed;
+        max-width: 100%;
 
-        > :is(div, button) {
-            width: 100% !important;
+        // First row / First column
+        .o_builder_matrix_sticky_left, .o_builder_matrix_sticky_top {
+            position: sticky;
+            background-color: $o-we-bg-lighter;
+
+            & > div {
+                background-color: $o-we-bg-dark;
+            }
         }
 
-        > div {
-            padding: 0;
-            border-radius: 0;
+        // First row
+        .o_builder_matrix_sticky_top {
+            top: 0;
         }
 
-
-        .btn {
-            padding: unset;
+        // First column
+        .o_builder_matrix_sticky_left {
+            min-width: 80px !important;
+            left: 0;
         }
 
-        .o-hb-input-number, .o-hb-input-base {
-            margin-right: 0;
-            padding: 1px 3px;
-            border-radius: 0;
+        // Top-Left corner
+        .o_builder_matrix_sticky_left.o_builder_matrix_sticky_top {
+            z-index: 1;
         }
 
-        &:last-child {
-            width: 28px;
-        }
-    }
+        td, th {
+            text-align: center;
 
-    .visually-hidden-focusable:not(:focus, :focus-within) {
-        position: static !important;
+            > :is(div, button) {
+                width: 100% !important;
+            }
+
+            > div {
+                padding: 0;
+                border-radius: 0;
+            }
+
+            .btn {
+                padding: unset;
+            }
+
+            .o-hb-input-number, .o-hb-input-base {
+                margin-right: 0;
+                padding: 1px 3px;
+                border-radius: 0;
+            }
+
+            &:not(:last-child) {
+                min-width: 55px;
+            }
+
+            &:last-child {
+                min-width: 28px;
+            }
+        }
+
+        .visually-hidden-focusable:not(:focus, :focus-within) {
+            position: static !important;
+        }
     }
 }
--- a/addons/website/static/src/builder/plugins/options/chart_option.xml
+++ b/addons/website/static/src/builder/plugins/options/chart_option.xml
@@ -14,144 +14,185 @@
             <BuilderSelectItem actionValue="'radar'">Radar</BuilderSelectItem>
         </BuilderSelect>
     </BuilderRow>
-    <BuilderRow t-if="this.domState.data.datasets.length > 1" label.translate="Stacked">
-        <BuilderCheckbox t-if="this.isActiveItem('bar_chart_opt') or this.isActiveItem('horizontal_bar_chart_opt') or this.isActiveItem('line_chart_opt')"
-            dataAttributeAction="'stacked'"
-            dataAttributeActionValue="'true'"/>
+    <BuilderRow label.translate="Stacked" level="1" t-if="this.domState.data.datasets.length > 1 and (this.isActiveItem('bar_chart_opt') or this.isActiveItem('horizontal_bar_chart_opt') or this.isActiveItem('line_chart_opt'))">
+        <BuilderCheckbox dataAttributeAction="'stacked'" dataAttributeActionValue="'true'"/>
     </BuilderRow>
     <BuilderRow label.translate="Legend">
-        <BuilderSelect dataAttributeAction="'legendPosition'">
-            <BuilderSelectItem dataAttributeActionValue="'none'">None</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'top'">Top</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'left'">Left</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'bottom'">Bottom</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'right'">Right</BuilderSelectItem>
-        </BuilderSelect>
+        <BuilderButtonGroup dataAttributeAction="'legendPosition'">
+            <t t-set="leftLegendButton">
+                <BuilderButton dataAttributeActionValue="'left'" title.translate="Left" iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"/>
+            </t>
+            <t t-set="rightLegendButton">
+                <BuilderButton dataAttributeActionValue="'right'" title.translate="Right" iconImg="'/website/static/src/img/snippets_options/pos_right.svg'"/>
+            </t>
+
+            <!-- In RTL/LTR, swap only the button display order without using
+                 ltrRtlMapping, since Chart.js does not mirror legend sides
+                 based on text direction. The values for "left" and "right"
+                 should not be swapped. -->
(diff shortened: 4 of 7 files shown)


addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -104,6 +106,42 @@ export class SetChartTypeAction extends BaseChartAction {
         this.updateDOMData(editingElement, data);
     }
 }
+export class SetColumnAction extends BaseChartAction {
+    static id = "setColumn";
+    apply({ editingElement, value }) {
+        const data = this.getData(editingElement);
+        const diff = parseInt(value) - data.datasets.length;
+
+        if (diff == 0) {
```
robinlej: Always `===`

or `if (!diff)`

addons/website/static/src/builder/plugins/options/chart_option.scss
```diff
@@ -1,36 +1,57 @@
...
+            position: sticky;
+            left: 0;
+            background-color: $o-we-bg-lighter;
         }
 
-        > div {
-            padding: 0;
-            border-radius: 0;
+        thead > tr > th:not(:last-child) > div,
+        tr:not(:last-child) > th:first-child > div {
```
robinlej: Maybe we could use an actual class instead of relying on such a chain of html tags 😬

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
-                        <BuilderTextInput
-                            action="'updateDatasetLabel'"
-                            actionParam="dataset.key"
-                            style="this.domState.isPieChart ? '' : ('border: 2px solid ' + (this.getColor(dataset.backgroundColor) || this.getColor(dataset.borderColor)))"/>
-                    </th>
-                </t>
+    <BuilderRow label.translate="Values">
+        <BuilderSlidingPanel
+            label.translate="Values"
+            extraClasses="'btn-primary'"
```
robinlej: 1. I know that's in the specs, but it's weird to have a purple button: since the sidebar design revamp, all the primary buttons are neon green
Shouldn't it be the case here too? So `btn-success` instead of `btn-primary`

2. Wouldn't it make sense to put the option either before or below the chart `Type` option? "Values" seems quite important to me

3. Wouldn't "Chart data" be a better (clearer) label than "Values"?

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
+            fullHeight="false"
+            darkBackground="false">
+            <i class="fa fa-paint-brush"/>
```
robinlej: Nitpick: the `paint-brush` looks a bit weird to me given the goal of the option. Why no `fa-pencil`?

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
+            <t t-set-slot="content" t-slot-scope="panel">
+                <BuilderRow label.translate="Row x Column">
```
robinlej: I'd use the × symbol => `Rows × Columns`

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
                 .classList.remove("visually-hidden-focusable");
         }
     }
+
+    onPasteValues(ev) {
```
robinlej, line 309: The paste feature is very neat and useful :)

- Could we do the same for the labels?

- And could we have a way to copy/paste a complete table with labels and data? (the issue is we don't have the "0-cell" on the top left, so nowhere to paste the full table) Maybe in a follow-up PR?

  | One | Two | Three | Four
-- | -- | -- | -- | --
1st | 10 | 20 | 30 | 50
2nd | 5 | 8 | 1 | 2
3rd | 42 | 18 | 24 | 31

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -193,6 +193,8 @@ export class ChartOption extends BaseOptionComponent {
         // click on a table inner cell
         else if (datasetIndex !== -1 && datasetIndex !== cellRowEl.children.length - 2) {
             this.updateCurrentCell({ datasetIndex, dataIndex });
+        } else if (datasetIndex === -1) {
+            this.updateCurrentCell({ datasetIndex: null, dataIndex: null });
```
robinlej, line 187: Why do you need it?

addons/website/static/src/builder/plugins/options/chart_option.js
robinlej: Nitpick: it's pretty obvious, but add some JS doc anyway to mention that `@param {ClipboardEvent} ev`

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
                 .classList.remove("visually-hidden-focusable");
         }
     }
+
+    onPasteValues(ev) {
+        ev.preventDefault();
+
+        const clipText = (ev.clipboardData || window.clipboardData).getData("text");
```
robinlej: > `ev.clipboardData || window.clipboardData`

When is `ev.clipboardData` undefined?

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
...
+        const clipRows = clipText.split("\n");
+        for (let i = 0; i < clipRows.length; i++) {
+            clipRows[i] = clipRows[i].split("\t");
+        }
```
robinlej, line 337: Shouldn't we create new rows/columns if we detect that there is more data than available cells?

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
...
+
+        const nbCols = tdEls.length - 1;
+        const nbRows = trEls.length - 1;
+
+        for (let i = 0; i < clipRows.length && i + rowIndex < nbRows; i++) {
+            for (let j = 0; j < clipRows[i].length && j + colIndex < nbCols; j++) {
+                const value = parseInt(clipRows[i][j]);
+                const inputEl = tdEl.querySelector("input");
+                inputEl.value = value;
+                const changeEv = new Event("change", { value: value });
```
robinlej: nitpick: `{ value: value }` can be written `{ value }`

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
...
+                inputEl.dispatchEvent(changeEv);
+                tdEl = tdEl.nextElementSibling;
+            }
+            trEl = trEl.nextElementSibling;
+            tdEl = [...trEl.querySelectorAll(":scope > td")].at(colIndex);
```
robinlej: You don't really need to keep updating `trEl` and `tdEl`.

If you store a matrix of the cells:
`const cells = trEls.map((trEl) => trEl.querySelectorAll(":scope > td"));`

you could just have a `currRowIndex = i + rowIndex` and `currColIndex = j + colIndex`
and query `cells[currRowIndex][currColIndex]` instead of `trEl`

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
+                    <BuilderNumberInput min="1" title.translate="row" action="'setRow'" preview="false"/>
+                    <BuilderNumberInput min="1" title.translate="column" action="'setColumn'" preview="false"/>
```
robinlej: Nitpick: `title.translate="Rows"` / `Columns` with capital letters (and I'd add the plural too)

addons/website/static/src/builder/plugins/options/chart_option.xml
robinlej: As the button only contains an icon, it should have accessible text with `<span class="visually-hidden">Some label</span>`

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
+                        getUsedCustomColors.bind="this.getColorPalette"
+                        action="'colorChange'"
+                        actionParam="{
+                            type: 'backgroundColor',
+                            datasetIndex: this.state.currentCell.datasetIndex,
+                            dataIndex: this.state.currentCell.dataIndex,
+                            backgroundColor: this.state.currentCell.backgroundColor,
+                        }"
+                        enabledTabs="['solid', 'custom']"
             ></BuilderColorPicker>
```
robinlej: this too should be indented to the right (but you can even just remove it and auto-close the `BuilderColorPicker`)

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
+                        getUsedCustomColors.bind="this.getColorPalette"
+                        action="'colorChange'"
+                        actionParam="{
+                            type: 'borderColor',
+                            datasetIndex: this.state.currentCell.datasetIndex,
+                            dataIndex: this.state.currentCell.dataIndex,
+                            borderColor: this.state.currentCell.borderColor,
+                        }"
+                        enabledTabs="['solid', 'custom']"
             ></BuilderColorPicker>
```
robinlej: same

addons/website/static/src/builder/plugins/options/chart_option.xml
robinlej: Actually I'd rename `setRows` and `setColumns` too

addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -104,6 +106,42 @@ export class SetChartTypeAction extends BaseChartAction {
...
+            const infoToAdd = [];
+            for (let i = 0; i < diff; i++) {
+                const newDataset = {
+                    key: DATASET_KEY_PREFIX + Date.now() + i,
+                    label: "",
+                    data: fillDatasetArray(0),
+                    backgroundColor: this.isPieChart(editingElement)
+                        ? data.labels.map(() => this.randomColor())
+                        : this.randomColor(),
+                    borderColor: this.isPieChart(editingElement) ? fillDatasetArray("") : "",
```
robinlej: make a reusable method in `BaseChartAction` to call that here and in `addColumn`
(you could even take the code before the loop and call the method with `diff === 0` in `addColumn`)

addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -132,6 +170,45 @@ export class RemoveColumnAction extends BaseChartAction {
         this.updateDOMData(editingElement, data);
     }
 }
+export class SetRowAction extends BaseChartAction {
+    static id = "setRow";
+    apply({ editingElement, value }) {
+        const data = this.getData(editingElement);
+        const diff = parseInt(value) - data.labels.length;
+
+        if (diff == 0) {
```
robinlej: same

addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -132,6 +170,45 @@ export class RemoveColumnAction extends BaseChartAction {
...
+                data.labels.push("");
+                data.datasets.forEach((dataset) => {
+                    dataset.data.push(0);
+                    if (this.isPieChart(editingElement)) {
+                        dataset.backgroundColor.push(this.randomColor());
+                        dataset.borderColor.push("");
+                    }
+                });
+            }
+            this.updateDOMData(editingElement, data);
```
robinlej: same kind of comment here for the rows as for the columns

addons/website/static/src/snippets/s_chart/chart.edit.js
```diff
@@ -9,9 +9,13 @@ const ChartEdit = (I) =>
         }
 
         start() {
-            super.start();
-            this.websiteEditService = this.services.website_edit;
-            this.websiteEditService.callShared("builderOverlay", "refreshOverlays");
+            // Prevent multiple start when multiple values of the dataset are
+            // changed quickly (on paste)
+            this.waitForAnimationFrame(() => {
```
robinlej: hmm, not sure about this one. 🤔  

Shouldn't it be done as a generic "batched start" at the `edit_interaction` level, during normalize? Or in the `website_edit_service`, on `refresh`?

Also, do you really need an animation frame, or just a tick would be enough?

addons/website/static/tests/builder/website_builder/chart_option.test.js
```diff
@@ -344,20 +366,89 @@ describe("Add & Delete buttons", () => {
...
+    expect(data.datasets[0].data).toHaveLength(10);
+    expect(".hb-sliding-panel table tbody tr").toHaveCount(11);
+    await contains(".hb-sliding-panel [data-action-id=setRow] input").edit(1);
+    data = JSON.parse(queryFirst(":iframe .s_chart").dataset.data);
+    expect(data.labels).toHaveLength(1);
+    expect(data.datasets[0].data).toHaveLength(1);
+    expect(".hb-sliding-panel table tbody tr").toHaveCount(2);
+});
+
+test("Pasting values update multiple chart values at once", async () => {
```
robinlej: nitpick: updates

addons/html_builder/static/src/core/building_blocks/builder_sliding_panel.scss
```diff
@@ -55,6 +55,8 @@
         position: absolute;
         animation-name: o_hb_sliding_panel_anim_hide;
     }
+
+    @include o-input-number-no-arrows();
```
robinlej: Note when you rebase: it's been added by another commit in the meantime: https://github.com/odoo/odoo/commit/f1eecc76ebaca6bf4c4fba0ee2ca6fb573b7f57b

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,163 @@
...
-            <BuilderSelectItem dataAttributeActionValue="'top'">Top</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'left'">Left</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'bottom'">Bottom</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'right'">Right</BuilderSelectItem>
-        </BuilderSelect>
+        <BuilderButtonGroup dataAttributeAction="'legendPosition'">
+            <BuilderButton dataAttributeActionValue="'none'" title.translate="Hide" icon="'fa-eye-slash'"/>
+            <BuilderButton dataAttributeActionValue="'top'" title.translate="Top" iconImg="'/website/static/src/img/snippets_options/pos_top.svg'"/>
+            <BuilderButton dataAttributeActionValue="'left'" title.translate="Left" iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"/>
+            <BuilderButton dataAttributeActionValue="'right'" title.translate="Right" iconImg="'/website/static/src/img/snippets_options/pos_right.svg'"/>
```
robinlej: You'll need some `ltrRtlMapping` for left and right, I think.
But because the value is applied on the chart canvas, it might be one of the few cases that is a bit finicky, though 🤔  To check with an actual RTL website (test with backend and frontend in different directions)

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +296,40 @@ export class ChartOption extends BaseOptionComponent {
...
         }
     }
+
+    onPasteValues(ev) {
+        ev.preventDefault();
+
+        const clipText = (ev.clipboardData || window.clipboardData).getData("text");
+        const clipRows = clipText.split("\n");
+        for (let i = 0; i < clipRows.length; i++) {
+            clipRows[i] = clipRows[i].split("\t");
```
robinlej: Can you double check that it works for RTL languages? 🤔

addons/website/static/src/builder/plugins/options/chart_option.xml
robinlej: Actually, let me tag @lebl-odoo, those are questions for you :)

lebl-odoo: Hello @robinlej ,

1. Good question and actually you may have right. We choose purple because it's a slidepanel and up to now any slide panel used purple button, but here it's a bit different because it's not just style, it's about inserting data. Green should be good :) We could also use fa-table icon

2. We had the discussion during the development IIRC, it has been done on purpose : Type have an impact on Data chart

3. At that time, we hesitate between "Values" and "Data", but data is too obscure. Adding "Chart" could be a nice improvement.

cc @romo-odoo

addons/website/static/src/builder/plugins/options/chart_option.js
romo-odoo: When the user click on the first column (labels), the colorpickers shouldn't be displayed. The "current cell" does not correspond to any value of the chart.

addons/website/static/src/builder/plugins/options/chart_option.xml
romo-odoo: I tested in Arabic and it seems correct
Image

romo-odoo:
@robinlej 
I updated the code and how `onPasteValues` works. Instead of writing the values in `BuilderNumberInput`, I change directly the dataset of the chart. That way, I can easily add rows / colums and the code is faster. It also solve the issue with the Chart interaction restarting multiple times.
Everything is in an history step and there no need anymore for `batchable` in `BuilderNumberInput`.

addons/website/static/src/builder/plugins/options/chart_option.scss
```diff
@@ -1,36 +1,56 @@
...
-    table-layout: fixed;
-    max-width: 100%;
-
-    td, th {
-        text-align: center;
-
-        > :is(div, button) {
-            width: 100% !important;
+.o_builder_matrix_wrapper {
+    overflow-x: scroll;
```
robinlej: Why not `auto`?

addons/website/static/src/builder/plugins/options/chart_option.scss
```diff
@@ -1,36 +1,56 @@
...
+    scrollbar-width: thin;
+
+    table.o_builder_matrix {
+        table-layout: fixed;
+        max-width: 100%;
+
+        tr > th:first-child,
+        tr > td:first-child {
+            min-width: 80px;
+            position: sticky;
```
robinlej, line 12: Maybe we should set a max-height on the table wrapper, and use `position: sticky` on the first row cells too?
I think it would be slightly more useable with lots of data

addons/website/static/src/builder/plugins/options/chart_option.xml
robinlej: As discussed: the buttons are pointing to the wrong direction in Arabic, that's why you need `ltrRtlMapping`

addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -104,61 +128,88 @@ export class SetChartTypeAction extends BaseChartAction {
...
+        if (!diff) {
+            return;
+        }
+        if (diff > 0) {
+            const isPieChart = this.isPieChart(editingElement);
+            for (let i = 0; i < diff; i++) {
+                addChartColumn(data, isPieChart, DATASET_KEY_PREFIX + Date.now() + i);
+            }
+        } else {
+            for (let i = 0; i < -diff; i++) {
```
robinlej: `Math.abs(diff)`

addons/website/static/src/builder/plugins/options/chart_option_plugin.js
```diff
@@ -104,61 +128,88 @@ export class SetChartTypeAction extends BaseChartAction {
...
+        if (!diff) {
+            return;
+        }
+        if (diff > 0) {
+            const isPieChart = this.isPieChart(editingElement);
+            for (let i = 0; i < diff; i++) {
+                addChartRow(data, isPieChart);
+            }
+        } else {
+            for (let i = 0; i < -diff; i++) {
```
robinlej: same

addons/website/static/src/builder/plugins/options/chart_option.scss
robinlej: cc @lebl-odoo for your opinion even if it's nothing big

lebl-odoo: As we discussed, let's move with what you proposed :)

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +286,112 @@ export class ChartOption extends BaseOptionComponent {
...
+
+        const startRow = isDatasetLabelCell ? 0 : dataIndex + 1;
+        const startCol = datasetIndex + 1;
+
+        const maxPastedColumnCount = Math.max(...clipRows.map((row) => row.length));
+        if (!maxPastedColumnCount) {
+            return;
+        }
+
+        this.env.editor.shared.operation.next(async () => {
```
robinlej: Since you have it as a dependency, use `this.dependencies.operation` instead

addons/website/static/src/builder/plugins/options/chart_option.js
```diff
@@ -294,6 +286,112 @@ export class ChartOption extends BaseOptionComponent {
...
+                        }
+                    }
+                }
+            }
+
+            if (!hasChanged) {
+                return;
+            }
+            editingElement.dataset.data = JSON.stringify(data);
+            this.env.editor.shared.history.addStep({ batchable: true });
```
robinlej: Since you have it as a dependency, use `this.dependencies.history` instead

addons/website/static/src/builder/plugins/options/chart_option.xml
```diff
@@ -14,144 +14,168 @@
...
-            <BuilderSelectItem dataAttributeActionValue="'top'">Top</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'left'">Left</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'bottom'">Bottom</BuilderSelectItem>
-            <BuilderSelectItem dataAttributeActionValue="'right'">Right</BuilderSelectItem>
-        </BuilderSelect>
+        <BuilderButtonGroup dataAttributeAction="'legendPosition'">
+            <BuilderButton dataAttributeActionValue="'none'" title.translate="Hide" icon="'fa-eye-slash'"/>
+            <BuilderButton dataAttributeActionValue="'top'" title.translate="Top" iconImg="'/website/static/src/img/snippets_options/pos_top.svg'"/>
+            <BuilderButton dataAttributeActionValue="'left'" ltrRtlMapping="'left-right'" title.translate="Left" iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"/>
+            <BuilderButton dataAttributeActionValue="'right'" ltrRtlMapping="'left-right'" title.translate="Right" iconImg="'/website/static/src/img/snippets_options/pos_right.svg'"/>
```
robinlej: That was my intuition: as it's applied on the Chart.js canvas, it's annoying. When the frontend is in RTL (Arabic), it doesn't work as expected.

Here Arabic/Arabic:
Image

Here English/Arabic:
Image

addons/html_builder/static/src/builder.scss
```diff
@@ -207,3 +207,12 @@
...
     scroll-behavior: smooth;
 }
+
+// Increase the size of the icon within a button to improve readability
+// (e.g., "fa-table" is hard to understand when the font-size is too small)
+.larger_button_icon {
+    --btn-padding-y: 0.05rem !important;
+    > i {
+        font-size: 14px;
+    }
```
robinlej: The font-size could be set on the class (the button) itself, the icon will get it by inheritance

addons/website/static/src/builder/plugins/options/chart_option.xml
robinlej: After checking the possibilities, I found that this is a case where `ltrRtlMapping` is not suited. `left` and `right` are applied as dataset values, to be passed to Chart.js. The mapping inverts the actions when the content is RTL, but the chart isn't aware that the content is in RTL, so it doesn't work.

Instead, we can just reorder the buttons:
```xml
<t t-set="leftLegendButton">
    <BuilderButton dataAttributeActionValue="'left'" title.translate="Left" iconImg="'/website/static/src/img/snippets_options/pos_left.svg'"/>
</t>
<t t-set="rightLegendButton">
    <BuilderButton dataAttributeActionValue="'right'" title.translate="Right" iconImg="'/website/static/src/img/snippets_options/pos_right.svg'"/>
</t>
<t t-if="this.env.langDir.builder === 'ltr'">
    <t t-out="leftLegendButton"/>
    <t t-out="rightLegendButton"/>
</t>
<t t-else="">
    <t t-out="rightLegendButton"/>
    <t t-out="leftLegendButton"/>
</t>
```

(In a hypothetical future, we may want to have an easier way to do that. It could go through the ltrRtlMapping with some new prop. But I'm not sure there is any other similar case in the codebase right now.)

robinlej:
@robodoo r+

PR #248476, opened 2026-02-12
[FIX] pos_*: correctly wait for order sync in tour

stva-odoo:
point_of_sale, pos_restaurant
This fix ensures that the tour waits for the order to fully sync before proceeding to the next step when clickPlanButton is triggered. This prevents the local order from being overwritten by the server version when the table is reopened.

runbot-23878,  runbot-238529, runbot-238892


davidmonnom (approved):
@robodoo r/

PR #248477, opened 2026-02-12
[FIX] point_of_sale: ensure amounts are properly rounded

pedrambiria:
Before this commit, the total amount of an order could be not properly rounded in some cases, and would get stored in the database with more decimal places than allowed by the currency.

opw-5919317

---


davidmonnom (approved):
@robodoo up to saas-18.4

davidmonnom (approved):
@robodoo r+

PR #248479, opened 2026-02-12
[FIX] l10n_latam_base: changes identification type on country change

legi-odoo:
**PROBLEM**
PR: https://github.com/odoo/odoo/pull/179078
Removed _onchange_country_id() which was used to set the identification type according to the country of the partner.
This PR reintroduce it, so id type and country remains consistent.

**STEP TO REPRODUCE**
1. install l10n_ar and l10n_co.
2. create a new partner.
3. set its country to Argentina, and select an argentinian id type.
4. set the country to Colombia and save. You end up with a partner from Colombia, with a id type that is used for Argentinian partners which shouldn't be possible.

opw-5801824


kitan191 (changes requested):
Thanks for the fix 👍 

However, the method was removed in saas-18.3.
Could you retarget to that version?

Thanks.

PR #248481, opened 2026-02-12
[FIX] pos_loyalty: prevent hiding a reward product

pedrambiria:
Before this commit, a product used as a loyalty reward could be hidden in the PoS.

opw-5918550

---

### DIFF
--- a/addons/pos_loyalty/models/product_product.py
+++ b/addons/pos_loyalty/models/product_product.py
@@ -36,8 +36,6 @@ class ProductProduct(models.Model):
             products = products.read(fields=res['fields'], load=False)
             self._process_pos_ui_product_product(products, config_id)
 
-            data['pos.session']['data'][0]['_pos_special_products_ids'] += [product.id for product in reward_products if product.id not in [p["id"] for p in res['data']]]
-
             # Identify special loyalty products (e.g., gift cards, e-wallets) to be displayed in the POS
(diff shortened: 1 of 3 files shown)


addons/pos_loyalty/models/product_product.py
```diff
@@ -36,7 +36,9 @@ def _load_pos_data(self, data):
             products = products.read(fields=res['fields'], load=False)
             self._process_pos_ui_product_product(products, config_id)
 
-            data['pos.session']['data'][0]['_pos_special_products_ids'] += [product.id for product in reward_products if product.id not in [p["id"] for p in res['data']]]
+            product_ids_to_hide = reward_products - self.env['product.product'].browse(classic_product_ids)
+            product_ids_to_hide = product_ids_to_hide - product_ids_to_hide.filtered_domain(self._load_pos_data_domain(data))
+            data['pos.session']['data'][0]['_pos_special_products_ids'] += product_ids_to_hide.ids
```
davidmonnom: I'm thinking, do you know if its really necessary to hide any products ?

pedrambiria: You are right. If a product is not marked as Available in POS or does not belong to the Limited Categories defined in your settings, it will be excluded from the ui. I modified the code.

davidmonnom (approved):
@robodoo r+

PR #248482, opened 2026-02-12
[FIX] hr_holidays: count valid allocations starting previous year

maedekey:
__

When setting the start date for a time off allocation to the previous year, it is not taken into account when computing the count of employee allocations on the time off type page.

## Reproduction Steps
1. Go to Time off > Configuration > Time off Types and click on any time off type.
2. A smart button Allocations should appear with a number in it. Note the number and click on the button.
3. If no allocation exists yet, create one. Otherwise, click on an already existing allocation.
4. Set the start date of the validity period to any date last year. Set the ending date so that the allocation is still valid as of now.
5. Go back to the Time off type page and look at the number on the Allocations smart button.

### Expected behavior
As the allocation we set is still valid, the number shouldn't have changed.

### Unexpected behavior
The allocation number has been decreased. However, when we click on the smart button, the same number of valid allocations will show. This creates an inconsistency between the smart button and the allocation page, as the smart button should show the number of valid allocations, and when landing on the allocation page, the results are automatically filtered by validity.

## Origin of the issue
The domain of the allocations to take into account when computing the count of valid allocations is defined here:
https://github.com/odoo/odoo/blob/2264f330859b79010b227e3a9fda1075de8ed4e8/addons/hr_holidays/models/hr_leave_type.py#L297-L304 This doesn't take into account valid allocations that started during the previous year.
The inconsistency with the allocation page can be seen here: https://github.com/odoo/odoo/blob/2264f330859b79010b227e3a9fda1075de8ed4e8/addons/hr_holidays/views/hr_leave_allocation_views.xml#L40-L46 Where the filter is defined based on today, rather than on the whole year, unlike above.
__
opw-5504272

### DIFF
--- a/addons/hr_holidays/models/hr_leave_type.py
+++ b/addons/hr_holidays/models/hr_leave_type.py
@@ -294,12 +294,11 @@ class HolidaysType(models.Model):
             holiday_status.virtual_remaining_leaves = leave_type_tuple[1].get('virtual_remaining_leaves', 0)
 
     def _compute_allocation_count(self):
-        min_datetime = fields.Datetime.to_string(datetime.now().replace(month=1, day=1, hour=0, minute=0, second=0, microsecond=0))
-        max_datetime = fields.Datetime.to_string(datetime.now().replace(month=12, day=31, hour=23, minute=59, second=59))
+        today = fields.Date.to_string(date.today())
         domain = [
             ('holiday_status_id', 'in', self.ids),
-            ('date_from', '>=', min_datetime),
-            ('date_from', '<=', max_datetime),
+            ('date_from', '<=', today),
+            '|', ('date_to', '=', False), ('date_to', '>=', today),
             ('state', 'in', ('confirm', 'validate')),
         ]
 
--- a/addons/hr_holidays/tests/test_allocations.py
+++ b/addons/hr_holidays/tests/test_allocations.py
@@ -570,3 +570,16 @@ class TestAllocations(TestHrHolidaysCommon):
             allocation_form.number_of_hours_display = 7.2
             allocation = allocation_form.save()
             self.assertEqual(allocation.duration_display, '7.2 hours')
+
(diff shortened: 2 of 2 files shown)


addons/hr_holidays/models/hr_leave_type.py
```diff
@@ -294,12 +294,10 @@ def _compute_leaves(self):
...
 
     def _compute_allocation_count(self):
-        min_datetime = fields.Datetime.to_string(datetime.now().replace(month=1, day=1, hour=0, minute=0, second=0, microsecond=0))
-        max_datetime = fields.Datetime.to_string(datetime.now().replace(month=12, day=31, hour=23, minute=59, second=59))
         domain = [
             ('holiday_status_id', 'in', self.ids),
-            ('date_from', '>=', min_datetime),
-            ('date_from', '<=', max_datetime),
+            ('date_from', '<=', fields.Datetime.to_string(datetime.now())),
+            '|', ('date_to', '=', False), ('date_to', '>=', fields.Datetime.to_string(datetime.now())),
```
yoba-odoo:
```suggestion
            ('date_from', '<=', fields.Date.to_string(date.today())),
            '|', ('date_to', '=', False), ('date_to', '>=', fields.Date.to_string(date.today())),
```



Since this commit https://github.com/odoo-dev/odoo/commit/254d6a71e8d5e7bdf8b302d774e024646921b72d the date_from and date_to of the allocation is no longer of type datetime so we can only use date.

Also since we are using the same date we can store it in a variable and then using it as much as we can rather than calling fields.Date multiple times

yoba-odoo (commented):
Thanks for your work!!

Left one comment otherwise LGTM!!

Bertrand2 (approved):
Thanks for the fix! I tested wrongly and indeed the bug was still there in 18.0 😄 
robodoo r+

PR #241459, opened 2025-12-26
[IMP] web: Inner filter support in search view

PiaCOS:
In some search views the list of filters can be huge which impairs the
readability.

This commit adds support for inner filters in `<search/>` allowing them
to be collapsed into a parent filter, reducing the number of visible
entries in the search view.

The logic for date filters was partially merged with the one for inner
filters since the syntax is the same.

It can be used inside a search arch as such:
```xml
<search>
    <filter string="Priority">
        <filter string="Urgent" domain="[('priority', '=', 3)]" name="urgent_priority"/>
        <filter string="High" domain="[('priority', '=', 2)]" name="high_priority"/>
        <filter string="Medium " domain="[('priority', '=', 1)]" name="medium_priority"/>
        <filter string="Low" domain="[('priority', '=', 0)]" name="low_priority"/>
    </filter>
</search>
```

task-5232012

PR #248489, opened 2026-02-12
[IMP] web: support for multi-select inner filter

PiaCOS:
Previously, the parents filter could only have one inner filter active at the same time.
This commits adds support for multi-select inner filters by combining their domain with an OR.

see #241459 

task-5933887


aab-odoo (approved):
robodoo r+

PR #248490, opened 2026-02-12
[FIX] delivery_stock_picking_batch: auto batch pickings with carrier set after confirmation to correct batch

HazemWaheed2050:
**Problem:**
If you have group by carrier option in automatic batches, when you create an SO with a stocked product, confirm the order, and in the delivery step you set/change a carrier, it creates a new batch transfer, and it doesn't group it with the appropriate batch with that carrier.

**Steps to reproduce:**
1) In the settings enable: Batch, Wave & Cluster transfers
2) Inventory > Configuration > Warehouse Management > Operation types
3) Enable Auto-batches, Batch grouping by carrier on Delivery orders
4) Make sure there is no batches made before contains the carrier you're going to use (to cleanly test this issue)
5) Make a sales order, add storable product, press on 'Add shipping' button and add 'standard delivery'
6) Go to the delivery of the pick from the smart button > validate it
7) Go to the next transfer to see the batch number
8) Make a new sales order, add storable product > confirm it > go to the delivery step > additional info
9) Add 'standard delivery' in the carrier field > validate the delivery > go to the next transfer button
10) check the batch it got added to, you will find it is a new one and didn't get grouped to the last batch created with the same carrier.

**Expected behaviour:**
It should get grouped in the batch containing that same carrier.

**Cause:**
When the SO is confirmed without a carrier, and then you edit the carrier in the delivery step (pick) then press on validate, it gets the carrier from the vals in the SO (blank), then along the way while validating the pick, `find_auto_batch` gets called with the empty carrier, so it creates a new batch, *AND then* it propagates the carrier from the SO to the pick using `get_new_picking_values` method, but by then it's too late as the batch is already created with the blank carrier.

**Solution:**
The problem happened because of this commit: https://github.com/odoo/odoo/commit/7c32adcd82119c35485addd1b198a7fcc0053c9f
We need to add the propagation of the carrier_id to the vals if it was already changed or set before the validation of the pick.

opw-5882310

### DIFF
--- a/addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
+++ b/addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
@@ -1,7 +1,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo import Command
-from odoo.tests import common, tagged
+from odoo.tests import common, tagged, Form
 
 
 @tagged('-at_install', 'post_install')
@@ -74,3 +74,48 @@ class TestDeliveryPickingBatch(common.TransactionCase):
         package = pack_wizard.action_put_in_pack()
         batch.action_done()
         self.assertEqual(package.weight, 3.0)
+
+    def test_auto_batch_carrier_change_after_confirmation(self):
+        """
+            Test an auto batch scenario where pickings correctly join the correct batches when carrier is set after confirmation.
+            The pickings looks like this:
+            - pick_1: confirmed with self.local_delivery_carrier
+            - pick_2: confirmed with carrier = False > self.local_delivery_carrier is set after confirmation
+            The expected result is:
+            - pick_1, pick_2 are in the same batch
+        """
+        warehouse = self.picking_type_out.warehouse_id
+        warehouse.delivery_steps = 'pick_ship'
+        warehouse.out_type_id.write({'auto_batch': True, 'batch_group_by_carrier': True})
+
+        partner_1, partner_2 = self.env['res.partner'].create([{
+            'name': f'{partner}',
+        } for partner in ('bob', 'marley')])
+        self.env['stock.quant']._update_available_quantity(self.product_a, self.stock_location, 40)
+        carrier_1 = self.local_delivery_carrier
+        pick_1, pick_2 = self.env['stock.picking'].create([{
+            'name': f"Pick {carrier_1.name if carrier else 'with no carrier'}",
+            'partner_id': partner,
+            'picking_type_id': warehouse.pick_type_id.id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': warehouse.wh_output_stock_loc_id.id,
+            'carrier_id': carrier,
+            'move_ids': [Command.create({
+                'product_id': self.product_a.id,
+                'product_uom_qty': 1,
+                'location_id': self.stock_location.id,
+                'location_dest_id': warehouse.wh_output_stock_loc_id.id,
+            })]
+        } for carrier, partner in [(carrier_1.id, partner_1.id), (False, partner_2.id)]])
+
+        pick_1.action_confirm()
+        pick_1.button_validate()
+        ship_1 = pick_1.move_ids.move_dest_ids.picking_id
+        self.assertEqual(ship_1, ship_1.batch_id.picking_ids)
+
+        pick_2.action_confirm()
+        with Form(pick_2) as picking_form:
+            picking_form.carrier_id = self.local_delivery_carrier
+        pick_2.button_validate()
+        ship_2 = pick_2.move_ids.move_dest_ids.picking_id
+        self.assertEqual(ship_1.batch_id.picking_ids, ship_1 | ship_2)
--- a/addons/stock_delivery/models/stock_move.py
+++ b/addons/stock_delivery/models/stock_move.py
@@ -41,6 +41,10 @@ class StockMove(models.Model):
     def _get_new_picking_values(self):
         vals = super(StockMove, self)._get_new_picking_values()
         carrier_id = self.reference_ids.sale_ids.carrier_id.id
+        # check if the previous picking have a carrier_id, then take carrier from that
+        # earlier we were taking carrier from sale but since carrier can be changed or updated in next steps so now we take carrier from prev picking
+        if len(self.move_orig_ids.picking_id.carrier_id) == 1:
+            carrier_id = self.move_orig_ids.picking_id.carrier_id.id
         vals['carrier_id'] = any(rule.propagate_carrier for rule in self.rule_id) and carrier_id
         return vals
 


addons/delivery_stock_picking_batch/models/stock_picking.py
```diff
@@ -28,6 +28,15 @@ def _get_batch_group_by_keys(self):
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    def write(self, vals):
+        res = super().write(vals)
+        if 'carrier_id' in vals:
+            for picking in self:
+                if picking.picking_type_id.batch_group_by_carrier and picking.state == 'assigned':
+                    picking.batch_id = False
```
lase-odoo: I do not think that this line is really expected because the `_find_auto_batch` method and feature is only expected to be triggered if the picking does not already have a batch:
https://github.com/odoo/odoo/blob/a5e1a0a60a387ea6dbedfbdfedc5868b3054bbf5/addons/stock_picking_batch/models/stock_picking.py#L197-L201
I understand the functional point of view that if the picking was auto_batched because of a certain carrier and we change the carrier it might be desirable to also re-evaluate the `auto_batch` to which our picking would be assigned to, but if the batch was manually created or was created for other reasons, IMO, we do not want to re-evaluate it ! Also, the fact that I can not find a similar mechanism for other types of batching (such as `batch_group_by_partner` which could also be reset the batch in case you change the partner) incentives me to think that it is not the expected behavior.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+            'name': 'Carrier 2',
+            'delivery_type': 'fixed',
+            'fixed_price': 15.0,
+            'product_id': self.env['product.product'].create({
+                'name': 'Delivery Carrier 2',
+                'type': 'service',
+                'categ_id': self.env.ref('delivery.product_category_deliveries').id,
+                'list_price': 15.0,
+            }).id
+        })
```
lase-odoo: It is best to create records in batch as it allows one to rely on the orm optimization of the amount of sql queries. required to alter data's in DB.

To be more precise, the create method of models can be given a list of vals.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+        warehouse = self.env['stock.warehouse'].search([], limit=1)
+        picking_type = self.env['stock.picking.type'].create({
+            'name': 'Test Delivery with Auto Batch',
+            'sequence_code': 'TEST',
+            'code': 'outgoing',
+            'company_id': self.env.company.id,
+            'warehouse_id': warehouse.id,
+            'auto_batch': True,
+            'batch_group_by_carrier': True,
+        })
```
lase-odoo: As a general rule of thumb it is best to use data's that are already available in the test class if there is no need to re-recreate a new record (ofc if you need to change half of the values of a given record, it is best to re-create it for readability). Here, you can simply use the `self.picking_type_out` as picking_type:
```suggestion
        self.picking_type_out.write({'auto_batch': True, 'batch_group_by_carrier': True})
```

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+            }).id
+        })
+        carrier_2 = self.env['delivery.carrier'].create({
+            'name': 'Carrier 2',
+            'delivery_type': 'fixed',
+            'fixed_price': 15.0,
+            'product_id': self.env['product.product'].create({
+                'name': 'Delivery Carrier 2',
+                'type': 'service',
+                'categ_id': self.env.ref('delivery.product_category_deliveries').id,
```
lase-odoo: In test class you can use `self.ref` rather than `self.env.ref(...).id`

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
lase-odoo: It is best to create records in batch as it allows one to rely on the orm optimization of the amount of sql queries. required to alter data's in DB.

To be more precise, the create method of models can be given a list of vals. I would therefore make a single create to create both delivery products and then a single create to create both carrier:
```suggestion
        delivery_services = self.env['product.product'].create([{
            'name': f"Delivery Service {i+1}",
            'type': 'service',
            'categ_id': self.ref('delivery.product_category_deliveries'),
            'list_price': 10.0,
        } for i in range(2)])
        carrier_1, carrier_2 =  self.env['delivery.carrier'].create([{
            'name': f'Carrier with {service.name}',
            'product_id': service.id,
        } for service in delivery_services])
```
Also, certain fields are computed and do not require to be set, in the present case, the `fixed_price` of the carrier is already computed from the `price` of its delivery service and its type is set by default to `fixed` so that there is no need to add these to the vals.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+                'list_price': 15.0,
+            }).id
+        })
+
+        self.env['stock.quant']._update_available_quantity(self.product_a, self.stock_location, 40)
+
+        picking_1 = self.env['stock.picking'].create({
```
lase-odoo: The same holds here for your deliveries, they can be created in batch and regarding my comment on the fix I do not think that it is useful to have that many deliveries. Also, as a general rule of thumb if you can be more precise in the variable name it is better. Here it is not any type of picking it is a delivery and it is best to name it delivery for clarity. Otherwise the next dev that will read the code will need to read the vals to determine the nature of the flow.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.customer_location.id,
+        })
+        picking_2.action_confirm()
+
+        # simulating setting carrier after confirming SO without setting a carrier
+        picking_2.write({'carrier_id': carrier_1.id})
+
+        self.assertEqual(picking_2.batch_id, picking_1.batch_id,
+                        "Picking 2 should join the same batch as picking 1 (carrier 1)")
```
lase-odoo: Imo, when the assert is clear on its own there is no need to add a description.
```suggestion
        self.assertEqual(picking_2.batch_id, picking_1.batch_id)
```

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+        picking_1 = self.env['stock.picking'].create({
+            'picking_type_id': picking_type.id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.customer_location.id,
+            'carrier_id': carrier_1.id,
+        })
+        self.env['stock.move'].create({
+            'product_id': self.product_a.id,
+            'product_uom_qty': 1,
+            'product_uom': self.product_a.uom_id.id,
```
lase-odoo: By contrast with `X2one` fields, the `one2Many` and `Many2Many` fields can be created during the creation of their parent record using `Command`. You can `from odoo.fields import Command` at the start of the file and:
in the vals of the `stock.picking` creation you can add a:
```py
'move_ids': [Command.create({
        'product_id': self.product_a.id,
        'product_uom_qty': 1,
        'location_dest_id': self.customer_location.id,
    })],
```
Here is a concrete example:
https://github.com/odoo/odoo/blob/a2b3a10255dba290ea462b9193ae11c54d8dd5e0/addons/mrp/tests/test_multicompany.py#L303-L313
Note that this version of the code will rely on the orm to once more minimize the amount of queries to create the related recordset.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+            'picking_id': picking_2.id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.customer_location.id,
+        })
+        picking_2.action_confirm()
+
+        # simulating setting carrier after confirming SO without setting a carrier
+        picking_2.write({'carrier_id': carrier_1.id})
```
lase-odoo: You should not call the write method directly, notably because a value can be set on a field without every calling the `write` method of the model it self. It is even more true in your use case since you chose to add logic to the write method but it is totally possible to edit the carrier_id of the record without ever calling this write method. I would rather process the flow as you describe in your steps to reproduce by editing the form of the picking. There is a `Form` class you can import from `odoo.tests` and rely on to simulate form edition:
```suggestion
        with Form(picking_2) as picking_form:
            picking_form.carrier_id = carrier_1
```
Note that in the current state of the test this will fail because you are not allowed to edit the carrier of a picking if you do not have a set `partner_id` (that is the delivery adress) ! You should definitely have one before confirming your deliveries. Hence you shoudl also create a partner:
```py
    partner = self.env['res.partner'].create({'name': 'bob'})
```
That you shoudl set in the `parnter_id` field of each of your deliveries (in the create vals). It is perfectly fine to have a single partner for all your deliveries.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,132 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+        """
+            Test an auto batch scenario where pickings correctly join the correct batches when carrier is set after confirmation.
+            The pickings looks like this:
+            - picking_1: confirmed with carrier_1
+            - picking_2: confirmed without carrier > carrier_1 is set after confirmation
+            - picking_3: confirmed with carrier_2
+            - picking_4: confirmed with carrier_1 > carrier_2 is set after confirmation
+            The expected result is:
+            - batch_1: picking_1, picking_2
+            - batch_2: picking_3, picking_4
```
lase-odoo: Since in the end we will not re-batch if the carrier is changed after the batch creation, I am not sure that that this many records are required to test the flow. I leave you decide about the minimal set up.

addons/delivery_stock_picking_batch/models/stock_picking.py
```diff
@@ -28,6 +28,15 @@ def _get_batch_group_by_keys(self):
...
+                    picking._find_auto_batch()
```
lase-odoo: Given this, I also would also avoid triggering the `_find_auto_batch` when we do write a falsy vals on `carrier_id`. hence I think it is probably for the condition to be something of the sort:
```suggestion
        if vals.get('carrier_id'):
            for picking in self:
                if picking.picking_type_id.batch_group_by_carrier and not picking.batch_id and picking.state == 'assigned':
                    picking._find_auto_batch()
```

lase-odoo (commented):

I personally think that the current fix goes to far as it also changes behavior that should not be (the batch re-evaluation in case the picking is already in a batch once the carrier is changed). 

I also left a few comments on the test. I'll re-review once all these changed are applied.

addons/delivery_stock_picking_batch/models/stock_move.py
```diff
@@ -0,0 +1,15 @@
...
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    def _get_new_picking_values(self):
+        vals = super()._get_new_picking_values()
+        if any(rule.propagate_carrier for rule in self.rule_id):
+            origin_picking = self.move_orig_ids.picking_id
+            if origin_picking.carrier_id:
+                vals['carrier_id'] = origin_picking.carrier_id.id
```
lase-odoo: I do not think that the fix should be done in the `delivery_stock_picking_batch` module. Actually, the issue seem to have been introduced by this change: https://github.com/odoo/odoo/commit/7c32adcd82119c35485addd1b198a7fcc0053c9f
The carrier used to be propagated at creation prior to the `_find_auto_batch` but the carrier propagation from origin to dest was post-pone to the validation to take carrier change into account. Imo, it make sense to keep both propagation. It is necessary at creation for certain flows such as yours with the `_find_auto_batch` so that I would porpagate it in the `_get_new_picking_values`:
https://github.com/odoo/odoo/blob/accaa246d5818572996660f967624978e906b5c3/addons/stock/models/stock_move.py#L1422
and it make sense at validation in order to consider the post creation carrier changes into account.

However, even if the carrier value used to be set just as you do, since self is not a single move, i think that the current formulation of the code is incorrect since we might end up with a `origin_picking.carrier_id` that is a record set of more than one carrier which would lead to a traceback because of `origin_picking.carrier_id.id`.

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,48 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+            'location_id': self.stock_location.id,
+            'location_dest_id': warehouse.wh_output_stock_loc_id.id,
+            'carrier_id': info[0],
+            'move_ids': [Command.create({
+                'product_id': self.product_a.id,
+                'product_uom_qty': 1,
+                'location_id': self.stock_location.id,
+                'location_dest_id': warehouse.wh_output_stock_loc_id.id,
+            })]
+        } for info in [(carrier.id, partner_1.id), (False, partner_2.id)]])
```
lase-odoo: I think it is clearer to name things explicitly: 
```suggestion
        } for carrier, partner in [(carrier.id, partner_1.id), (False, partner_2.id)]])
```

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,48 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+
+        pick_1.action_confirm()
+        pick_1.button_validate()
+        ship_1 = pick_1.move_ids.move_dest_ids.picking_id
+        self.assertTrue(ship_1.batch_id)
```
lase-odoo: I would rather make an assert on the `picking_ids` of the `ship_1.batch_id`:
```suggestion
        self.assertEqual(ship_1, ship_1.batch_id.picking_ids)
```

addons/delivery_stock_picking_batch/tests/test_delivery_picking_batch.py
```diff
@@ -74,3 +74,48 @@ def test_batch_picking_pack_shipping_weight_compute(self):
...
+
+        pick_2.action_confirm()
+        with Form(pick_2) as picking_form:
+            picking_form.carrier_id = self.local_delivery_carrier
+        pick_2.button_validate()
+        ship_2 = pick_2.move_ids.move_dest_ids.picking_id
+        self.assertEqual(ship_1.batch_id, ship_2.batch_id)
```
lase-odoo:
```suggestion
        self.assertEqual(ship_1.batch_id.picking_ids, ship_1 | ship_2)
```

lase-odoo (commented):
Thanks for the new proposition ! 🐝

PR #248503, opened 2026-02-12
[FIX] web: prevent crash when ContactImageField has no value

hubvd:
Steps to reproduce
==================

- Go to contacts
- Open studio
- Switch to the form view
- Open the x2many form view

=> TypeError: Cannot read properties of undefined (reading '0')

opw-5885953


aab-odoo (approved):
robodoo r+

PR #248505, opened 2026-02-12
[IMP] mail: cleanup and test tracking usage

tde-banana-odoo:
Main purposes

 * cover various use cases of tracking in addons. Several addons are doing
   manual calls to tracking API: '_mail_track' (generating tracking values),
   '_message_track' (generating valu and a tracking message) or calling
   'message_post' or '_message_log' with tracking values manually generated.
   Logging on a parent model is also a frequen need (e.g. move line change on
   move, esg factor line changes on factor, ...). Those should be covered by
   tests to assess the behavior.
 * update tracking API to remove low-level data manipulation and use simpler
   methods. Addons should not have to manipulate tracking values in any
   form.
 * locate tracking code being currently both generic for BaseModel and specific
   to mail.thread in a middle-ground mixin.

Specifications

Test various use case of tracking, notably when doing manual calls to
_message_track (which generates a tracking message), _mail_track
(which generates tracking values, to be included in a message_post)
or logging a message using either manually crafted tracking values or
a rendered body looking like tracking values.

Update docstring and arguments of tracking API. Add type annotations to
ease understanding of the various structures used in code.

Add a 'mail.track.mixin' mixin that allows to generate tracking values
without depending on 'mail.thread'. It also allows to remove part of the
code that was on BaseModel just for models not inheriting from 'mail.thread'.
Thread now mainly adds the posting capabilities based on tracking.

Update tracking API. Introduce '_track_add' when adding a manual tracking
on some values. Introduce '_track_record' when sending the tracking values
on another model (e.g. a parent for logging purpose).

Fix various crashing code bits discovered when adding tracking tests.
Note that most are going to be backported, hence may disappear from
this PR at merge time. Youpie.

Task-5935695 ([mail, various] Cleanup and test tracking usage)
Prepares Task-3645865 ([mail] In-body tracking)

Co-Authored-By: Prakash Prajapati <ppr@odoo.com>


addons/account_peppol/models/res_partner.py
```diff
@@ -76,36 +75,6 @@ def _compute_available_peppol_eas(self):
     # HELPERS
     # -------------------------------------------------------------------------
 
-    def _log_verification_state_update(self, company, old_value, new_value):
-        # log the update of the peppol verification state
-        # we do this instead of regular tracking because of the customized message
-        # and because we want to log the change for every company in the db
```
gawa-odoo, line 82: You lost that with your change.

The button (that updates that value mainly) only changes for a single company, and the tracking reflected that.
Now it does not

addons/sale/models/sale_order.py
```diff
@@ -2076,12 +2076,10 @@ def _track_finalize(self):
             len(self) == 1
             # The method _track_finalize is sometimes called too early or too late and it
             # might cause a desynchronization with the cache, thus this condition is needed.
-            and self.id in self._fields["state"]._get_cache(self.env)
+            and self.id in self._fields['state']._get_cache(self.env)
```
Feyensv:
```suggestion
            and self.id in self._fields["state"]._get_cache(self.env)
```
un-necessary diff (probably a recent conflict during a rebase, since we recently linted all our scope).

addons/sale/models/sale_order.py
```diff
@@ -2162,13 +2160,13 @@ def _phone_get_number_fields(self):
...
-        if "state" in init_values and self.state == "sale":
-            return self.env.ref("sale.mt_order_confirmed")
-        if "state" in init_values and self.state == "sent":
-            return self.env.ref("sale.mt_order_sent")
-        return super()._track_subtype(init_values)
+        if 'state' in track_init_values and self.state == 'sale':
+            return self.env.ref('sale.mt_order_confirmed')
+        elif 'state' in track_init_values and self.state == 'sent':
+            return self.env.ref('sale.mt_order_sent')
+        return super()._track_log_get_default_subtype(track_init_values)
```
Feyensv:
```suggestion
        if "state" in track_init_values and self.state == "sale":
            return self.env.ref("sale.mt_order_confirmed")
        if "state" in track_init_values and self.state == "sent":
            return self.env.ref("sale.mt_order_sent")
        return super()._track_log_get_default_subtype(track_init_values)
```
Same here, please keep the double quoted strings :pray: (and the if instead of elif)

addons/account_peppol/models/res_partner.py
tde-banana-odoo: Actually it does, since company_id is now stored in tracking value field_info. Display itself is not updated (aka does not display company) but that's because I won't change html / js code that is going to be scrapped soonish.

tde-banana-odoo:
@robodoo r+ rebase-merge

PR #248506, opened 2026-02-12
[IMP] payment: update iDEAL logo to Wero

snrav-odoo:
Before the commit:

The iDEAL payment method was using the existed legacy iDEAL logo.

After the commit:

- Updated the display name to "iDEAL / Wero".
- Replaced the legacy iDEAL logo with the new Wero logo.
- Introduced a separate "Wero" payment brand.

task-5922938


addons/payment/data/payment_method_data.xml
```diff
@@ -4442,6 +4451,15 @@
         />
     </record>
 
+     <record id="payment_method_wero" model="payment.method">
+        <field name="name">WERO</field>
+        <field name="code">wero</field>
+        <field name="primary_payment_method_id" eval="ref('payment.payment_method_ideal')"/>
+        <field name="sequence">1000</field>
+        <field name="active">True</field>
```
kcv-odoo: It should only be active if one of related provider is active

addons/payment/data/payment_method_data.xml
```diff
@@ -4227,6 +4227,15 @@
         <field name="image" type="base64" file="payment/static/img/hipercard.png"/>
     </record>
 
+    <record id="payment_method_ideal_brand" model="payment.method">
+        <field name="name">iDEAL Brand</field>
```
nipl-odoo: Name should be just iDEAL IMO

Image

addons/payment/data/payment_method_data.xml
```diff
@@ -4227,6 +4227,15 @@
...
+        <field name="code">ideal_brand</field>
+        <field name="primary_payment_method_id" eval="ref('payment.payment_method_ideal')"/>
+        <field name="sequence">1000</field>
+        <field name="active">True</field>
```
nipl-odoo: Why this brand is active by default, isn't there a way we can activate it when iDEAL PM is activated, same for `payment_method_wero` 🤔

nipl-odoo (changes requested):
Hello Ravij 👋🏻 Thank you for your work, I've left some comments

I think the logo here should be updated as well using the one attached in the chatter.
(don't forget to adjust the dimensions if they don’t match the other logos of other PMs)


Image

addons/payment_mollie/const.py
```diff
@@ -56,6 +56,8 @@
     'mastercard',
     'amex',
     'discover',
+    'ideal_imo',
+    'wero'
```
nipl-odoo: Why did you add this only in mollie? 🤔

addons/payment_mollie/const.py
```diff
@@ -56,6 +56,8 @@
     'mastercard',
     'amex',
     'discover',
+    'ideal_brand',
+    'wero',
```
nipl-odoo: Why only mollie? There are other providers supporting iDEAL too 🤔

addons/payment/static/img/ideal_wero.png
nipl-odoo, line 1: This image should be 240 × 160 to stay consistent with the other images in our codebase.

addons/payment_mollie/const.py
```diff
@@ -56,6 +56,8 @@
...
```
nipl-odoo: Why only Mollie? 🤔 There are other providers as well that support the iDEAL PM.

nipl-odoo (changes requested):

addons/payment/data/payment_method_data.xml
```diff
@@ -1407,7 +1407,7 @@
         <field name="code">ideal</field>
         <field name="sequence">1000</field>
         <field name="active">False</field>
-        <field name="image" type="base64" file="payment/static/img/ideal.png"/>
+        <field name="image" type="base64" file="payment/static/img/ideal_wero.png"/>
```
nipl-odoo, line 1342: Also, If I am not wrong you might need an upgrade script for this change since all these templates are marked as `noupdate=1` 👀

addons/payment_adyen/const.py
```diff
@@ -32,6 +32,8 @@
     'mastercard',
     'amex',
     'discover',
+    'ideal_brand',
+    'wero',
```
nipl-odoo: deserves little comment 👀

nipl-odoo (approved):
One non blocking comment otherwise Looks good to me! 😉
- Looks like you need to add more details in your commit message.
- pre-commit is failing in your upgrade PR 👀 
    - **tip**: You can install pre-commit in local if you want to check it before pushing the code, it will automatically does all checks when you commit in local.

Since `payment_method_ideal_brand` and `payment_method_wero` need to be activated whenever `payment_method_ideal` is enabled, we could either add some logic in `write`, or keep the current approach as it is. I’m probably not the best judge for it though, so I’ll leave the call to following reviewers.

addons/payment_worldline/const.py
```diff
@@ -9,6 +9,8 @@
     'discover',
     'mastercard',
     'visa',
+    'ideal_brand',
+    'wero',
```
kcv-odoo: We weren't enabling ideal PM before for worldline? why we need to do this now? Is it improvement? If yes then don't we need to add `idea` as Primary payment method?

Also don't add `wero` in Provider which still does not support it

same for adyen and mollie

addons/payment/data/payment_method_data.xml
```diff
@@ -1407,7 +1407,7 @@
         <field name="code">ideal</field>
```
kcv-odoo, line 1339: IMO it would be better to make this PM `payment_method_ideal_wero` and add two brands `iDeal` and `wero`. Wdyt?

```suggestion
<record id="payment_method_ideal_wero" model="payment.method">
<field name="name">iDEAL / WeRO</field>
<field name="code">ideal_wero</field>
```

So in [following task](https://www.odoo.com/odoo/project/4106/tasks/5400501) we just have to add `wero` code in related providers and do mapping if needed

@AntoineVDV can I have your opinion on this? as I am not quite sure which would be better

kcv-odoo (commented):
Thank you for your work! 🦆 

Added some questions

kcv-odoo (approved):
LGTM

AntoineVDV (changes requested):
Hello, given that this PR targets a stable version, we can't do what we want with stored record data. In particular, you change the `code` of the iDEAL payment method to `ideal_wero` both in the XML data file and in the Python mapping, but only the mapping will be updated for existing databases, while the records' `code` will remain `ideal`, thus breaking the Buckaroo and Stripe (and possibly other) integrations.

Adding an upgrade script is not a solution either, because the script will run only when the database is upgraded and only in the versions targeted by your script (currently 17.0).

Let's do this instead:
1. Only change the logo of the existing `ideal` payment method; keep the `name` and `code` untouched.
2. Add the individual Wero (not "weRO") and iDEAL logos as new brands of the existing `ideal` payment method. If existing databases are not updated, everything will still work, and database admins can update the logo themselves. If existing databases are updated or new databases are created, the two logos will be created as brand payment methods and replace the original iDEAL logo, thanks to the brand mechanism.
3. Drop the upgrade script completely.

addons/payment/data/payment_method_data.xml
```diff
@@ -3910,4 +3919,14 @@
         <field name="image" type="base64" file="payment/static/img/visa.png"/>
     </record>
 
+
```
AntoineVDV:
```suggestion
```

AntoineVDV (approved):

PR #248518, opened 2026-02-12
[FIX] l10n_mx_hr_payroll: Adjust demo data to avoid test failure

prdu-odoo:
The mx l10n module uses contract data to determine the holiday days included in a payslip. If an employee's earliest contract starts after a payslip's end date, the code throws an exception. 
See: https://runbot.odoo.com/odoo/error/237666
This should not happen during regular use, but the demo data for `hr_contract` and `hr_payroll` do not account for interactions between them. 
The first contract for Mitchell Admin was generated to start at the start of current year, while one of his payslips is set to end about a month ago.
This caused an exception if the current month was January. This PR adjusts the date of the contract to avoid this discrepancy.

---

### DIFF
--- a/addons/hr_contract/data/hr_contract_demo.xml
+++ b/addons/hr_contract/data/hr_contract_demo.xml
@@ -18,7 +18,7 @@
 
     <record id="hr_contract_admin" model="hr.contract">
         <field name="name">Mitchell Admin Contract</field>
-        <field name="date_start" eval="time.strftime('%Y')+'-1-1'"/>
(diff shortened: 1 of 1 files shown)


prdu-odoo:
@jugj-odoo 
Hey, I know this is a low-priority PR, but I'd rather not have it left stale.
So I'm just reminding you to please review it when you have the time.

prdu-odoo:
@robodoo up to saas-18.2

prdu-odoo:
Drat!

PR #248563, opened 2026-02-13
[FIX] stock: prevent errors on generating lots

bhna-odoo:
Currently, an error occurs when user tries to generate lots on a picking receipt for a product that is tracked by lots.

Steps to replicate:
- Install `stock` and create a product `test` tracked `by lots`.
- Open Inventory > Operation > Receipts , create a new one with product `test` and quantity 3.
- Open the form view of the stock move for `test` using the `view` button.
- Click `Generate Serials/Lots` and type `a` in `Quantity received` and click generate. (Error-1 will occur.)
- Click `Generate Serials/Lots` and type `a` in `Quantity per lot` and click generate. (Error-2 will occur.)

Error-1 :
```
File '/home/odoo/odoo18/community/addons/stock/models/stock_move.py', line 1055, in generate_lot_qty
    line_count = int(quantity // qty_per_lot)
                     ~~~~~~~~~^^~~~~~~~~~~~~
TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'
```

Error-2 :
```
File '/home/odoo/odoo18/community/addons/stock/models/stock_move.py', line 1053, in generate_lot_qty
    if qty_per_lot <= 0:
       ^^^^^^^^^^^^^^^^
TypeError: '<=' not supported between instances of 'NoneType' and 'int'
```

Cause:
- As the user typed in alphabets the [parsefloat] returns `NaN` which will be translated to `None` python side.
- As quantity is received as None the [line1] raises Error-1.
- As qty_per_lot is recieved as None the [line2] raises Error-2.

Solution:
- Raised Usererror when quantity received or quantity per lot is either negative or not a number.

[parsefloat]: https://github.com/odoo/odoo/blob/2da791df55762ae4dc1c06d2a43fad25b09dc11b/addons/stock/static/src/widgets/generate_serial.js#L50-L51

[line1]: https://github.com/odoo/odoo/blob/2da791df55762ae4dc1c06d2a43fad25b09dc11b/addons/stock/models/stock_move.py#L1001

[line2]: https://github.com/odoo/odoo/blob/2da791df55762ae4dc1c06d2a43fad25b09dc11b/addons/stock/models/stock_move.py#L999

related-sentry-7254849206

---

### DIFF
--- a/addons/stock/models/stock_move.py
+++ b/addons/stock/models/stock_move.py
@@ -996,8 +996,10 @@ Please change the quantity done or the rounding precision of your unit of measur
         default_vals = {}
 
         def generate_lot_qty(quantity, qty_per_lot):
-            if qty_per_lot <= 0:
-                raise UserError(_("The quantity per lot should always be a positive value."))
+            if not qty_per_lot or qty_per_lot <= 0:
+                raise UserError(self.env._("The quantity per lot should always be a positive value."))
+            if not quantity or quantity <= 0:
+                raise UserError(self.env._("The quantity received should always be a positive value."))
             line_count = int(quantity // qty_per_lot)
             leftover = quantity % qty_per_lot
             qty_array = [qty_per_lot] * line_count
--- a/addons/stock/tests/test_stock_lot.py
+++ b/addons/stock/tests/test_stock_lot.py
@@ -4,7 +4,7 @@
 from odoo import Command
 from odoo.addons.stock.tests.common import TestStockCommon
 from odoo.tests import Form
-from odoo.exceptions import ValidationError
+from odoo.exceptions import UserError, ValidationError
 
 
 class TestLotSerial(TestStockCommon):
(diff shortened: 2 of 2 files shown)


maan-odoo (changes requested):
The issue occurs because `parseFloat` returns `NaN` when it attempts to parse invalid or empty text. In my opinion, this should be handled on the JS side by defaulting the value to `0` whenever `parseFloat` returns `NaN`.

https://github.com/odoo/odoo/blob/1968e4bbeb6c92ca84304722054e53e9afceec89/addons/stock/static/src/widgets/generate_serial.js#L48-L49

___
Why are we accepting text input here? Should we use input type="number" instead?

https://github.com/odoo/odoo/blob/1968e4bbeb6c92ca84304722054e53e9afceec89/addons/stock/static/src/widgets/lots_dialog.xml#L43
https://github.com/odoo/odoo/blob/1968e4bbeb6c92ca84304722054e53e9afceec89/addons/stock/static/src/widgets/lots_dialog.xml#L56

bhna-odoo:
Solved here: https://github.com/odoo/odoo/pull/248529

PR #248565, opened 2026-02-13
[IMP] pos_restaurant: remove payment selection step from deposit tour

abhk-odoo:
### In this commit:
- Remove the payment method selection step from `test_no_kitchen_confirmation_for_deposit_money`
as deposits now automatically use the first available non–pay later payment method.

Task:5927077

Related PR:
- Enterprise: https://github.com/odoo/enterprise/pull/107368


davidmonnom (approved):
@robodoo r+

PR #248571, opened 2026-02-13
[FIX] point_of_sale: correct POS receipt tax label after fiscal position

paan-odoo:
Steps:

- Install l10n_be_pos_restaurant.
- Create a restaurant POS configuration with presets.
- Assign a fiscal position to one preset that replaces 21% tax with 6%.
- Open a POS session and process an order using that preset.

Issue:

- The POS receipt still displays the 21% tax's tax group label, even though the 6% tax is correctly applied.

Cause:

- Fiscal position was not taken into account when computing the tax group label for POS receipt orderlines.

Fix:

- Apply the fiscal position when determining the POS receipt tax group label.

Task-5899938


addons/l10n_be_pos_restaurant/static/tests/tours/pos_basic_flow_tour.js
```diff
@@ -0,0 +1,68 @@
...
+            Chrome.startPoS(),
+            Dialog.confirm("Open Register"),
+            FloorScreen.clickTable("2"),
+            ProductScreen.clickPartnerButton(),
+            ProductScreen.clickCustomer("Acme Corporation"),
+            ProductScreen.clickDisplayedProduct("Desk Organizer", false),
+            {
+                ...Dialog.confirm(),
+                content: "validate the variant dialog (with default values)",
+            },
```
jipr-odoo:
```suggestion
            ProductScreen.clickDisplayedProduct("Desk Pad", false),
```
You can choose a normal product to skip the extra step of `validating the variant dialog (with default values)` this dialog

addons/l10n_be_pos_restaurant/static/tests/tours/pos_basic_flow_tour.js
```diff
@@ -0,0 +1,68 @@
...
+                            "Tax group 'A' is not displayed next to the product price on the receipt."
+                        );
+                    }
+                },
+            },
+            ReceiptScreen.clickNextOrder(),
+            FloorScreen.clickTable("2"),
+            ProductScreen.clickPartnerButton(),
+            ProductScreen.clickCustomer("Acme Corporation"),
+            ProductScreen.clickDisplayedProduct("Desk Organizer", false),
```
jipr-odoo: same here

jipr-odoo (changes requested):

addons/l10n_be_pos_restaurant/tests/test_frontend.py
```diff
@@ -0,0 +1,48 @@
...
+        })
+        self.main_pos_config.write({
+            'use_presets': True,
+            'default_preset_id': self.preset_eat_in.id,
+            'available_preset_ids': [(6, 0, [
+                self.preset_eat_in.id,
+                self.preset_take_out.id,
+            ])],
+        })
+        self.start_pos_tour('test_pos_receipt_label_on_receipt')
```
paso-odoo: Also, assert the order line tax, fiscal and total after changing the preset

paso-odoo (changes requested):

addons/point_of_sale/static/src/app/components/orderline/orderline.xml
```diff
@@ -25,7 +25,7 @@
                         </span>
                     </div>
                     <div t-if="!props.basic_receipt" t-esc="line.getPriceString()" class="product-price price fw-bolder"/>
-                    <div t-if="props.showTaxGroup" t-esc="this.taxGroup" class="text-end" style="width: 2rem"/>
+                    <div t-if="props.showTaxGroup" t-esc="this.taxGroup" class="orderline-tax-group text-end" style="width: 2rem"/>
```
paso-odoo: It would be better if we use `line-details` class instead of adding new one in 18.2

paso-odoo (changes requested):

davidmonnom (approved):
@robodoo r+

PR #248572, opened 2026-02-13
[FIX] test_http: ensure log (add tests)

thle-odoo:
Insert unit tests that verify the presence of a log when a device is detected.


Julien00859 (approved):
@robodoo r+

PR #248576, opened 2026-02-13
[FIX] point_of_sale: only show LNA checkbox for ePOS

MaxTheWhale:
This commit ensures the LNA checkbox in the POS printer form only shows for ePOS printers. This is due to IoT printers now showing their own checkbox which comes from the field on the IoT box.

---


loouis-t (approved):
robodoo r+

PR #248585, opened 2026-02-13
[FIX] l10n_latam_base: changes identification type on country change

legi-odoo:
**PROBLEM**
PR: https://github.com/odoo/odoo/pull/179078
Removed _onchange_country_id() which was used to set the identification type according to the country of the partner.
This PR reintroduce it, so id type and country remains consistent.

**STEP TO REPRODUCE**
1. install l10n_ar and l10n_co.
2. create a new partner.
3. set its country to Argentina, and select an argentinian id type.
4. set the country to Colombia and save. You end up with a partner from Colombia, with a id type that is used for Argentinian partners which shouldn't be possible.

opw-5801824

### DIFF
--- a/addons/l10n_latam_base/models/res_partner.py
+++ b/addons/l10n_latam_base/models/res_partner.py
@@ -28,6 +28,17 @@ class ResPartner(models.Model):
         super(ResPartner, partners_vat)._check_vat(validation=validation)
         (self - partners_vat)._run_check_identification(validation=validation)
 
+    @api.onchange('country_id')
+    def _onchange_country_id(self):
+        country = self.country_id or self.company_id.account_fiscal_country_id or self.env.company.account_fiscal_country_id
+        identification_type = self.l10n_latam_identification_type_id
+        if not identification_type or identification_type.country_id != country:
+            self.l10n_latam_identification_type_id = (
+                self.env['l10n_latam.identification.type'].search(
+                [('country_id', '=', country.id), ('is_vat', '=', True)], limit=1) or
+                self.env.ref('l10n_latam_base.it_vat', raise_if_not_found=False)
+            )
+
     @api.onchange('vat', 'country_id', 'l10n_latam_identification_type_id')
     def _onchange_vat(self):
         super()._onchange_vat()


addons/l10n_latam_base/models/res_partner.py
```diff
@@ -28,6 +28,15 @@ def _check_vat(self, validation='error'):
...
         (self - partners_vat)._run_check_identification(validation=validation)
 
+    @api.onchange('country_id')
+    def _onchange_country_id(self):
+        country = self.country_id or self.company_id.account_fiscal_country_id or self.env.company.account_fiscal_country_id
+        identification_type = self.l10n_latam_identification_type_id
+        if not identification_type or (identification_type.country_id != country):
+            self.l10n_latam_identification_type_id = self.env['l10n_latam.identification.type'].search(
+                [('country_id', '=', country.id), ('is_vat', '=', True)], limit=1) or self.env.ref(
+                    'l10n_latam_base.it_vat', raise_if_not_found=False)
```
yosuanicolaus:
```suggestion
            self.l10n_latam_identification_type_id = (
                self.env['l10n_latam.identification.type'].search([('country_id', '=', country.id), ('is_vat', '=', True)], limit=1) or
                self.env.ref('l10n_latam_base.it_vat', raise_if_not_found=False)
            )
```

yosuanicolaus: But actually, I don't think this is the right way to fix it. You should look at where exactly is the code that is changing the `l10n_latam_identification_type_id` to the unmatching country one and prevent it from there.

yosuanicolaus (changes requested):

addons/l10n_latam_base/models/res_partner.py
legi-odoo: Not sure what you meant by that. There isn't any existing code changing the l10n_latam_identification_type_id, and that's the problem. You can set it, and then change the country, and the id type will not be updated, leading to inconsistencies between the id type and the partner country. This wasn't the case in previous version, because there was a onchange that updated the id type when the country changes.

In any case, I updated _run_check_identification to check that the id type match the partner country.

addons/l10n_latam_base/models/res_partner.py
```diff
@@ -18,7 +19,9 @@ def _commercial_fields(self):
         return super()._commercial_fields() + ['l10n_latam_identification_type_id']
 
     def _run_check_identification(self, validation='error'):
-        return
+        for partner in self:
+            if partner.country_id and partner.l10n_latam_identification_type_id.country_id and partner.l10n_latam_identification_type_id.country_id != partner.country_id:
+                raise ValidationError(_('The indentification type must match the partner country.'))
```
legi-odoo: This new check I added broke 2 tests in l10n_ec_edi, I fixed them in the linked enterprise PR

addons/l10n_latam_base/models/res_partner.py
```diff
@@ -28,6 +31,17 @@ def _check_vat(self, validation='error'):
         super(ResPartner, partners_vat)._check_vat(validation=validation)
         (self - partners_vat)._run_check_identification(validation=validation)
 
+    @api.onchange('country_id')
+    def _onchange_country_id(self):
+        country = self.country_id or self.company_id.account_fiscal_country_id or self.env.company.account_fiscal_country_id
+        identification_type = self.l10n_latam_identification_type_id
+        if not identification_type or (identification_type.country_id != country):
```
jorenvo:
```suggestion
        if not identification_type or identification_type.country_id != country:
```

addons/l10n_latam_base/models/res_partner.py
```diff
@@ -18,7 +19,9 @@ def _commercial_fields(self):
         return super()._commercial_fields() + ['l10n_latam_identification_type_id']
 
     def _run_check_identification(self, validation='error'):
-        return
+        for partner in self:
+            if partner.country_id and partner.l10n_latam_identification_type_id.country_id and partner.l10n_latam_identification_type_id.country_id != partner.country_id:
```
jorenvo:
```suggestion
            if validation == 'error' and partner.country_id and partner.l10n_latam_identification_type_id.country_id and partner.l10n_latam_identification_type_id.country_id != partner.country_id:
```

More in line with the idea behind that parameter I think.

jorenvo (commented):
Thanks, @legi-odoo! It's a bit odd that the `_run_check_identification` doesn't run for `is_vat` identification types though, those are probably the most widely used types.

addons/l10n_latam_base/models/res_partner.py
```diff
@@ -18,7 +19,9 @@ def _commercial_fields(self):
         return super()._commercial_fields() + ['l10n_latam_identification_type_id']
 
     def _run_check_identification(self, validation='error'):
-        return
+        for partner in self:
+            if validation == 'error' and partner.country_id and partner.l10n_latam_identification_type_id.country_id and partner.l10n_latam_identification_type_id.country_id != partner.country_id:
+                raise ValidationError(_('The indentification type must match the partner country.'))
```
jorenvo: Checking a bit deeper, before 18.3, we never had a validation error for this. We only relied on the onchange, which already makes it difficult for users to bypass the check in Odoo. 

And, as I mentioned in my previous review, putting it in this method doesn't seem right, because it's not run for the `is_vat` document types which are the most common. So we would need a new method and it seems more hassle than it's worth.

How do you feel about removing the validation error entirely and just relying on the onchange like in prior versions?

PS: there's a typo: i~~n~~dentification

legi-odoo: We can go with what we add in prior versions

jorenvo (approved):
Thanks!

@robodoo r+

PR #236251, opened 2025-11-18
[FIX] l10n_it_edi_withholding: relax import tax check on withholding

lordkrandel:
Some invoice come in with a wrong ENASARCO withholding reason. We now broaden the search to allow taxes with the same withholding type to be used during import even if the withholding reason doesn't match.

In the test, I change the Enasarco tax to reason Q to check that it gets correctly assigned.

Ticket [link](https://www.odoo.com/odoo/project.task/5175587), [link](https://www.odoo.com/odoo/project.task/5933699)
opw-5175587
opw-5933699

PR #248586, opened 2026-02-13
[FIX] l10n_it_edi_doi: Remove check on DOI tax being the only one on the line

lordkrandel:
We should be able to add more taxes with the 0% on the same line, like the Enasarco and 23% RIT. Indeed in italy it is possible to have invoices with Dichiarazione d'intento togheter with a withholding and Enasarco taxes.

See also: odoo/odoo#236251

Ticket [link](https://www.odoo.com/odoo/project.task/5933699)
opw-5933699


epictete (approved):
@robodoo r+

PR #248587, opened 2026-02-13
[IMP] html_editor*: add google translate and refactor the translation

Jinjiu96:
*: html_builder, mail, ai

Before this commit: we use chatgpt for translation. The translation has
less supported languages comparing to mainstream translation providers.

After this commit: we added the google translate and use it as the
default translation provider. We refactored the translation code so now
we can easily introduce new translation service by defining a new
translator instance. The translation now has its own toolbar group and
ai group is moved to the ai module on the enterprise branch.

In debug mode, the translate_dialog also shows all the current translate
services available:
1. it's possible to select a different translate service.
2. the translation is triggered on the change of translator.
3. user may select the preferred translation and insert it.

An IAP service is also created for google translate in the iap-app repo.

task-5155746

---


addons/html_editor/static/src/main/translate/translate_dialog.js
```diff
@@ -52,32 +54,42 @@ const POSTPROCESS_GENERATED_CONTENT = (content, baseContainer) => {
...
-        this.state = useState({ selectedMessageId: null });
-        onWillDestroy(() => this.pendingRpcPromise?.abort());
-    }
-
-    selectMessage(ev) {
-        this.state.selectedMessageId = +ev.currentTarget.getAttribute("data-message-id");
-    }
+        const google_translate = new GoogleTranslator("translate_google", "Google Translate");
+        const chatgpt_translate = new ChatGPTTranslator("translate_gpt", "ChatGPT");
+        this.translators = [google_translate, chatgpt_translate];
```
dmo-odoo: I didn't know that we would keep both? I thought the goal was for google translate to replace chatgpt. How do you enable/disable different translators?

I'm not against it on principle, but when only one is available then I don't think we should see the provider name:

Image

Jinjiu96: @dmo-odoo I adapted the PR so the provider name only shows when it's in debug mode and there are more than one translate service. 

So for now, we only have two translators and they are decided based on if we are in debug mode. In non-debug mode, only the google translator is enabled; In debug mode, both are enabled.

We can make a setting config in the future if we are going to introduce more translate service. But since we only have two options now, I didn't go that far 😬

PR #248588, opened 2026-02-13
[FIX] l10n_fr_account: fix vat exigibility mention on printed invoice

pasm-odoo:
The VAT exigibility mention on French invoices was incorrectly displayed.

 - For goods, VAT on debits is the default regime, therefore the mention is not mandatory and must not be shown.
 - For services, VAT is normally due on payment. The mention “TVA payée sur les débits” must be displayed only when a service VAT uses tax exigibility based on invoice.

 This commit updates the logic to display the mention only in this specific case
 and omits it otherwise.

 Task:5418502

### DIFF
--- a/addons/l10n_fr_account/views/report_invoice.xml
+++ b/addons/l10n_fr_account/views/report_invoice.xml
@@ -52,8 +52,9 @@
 
         <xpath expr="//div[@name='qr_code_placeholder']" position="before">
             <div class="mb-3">
-                <p t-if="o.l10n_fr_is_company_french and o.move_type.startswith('out_') and 'on_invoice' in o.invoice_line_ids.mapped('tax_ids.tax_exigibility')">
-                    Option to pay tax on debits
+                <p t-if="o.l10n_fr_is_company_french and o.is_sale_document(include_receipts=True) and
+                         any(l.product_id.type == 'service' and any(t.tax_exigibility == 'on_invoice' and t.tax_scope == 'consu' for t in l.tax_ids) for l in o.invoice_line_ids)">
+                    Tax exigibility on debits
                 </p>
             </div>
         </xpath>


addons/l10n_fr_account/models/account_move.py
```diff
@@ -5,6 +5,9 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     l10n_fr_is_company_french = fields.Boolean(compute='_compute_l10n_fr_is_company_french')
+    l10n_fr_account_show_tax_exigibility = fields.Boolean(
+        compute='_compute_l10n_fr_account_show_tax_exigibility'
+    )
```
hamo-odoo: Why do we need a compute field?

hamo-odoo (changes requested):

addons/l10n_fr_account/models/account_move.py
pasm-odoo: Good catch!! I am removing it now and shifting the logic directly into the q-web template. Thanks for pointing out.

hamo-odoo (approved):
LGTM

addons/l10n_fr_account/views/report_invoice.xml
```diff
@@ -52,8 +52,9 @@
 
         <xpath expr="//div[@name='qr_code_placeholder']" position="before">
             <div class="mb-3">
-                <p t-if="o.l10n_fr_is_company_french and o.move_type.startswith('out_') and 'on_invoice' in o.invoice_line_ids.mapped('tax_ids.tax_exigibility')">
-                    Option to pay tax on debits
+                <p t-if="o.company_id.country_code == 'FR' and o.move_type.startswith('out_') and
```
FlorianGilbert: We can use `is_sale_document`

FlorianGilbert: Why you change the condition on `country_code == 'FR'` instead of `l10n_fr_is_company_french`?

I don't find in the spec where we asked for that.

FlorianGilbert (changes requested):

addons/l10n_fr_account/views/report_invoice.xml
pasm-odoo: my bad :smile: , I have udpated now with field.

FlorianGilbert (approved):
@robodoo r+

PR #248593, opened 2026-02-13
[IMP] core: add session info in wz logs

thle-odoo:
It is difficult to cross-reference information when we do not know how
to filter queries made by a specific session.

The session identifier, which corresponds to the static part of the
session ID, can be used (or at least part of it) to identify a session.

Adding this session identification to the Werkzeug logs allows us to
determine which session was used for which request.

Werkzeug uses the Common Log Format ([CLF](https://en.wikipedia.org/wiki/Common_Log_Format)).
The `ident` field is not used. It will always be `-`.
Source: [`log function`](https://github.com/pallets/werkzeug/blob/3.0.1/src/werkzeug/serving.py#L470)

So we can insert this information here (Odoo is not picky about CLF).

Note:
If a hard rotation occurs, we add the new session id.

Task-5936062

### DIFF
--- a/odoo/addons/test_http/tests/test_session.py
+++ b/odoo/addons/test_http/tests/test_session.py
@@ -256,7 +256,7 @@ class TestHttpSession(TestHttpBase):
                     "csrf_token": self.csrf_token(),
                 },
             ).raise_for_status()
-        self.assertEqual(len(capture.output), 1)
+        self.assertEqual(len(capture.output), 2)
         self.assertRegex(capture.output[0],
             r"^WARNING:odoo.http:<function odoo\.addons\.\w+\.controllers\.\w+\.logout> "
             r"called ignoring args {('session_id', 'debug'|'debug', 'session_id')}$"
--- a/odoo/http/session.py
+++ b/odoo/http/session.py
@@ -8,6 +8,7 @@ import os
 import re
 import secrets
 import tempfile
+import threading
 import time
 import typing
 from collections.abc import MutableMapping
@@ -540,6 +541,11 @@ class SessionStore:
         else:
             self.delete(session)
             session.sid = self.generate_key()
+            if hasattr(threading.current_thread(), 'sess_id'):
+                _logger.info(
+                    'Session rotated: %s -> %s', threading.current_thread().sess_id, session.sid[:8]
+                )
+            threading.current_thread().sess_id = session.sid[:8]
         if session.uid:
             assert env, "saving this session requires an environment"
             update_session_token(session, env)
--- a/odoo/service/server.py
+++ b/odoo/service/server.py
@@ -141,6 +141,16 @@ class CommonRequestHandler(werkzeug.serving.WSGIRequestHandler):
         self._sent_server_header = None
         super().__init__(*args, **kwargs)
 
+    def log(self, type, message, *args):
+        # Common Log Format:
+        #   'ip-address ident auth [timestamp] "method path version" status size'
+        # The message function parameter holds everything after the timestamp
+        # Use the (shortened) session-id as ident, leave auth empty (a dash).
+        ident = getattr(threading.current_thread(), 'sess_id', '-')
+        # Note: `ident` is not the real ident of CLF but we recycle the column
+        CLF_message = f'{self.address_string()} {ident} - [{self.log_date_time_string()}] {message}\n'
+        werkzeug.serving._log(type, CLF_message, *args)
+
     def log_request(self, code='-', size='-'):
         try:
             path = uri_to_iri(self.path)
--- a/odoo/http/router.py
+++ b/odoo/http/router.py
@@ -252,6 +252,8 @@ class Application:
             del current_thread.dbname
         if hasattr(current_thread, 'uid'):
             del current_thread.uid
+        if hasattr(current_thread, 'sess_id'):
+            del current_thread.sess_id
         current_thread.rpc_model_method = ''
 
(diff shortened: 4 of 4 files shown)


Julien00859 (commented):
You can change the perf-info directly

thle-odoo:
ci/security
- `PY016`: `getattr` used with a hardcoded name
- `PY031`: `del current_thread.sess_id` delete specific attribute

Xavier-Do:
The format using a pipe looks a little arbitrary, the werkzeug logger currently respects one columns per piece of information. I would at least remove that.

Moreover, this information is not related to the werkzeug logger only (log_request) since the current_thread could have information about an ongoing request having a session for any operation and any log line. (it is  not the case of the perf_info, only making sense at the end of a request)

image

**I wonder if this should be part of the default format**, next to the db

It would allow to have this information on all lines, not only werkzeug requests
It would make it easier to spot visually, in one column

image

The cons is that the column is logged in all case shifting the log line on the right by two even without session on the thread.

**Alternatively if we really only want them in the werkzeug logs** it could be added next to the [werkzeug ip](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/serving.py#L478)

It looks like werkzeug follows [CLF](https://en.wikipedia.org/wiki/Common_Log_Format), so we have two useless column. The second one is a tcp ident, the third one is supposed to be the authenticated http user which we don't use, but it could make sense to have this session information there. To discuss but those columns are more likely always useless in odoo right now, and never used, it would be a nice way to use them)  

image
This could be a bad idea if we use some log parser that are aware that the message is in CLF but it looks unlikely since we have extra columns specific to odoo at the beginning

Julien00859:
Hi @thle-odoo, in master-h11-juc I'm removing the PerfFilter and am reimplementing it properly as we now have control over the logs emitted by the server. I appreciate your refactor to `s/perf_info/extra_info/` but as I'm redoing the feature entirely your refactor won't be long-lived and I think going for a minimal diff would be better. Care to do a minimal diff? i.e. modify `PerfFilter` directly.

Julien00859:
Another idea would be to do those changes directly inside my branch (you're free to push a commit on top). I think as we're already changing the logger name (from werkzeug to odoo.http.server) we can also use the opportunity to write the short session id closer to the ip address.

odoo/netsvc.py
```diff
@@ -356,3 +356,13 @@ def showwarning_with_traceback(message, category, filename, lineno, file=None, l
 def runbot(self, message, *args, **kws):
     self.log(logging.RUNBOT, message, *args, **kws)
 logging.Logger.runbot = runbot
+
+
+def _werkzeug_CLF_log(self, type, message, *args):
+    # Format: ip_address ident http_auth [timestamp] "message" status size
```
Xavier-Do: I would add a small comment saying that the ident is not the "real" ident of CLF but that we recycle the column since we don't care about keeping it exactly CLF

odoo/http/session.py
```diff
@@ -540,6 +541,8 @@ def rotate(self, session: Session, env: Environment, *, soft: bool = False) -> N
         else:
             self.delete(session)
             session.sid = self.generate_key()
+            if hasattr(threading.current_thread(), 'sess_id'):
+                threading.current_thread().sess_id += f'/{session.sid[:8]}'
         if session.uid:
             assert env, "saving this session requires an environment"
```
Xavier-Do, line 550: I think it would be better to set the new value and log the change.
This way the column meaning and sizes are always the same, the current value, not an arbitrary format. This would keep stuff aligned and easier to parse.

Also it is strange to set the new value only if we already have a sess_id on the thread, is it possible?  Shouldn't we always set the new value? 


```suggestion
            if hasattr(threading.current_thread(), 'sess_info'):
                _logger.info('Session rotated: %s -> %s', threading.current_thread().sess_info, session.static_sid[:8])
            threading.current_thread().sess_info = session.static_sid[:8]
```

thle-odoo: The rotation logic may be triggered during `save_session`, and thus within the context of a request; this request must be handled by the logic that sets the `sess_id` attribute on the current thread.

But that doesn't matter anymore with the change you are suggesting (which I have applied because of the advantages you mentioned)

odoo/service/server.py
```diff
@@ -141,6 +141,14 @@ def __init__(self, *args, **kwargs):
         self._sent_server_header = None
         super().__init__(*args, **kwargs)
 
+    def log(self, type, message, *args):
+        # CLF: 'ip_address ident http_auth [timestamp] "message" status size'
+        # `message` param represents '"message" status size' part of the CLF
```
Julien00859:
```suggestion
        # Common Log Format:
        #   ip-address ident auth [timestamp] "method path version" status size'
        # The message function parameter holds everything after the timestamp
        # Use the (shortened) session-id as ident, leave auth empty (a dash).
```

odoo/addons/test_http/tests/test_session.py
```diff
@@ -256,7 +256,7 @@ def test_session10_explicit_session(self):
                     "csrf_token": self.csrf_token(),
                 },
             ).raise_for_status()
-        self.assertEqual(len(capture.output), 1)
+        self.assertEqual(len(capture.output), 2)
```
Julien00859, line 259: why it changed?

thle-odoo: Because we add a log when the session rotates (hard rotation).
The goal being to have all the necessary information to trace the requests for a given session.

Julien00859:
@robodoo r-

Julien00859:
@robodoo r+

PR #248596, opened 2026-02-13
[IMP] mail, hr: improvements in Members

zel-odoo:
1. align better in channel member invitation popover
2. allow searching external users in channel member invitation popover
3. allow opening employee form view when clicking on employee name in the notification in Discuss
4. update strings used in some use cases.

task-5881048

---

### DIFF
--- a/addons/mail/static/src/core/common/notification_message.js
+++ b/addons/mail/static/src/core/common/notification_message.js
@@ -1,8 +1,10 @@
-import { useRef } from "@web/owl2/utils";
+import { useRef, useState } from "@web/owl2/utils";
+import { AvatarCardPopover } from "@mail/discuss/web/avatar_card/avatar_card_popover";
 import { useForwardRefsToParent } from "@mail/utils/common/hooks";
 import { Component, htmlEscape } from "@odoo/owl";
 import { _t } from "@web/core/l10n/translation";
 import { useService } from "@web/core/utils/hooks";
+import { usePopover } from "@web/core/popover/popover_hook";
 
 export class NotificationMessage extends Component {
     static template = "mail.NotificationMessage";
@@ -14,13 +16,33 @@ export class NotificationMessage extends Component {
         useForwardRefsToParent("messageRefs", (props) => props.message.id, this.root);
         this.htmlEscape = htmlEscape;
         this.store = useService("mail.store");
+        this.avatarCard = usePopover(AvatarCardPopover, {
+            arrow: false,
+            onClose: () => (this.state.isAvatarCardOpen = false),
+            popoverClass: "mx-2",
+        });
+        this.state = useState({
+            isAvatarCardOpen: false,
+        });
     }
 
     /**
      * @param {MouseEvent} ev
      */
     async onClickNotificationMessage(ev) {
-        this.store.handleClickOnLink(ev, this.props.thread);
+        const model = ev.target.dataset.oeModel;
+        const id = Number(ev.target.dataset.oeId);
+        if (ev.target.tagName === "A" && model === "res.partner" && id) {
+            if (!this.avatarCard.isOpen) {
+                this.avatarCard.open(ev.currentTarget, {
+                    id,
+                    model,
+                });
+                this.state.isAvatarCardOpen = true;
+            }
+        } else {
+            this.store.handleClickOnLink(ev, this.props.thread);
+        }
         const { oeType, oeId } = ev.target.dataset;
         if (oeType === "highlight") {
             await this.env.messageHighlight?.highlightMessage(
--- a/addons/mail/models/discuss/res_partner.py
+++ b/addons/mail/models/discuss/res_partner.py
@@ -86,7 +86,6 @@ class ResPartner(models.Model):
                 [("active", "=", True)],
                 [("user_ids", "!=", False)],
                 [("user_ids.active", "=", True)],
-                [("user_ids.share", "=", False)],
             ]
         )
         channel = self.env["discuss.channel"]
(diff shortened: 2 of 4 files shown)


addons/hr/static/src/core/web/store_service_patch.js
```diff
@@ -0,0 +1,31 @@
+import { Store } from "@mail/core/common/store_service";
+
+import { patch } from "@web/core/utils/patch";
+
+/** @type {import("models").Store} */
+const StorePatch = {
+    setup() {
+        super.setup(...arguments);
+    },
```
alexkuhn: Useless `setup()`

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -165,18 +165,21 @@ const StorePatch = {
...
-                    views: [[false, "form"]],
-                    res_id: id,
-                })
-            ).then(() => this.onLinkFollowed(thread));
+            this._handleClickOnRecord(ev, model, id, thread);
             return true;
         }
         return false;
     },
+    async _handleClickOnRecord(ev, model, id, thread) {
```
alexkuhn: Nitpick:
I'd prefer `(ev, { model, id, thread })` as signature:
- `ev` is a natural positional for the method
- `thread` looks optional
- ordering of `model` / `id` and the many positionals is cumbersome, so prefer explicitly named param for `model` / `id` at least.

addons/hr/static/src/core/web/store_service_patch.js
```diff
@@ -0,0 +1,31 @@
...
+                    type: "ir.actions.act_window",
+                    res_model: "hr.employee",
+                    views: [[false, "form"]],
+                    res_id: employee_id,
+                })
+                return;
+            }
+        }
+        return super._handleClickOnRecord(...arguments);
+    },
```
alexkuhn: Ok for click of mention to redirect to employee, but when CTRL/CMD+click this keeps redirect to the partner form view.

That's because the `href` still points to the partner.

I think that we should have fetch + update of href of the prettified message content, so that CTRL/CMD+click works... Or we do some shenanigans with intercepting CTRL/CMD+click to forge another href, but I have low expectations this would work.

CTRL/CMD+click is a frequent usage, therefore I think it's important to make this work consistently with the plain click to redirect to same record.

zel-odoo: But don't we want to get rid of `prepareMessageBody` finally?

alexkuhn: The discussions around removing `prepareMessageBody` comes more about being a side-effect on rendering of `Message` component that should instead be moved to logic part in `MessageModel.richBody` for example.
Functionally we need to parse and adjust the `href` for current user / target mention to have either partner or employee URL, so that CTRL/CMD+click works as expected by this IMP.

addons/hr/static/src/core/common/notification_message_patch.js
alexkuhn, line 1: I would expect to also do the same with shift + click on mention in a normal message.

zel-odoo: you mean you want to have this patch not in notification_message_patch but message_patch?

alexkuhn: Hm actually I would suggest something else slightly.

In master, there's inconsistent of `@mention` in Message and this `@mention` in the NotificationMessage. With this PR, this is a bit less inconsistent, thanks to click on NotificationMessage redirecting to the employee profile like the "View Profile" of the avatar card.

However, I think the 2 `@mention` should be more consistent, and I think the best is avatar card.
Therefore I would suggest for the click on `NotificationMessage` to open the avatar card, same as an `@mention` in `Message`.

zel-odoo: ok i see what you mean, then we need to update the spec in the task maybe. (i think your idea is better and we should do as you said)

addons/mail/static/src/core/common/notification_message.js
```diff
@@ -14,13 +16,33 @@ export class NotificationMessage extends Component {
...
+            if (!this.avatarCard.isOpen) {
+                this.avatarCard.open(ev.currentTarget, {
+                    id,
+                    model,
+                });
+                this.state.isAvatarCardOpen = true;
+            }
+        } else {
+            this.store.handleClickOnLink(ev, this.props.thread);
+        }
```
alexkuhn, line 45: I feel that the handling of opening an avatar card could be built-in to `handleClickOnLink()` based on `o_mail_redirect` or equivalent classname on the target (which is missing in `NotificationMessage` by the way, contributing to the inconsistent style with mentions in `Message`).

But I don't mind to do such cleaning later. This might be more complicated than it looks.

alexkuhn (approved):
LGTM.

What about re-targeting to 19.2, since in practice 19.2 is not really a "stable" branch until next week when actual DB will start using it? We'd have these improvements 2 months earlier thanks to being in 19.2 rather than 19.3. I let you make the final decision.

@robodoo delegate+

addons/mail/static/src/core/common/notification_message.js
seb-odoo: It should definitely be done in `handleClickOnLink`, this is the whole point of the method.
If we do it here, we'll only fix `onClickNotificationMessage` but not from other places.

alexkuhn: There's some copy-paste with `Message`, hence why this diff makes it consistent with `Message`. But I agree that it feels best to have it in `handleClickOnLink` to work in any context.

It's just that I see some other parts missing, like the `o_mail_redirect` classname or equivalent in `NotificationMessage`. The PR is pretty good and I feel other improvements could unnecessarily drag the task / PR.

seb-odoo: I'm fine with either ordering of PR.

PR #248597, opened 2026-02-13
[FIX] mail: prevent duplicate dropdown in pinned messages panel

akus-odoo:
----------------------------------------------
When multiple messages are pinned, repeatedly right-clicking different
messages in the pinned messages panel can open multiple context dropdowns
simultaneously. This may lead to UI inconsistencies and potential crashes.

----------------------------------------------
- Multiple pinned messages can be right-clicked consecutively
- Each right-click may open an additional dropdown menu
- Duplicate dropdowns can remain visible at the same time
- This can cause unstable or inconsistent UI behavior

----------------------------------------------
- Only one dropdown menu is displayed at a time in the pinned messages panel
- Repeated right-clicks properly reuse or replace the existing dropdown
- Prevents duplicate dropdowns and avoids potential UI crashes

Task-5935503

----------------------------------------------

### DIFF
--- a/addons/mail/static/src/core/common/message.js
+++ b/addons/mail/static/src/core/common/message.js
@@ -109,7 +109,7 @@ export class Message extends Component {
             emailHeaderOpen: false,
         });
         this.rightClickDropdownState = useDropdownState({
-            onClose: () => this.props.messageSelection.clearSelected(),
+            onClose: () => this.props.messageSelection?.clearSelected(),
         });
         this.rightClickAnchor = useChildRef("rightClickAnchor");
         /** @type {ShadowRoot} */


addons/mail/static/tests/discuss/message_pin/pinned_messages.test.js
```diff
@@ -189,3 +190,30 @@ test("Internal user can see unpin button", async () => {
...
+        {
+            body: "Pinned message 2",
+            model: "discuss.channel",
+            res_id: channelId,
+            pinned_at: "2024-01-01 10:05:00",
+        },
+    ]);
+    await start();
+    await openDiscuss(channelId);
+    await click(".o-mail-DiscussContent-header button[title='Pinned Messages']");
```
ramh-odoo: 🔨 Try to keep selectors concise to improve clarity and make the code easier to maintain.
```suggestion
    await click("button[title='Pinned Messages']");
```

ramh-odoo (approved):
LGTM otherwise

addons/mail/static/src/core/common/message.js
```diff
@@ -109,7 +109,7 @@ export class Message extends Component {
             emailHeaderOpen: false,
         });
         this.rightClickDropdownState = useDropdownState({
-            onClose: () => this.props.messageSelection.clearSelected(),
+            onClose: () => this.props.messageSelection?.clearSelected(),
```
ramh-odoo, line 112: It still makes sense to fix this, since `messageSelection` is an optional prop here.
Maybe we could remove the test, as a similar one has already been added in the another [PR](https://github.com/odoo/odoo/pull/249246).

cc @alexkuhn

alexkuhn: This test doesn't make sense with https://github.com/odoo/odoo/pull/249246 since the actual intentional behavior is no dropdown showing of message actions on right-click in pinned panel.

The assertion steps are negative, hence why I enriched an existing test instead of making another test.
So yes, the test is irrelevant.
The `messageSelection?.clearSelected()` is still a genuine fix to do, not sure this is reachable code in practice with fix of https://github.com/odoo/odoo/pull/249246 but I'd keep fixing it anyway :)

alexkuhn (approved):
@robodoo delegate+ r+

PR #248598, opened 2026-02-13
[IMP] hr_payroll: show entry date on payslip instead of contract start

FedericoBraidi:
In the paysplip definition, the current contract's start date is used. But if a person changes job or contract internally we don't want this value to change and we want it fixed to when the person joined the company. Notably, if a person worked at the same company in two well distinct periods, we want to consider the beginning of this period and not of the previous one(s).

Here we create a new function, similar to _get_first_version_date() which returns the beginning of the contract tied to the first version of the period, instead of the beginning of the version itself, which might be before or after the actual date of the start of the contract. We call this function _get_first_contract_date().

Task: 5909637
Enterprise PR: https://github.com/odoo/enterprise/pull/106692

### DIFF
--- a/addons/hr/models/hr_employee.py
+++ b/addons/hr/models/hr_employee.py
@@ -324,7 +324,7 @@ class HrEmployee(models.Model):
             versions = versions.filtered(lambda c: c.date_start <= self.env.context['before_date'])
         return versions
 
-    def _get_first_version_date(self, no_gap=True):
+    def _get_first_versions_filtered(self, no_gap=True):
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


ViviSOMT:
@FedericoBraidi thank you for your ticket! I struggled to find a working entry date field for the salary slip. We're using the Swiss-Payroll-App and the field hr.payslip_l10n_ch_entry is titled as the entry date but it takes the newest contract date. Maybe a specific field in the employee / payroll app could solve the problem.

tivisse:
@robodoo r+

PR #248602, opened 2026-02-13
[FIX] web: signature field notification service

LauraCForgeFlow:
The notification service is later used in this [method](https://github.com/odoo/odoo/blob/04f3473da52ec74f3955cadd58eb016537497d44/addons/web/static/src/views/fields/signature/signature_field.js#L120), but it was never declared so it was causing an error.

---

### DIFF
--- a/addons/web/static/tests/views/fields/signature_field.test.js
+++ b/addons/web/static/tests/views/fields/signature_field.test.js
@@ -1,6 +1,6 @@
 import { NameAndSignature } from "@web/core/signature/name_and_signature";
 
-import { expect, test } from "@odoo/hoot";
+import { expect, queryOne, test } from "@odoo/hoot";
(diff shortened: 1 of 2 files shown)


addons/web/static/tests/views/fields/signature_field.test.js
```diff
@@ -342,3 +342,22 @@ test("signature field should render initials", async () => {
     });
     expect.verifySteps(["V.B."]);
 });
+
+test("error loading url", async () => {
+    Partner._records = [{
+        id: 1,
+        sign: "1 kb",
+    }]git
```
kebeclibre: small mistake here

LauraCForgeFlow: My bad! It's already fixed @kebeclibre.

PR #248603, opened 2026-02-13
[IMP] point_of_sale: added reference number to KOT for retail pos

shag-odoo:
Following this commit:
====
- Product reference number is been added to retail pos

task-5920462
Related PR : https://github.com/odoo/enterprise/pull/107108

### DIFF
--- a/addons/point_of_sale/static/src/app/models/utils/order_change.js
+++ b/addons/point_of_sale/static/src/app/models/utils/order_change.js
@@ -144,7 +144,9 @@ export const getOrderChanges = (order, orderPreparationCategories) => {
             const lineDetails = {
                 uuid: orderline.uuid,
                 name: orderline.getFullProductName(),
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/static/src/app/models/utils/order_change.js
```diff
@@ -140,11 +140,12 @@ export const getOrderChanges = (order, orderPreparationCategories) => {
                 oldChanges[relatedKey] &&
                 (oldChanges[relatedKey].note !== note ||
                     oldChanges[relatedKey].customer_note !== customerNote);
-
```
paso-odoo, line 143: unnecessary change

paso-odoo (changes requested):
If possible write hoot for the `getOrderChanges`

davidmonnom (approved):
@robodoo r+

PR #248605, opened 2026-02-13
[FIX] project_timesheet_holidays: fix timesheet generation for flexible employee

bona-odoo:
Steps to reproduce:
-------------------
1. Install Timesheets and Time Off.
2. Create a new Time Off type with duration type "Half Day"
(disable "Require allocation" to skip allocations).
3. Enable "Timesheets" in Timesheets > Configuration.
4. Ensure the employee is configured with a "Flexible" working schedule.
5. Create a multi-day Time Off request (2 days) using this type and approve it.
6. Check the timesheet entries for the employee.

Issue:
------
When a multi-day time off request is submitted for a "Half Day" time off type,
the system fails to generate individual daily timesheet entries. Instead, it generates
a single timesheet entry with a duration calculated as calendar.hours_per_day / 2,
regardless of the actual number of days requested.

(e.g., if hours_per_day is 8, it generates one entry of 4 hours for a 2-day leave),

Cause:
------
After this commit 9512832, multi-day requests were allowed for time off types
where the **request_unit** is "half day" or "hours". However, the logic in `_generate_timesheets`
for flexible calendars was not updated to account for this.

Previously, `request_unit_half` was computed as:
https://github.com/odoo/odoo/blob/feb3cd520d7a736996621f7765184edca6ac390c/addons/hr_holidays/models/hr_leave.py#L319-L323

But now, the condition is:
https://github.com/odoo/odoo/blob/14ccd1f54e21c07ac68593c8dec7682371f8663c/addons/hr_holidays/models/hr_leave.py#L467-L470

Since `request_unit_half` now stays **True** regardless of the request duration
(as long as the Type is 'half_day'), the following condition in `_generate_timesheets`:
https://github.com/odoo/odoo/blob/14ccd1f54e21c07ac68593c8dec7682371f8663c/addons/project_timesheet_holidays/models/hr_leave.py#L39 always evaluates to **True**. This traps multi-day requests in the
"single-day" logic, preventing the code from reaching `_list_work_time_per_day`.

Solution:
---------
Update the condition to strictly check if the request is a single day before applying
the simplified single-entry logic. If the request spans multiple days, it will now
correctly fall into the else block, utilizing `_list_work_time_per_day` to generate
one timesheet entry per day of the leave.

opw-5715197

---

### DIFF
--- a/addons/project_timesheet_holidays/tests/test_timesheet_holidays.py
+++ b/addons/project_timesheet_holidays/tests/test_timesheet_holidays.py
@@ -54,6 +54,12 @@ class TestTimesheetHolidays(TestCommonTimesheet):
             'time_type': 'other',
         })
 
+        self.hr_leave_type_half_day = self.env['hr.leave.type'].sudo().create({
+            'name': 'Time Off Type (half day)',
+            'requires_allocation': False,
+            'request_unit': 'half_day',
+        })
+
         # HR Officer allocates some leaves to the employee 1
         self.Requests = self.env['hr.leave'].with_context(mail_create_nolog=True, mail_notrack=True)
         self.Allocations = self.env['hr.leave.allocation'].with_context(mail_create_nolog=True, mail_notrack=True)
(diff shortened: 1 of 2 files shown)


addons/project_timesheet_holidays/tests/test_timesheet_holidays.py
```diff
@@ -368,15 +374,47 @@ def test_one_day_timesheet_timeoff_flexible_employee(self):
...
+            }]
+        )
+
+        validation_data = [
+            (time_off, 1, 8.0),
+            (time_off_half_day, 1, 4.0),
+            (time_off_multi_day, 2, 16.0),
+        ]
+        for leave, expected_entries, expected_hours in validation_data:
+            leave.with_user(SUPERUSER_ID)._action_validate()
```
alsh-odoo: Instead of validating the leave one by one, you can do that in batch
`(time_off|time_off_half_day|timeoff_multi_day).action_approve()`

alsh-odoo (changes requested):

alsh-odoo (approved):

addons/project_timesheet_holidays/tests/test_timesheet_holidays.py
```diff
@@ -368,15 +374,47 @@ def test_one_day_timesheet_timeoff_flexible_employee(self):
                                                         "should be 10 hours")
 
     def test_timeoff_validation_fully_flexible_employee(self):
```
yoba-odoo, line 389: I am not sure I get why are we removing the old test? Seems like the old test was testing a flow for the fully flexible calendars and your test is testing a flow for flexible employees. 

So I guess we can keep the old test and you add a new one for your flow.

yoba-odoo (commented):
Thank you for your work!!

The fix LGTM but I left a comment regarding the test if you can explain more further please?

yoba-odoo (approved):
Just rebuild runbot and you are good to go 😉 🚀

xavierbol (commented):
LGTM, thanks for your work!

robodoo r+

bona-odoo:
robodoo retry

kitan191:
@robodoo delegate+

PR #248607, opened 2026-02-13
[IMP] account: try country_code to override the original one from base

jco-odoo:
We override, original country_code from base
from `country_id.code` to `account_fiscal_country_id.code`
because most of the fields are visible based on account_fiscal_country_id
in `account` and `l10n` modules. Which makes more sense the field should
be moreover visible on my fiscal country rather than my company country_id

for account_peppol and l10n_es_pos-
we fix the test cases

Possible effect after this commit technically, modules
not using account but using country_code should create
a new company during the test rather than using the existing one

task-5454775

Co-authored-by: Harsh Modi <hamo@odoo.com>

---

### DIFF
--- a/addons/account/models/account_journal.py
+++ b/addons/account/models/account_journal.py
@@ -165,7 +165,7 @@ class AccountJournal(models.Model):
     currency_id = fields.Many2one('res.currency', help='The currency used to enter statement', string="Currency")
     company_id = fields.Many2one('res.company', string='Company', required=True, readonly=True, index=True, default=lambda self: self.env.company,
         help="Company related to this journal")
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code', readonly=True)
+    country_code = fields.Char(related='company_id.country_code', readonly=True)
     account_fiscal_country_group_codes = fields.Json(related="company_id.account_fiscal_country_group_codes")
 
     refund_sequence = fields.Boolean(
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -335,7 +335,7 @@ class AccountMove(models.Model):
     made_sequence_gap = fields.Boolean(compute='_compute_made_sequence_gap', store=True)  # store wether this is the first move breaking the natural sequencing
     show_name_warning = fields.Boolean(store=False)
     type_name = fields.Char('Type Name', compute='_compute_type_name')
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code', readonly=True, depends=['company_id'])
+    country_code = fields.Char(related='company_id.country_code', readonly=True)
     account_fiscal_country_group_codes = fields.Json(related="company_id.account_fiscal_country_group_codes")
     company_price_include = fields.Selection(related='company_id.account_price_include', readonly=True)
     attachment_ids = fields.One2many('ir.attachment', 'res_id', domain=[('res_model', '=', 'account.move')], string='Attachments')
--- a/addons/account/models/account_payment.py
+++ b/addons/account/models/account_payment.py
@@ -197,7 +197,7 @@ class AccountPayment(models.Model):
     # used to know whether the field `partner_bank_id` needs to be required or not in the payments form views
     require_partner_bank_account = fields.Boolean(
         compute='_compute_show_require_partner_bank')
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code')
+    country_code = fields.Char(related='company_id.country_code', depends=['company_id'])
     amount_signed = fields.Monetary(
         currency_field='currency_id', compute='_compute_amount_signed', tracking=True,
         help='Negative value of amount field if payment_type is outbound')
--- a/addons/account/models/company.py
+++ b/addons/account/models/company.py
@@ -222,6 +222,8 @@ class ResCompany(models.Model):
         readonly=False,
         help="The country to use the tax reports from for this company")
     account_fiscal_country_group_codes = fields.Json(compute="_compute_account_fiscal_country_group_codes")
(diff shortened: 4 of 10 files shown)


jco-odoo:
@hamo-odoo Maybe I lost a commit of yours?

hamo-odoo:
@jco-odoo Now we have to fix the test cases 🫠

addons/account/models/account_journal.py
```diff
@@ -161,7 +161,7 @@ def _get_default_account_domain(self):
     currency_id = fields.Many2one('res.currency', help='The currency used to enter statement', string="Currency")
     company_id = fields.Many2one('res.company', string='Company', required=True, readonly=True, index=True, default=lambda self: self.env.company,
         help="Company related to this journal")
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code', readonly=True)
+    country_code = fields.Char(related='company_id.country_code', depends=['company_id'], readonly=True)
```
hamo-odoo: @jco-odoo As per my investigation on the failing test, we should do it as compute instead. there are few standalone modules that depends on the company.country_code and fiscal_country_id is only available when account is installed...so a compute seems much better here. What do you think?

hamo-odoo: i.e. sign_emsigner is only for India (and in account we override the field) but sign module does works as standalone and we can't set the fiscal_country_id in that particular module

hamo-odoo: ahh made the comment at the wrong line 🤦🏻

jco-odoo: @hamo-odoo 
-  Hmm, the idea we had was to indeed have the base country_code as the country_id.code and override it in account.  I do not see how a compute will help in that. There is no way to see what goes wrong in the test exactly?

jco-odoo: Normally account_fiscal_country_id should be equal to country_id in those cases.

hamo-odoo (approved):

jco-odoo:
What is still to check for this before passing to @RicGR98 :
- Payroll in enterprise: if they agree on the change in the tests. 
- We could change it everywhere to fiscal_country_code (but then that is everywhere, everywhere from base up and maybe not only for res_company, but also account_move...), but then also do a double check if in certain cases they would not need the regular country_id.code
- E.g. in D&B we might still need the company.country_id.code instead of the fiscal one. That should be checked.

addons/account/models/account_journal.py
RicGR98: I'm actually wondering if we really need the depends now since it's a simple dependency and not nested like before. If yes, why not in `addons/account/models/account_bank_statement_line.py` ?

jco-odoo: It is non-stored and you do not change the country_id of the company on the fly normally, so maybe we can just do without?

addons/account/models/company.py
```diff
@@ -216,6 +216,8 @@ class ResCompany(models.Model):
         readonly=False,
         help="The country to use the tax reports from for this company")
     account_fiscal_country_group_codes = fields.Json(compute="_compute_account_fiscal_country_group_codes")
+    # OVERRIDE country_code
+    country_code = fields.Char(related='account_fiscal_country_id.code', depends=['account_fiscal_country_id'])
```
jco-odoo, line 226:
```suggestion
    country_code = fields.Char(related='account_fiscal_country_id.code')
```
Still seems to work in the view.   But it is probably because it is there in base.

jco-odoo: Seems indeed logical that if we have it in base, we keep it here as well.

addons/account/models/account_payment.py
```diff
@@ -196,7 +196,7 @@ class AccountPayment(models.Model):
     # used to know whether the field `partner_bank_id` needs to be required or not in the payments form views
     require_partner_bank_account = fields.Boolean(
         compute='_compute_show_require_partner_bank')
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code')
+    country_code = fields.Char(related='company_id.country_code', depends=['company_id'])
```
RicGR98, line 200: Depends needed here ?

hamo-odoo: Thanks, I might have missed this one, while changing the others :D

addons/account/models/account_move.py
```diff
@@ -325,7 +325,7 @@ def _sequence_year_range_monthly_regex(self):
     made_sequence_gap = fields.Boolean(compute='_compute_made_sequence_gap', store=True)  # store wether this is the first move breaking the natural sequencing
     show_name_warning = fields.Boolean(store=False)
     type_name = fields.Char('Type Name', compute='_compute_type_name')
-    country_code = fields.Char(related='company_id.account_fiscal_country_id.code', readonly=True, depends=['company_id'])
+    country_code = fields.Char(related='company_id.country_code', readonly=True)
```
jco-odoo, line 338: In a way, it is true that here we do not need the depends as the company is not changed after, although it could be, but only to a child company, which could have the same fiscal country.  I am testing if we can do something the domain of the company_id field however.

jco-odoo: https://github.com/odoo/odoo/pull/254189

hamo-odoo: @jco-odoo I guess nothing needs to be changed here, then?

RicGR98 (approved):
@jco-odoo 

It's okay for me, so if it's okay for you too we can go ahead :)

@robodoo delegate+

I'll let you r+ the Enterprise one

addons/account/models/account_move.py
jco-odoo: indeed

PR #248619, opened 2026-02-13
[FIX] html_editor: misc URL autoconvert fixes

bso-odoo:
[FIX] html_editor: undo link autoconvert before space insertion

When adding a space after an URL, the URL text is converted to an URL.
When pressing undo, first the space is undone, then the link is undone.
This is wrong because if the user did not want a link, after undoing the
link, inserting a new space will again convert to a link.

This commit splits `handleAutomaticLinkInsertion` into two parts:
determining if a link must be created, and actually inserting the link.
This makes it possible to execute code within the condition before and
after the insertion.
To fix the similar behavior for enter and shift-enter, another before
input handler is also added in order to let the default before input be
executed before creating the link.

Steps to reproduce:
- Go to a "To do" note
- Type "odoo.com"
- Press space/enter/shift-enter
- Undo a single time

=> The insertion was undone instead of the link transform.

task-5936310

[FIX] html_editor: not include surrounding backtick in pasted URL

When pasting an URL surrounded by backticks, the ending backtick is included inside the link's HREF.

This commit fixes the regex for URL to also exclude backticks (like it did with `"` and `'`).

Steps to reproduce:
- Copy the following text in the clipboard:
```
`odoo.com`
```
- Go to a "To do" note
- Paste

=> The link's URL was
```
odoo.com`
```

task-5936310

### DIFF
--- a/addons/html_editor/static/src/main/link/link_plugin.js
+++ b/addons/html_editor/static/src/main/link/link_plugin.js
@@ -231,8 +231,15 @@ export class LinkPlugin extends Plugin {
         power_buttons: { commandId: "toggleLinkTools" },
 
         /** Handlers */
-        beforeinput_handlers: withSequence(5, this.onBeforeInput.bind(this)),
-        input_handlers: this.onInputDeleteNormalizeLink.bind(this),
+        beforeinput_handlers: [
+            withSequence(5, this.onBeforeInputPrepareConvertToLink.bind(this)),
+            this.onBeforeInputFixFirefoxSelection.bind(this),
+            withSequence(20, this.onBeforeInputConvertToLink.bind(this)),
+        ],
+        input_handlers: [
+            this.onInputDeleteNormalizeLink.bind(this),
+            this.onInputConvertToLink.bind(this),
+        ],
         before_delete_handlers: this.updateCurrentLinkSyncState.bind(this),
         delete_handlers: this.onInputDeleteNormalizeLink.bind(this),
         before_paste_handlers: this.updateCurrentLinkSyncState.bind(this),
@@ -813,36 +820,17 @@ export class LinkPlugin extends Plugin {
         }
     }
 
-    onBeforeInput(ev) {
-        if (ev.inputType === "insertParagraph" || ev.inputType === "insertLineBreak") {
-            const nodeForSelectionRestore = this.handleAutomaticLinkInsertion();
-            if (nodeForSelectionRestore) {
-                this.dependencies.selection.setCursorStart(nodeForSelectionRestore);
-                this.dependencies.history.addStep();
-            }
+    onBeforeInputPrepareConvertToLink(ev) {
+        if (
+            ev.inputType === "insertParagraph" ||
+            ev.inputType === "insertLineBreak" ||
+            (ev.inputType === "insertText" && ev.data === " ")
+        ) {
+            this.convertToLink = this.prepareConvertToLink();
         }
-        if (ev.inputType === "insertText" && ev.data === " ") {
-            const nodeForSelectionRestore = this.handleAutomaticLinkInsertion();
-            if (nodeForSelectionRestore) {
-                // Since we manually insert a space here, we will be adding a history step
-                // after link creation with selection at the end of the link and another
-                // after inserting the space. So first undo will remove the space, and the
-                // second will undo the link creation.
-                this.dependencies.selection.setSelection({
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 0,
-                });
-                this.dependencies.history.addStep();
-                nodeForSelectionRestore.textContent =
-                    "\u00A0" + nodeForSelectionRestore.textContent;
-                this.dependencies.selection.setSelection({
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 1,
-                });
-                this.dependencies.history.addStep();
-                ev.preventDefault();
-            }
-        }
+    }
+
+    onBeforeInputFixFirefoxSelection(ev) {
         // Firefox: avoid corrupted selection inside link.
         const selection = this.document.getSelection();
         if (
@@ -857,6 +845,26 @@ export class LinkPlugin extends Plugin {
             selection.collapse(selection.anchorNode, offset);
         }
         this.updateCurrentLinkSyncState();
+    }
+
+    onBeforeInputConvertToLink(ev) {
+        if (this.convertToLink) {
+            if (ev.inputType === "insertParagraph" || ev.inputType === "insertLineBreak") {
+                this.convertToLink();
+                this.dependencies.history.addStep();
+                delete this.convertToLink;
+            }
+        }
+    }
+
+    onInputConvertToLink(ev) {
+        if (this.convertToLink) {
+            if (ev.inputType === "insertText" && ev.data === " ") {
+                this.convertToLink();
+                this.dependencies.history.addStep();
+                delete this.convertToLink;
+            }
+        }
     }
 
     onInputDeleteNormalizeLink() {
@@ -894,8 +902,23 @@ export class LinkPlugin extends Plugin {
     /**
      * Inserts a link in the editor. Called after pressing space or (shif +) enter.
      * Performs a regex check to determine if the url has correct syntax.
+     * TODO Delete in master: kept for stable
      */
     handleAutomaticLinkInsertion() {
+        const convertToLink = this.prepareConvertToLink();
+        if (convertToLink) {
+            return convertToLink();
+        }
+    }
+
+    /**
+     * If a link must be inserted after pressing space or (shift +) enter,
+     * returns information that will be needed to insert the link.
+     * Performs a regex check to determine if the url has correct syntax.
+     *
+     * @returns {Function} that performs the link conversion
+     */
+    prepareConvertToLink() {
         let selection = this.dependencies.selection.getEditableSelection();
         if (
             isHtmlContentSupported(selection.anchorNode) &&
@@ -923,12 +946,16 @@ export class LinkPlugin extends Plugin {
                     startOffset,
                     startOffset + match[0].length
                 );
-                const link = this.createLink(url, text);
                 // split the text node and replace the url text with the link
                 const textNodeToReplace = selection.anchorNode.splitText(startOffset);
                 textNodeToReplace.splitText(match[0].length);
-                selection.anchorNode.parentElement.replaceChild(link, textNodeToReplace);
-                return nodeForSelectionRestore;
+                const link = this.createLink(url, text);
+                return () => {
+                    textNodeToReplace.splitText(match[0].length); // this will keep the space (that will have been added in the meantime)
+                    textNodeToReplace.parentNode.replaceChild(link, textNodeToReplace);
+                    // TODO Remove in master: kept for stable
+                    return nodeForSelectionRestore;
+                };
             }
         }
     }
--- a/addons/html_editor/static/tests/_helpers/user_actions.js
+++ b/addons/html_editor/static/tests/_helpers/user_actions.js
@@ -72,6 +72,47 @@ export async function insertText(editor, text) {
     }
 }
 
+/**
+ * @param {Editor} editor
+ */
+export async function insertSpace(editor) {
+    const keydownEvent = await manuallyDispatchProgrammaticEvent(editor.editable, "keydown", {
+        key: " ",
+    });
+    if (keydownEvent.defaultPrevented) {
+        return;
+    }
+    // InputEvent is required to simulate the insert text.
+    const [beforeinputEvent] = await manuallyDispatchProgrammaticEvent(
+        editor.editable,
+        "beforeinput",
+        {
+            inputType: "insertText",
+            data: " ",
+        }
+    );
+    if (beforeinputEvent.defaultPrevented) {
+        return;
+    }
+    const range = editor.document.getSelection().getRangeAt(0);
+    if (!range.collapsed) {
+        throw new Error("need to implement something... maybe");
+    }
+
+    // mimic the behavior of the browser when inserting a &nbsp
+    document.execCommand("insertText", false, " ");
+
+    const [inputEvent] = await manuallyDispatchProgrammaticEvent(editor.editable, "input", {
+        inputType: "insertText",
+        data: " ",
+    });
+    if (inputEvent.defaultPrevented) {
+        return;
+    }
+    // KeyUpEvent is not required but is triggered like the browser would.
+    await manuallyDispatchProgrammaticEvent(editor.editable, "keyup", { key: " " });
+}
+
 /** @param {Editor} editor */
 export function deleteForward(editor) {
     execCommand(editor, "deleteForward");
--- a/addons/html_editor/static/tests/link/popover.test.js
+++ b/addons/html_editor/static/tests/link/popover.test.js
@@ -17,7 +17,13 @@ import { contains, onRpc, patchWithCleanup } from "@web/../tests/web_test_helper
 import { setupEditor } from "../_helpers/editor";
 import { cleanLinkArtifacts } from "../_helpers/format";
 import { getContent, setContent, setSelection } from "../_helpers/selection";
-import { insertLineBreak, insertText, splitBlock, undo } from "../_helpers/user_actions";
+import {
+    insertLineBreak,
+    insertSpace,
+    insertText,
+    splitBlock,
+    undo,
+} from "../_helpers/user_actions";
 import { execCommand } from "../_helpers/userCommands";
 import { expectElementCount } from "../_helpers/ui_expectations";
 
@@ -249,7 +255,7 @@ describe("Link creation", () => {
         test("typing valid URL + space should convert to link", async () => {
             const { editor, el } = await setupEditor("<p>[]</p>");
             await insertText(editor, "http://google.co.in");
-            await insertText(editor, " ");
+            await insertSpace(editor);
             expect(cleanLinkArtifacts(getContent(el))).toBe(
                 '<p><a href="http://google.co.in">http://google.co.in</a>&nbsp;[]</p>'
             );
--- a/addons/html_editor/static/tests/utils/user_actions.test.js
+++ b/addons/html_editor/static/tests/utils/user_actions.test.js
@@ -1,7 +1,7 @@
 import { describe, expect, test } from "@odoo/hoot";
 import { setupEditor } from "../_helpers/editor";
 import { getContent } from "../_helpers/selection";
-import { simulateArrowKeyPress } from "../_helpers/user_actions";
+import { insertSpace, simulateArrowKeyPress } from "../_helpers/user_actions";
 
 describe("simulateArrowKeyPress method", () => {
     describe("move", () => {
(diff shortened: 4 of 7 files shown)


addons/html_editor/static/tests/link/popover.test.js
```diff
@@ -251,7 +251,7 @@ describe("Link creation", () => {
             await insertText(editor, "http://google.co.in");
             await insertText(editor, " ");
             expect(cleanLinkArtifacts(getContent(el))).toBe(
-                '<p><a href="http://google.co.in">http://google.co.in</a>&nbsp;[]</p>'
+                '<p><a href="http://google.co.in">http://google.co.in</a> []</p>'
```
dmo-odoo: This is wrong. A space at that position would not be rendered by the browser if it isn't a `&nbsp;`. Same for most of the other occurences below.

bso-odoo: @dmo-odoo Restored initial test results by introducing before & after callbacks in the method that creates links.
I used `defaultPrevented` to avoid that the default split and line break are executed, but in master I believe we should think about introducing a "stop propagation" mechanism.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -895,7 +908,7 @@ export class LinkPlugin extends Plugin {
      * Inserts a link in the editor. Called after pressing space or (shif +) enter.
      * Performs a regex check to determine if the url has correct syntax.
      */
-    handleAutomaticLinkInsertion() {
+    handleAutomaticLinkInsertion({ callbackBefore, callbackAfter }) {
```
dmo-odoo: EDIT: I actually got a better idea than this one, see other comment.

Original comment before edit:
Rather than introducing these callbacks, I would have inlined the code of the callbacks inside the function like this:

```diff
diff --git a/addons/html_editor/static/src/main/link/link_plugin.js b/addons/html_editor/static/src/main/link/link_plugin.js
index 97a761f39d80..2fed7d7d3e12 100644
--- a/addons/html_editor/static/src/main/link/link_plugin.js
+++ b/addons/html_editor/static/src/main/link/link_plugin.js
@@ -1084,35 +1084,7 @@ export class LinkPlugin extends Plugin {
     }
 
     onBeforeInput(ev) {
-        if (ev.inputType === "insertParagraph" || ev.inputType === "insertLineBreak") {
-            const nodeForSelectionRestore = this.handleAutomaticLinkInsertion();
-            if (nodeForSelectionRestore) {
-                this.dependencies.selection.setCursorStart(nodeForSelectionRestore);
-                this.dependencies.history.addStep();
-            }
-        }
-        if (ev.inputType === "insertText" && ev.data === " ") {
-            const nodeForSelectionRestore = this.handleAutomaticLinkInsertion();
-            if (nodeForSelectionRestore) {
-                // Since we manually insert a space here, we will be adding a history step
-                // after link creation with selection at the end of the link and another
-                // after inserting the space. So first undo will remove the space, and the
-                // second will undo the link creation.
-                this.dependencies.selection.setSelection({
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 0,
-                });
-                this.dependencies.history.addStep();
-                nodeForSelectionRestore.textContent =
-                    "\u00A0" + nodeForSelectionRestore.textContent;
-                this.dependencies.selection.setSelection({
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 1,
-                });
-                this.dependencies.history.addStep();
-                ev.preventDefault();
-            }
-        }
+        this.handleAutomaticLinkInsertion(ev);
         // Firefox: avoid corrupted selection inside link.
         const selection = this.document.getSelection();
         if (
@@ -1169,8 +1141,18 @@ export class LinkPlugin extends Plugin {
     /**
      * Inserts a link in the editor. Called after pressing space or (shif +) enter.
      * Performs a regex check to determine if the url has correct syntax.
+     *
+     * @param {InputEvent} ev
      */
-    handleAutomaticLinkInsertion() {
+    handleAutomaticLinkInsertion(ev) {
+        if (
+            ev.inputType !== "insertParagraph" &&
+            ev.inputType !== "insertLineBreak" &&
+            !(ev.inputType === "insertText" && ev.data === " ")
+        ) {
+            return;
+        }
+        ev.preventDefault();
         let selection = this.dependencies.selection.getEditableSelection();
         if (
             isHtmlContentSupported(selection) &&
@@ -1187,6 +1169,29 @@ export class LinkPlugin extends Plugin {
             const match = [...potentialUrl.matchAll(new RegExp(URL_REGEX, "g"))].pop();
 
             if (match && !EMAIL_REGEX.test(match[0])) {
+                if (ev.inputType === "insertText") {
+                    const text = selection.anchorNode.textContent;
+                    selection.anchorNode.textContent =
+                        text.substring(0, selection.anchorOffset) +
+                        "\u00A0" +
+                        text.substring(selection.anchorOffset);
+                    this.dependencies.selection.setSelection({
+                        anchorNode: selection.anchorNode,
+                        anchorOffset: selection.anchorOffset + 1,
+                    });
+                    this.dependencies.history.addStep();
+                } else {
+                    if (ev.inputType === "insertParagraph") {
+                        this.dependencies.split.splitBlock();
+                        this.dependencies.history.addStep();
+                    }
+                    if (ev.inputType === "insertLineBreak") {
+                        this.dependencies.lineBreak.insertLineBreak();
+                        // addStep included
+                    }
+                    this.dependencies.selection.modifySelection("move", "backward", "character");
+                    selection = this.dependencies.selection.getEditableSelection();
+                }
                 const nodeForSelectionRestore = selection.anchorNode.splitText(
                     selection.anchorOffset
                 );
@@ -1204,7 +1209,11 @@ export class LinkPlugin extends Plugin {
                 if (link.getAttribute("href") === link.textContent) {
                     this.newlyInsertedLinks.add(link);
                 }
-                return nodeForSelectionRestore;
+                if (ev.inputType !== "insertText") {
+                    this.dependencies.selection.setCursorStart(nodeForSelectionRestore);
+                    this.dependencies.selection.modifySelection("move", "forward", "character");
+                }
+                this.dependencies.history.addStep();
             }
         }
     }

```

What do you think?

addons/html_editor/static/tests/link/popover.test.js
dmo-odoo: I'm not sure there is a valid case for that outside of handlers pertaining to DOM events. Now that I think about it, I think that's actually the root cause of the bug here. We've been reimplementing `insertParagraph` and `insertLinebreak` in `LinkPlugin` just so we can check the text nodes before it just in case. We could do it with two handlers. One with a low sequence to check if the preceding text nodes are a url and which would set a variable on this like `this.urlTextNodeBeforeInput`, and the other with a high sequence to check if it's the right events (insertText space, insertParagraph, insertLinebreak) and if `this.urlTextNodeBeforeInput` is set. If it's the case then it can wrap the text node inside a `a` tag and be done with it. The order of changes is correct by construction and we can get rid of the reimplementation of those three behaviors in `LinkPlugin`.

bso-odoo: @dmo-odoo Adapted accordingly - thus letting the default insert paragraph and line break proceed.

dmo-odoo: Thanks, but the code does not correspond to what I said... ^^' I meant this:

1. LinkPlugin, onbeforeinput, low sequence: do the part that splits the text node to match a url, store the reference to the TextNode on `this` (and maybe the url too so that we don't have to re-match it)
2. Other plugins run their course, adds a space (to the second part of the text node that was split, so it doesn't affect the one that we stored) or adds a paragraph (doesn't affect the text node that we stored either)
3. LinkPlugin, onbeforeinput, high sequence: check if we stored a textnode (and url), if yes, wrap it in an a with the url

EDIT: I realize your code in fact does _almost_ that. What I didn't get is the `getAutomaticLinkInserter` function. The main difference is you don't split the node before, is there a particular reason not to do so? If I can avoid functions that return functions I'd be happier to be honest, I always found it a bit hard to follow and this case here is a prime example. If there is a real reason not to split the text node beforehand then I can also live with it (I'll just need to find a better name for the function though).

I'm also confused by the fact that we still insert the space ourselves. Why? Why not let the browser do it and recover `this.insertLink` like in the other case? I feel like it would be simpler rather than having two different cases, one where the link insertion is deffered and one where it is not.

bso-odoo: When using the default space insertion, it leads to the tests diff you highlighted in the first review:
```diff
-                '<p><a href="http://google.co.in">http://google.co.in</a>&nbsp;[]</p>'
+                '<p><a href="http://google.co.in">http://google.co.in</a> []</p>'
```
I will look into splitting the node before.

bso-odoo: Splitting before is ok for paragraph & line break.
Having the split happen before makes the whole text node - where the space was inserted - be replaced by a link within the space after it. (Similar with manual space insertion and default one)
```
1: odoo.com[]
2: odoo.com [] <= space within text node
3: <a href="odoo.com">odoo.com</a>[] <= no more space
```

dmo-odoo: We add the space manually anyway so we can add it in the second text node.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -923,15 +951,17 @@ export class LinkPlugin extends Plugin {
...
                 textNodeToReplace.splitText(match[0].length);
-                selection.anchorNode.parentElement.replaceChild(link, textNodeToReplace);
-                return nodeForSelectionRestore;
+                return { nodeForSelectionRestore, url, text, textNodeToReplace };
             }
         }
     }
+    replaceWithLink(selection, { url, text, textNodeToReplace }) {
+        const link = this.createLink(url, text);
+        selection.anchorNode.parentElement.replaceChild(link, textNodeToReplace);
```
dmo-odoo: Unless I'm mistaken, we don't need to pass the selection if we already have textNodeToReplace. We can do textNodeToReplace.parentNode.replaceChild(link, textNodeToReplace). If it works, we might not need to move the selection at all in https://github.com/odoo/odoo/pull/248619/changes/65d54cec59ce83d07ac4f2f27873da6088389110#diff-87d28773377778ad467903551beacdb07444ee2822684ba10c45a1bcff51f417R859, so no need for `nodeForSelectionRestore` either. At that point we might as well have `prepareAutomaticLinkInsertion` call `this.createLink` and return the `link` itself, or even do the replace in a callback. I'm sorry I know it was already in a callback before and I asked you to change it, but this time it's a bit different because it would only return a small callback with the `replaceChild`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
 
     onBeforeInput(ev) {
         if (ev.inputType === "insertParagraph" || ev.inputType === "insertLineBreak") {
-            const nodeForSelectionRestore = this.handleAutomaticLinkInsertion();
-            if (nodeForSelectionRestore) {
-                this.dependencies.selection.setCursorStart(nodeForSelectionRestore);
-                this.dependencies.history.addStep();
-            }
+            this.linkInsertion = this.prepareAutomaticLinkInsertion();
```
dmo-odoo: Rename to `this.convertToLink`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
...
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 0,
-                });
-                this.dependencies.history.addStep();
-                nodeForSelectionRestore.textContent =
-                    "\u00A0" + nodeForSelectionRestore.textContent;
+            const linkInsertion = this.prepareAutomaticLinkInsertion();
+            if (linkInsertion) {
+                const selection = this.dependencies.selection.getEditableSelection();
+                const space = this.document.createTextNode("\u00A0");
```
dmo-odoo: That variable name is misleading as it is not any space, it is a non-breaking space. I would call it `nbsp`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
...
-                // after inserting the space. So first undo will remove the space, and the
-                // second will undo the link creation.
-                this.dependencies.selection.setSelection({
-                    anchorNode: nodeForSelectionRestore,
-                    anchorOffset: 0,
-                });
-                this.dependencies.history.addStep();
-                nodeForSelectionRestore.textContent =
-                    "\u00A0" + nodeForSelectionRestore.textContent;
+            const linkInsertion = this.prepareAutomaticLinkInsertion();
```
dmo-odoo: Rename to `convertToLink`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -923,15 +951,17 @@ export class LinkPlugin extends Plugin {
                     startOffset,
                     startOffset + match[0].length
                 );
-                const link = this.createLink(url, text);
                 // split the text node and replace the url text with the link
                 const textNodeToReplace = selection.anchorNode.splitText(startOffset);
                 textNodeToReplace.splitText(match[0].length);
-                selection.anchorNode.parentElement.replaceChild(link, textNodeToReplace);
-                return nodeForSelectionRestore;
+                return { nodeForSelectionRestore, url, text, textNodeToReplace };
```
dmo-odoo:
```js
return () => {
    textNodeToReplace.splitText(match[0].length); // this will keep the space (that will have been added in the meantime)
    textNodeToReplace.parentNode.replaceChild(link, textNodeToReplace);
}
```

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -894,18 +905,35 @@ export class LinkPlugin extends Plugin {
...
+    }
+
+    /**
+     * If a link must be inserted after pressing space or (shift +) enter,
+     * returns information that will be needed to insert the link.
+     * Performs a regex check to determine if the url has correct syntax.
+     *
+     * @returns {Object} with nodeForSelectionRestore, url, text and textNodeToReplaces
+     */
+    prepareAutomaticLinkInsertion() {
```
dmo-odoo: Rename to `prepareConvertToLink`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -859,6 +853,23 @@ export class LinkPlugin extends Plugin {
         this.updateCurrentLinkSyncState();
     }
 
+    onBeforeInputAfterOtherPlugins(ev) {
```
dmo-odoo: Rename to `onBeforeInputConvertToLink`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
 
     onBeforeInput(ev) {
```
dmo-odoo: Rename to `onBeforeInputPrepareConvertToLink`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
...
-                    anchorOffset: 0,
-                });
-                this.dependencies.history.addStep();
-                nodeForSelectionRestore.textContent =
-                    "\u00A0" + nodeForSelectionRestore.textContent;
+            const linkInsertion = this.prepareAutomaticLinkInsertion();
+            if (linkInsertion) {
+                const selection = this.dependencies.selection.getEditableSelection();
+                const nbsp = this.document.createTextNode("\u00A0");
+                selection.anchorNode.after(nbsp);
```
dmo-odoo: I had an idea to ensure we use the correct space character. We could delay the creating of the link in case of `insertText` to after the input and delay the split on match length at that point, see the suggested return of `prepareAutomaticLinkInsertion`.

addons/html_editor/static/src/main/link/link_plugin.js
```diff
@@ -815,32 +818,23 @@ export class LinkPlugin extends Plugin {
...
                     anchorOffset: 1,
                 });
                 this.dependencies.history.addStep();
                 ev.preventDefault();
+                this.dependencies.selection.setCursorStart(linkInsertion.nodeForSelectionRestore);
+                this.replaceWithLink(selection, linkInsertion);
+                this.dependencies.history.addStep();
             }
         }
         // Firefox: avoid corrupted selection inside link.
```
dmo-odoo, line 834: `onBeforeInputFixFirefoxSelection`

addons/html_editor/static/tests/utils/user_actions.test.js
```diff
@@ -109,3 +109,31 @@ describe("simulateArrowKeyPress method", () => {
...
+        const { editor, el } = await setupEditor("<p>a[]b</p>");
+        await insertSpace(editor);
+        expect(getContent(el)).toBe("<p>a []b</p>");
+    });
+    test("insert second space", async () => {
+        const { editor, el } = await setupEditor("<p>a []b</p>");
+        await insertSpace(editor);
+        expect(getContent(el)).toBe("<p>a&nbsp; []b</p>");
+    });
+    test("insert space before space", async () => {
```
dmo-odoo, line 124: Isn't it missing "insert space after space" ?

addons/html_editor/static/tests/_helpers/user_actions.js
```diff
@@ -72,6 +72,47 @@ export async function insertText(editor, text) {
     }
 }
 
+/**
+ * @param {Editor} editor
+ */
+export async function insertSpace(editor) {
```
dmo-odoo, line 78: In the commit message, could you replace the last paragraph by the following, please? It was rather tricky to understand why the `prepareUpdate`/`getState` strategy didn't work, so I would rather have it written somewhere.

"Attempts were made to use the `prepareUpdate` and/or `getState` functions but to no avail. Those functions are meant to react to a change that happened around existing spaces in order to fix their visibility, but in this case the observed change would be the space itself, so the function would misbehave. This case was never taken into account when those functions were designed and is impossible to interweave properly into the existing code without rewriting them entirely.

In practice, those cases are handled by the native behavior of the browser when inserting a space through the `contenteditable`. This commit emulates that behavior for the test utils by using `execCommand`."

addons/html_editor/static/tests/_helpers/user_actions.js
```diff
@@ -72,6 +72,47 @@ export async function insertText(editor, text) {
...
+    if (beforeinputEvent.defaultPrevented) {
+        return;
+    }
+    const range = editor.document.getSelection().getRangeAt(0);
+    if (!range.collapsed) {
+        throw new Error("need to implement something... maybe");
+    }
+
+    // mimic the behavior of the browser when inserting a &nbsp
+    document.execCommand("insertText", false, " ");
```
bso-odoo, line 103: @dmo-odoo Ultimately did use the anti-climactic `execCommand`... I've noted the fact that other approaches were tried in the commit message. Runbot is green.

addons/html_editor/static/tests/utils/user_actions.test.js
bso-odoo: It's not missing, but I renamed the previous test which was doing that to make is more clear.

PR #248620, opened 2026-02-13
[FIX] product: fix Arabic alignment in Product Catalog

amahm-odoo:
Steps to reproduce:
1. Switch the user language to Arabic.
2. Open the Product Catalog from a Sales/Purchase order.
3. Observe that the Name, Internal Reference, and Price collapse into a single line with inconsistent ordering due to unmanaged horizontal inline flow.

Cause:
The use of generic span elements inside a portal encourages horizontal inline flow that fails to mirror correctly in RTL without explicit flex instructions.

Solution:
Update the Order Line portal template to use 'd-inline-flex' and 'align-items-baseline'. This treats the price and reference as a logical unit that respects the global direction and ensures consistent text alignment in both English and Arabic.

opw-5867434

### DIFF
--- a/addons/product/static/src/product_catalog/order_line/order_line.xml
+++ b/addons/product/static/src/product_catalog/order_line/order_line.xml
@@ -4,8 +4,10 @@
         <!-- Replace the element found using the css selector by the content of the portalled
              template.  -->
         <t t-portal="`#product-${props.productId}-price`">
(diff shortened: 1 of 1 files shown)


barracudapps (approved):
LGTM 👍 
Only worked with an SAR currency pricelist for me though.
Thanks for fixing

ltinel (approved):
LGTM, thanks :)
@robodoo delegate+

PR #227477, opened 2025-09-16
[IMP] orm: prevent use of falsy ids

kmagusiak:
Having a falsy real id is a programming mistake.
The following should be unusable `model.browse([False])`.

Add asserts that will check this and remove code from `OriginIds` that supports this case.

odoo/enterprise#95117

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248625, opened 2026-02-13
[FIX] microsoft_calendar: fix synchronization of cancelled events

pedrambiria:
Before this commit, when syncing events from Microsoft, if there was a cancelled event that was not yet synchronized with Odoo, calling browse on event.odoo_id would raise an error as it would be None.

opw-5917677

---

### DIFF
--- a/addons/microsoft_calendar/models/microsoft_sync.py
+++ b/addons/microsoft_calendar/models/microsoft_sync.py
@@ -338,8 +338,8 @@ class MicrosoftSync(models.AbstractModel):
         :return: synchronized odoo
         """
         existing = microsoft_events.match_with_odoo_events(self.env)
-        cancelled = microsoft_events.cancelled()
-        new = microsoft_events - existing - cancelled
+        cancelled = existing.cancelled()
+        new = microsoft_events - existing - microsoft_events.cancelled()
         new_recurrence = new.filter(lambda e: e.is_recurrent())
 
         # create new events and reccurrences


addons/microsoft_calendar/models/microsoft_sync.py
```diff
@@ -338,8 +338,8 @@ def _sync_microsoft2odoo(self, microsoft_events: MicrosoftEvent):
         :return: synchronized odoo
         """
         existing = microsoft_events.match_with_odoo_events(self.env)
-        cancelled = microsoft_events.cancelled()
-        new = microsoft_events - existing - cancelled
+        new = microsoft_events - existing - microsoft_events.cancelled()
+        cancelled = existing.cancelled()
```
reth-odoo: Doesn't seem like we needed to swap them around

reth-odoo (approved):

addons/microsoft_calendar/models/microsoft_sync.py
pedrambiria: Because we shouldn’t do anything if an event was canceled but doesn’t exist.

reth-odoo: I understand that

If you open the full diff, previously we set cancelled and then new, and now we do it in the inverse order.
That swap is not needed.

pedrambiria: Yes not needed, but I wanted to keep the same order as in the `_sync_google2odoo`.

pedrambiria: @reth-odoo Do I need to change the order? I wanted to keep the same order as in the `_sync_google2odoo`.

Odoonan: Hi, I think you can keep the original order, as it does not make a difference, let us have the smallest diff :)

Also, could you give a few steps to reproduce? Side question: is the issue already observed in 17.0? I saw the ticket was in 19.1, hence me asking

Cheers

pedrambiria: @Odoonan Hello,
I have modified the code as requested.

I did not attempt to reproduce the issue from scratch, as I performed my testing directly using the customer's calendar and the reproduction steps are quite specific. To reproduce this, you would need to sync a Microsoft Calendar with Odoo, create an event in Microsoft, and then cancel it before the initial sync to Odoo occurs.

While I did not see the issue in Version 17.0, the codebase is identical; therefore, I have decided to apply the fix starting from 17.0 to prevent future occurrences.

Odoonan: Hi @pedrambiria :wave: 

Thanks for the update. The code looks fine, but I would retarget 19.1, as the error is thrown due to recent work preventing use of falsy ids, which was 'supported' before. As the error was not reported before, I would avoid changes in old stable :) https://github.com/odoo/odoo/pull/227477 

Can you please retarget and include a reference to the commit in the linked PR? Thanks! We should be good to go then :rocket:

PR #248630, opened 2026-02-13
[IMP] mail: add copy suffix when duplicating canned responses

nees-odoo:
Previously, duplicating canned responses did not add any indication that 
the record was duplicated, making it difficult to distinguish 
between original and copied records.

- Adds `(copy)` suffix to the source field when duplicating canned responses.

task-[5873772](https://www.odoo.com/odoo/project/1519/tasks/5873772)

---

### DIFF
--- a/addons/mail/models/mail_canned_response.py
+++ b/addons/mail/models/mail_canned_response.py
@@ -71,6 +71,12 @@ class MailCannedResponse(models.Model):
         self._broadcast(delete=True)
         return super().unlink()
 
+    def copy_data(self, default=None):
+        vals_list = super().copy_data(default=default)
+        if "source" not in (default or {}):
+            return [dict(vals, source=self.env._("%s (copy)", rec.source)) for rec, vals in zip(self, vals_list)]
+        return vals_list
+
     def _broadcast(self, /, *, delete=False):
         for canned_response in self:
             stores = [Store(bus_channel=group) for group in canned_response.group_ids]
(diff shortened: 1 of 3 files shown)


addons/mail/tests/test_mail_canned_response.py
```diff
@@ -0,0 +1,20 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.canned_response = cls.env["mail.canned.response"].create({
+            "source": "hello",
+            "substitution": "Hello, how may I help you?",
+        })
+
+    def test_mail_canned_response_copy(self):
+        copied = self.canned_response.copy()
+        self.assertEqual(copied.source, f"{self.canned_response.source} (copy)")
```
maad-odoo: Since the added `copy_data` override  can handle multiple records at a time, it would be nice if the testcase can also demonstrate the same :)

maad-odoo (approved):

addons/mail/models/mail_canned_response.py
```diff
@@ -71,6 +71,13 @@ def unlink(self):
         self._broadcast(delete=True)
         return super().unlink()
 
+    def copy_data(self, default=None):
+        vals_list = super().copy_data(default=default)
+        if "source" not in (default or {}):
+            for rec, vals in zip(self, vals_list):
+                vals["source"] = self.env._("%s (copy)", rec.source)
+        return vals_list
+
```
alexkuhn, line 79: Could use shorter syntax version like https://github.com/odoo/odoo/blob/19.0/addons/resource/models/resource_resource.py#L81

alexkuhn (approved):
LGTM otherwise.
While it's usually preferred to put placeholder to help translators, e.g. `%(canned_response_source) (copy)`, `%s (copy)` has many occurrences and there's at least 1 existing occurrence in `@mail` that would lead to same and single POT entry.

@robodoo delegate+

PR #248633, opened 2026-02-13
[IMP] mail: cleanup poll votes panel

tsm-odoo:
This PR cleans up the poll votes panel:
- Remove useless option fetch when opening the panel: polls are always sent alongside their options (number of votes and label).
- Rely on `makeCachedFetchData` instead of custom caching.
- Add an empty placeholder.

image

### DIFF
--- a/addons/mail/static/src/core/common/mail_poll_option_model.js
+++ b/addons/mail/static/src/core/common/mail_poll_option_model.js
@@ -4,6 +4,16 @@ import { Record } from "@mail/model/record";
 export class MailPollOptionModel extends Record {
     static _name = "mail.poll.option";
 
+    static new() {
+        const option = super.new(...arguments);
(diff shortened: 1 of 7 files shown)


phenix-factory:
It's quite a strange pattern to have to fetch a poll option to get the poll itself. One would assume that you fetch the poll and it comes with its options.

addons/mail/static/src/core/common/mail_poll_option_model.js
```diff
@@ -16,19 +24,7 @@ export class MailPollOptionModel extends Record {
...
-                    poll_option_id: this.id,
-                });
-            } catch (e) {
-                this.votesFetched = false;
-                throw e;
-            }
-        }
-    }
+    /** @type {ReturnType<import("models").Store['makeCachedFetchData']>} */
+    fetchPollVotesCached;
```
seb-odoo: alphabetical

seb-odoo (approved):
@robodoo delegate+

PR #248634, opened 2026-02-13
[FIX] mrp: provide access to product variants in test

sawer-odoo:
The _/mrp:TestReportBom.test_mrp_report_bom_variant_selection_ test was failling because administrator user couldn't use the product variants. The needed dropdown menu to switch variant wasn't displayed and the test failed.

We add the "product.group_product_variant" rights to the user.

Runbot error: 238917

---

### DIFF
--- a/addons/mrp/tests/test_mrp_reports.py
+++ b/addons/mrp/tests/test_mrp_reports.py
@@ -6,6 +6,10 @@ from odoo.fields import Command
 class TestReportBom(HttpCase):
 
     def test_mrp_report_bom_variant_selection(self):
+        self.env.ref('base.user_admin').write({'group_ids': [
+            Command.link(self.env.ref('product.group_product_variant').id),
+        ]})
+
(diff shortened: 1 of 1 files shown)


addons/mrp/tests/test_mrp_reports.py
```diff
@@ -6,6 +6,8 @@
 class TestReportBom(HttpCase):
 
     def test_mrp_report_bom_variant_selection(self):
+        self.env.ref('base.user_admin').write({'group_ids': [(4, self.env.ref('product.group_product_variant').id)]})
```
clesgow:
```suggestion
        self.env.ref('base.user_admin').write({'group_ids': [
            Command.link(self.env.ref('product.group_product_variant').id),
        ]})
```
Always use [Commands](https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html?highlight=command#odoo.fields.Command) as they're more readable 🙂

PR #248639, opened 2026-02-13
[FIX] point_of_sale: prevent error when cancelling an empty order

meji-odoo:
Steps:
---
- Open a Restaurant session.
- Open a table with no order.
- Click the control button and click Cancel Order.

Issue:
---
- A traceback occurs when cancelling an order.

Cause:
--
- The order is cancelled first, and then `isSelectedLineCombo` tries to access the current order, which is already null.

Fix:
---
- Safely check for the selected order before accessing combo data.
- Disable the Cancel Order button when there are no order lines.

task-5934004

---

### DIFF
--- a/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
+++ b/addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
@@ -50,7 +50,11 @@
                 <button t-if="this.pos.isSelectedLineCombo" class="break-combo-button" t-att-class="buttonClass" t-on-click="() => this.breakSelectedCombo()">
                     <i class="fa fa-outdent" role="img" aria-label="Break combo" title="Brean combo" /> Break combo
                 </button>
-                <button t-if="this.pos.cashier._role != 'minimal'" t-att-class="buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())">
+                <button
+                    t-if="this.pos.cashier._role != 'minimal'"
+                    t-att-class="buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())"
+                    t-att-disabled="!pos.getOrder()?.getOrderlines()?.length"
+                >
                     <i class="fa fa-trash me-1" role="img" /> Cancel Order
                 </button>
             </div>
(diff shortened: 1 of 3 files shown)


addons/pos_restaurant/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -41,5 +41,8 @@
                 </button>
             </t>
         </xpath>
+        <xpath expr="//button[hasclass('cancel-order-button')]" position="attributes">
+            <attribute name="t-att-disabled">pos.config.module_pos_restaurant and !pos.getOrder()?.getOrderlines()?.length</attribute>
+        </xpath>
```
paso-odoo: No need of `pos.config.module_pos_restaurant` condition. you can use orderlines length in pos where button is defined
so this change will be removed

addons/point_of_sale/static/src/app/screens/product_screen/control_buttons/control_buttons.xml
```diff
@@ -47,7 +47,11 @@
                 <button t-if="this.pos.isSelectedLineCombo" class="break-combo-button" t-att-class="buttonClass" t-on-click="() => this.breakSelectedCombo()">
                     <i class="fa fa-outdent" role="img" aria-label="Break combo" title="Brean combo" /> Break combo
                 </button>
-                <button t-if="this.pos.cashier._role != 'minimal'" t-att-class="buttonClass" t-on-click="() => this.pos.onDeleteOrder(this.pos.getOrder())">
+                <button
+                    class="cancel-order-button"
+                    t-if="this.pos.cashier._role != 'minimal'"
```
paso-odoo:
```suggestion
                    t-if="this.pos.cashier._role != 'minimal'"
                    class="cancel-order-button"
```

addons/pos_restaurant/static/tests/tours/control_buttons_tour.js
```diff
@@ -19,6 +19,13 @@ registry.category("web_tour.tours").add("ControlButtonsTour", {
             Chrome.startPoS(),
             Dialog.confirm("Open Register"),
             FloorScreen.clickTable("2"),
+            ProductScreen.clickControlButtonMore(),
+            // check that cancel order button is disabled if there is no orderline in the order
+            {
+                content: "Check that cancel order button is disabled",
+                trigger: ".control-buttons .cancel-order-button:disabled",
+            },
+            Dialog.cancel(),
```
paso-odoo: also test in `point_of_sale`

paso-odoo (changes requested):

davidmonnom (approved):
@robodoo r+

PR #248645, opened 2026-02-13
[FIX] point_of_sale: error when combo contains archived product

parp-odoo:
Steps to reproduce:
- Archive a product used in a combo.
- Open Point of Sale.
- Create a new order.
- A blank screen appears (error in console)

Cause:
- The system attempts to access fields of an archived product included in a combo choice.

Fix:
- Skip loading the combo choice whose product is archived.

Task-5936290


davidmonnom (approved):
@robodoo r+

PR #248655, opened 2026-02-13
[IMP] website_slides: Improve slide_category_o2m_field tests

abz-odoo:
Since https://github.com/odoo/odoo/pull/248480 there is no need to add `is_category` in the view with `invisible='1'`


aab-odoo:
robodoo r+

PR #248661, opened 2026-02-13
[IMP] core: random serial for testing

kmagusiak:
When testing avoid having the first created user and partner having `id=1`. This detects bugs in tests where the id is compared and it matches by chance.

Types of errors:
- mixin related objects like
  - `company_id = self.env.user`
  - `user_id = self.env.user.partner_id.id`
  - product.product/product.template
  - etc.
- `record.abc_ids = (4, some_id)`

https://github.com/odoo/enterprise/pull/107420

---

### DIFF
--- a/addons/account/tests/test_account_journal.py
+++ b/addons/account/tests/test_account_journal.py
@@ -53,7 +53,7 @@ class TestAccountJournal(AccountTestInvoicingCommon, HttpCase):
         invoice_invalid.journal_id = journal
         invoice_invalid.action_post()
         self.assertTrue(invoice_invalid.payment_reference, "A payment reference should be generated.")
-        self.assertIn(str(journal.id), invoice_invalid.payment_reference, "The reference should fall back to using the journal ID.")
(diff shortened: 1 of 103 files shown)


addons/test_discuss_full/tests/test_performance.py
```diff
@@ -1353,7 +1353,7 @@ def _expected_result_for_message(self, channel):
                 ],
                 "record_name": "general",
                 "reply_to": '"Ernest Employee" <catchall.test@test.mycompany.com>',
-                "res_id": 1,
+                "res_id": self.channel_general.id,
```
seb-odoo, line 1361: Not a big deal as they are asserted equal above, but we use `channel.id` instead in other places.

Feyensv (approved):

addons/account/tests/test_account_journal.py
alialfie, line 1: What is the purpose of the changes in this file?

kmagusiak: You suppose that the id is less than 10000. At the beginning, I've put random ids between 10^3 and 10^5 which showed an issue here.

PR #248664, opened 2026-02-13
[FIX] account: cumulated balance not working on general ledger

leclerc-leo:
All cumulated balance are at 0 when opening the Journal Items from an account from the general ledger.

To reproduce:
- Open the general ledger
- Select "Journal Items" with any account with moves during the period
- Show the optional "cumulated_balance" column
- All values are at 0

---


hupo-odoo (approved):
@robodoo r+

PR #248672, opened 2026-02-13
[FIX] product: allow users with read access to print labels

qucol-odoo:
## Issue
Users who do not have the "write" right access on `product.template` cannot print product labels.

## Steps to reproduce
1. Install *Sales* (`sale_management`)
2. In Settings > User & Companies > Users, make sure Marc Demo does not have any write access on `product.template`
3. Log in as Marc Demo
4. In Sales > Products > Products, open a product and click *Print Labels* from the cogwheel menu
5. **An Access Error is shown, saying that the operation is allowed for the `Products/Create` group.**

## Cause
[This commit](https://github.com/odoo/odoo/commit/95ace0a694eaf83329b50e6b89f774f0c59fec5e) removed Products-related rights from the `base.group_user`. This made a difference in terms of access rights, as the `IrActionServe.run` method checks for the "write" access by calling `_can_execute_action_on_records`:

https://github.com/odoo/odoo/blob/d15685304f479541879fabd55ea1cae4252a2a90/odoo/addons/base/models/ir_actions.py#L1230-L1239

When a `group_ids` field is added to the action, this check is no longer performed.

opw-5914988

### DIFF
--- a/addons/product/views/product_template_views.xml
+++ b/addons/product/views/product_template_views.xml
@@ -197,6 +197,7 @@
     <record id="action_product_template_print_labels" model="ir.actions.server">
         <field name="name">Print Labels</field>
         <field name="model_id" ref="product.model_product_template"/>
+        <field name="group_ids" eval="[(4, ref('base.group_user'))]"/>
         <field name="binding_model_id" ref="product.model_product_template"/>
         <field name="state">code</field>
         <field name="code">
--- a/addons/product/tests/test_common.py
+++ b/addons/product/tests/test_common.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
+from odoo.exceptions import AccessError
 from odoo.tests import tagged
 
 from odoo.addons.product.tests.common import ProductCommon
@@ -24,3 +25,20 @@ class TestProduct(ProductCommon):
             self.pricelist,
         )
         self.assertEqual(self.pricelist.currency_id.name, self.currency.name)
+
+    def test_any_user_can_print_product_labels(self):
+        base_user = self.env['res.users'].create({
+            'name': 'Base user',
+            'login': 'base_user',
+            'email': 'base.user@test.com',
+            'group_ids': self.group_user,
+        })
+        print_label_action = self.env.ref('product.action_product_template_print_labels')
+        context = {
+            'active_model': 'product.template',
+            'active_id': self.product.product_tmpl_id,
+        }
+        try:
+            print_label_action.with_user(base_user).with_context(context).run()
+        except AccessError:
+            self.fail("AccessError raised while printing product label with base user.")


barracudapps (changes requested):
Thanks for the fix ! 👍
LGTM technically but would request a quick test.

addons/product/tests/test_barcode.py
```diff
@@ -172,3 +172,24 @@ def test_duplicated_barcodes_in_product_variants(self):
         self.assertEqual(variant_1.company_id, company_b)
         self.assertEqual(variant_2.barcode, 'barcode_3')
         self.assertEqual(variant_2.company_id, company_b)
+
+    def test_print_label_access_rights(self):
+        """
+        Ensure any user can print product template labels.
+        """
```
barracudapps: Could be single line

addons/product/tests/test_barcode.py
```diff
@@ -172,3 +172,24 @@ def test_duplicated_barcodes_in_product_variants(self):
         self.assertEqual(variant_1.company_id, company_b)
         self.assertEqual(variant_2.barcode, 'barcode_3')
         self.assertEqual(variant_2.company_id, company_b)
+
+    def test_print_label_access_rights(self):
```
barracudapps: Having a more explicit test name might lead to docstring removal

addons/product/tests/test_barcode.py
```diff
@@ -172,3 +172,24 @@ def test_duplicated_barcodes_in_product_variants(self):
...
+        """
+        base_user = self.env['res.users'].create({
+            'name': 'Base user',
+            'login': 'base_user',
+            'email': 'base.user@test.com',
+            'group_ids': [Command.link(self.env.ref('base.group_user').id)],
+        })
+        product_template = self.env['product.template'].create({
+            'name': 'Product',
+        })
```
barracudapps:
```suggestion
        product_template = self.env['product.template'].create({'name': 'Product'})
```
Very straight forward test so shouldn't need any product change and as it fits a single line, let's move on like this

addons/product/tests/test_barcode.py
```diff
@@ -172,3 +172,24 @@ def test_duplicated_barcodes_in_product_variants(self):
...
+        print_label_action = self.env.ref('product.action_product_template_print_labels')
+        context = {
+            'active_model': 'product.template',
+            'active_id': product_template.id,
+        }
+        res = print_label_action.with_user(base_user).with_context(context).run()
+        self.assertFalse(res)  # Finished correctly
```
barracudapps: As you know it should lead to an `AccessError` you might want to ensure it is not raised while running the method:
```suggestion
        try:
            print_label_action.with_user(base_user).with_context(context).run()
        except AccessError:
            self.fail("AccessError raised while printing product label with base user.")
```

And maybe this more complete version:
```suggestion
        try:
            print_label_action.with_user(base_user).with_context(context).run()
        except AccessError:
            self.fail("AccessError raised while printing product label with base user.")
        except Exception:
            self.fail("An exception was raised while printing product label with base user.")
```

FYI, based on https://github.com/odoo-dev/odoo/blob/06089ca89f6c0dd18e30f0ae04bfaf66bdc09591/addons/payment/tests/common.py#L215-L234

barracudapps (approved):
Thank you for the test but some nits here.
Feel free to pick the ones you want to apply then put the PR as ready.

addons/product/views/product_template_views.xml
```diff
@@ -203,6 +203,7 @@
             if records:
                 action = records.action_open_label_layout()
         </field>
+        <field name="group_ids" eval="[(4, ref('base.group_user'))]"/>
```
ticodoo: It's weird to put the `groups_ids` after the `code`, put it after the "model_id" instead so it's easier to see/in a more expected spot

addons/product/tests/test_barcode.py
```diff
@@ -172,3 +172,21 @@ def test_duplicated_barcodes_in_product_variants(self):
         self.assertEqual(variant_1.company_id, company_b)
         self.assertEqual(variant_2.barcode, 'barcode_3')
         self.assertEqual(variant_2.company_id, company_b)
+
+    def test_any_user_can_print_product_labels(self):
```
ticodoo: Test isn't related to barcodes, move this to `test_common.py` and use the existing `setUpClass` values like `self.product` and `self.group_user` instead of creating/querying from scratch for better performance.

It's also weird to have a test that never `assert`s anything. I'm not sure what `run` returns when it's successful, but if it's the `actions.act_window`, can you do an asserts that that is what is returned? If not, then the try/except is ok as is.

ticodoo (commented):
Commit analysis is incorrect, the cause is due to: 
https://github.com/odoo/odoo/pull/197286/changes/483b644c43d5a333c4f30d5907b846dd37c88294

Since before that group_users had write access to products. This is why the bug exists in saas-18.3+ and not saas-18.2 (which is when the linked commit in the current commit message was added).

Fix is ok with a little bit of cleanup 👍

ticodoo:
Thanks!

@robodoo r+

PR #248677, opened 2026-02-13
[FIX] snailmail: fix fetch attachment access error

amay-odoo:
Issue:
Before this commit, when sending a follow up report by post, an access error is thrown if the user doesn't have enough access to modify the res.company model

Fix:
modifying the external_report_layout_id as sudo

opw-5482855


reth-odoo (approved):
Probably fine, maybe add a comment to justify the sudo (even if it's obvious now)
@lobaudoux @jmak-odoo

lobaudoux (approved):

tde-banana-odoo:
@robodoo r+

PR #248679, opened 2026-02-13
[FIX] l10n_ro_edi_stock: fix report crash in non-English languages

amahm-odoo:
Steps to reproduce:
1. Switch the partner language to Spanish.
2. Print the Delivery Slip.
3. The report crashes due to translated XPath anchors.

Cause:
XPath targeted hardcoded English strings 'Tracking Number' and 'Total Weight'.

Solution:
Target technical fields o.carrier_tracking_ref and o.shipping_weight instead of text labels.

opw-5493498

### DIFF
--- a/addons/l10n_ro_edi_stock/report/report_deliveryslip.xml
+++ b/addons/l10n_ro_edi_stock/report/report_deliveryslip.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
     <template id="l10n_ro_edi_stock_report_delivery_document" inherit_id="stock.report_delivery_document" priority="100">
-        <xpath expr="//div/strong[text()='Tracking Number']/.. | //div/strong[text()='Total Weight']/.." position="after">
+        <xpath expr="//div[div[@t-field='o.carrier_tracking_ref']] | //div[div[span[@t-field='o.shipping_weight']]]" position="after">
             <div t-if="o.l10n_ro_edi_stock_enable and o.l10n_ro_edi_stock_document_uit" class="col-auto" name="div_etransport_uit">
                 <strong>eTransport UIT</strong>
                 <p t-field="o.l10n_ro_edi_stock_document_uit"/>


addons/l10n_ro_edi_stock/report/report_deliveryslip.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
     <template id="l10n_ro_edi_stock_report_delivery_document" inherit_id="stock.report_delivery_document" priority="100">
-        <xpath expr="//div/strong[text()='Tracking Number']/.. | //div/strong[text()='Total Weight']/.." position="after">
+        <xpath expr="//div[div[@t-field='o.carrier_tracking_ref']] | //div[span[@t-field='o.shipping_weight']]" position="after">
```
lase-odoo: While I agree it now targets un-translated isn't still a bad x-path ? both anchors are purely optional as they are in "t-if"'s.

lase-odoo: This solves the traceback issue but it creates a rendering issue, the part of the x-path related to the `shipping_weight` is not correct as the span of the-field is contained in a div we do not want to target ! 
https://github.com/odoo/odoo/blob/f688c6b66310438fa3e36a207770a63d0d8fffa5/addons/stock_delivery/views/report_deliveryslip.xml#L9-L19
Same as for the carrier, you need to target the div containing the div:
```suggestion
        <xpath expr="//div[div[@t-field='o.carrier_tracking_ref']] | //div[div[span[@t-field='o.shipping_weight']]]" position="after">
```
Here is the result with your proposition:

Image

Here is the result with mine:

Image

lase-odoo (commented):
Thanks a lot for the proposition ! 🦐

Good fix idea ! There is just a small issue with the current proposition.

lase-odoo:
robodoo r+

aboo-odoo (approved):
@robodoo r+

PR #248680, opened 2026-02-13
[PERF] account*: Improve upload of XML bill

auma-odoo:
### Description:

The upload and import process for large XML bills via Peppol or manual
upload was inefficient due to two primary bottlenecks. First, the
system performed individual queries per line to match products, taxes,
and accounts, leading to an N+1 query issue. Second, multiple write
operations were executed on each line to update various fields.

This commit introduces batching and improve caching for these
operations to reduce database call.

### Benchmark:

| N° of lines |  Before | After |
|-------------|---------|-------|
|       30264 | Timeout | 11min |

### Reference:
opw-5416612

### DIFF
--- a/addons/account/models/account_account.py
+++ b/addons/account/models/account_account.py
@@ -783,8 +783,18 @@ class AccountAccount(models.Model):
 
     @api.model
     def _get_most_frequent_account_for_partner(self, company_id, partner_id, move_type=None, journal_id=None):
-        most_frequent_account = self._get_most_frequent_accounts_for_partner(company_id, partner_id, move_type, filter_never_user_accounts=True, limit=1, journal_id=journal_id)
-        return most_frequent_account[0] if most_frequent_account else False
+
+        cache = self.env.cr.cache.setdefault('most_frequent_accounts_for_partner', {})
+        key = (company_id, partner_id, move_type, journal_id)
+
+        if key not in cache:
+            most_frequent_account = self._get_most_frequent_accounts_for_partner(
+                company_id, partner_id, move_type,
+                filter_never_user_accounts=True, limit=1, journal_id=journal_id
+            )
+            cache[key] = most_frequent_account[0] if most_frequent_account else False
+
+        return cache[key]
 
     @api.model
     def _order_accounts_by_frequency_for_partner(self, company_id, partner_id, move_type=None):
--- a/addons/account/models/account_move.py
+++ b/addons/account/models/account_move.py
@@ -3006,6 +3006,7 @@ class AccountMove(models.Model):
 
         to_delete = []
         to_create = []
+        grouped_update = defaultdict(set)
         for move in container['records']:
             if move.state != 'draft':
                 continue
@@ -3065,7 +3066,7 @@ class AccountMove(models.Model):
             for base_line, to_update in tax_results['base_lines_to_update']:
                 line = base_line['record']
                 if is_write_needed(line, to_update):
-                    line.write(to_update)
+                    grouped_update[line.currency_id.id, frozendict(to_update)].add(line.id)
 
             for tax_line_vals in tax_results['tax_lines_to_delete']:
                 to_delete.append(tax_line_vals['record'].id)
@@ -3080,8 +3081,12 @@ class AccountMove(models.Model):
             for tax_line_vals, grouping_key, to_update in tax_results['tax_lines_to_update']:
                 line = tax_line_vals['record']
                 if is_write_needed(line, to_update):
-                    line.write(to_update)
+                    grouped_update[line.currency_id.id, frozendict(to_update)].add(line.id)
 
+        if grouped_update:
+            # Need to use currency_id as a key to avoid writing with multiple currencies
+            for (currency_id, values), lines in grouped_update.items():
+                self.env['account.move.line'].browse(lines).write(dict(values))
         if to_delete:
             self.env['account.move.line'].browse(to_delete).with_context(dynamic_unlink=True).unlink()
         if to_create:
@@ -3202,9 +3207,13 @@ class AccountMove(models.Model):
         yield
         after = existing()
 
+        partner_id_to_update = defaultdict(set)
         for move in after:
             if changed('commercial_partner_id'):
-                move.line_ids.partner_id = after[move]['commercial_partner_id']
+                partner_id_to_update[after[move]['commercial_partner_id']].update(move.line_ids.ids)
+
+        for partner_id, line_ids in partner_id_to_update.items():
+            self.env['account.move.line'].browse(line_ids).partner_id = partner_id
 
     @contextmanager
     def _sync_dynamic_lines(self, container):
--- a/addons/account/models/account_tax.py
+++ b/addons/account/models/account_tax.py
@@ -4397,7 +4397,7 @@ class AccountTax(models.Model):
 
     @api.model
     def _import_retrieve_tax(self, search_plan, company, tax_values_list):
-        cache = {}
+        cache = self.env.cr.cache.setdefault('retrieved_tax_map', {})
 
         static_domain = expression.OR([
             [*self._check_company_domain(company), ('company_id', '!=', False)],
@@ -4453,9 +4453,9 @@ class AccountTax(models.Model):
                             'static_domain': expression.AND([tax_domain, static_domain]),
                         })
 
+                    if cache_key:
+                        cache[cache_key] = tax
                     if tax:
-                        if cache_key:
-                            cache[cache_key] = tax
                         tax_values['tax'] = tax
                         break
 
--- a/addons/account/tests/common.py
+++ b/addons/account/tests/common.py
@@ -208,6 +208,11 @@ class AccountTestInvoicingCommon(ProductCommon):
             ],
         })
 
+    def setUp(self):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+        super().setUp()
+
     @classmethod
     def change_company_country(cls, company, country):
         company.country_id = country
@@ -394,6 +399,9 @@ class AccountTestInvoicingCommon(ProductCommon):
     # -------------------------------------------------------------------------
 
     def group_of_taxes(self, taxes, **kwargs):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+
         self.tax_number += 1
         return self.env['account.tax'].create({
             'name': f"group_({self.tax_number})",
@@ -403,6 +411,9 @@ class AccountTestInvoicingCommon(ProductCommon):
         })
 
     def percent_tax(self, amount, **kwargs):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+
         self.tax_number += 1
         return self.env['account.tax'].create({
             'name': f"percent_{amount}_({self.tax_number})",
@@ -412,6 +423,9 @@ class AccountTestInvoicingCommon(ProductCommon):
         })
 
     def division_tax(self, amount, **kwargs):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+
         self.tax_number += 1
         return self.env['account.tax'].create({
             'name': f"division_{amount}_({self.tax_number})",
@@ -421,6 +435,9 @@ class AccountTestInvoicingCommon(ProductCommon):
         })
 
     def fixed_tax(self, amount, **kwargs):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+
         self.tax_number += 1
         return self.env['account.tax'].create({
             'name': f"fixed_{amount}_({self.tax_number})",
@@ -431,6 +448,10 @@ class AccountTestInvoicingCommon(ProductCommon):
 
     def python_tax(self, formula, **kwargs):
         self.ensure_installed('account_tax_python')
+
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
+
         self.tax_number += 1
         return self.env['account.tax'].create({
             'name': f"code_({self.tax_number})",
--- a/addons/account_edi_ubl_cii/models/account_edi_ubl.py
+++ b/addons/account_edi_ubl_cii/models/account_edi_ubl.py
@@ -2947,15 +2947,17 @@ class AccountEdiUBL(models.AbstractModel):
         if not self.module_installed('account_accountant'):
             # _predict_specific_account is defined in account_accountant
             return
+
+        accounts_map = {}
         lines_collected_values = collected_values['lines_collected_values']
         for line_collected_values in lines_collected_values:
             account_values = line_collected_values['account_values']
             if predictive := account_values.get('invoice_predictive'):
-                account_id = self.env['account.move.line']._predict_specific_account(
-                    move=predictive['invoice'],
-                    name=predictive['name'],
-                    partner=predictive['partner'],
-                )
+                account_params = {'move': predictive['invoice'], 'name': predictive['name'], 'partner': predictive['partner']}
+                account_key = tuple(account_params.values())
+                if account_key not in accounts_map:
+                    accounts_map[account_key] = self.env['account.move.line']._predict_specific_account(**account_params)
+                account_id = accounts_map.get(account_key)
                 if account_id:
                     account_values['account'] = self.env['account.account'].browse(account_id)
 


addons/account/models/account_move.py
```diff
@@ -3064,7 +3065,8 @@ def is_write_needed(line, values):
             for base_line, to_update in tax_results['base_lines_to_update']:
                 line = base_line['record']
                 if is_write_needed(line, to_update):
-                    line.write(to_update)
+                    to_update = self._clean_command(to_update)
+                    grouped_update[line.currency_id.id, frozenset(to_update.items())].add(line.id)
```
william-andre: Why do we need the currency?

addons/account/models/account_move.py
```diff
@@ -3079,13 +3081,35 @@ def is_write_needed(line, values):
             for tax_line_vals, grouping_key, to_update in tax_results['tax_lines_to_update']:
                 line = tax_line_vals['record']
                 if is_write_needed(line, to_update):
-                    line.write(to_update)
+                    to_update = self._clean_command(to_update)
+                    grouped_update[line.currency_id.id, frozenset(to_update.items())].add(line.id)
```
william-andre: Why not a frozendict instead of the frozenset?

addons/account/models/account_move.py
```diff
@@ -3079,13 +3081,35 @@ def is_write_needed(line, values):
...
 
         if to_delete:
             self.env['account.move.line'].browse(to_delete).with_context(dynamic_unlink=True).unlink()
         if to_create:
             self.env['account.move.line'].create(to_create)
 
+    def _clean_command(self, dict_to_clean):
+        clean_to_update = {}
+        for key, value in dict_to_clean.items():
+            if isinstance(value, (list, tuple)) and len(value) > 0 and isinstance(value[0], (list, tuple)):
```
william-andre: It would be way easier to understand with a `match` pattern matching

addons/account/models/account_move.py
```diff
@@ -3079,13 +3081,35 @@ def is_write_needed(line, values):
...
+                command = value[0]
```
william-andre: What if there are multiple commands?

addons/account/models/account_move.py
```diff
@@ -3064,7 +3065,8 @@ def is_write_needed(line, values):
             for base_line, to_update in tax_results['base_lines_to_update']:
                 line = base_line['record']
                 if is_write_needed(line, to_update):
-                    line.write(to_update)
+                    to_update = self._clean_command(to_update)
```
william-andre: Why do you need to "clean" anything?

addons/account/models/account_move.py
william-andre, line 1: What is the impact of batching the writes? What is ran at each call of `write` that is taking time?

addons/account/models/account_move.py
```diff
@@ -3079,13 +3081,35 @@ def is_write_needed(line, values):
...
+                    clean_to_update[key] = command[1]
+                else:
+                    clean_to_update[key] = tuple(command)
+            else:
+                clean_to_update[key] = tuple(value) if isinstance(value, list) else value
+
+        return clean_to_update
+
     @contextmanager
     def _sync_dynamic_line(self, existing_key_fname, needed_vals_fname, needed_dirty_fname, line_type, container):
```
william-andre, line 3096: I'm surprised that we don't need to batch anything in `_sync_dynamic_line` with `to_write`?

addons/account/models/account_move_line.py
```diff
@@ -601,6 +601,7 @@ def _compute_account_id(self):
                 line.account_id = account_id
 
         product_lines = self.filtered(lambda line: line.display_type == 'product' and line.move_id.is_invoice(True))
+        all_frequent_account_id_dict = defaultdict()
```
william-andre: This could be a simple `dict`
```suggestion
        frequent_account_cache = {}
```

addons/account/models/account_move_line.py
```diff
@@ -611,12 +612,22 @@ def _compute_account_id(self):
...
+                if key in all_frequent_account_id_dict:
+                    account_id = all_frequent_account_id_dict[key]
+                else:
+                    account_id = self.env['account.account']._get_most_frequent_account_for_partner(
+                        company_id=company_id,
+                        partner_id=partner_id,
+                        move_type=move_type,
+                        journal_id=journal_id,
+                    )
+                    all_frequent_account_id_dict[key] = account_id
```
william-andre:
```suggestion
                args = frozendict(
                    company_id=company_id,
                    partner_id=partner_id,
                    move_type=move_type,
                    journal_id=journal_id,
                )
                if args not in all_frequent_account_id_dict:
                    frequent_account_cache[key] = self.env['account.account']._get_most_frequent_account_for_partner(**args)
                account_id = frequent_account_cache[args]
```

addons/account/models/account_move_line.py
```diff
@@ -611,12 +612,22 @@ def _compute_account_id(self):
                 elif line.move_id.is_purchase_document(include_receipts=True):
                     line.account_id = accounts['expense'] or line.account_id
             elif line.partner_id:
-                account_id = self.env['account.account']._get_most_frequent_account_for_partner(
```
william-andre, line 616: One could also do something like this
```python
from functools import cache

def _compute_account_id(self):
    _get_most_frequent_account_for_partner = cache(self.env['account.account']._get_most_frequent_account_for_partner)

    ...

    account_id = _get_most_frequent_account_for_partner(
        company_id=line.company_id.id,
        partner_id=line.partner_id.id,
        move_type=line.move_id.move_type,
        journal_id=line.journal_id.id,
    )
```

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -895,6 +919,7 @@ def _retrieve_taxes(self, record, line_values, tax_type, tax_exigibility=False):
         # if no results, try to fetch the price_include=True taxes. If results, need to adapt the price_unit.
         logs = []
         taxes = []
+        stored_taxes = self.env.context.get('stored_taxes', defaultdict())
```
william-andre: Don't use the context, you can use `cr.cache` instead for instance

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -903,8 +928,8 @@ def _retrieve_taxes(self, record, line_values, tax_type, tax_exigibility=False):
                 ('type_tax_use', '=', tax_type),
                 ('amount', '=', amount),
             ]
-            tax = self.env['account.tax']
-            if hasattr(record, '_get_specific_tax'):
+            tax = stored_taxes.get(amount, self.env['account.tax'])
```
william-andre: The amount is not enough.
You also need the tax type and the company, the tax exigibility,... Basically the whole domain.

In the end, maybe adding a cache enclosed in `<account.tax>.search` would probably be better and solve more cases (e.g. other EDIs)

addons/account/models/product.py
```diff
@@ -335,3 +338,79 @@ def _retrieve_product(self, name=None, default_code=None, barcode=None, company=
...
+        if name_ilike_list:
+            Template = self.env['product.template']
+            subquery = Template._search([('active', '=', True)])
+            subquery.add_where(SQL(
+                "(jsonb_path_query_array(%(name)s, '$.*')::text) ILIKE ANY(%(values)s)",
+                name=SQL.identifier(Template._table, 'name'),
+                values=list(name_ilike_list)
+            ))
+
+            domains.append([('product_tmpl_id', 'in', subquery)])
```
william-andre: Why does the `ilike` domain operator not work here?

If I'm not mistaken, the difference with your code is only that you check all the langs? (which I'm not sure is good)

addons/account/models/product.py
```diff
@@ -335,3 +338,79 @@ def _retrieve_product(self, name=None, default_code=None, barcode=None, company=
...
+            ):
+                for domain in domains:
+                    if products := self.env['product.product']._read_group(
+                        domain=expression.AND([domain, company_domain, extra_domain or []]),
+                        aggregates=['id:recordset'],
+                        groupby=['barcode', 'default_code', 'name'],
+                    ):
+                        for barcode, default_code, name, product in products:
+                            keys = [barcode, default_code, name]
+                            all_products.update({key: product for key in keys if key})
```
william-andre: Don't we want 3 mappings: `barcode_mapping`, `code_mapping`, `name_mapping`, which allows finding the products using the priority barcode > code > name?

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -778,8 +792,15 @@ def _retrieve_line_vals(self, tree, document_type=False, qty_factor=1):
 
         # delivered_qty (mandatory)
         delivered_qty = 1
+        prefetched_products = self.env.context.get('prefetched_products', {})
```
william-andre: Use `cr.cache`

addons/account/models/product.py
```diff
@@ -335,3 +338,79 @@ def _retrieve_product(self, name=None, default_code=None, barcode=None, company=
...
+        for products_vals in split_every(1000, list_of_product_vals):
+            domains = self._get_products_domain_batched(products_vals)
+
+            company = company or self.env.company
+            for company_domain in (
+                [*self.env['res.partner']._check_company_domain(company), ('company_id', '!=', False)],
+                [('company_id', '=', False)],
+            ):
+                for domain in domains:
+                    if products := self.env['product.product']._read_group(
```
william-andre: Isn't a `_read_group` worse in the normal case, where there aren't too many products?

addons/account/models/account_move.py
auma-odoo: None. Didn't know we had a frozendict available in the tools 🙈

addons/account/models/account_move.py
auma-odoo: It was mostly for the frozenset since I can put an unhashable element into it. Could froze the command but easier to just "clean" it.

addons/account/models/product.py
auma-odoo: The issue is that we can't use a list of strings in `ilike` from what I remember. Also I used this definition so the index can be hit since we have a trigram index on `(jsonb_path_query_array(name, '$.*')::text)` by default.

addons/account/models/product.py
auma-odoo: I thought about it, but the issue is that it means to do all 3 queries anyway, since we can't check if we already have the product for each domain per product. Otherwise, it defeats the purpose of batching. 

Or do you mean just give 3 mappings as a return value? Could be a solution, but I fail to see the purpose in this case (other than keeping the function as is) 🤔. Do we need to prioritize `barcode` and `default_code` to match the product with the line?

addons/account/models/product.py
auma-odoo: It is true, but I am not sure if the speed difference is significant in this case. Will need to check that locally 🤔

addons/account_edi_ubl_cii/models/account_edi_common.py
auma-odoo: What do you mean by that? By adding a custom search in `account.tax` with the result of the search in cache?

william-andre: Caching only based on the amount will lead to conflicts because there are more parameters to it

auma-odoo: I agree with you that it is not the best way to store the tax. But I am wondering about how we could implement the `account.tax.search` caching? We could do the same as you proposed before with using the LRU cache 🤔

william-andre: Cache `search` by using the domain as a key?

addons/account/models/account_move.py
auma-odoo: In my case, the issue are calls on `modified` and `modified_triggers`. I wanted to batch most of them to reduce the amounts of calls to these parts of the code

addons/account/models/account_account.py
```diff
@@ -782,6 +782,7 @@ def _get_most_frequent_accounts_for_partner(self, company_id, partner_id, move_t
         ))]
 
     @api.model
+    @ormcache('company_id', 'partner_id', 'move_type', 'journal_id')
```
william-andre: I believe cache per partner is too big for ormcache.
It would be fine for a local/temporary cache on the registry though

addons/account/models/account_move_line.py
```diff
@@ -617,6 +617,7 @@ def _compute_account_id(self):
                     move_type=line.move_id.move_type,
                     journal_id=line.journal_id.id,
                 )
+
```
william-andre: 🏳️

addons/account_edi_ubl_cii/models/account_edi_common.py
william-andre, line 1: We shouldn't have to do anything special in `account_edi_ubl_cii` or any localization; the caching of the relevant methods should be enough on its own

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -934,6 +939,33 @@ def _retrieve_taxes(self, record, line_values, tax_type, tax_exigibility=False):
                     line_values['price_unit'] *= (1 + tax.amount / 100)
         return taxes, logs
 
+    @ormcache('company_id.id', 'tax_type', 'amount', 'tax_exigibility')
```
william-andre: Can't cache stuff like this, otherwise we would need to invalidate the cache on any write on `account.tax`.

Invalidating `ormcache` is expensive (it needs to invalidate everything on all the servers), so to be avoided.

addons/account/models/product.py
william-andre: That's a shame, because postgres does allow it...
Here is a simple PoC (I did it in 19.0 though)
```diff
diff --git a/odoo/orm/domains.py b/odoo/orm/domains.py
index a8cdc67b6084..d7a8aa20925d 100644
--- a/odoo/orm/domains.py
+++ b/odoo/orm/domains.py
@@ -1401,7 +1401,7 @@ def _optimize_like_str(condition, model):
         return condition
     if '=' in condition.operator:
         condition._raise("The pattern to match must be a string", error=TypeError)
-    return DomainCondition(condition.field_expr, condition.operator, str(value))
+    return DomainCondition(condition.field_expr, condition.operator, value)


 @field_type_optimization(['many2one', 'one2many', 'many2many'])
diff --git a/odoo/orm/fields.py b/odoo/orm/fields.py
index c4fc8f20f1e5..b8791babd17c 100644
--- a/odoo/orm/fields.py
+++ b/odoo/orm/fields.py
@@ -1320,15 +1320,17 @@ class Field(typing.Generic[T]):

             # add wildcard and unaccent depending on the operator
             need_wildcard = '=' not in operator
+            if not isinstance(value, list | tuple):
+                value = [value]
             if need_wildcard:
-                sql_value = SQL("%s", f"%{value}%")
+                sql_value = SQL("%s", [f"%{v}%" for v in value])
             else:
-                sql_value = SQL("%s", str(value))
+                sql_value = SQL("%s", [str(v) for v in value])
             if operator.endswith('ilike'):
                 sql_left = model.env.registry.unaccent(sql_left)
                 sql_value = model.env.registry.unaccent(sql_value)

-            sql = SQL("%s%s%s", sql_left, SQL_OPERATORS[operator], sql_value)
+            sql = SQL("%s%s ANY(%s)", sql_left, SQL_OPERATORS[operator], sql_value)
             if operator in Domain.NEGATIVE_OPERATORS and can_be_null:
                 sql = SQL("(%s OR %s IS NULL)", sql, sql_field)
             return sql
```

```python
In [1]: self.env['res.partner'].search([('name', 'ilike', ['Administ', 'My'])])
ic| value: ['Administ', 'My']
ic| sql_value: SQL('%s', ['%Administ%', '%My%'])
Out[1]: res.partner(3, 1)

In [2]: self.env['res.partner'].search([('name', 'ilike', ['Administ'])])
ic| value: ['Administ']
ic| sql_value: SQL('%s', ['%Administ%'])
Out[2]: res.partner(3,)

In [3]: self.env['res.partner'].search([('name', 'ilike', ['My'])])
ic| value: ['My']
ic| sql_value: SQL('%s', ['%My%'])
Out[3]: res.partner(1,)

In [4]: self.env['res.partner'].search([]).mapped('name')
Out[4]: ['Administrator', 'My Company']
```

@kmagusiak Do you have an opinion on this?

addons/account/models/product.py
```diff
@@ -339,28 +314,119 @@ def find_product_by_name_similarity(base_domain):
...
+                            if product.default_code:
+                                lookup['default_code', product.default_code] = product
+                            if product.name:
+                                lookup['name', product.name.lower()] = product
+
+                        for val_dict in product_vals:
+
+                            barcode = val_dict.get('barcode')
+                            code = val_dict.get('default_code')
+                            raw_name = (val_dict.get('name') or '').split('\n', 1)[0]
```
william-andre:
```suggestion
                            # cut Sales Description from the name
                            raw_name = (val_dict.get('name') or '').split('\n', 1)[0]
```

addons/account/models/product.py
```diff
@@ -339,28 +314,119 @@ def find_product_by_name_similarity(base_domain):
...
-            if name and len(name) > 4 and (product := find_product_by_name_similarity(base_domain)):
-                return product
+    def _retrieve_products_batched(self, list_of_product_vals, company=None, extra_domain=None):
+        '''Search all products and find one that matches one of the parameters.
 
-        return self.env['product.product']
+        :param company:         The company of the product.
+        :param extra_domain:    Any extra domain to add to the search.
+        :param product_vals:    Values the product should match.
+        :returns:               A product or an empty recordset if not found.
```
william-andre, line 479: Wrong docstring

addons/account/models/product.py
```diff
@@ -339,28 +314,119 @@ def find_product_by_name_similarity(base_domain):
...
+        '''
+
+        all_products = defaultdict(lambda: self.env['product.product'])
```
william-andre: misleading variable name

addons/account/models/product.py
```diff
@@ -339,28 +314,119 @@ def find_product_by_name_similarity(base_domain):
...
+                                all_products[raw_name] = match
+
+        return all_products
+
+    def _get_products_domain_batched(self, list_of_vals):
+        """Gives the domain for all values linked to a product.
+
+        :param list_of_vals:            The list of barcode, default_code and name for each product.
+        :returns:               The full domain based on all the values in the parameter.
+        :rtype: Domain
```
william-andre: wrong param, returns and rtype

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -657,8 +657,18 @@ def _import_rounding_amount(self, invoice, tree, xpath, qty_factor):
...
         logs = []
         lines_values = []
+        all_product_vals = []
         for line_tree in tree.iterfind(xpath):
-            line_values = self.with_company(invoice.company_id)._retrieve_invoice_line_vals(line_tree, invoice.move_type, qty_factor)
+            xpath_dict = self._get_line_xpaths(invoice.move_type, qty_factor)
+            product_vals = {k: self._find_value(v, line_tree) for k, v in xpath_dict['product'].items()}
+            all_product_vals.append(product_vals)
+
+        self.env.cr.cache['prefetched_products'] = self._import_products_batched(all_product_vals)
```
william-andre: Can't we avoid caching multiple times?

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -657,8 +657,18 @@ def _import_rounding_amount(self, invoice, tree, xpath, qty_factor):
...
+
+        for line_tree in tree.iterfind(xpath):
+            line_values = self.with_company(invoice.company_id)._retrieve_invoice_line_vals(
+                line_tree, invoice.move_type, qty_factor,
+            )
```
william-andre: I think that no diff is needed here?

addons/account/models/product.py
```diff
@@ -339,28 +314,119 @@ def find_product_by_name_similarity(base_domain):
...
+                            code = val_dict.get('default_code')
+                            raw_name = (val_dict.get('name') or '').split('\n', 1)[0]
+
+                            if barcode and (match := lookup.get(('barcode', barcode))):
+                                all_products[barcode] = match
+                            elif code and (match := lookup.get(('default_code', code))):
+                                all_products[code] = match
+                            elif raw_name and (match := lookup.get(('name', raw_name.lower()))):
+                                all_products[raw_name] = match
+
```
william-andre: Why don't we use the same key on `lookup` and `all_products`?

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -789,8 +799,15 @@ def _retrieve_line_vals(self, tree, document_type=False, qty_factor=1):
...
+        prefetched_products = self.env.cr.cache.get('prefetched_products', None)
         product_vals = {k: self._find_value(v, tree) for k, v in xpath_dict['product'].items()}
-        product = self._import_product(**product_vals)
+        if prefetched_products is None:
+            product = self._import_product(**product_vals)
+        else:
+            barcode = product_vals.get('barcode')
+            default_code = product_vals.get('default_code')
+            name = product_vals.get('name')
+            product = prefetched_products.get(barcode) or prefetched_products.get(default_code) or prefetched_products.get(name) or self.env['product.product']
```
william-andre:
```suggestion
            product = (
                prefetched_products.get(product_vals.get('barcode')) 
                or prefetched_products.get(product_vals.get('default_code'))
                or prefetched_products.get(product_vals.get('name'))
                or self.env['product.product']
            )
```

addons/account_edi_ubl_cii/models/account_edi_common.py
```diff
@@ -906,18 +926,26 @@ def _retrieve_taxes(self, record, line_values, tax_type, tax_exigibility=False):
         """
         # Taxes: all amounts are tax excluded, so first try to fetch price_include=False taxes,
         # if no results, try to fetch the price_include=True taxes. If results, need to adapt the price_unit.
+
+        # Avoid issue of cache not being cleaned between tests in test mode
```
william-andre: Why would it not be cleaned?
I thought savepoints cleaned the cache?

addons/account_edi_ubl_cii/models/account_edi_common.py
william-andre, line 939: Don't we want to store `None` or `account.tax()` here so that we don't search again for something that doesn't exist (especially since it is the most expensive case)

addons/account_edi_ubl_cii/models/account_edi_common.py
auma-odoo: From what I saw, it was not cleaned in some test between XML import, causing an error when trying to retrieve the tax with an ID that was not in the database anymore

william-andre: Would cleaning the cache when a tax is created/updated/deleted help?

addons/account/models/product.py
kmagusiak: If something, it would be for master. I am not against, but... The issues I see:
- There is code (in orm and in business modules) that assumes that the value of a ilike is a str.
- I have rand a small query and pg did not use the index with the any operator. My query/db might be too simple, so that would require some more performance investigation.

william-andre: > I have rand a small query and pg did not use the index with the any operator. My query/db might be too simple, so that would require some more performance investigation.

I already used this to improve queries in the past, where it was actually using the index with `x ILIKE ANY(ARRAY['a', 'b', 'c', 'd'])` but not with `x ILIKE 'a' OR x ILIKE 'b' OR x ILIKE 'c' OR x ILIKE 'd'`

addons/account/models/account_tax.py
william-andre, line 4435: use a normalized `Domain` object in 18.2

william-andre, line 4400: why is this not a transaction cache?

auma-odoo: Because now that the method has been batched, we don't really need to keep the value in transaction cache anymore. It is never cleared during the fetching of the taxes. It feels a little overkill when we can just use a dict located in the function

william-andre: > Because now that the method has been batched

It doesn't look like it is 🤔 

It is called in `_import_ubl_invoice_retrieve_taxes`, which is called in `_ubl_import_invoice`, which is clearly for only one invoice. The same tax will probably have to be detected on several invoices in the same transaction.

Same for `_import_ubl_invoice_retrieve_accounts` actually, but that one has the invoice being part of the cache key, so it would require a bigger refactor.

auma-odoo: I was mostly thinking of this function for one invoice. But it is true that we can cache for the case when we have multiple invoices.

addons/account_edi_ubl_cii/tests/test_ubl_import_bis3_invoice_be.py
```diff
@@ -50,6 +50,8 @@ def test_import_discount_per_line_price_on_big_quantity(self):
 
     @freeze_time('2020-01-01')
     def test_import_lot_of_decimals_in_quantities(self):
+        # Clear the cursor cache to avoid missing error
+        self.env.cr.cache.pop('retrieved_tax_map', None)
```
william-andre, line 54: Isn't the test suite supposed to do this?
Or maybe not in 18?

william-andre (approved):
@robodoo r+

addons/account_edi_ubl_cii/tests/test_ubl_import_bis3_invoice_be.py
auma-odoo: Not from my test and investigation. Seems the cache is not cleared between tests

william-andre: Clear the cache in `AccountTestInvoicingCommon.setUp` instead, as well as in the C~~R~~UD of `account.tax`

auma-odoo: Should be good now

william-andre:
@robodoo r+

StefanRijnhart:
This causes taxes to be assigned from the wrong company in our setup. Please consider the fix proposed in https://github.com/odoo/odoo/pull/262779

PR #248686, opened 2026-02-13
[FIX] sale_timesheet: add conditions to show timesheet portal link in mail template

aurb9:
Previously, the line displaying a link for customers to view the timesheets in the account move mail template was shown as soon as there were timesheets linked to the account move.

This means that the sentence appears in the invoice email even if the customer account is only accessible based on invitation and that the client was not invited to create one, or if they do not have portal access, leading to confusion.

In this commit, we make restrict further the visibility based on whether the customer already has portal access OR if there is a free sign up to create an account configured in the company.

Note that this bug targets master and no previous version because the only solution involves adding a field, forbidden in stable versions.

See odoo/upgrade#9499

task-5248969

### DIFF
--- a/addons/account/data/mail_template_data.xml
+++ b/addons/account/data/mail_template_data.xml
@@ -56,7 +56,7 @@
                 .
             </t>
         </t>
-        <t t-if="hasattr(object, 'timesheet_count') and object.timesheet_count">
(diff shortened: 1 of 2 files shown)


xavierbol:
LGTM, thanks for your work!

robodoo r+

PR #248687, opened 2026-02-13
[IMP] test_discuss_full: assertQueries

kmagusiak:
Use `assertQueries` instead of comments which became obsolete as refactors go.

---

### DIFF
--- a/addons/test_discuss_full/tests/test_performance.py
+++ b/addons/test_discuss_full/tests/test_performance.py
@@ -13,149 +13,11 @@ from odoo.tests.common import users, tagged, HttpCase, warmup
 
 @tagged("is_query_count")
 class TestDiscussFullPerformance(HttpCase, MailCommon):
-    # Queries for _query_count_init_store (in order):
-    #   1: search res_partner (odooot ref exists)
-    #   1: search res_groups (internalUserGroupId ref exists)
-    #   8: store add odoobot:
-    #       - fetch res_partner (_read_format)
-    #       - search res_users (_compute_im_status)
-    #       - search presence (_compute_im_status)
-    #       - fetch presence (_compute_im_status)
-    #       - _get_on_leave_ids (_compute_im_status hr_holidays override)
-    #       - search employee (_compute_im_status hr_homeworking override)
-    #       - fetch employee (_compute_im_status hr_homeworking override)
-    #       - fetch res_users (_read_format)
-    #       - fetch hr_employee (user)
-    #   5: settings:
-    #       - search res_users_settings (_find_or_create_for_user)
-    #       - fetch res_users_settings (_format_settings)
-    #       - search res_users_settings_volumes (_format_settings)
-    #       - search res_users_settings_embedded_action (_format_settings)
-    #       - search res_lang_res_users_settings_rel (_format_settings)
-    #       - search im_livechat_expertise_res_users_settings_rel (_format_settings)
-    #   2: hasCannedResponses
-    #       - fetch res_groups_users_rel
-    #       - search mail_canned_response
-    _query_count_init_store = 19
-    # Queries for _query_count_init_messaging (in order):
-    #   1: insert res_device_log
-    #   3: _search_is_member (for current user, first occurence _search_is_member for chathub given channel ids)
-    #       - fetch res_users
-    #       - search discuss_channel_member
-    #       - fetch discuss_channel
-    #   1: search bus_bus (_bus_last_id)
-    #   1. search discuss_channel (chathub given channel ids)
-    #   1: channels_as_member
-    #   2: _init_messaging_global_fields (discuss)
-    #       - fetch discuss_channel_member (is_self)
-    #       - _compute_message_unread
-    #   4: _init_messaging (mail)
-    #       - search bus_bus (_bus_last_id)
-    #       - _get_needaction_count (inbox counter)
-    #       - search mail_message (starred counter)
-    #           - _check_access
-    #   23: _process_request_for_all (discuss):
-    #       - search discuss_channel (channels_domain)
-    #       22: store add channel:
-    #           - read group member (prefetch _compute_self_member_id from _compute_is_member)
-    #           - read group member (_compute_invited_member_ids)
-    #           - search discuss_channel_rtc_session
-    #           - fetch discuss_channel_rtc_session
-    #           - search member (channel_member_ids)
-    #           - fetch discuss_channel_member (manual prefetch)
-    #           10: member:
-    #               10: partner:
-    #                   - fetch res_partner (partner)
-    #                   - fetch res_users (_compute_im_status)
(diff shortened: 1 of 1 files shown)


seb-odoo (commented):
Hello, please keep the comment after parenthesis associated with each query as it helps knowing where it is coming from.

Also make sure to keep the comments for all the "groups", I see you have done it for some results like in `queries_init_store` but not for others.

kmagusiak:
> Hello, please keep the comment after parenthesis associated with each query as it helps knowing where it is coming from.
> 
> Also make sure to keep the comments for all the "groups", I see you have done it for some results like in `queries_init_store` but not for others.

Names of functions changed, order of queries changed. I tried to map existing comments to queries, but it quickly started not matching and I gave up.

seb-odoo:
Don't try to do it from the old diff, just do it from the actual run.

I usually just put a break point in execute when log is enabled -> check the stack trace and put as comment next to the query what I see, press next and repeat until the end.

addons/test_discuss_full/tests/test_performance.py
```diff
@@ -354,9 +212,59 @@ def test_fn():
...
+            'SELECT r.id FROM ir_rule ...',
+            'SELECT "rule_group_rel"."rule_group_id", "rule_group_rel"."group_id" FROM "res_groups" ...',
+            'SELECT ... FROM "ir_rule" WHERE ... %s',
+            'SELECT "res_company"."id", ... %s',
+            'SELECT "res_users"."id", ...',
+            'SELECT "res_groups"."id" FROM "res_groups" WHERE "res_groups"."id" IN %s',
+            # settings
+            'SELECT "res_users_settings"."id", ...',
+            'SELECT "res_users_settings_embedded_action"."id", ...',
+            'SELECT "res_users_settings"."id", ... %s',
```
seb-odoo, line 237: This is still missing some useful comments that were previously there.
Why is there 2 `select res_users_settings`? -> `_find_or_create_for_user` and other from `_format_settings`.
Also only showing `SELECT id` prevents for seeing what are `search` and what are `(search_)fetch` (which was previously commented).

addons/test_discuss_full/tests/test_performance.py
```diff
@@ -354,9 +212,59 @@ def test_fn():
             field_list = self.users[0].with_user(self.users[0])._store_init_global_fields
             return Store().add_global_values(field_list).get_result()
 
+        queries_init_store = [
+            # _store_init_global_fields
+            'SELECT "res_partner"."id" FROM ... %s',
+            'SELECT r.id FROM ir_rule ...',
```
seb-odoo, line 221: Previous asserts were based on a warmup run to avoid noise from ir_rules and reading groups.

Do we really want to make the tests now depend on them?

The goal of the test is to catch scaling issues in our code, not in the ORM code.

addons/test_discuss_full/tests/test_performance.py
seb-odoo: What is the difference between `...` and `%s`?

addons/test_discuss_full/tests/test_performance.py
```diff
@@ -365,12 +273,53 @@ def test_fn():
...
+            'SELECT "discuss_channel_member"... FROM "discuss_channel" ...',
+            'SELECT "discuss_channel"."id" FROM ...',
+            'SELECT "bus_bus"."id" FROM ... LIMIT 1',  # _bus_last_id
+            # _store_init_messaging_global_fields
+            'SELECT "discuss_channel_member"."id", ...',
+            'SELECT count(mail_message.id) AS count, ...',  # _compute_message_unread
+            'SELECT count(*) as needaction_count FROM ...',  # _get_needaction_count
+            'SELECT "mail_message"."id", ... GROUP BY ...',
+            'SELECT "discuss_channel"."id", ...',
+            'SELECT "discuss_channel"."id", ...',
```
seb-odoo: Missing comment, unclear why there is what looks like is a duplicate search.

seb-odoo (commented):
It's probably easier if I just push on the branch to add the missing comments that I find relevant.

Let me know when it's green and ready for you and I'll push my changes and merge it.

addons/test_discuss_full/tests/test_performance.py
kmagusiak: "..." is to match anything. The queries ending with "%s" are fetches that end like "blabla.id IN %s".

addons/test_discuss_full/tests/test_performance.py
kmagusiak: Acutally... assertQueries runs `login = self.env.user.login` just before starting to count the queries. That means that the queries to resolve the user and their permissions are not counted.
If you prefer, I can add explicitly `self.env.user.login` in `_run_test` and update the queries.

In my opinion, if the sole goal is to catch scaling issues, then comments of what is called exactly are overkill. In any case, if we change the ORM, we will update these.

seb-odoo: We don't want just the number, we want to easily see where they are coming from.

The thing is we absolutely don't care about the raw queries themselves.

We only care about what they represent and on which flows they are called.
If that was me, I'd rather have the test explicit the method name in the stack rather than the query.

kmagusiak: From what you write, I see no added value in explicit queries. I think that comments like these are just not in line with what really happens and they drift apart after each change/refactor.
I am going to close this.

seb-odoo: I think it has value, but not in the current shape.

A comment is not good as it's hard to maintain and doesn't enforce any guarantee, I agree with you.
But an actual query is too low level for business code.

In an ideal world, what I'd like to do is to "tag" queries with business meaning and assert that in the test.
In our code, we would set a tag on important "group" of queries (like the current groups in the comments), like an important method or a specific flow.
In the ORM, I would expect the tag to be like model name/method.
This would allow us to track what happens without having to work with too low level concepts.

kmagusiak: We would need to have a context variable to pass that tag around and it would be just an overhead in most cases.

Testing the queries (count or assertQueries) is already quite low-level.
I don't really understand the reason for this test: to me, it is either:
- we need query count to prevent the use cases from exploding the number of queries
- we need to know which queries are executed to detect changes in behaviour (and keep the number of queries always to a minimum)

> In an ideal world, what I'd like to do is to "tag" queries with business meaning and assert that in the test.

Low level: possible using custom SQL where you add a comment and try to find it back.
Writing a test as you suggest it will become just an abstraction to maintain on top of queries.

> In the ORM, I would expect the tag to be like model name/method.

Which method of the stack? There's multitude of combinations, just for example a compute method can call another compute method which can prefetch records.

PR #248694, opened 2026-02-13
[FIX] point_of_sale: kit using different UoM

robinengels:
When selling a kit product in POS, if the component of the kit use a different UoM than the UoM defined on the product, the stock valuation lines are wrong.

Steps to reproduce:
-------------------
* Create a storable product A with a UoM "Dozen" and a cost price of 10€
* Create a kit product B with a BoM of 1 unit of product A
* Sell 1 unit of product B in POS
> Observation: The valuation lines have the wrong value

Why the fix:
------------
The product qty was not considering the UoM when computing the expense and stock valuation lines.

opw-5471923


davidmonnom (approved):
@robodoo r+

PR #229057, opened 2025-09-29
[IMP] orm: improve prefetching

rco-odoo:
Keep the prefetching info in the result of methods `filtered()`, `records[int_or_slice]`, `concat()` and `union()`.

task 5123411

PR #248770, opened 2026-02-15
[FIX] orm: speedup getting prefetch ids

kmagusiak:
Various improvements to avoid non-linear growth of returned values when iterating over prefetches.  Traversing a relational field introduces a duplication factor in the returned values.

`records |= more` is syntactic sugar for `records = records | more`, which does not modify the recordset in place but makes one from scratch.  Therefore the time complexity is not `O(len(more))` but `O(len(records) + len(more))`, which is what makes it quadratic when used in a loop.

Since refactor in https://github.com/odoo/odoo/pull/229057, this quadratic complexity is introduced in the prefetching, too.

Changes:
- Keep same prefetch when accessing a single key as when iterating over a recordset. Small gain and more consistent results.
- When iterating over prefetches, deduplicate at source. Without this, the overhead of the generator can be huge for each returned value, since the same value might be returned many times.

In less trivial cases (like traversing two relational fields and making unions with it), the time speedup is several orders of magnitude.

### DIFF
--- a/odoo/orm/utils.py
+++ b/odoo/orm/utils.py
@@ -5,7 +5,7 @@ from collections.abc import Set as AbstractSet
 import dateutil.relativedelta
 
 from odoo.exceptions import ValidationError
-from odoo.tools import SQL
+from odoo.tools import SQL, unique
 
 regex_alphanumeric = re.compile(r'^[a-z0-9_]+$')
 regex_object_name = re.compile(r'^[a-z0-9_.]+$')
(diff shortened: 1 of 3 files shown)


kmagusiak:
@robodoo squash

odoo/orm/utils.py
```diff
@@ -197,35 +204,58 @@ def __init__(self, ids: tuple, *prefetches: Reversible):
...
+                hash(obj)
+                return obj
+            except TypeError:  # unhashable
+                pass
+            r = object.__new__(cls)
+            r.obj = obj
+            return r
+
+        def __eq__(self, value):
+            return self.obj == value.obj
```
rco-odoo: If you're mixing objects of that class with other objects, you better
```suggestion
        def __eq__(self, value):
            return isinstance(value, __AlwaysHash) and self.obj == value.obj
```

rco-odoo:
@robodoo r-

PR #248775, opened 2026-02-16
[IMP] hr_attendance: add specific time option for auto check-out

pamd-odoo:
Before:
- Auto check-out only supported tolerance mode
- No possibility to define a fixed time

After:
- Added specific time mode in settings
- Users can define a fixed check-out time
- Implemented a dedicated cron job to process check-outs at the configured time

Impact:
- More predictable automatic check-out behavior

Task: 5895265

Current behavior before PR:

---


addons/hr_attendance/data/hr_attendance_data.xml
```diff
@@ -18,5 +18,14 @@
...
             <field name="interval_type">hours</field>
         </record>
+
+        <record id="hr_attendance_check_out_specific_time_cron" model="ir.cron">
+            <field name="name">Attendance: Automatically check-out employees (Specific Time)</field>
+            <field name="model_id" ref="model_hr_attendance"/>
+            <field name="state">code</field>
+            <field name="code">model._cron_auto_check_out_specific_time()</field>
+            <field name="interval_number">10</field>
+            <field name="interval_type">minutes</field>
```
ravb-odoo: Is this cron run frequency really necessary? We could schedule it to run every 1 or 2 hours instead.

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -17,6 +17,7 @@
 from odoo.tools import convert, format_datetime, format_duration, format_time
 from odoo.tools.date_utils import sum_intervals
 from odoo.tools.intervals import Intervals
+from odoo.tools.date_utils import float_to_time
```
ravb-odoo: We have already imported odoo.tools.date_utils, so we can directly use it instead of creating a new import.

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -669,6 +671,46 @@ def _cron_absence_detection(self):
 
         to_unlink.unlink()
 
+    def _cron_auto_check_out_specific_time(self):
+        """
+        Automatically check-out all employees still checked in
+        when company is in 'specific_time' mode.
+        """
+        now_utc = fields.Datetime.now()
+        companies = self.env['res.company'].search([
```
ravb-odoo:
```suggestion
        current_utc_dt = fields.Datetime.now()
        companies_with_auto_checkout = self.env['res.company'].search([
```
Add proper variable name

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -669,6 +671,46 @@ def _cron_absence_detection(self):
...
+            ('auto_check_out', '=', True),
+            ('auto_check_out_mode', '=', 'specific_time'),
+        ])
+        for company in companies:
+            company_timezone = ZoneInfo(company.partner_id.tz or 'UTC')
+            now_local = now_utc.astimezone(company_timezone)
+            specific_time_obj = float_to_time(company.auto_check_out_specific_time)
```
ravb-odoo: Also give proper variable names

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -669,6 +671,46 @@ def _cron_absence_detection(self):
...
+            hours, minutes = specific_time_obj.hour, specific_time_obj.minute
+            open_attendances = self.search([
+                ('check_out', '=', False),
+                ('employee_id.company_id', '=', company.id),
+            ])
```
ravb-odoo: Instead of using search inside the loop, we should use read_group outside the loop to improve performance.

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -669,6 +671,46 @@ def _cron_absence_detection(self):
...
+            for att in open_attendances:
+                check_in_local = att.check_in.astimezone(company_timezone)
+                cutoff_same_day = check_in_local.replace(
```
ravb-odoo: same here for the variable name

addons/hr_attendance/models/res_company.py
```diff
@@ -5,6 +5,7 @@
 from odoo import fields, models, api
 from odoo.fields import Domain
 from odoo.tools.urls import urljoin as url_join
+from odoo.exceptions import ValidationError
```
ravb-odoo: should be in alphabetical order

addons/hr_attendance/models/res_company.py
```diff
@@ -47,6 +50,15 @@ def _compute_attendance_kiosk_url(self):
         for company in self:
             company.attendance_kiosk_url = url_join(self.env['res.company'].get_base_url(), '/hr_attendance/%s' % company.attendance_kiosk_key)
 
+    @api.constrains('auto_check_out', 'auto_check_out_mode', 'partner_id')
+    def _check_auto_checkout_timezone(self):
+        for company in self:
+            if company.auto_check_out and company.auto_check_out_mode == 'specific_time':
+                if not company.partner_id.tz:
```
ravb-odoo: Why do we need to check the company’s partner timezone? As far as I know, there is no tz field available in the res.partner form view, so how would the user set the timezone?

addons/hr_attendance/views/res_config_settings_views.xml
```diff
@@ -22,8 +22,14 @@
                         <setting string="Automatic Check-Out" company_dependent="1" help="Automatically Check-Out Employees based on their working schedule with an additional tolerance. Does not apply to employees with a flexible working schedule.">
                             <field name="auto_check_out"/>
                             <div invisible="not auto_check_out">
+                                <span class="me-2">Based on</span><field name="auto_check_out_mode" widget="radio" options="{'horizontal': true}"/>
+                            </div>
+                            <div invisible="not auto_check_out or auto_check_out_mode == 'specific_time'">
                                 <span class="me-2">Tolerance</span><field name="auto_check_out_tolerance" class="text-center" style="width: 5ch;"/><span class="ms-2">Hours</span>
                             </div>
+                            <div invisible="not auto_check_out or auto_check_out_mode == 'tolerance'">
+                                <span class="me-2">Time</span><field name="auto_check_out_specific_time" class="text-center" style="width: 5ch;" widget="float_time"/>
```
ravb-odoo: avoid using custom styles and use Bootstrap classes instead.

ravb-odoo (changes requested):
Thanks for your work. I’ve added some comments. Also, use proper variable names and add relevant test cases.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: sure, my mistake.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: okay i will change that.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: sure.

addons/hr_attendance/data/hr_attendance_data.xml
pamd-odoo: i have made it run every 2 hours

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: sure.

addons/hr_attendance/models/res_company.py
pamd-odoo: okay. will do that.

addons/hr_attendance/views/res_config_settings_views.xml
pamd-odoo: i would but for float_time the cleanest solution is with custom css. this solution is already used in many places with float_time in codebase.

ravb-odoo: I don’t understand, if Bootstrap already provides width utility classes, why are we using custom CSS?

And how would that be the cleanest solution specifically for float_time?

addons/hr_attendance/models/res_company.py
pamd-odoo: sure i have changed it to take timezone from company. which could be changed from config setting.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: read_group is for aggregations (counting/summing), not fetching actual records. Since we need to process each attendance individually, I've moved the search outside the loop instead. This reduces multiple DB queries to just one.

addons/hr_attendance/models/res_company.py
ravb-odoo: I don’t think the user can change the timezone from the company configuration, as the tz field is not present in the company form view. Additionally, the tz field on the company is mandatory, so there is no need to check it explicitly.

addons/hr_attendance/views/res_config_settings_views.xml
pamd-odoo: okay i will bootstrap utility classes.

addons/hr_attendance/views/res_config_settings_views.xml
```diff
@@ -19,10 +19,19 @@
...
                         </setting>
-                        <setting string="Automatic Check-Out" company_dependent="1" help="Automatically Check-Out Employees based on their working schedule with an additional tolerance. Does not apply to employees with a flexible working schedule.">
+                        <setting string="Automatic Check-Out" company_dependent="1" help="Automatically check out employees based on their working schedule with an additional tolerance, or at a specific time each day. Does not apply to employees with a flexible working schedule.">
                             <field name="auto_check_out"/>
                             <div invisible="not auto_check_out">
-                                <span class="me-2">Tolerance</span><field name="auto_check_out_tolerance" class="text-center" style="width: 5ch;"/><span class="ms-2">Hours</span>
+                                <span class="me-2">Based on</span><field name="auto_check_out_mode" widget="radio" options="{'horizontal': true}"/>
+                            </div>
+                            <div invisible="not auto_check_out or auto_check_out_mode == 'specific_time'">
```
ravb-odoo:
```suggestion
                            <div invisible="auto_check_out_mode != 'tolerance'">
```
There is no need to check the not auto_check_out condition, as we already have a div with the same condition.

Also, we should not check auto_checkout_out_mode specifically for specific_time. In the future, if someone adds another option, the tolerance field would incorrectly become visible for that option as well.

addons/hr_attendance/views/res_config_settings_views.xml
```diff
@@ -19,10 +19,19 @@
...
+                                <span class="me-2">Tolerance</span>
+                                <field name="auto_check_out_tolerance" class="text-center w-25"/>
+                                <span class="ms-2">Hours</span>
+                            </div>
+                            <div invisible="not auto_check_out or auto_check_out_mode == 'tolerance'">
```
ravb-odoo: Same here

addons/hr_attendance/models/res_company.py
```diff
@@ -47,6 +50,15 @@ def _compute_attendance_kiosk_url(self):
         for company in self:
             company.attendance_kiosk_url = url_join(self.env['res.company'].get_base_url(), '/hr_attendance/%s' % company.attendance_kiosk_key)
 
+    @api.constrains('auto_check_out_specific_time')
+    def _check_auto_checkout_specific_time_range(self):
+        for company in self:
+            if company.auto_check_out and company.auto_check_out_mode == 'specific_time':
+                if company.auto_check_out_specific_time < 0 or company.auto_check_out_specific_time >= 24:
```
ravb-odoo:
```suggestion
            if company.auto_check_out and company.auto_check_out_mode == 'specific_time' and not 0 <= company.auto_check_out_specific_time < 24:
```

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -647,6 +648,51 @@ def _cron_absence_detection(self):
...
+        companies_with_auto_checkout = self.env['res.company'].search([
+            ('auto_check_out', '=', True),
+            ('auto_check_out_mode', '=', 'specific_time'),
+        ])
+        if not companies_with_auto_checkout:
+            return
+        all_open_attendances = self.search([
+            ('check_out', '=', False),
+            ('employee_id.company_id', 'in', companies_with_auto_checkout.ids),
+        ])
```
ravb-odoo, line 709:
```suggestion
        all_open_attendances = self._read_group(
            domain=[
                ('check_out', '=', False),
                ('employee_id.company_id', 'in', companies_with_auto_checkout.ids),
            ],
            groupby=['company_id'],
            aggregates=['id:recordset'],
)
```

We can do something like this to avoid using filtered inside the loop.

addons/hr_attendance/data/hr_attendance_data.xml
```diff
@@ -18,5 +18,14 @@
...
         </record>
+
+        <record id="hr_attendance_check_out_specific_time_cron" model="ir.cron">
+            <field name="name">Attendance: Automatically check-out employees (Specific Time)</field>
+            <field name="model_id" ref="model_hr_attendance"/>
+            <field name="state">code</field>
+            <field name="code">model._cron_auto_check_out_specific_time()</field>
+            <field name="interval_number">2</field>
+            <field name="interval_type">hours</field>
+        </record>
```
ravb-odoo: A cron (hr_attendance_check_out_cron) is already created for the auto check-out purpose. So why not reuse that cron and merge your logic into it? We should avoid creating a new cron if it is not necessary.

Secondly, calling the cron (hr_attendance_check_out_cron) every 4 hours may not be required. In my opinion, running it 3 or 4 times a day should be sufficient.

ravb-odoo (changes requested):
Here are some comments.

addons/hr_attendance/data/hr_attendance_data.xml
pamd-odoo: I’d prefer to keep this as a separate cron. The existing one is focused on tolerance and calendar-based logic, while this is a fixed wall-clock cutoff. Merging both would mix two different behaviors in one method and add unnecessary branching. Keeping them separate keeps responsibilities clear and easier to maintain.

I agree on the frequency though — I can reduce it to every 8 hours since the logic safely checks the cutoff time before applying anything.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: Although the performance gain in minimal with using _read_group, i have made that change for worst case scenario. (ie. thousands of open attendance records.)

addons/hr_attendance/models/res_company.py
pamd-odoo: Sure. will do this.

addons/hr_attendance/views/res_config_settings_views.xml
pamd-odoo: okay will make this small change.

addons/hr_attendance/models/hr_attendance.py
ravb-odoo: In terms of performance, _read_group is generally preferable over search() followed by filtering in a loop when dealing with large datasets. Using search() and then applying .filtered() will have more time complexity compared to _read_group, especially as the data volume increases.

However, in this case, I don’t think we can use _read_group because company_id is not a stored field on hr.attendance. Also, we can’t reliably use employee_id.company_id in the groupby. I’m not sure about this, but you can try it once to confirm.

addons/hr_attendance/data/hr_attendance_data.xml
ravb-odoo: Creating a new cron for this purpose would add unnecessary load to the server. Generally, cron jobs are not meant to separate logic for a similar flow. Since a cron already exists for the same purpose, we should reuse it instead of creating a new one.

If you want to keep the logic separate, you can create two different methods and call them based on the selection from a single cron. However, please ensure that there is no code duplication.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: I tried using _read_group, but grouping on employee_id.company_id is not supported because dot notation cannot be used in groupby, and company_id is not a direct stored field on hr.attendance. Because of that, I reverted back to search() with filtering

addons/hr_attendance/data/hr_attendance_data.xml
pamd-odoo: I merged the logic into the existing cron to avoid introducing an additional scheduled job.

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -555,13 +555,18 @@ def action_refuse_overtime(self):
         self.linked_overtime_ids.action_refuse()
 
     def _cron_auto_check_out(self):
+        self._cron_auto_check_out_tolerance()
+        self._cron_auto_check_out_specific_time()
```
yava-odoo, line 599: Not sure that we should merge these two cron

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -636,6 +636,347 @@ def test_auto_check_out_lunch_period(self):
             self.assertEqual(morning.worked_hours + afternoon.worked_hours, 9)  # 8 hours from calendar's attendances + 1 hour of tolerance
             self.assertEqual(afternoon.check_out, datetime(2024, 1, 1, 18, 0))
 
+    @freeze_time("2026-02-26 07:00:00")
```
yava-odoo, line 693: For the same specific_time you can create multiple attendances in a single test case (merge multiple test cases)


Instead of calling the write method everytime you can directly add this in the setup

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -636,6 +636,347 @@ def test_auto_check_out_lunch_period(self):
...
+    def test_auto_check_out_specific_time_normal_night_shift(self):
+        """Night shift: Check-in yesterday 11 PM, checkout today 6 AM"""
```
yava-odoo: Instead of AM/PM, use 24 hour format in the comments so it will be easy for us to get the idea of the case

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -636,6 +636,347 @@ def test_auto_check_out_lunch_period(self):
...
+    def test_auto_check_out_specific_time_before_cutoff(self):
+        """Check-in today 2 AM, checkout today 6 AM"""
+        self.company.write({
+            'auto_check_out': True,
+            'auto_check_out_mode': 'specific_time',
+            'auto_check_out_specific_time': 6.0,
+        })
+        attendance = self.env['hr.attendance'].create({
+            'employee_id': self.employee.id,
+            'check_in': datetime(2026, 2, 25, 20, 30),
```
yava-odoo: In the comment, you mentioned 2 AM, but in the datetime it's actually 8 PM?

yava-odoo (changes requested):
Thanks for the work!
Here are some comments 

Can you update the commit msg as well with the edge case of this cron?

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: Because having different cron will affect performance and unnecessary put more load on server. that is why i have called both method in single cron.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: This test file also contains tests for the tolerance mode (_cron_auto_check_out) which require a different configuration (auto_check_out_mode = 'tolerance').
Because of that I kept the configuration inside each specific_time test to avoid affecting the tolerance mode tests.

Regarding merging multiple scenarios into a single test: since an employee cannot have multiple open attendances, it would require creating additional employees for each case. I kept them separated to keep the scenarios clearer and easier to read.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: okay will do that.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: okay small mistake. i will correct it.

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -665,6 +670,51 @@ def _cron_absence_detection(self):
...
+        companies_with_auto_checkout = self.env['res.company'].search([
+            ('auto_check_out', '=', True),
+            ('auto_check_out_mode', '=', 'specific_time'),
+        ])
+        if not companies_with_auto_checkout:
+            return
+        all_open_attendances = self.search([
+            ('check_out', '=', False),
+            ('employee_id.company_id', 'in', companies_with_auto_checkout.ids),
+        ])
```
mepe-odoo:
```suggestion
        all_open_attendances = self.search([
            ('check_out', '=', False),
            ('employee_id.company_id', 'any', [
                ('auto_check_out', '=', True),
                ('auto_check_out_mode', '=', 'specific_time'),
            ]),
        ])
```

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -665,6 +670,51 @@ def _cron_absence_detection(self):
...
+        for company in companies_with_auto_checkout:
+            company_tz = ZoneInfo(company.tz or 'UTC')
+            current_company_dt = current_utc_dt.astimezone(company_tz)
+            cutoff_time = float_to_time(company.auto_check_out_specific_time)
+            cutoff_hour, cutoff_minute = cutoff_time.hour, cutoff_time.minute
+            company_attendances = all_open_attendances.filtered(
+                lambda a: a.employee_id.company_id == company
+            )
+            for att in company_attendances:
```
mepe-odoo:
```suggestion
        for company, company_attendances in all_open_attendances.grouped('company_id'):
            company_tz = ZoneInfo(company.tz or 'UTC')
            current_company_dt = current_utc_dt.astimezone(company_tz)
            cutoff_time = float_to_time(company.auto_check_out_specific_time)
            cutoff_hour, cutoff_minute = cutoff_time.hour, cutoff_time.minute
            company_attendances = all_open_attendances.filtered(
                lambda a: a.employee_id.company_id == company
            )
            for att in company_attendances:
```

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -665,6 +670,51 @@ def _cron_absence_detection(self):
...
+                check_in_company_dt = att.check_in.astimezone(company_tz)
```
mepe-odoo: the check_in should be timezoned with the employee_tz and not the company_tz :)

addons/hr_attendance/models/hr_attendance.py
```diff
@@ -665,6 +670,51 @@ def _cron_absence_detection(self):
...
+                same_day_cutoff_dt = check_in_company_dt.replace(
+                    hour=cutoff_hour, minute=cutoff_minute, second=0, microsecond=0
+                )
+                if check_in_company_dt.time() < same_day_cutoff_dt.time():
+                    next_cutoff = same_day_cutoff_dt
+                else:
+                    next_cutoff = same_day_cutoff_dt + relativedelta(days=1)
+                if current_company_dt < next_cutoff:
+                    continue
+                employee_checkout = next_cutoff.astimezone(ZoneInfo('UTC')).replace(tzinfo=None)
```
mepe-odoo: the employee checkout needs to be checked with the employee tz

addons/hr_attendance/models/res_company.py
```diff
@@ -48,6 +51,14 @@ def _compute_attendance_kiosk_url(self):
         for company in self:
             company.attendance_kiosk_url = url_join(self.env['res.company'].get_base_url(), '/hr_attendance/%s' % company.attendance_kiosk_key)
 
+    @api.constrains('auto_check_out_specific_time')
+    def _check_auto_checkout_specific_time_range(self):
+        for company in self:
+            if company.auto_check_out and company.auto_check_out_mode == 'specific_time' and not 0 <= company.auto_check_out_specific_time < 24:
+                raise ValidationError(self.env._(
+                    "Specific Time must be within a 24-hour range (0h 0m 0s to 23h 59m 59s).",
+                ))
```
mepe-odoo: hum I think you can do an sql constraint like
'''CHECK(
            (auto_check_out_mode='specific_time' AND auto_check_out_specific_time BETWEEN 0 AND 23.99) OR auto_check_out_mode='tolerance' 
        )'''
:)

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: okay i have changed it after po's comment on task.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: Yes it will be based on employee tz.

addons/hr_attendance/models/res_company.py
pamd-odoo: sure if you prefer sql constrain i would change it.

addons/hr_attendance/models/hr_attendance.py
pamd-odoo: sure but there is no field called company_id in attendance. so i have grouped it by employee company_id.

mepe-odoo: so grouped(lambda a: employee.company_id) :)

addons/hr_attendance/models/res_company.py
mepe-odoo: It's nicer to do it in sql because you will have a real constraint in python you can still bypass it :)

addons/hr_attendance/models/res_config_settings.py
```diff
@@ -22,7 +22,9 @@ class ResConfigSettings(models.TransientModel):
     attendance_overtime_validation = fields.Selection(related="company_id.attendance_overtime_validation", readonly=False)
     auto_check_out = fields.Boolean(related="company_id.auto_check_out", readonly=False)
     single_check_in = fields.Boolean(related="company_id.single_check_in", readonly=False)
+    auto_check_out_mode = fields.Selection(related="company_id.auto_check_out_mode", readonly=False)
     auto_check_out_tolerance = fields.Float(related="company_id.auto_check_out_tolerance", readonly=False)
+    auto_check_out_specific_time = fields.Float(related="company_id.auto_check_out_specific_time", readonly=False)
```
mepe-odoo, line 27: you should also have a constraint on this field as you can write on it :)

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -690,6 +690,123 @@ def test_auto_check_out_lunch_period(self):
...
+            {'employee_id': self.future_contract_employee.id, 'check_in': datetime(2026, 2, 26, 1, 0), 'check_out': datetime(2026, 2, 26, 4, 30)},
+            {'employee_id': self.flexible_employee.id, 'check_in': datetime(2026, 2, 26, 1, 0)},
+        ])
+        initial_checkout = already_checked.check_out
+        self.assertEqual(night_shift.check_out, False)
+        self.env['hr.attendance']._cron_auto_check_out_specific_time()
+        self.assertEqual(night_shift.check_out, datetime(2026, 2, 26, 6, 0))
+        self.assertEqual(night_shift.out_mode, 'auto_check_out')
+        self.assertEqual(before_cutoff.check_out, datetime(2026, 2, 26, 6, 0))
+        self.assertEqual(one_min_before.check_out, datetime(2026, 2, 25, 21, 0))
```
mepe-odoo, line 717: can you add a comment to explain that 9PM in utc is equal to 6AM in Japanese time please ? :)

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -690,6 +690,123 @@ def test_auto_check_out_lunch_period(self):
...
+        self.assertEqual(midnight.check_out, datetime(2026, 2, 25, 16, 0))
```
mepe-odoo, line 719: same for this one :)

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -690,6 +690,123 @@ def test_auto_check_out_lunch_period(self):
...
+        after_cutoff, exactly_at_cutoff, minimum_duration = self.env['hr.attendance'].create([
+            {'employee_id': self.employee.id, 'check_in': datetime(2026, 2, 26, 6, 0)},
+            {'employee_id': self.other_employee.id, 'check_in': datetime(2026, 2, 26, 6, 0)},
+            {'employee_id': self.jpn_employee.id, 'check_in': datetime(2026, 2, 25, 21, 0)},
+        ])
+        self.env['hr.attendance']._cron_auto_check_out_specific_time()
+        self.assertEqual(after_cutoff.check_out, datetime(2026, 2, 27, 6, 0))
+        self.assertEqual(exactly_at_cutoff.check_out, datetime(2026, 2, 27, 6, 0))
+        self.assertTrue(minimum_duration.check_out)
+        self.assertGreater(minimum_duration.check_out, minimum_duration.check_in)
```
mepe-odoo: you already do this check in your various test no ? 🤔

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -690,6 +690,123 @@ def test_auto_check_out_lunch_period(self):
...
+            'company_id': brussels_company.id,
+            'tz': 'Europe/Brussels',
+        })
+        att_main, att_brussels = self.env['hr.attendance'].create([
+            {'employee_id': self.employee.id, 'check_in': datetime(2026, 2, 26, 1, 0)},
+            {'employee_id': brussels_employee.id, 'check_in': datetime(2026, 2, 26, 7, 0)},
+        ])
+        self.env['hr.attendance']._cron_auto_check_out_specific_time()
+        self.assertEqual(att_main.check_out, datetime(2026, 2, 26, 6, 0))
+        self.assertEqual(att_brussels.check_out, datetime(2026, 2, 26, 9, 0))
```
mepe-odoo: same for this one, you test for honolulu and japanese timezone I think it's enough :)

addons/hr_attendance/tests/test_hr_attendance_overtime.py
```diff
@@ -690,6 +690,123 @@ def test_auto_check_out_lunch_period(self):
...
+            'auto_check_out_mode': 'specific_time',
+            'auto_check_out_specific_time': 23.98,
+        })
+        end_of_day = self.env['hr.attendance'].create({
+            'employee_id': self.employee.id,
+            'check_in': datetime(2026, 2, 26, 4, 30),
+        })
+        self.env['hr.attendance']._cron_auto_check_out_specific_time()
+        self.assertEqual(end_of_day.check_out, datetime(2026, 2, 26, 23, 59))
+        with freeze_time("2026-02-26 07:00:00"):
```
mepe-odoo, line 756: it's strange to go back in the past :/

addons/hr_attendance/models/res_config_settings.py
pamd-odoo: The field is related to company_id.auto_check_out_specific_time, so writes on it still go through res.company and are already covered by the SQL constraint there. I don't see the need for adding constraint again.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: sure no issue. i i wlll add comment for it.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: sure no issue. i i wlll add comment for it.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: This test was mainly added to explicitly cover the scenario where the check-in happens exactly at or after the specific time, so the automatic checkout should happen on the next day at the configured time.

Some overlap with the other tests definitely exists though, so I can remove it if you think it is redundant

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: Makes sense. I removed this test since the Japanese and Honolulu timezone cases already cover the employee timezone behavior sufficiently.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
pamd-odoo: I kept it in a single test since both edge cases are related to the same behavior. The nested `freeze_time` is mainly to avoid duplicating the setup/assertions in two separate tests.

addons/hr_attendance/tests/test_hr_attendance_overtime.py
mepe-odoo: yes if it's redundant please remove it we try to reduce the amount of tests (but we still want to cover everything) for the runbot :)

pamd-odoo: okay i believe it is redundant. so i have removed it.

mepe-odoo:
@robodoo r+

PR #248778, opened 2026-02-16
[IMP] mrp: reorder BoM and quantity fields in MO form view

dhha-odoo:
Reordered the BoM and quantity fields in the MO form view so that quantity
is the final input.

TaskID:- 5864744

### DIFF
--- a/addons/mrp/views/mrp_production_views.xml
+++ b/addons/mrp/views/mrp_production_views.xml
@@ -335,6 +335,22 @@
                             <field name="forecasted_issue" invisible="1"/>
                             <field name="company_id" invisible="1"/>
                             <field name="product_description_variants" invisible="product_description_variants in (False, '')" readonly="state != 'draft'"/>
+                            <label for="bom_id" name="bom_label"/>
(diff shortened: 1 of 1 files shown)


addons/mrp/models/mrp_production.py
```diff
@@ -449,9 +449,10 @@ def _compute_product_qty(self):
         for production in self:
             if production.state != 'draft':
                 continue
-            if production.bom_id and production._origin.bom_id != production.bom_id:
+            has_bom = production.bom_id
+            if not production._origin.bom_id and has_bom:
                 production.product_qty = production.bom_id.product_qty
-            elif not production.bom_id:
+            elif not has_bom:
                 production.product_qty = 1.0
```
tpma-odoo: Here, removing the BoM still `resets` the `quantity to 1.0`, which can override a quantity explicitly set by the user. For example, in MTO generated mo, the quantity comes from the sale order and should not be altered when the BoM is unset. The quantity should remain unchanged in this case as well.

svs-odoo:
robodoo r+

PR #248779, opened 2026-02-16
[FIX] sale: remove duplicated Delivery Date in SO report preview

khsr-odoo:
**Steps to reproduce:**
1. Install modules sale_management
2. Go to Settings, Configure Document Layout and set layout to DIN 5008
3. Create a new Sale Order
4. Set a customer, add a product, and fill in the Delivery Date (Other Info)
5. Click on Print and Preview

**Issue:**
The Delivery Date (commitment_date) is printed twice in the report.

**Solution:**
Remove duplicated Delivery Date entry from this module to avoid rendering it twice.

Issues : https://github.com/odoo/odoo/issues/247522

**opw-5490651**


Feyensv (approved):
robodoo r+ delegate+

PR #240683, opened 2025-12-19
[IMP] core, *: split odoo.http module

Julien00859:
The odoo.http module is a huge file with thousands of lines. In this work we split it in many smaller chunks. The desire to split odoo.http has been in the air for quite some time, notably with the introduction of the many facade objects.

This work only moves the constants, functions and classes in new dedicated odoo.http-submodules. In order to make the diff easier to gasp, we decided *not* to do other changes such as moving methods around. That's job for a future work.

We decided to re-expose under the "odoo.http" namespace only the four Controller, route, Response and request. All the other symbols are much less so used and are to be imported from their dedicated module.

`content_disposition` is kinda a special one, it is often imported by controllers that load an entire file in memory before sending it. We are opinated and think it would be a better idea to stream to files in order to keep memory usage in check. We moved the function next to the Stream class so the latter can be discovered.

PR #248785, opened 2026-02-16
[FIX] web: fix SessionExpiredException handling after http split

rame-odoo:
Steps to Reproduce:
1) Open Attendance app > Configuration > Settings
2) Open the Kiosk URL in another window
3) Go back to module in old window and try to interect with UI 4) A traceback appears instead of a proper session expired message

Issue:
- A traceback was shown when a session expired after switching kiosk configuration.
- After the recent odoo.http refactor, the exception path of SessionExpiredException changed from odoo.http.SessionExpiredException to odoo.http.session.SessionExpiredException.
- The web error registry still used the old path, causing full RPC tracebacks on session expiry instead of the expected session expired dialog/notification.
- Ref: PR #240683

Fix:
- Updated the web error registry to handle the new exception path odoo.http.session.SessionExpiredException, restoring the expected session expired behavior and preventing traceback.

task-5936753

### DIFF
--- a/addons/web/static/src/core/errors/error_dialogs.js
+++ b/addons/web/static/src/core/errors/error_dialogs.js
@@ -229,6 +229,6 @@ registry
     .add("odoo.exceptions.UserError", WarningDialog)
     .add("odoo.exceptions.ValidationError", WarningDialog)
     .add("odoo.exceptions.RedirectWarning", RedirectWarningDialog)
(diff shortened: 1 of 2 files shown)


kmagusiak (approved):

Einar-eian (approved):
Thank you for your work! LGTM 💪 

@robodoo r+

PR #248786, opened 2026-02-16
[IMP] {test_}mail: improve activity handling and add reuse support

adsh-odoo:
This PR makes the following improvements for the activities.

- Remove the reschedule button from the done activities.
- Add a rainbow_man effect for empty systray opening.
- Add support to reuse the done activity from the posted message record.
- Add the mail_activity_id field on the message, as earlier the activity
  and message were never connected so this feature will not work for the older
  messages.

Task-5344015

### DIFF
--- a/addons/mail/models/mail_message.py
+++ b/addons/mail/models/mail_message.py
@@ -228,6 +228,7 @@ class MailMessage(models.Model):
     mail_activity_type_id = fields.Many2one(
         'mail.activity.type', 'Mail Activity Type',
         index='btree_not_null', ondelete='set null')
+    mail_activity_id = fields.Many2one('mail.activity', 'Mail Activity', index='btree_not_null')
     is_internal = fields.Boolean('Employee Only', help='Hide to public / portal users, independently from subtype configuration.')
     # origin
     email_from = fields.Char('From', help="Email address of the sender. This field is set when no matching partner is found and replaces the author_id field in the chatter.")
@@ -1183,6 +1184,7 @@ class MailMessage(models.Model):
         res.attr("subject")
         # sudo: mail.message.subtype - reading subtype on accessible message is allowed
         res.one("subtype_id", ["description"], sudo=True)
+        res.one("mail_activity_id", ["activity_type_id", "summary", "note"], predicate=lambda m: m.mail_activity_id)
         res.attr("write_date")
         self._store_linked_messages_fields(res)
         self._store_message_link_previews_fields(res)
--- a/addons/mail/static/src/core/web/activity_list_popover.js
+++ b/addons/mail/static/src/core/web/activity_list_popover.js
@@ -51,6 +51,8 @@ export class ActivityListPopover extends Component {
                 this.props.resModel,
                 this.props.resIds ? this.props.resIds : [this.props.resId],
                 this.props.defaultActivityTypeId
+                    ? { default_activity_type_id: this.props.defaultActivityTypeId }
+                    : {}
             )
             .then(() => this.props.onActivityChanged());
         this.props.close();
--- a/addons/mail/static/src/core/web/activity_model_patch.js
+++ b/addons/mail/static/src/core/web/activity_model_patch.js
@@ -75,7 +75,15 @@ patch(Activity.prototype, {
         return action;
     },
     remove({ broadcast = true } = {}) {
-        this.delete();
+        // Only delete the activity if no message is linked to it.
+        // We keep it when marked as done because reuse action needs mail_activity_id.
+        if (
+            !Object.values(this.store["mail.message"].records).some(
+                (m) => m.mail_activity_id?.id === this.id
+            )
+        ) {
+            this.delete();
+        }
         if (broadcast) {
             this.activityBroadcastChannel?.postMessage({
                 type: "DELETE",
--- a/addons/mail/static/src/core/web/store_service_patch.js
+++ b/addons/mail/static/src/core/web/store_service_patch.js
@@ -89,16 +89,14 @@ const StorePatch = {
     /**
      * @param {string} resModel
      * @param {number[]} resIds
-     * @param {number|undefined} defaultActivityTypeId
+     * @param {Object} [defaultContextValues] default values to pass in the activity form view.
      */
-    async scheduleActivity(resModel, resIds, defaultActivityTypeId = undefined) {
+    async scheduleActivity(resModel, resIds, defaultContextValues = {}) {
         const context = {
             active_model: resModel,
             active_ids: resIds,
             active_id: resIds[0],
-            ...(defaultActivityTypeId !== undefined
-                ? { default_activity_type_id: defaultActivityTypeId }
-                : {}),
+            ...defaultContextValues,
         };
         await new Promise((resolve) =>
             this.env.services.action.doAction(
--- a/addons/test_mail/tests/test_mail_activity.py
+++ b/addons/test_mail/tests/test_mail_activity.py
@@ -261,6 +261,7 @@ class TestActivityFlow(TestActivityCommon):
 
     def test_activity_flow_employee(self):
         with self.with_user('employee'):
+            activity_type_id = self.env.ref('mail.mail_activity_data_email')
             test_record = self.env['mail.test.activity'].browse(self.test_record.id)
             self.assertEqual(test_record.activity_ids, self.env['mail.activity'])
 
@@ -268,7 +269,7 @@ class TestActivityFlow(TestActivityCommon):
             activity = self.env['mail.activity'].create({
                 'summary': 'Test Activity',
                 'date_deadline': date.today() + relativedelta(days=1),
-                'activity_type_id': self.env.ref('mail.mail_activity_data_email').id,
+                'activity_type_id': activity_type_id.id,
                 'res_model_id': self.env['ir.model']._get(test_record._name).id,
                 'res_id': test_record.id,
             })
@@ -285,7 +286,10 @@ class TestActivityFlow(TestActivityCommon):
             activity.action_feedback(feedback='So much feedback')
             self.assertEqual(activity.feedback, 'So much feedback')
             self.assertEqual(test_record.activity_ids, self.env['mail.activity'])
-            self.assertEqual(test_record.message_ids[0].subtype_id, self.env.ref('mail.mt_activities'))
+            last_message = test_record.message_ids[0]
+            self.assertEqual(last_message.subtype_id, self.env.ref('mail.mt_activities'))
+            self.assertEqual(last_message.mail_activity_type_id, activity_type_id)
+            self.assertEqual(last_message.mail_activity_id, activity)
 
     @mute_logger('odoo.addons.mail.models.mail_mail')
     def test_activity_notify_other_user(self):
--- a/addons/mail/models/mail_activity.py
+++ b/addons/mail/models/mail_activity.py
@@ -628,6 +628,7 @@ class MailActivity(models.Model):
                             'feedback': feedback,
                             'display_assignee': activity.user_id != self.env.user
                         },
+                        mail_activity_id=activity.id,
                         mail_activity_type_id=activity.activity_type_id.id,
                         subtype_xmlid='mail.mt_activities',
                     )
--- a/addons/mail/models/mail_thread.py
+++ b/addons/mail/models/mail_thread.py
@@ -3132,6 +3132,7 @@ class MailThread(models.AbstractModel):
             'incoming_email_to',
             'is_internal',
             'mail_activity_type_id',
+            'mail_activity_id',
             'mail_server_id',
             'message_id',
             'message_type',
--- a/addons/mail/models/res_users.py
+++ b/addons/mail/models/res_users.py
@@ -499,7 +499,7 @@ class ResUsers(models.Model):
             model = self.env["ir.model"]._get(model_name).with_prefetch(model_ids)
             user_activities[model_name] = {
                 "id": model.id,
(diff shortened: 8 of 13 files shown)


addons/mail/static/src/core/common/message_actions.js
```diff
@@ -213,6 +213,13 @@ registerMessageAction("end-poll", {
     onSelected: ({ message }) => rpc("/mail/poll/end", { poll_id: message.poll.id }),
     sequence: 115,
 });
+registerMessageAction("reuse-activity", {
+    condition: ({ message }) => message.isActivity,
+    icon: "fa fa-recycle",
+    name: _t("Reuse"),
+    onSelected: ({ message }) => message.reuseActivity(),
+    sequence: 125,
+});
```
mrsr-odoo: I think we can move all the code from `mail/static/src/core/common` to `mail/static/src/core/web` as it web feature not related to discuss part

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -91,14 +91,21 @@ const StorePatch = {
      * @param {number[]} resIds
      * @param {number|undefined} defaultActivityTypeId
      */
-    async scheduleActivity(resModel, resIds, defaultActivityTypeId = undefined) {
+    async scheduleActivity(
+        resModel,
+        resIds,
+        defaultActivityTypeId = undefined,
+        predifinedContext = undefined,
```
mrsr-odoo: we can merge these two into single context object `{}`

addons/mail/static/src/core/web/activity_model_patch.js
```diff
@@ -65,8 +65,10 @@ patch(Activity.prototype, {
         });
         return action;
     },
-    remove({ broadcast = true } = {}) {
-        this.delete();
+    remove(deleteActivity = true, { broadcast = true } = {}) {
+        if (deleteActivity) {
+            this.delete();
+        }
```
mrsr-odoo: why do we need this one ?

addons/mail/static/src/core/common/message_model.js
```diff
@@ -633,6 +638,22 @@ export class Message extends Record {
         this.store.env.services.notification.add(notification, { type });
     }
 
+    async reuseActivity() {
+        const context = {
+            default_activity_type_id: this.mail_activity_id.activity_type_id.id,
+            summary: this.mail_activity_id.summary,
```
mrsr-odoo:
```suggestion
            default_summary: this.mail_activity_id.summary,
``` 
?

mrsr-odoo (commented):
Hello @adsh-odoo ,
Left few comments for js part
Thanks

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -115,7 +122,10 @@ const StorePatch = {
                     context,
                 },
                 {
-                    onClose: resolve,
+                    onClose: () => {
+                        resolve();
+                        thread?.fetchThreadData(["activities", "messages"]);
+                    },
```
mrsr-odoo: let's keep it unchanged

addons/mail/static/src/core/common/message_model.js
```diff
@@ -633,6 +638,22 @@ export class Message extends Record {
...
+            default_user_id: this.mail_activity_id.user_id?.id,
+            default_note: this.mail_activity_id.note,
+        };
+        this.store.scheduleActivity(
+            this.thread.model,
+            [this.thread.id],
+            undefined,
+            context,
+            this.thread
+        );
```
mrsr-odoo: and maybe we can use `fetchThreadData` here

addons/mail/models/mail_message.py
```diff
@@ -194,6 +194,7 @@ def default_get(self, fields):
     mail_activity_type_id = fields.Many2one(
         'mail.activity.type', 'Mail Activity Type',
         index='btree_not_null', ondelete='set null')
```
mrsr-odoo, line 230: to check why do need this field?

addons/calendar/models/res_users.py
```diff
@@ -164,8 +168,9 @@ def _get_activity_groups(self):
                 'model': 'calendar.event',
                 'icon': modules.module.get_module_icon(EventModel._original_module),
                 'domain': [('active', 'in', [True, False])],
-                'meetings': meetings_lines,
+                'meetings': systray_meetings_lines,
                 "view_type": EventModel._systray_view,
+                "remaining_meetings":  len(meetings_lines) - len(systray_meetings_lines) if len(meetings_lines) - len(systray_meetings_lines) > 0 else 0
```
mrsr-odoo: when can we have negative value for `len(meetings_lines) - len(systray_meetings_lines)

addons/calendar/models/res_users.py
```diff
@@ -155,6 +155,10 @@ def _get_activity_groups(self):
             ['id', 'start', 'name', 'allday'],
             order='start')
         if meetings_lines:
+            all_day_meetings = [meeting_line for meeting_line in meetings_lines if meeting_line['allday']]
+            not_all_day_meetings = [meeting_line for meeting_line in meetings_lines if not meeting_line['allday']]
+            systray_meetings_lines = not_all_day_meetings[:2] + all_day_meetings[:1]
+
```
mrsr-odoo: can't we just show first three activities(without all day filter) ? so we only fetch first three records

addons/calendar/static/src/core/web/activity_menu_patch.xml
```diff
@@ -16,6 +16,11 @@
                         </span>
                     </div>
                 </t>
+                <div t-if="group.remaining_meetings" class="o-calendar-meeting d-flex px-2">
```
mrsr-odoo:
```suggestion
                <div t-if="group.remaining_meetings" class="p-2">
```
to add horizontal padding as well

mrsr-odoo (commented):
Hello @adsh-odoo ,
And the rest of the comments
Thanks

addons/calendar/models/res_users.py
adsh-odoo: I made the syntax complex, which caused this misunderstanding.

addons/mail/static/src/core/web/activity_model_patch.js
adsh-odoo: This method is called when we mark the activity done or when we cancel the activity. This delete method removes the activity from the store. We want to keep the activity in the store when it is marked done, so that's why we added this condition here to differentiate when to delete from the store and when not.

addons/mail/models/mail_message.py
adsh-odoo: We are using this field in the `crm.activity.report` view directly from the `mail_message` itself. If we want to remove this field we have to make it accessible via `mail_activity__id.mail_activity_type_id` but in that case we will loose the track of the previous records as mail_activity was not connected to mail_message before thhis commit.

addons/calendar/models/res_users.py
adsh-odoo: We can but showing the both types of meeting gives users a better overview of their schedule.
If we just show the first three records, the list might become biased toward one type.

adsh-odoo:
Hello @mrsr-odoo 
I have made the changes and answered your questions. Can you please have a look?
Thanks!!

addons/mail/static/src/core/common/thread_model.js
```diff
@@ -66,7 +66,7 @@ export class Thread extends Record {
     autofocus = 0;
     activities = fields.Many("mail.activity", {
         sort: (a, b) => compareDatetime(a.date_deadline, b.date_deadline) || a.id - b.id,
-        onDelete: (r) => r?.remove(),
+        onDelete: (r) => r?.remove(false),
```
mrsr-odoo: we can remove `onDelete` if we don't want it to delete ?

addons/mail/static/src/core/web/activity_list_popover.js
```diff
@@ -50,7 +50,9 @@ export class ActivityListPopover extends Component {
             .scheduleActivity(
                 this.props.resModel,
                 this.props.resIds ? this.props.resIds : [this.props.resId],
-                this.props.defaultActivityTypeId
+                ...(this.props.defaultActivityTypeId
+                    ? [{ default_activity_type_id: this.props.defaultActivityTypeId }]
+                    : [])
```
mrsr-odoo:
```suggestion
                this.props.defaultActivityTypeId
                    ? { default_activity_type_id: this.props.defaultActivityTypeId }
                    : {}
```

addons/mail/static/src/core/web/activity_model_patch.js
mrsr-odoo: custom flag for deleting doesn't look good here as this function is intent to delete it (see other comment if it is only used once)
and as per below code if multiple tab with same record are open this function will be called recursively https://github.com/odoo/odoo/blob/c8a1d17b873ddc381351cefdd00c069eaf970024/addons/mail/static/src/core/web/store_service_patch.js#L143 with current flag

addons/mail/static/src/core/web/activity_model_patch.js
```diff
@@ -65,8 +65,10 @@ patch(Activity.prototype, {
...
         if (broadcast) {
             this.activityBroadcastChannel?.postMessage({
```
mrsr-odoo, line 88: not your code but it is not working currently 🙁  maybe `this.store.activityBroadcastChannel` might work (see `markAsDone` above)?
should report to related team

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -89,16 +89,14 @@ const StorePatch = {
     /**
      * @param {string} resModel
      * @param {number[]} resIds
-     * @param {number|undefined} defaultActivityTypeId
+     * @param {Object|undefined} [predifinedContext] Context values to pass to the activity form view as default_*
      */
-    async scheduleActivity(resModel, resIds, defaultActivityTypeId = undefined) {
+    async scheduleActivity(resModel, resIds, predifinedContext = undefined) {
```
mrsr-odoo:
```suggestion
    async scheduleActivity(resModel, resIds, defaultContextValues = {}) {
```

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -89,16 +89,14 @@ const StorePatch = {
...
         const context = {
             active_model: resModel,
             active_ids: resIds,
             active_id: resIds[0],
-            ...(defaultActivityTypeId !== undefined
-                ? { default_activity_type_id: defaultActivityTypeId }
-                : {}),
+            ...(predifinedContext !== undefined ? { ...predifinedContext } : {}),
```
mrsr-odoo:
```suggestion
            ...defaultContextValues,
```

addons/mail/static/src/core/web/store_service_patch.js
```diff
@@ -89,16 +89,14 @@ const StorePatch = {
     /**
      * @param {string} resModel
      * @param {number[]} resIds
-     * @param {number|undefined} defaultActivityTypeId
+     * @param {Object|undefined} [predifinedContext] Context values to pass to the activity form view as default_*
```
mrsr-odoo: same for the docs

mrsr-odoo (commented):
Hello @adsh-odoo 👋 ,
Left some comments
Thanks

addons/calendar/models/res_users.py
adsh-odoo: Edit-- Had another look at this one, seems we can optimize performance if we stick to 3 records. Will share the concern with PO and will try to convince him also to modify this functionality. Meanwhile, i will made the changes

addons/mail/static/src/core/common/thread_model.js
adsh-odoo: The approach has been changed as recently we moved the delete button inside the activity form view, which leads to the same execution path for both markDone and cancel activity. Earlier, when cancelling the activity had a different execution flow, so that's why I had done it earlier. The current approach also seems cleaner compared to the previous one

addons/mail/models/mail_message.py
```diff
@@ -200,6 +200,7 @@ def default_get(self, fields):
...
+    mail_activity_id = fields.Many2one('mail.activity')
```
tde-banana-odoo: Should add a not null index (as it is going to be generally void), and also ondelete to set
```
index='btree_not_null', ondelete='set null',
```

addons/test_mail/tests/test_mail_activity.py
```diff
@@ -289,7 +289,9 @@ def test_activity_flow_employee(self):
             activity.action_feedback(feedback='So much feedback')
             self.assertEqual(activity.feedback, 'So much feedback')
             self.assertEqual(test_record.activity_ids, self.env['mail.activity'])
-            self.assertEqual(test_record.message_ids[0].subtype_id, self.env.ref('mail.mt_activities'))
+            last_message = test_record.message_ids[0]
+            self.assertEqual(last_message.subtype_id, self.env.ref('mail.mt_activities'))
+            self.assertEqual(last_message.mail_activity_id, activity)
```
tde-banana-odoo, line 292: Should also check activity_type_id while we are here :)

tde-banana-odoo (commented):
saw some light :wave:

addons/mail/models/mail_message.py
adsh-odoo: Will add an index, just have a query related to the ondelete property, Many2One field fallsback to `set null` bydefault so IMO we can skip that. Let me know if I missed something.

PR #248787, opened 2026-02-16
[FIX] point_of_sale: add shouldRoundChange getter for override

hadielyakhni:
We extract the condition into a getter so we can override it elsewhere. For now, it is overriden in `pos_settle_order`. See below for explanation.

`change` is the amount we return to the client, so it's in cash, and therefore, we always round it if `cash_rounding` is true.

However, in pos_settle_order, when a client 'deposits money', the order `change` is what we get paid by that client, and he might choose a payment method different than cash (card for e.g.). We change the `shouldRoundChange` logic to count for such cases.

opw-5222985


adgu-odoo (commented):
robodoo r+

PR #248789, opened 2026-02-16
[IMP] hr_holidays : adjust group request timeoff type

adak-ahmed:
-New domain has been added for the work_entry_type field, so as to show the no_allocation types and the valid types where employees have allocations.

### DIFF
--- a/addons/hr_holidays/wizard/hr_leave_generate_multi_wizard.py
+++ b/addons/hr_holidays/wizard/hr_leave_generate_multi_wizard.py
@@ -3,7 +3,7 @@
 from datetime import datetime, timedelta, UTC
 from zoneinfo import ZoneInfo
 
-from odoo import fields, models
+from odoo import api, fields, models
 from odoo.exceptions import UserError
 from odoo.fields import Domain
 
@@ -25,7 +25,7 @@ class HrLeaveGenerateMultiWizard(models.TransientModel):
 
     name = fields.Char("Description")
     work_entry_type_id = fields.Many2one(
-        "hr.work.entry.type", string="Time Off Type", required=True)
+        "hr.work.entry.type", string="Time Off Type", required=True, domain="[('id', 'in', valid_work_entry_type_ids)]")
     employee_ids = fields.Many2many('hr.employee', string='Employees', domain=lambda self: self._get_employee_domain())
     company_id = fields.Many2one('res.company', default=lambda self: self.env.company, required=True)
     date_from = fields.Date('Start Date', required=True, default=lambda self: fields.Date.today())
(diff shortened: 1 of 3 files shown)


addons/hr_holidays/wizard/hr_leave_generate_multi_wizard.py
```diff
@@ -120,3 +121,29 @@ def action_generate_time_off(self):
...
+                employee_data = allocation_data[employee._origin]
+                entry_allocation = next((item[1] for item in employee_data if item[0] == entry_type.name), None)
+                if entry_allocation is not None:
+                    result.append(entry_allocation)
+            invalid_allocation = any(data.get("virtual_remaining_leaves") <= 0 for data in result)
+            if not invalid_allocation or (invalid_allocation and entry_type.allows_negative):
+                valid_ids.append(entry_type.id)
+            else:
+                continue
+        self.valid_work_entry_type_ids = valid_ids
```
amah-odoo: not sure this work fine or not but the idea to go with more simple approach 
```suggestion
    @api.depends('employee_ids', 'date_from')
    def _compute_valid_work_entry_type_ids(self):
        res = self.env['hr.work.entry.type']
        work_entry_types = self.env['hr.work.entry.type'].search([]).grouped('requires_allocation')
        work_entry_types_no_alloc, work_entry_types_alloc = work_entry_types.values()
        res |= work_entry_types_no_alloc
        allocations = self.env['hr.leave.allocation'].search([
            ('work_entry_type_id', 'in', work_entry_types_alloc.ids),
            ('employee_id', 'in', self.employee_ids),
        ]).grouped('work_entry_type_id')
        for entry_type, entry_allocations in allocations:
            if entry_allocations.employee_id == self.employee_ids:
                res |= entry_type
        self.valid_work_entry_type_ids = res
```

amah-odoo (changes requested):
Thanks for your work 💯 

we need to work on the `_compute_valid_work_entry_type_ids` performance as  `get_allocation_data` seems overkill.

addons/hr_holidays/wizard/hr_leave_generate_multi_wizard.py
```diff
@@ -120,3 +121,24 @@ def action_generate_time_off(self):
...
+    def _compute_valid_work_entry_type_ids(self):
+        res = self.env["hr.work.entry.type"]
+        work_entry_types = (
+            self.env["hr.work.entry.type"].search([]).grouped("requires_allocation")
+        )
+        work_entry_types_alloc, work_entry_types_no_alloc = work_entry_types.values()
+        res |= work_entry_types_no_alloc
+        allocations = self.env['hr.leave.allocation'].search([
+            ('work_entry_type_id', 'in', work_entry_types_alloc.ids),
+            ('employee_id', 'in', self.employee_ids.ids),
```
amah-odoo, line 136: I think we can add a date check here to consider only the allocation that start before the leave and end after it 👀

amah-odoo (changes requested):
I step to go 🪜

amah-odoo (approved):
Thanks for your work 💯

PR #248792, opened 2026-02-16
[IMP] mail: fix attachment image scaling

phenix-factory:
When editing a message, newly attached images were displayed larger than the attachment cards.

This commit forces attachment images to use the same height as the attachment cards.

Task-5867464 (part of point 53)

Before
image
After
image

### DIFF
--- a/addons/mail/static/src/core/common/attachment_list.scss
+++ b/addons/mail/static/src/core/common/attachment_list.scss
@@ -1,11 +1,13 @@
+$o-mail-attachment-height: 10em;
+
 .o-mail-AttachmentContainer {
     &.o-inMessage.o-isImage:not(.o-inChatter) {
         min-width: 13em;
-        min-height: 10em;
+        min-height: $o-mail-attachment-height;
     }
     &:not(.o-inMessage.o-isImage), &.o-inChatter.o-inMessage.o-isImage {
         width: 13em;
-        height: 10em;
+        height: $o-mail-attachment-height;
     }
     background-color: $gray-200;
 
@@ -38,8 +40,12 @@
     line-height: 1.25;
 }
 
-.o-mail-AttachmentImage.o-inMessage {
+.o-mail-AttachmentImage.o-inMessage:not(.o-inComposer) {
     max-height: 300px;
+}
+
+.o-mail-AttachmentImage.o-inComposer {
+    max-height: $o-mail-attachment-height;
 }
 
 .o-viewable {
--- a/addons/mail/static/src/core/common/attachment_list.xml
+++ b/addons/mail/static/src/core/common/attachment_list.xml
@@ -44,6 +44,7 @@
                         t-elif="attachment.isImage"
                         class="o-mail-AttachmentImage img img-fluid w-100 rounded-3"
                         t-att-class="{
+                            'o-inComposer': env.inComposer,
                             'o-inMessage': env.inMessage,
                             'o-inChatter': env.inChatter,
                             'object-fit-contain': !env.inChatter and env.inMessage,


addons/mail/static/src/core/common/attachment_list.scss
```diff
@@ -39,7 +39,7 @@
 }
 
 .o-mail-AttachmentImage.o-inMessage {
-    max-height: 300px;
+    max-height: 140px;
```
seb-odoo: I would only change that if message has both attachments and images for consistency.
If there is only images, and in particular only one image, having the bigger size is much better.

alexkuhn: Agree, `140px` forces us to use the file viewer in dialog whereas `300px` makes the preview sufficient to see what the image is about.

The problem is really when combining non-images and images together, so I would limit to when mixing both together too.

I wouldn't even mind if we say like at 3+ images they are also enforced to `140px`, at this point this person is sharing an album and smaller thumbnails makes sense. Just the case of 1 or 2 images the `300px` height is preferred.

phenix-factory: The user already knows what he is uploading and doesn't need a giant thumbnail that takes up a substantial part of the screen when editing a message. We don't need the amount of code this is going to generate, it's overkill to me.

seb-odoo: Oh this is editing? Then yes I would tend to agree, but the diff does not seem like it targets editing only.

phenix-factory: You are right, it's not just for editing, it applies to the attachment list inside a message.

But if we agree that it can be smaller in editing, then it should just be small in edition and we keep the same behavior otherwise

seb-odoo: Yes ok to be smaller in editing.

phenix-factory: image

addons/mail/static/src/core/common/attachment_list.xml
```diff
@@ -44,7 +44,8 @@
                         t-elif="attachment.isImage"
                         class="o-mail-AttachmentImage img img-fluid w-100 rounded-3"
                         t-att-class="{
-                            'o-inMessage': env.inMessage,
+                            'o-inComposer': env.inComposer,
+                            'o-inMessage': env.inMessage and !env.inComposer,
```
alexkuhn: Keep `'o-Message': env.inMessage,` and instead adapt SCSS to have `:not(.o-inComposer)` when needed.

This is misleading for `o-inMessage` to not be present when inside a composer. Better have very explicit code of equivalent `env.inMessage` / `env.inComposer` reflected in `o-inMessage` / `o-inComposer`.

phenix-factory: That also works.
Not sure if it's better, the condition is just elsewhere.

addons/mail/static/src/core/common/attachment_list.scss
alexkuhn, line 11: This rule combines with the max-height `300px` / `140px` below, which is not intentional.

Thankfully value below are greater than `10em` so this works fine, but this would mistakenly combine if the value of `max-height` was lower.

addons/mail/static/src/core/common/attachment_list.scss
```diff
@@ -38,10 +38,14 @@
...
 
-.o-mail-AttachmentImage.o-inMessage {
+.o-mail-AttachmentImage.o-inMessage:not(.o-inComposer) {
     max-height: 300px;
 }
 
+.o-mail-AttachmentImage.o-inComposer {
+    max-height: 140px;
+}
+
```
alexkuhn, line 50: I'd use `height: 10em` for consistent sizing of attachments from rule above.
I'd suggest to have a variable to link the value together.

alexkuhn (approved):
LGTM otherwise

PR #248794, opened 2026-02-16
[FIX] mail: prevent editing mailbox thread name when channel is undefined

akth-odoo:
Steps to reproduce:

- Open Discuss.
- Go to Starred, Inbox, or History.
- Try to update the thread name.

The thread name input is editable even when no channel exists, causing a traceback when rename is called on an undefined channel.

This happens because `enabled` defaults to `true` and `thread.channel?.isAllowedToRename` return `undefined`. This commit ensures `enabled` is a boolean.

Task-5940328
---


seb-odoo (approved):
@robodoo r+

PR #248799, opened 2026-02-16
[FIX] web: Make navigation_hook not override default tab navigation

BastienFafchamps:
Before this commit, the navigation_hook used custom hotkey for handling `tab` and `shift+tab` navigation.
Now we use the browser default navigation system, cases which needed extra control (like dropdowns) have to add the `tab` handlers themselves.

If the `virtualFocus` option is true, the custom hotkeys are still used to avoid tabbing out of the focused input. 

---


kebeclibre:
A test involving a nested dropdown will help understand the case. I'm not sure to follow everything.
The commit message should be improved in that way too.

PR #248805, opened 2026-02-16
[FIX] hr_attendance: fix systray label

MissingNoShiny:


xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

addons/hr_attendance/static/src/components/attendance_menu/attendance_menu.xml
```diff
@@ -18,16 +18,11 @@
                 <div class="att_container d-flex flex-column">
                     <div class="d-flex justify-content-between align-items-center gap-5 border-bottom p-3">
                         <h3 class="mb-0">
```
xavierbol, line 20: to avoid replacing that h3, we could name it and use that name to add a t-else and add another h3 before that one to add "timesheets" title.

What do you think? 🙂

xavierbol (commented):

xavierbol:
robodoo r- 

Actually, we could improve here to avoid using "replace" in enterprise PR.

xavierbol (commented):
LGTM, thanks for your work! 🙂

robodoo r+

PR #248806, opened 2026-02-16
[IMP] hr, mail, project: no IM status for non-internal users

Atovange:
enterprise PR: https://github.com/odoo/enterprise/pull/107494

This commit removes the possibility for non-internal users to see the IM status of other users/guests.
Non-internal users don't need this information, making it unnecessarily invasive.


addons/hr/models/hr_employee.py
```diff
@@ -2037,7 +2037,12 @@ def action_open_versions(self):
 
     def _store_avatar_card_fields(self, res: Store.FieldList):
         res.one("department_id", ["name"])
-        res.one("user_id", lambda res: (res.attr("share"), res.one("partner_id", ["im_status", "tz"])))
+        res.one("user_id", lambda res: (
+            res.attr("share"),
+            res.one("partner_id", lambda res: (
+                res.attr("im_status") if res.is_for_internal_users() else None,
```
seb-odoo, line 2043: Same remark as enterprise, we should always be using `_store_im_status_fields` instead.

addons/hr/models/hr_employee_public.py
```diff
@@ -227,7 +227,12 @@ def init(self):
 
     def _store_avatar_card_fields(self, res: Store.FieldList):
         res.one("department_id", ["name"])
-        res.one("user_id", lambda res: (res.attr("share"), res.one("partner_id", ["im_status", "tz"])))
+        res.one("user_id", lambda res: (
+            res.attr("share"),
+            res.one("partner_id", lambda res: (
+                res.attr("im_status") if res.is_for_internal_users() else None,
```
seb-odoo, line 233: `_store_im_status_fields`

addons/mail/models/discuss/res_partner.py
```diff
@@ -145,7 +145,7 @@ def get_mention_suggestions_from_channel(self, channel_id, search, limit=8):
         store.add(
             partners,
             lambda res: (
-                res.from_method("_store_partner_fields"),
+                res.from_method("_store_partner_fields") if res.is_for_internal_users() else None,
```
seb-odoo, line 148: This looks incorrect, other fields from `_store_partner_fields` should likely still be returned.

addons/mail/models/ir_websocket.py
```diff
@@ -43,6 +43,8 @@ def _update_mail_presence(self, inactivity_period):
 
     def _prepare_subscribe_data(self, channels, last):
         data = super()._prepare_subscribe_data(channels, last)
+        if not self.env.user._is_internal():
+            return data
```
seb-odoo, line 47: Do we actually need this return?
If we do want to provide the token in very specific flow, it shouldn't be prevented.

addons/mail/models/ir_websocket.py
```diff
@@ -103,7 +105,7 @@ def _prepare_subscribe_data(self, channels, last):
 
     def _after_subscribe_data(self, data):
         user, guest = self.env["res.users"]._get_current_persona()
-        if user or guest:
+        if (user or guest) and self.env.user._is_internal():
```
seb-odoo, line 108: same remark/question

addons/hr/models/hr_employee.py
```diff
@@ -2037,7 +2037,12 @@ def action_open_versions(self):
...
+                res.attr("tz"),
```
seb-odoo, line 2044: Note: `tz` is inside the check in some other places (partner in particular) so we might want to move it here too?

addons/mail/static/src/core/common/discuss_avatar.js
```diff
@@ -50,6 +52,9 @@ export class DiscussAvatar extends Component {
         if (this.channel) {
             return this.channel.showThreadIcon({ ignoreTyping: !this.props.typing });
         }
+        if (!this.store.self_user || this.store.self_user.share === true) {
+            return false;
+        }
```
seb-odoo, line 57: Same kind of remark: I don't think this should be hard-coded, it should just implicitly not be shown if the data is not received.

addons/mail/tests/discuss/test_discuss_mail_presence.py
```diff
@@ -55,15 +55,9 @@ def test_presence_access(self):
...
         portal = new_test_user(self.env, login="portal_user", groups="base.group_portal")
-        other_portal = new_test_user(
-            self.env, login="other_portal_user", groups="base.group_portal"
-        )
         guest = self.env["mail.guest"].create({"name": "Guest"})
-        other_guest = self.env["mail.guest"].create({"name": "Other Guest"})
         for requested_by, target, has_token, allowed in [
             *product([internal], [guest, other_internal, portal], [True, False], [True]),
-            *product([guest, portal], [internal, other_guest, other_portal], [False], [False]),
-            *product([guest, portal], [internal, other_guest, other_portal], [True], [True]),
```
seb-odoo, line 66: I would keep these unless we delete the token feature entirely (I don't think we should just yet).

seb-odoo (commented):
Note: there are several places after this PR where we might want/need to send an extra bus store targeting the internal user sub-channel of the main channel.

Otherwise we might no longer receive the new im status for users for which it should be received/displayed.

In particular: new member (fully unknown in JS before that point) joining channel and displaying it in the member list -> we expect to have the status available.

But if the channel is not internal-only, it won't be sent at all.

PR #248812, opened 2026-02-16
[IMP] mail: add space after emoji

phenix-factory:
Task-5867464 (point 70)

### DIFF
--- a/addons/mail/static/src/core/common/composer.js
+++ b/addons/mail/static/src/core/common/composer.js
@@ -900,14 +900,14 @@ export class Composer extends Component {
     addEmoji(str) {
         const composer = toRaw(this.props.composer);
         if (this.editor) {
-            this.editor.shared.dom.insert(str);
+            this.editor.shared.dom.insert(str + " ");
             this.editor.shared.history.addStep();
         } else {
             const composerText = composer.composerText;
             const firstPart = composerText.slice(0, composer.selection.start);
             const secondPart = composerText.slice(composer.selection.end, composerText.length);
-            composer.composerText = firstPart + str + secondPart;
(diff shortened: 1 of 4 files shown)


addons/mail/static/src/core/common/composer.js
```diff
@@ -893,14 +893,14 @@ export class Composer extends Component {
...
         if (this.editor) {
-            this.editor.shared.dom.insert(str);
+            this.editor.shared.dom.insert(str + " ");
             this.editor.shared.history.addStep();
         } else {
             const composerText = composer.composerText;
             const firstPart = composerText.slice(0, composer.selection.start);
             const secondPart = composerText.slice(composer.selection.end, composerText.length);
-            composer.composerText = firstPart + str + secondPart;
-            this.selection.moveCursor((firstPart + str).length);
```
alexkuhn, line 910: The whitespace should not be triggered when using multi-selection in emojipicker, i.e. with SHIFT + click to select emojis in emoji picker.
Emoji picker provides `resetOnSelect` to the emoji picker `onSelected`:
https://github.com/odoo/odoo/blob/19.0/addons/web/static/src/core/emoji_picker/emoji_picker.js#L477

This should be provided to `addEmoji`:
https://github.com/odoo/odoo/blob/19.0/addons/mail/static/src/core/common/composer_actions.js#L105
```js
onSelect: (emoji, resetOnSelect) => owner.addEmoji(emoji, resetOnSelect),
```

So that the whitespace is added only when `resetOnSelect` is truthy.

This would make SHIFT + selection to not have whitespace, as is the expected behavior (same as Discord by the way).

alexkuhn (approved):
LGTM otherwise.
@robodoo delegate+

PR #248822, opened 2026-02-16
[FIX] account, l10n_sa: fix RTL direction in portal preview and subtotal layout

hisi-odoo:
When opening invoice HTML preview in the portal, the layout was displayed in LTR
even if the invoice partner language was RTL (e.g. Arabic). This happened when
the portal user language was LTR. PDF structure was correct, but HTML preview had
wrong text flow and alignment.

Root cause was that RTL asset processing depends on `env.lang`, not on `t-lang`
in templates. Since the ORM context language remained the portal user’s language,
CSS was compiled in LTR and hardcoded `direction: ltr;` rules from `report.scss`
were overriding the `<html dir="rtl">`.

To fix this, the portal controller now checks the partner language and, if it is
RTL, updates the request context before rendering. This ensures `_render_qweb_html`
runs with the correct `env.lang`, triggering proper RTL CSS conversion via rtlcss.

Additionally, the subtotal section in the Saudi invoice report could have its
values visually cut due to insufficient width allocated to the tax totals
column. The column width has been slightly increased to ensure totals are
properly displayed.

task-5878719

### DIFF
--- a/addons/account/controllers/portal.py
+++ b/addons/account/controllers/portal.py
@@ -172,6 +172,10 @@ class PortalAccount(CustomerPortal):
         elif report_type in ('html', 'pdf', 'text'):
             has_generated_invoice = bool(invoice_sudo.invoice_pdf_report_id)
             request.update_context(proforma_invoice=not has_generated_invoice)
+            # If the partner's language is RTL, set the context language to match, ensuring the report shows the correct text direction.
+            partner_lang = invoice_sudo.partner_id.lang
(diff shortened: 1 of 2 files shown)


khva-odoo (approved):

hisi-odoo:
Hello @alialfie 👋

When you get some time, could you please review this PR?
Thanks in advance! 😃

alialfie (approved):
Thanks for the fix! LGTM
robodoo r+

PR #248823, opened 2026-02-16
[IMP] html_editor: apply format/color to block if completely selected

deso-odoo:

When a paragraph-related element is fully selected, formatting commands (e.g. font-size, font-style, font-weight, etc.) and color are applied directly on the block element instead of new span or font element.

List items already behaved this way; this change aligns other block elements with the same behavior.

Applying the format on the block simplifies the DOM structure and ensures consistent behavior when formatting entire blocks.

task-5215987

---

### DIFF
--- a/addons/html_editor/static/src/core/format_plugin.js
+++ b/addons/html_editor/static/src/core/format_plugin.js
@@ -4,7 +4,7 @@ import { callbacksForCursorUpdate } from "@html_editor/utils/selection";
 import { _t } from "@web/core/l10n/translation";
 import { Plugin } from "../plugin";
 import { closestBlock, isBlock } from "../utils/blocks";
-import { cleanTextNode, fillEmpty, removeClass, splitTextNode, unwrapContents } from "../utils/dom";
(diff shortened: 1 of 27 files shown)


addons/html_editor/static/src/core/format_plugin.js
```diff
@@ -91,14 +99,14 @@ export class FormatPlugin extends Plugin {
                 id: "formatUnderline",
                 description: _t("Toggle underline"),
                 icon: "fa-underline",
-                run: this.formatSelection.bind(this, "underline"),
+                run: this.formatSelection.bind(this, "underline", { applyToBlock: false }),
```
deso-odoo, line 103: Unlike other text styles such as `font-weight` or `font-style`, `text-decoration` applied on a parent element propagates across the entire inline content and cannot be canceled by descendants using `text-decoration: none` which is the reason we can't have `addNeutralStyle` for underline and strikethrough, hence we don't apply them on block.

deso-odoo:
handled in https://github.com/odoo/odoo/pull/266372

PR #248825, opened 2026-02-16
[FIX] website_mass_mailing_event: correct aside fragment name

creaturemusings:
Commit [33685d7dfd8b](https://github.com/odoo/odoo/commit/33685d7dfd8b232802a5c46ed78d9411f328a5d3) added autofill email snippets
for products and events.

The name of one of the event fragments had a typo, resulting in
the autofill plugin trying to reach a template that doesn't exist.

This commit fixes that issue.

This commit also fixes an edge case issue that would occur if an event
had an associated "address" partner that didn't have a country,
resulting in an error when trying to select that event.

task-5940938

### DIFF
--- a/addons/website_mass_mailing_event/static/src/builder/options/event_snapshot_fragments.xml
+++ b/addons/website_mass_mailing_event/static/src/builder/options/event_snapshot_fragments.xml
@@ -100,7 +100,7 @@
 
     <t t-name="website_mass_mailing_event.event_address_template">
         <t t-set="has_city" t-value="address.city"/>
-        <t t-set="has_state" t-value="state and country.state_required"/>
+        <t t-set="has_state" t-value="state and country and country.state_required"/>
 
         <span class="fa fa-map-marker"></span> <t t-if="has_city">
             <t t-out="address.city"/><t t-if="address.country_id">, </t>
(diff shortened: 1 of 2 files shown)


addons/website_mass_mailing_event/static/src/builder/options/event_snapshot_fragments.xml
```diff
@@ -100,14 +100,14 @@
...
+        <t t-set="has_state" t-value="state and country and country.state_required"/>
 
         <span class="fa fa-map-marker"></span> <t t-if="has_city">
             <t t-out="address.city"/><t t-if="address.country_id">, </t>
         </t>
         <t t-if="has_state">
             <t t-out="state.code"/>,
         </t>
-        <t t-out="country.name"/>
+        <t t-out="country and country.name"/>
```
abd-msyukyu-odoo: not sure, but shouldn't you do `<t t-if="country and country.name"><t t-out="country.name">` ? Won't your current form display some falsy value?

abd-msyukyu-odoo (commented):
@creaturemusings Heya, thanks for the fix, instead of `19.2`, mention the [commit](https://github.com/odoo/odoo/commit/33685d7dfd8b232802a5c46ed78d9411f328a5d3) with with a link

addons/website_mass_mailing_event/static/src/builder/options/event_snapshot_fragments.xml
creaturemusings: It didn't when I tried it but it doesn't hurt to be prudent/explicit.

PR #248831, opened 2026-02-16
[FIX] mail: increase timeout of waitStoreFetch

seb-odoo:
The following tests fail no-deterministically on runbot.

The issue cannot be reproduced locally with more than 500 runs and the tests show no obvious coding issue.

The most realistic explanation is simply that heavy load would make some async steps timeout before they actually resolve: the expected payload is not caught at the start of the test when it should, but it is later caught at the end of the test.

runbot-239937
runbot-240553

### DIFF
--- a/addons/mail/static/tests/mail_test_helpers.js
+++ b/addons/mail/static/tests/mail_test_helpers.js
@@ -874,10 +874,11 @@ export function listenStoreFetch(nameOrNames = [], { logParams = [], onRpc: onRp
  * @param {boolean} [options.ignoreOrder=false]
  * @param {string[]} [options.stepsAfter=[]]
  * @param {string[]} [options.stepsBefore=[]]
+ * @param {number} [options.timeout=5000]
  */
 export async function waitStoreFetch(
     nameOrNames = [],
-    { ignoreOrder = false, stepsAfter = [], stepsBefore = [] } = {}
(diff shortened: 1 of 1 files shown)


addons/mail/static/tests/mail_test_helpers.js
```diff
@@ -874,10 +874,11 @@ export function listenStoreFetch(nameOrNames = [], { logParams = [], onRpc: onRp
  * @param {boolean} [options.ignoreOrder=false]
  * @param {string[]} [options.stepsAfter=[]]
  * @param {string[]} [options.stepsBefore=[]]
+ * @param {number} [options.timeout=5000]
  */
 export async function waitStoreFetch(
     nameOrNames = [],
-    { ignoreOrder = false, stepsAfter = [], stepsBefore = [] } = {}
+    { ignoreOrder = false, stepsAfter = [], stepsBefore = [], timeout = 5000 } = {}
```
seb-odoo, line 881: Note: default in `waitForSteps` is `2000` which is simply too low. Having such a low timeout has no benefit except making testes crash for no reason when they could just wait a little longer.

It's unrealistic to hope for all tests to run fast all the time.

If it was just me I'd just put `30000` or something, we don't want to test the performance of all tests all the time, we just want this timeout to prevent the test suite from hanging when there is an actual error.

phenix-factory (approved):
@robodoo r+ delegate+

PR #248833, opened 2026-02-16
[FIX] website: ensure footer copyright applies before mega footers

sben-odoo:
Steps to reproduce:
===================
1. Go to website and switch footer to mega footer
2. Add a language & edit copyright text
3. Save & change the language

=> breaks

Cause:
======
Editing triggers Odoo's Copy-On-Write (COW) mechanism, creating a
customized template with a higher database ID. This causes it to
execute after the mega footer templates, which expect col-sm classes
that were already changed to col-md. The xpaths fail and the footer
breaks.

The existing patch in html_editor/models/ir_ui_view.py (from [1])
that re-adds col-md when saving one of the affected views
was not working in this case because the write was done with
delayed_translations=True in context, causing the ancestor patch to be
skipped.

Solution:
=========
- Add `delayed_translations=False` to the ancestor write in `ir_ui_view.py`
  so the patch is applied immediately when editing the footer. This
  prevents the issue without requiring a module update.
- Add priority="15" to `footer_copyright_company_name` so that even for
  already-broken databases (where COW was created before this fix),
  the template runs before mega footer templates after a module update.
  The priority is inherited by COW-generated views.

[1]: https://github.com/odoo/odoo/commit/2d6699b4c460084110ccada0392845156af4a135
opw-5485986

### DIFF
--- a/addons/website/views/website_templates.xml
+++ b/addons/website/views/website_templates.xml
@@ -2453,7 +2453,7 @@
     </xpath>
 </template>
 
-<template id="footer_copyright_company_name" inherit_id="website.layout">
+<template id="footer_copyright_company_name" inherit_id="website.layout" priority="15">
     <xpath expr="//footer//span[hasclass('o_footer_copyright_name')]" position="replace">
         <span class="o_footer_copyright_name me-2 small">Copyright &amp;copy; Company name</span>
     </xpath>
--- a/addons/html_editor/models/ir_ui_view.py
+++ b/addons/html_editor/models/ir_ui_view.py
@@ -338,7 +338,7 @@ class IrUiView(models.Model):
                     node.set('class', node.get('class') + ' col-md')
                     has_change = True
             if has_change:
-                ancestor.with_context(no_cow=True).write({'arch': etree.tostring(arch, encoding='unicode')})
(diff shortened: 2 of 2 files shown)


addons/website/views/website_templates.xml
```diff
@@ -2453,7 +2453,7 @@
     </xpath>
 </template>
 
-<template id="footer_copyright_company_name" inherit_id="website.layout">
+<template id="footer_copyright_company_name" inherit_id="website.layout" priority="10">
```
nle-odoo: probably `priority="15"` should be enough (to prevent creating inverse issue where a higher priority view would be affected by changing this view too low)

nle-odoo (approved):
makes sense to me, the priority of inheriting views is "priority, id":

https://github.com/odoo/odoo/blob/fee6b32a8a57577bd8229c80dff6f93964f9f556/odoo/addons/base/models/ir_ui_view.py#L754

so since in this scenario we will create the `website.template_footer_mega` view before the COW of `website.footer_copyright_company_name` view, I don't see a solution besides changing the priority.

another solution could be to have a website.template_footer_mega have a XPATH that would work in both situation (if `website.footer_copyright_company_name` was applied after or before), but this would only work for people that update "website" module.

PR #248838, opened 2026-02-16
[FIX] point_of_sale: allow barcode scan for product packaging

aksi-odoo:
Step to reproduce
- Create a product
- Add two attributes:
1. One with Instantly creation mode
2. One with Never creation mode
- Define packaging from the Sales tab
- Add a barcode on the variant packaging ex: 111356,11357
- Scan the packaging barcode in POS

Observation:
- we get a traceback `TypeError: Cannot read properties of undefined (reading 'product_template_attribute_value_ids')`

Cause:
- when we do not have barcode on product, when opening `openConfigurator`
- (as we have few varianst) product get undefined.
https://github.com/odoo/odoo/blob/f229f23d7bf3d837ff5577c36145bf2ba410ea22/addons/point_of_sale/static/src/app/services/pos_store.js#L738
- Hence the traceback

Fix:
- For the case, when packaging has barcode but not the product,
we search for product in that case too.

opw-5886570

---

### DIFF
--- a/addons/point_of_sale/static/src/app/services/pos_store.js
+++ b/addons/point_of_sale/static/src/app/services/pos_store.js
@@ -687,6 +687,13 @@ export class PosStore extends WithLazyGetterTrap {
             let product;
             if (opts.code) {
                 product = this.models["product.product"].getBy("barcode", opts.code.base_code);
+                if (!product) {
+                    const productPackaging = this.models["product.uom"].getBy(
+                        "barcode",
+                        opts.code.base_code
+                    );
+                    product = productPackaging && productPackaging.product_id;
+                }
             } else {
                 product = opts.presetVariant;
             }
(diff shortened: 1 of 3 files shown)


alsh-odoo (approved):

stva-odoo (commented):
robodoo r+

addons/point_of_sale/static/src/app/services/pos_store.js
```diff
@@ -691,7 +691,7 @@ export class PosStore extends WithLazyGetterTrap {
                 product = opts.presetVariant;
             }
             attributeLinesValues = attributeLinesValues.map((values) =>
-                values[0].attribute_id.create_variant === "no_variant"
+                values[0].attribute_id.create_variant === "no_variant" || !product
```
stva-odoo: I suggest using the same code as in  18.4 https://github.com/odoo/odoo/blob/saas-18.4/addons/point_of_sale/static/src/app/services/pos_store.js#L792

```
           const attrValueIds = new Set(
                product?.product_template_attribute_value_ids?.map((v) => v.id) || []
            );

            attributeLinesValues = attributeLinesValues.map((values) =>
                values[0].attribute_id.create_variant === "no_variant"
                    ? values
                    : values.filter((value) => attrValueIds.has(value.id))
            );
```

It should fix your issue

stva-odoo:
robodoo r+

PR #248842, opened 2026-02-16
[FIX] l10n_hr: Receivable and Payable account swapped

leclerc-leo:
The receivable accounts were wrongly setup where the receivable account of the tax group was set with the payable account, ...

task-5045769

---


oco-odoo (approved):
@robodoo r+

PR #248845, opened 2026-02-16
[IMP] point_of_sale: add resilience to paymentMethodImage

jeanschoenlaub:
In preparation for owl3, we had to manually refactor the `paymentMethodImage` which was a bit fragile.

BEFORE:
`paymentMethodImage` uses `this.paymentMethod.image` even if paymentMethod is a template variable and not a componenent variable.

NOW:
`paymentMethodImage` uses `paymnentMethod.image` making it clear the variable is coming from the template.

task: OWL3 prep - add this. to template variables

---


davidmonnom (approved):
@robodoo r+

PR #240368, opened 2025-12-17
[FIX] stock: ensure write access to sequence of product's lot numbers

nsirjacobs:
**Behavior:**
Behavior:
When logged in as a user with group_user access (member), trying to generate the next few serial numbers for a tracked product from receipts will cause an access error.

This happens because there is a write operation on the ir.sequence linked to the serial number at the end of the process that happens only when the 'New' button was previously pressed.
And since access rights to ir.sequence are dependant on the base group of the user, changing the user Inventory rights to admin will not resolve the issue.

A test has been created to ensure no access error is created when generating sequence numbers as a lower access user, this test would pass without the fix since it is focused on the python function.

**Steps to reproduce:**
- Create a product that is tracked by serial number
- Log in as a user with group_user access (member role)
- Go to Inventory -> Operations -> Receipts
- Create a new picking and add the product
- After clicking on Mark as Todo you'll see Details pop up in the product line
- After clicking Details, select Generate Serial/Lots
- Click New, then Generate, and you'll get an access error for ir.Sequence

opw-5368553

PR #248846, opened 2026-02-16
[FIX] stock: prevent error on missing sequence

bhna-odoo:
Currently, an error occurs when a user tries to generate lots while providing a lot number.

**Steps to replicate:**
- Install purchase (without demo).
- Create a product `test`.
- Install stock and turn on `Lots and Serial Numbers`
- Open the product `test` and turn on `Track Inventory` `by Lots`.
- Open Receipts > add the product `test`> give demand as 3 > and go to its form view using view button.
- Click `Generate Lots` > type `lot1` in `First lot Number` > Generate > Error-1
- Click `Generate Lots` > type 0 in Quantity received > Generate > Error-2.

**Error-1:**
```
  File '/home/odoo/odoo18/community/addons/stock/models/stock_move.py', line 1026, in action_generate_lot_line_vals
    if (first_lot and first_lot == product.lot_sequence_id.get_next_char(first_number)):
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File '/home/odoo/odoo18/community/odoo/addons/base/models/ir_sequence.py', line 237, in get_next_char
    interpolated_prefix, interpolated_suffix = self._get_prefix_suffix()
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File '/home/odoo/odoo18/community/odoo/addons/base/models/ir_sequence.py', line 227, in _get_prefix_suffix
    self.ensure_one()
  File '/home/odoo/odoo18/community/odoo/orm/models.py', line 5640, in ensure_one
    raise ValueError('Expected singleton: %s' % self)
ValueError: Expected singleton: ir.sequence()
```

**Error-2:**
```
  File '/home/odoo/odoo18/community/addons/stock/models/stock_move.py', line 1025, in action_generate_lot_line_vals
    first_number = product.lot_sequence_id.number_next_actual - product.lot_sequence_id.number_increment
                   ^^^^^^^
UnboundLocalError: cannot access local variable 'product' where it is not associated with a value
```
---
**Cause:**
- Both errors originated through a recent [PR].

**Error-1 (Expected singleton: ir.sequence()):**
- As the product was already created before Inventory was installed, the `lot_sequence_id` was empty. (Note:`lot_sequence_id` field has a default value , but default value
       assignment triggers only during the record creation,  any records created
       before stock is installed will not be assigned any value for
      `lot_sequence_id`.)

- As no `lot_sequence_id` is assigned to `test` product the line [1] calls `get_next_char()` on an empty recordset which further calls `_get_prefix_suffix()` [2] and raises singletonerror from [here].

**Error-2 (UnboundLocalError: cannot access local variable 'product'):**
- As the `Received Quantity` was given 0, the `count` argument is received as 0 and as a result the `lot_qties` [3] and `lot_names` [4] are received as empty lists.
- This causes their [zip] to be empty list too and the loop never runs, so assignment to [product] variable never happens and causes the error to occur from here [5].
---
**Solution:**

**Error-1:**
- Now we perform write on `product.lot_sequence_id` only if it exists, otherwise we skip it.

**Error-2:**
- Moved the static assignment of variable `product` and `location_dest_id` outside the loop, this will also prevent the browse being called multiple times for browsing the same record.

[PR]: https://github.com/odoo/odoo/pull/240368

[1]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L1026

[2]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/odoo/addons/base/models/ir_sequence.py#L237

[here]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/odoo/addons/base/models/ir_sequence.py#L227

[3]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L989

[4]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L994

[zip]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L1000

[product]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L1004

[5]: https://github.com/odoo/odoo/blob/7ab52c1675b9764d11454c7b5216064bec4628f8/addons/stock/models/stock_move.py#L1025

sentry-7254849206,7265844194
---

### DIFF
--- a/addons/stock/models/stock_move.py
+++ b/addons/stock/models/stock_move.py
@@ -997,15 +997,15 @@ Please change the quantity done or the rounding precision in your settings.""",
             lot_qties = [1] * len(lot_names)
 
         vals_list = []
+        loc_dest = self.env['stock.location'].browse(default_vals['location_dest_id'])
+        product = self.env['product.product'].browse(default_vals['product_id'])
         for lot, qty in zip(lot_names, lot_qties):
             if not lot.get('quantity'):
                 lot['quantity'] = qty
-            loc_dest = self.env['stock.location'].browse(default_vals['location_dest_id'])
-            product = self.env['product.product'].browse(default_vals['product_id'])
-            loc_dest = loc_dest._get_putaway_strategy(product, lot['quantity'])
+            putaway_loc_dest = loc_dest._get_putaway_strategy(product, lot['quantity'])
             vals_list.append({**default_vals,
                              **lot,
-                             'location_dest_id': loc_dest.id,
+                             'location_dest_id': putaway_loc_dest.id,
                              'product_uom_id': product.uom_id.id,
                             })
         if default_vals.get('picking_type_id'):
@@ -1022,9 +1022,10 @@ Please change the quantity done or the rounding precision in your settings.""",
                         'id': value,
                         'display_name': self.env['stock.move.line'][key].browse(value).display_name
                     }
-        first_number = product.lot_sequence_id.number_next_actual - product.lot_sequence_id.number_increment
-        if (first_lot and first_lot == product.lot_sequence_id.get_next_char(first_number)):
-            product.lot_sequence_id.sudo().write({'number_next_actual': first_number + len(lot_qties)})
+        if product.lot_sequence_id:
+            first_number = product.lot_sequence_id.number_next_actual - product.lot_sequence_id.number_increment
+            if (first_lot and first_lot == product.lot_sequence_id.get_next_char(first_number)):
+                product.lot_sequence_id.sudo().write({'number_next_actual': first_number + len(lot_qties)})
         return vals_list
 
     def _push_apply(self):
--- a/addons/stock/tests/test_generate_serial_numbers.py
+++ b/addons/stock/tests/test_generate_serial_numbers.py
@@ -474,3 +474,39 @@ class StockGenerateCommon(TransactionCase):
             {'quantity': 1, 'lot_name': self.product_serial.lot_sequence_id.get_next_char(first_num + 8)},
             {'quantity': 1, 'lot_name': self.product_serial.lot_sequence_id.get_next_char(first_num + 9)},
         ])
+
+    def test_lots_generation_on_product_with_no_lot_sequence(self):
+        """ Test lots generation for a lot-tracked product without a defined lot_sequence_id.
+
+            When `lot_sequence_id` is not set, lot names are generated
+            from the user-provided `first_lot` value.
+
+            For example, if `first_lot='lot01'` and quantity is 4,
+            the generated lot names will be:
+            lot01, lot02, lot03, lot04, using `generate_lot_names()`.
+        """
+        product = self.product_serial
+        product.write({'tracking': 'lot', 'lot_sequence_id': False})
+        move_line_vals = self.env['stock.move'].action_generate_lot_line_vals({
+                'default_tracking': 'lot',
+                'default_product_id': product.id,
+                'default_location_dest_id': self.location_dest.id,
+                'default_quantity': 4,
+            }, "generate", "lot01", 1, "test")
+        lot_names = self.env['stock.lot'].generate_lot_names("lot01", len(move_line_vals))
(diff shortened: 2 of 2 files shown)


addons/stock/models/stock_move.py
```diff
@@ -997,11 +997,11 @@ def remove_prefix(text, prefix):
...
 
         vals_list = []
+        loc_dest = self.env['stock.location'].browse(default_vals['location_dest_id'])
+        product = self.env['product.product'].browse(default_vals['product_id'])
         for lot, qty in zip(lot_names, lot_qties):
             if not lot.get('quantity'):
                 lot['quantity'] = qty
-            loc_dest = self.env['stock.location'].browse(default_vals['location_dest_id'])
-            product = self.env['product.product'].browse(default_vals['product_id'])
             loc_dest = loc_dest._get_putaway_strategy(product, lot['quantity'])
```
maan-odoo: The variable `loc_dest` is reassigned within the loop, which overrides the original value. Consider using a separate variable name to prevent unintended side effects and improve code readability.

addons/stock/models/stock_move.py
```diff
@@ -1022,6 +1022,8 @@ def remove_prefix(text, prefix):
                         'id': value,
                         'display_name': self.env['stock.move.line'][key].browse(value).display_name
                     }
+        if not product.lot_sequence_id:
+            product.lot_sequence_id = self.env.ref('stock.sequence_production_lots', raise_if_not_found=False)
```
maan-odoo: I don’t think we should assign the default `sequence_production_lots` when the product has no `lot_sequence_id`. It would be better to simply skip this block in that case.

addons/stock/tests/test_generate_serial_numbers.py
```diff
@@ -474,3 +474,28 @@ def test_sequence_serial_numbers_access_rights(self):
             {'quantity': 1, 'lot_name': self.product_serial.lot_sequence_id.get_next_char(first_num + 8)},
             {'quantity': 1, 'lot_name': self.product_serial.lot_sequence_id.get_next_char(first_num + 9)},
         ])
+
+    def test_lots_generation_on_product_with_no_lot_sequence(self):
```
maan-odoo, line 478: Good to add doc string here

addons/stock/tests/test_generate_serial_numbers.py
```diff
@@ -474,3 +474,28 @@ def test_sequence_serial_numbers_access_rights(self):
...
+        product.write({'tracking': 'lot', 'lot_sequence_id': False})
+        move_line_vals = self.env['stock.move'].action_generate_lot_line_vals({
+                'default_tracking': 'lot',
+                'default_product_id': product.id,
+                'default_location_dest_id': self.location_dest.id,
+                'default_quantity': 4,
+            }, "generate", "lot01", 1, "test")
+        lot_names = self.env['stock.lot'].generate_lot_names("lot01", len(move_line_vals))
+        test_vals = [{'quantity': 1, 'lot_name': lot_names[i]['lot_name']}
+                    for i in range(len(move_line_vals))]
```
maan-odoo:
```suggestion
        test_vals = [
            {'quantity': 1, 'lot_name': lot['lot_name']}
            for lot in lot_names
        ]
```
- More readable
- No need range() here as we can access lot directctly

maan-odoo (approved):

bhna-odoo:
Hello @odoo/sentry_reviewers,

Please review this PR.

Thanks :smile: !

PR #248852, opened 2026-02-16
[IMP] web: replay action after check identity

thle-odoo:
The introduction of the mechanism for blocking untrusted devices allows the user to be re-authenticated automatically (via fingerprint)[^1]. If automatic re-authentication is successful, from a UX perspective, the user does not see their action being performed.

This commit introduces the ability to replay the action that was blocked with an identity check if the identity is verified. This way, the user does not experience any friction.

Note:
Correct linter alerts.

Task-5941988

[^1]: Commit: 61f22175ef3df37087887e7419dac54a620bbd55

### DIFF
--- a/addons/web/static/src/core/session/check_identity.js
+++ b/addons/web/static/src/core/session/check_identity.js
@@ -1,8 +1,8 @@
-import { Component, EventBus, onWillDestroy, onWillStart, useState } from "@odoo/owl";
+import { Component, onWillStart, useState } from "@odoo/owl";
 import { Dialog } from "@web/core/dialog/dialog";
 import { rpc, RPCError } from "@web/core/network/rpc";
 import { registry } from "@web/core/registry";
-import { useBus, useService } from "@web/core/utils/hooks";
+import { useService } from "@web/core/utils/hooks";
 import { redirect } from "@web/core/utils/urls";
 import { post } from "@web/core/network/http_service";
 import { session } from "@web/session";
@@ -35,18 +35,33 @@ export class CheckIdentityForm extends Component {
         };
 
         this.checkIdentityService = useService("check_identity");
-        this.checkIdentityService.bus.trigger("start");
+
+        if (!this.checkIdentityService.identityCheckPromise) {
+            // The form is mounted directly into an HTTP page (bypassing the dialog service)
+            const { promise, resolve } = Promise.withResolvers();
+            this.checkIdentityService.identityCheckPromise = promise;
+
+            this.checkIdentityService.identityCheckCleanUp = () => {
+                this.close();
+                resolve();
+                this.checkIdentityService.identityCheckPromise = null;
+                this.checkIdentityService.identityCheckCleanUp = null;
+            };
+        }
+
         this.state = useState({
             error: false,
             authMethod: null,
         });
+
         onWillStart(async () => {
             const data = await this.checkIdentityService.getInitData();
 
             // Attempting to verify the identity of the device using its fingerprint
             if (data.fingerprint_check && (await this.checkIdentityService.updateFingerprint())) {
                 // There is no need to re-authenticate the user explicitly via the form
-                this.checkIdentityService.checkSignaling();
+                this.checkIdentityService.identityCheckCleanUp();
+                this.checkIdentityService.channel.postMessage("identityChecked");
             }
 
             this.user = {
@@ -55,11 +70,6 @@ export class CheckIdentityForm extends Component {
             };
             this.setAuthMethods(data.auth_methods);
         });
-        onWillDestroy(() => {
-            this.checkIdentityService.bus.trigger("stop");
-        });
-
-        useBus(this.checkIdentityService.bus, "identityChecked", () => this.close());
     }
 
     setAuthMethods(authMethods) {
@@ -143,21 +153,17 @@ export class CheckIdentity {
     }
 
     setup(env, services) {
-        this.bus = new EventBus();
-        this.channel = new BroadcastChannel("check_identity");
         this.dialogService = services["dialog"];
-        this.started = false;
         this.fingerprint = null;
 
-        this.bus.addEventListener("start", () => {
-            this.started = true;
-        });
-        this.bus.addEventListener("stop", () => {
-            this.started = false;
-        });
+        this.identityCheckPromise = null;
+        this.identityCheckCleanUp = null;
+
+        // Multi-tab sync: if another tab verified the identity, clean up locally
+        this.channel = new BroadcastChannel("check_identity");
         this.channel.addEventListener("message", (event) => {
-            if (event.data === "identityChecked") {
-                this.bus.trigger("identityChecked");
+            if (event.data === "identityChecked" && this.identityCheckCleanUp) {
+                this.identityCheckCleanUp();
             }
         });
 
@@ -255,22 +261,24 @@ export class CheckIdentity {
         if (result?.mfa) {
             return { success: false, mfa: result.mfa, auth_methods: result.auth_methods };
         }
-        this.checkSignaling();
-    }
-
-    checkSignaling() {
-        this.bus.trigger("identityChecked");
+        this.identityCheckCleanUp();
         this.channel.postMessage("identityChecked");
     }
 
     async checkIdentity() {
-        if (!this.started) {
-            this.dialogService.add(CheckIdentityDialog);
+        if (!this.identityCheckPromise) {
+            const { promise, resolve } = Promise.withResolvers();
+            this.identityCheckPromise = promise;
+            const closeDialog = this.dialogService.add(CheckIdentityDialog);
+
+            this.identityCheckCleanUp = () => {
+                closeDialog();
+                resolve();
+                this.identityCheckPromise = null;
+                this.identityCheckCleanUp = null;
+            };
         }
-        // Mutex
-        await new Promise((resolve) => {
-            this.bus.addEventListener("identityChecked", resolve, { once: true });
-        });
+        return this.identityCheckPromise;
     }
 
     async run() {
--- a/addons/auth_timeout/static/src/services/check_identity/check_identity.js
+++ b/addons/auth_timeout/static/src/services/check_identity/check_identity.js
@@ -23,7 +23,7 @@ export class CheckIdentityTimeout {
             this.busService.addEventListener("connect", () => this.updatePresence(), { once: true });
 
             this.presenceService.bus.addEventListener("presence", () => {
-                if (!this.checkIdentityService.started) {
+                if (!this.checkIdentityService.identityCheckPromise) {
                     clearTimeout(this.inactivityTimer);
                     this.startInactivityTimer();
                 }


addons/web/static/src/core/session/check_identity.js
```diff
@@ -182,6 +183,38 @@ export class CheckIdentity {
...
+                            .catch(reject);
+                    });
+            });
+            // Attach custom function
+            promise.abort = function (rejectError = true) {
+                originalPromise?.abort?.(rejectError);
+            };
+            return promise;
+        };
+    }
```
mwath: It feels more complicated than what it should be 🤔 
Isn't the following equivalent, or is there a specific use-case that requires a Promise?
```suggestion
        rpc._rpc = async (...args) => {
            try {
                return await originalRpc(...args);
            } catch (error) {
                if (error.data?.name !== "odoo.http.session.CheckIdentityException") {
                    throw error;
                }
                // Check identity
                await this.run(false);
                // Replay
                return await originalRpc(...args);
            }
        };
    }
```

thle-odoo: Thanks for the suggestion.

Async function will create a new promise and so custom function (`abort`) will not be preserved.

It is therefore necessary to create a new promise object and add the custom function to it.

In this new promise (the function), using an asynchronous function is an anti-pattern in JavaScript. But I agree that it makes the code shorter and more "readable". Therefore, I agree with the suggestion to use an asynchronous function.

(I adapted the suggestion)

addons/web/static/src/core/session/check_identity.js
```diff
@@ -182,6 +184,49 @@ export class CheckIdentity {
...
+                    }
+                }
+            });
+            // Attach custom function
+            promise.abort = function (rejectError = true) {
+                isAborted = true;
+                activePromise?.abort?.(rejectError);
+            };
+            return promise;
+        };
```
aab-odoo: Couldn't this be simplified a bit? I'm not sure about the abort thing, but as far as I know, its usefulness is very limited. And with both your solution or mine, the rpc done by run can't be aborted.

```suggestion
            const originalProm = originalRpc(...args);
            const prom = originalProm.catch(async (error) => {
                if (error.data?.name === "odoo.http.session.CheckIdentityException") {
                    return this.run(false);  // Check identity (mutex)
                }
                throw e;
            });
            prom.abort = originalProm.abort;
            return prom;
        };
```

addons/web/static/src/core/session/check_identity.js
```diff
@@ -170,6 +170,8 @@ export class CheckIdentity {
         // Only for internal user
         env.bus.addEventListener("WEB_CLIENT_READY", () => {
             if (session.device_salt) {
+                // Re-authentication feature with fingerprint is activated
+                this.patch_rpc();
```
aab-odoo: I would simply patch rpc outside of this class (top level of the module). That way, no need to bother checking if it has been patched already.

addons/web/static/src/core/session/check_identity.js
```diff
@@ -262,17 +307,20 @@ export class CheckIdentity {
...
         // not even inspecting the dom or through the console using Javascript.
-        this.env.services.action && this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});
+        if (reload) {
+            this.env.services.action && this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});
+        }
         await new Promise((resolve) => {
             this.bus.addEventListener("identityChecked", resolve, { once: true });
         });
         // Reload the view to display back the data that was displayed before.
+        if (!reload) return;
```
aab-odoo: by convention we force the brackets in JS
```suggestion
        if (!reload) {
            return;
        }
```

By the way, could you enable the tooling? `./addons/web/tooling/enable.sh`
There're a lot of eslint/prettier alerts in this file.

aab-odoo: It would be more readable like this imo (as we already have the check in the beginning of the function)
```suggestion
        if (reload) {
          if (this.env.services.action) {
            if (this.env.services.action.currentController) {
                this.env.services.action.doAction("soft_reload");
            } else {
                // In the case of multiple concurrent ``CheckIdentityException``
                // managed by differents dispatchers.
                // We get render from HTTP but error from JSONRPC.
                // ``soft_reload`` leaves an empty page.
                // Example: full refresh with an untrusted device.
                this.env.services.action.doAction("reload");
            }
        }
```

aab-odoo: Other refactoring suggestion: introduce a `checkIdentity` function:

```js
    checkIdentity() {
        if (!this.started) {
            this.dialogService.add(CheckIdentityDialog);
        }
        return new Promise((resolve) => {
            this.bus.addEventListener("identityChecked", resolve, { once: true });
        });
   }

    async run() { // reload param not needed anymore
        // Empty the current view, to not let any confidential data displayed
        // not even inspecting the dom or through the console using Javascript.
        this.env.services.action && this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});

        await this.checkIdentity();

        // Reload the view to display back the data that was displayed before.
        if (this.env.services.action) {
            if (this.env.services.action.currentController) {
                this.env.services.action.doAction("soft_reload");
            } else {
                // In the case of multiple concurrent ``CheckIdentityException``
                // managed by differents dispatchers.
                // We get render from HTTP but error from JSONRPC.
                // ``soft_reload`` leaves an empty page.
                // Example: full refresh with an untrusted device.
                this.env.services.action.doAction("reload");
            }
        }
    }
```

Then, you can directly call `checkIdentity` in the rpc patch.

addons/web/static/src/core/session/check_identity.js
```diff
@@ -255,23 +256,30 @@ export class CheckIdentity {
...
+        // Mutex
         await new Promise((resolve) => {
             this.bus.addEventListener("identityChecked", resolve, { once: true });
         });
+    }
+
+    async run() {
+        // Empty the current view, to not let any confidential data displayed
+        // not even inspecting the dom or through the console using Javascript.
+        this.env.services.action && this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});
```
aab-odoo: by convention, we rather write

```suggestion
        if (this.env.services.action) {
            this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});
        }
```

aab-odoo: Note that you can actually trigger the event all the time. If there's an action service (i.e. backend), it will empty the screen, and if there isn't, the event will be ignored.

A comment above mentioning that there may not be an action service (e.g. frontend) might be useful though

addons/web/static/src/core/session/check_identity.js
```diff
@@ -255,23 +256,30 @@ export class CheckIdentity {
...
 
-    async run() {
+    async checkIdentity() {
         if (!this.started) {
             this.dialogService.add(CheckIdentityDialog);
         }
-        // Empty the current view, to not let any confidential data displayed
-        // not even inspecting the dom or through the console using Javascript.
-        this.env.services.action && this.env.bus.trigger("ACTION_MANAGER:UPDATE", {});
+        // Mutex
```
aab-odoo: mutex? Was it a comment left not to forget to add a mutex? Or does it state that the following is mutexified? Because I don't think it is.

addons/web/static/src/core/session/check_identity.js
```diff
@@ -255,23 +256,30 @@ export class CheckIdentity {
...
         await new Promise((resolve) => {
             this.bus.addEventListener("identityChecked", resolve, { once: true });
         });
```
aab-odoo: An alternative to this would be to add a new callback props to CheckIdentityDialog, which it'll call once the identity is checked.
```js
        if (!this.started) {
            const { promise, resolve } = Promise.withResolvers();
            this.identityCheckPromise = promise;
            this.dialogService.add(CheckIdentityDialog, { onIdentityChecked: resolve });
        }
        return this.identityCheckPromise;
```

You could even use this.identityCheckPromise as a flag to determine whether or not there's already a pending identity check and get rid of started (+ the event handlers start/stop to set it).

addons/web/static/src/core/session/check_identity.js
```diff
@@ -320,3 +328,19 @@ export const checkIdentityService = {
...
+    const promise = originalPromise.catch(async (error) => {
+        if (error.data?.name === "odoo.http.session.CheckIdentityException") {
+            await odoo.__WOWL_DEBUG__.root.env.services["check_identity"].checkIdentity();
+            return originalRpc(...args);
+        }
+        throw error;
+    });
+    promise.abort = originalPromise.abort;
+    return promise;
+};
```
aab-odoo, line 358: not a fan of this, but I don't really see another way to replay the rpc s.t. the caller doesn't notice anything.

addons/web/static/src/core/session/check_identity.js
thle-odoo: The mutex behaves like an asynchronous synchronization barrier to wait the bus event `identityChecked` and so not continue the logic after `await checkIdentity()` without confirming the identity (manage the case of multiple requests which will be blocked with a `CheckIdentityException`).

In practice, it works (but I agree that it is not perfect); this part of the code was moved from the `auth_timeout` module ([commit](https://github.com/odoo/odoo/commit/87592b98d631a07563dda5e51c0b0f4102ca2a9b)).
If we receive any suggestions for improving the code, we can apply them.
Thanks for the suggestion (see: https://github.com/odoo/odoo/pull/248852#discussion_r3267329463)

addons/web/static/src/core/session/check_identity.js
thle-odoo: I have applied the changes in a separate commit: https://github.com/odoo/odoo/pull/248852/changes/be9a002188edaadf0dad43e98720d877729be8bf

addons/web/static/src/core/session/check_identity.js
thle-odoo: I think this is the best way to do it.

The other option is to directly incorporate this logic into the original rpc, but that has the disadvantage of not having all the logic in the same file.

addons/web/static/src/core/session/check_identity.js
aab-odoo: > I have applied the changes in a separate commit: [be9a002](https://github.com/odoo/odoo/commit/be9a002188edaadf0dad43e98720d877729be8bf)

Ok thanks. If you think it's better like this, keep it. If you think it was better before, you can remove the commit :-)

thle-odoo: The only difference is that we no longer call the `close` function of the `CheckIdentityForm` component, since we destroy it using a "top-down" approach via `closeDialog` during `identityCheckCleanUp` (and the `redirect` prop is not used). I'm leaving this function in anyway, as this component could be reused elsewhere (because it is a `public_components`).

```diff
-        this.checkIdentityService.bus.addEventListener("identityChecked", () => {
-            this.close();
-        });
```

If the code is OK (?), I will keep the commit; the code is objectively better (and simpler).

aab-odoo: Yep, that looks fine :-) Thanks

aab-odoo: I didn't test it though, and I've a little concern: can you close the dialog without validating the check? e.g. with ESC or by clicking on the X? Because if yes, I think you won't go through your cleanup callback, so the promise will remain and you won't be able to open the dialog again.

thle-odoo: We can close the dialog box by pressing ESC and clicking on the X.
However, this will perform a logout and a redirect:
```js
this.env.dialogData.dismiss = async () => {
    const url = await post("/web/session/logout", { csrf_token: odoo.csrf_token }, "url");
    redirect(url);
};
```
The redirect triggers a full page reload, and the entire JavaScript context is destroyed.
As a result, the framework will re-execute the logic in the `check_identity.js` file, register the `CheckIdentity` service, and so on.
The result is that the state of the service will be properly reset to the values defined in the `setup`:
```js
this.identityCheckPromise = null;
this.identityCheckCleanUp = null;
```

> I think you won't go through your cleanup callback, so the promise will remain and you won't be able to open the dialog again.

We don't go through the cleanup, but the promise won't be kept.

Based on my investigation, the behavior is correct and won't cause any issues.
_Feel free to correct me if necessary_

aab-odoo: That makes sense :+1:

thle-odoo:
The double-rendering issue (see attachment) stems from a problem with determining the `rootId` if there is a shadow-DOM.
This behavior was introduced by the commit: https://github.com/odoo/odoo/commit/b3f0ca15db74e9dbb19020b0cb828fb90f21d2a4

This behavior is therefore independent of the commits in this PR.

The Discuss team is aware of this.
CC: @tsm-odoo

two_windows

UPDATE:
Fixed with the commit: https://github.com/odoo/odoo/commit/4279b7f2bbc76bfe069cc1d9300aa34dafe40063

beledouxdenis:
robodoo delegate=thle-odoo

thle-odoo:
robodoo rebase-ff

thle-odoo:
robodoo r+

PR #248855, opened 2026-02-16
[FIX] iot_drivers: scale compatibility for <v19

Ysoroko:
This PR adds the scale commands compatibility for the iot box images in >= saas-19.1 with the dbs in v18.0 and less

In v18.0 we read 'value' key sent by the iot box but the iot box doesn't send it anymore in 19.1
This PR makes the iot box send both 'result' as exepcted in v19 and 'value' expected in v18

Enterprise PR: https://github.com/odoo/enterprise/pull/107641


loouis-t (approved):
robodoo r+

PR #248857, opened 2026-02-16
[FIX] web: tooltip_service: do not kill tooltip if still in target

kebeclibre:
Have a tooltip on a parent.
Hover on a child of that parent.
Now, leave the child but stay in parent.

Before this commit, the tooltip would be killed and never respawn.

After this commit, the tooltip is not even killed if we stayed within the parent's physical space.

task-5346498

Current behavior before PR:

---

### DIFF
--- a/addons/web/static/tests/core/tooltip/tooltip_service.test.js
+++ b/addons/web/static/tests/core/tooltip/tooltip_service.test.js
@@ -386,3 +386,40 @@ test("touch rendering - tap-to-show", async () => {
     await animationFrame();
     expect(".o_popover").toHaveCount(0);
 });
+
+test.tags("desktop");
+test("tooltip from and to child element", async () => {
+    class MyComponent extends Component {
+        static props = ["*"];
+        static template = xml`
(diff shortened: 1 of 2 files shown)


addons/web/static/tests/core/tooltip/tooltip_service.test.js
```diff
@@ -386,3 +386,36 @@ test("touch rendering - tap-to-show", async () => {
     await animationFrame();
     expect(".o_popover").toHaveCount(0);
 });
+
+test.tags("desktop");
+test("tooltip from and to child element", async () => {
+    class MyComponent extends Component {
+        static props = ["*"];
+        static template = xml`<div class="no-tooltip">space</div><div class="p-5 parent-element" data-tooltip="hello"><button>Action</button></div>`;
```
adr-odoo: `.parent-element` is needed?

adr-odoo (approved):

adr-odoo:
@robodoo r+

PR #246809, opened 2026-02-02
[IMP] l10n_sk: P&L + Balance Sheet

aizu-odoo:
Adding the Slovakian P&L and Balance Sheet reports, as well as the combined version under Annual Statements with the xml for form UZPODv14.

task-5105102

enterprise/pull/97156

PR #248859, opened 2026-02-16
[IMP] l10n_sk: P&L + Balance Sheet

aizu-odoo:
Adding the Slovakian P&L and Balance Sheet reports, as well as the combined version under Annual Statements with the xml for form UZPODv14.

odoo/enterprise/pull/107535

task-5105102

Back-port of: odoo/odoo#246809


jco-odoo (approved):

cawo-odoo:
Please be sure to follow the process described in https://github.com/odoo/upgrade/wiki/Upgrade-exceptions
If and when you're confident that this PR will need no more changes, we can create the FW ports already and when all are "green", merge all at once.

PR #248860, opened 2026-02-16
[FIX] web_editor: improve performance by simplifying dirty check

walidsahli:
Problem:
When editing heavy content, each interaction can take up to 600 ms, sometimes freezing the editor completely.

Cause:
The `_isDirty` check is computationally expensive and is executed on every history step. This is unnecessary and leads to poor performance when handling large documents.

Solution:
Instead of recomputing `_isDirty` on each history step, directly mark the field as dirty whenever a history step is triggered. This provides a simpler and more efficient approach.

Before fix:
Each keyboard interaction to next paint takes around 200ms.
image

After fix:
Each keyboard interaction to next paint takes around 40ms.
image

opw-5892291

---


walidsahli:
@robodoo up to saas-18.4

PR #248869, opened 2026-02-16
[FIX] hr_expense: skip encrypted files from expenses report PDF

kitan191:
**Steps to reproduce:**
- Install hr_expense
- Create an expense
- Attach several receipts with at least one encrypted PDF
- Create Report
- From cog menu, print "Expenses Report"

**Issue:**
A traceback is raised while trying to add the attachments to the report: "File has not been decrypted".

**Cause:**
The encrypted file cannot be appended to the report PDF.

**Solution:**
Skip encrypted files and log a note with the reason in the report.

opw-5194501

---

### DIFF
--- a/addons/hr_expense/models/ir_actions_report.py
+++ b/addons/hr_expense/models/ir_actions_report.py
@@ -1,7 +1,7 @@
 import io
-from odoo import models
+from odoo import models, _
 from odoo.tools import pdf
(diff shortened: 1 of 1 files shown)


hupo-odoo (approved):
@robodoo r+
thanks for the fix!

PR #248871, opened 2026-02-16
[FIX] sale: resize portal SO section colspan

mano-odoo:
The colspan computation on the SO section is broken on mobile. This is because the unit price column is not rendered < sm viewports, and the taxes column is not rendered < md viewports. Thus we need to adapt the colspan accordingly to these responsive columns.

task-5942576

---

### DIFF
--- a/addons/sale/views/sale_portal_templates.xml
+++ b/addons/sale/views/sale_portal_templates.xml
@@ -782,10 +782,26 @@
                                         <t t-if="not show_section_total">
                                             <t t-set="section_name_colspan" t-value="99"/>
                                         </t>
+                                        <!-- Some columns (e.g. Discount and Unit Price)
+                                        are not displayed on smaller viewports -->
+                                        <td
+                                            name="td_section_name_xs"
+                                            t-att-colspan="section_name_colspan - 1 - (1 if display_discount else 0) - (1 if display_taxes else 0)"
+                                            t-attf-class="d-sm-none {{padding_class}}"
+                                        >
+                                            <span t-field="line.name"/>
+                                        </td>
+                                        <td
+                                            name="td_section_name_sm"
+                                            t-att-colspan="section_name_colspan - (1 if display_taxes else 0)"
+                                            t-attf-class="d-none d-sm-table-cell d-md-none {{padding_class}}"
+                                        >
+                                            <span t-field="line.name"/>
+                                        </td>
                                         <td
                                             name="td_section_name"
                                             t-att-colspan="section_name_colspan"
-                                            t-att-class="padding_class"
+                                            t-attf-class="d-none d-md-table-cell {{padding_class}}"
                                         >
                                             <span t-field="line.name"/>
                                         </td>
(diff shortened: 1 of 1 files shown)


addons/sale/views/sale_portal_templates.xml
```diff
@@ -782,10 +782,26 @@
                                         <t t-if="not show_section_total">
                                             <t t-set="section_name_colspan" t-value="99"/>
                                         </t>
+                                        <!-- Some columns eg display_discount and Unit Price
+                                        are not displayed on smaller viewports -->
+                                        <td
+                                            name="td_section_name_xs"
+                                            t-att-colspan="section_name_colspan - 2"
```
nipl-odoo: We don't show discount, taxes and unit price so it should be
First -1 is for unit price which is always hidden for XS screens
```js
t-att-colspan="section_name_colspan - 1 - (1 if display_discount else 0) - (1 if display_taxes else 0)"
```

addons/sale/views/sale_portal_templates.xml
```diff
@@ -782,10 +782,26 @@
...
+                                            t-attf-class="d-sm-none {{padding_class}}"
+                                        >
+                                            <span t-field="line.name"/>
+                                        </td>
+                                        <td
+                                            name="td_section_name_sm"
+                                            t-att-colspan="section_name_colspan - 1"
```
nipl-odoo: For these screens we don't show taxes only so it should be
```js
t-att-colspan="section_name_colspan - (1 if display_taxes else 0)"
```

addons/sale/views/sale_portal_templates.xml
nipl-odoo: These 3 columns also need to be handled for smaller screens we need to show hidden section lines just like product lines. Maybe applying `d-none d-sm/md-table-cell` would be just fine 👀 
(Just in case you don't know, you can get here by clicking on `Hide Composition` on some section)

nipl-odoo (changes requested):
Hello Mathieu 👋🏻
I’m not convinced we need to bring JS into the picture just to deal with colspans 🤔 feels a bit overkill IMO. The current solution looks good overall, apart from the two issues I pointed out below. I’ve left comments directly on the tricky cases. 😄

As for the stable question, I think this fix should be stable-friendly, but I’ll let VFE be the judge of that.

Also, just a small heads-up: please make sure not to break the `section_name_colspan` xpath in `/sale_subscription/views/sale_subscription_portal_templates.xml`. It shouldn’t be impacted by my current suggestion, but flagging it in case you decide to explore another route 👀

 - For normal sections but with discount column
image
 
- For Hidden sections
Image

nipl-odoo (approved):
Thank you for your work 😄 Looks good to me! 
@Feyensv Do you think this fix can land in stable? 👀 We are just adding new `td`s so it's safe I guess

addons/sale/views/sale_portal_templates.xml
```diff
@@ -782,10 +782,26 @@
                                         <t t-if="not show_section_total">
                                             <t t-set="section_name_colspan" t-value="99"/>
                                         </t>
+                                        <!-- Some columns eg display_discount and Unit Price
+                                        are not displayed on smaller viewports -->
```
Feyensv:
```suggestion
                                        <!-- Some columns (e.g. Discount and Unit Price)
                                        are not displayed on smaller viewports -->
```

`display_disount` is not a column, it's only a variable meant to know whether the discount is displayed or not :eyes:

Feyensv (approved):
Small nitpicking on the comment but LGTM (though I recognize I don't get how it works :D).

Thanks for the fix (and @nipl-odoo for the review :pray:)

robodoo delegate+

PR #248882, opened 2026-02-16
[FIX] l10n_es_edi_facturae: use currency dependent rounding for amounts

agr-odoo:
Steps to reproduce:
- Have facturae modules installed
- Generate invoice with any amount
- Send to Facturae

Issue:
Resulting XML has 8 digits after the decimal point on several fields, such as unit price, gross amount, and total cost.

When trying to validate such an XML, this results in validation error: "RCF06001: En facturas emitidas en euros, alguno de los importes de las líneas tiene más de dos decimales (regla 6a del anexo II de la Orden HAP/1650/2015)."

According to regulation HAP/1650/2015 [1]: For invoices issued in euros, it will be validated that the total line amounts related to the total cost are numeric and rounded, according to the common rounding method, to two decimal places.
This commit introduces dynamic decimal precision: 2 places for EUR and 8 places (the previous default) for other currencies.

[1] https://www.boe.es/diario_boe/txt.php?id=BOE-A-2015-8844
Machine translated [BOE-A-2015-8844 (1).pdf](https://github.com/user-attachments/files/25345382/BOE-A-2015-8844.1.pdf)

opw-5927356

### DIFF
--- a/addons/l10n_es_edi_facturae/data/facturae_templates.xml
+++ b/addons/l10n_es_edi_facturae/data/facturae_templates.xml
@@ -135,13 +135,13 @@
                 <Quantity t-out="line['Quantity']"/>
                 <UnitOfMeasure t-out="line.get('UnitOfMeasure')"/>
                 <UnitPriceWithoutTax t-out="float_repr(refund_multiplier*line['UnitPriceWithoutTax'], 8)"/>
-                <TotalCost t-out="float_repr(refund_multiplier*line['TotalCost'], 8)"/>
+                <TotalCost t-out="float_repr(refund_multiplier*line['TotalCost'], file_currency.decimal_places)"/>
                 <DiscountsAndRebates t-if="line.get('DiscountsAndRebates')">
                     <t t-foreach="line['DiscountsAndRebates']" t-as="discount">
                         <Discount>
                             <DiscountReason t-out="discount['DiscountReason']"/>
                             <DiscountRate t-out="discount.get('DiscountRate')"/>
-                            <DiscountAmount t-out="float_repr(refund_multiplier*discount['DiscountAmount'], 8)"/>
+                            <DiscountAmount t-out="float_repr(refund_multiplier*discount['DiscountAmount'], file_currency.decimal_places)"/>
                         </Discount>
                     </t>
                 </DiscountsAndRebates>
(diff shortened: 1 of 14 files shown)


yosuanicolaus (approved):
Looks good to me then. 👍 

Sending to final review

addons/l10n_es_edi_facturae/data/facturae_templates.xml
```diff
@@ -150,11 +150,11 @@
                         <Charge>
                             <ChargeReason t-out="charge['ChargeReason']"/>
                             <ChargeRate t-out="charge.get('ChargeRate')"/>
-                            <ChargeAmount t-out="float_repr(refund_multiplier*charge['ChargeAmount'], 8)"/>
+                            <ChargeAmount t-out="float_repr(refund_multiplier*charge['ChargeAmount'], file_currency.decimal_places)"/>
                         </Charge>
                     </t>
                 </Charges>
-                <GrossAmount t-out="float_repr(refund_multiplier*line['GrossAmount'], 8)"/>
+                <GrossAmount t-out="float_repr(refund_multiplier*line['GrossAmount'], file_currency.decimal_places)"/>
```
smetl, line 157: Overall this doesn't look incorrect when looking at the XSD saying
```xml
<xs:element name="GrossAmount" type="DoubleUpToEightDecimalType">
<xs:annotation>
<xs:documentation xml:lang="en"> Result: TotalCost - DiscountAmount + ChargeAmount. Up to eight decimal points.</xs:documentation>
<xs:documentation xml:lang="es"> Importe bruto. Resultado: TotalCost - DiscountAmount + ChargeAmount. Hasta ocho decimales.</xs:documentation>
</xs:annotation>
</xs:element>
```
same for the others

smetl (commented):
I'm a bit confused... Looking at the xsd, everything seems correct in the original code. Regarding your explanation, it seems it's accepted in some platform but not all of them... Idk what to say then 😅

addons/l10n_es_edi_facturae/data/facturae_templates.xml
chklop: Only UnitPriceWithoutTax is allowed to have a precision of more than 2 decimals. Other amounts can be written with more but if they do, they must all be 0.

image

In this example Unit price is allowed to have more precision. TotalCost and GrossAmount are only accepted with more precision because they are all 0.
image

This fix is correct.

smetl (approved):
@robodoo r+

PR #248885, opened 2026-02-16
[FIX] l10n_fr_hr_holidays: compute time off for partial time employee

maedekey:
When setting time off for a french employee whose working schedule is part time, some computations aren't performed correctly. Indeed, when we create a new working schedule, of type Fully Fixed, but we remove a day from the schedule, it will be considered as part time.
According to the french law:
https://code.travail.gouv.fr/fiche-service-public/quelle-est-lincidence-sur-les-droits-a-conges-payes-dun-salarie-a-temps-partiel when in this case, we create 2 days of time off on thursday and friday, 3 days are supposed to be deducted from the time off balance of the employee. Currently, only 2 days are deducted.

## Reproduction Steps
1. Create a part time working schedule for a french company, where the non-working days are Wednesday, Saturday
and Sunday.
2. Create an employee to which you assign this working schedule.
3. Go to Time Off. In Settings, in the French Time Off Localization block, make sure that the Company Paid Time Off Type is set to Paid Time Off. Make also sure that the Company Working Hours is srt at a full-time schedule, NOT to the part-time schedule you just created.
4. On the Management type, click Allocations. Create a new allocation for the employee you just created and validate it.
5. Click on Management > Time off. Click on new. Select the employee you just created and Paid Time off as Time Off Type.
6. Select the start date to be a Thursday and an end date to be a Friday.

### Expected behavior
In compliance with the French law about part-time working schedules, the duration of the time off should be 3 days.

### Unexpected behavior
The duration shown is 2 days.
Moreover, if we select an entire week, the duration will be 5 days instead of 6.

## Origin of the issue
In `_get_durations()` of the module, we haven't implemented the feature that takes into account that specific french law for part-time workers.

__
opw-5911616

### DIFF
--- a/addons/l10n_fr_hr_holidays/models/hr_leave.py
+++ b/addons/l10n_fr_hr_holidays/models/hr_leave.py
@@ -153,18 +153,15 @@ class HrLeave(models.Model):
                     date_end = leave.date_to
                     while not leave.resource_calendar_id._works_on_date(date_start):
                         date_start += relativedelta(days=1)
-                    extended_date_end = date_end
-                    while not company_cal._works_on_date(extended_date_end + relativedelta(days=1)):
-                        extended_date_end += relativedelta(days=1)
                     # Count number of days in company calendar
                     current = date_start.date()
-                    end_date = extended_date_end.date()
+                    end_date = date_end.date()
                     legal_days = 0.0
                     while current <= end_date:
                         if current in holidays_days_list:
                             current += relativedelta(days=1)
                             continue
-                        if company_cal._works_on_date(current):
+                        if current.weekday() < 6:
                             legal_days += 1.0
                         current += relativedelta(days=1)
                     standard_duration = super()._get_durations(resource_calendar=resource_calendar)
--- a/addons/l10n_fr_hr_holidays/tests/test_french_leaves.py
+++ b/addons/l10n_fr_hr_holidays/tests/test_french_leaves.py
@@ -581,3 +581,105 @@ class TestFrenchLeaves(TransactionCase):
         self.assertEqual(leave.date_from.date(), date(2026, 1, 12))
         self.assertEqual(leave.date_to.date(), date(2026, 1, 12))
         self.assertEqual(leave.number_of_hours, 8.0, 'Duration should be 8 hours (one working day)')
+
+    def test_employee_partial_time_holidays(self):
+        """checks that days off are correctly computed for employees with partial time schedules.
+        Law can be found here: https://code.travail.gouv.fr/fiche-service-public/quelle-est-lincidence-sur-les-droits-a-conges-payes-dun-salarie-a-temps-partiel"""
+        part_time_calendar = self.env['resource.calendar'].create({
+            'name': 'Part Time Calendar',
+            'company_id': self.company.id,
+            'attendance_ids': [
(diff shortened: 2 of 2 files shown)


yoba-odoo (approved):
Seems okay to me

addons/l10n_fr_hr_holidays/models/hr_leave.py
```diff
@@ -154,8 +154,12 @@ def _get_durations(self, check_leave_type=True, resource_calendar=None):
                     while not leave.resource_calendar_id._works_on_date(date_start):
                         date_start += relativedelta(days=1)
                     extended_date_end = date_end
-                    while not company_cal._works_on_date(extended_date_end + relativedelta(days=1)):
-                        extended_date_end += relativedelta(days=1)
+                    if not leave.resource_calendar_id.is_fulltime:
+                        if date_end.weekday() < 6 and not leave.resource_calendar_id._works_on_date(date_end + relativedelta(days=1)):
+                            date_end += relativedelta(days=1)
```
romaincarlier4: This will do nothing as `date_end` is not used afterwards

addons/l10n_fr_hr_holidays/models/hr_leave.py
```diff
@@ -164,7 +168,7 @@ def _get_durations(self, check_leave_type=True, resource_calendar=None):
                         if current in holidays_days_list:
                             current += relativedelta(days=1)
                             continue
-                        if company_cal._works_on_date(current):
+                        if (leave.resource_calendar_id.is_fulltime and leave.resource_calendar_id._works_on_date(current)) or (not leave.resource_calendar_id.is_fulltime and current.weekday() < 6):
```
romaincarlier4: Are you sure about this ? Looks like we wanted to add one day if the company calendar was working on that day. Also, I think the behavior will be wrong if you are part time, do not work on friday and ask for time off on monday-tuesday

romaincarlier4 (commented):
Thanks for the fix! Some questions

addons/l10n_fr_hr_holidays/models/hr_leave.py
maedekey: Indeed, I got confused with extended_date_end. And even so, the condition will never be met as the date_end is already updated correctly in `_compute_date_from_to`.

addons/l10n_fr_hr_holidays/models/hr_leave.py
maedekey: I am sure about this. In the link I have sent, the law mentions "jours ouvrables", which is in France from Monday to Saturday. There's a mention saying that the employer can compute days off based on "jours ouvrés", which would correspond to the company schedule, but it says that in this case, that way of computing days off has to guarantee to the employee right to days off at least equal to those computed based on opening days: 
>Dans ce cas, ce mode de calcul doit garantir au salarié des droits à congés au moins égaux à ceux calculés en jours ouvrables.

It also mentions that the computation of paid time off for the employee having a partial time schedule has to take into account the first day of the day off, and also all the opening days included in the absence period until the resumption of work:
>Le décompte des jours de congés payés du salarié travaillant à temps partiel est effectué de la manière suivante par l'employeur :
> - Prise en compte du 1er jour de départ en congé
> - Et prise en compte de tous les jours ouvrables inclus dans la période d'absence jusqu'à la reprise du travail.

Finally I have tried to create 2 days off on Monday and Tuesday for an employee who doesn't work on Friday, Saturday and Sunday. It only registers 2 days off, which is in compliance with the law above.

PR #233965, opened 2025-10-31
[IMP] base: show account_holder_name in res.partner.bank form view

Mohamed-Dallash:
Allow changing of the account holder name in case it is different from the partner name which was the computed default.

Task-5222712

[Related PR](https://github.com/odoo/enterprise/pull/98572)

PR #248886, opened 2026-02-16
[FIX] base: allow changing contact name without changing account holder name

HazemWaheed2050:
**steps to reproduce:**
1. Create a contact
2. On the accounting tab, add one bank account
3. Change the name of the contact
4. The record's name updates automatically the field "Account Holder Name"

**problem:**
When changing a profile's name in the contacts app, the account holder name changes to be the same as the profile's name.

**expected behaviour:**
"Make the account holder name update if partner.name is changed ONLY if they were the same before the change. Otherwise keep the account holder name as is." - Referenced from the feature ticket: 5222712

That means if contact's name is the same as the account holder name, then any change to the contact's name, should as well update 'account holder name' accordingly, but if you changed the account holder name itself, then you change contact's name (if they're not the same) then the account holder name shouldn't be updated to match the contact's name.

**cause:**
The account holder name depends on `partner_id.name`
https://github.com/odoo/odoo/blob/2394a9a7f0ece2d5a2185fa4e715d943fc044733/odoo/addons/base/models/res_bank.py#L104-L107

So whenever you change the contact's name, the account holder name changes accordingly.

**Fix:**
Upon creation, the account holder name is set to the partner name if no name is provided. Additionally, when the partner name changes, the account holder name is updated only if it previously matched the partnername; otherwise, it is left unchanged.

**NOTE:**
This is a backport of this PR: https://github.com/odoo/odoo/pull/233965

opw-5913464

---

### DIFF
--- a/addons/portal/tests/test_portal.py
+++ b/addons/portal/tests/test_portal.py
@@ -23,6 +23,13 @@ class TestUsersHttp(HttpCase):
             'acc_type': 'bank',
         })
 
+        bank_account_2 = self.env['res.partner.bank'].create({
+            'acc_number': '987654321',
+            'partner_id': portal_user.partner_id.id,
+            'acc_holder_name': 'Partner A',
+            'acc_type': 'bank',
+        })
+
         common_data = {
             'phone': '1234567890',
             'email': 'test@example.com',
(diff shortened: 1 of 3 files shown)


addons/portal/tests/test_portal.py
```diff
@@ -38,11 +38,6 @@ def test_account_holder_name_update(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(bank_account.acc_holder_name, 'Partner A Holder')
 
-        # request 2: request with changed partner name
-        response2 = self.url_open(url='/my/account', data={**common_data, 'name': 'Partner New Name', 'csrf_token': Request.csrf_token(self)})
-        self.assertEqual(response2.status_code, 200)
-        self.assertEqual(bank_account.acc_holder_name, 'Partner New Name')
```
kitan191, line 44: Instead of removing this part of the test, you can add a second bank account with its `acc_holder_name` set to the user name instead.
Then you can test that `acc_holder_name` doesn't change for the first one but change for the second one.

kmagusiak (approved):
LGTM

alialfie (approved):
LGTM, thanks!
robodoo r+

PR #248889, opened 2026-02-16
[FIX] sale: round so global discounted amount

mma1377:
Issue:
---
Global discount is not rounded as SOLs amounts, resulting non-zero total amount when price 100% global discount is applied.

Steps to reproduce:
---
1- Create a SO.
2- Add two lines:
   - qty:4, price_unit:2.8207, tax:15%
   - qty:5, price_unit:2.8207, tax:15%

3- Using Discount action button, apply 100% global discount.

As you see the total amount is not zero.

Cause:
---
Sol's `price_total`, `price_subtotal` are rounded to currency decimals `_add_tax_details_in_base_line`:
https://github.com/odoo/odoo/blob/ef3c443837b02badc9e20bdec7544c34ed402ef3/addons/account/models/account_tax.py#L1551-L1560

However, the same rounding is not applied in discount calculation, leading to this mismatch.

opw-5886749

### DIFF
--- a/addons/sale/wizard/sale_order_discount.py
+++ b/addons/sale/wizard/sale_order_discount.py
@@ -123,7 +123,9 @@ class SaleOrderDiscount(models.TransientModel):
             taxes = line.tax_id.flatten_taxes_hierarchy()
             fixed_taxes = taxes.filtered(lambda t: t.amount_type == 'fixed')
             taxes -= fixed_taxes
-            total_price_per_tax_groups[taxes] += line.price_unit * (1 - (line.discount or 0.0) / 100) * line.product_uom_qty
+            total_price_per_tax_groups[taxes] += self.company_id.currency_id.round(
+                line.price_unit * (1 - (line.discount or 0.0) / 100) * line.product_uom_qty
+            )
 
         discount_dp = self.env['decimal.precision'].precision_get('Discount')
         context = {'lang': self.sale_order_id._get_lang()}  # noqa: F841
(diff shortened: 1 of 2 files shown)


addons/sale/wizard/sale_order_discount.py
```diff
@@ -123,7 +123,9 @@ def _create_discount_lines(self):
             taxes = line.tax_id.flatten_taxes_hierarchy()
             fixed_taxes = taxes.filtered(lambda t: t.amount_type == 'fixed')
             taxes -= fixed_taxes
-            total_price_per_tax_groups[taxes] += line.price_unit * (1 - (line.discount or 0.0) / 100) * line.product_uom_qty
+            total_price_per_tax_groups[taxes] += self.company_id.currency_id.round(
+                line.price_unit * (1 - (line.discount or 0.0) / 100) * line.product_uom_qty
```
Feyensv, line 127:
```suggestion
                line._get_discounted_price() * line.product_uom_qty
```

Also, what's the difference with `price_subtotal` then ?

Feyensv (commented):
I fear this might not be so simple to fix (considering the deep changes @smetl had to do to fix tax issues here in 19).

@smetl is this the right approach or should we simply tell the client it's a technical limitation in stable ?

addons/sale/wizard/sale_order_discount.py
smetl: - You need to round using the foreign currency, not the one set on the company
- @Feyensv I guess it's to manage the case when the price is tax included.
- Whatever you do, this is a lost cause because there is no way to solve all the cases mathematically speaking. By rounding here, you might create new bugs for a different setup with others prices, others taxes, others percentages.

PR #248141, opened 2026-02-11
[IMP] mail:  Hide files and pinned messages panel when search is open

jayp-odoo:
When the search view is opened in Chatter, files and pinned messages panels are now hidden to avoid UI clutter and improve focus on search results.

task-[5873731](https://www.odoo.com/odoo/project/1519/tasks/5873731)

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248890, opened 2026-02-16
[FIX] mail: stabilize chatter search panels test due to thread not loaded

jayp-odoo:
The test "opening search in chatter hides files and pinned messages panels"
was flaky due to a race condition.

The attachment box toggle can be clicked before attachments are loaded, 
so the click is ignored and the assertion on `.o-mail-AttachmentBox` times out.

Fix by waiting for deterministic preconditions before toggling:
- wait for attach-files counter to display `1`
 
issue generated from this [PR](
https://github.com/odoo/odoo/pull/248141
)

---

### DIFF
--- a/addons/mail/static/tests/chatter/web/chatter_search.test.js
+++ b/addons/mail/static/tests/chatter/web/chatter_search.test.js
@@ -95,7 +95,7 @@ test("opening search in chatter hides files and pinned messages panels", async (
     });
     await start();
     await openFormView("res.partner", partnerId);
-    await click("button[aria-label='Attach files']");
+    await click("button[aria-label='Attach files']:text('1')");
     await click("button[title='Pinned Messages']");
     await contains(".o-mail-AttachmentBox");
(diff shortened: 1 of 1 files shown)


seb-odoo (approved):
It would be nice to have a more specific title.

@robodoo delegate+ priority

Priority as breaking staging (I might need to re-do it on actual r+)

addons/mail/static/tests/chatter/web/chatter_search.test.js
```diff
@@ -95,7 +95,7 @@ test("opening search in chatter hides files and pinned messages panels", async (
     });
     await start();
     await openFormView("res.partner", partnerId);
-    await click("button[aria-label='Attach files']");
+    await click("button[aria-label='Attach files']:text('1')");
```
seb-odoo, line 98: Note: `Attach files` opens upload rather than box if thread data are not yet loaded.

seb-odoo (approved):
@robodoo r+ priority

PR #248894, opened 2026-02-16
[REF] mail: add resilience to hasActionsMenu

jeanschoenlaub:
In preparation for owl3, we had to manually refactor the hasActionsMenu which was a bit fragile.

BEFORE:
hasActionsMenu uses this.partitionedAction even if partitionedAction is a template variable and not a componenent variable.

NOW:
hasActionsMenu uses this.threadActions.partition,
an actual component variable

task: OWL3 prep - add this. to template variables

Current behavior before PR:

---

### DIFF
--- a/addons/mail/static/src/core/common/chat_window.js
+++ b/addons/mail/static/src/core/common/chat_window.js
@@ -64,11 +64,12 @@ export class ChatWindow extends Component {
     }
 
     get hasActionsMenu() {
+        const partition = this.threadActions.partition;
         return (
-            this.partitionedActions.group.length > 0 ||
-            this.partitionedActions.other.length > 0 ||
-            (this.ui.isSmall && this.partitionedActions.quick.length > 2) ||
-            (!this.ui.isSmall && this.partitionedActions.quick.length > 3)
+            partition.group.length > 0 ||
+            partition.other.length > 0 ||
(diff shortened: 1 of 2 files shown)


addons/mail/static/src/core/common/chat_window.js
```diff
@@ -65,10 +65,10 @@ export class ChatWindow extends Component {
...
     get hasActionsMenu() {
         return (
-            this.partitionedActions.group.length > 0 ||
-            this.partitionedActions.other.length > 0 ||
-            (this.ui.isSmall && this.partitionedActions.quick.length > 2) ||
-            (!this.ui.isSmall && this.partitionedActions.quick.length > 3)
+            this.threadActions.partition.group.length > 0 ||
+            this.threadActions.partition.other.length > 0 ||
+            (this.ui.isSmall && this.threadActions.partition.quick.length > 2) ||
+            (!this.ui.isSmall && this.threadActions.partition.quick.length > 3)
```
alexkuhn: I would put `this.threadActions.partition` in a `const`, as the `.partition` builds an object based on condition of each action in a registry, so this is more efficient to do it once.

The reason the template has a `t-set` and uses it was because of concerns to build the object of `.partition` once, for more efficient code.

alexkuhn (approved):
LGTM otherwise!

@robodoo delegate+

PR #248895, opened 2026-02-16
[FIX] account_peppol: Avoid commercial_partner_id being empty

vmac-odoo:
Before this commit, when commercial_partner_id is on the view (possible with web_studio), the value by default is False.

When the autocomplete widget is used, many fields could be autofilled and raise _onchange_verify_peppol_status, that requires this field.

To avoid this issue we review that the value has been filled.

Steps to Reproduce:
1. Open the Contacts app
2. Open Studio on the contact form view
3. Add the field commercial_partner_id to the form view (make it visible)
4. Create a new contact
5. Type a name
6. Select a suggestion from the IAP autocomplete
7. An error is raised immediately

OPW-[5896847](https://www.odoo.com/odoo/action-4043/5896847)

---


clbr-odoo (approved):
@robodoo r+ squash

PR #248897, opened 2026-02-16
[FIX] point_of_sale: prioritize server date for when pos is online

pedrambiria:
Before this commit, if the PoS device had a wrong date, the orders created while being online would have a wrong date_order. This commit prioritizes the server date for the order that is currently being processed when the PoS is online.

opw-5884498

---

### DIFF
--- a/addons/point_of_sale/models/pos_order.py
+++ b/addons/point_of_sale/models/pos_order.py
@@ -86,6 +86,10 @@ class PosOrder(models.Model):
         if not order.get('company_id'):
             order['company_id'] = pos_session.config_id.company_id.id
 
+        if self.env.context.get('current_order_uuid') and order['uuid'] == self.env.context['current_order_uuid']:
+            # Prioritize the server date for the order that is currently being processed
+            order['date_order'] = fields.Datetime.now()
+
         pos_order = False
         record_uuid_mapping = order.pop('relations_uuid_mapping', {})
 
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/models/pos_order.py
```diff
@@ -121,8 +125,8 @@ def _process_order(self, order, existing_order):
 
         for model_name, mapping in record_uuid_mapping.items():
             owner_records = self.env[model_name].search([('uuid', 'in', mapping.keys())])
-            for uuid, fields in mapping.items():
-                for name, uuids in fields.items():
+            for uuid, field_names in mapping.items():
+                for name, uuids in field_names.items():
```
davidmonnom, line 140: That's not related to this PR, try to avoid this :)

davidmonnom (commented):
Isn't easier to just assign date_order server side in sync_from_ui when the order isn't yet known ?

addons/point_of_sale/models/pos_order.py
pedrambiria: It's needed because I want to use fields.Datetime.now()

pedrambiria: > Isn't easier to just assign date_order server side in sync_from_ui when the order isn't yet known ?

I didn't quite catch your meaning. I did this because I want to preserve the original date for orders that are created while offline.

davidmonnom (approved):
@robodoo r+

PR #248960, opened 2026-02-17
[IMP] l10n_in: optimize tax tag lookup

hash-odoo:
Before this commit:
- `_get_l10n_in_tax_tag_ids()` was executed for moves unrelated to Indian localization, causing unnecessary processing.
- Profiling showed repeated `env.ref` lookups (>≈1.6% transactions each) even for a simple invoice flow (create → add products → post).

After this commit:
- Restrict execution to Indian-specific moves only.
- Replace `env.ref` with `_xmlid_to_res_id` for faster XMLID resolution.

task-5935607

### DIFF
--- a/addons/l10n_in/models/account_invoice.py
+++ b/addons/l10n_in/models/account_invoice.py
@@ -721,6 +721,8 @@ class AccountMove(models.Model):
 
     def _get_sync_stack(self, container):
         stack, update_containers = super()._get_sync_stack(container)
+        if all(move.country_code != 'IN' for move in self):
+            return stack, update_containers
         _tax_container, invoice_container, misc_container = update_containers()
         moves = invoice_container['records'] + misc_container['records']
         stack.append((9, self._sync_l10n_in_gstr_section(moves)))
--- a/addons/l10n_in/models/account_move_line.py
+++ b/addons/l10n_in/models/account_move_line.py
@@ -83,16 +83,17 @@ class AccountMoveLine(models.Model):
         return False
 
     def _get_l10n_in_tax_tag_ids(self):
-
-        def get_tag_ids(*refs):
-            return [self.env.ref(ref).id for ref in refs]
-
+        xmlid_to_res_id = self.env['ir.model.data']._xmlid_to_res_id
+        tag_refs = {
+            'sgst': ['l10n_in.tax_tag_base_sgst', 'l10n_in.tax_tag_sgst'],
+            'cgst': ['l10n_in.tax_tag_base_cgst', 'l10n_in.tax_tag_cgst'],
+            'igst': ['l10n_in.tax_tag_base_igst', 'l10n_in.tax_tag_igst'],
+            'cess': ['l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'],
+            'eco_9_5': ['l10n_in.tax_tag_eco_9_5'],
+        }
         return {
-            'sgst': get_tag_ids('l10n_in.tax_tag_base_sgst', 'l10n_in.tax_tag_sgst'),
-            'cgst': get_tag_ids('l10n_in.tax_tag_base_cgst', 'l10n_in.tax_tag_cgst'),
-            'igst': get_tag_ids('l10n_in.tax_tag_base_igst', 'l10n_in.tax_tag_igst'),
-            'cess': get_tag_ids('l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'),
-            'eco_9_5': get_tag_ids('l10n_in.tax_tag_eco_9_5'),
+            categ: [xmlid_to_res_id(xml_id) for xml_id in ref]
+            for categ, ref in tag_refs.items()
         }
 
     def _get_l10n_in_gstr_section(self, tax_tags_dict):


addons/l10n_in/models/account_invoice.py
```diff
@@ -721,6 +721,8 @@ def _l10n_in_edi_get_iap_buy_credits_message(self):
 
     def _get_sync_stack(self, container):
         stack, update_containers = super()._get_sync_stack(container)
+        if 'IN' not in self.mapped('country_code'):
+            return stack, update_containers
```
hamo-odoo: @jva-odoo @jasa-odoo, why didn't we check for the Indian moves? I'm not sure what's best to check `self` or the container here?

addons/l10n_in/models/account_move_line.py
```diff
@@ -82,17 +82,19 @@ def _l10n_in_check_invalid_hsn_code(self):
...
+            'cess': ['l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'],
+            'eco_9_5': ['l10n_in.tax_tag_eco_9_5'],
+        }
         return {
-            'sgst': get_tag_ids('l10n_in.tax_tag_base_sgst', 'l10n_in.tax_tag_sgst'),
-            'cgst': get_tag_ids('l10n_in.tax_tag_base_cgst', 'l10n_in.tax_tag_cgst'),
-            'igst': get_tag_ids('l10n_in.tax_tag_base_igst', 'l10n_in.tax_tag_igst'),
-            'cess': get_tag_ids('l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'),
-            'eco_9_5': get_tag_ids('l10n_in.tax_tag_eco_9_5'),
+            tax: [xmlid_to_res_id(id) for id in ref]
```
hamo-odoo: id is a keyword
```suggestion
            tax: [xmlid_to_res_id(xml_id) for xml_id in ref]
```

addons/l10n_in/models/account_invoice.py
jasa-odoo: Hello,
I believe checking `self` here is a better approach because `container` is just a sync structure (dict) derived from `self`

addons/l10n_in/models/account_invoice.py
```diff
@@ -721,6 +721,8 @@ def _l10n_in_edi_get_iap_buy_credits_message(self):
 
     def _get_sync_stack(self, container):
         stack, update_containers = super()._get_sync_stack(container)
+        if 'IN' not in self.mapped('country_code'):
```
hamo-odoo:
```suggestion
        if all(move.country_code != 'IN' for move in self):
```
Reason: using l10n_in with a higher probability of Indian moves, this will make fewer iterations, rather using mapped will always iterate all the records

hamo-odoo (approved):
LGTM

addons/l10n_in/models/account_move_line.py
```diff
@@ -82,17 +82,19 @@ def _l10n_in_check_invalid_hsn_code(self):
...
+            'cess': ['l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'],
+            'eco_9_5': ['l10n_in.tax_tag_eco_9_5'],
+        }
         return {
-            'sgst': get_tag_ids('l10n_in.tax_tag_base_sgst', 'l10n_in.tax_tag_sgst'),
-            'cgst': get_tag_ids('l10n_in.tax_tag_base_cgst', 'l10n_in.tax_tag_cgst'),
-            'igst': get_tag_ids('l10n_in.tax_tag_base_igst', 'l10n_in.tax_tag_igst'),
-            'cess': get_tag_ids('l10n_in.tax_tag_base_cess', 'l10n_in.tax_tag_cess'),
-            'eco_9_5': get_tag_ids('l10n_in.tax_tag_eco_9_5'),
+            tax: [xmlid_to_res_id(xml_id) for xml_id in ref]
```
jco-odoo: tax is confusing here: tax_type?

addons/l10n_in/models/account_move_line.py
```diff
@@ -82,17 +82,19 @@ def _l10n_in_check_invalid_hsn_code(self):
             )
         return False
 
+    @tools.ormcache()
```
jco-odoo: xml_id_lookup is already cached as well...

jco-odoo: Although this is also the case for default_uom_id e.g. on the product

addons/l10n_in/models/account_move_line.py
jco-odoo: Maybe categ as further on in the code we also use category.

hash-odoo: ~~I think it can be `tax_tag` or `tag_id`, as those records belong to `account.account.tag`~~

We can have categ, that would be more consistent.

addons/l10n_in/models/account_move_line.py
william-andre:
```suggestion
```

william-andre (commented):
It makes sense since we don't care if the ids still exist in the db in this case because we are just trying to compare.

addons/l10n_in/models/account_invoice.py
```diff
@@ -721,6 +721,8 @@ def _l10n_in_edi_get_iap_buy_credits_message(self):
 
     def _get_sync_stack(self, container):
         stack, update_containers = super()._get_sync_stack(container)
+        if all(move.country_code != 'IN' for move in self):
```
william-andre, line 724: I personally find this easier to understand, maybe it's just me 🤷 
```suggestion
        if not any(move.country_code == 'IN' for move in self):
```

hash-odoo: I personally find the one with all() shorter.

Plus, why use 'not' and make it negative. Shouldn't we think positive? :thought_balloon:

addons/l10n_in/models/account_move_line.py
hash-odoo: Right, initially the method had multiple entries in the profiler. But `_xmlid_lookup` manages it. Will remove it :+1:

addons/l10n_in/models/account_invoice.py
william-andre: It's just more natural to me to think "if there aren't any indian moves, return" than "if all moves are not indian, return".
It's also the same logic as elsewhere with things like `if self.country_code != 'IN': return`

But ok, I won't block for that

william-andre (approved):
@robodoo r+

PR #248964, opened 2026-02-17
[FIX] html_editor: fix toolbar position on selected table

adsa-odoo:

Steps to reproduce:

- Create a 3 x 3 table.
- Select 3rd column and wait for toolbar.

Currently, the toolbar is positioned at the start of the table even when the last column is selected. This happens because selecting cells in the 3rd column creates a DOM selection range that starts at the first cell and ends at the last cell of the column, traversing all intermediate elements. As a result, browser's range rectangle does not match the actual custom-selection rect.

Now, Toolbar is positioned correctly above the custom selected cells.

task-5935587

---

### DIFF
--- a/addons/html_editor/static/src/core/overlay.js
+++ b/addons/html_editor/static/src/core/overlay.js
@@ -29,6 +29,7 @@ export class EditorOverlay extends Component {
         history: Object,
         close: Function,
         isOverlayOpen: Function,
+        getCustomRect: { type: Function, optional: true },
 
         // Props from createOverlay
         positionOptions: { type: Object, optional: true },
@@ -122,7 +123,10 @@ export class EditorOverlay extends Component {
             }
             range = this.lastSelection.range;
         }
-        let rect = range.getBoundingClientRect();
+        let rect =
+            this.props.getCustomRect?.() ||
+            this.lastSelection.rect ||
+            range.getBoundingClientRect();
         if (rect.x === 0 && rect.width === 0 && rect.height === 0) {
             // Attention, using disableObserver and enableObserver is always dangerous (when we add or remove nodes)
             // because if another mutation uses the target that is not observed, that mutation can never be applied
--- a/addons/html_editor/static/src/core/overlay_plugin.js
+++ b/addons/html_editor/static/src/core/overlay_plugin.js
@@ -31,6 +31,7 @@ export class OverlayPlugin extends Plugin {
             this.container = this.getScrollContainer();
         });
         this.addDomListener(this.topDocument.defaultView, "resize", this.throttledUpdateContainer);
+        this.targetRectProviders = this.getResource("overlay_selection_target_rect_providers");
     }
 
     destroy() {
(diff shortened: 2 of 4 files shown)


addons/html_editor/static/src/core/overlay_plugin.js
```diff
@@ -137,4 +139,14 @@ export class Overlay {
     updatePosition() {
         this.bus.trigger("updatePosition");
     }
+
+    getCustomSelectionRange() {
```
deso-odoo: This should belong to `selectionPlugin`.

addons/html_editor/static/src/core/overlay.js
```diff
@@ -114,7 +115,7 @@ export class EditorOverlay extends Component {
         const inEditable = this.props.editable.contains(selection.anchorNode);
         let range;
         if (inEditable) {
-            range = selection.getRangeAt(0);
+            range = this.props.getCustomSelectionRange?.() || selection.getRangeAt(0);
```
dmo-odoo: Since we don't have a usecase for the custom range in selection itself, I suggest introducing a new `overlay_selection_target_rect_providers` resource instead. The table plugin can use it to provide a rect directly for the overlay plugin to use at https://github.com/odoo/odoo/pull/248964/changes#diff-32cfddd25bd0af057c1370930957fd04c61c8ec008a5b9234226ec4237833d9fR144 without the need for using an intermediate fake range at all.

addons/html_editor/static/src/core/overlay_plugin.js
```diff
@@ -94,8 +104,9 @@ export class Overlay {
             const selection = this.plugin.editable.ownerDocument.getSelection();
             let initialSelection;
             if (selection && selection.type !== "None") {
+                const rect = this.plugin.getCustomRect();
                 initialSelection = {
-                    range: selection.getRangeAt(0),
+                    range: rect ? { getBoundingClientRect: () => rect } : selection.getRangeAt(0),
```
dmo-odoo: This still uses a fake range which only has `getBoundingClientRect` available. Is the range actually used for anything else than the rect? I suppose not, otherwise this implementation would have an issue. Can't we replace `initialSelection` by `initialRect`?

adsa-odoo: The range passed here is used at https://github.com/odoo/odoo/blob/18.0/addons/html_editor/static/src/core/overlay.js#L123 . There is one case where this range can be used that is range.cloneRange https://github.com/odoo/odoo/blob/18.0/addons/html_editor/static/src/core/overlay.js#L130-L132 . In our case this `if` condition will not be true that's why it is working fine but if there might be a case where even custom rect has x, y and width 0 and `inEditable` is false, we will have problem. Do you think we should patch the getBoundingClientRect of original range here? Something like:
```
const range = selection.getRangeAt(0)
if (rect) {
range.getBoundingClientRect = () => rect;
}
initialSelection = { range }
```

dmo-odoo: Can't we add `rect` to the `initialSelection` object? Then instead of doing `range.getBoundingClientRect()` we could do `const rect = this.lastSelection.rect ?? range.getBoundingClientRect()` without having to override the native Range object.

adsa-odoo: I've made the changes. Can you please have a look?

PR #240555, opened 2025-12-18
[IMP] web: beautiful duration

bastien-ipb:
In this commit I change the format of the durations (float_time). Now they are displayed as follow by default: 12h 30m 25s You can also specify the unit of time, like that you can specify to float_time what value it's reading. For example, if the unit of time is "minutes" and that you give it 30, it will translated to 30m. The value set to 0 are removed from the result. 0h 30m 0s => 30m

There is also a bottom popover on float_time fields that display how the typed value will be interpreted.

The format hh:mm:ss can be forced by putting the option "numeric" to true. The seconds are shown by default, they can be hide by putting the option "showSeconds" to false.

TASK-5347051

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #248966, opened 2026-02-17
[FIX] hr_holidays: fix invalid time display for hourly time off

nikj-odoo:
**Issue**:
- When users selected time off based on hours, "InvalidDateTime" was shown instead of the correct "From" and "To" time.
This happened after a recent
refactor:https://github.com/odoo/odoo/pull/240555/changes#diff-0beecbb730e1c7762f2a8d5f7038c02e123da10a85dec2e5d16a02748373f2fe of getFormatValue, which now returns
DateTime in a different format than before.

Fix:
- Updated  the method and adapt the new refactored format.

Impact:
- The correct DateTime values are now displayed instead of "InvalidDateTime".

Task-5940185


tivisse:
@robodoo r+

PR #248967, opened 2026-02-17
[FIX] website: handle menu URLs without leading slash and warn on spaces

aans-odoo:
When a menu item URL contains a space and has no leading slash (e.g.,
"some url"), creating a page from its 404 screen does not link the menu
to the newly created page. The menu keeps pointing to a 404.

**Steps to reproduce**:
1. Create a menu item with a URL containing spaces and no leading slash
   (e.g., "some url").
2. Click the menu item -> a 404 page is displayed (expected).
3. Click "Create Page" -> the page is created and saved.
4. Click the menu item again -> it still returns a 404 (unexpected).

**Issue**:
During page creation, the path is slugified (e.g., "some url" ->
"/some-url"). The controller then tries to link the menu to the new page
by setting `page_id`. However, `pagenew()` only searches for menu URLs
with a leading slash, so a menu saved as "some url" is not found when
searching for "/some url". As a result, `page_id` is never set, and
`_clean_url()` keeps resolving the menu to "/some url" instead of the
page's actual URL "/some-url".

Menus without spaces (e.g., "mypage") have the same issue where
`page_id` is not set, but since `_clean_url()` prepends "/" and renders
"/mypage", which matches the new page URL, the menu still resolves
correctly.

**Fix**:
Match both URL formats (with and without leading slash) when linking a
menu to a newly created page.

Additionally, a warning is now shown in the menu editor when a URL
contains spaces, since spaces are slugified during page creation,
which could cause confusion. (Note: this is only a warning - URLs with
spaces are still allowed.)

task-[5095646](https://www.odoo.com/odoo/project/974/tasks/5095646)

---

### DIFF
--- a/addons/website/controllers/main.py
+++ b/addons/website/controllers/main.py
@@ -666,7 +666,10 @@ class Website(Home):
         # If that URL is also a menu, we update it accordingly.
         # NB: we don't want to slugify on menu creation as it could redirect
         # towards files (with spaces, apostrophes, etc.).
-        menu = request.env['website.menu'].search([('url', '=', '/' + path), ('page_id', '=', False)])
+        # When searching for a menu, we also match URLs with or without a
+        # leading slash to prevent mismatches when records were created without
+        # a leading slash.
+        menu = request.env['website.menu'].search([('url', 'in', ['/' + path, path]), ('page_id', '=', False)])
         if menu:
             menu.page_id = page['page_id']
 
--- a/addons/website/static/src/components/dialog/edit_menu.xml
+++ b/addons/website/static/src/components/dialog/edit_menu.xml
@@ -17,6 +17,11 @@
             <div class="col-md-9">
                 <input id="url_input" t-ref="url-input" type="text" t-model="url.input.value" class="form-control" t-att-class="{ 'is-invalid': url.input.hasError }"/>
                 <small class="form-text">Hint: Type '/' to search an existing page and '#' to link to an anchor.</small>
+                <!-- TODO (19.0): Remove this warning once URLs are auto-filled
+                     from a slugified title. -->
+                <small class="text-warning d-block" t-att-class="{'invisible': !url.input.value?.includes(' ') || new RegExp('^[a-z]+:', 'i').test(url.input.value)}">
+                    Spaces are not allowed in URL.
+                </small>
             </div>
         </div>
     </WebsiteDialog>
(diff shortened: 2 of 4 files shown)


addons/website/controllers/main.py
```diff
@@ -666,7 +666,12 @@ def pagenew(self, path="", add_menu=False, template=False, redirect=False, **kwa
         # If that URL is also a menu, we update it accordingly.
         # NB: we don't want to slugify on menu creation as it could redirect
         # towards files (with spaces, apostrophes, etc.).
-        menu = request.env['website.menu'].search([('url', '=', '/' + path), ('page_id', '=', False)])
+        # When searching for a menu, we also match URLs with or without a
+        # leading slash to prevent mismatches when records were created without
+        # a leading slash.
+        menu = request.env['website.menu'].search(
+            [('url', 'in', ['/' + path, path]), ('page_id', '=', False)], limit=1
```
blse-odoo: It's a bit of an edge case, but it used to update all matching menus, not just the first one found 🤔 Is there a particular reason to add the `limit=1`?
(if there is none, I think it would be best to not add it to keep only the needed changes)

addons/website/static/src/components/dialog/edit_menu.xml
```diff
@@ -17,6 +17,9 @@
             <div class="col-md-9">
                 <input id="url_input" t-ref="url-input" type="text" t-model="url.input.value" class="form-control" t-att-class="{ 'is-invalid': url.input.hasError }"/>
                 <small class="form-text">Hint: Type '/' to search an existing page and '#' to link to an anchor.</small>
+                <small class="text-warning d-block" t-att-class="{'invisible': !url.input.value?.includes(' ')}">
+                    Spaces are not allowed in URL.
+                </small>
```
blse-odoo: From the [discussion in the task](https://www.odoo.com/mail/message/925568727), this warning should only appear for relative URLs with a space (not all URLs with a space) 🙂

blse-odoo (commented):
Thanks for your PR 👍
I had a discussion with lebl, the warning would not be needed in the forward ports starting from the version where the url is auto-filled with a slugified version of the title 🙂

addons/website/controllers/main.py
aans-odoo: I kept `limit=1` to ensure only the first result, but as you pointed out, that's opposite to the existing logic! I’ve reverted it.

aans-odoo:
Hello @blse-odoo :wave: thanks for the review, I've made the changes.
Rebasing the branch as runbot is failing because of it.

FrancoisGe:
@robodoo delegate=blse-odoo

blse-odoo:
@robodoo r+

PR #248970, opened 2026-02-17
[IMP] tests: configurable timeout for web calls

kmagusiak:
When debugging, routes opened during testing timeout after 12 seconds.
We add ODOO_TEST_DISABLE_TIMEOUT environment variable to overwrite the
default values. This allows to use breakpoints in the called URL and
avoid the timeout.

---

### DIFF
--- a/odoo/tests/common.py
+++ b/odoo/tests/common.py
@@ -73,7 +73,7 @@ from odoo.modules.registry import Registry
 from odoo.sql_db import Cursor
 from odoo.tools import SQL, DotDict, config, float_compare, mute_logger, profiler
 from odoo.tools.mail import single_email_re
-from odoo.tools.misc import diff_zip, find_in_path, lower_logging
+from odoo.tools.misc import diff_zip, find_in_path, lower_logging, str2bool
 from odoo.tools.xml_utils import _validate_xml
 
 import odoo.addons.base
@@ -114,6 +114,8 @@ CHECK_BROWSER_ITERATIONS = 100
 BROWSER_WAIT = CHECK_BROWSER_SLEEP * CHECK_BROWSER_ITERATIONS # seconds
 DEFAULT_SUCCESS_SIGNAL = 'test successful'
 TEST_CURSOR_COOKIE_NAME = 'test_request_key'
+
+DISABLE_TIMEOUTS = str2bool(os.getenv('ODOO_TEST_DISABLE_TIMEOUT', '0'))
 
 IGNORED_MSGS = re.compile(r"""
     failed\ to\ fetch  # base error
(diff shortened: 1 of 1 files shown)


odoo/tests/common.py
```diff
@@ -114,6 +114,9 @@
 BROWSER_WAIT = CHECK_BROWSER_SLEEP * CHECK_BROWSER_ITERATIONS # seconds
 DEFAULT_SUCCESS_SIGNAL = 'test successful'
 TEST_CURSOR_COOKIE_NAME = 'test_request_key'
+DEFAULT_WEB_TIMEOUT = float(os.getenv('ODOO_TEST_WEB_TIMEOUT', '0')) or 12
+DEFAULT_JS_TIMEOUT = DEFAULT_WEB_TIMEOUT * 5
+assert DEFAULT_WEB_TIMEOUT > 0
```
Julien00859: "AssertionError: False is not True" is a bad error message,

```suggestion
assert DEFAULT_WEB_TIMEOUT > 0, "bad ODOO_TEST_WEB_TIMEOUT environment variable"
```

odoo/tests/common.py
```diff
@@ -114,6 +114,9 @@
 BROWSER_WAIT = CHECK_BROWSER_SLEEP * CHECK_BROWSER_ITERATIONS # seconds
 DEFAULT_SUCCESS_SIGNAL = 'test successful'
 TEST_CURSOR_COOKIE_NAME = 'test_request_key'
+DEFAULT_WEB_TIMEOUT = float(os.getenv('ODOO_TEST_WEB_TIMEOUT', '0')) or 12
```
Julien00859: There must be as many options as there are different values. `12` is only the timeout for `url_open`, the other functions use `10` and `20`. All those values were carefully picked by the runbot team and are individually maintained.

Julien00859:
```suggestion
_DEFAULT_WEB_TIMEOUT = float(os.getenv('ODOO_TEST_WEB_TIMEOUT', '0')) or 12
```
As they are used as default value for the functions below, importing and changing those global variable will have no impact on the functions, the functions will still use the original value. Let's make them private, so it is a bit more explicit that it is useless to change them.

I would even advocate we don't use global variables, and instead have the `getenv` on the function definition.

Julien00859 (changes requested):

odoo/tests/common.py
kmagusiak: I have added a second timeout variable depending on the first timeout.
I will need the input of @Xavier-Do to check how many variables we need exactly as I am not sure all of them need to have a distinct value. The goal of the PR is also to be able to give a single environment variable to overwrite all timeouts for testing.

Julien00859: I'm fine with the current proposal

kmagusiak:
After discussion with @Xavier-Do, simplifying to a single env variable that disables timeouts.

Julien00859 (approved):

PR #248974, opened 2026-02-17
[FIX] mass_mailing: preserve changes when switching form tabs

walidsahli:
Problem:
While editing an email marketing record, switching tabs in the form view causes the last changes in the editor to be lost.

Cause:
The `blur` event on the iframe is not triggered when clicking outside of it. If the iframe is not fully ready when listeners are attached, the required events are not properly registered, so the latest changes are not saved before switching tabs.

Solution:
Attach the `blur` listener only once the iframe is fully loaded, ensuring that changes are correctly detected and saved when focus is lost.

Steps to reproduce:
- Create a new email marketing record.
- Add a text snippet.
- Change the text and save.
- Modify the text again.
- Switch to another tab in the form view, then return to the "Mail body" tab.
- Observe that the last changes are lost.

opw-5921800

---

### DIFF
--- a/addons/mass_mailing/static/src/iframe/mass_mailing_iframe.js
+++ b/addons/mass_mailing/static/src/iframe/mass_mailing_iframe.js
@@ -253,6 +253,7 @@ export class MassMailingIframe extends Component {
         this.iframeRef.el.contentWindow.addEventListener("beforeUnload", () => {
             this.iframeRef.el.removeAttribute("is-ready");
         });
+        this.iframeRef.el.contentWindow.addEventListener("blur", this.onBlur.bind(this));
         this.iframeLoaded.resolve({
             iframe: this.iframeRef.el,
             bundleControls: this.bundleControls,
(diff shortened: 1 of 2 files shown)


addons/mass_mailing/static/src/iframe/mass_mailing_iframe.js
```diff
@@ -253,6 +253,7 @@ export class MassMailingIframe extends Component {
         this.iframeRef.el.contentWindow.addEventListener("beforeUnload", () => {
             this.iframeRef.el.removeAttribute("is-ready");
         });
+        this.iframeRef.el.contentWindow.addEventListener("blur", this.props.onBlur);
```
abd-msyukyu-odoo: we should use the `onBlur` function of the component, not the props directly, because it is not needed in readonly mode

abd-msyukyu-odoo (commented):
Heya, thanks for the fix, added a small comment, but should be good to go otherwise

addons/mass_mailing/static/src/iframe/mass_mailing_iframe.js
walidsahli: Indeed, updated!

addons/mass_mailing/static/src/iframe/mass_mailing_iframe.js
```diff
@@ -253,6 +253,7 @@ export class MassMailingIframe extends Component {
         this.iframeRef.el.contentWindow.addEventListener("beforeUnload", () => {
             this.iframeRef.el.removeAttribute("is-ready");
         });
+        this.iframeRef.el.contentWindow.addEventListener("blur", this.onBlur);
```
abd-msyukyu-odoo: Also, I think you should bind the function to `this` for it to work, or use an arrow function, else `this` is undefined?

walidsahli: my bad yes, Thanks!

PR #248978, opened 2026-02-17
[FIX] base: resetting crons timeout_counter on timeout

abib-odoo:
CronJobs timed_out_counter is not getting reset when reaching the CONSECUTIVE_TIMEOUT_FOR_FAILURE due to a transaction rollback. This commit introduces a fix to ensure only records with an actual exception tuple pass through in method: 'method_direct_trigger'.

When a cron job reaches the timeout threshold (timed_out_counter >= 3), method_direct_trigger crashes because the ListLogHandler filter matches log records that have exc_info = None. But the original filter only checked hasattr(lr, 'exc_info') which is true for all LogRecord objects. This caused a TypeError when attempting to unpack exc_info (None) on the next line, which rolled back the transaction and prevented timed_out_counter from being reset.

Current behavior before PR:
Clicking "Run Manually" on a timed-out cron (timed_out_counter >= 3) raises TypeError: cannot unpack non-iterable NoneType object, rolling back the transaction. The timed_out_counter is never reset, leaving the cron permanently stuck.

The filter correctly skips log records where exc_info is None (i.e., non-exception errors like timeouts). method_direct_trigger returns True, the transaction commits, and timed_out_counter is properly reset to 0.

---

### DIFF
--- a/odoo/addons/base/models/ir_cron.py
+++ b/odoo/addons/base/models/ir_cron.py
@@ -167,7 +167,7 @@ class IrCron(models.Model):
 
         with ListLogHandler(_logger, logging.ERROR) as capture:
             self._process_job(cron_cr, job)
-        if log_record := next((lr for lr in capture if hasattr(lr, 'exc_info')), None):
+        if log_record := next((lr for lr in capture if getattr(lr, 'exc_info', None)), None):
             _exc_type, exception, _traceback = log_record.exc_info
             e = RuntimeError()
             e.__cause__ = exception


odoo/addons/base/models/ir_cron.py
```diff
@@ -167,8 +167,8 @@ def method_direct_trigger(self):
 
         with ListLogHandler(_logger, logging.ERROR) as capture:
             self._process_job(cron_cr, job)
-        if log_record := next((lr for lr in capture if hasattr(lr, 'exc_info')), None):
-            _exc_type, exception, _traceback = log_record.exc_info
+        if log_record := next((lr for lr in capture if hasattr(lr, 'exc_info') and lr.exc_info), None):
+            _, exception, _ = log_record.exc_info
```
kmagusiak: Nitpick: Why this line?

odoo/addons/base/tests/test_ir_cron.py
```diff
@@ -685,6 +685,39 @@ def mocked_run(self):
...
+            'done': 0,
+            'timed_out_counter': 3,
+        }])
+        self.env.flush_all()
+
+        with (
+            self.enter_registry_test_mode(),
+            mute_logger('odoo.addons.base.models.ir_cron'),
+        ):
+            # wrapping this in a try/except block to catch the exc_info bug, to continue assert statements execution
```
kmagusiak: The try-except should be in the ir_cron file. The expected behaviour is that `method_direct_trigger` does not crash.

kmagusiak (changes requested):

odoo/addons/base/models/ir_cron.py
abib-odoo: @kmagusiak Mix of a personal preference and technical reasoning, I don't like to keep unused variables, it can clutter up the code quickly

also since these variables are not _typed_ nor referenced anywhere, keeping them doesn't really provide any additional structural information, they just signal that they are unused.

that said I can undo this for consistency if that's the team preference.

odoo/addons/base/tests/test_ir_cron.py
abib-odoo: @kmagusiak You have a point, the reasoning behind the try/except in the test was to ensure the assertions still execute even without the fix (the assertions weren't executed because of the TypeError), so instead of a raw TypeError crash, we'd get a meaningful assertion failure like "timed_out_counter should be reset to 0". 
But I agree it's not ideal to have defensive handling there.

I think we can drop the try/except from the test entirely and instead strengthen the fix in method_direct_trigger with a stricter type check that its a tuple of 3 elements, and the variable we are using
is of a BaseException type

```
if log_record := next((
    lr for lr in capture
    if isinstance(lr.exc_info, tuple)
    and len(lr.exc_info) == 3
    and isinstance(lr.exc_info[1], BaseException)
), None):
```

then we won't need any try-catch unless we want to always reset the counter no matter what the exception is (could be bad since other problems can arise and we won't be able to know)

what do you think?

odoo/addons/base/models/ir_cron.py
kmagusiak: Nah, it's fine. (My preference is just to name them keeping the _ prefix to mark as unused.)

odoo/addons/base/models/ir_cron.py
```diff
@@ -167,8 +167,13 @@ def method_direct_trigger(self):
 
         with ListLogHandler(_logger, logging.ERROR) as capture:
             self._process_job(cron_cr, job)
-        if log_record := next((lr for lr in capture if hasattr(lr, 'exc_info')), None):
-            _exc_type, exception, _traceback = log_record.exc_info
+        if log_record := next((
```
kmagusiak: I think that it would be simpler to read if written as:
```py
for log_record in capture:
  try:
    exception = hasattr(log_record, 'exc_info') and log_record.exc_info[1]
    if not isinstance(exception, BaseException):
      continue
  except Exception:
    continue
  e = RuntimeError()
  ...
```

kmagusiak (commented):
@Julien00859 Can you have a look?

abib-odoo:
> What is `exc_info` when it is not a 3-values tuple?
> 
> * None, and other falsy values, can be ignored (we can do that changing the `hasattr` for a `getattr` with a `None` fallback).
> * True would be annoying, usually we should use `sys.exc_info` but since the exception occurred in another thread we have no way to get it :(

from what I can tell in the python Logger class inside `_log`, exc_info is always None or a tuple of 3 elements based on the code:

```
def _log(self, level, msg, args, exc_info=None, extra=None, stack_info=False,
             stacklevel=1):
        """
        Low-level logging routine which creates a LogRecord and then calls
        all the handlers of this logger to handle the record.
        """
        ...
        if exc_info:
            if isinstance(exc_info, BaseException):
                exc_info = (type(exc_info), exc_info, exc_info.__traceback__)
            elif not isinstance(exc_info, tuple):
                exc_info = sys.exc_info()
        record = self.makeRecord(self.name, level, fn, lno, msg, args,
                                 exc_info, func, extra, sinfo)
        self.handle(record)
```

Possible Conversions with different arguments:
1-None: we skip it so it's None on the LogRecord
2- Non-Faulty value that is not an Exception nor a Tuple: sys.exc_info()
3- BaseException: convert it to a tuple
4- Tuple: passed as it is

regardless of how we re-write the fix for readability, I think its safe to keep the checks and maybe add back the
`hasattr` check in case there is a version of python that doesn't has it or if its removed in the future

@kmagusiak

abib-odoo:
> ```
>         if exc_info:
>             if isinstance(exc_info, BaseException):
>                 exc_info = (type(exc_info), exc_info, exc_info.__traceback__)
>             elif not isinstance(exc_info, tuple):
>                 exc_info = sys.exc_info()
> ```
> 
> so it always is either falsy, either a valid tuple. Let's change the `hasattr` for a `getattr` with a `None` fallback then. 1 line change. No need to bother with a test.

To be honest I want to keep the type check, because we're creating `RuntimeException` variable and assigning the `__cause__` attribute to the exception from the tuple, this attribute has a type check on it

builtin.py
```
@disjoint_base
class BaseException:
    args: tuple[Any, ...]
    __cause__: BaseException | None
    __context__: BaseException | None

```
which will fail with `TypeError: exception cause must be None or derive from BaseException`

yes the exc_info will always be None or a tuple but there was no check whether its a valid tuple or not, passing a tuple with any data type will pass and eventually fail in the `method_direct_trigger`

@kmagusiak

odoo/addons/base/models/ir_cron.py
Julien00859: We have a minimal diff policy in stable, don't change code unless you need to.

kmagusiak (approved):
@robodoo r+

Thanks for your work 👍

PR #248979, opened 2026-02-17
[FIX] hr_attendance: hide archived employees

YoussefM890:
**Problem:**
When an employee is archived, their attendance records still appear in the attendance list and reporting views.

**Steps to reproduce:**
1. Create an employee and record some attendances
2. Archive the employee
3. Go to Attendances > Overview
4. Observe the archived employee still appears in the list

**Current behavior:**
Archived employees remain visible in all attendance views.

**Expected behavior:**
Archived employees should be hidden by default, with an option to view them via search filters.

**Cause of the issue:**
The hr.attendance model has no `active` field itself, so the ORM never auto-filters records of archived employees. The list view loads all attendance records regardless of the employee's active status.

**Fix:**
A custom JS list model injects `employee_id.active = True` into the domain on every load, filtering out archived employees by default. Search filters allow toggling between active and archived employees when needed.

Backport of: 3d6d89a49a9d168eb87587f82c33f279a4ff22a3

opw-5930149

### DIFF
--- a/addons/hr_attendance/static/src/views/attendance_list_view.js
+++ b/addons/hr_attendance/static/src/views/attendance_list_view.js
@@ -1,0 +1,25 @@
+/** @odoo-module **/
+
+import { registry } from "@web/core/registry";
+import { listView } from "@web/views/list/list_view";
+
+export class AttendanceListModel extends listView.Model {
+
+    /** @override **/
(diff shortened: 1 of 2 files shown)


taqu-odoo (commented):
LGTM but not sure if it's worth fixing in 17

PR #248984, opened 2026-02-17
[FIX] hr_timesheet: Sort timesheet lines in PDF

VincentJanssen-Code:
## Behavior before the PR:
The printed version of the timesheet always displayed the account_analytic_lines in the same order as in the view. This meant that if a user added a new line to the timesheet and then printed it, the order of the lines could be incorrect.

## Expected behavior after the PR:
The timesheet is automatically sorted before being printed, ensuring the correct order of lines.

task-[5349814](https://www.odoo.com/odoo/project/4105/tasks/5349814)

---

### DIFF
--- a/addons/hr_timesheet/report/report_timesheet_templates.xml
+++ b/addons/hr_timesheet/report/report_timesheet_templates.xml
@@ -24,7 +24,7 @@
                         </tr>
                    </thead>
                    <tbody>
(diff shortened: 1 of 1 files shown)


deneuvillem (commented):
LGTM, thanks for your improvement!

robodoo r+

